PR #4208, opened 2021-01-08
Fix TO trying to create internal topic

sknot-rh:

### Type of change
- Bugfix

### Description
TO is trying to create internal topics. That produces errors like this 
```
2021-01-08 08:21:27,45461 WARN  [vert.x-eventloop-thread-0] ZkTopicsWatcher:153 - 6|/brokers/topics 3:+__consumer_offsets: Error responding to creation of topic __consumer_offsets
org.apache.kafka.common.errors.UnknownTopicOrPartitionException:
```

### Checklist

- [ ] Write tests
- [ ] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/topic-operator/src/main/java/io/strimzi/operator/topic/ZkTopicsWatcher.java
+++ b/topic-operator/src/main/java/io/strimzi/operator/topic/ZkTopicsWatcher.java
@@ -146,13 +146,17 @@ class ZkTopicsWatcher {
                     tcw.addChild(topicName);
                     tw.addChild(topicName);
                     LogContext logContext = LogContext.zkWatch(TOPICS_ZNODE, watchCount + ":+" + topicName);
-                    topicOperator.onTopicCreated(logContext, new TopicName(topicName)).onComplete(ar -> {
-                        if (ar.succeeded()) {
-                            LOGGER.debug("{}: Success responding to creation of topic {}", logContext, topicName);
-                        } else {
-                            LOGGER.warn("{}: Error responding to creation of topic {}", logContext, topicName, ar.cause());
-                        }
-                    });
+                    if ("__consumer_offsets".equals(topicName) || "__transaction_state".equals(topicName)) {
+                        LOGGER.debug("{}: Cannot create internal topic {}", logContext, topicName);
+                    } else {
+                        topicOperator.onTopicCreated(logContext, new TopicName(topicName)).onComplete(ar -> {
+                            if (ar.succeeded()) {
+                                LOGGER.debug("{}: Success responding to creation of topic {}", logContext, topicName);
+                            } else {
+                                LOGGER.warn("{}: Error responding to creation of topic {}", logContext, topicName, ar.cause());
+                            }
+                        });
+                    }
                 }
             }
         }
--- a/topic-operator/src/main/java/io/strimzi/operator/topic/TopicOperator.java
+++ b/topic-operator/src/main/java/io/strimzi/operator/topic/TopicOperator.java
@@ -940,6 +940,9 @@ class TopicOperator {
                 });
             }
         };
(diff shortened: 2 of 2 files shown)


topic-operator/src/main/java/io/strimzi/operator/topic/ZkTopicsWatcher.java
```diff
@@ -146,13 +146,17 @@ public void handle(AsyncResult<List<String>> childResult) {
...
                     tw.addChild(topicName);
                     LogContext logContext = LogContext.zkWatch(TOPICS_ZNODE, watchCount + ":+" + topicName);
-                    topicOperator.onTopicCreated(logContext, new TopicName(topicName)).onComplete(ar -> {
-                        if (ar.succeeded()) {
-                            LOGGER.debug("{}: Success responding to creation of topic {}", logContext, topicName);
-                        } else {
-                            LOGGER.warn("{}: Error responding to creation of topic {}", logContext, topicName, ar.cause());
-                        }
-                    });
+                    if ("__consumer_offsets".equals(created)) {
```
ppatierno, line 149: I think we should be more general here. AFAIK there is a `__transaction_state` internal topic as well for tracking transaction state. And what about the topics created by Kafka Connect and Kafka streams API? @tombentley wdyt?

topic-operator/src/main/java/io/strimzi/operator/topic/ZkTopicsWatcher.java
```diff
@@ -146,13 +146,17 @@ public void handle(AsyncResult<List<String>> childResult) {
...
                     LogContext logContext = LogContext.zkWatch(TOPICS_ZNODE, watchCount + ":+" + topicName);
-                    topicOperator.onTopicCreated(logContext, new TopicName(topicName)).onComplete(ar -> {
-                        if (ar.succeeded()) {
-                            LOGGER.debug("{}: Success responding to creation of topic {}", logContext, topicName);
-                        } else {
-                            LOGGER.warn("{}: Error responding to creation of topic {}", logContext, topicName, ar.cause());
-                        }
-                    });
+                    if ("__consumer_offsets".equals(topicName) || "__transaction_state".equals(topicName)) {
+                        LOGGER.debug("{}: Cannot create internal topic {}", logContext, topicName);
```
tombentley, line 150: It's not enough to log this (even at INFO or WARN), because the user manipulating the `KafkaTopic` might not have access to the TO logs. We need to something in `status.conditions`. 

It's also not enough to just have this in the `ZkTopicsWatcher`. You can't eliminate the possibility that the topic creation is handled first by the periodic reconciliation. Probably something in `TopicOperator.reconcile` will allow you to make the check in a single place.

topic-operator/src/main/java/io/strimzi/operator/topic/ZkTopicsWatcher.java
tombentley: FTR the issue here is that `__consumer_offsets` and `__transaction_state` are internal in the sense that they can only be created by a broker internally and not directly by any client. Thus Kafka Connect and Kafka Streams "internal" topics are not impacted.

topic-operator/src/main/java/io/strimzi/operator/topic/ZkTopicsWatcher.java
sknot-rh: This exception is thrown in the cluster where no custom `KafkaTopic` resources have been created. What happens is

- kafka creates internal topic `__consumer_offsets`
- TO watches this and creates k8s resource + entry in the ZK
- during that, `UnknownTopicOrPartitionException` is thrown

scholzj:
@ppatierno @tombentley Do you have more comments on this?

PR #4214, opened 2021-01-08
Fix #4213: Bug renaming KafkaTopics to remove the hash

### LINKED ISSUE #4213: Bug renaming KafkaTopics to remove the hash
When a topic in Kafka has a name which can't be used as the name of a `KafkaTopic` in kube, the TO creates the KafkaTopic resources with a hash appended, e.g. `consumer-offsets---84e7a678d08f4bd226872e5cdd4eb527fadc1c6a`

If the user wants to rename this CR, the following steps should work:

1. disable the TO
1. delete the old CR with the name ‘kafka-topic-name’+‘hash’
1. create a CR with the name ‘kafka-topic-name’ (and `spec.topicName`)
1. delete TO internal storage deleteall /strimzi (or just delete the node for the topic concerned)
1. enable the TO

But they don't because when the TO starts in step 5 the topic store JSON incorrectly uses the name with the hash.

tombentley:
### Type of change

- Bugfix

### Description

Fixes #4213.


tombentley:
Still need to add test.

ppatierno (approved):

sknot-rh (approved):

Issue #3904, opened 2020-11-02
[Doc] Update resource values in Cruise Control docs example with more realistic values


kyguy:
**Suggestion / Problem**
As a Java application, Cruise Control can consume a good amount of resources, however the resource `request` and `limits` values listed in the yaml example in the `Deploying Cruise Control` docs [1] are quite low. We should update these values with more realistic resource values so users don't starve their Cruise Control instances!
```
spec:
  cruiseControl:
    resources: (3)
      requests:
        cpu: <UPDATE HERE>
        memory: <UPDATE HERE>
      limits:
        cpu: <UPDATE HERE>
        memory: <UPDATE HERE>
```

**Documentation Link**
[1] https://github.com/strimzi/strimzi-kafka-operator/blob/c836a4ed8cc3e44a544a288fa172a3d255a85006/documentation/modules/cruise-control/proc-deploying-cruise-control.adoc

PR #4219, opened 2021-01-11
Change to more realistic cpu and memory requests and limits

tom1299:
### Type of change
* Documentation

### Description
The values in the documentation of the cruise control deployment description should reflect more realistic values, see issue #3904 as well. This pull request includes the change.

### Checklist
- [ ] Write tests
- [x] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/documentation/modules/cruise-control/proc-deploying-cruise-control.adoc
+++ b/documentation/modules/cruise-control/proc-deploying-cruise-control.adoc
@@ -44,11 +44,11 @@ spec:
       # ...
     resources: <3>
       requests:
-        cpu: 500m
+        cpu: 1
         memory: 512Mi
       limits:
         cpu: 2
-        memory: 1Gi
+        memory: 2Gi
     logging: <4>
         type: inline
         loggers:
--- a/documentation/modules/cruise-control/proc-deploying-cruise-control.adoc
+++ b/documentation/modules/cruise-control/proc-deploying-cruise-control.adoc
@@ -44,11 +44,11 @@ spec:
       # ...
     resources: <3>
       requests:
-        cpu: 200m
-        memory: 64Mi
-      limits:
         cpu: 500m
-        memory: 128Mi
+        memory: 512Mi
+      limits:
+        cpu: 2
+        memory: 1Gi
     logging: <4>
         type: inline
         loggers:


kyguy:
Thanks for the PR @tom1299!

These numbers are certainly better than what we have. If I may ask, how were these new numbers determined and/or under what conditions were these numbers tested (e.g. how much load and how many brokers, partitions etc)? The numbers look good to me for a minimal cluster but I want to justify the new number changes to the maintainers.

tom1299:
To be honest, I haven‘t tested these numbers. I studied the cruise-control configuration [here](https://github.com/linkedin/cruise-control/wiki/Configurations) and found that (besides load / brokers and partitions) memory and cpu consumption depends on the number of threads configured (default is 8 & 1, but probably configured much higher in big clusters). With a default stack size of ~1MB per thread I estimated the minimum memory setting.
But I can‘t provide any numbers that these settings will work for, I‘m afraid. I would need to do more research for that.

tomncooper:
Thanks for this @tom1299, we definitely need more memory, I am not sure how mutli-threaded Cruise Control is but I think we should probably make the request 1 CPU as the model building can be quite a task.

documentation/modules/cruise-control/proc-deploying-cruise-control.adoc
```diff
@@ -44,11 +44,11 @@ spec:
       # ...
     resources: <3>
       requests:
-        cpu: 200m
-        memory: 64Mi
-      limits:
         cpu: 500m
```
tomncooper, line 47: This should probably be 1

ppatierno:
I agree with @tomncooper and @kyguy but to say that these numbers are more realistic we should have a kind of proof for it.

tom1299:
Thank you for your feedback. I took a look at the [start script for CC](https://github.com/linkedin/cruise-control/blob/master/kafka-cruise-control-start.sh#L131) and if not otherwise specified the max. heap memory setting is `1Gb`. So I would think that a resource limit of `1Gb` is indeed a reasonable (although Stack size might actually increase the overall memory consumption). I also asked for help for some reasonable values on the CC glitter forum.

tom1299:
I adjusted the values a little bit. I think that the values for memory `request 512M` and `limit 2G` are in line when using the default values provided in the CC start script (heap space size `xmx=1G`). If the java process exceeds this heap space (for example if the model needs a more space) a `OutOfMemory` exception should be thrown. That leaves ~`1G` for the stack, which actually could accommodate quiet a lot of threads running in parallel.
Regarding CPU, I have to admit that I did not find any means nor do I have runtime performance data to deduct a reasonable value.

documentation/modules/cruise-control/proc-deploying-cruise-control.adoc
tom1299: I updated this value to 1 as you suggested

tomncooper (approved):

tomncooper:
I think we need to do some more extensive testing. However, until we have the opportunity to do that I think these are much more sensible settings.

tomncooper:
@tom1299 you will need to sign off your commit to pass the DCO check. If you look at the DCO bot details it will give you the commands to amend the commit.

kyguy (approved):
Sounds reasonable to me!

Thanks again for looking into this @tom1299!

ppatierno (approved):

scholzj (approved):

Issue #3112, opened 2020-05-28
Use Kafka Admin API for managing user quotas in User Operator


ppatierno:
Once released, we should use the relevant Kafka AdminClient APIs for managing user quotas instead of using ZooKeeper directly. The work is being done as part of [KIP-546](https://cwiki.apache.org/confluence/display/KAFKA/KIP-546%3A+Add+Client+Quota+APIs+to+the+Admin+Client) and should be released with Kafka 2.6.0.

PR #4221, opened 2021-01-11
Removed ZooKeeper dependency and used Admin Client for user quotas

ppatierno:
### Type of change

- Refactoring

### Description

This PR fixes https://github.com/strimzi/strimzi-kafka-operator/issues/3112 actually removing the ZooKeeper dependency from the `KafkaUserQuotasOperator` and using the Admin Client API instead. 

### Checklist

- [x] Write tests

### DIFF
--- a/user-operator/src/main/java/io/strimzi/operator/user/operator/KafkaUserQuotasOperator.java
+++ b/user-operator/src/main/java/io/strimzi/operator/user/operator/KafkaUserQuotasOperator.java
@@ -4,36 +4,34 @@
  */
 package io.strimzi.operator.user.operator;
 
-import com.fasterxml.jackson.databind.JsonNode;
-import com.fasterxml.jackson.databind.ObjectMapper;
-import io.fabric8.zjsonpatch.JsonDiff;
 import io.strimzi.api.kafka.model.KafkaUserQuotas;
 import io.strimzi.operator.common.operator.resource.ReconcileResult;
 import io.vertx.core.Future;
 import io.vertx.core.Promise;
 import io.vertx.core.Vertx;
-import io.vertx.core.json.JsonObject;
-import org.I0Itec.zkclient.ZkClient;
-import org.I0Itec.zkclient.serialize.BytesPushThroughSerializer;
+import org.apache.kafka.clients.admin.Admin;
+import org.apache.kafka.common.quota.ClientQuotaAlteration;
+import org.apache.kafka.common.quota.ClientQuotaFilter;
+import org.apache.kafka.common.quota.ClientQuotaFilterComponent;
+import org.apache.kafka.common.quota.ClientQuotaEntity;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 
-import java.io.IOException;
-import java.io.UnsupportedEncodingException;
-import java.net.URLEncoder;
-import java.nio.charset.StandardCharsets;
+import java.util.Collections;
+import java.util.HashSet;
+import java.util.Map;
+import java.util.Objects;
+import java.util.Set;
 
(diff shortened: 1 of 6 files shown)


scholzj:
This can be probably merged only after the support for Kafka 2.5.x is removed after branching 0.21, or?

ppatierno:
Right. I opened this issue https://github.com/strimzi/strimzi-kafka-operator/issues/4224 to work on for it.

ppatierno:
This PR can be merged only after #4491 is merged.

user-operator/src/main/java/io/strimzi/operator/user/operator/KafkaUserQuotasOperator.java
```diff
@@ -202,119 +92,105 @@ boolean isPathExist(String path)    {
...
-            }
-            if (config.getString("request_percentage") != null) {
-                config.remove("request_percentage");
+    protected KafkaUserQuotas describeUserQuotas(String username) throws Exception {
+        ClientQuotaFilterComponent c = ClientQuotaFilterComponent.ofEntity(ClientQuotaEntity.USER, username);
+        ClientQuotaFilter f =  ClientQuotaFilter.contains(Collections.singleton(c));
+        KafkaUserQuotas current = null;
+        try {
+            ClientQuotaEntity cqe = new ClientQuotaEntity(Collections.singletonMap(ClientQuotaEntity.USER, username));
+            Map<ClientQuotaEntity, Map<String, Double>> map = adminClient.describeClientQuotas(f).entities().get();
```
scholzj, line 136: Should this be done asynchronously? Or is this whole block in some `executeBlocking`?

user-operator/src/main/java/io/strimzi/operator/user/operator/KafkaUserQuotasOperator.java
```diff
@@ -202,119 +92,105 @@ boolean isPathExist(String path)    {
...
-     *
-     * @return  Returns the updated JSON without the quotas
-     */
-    protected JsonObject removeQuotasFromJsonUser(byte[] userConfig)   {
-        JsonObject json = new JsonObject(new String(userConfig, StandardCharsets.UTF_8));
+    protected void alterUserQuotas(String username, Set<ClientQuotaAlteration.Op> ops) throws Exception {
+        ClientQuotaEntity cqe = new ClientQuotaEntity(Collections.singletonMap(ClientQuotaEntity.USER, username));
+        ClientQuotaAlteration cqa = new ClientQuotaAlteration(cqe, ops);
+        try {
+            adminClient.alterClientQuotas(Collections.singleton(cqa)).all().get();
```
scholzj, line 123: Should this be done asynchronously? Or is this whole block in some `executeBlocking`?

scholzj (commented):
I left few comments. I also wonder if you tried whether it works well in combination with the SCRAM-SHA users? I think that in the past we had problem with it because they are stored in the same path in ZK.

user-operator/src/main/java/io/strimzi/operator/user/operator/KafkaUserQuotasOperator.java
ppatierno: Yes, all the adminClient calls are done from inside the `reconcile` method (which calls `createOrUpdate`) and the `reconcile` does these calls in an `executeBlocking`. The same happens for the `SimpleAclOperator` for example.

user-operator/src/main/java/io/strimzi/operator/user/operator/KafkaUserQuotasOperator.java
ppatierno: Ditto as above.

scholzj (approved):

PR #4222, opened 2021-01-11
Fix default docker registry in dev guide

ppatierno:
### Type of change

- Documentation

### Description

Trivial PR about specifying `quay.io` as default container registry instead of `docker.io`.

### Checklist

- [x] Update documentation


scholzj:
The default registry is still `docker.io`. We did not changed it. The reason was that it would break any existing workflows for users who never needed to redefine this variable.

ppatierno:
Ok my bad. Thanks.

PR #4227, opened 2021-01-12
Remove Helm 2 support

tomncooper:
### Type of change

Refactoring

### Description

Helm2 support ended in [November last year](https://helm.sh/blog/helm-2-becomes-unsupported/) so this PR removes the helm 2 build option.

### Checklist

- [X] Update documentation
- [X] Update CHANGELOG.md


Makefile
```diff
@@ -12,7 +12,7 @@ ifneq ($(RELEASE_VERSION),latest)
   GITHUB_VERSION = $(RELEASE_VERSION)
 endif
 
-SUBDIRS=kafka-agent mirror-maker-agent tracing-agent crd-annotations test crd-generator api mockkube certificate-manager operator-common config-model config-model-generator cluster-operator topic-operator user-operator kafka-init docker-images helm-charts/helm2 helm-charts/helm3 install examples
+SUBDIRS=kafka-agent mirror-maker-agent tracing-agent crd-annotations test crd-generator api mockkube certificate-manager operator-common config-model config-model-generator cluster-operator topic-operator user-operator kafka-init docker-images helm-charts/helm3 install examples
```
scholzj, line 15: There is more stuff in this file ... check the targets `bridge_version`, `release_prepare`, `release_helm_version`, `helm_pkg` ... they all need to be fixed.

scholzj (commented):
Right now, we expected two binaries for the build: `helm2` for Helm 2 and `helm3` for Helm 3. Should we update this and move to using only `helm` with Helm 3? It might be easier for users to use since they would be able to use the default Helm and not some manually created names.

tomncooper:
Good point, fixing now.

scholzj (approved):
Looks good. Lets see if all the tests pass.

scholzj:
```
Run 'helm init' to configure helm.
+ echo 'Verifying the installation of helm3 binary...'
+ helm3 --help
Verifying the installation of helm3 binary...
.azure/scripts/setup-helm.sh: line 18: helm3: command not found
+ helmCommandOutput=127
+ '[' 127 '!=' 0 ']'
+ echo 'helm binary hasn'\''t been installed properly - exiting...'
+ exit 1
helm binary hasn't been installed properly - exiting...
```

tomncooper:
Sorry I am still fixing some issues with the main make file.

scholzj:
Can you resolve the conflicts so that I can run tests on it?

tomncooper:
Done

scholzj (approved):

scholzj:
Everything passed ... so I guess a good job. I will merge it after we get all the needed approvals.

Frawless (approved):

Makefile
```diff
@@ -70,19 +68,17 @@ release_pkg: helm_pkg
...
-		$(SED) -i 's/\(image\.tagPrefix[^\n]*| \)`.*`/\1`$(RELEASE_VERSION)`/g' $$CHART_PATH/README.md;	\
-		$(SED) -i '/name: quay\.io\/kafka-bridge/{n;s/\(tag: \).*/\1$(BRIDGE_VERSION)/g}' $$CHART_PATH/values.yaml;	\
-		$(SED) -i 's/\(kafkaBridge.image\.tag[^\n]*| \)`.*`/\1`$(BRIDGE_VERSION)`/g' $$CHART_PATH/README.md;	\
-	done
+	CHART_PATH=./helm-charts/helm3/strimzi-kafka-operator; \
+	$(SED) -i 's/\(tag: \).*/\1$(RELEASE_VERSION)/g' $$CHART_PATH/values.yaml; \
+	$(SED) -i 's/\(tagPrefix: \).*/\1$(RELEASE_VERSION)/g' $$CHART_PATH/values.yaml; \
+	$(SED) -i 's/\(image\.tag[^\n]*| \)`.*`/\1`$(RELEASE_VERSION)`/g' $$CHART_PATH/README.md; \
+	$(SED) -i 's/\(image\.tagPrefix[^\n]*| \)`.*`/\1`$(RELEASE_VERSION)`/g' $$CHART_PATH/README.md; \
+	$(SED) -i '/name: quay.io\/kafka-bridge/{n;s/\(tag: \).*/\1$(BRIDGE_VERSION)/g}' $$CHART_PATH/values.yaml; \
```
ppatierno, line 76: as done by @scholzj in this PR https://github.com/strimzi/strimzi-kafka-operator/pull/4225, we need to escape the "dot" in quay.io `'/name: quay\.io\/kafka-bridge/{n;s/\(tag: \).*/\1$(BRIDGE_VERSION)/g}'`

install/Makefile
```diff
@@ -59,7 +29,7 @@ crd_install_helm3:
 	yq write -i ../helm-charts/helm3/strimzi-kafka-operator/crds/048-Crd-kafkamirrormaker2.yaml metadata.labels.component "kafkamirrormaker2.kafka.strimzi.io-crd"
 	yq write -i ../helm-charts/helm3/strimzi-kafka-operator/crds/049-Crd-kafkarebalance.yaml metadata.labels.component "kafkarebalances.kafka.strimzi.io-crd"
 
-crd_install: crd_install_helm2 crd_install_helm3
+crd_install: crd_install_helm3
```
ppatierno, line 32: at this point I think we could even renaming from `crd_install_helm3` to just `crd_install_helm`

ppatierno (changes requested):

install/Makefile
tomncooper: I did think about doing that but I guess at some point there will be a helm 4 and we may need to run both again. If we leave it like this there is less to change. WDYT?

Makefile
tomncooper: Good catch! I have patched that.

install/Makefile
ppatierno: Makes sense. I was wondering we were going to support just one Helm version from now on. If not, then it's fine helm3 ;-)

ppatierno (approved):

install/Makefile
scholzj: Well, I guess there will be a long time until Helm4. So I do not think that is something what we will add in next months. (at the same time, I do not have problem to keep the target called helm3)

PR #4236, opened 2021-01-13
Make the Kafka Exporter Grafana Dashboard always find the Kafka cluster even when there are no consumer groups

scholzj:
### Type of change

- Bugfix

### Description

The Kafka Exporter dashboard detects the namespaces / clusters based on the `kafka_consumergroup_current_offset` metric. This metric is not present in some situation (probably when there are no committed offsets after startup etc.). While without this metric the dashabord has only limited value, using another metric which would be there all the time to detect the Exporter even without consumer groups could help to avoid the confusions. 

In this PR I use the `kafka_exporter_build_info` metric which should be always present. The value it self is not so important => it is used only to detect the exporters / clusters presence.

### Checklist


ppatierno (approved):
LGTM. That metric seems to be exported by the Go client used to instrument Prometheus metrics in the exporter and not by the exporter itself, so it should be always there.

PR #4238, opened 2021-01-13
Upgrade kubernetes-client to 5.0.0

sknot-rh:
### Type of change

- Refactoring

### Description

Upgrading dependency. Doneable classes removed. `Spec` and `Status` are now part of CustomResource (I think we can extend `io.fabric8.kubernetes.api.model.Status` for our status classes). 

### Checklist

- [ ] Write tests
- [ ] Update documentation
- [ ] Update CHANGELOG.md


scholzj (approved):

im-konge (approved):
Thanks for updating the STs :) LGTM

ppatierno (approved):

scholzj:
@Frawless @tombentley Do you plan to review this? Or can it be merged? It passed the tests and is already approved by me and Paolo.

tombentley:
I have looked at it, and my only reservation is that I'm not completely sure how well it will work with the move to `v1beta2`. The `CustomResource` constructor now figures out the API version from the annotations on the class, meaning that our subclasses of `CustomResource` (e.g. `Kafka` etc) need to be annotated with `@Version` and so are version specific. It don't fully understand the consequences of that. Maybe there are none, maybe we can work around it, or maybe for those subclasses we'd end up needing per-version "CR root" classes (I'm pretty sure the rest of the classes could stay the same). I guess there's only one way we'll find out, and there isn't a good alternative which doesn't involve merging this PR.

tombentley (approved):

PR #4239, opened 2021-01-13
[DOC] Improve presentation of the schema content in the Using Guide

PaulRMellor:

**Documentation**

- Using Guide split into parts
   -- Overview
   -- Configuring
   -- Managing w/ FAQ
   -- Schema reference
- FAQs moved to _Managing Strimzi_ section
- New schema titles -- without `schema reference`
- Table title says _<schema-name> schema properties_
- Link to table added where additional descriptions present in section

### Checklist

- [ ] Write tests
- [x] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/documentation/api/io.strimzi.api.kafka.model.AclRule.adoc
+++ b/documentation/api/io.strimzi.api.kafka.model.AclRule.adoc
@@ -1,5 +1,3 @@
-xref:type-AclRule-schema-{context}[Full list of `AclRule` schema properties]
-
 Configures access control rule for a `KafkaUser` when brokers are using the `AclAuthorizer`.
 
 .Example `KafkaUser` configuration with authorization
--- a/documentation/api/io.strimzi.api.kafka.model.AclRule.adoc
+++ b/documentation/api/io.strimzi.api.kafka.model.AclRule.adoc
@@ -1,3 +1,5 @@
+xref:type-AclRule-schema-{context}[Full list of `AclRule` schema properties]
+
 Configures access control rule for a `KafkaUser` when brokers are using the `AclAuthorizer`.
 
 .Example `KafkaUser` configuration with authorization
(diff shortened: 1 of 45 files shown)


documentation/api/io.strimzi.api.kafka.model.AclRule.adoc
```diff
@@ -1,3 +1,5 @@
+xref:type-AclRule-schema-{context}[Full list of `AclRule` schema properties]
```
scholzj, line 1: I wonder if instead of adding these manually, we should generate these in the generator. Otherwise we might soon add new modules and forget to add this. Since these are just includes, it should be easy to generate it before it It could be just added before the line 419 in `DocGenerator`.

scholzj (commented):
The build failed, because it seems that some of the tests in the `crd-generator` are failing.

documentation/api/io.strimzi.api.kafka.model.AclRule.adoc
PaulRMellor: Yeah. I thought about this yesterday. Not sure why I went for the manual in the end.

PaulRMellor:
> The build failed, because it seems that some of the tests in the `crd-generator` are failing.

I removed _schema reference_ from section titles and matched in SimpleTest. Not sure why that's an issue?

scholzj:
> I removed schema reference from section titles and matched in SimpleTest. Not sure why that's an issue?

I guess the tests would fail even with just some different whitespaces etc. I can look at it later if you want.

laidan6000 (approved):
These changes look sensible to me. I like the new title for Deployment configuration.

After this is merged, please can you share a link to the Using Strimzi guide on `master`. It would be a good idea to check the changes in production, e.g. use of Asciidoc parts, new headings in the schema reference. Thanks.

scholzj (commented):
I'm not sure the move of the FAQs into _Managing Strimzi_ makes really sense to me. But I can live with it if you think it is better there. Can you look at the failing build? It looks like some of the tests of the docs generator are failing. Or do you need some help with it?

PaulRMellor:
> I'm not sure the move of the FAQs into _Managing Strimzi_ makes really sense to me. But I can live with it if you think it is better there. Can you look at the failing build? It looks like some of the tests of the docs generator are failing. Or do you need some help with it?

Removing _schema reference_ from the schema headings was causing the issue. Is there any way we can retain that change?

scholzj (approved):

PR #4261, opened 2021-01-17
Fix the network-policies handling when metrics config from CM is used

scholzj:
### Type of change

- Bugfix

### Description

When the new metrics configuration from ConfigMap is used, the Prometheus JMX Exporter port is not opened properly on the network policies and that makes the metrics not accessible. This PR properly sets the `metricsEnabled` value when one of the currently supported options is used.

This is a bug in the 0.21.0 release. It needs to be backported to 0.21 release branch and we need to do 0.21.1 release.

### Checklist

- [x] Write tests

### DIFF
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/model/ModelUtils.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/model/ModelUtils.java
@@ -519,12 +519,12 @@ public class ModelUtils {
      * @param resourceWithMetrics       The section of the resource with metrics configuration
      */
     public static void parseMetrics(AbstractModel model, HasConfigurableMetrics resourceWithMetrics)   {
-        if (resourceWithMetrics.getMetrics() != null) {
-            model.setMetricsEnabled(true);
-            model.setMetricsConfig(resourceWithMetrics.getMetrics().entrySet());
-        } else if (resourceWithMetrics.getMetricsConfig() != null)    {
+        if (resourceWithMetrics.getMetricsConfig() != null)    {
             model.setMetricsEnabled(true);
             model.setMetricsConfigInCm(resourceWithMetrics.getMetricsConfig());
+        } else if (resourceWithMetrics.getMetrics() != null) {
+            model.setMetricsEnabled(true);
+            model.setMetricsConfig(resourceWithMetrics.getMetrics().entrySet());
         }
     }
 }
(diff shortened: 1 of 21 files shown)


cluster-operator/src/main/java/io/strimzi/operator/cluster/model/ModelUtils.java
```diff
@@ -509,6 +510,21 @@ public static void setClusterOperatorNetworkPolicyNamespaceSelector(NetworkPolic
...
+     * @param resourceWithMetrics       The section of the resource with metrics configuration
+     */
+    public static void parseMetrics(AbstractModel model, HasConfigurableMetrics resourceWithMetrics)   {
+        if (resourceWithMetrics.getMetrics() != null) {
+            model.setMetricsEnabled(true);
+            model.setMetricsConfig(resourceWithMetrics.getMetrics().entrySet());
+        } else if (resourceWithMetrics.getMetricsConfig() != null)    {
+            model.setMetricsEnabled(true);
+            model.setMetricsConfigInCm(resourceWithMetrics.getMetricsConfig());
+        }
```
ppatierno, line 528: If the user, by mistake, specifies both `metrics` and `metricsConfig` fields in the `Kafka` resource we are giving precedence to the "deprecated/old" way. Is this on purpose for backward compatibility or we should use the "if" in the other way around?

scholzj: Indeed ... it probably makes sense to use the new first. Thanks

cluster-operator/src/test/java/io/strimzi/operator/cluster/model/ModelUtilsTest.java
```diff
@@ -421,4 +421,9 @@ public void testCONetworkPolicyPeerNamespaceSelectorDifferentNSWithLabels()  {
         ModelUtils.setClusterOperatorNetworkPolicyNamespaceSelector(peer, "my-ns", "my-operator-ns", nsLabels);
         assertThat(peer.getNamespaceSelector().getMatchLabels(), is(nsLabels.toMap()));
     }
+
+    @Test
+    public void testMetricsParsing()    {
```
sknot-rh, line 426: Maybe remove this.

cluster-operator/src/test/java/io/strimzi/operator/cluster/model/KafkaMirrorMaker2ClusterTest.java
```diff
@@ -111,7 +116,7 @@
 
     private final KafkaMirrorMaker2 resource = new KafkaMirrorMaker2Builder(ResourceUtils.createEmptyKafkaMirrorMaker2(namespace, cluster))
             .withNewSpec()
-                .withMetricsConfig(jmxMetricsConfig)
+                //.withMetricsConfig(jmxMetricsConfig)
```
sknot-rh, line 119: Can be removed

sknot-rh (approved):
Few nits, otherwise LGTM

ppatierno (approved):

Issue #307, opened 2018-03-12
Topic controller's AdminClient config


tombentley:
The topic controller should set `client.id` in its AdminClient config, so Kafka logging has a bit more contextual info for connections from the TC.

See https://kafka.apache.org/documentation/#adminclientconfigs

scholzj:
Reviewed on 10th September 2020: This is still valid.

ShubhamRwt:
I would like to work on this issue. Is it still valid?

tombentley:
@ShubhamRwt it is still valid, so feel free to work on it.

ShubhamRwt:
Hello @tombentley. I have added the enhancements and written the tests but I am stuck with how I can test it up.

tombentley:
@ShubhamRwt yeah it's a difficult one to test right now, since `Admin` doesn't expose its config once created. I would be inclined to:

1. Refactor `Session` so that the constructor takes a `Function<Properties, Admin>` which would be used to create the client. `Main` would pass as `Admin::create`.
2. Write a `SessionTest` which created a `Session`, using your own impl of the function about which you could later assert on the entries in the `Properties`. Probably no need to actually call `Session.start()` if you factor out a method for creating the client. 

This would open the door to more tests for `Session` as we could then start using mocked `Admin`.

PR #4270, opened 2021-01-19
Topic controller's AdminClient config #307

ShubhamRwt:
### Type of change
- Enhancement 

### Checklist

- [x] Write tests
- [ ] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/topic-operator/src/main/java/io/strimzi/operator/topic/Config.java
+++ b/topic-operator/src/main/java/io/strimzi/operator/topic/Config.java
@@ -10,6 +10,7 @@ import java.util.HashMap;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Set;
+import java.util.UUID;
 
 public class Config {
 
@@ -100,6 +101,7 @@ public class Config {
     public static final String TC_RESOURCE_LABELS = "STRIMZI_RESOURCE_LABELS";
     public static final String TC_KAFKA_BOOTSTRAP_SERVERS = "STRIMZI_KAFKA_BOOTSTRAP_SERVERS";
     public static final String TC_NAMESPACE = "STRIMZI_NAMESPACE";
+    public static final String TC_CLIENT_ID = "STRIMZI_CLIENT_ID";
     public static final String TC_ZK_CONNECT = "STRIMZI_ZOOKEEPER_CONNECT";
     public static final String TC_ZK_SESSION_TIMEOUT_MS = "STRIMZI_ZOOKEEPER_SESSION_TIMEOUT_MS";
     public static final String TC_ZK_CONNECTION_TIMEOUT_MS = "TC_ZK_CONNECTION_TIMEOUT_MS";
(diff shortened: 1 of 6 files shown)


sknot-rh (approved):

ppatierno (commented):
Not sure we would need an env var for configuring the client-id from the outside.
It could be useful when you need to run more TO instances maybe related to the same Kafka cluster? Something that doesn't sound to make much sense. I would say that a kind of "hard-coded" client-id containing at least the name of the corresponding Kafka cluster (if possible) could be enough. But @tombentley could have a different opinion.

topic-operator/src/main/java/io/strimzi/operator/topic/Config.java
```diff
@@ -119,6 +120,9 @@ private Value(String key, Type<? extends T> type, boolean required) {
     /** The kubernetes namespace in which to operate. */
     public static final Value<String> NAMESPACE = new Value<>(TC_NAMESPACE, STRING, true);
 
+    /** Cluster-ID */
+    public static final Value<String> CLIENT_ID = new Value<>(TC_CLIENT_ID, STRING, "");
```
tombentley, line 124: I think the default should be something more like `"strimzi-topic-operator-" + uuid`, that way it's at least vaguely useful even when not configured.

tombentley:
> Not sure we would need an env var for configuring the client-id from the outside.
It could be useful when you need to run more TO instances maybe related to the same Kafka cluster? Something that doesn't sound to make much sense. I would say that a kind of "hard-coded" client-id containing at least the name of the corresponding Kafka cluster (if possible) could be enough. But @tombentley could have a different opinion.

I think there's two things here: 

1. Being able to specify a particular client id, which is what's implemented here and makes sense in standalone TO use cases.
2. Having a good client id in the CO-deploys-TO scenario. We can build on this PR to implement that. Possibly we should expose it through the API, but I think having the CO configure a good default (e.g. using the namespace and cluster name) should be done even if we don't expose it though the API.

ppatierno:
Yeah I always forget that standalone TO is a kind of thing we still support :-)

topic-operator/src/main/java/io/strimzi/operator/topic/Config.java
ShubhamRwt: Ok sure@tombentley , I will make this change. Regarding the UUID, can we use UUID.randomUUID() to generate it ?

tombentley: `UUID.randomUUID()` is fine with me.

ShubhamRwt: Okay. Thanks

ShubhamRwt: Hello @tombentley,  I have made the necessary changes as per mentioned above but I have not changed the tests. Do I need to change them or are they fine ?

ShubhamRwt:
Hello @ppatierno , @tombentley , @sknot-rh , are the above  changes good to go or do we require some more changes?

tombentley (approved):

ppatierno (approved):

scholzj:
Thanks for the PR.

PR #4314, opened 2021-01-25
Simplify Kafka upgrades and downgrades

scholzj:
### Type of change

- Enhancement / new feature

### Description

This PR implements improved support for Kafka upgrades and downgrades. The current process has following limitations:
* Insists on having the versions (Kafka version, but also `log.message.format.version` or `inter.broker.protocol.version`) written in the custom resource before the upgrade. That makes the upgrades hard for example when using automated upgrades through Operator Hub etc.
* Doesn't easily allow to skip multiple Strimzi and Kafka versions during the upgrade because the Kafka upgrade process requires detailed knowledge of the previous version (to know how to modify the Kafka StatefulSet etc.)

To address these, this change:
* Tries to detect as much of the existing configuration in reliable way (from annotation rather then parsing config maps which might change in the future)
* Tries to guess the used versions (Kafka version X has either pre-configured `log.message.format.version` or `inter.broker.protocol.version` or uses the default which can be derived from the Kafka version)
* Tries to auto-configure the versions where possible (If not set by the user, lets set it explicitly to the default value)

The main change is that to roll out the new Kafka version it uses the regular rolling update of the Kafka pods instead of the special one which was causing lot of issues in the past. It does also some smaller improvements where the annotations are carried on pods to make sure we can handle failures during upgrade and recover from them etc.

Despite the improvements, the process tries to follow the same rules as used before and which are the usual Kafka upgrade process:
* For upgrade, roll out the new brokers while first using the older `log.message.format.version` or `inter.broker.protocol.version` and only afterwards change to the new versions for message format and inter-broker protocol.
* For downgrade, do not allow downgrade when the new `log.message.format.version` or `inter.broker.protocol.version` are already used

In order to be able to handle the upgrade from older Kafka versions not supported by the current version of the operator, I had to add the the `kafka-versions.yaml` also older Kafka versions. They are marked by a new `supported` field. This allows us to handle the upgrade by configuring their native `log.message.format.version` and `inter.broker.protocol.version`

While it makes the process easier for those who do not want to maintain all the versions manually, it of course still supports the old way of having the users maintaining them as in the past. This should also stay the recommended approach.

### Checklist

- [x] Write tests
- [ ] Update CHANGELOG.md


cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaAssemblyOperator.java
```diff
@@ -267,6 +259,7 @@ public KafkaAssemblyOperator(Vertx vertx, PlatformFeaturesAvailability pfa,
                 .compose(state -> state.reconcileCas(this::dateSupplier))
                 .compose(state -> state.clusterOperatorSecret(this::dateSupplier))
                 .compose(state -> state.getKafkaClusterDescription())
+                .compose(state -> state.prepareVersionChange())
```
tombentley, line 262: Is there a reason this has to be done so early in the reconciliation?

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaAssemblyOperator.java
```diff
@@ -298,8 +291,6 @@ public KafkaAssemblyOperator(Vertx vertx, PlatformFeaturesAvailability pfa,
                 .compose(state -> state.kafkaManualPodCleaning())
                 .compose(state -> state.kafkaNetPolicy())
                 .compose(state -> state.kafkaManualRollingUpdate())
-                .compose(state -> state.kafkaVersionChangeCheck())
-                .compose(state -> state.kafkaVersionChange(false))
```
tombentley, line 302: The original idea with having the version change roll early in the kafka part of the reconciliation is the operator hasn't done anything to the kafka pods at this point, so it doesn't needed to make any assumptions about how the image behaves/what it expects, which is relevant when the operator is upgraded and does the forced operand upgrade.

By deferring the forced operand upgrade until after we've done things like changed the networking and updated the ancillary CM there is a higher risk that one of the brokers suffers a spontaneous failure, restarts and doesn't see compatible config (because the image in the sts hasn't been changed yet). The KafkaConfig algorithm should be enough because it always tries to roll brokers which aren't ready first. So any broker which had spontaneously restarted would restart first and because it's then using the right image it should be able to start. So I _think_ this is a recoverable error, assuming that all such failures result in an unready broker.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaAssemblyOperator.java
```diff
@@ -1823,10 +1451,79 @@ String zkConnectionString(int connectToReplicas, Function<Integer, String> zkNod
...
+                                }
+                            }
+                        }
+
+                        // We decide what is the current Kafka version used and create the KafkaVersionChange object
+                        // describing the situation.
+                        if (lowestKafkaVersion == null)   {
+                            // No version found in Pods or StatefulSet => New cluster or cluster with deleted
+                            // StatefulSet. We treat it as if current version and desired are the same which results in
+                            // noop.
```
tombentley, line 1514:
```suggestion
                            // a noop reconciliation followed by a 2nd reconciliation for actual upgrade/downgrade
```

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaAssemblyOperator.java
```diff
@@ -1823,10 +1451,79 @@ String zkConnectionString(int connectToReplicas, Function<Integer, String> zkNod
...
+
+                        //return Future.succeededFuture(this);
+                        return podOperations.listAsync(namespace, this.kafkaCluster.getSelectorLabels());
+                    }).compose(pods -> {
+                        String lowestKafkaVersion = currentVersion;
+                        String highestKafkaVersion = currentVersion;
+
+                        for (Pod pod : pods)    {
+                            // Collect the different annotations from the pods
+                            String kafkaVersionAnnotation = Annotations.stringAnnotation(pod, ANNO_STRIMZI_IO_KAFKA_VERSION, null);
```
tombentley, line 1468: I think names like `currentVersion`, `currentMessageFormat` and `currentIbp` would make this bit of code clearer.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaAssemblyOperator.java
```diff
@@ -1823,10 +1451,79 @@ String zkConnectionString(int connectToReplicas, Function<Integer, String> zkNod
                         kafkaCurrentReplicas = 0;
                         if (sts != null && sts.getSpec() != null)   {
                             kafkaCurrentReplicas = sts.getSpec().getReplicas();
+                            this.currentVersion = Annotations.annotations(sts).get(ANNO_STRIMZI_IO_KAFKA_VERSION);
```
tombentley, line 1454: Let's rename the field `stsCurrentVersion` since there's STS and Pod ones.

tombentley (commented):
Some initial comments.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaAssemblyOperator.java
scholzj: I think it can technically be also later on. But this is what decides if the change is feasible or not. So I think having it up top is better because it stops the upgrade / downgrade which would not work right away. Without for example already upgrading Zookeeper just to find out that Kafka cannot be upgraded. At the same time, all information is well known already here. So I did not see any reasons to do it later.

tombentley: Makes sense, thanks

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaAssemblyOperator.java
scholzj: Yeah, I know that was the original motivation. But it was also the part which caused all the issues because it required knowledge of prior versions. In theory yes, something can happen and break the cluster. But I think the time is fairly short and the benefits from using the regular rolling update overweight the negatives of having to fight with the prior knowledge.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaAssemblyOperator.java
tombentley: Can you explain the new cluster case mentioned?

scholzj: You mean here or in the comment? 

New cluster means that you just created the Kafka CR and this is the first reconciliation of the cluster => so everything is created for the first time and no upgrade or downgrade matters. In this case the StatefulSet does not exist yet. This is a legitimate noop. There is no version change.

Other case would be that the StatefulSet was deleted either manually or in some disaster situation. In that case all we can do is roll out the new STS based on the CR. We do not know what version was used before it was deleted, but in any case there is nothing to roll. So again, the noop and rolling out the version as per the Kafka CR is ok I think.

The third use case - this is the one which we can filter out and possibly fail the reconciliation - would be if the StatefulSet exists but does not have the annotation. The version annotation is used since ages. So this would mean some super old version before you added the initial upgrade support. My thoughts were that if someone really tries this, they cannot really expect to have a smooth update so it just rolls through it. But as I said, we can also easily fail this.

tombentley: Thanks for the explanation. I think for the 3rd case failing the reconciliation might be best. We could tell the user to add the annotation themselves?

scholzj: Ok, I updated it accordingly.

tombentley (approved):

PR #4322, opened 2021-01-27
Renamed Condition to KafkaCondition

ppatierno:
### Type of change

- Refactoring

### Description

Using the Strimzi API in another application happens that the `Condition` class we provide clashes with the `Condition` class in the fabric8 Kubernetes client. Other than this, the same fabric8 provides specific classes like `PodCondition`, `DeploymentCondition` and so on. 
For avoiding clashes and following kind of the same pattern, this PR just renames `Condition` to `KafkaCondition`.

Maybe is it a kind of breaking changes in the API if someone is using them but I think it worths it. I can add this to the CHANGELOG if you agree to this change.

### Checklist

- [ ] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/api/src/main/java/io/strimzi/api/kafka/model/status/KafkaCondition.java
+++ b/api/src/main/java/io/strimzi/api/kafka/model/status/KafkaCondition.java
@@ -27,7 +27,7 @@ import static java.util.Collections.emptyMap;
 @JsonPropertyOrder({ "type", "status", "lastTransitionTime", "reason", "message" })
 @EqualsAndHashCode
 @ToString
-public class Condition implements UnknownPropertyPreserving, Serializable {
+public class KafkaCondition implements UnknownPropertyPreserving, Serializable {
     private static final long serialVersionUID = 1L;
 
     private String status;
--- a/api/src/main/java/io/strimzi/api/kafka/model/status/Status.java
+++ b/api/src/main/java/io/strimzi/api/kafka/model/status/Status.java
@@ -33,33 +33,33 @@ import static java.util.Collections.emptyMap;
 @EqualsAndHashCode
 @ToString
 public abstract class Status implements UnknownPropertyPreserving, Serializable {
-    private List<Condition> conditions;
+    private List<KafkaCondition> conditions;
     private long observedGeneration;
     private Map<String, Object> additionalProperties;
 
     @Description("List of status conditions")
-    public List<Condition> getConditions() {
+    public List<KafkaCondition> getConditions() {
         return conditions;
     }
 
-    public void setConditions(List<Condition> conditions) {
+    public void setConditions(List<KafkaCondition> conditions) {
         this.conditions = conditions;
     }
 
-    private List<Condition> prepareConditionsUpdate() {
-        List<Condition> oldConditions = getConditions();
-        List<Condition> newConditions = oldConditions != null ? new ArrayList<>(oldConditions) : new ArrayList<>(0);
+    private List<KafkaCondition> prepareConditionsUpdate() {
+        List<KafkaCondition> oldConditions = getConditions();
+        List<KafkaCondition> newConditions = oldConditions != null ? new ArrayList<>(oldConditions) : new ArrayList<>(0);
         return newConditions;
     }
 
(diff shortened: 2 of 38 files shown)


samuel-hawker (approved):
LGTM

scholzj:
@ppatierno Isn't this backwards incompatible change of the API?

ppatierno:
@scholzj as I mentioned yes, it is but I think it's worth instead of having the explained clashing forever and this kind of "bad" naming not following the pattern fabric8 is using. I would say that a breaking changes statement in the changelog could be enough.

tombentley:
@ppatierno I disagree that this is worth breaking compatibility for. Yes, it's annoying if you have to fully qualify the class name everywhere, but compatibility is more important in my view. I suspect you can refactor/write some helper functions so that the need for fully qualified names is at least isolated to a single class.

ppatierno:
@tombentley tbh I really think that breaking this now instead of having it forever and even after a 1.0 release would be better imho.
How we should move forward and decide on this?

ppatierno:
@tombentley other than this how don't know how much the Maven Central statistics are valid but it seems no one is using the Strimzi API right now. https://mvnrepository.com/artifact/io.strimzi/api

tombentley:
@ppatierno according to https://semver.org/ item 4 because we still have major version 0 we're OK to do as you propose because people should not be assuming that our API is stable. But _in practice_ we've done a pretty good job of not making incompatible changes. If the problem you were trying to solve didn't have a solution as simple as fully qualifying the class name then I would think differently. But there is no pressing need to make this change and break people's code, because a simple (if annoying) solution exists. If we want to fix it for 1.0 that would be fine imho.

Indeed, I think we should probably start opening issues with 1.0 as milestone for any compatibility breaks we want to make for 1.0.

Ultimately, if the other maintainers think differently I'm not going to stand in your way.

ppatierno:
@tombentley thanks for raising the semver point, so it sounds like you can brake API and it's better sooner than later unless as you are saying we don't want never brake anything because there is a simple solution like using the fqn for the `Condition` class.
I see your point but I still think that it's not a so major change even if breaking "not stable" API.

scholzj:
AFAIK the numbers on the mvnrepository.org show only use of the artefacts which it self is published in Maven central repository. It does not show any user who never published there. So I do not think it tells much. 

I also wonder how would you feel about this if similar change is done again in few months when you are already using the `api` in production and have to update accordingly. Would you still see it as totally fine?

ppatierno:
I could feel bad but right now I am aware of using a 0.x.y version that exposes not stable API; isn't it the case of new Kubernetes features as well?

ppatierno:
Just to point out that updating to fabric8 5.0.0 kind of broke things already for people who were using the `Kafka` resource k8s client creation without providing a CRD context as then fixed in one of my latest PR.

scholzj:
> but right now I am aware of using a 0.x.y version that exposes not stable API

Great. Arguments like this make it really hard to explain to others why Strimzi is production ready even when it is not 1.0. If someone finds this you are basically telling them that they should not use Strimzi because it is not stable, it is still under development and can break at any point. :-(

---

I do not have a very strong opinion on this. I think this should not be done because it does add little to no value, but I do not feel strong enough about it to block it in any way. I just hope you will be consistent if someone comes with a similar change in the future when it breaks your code.

ppatierno:
> Great. Arguments like this make it really hard to explain to others why Strimzi is production ready even when it is not 1.0. If someone finds this you are basically telling them that they should not use Strimzi because it is not stable, it is still under development and can break at any point. :-(

I think it's kind of different. I am not saying that the operator is not stable and not production-ready. Changing or breaking the API doesn't mean breaking internal stuff that just works today as Strimzi does. And this PR doesn't break the stability of the operator for sure; it's maybe just "cosmetic" but annoying.

Regarding moving forward or not I could just merge because two maintainers (me and Sam) are fine but I don't want to go that way. I would love to have a strong majority of maintainers and even if @Frawless would be ok with this, it's going to be just 3 votes against 2.
For this reason, I will close this PR.

PR #4341, opened 2021-01-29
Tests - rename cert with expired prefix

michalxo:
Simple rename of static expired certificates, which are used in tests


michalxo:
@strimzi-ci run tests true profile=regression testcase=io.strimzi.systemtest.security.SecurityST#testAutoRenewCaCertsTriggerByExpiredCertificate,io.strimzi.systemtest.security.SecurityST#testCaRenewalBreakInMiddle,io.strimzi.systemtest.connectConnectS2IST#testChangeConnectS2IConfig

michalxo:
@see-quick do you think there are some other security related tests that might use these expired certificates? I haven't found any other tests.
Could anyone please fire up azure pipelines? 
Thanks

scholzj (commented):
Shouldn't the expired cert be generated as well? Or is that not possible?

michalxo:
> Shouldn't the expired cert be generated as well? Or is that not possible?

IMO it does not give any benefit to currently used tests, to generate dynamically invalid certificate. (But I might be wrong for some specific use cases..)
I also seen similar story with `operator-common/src/test/java/io/strimzi/operator/common/operator/MockCertManager.java` but it's stored as a string instead of file.

scholzj (approved):

ppatierno (approved):

Frawless (approved):

see-quick (approved):

Issue #4181, opened 2021-01-05
Kafka Connect Build: Keep a copy of the Kaniko builder under the Strimzi repositories


scholzj:
Currently, we are using the Kaniko container image directly from their repository. I'm not sure this is optimal - it is not visible that we use it. It is also stored in different infra etc. It might make sense to maintain a copy of it under our own container respository. I think Kaniko is trusted community, but we never know if they for example have to move the images etc. So keeping our own copy might make sense.

PR #4343, opened 2021-01-29
Keep a copy of the Kaniko builder under the Strimzi repositories

scholzj:
### Type of change

- Enhancement / new feature

### Description

As described in #4181 ... Currently, we are using the Kaniko container image directly from their repository. I'm not sure this is optimal - it is not visible that we use it. It is also stored in different infra etc. It might make sense to maintain a copy of it under our own container repository. I think Kaniko is trusted community, but we never know if they for example have to move the images etc. So keeping our own copy makes sense.

This PR does not build Kaniko from source code => it just pulls the image build by the Kaniko community, retags it and pushes it to Strimzi Quay.io registry.

### Checklist

### DIFF
--- a/docker-images/kaniko-executor/Makefile
+++ b/docker-images/kaniko-executor/Makefile
@@ -1,0 +1,14 @@
+PROJECT_NAME := kaniko-executor
+KANIKO_EXECUTOR = gcr.io/kaniko-project/executor:v1.3.0
+
+docker_build:
+	# The Kaniko executor image used for building new Kafka Connect images with additional connectors is not build from
+	# scratch. We just pull the one released by Kaniko, retag it and push it to our repository to have our versioning on
+	# it and have it stored there in our other images.
+	$(DOCKER_CMD) pull $(KANIKO_EXECUTOR)
+	$(DOCKER_CMD) tag $(KANIKO_EXECUTOR) strimzi/$(PROJECT_NAME):latest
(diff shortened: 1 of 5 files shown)


scholzj:
@strimzi/maintainers I named the new image `kaniko-executor` (`quay.io/strimzi/kaniko-executor:latest`). I was a bit undecided between a neutral name such as `connect-builder` and a name which references the Kaniko project and makes it more clear what is inside. But at the end I went for `kaniko-executor`. Do you have any strong opinions about this? If needed, it should not be a problem to rename it fo course.

docker-images/kaniko-executor/Makefile
```diff
@@ -0,0 +1,14 @@
+PROJECT_NAME := kaniko-executor
+KANIKO_EXECUTOR = gcr.io/kaniko-project/executor:v1.3.0
```
ppatierno, line 2: why it's not `latest`?

scholzj: I think we should have a fixed tag and when we decide to upgrade it we should do that intentionally with proper testing etc. Not just because Kaniko changes the latest to something else.

ppatierno (approved):

Frawless (approved):

scholzj:
@tombentley It helps us to avoid issues such as Helm had for example when they needed to move the repository and everyone had to change the URL. This protects us against these changes. I think it is a little effort which might not add any significant value, but also can save us a lot of troubles. I think it is worth it.

tombentley (approved):

PR #4344, opened 2021-01-30
Allow to pass additional options to Kaniko executor for Kafka Connect builds

### LINKED ISSUE #4180: Kafka Connect Build: Add advanced option for Kaniko builds
The Kaniko builds used in Kafka Connect build can have different options. For example to enable use of insecure push / pull container repositories, doing reproducible builds or squashing image layers. These might be useful in some cases. So it would be handy to add support for them in the API and pass them into Kaniko builder.

scholzj:
### Type of change

- Enhancement / new feature

### Description

The Kaniko executor used for the Kafka Connect builds has all kind of options which might be useful in some situation. From logging configuration, to handling insecure container registries etc. It might be useful in some situation to be able to configure these.

This PR adds a new field `additionalKanikoOptions` which allows to configure such options. The options are validated through allow-list so that they don't conflict with the options which Strimzi needs to configure. This should close #4180.

### Checklist

- [x] Write tests
- [x] Update documentation
- [x] Update CHANGELOG.md

### DIFF
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaConnectBuild.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaConnectBuild.java
@@ -164,7 +164,7 @@ public class KafkaConnectBuild extends AbstractModel {
     private static void validateAdditionalKanikoOptions(List<String> desiredOptions)    {
         List<String> allowedOptions = Arrays.asList(DockerOutput.ALLOWED_KANIKO_OPTIONS.split("\\s*,+\\s*"));
         List<String> forbiddenOptions = desiredOptions.stream()
-                .filter(option -> !allowedOptions.stream().anyMatch(allowed -> option.startsWith(allowed)))
+                .filter(option -> allowedOptions.stream().noneMatch(option::startsWith))
                 .collect(Collectors.toList());
 
         if (!forbiddenOptions.isEmpty())    {
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaConnectBuild.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaConnectBuild.java
@@ -92,9 +92,8 @@ public class KafkaConnectBuild extends AbstractModel {
 
                 if (dockerOutput.getAdditionalKanikoOptions() != null
                         && !dockerOutput.getAdditionalKanikoOptions().isEmpty())  {
-                    List<String> desiredOptions = Arrays.asList(dockerOutput.getAdditionalKanikoOptions().split("\\s+"));
-                    validateAdditionalKanikoOptions(desiredOptions);
-                    build.additionalKanikoOptions = desiredOptions;
+                    validateAdditionalKanikoOptions(dockerOutput.getAdditionalKanikoOptions());
+                    build.additionalKanikoOptions = dockerOutput.getAdditionalKanikoOptions();
                 }
             }
         }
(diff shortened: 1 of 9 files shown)


cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaConnectBuild.java
```diff
@@ -268,13 +300,20 @@ public Pod generateBuilderPod(boolean isOpenShift, ImagePullPolicy imagePullPoli
     protected List<Container> getContainers(ImagePullPolicy imagePullPolicy) {
         List<Container> containers = new ArrayList<>(1);
 
+        List<String> args = new ArrayList<>(4);
```
ppatierno, line 303: should be this 4 + size of the additionalKanikoOptions (but only if additionalKanikoOptions != null)?
So kind of

```java
List<String> args = additionalKanikoOptions != null ? new ArrayList<>(4 + additionalKanikoOptions.size()) : new ArrayList<>(4)
```

ppatierno (approved):
LGTM. Just a nit, but no need to change if you would leave as it is.

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaConnectBuild.java
```diff
@@ -81,6 +83,19 @@ public static KafkaConnectBuild fromCrd(KafkaConnect kafkaConnect, KafkaVersion.
...
             validateBuildConfiguration(spec.getBuild());
+
+            // The additionalKanikoOptions are validated separately to avoid parsing the list twice
+            if (spec.getBuild().getOutput() != null
+                    && spec.getBuild().getOutput() instanceof DockerOutput) {
+                DockerOutput dockerOutput = (DockerOutput) spec.getBuild().getOutput();
+
+                if (dockerOutput.getAdditionalKanikoOptions() != null
+                        && !dockerOutput.getAdditionalKanikoOptions().isEmpty())  {
+                    List<String> desiredOptions = Arrays.asList(dockerOutput.getAdditionalKanikoOptions().split("\\s+"));
```
tombentley, line 94: Is is possible for an option to contain whitespace? Even if none of them allow this today we can't rule it out in the future. Perhaps it would be better if the `addtionalOptions` were a list?

scholzj: I think with the `jvmOptions` we had the problem that list has not guaranteed ordering, right? But in general, I'm fine with list as well ... so I can change it.

scholzj: Ok, this should be now changed.

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaConnectBuild.java
```diff
@@ -139,6 +153,23 @@ private static void validateBuildConfiguration(Build build)    {
...
+    /**
+     * Validates the additional Kaniko options configured by the user against the list of allowed options. If any
+     * options which are not allowed are found, it raises an InvalidResourceException exception.
+     *
+     * @param desiredOptions    List of additional Kaniko options configured by the user
+     */
+    private static void validateAdditionalKanikoOptions(List<String> desiredOptions)    {
+        List<String> allowedOptions = Arrays.asList(DockerOutput.ALLOWED_KANIKO_OPTIONS.split("\\s*,+\\s*"));
+        List<String> forbiddenOptions = desiredOptions.stream()
+                .filter(option -> allowedOptions.stream().noneMatch(option::startsWith))
```
tombentley, line 165: `startsWith` risks confusing allowed and disallowed options with differ only by suffix. For example there is already `--insecure` and `--insecure-pull`, but what if `--insecure` was disallowed but `--insecure-pull` was allowed?

I think you've used `startsWith` to avoid having to parse/distinguish from pure `--options` v.s. `--options=with=arguments`? I think either we have the API be a list of pairs with mandatory name and optional argument, or we parse the option name out of the option with optional argument.

scholzj: Ok, I improved the parsing to detect the option name and compare it with equals. Please let me know if that works for you.

tombentley (approved):
Thanks!

Issue #4327, opened 2021-01-27
Add failure_reason label to the strimzi_resource_state metric


davidffrench:
**Is your feature request related to a problem? Please describe.**
Currently, the strimzi_resource_state metric reports a boolean state to the readiness of a Kafka cluster. This means it is not currently possible to differentiate between a failed state and an in progress state.

**Describe the solution you'd like**
An easy solution to differentiate between a failed and in progress state is to add a label to the metric with a failure reason which gets added when the resource is in a failed state.

**Describe alternatives you've considered**
A separate metric or using a gauge metric with values for each different state. These are both unnecessary to solve the problem.

sknot-rh:
Do I understand correctly you want the metric to look like this? 
```
strimzi_resource_state{container="strimzi-cluster-operator", endpoint="http", failure_reason="Version 2.8.0 is not supported. Supported versions are: 2.5.0, 2.5.1, 2.6.0, 2.7.0.", instance="10.129.2.121:8080", job="myproject/cluster-operator-metrics", kind="Kafka", name="my-cluster", namespace="myproject", pod="strimzi-cluster-operator-56b59d667d-vj7ln", resource_namespace="myproject"}
--
```

davidffrench:
Thanks for the response @sknot-rh , Yes that looks good to me.

sknot-rh:
Should we reopen this, as th Topic operator part is still opened in https://github.com/strimzi/strimzi-kafka-operator/pull/4502 ?

PR #4354, opened 2021-02-02
State metric contains a message

sknot-rh:

### Type of change
- Enhancement / new feature

### Description
Fixes https://github.com/strimzi/strimzi-kafka-operator/issues/4327

Currently, the state metric gives just info on whether the resource is ready or not. In case of not-ready, we cannot say, whether it is still created or failed. This PR adds a message to the metrics tags.

The metric in case of ready resource:
```
strimzi_resource_state{container="strimzi-cluster-operator", endpoint="http", failure_reason="none", instance="10.129.2.121:8080", job="myproject/cluster-operator-metrics", kind="Kafka", name="my-cluster", namespace="myproject", pod="strimzi-cluster-operator-56b59d667d-vj7ln", resource_namespace="myproject"} | 1
```

The metric in case of *not* ready resource:
```
strimzi_resource_state{container="strimzi-cluster-operator", endpoint="http", failure_reason="Version 2.8.0 is not supported. Supported versions are: 2.5.0, 2.5.1, 2.6.0, 2.7.0.", instance="10.129.2.121:8080", job="myproject/cluster-operator-metrics", kind="Kafka", name="my-cluster", namespace="myproject", pod="strimzi-cluster-operator-56b59d667d-vj7ln", resource_namespace="myproject"} 0
```

### Checklist
- [ ] Write tests
- [ ] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/operator-common/src/main/java/io/strimzi/operator/common/AbstractOperator.java
+++ b/operator-common/src/main/java/io/strimzi/operator/common/AbstractOperator.java
@@ -540,15 +540,23 @@ public abstract class AbstractOperator<
     private void updateResourceState(Reconciliation reconciliation, boolean ready, Throwable cause) {
         String key = reconciliation.namespace() + ":" + reconciliation.kind() + "/" + reconciliation.name();
 
-        Tags metricTagsWithReason = Tags.of(
-                Tag.of("kind", reconciliation.kind()),
-                Tag.of("name", reconciliation.name()),
-                Tag.of("resource-namespace", reconciliation.namespace()),
-                Tag.of("failure-reason", cause == null ? "none" : cause.getMessage() == null ? "unknown error" : cause.getMessage()));
+        Tags metricTags;
+        if (cause == null) {
+            metricTags = Tags.of(
+                    Tag.of("kind", reconciliation.kind()),
+                    Tag.of("name", reconciliation.name()),
+                    Tag.of("resource-namespace", reconciliation.namespace()));
+        } else {
+            metricTags = Tags.of(
+                    Tag.of("kind", reconciliation.kind()),
+                    Tag.of("name", reconciliation.name()),
+                    Tag.of("resource-namespace", reconciliation.namespace()),
+                    Tag.of("reason", cause.getMessage() == null ? "unknown error" : cause.getMessage()));
+        }
 
         T cr = resourceOperator.get(reconciliation.namespace(), reconciliation.name());
 
-        Optional<Meter> gauge = metrics.meterRegistry().getMeters()
+        Optional<Meter> metric = metrics.meterRegistry().getMeters()
                 .stream()
                 .filter(meter -> meter.getId().getName().equals(METRICS_PREFIX + "resource.state") &&
                         meter.getId().getTags().contains(Tag.of("kind", reconciliation.kind())) &&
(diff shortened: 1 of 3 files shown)


operator-common/src/main/java/io/strimzi/operator/common/AbstractOperator.java
```diff
@@ -488,31 +488,38 @@ public AtomicInteger getResourceCounter() {
...
-        Tags metricTags = Tags.of(
+    private void updateResourceState(Reconciliation reconciliation, boolean ready, Throwable cause) {
+        String key = reconciliation.namespace() + ":" + reconciliation.kind() + "/" + reconciliation.name();
+
+        Tags metricTagsWithReason = Tags.of(
                 Tag.of("kind", reconciliation.kind()),
                 Tag.of("name", reconciliation.name()),
-                Tag.of("resource-namespace", reconciliation.namespace()));
+                Tag.of("resource-namespace", reconciliation.namespace()),
+                Tag.of("failure-reason", cause == null ? "none" : cause.getMessage()));
```
ppatierno, line 498: why did you change the key of the map, using a string with the same field unless failure-reason. Why it should not be part of the key in that map? The tags are a way to identify a metric making a kind of compound key. Why failure-reason should be out?

operator-common/src/main/java/io/strimzi/operator/common/AbstractOperator.java
```diff
@@ -488,31 +488,38 @@ public AtomicInteger getResourceCounter() {
...
+                        meter.getId().getTags().contains(Tag.of("name", reconciliation.name())) &&
+                        meter.getId().getTags().contains(Tag.of("resource-namespace", reconciliation.namespace()))
+                ).findFirst();
+
+        if (gauge.isPresent()) {
+            // remove metric so it can be re-added with new tags
+            metrics.meterRegistry().remove(gauge.get().getId());
+            resourcesStateCounter.remove(key);
+            log.debug("{}: Removed metric " + METRICS_PREFIX + "resource.state{}", reconciliation, key);
+        }
```
ppatierno, line 515: why this part is not in the "else" condition when the cr is null?

operator-common/src/main/java/io/strimzi/operator/common/AbstractOperator.java
sknot-rh: If the failure reason changes, the key changes, and thus we have two different metrics.

operator-common/src/main/java/io/strimzi/operator/common/AbstractOperator.java
sknot-rh: Basically we do remove the metric and only if the cr is not null, re-create it.

operator-common/src/main/java/io/strimzi/operator/common/AbstractOperator.java
```diff
@@ -488,31 +488,38 @@ public AtomicInteger getResourceCounter() {
...
+        String key = reconciliation.namespace() + ":" + reconciliation.kind() + "/" + reconciliation.name();
+
+        Tags metricTagsWithReason = Tags.of(
                 Tag.of("kind", reconciliation.kind()),
                 Tag.of("name", reconciliation.name()),
-                Tag.of("resource-namespace", reconciliation.namespace()));
+                Tag.of("resource-namespace", reconciliation.namespace()),
+                Tag.of("failure-reason", cause == null ? "none" : cause.getMessage() == null ? "unknown error" : cause.getMessage()));
 
         T cr = resourceOperator.get(reconciliation.namespace(), reconciliation.name());
```
scholzj, line 500: Do we really need a synchronous get here just to check if the resource is not null? Can't it be derived from the reconciliation in some other way?

operator-common/src/main/java/io/strimzi/operator/common/AbstractOperator.java
```diff
@@ -488,31 +488,38 @@ public AtomicInteger getResourceCounter() {
...
-        Tags metricTags = Tags.of(
+    private void updateResourceState(Reconciliation reconciliation, boolean ready, Throwable cause) {
+        String key = reconciliation.namespace() + ":" + reconciliation.kind() + "/" + reconciliation.name();
+
+        Tags metricTagsWithReason = Tags.of(
                 Tag.of("kind", reconciliation.kind()),
                 Tag.of("name", reconciliation.name()),
-                Tag.of("resource-namespace", reconciliation.namespace()));
+                Tag.of("resource-namespace", reconciliation.namespace()),
+                Tag.of("failure-reason", cause == null ? "none" : cause.getMessage() == null ? "unknown error" : cause.getMessage()));
```
scholzj, line 498: Should the Tag be present if there is no error? What is the value of it?

scholzj (commented):
Should this be added also to the Topic Operator? I don't think that is using this code.

sknot-rh:
@ShubhamRwt wanted to add this to the TO. I am not sure whether it should be done in separate PR or he should push in this branch.

scholzj:
I guess it is fine to do it in a separate PR. But I do not like the idea of having this partially implemented for a release. Given the planned 0.22 release date is coming closer, we should then wait for that Pr being ready as well probably.

ppatierno:
Do we need a test for the new failure message addition?

sknot-rh:
I can try to write one

ppatierno (approved):

scholzj (approved):

scholzj:
**Should be merged only together with the next which does the same for the TO.**

operator-common/src/main/java/io/strimzi/operator/common/AbstractOperator.java
sknot-rh: It seems that Prometheus needs the same set of tags.
```
Prometheus requires that all meters with the same name have the same set of tag keys. There is already an existing meter named 'strimzi_resource_state' containing tag keys [kind, name, reason, resource_namespace]. The meter you are attempting to register has keys [kind, name, resource_namespace]
```

scholzj:
@sknot-rh Looks like this is breaking the regression tests.

PR #4291, opened 2021-01-20
Prepare KRAO for deprecated properties

sknot-rh:
Signed-off-by: Stanislav Knot <sknot@redhat.com>

### Type of change
- Enhancement / new feature

### Description
Fixes https://github.com/strimzi/strimzi-kafka-operator/issues/3637

The `KafkaRebalanceAssemblyOperator` does not put any warning about deprecated properties used in the CR.
Since there is no deprecated property yet, I marked one of the present ones as deprecated just for the test and then reverted it.
The status of the KafkaRebalance CR with deprecated property then looks like this:
```
status:
  conditions:
  - lastTransitionTime: 2021-01-20T12:26:14.219971Z
    status: "True"
    type: ProposalReady
  - lastTransitionTime: 2021-01-20T13:52:50.022206Z
    message: In API version kafka.strimzi.io/v1alpha1 the property goals at path spec.goals
      has been deprecated. This feature should now be configured at path test.
    reason: DeprecatedFields
    status: "True"
    type: Warning
  - lastTransitionTime: 2021-01-20T13:52:50.708662Z
    status: "True"
    type: ProposalReady
```
### Checklist

_Please go through this checklist and make sure all applicable tasks have been done_

- [ ] Write tests
- [ ] Make sure all tests pass
- [ ] Update documentation
- [ ] Check RBAC rights for Kubernetes / OpenShift roles
- [ ] Try your changes from Pod inside your Kubernetes and OpenShift cluster, not just locally
- [ ] Reference relevant issue(s) and close them after merging
- [ ] Update CHANGELOG.md
- [ ] Supply screenshots for visual changes, such as Grafana dashboards

PR #4357, opened 2021-02-02
Fix KafkaRebalance status when deprecated or unknown property is in spec

sknot-rh:

### Type of change
- Bugfix

### Description
KafkaRebalanceAbstractOperator in difference to other operators does not create a new status for each reconciliation. Instead of that, it reuses the previous one. In this PR, new status is built but data from the old one is used. 
Status like bellow does not appear anymore. Also removing non-valid conditions work now (e.g. unknown property is removed)

```
  - lastTransitionTime: 2021-02-02T11:34:16.034937Z
    message: 'Contains object at path spec with an unknown property: asd'
    reason: UnknownFields
    status: "True"
    type: Warning
  - lastTransitionTime: 2021-02-02T11:34:16.052363Z
    message: 'Contains object at path spec with an unknown property: asd'
    reason: UnknownFields
    status: "True"
    type: Warning
```

### Checklist
- [x] Write tests
- [ ] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceAssemblyOperator.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceAssemblyOperator.java
@@ -48,6 +48,7 @@ import io.vertx.core.json.JsonObject;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 
+import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
 import java.util.Map;
@@ -320,8 +321,7 @@ public class KafkaRebalanceAssemblyOperator
 
     private Future<Void> reconcile(Reconciliation reconciliation, String host,
                                    CruiseControlApi apiClient, KafkaRebalance kafkaRebalance,
-                                   KafkaRebalanceState currentState, KafkaRebalanceAnnotation rebalanceAnnotation,
-                                   Set<Condition> unknownAndDeprecatedConditions) {
+                                   KafkaRebalanceState currentState, KafkaRebalanceAnnotation rebalanceAnnotation) {
 
         log.info("{}: Rebalance action from state [{}]", reconciliation, currentState);
 
@@ -335,7 +335,6 @@ public class KafkaRebalanceAssemblyOperator
                return kafkaRebalanceOperator.getAsync(reconciliation.namespace(), reconciliation.name())
                             .compose(currentKafkaRebalance -> {
                                 if (currentKafkaRebalance != null) {
-                                    addWarningsToStatus(desiredStatus, unknownAndDeprecatedConditions);
                                     return updateStatus(currentKafkaRebalance, desiredStatus, null)
                                             .compose(updatedKafkaRebalance -> {
                                                 log.info("{}: State updated to [{}] with annotation {}={} ",
@@ -367,7 +366,6 @@ public class KafkaRebalanceAssemblyOperator
                                 }
                             }, exception -> {
                                     log.error("{}: Status updated to [NotReady] due to error: {}", reconciliation, exception.getMessage());
-                                    addWarningsToStatus(desiredStatus, unknownAndDeprecatedConditions);
                                     return updateStatus(kafkaRebalance, new KafkaRebalanceStatus(), exception)
                                             .mapEmpty();
                                 }); },
@@ -388,7 +386,7 @@ public class KafkaRebalanceAssemblyOperator
                                                                         KafkaRebalanceAnnotation rebalanceAnnotation, RebalanceOptions.RebalanceOptionsBuilder rebalanceOptionsBuilder) {
         switch (currentState) {
             case New:
-                return onNew(reconciliation, host, apiClient, rebalanceOptionsBuilder);
+                return onNew(reconciliation, host, apiClient, rebalanceOptionsBuilder, kafkaRebalance);
             case PendingProposal:
                 return onPendingProposal(reconciliation, host, apiClient, kafkaRebalance, rebalanceAnnotation, rebalanceOptionsBuilder);
             case ProposalReady:
@@ -396,10 +394,10 @@ public class KafkaRebalanceAssemblyOperator
             case Rebalancing:
                 return onRebalancing(reconciliation, host, apiClient, kafkaRebalance, rebalanceAnnotation);
             case Stopped:
-                return onStop(reconciliation, host, apiClient, rebalanceAnnotation, rebalanceOptionsBuilder);
+                return onStop(reconciliation, host, apiClient, rebalanceAnnotation, rebalanceOptionsBuilder, kafkaRebalance);
             case Ready:
                 // Rebalance Complete
-                return Future.succeededFuture(kafkaRebalance.getStatus());
+                return Future.succeededFuture(buildRebalanceStatusFromPreviousStatus(kafkaRebalance.getStatus(), validate(kafkaRebalance)));
             case NotReady:
                 // Error case
                 return onNotReady(reconciliation, host, apiClient, kafkaRebalance, rebalanceAnnotation, rebalanceOptionsBuilder);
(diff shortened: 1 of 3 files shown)


tomncooper (approved):

operator-common/src/main/java/io/strimzi/operator/common/operator/resource/StatusUtils.java
```diff
@@ -100,19 +100,30 @@ public static Condition buildRebalanceCondition(String type) {
...
         List<Condition> conditionList = new ArrayList<>();
         Condition condition = StatusUtils.buildCondition(type, conditionStatus, null);
         conditionList.add(condition);
 
         if (status.getConditions() != null) {
             status.getConditions().forEach(cond -> {
                 if ("UnknownFields".equals(cond.getReason()) || "DeprecatedFields".equals(cond.getReason())) {
-                    conditionList.add(cond);
+                    if (!conditionAlreadyPresent(conditionList, cond)) {
+                        conditionList.add(cond);
```
ppatierno, line 112: just trying to understand, why we should not put an already existing condition. If the same condition is happening, we should have the same one but with a different timestamp to show its latest update.

operator-common/src/main/java/io/strimzi/operator/common/operator/resource/StatusUtils.java
```diff
@@ -100,19 +100,30 @@ public static Condition buildRebalanceCondition(String type) {
...
+        status.setConditions(new ArrayList<>(conditionList));
+    }
 
-        status.setConditions(conditionList);
+    private static boolean conditionAlreadyPresent(List<Condition> conditions, Condition condition) {
+        return conditions.stream().filter(cond -> {
+            return condition.getReason().equals(cond.getReason()) &&
+                    condition.getType().equals(cond.getType()) &&
+                    condition.getStatus().equals(cond.getStatus()) &&
+                    condition.getMessage().equals(cond.getMessage());
```
ppatierno, line 125: isn't it enough to filter on type? That should be the "key", or?

sknot-rh: Type is for example a `warning`. So using it as a key would result to having just one warning even in the case when there are more ones. I think more suitable one is `reason`.

operator-common/src/main/java/io/strimzi/operator/common/operator/resource/StatusUtils.java
sknot-rh: It would result in this: https://gist.github.com/im-konge/b26e068fce4c5c1395717ef7c4a807ab

ppatierno: how does it work today without warnings then?

operator-common/src/main/java/io/strimzi/operator/common/operator/resource/StatusUtils.java
ppatierno: Yes maybe

operator-common/src/main/java/io/strimzi/operator/common/operator/resource/StatusUtils.java
sknot-rh: Before #4291 there is nothing besides of Proposal status.

ppatierno: this is part of a proposal status but we didn't have these warnings conditions and the condition was always just one, or? Now we are moving to have more conditions then? @tomncooper do I remember wrong?

sknot-rh: Correct. The issue is described in https://github.com/strimzi/strimzi-kafka-operator/issues/3637

operator-common/src/main/java/io/strimzi/operator/common/operator/resource/StatusUtils.java
```diff
@@ -100,19 +100,24 @@ public static Condition buildRebalanceCondition(String type) {
...
         }
+        // condition contains a timestamp so using a set does not help
         List<Condition> conditionList = new ArrayList<>();
         Condition condition = StatusUtils.buildCondition(type, conditionStatus, null);
         conditionList.add(condition);
 
-        if (status.getConditions() != null) {
-            status.getConditions().forEach(cond -> {
-                if ("UnknownFields".equals(cond.getReason()) || "DeprecatedFields".equals(cond.getReason())) {
+        status.getConditions().forEach(cond -> {
```
scholzj, line 108: Why do you think we don't need the null check anymore?

operator-common/src/main/java/io/strimzi/operator/common/operator/resource/StatusUtils.java
```diff
@@ -100,19 +100,24 @@ public static Condition buildRebalanceCondition(String type) {
...
+            if ("UnknownFields".equals(cond.getReason()) || "DeprecatedFields".equals(cond.getReason())) {
```
scholzj, line 109: Should `DeprecatedObjects` be added here as well to this condition?

operator-common/src/main/java/io/strimzi/operator/common/operator/resource/StatusUtils.java
```diff
@@ -100,19 +100,24 @@ public static Condition buildRebalanceCondition(String type) {
...
+                if (!conditionAlreadyPresent(conditionList, cond)) {
+                    cond.setLastTransitionTime(iso8601Now());
                     conditionList.add(cond);
                 }
-            });
-        }
+            }
```
scholzj, line 114: Could you please explain what this code does? It is common to have the messages with the same reason present multiple times. For example:

```yaml
    - lastTransitionTime: "2021-02-15T13:05:24.039453Z"
      message: In API version kafka.strimzi.io/v1beta1 the metrics property at path
        spec.kafka.metrics has been deprecated, and should now be configured using
        spec.kafka.metricsConfig.
      reason: DeprecatedFields
      status: "True"
      type: Warning
    - lastTransitionTime: "2021-02-15T13:05:24.039453Z"
      message: In API version kafka.strimzi.io/v1beta1 the metrics property at path
        spec.zookeeper.metrics has been deprecated, and should now be configured using
        spec.zookeeper.metricsConfig.
      reason: DeprecatedFields
      status: "True"
      type: Warning
```

So I'm not entirely sure if what you are doing here is right.

ppatierno:
I have no more comments than the ones from Jakub. I will wait for the answers there.

operator-common/src/main/java/io/strimzi/operator/common/operator/resource/StatusUtils.java
sknot-rh: https://gist.github.com/im-konge/b26e068fce4c5c1395717ef7c4a807ab

scholzj: That is without of context => so hard to understand the issue there. But I do not think it goes against my comment => I'm not arguing that you should have there the same warning 100 times. I'm saying that the Reason is not expected to be unique in general and should not be treated as such.

Yesterday when I checked this code in IDE it seemed to be used by KafkaRebalance only ... so maybe it never worked properly 🤷🏻‍♂️.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceAssemblyOperator.java
```diff
@@ -382,18 +380,18 @@ private String rebalanceStateConditionType(KafkaRebalanceStatus status) {
...
+                return onNew(reconciliation, host, apiClient, rebalanceOptionsBuilder, kafkaRebalance);
             case PendingProposal:
                 return onPendingProposal(reconciliation, host, apiClient, kafkaRebalance, rebalanceAnnotation, rebalanceOptionsBuilder);
             case ProposalReady:
                 return onProposalReady(reconciliation, host, apiClient, kafkaRebalance, rebalanceAnnotation, rebalanceOptionsBuilder);
             case Rebalancing:
                 return onRebalancing(reconciliation, host, apiClient, kafkaRebalance, rebalanceAnnotation);
             case Stopped:
-                return onStop(reconciliation, host, apiClient, rebalanceAnnotation, rebalanceOptionsBuilder);
+                return onStop(reconciliation, host, apiClient, rebalanceAnnotation, rebalanceOptionsBuilder, kafkaRebalance);
```
scholzj, line 391: Should we try to keep the same order of the parameters here? I see 4 methods hare which have already `reconciliation, host, apiClient, kafkaRebalance, ...`. So maybe you can insert it after the `apiClient` in the two new places as well?

scholzj (commented):
One nit, but LGTM otherwise. Maybe we can also have some test for it?

api/src/test/resources/io/strimzi/api/kafka/model/047-Crd-kafkaconnector.yaml
```diff
@@ -110,4 +110,9 @@ spec:
               tasksMax:
                 type: integer
                 description: The maximum number of tasks for the Kafka Connector.
+              topics:
+                type: array
+                items:
+                  type: string
+                description: The list of topics used by the Kafka Connector.
```
ppatierno, line 117: how is that related? a rebase problem?

sknot-rh: Yeah

ppatierno (approved):

scholzj (approved):

PR #4375, opened 2021-02-08
Add support for Ingress V1

scholzj:
### Type of change

- Enhancement / new feature

### Description

Kubernetes 1.19 introduced `v1` of the Ingress resource. The previous versions (`extension/v1beta1` and `networking/v1beta1`) will be removed in Kubernetes 1.22. This is relatively short time. This PR adds support for v1 Ingress but keeps the `networking/v1beta1` support present to keep it working on Kubernetes 1.16-1.18. The way it is designed makes it easy to remove the v1beta1 part in the future.

Kubernetes now also have new resource IngressClass which has the concept of default Ingress class. To make this work for the v1 Ingress in Strimzi, this also changes the default `nginx` class which was always set in the previous Strimzi versions.

This has been tested manually against Kube 1.20 and 1.16 to test both v1 and v1beta1 variants.

### Checklist

- [x] Write tests
- [x] Update documentation
- [x] Update CHANGELOG.md

### DIFF
--- a/api/src/main/java/io/strimzi/api/kafka/model/listener/KafkaListenerExternalIngress.java
+++ b/api/src/main/java/io/strimzi/api/kafka/model/listener/KafkaListenerExternalIngress.java
@@ -80,8 +80,7 @@ public class KafkaListenerExternalIngress extends KafkaListenerExternal {
         this.configuration = configuration;
     }
 
-    @Description("Configures the `Ingress` class that defines which `Ingress` controller will be used. " +
-            "If not set, the `Ingress` class is set to `nginx`.")
+    @Description("Configures the `Ingress` class that defines which `Ingress` controller will be used.")
     @JsonInclude(JsonInclude.Include.NON_NULL)
     @JsonProperty("class")
     public String getIngressClass() {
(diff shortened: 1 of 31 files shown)


api/src/main/java/io/strimzi/api/kafka/model/listener/KafkaListenerExternalIngress.java
```diff
@@ -80,8 +80,7 @@ public void setConfiguration(IngressListenerConfiguration configuration) {
         this.configuration = configuration;
     }
 
-    @Description("Configures the `Ingress` class that defines which `Ingress` controller will be used. " +
-            "If not set, the `Ingress` class is set to `nginx`.")
+    @Description("Configures the `Ingress` class that defines which `Ingress` controller will be used.")
```
ppatierno, line 83: does it mean that it's mandatory to be set now? what happens if not set and not default to nginx anymore?

scholzj: No, it is not mandatory now. But the default value in the past was `nginx`. Now it will have no default value and when not configured it will not be set at all (which in my case for example doesn't matter).

scholzj: I improved the description to make it more clear.

ppatierno (approved):

PR #4379, opened 2021-02-08
CA:true

tombentley:
### Type of change

- Bugfix

### Description

When renewing self-signed certs we weren't generating certs with `CA:true` or the `pathlen` constraint. Doing that requires passing an openssl config file, which previously we only did when the SAN were specified. 

This PR:

* Changes how we construct the renewed certificate using the `openssl` commands described https://serverfault.com/a/501513. Previously we used the accepted answer to that same serverfault question, but it doesn't work for passing the constraints for the new cert.
* Refactors how we build the config file so that we don't always need to have SAN specified in order for the config to be generated.
* Simplifies how the file is constructed to avoid a redundant file copy (we were writing the resource to a file, then copying it and appending the extra bits).
* Generally tries to tidy up the code a bit. 
* Adds a couple of tests and some assertions to the existing tests

Note that it also sets the `pathlen=0` where previously our CA certs got the default pathlen (which meant people could generate intermediate CAs, signed by our root CA). If there are usecases where people in fact use intermediate CAs we should change pathlen.

### DIFF
--- a/certificate-manager/src/main/java/io/strimzi/certs/OpenSslCertManager.java
+++ b/certificate-manager/src/main/java/io/strimzi/certs/OpenSslCertManager.java
@@ -180,40 +180,14 @@ public class OpenSslCertManager implements CertManager {
 
         //openssl req -new -key root.key -out newcsr.csr
         File csrFile = File.createTempFile("renewal", ".csr");
-        List<String> cmd = new ArrayList<>(asList("openssl", "req",
-                "-new",
-                "-batch",
-                "-out", csrFile.getAbsolutePath(),
-                "-key", keyFile.getAbsolutePath()));
 
-        File sna = null;
-        File openSslConf = null;
-        if (sbj != null) {
-            if (sbj.subjectAltNames() != null && sbj.subjectAltNames().size() > 0) {
-
-                // subject alt names need to be in an openssl configuration file
-                InputStream is = getClass().getClassLoader().getResourceAsStream("openssl.conf");
-                openSslConf = File.createTempFile("openssl-", ".conf");
-                Files.copy(is, openSslConf.toPath(), StandardCopyOption.REPLACE_EXISTING);
-
-                sna = addSubjectAltNames(openSslConf, sbj);
-                cmd.addAll(asList("-config", sna.toPath().toString(), "-extensions", "v3_req"));
-            }
-            cmd.addAll(asList("-subj", sbj.toString()));
-        }
+        List<String> cmd = new ArrayList<>(asList("openssl", "x509",
+                "-x509toreq",
+                "-in", certFile.getAbsolutePath(),
+                "-signkey", keyFile.getAbsolutePath(),
+                "-out", csrFile.getAbsolutePath()));
 
         exec(cmd);
-
-        if (sna != null) {
-            if (!sna.delete()) {
-                log.warn("{} cannot be deleted", sna.getName());
-            }
-        }
-        if (openSslConf != null) {
-            if (!openSslConf.delete()) {
-                log.warn("{} cannot be deleted", openSslConf.getName());
-            }
-        }
 
         //openssl x509 -req -days 3650 -in newcsr.csr -signkey root.key -out newroot.pem
         List<String> cmd2 = new ArrayList<>(asList("openssl", "x509",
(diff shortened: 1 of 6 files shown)


scholzj (approved):
LGTM. Although @ppatierno probably knows this area more than I do ... so his review might be more important.

certificate-manager/src/main/java/io/strimzi/certs/OpenSslCertManager.java
```diff
@@ -176,45 +144,26 @@ public void deleteFromTrustStore(List<String> aliases, File trustStoreFile, Stri
 
     @Override
     public void renewSelfSignedCert(File keyFile, File certFile, Subject sbj, int days) throws IOException {
-        // See https://serverfault.com/questions/306345/certification-authority-root-certificate-expiry-and-renewal
+        // See https://serverfault.com/a/501513
 
         //openssl req -new -key root.key -out newcsr.csr
-        File csrFile = File.createTempFile("renewal", ".csr");
+        File csrFile = Files.createTempFile("renewal", ".csr").toFile();
```
ppatierno, line 150: why this change? What's the actual difference between `createTempFile` in `File` vs `Files` (other than the need to call one more `toFile` on the second one?)

certificate-manager/src/main/java/io/strimzi/certs/OpenSslCertManager.java
```diff
@@ -341,23 +283,26 @@ public void generateCert(File csrFile, File caKey, File caCert, File crtFile, Su
...
 
         // We need to remove CA serial file
         Files.deleteIfExists(Paths.get(caCert.getPath().replace(".crt", ".srl")));
     }
 
+    private Path createDefaultConfig() throws IOException {
+        try (InputStream is = getClass().getClassLoader().getResourceAsStream("openssl.conf")) {
+            Path openSslConf = Files.createTempFile("openssl-", ".conf");
+            Files.copy(is, openSslConf, StandardCopyOption.REPLACE_EXISTING);
+            return openSslConf;
```
ppatierno, line 295: again here, I see that where we use `createDefaultConfig`, we need a `File` to we call `toFile` on the returned `Path`. Why not return the `File` instance directly?

certificate-manager/src/test/java/io/strimzi/certs/OpenSslCertManagerTest.java
```diff
@@ -208,30 +235,25 @@ private void testGenerateSignedCert(File caKey, File caCert, Subject caSbj, File
 
         ssl.addKeyAndCertToKeyStore(caKey, caCert, "ca", keyStore, keyStorePassword);
 
-        CertificateFactory cf = CertificateFactory.getInstance("X.509");
-        Certificate c = cf.generateCertificate(new FileInputStream(cert));
-        Certificate ca = cf.generateCertificate(new FileInputStream(caCert));
+        X509Certificate c = loadCertificate(cert);
+        assertCaCertificate(c, false);
```
ppatierno, line 239: I am missing something, isn't this going to fail anyway because of the `false` parameter you are passing in?

certificate-manager/src/main/java/io/strimzi/certs/OpenSslCertManager.java
tombentley: The javadoc for `File.createTempFile` includes:

> The Files.createTempFile method provides an alternative method to create an
> empty file in the temporary-file directory. Files created by that method
> may have more restrictive access permissions to files created by this
> method and so may be more suited to security-sensitive applications.

IDK the details, and maybe in a container there's no real difference, but since we're messing around with keys in this code, I felt this was worth changing, just to be on the safe side.

certificate-manager/src/main/java/io/strimzi/certs/OpenSslCertManager.java
tombentley: Same reason really, and I found using Paths everywhere was just easier.

certificate-manager/src/test/java/io/strimzi/certs/OpenSslCertManagerTest.java
tombentley: The false means we're asserting that the issued cert is not a CA cert (i.e. _does_ not have `CA: true`). This is in a test about the issued certs, not the CA certs. The issued certs should not have `CA: true` because they're strictly client certs.

ppatierno (approved):

PR #4392, opened 2021-02-10
[DOC] Use tls: true in Ingress examples

scholzj:
### Type of change

- Bugfix

### Description

Ingress type listeners require TLS to be always enabled because it uses TLS passthrough. So we should use `tls: true` in all the examples.

### Checklist

- [x] Update documentation


ppatierno (approved):

PaulRMellor (approved):
:+1:

PR #4405, opened 2021-02-12
Use images from quay in strimzi-test-container

Frawless:

### Type of change

- Bugfix

### Description

We should use images from quay in test-container istead of images from docker hub.

### Checklist


see-quick (approved):
Can you please add commend to testStartContainer that if you want to work it properly it is mandatory that you have set `TESTCONTAINERS_RYUK_DISABLED` env to `true`. Anyway LGTM

.azure/build-pipeline.yaml
```diff
@@ -34,11 +34,16 @@ jobs:
     # Pipeline steps
     steps:
       - template: "templates/general_steps.yaml"
+
+      - bash: "mvn test -f test-container/pom.xml"
+        displayName: Run tests for strimzi-test-container
+
```
scholzj, line 40: I think this should be after the cache task. I also wonder why does this not run as part of the regular build. Wouldn't it be easier to have it just included as part of it?

Frawless: Unit and ITs are executed during image build where not all packages are built. `test-container` build is executed during  `build_strimzi.yaml` execution, however, here we skip tests. TBH I am not sure what's the right solution here. Maybe execute unit and IT tests during the project build and during image build just skip them?

scholzj: I remembered probably what the problem was. The containers are built at the end of the pipeline. So when you run this as part of the regular build, it will possibly not have the images yet. So if we want to have it in the pipeline, it should be at the end after the regular build when the images are already built I guess?

Frawless: yes make sense, just thinking if the images are by default built as `quay.io/strimzi/operator...` and then the tag is done for `DOCKER_REGISTRY` etc. because if it's built already with different org, it won't work for test-containers anyway since the org is hardcoded here.

scholzj: AFAIK they are built under different named by later retagged for the real name. However, I guess they will almost always use the `latest` tag. How does the test-container deal with Strimzi versions?

Frawless: you can specify a version as a parameter passed to the constructor

scholzj: But that is version of Kafka, not Strimzi, or?

Frawless: you can pass whole tag like `0.20.0-kafka-2.6.0`

scholzj (approved):

Issue #4394, opened 2021-02-11
Make additional labels configurable


maxbrunet:
**Is your feature request related to a problem? Please describe.**

We use Argo CD which tracks resources it manages by injecting a label:
https://argoproj.github.io/argo-cd/faq/#why-is-my-app-out-of-sync-even-after-syncing
The operator propagates the label from the `Kafka` resource to the `kafka-init` ClusterRoleBinding resource, and Argo CD wants to delete the later since it does not manage it.  

In our case this Argo CD label is customized to `app.argoproj.io/instance` (instead of `app.kubernetes.io/instance` which we already use in our manifests).

**Describe the solution you'd like**

Making this condition configurable could solve that:
https://github.com/strimzi/strimzi-kafka-operator/blob/192ec6a495212d3daaad31af167b5c771786648b/operator-common/src/main/java/io/strimzi/operator/common/model/Labels.java#L144

The `Kafka` resource could have two fields:

* `/spec/resourceLabels/filters/exludeStartsWith` (default: `["app.kubernetes.io/"]`)
* `/spec/resourceLabels/filters/keep` (default: `["app.kubernetes.io/part-of"]`)

This should be backward compatible and setting them to an empty list `[]` would disable them.

**Describe alternatives you've considered**

* Take a glob/regex pattern to filter labels
* Have a label whitelist/blacklist mechanism

But overall, I think propagating labels is not something operators do as it creates more issues than it solves, so maybe:

* supporting `somethingMetadata` fields would be a simpler approach, like `podMetadata`, `initClusterRoleBindingMetadata`... and let the user decide what labels and annotations are set (Helm/Kustomize/... can take of that)?

**Additional context**

Discussion started here #4039

PR #4411, opened 2021-02-13
Make it possible to use a Label selector to operate only some CRs

scholzj:
### Type of change

- Enhancement / new feature

### Description

Sometimes, it could be handy to be able to specify a selector which will be used by the operator and make it see only some resources. This can be used for example for A-B testing, blue-green deployments etc. This PR adds support for such selector for all cluster based resources (Kafka, KafkaConnect, KafkaConnectS2I, KafkaMirrorMaker, KafkaMirrorMaker2, KafkaBridge). The KafkaConnector and KafkaRebalance resources will be operated by the operator which also operates the main resource they belong to.

Right now, the labels used by the selector propagate into the operand resources. That means that while handing over between operator instances, the operands will be always rolled. We should figure out how to avoid this, but that should be done in a separate PR - perhaps while addressing #4394.

### Checklist

- [x] Write tests
- [x] Update documentation

### DIFF
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/ClusterOperatorConfig.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/ClusterOperatorConfig.java
@@ -345,8 +345,7 @@ public class ClusterOperatorConfig {
     }
 
     /**
-     * Parse labels from String into the Labels format. This is used to parse the custom resource labels and the
-     * namespace labels.
+     * Parse labels from String into the Labels format.
      *
      * @param vars              Map with configuration variables
      * @param configurationKey  Key containing the string with labels
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/ClusterOperatorConfig.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/ClusterOperatorConfig.java
@@ -42,6 +42,7 @@ public class ClusterOperatorConfig {
     public static final String STRIMZI_IMAGE_PULL_SECRETS = "STRIMZI_IMAGE_PULL_SECRETS";
     public static final String STRIMZI_OPERATOR_NAMESPACE = "STRIMZI_OPERATOR_NAMESPACE";
     public static final String STRIMZI_OPERATOR_NAMESPACE_LABELS = "STRIMZI_OPERATOR_NAMESPACE_LABELS";
+    public static final String STRIMZI_CUSTOM_RESOURCE_SELECTOR = "STRIMZI_CUSTOM_RESOURCE_SELECTOR";
 
     // Feature Flags
     public static final String STRIMZI_RBAC_SCOPE = "STRIMZI_RBAC_SCOPE";
@@ -81,6 +82,7 @@ public class ClusterOperatorConfig {
     private final String operatorNamespace;
     private final Labels operatorNamespaceLabels;
     private final RbacScope rbacScope;
+    private final Labels customResourceSelector;
 
     /**
      * Constructor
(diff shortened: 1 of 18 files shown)


cluster-operator/src/main/java/io/strimzi/operator/cluster/ClusterOperatorConfig.java
```diff
@@ -337,18 +344,27 @@ private static ImagePullPolicy parseImagePullPolicy(String imagePullPolicyEnvVar
         return imagePullSecrets;
     }
 
-    private static Labels parseOperatorNamespaceLabels(String operatorNamespaceLabelsString) {
-        Labels operatorNamespaceLabels = null;
+    /**
+     * Parse labels from String into the Labels format. This is used to parse the custom resource labels and the
```
ppatierno, line 348: I would avoid adding "This is used to ... " because at some point we could use this method for parsing some other labels and if the JavaDoc is not updated it would be stale.

operator-common/src/main/java/io/strimzi/operator/common/AbstractOperator.java
```diff
@@ -189,6 +193,13 @@ private String getLockName(String namespace, String name) {
             T cr = resourceOperator.get(namespace, name);
 
             if (cr != null) {
+                if (!Util.matchesSelector(selector(), cr))  {
+                    // The the labels matching the selector are removed, the DELETED event is triggered. We have to
+                    // filter these situations out and ignore the reconciliation.
+                    log.debug("{}: {} {} in namespace {} does not match label selector {} and will be ignored", reconciliation, kind(), name, namespace, selector().get().getMatchLabels());
```
ppatierno, line 199: can you elaborate a little bit on the comment? It's not clear to me. This is the case where the CR exists but doesn't match the label selectors so we are ignoring it. How the comment is related to this?

scholzj: The comment is there because this is tricky and possibly unexpected. The watch is configured to watch for resources with some labels. So one would expect to get only events when the resource has these labels. But that is not the case. If you remove the label from the CR, it triggers a DELETE event for the watch despite the resource not being technically deleted and the CR labels not matching the watch filter anymore. So we need to filter these situations out and that is why this additional check had to be added there. 

And because I think it is not obvious why we check labels again in a watch which already has them in the selector I added the comment there

scholzj: I updated the comment to make it a bit more clear.

ppatierno: This triggered DELETE is something "bad" in K8S API server or fabric8 watcher? Doesn't sound normal to trigger a DELETE because labels are removed :-o

scholzj: That would not be Fabric8 but Kube API server if something. But I think this is a correct and desired behaviour. Essentially, by deleting the labels, you delete the resource from the watch. So I think it makes sense and there will be many cases where this would be actually important (even in Strimzi, I imagine in User or Topic Operators or in the Connector watch this is exactly what we want). But in our use-case with the handover it does not work so we need to identify it and ignore it.

ppatierno: Tbh don't know if it's really correct. You are actually removing it from the watcher so "not watching it anymore" but the resource is still in etcd and a DELETE event should be related to what happens to the resource in etcd I think.
Anyway, there is nothing we can do, we can just deal with it.

scholzj: Ok, great. In that case your comments should be incorporated.

documentation/modules/ref-operator-cluster.adoc
```diff
@@ -45,6 +45,27 @@ env:
...
+[source,yaml,options="nowrap"]
+----
+env:
+  - name: STRIMZI_OPERATOR_NAMESPACE_LABELS
+    value: label1=value1,label2=value2
+----
+
+`STRIMZI_CUSTOM_RESOURCE_SELECTOR`:: Optional.
+Specifies label selector used to filter the custom resources handled by the operator.
+The operator will operate only the custom resource which will have the labels set.
```
tomncooper, line 58: "The operator will operate only on those custom resources which will have the specified labels set."

tomncooper (approved):

ppatierno (approved):

PR #4415, opened 2021-02-15
[DOC] Topic Operator topic store

PaulRMellor:

**Documentation**

New sections to describe the Topic Operator _topic store_:

**Using Guide**
- _Topic Operator topic store_ (con-topic-operator-store.adoc): new section to describe the topic store
- _Handling changes to topics_ moves to new section^^

**Deploying Guide**
- _Downgrading Kafka brokers and client applications_: new step to delete topic store topics

**Overview**
- _Topic Operator_: Brief description of the topic store + link

Plus restructuring of files to move the operator content into assembly and module folders

### Checklist

- [ ] Write tests
- [x] Update documentation
- [ ] Update CHANGELOG.md


documentation/modules/upgrading/proc-downgrade-brokers-older-kafka.adoc
```diff
@@ -75,3 +76,16 @@ Reconciliation #_NUM_(watch) Kafka(_NAMESPACE_/_NAME_): Kafka version downgrade
...
 The Kafka cluster and clients are now using the previous Kafka version.
+
+. If you are reverting back to a version of Strimzi that uses ZooKeeper for the storage of topic metadata, delete the internal topic store topics from the Kafka cluster.
++
+For example, where _my-cluster-kafka-0_ is the name of the broker pod:
++
+[source,shell]
+----
+kubectl exec -ti my-cluster-kafka-0 -- ./bin/kafka-topics.sh --bootstrap-server localhost:9092 --topic __strimzi-topic-operator-kstreams-topic-store-changelog --delete && ./bin/kafka-topics.sh --bootstrap-server localhost:9092 --topic __strimzi_store_topic --delete
+----
```
PaulRMellor, line 87: Right place? Do we want to say _Optional_ or _earlier than 0.22_?

documentation/modules/managing/proc-cluster-recovery-volume.adoc
```diff
@@ -193,20 +193,15 @@ kubectl apply -f _kafka.yaml_
...
-----
-kubectl exec -ti _my-cluster-zookeeper-0_ -- bin/zookeeper-shell.sh localhost:12181
-----
-. Delete the whole `/strimzi` path to remove the Topic Operator storage:
-+
-[source,shell,subs="+quotes,attributes"]
+[source,shell]
 ----
-deleteall /strimzi
+kubectl exec -ti my-cluster-kafka-0 -- ./bin/kafka-topics.sh --bootstrap-server localhost:9092 --topic __strimzi-topic-operator-kstreams-topic-store-changelog --delete && ./bin/kafka-topics.sh --bootstrap-server localhost:9092 --topic __strimzi_store_topic --delete
```
scholzj, line 202: We should use the following:

```
kubectl -n kafka run kafka-admin -ti --image={DockerKafkaImageCurrent} --rm=true --restart=Never -- ./bin/kafka-topics.sh --bootstrap-server localhost:9092 --topic __strimzi-topic-operator-kstreams-topic-store-changelog --delete && ./bin/kafka-topics.sh --bootstrap-server localhost:9092 --topic __strimzi_store_topic --delete
```

We should probably also add some caveat that when needed the command has to be adapted to correspond to user's configuration of listeners and authentication.

documentation/modules/operators/con-topic-operator-store.adoc
```diff
@@ -0,0 +1,115 @@
+// Module included in the following assemblies:
+//
+// assembly-using-the-topic-operator.adoc
+
+[id='ref-topic-operator-store-{context}']
+= Topic Operator topic store
+
+[role="_abstract"]
+The Topic Operator uses persistent storage to store topic metadata describing topic configuration as key-value pairs.
+The _topic store_ is based on the Kafka Streams key-value mechanism, which uses persistent storage with local and remote components.
```
scholzj, line 10: The term `persistent storage` is misleading here since the topic operator does not use persistent storage. Maybe this should say something like this? Also, no clue what _persistent storage with local and remote components_ means.

```suggestion
The Topic Operator uses Kafka to store topic metadata describing topic configuration as key-value pairs.
The _topic store_ is based on the Kafka Streams key-value mechanism, which uses Kafka topics to persist the state.
```

documentation/modules/operators/con-topic-operator-store.adoc
```diff
@@ -0,0 +1,115 @@
...
+
+Topic metadata is stored and accessed locally in-memory.
+Updates from operations applied to the local in-memory topic store are persisted to a backup topic store on disk.
```
scholzj, line 13: This is again confusing since the previous paragraph just explained how it uses persistent storage and here it suddenly said it is in-memory. Maybe this would be better?

```suggestion
Topic metadata are cached in-memory and accessed locally.
Updates from operations applied to the local in-memory cache are persisted to a backup topic store on disk.
```

documentation/modules/operators/con-topic-operator-store.adoc
```diff
@@ -0,0 +1,115 @@
...
+The topic store is continually synchronized with updates from Kafka topics or Kubernetes `KafkaTopic` custom resources.
+Operations are handled rapidly with the topic store set up this way,
+but should the in-memory topic store crash it is automatically repopulated from the persistent storage.
```
scholzj, line 16:
```suggestion
but should the in-memory cache crash it is automatically repopulated from the persistent storage.
```

documentation/modules/operators/con-topic-operator-store.adoc
```diff
@@ -0,0 +1,115 @@
...
+
+== Internal topic store topics
+
+Internal topics support the handling of topic metadata.
+
+`__strimzi_store_topic`:: Input topic for storing the topic metadata
+`__strimzi-topic-operator-kstreams-topic-store-changelog`:: Retains a log of compacted topic store values
```
scholzj, line 23: Can we add something like:

```
IMPORTANT: Do not delete these topics. Their deletion can cause Topic Operator failures.
```

documentation/modules/operators/con-topic-operator-store.adoc
```diff
@@ -0,0 +1,115 @@
...
+
+== Migrating topic metadata from ZooKeeper
+
+In previous releases of Strimzi, topic metadata was stored in ZooKeeper.
+The new process removes this requirement, bringing the metadata into the Kafka cluster, and under the control of the Topic Operator.
+
+When upgrading to Strimzi {ProductVersion}, the transition to Topic Operator control of the topic store is seamless.
+Metadata is found and migrated from ZooKeeper, and the old store is cleansed.
```
scholzj, line 31: Is the old store deleted? Or just cleansed? I think that for lot of Kafka users this might mean two different things.

documentation/modules/operators/con-topic-operator-store.adoc
```diff
@@ -0,0 +1,115 @@
...
+If you are reverting back to a version of Strimzi that uses ZooKeeper for the storage of topic metadata,
+you downgrade your Cluster Operator to the previous version,
+then downgrade Kafka brokers and client applications to the previous Kafka version as standard.
+
+Additionally, you must delete the topics created for the topic store using an `exec` command, specifying the bootstrap address of the Kafka cluster and the name of a broker pod.
+For example, where `my-cluster-kafka-0` is the name of the broker pod:
+
+[source,shell]
+----
+kubectl exec -ti my-cluster-kafka-0 -- ./bin/kafka-topics.sh --bootstrap-server localhost:9092 --topic __strimzi-topic-operator-kstreams-topic-store-changelog --delete && ./bin/kafka-topics.sh --bootstrap-server localhost:9092 --topic __strimzi_store_topic --delete
```
scholzj, line 44: Thsi should use its own pod as in the command I suggested above.

documentation/modules/operators/con-topic-operator-store.adoc
```diff
@@ -0,0 +1,115 @@
...
+----
+
+The Topic Operator will reconstruct the Zookeeper topic metadata from the state of the topics in Kafka.
+
+NOTE: When scaling a Kafka cluster, the topic store topics must also be deleted so that partitions can be reassigned.
```
scholzj, line 49: I don't think this is true. Why would we delete the topic just because of scaling?

documentation/modules/operators/con-topic-operator-store.adoc
```diff
@@ -0,0 +1,115 @@
...
+
+== Scaling and topics
+
+The recommended configuration for topics is to have a topic replication factor of 3, so that there are 2 in-sync replicas on other brokers.
```
scholzj, line 53:
```suggestion
The recommended configuration for the Topic Operator topics is to have a topic replication factor of 3 and minimum of 2 in-sync replicas.
```

documentation/modules/operators/con-topic-operator-store.adoc
```diff
@@ -0,0 +1,115 @@
...
+If a broker is unavailable, at least one other broker is in-sync.
```
scholzj, line 54:
```suggestion
```

documentation/modules/operators/con-topic-operator-store.adoc
```diff
@@ -0,0 +1,115 @@
...
+[source,shell,subs="+quotes,attributes"]
+----
+apiVersion: {KafkaTopicApiVersion}
+kind: KafkaTopic
+metadata:
+  name: my-topic
+  labels:
+    strimzi.io/cluster: my-cluster
+spec:
+  partitions: 1
```
scholzj, line 65: Is 1 partition just an example? Or is it a requirement? Can the TO work with multiple partitions @alesj?

documentation/modules/operators/con-topic-operator-store.adoc
```diff
@@ -0,0 +1,115 @@
...
+  replicas: 3 <1>
+  config:
+    min.insync.replicas=2 <2>
+  #...
+----
+<1> The number of replica topics. Currently, this cannot be changed in the `KafkaTopic` resource, but it can be changed using the `kafka-reassign-partitions.sh` tool.
+<2> The minimum number of replica partitions that a message must be successfully written to, or an exception is raised.
+
+NOTE: In-sync replicas are used in conjunction with the `acks` configuration for a producer application.
+The `acks` configuration determines the number of follower partitions a message must be replicated to before the message is acknowledged as successfully received.
```
scholzj, line 75: This is true. But what does it mean for the user? We should either add how they can configure it for the TO. Or we should just say that the TO always runs with `acks=all`.

documentation/modules/operators/con-topic-operator-store.adoc
```diff
@@ -0,0 +1,115 @@
...
+
+When scaling Kafka clusters by adding or removing brokers, replicas are not reassigned automatically.
```
scholzj, line 77:
```suggestion
When scaling Kafka clusters by adding or removing brokers, replication factor configuration is not changed and replicas are not reassigned automatically.
```

documentation/modules/operators/con-topic-operator-store.adoc
```diff
@@ -0,0 +1,115 @@
...
+However, you can use the `kafka-reassign-partitions.sh` tool to change the replication factor, and manually reassign replicas to brokers.
```
scholzj, line 78: Should we add link to https://strimzi.io/docs/operators/latest/full/using.html#con-partition-reassignment-str here?

documentation/modules/operators/con-topic-operator-store.adoc
```diff
@@ -0,0 +1,115 @@
...
+
+Alternatively, though Cruise Control cannot change the replication factor for topics,
```
scholzj, line 80: I assume Cruise Control can do this, but the Strimzi - CruiseControl integration cannot. Maybe we should make it more clear to not mislead anyone about Cruise Control itself.

scholzj (commented):
Are there an changes needed to https://strimzi.io/docs/operators/latest/full/deploying.html#deploying-the-topic-operator-standalone-str? Or does that remain as it is? Were there no new options which we might want to document?

documentation/modules/managing/proc-cluster-recovery-volume.adoc
PaulRMellor: Changed to:

> `kubectl -n kafka run kafka-admin -ti --image={DockerKafkaImageCurrent} --rm=true --restart=Never -- ./bin/kafka-topics.sh --bootstrap-server localhost:9092 --topic __strimzi-topic-operator-kstreams-topic-store-changelog --delete && ./bin/kafka-topics.sh --bootstrap-server localhost:9092 --topic __strimzi_store_topic --delete`
> 
> The command must correspond to the type of listener and authentication used to access the Kafka cluster.

documentation/modules/operators/con-topic-operator-store.adoc
PaulRMellor: Updated as suggested. I was trying to convey the two aspects: in-memory and backup

documentation/modules/operators/con-topic-operator-store.adoc
PaulRMellor: Thanks. _Cache_ helps here.

documentation/modules/operators/con-topic-operator-store.adoc
PaulRMellor: IMPORTANT: Do not delete these topics, as they are essential to the running of the Topic Operator.

documentation/modules/operators/con-topic-operator-store.adoc
PaulRMellor: @alesj - Ales -- when upgrading is the old store just cleansed or deleted?

documentation/modules/operators/con-topic-operator-store.adoc
PaulRMellor: I've replaced with the kube-admin command, as suggested above

documentation/modules/operators/con-topic-operator-store.adoc
PaulRMellor: @ales - I'm dropping this note

documentation/modules/operators/con-topic-operator-store.adoc
PaulRMellor: Not sure if I'm capturing what you think we need in this update:

> NOTE: In-sync replicas are used in conjunction with the `acks` configuration for a producer application.
> The `acks` configuration determines the number of follower partitions a message must be replicated to before the message is acknowledged as successfully received. The Topic Operator runs with `acks=all` by default, whereby messages must be acknowledged by all in-sync replicas.

documentation/modules/operators/con-topic-operator-store.adoc
PaulRMellor: I've put links to all relevant sections at the end , including this one

documentation/modules/operators/con-topic-operator-store.adoc
PaulRMellor: Updated:

> Alternatively, though the integration of Cruise Control for Strimzi cannot change the replication factor for topics,
> the optimization proposals it generates for rebalancing Kafka include commands that transfer partition replicas and change partition leadership.

PaulRMellor:
> Are there an changes needed to https://strimzi.io/docs/operators/latest/full/deploying.html#deploying-the-topic-operator-standalone-str? Or does that remain as it is? Were there no new options which we might want to document?

Will check with @alesj

documentation/modules/operators/con-topic-operator-store.adoc
scholzj: `runs with acks=all by default` ... I think it is not possible to change it. So I would leave out the `by default` maybe?

PaulRMellor: Okay.

documentation/modules/managing/proc-cluster-recovery-volume.adoc
```diff
@@ -50,9 +50,6 @@ Example output showing columns important to this procedure:
 [source,shell,subs="+quotes,attributes"]
 ----
 NAME                                         RECLAIMPOLICY CLAIM
-pvc-5e9c5c7f-3317-11ea-a650-06e1eadd9a4c ... Retain ...    myproject/data-my-cluster-zookeeper-1
-pvc-5e9cc72d-3317-11ea-97b0-0aef8816c7ea ... Retain ...    myproject/data-my-cluster-zookeeper-0
-pvc-5ead43d1-3317-11ea-97b0-0aef8816c7ea ... Retain ...    myproject/data-my-cluster-zookeeper-2
```
scholzj, line 55: I think this should stay. While the Topic Operator does not use Zookeeper to store the data anymore, it is still used to store Kafka metadata and the volumes need to be recovered!

PaulRMellor: Returned!

documentation/modules/managing/proc-cluster-recovery-volume.adoc
```diff
@@ -193,20 +193,15 @@ kubectl apply -f _kafka.yaml_
...
-. Run an `exec` command to one of the Kafka broker pods to open the ZooKeeper shell script.
+. Delete the internal topic store topics from the Kafka cluster:
 +
-For example, where _my-cluster-kafka-0_ is the name of the broker pod:
-+
-[source,shell,subs="+quotes,attributes"]
+[source,shell]
 ----
-kubectl exec -ti _my-cluster-zookeeper-0_ -- bin/zookeeper-shell.sh localhost:12181
+kubectl -n kafka run kafka-admin -ti --image={DockerKafkaImageCurrent} --rm=true --restart=Never -- ./bin/kafka-topics.sh --bootstrap-server localhost:9092 --topic __strimzi-topic-operator-kstreams-topic-store-changelog --delete && ./bin/kafka-topics.sh --bootstrap-server localhost:9092 --topic __strimzi_store_topic --delete
```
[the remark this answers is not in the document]
PaulRMellor: Removed

documentation/modules/upgrading/proc-downgrade-brokers-older-kafka.adoc
scholzj: To make it future proof you should probably say something like _when downgrading to operator version earlier then 0.22, you should ..._

scholzj (commented):
LGTM apart from the comments waiting for @alesj

documentation/modules/upgrading/proc-downgrade-brokers-older-kafka.adoc
PaulRMellor: Added

documentation/modules/operators/con-topic-operator-store.adoc
```diff
@@ -0,0 +1,117 @@
...
+
+Topic metadata is cached in-memory and accessed locally.
+Updates from operations applied to the local in-memory cache are persisted to a backup topic store on disk.
+The topic store is continually synchronized with updates from Kafka topics or Kubernetes `KafkaTopic` custom resources.
+Operations are handled rapidly with the topic store set up this way,
+but should the in-memory cache crash it is automatically repopulated from the persistent storage.
+
+== Internal topic store topics
+
+Internal topics support the handling of topic metadata.
```
laidan6000, line 20:
```suggestion
Internal topics support the handling of topic metadata stored in the topic store.
```

documentation/modules/operators/con-topic-operator-store.adoc
```diff
@@ -0,0 +1,117 @@
...
+
+`__strimzi_store_topic`:: Input topic for storing the topic metadata
+`__strimzi-topic-operator-kstreams-topic-store-changelog`:: Retains a log of compacted topic store values
+
+IMPORTANT: Do not delete these topics, as they are essential to the running of the Topic Operator.
```
laidan6000, line 25: Maybe use a warning here.

documentation/modules/operators/con-topic-operator-store.adoc
```diff
@@ -0,0 +1,117 @@
...
+
+== Migrating topic metadata from ZooKeeper
+
+In previous releases of Strimzi, topic metadata was stored in ZooKeeper.
```
laidan6000, line 29: This information might be helpful earlier on as well.

documentation/modules/operators/con-topic-operator-store.adoc
laidan6000: Is erased a better verb?

documentation/modules/operators/con-topic-operator-store.adoc
```diff
@@ -0,0 +1,117 @@
...
+When upgrading to Strimzi {ProductVersion}, the transition to Topic Operator control of the topic store is seamless.
+Metadata is found and migrated from ZooKeeper, and the old store is cleansed.
+
+== Downgrading to a Strimzi version that uses ZooKeeper to store topic metadata
+
+If you are reverting back to a version of Strimzi earlier than 0.22, which uses ZooKeeper for the storage of topic metadata,
+you still downgrade your Cluster Operator to the previous version,
+then downgrade Kafka brokers and client applications to the previous Kafka version as standard.
+
+However, you must also delete the topics created for the topic store using a `kafka-admin` command, specifying the bootstrap address of the Kafka cluster.
```
laidan6000, line 41:
```suggestion
However, you must also delete the topics that were created for the topic store using a `kafka-admin` command, specifying the bootstrap address of the Kafka cluster.
```

documentation/modules/operators/con-topic-operator-store.adoc
```diff
@@ -0,0 +1,117 @@
...
+kubectl run kafka-admin -ti --image={DockerKafkaImageCurrent} --rm=true --restart=Never -- ./bin/kafka-topics.sh --bootstrap-server localhost:9092 --topic __strimzi-topic-operator-kstreams-topic-store-changelog --delete && ./bin/kafka-topics.sh --bootstrap-server localhost:9092 --topic __strimzi_store_topic --delete
+----
+
+The command must correspond to the type of listener and authentication used to access the Kafka cluster.
+
+The Topic Operator will reconstruct the ZooKeeper topic metadata from the state of the topics in Kafka.
+
+== Topic Operator topic replication and scaling
+
+The recommended configuration for the Topic Operator topics is to have a topic replication factor of 3, and a minimum of 2 in-sync replicas.
```
laidan6000, line 55:
```suggestion
The recommended configuration for topics managed by the Topic Operator is a topic replication factor of 3, and a minimum of 2 in-sync replicas.
```

documentation/modules/operators/con-topic-operator-store.adoc
```diff
@@ -0,0 +1,117 @@
...
+  labels:
+    strimzi.io/cluster: my-cluster
+spec:
+  partitions: 1
+  replicas: 3 <1>
+  config:
+    min.insync.replicas=2 <2>
+  #...
+----
+<1> The number of replica topics. Currently, this cannot be changed in the `KafkaTopic` resource, but it can be changed using the `kafka-reassign-partitions.sh` tool.
```
laidan6000, line 72:
```suggestion
<1> The number of partition replicas. Currently, this cannot be changed in the `KafkaTopic` resource, but it can be changed using the `kafka-reassign-partitions.sh` tool.
```

documentation/modules/operators/con-topic-operator-store.adoc
```diff
@@ -0,0 +1,117 @@
...
+<2> The minimum number of replica partitions that a message must be successfully written to, or an exception is raised.
+
+NOTE: In-sync replicas are used in conjunction with the `acks` configuration for a producer application.
```
laidan6000, line 75:
```suggestion
NOTE: In-sync replicas are used in conjunction with the `acks` configuration for producer applications.
```

documentation/modules/operators/con-topic-operator-store.adoc
```diff
@@ -0,0 +1,117 @@
...
+
+When scaling Kafka clusters by adding or removing brokers, replication factor configuration is not changed and replicas are not reassigned automatically.
+However, you can use the `kafka-reassign-partitions.sh` tool to change the replication factor, and manually reassign replicas to brokers.
+
+Alternatively, though the integration of Cruise Control for Strimzi cannot change the replication factor for topics,
+the optimization proposals it generates for rebalancing Kafka include commands that transfer partition replicas and change partition leadership.
+
+== Handling changes to topics
+
+A fundamental problem that the Topic Operator has to solve is that there is no single source of truth:
```
laidan6000, line 87:
```suggestion
A fundamental problem that the Topic Operator needs to solve is that there is no single source of truth:
```

documentation/modules/operators/con-topic-operator-store.adoc
```diff
@@ -0,0 +1,117 @@
...
+Both the `KafkaTopic` resource and the Kafka topic can be modified independently of the Topic Operator.
```
laidan6000, line 88:
```suggestion
both the `KafkaTopic` resource and the Kafka topic can be modified independently of the Topic Operator.
```

documentation/modules/operators/con-topic-operator-store.adoc
```diff
@@ -0,0 +1,117 @@
...
+
+For example, suppose the Topic Operator is not running, and a `KafkaTopic` called _my-topic_ is created.
+When the Topic Operator starts, the topic store does not contain information on my-topic, so it can infer that the `KafkaTopic` was created after it was last running.
+The Topic Operator creates the topic corresponding to my-topic, and also stores metadata for my-topic in the topic store.
+
+If you update Kafka topic configuration or apply a change through the `KafkaTopic` custom resource,
+the topic store is updated after the Kafka cluster is reconciled.
+
+The topic store also allows the Topic Operator to manage scenarios where the topic configuration is changed in Kafka topics _and_ updated through Kubernetes `KafkaTopic` custom resources,
+so long as the changes are not incompatible.
```
laidan6000, line 104:
```suggestion
as long as the changes are not incompatible.
```

laidan6000 (changes requested):
Great content. I made a few suggested changes.

documentation/modules/managing/proc-cluster-recovery-volume.adoc
```diff
@@ -17,7 +17,7 @@ The `KafkaTopic` resources must therefore be recovered before the cluster is sta
 
 . Use _Option 2_ when you are unable to recover all `KafkaTopic` resources.
 +
-This time you deploy your cluster without the Topic Operator, delete the Topic Operator data in ZooKeeper, and then redeploy it so that the Topic Operator can recreate the `KafkaTopic` resources from the corresponding topics.
+This time you deploy your cluster without the Topic Operator, delete the Topic Operator topic store metadata, and then redeploy it so that the Topic Operator can recreate the `KafkaTopic` resources from the corresponding topics.
```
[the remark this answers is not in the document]
PaulRMellor: Changed the end bit to:
_then redeploy the Kafka cluster with the Topic Operator so it can recreate the `KafkaTopic` resources from the corresponding topics

documentation/modules/overview/con-overview-components-topic-operator.adoc
```diff
@@ -29,4 +29,6 @@ Working in the other direction, if a topic is:
 This allows you to declare a `KafkaTopic` as part of your application's deployment and the Topic Operator will take care of creating the topic for you.
 Your application just needs to deal with producing or consuming from the necessary topics.
 
-If the topic is reconfigured or reassigned to different Kafka nodes, the `KafkaTopic` will always be up to date.
+The Topic Operator maintains information about each topic in a _topic store_, which is continually synchronized with updates from Kafka topics or Kubernetes `KafkaTopic` custom resources.
+Updates from operations applied to a local in-memory topic store are persisted to a backup topic store on disk.
+If a topic is reconfigured or reassigned to different Kafka nodes, the `KafkaTopic` will always be up to date.
```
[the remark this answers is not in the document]
PaulRMellor: Changed to: _or reassigned to other brokers_

documentation/modules/operators/con-topic-operator-store.adoc
PaulRMellor: If we keep the mention only here, we can drop this section cleanly when it is no longer needed.

laidan6000 (approved):

documentation/modules/operators/con-topic-operator-store.adoc
```diff
@@ -0,0 +1,117 @@
...
+the topic store is updated after the Kafka cluster is reconciled.
+
+The topic store also allows the Topic Operator to manage scenarios where the topic configuration is changed in Kafka topics _and_ updated through Kubernetes `KafkaTopic` custom resources,
+as long as the changes are not incompatible.
+For example, it is possible to make changes to the same topic config key, but to different values.
+For incompatible changes, the Kafka configuration takes priority, and the `KafkaTopic` is updated accordingly.
+
+NOTE: You can also use the `KafkaTopic` resource to delete topics using a `kubectl delete -f _KAFKA-TOPIC-CONFIG-FILE_` command.
+To be able to do this, `delete.topic.enable` must be set to `true` (default) in the `spec.kafka.config` of the Kafka resource.
+
```
laidan6000, line 110: Extra space.

documentation/modules/operators/con-topic-operator-store.adoc
alesj:
```                    log.info("Deleting ZK topics path: {}", topicsPath);
                    zk.delete(topicsPath, -1);
```

alesj: The ZK node holding the topcs data is deleted ...

documentation/modules/operators/con-topic-operator-store.adoc
alesj: Hmmm, good question, since this is the topic controlled by Kafka Streams -- this is the input topic into our topology.
@tombentley ^^ how with the topics controlled by Kafka Streams?

alesj:
> Are there an changes needed to https://strimzi.io/docs/operators/latest/full/deploying.html#deploying-the-topic-operator-standalone-str? Or does that remain as it is? Were there no new options which we might want to document?

> Will check with @alesj

@PaulRMellor what's the question here?

alesj:
> @PaulRMellor what's the question here?

No changes needed to the deploy process afais ...

documentation/modules/operators/con-topic-operator-store.adoc
scholzj: @PaulRMellor So I guess we can just say _deleted_?

documentation/modules/operators/con-topic-operator-store.adoc
alesj: OK, afais, it goes like this:
* you (can) configure number of Streams threads

```   
public static final String NUM_STREAM_THREADS_CONFIG = "num.stream.threads";
private static final String NUM_STREAM_THREADS_DOC = "The number of threads to execute stream processing.";
```

and then those input topic partitions get distributed among all cluster threads / consumers.

tombentley: I can't think of a reason why it wouldn't work with >1 partition, but nor do I see that >1 would be needed.

documentation/modules/operators/con-topic-operator-store.adoc
PaulRMellor: Updated

documentation/modules/operators/con-topic-operator-store.adoc
PaulRMellor: I can add another callout:

> <1> The number of partitions for the topic. Generally, 1 partition is sufficient.

PaulRMellor:
> > @PaulRMellor what's the question here?
> 
> No changes needed to the deploy process afais ...

Thanks @alesj . The question was around the need to document any other `.env` properties in that section describing how to deploy the TO standalone.

scholzj (approved):

tombentley (approved):

PR #4435, opened 2021-02-17
[setup-kubernetes.sh] Add check when KUBE_VERSION is set to latest

im-konge:

### Type of change

- Enhancement

### Description

In case when we set the `KUBE_VERSION` env to `latest` (for example we don't know which is the latest released version etc.) we should grab it from GH releases. This PR adds this possibility.

### DIFF
--- a/.azure/scripts/setup-kubernetes.sh
+++ b/.azure/scripts/setup-kubernetes.sh
@@ -38,6 +38,11 @@ if [ "$TEST_CLUSTER" = "minikube" ]; then
     else
         TEST_MINIKUBE_URL=https://github.com/kubernetes/minikube/releases/download/${TEST_MINIKUBE_VERSION}/minikube-linux-amd64
     fi
+
+    if [ "$KUBE_VERSION" = "latest" ]; then
+        KUBE_VERSION=$(curl -s https://github.com/kubernetes/kubernetes/releases/latest | grep -o "[0-9].[0-9][0-9].[0-9]")
+    fi
+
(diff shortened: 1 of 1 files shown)


.azure/scripts/setup-kubernetes.sh
```diff
@@ -38,6 +38,11 @@ if [ "$TEST_CLUSTER" = "minikube" ]; then
     else
         TEST_MINIKUBE_URL=https://github.com/kubernetes/minikube/releases/download/${TEST_MINIKUBE_VERSION}/minikube-linux-amd64
     fi
+
+    if [ "$KUBE_VERSION" = "latest" ]; then
+        KUBE_VERSION=$(curl -s https://github.com/kubernetes/kubernetes/releases/latest | grep -o "[0-9].[0-9][0-9].[0-9]")
+    fi
+
```
Frawless, line 45: I would prefer to do not set `--kubernetes-version` in that case instead of getting a version from a website which could be different than minikube supports (at least AFAIU from our offline conversation)

Frawless (changes requested):

.azure/scripts/setup-kubernetes.sh
scholzj: That is quite common that the latest Kube release is not supported by Minikube yet.

scholzj (approved):
I'm not sure you should be running anything against _latest_ version ... that is for me just asking for troubles. Code wise it is fine with me.

Frawless (approved):

PR #4438, opened 2021-02-18
Fixed wrong ConfigMap apiVersion

ppatierno:
### Type of change

- Documentation

### Description

Fixed ConfigMap example for metrics with bad apiVersion, using the Strimzi one and not the right Kubernetes v1 one.
It also removes adoc related information into another example.

### Checklist

- [x] Update documentation

### DIFF
--- a/documentation/modules/con-common-configuration-properties.adoc
+++ b/documentation/modules/con-common-configuration-properties.adoc
@@ -354,7 +354,7 @@ When referencing an empty file, all metrics are exposed as long as they have not
 [source,yaml,subs="+attributes"]
 ----
 kind: ConfigMap
-apiVersion: {KafkaApiVersion}
+apiVersion: v1
 metadata:
   name: my-configmap
 data:
@@ -375,7 +375,6 @@ data:
 .Example metrics configuration for Kafka
 [source,yaml,subs="+attributes"]
 ----
-[source,shell,subs="+attributes"]
 apiVersion: {KafkaApiVersion}
 kind: Kafka
 metadata:


scholzj (approved):

documentation/modules/con-common-configuration-properties.adoc
```diff
@@ -354,7 +354,7 @@ When referencing an empty file, all metrics are exposed as long as they have not
 [source,yaml,subs="+attributes"]
 ----
 kind: ConfigMap
-apiVersion: {KafkaApiVersion}
+apiVersion: v1
```
PaulRMellor, line 357: We use a few _v1_ versions in the guides. For secrets, pods, service, volumes and so on. Does it make sense to use an attribute?

PaulRMellor (approved):
Looks good. I left a comment about use of attribute

documentation/modules/con-common-configuration-properties.adoc
ppatierno: It could make sense but because it impacts other files, let's do it in a different PR.

ppatierno: Even if, thinking more, it's not something that is going to change. It's different from Strimzi version where we are in beta1 then beta2, v1, ... With ConfigMap the v1 version is well defined, as for Secrets, and so on. @scholzj wdyt?

scholzj: If you make this attribute, you need to have many attributes ... SecretApiVersion, ConfigMapAPiVersion, ServiceApiVersion etc. as they might change independently. So you would end up with many attributes for something what changes rarely if ever. IMHO not worth it. (the issue here is typo in the example - not something what changed)

ppatierno:
@scholzj I guess we can merge this one?

PR #4445, opened 2021-02-18
Test Strimzi with CRD v1 resources / v1beta2 API only

scholzj:
This is just a test PR to run the tests with the CRD v1 resources / Strimzi v1beta2 API. This is not expected to be merged.


scholzj:
The regression tests are passing. The served its purpose and can be closed.

Issue #1160, opened 2018-12-13
Use Info level k8s `Event` to tell user about phases of Kafka upgrade


tombentley:
I'm trying to document the Kafka upgrade process. The user needs to take certain action (upgrade clients) part way through. Currently the only way for the user to know that it's time to do this is by grepping the logs for a line which looks like this:

```
2018-12-13 12:04:22 INFO  KafkaAssemblyOperator:615 - Reconciliation #53(watch) Kafka(myproject/my-cluster): Kafka version upgrade from 2.0.0 to 2.1.0, phase 2 of 2 completed
```

Coupling the documentation to particular log messages is not ideal because it means that we need to know which log messages have been documented, so we can either avoid changing them or at least update the documentation when the log message is changed.

Using K8s an `Info`-level `Event` would be a better approach.

scholzj:
We have now warning in status when the message-format uses old version. Should that be enough?

scholzj:
Discussed on the Community Call on 25th March => can be closed

PR #4447, opened 2021-02-19
[DOC] Upgrades from older Strimzi versions

laidan6000:
Signed-off-by: Daniel Laing <dlaing@redhat.com>

### Type of change
- Documentation

### Description

Guide(s) updated: _Deploying and Upgrading Strimzi_

This pull request makes further updates to the "Upgrading Strimzi" chapter based the redesigned Kafka upgrade process completed in PR #4314 .

Outstanding to do / discuss
- Add a new procedure for "Upgrading from an older Strimzi version". Older = older than `Strimzi latest - 1`.
- Is adding a new procedure the best approach or should we modify the existing procedures?

### Checklist

- [ ] Write tests
- [X] Update documentation
 - RELATED TO ISSUE **#1160** - Use Info level k8s `Event` to tell user about phases of Kafka upgrade
- [ ] Update CHANGELOG.md

### DIFF
--- a/documentation/assemblies/upgrading/assembly-upgrade-kafka-cluster-operator.adoc
+++ b/documentation/assemblies/upgrading/assembly-upgrade-kafka-cluster-operator.adoc
@@ -7,7 +7,7 @@
 
 The steps to upgrade your Cluster Operator deployment to use Strimzi {ProductVersion} are described in this section.
 
-Follow this procedure if you deployed the Cluster Operator using the installation YAML files.
+Follow this procedure if you deployed the Cluster Operator using the installation YAML files rather than {OperatorHub}.
 
 The availability of Kafka clusters managed by the Cluster Operator is not affected by the upgrade operation.
 
--- a/documentation/assemblies/upgrading/assembly-upgrade-kafka-versions.adoc
+++ b/documentation/assemblies/upgrading/assembly-upgrade-kafka-versions.adoc
@@ -6,7 +6,7 @@
 
 = Upgrading Kafka
 
-After you have upgraded your Cluster Operator, the next step is to upgrade all Kafka brokers to a later supported version of Kafka.
+After you have upgraded your Cluster Operator to {ProductVersion}, the next step is to upgrade all Kafka brokers to the latest supported version of Kafka.
 
 Kafka upgrades are performed by the Cluster Operator through rolling updates of the Kafka brokers. 
 
@@ -29,8 +29,10 @@ The Cluster Operator initiates rolling updates based on the Kafka cluster config
 
 |===
 
-Additional rolling updates occur if the new version of Kafka requires a new ZooKeeper version.
-//check the above statement is still accurate
+As part of the Kafka upgrade, the Cluster Operator initiates rolling updates for ZooKeeper.
+
+* A single rolling update occurs even if the ZooKeeper version is unchanged.
+* Additional rolling updates occur if the new version of Kafka requires a new ZooKeeper version.
 
 .Additional resources
 
--- a/documentation/assemblies/upgrading/assembly-upgrade-kafka-versions.adoc
+++ b/documentation/assemblies/upgrading/assembly-upgrade-kafka-versions.adoc
@@ -6,7 +6,7 @@
 
 = Upgrading Kafka
 
-After you have upgraded your Cluster Operator, the next step is to upgrade all Kafka brokers to a higher supported version of Kafka.
+After you have upgraded your Cluster Operator, the next step is to upgrade all Kafka brokers to a later supported version of Kafka.
 
 Kafka upgrades are performed by the Cluster Operator through rolling updates of the Kafka brokers. 
 
--- a/documentation/assemblies/upgrading/assembly-upgrade-kafka-versions.adoc
+++ b/documentation/assemblies/upgrading/assembly-upgrade-kafka-versions.adoc
@@ -6,18 +6,31 @@
 
 = Upgrading Kafka
 
-After you have upgraded your Cluster Operator, you can upgrade all Kafka brokers to a higher supported version of Kafka.
+After you have upgraded your Cluster Operator, the next step is to upgrade all Kafka brokers to a higher supported version of Kafka.
 
-Kafka upgrades are performed using the Cluster Operator. 
-How the Cluster Operator performs an upgrade depends on the differences between the versions of:
+Kafka upgrades are performed by the Cluster Operator through rolling updates of the Kafka brokers. 
 
-* Inter-broker protocol
-* Log message format
-* ZooKeeper
+The Cluster Operator initiates rolling updates based on the Kafka cluster configuration.
 
-When the versions are the same for the current and target Kafka version, as is typically the case for a patch level upgrade, the Cluster Operator can upgrade through a single rolling update of the Kafka brokers.
+[cols="2*",options="header",stripes="none",separator=¦]
+|===
 
-When one or more of these versions differ, the Cluster Operator requires two or three rolling updates of the Kafka brokers to perform the upgrade.
+¦If `Kafka.spec.kafka.config` contains...
+¦The Cluster Operator initiates...
+
+¦Both the `inter.broker.protocol.version` and the `log.message.format.version`.
+¦A single rolling update.
+
+¦Either the `inter.broker.protocol.version` or the `log.message.format.version`.
+¦Two rolling updates.
+
+¦No configuration for the `inter.broker.protocol.version` or the `log.message.format.version`.
+¦Two rolling updates.
+
+|===
+
+Additional rolling updates occur if the new version of Kafka requires a new ZooKeeper version.
+//check the above statement is still accurate
 
 .Additional resources
 
--- a/documentation/assemblies/upgrading/assembly-upgrade-kafka.adoc
+++ b/documentation/assemblies/upgrading/assembly-upgrade-kafka.adoc
@@ -8,13 +8,13 @@
 Upgrading Strimzi is a three-stage process. 
 To upgrade brokers and clients without downtime, you _must_ complete the upgrade procedures in the following order:
 
-. Update your Cluster Operator to a later Strimzi version.
+. Update your Cluster Operator to Strimzi version {ProductVersion}.
 +
 The approach you take depends on how you xref:cluster-operator-{context}[deployed the Cluster Operator].
 +
 * If you deployed the Cluster Operator using the installation YAML files, perform your upgrade by modifying the Operator installation files, as described in xref:assembly-upgrade-cluster-operator-{context}[Upgrading the Cluster Operator]. 
 +
-* If you deployed the Cluster Operator from {OperatorHub}, use the Operator Lifecycle Manager (OLM) to change the update channel for the Strimzi Operators to the new Strimzi version.
+* If you deployed the Cluster Operator from {OperatorHub}, use the Operator Lifecycle Manager (OLM) to change the update channel for the Strimzi Operators to the Strimzi version {ProductVersion}.
 +
 Depending on your chosen upgrade strategy, after updating the channel, either:
 +
@@ -24,7 +24,7 @@ Depending on your chosen upgrade strategy, after updating the channel, either:
 +
 For more information on using {OperatorHub} to upgrade Operators, see the {OLMOperatorDocs}.
 
-. Upgrade all Kafka brokers and client applications to a later supported Kafka version.
+. Upgrade all Kafka brokers and client applications to the latest supported Kafka version.
 +
 * xref:assembly-upgrading-kafka-versions-{context}[]
 * xref:con-strategies-for-upgrading-clients-{context}[] 
--- a/documentation/assemblies/upgrading/assembly-upgrade-kafka.adoc
+++ b/documentation/assemblies/upgrading/assembly-upgrade-kafka.adoc
@@ -5,9 +5,10 @@
 [id='assembly-upgrade-kafka-{context}']
 = Strimzi and Kafka upgrades
 
-Upgrading Strimzi is a three-stage process. To upgrade brokers and clients without downtime, you _must_ complete the upgrade procedures in the following order:
+Upgrading Strimzi is a three-stage process. 
+To upgrade brokers and clients without downtime, you _must_ complete the upgrade procedures in the following order:
 
-. Update your Cluster Operator to the latest Strimzi version.
+. Update your Cluster Operator to a later Strimzi version.
 +
 The approach you take depends on how you xref:cluster-operator-{context}[deployed the Cluster Operator].
 +
@@ -23,7 +24,7 @@ Depending on your chosen upgrade strategy, after updating the channel, either:
 +
 For more information on using {OperatorHub} to upgrade Operators, see the {OLMOperatorDocs}.
 
-. Upgrade all Kafka brokers and client applications to the latest Kafka version.
+. Upgrade all Kafka brokers and client applications to a later supported Kafka version.
 +
 * xref:assembly-upgrading-kafka-versions-{context}[]
 * xref:con-strategies-for-upgrading-clients-{context}[] 
--- a/documentation/assemblies/upgrading/assembly-upgrade-kafka.adoc
+++ b/documentation/assemblies/upgrading/assembly-upgrade-kafka.adoc
@@ -14,7 +14,7 @@ The approach you take depends on how you xref:cluster-operator-{context}[deploye
 +
 * If you deployed the Cluster Operator using the installation YAML files, perform your upgrade by modifying the Operator installation files, as described in xref:assembly-upgrade-cluster-operator-{context}[Upgrading the Cluster Operator]. 
 +
-* If you deployed the Cluster Operator from {OperatorHub}, use the Operator Lifecycle Manager (OLM) to change the update channel for the Strimzi Operators to the Strimzi version {ProductVersion}.
+* If you deployed the Cluster Operator from {OperatorHub}, use the Operator Lifecycle Manager (OLM) to change the update channel for the Strimzi Operators to a new Strimzi version.
 +
 Depending on your chosen upgrade strategy, after updating the channel, either:
 +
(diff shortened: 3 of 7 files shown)


documentation/assemblies/upgrading/assembly-upgrade.adoc
```diff
@@ -6,10 +6,41 @@
...
+
+.Kafka version support
+
+You can review supported Kafka versions in the link:https://strimzi.io/downloads/[Supported versions^] table.
+
+* The *Operators* column lists all released Strimzi versions (the Strimzi version is often called the "Operator version"). 
+
+* The *Kafka versions* column lists the Kafka versions that are supported by each Strimzi version. 
+
+//When upgrading to an Operator version that drops support for the Kafka version you are upgrading from, the upgraded Cluster Operator returns an error message, for example:
```
laidan6000, line 38: TBC: Decide if this information is better here, or in the Kafka upgrade procedure.

documentation/assemblies/upgrading/assembly-upgrade-kafka-versions.adoc
```diff
@@ -6,18 +6,31 @@
...
+¦Either the `inter.broker.protocol.version` or the `log.message.format.version`.
+¦Two rolling updates.
+
+¦No configuration for the `inter.broker.protocol.version` or the `log.message.format.version`.
+¦Two rolling updates.
+
+|===
+
+Additional rolling updates occur if the new version of Kafka requires a new ZooKeeper version.
+//check the above statement is still accurate
```
laidan6000, line 33: FYI - @scholzj

documentation/modules/upgrading/proc-upgrade-cluster-operator.adoc
```diff
@@ -36,6 +36,16 @@ kubectl apply -f install/cluster-operator
...
++
+The Cluster Operator might return an error message similar to:
++
+[source,shell,subs="+quotes"]
+----
+"Version {KafkaVersionLower} is not supported. Supported versions are: #...".
+----
++
+This error occurs when upgrading to an Operator version that drops support for the previous Kafka version. 
+You will upgrade to a supported Kafka version in the next procedure, when you upgrade all brokers to a later Kafka version.
```
laidan6000, line 49: At the moment, this is presented as "information only"? 
Is that right?
Or, does the user need to update `spec.Kafka.version` now?

documentation/assemblies/upgrading/assembly-upgrade.adoc
```diff
@@ -6,10 +6,42 @@
 [id='assembly-upgrade-{context}']
 = Upgrading Strimzi
 
-Strimzi can be upgraded from a lower version to a higher version to take advantage of new features and enhancements, performance improvements, and security options.
+Strimzi can be upgraded from a given version to a later version to take advantage of new features and enhancements, performance improvements, and security options.
```
scholzj, line 9: I wonder if we should really use here the actual version number ... since as this stands, it applies for Strimzi 0.22.0. It does not apply to 0.21 if upgrading from 0.19. And we do not know how it applies to 0.23 in the future (I hope it will, but 0.23 will have its own docs with possible changes if there are any). Maybe that is why the `later` is so confusing for me.

documentation/assemblies/upgrading/assembly-upgrade.adoc
```diff
@@ -6,10 +6,42 @@
...
+
+Two upgrade paths are possible:
+
+Incremental:: 
+Upgrading Strimzi from the current version to the next version.
++
+For example, upgrading from Strimzi 0.20.1 to 0.21.0.
+
+Multi-version::
+Upgrading Strimzi from the current version to any later version within a single upgrade (skipping one or more intermediate versions).
```
scholzj, line 26: Again, this is documentation for Strimzi 0.22. So this is about upgrading _from older versions to current version_ or possibly _from older versions to Strimzi {ProductVersion}_.

documentation/modules/upgrading/proc-upgrade-cluster-operator.adoc
```diff
@@ -37,14 +37,29 @@ kubectl apply -f install/cluster-operator
 +
 Wait for the rolling updates to complete.
 
+. If the new Operator version no longer supports the Kafka version you are upgrading from, an error message is returned by the Cluster Operator. For example:
++
+[source,shell,subs="+quotes,attributes"]
+----
+"Version {KafkaVersionLower} is not supported. Supported versions are: #...".
```
scholzj, line 44: `{KafkaVersionLower}` is technically still supported so the message might be confusing. We should either add a new variable for the _unssuported_ version or just hardcode here something ... for example `2.4.0`.

scholzj (commented):
I think technically it is good. It is just some of the wording which I commented about.

documentation/modules/upgrading/proc-upgrade-brokers-newer-kafka.adoc
```diff
@@ -6,7 +6,7 @@
 
 = Upgrading Kafka brokers and client applications
 
-This procedure describes how to upgrade a Strimzi Kafka cluster to a higher version of Kafka.
+This procedure describes how to upgrade a Strimzi Kafka cluster to a later version of Kafka.
 
 .Prerequisites
```
laidan6000, line 11: I don't think this prerequisite is still true:

> The Cluster Operator, which supports both versions of Kafka, is up and running.

laidan6000:
@scholzj - I addressed your comments, thank you. Also:

- Changed the ProductVersion attribute to `0.22.0`.

- Updated the terminology to make the content specific to Strimzi 0.22 as far as possible. As Jakub pointed out, there was a mix of version-specific conent and specific content.

- Now using "Strimzi version 0.22.0" (with the ProductVersion attribute). The phrase "new Strimzi version" is sometimes used instead.

- Now using "the latest supported Kafka version" instead of "a later version".

- Removed the `KafkaVersionEarlier` and `KafkaVersionLater` attributes because they add too much complexity.

documentation/modules/upgrading/proc-upgrade-cluster-operator.adoc
laidan6000: I hardcoded the error message for Kafka 2.4.0. That's much simpler than using variables.

documentation/assemblies/upgrading/assembly-upgrade.adoc
```diff
@@ -6,10 +6,40 @@
...
+Each Kafka release introduces new features, improvements, and bug fixes to your Strimzi deployment.
+
+Strimzi can be xref:assembly-downgrade-{context}[downgraded] to the previous version if you encounter issues with the newer version.
+
+.Upgrade paths
+
+Two upgrade paths are possible:
+
+Incremental:: 
+Upgrading Strimzi from version 0.21.1 to version {ProductVersion}.
```
scholzj, line 21: We should write around the 0.21.1 release otherwise we would need to update it all the time. MAybe ...

```suggestion
Upgrading Strimzi from previous version to version {ProductVersion}.
```

laidan6000: OK, I guess it's likely to get missed. I'll change it to "the previous version".

documentation/assemblies/upgrading/assembly-upgrade-kafka-cluster-operator.adoc
```diff
@@ -7,7 +7,7 @@
 
 The steps to upgrade your Cluster Operator deployment to use Strimzi {ProductVersion} are described in this section.
 
-Follow this procedure if you deployed the Cluster Operator using the installation YAML files.
+Follow this procedure if you deployed the Cluster Operator using the installation YAML files rather than {OperatorHub}.
```
tombentley, line 10: What about helm?

documentation/assemblies/upgrading/assembly-upgrade.adoc
tombentley: But does previous version mean 0.20.0 or a 0.20.1? It's a bit ambiguous. Do we mean previous _minor_ version?

documentation/modules/upgrading/proc-upgrade-brokers-newer-kafka.adoc
```diff
@@ -6,19 +6,19 @@
...
 
 For the `Kafka` resource to be upgraded, check:
 
 * The Cluster Operator, which supports both versions of Kafka, is up and running.
-* The `Kafka.spec.kafka.config` does _not_ contain options that are not supported in the version of Kafka that you are upgrading to.
-* Whether the `log.message.format.version` and `inter.broker.protocol.version` for the current Kafka version need to be updated for the new version.
+* The `Kafka.spec.kafka.config` does _not_ contain options that are not supported in the new Kafka version.
+* Whether the `log.message.format.version` and `inter.broker.protocol.version` for the current Kafka version need to be updated for the new Kafka version.
```
tombentley, line 17: I wonder whether we're phrasing this right. It's not like these things necessarily _need_ to be changed. Rather it's up to the user to manage them and thus understand what they do and so decide when it's right for them to change them.

documentation/modules/upgrading/proc-upgrade-brokers-newer-kafka.adoc
```diff
@@ -52,7 +52,9 @@ If the `log.message.format.version` and `inter.broker.protocol.version` are unse
 +
 NOTE: The value of `log.message.format.version` and `inter.broker.protocol.version` must be strings to prevent them from being interpreted as floating point numbers.
 
-. Change the `Kafka.spec.kafka.version` to specify the new version (leaving the `log.message.format.version` and `inter.broker.protocol.version` at the _current_ versions).
+. Change the `Kafka.spec.kafka.version` to specify the new Kafka version.
```
tombentley, line 55: Unless it's already explained elsewhere, I think it would be helpful to explain here that this change is about ensuring that all the brokers in the cluster are using the new binary, and during this change they need to continue using the old IBP so that the brokers can continue communicating.

documentation/modules/upgrading/proc-upgrade-brokers-newer-kafka.adoc
```diff
@@ -83,39 +85,19 @@ See xref:con-versions-and-images-str[Kafka version and image mappings]
...
 . Save and exit the editor, then wait for rolling updates to complete.
 +
-NOTE: Additional rolling updates occur if the new version of Kafka has a new ZooKeeper version.
-+
-Check the update in the logs or by watching the pod state transitions:
-+
-[source,shell,subs=+quotes]
-----
-kubectl logs -f _CLUSTER-OPERATOR-POD-NAME_ | grep -E "Kafka version upgrade from KafkaVersion.*[0-9.]+.* to KafkaVersion.*[0-9.]+.*completed"
-----
```
tombentley, line 93: Glad to see these go :-)

documentation/modules/upgrading/proc-upgrade-brokers-newer-kafka.adoc
```diff
@@ -83,39 +85,19 @@ See xref:con-versions-and-images-str[Kafka version and image mappings]
...
-+
 The rolling updates:
 +
 * Ensure each pod is using the broker binaries for the new version of Kafka
-* Configure the brokers to send messages using the inter-broker protocol of the new version of Kafka
+* Configure the brokers to send messages using the inter-broker protocol version of the new version of Kafka
 +
-NOTE: Clients are still using the old version, so brokers will convert messages to the old version before sending them to the clients. To minimize this additional load, update the clients as quickly as possible.
+IMPORTANT: Clients are still using the old Kafka version, so brokers will convert messages to the old version before sending them to the clients. To minimize this additional load, update all client applications as soon as possible.
```
tombentley, line 100: I think this is misleading. It's not the client version itself that matters. Message conversion can be caused by different message format between producer and topic (which, yes defaults to log.message.format.version, but can be overridden per topic), but also things like mismatched compression. Can we remove this? I think the following step makes it a bit clearer, since the strategies are explained in a bit more detail.

documentation/assemblies/upgrading/assembly-upgrade.adoc
scholzj: I guess that is good point @laidan6000 ^^^ ... _previous minor version_ would be better.

PaulRMellor: Maybe an _(X.Y)_ example in

documentation/modules/upgrading/proc-upgrade-cluster-operator.adoc
```diff
@@ -37,20 +37,35 @@ kubectl apply -f install/cluster-operator
...
+[source,shell,subs="+quotes,attributes"]
+----
+"Version 2.4.0 is not supported. Supported versions are: 2.6.0, 2.6.1, 2.7.0."
+----
++
+If you see this message, upgrade to a Kafka version that is supported by the new Cluster Operator version:
+
+.. Edit the Kafka custom resource.
+
+.. Change the `spec.Kafka.version` property to a supported Kafka version.
```
PaulRMellor, line 51:
```suggestion
.. Change the `spec.kafka.version` property to a supported Kafka version.
```

PaulRMellor (approved):
Great improvement. Clear instructions.

documentation/assemblies/upgrading/assembly-upgrade-kafka.adoc
```diff
@@ -23,7 +24,7 @@ Depending on your chosen upgrade strategy, after updating the channel, either:
 +
 For more information on using {OperatorHub} to upgrade Operators, see the {OLMOperatorDocs}.
 
-. Upgrade all Kafka brokers and client applications to the latest Kafka version.
+. Upgrade all Kafka brokers and client applications to a later supported Kafka version.
```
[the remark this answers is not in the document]
laidan6000: Agreed. I've updated the wording to "latest supported Kafka version" in multiple places.

documentation/assemblies/upgrading/assembly-upgrade-kafka.adoc
```diff
@@ -5,9 +5,10 @@
 [id='assembly-upgrade-kafka-{context}']
 = Strimzi and Kafka upgrades
 
-Upgrading Strimzi is a three-stage process. To upgrade brokers and clients without downtime, you _must_ complete the upgrade procedures in the following order:
+Upgrading Strimzi is a three-stage process. 
+To upgrade brokers and clients without downtime, you _must_ complete the upgrade procedures in the following order:
 
-. Update your Cluster Operator to the latest Strimzi version.
+. Update your Cluster Operator to a later Strimzi version.
```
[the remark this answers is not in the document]
laidan6000: Done. "Update your Cluster Operator to a new Strimzi version."

documentation/assemblies/upgrading/assembly-upgrade.adoc
laidan6000: Agreed. I've changed this to use the actual version number.

> Strimzi can be upgraded to version {ProductVersion} to take advantage of new features and enhancements, performance improvements, and security options.

documentation/assemblies/upgrading/assembly-upgrade.adoc
laidan6000: Changed to:

documentation/assemblies/upgrading/assembly-upgrade-kafka-cluster-operator.adoc
laidan6000: I think we'll need to open a new issue for Helm.

The Deploying and upgrading Strimzi guide should be updated to reflect deployment using Helm charts. The Upgrading Strimzi chapter needs to describe how to upgrade Strimzi if you deployed via Helm.

What do you think, @scholzj ?

documentation/modules/upgrading/proc-upgrade-brokers-newer-kafka.adoc
laidan6000: I'm certainly not keen on the second bullet. How would the user find out which options are _not_ supported in the new Kafka version? By checking the Kafka release notes? Are we just covering our backs or is this an actual concern, I wonder.

As for the third bullet, I can rephrase the sentence and perhaps move it up to the introduction.

documentation/assemblies/upgrading/assembly-upgrade-kafka-cluster-operator.adoc
scholzj: I think that Kafka upgrades are the same regardless of OLM or Helm. For the operator upgrade, I'm not sure we have anyone who knows how to do it with Helm :-/

scholzj (approved):

documentation/assemblies/upgrading/assembly-upgrade.adoc
laidan6000: Changed to "the previous minor version". I tried an X.Y example but it seemed confusing.

documentation/modules/upgrading/proc-upgrade-brokers-newer-kafka.adoc
laidan6000: Line 17 is now part of the procedure introduction:

> Compared to your current Kafka version, the new version might support a higher _log message format version_ or _inter-broker protocol version_, or both. Follow the steps to upgrade these versions, if required. For more information, see Kafka versions.

documentation/modules/upgrading/proc-upgrade-brokers-newer-kafka.adoc
laidan6000: Yes, I think that's useful to know at this stage. I added a note:

> NOTE: Changing the `kafka.version` ensures that all brokers in the cluster will be upgraded to start using the new broker binaries. During this process, some brokers are using the old binaries while others have already upgraded to the new ones. Leaving the `inter.broker.protocol.version` unchanged ensures that the brokers can continue to communicate with each other throughout the upgrade.

documentation/modules/upgrading/proc-upgrade-brokers-newer-kafka.adoc
laidan6000: Removed the note.

laidan6000:
@tombentley - Thanks for the review. The PR is now ready for further checks.

Frawless (approved):

tombentley (approved):

PR #4448, opened 2021-02-19
Add OLM manifest generator scripts

kyguy:
### Type of change

- Enhancement / new feature

### Description

Adds OLM manifest generator scripts to generate the CRD and CSV manifest files that we upload to OperatorHub.

The intended bundle workflow is:
1. Build project (including pushing images to remote registry)
2. Run bundle generator
3. Upload the generated `tools/olm-bundle/bundle/manifest` files to OperatorHub

(Edited to describe new direction of PR )

### Checklist

- [ ] Write tests
- [ ] Update documentation
- [ ] Update CHANGELOG.md


docker-images/bundle/Makefile
```diff
@@ -0,0 +1,70 @@
...
+	$(CP) $(CRD_DIR)/043-Crd-kafkatopic.yaml $(MANIFESTS)/kafkatopics.kafka.strimzi.io.crd.yaml
+	$(CP) $(CRD_DIR)/044-Crd-kafkauser.yaml $(MANIFESTS)/kafkausers.kafka.strimzi.io.crd.yaml
+	$(CP) $(CRD_DIR)/040-Crd-kafka.yaml $(MANIFESTS)/kafkas.kafka.strimzi.io.crd.yaml
+	$(CP) $(CRD_DIR)/041-Crd-kafkaconnect.yaml $(MANIFESTS)/kafkaconnects.kafka.strimzi.io.crd.yaml
+	$(CP) $(CRD_DIR)/042-Crd-kafkaconnects2i.yaml $(MANIFESTS)/kafkaconnects2is.kafka.strimzi.io.crd.yaml
+	$(CP) $(CRD_DIR)/045-Crd-kafkamirrormaker.yaml $(MANIFESTS)/kafkamirrormakers.kafka.strimzi.io.crd.yaml
+	$(CP) $(CRD_DIR)/046-Crd-kafkabridge.yaml $(MANIFESTS)/kafkabridges.kafka.strimzi.io.crd.yaml
+	$(CP) $(CRD_DIR)/047-Crd-kafkaconnector.yaml $(MANIFESTS)/kafkaconnectors.kafka.strimzi.io.crd.yaml
+	$(CP) $(CRD_DIR)/048-Crd-kafkamirrormaker2.yaml $(MANIFESTS)/kafkamirrormaker2s.kafka.strimzi.io.crd.yaml
+	$(CP) $(CRD_DIR)/049-Crd-kafkarebalance.yaml $(MANIFESTS)/kafkarebalances.kafka.strimzi.io.crd.yaml
```
scholzj, line 23: I think the CRDs should either be:
* Copied here by `make crd_install` as is done for other CRD locations
* Or they should be copied at build time and instead of committing them you should just add them to .gitignore (this might be better maybe)

docker-images/bundle/Makefile
```diff
@@ -0,0 +1,70 @@
+PROJECT_NAME=bundle
+
+CRD_DIR := ../../install/cluster-operator
+DEPLOYMENT_FILE := $(CRD_DIR)/060-Deployment-strimzi-cluster-operator.yaml
+MANIFESTS := ./manifests
```
scholzj, line 5: `manifests` seems very generic. Maybe `olm-manifests` would be more corresponding? I'm also not sure what do you need it for. Why not just create some `tmp` file here to keep the files? And maybe the CSV file can live either in `install` or here as well.

docker-images/bundle/Dockerfile
```diff
@@ -0,0 +1,11 @@
+FROM scratch
+
+LABEL operators.operatorframework.io.bundle.package.v1="strimzi-kafka-operator"
+LABEL operators.operatorframework.io.bundle.manifests.v1="manifests/"
+LABEL operators.operatorframework.io.bundle.metadata.v1="metadata/"
+LABEL operators.operatorframework.io.bundle.channel.default.v1="stable"
+LABEL operators.operatorframework.io.bundle.channels.v1="stable, strimzi-0.21.x"
```
scholzj, line 7: Should this be 0.22.0-snapshot or something?

docker-images/bundle/Makefile
```diff
@@ -0,0 +1,70 @@
+PROJECT_NAME=bundle
```
scholzj, line 1: Maybe we can call it `olm-bundle` (and rename the directory as well)? `bundle` is very generic term (for me the YAML installation files are bundle)

docker-images/bundle/Makefile
```diff
@@ -0,0 +1,70 @@
...
+	do \
+		VERSION=`echo "$${val}" | cut -d'=' -f1 | tr -d '.'`;\
+		IMAGE=`echo "$${val}" | cut -d'=' -f2`;\
+		NAME="strimzi-kafka-$${VERSION}"\
+		name=$${NAME} image="$${IMAGE}" yq ea -i '.spec.relatedImages += [{"name": env(name), "value": env(image)}]' $(CSV_FILE); \
+	done\
+	
+	yq ea -i '.spec.relatedImages += [{"name": "strimzi-bridge", "value": "$(KAFKA_BRIDGE)"}]' $(CSV_FILE)
+	yq ea -i '.spec.relatedImages += [{"name": "strimzi-jmxtrans", "value": "$(JMXTRANS_IMAGE)"}]' $(CSV_FILE)
+	yq ea -i '.spec.relatedImages += [{"name": "strimzi-kaniko-executor", "value": "$(KANIKO_EXECUTOR_IMAGE)"}]' $(CSV_FILE)
```
scholzj, line 62: Any chance we can instead get the digest from the just built images? I assume the digest does not change by re-tagging etc., or?

scholzj (commented):
Nice PR ... good work. 

What would be the expected flow? Right now it does not seem to be plugged into the main Makefile. I'm also not sure what would it do at release builds. We would also need to figure out how to deal with OperatorHub which does not use the container image but files in git repository (from which they build their own container I guess).

Some additional notes:
* The hardest part when updating the YAML file is the RBAC which seems to be missing here so far.
* How does this handle the channels?

docker-images/bundle/Dockerfile
tomncooper: Could you read the value from the release.version file in the repo root, might be an argument for having a make target and/or build script for the whole process.

docker-images/bundle/Makefile
tomncooper: I agree, there are so many bundles in the world

docker-images/bundle/metadata/annotations.yaml
```diff
@@ -0,0 +1,8 @@
+annotations:
+  operators.operatorframework.io.bundle.channel.default.v1: stable
+  operators.operatorframework.io.bundle.channels.v1: stable, strimzi-0.21.x
```
tomncooper, line 3: As above we should probably have some logic for deciding the channel from the `release.version` file?

tomncooper:
Great PR @kyguy, this will be very useful. As Jakub said I think this needs to be tied into the top level make file and the release process (which I realize, is easier said than done but the fewer manual steps, the better).

install/Makefile
```diff
@@ -29,7 +38,90 @@ crd_install_helm3:
...
+	$(CP) $(CRD_DIR)/030-ClusterRole-strimzi-kafka-broker.yaml $(MANIFESTS)/strimzikafkabroker.clusterrole.yaml
+	$(CP) $(CRD_DIR)/031-ClusterRole-strimzi-entity-operator.yaml $(MANIFESTS)/strimzientityoperator.clusterrole.yaml
+	$(CP) $(CRD_DIR)/032-ClusterRole-strimzi-topic-operator.yaml $(MANIFESTS)/strimzitopicoperator.clusterrole.yaml
+	$(CP) $(CRD_DIR)/033-ClusterRole-strimzi-kafka-client.yaml $(MANIFESTS)/strimzikafkaclient.clusterrole.yaml
+	$(CP) $(CRD_DIR)/050-ConfigMap-strimzi-cluster-operator.yaml $(MANIFESTS)/strimziclusteroperator.configmap.yaml
+
+	# Update exisiting Cluster Service Version (CSV) file
+	yq ea -i 'select(fi==0).spec.install.spec.deployments[0].spec = select(fi==1).spec | select(fi==0)' $(CSV_FILE) $(DEPLOYMENT_FILE)
+	yq ea -i 'select(fi==0).spec.install.spec.permissions[0].rules = select(fi==1).rules | select(fi==0)' $(CSV_FILE) $(CRD_DIR)/020-ClusterRole-strimzi-cluster-operator-role.yaml
+	yq ea -i 'select(fi==0).spec.install.spec.clusterPermissions = select(fi==1).rules | select(fi==0)' $(CSV_FILE) $(CRD_DIR)/021-ClusterRole-strimzi-cluster-operator-role.yaml
```
tomncooper, line 62: You need to check that this command correctly creates the `spec.install.spec.clusterPermissions.rules` list. When I tried it, it copied the entries below the `- rules:` key but missed the `- rules:` itself.

tomncooper: The `clusterPermissions` was also missing a `serviceAccountName` entry

install/olm-manifests/bundle.clusterserviceversion.yaml
```diff
@@ -0,0 +1,1323 @@
...
+              verbs:
+                - get
+                - list
+                - watch
+                - create
+                - delete
+                - patch
+                - update
+          serviceAccountName: strimzi-cluster-operator
+      clusterPermissions:
```
tomncooper, line 1106: Ok this may explain the errors I saw there should be a `- rules:` below `clusterPermissions:`

tomncooper: And yep this is template is also missing the `serviceAccountName` entry for the `clusterPermissions`.

install/Makefile
tomncooper: Actually this looks like an issue with your bundle template, see comments below.

kyguy: Thanks for the catch, just fixed it up!

kyguy:
Made a couple changes to this PR:
- Updated the PR to only generate the CRD and CSV manifest files that we upload to OperatorHub.
- Added SHA digest pinning
- Added the generated manifest to the `.gitignore` file
- Added the olm-manifest generator script as part of the parent Makefile
- Added a method of generation using the operator-sdk 

**NOTE**: I have generated the manifests from the latest `main` branch and explicitly committed it here so it can be examined (it had to be force added) and will remove them once the PR has been reviewed.

The intended bundle workflow is:
1. Build project
2. Upload the generated manifest files to OperatorHub
3. Update the CSV template  ( e.g. `packaging/olm-manifests/csv-template/bases/strimzi-cluster-operator.clusterserviceversion.yaml`)  with the latest released CSV file in OperatorHub.

Open concerns/questions:
- For the RBAC cluster roles in the CSV, `.spec.install.spec.permissions.rules`, I have noticed we have added the Entity Operator ClusterRoles in past OperatorHub bundle releases. Since the Enitity Operator ClusterRoles are included in the bundle as individual files do we have to list them to the CSV as well?

docker-images/bundle/metadata/annotations.yaml
kyguy: Just updated the direction of the PR and have dropped support for generating the metadata files for now since we maintain these versions in OperatorHub

packaging/olm-manifests/csv-template/bases/strimzi-cluster-operator.clusterserviceversion.yaml
```diff
@@ -0,0 +1,1345 @@
...
+      [
+        {
+          "apiVersion":"kafka.strimzi.io/v1beta2",
+          "kind":"Kafka",
+          "metadata":{
+            "name":"my-cluster"
+          },
+          "spec":{
+            "kafka":{
+                "version":"2.7.0",
```
scholzj, line 48: This should be 2.8.0 now.

packaging/olm-manifests/csv-template/bases/strimzi-cluster-operator.clusterserviceversion.yaml
```diff
@@ -0,0 +1,1345 @@
...
+                    "type": "internal",
+                    "tls": true
+                  }
+                ],
+                "config":{
+                  "offsets.topic.replication.factor":3,
+                  "transaction.state.log.replication.factor":3,
+                  "transaction.state.log.min.isr":2,
+                  "log.message.format.version":"2.7",
+                  "inter.broker.protocol.version":"2.7"
```
scholzj, line 69: These should be 2.8 now.

packaging/olm-manifests/csv-template/bases/strimzi-cluster-operator.clusterserviceversion.yaml
```diff
@@ -0,0 +1,1345 @@
...
+          }
+        },
+        {
+          "apiVersion":"kafka.strimzi.io/v1beta2",
+          "kind":"KafkaConnect",
+          "metadata":{
+            "name":"my-connect-cluster"
+          },
+          "spec":{
+            "version":"2.7.0",
```
scholzj, line 96: 2.8.0 ... there are few more places below ... I will not comment on all of them.

packaging/olm-manifests/csv-template/bases/strimzi-cluster-operator.clusterserviceversion.yaml
```diff
@@ -0,0 +1,1345 @@
...
+    certified: 'false'
+    containerImage: quay.io/strimzi/operator@sha256:0e37f671e6d489c3e56ffb8f51f19a2a014b2cb3f57eafcc5e2931090932e725
+    createdAt: 2021-03-23 13:13:13
+    description: Strimzi provides a way to run an Apache Kafka cluster on Kubernetes
+      or OpenShift in various deployment configurations.
+    repository: https://github.com/strimzi/strimzi-kafka-operator
+    support: Strimzi
+    operators.openshift.io/infrastructure-features: |-
+      ["Disconnected", "Proxy"]
+  name: strimzi-cluster-operator.v0.22.1
```
scholzj, line 356: It is a bit confusing to have this based on a specific version. Would having some placeholder here be better?

packaging/olm-manifests/manifests/strimzi-cluster-operator.clusterserviceversion.yaml
```diff
@@ -0,0 +1,1310 @@
...
+      mediatype: image/svg+xml
+  install:
+    spec:
+      permissions:
```
scholzj, line 879: Are the permissions automatically updated? Or does it have to be maintained manually?

packaging/olm-manifests/manifests/strimzi-cluster-operator.clusterserviceversion.yaml
```diff
@@ -0,0 +1,1310 @@
...
+      url: https://cloud-native.slack.com/messages/strimzi
+    - name: GitHub
+      url: https://github.com/strimzi/strimzi-kafka-operator
+  maintainers:
+    - email: cncf-strimzi-users@lists.cncf.io
+      name: Strimzi
+  maturity: stable
+  provider:
+    name: Strimzi
+  replaces: 0.22.1
```
scholzj, line 1306: How do you know what was the previous version?

Frawless:
Maybe we could create azure olm pipeline (not as part of this PR I guess). AFAIK OLM could be enabled on Minikube and it will be a nice verification of Kyles implementation.

packaging/olm-manifests/generate-manifests.sh
```diff
@@ -0,0 +1,152 @@
...
+  do
+    tag=$(echo $image_tag | cut -d':' -f2);
+    image=$(echo $image_tag | cut -d'@' -f1 | cut -d':' -f1);
+    
+    registry=$(echo $image | cut -d'/' -f1);
+    org=$(echo $image | rev | cut -d'/' -f2 | rev);
+    repo=$(echo $image | rev | cut -d'/' -f1 | rev);
+    
+    echo "Get digest from remote registry for: $image_tag"; 
+    digest="$(curl -s -H "Accept: application/vnd.docker.distribution.manifest.v2+json" https://${registry}/v2/${org}/${repo}/manifests/${tag} | sha256sum | head -c 64)";
```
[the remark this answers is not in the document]
kyguy: It will retrieve the image specified in the yaml files, more specifically the `packaging/install/cluster-operator/060-Deployment-strimzi-cluster-operator.yaml` file. 

The script assumes that the generated deployment yaml has the latest images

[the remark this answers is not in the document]
scholzj: I'm not completely sure this is the case with:
* The images often being pushed into different repositories
* The images being pushed only at the end of the build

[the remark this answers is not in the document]
kyguy: We could restrict the generator script to only be used as part of the Strimzi release workflow after the deployment yaml has been updated to point to the officially supported images of the release. Would this solve the problem?

Of course, the logic can be changed to retrieve the image path information from the parent `Makefile.docker` file but it would be a little more complicated. Would that be preferable?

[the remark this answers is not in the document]
scholzj: Well, I'm not sure what would solve the problem TBH. It depends how and what for is it supposed to be used. Maybe it should be moved to `tools` to make it clear it is not used anywhere during the build / release process but it is just a tool for a manual use?

packaging/olm-manifests/manifests/strimzi-cluster-operator.clusterserviceversion.yaml
kyguy: I just added a curl script to the generator which will set this to the latest released version from OperatorHub

packaging/olm-manifests/manifests/strimzi-cluster-operator.clusterserviceversion.yaml
kyguy: It's only partially automated but I am going to fully automate it as part of this PR. Right now the generator only adds the Cluster Operator permissions but I am working on adding the Entity Operator permissions as well

packaging/olm-manifests/csv-template/bases/strimzi-cluster-operator.clusterserviceversion.yaml
kyguy: Done

packaging/olm-manifests/csv-template/bases/strimzi-cluster-operator.clusterserviceversion.yaml
kyguy: Done

packaging/olm-manifests/csv-template/bases/strimzi-cluster-operator.clusterserviceversion.yaml
kyguy: Done

packaging/olm-manifests/csv-template/bases/strimzi-cluster-operator.clusterserviceversion.yaml
kyguy: Done

Makefile
```diff
@@ -12,7 +12,7 @@ ifneq ($(RELEASE_VERSION),latest)
   GITHUB_VERSION = $(RELEASE_VERSION)
 endif
 
-SUBDIRS=kafka-agent mirror-maker-agent tracing-agent crd-annotations test crd-generator api api-conversion mockkube certificate-manager operator-common config-model config-model-generator cluster-operator topic-operator user-operator kafka-init docker-images packaging/helm-charts/helm3 packaging/install packaging/examples
+SUBDIRS=kafka-agent mirror-maker-agent tracing-agent crd-annotations test crd-generator api api-conversion mockkube certificate-manager operator-common config-model config-model-generator cluster-operator topic-operator user-operator kafka-init docker-images packaging/helm-charts/helm3 packaging/install packaging/examples packaging/olm-manifests
```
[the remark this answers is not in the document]
kyguy: I have moved the generator scripts to the `tools/olm-manifest` directory and taken the references out of the Makefile

packaging/olm-manifests/generate-manifests.sh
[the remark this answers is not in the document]
kyguy: I have moved the generator scripts to the tools/olm-manifest directory for manual execution for now

packaging/olm-manifests/generate-manifests.sh
```diff
@@ -0,0 +1,152 @@
...
+# There are a couple of parts which the operator-sdk does not fully support:
+#   - Copying ClusterRole resources, including: 
+#       a) Operator namespaced Cluster Roles to CSV file from CRDs dir
+#       b) Strimzi Kafka client Cluster Roles to manifests from CRDs dir
+#   - Copying alm-examples field to CSV file from CSV template
+# 
+# Outstanding RFEs to cover these issues:
+#   - https://github.com/operator-framework/operator-sdk/issues/4503
+#
+generate_bundle_with_operator_sdk() {
```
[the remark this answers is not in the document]
kyguy: Yes you are right, I have stripped out the operator-sdk method for now

.gitignore
```diff
@@ -34,6 +34,8 @@ master.html
 helm-charts/helm3/strimzi-kafka-operator-*.*.*.tgz
 packaging/helm-charts/helm3/strimzi-kafka-operator-*.*.*.tgz
 
+packaging/olm-manifests/manifests
```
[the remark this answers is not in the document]
kyguy: Just changed this to `tools/olm-manifests/manifests` for now

packaging/olm-manifests/generate-manifests.sh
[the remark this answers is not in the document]
tomncooper: @kyguy could we not inspect the various `$DOCKER_xxxxxxxx` env vars? You may have to check where in the build chain the defaults are set though?

Makefile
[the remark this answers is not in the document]
tomncooper: I think a separate make target makes sense as this won't need to be done for every build.

packaging/olm-manifests/Makefile
```diff
@@ -0,0 +1,20 @@
...
+
+generate_manifests:
+	./generate-manifests.sh $(RELEASE_VERSION)
+
+docker_build: generate_manifests
+docker_tag:
+docker_push:
+
+all: docker_build
+clean: manfests_clean
```
tomncooper: do you mean `manifests_clean`?

tools/olm-manifests/README.md
```diff
@@ -0,0 +1,11 @@
+# OLM manifests generator
+Bash script for generating OLM manifest files for OperatorHub.
+
+## Requirements
+- bash 5+ (GNU)
+- yq 4.6+ (YAML processor)
+
+## Usage
+```
+./generate-manifests.sh <BUNDLE_VERSION>
```
tomncooper, line 10: Can you add an example bundle version "eg `./generate-manifests.sh 0.22.1`. Also maybe you should fail the script if someone runs it without the bundle version? Unless producing a bundle called `strimzi-cluster-operator-v` is ok?

tomncooper:
Looks good @kyguy! 

As I understand it this will produce OLM bundles in the package manifest format that the OperatorHub is using. But OLM is moving to the bundle format and other OLM based systems will use that format. I am not saying it needs to be in this PR but it would also be nice to use the logic here to give and option to create the bundle format.

Again not needed for this PR, but there is a [python tool](https://olm.operatorframework.io/docs/tasks/validate-package/) maintained by OLM that validates the bundles which might be nice to integrate into this tool to catch issue early.

scholzj:
> Again not needed for this PR, but there is a python tool maintained by OLM that validates the bundles which might be nice to integrate into this tool to catch issue early.

Is that still _supported_? My understanding was that the validation was now moving into the SDK tool.

tomncooper:
> Is that still _supported_? My understanding was that the validation was now moving into the SDK tool.

I was just going off the current OLM docs, if it is in the SDK tool then we could use that.

packaging/olm-manifests/manifests/strimzi-cluster-operator.clusterserviceversion.yaml
kyguy: Fixed

kyguy:
Switching back to using the operator-sdk as it has commands for bundle generation and validation. Although we don't directly need all of the parts that are generated by the operator-sdk for OperatorHub e.g. `metadata/annotations` and `bundle.Dockerfile`, these parts make it easier/possible for us to test and validate the `manifest` files we do need for OperatorHub.

packaging/olm-manifests/generate-manifests.sh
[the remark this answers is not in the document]
kyguy: @tomncooper I am not sure I understand, by `$DOCKER_xxxxxxxx` env vars do you mean the env vars listed in the `packaging/install/cluster-operator/060-Deployment-strimzi-cluster-operator.yaml` file?

tools/olm-bundle/generate-olm-bundle.sh
```diff
@@ -0,0 +1,120 @@
...
+fi
+
+# Generates bundle from existing CRDs
+generate_olm_bundle() {
+  # To generate a bundle from existing CRDs using the operator-sdk we must:
+  #   1.) Specify package name "--package" and bundle version "--version"
+  #   2.) Set `--input-dir` to the directory containing CRDs
+  #   3.) Place our template CSV file in a directory in the kustomize format: 
+  #       "<some-dir>/bases/<package-name>.clusterserviceversion.yaml"
+  # *Note* the package name must match the name used in the CSV template
```
tomncooper, line 35: What will happen if it doesn't? If it would result in subtle bugs could we do a check here as we know `$PACKAGE_NAME` and can use yq to inspect the CSV?

packaging/olm-manifests/generate-manifests.sh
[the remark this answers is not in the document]
tomncooper: No, in the docker build script we use `DOCKER_REGISTRY`, `DOCKER_ORG` and `DOCKER_TAG` env vars to control how images are named, tagged and what repo they are pushed too. If these are set for a custom build then you can construct the container image address from them and pull the image that was just built.

scholzj (commented):
I'm confused ... will the CRDs be committed here or not? I think they should be not if the generated CSV is not. They should copied on the fly from the `packaging/...` directories. Same for the roles and other resources which are just copies I guess?

kyguy:
@scholzj Sorry for the confusion, I was not planning on including the copied CRDs or anything else copied/generated by the generator. I have temporarily committed the copied/generated files for review but am planning on taking them down after review and before the PR is merged.

scholzj:
@kyguy Would it be maybe easier to just leave exactly what do you plan to commit? Everyone can check it out and try it or maybe you can upload a ZIP file with example output. It would make the review IMHO more straight forward when knowing what exactly is being reviewed.

kyguy:
@scholzj Yes, I will do that

tools/olm-bundle/README.md
```diff
@@ -0,0 +1,17 @@
...
+Bash script for generating OLM bundle for OperatorHub.
+
+## Requirements
+- bash 5+ (GNU)
+- yq 4.6+ (YAML processor)
+- operator-sdk 1.19+
+
+## Usage
+```
+./generate-olm-bundle.sh <BUNDLE_VERSION>
```
scholzj, line 11: What is bundle version? Is that the same as operator version?

tools/olm-bundle/generate-olm-bundle.sh
```diff
@@ -0,0 +1,120 @@
...
+    
+    registry=$(echo $image | cut -d'/' -f1);
+    org=$(echo $image | rev | cut -d'/' -f2 | rev);
+    repo=$(echo $image | rev | cut -d'/' -f1 | rev);
+    
+    echo "Get digest from remote registry for: $image_tag"; 
+    digest="$(curl -s -H "Accept: application/vnd.docker.distribution.manifest.v2+json" https://${registry}/v2/${org}/${repo}/manifests/${tag} | sha256sum | head -c 64)";
+    
+    image_digest="${image}@sha256:${digest}"
+    sed -i "s|${image_tag}|${image_digest}|g" ${CSV_FILE};
```
scholzj, line 115: You should probably load the multiplatform support and use `$SED` and `$CP` vars to make sure this works on MacOS.

see-quick: https://gist.github.com/see-quick/57aa46cfb46e5598dbec83f7819299d6 something like this should be great

scholzj: You mean like this? https://github.com/strimzi/strimzi-kafka-operator/blob/main/tools/multi-platform-support.sh :-o

tools/olm-bundle/README.md
kyguy: TLDR; The bundle version is the version of the operator metadata package used by OLM. While OperatorHub uses the `package manifest` format[1], the `<BUNDLE_VERSION>` will be the same as the operator version. However, once OperatorHub moves to the `bundle` format, the `<BUNDLE_VERSION>` will be different from the operator version.

OperatorHub currently uses the `package manifest` format to release operator metadata packages, CRDs + CSV files, for OLM installation. The `package manifest` format does not require strict versioning of operator metadata packages so we can make updates to operator metadata of an operator version without versioning the operator metadata package. So while OperatorHub uses the `package manifest` format, we can keep this <BUNDLE_VERSION> as the same as the Operator version.

OLM has been moving towards using a new metadata package format called the `bundle` format, which is what OperatorHub will use in the future. The `bundle` format requires strict versioning of operator metadata packages, so whenever we update the operator metadata of an operator version, we must bump the operator metadata package version. This will force us to decouple the operator version from the operator metadata package version. For example, in the future we may have an operator of version `0.23.0` but then have an operator metadata package of version `0.23.0+0`

[1] https://docs.openshift.com/container-platform/4.7/operators/understanding/olm-packaging-format.html#olm-package-manifest-format_olm-packaging-format

scholzj: That is fine and I understand that. But I do not think this is clear from the readme. Especially the example with version `2.3.0` is confusing. Because in reality, even when the versioning is not exactly the same, it will still follow the operator versioning, right? It will not be _random-number+N_ but _operator-version+N_ for example. So having the example more reaslistic and setting it for example to 0.22.1 instead of 2.3.0 might make it more clear.

tools/olm-bundle/README.md
```diff
@@ -0,0 +1,17 @@
+# OLM bundle generator
```
scholzj, line 1: I guess the README should describe things which need to be updated manually? Following come to my mind:
* Description inside the CSV
* `permissions` and `clusterPermissions` inside the CSV (or is this in some way generated)?
* Examples in the CSV annotation?

scholzj (commented):
Few more nits for the README. I'm also going to try it out - so I might have more comments later.

tools/olm-bundle/generate-olm-bundle.sh
```diff
@@ -0,0 +1,141 @@
+#!/usr/bin/env bash
+#
+# Generates OLM bundle using existing CRDs
+#
+#
+source ../multi-platform-support.sh
```
scholzj, line 6: I think this should look like this so that it can be run from different directories (e.g. `./tools/olm-bundle/generate-olm-bundle.sh ...`)

```suggestion
source "$(dirname $(realpath $0))/../multi-platform-support.sh"
```

tools/olm-bundle/generate-olm-bundle.sh
```diff
@@ -0,0 +1,141 @@
...
+
+PACKAGE_NAME=strimzi-cluster-operator
+VERSION=$1
+
+CHANNELS=latest
+PREVIOUS_BUNDLE_VERSION=$(curl -s https://raw.githubusercontent.com/operator-framework/community-operators/master/community-operators/strimzi-kafka-operator/strimzi-kafka-operator.package.yaml | yq e '.channels[0].currentCSV' -)
+
+CSV_TEMPLATE_DIR=./csv-template
```
scholzj, line 14: Actually, I think you need to use `$(dirname $(realpath $0))/` for all the paths.

tools/olm-bundle/README.md
```diff
@@ -0,0 +1,17 @@
...
+Bash script for generating OLM bundle for OperatorHub.
+
+## Requirements
+- bash 5+ (GNU)
+- yq 4.6+ (YAML processor)
+- operator-sdk 1.19+
```
scholzj, line 7: Hey, there is no operator SDK version 1.19 ... the newest is 1.7.1

scholzj:
Any chance we can keep the original names which we use in Strimzi OperatorHub today?

kyguy:
> Any chance we can keep the original names which we use in Strimzi OperatorHub today?

Unfortunately, the operator-sdk generates its own names for these files when it copies them over so I have added some scripts to change them back to the original names we have in OperatorHub.

tools/olm-bundle/README.md
kyguy: Just added a section in the README that describes what is generated automatically and what needs to be updated manually

tools/olm-bundle/README.md
kyguy: Yes you are right, sorry, I meant to put `0.23.0` there but accidentally put `2.3.0`. I have just updated the example to `0.22.1`.

tools/olm-bundle/README.md
```diff
@@ -0,0 +1,32 @@
...
+Then the generator creates a CSV file generating the following CSV fields automatically:
+- `metadata.name`
+- `metadata.annotations.containerImage`
+- `metadata.annotations.createdAt`
+- `spec.name`
+- `spec.version`
+- `spec.replaces`
+- `spec.install.spec.permissions`
+- `spec.install.spec.clusterPermissions`
+- `spec.install.spec.deployments`
```
scholzj, line 14: How does this work? I guess you do it with `yq`?

scholzj (commented):
* I seem to be getting following error when I run it now:
  ```
  INFO[0000] Unpacking image layers
  INFO[0000] running /usr/local/bin/docker pull ./bundle/
  ERRO[0000] invalid reference format
  FATA[0000] error unpacking image ./bundle/: error pulling image: invalid reference format
  . exit status 1
  ```
  That seemed to have screwed up the replacing of the container image names with digests.
* If you rebase and run it, it also looks like the new `STRIMZI_FEATURE_GATES` env var is wrong (missing value).
* The deployment keeps the resources. At least in the past, that used to break the OLM CIs. So we do not set the resources in OLM.
* We also name the deployment per version. E.g. `strimzi-cluster-operator-v0.22.1` and set Recreate strategy because of some past issues.
* The permissions are a bit weird since they have duplicate records. Will it work with that? Did you actually tested the OLM deployment with it?
  ```
              - apiGroups:
                - ""
              resources:
                - nodes
              verbs:
                - get
            - apiGroups:
                - ""
              resources:
                - nodes
              verbs:
                - get
  ```

It might be good to try to diff the output against the last submission which we have.

kyguy:
> I seem to be getting following error when I run it now:

I have just fixed this so you can now run the generator from any directory.

> If you rebase and run it, it also looks like the new STRIMZI_FEATURE_GATES env var is wrong (missing value).

I have rebased to pick up this change and fixed the missing value (operator-sdk removes fields with empty string values)

> The deployment keeps the resources. At least in the past, that used to break the OLM CIs. So we do not set the resources in OLM.

I have added some logic to remove the deployment's resource section after copying it

> We also name the deployment per version. E.g. strimzi-cluster-operator-v0.22.1 and set Recreate strategy because of some past issues.

I have added the version to the deployment name

> The permissions are a bit weird since they have duplicate records. Will it work with that? Did you actually tested the OLM deployment with it? 

Yes, I have tested this and it and it works but I agree that it looks weird. I have removed the duplicated values. Generating the RBAC rules in the manner that we have done in OperatorHub (where similar rules of different ClusterRoles are merged together) is non-trivial with bash + yq scripts. The best thing we can do is aggregate the rules from the different ClusterRoles and remove duplicates

tools/olm-bundle/README.md
kyguy: Yes, these are copied over from the existing CRD/install files using `yq`

Do you think I should add more information about how it is done to the README?

packaging/olm-manifests/generate-manifests.sh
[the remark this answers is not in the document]
kyguy: Thanks for the clarification Tom! I have just added a Makefile in the `tools/olm-bundle` directory which will pulls in the env vars from the parent `Makefile.docker` file. Now we can execute the generator by running:
```
make BUNDLE_VERSION=0.23.0
```
and it will use the env vars defined in the Makefile.docker file: `DOCKER_REGISTRY`, `DOCKER_ORG`, and `DOCKER_TAG` to populate the image fields in the CSV file.

kyguy:
This is ready for another round of reviews @scholzj @tomncooper , note that I updated how the tool is executed:
```
make BUNDLE_VERSION=0.23.0
```
from the `tools/olm-bundle` dir. Still thinking if it is worth linking this Makefile into the project's parent Makefile so we could do something like:
```
make olm_install BUNDLE_VERSION=0.23.0
```
from the project's parent directory. Anyays, let me know what you think!

tools/olm-bundle/Makefile
```diff
@@ -0,0 +1,12 @@
+PROJECT_NAME=olm-bundle
+
+BUNDLE_VERSION ?= $(shell cat ../../release.version | cut -d'-' -f1)
```
tomncooper, line 3: For robustness you should probably use:
```
$(dirname $(realpath $0))/../../release.version
```

tomncooper:
I am trying to run the make file from the `tools/olm-bundle` directory and I am getting the following error:
```
Get digest from remote registry for image: docker.io/tom/null:latest
ERROR: Could not find image in remote registry: 'docker.io/tom/null:latest'
make: *** [Makefile:10: olm_install] Error 1
```
Now that error is expected as that image indeed doesn't exist. But the issue is the `null` in the image path. I think this is because `$repo` is not being set/calculated properly in the ` generate_image_digests()` function?

kyguy:
> But the issue is the null in the image path. I think this is because $repo is not being set/calculated properly in the generate_image_digests() function?

The generator reuses the repo section of the image paths specified in `packaging/install/cluster-operator/060-Deployment-strimzi-cluster-operator.yaml` file. I am guessing that either one of those images paths were set to `null` or there is some logic error due to some image path format I did not anticipate. Regardless, I will add some logic to defend against this and ping you to try again.

If you still have the partially generated CSV file, send it my way for a post mortem!

scholzj (commented):
Mostly good ... but one bug which I found ... in the generated CSV, in the deployment, the operator image is as `quay.io/strimzi/operator:latest` instead of using the digest. Also the related images seem to be screwed up:

```
  relatedImages:
    - name: strimzi-
      image: ""
    - name: strimzi-
      image: ""
    - name: strimzi-
      image: ""
    - name: strimzi-kafka-bridge
      image: quay.io/strimzi/kafka-bridge@sha256:bed07186355a4c04ee50e90048e3bd3a023264093a9ef28f0278cce44640e93f
    - name: strimzi-jmxtrans
      image: quay.io/strimzi/jmxtrans@sha256:3b447d338350569f2056457b1152a8dcb9956bf0883e3a2e5692c0f96171c7a2
    - name: strimzi-kaniko-executor
      image: quay.io/strimzi/kaniko-executor@sha256:b0ca0b4fb16822e55ea1250c54c73c7c4248282908065422ededa49fa400436a
    - name: strimzi-kafka-270
      image: quay.io/strimzi/kafka@sha256:ad3e0a30303a58cc687212b475f48ec53a39248a55df8543cc759682d0fc088b
    - name: strimzi-kafka-271
      image: quay.io/strimzi/kafka@sha256:47da9306caa53e0cdcc50fd12f5d943998fb0ce306f416028eaadfd73fd49ae4
    - name: strimzi-kafka-280
      image: quay.io/strimzi/kafka@sha256:ea9cc37e995cba2688a2a7dba216d09f3eed1cfe802dc109dacf4feeadc0e26e
```

tools/olm-bundle/generate-olm-bundle.sh
```diff
@@ -0,0 +1,227 @@
...
+ 
+  # Move Dockerfile generated by operator-sdk
+  mv ./bundle.Dockerfile $DOCKERFILE
+  
+  # Remove last three lines of Dockerfile with incorrectly generated paths
+  head -n -3 $DOCKERFILE | tee $DOCKERFILE 1> /dev/null 
+  
+  # Add copy commands with correct paths
+  echo "COPY ./manifests/ /manifests/" >> $DOCKERFILE
+  echo "COPY ./metadata/ /metadata/" >> $DOCKERFILE
```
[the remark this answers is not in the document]
kyguy: Thanks for spotting, this is an error due to a platform specific (GNU specific) argument used by the `head` command. I have updated the `multi-platform-support.sh` to enforce the use of GNU version of `head` when running on OSX. This should fix the issue!

tools/olm-bundle/Makefile
```diff
@@ -0,0 +1,11 @@
...
+
+BUNDLE_VERSION ?= $(shell cat $(dirname $(realpath $0))../../release.version | cut -d'-' -f1)
+.DEFAULT_GOAL := default
+
+include ../../Makefile.docker
+
+olm_install:
+	./generate-olm-bundle.sh BUNDLE_VERSION=$(BUNDLE_VERSION) DOCKER_REGISTRY=$(DOCKER_REGISTRY) DOCKER_ORG=$(DOCKER_ORG) DOCKER_TAG=$(DOCKER_TAG)
+
+default: olm_install
```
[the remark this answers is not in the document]
kyguy: The intented benefit of running the `make` command is to use the Docker env vars defined in the `../../Makefile.docker` file: `DOCKER_REGISTRY`, `DOCKER_ORG`, and `DOCKER_TAG` so users can create bundles based on the container images they have just built.

I have just plugged it into the main Makefile now so that users can run:
```
make olm_install
```
to generate their bundle files

kyguy:
> Also the related images seem to be screwed up:

The most recent commits should fix the issue surrounding the digest pinning of the operator image and the blank value for the operator in the `relatedImages` section. I am not sure yet why there were duplicated empty value entries in the `relatedImages` section, I am still investigating potential causes

tools/multi-platform-support.sh
```diff
@@ -24,4 +25,5 @@ then
     WC=gwc
     UNIQ=guniq
     SORT=gsort
+		HEAD=ghead
```
scholzj, line 28: This seems to have a weird alignment.

scholzj:
>The most recent commits should fix the issue surrounding the digest pinning of the operator image and the blank value for the operator in the relatedImages section. I am not sure yet why there were duplicated empty value entries in the relatedImages section, I am still investigating potential causes

That seems to be fixed now.

> Thanks for spotting, this is an error due to a platform specific (GNU specific) argument used by the head command. I have updated the multi-platform-support.sh to enforce the use of GNU version of head when running on OSX. This should fix the issue!

This did not seemed to help. Maybe you also need to use `gtee`?

scholzj:
Also, the latest change around the Makefile seems to break the builds:

```
make[1]: Nothing to be done for `all'.
/Library/Developer/CommandLineTools/usr/bin/make -C tools/olm-bundle all
# Build Docker image ...
docker  build  --build-arg strimzi_version=0.24.0-SNAPSHOT -t strimzi/olm-bundle:latest ./
[+] Building 0.1s (1/2)
 => [internal] load build definition from Dockerfile                                                                                                                                                                                        0.0s
 => => transferring dockerfile: 2B                                                                                                                                                                                                          0.0s
failed to solve with frontend dockerfile.v0: failed to read dockerfile: open /var/lib/docker/tmp/buildkit-mount181801806/Dockerfile: no such file or directory
make[1]: *** [docker_build_default] Error 1
make: *** [tools/olm-bundle] Error 2
```

I'm not sure adding it to `SUBDIRS` is a good idea.

kyguy:
> This did not seemed to help. Maybe you also need to use gtee?

Just added `gtee` now, if that doesn't work I have a sed trick that should work

> I'm not sure adding it to SUBDIRS is a good idea.

I was having trouble getting the `make olm_install` working without adding it to the subdir section of the main Makefile. I'll try for a better way around it

scholzj:
> I was having trouble getting the make olm_install working without adding it to the subdir section of the main Makefile. I'll try for a better way around it

Then maybe keep the make file inside the olm-generator but don't integrate it? Otherwise you would probably need to make sure it works in the CIs which might need more Makefile work but also installing the OperatorSDK tooling etc.

scholzj:
BTW: Dockerfile looks good now.

kyguy:
> Then maybe keep the make file inside the olm-generator but don't integrate it? Otherwise you would probably need to make sure it works in the CIs which might need more Makefile work but also installing the OperatorSDK tooling etc.

Yeah you are right, it's probably best to streamline the tooling before we integrate it into the CI too. I will remove the integration from the main Makefile

scholzj (approved):

kyguy:
> Now that error is expected as that image indeed doesn't exist. But the issue is the null in the image path. I think this is because $repo is not being set/calculated properly in the generate_image_digests() function?

The recent changes should prevent this from happening unless the images paths are set to "null" explicitly in the `packaging/install/cluster-operator/060-Deployment-strimzi-cluster-operator.yaml` file.

All open comments have been addressed and unless there are any more concerns, this PR should be good to go!

samuel-hawker:
I wonder if the fact some of the variables aren't being set might suggest that the `shellcheck` scripts aren't properly linting this script, has `tools/olm-bundle` been added to the shellchecker script? https://github.com/strimzi/strimzi-kafka-operator/blob/f2ce3ea17c28846ad61640df55f46231a4cea17f/.azure/scripts/shellcheck.sh#L7

scholzj:
@kyguy Can you have a look at the comment from @samuel-hawker? I do not think we need this for 0.24 - it is more or less not tied to any release ... so whenever you have time.

kyguy:
@samuel-hawker I have correct all the warnings raised by the shellcheck linting and added the generator script to the shellcheck linter script

samuel-hawker:
@kyguy all the changes to pass shellcheck look good to me!
This PR makes sense to me, though I would be more comfortable approving once I know @tomncooper has confirmed that he's happy with this PR as well, since I saw he had some earlier concerns/comments (which I believe have all been addressed).

tools/olm-bundle/generate-olm-bundle.sh
```diff
@@ -0,0 +1,230 @@
...
+    repo=$(echo "$image" | rev | cut -d'/' -f1 | rev);
+    
+    if [ "$repo" != "kafka-bridge" ]; then
+      tag=$(echo "$tag" | $SED "s/$(echo "$tag" | cut -d'-' -f1)/${DOCKER_TAG}/g")
+    fi
+    registry=$DOCKER_REGISTRY;
+    org=$DOCKER_ORG;
+
+    echo "Get digest from remote registry for image: ${registry}/${org}/${repo}:${tag}"
+    resp="$(curl -s -w '\n%{http_code}' -H "Accept: application/vnd.docker.distribution.manifest.v2+json" "https://${registry}/v2/${org}/${repo}/manifests/${tag}")"
```
tomncooper, line 211: When I set registry to `docker.io` (with my own repo) this line results in a `301` return code from `curl` suggesting the URL format has changed. When I use `quay.io` I get a `401` unauthorized return code. 

Is there a way to get these digests by using the `docker` or `podman` CLI's that would result in a less brittle implementation.

tomncooper: For example I can run:
```
$ docker image inspect docker.io/tomncooper/operator:latest | jq -r ".[0].RepoDigests"  
[
  "tomncooper/operator@sha256:906a7e45f653957d13ff9c2727c2db98c06dab385f811ab686743fbcd1e2944a"
]
```

tomncooper: This works with `podman` too. Of course the issue with this is you need to pull the images first.

tomncooper:
Great work plugging away at this Kyle.

I pulled the PR branch. When I run the build with custom docker repo I get the following:
```
make olm_install 
./generate-olm-bundle.sh BUNDLE_VERSION=0.24.0 DOCKER_REGISTRY=docker.io DOCKER_ORG=tomncooper DOCKER_TAG=latest
Generating bundle version 0.24.0
Generating bundle manifests
WARN[0000] ClusterServiceVersion validation: [OperationFailed] provided API should have an example annotation 
...
Bundle manifests generated successfully in /home/tom/repos/red_hat/strimzi-kafka-operator/tools/olm-bundle/bundle
Generating bundle metadata
INFO[0000] Creating /home/tom/repos/red_hat/strimzi-kafka-operator/tools/olm-bundle/bundle/metadata/annotations.yaml 
INFO[0000] Creating bundle.Dockerfile                   
INFO[0000] Bundle metadata generated suceessfully       
rm: cannot remove '/home/tom/repos/red_hat/strimzi-kafka-operator/tools/olm-bundle/bundle/manifests/strimzi-cluster-operator-namespaced_rbac.authorization.k8s.io_v1_clusterrole.yaml': No such file or directory
Update CRD filename kafka.strimzi.io_kafkabridges.yaml -> kafkabridges.kafka.strimzi.io.crd.yaml
Update CRD filename kafka.strimzi.io_kafkaconnectors.yaml -> kafkaconnectors.kafka.strimzi.io.crd.yaml
Update CRD filename kafka.strimzi.io_kafkaconnects2is.yaml -> kafkaconnects2is.kafka.strimzi.io.crd.yaml
Update CRD filename kafka.strimzi.io_kafkaconnects.yaml -> kafkaconnects.kafka.strimzi.io.crd.yaml
Update CRD filename kafka.strimzi.io_kafkamirrormaker2s.yaml -> kafkamirrormaker2s.kafka.strimzi.io.crd.yaml
Update CRD filename kafka.strimzi.io_kafkamirrormakers.yaml -> kafkamirrormakers.kafka.strimzi.io.crd.yaml
Update CRD filename kafka.strimzi.io_kafkarebalances.yaml -> kafkarebalances.kafka.strimzi.io.crd.yaml
Update CRD filename kafka.strimzi.io_kafkas.yaml -> kafkas.kafka.strimzi.io.crd.yaml
Update CRD filename kafka.strimzi.io_kafkatopics.yaml -> kafkatopics.kafka.strimzi.io.crd.yaml
Update CRD filename kafka.strimzi.io_kafkausers.yaml -> kafkausers.kafka.strimzi.io.crd.yaml
Update CRD filename strimzi-cluster-operator_v1_configmap.yaml -> strimziclusteroperator.configmap.yaml
Get digest from remote registry for image: docker.io/tomncooper/operator:latest
ERROR: Could not find image in remote registry: 'docker.io/tomncooper/operator:latest'
make: *** [Makefile:9: olm_install] Error 1
```
So I think you need to fix image digest pull issue. 

Also are those warnings from the olm validation an issue with our community bundle as well? Either way we should probably address them. 

Finally, I am not sure why: `rm: cannot remove '/home/tom/repos/red_hat/strimzi-kafka-operator/tools/olm-bundle/bundle/manifests/strimzi-cluster-operator-namespaced_rbac.authorization.k8s.io_v1_clusterrole.yaml': No such file or directory` is happening?

tomncooper (changes requested):
You need to have a look at he digest image fetching and a couple of warning/errors when running the make command.

kyguy:
Thanks for the review @tomncooper! TL;DR the issues should now be addressed with the latest commit

> When I set registry to docker.io (with my own repo) this line results in a 301 return code from curl suggesting the URL format has changed. When I use quay.io I get a 401 unauthorized return code.

To address the authorization/digest pinning issues, I have replaced `curl` with `skopeo` [1]. It is another depedency that will have to be installed separately in order to use the bundle generator tool but I think it is worth using it as it

- Avoids messy curl/bash script
- Handles credentials to private container image registries/repos
- Can retrieve digests without pulling the container images

> Also are those warnings from the olm validation an issue with our community bundle as well? Either way we should probably address them.

Yes those warnings are flagged in our community bundle as well but they can safely be ignored for now. There are two issues causing the warnings here:

1. A bug in the operator-sdk generator that prevents it from seeing/handling the `metadata.annotations.alm-examples` in the `csv-template/bases/strimzi-cluster-operator.clusterserviceversion.yaml` file [2] (I need to chase this issue up upstream)
2. The lack of examples provided by us in the `metadata.annotations.alm-examples` section of our CSV file for CRDs using older Strimzi APIs like `v1alpha1` and `v1beta1` (i.e. `KafkaTopic` and `KafkaUser` resources). Note that we already provide examples for `v1beta2` versions of these CRDs in the CSV's `metadata.annotations.alm-examples` section but the operator-sdk complains about not having a example for every listed API version of a CRD.

> Finally, I am not sure why: rm: cannot remove '/home/tom/repos/red_hat/strimzi-kafka-operator/tools/olm-bundle/bundle/manifests/strimzi-cluster-operator-namespaced_rbac.authorization.k8s.io_v1_clusterrole.yaml': No such file or directory is happening?

I have removed the offending script, it was a work-around for an operator-sdk bug but it looks like the bug got fixed in the latest version. Since the workaround is now unecessary, I have removed it

I also rebased the PR on the `main` branch and removed stale CRD/CRD section i.e KafkaConnectS2I

Let me know what you think!

[1] https://github.com/containers/skopeo
[2] https://github.com/operator-framework/operator-sdk/issues/4503

scholzj:
> To address the authorization/digest pinning issues, I have replaced curl with skopeo [1]. It is another depedency that will have to be installed separately in order to use the bundle generator tool but I think it is worth using it as it

Did anyone investigated why does Tom have the HTTP 401 issue? I did not seemed to have it. Also, how does `skopeo` work on MacOS / Windows? Wouldn't pulling the images and getting the digest that way be better? How do the different options handle multi-arch images?

kyguy:
> Did anyone investigated why does Tom have the HTTP 401 issue?

What I found was that DockerHub does not allow anonymous requests of public container images anymore, so curl failed when requesting the image metadata. 

The curl request to Quay failed because one or more of the images being queried was private. 

Alternatively, I could have created registry tokens for curl to use to get this private metadata but it makes the script a little more complicated.

> Also, how does skopeo work on MacOS / Windows

Skopeo should work and be installable on MacOS, I need to investigate about Windows though.

> Wouldn't pulling the images and getting the digest that way be better? 

I would like to avoid pulling all of the images just to retrieve their digests, unfortunately, docker and podman do not offer this. I can look deeper into using auth tokens with curl if skopeo is undesirable.

> How do the different options handle multi-arch images?

Assuming we group the image reference targeting different architectures under one virtual image/manifest digest list, all tools should work pretty similarly from what I understand, but let me play around with it tomorrow to confirm.

scholzj:
> Assuming we group the image reference targeting different architectures under one virtual image/manifest digest list, all tools should work pretty similarly from what I understand, but let me play around with it tomorrow to confirm.

If we should introduce another dependency such as `scopeo`, it would be good to double check it returns the right digest for multiarch as we might need it probably soon.

kyguy:
> If we should introduce another dependency such as scopeo, it would be good to double check it returns the right digest for multiarch as we might need it probably soon.

I have tested `skopeo` on a multiarch image I have built and it works as expected, retrieving the digest of the manifest list which contains the digests for the platform specific digests.

I have also found an easy way to retrieve and use auth tokens with curl, I am just testing a few things but I will compare the different options here and we can decide which way forward then

scholzj:
> I have also found an easy way to retrieve and use auth tokens with curl, I am just testing a few things but I will compare the different options here and we can decide which way forward then

I guess auth tokens will suck unless you can easily extract them from the Docker configuration. Is that possible?

kyguy:
> I guess auth tokens will suck unless you can easily extract them from the Docker configuration. Is that possible?

Not easily, and maintaining a cross platform bash script that does so will be a pain, unless we prompt for docker username/passwords at the start of the run of the script.

So I have been experimenting with this today and from what I found,  I think we should stick with `skopeo` for digest retrieval. The difference in code between using `curl` vs `skopeo` for digest retrieval amounts to this:

## Using `curl`:
```
HEADERS="-H \"Accept: application/vnd.docker.distribution.manifest.v2+json\""
if [ "$registry" = "docker.io" ]; then
  TOKEN=$(curl -s "https://auth.docker.io/token?service=registry.docker.io&scope=repository:$org/$repo:pull" | yq ea -P | yq ea ".token" -)
   HEADERS="$HEADERS  -H \"Authorization: Bearer ${TOKEN}\""
fi  

resp="$(curl -s -w '\n%{http_code}' "${HEADERS}" "https://${registry}/v2/${org}/${repo}/manifests/${tag}")"
code=$(tail -1 <<< "$resp")
body=$($SED '$ d' <<< "$resp")

if [ "$code" != "200" ]; then
  echo "ERROR: Could not find image in remote registry: '$registry/$org/$repo:$tag' $code"
  exit 1;
fi

digest=$(echo -n "$body" | sha256sum | $HEAD -c 64);
image_digest="${image}@sha256:${digest}"
```
Dockerhub requires tokens for pulling images/image metadata, even for anonymous users. So we have to retrieve the tokens before making any  metadata requests. We would also need to add some logic here to get registry usernames/passwords if we want to retrieve the metadata of private images as well.

## Using `skopeo`:
```
digest=$(skopeo inspect --format "{{ .Digest }}" "docker://$registry/$org/$repo:$tag")
image_digest="${image}@${digest}"
```
Skopeo will automatically handle DockerHub tokens and uses your docker/quay credentials to pull metadata of private images.

scholzj:
Ok, I guess I leave it up to you to decide which one to use.

kyguy:
Awesome, let's stick with `skopeo` then. I just want to run the change by @tomncooper to verify that we have thoroughly addressed the issues raised. Tom, I know you have a lot cooking at the moment so no pressure on this!

scholzj:
@tomncooper Does it work for you now?

tomncooper:
Hi @kyguy, I ran a `make clean all` build on this branch with `export DOCKER_ORG=tomncooper && export DOCKER_TAG=latest && export DOCKER_REGISTRY=docker.io`.

I then ran `make` in the `tools/olm-bundle` dir:
```
⇒  make 
./generate-olm-bundle.sh BUNDLE_VERSION=0.25.0 DOCKER_REGISTRY=docker.io DOCKER_ORG=tomncooper DOCKER_TAG=latest
Generating bundle version 0.25.0
Generating bundle manifests
WARN[0000] ClusterServiceVersion validation: [OperationFailed] provided API should have an example annotation 
...
Bundle manifests generated successfully in /home/tom/repos/red_hat/strimzi-kafka-operator/tools/olm-bundle/bundle
Generating bundle metadata
INFO[0000] Creating bundle.Dockerfile                   
INFO[0000] Creating /home/tom/repos/red_hat/strimzi-kafka-operator/tools/olm-bundle/bundle/metadata/annotations.yaml 
INFO[0000] Bundle metadata generated suceessfully       
Update CRD filename kafka.strimzi.io_kafkabridges.yaml -> kafkabridges.kafka.strimzi.io.crd.yaml
Update CRD filename kafka.strimzi.io_kafkaconnectors.yaml -> kafkaconnectors.kafka.strimzi.io.crd.yaml
Update CRD filename kafka.strimzi.io_kafkaconnects.yaml -> kafkaconnects.kafka.strimzi.io.crd.yaml
Update CRD filename kafka.strimzi.io_kafkamirrormaker2s.yaml -> kafkamirrormaker2s.kafka.strimzi.io.crd.yaml
Update CRD filename kafka.strimzi.io_kafkamirrormakers.yaml -> kafkamirrormakers.kafka.strimzi.io.crd.yaml
Update CRD filename kafka.strimzi.io_kafkarebalances.yaml -> kafkarebalances.kafka.strimzi.io.crd.yaml
Update CRD filename kafka.strimzi.io_kafkas.yaml -> kafkas.kafka.strimzi.io.crd.yaml
Update CRD filename kafka.strimzi.io_kafkatopics.yaml -> kafkatopics.kafka.strimzi.io.crd.yaml
Update CRD filename kafka.strimzi.io_kafkausers.yaml -> kafkausers.kafka.strimzi.io.crd.yaml
Update CRD filename strimzi-cluster-operator_v1_configmap.yaml -> strimziclusteroperator.configmap.yaml
Get digest from remote registry for image: docker.io/tomncooper/operator:latest
Replacing quay.io/strimzi/operator:latest with quay.io/strimzi/operator@sha256:36457d91c504dfc6451f4d4f046c10ee74578738eadea69400234a05e460f2e8
Get digest from remote registry for image: docker.io/tomncooper/kafka:latest-kafka-2.7.0
Replacing quay.io/strimzi/kafka:latest-kafka-2.7.0 with quay.io/strimzi/kafka@sha256:a8fe5b7489459cf3adf710ccc2f581a84def41c35fe50768402f0b1d2d7a0908
Get digest from remote registry for image: docker.io/tomncooper/kafka:latest-kafka-2.7.1
Replacing quay.io/strimzi/kafka:latest-kafka-2.7.1 with quay.io/strimzi/kafka@sha256:2865b95b32deeb76584ae01e4c8c5f3e2b34c4b39aa3e5490eec953784e966a1
Get digest from remote registry for image: docker.io/tomncooper/kafka:latest-kafka-2.8.0
Replacing quay.io/strimzi/kafka:latest-kafka-2.8.0 with quay.io/strimzi/kafka@sha256:a7985134c92b1a16b0013b57f1e7d5b5054fc3a87dcb4a542f6a762ffa5251d6
Get digest from remote registry for image: docker.io/tomncooper/kafka-bridge:0.20.1
FATA[0001] Error parsing image name "docker://docker.io/tomncooper/kafka-bridge:0.20.1": Error reading manifest 0.20.1 in docker.io/tomncooper/kafka-bridge: manifest unknown: manifest unknown 
```

kyguy:
Sorry @tomncooper I must have messed that up when refactoring. Issue should now be fixed!

tomncooper (approved):
I did a custom build with my own repo, tags etc and it worked great! Awesome work @kyguy!

Frawless (approved):

PR #4453, opened 2021-02-22
ST: Use cmd client for deploy in upgrade tests

Frawless:

### Type of change

- Refactoring

### Description

We should use cmd client for each CR creation during the upgrade because at some point, we will need to deal with different api versions and it cannot be handled by fabric8 client easily.

### Checklist


systemtest/src/main/java/io/strimzi/systemtest/utils/StUtils.java
```diff
@@ -322,4 +324,13 @@ public static String changeDeploymentNamespace(File deploymentFile, String names
...
     }
+
+    public static void waitForResourceReadiness(String resourceType, String resourceName) {
+        LOGGER.info("Waiting for " + resourceType + "/" + resourceName + " readiness");
+
+        TestUtils.waitFor("resource " + resourceType + "/" + resourceName + " readiness",
+                Constants.GLOBAL_POLL_INTERVAL, Constants.GLOBAL_CMD_CLIENT_TIMEOUT,
+            () -> ResourceManager.cmdKubeClient().getResourceReadiness(resourceType, resourceName));
+        LOGGER.info("Resource " + resourceType + "/" + resourceName + " is ready");
+    }
```
im-konge, line 335: Wouldn't be better to add this to `ResourceManager`? Both options - have it in `StUtils` or `ResourceManager` - makes sense to me. So it's up to you

im-konge (approved):
Just one suggestion. LGTM

scholzj (approved):

systemtest/src/main/java/io/strimzi/systemtest/utils/StUtils.java
Frawless: Make sense to move it to ResourceManager

PR #4465, opened 2021-02-23
Update template with smaller volume size

gabrielhao:
### Type of change

_Select the type of your PR_

- Bugfix

### Description

This template is used in Quicstart document, it requires 100GB volume in local environment. This is often impractical and leads local creation of cluster to fail. Therefore, a smaller volume would be better for local environment.

### Checklist

- [ ] Write tests
- [ ] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/examples/kafka/kafka-persistent-single.yaml
+++ b/examples/kafka/kafka-persistent-single.yaml
@@ -26,13 +26,13 @@ spec:
       volumes:
       - id: 0
         type: persistent-claim
-        size: 100Gi
+        size: 2Gi
         deleteClaim: false
   zookeeper:
     replicas: 1
     storage:
       type: persistent-claim
-      size: 100Gi
+      size: 2Gi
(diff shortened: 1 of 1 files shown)


scholzj:
`2Gi` is by far not enough for lot of Kafka work. This was discussed in the past and I still think `100Gi` is a good value. And everyone who wants to use something different can change it on their own.

gabrielhao:
> `2Gi` is by far not enough for lot of Kafka work. This was discussed in the past and I still think `100Gi` is a good value. And everyone who wants to use something different can change it on their own.

I see. Then we need to put this in quickstart document. Otherwise people who follows the steps there will fail to start the demo cluster.

scholzj:
Thanks for the PR. We discussed this on the Strimzi Community meeting. We will check the docs if there is some improvement, but want to stick with the current example.

PR #4473, opened 2021-02-25
[DOC] mm2 topic offset synchronization

PaulRMellor:

**Documentation**

New section describing _synchronizing consumer group offsets_ added to the MirrorMaker 2.0 _Cluster Configuration_ section
Example config added to the procedure in _Synchronizing data between Kafka clusters using MirrorMaker 2.0_

### Checklist

- [ ] Write tests
- [ ] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/documentation/modules/configuring/con-config-mirrormaker2-replication.adoc
+++ b/documentation/modules/configuring/con-config-mirrormaker2-replication.adoc
@@ -15,10 +15,6 @@ MirrorMaker 2.0 uses:
 MirrorMaker 2.0 is based on the Kafka Connect framework, _connectors_ managing the transfer of data between clusters.
 A MirrorMaker 2.0 `MirrorSourceConnector` replicates topics from a source cluster to a target cluster.
 
-By default, a check for new topics in the source cluster is made every 10 minutes.
-You can change the frequency by adding `refresh.topics.interval.seconds` to the source connector configuration of the `KafkaMirrorMaker2` resource.
-However, increasing the frequency of the operation might affect overall performance.
-
 The process of _mirroring_ data from one cluster to another cluster is asynchronous.
 The recommended pattern is for messages to be produced locally alongside the source Kafka cluster, then consumed remotely close to the target Kafka cluster.
 
@@ -26,6 +22,10 @@ MirrorMaker 2.0 can be used with more than one source cluster.
 
 .Replication across two clusters
 image::mirrormaker.png[MirrorMaker 2.0 replication]
+
+By default, a check for new topics in the source cluster is made every 10 minutes.
+You can change the frequency by adding `refresh.topics.interval.seconds` to the source connector configuration of the `KafkaMirrorMaker2` resource.
+However, increasing the frequency of the operation might affect overall performance.
 
 = Cluster configuration
 
@@ -87,25 +87,27 @@ MirrorMaker 2.0 uses its `MirrorCheckpointConnector` to emit _checkpoints_ for o
 
 == Synchronizing consumer group offsets
 
-The `__consumer_offsets` topic stores information on committed offsets, according to consumer group.
+The `__consumer_offsets` topic stores information on committed offsets, for each consumer group.
 Offset synchronization periodically transfers the consumer offsets for the consumer groups of a source cluster into the consumer offsets topic of a target cluster.
 
 Offset synchronization is particularly useful in an _active/passive_ configuration.
-Should the active cluster go down, consumer applications can switch to the passive (standby) cluster and pick up from the last offset position.
+If the active cluster goes down, consumer applications can switch to the passive (standby) cluster and pick up from the last transferred offset position.
 
 To use topic offset synchronization:
 
 * Enable the synchronization by adding `sync.group.offsets.enabled` to the checkpoint connector configuration of the `KafkaMirrorMaker2` resource,
 and setting the property to `true`. Synchronization is disabled by default.
-* Add the `IdentityReplicationPolicy` to the source connector configuration so that topics in the target cluster retain their original names.
+* Add the `IdentityReplicationPolicy` to the source and checkpoint connector configuration so that topics in the target cluster retain their original names.
 
-For topic offset synchronization to work, consumer groups in the source cluster cannot be active in the target cluster.
-Consequently, if you are using an _active/active_ configuration that does not support or require offset synchronization, do not enable topic offset synchronization.
+For topic offset synchronization to work, consumer groups in the target cluster cannot use the same ids as groups in the source cluster.
 
 If enabled, the synchronization of offsets from the source cluster is made periodically.
 You can change the frequency by adding `sync.group.offsets.interval.seconds` and `emit.checkpoints.interval.seconds` to the checkpoint connector configuration.
 The properties specify the frequency in seconds that the consumer group offsets are synchronized, and the frequency of checkpoints emitted for offset tracking.
 The default for both properties is 60 seconds.
+You might also want to change the frequency of checks for new consumer groups using the `refresh.groups.interval.seconds` property, which is performed every 10 minutes by default.
+
+Because the synchronization is time-based, any switchover by consumers to a passive cluster will likely result in some duplication of messages.
 
 == Connectivity checks
 
--- a/documentation/modules/configuring/con-config-mirrormaker2-replication.adoc
+++ b/documentation/modules/configuring/con-config-mirrormaker2-replication.adoc
@@ -105,7 +105,7 @@ If enabled, the synchronization of offsets from the source cluster is made perio
 You can change the frequency by adding `sync.group.offsets.interval.seconds` and `emit.checkpoints.interval.seconds` to the checkpoint connector configuration.
 The properties specify the frequency in seconds that the consumer group offsets are synchronized, and the frequency of checkpoints emitted for offset tracking.
 The default for both properties is 60 seconds.
-You might also want to change the frequency of checks for new consumer groups using the `refresh.groups.interval.seconds` property, which is performed every 10 minutes by default.
+You can also change the frequency of checks for new consumer groups using the `refresh.groups.interval.seconds` property, which is performed every 10 minutes by default.
 
 Because the synchronization is time-based, any switchover by consumers to a passive cluster will likely result in some duplication of messages.
 
--- a/documentation/modules/configuring/con-config-mirrormaker2-replication.adoc
+++ b/documentation/modules/configuring/con-config-mirrormaker2-replication.adoc
@@ -1,6 +1,6 @@
 // Module included in the following assemblies:
 //
-// assembly-mirrormaker.adoc
+// assembly-config-mirrormaker2.adoc
 
 [id='con-mirrormaker-{context}']
 = MirrorMaker 2.0 data replication
@@ -14,6 +14,10 @@ MirrorMaker 2.0 uses:
 
 MirrorMaker 2.0 is based on the Kafka Connect framework, _connectors_ managing the transfer of data between clusters.
 A MirrorMaker 2.0 `MirrorSourceConnector` replicates topics from a source cluster to a target cluster.
+
+By default, a check for new topics in the source cluster is made every 10 minutes.
+You can change the frequency by adding `refresh.topics.interval.seconds` to the source connector configuration of the `KafkaMirrorMaker2` resource.
+However, increasing the frequency of the operation might affect overall performance.
 
 The process of _mirroring_ data from one cluster to another cluster is asynchronous.
 The recommended pattern is for messages to be produced locally alongside the source Kafka cluster, then consumed remotely close to the target Kafka cluster.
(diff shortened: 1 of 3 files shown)


documentation/modules/configuring/con-config-mirrormaker2-replication.adoc
```diff
@@ -81,6 +85,28 @@ Both topics enable replication to be fully restored from the correct offset posi
...
+Should the active cluster go down, consumer applications can switch to the passive (standby) cluster and pick up from the last offset position.
+
+To use topic offset synchronization:
+
+* Enable the synchronization by adding `sync.group.offsets.enabled` to the checkpoint connector configuration of the `KafkaMirrorMaker2` resource,
+and setting the property to `true`. Synchronization is disabled by default.
+* Add the `IdentityReplicationPolicy` to the source connector configuration so that topics in the target cluster retain their original names.
+
+For topic offset synchronization to work, consumer groups in the source cluster cannot be active in the target cluster.
+Consequently, if you are using an _active/active_ configuration that does not support or require offset synchronization, do not enable topic offset synchronization.
```
scholzj, line 103: I'm not sure I understand this sentence. In general, you can use offset synchronization also in active-active clusters. And since you anyway cannot easily use a single consumer group over multiple clusters, this might not be a big limitation I think. So maybe we can remove this sentence?

documentation/modules/configuring/con-config-mirrormaker2-replication.adoc
```diff
@@ -81,6 +85,28 @@ Both topics enable replication to be fully restored from the correct offset posi
...
+
+If enabled, the synchronization of offsets from the source cluster is made periodically.
+You can change the frequency by adding `sync.group.offsets.interval.seconds` and `emit.checkpoints.interval.seconds` to the checkpoint connector configuration.
+The properties specify the frequency in seconds that the consumer group offsets are synchronized, and the frequency of checkpoints emitted for offset tracking.
+The default for both properties is 60 seconds.
```
scholzj, line 108: Should this mention also the `refresh.groups.interval.seconds` interval? (default 10 minutes)

documentation/modules/configuring/proc-config-mirrormaker2-replication.adoc
```diff
@@ -111,38 +111,42 @@ spec:
         replication.factor: 1 <20>
         offset-syncs.topic.replication.factor: 1 <21>
         sync.topic.acls.enabled: "false" <22>
-        replication.policy.separator: "" <23>
-        replication.policy.class: "io.strimzi.kafka.connect.mirror.IdentityReplicationPolicy" <24>
-    heartbeatConnector: <25>
+        refresh.topics.interval.seconds: 60 <23>
+        replication.policy.separator: "" <24>
+        replication.policy.class: "io.strimzi.kafka.connect.mirror.IdentityReplicationPolicy" <25>
```
scholzj, line 116: To make this example valid, you have to also add this option to the `checkpointConnector.config`

scholzj (commented):
Soma smaller comments. But looks good otherwise.

documentation/modules/configuring/con-config-mirrormaker2-replication.adoc
```diff
@@ -15,6 +15,10 @@ MirrorMaker 2.0 uses:
 MirrorMaker 2.0 is based on the Kafka Connect framework, _connectors_ managing the transfer of data between clusters.
 A MirrorMaker 2.0 `MirrorSourceConnector` replicates topics from a source cluster to a target cluster.
 
+By default, a check for new topics in the source cluster is made every 10 minutes.
+You can change the frequency by adding `refresh.topics.interval.seconds` to the source connector configuration of the `KafkaMirrorMaker2` resource.
+However, increasing the frequency of the operation might affect overall performance.
+
```
laidan6000, line 21: I have limited knowledge of MM2, but the new paragraph seems to interrupt the flow of the original introduction (which is great). It talks about the `KafkaMirrorMaker2` resource very early on.

Perhaps it could be positioned after line 25, or even in the "Cluster configuration" section.

documentation/modules/configuring/con-config-mirrormaker2-replication.adoc
```diff
@@ -81,6 +85,28 @@ Both topics enable replication to be fully restored from the correct offset posi
 
 MirrorMaker 2.0 uses its `MirrorCheckpointConnector` to emit _checkpoints_ for offset tracking.
 
+== Synchronizing consumer group offsets
+
+The `__consumer_offsets` topic stores information on committed offsets, according to consumer group.
+Offset synchronization periodically transfers the consumer offsets for the consumer groups of a source cluster into the consumer offsets topic of a target cluster.
```
laidan6000, line 91: "Offset synchronization" is a MM2 concept, correct?

documentation/modules/configuring/con-config-mirrormaker2-replication.adoc
```diff
@@ -81,6 +85,28 @@ Both topics enable replication to be fully restored from the correct offset posi
...
+
+Offset synchronization is particularly useful in an _active/passive_ configuration.
+Should the active cluster go down, consumer applications can switch to the passive (standby) cluster and pick up from the last offset position.
```
laidan6000, line 94:
```suggestion
If the active cluster goes down, consumer applications can switch to the passive (standby) cluster and pick up from the last offset position.
```

documentation/modules/configuring/con-config-mirrormaker2-replication.adoc
```diff
@@ -81,6 +85,28 @@ Both topics enable replication to be fully restored from the correct offset posi
...
+
+To use topic offset synchronization:
+
+* Enable the synchronization by adding `sync.group.offsets.enabled` to the checkpoint connector configuration of the `KafkaMirrorMaker2` resource,
+and setting the property to `true`. Synchronization is disabled by default.
+* Add the `IdentityReplicationPolicy` to the source connector configuration so that topics in the target cluster retain their original names.
+
+For topic offset synchronization to work, consumer groups in the source cluster cannot be active in the target cluster.
```
laidan6000, line 102:
```suggestion
For topic offset synchronization to work, consumer groups in the source cluster cannot also be active in the target cluster.
```

laidan6000 (changes requested):
New feature is well explained. I asked a question about the introduction.

documentation/modules/configuring/con-config-mirrormaker2-replication.adoc
```diff
@@ -81,6 +85,28 @@ Both topics enable replication to be fully restored from the correct offset posi
 
 MirrorMaker 2.0 uses its `MirrorCheckpointConnector` to emit _checkpoints_ for offset tracking.
 
+== Synchronizing consumer group offsets
+
+The `__consumer_offsets` topic stores information on committed offsets, according to consumer group.
```
tombentley, line 90:
```suggestion
The `__consumer_offsets` topic stores information on committed offsets, for each consumer group.
```

documentation/modules/configuring/con-config-mirrormaker2-replication.adoc
tombentley, line 94:
```suggestion
Should the active cluster go down, consumer applications can switch to the passive (standby) cluster and pick up from the last transferred offset position.
```

documentation/modules/configuring/con-config-mirrormaker2-replication.adoc
tombentley, line 102: I think we can phrase this clearer, since AFAIU the problem is just about colliding group ids, rather than use the of consumer groups _per se_.

```suggestion
For topic offset synchronization to work, consumer groups in the target cluster cannot use the same ids as groups in  the source cluster.
```

documentation/modules/configuring/con-config-mirrormaker2-replication.adoc
tombentley: We should also be explicit that because this is time based it means that consumers which switch to the passive on failover will process some (hopefully small) number of duplicates.

documentation/modules/configuring/con-config-mirrormaker2-replication.adoc
PaulRMellor: Removed

documentation/modules/configuring/con-config-mirrormaker2-replication.adoc
PaulRMellor: Updated to:

> If enabled, the synchronization of offsets from the source cluster is made periodically.
> You can change the frequency by adding `sync.group.offsets.interval.seconds` and `emit.checkpoints.interval.seconds` to the checkpoint connector configuration.
> The properties specify the frequency in seconds that the consumer group offsets are synchronized, and the frequency of checkpoints emitted for offset tracking.
> The default for both properties is 60 seconds.
> You might also want to change the frequency of checks for new consumer groups using the `refresh.groups.interval.seconds` property, which is performed every 10 minutes by default.
> 
> Because the synchronization is time-based, any switchover by consumers to a passive cluster will likely result in some duplication of messages.

Also added `refresh.groups.interval.seconds` to example config

documentation/modules/configuring/proc-config-mirrormaker2-replication.adoc
PaulRMellor: Added. I've left the description here, but expanded: _To configure topic offset synchronization, this property must also be set for the `checkpointConnector.config`._

Also changed the concepts: _Add the `IdentityReplicationPolicy` to the source and checkpoint connector configuration so that topics in the target cluster retain their original names._

documentation/modules/configuring/con-config-mirrormaker2-replication.adoc
PaulRMellor: Fair point. I've moved to after the image

documentation/modules/configuring/con-config-mirrormaker2-replication.adoc
PaulRMellor: Yes

tombentley (approved):

documentation/modules/configuring/con-config-mirrormaker2-replication.adoc
```diff
@@ -81,6 +85,30 @@ Both topics enable replication to be fully restored from the correct offset posi
...
+and setting the property to `true`. Synchronization is disabled by default.
+* Add the `IdentityReplicationPolicy` to the source and checkpoint connector configuration so that topics in the target cluster retain their original names.
+
+For topic offset synchronization to work, consumer groups in the target cluster cannot use the same ids as groups in the source cluster.
+
+If enabled, the synchronization of offsets from the source cluster is made periodically.
+You can change the frequency by adding `sync.group.offsets.interval.seconds` and `emit.checkpoints.interval.seconds` to the checkpoint connector configuration.
+The properties specify the frequency in seconds that the consumer group offsets are synchronized, and the frequency of checkpoints emitted for offset tracking.
+The default for both properties is 60 seconds.
+You might also want to change the frequency of checks for new consumer groups using the `refresh.groups.interval.seconds` property, which is performed every 10 minutes by default.
```
scholzj, line 108: TBH, I'm not sure we should suggest the change. We do not really know what impact these have on performance etc. So I think saying _You can also change ..._ would be better.

documentation/modules/configuring/proc-config-mirrormaker2-replication.adoc
```diff
@@ -111,38 +111,44 @@ spec:
         replication.factor: 1 <20>
         offset-syncs.topic.replication.factor: 1 <21>
         sync.topic.acls.enabled: "false" <22>
-        replication.policy.separator: "" <23>
-        replication.policy.class: "io.strimzi.kafka.connect.mirror.IdentityReplicationPolicy" <24>
-    heartbeatConnector: <25>
+        refresh.topics.interval.seconds: 60 <23>
+        refresh.groups.interval.seconds: 600 <24>
```
scholzj, line 115: I think this needs to be in the `checkpointConnector` section.

scholzj (commented):
Some more nits ... LGTM otherwise.

documentation/modules/configuring/con-config-mirrormaker2-replication.adoc
PaulRMellor: Changed to: _You can also change the frequency of checks for new consumer groups using the `refresh.groups.interval.seconds` property, which is performed every 10 minutes by default._

documentation/modules/configuring/proc-config-mirrormaker2-replication.adoc
PaulRMellor: Moved

scholzj (approved):

scholzj:
@laidan6000 Does this look good to merge now?

laidan6000 (approved):

PR #4485, opened 2021-03-01
Use ASF mirrors to download Kafka binaries

scholzj:
### Type of change

- Bugfix

### Description

This PR tries to change the download URLs for the Kafka binaries to use the mirrors instead of the main dist site.


scholzj:
Turns out it is not easy as this. Right now, only the last two releases (2.7.0 and 2.6.1) are on the mirrors. the rest is only in the archives. Closing this PR as it would not help much and need changes quite often.

PR #4493, opened 2021-03-02
Try to speed up downloads of Kafka binaries - Attempt No 2

scholzj:
### Type of change

- Bugfix

### Description

This PR tries to use two mechanisms to speed up things:
* It tries to cache the Kafka artifacts
* It tries to use the mirrors for the latest versions available on them

It also increases the maximal build time because we need to finish the build to get the cache first.


sknot-rh (approved):
LGTM

Frawless (approved):

PR #4503, opened 2021-03-03
Make it possible to generate CRDs without descriptions for some versions

scholzj:
### Type of change

- Enhancement / new feature

### Description

In some cases it might be useful to be able to disable generating descriptions for all API versions in the CRDs in order to make the files smaller. This enhancement to the CrdGenerator allows to specify versions for which should the descriptions be generated. This is currently not used for the CRDs in the repo, but it is available when needed for example to generate OLM files etc.

### Checklist

- [x] Write tests

### DIFF
--- a/crd-generator/src/main/java/io/strimzi/crdgenerator/CrdGenerator.java
+++ b/crd-generator/src/main/java/io/strimzi/crdgenerator/CrdGenerator.java
@@ -172,7 +172,7 @@ public class CrdGenerator {
     private final List<ApiVersion> generateVersions;
     private final ApiVersion storageVersion;
     private final VersionRange<ApiVersion> servedVersion;
-    private final List<ApiVersion> describeVersions;
+    private final VersionRange<ApiVersion> describeVersions;
     // TODO CrdValidator
     // extraProperties
     // @Buildable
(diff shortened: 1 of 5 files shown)


crd-generator/src/main/java/io/strimzi/crdgenerator/CrdGenerator.java
```diff
@@ -280,13 +281,15 @@ public CrdGenerator(VersionRange<KubeVersion> targetKubeVersions, ApiVersion crd
...
+     * @param describeVersions The API versions to which descriptions should be added
      */
     public CrdGenerator(VersionRange<KubeVersion> targetKubeVersions, ApiVersion crdApiVersion,
                  ObjectMapper mapper, Map<String, String> labels, Reporter reporter,
                  List<ApiVersion> apiVersions,
                  ApiVersion storageVersion,
                  VersionRange<ApiVersion> servedVersions,
-                 ConversionStrategy conversionStrategy) {
+                 ConversionStrategy conversionStrategy,
+                 List<ApiVersion> describeVersions) {
```
tombentley, line 292: Does it need to be a list? Our could it be a `VersionRange`? `v1beta1+` saves having to update  the endpoint for `v1beta3`

tombentley (approved):
One minor optional comment, otherwise LGTM.

crd-generator/src/main/java/io/strimzi/crdgenerator/CrdGenerator.java
scholzj: Ok, I changed it to use range instead of list.

PR #4517, opened 2021-03-05
[systemtest] Add test for configurating logging with CM which doesn't exist

im-konge:

### Type of change

- New test

### Description

After fix in #4490 when we are trying to set logging to use `ConfigMap` which doesn't exist, the default logging is used (in `custom-config/log4j.properties`). This PR adds test for it.

### Checklist

- [x] Write tests

### DIFF
--- a/systemtest/src/test/java/io/strimzi/systemtest/log/LoggingChangeST.java
+++ b/systemtest/src/test/java/io/strimzi/systemtest/log/LoggingChangeST.java
@@ -1130,8 +1130,7 @@ class LoggingChangeST extends AbstractST {
             "log4j.logger.org.I0Itec.zkclient.ZkClient=INFO\n" +
             "log4j.logger.org.apache.zookeeper=INFO\n" +
             "log4j.logger.kafka=INFO\n" +
-            "log4j.logger.org.apache.kafka=INFO\n" +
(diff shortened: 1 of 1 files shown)


im-konge:
@strimzi-ci run tests test_only profile=regression testcase=LoggingChangeST#testNotExistingCMSetsDefaultLogging

Frawless (approved):

sknot-rh (approved):

scholzj (commented):
And is that really the behaviour we want? I think if the user told us to use logging from config map which does not exist it should throw InvalidResourceException and not just override it.

sknot-rh:
@scholzj We are just printing the warning. IF you think we should raise an exception, please let me know 
https://github.com/strimzi/strimzi-kafka-operator/blob/a8292e655731376c0a21ddb09f4fd1951a43bb47/cluster-operator/src/main/java/io/strimzi/operator/cluster/model/AbstractModel.java#L510

scholzj:
@sknot-rh Maybe worth discussion on Slack or something to get more opinions on this?

im-konge:
@strimzi-ci run tests test_only profile=regression testcase=LoggingChangeST#testNotExistingCMSetsDefaultLogging

scholzj (approved):

PR #4521, opened 2021-03-06
Remove confusing failed to acquire lock warning

scholzj:
### Type of change

- Bugfix
- Enhancement / new feature

### Description

The `Failed to acquire lock` warning is confusing for the users. The main purpose of this warning is to know from the log when reconciliation gets stuck. But in 99% of cases it does not mean anything wrong =>  it just means another reconciliation is in progress. Bu tit gives users the impression that this indicates some problem.

This PR moves this log message to debug level. And instead, adds a periodic timer which for every minute logs an infor message that the reconciliation is still running:

```
2021-03-06 22:48:32 INFO  AbstractOperator:397 - Reconciliation #2(watch) Kafka(myproject/my-cluster): Reconciliation is in progress
```

This should be less confusing but should serve equally well to see when a reconciliation would be taking too long.

### DIFF
--- a/operator-common/src/main/java/io/strimzi/operator/common/AbstractOperator.java
+++ b/operator-common/src/main/java/io/strimzi/operator/common/AbstractOperator.java
@@ -75,6 +75,7 @@ public abstract class AbstractOperator<
 
     private static final Logger log = LogManager.getLogger(AbstractOperator.class);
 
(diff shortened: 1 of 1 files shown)


kornys:
Thank you for this 😃

ppatierno (approved):

PR #4536, opened 2021-03-09
Move Topic and User Operators back to v1beta1 to allow downgrades

scholzj:
### Type of change

- Bugfix

### Description

We originally decided to use v1beta2 resources in User and Topic Operators. This was because when upgrading to 0.23, the v1beta1 version would not be available which would result in deleting the topics (Fabric8 treats HTTP 404 as _resource does not exist_ which tricks the operator into thinking that the KafkaTopic or KafkaUser resource does not exist anymore). However, this obviously screws up the downgrade.

So this PR approaches it differently. It keeps using v1beta1 in User and Topic Operator. But it also keeps the v1alpha1 and v1beta1 versions in the KafkaTopic and KafkaUser CRDs v1s. That means that even during the upgrade to 0.23, we will still have these versions and they will be removed only later after the UO and TO are fully migrated to v1beta2.

Since the KafkaTopic and KafkaUser CRDs actually do not have any deprecated or non-structural fields, it should be possible to keep the older versions even in the CRD v1 APIs and should allow this workaround which would otherwise not be possible for example for the Kafka CRD.

### Checklist

- [x] Write tests


api-conversion/src/test/java/io/strimzi/kafka/api/conversion/cli/CliTestUtils.java
```diff
@@ -233,8 +234,13 @@ public static void crdHasV1Beta2Only(KubernetesClient client)    {
             String crdName = CrdUpgradeCommand.CRD_NAMES.get(kind);
             CustomResourceDefinition crd = client.apiextensions().v1beta1().customResourceDefinitions().withName(crdName).get();
 
-            assertThat(crd.getSpec().getVersions().size(), is(1));
-            assertThat(crd.getSpec().getVersions().stream().map(CustomResourceDefinitionVersion::getName).collect(toList()), contains("v1beta2"));
+            if (kind.equals("KafkaTopic") || kind.equals("KafkaUser"))  {
```
tombentley, line 237: Can we add a comment for this special case?

api/pom.xml
```diff
@@ -214,6 +212,30 @@
                             </arguments>
                         </configuration>
                     </execution>
+                    <execution>
+                        <id>generate-crd-co-install-v1-eo</id>
+                        <phase>process-classes</phase>
```
tombentley, line 217: Again, let's comment the special case here.

tombentley (approved):
Good work figuring this out. Nits only, otherwise LGTM.

PR #4399, opened 2021-02-12
Pause reconciliations

sknot-rh:
Signed-off-by: Stanislav Knot <sknot@redhat.com>

### Type of change
- Enhancement / new feature

### Description
`strimzi.io/pause-reconciliation: "true"` annotation allows to pause reconciliations.
That may come useful if we want to create k8s resource but not the `$resource`.

Each operator watches resources of some `Kind`. When some change is done in k8s resource, this change is reflected. If reconciliations are paused, no change is reflected.
The TopicOperator is special because it works bidirectionally. That means if the kafka topic is created in kafka, the k8s resource is still created. (This may be discussed).

### Checklist

_Please go through this checklist and make sure all applicable tasks have been done_

- [ ] Write tests
- [ ] Make sure all tests pass
- [ ] Update documentation
- [ ] Check RBAC rights for Kubernetes / OpenShift roles
- [ ] Try your changes from Pod inside your Kubernetes and OpenShift cluster, not just locally
- [ ] Reference relevant issue(s) and close them after merging
- [ ] Update CHANGELOG.md
- [ ] Supply screenshots for visual changes, such as Grafana dashboards

PR #4537, opened 2021-03-09
[systemtest] Add tests for pausing reconciliation feature

im-konge:
### Type of change

- New tests

### Description

After #4399 we are able to add annotation to resources about pausing the reconciliations (`strimzi.io/pause-reconciliation: "true"`), which causes that resource is not created (if the anno is added when we deploying it) or changes are not reflected. This PR adds tests for this new feature.

_**NOTE**: currently KafkaTopic checks are commented because of issues connected to reconciliation feature, after the problems will be fixed, these lines will be uncommented._

### Checklist

- [X] Write tests

### DIFF
--- a/systemtest/src/main/java/io/strimzi/systemtest/utils/kafkaUtils/KafkaConnectorUtils.java
+++ b/systemtest/src/main/java/io/strimzi/systemtest/utils/kafkaUtils/KafkaConnectorUtils.java
@@ -127,7 +127,10 @@ public class KafkaConnectorUtils {
         return getConnectorConfig(podName, connectorName, apiUrl);
     }
 
-    public static void waitForConnectorsSpecStability(String podName, String connectorName, String oldSpec) {
+    /**
+     * Checks stability of Connector's spec on Connect API, which should be same like before changes
+     */
+    public static void waitForConnectorSpecFromConnectAPIStability(String podName, String connectorName, String oldSpec) {
         int[] stableCounter = {0};
 
         TestUtils.waitFor("Connector's spec will be stable", Constants.GLOBAL_POLL_INTERVAL, Constants.GLOBAL_STATUS_TIMEOUT, () -> {
--- a/systemtest/src/main/java/io/strimzi/systemtest/utils/kafkaUtils/KafkaConnectorUtils.java
+++ b/systemtest/src/main/java/io/strimzi/systemtest/utils/kafkaUtils/KafkaConnectorUtils.java
@@ -126,4 +126,24 @@ public class KafkaConnectorUtils {
             () -> !oldConfig.equals(getConnectorConfig(podName, connectorName, apiUrl)));
         return getConnectorConfig(podName, connectorName, apiUrl);
     }
+
+    public static void waitForConnectorsSpecStability(String podName, String connectorName, String oldSpec) {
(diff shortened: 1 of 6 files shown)


systemtest/src/main/java/io/strimzi/systemtest/utils/kafkaUtils/KafkaConnectorUtils.java
```diff
@@ -126,4 +126,24 @@ public static String waitForConnectorConfigUpdate(String podName, String connect
             () -> !oldConfig.equals(getConnectorConfig(podName, connectorName, apiUrl)));
         return getConnectorConfig(podName, connectorName, apiUrl);
     }
+
+    public static void waitForConnectorsSpecStability(String podName, String connectorName, String oldSpec) {
```
Frawless, line 130: Shouldn't be `spec` not used as string but as farbci8 `spec` ?

Why just spec? Shouldn't we check also generation for changes? Or spec is enough?

im-konge: I thought that it would be better to grep `spec` from the API. 

And idea of getting whole config is good, I'll change it. Thanks

scholzj: The `.spec` is actually changed only by the user (the STs here). The operator would change only the status. So it might be good to clarify what does it actually test and what needs to be stable.

im-konge: I wanted to test that when we change anything (in the `KafkaConnector` case the `taskMax`), the changes will not be applied if the reconciliations are paused. So, that's why I added there these stability "waits" and why I'm checking the specs. Maybe I can add some more comments to the tests.

scholzj: So ... it compares the REST API connector with the custom resource? I think that makes sense ... maybe the name of the method could be more explanatory and you can add some comment.

im-konge: Yeah @Frawless said completely same thing :smile: I'll change it, thanks :smile:

scholzj:
FYI: The test failure seems to be related to the PR.

im-konge:
I'll take a look at it. Is it possible that images were not builded yet?

scholzj:
> I'll take a look at it. Is it possible that images were not builded yet? Also I should do the rebase.

The images are build as part of the regression pipeline AFAIK. So I do not think that should be a problem.

im-konge:
@strimzi-ci run tests test_only profile=regression testcase=ReconciliationST

im-konge:
@strimzi-ci run tests test_only profile=regression testcase=ReconciliationST

im-konge:
@strimzi-ci run tests test_only profile=regression testcase=ReconciliationST

Frawless (approved):

scholzj (approved):

Issue #4519, opened 2021-03-05
[Feature] Provide ability to configure a custom Authorizer class


rareddy:
Provide the ability to configure a custom Authorizer implementation other than Simple, OAuth and OPA.  Currently, there is no way to achieve configure a custom Authorizer. Providing the ability to user-defined Authorizer can extend this feature to other variety of Authorization implementations and will keep the feature extensible.

**Describe**
Define a new type called "custom" in the "authorization" section of the CRD, and also configure the class name for the same. If the defined class is found in the classpath then it will be used for configuring the Authorization module for Kafka.

In order to make the custom Authorizer class available in the Strimi image is not part of this request, it will be left as a user exercise to build a private image that has this class.

**Alternatives**
One could use the OPA endpoint to provide an alternative implementation, but that would need a separate process.

**Additional context**
This feature can be useful when defining simple ACLs that may apply to all the users of the cluster rather than per user based ACLs provided by Simple and OAuth methods

scholzj:
Done in #4540

rareddy:
My first Strimzi contribution!!! Thank you @scholzj

PR #4137, opened 2020-12-21
[MO] - [2nd-3rd step paralelism] -> templates, re-worked resources, re-writed ∀ tests

see-quick:
### Type of change

- Enhancement / new feature
- Refactoring

### Description

This PR introducing parallel execution of system tests on the method level. The whole code is re-worked:

1. Change of ResourceMananger -  2 base stacks (method, class) are re-worked to map of stacks that hold every test resource and after the test is complete then delete all
2. Change of Resources - methods, which were used for creating some resources were moved to `ResourceNameTemplates.java`
3. Change system test API for resources - every kind of resources is now encapsulated and only the resource manager can create a resource based on a specific template

- Old API
   -  ResourceName.specificResource()
- New API
   - `resourceManager.create(testContext, ResourceNameTemplate.specificTemlate()..)`
 
### Checklist

- [x] Write tests
- [x] Make sure all tests pass
- [ ] Update documentation

PR #4540, opened 2021-03-09
#4519: Adding ability to configure custom Authorizer implementation

rareddy:

### Type of change
Enhancement

### Description
Adds the ability to configure a custom Authorization implementation class.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,6 +35,7 @@
 * List topics used by a Kafka Connect connector in the `.status` section of the `KafkaConnector` custom resource
 * Bump Cruise Control to v2.5.37 for Kafka 2.7 support. Note this new version of Cruise Control uses `Log4j 2` and is supported by dynamic logging configuration (where logging properties are defined in a ConfigMap). However, existing `Log4j` configurations must be updated to `Log4j 2` configurations.
 * Support pausing reconciliation of CR with annotation `strimzi.io/pause-reconciliation`
+* Support for configuring custom Authorizer implementation 
 
 ### Changes, deprecations and removals
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 * Remove support for Kafka 2.5.x
 * Remove direct ZooKeeper access for handling user quotas in the User Operator. Add usage of Admin Client API instead.
 * Migrate to CRD v1 (required by Kubernetes 1.22+)
+* Support for configuring custom Authorizer implementation 
 
 ### Changes, deprecations and removals
 
@@ -35,7 +36,6 @@
 * List topics used by a Kafka Connect connector in the `.status` section of the `KafkaConnector` custom resource
 * Bump Cruise Control to v2.5.37 for Kafka 2.7 support. Note this new version of Cruise Control uses `Log4j 2` and is supported by dynamic logging configuration (where logging properties are defined in a ConfigMap). However, existing `Log4j` configurations must be updated to `Log4j 2` configurations.
 * Support pausing reconciliation of CR with annotation `strimzi.io/pause-reconciliation`
-* Support for configuring custom Authorizer implementation 
 
 ### Changes, deprecations and removals
 
--- a/api/src/main/java/io/strimzi/api/kafka/model/KafkaAuthorization.java
+++ b/api/src/main/java/io/strimzi/api/kafka/model/KafkaAuthorization.java
@@ -38,7 +38,7 @@ public abstract class KafkaAuthorization implements UnknownPropertyPreserving, S
             "`simple` authorization type uses Kafka's `kafka.security.authorizer.AclAuthorizer` class for authorization. " +
             "`keycloak` authorization type uses Keycloak Authorization Services for authorization. " +
             "`opa` authorization type uses Open Policy Agent based authorization." +
-            "`custom` authorization type uses user defined implementation for authorization.")
+            "`custom` authorization type uses user-provided implementation for authorization.")
     public abstract String getType();
 
     @Override
--- a/api/src/main/java/io/strimzi/api/kafka/model/KafkaAuthorization.java
+++ b/api/src/main/java/io/strimzi/api/kafka/model/KafkaAuthorization.java
@@ -23,7 +23,8 @@ import java.util.Map;
 @JsonSubTypes({
         @JsonSubTypes.Type(name = KafkaAuthorizationSimple.TYPE_SIMPLE, value = KafkaAuthorizationSimple.class),
         @JsonSubTypes.Type(name = KafkaAuthorizationOpa.TYPE_OPA, value = KafkaAuthorizationOpa.class),
-        @JsonSubTypes.Type(name = KafkaAuthorizationKeycloak.TYPE_KEYCLOAK, value = KafkaAuthorizationKeycloak.class)
+        @JsonSubTypes.Type(name = KafkaAuthorizationKeycloak.TYPE_KEYCLOAK, value = KafkaAuthorizationKeycloak.class),
+        @JsonSubTypes.Type(name = KafkaAuthorizationCustom.TYPE_CUSTOM, value = KafkaAuthorizationCustom.class)
 })
 @JsonInclude(JsonInclude.Include.NON_NULL)
 @EqualsAndHashCode
@@ -33,10 +34,11 @@ public abstract class KafkaAuthorization implements UnknownPropertyPreserving, S
     private Map<String, Object> additionalProperties;
 
     @Description("Authorization type. " +
-            "Currently, the supported types are `simple`, `keycloak`, and `opa`. " +
+            "Currently, the supported types are `simple`, `keycloak`, `opa` and `custom`. " +
             "`simple` authorization type uses Kafka's `kafka.security.authorizer.AclAuthorizer` class for authorization. " +
             "`keycloak` authorization type uses Keycloak Authorization Services for authorization. " +
-            "`opa` authorization type uses Open Policy Agent based authorization.")
+            "`opa` authorization type uses Open Policy Agent based authorization." +
+            "`custom` authorization type uses user defined implementation for authorization.")
     public abstract String getType();
 
     @Override
--- a/api/src/main/java/io/strimzi/api/kafka/model/KafkaAuthorizationCustom.java
+++ b/api/src/main/java/io/strimzi/api/kafka/model/KafkaAuthorizationCustom.java
@@ -14,10 +14,9 @@ import io.sundr.builder.annotations.Buildable;
 import lombok.EqualsAndHashCode;
 
 import java.util.List;
-import java.util.Map;
 
 /**
- * Configures the broker authorization
+ * Configures the broker for custom authorization module
  */
 @DescriptionFile
 @Buildable(
@@ -25,7 +24,7 @@ import java.util.Map;
         builderPackage = Constants.FABRIC8_KUBERNETES_API
 )
 @JsonInclude(JsonInclude.Include.NON_NULL)
-@JsonPropertyOrder({"type", "authorizerClass", "config", "superUsers"})
+@JsonPropertyOrder({"type", "authorizerClass", "superUsers"})
 @EqualsAndHashCode
 public class KafkaAuthorizationCustom extends KafkaAuthorization {
     private static final long serialVersionUID = 1L;
@@ -34,7 +33,6 @@ public class KafkaAuthorizationCustom extends KafkaAuthorization {
 
     private String authorizerClass;
     private List<String> superUsers;
-    private Map<String, Object> config;
 
     @Description("Must be `" + TYPE_CUSTOM + "`")
     @Override
@@ -61,18 +59,5 @@ public class KafkaAuthorizationCustom extends KafkaAuthorization {
 
     public void setAuthorizerClass(String clazz) {
         this.authorizerClass = clazz;
-    }
-
-    @Description("Configuration properties for custom authorizer")
-    @Example("config:\n" +
-             "  prop1: value1\n" +
-             "  prop2: value2")
-    @JsonInclude(JsonInclude.Include.NON_NULL)
-    public Map<String, Object> getConfig() {
-        return config;
-    }
-
-    public void setConfig(Map<String, Object> configProperties) {
-        this.config = configProperties;
     }
 }
--- a/api/src/main/java/io/strimzi/api/kafka/model/KafkaAuthorizationCustom.java
+++ b/api/src/main/java/io/strimzi/api/kafka/model/KafkaAuthorizationCustom.java
@@ -1,0 +1,62 @@
+/*
+ * Copyright Strimzi authors.
+ * License: Apache License 2.0 (see the file LICENSE or http://apache.org/licenses/LICENSE-2.0.html).
+ */
+package io.strimzi.api.kafka.model;
+
+import com.fasterxml.jackson.annotation.JsonInclude;
+import com.fasterxml.jackson.annotation.JsonPropertyOrder;
+import io.strimzi.crdgenerator.annotations.Description;
+import io.strimzi.crdgenerator.annotations.DescriptionFile;
+import io.strimzi.crdgenerator.annotations.Example;
+import io.sundr.builder.annotations.Buildable;
+import lombok.EqualsAndHashCode;
+
+import java.util.List;
+
+/**
+ * Configures the broker authorization
+ */
+@DescriptionFile
+@Buildable(
+        editableEnabled = false,
+        builderPackage = Constants.FABRIC8_KUBERNETES_API
+)
+@JsonInclude(JsonInclude.Include.NON_NULL)
+@JsonPropertyOrder({"type", "implementionClass", "superUsers"})
+@EqualsAndHashCode
+public class KafkaAuthorizationCustom extends KafkaAuthorization {
+    private static final long serialVersionUID = 1L;
+
+    public static final String TYPE_CUSTOM = "custom";
+
+    private String implementionClass; 
+    private List<String> superUsers;
+
+    @Description("Must be `" + TYPE_CUSTOM + "`")
+    @Override
+    public String getType() {
+        return TYPE_CUSTOM;
+    }
+
+    @Description("List of super users. Should contain list of user principals which should get unlimited access rights.")
+    @Example("- CN=my-user\n" +
+             "- CN=my-other-user")
+    @JsonInclude(JsonInclude.Include.NON_NULL)
+    public List<String> getSuperUsers() {
+        return superUsers;
+    }
+
+    public void setSuperUsers(List<String> superUsers) {
+        this.superUsers = superUsers;
+    }
+
+    @Description("Authorization Implemention class, must be available in classpath")
+    public String getImplementionClass() {
+        return implementionClass;
+    }
+
+    public void setImplementionClass(String implementionClass) {
+        this.implementionClass = implementionClass;
+    }
+}
--- a/api/src/main/java/io/strimzi/api/kafka/model/KafkaAuthorizationCustom.java
+++ b/api/src/main/java/io/strimzi/api/kafka/model/KafkaAuthorizationCustom.java
@@ -5,7 +5,6 @@
 package io.strimzi.api.kafka.model;
 
 import com.fasterxml.jackson.annotation.JsonInclude;
-import com.fasterxml.jackson.annotation.JsonProperty;
 import com.fasterxml.jackson.annotation.JsonPropertyOrder;
 
 import io.strimzi.crdgenerator.annotations.Description;
@@ -35,7 +34,7 @@ public class KafkaAuthorizationCustom extends KafkaAuthorization {
 
     private String authorizerClass;
     private List<String> superUsers;
-    private Map<String, Object> configProperties;
+    private Map<String, Object> config;
 
     @Description("Must be `" + TYPE_CUSTOM + "`")
     @Override
@@ -43,7 +42,7 @@ public class KafkaAuthorizationCustom extends KafkaAuthorization {
         return TYPE_CUSTOM;
     }
 
-    @Description("List of super users. which are user principals with unlimited access rights.")
+    @Description("List of super users, which are user principals with unlimited access rights.")
     @Example("- CN=my-user\n" +
              "- CN=my-other-user")
     @JsonInclude(JsonInclude.Include.NON_NULL)
@@ -69,12 +68,11 @@ public class KafkaAuthorizationCustom extends KafkaAuthorization {
              "  prop1: value1\n" +
              "  prop2: value2")
     @JsonInclude(JsonInclude.Include.NON_NULL)
-    @JsonProperty(value = "config")
-    public Map<String, Object> getConfigProperties() {
-        return configProperties;
+    public Map<String, Object> getConfig() {
+        return config;
     }
 
-    public void setConfigProperties(Map<String, Object> configProperties) {
-        this.configProperties = configProperties;
+    public void setConfig(Map<String, Object> configProperties) {
+        this.config = configProperties;
     }
 }
--- a/api/src/main/java/io/strimzi/api/kafka/model/KafkaAuthorizationCustom.java
+++ b/api/src/main/java/io/strimzi/api/kafka/model/KafkaAuthorizationCustom.java
@@ -5,7 +5,9 @@
 package io.strimzi.api.kafka.model;
 
 import com.fasterxml.jackson.annotation.JsonInclude;
+import com.fasterxml.jackson.annotation.JsonProperty;
 import com.fasterxml.jackson.annotation.JsonPropertyOrder;
+
 import io.strimzi.crdgenerator.annotations.Description;
 import io.strimzi.crdgenerator.annotations.DescriptionFile;
 import io.strimzi.crdgenerator.annotations.Example;
@@ -13,6 +15,7 @@ import io.sundr.builder.annotations.Buildable;
 import lombok.EqualsAndHashCode;
 
 import java.util.List;
+import java.util.Map;
 
 /**
  * Configures the broker authorization
@@ -23,15 +26,16 @@ import java.util.List;
         builderPackage = Constants.FABRIC8_KUBERNETES_API
 )
 @JsonInclude(JsonInclude.Include.NON_NULL)
-@JsonPropertyOrder({"type", "implementionClass", "superUsers"})
+@JsonPropertyOrder({"type", "authorizerClass", "config", "superUsers"})
 @EqualsAndHashCode
 public class KafkaAuthorizationCustom extends KafkaAuthorization {
     private static final long serialVersionUID = 1L;
 
     public static final String TYPE_CUSTOM = "custom";
 
-    private String implementionClass; 
+    private String authorizerClass;
     private List<String> superUsers;
+    private Map<String, Object> configProperties;
 
     @Description("Must be `" + TYPE_CUSTOM + "`")
     @Override
@@ -39,7 +43,7 @@ public class KafkaAuthorizationCustom extends KafkaAuthorization {
         return TYPE_CUSTOM;
     }
 
-    @Description("List of super users. Should contain list of user principals which should get unlimited access rights.")
+    @Description("List of super users. which are user principals with unlimited access rights.")
     @Example("- CN=my-user\n" +
              "- CN=my-other-user")
     @JsonInclude(JsonInclude.Include.NON_NULL)
@@ -51,12 +55,26 @@ public class KafkaAuthorizationCustom extends KafkaAuthorization {
         this.superUsers = superUsers;
     }
 
-    @Description("Authorization Implemention class, must be available in classpath")
-    public String getImplementionClass() {
-        return implementionClass;
+    @Description("Authorization implementation class, which must be available in classpath")
+    public String getAuthorizerClass() {
+        return authorizerClass;
     }
 
-    public void setImplementionClass(String implementionClass) {
-        this.implementionClass = implementionClass;
+    public void setAuthorizerClass(String clazz) {
+        this.authorizerClass = clazz;
+    }
+
+    @Description("Configuration properties for custom authorizer")
+    @Example("config:\n" +
+             "  prop1: value1\n" +
+             "  prop2: value2")
+    @JsonInclude(JsonInclude.Include.NON_NULL)
+    @JsonProperty(value = "config")
+    public Map<String, Object> getConfigProperties() {
+        return configProperties;
+    }
+
+    public void setConfigProperties(Map<String, Object> configProperties) {
+        this.configProperties = configProperties;
     }
 }
--- a/api/src/test/resources/io/strimzi/api/kafka/model/040-Crd-kafka.yaml
+++ b/api/src/test/resources/io/strimzi/api/kafka/model/040-Crd-kafka.yaml
@@ -742,6 +742,10 @@ spec:
                           so the sooner the job completes. However, using more threads
                           places a heavier load on the authorization server. The default
                           value is 5.
+                      implementionClass:
+                        type: string
+                        description: Authorization Implemention class, must be available
+                          in classpath.
                       initialCacheCapacity:
                         type: integer
                         description: Initial capacity of the local cache used by the
@@ -784,12 +788,15 @@ spec:
                         - simple
                         - opa
                         - keycloak
+                        - custom
                         description: Authorization type. Currently, the supported
-                          types are `simple`, `keycloak`, and `opa`. `simple` authorization
-                          type uses Kafka's `kafka.security.authorizer.AclAuthorizer`
+                          types are `simple`, `keycloak`, `opa` and `custom`. `simple`
+                          authorization type uses Kafka's `kafka.security.authorizer.AclAuthorizer`
                           class for authorization. `keycloak` authorization type uses
                           Keycloak Authorization Services for authorization. `opa`
-                          authorization type uses Open Policy Agent based authorization.
+                          authorization type uses Open Policy Agent based authorization.`custom`
+                          authorization type uses user defined implementation for
+                          authorization.
                       url:
                         type: string
                         example: http://opa:8181/v1/data/kafka/authz/allow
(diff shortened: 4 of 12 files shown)


CHANGELOG.md
```diff
@@ -19,6 +19,7 @@
 * List topics used by a Kafka Connect connector in the `.status` section of the `KafkaConnector` custom resource
 * Bump Cruise Control to v2.5.37 for Kafka 2.7 support. Note this new version of Cruise Control uses `Log4j 2` and is supported by dynamic logging configuration (where logging properties are defined in a ConfigMap). However, existing `Log4j` configurations must be updated to `Log4j 2` configurations.
 * Support pausing reconciliation of CR with annotation `strimzi.io/pause-reconciliation`
+* Support for configuring custom Authorizer implementation
```
scholzj, line 22: This will end up in Strimzi 0.23. So you will need to update this.

api/src/main/java/io/strimzi/api/kafka/model/KafkaAuthorization.java
```diff
@@ -33,10 +34,11 @@
     private Map<String, Object> additionalProperties;
 
     @Description("Authorization type. " +
-            "Currently, the supported types are `simple`, `keycloak`, and `opa`. " +
+            "Currently, the supported types are `simple`, `keycloak`, `opa` and `custom`. " +
             "`simple` authorization type uses Kafka's `kafka.security.authorizer.AclAuthorizer` class for authorization. " +
             "`keycloak` authorization type uses Keycloak Authorization Services for authorization. " +
-            "`opa` authorization type uses Open Policy Agent based authorization.")
+            "`opa` authorization type uses Open Policy Agent based authorization." +
+            "`custom` authorization type uses user defined implementation for authorization.")
```
scholzj, line 41:
```suggestion
            "`custom` authorization type uses user provided implementation for authorization.")
```

api/src/main/java/io/strimzi/api/kafka/model/KafkaAuthorizationCustom.java
```diff
@@ -0,0 +1,62 @@
...
+ */
+@DescriptionFile
+@Buildable(
+        editableEnabled = false,
+        builderPackage = Constants.FABRIC8_KUBERNETES_API
+)
+@JsonInclude(JsonInclude.Include.NON_NULL)
+@JsonPropertyOrder({"type", "implementionClass", "superUsers"})
+@EqualsAndHashCode
+public class KafkaAuthorizationCustom extends KafkaAuthorization {
```
scholzj, line 28: I think we will need to add some `config` section here. I think that would need to be passed to the user configuration section.

api/src/main/java/io/strimzi/api/kafka/model/KafkaAuthorizationCustom.java
```diff
@@ -0,0 +1,62 @@
...
+    public List<String> getSuperUsers() {
+        return superUsers;
+    }
+
+    public void setSuperUsers(List<String> superUsers) {
+        this.superUsers = superUsers;
+    }
+
+    @Description("Authorization Implemention class, must be available in classpath")
+    public String getImplementionClass() {
```
scholzj, line 55: Maybe we can name it just `authorizerClass`? I think that might be more explaining than `implementationClass`.

documentation/api/io.strimzi.api.kafka.model.KafkaAuthorizationCustom.adoc
```diff
@@ -0,0 +1,32 @@
+Custom authorization in Strimzi lets user configure their own `Authorizer` plugin to define the Access Control Lists (ACLs)
+
+ACLs allow you to define which users have access to which resources at a granular level.
+
+Configure the `Kafka` custom resource to use custom authorization.
+Set the `type` property in the `authorization` section to the value `custom`,
+and the set following properties.
```
scholzj, line 7: Can we add here something like ...

```
IMPORTANT: The custom authorizer needs to implement the `org.apache.kafka.server.authorizer.Authorizer` interface and support configuration of super-users from the `super.users` configuration field.
```

or something like that? This is important, because this is needed to configure the Strimzi compoenents.

scholzj (commented):
Thanks for the PR. I left some comments. You will also need to run `make crd_install` to update the CRDs in the Helm Chart.

api/src/main/java/io/strimzi/api/kafka/model/KafkaAuthorizationCustom.java
rareddy: What I have seen with other types in `KafkaBrokerConfigurationBuilder` is passing the key/value pairs, and I believe they get converted `configure(Map<String, Object> config)` method on `Authorizer` class. Since we can not send a complex type as I suspect this content is being serialized, thus I was saying yaml/.properties contents but was afraid of any escaping rules of the content.

scholzj: `Map<String, Object>` is already used on many places through the code ... just look at `.spec.kafka.config` for example. You do not need to deal with any properties or anything.

rareddy: ok, Thanks.  I see there is already `addtionalProperties` with type `Map<String, Object>` in the base class, I will introduce it as a `configuration` section and pass on the properties. so, basically would look like

```
authorization:
  type: custom
  authorizerClass: io.mycompany.CustomAuthorizer
  configuration:
    prop1=value1
    prop2=value2
  superUsers:
    - CN=client_1
    - user_2
    - CN=client_3
```

scholzj: The additional properties have to stay as a separate field. That is used for any fields which would be unknown. Also, the YAML will look like this at the end:

```yaml
authorization:
  type: custom
  authorizerClass: io.mycompany.CustomAuthorizer
  configuration:
    prop1: value1
    prop2: value2
  superUsers:
    - CN=client_1
    - user_2
    - CN=client_3
```

api/src/main/java/io/strimzi/api/kafka/model/KafkaAuthorization.java
PaulRMellor: _user-provided implementation_

documentation/api/io.strimzi.api.kafka.model.KafkaAuthorizationCustom.adoc
```diff
@@ -0,0 +1,32 @@
+Custom authorization in Strimzi lets user configure their own `Authorizer` plugin to define the Access Control Lists (ACLs)
```
PaulRMellor, line 1:
```suggestion
To use custom authorization in Strimzi, you can configure your own `Authorizer` plugin to define Access Control Lists (ACLs).
```

documentation/api/io.strimzi.api.kafka.model.KafkaAuthorizationCustom.adoc
PaulRMellor: IMPORTANT: The custom `Authorizer` must implement the `org.apache.kafka.server.authorizer.Authorizer` interface, and support configuration of super users using the `super.users` configuration property.

documentation/api/io.strimzi.api.kafka.model.KafkaAuthorizationCustom.adoc
```diff
@@ -0,0 +1,32 @@
...
+=== `implementionClass`
+Implementation class of the `org.apache.kafka.server.authorizer.Authorizer` interface to support the custom ACLs
+**Required.**
+
+[id='property-simple-authorization-superusers-{context}']
+=== `superUsers`
+A list of user principals treated as super users, so that they are always allowed without querying ACL rules.
+For more information see xref:con-securing-kafka-authorization-str[Kafka authorization].
+
+.An example of simple authorization configuration
```
PaulRMellor, line 18:
```suggestion
.An example of custom authorization configuration
```

documentation/modules/appendix_crds.adoc
```diff
@@ -985,6 +985,32 @@ It must have the value `keycloak` for the type `KafkaAuthorizationKeycloak`.
...
+
+
+The `type` property is a discriminator that distinguishes use of the `KafkaAuthorizationCustom` type from xref:type-KafkaAuthorizationSimple-{context}[`KafkaAuthorizationSimple`], xref:type-KafkaAuthorizationOpa-{context}[`KafkaAuthorizationOpa`], xref:type-KafkaAuthorizationKeycloak-{context}[`KafkaAuthorizationKeycloak`].
+It must have the value `custom` for the type `KafkaAuthorizationCustom`.
+[options="header"]
+|====
+|Property                  |Description
+|type               1.2+<.<|Must be `custom`.
+|string
+|implementionClass  1.2+<.<|Authorization Implemention class, must be available in classpath.
```
PaulRMellor, line 1008:
```suggestion
|implementionClass  1.2+<.<|Authorization Implementation class, which must be available in classpath.
```

documentation/modules/appendix_crds.adoc
```diff
@@ -985,6 +985,32 @@ It must have the value `keycloak` for the type `KafkaAuthorizationKeycloak`.
...
+|string
+|superUsers         1.2+<.<|List of super users. Should contain list of user principals which should get unlimited access rights.
```
PaulRMellor, line 1010:
```suggestion
|superUsers         1.2+<.<|List of super users, which are user principals with unlimited access rights.
```

install/cluster-operator/040-Crd-kafka.yaml
```diff
@@ -7524,6 +7531,10 @@ spec:
                           so the sooner the job completes. However, using more threads
                           places a heavier load on the authorization server. The default
                           value is 5.
+                      implementionClass:
+                        type: string
+                        description: Authorization Implemention class, must be available
+                          in classpath.
```
PaulRMellor, line 7537:
```suggestion
                        description: Authorization Implementation class, which must be available in the classpath.
```

install/cluster-operator/040-Crd-kafka.yaml
```diff
@@ -15899,6 +15913,10 @@ spec:
...
```
PaulRMellor, line 15919:
```suggestion
                        description: Authorization Implementation class, which must be available in the classpath.
```

install/cluster-operator/040-Crd-kafka.yaml
```diff
@@ -15941,12 +15959,15 @@ spec:
...
-                          types are `simple`, `keycloak`, and `opa`. `simple` authorization
-                          type uses Kafka's `kafka.security.authorizer.AclAuthorizer`
+                          types are `simple`, `keycloak`, `opa` and `custom`. `simple`
+                          authorization type uses Kafka's `kafka.security.authorizer.AclAuthorizer`
                           class for authorization. `keycloak` authorization type uses
                           Keycloak Authorization Services for authorization. `opa`
-                          authorization type uses Open Policy Agent based authorization.
+                          authorization type uses Open Policy Agent based authorization.`custom`
+                          authorization type uses user defined implementation for
+                          authorization.
```
PaulRMellor, line 15970:
```suggestion
                          authorization type uses user-provided implementation for
                          authorization.
```

api/src/test/resources/io/strimzi/api/kafka/model/040-Crd-kafka.yaml
```diff
@@ -781,12 +785,15 @@ spec:
...
-                          types are `simple`, `keycloak`, and `opa`. `simple` authorization
-                          type uses Kafka's `kafka.security.authorizer.AclAuthorizer`
+                          types are `simple`, `keycloak`, `opa` and `custom`. `simple`
+                          authorization type uses Kafka's `kafka.security.authorizer.AclAuthorizer`
                           class for authorization. `keycloak` authorization type uses
                           Keycloak Authorization Services for authorization. `opa`
-                          authorization type uses Open Policy Agent based authorization.
+                          authorization type uses Open Policy Agent based authorization.`custom`
+                          authorization type uses user defined implementation for
+                          authorization.
```
PaulRMellor, line 796:
```suggestion
                          authorization type uses user-provided implementation for
                          authorization.
```

install/cluster-operator/040-Crd-kafka.yaml
```diff
@@ -7566,12 +7577,15 @@ spec:
...
```
PaulRMellor, line 7588:
```suggestion
                          authorization type uses user-provided implementation for
                          authorization.
```

PaulRMellor:
I guess we should update the _Kafka Authorization_ section in the Using Guide.
https://strimzi.io/docs/operators/master/using.html#con-securing-kafka-authorization-str

There are also mentions of authorization options (simple, OPA) in the _Deploying Guide- and _Overview_. 

I can pick that up in a separate PR?

rareddy:
@scholzj @ppatierno is there any way I can create a new custom authorizer and have it part of the image for system test? I spent 90% of my time trying to write a test and I failed :(

scholzj:
> is there any way I can create a new custom authorizer and have it part of the image for system test?

I do not think we would really want it. I think the right way is to test with one of the existing authorizers and just configure it using the `custom` type.

rareddy:
> I think the right way is to test with one of the existing authorizers and just configure it using the custom type.

sounds good, that is the direction I took with test case

api/src/main/java/io/strimzi/api/kafka/model/KafkaAuthorizationCustom.java
```diff
@@ -23,23 +26,24 @@
...
 @EqualsAndHashCode
 public class KafkaAuthorizationCustom extends KafkaAuthorization {
     private static final long serialVersionUID = 1L;
 
     public static final String TYPE_CUSTOM = "custom";
 
-    private String implementionClass; 
+    private String authorizerClass;
     private List<String> superUsers;
+    private Map<String, Object> configProperties;
```
ppatierno, line 38: why using a different name for the field and then the need for the `@JsonProperty(value = "config")` on the getter? I think that just `config` field sounds good imho.

systemtest/src/test/java/io/strimzi/systemtest/security/custom/CustomAuthorizerST.java
```diff
@@ -0,0 +1,80 @@
...
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.is;
+
+@Tag(INTERNAL_CLIENTS_USED)
+public class CustomAuthorizerST extends AbstractST {
+    static final String NAMESPACE = "custom-authorizer-test";
+    static final String CLUSTER_NAME = "custom-authorizer";
+    static final String ADMIN = "sre-admin";
+
+    @BeforeAll
```
Frawless, line 35: Please, move `beforeAll` to the bottom of the class as we have it in all other classes.

systemtest/src/test/java/io/strimzi/systemtest/security/custom/CustomAuthorizerST.java
```diff
@@ -0,0 +1,80 @@
...
+import java.util.Map;
+
+import kafka.security.authorizer.AclAuthorizer;
+
+import static io.strimzi.systemtest.Constants.INTERNAL_CLIENTS_USED;
+import static io.strimzi.test.k8s.KubeClusterResource.kubeClient;
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.is;
+
+@Tag(INTERNAL_CLIENTS_USED)
```
Frawless, line 29: `regression` tag should be added here I guess.

api/src/main/java/io/strimzi/api/kafka/model/KafkaAuthorizationCustom.java
rareddy: sure, I can change it. I was just trying to follow the pattern.

api/src/main/java/io/strimzi/api/kafka/model/KafkaAuthorizationCustom.java
```diff
@@ -23,23 +26,24 @@
...
 
     @Description("Must be `" + TYPE_CUSTOM + "`")
     @Override
     public String getType() {
         return TYPE_CUSTOM;
     }
 
-    @Description("List of super users. Should contain list of user principals which should get unlimited access rights.")
+    @Description("List of super users. which are user principals with unlimited access rights.")
```
PaulRMellor, line 46:
```suggestion
    @Description("List of super users, which are user principals with unlimited access rights.")
```

PaulRMellor: same elsewhere

api/src/main/java/io/strimzi/api/kafka/model/KafkaAuthorizationCustom.java
rareddy: done

systemtest/src/test/java/io/strimzi/systemtest/security/custom/CustomAuthorizerST.java
```diff
@@ -0,0 +1,80 @@
...
+            .withClusterName(CLUSTER_NAME)
+            .withKafkaUsername(USER_NAME)
+            .withMessageCount(MESSAGE_COUNT)
+            .withListenerName(Constants.PLAIN_LISTENER_DEFAULT_NAME)
+            .build();
+
+        // AclAuthorizer without user config will not allow any permission
+        internalKafkaClient.sendMessagesPlain();
+        int received = internalKafkaClient.receiveMessagesPlain();
+        assertThat(received, is(0));
```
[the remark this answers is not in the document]
rareddy: Improved the test now to test the user to send/recv messages and test validity

api/src/main/java/io/strimzi/api/kafka/model/KafkaAuthorizationCustom.java
rareddy: updated docs

mstruk:
I looked at this a bit, and one issue is that currently you can abuse `config` to inject any kind of `key=value` into `strimzi.properties`, circumventing protections declaratively enforced for Kafka CRD `config` for example. Since the actual config keys for custom Authorizer could be anything, they can't be part of declarative constraints (we don't know them ahead of time).

One way to solve this would be to write a validation function that performs these checks programmatically (basically replicates the declarative rules - that's simpler in the code in fact than to enumerate declarations for each key in a json file). 

Another would maybe be to force a declaration of a prefix which would be used to prefix all the custom authorizer config keys in the `strimzi.properties`, but since the Authorizer code will only see the full keys it needs to know the prefix. Since there is only one authorizer in the system we could have a reserved key: `custom.authorizer.prefix`, and the custom authorizer can then filter the passed config using this prefix. Although you need again a validation function for the prefix so that's then just an additional complication, and an inferior solution.

scholzj:
> I looked at this a bit, and one issue is that currently you can abuse config to inject any kind of key=value into strimzi.properties, circumventing protections declaratively enforced for Kafka CRD config for example. Since the actual config keys for custom Authorizer could be anything, they can't be part of declarative constraints (we don't know them ahead of time).

I didn't got to review it again yet. But yeah, the config section should probably be added to the `.spec.kafka.config` values and passed through the same validation rules? Or maybe we should think if it is even needed and if the values should not be provided directly there to make things easier? Maybe I lead it in the wrong direction by suggesting the config should be there.

rareddy:
>Another would maybe be to force a declaration of a prefix which would be used to prefix all the custom authorizer config keys in the strimzi.properties, but since the Authorizer code will only see the full keys it needs to know the prefix.

I have taken exactly this approach, but the Authorizer does not need to know the prefix as it is custom it can be designed to read for that fixed prefix, but yes validation will be an issue. Alternatively, for validation one other simple way is to use the simple class name of the custom authorizer as prefix and not include any property that does not prefixed with that name. For Ex if custom authorizer is `io.project.Foobar` we can simply use `FooBar` as prefix and filter them when configuring it. This can be documented to avoid any confusion.

>maybe we should think if it is even needed and if the values should not be provided directly there to make things easier? >Maybe I lead it in the wrong direction by suggesting the config should be there.

having the configuration definitely made my implementation a lot simpler so recommend we keep it. otherwise, especially with the container environment it will hard to access a file/configmap etc for configuration. In my usecase, it kept me by not using kubeclient to go grab a known configmap in Authorizer implementation.

But I will make changes however you guys suggest as I am just getting familiar with Strimzi.

scholzj:
> I have taken exactly this approach, but the Authorizer does not need to know the prefix as it is custom it can be designed to read for that fixed prefix, but yes validation will be an issue. Alternatively, for validation one other simple way is to use the simple class name of the custom authorizer as prefix and not include any property that does not prefixed with that name. For Ex if custom authorizer is io.project.Foobar we can simply use FooBar as prefix and filter them when configuring it. This can be documented to avoid any confusion.

No, I do not think this would make sense ... you want this to work with any existing authorizers as well. Imposing a limitation like this might block them from working.

> having the configuration definitely made my implementation a lot simpler so recommend we keep it. otherwise, especially with the container environment it will hard to access a file/configmap etc for configuration. In my usecase, it kept me by not using kubeclient to go grab a known configmap in Authorizer implementation.

What I meant was this ... at the end the configuration fields anyway end up in the single properties file. So what I was thinking about was whether we really need to have the configuration section there => maybe the user can just add the options into the user configuration. That way, it will be automatically filter from forbidden options. The only thing I do not know for sure is what kind of validation we do there. 

I guess the other option would be ...
* to keep the configuration in the authorization `config` section and just filter it against the forbidden options (that way there is not duplicate detection against the other `config` section but we probably don't care)
* Or inject it probably already in the `fromCrd` method into the suer-provided configuration.

scholzj:
You cannot do the prefix thing as mentioned above. And it should IMHO not be unvalidated. The duplicates we IMHO care less about. So that leaves 2, 3 and 5 as otpions IMHO.

rareddy:
IMO between 2 & 3 the difference is where the configuration is defined in the CR. Since 3 is already in place, I vote for using that and remove the `config` section from the current PR.

rareddy:
@scholzj @mstruk any further guidance on this how-to proceed?

tombentley:
FWIW, my preference is that the authoriser's config should be put in `spec.kafka.config`, because if we supported other plugins in the future it would become confusing if the overall config consisted of:

1. The implicit stuff which the operator supplies
2. The explicit `spec.kafka.config`
3. Several _other_ places in the CR which might be sever screens of text away from `spec.kafka.config`.

There are further consequences for documentation if `spec.kafka.config` wasn't the single point for configuring the explicit configs.

scholzj:
Ok, lets go with that then. I checked that custom configs can be passed through `.spec.kafka.config` so it should be fine.

rareddy:
+1, I will remove the `config` section from this PR and update it.

rareddy:
Updated the PR by removing the `config` section from Authorizer and updated docs

documentation/api/io.strimzi.api.kafka.model.KafkaAuthorizationCustom.adoc
```diff
@@ -0,0 +1,131 @@
...
+authorization:
+  type: custom
+  authorizerClass: io.mycompany.CustomAuthorizer
+  superUsers:
+    - CN=client_1
+    - user_2
+    - CN=client_3
+config:
+  strimzi.authorization.custom.property1=value1
+  strimzi.authorization.custom.property2=value2
```
scholzj, line 35: I wonder if this makes it a bit easier to read.

```suggestion
spec:
  kafka:
    # ...
    authorization:
      type: custom
      authorizerClass: io.mycompany.CustomAuthorizer
      superUsers:
        - CN=client_1
        - user_2
        - CN=client_3
    # ...
    config:
      authorization.custom.property1=value1
      authorization.custom.property2=value2
    # ...
```

documentation/api/io.strimzi.api.kafka.model.KafkaAuthorizationCustom.adoc
```diff
@@ -0,0 +1,131 @@
...
+            return new UserSpec(principal.substring(0, pos), principal.substring(pos + 1));
+        }
+
+        @Override
+        public String toString() {
+            return super.toString() + " " + type + ":" + name;
+        }
+    }
+}
+----
```
scholzj, line 127: I not sure this example makes sense ... the docs would need to be checked and maintained when the Authorizer interface ever changes. Maybe we should leave it to Kafka.

install/cluster-operator/040-Crd-kafka.yaml
```diff
@@ -696,6 +696,10 @@ spec:
                           the Open Policy Agent, for example, when it is temporarily
                           unavailable). Defaults to `false` - all actions will be
                           denied.
+                      authorizerClass:
```
scholzj, line 699: The changes in this file need to go into `packaging/install/cluster-operator/...`

systemtest/src/test/java/io/strimzi/systemtest/security/custom/CustomAuthorizerST.java
```diff
@@ -0,0 +1,237 @@
...
+@Tag(REGRESSION)
+public class CustomAuthorizerST extends AbstractST {
+    static final String NAMESPACE = "custom-authorizer-test";
+    static final String CLUSTER_NAME = "custom-authorizer";
+    static final String ADMIN = "sre-admin";
+    private static final Logger LOGGER = LogManager.getLogger(CustomAuthorizerST.class);
+
+    @Test
+    @Tag(NODEPORT_SUPPORTED)
+    @Tag(EXTERNAL_CLIENTS_USED)
```
scholzj, line 48: @Frawless @im-konge I do not think these tests really need to use external clients, or? Can you help @rareddy use internal clients for this?

scholzj (commented):
The code looks good. Sorry for leading you into dead end before with the config. I left few comments for the docs and the test. Also, could you please rebase it so that we can run the tests?

install/cluster-operator/040-Crd-kafka.yaml
rareddy: I did not follow this suggestion

scholzj: The files in `install/...` should change only during release. The files which need to be changed in a PR like this are in `packaging/...` That is one of the reasons why this needs to be rebased. (more details about it are in https://github.com/strimzi/proposals/blob/master/020-rename-default-branch-of-strimzi-github-repositories.md)

documentation/api/io.strimzi.api.kafka.model.KafkaAuthorizationCustom.adoc
rareddy: Yes, agree. Will remove it

systemtest/src/test/java/io/strimzi/systemtest/security/custom/CustomAuthorizerST.java
Frawless: @rareddy  could you please check test `testSendMessagesTlsAuthenticated` how we use `InternalKafkaClient` ? I think that simply replace of `BasicExternalKafkaClient` exactly with the code form test I mentioned should be enough.

rareddy: Thanks, @Frawless & @scholzj  updated test to the internal client now.

documentation/api/io.strimzi.api.kafka.model.KafkaAuthorizationCustom.adoc
```diff
@@ -0,0 +1,45 @@
+To use custom authorization in Strimzi, you can configure your own `Authorizer` plugin to define Access Control Lists (ACLs).
```
scholzj, line 1: Can we also (somewhere here) add some explanation that users need to build their custom image and add the custom authorizer JARs to the classpath (`/opt/kafka/libs`)? Otherwise some might expect that it would be magically added just by using the API in the Kafka CR ;-).

scholzj (commented):
Two more nits, but LGTM otherwise.

scholzj:
There seem to be also some docs build issue which probably killed the build:

```
2021-03-19T16:19:45.8846160Z asciidoctor: WARNING: api/io.strimzi.api.kafka.model.KafkaAuthorizationCustom.adoc: line 17: id assigned to section already in use: property-simple-authorization-superusers-reference
```

I think that is about this part:

```
[id='property-simple-authorization-superusers-{context}']
```

which probably already exists in the simple authroizer docs.

documentation/api/io.strimzi.api.kafka.model.KafkaAuthorizationCustom.adoc
rareddy: how about adding the below text? @scholzj @PaulRMellor 

For Strimzi to instantiate and configure the custom Authorizer class, apart from the additions to the custom resource as shown above, the JAR file containing the custom Authorizer class along with its dependencies must be available on the classpath of Kafka broker.

Strimzi maven build process provides a mechanism to add these custom third-party libraries to the generated Kafka broker container image by adding them as dependencies in the `pom.xml` file under `docker-images/kafka/kafka-thirdparty-libs` directory. Under this directory there are different folders for different Kafka versions, choose the folder that fits your needs. Before `pom.xml` is modified the third-party library must be available in a Maven repository, and that maven repository is accessible to the Strimzi build process.

scholzj: That sounds good to me! 👍

scholzj (approved):
LGTM. Thanks for the work.

scholzj:
@PaulRMellor @ppatierno @Frawless Could you have another look at this?

Frawless (approved):
LGTM, good job :)

ppatierno (approved):

see-quick:
Hi, just a note. Recently, we merged a huge PR (https://github.com/strimzi/strimzi-kafka-operator/pull/4137), relating to `system test module`. Almost everything is changed. So, when you rebase you will have all changes which bind to our refactor. If you have some problems with re-writing the tests I am here to help.

scholzj:
@see-quick Maybe you can provide a patch?

see-quick:
After you rebase just apply this `git patch` file. https://gist.github.com/see-quick/e5a16b1fc375fb98c03284d606b2c33f

documentation/api/io.strimzi.api.kafka.model.KafkaAuthorizationCustom.adoc
```diff
@@ -0,0 +1,49 @@
...
+
+ACLs allow you to define which users have access to which resources at a granular level.
+
+Configure the `Kafka` custom resource to use custom authorization.
+Set the `type` property in the `authorization` section to the value `custom`,
+and the set following properties.
+
+IMPORTANT: The custom Authorizer must implement the org.apache.kafka.server.authorizer.Authorizer interface, and support configuration of super users using the super.users configuration property.
```
PaulRMellor, line 9:
```suggestion
IMPORTANT: The custom Authorizer must implement the `org.apache.kafka.server.authorizer.Authorizer` interface, and support configuration of super users using the super.users configuration property.
```

documentation/api/io.strimzi.api.kafka.model.KafkaAuthorizationCustom.adoc
```diff
@@ -0,0 +1,49 @@
...
+
+[id='property-custom-authorization-authorizerclass-{context}']
+=== `authorizerClass`
+(Required) Java class that implements the `org.apache.kafka.server.authorizer.Authorizer` interface to support the custom ACLs.
```
PaulRMellor, line 13:
```suggestion
(Required) Java class that implements the `org.apache.kafka.server.authorizer.Authorizer` interface to support custom ACLs.
```

documentation/api/io.strimzi.api.kafka.model.KafkaAuthorizationCustom.adoc
```diff
@@ -0,0 +1,49 @@
...
+
+
+[id='property-custom-authorization-superusers-{context}']
+=== `superUsers`
+A list of user principals treated as super users, so that they are always allowed without querying ACL rules.
+For more information see xref:con-securing-kafka-authorization-str[Kafka authorization].
+
+Configuration required for the custom authorizer can be provided using `Kafka.spec.kafka.config` and those properties will be supplied to custom authorizer in `config` method for initialization.
```
PaulRMellor, line 21:
```suggestion
You can add configuration for initializing the custom authorizer using `Kafka.spec.kafka.config`.
```

documentation/api/io.strimzi.api.kafka.model.KafkaAuthorizationCustom.adoc
```diff
@@ -0,0 +1,49 @@
...
+        - user_2
+        - CN=client_3
+    # ...
+    config:
+      authorization.custom.property1=value1
+      authorization.custom.property2=value2
+    # ...
+----
+
+For Strimzi to instantiate and configure the custom Authorizer class, apart from the additions to the custom resource as shown above, the JAR file containing the custom Authorizer class along with its dependencies must be available on the classpath of Kafka broker.
```
PaulRMellor, line 43:
```suggestion
In addition to the `Kafka` custom resource configuration, the JAR file containing the custom authorizer class along with its dependencies must be available on the classpath of the Kafka broker. 
```

documentation/api/io.strimzi.api.kafka.model.KafkaAuthorizationCustom.adoc
```diff
@@ -0,0 +1,49 @@
...
+
+Strimzi maven build process provides a mechanism to add these custom third-party libraries to the generated Kafka broker container image by adding them as dependencies in the pom.xml file under docker-images/kafka/kafka-thirdparty-libs directory. Under this directory there are different folders for different Kafka versions, choose the folder that fits your needs. Before pom.xml is modified the third-party library must be available in a Maven repository, and that maven repository is accessible to the Strimzi build process.
```
PaulRMellor, line 45:
```suggestion
The Strimzi Maven build process provides a mechanism to add custom third-party libraries to the generated Kafka broker container image by adding them as dependencies in the `pom.xml` file under the `docker-images/kafka/kafka-thirdparty-libs` directory. The directory contains different folders for different Kafka versions. Choose the appropriate folder. Before modifying the `pom.xml` file, the third-party library must be available in a Maven repository, and that Maven repository must be accessible to the Strimzi build process.
```

PaulRMellor (approved):
Few minor suggestions for the doc. Looks good.

scholzj:
Thanks for the PR and your patience with all the rebasing @rareddy ... you hit a bad time where lot of changes impacting almost everything were merged after we released 0.22.

PR #4553, opened 2021-03-11
Remove gRPC dependency.

alesj:
### Type of change

- Bugfix

### Description

The Topic Operator is designed to run always only in a single instance. It has no way to run in multiple replicas because it cannot partitions the KafkaTopic CRs or Kafka topics. We should therefore remove the distributed topic store from it since it is never used and removing it reduces the dependency footprint.


scholzj (commented):
The code looks good to me. Can you update the PR description?

sknot-rh (approved):

ppatierno:
Out of curiosity, why we are removing the gRPC based implemantation of the store?

scholzj:
@ppatierno Because it is not needed. The TO anyway doesn't know how to run in multiple instances for many other reasons then the Topic Store.

ppatierno (approved):

ppatierno:
I had doubts since the beginning why we had this, so was curious why removing now. Thanks.

scholzj (commented):
Ok, I updated the description my self.

PR #4573, opened 2021-03-15
Openssl refactor

tombentley:
### Type of change

- Refactoring

### Description

This is a refactoring and feature of the `OpenSslCertManager`. It does the following:

* Uses `openssl ca` for CA operations, which is a more powerful, but less friendly, tool. This is necessary to support creating certificates with defined end dates, rather than cert validity periods always being defined as `[now, now+days]`.
* Adds support for generating intermediate CA certificates.
* Generates the RSA keypair explicitly. There's currently no actual benefit to this, but it would allow us to support ECDSA certificates very easily should we decide that's worth doing.
* Removes some of the overload that were only used by tests (i.e. added no value).
* Adds more assertions to the tests.
* Cleans up how the list of `openssl` process options are constructed, using a helper class. There's now the ability to log (at trace level) the config files being used, which can be helpful when debugging. Keys and certs are not logged.
* Adds a `Builder` for `Subject`, because it had a super-awkward API. I've not refactored all the old call sites though.

### DIFF
--- a/certificate-manager/src/main/java/io/strimzi/certs/OpenSslCertManager.java
+++ b/certificate-manager/src/main/java/io/strimzi/certs/OpenSslCertManager.java
@@ -276,7 +276,7 @@ public class OpenSslCertManager implements CertManager {
                     .newCertsDir(newCertsDir)
                     .basicConstraints("critical,CA:true,pathlen:" + pathLength)
                     .keyUsage("critical,keyCertSign,cRLSign")
-                    .exec();
+                    .exec(false);
 
             // The key will be in pkcs#1 format (bracketed by BEGIN/END RSA PRIVATE KEY)
             // Convert it to pkcs#8 format (bracketed by BEGIN/END PRIVATE KEY)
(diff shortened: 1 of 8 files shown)


scholzj (approved):
LGTM. But it should be probably also reviewed by @ppatierno who is IIRC the original author of this code.

Also, did you checked if it causes any problems with existing CAs generated before this PR?

certificate-manager/src/main/java/io/strimzi/certs/OpenSslCertManager.java
```diff
@@ -216,139 +126,508 @@ private File buildConfigFile(Subject sbj, boolean isCa) throws IOException {
...
+        Path newCertsDir = null;
+        Path database = null;
 
-            cmd.addAll(asList("-subj", sbj.toString()));
+        try {
+            tmpKey = Files.createTempFile(null, null);
+            // Generate a key pair
+            new OpensslArgs("openssl", "genrsa")
+                    .optArg("-out", tmpKey)
+                    .opt("4096")
```
ppatierno, line 243: because it's used in a couple of places, maybe it would make sense to use a key lenght constant?

ppatierno (approved):
LGTM. Just a nit.

PR #4584, opened 2021-03-16
[systemtest] Remove ZK check from user quotas test

im-konge:

### Type of change

- Enhancement

### Description

When I was testing the #4221 I found that we don't really need to check user quotas in ZK as we already checking it via `bin/kafka-configs.sh`, so the "wait for" and checks around it aren't necessary. 

This PR removes ZK checks from `testUserWithQuotas` and add new one for output from `./bin/kafka-configs.sh` binary.

### Checklist


im-konge:
@strimzi-ci run tests test_only profile=regression testcase=UserST

Frawless (approved):

scholzj (approved):

ppatierno (approved):

Issue #2617, opened 2020-02-28
[Enhancement] .Support for multiple authentication mechanism for Kafka


arun-blacklane:
**Is your feature request related to a problem? Please describe.**
We want to use OAuth as the main authentication mechanism for Kafka using Strimzi.
But there are some clients which does-not support OAuth by default(Some of them are third party consumers for which we don't have direct support to modify clients)

**Describe the solution you'd like**
In kafka there is an option to support multiple authentication mechanism for brokers.
`sasl.enabled.mechanisms` option in server.properties
[(https://cwiki.apache.org/confluence/display/KAFKA/KIP-43%3A+Kafka+SASL+enhancements#KIP-43:KafkaSASLenhancements-Supportformultiplemechanismsinabroker)]
Was looking at Strimzi documentation and seems like we don't have a way to support this.
Is this currently supported in strimzi?

**Describe alternatives you've considered**
For the existing clients that we have control over, we can make changes in client side to use OAuth authentication. 
But we are stuck on making it available with clients which doesn't support a specific authentication mechanism.(We want to have a restricted access to these clients based a different authentication mechanism)

scholzj:
As answered in #2616, this is currently not supported. We plan to add support for more listeners - would being able to define one listener with SCRAM and one with OAUth work for you? Supporting multiple authentication types is a bit more complicated because you can in theory combine OAuth and SCRAM, but not OAuth and TLS. So it is more complicated to handle.

arun-blacklane:
@scholzj 
Thanks for the response.
I think first option should work for us. We can have a dedicated listener for the clients that don't use OAuth.
Is there a timeline on when this is to be implemented with Strimzi?

scholzj:
I would love to work on it in one of the next releases. So lets sea before the summer. But as everything - that is just tentative.

oulydna:
Hey guys, do we have a more clear idea when this may be available?

scholzj:
I think this is slowly moving forward. There is now an approved proposal for this change: https://github.com/strimzi/proposals/blob/master/005-improving-configurability-of-kafka-listeners.md ... I hope to get to the implementation during the summer.

oulydna:
Awesome!

scholzj:
Triaged on 17.3.2022: Multiple authentications can be used with multiple listeners. So this should be done.

PR #4605, opened 2021-03-19
[DOC] Add extra description of the SASL_PLAIN authentication mechanism

laidan6000:
### Type of change

- Documentation

Guide(s) updated: Using Strimzi

### Description

This pull request adds a description of the SASL_PLAIN authentication mechanism to the existing module named "OAuth 2.0 authentication mechanisms". At the moment, this topic only mention SASL_OAUTHBEARER.

It also updates the "Session re-authentication for Kafka brokers" module with new information for SASL_PLAIN. I tried to improve the structure of the content as well.

The words are basically the same as used in the procedure for broker configuration. 

It also fixes some "Module included in" comments in multiple OAuth 2 modules.

### Checklist

- [ ] Write tests
- [X] Update documentation
- [X] Reference relevant issue(s) and close them after merging - RELATED TO ISSUE #2617
- [ ] Update CHANGELOG.md


documentation/modules/oauth/con-oauth-authentication-flow.adoc
```diff
@@ -1,12 +1,42 @@
...
-The Kafka _SASL OAUTHBEARER_ mechanism is used to establish authenticated sessions with a Kafka broker.
+You can configure OAuth 2.0 authentication using one of the following authentication mechanisms:
 
-A Kafka client initiates a session with the Kafka broker using the _SASL OAUTHBEARER_ mechanism for credentials exchange, where credentials take the form of an access token.
+* SASL_OAUTHBEARER
+* SASL_PLAIN
 
-Kafka brokers and clients need to be configured to use OAuth 2.0.
+Both mechanisms allow Kafka clients to establish authenticated sessions with Kafka brokers. 
+However, the authentication flow between a client, the authorization server, and the Kafka broker is different for each mechanism.
```
scholzj, line 14: I think it is really important to explain the differences and recommendations. @mstruk can provide more details, but AFAIU:
* Whenever possible, OAUTHBEARER should be used
* Only when the user has clients not supporting OAUTHBEARER, they should consider using PLAIN.

documentation/modules/oauth/con-oauth-authentication-flow.adoc
```diff
@@ -1,12 +1,42 @@
...
+SASL_PLAIN allows client authentication using either of the following methods:
+
+* OAuth access token
+* OAuth `clientId` and `secret` (the client credentials)
+
+The client must enable SASL_PLAIN and set the `username` and `password`. 
+If the client sets `username` to `access-token`, the authorization server will interpret the password as the access token. 
+Otherwise, the authorization server will interpret the user as the `clientId` and the password as the client secret.
+
+Kafka brokers and clients both need to be configured to use OAuth 2.0. SASL_PLAIN must be enabled in the listener configuration for the Kafka broker.
```
scholzj, line 38: I think this makes it more clear that this is just a flag in the OAuth configuration. I think we should also have some example here maybe?

```suggestion
Kafka brokers and clients both need to be configured to use OAuth 2.0. SASL_PLAIN must be enabled in the listener OAuth 2.0 authentication configuration for the Kafka broker.
```

documentation/modules/oauth/con-oauth-authentication-flow.adoc
laidan6000: That would be great if @mstruk can add those guidelines.

documentation/modules/oauth/con-oauth-authentication-flow.adoc
laidan6000: I can add a short example YAML. "Configuring OAuth 2.0 support for Kafka brokers" already has three examples.

documentation/modules/oauth/con-oauth-authentication-flow.adoc
```diff
@@ -35,7 +45,18 @@ The client must enable SASL_PLAIN and set the `username` and `password`.
...
+Kafka brokers and clients both need to be configured to use OAuth 2.0. SASL_PLAIN must be enabled in the listener OAuth 2.0 authentication configuration for the Kafka broker.
+
+[source,yaml,subs="+quotes,attributes+"]
+----
+  # ...
+  authentication:
+    type: oauth
+    # ...
+    enableOauthBearer: false
+    enablePlain: true
```
scholzj, line 57: Will it be clear tot he users that they can also have both set to true?

documentation/modules/oauth/con-oauth-authentication-flow.adoc
```diff
@@ -1,12 +1,63 @@
...
 
-Kafka brokers and clients need to be configured to use OAuth 2.0.
+Both mechanisms allow Kafka clients to establish authenticated sessions with Kafka brokers. 
+However, the authentication flow between a client, the authorization server, and the Kafka broker is different for each mechanism.
+
+.SASL_OAUTHBEARER
+
+Support for the SASL_OAUTHBEARER mechanism is built in to many Kafka clients.
+
+The client initiates a session with the Kafka broker using the SASL OAUTHBEARER mechanism for credentials exchange, where credentials take the form of a JWT access token.
```
mstruk, line 20:
```suggestion
The client initiates a session with the Kafka broker using the SASL OAUTHBEARER mechanism for credentials exchange, where credentials take the form of a bearer token provided by the callback code. Using Strimzi OAuth callbacks one can configure three ways to provide the token:
* Client ID and Secret (by using OAuth2 client credentials mechanism)
* A long-lived access token, obtained manually by the user at configuration time
* A long-lived refresh token, obtained manually by the user at configuration time
```

documentation/modules/oauth/con-oauth-authentication-flow.adoc
```diff
@@ -1,12 +1,63 @@
...
+    # ...
+    enableOauthBearer: true
+    enablePlain: false
+----
+
+.SASL_PLAIN
+
+SASL_PLAIN is a simple authentication mechanism that is supported by all Kafka client tools, including developer tools such as kafkacat.
+
+SASL_PLAIN allows client authentication using either of the following methods:
```
mstruk, line 39:
```suggestion
SASL_PLAIN with the Strimzi provided callback is what we also call OAuth over PLAIN, and allows a client to authenticate using either of the following methods:
```

documentation/modules/oauth/con-oauth-authentication-flow.adoc
```diff
@@ -1,12 +1,63 @@
...
+
+* OAuth access token
```
mstruk, line 41:
```suggestion
```

documentation/modules/oauth/con-oauth-authentication-flow.adoc
```diff
@@ -1,12 +1,63 @@
...
+* OAuth `clientId` and `secret` (the client credentials)
```
mstruk, line 42:
```suggestion
* OAuth `clientId` and `secret` (the client credentials)
* OAuth access token
```

documentation/modules/oauth/con-oauth-authentication-flow.adoc
```diff
@@ -1,12 +1,63 @@
...
+
+The client must enable SASL_PLAIN and set the `username` and `password`. 
+If the client sets `username` to `access-token`, the authorization server will interpret the password as the access token.
```
mstruk, line 45:
```suggestion
If the client prefixes the `password` with `$accessToken:` followed by the value of the access token (access token), the authorization server will interpret the password as the access token.
```

documentation/modules/oauth/con-oauth-authentication-flow.adoc
```diff
@@ -1,12 +1,63 @@
...
+SASL_PLAIN is a simple authentication mechanism that is supported by all Kafka client tools, including developer tools such as kafkacat.
+
+SASL_PLAIN allows client authentication using either of the following methods:
+
+* OAuth access token
+* OAuth `clientId` and `secret` (the client credentials)
+
+The client must enable SASL_PLAIN and set the `username` and `password`. 
+If the client sets `username` to `access-token`, the authorization server will interpret the password as the access token. 
+Otherwise, the authorization server will interpret the user as the `clientId` and the password as the client secret.
```
mstruk, line 46:
```suggestion
Otherwise, the authorization server will interpret the `username` as the client ID and the `password` as the client secret.

If the `password` is set as the access token, the `username` has to be set to the same principal name the Kafka broker will obtain from the access token. The specifics depend on how you configure username extraction (using `userNameClaim`, `fallbackUserNameClaim`, `fallbackUsernamePrefix`, `userInfoEndpointUri`), but also on your authorization server - how it maps client Ids to account names.
```

documentation/modules/oauth/con-oauth-authentication-flow.adoc
```diff
@@ -1,12 +1,63 @@
...
+However, the authentication flow between a client, the authorization server, and the Kafka broker is different for each mechanism.
+
+.SASL_OAUTHBEARER
+
+Support for the SASL_OAUTHBEARER mechanism is built in to many Kafka clients.
+
+The client initiates a session with the Kafka broker using the SASL OAUTHBEARER mechanism for credentials exchange, where credentials take the form of a JWT access token.
+
+Kafka brokers and clients both need to be configured to use OAuth 2.0. 
+SASL_OAUTHBEARER must be enabled in the listener configuration for the Kafka broker.
```
mstruk, line 23:
```suggestion
SASL_OAUTHBEARER is automatically enabled in the listener configuration for the Kafka broker, so it is not necessary to explicitly enable it.
```

documentation/modules/oauth/con-oauth-authentication-flow.adoc
```diff
@@ -1,12 +1,63 @@
...
+
+[source,yaml,subs="attributes+"]
+----
+  # ...
+  authentication:
+    type: oauth
+    # ...
+    enableOauthBearer: true
+    enablePlain: false
```
mstruk, line 32:
```suggestion
```

documentation/modules/oauth/con-oauth-authentication-flow.adoc
```diff
@@ -1,12 +1,63 @@
...
+
+Kafka brokers and clients both need to be configured to use OAuth 2.0. SASL_PLAIN must be enabled in the listener OAuth 2.0 authentication configuration for the Kafka broker.
+
+[source,yaml,subs="+quotes,attributes+"]
+----
+  # ...
+  authentication:
+    type: oauth
+    # ...
+    enableOauthBearer: false
```
mstruk, line 56:
```suggestion
```

documentation/modules/oauth/con-oauth-authentication-flow.adoc
```diff
@@ -1,12 +1,63 @@
...
+SASL_PLAIN allows client authentication using either of the following methods:
+
+* OAuth access token
+* OAuth `clientId` and `secret` (the client credentials)
+
+The client must enable SASL_PLAIN and set the `username` and `password`. 
+If the client sets `username` to `access-token`, the authorization server will interpret the password as the access token. 
+Otherwise, the authorization server will interpret the user as the `clientId` and the password as the client secret.
+
+Kafka brokers and clients both need to be configured to use OAuth 2.0. SASL_PLAIN must be enabled in the listener OAuth 2.0 authentication configuration for the Kafka broker.
```
mstruk, line 48:
```suggestion
Kafka brokers and clients both need to be configured to use OAuth 2.0. SASL_PLAIN must be enabled in the listener OAuth 2.0 authentication configuration for the Kafka broker.

The following enables SASL_PLAIN in addition to SASL_OAUTHBEARER which is enabled by default. If you want only SASL_PLAIN you can disable SASL_OAUTHBEARER by setting `enableOauthBearer` to `false`.
```

documentation/modules/oauth/con-oauth-reauthentication.adoc
```diff
@@ -1,11 +1,12 @@
...
 //
-// assembly-oauth.adoc
+// assembly-oauth-authentication.adoc
 
 [id='con-oauth-reauthentication-{context}']
 = Session re-authentication for Kafka brokers
 
-The Kafka _SASL OAUTHBEARER_ mechanism, which is used for OAuth 2.0 authentication in Strimzi, supports a Kafka feature called the _re-authentication_ mechanism.
+The _SASL OAUTHBEARER_ authentication mechanism supports a Kafka feature called the _session re-authentication_ mechanism. 
+Session re-authentication is not supported if _SASL_PLAIN_ authentication is used.
```
mstruk, line 9: There is some nuance to this. And maybe we could better explain it. Re-authentication happens in both cases. `maxSecondsWithoutReauthentication` setting sets the maximum session timeout regardless of the mechanism.
The difference between OAUTHBEARER behavior and PLAIN behavior is that for OAUTHBEARER the token expires after some time, and that time takes precedence if shorter than the configured `maxSecondsWithoutReauthentication`, whereas for PLAIN the re-authentication is forced at the configured time. The same `maxSecondsWithoutReauthentication` config is used for both OAUTHBEARER and PLAIN on the same listener. If `maxSecondsWithoutReauthentication` is not set then PLAIN clients can stay connected indefinitely without ever having to re-authenticate.

mstruk (changes requested):

documentation/modules/oauth/con-oauth-authentication-flow.adoc
PaulRMellor: we have a warning about using SASL plain in `KafkaClientAuthenticationPlain` schema reference

PaulRMellor (approved):

laidan6000:
Thanks for all the changes, @mstruk . This is now ready for a second review.

documentation/assemblies/oauth/assembly-oauth-authentication.adoc
```diff
@@ -5,15 +5,19 @@
 [id='assembly-oauth-authentication_{context}']
 = Using OAuth 2.0 token-based authentication
 
-Strimzi supports the use of OAuth 2.0 authentication using the _SASL OAUTHBEARER_ mechanism.
+Strimzi supports the use of OAuth 2.0 authentication using the _SASL_OAUTHBEARER_ and _SASL_PLAIN_ mechanisms.
```
scholzj, line 8: Why do we call it _SASL_OAUTHBEARER_ and _SASL_PLAIN_ I would everywhere use just _SASL OAUTHBEARER_ and _SASL PLAIN_. Is the form with the underscore used in any SASL, JaaS or Kafka documentation?

documentation/assemblies/oauth/assembly-oauth-authentication.adoc
```diff
@@ -5,15 +5,19 @@
...
 
 OAuth 2.0 enables standardized token-based authentication and authorization between applications, using a central authorization server to issue tokens that grant limited access to resources.
 
+Kafka brokers and clients both need to be configured to use OAuth 2.0. 
 You can configure OAuth 2.0 authentication, then xref:assembly-oauth-authorization_{context}[OAuth 2.0 authorization].
 
-OAuth 2.0 authentication can also be used in conjunction with `simple` or OPA-based xref:con-securing-kafka-authorization-str[Kafka authorization].
+[NOTE]
+====
+OAuth 2.0 authentication can also be used in conjunction with `simple` or Open Policy Agent (OPA) based xref:con-securing-kafka-authorization-str[Kafka authorization].
```
scholzj, line 17: It is a bit weird to use only the `type` name for one and full name for the other. Would this be better?

```suggestion
OAuth 2.0 authentication can also be used in conjunction with Simple or Open Policy Agent (OPA) based xref:con-securing-kafka-authorization-str[Kafka authorization].
```

documentation/modules/oauth/con-oauth-authentication-flow.adoc
```diff
@@ -1,12 +1,79 @@
...
+.SASL_PLAIN overview
+
+SASL_PLAIN, also referred to as _OAuth 2.0 over PLAIN_, is a simple authentication mechanism supported by all Kafka client tools (including developer tools such as kafkacat). 
+Client credentials are handled centrally, behind an authorization server that complies with OAuth 2.0 and OpenID Connect (OIDC).
+
+When used with the provided callbacks, SASL_PLAIN allows Kafka clients to authenticate with Kafka brokers using either of the following methods:
+
+* OAuth access token
+
+* OAuth `clientId` and `secret` (the client credentials)
```
scholzj, line 53: Should we use the same terms as in the list on lines 24-30?

```suggestion
* Client ID and Secret (by using the OAuth 2.0 client credentials mechanism)

* A long-lived access token, obtained manually at configuration time
```

documentation/modules/oauth/con-oauth-authentication-flow.adoc
```diff
@@ -1,12 +1,79 @@
...
+
+The client must enable SASL_PLAIN and set the `username` and `password`. 
+If the client prefixes the `password` with `$accessToken:` followed by the value of the access token (access token ???), the authorization server will interpret the password as the access token.
```
scholzj, line 56: I guess the `(access token ???)`  should not be here?

documentation/modules/oauth/con-oauth-reauthentication.adoc
```diff
@@ -1,21 +1,40 @@
...
+The duration of the access token always take precedence over `maxSecondsWithoutReauthentication`.
+
+.Session expiry with PLAIN
+
+When session re-authentication is configured with SASL_PLAIN, the broker's authenticated session will expire at the configured `maxSecondsWithoutReauthentication`. 
+If `maxSecondsWithoutReauthentication` is not configured, SASL_PLAIN clients can remain connected to brokers indefinitely, without needing to re-authenticate.
+
+.Client re-authentication
+
+When the broker's authenticated session expires, the client must re-authenticate to the existing session by sending a new, valid access token to the broker, without dropping the connection.
```
scholzj, line 28: This section seems a bit disconnected. Does it relate to _Session expiry with PLAIN_? Or both _Session expiry with PLAIN_ and _Session expiry with OAUTHBEARER_?

laidan6000: I think this part applies to both OAUTHBEARER and PLAIN.

 @mstruk - Please can you read the updated doc for Session re-authentication and let me know if anything looks out of place.

Do we need more information about _client re-authentication_ when using PLAIN?

documentation/assemblies/oauth/assembly-oauth-authentication.adoc
laidan6000: Underscores are used in the Kafka documentation, for example, https://kafka.apache.org/documentation/#security_sasl_config

@mstruk and @PaulRMellor  - Do you have any objections to me removing the underscores?

documentation/modules/oauth/con-oauth-authentication-flow.adoc
laidan6000: Sorry, that was my mistake.

documentation/assemblies/oauth/assembly-oauth-authentication.adoc
scholzj: `SASL_PLAINTEXT` or `SASL_SSL` used there are Kafka's security protocols which (while absolutely confusing) are something different than SASL mechanism you talk about here. I do not think the Kafka doc is using `SASL_PLAIN` or `SASL_OAUTHBEARER`.

documentation/modules/oauth/con-oauth-authentication-flow.adoc
scholzj: No worries ... that is what reviews are for ;-)

documentation/modules/oauth/con-oauth-authentication-flow.adoc
laidan6000: Probably not...

I can add something like this:

SASL_OAUTHBEARER and SASL_PLAIN can be enabled at the same time to support both types of clients.

documentation/modules/oauth/con-oauth-reauthentication.adoc
laidan6000: Thanks for the explanation! I re-worked this module based on your text - please can you check it carefully.

documentation/assemblies/oauth/assembly-oauth-authentication.adoc
mstruk: Checking the Kafka Docs, they mostly use just OAUTHBEARER, sometimes SASL/OAUTHBEARER, and only here and there SASL OAUTHBEARER. In light of that we may simply use OAUTHBEARER, and PLAIN.

documentation/modules/oauth/con-oauth-reauthentication.adoc
mstruk: Re-authentication is related to the enforcement of session expiry - it allows a new session authentication from the client, effectively the start of a new session by reusing existing connection rather than dropping it. The mechanics is really in the Kafka Client library, not something the client has to explicitly take care of in the code. I see this chapter more as an explanation what's going on in the background for user to better understand the mechanics.

documentation/assemblies/oauth/assembly-oauth-authentication.adoc
laidan6000: Great. I'm all for making this simpler.

documentation/modules/oauth/con-oauth-reauthentication.adoc
laidan6000: Great info, thanks. I updated the module again based on this text.

documentation/assemblies/oauth/assembly-oauth-authentication.adoc
laidan6000: Changed to OAUTHBEARER and PLAIN throughout.

I was careful to NOT change any occurences of `SASL_SSL` and `SASL_PLAINTEXT`.

documentation/modules/oauth/con-oauth-authentication-flow.adoc
laidan6000: Added:

> OAUTHBEARER and PLAIN can be enabled at the same time, on the same `oauth` listener, if both types of OAuth clients need to authenticate.

scholzj (approved):

scholzj:
@laidan6000 I approved it, the content seems ok. But it looks like the DCO is broken. Can you have a look at it?

documentation/modules/oauth/con-oauth-authentication-flow.adoc
```diff
@@ -1,12 +1,81 @@
...
+  # ...
+  authentication:
+    type: oauth
+    # ...
+    enableOauthBearer: true
+----
+
+.PLAIN overview
+
+PLAIN, also referred to as _OAuth 2.0 over PLAIN_, is a simple authentication mechanism supported by all Kafka client tools (including developer tools such as kafkacat).
```
mstruk, line 48:
```suggestion
PLAIN is a simple authentication mechanism supported by all Kafka client tools (including developer tools such as kafkacat). In order to use it together with OAuth authentication we install additional server-side callbacks and call this _OAuth 2.0 over PLAIN_.
```

documentation/modules/oauth/con-oauth-authentication-flow.adoc
```diff
@@ -1,12 +1,81 @@
...
+Client credentials are handled centrally, behind an authorization server that complies with OAuth 2.0 and OpenID Connect (OIDC).
+
+When used with the provided callbacks, PLAIN allows Kafka clients to authenticate with Kafka brokers using either of the following methods:
+
+* Client ID and Secret (by using the OAuth 2.0 client credentials mechanism)
+
+* A long-lived access token, obtained manually at configuration time
+
+The client must enable PLAIN and set the `username` and `password`. 
+If the client prefixes the `password` with `$accessToken:` followed by the value of the access token, the authorization server will interpret the password as the access token.
```
mstruk, line 58:
```suggestion
If the client prefixes the `password` with `$accessToken:` followed by the value of the access token, the Kafka broker will interpret the password as the access token. 
```

documentation/modules/oauth/con-oauth-authentication-flow.adoc
```diff
@@ -1,12 +1,81 @@
...
+
+When used with the provided callbacks, PLAIN allows Kafka clients to authenticate with Kafka brokers using either of the following methods:
+
+* Client ID and Secret (by using the OAuth 2.0 client credentials mechanism)
+
+* A long-lived access token, obtained manually at configuration time
+
+The client must enable PLAIN and set the `username` and `password`. 
+If the client prefixes the `password` with `$accessToken:` followed by the value of the access token, the authorization server will interpret the password as the access token. 
+Otherwise, the authorization server will interpret the `username` as the client ID and the `password` as the client secret.
```
mstruk, line 59:
```suggestion
Otherwise, the Kafka broker will interpret the `username` as the client ID and the `password` as the client secret.
```

documentation/modules/oauth/con-oauth-reauthentication.adoc
```diff
@@ -1,23 +1,39 @@
...
-You enable session re-authentication in the `Kafka` resource. Set the `maxSecondsWithoutReauthentication` property for a TLS listener with `type: oauth` authentication.
-Session re-authentication is supported for both types of token validation (fast local JWT and introspection endpoint). For an example configuration, see xref:proc-oauth-authentication-broker-config-{context}[].
+.Differences in session expiry for OAUTHBEARER and PLAIN
 
-For more information about the re-authentication mechanism, which was added in Kafka version 2.2, see link:https://cwiki.apache.org/confluence/display/KAFKA/KIP-368%3A+Allow+SASL+Connections+to+Periodically+Re-Authenticate[KIP-368^].
+There are some differences in how session expiry works when session re-authentication is configured for OAUTHBEARER and PLAIN authentication.
+
+* For OAUTHBEARER, the broker's authenticated session will expire at the configured `maxSecondsWithoutReauthentication`. 
+The session will expire earlier if the access token expires before the `maxSecondsWithoutReauthentication`. 
+The duration of the access token always take precedence over `maxSecondsWithoutReauthentication`.
```
mstruk, line 33: This last statement seems to me to sow confusion.
First we say session will expire at the configured time unless the token expires beforehand in which case the session will expire earlier.
Then we say token expiry overrides the configured time - meaning if token lifetime is longer it also overrides the configured time.

If `maxSecondsWithoutReauthentication` is not configured, the session will not expire with the expiry of the access token. This is the same as for PLAIN - without `maxSecondsWithoutReauthentication` the sessions just don't expire by themselves. 

```suggestion
```

documentation/modules/oauth/con-oauth-reauthentication.adoc
```diff
@@ -1,23 +1,39 @@
...
+
+* For PLAIN, the broker's authenticated session will expire at the configured `maxSecondsWithoutReauthentication`. 
+If `maxSecondsWithoutReauthentication` is not configured, PLAIN clients can remain connected to brokers indefinitely, without needing to re-authenticate.
```
mstruk, line 36:
```suggestion

If `maxSecondsWithoutReauthentication` is not configured, OAUTHBEARER and PLAIN clients can remain connected to brokers indefinitely, without needing to re-authenticate. In that case the sessions don't end with access token expiry, but access token expiry can be taken into account when configuring proper authorization. For example by using `keycloak` authorization or installing a custom authorizer.
```

documentation/modules/oauth/proc-oauth-authentication-broker-config.adoc
```diff
@@ -133,8 +133,8 @@ Depending on how you apply OAuth 2.0 authentication, and the type of authorizati
...
+<7> Set this to `false` in order to disable the OAUTHBEARER mechanism on the listener. At least one of PLAIN or OAUTHBEARER has to be enabled. Default is `true`.
+<8> Set this to `true` in order to enable the PLAIN mechanism on the listener, which is supported by all clients on all platforms. The Kafka client has to enable the PLAIN mechanism and set the `username` and the `password`. This mechanism can be used to authenticate either by using the OAuth access token, or by using the OAuth clientId and secret (client credentials). If the client sets `username` to `access-token` then the password is interpreted as access token on the server, otherwise the user is interpreted as the clientId, and password as the client secret. Default is `false`.
```
mstruk, line 137:
```suggestion
<8> Set this to `true` in order to enable the PLAIN mechanism on the listener, which is supported by all clients on all platforms. The Kafka client has to enable the PLAIN mechanism and set the `username` and the `password`. This mechanism can be used to authenticate either by using the OAuth access token, or by using the OAuth clientId and secret (client credentials). If the client sets `password` to start with a string `$accessToken:` then the password is interpreted as access token on the server, and `username` as the account username, otherwise the user is interpreted as the clientId, and password as the client secret. Default is `false`.
```

mstruk (changes requested):

documentation/assemblies/oauth/assembly-oauth-authentication.adoc
```diff
@@ -5,15 +5,19 @@
...
 
 OAuth 2.0 enables standardized token-based authentication and authorization between applications, using a central authorization server to issue tokens that grant limited access to resources.
 
+Kafka brokers and clients both need to be configured to use OAuth 2.0. 
 You can configure OAuth 2.0 authentication, then xref:assembly-oauth-authorization_{context}[OAuth 2.0 authorization].
 
-OAuth 2.0 authentication can also be used in conjunction with `simple` or OPA-based xref:con-securing-kafka-authorization-str[Kafka authorization].
+[NOTE]
+====
+OAuth 2.0 authentication can also be used in conjunction with Simple or Open Policy Agent (OPA) based xref:con-securing-kafka-authorization-str[Kafka authorization].
```
PaulRMellor, line 17: What about the new `custom` option?

documentation/modules/oauth/con-oauth-authentication-flow.adoc
```diff
@@ -1,12 +1,83 @@
...
+  # ...
+  authentication:
+    type: oauth
+    # ...
+    enableOauthBearer: true
+----
+
+.PLAIN overview
+
+PLAIN is a simple authentication mechanism supported by all Kafka client tools (including developer tools such as kafkacat).
```
PaulRMellor, line 48: We say earlier that _Support for OAUTHBEARER is built in to many clients_ and here we say PLAIN is supported by all _client tools_. Does this mean we would normally have to configure OAUTHBEARER, both or neither for clients and client tools?

documentation/modules/oauth/con-oauth-authentication-flow.adoc
```diff
@@ -1,12 +1,83 @@
...
+  authentication:
+    type: oauth
+    # ...
+    enableOauthBearer: true
+----
+
+.PLAIN overview
+
+PLAIN is a simple authentication mechanism supported by all Kafka client tools (including developer tools such as kafkacat). 
+To enable PLAIN to be used together with OAuth authentication, Strimzi includes additional server-side callbacks and calls this _OAuth 2.0 over PLAIN_.
```
PaulRMellor, line 49:
```suggestion
To enable PLAIN to be used together with OAuth authentication, Strimzi includes server-side callbacks and calls this _OAuth 2.0 over PLAIN_. 
```

documentation/modules/oauth/con-oauth-authentication-flow.adoc
```diff
@@ -1,12 +1,83 @@
...
+    # ...
+    enableOauthBearer: true
+----
+
+.PLAIN overview
+
+PLAIN is a simple authentication mechanism supported by all Kafka client tools (including developer tools such as kafkacat). 
+To enable PLAIN to be used together with OAuth authentication, Strimzi includes additional server-side callbacks and calls this _OAuth 2.0 over PLAIN_. 
+
+With PLAIN, Client credentials are handled centrally, behind an authorization server that complies with OAuth 2.0 and OpenID Connect (OIDC).
```
PaulRMellor, line 51:
```suggestion
With PLAIN, client credentials are handled centrally, behind an authorization server that complies with OAuth 2.0 and OpenID Connect (OIDC).
```

PaulRMellor: Do we need to say you will need a compliant authorzation server?

documentation/modules/oauth/con-oauth-authentication-flow.adoc
```diff
@@ -1,12 +1,83 @@
...
+
+When used with the provided callbacks, PLAIN allows Kafka clients to authenticate with Kafka brokers using either of the following methods:
```
PaulRMellor, line 53:
```suggestion
When used with the OAuth 2.0 over PLAIN callbacks, Kafka clients authenticate with Kafka brokers using either of the following methods:
```

documentation/modules/oauth/con-oauth-authentication-flow.adoc
```diff
@@ -1,12 +1,83 @@
...
+
+* Client ID and Secret (by using the OAuth 2.0 client credentials mechanism)
+
+* A long-lived access token, obtained manually at configuration time
+
+The client must enable PLAIN and set the `username` and `password`.
```
PaulRMellor, line 59:
```suggestion
The client must be enabled to use PLAIN authentication with a `username` and `password`. 
```

documentation/modules/oauth/con-oauth-authentication-flow.adoc
```diff
@@ -1,12 +1,83 @@
...
+.PLAIN overview
+
+PLAIN is a simple authentication mechanism supported by all Kafka client tools (including developer tools such as kafkacat). 
+To enable PLAIN to be used together with OAuth authentication, Strimzi includes additional server-side callbacks and calls this _OAuth 2.0 over PLAIN_. 
+
+With PLAIN, Client credentials are handled centrally, behind an authorization server that complies with OAuth 2.0 and OpenID Connect (OIDC).
+
+When used with the provided callbacks, PLAIN allows Kafka clients to authenticate with Kafka brokers using either of the following methods:
+
+* Client ID and Secret (by using the OAuth 2.0 client credentials mechanism)
```
PaulRMellor, line 55: We're using lower case secret elsewhere.

documentation/modules/oauth/con-oauth-authentication-flow.adoc
```diff
@@ -1,12 +1,83 @@
...
+
+* A long-lived access token, obtained manually at configuration time
+
+The client must enable PLAIN and set the `username` and `password`. 
+If the client prefixes the `password` with `$accessToken:` followed by the value of the access token, the Kafka broker will interpret the password as the access token.
```
PaulRMellor, line 60:
```suggestion
If the password is prefixed with `$accessToken:` followed by the value of the access token, the Kafka broker will interpret the password as the access token. 
```

documentation/modules/oauth/con-oauth-authentication-flow.adoc
```diff
@@ -1,12 +1,83 @@
...
+* Client ID and Secret (by using the OAuth 2.0 client credentials mechanism)
+
+* A long-lived access token, obtained manually at configuration time
+
+The client must enable PLAIN and set the `username` and `password`. 
+If the client prefixes the `password` with `$accessToken:` followed by the value of the access token, the Kafka broker will interpret the password as the access token. 
+Otherwise, the Kafka broker will interpret the `username` as the client ID and the `password` as the client secret.
+
+If the `password` is set as the access token, the `username` must be set to the same principal name that the Kafka broker obtains from the access token. 
+The process depends on how you configure username extraction (using `userNameClaim`, `fallbackUserNameClaim`, `fallbackUsernamePrefix`, and `userInfoEndpointUri`), and also your authorization server (in particular, how it maps client IDs to account names).
```
PaulRMellor, line 64:
```suggestion
The process depends on how you configure username extraction (using `userNameClaim`, `fallbackUserNameClaim`, `fallbackUsernamePrefix`, or `userInfoEndpointUri`), and also your authorization server (in particular, how it maps client IDs to account names).
```

PaulRMellor: This is a long sentence. Maybe it could be split and the brackets brought into main sentences.

documentation/modules/oauth/con-oauth-reauthentication.adoc
```diff
@@ -1,23 +1,41 @@
...
 //
-// assembly-oauth.adoc
+// assembly-oauth-authentication.adoc
 
-[id='con-oauth-reauthentication-{context}']
+[id='{context}']
 = Session re-authentication for Kafka brokers
 
-The Kafka _SASL OAUTHBEARER_ mechanism, which is used for OAuth 2.0 authentication in Strimzi, supports a Kafka feature called the _re-authentication_ mechanism.
+OAuth 2.0 authentication in Strimzi supports _session re-authentication_, which is part of the Kafka Client libraries.
```
PaulRMellor, line 8: Not sure. Is it this: Strimzi's Kafka Client libraries support session re-athentication for OAuth 2.0 authentication?

documentation/modules/oauth/con-oauth-authentication-flow.adoc
PaulRMellor: Do we provide example clients configured for PLAIN and OAUTHBEARER?

documentation/modules/oauth/con-oauth-reauthentication.adoc
```diff
@@ -1,23 +1,41 @@
...
-// assembly-oauth.adoc
+// assembly-oauth-authentication.adoc
 
-[id='con-oauth-reauthentication-{context}']
+[id='{context}']
 = Session re-authentication for Kafka brokers
 
-The Kafka _SASL OAUTHBEARER_ mechanism, which is used for OAuth 2.0 authentication in Strimzi, supports a Kafka feature called the _re-authentication_ mechanism.
+OAuth 2.0 authentication in Strimzi supports _session re-authentication_, which is part of the Kafka Client libraries. 
+This Kafka mechanism is used to enforce the expiry of authenticated sessions between OAuth clients and Kafka brokers after a defined period of time.
```
PaulRMellor, line 9:
```suggestion
Session re-authentication enforces the expiry of authenticated sessions between OAuth clients and Kafka brokers after a defined period of time. 
```

documentation/modules/oauth/con-oauth-reauthentication.adoc
```diff
@@ -1,23 +1,41 @@
...
+When a session expires, the client immediately starts a new session by reusing the existing connection rather than dropping it.
 
-When the re-authentication mechanism is enabled in the configuration of an `oauth` type listener, the broker's authenticated session expires when the access token expires.
-The client must then re-authenticate to the existing session by sending a new, valid access token to the broker, without dropping the connection.
+Session re-authentication is disabled by default. 
+To enable and configure it, you set a time value for `maxSecondsWithoutReauthentication` in the `oauth` listener configuration in the `Kafka` custom resource. 
+The same property is used to configure session re-authentication for OAUTHBEARER and PLAIN authentication. 
+For an example configuration, see xref:proc-oauth-authentication-broker-config-{context}[].
+
+Session re-authentication can be used with _fast local JWT_ or _introspection endpoint_ token validation.
```
PaulRMellor, line 17:
```suggestion
Session re-authentication is used with _fast local JWT_ or _introspection endpoint_ token validation.
```

documentation/modules/oauth/con-oauth-reauthentication.adoc
```diff
@@ -1,23 +1,41 @@
...
 
-You enable session re-authentication in the `Kafka` resource. Set the `maxSecondsWithoutReauthentication` property for a TLS listener with `type: oauth` authentication.
-Session re-authentication is supported for both types of token validation (fast local JWT and introspection endpoint). For an example configuration, see xref:proc-oauth-authentication-broker-config-{context}[].
+.Differences in session expiry for OAUTHBEARER and PLAIN
+
+There are some differences in how session expiry works when session re-authentication is configured for OAUTHBEARER and PLAIN authentication.
 
-For more information about the re-authentication mechanism, which was added in Kafka version 2.2, see link:https://cwiki.apache.org/confluence/display/KAFKA/KIP-368%3A+Allow+SASL+Connections+to+Periodically+Re-Authenticate[KIP-368^].
+* For OAUTHBEARER, the broker's authenticated session will expire at the configured `maxSecondsWithoutReauthentication`. 
+The session will expire earlier if the access token expires before the `maxSecondsWithoutReauthentication`.
```
PaulRMellor, line 32: What about refresh token?

documentation/modules/oauth/con-oauth-reauthentication.adoc
```diff
@@ -1,23 +1,41 @@
...
-Session re-authentication is supported for both types of token validation (fast local JWT and introspection endpoint). For an example configuration, see xref:proc-oauth-authentication-broker-config-{context}[].
+.Differences in session expiry for OAUTHBEARER and PLAIN
+
+There are some differences in how session expiry works when session re-authentication is configured for OAUTHBEARER and PLAIN authentication.
 
-For more information about the re-authentication mechanism, which was added in Kafka version 2.2, see link:https://cwiki.apache.org/confluence/display/KAFKA/KIP-368%3A+Allow+SASL+Connections+to+Periodically+Re-Authenticate[KIP-368^].
+* For OAUTHBEARER, the broker's authenticated session will expire at the configured `maxSecondsWithoutReauthentication`. 
+The session will expire earlier if the access token expires before the `maxSecondsWithoutReauthentication`. 
+
+* For PLAIN, the broker's authenticated session will expire at the configured `maxSecondsWithoutReauthentication`.
```
PaulRMellor, line 34: What if PLAIN config uses a long-lived access token?

documentation/modules/oauth/con-oauth-reauthentication.adoc
```diff
@@ -1,23 +1,41 @@
...
+There are some differences in how session expiry works when session re-authentication is configured for OAUTHBEARER and PLAIN authentication.
 
-For more information about the re-authentication mechanism, which was added in Kafka version 2.2, see link:https://cwiki.apache.org/confluence/display/KAFKA/KIP-368%3A+Allow+SASL+Connections+to+Periodically+Re-Authenticate[KIP-368^].
+* For OAUTHBEARER, the broker's authenticated session will expire at the configured `maxSecondsWithoutReauthentication`. 
+The session will expire earlier if the access token expires before the `maxSecondsWithoutReauthentication`. 
+
+* For PLAIN, the broker's authenticated session will expire at the configured `maxSecondsWithoutReauthentication`. 
+
+If `maxSecondsWithoutReauthentication` is not configured, OAUTHBEARER and PLAIN clients can remain connected to brokers indefinitely, without needing to re-authenticate. 
+In that case, the sessions do not end with access token expiry, but this can be considered when configuring proper authorization.
```
PaulRMellor, line 37:
```suggestion
In this case, the sessions do not end with access token expiry, but this can be considered when configuring proper authorization. 
```

PaulRMellor: Not sure what _but this can be considered when configuring proper authorization._ is trying to say. Or what _proper authorization_ might be.

documentation/modules/oauth/proc-oauth-authentication-broker-config.adoc
```diff
@@ -133,8 +133,8 @@ Depending on how you apply OAuth 2.0 authentication, and the type of authorizati
...
-<7> Set this to `false` in order to disable the SASL_OAUTHBEARER mechanism on the listener. At least one of SASL_PLAIN or SASL_OAUTHBEARER has to be enabled. Default is `true`.
-<8> Set this to `true` in order to enable the SASL_PLAIN mechanism on the listener, which is supported by all clients on all platforms. The Kafka client has to enable the SASL_PLAIN mechanism and set the `username` and the `password`. This mechanism can be used to authenticate either by using the OAuth access token, or by using the OAuth clientId and secret (client credentials). If the client sets `username` to `access-token` then the password is interpreted as access token on the server, otherwise the user is interpreted as the clientId, and password as the client secret. Default is `false`.
+<7> Set this to `false` in order to disable the OAUTHBEARER mechanism on the listener. At least one of PLAIN or OAUTHBEARER has to be enabled. Default is `true`.
```
PaulRMellor, line 136:
```suggestion
<7> Set this to `false`to disable the OAUTHBEARER mechanism on the listener. At least one of PLAIN or OAUTHBEARER has to be enabled. Default is `true`.
```

documentation/modules/oauth/proc-oauth-authentication-broker-config.adoc
```diff
@@ -133,8 +133,8 @@ Depending on how you apply OAuth 2.0 authentication, and the type of authorizati
...
+<8> Set this to `true` in order to enable the PLAIN mechanism on the listener, which is supported by all clients on all platforms. The Kafka client has to enable the PLAIN mechanism and set the `username` and the `password`. This mechanism can be used to authenticate either by using the OAuth access token, or by using the OAuth clientId and secret (client credentials). If the client sets `password` to start with a string `$accessToken:` then the password is interpreted as access token on the server, and `username` as the account username, otherwise the user is interpreted as the clientId, and password as the client secret. Default is `false`.
```
PaulRMellor, line 137:
```suggestion
<8> Set this to `true` to enable the PLAIN mechanism on the listener, which is supported by all clients on all platforms. The Kafka client has to enable the PLAIN mechanism and set the `username` and the `password`. This mechanism can be used to authenticate either by using the OAuth access token, or by using the OAuth client id and secret (client credentials). If the client sets `password` to start with the string `$accessToken:`, the password is interpreted as the access token on the server, and `username` as the account username, otherwise the user is interpreted as the client id, and password as the client secret. Default is `false`.
```

PaulRMellor (commented):
Few more comments based on updates

documentation/assemblies/oauth/assembly-oauth-authentication.adoc
mstruk: What about 'keycloak' for that matter? You can configure any supported authorization mechanism independently of 'oauth' authentication.

documentation/modules/oauth/con-oauth-authentication-flow.adoc
mstruk: I don't know about the choice of workd but let me just clarify a bit ...
Many client tools use libraries that have basic OAUTHBEARER support (they understand it at protocol level), but require the client to write and install a callback handler that provides the access token. We provide such callback handler for upstream Kafka Client java library, but not for any other Kafka client library in other languages like Go, C/C++, JS ... In that case the client would have to write their own code for connecting to authorization server and obtaining the token.
Some Kafka client libraries may not even support OAUTHBEARER at protocol level or handle it properly by providing a callback mechanism. In that case the client simply can't use OAUTHBEARER.

documentation/modules/oauth/con-oauth-reauthentication.adoc
mstruk: That doesn't sound right, rather it is that the listener on the Kafka broker can be configured with re-authentication (you can turn it on using `maxSecondsWithoutReauthentication` value greater than 0 - by default it is turned off).
But if enabled on the listener it will only work if client libraries support it. Upstream Kafka client libraries support it.

documentation/modules/oauth/con-oauth-reauthentication.adoc
mstruk: Refresh token is never sent to the Kafka broker for authentication. It is only saved in Kafka client configuration, when configured, and is used by client to obtain a fresh new access token as needed, which is then sent to the Kafka broker.

documentation/modules/oauth/con-oauth-reauthentication.adoc
mstruk: If the access token expires before the `maxSecondsWithoutReauthentication` then the re-authentication is triggered then. However, in that case the client will send an expired token as part of re-authentication because it has no mechanism in PLAIN to refresh the token. As a result the authentication will fail and the client will not be able to continue without configuring a new access token as the password, that is typically a manual step. This is where OAUTHBEARER would have no problem, and using clientId + secret over PLAIN would also have no problem.

documentation/modules/oauth/con-oauth-authentication-flow.adoc
laidan6000: To reduce the ambiguity, I changed "built in" and "supported" to just "used". Apparently all clients use PLAIN. I added a new sentence: "OAUTHBEARER can only be used for Kafka broker authentication by Kafka clients that support this mechanism".

documentation/modules/oauth/con-oauth-authentication-flow.adoc
laidan6000: @mstruk - Do we provide this?

documentation/modules/oauth/con-oauth-reauthentication.adoc
laidan6000: @mstruk - Does session re-auth (with OAUTHBEARER) apply to refresh tokens?

documentation/modules/oauth/con-oauth-reauthentication.adoc
laidan6000: I don't think that matters.

documentation/modules/oauth/con-oauth-reauthentication.adoc
mstruk: Re-authentication is configured and triggered by Kafka broker so it applies to Kafka clients regardless of how they are configured. The refresh token is on the Kafka client. Re-authentication applies. The client refreshes the access token by using the refresh token, and uses the new access token to re-authenticate.

documentation/modules/oauth/con-oauth-reauthentication.adoc
laidan6000: I may have distorted the meaning. @mstruk - Please can you clarify?

Updated to:

> Authenticated sessions do not end with access token expiry. 
> However, this can be considered when configuring authorization, for example, by using `keycloak` authorization or installing a custom authorizer.

documentation/modules/oauth/con-oauth-reauthentication.adoc
laidan6000: That sounds much better to me. @mstruk - FYI

documentation/modules/oauth/con-oauth-authentication-flow.adoc
mstruk: You mean example of client configuration for using PLAIN to connect to the Kafka broker?
There's documentation for it [here](https://github.com/strimzi/strimzi-kafka-oauth/blob/main/README.md#configuring-the-kafka-client-with-sasl_plain). PLAIN mechanism doesn't require any additional Strimzi libraries on the client.

OAUTHBEARER on the other hand requires strimzi-kafka-oauth client libraries as explained [here](https://github.com/strimzi/strimzi-kafka-oauth/blob/main/README.md#configuring-the-kafka-client-with-sasl_oauthbearer)

documentation/assemblies/oauth/assembly-oauth-authentication.adoc
laidan6000: This note is becoming unwieldy! I think we should just go back to a simple link.

NOTE: OAuth 2.0 authentication can be used in conjunction with __Kafka authorization__.

scholzj:
So, @mstruk is this now OK? @laidan6000 could you please fix the DCO?

laidan6000:
I am still working on some edits related to @mstruk 's recent comments.
I will update the PR ASAP.

documentation/modules/oauth/con-oauth-reauthentication.adoc
laidan6000: OK, I attempted to add an explanation of the behaviour.

> For PLAIN using the long-lived access token method:
> 
> * The broker's authenticated session will expire at the configured `maxSecondsWithoutReauthentication`. 
> 
> * Re-authentication will fail if the access token expires before the configured time. Although session re-authentication is attempted, PLAIN has no mechanism for refreshing tokens.

documentation/modules/oauth/con-oauth-reauthentication.adoc
laidan6000: Thanks, well spotted. I added a new part about refresh tokens.

> Session re-authentication also applies to refresh tokens, if used. When the session expires, the client refreshes the access token by using its refresh token. The client then uses the new access token to re-authenticate to the existing session.

laidan6000:
Fixed the DCO.
@mstruk - This is ready for you to check again. Thanks.

documentation/modules/oauth/con-oauth-authentication-flow.adoc
```diff
@@ -1,12 +1,91 @@
...
+Configure clients to use OAUTHBEARER whenever possible. 
+This mechanism provides a higher level of security because client credentials are _never_ shared with Kafka brokers. 
+Consider using PLAIN only with Kafka clients that do not support OAUTHBEARER. 
 
-Kafka brokers and clients need to be configured to use OAuth 2.0.
+If necessary, OAUTHBEARER and PLAIN can be enabled together, on the same `oauth` listener.
+
+.OAUTHBEARER overview
+
+OAUTHBEARER is the default authentication mechanism included as part of Kafka.
```
mstruk, line 20: If I didn't know better, I would understand this sentence as if OAUTHBEARER is the mechanism everyone uses with Kafka because all clients and server use it by default. That is of course not the case. OAUTHBEARER is supported, but has to be explicitly configured, and if you don't want to write your own callback handlers, you don't have many other options (no other option that I know of in fact) but to use the callbacks provided by Strimzi.

documentation/modules/oauth/con-oauth-authentication-flow.adoc
```diff
@@ -1,12 +1,91 @@
...
+====
+OAUTHBEARER authentication can only be used by Kafka clients that support the OAUTHBEARER mechanism at the protocol level.
+====
+
+.PLAIN overview
+
+PLAIN is a simple authentication mechanism used by all Kafka client tools (including developer tools such as kafkacat). 
+To enable PLAIN to be used together with OAuth 2.0 authentication, Strimzi includes server-side callbacks and calls this _OAuth 2.0 over PLAIN_. 
+
+With PLAIN, client credentials are handled centrally, behind a compliant authorization server.
```
mstruk, line 58:
```suggestion
With Strimzi implementation of PLAIN, the client credentials are not stored in Zookeeper but are handled centrally, behind a compliant authorization server, just like when OAUTHBEARER is used.
```

documentation/modules/oauth/con-oauth-authentication-flow.adoc
laidan6000: I will update the explanation, thank you.

documentation/modules/oauth/con-oauth-authentication-flow.adoc
laidan6000: I applied the suggestion but reworded slightly.

mstruk (approved):

laidan6000:
@scholzj - Please merge this PR into `main` and then cherry-pick the commit into `release-0.22.x`.

This PR adds additional documentation on SASL_PLAIN for OAuth 2.0, which was released in Strimzi 0.22.0.

PR #4564, opened 2021-03-12
Move master to CRD v1

scholzj:
### Type of change

- Bugfix

### Description

Now after 0.22 is branched, we can move master to Strimzi `v1beta2` API / CRD `apiextensions/v1` to get it ready for 0.23.

This should close #4118 and close #3875.

### Checklist

- [x] Write tests
- [x] Make sure all tests pass
- [x] Try your changes from Pod inside your Kubernetes and OpenShift cluster, not just locally
- [x] Reference relevant issue(s) and close them after merging
- [x] Update CHANGELOG.md

PR #4608, opened 2021-03-19
[systemtest] Make upgrade work with new CRDs

im-konge:
### Type of change

- Enhancement

### Description

After `0.22.0` was released and also #4564 has been merged, our upgrade tests starts to fail, because of CRDs upgrade. This PR fixes the problem in upgrade: when we want to upgrade from version `X` to `HEAD`, we have to do the mid step and upgrade to `0.22.0`, use the `api-conversion.sh` script - upgrade CRs and CRDs, and then upgrade to `HEAD`

### Checklist


im-konge:
@strimzi-ci run tests test_only profile=upgrade

im-konge:
@strimzi-ci run tests test_only profile=upgrade

im-konge:
@strimzi-ci run tests test_only profile=upgrade

systemtest/src/test/java/io/strimzi/systemtest/upgrade/AbstractUpgradeST.java
```diff
@@ -97,23 +100,74 @@ protected static JsonArray readUpgradeJson(String fileName) {
...
+                data.put("toExamples", "HEAD");
+
+                // Generate procedures for upgrade
+                JsonObject procedures = new JsonObject();
+                procedures.put("kafkaVersion", testKafkaVersion.version());
+                procedures.put("logMessageVersion", testKafkaVersion.messageVersion());
+                procedures.put("interBrokerProtocolVersion", testKafkaVersion.protocolVersion());
+                data.put("proceduresAfterOperatorUpgrade", procedures);
+
+                parameters.add(Arguments.of(data.getString("fromVersion"), "0.22.0", "HEAD", data));
```
Frawless, line 116:
```suggestion
                parameters.add(Arguments.of(data.getString("fromVersion"), "0.22.1", "HEAD", data));
```

scholzj (approved):
I don't understand this much. But if it works and if Jakub is fine with it, it should be good. Should we ask for review from @see-quick as well?

Frawless:
@scholzj  we had discussion offline about the implementation and we are going to continue in it today so it's not ready yet.

systemtest/src/test/java/io/strimzi/systemtest/upgrade/AbstractUpgradeST.java
```diff
@@ -108,12 +112,56 @@ protected static JsonArray readUpgradeJson(String fileName) {
...
         });
 
         return parameters.stream();
     }
 
+    protected static List<JsonObject> buildMidStepUpgradeData(JsonObject jsonData) {
+        List<TestKafkaVersion> testKafkaVersions = TestKafkaVersion.getKafkaVersions();
+        TestKafkaVersion testKafkaVersion = testKafkaVersions.get(testKafkaVersions.size() - 1);
+
+        List<JsonObject> steps = new ArrayList<>();
```
Frawless, line 126: Maybe would be better to have Map instead of array? To avoid magic numbers later in getting data?

Frawless (approved):

systemtest/src/test/java/io/strimzi/systemtest/upgrade/AbstractUpgradeST.java
```diff
@@ -54,6 +56,7 @@
 public class AbstractUpgradeST extends AbstractST {
 
     private static final Logger LOGGER = LogManager.getLogger(AbstractUpgradeST.class);
+    private io.strimzi.systemtest.Constants systemtestConstants;
```
see-quick, line 59: Just a nit. `io.strimzi.systemtest.Constants` -> Constants

systemtest/src/test/java/io/strimzi/systemtest/upgrade/AbstractUpgradeST.java
```diff
@@ -108,12 +112,56 @@ protected static JsonArray readUpgradeJson(String fileName) {
...
+
+        midStep.put("urlTo", midStepUrl);
+        midStep.put("toVersion", midStepVersion);
+        midStep.put("toExamples", midStepExamples);
+        midStep.put("urlToConversionTool", conversionTool.getString("urlToConversionTool"));
+        midStep.put("toConversionTool", conversionTool.getString("toConversionTool"));
+
+        JsonObject midStepProcedures = new JsonObject();
+        midStepProcedures.put("kafkaVersion", testKafkaVersion.version());
+        midStepProcedures.put("logMessageVersion", testKafkaVersion.messageVersion());
```
see-quick, line 148: Would be great to have some abstraction class, which will stores all properties inside JSON so we can avoid typo errors.

systemtest/src/test/java/io/strimzi/systemtest/upgrade/AbstractUpgradeST.java
im-konge: There are already different Constants, so you need to do something like this - or you have to do `io.strimzi.systemtest.Constants.GLOBAL_TIMEOUT` etc. everywhere where it's needed.

systemtest/src/test/java/io/strimzi/systemtest/upgrade/AbstractUpgradeST.java
im-konge: Yes, that's sounds good, let me do it in next PR

PR #3471, opened 2020-08-06
Add cluster before/after load information to rebalance optimization proposal

tomncooper:
### Type of change

Enhancement

### Description

This PR addresses #3279 and adds broker load information to the rebalance optimization proposal. The load information is added as a new field in the `KafkaRebalance` resource's status. Each broker has its own entry with multiple metrics that show arrays consisting of [before, after, difference] values. An example status is shown below:

```
Status:
  Conditions:
    Last Transition Time:  2020-08-05T14:59:40.766999Z
    Status:                True
    Type:                  ProposalReady
  Observed Generation:     1
  Optimization Result:
    Broker Load:
      0:
        Cpu Percentage:
          0.36882564425468445
          0.3104616701602936
          -0.05836397409439087
        Disk Used MB:
          0.08218097686767578
          0.061748504638671875
          -0.020432472229003906
        Disk Used Percentage:
          8.218097686767578e-05
          6.174850463867188e-05
          -2.0432472229003904e-05
        Follower Network In Rate:
          0.01287756860256195
          0.021891381591558456
          0.009013812988996506
        Leader Network In Rate:
          0.04915750026702881
          0.03794153034687042
          -0.011215969920158386
        Leaders:
          22
          1
          -21
        Network Out Rate:
          0.08511605020612478
          0.07390008121728897
          -0.011215968988835812
        Potential Max Network Out Rate:
          0.09799361787736416
          0.09579146280884743
          -0.0022021550685167313
        Replicas:
          43
          39
          -4
      1:
        Cpu Percentage:
          0.0057828957214951515
          0.041537750512361526
          0.035754854790866375
        Disk Used MB:
          0.014960289001464844
          0.029038429260253906
          0.014078140258789062
        Disk Used Percentage:
          1.4960289001464843e-05
          2.9038429260253908e-05
          1.4078140258789065e-05
        Follower Network In Rate:
          0.01352857705205679
          0.008375635370612144
          -0.005152941681444645
        Leader Network In Rate:
          0.011323668994009495
          0.0182796698063612
          0.006956000812351704
        Leaders:
          22
          34
          12
        Network Out Rate:
          0.011323668994009495
          0.0182796698063612
          0.006956000812351704
        Potential Max Network Out Rate:
          0.024852246046066284
          0.026655305176973343
          0.0018030591309070587
        Replicas:
          44
          47
          3
      2:
        Cpu Percentage:
          0.005323501303792
          0.027932610362768173
          0.022609109058976173
        Disk Used MB:
[...]

Issue #3279, opened 2020-07-02
[Enhancement] Add load before and after information to rebalance optimization proposal


tomncooper:
## Optimization Summary

Currently the status of the `KafkaRebalance` CR shows the `Summary` field of the json returned from Cruise Control rebalance endpoint. This provides an overview of what the rebalance would entail:

```
{
  "numIntraBrokerReplicaMovements": 0,
  "excludedBrokersForLeadership": [],
  "numReplicaMovements": 30,
  "onDemandBalancednessScoreAfter": 74.54047494896054,
  "onDemandBalancednessScoreBefore": 74.54047494896054,
  "intraBrokerDataToMoveMB": 0,
  "recentWindows": 1,
  "dataToMoveMB": 0,
  "monitoredPartitionsPercentage": 100,
  "excludedTopics": [],
  "numLeaderMovements": 12,
  "excludedBrokersForReplicaMove": []
}
```
This lets a user judge the size of the changes involved in this proposed rebalance. However, the only comparison between the cluster before and after are the `onDemandBalancednessScore`s. Due to how these are calculated, they may not actually show a difference despite the fact that cluster will actually change. Therefore a user cannot always tell from the summary what changes are involved in a proposed rebalance.

## Broker Load Before/After

The json returned by the rebalance endpoint also contains details of the load on each broker before and after the proposed rebalance. 

### Before Optimization

```
curl -X POST 'localhost:9090/kafkacruisecontrol/rebalance?json=true&verbose=true' | jq .loadBeforeOptimization.brokers
[
  {
    "FollowerNwInRate": 0.013525374233722687,
    "BrokerState": "ALIVE",
    "Broker": 0,
    "NwOutRate": 0.09418795816600323,
    "NumCore": 1,
    "Host": "my-cluster-kafka-0.my-cluster-kafka-brokers.myproject.svc",
    "CpuPct": 0.7460259795188904,
    "Replicas": 52,
    "NetworkInCapacity": 10000,
    "Rack": "my-cluster-kafka-0.my-cluster-kafka-brokers.myproject.svc",
    "Leaders": 24,
    "DiskCapacityMB": 100000,
    "DiskMB": 1.9566116333007812,
    "PnwOutRate": 0.10786869935691357,
    "NetworkOutCapacity": 10000,
    "LeaderNwInRate": 0.07638948410749435,
    "DiskPct": 0.001956611633300781
  },
  {
    "FollowerNwInRate": 0.011713990941643715,
[...]

PR #4622, opened 2021-03-24
Add cluster before/after load information to rebalance optimization proposal in a ConfigMap

ShubhamRwt:
### Type of change

- Enhancement / new feature
- 
### Description

Enhancement

Description
This PR addresses #3279 and it is an extension of the PR #3471.  The load information is added in a configMap which is attached to the Kafka Rebalance Assembly Operator. Each broker has its own entry with multiple metrics that show arrays consisting of [before, after, difference] values. An example status is shown below: 

```Status:
  Conditions:
    Last Transition Time:  2020-08-05T14:59:40.766999Z
    Status:                True
    Type:                  ProposalReady
  Observed Generation:     1
  Optimization Result:
    Broker Load:
      0:
        Cpu Percentage:
          0.36882564425468445
          0.3104616701602936
          -0.05836397409439087
        Disk Used MB:
          0.08218097686767578
          0.061748504638671875
          -0.020432472229003906
        Disk Used Percentage:
          8.218097686767578e-05
          6.174850463867188e-05
          -2.0432472229003904e-05
        Follower Network In Rate:
          0.01287756860256195
          0.021891381591558456
          0.009013812988996506
        Leader Network In Rate:
          0.04915750026702881
          0.03794153034687042
          -0.011215969920158386
        Leaders:
          22
          1
          -21
        Network Out Rate:
          0.08511605020612478
          0.07390008121728897
          -0.011215968988835812
        Potential Max Network Out Rate:
          0.09799361787736416
          0.09579146280884743
          -0.0022021550685167313
        Replicas:
          43
          39
          -4
      1:
        Cpu Percentage:
          0.0057828957214951515
          0.041537750512361526
          0.035754854790866375
        Disk Used MB:
          0.014960289001464844
          0.029038429260253906
          0.014078140258789062
        Disk Used Percentage:
          1.4960289001464843e-05
          2.9038429260253908e-05
          1.4078140258789065e-05
        Follower Network In Rate:
          0.01352857705205679
          0.008375635370612144
          -0.005152941681444645
        Leader Network In Rate:
          0.011323668994009495
          0.0182796698063612
          0.006956000812351704
        Leaders:
          22
          34
          12
        Network Out Rate:
          0.011323668994009495
          0.0182796698063612
          0.006956000812351704
        Potential Max Network Out Rate:
          0.024852246046066284
          0.026655305176973343
          0.0018030591309070587
        Replicas:
          44
          47
          3
      2:
        Cpu Percentage:
          0.005323501303792
          0.027932610362768173
          0.022609109058976173
        Disk Used MB:
          0.02252483367919922
          0.028879165649414062
          0.006354331970214844
        Disk Used Percentage:
          2.252483367919922e-05
          2.8879165649414063e-05
          6.354331970214844e-06
        Follower Network In Rate:
          0.010815710760653019
          0.006954839453101158
          -0.003860871307551861
        Leader Network In Rate:
          0.014682217501103878
          0.01894218660891056
          0.004259969107806683
        Leaders:
          21
          30
          9
        Network Out Rate:
          0.014682217501103878
          0.01894218660891056
          0.004259969107806683
        Potential Max Network Out Rate:
          0.025497928261756897
          0.02589702606201172
          0.0003990978002548218
        Replicas:
          42
          43
          1
    Summary:
      Data To Move MB:  0
      Excluded Brokers For Leadership:
      Excluded Brokers For Replica Move:
      Excluded Topics:
      Intra Broker Data To Move MB:         0
      Monitored Partitions Percentage:      100
      Num Intra Broker Replica Movements:   0
      Num Leader Movements:                 9
      Num Replica Movements:                36
      On Demand Balancedness Score After:   78.01176356230222
      On Demand Balancedness Score Before:  78.01176356230222
      Recent Windows:                       1
  Session Id:                               1b7a2642-e49a-4012-ba3d-afa5c6ca9af4
Events:                                     <none> ```

### Checklist

- [ ] Write tests
- [ ] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceAssemblyOperator.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceAssemblyOperator.java
@@ -142,7 +142,7 @@ public class KafkaRebalanceAssemblyOperator
 
     private static final long REBALANCE_POLLING_TIMER_MS = 5_000;
     private static final int MAX_API_RETRIES = 5;
-    protected static final String BROKER_LOAD_KEY = "brokerLoad";
+    protected static final String BROKER_LOAD_KEY = "brokerLoad.json";
     private final CrdOperator<KubernetesClient, KafkaRebalance, KafkaRebalanceList> kafkaRebalanceOperator;
     private final CrdOperator<KubernetesClient, Kafka, KafkaList> kafkaOperator;
     private final PlatformFeaturesAvailability pfa;
@@ -409,7 +409,7 @@ public class KafkaRebalanceAssemblyOperator
     }
 
     /**
-     * computeNextStatus returns a future to a KafkaRebalanceStatus that will be computed depending on the given
+     * computeNextStatus returns a future to a wrapper class containing ConfigMap and KafkaRebalanceStatus that will be computed depending on the given
      * KafkaRebalance state, status and annotations.
      */
     /* test */ protected Future<MapAndStatus<ConfigMap, KafkaRebalanceStatus>> computeNextStatus(Reconciliation reconciliation,
@@ -426,7 +426,7 @@ public class KafkaRebalanceAssemblyOperator
             case Rebalancing:
                 return onRebalancing(reconciliation, host, apiClient, kafkaRebalance, rebalanceAnnotation);
             case Stopped:
-                return onStop(kafkaRebalance, reconciliation, host, apiClient, rebalanceAnnotation, rebalanceOptionsBuilder);
+                return onStop(reconciliation, host, apiClient, kafkaRebalance, rebalanceAnnotation, rebalanceOptionsBuilder);
             case Ready:
                 // Rebalance Complete
                 return onReady(reconciliation, host, apiClient, kafkaRebalance, rebalanceAnnotation, rebalanceOptionsBuilder);
@@ -510,7 +510,7 @@ public class KafkaRebalanceAssemblyOperator
      *                              returned by the Cruise Control rebalance endpoint.
      * @param brokerLoadAfterArray The JSONArray of broker load JSONObjects, for after the optimization proposal is applied,
      *                             returned by the Cruise Control rebalance endpoint.
-     * @return A map linking from broker ID integer to a map of load parameter to [before, after, difference] arrays.
+     * @return A JsonObject linking from broker ID integer to a map of load parameter to [before, after, difference] arrays.
      */
     protected static JsonObject parseLoadStats(JsonArray brokerLoadBeforeArray, JsonArray brokerLoadAfterArray) {
 
@@ -582,18 +582,17 @@ public class KafkaRebalanceAssemblyOperator
         return brokersStats;
     }
 
+    /**
+     * A wrapper class containing used to bind the ConfigMap and the status together.
+     */
+
     static class MapAndStatus<T, K> {
 
         T loadMap;
-
         K status;
 
         public T getLoadMap() {
             return loadMap;
-        }
-
-        public void setLoadMap(T loadMap) {
-            this.loadMap = loadMap;
         }
 
         public K getStatus() {
@@ -616,7 +615,7 @@ public class KafkaRebalanceAssemblyOperator
      * load map.
      *
      * @param  proposalJson The JSONObject representing the response from the Cruise Control rebalance endpoint.
-     * @return A map containing the proposal summary and broker load maps.
+     * @return A wrapper class containing the proposal summary map and a config map containing broker load.
      */
     protected static MapAndStatus<ConfigMap, Map<String, Object>> processOptimizationProposal(KafkaRebalance kafkaRebalance, JsonObject proposalJson) {
 
@@ -647,10 +646,7 @@ public class KafkaRebalanceAssemblyOperator
                 .withData(Collections.singletonMap(BROKER_LOAD_KEY, beforeAndAfterBrokerLoad.encodePrettily()))
                 .build();
 
-        Map<String, Object> optimizationProposal = new HashMap<>();
-        optimizationProposal.put(CruiseControlRebalanceKeys.SUMMARY.getKey(),
-                proposalJson.getJsonObject(CruiseControlRebalanceKeys.SUMMARY.getKey()).getMap());
-        return new MapAndStatus<>(rebalanceMap, optimizationProposal);
+        return new MapAndStatus<>(rebalanceMap, proposalJson.getJsonObject(CruiseControlRebalanceKeys.SUMMARY.getKey()).getMap());
     }
 
     private MapAndStatus<ConfigMap, KafkaRebalanceStatus> buildRebalanceStatus(KafkaRebalance kafkaRebalance, String sessionID, KafkaRebalanceState cruiseControlState, JsonObject proposalJson) {
@@ -673,7 +669,7 @@ public class KafkaRebalanceAssemblyOperator
      * @param host Cruise Control service to which sending the rebalance proposal request
      * @param apiClient Cruise Control REST API client instance
      * @param rebalanceOptionsBuilder builder for the Cruise Control REST API client options
-     * @return a Future with the next {@code KafkaRebalanceStatus} including the state
+     * @return a Future with the next {@code MapAndStatus<ConfigMap, KafkaRebalanceStatus>} including the ConfigMap and state
      */
 
     private Future<MapAndStatus<ConfigMap, KafkaRebalanceStatus>> onNew(Reconciliation reconciliation,
@@ -693,7 +689,7 @@ public class KafkaRebalanceAssemblyOperator
      * @param kafkaRebalance Current {@code KafkaRebalance} resource
      * @param rebalanceAnnotation The current value for the strimzi.io/rebalance annotation
      * @param rebalanceOptionsBuilder builder for the Cruise Control REST API client options
-     * @return a Future with the next {@code KafkaRebalanceStatus} including the state
+     * @return a Future with the next {@code MapAndStatus<ConfigMap, KafkaRebalanceStatus> } including the ConfigMap and state
      */
     private Future<MapAndStatus<ConfigMap, KafkaRebalanceStatus>> onNotReady(Reconciliation reconciliation,
                                                     String host, CruiseControlApi apiClient,
@@ -726,7 +722,7 @@ public class KafkaRebalanceAssemblyOperator
      * @param kafkaRebalance Current {@code KafkaRebalance} resource
      * @param rebalanceAnnotation The current value for the strimzi.io/rebalance annotation
      * @param rebalanceOptionsBuilder builder for the Cruise Control REST API client options
-     * @return a Future with the next {@code KafkaRebalanceStatus} including the state
+     * @return a Future with the next {@code MapAndStatus<ConfigMap, KafkaRebalanceStatus>} including the ConfigMap and state.
      */
     private Future<MapAndStatus<ConfigMap, KafkaRebalanceStatus>> onPendingProposal(Reconciliation reconciliation,
                                                            String host, CruiseControlApi apiClient,
@@ -807,7 +803,7 @@ public class KafkaRebalanceAssemblyOperator
      * @param kafkaRebalance Current {@code KafkaRebalance} resource
      * @param rebalanceAnnotation The current value for the strimzi.io/rebalance annotation
      * @param rebalanceOptionsBuilder builder for the Cruise Control REST API client options
-     * @return a Future with the next {@code KafkaRebalanceStatus} including the state
+     * @return a Future with the next {@code MapAndStatus<ConfigMap, KafkaRebalanceStatus>} including the ConfigMap and state
      */
     private Future<MapAndStatus<ConfigMap, KafkaRebalanceStatus>> onProposalReady(Reconciliation reconciliation,
                                                          String host, CruiseControlApi apiClient,
@@ -845,7 +841,7 @@ public class KafkaRebalanceAssemblyOperator
      * @param apiClient Cruise Control REST API client instance
      * @param kafkaRebalance Current {@code KafkaRebalance} resource
      * @param rebalanceAnnotation The current value for the strimzi.io/rebalance annotation
-     * @return a Future with the next {@code KafkaRebalanceStatus} including the state
+     * @return a Future with the next {@code MapAndStatus<ConfigMap, KafkaRebalanceStatus>} including the state
      */
     private Future<MapAndStatus<ConfigMap, KafkaRebalanceStatus>> onRebalancing(Reconciliation reconciliation,
                                                        String host, CruiseControlApi apiClient,
@@ -972,11 +968,11 @@ public class KafkaRebalanceAssemblyOperator
      * @param apiClient Cruise Control REST API client instance
      * @param rebalanceAnnotation The current value for the strimzi.io/rebalance annotation
      * @param rebalanceOptionsBuilder builder for the Cruise Control REST API client options
-     * @return a Future with the next {@code KafkaRebalanceStatus} bringing the state
+     * @return a Future with the next {@code MapAndStatus<ConfigMap, KafkaRebalanceStatus>} bringing the ConfigMap and state.
      */
 
-    private Future<MapAndStatus<ConfigMap, KafkaRebalanceStatus>> onStop(KafkaRebalance kafkaRebalance, Reconciliation reconciliation,
-                                                String host, CruiseControlApi apiClient, KafkaRebalanceAnnotation rebalanceAnnotation,
+    private Future<MapAndStatus<ConfigMap, KafkaRebalanceStatus>> onStop(Reconciliation reconciliation,
+                                                String host, CruiseControlApi apiClient, KafkaRebalance kafkaRebalance, KafkaRebalanceAnnotation rebalanceAnnotation,
                                                 RebalanceOptions.RebalanceOptionsBuilder rebalanceOptionsBuilder) {
         if (rebalanceAnnotation == KafkaRebalanceAnnotation.refresh) {
             return requestRebalance(reconciliation, host, apiClient, kafkaRebalance, true, rebalanceOptionsBuilder);
@@ -998,7 +994,7 @@ public class KafkaRebalanceAssemblyOperator
      * @param apiClient Cruise Control REST API client instance
      * @param rebalanceAnnotation The current value for the strimzi.io/rebalance annotation
      * @param rebalanceOptionsBuilder builder for the Cruise Control REST API client options
-     * @return a Future with the next {@code KafkaRebalanceStatus} bringing the state
+     * @return a Future with the next {@code MapAndStatus<ConfigMap, KafkaRebalanceStatus>} bringing the ConfigMap and state
      */
     private Future<MapAndStatus<ConfigMap, KafkaRebalanceStatus>> onReady(Reconciliation reconciliation,
                                                  String host, CruiseControlApi apiClient,
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceAssemblyOperator.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceAssemblyOperator.java
@@ -358,7 +358,7 @@ public class KafkaRebalanceAssemblyOperator
         RebalanceOptions.RebalanceOptionsBuilder rebalanceOptionsBuilder = convertRebalanceSpecToRebalanceOptions(kafkaRebalance.getSpec());
 
         return computeNextStatus(reconciliation, host, apiClient, kafkaRebalance, currentState, rebalanceAnnotation, rebalanceOptionsBuilder)
-           .compose(desiredStatusandMap -> {
+           .compose(desiredStatusAndMap -> {
                // More events related to resource modification might be queued with a stale state. (potentially updated by the rebalance holding the lock)
                // Due to possible long rebalancing operations that take the lock for the entire period,
                // do a new get to retrieve the current resource state.
@@ -366,8 +366,8 @@ public class KafkaRebalanceAssemblyOperator
                             .compose(currentKafkaRebalance -> {
                                 if (currentKafkaRebalance != null) {
                                     return configMapOperator.reconcile(kafkaRebalance.getMetadata().getNamespace(),
-                                            kafkaRebalance.getMetadata().getName(), desiredStatusandMap.getLoadMap())
-                                            .compose(i -> updateStatus(currentKafkaRebalance, desiredStatusandMap.getStatus(), null))
+                                            kafkaRebalance.getMetadata().getName(), desiredStatusAndMap.getLoadMap())
+                                            .compose(i -> updateStatus(currentKafkaRebalance, desiredStatusAndMap.getStatus(), null))
                                             .compose(updatedKafkaRebalance -> {
                                                 log.info("{}: State updated to [{}] with annotation {}={} ",
                                                         reconciliation,
@@ -585,7 +585,6 @@ public class KafkaRebalanceAssemblyOperator
     /**
      * A wrapper class containing used to bind the ConfigMap and the status together.
      */
-
     static class MapAndStatus<T, K> {
 
         T loadMap;
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceAssemblyOperator.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceAssemblyOperator.java
@@ -3,8 +3,9 @@
  * License: Apache License 2.0 (see the file LICENSE or http://apache.org/licenses/LICENSE-2.0.html).
  */
 package io.strimzi.operator.cluster.operator.assembly;
-
 import io.fabric8.kubernetes.api.model.LabelSelector;
+import io.fabric8.kubernetes.api.model.ConfigMap;
+import io.fabric8.kubernetes.api.model.ConfigMapBuilder;
 import io.fabric8.kubernetes.client.KubernetesClient;
 import io.fabric8.kubernetes.client.KubernetesClientException;
 import io.fabric8.kubernetes.client.Watcher;
@@ -28,23 +29,27 @@ import io.strimzi.operator.cluster.model.CruiseControl;
 import io.strimzi.operator.cluster.model.InvalidResourceException;
 import io.strimzi.operator.cluster.model.NoSuchResourceException;
 import io.strimzi.operator.cluster.model.StatusDiff;
+import io.strimzi.operator.cluster.operator.resource.ResourceOperatorSupplier;
 import io.strimzi.operator.cluster.operator.resource.cruisecontrol.CruiseControlApi;
 import io.strimzi.operator.cluster.operator.resource.cruisecontrol.CruiseControlApiImpl;
+import io.strimzi.operator.cluster.operator.resource.cruisecontrol.CruiseControlLoadParameters;
+import io.strimzi.operator.cluster.operator.resource.cruisecontrol.CruiseControlRebalanceKeys;
 import io.strimzi.operator.cluster.operator.resource.cruisecontrol.CruiseControlRestException;
 import io.strimzi.operator.cluster.operator.resource.cruisecontrol.CruiseControlUserTaskStatus;
 import io.strimzi.operator.cluster.operator.resource.cruisecontrol.RebalanceOptions;
-import io.strimzi.operator.cluster.operator.resource.ResourceOperatorSupplier;
 import io.strimzi.operator.common.AbstractOperator;
 import io.strimzi.operator.common.Annotations;
 import io.strimzi.operator.common.Reconciliation;
 import io.strimzi.operator.common.Util;
 import io.strimzi.operator.common.model.Labels;
 import io.strimzi.operator.common.operator.resource.AbstractWatchableStatusedResourceOperator;
+import io.strimzi.operator.common.operator.resource.ConfigMapOperator;
 import io.strimzi.operator.common.operator.resource.CrdOperator;
 import io.strimzi.operator.common.operator.resource.StatusUtils;
 import io.vertx.core.Future;
 import io.vertx.core.Promise;
 import io.vertx.core.Vertx;
+import io.vertx.core.json.JsonArray;
 import io.vertx.core.json.JsonObject;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
@@ -52,6 +57,7 @@ import org.apache.logging.log4j.Logger;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
+import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
@@ -60,7 +66,6 @@ import java.util.concurrent.atomic.AtomicInteger;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
-import static io.strimzi.operator.cluster.operator.resource.cruisecontrol.CruiseControlApi.CC_REST_API_SUMMARY;
 import static io.strimzi.operator.common.Annotations.ANNO_STRIMZI_IO_REBALANCE;
 
 /**
@@ -137,12 +142,13 @@ public class KafkaRebalanceAssemblyOperator
 
     private static final long REBALANCE_POLLING_TIMER_MS = 5_000;
     private static final int MAX_API_RETRIES = 5;
-
+    protected static final String BROKER_LOAD_KEY = "brokerLoad";
     private final CrdOperator<KubernetesClient, KafkaRebalance, KafkaRebalanceList> kafkaRebalanceOperator;
     private final CrdOperator<KubernetesClient, Kafka, KafkaList> kafkaOperator;
     private final PlatformFeaturesAvailability pfa;
     private final Optional<LabelSelector> kafkaSelector;
 
+    private final ConfigMapOperator configMapOperator;
     /**
      * @param vertx The Vertx instance
      * @param pfa Platform features availability properties
@@ -156,6 +162,7 @@ public class KafkaRebalanceAssemblyOperator
         this.pfa = pfa;
         this.kafkaRebalanceOperator = supplier.kafkaRebalanceOperator;
         this.kafkaOperator = supplier.kafkaOperator;
+        this.configMapOperator = supplier.configMapOperations;
     }
 
     /**
@@ -336,6 +343,8 @@ public class KafkaRebalanceAssemblyOperator
                                    KafkaRebalanceState currentState, KafkaRebalanceAnnotation rebalanceAnnotation) {
 
         log.info("{}: Rebalance action from state [{}]", reconciliation, currentState);
+        log.info("" + kafkaRebalance.getMetadata().getNamespace());
+        log.info("" + reconciliation.namespace());
 
         if (Annotations.isReconciliationPausedWithAnnotation(kafkaRebalance)) {
             // we need to do this check again because it was triggered by a watcher
@@ -351,14 +360,16 @@ public class KafkaRebalanceAssemblyOperator
         RebalanceOptions.RebalanceOptionsBuilder rebalanceOptionsBuilder = convertRebalanceSpecToRebalanceOptions(kafkaRebalance.getSpec());
 
         return computeNextStatus(reconciliation, host, apiClient, kafkaRebalance, currentState, rebalanceAnnotation, rebalanceOptionsBuilder)
-           .compose(desiredStatus -> {
+           .compose(desiredStatusandMap -> {
                // More events related to resource modification might be queued with a stale state. (potentially updated by the rebalance holding the lock)
                // Due to possible long rebalancing operations that take the lock for the entire period,
                // do a new get to retrieve the current resource state.
                return kafkaRebalanceOperator.getAsync(reconciliation.namespace(), reconciliation.name())
                             .compose(currentKafkaRebalance -> {
                                 if (currentKafkaRebalance != null) {
-                                    return updateStatus(currentKafkaRebalance, desiredStatus, null)
+                                    return configMapOperator.reconcile(kafkaRebalance.getMetadata().getNamespace(),
+                                            kafkaRebalance.getMetadata().getName(), desiredStatusandMap.getLoadMap())
+                                            .compose(i -> updateStatus(currentKafkaRebalance, desiredStatusandMap.getStatus(), null))
                                             .compose(updatedKafkaRebalance -> {
                                                 log.info("{}: State updated to [{}] with annotation {}={} ",
                                                         reconciliation,
@@ -403,7 +414,7 @@ public class KafkaRebalanceAssemblyOperator
      * computeNextStatus returns a future to a KafkaRebalanceStatus that will be computed depending on the given
      * KafkaRebalance state, status and annotations.
      */
-    /* test */ protected Future<KafkaRebalanceStatus> computeNextStatus(Reconciliation reconciliation,
+    /* test */ protected Future<MapandStatus<ConfigMap, KafkaRebalanceStatus>> computeNextStatus(Reconciliation reconciliation,
                                                                         String host, CruiseControlApi apiClient,
                                                                         KafkaRebalance kafkaRebalance, KafkaRebalanceState currentState,
                                                                         KafkaRebalanceAnnotation rebalanceAnnotation, RebalanceOptions.RebalanceOptionsBuilder rebalanceOptionsBuilder) {
@@ -417,7 +428,7 @@ public class KafkaRebalanceAssemblyOperator
             case Rebalancing:
                 return onRebalancing(reconciliation, host, apiClient, kafkaRebalance, rebalanceAnnotation);
             case Stopped:
-                return onStop(reconciliation, host, apiClient, kafkaRebalance, rebalanceAnnotation, rebalanceOptionsBuilder);
+                return onStop(kafkaRebalance, reconciliation, host, apiClient, rebalanceAnnotation, rebalanceOptionsBuilder);
             case Ready:
                 // Rebalance Complete
                 return onReady(reconciliation, host, apiClient, kafkaRebalance, rebalanceAnnotation, rebalanceOptionsBuilder);
@@ -427,16 +438,6 @@ public class KafkaRebalanceAssemblyOperator
             default:
                 return Future.failedFuture(new RuntimeException("Unexpected state " + currentState));
         }
-    }
-
-    private KafkaRebalanceStatus buildRebalanceStatus(String sessionID, KafkaRebalanceState cruiseControlState, Set<Condition> validation) {
-        List<Condition> conditions = new ArrayList<>();
-        conditions.add(StatusUtils.buildRebalanceCondition(cruiseControlState.toString()));
-        conditions.addAll(validation);
-        return new KafkaRebalanceStatusBuilder()
-                .withSessionId(sessionID)
-                .withConditions(conditions)
-                .build();
     }
 
     private KafkaRebalanceStatus buildRebalanceStatusFromPreviousStatus(KafkaRebalanceStatus currentStatus, Set<Condition> validation) {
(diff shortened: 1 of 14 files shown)

### CODE REVIEW

File: cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceAssemblyOperator.java [hunk 1]
```diff
@@ -351,14 +360,16 @@ private String rebalanceStateConditionType(KafkaRebalanceStatus status) {
         RebalanceOptions.RebalanceOptionsBuilder rebalanceOptionsBuilder = convertRebalanceSpecToRebalanceOptions(kafkaRebalance.getSpec());
 
         return computeNextStatus(reconciliation, host, apiClient, kafkaRebalance, currentState, rebalanceAnnotation, rebalanceOptionsBuilder)
-           .compose(desiredStatus -> {
+           .compose(desiredStatusandMap -> {
```
scholzj, line 363: Maybe `desiredStatusAndMap` would be better (i.e. uppercase `A` in `and`)?

ShubhamRwt: Okay. I have applied that now

documentation/modules/cruise-control/con-optimization-proposals.adoc
```diff
@@ -94,10 +101,46 @@ m¦excludedBrokersForReplicaMove
...
+
+m| leaders                     | The number of replicas on this broker that are partition leaders.
+
+m| replicas                    | The number of replicas on this broker.
+
+m| cpuPercentage               | The CPU utilization as a percentage of the defined capacity.
+
+m| diskUsedPercentage          | The disk utilization as a percentage of the defined capacity.
+
+m| diskUsedMB                  | The absolute disk usage in megabytes.
```
laidan6000, line 128:
```suggestion
m| diskUsedMB                  | The absolute disk usage in MB.
```

laidan6000 (changes requested):
This is excellent content, thank you. I made some suggested changes. After applying the changes and pulling them to your local branch, you will need to fix the DCO sign-off check.

ShubhamRwt:
@laidan6000 thanks a lot for the changes. I will apply them.

documentation/modules/cruise-control/con-optimization-proposals.adoc
```diff
@@ -94,10 +101,46 @@ m¦excludedBrokersForReplicaMove
 
 |===
 
+
+[discrete]
+=== Broker load
+
+In the broker load section, each broker is listed under its broker ID number. 
+For each broker a number of metrics (described in the table below) are listed.
```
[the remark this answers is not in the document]
laidan6000: @ShubhamRwt - I don't think you need line 109. The table is introduced in line 113.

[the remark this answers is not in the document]
ShubhamRwt: @laidan6000 thanks. I guess I missed this one. Sorry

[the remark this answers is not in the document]
laidan6000: No problem!

laidan6000 (approved):
Looks good to me.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceAssemblyOperator.java
```diff
@@ -403,24 +414,24 @@ private String rebalanceStateConditionType(KafkaRebalanceStatus status) {
...
      */
-    /* test */ protected Future<KafkaRebalanceStatus> computeNextStatus(Reconciliation reconciliation,
+    /* test */ protected Future<MapandStatus<ConfigMap, KafkaRebalanceStatus>> computeNextStatus(Reconciliation reconciliation,
                                                                         String host, CruiseControlApi apiClient,
                                                                         KafkaRebalance kafkaRebalance, KafkaRebalanceState currentState,
                                                                         KafkaRebalanceAnnotation rebalanceAnnotation, RebalanceOptions.RebalanceOptionsBuilder rebalanceOptionsBuilder) {
         switch (currentState) {
             case New:
-                return onNew(reconciliation, host, apiClient, kafkaRebalance, rebalanceOptionsBuilder);
+                return onNew(kafkaRebalance, reconciliation, host, apiClient, rebalanceOptionsBuilder);
```
tomncooper, line 423: Can you make sure the order of these arguments is maintained as they were previouly. There is is no reason to change the order here. Same for the `Stopped` method below.

cluster-operator/src/test/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceAssemblyOperatorTest.java
```diff
@@ -66,16 +70,26 @@
 import static org.junit.jupiter.api.Assertions.assertNotNull;
 import static org.junit.jupiter.api.Assertions.assertTrue;
 import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.ArgumentMatchers.anyString;
```
tomncooper, line 73: You need to remove this unused import in order for checkstyle test to pass

cluster-operator/src/test/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceAssemblyOperatorTest.java
```diff
@@ -66,16 +70,26 @@
...
 import static org.mockito.Mockito.when;
 
 @ExtendWith(VertxExtension.class)
 public class KafkaRebalanceAssemblyOperatorTest {
-
     private static final String HOST = "localhost";
     private static final String RESOURCE_NAME = "my-rebalance";
     private static final String CLUSTER_NAMESPACE = "cruise-control-namespace";
     private static final String CLUSTER_NAME = "kafka-cruise-control-test-cluster";
+    private static ConfigMap map = new ConfigMapBuilder()
```
tomncooper, line 84: This map doesn't seem to be used for anything?

cluster-operator/src/test/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceStatusTest.java
```diff
@@ -0,0 +1,130 @@
+/*
```
tomncooper, line 1: Why is this class commented out?

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceAssemblyOperator.java
```diff
@@ -609,24 +811,25 @@ private KafkaRebalanceStatus buildRebalanceStatus(String sessionID, KafkaRebalan
...
+    private Future<MapandStatus<ConfigMap, KafkaRebalanceStatus>> onProposalReady(Reconciliation reconciliation,
                                                          String host, CruiseControlApi apiClient,
                                                          KafkaRebalance kafkaRebalance,
                                                          KafkaRebalanceAnnotation rebalanceAnnotation,
                                                          RebalanceOptions.RebalanceOptionsBuilder rebalanceOptionsBuilder) {
         switch (rebalanceAnnotation) {
             case none:
                 log.debug("{}: No {} annotation set", reconciliation, ANNO_STRIMZI_IO_REBALANCE);
-                return Future.succeededFuture(buildRebalanceStatusFromPreviousStatus(kafkaRebalance.getStatus(), validate(kafkaRebalance)));
+                configMapOperator.getAsync(kafkaRebalance.getMetadata().getNamespace(), kafkaRebalance.getMetadata().getName())
```
tomncooper, line 822: So this call looks correct, but the main issue is that you aren't returning anything, so this does nothing?

tomncooper:
`KafkaRebalanceStateMachineTest#testProposalReadyNoChange` is failing with a null pointer exception. I think this is probably beacuse the configmap.getAsync is not mocked properly. Have a look in the `KafkaAssemblyOperatorTest` class and add the `configMapOperations` mocking code to the state machine test, that should allow this test to pass.

cluster-operator/src/test/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceStateMachineTest.java
```diff
@@ -58,6 +62,7 @@
     private static final String RESOURCE_NAME = "my-rebalance";
     private static final String CLUSTER_NAMESPACE = "cruise-control-namespace";
     private static final String CLUSTER_NAME = "kafka-cruise-control-test-cluster";
+    ConfigMapOperator mockCmOps;
```
tomncooper, line 65: Why are you making this mock a field. The other mocks are just method instances?

tomncooper, line 65: Why is this a being stored as a field in the class? The other mocks are method instances?

cluster-operator/src/test/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceStateMachineTest.java
```diff
@@ -162,15 +168,18 @@ public String cruiseControlHost(String clusterName, String clusterNamespace) {
                 KafkaRebalance,
                 KafkaRebalanceList> mockRebalanceOps = supplier.kafkaRebalanceOperator;
 
+        when(mockCmOps.getAsync(eq(CLUSTER_NAMESPACE), eq(RESOURCE_NAME))).thenReturn(Future.succeededFuture(new ConfigMap()));
```
tomncooper, line 171: Do you need the `eq()`?

ShubhamRwt: I have tried with both with and without eq(), looks like we dont require it since both leads the same result

cluster-operator/src/test/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceStatusTest.java
```diff
@@ -0,0 +1,202 @@
...
+
+        assertTrue(output.getStatus().containsKey(CruiseControlRebalanceKeys.SUMMARY.getKey()));
+
+        Map<String, String> brokerMap = output.getLoadMap().getData();
+
+        try {
+
+            ObjectMapper mapper = new ObjectMapper();
+
+            Map<String, LinkedHashMap<String, String>> brokerLoadMap = mapper.readValue(brokerMap.get("brokerLoad"), LinkedHashMap.class);
```
tomncooper, line 169: Similar to above you should test that the ConfigMap you were given has the data in you expected. You can test its fields directly, you don't need to force into a map.

cluster-operator/src/test/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceStatusTest.java
```diff
@@ -0,0 +1,202 @@
...
+                .getJsonArray(CruiseControlRebalanceKeys.BROKERS.getKey());
+        JsonArray loadAfterArray = proposal.getJsonObject(CruiseControlRebalanceKeys.LOAD_AFTER_OPTIMIZATION.getKey())
+                .getJsonArray(CruiseControlRebalanceKeys.BROKERS.getKey());
+
+        JsonObject output = KafkaRebalanceAssemblyOperator.parseLoadStats(
+                loadBeforeArray, loadAfterArray);
+
+        try {
+            ObjectMapper mapper = new ObjectMapper();
+            Map<String, LinkedHashMap<String, String>> outputMap = mapper.readValue(output.encodePrettily(), Map.class);
```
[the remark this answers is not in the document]
ShubhamRwt: Okay. I will write the logic corresponding to the changes. I thought we need to stick to the previous logic

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceAssemblyOperator.java
```diff
@@ -453,21 +454,220 @@ private KafkaRebalanceStatus buildRebalanceStatusFromPreviousStatus(KafkaRebalan
...
+                brokerLoadBeforeOptimization, brokerLoadAfterOptimization);
+
+        ConfigMap rebalanceMap = new ConfigMapBuilder()
+                .withApiVersion("v1")
+                .withNewMetadata()
+                .withNamespace(kafkaRebalance.getMetadata().getNamespace())
+                .withName(kafkaRebalance.getMetadata().getName())
+                .withLabels(Collections.singletonMap("app", "strimzi"))
+                .endMetadata()
+                .withData(Collections.singletonMap(BROKER_LOAD_KEY, beforeAndAfterBrokerLoad.encodePrettily()))
```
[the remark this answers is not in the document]
tomncooper: Then you would just be able to pass that map directly to the ConfigMap builder

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceAssemblyOperator.java
```diff
@@ -336,6 +343,8 @@ private String rebalanceStateConditionType(KafkaRebalanceStatus status) {
                                    KafkaRebalanceState currentState, KafkaRebalanceAnnotation rebalanceAnnotation) {
 
         log.info("{}: Rebalance action from state [{}]", reconciliation, currentState);
+        log.info("" + kafkaRebalance.getMetadata().getNamespace());
+        log.info("" + reconciliation.namespace());
```
scholzj, line 347: This looks like some debug logging ... maybe it can be removed?

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceAssemblyOperator.java
scholzj, line 363: I guess this should be `desiredStatusAndMap` with capital `A`?

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceAssemblyOperator.java
```diff
@@ -453,21 +454,220 @@ private KafkaRebalanceStatus buildRebalanceStatusFromPreviousStatus(KafkaRebalan
...
+
+            }
+
+            brokersStats.put(String.valueOf(loadBeforeEntry.getKey()), brokerStats);
+        }
+
+        return brokersStats;
+    }
+
+    static class MapandStatus<T, K> {
```
scholzj, line 587: `MapAndStatus` (capital `A`) would be a better name.

scholzj (commented):
* This might be worth adding a record to the CHANGELOG.md file
* * Can you share an example of the resulting Config Map?

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceAssemblyOperator.java
ShubhamRwt: I have made these changes. Pushing it to the branch now

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceAssemblyOperator.java
ShubhamRwt: Sorry. I have removed them now

ShubhamRwt:
> * This might be worth adding a record to the CHANGELOG.md file
> * * Can you share an example of the resulting Config Map?

@scholzj Hello, Here is the Link to the ConfigMap that is getting generated https://gist.github.com/ShubhamRwt/e044a189dab03f2313f991b6dd074d66

scholzj:
Thanks for sharing the ConfigMap. I guess I have some general questions about it:
* Given it is a JSON, should the key be `brokerLoad.json` to make it clear what is the format?
* Do we expect it to be used programatically? Or read by humans? If we expect it to be read by humans, should we have some more _human friendly_ format? Some CSV they can download and export to Excel? HTML they can download and read in browser?

_(this is more general question, not just for Shubham, but also @tomncooper, @kyguy and @strimzi/maintainers)_

kyguy:
> Given it is a JSON, should the key be brokerLoad.json to make it clear what is the format?

That sounds good to me

> Do we expect it to be used programatically? Or read by humans? If we expect it to be read by humans, should we have some more human friendly format? Some CSV they can download and export to Excel? HTML they can download and read in browser?

I would expect programatically, people can ingest the JSON data with their own tools (maybe we could provide a tool in a future PR)

ShubhamRwt:
> * Do we expect it to be used programatically? Or read by humans? If we expect it to be read by humans, should we have some more _human friendly_ format? Some CSV they can download and export to Excel? HTML they can download and read in browser?

I think that JSON format is quite easy to read so we can keep it this way and also as @kyguy suggested the JSON can be easily mended by the user into readable format and we can also append some tool later on which can make things more easier for the user.

scholzj:
Ok. You should still add some record to the CHANGELOG.md file.

ShubhamRwt:
> Ok. You should still add some record to the CHANGELOG.md file.

Okay. I will add it

operator-common/src/main/java/io/strimzi/operator/cluster/operator/resource/cruisecontrol/CruiseControlLoadParameters.java
```diff
@@ -0,0 +1,70 @@
...
+    }
+
+    public String getType() {
+        return type;
+    }
+
+    private static List<CruiseControlLoadParameters> filterByType(String filterType) {
+        return Arrays.stream(CruiseControlLoadParameters.values())
+                .filter(loadParameter -> loadParameter.getType() != null)
+                .filter(loadParameter -> loadParameter.getType().equals(filterType))
```
sknot-rh, line 58: If you use `filterType.equals(loadParameter.getType())` you can omit the upper filtering.

documentation/modules/cruise-control/con-optimization-proposals.adoc
```diff
@@ -34,7 +34,14 @@ Consider a shorter interval for fast changing clusters, although this increases
 [discrete]
 == Contents of optimization proposals
 
-The following table describes the properties contained in an optimization proposal:
+An optimization proposal consists of two main sections: the _summary_ and the _broker load_.
+The summary provides an overview of the proposed cluster rebalance and indicates the scale of the changes involved.
+The broker load allows you to see the impact of the proposed rebalance on each of the brokers in the cluster.
```
laidan6000, line 39: _Suggested change:_

The broker load allows you to see the impact of the proposed rebalance on each of the brokers in the cluster.

>

The broker load shows before and after values for the proposed rebalance, so you can see the impact on each of the brokers in the cluster.

ShubhamRwt: Okay. I will apply this change

scholzj (approved):

operator-common/src/main/java/io/strimzi/operator/cluster/operator/resource/cruisecontrol/CruiseControlLoadParameters.java
ShubhamRwt: Done @sknot-rh

sknot-rh (approved):

scholzj:
@tomncooper Do you have some more comments to this?

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceAssemblyOperator.java
tomncooper: This may be an issue at my end (I have refreshed the page several times) but I am not seeing this change to `desiredStatusAndMap`. Are you sure you have pushed your latest changes?

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceAssemblyOperator.java
```diff
@@ -417,7 +426,7 @@ private String rebalanceStateConditionType(KafkaRebalanceStatus status) {
             case Rebalancing:
                 return onRebalancing(reconciliation, host, apiClient, kafkaRebalance, rebalanceAnnotation);
             case Stopped:
-                return onStop(reconciliation, host, apiClient, kafkaRebalance, rebalanceAnnotation, rebalanceOptionsBuilder);
+                return onStop(kafkaRebalance, reconciliation, host, apiClient, rebalanceAnnotation, rebalanceOptionsBuilder);
```
tomncooper, line 429: As I have said before, can you keep the method signatures (argument ordering) consistent. `reconciliation` should be first with `kafkaRebalance` after `apiClient`.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceAssemblyOperator.java
```diff
@@ -453,21 +452,220 @@ private KafkaRebalanceStatus buildRebalanceStatusFromPreviousStatus(KafkaRebalan
...
-     * When a new {@code KafkaRebalance} is created, it calls the Cruise Control API requesting a rebalance proposal.
+     * Converts the supplied before and after broker load arrays into a map linking from broker ID integer to a map linking
+     * from load parameter to an array of [before, after, difference]. The load paramters included in the map are dictated
+     * by the values in he {@link CruiseControlLoadParameters} enum.
      *
+     * @param brokerLoadBeforeArray The JSONArray of broker load JSONObjects, for before the optimization proposal is applied,
+     *                              returned by the Cruise Control rebalance endpoint.
+     * @param brokerLoadAfterArray The JSONArray of broker load JSONObjects, for after the optimization proposal is applied,
+     *                             returned by the Cruise Control rebalance endpoint.
+     * @return A map linking from broker ID integer to a map of load parameter to [before, after, difference] arrays.
```
tomncooper, line 513: Can you fix the javadoc for this method. It returns a `JsonObject` not a map.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceAssemblyOperator.java
```diff
@@ -453,21 +452,220 @@ private KafkaRebalanceStatus buildRebalanceStatusFromPreviousStatus(KafkaRebalan
...
+            brokersStats.put(String.valueOf(loadBeforeEntry.getKey()), brokerStats);
+        }
+
+        return brokersStats;
+    }
+
+    static class MapAndStatus<T, K> {
+
+        T loadMap;
+
```
tomncooper, line 588: Not sure we need this space. Also maybe add a brief bit of javadoc/comment saying what this class is for?

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceAssemblyOperator.java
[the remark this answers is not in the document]
tomncooper: After further discussion I am happy to leave this the way it is.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceAssemblyOperator.java
```diff
@@ -453,21 +452,220 @@ private KafkaRebalanceStatus buildRebalanceStatusFromPreviousStatus(KafkaRebalan
...
+        }
+    }
+
+    /**
+     * Converts the supplied JSONObject containing the response from the {@link CruiseControlApi#rebalance} or
+     * {@link CruiseControlApi#getUserTaskStatus} methods, into a map linking to a proposal summary map and a broker
+     * load map.
+     *
+     * @param  proposalJson The JSONObject representing the response from the Cruise Control rebalance endpoint.
+     * @return A map containing the proposal summary and broker load maps.
```
tomncooper, line 619: Again, can we make sure the javadoc is accurate. This returns a `MapAndStatus` instance not a map.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceAssemblyOperator.java
```diff
@@ -453,21 +452,220 @@ private KafkaRebalanceStatus buildRebalanceStatusFromPreviousStatus(KafkaRebalan
...
+                .withNewMetadata()
+                .withNamespace(kafkaRebalance.getMetadata().getNamespace())
+                .withName(kafkaRebalance.getMetadata().getName())
+                .withLabels(Collections.singletonMap("app", "strimzi"))
+                .endMetadata()
+                .withData(Collections.singletonMap(BROKER_LOAD_KEY, beforeAndAfterBrokerLoad.encodePrettily()))
+                .build();
+
+        Map<String, Object> optimizationProposal = new HashMap<>();
+        optimizationProposal.put(CruiseControlRebalanceKeys.SUMMARY.getKey(),
```
tomncooper, line 651: As far as I can see this is a change from the current behavior where the `optimizationResult` given to the `KafkaRebalanceStatusBuilder` is not under an additional summary key it is just the map of the contents of the summary. For example [here](https://github.com/strimzi/strimzi-kafka-operator/blob/1554521a7a0ef5a641717e94ad762834fec89608/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceAssemblyOperator.java#L926). So this would change the format of status compare to the current code.

I think you just need to return the summary map as the second argument in the return statement. No need to create a new map for this one entry: `return new MapAndStatus<>(rebalanceMap, proposalJson.getJsonObject(CruiseControlRebalanceKeys.SUMMARY.getKey()).getMap())`

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceAssemblyOperator.java
```diff
@@ -774,12 +974,12 @@ private KafkaRebalanceStatus buildRebalanceStatus(String sessionID, KafkaRebalan
      * @param rebalanceOptionsBuilder builder for the Cruise Control REST API client options
      * @return a Future with the next {@code KafkaRebalanceStatus} bringing the state
```
tomncooper, line 975: You need to update the javadoc for all the `on<state>` methods as they now return a `Future<MapAndStatus>` not just the status future.

cluster-operator/src/test/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceStateMachineTest.java
tomncooper: Have you addressed this @ShubhamRwt?

cluster-operator/src/test/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceStateMachineTest.java
```diff
@@ -162,15 +165,18 @@ public String cruiseControlHost(String clusterName, String clusterNamespace) {
...
 
+        when(mockCmOps.getAsync(CLUSTER_NAMESPACE, RESOURCE_NAME)).thenReturn(Future.succeededFuture(new ConfigMap()));
         when(mockRebalanceOps.get(CLUSTER_NAMESPACE, RESOURCE_NAME)).thenReturn(kcRebalance);
         when(mockRebalanceOps.getAsync(CLUSTER_NAMESPACE, RESOURCE_NAME)).thenReturn(Future.succeededFuture(kcRebalance));
 
         return kcrao.computeNextStatus(recon, HOST, client, kcRebalance, currentState, initialAnnotation, rbOptions)
                 .compose(result -> {
                     context.verify(() -> {
-                        assertThat(result.getConditions(), StateMatchers.hasStateInConditions(nextState));
+                        System.out.println(result.getStatus().getConditions());
```
tomncooper, line 175: There shouldn't be print lines in you code, either remove them or add a log statement.

cluster-operator/src/test/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceStatusTest.java
tomncooper: You can resolve this comment

cluster-operator/src/test/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceStatusTest.java
```diff
@@ -0,0 +1,200 @@
...
+import static org.hamcrest.Matchers.hasKey;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+import static org.keycloak.util.JsonSerialization.mapper;
+
+
+public class KafkaRebalanceStatusTest {
+
+    private static final int BROKER_ONE_KEY = 1;
+
+    private static final String BROKER_LOAD_KEY = "brokerLoad.json";
```
tomncooper, line 39: Instead of having this key defined twice can you pull it from the KRAO class or better yet move it to one of (or a new) enum/class in the operator.common module (with the other CC keys)? That way if it ever changes in the future it will not cause test failures.

cluster-operator/src/test/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceStatusTest.java
```diff
@@ -0,0 +1,200 @@
...
+        JsonObject output = KafkaRebalanceAssemblyOperator.parseLoadStats(
+                loadBeforeArray, loadAfterArray);
+
+        assertThat(output.getMap(), hasKey("1"));
+
+        assertThat(output.getJsonObject("1").getMap(), hasKey(CruiseControlLoadParameters.REPLICAS.getKafkaRebalanceStatusKey()));
+
+        JsonObject replicas = output.getJsonObject("1").getJsonObject("replicas");
+
+        ArrayList<Integer> replicasList = new ArrayList<>();
```
tomncooper, line 124: Why do we need an Arraylist here? Why not just do: `assertThat(replicas.getInteger("before"), is(10));` below?

cluster-operator/src/test/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceStatusTest.java
```diff
@@ -0,0 +1,200 @@
...
+
+        assertThat(replicasList.get(0), is(10));
+        assertThat(replicasList.get(1), is(5));
+        assertThat(replicasList.get(2), is(-5));
+
+        assertThat(output.getJsonObject("1").getMap(), hasKey(CruiseControlLoadParameters.CPU_PERCENTAGE.getKafkaRebalanceStatusKey()));
+
+        JsonObject cpus = output.getJsonObject("1").getJsonObject("cpuPercentage");
+
+        ArrayList<Double> cpu = new ArrayList<>();
```
tomncooper, line 137: Same as for replicas, this ArrayList seems like a pointless step?

cluster-operator/src/test/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceStatusTest.java
```diff
@@ -0,0 +1,200 @@
...
+
+            Map<String, LinkedHashMap<String, String>> brokerLoadMap = mapper.readValue(brokerMap.get(BROKER_LOAD_KEY), LinkedHashMap.class);
+
+            assertThat(brokerMap, hasKey(BROKER_LOAD_KEY));
+
+            LinkedHashMap<String, LinkedHashMap<String, Object>> m = (LinkedHashMap) brokerLoadMap.get("1");
+
+            assertThat(m, hasKey(CruiseControlLoadParameters.CPU_PERCENTAGE.getKafkaRebalanceStatusKey()));
+
+            ArrayList<Double> cpu = new ArrayList<>();
```
tomncooper, line 175: Again why not just do: ` assertThat((Double) m.get("cpuPercentage").get("before"), is(10.0));`?

cluster-operator/src/test/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceStatusTest.java
```diff
@@ -0,0 +1,200 @@
...
+            cpu.add((Double) m.get("cpuPercentage").get("diff"));
+
+            assertThat(cpu.get(0), is(10.0));
+            assertThat(cpu.get(1), is(20.0));
+            assertThat(cpu.get(2), is(10.0));
+
+
+            assertThat(m, hasKey(CruiseControlLoadParameters.REPLICAS.getKafkaRebalanceStatusKey()));
+
+            ArrayList<Integer> replicas = new ArrayList<>();
```
tomncooper, line 187: Same comment as for other lists in this test.

documentation/modules/cruise-control/con-optimization-proposals.adoc
```diff
@@ -94,10 +101,45 @@ m¦excludedBrokersForReplicaMove
 
 |===
 
+
+[discrete]
+=== Broker load
+
+The broker load section lists all broker IDs and provides a number of metrics for each broker.
```
tomncooper, line 108: Can we make it clear that the broker load is a json string in a config map. This reads like it is still a _section_ of the status message.

tomncooper: It would also be nice to give an example command as to how a user might retrieve this load information json.

systemtest/src/test/java/io/strimzi/systemtest/cruisecontrol/CruiseControlIsolatedST.java
```diff
@@ -159,9 +159,11 @@ void testCruiseControlTopicExclusion(ExtensionContext extensionContext) {
 
         LOGGER.info("Checking status of KafkaRebalance");
         KafkaRebalanceStatus kafkaRebalanceStatus = KafkaRebalanceResource.kafkaRebalanceClient().inNamespace(NAMESPACE).withName(clusterName).get().getStatus();
-        assertThat(kafkaRebalanceStatus.getOptimizationResult().get("excludedTopics").toString(), containsString(excludedTopic1));
-        assertThat(kafkaRebalanceStatus.getOptimizationResult().get("excludedTopics").toString(), containsString(excludedTopic2));
-        assertThat(kafkaRebalanceStatus.getOptimizationResult().get("excludedTopics").toString(), not(containsString(includedTopic)));
+        Map<String, Object> optimizationResult = kafkaRebalanceStatus.getOptimizationResult();
+        Map<String, Object> optimizationSummary = (Map<String, Object>) optimizationResult.get("summary");
```
tomncooper, line 163: As commented above, currently the optimization result is not under an additional "summary" key and your code puts it under one, which it shouldn't. After you move it back to be the map itself you will need to change this code as well.

tomncooper (changes requested):
Left some comments and small changes that I think need to be made before we can merge this.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceAssemblyOperator.java
ShubhamRwt: Iam sorry, I guess it got skipped. I had resolved the ordering earlier. I have resolved it now

documentation/modules/cruise-control/con-optimization-proposals.adoc
```diff
@@ -34,7 +34,14 @@ Consider a shorter interval for fast changing clusters, although this increases
 [discrete]
 == Contents of optimization proposals
 
-The following table describes the properties contained in an optimization proposal:
+An optimization proposal consists of two main sections: the _summary_ and the _broker load_.
```
tomncooper, line 37: This still looks like the old status only PR, it should read something like:

"An optimization proposal consists of two main sections: the _summary_ which is stored in the status of the KafkaRebalance custom resource and the _broker load_ which is stored in a ConfigMap."

documentation/modules/cruise-control/con-optimization-proposals.adoc
```diff
@@ -94,10 +101,45 @@ m¦excludedBrokersForReplicaMove
 
 |===
 
+
+[discrete]
+=== Broker load
+
+The broker load is a json string which consist of all broker IDs and provides a number of metrics for each broker.
```
tomncooper, line 108: "The broker load is stored in a ConfigMap (with the same name as the KafakRebalance custom resource) as a JSON formatted string. This JSON string consists of a JSON object with keys for each broker IDs linking to a number of metrics for each broker."

documentation/modules/cruise-control/con-optimization-proposals.adoc
```diff
@@ -94,10 +101,45 @@ m¦excludedBrokersForReplicaMove
...
+Each metric consist of three values.
+The first is the metric value before the optimization proposal is applied, the second is the expected value of the metric after the proposal is applied, and the third is the difference between the first two values (after minus before).
+
+The following table explains the properties contained in the optimization proposal's broker load section:
```
tomncooper, line 112: "The following table explains the properties contained in the optimization proposal's broker load ConfigMap"

documentation/modules/cruise-control/con-optimization-proposals.adoc
tomncooper: Should we also include an example command of how to extract the JSON string from the config map using `kubectl`? 

WDYT @laidan6000?

documentation/modules/cruise-control/con-optimization-proposals.adoc
laidan6000, line 37:
```suggestion
An optimization proposal consists of two main sections: 

* The _summary_ is stored in the `status` of the `KafkaRebalance` resource.
* The _broker load_ is stored in a ConfigMap that contains data as a JSON string.
```

documentation/modules/cruise-control/con-optimization-proposals.adoc
laidan6000: Sounds like that would be useful.

To extract the JSON string from the ConfigMap, use the following command:

laidan6000 (changes requested):

ShubhamRwt:
Thanks for the review @tomncooper and @laidan6000. I will be thpushing the changes in some time.

documentation/modules/cruise-control/con-optimization-proposals.adoc
tomncooper: @ShubhamRwt can you add the `kubectl` command to extract the json?

laidan6000:
Please use a source block for the kubectl command and the proper formatting for a variable. Thanks!

Asciidoc markup:

```
To extract the JSON string from the ConfigMap, use the following command:

[source,shell,subs=+quotes]
----
kubectl describe configmap _MY-REBALANCE_
----
```

laidan6000:
Does the output of the command need to be piped to anything to make it more readable?

documentation/modules/cruise-control/con-optimization-proposals.adoc
ShubhamRwt: just the json right ? ```kubectl get configmap my-rebalance -o json```

ShubhamRwt:
> Does the output of the command need to be piped to anything to make it more readable?

Umm Iam not sure about that @laidan6000, we are using just going with json for now but later on we can add some tool which prettifies the json

laidan6000 (approved):
Looks good to me.

documentation/modules/cruise-control/con-optimization-proposals.adoc
```diff
@@ -94,10 +105,52 @@ m¦excludedBrokersForReplicaMove
 
 |===
 
+
+[discrete]
+=== Broker load
+
+The broker load is stored in a ConfigMap (with the same name as the KafakRebalance custom resource) as a JSON formatted string. This JSON string consists of a JSON object with keys for each broker IDs linking to a number of metrics for each broker
```
tomncooper, line 112: You are missing a full stop at the end of this sentence.

ShubhamRwt: I have added it now.

documentation/modules/cruise-control/con-optimization-proposals.adoc
```diff
@@ -94,10 +105,52 @@ m¦excludedBrokersForReplicaMove
...
+
+The broker load is stored in a ConfigMap (with the same name as the KafakRebalance custom resource) as a JSON formatted string. This JSON string consists of a JSON object with keys for each broker IDs linking to a number of metrics for each broker.
+Each metric consist of three values.
+The first is the metric value before the optimization proposal is applied, the second is the expected value of the metric after the proposal is applied, and the third is the difference between the first two values (after minus before).
+
+To extract the JSON string from the ConfigMap, use the following command:
+
+[source,shell,subs=+quotes]
+----
+kubectl get configmap _my-rebalance_ -o json | jq .data
```
tomncooper, line 120: We shouldn't use emphasis ( `_`) in a code example.

tomncooper: Also it would be nice to provide a jq example that spits out just the load json:
```
jq '.["data"]["brokerLoad.json"]'
```

tomncooper: When you do that the actual JSON is not parseable by `jq`. It contains a load of escape characters etc. This may be linked to the way we are encoding the string.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceAssemblyOperator.java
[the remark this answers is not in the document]
tomncooper: I think the use of `encodePrettily` is causing the final JSON string to be miss formatted. We just need a straight dump of the json here not a load of new lines added in.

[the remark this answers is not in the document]
tomncooper: I think changing this to just `.encode()` would be better.

documentation/modules/cruise-control/con-optimization-proposals.adoc
tomncooper: I think we shouldn't be using `encodePrettily` is should just be `encode`

tomncooper (changes requested):
I think we need to change the JSON encoding so that it is parseable by `jq` when extracted via `kubectl`

documentation/modules/cruise-control/con-optimization-proposals.adoc
ShubhamRwt: @tomncooper,Hello, Actually  ( _)  this change was suggested by @laidan6000 , mostly due to consistency If I remember correctly

ShubhamRwt: I have changed it to encode() now

tomncooper: Ok I tried this change and I think it should def be `encode`. The JSON can be parsed with `jq` using this command:
```
jq '.["data"]["brokerLoad.json"]|fromjson|.'
```
The user can then add their query onto the end of that string.

tomncooper: Is still wonder if it would have been possible to encode the data directly into the `ConfigMap` builder and avoid JSON all together. Then the user could convert the config map data with `-o json` or `-o yaml`.

ShubhamRwt: Okay, Got it. I will push these changes in some minutes

ShubhamRwt:
@tomncooper Hello,  Do we have more comments or sugesstions for this ?

documentation/modules/cruise-control/con-optimization-proposals.adoc
```diff
@@ -94,10 +105,52 @@ m¦excludedBrokersForReplicaMove
...
 
+
+[discrete]
+=== Broker load
+
+The broker load is stored in a ConfigMap (with the same name as the KafakRebalance custom resource) as a JSON formatted string. This JSON string consists of a JSON object with keys for each broker IDs linking to a number of metrics for each broker.
+Each metric consist of three values.
+The first is the metric value before the optimization proposal is applied, the second is the expected value of the metric after the proposal is applied, and the third is the difference between the first two values (after minus before).
+
+To extract the JSON string from the ConfigMap, use the following command:
```
tomncooper, line 116: `jq` is just one option for doing this and we don't list `jq` as something you *need* in order to use Strimzi. So I would change this to something like: 

> To extract the JSON string from the ConfigMap you could use the following command, which uses the `jq` command line JSON parser tool:

documentation/modules/cruise-control/con-optimization-proposals.adoc
```diff
@@ -94,10 +105,52 @@ m¦excludedBrokersForReplicaMove
...
+
+[source,shell,subs=+quotes]
+----
+kubectl get configmap _my-rebalance_ -o json | jq '.["data"]["brokerLoad.json"]|fromjson|.'
```
tomncooper, line 120: I am still not clear on why we are adding emphasis to text in a verbatim code example? When rendered, if I copy code would I get the markup as well?

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceAssemblyOperator.java
```diff
@@ -351,14 +358,16 @@ private String rebalanceStateConditionType(KafkaRebalanceStatus status) {
         RebalanceOptions.RebalanceOptionsBuilder rebalanceOptionsBuilder = convertRebalanceSpecToRebalanceOptions(kafkaRebalance.getSpec());
 
         return computeNextStatus(reconciliation, host, apiClient, kafkaRebalance, currentState, rebalanceAnnotation, rebalanceOptionsBuilder)
-           .compose(desiredStatus -> {
+           .compose(desiredStatusandMap -> {
```
tomncooper, line 361: It looks like you still haven't applied the change from `desiredStatusandMap` to `desiredStatusAndMap`?

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceAssemblyOperator.java
```diff
@@ -453,34 +452,230 @@ private KafkaRebalanceStatus buildRebalanceStatusFromPreviousStatus(KafkaRebalan
...
+            brokersStats.put(String.valueOf(loadBeforeEntry.getKey()), brokerStats);
+        }
+
+        return brokersStats;
+    }
+
+    /**
+     * A wrapper class containing used to bind the ConfigMap and the status together.
+     */
+
```
tomncooper, line 588: Can you get rid of this line?

tomncooper (changes requested):
A few nits and then I think we are there

documentation/modules/cruise-control/con-optimization-proposals.adoc
ShubhamRwt: I tried this but  rendering we are not getting the the markup, but before removing them I guess we should confirm it with @laidan6000 .

documentation/modules/cruise-control/con-optimization-proposals.adoc
laidan6000: Sounds good.

documentation/modules/cruise-control/con-optimization-proposals.adoc
laidan6000: The markup would not be copied. @ShubhamRwt - Please use capitals and hyphens, so that would be:

MY-REBALANCE

Apologies for the confusion.

ShubhamRwt:
@tomncooper I have made the changes.

tomncooper (approved):
LGTM. Great work @ShubhamRwt!

PR #4633, opened 2021-03-24
ST: Enable loadbalancer tests for aws and cover finalizer testing

Frawless:

### Type of change

- Enhancement / new feature

### Description

This PR add finalizer to loadbalancers, which is needed for proper deletion of loadbalancers on aws. 

### Checklist


scholzj (commented):
The code looks good. But I'm not sure we want to have it hardcoded since it is not supported everywhere. Maybe we should make it conditional based on some env var?

Frawless:
@scholzj finalizers are not supproted or that anno? Guess we can set finalizers to `null` or use anno base on condition right?

scholzj:
Well, it is not an annotation (so you should maybe rename the variable) ... it is a name of the finalizer. Finalizers in general work everywhere. But this particular `service.kubernetes.io/load-balancer-cleanup` finalizer works only on some environments which support it. So having it hardcoded and unconditional probably breaks other environments (such as my home cluster ;-))

scholzj (approved):

PR #4634, opened 2021-03-24
Generate the API Reference docs for `v1beta2` only

scholzj:
### Type of change

- Documentation

### Description

The Operators now support only Strimzi API version `v1beta2`. So lot of the fields which we had in `v1alpha1` or `v1beta1` cannot be used anymore. Therefore they should be removed from the API Reference. This can be done by switching the API version in the `DocGenerator` to `v1beta2`.


tombentley (approved):

laidan6000:
We should probably do some manual testing of incoming links to the schema reference after this is merged.

PR #4640, opened 2021-03-25
Fix doc about CC logging

sknot-rh:
Signed-off-by: Stanislav Knot <sknot@redhat.com>

### Type of change
- Documentation

### Description
Doc is mentioning deprecated property.

PR #4641, opened 2021-03-25
Test for CruiseControl rootlogger level change

michalxo:
Signed-off-by: Michal Tóth <mtoth@redhat.com>

Test which verifies functionality of CruiseControl logger level change.

- [x] Write tests
- [x] Update documentation #4640 

### DIFF
--- a/systemtest/src/test/java/io/strimzi/systemtest/log/LogSettingST.java
+++ b/systemtest/src/test/java/io/strimzi/systemtest/log/LogSettingST.java
@@ -7,6 +7,7 @@ package io.strimzi.systemtest.log;
 import io.fabric8.kubernetes.api.model.Container;
 import io.fabric8.kubernetes.api.model.EnvVar;
 import io.fabric8.kubernetes.api.model.Pod;
+import io.strimzi.api.kafka.model.InlineLogging;
 import io.strimzi.api.kafka.model.JvmOptions;
 import io.strimzi.api.kafka.model.JvmOptionsBuilder;
 import io.strimzi.api.kafka.model.KafkaBridgeResources;
@@ -36,10 +37,12 @@ import io.strimzi.systemtest.templates.crd.KafkaMirrorMakerTemplates;
 import io.strimzi.systemtest.templates.crd.KafkaTemplates;
 import io.strimzi.systemtest.templates.crd.KafkaTopicTemplates;
 import io.strimzi.systemtest.templates.crd.KafkaUserTemplates;
+import io.strimzi.systemtest.utils.StUtils;
 import io.strimzi.systemtest.utils.kubeUtils.controllers.DeploymentConfigUtils;
 import io.strimzi.systemtest.utils.kubeUtils.controllers.DeploymentUtils;
 import io.strimzi.systemtest.utils.kubeUtils.controllers.StatefulSetUtils;
 import io.strimzi.systemtest.utils.kubeUtils.objects.PodUtils;
+import io.strimzi.test.TestUtils;
 import io.strimzi.test.timemeasuring.Operation;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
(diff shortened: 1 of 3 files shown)


michalxo:
@strimzi-ci run tests test_only profile=regression testcase=LogSettingST

scholzj (approved):

sknot-rh (approved):

systemtest/src/test/java/io/strimzi/systemtest/log/LogSettingST.java
```diff
@@ -412,6 +417,39 @@ void testBridgeLogSetting(ExtensionContext extensionContext) {
...
+
+        try {
+            Thread.sleep(CO_OPERATION_TIMEOUT_SHORT);
+        } catch (InterruptedException e) {
+            e.printStackTrace();
+        }
+
+        LOGGER.info("Verify changed root logger in {}.", cruiseControlPodName);
+        containerLogLevel = cmdKubeClient().execInPod(cruiseControlPodName, "grep", "-i", "rootlogger.level", Constants.CRUISE_CONTROL_LOG_FILE_PATH).out().trim().split("=")[1];
+        assertThat(containerLogLevel.toUpperCase(Locale.ENGLISH), is(debugText.strip()));
```
im-konge, line 446: Wouldn't it be better to add this to `waitFor` and wait for `DEBUG` level to appear to the `rootlogger.level`? I guess it's better to wait for something to happen that just sleep for few seconds and then assert the result.

systemtest/src/test/java/io/strimzi/systemtest/log/LogSettingST.java
```diff
@@ -412,6 +417,39 @@ void testBridgeLogSetting(ExtensionContext extensionContext) {
...
+
+        LOGGER.info("Check cruise control logs in pod {} and it's container {} .", cruiseControlPodName, Constants.CRUISE_CONTROL_CONTAINER_NAME);
+        logOut = cmdKubeClient().namespace(cluster.getNamespace()).execInCurrentNamespace("logs", cruiseControlPodName, "-c", Constants.CRUISE_CONTROL_CONTAINER_NAME, "--since=20s").out();
+        assertThat(logOut, containsString(debugText));
```
im-konge, line 450: Well, when you want to check the log, I would use the `assumeThat` or `assumeTrue`, and then do the assertion -> in case that the log will not contain the `DEBUG` text, it will end without error - we don't want flaky tests, right?

systemtest/src/test/java/io/strimzi/systemtest/log/LogSettingST.java
im-konge: (look at tests in `LoggingChangeST` where these waits are)

michalxo: Good idea, I will change it tomorrow.

systemtest/src/test/java/io/strimzi/systemtest/log/LogSettingST.java
michalxo: Oh really? I thought it would fail with an error if debugText is not in logOut. I will check it out as well.
Thanks.

im-konge: Well I would at least try it, thanks :)

scholzj: @im-konge What is the point of the assert if you skip it  instead of letting it fail?

im-konge: We discussed offline that checking log is not good and we're trying to avoid it. I guess it's better (if this assert is only optional and we're checking the log change in the log4j file) to do the `assumeTrue` and then do some checks or remove the assert, because having flaky test is not wanted from my POV.

im-konge: Or add some `waitFor` for this

michalxo: So I read what `assumeThat|True` does and I think it does not think it fits well into this test. I am going with waitFor usage instead. ;-)

scholzj: @im-konge The thing is that if it is optional it doesn't really matter, or? I can understand that the tests with real logs can be flaky. However:
* What value does it have to check it if you ignore the result? You can probably just remove it in that case and save time at least, or?
* Without checking the actual logs - do we really need STs? Updating the ConfigMap resource can be quite easily tested in UTs. Updating the file in the Pod is probably tested quite well in Kube.

im-konge: Yeah I know, that without the log check it's useless to have these kind of tests. All I don't want is to have a flaky test. So if the `waitFor` will "fix" it, it's good. Thanks @scholzj and @michalxo

im-konge (approved):

michalxo:
@strimzi-ci run tests test_only profile=regression testcase=LogSettingST

systemtest/src/test/java/io/strimzi/systemtest/log/LogSettingST.java
```diff
@@ -412,6 +421,37 @@ void testBridgeLogSetting(ExtensionContext extensionContext) {
...
+        LOGGER.info("Check logs in CruiseControl - make sure no DEBUG is found there.");
+        String logOut = StUtils.getLogFromPodByTime(cruiseControlPodName, Constants.CRUISE_CONTROL_CONTAINER_NAME, "20s");
+        assertThat(logOut.toUpperCase(Locale.ENGLISH), not(containsString(debugText)));
+
+        InlineLogging logging = new InlineLogging();
+        logging.setLoggers(Collections.singletonMap("rootLogger.level", debugText.strip()));
+        KafkaResource.replaceKafkaResource(LOG_SETTING_CLUSTER_NAME, kafka -> kafka.getSpec().getCruiseControl().setLogging(logging));
+
+        LOGGER.info("Wait for change of root logger in {}.", cruiseControlPodName);
+        TestUtils.waitFor("Waiting for log to be changed", Duration.ofSeconds(5).toMillis(), CO_OPERATION_TIMEOUT_MEDIUM, () -> {
```
Frawless, line 443: Maybe you can use some constant?

Frawless (approved):
Just one nit

systemtest/src/test/java/io/strimzi/systemtest/log/LogSettingST.java
michalxo: `Duration.ofSeconds(5).toMillis()` but the name would be inappropriate.. I searched for them, but I did not like any :-)

Frawless: so create new one?

michalxo:
@strimzi-ci run tests test_only profile=regression testcase=LogSettingST

systemtest/src/test/java/io/strimzi/systemtest/log/LogSettingST.java
michalxo: done.

michalxo:
@strimzi-ci run tests test_only profile=regression testcase=LogSettingST

PR #4644, opened 2021-03-25
Removed rm line in config-model-generator Makefile

npecka:
### Type of change

- Bugfix
- Refactoring

### Description

Removed line in config-model-generator Makefile as it seemed unnecessary when running overall `make clean` and it deleted various files as shown below. After the change these files still exist after `make clean` is run. Maybe this line should be changed to a different spot within the project?

        deleted:    cluster-operator/src/main/resources/kafka-2.6.0-config-model.json
        deleted:    cluster-operator/src/main/resources/kafka-2.6.1-config-model.json
        deleted:    cluster-operator/src/main/resources/kafka-2.7.0-config-model.json

### Checklist


scholzj:
I do not think this is the right way because it will mean we will have more and more models over time. The solution is to not have these committed in the first place and having them on .gitignore.

scholzj (commented):
I think you still need to delete the 3 existing files otherwise we will keep them forever.

scholzj (approved):

Frawless (approved):

sknot-rh (approved):

PR #4645, opened 2021-03-25
Resolve minor build and linting issues, rely on server validation for IT CRDs

MikeEdgar:
### Type of change

- Refactoring

### Description

While setting up Strimzi for local build and IDE import, I came across several warnings and errors (related to generating the tar.gz for api-conversion). I'm still working through my local build/configuration, but I want to open this PR to get feedback/comments sooner rather than later in case changes are needed/desired.

- Update .gitignore for sub-module Eclipse/Maven output & configuration
- Use tarLongFileMode=posix for maven-assembly-plugin to accommodate larger group ID values (see https://maven.apache.org/plugins/maven-assembly-plugin/faq.html#tarFileModes) that may occur in local builds
- Specify generic type parameter to resolve Eclipse JDT linting error
- Log message of caught exception in `TestUtils#waitFor`
- Disable local validation for CRD integration tests 
- Remove specification of dependency/plugin versions when duplicate of managed artifact
- Move `combine.self="override"` within plugin `<configuration>` element to the applicable option
- Replace deprecated `pom.basedir` with `project.basedir`

### Checklist

- [ ] Write tests
- [ ] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/api-conversion/src/main/java/io/strimzi/kafka/api/conversion/converter/KafkaConverter.java
+++ b/api-conversion/src/main/java/io/strimzi/kafka/api/conversion/converter/KafkaConverter.java
@@ -72,9 +72,9 @@ public class KafkaConverter extends Converter<Kafka> {
         Conversion.replaceLogging("/spec/entityOperator/topicOperator/logging", "log4j2.properties"),
         Conversion.replaceLogging("/spec/entityOperator/userOperator/logging", "log4j2.properties"),
         Conversion.replaceLogging("/spec/cruiseControl/logging", "log4j2.properties"),
-        new MetricsConversion<>("/spec/kafka", KafkaClusterSpec.class, "kafka"),
-        new MetricsConversion<>("/spec/zookeeper", ZookeeperClusterSpec.class, "zookeeper"),
-        new MetricsConversion<>("/spec/cruiseControl", CruiseControlSpec.class, "cruise-control")
+        new MetricsConversion<Kafka>("/spec/kafka", KafkaClusterSpec.class, "kafka"),
+        new MetricsConversion<Kafka>("/spec/zookeeper", ZookeeperClusterSpec.class, "zookeeper"),
+        new MetricsConversion<Kafka>("/spec/cruiseControl", CruiseControlSpec.class, "cruise-control")
     );
 
     public KafkaConverter() {
--- a/config-model-generator/pom.xml
+++ b/config-model-generator/pom.xml
@@ -24,12 +24,10 @@
         <dependency>
             <groupId>org.apache.kafka</groupId>
             <artifactId>kafka_2.12</artifactId>
-            <version>${kafka-metadata-version}</version>
         </dependency>
         <dependency>
             <groupId>org.apache.kafka</groupId>
             <artifactId>kafka-clients</artifactId>
-            <version>${kafka-metadata-version}</version>
         </dependency>
         <dependency>
             <groupId>com.fasterxml.jackson.core</groupId>
--- a/config-model-generator/pom.xml
+++ b/config-model-generator/pom.xml
@@ -16,6 +16,22 @@
         <config-model-file>../cluster-operator/src/main/resources/kafka-${kafka-metadata-version}-config-model.json</config-model-file>
     </properties>
 
+    <dependencyManagement>
+        <!-- Manage versions to allow for override via the CLI with system property. -->
+        <dependencies>
+            <dependency>
+                <groupId>org.apache.kafka</groupId>
+                <artifactId>kafka_2.12</artifactId>
+                <version>${kafka-metadata-version}</version>
+            </dependency>
+            <dependency>
+                <groupId>org.apache.kafka</groupId>
+                <artifactId>kafka-clients</artifactId>
+                <version>${kafka-metadata-version}</version>
+            </dependency>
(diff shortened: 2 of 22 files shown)


config-model-generator/pom.xml
```diff
@@ -24,12 +24,10 @@
         <dependency>
             <groupId>org.apache.kafka</groupId>
             <artifactId>kafka_2.12</artifactId>
-            <version>${kafka-metadata-version}</version>
```
scholzj, line 27: I think these were definitely intentionally there.

config-model-generator/pom.xml
```diff
@@ -24,12 +24,10 @@
...
         </dependency>
         <dependency>
             <groupId>org.apache.kafka</groupId>
             <artifactId>kafka-clients</artifactId>
-            <version>${kafka-metadata-version}</version>
```
scholzj, line 32: Same as above.

config-model-generator/pom.xml
MikeEdgar: Thanks for spotting this - I added a management section in this module with a comment to support the overrides, let me know what you think about that approach. Otherwise, I'm happy to restore this to the original and ignore Eclipse's warnings if it's too much :)

scholzj (approved):
LGTM if it works.

test/src/main/java/io/strimzi/test/TestUtils.java
```diff
@@ -125,6 +125,7 @@ public static long waitFor(String description, long pollIntervalMs, long timeout
             try {
                 result = ready.getAsBoolean();
             } catch (Exception e) {
+                LOGGER.info("Exception waiting for {}, {}", description, e.getMessage());
```
Frawless, line 128: I am not sure if this is a good solution. We had discussion around it and this exception should be printed only when the wait fails. This info log can easilly log a lot of same lines without any useful info from it. I think we maybe should try to execute same code again on line 136 (before `onTimeout.run()`) and print the exception here.

MikeEdgar: I see your point about logging a lot of lines and I agree. The situation I encountered was that in my initial setup `kubectl` was not found and therefore the process could never be started. The exception message clearly indicated what was wrong, but based on the build output on the console, the process seemed to be stuck. Ideally for situations like that, a user would be notified for fatal errors and be able to cancel the build, etc.

What are your thoughts on a solution where the message is logged only the first time it is encountered?

ppatierno (approved):
Same as Jakub, I am OK if it works.

test/src/main/java/io/strimzi/test/TestUtils.java
Frawless: Ye sounds like a good option. The first and last exception on the info log level would be great.

scholzj:
@Frawless Is this ok for you now?
@tombentley I thought that it might be useful if you could do a quick review as well.

Frawless (approved):
Fine by me, thanks!

api-conversion/src/main/java/io/strimzi/kafka/api/conversion/converter/KafkaConverter.java
```diff
@@ -72,9 +72,9 @@ public ArrayOrObjectKafkaListeners apply(ArrayOrObjectKafkaListeners arrayOrObje
         Conversion.replaceLogging("/spec/entityOperator/topicOperator/logging", "log4j2.properties"),
         Conversion.replaceLogging("/spec/entityOperator/userOperator/logging", "log4j2.properties"),
         Conversion.replaceLogging("/spec/cruiseControl/logging", "log4j2.properties"),
-        new MetricsConversion<>("/spec/kafka", KafkaClusterSpec.class, "kafka"),
-        new MetricsConversion<>("/spec/zookeeper", ZookeeperClusterSpec.class, "zookeeper"),
-        new MetricsConversion<>("/spec/cruiseControl", CruiseControlSpec.class, "cruise-control")
+        new MetricsConversion<Kafka>("/spec/kafka", KafkaClusterSpec.class, "kafka"),
+        new MetricsConversion<Kafka>("/spec/zookeeper", ZookeeperClusterSpec.class, "zookeeper"),
+        new MetricsConversion<Kafka>("/spec/cruiseControl", CruiseControlSpec.class, "cruise-control")
```
tombentley, line 77: What was the problem here?

tombentley (approved):
Just a question

api-conversion/src/main/java/io/strimzi/kafka/api/conversion/converter/KafkaConverter.java
MikeEdgar: Eclipse Java development tools flagged the assignment of `V1BETA1_TO_V1BETA2` with this error:
> `Type mismatch: cannot convert from VersionConversion<HasMetadata> to VersionConversion<Kafka>`

The added type parameter gives JDT an extra hint. I saw this error in both Eclipse itself as well as VSCode with the Java tools which I believe are based on JDT. I'd be curious to hear if others are using different IDEs like IntelliJ or older versions of Eclipse. This was seen with the current Eclipse release.

scholzj:
Thanks for the PR.

PR #4654, opened 2021-03-26
[MO] - additional logging + check emptiness

see-quick:

### Type of change

_Select the type of your PR_

- Enhancement / new feature

### Description

This PR adding logging to `TestKafkaVersion.class` and with that check for the emptiness of supported Kafka. 

Why?

We once got into a situation where our CI tests could spin up Kafka cluster and there was an NPE. This problem was not simply found at all. In the end, we found out that it was because the Kafka version of the map was empty. Thanks to this check, the problem will be eliminated.

### Checklist


im-konge (approved):

Frawless (approved):

scholzj:
I don't necessarily have problem with this, but I'm just curious ... what is the reason for this? We do not have the kafka versions file every empty, or?

scholzj (approved):

PR #4665, opened 2021-03-29
Update VertX version to 4.0.3

sknot-rh:

### Type of change
- Dependency update

### Description
The currently used version 3.9.1 seems to be quite obsolete. Updating to 4.0.3.

The netty version changed in vert.x, so the API requests had to be adjusted.

Also fasterxm versions updated with one excedption. `fasterxml.jackson-dataformat.version` has to be with older version as there is an issue https://github.com/FasterXML/jackson-dataformats-text/pull/246

### DIFF
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaConnectApiImpl.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaConnectApiImpl.java
@@ -18,6 +18,7 @@ import io.vertx.core.Promise;
 import io.vertx.core.Vertx;
 import io.vertx.core.buffer.Buffer;
 import io.vertx.core.http.HttpClientOptions;
+import io.vertx.core.http.HttpMethod;
 import io.vertx.core.json.JsonArray;
 import io.vertx.core.json.JsonObject;
 import org.apache.logging.log4j.LogManager;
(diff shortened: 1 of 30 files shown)


scholzj (approved):
Assuming it works ... 🤞

ppatierno (approved):
LGTM if all tests passed. Right now this version still has support for async callback but at some point I guess it will be deprecated in favour of using returned `Future` and maybe we should do some changes.

pom.xml
```diff
@@ -136,7 +136,7 @@
         <junit-platform-surefire-provider.version>1.3.2</junit-platform-surefire-provider.version>
         <opentest4j.version>1.2.0</opentest4j.version>
         <gson.version>2.8.2</gson.version>
-        <vertx.kafka.client>3.9.1</vertx.kafka.client>
+        <vertx.kafka.client>4.0.3</vertx.kafka.client>
         <exec-maven-plugin.version>1.6.0</exec-maven-plugin.version>
         <netty.version>4.1.60.Final</netty.version>
```
scholzj, line 141: Could you please try to bump Netty to 4.1.62.Final and see if it works with Vert.x 4.0.3?

sknot-rh: Does not work even with 4.1.63.Final.

scholzj: So ... does it work at least with the 4.1.60 which is default in VErt.x 4.0.3?

sknot-rh: Yes

scholzj: Then I guess that is the way to go ... and hopefully Vert.x release another fix later to work with the fixed Netty 🤷‍♂️

scholzj (approved):

ppatierno (approved):

tombentley (approved):

scholzj:
@sknot-rh Could you please look at the conflict / rebase it?

scholzj:
@sknot-rh I'm really sorry ... it looks like the the Pr from Tom Cooper created some more conflicts :-/

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/resource/cruisecontrol/CruiseControlApiImpl.java
```diff
@@ -105,68 +112,77 @@ public CruiseControlApiImpl(Vertx vertx, int idleTimeout) {
...
-                        JsonObject json = buffer.toJsonObject();
-                        if (json.containsKey(CC_REST_API_ERROR_KEY)) {
-                            // If there was a client side error, check whether it was due to not enough data being available
-                            if (json.getString(CC_REST_API_ERROR_KEY).contains("NotEnoughValidWindowsException")) {
-                                CruiseControlRebalanceResponse ccResponse = new CruiseControlRebalanceResponse(userTaskID, json);
-                                ccResponse.setNotEnoughDataForProposal(true);
-                                result.complete(ccResponse);
+            httpClient.request(HttpMethod.POST, port, host, path, request -> {
+                if (request.succeeded()) {
+                    if (idleTimeout != HTTP_DEFAULT_IDLE_TIMEOUT_SECONDS) {
```
tomncooper, line 117: I might be wrong but here you seem to be setting the idle timeout for a request that has already suceeded and returned?

tomncooper: The httpClient.request docs say that the method:
Create an HTTP request to send to the server at the default host and port. The handler is called when the request is ready to be sent.

So this might be right, I am just confused as to what `request.succeeded()` means here?

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/resource/cruisecontrol/CruiseControlApiImpl.java
```diff
@@ -105,68 +112,77 @@ public CruiseControlApiImpl(Vertx vertx, int idleTimeout) {
...
+                        request.result().setTimeout(idleTimeout * 1000);
+                    }
+
+                    if (userTaskId != null) {
+                        request.result().putHeader(CC_REST_API_USER_ID_HEADER, userTaskId);
+                    }
+                    request.result().end();
```
tomncooper, line 124: Doesn't `.end()` send the request? The [javadoc](https://vertx.io/docs/apidocs/io/vertx/core/http/HttpClientRequest.html#end--) says that "Once the request has ended, it cannot be used any more,"?

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/resource/cruisecontrol/CruiseControlApiImpl.java
sknot-rh: The last argument of `request` method is `Handler<AsyncResult<HttpClientRequest>>` so it's result is a `HttpClientRequest` object.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/resource/cruisecontrol/CruiseControlApiImpl.java
sknot-rh: Good catch.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/resource/cruisecontrol/CruiseControlApiImpl.java
tomncooper: Ah right it is actually the `AsyncResult` that contains the request you are calling `.suceeded()` on. Ok.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaConnectApiImpl.java
```diff
@@ -63,34 +64,41 @@ public KafkaConnectApiImpl(Vertx vertx) {
...
+                    result.fail(request.cause());
                 }
-            })
-            .exceptionHandler(result::tryFail)
-            .setFollowRedirects(true)
-            .putHeader("Accept", "application/json")
-            .putHeader("Content-Type", "application/json")
-            .putHeader("Content-Length", String.valueOf(data.length()))
-            .write(data)
-            .end());
```
tomncooper, line 93: In some places you have called `request.result().end();` and others like here you don't seem to `.end()` the request? I think you need (to be safe) to explicitly end all the requests. Unless vertx 4 does things differently?

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/resource/cruisecontrol/CruiseControlApiImpl.java
sknot-rh: From this https://access.redhat.com/documentation/en-us/red_hat_build_of_eclipse_vert.x/4.0/pdf/eclipse_vert.x_4.0_migration_guide/Red_Hat_build_of_Eclipse_Vert.x-4.0-Eclipse_Vert.x_4.0_Migration_Guide-en-US.pdf
 I understand the `.end()` is not needed anymore.

PR #4690, opened 2021-03-31
Preserve unknown fields for lists of maps

scholzj:
### Type of change

- Bugfix

### Description

When using Maps in CRDs with `apiextensions/v1`, `x-kubernetes-preserve-unknown-fields` needs to be set to `true`. The CRD generator currently does that for maps, but not for lists of maps. This PR fixes it.

Without this, the CR will not store the MM2 connector data. And as a result when MM2 is deployed, it will be reconciled in a never ending loop because the status stored by Kube is never the same as what the operator updates it to (because Kube removes the unstructured data). Once the preservation of unknown fields is properly enabled, the problem seems to disappear.

### Checklist

### DIFF
--- a/crd-generator/src/main/java/io/strimzi/crdgenerator/CrdGenerator.java
+++ b/crd-generator/src/main/java/io/strimzi/crdgenerator/CrdGenerator.java
@@ -867,7 +867,7 @@ public class CrdGenerator {
                 || long.class.equals(elementType)) {
             itemResult.put("type", "integer");
         } else if (Map.class.equals(elementType)) {
-            preserveUnknownFields(itemResult, elementType);
+            preserveUnknownFields(itemResult);
             itemResult.put("type", "object");
         } else  {
             buildObjectSchema(crApiVersion, itemResult, elementType, true, description);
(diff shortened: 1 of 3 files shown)


crd-generator/src/main/java/io/strimzi/crdgenerator/CrdGenerator.java
```diff
@@ -867,6 +867,7 @@ private ObjectNode buildArraySchema(ApiVersion crApiVersion, Property property,
                 || long.class.equals(elementType)) {
             itemResult.put("type", "integer");
         } else if (Map.class.equals(elementType)) {
+            preserveUnknownFields(itemResult, elementType);
```
tombentley, line 870: At this call site we know from the line above that the element type is map, but at the call site on line 885 we don't and you've put the check on the element type within `preserveUnknownFields()`. It would probably be better to keep the check outside the method. There will be no change at this call site, but and if wrapping the call site on 885.

tombentley (approved):
One nit, otherwise LGTM.

Issue #4627, opened 2021-03-24
Remove support for externalTrafficPolicy and loadBalancerSourceRanges from template in CO


scholzj:
`v1beta2` Strimzi API does not support the `externalTrafficPolicy` and `loadBalancerSourceRanges` from the `externalBootstrapService` and `perPodService` templates. They can now be configured only in the listener configuration (this is changed in order to make it configurable _per listener_). Therefore we should remove any use of these legacy fields from the CO. 

_(The fields should still remain present in the `api` module for the time being)_

PR #4692, opened 2021-03-31
Removed stale code linked to v1beta1 in kafkaCluster.java

npecka:
### Type of change

- Refactoring

### Description

externalTrafficPolicy and loadBalancingSourceRanges original input methods have been discontinued in favor of listener configurations in v1beta2 to make them per listener. Removed stale code from kafkaCluster.java and reconfigured unit tests with changes in mind.

Resolves --> #4627 

### Checklist

- [x] Write tests
- [ ] Update documentation
- [x] Update CHANGELOG.md

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
 * Support for configuring custom Authorizer implementation 
 * Changed Reconciliation interval for Topic Operator from 90 to 120 seconds (to keep it the same as for other operators)
 * Changed Zookeeper session timeout default value to 18 seconds for Topic and User Operators (for improved resiliency)
+* Removed stale code for externalTrafficPolicy and loadBalanceRangeSources from v1beta1 templates
 
 ### Changes, deprecations and removals
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,6 @@
 * Support for configuring custom Authorizer implementation 
 * Changed Reconciliation interval for Topic Operator from 90 to 120 seconds (to keep it the same as for other operators)
 * Changed Zookeeper session timeout default value to 18 seconds for Topic and User Operators (for improved resiliency)
-* Removed stale code for externalTrafficPolicy and loadBalanceRangeSources from v1beta1 templates
 
 ### Changes, deprecations and removals
 
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaCluster.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaCluster.java
@@ -517,40 +517,6 @@ public class KafkaCluster extends AbstractModel {
                 result.templateHeadlessServiceAnnotations = template.getBrokersService().getMetadata().getAnnotations();
             }
 
-            if (template.getExternalBootstrapService() != null) {
-                if (template.getExternalBootstrapService().getMetadata() != null) {
-                    result.templateExternalBootstrapServiceLabels = template.getExternalBootstrapService().getMetadata().getLabels();
-                    result.templateExternalBootstrapServiceAnnotations = template.getExternalBootstrapService().getMetadata().getAnnotations();
-                }
-
-                result.templateExternalBootstrapServiceTrafficPolicy = template.getExternalBootstrapService().getExternalTrafficPolicy();
-
-                if (result.isExposedWithLoadBalancer()) {
-                    result.templateExternalBootstrapServiceLoadBalancerSourceRanges = template.getExternalBootstrapService().getLoadBalancerSourceRanges();
-                } else if (template.getExternalBootstrapService().getLoadBalancerSourceRanges() != null
-                        && template.getExternalBootstrapService().getLoadBalancerSourceRanges().size() > 0) {
-                    // LoadBalancerSourceRanges have been set, but LaodBalancers are not used
-                    log.warn("The Kafka.spec.kafka.template.externalBootstrapService.loadBalancerSourceRanges option can be used only with load balancer type listeners");
-                }
-            }
-
-            if (template.getPerPodService() != null) {
-                if (template.getPerPodService().getMetadata() != null) {
-                    result.templatePerPodServiceLabels = template.getPerPodService().getMetadata().getLabels();
-                    result.templatePerPodServiceAnnotations = template.getPerPodService().getMetadata().getAnnotations();
-                }
-
-                result.templatePerPodServiceTrafficPolicy = template.getPerPodService().getExternalTrafficPolicy();
-
-                if (result.isExposedWithLoadBalancer()) {
-                    result.templatePerPodServiceLoadBalancerSourceRanges = template.getPerPodService().getLoadBalancerSourceRanges();
-                } else if (template.getPerPodService().getLoadBalancerSourceRanges() != null
-                        && template.getPerPodService().getLoadBalancerSourceRanges().size() > 0) {
-                    // LoadBalancerSourceRanges have been set, but LoadBalancers are not used
-                    log.warn("The Kafka.spec.kafka.template.perPodService.loadBalancerSourceRanges option can be used only with load balancer type listeners");
-                }
-            }
-
             if (template.getExternalBootstrapRoute() != null && template.getExternalBootstrapRoute().getMetadata() != null) {
                 result.templateExternalBootstrapRouteLabels = template.getExternalBootstrapRoute().getMetadata().getLabels();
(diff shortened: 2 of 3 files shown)


cluster-operator/src/test/java/io/strimzi/operator/cluster/model/KafkaClusterTest.java
```diff
@@ -896,12 +896,12 @@ public void testLoadBalancerExternalTrafficPolicyLocalFromTemplate() {
 
         // Check external bootstrap service
         Service ext = kc.generateExternalBootstrapServices().get(0);
-        assertThat(ext.getSpec().getExternalTrafficPolicy(), is(ExternalTrafficPolicy.LOCAL.toValue()));
+        assertThat(ext.getSpec().getExternalTrafficPolicy(), is(nullValue()));
```
scholzj, line 899: I guess you can just remove these test since it was there just test this feature. Just make sure we have the tests for the current was how to set it.

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaCluster.java
```diff
@@ -517,40 +517,6 @@ public static KafkaCluster fromCrd(Kafka kafkaAssembly, KafkaVersion.Lookup vers
...
-
-                if (result.isExposedWithLoadBalancer()) {
-                    result.templatePerPodServiceLoadBalancerSourceRanges = template.getPerPodService().getLoadBalancerSourceRanges();
-                } else if (template.getPerPodService().getLoadBalancerSourceRanges() != null
-                        && template.getPerPodService().getLoadBalancerSourceRanges().size() > 0) {
-                    // LoadBalancerSourceRanges have been set, but LoadBalancers are not used
-                    log.warn("The Kafka.spec.kafka.template.perPodService.loadBalancerSourceRanges option can be used only with load balancer type listeners");
-                }
-            }
-
```
scholzj, line 553: Are you sure we can remove the whole block? IIRC we did not deprecated / removed the metadata part, or?

npecka: No you are right. I brought back the changes in the latest commit while still removing the non metadata parts related to the PR.

kyguy (approved):

CHANGELOG.md
```diff
@@ -12,6 +12,7 @@
 * Support for configuring custom Authorizer implementation 
 * Changed Reconciliation interval for Topic Operator from 90 to 120 seconds (to keep it the same as for other operators)
 * Changed Zookeeper session timeout default value to 18 seconds for Topic and User Operators (for improved resiliency)
+* Removed stale code for externalTrafficPolicy and loadBalanceRangeSources from v1beta1 templates
```
scholzj, line 15: I do not think we need this. We normally put into the CHANGELOG.md only things which impact users which this doesn't really, it is just internal cleanup.

cluster-operator/src/test/java/io/strimzi/operator/cluster/model/KafkaClusterTest.java
```diff
@@ -1130,48 +1054,6 @@ public void testLoadBalancerSourceRangeFromListener() {
...
-
-        // Check per pod services
-        for (int i = 0; i < replicas; i++)  {
-            Service srv = kc.generateExternalServices(i).get(0);
-            assertThat(srv.getSpec().getLoadBalancerSourceRanges(), is(sourceRanges));
-        }
-    }
-
     @Test
     public void testLoadBalancerSourceRangeConflict() {
```
scholzj, line 1058: I think you can remove this test as well since the conflict does not exist anymore, or?

cluster-operator/src/test/java/io/strimzi/operator/cluster/model/KafkaClusterTest.java
```diff
@@ -979,44 +941,6 @@ public void testLoadBalancerExternalTrafficPolicyClusterFromListener() {
...
-
-        // Check per pod services
-        for (int i = 0; i < replicas; i++)  {
-            Service srv = kc.generateExternalServices(i).get(0);
-            assertThat(srv.getSpec().getExternalTrafficPolicy(), is(ExternalTrafficPolicy.CLUSTER.toValue()));
-        }
-    }
-
     @Test
     public void testLoadBalancerExternalTrafficPolicyClusterConflict() {
```
scholzj, line 945: I think you can remove this test as well since the conflict does not exist anymore, or?

cluster-operator/src/test/java/io/strimzi/operator/cluster/model/KafkaClusterTest.java
```diff
@@ -867,44 +867,6 @@ public void testLoadBalancerExternalTrafficPolicyLocalFromListener() {
...
-
-        // Check per pod services
-        for (int i = 0; i < replicas; i++)  {
-            Service srv = kc.generateExternalServices(i).get(0);
-            assertThat(srv.getSpec().getExternalTrafficPolicy(), is(ExternalTrafficPolicy.LOCAL.toValue()));
-        }
-    }
-
     @Test
     public void testLoadBalancerExternalTrafficPolicyLocalConflict() {
```
scholzj, line 871: I think you can remove this test as well since the conflict does not exist anymore, or?

scholzj (commented):
Some more nits, but LGTM otherwise.

cluster-operator/src/test/java/io/strimzi/operator/cluster/model/KafkaClusterTest.java
npecka: I think this test is still needed as it's part of the `testLoadBalancerSourcesRangeFromListener` and we still wanted to make sure the source ranges exist with the new configuration correct?

cluster-operator/src/test/java/io/strimzi/operator/cluster/model/KafkaClusterTest.java
npecka: I think this test is still needed as it's part of the `testLoadBalancerExternalTrafficPolicyClusterFromListener` and we still wanted to make sure the policy exists with the new configuration correct?

cluster-operator/src/test/java/io/strimzi/operator/cluster/model/KafkaClusterTest.java
npecka: I think this test is still needed as it's part of the `testLoadBalancerExternalTrafficPolicyLocalFromListener` and we still wanted to make sure the policy exists with the new configuration correct?

scholzj:
@npecka We seem to have `testLoadBalancerExternalTrafficPolicyClusterFromListener`, `testLoadBalancerExternalTrafficPolicyLocalFromListener` and `testLoadBalancerSourceRangeFromListener`. These seem to test the configuration through the listener section. The `*Conflict` seem to only test specifically what happened when both ways were configured at the same time. But that should not happen anymore. So why do we need those tests? Did I missed something?

npecka:
> @npecka We seem to have `testLoadBalancerExternalTrafficPolicyClusterFromListener`, `testLoadBalancerExternalTrafficPolicyLocalFromListener` and `testLoadBalancerSourceRangeFromListener`. These seem to test the configuration through the listener section. The `*Conflict` seem to only test specifically what happened when both ways were configured at the same time. But that should not happen anymore. So why do we need those tests? Did I missed something?

@scholzj Ah I understand. I misread and thought it was strictly testing for the configuration and not both being configured at the same time. So if we omit these tests it doesn't look like there are current ones that test if the listener config is set properly for externalTrafficPolicy and loadBalancerSourceRanges. Would it be appropriate to keep these tests and change the assert section to something like these `assertThat(kc.getListeners().get(0).getConfiguration().getExternalTrafficPolicy(), is(ExternalTrafficPolicy.CLUSTER));` and `assertThat(kc.getListeners().get(0).getConfiguration().getLoadBalancerSourceRanges(), is(sourceRanges));` ? Or do you think we should just do away with all of them all together?

scholzj:
The Three tests I commented on are testing the conflicts -> those should be deleted. The three tests I named here:

> `testLoadBalancerExternalTrafficPolicyClusterFromListener`, `testLoadBalancerExternalTrafficPolicyLocalFromListener` and `testLoadBalancerSourceRangeFromListener`

Are testing the fucntionality still used. So if you keep those that should be IMHO enough.

npecka:
@scholzj Ah I see now. Thank you for the clarification, I didn't notice there were the conflicts and then the ones without. I have gone ahead and deleted the conflicts in favor of the non conflict tests.

scholzj (approved):

ppatierno (approved):

Issue #4694, opened 2021-04-01
topic-store-changelog created with min.isr of 2 but replication factor 1


cablespaghetti:
**Describe the bug**
The `__strimzi-topic-operator-kstreams-topic-store-changelog` topic was created when installing 0.22.1 (on top of old volumes from 0.19 which might make a difference...) the topic was created with min.isr of 2 but replication factor 1.  

```
[kafka@kafka-kafka-0 bin]$ ./kafka-topics.sh --bootstrap-server localhost:9092 --describe --topic __strimzi-topic-operator-kstreams-topic-store-changelog
Topic: __strimzi-topic-operator-kstreams-topic-store-changelog	PartitionCount: 1	ReplicationFactor: 1	Configs: min.insync.replicas=2,cleanup.policy=compact,segment.bytes=67108864,message.format.version=2.7-IV2,min.compaction.lag.ms=0,message.timestamp.type=CreateTime
	Topic: __strimzi-topic-operator-kstreams-topic-store-changelog	Partition: 0	Leader: 1	Replicas: 1	Isr: 1
```
We're running 3 Kafka nodes with this config:
```
    config:
      auto.create.topics.enable: "false"
      offsets.topic.replication.factor: 3
      transaction.state.log.replication.factor: 3
      transaction.state.log.min.isr: 2
      default.replication.factor: 3
      replica.fetch.max.bytes: 10485760
      min.insync.replicas: 2
```

**Expected behavior**
The topic to be created with the default isr and replication factor of the cluster.

**Environment (please complete the following information):**
 - Strimzi version: 0.22.1
 - Installation method: Helm Chart
 - Kubernetes cluster: Kubernetes 1.19
 - Infrastructure: Amazon EKS

I'm happy to supply more information and try further deployments if this isn't easy to reproduce. Thanks

alesj:
This topic is created internally by Kafka Streams (for our Kafka Streams store), and we don't provide any defaults for it.

```
   StreamsConfig ...

    public static final String REPLICATION_FACTOR_CONFIG = "replication.factor";
    private static final String REPLICATION_FACTOR_DOC = "The replication factor for change log topics and repartition topics created by the stream processing application.";

            .define(REPLICATION_FACTOR_CONFIG,
                    Type.INT,
                    1,
                    Importance.HIGH,
                    REPLICATION_FACTOR_DOC)
```

We need to find a good way for the users to provide various properties ... as we're currently limited to

```
[...]

PR #4697, opened 2021-04-02
Allow for additional Kafka (Streams) properties - #4694.

alesj:
Signed-off-by: Ales Justin <ales.justin@gmail.com>

### Type of change

- Enhancement / new feature

### Description

Add a way to specify additional Kafka client / Streams properties.

### Checklist

- [X] Write tests
- [ ] Update documentation -- @PaulRMellor this needs to be added ... if/when this gets merged

### DIFF
--- a/topic-operator/src/main/java/io/strimzi/operator/topic/Config.java
+++ b/topic-operator/src/main/java/io/strimzi/operator/topic/Config.java
@@ -126,6 +126,9 @@ public class Config {
 
     public static final String TC_USE_ZOOKEEPER_TOPIC_STORE = "STRIMZI_USE_ZOOKEEPER_TOPIC_STORE";
 
+    public static final String TC_ADDITIONAL_KAFKA_PROPERTIES = "STRIMZI_ADDITIONAL_KAFKA_PROPERTIES";
+    public static final String TC_ADDITIONAL_KAFKA_PROPERTIES_DELIMITER = "STRIMZI_ADDITIONAL_KAFKA_PROPERTIES_DELIMITER";
+
     private static final Map<String, Value<?>> CONFIG_VALUES = new HashMap<>();
 
     /** A comma-separated list of key=value pairs for selecting Resources that describe topics. */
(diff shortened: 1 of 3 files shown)


scholzj:
Well, if this is the only solution, you would also need to add support for this to the Cluster Operator and the Kafka CR. But I'm not sure I understand the logic ... who creates the topic? Is it us or Kafka Streams API? Who decides that the topic should have 1 replica only? Because whoever does this is IMHO responsible for setting the other relevant options as well. So if it is Kafka Streams API, it is maybe bug in it? Of it is us, we should maybe just set it automatically?

alesj:
The topic is created by Kafka Streams -- this is the topic that holds the data for the store.
The Streams (topology) configuration is picked up from the Properties instance that's feed to the KafkaStreams constructor.
Apart from bootstrap servers and application id (afaik ...), the rest is optional and has some defaults.
And the default -- from the code I posted in #4694 -- for replication factor is 1.

Maybe we can set this one ourselves?
But then we go down the rabbit hole ... today is replication factor, tomorrow is gonna be property1, then property2, ...

This way -- I'm not saying is the best, but it was easy to impl some quick poc/pr -- we leave the choice and all possible options to the user.

tomncooper:
if you set the kafka streams `replication.factor` to -1 it will match whatever the cluster's default replication factor is. The user should make sure that the default replication factor is greater than the minISR so that would avoid having to set it manually for Kafka Streams.

alesj:
OK, this solves the problem for replication factor ... what about any other property ... 
(I guess that's what we should do in any case -- set the property to -1, if it's not been set yet (once we have a generic property setting solution that is :-)) ... ?)

scholzj:
> OK, this solves the problem for replication factor ... what about any other property ...
(I guess that's what we should do in any case -- set the property to -1, if it's not been set yet (once we have a generic property setting solution that is :-)) ... ?)

The default replication factor can be expected to be in sync with the `min.insync.replicas` ... so at least in the issue linked to this, setting replication factor to `-1` should IMHO be sufficient.

scholzj:
I guess we can close this as obsolete after #4706?

alesj:
Yes

PR #4699, opened 2021-04-04
Provide the parametrized type for future

cutiechi:
Signed-off-by: cutiechi <superchijinpeng@gmail.com>

### Type of change

_Select the type of your PR_

- Enhancement / new feature

### Description

Provide the parametrized type for future

### Checklist

- [ ] Write tests
- [ ] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/Main.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/Main.java
@@ -22,7 +22,6 @@ import io.strimzi.operator.cluster.operator.resource.ResourceOperatorSupplier;
 import io.strimzi.operator.common.PasswordGenerator;
 import io.strimzi.operator.common.Util;
 import io.strimzi.operator.common.operator.resource.ClusterRoleOperator;
-import io.strimzi.operator.common.operator.resource.ReconcileResult;
 import io.vertx.core.CompositeFuture;
 import io.vertx.core.Future;
 import io.vertx.core.Promise;
@@ -98,6 +97,7 @@ public class Main {
         });
     }
 
+    @SuppressWarnings("rawtypes")
     static CompositeFuture run(Vertx vertx, KubernetesClient client, PlatformFeaturesAvailability pfa, ClusterOperatorConfig config) {
         Util.printEnvInfo();
 
@@ -134,7 +134,7 @@ public class Main {
         KafkaRebalanceAssemblyOperator kafkaRebalanceAssemblyOperator =
                 new KafkaRebalanceAssemblyOperator(vertx, pfa, resourceOperatorSupplier, config);
 
-        List<Future<String>> futures = new ArrayList<>(config.getNamespaces().size());
+        List<Future> futures = new ArrayList<>(config.getNamespaces().size());
         for (String namespace : config.getNamespaces()) {
             Promise<String> prom = Promise.promise();
             futures.add(prom.future());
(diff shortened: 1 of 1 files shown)


cluster-operator/src/main/java/io/strimzi/operator/cluster/Main.java
```diff
@@ -159,12 +160,12 @@ static CompositeFuture run(Vertx vertx, KubernetesClient client, PlatformFeature
                     prom.handle(res);
                 });
         }
-        return CompositeFuture.join(futures);
+        return CompositeFuture.join(new ArrayList<>(futures));
```
scholzj, line 163: Why do we need to create a new list here?

cluster-operator/src/main/java/io/strimzi/operator/cluster/Main.java
```diff
@@ -193,7 +194,7 @@ static CompositeFuture run(Vertx vertx, KubernetesClient client, PlatformFeature
             }
 
             Promise<Void> returnPromise = Promise.promise();
-            CompositeFuture.all(futures).onComplete(res -> {
+            CompositeFuture.all(new ArrayList<>(futures)).onComplete(res -> {
```
scholzj, line 197: Same as above ... Why do we need to create a new list here?

cluster-operator/src/main/java/io/strimzi/operator/cluster/Main.java
cutiechi: 截屏2021-04-05 02 47 29

Because `CompositeFuture.join` required type `List<Future>`

cluster-operator/src/main/java/io/strimzi/operator/cluster/Main.java
cutiechi: Because `CompositeFuture.all` required type` List<Future>`

cluster-operator/src/main/java/io/strimzi/operator/cluster/Main.java
scholzj: So why not keep it as `List<Future>` in the first place? It seems like a better solution than this.

cutiechi: Generic types shouldn't be used raw (without type parameters) in variable declarations or return values. Doing so bypasses generic type checking, and defers the catch of unsafe code to runtime.

scholzj:
I added it to the agenda of the Community meeting tomorrow (https://github.com/strimzi/strimzi-kafka-operator#strimzi-community-meetings) to discuss it with other maintainers.

cluster-operator/src/main/java/io/strimzi/operator/cluster/Main.java
tombentley: The signature of `CompositeFuture.join()` using the raw type `Future` is the root cause of the problem, and there isn't much we can do about that, since it's not our code. It forces a raw type to be used _somewhere_. Using `new ArrayList<>(futures)` just lets the compiler infer the type argument to `ArrayList` to be the raw type `Future`, so there's still a raw type, it's just you didn't write it. It comes with the cost of allocating a pointless list, and it this case it doesn't actually make the code any more type safe. (I agree with the general point of avoiding raw types where possible, the problem is that, as here, it's not always possible.)

Perhaps the best thing to do would to be add the following method to `Util`:

```
    @SuppressWarnings({"unchecked", "rawtypes"})
    static <T> CompositeFuture join(List<Future<T>> futures) {
        return CompositeFuture.join((List) futures);
    }
```

and use it everywhere we currently call `CompositeFuture.join`. This isolates the use of raw types to a single method, so avoiding `@SuppressWarnings` at all the call sites.

cutiechi: @tombentley Ok, i see

scholzj: > This isolates the use of raw types to a single method, so avoiding @SuppressWarnings at all the call sites.

There do not seem to be any `@SuppressWarnings` triggered by this. So would it be better to just try to address this at the source? (in Vert.x)

cutiechi: @scholzj See eclipse-vertx/vert.x#3595 why Vert.x doesn't have this.

tombentley:
I think the best solution is to add the helper method to `Util`, as I suggested. Any reason why not to do that?

scholzj:
@cutiechi Any chance you can get back to it and implement it as suggested by @tombentley?

cutiechi:
@scholzj @tombentley  I think even if you do, `Future` will still have a warning when calling this wrapper method

scholzj:
Closing this as discussed on the Strimzi Community Call. We might open a new PR later to move this to Util, but in general there is no major issue with it as it is right now.

Issue #2529, opened 2020-02-10
[Question] spec.replicas in body is required


dorsany:
I'm getting the following error
```
spec.replicas in body is required
```

This is part of my `kafka.yaml` for the default kafka config:
```
...
  kafka:
    config:
      auto.create.topics.enable: "false"
      default.replication.factor: 1
      log.retention.bytes: 2147483648
      log.retention.hours: 48
      log.roll.hours: 24
      message.max.bytes: 2097152
      min.insync.replicas: 1
      num.io.threads: 8
      num.network.threads: 6
      num.partitions: 6
      offsets.topic.replication.factor: 1
      offsets.topic.segment.bytes: 26214400
      transaction.state.log.min.isr: 1
      transaction.state.log.replication.factor: 1
...
```

Here is my `KafkaTopic.yaml` :
```
apiVersion: kafka.strimzi.io/v1beta1
kind: KafkaTopic
metadata:
  name: my.topic
  labels:
    strimzi.io/cluster: kafka-cluster
spec:
  partitions: 18
```
If I will remove the *specs* (in my case spec line and partions line) it will work 
or 
if I will add the *replicas* it will work too

my questions is why I can't configure only the `partitions` and the `KafkaTopic` will use the default value of the replicas (in my case is `default.replication.factor: 1`) from the `kafka.yaml` ?

Version: `0.16.2`

scholzj:
TBH, This looks like a validation issue to me. I think if you don't specify the `spec` section it should error as well. Cc @tombentley @stanlyDoge

dorsany:
Hi @scholzj ,
I understand that it is a validation issue @JsonProperty(required = true) in 
strimzi-kafka-operator/api/src/main/java/io/strimzi/api/kafka/model/KafkaTopicSpec.java

I am asking if can we make both partitions & replicas optional, as the Kafka cluster have these defaults...
example: I want to create a specific topic without specifying the replica count (it depends on the broker replicas) and i want that the topic will be created with the cluster default.

I see that in: strimzi-kafka-operator/topic-operator/src/main/java/io/strimzi/operator/topic/Topic.java there is an option to create without replicas or partitions specifications...

Can we consider it as a feature request?

10x!

ppatierno:
[...]

PR #4711, opened 2021-04-07
Feature/issue 2529 optional topic specs

echernous:
### Type of change

- Enhancement

### Description
https://github.com/strimzi/strimzi-kafka-operator/issues/2529 
Added support for optional replicas and partitions values during new kafkatopic creation, the values will default to broker's default.replication.factor and num.partitions.

### Checklist
- [x] Update CHANGELOG.md


echernous:
Hi @npecka, please could you check this PR? It's for one of your tickets. Thank you!

api/src/test/resources/io/strimzi/api/kafka/model/KafkaTopic-minimal.yaml
```diff
@@ -2,6 +2,3 @@ apiVersion: kafka.strimzi.io/v1beta2
 kind: KafkaTopic
 metadata:
   name: my-topic
-spec:
```
scholzj, line 5: Should the `spec` section be required?

scholzj (commented):
I left one minor comment. I also wonder if we should have some IT test for creating the topic with the defaults (more or less check that it passes and the topic is created)?

api/src/test/resources/io/strimzi/api/kafka/model/KafkaTopic-minimal.yaml
echernous: Hi Jakub, thanks for checking the PR. 

There is the https://github.com/strimzi/strimzi-kafka-operator/blob/main/api/src/test/java/io/strimzi/api/kafka/model/KafkaTopicCrdIT.java#L44 test that verifies resource creation with minimum parameters - now it's without replicas and partitions. I'm checking the existing tests for topic operator, possible we could add a test for topic without replicas and partitions, will come back on this one) 

As there are no other required parameters for the topic to be created we could pass an empty json as it's currently done for the Rebalance test - https://github.com/strimzi/strimzi-kafka-operator/blob/main/api/src/test/resources/io/strimzi/api/kafka/model/KafkaRebalance-minimal.yaml, but it also works without the spec field too. I guess it won't be a big difference to pass empty json or just don't pass the field at all, maybe for visibility to see it exists and can be used.

I noticed there is another issue, if I set default.replication.factor higher than number of brokers - there is no exception to provision a cluster or create a topic with default value, only if I try to use the topic it says INVALID_REPLICATION_FACTOR. Should I create a separate issue for this one or it's expected? Thanks!

scholzj (approved):

ppatierno (approved):

api/src/main/java/io/strimzi/api/kafka/model/KafkaTopicSpec.java
```diff
@@ -50,9 +49,9 @@ public void setTopicName(String topicName) {
             "This cannot be decreased after topic creation. " +
             "It can be increased after topic creation, " +
             "but it is important to understand the consequences that has, " +
-            "especially for topics with semantic partitioning.")
+            "especially for topics with semantic partitioning. " +
+            "When absent this will default to the broker configuration for num.partitions.")
```
tombentley, line 53:
```suggestion
            "When absent this will default to the broker configuration for `num.partitions`.")
```

api/src/main/java/io/strimzi/api/kafka/model/KafkaTopicSpec.java
```diff
@@ -61,10 +60,10 @@ public void setPartitions(Integer partitions) {
         this.partitions = partitions;
     }
 
-    @Description("The number of replicas the topic should have.")
+    @Description("The number of replicas the topic should have. " +
+            "When absent this will default to the broker configuration for default.replication.factor.")
```
tombentley, line 64:
```suggestion
            "When absent this will default to the broker configuration for `default.replication.factor`.")
```

tombentley (approved):

echernous:
Hi all, thanks for checking! I found another issue during testing, still working on it

sknot-rh (approved):

PR #4727, opened 2021-04-09
Add topic name to the topic status

sknot-rh:

### Type of change
- Refactoring

### Description

The K8S KafkaTopic resource does contain `topicName` field. It should not be changed since K8S resource should be bounded to exactly one Kafka Topic. Changing `spec.topicName` brings inconsistency between K8S and Kafka topics.

Putting (and not changing) the `topicName` to the status makes it unambiguous with what `spec.topicName` was topic created.

### Checklist

- [ ] Write tests
- [ ] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/api/src/main/java/io/strimzi/api/kafka/model/status/KafkaTopicStatus.java
+++ b/api/src/main/java/io/strimzi/api/kafka/model/status/KafkaTopicStatus.java
@@ -7,6 +7,7 @@ package io.strimzi.api.kafka.model.status;
 import com.fasterxml.jackson.annotation.JsonInclude;
 import com.fasterxml.jackson.annotation.JsonPropertyOrder;
 import io.strimzi.api.kafka.model.Constants;
+import io.strimzi.crdgenerator.annotations.Description;
 import io.sundr.builder.annotations.Buildable;
 import lombok.EqualsAndHashCode;
 import lombok.ToString;
@@ -19,9 +20,20 @@ import lombok.ToString;
         builderPackage = Constants.FABRIC8_KUBERNETES_API
 )
 @JsonInclude(JsonInclude.Include.NON_NULL)
-@JsonPropertyOrder({ "conditions", "observedGeneration" })
+@JsonPropertyOrder({ "topicName", "conditions", "observedGeneration" })
 @EqualsAndHashCode
 @ToString(callSuper = true)
 public class KafkaTopicStatus extends Status {
     private static final long serialVersionUID = 1L;
+
+    private String topicName;
+
+    @Description("Topic name")
+    public String getTopicName() {
+        return topicName;
+    }
+
+    public void setTopicName(String topicName) {
+        this.topicName = topicName;
+    }
 }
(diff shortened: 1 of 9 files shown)


api/src/main/java/io/strimzi/api/kafka/model/status/KafkaTopicStatus.java
```diff
@@ -19,9 +20,20 @@
         builderPackage = Constants.FABRIC8_KUBERNETES_API
 )
 @JsonInclude(JsonInclude.Include.NON_NULL)
-@JsonPropertyOrder({ "conditions", "observedGeneration" })
+@JsonPropertyOrder({ "topicName", "conditions", "observedGeneration" })
```
scholzj, line 23: I think it should be last after conditions and observed geenration?

scholzj (commented):
Can you please explain a bit more in the PR description what does it mean and why is it important? Is this related to the operator renaming topics on its own? Or what does it relate to?

scholzj:
So what does it do when the `.spec` name is changed? Set some error condition and send an event? Or does it revert the value and updates the CR?

sknot-rh:
It does create an event and sets status similar to this 
```
spec:
  config:
    retention.ms: 7200000
    segment.bytes: 1073741824
  partitions: 1
  replicas: 1
  topicName: my-topic-renamed
status:
  conditions:
  - lastTransitionTime: 2021-04-09T08:53:23.915275Z
    message: Kafka topics cannot be renamed, but KafkaTopic's spec.topicName has changed.
    reason: IllegalArgumentException
    status: "True"
    type: NotReady
  observedGeneration: 2
  topicName: my-topic
```

scholzj (approved):

topic-operator/src/main/java/io/strimzi/operator/topic/TopicOperator.java
```diff
@@ -1010,6 +1010,15 @@ protected void observedTopicFuture(KafkaTopic observedTopic) {
                     KafkaTopicStatus kts = new KafkaTopicStatus();
                     StatusUtils.setStatusConditionAndObservedGeneration(topic, kts, result);
 
+
+
```
tombentley, line 1014: No need for 3 blank lines here

tombentley (commented):
You've only done the check when we see the Kafka topic change in kube. I think we need similar logic in `getKafkaAndReconcile` (which is on the path when reconciliation is timer-triggered). Best to factor out a method.

topic-operator/src/main/java/io/strimzi/operator/topic/TopicDiff.java
```diff
@@ -225,7 +225,7 @@ public static TopicDiff diff(Topic source, Topic target) {
         Objects.requireNonNull(source.getTopicName());
         Objects.requireNonNull(target.getTopicName());
         if (!source.getTopicName().equals(target.getTopicName())) {
-            throw new IllegalArgumentException();
+            throw new IllegalArgumentException("Kafka topics cannot be renamed, but KafkaTopic's spec.topicName has changed.");
```
tombentley, line 228: I wonder if we really still need this. I would have thought that we were not detecting name changes before we get as far as doing the diff.

tombentley (approved):

topic-operator/src/main/java/io/strimzi/operator/topic/TopicDiff.java
sknot-rh: IIRC this path is active in periodical reconciliation.

PR #4732, opened 2021-04-09
ConnectorMockTest flaky sometimes

rareddy:

### Type of change

- Bugfix for Testcase

### Description

This test fixes part of the code that fails due to timing. The code is written with expetation that the reconciler will be run twice before the kafka connect status with ready condition is made, however with mockkube ready status is set on the first reconciler run, then based on timing the test fails due to that, the assertion is changed for that to make sure test passes in all scenarios


rareddy:
@tombentley fyi

tombentley (approved):
Thanks!

scholzj (approved):

PR #4740, opened 2021-04-13
Do certificate reconciliation at a single point

tombentley:
PR opened for CI purposes. This can currently be ignored.


scholzj:
Just FYI: IIRC the main reason why we didn't had it in one place was because we had only weak mechanism to see if the certificate updated or not. This is currently not addressed here. Just saying ... sorry if stating the obvious ;-).

PR #4717, opened 2021-04-08
Remove support for legacy logging configuration from the CO code

ShubhamRwt:
### Type of change

- Refactoring

### Description

Removed the support for legacy logging configuration from the CO code

### Checklist

_Please go through this checklist and make sure all applicable tasks have been done_

- [ ] Write tests
- [ ] Make sure all tests pass
- [ ] Update documentation
- [ ] Check RBAC rights for Kubernetes / OpenShift roles
- [ ] Try your changes from Pod inside your Kubernetes and OpenShift cluster, not just locally
- [ ] Reference relevant issue(s) and close them after merging
- [ ] Update CHANGELOG.md
- [ ] Supply screenshots for visual changes, such as Grafana dashboards

PR #4761, opened 2021-04-15
[systemtest] Fixups for LoggingChangesST tests and fix of exception message inside the `AbstractModel.java`

im-konge:

### Type of change

- Test fixes

### Description

This PR fixes two tests from `LoggingChangeST`, which starts to fail after #4717.
Also I added fix for exception message in `AbstractModel.java`. This exception appears when we set CM (to external logging) which doesn't exist.

Before:
```
    - lastTransitionTime: 2021-04-15T21:22:41.625Z
      message: ConfigMap non-existing-cm-namewith external logging configuration does
        not exist or doesn't contain the configuration under the {} keylog4j.properties.
      reason: InvalidResourceException
      status: "True"
      type: NotReady
```
After:
```
- lastTransitionTime: 2021-04-15T21:33:31.134Z
      message: ConfigMap non-existing-cm-name with external logging configuration
        does not exist or doesn't contain the configuration under the log4j.properties
        key.
      reason: InvalidResourceException
      status: "True"
      type: NotReady
```

### Checklist

### DIFF
--- a/.azure/regression-pipeline.yaml
+++ b/.azure/regression-pipeline.yaml
@@ -67,8 +67,7 @@ jobs:
     parameters:
       name: 'regression_all_remaining'
       display_name: 'regression-bundle VII. - remaining system tests'
-      # !LoggingChangeST is skipped because it can be flaky on Azure
-      test_case: '!kafka/**/*ST,!mirrormaker/**/*ST,!connect/**/*ST,!security/**/*ST,!LoggingChangeST,!operators/**/*ST,!rollingupdate/**/*ST,!watcher/**/*ST,!tracing/**/*ST'
+      test_case: '!kafka/**/*ST,!mirrormaker/**/*ST,!connect/**/*ST,!security/**/*ST,!operators/**/*ST,!rollingupdate/**/*ST,!watcher/**/*ST,!tracing/**/*ST'
       groups: 'regression'
       cluster_operator_install_type: 'bundle'
       timeout: 360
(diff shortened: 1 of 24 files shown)


im-konge:
@strimzi-ci run tests testcase=LoggingChangeST

scholzj (approved):

scholzj (commented):
The code looks ok. But ...
* #4717 passed all regression tests without any issues. So are these not run in regression? Or why did it not failed there?
* After #4717 we throw an exception when the logging is badly configured. But this particular logging configuration did IMHO not worked even before it. The issue was just silently ignored. So I wonder if the logic of the test is correct or not if it before didn't recognized the issue.

im-konge:
@scholzj I'm looking into the Azure regression pipeline, where is this comment
` # !LoggingChangeST is skipped because it can be flaky on Azure`, so this is the reason why it didn't failed in the #4717

scholzj:
@im-konge So is it tracked anywhere? Is anyone working on it?

scholzj:
It is disabled for 9 months now ... so maybe someone should at least try out if it works now?

im-konge:
And about the second point -> in the past, the test was designed to check if the default CM is set when user set CM to some non-existing one. IIRC there was discussion about this, if it's correct or if there should be exception. Now the test checks that the previous CM is kept, the status of Kafka is `NotReady` and message appears. Or should I somehow change the test to check more things?

im-konge:
@scholzj yeah sure, I can try it in this PR

Frawless (approved):

im-konge:
@scholzj @Frawless tests are passing, so I think this is ready for merge?

PR #3827, opened 2020-10-15
Fix leaking HttpClient connections

tombentley:
Signed-off-by: Tom Bentley <tbentley@redhat.com>

### Type of change

- Bugfix

### Description

Fix leaking HttpClient connections.

PR #4762, opened 2021-04-15
Fix leaking HttpClient connections in Cruise Control client

tomncooper:
### Type of change

Bugfix

### Description

This PR fixes a potential issue with unclosed connections in the cruise control client api. The fix applies the same approach as #3827 so a common `HttpClientUtils` class has been created that is used by both the CC and Kafka Connect client implementations.

### Checklist

### DIFF
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaConnectApiImpl.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaConnectApiImpl.java
@@ -8,6 +8,7 @@ package io.strimzi.operator.cluster.operator.assembly;
 import com.fasterxml.jackson.core.type.TypeReference;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import io.strimzi.api.kafka.model.connect.ConnectorPlugin;
+import io.strimzi.operator.cluster.operator.resource.HttpClientUtils;
 import io.strimzi.operator.common.BackOff;
 import io.strimzi.operator.common.Util;
 import io.strimzi.operator.common.model.OrderedProperties;
@@ -16,7 +17,6 @@ import io.vertx.core.Handler;
 import io.vertx.core.Promise;
 import io.vertx.core.Vertx;
 import io.vertx.core.buffer.Buffer;
-import io.vertx.core.http.HttpClient;
 import io.vertx.core.http.HttpClientOptions;
 import io.vertx.core.json.JsonArray;
 import io.vertx.core.json.JsonObject;
@@ -32,7 +32,6 @@ import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
-import java.util.function.BiConsumer;
 import java.util.function.Supplier;
 
 import static java.util.Arrays.asList;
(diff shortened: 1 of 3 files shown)


cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/resource/cruisecontrol/CruiseControlApiImpl.java
```diff
@@ -189,126 +181,141 @@ public CruiseControlApiImpl(Vertx vertx, int idleTimeout) {
...
+        return promise.future().compose(
+            result -> {
+                httpClient.close();
+                return Future.succeededFuture(result);
+            },
+            error -> {
+                httpClient.close();
+                return Future.failedFuture(error);
+            });
+    }
```
scholzj, line 320: This looks identical to the one Tom has in `KafkaConnectApiImpl`. Can we move it to some `Util` or `HttpUtil` class and share it? Or does it actually use something from the class and cannot be static?

scholzj (commented):
One nit. LGTM otherwise.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/resource/cruisecontrol/CruiseControlApiImpl.java
tomncooper: That is a good idea. I have moved the method to an external util class. Should I edit the Connect client to use it in this same PR or open another after this is merged?

scholzj: I think you can do it in this PR.

tomncooper: Done

scholzj (approved):

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaConnectApiImpl.java
```diff
@@ -63,7 +62,7 @@ public KafkaConnectApiImpl(Vertx vertx) {
         Buffer data = configJson.toBuffer();
         String path = "/connectors/" + connectorName + "/config";
         log.debug("Making PUT request to {} with body {}", path, configJson);
-        return withHttpClient((httpClient, result) ->
+        return HttpClientUtils.withHttpClient(vertx, new HttpClientOptions().setLogActivity(true), (httpClient, result) ->
```
tombentley, line 65: Just checking that this wasn't left over from debugging?

tombentley (approved):

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaConnectApiImpl.java
tomncooper: AFAIK that matches what the old KafkaConnect API was setting for all HttpClientOptions? So I left it the same.

ppatierno (approved):

Issue #3324, opened 2020-07-15
Make sure kafka pods are ready after the dynamic configuration


sknot-rh:
More info: https://github.com/strimzi/strimzi-kafka-operator/pull/2389#discussion_r454997887

PR #2389, opened 2020-01-13
Dynamic configuration of Kafka brokers

stanlyDoge:
### Type of change
- Enhancement / new feature

### Description
The algorithm detects whether Kafka pods can be reconfigured without rolling update or not. If yes, the configuration is updated dynamically.

### Checklist

- [ ] Update/write design documentation in `./design`
- [x] Write tests
- [x] Make sure all tests pass
- [ ] Update documentation
- [ ] Check RBAC rights for Kubernetes / OpenShift roles
- [x] Try your changes from Pod inside your Kubernetes and OpenShift cluster, not just locally
- [ ] Reference relevant issue(s) and close them after merging
- [ ] Update CHANGELOG.md

PR #4763, opened 2021-04-15
Dynamic Config TODO should be closed out

npecka:
### Type of change

- Refactoring

### Description

After investigation I do not feel this TODO is needed anymore and can be closed out. As we were investigating, we found that the only things that could be Dynamically Configured were the loggers. After testing the functionality with potential configurations aimed to make the loggers unhealthy to trigger a bad broker config, we found the broker object not reporting that the logger was misconfigured but the Cluster Operator would report back with this. Even though it was broken, the Kafka Broker still reported back in a healthy state and seemed to function fine without the logger. So no matter the change that can be applied Dynamically, it all seems to be related to the logger in this context which doesn't seem to be an essential component to the Broker to function and report back healthy. 

This is in response too --> #2389 and #3324

For future it might be a good idea to incorporate some kind of check to the CO for bad logger configuration and stop a dynamic update rather than continue to apply a potential "bad" config.

### Checklist

- [x] Write tests
- [ ] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/resource/KafkaRoller.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/resource/KafkaRoller.java
@@ -360,7 +360,6 @@ public class KafkaRoller {
                             log.debug("{}: Pod {} can be rolled now", reconciliation, podId);
                             restartAndAwaitReadiness(pod, operationTimeoutMs, TimeUnit.MILLISECONDS);
                         } else {
-                            // TODO do we need some check here that the broker is still OK?
                             awaitReadiness(pod, operationTimeoutMs, TimeUnit.MILLISECONDS);
                         }
                     } else {


kyguy (approved):

sknot-rh:
Hi. Thanks for the investigation. However, the Kafka broker config (not just the loggers) is dynamically changeable. Check the usages of `io.strimzi.operator.common.Util#getBrokersConfig(java.lang.String)` to see where it is performed.
IIRC this TODO is about checking that pod is ok after the dynamic configuration. Some cases like validation are caught by Kafka broker itself. We need to think about cases where the config is successfully changed but breaks the broker. 
Please check https://kafka.apache.org/documentation/#brokerconfigs The options with `Update Mode` with value `per-broker` or `cluster-wide` are dynamically changeable. We need to pick some from this set, hypothetically `num.replica.fetchers`, set it some value which passes Kafka brokers validation, but breaks its state (it is just example, @tombentley can help you to pick up the option that can break the broker).

tombentley:
@sknot-rh's comment is spot on. Setting `num.io.threads` to a very high value is a great example. I assume the broker would go and try to create that many threads, hit some JVM or OS limit and possibly not function correctly, even though the config update returned OK. I was concerned that a readiness check alone wouldn't detect that (hence the TODO). The consequence of having an inadequate check are that the KafkaRoller moves on to apply the change to the next broker, taking it down too. And so on. 

Reviewing the broker configs, in addition to the things like `num.io.threads` which we might be able to detect at the time we're diing the rolling update, there are some configs which could result in failure much later:

* `num.recovery.threads.per.data.dir`: it would probably lurk in a bad state until the pod was restarted (possibly crashing the broker during shutdown), then prevent the broker starting up.
* `num.replica.alter.log.dirs.threads` similarly might get set OK and only become a problem at an arbitrary future point (reassignment)

So I'm now inclined to think that this is, in general, orthogonal to rolling reconfiguration, and not something that the operator can hope to protect against. So I'm +1 on removing the comment for that reason.

tombentley (approved):

sknot-rh (approved):

PR #4765, opened 2021-04-16
Fix double dots and asciidoc formatting in tables

scholzj:
### Type of change

- Documentation

### Description

This PR fixes some issues in the APi reference documentation:

* It fixes the way we check for `.` at the end of the description to work fine with multiline descriptions. In thepast, we had two places where this generated `..` at the end of the description. This should be now fixed by using the `DOTALL` flag in the Java pattern matcher.
* It enabled asciidoc formatting in the schema tables. That was disabled in the past and as result, it was not formatting for example lists properly. See the example of this before this change:
    Screenshot 2021-04-16 at 13 52 07
  and after this change:
    Screenshot 2021-04-16 at 13 51 33

These are not major issues. But it makes the docs little bit better.

### Checklist

- [x] Update documentation

### DIFF
--- a/documentation/modules/appendix_crds.adoc
+++ b/documentation/modules/appendix_crds.adoc
@@ -8,9 +8,9 @@
 [options="header"]
 |====
 |Property       |Description
-|spec    1.2+<.<|The specification of the Kafka and ZooKeeper clusters, and Topic Operator.
+|spec    1.2+<.<a|The specification of the Kafka and ZooKeeper clusters, and Topic Operator.
 |xref:type-KafkaSpec-{context}[`KafkaSpec`]
-|status  1.2+<.<|The status of the Kafka and ZooKeeper clusters, and Topic Operator.
+|status  1.2+<.<a|The status of the Kafka and ZooKeeper clusters, and Topic Operator.
 |xref:type-KafkaStatus-{context}[`KafkaStatus`]
 |====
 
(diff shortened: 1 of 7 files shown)


tombentley (approved):

documentation/modules/appendix_crds.adoc
```diff
@@ -279,34 +279,34 @@ include::../api/io.strimzi.api.kafka.model.listener.arraylistener.GenericKafkaLi
...
+|preferredNodePortAddressType  1.2+<.<a|Defines which address type should be used as the node address. Available types are: `ExternalDNS`, `ExternalIP`, `InternalDNS`, `InternalIP` and `Hostname`. By default, the addresses will be used in the following order (the first one found will be used):
 * `ExternalDNS`
 * `ExternalIP`
 * `InternalDNS`
 * `InternalIP`
 * `Hostname`
 
-This field can be used to select the address type which will be used as the preferred type and checked first. In case no address will be found for this address type, the other types will be used in the default order.This field can be used only with `nodeport` type listener..
+This field can be used to select the address type which will be used as the preferred type and checked first. In case no address will be found for this address type, the other types will be used in the default order.This field can be used only with `nodeport` type listener.
```
PaulRMellor, line 307:
```suggestion
This field is used to select the preferred address type, which is checked first. If no address is found for this address type, the other types are checked in the default order. This field can only be used with `nodeport` type listener.
```

PaulRMellor (approved):

laidan6000 (approved):

PR #4770, opened 2021-04-16
[MO] - [checkstyle] -> able to run it via intelij checkstyle plugin

see-quick:

### Type of change

- Enhancement / new feature
- Refactoring

### Description

This PR removing `java.header` file and improving regex for the header. Moreover, everyone is able to set up checkstyle plugin and see problems in a real-time when editing code. 

For instance:
image
image

The problem was that with in that `java.header` file was an issue with parsing regex inside that file. I tried to escape the `\{` and `\}` but it does not work. So I added regex inside value attribute and it works.  

image.


scholzj (approved):
If it works ... why not.

Frawless (approved):

tombentley (approved):

ppatierno (approved):

PR #4773, opened 2021-04-18
Add support for Kafka 2.8.0

scholzj:
### Type of change

- Enhancement / new feature

### Description

This PR adds support for Apache Kafka 2.8.0. Debezium - used as the test cluster in some integration tests - is currently not ready for Kafka 2.8.0. So this PR does not move the operators to use 2.8.0 clients - just adds support for running 2.8.0 as operand. Since we currently do not use any 2.8.0 feature in the operator code, this should not be a blocker and can be fixed later when needed / when Debezium supports 2.8.

_Note: The failing regression tests are related to Helm Charts. That seems to be caused by a bug in the tests. I raised #4778 for this._

### Checklist

- [x] Update documentation
- [x] Update CHANGELOG.md


tombentley (approved):
I've only looked at the config model part, but it looks ok to me.

sknot-rh (approved):

PR #4775, opened 2021-04-19
Reduce the number of Kubernetes API calls

### LINKED ISSUE #4669: Minimize empty patch API calls

Rancher management server appears to listen on all Service patch requests, and could have increased overhead applied to it with a large number of Kafka installations in the same cluster. Some patches to Services appear empty and cause Rancher to churn on its own reconciliation as a result.

Perform diffing or caching when reconciling resources around the Kafka installation to minimize API calls made to the k8s server.

Rancher is investigating if any alternatives exist to listening for all API calls surrounding Services to power its functionality around these objects.

**Additional context**

The combination of these two issues (Strimzi patching Services every reconciliation loop + Rancher reconciling its data on every Service patch) is the only apparent cause of degraded behavior. Although either fix will provide some relief, the ideal is to have both fixes implemented to guarantee minimal overhead in both contexts.

scholzj:
### Type of change

- Enhancement / new feature

### Description

Right, in every reconciliation we patch every resource we manage although in many cases there are no real changes. That sometimes causes problems since different resources handle these events differently. For example for Pod Disruption Budgets or Persistent Volume Claims, even through we do not see any change it triggers the resource updates. Also - while I'm not aware of any specific performance issues related to this - it will also cost some performance both in the Strimzi operator and in the Kube API server. And in some specific cases such as the one described in #4669 it can cause issues to the Kubernetes cluster.

This PR adds a new diffing mechanism before we actually patch the resources. It compares the desired and current resource and does the patch only in case we actually find some difference. By default, it ignores just some basic fields (`.status` or `.metadata.managedFields`), but different resources can customize what should or should not be ignored. In some cases, the patching is just caused by default values set by Kubernetes. So this PR also modifies some of our code to set the defaults explicitly and avoid patching in these cases. 

The general approach I chose is to:
* Ignore differences in the fields which we do not configure anywhere
* Set the defaults explicitly for the fields which can be optionally configured

There are many resource kinds we manage. This PR does not address all of them. In particular StatefulSets and Deployments have a lot of different fields with different defaults. It focuses instead on the following resources where the patching seems to cause issues or recreate the resources (or where the fix was simple and obvious):
* Network policies
* Secrets
* Services (this is the issue in #4669)
* Pod Disruption Budgets (they already had a custom fix for this, now they rely on the generic implementation => so it was just redesigned)
* Persistent Volume Claims

Further PRs might look into more resources.

This should close #4669 

### Checklist

- [x] Write tests


tombentley:
I've not looked at tha PR to great detail yet, but have some initial questions. This is an interesting PR because it starts doing some similar logic to fabrics8's own `patch()` implementation. Fabric8's `patch()` gets the resource, diffs that resource with the desired resource and sends the patch. Here we end up doing a similar thing except we're adding more resource-specific logic to decide when the patch would be a no-op and in those cases not actually making a `PATCH` request. 

I've thought several times in the past about whether fabric8's patch method is always the right tool for what we want. Even before this PR we had several places where we computed a diff ourselves and decided what to do. And that adds inefficiency because fabirc will have to get the resource again, and figure out the diff again.

So my hypothetical question is: If fabric8 had a variant of `patch()` which took an actual `JsonDiff` argument would that help here, and elsewhere in Strimzi, do you think? If would allow us to get the resource, diff it, decide whether we really wanted update the resource based on the diff and optionally patch it.

A related question is: How much of this PR is really papering over inadequacies of fabric's logic for deciding whether a patch is _effectively_ empty? Because that's what all the default stuff here is about, isn't it? Wouldn't it be better for that to be in fabric8?

scholzj:
@tombentley TBH, I'm not completely sure how much of this is caused by some (possibly unique) way how we use the APIs, how much of this is caused by Fabric8 client and how much of this is just bad design in the Kubernetes API. One of the problems is that there are many conditional defaults which are hard to handle both and that is what is causing the issues.

I suspect lot of these would be solved if we didn't generate new resource every-time but queried the the existing resource and edited it as that would better deal with the defaults and you would not need so much logic to decide if it is patch or NOOP. But I'm not 100% sure there are no hidden issues with doing it that way. This PR is obviously much easier fix than that. It also really focuses just on few resources and not on all of them.

Having some way to plug-in our own diffing into the Fabric8 patch method would probably save some cycles over the current solution. But I think it is still rather a hack than a real solution.

Frawless (approved):

PR #4779, opened 2021-04-20
[systemtest] HelmChartST - change image registry and repo for all images

### LINKED ISSUE #4778: [QE] Regression STs are not testing the right images in HelmChartST
The `HelmChartST` is running as part of the regression test suite. But it does not seem to modify the image names. That means that:

* When new version of Kafka is added, the tests always fail because the images are not in Quay.io
* When running tests for a PR not adding new version of Kafka it is using the images from Quay.io which are not those from the PR.

This should be fixed.

im-konge:

### Type of change

- Bugfix

### Description

When we were deploying CO by Helm in `HelmChartST`, we didn't change image repository and registry for Kafka images, which were pushed to local registry -> so on Azure regression this ST fails when new Kafka version is added and not yet present in `quay.io` repo.

This PR adds  `imageRegistryOverride` and `imageRepositoryOverride` to override registry and repository in all images.  

Closes #4778 

### Checklist


scholzj (approved):

Frawless (approved):

Frawless:
LGTM, I assume that everything was replaced properly and it's working in CI as it should

im-konge:
@Frawless yes, I checked it in CI and also tested it locally.

PR #4785, opened 2021-04-21
ST: Downgrade Strimzi without upgraded Kafka

Frawless:

### Type of change

- Bugfix

### Description

We support downgrade only when `inter.broker.protocol.version` and `log.message.format.version` wasn't upgraded to the new Kafka version. We should reflect this in tests.

https://strimzi.io/docs/operators/latest/full/deploying.html#con-target-downgrade-version-str

### Checklist

### DIFF
--- a/systemtest/src/main/java/io/strimzi/systemtest/utils/TestKafkaVersion.java
+++ b/systemtest/src/main/java/io/strimzi/systemtest/utils/TestKafkaVersion.java
@@ -30,7 +30,6 @@ public class TestKafkaVersion implements Comparable<TestKafkaVersion> {
     private static List<TestKafkaVersion> kafkaVersions;
     private static List<TestKafkaVersion> supportedKafkaVersions;
 
-
     static {
         try {
             kafkaVersions = parseKafkaVersions(TestUtils.USER_PATH + "/../kafka-versions.yaml");
@@ -42,7 +41,7 @@ public class TestKafkaVersion implements Comparable<TestKafkaVersion> {
                 throw new Exception("There is no one Kafka version supported inside " + TestUtils.USER_PATH + "/../kafka-versions.yaml file");
             }
 
-            LOGGER.debug("These are following supported Kafka versions:\n{}", kafkaVersions.toString());
+            LOGGER.debug("These are following Kafka versions:\n{}", kafkaVersions.toString());
             LOGGER.debug("These are following supported Kafka versions:\n{}", supportedKafkaVersions.toString());
         } catch (Exception e) {
             e.printStackTrace();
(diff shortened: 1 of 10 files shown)


systemtest/src/test/java/io/strimzi/systemtest/upgrade/AbstractUpgradeST.java
```diff
@@ -412,7 +412,15 @@ protected void setupEnvAndUpgradeClusterOperator(ExtensionContext extensionConte
         if (!cmdKubeClient().getResources(getResourceApiVersion(Kafka.RESOURCE_PLURAL, operatorVersion)).contains(clusterName)) {
             // Deploy a Kafka cluster
             if ("HEAD".equals(testParameters.getString("fromVersion"))) {
-                resourceManager.createResource(extensionContext, KafkaTemplates.kafkaPersistent(clusterName, 3, 3).build());
+                resourceManager.createResource(extensionContext, KafkaTemplates.kafkaPersistent(clusterName, 3, 3)
+                    .editSpec()
+                        .editKafka()
+                            .withVersion(kafkaVersion)
+                            .addToConfig("log.message.format.version", kafkaVersion.substring(0, 3))
+                            .addToConfig("inter.broker.protocol.version", kafkaVersion.substring(0, 3))
```
scholzj, line 420: I wonder if you should use some smarter way than `substring(0, 3)` to extract the format. While it is a long time before we have a minor version with more digits, one day it might happen.

Frawless: Maybe get it from `kafka-versions.yaml` is better

Frawless:
I also need to double-check `KafkaUpgradeDowngradeST` for the same changes.

Frawless:
I went through `KafkaUpgradeDowngradeST` and it doesn't need any changes from my POV.

systemtest/src/main/java/io/strimzi/systemtest/utils/TestKafkaVersion.java
```diff
@@ -28,20 +28,22 @@
 
     private static final Logger LOGGER = LogManager.getLogger(TestKafkaVersion.class);
     private static List<TestKafkaVersion> kafkaVersions;
+    private static List<TestKafkaVersion> supportedKafkaVersions;
+
```
scholzj, line 32: Is the seocnd empty line needed?

systemtest/src/main/java/io/strimzi/systemtest/utils/TestKafkaVersion.java
```diff
@@ -28,20 +28,22 @@
...
-            kafkaVersions = supportedKafkaVersions;
-
-            if (kafkaVersions == null || kafkaVersions.size() == 0) {
+            if (supportedKafkaVersions == null || supportedKafkaVersions.size() == 0) {
                 throw new Exception("There is no one Kafka version supported inside " + TestUtils.USER_PATH + "/../kafka-versions.yaml file");
             }
 
-            LOGGER.info("These are following supported Kafka versions:\n{}", kafkaVersions.toString());
+            LOGGER.debug("These are following supported Kafka versions:\n{}", kafkaVersions.toString());
+            LOGGER.debug("These are following supported Kafka versions:\n{}", supportedKafkaVersions.toString());
```
scholzj, line 46: Is this supposed to be twice here? If yes, I guess the test should be different?

scholzj (approved):

PR #4793, opened 2021-04-22
Use default values in CRD generation

LiamClarkeNZ:
### Type of change

- Enhancement

### Description

Use defaults from @DefaultValue annotations on model objects to populate the `default` field of CRDs generated for API version v1.

### Checklist

- [X] Write tests
- [ ] Update documentation
- [ ] Update CHANGELOG.md


crd-generator/src/main/java/io/strimzi/crdgenerator/DefaultValueSelector.java
```diff
@@ -0,0 +1,12 @@
+/*
+ * Copyright Strimzi authors.
+ * License: Apache License 2.0 (see the file LICENSE or http://apache.org/licenses/LICENSE-2.0.html).
+ */
+package io.strimzi.crdgenerator;
+
+public class DefaultValueSelector {
```
LiamClarkeNZ, line 7: Err, this shouldn't have been committed. Will push a change to PR to remove it tomorrow.

scholzj (commented):
I think you need to run `make crd_install` to get the build passing. I also wonder what value does this add. Where do the defaults in the CRDs show up? How do they affect the API?

crd-generator/src/main/java/io/strimzi/crdgenerator/CrdGenerator.java
```diff
@@ -305,9 +305,16 @@ public CrdGenerator(VersionRange<KubeVersion> targetKubeVersions, ApiVersion crd
         this.storageVersion = storageVersion;
         this.servedVersion = servedVersions;
         this.conversionStrategy = conversionStrategy;
+
+        //noinspection ReplaceNullCheck
```
tombentley, line 309: What's this?

crd-generator/src/main/java/io/strimzi/crdgenerator/DefaultValueHandler.java
```diff
@@ -0,0 +1,96 @@
...
+
+import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.Method;
+import java.util.Arrays;
+import java.util.Optional;
+import java.util.function.Consumer;
+
+import static io.strimzi.api.annotations.ApiVersion.V1;
+
+class DefaultValueHandler {
```
tombentley, line 20: Can we add a sentence of Javadoc saying what this class is for?

crd-generator/src/main/java/io/strimzi/crdgenerator/DefaultValueSelector.java
```diff
@@ -0,0 +1,12 @@
...
+
+    public <T> T doFoo() {
```
tombentley, line 9: Are, you're a fellow `foo` user :smile:

crd-generator/src/test/java/io/strimzi/crdgenerator/DefaultValueHandlerTest.java
```diff
@@ -0,0 +1,222 @@
...
+    }
+
+    @Test
+    void testCustomisedEnumDefaults() throws NoSuchMethodException {
+        Property property = new Property(DefaultsGalore.class.getDeclaredMethod("getCustomisedEnum"));
+        ObjectNode result = v1DefaultHandler.addDefaultValueIfApplicable(blankNode, property);
+        assertEquals(CustomisedEnum.TWO.toValue(), result.get("default").asText());
+    }
+
+
```
tombentley, line 119: Can we remove this excessive whitespace?

tombentley (commented):
What benefit (apart from being more self documenting) does putting the defaults in the CRD schema bring? Does it get used by kube itself, or tools which interact with the API server? I guess `kubectl explain` would show it, but are you aware of anything else?

It would be good to support the annotation in the `DocGenerator` too. And we should add to the Javadoc of `@DefaultValue` exactly what the semantics are (CRD schema and/or docs). 

And probably review all the api classes and figure out where we can use the annotation and remove it from the description. I wonder if there's a way to prevent us from slipping back to our old sloppy ways and forgetting about this entirely. Maybe a `@NoDefaultValue` annotation, and have the tools force that one or the other is present on all scalar-types fields?

LiamClarkeNZ:
> I think you need to run `make crd_install` to get the build passing. 

Cheers :) 


> I also wonder what value does this add. 

Haha, I made these changes based on a comment of yours from 2019 that we should use them for v1 CRDs. ;) 

> Where do the defaults in the CRDs show up? How do they affect the API?

They show up in the field schema, [for example](https://github.com/strimzi/strimzi-kafka-operator/blob/c474426d0acfba8b86f0c8254cdee72892fac61d/packaging/install/cluster-operator/041-Crd-kafkaconnect.yaml#L263). 

According to the docs, defaults will be applied if explicit values aren't specified:
https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#defaulting

LiamClarkeNZ:
> What benefit (apart from being more self documenting) does putting the defaults in the CRD schema bring? Does it get used by kube itself, or tools which interact with the API server? I guess `kubectl explain` would show it, but are you aware of anything else?

The defaults are used by kubectl etc. https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#defaulting

> It would be good to support the annotation in the `DocGenerator` too. 

I was intending to do that with a subsequent branch, but happy to do it in this one if you'd prefer both in at the same time.

> And we should add to the Javadoc of `@DefaultValue` exactly what the semantics are (CRD schema and/or docs).

`@DefaultValue` is a Vertx annotation, so I don't control it, however, can easily replace it with one of our own. 


> And probably review all the api classes and figure out where we can use the annotation and remove it from the description.

Will do.  

> I wonder if there's a way to prevent us from slipping back to our old sloppy ways and forgetting about this entirely. Maybe a `@NoDefaultValue` annotation, and have the tools force that one or the other is present on all scalar-types fields?

I lack a bit of context here, what issues happened in the past?

crd-generator/src/main/java/io/strimzi/crdgenerator/CrdGenerator.java
LiamClarkeNZ: A hint to Intellij to stop suggesting a particular refactoring. It keeps suggesting replacing this:

```
        if (defaultValueHandler == null) {
            this.defaultValueHandler = new DefaultValueHandler(crdApiVersion, this::err);
        } else {
            this.defaultValueHandler = defaultValueHandler;
        }
```

With 

```
        this.defaultValueHandler = Objects.requireNonNullElseGet(defaultValueHandler, 
                () -> new DefaultValueHandler(crdApiVersion, this::err));
                
```

And highlights it for IDEA users as a warning, accordingly, so that comment turns it off for that statement.

scholzj:
> Haha, I made these changes based on a comment of yours from 2019 that we should use them for v1 CRDs. ;)

Did I? Interesting. Looks like until 2021 I forgot that the schema even has some `default` field. I think the main value might be docs.

> The defaults are used by kubectl etc. https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#defaulting

Hmm, is that really what we want? It will make the CRs have a lot more fields if these values are automatically added into the actual resource. So we would need to carefully decide what is important enough to add and what is not (like it might make sense for replicas which are required ... but probably not for the five different probe timeouts). But if we cannot use the default value for all fields with default, it will quickly loose any value it might have for the documentation :-/

LiamClarkeNZ:
> > Haha, I made these changes based on a comment of yours from 2019 that we should use them for v1 CRDs. ;)
> 
> Did I? Interesting. Looks like until 2021 I forgot that the schema even has some `default` field. I think the main value might be docs.
> 
> > The defaults are used by kubectl etc. https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#defaulting
> 
> Hmm, is that really what we want? It will make the CRs have a lot more fields if these values are automatically added into the actual resource. So we would need to carefully decide what is important enough to add and what is not (like it might make sense for replicas which are required ... but probably not for the five different probe timeouts). But if we cannot use the default value for all fields with default, it will quickly loose any value it might have for the documentation :-/

From my understanding of the docs, default valued fields are pruned, but applied by the API Server or etcd.

scholzj:
So, if this is not supported on older Kube versions which we still support, should we close this?

LiamClarkeNZ:
> So, if this is not supported on older Kube versions which we still support, should we close this?

Yep, will do so now.

LiamClarkeNZ:
Because of the K8s bug described here: https://github.com/kubernetes/kubernetes/issues/92849

Using defaults in CRDs breaks in K8s versions we support, so this PR should be closed.

Issue #4766, opened 2021-04-16
Standalone Topic Operator can't connect via TLS without authentication


jozenstar:
_Please use this to only for bug reports. For questions or when you need help, you can use the [GitHub Discussions](https://github.com/strimzi/strimzi-kafka-operator/discussions), our [#strimzi Slack channel](https://slack.cncf.io/) or out [user mailing list](https://lists.cncf.io/g/cncf-strimzi-users/topics)._

**Describe the bug**
I'm using Strimzi to operate on AWS MSK cluster which requires TLS connection but doesn't have any authentication enabled at the moment. I've set the KafkaConnect successfully with this codeblock
```
tls:
    trustedCertificates: []

```
However, I can't set the Standalone Topic Operator to behave the same. If you set `STRIMZI_TLS_ENABLED` to `true` then the topic operator requires trust- and keystore locations and passwords to be set.  If you set `STRIMZI_TLS_ENABLED` to `false` then you get the Java OOM (because of nonTLS connection). 
I've tried to set `STRIMZI_TLS_ENABLED` to `false`  and set the `security.protocol` to `SSL` via `STRIMZI_JAVA_SYSTEM_PROPERTIES`, however this value is being overwritten by AdminConfig.
 

**To Reproduce**
1. Create an MSK cluster with strict TLS connection but without authentication
2. Create a TopicOperator to connect to it
3. Check logs 

**Expected behavior**
I'd like the standalone topic operator to be able to communicate via TLS without an authentication

**Environment (please complete the following information):**
 - Strimzi version: 0.22.1
 - Installation method: YAML files
 - Kubernetes cluster: Kubernetes v1.18.9-eks-d1db3c
 - Infrastructure: Amazon EKS

**YAML files and logs**
I've tried different configurations. I believe that this one should work, but it doesn't.
```
apiVersion: apps/v1
kind: Deployment
metadata:
  name: strimzi-topic-operator
  labels:
    app: strimzi
  namespace: strimzi
spec:
  replicas: 1
  selector:
    matchLabels:
      name: strimzi-topic-operator
  template:
    metadata:
      labels:
        name: strimzi-topic-operator
    spec:
      serviceAccountName: strimzi-topic-operator
      containers:
        - name: strimzi-topic-operator
          image: quay.io/strimzi/operator:0.22.1
          args:
          - /opt/strimzi/bin/topic_operator_run.sh
          env:
            - name: STRIMZI_RESOURCE_LABELS
[...]

Issue #2761, opened 2020-03-29
[Question] Standalone topic operator with AWS MSK (TLS error)


elisiano:
I stood up a standalone topic operator following the docs (latest 0.17.0) and I'm trying to have it communicate with an MSK cluster.

The MSK cluster is configured to allow only TLS connections to the brokers.

If I just set `STRIMZI_TLS_ENABLED` to true, the startup scripts fail because it expects the certificates in the tls-sidecar path, but those directories do not exist (there is no side car running, nor I could find instructions on how to do so, but I assume that the sidecar gets deployed when using the topic operator within the cluster operator).

Looking at the startup scripts it seems that the logic to create a keystore is only executed if the truststore and keystore variables are not used.

The upstream AWS MSK [docs](https://docs.aws.amazon.com/msk/latest/developerguide/produce-consume.html) suggest to use the default truststore since the certificates used in the MSK clusters have public certs.

So I tried setting `STRIMZI_TRUSTSTORE_LOCATION` to `/usr/lib/jvm/jre/lib/security/cacerts` which, after a series of symlink, resolve to the correct file (I tried listing the content with keytool with an empty password and the list of CAs is pretty big).

this is the startup logs (and failure):
```
➭ kubectl logs strimzi-topic-operator-msk-0-575b9d9df7-jbpkh
+ shift
+ . /opt/strimzi/bin/dynamic_resources.sh
++ get_heap_size
+++ cat /sys/fs/cgroup/memory/memory.limit_in_bytes
++ CONTAINER_MEMORY_IN_BYTES=524288000
++ DEFAULT_MEMORY_CEILING=32505856
++ '[' 524288000 -lt 32505856 ']'
+ MAX_HEAP=
+ '[' -n '' ']'
+ export MALLOC_ARENA_MAX=2
+ MALLOC_ARENA_MAX=2
+ JAVA_OPTS=' -Dvertx.cacheDirBase=/tmp -Djava.security.egd=file:/dev/./urandom'
++ get_gc_opts
++ '[' '' == true ']'
++ echo ''
+ JAVA_OPTS=' -Dvertx.cacheDirBase=/tmp -Djava.security.egd=file:/dev/./urandom '
[...]

PR #4796, opened 2021-04-22
Topic operator tls auth switch

jozenstar:
### Type of change

- Enhancement / new feature

### Description

Previously standalone topic-operator didn't support the TLS encrypted connection without authentication.
This fix allows you to establish an encrypted TLS connection without authentication to your kafka cluster.
This fix won't affect the current approach until you set the new variable `STRIMZI_TLS_AUTH_ENABLED` to `false` in the `05-Deployment-strimzi-topic-operator.yaml`

### Checklist

- [x] Update documentation

### DIFF
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/model/EntityTopicOperator.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/model/EntityTopicOperator.java
@@ -276,6 +276,7 @@ public class EntityTopicOperator extends AbstractModel {
         varList.add(buildEnvVar(ENV_VAR_ZOOKEEPER_SESSION_TIMEOUT_MS, Integer.toString(zookeeperSessionTimeoutMs)));
         varList.add(buildEnvVar(ENV_VAR_TOPIC_METADATA_MAX_ATTEMPTS, String.valueOf(topicMetadataMaxAttempts)));
         varList.add(buildEnvVar(ENV_VAR_TLS_ENABLED, Boolean.toString(true)));
+        varList.add(buildEnvVar(ENV_VAR_TLS_AUTH_ENABLED, Boolean.toString(true)));
         varList.add(buildEnvVar(ENV_VAR_STRIMZI_GC_LOG_ENABLED, String.valueOf(gcLoggingEnabled)));
         EntityOperator.javaOptions(varList, getJvmOptions(), javaSystemProperties);
 
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/model/EntityTopicOperator.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/model/EntityTopicOperator.java
@@ -55,7 +55,7 @@ public class EntityTopicOperator extends AbstractModel {
     public static final String ENV_VAR_ZOOKEEPER_SESSION_TIMEOUT_MS = "STRIMZI_ZOOKEEPER_SESSION_TIMEOUT_MS";
     public static final String ENV_VAR_TOPIC_METADATA_MAX_ATTEMPTS = "STRIMZI_TOPIC_METADATA_MAX_ATTEMPTS";
     public static final String ENV_VAR_TLS_ENABLED = "STRIMZI_TLS_ENABLED";
-    public static final String ENV_VAR_TLS_ENABLED = "STRIMZI_TLS_AUTH_ENABLED";
+    public static final String ENV_VAR_TLS_AUTH_ENABLED = "STRIMZI_TLS_AUTH_ENABLED";
     public static final Probe DEFAULT_HEALTHCHECK_OPTIONS = new ProbeBuilder()
             .withInitialDelaySeconds(EntityTopicOperatorSpec.DEFAULT_HEALTHCHECK_DELAY)
             .withTimeoutSeconds(EntityTopicOperatorSpec.DEFAULT_HEALTHCHECK_TIMEOUT).build();
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/model/EntityTopicOperator.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/model/EntityTopicOperator.java
@@ -55,6 +55,7 @@ public class EntityTopicOperator extends AbstractModel {
     public static final String ENV_VAR_ZOOKEEPER_SESSION_TIMEOUT_MS = "STRIMZI_ZOOKEEPER_SESSION_TIMEOUT_MS";
     public static final String ENV_VAR_TOPIC_METADATA_MAX_ATTEMPTS = "STRIMZI_TOPIC_METADATA_MAX_ATTEMPTS";
     public static final String ENV_VAR_TLS_ENABLED = "STRIMZI_TLS_ENABLED";
+    public static final String ENV_VAR_TLS_ENABLED = "STRIMZI_TLS_AUTH_ENABLED";
     public static final Probe DEFAULT_HEALTHCHECK_OPTIONS = new ProbeBuilder()
             .withInitialDelaySeconds(EntityTopicOperatorSpec.DEFAULT_HEALTHCHECK_DELAY)
             .withTimeoutSeconds(EntityTopicOperatorSpec.DEFAULT_HEALTHCHECK_TIMEOUT).build();
--- a/documentation/modules/deploying/proc-deploy-topic-operator-standalone.adoc
+++ b/documentation/modules/deploying/proc-deploy-topic-operator-standalone.adoc
@@ -34,14 +34,16 @@ Default `20000` (20 seconds).
 Default `/strimzi/topics`.
 .. `STRIMZI_TLS_ENABLED` to enable TLS support for encrypting the communication with Kafka brokers.
 Default `true`.
+.. `STRIMZI_TLS_AUTH_ENABLED` to enable authenticated communication with Kafka brokers.
+Default `true`.
 .. `STRIMZI_TRUSTSTORE_LOCATION` to the path to the truststore containing certificates for enabling TLS based communication.
-Mandatory only if TLS is enabled through `STRIMZI_TLS_ENABLED`.
+Mandatory only if TLS and authentication are enabled through `STRIMZI_TLS_ENABLED` and `STRIMZI_TLS_AUTH_ENABLED`.
 .. `STRIMZI_TRUSTSTORE_PASSWORD` to the password for accessing the truststore defined by `STRIMZI_TRUSTSTORE_LOCATION`.
-Mandatory only if TLS is enabled through `STRIMZI_TLS_ENABLED`.
+Mandatory only if TLS and authentication are enabled through `STRIMZI_TLS_ENABLED` and `STRIMZI_TLS_AUTH_ENABLED`.
 .. `STRIMZI_KEYSTORE_LOCATION` to the path to the keystore containing private keys for enabling TLS based communication.
-Mandatory only if TLS is enabled through `STRIMZI_TLS_ENABLED`.
+Mandatory only if TLS and authentication are enabled through `STRIMZI_TLS_ENABLED` and `STRIMZI_TLS_AUTH_ENABLED`.
 .. `STRIMZI_KEYSTORE_PASSWORD` to the password for accessing the keystore defined by `STRIMZI_KEYSTORE_LOCATION`.
-Mandatory only if TLS is enabled through `STRIMZI_TLS_ENABLED`.
+Mandatory only if TLS and authentication are enabled through `STRIMZI_TLS_ENABLED` and `STRIMZI_TLS_AUTH_ENABLED`.
 .. `STRIMZI_LOG_LEVEL` to the level for printing logging messages.
 The value can be set to: `ERROR`, `WARNING`, `INFO`, `DEBUG`, and `TRACE`.
 Default `INFO`.
--- a/topic-operator/scripts/topic_operator_run.sh
+++ b/topic-operator/scripts/topic_operator_run.sh
@@ -14,7 +14,7 @@ if [ -n "$STRIMZI_JAVA_OPTS" ]; then
     export JAVA_OPTS="${JAVA_OPTS} ${STRIMZI_JAVA_OPTS}"
 fi
 
-if [ "$STRIMZI_TLS_ENABLED" = "true" ]; then
+if [ "$STRIMZI_TLS_ENABLED" = "true" ] && [ "$STRIMZI_TLS_AUTH_ENABLED" = "true" ]; then
     if [ -z "$STRIMZI_TRUSTSTORE_LOCATION" ] && [ -z "$STRIMZI_KEYSTORE_LOCATION" ]; then
         # Generate temporary keystore password
         CERTS_STORE_PASSWORD=$(< /dev/urandom tr -dc _A-Z-a-z-0-9 | head -c32)
--- a/topic-operator/src/main/java/io/strimzi/operator/topic/Session.java
+++ b/topic-operator/src/main/java/io/strimzi/operator/topic/Session.java
@@ -156,11 +156,15 @@ public class Session extends AbstractVerticle {
 
         if (Boolean.parseBoolean(config.get(Config.TLS_ENABLED))) {
             kafkaClientProps.setProperty(AdminClientConfig.SECURITY_PROTOCOL_CONFIG, "SSL");
-            kafkaClientProps.setProperty(SslConfigs.SSL_TRUSTSTORE_LOCATION_CONFIG, config.get(Config.TLS_TRUSTSTORE_LOCATION));
-            kafkaClientProps.setProperty(SslConfigs.SSL_TRUSTSTORE_PASSWORD_CONFIG, config.get(Config.TLS_TRUSTSTORE_PASSWORD));
-            kafkaClientProps.setProperty(SslConfigs.SSL_KEYSTORE_LOCATION_CONFIG, config.get(Config.TLS_KEYSTORE_LOCATION));
-            kafkaClientProps.setProperty(SslConfigs.SSL_KEYSTORE_PASSWORD_CONFIG, config.get(Config.TLS_KEYSTORE_PASSWORD));
-            kafkaClientProps.setProperty(SslConfigs.SSL_ENDPOINT_IDENTIFICATION_ALGORITHM_CONFIG, config.get(Config.TLS_SSL_ENDPOINT_IDENTIFICATION_ALGORITHM));
+            if (Boolean.parseBoolean(config.get(Config.TLS_AUTH_ENABLED))) {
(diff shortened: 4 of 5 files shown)


jozenstar:
Relevant issues and discussions
https://github.com/strimzi/strimzi-kafka-operator/discussions/4787
https://github.com/strimzi/strimzi-kafka-operator/issues/4766
https://github.com/strimzi/strimzi-kafka-operator/issues/2761

jozenstar:
@scholzj Hi,
I've made a fix for https://github.com/strimzi/strimzi-kafka-operator/discussions/4787
Can you review it, please?

Thanks

scholzj (commented):
Thanks for the PR. I think you will need to integrate this also into the Cluster Operator otherwise this will break it. More or less add the new variable in here like it is done for the `STRIMZI_TLS_ENABLED`: https://github.com/strimzi/strimzi-kafka-operator/blob/f136bbc77913f1d6ef837057008a766f67e67329/cluster-operator/src/main/java/io/strimzi/operator/cluster/model/EntityTopicOperator.java#L268

scholzj:
@jozenstar Do you think you will be bale to get back to the comment above?

jozenstar:
@scholzj Hi,
Sorry for the late update.
I've applied the required fix.
Please see last commit.

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/EntityTopicOperator.java
```diff
@@ -55,6 +55,7 @@
     public static final String ENV_VAR_ZOOKEEPER_SESSION_TIMEOUT_MS = "STRIMZI_ZOOKEEPER_SESSION_TIMEOUT_MS";
     public static final String ENV_VAR_TOPIC_METADATA_MAX_ATTEMPTS = "STRIMZI_TOPIC_METADATA_MAX_ATTEMPTS";
     public static final String ENV_VAR_TLS_ENABLED = "STRIMZI_TLS_ENABLED";
+    public static final String ENV_VAR_TLS_AUTH_ENABLED = "STRIMZI_TLS_AUTH_ENABLED";
```
scholzj, line 58: I guess you will need to also use the new variable here? https://github.com/strimzi/strimzi-kafka-operator/blob/f2ce3ea17c28846ad61640df55f46231a4cea17f/cluster-operator/src/main/java/io/strimzi/operator/cluster/model/EntityTopicOperator.java#L268

topic-operator/scripts/topic_operator_run.sh
```diff
@@ -14,7 +14,7 @@ if [ -n "$STRIMZI_JAVA_OPTS" ]; then
     export JAVA_OPTS="${JAVA_OPTS} ${STRIMZI_JAVA_OPTS}"
 fi
 
-if [ "$STRIMZI_TLS_ENABLED" = "true" ]; then
+if [ "$STRIMZI_TLS_ENABLED" = "true" ] && [ "$STRIMZI_TLS_AUTH_ENABLED" = "true" ]; then
```
scholzj, line 17: I think the TRUSTSTORE is needed even when TLS authentication is not enabled. So it needs to be set even if `STRIMZI_TLS_ENABLED` is set to `true` but `STRIMZI_TLS_AUTH_ENABLED` is `false`.

topic-operator/src/main/java/io/strimzi/operator/topic/Session.java
```diff
@@ -156,11 +156,13 @@ public void start(Promise<Void> start) {
...
         if (Boolean.parseBoolean(config.get(Config.TLS_ENABLED))) {
             kafkaClientProps.setProperty(AdminClientConfig.SECURITY_PROTOCOL_CONFIG, "SSL");
-            kafkaClientProps.setProperty(SslConfigs.SSL_TRUSTSTORE_LOCATION_CONFIG, config.get(Config.TLS_TRUSTSTORE_LOCATION));
-            kafkaClientProps.setProperty(SslConfigs.SSL_TRUSTSTORE_PASSWORD_CONFIG, config.get(Config.TLS_TRUSTSTORE_PASSWORD));
-            kafkaClientProps.setProperty(SslConfigs.SSL_KEYSTORE_LOCATION_CONFIG, config.get(Config.TLS_KEYSTORE_LOCATION));
-            kafkaClientProps.setProperty(SslConfigs.SSL_KEYSTORE_PASSWORD_CONFIG, config.get(Config.TLS_KEYSTORE_PASSWORD));
             kafkaClientProps.setProperty(SslConfigs.SSL_ENDPOINT_IDENTIFICATION_ALGORITHM_CONFIG, config.get(Config.TLS_SSL_ENDPOINT_IDENTIFICATION_ALGORITHM));
+            if (Boolean.parseBoolean(config.get(Config.TLS_AUTH_ENABLED))) {
+                kafkaClientProps.setProperty(SslConfigs.SSL_TRUSTSTORE_LOCATION_CONFIG, config.get(Config.TLS_TRUSTSTORE_LOCATION));
+                kafkaClientProps.setProperty(SslConfigs.SSL_TRUSTSTORE_PASSWORD_CONFIG, config.get(Config.TLS_TRUSTSTORE_PASSWORD));
```
scholzj, line 162: Same as above => the truststore options need to be gated by `TLS_ENABLED`. Truststore is needed even without authentication.

documentation/modules/deploying/proc-deploy-topic-operator-standalone.adoc
```diff
@@ -34,14 +34,16 @@ Default `20000` (20 seconds).
...
 .. `STRIMZI_TLS_ENABLED` to enable TLS support for encrypting the communication with Kafka brokers.
 Default `true`.
+.. `STRIMZI_TLS_AUTH_ENABLED` to enable authenticated communication with Kafka brokers.
+Default `true`.
 .. `STRIMZI_TRUSTSTORE_LOCATION` to the path to the truststore containing certificates for enabling TLS based communication.
-Mandatory only if TLS is enabled through `STRIMZI_TLS_ENABLED`.
+Mandatory only if TLS and authentication are enabled through `STRIMZI_TLS_ENABLED` and `STRIMZI_TLS_AUTH_ENABLED`.
 .. `STRIMZI_TRUSTSTORE_PASSWORD` to the password for accessing the truststore defined by `STRIMZI_TRUSTSTORE_LOCATION`.
-Mandatory only if TLS is enabled through `STRIMZI_TLS_ENABLED`.
+Mandatory only if TLS and authentication are enabled through `STRIMZI_TLS_ENABLED` and `STRIMZI_TLS_AUTH_ENABLED`.
```
scholzj, line 42: Same as below - `STRIMZI_TRUSTSTORE_LOCATION` and `STRIMZI_TRUSTSTORE_PASSWORD` should require only `STRIMZI_TLS_ENABLED`.

topic-operator/scripts/topic_operator_run.sh
jozenstar: Sorry, I'm not a Java developer, so I can not be right here, but the current setup works perfectly for me.
Why would we need the TRUSTSTORE to be set when the authentication is disabled?
I suppose in that case java uses its default truststore (cacerts file).

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/EntityTopicOperator.java
jozenstar: Thanks for noticing!
Fixed.

topic-operator/scripts/topic_operator_run.sh
scholzj: Truststore is the same as for example `caFile` in other languages. It is a set of public keys used to verify the identity of the server it is connecting to. You indeed do not need to configure it if you are fine with using the default certificates bundled with JVM (bascially if your server uses certificate signed by some trusted CA). But that doesn't make it part of TLS client authentication. It could be that this works for you but it is not the right solution and it will not work for others.

If your problem is that you do not want to set the truststore, maybe you can deal with it differently:
* Scrap the whole `TLS_AUTH_ENABLED` thing
* And change the code to something like this (not tested, just sample code):
  ```java
        if (Boolean.parseBoolean(config.get(Config.TLS_ENABLED))) {
            kafkaClientProps.setProperty(AdminClientConfig.SECURITY_PROTOCOL_CONFIG, "SSL");
            
            if (!config.get(Config.TLS_TRUSTSTORE_LOCATION).isEmpty() && !config.get(Config.TLS_TRUSTSTORE_PASSWORD).isEmpty()) {
                kafkaClientProps.setProperty(SslConfigs.SSL_TRUSTSTORE_LOCATION_CONFIG, config.get(Config.TLS_TRUSTSTORE_LOCATION));
                kafkaClientProps.setProperty(SslConfigs.SSL_TRUSTSTORE_PASSWORD_CONFIG, config.get(Config.TLS_TRUSTSTORE_PASSWORD));
            }

            if (!config.get(Config.TLS_KEYSTORE_LOCATION).isEmpty() && !config.get(Config.TLS_KEYSTORE_PASSWORD).isEmpty()) {
                kafkaClientProps.setProperty(SslConfigs.SSL_KEYSTORE_LOCATION_CONFIG, config.get(Config.TLS_KEYSTORE_LOCATION));
                kafkaClientProps.setProperty(SslConfigs.SSL_KEYSTORE_PASSWORD_CONFIG, config.get(Config.TLS_KEYSTORE_PASSWORD));
            }
            
            kafkaClientProps.setProperty(SslConfigs.SSL_ENDPOINT_IDENTIFICATION_ALGORITHM_CONFIG, config.get(Config.TLS_SSL_ENDPOINT_IDENTIFICATION_ALGORITHM));
        }
  ```

This way, you do not need to change the other operator and simply if the truststore environment variables are set, the will be used. If the keystore variables are used, they will be used. (at the end, Kafka clients do not have any switch for TLS client authentication, they just do what the broker says IIRC).

jozenstar: Well, I can do it, but then should we disable TRUSTSTORE and KEYSTORE generation in the topic_operator_run.sh?
I'm talking about this code block
```
if [ "$STRIMZI_TLS_ENABLED" = "true" ]; then
    if [ -z "$STRIMZI_TRUSTSTORE_LOCATION" ] && [ -z "$STRIMZI_KEYSTORE_LOCATION" ]; then
        # Generate temporary keystore password
        CERTS_STORE_PASSWORD=$(< /dev/urandom tr -dc _A-Z-a-z-0-9 | head -c32)
        export CERTS_STORE_PASSWORD

        mkdir -p /tmp/topic-operator

        # Import certificates into keystore and truststore
        "${STRIMZI_HOME}/bin/tls_prepare_certificates.sh"

        export STRIMZI_TRUSTSTORE_LOCATION=/tmp/topic-operator/replication.truststore.p12
        export STRIMZI_TRUSTSTORE_PASSWORD="$CERTS_STORE_PASSWORD"

        export STRIMZI_KEYSTORE_LOCATION=/tmp/topic-operator/replication.keystore.p12
        export STRIMZI_KEYSTORE_PASSWORD="$CERTS_STORE_PASSWORD"
    fi
fi
```

scholzj: Hm, right. You cannot disable it because it is needed. You would need to make it conditional somehow as well I guess.

jozenstar: I can keep `STRIMZI_TLS_AUTH_ENABLED` variable, but change the Session.java so the `SSL_TRUSTSTORE_LOCATION_CONFIG` and `SSL_TRUSTSTORE_PASSWORD_CONFIG` will be set if it's set as `STRIMZI_TRUSTSTORE_LOCATION` and `STRIMZI_TRUSTSTORE_PASSWORD` env vars.

Like this
```
if (Boolean.parseBoolean(config.get(Config.TLS_ENABLED))) {
            kafkaClientProps.setProperty(AdminClientConfig.SECURITY_PROTOCOL_CONFIG, "SSL");
            kafkaClientProps.setProperty(SslConfigs.SSL_ENDPOINT_IDENTIFICATION_ALGORITHM_CONFIG, config.get(Config.TLS_SSL_ENDPOINT_IDENTIFICATION_ALGORITHM));

            if (!config.get(Config.TLS_TRUSTSTORE_LOCATION).isEmpty() && !config.get(Config.TLS_TRUSTSTORE_PASSWORD).isEmpty()) {
                kafkaClientProps.setProperty(SslConfigs.SSL_TRUSTSTORE_LOCATION_CONFIG, config.get(Config.TLS_TRUSTSTORE_LOCATION));
                kafkaClientProps.setProperty(SslConfigs.SSL_TRUSTSTORE_PASSWORD_CONFIG, config.get(Config.TLS_TRUSTSTORE_PASSWORD));
            }
            if (Boolean.parseBoolean(config.get(Config.TLS_AUTH_ENABLED))) {
                kafkaClientProps.setProperty(SslConfigs.SSL_KEYSTORE_LOCATION_CONFIG, config.get(Config.TLS_KEYSTORE_LOCATION));
                kafkaClientProps.setProperty(SslConfigs.SSL_KEYSTORE_PASSWORD_CONFIG, config.get(Config.TLS_KEYSTORE_PASSWORD));
            }
        }
```


And the `topic-operator-run.sh` script will remain the same

```
if [ "$STRIMZI_TLS_ENABLED" = "true" ] && [ "$STRIMZI_TLS_AUTH_ENABLED" = "true" ]; then
    if [ -z "$STRIMZI_TRUSTSTORE_LOCATION" ] && [ -z "$STRIMZI_KEYSTORE_LOCATION" ]; then
        # Generate temporary keystore password
        CERTS_STORE_PASSWORD=$(< /dev/urandom tr -dc _A-Z-a-z-0-9 | head -c32)
        export CERTS_STORE_PASSWORD

        mkdir -p /tmp/topic-operator

        # Import certificates into keystore and truststore
        "${STRIMZI_HOME}/bin/tls_prepare_certificates.sh"

        export STRIMZI_TRUSTSTORE_LOCATION=/tmp/topic-operator/replication.truststore.p12
        export STRIMZI_TRUSTSTORE_PASSWORD="$CERTS_STORE_PASSWORD"

        export STRIMZI_KEYSTORE_LOCATION=/tmp/topic-operator/replication.keystore.p12
        export STRIMZI_KEYSTORE_PASSWORD="$CERTS_STORE_PASSWORD"
    fi
fi
```

This way both full strimzi and standalone deployments will work, right?

scholzj: I think this possibly breaks it for other users who want to have just the truststore configured.

jozenstar: Why? If they have specified the `STRIMZI_TRUSTSTORE_LOCATION` and `STRIMZI_TRUSTSTORE_LOCATION` - then it'll be set as a property according to my latest example of Session.java:
```
if (Boolean.parseBoolean(config.get(Config.TLS_ENABLED))) {
            kafkaClientProps.setProperty(AdminClientConfig.SECURITY_PROTOCOL_CONFIG, "SSL");
            kafkaClientProps.setProperty(SslConfigs.SSL_ENDPOINT_IDENTIFICATION_ALGORITHM_CONFIG, config.get(Config.TLS_SSL_ENDPOINT_IDENTIFICATION_ALGORITHM));

            if (!config.get(Config.TLS_TRUSTSTORE_LOCATION).isEmpty() && !config.get(Config.TLS_TRUSTSTORE_PASSWORD).isEmpty()) {
                kafkaClientProps.setProperty(SslConfigs.SSL_TRUSTSTORE_LOCATION_CONFIG, config.get(Config.TLS_TRUSTSTORE_LOCATION));
                kafkaClientProps.setProperty(SslConfigs.SSL_TRUSTSTORE_PASSWORD_CONFIG, config.get(Config.TLS_TRUSTSTORE_PASSWORD));
            }
            if (Boolean.parseBoolean(config.get(Config.TLS_AUTH_ENABLED))) {
                kafkaClientProps.setProperty(SslConfigs.SSL_KEYSTORE_LOCATION_CONFIG, config.get(Config.TLS_KEYSTORE_LOCATION));
                kafkaClientProps.setProperty(SslConfigs.SSL_KEYSTORE_PASSWORD_CONFIG, config.get(Config.TLS_KEYSTORE_PASSWORD));
            }
        }
```

jozenstar: We can even specify a conditional set of keystore properties even if `TLS_AUTH_ENABLED` is set to `true` (it's set to `true` by default)

scholzj: I'm sorry, but the truststore is simply not part of TLS authentication and should not be gated by it. I appreciate that you are trying to contribute, but it needs to be done right. There might be users using only TLS without authentication but who still need to provide the truststore. So it needs to be possible without enabling the TLS client authentication.

jozenstar: This is actually what I'm talking about.
Regarding my latest example they could specify `TRUSTORE` location and password even if the `TLS_AUTH_ENABLED` is set to `false`.

Maybe I got you wrong, so could you please clarify if this is the case.

P.S. Thanks for your active participation <3

scholzj: The last Java code looks fine (although I guess we do not need `TLS_AUTH_ENABLED` really here since we can check the keystore the same way as truststore?). But the Bash code has it still gated both through `TLS_AUTH_ENABLED`

jozenstar: Well, as far as I see, the bash code generates certificates if `STRIMZI_TRUSTSTORE_LOCATION` and `STRIMZI_KEYSTORE_LOCATION` are empty. This certificate generation only works with the full strimzi deployment. It doesn't work for standalone topic operator deployment.
The idea is:
If you use the full strimzi deployment (with kafka, zookeper etc. managed by strimzi in k8s) and `TLS_ENABLED` set to `true` then you'll always have both TLS encryption and authentication enabled. And the bash script `if` condition will pass. They couldn't disable authentication (why would they want to do it?)
If you use the standalone topic operator deployment - then the bash conditions will never pass because you have either `TLS_AUTH_ENABLED` set to `true` with `TRUSTSTORE` and `KEYSTORE` specified or `TLS_AUTH_ENABLED` set to `false` with `TRUSTSTORE` either specified or not.

Do you want me to also implement TLS_AUTH switch for the full strimzi deployment?

scholzj: No, you can have standalone users who:
* Want to use TLS encryption
* Want to use it with self signed certificate (i.e. they need the truststore to be generated)
* But do not want to use TLS authentication => so `TLS_AUTH_ENABLED=false` and no keystore

scholzj:
@jozenstar Any chance you can get back to this?

jozenstar:
> @jozenstar Any chance you can get back to this?

@scholzj At the moment I'm working on some high priority tasks, so I'll come back to this when I'll finish

scholzj:
@jozenstar Ok, great. Thanks.

scholzj:
There is a similar PR #5201 which seems to do the same in a way but seems to be moving forward. I think we can maybe close this one?

jozenstar:
@scholzj Sure, if it does the same

PR #4807, opened 2021-04-23
Mark CRD versions as deprecated

scholzj:
### Type of change

- Enhancement / new feature

### Description

We should have used this earlier. But I did not know CRDs have such option. This PR adds support for the `deprecated` field in the CRD version specification. It is used to:
* Mark all `KafkaConnectS2I` versions as deprecated together with a custom message pointing to the migration doc:
    ```
    Warning: KafkaConnectS2I resource is deprecated. Use the KafkaConnect resource instead. See https://strimzi.io/docs/operators/latest/using.html#proc-migrating-kafka-connect-s2i-str for instructions how to migrate.
    kafkaconnects2i.kafka.strimzi.io/my-connect-s2i created
    ```
* Mark all v1beta1 and v1alpha1 versions in the other resources as deprecated. In the new CRDs, the older verions are actually used only for `KafkaUser` and `KafkaTopics` resources, but in case we need to generate some v1beta1 CRDs as an emergency it might be handy:
    ```
    Warning: kafka.strimzi.io/v1alpha1 KafkaTopic is deprecated; use kafka.strimzi.io/v1beta2 KafkaTopic
    kafkatopic.kafka.strimzi.io/kafka-test-apps created
    Warning: kafka.strimzi.io/v1alpha1 KafkaUser is deprecated; use kafka.strimzi.io/v1beta2 KafkaUser
    kafkauser.kafka.strimzi.io/my-user created
    ```
* I noticed we did not had the `v1beta2` versions set as stored versions (although they ar eused as stored version in the generated v1 CRDs). So I updated this as part of this PR as well.

### Checklist

### DIFF
--- a/api/src/main/java/io/strimzi/api/kafka/model/Kafka.java
+++ b/api/src/main/java/io/strimzi/api/kafka/model/Kafka.java
@@ -43,9 +43,9 @@ import static java.util.Collections.unmodifiableList;
                 group = Kafka.RESOURCE_GROUP,
                 scope = Kafka.SCOPE,
                 versions = {
-                        @Crd.Spec.Version(name = Kafka.V1BETA2, served = true, storage = false),
-                        @Crd.Spec.Version(name = Kafka.V1BETA1, served = true, storage = true),
-                        @Crd.Spec.Version(name = Kafka.V1ALPHA1, served = true, storage = false)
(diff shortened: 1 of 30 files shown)


scholzj:
Stupid ... this is 1.19+ only feature. Kube 1.16 rejects the CRDs with these values 🤦

PR #4809, opened 2021-04-24
Mark v1beta2 version as stored version

scholzj:
### Type of change

- Task

### Description

Originally done as part of #4807 which was closed because it is not backwards compatible with Kube 1.18 and lower. This PR updates the stored version in our CRDs to `v1beta2`. This was already used in the `apiextensions/v1` CRDs anyway, so this affects only the old `apiextensions/v1beta1` used for testing.


scholzj:
@$#& ... this of course breaks the tests for the upgrade of the stored version 🤦

PR #4812, opened 2021-04-24
[MO] - [Azure] -> MM pipeline paralelism disabled

see-quick:

### Type of change

- Bugfix

### Description

Same problem as with https://github.com/strimzi/strimzi-kafka-operator/pull/4806 PR. Mirror maker regression pipeline sometimes end up because of the OOM problem. 

### Checklist

### DIFF
--- a/.azure/regression-pipeline.yaml
+++ b/.azure/regression-pipeline.yaml
@@ -62,6 +62,7 @@ jobs:
       test_case: 'mirrormaker/**/*ST,kafka/dynamicconfiguration/**/*ST'
       groups: 'regression'
       cluster_operator_install_type: 'bundle'
+      run_parallel: false
       timeout: 360
 
   - template: 'templates/system_test_general.yaml'


scholzj (approved):

Frawless (approved):

.azure/regression-pipeline.yaml
```diff
@@ -62,6 +62,7 @@ jobs:
       test_case: 'mirrormaker/**/*ST,kafka/dynamicconfiguration/**/*ST'
       groups: 'regression'
       cluster_operator_install_type: 'bundle'
+      run_parallel: false
```
Frawless, line 65: do we really need to set this? We use also `parallel` so maybe it makes sense to set it to `1` ?

see-quick: Yes, we do. Based on my observation I saw some edge cases when you allow paralelism `(junit.jupiter.execution.parallel.enabled=true)` and let `( junit.jupiter.execution.parallel.config.fixed.parallelism=1)` can run 2 tests at once. It is maybe bug inside `junit` but practically speaking it can be resolved but setting this flag to `false`. Moreover it ensures that only `main` thread will executed tests instead of `fork-join` pool thread selected by JVM.

PR #4828, opened 2021-04-27
[systemtest] Fixes for failing tests

im-konge:

### Type of change

- Test fixes

### Description

`MetricsST` - problem with extra created topic, which affected another test
`MirrorMakerST` - we didn't wait for the clients pod, also there was some random issue with producer wasn't able to connect to Kafka bootstrap, even if everything was correctly set and Kafka pods up and running
`CustomResourceStatusST` - connector was overriden by another test
`ColdBackupScriptST` - on `minikube` we wasn't able to pull the `centos:7` image from docker hub - because of reaching maximum pull limit every time - this PR changes it to pull it from quay

### Checklist


Frawless (approved):

tools/cold-backup/templates/patch.json
```diff
@@ -3,7 +3,7 @@
         "containers": [
             {
                 "name": "dummy",
-                "image": "centos:7",
+                "image": "quay.io/centos/centos:8",
```
scholzj, line 6: You should not change this. Last time I checked, the Quay repositories were not up to date.

im-konge: So, what image will we add here? Because everything from docker hub will not be pulled. @Frawless suggested Strimzi image. Or we can change it via `sed` command directly in ST

scholzj: Well, the base image also pulls from DockerHub. So if you cannot pull from there you should not even get to this test.

scholzj: After / if https://github.com/strimzi/proposals/pull/31 is approved, we can change it to UBI8 as well.

im-konge: Okay, so I'll revert it and we'll keep it fail until/if the proposal will be merged. Thanks @scholzj

scholzj (approved):

PR #4847, opened 2021-04-28
[MO] - [Strimzi Kafka container] -> unique naming of advertised liste…

see-quick:

### Type of change

- Bugfix
- Enhancement / new feature

### Description

This PR fixing problem for the Strimzi Kafka container. The problem is that when you have more than 1 network we are configuring advertised listeners which has the same naming and that ends up in the following problem:

```
[2021-04-28 10:10:34,553] ERROR Exiting Kafka due to fatal exception (kafka.Kafka$)
java.lang.IllegalArgumentException: requirement failed: Each listener must have a different name, listeners: PLAINTEXT://localhost:49185,BROKER://172.19.0.2:9093,BROKER://172.17.0.5:9093
```

This change should eliminate this problem.

### Checklist

### DIFF
--- a/test-container/src/main/java/io/strimzi/StrimziKafkaContainer.java
+++ b/test-container/src/main/java/io/strimzi/StrimziKafkaContainer.java
@@ -97,19 +97,40 @@ public class StrimziKafkaContainer extends GenericContainer<StrimziKafkaContaine
 
         Collection<ContainerNetwork> cns = containerInfo.getNetworkSettings().getNetworks().values();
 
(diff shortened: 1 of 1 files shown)


scholzj (approved):
So, I can see how this fixes the problem you described. But I'm a bit confused how does the user find the right listener to use.

Frawless (approved):

Frawless:
@scholzj  shouldn't users use `PLAINTEXT` one in that case?

scholzj:
What are the other listeners for then?

PR #4858, opened 2021-04-29
[MO] - [unit tests] -> model package (class + method parallelism)

see-quick:

### Type of change

- Enhancement / new feature

### Description

This PR adding support for all tests inside `cluster-operator/io/strimzi/operator/cluster/model` package which can be executed by method and class-wide parallelism.

### Checklist

### DIFF
--- a/cluster-operator/src/test/java/io/strimzi/operator/cluster/model/AbstractConfigurationTest.java
+++ b/cluster-operator/src/test/java/io/strimzi/operator/cluster/model/AbstractConfigurationTest.java
@@ -10,8 +10,9 @@ import java.util.Map;
 
 import io.strimzi.operator.common.InvalidConfigParameterException;
 import io.strimzi.operator.common.model.OrderedProperties;
+import io.strimzi.test.annotations.ParallelSuite;
+import io.strimzi.test.annotations.ParallelTest;
 import io.vertx.core.json.JsonObject;
-import org.junit.jupiter.api.Test;
 
 import static io.strimzi.test.TestUtils.LINE_SEPARATOR;
 import static java.util.Arrays.asList;
@@ -20,6 +21,7 @@ import static org.hamcrest.CoreMatchers.nullValue;
 import static org.hamcrest.MatcherAssert.assertThat;
 import static org.junit.jupiter.api.Assertions.fail;
 
+@ParallelSuite
 public class AbstractConfigurationTest {
 
     static OrderedProperties createProperties(String... keyValues) {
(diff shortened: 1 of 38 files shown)


scholzj (approved):
I guess if it works then it is good.

see-quick:
Yes, this package was simple I would say....but in the `operator` package it will be challenge for me :D

tombentley (commented):
I don't really know junit 5 well enough to comment about the implementation. But:

1. is it really necessary to introduce `@ParallelTest`? Couldn't we make `@Test` be parallel by default, and use `@IsolatedTest` if we need something to run on its own?

2. What if we wanted to use some other lock? Does the locking be overridden on a per test basis?
3. Is there some way to base the degree of parallelism/#threads on the number of cores, rather than picking an arbitrary number?

see-quick:
> is it really necessary to introduce @ParallelTest? Couldn't we make @Test be parallel by default, and use @IsolatedTest if we need something to run on its own?

I think from the user perspective definitely. I followed convention, which I Introduced in system tests so I wanted to make all tests consistent. I see your point but you have seen that these types of tests will run in parallel. Otherwise, if I will let `@Test` annotation somebody can think these tests are in sequence mode because they do not know about `junit.properties` file

> What if we wanted to use some other lock? Does the locking be overridden on a per-test basis?

I think you can write your own lock as I have to implement my own for `IsolatedTest`.  Moreover,  I think you can also implement some multicast lock that will lock a group of tests for synchronization...

> Is there some way to base the degree of parallelism/#threads on the number of cores, rather than picking an arbitrary number?

There are few strategies I guess. For instance `dynamic` strategy computes the desired parallelism based on the number of available processors/cores multiplied by the junit.jupiter.execution.parallel.config.dynamic.factor configuration parameter (defaults to 1). What is also worth mentioning is the custom strategy if we wanted more crazy behaviour...

see-quick:
@tombentley

tombentley:
> I think you can write your own lock as I have to implement my own for `IsolatedTest`. Moreover, I think you can also implement some multicast lock that will lock a group of tests for synchronization...

I guess I meant whether it was possible to do this without having to define an annotation type. For example, could we annotate with `@ParallelTest` but override the locking by using `@ResourceLock` on the test method/class. I'm fine if the answer is no, since defining an annotation type isn't so very hard. I was just wondering.

My gut feeling is that `dynamic` makes for a better default that any hard-coded number.

tombentley (approved):

sknot-rh (approved):

see-quick:
@tombentley 
> For example, could we annotate with @ParallelTest but override the locking by using @ResourceLock on the test method/class

Yes, you can `override lock` it as you describe... ).

PR #4860, opened 2021-04-30
 [MO] - [unit + integration co tests] -> method wide parallelism

see-quick:
### Type of change

- Enhancement / new feature
- Refactoring

### Description

This PR is a continuation of https://github.com/strimzi/strimzi-kafka-operator/pull/4858. Specifically covering all `operator/assembly` package and there is some re-work in test cases. Most of these problems are because all test cases used shared object, which causes race conditions. I have re-worked it to use the local object for each test case which is thread-safe and there is no chance that race condition will appear in this scenario.

### Checklist


see-quick:
Hmmm, it seems that the executor (which is implemented inside junit5 parallelism) cannot take any more tasks under its responsibility. This happens when the limit of its local “memory” is reached....I will try maybe to not run in parallel (Parametrized tests...)

scholzj (approved):

tombentley (approved):

Issue #4872, opened 2021-04-30
NPE error whilst handling failure to roll


k-wall:
**Describe the bug**

Whilst investigating a Strimzi failure case, I noticed the following NPE during a roll pod attempt which I would like to report.  The stack trace is from a derivative built from 0.22.1. Line numbers are out by 1.

 ```
INFO 2021-04-29 17:30:53 INFO  KafkaRoller:289 - Reconciliation #149(watch) Kafka(foo-orzjkppsvfzatbfw/foo-orzjkppsvfzatbfw): Could not roll pod 2, giving up after 10 attempts. Total delay between attempts 127750ms
java.lang.NullPointerException: null
	at io.strimzi.operator.cluster.operator.resource.KafkaRoller.restartPlan(KafkaRoller.java:457)
	at io.strimzi.operator.cluster.operator.resource.KafkaRoller.restartIfNecessary(KafkaRoller.java:351)
	at io.strimzi.operator.cluster.operator.resource.KafkaRoller.lambda$schedule$5(KafkaRoller.java:274)
	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:515) [?:?]
	at java.util.concurrent.FutureTask.run(FutureTask.java:264) [?:?]
	at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.run(ScheduledThreadPoolExecutor.java:304) [?:?]
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128) [?:?]
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628) [?:?]
	at java.lang.Thread.run(Thread.java:829) [?:?]
```

This is the line in `io.strimzi.operator.cluster.operator.resource.KafkaRoller#restartPlan` where the NPE occurs is:

```
        if (podStuck && !reasonToRestartPod.contains("Pod has old generation")) {
```

It occurs when reasonToRestartPod is null. The code's handling of reasonToRestartPod is inconsistent as it later is guards the reasonToRestartPod == null case.

I think either:

* the `if` should guard the `null` case too, or
* in the anonymous function within `io.strimzi.operator.cluster.operator.assembly.KafkaAssemblyOperator.ReconciliationState#maybeRollKafkaInSequence()` the empty list should be returned rather than null.

**To Reproduce**

I don't currently have a way to reliably reproduce the problem.

[...]

PR #4881, opened 2021-05-03
Fix NPE when no reasons to roll kafka pods

sknot-rh:

### Type of change
- Bugfix

### Description

The handling of situations where no reasons for rolling kafka pods was inconsistent. We should use only null or emptyList, not both.

Fixes https://github.com/strimzi/strimzi-kafka-operator/issues/4872

### Checklist
- [ ] Write tests
- [ ] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/resource/KafkaRoller.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/resource/KafkaRoller.java
@@ -14,6 +14,7 @@ import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
 import java.util.concurrent.CompletableFuture;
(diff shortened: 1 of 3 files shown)


Frawless (approved):

scholzj (approved):

ppatierno (approved):

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/resource/KafkaRoller.java
```diff
@@ -460,7 +460,7 @@ private RestartPlan restartPlan(int podId, Pod pod, RestartContext restartContex
             throw new FatalProblem("Pod is unschedulable");
         }
         // Unless the annotation is present, check the pod is at least ready.
-        boolean needsRestart = reasonToRestartPod != null && !reasonToRestartPod.isEmpty();
+        boolean needsRestart = !reasonToRestartPod.isEmpty();
```
tombentley, line 463: Can we stick an `Objects.requireNonNull` around the `List<String> reasonToRestartPod = podNeedsRestart.apply(pod);` a few lines above, so if a function does return null we have a nice obvious NPE?

tombentley (approved):
One nit, but LGTM.

PR #4889, opened 2021-05-03
ST: Add check for MM2 connectors readiness

Frawless:
### Type of change

- Enhancement / new feature

### Description

This PR enhances the current MM2 status test with additional check for its connectors.

### Checklist


scholzj (commented):
LGTM assuming it works. Are you sure the connectors will be always there? Or do we need to check their count or something.

Frawless:
Good point....I think there always should be connectors created when mm2 CR is ready, but tbh not sure how many. I think I always saw 3 or 4.

scholzj:
Should be 3 normally ... source, checkpoint and heartbeat => they correspond to the sections in the CR.

Frawless:
Ok, I will add check that there are 3 connectors in the status

scholzj (approved):

PR #4895, opened 2021-05-04
Dual Stack networking support

scholzj:
### Type of change

- Enhancement / new feature

### Description

This PR adds better support for dual stack networking (IPv4+IPv6). It lets users to choose the preferred policy and the preferred stack. This has to be done on two places:
* All internal Kafka listeners (both `type: internal` but also replication and control plane listeners share the `...-kafka-bootstrap` and `...-kafka-brokers` services), as well as Zookeeper, Connect, CruiseControl etc., have this configuration in the `template` section.
* External Kafka listener have their own services on a per listener basis and have this configuration in the `listeners` section to allow different configuration for each listener.

This PR also adds support for Enum based Lists to our CRDs and updates the `ServiceOperator` to handle the dual stack networking defaults in case user doesn't configure anything (since in that case, Kubernetes will pick some defaults on their own).

### Checklist

- [x] Write tests
- [x] Update documentation
- [x] Update CHANGELOG.md

### DIFF
--- a/api/src/main/java/io/strimzi/api/kafka/model/listener/arraylistener/GenericKafkaListenerConfiguration.java
+++ b/api/src/main/java/io/strimzi/api/kafka/model/listener/arraylistener/GenericKafkaListenerConfiguration.java
@@ -12,8 +12,11 @@ import io.strimzi.api.kafka.model.Constants;
 import io.strimzi.api.kafka.model.UnknownPropertyPreserving;
 import io.strimzi.api.kafka.model.listener.NodeAddressType;
 import io.strimzi.api.kafka.model.template.ExternalTrafficPolicy;
+import io.strimzi.api.kafka.model.template.IpFamily;
+import io.strimzi.api.kafka.model.template.IpFamilyPolicy;
 import io.strimzi.crdgenerator.annotations.Description;
 import io.strimzi.crdgenerator.annotations.DescriptionFile;
+import io.strimzi.crdgenerator.annotations.PresentInVersions;
 import io.sundr.builder.annotations.Buildable;
 import lombok.EqualsAndHashCode;
 
@@ -28,7 +31,8 @@ import static java.util.Collections.emptyMap;
  * Configures Kafka listeners
  */
 @DescriptionFile
-@JsonPropertyOrder({"brokerCertChainAndKey", "ingressClass", "preferredAddressType", "externalTrafficPolicy", "loadBalancerSourceRanges", "bootstrap", "brokers"})
+@JsonPropertyOrder({"brokerCertChainAndKey", "ingressClass", "preferredAddressType", "externalTrafficPolicy",
+        "loadBalancerSourceRanges", "bootstrap", "brokers", "ipFamilyPolicy", "ipFamilies"})
 @JsonInclude(JsonInclude.Include.NON_NULL)
 @Buildable(
     editableEnabled = false,
@@ -49,6 +53,8 @@ public class GenericKafkaListenerConfiguration implements Serializable, UnknownP
     private List<GenericKafkaListenerConfigurationBroker> brokers;
     private Integer maxConnections;
     private Integer maxConnectionCreationRate;
+    private IpFamilyPolicy ipFamilyPolicy;
+    private List<IpFamily> ipFamilies;
 
     private Map<String, Object> additionalProperties = new HashMap<>(0);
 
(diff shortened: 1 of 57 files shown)


api/src/main/java/io/strimzi/api/kafka/model/template/InternalServiceTemplate.java
```diff
@@ -0,0 +1,89 @@
...
+import io.sundr.builder.annotations.Buildable;
+import lombok.EqualsAndHashCode;
+
+import java.io.Serializable;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+/**
+ * Representation of a template for Strimzi external services.
```
tombentley, line 22: Did you mean internal services here?

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaCluster.java
```diff
@@ -798,7 +790,8 @@ public Service generateService() {
                     ports,
                     getLabelsWithStrimziName(name, Util.mergeLabelsOrAnnotations(templateExternalBootstrapServiceLabels, ListenersUtils.bootstrapLabels(listener))),
                     getSelectorLabels(),
-                    Util.mergeLabelsOrAnnotations(ListenersUtils.bootstrapAnnotations(listener), templateExternalBootstrapServiceAnnotations)
+                    Util.mergeLabelsOrAnnotations(ListenersUtils.bootstrapAnnotations(listener), templateExternalBootstrapServiceAnnotations),
+                    ListenersUtils.ipFamilyPolicy(listener), ListenersUtils.ipFamilies(listener)
```
tombentley, line 794: Can we put the args of different lines?

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaCluster.java
```diff
@@ -864,7 +857,8 @@ public Service generateService() {
                     ports,
                     getLabelsWithStrimziName(name, Util.mergeLabelsOrAnnotations(templatePerPodServiceLabels, ListenersUtils.brokerLabels(listener, pod))),
                     selector,
-                    Util.mergeLabelsOrAnnotations(ListenersUtils.brokerAnnotations(listener, pod), templatePerPodServiceAnnotations)
+                    Util.mergeLabelsOrAnnotations(ListenersUtils.brokerAnnotations(listener, pod), templatePerPodServiceAnnotations),
+                    ListenersUtils.ipFamilyPolicy(listener), ListenersUtils.ipFamilies(listener)
```
tombentley, line 861: Can we put the args of different lines?

tombentley (commented):
Could we add a test for the list-of-enum schema generation?

Did you test jmxtrans (as an example of something we don't religiously test but which binds to addresses, maybe there are others too).

scholzj:
> Could we add a test for the list-of-enum schema generation?

Thanks I forgot about these.

> Did you test jmxtrans (as an example of something we don't religiously test but which binds to addresses, maybe there are others too).

I did not. TBH, I have no idea how it works. But this PR does not change anything, it just gives users more options. The _default_ values without this will be still the same as after this PR. So what worked before should keep working. And if anything has issues with IPv6 or DualStack, this gives users a way how to control it.

tombentley (approved):

api/src/main/java/io/strimzi/api/kafka/model/listener/arraylistener/GenericKafkaListenerConfiguration.java
```diff
@@ -193,6 +199,36 @@ public void setMaxConnectionCreationRate(Integer maxConnectionCreationRate) {
         this.maxConnectionCreationRate = maxConnectionCreationRate;
     }
 
+    @Description("Specifies which IP Family Policy should be used by this service. " +
+            "Available options are `SingleStack` (a single IP family), " +
+            "`PreferDualStack` (two IP families on dual-stack configured clusters or a single IP family on single-stack clusters), " +
+            "and `RequireDualStack` (two IP families on dual-stack configured clusters, otherwise fail). " +
+            "If unspecified, Kubernetes will choose the default value based on the service type. " +
+            "Available on Kubernetes 1.20 and newer.")
```
PaulRMellor, line 207:
```suggestion
    @Description("Specifies the IP Family Policy used by the service. " +
            "Available options are `SingleStack`, `PreferDualStack` and `RequireDualStack`. " +
            "`SingleStack` is for a single IP family. " +
            "`PreferDualStack` is for two IP families on dual-stack configured clusters or a single IP family on single-stack clusters. " +
            "`RequireDualStack` fails unless there are two IP families on dual-stack configured clusters. " +
            "If unspecified, Kubernetes will choose the default value based on the service type. " +
            "Available on Kubernetes 1.20 and newer.")
```

api/src/main/java/io/strimzi/api/kafka/model/listener/arraylistener/GenericKafkaListenerConfiguration.java
```diff
@@ -193,6 +199,36 @@ public void setMaxConnectionCreationRate(Integer maxConnectionCreationRate) {
...
+    }
+
+    public void setIpFamilyPolicy(IpFamilyPolicy ipFamilyPolicy) {
+        this.ipFamilyPolicy = ipFamilyPolicy;
+    }
+
+    @Description("Specifies which IP Families should be used by this service. " +
+            "Available options are `IPv4` and `IPv6. " +
+            "If unspecified, Kubernetes will choose the default value based on the `ipFamilyPolicy` setting. " +
+            "Available on Kubernetes 1.20 and newer.")
```
PaulRMellor, line 221:
```suggestion
    @Description("Specifies the IP Families used by the service. " +
            "Available options are `IPv4` and `IPv6. " +
            "If unspecified, Kubernetes will choose the default value based on the `ipFamilyPolicy` setting. " +
            "Available on Kubernetes 1.20 and newer.")
```

PaulRMellor (approved):

PR #4900, opened 2021-05-05
[systemtest] Make logging of errors when resource is not created clearer

im-konge:

### Type of change

- Fix

### Description

This PR fixes logging when the resource is not created or is currently `null` when we are waiting for readiness.

When the resource is not created and we reach the timeout, we get this kind of error:
```
2021-04-16T12:58:50.0249926Z io.strimzi.test.WaitException: Timeout after 180000 ms waiting for Resource condition:io.strimzi.systemtest.resources.ResourceManager$$Lambda$659/0x0000000840872c40@569b65b8 is fulfilled 
```
which will not tell us on which condition we are really waiting.

Because the `Predicate<T>` doesn't contain any information about method name that he "carries", I had to create this encapsulation.

After this, the log message looks clearer:
```
2021-05-05 08:32:31 [main] ERROR [TestUtils:138] Exception waiting for Resource condition: readiness is fulfilled for resource KafkaTopic:some-topic
```

Also, this PR fixes behavior when we just started with readiness wait, the resource is not created and we checking it's status. Then this line appears:
```
2021-04-16 14:28:41 [main] ERROR [TestUtils:138] Exception waiting for Wait for Kafka: my-cluster-1951300410 will have desired state: Ready, null
```

### Checklist

### DIFF
--- a/test/src/main/java/io/strimzi/test/TestUtils.java
+++ b/test/src/main/java/io/strimzi/test/TestUtils.java
@@ -133,7 +133,7 @@ public final class TestUtils {
                 result = ready.getAsBoolean();
             } catch (Exception e) {
                 exceptionMessage = e.getMessage();
-                if (++exceptionCount == 1) {
+                if (++exceptionCount == 1 && exceptionMessage != null) {
                     // Log the first exception as soon as it occurs
                     LOGGER.error("Exception waiting for {}, {}", description, exceptionMessage);
                     // log the stacktrace
@@ -148,8 +148,11 @@ public final class TestUtils {
             if (timeLeft <= 0) {
                 if (exceptionCount > 1) {
                     LOGGER.error("Exception waiting for {}, {}", description, exceptionMessage);
-                    // printing handled stacktrace
-                    LOGGER.error(stackTraceError.toString());
+
+                    if (!stackTraceError.toString().isEmpty()) {
+                        // printing handled stacktrace
+                        LOGGER.error(stackTraceError.toString());
+                    }
                 }
                 onTimeout.run();
                 WaitException waitException = new WaitException("Timeout after " + timeoutMs + " ms waiting for " + description);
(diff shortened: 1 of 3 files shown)


test/src/main/java/io/strimzi/test/TestUtils.java
```diff
@@ -133,7 +133,7 @@ public static long waitFor(String description, long pollIntervalMs, long timeout
                 result = ready.getAsBoolean();
             } catch (Exception e) {
                 exceptionMessage = e.getMessage();
-                if (++exceptionCount == 1) {
+                if (++exceptionCount == 1 && exceptionMessage != null) {
```
Frawless, line 136: I think you should print a warning even if the `exceptionMessage` is null and print `exceptionMessage` only if it's filled.

test/src/main/java/io/strimzi/test/TestUtils.java
```diff
@@ -133,7 +133,7 @@ public static long waitFor(String description, long pollIntervalMs, long timeout
...
                     // Log the first exception as soon as it occurs
                     LOGGER.error("Exception waiting for {}, {}", description, exceptionMessage);
```
scholzj, line 138: Should this be really ERROR level? I noticed it yesterday and found it very confusing because this _ERROR_ seemed to be in many test cases without meaning anything.

see-quick: I think you mean something like this...:
```
2021-05-05 12:23:12 [ForkJoinPool-1-worker-11] ERROR [TestUtils:138] Exception waiting for Wait for Kafka: my-cluster-1594415598 will have desired state: Ready, null
```
but that will disappear after `exceptionMessage != null`.

scholzj: Ok, than maybe it is covered 🤷‍♂️

scholzj (approved):

see-quick (approved):

PR #4903, opened 2021-05-05
Make (finished) KafkaRebalance in Ready state refreshable

### LINKED ISSUE #3296: [Enhancement] Should KafkaRebalance custom resource status `Ready` mark end of usable life?
From reading the documentation it was unclear that once a `KafkaRebalance` proposal is successful and the status turns `Ready` you are unable to get a new proposal from that `KafkaRebalance` custom resource. It was only through looking through the [code](https://github.com/strimzi/strimzi-kafka-operator/blob/2e950039d7559ddd7ce6d46aa586f0e66b50e702/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceAssemblyOperator.java#L368-L370) that we noticed that once the status is `Ready` that you won't be able to generate a new proposal from this custom resource.  

This seems to be intentional but we were wondering if this should be so? We were wondering if there was any compelling reason that you shouldn't be able to use the same custom resource for generating a new proposal? This would also make keeping a KafkaRebalance custom resource in your cluster make more sense as currently all it does is act as an old artefact of a previous rebalance.

tomncooper:
### Type of change

- Enhancement / new feature
- Documentation

### Description

This PR closes #3296.

Currently, when a Cruise Control rebalance completes, the `KafkaRebalance` CR will be in the `Ready` state. This state, unlike the other end states of `NotReady` and `Stopped` cannot be "refreshed". The user has to delete the `KafkaRebalance` (possibly with all their custom settings) and create a new one if they want to reuse all the same settings.

This PR allow a user to "refresh" a finished (`Ready`) `KafkaRebalance` CR which will send a new rebalance request to Cruise Control and move the CR into `ProposalPending`/`ProposalReady` ready for approval. 

### Checklist

- [x] Write tests
- [x] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/cluster-operator/src/test/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceAssemblyOperatorTest.java
+++ b/cluster-operator/src/test/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceAssemblyOperatorTest.java
@@ -436,6 +436,82 @@ public class KafkaRebalanceAssemblyOperatorTest {
     }
 
     /**
+     * Tests the transition from 'New' to 'ProposalReady'
+     * The rebalance proposal is approved and the resource moves to 'Rebalancing' and finally to 'Ready'
+     * Then the Ready KafkaRebalance is refreshed and a moved to 'ProposalReady' again.
+     *
+     * 1. A new KafkaRebalance resource is created; it is in the 'New' state
+     * 2. The operator requests a rebalance proposal through the Cruise Control REST API
+     * 3. The rebalance proposal is ready on the first call
+     * 4. The KafkaRebalance resource transitions to the 'ProposalReady' state
+     * 5. The KafkaRebalance resource is annotated with 'strimzi.io/rebalance=approve'
+     * 6. The operator requests the rebalance operation through the Cruise Control REST API
+     * 7. The rebalance operation is not done immediately; the operator starts polling the Cruise Control REST API
+     * 8. The KafkaRebalance resource moves to the 'Rebalancing' state
+     * 9. The rebalance operation is done
+     * 10. The KafkaRebalance resource moves to the 'Ready' state
+     * 11. The KafkaRebalance resource is annotated with 'strimzi.io/rebalance=refresh'
+     * 12. The KafkaRebalance resource moves to the 'ProposalReady' state
+     */
+    @Test
+    public void testNewToProposalReadyToRebalancingToReadyThenRefreshRebalance(VertxTestContext context) throws IOException, URISyntaxException {
+
+        // Setup the rebalance and user tasks endpoints with the number of pending calls before a response is received.
+        MockCruiseControl.setupCCRebalanceResponse(ccServer, 0);
+        MockCruiseControl.setupCCUserTasksResponseNoGoals(ccServer, 0, 0);
+
+        KafkaRebalance kr =
+                createKafkaRebalance(CLUSTER_NAMESPACE, CLUSTER_NAME, RESOURCE_NAME, new KafkaRebalanceSpecBuilder().build());
+
+        Crds.kafkaRebalanceOperation(kubernetesClient).inNamespace(CLUSTER_NAMESPACE).create(kr);
+
+        when(mockKafkaOps.getAsync(CLUSTER_NAMESPACE, CLUSTER_NAME)).thenReturn(Future.succeededFuture(kafka));
+        mockRebalanceOperator(mockRebalanceOps, CLUSTER_NAMESPACE, RESOURCE_NAME, kubernetesClient);
+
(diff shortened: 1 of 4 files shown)


scholzj (approved):

cluster-operator/src/test/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceAssemblyOperatorTest.java
```diff
@@ -435,6 +435,82 @@ public void testNewToProposalReadyToRebalancingToReadyRebalance(VertxTestContext
             }));
     }
 
+    /**
+     * Tests the transition from 'New' to to 'ProposalReady'
```
see-quick, line 439:
```suggestion
     * Tests the transition from 'New' to 'ProposalReady'
```
typo I think :D

cluster-operator/src/test/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceAssemblyOperatorTest.java
```diff
@@ -435,6 +435,82 @@ public void testNewToProposalReadyToRebalancingToReadyRebalance(VertxTestContext
...
+     * The rebalance proposal is approved and the resource moves to 'Rebalancing' and finally to 'Ready'
+     * Then the Ready KafkaRebalance is refreshed and a moved to 'ProposalReady' again.
+     *
+     * 1. A new KafkaRebalance resource is created; it is in the 'New' state
+     * 2. The operator requests a rebalance proposal through the Cruise Control REST API
+     * 3. The rebalance proposal is ready on the first call
+     * 4. The KafkaRebalance resource transitions to the 'ProposalReady' state
+     * 9. The KafkaRebalance resource is annotated with 'strimzi.io/rebalance=approve'
```
see-quick, line 447: Do these numbers refer to some state number or it is just leftover and it should be 4.....5. etc?

cluster-operator/src/test/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceStateMachineTest.java
```diff
@@ -504,6 +504,72 @@ public void testStoppedRefreshToPendingProposalNotEnoughData(Vertx vertx, VertxT
 
     }
 
+    @Test
```
see-quick, line 507: If I will be quick enough (https://github.com/strimzi/strimzi-kafka-operator/pull/4860). You can use parallel annotation to make these tests parallel.

see-quick (approved):

ppatierno (approved):

documentation/modules/cruise-control/proc-approving-optimization-proposal.adoc
```diff
@@ -68,7 +68,9 @@ kubectl describe kafkarebalance _rebalance-cr-name_
 
 ** Rebalancing: The cluster rebalance operation is in progress. 
 
-** Ready: The cluster rebalancing operation completed successfully. The `KafkaRebalance` custom resource cannot be reused.
+** Ready: The cluster rebalancing operation completed successfully.
+The `KafkaRebalance` custom resource can be re-used by annotating the resource with `refresh`.
```
laidan6000, line 72:
```suggestion
To use the same `KafkaRebalance` custom resource to generate another optimization proposal, apply the `refresh` annotation to the custom resource. 
```

documentation/modules/cruise-control/proc-approving-optimization-proposal.adoc
```diff
@@ -68,7 +68,9 @@ kubectl describe kafkarebalance _rebalance-cr-name_
...
+This will move the resource to the `ProposalPending`/`ProposalReady` state where the relevant instructions above should then be followed.
```
laidan6000, line 73:
```suggestion
This moves the custom resource to the `PendingProposal` or `ProposalReady` state. You can then review the optimization proposal and approve it, if desired.
```

laidan6000 (changes requested):
Made a few changes to make the phrasing consistent with the existing doc.
Also, I think the state is PendingProposal.

cluster-operator/src/test/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceStateMachineTest.java
tomncooper: Might be better to leave that change to another PR?

laidan6000 (commented):
Looks good!

PR #4910, opened 2021-05-06
[MO] - [system tests] -> removing unnecessarity test cases for host alias

see-quick:
### Type of change

- Refactoring

### Description

This PR removing unnecessarily test cases for host aliases. The main reason is that all these cases are covered in the UTs and there is no need to do it at the system level. The procedure is to inject host aliases settings in the components CRs and afterwards, is to check inside the specific pods.

### Checklist

### DIFF
--- a/systemtest/src/test/java/io/strimzi/systemtest/AbstractST.java
+++ b/systemtest/src/test/java/io/strimzi/systemtest/AbstractST.java
@@ -122,11 +122,6 @@ public abstract class AbstractST implements TestSeparator {
     public static final String USER_NAME = KafkaUserUtils.generateRandomNameOfKafkaUser();
     public static final String TOPIC_NAME = KafkaTopicUtils.generateRandomNameOfTopic();
 
-    // Constants for host aliases tests
-    protected final String aliasIp = "34.89.152.196";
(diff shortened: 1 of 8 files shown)


Frawless (approved):

scholzj (approved):
Is it really worth removing them if we already have them? Should we maybe just use them at releases or something to not waste time in regular runs but keep the code?

see-quick:
@scholzj then we can introduce new `@Tag` something like `RELEASE_ONLY` and filter them. But I think it will only cause only problems with rising questions like how to improve this code...and maybe for newcomers cause misunderstanding and they will create same scenarios, which are not good I would say.

im-konge:
@scholzj I remember we discussed this option with @Frawless, I found just 2 tests which should be moved under some `RELEASE` tag, which is not quite lot -> we agreed that execution time will be decreased by the parallelism, so we don't need it.

im-konge (approved):

scholzj:
Didn't we already had tests like that? At least I know that it was discussed mnultiple times in the past.

im-konge:
I remember  `testConfigureDeploymentStrategy` and these host aliases tests. I don't think there are any other, but I would have to make sure.

PR #4925, opened 2021-05-07
[systemtest] Split Kafka clusters to separate namespaces in MetricsST

im-konge:

### Type of change

- Fix

### Description

Currently, we have two Kafka clusters deployed in one namespace. That can cause some issues - like topic store which is connected to second Kafka cluster, so when we collect metrics for the first Kafka cluster and we want to check number of topics, the check fails, because store topic is missing.

This PR separates these two Kafka clusters to each own namespace and fixes tests which are affected by changes to first or second Kafka cluster.

### Checklist


Frawless (approved):

Frawless:
I think we should cherry-pick this to `release-0.23.x`

PR #4945, opened 2021-05-11
Support pausing KafkaTopic reconciliations

sknot-rh:

### Type of change
- new feature

### Description
We had to revert https://github.com/strimzi/strimzi-kafka-operator/pull/4546 because issues with status. These issues were fixed in https://github.com/strimzi/strimzi-kafka-operator/pull/4727 so we can reopen this.

### Checklist
- [x] Write tests
- [ ] Update documentation
- [x] Update CHANGELOG.md

### DIFF
--- a/topic-operator/src/main/java/io/strimzi/operator/topic/TopicOperator.java
+++ b/topic-operator/src/main/java/io/strimzi/operator/topic/TopicOperator.java
@@ -17,6 +17,7 @@ import io.strimzi.api.kafka.model.KafkaTopic;
 import io.strimzi.api.kafka.model.KafkaTopicBuilder;
 import io.strimzi.api.kafka.model.status.KafkaTopicStatus;
 import io.strimzi.operator.cluster.model.StatusDiff;
+import io.strimzi.operator.common.Annotations;
 import io.strimzi.operator.common.BackOff;
 import io.strimzi.operator.common.MaxAttemptsExceededException;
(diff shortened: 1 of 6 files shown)


topic-operator/src/main/java/io/strimzi/operator/topic/TopicOperator.java
```diff
@@ -1309,8 +1323,12 @@ public void setKafkaTopics(List<KafkaTopic> ktList) {
             });
         }).compose(reconcileState -> {
             List<Future> futs = new ArrayList<>();
+            pausedTopicCounter.set(0);
```
tombentley, line 1326: How can we be sure it won't get out of sync (e.g. with event-based reconciliation happening concurrently)?

sknot-rh: The count is refreshed with each periodic reconciliation. So there may be the "lag" when paused topic is created or the topic is paused. This lag should be at max the periodic reconciliation time. Do you think we should decrease/increase the counter everytime the topic is un/paused?

tombentley (commented):
Can we have some test coverage for the metric. It doesn't have to be a new test, just adding assertions to some existing tests. There are actually a few different cases worth testing there.

tombentley (approved):

ppatierno (approved):

PR #4949, opened 2021-05-12
[DOC] Kafka Exporter link updates

PaulRMellor:
**Documentation**

Updates to links for _Configuring Kafka Exporter_ section in the _Deploying_ doc.

### Checklist

- [ ] Write tests
- [x] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/documentation/modules/metrics/kafka-exporter/proc_kafka-exporter-configuring.adoc
+++ b/documentation/modules/metrics/kafka-exporter/proc_kafka-exporter-configuring.adoc
@@ -33,7 +33,7 @@ metadata:
 spec:
   # ...
   kafkaExporter:
-    image: my-org/my-image:latest <1>
+    image: my-registry.io/my-org/my-exporter-cluster:latest <1>
     groupRegex: ".*" <2>
     topicRegex: ".*" <3>
     resources: <4>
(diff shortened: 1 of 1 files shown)


PaulRMellor:
@scholzj I've been asked about replacing the example `image` with _working_ image name.

scholzj:
@PaulRMellor I think there is a reason for having an _example_ image name there. The real image name does not need to be specified there, so having it there would be misleading. So if the example image name is confusing, I guess the alternative would be something what is a bit more clear placeholder.

PaulRMellor:
> @PaulRMellor I think there is a reason for having an _example_ image name there. The real image name does not need to be specified there, so having it there would be misleading. So if the example image name is confusing, I guess the alternative would be something what is a bit more clear placeholder.

@scholzj - thanks. Maybe we could change to something like _my-registry.io/my-org/my-exporter-cluster:latest_
(`my-org/my-image:latest` is used in a few places)

scholzj:
@PaulRMellor `my-registry.io/my-org/my-exporter-cluster:latest` would be totally fine for me.

scholzj (approved):

laidan6000 (approved):

PR #4963, opened 2021-05-13
Configuration for kubernetes-ops-pool

rareddy:

### Type of change
- Bugfix
- Enhancement

### Description

Added support to configure the thread pool size using an environment variable for `kubernetes-ops-pool`, which is used for various sync and async operations. The env variable is "STRIMZI_OPERATIONS_THREAD_POOL_SIZE". The default size kept at 10 threads.

Also corrected usage of pool `kubernetes-ops-tool` which seems to be bug, replaced all instances of usage of this thread pool to use `kubernetes-ops-pool`. Effectively there is only single thread pool now in use instead of two.

### DIFF
--- a/.checkstyle/checkstyle.xml
+++ b/.checkstyle/checkstyle.xml
@@ -94,7 +94,7 @@
         <module name="MethodLength"/>
         <module name="ParameterNumber">
             <!-- default is 8 -->
-            <property name="max" value="14"/>
+            <property name="max" value="13"/>
         </module>
         <module name="ClassDataAbstractionCoupling">
             <!-- default is 7 -->
--- a/.checkstyle/checkstyle.xml
+++ b/.checkstyle/checkstyle.xml
@@ -94,7 +94,7 @@
         <module name="MethodLength"/>
         <module name="ParameterNumber">
             <!-- default is 8 -->
-            <property name="max" value="13"/>
+            <property name="max" value="14"/>
         </module>
         <module name="ClassDataAbstractionCoupling">
             <!-- default is 7 -->
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## 0.24.0
 
 * Add support for Kafka 2.7.1 and remove support for 2.6.0, 2.6.1, and 2.6.2
+* Added support to configure the thread pool size for `kubernetes-ops-pool`
 
 ## 0.23.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 ## 0.24.0
 
 * Add support for Kafka 2.7.1 and remove support for 2.6.0, 2.6.1, and 2.6.2
-* Added support to configure the worker thread pool size
+* Added support for configuring cluster-operator's worker thread pool size that is used for various sync and async tasks
  
 ## 0.23.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,8 +3,8 @@
 ## 0.24.0
 
 * Add support for Kafka 2.7.1 and remove support for 2.6.0, 2.6.1, and 2.6.2
-* Added support to configure the thread pool size for `kubernetes-ops-pool`
-
+* Added support to configure the worker thread pool size
+ 
 ## 0.23.0
 
 * Add support for Kafka 2.8.0 and 2.6.2, remove support for Kafka 2.5.x
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/ClusterOperatorConfig.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/ClusterOperatorConfig.java
@@ -107,6 +107,7 @@ public class ClusterOperatorConfig {
      * @param featureGates Configuration string with feature gates settings
      * @param operationsThreadPoolSize The size of the thread pool used for various operations
      */
+    @SuppressWarnings("checkstyle:ParameterNumber")
     public ClusterOperatorConfig(
             Set<String> namespaces,
             long reconciliationIntervalMs,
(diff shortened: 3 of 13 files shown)


.checkstyle/checkstyle.xml
```diff
@@ -94,7 +94,7 @@
         <module name="MethodLength"/>
         <module name="ParameterNumber">
             <!-- default is 8 -->
-            <property name="max" value="13"/>
+            <property name="max" value="14"/>
```
scholzj, line 97: Can you please just use annotation to surpress the warning instead of changing the `checkstyle.xml`?

CHANGELOG.md
```diff
@@ -2,7 +2,7 @@
 
 ## 0.24.0
 
-
+ * Added support to configure the thread pool size for `kubernetes-ops-pool`
```
scholzj, line 5: Please add empty line after the bullet point. Maybe you could also make it a bit more generic? Nobody will know what does this mean.

```suggestion
 * Added support to configure the worker thread pool size
 
```

install/cluster-operator/060-Deployment-strimzi-cluster-operator.yaml
```diff
@@ -42,6 +42,8 @@ spec:
               valueFrom:
                 fieldRef:
                   fieldPath: metadata.namespace
+            - name: STRIMZI_OPERATIONS_THREAD_POOL_SIZE
+              value: "10"
```
scholzj, line 46: * The `install/` files are released installation files. You should not change them as a regular PRs. The once to be changed are in `packaging/` (+ the 060... file is actually generated from the Helm Chart)
* But I do not think we expect the users to change these. So I would not add the defaults to the deployments anyway. I would just remove it from here.

documentation/modules/managing/con-cluster-operator-faqs.adoc
```diff
@@ -105,3 +105,8 @@ When configuring the client directly in Java, set the configuration option to an
 
 [source,java]
 props.put("ssl.endpoint.identification.algorithm", "");
+
+
+=== Control the thread pool size of `kubernetes-ops-pool`
+
+`kubernetes-ops-pool` thread pool is used for various async and blocking operations by the cluster operator. The default of the pool is set at 10. In rare situations if one need to adjust the thread pool size they can set a environment variable `STRIMZI_OPERATIONS_THREAD_POOL_SIZE` to resize the pool size.
```
scholzj, line 112: No problem with having this in the FAQ. But nobody will find it here. Please add this to the `ref-operator-cluater.adoc` file.

scholzj (approved):

cluster-operator/src/main/java/io/strimzi/operator/cluster/ClusterOperatorConfig.java
```diff
@@ -233,6 +242,16 @@ private static long parseTimeout(String timeoutEnvVar, long defaultTimeout) {
         return timeout;
     }
 
+    private static int parseInt(String envVar, int defaultValue) {
+        int defalt = defaultValue;
```
ppatierno, line 246: minor typo "defalt"

CHANGELOG.md
```diff
@@ -2,7 +2,8 @@
 
 ## 0.24.0
 
-
+ * Added support to configure the worker thread pool size
```
ppatierno, line 5: I would add a little bit more of context on how this pool is used. Just a few more info as in the section where you describe the corresponding env var.

scholzj: Wouldn't it just confuse users who until today had absolutely no clue there is some worker pool? Are you ready to answer the questions how should it be configured etc.?

ppatierno: So I guess you are suggesting to remove it? I would agree on it, but if we leave the description we should be more precise on what "worker thread pool" is.

scholzj: I was thinking about it when doing my review. I do not have problem with it being there like this. My issue is that the CHANGELOG needs to be short and I'm afraid that by adding more details you will not be able to explain it and will just confuse people into thinking it is something important they should care about. Just my personal view of course.

ppatierno: Tbh as it is now, I would remove it because it doesn't explain nothing clearly. It would raise more questions.

cluster-operator/src/main/java/io/strimzi/operator/cluster/ClusterOperatorConfig.java
rareddy: not a typo, since `default` is a reserved keyword.  I can change to something else if you prefer.

CHANGELOG.md
rareddy: how about `Support for configuring cluster-operator's worker thread pool size that is used for various sync and async tasks`?

cluster-operator/src/main/java/io/strimzi/operator/cluster/ClusterOperatorConfig.java
ppatierno: right :-) yeah another name would be better then!

rareddy:
@ppatierno updated now

CHANGELOG.md
ppatierno: I am fine with that one, let's see @scholzj before making the change.

scholzj: Just approve it so that we can merge it ... ;-)

ppatierno (approved):
LGTM. Just fix conflicts first ;-)

documentation/modules/managing/con-cluster-operator-faqs.adoc
```diff
@@ -105,3 +105,4 @@ When configuring the client directly in Java, set the configuration option to an
 
 [source,java]
 props.put("ssl.endpoint.identification.algorithm", "");
+
```
sknot-rh, line 108: This empty line is somehow generated?

sknot-rh (approved):
LGTM, one nit

PR #4894, opened 2021-05-04
Re-factor ZK client config setup and close methods to use worker threads

MikeEdgar:
Related to #4855 

Signed-off-by: Michael Edgar <medgar@redhat.com>

### Type of change
- Bugfix

### Description

Additional updates to `ZookeeperScaler` (beyond #4890) to execute blocking operations using Vert.x worker threads.

### Checklist

_Please go through this checklist and make sure all applicable tasks have been done_

- [ ] Write tests
- [ ] Make sure all tests pass
- [ ] Update documentation
- [ ] Check RBAC rights for Kubernetes / OpenShift roles
- [ ] Try your changes from Pod inside your Kubernetes and OpenShift cluster, not just locally
- [ ] Reference relevant issue(s) and close them after merging
- [ ] Update CHANGELOG.md
- [ ] Supply screenshots for visual changes, such as Grafana dashboards

PR #4968, opened 2021-05-14
Handle ZooKeeper close on error; add unit test and assertions

MikeEdgar:
### Type of change

- Bugfix

### Description

Fix issue where the Zookeeper client is not closed following an unsuccessful `getConfig` or `scaleTo`, introduced with re-factoring in #4894.

### Checklist

- [X] Write tests
- [ ] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/resource/ZookeeperScaler.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/resource/ZookeeperScaler.java
@@ -99,9 +99,16 @@ public class ZookeeperScaler implements AutoCloseable {
 
                     getCurrentConfig(zkAdmin)
                             .compose(servers -> scaleTo(zkAdmin, servers, scaleTo))
-                            .compose(nothing -> closeConnection(zkAdmin))
-                            .onSuccess(scalePromise::complete)
-                            .onFailure(scalePromise::fail);
+                            .onComplete(res ->
+                                closeConnection(zkAdmin)
+                                    .onComplete(closeResult -> {
+                                        // Ignoring the result of `closeConnection`
+                                        if (res.succeeded()) {
+                                            scalePromise.complete();
+                                        } else {
+                                            scalePromise.fail(res.cause());
+                                        }
+                                    }));
 
                     return scalePromise.future();
                 });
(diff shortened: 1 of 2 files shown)


cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/resource/ZookeeperScaler.java
```diff
@@ -99,9 +99,16 @@ protected ZookeeperScaler(Vertx vertx, ZooKeeperAdminProvider zooAdminProvider,
 
                     getCurrentConfig(zkAdmin)
                             .compose(servers -> scaleTo(zkAdmin, servers, scaleTo))
-                            .compose(nothing -> closeConnection(zkAdmin))
-                            .onSuccess(scalePromise::complete)
-                            .onFailure(scalePromise::fail);
+                            .onComplete(res ->
+                                closeConnection(zkAdmin)
+                                    .onComplete(closeResult -> {
```
scholzj, line 104: Should you log something in case `closeResult` fails? I think it might be useful.

MikeEdgar: Ok, I'll add that. Do you think debug or info (or even warn) level?

scholzj: I think that unless we know for sure that any issues will print other errors from the Zoo client, then `warn` level might be good.

MikeEdgar: It's actually already logging exceptions from `ZooKeeperAdmin#close` in `closeConnection`. Maybe just bumping that up from `debug` to `warn` will do it?

scholzj: Ok, that makes sense I guess.

MikeEdgar: Done. That will cover failures here as well as any that may occur while closing from `connect`.

scholzj (approved):

ppatierno (approved):

PR #4969, opened 2021-05-14
ST: Add 0.23 to StrimziUpgradeST

Frawless:

### Type of change

- Enhancement / new feature

### Description

This PR slightly changes upgrade tests and adds 0.23 to it.

### Checklist

### DIFF
--- a/systemtest/src/test/java/io/strimzi/systemtest/upgrade/AbstractUpgradeST.java
+++ b/systemtest/src/test/java/io/strimzi/systemtest/upgrade/AbstractUpgradeST.java
@@ -566,7 +566,9 @@ public class AbstractUpgradeST extends AbstractST {
 
         LOGGER.info("Converting CRs ...");
         // run conversion of crs
-        ExecResult execResult = Exec.exec(convertorPath, "cr", "-n=" + namespace);
+        // CRs conversion may fail, because for 0.23 for example it's already done
+        // CRS conversion needs old versions of Strimzi CRDs, which are not available after 0.22
+        ExecResult execResult = Exec.exec(null, Arrays.asList(convertorPath, "cr", "-n=" + namespace), 0, true, false);
         LOGGER.debug("CRs conversion STDOUT:");
         LOGGER.debug(execResult.out());
         LOGGER.debug("CRs conversion STDERR:");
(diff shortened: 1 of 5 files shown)


systemtest/src/test/java/io/strimzi/systemtest/upgrade/AbstractUpgradeST.java
```diff
@@ -557,12 +564,24 @@ protected void convertCRDs(JsonObject conversionTool, String namespace) throws I
...
+        LOGGER.info("CRs conversion done!");
         // run crd-upgrade
         LOGGER.info("Converting CRDs");
-        Exec.exec(convertorPath, "crd");
+        execResult = Exec.exec(convertorPath, "crd");
+        LOGGER.debug("CRDs conversion STDOUT:");
+        LOGGER.debug(execResult.out());
+        LOGGER.debug("CRDs conversion STDERR:");
+        LOGGER.debug(execResult.err());
+        LOGGER.info("CRDs conversion done!");
```
scholzj, line 584: I'm not sure I follow this part ... are we just ignoring if it fails?

Frawless: yes, I am open to any better solution, but in general, CRs conversion is not working in case that 0.23 is installed, because v1alpha1 is not instaled.

scholzj: I guess we can detect it what versions the CRDs have. Or maybe we should have some flag if it should be done or not? I guess we will remove the api-conversion tool now as well, so that would cause further problems.

Frawless: ok, detect CRD version make sense. I will change it.

systemtest/src/test/java/io/strimzi/systemtest/upgrade/AbstractUpgradeST.java
```diff
@@ -557,12 +564,28 @@ protected void convertCRDs(JsonObject conversionTool, String namespace) throws I
...
+        }
         // run crd-upgrade
         LOGGER.info("Converting CRDs");
-        Exec.exec(convertorPath, "crd");
+        ExecResult execResult = Exec.exec(convertorPath, "crd");
+        LOGGER.debug("CRDs conversion STDOUT:");
+        LOGGER.debug(execResult.out());
+        LOGGER.debug("CRDs conversion STDERR:");
+        LOGGER.debug(execResult.err());
+        LOGGER.info("CRDs conversion done!");
```
scholzj, line 588: I don't think the logic of this works ... IMHO you should:
* Include this whole thing into the block
* You should check if it has something else than `v1beta2` => I doubt the Kafka CRD for example has ever `v1alpha1` there

scholzj (approved):

Issue #4833, opened 2021-04-28
Document increasing maxTasks to improve MirrorMaker2 performance


jordan-thoms:
**Suggestion / Problem**
After setting up MirrorMaker2, the mirroring speed was slow and adding more replica nodes did not help. Eventually I found that increasing the maxTasks setting on the sourceConnector helped, but that setting is only mentioned in the spec and not in the main documentation. 
**Documentation Link**
https://strimzi.io/docs/operators/latest/using.html#proc-mirrormaker-replication-str should include maxTasks like this:
```
  mirrors:
  - sourceCluster: "source-cluster"
    targetCluster: "strimzi-cluster"
    sourceConnector:
      tasksMax: 100
      config:
        replication.factor: 1
        offset-syncs.topic.replication.factor: 3
```

and suggest increasing it if the replication is not fast enough.

PR #4995, opened 2021-05-19
[DOC] Add tasksMax to the MM2 example config

PaulRMellor:

**Documentation**
Update to the _Using Guide_ to show `tasksMax' used in the example MM2 config.
Describes the property

### Checklist

- [ ] Write tests
- [ ] Update documentation
- [ ] Update CHANGELOG.md


PaulRMellor:
This PR fixes #4833

scholzj (commented):
Is it intentional to use the tasksMax only for the source connector?

laidan6000 (approved):
Clear description.

PaulRMellor:
> Is it intentional to use the tasksMax only for the source connector?

The issue mentions that it helped with source connector. Do we want to show it elsewhere?

scholzj:
> The issue mentions that it helped with source connector. Do we want to show it elsewhere?

I don't know ... TBH, I do not understand it ... that is why I was wondering.

scholzj (approved):

documentation/modules/configuring/proc-config-mirrormaker2-replication.adoc
```diff
@@ -204,33 +205,34 @@ Standard Apache Kafka configuration may be provided, restricted to those propert
...
-<20> Replication factor for mirrored topics created at the target cluster.
-<21> Replication factor for the `MirrorSourceConnector` `offset-syncs` internal topic that maps the offsets of the source and target clusters.
-<22> When xref:con-mirrormaker-acls-{context}[ACL rules synchronization] is enabled, ACLs are applied to synchronized topics. The default is `true`.
-<23> Optional setting to change the frequency of checks for new topics. The default is for a check every 10 minutes.
-<24> Defines the separator used for the renaming of remote topics.
-<25> Adds a policy that overrides the automatic renaming of remote topics. Instead of prepending the name with the name of the source cluster, the topic retains its original name. This optional setting is useful for active/passive backups and data migration.
+<20> The maximum number of tasks for the connector. Tasks handle the data replication and run in parallel. If the infrastructure supports the processing overhead, increasing this value can improve throughput. Kafka Connect distributes the workload between tasks. Kafka Connect creates fewer tasks if the maximum number of tasks is higher than the number of partitions served.
```
tombentley, line 208: "Kafka Connect creates fewer tasks if the maximum number of tasks is higher than the number of partitions served."

Is this really a thing?

PaulRMellor: @tombentley - There are a couple of answers here about ideal setting of `max.tasks`: 
https://stackoverflow.com/questions/41900694/ideal-value-for-kafka-connect-distributed-tasks-max-configuration-setting
In Kafka doc, it says of max tasks: _The connector may create fewer tasks if it cannot achieve this level of parallelism._ 

Maybe we could advise as folows: _Kafka Connect creates fewer than the maximum number of tasks if it cannot achieve the parallelism. Aim to have one task for each partition served._

tombentley: I think you're confusing the _connector_ being allowed to create fewer tasks with the connect _runtime_ doing something to cap the number of tasks. That's not how it works. In reality:

1. the runtime asks the connector to create up to `max.tasks`, 
2. the connector may return fewer configs, it should not return more configs, but it _could_ and the runtime can't do anything about that. 
3. If there are more tasks than workers then workers will be assigned multiple tasks

So at least the language is confusing because it suggests that the runtime does things it doesn't do. 

Also the word "partition" is open to misinterpretation here. When talking about a sink connector, partition == topic partition, but that's not true for a source connector, where the number of tasks which can be run in parallel depends very much on the source system and may not be related to the number of topic partitions in the topic(s) being produced to.

PaulRMellor: Thanks @tombentley How's this. Trying to keep it brief but informative.

> The maximum number of tasks created by the connector. Tasks handle the data replication and run in parallel. If the infrastructure supports the processing overhead, increasing this value can improve throughput. Kafka Connect distributes the workload between tasks. If there are more tasks than workers, workers are assigned multiple tasks. For sink connectors, aim to have one task for each topic partition served. For source connectors, the number of tasks that can run in parallel also depends on the external system. The connector creates fewer than the maximum number of tasks if it cannot achieve the parallelism.

PaulRMellor: BTW, we're going to add some concepts on Kafka Connect for tasks, workers etc.

tombentley: @PaulRMellor how about this slightly tweaked version:

The maximum number of tasks that the connector may create. Tasks handle the data replication and run in parallel. If the infrastructure supports the processing overhead, increasing this value can improve throughput. Kafka Connect distributes the tasks between members of the cluster. If there are more tasks than workers, workers are assigned multiple tasks. For sink connectors, aim to have one task for each topic partition consumed. For source connectors, the number of tasks that can run in parallel may also depend on the external system. The connector creates fewer than the maximum number of tasks if it cannot achieve the parallelism.

tombentley (approved):

PR #4996, opened 2021-05-20
DRAFT: Exit `KafkaRoller` schedule loop on detection of cancelled reconciliation

### LINKED ISSUE #4869: KafkaRoller may continue to try to roll a kafka's pod long after the kafka is deleted
**Describe the bug**

Strimzi (latest)

If a kafka is deleted (kafka CR removed), it is possible for KafkaRoller to continue to try act on the kafka for many minutes, needlessly, continually failing at each retry.  This is wasteful of resources.

In our use-case case, Strimzi will be managing a large number of kafkas with the set of kafka mutating relatively quick. There is the real possibility that useful work is delayed.

In the example I highlight below, Strimzi is still needlessless processing the CR, 17 minutes after it was deleted.

```
2021-04-30 11:04:19 INFO  OperatorWatcher:40 - Reconciliation #20(watch) Kafka(foo-fzz4yx57aem1j0b/foo-fzz4yx57aem1j0b): Kafka foo-fzz4yx57aem1j0b in namespace foo-fzz4yx57aem1j0b was ADDED
..
2021-04-30 11:06:40 INFO  OperatorWatcher:40 - Reconciliation #114(watch) Kafka(foo-fzz4yx57aem1j0b/foo-fzz4yx57aem1j0b): Kafka foo-fzz4yx57aem1j0b in namespace foo-fzz4yx57aem1j0b was DELETED


2021-04-30 11:07:48 INFO  KafkaRoller:296 - Reconciliation #20(watch) Kafka(foo-fzz4yx57aem1j0b/foo-fzz4yx57aem1j0b): Could not roll pod 1 due to io.strimzi.operator.cluster.operator.resource.KafkaRoller$ForceableProblem: Error getting broker config, retrying after at least 250ms
...
2021-04-30 11:07:48 DEBUG KafkaRoller:272 - Reconciliation #20(watch) Kafka(foo-fzz4yx57aem1j0b/foo-fzz4yx57aem1j0b): Considering restart of pod 2 after delay of 0 MILLISECONDS
2021-04-30 11:08:18 INFO  KafkaRoller:296 - Reconciliation #20(watch) Kafka(foo-fzz4yx57aem1j0b/foo-fzz4yx57aem1j0b): Could not roll pod 2 due to io.strimzi.operator.cluster.operator.resource.KafkaRoller$ForceableProblem: Error getting broker config, retrying after at least 250ms
2021-04-30 11:08:18 DEBUG KafkaRoller:272 - Reconciliation #20(watch) Kafka(foo-fzz4yx57aem1j0b/foo-fzz4yx57aem1j0b): Considering restart of pod 0 after delay of 250 MILLISECONDS
2021-04-30 11:08:48 INFO  KafkaRoller:296 - Reconciliation #20(watch) Kafka(foo-fzz4yx57aem1j0b/foo-fzz4yx57aem1j0b): Could not roll pod 0 due to io.strimzi.operator.cluster.operator.resource.KafkaRoller$ForceableProblem: Error getting broker config, retrying after at least 500ms
2021-04-30 11:08:48 DEBUG KafkaRoller:272 - Reconciliation #20(watch) Kafka(foo-fzz4yx57aem1j0b/foo-fzz4yx57aem1j0b): Considering restart of pod 1 after delay of 250 MILLISECONDS
2021-04-30 11:09:18 INFO  KafkaRoller:296 - Reconciliation #20(watch) Kafka(foo-fzz4yx57aem1j0b/foo-fzz4yx57aem1j0b): Could not roll pod 1 due to io.strimzi.operator.cluster.operator.resource.KafkaRoller$ForceableProblem: Error getting broker config, retrying after at least 500ms
....
2021-04-30 11:23:53 INFO  KafkaRoller:289 - Reconciliation #20(watch) Kafka(foo-fzz4yx57aem1j0b/foo-fzz4yx57aem1j0b): Could not roll pod 2, giving up after 10 attempts. Total delay between attempts 127750ms
```

**To Reproduce**
https://github.com/k-wall/strzimi-del-problem/blob/main/create_kafkas.sh

1. Install Strimzi using quick start, configure for `STRIMZI_NAMESPACE` `*` following docs
2. `create_kafkas.sh 50` to create 50 kafka
3. `oc delete k -l kafka=true --all-namespaces`
3.  Watch logs

**Expected behavior**
Efficient handling of the kafka delete case.

 - Strimzi version: 0.22.1
 - Installation method: Yaml
 - Kubernetes cluster: 4.7.2
 - Infrastructure: AWS multi region.

**YAML files and logs**

Attached

**Additional context**

MikeEdgar:
### Type of change

- Bugfix
- Enhancement / new feature

### Description

Add boolean to `Reconciliation` to indicate reconcile action cancelled due to observation of DELETED action

Fixes #4869 

This is still a work-in-progress, but I would like to get feedback on whether the approach seems reasonable, etc. Note that the final version of this PR will not include changes to the `Dockerfile`, please ignore.

### Checklist

- [ ] Write tests
- [ ] Update documentation
- [ ] Update CHANGELOG.md

Signed-off-by: Michael Edgar <medgar@redhat.com>

### DIFF
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/resource/KafkaRoller.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/resource/KafkaRoller.java
@@ -273,7 +273,14 @@ public class KafkaRoller {
         RestartContext ctx = podToContext.computeIfAbsent(podId,
             k -> new RestartContext(backoffSupplier));
         singleExecutor.schedule(() -> {
+            if (reconciliation.isCancelled()) {
+                log.info("{}: reconciliation cancelled, no longer considering restart of pod {}", reconciliation, podId);
+                ctx.promise.complete();
+                return;
+            }
+
             log.debug("{}: Considering restart of pod {} after delay of {} {}", reconciliation, podId, delay, unit);
+
             try {
                 restartIfNecessary(podId, ctx);
                 ctx.promise.complete();
@@ -295,6 +302,9 @@ public class KafkaRoller {
                     ctx.promise.fail(e instanceof TimeoutException ?
                             new io.strimzi.operator.common.operator.resource.TimeoutException() :
                             e);
+                } else if (reconciliation.isCancelled()) {
+                    log.info("{}: Could not roll pod {} due to {}, giving up due to cancelled reconciliation", reconciliation, podId, e);
+                    ctx.promise.complete();
                 } else {
                     long delay1 = ctx.backOff.delayMs();
                     log.info("{}: Could not roll pod {} due to {}, retrying after at least {}ms",
@@ -346,7 +356,7 @@ public class KafkaRoller {
             throws Exception {
         Pod pod;
         try {
-            pod = podOperations.get(namespace, KafkaCluster.kafkaPodName(cluster, podId));
+            pod = podOperations.get(namespace, podName(podId));
         } catch (KubernetesClientException e) {
             throw new UnforceableProblem("Error getting pod " + podName(podId), e);
         }
@@ -516,10 +526,13 @@ public class KafkaRoller {
      * @return a Future which completes with the config of the given broker.
      */
     protected Config brokerConfig(int brokerId) throws ForceableProblem, InterruptedException {
-        ConfigResource resource = new ConfigResource(ConfigResource.Type.BROKER, String.valueOf(brokerId));
+        ConfigResource resource = Util.getBrokersConfig(brokerId);
         return await(Util.kafkaFutureToVertxFuture(vertx, allClient.describeConfigs(singletonList(resource)).values().get(resource)),
             30, TimeUnit.SECONDS,
-            error -> new ForceableProblem("Error getting broker config", error)
+            error -> {
+                log.debug("Error getting broker config: {}", error.getMessage(), error);
+                return new ForceableProblem("Error getting broker config", error);
+            }
         );
     }
 
@@ -532,7 +545,10 @@ public class KafkaRoller {
         ConfigResource resource = Util.getBrokersLogging(brokerId);
         return await(Util.kafkaFutureToVertxFuture(vertx, allClient.describeConfigs(singletonList(resource)).values().get(resource)),
                 30, TimeUnit.SECONDS,
-            error -> new ForceableProblem("Error getting broker logging", error)
+            error -> {
+                log.debug("Error getting broker logging: {}", error.getMessage(), error);
+                return new ForceableProblem("Error getting broker logging", error);
+            }
         );
     }
 
@@ -719,7 +735,7 @@ public class KafkaRoller {
     String podName(int podId) {
         return KafkaCluster.kafkaPodName(this.cluster, podId);
     }
-    
+
     /**
      * Return true if the given {@code podId} is the controller and there are other brokers we might yet have to consider.
      * This ensures that the controller is restarted/reconfigured last.
--- a/docker-images/base/Dockerfile
+++ b/docker-images/base/Dockerfile
@@ -5,7 +5,7 @@ ARG TARGETPLATFORM
 USER root
 
 RUN microdnf update \
-    && microdnf install java-${JAVA_VERSION}-openjdk-headless openssl procps shadow-utils \
+    && microdnf install java-${JAVA_VERSION}-openjdk-headless java-${JAVA_VERSION}-openjdk-devel openssl procps shadow-utils \
     && microdnf clean all
 
 ENV JAVA_HOME /usr/lib/jvm/jre-11
--- a/operator-common/src/main/java/io/strimzi/operator/common/AbstractOperator.java
+++ b/operator-common/src/main/java/io/strimzi/operator/common/AbstractOperator.java
@@ -8,6 +8,7 @@ import io.fabric8.kubernetes.api.model.LabelSelector;
 import io.fabric8.kubernetes.api.model.rbac.ClusterRoleBinding;
 import io.fabric8.kubernetes.client.CustomResource;
 import io.fabric8.kubernetes.client.Watch;
+import io.fabric8.kubernetes.client.Watcher.Action;
 import io.fabric8.kubernetes.client.WatcherException;
 import io.micrometer.core.instrument.Counter;
 import io.micrometer.core.instrument.Tag;
@@ -38,6 +39,7 @@ import org.apache.logging.log4j.Logger;
 
 import java.util.ArrayList;
 import java.util.LinkedHashSet;
+import java.util.List;
 import java.util.Map;
 import java.util.Collections;
 import java.util.Optional;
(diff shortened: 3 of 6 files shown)


cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/resource/KafkaRoller.java
```diff
@@ -273,7 +273,14 @@ public String toString() {
         RestartContext ctx = podToContext.computeIfAbsent(podId,
             k -> new RestartContext(backoffSupplier));
         singleExecutor.schedule(() -> {
+            if (reconciliation.isCancelled()) {
```
tombentley, line 276: Using the `Reconciliation` for checking for cancellation seems reasonable. But I think we should encapsulate the checking within `Reconcilation` with a `checkForCancellation` method, which throws a `ReconciliationCancelledException`. Within `KafkaRoller` we should be able to make that exit the rolling immediately, most likely propagating the/a `ReconciliationCancelledException` to the `KafkaAssemblyOperator` itself.

docker-images/base/Dockerfile
```diff
@@ -5,7 +5,7 @@ ARG TARGETPLATFORM
 USER root
 
 RUN microdnf update \
-    && microdnf install java-${JAVA_VERSION}-openjdk-headless openssl procps shadow-utils \
+    && microdnf install java-${JAVA_VERSION}-openjdk-headless java-${JAVA_VERSION}-openjdk-devel openssl procps shadow-utils \
```
tombentley, line 8: Why do we need this?

operator-common/src/main/java/io/strimzi/operator/common/AbstractOperator.java
```diff
@@ -139,6 +142,7 @@ public AbstractOperator(Vertx vertx, String kind, O resourceOperator, MetricsPro
                 metricTags);
 
         resourcesStateCounter = new ConcurrentHashMap<>();
+        reconciliations = new ConcurrentHashMap<>();
```
tombentley, line 145: It seems to me that the map (internal to vertx) used for the locks is exactly what we need to provide the necessary semantics, and having two maps, as here, opens the door to a race (though I think it's unlikely to be a problem in practice). I'm not sure how easy it would be to implement our own replacement for the locking functionality that supported cancellation.

tombentley (commented):
Hi @MikeEdgar thanks for taking a look at this and sorry for the delay. I think this is along the right lines, but I think we should explore whether we can integrate this with the locking.

docker-images/base/Dockerfile
MikeEdgar: This is only to support collection of JVM thread dumps in my testing. The final PR will not have any changes to `Dockerfile`.

operator-common/src/main/java/io/strimzi/operator/common/AbstractOperator.java
MikeEdgar: Are you thinking something like `vertx.sharedData().getLocalMap("reconciliations")` ? I may not be following you completely.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/resource/KafkaRoller.java
MikeEdgar: Just to clarify - `checkForCancellation` would be a drop-in replacement for the current `isCancelled` calls, correct? Of course, with the logging/promise logic removed. I'm also assuming that `ReconciliationCancelledException` would be best as a `RuntimeException`.

operator-common/src/main/java/io/strimzi/operator/common/AbstractOperator.java
tombentley: The first time I looked at this PR I was wondering whether we really needed a map of _all_ the reconciliations, or whether we were just interested in the cancelled ones. I think the original comment was motivated by this. Now I look again, I don't see the original problem. 

That the cancellation doesn't necessarily affect just the running reconciliation, but also any that are queued up waiting on the lock makes me wonder whether anything guarantees that at the time we process a DELETED event (which will cause the cancellation) that the queue cannot have a recreation ADDED event. This would be guaranteed by the assumption that fabric8 dispatches all watch events via a single thread. But off the top of my head I'm not certain whether that assumption is true. In any case I think this would be super-rare in practice and would be fixed by the next periodic reconciliation.

tombentley:
The ask, according to #4869, is for deletion to be swiftly honoured.

In its current form this PR relies on checking for cancellation prior to each restart attempt (and thus after a certain time delay).
While that might be an improvement it's not every effective if cancellation happens just 
after we've scheduled a rolling attempt minutes into the future, because it will be minutes before the restart task runs and notices that the reconciliation is cancelled.
We need support for proper cooperative cancellation using thread interruption and/or APIs that support it. 

I can see how we can make interruption work for the KafkaRoller, by using `shutdownNow()` on the executor service. 
But that's not the only place where interruptible blocking is being used.
For example there is a call to `KafkaFuture.get` in the KAO which could block for a long time (though it should eventually timeout) and
ZookeeperScaler has a very dodgy-looking `catch (InterruptedException e)`.

So I don't think this PR in its current for will yield what #4869 is asking for.
It would be an improvement on the 17min wait time shown there, but minutes long waits would still be possible. 
It would not remove the need for machinery in the orchestrating code to account for "slow deletion". 

We can refactor these calls to interruptibly blocking APIs to support thread interruption if those methods are being called on "our own" threads where we control/can define an interruption policy.  
But currently they're being called on Vertx threads and so we need to pass off the job of calling those methods to our own threads. That means either we fire up a thread just for making those calls (not great for performance) or we have our own pool where we can carefully control when a thread runs a task, so we know which thread to interrupt in order to cancel a reconciliation.

Doing this is not very compatible with vertx. I don't think we should try to make it work with vertx. If we did we'd basically have two ways of executing code, on vertx and on some threads of our own. We'd have a _lot_ of complexity and it will be really hard to reason correctly about threading. It would be better to get rid of vertx and use a thread-per-reconciliation model. That would have other benefits (e.g. meaingful stacktraces), and would be a better fit for the sort of scale that the operator runs at.

MikeEdgar:
@tombentley - thanks for following up and sorry to leave this open for so long. Based on your comment, it sounds as though the correct approach will be something more comprehensive. Do you see anything worth salvaging here?

tombentley:
> @tombentley - thanks for following up and sorry to leave this open for so long.

It's me who should be apologising to you for being so slow in reviewing it.

>  Based on your comment, it sounds as though the correct approach will be something more comprehensive. Do you see anything worth salvaging here?

Possibly not, but don't close it just yet. I've opened #5438 to illustrate the sort of changes I'm talking about. I'd like it to see some discussion before we decide what to do about this.

scholzj:
Thanks for the PR. This was discussed many times not he community calls. The conclusion always was that this is not the implementation we want to accept - the issues is relatively minor and the risk connected to this and the added complexity is too big. We will try to come up with some better solution in the future.

PR #5008, opened 2021-05-21
Various small updates to test log statements

michalxo:
Altered few log level statements to more useful (imo) logging levels.
Also, some of the log statements had bad wording and some non-functional values (printed empty stuff).

### DIFF
--- a/systemtest/src/main/java/io/strimzi/systemtest/utils/kafkaUtils/KafkaUtils.java
+++ b/systemtest/src/main/java/io/strimzi/systemtest/utils/kafkaUtils/KafkaUtils.java
@@ -199,20 +199,22 @@ public class KafkaUtils {
             boolean kafkaSameAsLast = kafkaSnaptop.equals(kafkaPods[0]);
             boolean eoSameAsLast = eoSnapshot.equals(eoPods[0]);
             if (!zkSameAsLast) {
-                LOGGER.info("ZK Cluster not stable");
+                LOGGER.warn("ZK Cluster not stable");
             }
             if (!kafkaSameAsLast) {
-                LOGGER.info("Kafka Cluster not stable");
+                LOGGER.warn("Kafka Cluster not stable");
             }
             if (!eoSameAsLast) {
-                LOGGER.info("EO not stable");
+                LOGGER.warn("EO not stable");
             }
-            if (zkSameAsLast
-                    && kafkaSameAsLast
-                    && eoSameAsLast) {
+            if (zkSameAsLast && kafkaSameAsLast && eoSameAsLast) {
                 int c = count[0]++;
-                LOGGER.info("All stable for {} polls", c);
-                return c > 60;
+                LOGGER.debug("All stable after {} polls", c);
+                if (c > 60) {
+                    LOGGER.info("Kafka cluster is stable after {} polls.", c);
+                    return true;
+                }
+                return false;
             }
             zkPods[0] = zkSnapshot;
             kafkaPods[0] = kafkaSnaptop;
@@ -259,7 +261,7 @@ public class KafkaUtils {
      * @param value value of specific property
      */
     public static boolean verifyCrDynamicConfiguration(String clusterName, String brokerConfigName, Object value) {
-        LOGGER.info("Dynamic Configuration in Kafka CR is {}={} and excepted is {}={}",
+        LOGGER.info("Dynamic Configuration in Kafka CR is {}={} and expected is {}={}",
             brokerConfigName,
             KafkaResource.kafkaClient().inNamespace(kubeClient().getNamespace()).withName(clusterName).get().getSpec().getKafka().getConfig().get(brokerConfigName),
             brokerConfigName,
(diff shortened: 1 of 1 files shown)


scholzj (commented):
Might be worth explaining why do you think these changes should be done instead of the empty description ;-).

michalxo:
> Might be worth explaining why do you think these changes should be done instead of the empty description ;-).

I thought it's self-explanatory. :-) 
Basically too much debug data was printed in test (lowered to debug from info level).
Some of the log statements had bad wording and some non-functional values (printed empty stuff).
These are really small changes, if not needed then this PR can be happily rejected :)

systemtest/src/main/java/io/strimzi/systemtest/utils/kafkaUtils/KafkaUtils.java
```diff
@@ -199,19 +199,19 @@ public static void waitForClusterStability(String namespaceName, String clusterN
...
             if (!eoSameAsLast) {
-                LOGGER.info("EO not stable");
+                LOGGER.warn("EO not stable");
             }
             if (zkSameAsLast
                     && kafkaSameAsLast
                     && eoSameAsLast) {
                 int c = count[0]++;
-                LOGGER.info("All stable for {} polls", c);
+                LOGGER.debug("All stable after {} polls", c);
```
Frawless, line 214: I think you should add some final info log that pods are stable for X seconds or something like that. It seems it's not in this method.

systemtest/src/main/java/io/strimzi/systemtest/utils/kafkaUtils/KafkaUtils.java
```diff
@@ -330,9 +330,9 @@ public static boolean verifyPodDynamicConfiguration(String kafkaPodNamePrefix, S
 
         Map<String, ConfigModel> configs = KafkaUtils.readConfigModel(kafkaVersion);
 
-        LOGGER.info("This is configs {}", configs.toString());
+        LOGGER.debug("Kafka config {}", configs.toString());
```
Frawless, line 333: @see-quick is this log on INFO level worthy in case of some debug?

scholzj:
> I thought it's self-explanatory. :-)
Basically too much debug data was printed in test (lowered to debug from info level).
Some of the log statements had bad wording and some non-functional values (printed empty stuff).
These are really small changes, if not needed then this PR can be happily rejected :)

Nothing is self-explanatory few years later.  I think that every PR deserves some description even if it is just that your personal view is that it should be logged like this. Also, you change some from info -> debug, but some from info -> warn ... always worth few words to make it clear what and why is this PR suggesting.

systemtest/src/main/java/io/strimzi/systemtest/utils/kafkaUtils/KafkaUtils.java
im-konge, line 333: I guess that it's good to know which dynamic configuration will be tested. The sentence is good, just change it back to the `info` level.

im-konge (commented):
I have same comments as @Frawless and @scholzj - you should definitely add some description.

systemtest/src/main/java/io/strimzi/systemtest/utils/kafkaUtils/KafkaUtils.java
michalxo: reverting back to info.

systemtest/src/main/java/io/strimzi/systemtest/utils/kafkaUtils/KafkaUtils.java
michalxo: Provided.

systemtest/src/main/java/io/strimzi/systemtest/utils/kafkaUtils/KafkaUtils.java
```diff
@@ -199,20 +199,25 @@ public static void waitForClusterStability(String namespaceName, String clusterN
...
                 int c = count[0]++;
-                LOGGER.info("All stable for {} polls", c);
-                return c > 60;
+                LOGGER.debug("All stable after {} polls", c);
+                if (c > 60) {
+                    LOGGER.info("Kafka cluster is stable after {} polls.", c);
+                    return true;
+                } else {
+                    return false;
+                }
```
im-konge, line 220: You can simplify it like this:
```suggestion
                if (c > 60) {
                    LOGGER.info("Kafka cluster is stable after {} polls.", c);
                    return true;
                }
                return false;
```

Frawless (approved):
Please, fill up the PR description.

im-konge (approved):

systemtest/src/main/java/io/strimzi/systemtest/utils/kafkaUtils/KafkaUtils.java
michalxo: fixed. thanks

scholzj:
> Too much debug data was printed in test (lowered to debug from info level).

So, we ended up with:
* 3 log levels increased to warn
* One new info level added
* And one changed to debug

... just saying ;-)

michalxo:
> > Too much debug data was printed in test (lowered to debug from info level).
> 
> So, we ended up with:
> 
>     * 3 log levels increased to warn
> 
>     * One new info level added
> 
>     * And one changed to debug
> 
> 
> ... just saying ;-)

Yeah, you were right... :-D

scholzj (approved):

PR #5018, opened 2021-05-24
Pass proxy configuration into the Connect build as arguments

### LINKED ISSUE #4814: Make Kafka Connect Build work behind proxies
As discussed in #4808, Kafka Connect build currently does not seem to work behind HTTP proxies. We should make it possible.

scholzj:
### Type of change

- Enhancement / new feature

### Description

The operator accepts the proxy configuration in the form of environment variables (`HTTP_PROXY`, `HTTPS_PROXY` and `NO_PROXY`) and passes it over to the operands. However, we do not currently pass them into the Kafka Connect Build. So when user runs it behind a proxy and the build tries to download the connector plugins, it fails. This PR passes the proxy configuration into the container as build arguments (so that they do not propagate into the container when it is later used). Because `curl` doesn't like uppercase `HTTP_PROXY`, they are passed in lowercase.

This should close #4814.

### Checklist


scholzj:
@strimzi/maintainers Any chance someone can review this?

tombentley (approved):

PR #4888, opened 2021-05-03
Add and enable Control Plane listener

scholzj:
### Type of change

- Enhancement / new feature

### Description

This PR adds support for the Control Plane listener as described in [proposal 25](https://github.com/strimzi/proposals/blob/main/025-control-plain-listener.md).

It adds a new listener called `CONTROLPLANE-9090` listening on port 9090. This listener is added by default. It also adds a new Feature Gate called `ControlPlaneListener` which is disabled by default. It can be enabled by adding `+ControlPlaneListener` to the `STRIMZI_FEATURE_GATES` environment variable in the Cluster Operator. When enabled, it activates the control plane listener (tells Kafka to use it for control plane communication).

System Tests and docs will be done in separate PR.

### Checklist

- [x] Write tests
- [x] Make sure all tests pass
- [x] Try your changes from Pod inside your Kubernetes and OpenShift cluster, not just locally
- [x] Update CHANGELOG.md

PR #5033, opened 2021-05-26
New SystemTest for FeatureGate: ControlPlaneListener

michalxo:
Signed-off-by: Michal Tóth <mtoth@redhat.com>

### Type of change

- New test for FeatureGatesST.testControlPlaneListenerFeatureGate
- Addition of STRIMZI_FEATURE_GATES Environment variable to OLM, Helm and Bundle installation types
- Passing additionalEnvVars to ClusterOperator creation method(s)
- Small changes in OLMResource (problems with instance variables)

### Description

This is a SystemTest for first Feature Gate - ControlPlainListener
Work is related to Add and enable Control Plane listener #4888
### Checklist

- [x] Write tests

### DIFF
--- a/systemtest/src/main/java/io/strimzi/systemtest/Environment.java
+++ b/systemtest/src/main/java/io/strimzi/systemtest/Environment.java
@@ -125,6 +125,11 @@ public class Environment {
     private static final String LB_FINALIZERS_ENV = "LB_FINALIZERS";
 
     /**
+     * CO Features gates variable
+     */
+    public static final String STRIMZI_FEATURE_GATES_ENV = "STRIMZI_FEATURE_GATES";
+
+    /**
      * Defaults
      */
     private static final String ST_KAFKA_VERSION_DEFAULT = "2.8.0";
@@ -146,6 +151,7 @@ public class Environment {
     private static final boolean DEFAULT_TO_DENY_NETWORK_POLICIES_DEFAULT = true;
     private static final ClusterOperatorInstallType CLUSTER_OPERATOR_INSTALL_TYPE_DEFAULT = ClusterOperatorInstallType.BUNDLE;
     private static final boolean LB_FINALIZERS_DEFAULT = false;
+    private static final String STRIMZI_FEATURE_GATES_DEFAULT = "";
 
     /**
      * Set values
@@ -162,6 +168,7 @@ public class Environment {
     public static final String KUBERNETES_DOMAIN = getOrDefault(KUBERNETES_DOMAIN_ENV, KUBERNETES_DOMAIN_DEFAULT);
     public static final boolean SKIP_TEARDOWN = getOrDefault(SKIP_TEARDOWN_ENV, Boolean::parseBoolean, false);
     public static final String STRIMZI_RBAC_SCOPE = getOrDefault(STRIMZI_RBAC_SCOPE_ENV, STRIMZI_RBAC_SCOPE_DEFAULT);
+    public static final String STRIMZI_FEATURE_GATES = getOrDefault(STRIMZI_FEATURE_GATES_ENV, STRIMZI_FEATURE_GATES_DEFAULT);
     // variables for test-client image
     private static final String TEST_CLIENT_IMAGE_DEFAULT = STRIMZI_REGISTRY + "/" + STRIMZI_ORG + "/test-client:" + STRIMZI_TAG + "-kafka-" + ST_KAFKA_VERSION;
     public static final String TEST_CLIENT_IMAGE = getOrDefault(TEST_CLIENT_IMAGE_ENV, TEST_CLIENT_IMAGE_DEFAULT);
--- a/systemtest/src/main/java/io/strimzi/systemtest/resources/operator/BundleResource.java
+++ b/systemtest/src/main/java/io/strimzi/systemtest/resources/operator/BundleResource.java
@@ -56,38 +56,42 @@ public class BundleResource implements ResourceType<Deployment> {
     }
 
     public static DeploymentBuilder clusterOperator(String namespace, long operationTimeout) {
-        return defaultClusterOperator(Constants.STRIMZI_DEPLOYMENT_NAME, namespace, namespace, operationTimeout, Constants.RECONCILIATION_INTERVAL);
+        return defaultClusterOperator(Constants.STRIMZI_DEPLOYMENT_NAME, namespace, namespace, operationTimeout, Constants.RECONCILIATION_INTERVAL, null);
     }
 
     public static DeploymentBuilder clusterOperator(String namespace, String namespaceEnv, long operationTimeout, long reconciliationInterval) {
-        return defaultClusterOperator(Constants.STRIMZI_DEPLOYMENT_NAME, namespace, namespaceEnv, operationTimeout, reconciliationInterval);
+        return defaultClusterOperator(Constants.STRIMZI_DEPLOYMENT_NAME, namespace, namespaceEnv, operationTimeout, reconciliationInterval, null);
     }
 
     public static DeploymentBuilder clusterOperator(String namespace, String namespaceEnv, long reconciliationInterval) {
-        return defaultClusterOperator(Constants.STRIMZI_DEPLOYMENT_NAME, namespace, namespaceEnv, Constants.CO_OPERATION_TIMEOUT_DEFAULT, reconciliationInterval);
+        return defaultClusterOperator(Constants.STRIMZI_DEPLOYMENT_NAME, namespace, namespaceEnv, Constants.CO_OPERATION_TIMEOUT_DEFAULT, reconciliationInterval, null);
     }
 
     public static DeploymentBuilder clusterOperator(String namespace, String namespaceEnv) {
-        return defaultClusterOperator(Constants.STRIMZI_DEPLOYMENT_NAME, namespace, namespaceEnv, Constants.CO_OPERATION_TIMEOUT_DEFAULT, Constants.RECONCILIATION_INTERVAL);
+        return defaultClusterOperator(Constants.STRIMZI_DEPLOYMENT_NAME, namespace, namespaceEnv, Constants.CO_OPERATION_TIMEOUT_DEFAULT, Constants.RECONCILIATION_INTERVAL, null);
     }
 
     public static DeploymentBuilder clusterOperator(String namespace, long operationTimeout, long reconciliationInterval) {
-        return defaultClusterOperator(Constants.STRIMZI_DEPLOYMENT_NAME, namespace, namespace, operationTimeout, reconciliationInterval);
+        return defaultClusterOperator(Constants.STRIMZI_DEPLOYMENT_NAME, namespace, namespace, operationTimeout, reconciliationInterval, null);
     }
 
     public static DeploymentBuilder clusterOperator(String name, String namespace, String namespaceEnv, long operationTimeout, long reconciliationInterval) {
-        return defaultClusterOperator(name, namespace, namespaceEnv, operationTimeout, reconciliationInterval);
+        return defaultClusterOperator(name, namespace, namespaceEnv, operationTimeout, reconciliationInterval, null);
     }
 
     public static DeploymentBuilder clusterOperator(String namespace) {
-        return defaultClusterOperator(Constants.STRIMZI_DEPLOYMENT_NAME, namespace, namespace, Constants.CO_OPERATION_TIMEOUT_DEFAULT, Constants.RECONCILIATION_INTERVAL);
+        return defaultClusterOperator(Constants.STRIMZI_DEPLOYMENT_NAME, namespace, namespace, Constants.CO_OPERATION_TIMEOUT_DEFAULT, Constants.RECONCILIATION_INTERVAL, null);
+    }
+
+    public static DeploymentBuilder clusterOperator(String namespace, List<EnvVar> extraEnvVars) {
+        return defaultClusterOperator(Constants.STRIMZI_DEPLOYMENT_NAME, namespace, namespace, Constants.CO_OPERATION_TIMEOUT_DEFAULT, Constants.RECONCILIATION_INTERVAL, extraEnvVars);
     }
 
     public static DeploymentBuilder defaultClusterOperator(String namespace) {
-        return defaultClusterOperator(Constants.STRIMZI_DEPLOYMENT_NAME, namespace, namespace, Constants.CO_OPERATION_TIMEOUT_DEFAULT, Constants.RECONCILIATION_INTERVAL);
+        return defaultClusterOperator(Constants.STRIMZI_DEPLOYMENT_NAME, namespace, namespace, Constants.CO_OPERATION_TIMEOUT_DEFAULT, Constants.RECONCILIATION_INTERVAL, null);
     }
 
-    private static DeploymentBuilder defaultClusterOperator(String name, String namespace, String namespaceEnv, long operationTimeout, long reconciliationInterval) {
+    private static DeploymentBuilder defaultClusterOperator(String name, String namespace, String namespaceEnv, long operationTimeout, long reconciliationInterval, List<EnvVar> additionalEnvVars) {
 
         Deployment clusterOperator = DeploymentResource.getDeploymentFromYaml(PATH_TO_CO_CONFIG);
 
@@ -109,6 +113,8 @@ public class BundleResource implements ResourceType<Deployment> {
                 case "STRIMZI_OPERATION_TIMEOUT_MS":
                     envVar.setValue(Long.toString(operationTimeout));
                     break;
+                case "STRIMZI_FEATURE_GATES":
+                    envVar.setValue(Environment.STRIMZI_FEATURE_GATES);
                 default:
                     if (envVar.getName().contains("KAFKA_BRIDGE_IMAGE")) {
                         envVar.setValue(Environment.useLatestReleasedBridge() ? envVar.getValue() : Environment.BRIDGE_IMAGE);
@@ -123,6 +129,15 @@ public class BundleResource implements ResourceType<Deployment> {
         envVars.add(new EnvVar("STRIMZI_IMAGE_PULL_POLICY", Environment.COMPONENTS_IMAGE_PULL_POLICY, null));
         envVars.add(new EnvVar("STRIMZI_LOG_LEVEL", Environment.STRIMZI_LOG_LEVEL, null));
         envVars.add(new EnvVar("STRIMZI_RBAC_SCOPE", Environment.STRIMZI_RBAC_SCOPE, null));
+
+        if (additionalEnvVars != null) {
+            envVars.forEach(envVar -> additionalEnvVars.stream()
+                    .filter(aVar -> envVar.getName().equals(aVar.getName()))
+                    .findFirst()
+                    .ifPresent(xVar -> envVar.setValue(xVar.getValue()))
+            );
+        }
+
         // Apply updated env variables
         clusterOperator.getSpec().getTemplate().getSpec().getContainers().get(0).setEnv(envVars);
 
--- a/systemtest/src/main/java/io/strimzi/systemtest/resources/specific/HelmResource.java
+++ b/systemtest/src/main/java/io/strimzi/systemtest/resources/specific/HelmResource.java
@@ -115,7 +115,8 @@ public class HelmResource implements SpecificResourceType {
                 entry("resources.limits.cpu", LIMITS_CPU),
                 entry("logLevelOverride", Environment.STRIMZI_LOG_LEVEL),
                 entry("fullReconciliationIntervalMs", Long.toString(reconciliationInterval)),
-                entry("operationTimeoutMs", Long.toString(operationTimeout)))
+                entry("operationTimeoutMs", Long.toString(operationTimeout)),
+                entry("featureGates", Environment.STRIMZI_FEATURE_GATES))
                 .collect(TestUtils.entriesToMap()));
 
         Path pathToChart = new File(HELM_CHART).toPath();
--- a/systemtest/src/main/java/io/strimzi/systemtest/resources/specific/OlmResource.java
+++ b/systemtest/src/main/java/io/strimzi/systemtest/resources/specific/OlmResource.java
@@ -268,7 +268,8 @@ public class OlmResource implements SpecificResourceType {
                     .replace("${OLM_INSTALL_PLAN_APPROVAL}", installationStrategy.toString())
                     .replace("${STRIMZI_FULL_RECONCILIATION_INTERVAL_MS}", Long.toString(reconciliationInterval))
                     .replace("${STRIMZI_OPERATION_TIMEOUT_MS}", Long.toString(operationTimeout))
-                    .replace("${STRIMZI_RBAC_SCOPE}", Environment.STRIMZI_RBAC_SCOPE));
+                    .replace("${STRIMZI_RBAC_SCOPE}", Environment.STRIMZI_RBAC_SCOPE)
+                    .replace("${STRIMZI_FEATURE_GATES}", Environment.STRIMZI_FEATURE_GATES));
 
 
             ResourceManager.cmdKubeClient().apply(subscriptionFile);
--- a/systemtest/src/main/java/io/strimzi/systemtest/resources/specific/OlmResource.java
+++ b/systemtest/src/main/java/io/strimzi/systemtest/resources/specific/OlmResource.java
@@ -67,7 +67,7 @@ public class OlmResource implements SpecificResourceType {
 
     @Override
     public void delete() {
-        this.deleteOlm(deploymentName, namespace, csvName);
+        deleteOlm(deploymentName, namespace, csvName);
     }
 
     public OlmResource() { }
@@ -97,8 +97,6 @@ public class OlmResource implements SpecificResourceType {
             createOperatorGroup(namespace);
         }
 
-        String csvName;
-
         if (fromVersion != null) {
             createAndModifySubscription(namespace, operationTimeout, reconciliationInterval, olmInstallationStrategy, fromVersion);
             // must be strimzi-cluster-operator.v0.18.0
@@ -119,7 +117,7 @@ public class OlmResource implements SpecificResourceType {
         TestUtils.waitFor("Cluster Operator deployment creation", Constants.GLOBAL_POLL_INTERVAL, CR_CREATION_TIMEOUT,
             () -> ResourceManager.kubeClient().getDeploymentNameByPrefix(Environment.OLM_OPERATOR_DEPLOYMENT_NAME) != null);
 
-        String deploymentName = ResourceManager.kubeClient().getDeploymentNameByPrefix(Environment.OLM_OPERATOR_DEPLOYMENT_NAME);
+        deploymentName = ResourceManager.kubeClient().getDeploymentNameByPrefix(Environment.OLM_OPERATOR_DEPLOYMENT_NAME);
         ResourceManager.setCoDeploymentName(deploymentName);
 
         // Wait for operator creation
--- a/systemtest/src/main/java/io/strimzi/systemtest/utils/kubeUtils/objects/ServiceUtils.java
+++ b/systemtest/src/main/java/io/strimzi/systemtest/utils/kubeUtils/objects/ServiceUtils.java
@@ -20,7 +20,6 @@ import static io.strimzi.test.k8s.KubeClusterResource.kubeClient;
 public class ServiceUtils {
 
     private static final Logger LOGGER = LogManager.getLogger(ServiceUtils.class);
-    private static final long READINESS_TIMEOUT = ResourceOperation.getTimeoutForResourceReadiness(Constants.SERVICE);
     private static final long DELETION_TIMEOUT = ResourceOperation.getTimeoutForResourceDeletion();
 
     private ServiceUtils() { }
@@ -40,10 +39,6 @@ public class ServiceUtils {
         }
     }
 
-    public static void waitForServiceLabelsChange(String serviceName, Map<String, String> labels) {
-        waitForServiceLabelsChange(kubeClient().getNamespace(), serviceName, labels);
-    }
-
     public static void waitForServiceLabelsDeletion(String namespaceName, String serviceName, String... labelKeys) {
         for (final String labelKey : labelKeys) {
             LOGGER.info("Service label {} change to {}", labelKey, null);
@@ -52,36 +47,6 @@ public class ServiceUtils {
                     kubeClient(namespaceName).getService(namespaceName, serviceName).getMetadata().getLabels().get(labelKey) == null
             );
         }
-    }
-
-    public static void waitForLoadBalancerService(String serviceName) {
-        LOGGER.info("Waiting for Service {} in namespace {}", serviceName, kubeClient().getNamespace());
-
-        TestUtils.waitFor("LoadBalancer service " + serviceName + " to be ready", Constants.POLL_INTERVAL_FOR_RESOURCE_READINESS, READINESS_TIMEOUT,
-            () -> kubeClient().getClient().services().inNamespace(kubeClient().getNamespace()).withName(serviceName).get().getSpec().getExternalIPs().size() > 0);
-        LOGGER.info("Service {} in namespace {} is ready", serviceName, kubeClient().getNamespace());
-    }
-
-    public static void waitForNodePortService(String serviceName) throws InterruptedException {
-        LOGGER.info("Waiting for Service {} in namespace {}", serviceName, kubeClient().getNamespace());
-
-        TestUtils.waitFor("NodePort service " + serviceName + " to be ready", Constants.POLL_INTERVAL_FOR_RESOURCE_READINESS, READINESS_TIMEOUT,
-            () -> kubeClient().getClient().services().inNamespace(kubeClient().getNamespace()).withName(serviceName).get().getSpec().getPorts().get(0).getNodePort() != null);
-
-        Thread.sleep(10000);
-        LOGGER.info("Service {} in namespace {} is ready", serviceName, kubeClient().getNamespace());
-    }
-
-    /**
-     * Wait until Service of the given name will be deleted.
-     * @param serviceName service name
-     */
-    public static void waitForServiceDeletion(String serviceName) {
-        LOGGER.info("Waiting for Service {} deletion in namespace {}", serviceName, kubeClient().getNamespace());
-
-        TestUtils.waitFor("Service " + serviceName + " to be deleted", Constants.POLL_INTERVAL_FOR_RESOURCE_READINESS, DELETION_TIMEOUT,
-            () -> kubeClient().getService(serviceName) == null);
-        LOGGER.info("Service {} in namespace {} was deleted", serviceName, kubeClient().getNamespace());
     }
 
     /**
--- a/systemtest/src/test/java/io/strimzi/systemtest/operators/FeatureGatesST.java
+++ b/systemtest/src/test/java/io/strimzi/systemtest/operators/FeatureGatesST.java
@@ -54,7 +54,6 @@ public class FeatureGatesST extends AbstractST {
     private static final Logger LOGGER = LogManager.getLogger(FeatureGatesST.class);
 
     protected void deployClusterOperatorWithEnvVars(ExtensionContext extensionContext, List<EnvVar> envVars) {
-        assumeFalse(Environment.isOlmInstall() || Environment.isHelmInstall());
         prepareEnvForOperator(extensionContext, NAMESPACE, Arrays.asList(NAMESPACE, NAMESPACE));
         applyBindings(extensionContext, NAMESPACE);
         resourceManager.createResource(extensionContext, BundleResource.clusterOperator(NAMESPACE, envVars).build());
@@ -66,7 +65,9 @@ public class FeatureGatesST extends AbstractST {
      */
     @IsolatedTest("Feature Gates test for ControlPlainListener")
     @Tag(INTERNAL_CLIENTS_USED)
-    public void testControlPlaneListenerFeatureGate(ExtensionContext extensionContext) throws InterruptedException {
+    public void testControlPlaneListenerFeatureGate(ExtensionContext extensionContext) {
+        assumeFalse(Environment.isOlmInstall() || Environment.isHelmInstall());
+
         final String clusterName = mapWithClusterNames.get(extensionContext.getDisplayName());
         final String producerName = "producer-test-" + new Random().nextInt(Integer.MAX_VALUE);
         final String consumerName = "consumer-test-" + new Random().nextInt(Integer.MAX_VALUE);
@@ -75,10 +76,7 @@ public class FeatureGatesST extends AbstractST {
         List<EnvVar> testEnvVars = new ArrayList<>();
         int kafkaReplicas = 3;
 
-        String systemEnvFG = System.getenv(Environment.STRIMZI_FEATURE_GATES_ENV);
-        if (systemEnvFG == null || !systemEnvFG.contains("ControlPlaneListener")) {
-            testEnvVars.add(new EnvVar(Environment.STRIMZI_FEATURE_GATES_ENV, "+ControlPlaneListener", null));
-        }
+        testEnvVars.add(new EnvVar(Environment.STRIMZI_FEATURE_GATES_ENV, "+ControlPlaneListener", null));
         deployClusterOperatorWithEnvVars(extensionContext, testEnvVars);
         resourceManager.createResource(extensionContext, KafkaTemplates.kafkaPersistent(clusterName, kafkaReplicas).build());
 
@@ -90,7 +88,7 @@ public class FeatureGatesST extends AbstractST {
 
         Map<String, String> kafkaPods = StatefulSetUtils.ssSnapshot(NAMESPACE, kafkaStatefulSetName(clusterName));
 
-        LOGGER.info("Try to send some messages to Kafka over the next {} seconds.", messageCount);
+        LOGGER.info("Try to send some messages to Kafka over next few minutes.");
         KafkaTopic kafkaTopic = KafkaTopicTemplates.topic(clusterName, topicName)
                 .editSpec()
                     .withReplicas(kafkaReplicas)
@@ -131,7 +129,5 @@ public class FeatureGatesST extends AbstractST {
         LOGGER.info("Waiting for clients to finish sending/receiving messages.");
         ClientUtils.waitForClientSuccess(producerName, NAMESPACE, MESSAGE_COUNT);
         ClientUtils.waitForClientSuccess(consumerName, NAMESPACE, MESSAGE_COUNT);
-        JobUtils.deleteJobWithWait(NAMESPACE, producerName);
-        JobUtils.deleteJobWithWait(NAMESPACE, consumerName);
     }
 }
--- a/systemtest/src/test/java/io/strimzi/systemtest/operators/FeatureGatesST.java
+++ b/systemtest/src/test/java/io/strimzi/systemtest/operators/FeatureGatesST.java
@@ -1,0 +1,137 @@
+/*
+ * Copyright Strimzi authors.
+ * License: Apache License 2.0 (see the file LICENSE or http://apache.org/licenses/LICENSE-2.0.html).
+ */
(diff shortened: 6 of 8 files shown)


systemtest/src/main/java/io/strimzi/systemtest/Environment.java
```diff
@@ -162,6 +168,7 @@
     public static final String KUBERNETES_DOMAIN = getOrDefault(KUBERNETES_DOMAIN_ENV, KUBERNETES_DOMAIN_DEFAULT);
     public static final boolean SKIP_TEARDOWN = getOrDefault(SKIP_TEARDOWN_ENV, Boolean::parseBoolean, false);
     public static final String STRIMZI_RBAC_SCOPE = getOrDefault(STRIMZI_RBAC_SCOPE_ENV, STRIMZI_RBAC_SCOPE_DEFAULT);
+    public static final String STRIMZI_FEATURE_GATES = getOrDefault(STRIMZI_FEATURE_GATES_ENV, STRIMZI_FEATURE_GATES_DEFAULT);
```
michalxo, line 171: @Frawless do you think this should be removed, or better to keep it there for future? I don't use it in my tests so far...

systemtest/src/test/java/io/strimzi/systemtest/operators/FeatureGatesST.java
```diff
@@ -0,0 +1,93 @@
...
+    void afterEach(ExtensionContext extensionContext) {
+        clusterOperator = redeployCOWithEnvVars(defaultEnvVars);
+        testEnvVars = null;
+    }
+
+    @BeforeEach
+    void beforeEachSetClusterOperator(ExtensionContext extensionContext) {
+        clusterOperator = DeploymentResource.getDeploymentFromYaml(BundleResource.PATH_TO_CO_CONFIG);
+        testEnvVars = clusterOperator.getSpec().getTemplate().getSpec().getContainers().get(0).getEnv();
+    }
```
im-konge, line 63: I know it makes sense to have these `@BeforeAll`, `@BeforeEach` etc. on the beginning of the class, but right now we have it at the end of all ST classes, can you please move it there to follow this "pattern"?

systemtest/src/test/java/io/strimzi/systemtest/operators/FeatureGatesST.java
```diff
@@ -0,0 +1,93 @@
...
+    private Deployment clusterOperator;
+    private List<EnvVar> defaultEnvVars;
+    private List<EnvVar> testEnvVars;
+
+    @BeforeAll
+    void setup(ExtensionContext extensionContext) {
+        installClusterWideClusterOperator(extensionContext, NAMESPACE, Constants.CO_OPERATION_TIMEOUT_DEFAULT, Constants.RECONCILIATION_INTERVAL);
+        clusterOperator = DeploymentResource.getDeploymentFromYaml(BundleResource.PATH_TO_CO_CONFIG);
+        defaultEnvVars = clusterOperator.getSpec().getTemplate().getSpec().getContainers().get(0).getEnv();
+    }
```
im-konge, line 51: If you know that for each test you'll need CO with different env vars or configuration, I would deploy the CO in each testcase, like we have it in `NetworkPoliciesST`.

Then you can just do this:
```
resourceManager.createResource(extensionContext, BundleResource.clusterOperator(NAMESPACE, "*", Constants.CO_OPERATION_TIMEOUT_DEFAULT)
            .editOrNewSpec()
                .editOrNewTemplate()
                    .editOrNewSpec()
                        .editContainer(0)
                            .addToEnv(operatorLabelsEnv)
                        .endContainer()
                    .endSpec()
                .endTemplate()
            .endSpec()
            .build());
```
Or something similar.

systemtest/src/test/java/io/strimzi/systemtest/operators/FeatureGatesST.java
```diff
@@ -0,0 +1,93 @@
...
+        clusterOperator.getSpec().getTemplate().getSpec().getContainers().get(0).setEnv(envVarList);
+        return new DeploymentBuilder(clusterOperator).build();
+    }
+
+    /**
+     * Control Plane Listener
+     * https://github.com/strimzi/proposals/blob/main/025-control-plain-listener.md
+     */
+    @Test
+    @ParallelNamespaceTest
```
im-konge, line 76: `@IsolatedTest` here will be much better. Imagine running for example 4 tests in parallel, which will change something in CO. 
`@ParallelNamespaceTest` is for STs which will have same beginning, like the CO, and we want to deploy multiple Kafka clusters.

systemtest/src/test/java/io/strimzi/systemtest/operators/FeatureGatesST.java
```diff
@@ -0,0 +1,93 @@
...
+        redeployCOWithEnvVars(testEnvVars);
+        // Following line adds dependency on cluster-operator
+        resourceManager.createResource(extensionContext,
+                KafkaTemplates.kafkaPersistent(clusterName, 3).build());
+
+        LOGGER.info("Check for presence of ContainerPort 9090/tcp (tcp-ctrlplane) in first Kafka pod.");
+        final Pod kafkaPod = PodUtils.getPodsByPrefixInNameWithDynamicWait(namespaceName, clusterName + "-kafka-").get(0);
+        ContainerPort expectedControlPlaneContainerPort = new ContainerPort(9090, null, null, "tcp-ctrlplane", "TCP");
+        List<ContainerPort> kafkaPodPorts = kafkaPod.getSpec().getContainers().get(0).getPorts();
+        assertTrue(kafkaPodPorts.contains(expectedControlPlaneContainerPort));
```
im-konge, line 91: Now you are testing that CO is deployed with enabled ControlPlane listener, which is then present in the pod spec. But, this can be tested with some simple UT.

When it's the "listener", I would test if we are able to send messages or something like that.

systemtest/src/test/java/io/strimzi/systemtest/operators/FeatureGatesST.java
```diff
@@ -0,0 +1,93 @@
...
+    @ParallelNamespaceTest
+    public void testControlPlaneListenerFeatureGate(ExtensionContext extensionContext) {
+        final String namespaceName = extensionContext.getStore(ExtensionContext.Namespace.GLOBAL).get(Constants.NAMESPACE_KEY).toString();
+        final String clusterName = mapWithClusterNames.get(extensionContext.getDisplayName());
+
+        testEnvVars.add(new EnvVar(Environment.STRIMZI_FEATURE_GATES_ENV, "+ControlPlaneListener", null));
+        redeployCOWithEnvVars(testEnvVars);
+        // Following line adds dependency on cluster-operator
+        resourceManager.createResource(extensionContext,
+                KafkaTemplates.kafkaPersistent(clusterName, 3).build());
```
im-konge, line 85: From my POV you don't need persistent Kafka, `kafkaEphemeral` should be enough

systemtest/src/test/java/io/strimzi/systemtest/operators/FeatureGatesST.java
michalxo: Yeah I will look on it.

systemtest/src/test/java/io/strimzi/systemtest/operators/FeatureGatesST.java
michalxo: I need to change logic of this test class, so I will fix that tomorrow ;)

scholzj (commented):
If you want a single ST for this, it should probably combine sending / receiving messages (through a regular port) combined with some broker failovers. Checking the port is useless, especially since it is there regardless the feature gate.

michalxo:
> If you want a single ST for this, it should probably combine sending / receiving messages (through a regular port) combined with some broker failovers. Checking the port is useless, especially since it is there regardless the feature gate.

So I don't need to worry with exposing 9090 (CPL port) on Kafka and send/recv some messages over there? That's what @im-konge  suggested and made sense to me.
Thank you

scholzj:
> So I don't need to worry with exposing 9090 (CPL port) on Kafka and send/recv some messages over there? That's what @im-konge suggested and made sense to me.

No, the port is not designed for sending and receiving messages by clients. It is the same as 9091 - used internally by Kafka only.

systemtest/src/main/java/io/strimzi/systemtest/Environment.java
michalxo: Partially done, as I need to have name of the ENV, not just it's value.

systemtest/src/test/java/io/strimzi/systemtest/operators/FeatureGatesST.java
michalxo: Changed all.

systemtest/src/main/java/io/strimzi/systemtest/utils/kubeUtils/objects/ServiceUtils.java
```diff
@@ -79,11 +80,24 @@ public static void waitForNodePortService(String serviceName) throws Interrupted
     public static void waitForServiceDeletion(String serviceName) {
         LOGGER.info("Waiting for Service {} deletion in namespace {}", serviceName, kubeClient().getNamespace());
 
+        kubeClient().deleteService(serviceName);
```
im-konge, line 83: This should be just `wait` for the `service` deletion. That means, you need to delete it in the test and then use the `waitForServiceDeletion`.

systemtest/src/main/java/io/strimzi/systemtest/utils/kubeUtils/objects/ServiceUtils.java
```diff
@@ -79,11 +80,24 @@ public static void waitForNodePortService(String serviceName) throws Interrupted
...
+     * @param serviceName service name
+     */
+    public static void waitForServiceCreation(String serviceName) {
+        LOGGER.info("Waiting for Service {} creation in namespace {}", serviceName, kubeClient().getNamespace());
+
+        TestUtils.waitFor("Service " + serviceName + " to be created", Constants.POLL_INTERVAL_FOR_RESOURCE_READINESS, CREATION_TIMEOUT,
+                () -> kubeClient().getService(serviceName) != null);
+        LOGGER.info("Service {} in namespace {} has been created", serviceName, kubeClient().getNamespace());
+    }
+
```
im-konge, line 100: You should add the `String namespaceName` to the parameters and work with that

systemtest/src/test/java/io/strimzi/systemtest/operators/FeatureGatesST.java
```diff
@@ -0,0 +1,110 @@
...
+public class FeatureGatesST extends AbstractST {
+
+    static final String NAMESPACE = "feature-gates-tests";
+    private static final Logger LOGGER = LogManager.getLogger(FeatureGatesST.class);
+
+    protected void redeployClusterOperatorWithEnvVars(ExtensionContext extensionContext, String namespace, List<EnvVar> envVars) {
+        prepareEnvForOperator(extensionContext, namespace, Arrays.asList(namespace, namespace));
+        applyBindings(extensionContext, namespace);
+        resourceManager.createResource(extensionContext, BundleResource.clusterOperator(namespace, envVars).build());
+    }
```
im-konge, line 51: I would name it `deployClusterOperatorWithEnvVars`. The `namespace` here will be always the same from my POV, so you can work just with the `NAMESPACE` var.

systemtest/src/test/java/io/strimzi/systemtest/operators/FeatureGatesST.java
```diff
@@ -0,0 +1,110 @@
...
+
+    /**
+     * Control Plane Listener
+     * https://github.com/strimzi/proposals/blob/main/025-control-plain-listener.md
+     */
+    @IsolatedTest("Feature Gates test for ControlPlainListener")
+    @Tag(INTERNAL_CLIENTS_USED)
+    public void testControlPlaneListenerFeatureGate(ExtensionContext extensionContext) throws InterruptedException {
+        final String namespaceName = NAMESPACE;
```
im-konge, line 60: This is not needed, right?

systemtest/src/main/java/io/strimzi/systemtest/utils/kubeUtils/objects/ServiceUtils.java
michalxo: Why do I need to add namespace there? It's already there `kubeClient().getNamespace()`.

im-konge: What if we want to wait for service in different namespace? You can add there some encapsulation, but don't forget we have parallel namespaces and the context in the `kubeClient().getNamespace()` can point to completely different namespace.

michalxo:
@strimzi-ci run tests test_only profile=regression testcase=FeatureGatesST

systemtest/src/main/java/io/strimzi/systemtest/resources/specific/HelmResource.java
```diff
@@ -115,7 +115,8 @@ private void clusterOperator(long operationTimeout, long reconciliationInterval)
                 entry("resources.limits.cpu", LIMITS_CPU),
                 entry("logLevelOverride", Environment.STRIMZI_LOG_LEVEL),
                 entry("fullReconciliationIntervalMs", Long.toString(reconciliationInterval)),
-                entry("operationTimeoutMs", Long.toString(operationTimeout)))
+                entry("operationTimeoutMs", Long.toString(operationTimeout)),
+                entry("featureGates", Environment.STRIMZI_LOG_LEVEL))
```
scholzj, line 119: `Environment.STRIMZI_LOG_LEVEL` does not look to be the right env.

systemtest/src/test/java/io/strimzi/systemtest/operators/FeatureGatesST.java
```diff
@@ -0,0 +1,112 @@
...
+                .withNamespaceName(NAMESPACE)
+                .build();
+
+        resourceManager.createResource(extensionContext, kafkaBasicClientJob.producerStrimzi().build());
+        resourceManager.createResource(extensionContext, kafkaBasicClientJob.consumerStrimzi().build());
+
+        LOGGER.info("Remove kafka-broker service and wait for it's recreation (should be done within 30secs)");
+        KubeClusterResource.kubeClient().deleteService(KafkaResources.brokersServiceName(clusterName));
+        ServiceUtils.waitForServiceDeletion(KafkaResources.brokersServiceName(clusterName), NAMESPACE);
+        ServiceUtils.waitForServiceCreation(KafkaResources.brokersServiceName(clusterName), NAMESPACE);
```
scholzj, line 100: I would remove this. I do not think it adds any value in this test.

systemtest/src/test/java/io/strimzi/systemtest/operators/FeatureGatesST.java
```diff
@@ -0,0 +1,112 @@
...
+
+        LOGGER.info("Check for presence of ContainerPort 9090/tcp (tcp-ctrlplane) in first Kafka pod.");
+        final Pod kafkaPod = PodUtils.getPodsByPrefixInNameWithDynamicWait(NAMESPACE, clusterName + "-kafka-").get(0);
+        ContainerPort expectedControlPlaneContainerPort = new ContainerPort(9090, null, null, "tcp-ctrlplane", "TCP");
+        List<ContainerPort> kafkaPodPorts = kafkaPod.getSpec().getContainers().get(0).getPorts();
+        assertTrue(kafkaPodPorts.contains(expectedControlPlaneContainerPort));
+
+        LOGGER.info("Try to send some messages to Kafka over the next {} seconds.", sendingTime);
+        // use example clients for jobs - sending messages for 3-5mins
+        KafkaBasicExampleClients kafkaBasicClientJob = new KafkaBasicExampleClients.Builder()
```
scholzj, line 84: Where do we create the topic used by this? Can we make sure it has multiple replicas?

systemtest/src/test/java/io/strimzi/systemtest/operators/FeatureGatesST.java
```diff
@@ -0,0 +1,112 @@
...
+        resourceManager.createResource(extensionContext, kafkaBasicClientJob.consumerStrimzi().build());
+
+        LOGGER.info("Remove kafka-broker service and wait for it's recreation (should be done within 30secs)");
+        KubeClusterResource.kubeClient().deleteService(KafkaResources.brokersServiceName(clusterName));
+        ServiceUtils.waitForServiceDeletion(KafkaResources.brokersServiceName(clusterName), NAMESPACE);
+        ServiceUtils.waitForServiceCreation(KafkaResources.brokersServiceName(clusterName), NAMESPACE);
+
+        LOGGER.info("Force Rolling Update of Kafka by upscaling replicas to {}", kafkaReplicas + 2);
+        KafkaResource.replaceKafkaResourceInSpecificNamespace(clusterName, k -> k.getSpec().getKafka().setReplicas(kafkaReplicas + 2), NAMESPACE);
+        StatefulSetUtils.waitForAllStatefulSetPodsReady(NAMESPACE, KafkaResources.kafkaStatefulSetName(clusterName), kafkaReplicas + 2, ResourceOperation.getTimeoutForResourceReadiness(Constants.STATEFUL_SET));
```
scholzj, line 104: This might do the job. But I wonder if this is the best solution. What about just deleting one of the broker pods?

systemtest/src/main/java/io/strimzi/systemtest/resources/specific/HelmResource.java
michalxo: Thanks, that's wrong copy/paste. Fixed.

systemtest/src/test/java/io/strimzi/systemtest/operators/FeatureGatesST.java
michalxo: I can do both - delete 1 of 2 kafka pods before RU.

systemtest/src/main/java/io/strimzi/systemtest/utils/kubeUtils/objects/ServiceUtils.java
```diff
@@ -76,12 +77,24 @@ public static void waitForNodePortService(String serviceName) throws Interrupted
      * Wait until Service of the given name will be deleted.
      * @param serviceName service name
      */
-    public static void waitForServiceDeletion(String serviceName) {
-        LOGGER.info("Waiting for Service {} deletion in namespace {}", serviceName, kubeClient().getNamespace());
+    public static void waitForServiceDeletion(String serviceName, String namespace) {
```
Frawless, line 80: Is there any other use case for this method?

systemtest/src/main/java/io/strimzi/systemtest/utils/kubeUtils/objects/ServiceUtils.java
```diff
@@ -76,12 +77,24 @@ public static void waitForNodePortService(String serviceName) throws Interrupted
...
-        LOGGER.info("Service {} in namespace {} was deleted", serviceName, kubeClient().getNamespace());
+            () -> kubeClient().namespace(namespace).getService(serviceName) == null);
+        LOGGER.info("Service {} in namespace {} was deleted", serviceName, namespace);
+    }
+
+    /**
+     * Wait until Service of the given name will be created.
+     * @param serviceName service name
+     */
+    public static void waitForServiceCreation(String serviceName, String namespace) {
```
Frawless, line 92: Do we need to wait for service creation or we can just simply wait for some Kafka readiness with clients?

systemtest/src/test/java/io/strimzi/systemtest/operators/FeatureGatesST.java
Frawless: auto-creation is used I guess

systemtest/src/test/java/io/strimzi/systemtest/operators/FeatureGatesST.java
Frawless: Or just trigger manual RU via annotation?

systemtest/src/test/java/io/strimzi/systemtest/operators/FeatureGatesST.java
```diff
@@ -0,0 +1,112 @@
...
+
+        LOGGER.info("Waiting for clients to finish sending/receiving messages");
+        ClientUtils.waitForClientSuccess(consumerName, NAMESPACE, MESSAGE_COUNT);
+        ClientUtils.waitForClientSuccess(consumerName, NAMESPACE, MESSAGE_COUNT);
+        JobUtils.deleteJobWithWait(NAMESPACE, consumerName);
+        JobUtils.deleteJobWithWait(NAMESPACE, consumerName);
```
Frawless, line 110: You use consumer in both lines. Do we also need to delete them manually? I think `ResourceManager` should do it automatically.

michalxo: I fixed that locally, but hasn't pushed. I rather delete jobs to be sure.

systemtest/src/main/java/io/strimzi/systemtest/utils/kubeUtils/objects/ServiceUtils.java
michalxo: I think this will be removed, as this whole part of service killing is redundant (as per Jakub's comment above) and it just creates problems sometimes - I was unable to debug the issue.

systemtest/src/test/java/io/strimzi/systemtest/operators/FeatureGatesST.java
michalxo: removed

systemtest/src/test/java/io/strimzi/systemtest/operators/FeatureGatesST.java
michalxo: Yes, I will create it before with replication=2

systemtest/src/test/java/io/strimzi/systemtest/operators/FeatureGatesST.java
```diff
@@ -0,0 +1,126 @@
...
+        resourceManager.createResource(extensionContext, kafkaBasicClientJob.consumerStrimzi().build());
+        JobUtils.waitForJobRunning(consumerName, NAMESPACE, Duration.ofSeconds(20).toMillis());
+
+        LOGGER.info("Delete first found Kafka broker pod.");
+        kubeClient(NAMESPACE).deletePod(NAMESPACE, kafkaPod);
+        StatefulSetUtils.waitForAllStatefulSetPodsReady(KafkaResources.kafkaStatefulSetName(clusterName), kafkaReplicas);
+
+        LOGGER.info("Force Rolling Update of Kafka by upscaling replicas to {}.", kafkaReplicas + 2);
+        KafkaResource.replaceKafkaResourceInSpecificNamespace(clusterName, k -> k.getSpec().getKafka().setReplicas(kafkaReplicas + 2), NAMESPACE);
+        StatefulSetUtils.waitForAllStatefulSetPodsReady(NAMESPACE, KafkaResources.kafkaStatefulSetName(clusterName), kafkaReplicas + 2, ResourceOperation.getTimeoutForResourceReadiness(Constants.STATEFUL_SET));
```
scholzj, line 118: You don't need this ... you can just use anno to trigger a regular rolling update if you want. That will at least need less resources and thus be more stable.

systemtest/src/test/java/io/strimzi/systemtest/operators/FeatureGatesST.java
```diff
@@ -0,0 +1,126 @@
...
+        final String topicName = "topic-fg-cpl";
+        int sendingTime = 300;
+        List<EnvVar> testEnvVars = new ArrayList<>();
+        int kafkaReplicas = 2;
+
+        testEnvVars.add(new EnvVar(Environment.STRIMZI_FEATURE_GATES_ENV, "+ControlPlaneListener", null));
+        deployClusterOperatorWithEnvVars(extensionContext, testEnvVars);
+
+        resourceManager.createResource(extensionContext, KafkaTemplates.kafkaPersistent(clusterName, kafkaReplicas).build());
+        StatefulSetUtils.waitForAllStatefulSetPodsReady(KafkaResources.kafkaStatefulSetName(clusterName), kafkaReplicas);
```
im-konge, line 81: This is not needed. You are waiting for the `StatefulSet` and pods to be ready, which is handled by `ResourceManager`

systemtest/src/test/java/io/strimzi/systemtest/operators/FeatureGatesST.java
```diff
@@ -0,0 +1,126 @@
...
+                .withNamespaceName(NAMESPACE)
+                .build();
+
+        resourceManager.createResource(extensionContext, kafkaBasicClientJob.producerStrimzi().build());
+        resourceManager.createResource(extensionContext, kafkaBasicClientJob.consumerStrimzi().build());
+        JobUtils.waitForJobRunning(consumerName, NAMESPACE, Duration.ofSeconds(20).toMillis());
+
+        LOGGER.info("Delete first found Kafka broker pod.");
+        kubeClient(NAMESPACE).deletePod(NAMESPACE, kafkaPod);
+        StatefulSetUtils.waitForAllStatefulSetPodsReady(KafkaResources.kafkaStatefulSetName(clusterName), kafkaReplicas);
```
im-konge, line 114: I would use this `PodUtils.deletePodWithWait()`. If it not work, you should take a look why (I remember you mentioned it offline). Because in different STs it's working without problem.

systemtest/src/test/java/io/strimzi/systemtest/operators/FeatureGatesST.java
```diff
@@ -0,0 +1,126 @@
...
+     * https://github.com/strimzi/proposals/blob/main/025-control-plain-listener.md
+     */
+    @IsolatedTest("Feature Gates test for ControlPlainListener")
+    @Tag(INTERNAL_CLIENTS_USED)
+    public void testControlPlaneListenerFeatureGate(ExtensionContext extensionContext) throws InterruptedException {
+        final String clusterName = mapWithClusterNames.get(extensionContext.getDisplayName());
+        final String producerName = "producer-test";
+        final String consumerName = "consumer-test";
+        final String topicName = "topic-fg-cpl";
+        int sendingTime = 300;
```
im-konge, line 73: Is this variable for message count? If yes, shouldn't it have different name?

systemtest/src/test/java/io/strimzi/systemtest/operators/FeatureGatesST.java
```diff
@@ -0,0 +1,126 @@
...
+    private static final Logger LOGGER = LogManager.getLogger(FeatureGatesST.class);
+
+    protected void deployClusterOperatorWithEnvVars(ExtensionContext extensionContext, List<EnvVar> envVars) {
+        if (Environment.isOlmInstall()) {
+            cluster.createNamespaces(NAMESPACE, Arrays.asList(NAMESPACE, NAMESPACE));
+        } else {
+            prepareEnvForOperator(extensionContext, NAMESPACE, Arrays.asList(NAMESPACE, NAMESPACE));
+        }
+        applyBindings(extensionContext, NAMESPACE);
+        resourceManager.createResource(extensionContext, BundleResource.clusterOperator(NAMESPACE, envVars).build());
```
im-konge, line 59: If you want to make it work with OLM, Helm etc. you should do it differently. 
Now you deploy CO by bundle installation (because of `BundleResource`). 
Look at the `installClusterOperator()` method, where we are checking, by which method should be CO installed.

systemtest/src/test/java/io/strimzi/systemtest/operators/FeatureGatesST.java
michalxo: I am triggerring RU by changing KafkaReplicas from 2 -> 4.

systemtest/src/main/java/io/strimzi/systemtest/utils/kubeUtils/objects/ServiceUtils.java
michalxo: removed...

systemtest/src/test/java/io/strimzi/systemtest/operators/FeatureGatesST.java
Frawless: Ye this is definitely wrong. I think you should use only the Bundle install type in this test and leave OLM and HELM to setup feature gates via env var.

systemtest/src/test/java/io/strimzi/systemtest/operators/FeatureGatesST.java
michalxo: This method is not suitable for my needs, as it expects that pod will be already in terminating phase, which is not happening (too quick), so it's ending up in loop.
Imo `PodUtils.deletePodWithWait()` is wrongly named method and should be - `waitForDeletionOfTerminatedPod` instead.

im-konge: And why you don't update the method? I guess there is just wrong `if` statement.
Before:
```
if (pod.getStatus().getPhase().equals("Terminating")) {
                            LOGGER.debug("Deleting pod {}", pod.getMetadata().getName());
                            cmdKubeClient(namespaceName).deleteByName("pod", pod.getMetadata().getName());
                        }
```
After:
```
if (!pod.getStatus().getPhase().equals("Terminating")) {
                            LOGGER.debug("Deleting pod {}", pod.getMetadata().getName());
                            cmdKubeClient(namespaceName).deleteByName("pod", pod.getMetadata().getName());
                        }
```
It seems that it's just some leftover, so if it doesn't work how it should, just fix it.

systemtest/src/test/java/io/strimzi/systemtest/operators/FeatureGatesST.java
michalxo: Thank you for ideas. 
Helm with externally defined ENV var works fine. 
Now to resolve OLM and I will push more changes here.

michalxo: Tested on OLM as well with 
```
CLUSTER_OPERATOR_INSTALL_TYPE=OLM
STRIMZI_FEATURE_GATES=+ControlPlaneListener
OLM_OPERATOR_VERSION=0.23.0
```

systemtest/src/test/java/io/strimzi/systemtest/operators/FeatureGatesST.java
```diff
@@ -0,0 +1,128 @@
...
+        resourceManager.createResource(extensionContext, kafkaBasicClientJob.consumerStrimzi().build());
+        JobUtils.waitForJobRunning(consumerName, NAMESPACE, Duration.ofSeconds(20).toMillis());
+
+        LOGGER.info("Delete first found Kafka broker pod.");
+        kubeClient(NAMESPACE).deletePod(NAMESPACE, kafkaPod);
+        StatefulSetUtils.waitForAllStatefulSetPodsReady(KafkaResources.kafkaStatefulSetName(clusterName), kafkaReplicas);
+
+        LOGGER.info("Force Rolling Update of Kafka via annotation.");
+        KafkaResource.replaceKafkaResourceInSpecificNamespace(clusterName, k -> k.getMetadata().setAnnotations(Map.of(Annotations.ANNO_STRIMZI_IO_MANUAL_ROLLING_UPDATE, "true")), NAMESPACE);
+        StatefulSetUtils.waitForAllStatefulSetPodsReady(NAMESPACE, KafkaResources.kafkaStatefulSetName(clusterName), kafkaReplicas, ResourceOperation.getTimeoutForResourceReadiness(Constants.STATEFUL_SET));
```
scholzj, line 120: The rolling update will not happen immediately here. So I wonder if this `waitForAllStatefulSetPodsReady` really waits for it. I would expect that it passes in most cases before the rolling update actually happens.

michalxo:
@strimzi-ci run tests test_only profile=regression testcase=FeatureGatesST

michalxo:
@strimzi-ci run tests test_only profile=regression testcase=FeatureGatesST

scholzj (approved):

systemtest/src/test/java/io/strimzi/systemtest/operators/FeatureGatesST.java
```diff
@@ -0,0 +1,134 @@
...
+    static final String NAMESPACE = "feature-gates-tests";
+    private static final Logger LOGGER = LogManager.getLogger(FeatureGatesST.class);
+
+    protected void deployClusterOperatorWithEnvVars(ExtensionContext extensionContext, List<EnvVar> envVars) {
+        if (Environment.isOlmInstall() || Environment.isHelmInstall()) {
+            installClusterOperator(extensionContext, NAMESPACE);
+        } else {
+            prepareEnvForOperator(extensionContext, NAMESPACE, Arrays.asList(NAMESPACE, NAMESPACE));
+            applyBindings(extensionContext, NAMESPACE);
+            resourceManager.createResource(extensionContext, BundleResource.clusterOperator(NAMESPACE, envVars).build());
```
Frawless, line 62: What if I will not set env var for feature gates and will try to install CO via OLM?

systemtest/src/test/java/io/strimzi/systemtest/operators/FeatureGatesST.java
```diff
@@ -0,0 +1,134 @@
...
+        LOGGER.info("Delete first found Kafka broker pod.");
+        kubeClient(NAMESPACE).deletePod(NAMESPACE, kafkaPod);
+        StatefulSetUtils.waitForAllStatefulSetPodsReady(KafkaResources.kafkaStatefulSetName(clusterName), kafkaReplicas);
+
+        LOGGER.info("Force Rolling Update of Kafka via annotation.");
+        KafkaResource.replaceKafkaResourceInSpecificNamespace(clusterName, k -> k.getMetadata().setAnnotations(Map.of(Annotations.ANNO_STRIMZI_IO_MANUAL_ROLLING_UPDATE, "true")), NAMESPACE);
+        // Wait some time for Strimzi Operator reconciliation to happen (so annotation is parsed and evaluated).
+        LOGGER.info("Wait {} ms for next reconciliation", Constants.RECONCILIATION_INTERVAL);
+        Thread.sleep(Constants.RECONCILIATION_INTERVAL);
+        StatefulSetUtils.waitForAllStatefulSetPodsReady(NAMESPACE, KafkaResources.kafkaStatefulSetName(clusterName), kafkaReplicas, ResourceOperation.getTimeoutForResourceReadiness(Constants.STATEFUL_SET));
```
Frawless, line 126: Why don't you wait for pods rolling update? You just opened a door for race condition by this code I think.

Frawless: `StatefulSetUtils.waitTillSsHasRolled`

systemtest/src/test/java/io/strimzi/systemtest/operators/FeatureGatesST.java
```diff
@@ -0,0 +1,134 @@
...
+                .withBootstrapAddress(KafkaResources.plainBootstrapAddress(clusterName))
+                .withTopicName(topicName)
+                .withMessageCount(messageCount)
+                .withDelayMs(500)
+                .withNamespaceName(NAMESPACE)
+                .build();
+
+        resourceManager.createResource(extensionContext, kafkaBasicClientJob.producerStrimzi().build());
+        resourceManager.createResource(extensionContext, kafkaBasicClientJob.consumerStrimzi().build());
+        JobUtils.waitForJobRunning(consumerName, NAMESPACE, Duration.ofSeconds(20).toMillis());
```
Frawless, line 115: Why you don't wait for job success?

systemtest/src/test/java/io/strimzi/systemtest/operators/FeatureGatesST.java
michalxo: Then you are running this test incorrectly. It has been stated multiple times to me, that only Bundle installation type should be used w/o env var. :-) 
OLM and Helm should use external Var set (optionally for Bundle).

systemtest/src/test/java/io/strimzi/systemtest/operators/FeatureGatesST.java
michalxo: I did not know about that fucntion. Going to use it. thanks!

systemtest/src/test/java/io/strimzi/systemtest/operators/FeatureGatesST.java
michalxo: I used it at the end of test. Here - I just make sure, that job/clients are running and sending messages. So I can proceed with restarting/rolling brokers etc.

systemtest/src/test/java/io/strimzi/systemtest/operators/FeatureGatesST.java
Frawless: so what's the purpose of this test for HELM and OLM as you have it now? It kinda doesn't make sense because it will test the same thing which is already tested by other tests. You should have there only bundle install type and assumption, that `CLUSTER_OPERATOR_INSTALL_TYPE=bundle`

michalxo: Removed running this testcase with OLM and Helm installations after offline talk. -> `assumeFalse()`

michalxo:
@strimzi-ci run tests test_only profile=regression testcase=FeatureGatesST

systemtest/src/test/java/io/strimzi/systemtest/operators/FeatureGatesST.java
```diff
@@ -0,0 +1,137 @@
...
+
+        LOGGER.info("Delete first found Kafka broker pod.");
+        kubeClient(NAMESPACE).deletePod(NAMESPACE, kafkaPod);
+        StatefulSetUtils.waitForAllStatefulSetPodsReady(KafkaResources.kafkaStatefulSetName(clusterName), kafkaReplicas);
+
+        LOGGER.info("Force Rolling Update of Kafka via annotation.");
+        kafkaPods.keySet().forEach(podName -> {
+            kubeClient(NAMESPACE).editPod(podName).edit(pod -> new PodBuilder(pod)
+                    .editMetadata()
+                    .addToAnnotations(Annotations.ANNO_STRIMZI_IO_MANUAL_ROLLING_UPDATE, "true")
```
im-konge, line 124: indent

systemtest/src/test/java/io/strimzi/systemtest/operators/FeatureGatesST.java
```diff
@@ -0,0 +1,137 @@
...
+                .withBootstrapAddress(KafkaResources.plainBootstrapAddress(clusterName))
+                .withTopicName(topicName)
+                .withMessageCount(messageCount)
+                .withDelayMs(500)
+                .withNamespaceName(NAMESPACE)
+                .build();
+
+        resourceManager.createResource(extensionContext, kafkaBasicClientJob.producerStrimzi().build());
+        resourceManager.createResource(extensionContext, kafkaBasicClientJob.consumerStrimzi().build());
+        JobUtils.waitForJobRunning(consumerName, NAMESPACE, Duration.ofSeconds(20).toMillis());
```
im-konge, line 114: Just a nit - if you want to check that producer is running/sending messages, shouldn't it be 
```suggestion
        JobUtils.waitForJobRunning(producerName, NAMESPACE, Duration.ofSeconds(20).toMillis());
```
?

michalxo: Here I wanted to make sure (or rather assume), that if second started client is running, then first is running as well, so we can proceed with restarting broker pods. Fine by me with change or not.

michalxo:
I assume we're good to merge now.

michalxo:
@strimzi-ci run tests test_only profile=regression testcase=FeatureGatesST

systemtest/src/test/java/io/strimzi/systemtest/operators/FeatureGatesST.java
```diff
@@ -0,0 +1,137 @@
...
+ * https://github.com/strimzi/proposals/blob/main/022-feature-gates.md
+ */
+@Tag(REGRESSION)
+public class FeatureGatesST extends AbstractST {
+
+    static final String NAMESPACE = "feature-gates-tests";
+    private static final Logger LOGGER = LogManager.getLogger(FeatureGatesST.class);
+
+    protected void deployClusterOperatorWithEnvVars(ExtensionContext extensionContext, List<EnvVar> envVars) {
+        assumeFalse(Environment.isOlmInstall() || Environment.isHelmInstall());
```
Frawless, line 57: wouldn't be better to have it on the first line of the test/class ? Annotation would be nice, but that could be done in the next PR.

systemtest/src/test/java/io/strimzi/systemtest/operators/FeatureGatesST.java
```diff
@@ -0,0 +1,137 @@
...
+
+        LOGGER.info("Check for presence of ContainerPort 9090/tcp (tcp-ctrlplane) in first Kafka pod.");
+        final Pod kafkaPod = PodUtils.getPodsByPrefixInNameWithDynamicWait(NAMESPACE, clusterName + "-kafka-").get(0);
+        ContainerPort expectedControlPlaneContainerPort = new ContainerPort(9090, null, null, "tcp-ctrlplane", "TCP");
+        List<ContainerPort> kafkaPodPorts = kafkaPod.getSpec().getContainers().get(0).getPorts();
+        assertTrue(kafkaPodPorts.contains(expectedControlPlaneContainerPort));
+
+        Map<String, String> kafkaPods = StatefulSetUtils.ssSnapshot(NAMESPACE, kafkaStatefulSetName(clusterName));
+
+        LOGGER.info("Try to send some messages to Kafka over the next {} seconds.", messageCount);
```
Frawless, line 93: You are sending 300 messages but every message is sent with 500ms delay which won't take 300s in total.

im-konge (approved):
LGTM, @see-quick can you please verify the deletion of OLM?

systemtest/src/main/java/io/strimzi/systemtest/resources/operator/BundleResource.java
```diff
@@ -56,38 +56,42 @@ public boolean waitForReadiness(Deployment resource) {
...
 
     public static DeploymentBuilder clusterOperator(String namespace, long operationTimeout, long reconciliationInterval) {
-        return defaultClusterOperator(Constants.STRIMZI_DEPLOYMENT_NAME, namespace, namespace, operationTimeout, reconciliationInterval);
+        return defaultClusterOperator(Constants.STRIMZI_DEPLOYMENT_NAME, namespace, namespace, operationTimeout, reconciliationInterval, null);
     }
 
     public static DeploymentBuilder clusterOperator(String name, String namespace, String namespaceEnv, long operationTimeout, long reconciliationInterval) {
-        return defaultClusterOperator(name, namespace, namespaceEnv, operationTimeout, reconciliationInterval);
+        return defaultClusterOperator(name, namespace, namespaceEnv, operationTimeout, reconciliationInterval, null);
     }
```
see-quick, line 80: This is not really a good sign of clean code...When you have more than 5-6 overloaded methods it's better to use the Builder pattern. I know that these methods have been there before, but if they are extended, the alternatives I have mentioned should be considered. Like I have done in my current PR https://github.com/strimzi/strimzi-kafka-operator/pull/5105.

systemtest/src/main/java/io/strimzi/systemtest/resources/specific/OlmResource.java
```diff
@@ -67,7 +67,7 @@ public void create(String namespace, OlmInstallationStrategy olmInstallationStra
 
     @Override
     public void delete() {
-        this.deleteOlm(deploymentName, namespace, csvName);
```
see-quick, line 70: What is the point of removing `this` reference? Did you face some issues with that?

michalxo: Yes I did have problems with accessibility of variables. See more changes in that file for example.

see-quick (approved):

Frawless (approved):

PR #5060, opened 2021-06-01
ST: Add test for Kafka Quotas Plugin integration

Frawless:

### Type of change

- Enhancement / new feature

### Description

This PR adds system test which checks the integration of Kafka Quotas Plugin. The test only checks if the integration works based on disk-space quota. More scenarios are hard to implement with the current test suite and they will be part of manual checks for the releases.
### Checklist

- [x] Write tests

### DIFF
--- a/systemtest/src/test/java/io/strimzi/systemtest/kafka/QuotasST.java
+++ b/systemtest/src/test/java/io/strimzi/systemtest/kafka/QuotasST.java
@@ -1,0 +1,95 @@
+/*
+ * Copyright Strimzi authors.
+ * License: Apache License 2.0 (see the file LICENSE or http://apache.org/licenses/LICENSE-2.0.html).
+ */
+package io.strimzi.systemtest.kafka;
+
+import io.strimzi.api.kafka.model.KafkaResources;
+import io.strimzi.systemtest.AbstractST;
+import io.strimzi.systemtest.Constants;
+import io.strimzi.systemtest.annotations.ParallelNamespaceTest;
+import io.strimzi.systemtest.resources.crd.kafkaclients.KafkaBasicExampleClients;
+import io.strimzi.systemtest.templates.crd.KafkaTemplates;
+import io.strimzi.systemtest.templates.crd.KafkaTopicTemplates;
+import io.strimzi.systemtest.utils.kubeUtils.controllers.JobUtils;
+import io.strimzi.test.WaitException;
+import org.apache.logging.log4j.LogManager;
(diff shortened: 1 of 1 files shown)


scholzj (approved):

im-konge (approved):

systemtest/src/test/java/io/strimzi/systemtest/kafka/QuotasST.java
```diff
@@ -0,0 +1,95 @@
...
+            .build();
+
+        resourceManager.createResource(extensionContext, basicClients.producerStrimzi().build());
+        // Kafka Quotas Plugin should stop producer in around 10-20 seconds with configured throughput
+        assertThrows(WaitException.class, () -> JobUtils.waitForJobFailure(producerName, NAMESPACE, 120_000));
+
+        String kafkaLog = kubeClient(namespaceName).logs(KafkaResources.kafkaPodName(clusterName, 0));
+        String softLimitLog = "disk is beyond soft limit";
+        String hardLimitLog = "disk is full";
+        assertThat("Kafka log doesn't contain '" + softLimitLog + "' log", kafkaLog, CoreMatchers.containsString(softLimitLog));
```
samuel-hawker, line 81: This might be a contrived test suggestion, but would it be worth writing a similar test where we only hit the soft limit, but not the hard limit?
I guess the question is who owns the soft/hard limiting capability, this test assumes that soft limiting/hard limiting just works, which is a fine assumption probably.
I guess my question is how difficult would it be to engineer a test that checks the soft limit has been passed, but not the hard limit?

Frawless: With the current capabilities of system tests it might be possible to basically duplicate this test and see that producer finished at some point. I think it might be also part of this test, but it opens a lot of space for flakiness from my POV. We agreed that we will more focus on Kafka Quota Plugin tests during releases and our manual verification where it's much more easy without any proper performance tests.

Frawless: I think we can create an issue for some performance tests which could cover it. WDYT?

samuel-hawker: Sorry didn't see your response till now!
I think that sounds more than reasonable,  and I don't think we want to introduce flaky tests, I think manual checks and a trust that the quota plugin works as intended is more than enough.

Issue #1559, opened 2019-04-18
Previous versions of chart broken after new release


bbrewer-papa-x:
After 0.11.2 release downloads of 0.10.0 are failing. This has happened before, see https://github.com/strimzi/strimzi-kafka-operator/issues/1411. The error message received is
```
Downloading strimzi-kafka-operator from repo https://strimzi.io/charts
Save error occurred:  could not download https://github.com/strimzi/strimzi-kafka-operator/releases/download/0.11.0-rc1/helm-charts/strimzi-kafka-operator-0.10.0.tgz: Failed to fetch https://github.com/strimzi/strimzi-kafka-operator/releases/download/0.11.0-rc1/helm-charts/strimzi-kafka-operator-0.10.0.tgz : 404 Not Found
```

1. In a helm repository
2. Create or edit `requirements.yaml` to include
```
dependencies:
  - name: strimzi-kafka-operator
    version: 0.10.0
    repository: https://strimzi.io/charts
    condition: strimzi.enable
```
3. Run commands
```
helm repo add strimzi https://strimzi.io/charts
helm repo update
helm dependency build chart
```
```
Hang tight while we grab the latest from your chart repositories...
...Successfully got an update from the "strimzi" chart repository
...Successfully got an update from the "stable" chart repository
Update Complete. ⎈Happy Helming!⎈
Saving 1 charts
Downloading strimzi-kafka-operator from repo https://strimzi.io/charts
Save error occurred:  could not download https://github.com/strimzi/strimzi-kafka-operator/releases/download/0.11.0-rc1/helm-charts/strimzi-kafka-operator-0.10.0.tgz: Failed to fetch https://github.com/strimzi/strimzi-kafka-operator/releases/download/0.11.0-rc1/helm-charts/strimzi-kafka-operator-0.10.0.tgz : 404 Not Found
Deleting newly downloaded charts, restoring pre-update state
Error: could not download https://github.com/strimzi/strimzi-kafka-operator/releases/download/0.11.0-rc1/helm-charts/strimzi-kafka-operator-0.10.0.tgz: Failed to fetch https://github.com/strimzi/strimzi-kafka-operator/releases/download/0.11.0-rc1/helm-charts/strimzi-kafka-operator-0.10.0.tgz : 404 Not Found
```

Expected behavior:
Should not get a 404

Environment:
 - Strimzi version: 0.10.0
 - Installation method: Helm chart
 - Kubernetes cluster: Kubernetes 1.11
[...]

PR #5066, opened 2021-06-01
Bump Cruise Control to latest

Boojapho:
### Type of change

- Enhancement / new feature

### Description

Bump Cruise Control to latest to inherit bug and security fixes

### Checklist

- [x] Write tests
- [x] Update documentation
- [x] Update CHANGELOG.md

### DIFF
--- a/documentation/modules/cruise-control/con-rebalance-performance.adoc
+++ b/documentation/modules/cruise-control/con-rebalance-performance.adoc
@@ -33,11 +33,12 @@ Cluster rebalance performance is also influenced by the _replica movement strate
 By default, Cruise Control uses the `BaseReplicaMovementStrategy`, which simply applies the commands in the order they were generated.
 However, if there are some very large partition reassignments early in the proposal, this strategy can slow down the application of the other reassignments.
 
-Cruise Control provides three alternative replica movement strategies that can be applied to optimization proposals:
+Cruise Control provides four alternative replica movement strategies that can be applied to optimization proposals:
 
 * `PrioritizeSmallReplicaMovementStrategy`: Order reassignments in order of ascending size.
 * `PrioritizeLargeReplicaMovementStrategy`: Order reassignments in order of descending size. 
 * `PostponeUrpReplicaMovementStrategy`: Prioritize reassignments for replicas of partitions which have no out-of-sync replicas.
+* `PrioritizeMinIsrWithOfflineReplicasStrategy`: Prioritize reassignments with (At/Under)MinISR partitions with offline replicas.
 
 These strategies can be configured as a sequence.
 The first strategy attempts to compare two partition reassignments using its internal logic. 
--- a/documentation/modules/cruise-control/con-rebalance-performance.adoc
+++ b/documentation/modules/cruise-control/con-rebalance-performance.adoc
@@ -38,7 +38,7 @@ Cruise Control provides four alternative replica movement strategies that can be
 * `PrioritizeSmallReplicaMovementStrategy`: Order reassignments in order of ascending size.
 * `PrioritizeLargeReplicaMovementStrategy`: Order reassignments in order of descending size. 
 * `PostponeUrpReplicaMovementStrategy`: Prioritize reassignments for replicas of partitions which have no out-of-sync replicas.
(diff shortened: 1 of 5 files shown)


scholzj (commented):
The code looks good. I will run the system tests. Could you please fix the DCO sign-off? The instructions should be under the _Details_ link next to the DCO status.

samuel-hawker (approved):
LGTM
Quick question, I see that our CC versions in 2.7.x vs 2.8.x were different
image
Was there a reason for this and should this be taken into account for the current PR.
Given this is a minor update, I don't think it matters, but thought it worth mentioning.

samuel-hawker:
^^ The above is more of a question for @scholzj and the other maintainers.

tomncooper:
Thanks for the PR!

This new version adds a new replica movement strategy options from [CC #1559](https://github.com/linkedin/cruise-control/pull/1559) which should be added to the relevant sections of the [documentation](https://strimzi.io/docs/operators/latest/using.html#con-rebalance-str). It will be important to note that:
```
Current PR assumes that if users intend to use the new PrioritizeMinIsrWithOfflineReplicasStrategy
strategy, then minISR-based concurrency adjustment is enabled. Otherwise, the 
PrioritizeMinIsrWithOfflineReplicasStrategy strategy is just a no-op.
```

tomncooper:
> ^^ The above is more of a question for @scholzj and the other maintainers.

As far as I understand it, there is no reason for them to be different. That was just an oversight on our part.

Boojapho:
> The code looks good. I will run the system tests. Could you please fix the DCO sign-off? The instructions should be under the _Details_ link next to the DCO status.

I've fixed the DCO now.

Boojapho:
> Thanks for the PR!
> 
> This new version adds a new replica movement strategy options from [CC #1559](https://github.com/linkedin/cruise-control/pull/1559) which should be added to the relevant sections of the [documentation](https://strimzi.io/docs/operators/latest/using.html#con-rebalance-str). It will be important to note that:
> 
> ```
> Current PR assumes that if users intend to use the new PrioritizeMinIsrWithOfflineReplicasStrategy
> strategy, then minISR-based concurrency adjustment is enabled. Otherwise, the 
> PrioritizeMinIsrWithOfflineReplicasStrategy strategy is just a no-op.
> ```

@tomncooper I think I have updated the docs for what is needed.  Appreciate if you could take a look.

scholzj (approved):

documentation/modules/cruise-control/con-rebalance-performance.adoc
```diff
@@ -33,11 +33,12 @@ Cluster rebalance performance is also influenced by the _replica movement strate
 By default, Cruise Control uses the `BaseReplicaMovementStrategy`, which simply applies the commands in the order they were generated.
 However, if there are some very large partition reassignments early in the proposal, this strategy can slow down the application of the other reassignments.
 
-Cruise Control provides three alternative replica movement strategies that can be applied to optimization proposals:
+Cruise Control provides four alternative replica movement strategies that can be applied to optimization proposals:
 
 * `PrioritizeSmallReplicaMovementStrategy`: Order reassignments in order of ascending size.
 * `PrioritizeLargeReplicaMovementStrategy`: Order reassignments in order of descending size. 
 * `PostponeUrpReplicaMovementStrategy`: Prioritize reassignments for replicas of partitions which have no out-of-sync replicas.
+* `PrioritizeMinIsrWithOfflineReplicasStrategy`: Prioritize reassignments with (At/Under)MinISR partitions with offline replicas.  This strategy will only work if Cruise Control has been configured with `min_isr_based_concurrency_adjustment` set to true.
```
tomncooper, line 41: We don't support the CC REST API `/admin` endpoint so a Strimzi user cannot set `min_isr_based_concurrency_adjustment`. They would need to set the `concurrency.adjuster.min.isr.check.enabled` in the Cruise Control section of the Kafka CR.

Boojapho: Updated to "if `cruiseControl.config.concurrency.adjuster.min.isr.check.enabled` is set to `true` in the `Kafka` custom resource's spec."

scholzj:
@tomncooper Does it look good now? I can deal with the conflict later.

tomncooper (approved):
LGTM

scholzj:
Thanks for the PR.

Issue #3986, opened 2020-11-20
Metrics ConfigMap contains JSON under YML property


sknot-rh:
While working on https://github.com/strimzi/strimzi-kafka-operator/pull/3855 I have noticed we are using [metrics-config.yml](https://github.com/strimzi/strimzi-kafka-operator/blob/5f09629ec9c0b2d60075d999017376f546df4d53/cluster-operator/src/main/java/io/strimzi/operator/cluster/model/AbstractModel.java#L113) as a property for the generated Logging and Metrics ConfigMap.
From the code we can see we are putting a JSON to the YML property: https://github.com/strimzi/strimzi-kafka-operator/blob/5f09629ec9c0b2d60075d999017376f546df4d53/cluster-operator/src/main/java/io/strimzi/operator/cluster/model/AbstractModel.java#L530
We should be consistent with the data format.

scholzj:
What is the solution for this @sknot-rh? Can we use YAML to pass to the JMX exporter? Or should we rename the key in the ConfigMap?

sknot-rh:
Renaming the key should be enough.

sknot-rh:
Fixed in https://github.com/strimzi/strimzi-kafka-operator/pull/5072

PR #5072, opened 2021-06-02
Fix for Metrics ConfigMap containing JSON under YML property #3986

ShubhamRwt:

### Type of change

_Select the type of your PR_

- Bugfix

### Description

It is fix for issue #3986. We were using metrics-config.yml as a property for the generated Logging and Metrics ConfigMap and mapping json to yml.  So I renamed the property to `metrics-config.json`.

### Checklist

- [ ] Write tests
- [ ] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/cluster-operator/src/test/java/io/strimzi/operator/cluster/TestUtils.java
+++ b/cluster-operator/src/test/java/io/strimzi/operator/cluster/TestUtils.java
@@ -12,7 +12,6 @@ import io.fabric8.kubernetes.client.dsl.Resource;
 import io.strimzi.api.kafka.model.JmxPrometheusExporterMetrics;
 import io.strimzi.api.kafka.model.JmxPrometheusExporterMetricsBuilder;
 import io.strimzi.api.kafka.model.status.Status;
-import io.strimzi.operator.cluster.model.AbstractModel;
 
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
@@ -35,12 +34,12 @@ public class TestUtils {
         return metricsConfig;
     }
 
-    public static ConfigMap getJmxMetricsCm(String data, String metricsCMName) {
+    public static ConfigMap getJmxMetricsCm(String data, String metricsCMName, String metricsConfigYaml) {
         ConfigMap metricsCM = new ConfigMapBuilder()
                 .withNewMetadata()
                 .withName(metricsCMName)
                 .endMetadata()
-                .withData(singletonMap(AbstractModel.ANCILLARY_CM_KEY_METRICS, data))
+                .withData(singletonMap(metricsConfigYaml, data))
                 .build();
         return metricsCM;
     }
--- a/cluster-operator/src/test/java/io/strimzi/operator/cluster/operator/assembly/KafkaAssemblyOperatorTest.java
+++ b/cluster-operator/src/test/java/io/strimzi/operator/cluster/operator/assembly/KafkaAssemblyOperatorTest.java
@@ -771,7 +771,7 @@ public class KafkaAssemblyOperatorTest {
         Map<String, Object> metricsCmJson = metrics ? METRICS_CONFIG : null;
         KafkaExporterSpec exporter = metrics ? new KafkaExporterSpec() : null;
         String metricsCMName = "metrics-cm";
-        JmxPrometheusExporterMetrics jmxMetricsConfig = metrics ? null : io.strimzi.operator.cluster.TestUtils.getJmxPrometheusExporterMetrics(AbstractModel.ANCILLARY_CM_KEY_METRICS, metricsCMName);
+        JmxPrometheusExporterMetrics jmxMetricsConfig = metrics ? null : io.strimzi.operator.cluster.TestUtils.getJmxPrometheusExporterMetrics("metrics-config.yml", metricsCMName);
 
         Kafka resource = ResourceUtils.createKafka(clusterNamespace, clusterName, replicas, image, healthDelay, healthTimeout, jmxMetricsConfig, kafkaConfig, zooConfig, kafkaStorage, zkStorage, LOG_KAFKA_CONFIG, LOG_ZOOKEEPER_CONFIG, exporter, null);
 
@@ -988,7 +988,7 @@ public class KafkaAssemblyOperatorTest {
                 .withNewMetadata()
                     .withName("metrics-cm")
                 .endMetadata()
-                .withData(singletonMap("metrics-config.json", ""))
+                .withData(singletonMap("metrics-config.yml", ""))
                 .build();
         ConfigMap metricsAndLoggingCm = originalKafkaCluster.generateAncillaryConfigMap(new MetricsAndLogging(metricsCm, null), emptySet(), emptySet(), false);
         when(mockCmOps.get(clusterNamespace, KafkaCluster.metricAndLogConfigsName(clusterName))).thenReturn(metricsAndLoggingCm);
--- a/cluster-operator/src/test/java/io/strimzi/operator/cluster/operator/assembly/KafkaAssemblyOperatorTest.java
+++ b/cluster-operator/src/test/java/io/strimzi/operator/cluster/operator/assembly/KafkaAssemblyOperatorTest.java
@@ -155,7 +155,7 @@ public class KafkaAssemblyOperatorTest {
     private final String metricsCmJson = "{\"foo\":\"bar\"}";
     private final String metricsCMName = "metrics-cm";
     private final String differentMetricsCMName = "metrics-cm-2";
-    private final ConfigMap metricsCM = io.strimzi.operator.cluster.TestUtils.getJmxMetricsCm(metricsCmJson, metricsCMName);
+    private final ConfigMap metricsCM = io.strimzi.operator.cluster.TestUtils.getJmxMetricsCm(metricsCmJson, metricsCMName, "metrics-config.yml");
 
     private final KubernetesVersion kubernetesVersion = KubernetesVersion.V1_20;
 
@@ -881,7 +881,7 @@ public class KafkaAssemblyOperatorTest {
     public void testUpdateZkClusterMetricsConfig(Params params, VertxTestContext context) {
         setFields(params);
         Kafka kafkaAssembly = getKafkaAssembly("bar");
-        JmxPrometheusExporterMetrics jmxMetricsConfig = io.strimzi.operator.cluster.TestUtils.getJmxPrometheusExporterMetrics(AbstractModel.ANCILLARY_CM_KEY_METRICS, differentMetricsCMName);
+        JmxPrometheusExporterMetrics jmxMetricsConfig = io.strimzi.operator.cluster.TestUtils.getJmxPrometheusExporterMetrics("metrics-config.yml", differentMetricsCMName);
         kafkaAssembly.getSpec().getKafka().setMetricsConfig(jmxMetricsConfig);
         updateCluster(context, getKafkaAssembly("bar"), kafkaAssembly);
     }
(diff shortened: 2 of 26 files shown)


scholzj (commented):
Have you tested this manually? Does it work? I think if nothing else you would need to change the name of the file as it is referred to in all of the container images.

ShubhamRwt:
> Have you tested this manually? Does it work? I think if nothing else you would need to change the name of the file as it is referred to in all of the container images.

@scholzj Hii, I ran all the unit test after change the key, all those went well and the I am running the integration tests locally too, they are running

scholzj:
The thing you should ideally do with every PR ... build the images, try it live. Unit or integration tests can never replace that ;-)

ShubhamRwt:
Ohh I got it now, Sorry I took the manual testing part as unit and integration test , I built the project earlier and the build was successful. I am testing out by deploying it now

ShubhamRwt:
@scholzj I got what you were suggesting above. I am making those changes and will push them by tomorrow

ShubhamRwt:
@scholzj I have tested the above changes after deploying the operator image. Now it is working good.
.

examples/metrics/kafka-connect-metrics.yaml
```diff
@@ -13,7 +13,7 @@ spec:
     valueFrom:
       configMapKeyRef:
         name: connect-metrics
-        key: metrics-config.yml
+        key: metrics-config.json
```
scholzj, line 16: This is actually YAML ... so this should stay as it was ... no cheating with _replace all_ ;-)

examples/metrics/kafka-connect-metrics.yaml
```diff
@@ -23,7 +23,7 @@ metadata:
   labels:
     app: strimzi
 data:
-  metrics-config.yml: |
+  metrics-config.json: |
```
scholzj, line 26: Same as above ...

examples/metrics/kafka-cruise-control-metrics.yaml
```diff
@@ -36,7 +36,7 @@ spec:
       valueFrom:
         configMapKeyRef:
           name: cruise-control-metrics
-          key: metrics-config.yml
+          key: metrics-config.json
```
scholzj, line 39: Same as above ... I will not repeat it anymore, it should apply to all examples.

systemtest/src/main/java/io/strimzi/systemtest/Constants.java
```diff
@@ -138,7 +138,7 @@
     String PATH_TO_KAFKA_CONNECT_S2I_CONFIG = PATH_TO_PACKAGING_EXAMPLES + "/connect/kafka-connect-s2i.yaml";
     String PATH_TO_KAFKA_MIRROR_MAKER_2_METRICS_CONFIG = PATH_TO_PACKAGING_EXAMPLES + "/metrics/kafka-mirror-maker-2-metrics.yaml";
 
-    String METRICS_CONFIG_YAML_NAME = "metrics-config.yml";
+    String METRICS_CONFIG_JSON_NAME = "metrics-config.json";
```
scholzj, line 141: Same as with the examples ... I suspect in the system tests these should stay as YAML files.

scholzj (changes requested):

systemtest/src/main/java/io/strimzi/systemtest/Constants.java
ShubhamRwt: I had some confusion with the changes, Is only renaming this key to metrics-key.json sufficient. In system tests where the metric-config.yml is used, there is some yaml to json conversion going on.

scholzj: Users are using YAML -> in the custom resources and in the Config Maps. That should not change. Only the internal use of it is changed. If this is used for both internal and external, you would need to create two keys and update the STs.

ShubhamRwt: Okay. Now I get it. I was thinking that the for both external and internal thing we are going the json format. Thanks a lot for clearing my doubt.

cluster-operator/src/test/java/io/strimzi/operator/cluster/operator/assembly/KafkaAssemblyOperatorTest.java
```diff
@@ -1001,7 +1001,7 @@ private void updateCluster(VertxTestContext context, Kafka originalAssembly, Kaf
                 .withNewMetadata()
                     .withName("metrics-cm")
                 .endMetadata()
-                .withData(singletonMap("metrics-config.yml", ""))
+                .withData(singletonMap("metrics-config.json", ""))
```
scholzj, line 1004: Are you sure this should be `.json`? This looks like the user Config Map to me.

ShubhamRwt: @scholzj Yes I guess, If we run the tests with `.yml`, then we get a InvalidResourceEception error, since its searching for the key `metrics-config.json` in `metrics-cm'

systemtest/src/main/java/io/strimzi/systemtest/Constants.java
scholzj: I think it is probably because of this line which uses the internal name as a name for the external ConfigMap field: https://github.com/strimzi/strimzi-kafka-operator/blob/50b22d2490aed3ca19c20d89f31cdb60b73a148f/cluster-operator/src/test/java/io/strimzi/operator/cluster/operator/assembly/KafkaAssemblyOperatorTest.java#L774

I think you should use here some custom name and not `AbstractModel.ANCILLARY_CM_KEY_METRICS` and then use the custom name also in this config map here. Tat would make it much more clear what exactly this config map is and make the code more readable.

ShubhamRwt: Okay got it

scholzj: @ShubhamRwt Sorry, this was supposed to be as a comment to https://github.com/strimzi/strimzi-kafka-operator/pull/5072#discussion_r646905630 and not here.

ShubhamRwt: Okay np. I am making the changes

ShubhamRwt: The custom name would be corresponding to .yml right? Since the configmap is corresponding to user

scholzj: Yes

ShubhamRwt:
@scholzj Seems like there was one test which in `TestUtils` which was causing all the unit tests to fail earlier, I changed it to the custom name and the test were running successfully.

ShubhamRwt:
> @scholzj Seems like there was one test which in `TestUtils` which was causing all the unit tests to fail earlier, I changed it to the custom name and the test were running successfully.

Also we had to use the same custom name for `metrics-cm-2` also

scholzj (approved):

ShubhamRwt:
@scholzj Hii, Are the regression test breaking due to new changes? The same errors occurred earlier also

ShubhamRwt:
Are they expecting the key to be 'metrics.json' in the System Tests?

scholzj:
It looks like they are broker by this PR. Not sure exactly why, but seems to be related.

scholzj:
@ShubhamRwt I restarted the regression since the LogCollector issues are fixed. But I suspect that it might fail again because the failures will be related to this PR and were just hidden by the other issue. We will see.

cluster-operator/src/test/java/io/strimzi/operator/cluster/TestUtils.java
```diff
@@ -40,7 +39,7 @@ public static ConfigMap getJmxMetricsCm(String data, String metricsCMName) {
                 .withNewMetadata()
                 .withName(metricsCMName)
                 .endMetadata()
-                .withData(singletonMap(AbstractModel.ANCILLARY_CM_KEY_METRICS, data))
+                .withData(singletonMap("metrics-config.yml", data))
```
Frawless, line 42: Wouldn't be better to introduce some constant for the file name?

scholzj: I guess this relateds to the `getJmxPrometheusExporterMetrics` method above it. Maybe it should be a method parameter?

scholzj:
@ShubhamRwt Still seems to be failing in MetricsST

cluster-operator/src/test/java/io/strimzi/operator/cluster/TestUtils.java
ShubhamRwt: I have defined a variable with `metrics-config.json` in the `Constant' Interface also, so if we want to use a constant instead of passing the parameter we can do that also

scholzj (approved):

scholzj:
@Frawless @sknot-rh Could you please have a look at this?

sknot-rh (approved):

systemtest/src/test/java/io/strimzi/systemtest/rollingupdate/RollingUpdateST.java
```diff
@@ -780,9 +780,9 @@ void testMetricsChange(ExtensionContext extensionContext) throws JsonProcessingE
         Object kafkaMetricsJsonToYaml = yamlReader.readValue(kafkaMetricsConf, Object.class);
         Object zkMetricsJsonToYaml = yamlReader.readValue(zkMetricsConf, Object.class);
         ObjectMapper jsonWriter = new ObjectMapper();
-        assertThat(kubeClient(NAMESPACE).getClient().configMaps().inNamespace(NAMESPACE).withName(KafkaResources.kafkaMetricsAndLogConfigMapName(clusterName)).get().getData().get("metrics-config.yml"),
+        assertThat(kubeClient(NAMESPACE).getClient().configMaps().inNamespace(NAMESPACE).withName(KafkaResources.kafkaMetricsAndLogConfigMapName(clusterName)).get().getData().get("metrics-config.json"),
```
Frawless, line 783:
```suggestion
        assertThat(kubeClient(NAMESPACE).getClient().configMaps().inNamespace(NAMESPACE).withName(KafkaResources.kafkaMetricsAndLogConfigMapName(clusterName)).get().getData().get(Constants.METRICS_CONFIG_JSON_NAME),
```

systemtest/src/test/java/io/strimzi/systemtest/rollingupdate/RollingUpdateST.java
```diff
@@ -780,9 +780,9 @@ void testMetricsChange(ExtensionContext extensionContext) throws JsonProcessingE
...
                 is(jsonWriter.writeValueAsString(kafkaMetricsJsonToYaml)));
-        assertThat(kubeClient(NAMESPACE).getClient().configMaps().inNamespace(NAMESPACE).withName(KafkaResources.zookeeperMetricsAndLogConfigMapName(clusterName)).get().getData().get("metrics-config.yml"),
+        assertThat(kubeClient(NAMESPACE).getClient().configMaps().inNamespace(NAMESPACE).withName(KafkaResources.zookeeperMetricsAndLogConfigMapName(clusterName)).get().getData().get("metrics-config.json"),
```
Frawless, line 785:
```suggestion
        assertThat(kubeClient(NAMESPACE).getClient().configMaps().inNamespace(NAMESPACE).withName(KafkaResources.zookeeperMetricsAndLogConfigMapName(clusterName)).get().getData().get(Constants.METRICS_CONFIG_JSON_NAME),
```

Frawless (approved):

systemtest/src/test/java/io/strimzi/systemtest/rollingupdate/RollingUpdateST.java
ShubhamRwt: Applied the suggestion @Frawless

Issue #3770, opened 2020-10-09
[Enhancement] Lock down Cruise Control API


kyguy:
**Is your feature request related to a problem? Please describe.**

The full Cruise Control API can be accessed directly by users by sending HTTP requests to the application on the Cruise Control pod. This allows users to preform potentially destructive operations on Kafka clusters without being properly vetted by the Cluster Operator. All non-read interactions with Cruise Control should happen through the supported `KafkaRebalance` custom resource and should not be allowed by any other method.

**Describe the solution you'd like**

Prevent users from accessing the Cruise Control API directly.

OuesFa:
Hello @kyguy 👋 
I am using cruise-control-ui on top of the cruise-control-api created by Strimzi operator. (juste to have the status of the cluster)
Could the improvement you suggest break the UI deployment?

kyguy:
After some investigation it looks like we can best accomplish this by using Cruise Control's built-in support for HTTP basic authentication and HTTPS [1]. With the basic authentication we can create two users, one with an `ADMIN` role reserved for the Cluster Operator and one with a `VIEWER` role reserved for developers. The username and password for the user with the `ADMIN` role can be secured with HTTPS requests. The username and password for the user with the `VIEWER` role can be shared in the documentation so that developers may access the read-only endpoints of the Cruise Control API.

With these changes we could also provide a cruise-control-entrance hack much like Jakub Scholz's zoo-entrance hack [2] which would create a Kubernetes service which reuses Strimzi certificates for making unencrypted requests to the Cruise Control API. This hack would allow third-party applications like the cruise-control-ui [3] read access to the Cruise Control API.

[1] https://github.com/linkedin/cruise-control/wiki/Security
[2] https://github.com/scholzj/zoo-entrance
[3] https://github.com/linkedin/cruise-control-ui

PR #5073, opened 2021-06-02
Configurable authentication, authorization, and SSL for Cruise Control API

kyguy:
### Type of change

- Enhancement / new feature

### Description
Locks down the Cruise Control API by enforcing API authentication, authorization, and SSL by default. But gives users the flexibility to disable these security configurations [1] if needed through the `spec.cruiseControl.config` section the the `Kafka` resource. e.g.
```
kind: Kafka
metadata:
  ...
spec:
  cruiseControl:
    config:
      webserver.security.enable: false  # Disables CC HTTP Basic authentication
      webserver.ssl.enable: false       # Disables CC API SSL
```

Addresses https://github.com/strimzi/strimzi-kafka-operator/issues/3770

[1] https://github.com/linkedin/cruise-control/wiki/Security
### Checklist

- [x] Write tests
- [x] Update documentation
- [x] Update CHANGELOG.md


cluster-operator/src/main/java/io/strimzi/operator/cluster/model/CruiseControl.java
```diff
@@ -173,6 +198,34 @@ protected static String defaultBootstrapServers(String cluster) {
     }
 
     @SuppressWarnings("deprecation")
+    public static String encodeToBase64(String s) {
+        return Base64.getEncoder().encodeToString(s.getBytes(StandardCharsets.US_ASCII));
+    }
```
scholzj, line 203: What is deprecated on this?

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/ModelUtils.java
```diff
@@ -129,8 +129,10 @@ static EnvVar tlsSidecarLogEnvVar(TlsSidecar tlsSidecar) {
     }
 
     public static Secret buildSecret(Reconciliation reconciliation, ClusterCa clusterCa, Secret secret, String namespace, String secretName,
-                                     String commonName, String keyCertName, Labels labels, OwnerReference ownerReference, boolean isMaintenanceTimeWindowsSatisfied) {
-        Map<String, String> data = new HashMap<>(4);
+                                     String commonName, String keyCertName, Labels labels, Map<String, String> data, OwnerReference ownerReference, boolean isMaintenanceTimeWindowsSatisfied) {
```
scholzj, line 132: This method main focus is to deal with the certificate secrets. I do not think you should mix any other data into it. You should add it in some other way or maybe even better, create additional secret instead of using the certificate secrets.

operator-common/src/test/java/io/strimzi/operator/common/operator/MockCertManager.java
```diff
@@ -156,6 +158,10 @@ public static String clientsCaCertStore() {
         return Base64.getEncoder().encodeToString(CLIENTS_CERT_STORE);
     }
 
+    public static String clientsCaCertStorePassword() {
+        return Base64.getEncoder().encodeToString(CLIENTS_CERT_STORE_PASSWORD.getBytes(Charset.defaultCharset()));
+    }
+
```
scholzj, line 164: How is the secret mocked in other tests without this method? Maybe you can use the same instead of adding this?

docker-images/kafka/cruise-control-scripts/cruise_control_config_generator.sh
```diff
@@ -43,12 +43,16 @@ cluster.configs.file=$CC_CLUSTER_CONFIG_FILE
 webserver.accesslog.path=$CC_ACCESS_LOG
 webserver.http.address=0.0.0.0
 webserver.http.cors.allowmethods=OPTIONS,GET
+webserver.ssl.keystore.location=/tmp/cruise-control/client-server.keystore.p12
```
scholzj, line 46: How does this deal with the TLS hostname verification when you use the same secret for both connecting to Kafka and as a server certificate for the webserver?

scholzj (commented):
* The file `api/src/test/resources/io/strimzi/api/kafka/model/040-Crd-kafka.yaml` should not exist anymore. You might need to rebase the PR.

docker-images/kafka/cruise-control-scripts/cruise_control_config_generator.sh
kyguy: Thanks for pointing this out. I have now added Cruise Control's public certificate to the truststore of the KafkaRebalance client so that TLS hostname verification is preformed

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/ModelUtils.java
kyguy: Following your suggestion, I have just separated the Cruise Control API credentials its own secret in the latest commit

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/CruiseControl.java
kyguy: That was a typo and has been removed

operator-common/src/test/java/io/strimzi/operator/common/operator/MockCertManager.java
kyguy: It is hard-coded in the other tests e.g. [1] [2] I figured we could refactor the other occurances in a separate PR as a lot of other reference [3]

[1] https://github.com/strimzi/strimzi-kafka-operator/blob/201540481bdf48c45056e7204ee799eaf3b7b4cf/certificate-manager/src/test/java/io/strimzi/certs/OpenSslCertManagerTest.java#L74
[2] https://github.com/strimzi/strimzi-kafka-operator/blob/76cd8065b22c8cf317d6666b271f306d54ea0bd3/cluster-operator/src/test/java/io/strimzi/operator/cluster/ResourceUtils.java#L211
[3] https://github.com/strimzi/strimzi-kafka-operator/search?q=123456

kyguy:
Currently working on the Cruise Control mock-server tests to handle TLS requests, I will ask for another round of reviews once that is fixed!

kyguy:
This PR is now ready for another round of reviews @scholzj @tomncooper @ppatierno !

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/CruiseControl.java
```diff
@@ -216,6 +271,15 @@ public static CruiseControl fromCrd(Reconciliation reconciliation, Kafka kafkaAs
             cruiseControl.setTlsSidecar(tlsSidecar);
 
             cruiseControl = cruiseControl.updateConfiguration(spec);
+            CruiseControlConfiguration c = (CruiseControlConfiguration) cruiseControl.getConfiguration();
+            Boolean sslEnabled = isApiAuthenticationEnabled(c);
+            if (sslEnabled) {
+                cruiseControl.uriScheme = HTTPS_SCHEME;
+            } else {
+                cruiseControl.uriScheme = HTTP_SCHEME;
+            }
```
scholzj, line 280: I'm a bit confused. From some parts it seems like the authentication is based on some username / password. But here you seem to use it only to enable the TLS.

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/CruiseControl.java
```diff
@@ -420,8 +493,8 @@ public Deployment generateDeployment(boolean isOpenShift, Map<String, String> an
                 .withCommand("/opt/cruise-control/cruise_control_run.sh")
                 .withEnv(getEnvVars())
                 .withPorts(getContainerPortList())
-                .withLivenessProbe(ProbeGenerator.httpProbe(livenessProbeOptions, livenessPath, REST_API_PORT_NAME))
-                .withReadinessProbe(ProbeGenerator.httpProbe(readinessProbeOptions, readinessPath, REST_API_PORT_NAME))
+                .withLivenessProbe(ProbeGenerator.httpProbe(livenessProbeOptions, livenessPath, REST_API_PORT_NAME, uriScheme, getHttpHeaders()))
+                .withReadinessProbe(ProbeGenerator.httpProbe(readinessProbeOptions, readinessPath, REST_API_PORT_NAME, uriScheme, getHttpHeaders()))
```
scholzj, line 497: This will expose the username and password in the Pod YAML. I do not think this is secure. Doesn't Cruise Control have some health check endpoint outside of authentication?

docker-images/kafka/cruise-control-scripts/cruise_control_config_generator.sh
scholzj: Maybe some better name than `client-server` would be better? Is it a client certificate? Server certificate? What client and what server? Maybe it should be called `cruise-control.keystore.p12`?

docker-images/kafka/cruise-control-scripts/cruise_control_config_generator.sh
```diff
@@ -43,12 +43,16 @@ cluster.configs.file=$CC_CLUSTER_CONFIG_FILE
...
+webserver.ssl.keystore.type=PKCS12
+webserver.ssl.key.password=$CERTS_STORE_PASSWORD
 security.protocol=SSL
 ssl.keystore.type=PKCS12
-ssl.keystore.location=/tmp/cruise-control/replication.keystore.p12
+ssl.keystore.location=/tmp/cruise-control/client-server.keystore.p12
 ssl.keystore.password=$CERTS_STORE_PASSWORD
 ssl.truststore.type=PKCS12
-ssl.truststore.location=/tmp/cruise-control/replication.truststore.p12
+ssl.truststore.location=/tmp/cruise-control/client.truststore.p12
```
scholzj, line 55: Again, the naming seems to be a bit unintuitice ... should this be called something like `kafka.truststore` or something? Or keep the old name whicih was much more clear than this.

docker-images/kafka/cruise-control-scripts/cruise_control_run.sh
```diff
@@ -14,10 +14,10 @@ mkdir -p /tmp/cruise-control
 # Import certificates into keystore and truststore
 "$CRUISE_CONTROL_HOME"/cruise_control_tls_prepare_certificates.sh
 
-export STRIMZI_TRUSTSTORE_LOCATION=/tmp/cruise-control/replication.truststore.p12
+export STRIMZI_TRUSTSTORE_LOCATION=/tmp/cruise-control/client-server.truststore.p12
```
scholzj, line 17: Is the file name correct? In other places you use `client.truststore.p12` (which as I commented above might not be the best choice, but here I wonder if the name is correct in the first place).

documentation/modules/cruise-control/ref-cruise-control-configuration.adoc
```diff
@@ -165,3 +165,42 @@ spec:
...
+kind: Kafka
+metadata:
+  name: my-cluster
+spec:
+  # ...
+  cruiseControl:
+    config:
+       webserver.ssl.enable: false
+# ...
+----
```
scholzj, line 205: This is weird. Does Cruise Control really enable HTTP Basic authentication with `.ssl.enable` option? And Authorization with `security.enable`? If it is like than then I guess we have no choice, but this seems highly misleading.

scholzj (commented):
I think the `api/src/test/resources/io/strimzi/api/kafka/model/040-Crd-kafka.yaml` file should not be there anymore.

documentation/modules/appendix_crds.adoc
```diff
@@ -1274,7 +1274,7 @@ Used in: xref:type-KafkaSpec-{context}[`KafkaSpec`]
...
+|config          1.2+<.<a|The Cruise Control configuration. For a full list of configuration options refer to https://github.com/linkedin/cruise-control/wiki/Configurations. Note that properties with the following prefixes cannot be set: bootstrap.servers, client.id, zookeeper., network., security., failed.brokers.zk.path,webserver.http., webserver.api.urlprefix, webserver.session.path, webserver.accesslog., two.step., request.reason.required,metric.reporter.sampler.bootstrap.servers, metric.reporter.topic, partition.metric.sample.store.topic, broker.metric.sample.store.topic,capacity.config.file, self.healing., anomaly.detection., ssl. (with the exception of: ssl.cipher.suites, ssl.protocol, ssl.enabled.protocols, webserver.http.cors.enabled,webserver.http.cors.origin, webserver.http.cors.exposeheaders, webserver.security.enable, webserver.ssl.enable).
 |map
```
laidan6000, line 1278: Add a space after "webserver.http.cors.enabled,".

documentation/modules/cruise-control/ref-cruise-control-configuration.adoc
```diff
@@ -165,3 +165,42 @@ spec:
           key: cruise-control-log4j.properties
     # ...
 ----
+
+[[api-security-configuration]]
+[discrete]
+== API security
```
laidan6000, line 171:
```suggestion
== Disabling API security
```

documentation/modules/cruise-control/ref-cruise-control-configuration.adoc
```diff
@@ -165,3 +165,42 @@ spec:
...
+
+The Cruise Control API is secured by default to protect the cluster against any potentially destructive Cruise Control operations.
```
laidan6000, line 173:
```suggestion
The Cruise Control REST API is secured by default to protect the cluster against potentially destructive Cruise Control operations.
```

laidan6000: Please give an example of such an operation. 

"The Cruise Control REST API is secured by default to protect the cluster against potentially destructive Cruise Control operations, such as decommissioning Kafka brokers."

documentation/modules/cruise-control/ref-cruise-control-configuration.adoc
```diff
@@ -165,3 +165,42 @@ spec:
...
+
+However, these security settings can be disabled.
```
laidan6000, line 175:
```suggestion
However, you can disable these security settings if needed.
```

documentation/modules/cruise-control/ref-cruise-control-configuration.adoc
```diff
@@ -165,3 +165,42 @@ spec:
...
+
+.An example of disabling Cruise Control API authorization.
```
laidan6000, line 177:
```suggestion
.Example Cruise Control configuration to disable API authorization.
```

documentation/modules/cruise-control/ref-cruise-control-configuration.adoc
```diff
@@ -165,3 +165,42 @@ spec:
...
+  name: my-cluster
+spec:
+  # ...
+  cruiseControl:
+    config:
+      webserver.security.enable: false
+# ...
+----
+
+.An example of disabling Cruise Control API authentication.
```
laidan6000, line 192:
```suggestion
.Example Cruise Control configuration to disable API authentication.
```

documentation/modules/cruise-control/ref-cruise-control-configuration.adoc
```diff
@@ -165,3 +165,42 @@ spec:
...
 ----
+
+[[api-security-configuration]]
+[discrete]
+== API security
+
+The Cruise Control API is secured by default to protect the cluster against any potentially destructive Cruise Control operations.
+
+However, these security settings can be disabled.
+
```
laidan6000, line 176: _Add a bulleted list:_

* To disable the built-in authorization, set `webserver.security.enable` to `false`.
* To disable the built-in authentication, set `webserver.ssl.enable` `false`.

laidan6000 (changes requested):
Looks good. I made a few more comments.

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/CruiseControl.java
kyguy: Unfortunately Cruise Control doesn't have a health check outside of the authentication but the credentials of the role exposed is meant to be public anyway. The role exposed has `USER` privileges which only has access to the GET endpoints except bootstrap and train. This role is exposed so users can get connect third-party tools for querying Cruise Control for information about their cluster

scholzj: I don't think this makes sense. Why do we build the authentication if we say that the secret credentials are not secret?

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/CruiseControl.java
kyguy: Yes, this part only to sets the scheme used by the probes without setting up the certs/keys since the kubelet skips certificate verification when the `scheme` is set to `HTTPS` [1] 

The username/password bits in the other parts is setting up HTTP Basic authentication primarily used here for role based(authorization) access to the API 

[1] https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/#http-probes

docker-images/kafka/cruise-control-scripts/cruise_control_config_generator.sh
kyguy: I like your suggestion for the keystore:  `cruise-control.keystore.p12` and  I was thinking we could use the same pattern for the truststore: `cruise-control.trustore.p12` since it is cruise control's truststore. Although I do understand the reasoning behind the original naming scheme that the truststore contains the public Kafka certs. What do you think?

docker-images/kafka/cruise-control-scripts/cruise_control_run.sh
kyguy: Thanks for the catch, this was a typo

documentation/modules/cruise-control/ref-cruise-control-configuration.adoc
kyguy: Sorry I mixed these up in the original description of the PR, it should be:
- webserver.security.enable  -> HTTP Basic authentication (which I am using for authorization)
- webserver.ssl.enable  ->  HTTPS authentication

docker-images/kafka/cruise-control-scripts/cruise_control_config_generator.sh
scholzj: `cruise-control.trustore.p12` reads as _This is the truststore used for mTLS authentication of users connecting to the Cruise Control APIs._. But it is not that, or? It is the _Truststore used to trust the Kafka brokers when connecting to them_. So if I got it right, I think something what relates it to the Kafka cluster instead of Cruise Control makes more sense.

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/CruiseControl.java
kyguy: There are two roles:
- ADMIN role which has access to all of the endpoints, these credential for this role are kept secret and used exclusively by Strimzi for Cruise Control admin operations.
- USER role which has access to the GET endpoints, these credentials are exposed for Strimzi users to use as they please
The probes only need/expose the USER role to check the health of Cruise Control

docker-images/kafka/cruise-control-scripts/cruise_control_config_generator.sh
kyguy: I was reading `cruise-control.truststore.p12` as _This is the truststore holding the certs used for identifying apps which Cruise Control trusts_. From what I understood a truststore can hold multiple certificate entries used for identifying different apps ( although in this specific case, Cruise Control only needs the certs for identifying/trusting the Kafka brokers).

scholzj: Well, I guess this _This is the truststore holding the certs used for identifying apps which Cruise Control trusts_ applies for both cases ... but at least me personally, I would by the name link it to the Cruise Control API which for me is the main CC interface. Not to the Kafka client which is inside.

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/CruiseControl.java
scholzj: Well, I don't think this is right:
* You cannot say you implement security and share the credentials publicly like this. You need to find some other way. This would if nothing else trigger a lot of questions and issues about exposed credentials etc.
* You say this is intended to be used by the users. But this is not the impression I got from the docs changes or PR description ... it does not explain how to create the credentials for the users or how to get them etc. Or did I missed it somewhere?

documentation/modules/cruise-control/ref-cruise-control-configuration.adoc
```diff
@@ -165,3 +165,45 @@ spec:
...
+[[api-security-configuration]]
+[discrete]
+== Disabling API security
+
+The Cruise Control REST API is secured by default to protect the cluster against potentially destructive Cruise Control operations, such as decommissioning Kafka brokers.
+
+However, you can disable these security settings if needed.
+
+* To disable the built-in authorization, set webserver.security.enable to false.
```
laidan6000, line 177:
```suggestion
* To disable the built-in authorization, set `webserver.security.enable` to `false`.
```

documentation/modules/cruise-control/ref-cruise-control-configuration.adoc
```diff
@@ -165,3 +165,45 @@ spec:
...
+* To disable the built-in authentication, set webserver.ssl.enable false.
```
laidan6000, line 178:
```suggestion
* To disable the built-in authentication, set `webserver.ssl.enable` to `false`.
```

documentation/modules/cruise-control/ref-cruise-control-configuration.adoc
```diff
@@ -165,3 +165,45 @@ spec:
...
+
+.Example Cruise Control configuration to disable API authorization.
```
laidan6000, line 180:
```suggestion
.Example Cruise Control configuration to disable API authorization
```

documentation/modules/cruise-control/ref-cruise-control-configuration.adoc
```diff
@@ -165,3 +165,45 @@ spec:
...
+  name: my-cluster
+spec:
+  # ...
+  cruiseControl:
+    config:
+      webserver.security.enable: false
+# ...
+----
+
+.Example Cruise Control configuration to disable API authentication.
```
laidan6000, line 195:
```suggestion
.Example Cruise Control configuration to disable API authentication
```

laidan6000 (changes requested):
A few nits

docker-images/kafka/cruise-control-scripts/cruise_control_config_generator.sh
kyguy: I understand, in that case I will switch it back to what is was originally, `replication.truststore.p12`, I want it to be as clear as possible and there is no sense in me changing the name if it is going to mislead people!

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/CruiseControl.java
kyguy: You are right. To address these concerns I have done the following:

- Updated the probes to  use healthcheck script to protect the username and password from being exposed in the Pod Yaml
-  Removed the static password for the user role. If people want direct access to the REST API whether it be for reading or writing, they will have to disable the security configurations. We can always make the API roles configurable in a later PR.

laidan6000 (approved):
New version looks good.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaAssemblyOperator.java
```diff
@@ -3419,6 +3420,21 @@ private boolean isPodUpToDate(StatefulSet sts, Pod pod) {
...
+                return secretFuture.compose(res -> {
+                    if (res == null) {
+                        return withVoid(secretOperations.reconcile(reconciliation, namespace, CruiseControl.apiSecretName(name),
+                                cruiseControl.generateApiSecret()));
+                    }
+                    return withVoid(Future.succeededFuture(this));
+                });
+
+            }
+            return withVoid(secretOperations.reconcile(reconciliation, namespace, cruiseControl.apiSecretName(name), null));
```
scholzj, line 3435: Can you use if/else? Using if/else is better because it allows to detect mistakes such as missing return etc.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceAssemblyOperator.java
```diff
@@ -1050,33 +1063,62 @@ public MapAndStatus(T loadMap, K status) {
...
-                            // cluster rebalance is new or it is in one of the others states
-                            if (kafkaRebalanceStatus == null || kafkaRebalanceStatus.getConditions().stream().filter(cond -> "ReconciliationPaused".equals(cond.getType())).findAny().isPresent()) {
-                                currentState = KafkaRebalanceState.New;
-                            } else {
-                                String rebalanceStateType = rebalanceStateConditionType(kafkaRebalanceStatus);
-                                if (rebalanceStateType == null) {
-                                    throw new RuntimeException("Unable to find KafkaRebalance State in current KafkaRebalance status");
+                    Future<Secret> ccSecretFuture = secretOperations.getAsync(clusterNamespace, ccSecretName);
+                    Future<Secret> ccApiSecretFuture = secretOperations.getAsync(clusterNamespace, ccApiSecretName);
+                    Future<Secret> coSecretFuture = secretOperations.getAsync(clusterNamespace, coSecretName);
```
scholzj, line 1072: How often does this loop run? I remember in the past that the rebalance assembly was quite aggressive in reconciling. You should make sure to get the secrets only once and not once per second (assuming this is the loop which runs in very short intervals).

cluster-operator/src/test/java/io/strimzi/operator/cluster/ResourceUtils.java
```diff
@@ -205,6 +205,16 @@ public static Kafka createKafka(String namespace, String name, int replicas,
                 .build();
     }
 
+    public static Secret createSecret(String name, String namespace, Map<String, String> data) {
```
scholzj, line 208: Don't we have this already somewhere? What about AbstractModel or something like that?

scholzj (commented):
I saw lot of updated tests. But I do not seem to see any new tests which would check what happens when the authentication is disabled etc. That should be probably covered in Unit tests, but we should also have some system tests to cover that it still works.

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/CruiseControl.java
```diff
@@ -184,7 +210,34 @@ protected static String defaultBootstrapServers(String cluster) {
...
     }
 
-    @SuppressWarnings("deprecation")
+    public static String encodeToBase64(String s) {
+        return Base64.getEncoder().encodeToString(s.getBytes(StandardCharsets.US_ASCII));
+    }
+
+    public static String getAuthCredentialsFile() {
+        return API_AUTH_CONFIG_VOLUME_MOUNT + API_AUTH_FILE_KEY;
+    }
```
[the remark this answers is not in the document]
kyguy: Just removed

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/CruiseControl.java
```diff
@@ -420,8 +478,14 @@ public Deployment generateDeployment(boolean isOpenShift, Map<String, String> an
...
-                .withLivenessProbe(ProbeGenerator.httpProbe(livenessProbeOptions, livenessPath, REST_API_PORT_NAME))
-                .withReadinessProbe(ProbeGenerator.httpProbe(readinessProbeOptions, readinessPath, REST_API_PORT_NAME))
+                .withLivenessProbe(ProbeGenerator.defaultBuilder(livenessProbeOptions)
+                        .withNewExec()
+                            .withCommand("/opt/cruise-control/cruise_control_healthcheck.sh")
+                        .endExec().build())
+                .withReadinessProbe(ProbeGenerator.defaultBuilder(readinessProbeOptions)
+                        .withNewExec()
+                            .withCommand("/opt/cruise-control/cruise_control_healthcheck.sh")
+                        .endExec().build())
```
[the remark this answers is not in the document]
kyguy: Just move occurrences to the following line

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/CruiseControl.java
```diff
@@ -528,6 +598,16 @@ public static String cruiseControlServiceAccountName(String cluster) {
         return varList;
     }
 
+    /**
+     * Generates the name of the Cruise Control secret with API authorization credentials
+     *
+     * @param kafkaCluster  Name of the Kafka Custom Resource
+     * @return  Name of the Cruise Control secret
+     */
+    public static String apiSecretName(String kafkaCluster) {
```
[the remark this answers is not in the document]
kyguy: We don't, removed

api/src/main/java/io/strimzi/api/kafka/model/CruiseControlSpec.java
```diff
@@ -34,9 +34,9 @@
     public static final String FORBIDDEN_PREFIXES = "bootstrap.servers, client.id, zookeeper., network., security., failed.brokers.zk.path,"
         + "webserver.http., webserver.api.urlprefix, webserver.session.path, webserver.accesslog., two.step., request.reason.required,"
         + "metric.reporter.sampler.bootstrap.servers, metric.reporter.topic, partition.metric.sample.store.topic, broker.metric.sample.store.topic,"
-        + "capacity.config.file, self.healing., ssl.";
-    public static final String FORBIDDEN_PREFIX_EXCEPTIONS = "ssl.cipher.suites, ssl.protocol, ssl.enabled.protocols, webserver.http.cors.enabled," +
-            "webserver.http.cors.origin, webserver.http.cors.exposeheaders";
+        + "capacity.config.file, self.healing., anomaly.detection., ssl.";
+    public static final String FORBIDDEN_PREFIX_EXCEPTIONS = "ssl.cipher.suites, ssl.protocol, ssl.enabled.protocols, webserver.http.cors.enabled, " 
+				+ "webserver.http.cors.origin, webserver.http.cors.exposeheaders, webserver.security.enable, webserver.ssl.enable";
```
scholzj, line 39: This alignment seems a bit weird. It looks different than on the lines 35-37 for example.

docker-images/kafka/cruise-control-scripts/cruise_control_healthcheck.sh
```diff
@@ -0,0 +1,22 @@
+#!/usr/bin/env bash
+set -e
+
+SCHEME="http"
+HOST="$(hostname | rev | cut -d "-" -f3- | rev):${API_PORT}"
```
scholzj, line 5: Why not just use localhost / 127.0.0.1? the healthcheck should always run locally.

scholzj (commented):
Some nits, but LGTM otherwise. Needs to be reviewed by the SMEs (@ppatierno @tomncooper) as well.

docker-images/kafka/cruise-control-scripts/cruise_control_healthcheck.sh
kyguy: The `--resolve "${HOST}:127.0.0.1"`arg to curl resolves the host address to `0.0.0.0` (I will update 0.0.0.0 to 127.0.0.1 in next commit). I had to do it this way to get the  TLS hostname verification working using localhost/127.0.0.1 since the Common Name of the CC public certificate lists the hostname of the pod

scholzj: Why don't you just do 127.0.0.1 and disable hostname verification?

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceAssemblyOperator.java
kyguy: Once every two minutes. It also gets hit once after a user updates the KafkaRebalance resource. Is this infrequent enough to leave here? Or do you still think I should refactor?

scholzj: If it is really only every 2 minutes it should be ok. But I know that at some point the KafkaRebalanceAssemblyOperator was running a separate reconciliation loop which was running every second or so. That would be too often for sure.

docker-images/kafka/cruise-control-scripts/cruise_control_healthcheck.sh
kyguy: I guess I could do this, this is what k8s HTTPS probes do anyway and it would simplify the script. From what I understand it shouldn't be much of a risk if I am using the 127.0.0.1 address

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/resource/cruisecontrol/CruiseControlApiImpl.java
```diff
@@ -24,35 +34,98 @@
...
-    public CruiseControlApiImpl(Vertx vertx) {
-        this(vertx, HTTP_DEFAULT_IDLE_TIMEOUT_SECONDS);
+    private Secret ccSecret;
+    private Secret ccApiSecret;
+    private Secret coSecret;
+    private boolean apiAuthorizationEnabled;
+    private boolean apiAuthenticationEnabled;
+    private HTTPHeader authHttpHeader;
+
+    public CruiseControlApiImpl(Vertx vertx, Secret ccSecret, Secret ccApiSecret, Secret coSecret, Boolean apiAuthorizationEnabled, boolean apiAuthenticationEnabled) {
```
tomncooper, line 44: Why not just call the constructor below with the `HTTP_DEFAULT_IDLE_TIMEOUT_SECONDS` as the `idleTimeout` argument, instead of rewriting all these assignments?

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/resource/cruisecontrol/CruiseControlApiImpl.java
```diff
@@ -24,35 +34,98 @@
...
 
+    public HttpClientOptions getHttpClientOptions() {
+
+        if (apiAuthenticationEnabled) {
+            String trustStorePassword = null;
+            File truststoreFile = null;
+
+            PasswordGenerator pg = new PasswordGenerator(12);
+            trustStorePassword = pg.generate();
+            truststoreFile = Util.createFileTrustStore(getClass().getName(), "ts", Ca.cert(ccSecret, "cruise-control.crt"), trustStorePassword.toCharArray());
```
tomncooper, line 79: This method is called for every client request. So we are recreating this trust store file and password for every single request? Is that strictly required? Can this not be done once when the client is created? Or is there some security thing I am missing?

cluster-operator/src/test/java/io/strimzi/operator/cluster/operator/resource/cruisecontrol/MockCruiseControl.java
```diff
@@ -73,6 +121,12 @@ public static ClientAndServer server(int port) throws IOException {
         return new ClientAndServer(port);
     }
 
+    public static Header getAuthHeader() {
```
tomncooper, line 124: If we are using this method in both the operator and the tests it might be safer to move it in with the other common CC stuff in the operator-common module?

docker-images/kafka/cruise-control-scripts/cruise_control_healthcheck.sh
kyguy: After adding the appropriate domains to the Cruise Control certificate, the `resolve` argument is now unnecessary. Now we have certificate verification without the extra scripts

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/resource/cruisecontrol/CruiseControlApiImpl.java
kyguy: You are right, done!

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/resource/cruisecontrol/CruiseControlApiImpl.java
kyguy: Good call, I've moved this logic to the constructor so that it is only instantiated once per Rebalance reconcile

cluster-operator/src/test/java/io/strimzi/operator/cluster/operator/resource/cruisecontrol/MockCruiseControl.java
kyguy: I was having  trouble reducing this since the CruiseControlApiImpl and  MockCruiseControl use different header objects, HTTPHeader and Header respectively. What I ended up doing was moving the core  logic to CruiseControlApiImpl and calling it from MockCruiseControl, casting HTTPHeader to Header where necessary. Does this look better?

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/ClusterCa.java
```diff
@@ -115,6 +115,38 @@ public Secret cruiseControlSecret() {
...
+            Subject.Builder subject = new Subject.Builder()
+                    .withOrganizationName("io.strimzi")
+                    .withCommonName(CruiseControl.serviceName(cluster));
+
+            subject.addDnsName(CruiseControl.serviceName(cluster));
+            subject.addDnsName(String.format("%s.%s",  CruiseControl.serviceName(cluster), namespace));
+            subject.addDnsName(ccDnsGenerator.serviceDnsNameWithoutClusterDomain());
+            subject.addDnsName(ccDnsGenerator.serviceDnsName());
+            subject.addDnsName(ccDnsGenerator.wildcardServiceDnsNameWithoutClusterDomain());
+            subject.addDnsName(ccDnsGenerator.wildcardServiceDnsName());
```
scholzj, line 134: You probably don't need these two.

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/ClusterCa.java
```diff
@@ -115,6 +115,38 @@ public Secret cruiseControlSecret() {
...
+            subject.addDnsName(CruiseControl.serviceName(cluster));
+            subject.addDnsName("localhost");
+            return subject.build();
+        };
+
+        LOGGER.debugCr(reconciliation, "{}: Reconciling zookeeper certificates", this);
```
scholzj, line 140: Wrong log message.

scholzj (approved):
LGTM. As I said before, it needs to be reviewed by the SMEs as well.

documentation/modules/cruise-control/ref-cruise-control-configuration.adoc
```diff
@@ -165,3 +165,30 @@ spec:
           key: cruise-control-log4j.properties
     # ...
 ----
+
+[[api-security-configuration]]
+[discrete]
+== Disabling API security
+
+The Cruise Control REST API is secured by default to protect the cluster against potentially destructive Cruise Control operations, such as decommissioning Kafka brokers.
```
laidan6000, line 173:
```suggestion
The Cruise Control REST API is secured with authentication and authorization to protect the cluster against potentially destructive Cruise Control operations, such as decommissioning Kafka brokers.
```

documentation/modules/cruise-control/ref-cruise-control-configuration.adoc
```diff
@@ -165,3 +165,30 @@ spec:
...
+
+However, you can disable these security settings if needed.
```
laidan6000, line 175:
```suggestion
You can disable the built-in security if necessary.
```

laidan6000, line 175:
```suggestion

```

documentation/modules/cruise-control/ref-cruise-control-configuration.adoc
```diff
@@ -165,3 +165,30 @@ spec:
           key: cruise-control-log4j.properties
     # ...
 ----
+
+[[api-security-configuration]]
+[discrete]
+== Disabling API security
+
+The Cruise Control REST API is secured by default to protect the cluster against potentially destructive Cruise Control operations, such as decommissioning Kafka brokers.
+
```
laidan6000, line 174:
```suggestion
It is recommended that Cruise Control is always run with the default authentication and authorization enabled.
However, you can disable the built-in security settings if needed.
```

documentation/modules/cruise-control/ref-cruise-control-configuration.adoc
laidan6000: Resolved.

documentation/modules/cruise-control/ref-cruise-control-configuration.adoc
```diff
@@ -165,3 +165,30 @@ spec:
           key: cruise-control-log4j.properties
     # ...
 ----
+
+[[api-security-configuration]]
+[discrete]
+== Disabling API security
```
laidan6000, line 171:
```suggestion
== Cruise Control REST API security
```

laidan6000 (changes requested):
Few more comments, thanks.

scholzj:
> Disable CC security for system tests that test API via curl

Is there some plan to add additional tests with security? If we have enabled by default, it is quite important.

kyguy:
Yes, I am currently working on it, please wait on running the system tests until I give the signal

laidan6000 (approved):
LGTM

kyguy:
@scholzj I updated the Cruise Control systemtest client logic as well as added some tests to confirm the Cruise Control REST API is reachable when the security is enabled and disabled. Let me know what you think!

This is ready for another automated system tests run!

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/CruiseControl.java
```diff
@@ -105,6 +128,8 @@
     private TlsSidecar tlsSidecar;
     private String tlsSidecarImage;
     private String minInsyncReplicas = "1";
+    private boolean authenticationEnabled;
+    private boolean authorizationEnabled;
```
ppatierno, line 132: I am not sure about the naming of these ... I mean they should just reflect what they are in Cruise Control so "security" and "ssl". AFAIU from CC doc, the `webserver.security.enable` enables the HTTP basic authentication so why using the `isApiAuthorizationEnabled` method to fill the `authorizationEnabled` field? it's about enabling the HTTP basic authentication; for `webserver.ssl.enable`, it's enabling the SSL/TLS layer on HTTP, so HTTPS but it is not related to authentication, so why using `isApiAuthenticationEnabled` to fill the corresponding `authenticationEnabled`? I would say that fields should just be `securityEnabled` and `sslEnabled` and so the corresponding methods getting the config from CC. Am I misunderstading anything here?

ppatierno (changes requested):

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/CruiseControl.java
ppatierno: Or maybe you could use `authenticationEnabled` for the `webserver.security.enable` and something like `encryptionEnabled` for `webserver.ssl.enable`

ppatierno (changes requested):

docker-images/kafka/cruise-control-scripts/cruise_control_run.sh
```diff
@@ -17,7 +17,7 @@ mkdir -p /tmp/cruise-control
 export STRIMZI_TRUSTSTORE_LOCATION=/tmp/cruise-control/replication.truststore.p12
 export STRIMZI_TRUSTSTORE_PASSWORD="$CERTS_STORE_PASSWORD"
 
-export STRIMZI_KEYSTORE_LOCATION=/tmp/cruise-control/replication.keystore.p12
+export STRIMZI_KEYSTORE_LOCATION=/tmp/cruise-control/cruise-control.keystore.p12
```
ppatierno, line 20: why this change in naming? Let me see if I got it correctly ... before this PR, this keystore was used for containing CC certificate and private key in order to have "mutual" authentication (so TLS client authentication) of CC against Kafka cluster (when it connects to the cluster for doing its stuff). 
With this PR you are adding in the same keystore, the cert and key that are used by CC for TLS encryption versus clients connecting to its API (so the cluster operator).
Would it make sense to change the name of replication.truststore.p12 in the same way? In the end it's not about replication here. Wdyt @kyguy @tomncooper @scholzj ?

documentation/modules/cruise-control/ref-cruise-control-configuration.adoc
```diff
@@ -165,3 +165,31 @@ spec:
...
+
+The Cruise Control REST API is secured with HTTP Basic authentication and SSL to protect the cluster against potentially destructive Cruise Control operations, such as decommissioning Kafka brokers.
+
+We recommend that Cruise Control in Strimzi is **only used with these settings enabled**.
+You should not disable the built-in HTTP Basic authentication or SSL settings described below.
+
+* To disable the built-in HTTP Basic authentication, set `webserver.security.enable` to `false`.
+* To disable the built-in SSL, set `webserver.ssl.enable` to `false`.
+
+.Example Cruise Control configuration to disable API authorization and authentication
```
ppatierno, line 181: "authorization and authentication" now it's more authentication/authorization and SSL

docker-images/kafka/cruise-control-scripts/cruise_control_run.sh
kyguy: Exactly right Paolo, this was the reasoning behind the name change for the keystore. As for the truststore, Jakub and I had the same discussion here [1] and decided that it was best to leave it as was. Take a look over the discussion and let us know what you think!

[1] https://github.com/strimzi/strimzi-kafka-operator/pull/5073#discussion_r693381411

docker-images/kafka/cruise-control-scripts/cruise_control_config_generator.sh
```diff
@@ -43,9 +43,13 @@ cluster.configs.file=$CC_CLUSTER_CONFIG_FILE
 webserver.accesslog.path=$CC_ACCESS_LOG
 webserver.http.address=0.0.0.0
 webserver.http.cors.allowmethods=OPTIONS,GET
+webserver.ssl.keystore.location=/tmp/cruise-control/cruise-control.keystore.p12
+webserver.ssl.keystore.password=$CERTS_STORE_PASSWORD
+webserver.ssl.keystore.type=PKCS12
+webserver.ssl.key.password=$CERTS_STORE_PASSWORD
```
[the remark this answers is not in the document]
kyguy: I have the same understanding and although we shouldn't need it because we don't encrypt keystore keys(from what I understand), Cruise Control with throw a NPE if we have don't include the `webserver.ssl.key.password` when `webserver.ssl.enable` is enabled [1]

[1] https://github.com/linkedin/cruise-control/blob/0ba9ddd44aecf8f278d9fca3d38df6d54b1ac55b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/KafkaCruiseControlApp.jka/cruisecontrol/KafkaCruiseControlApp.java#L107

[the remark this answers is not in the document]
ppatierno: From the CC code I see that it just sets the password on the Jetty `SslContextFactory`, and without specifying it we are passing `null`. From the Jetty code [1] I see that passing `null` should not hurt, it should mean just "no password on the key". Have you tried to avoid using the password? Maybe it's worth giving it a try?

[1] https://github.com/eclipse/jetty.project/blob/b56edf511ab4399122ea2c6162a4a5988870f479/jetty-util/src/main/java/org/eclipse/jetty/util/ssl/SslContextFactory.java#L816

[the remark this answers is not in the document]
kyguy: So when I tested this yesterday, I removed the whole line ` webserver.ssl.key.password=$CERTS_STORE_PASSWORD`, that's probably why CC threw the NPE from [1],  Cruise Control goes to look for the` webserver.ssl.key.password` key and throws a NPE when it's not found. Let me try testing with just `webserver.ssl.key.password=` or webserver.ssl.key.password="" in the config to see if we get the same result

[1] https://github.com/linkedin/cruise-control/blob/0ba9ddd44aecf8f278d9fca3d38df6d54b1ac55b/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/KafkaCruiseControlApp.java#L107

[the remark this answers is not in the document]
kyguy: So in the CC config, I just tested with:

- `webserver.ssl.key.password=`
- `webserver.ssl.key.password=""`

Both of which cause CC to throw the exception below. From what I understand from SO, this can happen when you use the wrong key leading me to think that the `$CERT_STORE_PASSWORD` is used/needed afterall
```
 java.security.UnrecoverableKeyException: Get Key failed: Given final block not properly padded. Such issues can arise if a bad key is used during decryption.
     at sun.security.pkcs12.PKCS12KeyStore.engineGetKey(PKCS12KeyStore.java:446) ~[?:?]
     at sun.security.util.KeyStoreDelegator.engineGetKey(KeyStoreDelegator.java:90) ~[?:?]
     at java.security.KeyStore.getKey(KeyStore.java:1057) ~[?:?]
     at sun.security.ssl.SunX509KeyManagerImpl.<init>(SunX509KeyManagerImpl.java:145) ~[?:?]
     at sun.security.ssl.KeyManagerFactoryImpl$SunX509.engineInit(KeyManagerFactoryImpl.java:70) ~[?:?]
     at javax.net.ssl.KeyManagerFactory.init(KeyManagerFactory.java:271) ~[?:?]    
     at org.eclipse.jetty.util.ssl.SslContextFactory.getKeyManagers(SslContextFactory.java:1249) ~[jetty-util-9.4.42.v20210604.jar:9.4.42.v20210604]
     at org.eclipse.jetty.util.ssl.SslContextFactory$Server.getKeyManagers(SslContextFactory.java:2363) ~[jetty-util-9.4.42.v20210604.jar:9.4.42.v20210604]
     at org.eclipse.jetty.util.ssl.SslContextFactory.load(SslContextFactory.java:373) ~[jetty-util-9.4.42.v20210604.jar:9.4.42.v20210604]
     at org.eclipse.jetty.util.ssl.SslContextFactory.doStart(SslContextFactory.java:244) ~[jetty-util-9.4.42.v20210604.jar:9.4.42.v20210604]
     at org.eclipse.jetty.util.component.AbstractLifeCycle.start(AbstractLifeCycle.java:73) ~[jetty-util-9.4.42.v20210604.jar:9.4.42.v20210604]
     at org.eclipse.jetty.util.component.ContainerLifeCycle.start(ContainerLifeCycle.java:169) ~[jetty-util-9.4.42.v20210604.jar:9.4.42.v20210604]
     at org.eclipse.jetty.util.component.ContainerLifeCycle.doStart(ContainerLifeCycle.java:117) ~[jetty-util-9.4.42.v20210604.jar:9.4.42.v20210604]
     at org.eclipse.jetty.server.SslConnectionFactory.doStart(SslConnectionFactory.java:97) ~[jetty-server-9.4.42.v20210604.jar:9.4.42.v20210604]
     at org.eclipse.jetty.util.component.AbstractLifeCycle.start(AbstractLifeCycle.java:73) ~[jetty-util-9.4.42.v20210604.jar:9.4.42.v20210604]   
     at org.eclipse.jetty.util.component.ContainerLifeCycle.start(ContainerLifeCycle.java:169) ~[jetty-util-9.4.42.v20210604.jar:9.4.42.v20210604]
     at org.eclipse.jetty.util.component.ContainerLifeCycle.doStart(ContainerLifeCycle.java:117) ~[jetty-util-9.4.42.v20210604.jar:9.4.42.v20210604]
     at org.eclipse.jetty.server.AbstractConnector.doStart(AbstractConnector.java:321) ~[jetty-server-9.4.42.v20210604.jar:9.4.42.v20210604]
     at org.eclipse.jetty.server.AbstractNetworkConnector.doStart(AbstractNetworkConnector.java:81) ~[jetty-server-9.4.42.v20210604.jar:9.4.42.v20210604]
     at org.eclipse.jetty.server.ServerConnector.doStart(ServerConnector.java:234) ~[jetty-server-9.4.42.v20210604.jar:9.4.42.v20210604]
     at org.eclipse.jetty.util.component.AbstractLifeCycle.start(AbstractLifeCycle.java:73) ~[jetty-util-9.4.42.v20210604.jar:9.4.42.v20210604]
     at org.eclipse.jetty.server.Server.doStart(Server.java:401) ~[jetty-server-9.4.42.v20210604.jar:9.4.42.v20210604]
     at org.eclipse.jetty.util.component.AbstractLifeCycle.start(AbstractLifeCycle.java:73) ~[jetty-util-9.4.42.v20210604.jar:9.4.42.v20210604]
     at com.linkedin.kafka.cruisecontrol.KafkaCruiseControlApp.start(KafkaCruiseControlApp.java:66) ~[cruise-control-2.5.57.jar:?]
     at com.linkedin.kafka.cruisecontrol.KafkaCruiseControlMain.main(KafkaCruiseControlMain.java:40) ~[cruise-control-2.5.57.jar:?]
 Caused by: javax.crypto.BadPaddingException: Given final block not properly padded. Such issues can arise if a bad key is used during decryption.
     at com.sun.crypto.provider.CipherCore.unpad(CipherCore.java:975) ~[?:?]
     at com.sun.crypto.provider.CipherCore.fillOutputBuffer(CipherCore.java:1056) ~[?:?]
     at com.sun.crypto.provider.CipherCore.doFinal(CipherCore.java:853) ~[?:?]
     at com.sun.crypto.provider.PKCS12PBECipherCore.implDoFinal(PKCS12PBECipherCore.java:408) ~[?:?]    
     at com.sun.crypto.provider.PKCS12PBECipherCore$PBEWithSHA1AndDESede.engineDoFinal(PKCS12PBECipherCore.java:440) ~[?:?]
     at javax.crypto.Cipher.doFinal(Cipher.java:2202) ~[?:?]
     at sun.security.pkcs12.PKCS12KeyStore.lambda$engineGetKey$0(PKCS12KeyStore.java:387) ~[?:?]
     at sun.security.pkcs12.PKCS12KeyStore$RetryWithZero.run(PKCS12KeyStore.java:283) ~[?:?]     
     at sun.security.pkcs12.PKCS12KeyStore.engineGetKey(PKCS12KeyStore.java:381) ~[?:?]
     ... 24 more
```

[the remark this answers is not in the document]
ppatierno: So it seems to be Jetty that sucks! We have to keep the password for now but I would open an issue and PR on CC repo to avoiding running `setKeyManagerPassword` if it's not provided in the configuration.

ppatierno (approved):

tomncooper (approved):
Really great work Kyle. My only outstanding question is why you have included the `USER` role as well as the `ADMIN` one? You don't seem (but I may have missed something) to be using it?

kyguy:
@tomncooper Nice catch! The `USER` role is only used by the probes, but the probes could have easily been set up to use the `ADMIN` role instead. I originally included the `USER` role for Strimzi users to have read access to the Cruise Control API but stripped out the public password to simplify the PR(with the intention of adding it and documentation in another PR!). Since the probes only need read access to the CC API I left them using the `USER` role

kyguy:
@scholzj when you get a chance, could you run the automated system tests again?

docker-images/kafka/cruise-control-scripts/cruise_control_config_generator.sh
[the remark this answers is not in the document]
kyguy: Opened an issue here [1], I'll follow up from there

[1] https://github.com/linkedin/cruise-control/issues/1707

PR #5084, opened 2021-06-05
Experimental ZooKeeper-less implementation

scholzj:
### Type of change

- Enhancement / new feature

### Description

Prototype / experimental implementation of ZooKeeper-less Kafka. This is not intended as future _production grade_ implementation, but might help with development and testing of different sub-components to work with Kafka after ZooKeeper removal.

It is protected by a feature gate `EscapeTheZoo` => to activate, set in the `STRIMZI_FEATURE_GATES` environment variable `+EscapeTheZoo` and create the Kafka CR without the `.spec.zookeeper` section (you should also avoid `.spec.cruiseControl` and `spec.entityOperator` unless you want to develop them to work without ZooKeeper ;-)).

### Checklist

- [ ] Write tests
- [ ] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaBrokerConfigurationBuilder.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaBrokerConfigurationBuilder.java
@@ -65,6 +65,47 @@ public class KafkaBrokerConfigurationBuilder {
     }
 
     /**
+     * Adds the node ID template. The actual node ID will be replaced in the container from the
+     * {@code STRIMZI_BROKER_ID} environment variable.
+     *
+     * @return Returns the builder instance
+     */
+    public KafkaBrokerConfigurationBuilder withNodeId()   {
+        printSectionHeader("Node ID");
+        writer.println("node.id=${STRIMZI_BROKER_ID}");
+
+        writer.println();
+
+        return this;
+    }
+
+    /**
+     * Adds the Kraft configuration.
+     *
+     * @param clusterName                   Name of the cluster (important for the advertised hostnames)
+     * @param namespace                     Namespace (important for generating the advertised hostname)
+     * @param replicas Number of replicas to configure the Kraft quorum
+     *
+     * @return Returns the builder instance
+     */
+    public KafkaBrokerConfigurationBuilder withKraft(String clusterName, String namespace, int replicas)   {
+        printSectionHeader("Kraft configuration");
+        writer.println("process.roles=broker,controller");
+        writer.println("controller.listener.names=CONTROLPLANE-9090");
+
+        List<String> quorum = new ArrayList<>(replicas);
+        for (int i = 0; i < replicas; i++)  {
+            quorum.add(i + "@" + KafkaCluster.podDnsName(namespace, clusterName, i) + ":9090");
+        }
+
+        writer.println("controller.quorum.voters=" + String.join(",", quorum));
+
+        writer.println();
+
+        return this;
+    }
+
+    /**
      * Configures the Cruise Control metric reporter. It is set only if user enabled the Cruise Control.
      *
      * @param clusterName Name of the cluster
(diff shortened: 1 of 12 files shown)


scholzj:
@tombentley FYI, as discussed earlier this week in the community meeting.

docker-images/kafka/scripts/kafka_run.sh
```diff
@@ -47,6 +43,26 @@ echo "Starting Kafka with configuration:"
 ./kafka_config_generator.sh | tee /tmp/strimzi.properties | sed -e 's/sasl.jaas.config=.*/sasl.jaas.config=[hidden]/g' -e 's/password=.*/password=[hidden]/g'
 echo ""
 
+# Prepare for Kraft
+if [ -n "$CLUSTER_ID" ]; then
```
tombentley, line 47: Maybe we should have an explicit var for enabling ZK in the scripts. Or rename it so it's `KRAFT_CLUSTER_ID`

docker-images/kafka/scripts/kafka_run.sh
```diff
@@ -47,6 +43,26 @@ echo "Starting Kafka with configuration:"
...
+  KRAFT_LOG_DIR=$(grep "log\.dirs=" /tmp/strimzi.properties | sed "s/log\.dirs=*//")
+
+  if [ ! -f "$KRAFT_LOG_DIR/meta.properties" ]; then
+    echo "Formatting Kraft storage"
+    mkdir -p "$KRAFT_LOG_DIR"
+    ./bin/kafka-storage.sh format -t "$CLUSTER_ID" -c /tmp/strimzi.properties
+  else
+    echo "Kraft storage is already formatted"
+  fi
```
tombentley, line 56: Should the trigger for this be the non-existence of `metadata.properties`. Maybe it should be done in an initcontainer?

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaAssemblyOperator.java
```diff
@@ -366,6 +376,73 @@ public KafkaAssemblyOperator(Vertx vertx, PlatformFeaturesAvailability pfa,
         return chainPromise.future();
     }
 
+    Future<Void> reconcileWithoutZoo(ReconciliationState reconcileState)  {
```
tombentley, line 379: We're going to have to live with both modes for a while, during which there will be continued churn to the reconciliation process. So I wonder if it would be better to pass the mode flag and branch within the compose steps.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaAssemblyOperator.java
```diff
@@ -255,6 +257,14 @@ public KafkaAssemblyOperator(Vertx vertx, PlatformFeaturesAvailability pfa,
     }
 
     Future<Void> reconcile(ReconciliationState reconcileState)  {
+        if (featureGates.escapeTheZooEnabled()) {
```
tombentley, line 260: `kraftMode` is probably a better name.

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaBrokerConfigurationBuilder.java
```diff
@@ -157,23 +198,29 @@ public KafkaBrokerConfigurationBuilder withZookeeper(String clusterName)  {
...
+        if (!escapeTheZoo) {
+            advertisedListeners.add(String.format("CONTROLPLANE-9090://%s:9090",
+                    // Pod name constructed to be templatable for each individual ordinal
+                    DnsNameGenerator.podDnsNameWithoutClusterDomain(namespace, KafkaResources.brokersServiceName(clusterName),
+                            KafkaResources.kafkaStatefulSetName(clusterName) + "-${STRIMZI_BROKER_ID}")
+            ));
+            securityProtocol.add("CONTROLPLANE-9090:SSL");
+            configureControlPlaneListener();
+        } else {
+            securityProtocol.add("CONTROLPLANE-9090:PLAINTEXT");
```
tombentley, line 222: I guess for each of these places where we're disabling functionality we should raise an issue so we can tackle it bit by bit and not lose track of temporarily missing features in kraft mode. Adding the issue number in the code is probably also worthwhile in this case.

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaCluster.java
```diff
@@ -1969,27 +1969,40 @@ public NodeAddressType getPreferredNodeAddressType(GenericKafkaListener listener
...
-                .withListeners(cluster, namespace, listeners, controlPlaneListener)
-                .withAuthorization(cluster, authorization)
-                .withCruiseControl(cluster, cruiseControlSpec, ccNumPartitions, ccReplicationFactor, ccMinInSyncReplicas)
-                .withUserConfiguration(configuration)
-                .build().trim();
+    private String generateBrokerConfiguration(boolean controlPlaneListener, boolean escapeTheZoo)   {
+        if (escapeTheZoo)   {
+            return new KafkaBrokerConfigurationBuilder()
+                    .withNodeId()
+                    .withKraft(cluster, namespace, replicas)
```
tombentley, line 1976: So currently every broker must also be a voter, right?

scholzj: Yes. Every broker is controller.

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaBrokerConfigurationBuilder.java
scholzj: I'm not sure this is something what should evolve into production implementation - so I'm not sure it makes sense to _tackle it bit by bit_. I guess having some comments pointing that out might be good.

scholzj:
@tombentley I guess the main question to answer is whether we actually want to have this merged which to be honest I'm not sure my self.

scholzj:
Closing as discussed on the Strimzi Community Call. If anyone wants to experiment with this, I will keep the branch so that it can be checked out and built.

Issue #4769, opened 2021-04-16
Unable to renew Custom Cluster CA Certificates


tmitchell10:
**Describe the bug**
I am attempting to renew the CA certificates in my Strimzi cluster and when I follow the instructions and replace the Cluster CA certificates, no pods roll, but the zookeeper pods issue error messages in the logs:

2021-04-16 15:07:59,715 WARN Exception caught (org.apache.zookeeper.server.NettyServerCnxnFactory) [nioEventLoopGroup-7-1]
io.netty.handler.codec.DecoderException: javax.net.ssl.SSLHandshakeException: Received fatal alert: certificate_unknown
	at io.netty.handler.codec.ByteToMessageDecoder.callDecode(ByteToMessageDecoder.java:471)
	at io.netty.handler.codec.ByteToMessageDecoder.channelRead(ByteToMessageDecoder.java:276)
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:379)
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:365)
	at io.netty.channel.AbstractChannelHandlerContext.fireChannelRead(AbstractChannelHandlerContext.java:357)
	at io.netty.channel.DefaultChannelPipeline$HeadContext.channelRead(DefaultChannelPipeline.java:1410)
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:379)
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:365)
	at io.netty.channel.DefaultChannelPipeline.fireChannelRead(DefaultChannelPipeline.java:919)
	at io.netty.channel.nio.AbstractNioByteChannel$NioByteUnsafe.read(AbstractNioByteChannel.java:163)
	at io.netty.channel.nio.NioEventLoop.processSelectedKey(NioEventLoop.java:714)
	at io.netty.channel.nio.NioEventLoop.processSelectedKeysOptimized(NioEventLoop.java:650)
	at io.netty.channel.nio.NioEventLoop.processSelectedKeys(NioEventLoop.java:576)
	at io.netty.channel.nio.NioEventLoop.run(NioEventLoop.java:493)
	at io.netty.util.concurrent.SingleThreadEventExecutor$4.run(SingleThreadEventExecutor.java:989)
[...]

PR #5103, opened 2021-06-09
Make certificate preparation in TLS sidecars less sensitive on new-lines

scholzj:
### Type of change

- Bugfix

### Description

When the CA secret contains multiple public keys, we have to merge them into a single file in the TLS sidecars for CC and EO. But the current way of merging them is sensitive to whether the do or do not contain a new-line at the end. When they do not have the new-line, they end up like this:

```
-----BEGIN CERTIFICATE-----
...
-----END CERTIFICATE----------BEGIN CERTIFICATE-----
...
-----END CERTIFICATE-----
```

And `stunnel` cannot read them and fails.

This PR improves the way the certificates are merged to make sure it works regardless the new line. It first removes the last new-line if there is any (to avoid double new-lines) and then adds another new-line. This way the file should always look the same.

This issue was originally raised as part of #4769.

### Checklist

### DIFF
--- a/docker-images/kafka/stunnel-scripts/cruise_control_stunnel_config_generator.sh
+++ b/docker-images/kafka/stunnel-scripts/cruise_control_stunnel_config_generator.sh
@@ -5,7 +5,10 @@ set -e
 CC_CERTS_KEYS=/etc/tls-sidecar/cc-certs
 # Combine all the certs in the cluster CA into one file
 CA_CERTS=/tmp/cluster-ca.crt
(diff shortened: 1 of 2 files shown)


tombentley (approved):

ppatierno (approved):

chrispatmore:
I'm seeing another problem
https://github.com/strimzi/strimzi-kafka-operator/issues/4769#issuecomment-857803971
That would be good to see if it can be fixed here as well

scholzj:
@chrispatmore I merged it since it fixes at leats something. If we find some fix for the next problem we can open a new PR.

PR #5104, opened 2021-06-09
helm chart: add securitycontext for operator

mblaschke:
### Type of change

- Bugfix
x Enhancement / new feature
- Refactoring
- Documentation

### Description

add pod and container `securityContext` for operator into helm chart

### Checklist

- [ ] Write tests
- [ ] Update documentation
- [ ] Update CHANGELOG.md


scholzj (commented):
Two comments ...
* You need to do these changes in the `packaging/helm` folder which is the _development_ Helm Chart.
* I'm wondering if _no `securityContext`_ has the same semantics as _empty `securityContext`_. For example in different components injecting customised security context etc. Maybe it would be more safe to have it whole wrapped in some `if`? That way we will know for sure it will not break anything.

scholzj (commented):
Looks good. Would you mind adding the two options also to https://github.com/strimzi/strimzi-kafka-operator/blob/main/packaging/helm-charts/helm3/strimzi-kafka-operator/README.md? Thanks

scholzj (approved):

Frawless (approved):

PR #5083, opened 2021-06-04
Feature/lables exclude in value

vl4deee11:
### Type of change
- Enhancement / new feature

### Description

For fluxcd it is useful to exclude labels in values.yaml.

It can be useful with problem https://github.com/strimzi/strimzi-kafka-operator/issues/2558

### Checklist

_Please go through this checklist and make sure all applicable tasks have been done_

- [ ] Write tests
- [ ] Make sure all tests pass
- [x] Update documentation
- [ ] Check RBAC rights for Kubernetes / OpenShift roles
- [ ] Try your changes from Pod inside your Kubernetes and OpenShift cluster, not just locally
- [ ] Reference relevant issue(s) and close them after merging
- [ ] Update CHANGELOG.md
- [ ] Supply screenshots for visual changes, such as Grafana dashboards

PR #5106, opened 2021-06-09
Fix wrongly committed Helm Chart improvement

scholzj:
### Type of change

- Bugfix

### Description

In #5083 we screwed up during the review process and committed the changes to the released Helm Chart in `helm-chart`. This PR tries to fixit. It:
* Reverts the changes in the `/helm-charts`
* Adds the same changes to `packaging/helm-charts`

### DIFF
--- a/helm-charts/helm3/strimzi-kafka-operator/README.md
+++ b/helm-charts/helm3/strimzi-kafka-operator/README.md
@@ -169,7 +169,6 @@ the documentation for more details.
 | `readinessProbe.periodSeconds`       | Readiness probe period in seconds         | 30                                                   |
 | `imageRegistryOverride`              | Override all image registry config        | `nil`                                                |
 | `imageRepositoryOverride`            | Override all image repository config      | `nil`                                                |
-| `labelsExclusionPattern`             | Override the exclude pattern for exclude some labels             | `""`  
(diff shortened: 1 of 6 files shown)


samuel-hawker:
This might sound heavy-handed, but might it be worth doing something like a hash on the `helm-charts` directory and then storing that in source, that way, if someone changes the wrong helm-chart, we could fail builds if they don't change the hash, and if they do, we'd see it in the code changes and be like 'im not sure we should be changing this'

https://stackoverflow.com/questions/545387/linux-compute-a-single-hash-for-a-given-folder-contents
Suggested methods of doing this would be here, but essentially I think a unique ID of our public facing files we don't want changed often might be a good idea.
I think checking for a hash change is preferable to checking file names, but maybe I am being overly zealous in finding a technical solution to a small problem.

scholzj:
I'm not sure you need to store the hash somewhere to check this. You can maybe just diff it against the master branch? But in any case, someone would need to implement it (have time for it).

scholzj:
> If you and others agree this would have some value I'd happily raise an issue and look into doing this, assumedly it would be best integrated into the build and test azure pipeline?

Yeah, sure ... go for it. The Build and Test pipeline is best I think.

PR #5107, opened 2021-06-09
feat: shasum published charts

samuel-hawker:
Proposed solution that was suggested in
https://github.com/strimzi/strimzi-kafka-operator/pull/5106#issuecomment-857923126
where some changes were checked in and passed review,
but for the wrong chart
this check should make this a lot harder to happen

Signed-off-by: Samuel Hawker <sam.b.hawker@gmail.com>

### Type of change

_Select the type of your PR_

- Bugfix
- Enhancement / new feature
- Refactoring
- Documentation

### Description

### Checklist

- [ ] Write tests
- [ ] Update documentation
- [ ] Update CHANGELOG.md


samuel-hawker:
Raising as a draft first, verifying that it works as intended on azure, and that output is useful in all situations. Will remove the second, dummy commit after.

scholzj (commented):
Some thoughts to consider:
* We need this also for `install` and `examples`
* Should we try to run it as early as possible in `make all`?
* Does the Azure build actually run `make all`? I don't think so. I think you should run the shell script directly from `build.sh`.
* I wonder if we should have the checksums in some separate files instead of directly int he script.

samuel-hawker:
Good points I will action on these now.

.azure/scripts/build.sh
```diff
@@ -57,6 +57,8 @@ if [ -n "$CHANGED_DERIVED" ] || [ -n "$GENERATED_FILES" ] ; then
     exit 1
 fi
 
+source .azure/scripts/publish_check.sh
```
samuel-hawker, line 60: I _think_  I should move this to before the mvn step.

scholzj: Yeah, that way it should fail faster

Makefile
```diff
@@ -16,7 +16,7 @@ endif
 SUBDIRS=kafka-agent mirror-maker-agent tracing-agent crd-annotations test crd-generator api api-conversion mockkube certificate-manager operator-common config-model config-model-generator cluster-operator topic-operator user-operator kafka-init docker-images packaging/helm-charts/helm3 packaging/install packaging/examples
 DOCKER_TARGETS=docker_build docker_push docker_tag
 
-all: prerequisites_check $(SUBDIRS) crd_install helm_install shellcheck docu_versions docu_check
+all: prerequisites_check publish_check $(SUBDIRS) crd_install helm_install shellcheck docu_versions docu_check
```
scholzj, line 19: Out of curiosity ... why do you call it publish_check?

samuel-hawker: This was more of a placeholder, I wasn't 100% sure what to call it, i put publish as these are like our 'published' materials, maybe `release_files_check` might be better?
Is there a specific word for what you call the public facing, release versioned files for helm-charts/install/examples?

scholzj: I think `release_files_check` sounds good to me. I think we can use it. _release files_ is what I would normally call it I guess.

scholzj:
@samuel-hawker Any plans to get back to this and finish it? There were several PRs where we missed commits into these directories during the review. So it would be super useful to have this in the CI.

samuel-hawker:
> @samuel-hawker Any plans to get back to this and finish it? There were several PRs where we missed commits into these directories during the review. So it would be super useful to have this in the CI.

@scholzj apologies for the delays on this, haven't had much free time.
As you can probably tell from the commit history, I hit a blocker of sorts:
The command I use to generate the shas are:
```
find ./helm-charts/ -type f -print0 | sort -z | xargs -0 $SHA1SUM | $SHA1SUM
```
This on a mac, always returns a different sha from what the server generates, even when I substituted:
 find -> gfind, sort -> gsort, xargs -> gxargs
 
 Looking into this it seems that things like file owner, user etc are used as part of the hash, so different systems will return different shas, however the azure pipeline does consistently return the same sha.
 
 I think the pragmatic solution here, for now, is to not run the checks locally, but only on PR, and it checks the checked in shas against what the azure pipeline generates.
 This does come with a drawback:
 When we want to update these files, a PR will have a two stage commit process,
 1 - checkin the 'release' files, i.e. helm-charts/*
 2 - have an azure job fail, and spit out the new updated shas, copy these into the corresponding .checksums file and then push up the second commit
 
 This would be the process as of now, and it means we get the value and a followup PR could look into aligning local generates shasums with what the CI/CD server generates.
 I looked into using tar, but again similar issues, plus apparently differenty systems use different headers in the tar files.
 I don't know if you or anyone you know might be knowledgable on different OS bash commands, that could solve the above problem.
 
 If you confirm that the above proposed plan sounds good I can cleanup this PR (removing it from the make all) and only call it in the pipeline.

scholzj:
Hmm, that sucks. But it might be better then nothing 🤷🏻‍♂️. Have you tried some tricks like doing a `tar` of the directory and trying to hash that? Other option which comes to my mind is to run a Docker container which mounts the data as a volume and hash it there ... but that would be quite obscure solution to a simple problem :-D ... that is probably not worth it.

samuel-hawker:
> Hmm, that sucks. But it might be better then nothing 🤷🏻‍♂️. Have you tried some tricks like doing a `tar` of the directory and trying to hash that? Other option which comes to my mind is to run a Docker container which mounts the data as a volume and hash it there ... but that would be quite obscure solution to a simple problem :-D ... that is probably not worth it.

The docker mounting solution is a good idea...
I will put this on my TODOs and try to get this done asap.
I did start to try to do it with Tar but many stackoverflow posts were claiming this was less portable than my current method, but it's worth a try as well.

samuel-hawker:
@scholzj unfortunately I could not get it working with a docker container either, the shas were continuing to yield different shas, even after a chown/chmod
Similar story for tar, for now I think we should opt for using this more simple approach and improve it over time.

scholzj:
@samuel-hawker Ok, lets have at least something.

samuel-hawker:
@strimzi/maintainers can I get a review on this when possible?
I will try and improve this in a followup, as mentioned previously, all  attempts at standardizing shas between Azure and local builds have unfortunately  failed, even when running from inside a container.

scholzj (approved):

scholzj:
@strimzi/maintainers Anyone else wants to have a look? Or can we merge it?

Frawless (approved):

Issue #5088, opened 2021-06-07
Adding custom resource selector tag on Strimzi operator metrics


ppatierno:
With the introduction of the `STRIMZI_CUSTOM_RESOURCE_SELECTOR` env var for selecting which Kafka customer resources the operator should reconcile via labels, it would be useful to have that information in the metrics exposed by the Strimzi operator.
For example, currently we have something like:

```shell
strimzi_reconciliations_total{kind="Kafka",} 13.0
```
but could be useful to have distinction across custom resource selector labels, so something like:

```shell
strimzi_reconciliations_total{kind="Kafka", selector="io.ppatierno/my-label-to-select=my-value"} 13.0
strimzi_reconciliations_total{kind="Kafka", selector="io.ppatierno/my-label-to-select-2=my-value-2"} 5.0
```

The same could be applied to all other Strimzi operator metrics already grouped by "kind" tag.

ShubhamRwt:
@ppatierno Hello, Is this issue free to work upon?

ppatierno:
@ShubhamRwt sure, feel free to work on it if you like ;-)

PR #5114, opened 2021-06-10
Adding custom resource selector tag on Strimzi operator metrics

ShubhamRwt:
### Type of change

- Enhancement / new feature

### Description

This PR is for issue #5088, which is adding the `selector` label to the the tags of the metrics.

### Checklist

- [ ] Write tests
- [ ] Update documentation
- [ ] Update CHANGELOG.md


cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/AbstractConnectOperator.java
```diff
@@ -144,10 +144,15 @@ public AbstractConnectOperator(Vertx vertx, PlatformFeaturesAvailability pfa, St
...
         // Setup metrics for connectors
-        Tags metricTags = Tags.of(Tag.of("kind", KafkaConnector.RESOURCE_KIND));
 
+        Tags metricTags;
+
+        if (config.getCustomResourceSelector() != null) {
+            metricTags = Tags.of(Tag.of("kind", KafkaConnector.RESOURCE_KIND), Tag.of("selector", config.getCustomResourceSelector().toSelectorString()));
+        } else {
+            metricTags = Tags.of(Tag.of("kind", KafkaConnector.RESOURCE_KIND));
+        }
```
scholzj, line 155: Does the Connector operator really follow the selector labels? Doesn't it follow only the connect labels? If that is the case, I'm not sure this should apply to the metric.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceAssemblyOperator.java
```diff
@@ -156,7 +156,7 @@
      */
     public KafkaRebalanceAssemblyOperator(Vertx vertx, PlatformFeaturesAvailability pfa,
                                           ResourceOperatorSupplier supplier, ClusterOperatorConfig config) {
-        super(vertx, KafkaRebalance.RESOURCE_KIND, supplier.kafkaRebalanceOperator, supplier.metricsProvider, null);
+        super(vertx, KafkaRebalance.RESOURCE_KIND, supplier.kafkaRebalanceOperator, supplier.metricsProvider, config.getCustomResourceSelector());
```
scholzj, line 159: Does the Rebalance operator really follow the selector labels? Doesn't it follow only the cluster labels? If that is the case, I'm not sure this should apply to the metric.

operator-common/src/test/java/io/strimzi/operator/common/OperatorMetricsTest.java
```diff
@@ -68,7 +70,7 @@ public void testSuccessfulReconcile(VertxTestContext context)  {
 
         AbstractWatchableStatusedResourceOperator resourceOperator = resourceOperatorWithExistingResource();
 
-        AbstractOperator operator = new AbstractOperator(vertx, "TestResource", resourceOperator, metrics, null) {
+        AbstractOperator operator = new AbstractOperator(vertx, "TestResource", resourceOperator, metrics, Labels.fromMap(emptyMap())) {
```
scholzj, line 73: You updated the tests. But I think you should also add some new tests to cover the new feature you are adding.

scholzj (commented):
* Please also reference the issue Paolo opened for this.

operator-common/src/test/java/io/strimzi/operator/common/OperatorMetricsTest.java
ShubhamRwt: Yeah, that was something I wanted to ask in the morning

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/AbstractConnectOperator.java
ShubhamRwt: I ran the tests

ppatierno:
A more general comment I have ...

AFAIU you are adding the selector got from the overall configuration on the cluster operator. The tag should be more related to the label applied on the custom resource that is part of the overall selector. I mean ...
The selector configured on the cluster operator via the env var could be something like:

`labelKey1=value1, labelKey2=value2`

It could happen that some Kafka CRs have `labelKey1=value1` in its labels and other Kafka CRs has `labelKey2=value2` so they are reconciled by the cluster operator but what I want to see is metrics like:

```
strimzi_reconciliations_successful_total{kind="Kafka", selector="labelKey1=value1"} 2.0
strimzi_reconciliations_successful_total{kind="Kafka", selector="labelKey2=value2"} 5.0
```

and not

```shell
strimzi_reconciliations_successful_total{kind="Kafka", selector="labelKey1=value1,labelKey1=value1"} 7.0
```

does it make sense?

ShubhamRwt:
> A more general comment I have ...
> 
> AFAIU you are adding the selector got from the overall configuration on the cluster operator. The tag should be more related to the label applied on the custom resource that is part of the overall selector. I mean ...
> The selector configured on the cluster operator via the env var could be something like:
> 
> `labelKey1=value1, labelKey2=value2`
> 
> It could happen that some Kafka CRs have `labelKey1=value1` in its labels and other Kafka CRs has `labelKey2=value2` so they are reconciled by the cluster operator but what I want to see is metrics like:
> 
> ```
> strimzi_reconciliations_successful_total{kind="Kafka", selector="labelKey1=value1"} 2.0
> strimzi_reconciliations_successful_total{kind="Kafka", selector="labelKey2=value2"} 5.0
> ```
> 
> and not
> 
> ```shell
> strimzi_reconciliations_successful_total{kind="Kafka", selector="labelKey1=value1,labelKey1=value1"} 7.0
> ```
> 
> does it make sense?
> 
> Of course, if a selector is not configured, the cluster operator will just reconcile every Kafka CR and the output could be just what we have today:
> 
> ```shell
> strimzi_reconciliations_successful_total{kind="Kafka"} 7.0
> ```
Thanks for the brief explanation @ppatierno. Now I have got the gist of what you expecting, I will start working on the fixes for this

scholzj:
@ppatierno I do not get your point. IIRC the selector is not `OR` but `AND`. If you specify multiple values, you will reconcile only resources which match the whole selector and just some parts of it.

ppatierno:
@scholzj if that's an AND then you are right, I thought we implemented that as an OR.

scholzj:
> @scholzj if that's an AND then you are right, I thought we implemented that as an OR.

It is some time since I implemented it. But we do not have multiple watches, we do not list the resources multiple times etc. And there is no way you could treat these as `OR` in a single watch or single list query as far as I know. So I'm pretty sure these are treated as `AND` (although not 100% sure I guess).

operator-common/src/main/java/io/strimzi/operator/common/AbstractOperator.java
```diff
@@ -102,8 +102,13 @@ public AbstractOperator(Vertx vertx, String kind, O resourceOperator, MetricsPro
...
 
         // Setup metrics
-        Tags metricTags = Tags.of(Tag.of("kind", kind()));
+        Tags metricTags;
 
+        if (selectorLabels != null) {
+            metricTags = Tags.of(Tag.of("kind", kind()), Tag.of("selector", selectorLabels.toSelectorString()));
+        } else {
+            metricTags = Tags.of(Tag.of("kind", kind()), Tag.of("selector", ""));
+        }
```
ppatierno, line 111: what about?

```java
metricTags = Tags.of(Tag.of("kind", kind()), selectorLabels != null ? Tag.of("selector", selectorLabels.toSelectorString()) : Tag.of("selector", ""));
```
of for clarity
```java
String selectorValue = selectorLabels != null ? selectorLabels.toSelectorString() : "";
metricTags = Tags.of(Tag.of("kind", kind()), Tag.of("selector", selectorValue));
```

or have you tried what happens when you just put `null` as tag value instead of empty string?

ShubhamRwt: Yeah I tried it with null but it gives us error and cause the operator to fail,  since its a `not null` property. So the above workaround will work fine but we cant go with the null one

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/AbstractConnectOperator.java
```diff
@@ -144,9 +144,9 @@ public AbstractConnectOperator(Vertx vertx, PlatformFeaturesAvailability pfa, St
         this.operatorNamespaceLabels = config.getOperatorNamespaceLabels();
         this.pfa = pfa;
         this.port = port;
-
         // Setup metrics for connectors
-        Tags metricTags = Tags.of(Tag.of("kind", KafkaConnector.RESOURCE_KIND));
+
```
scholzj, line 148: A nit, but is the empty line really needed here?

operator-common/src/main/java/io/strimzi/operator/common/AbstractOperator.java
```diff
@@ -462,6 +467,10 @@ protected void addWarningsToStatus(Status status, Set<Condition> unknownAndDepre
         return selector;
     }
 
+    public Map<String, String> getMatchLables() {
+        return selector.get().getMatchLabels();
+    }
```
scholzj, line 472: Is this really needed? It does not seem to be used in any production code.

operator-common/src/test/java/io/strimzi/operator/common/OperatorMetricsTest.java
```diff
@@ -113,6 +116,61 @@ protected Status createStatus() {
                 })));
     }
 
+    @Test
+    public void testSuccessfulReconcileWithSelectorLabel(VertxTestContext context)  {
+        MetricsProvider metrics = createCleanMetricsProvider();
+
+        AbstractWatchableStatusedResourceOperator resourceOperator = resourceOperatorWithExistingResource();
+
+        Labels nsLabel = Labels.fromMap(Collections.singletonMap("nsLabelKey", "nsLabelValue"));
```
scholzj, line 125: `nsLabels` is something I would read s namespace ... but this looks like a resource labels? Should we give it some better name? Same for the values I guess?

operator-common/src/test/java/io/strimzi/operator/common/OperatorMetricsTest.java
```diff
@@ -113,6 +116,61 @@ protected Status createStatus() {
...
+                .onComplete(context.succeeding(v -> context.verify(() -> {
+                    MeterRegistry registry = metrics.meterRegistry();
+
+                    assertThat(registry.get(AbstractOperator.METRICS_PREFIX + "reconciliations").tag("kind", "TestResource").counter().count(), is(1.0));
+                    assertThat(registry.get(AbstractOperator.METRICS_PREFIX + "reconciliations.successful").tag("kind", "TestResource").counter().count(), is(1.0));
+                    assertThat(registry.get(AbstractOperator.METRICS_PREFIX + "reconciliations.failed").tag("kind", "TestResource").counter().count(), is(0.0));
+                    assertThat(registry.get(AbstractOperator.METRICS_PREFIX + "reconciliations.locked").tag("kind", "TestResource").counter().count(), is(0.0));
+
+                    assertThat(registry.get(AbstractOperator.METRICS_PREFIX + "reconciliations.duration").tag("kind", "TestResource").timer().count(), is(1L));
+                    assertThat(registry.get(AbstractOperator.METRICS_PREFIX + "reconciliations.duration").tag("kind", "TestResource").timer().totalTime(TimeUnit.MILLISECONDS), greaterThan(0.0));
```
scholzj, line 162: You seem to test the counters. But I do not see the labels used anywhere here. Should it be asserted somewhere as well?

operator-common/src/test/java/io/strimzi/operator/common/OperatorMetricsTest.java
```diff
@@ -166,13 +224,68 @@ protected Status createStatus() {
...
+                return null;
+            }
+
+            @Override
+            protected Status createStatus() {
+                return new Status() { };
+            }
+        };
+
+//        assertThat(operator.getMatchLables().get("nsLabelKey").toString(), is("nsLabelValue"));
```
scholzj, line 255: Should this be deleted?

operator-common/src/test/java/io/strimzi/operator/common/OperatorMetricsTest.java
```diff
@@ -166,13 +224,68 @@ protected Status createStatus() {
...
+                .onComplete(context.failing(v -> context.verify(() -> {
+                    MeterRegistry registry = metrics.meterRegistry();
+
+                    assertThat(registry.get(AbstractOperator.METRICS_PREFIX + "reconciliations").tag("kind", "TestResource").counter().count(), is(1.0));
+                    assertThat(registry.get(AbstractOperator.METRICS_PREFIX + "reconciliations.successful").tag("kind", "TestResource").counter().count(), is(0.0));
+                    assertThat(registry.get(AbstractOperator.METRICS_PREFIX + "reconciliations.failed").tag("kind", "TestResource").counter().count(), is(1.0));
+                    assertThat(registry.get(AbstractOperator.METRICS_PREFIX + "reconciliations.locked").tag("kind", "TestResource").counter().count(), is(0.0));
+
+                    assertThat(registry.get(AbstractOperator.METRICS_PREFIX + "reconciliations.duration").tag("kind", "TestResource").timer().count(), is(1L));
+                    assertThat(registry.get(AbstractOperator.METRICS_PREFIX + "reconciliations.duration").tag("kind", "TestResource").timer().totalTime(TimeUnit.MILLISECONDS), greaterThan(0.0));
```
scholzj, line 268: Again, might be nice to assert that the tag is set properly (as empty string).

operator-common/src/test/java/io/strimzi/operator/common/OperatorMetricsTest.java
```diff
@@ -166,13 +224,68 @@ protected Status createStatus() {
                 })));
     }
 
+    @Test
+    public void testFailingReconcileWithSelectorLabel(VertxTestContext context)  {
```
scholzj, line 228: Should this be `testFailingReconcileWithoutSelectorLabel`? I do not see any label. Same in the further tests.

scholzj (changes requested):
The code looks mostly good apart from the nits. But I think the tests need some more work on. I left some comments. They also seem to create lot of duplicate functionality. I wonder if you can have some method with _selector labels_ parameter and use it to have the code only once but test different values of labels ... null, empty map, some labels etc.

operator-common/src/main/java/io/strimzi/operator/common/AbstractOperator.java
ShubhamRwt: Actually I created this to use in the test. In the Operator metrics test, when Iam asserting to check if the labels is correct, I need this method

scholzj: I do not think you need this. In the tests:
* You set the selector, so you need its value
* In this PR, you do not care about the selector value really. You care about the selector tag on the metric. That is what you need to test.

In case it would really be needed, it should probably:
* Not be public
* Have a Javadoc explaining it why and where it is needed

operator-common/src/main/java/io/strimzi/operator/common/AbstractOperator.java
```diff
@@ -461,7 +462,6 @@ protected void addWarningsToStatus(Status status, Set<Condition> unknownAndDepre
     public Optional<LabelSelector> selector() {
         return selector;
     }
-
```
scholzj, line 464: I think we want to keep this empty line between methods?

operator-common/src/main/java/io/strimzi/operator/common/model/Labels.java
```diff
@@ -5,7 +5,6 @@
 package io.strimzi.operator.common.model;
 
 import io.fabric8.kubernetes.api.model.HasMetadata;
-
```
scholzj, line 8: Do we need this change?

operator-common/src/test/java/io/strimzi/operator/common/OperatorMetricsTest.java
```diff
@@ -48,6 +52,7 @@
 public class OperatorMetricsTest {
     private static Vertx vertx;
 
+
```
scholzj, line 55: Do we need the double empty line?

operator-common/src/test/java/io/strimzi/operator/common/OperatorMetricsTest.java
```diff
@@ -62,13 +67,13 @@ public static void after() {
         vertx.close();
     }
 
-    @Test
-    public void testSuccessfulReconcile(VertxTestContext context)  {
+
+    public void testSuccessfulReconcile(VertxTestContext context, Labels selectorLabels)  {
```
scholzj, line 71: If it is not a test anymore, it might be best to rename it to not start with `test`. Also, the double empty line above is probably not needed?

operator-common/src/test/java/io/strimzi/operator/common/OperatorMetricsTest.java
```diff
@@ -95,11 +100,12 @@ protected Status createStatus() {
                 .onComplete(context.succeeding(v -> context.verify(() -> {
                     MeterRegistry registry = metrics.meterRegistry();
 
+                    assertThat(registry.getMeters().get(0).getId().getTags().size(), is(2));
+                    assertThat(registry.getMeters().get(0).getId().getTags().get(1), is(Tag.of("selector", selectorLabels != null ? selectorLabels.toSelectorString() : "")));
```
scholzj, line 104: This tests the label only for the first metric? What metric is that? Can't you get it by name? That would probably be more stable.

operator-common/src/test/java/io/strimzi/operator/common/OperatorMetricsTest.java
```diff
@@ -147,6 +168,8 @@ protected Status createStatus() {
                 .onComplete(context.failing(v -> context.verify(() -> {
                     MeterRegistry registry = metrics.meterRegistry();
 
+                    assertThat(registry.getMeters().get(0).getId().getTags().size(), is(2));
+                    assertThat(registry.getMeters().get(0).getId().getTags().get(1), is(Tag.of("selector", "")));
```
scholzj, line 172: Should we apply the same logic here as we did above? I.e. test it for different selectors? Or if we think that it is not needed to test the selector here anymore, do we really need these asserts?

Plus the same as above ... the `get(0)` looks weird. Can't we get the metric by name?

This applies also to the same tests below.

operator-common/src/test/java/io/strimzi/operator/common/OperatorMetricsTest.java
```diff
@@ -166,6 +189,7 @@ protected Status createStatus() {
                 })));
     }
 
+
```
scholzj, line 192: Is this needed?

operator-common/src/test/java/io/strimzi/operator/common/OperatorMetricsTest.java
ShubhamRwt: assertThat(registry.find(AbstractOperator.METRICS_PREFIX + "reconciliations").meter().getId().getTags().size(), is(2));
Will something like this work @scholzj ?

scholzj: That is more like what I had in mind, yes. All the other asserts seem to be using `get` and not `find`. So maybe you can do the same?

ShubhamRwt: Yeah, Iam changing them

operator-common/src/test/java/io/strimzi/operator/common/OperatorMetricsTest.java
ShubhamRwt: for failing reconcile I tried using a non empty label but seem like it breaks the unit test, I get some sort of async error

scholzj: But I guess the label should be there even in that case, or? So it probably needs to be addressed one way or another.

ShubhamRwt: ```java.lang.AssertionError: java.lang.AssertionError: The asynchronous result was expected to have failed``` i get this error when i run it with valued selector
Empty labels work fine

scholzj: I guess that either means there is some error in the code or in the test.

operator-common/src/main/java/io/strimzi/operator/common/AbstractOperator.java
```diff
@@ -204,6 +205,7 @@ private String getLockName(String namespace, String name) {
                     // and might not be really deleted. We have to filter these situations out and ignore the
                     // reconciliation because such resource might be already operated by another instance (where the
                     // same change triggered ADDED event).
+
```
scholzj, line 208: I do not think there is any reason to remove and add empty lines at random places - especially since you are not changing them at all. This comment is directly related to the log message below so the empty line is IMHO not needed.

operator-common/src/main/java/io/strimzi/operator/common/AbstractOperator.java
scholzj: Same here ... we normally always have ampty line between two methods. No reason to delete it here.

operator-common/src/test/java/io/strimzi/operator/common/OperatorMetricsTest.java
scholzj: Here you for some reason add more empty lines again.

operator-common/src/test/java/io/strimzi/operator/common/OperatorMetricsTest.java
```diff
@@ -199,12 +235,13 @@ protected Status createStatus() {
                 .onComplete(context.succeeding(v -> context.verify(() -> {
                     MeterRegistry registry = metrics.meterRegistry();
 
+                    assertThat(registry.find(AbstractOperator.METRICS_PREFIX + "reconciliations").meter().getId().getTags().size(), is(2));
+                    assertThat(registry.find(AbstractOperator.METRICS_PREFIX + "reconciliations").meter().getId().getTags().get(1), is(Tag.of("selector", "")));
```
scholzj, line 239: Is using `find` here and `get` below caused by some difference between those?

operator-common/src/test/java/io/strimzi/operator/common/OperatorMetricsTest.java
```diff
@@ -199,12 +235,13 @@ protected Status createStatus() {
...
                     assertThat(registry.get(AbstractOperator.METRICS_PREFIX + "reconciliations").tag("kind", "TestResource").counter().count(), is(1.0));
                     assertThat(registry.get(AbstractOperator.METRICS_PREFIX + "reconciliations.successful").tag("kind", "TestResource").counter().count(), is(1.0));
                     assertThat(registry.get(AbstractOperator.METRICS_PREFIX + "reconciliations.failed").tag("kind", "TestResource").counter().count(), is(0.0));
                     assertThat(registry.get(AbstractOperator.METRICS_PREFIX + "reconciliations.locked").tag("kind", "TestResource").counter().count(), is(0.0));
                     assertThat(registry.get(AbstractOperator.METRICS_PREFIX + "resources.paused").tag("kind", "TestResource").gauge().value(), is(1.0));
-
                     assertThat(registry.get(AbstractOperator.METRICS_PREFIX + "reconciliations.duration").tag("kind", "TestResource").timer().count(), is(1L));
                     assertThat(registry.get(AbstractOperator.METRICS_PREFIX + "reconciliations.duration").tag("kind", "TestResource").timer().totalTime(TimeUnit.MILLISECONDS), greaterThan(0.0));
```
scholzj, line 246: Is it intended that the new tag is only on the `reconciliations` metric? Or should it be tested on these other metrics as well?

operator-common/src/test/java/io/strimzi/operator/common/OperatorMetricsTest.java
```diff
@@ -252,6 +289,8 @@ protected Status createStatus() {
                 .onComplete(context.failing(v -> context.verify(() -> {
                     MeterRegistry registry = metrics.meterRegistry();
 
+                    assertThat(registry.find(AbstractOperator.METRICS_PREFIX + "reconciliations").meter().getId().getTags().size(), is(2));
```
scholzj, line 292: Same comments as above.

operator-common/src/test/java/io/strimzi/operator/common/OperatorMetricsTest.java
```diff
@@ -313,6 +352,8 @@ protected Status createStatus() {
                 .onComplete(context.succeeding(v -> context.verify(() -> {
                     MeterRegistry registry = metrics.meterRegistry();
 
+                    assertThat(registry.find(AbstractOperator.METRICS_PREFIX + "reconciliations").meter().getId().getTags().size(), is(2));
+                    assertThat(registry.find(AbstractOperator.METRICS_PREFIX + "reconciliations").meter().getId().getTags().get(1), is(Tag.of("selector", "")));
```
scholzj, line 356: Same comments as above.

operator-common/src/test/java/io/strimzi/operator/common/OperatorMetricsTest.java
```diff
@@ -383,6 +424,8 @@ protected Status createStatus() {
         reconcileAllPromise.future().onComplete(context.succeeding(v -> context.verify(() -> {
             MeterRegistry registry = metrics.meterRegistry();
 
+            assertThat(registry.find(AbstractOperator.METRICS_PREFIX + "reconciliations").meter().getId().getTags().size(), is(2));
+            assertThat(registry.find(AbstractOperator.METRICS_PREFIX + "reconciliations").meter().getId().getTags().get(1), is(Tag.of("selector", "")));
```
scholzj, line 428: Same comments as above.

operator-common/src/test/java/io/strimzi/operator/common/OperatorMetricsTest.java
ShubhamRwt: I guessI have made wrong configuration in my ide --> It sometimes add additional lines. I will remove the unnecessory spaces and make the indentation configs proper there

ShubhamRwt:
@scholzj I have added the suggestions and I guess now all the extra line addition issues are resolved

operator-common/src/test/java/io/strimzi/operator/common/OperatorMetricsTest.java
```diff
@@ -95,11 +98,16 @@ protected Status createStatus() {
...
                     assertThat(registry.get(AbstractOperator.METRICS_PREFIX + "reconciliations.successful").tag("kind", "TestResource").counter().count(), is(1.0));
+                    assertThat(registry.get(AbstractOperator.METRICS_PREFIX + "reconciliations.failed").meter().getId().getTags().get(1), is(Tag.of("selector", selectorLabels != null ? selectorLabels.toSelectorString() : "")));
                     assertThat(registry.get(AbstractOperator.METRICS_PREFIX + "reconciliations.failed").tag("kind", "TestResource").counter().count(), is(0.0));
+                    assertThat(registry.get(AbstractOperator.METRICS_PREFIX + "reconciliations.locked").meter().getId().getTags().get(1), is(Tag.of("selector", selectorLabels != null ? selectorLabels.toSelectorString() : "")));
                     assertThat(registry.get(AbstractOperator.METRICS_PREFIX + "reconciliations.locked").tag("kind", "TestResource").counter().count(), is(0.0));
 
+                    assertThat(registry.get(AbstractOperator.METRICS_PREFIX + "reconciliations.duration").meter().getId().getTags().get(1), is(Tag.of("selector", selectorLabels != null ? selectorLabels.toSelectorString() : "")));
```
scholzj, line 110: You seem to repeat `Tag.of("selector", selectorLabels != null ? selectorLabels.toSelectorString() : "")` again and again. Maybe it is worth creating it once before all the asserts and just reuse it? That way, the asserts would be much more readable.

operator-common/src/test/java/io/strimzi/operator/common/OperatorMetricsTest.java
```diff
@@ -147,11 +169,16 @@ protected Status createStatus() {
...
                     assertThat(registry.get(AbstractOperator.METRICS_PREFIX + "reconciliations.successful").tag("kind", "TestResource").counter().count(), is(0.0));
+                    assertThat(registry.get(AbstractOperator.METRICS_PREFIX + "reconciliations.failed").meter().getId().getTags().get(1), is(Tag.of("selector", selectorLabels != null ? selectorLabels.toSelectorString() : "")));
                     assertThat(registry.get(AbstractOperator.METRICS_PREFIX + "reconciliations.failed").tag("kind", "TestResource").counter().count(), is(1.0));
+                    assertThat(registry.get(AbstractOperator.METRICS_PREFIX + "reconciliations.locked").meter().getId().getTags().get(1), is(Tag.of("selector", selectorLabels != null ? selectorLabels.toSelectorString() : "")));
                     assertThat(registry.get(AbstractOperator.METRICS_PREFIX + "reconciliations.locked").tag("kind", "TestResource").counter().count(), is(0.0));
 
+                    assertThat(registry.get(AbstractOperator.METRICS_PREFIX + "reconciliations.duration").meter().getId().getTags().get(1), is(Tag.of("selector", selectorLabels != null ? selectorLabels.toSelectorString() : "")));
```
scholzj, line 181: Same as above ...

operator-common/src/test/java/io/strimzi/operator/common/OperatorMetricsTest.java
```diff
@@ -199,12 +241,18 @@ protected Status createStatus() {
...
+                    assertThat(registry.get(AbstractOperator.METRICS_PREFIX + "reconciliations.failed").meter().getId().getTags().get(1), is(Tag.of("selector", "")));
                     assertThat(registry.get(AbstractOperator.METRICS_PREFIX + "reconciliations.failed").tag("kind", "TestResource").counter().count(), is(0.0));
+                    assertThat(registry.get(AbstractOperator.METRICS_PREFIX + "reconciliations.locked").meter().getId().getTags().get(1), is(Tag.of("selector", "")));
                     assertThat(registry.get(AbstractOperator.METRICS_PREFIX + "reconciliations.locked").tag("kind", "TestResource").counter().count(), is(0.0));
+                    assertThat(registry.get(AbstractOperator.METRICS_PREFIX + "resources.paused").meter().getId().getTags().get(1), is(Tag.of("selector", "")));
                     assertThat(registry.get(AbstractOperator.METRICS_PREFIX + "resources.paused").tag("kind", "TestResource").gauge().value(), is(1.0));
 
+                    assertThat(registry.get(AbstractOperator.METRICS_PREFIX + "reconciliations.duration").meter().getId().getTags().get(1), is(Tag.of("selector", "")));
```
scholzj, line 255: Same as above ... but also, why do we have the selector always empty here unlike the previous test cases? If you think that we do not need to check for the proper labels / tags in every of these tests, it might be ok. But I wonder if having the empty tag test really makes sense here in that case or should be just removed.

operator-common/src/test/java/io/strimzi/operator/common/OperatorMetricsTest.java
```diff
@@ -252,11 +300,16 @@ protected Status createStatus() {
...
                     assertThat(registry.get(AbstractOperator.METRICS_PREFIX + "reconciliations.successful").tag("kind", "TestResource").counter().count(), is(0.0));
+                    assertThat(registry.get(AbstractOperator.METRICS_PREFIX + "reconciliations.failed").meter().getId().getTags().get(1), is(Tag.of("selector", "")));
                     assertThat(registry.get(AbstractOperator.METRICS_PREFIX + "reconciliations.failed").tag("kind", "TestResource").counter().count(), is(0.0));
+                    assertThat(registry.get(AbstractOperator.METRICS_PREFIX + "reconciliations.locked").meter().getId().getTags().get(1), is(Tag.of("selector", "")));
                     assertThat(registry.get(AbstractOperator.METRICS_PREFIX + "reconciliations.locked").tag("kind", "TestResource").counter().count(), is(1.0));
 
+                    assertThat(registry.get(AbstractOperator.METRICS_PREFIX + "reconciliations.duration").meter().getId().getTags().get(1), is(Tag.of("selector", "")));
                     assertThat(registry.get(AbstractOperator.METRICS_PREFIX + "reconciliations.duration").tag("kind", "TestResource").timer().count(), is(0L));
```
scholzj, line 313: Same as above.

operator-common/src/test/java/io/strimzi/operator/common/OperatorMetricsTest.java
```diff
@@ -313,11 +366,16 @@ protected Status createStatus() {
...
+                    assertThat(registry.get(AbstractOperator.METRICS_PREFIX + "reconciliations.successful").meter().getId().getTags().get(1), is(Tag.of("selector", "")));
                     assertThat(registry.get(AbstractOperator.METRICS_PREFIX + "reconciliations.successful").tag("kind", "TestResource").counter().count(), is(1.0));
+                    assertThat(registry.get(AbstractOperator.METRICS_PREFIX + "reconciliations.failed").meter().getId().getTags().get(1), is(Tag.of("selector", "")));
                     assertThat(registry.get(AbstractOperator.METRICS_PREFIX + "reconciliations.failed").tag("kind", "TestResource").counter().count(), is(0.0));
+                    assertThat(registry.get(AbstractOperator.METRICS_PREFIX + "reconciliations.locked").meter().getId().getTags().get(1), is(Tag.of("selector", "")));
                     assertThat(registry.get(AbstractOperator.METRICS_PREFIX + "reconciliations.locked").tag("kind", "TestResource").counter().count(), is(0.0));
 
+                    assertThat(registry.get(AbstractOperator.METRICS_PREFIX + "reconciliations.duration").meter().getId().getTags().get(1), is(Tag.of("selector", "")));
```
scholzj, line 378: Same as above.

operator-common/src/test/java/io/strimzi/operator/common/OperatorMetricsTest.java
```diff
@@ -383,12 +441,18 @@ protected Status createStatus() {
...
+            assertThat(registry.get(AbstractOperator.METRICS_PREFIX + "reconciliations.successful").meter().getId().getTags().get(1), is(Tag.of("selector", "")));
             assertThat(registry.get(AbstractOperator.METRICS_PREFIX + "reconciliations.successful").tag("kind", "TestResource").counter().count(), is(3.0));
+            assertThat(registry.get(AbstractOperator.METRICS_PREFIX + "reconciliations.failed").meter().getId().getTags().get(1), is(Tag.of("selector", "")));
             assertThat(registry.get(AbstractOperator.METRICS_PREFIX + "reconciliations.failed").tag("kind", "TestResource").counter().count(), is(0.0));
+            assertThat(registry.get(AbstractOperator.METRICS_PREFIX + "reconciliations.locked").meter().getId().getTags().get(1), is(Tag.of("selector", "")));
             assertThat(registry.get(AbstractOperator.METRICS_PREFIX + "reconciliations.locked").tag("kind", "TestResource").counter().count(), is(0.0));
 
+            assertThat(registry.get(AbstractOperator.METRICS_PREFIX + "reconciliations.duration").meter().getId().getTags().get(1), is(Tag.of("selector", "")));
```
scholzj, line 455: Same as above.

operator-common/src/test/java/io/strimzi/operator/common/OperatorMetricsTest.java
ShubhamRwt: In these cases we are not passing any selector label. So the selector label is always going to be empty only. I guess this assert is fine in the cases where we are not passing any Labels and keeping them null

operator-common/src/test/java/io/strimzi/operator/common/OperatorMetricsTest.java
ShubhamRwt: @scholzj I have applied these suggestions

operator-common/src/test/java/io/strimzi/operator/common/OperatorMetricsTest.java
ShubhamRwt: I didnt check for empty, null and valued label in all the test, Only the failing reconcile and successful reconcile once. But even if I dont check for labels in the leftiver test it means the labels their are null, so we will definitely get a selector label as value ` selector :  "" `

scholzj (approved):

scholzj:
@ppatierno Can you have a look at this?

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/AbstractConnectOperator.java
```diff
@@ -146,7 +146,7 @@ public AbstractConnectOperator(Vertx vertx, PlatformFeaturesAvailability pfa, St
         this.port = port;
 
         // Setup metrics for connectors
-        Tags metricTags = Tags.of(Tag.of("kind", KafkaConnector.RESOURCE_KIND));
+        Tags metricTags = Tags.of(Tag.of("kind", KafkaConnector.RESOURCE_KIND), Tag.of("selector", ""));
```
sknot-rh, line 149: So Kafka connect does have this selector always empty?

ppatierno: this is KafkaConnector not KafkaConnect

ppatierno (approved):

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/AbstractConnectOperator.java
ShubhamRwt: @sknot-rh Hello, we are keeping it as empty always because of the prometheus meter property. If we dont put the selector here then, it will lead to something like this  --->`Prometheus requires that all meters with the same name have the same set of tag keys. There is already an existing meter named 'strimzi_reconciliations_total' containing tag keys [kind, selector]. The meter you are attempting to register has keys [kind].`

ShubhamRwt: If your question is why we are using an empty selector label.

sknot-rh: Ah, thanks @ppatierno

sknot-rh (approved):

Issue #3204, opened 2020-06-16
[Doc] MIrrorMaker2 docs don't describe how to enable compression in the target cluster


brokenjacobs:
**Suggestion / Problem**
When trying to configure mirrormaker2 I ran into numerous issues trying to get my messages to be compressed in the target cluster. It turns out that mirrormaker2 hands it's messages off to kafka connect. So any producer settings you make in the sourceConnector have no affect (other than increasing cpu usage due to decompression). To properly enable compression on the destination cluster you need to make the producer setting to the kafkaconnect pieces, under clusters.
Like so:
```
  clusters:
  - alias: "sourceCluster"
    bootstrapServers: my-source-cluster-bootstrap:9094
    config:
      config.storage.replication.factor: 1
      offset.storage.replication.factor: 1
      status.storage.replication.factor: 1
  - alias: "targetCluster"
    bootstrapServers: my-target-cluster-bootstrap:9092
    config:
      producer.compression.type: zstd
```
By contrast consumer specific configuration DOES go in the sourceConnector:
mirrors:
  - sourceCluster: "sourceCluster"
    targetCluster: "targetCluster
    sourceConnector:
      config:
        sync.topic.acls.enabled: "false"
        consumer.fetch.min.bytes: 262144
        consumer.fetch.max.wait.ms: 10000
        consumer.max.poll.records: 5000
        consumer.receive.buffer.bytes: -1
        consumer.send.buffer.bytes: -1

**Documentation Link**
https://strimzi.io/docs/operators/master/using.html#proc-mirrormaker-replication-str
I'd like to see this as part of the examples in this section.

Also... it would be good to clarify that kafkamirrormaker2 will NOT work if not colocated with the target cluster. (Pull not push, mm2 has no push support).

Koushikk-24:
@brokenjacobs  Hello, Is this issue free to work upon?

scholzj:
@Koushikk-24 I think you can work on this if you want

Koushikk-24:
> @Koushikk-24 I think you can work on this if you want

Thanks. 
@brokenjacobs The documentation should be updated in this [file](https://github.com/strimzi/strimzi-kafka-operator/blob/main/examples/mirror-maker/kafka-mirror-maker-2-custom-replication-policy.yaml) right?

brokenjacobs:
Yes, thanks all!

Koushikk-24:
This issue is available to resolve

PR #5119, opened 2021-06-11
Updated Mirrormaker2 docs #3204

Koushikk-24:
Signed-off-by: Rupesh <punnarupeshkoushik@gmail.com>

### Updated docs don't describe how to enable compression in the target cluster

_Select the type of your PR_
- Documentation

### Description

This PR updated the Mirror maker 2 docs [clusters and mirrors] issue #3204 

### Checklist

- [ ] Write tests
- [ ] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/documentation/modules/configuring/proc-config-mirrormaker2-replication.adoc
+++ b/documentation/modules/configuring/proc-config-mirrormaker2-replication.adoc
@@ -34,14 +34,26 @@ spec:
   version: {DefaultKafkaVersion}
   connectCluster: "my-cluster-target"
   clusters:
-  - alias: "my-cluster-source"
-    bootstrapServers: my-cluster-source-kafka-bootstrap:9092
-  - alias: "my-cluster-target"
-    bootstrapServers: my-cluster-target-kafka-bootstrap:9092
+  - alias: "sourceCluster"
+    bootstrapServers: my-source-cluster-bootstrap:9094
+    config:
+      config.storage.replication.factor: 1
+      offset.storage.replication.factor: 1
+      status.storage.replication.factor: 1
+  - alias: "targetCluster"
+    bootstrapServers: my-target-cluster-bootstrap:9092
+    config:
+      producer.compression.type: zstd
   mirrors:
-  - sourceCluster: "my-cluster-source"
-    targetCluster: "my-cluster-target"
-    sourceConnector: {}
+  - sourceCluster: "my-source-cluster"
+    targetCluster: "my-target-cluster"
+    sourceConnector:
+      config:
+        replication.factor: 1
+        offset-syncs.topic.replication.factor: 1
+        sync.topic.acls.enabled: "false"
+        replication.policy.separator: ""
+        replication.policy.class: "io.strimzi.kafka.connect.mirror.IdentityReplicationPolicy
 ----
 
 You can configure access control for source and target clusters using TLS or SASL authentication.
--- a/examples/mirror-maker/kafka-mirror-maker-2-custom-replication-policy.yaml
+++ b/examples/mirror-maker/kafka-mirror-maker-2-custom-replication-policy.yaml
@@ -7,25 +7,30 @@ spec:
   replicas: 1
   connectCluster: "my-target-cluster"
   clusters:
-  - alias: "my-source-cluster"
-    bootstrapServers: my-source-cluster-kafka-bootstrap:9092
-  - alias: "my-target-cluster"
-    bootstrapServers: my-target-cluster-kafka-bootstrap:9092
+  - alias: "sourceCluster"
+    bootstrapServers: my-source-cluster-bootstrap:9094
     config:
-      # -1 means it will use the default replication factor configured in the broker
-      config.storage.replication.factor: -1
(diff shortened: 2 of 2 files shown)


scholzj (changes requested):
Thanks for the PR. But I do not think it really solves the issue since it does not change docs, only example files. In addition:
* You should not change anything in the `examples` directory - these are released files as noted there. If you want to change an example, please do it in `packaging/examples`.
* You took an example of how to use custom replication policy and completely removed the custom replication policy. This example is there for a reason. You cannot just remove something to add something completely different. If you want to add an example showing how to add compression, fine. But add a new file for it and name it accordingly.
* All the example follow the same style how things are named etc. Do not change it. Use the same aliases and basic options as all other MM2 examples.

If you want to add this to the actual docs, you can find them here: https://github.com/strimzi/strimzi-kafka-operator/tree/main/documentation ... there is also a guide about how to contribute to the docs: https://strimzi.io/contributing/guide/

Koushikk-24:
@scholzj, thanks for the review. I have updated the doc file

scholzj:
You still need to revert the changes tot he example files. It would also need some explanation - you should not enable the comperession like this without explaining it as it might be confusing and cause problems to users. And the same as what I said for the example applies also for the docs ... do not change everything what is completely unrelated to the issue.

Maybe @PaulRMellor or @laidan6000 can point you to the right place where to add it.

Koushikk-24:
@scholzj, I have reverted the changes of example files and updated the doc file. Can you please review once again?

documentation/modules/configuring/proc-config-mirrormaker2-replication.adoc
```diff
@@ -34,14 +34,26 @@ spec:
...
-    sourceConnector: {}
+  - sourceCluster: "my-source-cluster"
+    targetCluster: "my-target-cluster"
+    sourceConnector:
+      config:
+        replication.factor: 1
+        offset-syncs.topic.replication.factor: 1
+        sync.topic.acls.enabled: "false"
+        replication.policy.separator: ""
+        replication.policy.class: "io.strimzi.kafka.connect.mirror.IdentityReplicationPolicy
```
scholzj, line 56: Why are we adding this configuration here? It seems to be unrelated.

documentation/modules/configuring/proc-config-mirrormaker2-replication.adoc
```diff
@@ -34,14 +34,26 @@ spec:
...
-  - alias: "my-cluster-source"
-    bootstrapServers: my-cluster-source-kafka-bootstrap:9092
-  - alias: "my-cluster-target"
-    bootstrapServers: my-cluster-target-kafka-bootstrap:9092
+  - alias: "sourceCluster"
+    bootstrapServers: my-source-cluster-bootstrap:9094
+    config:
+      config.storage.replication.factor: 1
+      offset.storage.replication.factor: 1
+      status.storage.replication.factor: 1
```
scholzj, line 42: Why are we adding this configuration here? It seems to be unrelated.

documentation/modules/configuring/proc-config-mirrormaker2-replication.adoc
```diff
@@ -34,14 +34,26 @@ spec:
   version: {DefaultKafkaVersion}
   connectCluster: "my-cluster-target"
   clusters:
-  - alias: "my-cluster-source"
-    bootstrapServers: my-cluster-source-kafka-bootstrap:9092
-  - alias: "my-cluster-target"
-    bootstrapServers: my-cluster-target-kafka-bootstrap:9092
+  - alias: "sourceCluster"
```
scholzj, line 37: Please use the original name. This now anyway breaks the whole example because you change the aliases here but the references use the old names.

documentation/modules/configuring/proc-config-mirrormaker2-replication.adoc
```diff
@@ -34,14 +34,26 @@ spec:
...
+    bootstrapServers: my-source-cluster-bootstrap:9094
+    config:
+      config.storage.replication.factor: 1
+      offset.storage.replication.factor: 1
+      status.storage.replication.factor: 1
+  - alias: "targetCluster"
```
scholzj, line 43: Please use the original name.

documentation/modules/configuring/proc-config-mirrormaker2-replication.adoc
```diff
@@ -34,14 +34,26 @@ spec:
...
+    bootstrapServers: my-target-cluster-bootstrap:9092
+    config:
+      producer.compression.type: zstd
```
scholzj, line 46: As I said before. The easiest way to add this would be to add a small paragraphs saying something like _If you want to use compression to ...._ And add a separate snippet with this to it.

scholzj (changes requested):

Koushikk-24:
> @scholzj, I have reverted the changes of example files and updated the doc file. Can you please review once again?

Thanks for the review,  it looks like I still didn't properly understand the solution for this issue.

PR #5121, opened 2021-06-12
Fix file-name issue in Kafka Connect Build

scholzj:
### Type of change

- Bugfix
- Enhancement / new feature

### Description

When Kafka Connect Build is downloading artifacts, it is currently using the last segment of the URL as a filename. However, as discussed in #5117 this does not really work in all cases. For example with he download URL `https://drive.google.com/uc?export=download&id=1sEZHDMskmwV31vTx7nUI_yoJM7AgdpXX`, the resulting file name is `uc?export=download&id=1sEZHDMskmwV31vTx7nUI_yoJM7AgdpXX` which is not valid.

The easy fix is to use the URL hash as the file name. This is already used to create the directories where the files are added. This works without issues for ZIP and TGZ archives where the files are anyway just temporary and are deleted after unpacking. It works also for JARs which can be named `<hash>.jar`.

But `type: jar` artifacts are also used to simply download and place any file. This could be for example an `so` library (used by some Debezium connectors as an example) or it could be a LICENSE file etc. Naming them `<hash>.jar` makes little sense and does not work. therefore this PR adds a new artifact type `other`. This an be used to download assorted artifacts without any expectation of a file type. Optionally, user can also specify `fileName` option for this type of artifact which will be the name of the file used in the target image. For example:

```yaml
        - fileName: my-db-driver.so
          type: other
          url: https://some-url.tld/downaload?artifact=12334&path=some/random/path&someOther=option
```

If the `fileName` is not specified, the file will be simply named `<hash>` based on the URL hash. This should fix the issue with weird file names causing errors. 

The downside is that any user who currently uses `type: jar` artifact to download a file under some specific name will now experience a change which might break his deployment and might need to change the CR to use the `type: other`artifact. The alternative options to this would be to leave this unfixed for `type: jar` artifacts. But this seems like a cleaner way to go for the future.

### Checklist

- [x] Write tests
- [x] Update documentation
- [x] Update CHANGELOG.md


tombentley (commented):
So `other` also get added to the plugin path in exactly the same way as `jar`, they just don't need to be packaged as a jar, or have `.jar` suffix?

What provision is there for detecting/handling name collisions?

scholzj:
> So other also get added to the plugin path in exactly the same way as jar, they just don't need to be packaged as a jar, or have .jar suffix?

Yes. The `other` type can be used by any file type basically and the user can specify the end name.

> What provision is there for detecting/handling name collisions?

Each artifact is placed in its own sub-directory named based on the download URL hash (this is not new => this was done already before). So leaving aside possible hash collisions, to have colliding names you would need to have basically a duplicate artifact (the same URL or the same URL and fileName in multiple artifacts of the same type).

tombentley (approved):

sknot-rh (approved):

CHANGELOG.md
```diff
@@ -21,6 +21,8 @@
 * The field `whitelist` in the `KafkaMirrorMaker` resource is deprecated and will be removed in the future.
   It is replaced with a new field `include`.
 * `bind-utils` removed from containers to improve security posture.
+* From Strimzi 0.24 on, Kafka Connect Build is using hashes instead of the last segment of the download URL to name the downloaded files.
+  In case you need to have your artifact to use some specific name, you can use the new `type: other` artifact and its `fileName` field.
```
PaulRMellor, line 25:
```suggestion
* Kafka Connect Build now uses hashes to name downloaded artifact files. Previously, it was using the last segment of the download URL.
  If your artifact requires a specific name, you can use the new `type: other` artifact and its `fileName` field.
```

api/src/main/java/io/strimzi/api/kafka/model/connect/build/OtherArtifact.java
```diff
@@ -0,0 +1,46 @@
...
+    String fileName;
+
+    @Description("Must be `" + TYPE_OTHER + "`")
+    @JsonInclude(JsonInclude.Include.NON_NULL)
+    @Override
+    public String getType() {
+        return TYPE_OTHER;
+    }
+
+    @Description("Name under which the artifact will be stored.")
```
PaulRMellor, line 37:
```suggestion
    @Description("Name under which the artifact is stored.")
```

documentation/api/io.strimzi.api.kafka.model.connect.build.Build.adoc
```diff
@@ -115,14 +115,14 @@ spec:
 Strimzi supports two types of artifacts:
```
PaulRMellor, line 115:
```suggestion
Strimzi supports the following types of artifacts:
```

documentation/api/io.strimzi.api.kafka.model.connect.build.Build.adoc
```diff
@@ -115,14 +115,14 @@ spec:
...
 * JAR files, which are downloaded and used directly
 * TGZ archives, which are downloaded and unpacked
+* Other files, which are downloaded and used directly
```
PaulRMellor, line 118: Can we give an example of _other_ files?

documentation/api/io.strimzi.api.kafka.model.connect.build.Build.adoc
```diff
@@ -115,14 +115,14 @@ spec:
...
 
 IMPORTANT: Strimzi does not perform any security scanning of the downloaded artifacts.
 For security reasons, you should first verify the artifacts manually, and configure the checksum verification to make sure the same artifact is used in the automated build and in the Kafka Connect deployment.
 
 .Using JAR artifacts
 
-JAR artifacts represent a resource which is downloaded and added to a container image.
-JAR artifacts are mainly used for downloading JAR files, but they can also used to download other file types.
+JAR artifacts represent a JAR file which is downloaded and added to a container image.
```
PaulRMellor, line 125:
```suggestion
JAR artifacts represent a JAR file that is downloaded and added to a container image.
```

documentation/api/io.strimzi.api.kafka.model.connect.build.Build.adoc
PaulRMellor, line 118:
```suggestion
* Other artifacts, which are downloaded and used directly
```

PaulRMellor, line 118:
```suggestion
* Other artifacts, which are downloaded and used directly
```

documentation/api/io.strimzi.api.kafka.model.connect.build.Build.adoc
```diff
@@ -187,3 +187,37 @@ spec:
 <1> (Required) Type of artifact.
 <2> (Required) URL from which the archive is downloaded.
 <3> (Optional) SHA-512 checksum to verify the artifact.
+
+.Using `other` artifacts
+
+`other` artifacts represent any kind of file which is downloaded and added to a container image.
```
PaulRMellor, line 193:
```suggestion
`other` artifacts represent any kind of file that is downloaded and added to a container image.
```

PaulRMellor: Can mean JAR or TGZ artifacts? Or anything other than those types of artifacts?

documentation/api/io.strimzi.api.kafka.model.connect.build.Build.adoc
```diff
@@ -187,3 +187,37 @@ spec:
...
+A file name which will be used in the resulting container image can be also specified using the `fileName` field.
```
PaulRMellor, line 194:
```suggestion
If you want to use a specific name for the artifact in the resulting container image, use the `fileName` field.
```

documentation/api/io.strimzi.api.kafka.model.connect.build.Build.adoc
```diff
@@ -187,3 +187,37 @@ spec:
...
+  build:
+    output:
+      #...
+    plugins:
+      - name: my-plugin
+        artifacts:
+          - type: other # <1>
+            url: https://my-domain.tld/my-other-file.ext # <2>
+            sha512sum: 589...ab4 # <3>
+            fileName: name-the-file.ext # <4>
```
PaulRMellor, line 217:
```suggestion
          - type: other  <1>
            url: https://my-domain.tld/my-other-file.ext  <2>
            sha512sum: 589...ab4  <3>
            fileName: name-the-file.ext  <4>
```

documentation/api/io.strimzi.api.kafka.model.connect.build.Build.adoc
```diff
@@ -187,3 +187,37 @@ spec:
...
+  #...
+----
+<1> (Required) Type of artifact.
+<2> (Required) URL from which the artifact is downloaded.
+<3> (Optional) SHA-512 checksum to verify the artifact.
+<4> (Optional) The name under which the file will be stored in the resulting container image.
```
PaulRMellor, line 223:
```suggestion
<4> (Optional) The name under which the file is stored in the resulting container image.
```

PaulRMellor (commented):
Looks good. Few minor comments

documentation/api/io.strimzi.api.kafka.model.connect.build.Build.adoc
PaulRMellor: Maybe add what happens if you don't add one?
_If the fileName is not specified, the file is named based on the URL hash._

scholzj:
@PaulRMellor Thanks. The review comments should be implemented now.

scholzj:
@PaulRMellor Sorry, I merged this by mistake without realising you didn't approved it. I guess the PR might be fine since I added all your comments / suggestions. But if you have something more please let me know and I will open a new PR for it.

PR #5128, opened 2021-06-15
Make sure `getCruiseControlDescription` returns only after it collected / created the description

### LINKED ISSUE #5126: ServiceAccount gets periodically deleted/created if CruiseControl CM metrics are used
We currently use the inline metrics from the examples in our Kafka setup, which works fine. The setup uses Istio as service mesh. Since these metrics are deprecated, we want to switch to the CM metrics instead. This seems to work fine for all resources except of CruiseControl.
After we use the new `metricsConfig` there, we observed connection issues in the Istio proxy and deamon after some time of execution. This leads to a memory leakage in the Istio proxy.

If I look into the operator logs (in DEBUG mode), I can see that the operator deletes the CruiseControl ServiceAccount and recreates it in one of the following reconciliation periods. This also leads to a new created token which belongs to the SA.
The process occurs from time to time. I've noticed no fix interval, but it seems to occur at least once per hour.
I cannot see any detail on the log why the SA was deleted by the operator.

I can reproduce the behavior with the current Helm chart release and the sample CruiseControl metrics config.

So far I only have noticed this issue on CruiseControl.

**How To Reproduce**
1. Install the Helm chart with debug log level: `helm install my-release strimzi/strimzi-kafka-operator --set logLevel=DEBUG`
2. Install [CruiseControl metrics example](https://github.com/strimzi/strimzi-kafka-operator/blob/main/examples/metrics/kafka-cruise-control-metrics.yaml)
3. Watch SA of CruiseControl
4. If the SA was deleted/recreated see operator logs

**Expected behavior**
The CruiseControl ServiceAccount should be stable to avoid recreations of the SA token.

 - Strimzi version: 0.22.1, 0.23.0
 - Kubernetes cluster: Kubernetes 1.19.7
 - Infrastructure: Kind

**YAML files and logs**
ServiceAccountOperator logs:
```
2021-06-15 10:09:02 DEBUG ServiceAccountOperator:111 - ServiceAccount strimzi2/my-cluster-cruise-control already exists, patching it
2021-06-15 10:11:02 DEBUG ServiceAccountOperator:111 - ServiceAccount strimzi2/my-cluster-cruise-control already exists, patching it
2021-06-15 10:13:02 DEBUG ServiceAccountOperator:111 - ServiceAccount strimzi2/my-cluster-cruise-control already exists, patching it
2021-06-15 10:15:02 DEBUG ServiceAccountOperator:117 - ServiceAccount strimzi2/my-cluster-cruise-control exist, deleting it
2021-06-15 10:15:02 DEBUG ResourceSupport:148 - Opened watch io.fabric8.kubernetes.client.dsl.internal.WatchConnectionManager@18071fc for evaluation of observe deletion of ServiceAccount strimzi2/my-cluster-cruise-control
2021-06-15 10:15:02 DEBUG ResourceSupport:165 - Not yet satisfied: observe deletion of ServiceAccount strimzi2/my-cluster-cruise-control
2021-06-15 10:15:02 DEBUG ResourceSupport:165 - Not yet satisfied: observe deletion of ServiceAccount strimzi2/my-cluster-cruise-control
2021-06-15 10:15:02 DEBUG ServiceAccountOperator:166 - ServiceAccount strimzi2/my-cluster-cruise-control has been deleted
2021-06-15 10:17:02 DEBUG ServiceAccountOperator:108 - ServiceAccount strimzi2/my-cluster-cruise-control does not exist, creating it
2021-06-15 10:19:02 DEBUG ServiceAccountOperator:111 - ServiceAccount strimzi2/my-cluster-cruise-control already exists, patching it
```
Full operator and CruiseControl log attached.
[strimzi-operator.log](https://github.com/strimzi/strimzi-kafka-operator/files/6654898/strimzi-operator.log)
[cruise-control.log](https://github.com/strimzi/strimzi-kafka-operator/files/6654899/cruise-control.log)

I can't find anything suspicious in the [ServiceAccountOperator](https://github.com/strimzi/strimzi-kafka-operator/blob/3b3a419dd17cac0458b2a96d3d2b6f9d47683ebe/operator-common/src/main/java/io/strimzi/operator/common/operator/resource/ServiceAccountOperator.java) or the [AbstractResourceOperator](https://github.com/strimzi/strimzi-kafka-operator/blob/3b3a419dd17cac0458b2a96d3d2b6f9d47683ebe/operator-common/src/main/java/io/strimzi/operator/common/operator/resource/AbstractResourceOperator.java#L81).

Does this behavior also exists on the current main branch, or was this fixed by the [SA patching changes](https://github.com/strimzi/strimzi-kafka-operator/pull/4974)?

Thanks for the help!

scholzj:
### Type of change

- Bugfix

### Description

The `getCruiseControlDescription` in `KafkaAssemblyOperator` is missing return statement when Cruise Control is configured and instead returns as if it was not. That causes the description to be collected asynchronously while the next methods are already running. #5126 describes how it can cause the SA or Network Policy to be deleted although Cruise Control itself is enabled. This bug probably exists for some time, but it was less visible when the metrics configuration was done inline in the Kafka CR since now with the metrics configuration in a ConfigMap, we actually query the ConfigMap which takes more time.

This PR fixes the code to return only after the description is collected / created and the network policies or service accounts should be created properly every recocniliation.

This should close #5126 

### Checklist

### DIFF
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaAssemblyOperator.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaAssemblyOperator.java
@@ -3348,8 +3348,9 @@ public class KafkaAssemblyOperator extends AbstractAssemblyOperator<KubernetesCl
 
         private final Future<ReconciliationState> getCruiseControlDescription() {
             CruiseControl cruiseControl = CruiseControl.fromCrd(reconciliation, kafkaAssembly, versions);
+
             if (cruiseControl != null) {
-                Util.metricsAndLogging(reconciliation, configMapOperations, kafkaAssembly.getMetadata().getNamespace(),
+                return Util.metricsAndLogging(reconciliation, configMapOperations, kafkaAssembly.getMetadata().getNamespace(),
                         cruiseControl.getLogging(), cruiseControl.getMetricsConfigInCm())
                         .compose(metricsAndLogging -> {
                             ConfigMap logAndMetricsConfigMap = cruiseControl.generateMetricsAndLogConfigMap(metricsAndLogging);
@@ -3358,12 +3359,13 @@ public class KafkaAssemblyOperator extends AbstractAssemblyOperator<KubernetesCl
 
                             this.cruiseControlMetricsAndLogsConfigMap = logAndMetricsConfigMap;
                             this.cruiseControl = cruiseControl;
-
                             this.ccDeployment = cruiseControl.generateDeployment(pfa.isOpenshift(), annotations, imagePullPolicy, imagePullSecrets);
+
                             return Future.succeededFuture(this);
                         });
+            } else {
+                return withVoid(Future.succeededFuture());
             }
-            return withVoid(Future.succeededFuture());
         }
 
         Future<ReconciliationState> cruiseControlServiceAccount() {
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaAssemblyOperator.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaAssemblyOperator.java
@@ -3346,7 +3346,7 @@ public class KafkaAssemblyOperator extends AbstractAssemblyOperator<KubernetesCl
             return stsGeneration == podGeneration;
         }
 
-        private final Future<ReconciliationState> getCruiseControlDescription() {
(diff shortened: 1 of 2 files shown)


steadyk (approved):
Looks good to me! Thanks for the quick change!

tombentley (commented):
No test?

scholzj:
@tombentley TBH, I'm not sure it is really needed. But I added a test which should IMHO cover it. Please have a look.

PR #5130, opened 2021-06-15
proper link to examples/metrics

tringuyen-yw:
### Type of change

- Documentation

### Description

Fix the quote `examples/metrics` to a proprer clickable link.

### Checklist

- [ ] Write tests
- [ X] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/documentation/assemblies/metrics/assembly_metrics-cruise-control.adoc
+++ b/documentation/assemblies/metrics/assembly_metrics-cruise-control.adoc
@@ -43,7 +43,7 @@ spec:
 
 If you deployed Cruise Control with your Kafka cluster with the metrics enabled, you can enable Grafana to present the metrics data it exposes.
 
-A Cruise Control dashboard is provided in the `examples/metrics` directory as a JSON file:
+A Cruise Control dashboard is provided in the link:/examples/metrics[examples/metrics] directory as a JSON file:
 
 * `strimzi-cruise-control.json`
 
--- a/documentation/assemblies/metrics/assembly_metrics-cruise-control.adoc
+++ b/documentation/assemblies/metrics/assembly_metrics-cruise-control.adoc
@@ -43,7 +43,7 @@ spec:
 
 If you deployed Cruise Control with your Kafka cluster with the metrics enabled, you can enable Grafana to present the metrics data it exposes.
 
-A Cruise Control dashboard is provided in the link:/examples/metrics[examples/metrics] directory as a JSON file:
+A Cruise Control dashboard is provided in the link:https://github.com/strimzi/strimzi-kafka-operator/tree/{GithubVersion}/examples/metrics[examples/metrics] directory as a JSON file:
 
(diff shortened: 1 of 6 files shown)


scholzj (commented):
I'm not sure there is a single link which would work in every place, or? How will the link to `/examples/metrics` work for example on the website or when the user unpacks it from the release ZIP?

Maybe using something like this could work:
```
link:https://github.com/strimzi/strimzi-kafka-operator/tree/{GithubVersion}/examples/metrics[examples/metrics]
```

scholzj:
@tringuyen-yw Also, can you please fix the DCO signoff? The instructions should be under the _Details_ link next to the DCO status.

tringuyen-yw:
@scholzj the link `/examples/metrics` refers to the root of the project. When Asciidoc builds the HTML, it would generate the proprer link. On [Github, working branch](https://github.com/YoppWorks/strimzi-kafka-operator/blob/Tri_Strimzi-doc-fix-link-examples-metrics/documentation/modules/metrics/ref_metrics-config-files.adoc), the link is working OK.

But to play it safe, we can of course set the absolute link. What does `{GithubVersion}` expands to? 

Will make change now and do the DCO signoff

tringuyen-yw:
@scholzj Committed the link you suggested 
You can check on the working branch that the link to `examples/metrics` is now broken because of the `{GithubVersion}` attributed which is not rendered: [ref_metrics-config-files.adoc](https://github.com/YoppWorks/strimzi-kafka-operator/blob/Tri_Strimzi-doc-fix-link-examples-metrics/documentation/modules/metrics/ref_metrics-config-files.adoc)

Would you be OK if I hardcode the link to `tree/main` instead of `tree/{GithubVersion}`

```
link:https://github.com/strimzi/strimzi-kafka-operator/tree/main/examples/metrics[examples/metrics]
```
This would make a link always pointing to `master` [examples/metrics](https://github.com/strimzi/strimzi-kafka-operator/tree/main/examples/metrics)

scholzj:
> You can check on the working branch that the link to examples/metrics is now broken because of the {GithubVersion} attributed which is not rendered: ref_metrics-config-files.adoc

When the documentation is built, the argument with the `GithubVersion` is passed into it. So when you do `make docu_html`, you should see the linked rendered to either point to the `main` branch or in case of releases it will point to the right tag (this is important, because the current `main` branch metric examples would not work with older Strimzi versions etc.)

If you search the `documentation` folder, there should be a file `attributes.adoc`. Maybe you can also add the `GithubVersion` attribute there with the `main` as a default value. That should make sure that in case you do not use one of the `make` command to build the docs, you get at least to the `main` branch. You can just add there something like:
```
:GithubVersion: main
```

scholzj (approved):

Frawless (approved):

tombentley (approved):

scholzj:
Thanks for the PR @tringuyen-yw

PR #5137, opened 2021-06-16
[MO] - [system test] -> rbac support

see-quick:

### Type of change

- Bugfix
- Enhancement / new feature

### Description

This PR adding support in our system tests to run to RBAC. Moreover there is little change in LogCollector which fixing bug that if there was some problem of deploying Cluster Operator it will no longer masked original error. 

### Checklist

### DIFF
--- a/systemtest/src/main/java/io/strimzi/systemtest/utils/StUtils.java
+++ b/systemtest/src/main/java/io/strimzi/systemtest/utils/StUtils.java
@@ -376,11 +376,11 @@ public class StUtils {
 
     /**
      * Retrieve namespace based on the cluster configuration
-     * @param singleNamespace suite namespace
+     * @param namespace suite namespace
      * @param extensionContext test context for get the parallel namespace
      * @return single or parallel namespace based on cluster configuration
      */
-    public static String getNamespaceBasedOnRbac(String singleNamespace, ExtensionContext extensionContext) {
-        return Environment.isNamespaceRbacScope() ? singleNamespace : extensionContext.getStore(ExtensionContext.Namespace.GLOBAL).get(Constants.NAMESPACE_KEY).toString();
+    public static String getNamespaceBasedOnRbac(String namespace, ExtensionContext extensionContext) {
+        return Environment.isNamespaceRbacScope() ? namespace : extensionContext.getStore(ExtensionContext.Namespace.GLOBAL).get(Constants.NAMESPACE_KEY).toString();
     }
 }
--- a/systemtest/src/main/java/io/strimzi/systemtest/utils/StUtils.java
+++ b/systemtest/src/main/java/io/strimzi/systemtest/utils/StUtils.java
@@ -373,4 +373,14 @@ public class StUtils {
                 .toLowerCase(Locale.ENGLISH)
                 .contains(PARALLEL_NAMESPACE)).count() == 1;
     }
+
+    /**
+     * Retrieve namespace based on the cluster configuration
+     * @param singleNamespace suite namespace
+     * @param extensionContext test context for get the parallel namespace
+     * @return single or parallel namespace based on cluster configuration
+     */
+    public static String getNamespaceBasedOnRbac(String singleNamespace, ExtensionContext extensionContext) {
+        return Environment.isNamespaceRbacScope() ? singleNamespace : extensionContext.getStore(ExtensionContext.Namespace.GLOBAL).get(Constants.NAMESPACE_KEY).toString();
+    }
 }
--- a/systemtest/src/test/java/io/strimzi/systemtest/AbstractST.java
+++ b/systemtest/src/test/java/io/strimzi/systemtest/AbstractST.java
@@ -534,11 +534,13 @@ public abstract class AbstractST implements TestSeparator {
 
             // if 'parallel namespace test' we are gonna delete namespace
             if (StUtils.isParallelNamespaceTest(extensionContext)) {
+                // if RBAC is enable we don't run tests in parallel mode and with that said we don't create another namespaces
+                if (!Environment.isNamespaceRbacScope()) {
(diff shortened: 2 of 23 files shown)


systemtest/src/main/java/io/strimzi/systemtest/utils/StUtils.java
```diff
@@ -373,4 +373,14 @@ public static boolean isParallelNamespaceTest(ExtensionContext extensionContext)
...
                 .contains(PARALLEL_NAMESPACE)).count() == 1;
     }
+
+    /**
+     * Retrieve namespace based on the cluster configuration
+     * @param singleNamespace suite namespace
+     * @param extensionContext test context for get the parallel namespace
+     * @return single or parallel namespace based on cluster configuration
+     */
+    public static String getNamespaceBasedOnRbac(String singleNamespace, ExtensionContext extensionContext) {
```
Frawless, line 383:
```suggestion
    public static String getNamespaceBasedOnRbac(String coNamespace, ExtensionContext extensionContext) {
```

or maybe just a simple `namespace` ?

systemtest/src/test/java/io/strimzi/systemtest/kafka/KafkaST.java
```diff
@@ -813,7 +813,7 @@ void testEntityOperatorWithoutTopicOperator(ExtensionContext extensionContext) {
 
     @ParallelNamespaceTest
     void testEntityOperatorWithoutUserOperator(ExtensionContext extensionContext) {
-        final String namespaceName = extensionContext.getStore(ExtensionContext.Namespace.GLOBAL).get(Constants.NAMESPACE_KEY).toString();
+        final String namespaceName = Environment.isNamespaceRbacScope() ? NAMESPACE : extensionContext.getStore(ExtensionContext.Namespace.GLOBAL).get(Constants.NAMESPACE_KEY).toString();
```
Frawless, line 816: Why this (and following) line is different? Shouldn't you use the same method for getting the namespace?

see-quick: Correct

see-quick:
Regression pipeline passed:  https://dev.azure.com/cncf/strimzi/_build/results?buildId=79249&view=results

systemtest/src/test/java/io/strimzi/systemtest/metrics/MetricsST.java
```diff
@@ -604,6 +606,8 @@ private void assertCoMetricResources(String kind, int value) {
 
     @BeforeAll
     void setupEnvironment(ExtensionContext extensionContext) throws Exception {
+        assumeFalse(Environment.isNamespaceRbacScope());
```
scholzj, line 609: Why do metrics tests need to be disabled?

see-quick: because this suite works with 3 namespaces and AFAIK the RBAC can work with only one namespace at once so that's why I disabled it.

scholzj: Why does it need multiple namespace? Metrics do not have anything what would be dependent on global scope.

see-quick: but in that suite we deploy 2 Kafka clusters in separate namespace

Frawless: You can add this assume just only to mm test in metrics I think

see-quick: > You can add this assume just only to mm test in metrics I think

No, we can't already discussed offline.

Frawless (approved):

systemtest/src/test/java/io/strimzi/systemtest/AbstractST.java
```diff
@@ -581,14 +583,17 @@ protected void beforeEachMayOverride(ExtensionContext extensionContext) {
 
             // if 'parallel namespace test' we are gonna create
```
Frawless, line 584:
```suggestion
            // if 'parallel namespace test' we are gonna create namespace
```

see-quick:
Blocked by https://github.com/strimzi/strimzi-kafka-operator/pull/5223

systemtest/src/test/java/io/strimzi/systemtest/metrics/MetricsST.java
```diff
@@ -256,6 +257,7 @@ void testKafkaExporterDifferentSetting() throws InterruptedException, ExecutionE
     }
 
     @ParallelTest
+    @Disabled("Because of this PR https://github.com/strimzi/strimzi-kafka-operator/pull/5137.")
```
Frawless, line 260: It's not working even for bundle install type with cluster-wide operator?

Frawless: discussed offline. Make sense now.

Frawless (approved):

systemtest/src/test/java/io/strimzi/systemtest/metrics/MetricsST.java
scholzj: So, what does it mean? Can you explain it here as well?

see-quick:
We want to cherry-pick to `0.24`

systemtest/src/test/java/io/strimzi/systemtest/metrics/MetricsST.java
see-quick: It works with the cluster-wide operator but I let the implementation of the test for explicit 2 namespaces (which currently doesn't work but with Standa's PR it will work). That's it.

PR #5138, opened 2021-06-16
Fix connect logging hierarchy

sknot-rh:

### Type of change
- Bugfix

### Description
The logging hierarchy causes issues when the root (or any upper level) logger level is updated. As an optimization step, we are not sending all the desired log levels but the changed ones only. If the root level is changed, all its children are set to parent level.
So if the root level is changed, we need to reset all the children logger levels. For example, if we change
`io.com` and we are using also `io.com.logger1`, the operator needs to set also the `io.com.logger1`

### Checklist

- [x] Write tests
- [ ] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/cluster-operator/src/test/java/io/strimzi/operator/cluster/operator/assembly/KafkaConnectApiTest.java
+++ b/cluster-operator/src/test/java/io/strimzi/operator/cluster/operator/assembly/KafkaConnectApiTest.java
@@ -7,6 +7,7 @@ package io.strimzi.operator.cluster.operator.assembly;
 import io.strimzi.api.kafka.model.connect.ConnectorPlugin;
 import io.strimzi.operator.common.BackOff;
 import io.strimzi.operator.common.Reconciliation;
+import io.strimzi.operator.common.Util;
 import io.strimzi.operator.common.model.OrderedProperties;
 import io.strimzi.test.TestUtils;
 import io.strimzi.test.annotations.IsolatedTest;
@@ -37,6 +38,7 @@ import java.util.concurrent.CountDownLatch;
 
 import static java.util.Collections.emptyMap;
 import static java.util.Collections.singletonList;
+import static java.util.Collections.singletonMap;
 import static org.hamcrest.MatcherAssert.assertThat;
 import static org.hamcrest.Matchers.containsString;
 import static org.hamcrest.Matchers.empty;
@@ -47,6 +49,7 @@ import static org.hamcrest.Matchers.is;
 import static org.hamcrest.Matchers.not;
 import static org.hamcrest.Matchers.nullValue;
 import static org.junit.jupiter.api.Assertions.assertNotNull;
+import static org.junit.jupiter.api.Assertions.assertTrue;
 
 @ExtendWith(VertxExtension.class)
 public class KafkaConnectApiTest {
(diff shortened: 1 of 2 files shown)


scholzj (commented):
To be honest, I never looked into the code for this. So I wonder:
* Do we always identify the changes since last reconciliation or do we always update all loggers?
* Is this really the way the API is expected to be used? It sounds like this is a weird design which can lead to missing log messages. Because it sounds like change in the higher level will basically temporarily affect all levels below it, so messages might be missed.

sknot-rh:
We do identify the changes. Yes, sadly the API cannot do the changes in an atomic way. So there might be some tiny window with the wrong log level. This happens only when the parent level is changed.

scholzj:
@tombentley Do you know if this expected behaviour? ^^^

I guess one day the Log4j2 solves that :-/.

cluster-operator/src/test/java/io/strimzi/operator/cluster/operator/assembly/KafkaConnectApiTest.java
```diff
@@ -279,4 +282,34 @@ public void testChangeLoggers(VertxTestContext context) throws InterruptedExcept
...
+        // root logger level changed
+        assertTrue(client.parentLogLevelChanged(fetched, ops, Map.entry("io.org.com", "WARN")));
+
+        fetched = new HashMap();
+        fetched.put("root", singletonMap("level", "TRACE"));
+        fetched.put("io.org", singletonMap("level", "ERROR"));
+        fetched.put("io.org.com", singletonMap("level", "FATAL"));
+
+        // root logger level did not change. The parent logger level did
+        assertTrue(client.parentLogLevelChanged(fetched, ops, Map.entry("io.org.com", "WARN")));
```
scholzj, line 312: Should we also add a test that logger from other hierarchy is not affected here? E.g. have a logger `com.org.io` and check that it returns false?

cluster-operator/src/test/java/io/strimzi/operator/cluster/operator/assembly/KafkaConnectApiTest.java
```diff
@@ -279,4 +282,34 @@ public void testChangeLoggers(VertxTestContext context) throws InterruptedExcept
...
+
+        KafkaConnectApiImpl client = new KafkaConnectApiImpl(vertx);
+
+        Map<String, Map<String, String>> fetched = new HashMap();
+        fetched.put("root", singletonMap("level", "INFO"));
+        fetched.put("io.org", singletonMap("level", "WARN"));
+        fetched.put("io.org.com", singletonMap("level", "FATAL"));
+
+        // root logger level changed
+        assertTrue(client.parentLogLevelChanged(fetched, ops, Map.entry("io.org.com", "WARN")));
```
scholzj, line 304: Should we also have some negative test case to check that if nothing changed it returns false?

scholzj (approved):

scholzj (approved):

tombentley (commented):
This looks surprisingly involved for what it's trying to do. It's not simple logic to follow. TBH I wonder whether the "just send what's changed" is really an optimization worth making. If we didn't do that we'd just need to sort the loggers (as we current do) and make the REST call for each of them, right?

scholzj:
> This looks surprisingly involved for what it's trying to do. It's not simple logic to follow. TBH I wonder whether the "just send what's changed" is really an optimization worth making. If we didn't do that we'd just need to sort the loggers (as we current do) and make the REST call for each of them, right?

Well, given how it works you would every two minutes have a window where you might miss messages or get unwanted messages because of the inheritance. Plus it might have some perf impact.

sknot-rh:
Yeah, sending everything each reconciliation means you do send even the root level logger each reconciliation. That results in overriding  _each_ child logger for a brief moment. 
Another thing is - let's say you have (for some reason) 200 loggers. Even when you do not change anything, you would send 200 put requests every reconcliliation.

tombentley:
Yes, but we're getting the current logger levels anyway, so what I'm suggesting is we compute the _complete_ update Map (rather than trying to compute a minimal update map), but only actually do the update if that differs from the current map. That way most of the time we don't do any update, but when we do we update all the loggers.

tombentley:
I've been working on a patch to do this. If the tests pass I'll push it and you can tell me if I've overlooked something.

PR #5160, opened 2021-06-19
Print deprecation warning for the enableECDSA field only when really set

scholzj:
### Type of change

- Bugfix

### Description

The `enableECDSA` field has been deprecated. But because it is a `boolean` which is set to `true` by default, it cause the operator to always print the deprecation warning whenever OAuth authentication is enabled in the broker even when the `enableECDSA` field is not used:

```
In API version kafka.strimzi.io/v1beta2 the enableECDSA property at path spec.kafka.listeners.genericKafkaListeners.auth.enableECDSA has been deprecated.
```

Similar warning is also in the Kafka CR status.

This PR changes the field type to `Boolean` and sets it to null by default. That way, if the `enableECDSA` fiels is not set, no warning is shown. Only when the field is really set in the Kafka CR to `true` or `false` the warning is printed and set in the resource status.

This should be picked for the 0.24.0 release.

### Checklist


ppatierno (approved):

mstruk (approved):
The fix looks good. Tried it locally, and it works.

Issue #3741, opened 2020-10-02
[Enhancement] Add SASL support to Topic Operator for standalone use


Coqueiro:
I've followed the [documentation on how to deploy a standalone Topic Operator](https://strimzi.io/docs/operators/master/deploying.html#deploying-the-topic-operator-standalone-str).

It asks for those variables:
```
STRIMZI_TRUSTSTORE_LOCATION
STRIMZI_TRUSTSTORE_PASSWORD
STRIMZI_KEYSTORE_LOCATION
STRIMZI_KEYSTORE_PASSWORD
```

That said, we have a MSK cluster using SCRAM over SASL_SSL (https://docs.aws.amazon.com/msk/latest/developerguide/msk-password.html), we use signed certificates, so we don't need a store.
We also started to work with Confluent Cloud, doing PLAIN over SASL_SSL, also signed certificates.

Is there a way to setup the Topic Operator over those specific scenarios?

ppatierno:
@scholzj I did a research on issues but even taking a look at the code, we don't have SASL support in the Topic Operator.

Coqueiro:
Since Confluent and MSK already support those authentications methods, this would be a great addition.

scholzj:
If you want, we can convert this to enhancement. But I will be honest with you, from the core maintainers our main focus is to have the ecosystem working around the cluster operator and our own Kafka and not around Amazon MSK. So you should probably not expect this to get implemented any time soon unless someone contributes it. (we can of course provide help and guidance during the implementation)

Coqueiro:
Totally understandable reasoning. Thanks for the clarity, @scholzj !

Would be great to use Strimzi's Topic and User Operator in our current setup, they are good solutions from a DevOps perspective. Sooner of later I may try to implement this.

AlbertoPeon:
Hi @ppatierno,

Does the same apply to the Kafka User operator? We were planning to use both with MSK and SASL_SSL (we have a very similar usecase as OP) but this is a bit of a blocker for us.

ppatierno:
Yes that's the same

neuromantik33:
Would love to use strimzi's connect operator to also be able to connect to confluent's managed kafka cluster using plain SASL_SSL. Of course we plan to be running our own kafka eventually once we get the necessary hardware but in the meantime it would be great to be able to use strimzi for connecting to _foreign_ clusters.

scholzj:
[...]

PR #5201, opened 2021-06-25
feature/topic-operator: Add SASL, SCRAM and Public CA connectivity options

adinhodovic:
### Type of change

- Enhancement / new feature

### Description

Hello Strimzi maintainers,

I'd like to use the Strimzi Operator with an external MSK cluster, however it has a Public CA (need possibility to disable truststore), SCRAM/SASL authentication which requires security protocol change (sasl_ssl), sasl mechanism definitions and JAAS config definitions.

I'll add documentation for all the options if the approach is correct/fine. Not familiar with Java or the various Kafka connectivity/auth methods there are, looking for feedback.

The following environment variables works with MSK and a standalone Topic Operator:

```
# Flag for SSL_ENDPOINT_IDENTIFICATION_ALGORITHM_CONFIG
- name: STRIMZI_TLS_ENABLED
  value: "true"
# Disables truststore generation even if TLS enabled
- name: STRIMZI_PUBLIC_CA
  value: "true"
# Flag for auth/keystore
- name: STRIMZI_TLS_AUTH_ENABLED
  value: "false"
# SASL flag
- name: STRIMZI_SASL_ENABLED
  value: "true"
# SASL mechanism
- name: STRIMZI_SASL_MECHANISM
  value: "scram-sha-512"
# Username
- name: STRIMZI_SASL_USERNAME
  value: "admin"
# Password
- name: STRIMZI_SASL_PASSWORD
  value: "hello strimzi"
# Security protocol (default SSL)
- name: STRIMZI_SECURITY_PROTOCOL
  value: "SASL_SSL"

```

Issues: https://github.com/strimzi/strimzi-kafka-operator/issues/2761 https://github.com/strimzi/strimzi-kafka-operator/issues/4766 https://github.com/strimzi/strimzi-kafka-operator/issues/3741
### Checklist

- [x] Write tests
- [x] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/model/EntityTopicOperator.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/model/EntityTopicOperator.java
@@ -58,7 +58,6 @@ public class EntityTopicOperator extends AbstractModel {
     public static final String ENV_VAR_SECURITY_PROTOCOL = "STRIMZI_SECURITY_PROTOCOL";
 
     public static final String ENV_VAR_TLS_ENABLED = "STRIMZI_TLS_ENABLED";
-    public static final String ENV_VAR_TLS_AUTH_ENABLED = "STRIMZI_TLS_AUTH_ENABLED";
 
     public static final Probe DEFAULT_HEALTHCHECK_OPTIONS = new ProbeBuilder()
             .withInitialDelaySeconds(EntityTopicOperatorSpec.DEFAULT_HEALTHCHECK_DELAY)
@@ -79,7 +78,6 @@ public class EntityTopicOperator extends AbstractModel {
     private int topicMetadataMaxAttempts;
     protected List<ContainerEnvVar> templateContainerEnvVars;
     protected SecurityContext templateContainerSecurityContext;
-    private String securityProtocol;
 
     /**
      * @param reconciliation   The reconciliation
@@ -107,7 +105,6 @@ public class EntityTopicOperator extends AbstractModel {
         this.zookeeperSessionTimeoutMs = EntityTopicOperatorSpec.DEFAULT_ZOOKEEPER_SESSION_TIMEOUT_SECONDS * 1_000;
         this.resourceLabels = ModelUtils.defaultResourceLabels(cluster);
         this.topicMetadataMaxAttempts = EntityTopicOperatorSpec.DEFAULT_TOPIC_METADATA_MAX_ATTEMPTS;
-        this.securityProtocol = EntityTopicOperatorSpec.DEFAULT_SECURITY_PROTOCOL;
 
         this.ancillaryConfigMapName = metricAndLogConfigsName(cluster);
         this.logAndMetricsConfigVolumeName = "entity-topic-operator-metrics-and-logging";
@@ -283,9 +280,8 @@ public class EntityTopicOperator extends AbstractModel {
         varList.add(buildEnvVar(ENV_VAR_FULL_RECONCILIATION_INTERVAL_MS, Integer.toString(reconciliationIntervalMs)));
         varList.add(buildEnvVar(ENV_VAR_ZOOKEEPER_SESSION_TIMEOUT_MS, Integer.toString(zookeeperSessionTimeoutMs)));
         varList.add(buildEnvVar(ENV_VAR_TOPIC_METADATA_MAX_ATTEMPTS, String.valueOf(topicMetadataMaxAttempts)));
-        varList.add(buildEnvVar(ENV_VAR_SECURITY_PROTOCOL, securityProtocol));
+        varList.add(buildEnvVar(ENV_VAR_SECURITY_PROTOCOL, EntityTopicOperatorSpec.DEFAULT_SECURITY_PROTOCOL));
         varList.add(buildEnvVar(ENV_VAR_TLS_ENABLED, Boolean.toString(true)));
-        varList.add(buildEnvVar(ENV_VAR_TLS_AUTH_ENABLED, Boolean.toString(true)));
         varList.add(buildEnvVar(ENV_VAR_STRIMZI_GC_LOG_ENABLED, String.valueOf(gcLoggingEnabled)));
         EntityOperator.javaOptions(varList, getJvmOptions(), javaSystemProperties);
 
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/model/EntityTopicOperator.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/model/EntityTopicOperator.java
@@ -55,7 +55,11 @@ public class EntityTopicOperator extends AbstractModel {
     public static final String ENV_VAR_FULL_RECONCILIATION_INTERVAL_MS = "STRIMZI_FULL_RECONCILIATION_INTERVAL_MS";
     public static final String ENV_VAR_ZOOKEEPER_SESSION_TIMEOUT_MS = "STRIMZI_ZOOKEEPER_SESSION_TIMEOUT_MS";
     public static final String ENV_VAR_TOPIC_METADATA_MAX_ATTEMPTS = "STRIMZI_TOPIC_METADATA_MAX_ATTEMPTS";
+    public static final String ENV_VAR_SECURITY_PROTOCOL = "STRIMZI_SECURITY_PROTOCOL";
+
     public static final String ENV_VAR_TLS_ENABLED = "STRIMZI_TLS_ENABLED";
+    public static final String ENV_VAR_TLS_AUTH_ENABLED = "STRIMZI_TLS_AUTH_ENABLED";
+
     public static final Probe DEFAULT_HEALTHCHECK_OPTIONS = new ProbeBuilder()
             .withInitialDelaySeconds(EntityTopicOperatorSpec.DEFAULT_HEALTHCHECK_DELAY)
             .withTimeoutSeconds(EntityTopicOperatorSpec.DEFAULT_HEALTHCHECK_TIMEOUT).build();
@@ -75,6 +79,7 @@ public class EntityTopicOperator extends AbstractModel {
     private int topicMetadataMaxAttempts;
     protected List<ContainerEnvVar> templateContainerEnvVars;
     protected SecurityContext templateContainerSecurityContext;
+    private String securityProtocol;
 
     /**
      * @param reconciliation   The reconciliation
@@ -102,6 +107,7 @@ public class EntityTopicOperator extends AbstractModel {
         this.zookeeperSessionTimeoutMs = EntityTopicOperatorSpec.DEFAULT_ZOOKEEPER_SESSION_TIMEOUT_SECONDS * 1_000;
         this.resourceLabels = ModelUtils.defaultResourceLabels(cluster);
         this.topicMetadataMaxAttempts = EntityTopicOperatorSpec.DEFAULT_TOPIC_METADATA_MAX_ATTEMPTS;
+        this.securityProtocol = EntityTopicOperatorSpec.DEFAULT_SECURITY_PROTOCOL;
 
         this.ancillaryConfigMapName = metricAndLogConfigsName(cluster);
         this.logAndMetricsConfigVolumeName = "entity-topic-operator-metrics-and-logging";
@@ -277,7 +283,9 @@ public class EntityTopicOperator extends AbstractModel {
         varList.add(buildEnvVar(ENV_VAR_FULL_RECONCILIATION_INTERVAL_MS, Integer.toString(reconciliationIntervalMs)));
         varList.add(buildEnvVar(ENV_VAR_ZOOKEEPER_SESSION_TIMEOUT_MS, Integer.toString(zookeeperSessionTimeoutMs)));
         varList.add(buildEnvVar(ENV_VAR_TOPIC_METADATA_MAX_ATTEMPTS, String.valueOf(topicMetadataMaxAttempts)));
+        varList.add(buildEnvVar(ENV_VAR_SECURITY_PROTOCOL, securityProtocol));
         varList.add(buildEnvVar(ENV_VAR_TLS_ENABLED, Boolean.toString(true)));
+        varList.add(buildEnvVar(ENV_VAR_TLS_AUTH_ENABLED, Boolean.toString(true)));
         varList.add(buildEnvVar(ENV_VAR_STRIMZI_GC_LOG_ENABLED, String.valueOf(gcLoggingEnabled)));
         EntityOperator.javaOptions(varList, getJvmOptions(), javaSystemProperties);
 
--- a/documentation/modules/deploying/proc-deploy-topic-operator-standalone.adoc
+++ b/documentation/modules/deploying/proc-deploy-topic-operator-standalone.adoc
@@ -99,24 +99,25 @@ You can set the level to `ERROR`, `WARNING`, `INFO`, `DEBUG`, or `TRACE`.
 <9> Enables TLS support for encrypted communication with the Kafka brokers.
 <10> (Optional) The Java options used by the JVM running the Topic Operator.
 <11> (Optional) The debugging (`-D`) options set for the Topic Operator.
-<12> (Optional) Skip generating trust store certificates if TLS is enabled. If this setting is enabled then the brokers have to use a public trusted certificate authority for their TLS certificates.
-The default is "false".
-<13> (Optional) Generate key store certificates for mutual TLS authentication. Setting this to false will disable client authentication with TLS to the Kafka brokers.
-The default is "true".
-<14> (Optional) Enable SASL support for client authentication when connecting to Kafka brokers.
-The default is "false".
+<12> (Optional) Skips the generation of trust store certificates if TLS is enabled through `STRIMZI_TLS_ENABLED`. If this environment variable is enabled, the brokers must use a public trusted certificate authority for their TLS certificates.
+The default is `false`.
+<13> (Optional) Generates key store certificates for mutual TLS authentication. Setting this to `false` disables client authentication with TLS to the Kafka brokers.
+The default is `true`.
+<14> (Optional) Enables SASL support for client authentication when connecting to Kafka brokers.
+The default is `false`.
 <15> (Optional) The SASL username for client authentication.
 Mandatory only if SASL is enabled through `STRIMZI_SASL_ENABLED`.
 <16> (Optional) The SASL password for client authentication.
 Mandatory only if SASL is enabled through `STRIMZI_SASL_ENABLED`.
 <17> (Optional) The SASL mechanism for client authentication.
 Mandatory only if SASL is enabled through `STRIMZI_SASL_ENABLED`.
-The value can be set to: `plain`, `scram-sha-256` and `scram-sha-512`.
+You can set the value to `plain`, `scram-sha-256`, or `scram-sha-512`.
 <18> (Optional) The security protocol used for communication with Kafka brokers.
 The default value is "PLAINTEXT".
-The value can be set to: `PLAINTEXT`, `SSL`, `SASL_PLAINTEXT` and `SASL_SSL`.
+You can set the value to `PLAINTEXT`, `SSL`, `SASL_PLAINTEXT`, or `SASL_SSL`.
 
 . If you enabled TLS with the `STRIMZI_TLS_ENABLED` environment variable, specify the keystore and truststore used to authenticate connection to the Kafka cluster.
+. If you want to connect to Kafka brokers which are using certificates from public CAs as for example AWS MSK then you'll need to set `STRIMZI_PUBLIC_CA` to `true`.
 +
 .Example TLS configuration
 [source,shell,subs=+quotes]
--- a/documentation/modules/deploying/proc-deploy-topic-operator-standalone.adoc
+++ b/documentation/modules/deploying/proc-deploy-topic-operator-standalone.adoc
@@ -101,8 +101,8 @@ You can set the level to `ERROR`, `WARNING`, `INFO`, `DEBUG`, or `TRACE`.
 <11> (Optional) The debugging (`-D`) options set for the Topic Operator.
 <12> (Optional) Skip generating trust store certificates if TLS is enabled. If this setting is enabled then the brokers have to use a public trusted certificate authority for their TLS certificates.
 The default is "false".
-<13> (Optional) Skip generating key store certificates for mutual TLS authentication. This will disable client authentication with TLS to the Kafka brokers.
-The default is "false".
+<13> (Optional) Generate key store certificates for mutual TLS authentication. Setting this to false will disable client authentication with TLS to the Kafka brokers.
+The default is "true".
 <14> (Optional) Enable SASL support for client authentication when connecting to Kafka brokers.
 The default is "false".
 <15> (Optional) The SASL username for client authentication.
--- a/documentation/modules/deploying/proc-deploy-topic-operator-standalone.adoc
+++ b/documentation/modules/deploying/proc-deploy-topic-operator-standalone.adoc
@@ -65,6 +65,20 @@ spec:
               value: "-Xmx=512M -Xms=256M"
             - name: STRIMZI_JAVA_SYSTEM_PROPERTIES <11>
               value: "-Djavax.net.debug=verbose -DpropertyName=value"
+            - name: STRIMZI_PUBLIC_CA <12>
+              value: "false"
+            - name: STRIMZI_TLS_AUTH_ENABLED <13>
+              value: "false"
+            - name: STRIMZI_SASL_ENABLED <14>
+              value: "false"
+            - name: STRIMZI_SASL_USERNAME <15>
+              value: "admin"
+            - name: STRIMZI_SASL_PASSWORD <16>
+              value: "password"
+            - name: STRIMZI_SASL_MECHANISM <17>
+              value: "scram-sha-512"
+            - name: STRIMZI_SECURITY_PROTOCOL <18>
+              value: "SSL"
 ----
 <1> The Kubernetes namespace for the Topic Operator to watch for `KafkaTopic` resources. Specify the namespace of the Kafka cluster.
 <2> The host and port pair of the bootstrap broker address to discover and connect to all brokers in the Kafka cluster.
@@ -85,6 +99,22 @@ You can set the level to `ERROR`, `WARNING`, `INFO`, `DEBUG`, or `TRACE`.
 <9> Enables TLS support for encrypted communication with the Kafka brokers.
 <10> (Optional) The Java options used by the JVM running the Topic Operator.
 <11> (Optional) The debugging (`-D`) options set for the Topic Operator.
+<12> (Optional) Skip generating trust store certificates if TLS is enabled. If this setting is enabled then the brokers have to use a public trusted certificate authority for their TLS certificates.
+The default is "false".
+<13> (Optional) Skip generating key store certificates for mutual TLS authentication. This will disable client authentication with TLS to the Kafka brokers.
+The default is "false".
+<14> (Optional) Enable SASL support for client authentication when connecting to Kafka brokers.
+The default is "false".
+<15> (Optional) The SASL username for client authentication.
+Mandatory only if SASL is enabled through `STRIMZI_SASL_ENABLED`.
+<16> (Optional) The SASL password for client authentication.
+Mandatory only if SASL is enabled through `STRIMZI_SASL_ENABLED`.
+<17> (Optional) The SASL mechanism for client authentication.
+Mandatory only if SASL is enabled through `STRIMZI_SASL_ENABLED`.
+The value can be set to: `plain`, `scram-sha-256` and `scram-sha-512`.
+<18> (Optional) The security protocol used for communication with Kafka brokers.
+The default value is "PLAINTEXT".
+The value can be set to: `PLAINTEXT`, `SSL`, `SASL_PLAINTEXT` and `SASL_SSL`.
 
 . If you enabled TLS with the `STRIMZI_TLS_ENABLED` environment variable, specify the keystore and truststore used to authenticate connection to the Kafka cluster.
 +
--- a/documentation/modules/deploying/proc-deploy-topic-operator-standalone.adoc
+++ b/documentation/modules/deploying/proc-deploy-topic-operator-standalone.adoc
@@ -116,8 +116,8 @@ You can set the value to `plain`, `scram-sha-256`, or `scram-sha-512`.
 The default value is "PLAINTEXT".
 You can set the value to `PLAINTEXT`, `SSL`, `SASL_PLAINTEXT`, or `SASL_SSL`.
 
+. If you want to connect to Kafka brokers that are using certificates from a public certificate authority, set `STRIMZI_PUBLIC_CA` to `true`. Set this property to `true`, for example, if you are using Kafka on Amazon Web Services (AWS).
 . If you enabled TLS with the `STRIMZI_TLS_ENABLED` environment variable, specify the keystore and truststore used to authenticate connection to the Kafka cluster.
-. If you want to connect to Kafka brokers which are using certificates from public CAs as for example AWS MSK then you'll need to set `STRIMZI_PUBLIC_CA` to `true`.
 +
 .Example TLS configuration
 [source,shell,subs=+quotes]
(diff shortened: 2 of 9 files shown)


scholzj (commented):
Thanks for the PR. 

I left few smaller comments. I think this also breaks the non-standalone mode, because you added `STRIMZI_TLS_AUTH_ENABLED` environment variable but it is not set by the Cluster Operator when it deploys the Topic Operator. I think you would need to add it here: https://github.com/strimzi/strimzi-kafka-operator/blob/3b3a419dd17cac0458b2a96d3d2b6f9d47683ebe/cluster-operator/src/main/java/io/strimzi/operator/cluster/model/EntityTopicOperator.java#L271 (plus there will be some unit tests for it I guess)

I think the overall idea is definitely welcomed addition. But the environment variables basically form an API which ideally should not change later. So I think we also need to think a bit more whether this is the right API. But it is for sure in the right direction.

topic-operator/src/main/java/io/strimzi/operator/topic/Session.java
```diff
@@ -153,12 +157,40 @@ public void start(Promise<Void> start) {
...
+        if (Boolean.parseBoolean(config.get(Config.SASL_ENABLED))) {
+            String saslMechanism = null;
+            String jaasConfig = null;
+            String username = config.get(Config.SASL_USERNAME);
+            String password = config.get(Config.SASL_PASSWORD);
+
+            if (KafkaClientAuthenticationPlain.TYPE_PLAIN.equals(config.get(Config.SASL_MECHANISM))) {
+                saslMechanism = "PLAIN";
+                jaasConfig = "org.apache.kafka.common.security.plain.PlainLoginModule required username=\"" + username + "\" password=\"" + password + "\";";
+            } else if (KafkaClientAuthenticationScramSha512.TYPE_SCRAM_SHA_512.equals(config.get(Config.SASL_MECHANISM))) {
```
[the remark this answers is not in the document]
adinhodovic: Check updated changes in `Session.java`.

topic-operator/src/main/java/io/strimzi/operator/topic/Session.java
```diff
@@ -153,12 +157,40 @@ public void start(Promise<Void> start) {
         kafkaClientProps.setProperty(StreamsConfig.APPLICATION_ID_CONFIG, config.get(Config.APPLICATION_ID));
 
         if (Boolean.parseBoolean(config.get(Config.TLS_ENABLED))) {
-            kafkaClientProps.setProperty(AdminClientConfig.SECURITY_PROTOCOL_CONFIG, "SSL");
-            kafkaClientProps.setProperty(SslConfigs.SSL_TRUSTSTORE_LOCATION_CONFIG, config.get(Config.TLS_TRUSTSTORE_LOCATION));
-            kafkaClientProps.setProperty(SslConfigs.SSL_TRUSTSTORE_PASSWORD_CONFIG, config.get(Config.TLS_TRUSTSTORE_PASSWORD));
-            kafkaClientProps.setProperty(SslConfigs.SSL_KEYSTORE_LOCATION_CONFIG, config.get(Config.TLS_KEYSTORE_LOCATION));
-            kafkaClientProps.setProperty(SslConfigs.SSL_KEYSTORE_PASSWORD_CONFIG, config.get(Config.TLS_KEYSTORE_PASSWORD));
+            kafkaClientProps.setProperty(AdminClientConfig.SECURITY_PROTOCOL_CONFIG, config.get(config.SECURITY_PROTOCOL));
```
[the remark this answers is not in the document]
adinhodovic: I think we would need to move `SECURITY_PROTOCOL_CONFIG` outside of the `TLS_AUTH` gate. Check the added commits. I've also added defaults for the EntityTopicOperator where TLS and TLS AUTH is enabled and `SECURITY_PROTOCOL` is set to "SSL".

adinhodovic:
@scholzj Thanks for the initial review.

> I left few smaller comments. I think this also breaks the non-standalone mode, because you added STRIMZI_TLS_AUTH_ENABLED environment variable but it is not set by the Cluster Operator when it deploys the Topic Operator. I think you would need to add it here: 

Does the scope of the PR also cover the intention of adding SASL/SCRAM auth to the TopicOperator entity, i.e that you can configure SCRAM and SASL in the API spec of the K8s resource? Or is it only to not break current configs by adding sane/working default vars to the TopicOperator.

scholzj:
> Does the scope of the PR also cover the intention of adding SASL/SCRAM auth to the TopicOperator entity, i.e that you can configure SCRAM and SASL in the API spec of the K8s resource? Or is it only to not break current configs by adding sane/working default vars to the TopicOperator. I've done the latter one I believe, for the first one pointers would be good if needed.

No, the SASL does not need to be added there - the internal listeners used by Strimzi do not use it. So all was needed was to make sure that what worked before will still work.

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/EntityTopicOperator.java
```diff
@@ -277,7 +283,9 @@ public static EntityTopicOperator fromCrd(Reconciliation reconciliation, Kafka k
         varList.add(buildEnvVar(ENV_VAR_FULL_RECONCILIATION_INTERVAL_MS, Integer.toString(reconciliationIntervalMs)));
         varList.add(buildEnvVar(ENV_VAR_ZOOKEEPER_SESSION_TIMEOUT_MS, Integer.toString(zookeeperSessionTimeoutMs)));
         varList.add(buildEnvVar(ENV_VAR_TOPIC_METADATA_MAX_ATTEMPTS, String.valueOf(topicMetadataMaxAttempts)));
+        varList.add(buildEnvVar(ENV_VAR_SECURITY_PROTOCOL, securityProtocol));
```
scholzj, line 286: Given the value is always the same, maybe we can just use `DEFAULT_SECURITY_PROTOCOL` here and just remove the `securityProtocol` field?

scholzj (commented):
One nit, but LGTM otherwise.

topic-operator/src/main/java/io/strimzi/operator/topic/Session.java
```diff
@@ -151,14 +156,42 @@ public void start(Promise<Void> start) {
...
+            String username = config.get(Config.SASL_USERNAME);
+            String password = config.get(Config.SASL_PASSWORD);
+
+            if (SASL_TYPE_PLAIN.equals(config.get(Config.SASL_MECHANISM))) {
+                saslMechanism = "PLAIN";
+                jaasConfig = "org.apache.kafka.common.security.plain.PlainLoginModule required username=\"" + username + "\" password=\"" + password + "\";";
+            } else if (SASL_TYPE_SCRAM_SHA_512.equals(config.get(Config.SASL_MECHANISM))) {
+                saslMechanism = "SCRAM-SHA-512";
+                jaasConfig = "org.apache.kafka.common.security.scram.ScramLoginModule required username=\"" + username + "\" password=\"" + password + "\";";
+            }
```
tombentley, line 187: I guess:

* this should be factored into a separate method (which takes the `kafkaClientProps`) and adds to it.
* We should support SCRAM-SHA-256 (even though it's not supported by the Cluster Operator, some people use the Topic Operator standalone, and there's no harm in letting them use 256).
* Throw some meaningful exception and message in a final `else` clause.

topic-operator/src/main/java/io/strimzi/operator/topic/Session.java
```diff
@@ -151,14 +156,42 @@ public void start(Promise<Void> start) {
...
-            kafkaClientProps.setProperty(SslConfigs.SSL_TRUSTSTORE_LOCATION_CONFIG, config.get(Config.TLS_TRUSTSTORE_LOCATION));
-            kafkaClientProps.setProperty(SslConfigs.SSL_TRUSTSTORE_PASSWORD_CONFIG, config.get(Config.TLS_TRUSTSTORE_PASSWORD));
-            kafkaClientProps.setProperty(SslConfigs.SSL_KEYSTORE_LOCATION_CONFIG, config.get(Config.TLS_KEYSTORE_LOCATION));
-            kafkaClientProps.setProperty(SslConfigs.SSL_KEYSTORE_PASSWORD_CONFIG, config.get(Config.TLS_KEYSTORE_PASSWORD));
             kafkaClientProps.setProperty(SslConfigs.SSL_ENDPOINT_IDENTIFICATION_ALGORITHM_CONFIG, config.get(Config.TLS_SSL_ENDPOINT_IDENTIFICATION_ALGORITHM));
+
+            if (!config.get(Config.TLS_TRUSTSTORE_LOCATION).isEmpty() && !config.get(Config.TLS_TRUSTSTORE_PASSWORD).isEmpty()) {
+                kafkaClientProps.setProperty(SslConfigs.SSL_TRUSTSTORE_LOCATION_CONFIG, config.get(Config.TLS_TRUSTSTORE_LOCATION));
+                kafkaClientProps.setProperty(SslConfigs.SSL_TRUSTSTORE_PASSWORD_CONFIG, config.get(Config.TLS_TRUSTSTORE_PASSWORD));
+            }
```
tombentley, line 167: Shouldn't we just have two `if` statements here?

topic-operator/src/main/java/io/strimzi/operator/topic/Session.java
```diff
@@ -151,14 +156,42 @@ public void start(Promise<Void> start) {
...
+
+            if (Boolean.parseBoolean(config.get(Config.TLS_AUTH_ENABLED))) {
```
tombentley, line 169: Do we really need `TLS_AUTH_ENABLED`? Couldn't we only add the keystore location and password when they're provided? That would be more consistent with how we're handling the truststore.

topic-operator/src/main/java/io/strimzi/operator/topic/Config.java
```diff
@@ -179,6 +189,16 @@ private Value(String key, Type<? extends T> type, boolean required) {
...
+    /** If SASL should be used to authenticate with */
+    public static final Value<String> SASL_ENABLED = new Value<>(TC_SASL_ENABLED, STRING, "false");
+    /** The SASL mechanism to be used */
+    public static final Value<String> SASL_MECHANISM = new Value<>(TC_SASL_MECHANISM, STRING, "");
+    /** The SASL username to be used for authentication */
+    public static final Value<String> SASL_USERNAME = new Value<>(TC_SASL_USERNAME, STRING, "");
+    /** The SASL password to be used for authentication */
+    public static final Value<String> SASL_PASSWORD = new Value<>(TC_SASL_PASSWORD, STRING, "");
+    /** The security protocol to be used */
+    public static final Value<String> SECURITY_PROTOCOL = new Value<>(TC_SECURITY_PROTOCOL, STRING, "PLAINTEXT");
```
tombentley, line 201: Adding the security protocol, which can take the values `PLAINTEXT`, `SSL`, `SASL_PLAINTEXT` and `SASL_SSL`, means there's some duplication between this and the existing `TLS_ENABLED` config. I think we should give `TLS_ENABLED` no default here and in `Session` we should:

1. If `SECURITY_PROTOCOL` and `TLS_ENABLED` are both given ensure they're consistent and use the `SECURITY_PROTOCOL`
2. If only `SECURITY_PROTOCOL` is given then use it.
3. If only `TLS_ENABLED` is enabled then use `SSL` or `PLAINTEXT`
4. If neither are given then use `SSL`

We could deprecate the `TLS_ENABLED` at the same time (but would continue to configure both `SECURITY_PROTOCOL` and `TLS_ENABLED` in the EntityTopicOperator for now). Wdyt @scholzj @sknot-rh?

tombentley (commented):
This looks mostly reasonable to me. I left a few comments. I also think there would be some changes to the docs needed, to document the new options.

topic-operator/src/main/java/io/strimzi/operator/topic/Session.java
```diff
@@ -151,14 +156,42 @@ public void start(Promise<Void> start) {
...
+        }
+
+        if (Boolean.parseBoolean(config.get(Config.SASL_ENABLED))) {
+            String saslMechanism = null;
+            String jaasConfig = null;
+            String username = config.get(Config.SASL_USERNAME);
+            String password = config.get(Config.SASL_PASSWORD);
+
+            if (SASL_TYPE_PLAIN.equals(config.get(Config.SASL_MECHANISM))) {
+                saslMechanism = "PLAIN";
```
sknot-rh, line 182: Cannot we just use `SASL_TYPE_PLAIN`?

topic-operator/src/main/java/io/strimzi/operator/topic/Session.java
```diff
@@ -37,6 +38,9 @@
 
     private final static Logger LOGGER = LogManager.getLogger(Session.class);
 
+    private final static String SASL_TYPE_PLAIN = "PLAIN";
+    private final static String SASL_TYPE_SCRAM_SHA_512 = "PLAIN";
```
sknot-rh, line 42: Is this a typo? Should it be `SCRAM-SHA-512`?

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/EntityTopicOperator.java
adinhodovic: Should be fixed.

topic-operator/src/main/java/io/strimzi/operator/topic/Session.java
adinhodovic: Extracted it to a separate function and added 256 as an option.

topic-operator/src/main/java/io/strimzi/operator/topic/Session.java
adinhodovic: Not following? Two separate if statements? Should they not have both set?

topic-operator/src/main/java/io/strimzi/operator/topic/Session.java
adinhodovic: Yeah, makes sense. Removed it, however I still use `TLS_AUTH_ENABLED` to not generate key store certs which also sets the variables `TLS_KEYSTORE_LOCATION` and `TLS_KEYSTORE_PASSWORD`. Also, there's a usage for `PUBLIC_CA` which disables generation of `TLS_TRUSTORE_LOCATION`. By default both Keystore and Truststore are set by the `prepare_certificates.sh` script.

topic-operator/src/main/java/io/strimzi/operator/topic/Session.java
adinhodovic: Yes, typo.

topic-operator/src/main/java/io/strimzi/operator/topic/Session.java
adinhodovic: No we cannot, the configuration takes lower case sasl mechanism and the Kafka config accepts upper case. I think the lower case is aligned with the rest of the strimzi code.

topic-operator/src/main/java/io/strimzi/operator/topic/Config.java
adinhodovic: Should I proceed with this?

adinhodovic:
> This looks mostly reasonable to me. I left a few comments. I also think there would be some changes to the docs needed, to document the new options.

Thanks for the reviews! I've added documentation.

topic-operator/src/main/java/io/strimzi/operator/topic/Session.java
tombentley: I don't think a truststore _has_ to have a password. I suppose supplying a password without a location should be an error.

topic-operator/src/main/java/io/strimzi/operator/topic/Config.java
tombentley: Yes. @scholzj @sknot-rh can speak up if they disagree.

documentation/modules/deploying/proc-deploy-topic-operator-standalone.adoc
```diff
@@ -34,6 +34,22 @@ Default `20000` (20 seconds).
 Default `/strimzi/topics`.
 .. `STRIMZI_TLS_ENABLED` to enable TLS support for encrypting the communication with Kafka brokers.
 Default `true`.
+.. `STRIMZI_PUBLIC_CA` to skip generating trust store certificates if TLS is enabled.
```
tombentley, line 37: Let's explain the effect that this will have on the operator.

documentation/modules/deploying/proc-deploy-topic-operator-standalone.adoc
```diff
@@ -34,6 +34,22 @@ Default `20000` (20 seconds).
...
+Default `false`.
+.. `STRIMZI_TLS_AUTH_ENABLED` to skip generating key store certificates if TLS is enabled.
```
tombentley, line 39: Again, what does that mean for the operator?

tombentley (approved):
@adinhodovic I took the liberty of factoring out a method for building the client props, hope that's OK.

LGTM, thanks!

adinhodovic:
> @adinhodovic I took the liberty of factoring out a method for building the client props, hope that's OK.
> 
> LGTM, thanks!

@tombentley Understandable :D and thanks!

Two questions:

1) I rebased and saw that there are two docs now, one for standalone operator and the other for the cluster default topic operator. Should I add the docs to the cluster operator too? I believe @scholzj mentioned that this should not be configurable or have any impact on the strimzi-operator.

2) Would there be OK for the maintainers if we add the standalone-topic-operator to the Helm chart. By default it is disabled but a flag would deploy the topic-operator as a deployment. This would only be a extra file(RBAC, serviceaccount and CRDs can be reused) and it could make use of the helm chart version updates and CRDs for topics. We use the strimzi-operator for kafka-connect with an external cluster and would be nice to use the helm chart to deploy the strimzi-operator and standalone-topic-operator than maintaining a chart for the standalone-topic-operator ourselves. Understandable if not preferred!

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/EntityTopicOperator.java
```diff
@@ -277,6 +280,7 @@ public static EntityTopicOperator fromCrd(Reconciliation reconciliation, Kafka k
         varList.add(buildEnvVar(ENV_VAR_FULL_RECONCILIATION_INTERVAL_MS, Integer.toString(reconciliationIntervalMs)));
         varList.add(buildEnvVar(ENV_VAR_ZOOKEEPER_SESSION_TIMEOUT_MS, Integer.toString(zookeeperSessionTimeoutMs)));
         varList.add(buildEnvVar(ENV_VAR_TOPIC_METADATA_MAX_ATTEMPTS, String.valueOf(topicMetadataMaxAttempts)));
+        varList.add(buildEnvVar(ENV_VAR_SECURITY_PROTOCOL, EntityTopicOperatorSpec.DEFAULT_SECURITY_PROTOCOL));
```
scholzj, line 283: Maybe I missed something. But don't we need to also set `STRIMZI_TLS_AUTH_ENABLED` here?

scholzj:
> Would there be OK for the maintainers if we add the standalone-topic-operator to the Helm chart. By default it is disabled but a flag would deploy the topic-operator as a deployment. This would only be a extra file(RBAC, serviceaccount and CRDs can be reused) and it could make use of the helm chart version updates and CRDs for topics that are automated within the repo I believe. We use the strimzi-operator for kafka-connect with an external cluster and would be nice to use the helm chart to deploy the strimzi-operator and standalone-topic-operator than maintaining a chart for the standalone-topic-operator ourselves. Understandable if not preferred!

I'm not a Helm user - how does it work with multiple different deployments in a single Helm Chart? Wouldn't it be better to have a separate Helm Chart for the Topic Operator. (In general, I don't think it is a problem to add it if you can contribute it and ideally help to maintain it since nobody from the core team knows much about Helm Charts), but trying to understand what is the best way.

> I rebased and saw that there are two docs now, one for standalone operator and the other for the cluster default topic operator. Should I add the docs to the cluster operator too? I believe @scholzj mentioned that this should not be configurable or have any impact on the strimzi-operator.

For the users of the Cluster Operator, this should IMHO be just an implementation detail. It does not add any new options. So it should be only the standalone docs what matters.

adinhodovic:
> > Would there be OK for the maintainers if we add the standalone-topic-operator to the Helm chart. By default it is disabled but a flag would deploy the topic-operator as a deployment. This would only be a extra file(RBAC, serviceaccount and CRDs can be reused) and it could make use of the helm chart version updates and CRDs for topics that are automated within the repo I believe. We use the strimzi-operator for kafka-connect with an external cluster and would be nice to use the helm chart to deploy the strimzi-operator and standalone-topic-operator than maintaining a chart for the standalone-topic-operator ourselves. Understandable if not preferred!
> 
> I'm not a Helm user - how does it work with multiple different deployments in a single Helm Chart? Wouldn't it be better to have a separate Helm Chart for the Topic Operator. (In general, I don't think it is a problem to add it if you can contribute it and ideally help to maintain it since nobody from the core team knows much about Helm Charts), but trying to understand what is the best way.
> 
> > I rebased and saw that there are two docs now, one for standalone operator and the other for the cluster default topic operator. Should I add the docs to the cluster operator too? I believe @scholzj mentioned that this should not be configurable or have any impact on the strimzi-operator.
> 
> For the users of the Cluster Operator, this should IMHO be just an implementation detail. It does not add any new options. So it should be only the standalone docs what matters.

Ah, yeah did not think of multiple deployments. Can contribute with a standalone Helm chart :+1:

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/EntityTopicOperator.java
adinhodovic: I think you are somewhat correct (I've misconfigured it but should be fixed in latest commit). `STRIMZI_TLS_AUTH_ENABLED` and `STRIMZI_PUBLIC_CA` only gate if certificates should be generated, they're not used in the operator. The issue I ran into is that we need to explicitly set the generation of truststore and keystore to false otherwise the `tls_operator_run.sh` will generate them by checking if for example `STRIMZI_TRUSTORE_LOCATION` is undefined(`[ -z "$STRIMZI_TRUSTSTORE_LOCATION" ]` ) which always defaults to that when the string is empty. This conditional sets the `STRIMZI_TRUSTSTORE_LOCATION` and related keystore/truststore variables which we don't want with an external cluster. They're not used in the topic operator anymore, only in the sh scripts `topic_operator_run.sh`. 

They're used here:

https://github.com/strimzi/strimzi-kafka-operator/blob/bbef892e8f398098c45f23a9cfb0344ecf7cf924/topic-operator/scripts/topic_operator_run.sh#L27-L39 to avoid setting the keystore/truststore variables. And they skip setting/generating stores when the `PUBLIC_CA` is true and `TLS_AUTH_ENABLED` is false. In all other cases the checks should not affect anything. Let me know if there is a better/cleaner way.

In the operator we just check if the truststore/keystore locations/passwords are not empty: https://github.com/strimzi/strimzi-kafka-operator/pull/5201/files#diff-0ed74d5df887eaebfe8c8039af02c2bd34e16cf69e9a094bd0814a6879b58e94R325-R339

topic-operator/src/main/java/io/strimzi/operator/topic/Session.java
```diff
@@ -267,6 +297,54 @@ public void handle(Long oldTimerId) {
...
+    private Properties adminClientProperties() {
+        Properties kafkaClientProps = new Properties();
+        kafkaClientProps.setProperty(AdminClientConfig.BOOTSTRAP_SERVERS_CONFIG, config.get(Config.KAFKA_BOOTSTRAP_SERVERS));
+        kafkaClientProps.setProperty(StreamsConfig.APPLICATION_ID_CONFIG, config.get(Config.APPLICATION_ID));
+
+        String securityProtocol = config.get(Config.SECURITY_PROTOCOL);
+        boolean tlsEnabled = Boolean.parseBoolean(config.get(Config.TLS_ENABLED));
+
+        if (tlsEnabled && !securityProtocol.isEmpty()) {
+            if (!securityProtocol.equals("SASL_SSL") || !securityProtocol.equals("SSL")) {
```
scholzj, line 309: @tombentley @adinhodovic I think this should be `&&`? IMHO this is what failed all the system tests.

scholzj (commented):
@tombentley @adinhodovic I do not know the TO that much. But two thoughts:
* Can this be unit tested?
* As we are changing the _API_ of the standalone TO, should we move to use PEM files instead of JKS/PKCS12 files? (maybe we can do it in a separate PR as this would be a very different change from what @adinhodovic really started with) I wonder if that could simplify the configuration by moving more of the logic from the shell script into the Java code and make it easier to unit test it.

documentation/modules/deploying/proc-deploy-topic-operator-standalone.adoc
```diff
@@ -85,6 +99,22 @@ You can set the level to `ERROR`, `WARNING`, `INFO`, `DEBUG`, or `TRACE`.
 <9> Enables TLS support for encrypted communication with the Kafka brokers.
 <10> (Optional) The Java options used by the JVM running the Topic Operator.
 <11> (Optional) The debugging (`-D`) options set for the Topic Operator.
+<12> (Optional) Skip generating trust store certificates if TLS is enabled. If this setting is enabled then the brokers have to use a public trusted certificate authority for their TLS certificates.
```
PaulRMellor, line 102:
```suggestion
<12> (Optional) Skips the generation of trust store certificates if TLS is enabled. If this environment variable is enabled, the brokers must use a public trusted certificate authority for their TLS certificates.
```

documentation/modules/deploying/proc-deploy-topic-operator-standalone.adoc
```diff
@@ -85,6 +99,22 @@ You can set the level to `ERROR`, `WARNING`, `INFO`, `DEBUG`, or `TRACE`.
...
+The default is "false".
```
PaulRMellor, line 103:
```suggestion
The default is `false`.
```

documentation/modules/deploying/proc-deploy-topic-operator-standalone.adoc
```diff
@@ -85,6 +99,22 @@ You can set the level to `ERROR`, `WARNING`, `INFO`, `DEBUG`, or `TRACE`.
...
+<13> (Optional) Generate key store certificates for mutual TLS authentication. Setting this to false will disable client authentication with TLS to the Kafka brokers.
+The default is "true".
```
PaulRMellor, line 105:
```suggestion
<13> (Optional) Generates key store certificates for mutual TLS authentication. Setting this to `false` disables client authentication with TLS to the Kafka brokers.
The default is `true`.
```

documentation/modules/deploying/proc-deploy-topic-operator-standalone.adoc
```diff
@@ -85,6 +99,22 @@ You can set the level to `ERROR`, `WARNING`, `INFO`, `DEBUG`, or `TRACE`.
...
+<14> (Optional) Enable SASL support for client authentication when connecting to Kafka brokers.
+The default is "false".
```
PaulRMellor, line 107:
```suggestion
<14> (Optional) Enables SASL support for client authentication when connecting to Kafka brokers.
The default is `false`.
```

documentation/modules/deploying/proc-deploy-topic-operator-standalone.adoc
```diff
@@ -85,6 +99,22 @@ You can set the level to `ERROR`, `WARNING`, `INFO`, `DEBUG`, or `TRACE`.
...
+<15> (Optional) The SASL username for client authentication.
+Mandatory only if SASL is enabled through `STRIMZI_SASL_ENABLED`.
+<16> (Optional) The SASL password for client authentication.
+Mandatory only if SASL is enabled through `STRIMZI_SASL_ENABLED`.
+<17> (Optional) The SASL mechanism for client authentication.
+Mandatory only if SASL is enabled through `STRIMZI_SASL_ENABLED`.
+The value can be set to: `plain`, `scram-sha-256` and `scram-sha-512`.
```
PaulRMellor, line 114:
```suggestion
You can set the value to `plain`, `scram-sha-256`, or `scram-sha-512`.
```

documentation/modules/deploying/proc-deploy-topic-operator-standalone.adoc
```diff
@@ -85,6 +99,22 @@ You can set the level to `ERROR`, `WARNING`, `INFO`, `DEBUG`, or `TRACE`.
...
+<18> (Optional) The security protocol used for communication with Kafka brokers.
+The default value is "PLAINTEXT".
+The value can be set to: `PLAINTEXT`, `SSL`, `SASL_PLAINTEXT` and `SASL_SSL`.
```
PaulRMellor, line 117:
```suggestion
You can set the value to `PLAINTEXT`, `SSL`, `SASL_PLAINTEXT`, or `SASL_SSL`.
```

documentation/modules/deploying/proc-deploy-topic-operator-standalone.adoc
```diff
@@ -85,6 +99,22 @@ You can set the level to `ERROR`, `WARNING`, `INFO`, `DEBUG`, or `TRACE`.
...
 
 . If you enabled TLS with the `STRIMZI_TLS_ENABLED` environment variable, specify the keystore and truststore used to authenticate connection to the Kafka cluster.
```
PaulRMellor, line 119: Do we need to update this to mention `STRIMZI_PUBLIC_CA`?  Would we better moving the new optional connectivity environment variables into another separate step so we can give the variables a bit of context? _If you want to connect...use these variables..._

PaulRMellor (commented):
Left a few suggestions for the doc

topic-operator/src/main/java/io/strimzi/operator/topic/Session.java
tombentley: Yep

tombentley:
> Can this be unit tested?

Simplest way would be to make `adminClientProperties` package access and assert the Properties it returns for different configurations. Personally I think that would be sufficient.

> As we are changing the API of the standalone TO, should we move to use PEM files instead of JKS/PKCS12 files?

If we're going to do that it should be a separate PR I think.

adinhodovic:
Added unit tests and changed the docs according to the review, but still lacking coverage on tests for the `.sh` script. Two env vars are used to avoid generating and setting keystore/truststore variables. . Would prefer to skip the PEM12 changes if possible. LMK if anything is lacking/wrong approach.

scholzj:
Thanks for the changes. I will have a look at them. I also started the system tests to see if they pass now.

scholzj (approved):

sknot-rh (approved):

documentation/modules/deploying/proc-deploy-topic-operator-standalone.adoc
```diff
@@ -85,8 +99,25 @@ You can set the level to `ERROR`, `WARNING`, `INFO`, `DEBUG`, or `TRACE`.
...
+Mandatory only if SASL is enabled through `STRIMZI_SASL_ENABLED`.
+<17> (Optional) The SASL mechanism for client authentication.
+Mandatory only if SASL is enabled through `STRIMZI_SASL_ENABLED`.
+You can set the value to `plain`, `scram-sha-256`, or `scram-sha-512`.
+<18> (Optional) The security protocol used for communication with Kafka brokers.
+The default value is "PLAINTEXT".
+You can set the value to `PLAINTEXT`, `SSL`, `SASL_PLAINTEXT`, or `SASL_SSL`.
 
 . If you enabled TLS with the `STRIMZI_TLS_ENABLED` environment variable, specify the keystore and truststore used to authenticate connection to the Kafka cluster.
+. If you want to connect to Kafka brokers which are using certificates from public CAs as for example AWS MSK then you'll need to set `STRIMZI_PUBLIC_CA` to `true`.
```
PaulRMellor, line 120:
```suggestion
. If you want to connect to Kafka brokers that are using certificates from public CAs, set `STRIMZI_PUBLIC_CA` to `true`. Set this property to `true`, for example, if you are using Kafka on Amazon Web Services (AWS).
```

documentation/modules/deploying/proc-deploy-topic-operator-standalone.adoc
```diff
@@ -85,8 +99,25 @@ You can set the level to `ERROR`, `WARNING`, `INFO`, `DEBUG`, or `TRACE`.
...
 +
 .Example TLS configuration
```
PaulRMellor, line 122: Should step 2 and 3 swap places so that the example truststore confif sits with _If you enabled TLS with the STRIMZI_TLS_ENABLED environment variable,..._

PaulRMellor: Could you remove the quote from the example config?
` # ..."` - > `# ...`

PaulRMellor (commented):
I think steps 2 and 3 should swap round so the example trusctore/keystore config sits with _If you enabled TLS with the STRIMZI_TLS_ENABLED environment variable_

PaulRMellor (approved):
Looks good. Thanks for the doc changes. :+1:

scholzj:
Thanks for the PR!

PR #5215, opened 2021-06-29
Fix Helm install type FeatureGates variables

michalxo:
- Bugfix

### Description
There was a problem with non-escaped comma in new variable `STRIMZI_FEATURE_GATES` while using helm as CO installation type. This fixes the issue.

### Checklist

### DIFF
--- a/systemtest/src/main/java/io/strimzi/systemtest/resources/specific/HelmResource.java
+++ b/systemtest/src/main/java/io/strimzi/systemtest/resources/specific/HelmResource.java
@@ -126,7 +126,8 @@ public class HelmResource implements SpecificResourceType {
                 entry("logLevelOverride", Environment.STRIMZI_LOG_LEVEL),
                 entry("fullReconciliationIntervalMs", Long.toString(reconciliationInterval)),
                 entry("operationTimeoutMs", Long.toString(operationTimeout)),
-                entry("featureGates", Environment.STRIMZI_FEATURE_GATES),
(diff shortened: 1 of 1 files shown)


scholzj:
Can you please fix the DCO signoff @michalxo?

systemtest/src/main/java/io/strimzi/systemtest/resources/specific/HelmResource.java
```diff
@@ -126,7 +126,7 @@ private void clusterOperator(long operationTimeout, long reconciliationInterval)
                 entry("logLevelOverride", Environment.STRIMZI_LOG_LEVEL),
                 entry("fullReconciliationIntervalMs", Long.toString(reconciliationInterval)),
                 entry("operationTimeoutMs", Long.toString(operationTimeout)),
-                entry("featureGates", Environment.STRIMZI_FEATURE_GATES),
+                entry("featureGates", Environment.STRIMZI_FEATURE_GATES.replaceAll(",", "\\\\,")),
```
scholzj, line 129: Do we want some explanatory comment about why it needs to be escaped like this?

michalxo: Good idea, I would put it there as it's easy to miss.

Frawless (approved):

im-konge (approved):

scholzj (approved):

PR #5217, opened 2021-06-29
Make it possible to predefine password for SCRAM-SHA users

### LINKED ISSUE #5129: Strimzi operator removes annotations from secrets with custom credentials
**Describe the bug**
A way to define custom credentials for a KafkaUser is to define define the credentials in a secret with a matching name and labeled with `strimzi.io/cluster` and `strimzi.io/kind` labels, for example:

```
apiVersion: v1
kind: Secret
metadata:
  name: cyndi
  labels:
    strimzi.io/kind: KafkaUser
  annotations:
    foo: bar
data:
  password: aGF2ZWZ1bg==
  sasl.jaas.config: b3JnLmFwYWNoZS5rYWZrYS5jb21tb24uc2VjdXJpdHkuc2NyYW0uU2NyYW1Mb2dpbk1vZHVsZSByZXF1aXJlZCB1c2VybmFtZT0iY3luZGkiIHBhc3N3b3JkPSJoYXZlZnVuIjs=
type: Opaque
```

The strimzi operator then uses these credentials for the user instead of generating new ones, which is great.

However, during the reconciliation, the operator also modifies the secret and removes all the annotations. This is not such a problem in the `foo: bar` scenario but it becomes a problem when a gitops tools is used to provision (and rotate) the credentials in the secret. **These tools typically make use of annotations and get confused or work unexpectedly if the strimzi operator interferes with the annotations they use.**

**To Reproduce**
1. Deploy Strimzi/AMQ Streams (I am using amqstreams.v1.7.2)
2. Create a custom secret with kafka user credentials (see above)
3. Create a KafkaUser with the matching name
4. Observe the strimzi-operator take ownership of the secret and remove all annotations

**Expected behavior**
The operator should read the provided credentials (and add custom labels/annotations if needed) but should otherwise respect the state of the resource and should not touch labels/annotations it had not set.

 - Strimzi version: amqstreams.v1.7.2
 - Installation method: OLM
 - Kubernetes cluster: CodeReady Containers version: 1.26.0+31f06c09
 - Infrastructure: Run locally on my Fedora laptop

### LINKED ISSUE #5141: Support providing custom KafkaUser credentials

By default, Strimzi operator auto-generates authentication credentials for each KafkaUser and puts them in a Kubernetes secret. This secret can only be used from within the same Kubernetes namespace. This is a limitation in situations where e.g. a KafkaConnect cluster resides in a different namespace than the Kafka cluster. The credentials need to be copied and kept in sync with other namespaces where they are needed.

A way around this limitation is to provide custom authentication credentials to Strimzi. There are many tools available that
* let one define their credentials in a central source-of-truth place (e.g. Vault), and
* will keep Kubernetes secrets in given namespaces in sync with the given source of truth

The piece that is currently missing is convincing the Strimzi operator to use the custom credentials instead of the auto-generated ones.

I'd like to be able to specify the credentials to be used as part of the KafkaUser definition, e.g:

```
kind: KafkaUser
spec:
  authentication:
    type: scram-sha-512
    secret: kafka-user-secret
```
or
```
kind: KafkaUser
spec:
  authentication:
    type: scram-sha-512
    password:
      secretKeyRef:
        name: kafka-user-secret
        key: password
```

I've tried tricking the operator into using my credentials by defining a Kubernetes secret with the same name and structure as it normally uses. This gets the job done but is not ideal as the fact that the Strimzi operator takes ownership and control over the provided secret causes conflicts with the cluster management tools used to provision the Secret in the first place. See https://github.com/strimzi/strimzi-kafka-operator/issues/5129#issuecomment-862613056 for more details.

I've also considered taking the auto-generate credentials, storing them in Vault and getting them to the other namespaces that way. The problem with that is that the Vault is no longer the single source of truth which comes with its own set of problems.

**Additional context**
See https://github.com/strimzi/strimzi-kafka-operator/issues/5129 for more context.

scholzj:
### Type of change

- Enhancement / new feature

### Description

Sometimes, users want to provide their own passwords for SCRAM-SHA users created by the User Operator instead of having them auto.generated by Strimzi. As a hack, one can first create the user secret and only afterwards the `KafkaUser` resource. But this approach has several challenges:
* Creating the user secret manually is sensitive for timing. If the secret and `KafkaUser` are created at the same time, there is a good chance that the operator might be faster and overwrite the password in the secret with the generated one. So one needs to first create the Secret and make sure it exists before creating the `KafkaUser`.
* Another issue is described in #5141 / #5129. When using GitOps patterns, the GitOps tooling often expects to own the created secret. But because the operator owns it as well, that might create conflict.

This PR adds a proper method for specifying custom password. Users can create a secret with their own name and store the desired password inside under some key. Then they can reference the key in the `KafkaUser` resource:

```yaml
kind: KafkaUser
metadata:
  name: my-user
  labels:
spec:
  authentication:
    type: scram-sha-512
    password:
      valueFrom:
        secretKeyRef:
          name: my-secret
          key: my-password
  # ...
```

The operator will read the desired password from the secret and use it instead of the generated one. It address both of the issues listed above:
* If the referenced secret does not exist yet, the reconciliation will fail with an error and will retry next time. That way, the secret and `KafkaUser` can be created together and will be eventually consistent.
* The secret which is referenced is owned by the user (or the software which created it). So when using any GitOps SW, it can set any labels and annos on this secret since it is owned by the tool and not by Strimzi.

As part of this work which caused fairly big changes to the code, this PR also does some other minor changes:
* Add renewal and validity as a proper members of `UserOperatorConfig` to avoid parsing the environment variables again and again
* Makes sure async calls tot he Kubernetes APi are used instead of the sync calls in the old code
* Tries to streamline the `createOrUpdate` method and the `KafkaUserModel` `fromCrd` method

This PR should close #5141 and close #5129.

### Checklist

- [x] Write tests
- [x] Update documentation
- [x] Update CHANGELOG.md


sknot-rh (approved):

CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ## 0.25.0
 
+* Allow providing custom password for SCRAM-SHA-512 users by referencing a secret in the `KafkaUser` resource
```
PaulRMellor, line 5:
```suggestion
* Allow a custom password to be provided for SCRAM-SHA-512 users by referencing a secret in the `KafkaUser` resource
```

api/src/main/java/io/strimzi/api/kafka/model/KafkaUserScramSha512ClientAuthentication.java
```diff
@@ -20,6 +20,17 @@
 
     public static final String TYPE_SCRAM_SHA_512 = "scram-sha-512";
 
+    private Password password;
+
+    @Description("Specify the password which should be used by this user. If not set, a new password will be generated by the User Operator.")
```
PaulRMellor, line 25:
```suggestion
    @Description("Specify the password for the user. If not set, a new password is generated by the User Operator.")
```

documentation/modules/con-securing-client-authentication.adoc
```diff
@@ -112,3 +112,30 @@ Decoding the generated password:
 ----
 echo "Z2VuZXJhdGVkcGFzc3dvcmQ=" | base64 --decode
 ----
+
+=== Providing custom password
```
PaulRMellor, line 116:
```suggestion
=== Custom password configuration
```

documentation/modules/con-securing-client-authentication.adoc
```diff
@@ -112,3 +112,30 @@ Decoding the generated password:
...
+
+When the user is created, Strimzi will normally generate a new random password.
```
PaulRMellor, line 118:
```suggestion
When a user is created, Strimzi generates a random password.
```

documentation/modules/con-securing-client-authentication.adoc
```diff
@@ -112,3 +112,30 @@ Decoding the generated password:
...
+In case you want to use your own password instead of the one generated by Strimzi, you can create a secret with the password and reference it in the `KafkaUser` resource:
```
PaulRMellor, line 119:
```suggestion
You can use your own password instead of the one generated by Strimzi. To do so, create a secret with the password and reference it in the `KafkaUser` resource:
```

documentation/modules/con-securing-client-authentication.adoc
```diff
@@ -112,3 +112,30 @@ Decoding the generated password:
...
+
+.An example `KafkaUser` with SCRAM-SHA-512 authentication and providing password
```
PaulRMellor, line 121:
```suggestion
.An example `KafkaUser` with a password set for SCRAM-SHA-512 authentication
```

documentation/modules/con-securing-client-authentication.adoc
```diff
@@ -112,3 +112,30 @@ Decoding the generated password:
...
+  authentication:
+    type: scram-sha-512
+    password:
+      valueFrom:
+        secretKeyRef:
+          name: my-secret <1>
+          key: my-password <2>
+  # ...
+----
+<1> Name of the secret with your predefined password
```
PaulRMellor, line 140:
```suggestion
<1> Name of the secret providing the predefined password
```

documentation/modules/con-securing-client-authentication.adoc
```diff
@@ -112,3 +112,30 @@ Decoding the generated password:
...
+<2> Key under which the password is stored inside the secret
```
PaulRMellor, line 141:
```suggestion
<2> The key for the password stored inside the secret.
```

PaulRMellor (approved):

PaulRMellor:
@scholzj A few minor comments. looks good.

PR #5222, opened 2021-06-30
[systemtest] Add 0.24.0 to upgrade tests

im-konge:
### Type of change

- New check

### Description

This PR adds `0.24.0` version to our upgrade tests and adding one more upgrade stage:
```
0.22.1 >  HEAD
0.23.0 > HEAD
0.24.0 > HEAD
```
Also this PR fixes naming of the parametrized test which was broken by adding additional stage.

### Checklist

### DIFF
--- a/systemtest/src/test/java/io/strimzi/systemtest/upgrade/AbstractUpgradeST.java
+++ b/systemtest/src/test/java/io/strimzi/systemtest/upgrade/AbstractUpgradeST.java
@@ -38,6 +38,7 @@ import org.junit.jupiter.params.provider.Arguments;
 import java.io.File;
 import java.io.FileReader;
 import java.io.IOException;
+import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.HashMap;
 import java.util.LinkedList;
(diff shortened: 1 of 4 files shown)


Frawless (approved):

scholzj:
Why do both 0.22.1 > 0.24.0 > HEAD and 0.24.0 > HEAD? Wouldn't it make more sense to do `0.22.1 -> HEAD` and `0.24.0 -> HEAD`? Also, what about 0.23 -> HEAD? Does it make sense or not?

im-konge:
I thought we discussed that we will do `0.22.1 > 0.24.0 > HEAD` (because of API conversion) and `0.24.0 > HEAD`.  When we tested upgrade from `0.22.1 > 0.23.0 > HEAD` I guess that we don't need to test it now as it already works (from my POV). But if you want to change it that way how you described, no problem.

scholzj:
> I thought we discussed that we will do 0.22.1 > 0.24.0 > HEAD (because of API conversion)

Well, technically, you can do 0.22.1 -> API conversion -> 0.24 -> HEAD equally well as 0.22.1 -> API conversion -> HEAD. So I'm not sure the conversion matters. 

The question is whether we want to focus more on:
* Upgrading from upgraded version which is what you proposed
* Upgrading directly

Each tests something little bit different. But since we now support version jumping, I wonder if 0.22.1 -> HEAD is better. The think to consider is how does 0.22.1 -> 0.24 -> HEAD differ from 0.24 -> HEAD. In theory, the patching might cause it to be little bit different ... but not sure we ever saw it. Also, if that makes sense, you should probably start the chain long before 0.22.1 as we had it in the past.

im-konge:
Yeah, you are right, we discussed this with @Frawless and agreed that we didn't really tested the `0.22.1 > HEAD` scenario. And, because the upgrade `0.23.0 > HEAD` is theoretically the same as `0.24.0 > HEAD` (no api conversion etc.), I think it would be better to do the `0.22.1 > HEAD` and `0.24.0 > HEAD`, if it makes sense to you.

scholzj:
> And, because the upgrade 0.23.0 > HEAD is theoretically the same as 0.24.0 > HEAD (no api conversion etc.), I think it would be better to do the 0.22.1 > HEAD and 0.24.0 > HEAD, if it makes sense to you.

I would not call it the same, it doesn't have the API conversion, but apart from that these are different versions. HEnce I wonder if it would make sense to have 0.23 -> HEAD as well. But yeah, 0.22.1 -> HEAD an 0.24 -> HEAD are more important I guess.

im-konge:
I guess it will not be a big deal to add the `0.23.0 > HEAD` as well, so maybe I can add it and leave others to say their opinion too. Thanks @scholzj

scholzj (approved):

Frawless (approved):

PR #5223, opened 2021-06-30
Fix resource count metric when multiple ns are configured

sknot-rh:

### Type of change
- Bugfix

### Description
When CO is configured to watch multiple namespaces, for each namespace is created a new verticle. That results in a race where the "slowest" verticle sets the value of the `strimzi_resources{kind="*",selector="",}` metric.
In this PR, I used maps, where keys are namespaces. For each namespace, a set of metrics is created.

```
$ oc get k --all-namespaces
NAMESPACE    NAME           DESIRED KAFKA REPLICAS   DESIRED ZK REPLICAS   READY     WARNINGS
myproject    my-cluster     3                        3                               
myproject    my-cluster-3   3                        3                               
myproject2   my-cluster     3                        3                     True      
myproject2   my-cluster-2   3                        3                     True      
myproject2   my-cluster-3   3                        3                     True 
```
Snímek z 2021-07-01 13-47-39

Note:
before this PR the metrics (for example succeeded or failed reconciliations count) had a value (`0`) from the start of the operator.
After this PR the metric is not available until its value is written.

### Checklist
- [ ] Write tests
- [ ] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,7 +19,7 @@
 * User Operator now uses Kafka Admin API to manage SCRAM-SHA-512 credentials.
   All operations done by the User Operator now use Kafka Admin API and connect directly to Kafka instead of ZooKeeper.
   As a result, the environment variables `STRIMZI_ZOOKEEPER_CONNECT` and `STRIMZI_ZOOKEEPER_SESSION_TIMEOUT_MS` were removed from the User Operator configuration.
- * Annotate Cluster Operator resource metrics by a namespace label
+* Annotate Cluster Operator resource metrics with a namespace label
 
 ## 0.24.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,6 @@
 * Add support for `tls-external` authentication to User Operator to allow management of ACLs and Quotas for TLS users with user certificates generated externally (#5249) 
 * Support for disabling the automatic generation of network policies by the Cluster Operator. Set the Cluster Operator's `STRIMZI_NETWORK_POLICY_GENERATION` environment variable to `false` to disable network policies. (#5258)
 * Update User Operator to use Admin API for managing SCRAM-SHA-512 users 
-* Annotate Cluster Operator resource metrics by a namespace label
 
 ### Changes, deprecations and removals
 
@@ -20,6 +19,7 @@
 * User Operator now uses Kafka Admin API to manage SCRAM-SHA-512 credentials.
   All operations done by the User Operator now use Kafka Admin API and connect directly to Kafka instead of ZooKeeper.
   As a result, the environment variables `STRIMZI_ZOOKEEPER_CONNECT` and `STRIMZI_ZOOKEEPER_SESSION_TIMEOUT_MS` were removed from the User Operator configuration.
+ * Annotate Cluster Operator resource metrics by a namespace label
 
 ## 0.24.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 * Add support for `tls-external` authentication to User Operator to allow management of ACLs and Quotas for TLS users with user certificates generated externally (#5249) 
 * Support for disabling the automatic generation of network policies by the Cluster Operator. Set the Cluster Operator's `STRIMZI_NETWORK_POLICY_GENERATION` environment variable to `false` to disable network policies. (#5258)
 * Update User Operator to use Admin API for managing SCRAM-SHA-512 users 
+* Annotate Cluster Operator resource metrics by a namespace label
 
 ### Changes, deprecations and removals
 
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/AbstractConnectOperator.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/AbstractConnectOperator.java
@@ -17,8 +17,6 @@ import io.fabric8.kubernetes.client.Watcher;
 import io.fabric8.kubernetes.client.WatcherException;
 import io.fabric8.kubernetes.client.dsl.Resource;
 import io.micrometer.core.instrument.Counter;
-import io.micrometer.core.instrument.Tag;
-import io.micrometer.core.instrument.Tags;
 import io.micrometer.core.instrument.Timer;
 import io.netty.channel.ConnectTimeoutException;
 import io.strimzi.api.kafka.KafkaConnectList;
@@ -47,6 +45,7 @@ import io.strimzi.operator.cluster.operator.resource.ResourceOperatorSupplier;
 import io.strimzi.operator.common.AbstractOperator;
 import io.strimzi.operator.common.Annotations;
 import io.strimzi.operator.common.BackOff;
+import io.strimzi.operator.common.Operator;
 import io.strimzi.operator.common.ReconciliationLogger;
 import io.strimzi.operator.common.Reconciliation;
 import io.strimzi.operator.common.Util;
@@ -79,6 +78,7 @@ import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
 import java.util.TreeMap;
+import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.function.BiFunction;
 import java.util.function.Function;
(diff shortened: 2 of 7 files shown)


scholzj (commented):
* What happens when the operator is watching all namespaces and the namespace is `*`? Or is the namespace in the metrics already the actual namespace of the resource?
* Maybe adding the namespace label deserves a note in https://github.com/strimzi/strimzi-kafka-operator/blob/main/CHANGELOG.md#changes-deprecations-and-removals? I'm not sure if added label can cause some issues.
* Also, is there any risk the `namespace` collides with some regular Prometheus labels? Such as the namespace from which the metric was colelcted?

sknot-rh:
- the namespace is set to `*` (`strimzi_resources{kind="Kafka",namespace="*",selector="",} 2.0`
- good point
- I do not think so. I deployed entire stack (Prometheus + grafana) and there was no issue with that

scholzj:
> the namespace is set to * (strimzi_resources{kind="Kafka",namespace="*",selector="",} 2.0

I think we need to detect this and leave out the tag. The `*` IMHO should no be there.

CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 ## 0.25.0
 
 * Allow a custom password to be set for SCRAM-SHA-512 users by referencing a secret in the `KafkaUser` resource
+* Annotate Cluster Operator resource metrics by a namespace label
```
scholzj, line 6: Should it be in the changes section below? That way people looking for compatibility changes will be more likely to notice.

scholzj (commented):
The code looks good. But you need to re-enable the system tests disabled by @see-quick in #5137

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/AbstractConnectOperator.java
```diff
@@ -112,11 +112,11 @@
...
-    private final Counter connectorsReconciliationsCounter;
-    private final Counter connectorsFailedReconciliationsCounter;
-    private final Counter connectorsSuccessfulReconciliationsCounter;
-    private final AtomicInteger connectorsResourceCounter;
-    private final Timer connectorsReconciliationsTimer;
+    private Map<String, Counter> connectorsReconciliationsCounterMap;
+    private Map<String, Counter> connectorsFailedReconciliationsCounterMap;
+    private Map<String, Counter> connectorsSuccessfulReconciliationsCounterMap;
+    private Map<String, AtomicInteger> connectorsResourceCounterMap;
+    private Map<String, Timer> connectorsReconciliationsTimerMap;
```
tombentley, line 119: Less code to initialise the field here as part of the declaration.

operator-common/src/main/java/io/strimzi/operator/common/AbstractOperator.java
```diff
@@ -84,60 +84,35 @@
...
+    private final Labels selectorLabels;
+    private Map<String, AtomicInteger> resourcesStateCounter;
+    private Map<String, AtomicInteger> resourceCounterMap;
+    private Map<String, AtomicInteger> pausedResourceCounterMap;
+    private Map<String, Counter> periodicReconciliationsCounterMap;
+    private Map<String, Counter> reconciliationsCounterMap;
+    private Map<String, Counter> failedReconciliationsCounterMap;
+    private Map<String, Counter> successfulReconciliationsCounterMap;
+    private Map<String, Counter> lockedReconciliationsCounterMap;
+    private Map<String, Timer> reconciliationsTimerMap;
```
tombentley, line 97: Same comment, just initialize the field here/

operator-common/src/main/java/io/strimzi/operator/common/AbstractOperator.java
```diff
@@ -496,38 +471,69 @@ public void accept(WatcherException e) {
...
+                reconciliationTimerSample.stop(getReconciliationsTimer(reconciliation.namespace()));
                 LOGGER.warnCr(reconciliation, "Failed to reconcile", cause);
             }
         }
     }
 
-    public Counter getPeriodicReconciliationsCounter() {
-        return periodicReconciliationsCounter;
+    @Override
+    public Counter getPeriodicReconciliationsCounter(String namespace) {
```
tombentley, line 497: My usual pet peeve about methods that return a value not needing to be called `get${Something}` when they can just be called `${something}`.

operator-common/src/main/java/io/strimzi/operator/common/Operator.java
```diff
@@ -94,9 +99,60 @@ default void reconcileThese(String trigger, Set<NamespaceAndName> desiredNames,
...
+            metricTags = Tags.of(Tag.of("kind", kind), Tag.of("selector", selectorValue));
+        } else {
+            metricTags = Tags.of(Tag.of("kind", kind), Tag.of("namespace", namespace), Tag.of("selector", selectorValue));
+        }
+
+        Counter counter = counterMap.get(namespace + "/" + kind);
+        if (counter == null) {
+            counter = metrics.counter(metricName, metricHelp, metricTags);
+            counterMap.put(namespace + "/" + kind, counter);
+        }
```
tombentley, line 121: Use `computeIfAbsent` rather than separate `get` & `put`. With the ConcurrentHashMap that will give you atomicity which you need if you're not going to miscount things when different threads can access the map.

operator-common/src/main/java/io/strimzi/operator/common/Operator.java
```diff
@@ -94,9 +99,60 @@ default void reconcileThese(String trigger, Set<NamespaceAndName> desiredNames,
...
+        } else {
+            metricTags = Tags.of(Tag.of("kind", kind), Tag.of("namespace", namespace), Tag.of("selector", selectorValue));
+        }
+
+        AtomicInteger gauge = gaugeMap.get(namespace + "/" + kind);
+        if (gauge == null) {
+            gauge = metrics.gauge(metricName, metricHelp, metricTags);
+            gaugeMap.put(namespace + "/" + kind, gauge);
+        }
+        return gauge;
```
tombentley, line 139: This is basically identical to `getCounter` and `getTimer`, the only difference is which method of `metrics` we call. You should be able to write a single `private` generic method with abstracts across this. 

Something like 

```java
static <M> AtomicInteger metric(String namespace, String kind, MetricsProvider metrics, Labels selectorLabels, Map<String, AtomicInteger> gaugeMap, Function<Tags, M> fn) {
    //...
    return metrics.computeIfAbsent(metricName, x -> fn.apply(metricTags);
}
```

and implement each of `getGuage`, `getCounter` and `getTimer` with `return metric(..., tags -> metrics.guage(metricName, metricHelp, tags)`.

tombentley (approved):
Thanks!

sknot-rh:
> > the namespace is set to * (strimzi_resources{kind="Kafka",namespace="*",selector="",} 2.0
> 
> I think we need to detect this and leave out the tag. The `*` IMHO should no be there.

Hmm, because AbstractConnectOperator registers metrics, we cannot omit the namespace label. Otherwise, we get an error about metric tags conflict.

scholzj:
> Hmm, because AbstractConnectOperator registers metrics, we cannot omit the namespace label. Otherwise, we get an error about metric tags conflict.

But you cannot use `*` ... that is just an internal marker.

sknot-rh:
> > Hmm, because AbstractConnectOperator registers metrics, we cannot omit the namespace label. Otherwise, we get an error about metric tags conflict.
> 
> But you cannot use `*` ... that is just an internal marker.

Does `any` sound any better? I am aware of conflicts it could create :(

scholzj:
> Does any sound any better? I am aware of conflicts it could create :(

I'm not Prometheus expert, but it sounds equally arbitrary. Can we leave the label empty? (Just empty string value?)

If not, maybe we need to rework it and put there the namespace of the custom resource instead of the watched namespace? (I think that makes more sense in general, but might be harder to do . especially to handle removed CRs)

tombentley:
I don't think you can use `any`, since there might be a namespace with that name. I guess you could use something which was illegal as a namespace name (dns subdomain), e.g. `<all>` or `ALL`, assuming that worked with Prometheus. I think that would be clearer than using the empty string.

scholzj:
> I think that would be clearer than using the empty string.

I think empty string fits better with the concept of labels and their use than something like `<any>`. Also, we already use the empty label for the selector IIRC.

see-quick:
@sknot-rh btw you can trigger also macro-pipeline using `/azp run all-strimzi-system-tests`

scholzj (approved):
If it has to be, I can probably live with this. But I still think it would make much more sense to use the namespace of the actual resource in the metric label rather than having it empty. That would give clear and consistent metrics regardless the operator configuration.

CHANGELOG.md
```diff
@@ -19,6 +19,7 @@
 * User Operator now uses Kafka Admin API to manage SCRAM-SHA-512 credentials.
   All operations done by the User Operator now use Kafka Admin API and connect directly to Kafka instead of ZooKeeper.
   As a result, the environment variables `STRIMZI_ZOOKEEPER_CONNECT` and `STRIMZI_ZOOKEEPER_SESSION_TIMEOUT_MS` were removed from the User Operator configuration.
+ * Annotate Cluster Operator resource metrics by a namespace label
```
samuel-hawker, line 22:
```suggestion
 * Annotate Cluster Operator resource metrics with a namespace label
```
I think we want to use with here, but might be worth asking someone with a better knowledge of the English language than myself!

sknot-rh: As my English is as good as stepping on the lego, I am pinging @PaulRMellor and @laidan6000 to check.

PaulRMellor: @sknot-rh Yes. I'd change to _with_

sknot-rh:
> If it has to be, I can probably live with this. But I still think it would make much more sense to use the namespace of the actual resource in the metric label rather than having it empty. That would give clear and consistent metrics regardless the operator configuration.

I added the resource ns to the metrics labels even when the `*` namespace is configured.

Issue #5229, opened 2021-06-30
Deploying Standalone User Operator CA Secret Format


pecigonzalo:
**Suggestion / Problem**
I was trying to deploy the Standalone User Operator against AWS MSK and using TLS as Auth. For this I created an AWS Private certificate authority, and then a Subordinate CA that could issue certificates.
I added them as secrets using
```
kubectl create --save-config --dry-run=client secret generic strimzi-subordinate-cert --from-file=keys/subordinate.crt -o yaml | kubectl apply -f -
kubectl create --save-config --dry-run=client secret generic strimzi-subordinate-key --from-file=keys/subordinate.key -o yaml | kubectl apply -f -
```
and tried my operator but it was failing with a NPE pointing at https://github.com/strimzi/strimzi-kafka-operator/blob/main/operator-common/src/main/java/io/strimzi/operator/cluster/model/Ca.java#L731.

```
2021-06-30 21:53:09 DEBUG Ca:282 - Reconciliation #3(timer) KafkaUser(default/kafka.user.kafka-simple-client): Generating certificate /CN=kafka.user.kafka-simple-client with SAN null, signed by CA clients-ca
2021-06-30 21:53:09 DEBUG OpenSslCertManager:617 - Running command [openssl, req, -new, -batch, -nodes, -keyout, /tmp/tls5600546536714548180key, -out, /tmp/tls7779136491457340388csr, -subj, /CN=kafka.user.kafka-simple-client]
2021-06-30 21:53:09 DEBUG OpenSslCertManager:640 - Got result 0
2021-06-30 21:53:09 WARN  KafkaUserOperator:149 - Reconciliation #3(timer) KafkaUser(default/kafka.user.kafka-simple-client): java.lang.NullPointerException
2021-06-30 21:53:09 ERROR AbstractOperator:274 - Reconciliation #3(timer) KafkaUser(default/kafka.user.kafka-simple-client): createOrUpdate failed
java.lang.NullPointerException: null
	at java.util.Base64$Decoder.decode(Base64.java:561) ~[?:?]
	at io.strimzi.operator.cluster.model.Ca.currentCaKey(Ca.java:732) ~[io.strimzi.operator-common-0.24.0.jar:0.24.0]
	at io.strimzi.operator.cluster.model.Ca.generateSignedCert(Ca.java:285) ~[io.strimzi.operator-common-0.24.0.jar:0.24.0]
[...]

PR #5237, opened 2021-07-01
Properly validate CA secrets in User Operator

scholzj:
### Type of change

- Bugfix

### Description

The User Operator currently validates only whether the CA secrets exist. But it does not validate whether it has inside the right fields. So when the user passes a secret which for example uses `tls.crt` instead of `ca.crt`, it just throws an NPE. This PR improves the validation to also check if the right key exists in the secret.

This was originally raised as part of #5229 

### Checklist

- [x] Write tests

### DIFF
--- a/user-operator/src/main/java/io/strimzi/operator/user/model/KafkaUserModel.java
+++ b/user-operator/src/main/java/io/strimzi/operator/user/model/KafkaUserModel.java
@@ -260,13 +260,13 @@ public class KafkaUserModel {
 
     private void validateCACertificates(Secret clientsCaCertSecret, Secret clientsCaKeySecret)   {
         if (clientsCaCertSecret == null) {
-            // CA certificate secret does nto exist
+            // CA certificate secret does not exist
             throw new InvalidCertificateException("The Clients CA Cert Secret is missing");
         } else if (clientsCaCertSecret.getData() == null || clientsCaCertSecret.getData().get("ca.crt") == null)    {
             // CA certificate secret exists, but does not have the ca.crt key
             throw new InvalidCertificateException("The Clients CA Cert Secret is missing the ca.crt file");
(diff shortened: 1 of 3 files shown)


user-operator/src/main/java/io/strimzi/operator/user/model/KafkaUserModel.java
```diff
@@ -198,68 +198,79 @@ public Secret generateSecret()  {
...
+        try {
+            this.userCertAndKey = clientsCa.generateSignedCert(name);
+        } catch (IOException e) {
+            LOGGER.errorCr(reconciliation, "Error generating signed certificate for user {}", name, e);
+        }
+    }
 
+    private void validateCACertificates(Secret clientsCaCertSecret, Secret clientsCaKeySecret)   {
+        if (clientsCaCertSecret == null) {
+            // CA certificate secret does nto exist
```
ppatierno, line 263: not

user-operator/src/main/java/io/strimzi/operator/user/model/KafkaUserModel.java
```diff
@@ -198,68 +198,79 @@ public Secret generateSecret()  {
...
+            throw new InvalidCertificateException("The Clients CA Cert Secret is missing");
+        } else if (clientsCaCertSecret.getData() == null || clientsCaCertSecret.getData().get("ca.crt") == null)    {
+            // CA certificate secret exists, but does not have the ca.crt key
+            throw new InvalidCertificateException("The Clients CA Cert Secret is missing the ca.crt file");
+        } else if (clientsCaKeySecret == null) {
+            // CA certificate secret does nto exist
```
ppatierno, line 269: not

ppatierno (approved):
LGTM. Just two typos.

Issue #5242, opened 2021-07-01
strimz-kafka-operator restarting kafka due to fluxcd


kangzhang:
**Describe the bug**
We started using strimz-kafka-operator recently and it was overall smooth (thank you!). 

We ran into some problems with fluxcd (The GitOps tool we use https://fluxcd.io/): strimz-kafka-operator would restart kafka sometimes when we push a change. Looking a bit into it and it seems that some lables added by fluxcd, triggered strimz-kafka-operator to do a rolling update.

**Expected behavior**
strimzi-kafka-operator ignore all fluxcd tags. 

**Environment (please complete the following information):**
 - Strimzi version: main
 - Installation method: YAML files via fluxcd
 - Kubernetes cluster: Kubernetes 1.20
 - Infrastructure: Amazon EKS

**YAML files and logs**

This is the log from the operator: https://gist.github.com/kangzhang/13cec69ce30079b94d4f3bf41631d6a6 . You can see that the operator detected some label changes by fluxcd: https://gist.github.com/kangzhang/13cec69ce30079b94d4f3bf41631d6a6#file-gistfile1-txt-L244 and triggered a rolling update: https://gist.github.com/kangzhang/13cec69ce30079b94d4f3bf41631d6a6#file-gistfile1-txt-L267

**Additional context**

I'm wondering if we could black list fluxcd labels? The are only used by fluxcd to track the deployment and should not be used by strimzi. 

Here are some example labels fluxcd would add:

```
              kustomize.toolkit.fluxcd.io/checksum=77dd8897619d50e2035bafbfe0b41e22cea893cd
              kustomize.toolkit.fluxcd.io/name=environment
              kustomize.toolkit.fluxcd.io/namespace=flux-system
```

I also read thru #2558 and #5083 . The flag `STRIMZI_LABELS_EXCLUSION_PATTERN ` does not seem to help. (I already set to `"(^app.kubernetes.io/(?!part-of).*|^kustomize.toolkit.fluxcd.io.*)"` and you can see it in log file too: https://gist.github.com/kangzhang/13cec69ce30079b94d4f3bf41631d6a6#file-gistfile1-txt-L69

Thanks!

scholzj:
What version of Strimzi do you use? `main` doesn't mean anything - it can be build which is several months old as well as build from yesterday. Did you tried it with 0.23.0 or 0.24.0 versions?

scholzj:
[...]

PR #5245, opened 2021-07-02
Add FluxCD Kustomize labels to the exclude pattern

scholzj:
### Type of change

- Enhancement / new feature

### Description

As discussed #5242, FluxCD Kustomize is adding bucnh of labels containing among other things some checksums which might change fairly often. Right now, we mirror these labels to the operands and it is causing unnecessary rolling updates. This Pr updates the default value for the `STRIMZI_LABELS_EXCLUSION_PATTERN` environment variable to exclude these labels by default and not mirror them to the operands.

### Checklist

- [x] Update documentation
- [x] Update CHANGELOG.md

### DIFF
--- a/operator-common/src/main/java/io/strimzi/operator/common/model/Labels.java
+++ b/operator-common/src/main/java/io/strimzi/operator/common/model/Labels.java
@@ -97,7 +97,7 @@ public class Labels {
      * Used to exclude parent CR's labels from being assigned to provisioned subresources
      */
     public static final Pattern STRIMZI_LABELS_EXCLUSION_PATTERN = Pattern.compile(System.getenv()
-            .getOrDefault("STRIMZI_LABELS_EXCLUSION_PATTERN", "^app.kubernetes.io/(?!part-of).*"));
+            .getOrDefault("STRIMZI_LABELS_EXCLUSION_PATTERN", "(^app.kubernetes.io/(?!part-of).*|^kustomize.toolkit.fluxcd.io.*)"));
 
     /**
      * The empty set of labels.


operator-common/src/main/java/io/strimzi/operator/common/model/Labels.java
```diff
@@ -97,7 +97,7 @@
      * Used to exclude parent CR's labels from being assigned to provisioned subresources
      */
     public static final Pattern STRIMZI_LABELS_EXCLUSION_PATTERN = Pattern.compile(System.getenv()
-            .getOrDefault("STRIMZI_LABELS_EXCLUSION_PATTERN", "^app.kubernetes.io/(?!part-of).*"));
+            .getOrDefault("STRIMZI_LABELS_EXCLUSION_PATTERN", "(^app.kubernetes.io/(?!part-of).*|^kustomize.toolkit.fluxcd.io.*)"));
```
ppatierno, line 100: why are we setting the kustomize exclusion by default and not letting the user to do so using the `STRIMZI_LABELS_EXCLUSION_PATTERN` env var?

scholzj: So, you prefer to:
* Let the user run into problems
* Search for a solution
* Open an issue / discussion
* Wait for Jakub to help them?

ppatierno: I am just saying that this list could grow with any kind of other projects labels if we start to add the first one.

ppatierno (approved):

operator-common/src/main/java/io/strimzi/operator/common/model/Labels.java
scholzj: I agree, but I'm not sure I see anything wrong with it. At the end that was one of the purposes with which we created it, or?

Issue #3467, opened 2020-08-06
[Enhancement] Allow setting NetworkPolicy for Metrics Port


forsberg:
I would like to isolate the metrics port (9404) so that only Prometheus can talk to it. As far as I can see, there's no built in support in the Kafka CRD for doing so, although I can of course do it by adding a separate NetworkPolicy.

davidjlynn:
This would also be valuable for me.

How to do this, I was hoping that [NetworkPolicyPeers](https://strimzi.io/docs/operators/latest/using.html#configuration-listener-network-policy-reference), which is used to restrict the regular listeners, could be used when we define metrics.

We define metrics here: https://strimzi.io/docs/operators/latest/using.html#type-JmxPrometheusExporterMetrics-reference
So perhaps another property `networkPolicyPeers` to align with the listeners would be sensible.

Note there is also the deprecated metrics `metrics` rather than the replacement of `metricsConfig`, we can likely disregard this when adding this feature, as the migration required to get this enhancement does not seem particularly hard.

scholzj:
TBH, I'm not sure where it fits in the API. It might fit into the template section maybe. Something I also considered was to have a feature gate to just disable network policy creation and let user choose if they want to manage all network policies on their own (I think the metrics as they are right now work fine for 9X% people - but some people have different some special requirements and it might be hard to them all into the CRD APIs).

davidjlynn:
I think a feature gate is a really good idea.
Otherwise Strimzi may get more requests for functionality around NetworkPolicy, (Like [Egress](https://github.com/strimzi/strimzi-kafka-operator/issues/2212) for example)

If a consumer is locking down access to Kafka, they are likely comfortable creating their own network policies, so disabling the network policy creation will solve the issue nicely.

I do still see benefit for Strimzi supporting the existing behaviour of by default adding "allow all to access kafka" network policies, it gets Kafka up and running even in a deployment with a deny-all network policy.

Then if a consumer is wanting granular network policies they can disable and write their own.

scholzj:
> I do still see benefit for Strimzi supporting the existing behaviour of by default adding "allow all to access kafka" network policies, it gets Kafka up and running even in a deployment with a deny-all network policy.

Yeah, that is really one of the main reasons why we added it in the first place.

scholzj:
[...]

PR #5258, opened 2021-07-06
issue-3467 Add option to disable Network Policy Generation.

davidjlynn:
### Type of change

- Enhancement / new feature

### Description

This pull request aims to solve #3467 by adding a feature gate to toggle the generation of network policy generation by Strimzi. This feature gate is by default enabled (as this maintains current behaviour) and can be used by specifying the `STRIMZI_NETWORK_POLICY_GENERATION` environment variable as `false` in the configuration.

### Checklist

- [x] Write tests
- [x] Update documentation
- [x] Update CHANGELOG.md


davidjlynn:
As mentioned in strimzi/proposals#35 an environment variable should be used instead.

This PR switched over to this approach, rather than the previous approach of feature gate.

documentation/modules/operators/ref-operator-cluster.adoc
```diff
@@ -168,6 +168,14 @@ If you are using a different DNS domain name suffix in your cluster, change the
 The timeout for building new Kafka Connect images with additional connectots, in milliseconds.
 This value should be increased when using Strimzi to build container images containing many connectors or using a slow container registry.
 
+`STRIMZI_NETWORK_POLICY_GENERATION` :: Optional, default `true`.
+The boolean value, controlling whether Strimzi should generate Network Policy files using the operator.
```
scholzj, line 172: Maybe `Network Policy resources` is better than `Network Policy files`?

scholzj (commented):
The way it is implemented here is that it will always reconcile the network policies. This has two side-effects:
* If your custom network policy has the same name as the one generated by the operator, the operator will wipe it.
* It still requires the RBAC rights to work with network policies. That means it will not work in use-cases where the user is not allowed to use network policies (we had such as well).

I wonder if it would be better to instead of changing the models, to instead not reconcile the network policies. Either by using some _dummy_ NetworkPolicyOperator would would do nothing (that might make it easier to unit test). Or by handling it in the operator code for example like this (from the `KafkaAssemblyOperator` class):

```
        Future<ReconciliationState> kafkaNetPolicy() {
            if (isNetworkPolicyGeneration) {
                return withVoid(networkPolicyOperator.reconcile(reconciliation, namespace, KafkaCluster.networkPolicyName(name), kafkaCluster.generateNetworkPolicy(operatorNamespace, operatorNamespaceLabels)));
            } else {
                return withVoid(Future.succeededFuture());
            }
        }
```

davidjlynn:
> I wonder if it would be better to instead of changing the models, to instead not reconcile the network policies. Either by using some _dummy_ NetworkPolicyOperator would would do nothing (that might make it easier to unit test). Or by handling it in the operator code for example like this (from the `KafkaAssemblyOperator` class)

@scholzj I am happy to switch to the example you gave above, indeed this avoids the reconciliation step completely which is sensible.

documentation/modules/operators/ref-operator-cluster.adoc
davidjlynn: Switched to better naming.

scholzj (approved):

CHANGELOG.md
```diff
@@ -6,6 +6,7 @@
 * Open Policy Agent authorizer updated to a new version supporting Scala 2.13. See the _Changes, deprecations and removals_ sections for more details. (#5192)
 * Allow a custom password to be set for SCRAM-SHA-512 users by referencing a secret in the `KafkaUser` resource
 * Add support for `tls-external` authentication to User Operator to allow management of ACLs and Quotas for TLS users with user certificates generated externally (#5249) 
+* Add support for disabling generation of network policies (Available via the `STRIMZI_NETWORK_POLICY_GENERATION` environment variable, disabling this will disable network policies)
```
PaulRMellor, line 9:
```suggestion
* Support for disabling the automatic generation of network policies for listeners. Users can now use the Cluster Operator's `STRIMZI_NETWORK_POLICY_GENERATION` environment variable to disable network policies.
```

davidjlynn: Thanks @PaulRMellor , I like your phrasing better but just wanted to also remove the "for listeners" part.
This applies to all Network Policies and the Cluster Operator generates policies for Metrics too.
I have change to:
`* Support for disabling the automatic generation of network policies by the Cluster Operator. Set the Cluster Operator's `STRIMZI_NETWORK_POLICY_GENERATION` environment variable to `false` to disable network policies.`
Let me know if this is acceptable.

documentation/modules/operators/ref-operator-cluster.adoc
```diff
@@ -168,6 +168,14 @@ If you are using a different DNS domain name suffix in your cluster, change the
...
 This value should be increased when using Strimzi to build container images containing many connectors or using a slow container registry.
 
+`STRIMZI_NETWORK_POLICY_GENERATION` :: Optional, default `true`.
+The boolean value, controlling whether Strimzi should generate Network Policy resources using the operator.
+
+By default, this is enabled, causing Network policies to be generated which allow Kafka connections between the relevant components involved.
+
+However, if the user desires custom network policies, they may disable this generation and take the responsibility for maintaining connections between the components.
+
+
```
PaulRMellor, line 178:
```suggestion
`STRIMZI_NETWORK_POLICY_GENERATION` :: Optional, default `true`.
Controls whether Strimzi generates network policy resources.
Network policies allow connections between Kafka components.

Set this environment variable to `false` to disable network policy generation. You might do this, for example, if you want to use custom network policies. Custom network policies allow more control over maintaining the connections between components.

```

PaulRMellor (approved):
Looks good. Couple of suggestions

CHANGELOG.md
PaulRMellor: Thanks @davidjlynn - i took out the listener part when I realised what this was about. :-) See above^

CHANGELOG.md
```diff
@@ -6,6 +6,7 @@
 * Open Policy Agent authorizer updated to a new version supporting Scala 2.13. See the _Changes, deprecations and removals_ sections for more details. (#5192)
 * Allow a custom password to be set for SCRAM-SHA-512 users by referencing a secret in the `KafkaUser` resource
 * Add support for `tls-external` authentication to User Operator to allow management of ACLs and Quotas for TLS users with user certificates generated externally (#5249) 
+* Support for disabling the automatic generation of network policies by the Cluster Operator. Set the Cluster Operator's `STRIMZI_NETWORK_POLICY_GENERATION` environment variable to `false` to disable network policies.
```
PaulRMellor, line 9:
```suggestion
* Support for disabling the automatic generation of network policies by the Cluster Operator. Set the Cluster Operator's `STRIMZI_NETWORK_POLICY_GENERATION` environment variable to `false` to disable network policies. (#5258)
```

scholzj:
@davidjlynn Can you have a look at the broken DCO sign-off? Under the _Details_ link, there should be some instructions how to fix it.

tombentley (approved):
Thanks!

davidjlynn:
I have messed up the squash, (The sign off command wasn't working) I will fix ASAP.

davidjlynn:
Squash and DCO should be fixed now.

scholzj:
Thanks. I will merge it after the tests pass.

PR #4746, opened 2021-04-13
Allow empty tokenEndpointUri in OAuth PLAIN to assume token in password

elakito:
### Type of change

_Select the type of your PR_
- Enhancement / new feature

### Description

_Please describe your pull request_

When the oauth plain mode is updated with https://github.com/strimzi/strimzi-kafka-oauth/pull/107, this change in strimzi-kafka-operator will allow this configuration by not forbidding an empty `TokenEndpointUri` value.

### Checklist

_Please go through this checklist and make sure all applicable tasks have been done_

- [ ] Write tests
- [ ] Make sure all tests pass
- [ ] Update documentation
- [ ] Check RBAC rights for Kubernetes / OpenShift roles
- [ ] Try your changes from Pod inside your Kubernetes and OpenShift cluster, not just locally
- [ ] Reference relevant issue(s) and close them after merging
- [ ] Update CHANGELOG.md
- [ ] Supply screenshots for visual changes, such as Grafana dashboards

PR #5259, opened 2021-07-06
Add explanation for `enablePlain` in `no-client-credentials` mode when `tokenEndpointUri` is not set

mstruk:
Signed-off-by: Marko Strukelj <marko.strukelj@gmail.com>

### Type of change

_Select the type of your PR_

- Documentation

### Description

This is additional documentation for #4746.

### Checklist

- [ ] Write tests
- [x] Update documentation
- [ ] Update CHANGELOG.md


laidan6000:
Remember to cherry-pick into Strimzi 0.24.x.

scholzj (approved):

documentation/modules/oauth/proc-oauth-authentication-broker-config.adoc
```diff
@@ -134,7 +134,7 @@ Depending on how you apply OAuth 2.0 authentication, and the type of authorizati
...
...PLAIN mechanism on the listener, which is supported by all clients on all platforms. The Kafka client has to enable the PLAIN mechanism and set the `username` and the `password`. This mechanism can be used to authenticate either by using the OAuth access token, or by using the OAuth `clientId` and `secret` (client credentials). The behavior is additionally controlled by whether `tokenEndpointUri` is specified or not. If `tokenEndpointUri` is specified, then if the client sets `password` to start with the string `$accessToken:`, the password is interpreted as the access token on the server, and `username` as the account username, otherwise the `username` is interpreted as the `clientId`, and `password` as the client `secret`, which the broker uses to obtain the access token in client's name. However, if `tokenEndpointUri` is not specified, the `password` is always interpreted as an access token, and the `username` is always interpreted as the account username which has to match the principal id extracted from the token (a so called 'no-client-credentials' mode, because client always has to obtain the access token by itself, and can't use `clientId` and `secret`). Default is `false`.
```
laidan6000, line 137:
```suggestion
<8> Set to `true` to enable PLAIN authentication on the listener, which is supported by all clients on all platforms. The Kafka client must enable the PLAIN mechanism and set the `username` and `password`. PLAIN can be used to authenticate either by using the OAuth access token, or the OAuth `clientId` and `secret` (the client credentials). The behavior is additionally controlled by whether `tokenEndpointUri` is specified or not (default is `false`). 
+
If `tokenEndpointUri` is specified and the client sets `password` to start with the string `$accessToken:`, the server interprets the password as the access token and the `username` as the account username. Otherwise, the `username` is interpreted as the `clientId` and the `password` as the client `secret`, which the broker uses to obtain the access token in the client's name.
+
If `tokenEndpointUri` is not specified, the `password` is always interpreted as an access token and the `username` is always interpreted as the account username, which must match the principal id extracted from the token. This is known as 'no-client-credentials' mode because the client must always obtain the access token by itself, and can't use `clientId` and `secret`. 
```

laidan6000 (changes requested):
Looks good. I made a few changes to the call out text.

documentation/modules/oauth/proc-oauth-authentication-broker-config.adoc
```diff
@@ -134,7 +134,7 @@ Depending on how you apply OAuth 2.0 authentication, and the type of authorizati
...
 <9> This has to be set to support the client credentials authentication when `enablePlain` is set to true, as described in previous point.
```
laidan6000, line 138: @mstruk - I think number 9 also needs to be changed slightly.

documentation/modules/oauth/proc-oauth-authentication-broker-config.adoc
mstruk:
```
The behavior is additionally controlled by whether `tokenEndpointUri` is specified or not (default is `false`). 
```

I find the final part, that used to be its own sentence, confusing as it appears to refer to `tokenEndpointUri` when in fact it should refer to `enablePlain`.  I suggest putting it back into its own sentence as is the pattern for all the options.

documentation/modules/oauth/proc-oauth-authentication-broker-config.adoc
mstruk: @laidan6000 Ah, I see what you mean, it's now optional so 'has to be set' is not ok. 
```suggestion
<9> Additional configuration for PLAIN mechanism to allow clients to authenticate by passing `clientId` and `secret` as `username` and `password` as described in the previous point. If not specified the clients can authenticate over PLAIN only by passing an access token as `password` parameter.
```

laidan6000: Looks good.

documentation/modules/oauth/proc-oauth-authentication-broker-config.adoc
laidan6000: OK, please go ahead and change it.

laidan6000 (approved):
Looks good.

scholzj:
@strimzi/maintainers This changes the Java code ... so we need another reviewer to review / approve this. Can someone have a look?

documentation/modules/oauth/proc-oauth-authentication-broker-config.adoc
```diff
@@ -134,8 +134,12 @@ Depending on how you apply OAuth 2.0 authentication, and the type of authorizati
...
++
+If `tokenEndpointUri` is specified and the client sets `password` to start with the string `$accessToken:`, the server interprets the password as the access token and the `username` as the account username. Otherwise, the `username` is interpreted as the `clientId` and the `password` as the client `secret`, which the broker uses to obtain the access token in the client's name.
++
+If `tokenEndpointUri` is not specified, the `password` is always interpreted as an access token and the `username` is always interpreted as the account username, which must match the principal id extracted from the token. This is known as 'no-client-credentials' mode because the client must always obtain the access token by itself, and can't use `clientId` and `secret`. 
+<9> Additional configuration for PLAIN mechanism to allow clients to authenticate by passing `clientId` and `secret` as `username` and `password` as described in the previous point. If not specified the clients can authenticate over PLAIN only by passing an access token as `password` parameter.
```
tombentley, line 142: This is sufficiently complicated that I think it would be good to have examples (possibly with pictures) which illustrated these alternatives. I'm not suggesting we should do that in this PR though.

tombentley (approved):

PR #5266, opened 2021-07-07
ST: Apply Roles when CO use namespace-rbac and cluster wide options properly

Frawless:

### Type of change

- Enhancement / new feature

### Description

This PR fixes the way how roles are created for namespace-rbac mode of CO.

### Checklist

### DIFF
--- a/systemtest/src/main/java/io/strimzi/systemtest/resources/kubernetes/RoleResource.java
+++ b/systemtest/src/main/java/io/strimzi/systemtest/resources/kubernetes/RoleResource.java
@@ -1,0 +1,62 @@
+/*
+ * Copyright Strimzi authors.
+ * License: Apache License 2.0 (see the file LICENSE or http://apache.org/licenses/LICENSE-2.0.html).
+ */
+package io.strimzi.systemtest.resources.kubernetes;
+
+import io.fabric8.kubernetes.api.model.rbac.Role;
+import io.fabric8.kubernetes.api.model.rbac.RoleBuilder;
+import io.strimzi.systemtest.Constants;
+import io.strimzi.systemtest.resources.ResourceManager;
+import io.strimzi.systemtest.resources.ResourceType;
+import io.strimzi.test.TestUtils;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+
+public class RoleResource implements ResourceType<Role> {
+
+    private static final Logger LOGGER = LogManager.getLogger(RoleResource.class);
+
+    @Override
+    public String getKind() {
+        return Constants.ROLE;
+    }
+    @Override
+    public Role get(String namespace, String name) {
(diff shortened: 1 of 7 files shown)


see-quick (approved):

systemtest/src/main/java/io/strimzi/systemtest/resources/kubernetes/RoleResource.java
```diff
@@ -0,0 +1,62 @@
...
+    public Role get(String namespace, String name) {
+        return ResourceManager.kubeClient().namespace(namespace).getRole(name);
+    }
+    @Override
+    public void create(Role resource) {
+        ResourceManager.kubeClient().namespace(resource.getMetadata().getNamespace()).createOrReplaceRole(resource);
+    }
+    @Override
+    public void delete(Role resource) {
+        ResourceManager.kubeClient().namespace(resource.getMetadata().getNamespace()).deleteRole(resource.getMetadata().getName());
```
see-quick, line 34: I think better would be use deleteRole with signature `deleteRole(namespaceName, resourceName)` inside KubeClient. I know in my refactor that these types of methods without specifying namespace explicitly end up with race-condition...

Frawless: Done

systemtest/src/test/java/io/strimzi/systemtest/metrics/MetricsST.java
```diff
@@ -597,16 +597,14 @@ void setupEnvironment(ExtensionContext extensionContext) throws Exception {
...
-        install.applyClusterOperatorInstallFiles(SECOND_NAMESPACE);
 
         cluster.setNamespace(FIRST_NAMESPACE);
 
         install = new SetupClusterOperator.SetupClusterOperatorBuilder()
             .withExtensionContext(extensionContext)
             .withNamespace(FIRST_NAMESPACE)
             .withWatchingNamespaces(FIRST_NAMESPACE + "," + SECOND_NAMESPACE)
-            .withBindingsNamespaces(Collections.singletonList(FIRST_NAMESPACE))
+            .withBindingsNamespaces(Arrays.asList(FIRST_NAMESPACE, SECOND_NAMESPACE))
```
scholzj, line 607: `List.of(...)`?

scholzj (approved):
If you want to merge this, I'm fine with it. However, I wonder a bit if it is wise. If the cluster is not clean and needs the `--replace` ... can you trust it to execute the tests? It could mean a lot of things ... e.g. another Strimzi operator installed somewhere which would interfere with the tests etc.

Frawless:
@scholzj I think we can even remove it now. The issue was caused by creating ClusterRoleBindings by applying yaml files instead of letting helm do it. I will remove it and also working your comment. Then tests again just for sure.

PR #5281, opened 2021-07-09
[MO] - [system tests] -> fix for testKafkaConnectAndConnectorStatus

see-quick:

### Type of change

- Bugfix

### Description

This PR fix race condition in `testKafkaConnectAndConnectorStatus` test case. The problem here was that when we configure KafkaConnetor with additional config:

```
.editSpec()
   .addToConfig("topic", EXAMPLE_TOPIC_NAME)
.endSpec()
```

and afterwards, check `KafkaConnectorStatus` there was a time window that this status gets updated. So I decrease the TopicReconciliation interval from defaults 120s to 15s and after 5 runs tests always passed. Otherwise with default reconciliation interval test fails once a time...which is not an acceptable behaviour...

### Checklist

### DIFF
--- a/systemtest/src/test/java/io/strimzi/systemtest/operators/CustomResourceStatusST.java
+++ b/systemtest/src/test/java/io/strimzi/systemtest/operators/CustomResourceStatusST.java
@@ -268,8 +268,6 @@ class CustomResourceStatusST extends AbstractST {
             .endSpec()
             .build());
 
-
-
         assertKafkaConnectStatus(1, connectUrl);
         assertKafkaConnectorStatus(CUSTOM_RESOURCE_STATUS_CLUSTER_NAME, 1, "RUNNING|UNASSIGNED", 0, "RUNNING", "source", List.of(EXAMPLE_TOPIC_NAME));
 
@@ -577,13 +575,13 @@ class CustomResourceStatusST extends AbstractST {
     void assertKafkaConnectorStatus(String clusterName, long expectedObservedGeneration, String connectorStates, int taskId, String taskState, String type, List<String> topics) {
         TestUtils.waitFor("KafkaConnector has following values", Constants.GLOBAL_POLL_INTERVAL, Constants.GLOBAL_TIMEOUT,
             () -> {
-                boolean verificationFormula = false;
+                boolean isKafkaConnectorStatusCorrect = false;
 
                 KafkaConnectorStatus kafkaConnectorStatus = KafkaConnectorResource.kafkaConnectorClient().inNamespace(NAMESPACE).withName(CUSTOM_RESOURCE_STATUS_CLUSTER_NAME).get().getStatus();
                 Map<String, Object> connectorStatus = kafkaConnectorStatus.getConnectorStatus();
                 String currentState = ((LinkedHashMap<String, String>) connectorStatus.get("connector")).get("state");
 
-                verificationFormula =
+                isKafkaConnectorStatusCorrect =
                     kafkaConnectorStatus.getObservedGeneration() == expectedObservedGeneration &&
                         connectorStates.contains(currentState) &&
                         connectorStatus.get("name").equals(CUSTOM_RESOURCE_STATUS_CLUSTER_NAME) &&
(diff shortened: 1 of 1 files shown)


systemtest/src/test/java/io/strimzi/systemtest/operators/CustomResourceStatusST.java
```diff
@@ -569,15 +573,24 @@ void validateConnectPlugins(List<ConnectorPlugin> pluginsList) {
...
-        KafkaConnectorStatus kafkaConnectorStatus = KafkaConnectorResource.kafkaConnectorClient().inNamespace(NAMESPACE).withName(CUSTOM_RESOURCE_STATUS_CLUSTER_NAME).get().getStatus();
-        assertThat(kafkaConnectorStatus.getObservedGeneration(), is(expectedObservedGeneration));
-        Map<String, Object> connectorStatus = kafkaConnectorStatus.getConnectorStatus();
-        String currentState = ((LinkedHashMap<String, String>) connectorStatus.get("connector")).get("state");
-        assertThat(connectorStates, containsString(currentState));
-        assertThat(connectorStatus.get("name"), is(CUSTOM_RESOURCE_STATUS_CLUSTER_NAME));
-        assertThat(connectorStatus.get("type"), is(type));
-        assertThat(connectorStatus.get("tasks"), notNullValue());
-        assertThat(kafkaConnectorStatus.getTopics(), is(topics));
+        TestUtils.waitFor("KafkaConnector has following values", Constants.GLOBAL_POLL_INTERVAL, Constants.GLOBAL_TIMEOUT,
```
Frawless, line 576: maybe you could print the expected values as well?

Frawless (approved):
Just one nit

scholzj (commented):
Can you please explain what exactly is the problem and why does this solve it in more detail? Because I do not get it. The topic operator has nothing to do with the connector status. So we need to understand this to make sure it does not indicate some bug.

see-quick:
In that tests we setup KafkaConnector with the following:

```
resourceManager.createResource(extensionContext, KafkaConnectorTemplates.kafkaConnector(CUSTOM_RESOURCE_STATUS_CLUSTER_NAME)
            .editSpec()
                .addToConfig("topic", EXAMPLE_TOPIC_NAME)
            .endSpec()
            .build());
```

and then afterwards checking KafkaConnectorStatus that the `EXAMPLE_TOPIC_NAME` is there. The default `TopicOperator` reconciliation is 120s I have changed it to 15s because I assume that after 120s seconds that topic inside the config of KafkaConnector will not be propagated once a time (race condition). But when I set the reconciliation interval to 15s test case pass. 
image

scholzj:
> The default TopicOperator reconciliation is 120s I have changed it to 15s because I assume that after 120s seconds that topic inside the config of KafkaConnector will not be propagated once a time (race condition). But when I set the reconciliation interval to 15s test case pass.

This is an assumption which would need to be explained. Topic Operator is not involved anywhere in the path which you described. It does not update the connector status with the topic and Connect does not care whether the KafkaTopic exists or not. So it makes little sense.

see-quick:
As we discussed on the community call. Here is a reproducer for the fails:

1. Deploy latest Cluster operator with operation timeout SHORT (STRIMZI_OPERATION_TIMEOUT_MS=Duration.ofSeconds(30).toMilis())
2. Deploy Kafka cluster with plain, tls and external node port listener  
3. Deploy Kafka connect with annotations set to `strimzi.io/use-connector-resources -> true`
4. Deploy Kafka connector with additional config set to `("topic", "example-topic-name")`\
5. Immediately fetch Kafka connector status and topics list has to be `["example-topic-name"]`, but sometimes it's empty  `[ ]`

`Note` that sometimes it passes what makes it flaky...

image

scholzj:
I was trying to reproduce this, but I failed. This part:

> Immediately fetch Kafka connector status and topics list has to be ["example-topic-name"], but sometimes it's empty [ ]

is true, but this is expected. The topics are usually provided by Kafka only later. So they show up only in the next reconciliation of the Cluster Operator. This is not handled in the old code which does not wait for it. So if you add the expected topics into it, it would fail.

But I did not managed to reproduce any relationship to the Topic Operator. It seems to be bound purely to the CO reconciliation in my manual tests. And if I follow your logic, disabling the TO or setting the reconciliation should make the topics not show at all. But that does not happen either.

I also checked out your code and tried to run your test. But it seems to pass for me regardless whether the TO reco interval is set to 15 seconds, 1500 seconds or not set at all.

So I don't know, as I said - you are right that the topics are not there right from the start, but that is IMHO expected. And I didn't managed to find any connection to the TO interval.

PR #5285, opened 2021-07-11
Tell User Operator whether Kafka Admin API can be used to manage ACLs or not

scholzj:
### Type of change

- Bugfix

### Description

In Kafka, the Admin API ACL management delegates to the configured ACL authorizer class. However, it is fairly common the the authorizer does not implement these methods. This applies especially in cases where the authorization data are stored in completely separate system such as with the OPA or Keycloak authorizers.

Today, when the `KafkaUser` CR contains the authorization section, the User Operator is always tries to set the ACLs using the Kafka Admin API. While managing the ALCs, it checks for some specific exceptions to detect if the ACL management is not supported and provide more specific errors to the user. But this solution is not perfect since it still calls the Admin API again and again. That can cause unnecessary load to the Kafka cluster but also means that exceptions and errors might be printed in the Kafka logs.

This PR introduces a better solution - it adds a new configuration for the User Operator to indicate whether the configured authorizer does or does not support the management of ACLs using the Kafka Admin API. The User Operator uses the information and either uses the Admin API to manage the ACLs or checks if the resource contains any ACLs and throws an `InvalidResourceConiguration` exception. This reduces the unnecessary Admin API calls and errors int he broker logs.

When the UO is deployed via the CO, it is configured automatically based on the authorization configuration. For OPA authorization, it is always disabled. For Keycloak authorization, it is enabled only when delegation to the simple authorizer is enabled. For simple authorizer it is always enabled. And for the custom authorizer, user can configure whether the custom authorizer implementation does or doesn't support the Admin API.

For the standalone UO, users have to configure this option themselves. The default value is `true` to enable it - this helps to maintain the backwards compatibility => for existing users, this will work as before unless they start using the new configuration value.

This Pr also updates the docs - it also adds the docs for the `STRIMZI_SECRET_PREFIX` option which seems to be missing.

### Checklist

- [x] Write tests
- [x] Update documentation

### DIFF
--- a/api/src/main/java/io/strimzi/api/kafka/model/KafkaAuthorization.java
+++ b/api/src/main/java/io/strimzi/api/kafka/model/KafkaAuthorization.java
@@ -41,6 +41,13 @@ public abstract class KafkaAuthorization implements UnknownPropertyPreserving, S
             "`custom` authorization type uses user-provided implementation for authorization.")
     public abstract String getType();
 
+    /**
+     * Indicates whether the Authorizer supports the Kafka Admin API to manage ACLs or not.
+     *
+     * @return True if ACLs can be managed using Kafka Admin API. False otherwise.
+     */
+    public abstract boolean supportsAdminApi();
+
     @Override
     public Map<String, Object> getAdditionalProperties() {
         return this.additionalProperties;
--- a/documentation/modules/con-securing-client-authorization.adoc
+++ b/documentation/modules/con-securing-client-authorization.adoc
@@ -9,11 +9,20 @@ User authorization is configured using the `authorization` property in `KafkaUse
 The authorization type enabled for a user is specified using the `type` field.
 
 To use simple authorization, you set the `type` property to `simple` in `KafkaUser.spec.authorization`.
-Simple authorization uses the default Kafka authorization plugin, `AclAuthorizer`.
+The simple authorization will use the Kafka Admin API to manage the ACL rules inside your Kafka cluster.
+Whether the ACL management in the User Operator is enabled or not depends on your authorization configuration in the Kafka cluster.
 
-Alternatively, you can use xref:type-KafkaAuthorizationOpa-reference[OPA authorization],
-or if you are already using OAuth 2.0 token based authentication,
-you can also use xref:assembly-oauth-authorization_str[OAuth 2.0 authorization].
+* For simple authorization, ACL management is always enabled.
+* For OPA authorization, ACL management is always disabled and authorization rules have to be configured in the OPA server.
+* For Keycloak authorization, you can manage the ACL rules directly in Keycloak.
+  You can also delegate authorization to the simple authorizer as a fallback option in the configuration.
+  When the delegation to the simple authorizer is enabled, User Operator will enable management of ACL rules as well.
+* When using custom authorization plugin, use the `supportsAdminApi` field in `.spec.kafka.authorization` in the `Kafka` custom resource to enable or disable the support.
+
+If ACL managment is not enabled, Strimzi rejects the resource in case it contains any ACL rules.
+
+If you're using a standalone deployment of the User Operator, the ACL management is enabled by default.
+You can disable it using the `STRIMZI_ACLS_ADMIN_API_SUPPORTED` environment variable.
 
 If no authorization is specified, the User Operator does not provision any access rights for the user.
 Whether such a `KafkaUser` can still access resources depends on the authorizer being used.
--- a/documentation/modules/con-securing-client-authorization.adoc
+++ b/documentation/modules/con-securing-client-authorization.adoc
@@ -10,18 +10,19 @@ The authorization type enabled for a user is specified using the `type` field.
 
 To use simple authorization, you set the `type` property to `simple` in `KafkaUser.spec.authorization`.
 The simple authorization will use the Kafka Admin API to manage the ACL rules inside your Kafka cluster.
-Whether the ACL management in the User Operator is enabled or not depends on your authorization configuration in the Kafka cluster.
+Whether ACL management in the User Operator is enabled or not depends on your authorization configuration in the Kafka cluster.
 
 * For simple authorization, ACL management is always enabled.
-* For OPA authorization, ACL management is always disabled and authorization rules have to be configured in the OPA server.
+* For OPA authorization, ACL management is always disabled.
+  Authorization rules are configured in the OPA server.
 * For Keycloak authorization, you can manage the ACL rules directly in Keycloak.
   You can also delegate authorization to the simple authorizer as a fallback option in the configuration.
-  When the delegation to the simple authorizer is enabled, User Operator will enable management of ACL rules as well.
-* When using custom authorization plugin, use the `supportsAdminApi` field in `.spec.kafka.authorization` in the `Kafka` custom resource to enable or disable the support.
+  When delegation to the simple authorizer is enabled, the User Operator will enable management of ACL rules as well.
+* For custom authorization using a custom authorization plugin, use the `supportsAdminApi` property in the `.spec.kafka.authorization` configuration of the `Kafka` custom resource to enable or disable the support.
 
-If ACL managment is not enabled, Strimzi rejects the resource in case it contains any ACL rules.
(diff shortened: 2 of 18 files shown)


api/src/main/java/io/strimzi/api/kafka/model/KafkaAuthorization.java
```diff
@@ -41,6 +41,13 @@
             "`custom` authorization type uses user-provided implementation for authorization.")
     public abstract String getType();
 
+    /**
+     * Indicates whether the Authorizer supports the Kafka Admin API to manage ACLs or not.
+     *
+     * @return True if ACLs can be managed using Kafka Admin API. False otherwise.
+     */
+    public abstract boolean supportsAdminApi();
```
tombentley, line 49: Putting this in the baseclass means that the schema will suggest that each of the subtypes support this property, which is not true: Only the custom one supports a property which can actually be written. The schema ought to reflect this fact.

scholzj: That will happen anyway, or? The Schema does not distinguish to which type the field belongs. You can see all the fields in one long list and I do not think it is there because of this but because of the field in `KafkaAuthorizationCustom`.

tombentley: Oh yeah, you're right, because of the rubbish way we try to simulate inheritance in the JSON schema. Yuck.

scholzj: So, I guess there is not much we can do about it here and it should not matter?

documentation/modules/deploying/proc-deploy-user-operator-standalone.adoc
```diff
@@ -98,6 +102,11 @@ The default is `365` days.
 The default is `30` days to initiate certificate renewal before the old certificates expire.
 <14> (Optional) The Java options used by the JVM running the User Operator
 <15> (Optional) The debugging (`-D`) options set for the User Operator
+<16> (Optional) Prefix which will be used in the names of Kubernetes secrets created by the User Operator
```
PaulRMellor, line 105:
```suggestion
<16> (Optional) Prefix for the names of Kubernetes secrets created by the User Operator
```

documentation/modules/deploying/proc-deploy-user-operator-standalone.adoc
```diff
@@ -98,6 +102,11 @@ The default is `365` days.
...
+<17> (Optional) Indicates whether the Kafka cluster supports management of authorization ACL rules using the Kafka Admin API.
+When set to `false`, the User Operator will reject all resources with `simple` authorization ACL rules.
+This helps to avoid unnecessary exceptions in the Kafka cluster logs.
+The default is `true`.
```
PaulRMellor, line 109: Maybe some context about when to set and not set this in https://strimzi.io/docs/operators/in-development/using.html#con-securing-client-authorization-str

> **ACL management**
> Implementation of ACL management depends on your client authorization configuration. 
> 
> - For simple authorization, it's always enabled. 
> - For OPA authorization, it's always disabled. 
> - For Keycloak authorization, you can delegate to the simple authorizer in the configuration. This means you can use Keycloak authorization and ACLs. 
>  
> If ACL managment is not enabled, Strimzi throws an error if a resource contains ACLs.
> 
> If you're using a standalone deployment of the User Operator, use configuration to manage ACLs. Enabled by default, you can disable ACL management using the `STRIMZI_ACLS_ADMIN_API_SUPPORTED` environment variable.

PaulRMellor (approved):
Looks good. I've made a suggestion for adding doc content in another part of the guide. I can pick that up if you think we need it. Maybe we'll want to add something similar in the deploying guide too.

tombentley (approved):

documentation/modules/deploying/proc-deploy-user-operator-standalone.adoc
scholzj: @PaulRMellor I updated it, but didn't use exactly the same text as you suggested above. Could you please have another look?

documentation/modules/con-securing-client-authorization.adoc
```diff
@@ -9,11 +9,20 @@ User authorization is configured using the `authorization` property in `KafkaUse
 The authorization type enabled for a user is specified using the `type` field.
 
 To use simple authorization, you set the `type` property to `simple` in `KafkaUser.spec.authorization`.
-Simple authorization uses the default Kafka authorization plugin, `AclAuthorizer`.
+The simple authorization will use the Kafka Admin API to manage the ACL rules inside your Kafka cluster.
+Whether the ACL management in the User Operator is enabled or not depends on your authorization configuration in the Kafka cluster.
```
PaulRMellor, line 13:
```suggestion
Whether ACL management in the User Operator is enabled or not depends on your authorization configuration in the Kafka cluster.
```

documentation/modules/con-securing-client-authorization.adoc
```diff
@@ -9,11 +9,20 @@ User authorization is configured using the `authorization` property in `KafkaUse
...
 
-Alternatively, you can use xref:type-KafkaAuthorizationOpa-reference[OPA authorization],
-or if you are already using OAuth 2.0 token based authentication,
-you can also use xref:assembly-oauth-authorization_str[OAuth 2.0 authorization].
+* For simple authorization, ACL management is always enabled.
+* For OPA authorization, ACL management is always disabled and authorization rules have to be configured in the OPA server.
```
PaulRMellor, line 16:
```suggestion
* For OPA authorization, ACL management is always disabled. Authorization rules are configured in the OPA server.
```

documentation/modules/con-securing-client-authorization.adoc
```diff
@@ -9,11 +9,20 @@ User authorization is configured using the `authorization` property in `KafkaUse
...
+* For Keycloak authorization, you can manage the ACL rules directly in Keycloak.
+  You can also delegate authorization to the simple authorizer as a fallback option in the configuration.
+  When the delegation to the simple authorizer is enabled, User Operator will enable management of ACL rules as well.
```
PaulRMellor, line 19:
```suggestion
  When delegation to the simple authorizer is enabled, the User Operator will enable management of ACL rules as well.
```

documentation/modules/con-securing-client-authorization.adoc
```diff
@@ -9,11 +9,20 @@ User authorization is configured using the `authorization` property in `KafkaUse
...
+* When using custom authorization plugin, use the `supportsAdminApi` field in `.spec.kafka.authorization` in the `Kafka` custom resource to enable or disable the support.
```
PaulRMellor, line 20:
```suggestion
* For custom authorization using a custom authorization plugin, use the `supportsAdminApi` property in the `.spec.kafka.authorization` configuration of the `Kafka` custom resource to enable or disable the support.
```

documentation/modules/con-securing-client-authorization.adoc
```diff
@@ -9,11 +9,20 @@ User authorization is configured using the `authorization` property in `KafkaUse
...
+
+If ACL managment is not enabled, Strimzi rejects the resource in case it contains any ACL rules.
```
PaulRMellor, line 22:
```suggestion
If ACL managment is not enabled, Strimzi rejects a resource if it contains any ACL rules.
```

documentation/modules/con-securing-client-authorization.adoc
```diff
@@ -9,11 +9,20 @@ User authorization is configured using the `authorization` property in `KafkaUse
...
+
+If you're using a standalone deployment of the User Operator, the ACL management is enabled by default.
```
PaulRMellor, line 24:
```suggestion
If you're using a standalone deployment of the User Operator, ACL management is enabled by default.
```

documentation/modules/deploying/proc-deploy-user-operator-standalone.adoc
PaulRMellor: @scholzj  Thanks Jakub. The custom option completes it. Looks good. Just a few minor suggestions.

PR #5286, opened 2021-07-12
Config-model-generator execution on Maven build

LiamClarkeNZ:
### Type of change

- Refactoring

### Description

Binds generation of Kafka config model files to a Maven build phase to prevent cluster-operator tests failing in a couple of scenarios when `maven verify/install` etc. are been run without running `make` goals first. 

Scenarios were:
1) Tests failing in cluster-operator because no Kafka config files existed in its resources directory
2) Tests failing in cluster-operator because config files for now unsupported Kafka versions still exist in its resource directory.

### Checklist

- [ ] Write tests
- [ ] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/config-model-generator/Makefile
+++ b/config-model-generator/Makefile
@@ -7,7 +7,6 @@ docker_tag:
 docker_push:
 
 java_build:
-	./build-config-models.sh build
 
 spotbugs:
 	mvn $(MVN_ARGS) spotbugs:check
@@ -18,7 +17,7 @@ clean:
 	rm ../cluster-operator/src/main/resources/kafka-*-config-model.json || true
 	./build-config-models.sh clean
 
-#include ../Makefile.maven
+include ../Makefile.maven
 
 .PHONY: build clean release
 
--- a/config-model-generator/pom.xml
+++ b/config-model-generator/pom.xml
@@ -67,6 +67,32 @@
                         <argument>${config-model-file}</argument>
                     </arguments>
                 </configuration>
+                <executions>
+                    <execution>
+                        <!-- This execution is included here to generate the required files as part of the regular
+                             Maven build so that a successful `mvn clean install` isn't contingent on running make first.
+                             This helps with two scenarios:
+                                1) Tests failing in cluster-operator because no Kafka configuration existed
+                                2) Tests failing in cluster-operator because config files for now unsupported Kafka versions still exist -->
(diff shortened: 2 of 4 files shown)


config-model-generator/Makefile
```diff
@@ -7,7 +7,6 @@ docker_tag:
 docker_push:
 
 java_build:
-	./build-config-models.sh build
```
scholzj, line 10: How does this empty target work? Doesn't it exclude the Maven build from happening when `make all` is used?

config-model-generator/pom.xml
```diff
@@ -67,6 +67,32 @@
...
+                             once for every supported version, and the existing shell scripts already handle determing what those versions are -->
+                        <configuration>
+                            <executable>${project.basedir}${file.separator}build-config-models.sh</executable>
+                            <arguments>
+                                <argument>build</argument>
+                            </arguments>
+                            <workingDirectory>${project.basedir}</workingDirectory>
+                        </configuration>
+                    </execution>
+                </executions>
```
scholzj, line 95: Should `clean` be also done via Maven then? It seems a bit weird to have only one part driven by Maven.

config-model-generator/Makefile
LiamClarkeNZ: Ahhh, good point.

config-model-generator/pom.xml
LiamClarkeNZ: Certainly can. However, it looks like this change broke CodeQL, and given it was only a minor annoyance (that I only really noticed because I had to rebuild my laptop), not sure it's worth merging if it's breaking tooling. 

What do you think?

scholzj: You mean to just close this PR and leave it be as it is?

LiamClarkeNZ: Yep, forget about the PR - it's breaking CodeQL because `autobuild` runs with a clean local repo, and only runs to `package`, not `install`, so adding the execution  into the Maven build for `config-model-generator` will actually break this for anyone running anything less than `install` in the Maven lifecycle on a clean repo, which really defeats the purpose of it.  

It's only a minor problem occasionally encountered. A better solution might be to document in the dev guide that running `make java_build` after first cloning the repository ensures a straightforward Maven experience.

PR #5299, opened 2021-07-14
Add pre-check to self-closing-watch to handle missed events

scholzj:
### Type of change

- Bugfix

### Description

When deleting resources we usually go through the following process:
1) Check if the resource exists and if it does ...
2) ... Create a deletion watch
3) ... Delete the resource
4) ... Wait for steps 2 and 3 to succeed

In Kubernetes API, each of these steps is separate. So it can happen that the step 1 says the resource still exists, but in steps 2 and 3 it will be already deleted. This seems to happen regularly in the User Operator. Unlike CO, the UO handles the deletion of resources and does not rely purely on garbage collection. So the user secrets seem to be deleted in parallel by GC and by UO. That seems to very often raise the race condition between step 1 and steps 2 and 3 and cause the reconciliation to fail.

It fails in two different way:
a) The deletion watch is started only after the resource is deleted. So it never receives the `DELETE` event and completes
b) The `delete()` call returns `false` because the resource does not exist anymore.

To handle a), this PR adds additional pre-check function into the self-closing-watch. The pre-check function is called after the watch is opened and checks whether the resource is deleted. So even if the resource is deleted before the watch is created and we get no deletion event, the pre-check should find it and close the watch. In the worst case, both pre-check and the close the watch which should not be a problem.

For b), we should treat the `false` result as success. If the resource was not found, then maybe we didn't deleted it ... but who cares, the desired state was achieved. This behaviour is also changed in this PR.

### Checklist

- [x] Write tests

### DIFF
--- a/operator-common/src/main/java/io/strimzi/operator/common/operator/resource/ResourceSupport.java
+++ b/operator-common/src/main/java/io/strimzi/operator/common/operator/resource/ResourceSupport.java
@@ -93,17 +93,18 @@ public class ResourceSupport {
      * returning a Future which completes when {@code watchFn} returns non-null
      * to some event on the watchable, or after a timeout.
      *
-     * The given {@code watchFn} will be invoked on a worked thread when the
+     * The given {@code watchFn} will be invoked on a worker thread when the
      * Kubernetes resources changes, so may block.
      * When the {@code watchFn} returns non-null the watch will be closed and then
      * the future returned from this method will be completed on the context thread.
      *
      * In some cases such as resource deletion, it might happen that the resource is deleted already before the watch is
-     * started and as a result the watch never completes. The {@code preCheckFn} will be invoked on a worked thread
+     * started and as a result the watch never completes. The {@code preCheckFn} will be invoked on a worker thread
      * after the watch has been created. It is expected to double check if we still need to wait for the watch to fire.
      * When the {@code preCheckFn} returns non-null the watch will be closed and the future returned from this method
-     * will be completed on the context thread. In the deletion example described above, the {@code preCheckFn} can
-     * check if the resource still exists and close the watch in case it was already deleted.
+     * will be completed with the result of the {@code preCheckFn} on the context thread. In the deletion example
+     * described above, the {@code preCheckFn} can check if the resource still exists and close the watch in case it was
+     * already deleted.
      *
      * @param reconciliation Reconciliation marker used for logging
      * @param watchable The watchable - used to watch the resource.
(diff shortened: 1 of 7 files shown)


scholzj:
@tombentley Both a) and b) happen in parallel, so there is just one test. I also removed the blocking execution.

operator-common/src/main/java/io/strimzi/operator/common/operator/resource/ResourceSupport.java
```diff
@@ -96,21 +98,34 @@ private Throwable collectCauses(AsyncResult<?> primary,
      * When the {@code watchFn} returns non-null the watch will be closed and then
      * the future returned from this method will be completed on the context thread.
      *
-     * @param watchable The watchable.
+     * In some cases such as resource deletion, it might happen that the resource is deleted already before the watch is
+     * started and as a result the watch never completes. The {@code preCheckFn} will be invoked on a worked thread
```
tombentley, line 102:
```suggestion
     * started and as a result the watch never completes. The {@code preCheckFn} will be invoked on a worker thread
```

operator-common/src/main/java/io/strimzi/operator/common/operator/resource/ResourceSupport.java
```diff
@@ -96,21 +98,34 @@ private Throwable collectCauses(AsyncResult<?> primary,
...
+     * after the watch has been created. It is expected to double check if we still need to wait for the watch to fire.
+     * When the {@code preCheckFn} returns non-null the watch will be closed and the future returned from this method
+     * will be completed on the context thread. In the deletion example described above, the {@code preCheckFn} can
```
tombentley, line 105:
```suggestion
     * will be completed with the result of the {@code preCheckFn} on the context thread. In the deletion example described above, the {@code preCheckFn} can
```

operator-common/src/test/java/io/strimzi/operator/common/operator/resource/AbstractNonNamespacedResourceOperatorTest.java
```diff
@@ -465,12 +466,57 @@ public void testReconcileDeletionThrowsWhenDeleteMethodReturnsFalse(VertxTestCon
...
+    // This tests the pre-check which should stop the self-closing-watch in case the resource is deleted before the
+    // watch is opened.
+    @Test
+    @SuppressWarnings("unchecked")
+    public void testReconcileDeleteDoesNotTimeoutWhenResourceIsAlreadyDeleted(VertxTestContext context) {
+        T resource = resource();
+        AtomicBoolean watchWasClosed = new AtomicBoolean(false);
+        Resource mockResource = mock(resourceType());
+
+        AtomicInteger getInvocations = new AtomicInteger(0);
```
tombentley, line 485: Rather than counting invocations why not create this mock after the mock for `watch()` and have the return from this method predicated on whether the watch was created?

scholzj: Please have a look if the change I made is what you meant.

scholzj:
@tombentley Does this look good now? ;-)

tombentley (approved):

PR #5318, opened 2021-07-19
[MO] - [system test] -> unify installation

see-quick:

### Type of change

- Enhancement / new feature
- Refactoring

### Description

This PR is trying to unify installation Cluster Operators in our system tests. Some tests cases using  installation with the following format:
```
resourceManager.createResource(extensionContext,
            new BundleResource.BundleResourceBuilder()
                 ...
```

This approach is in most cases bad because it explicitly deploys BundleFormat and completely ignores other types of installation such as Helm or Olm. 

Correct way how to deploy operator is:

```
install = new SetupClusterOperator.SetupClusterOperatorBuilder()
            .withExtensionContext(extensionContext)
                ...
```

### Checklist

### DIFF
--- a/systemtest/src/main/java/io/strimzi/systemtest/SetupClusterOperator.java
+++ b/systemtest/src/main/java/io/strimzi/systemtest/SetupClusterOperator.java
@@ -4,6 +4,7 @@
  */
 package io.strimzi.systemtest;
 
+import io.fabric8.kubernetes.api.model.EnvVar;
 import io.fabric8.kubernetes.api.model.rbac.ClusterRoleBinding;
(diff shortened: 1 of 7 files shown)


Frawless (approved):

scholzj:
FYI: I think that `SetupClusterOperator` and `SetupClusterOperatorBuilder` are really weird names. Wouldn't just `ClusterOperator` and `ClusterOperatorBuilder` be better? Naming a class after an action (_Setup Cluster Operator_) is weird. (Anyway, that was there before, so not an issue for this PR - just something I noticed)

PR #5333, opened 2021-07-20
ST: Change deployment of Roles/RoleBingins to fix namespace-rbac pipelines

Frawless:

### Type of change

- Bugfix

### Description

This PR fixes several issues:

- option to deploy cluster operator in test method itself 
- delete all operator files even when you create new clsuter operator in test method
- create roles/rolebindings properly when CO is in namespace-rbac scope

All tests should pass now.

### Checklist


systemtest/src/test/java/io/strimzi/systemtest/specific/SpecificST.java
```diff
@@ -262,16 +259,14 @@ public void testRackAwareConnectCorrectDeployment(ExtensionContext extensionCont
...
-                .withWatchingNamespaces(NAMESPACE)
-                .withOperationTimeout(CO_OPERATION_TIMEOUT_SHORT)
-                .withReconciliationInterval(Constants.RECONCILIATION_INTERVAL)
-                .buildBundleInstance()
-                .buildBundleDeployment()
-                .build());
+        install = new SetupClusterOperator.SetupClusterOperatorBuilder()
+            .withExtensionContext(extensionContext)
+            .withNamespace(NAMESPACE)
+            .withWatchingNamespaces(NAMESPACE)
```
see-quick, line 265: You can remove this `.withWatchingNamespaces(NAMESPACE)` because inside the constructor we assign default values which in this case would be `NAMESPACE`

see-quick (approved):

Issue #5293, opened 2021-07-13
strimzi_resource_state not updated to 0 when operator doesn't see a Kafka resource anymore due to changed selector label


ppatierno:
The `strimzi_resource_state` metric is used to report when a specific (named) custom resource (i.e. `Kafka` kind) is ready or not.
It seems there is a bug when using more than one Strimzi operator handing over a `Kafka` custom resource from one to the other changing the so called selector label.
For example, let's consider having Strimzi 0.22.0 and 0.23.0 operator instances running together and watching all namespaces in a Kubernetes cluster but using the `STRIMZI_CUSTOM_RESOURCE_SELECTOR` env var in order to specify a label for selecting the Kafka custom resource to take care of.

1. Let's deploy a Kafka CR labeled for 0.22.0 so that it will reconcile it updating the `strimzi_resource_state` metric to 1 when it's ready.
2. Let's change the label on the Kafka CR to 0.23.0; the Strimzi 0.23.0 operator will take care of it while 0.22.0 won't see it anymore but the `strimzi_resource_state` metric will still stay 1 in the Strimzi 0.22.0 (of course it will be 1 in the Strimzi 0.23.0 as it should be). 

In this scenario, some aggregated metrics on complex dashboards could be doubled.
Taking a look at code it seems that the readiness of a Kafka resource on the `strimzi_resource_state` metric is set to 0 (not ready) only when there are some exception but not when the Kafka resource is not handled by the operator anymore which actually means a DELETED event.

sknot-rh:
I tried to reproduce this. In the single namespace I created two deployments of the Strimzi. One had configured
```
            - name: STRIMZI_CUSTOM_RESOURCE_SELECTOR
              value: "strimzi.io/cluster=my-cluster-1"
```
the second one had configured
```
            - name: STRIMZI_CUSTOM_RESOURCE_SELECTOR
              value: "strimzi.io/cluster=my-cluster-2"
```
Then I created a Kafka resource with label 
```  labels:
    strimzi.io/cluster: my-cluster-1
```
As expected, the metric in the first operator was set to 
`strimzi_resources{kind="Kafka",selector="strimzi.io/cluster=my-cluster-1",} 1.0`
and in the second one to
`strimzi_resources{kind="Kafka",selector="strimzi.io/cluster=my-cluster-2",} 0.0`
Then, I changed Kafka resource label to 
```
  labels:
    strimzi.io/cluster: my-cluster-2
```
[...]

PR #5342, opened 2021-07-21
Fix resource state metric

sknot-rh:

### Type of change
- Bugfix

### Description
Changing resources selector label was ignored by CO.

Fixes https://github.com/strimzi/strimzi-kafka-operator/issues/5293
### Checklist
- [ ] Write tests
- [ ] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/operator-common/src/main/java/io/strimzi/operator/common/AbstractOperator.java
+++ b/operator-common/src/main/java/io/strimzi/operator/common/AbstractOperator.java
@@ -563,7 +563,7 @@ public abstract class AbstractOperator<
             LOGGER.debugCr(reconciliation, "Removed metric " + METRICS_PREFIX + "resource.state{}", key);
         }
 
-        if (cr != null) {
+        if (cr != null && Util.matchesSelector(selector(), cr)) {
             resourcesStateCounter.computeIfAbsent(key, tags ->
                     metrics.gauge(METRICS_PREFIX + "resource.state", "Current state of the resource: 1 ready, 0 fail", metricTags)
             );


operator-common/src/main/java/io/strimzi/operator/common/AbstractOperator.java
```diff
@@ -563,7 +563,7 @@ private void updateResourceState(Reconciliation reconciliation, boolean ready, T
             LOGGER.debugCr(reconciliation, "Removed metric " + METRICS_PREFIX + "resource.state{}", key);
         }
 
-        if (cr != null) {
+        if (cr != null && Util.matchesSelector(selector(), cr)) {
```
scholzj, line 566: Why is this ever called for resources not matching the selector?

sknot-rh: I guess because we need to delete metrics for the resources not matching the selector.

scholzj: But this does not even have an else. So how does it delete the metric?

sknot-rh: In the code above (line 562). It looks for the metric, removes it, and only if the CR is nonNull (and matches the selector), the metric is put back - updated.

scholzj: Ok. 

But I'm still not sure this fixes the issue. `updateResourceState` should be called at the end of the reconciliation. So the  change you did should catch situations when the CR moved out of the selector during the reconciliation it self. However, that:
* Should not be done AFAIK if the handover happens properly and the reconciliation is paused, or? It can be done only when it is done without pausing in which case all kind of bad things can happen
* If that is really the case where it happened in #5293, the metric should be there only for a short time and be removed later anyway.

Isn't it the case that we need to delete the metrics when the resource is deleted?

scholzj:
@ppatierno Can you please have a look at this? This was your issue.

scholzj (approved):

Issue #3663, opened 2020-09-16
[Enhancement] Add histogram latency buckets for reconciliations duration


ppatierno:
The operators expose a `strimzi_reconciliations_duration_XXXX` providing max, avg, and so on but having histogram with latency buckets would be better in order to avoid spurious timing providing a wrong average for example.
Think about more reconciles taking a few seconds but just one with a bigger value screweing up an average and the other way around. From histogram you can get percentiles that are really useful in these scenarios.

ShubhamRwt:
@ppatierno Hello, Is this issue still open?

ppatierno:
I would say yes, it could be useful imho.

scholzj:
Done in #5349

PR #5349, opened 2021-07-22
Adding histogram latency buckets for reconciliation duration

ShubhamRwt:

### Type of change

_Select the type of your PR_

- Enhancement / new feature

### Description

This PR address the issue mentioned in #3663. This PR adds histogram latency buckets for `strimzi_reconciliation_duration` metrics. After the addition of this code this we can get the metrics like this :-  

`strimzi_reconciliations_duration_seconds_bucket{kind="KafkaRebalance",selector="",le="0.01",} 0.0
strimzi_reconciliations_duration_seconds_bucket{kind="KafkaRebalance",selector="",le="0.025",} 0.0
strimzi_reconciliations_duration_seconds_bucket{kind="KafkaRebalance",selector="",le="0.05",} 0.0
strimzi_reconciliations_duration_seconds_bucket{kind="KafkaRebalance",selector="",le="0.1",} 0.0
strimzi_reconciliations_duration_seconds_bucket{kind="KafkaRebalance",selector="",le="0.5",} 0.0
strimzi_reconciliations_duration_seconds_bucket{kind="KafkaRebalance",selector="",le="1.0",} 0.0
strimzi_reconciliations_duration_seconds_bucket{kind="KafkaRebalance",selector="",le="5.0",} 0.0
strimzi_reconciliations_duration_seconds_bucket{kind="KafkaRebalance",selector="",le="+Inf",} 0.0
strimzi_reconciliations_duration_seconds_count{kind="KafkaRebalance",selector="",} 0.0
strimzi_reconciliations_duration_seconds_sum{kind="KafkaRebalance",selector="",} 0.0 `

For complete metrics :-   
--> https://gist.github.com/ShubhamRwt/31eeb8992d869a87f8bcba61bbd8d787

### Checklist

- [ ] Write tests
- [ ] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/operator-common/src/main/java/io/strimzi/operator/common/MicrometerMetricsProvider.java
+++ b/operator-common/src/main/java/io/strimzi/operator/common/MicrometerMetricsProvider.java
@@ -11,6 +11,7 @@ import io.micrometer.core.instrument.Tags;
 import io.micrometer.core.instrument.Timer;
 import io.vertx.micrometer.backends.BackendRegistries;
 
+import java.time.Duration;
 import java.util.concurrent.atomic.AtomicInteger;
 
 /**
@@ -64,6 +65,7 @@ public class MicrometerMetricsProvider implements MetricsProvider {
     public Timer timer(String name, String description, Tags tags) {
         return Timer.builder(name)
                 .description(description)
+                .sla(Duration.ofMillis(10), Duration.ofMillis(25), Duration.ofMillis(50), Duration.ofMillis(100), Duration.ofMillis(500), Duration.ofMillis(1000), Duration.ofMillis(5000))
                 .tags(tags)
                 .register(metrics);
     }
--- a/operator-common/src/main/java/io/strimzi/operator/common/MicrometerMetricsProvider.java
+++ b/operator-common/src/main/java/io/strimzi/operator/common/MicrometerMetricsProvider.java
@@ -65,7 +65,7 @@ public class MicrometerMetricsProvider implements MetricsProvider {
     public Timer timer(String name, String description, Tags tags) {
         return Timer.builder(name)
                 .description(description)
-                .sla(Duration.ofMillis(10), Duration.ofMillis(25), Duration.ofMillis(50), Duration.ofMillis(100), Duration.ofMillis(500), Duration.ofMillis(1000), Duration.ofMillis(5000))
+                .sla(Duration.ofMillis(1000), Duration.ofMillis(5000), Duration.ofMillis(10000), Duration.ofMillis(30000), Duration.ofMillis(60000), Duration.ofMillis(120000), Duration.ofMillis(300000))
                 .tags(tags)
                 .register(metrics);
     }


operator-common/src/main/java/io/strimzi/operator/common/MicrometerMetricsProvider.java
```diff
@@ -64,6 +65,7 @@ public Counter counter(String name, String description, Tags tags) {
     public Timer timer(String name, String description, Tags tags) {
         return Timer.builder(name)
                 .description(description)
+                .sla(Duration.ofMillis(10), Duration.ofMillis(25), Duration.ofMillis(50), Duration.ofMillis(100), Duration.ofMillis(500), Duration.ofMillis(1000), Duration.ofMillis(5000))
```
ppatierno, line 68: I would not expect to see all these buckets with this level of granularity. I would say that everything under 500 or 1000 ms would be ok when the reconcile is fast without actually doing nothing or a few then we could have longer ones due to rolling pods, even over the minutes. Maybe 500, 1000, 30000, 60000, 120000, 240000 could have more sense? @scholzj wdyt?

scholzj: Well, it is hard. I guess different resources have different reconciliation times. I would maybe go for:
* 1s
* 5s
* 10s
* 30s
* 1m
* 2m
* 5m
* Inf

ppatierno: 5 min is our upper default limit on reconcile right? maybe that bucket can just be in "Inf", for this reason I was suggesting 4 min as last bucket. For the rest I am fine with your suggestions, they are more or less the same I did.

scholzj: I think 5 minutes bucket might be interesting ... 5+ minutes would suggests there is a high probability that something is timing out in the reconciliation. I'm fine if we want to add more between 2 and 5 minutes ... but I think the 5 minute is significant as the timeout and should be there.

ppatierno: but the bucket should be "less equal" so the 5 min bucket will be filled with values between 2m:01sec to 4m:59sec, everything above 5 min will be in Inf bucket.

scholzj: Yeah, exactly ... everything in the bucket above 5 minutes is suspicious. 2-5 minutes is is long, but not necessarily suspicious since it means that for sure nothing timed out.

ppatierno: Ok so just to clarify it will be interesting to check the "Inf" bucket not the "5m" bucket :-)
Anyway, I am fine with yours suggestions.

scholzj: Right ... I think the 5m-Inf bucket is more useful than 4m-Inf bucket.

ShubhamRwt:
Metrics log with new buckets --> https://gist.github.com/ShubhamRwt/e097b862013c089912e336994b4c6b09

scholzj:
@ShubhamRwt So, do you plan to add it to the Grafana dashboards as well?

ShubhamRwt:
Yeah Sure, Just need to look a bit on how the metrics are updated on grafana dashboard.

ShubhamRwt:
Is it just like how other metrics are added to the dashboard?I guess using .json files?  @scholzj

scholzj:
Well, in the PR it is the JSON file, yes: https://github.com/strimzi/strimzi-kafka-operator/blob/main/packaging/examples/metrics/grafana-dashboards/strimzi-operators.json

I normally load it into Grafana, edit it in Grafana through the UI, and export it. And after I export it I try to remove the unrelated changes which it does to the other fields.

ShubhamRwt:
> Well, in the PR it is the JSON file, yes: https://github.com/strimzi/strimzi-kafka-operator/blob/main/packaging/examples/metrics/grafana-dashboards/strimzi-operators.json
> 
> I normally load it into Grafana, edit it in Grafana through the UI, and export it. And after I export it I try to remove the unrelated changes which it does to the other fields.

Got it. I will work on changes later today and try to get them done.

ShubhamRwt:
image
@ppatierno @scholzj  Hello, Is this Histogram good to go with?

Or there is one more way of representing the buckets like this (No the histogram way)-->
image

scholzj:
I would expect something more like a distribution ... more something like this: 
image

ShubhamRwt:
Like this @scholzj 
image

scholzj:
No, that shows timestamps and not buckets.

ShubhamRwt:
image
@scholzj I have removed the time stamps from x axis and set them as histogram buckets

ShubhamRwt:
Should I keep the y axis as time ?

scholzj:
Well, you would need to adjust the buckets somehow to match what you have in the metrics. Also, it suggests you had 3 reconciliations below 10 seconds and 18 in the 300 seconds bucket in the last 5 minutes? That seems weird. It also also missing the split based on different resources.

ppatierno:
If there are a lot of problems getting a good looking Grafana, I think that we could move forward and open an issue for it.
Maybe someone will contribute to it if he finds it useful. 
Anyway I think that exposing the metrics is really useful, because someone else could not only using his own Grafana but even using the metrics just for alerts without visualizing data on dashboards.

ppatierno (approved):

scholzj:
> If there are a lot of problems getting a good looking Grafana, I think that we could move forward and open an issue for it.
Maybe someone will contribute to it if he finds it useful.
Anyway I think that exposing the metrics is really useful, because someone else could not only using his own Grafana but even using the metrics just for alerts without visualizing data on dashboards.

I don't agree with this. It looks like we right now have no idea how to use this metric. So I think it is questionable why it is needed / how it is useful. 

Also, AFAIU I think it removes the existing maximum reconciliation time metric used in the dashboard ... or does that stay?

ShubhamRwt:
```strimzi_reconciliations_duration_seconds_max{kind="KafkaRebalance",selector="",} 0.0
strimzi_reconciliations_duration_seconds_max{kind="Kafka",selector="",} 0.0
strimzi_reconciliations_duration_seconds_max{kind="KafkaConnect",selector="",} 0.0
strimzi_reconciliations_duration_seconds_max{kind="KafkaMirrorMaker2",selector="",} 0.0
strimzi_reconciliations_duration_seconds_max{kind="KafkaMirrorMaker",selector="",} 0.0
strimzi_reconciliations_duration_seconds_max{kind="KafkaConnector",selector="",} 0.0
strimzi_reconciliations_duration_seconds_max{kind="KafkaBridge",selector="",} 0.0```  , these metrics @scholzj ?

ppatierno:
> It looks like we right now have no idea how to use this metric

I already explained the usefulness of these metrics compared to have just avg and max. I don't think that metrics should we always on a dashboard or at least it's not the only way I know they are used. People could even just set alerts on them with no visualization or just using Prometheus UI but not Grafana.

scholzj:
> I already explained the usefulness of these metrics compared to have just avg and max. I don't think that metrics should we always on a dashboard or at least it's not the only way I know they are used. People could even just set alerts on them with no visualization or just using Prometheus UI but not Grafana.

Great, then please provide the example alerts. I'm not sure how do you do alerting on histograms, but that is exactly what this needs, an use-case which shows why is it useful. I do not think they are useful unless we have a way to use them. So we should provide it.

scholzj (approved):

Issue #4649, opened 2021-03-26
Observability of rolling restarts in Grafana


tombentley:
It would be useful to expose rolling restarts and their causes in the Grafana dashboard somehow, maybe through an [alert panel](https://grafana.com/grafana/plugins/alertlist/), or something similar. Having the context of recent restarts in Grafana is very useful in understanding the context of the rest of a visualisations.

This is similar to #4648, but not mutually exclusive.

LiamClarkeNZ:
Kia ora @tombentley

The alert panel only shows failing alerts, it wouldn't provide any useful data for you. I've used the [Table panel ](https://grafana.com/grafana/plugins/table/) in the past for similar purposes

However, the issue is that Prometheus is terrible for recording discrete events. The only way to capture the info you need via Prom is in labels, and Prometheus performance degrades as label cardinality increases. Other back-ends like [Victoria]](https://victoriametrics.com) which claim to be Prometheus-compatible are better at handling high cardinality labels, but that's not really an option for this.

The other problem with Prometheus is that all data-types are some form of time-series, and in combination with scrape intervals, there's no guarantee that a given event will be scraped. So exposing these events as a Prometheus metric could be lossy.

The best solution I've found in the past for incorporating discrete events into Grafana is to use the [annotations API](https://grafana.com/docs/grafana/latest/http_api/annotations/). Would that be an acceptable solution?

A potential solution would involve configuring the cluster operator with a Grafana endpoint and auth token, and then having it make POSTs to the given Grafana (in a manner that doesn't harm the operator).  

Labels/tags can also be included for precise targeting of annotations on a given dashboard. There are some UI limitations to how much information can be included. For example, here's two examples. 

1. The first is an annotation showing a new version of an app was released. 
2. The second is an annotation showing a K8s config diff. You can see that there's limited space for text. 

However, a good way of approaching larger strings is to use a "correlation id" or similar - a UUID placed into the annotations and the logs, for easy grepping. I haven't yet checked out how big the strings describing restart reasons are in the Cluster Operator, but wanted to get your feedback on this approach before I dived to deep.

Also worth mentioning that annotation text in Grafana is HTML, so it's possible to link directly to relevant searches in an ELK stack or similar, depending on how far you want to integrate. If you can handle unseemly line breaks, then the string being directly in the annotation is a pretty straight-forward approach.

[...]

Issue #4648, opened 2021-03-26
Observability of rolling restarts via Events


tombentley:
Currently it's not very easy to see whether a particular pod was restarted by the operator and why such a CO-initiated restart was necessary. I think the only place this is really observable is in the CO logs. It would be good if this information was a little more accessible (searching the logs is tedious, especially when they're verbose, or the information might not be available depending on how the logging is configured).

Generating a Kubernetes `Event` against each pod that's rolled might be a better way of surfacing this. I think that's better than putting a condition in the CR status because it remains even after the next reconciliation. WDYT @strimzi/maintainers?

scholzj:
Done in #6663

PR #5373, opened 2021-07-29
Publish Strimzi initiated restarts to Kubernetes as events, and to Micrometer as a counter #4648 #4649

LiamClarkeNZ:
### Type of change

- Enhancement / new feature

### Description

When Strimzi chooses to restart a pod, the reason(s) why are now expressed with an enum value - this has a default note/message for when more human readable detail is required. Explicit messages can be  set, and these will override the default message. If multiple restarts reasons are recorded in one reconciliation - e.g., multiple CA certs are renewed, only one reason (e.g., CA_CERT_RENEWED) will be retained for publishing, but all explicit messages for that reason are appended. 

These reasons are then published to Kubernetes as an event, and to the Micrometer registry. 

Default and explicit messages are used for logging, and also to create the message/note for the K8s event, truncated at the 1kB limit set by the K8s servers.

Micrometer has no capacity to serve such long strings, so instead only the reason is added as a tag, allongside the pod's name and namespace.

The Kubernetes events integration tries to use the highest version events API. It prefers events.k8s.io/v1 to events.k8s.io/v1beta1, but if neither are present (or perhaps been deactivated) then it will fall back to the legacy core events API.

From the point of view of an end user running `kubectl get events`, which API as used won't be noticeable. 

### Checklist

- [X] Write tests
- [To/do - monitoring] Update documentation
- [X] Check RBAC rights for Kubernetes / OpenShift roles
- [X] Update CHANGELOG.md


cluster-operator/src/main/java/io/strimzi/operator/cluster/ClusterOperatorConfig.java
```diff
@@ -77,6 +77,9 @@
     public static final long DEFAULT_CONNECT_BUILD_TIMEOUT_MS = 300_000;
     public static final int DEFAULT_STRIMZI_OPERATIONS_THREAD_POOL_SIZE = 10;
 
+    //Hostname as injected by K8s is used to identify an "instance" of a controller/operator, needed for K8s events
+    public static final String CLUSTER_OPERATOR_HOST = "HOSTNAME";
```
scholzj, line 81: I think this is confusing ...
* Value is `HOSTNAME`
* Constant is `CLUSTER_OPERATOR_HOST`
* And the instance value is `operatorInstance`

Please decide what are you configuring. Is it a hostname? Is it some instance ID? Now it is not clear what it is. The environment variable should be maybe also prefixed with `STRIMZI_` as all other options?

cluster-operator/src/main/java/io/strimzi/operator/cluster/ClusterOperatorConfig.java
```diff
@@ -192,6 +198,8 @@ public static ClusterOperatorConfig fromMap(Map<String, String> map, KafkaVersio
         RbacScope rbacScope = parseRbacScope(map.get(STRIMZI_RBAC_SCOPE));
         Labels customResourceSelector = parseLabels(map, STRIMZI_CUSTOM_RESOURCE_SELECTOR);
         String featureGates = map.getOrDefault(STRIMZI_FEATURE_GATES, "");
+        String operatorInstance = map.get(CLUSTER_OPERATOR_HOST);
```
scholzj, line 201: Is it required? Should we check here if it is set or not?

cluster-operator/src/test/java/io/strimzi/operator/cluster/ClusterOperatorConfigTest.java
```diff
@@ -32,23 +32,24 @@
 
 public class ClusterOperatorConfigTest {
 
-    private static Map<String, String> envVars = new HashMap<>(8);
+    private static final Map<String, String> ENV_VARS = new HashMap<>(8);
```
scholzj, line 35: Wouldn't keeping this for a separate PR make the review much easier?

operator-common/pom.xml
```diff
@@ -131,6 +135,7 @@
         <dependency>
             <groupId>org.mockito</groupId>
             <artifactId>mockito-core</artifactId>
+            <scope>test</scope>
```
scholzj, line 138: Is this needed? Isn't it specified as `test` in the main pom.xml?

operator-common/pom.xml
```diff
@@ -139,6 +144,7 @@
         <dependency>
             <groupId>io.strimzi</groupId>
             <artifactId>mockkube</artifactId>
+            <scope>test</scope>
```
scholzj, line 147: Is this needed? Isn't it specified as `test` in the main pom.xml?

operator-common/src/main/java/io/strimzi/operator/PlatformFeaturesAvailability.java
```diff
@@ -56,13 +69,57 @@
...
         .onComplete(pfaPromise);
 
         return pfaPromise.future();
     }
 
+    /**
+     * Brief rundown - the various APIs and K8s versions
+     *   events.k8s.io/v1      - introduced in Kubernetes 1.19
+     *   events.k8s.io/v1beta1 - introduced in Kubernetes 1.8, deprecated in 1.19, will be removed in 1.25
+     *   core events           - Always around, but has significant performance issues, so a new event API was introduced.
```
scholzj, line 86: Why do we need the core events? It is parallel to the `events.k8s.io` so I think we should use one or another. My undertsanding is that `events.k8s.io` is _ewer and better_ - so why not just use that? What are the situation when you expect us to use core events?

operator-common/src/main/java/io/strimzi/operator/common/model/RestartReason.java
```diff
@@ -0,0 +1,42 @@
...
+ * Copyright Strimzi authors.
+ * License: Apache License 2.0 (see the file LICENSE or http://apache.org/licenses/LICENSE-2.0.html).
+ */
+
+package io.strimzi.operator.common.model;
+
+/**
+ * Defines all the (existing) reasons a pod may need to be restarted
+ */
+public enum RestartReason {
```
scholzj, line 11: Why is this in operator-common? It seems to be CO specific.

operator-common/src/main/java/io/strimzi/operator/common/operator/resource/DeploymentOperator.java
```diff
@@ -61,25 +59,28 @@ protected Integer currentScale(String namespace, String name) {
...
+     * Asynchronously restart the given pod.
      * @param reconciliation The reconciliation
      * @param namespace The namespace of the pod.
      * @param name The name of the pod.
+     * @param reasons why the pod needs to be terminated
+     * @param timeoutMs how long the restart operation has to complete
      * @return A Future which will complete once all the pods has been deleted.
      */
-    public Future<ReconcileResult<Pod>> deletePod(Reconciliation reconciliation, String namespace, String name) {
+    public Future<Void> restartSingletonPod(Reconciliation reconciliation, String namespace, String name, RestartReasons reasons, long timeoutMs) {
```
scholzj, line 80: This is wrong. Delete and restart pod are different things. Deployment also doesn't give you control over restarts the way it is used in Strimzi. So you should not pull this into it.

scholzj (commented):
I left some comments. But for me, this seems weird. The idea that you can track pod restarts is misleading and complicated to implement. Misleading because you do not control it in many cases so you cannot inform about it. Complicated to implement because it increases the complexity of the resource operators significantly. 

I think - and that was my expectation how this will be implemented - that this will be reporting about the rolling updates from Kafka Roller and nothing else. I think the current state deserves a lot more explanation (and maybe a proposal?) about what is it doing, what events / actions will be tracked etc.

It would be also good to show examples of the events this produce and of the metrics. AFAIK Prometheus does not work very well with carrying information in labels. That is why for example the resource state reporting from CO uses just integers to report ready / unready states and not labels with full explanation. I'm also not sure how will such metrics be really utilised (in other words I do not understand how useful they are). Having some Grafana integration or Prometheus Alerts as part of this PR would help to understand this and prove the usefulness of this.

operator-common/src/main/java/io/strimzi/operator/common/operator/resource/DeploymentOperator.java
LiamClarkeNZ: I renamed it to match the semantics of [where it's being called from](https://github.com/strimzi/strimzi-kafka-operator/blob/72a0260e828adad35d44e590fdbebd4b407df275/operator-common/src/main/java/io/strimzi/operator/common/operator/resource/DeploymentOperator.java#L66). So either renaming it to restart vs. delete is wrong, or calling it from `rollingUpdate` with a `Reconciliation` attached is wrong.

operator-common/src/main/java/io/strimzi/operator/PlatformFeaturesAvailability.java
LiamClarkeNZ: When someone is using Kubernetes < 1.19 and has disabled either the beta events.k8s.io API, or all beta APIs using `--feature-gates AllBeta=true` as an argument to the api-server.

operator-common/pom.xml
LiamClarkeNZ: True.

operator-common/src/main/java/io/strimzi/operator/common/operator/resource/DeploymentOperator.java
scholzj: It deletes the pod. The StatefulSet controller is what starts it again. Especially since we are removing the StatefulSets and in the future something else will start it it is important to call it what it is.

cluster-operator/src/test/java/io/strimzi/operator/cluster/ClusterOperatorConfigTest.java
LiamClarkeNZ: I guess so. I have a habit of making things `final` that can be final, but this then invoked the Checkstyle `ConstantName` rules. I'll remove the `final`.

operator-common/src/main/java/io/strimzi/operator/PlatformFeaturesAvailability.java
scholzj: Well, did you checked if Strimzi really works in that case? We do not do it for anything else and I doubt this would be the only problem then. So it is maybe not worth the effort? (Plus if you see it like this, users can disable any API IIRC)

cluster-operator/src/test/java/io/strimzi/operator/cluster/ClusterOperatorConfigTest.java
scholzj: I normally note things like this and open a small separate PR for them later.

cluster-operator/src/main/java/io/strimzi/operator/cluster/ClusterOperatorConfig.java
LiamClarkeNZ: Yep, good point, it would be better to explicitly inject it from the metadata like we do for the namespace, and prefix it as you suggest.

cluster-operator/src/main/java/io/strimzi/operator/cluster/ClusterOperatorConfig.java
LiamClarkeNZ: A controller instance is required for v1 events, so I'll ensure it's injected.

LiamClarkeNZ:
@scholzj 

Fair point. 

> It would be also good to show examples of the events this produce and of the metrics. AFAIK Prometheus does not work very well with carrying information in labels. That is why for example the resource state reporting from CO uses just integers to report ready / unready states and not labels with full explanation. I'm also not sure how will such metrics be really utilised (in other words I do not understand how useful they are). Having some Grafana integration or Prometheus Alerts as part of this PR would help to understand this and prove the usefulness of this.

>  I think - and that was my expectation how this will be implemented - that this will be reporting about the rolling updates from Kafka Roller and nothing else. 
So looking back at the original issue, you're right, and I've fallen prey to scope creep. I'm happy to dial it back.

>  The idea that you can track pod restarts is misleading and complicated to implement. Misleading because you do not control it in many cases so you cannot inform about it. 

I'm unclear on what you mean - this only tracks pod restarts initiated by the operator, to make it easier to distinguish those from pod restarts caused by crashes, or the Kube scheduler. 

> It would be also good to show examples of the events this produce and of the metrics. AFAIK Prometheus does not work very well with carrying information in labels. 

Yep, documentation is in progress and definitely includes examples of working with both events and metrics. Prometheus doesn't do super well with very high cardinality sets of labels - hence why the Prometheus metric is confined to a predefined set of label values, rather than using the existing logging messages, some of which are dynamically created via string interpolation. 

Happy to add an example Grafana dashboard using them. 

So, next steps on this draft PR:

1. Remove smaller issues flagged
2. Explicitly inject operator instance
3. Reduce scope back to the original scope of `KafkaRoller`, which removes a bunch of other changes, such as DeploymentOperator, code in `operator-common` etc. 
4. Include example Grafana dashboard incorporating metrics

operator-common/src/main/java/io/strimzi/operator/common/operator/resource/DeploymentOperator.java
LiamClarkeNZ: Okay, that's fine (although `KafkaRoller` has similar naming issues), this'll be reverted when I reduce scope back to the `KafkaRoller` only.

scholzj:
> I'm unclear on what you mean here, this only informs of pod restarts initiated by the operator, to make it easier to distinguish those from pod restarts caused by crashes, or the Kube scheduler. For example, you may see in the K8s events that a Kafka pod was restarted, but at the moment, the only indication as to why is in logging.

In many cases the operator does not restart pods explicitly. It just updates deployments and lets kubernetes do its job. So the promise to track pod restarts is false. You can really do it only in some special cases such as in the KafkaRoller, possibly ZooKeeper restarts. For other components, you will just not cover the restarts at all or even worse, cover only subset of them.

> Prometheus doesn't do super well with very high cardinality sets of labels - hence why the Prometheus metric is confined to a predefined set of label values, rather than using the existing logging messages, some of which are dynamically created via string interpolation.

The list already looked bigger that what we considered too much in the past (for example in relationship to the resource states). TBH, I personally didn't do any testing or any investigation when this becomes a problem, so I do not know what is too much and what is still ok.

> Explicitly inject operator instance

I'm not sure what exactly you mean by this. So just to clarify ... I think the current naming is confusing. If we assume that hostname is always the right value for this, we can keep calling it hostname. But in that case it might get other use in the future as well as a hostname and it might be harder to change it later if the resource instance needs to use something else. The other option is to use some env. var which is specifically for operator instance. In that case, it can be defaulting to hostname but if someone changes it it will be fine since it will be just operator instance.

Two other thoughts ... 
* If it is always hostname ... can we inject it from the startup script instead of YAML?
* Should the resource (assume we are talking about the `reportingInstance` field in the Events API?) also include namespace?

cluster-operator/src/test/java/io/strimzi/operator/cluster/ClusterOperatorConfigTest.java
LiamClarkeNZ: Will do :)

operator-common/src/main/java/io/strimzi/operator/PlatformFeaturesAvailability.java
LiamClarkeNZ: I have only checked it worked on 1.16.0 with the beta events API disabled. I'll fire up Minikube now with all beta APIs disabled, see what happens :D

LiamClarkeNZ:
> > I'm unclear on what you mean here, this only informs of pod restarts initiated by the operator, to make it easier to distinguish those from pod restarts caused by crashes, or the Kube scheduler. For example, you may see in the K8s events that a Kafka pod was restarted, but at the moment, the only indication as to why is in logging.
> 
> In many cases the operator does not restart pods explicitly. It just updates deployments and lets kubernetes do its job. So the promise to track pod restarts is false. You can really do it only in some special cases such as in the KafkaRoller, possibly ZooKeeper restarts. For other components, you will just not cover the restarts at all or even worse, cover only subset of them.

Gotcha - I named the K8s action and metric `StrimziInitiatedPodRestart/strimzi_initiated_pod_restart` to try to make it explicit, but perhaps `strimzi_inititated_pod_removal` is better?

Incidentally, the naming of things is something I'd appreciate feedback on when I push reduced scope version of this draft. :) 

> > Prometheus doesn't do super well with very high cardinality sets of labels - hence why the Prometheus metric is confined to a predefined set of label values, rather than using the existing logging messages, some of which are dynamically created via string interpolation.
> 
> The list already looked bigger that what we considered too much in the past (for example in relationship to the resource states). TBH, I personally didn't do any testing or any investigation when this becomes a problem, so I do not know what is too much and what is still ok.

True, there isn't really a hard and fast rule, as it depends on what else is going on in your Prometheus DB. The number of causes will be reduced by the reduction in scope, so will re-evaluate. 

> > Explicitly inject operator instance
> 
> I'm not sure what exactly you mean by this. So just to clarify ... I think the current naming is confusing. If we assume that hostname is always the right value for this, we can keep calling it hostname. But in that case it might get other use in the future as well as a hostname and it might be harder to change it later if the resource instance needs to use something else. The other option is to use some env. var which is specifically for operator instance. In that case, it can be defaulting to hostname but if someone changes it it will be fine since it will be just operator instance.

Definitely agree, the naming was contradictory - HOSTNAME was already there in the env vars, so I used that. 

The K8s docs are rather light on what "reportingControllerInstance" should actually refer to. It tends to be, in K8s code, "reporting-controller-" + "host". Host itself though, is quite vague. Ultimately found its derived from this...

https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apimachinery/pkg/types/nodename.go

So for the CO running in Minikube, which names the single node it's running `minikube`, it'd be "strimzi-cluster-operator-minikube". This seems to be consistent with what usages of this field I've found in the wild, and `spec.nodeName` could be pushed down in the YAML.

However, I don't think there's any hard or fast rules around it, yet. 

> Two other thoughts ...
> 
> * If it is always hostname ... can we inject it from the startup script instead of YAML?
> * Should the resource (assume we are talking about the `reportingInstance` field in the Events API?) also include namespace?

1. If we want to go with just using hostname, that would be a lot better than using YAML. 
2. Adding namespace would be good, but the field is limited to 128 characters, not sure if that could cause difficulties. 

I was thinking the operator's namespace could be added as an annotation in the event's metadata though. I'd also like to add the reconciliation id to the event, and annotations seemed the best place.

So on these fields, I'd propose:

1. When the operator publishes an event, it annotates the event with `"strimzi.io/cluster-operator-namespace" : "ns here"` 
2. It annotates it with `"strimzi.io/reconciliation-id": "id"`
3. `reportingController = "strimzi.io/strimzi-cluster-operator"`
4. `reportingInstance =  "strimzi-cluster-operator-$(hostname)"`
5. `$(hostname)` is evaluated at start-up and passed in as an argument instead of using pushed-down env vars. 

What do you think?

operator-common/src/main/java/io/strimzi/operator/PlatformFeaturesAvailability.java
LiamClarkeNZ: [So I gave it a try](https://gist.github.com/LiamClarkeNZ/9c7fda2fdcf48b7533f441ca8c2c0ee9), and Strimzi does hit an issue with one beta API, but K8s itself hit far more issues - so I'd be happy to a) document that K8s events require events.k8s.io/v1beta1 or events.k8s.io/v1 and b) log at WARN level when neither is available.

scholzj: > a) document that K8s events require events.k8s.io/v1beta1 or events.k8s.io/v1 and

TBH, we do not document this for any other API. We also haven't run into anyone with disabled beta APIs so far.

scholzj:
Hmm, I actually thought the hostname was the pod name which IMHO makes more sense than worker node name. The worker node name IMHO means little. Unlike the kubelet in some of the examples which is only one on the node, it does not identify the operator instance. `<namespace>/<pod-name>` would IMHO uniquely identify the operator as long as we stay on the same cluster.

operator-common/src/main/java/io/strimzi/operator/PlatformFeaturesAvailability.java
LiamClarkeNZ: Cool, 'll leave it as a log.warn() then.

LiamClarkeNZ:
@scholzj - `<namespace>/<pod-name>` sounds good.

scholzj:
There was not much going on on this for a long time. Let's close this Pr for the time being to clean up the PR pipeline. We can always reopen it later when we have more time for it.

PR #5375, opened 2021-07-29
Set fixed size limit for `emptyDir` volumes used for temporary files

### LINKED ISSUE #5340: Provide ability to set sizeLimit on strimzi-tmp emptyDir volume
Our K8s cluster has a policy that blocks objects if sizeLimit is not set for emptyDir volumes.  This is to prevent a single pod from accidentally consuming all of a node's resources and causing impact to other pods running on the same node.  In order to use the Strimzi operator we had to disable that policy on the cluster running the Strimzi operator since there is no way to set that parameter for the kafka and zookeeper statefulsets that are created.

It would be good to have a parameter within the kafka object that allowed defining a sizeLimit for the emptyDir volume.

Possibly in the kafka CRD under spec.kafka.statefulset.emptyDir you can have an object sizeLimit that will set the appropriate size limit for emptyDir with an appropriate default size if it isn't set.  The same could be done for zookeeper.

```
kind: Kafka
metadata:
  name: kafka-prod
spec:
  kafka:
    statefulset:
      emptyDir:
        sizeLimit: 1Gi
  zookeeper:
    statefulset:
      emptyDir:
        sizeLimit: 1Gi
```

scholzj:
### Type of change

- Enhancement / new feature

### Description

**This is a draft used for testing!**

This PR tries to address and fix #5340:
* It removes the `--context` volume / directory / option from Kaniko build. This actually does not seem to be needed (since we download everything from the internet and don't use any local context files).
* It adds fixed size limit for all `emptyDir` volumes used by Strimzi for temporary files
    * `1Mi` for the `/tmp` directories
    * `1Ki` for the Kafka cluster `ready-files` volume
    * `100Ki` for the Kafka cluster init container data
    * `10Ki` for the Kafka Connect cluster init container data

### Checklist

- [ ] Write tests
- [x] Update CHANGELOG.md


cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaConnectBuild.java
```diff
@@ -322,7 +320,7 @@ public Pod generateBuilderPod(boolean isOpenShift, ImagePullPolicy imagePullPoli
 
         List<String> args = additionalKanikoOptions != null ? new ArrayList<>(4 + additionalKanikoOptions.size()) : new ArrayList<>(4);
         args.add("--dockerfile=/dockerfile/Dockerfile");
-        args.add("--context=dir://workspace");
+        //args.add("--context=dir://workspace");
```
sknot-rh, line 323: Do you plan to remove it or leaving it as a comment does have any intention?

sknot-rh (approved):

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaConnectBuild.java
scholzj: Sorry, my mistake - I deleted the line. Thanks for pointing it out.

Issue #5064, opened 2021-06-01
[CRD Generator] Validate that type fields have NON_NULL include policy


scholzj:
When using the `type` as discriminators for different sub-classes, it is important to make sure the `NON_NULL` include policy (`@JsonInclude(JsonInclude.Include.NON_NULL)`). Using other policies such as for example `NON_DEFAULT` will cause the `type` field to be omitted and generate invalid resource. That can cause issues such as the one described in #5054. Ideally, we should introduce. validation check for this in the CrdGenertor classes to make sure all classes are properly set.

ShubhamRwt:
@scholzj Hello, Is this issue still open or is it resolved?

tombentley:
@ShubhamRwt I suspect so, though the need for it isn't so pressing now that we're done the CRD conversion.

ShubhamRwt:
Okay got it, Thanks @tombentley

PR #5387, opened 2021-08-04
Add annotation check to CrdGenerator

sknot-rh:

### Type of change
- Enhancement / new feature

### Description
Fixes https://github.com/strimzi/strimzi-kafka-operator/issues/5064
Per issue description: The `type` property should be annotated by `@JsonInclude(JsonInclude.Include.NON_NULL)`. That will not omit this property while generating the CRD.

### Checklist

- [x] Write tests
- [ ] Update documentation
- [ ] Update CHANGELOG.md


scholzj (commented):
I think this works to some extent. But it is checking it purely on the name of the field being `type`. That is why you needed to add it to the `api` classes.

I think a better solution would be to detect the fields through the `JsonTypeInfo` / `JsonSubTypes` annotation instead of relying on a field name. That way you would also not need to add new CRD for the test because the ExampleCrd in tests already has the `discrim` field.

Not sure how more complicated the code will be.

scholzj (approved):
One nit. LGTM otherwise.

api/src/main/java/io/strimzi/api/kafka/model/AclRule.java
```diff
@@ -51,7 +51,6 @@ public AclRule(AclRuleType type, AclRuleResource resource, String host, AclOpera
             "ACL rules with type `allow` are used to allow user to execute the specified operations. " +
             "Default value is `allow`.")
     @DefaultValue("allow")
-    @JsonInclude(JsonInclude.Include.NON_DEFAULT)
```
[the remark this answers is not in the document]
sknot-rh: Ah, I thought is it an abstract class.

crd-generator/src/main/java/io/strimzi/crdgenerator/CrdGenerator.java
```diff
@@ -719,6 +731,11 @@ private void checkClass(Class<?> crdClass) {
         checkClassOverrides(crdClass, "equals", Object.class);
     }
 
+    private boolean hasAnnotatedType(Method method) {
```
tombentley, line 734: Surely something like `isAnnotatedWithIncludeNonNull` would be a better name.

crd-generator/src/main/java/io/strimzi/crdgenerator/CrdGenerator.java
```diff
@@ -707,6 +707,18 @@ private void checkClass(Class<?> crdClass) {
...
+                    String methodName = "get" + typePropertyName.substring(0, 1).toUpperCase(Locale.ENGLISH) + typePropertyName.substring(1).toLowerCase(Locale.ENGLISH);
+                    @SuppressWarnings("unchecked")
+                    Method method = c.getMethod(methodName);
+
+                    if (!hasAnnotatedType(method)) {
+                        err(c.getCanonicalName() + "#" + methodName + " is not annotated with @JsonInclude(JsonInclude.Include.NON_NULL)");
+                    }
+                } catch (NoSuchMethodException e) {
+                    err(e.getMessage());
+                }
```
tombentley, line 721: This could be factored into a method called something like `checkDiscriminatorIsIncluded`

tombentley (approved):
A couple of nits, LGTM otherwise.

scholzj:
@sknot-rh Does this now address Tom's comments? Can it be merged?

PR #5392, opened 2021-08-04
Update tags and make client examples images configurable

echernous:
### Type of change
- Enhancement

### Description

Updated missing tags for system tests, and made client examples images configurable

### Checklist


echernous:
Hi @Frawless, I was going through these tests in our midstream and thought it would be good to have these changes here too

Frawless (approved):
I think it makes sense to have clients' images configurable. JFYI we will switch from `clients-examples` to https://github.com/strimzi/test-clients in upcoming releases, we just need to implement some more functionality which we want.

LGTM.

echernous:
Hey @Frawless and @scholzj , I checked the failed tests and tried to run them on an osd cluster, it seems working - the correct client images are getting pulled and tests are passing. I'm not sure what happened with the build-kafka image. Please could we try to run it again? Thanks!

Frawless:
@echernous Maybe the clients image is configured differently for specific tests. I found this in for `FutureGateST`:

```Failed to pull image "10.99.250.171:80/strimzi-examples/java-kafka-producer:latest": rpc error: code = Unknown desc = Error response from daemon: manifest for 10.99.250.171:80/strimzi-examples/java-kafka-producer:latest not found: manifest unknown: manifest unknown```

I restarted the pipeline so we will see, maybe it was just some env issue.  If it will fail, I will check on Monday what's going on.

scholzj:
I think it will not work. The `STRIMZI_REGISTRY` is overriden in the STs to the local registry which contains the current build images. But the example clients (or in the future test clients) are not part of this build. So they will not exist in the local registry. So IMHO you need to have some environment variable `STRIMZI_TEST_CLIENTS_REGISTRY etc., which would be independent on the ones used for the operator / operand images.

echernous:
@scholzj , @Frawless thank you! got it, let change STRIMZI_REGISTRY to STRIMZI_REGISTRY_DEFAULT

see-quick (approved):

PR #5397, opened 2021-08-04
[DOC] Clarify that `*` can be used as a wildcard in ACL rules

scholzj:
### Type of change

- Enhancement / new feature
- Documentation

### Description

In the ACL rules, the `name` field can be set to `*` to indicate that the rule applies to all resources. This was missing in the docs and it often comes up as a question. This PR adds it to the docs to have it covered.

### DIFF
--- a/documentation/api/io.strimzi.api.kafka.model.AclRule.adoc
+++ b/documentation/api/io.strimzi.api.kafka.model.AclRule.adoc
@@ -53,6 +53,19 @@ A name is specified as a `literal` or a `prefix` using the `patternType` propert
 * Literal names are taken exactly as they are specified in the `name` field.
 * Prefix names use the value from the `name` as a prefix, and will apply the rule to all resources with names starting with the value.
 
+When `patternType` is set as `literal`, you can set the name to `*` to indicate that the rule applies to all resources.
+
+.Example ACL rule which allows the user to read messages from all topics
+[source,yaml,subs="attributes+"]
+----
+    acls:
+      - resource:
+          type: topic
+          name: "*"
+          patternType: literal
+        operation: Read
+----
+
 [id='property-acl-type-{context}']
 === `type`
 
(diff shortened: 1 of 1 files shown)


ppatierno (approved):

documentation/api/io.strimzi.api.kafka.model.AclRule.adoc
```diff
@@ -53,6 +53,19 @@ A name is specified as a `literal` or a `prefix` using the `patternType` propert
 * Literal names are taken exactly as they are specified in the `name` field.
 * Prefix names use the value from the `name` as a prefix, and will apply the rule to all resources with names starting with the value.
 
+When `patternType` is set as `literal`, you can set the name to `*` to indicate that the rule applies to all resources.
+
+.Example ACL rule which allows the user to read messages from all topics
```
laidan6000, line 58:
```suggestion
.Example ACL rule that allows the user to read messages from all topics
```

documentation/api/io.strimzi.api.kafka.model.AclRule.adoc
```diff
@@ -53,6 +53,19 @@ A name is specified as a `literal` or a `prefix` using the `patternType` propert
 * Literal names are taken exactly as they are specified in the `name` field.
 * Prefix names use the value from the `name` as a prefix, and will apply the rule to all resources with names starting with the value.
```
laidan6000, line 54:
```suggestion
* Prefix names use the `name` value as a prefix and then apply the rule to all resources with names starting with that value.
```

laidan6000 (changes requested):
Looks like a useful addition. I made a couple of suggestions.

scholzj:
@laidan6000 Thanks, should be fixed now.

laidan6000 (approved):

PR #5401, opened 2021-08-06
Fix the offset and transaction state topic configurations in Quickstart book

### LINKED ISSUE #5400: Docs: Strimzi Quick Start guide (0.24.0) incorrect Kafka CRD example
Kafka CRD's example deployment described in https://strimzi.io/docs/operators/latest/quickstart.html (which is based on
https://github.com/strimzi/strimzi.github.io/blob/main/docs/operators/0.24.0/quickstart-book.html#L432)
is incorrect. It causes `InvalidResourceException` in Kafka CRD with message: `Kafka configuration option 'offsets.topic.replication.factor' should be set to 1 because 'spec.kafka.replicas' is 1`.

To make it work with a defined config, it is required to have 3 replicas.

scholzj:
### Type of change

- Bugfix
- Documentation

### Description

The Quickstart book currently uses a Kafka cluster with one Zoo and one Kafka node. But it uses replication factor 3 for offsets and transaction states. That does not work with single Kafka node. This PR changes the settings to make them work.

This should fix #5400.

### Checklist

- [x] Update documentation

### DIFF
--- a/documentation/modules/quickstart/proc-creating-kafka-cluster.adoc
+++ b/documentation/modules/quickstart/proc-creating-kafka-cluster.adoc
@@ -57,9 +57,9 @@ spec:
         size: 100Gi
         deleteClaim: false
     config:
-      offsets.topic.replication.factor: 3
-      transaction.state.log.replication.factor: 3
-      transaction.state.log.min.isr: 2
+      offsets.topic.replication.factor: 1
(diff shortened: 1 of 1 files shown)


sknot-rh (approved):

samuel-hawker (approved):
LGTM

Issue #4179, opened 2021-01-05
Kafka Connect Build: Add support for Maven coordinates artifacts


scholzj:
We should add support for Maven coordinates to Kafka Connect build. The idea is the user would specify Maven artifact by using group ID, artifact ID and version. And Strimzi would pull it including all dependencies. This would probably also need to support configuration of different Maven repositories etc.

PR #5407, opened 2021-08-09
Support maven-coords based connector build

sknot-rh:

### Type of change
- Enhancement / new feature

### Description
Fixes https://github.com/strimzi/strimzi-kafka-operator/issues/4179
Add support for maven-coords based connector build.

`jar`, `zip` and `tgz` types of build do have a common required parameter -`url`. That does not apply for `maven`, where an optional repository can be and required GAV has to be specified.

### Checklist

- [x] Write tests
- [x] Update documentation
- [x] Update CHANGELOG.md


docker-images/kafka/Dockerfile
```diff
@@ -6,7 +6,7 @@ ARG THIRD_PARTY_LIBS
 ARG strimzi_version
 ARG TARGETPLATFORM
 
-RUN microdnf install gettext nmap-ncat stunnel net-tools unzip hostname findutils tar \
+RUN microdnf install gettext nmap-ncat stunnel net-tools unzip hostname findutils tar maven \
```
scholzj, line 9: I don't think we want this for security reason. As discussed before offline - I think this should use multistage build or some other way to download it without having Maven in all the container images.

tombentley: @sknot-rh have you seen https://github.com/shrinkwrap/resolver?

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaConnectDockerfile.java
```diff
@@ -144,6 +148,24 @@ private void addPlugin(PrintWriter writer, Plugin plugin)    {
         }
     }
 
+    private void validateUrlPresence(DownloadableArtifact art) {
+        if (art.getUrl() == null) {
+            throw new InvalidConfigurationException(art.getType() + " artifact is missing an URL.");
+        }
+    }
+
+    private void validateGavPresence(MavenArtifact art) {
```
tombentley, line 157: Wouldn't it be simpler to have a `validate` method on `Artifact` which was implemented by `DownloadableArtifact` to check the URL and by `MavenArtifact` to check the GAV? Or is `InvalidConfigurationException` not in `api`?

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaConnectDockerfile.java
```diff
@@ -47,16 +49,18 @@
      */
     public KafkaConnectDockerfile(String fromImage, Build connectBuild) {
         StringWriter stringWriter = new StringWriter();
+        StringWriter prefixStringWriter = new StringWriter();
+        PrintWriter prefixWriter = new PrintWriter(prefixStringWriter);
```
scholzj, line 53: Why is this needed? I don't understand this

sknot-rh: From [this article](https://docs.docker.com/develop/develop-images/multistage-build/#use-multi-stage-builds) that the parts have to be done in a specific order. I did not try the unordered parts. Do you think it works the same way?

scholzj: Ah, ok now I understand it better ... but:
* The naming is unintuitive => I do not think anyone would understand that the prefix means the initial stage here.
* I wonder if we want to have the _one stage per artifact_ approach there. Wouldn't it be better to have a single writer with something like this and generate only the build stage + the regular container instead of having many stages?
  ```
        printHeader(writer); // Print initial comment
        connectorPluginsPreStage(writer, connectBuild.getPlugins());
        from(writer, fromImage); // Create FROM statement
        user(writer, ROOT_USER); // Switch to root user to be able to add plugins
        proxy(writer); // Configures proxy environment variables
        connectorPlugins(writer, connectBuild.getPlugins());
  ```

I also think the tool linked by Tom would be worth investigating. Keeping it inside the operator and generating just a regular Dockerfile might make it easier to test it (given the different build on Kube and OCP, the multistage build needs to be tested there twice).

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaConnectDockerfile.java
```diff
@@ -284,6 +313,61 @@ private void addZipArtifact(PrintWriter writer, String connectorPath, ZipArtifac
...
+                mvn.getArtifact(),
+                mvn.getVersion(),
+                repo,
+                mvn.getGroup().replace(".", "/"), //org.apache.camel is translated as org/apache/camel in the URL
+                mvn.getArtifact().replace(".", "/"),
+                mvn.getVersion(),
+                mvn.getArtifact(),
+                mvn.getVersion());
+
+        prefixWriter.println("FROM maven:3.8.1-openjdk-17-slim AS download" + artifactHash);
```
scholzj, line 347: * I'm not sure we should use Java 17 for this. Can't there be different artifacts for Java 17 in some cases? Does that impact what is downloaded?
* Since we use UBI8 as base image, I wonder if we should use https://catalog.redhat.com/software/containers/ubi8/openjdk-11/5dd6a4b45a13461646f677f4?container-tabs=overview instead
* I wonder if we should repackage this image similarly how we do it for Kaniko.

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaConnectDockerfile.java
```diff
@@ -61,6 +66,32 @@ public KafkaConnectDockerfile(String fromImage, Build connectBuild) {
...
+                        repo,
+                        ((MavenArtifact) mvn).getGroup().replace(".", "/"), //org.apache.camel is translated as org/apache/camel in the URL
+                        ((MavenArtifact) mvn).getArtifact().replace(".", "/"),
+                        ((MavenArtifact) mvn).getVersion(),
+                        ((MavenArtifact) mvn).getArtifact(),
+                        ((MavenArtifact) mvn).getVersion());
+
+                writer.println("FROM " + UBI_MAVEN_IMAGE + " AS download" + artifactHash);
+                writer.println("RUN " + downloadPomCmd + " \\");
+                writer.println("      && mvn dependency:copy-dependencies -DoutputDirectory=/tmp/artifacts/");
```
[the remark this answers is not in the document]
sknot-rh: It seems to be dowloaded each build -> 
```
Downloaded from central: https://repo1.maven.org/maven2/org/apache/maven/plugins/maven-dependency-plugin/3.1.1/maven-dependency-plugin-3.1.1.jar (167 kB at 6.2 MB/s)
```

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaConnectDockerfile.java
sknot-rh: `ICE` is not in the `api`

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaConnectDockerfile.java
```diff
@@ -61,6 +72,42 @@ public KafkaConnectDockerfile(String fromImage, Build connectBuild) {
...
+                        ((MavenArtifact) mvn).getGroup().replace(".", "/"), //org.apache.camel is translated as org/apache/camel in the URL
+                        ((MavenArtifact) mvn).getArtifact().replace(".", "/"),
+                        ((MavenArtifact) mvn).getVersion(),
+                        ((MavenArtifact) mvn).getArtifact(),
+                        ((MavenArtifact) mvn).getVersion());
+
+                writer.println("FROM " + mavenImage + " AS download" + artifactHash);
+                writer.println("RUN " + downloadPomCmd + " \\");
+                writer.println("      && mvn dependency:copy-dependencies -DoutputDirectory=/tmp/artifacts/");
+                writer.println();
```
scholzj, line 99: Can we try to do this as multiple commands in a single stage? I.e. instead of stage per artifact use one stage and output directory per artifact?

scholzj (commented):
One nit. I also wonder if `maven-builder` would be better name for the image than `maven-image`.

CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 ## 0.26.0
 
 * Update the Open Policy Agent Authorizer to version [1.1.0](https://github.com/Bisnode/opa-kafka-plugin/releases/tag/v1.1.0)
+* Add support for building connector images from the maven coordinates
```
PaulRMellor, line 6:
```suggestion
* Add support for building connector images from the maven coordinates (#4179)
```

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaConnectDockerfile.java
```diff
@@ -144,6 +196,24 @@ private void addPlugin(PrintWriter writer, Plugin plugin)    {
         }
     }
 
+    private void validateUrlPresence(DownloadableArtifact art) {
+        if (art.getUrl() == null) {
+            throw new InvalidConfigurationException(art.getType() + " artifact is missing an URL.");
```
PaulRMellor, line 201:
```suggestion
            throw new InvalidConfigurationException(art.getType() + " artifact is missing a URL.");
```

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaConnectDockerfile.java
```diff
@@ -144,6 +196,24 @@ private void addPlugin(PrintWriter writer, Plugin plugin)    {
...
+        }
+    }
+
+    private void validateGavPresence(MavenArtifact art) {
+        if (art.getGroup() == null) {
+            throw new InvalidConfigurationException(art.getType() + " artifact is missing Group.");
```
PaulRMellor, line 207: _group ID_?

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaConnectDockerfile.java
```diff
@@ -144,6 +196,24 @@ private void addPlugin(PrintWriter writer, Plugin plugin)    {
...
+        }
+        if (art.getArtifact() == null) {
+            throw new InvalidConfigurationException(art.getType() + " artifact is missing Artifact.");
```
PaulRMellor, line 210: `artifactID`

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaConnectDockerfile.java
```diff
@@ -144,6 +196,24 @@ private void addPlugin(PrintWriter writer, Plugin plugin)    {
...
+        }
+        if (art.getVersion() == null) {
+            throw new InvalidConfigurationException(art.getType() + " artifact is missing Version.");
```
PaulRMellor, line 213:
```suggestion
            throw new InvalidConfigurationException(art.getType() + " artifact is missing a version number.");
```

cluster-operator/src/test/java/io/strimzi/operator/cluster/model/KafkaConnectDockerfileTest.java
```diff
@@ -396,4 +401,134 @@ public void testDockerfileWithComments()   {
...
+        Build connectBuildTgz = new BuildBuilder()
+                .withPlugins(new PluginBuilder()
+                        .withName("my-connector-plugin")
+                        .withArtifacts(singletonList(tgzArtifact))
+                        .build())
+                .build();
+
+        Throwable e = assertThrows(InvalidConfigurationException.class, () -> new KafkaConnectDockerfile("myImage:latest", connectBuildTgz));
+
+        assertThat(e.getMessage(), is("tgz artifact is missing an URL."));
```
PaulRMellor, line 424:
```suggestion
        assertThat(e.getMessage(), is(".tgz artifact is missing a URL."));
```

cluster-operator/src/test/java/io/strimzi/operator/cluster/model/KafkaConnectDockerfileTest.java
```diff
@@ -396,4 +401,134 @@ public void testDockerfileWithComments()   {
...
+        Build connectBuildZip = new BuildBuilder()
+                .withPlugins(new PluginBuilder()
+                        .withName("my-connector-plugin")
+                        .withArtifacts(singletonList(zipArtifact))
+                        .build())
+                .build();
+
+        e = assertThrows(InvalidConfigurationException.class, () -> new KafkaConnectDockerfile("myImage:latest", connectBuildZip));
+
+        assertThat(e.getMessage(), is("zip artifact is missing an URL."));
```
PaulRMellor, line 436:
```suggestion
        assertThat(e.getMessage(), is(".zip artifact is missing a URL."));
```

cluster-operator/src/test/java/io/strimzi/operator/cluster/model/KafkaConnectDockerfileTest.java
```diff
@@ -396,4 +401,134 @@ public void testDockerfileWithComments()   {
...
+        Build connectBuildJar = new BuildBuilder()
+                .withPlugins(new PluginBuilder()
+                        .withName("my-connector-plugin")
+                        .withArtifacts(singletonList(jarArtifact))
+                        .build())
+                .build();
+
+        e = assertThrows(InvalidConfigurationException.class, () -> new KafkaConnectDockerfile("myImage:latest", connectBuildJar));
+
+        assertThat(e.getMessage(), is("jar artifact is missing an URL."));
```
PaulRMellor, line 448:
```suggestion
        assertThat(e.getMessage(), is(".jar artifact is missing a URL."));
```

packaging/helm-charts/helm3/strimzi-kafka-operator/crds/041-Crd-kafkaconnect.yaml
```diff
@@ -1727,27 +1727,39 @@ spec:
                             items:
                               type: object
                               properties:
+                                artifact:
+                                  type: string
+                                  description: Name of the artifact. Applicable for 'maven' type of artifact only.
```
PaulRMellor, line 1732:
```suggestion
                                  description: Maven artifact id. Applicable to the 'maven' artifact type only.
```

packaging/helm-charts/helm3/strimzi-kafka-operator/crds/041-Crd-kafkaconnect.yaml
```diff
@@ -1727,27 +1727,39 @@ spec:
...
                                 fileName:
                                   type: string
                                   description: Name under which the artifact will be stored.
+                                group:
+                                  type: string
+                                  description: Group of the artifact. Applicable for 'maven' type of artifact only.
```
PaulRMellor, line 1738:
```suggestion
                                  description: Maven project group id. Applicable to the 'maven' artifact type only.
```

packaging/helm-charts/helm3/strimzi-kafka-operator/crds/041-Crd-kafkaconnect.yaml
```diff
@@ -1727,27 +1727,39 @@ spec:
...
+                                repository:
+                                  type: string
+                                  description: The repository to download the artifact from. Applicable for 'maven' type of artifact only.
```
PaulRMellor, line 1741:
```suggestion
                                  description: Maven repository to download the artifact from. Applicable to the 'maven' artifact type only.
```

packaging/helm-charts/helm3/strimzi-kafka-operator/crds/041-Crd-kafkaconnect.yaml
```diff
@@ -1727,27 +1727,39 @@ spec:
...
                                 sha512sum:
                                   type: string
-                                  description: SHA512 checksum of the artifact. Optional. If specified, the checksum will be verified while building the new container. If not specified, the downloaded artifact will not be verified.
+                                  description: SHA512 checksum of the artifact. Optional. If specified, the checksum will be verified while building the new container. If not specified, the downloaded artifact will not be verified. Not applicable for 'maven' type of artifact.
```
PaulRMellor, line 1744:
```suggestion
                                  description: SHA512 checksum of the artifact. Optional. If specified, the checksum will be verified while building the new container. If not specified, the downloaded artifact will not be verified. Not applicable to the `maven` artifact type.
```

packaging/helm-charts/helm3/strimzi-kafka-operator/crds/041-Crd-kafkaconnect.yaml
```diff
@@ -1727,27 +1727,39 @@ spec:
...
+                                  description: Artifact type. Currently, the supported artifact types are `tgz`, `jar`, `zip` and `maven`.
                                 url:
                                   type: string
                                   pattern: ^(https?|ftp)://[-a-zA-Z0-9+&@#/%?=~_|!:,.;]*[-a-zA-Z0-9+&@#/%=~_|]
-                                  description: URL of the artifact which will be downloaded. Strimzi does not do any security scanning of the downloaded artifacts. For security reasons, you should first verify the artifacts manually and configure the checksum verification to make sure the same artifact is used in the automated build. Required.
+                                  description: URL of the artifact which will be downloaded. Strimzi does not do any security scanning of the downloaded artifacts. For security reasons, you should first verify the artifacts manually and configure the checksum verification to make sure the same artifact is used in the automated build. Required. Not applicable for 'maven' type of artifact.
```
PaulRMellor, line 1757:
```suggestion
                                  description: URL of the artifact which will be downloaded. Strimzi does not do any security scanning of the downloaded artifacts. For security reasons, you should first verify the artifacts manually and configure the checksum verification to make sure the same artifact is used in the automated build. Required. Not applicable to the `maven` artifact type.
```

packaging/helm-charts/helm3/strimzi-kafka-operator/crds/041-Crd-kafkaconnect.yaml
```diff
@@ -1727,27 +1727,39 @@ spec:
...
+                                version:
+                                  type: string
+                                  description: Version of the artifact. Applicable for 'maven' type of artifact only.
```
PaulRMellor, line 1760:
```suggestion
                                  description: Version of the artifact. Applicable to the `maven` artifact type only.
```

api/src/main/java/io/strimzi/api/kafka/model/connect/build/MavenArtifact.java
```diff
@@ -0,0 +1,77 @@
...
+    @Description("Group of the artifact. Applicable for 'maven' type of artifact only.")
+    public String getGroup() {
+        return group;
+    }
+
+    public void setGroup(String group) {
+        this.group = group;
+    }
+
+    @Description("Name of the artifact. Applicable for 'maven' type of artifact only.")
```
PaulRMellor, line 50:
```suggestion
    @Description("Maven artifact id. Applicable to the `maven` artifact type only.")
```

api/src/main/java/io/strimzi/api/kafka/model/connect/build/MavenArtifact.java
```diff
@@ -0,0 +1,77 @@
...
+    private String repository;
+
+    @Description("Must be `" + TYPE_MVN + "`")
+    @JsonInclude(JsonInclude.Include.NON_NULL)
+    @Override
+    public String getType() {
+        return TYPE_MVN;
+    }
+
+    @Description("Group of the artifact. Applicable for 'maven' type of artifact only.")
```
PaulRMellor, line 41:
```suggestion
    @Description("Maven group id. Applicable to the `maven` artifact type only.")
```

packaging/helm-charts/helm3/strimzi-kafka-operator/crds/041-Crd-kafkaconnect.yaml
sknot-rh: We are specifying artifact, not maven version. I think this is ambiguous.

api/src/main/java/io/strimzi/api/kafka/model/connect/build/MavenArtifact.java
```diff
@@ -0,0 +1,77 @@
...
+    @Description("Version of the artifact. Applicable for 'maven' type of artifact only.")
+    public String getVersion() {
+        return version;
+    }
+
+    public void setVersion(String version) {
+        this.version = version;
+    }
+
+    @Description("The repository to download the artifact from. Applicable for 'maven' type of artifact only.")
```
PaulRMellor, line 68:
```suggestion
    @Description("Maven repository to download the artifact from. Applicable to the `maven` artifact type only.")
```

api/src/main/java/io/strimzi/api/kafka/model/connect/build/DownloadableArtifact.java
```diff
@@ -30,8 +29,7 @@
     @Description("URL of the artifact which will be downloaded. " +
             "Strimzi does not do any security scanning of the downloaded artifacts. " +
             "For security reasons, you should first verify the artifacts manually and configure the checksum verification to make sure the same artifact is used in the automated build. " +
-            "Required.")
-    @JsonProperty(required = true)
+            "Required. Not applicable for 'maven' type of artifact.")
```
PaulRMellor, line 32:
```suggestion
            "Required. Not applicable to the `maven` artifact type.")
```

api/src/main/java/io/strimzi/api/kafka/model/connect/build/MavenArtifact.java
```diff
@@ -0,0 +1,77 @@
...
+    @Description("Name of the artifact. Applicable for 'maven' type of artifact only.")
+    public String getArtifact() {
+        return artifact;
+    }
+
+    public void setArtifact(String artifact) {
+        this.artifact = artifact;
+    }
+
+    @Description("Version of the artifact. Applicable for 'maven' type of artifact only.")
```
PaulRMellor, line 59:
```suggestion
    @Description("Maven version number. Applicable to the `maven` artifact type only.")
```

api/src/main/java/io/strimzi/api/kafka/model/connect/build/DownloadableArtifact.java
```diff
@@ -44,7 +42,7 @@ public void setUrl(String url) {
     @Description("SHA512 checksum of the artifact. " +
             "Optional. " +
             "If specified, the checksum will be verified while building the new container. " +
-            "If not specified, the downloaded artifact will not be verified.")
+            "If not specified, the downloaded artifact will not be verified. Not applicable for 'maven' type of artifact.")
```
PaulRMellor, line 45:
```suggestion
            "If not specified, the downloaded artifact will not be verified. Not applicable to the `maven` artifact type.")
```

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaConnectDockerfile.java
PaulRMellor: Is this for any artifact type?

sknot-rh: Any other.

documentation/modules/appendix_crds.adoc
```diff
@@ -2024,14 +2024,37 @@ Used in: xref:type-Plugin-{context}[`Plugin`]
 [options="header"]
 |====
 |Property          |Description
-|url        1.2+<.<a|URL of the artifact which will be downloaded. Strimzi does not do any security scanning of the downloaded artifacts. For security reasons, you should first verify the artifacts manually and configure the checksum verification to make sure the same artifact is used in the automated build. Required.
+|url        1.2+<.<a|URL of the artifact which will be downloaded. Strimzi does not do any security scanning of the downloaded artifacts. For security reasons, you should first verify the artifacts manually and configure the checksum verification to make sure the same artifact is used in the automated build. Required. Not applicable for 'maven' type of artifact.
```
PaulRMellor: I don't suppose we can avoid the _Not applicable..._ bit appearing in the sections where it is not relevant?

documentation/modules/overview/con-configuration-points-connect.adoc
```diff
@@ -48,7 +48,8 @@ Kafka provides two built-in connectors:
 * `FileStreamSinkConnector` streams data from Kafka to an external system, reading messages from a Kafka topic and creating a line for each in an output file.
 --
 
-You can add other connectors using connector plugins, which are a set of JAR files or TGZ archives that define the implementation required to connect to certain types of external system.
+You can add other connectors using connector plugins, which are a set of JAR files, TGZ or ZIP archives that define the implementation required to connect to certain types of external system.
+You can also specify a Maven coordinates to fetch plugin from the Maven repository.
```
PaulRMellor, line 52:
```suggestion
You can also specify Maven coordinates to fetch plugins from a Maven repository.
```

documentation/modules/appendix_crds.adoc
sknot-rh: It should appear only in the sections describing jar, zip, tgz and maven type of Connect build.

documentation/modules/overview/con-configuration-points-connect.adoc
PaulRMellor: Nice to have the Overview updated :+1:

PaulRMellor (approved):
Look good. Few suggestions

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaConnectDockerfile.java
```diff
@@ -39,17 +45,23 @@
 
     private final String dockerfile;
 
+    private final Pattern special = Pattern.compile("[!@#$%&*()+=\\\\|<>?{};\\[\\]~]");
+    private static final String DEFAULT_MAVEN_IMAGE = "registry.access.redhat.com/ubi8/openjdk-11:1.3-18";
```
scholzj, line 49: This should be the Strimzi image, or?

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaConnectDockerfile.java
```diff
@@ -284,6 +357,47 @@ private void addZipArtifact(PrintWriter writer, String connectorPath, ZipArtifac
...
+     * @param connectorPath     Path where the connector to which this artifact belongs should be downloaded
+     * @param mvn               The maven artifact
+     */
+    private void addMavenArtifact(PrintWriter writer, String connectorPath, MavenArtifact mvn) {
+        validateGavPresence(mvn);
+        String repo = mvn.getRepository() == null ? MavenArtifact.DEFAULT_REPOSITORY : maybePatchRepository(mvn.getRepository());
+        String artifactHash = Util.sha1Prefix(mvn.getGroup() + mvn.getArtifact() + mvn.getVersion());
+        String artifactDir = connectorPath + "/" + artifactHash;
+
+        String downloadJarCmd = String.format("curl -L --create-dirs --output %s/%s-%s.jar %s%s/%s/%s/%s-%s.jar",
```
scholzj, line 373: Why do we have to download the JAR? Isn't it downloaded in the pre-step? If we have to do it, wouldn't it be better to do it in the pre-stage as well? It would save us step / layer in the final image.

Makefile
```diff
@@ -55,6 +55,7 @@ release_version:
 	$(FIND) ./packaging/install -name '*.yaml' -type f -exec $(SED) -i '/[0-9.]\+=quay.io\/strimzi\/kafka[a-zA-Z0-9_.-]\?\+:[a-zA-Z0-9_.-]\+-kafka-[0-9.]\+"\?/s/:[a-zA-Z0-9_.-]\+-kafka-\([0-9.]\+\)/:$(RELEASE_VERSION)-kafka-\1/g' {} \;
 	$(FIND) ./packaging/install -name '*.yaml' -type f -exec $(SED) -i '/value: "\?quay.io\/strimzi\/jmxtrans:[a-zA-Z0-9_.-]\+"\?/s/quay.io\/strimzi\/jmxtrans:[a-zA-Z0-9_.-]\+/quay.io\/strimzi\/jmxtrans:$(RELEASE_VERSION)/g' {} \;
 	$(FIND) ./packaging/install -name '*.yaml' -type f -exec $(SED) -i '/value: "\?quay.io\/strimzi\/kaniko-executor:[a-zA-Z0-9_.-]\+"\?/s/quay.io\/strimzi\/kaniko-executor:[a-zA-Z0-9_.-]\+/quay.io\/strimzi\/kaniko-executor:$(RELEASE_VERSION)/g' {} \;
+	$(FIND) ./packaging/install -name '*.yaml' -type f -exec $(SED) -i '/value: "\?quay.io\/strimzi\/maven-builder:[a-zA-Z0-9_.-]\+"\?/s/quay.io\/strimzi\/maven-image:[a-zA-Z0-9_.-]\+/quay.io\/strimzi\/maven-image:$(RELEASE_VERSION)/g' {} \;
```
scholzj, line 58: Does this need to be replaced everywhere?

```suggestion
	$(FIND) ./packaging/install -name '*.yaml' -type f -exec $(SED) -i '/value: "\?quay.io\/strimzi\/maven-builder:[a-zA-Z0-9_.-]\+"\?/s/quay.io\/strimzi\/maven-builder:[a-zA-Z0-9_.-]\+/quay.io\/strimzi\/maven-builder:$(RELEASE_VERSION)/g' {} \;
```

api/src/main/java/io/strimzi/api/kafka/model/connect/build/DownloadableArtifact.java
```diff
@@ -30,8 +29,8 @@
     @Description("URL of the artifact which will be downloaded. " +
             "Strimzi does not do any security scanning of the downloaded artifacts. " +
             "For security reasons, you should first verify the artifacts manually and configure the checksum verification to make sure the same artifact is used in the automated build. " +
-            "Required.")
-    @JsonProperty(required = true)
+            "Required." +
+            "Not applicable to the `maven` artifact type.")
```
scholzj, line 33: Missing spaces? + is it really `Required` if it is not applicable to ...? Maybe you should try to reweord it something like this?

```suggestion
            "Required for `jar`, `zip`, and `tgz` artifacts. " +
            "Not applicable to the `maven` artifact type.")
```

api/src/main/java/io/strimzi/api/kafka/model/connect/build/DownloadableArtifact.java
```diff
@@ -44,7 +43,8 @@ public void setUrl(String url) {
     @Description("SHA512 checksum of the artifact. " +
             "Optional. " +
             "If specified, the checksum will be verified while building the new container. " +
-            "If not specified, the downloaded artifact will not be verified.")
+            "If not specified, the downloaded artifact will not be verified. Not applicable for 'maven' type of artifact." +
+            "Not applicable to the `maven` artifact type.")
```
scholzj, line 47: Missing space? + you have `Not applicable for 'maven' type of artifact.` twice.

scholzj (commented):
Some nits. You also have to rebase the CHANGELOG :-/. Sorry.

CHANGELOG.md
scholzj: Should `maven` be `Maven` with capital `M`?

api/src/main/java/io/strimzi/api/kafka/model/connect/build/DownloadableArtifact.java
```diff
@@ -30,8 +29,8 @@
     @Description("URL of the artifact which will be downloaded. " +
             "Strimzi does not do any security scanning of the downloaded artifacts. " +
             "For security reasons, you should first verify the artifacts manually and configure the checksum verification to make sure the same artifact is used in the automated build. " +
-            "Required.")
-    @JsonProperty(required = true)
+            "Required for `jar`, `zip`, and `tgz` artifacts. " +
```
scholzj, line 32: And `other`?

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaConnectDockerfile.java
```diff
@@ -144,6 +209,24 @@ private void addPlugin(PrintWriter writer, Plugin plugin)    {
         }
     }
 
+    private void validateUrlPresence(DownloadableArtifact art) {
+        if (art.getUrl() == null) {
```
scholzj, line 213: I know URL validation is hard. But should we check it is at least not empty?

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaConnectDockerfile.java
```diff
@@ -144,6 +209,24 @@ private void addPlugin(PrintWriter writer, Plugin plugin)    {
...
+    }
+
+    private void validateGavPresence(MavenArtifact art) {
+        if (art.getGroup() == null) {
+            throw new InvalidConfigurationException("`" + art.getType() + "` artifact is missing a group ID.");
+        }
+        if (art.getArtifact() == null) {
+            throw new InvalidConfigurationException("`" + art.getType() + "` artifact is missing an artifact ID.");
+        }
+        if (art.getVersion() == null) {
```
scholzj, line 225: Same as above, should we check it is not empty string as well?

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaConnectDockerfile.java
```diff
@@ -61,6 +73,57 @@ public KafkaConnectDockerfile(String fromImage, Build connectBuild) {
...
+
+        if (mvnArtifacts.size() > 0) {
+            writer.println("FROM " + mavenImage + " AS downloadArtifacts");
+            mvnArtifacts.forEach(mvn -> {
+                checkForShellMetachars(((MavenArtifact) mvn).getGroup());
+                checkForShellMetachars(((MavenArtifact) mvn).getArtifact());
+                checkForShellMetachars(((MavenArtifact) mvn).getVersion());
+                String repo = ((MavenArtifact) mvn).getRepository() == null ? MavenArtifact.DEFAULT_REPOSITORY : maybePatchRepository(((MavenArtifact) mvn).getRepository());
+                String artifactHash = Util.sha1Prefix(((MavenArtifact) mvn).getGroup() + ((MavenArtifact) mvn).getArtifact() + ((MavenArtifact) mvn).getVersion());
+                String downloadPomCmd = String.format("curl -L --create-dirs --output /tmp/%s/pom.xml %s%s/%s/%s/%s-%s.pom",
```
scholzj, line 92: You do not seem to include connector into the path. What will happen if two connectors have the same artifact? Will it not fail because of some _pre-existing_ files etc.? (same of course for the other commands)

cluster-operator/src/test/java/io/strimzi/operator/cluster/model/KafkaConnectDockerfileTest.java
```diff
@@ -396,4 +402,146 @@ public void testDockerfileWithComments()   {
...
                 "USER 1001\n\n"));
     }
+
+    @ParallelTest
+    public void testNoUrlWhenRequired() {
+        Artifact tgzArtifact = new TgzArtifactBuilder()
+                .build();
+        Artifact jarArtifact = new JarArtifactBuilder()
+                .build();
+        Artifact zipArtifact = new ZipArtifactBuilder()
```
scholzj, line 412: Should this also cover `other` artifact?

scholzj (commented):
Some more nits.

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaConnectDockerfile.java
```diff
@@ -61,6 +73,57 @@ public KafkaConnectDockerfile(String fromImage, Build connectBuild) {
...
+                writer.println("      && " + downloadJarCmd);
+                writer.println();
+            });
+        }
+    }
+
+    private void checkForShellMetachars(String toBeChecked) {
+        Matcher hasSpecial = special.matcher(toBeChecked);
+        if (hasSpecial.find()) {
+            throw new InvalidConfigurationException("String '" + toBeChecked + "' contains forbidden character");
```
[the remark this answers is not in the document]
sknot-rh: Could we do the URL in another PR? I mean, it is better readable if we are not mixing up two things.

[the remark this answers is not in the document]
tombentley: Don't validate, quote. It's not that hard. You just enclose each argument in `'`, except `'` which you enclose in `"`. I.e.

* `foo` -> `'foo'`
* `foo'bar` -> `'foo'"'"'bar'`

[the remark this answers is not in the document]
tombentley: So you write a little helper class with an API a bit like StringBuilder, where you `append()` whole arguments at a time and do the quoting there. Then you have a `build()` method which gives you the fully quoted command.

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaConnectDockerfile.java
```diff
@@ -95,25 +96,8 @@ private void connectorPluginsPreStage(PrintWriter writer, List<Plugin> plugins)
                     checkForShellMetachars(repo);
                     String artifactHash = Util.sha1Prefix(((MavenArtifact) mvn).getGroup() + ((MavenArtifact) mvn).getArtifact() + ((MavenArtifact) mvn).getVersion());
                     String artifactDir = plugin.getKey() + "/" + artifactHash;
```
tombentley, line 98: `plugin.getKey()` is user supplied, right? In which case we need to use `Cmd` on the `mvn dependency:copy-dependencies` part.

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaConnectDockerfile.java
```diff
@@ -264,9 +259,9 @@ private void addOtherArtifact(PrintWriter writer, String connectorPath, OtherArt
         String artifactDir = connectorPath + "/" + artifactHash;
         String fileName = other.getFileName() != null ? other.getFileName() : artifactHash;
         String artifactPath = artifactDir + "/" + fileName;
-        String downloadCmd =  "curl -L --output " + artifactPath + " " + other.getUrl();
+        Cmd downloadCmd =  new Cmd("curl", "-L", "--output").append(artifactPath).append(other.getUrl());
```
tombentley, line 262: We don't really need the `.append` here, they're just extra arguments to the `...` parameter.

tombentley: Also, doesn't `artifactPath` depend on untrusted input?

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaConnectDockerfile.java
```diff
@@ -124,6 +108,17 @@ private void connectorPluginsPreStage(PrintWriter writer, List<Plugin> plugins)
...
+                repo,
+                mvn.getGroup().replace(".", "/"),
+                mvn.getArtifact().replace(".", "/"),
+                mvn.getVersion(),
+                mvn.getArtifact(),
+                mvn.getVersion(),
+                extension);
+    }
+
     private void checkForShellMetachars(String toBeChecked) {
```
tombentley, line 122: Do we still need this?

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaConnectDockerfile.java
```diff
@@ -264,9 +259,9 @@ private void addOtherArtifact(PrintWriter writer, String connectorPath, OtherArt
         String artifactDir = connectorPath + "/" + artifactHash;
         String fileName = other.getFileName() != null ? other.getFileName() : artifactHash;
         String artifactPath = artifactDir + "/" + fileName;
-        String downloadCmd =  "curl -L --output " + artifactPath + " " + other.getUrl();
+        Cmd downloadCmd =  new Cmd("curl", "-L", "--output").append(artifactPath).append(other.getUrl());
 
-        addUnmodifiedArtifact(writer, other, artifactDir, downloadCmd, artifactPath);
+        addUnmodifiedArtifact(writer, other, artifactDir, downloadCmd.toString(), artifactPath);
```
tombentley, line 264: It would be more typesafe to pass `downloadCmd` as a `Cmd` rather than a `String`, that within `addUnmodifiedArtifact` where we use it we know it must already be quoted.

packaging/helm-charts/helm3/strimzi-kafka-operator/values.yaml
```diff
@@ -120,6 +120,12 @@ kanikoExecutor:
     repository: strimzi
     name: kaniko-executor
     tag: latest
+mavenImage:
+  image:
+    registry: quay.io
+    repository: strimzi
+    name: maven-builder
+    tag: latest
```
scholzj, line 128: Should these be also added the the README.md of HelmChart?

api/src/main/java/io/strimzi/api/kafka/Cmd.java
```diff
@@ -0,0 +1,49 @@
+/*
+ * Copyright Strimzi authors.
+ * License: Apache License 2.0 (see the file LICENSE or http://apache.org/licenses/LICENSE-2.0.html).
+ */
+package io.strimzi.api.kafka;
+
+import java.util.Objects;
+
+public class Cmd {
```
scholzj, line 9: Are you sure the `api` is the best place for this? Would it be better in operator common if it needs to be shared or something like that?

scholzj (commented):
Two nits, LGTM otherwise.

packaging/helm-charts/helm3/strimzi-kafka-operator/values.yaml
sknot-rh: https://github.com/strimzi/strimzi-kafka-operator/blob/main/helm-charts/DO_NOT_EDIT.md

scholzj (approved):

tombentley (approved):

Issue #5372, opened 2021-07-29
Make resources section in kafka initContainer configurable as this is hardcoded currently


hari819:
### Discussed in https://github.com/strimzi/strimzi-kafka-operator/discussions/5369

<div type='discussions-op-text'>

<sup>Originally posted by **hari819** July 29, 2021</sup>
Hello Team,

i would like to remove the resources section by adding empty {} in the kafka STS section for the initContainer , because sometimes PODs are unable to start with insufficient cpu error ,  I was able to do this for the main Kafka container using 

  ```
apiVersion: kafka.strimzi.io/v1beta2
kind: Kafka
metadata:
  name: kafka
spec:
  kafka:
    version: 2.7.0
    replicas: 3
    template:
      pod:
        resources: {} #disable requests
```

but i am unable to set the same for the initContainer , 

  ```
initContainers:
    - name: kafka-init
      image: 'quay.io/strimzi/operator:0.22.1'
      args:
        - /opt/strimzi/bin/kafka_init_run.sh
      resources:
        limits:
          cpu: '1'
          memory: 256Mi
        requests:
          cpu: 100m
          memory: 128Mi
```

could you please help me on this ,</div>

sknot-rh:
It seems to be hardcoded for [Kafka](https://github.com/strimzi/strimzi-kafka-operator/blob/33da771f49456935ab6f2122695db4f925879c96/cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaCluster.java#L1536) and [Kafka Connect](https://github.com/strimzi/strimzi-kafka-operator/blob/d7161c6db23d384a4e8f5ec810dbc53418ed4122/cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaConnectCluster.java#L539).
@strimzi/maintainers what do you think about configuring `initContainer` resources in the template?

scholzj:
That was my original thought - we could add resources into the `initContainer` template field. My only concern is if removing the defaults causes problem for someone with crazy small amount of resource and have the init-container failing.

But now I had another thought. The init-container is part of the Kafka pod, but should run always before the main container. Can't we just use the resources of the main Kafka / Connect container for the init container?
* If the user has a crazy small Limit Range default, they will need to configure the resources for the main container as well. If the main container runs without resource requests / limits, the init container should run as well.
[...]

PR #5408, opened 2021-08-09
Init Container uses same value of resources as kafka resources

ShubhamRwt:
### Type of change

_Select the type of your PR_

- Enhancement / new feature

### Description

This PR resolves the issue mentioned in #5372. This PR makes the init Container resources configurable in such a way that the value of resources which were configured for kafka resources are used for setting the resources of the Init Containter also.

### Checklist

- [ ] Write tests
- [ ] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaCluster.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaCluster.java
@@ -1545,28 +1545,18 @@ public class KafkaCluster extends AbstractModel {
         List<Container> initContainers = new ArrayList<>(1);
 
         if (rack != null || !ListenersUtils.nodePortListeners(listeners).isEmpty()) {
-            Container initContainer;
+            Container initContainer = new ContainerBuilder()
+                    .withName(INIT_NAME)
+                    .withImage(initImage)
+                    .withArgs("/opt/strimzi/bin/kafka_init_run.sh")
+                    .withEnv(getInitContainerEnvVars())
+                    .withVolumeMounts(VolumeUtils.createVolumeMount(INIT_VOLUME_NAME, INIT_VOLUME_MOUNT))
+                    .withImagePullPolicy(determineImagePullPolicy(imagePullPolicy, initImage))
+                    .withSecurityContext(templateInitContainerSecurityContext)
+                    .build();
+
             if (getResources() != null) {
-                initContainer = new ContainerBuilder()
-                        .withName(INIT_NAME)
-                        .withImage(initImage)
-                        .withArgs("/opt/strimzi/bin/kafka_init_run.sh")
-                        .withResources(getResources())
-                        .withEnv(getInitContainerEnvVars())
-                        .withVolumeMounts(VolumeUtils.createVolumeMount(INIT_VOLUME_NAME, INIT_VOLUME_MOUNT))
-                        .withImagePullPolicy(determineImagePullPolicy(imagePullPolicy, initImage))
-                        .withSecurityContext(templateInitContainerSecurityContext)
-                        .build();
-            } else {
-                initContainer = new ContainerBuilder()
-                        .withName(INIT_NAME)
-                        .withImage(initImage)
-                        .withArgs("/opt/strimzi/bin/kafka_init_run.sh")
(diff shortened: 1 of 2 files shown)


cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaCluster.java
```diff
@@ -1548,24 +1545,29 @@ protected Affinity getMergedAffinity() {
...
+            } else {
+                initContainer = new ContainerBuilder()
+                        .withName(INIT_NAME)
+                        .withImage(initImage)
+                        .withArgs("/opt/strimzi/bin/kafka_init_run.sh")
+                        .withEnv(getInitContainerEnvVars())
+                        .withVolumeMounts(VolumeUtils.createVolumeMount(INIT_VOLUME_NAME, INIT_VOLUME_MOUNT))
+                        .withImagePullPolicy(determineImagePullPolicy(imagePullPolicy, initImage))
+                        .withSecurityContext(templateInitContainerSecurityContext)
+                        .build();
```
scholzj, line 1569: This seems like it should work fine. But I think it duplicates too much code. Why don't you create the init container first and only then set the resources ... not a complete code, but something like this:

```
Container initContainer = new ContainerBuilder().......build();

if (getResources() != null) {
    initContainer.setResources(getResources());
}
```

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaConnectCluster.java
```diff
@@ -531,30 +529,37 @@ public Deployment generateDeployment(Map<String, String> annotations, boolean is
...
+                initContainer = new ContainerBuilder()
+                        .withName(INIT_NAME)
+                        .withImage(initImage)
+                        .withArgs("/opt/strimzi/bin/kafka_init_run.sh")
+                        .withEnv(getInitContainerEnvVars())
+                        .withVolumeMounts(VolumeUtils.createVolumeMount(INIT_VOLUME_NAME, INIT_VOLUME_MOUNT))
+                        .withImagePullPolicy(determineImagePullPolicy(imagePullPolicy, initImage))
+                        .withSecurityContext(templateInitContainerSecurityContext)
+                        .build();
+            }
```
scholzj, line 554: Same as above.

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaConnectCluster.java
```diff
@@ -531,30 +529,37 @@ public Deployment generateDeployment(Map<String, String> annotations, boolean is
...
     }
 
     private ResourceRequirements getInitContainerResourceResourceRequirements() {
-        return new ResourceRequirementsBuilder()
-                .addToRequests("cpu", new Quantity("100m"))
-                .addToRequests("memory", new Quantity("128Mi"))
-                .addToLimits("cpu", new Quantity("1"))
-                .addToLimits("memory", new Quantity("256Mi"))
-                .build();
+        return getResources();
```
scholzj, line 562: Is this `getInitContainerResourceResourceRequirements` method used also somewhere else? If not, do we really need it? Why not call `getResources()` directly from `getInitContainers()`? It seems not needed if it is not called from anywhere else.

ShubhamRwt: Yeah, that was my initial thought. I will check if it is being used somewhere and if not, I will exclude it.

scholzj: Even if it is used somewhere else, you should think if it makes sense to have a separate method for it or not.

ShubhamRwt: It is not being used at any other place

scholzj (commented):
The code looks good. But checkstyle is failing, so you will need to have a look at it.

ShubhamRwt:
Resolved @scholzj

scholzj (approved):

ppatierno (approved):

sknot-rh (approved):

PR #5414, opened 2021-08-10
[MO] - upgrading test-container to be more flexible + multinode support

see-quick:

### Type of change

- Enhancement / new feature
- Refactoring

### Description

This PR adding support for custom configuration of `strimzi-test-container`.  Kafka configs are injected by map which can be specified in the Container constructor. Moreover, I have removed dependency on EmbeddedKafkaCluster and replaced it with  `strimzi-test-container`.

### Checklist

### DIFF
--- a/cluster-operator/pom.xml
+++ b/cluster-operator/pom.xml
@@ -279,11 +279,46 @@
             <version>${mockserver.version}</version>
             <scope>test</scope>
         </dependency>
+        <!-- Used by EmbeddedKafkaCluster -->
         <dependency>
-            <groupId>io.strimzi</groupId>
-            <artifactId>strimzi-test-container</artifactId>
-            <version>0.26.0-SNAPSHOT</version>
+            <groupId>org.apache.kafka</groupId>
+            <artifactId>kafka-streams</artifactId>
+            <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.apache.kafka</groupId>
+            <artifactId>kafka-streams</artifactId>
+            <classifier>test</classifier>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.kafka</groupId>
+            <artifactId>kafka-clients</artifactId>
+            <classifier>test</classifier>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.kafka</groupId>
+            <artifactId>kafka_2.13</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.kafka</groupId>
+            <artifactId>kafka_2.13</artifactId>
+            <classifier>test</classifier>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.junit.jupiter</groupId>
+            <artifactId>junit-jupiter-migrationsupport</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.kafka</groupId>
+            <artifactId>kafka-streams</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <!-- ^^^^^ Used by EmbeddedKafkaCluster -->
     </dependencies>
 
     <build>
--- a/cluster-operator/pom.xml
+++ b/cluster-operator/pom.xml
@@ -199,11 +199,6 @@
         <dependency>
             <groupId>org.apache.kafka</groupId>
             <artifactId>kafka-streams</artifactId>
-            <scope>test</scope>
-        </dependency>
-        <dependency>
-            <groupId>org.apache.kafka</groupId>
-            <artifactId>kafka-streams</artifactId>
             <classifier>test</classifier>
             <scope>test</scope>
         </dependency>
@@ -262,8 +257,6 @@
         <dependency>
             <groupId>com.github.spotbugs</groupId>
             <artifactId>spotbugs-annotations</artifactId>
-            <version>${spotbugs.version}</version>
-            <scope>provided</scope>
         </dependency>
         <dependency>
             <groupId>org.apache.kafka</groupId>
--- a/cluster-operator/pom.xml
+++ b/cluster-operator/pom.xml
@@ -195,41 +195,6 @@
             <artifactId>connect-api</artifactId>
             <scope>test</scope>
         </dependency>
-        <!-- Used by EmbeddedKafkaCluster -->
-        <dependency>
-            <groupId>org.apache.kafka</groupId>
-            <artifactId>kafka-streams</artifactId>
-            <scope>test</scope>
-        </dependency>
-        <dependency>
-            <groupId>org.apache.kafka</groupId>
-            <artifactId>kafka-streams</artifactId>
-            <classifier>test</classifier>
-            <scope>test</scope>
-        </dependency>
-        <dependency>
-            <groupId>org.apache.kafka</groupId>
-            <artifactId>kafka-clients</artifactId>
-            <classifier>test</classifier>
-            <scope>test</scope>
-        </dependency>
-        <dependency>
-            <groupId>org.apache.kafka</groupId>
-            <artifactId>kafka_2.13</artifactId>
-            <scope>test</scope>
-        </dependency>
-        <dependency>
-            <groupId>org.apache.kafka</groupId>
-            <artifactId>kafka_2.13</artifactId>
-            <classifier>test</classifier>
-            <scope>test</scope>
-        </dependency>
-        <dependency>
-            <groupId>org.junit.jupiter</groupId>
-            <artifactId>junit-jupiter-migrationsupport</artifactId>
-            <scope>test</scope>
-        </dependency>
-        <!-- ^^^^^ Used by EmbeddedKafkaCluster -->
         <dependency>
             <groupId>org.apache.kafka</groupId>
             <artifactId>connect-runtime</artifactId>
@@ -313,6 +278,11 @@
             <artifactId>mockserver-client-java</artifactId>
             <version>${mockserver.version}</version>
             <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>io.strimzi</groupId>
+            <artifactId>strimzi-test-container</artifactId>
+            <version>0.26.0-SNAPSHOT</version>
         </dependency>
     </dependencies>
 
--- a/cluster-operator/pom.xml
+++ b/cluster-operator/pom.xml
@@ -309,6 +309,11 @@
             <scope>test</scope>
         </dependency>
         <dependency>
+            <groupId>org.apache.kafka</groupId>
+            <artifactId>kafka_2.13</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
             <groupId>org.junit.jupiter</groupId>
             <artifactId>junit-jupiter-migrationsupport</artifactId>
             <scope>test</scope>
--- a/cluster-operator/pom.xml
+++ b/cluster-operator/pom.xml
@@ -199,6 +199,11 @@
         <dependency>
             <groupId>org.apache.kafka</groupId>
             <artifactId>kafka-streams</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.kafka</groupId>
+            <artifactId>kafka-streams</artifactId>
             <classifier>test</classifier>
             <scope>test</scope>
         </dependency>
@@ -222,11 +227,6 @@
         <dependency>
             <groupId>org.junit.jupiter</groupId>
             <artifactId>junit-jupiter-migrationsupport</artifactId>
-            <scope>test</scope>
-        </dependency>
-        <dependency>
-            <groupId>org.apache.kafka</groupId>
-            <artifactId>kafka-streams</artifactId>
             <scope>test</scope>
         </dependency>
         <!-- ^^^^^ Used by EmbeddedKafkaCluster -->
--- a/cluster-operator/pom.xml
+++ b/cluster-operator/pom.xml
@@ -195,6 +195,51 @@
             <artifactId>connect-api</artifactId>
             <scope>test</scope>
         </dependency>
+        <!-- Used by EmbeddedKafkaCluster -->
+        <dependency>
+            <groupId>org.apache.kafka</groupId>
+            <artifactId>kafka-streams</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.kafka</groupId>
+            <artifactId>kafka-streams</artifactId>
+            <classifier>test</classifier>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.kafka</groupId>
+            <artifactId>kafka-clients</artifactId>
+            <classifier>test</classifier>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.kafka</groupId>
+            <artifactId>kafka_2.13</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.kafka</groupId>
+            <artifactId>kafka_2.13</artifactId>
+            <classifier>test</classifier>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.kafka</groupId>
+            <artifactId>kafka_2.13</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.junit.jupiter</groupId>
+            <artifactId>junit-jupiter-migrationsupport</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.kafka</groupId>
+            <artifactId>kafka-streams</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <!-- ^^^^^ Used by EmbeddedKafkaCluster -->
         <dependency>
             <groupId>org.apache.kafka</groupId>
             <artifactId>connect-runtime</artifactId>
@@ -279,51 +324,6 @@
             <version>${mockserver.version}</version>
             <scope>test</scope>
         </dependency>
-        <!-- Used by EmbeddedKafkaCluster -->
-        <dependency>
-            <groupId>org.apache.kafka</groupId>
-            <artifactId>kafka-streams</artifactId>
-            <scope>test</scope>
-        </dependency>
-        <dependency>
-            <groupId>org.apache.kafka</groupId>
-            <artifactId>kafka-streams</artifactId>
-            <classifier>test</classifier>
-            <scope>test</scope>
-        </dependency>
-        <dependency>
-            <groupId>org.apache.kafka</groupId>
-            <artifactId>kafka-clients</artifactId>
-            <classifier>test</classifier>
-            <scope>test</scope>
-        </dependency>
-        <dependency>
-            <groupId>org.apache.kafka</groupId>
-            <artifactId>kafka_2.13</artifactId>
-            <scope>test</scope>
-        </dependency>
-        <dependency>
-            <groupId>org.apache.kafka</groupId>
-            <artifactId>kafka_2.13</artifactId>
-            <classifier>test</classifier>
-            <scope>test</scope>
-        </dependency>
-        <dependency>
-            <groupId>org.apache.kafka</groupId>
-            <artifactId>kafka_2.13</artifactId>
-            <scope>test</scope>
-        </dependency>
-        <dependency>
-            <groupId>org.junit.jupiter</groupId>
-            <artifactId>junit-jupiter-migrationsupport</artifactId>
-            <scope>test</scope>
-        </dependency>
-        <dependency>
-            <groupId>org.apache.kafka</groupId>
-            <artifactId>kafka-streams</artifactId>
-            <scope>test</scope>
-        </dependency>
-        <!-- ^^^^^ Used by EmbeddedKafkaCluster -->
     </dependencies>
 
     <build>
--- a/cluster-operator/pom.xml
+++ b/cluster-operator/pom.xml
@@ -225,11 +225,6 @@
             <scope>test</scope>
         </dependency>
         <dependency>
-            <groupId>org.apache.kafka</groupId>
-            <artifactId>kafka_2.13</artifactId>
-            <scope>test</scope>
-        </dependency>
-        <dependency>
             <groupId>org.junit.jupiter</groupId>
             <artifactId>junit-jupiter-migrationsupport</artifactId>
             <scope>test</scope>
@@ -267,6 +262,8 @@
         <dependency>
             <groupId>com.github.spotbugs</groupId>
             <artifactId>spotbugs-annotations</artifactId>
+            <version>${spotbugs.version}</version>
+            <scope>provided</scope>
         </dependency>
         <dependency>
             <groupId>org.apache.kafka</groupId>
--- a/cluster-operator/src/test/java/io/strimzi/operator/cluster/operator/assembly/KafkaConnectorIT.java
+++ b/cluster-operator/src/test/java/io/strimzi/operator/cluster/operator/assembly/KafkaConnectorIT.java
@@ -42,7 +42,6 @@ import java.io.IOException;
 import java.util.Collections;
 import java.util.LinkedHashMap;
 import java.util.Map;
-import java.util.concurrent.ExecutionException;
 import java.util.concurrent.TimeUnit;
 
 import static org.hamcrest.CoreMatchers.instanceOf;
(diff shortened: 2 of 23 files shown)


scholzj (commented):
Looks good. I think @tombentley and @sknot-rh  might know better about how needed is the multi-broker setup. 

Do you plan to do the same also for the User and Topic Operators?

Frawless:
I think a multinode setup would be useful not just for us, but for the community as well.

test-container/src/main/java/io/strimzi/StrimziKafkaContainer.java
```diff
@@ -20,7 +20,9 @@
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
+import java.util.HashMap;
 import java.util.List;
+import java.util.Map;
 
 public class StrimziKafkaContainer extends GenericContainer<StrimziKafkaContainer> {
```
tombentley, line 27: Can we add a sentence of two of Javadoc explaining what this is and how it's intended to be used?

test-container/src/main/java/io/strimzi/StrimziKafkaContainer.java
```diff
@@ -73,18 +76,24 @@
         LOGGER.info("Supported Strimzi version: {}", STRIMZI_VERSION);
     }
 
-    public StrimziKafkaContainer(final String version) {
+    public StrimziKafkaContainer(final String version, Map<String, String> additionalKafkaConfiguration) {
```
tombentley, line 79: Perhaps rename `version` to `imageVersion`, since that's what it seems to be (rather than a Kafka version string).

test-container/src/main/java/io/strimzi/StrimziKafkaContainer.java
```diff
@@ -133,13 +142,23 @@ protected void containerIsStarting(InspectContainerResponse containerInfo, boole
...
+        // default listener config
+        kafkaConfiguration.append(" --override listeners=" + kafkaListeners + "PLAINTEXT://0.0.0.0:" + KAFKA_PORT);
+        kafkaConfiguration.append(" --override advertised.listeners=" + advertisedListeners);
+        kafkaConfiguration.append(" --override zookeeper.connect=localhost:" + ZOOKEEPER_PORT);
+        kafkaConfiguration.append(" --override listener.security.protocol.map=" + kafkaListenerSecurityProtocol + "PLAINTEXT:PLAINTEXT");
+        kafkaConfiguration.append(" --override inter.broker.listener.name=BROKER1");
+
+        // additional kafka config
+        kafkaConfigurationMap.forEach((configName, configValue) -> {
+            kafkaConfiguration.append(" --override " + configName + "=" + configValue);
```
tombentley, line 156: Validate that none of the `configName` is one of the hard coded configs.

Also, this could be a `.map(...).forEach(kafkaConfiguration::append)`, I think.

test-container/src/main/java/io/strimzi/StrimziKafkaContainer.java
```diff
@@ -133,13 +142,23 @@ protected void containerIsStarting(InspectContainerResponse containerInfo, boole
...
         });
 
+        StringBuilder kafkaConfiguration = new StringBuilder();
+
+        // default listener config
+        kafkaConfiguration.append(" --override listeners=" + kafkaListeners + "PLAINTEXT://0.0.0.0:" + KAFKA_PORT);
+        kafkaConfiguration.append(" --override advertised.listeners=" + advertisedListeners);
+        kafkaConfiguration.append(" --override zookeeper.connect=localhost:" + ZOOKEEPER_PORT);
+        kafkaConfiguration.append(" --override listener.security.protocol.map=" + kafkaListenerSecurityProtocol + "PLAINTEXT:PLAINTEXT");
+        kafkaConfiguration.append(" --override inter.broker.listener.name=BROKER1");
```
tombentley, line 152: You could chain the appends together, and onto the `new StringBuilder`.

tombentley (commented):
We should be able to update `TopicOperatorIT` to use this too, right? (Can be a separate PR)

see-quick:
> I think a multinode setup would be useful not just for us, but for the community as well.

Your wishes come true 🧙‍♀️🦄

test-container/src/main/java/io/strimzi/StrimziKafkaContainer.java
see-quick: I don't want to forbid users from these options. For instance, if the user wants to override`listeners` or `zookeeper.connect` config is ok or?

see-quick:
I will do update for `TopicOperatorIT` in separate PR @tombentley

test-container/src/main/java/io/strimzi/StrimziKafkaContainer.java
tombentley: Fair enough, but does `--override` support later options overriding previous ones?

see-quick: Yes It does )

Frawless (approved):
I think it would be nice to write a few sentences about it (how to use it, etc), but maybe after the migration to the new repo.

see-quick:
@Frawless in the brand new repository, which should be created as this PR will be merged I will do it. Moreover, I thought about some BlogPost to write about it...I'm wondering if it's a good idea.

cluster-operator/src/test/java/io/strimzi/operator/cluster/operator/assembly/KafkaConnectorIT.java
```diff
@@ -67,24 +76,40 @@ public static void before() throws IOException {
...
         String connectClusterName = getClass().getSimpleName();
-        cluster.createTopics(connectClusterName + "-offsets", connectClusterName + "-config", connectClusterName + "-status");
+
+        CreateTopicsResult createTopicsResult = adminClient.createTopics(
+            Arrays.asList(
+                new NewTopic(connectClusterName + "-offsets", 3, (short) 3),
+                new NewTopic(connectClusterName + "-config", 3,  (short) 3),
+                new NewTopic(connectClusterName + "-status", 3,  (short) 3)
+            ));
+        createTopicsResult.all().get();
```
scholzj, line 107: Why is this needed? Shouldn't Connect create this automatically?

test-container/pom.xml
```diff
@@ -23,6 +23,16 @@
...
+        <dependency>
+            <groupId>com.github.spotbugs</groupId>
+            <artifactId>spotbugs-annotations</artifactId>
+        </dependency>
+        <!-- for TestUtils.waifFor       -->
+        <dependency>
+            <groupId>io.strimzi</groupId>
+            <artifactId>test</artifactId>
+            <scope>compile</scope>
+        </dependency>
```
scholzj, line 35: This will not work well once we move it to the sepsrate repo.

see-quick: I know but should I insert also the implementation of `waitFor` in some utils?

cluster-operator/src/test/java/io/strimzi/operator/cluster/operator/assembly/KafkaConnectorIT.java
see-quick: I don't know I just re-write this code to be equivalent by instead of using Kafka cluster to create it I am using AdminClient. I can give it a try....and see what happens...

see-quick: I see that @tombentley was first, which wrote that `cluster.createTopics(getClass().getSimpleName() + "-offsets", getClass().getSimpleName() + "-config", getClass().getSimpleName() + "-status");`... I tried to remove that code and test also passed so I don't really know :D 🤷‍♂️

cluster-operator/src/test/java/io/strimzi/operator/cluster/operator/assembly/KafkaConnectorIT.java
```diff
@@ -67,8 +70,11 @@ public static void before() throws IOException {
                         .setEnabled(true)
         ));
 
-        cluster = new EmbeddedKafkaCluster(3);
-        cluster.start();
+        kafkaCluster = new EmbeddedKafkaCluster(3);
+        kafkaCluster.start();
+
+        Properties properties = new Properties();
+        properties.setProperty(AdminClientConfig.BOOTSTRAP_SERVERS_CONFIG, kafkaCluster.bootstrapServers());
```
scholzj, line 77: I guess this is not needed?

test-container/pom.xml
```diff
@@ -23,6 +23,10 @@
             <groupId>org.testcontainers</groupId>
             <artifactId>testcontainers</artifactId>
         </dependency>
+        <dependency>
+            <groupId>com.github.spotbugs</groupId>
+            <artifactId>spotbugs-annotations</artifactId>
+        </dependency>
```
scholzj, line 29: Should this be set as provided so it is not pulled unnecessarily?

cluster-operator/pom.xml
```diff
@@ -314,6 +279,51 @@
             <version>${mockserver.version}</version>
             <scope>test</scope>
         </dependency>
+        <!-- Used by EmbeddedKafkaCluster -->
```
scholzj, line 282: Can we maybe move this to the original place to avoid unnecessary changes?

scholzj (approved):

scholzj:
@tombentley Do you have something more to this?

test-container/src/main/java/io/strimzi/StrimziKafkaCluster.java
```diff
@@ -0,0 +1,143 @@
...
+    @Override
+    public void start() {
+        Stream<Startable> startables = this.brokers.stream().map(Startable.class::cast);
+        try {
+            Startables.deepStart(startables).get(60, TimeUnit.SECONDS);
+        } catch (InterruptedException | ExecutionException | TimeoutException e) {
+            e.printStackTrace();
+        }
+
+        TestUtils.waitFor("Zookeeper readiness", Duration.ofSeconds(1).toMillis(), Duration.ofSeconds(30).toMillis(),
```
tombentley, line 112: We're not waiting for "Zookeeper readiness" here. We're waiting for the broker znode.

test-container/src/main/java/io/strimzi/StrimziKafkaCluster.java
```diff
@@ -0,0 +1,143 @@
...
+                    e.printStackTrace();
+                    return false;
+                }
+            });
+    }
+
+    @Override
+    public void stop() {
+        // stop all containers in parallel
+        Stream.concat(this.brokers.stream(), Stream.of(this.zookeeper))
```
tombentley, line 135: On the command line if I kill ZK first it sometimes prevents a broker from shutting down quickly. So maybe it's better to do the ZK shudown first, and then shut down the brokers in parallel.

test-container/src/main/java/io/strimzi/StrimziZookeeperContainer.java
```diff
@@ -0,0 +1,97 @@
...
+        LOGGER.info("This is mapped port {}", zookeeperExposedPort);
+
+        String command = "#!/bin/bash \n";
+
+        command += "bin/zookeeper-server-start.sh config/zookeeper.properties\n";
+
+        LOGGER.info("Copying command to 'STARTER_SCRIPT' script.");
+
+        copyFileToContainer(
+            Transferable.of(command.getBytes(StandardCharsets.UTF_8), 700),
```
tombentley, line 80: What's the intent behind having this starter script, if it just calls `zookeeper-server-start.sh`?

test-container/src/main/java/io/strimzi/StrimziZookeeperContainer.java
```diff
@@ -0,0 +1,97 @@
...
+    public boolean equals(Object o) {
+        if (this == o) return true;
+        if (o == null || getClass() != o.getClass()) return false;
+        if (!super.equals(o)) return false;
+        StrimziZookeeperContainer that = (StrimziZookeeperContainer) o;
+        return zookeeperExposedPort == that.zookeeperExposedPort;
+    }
+    @Override
+    public int hashCode() {
+        return Objects.hash(super.hashCode(), zookeeperExposedPort);
```
tombentley, line 95: Including `super.hashCode` like this seems to result in a `hashCode()` that's incompatible with `equals()`.

test-container/src/main/java/io/strimzi/StrimziZookeeperContainer.java
```diff
@@ -0,0 +1,97 @@
...
+        );
+    }
+
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) return true;
+        if (o == null || getClass() != o.getClass()) return false;
+        if (!super.equals(o)) return false;
+        StrimziZookeeperContainer that = (StrimziZookeeperContainer) o;
+        return zookeeperExposedPort == that.zookeeperExposedPort;
```
tombentley, line 91: This isn't really sufficient for equality. I could have two instances, and the exposed port constraint means only one can run at once, but both instances can exist. Do we need these containers to override equals?

test-container/src/main/java/io/strimzi/utils/WaitException.java
```diff
@@ -0,0 +1,15 @@
+/*
+ * Copyright Strimzi authors.
+ * License: Apache License 2.0 (see the file LICENSE or http://apache.org/licenses/LICENSE-2.0.html).
+ */
+package io.strimzi.utils;
+
```
tombentley, line 6: A line of Javadoc please, to explain what a WaitException is supposed to be conceptually.

test-container/src/main/java/io/strimzi/StrimziKafkaCluster.java
see-quick: good point yes

test-container/src/main/java/io/strimzi/StrimziZookeeperContainer.java
see-quick: It will never call `just` zookeeper-server-start.sh. There are two paths:

- configured to use external zookeeper -> 1. kafka-server-start.sh
- configured to use embedded zookeeper -> 1. zookeeper-server-start.sh, 2. kafka-server-start.sh 

the whole idea of this STARTING script is because we need to change the advertised configuration of the Kafka container before it starts...

test-container/src/main/java/io/strimzi/StrimziZookeeperContainer.java
see-quick: I think we don't need them and I could add spot bugs suppress annotation

scholzj:
@see-quick I'm not completely sure if we should merge it without approval from @tombentley. Is it blocking you in some further work on the test container?

test-container/src/main/java/io/strimzi/StrimziKafkaContainer.java
```diff
@@ -94,13 +112,18 @@ protected void doStart() {
         super.doStart();
     }
 
+    public StrimziKafkaContainer withExternalZookeeper(String connectString) {
+        this.externalZookeeperConnect = connectString;
+        return self();
+    }
```
tombentley, line 118: Perhaps private constructors and public factory methods to separate the two flavours would be better that this mutable field needing to be set before starting the container?

test-container/src/main/java/io/strimzi/StrimziZookeeperContainer.java
```diff
@@ -0,0 +1,85 @@
...
+import org.testcontainers.containers.Network;
+import org.testcontainers.images.builder.Transferable;
+
+import java.nio.charset.StandardCharsets;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+
+/**
+ * StrimziZookeeperContainer is an instance of the Zookeeper encapsulated inside a docker container using the latest image
```
tombentley, line 21: Similar rewording for the parametrised image version.

test-container/src/main/java/io/strimzi/utils/TestUtils.java
```diff
@@ -0,0 +1,78 @@
...
+                waitException.printStackTrace();
+                throw waitException;
+            }
+            long sleepTime = Math.min(pollIntervalMs, timeLeft);
+            if (LOGGER.isTraceEnabled()) {
+                LOGGER.trace("{} not ready, will try again in {} ms ({}ms till timeout)", description, sleepTime, timeLeft);
+            }
+            try {
+                Thread.sleep(sleepTime);
+            } catch (InterruptedException e) {
```
tombentley, line 73: Probably better to throw `InterruptedException`.

test-container/src/main/java/io/strimzi/utils/TestUtils.java
```diff
@@ -0,0 +1,78 @@
...
+            if (timeLeft <= 0) {
+                if (exceptionCount > 1) {
+                    LOGGER.error("Exception waiting for {}, {}", description, exceptionMessage);
+
+                    if (!stackTraceError.toString().isEmpty()) {
+                        // printing handled stacktrace
+                        LOGGER.error(stackTraceError.toString());
+                    }
+                }
+                onTimeout.run();
```
tombentley, line 62: This whole thing with `stackTraceError` and the calls to `printStackTrace` is a bit horrible. It's not really your job here to decide what should be printed to stderr. Perhaps using suppressed exceptions would be better? Better if you only have a single suppressed exception though.

test-container/src/main/java/io/strimzi/utils/TestUtils.java
```diff
@@ -0,0 +1,78 @@
...
+                WaitException waitException = new WaitException("Timeout after " + timeoutMs + " ms waiting for " + description);
+                waitException.printStackTrace();
+                throw waitException;
+            }
+            long sleepTime = Math.min(pollIntervalMs, timeLeft);
+            if (LOGGER.isTraceEnabled()) {
+                LOGGER.trace("{} not ready, will try again in {} ms ({}ms till timeout)", description, sleepTime, timeLeft);
```
tombentley, line 69: Perhaps "satisfied" is better than "ready"?

test-container/src/test/java/io/strimzi/StrimziKafkaClusterTest.java
```diff
@@ -0,0 +1,160 @@
...
+
+        Map<String, String> kafkaClusterConfiguration = new HashMap<>();
+        kafkaClusterConfiguration.put("zookeeper.connect", "zookeeper:2181");
+
+        systemUnderTest = new StrimziKafkaCluster(StrimziKafkaContainer.getStrimziVersion() + "-kafka-" + StrimziKafkaContainer.getLatestKafkaVersion(),
+            numberOfBrokers,
+            numberOfReplicas,
+            kafkaClusterConfiguration);
+        systemUnderTest.start();
+
```
tombentley, line 62: This is identical for both tests -> `@BeforeEach`

test-container/src/test/java/io/strimzi/StrimziKafkaClusterTest.java
```diff
@@ -0,0 +1,160 @@
...
+        );
+
+        final String brokers = result.getStdout();
+        // verify that all kafka brokers are bind to zookeeper
+        assertThat(brokers, notNullValue());
+        assertThat(brokers.split(",").length, is(numberOfBrokers));
+
+        LOGGER.info("Brokers are {}", systemUnderTest.getBootstrapServers());
+
+        systemUnderTest.stop();
```
tombentley, line 76: `@AfterEach`

test-container/src/test/java/io/strimzi/StrimziKafkaClusterTest.java
```diff
@@ -0,0 +1,160 @@
...
+                    return true;
+                });
+
+            consumer.unsubscribe();
+        } catch (InterruptedException e) {
+            e.printStackTrace();
+        } catch (ExecutionException e) {
+            e.printStackTrace();
+        } catch (TimeoutException e) {
+            e.printStackTrace();
```
tombentley, line 155: Calling `printStackTrace` is almost never the right thing to do, it's usually a sign that exception handling isn't really being done properly. In this case, because it's a test, just declare `throws Exception`.

test-container/src/main/java/io/strimzi/StrimziKafkaCluster.java
```diff
@@ -0,0 +1,146 @@
...
+
+        // multi-node set up
+        this.brokers = IntStream
+            .range(0, this.brokersNum)
+            .mapToObj(brokerId -> {
+                LOGGER.info("Starting broker with id {}", brokerId);
+                // adding broker id for each kafka container
+                additionalKafkaConfiguration.put("broker.id", String.valueOf(brokerId));
+
+                return new StrimziKafkaContainer(imageVersion, additionalKafkaConfiguration)
```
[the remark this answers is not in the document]
see-quick: Done

test-container/src/main/java/io/strimzi/cluster/StrimziKafkaCluster.java
```diff
@@ -0,0 +1,146 @@
+/*
+ * Copyright Strimzi authors.
+ * License: Apache License 2.0 (see the file LICENSE or http://apache.org/licenses/LICENSE-2.0.html).
+ */
+package io.strimzi.cluster;
```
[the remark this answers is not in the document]
see-quick: > Sorry, I should have been more specific before, let's use something like io.strimzi.test.container.

Done

test-container/src/main/java/io/strimzi/test/container/StrimziKafkaCluster.java
```diff
@@ -0,0 +1,144 @@
...
+public class StrimziKafkaCluster implements Startable {
+
+    private static final Logger LOGGER = LogManager.getLogger(StrimziKafkaCluster.class);
+
+    private final int brokersNum;
+    private final Network network;
+    private final StrimziZookeeperContainer zookeeper;
+    private final Collection<StrimziKafkaContainer> brokers;
+
+    @SuppressFBWarnings("BC_UNCONFIRMED_CAST_OF_RETURN_VALUE")
```
tombentley, line 44: Do we actually need this?

test-container/src/main/java/io/strimzi/test/container/StrimziKafkaCluster.java
```diff
@@ -0,0 +1,144 @@
...
+                return StrimziKafkaContainer.createWithAdditionalConfiguration(brokerId, imageVersion, additionalKafkaConfiguration)
+                    .withNetwork(this.network)
+                    .withNetworkAliases("broker-" + brokerId)
+                    .dependsOn(this.zookeeper)
+                    .withExternalZookeeper("zookeeper:" + StrimziZookeeperContainer.ZOOKEEPER_PORT);
+            })
+            .collect(Collectors.toList());
+    }
+
+    public StrimziKafkaCluster(Map<String, String> additionalKafkaConfiguration) {
```
tombentley, line 84: Is this actually used?

test-container/src/main/java/io/strimzi/test/container/StrimziKafkaContainer.java
```diff
@@ -20,20 +20,33 @@
...
-    private static final int KAFKA_PORT = 9092;
-    private static final int ZOOKEEPER_PORT = 2181;
     private static final String LATEST_KAFKA_VERSION;
     private static final List<String> SUPPORTED_KAFKA_VERSIONS = new ArrayList<>(5);
     private static final String STRIMZI_VERSION;
 
+    private Map<String, String> kafkaConfigurationMap;
+    private String externalZookeeperConnect;
     private int kafkaExposedPort;
+    private int brokerId;
```
tombentley, line 49: This field is never used.

test-container/src/main/java/io/strimzi/test/container/StrimziKafkaContainer.java
```diff
@@ -94,13 +118,18 @@ protected void doStart() {
         super.doStart();
     }
 
+    public StrimziKafkaContainer withExternalZookeeper(String connectString) {
```
tombentley, line 121: When I made the comment here before about the factory methods I was expecting the external ZK to be one of them. Now I see that you've got three overloads for those methods. Another 3 seems a bit silly, but maybe the external ZK should just be extra parameter for a 4th factory?

test-container/src/main/java/io/strimzi/test/container/StrimziKafkaCluster.java
```diff
@@ -0,0 +1,144 @@
...
+                        "sh", "-c",
+                        "bin/zookeeper-shell.sh zookeeper:" + StrimziKafkaContainer.ZOOKEEPER_PORT + " ls /brokers/ids | tail -n 1"
+                    );
+                    String brokers = result.getStdout();
+
+                    LOGGER.info("Stdout from zookeeper container....{}", result.getStdout());
+
+                    return brokers != null && brokers.split(",").length == this.brokersNum;
+                } catch (IOException | InterruptedException e) {
+                    e.printStackTrace();
```
tombentley, line 124: Remind me again why it's OK to just ignore the `InterruptedException`? I know it's an annoying Exception to have to deal with, and I know you're not expecting to need to support cancellation here, but you never know how code will evolve and be reused. It would be better to do `Thread.currentThread().interrupt();`, that way `waitFor`'s call to `sleep()` would throw `InterruptedException` and we can declare `waitFor() throws InterruptedException`, and then the whole thing supports cancellation.

test-container/src/main/java/io/strimzi/test/container/StrimziKafkaCluster.java
see-quick: No, we don't... Thanks

tombentley (approved):

Issue #5402, opened 2021-08-06
Make Scram-Sha password length configurable


pantaoran:
**Is your feature request related to a problem? Please describe.**
I want to enable SASL authentication with Scram-Sha-512 for a group of Kafka users on my cluster. Corporate security tells me that they will only allow it if the used passwords are at least 30 characters long, but the passwords generated by Strimzi are only 12 characters long.

**Describe the solution you'd like**
Ideally I would like the generated password length to be configurable, so that I can set this length according to my needs. I don't really care if such config would be on a per user basis or a per cluster basis.

**Describe alternatives you've considered**
Another option would be to just set the hardcoded length to a bigger number for increased security. May I suggest 30 characters? 😇 

**Additional context**
None

tombentley:
_If_ we were going to make it configurable then we should also consider making the alphabet from which the characters are drawn configurable too, since I'm sure there'll be some corporate security department somewhere that has opinions about that too.

scholzj:
I think that makes sense @tombentley

PR #5415, opened 2021-08-10
Make the Scram Sha -512 password length configurable

ShubhamRwt:
### Type of change

_Select the type of your PR_

- Enhancement / new feature

### Description

This PR address the issue mentioned in #5402. This PR adds an env var to the UserOperatorConfig to configure the length for the Scram Sha password.
 
### Checklist

- [ ] Write tests
- [ ] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/user-operator/src/main/java/io/strimzi/operator/user/UserOperatorConfig.java
+++ b/user-operator/src/main/java/io/strimzi/operator/user/UserOperatorConfig.java
@@ -29,13 +29,11 @@ public class UserOperatorConfig {
     public static final String STRIMZI_SECRET_PREFIX = "STRIMZI_SECRET_PREFIX";
     public static final String STRIMZI_ACLS_ADMIN_API_SUPPORTED = "STRIMZI_ACLS_ADMIN_API_SUPPORTED";
     public static final String STRIMZI_SCRAM_SHA_PASSWORD_LENGTH = "STRIMZI_SCRAM_SHA_PASSWORD_LENGTH";
-    public static final String STRIMZI_SCRAM_SHA_PASSWORD_ALPHABET = "STRIMZI_SCRAM_SHA_PASSWORD_ALPHABET";
 
     public static final long DEFAULT_FULL_RECONCILIATION_INTERVAL_MS = 120_000;
     public static final String DEFAULT_KAFKA_BOOTSTRAP_SERVERS = "localhost:9091";
     public static final String DEFAULT_SECRET_PREFIX = "";
     public static final int DEFAULT_SCRAM_SHA_PASSWORD_LENGTH = 12;
-    public static final String DEFAULT_SCRAM_SHA_PASSWORD_ALPHABET = "abcdefghijklmnopqrstuvwxyz" + "ABCDEFGHIJKLMNOPQRSTUVWXYZ" + "0123456789";
     // Defaults to true for backwards compatibility in standalone UO deployments
     public static final boolean DEFAULT_STRIMZI_ACLS_ADMIN_API_SUPPORTED = true;
 
@@ -53,7 +51,6 @@ public class UserOperatorConfig {
     private final int clientsCaRenewalDays;
     private final boolean aclsAdminApiSupported;
     private final int scramPasswordLength;
-    private final String scramPasswordAlphabet;
 
     /**
      * Constructor
@@ -72,7 +69,6 @@ public class UserOperatorConfig {
      * @param clientsCaValidityDays Number of days for which the certificate should be valid
      * @param clientsCaRenewalDays How long before the certificate expiration should the user certificate be renewed
      * @param scramPasswordLength Length used for the Scram-Sha Password
-     * @param scramPasswordAlphabet String with characters which will be used to generate the character of the password starting with the second character
      */
     @SuppressWarnings({"checkstyle:ParameterNumber"})
     public UserOperatorConfig(String namespace,
@@ -88,8 +84,7 @@ public class UserOperatorConfig {
                               boolean aclsAdminApiSupported,
                               int clientsCaValidityDays,
                               int clientsCaRenewalDays,
-                              int scramPasswordLength,
-                              String scramPasswordAlphabet) {
+                              int scramPasswordLength) {
         this.namespace = namespace;
         this.reconciliationIntervalMs = reconciliationIntervalMs;
         this.kafkaBootstrapServers = kafkaBootstrapServers;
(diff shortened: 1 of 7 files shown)


user-operator/src/main/java/io/strimzi/operator/user/UserOperatorConfig.java
```diff
@@ -114,6 +120,12 @@ public static UserOperatorConfig fromMap(Map<String, String> map) {
             reconciliationInterval = Long.parseLong(reconciliationIntervalEnvVar);
         }
 
+        int scramPasswordLength = DEFAULT_SCRAM_SHA_PASSWORD_LENGTH;
+        String scramPasswordLengthEnvVar = map.get(UserOperatorConfig.STRIMZI_SCRAM_SHA_PASSWORD_LENGTH);
+        if (scramPasswordLengthEnvVar != null) {
+            scramPasswordLength = Integer.parseInt(scramPasswordLengthEnvVar);
+        }
```
see-quick, line 127: I think this one could be elegantly be written by a ternary operator and remove unneeded variables what do you think? Something like:
```
scramPasswordLength =  map.get(UserOperatorConfig.STRIMZI_SCRAM_SHA_PASSWORD_LENGTH) != null ? Integer.parseInt(scramPasswordLengthEnvVar) : DEFAULT_SCRAM_SHA_PASSWORD_LENGTH;
```

scholzj (approved):
LGTM. Maybe you should think about adding some tests for this.

ShubhamRwt:
Added the tests @scholzj

user-operator/src/main/java/io/strimzi/operator/user/UserOperatorConfig.java
```diff
@@ -282,6 +294,13 @@ public String getSecretPrefix() {
         return secretPrefix;
     }
 
+    /**
+     * @return  Indicates whether the Kafka Admin API for managing ACLs is supported by the Kafka cluster or not
+     */
```
tombentley, line 299: Let's fix this.

tombentley (commented):
This basically looks good, but IIRC in the issue I suggested we also support configuring the alphabet from which the characters of the password are drawn. `PasswordGenerator` already has a constructor supporting that, so it should just be another env var.

ShubhamRwt:
Hello @tombentley , Actually I was thinking of opening other PR for that due to these issues 1) As per the alphabet part in Password Generator there are two Strings firstCharacterAlphabet( String used to generate only the first character) and one is only alphabet(String used to generate rest of the password) so I guess both would require to be made configurable .......  2) We don't have any constructor where we pass only two arguments (Like for eg if we make firstCharacterAlphabet and alphabet  configurable, so what if the user wants only set the firstCharacterAlphabet part but for the rest of part it want it to be default so we will have to use a constructor like PasswordGeneratot( length, firstCharacterAlphabet) )

scholzj:
@ShubhamRwt Maybe you can also add some test to `KafkaUserOperatorTest` to check that it is actually passed through and the password generated has the correct length.

user-operator/src/main/java/io/strimzi/operator/user/UserOperatorConfig.java
ShubhamRwt: Fixed

tombentley:
@ShubhamRwt the reason for supporting a different alphabet for the first letter was https://github.com/strimzi/strimzi-kafka-operator/pull/1630. I guess `org.apache.kafka.common.security.JaasConfig` might also have problems with/need quoting for special symbols. We should investigate those constraints before trying to implement something, so I'm +1 on a separate PR.

ShubhamRwt:
@tombentley @scholzj After some good amount of searching I guess the problem of using special characters with JAAS still persists and the only workaround for this is using `single quotes` around the password as Jakub mentioned on the PR [#1630](https://github.com/strimzi/strimzi-kafka-operator/pull/1630). So we have two choices either to constraint the alphabets such that no special character can be used(If passed a special character maybe some sort of warning and then using a default password). And second would be to enclose the password in quotes

ShubhamRwt:
Someone in the apicurio register also had this sort of issue --> https://github.com/Apicurio/apicurio-registry-operator/pull/75. So they fixed it by adding quotes

ShubhamRwt:
I guess Apicurio registry also has this problem with Jaas. They resolved it using the quotes.

tombentley:
@ShubhamRwt people using strimzi would need to know to use quotes when using a password generated by the UO. It was to avoid this (which is not very discoverable due to the poor error message from Kafka when quotes are absent) that Jakub did #1630 in the first place. 

@ShubhamRwt is it possible to have a password that includes quotes (needing a quote character to do so)? I.e. Is something like this possible?

```
sasl.jaas.config=org.apache.kafka.common.security.scram.ScramLoginModule required username=my-user2 password='foobar\'baz';
```

I think it would be OK to say that when a user configures their own alphabet (to be used for all characters) they should be prepared to use quotes. We can document is. It's not a regression because the default would still avoid numbers in the first letter.

ShubhamRwt:
Yeah I totally agree with the both the points @tombentley. I will check if the above password works fine, If it works fine then I will start working on the code. Thanks

scholzj:
So, what is the plan now @ShubhamRwt @tombentley? Do we review this just for the length and leave the alphabet for separate Pr as planned originaly? Or should I wait with review for @ShubhamRwt to add the custom alphabet support?

ShubhamRwt:
For the alphabet part part I guess Tom also agree for a seperate PR. For the test, I wrote the test but it didn't worked fine, so I was looking at it just now

ShubhamRwt:
I have started with the code for Alphabet part also but I have it in the second branch now, which I will push tomorrow

user-operator/src/test/java/io/strimzi/operator/user/operator/KafkaUserOperatorTest.java
```diff
@@ -933,6 +932,89 @@ public void testReconcileNewScramShaUser(VertxTestContext context)    {
             })));
     }
 
+    @Test
+    public void testReconcileNewScramShaUserwithConfigurableLength(VertxTestContext context)    {
```
scholzj, line 936: This looks good. I wonder if it would make more sense to focus on the password only and remove the assertion of the other things such as ACLs - since we would need to maintain those parts even when the test really tests the password length. But LG otherwise.

tombentley (approved):

pantaoran:
Will this also be documented? I don't have knowledge about the process, I just see that in this PR there is no documentation update, so I'm asking.

scholzj:
You would in general do something like this:

```yaml
# ...
spec:
  # ...
  entityOperator:
    topicOperator: {}
    userOperator: {}
    template:
      userOperatorContainer:
        env:
          - name: STRIMZI_SCRAM_SHA_PASSWORD_LENGTH
             value: "20"
```

Keep in mind that this is not yet released. It will be only as part of Strimzi 0.26.0.

PR #5438, opened 2021-08-17
A sketch of how we could support cancellation

### LINKED ISSUE #4869: KafkaRoller may continue to try to roll a kafka's pod long after the kafka is deleted

tombentley:
This PR is a sketch of how we might support cancellation and thus fix #4869 based on thoughts following [this comment](https://github.com/strimzi/strimzi-kafka-operator/pull/4996#issuecomment-899558964).

It illustrates:

* Moving away from using vertx. This is done by having an executor service bound to a `ReconciliationState`. In this case we're using a new executor for each `ReconciliationState` instance, but I don't think that's really where we'd end up. Rather we'd likely use an executor service with a pool of threads. 
* I've changed the `ZookeeperScaler` to not use vertx, and also the places where we were using `waitFor` for readiness etc use a busy loop with calls to `sleep`. The main thing about these changes is we allow `InterruptedException` to propagate all the way out to the executor.  It should be possible to use the executor service and schedule a task to check for the poll condition, but using sleep was quicker for now. 
* `Executor.submit` returns a `java.util.concurrent.Future` which supports cancellation. Calling `cancel(true)` on such a future (with a util.concurrent `ExecutorService`) will interrupt any thread that's running the task corresponding to that Future. 
* Using the `Cancellable` abstraction, we can cancel reconciliations for individual CRs. If they're running at the time of cancellation then the Future is used to cancel the task. Otherwise we just mark the task cancelled and it won't be possible to `run` any task for it. I've not yet hooked to the `KAO.cancel` to the watcher. In reality some of this really belongs in the `AbstractOperator`.

There's a whole bunch of stuff still to decide including:
* should we sleep or use the `ScheduledExecutorService` when awaiting conditions?
* We'd need to refactor KafkaRoller to use the common `ScheduledExecutorService`
* Is `cancellations` keyed on `metadata.uid` and only used for CR deletion?
* Is removing the dependency on vertx something we have/ought to do as part of this? Or are we prepared to accept the complexity of a hybrid model?
* Can we actually use a pool of threads? If we're not dropping vertx this assumes we have the resources for a bunch more threads.

But hopefully this shows enough of the moving parts to stimulate some discussion.

### DIFF
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/Cancellable.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/Cancellable.java
@@ -1,0 +1,42 @@
+/*
+ * Copyright Strimzi authors.
+ * License: Apache License 2.0 (see the file LICENSE or http://apache.org/licenses/LICENSE-2.0.html).
+ */
+package io.strimzi.operator.cluster.operator.assembly;
+
+interface Cancellable {
+    boolean isCancelled();
+
+    void cancel();
+
+    static Cancellable atomic() {
+        return new Cancellable() {
+
+            private volatile boolean cancelled = false;
+
+            @Override
+            public boolean isCancelled() {
+                return cancelled;
+            }
+
+            @Override
+            public void cancel() {
+                cancelled = true;
+            }
+        };
+    }
+
+    static Cancellable future(java.util.concurrent.Future<?> future) {
+        return new Cancellable() {
+            @Override
+            public boolean isCancelled() {
+                return future.isCancelled();
+            }
+
+            @Override
+            public void cancel() {
+                future.cancel(true);
+            }
+        };
+    }
+}
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaAssemblyOperator.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaAssemblyOperator.java
@@ -92,6 +92,7 @@ import io.strimzi.operator.common.Reconciliation;
 import io.strimzi.operator.common.ReconciliationException;
 import io.strimzi.operator.common.Util;
 import io.strimzi.operator.common.model.Labels;
+import io.strimzi.operator.common.operator.resource.AbstractReadyResourceOperator;
 import io.strimzi.operator.common.operator.resource.AbstractScalableResourceOperator;
 import io.strimzi.operator.common.operator.resource.CrdOperator;
 import io.strimzi.operator.common.operator.resource.DeploymentOperator;
@@ -131,6 +132,9 @@ import java.util.Locale;
 import java.util.Map;
 import java.util.Set;
 import java.util.TimeZone;
+import java.util.concurrent.Callable;
+import java.util.concurrent.Executors;
+import java.util.concurrent.ScheduledExecutorService;
 import java.util.function.Function;
 import java.util.TreeSet;
 import java.util.concurrent.ExecutionException;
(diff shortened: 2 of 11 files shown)


tombentley:
> Rather we'd likely use an executor service with a pool of threads.
> ...
> Can we actually use a pool of threads? If we're not dropping vertx this assumes we have the resources for a bunch more threads.

I think we _have_ to use some kind of shared thread pool in order to get something that uses resources (threads) about as efficiently as vertx. 

> should we sleep or use the ScheduledExecutorService when awaiting conditions?

If we're using a thread pool then we must use the latter.

> Is removing the dependency on vertx something we have/ought to do as part of this? Or are we prepared to accept the complexity of a hybrid model?

The ARO hierarchy only needs vertx for two things: i) Scheduling work in the future (basically to pass to `waitFor`) and ii) Running blocking methods on worker threads. We could replace the direct dependency with one on an interface and implement that interface using a `ScheduledExecutorService`. So replacing vertx in these parts maybe isn't so complicated. There are places where we use other parts of vertx, like the HTTP server and client. But maybe those are isolated enough that we'd not need to change those parts.

tombentley:
I'm closing this because I don't think we really move to a thread-per-cluster model while those threads are native threads. That will change if/when Project Loom happens on the JVM, but until then it's easiest to stick with vertx.

Issue #4014, opened 2020-11-29
defaultImageRepository instead of imageRepositoryOverride in helm chart


wtayyeb:
Dear strimzi team

In the operator helm chart, there is `imageRepositoryOverride` and `imageTagOverride` in values.yaml
and if one set them, he could not change `STRIMZI_KAFKA_xx_IMAGES` as they are Override like their name.

but let think this way, if one need to define private/cache registry for all images, and set `kafkaConnect` image to a custom repository, above variables will prevent him from that custom image. so he must set all .Values.xx.image.repository to private/cache registry and also set that custom image for kafkaconnect.

I have read and install several charts (>20) and there is a pattern letting you to define defaultRepository or defaultRegistry, to use them as default value not Override value, and the .Values.xx.image.repository is the final value and not the default value.

for example following line (from _kafka_image_map.tpl)
```
2.5.0={{ default .Values.kafka.image.repository .Values.imageRepositoryOverride }}/{{ .Values.kafka.image.name }}:{{ default .Values.kafka.image.tagPrefix .Values.imageTagOverride }}-kafka-2.5.0
```
will be something like this:
```
2.5.0={{ default .Values.defaultImageRepository .Values.kafka.image.repository }}/{{ .Values.kafka.image.name }}:{{ default .Values.defaultImageTag .Values.kafka.image.tagPrefix }}-kafka-2.5.0
```
and empty all values like .Values.xx.image.repository, .Values.xx.image.tag to let user set them.

scholzj:
That sounds like it makes sense. Can you open a PR for it then?

ShubhamRwt:
@wtayyeb Hello, do you plan to create a PR for this?

PR #5440, opened 2021-08-17
defaultImageRepository instead of imageOverrideRepository

ShubhamRwt:
### Type of change

_Select the type of your PR_

- Enhancement / new feature
- Refactoring

### Description

This PR fixes the issue mentioned in #4014.
Earlier we were making use of imageOverrideTag, imageOverrideRepository and imageOverrideRegistry. But the problem with this approach was if anyone set them then it will override the tags, registry and repository for all the images ie.all the images will now use same tags, registry, and repository. If someone want to set the registry, repository or tag for a single image he can't do that since all the override will apply to all the image. This PR resolves this issue by introducing defaultImageRepository, defaultImageRegistry, defaultImageTag which points to default repo, registry and tags. If the user wants wants to set these parameters for a single image now they can do it easily.

### Checklist

- [ ] Write tests
- [ ] Update documentation
- [ ] Update CHANGELOG.md


packaging/helm-charts/helm3/strimzi-kafka-operator/values.yaml
```diff
@@ -32,94 +36,94 @@ securityContext: {}
...
-    registry: quay.io
-    repository: strimzi
+    registry: ""
+    repository: ""
     name: kafka
-    tagPrefix: latest
+    tagPrefix: ""
 kanikoExecutor:
   image:
     registry: quay.io
```
scholzj, line 123: Should this still be here?

scholzj (commented):
One nit. Also, I think this deserves mentioning in the CHANGELOG _Changes, deprecations and removals_ section (https://github.com/strimzi/strimzi-kafka-operator/blob/main/CHANGELOG.md#changes-deprecations-and-removals)

scholzj (commented):
I think you will also need to update the release tooling which was updating the `tag` / `tagPrefix` fields and now will need to update `defaultImageTag`. I think you also need to update the README file (including the related release tooling).

scholzj (approved):

sknot-rh (approved):
This will cause conflicts in https://github.com/strimzi/strimzi-kafka-operator/pull/5407
I am happy with fixing them, so this can be merged.

PR #5444, opened 2021-08-17
fix(cluster-operator): correct auto-detection of KUBERNETES_SERVICE_DNS_DOMAIN

moustafab:
### Type of change
- Bugfix

### Description

The auto-detection script fails incorrectly when a match is not found and instead of defaulting to the correct `cluster.local` if it doesn't find a cluster domain name, it actually uses all the entries from `getent hosts kubernetes.default`

### Checklist

- [ ] Write tests
- [ ] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/cluster-operator/scripts/cluster_operator_run.sh
+++ b/cluster-operator/scripts/cluster_operator_run.sh
@@ -3,7 +3,7 @@ export JAVA_CLASSPATH=lib/io.strimzi.@project.build.finalName@.@project.packagin
 export JAVA_MAIN=io.strimzi.operator.cluster.Main
 
 if [ -z "$KUBERNETES_SERVICE_DNS_DOMAIN" ]; then
-  KUBERNETES_SERVICE_DNS_DOMAIN=$(getent hosts kubernetes.default | head -1 | sed "s/.*\skubernetes.default.svc.//")
+  KUBERNETES_SERVICE_DNS_DOMAIN=$(getent hosts kubernetes.default | head -1 | sed "s/.*\skubernetes.default.svc//" | sed "s/\.//")
   if [ -n "$KUBERNETES_SERVICE_DNS_DOMAIN" ]; then
     echo "Auto-detected KUBERNETES_SERVICE_DNS_DOMAIN: $KUBERNETES_SERVICE_DNS_DOMAIN"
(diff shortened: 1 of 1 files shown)


scholzj (commented):
The old code returns `cluster.local`. The new code returns `.cluster.local`. That has to show up somewhere as a problem, or?

moustafab:
> The old code returns `cluster.local`. The new code returns `.cluster.local`. That has to show up somewhere as a problem, or?

Yea you're correct, I added an additional `sed` statement to handle that case properly. I think using `getent` is a bit of a stretch for autodetection, as it isn't a reliable assumption however it isn't such a big issue because users can override the variable. I have this auto-detection failing inside of a k3s cluster which is part of a [vcluster](https://www.vcluster.com/docs/what-are-virtual-clusters).

scholzj (approved):
I guess this looks good. Do you have some example where this works better than the previous version? To be able to better understand it but also out of curiosity ... I mean as an example of what `getent` returns in that case.

scholzj:
> Yea you're correct, I added an additional sed statement to handle that case properly. I think using getent is a bit of a stretch for autodetection

It used to be done with nslookup. But that requires some additional packages to be installed which were often source of CVEs. So that is why we use `getent`

moustafab:
> I guess this looks good. Do you have some example where this works better than the previous version? To be able to better understand it but also out of curiosity ... I mean as an example of what `getent` returns in that case.

From a strimzi pod running inside of a vcluster:
```
$ getent hosts kubernetes.default | head -1 | sed "s/.*\skubernetes.default.svc.//"
100.66.163.43   kubernetes kubernetes.default kubernetes.default.svc
```

scholzj:
@moustafab Yeah, but what is the original value of `getent hosts kubernetes.default`?

moustafab:
> @moustafab Yeah, but what is the original value returned by `getent hosts kubernetes.default`?
@scholzj It is the same because the regex for the sed command doesn't match
```
$ getent hosts kubernetes.default
100.66.163.43   kubernetes kubernetes.default kubernetes.default.svc
```

scholzj:
Ahh, ok. Understand. Thanks.

ppatierno (approved):

scholzj:
Thanks for the PR

PR #5451, opened 2021-08-18
Add custom labels and annotations for JMX secrets

OuesFa:
Signed-off-by: faresoueslati <fares.oueslati@blablacar.com>

### Type of change

- Enhancement / new feature

### Description

Allowing the user to annotate and add labels to the secret containing the JMX credentials generated by the operator. Like what it's already done for the secret containing the cluster ca certificate for instance.

### Checklist

- [x] Write tests
- [x] Update documentation
- [x] Update CHANGELOG.md

### DIFF
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/model/AbstractModel.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/model/AbstractModel.java
@@ -286,6 +286,8 @@ public abstract class AbstractModel {
     protected Map<String, String> templateClusterRoleBindingAnnotations;
     protected Map<String, String> templateServiceAccountLabels;
     protected Map<String, String> templateServiceAccountAnnotations;
+    protected Map<String, String> templateJmxSecretLabels;
+    protected Map<String, String> templateJmxSecretAnnotations;
 
     protected List<Condition> warningConditions = new ArrayList<>(0);
 
@@ -923,8 +925,8 @@ public abstract class AbstractModel {
                 .build();
     }
 
-    protected Secret createSecret(String name, Map<String, String> data) {
-        return ModelUtils.createSecret(name, namespace, labels, createOwnerReference(), data);
+    protected Secret createSecret(String name, Map<String, String> data, Map<String, String> customAnnotations, Map<String, String> customLabels) {
+        return ModelUtils.createSecret(name, namespace, labels, createOwnerReference(), data, customAnnotations, customLabels);
     }
 
     protected Service createService(String type, List<ServicePort> ports, Map<String, String> annotations) {
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/model/AbstractModel.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/model/AbstractModel.java
@@ -101,6 +101,8 @@ import java.util.Optional;
 import java.util.Set;
 import java.util.stream.Collectors;
 
+import static java.util.Collections.emptyMap;
+
 /**
  * AbstractModel an abstract base model for all components of the {@code Kafka} custom resource
  */
@@ -354,7 +356,7 @@ public abstract class AbstractModel {
      * @return The selector labels as an instance of the Labels object.
      */
     public Labels getSelectorLabels() {
-        return getLabelsWithStrimziName(name, Collections.emptyMap()).strimziSelectorLabels();
+        return getLabelsWithStrimziName(name, emptyMap()).strimziSelectorLabels();
     }
 
     /**
@@ -925,7 +927,11 @@ public abstract class AbstractModel {
                 .build();
     }
 
-    protected Secret createSecret(String name, Map<String, String> data, Map<String, String> customAnnotations, Map<String, String> customLabels) {
+    protected Secret createSecret(String name, Map<String, String> data) {
+        return ModelUtils.createSecret(name, namespace, labels, createOwnerReference(), data, emptyMap(), emptyMap());
+    }
+
+    protected Secret createJmxSecret(String name, Map<String, String> data, Map<String, String> customAnnotations, Map<String, String> customLabels) {
         return ModelUtils.createSecret(name, namespace, labels, createOwnerReference(), data, customAnnotations, customLabels);
     }
 
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/model/AbstractModel.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/model/AbstractModel.java
@@ -931,8 +931,8 @@ public abstract class AbstractModel {
         return ModelUtils.createSecret(name, namespace, labels, createOwnerReference(), data, emptyMap(), emptyMap());
     }
 
-    protected Secret createJmxSecret(String name, Map<String, String> data, Map<String, String> customAnnotations, Map<String, String> customLabels) {
-        return ModelUtils.createSecret(name, namespace, labels, createOwnerReference(), data, customAnnotations, customLabels);
+    protected Secret createJmxSecret(String name, Map<String, String> data) {
+        return ModelUtils.createSecret(name, namespace, labels, createOwnerReference(), data, templateJmxSecretAnnotations, templateJmxSecretLabels);
     }
 
     protected Service createService(String type, List<ServicePort> ports, Map<String, String> annotations) {
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaCluster.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaCluster.java
@@ -1329,7 +1329,7 @@ public class KafkaCluster extends AbstractModel {
             data.put(key, Base64.getEncoder().encodeToString(passwordGenerator.generate().getBytes(StandardCharsets.US_ASCII)));
         }
 
-        return createJmxSecret(KafkaCluster.jmxSecretName(cluster), data, templateJmxSecretAnnotations, templateJmxSecretLabels);
+        return createJmxSecret(KafkaCluster.jmxSecretName(cluster), data);
     }
 
     private List<ContainerPort> getContainerPortList() {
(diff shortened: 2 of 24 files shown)


cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaAssemblyOperator.java
```diff
@@ -1230,11 +1230,24 @@ protected CompositeFuture adminClientSecrets() {
...
+            }
             if (zkCluster.isJmxAuthenticated()) {
                 Future<Secret> secretFuture = secretOperations.getAsync(namespace, ZookeeperCluster.jmxSecretName(name));
+                Map<String, String> finalJmxSecretAnnotations = jmxSecretAnnotations;
+                Map<String, String> finalJmxSecretLabels = jmxSecretLabels;
                 return secretFuture.compose(secret -> {
                     if (secret == null) {
-                        return withVoid(secretOperations.reconcile(reconciliation, namespace, ZookeeperCluster.jmxSecretName(name), zkCluster.generateJmxSecret()));
+                        return withVoid(secretOperations.reconcile(reconciliation, namespace, ZookeeperCluster.jmxSecretName(name),
+                                zkCluster.generateJmxSecret(finalJmxSecretAnnotations, finalJmxSecretLabels)));
```
scholzj, line 1250: I don't think this should be done in the `KafkaAssemblyOperator` class. This should be all encapsulated by the model classes (`KafkaCluster`, ZooKeeperCluster`, `AbstractModel` etc.). Same for the Kafka secret below. These classes already handle the template properties for many other resources, so you should be able to find there how it is done and used.

cluster-operator/src/test/java/io/strimzi/operator/cluster/model/ZookeeperClusterTest.java
```diff
@@ -273,12 +274,52 @@ public void testCreateClusterWithZookeeperJmxEnabled() {
...
+                .withNewJmxSecret()
+                .withNewMetadata()
+                .withAnnotations(customAnnotations)
+                .withLabels(customLabels)
+                .endMetadata()
+                .endJmxSecret()
+                .endTemplate()
+                .endZookeeper()
+                .endSpec()
+                .build();
```
scholzj, line 310: The code using the builder (here and in all other places) ismuch more readable when you use indentation to separate the different `withNew...` and `end...` calls. Have a look at the test below this one for example.

documentation/modules/security/ref-certificates-and-secrets.adoc
```diff
@@ -190,6 +190,46 @@ spec:
...
+        metadata:
+          labels:
+            label1: value1
+            label2: value2
+          annotations:
+            annotation1: value1
+            annotation2: value2
+    # ...
+----
+
```
scholzj, line 232: Do we really want a dedicated documentation and example for this? Most of the other template is IMHO not covered to such detail. It increases the maintenance effort and I'm not sure it really adds value. Just think about it.

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaCluster.java
```diff
@@ -1314,17 +1315,29 @@ public Secret generateBrokersSecret() {
...
+                .withNewMetadata()
+                .withName(KafkaCluster.jmxSecretName(cluster))
+                .withOwnerReferences(createOwnerReference())
+                .withNamespace(namespace)
+                .withLabels(Util.mergeLabelsOrAnnotations(labels.toMap(), jmxSecretLabels))
+                .withAnnotations(jmxSecretAnnotations)
+                .endMetadata()
+                .withType("Opaque")
+                .withData(data)
+                .build();
```
scholzj, line 1340: Wouldn't it be possible to extend the `createSecret` method with some additional optional parameters? This duplicates a code which will be hard to maintain.

scholzj (commented):
I left some comments. Ideally, this should be done also for Kafka Connect (and KafkaMirrorMaker2 based on it I guess).

documentation/modules/security/ref-certificates-and-secrets.adoc
```diff
@@ -190,7 +190,7 @@ spec:
     # ...
 ----
 
-For more information on configuring template properties, see xref:assembly-customizing-kubernetes-resources-str[].
+For more information on configuring template properties, see xref:assembly-customizing-kubernetes-resources-str[].== Adding labels and annotations to Secrets
```
scholzj, line 193: Is this change intentional? This seems like it adds a title at the end of a sentence ...

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/AbstractModel.java
```diff
@@ -923,8 +925,8 @@ protected ConfigMap createConfigMap(String name, Map<String, String> data) {
                 .build();
     }
 
-    protected Secret createSecret(String name, Map<String, String> data) {
-        return ModelUtils.createSecret(name, namespace, labels, createOwnerReference(), data);
+    protected Secret createSecret(String name, Map<String, String> data, Map<String, String> customAnnotations, Map<String, String> customLabels) {
+        return ModelUtils.createSecret(name, namespace, labels, createOwnerReference(), data, customAnnotations, customLabels);
```
scholzj, line 929: I wonder if we can do this part a bit better ... 
* You defined here the `templateJmxSecretLabels` and `templateJmxSecretAnnotations`
* Maybe you can add a new method `createJmxSecret` which will automatically call `ModelUtils.createSecret(name, namespace, labels, createOwnerReference(), data, templateJmxSecretAnnotations, templateJmxSecretLabels)`
* Keep the `createSecret` method signature as it was to reduce the amount of changes where you just call it with `emptyMap()` anyway.

packaging/install/cluster-operator/040-Crd-kafka.yaml
```diff
@@ -2039,6 +2039,27 @@ spec:
...
+                        properties:
+                          metadata:
+                            type: object
+                            properties:
+                              labels:
+                                x-kubernetes-preserve-unknown-fields: true
+                                type: object
+                                description: Labels added to the resource template.
+                                  Can be applied to different resources such as `StatefulSets`,
+                                  `Deployments`, `Pods`, and `Services`.
```
scholzj, line 2053: In all the CRDs, you seem to have the descriptions in Helm Chart as single line and in install as multiline. Did you changed that manually? Or is this how it was generated by the make commands? If you changed it manually, it would fail in the build, the files need to be as they were generated. (if it was generated automatically like this, then please just ignore this comment)

scholzj (commented):
Some nits, but looks mostly good.

tombentley (approved):
No comments additional to Jakub's. Thanks!

documentation/modules/security/ref-certificates-and-secrets.adoc
OuesFa: No it is not I'm removing it!

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/ModelUtils.java
```diff
@@ -188,16 +190,18 @@ public static Secret buildSecret(Reconciliation reconciliation, ClusterCa cluste
...
+    public static Secret createSecret(String name, String namespace, Labels labels, OwnerReference ownerReference,
+                                      Map<String, String> data, Map<String, String> customAnnotations, Map<String, String> customLabels) {
         if (ownerReference == null) {
             return new SecretBuilder()
                     .withNewMetadata()
                         .withName(name)
                         .withNamespace(namespace)
-                        .withLabels(labels.toMap())
+                        .withLabels(Util.mergeLabelsOrAnnotations(labels.toMap(), customLabels))
+                    .withAnnotations(customAnnotations)
```
sknot-rh, line 204: Can you please indent `withAnnotations`? If you will implement Jakub's suggestion about `createJmxSecret`, ignore this comment.

sknot-rh (approved):
One nit. Thank you for the PR :)

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/ModelUtils.java
OuesFa: Maybe I misunderstood something but even following Jakub's suggestion we gonna still have this line in this method right ?

scholzj: Yeah, m idea to have the different method in `AbstractModel`, but still call from there this method. So I think this should stay, yes.

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/AbstractModel.java
```diff
@@ -924,7 +928,11 @@ protected ConfigMap createConfigMap(String name, Map<String, String> data) {
     }
 
     protected Secret createSecret(String name, Map<String, String> data) {
-        return ModelUtils.createSecret(name, namespace, labels, createOwnerReference(), data);
+        return ModelUtils.createSecret(name, namespace, labels, createOwnerReference(), data, emptyMap(), emptyMap());
+    }
+
+    protected Secret createJmxSecret(String name, Map<String, String> data, Map<String, String> customAnnotations, Map<String, String> customLabels) {
+        return ModelUtils.createSecret(name, namespace, labels, createOwnerReference(), data, customAnnotations, customLabels);
```
scholzj, line 935: I don't think you need the `Map<String, String> customAnnotations, Map<String, String> customLabels` here, or? Can't you do just this?

```
    protected Secret createJmxSecret(String name, Map<String, String> data) {
        return ModelUtils.createSecret(name, namespace, labels, createOwnerReference(), data, templateJmxSecretAnnotations, templateJmxSecretLabels);
```

scholzj (commented):
I left one more comment (+ the doc issue is still there), LGTM otherwise.

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/AbstractModel.java
OuesFa: yes much more elegant !

scholzj (approved):

scholzj:
Thanks for the PR @OuesFa

OuesFa:
> Thanks for the PR @OuesFa

Thanks for your quick and helpful review 🙏

PR #5494, opened 2021-08-30
Fix incompatible types issue when using OpenJ9 JVM

kun-lu20:
1. Replace withNew*(String) methods with with*(String)
2. Replace addNew*(String) methods with addTo*s(String)

### Type of change

_Select the type of your PR_

- Bugfix

### Description

This PR is to fix the incompatible types issue raised when compiling the project with AdoptOpenJDK 11 (Eclipse OpenJ9 JVM), so that this project will be successfully compiled and built across different JVMs. 

The root cause of this issue is that when using OpenJ9 JVM to compile, only `withNew*(byte[])` methods will be generated in `Sundrio` produced classes such as `*Builder` classes, while `withNew*(String)` methods are expected in the code. Furthermore, the overloaded` withNew*()` methods are deprecated in recent versions and we could use `with*(String)` methods instead which would not cause the issue. The situation of `addNew*(String)` methods is similar.

To address this issue,  deprecated `withNew*(String)` methods are replaced with `with*(String)` methods and deprecated `addNew*(String)` methods are replaced with `addTo*s(String)` methods.

### Checklist

- [ ] Write tests
- [ ] Update documentation
- [ ] Update CHANGELOG.md


scholzj (approved):

scholzj (approved):
Thanks for the PR. I think it looks good assuming all the tests pass etc.

kun-lu20:
Thanks @scholzj !

Frawless (approved):

scholzj:
Thanks for the PR.

kun-lu20:
@Frawless @scholzj Thanks!

PR #5525, opened 2021-09-03
Merging waitForObserved and readiness method in DeploymentOperator

ShubhamRwt:

### Type of change

_Select the type of your PR_

- Refactoring

### Description

This PR introduces a new method deploymentReadiness which calls `waitforObserved` method first and then call the readiness method of the parent. This is done since both the `waitforObserved` method and `readiness` method were called in tandem.

### Checklist

- [ ] Write tests
- [ ] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/cluster-operator/src/test/java/io/strimzi/operator/cluster/operator/assembly/KafkaAssemblyOperatorTest.java
+++ b/cluster-operator/src/test/java/io/strimzi/operator/cluster/operator/assembly/KafkaAssemblyOperatorTest.java
@@ -627,9 +627,9 @@ public class KafkaAssemblyOperatorTest {
         when(mockDepOps.getAsync(anyString(), anyString())).thenReturn(
                 Future.succeededFuture()
         );
-        when(mockDepOps.waitForObserved(any(), anyString(), anyString(), anyLong(), anyLong())).thenReturn(
-                Future.succeededFuture()
-        );
+//        when(mockDepOps.waitForObserved(any(), anyString(), anyString(), anyLong(), anyLong())).thenReturn(
+//                Future.succeededFuture()
+//        );
         when(mockDepOps.readiness(any(), anyString(), anyString(), anyLong(), anyLong())).thenReturn(
                 Future.succeededFuture()
         );
@@ -1163,9 +1163,6 @@ public class KafkaAssemblyOperatorTest {
             when(mockDepOps.getAsync(clusterNamespace, EntityOperator.entityOperatorName(clusterName))).thenReturn(
                     Future.succeededFuture(originalEntityOperator.generateDeployment(true, Map.of(), null, null))
             );
-            when(mockDepOps.waitForObserved(any(), anyString(), anyString(), anyLong(), anyLong())).thenReturn(
-                    Future.succeededFuture()
-            );
             when(mockDepOps.readiness(any(), anyString(), anyString(), anyLong(), anyLong())).thenReturn(
                     Future.succeededFuture()
             );
@@ -1178,9 +1175,6 @@ public class KafkaAssemblyOperatorTest {
             when(mockDepOps.getAsync(clusterNamespace, EntityOperator.entityOperatorName(clusterName))).thenReturn(
                     Future.succeededFuture(originalCruiseControl.generateDeployment(true, Map.of(), null, null))
             );
-            when(mockDepOps.waitForObserved(any(), anyString(), anyString(), anyLong(), anyLong())).thenReturn(
-                    Future.succeededFuture()
-            );
             when(mockDepOps.readiness(any(), anyString(), anyString(), anyLong(), anyLong())).thenReturn(
                     Future.succeededFuture()
             );
@@ -1192,9 +1186,6 @@ public class KafkaAssemblyOperatorTest {
             );
             when(mockDepOps.getAsync(clusterNamespace, KafkaExporter.kafkaExporterName(clusterName))).thenReturn(
                     Future.succeededFuture(originalKafkaExporter.generateDeployment(true, null, null))
-            );
-            when(mockDepOps.waitForObserved(any(), anyString(), anyString(), anyLong(), anyLong())).thenReturn(
-                    Future.succeededFuture()
             );
             when(mockDepOps.readiness(any(), anyString(), anyString(), anyLong(), anyLong())).thenReturn(
                     Future.succeededFuture()
--- a/operator-common/src/main/java/io/strimzi/operator/common/operator/resource/DeploymentOperator.java
+++ b/operator-common/src/main/java/io/strimzi/operator/common/operator/resource/DeploymentOperator.java
@@ -105,7 +105,7 @@ public class DeploymentOperator extends AbstractScalableResourceOperator<Kuberne
         return waitFor(reconciliation, namespace, name, "observed", pollIntervalMs, timeoutMs, this::isObserved);
     }
 
-    public Future<Void> deploymentReadiness(Reconciliation reconciliation, String namespace, String name, long pollIntervalMs, long timeoutMs) {
+    public Future<Void> readiness(Reconciliation reconciliation, String namespace, String name, long pollIntervalMs, long timeoutMs) {
         return getAsync(namespace, name)
                 .compose(observe -> waitForObserved(reconciliation, namespace, name, pollIntervalMs, timeoutMs)
                 .compose(ready -> super.readiness(reconciliation, namespace, name, pollIntervalMs, timeoutMs)));
--- a/operator-common/src/main/java/io/strimzi/operator/common/operator/resource/DeploymentOperator.java
+++ b/operator-common/src/main/java/io/strimzi/operator/common/operator/resource/DeploymentOperator.java
@@ -66,7 +66,7 @@ public class DeploymentOperator extends AbstractScalableResourceOperator<Kuberne
     public Future<Void> rollingUpdate(Reconciliation reconciliation, String namespace, String name, long operationTimeoutMs) {
         return getAsync(namespace, name)
                 .compose(deployment -> deletePod(reconciliation, namespace, name))
-                .compose(ignored -> readiness(reconciliation, namespace, name, 1_000, operationTimeoutMs));
+                .compose(ignored -> super.readiness(reconciliation, namespace, name, 1_000, operationTimeoutMs));
     }
 
     /**
@@ -103,6 +103,12 @@ public class DeploymentOperator extends AbstractScalableResourceOperator<Kuberne
      */
     public Future<Void> waitForObserved(Reconciliation reconciliation, String namespace, String name, long pollIntervalMs, long timeoutMs) {
         return waitFor(reconciliation, namespace, name, "observed", pollIntervalMs, timeoutMs, this::isObserved);
+    }
+
+    public Future<Void> deploymentReadiness(Reconciliation reconciliation, String namespace, String name, long pollIntervalMs, long timeoutMs) {
+        return getAsync(namespace, name)
+                .compose(observe -> waitForObserved(reconciliation, namespace, name, pollIntervalMs, timeoutMs)
+                .compose(ready -> super.readiness(reconciliation, namespace, name, pollIntervalMs, timeoutMs)));
     }
 
     /**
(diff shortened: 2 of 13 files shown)


operator-common/src/main/java/io/strimzi/operator/common/operator/resource/DeploymentOperator.java
```diff
@@ -105,6 +105,12 @@ protected Integer currentScale(String namespace, String name) {
         return waitFor(reconciliation, namespace, name, "observed", pollIntervalMs, timeoutMs, this::isObserved);
     }
 
+    public Future<Void> deploymentReadiness(Reconciliation reconciliation, String namespace, String name, long pollIntervalMs, long timeoutMs) {
```
scholzj, line 108: Why don't you overwide the `readiness(...)` method here?

scholzj: That way i would also be covered by whatever tests there already are.

ShubhamRwt: This method calls both waitforobserve and readiness so, would it be okay to remove the part of waitforObserve from the tests
?

scholzj: I'm not sure I follow the question. The original idea was that the `waitForObserved` should be called from the regular readiness methods.

ShubhamRwt: Earlier `waitForObserve` method and `readiness` both of them were different right, but now since Iam calling `waitforObserve` from inside of `readiness` method. Its working fine for the Assembly Operator classes but I was asking that we are also using mocking in the test for both `readiness` and `waitforObserve` so if we call readiness, It also calls `waitforObserve` and again we also have a seperate mock for `waitforObserve`. for eg this--> ```        when(mockDepOps.waitForObserved(any(), anyString(), anyString(), anyLong(), anyLong())).thenReturn(
                Future.succeededFuture()
        );
        when(mockDepOps.readiness(any(), anyString(), anyString(), anyLong(), anyLong())).thenReturn(
                Future.succeededFuture()
        );```

ShubhamRwt: So here readiness will make a call to `waitForObserve` so is it okay to keep the above mocked `waitforObserve`?

scholzj: But you didn't added it to `readiness` ... you created a brand new method `deploymentReadiness`.

scholzj: Is mocked `waitForObserve` still needed? If yes, then it is ok to keep it. If it si not needed anymore, then don't keep it.

ShubhamRwt: > But you didn't added it to `readiness` ... you created a brand new method `deploymentReadiness`.

Actually I overrided `readiness` earlier but was a bit cinfused and didnt wanted to touch the test part much, that why I created a new method

ShubhamRwt: I will override the readiness method now then see if the test are passing perfectly or not

ShubhamRwt: @scholzj Overrriding the readiness method will also cause some problems since it is meddling with the DeploymenOperatorTest.

ShubhamRwt: I guess the error comes because of the `waitforObserve` call.

scholzj: Then you might need to fix the text 😉  ... you can override it to add more mocks if needed etc.

cluster-operator/src/test/java/io/strimzi/operator/cluster/operator/assembly/KafkaAssemblyOperatorTest.java
```diff
@@ -627,9 +627,9 @@ private void createCluster(VertxTestContext context, Kafka kafka, List<Secret> s
         when(mockDepOps.getAsync(anyString(), anyString())).thenReturn(
                 Future.succeededFuture()
         );
-        when(mockDepOps.waitForObserved(any(), anyString(), anyString(), anyLong(), anyLong())).thenReturn(
-                Future.succeededFuture()
-        );
+//        when(mockDepOps.waitForObserved(any(), anyString(), anyString(), anyLong(), anyLong())).thenReturn(
+//                Future.succeededFuture()
+//        );
```
tombentley, line 632: Just remove it, rather than commenting out code.

operator-common/src/main/java/io/strimzi/operator/common/operator/resource/DeploymentOperator.java
```diff
@@ -105,6 +105,12 @@ protected Integer currentScale(String namespace, String name) {
         return waitFor(reconciliation, namespace, name, "observed", pollIntervalMs, timeoutMs, this::isObserved);
     }
 
+    public Future<Void> readiness(Reconciliation reconciliation, String namespace, String name, long pollIntervalMs, long timeoutMs) {
```
tombentley, line 108: `AbstractReadyResourceOperator#isReady` has the javadoc `Check if a resource is in the Ready state.`, so the implied contract for `AbstractReadyResourceOperator#readiness` is "Wait until the resource is in the Ready state.". By overriding readiness here we're implicitly changing that contract for Deployments to be something slightly different, i.e. `status==Ready && metadata.generation==status.observedGeneration`. That could be confusing. 

I'm not sure if there are any places where we might want to wait for `status==Ready` but not `metadata.generation==status.observedGeneration`. If so, perhaps we should make the `&&` version the default semantic for `readiness()` (possibly moving `waitForObserved` up the class hierarchy, though I'm not sure all of the subclasses of ARO have a `status.observedGeneration`)?

In any case, I think we should try to define the contract for readiness as part of this PR.

operator-common/src/main/java/io/strimzi/operator/common/operator/resource/DeploymentOperator.java
```diff
@@ -105,6 +105,12 @@ protected Integer currentScale(String namespace, String name) {
...
+        return getAsync(namespace, name)
+                .compose(observe -> waitForObserved(reconciliation, namespace, name, pollIntervalMs, timeoutMs)
```
tombentley, line 110: If we're not using `observe` why the need for `getAsync`?

operator-common/src/main/java/io/strimzi/operator/common/operator/resource/DeploymentOperator.java
```diff
@@ -105,6 +105,12 @@ protected Integer currentScale(String namespace, String name) {
...
+                .compose(ready -> super.readiness(reconciliation, namespace, name, pollIntervalMs, timeoutMs)));
```
tombentley, line 111: Rather than composing both `waitForObserved` and `readiness` we could just `waitFor(isReady && isObserved)`, thus avoiding needing a `Future`.

operator-common/src/main/java/io/strimzi/operator/common/operator/resource/DeploymentOperator.java
scholzj: I think this is unique to Deployments as it controls the Deployment rollout basically.

operator-common/src/main/java/io/strimzi/operator/common/operator/resource/DeploymentOperator.java
scholzj: I'm not sure. We want to check if it is ready after it was rolled out. I guess in theory it should work the way you suggest, but should be tested on a real deployment.

ShubhamRwt: @tombentley Hi , I didn't quite got the complete gist of this `waitFor(isReady && isObserved)`

operator-common/src/main/java/io/strimzi/operator/common/operator/resource/DeploymentOperator.java
ShubhamRwt: So @tombentley basically you mean that we should have the waitForObserve() method from inside of readiness in AbstractReadyResourceOperator? Since I guess

ShubhamRwt: But yeah at the same time I guess `waitforObserve` / metadata.generation==status.observedGeneration  is common for deployment ?

operator-common/src/main/java/io/strimzi/operator/common/operator/resource/DeploymentOperator.java
tombentley: @ShubhamRwt look at the impl of both`waitForObserved` and `super.readiness` and you'll see they both poll the apiserver, using a method called `waitFor()`, waiting until some condition on the resource is satisfied. It should be equivalent to test for the conjunction of both conditions using a single call to `waitFor`.

ShubhamRwt: Yeah now I get it. Thanks @tombentley.

operator-common/src/main/java/io/strimzi/operator/common/operator/resource/DeploymentOperator.java
tombentley: I think our `readiness()` should have a well-defined contract so that the returned Future only returns when we know that the resource is ready and that that readiness pertains to the current generation of the resource.

Our `readiness()` is defined on `AbstractReadyResourceOperator`, which is a superclass of `PodOperator`, `StatefulSetOperator` and `DeploymentConfigOperator`, the resources for which all have statuses (I'm assuming with a `observedGeneration`, but I've not checked), but also the `EndpointOperator` where `Endpoints` doesn't have a `status`. While we can use the `status.observedGeneration` to check that the readiness reported by `Fabric8` (defined by `io.fabric8.kubernetes.client.internal.readiness.Readiness#isResourceReady`)  pertains to the current generation of the resource we can't do that for `Endpoints`. Thus I don't see how you can avoid a race when checking for `Endpoints` readiness.

I think it would be weird to special case `Deployments` as requiring the generation check but not the other status-having resources. Why is it OK for them to have a racy semantic but not Deployments?

scholzj: It is that the resources have different semantics and at the end their readiness means something slightly different in each case I would say. At the end, Fabric8 already implements it differently for every resource. Also the way the resources apply their updates is different. So I do not think this can be done the same way for all of them.

But I'm fine if you want to keep our `readiness` method clean and do always the same. But in that case I would suggest to close this PR and leave this as it is in main with separate methods which need to be called both instead of trying to force the Deployment rollout mechanism onto other resources.

ShubhamRwt: @scholzj I agree on keeping the seperate methods. So should I close this PR?

scholzj: Well, if @tombentley thinks this breaks the intention of the `readiness()` method then I would say yes, we should keep it as it is and close this and the related issues.

tombentley: I think it can be argued that Fabric8's readiness checks should take into account `status.observedGeneration` (where present on the resource). I've opened https://github.com/fabric8io/kubernetes-client/issues/3474 about this. Let's see what they say before closing this issue.

ShubhamRwt:
Closing this PR as per discussed on the Community Call 18/11/2021

PR #5542, opened 2021-09-09
Add `insecure` flag to allow downloadable artifacts to be pulled from badly secured servers

scholzj:
### Type of change

- Enhancement / new feature

### Description

Kafka Connect Build has currently several artifact types which just download the artifacts from a remote server - usually using HTTP(S). In some cases, the remote HTTPS server might be badly secured (e.g. expired cert, self-signed cert, wrong hostname etc.) and the download might fail. This is correct as a default behaviour. But it would be nice to give users possibility to override this if they really want to do it.

This PR adds the option `insecure` to indicate that the artifact should be downloaded even from insecure server. Internally, this results in the `-k` option being added to the `curl` command. But most tools have similar options in case we use different tool in the future for the downloads. The new option applies to `jar`,  `other`, `zip` and `tgz` artifacts.

### Checklist

- [x] Write tests
- [x] Update documentation
- [x] Update CHANGELOG.md

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@
 * Allow configuring labels and annotations for JMX authentication secrets
 * Enable Cruise Control anomaly.detection configurations
 * Add support for building connector images from the Maven coordinates
-* Allow Kafka Connect Build artifacts to be downloaded from insecure servers 
+* Allow Kafka Connect Build artifacts to be downloaded from insecure servers (#5542)
 
 ### Changes, deprecations and removals
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 * Allow configuring labels and annotations for JMX authentication secrets
 * Enable Cruise Control anomaly.detection configurations
 * Add support for building connector images from the Maven coordinates
+* Allow Kafka Connect Build artifacts to be downloaded from insecure servers 
 
 ### Changes, deprecations and removals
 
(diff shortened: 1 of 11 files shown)


CHANGELOG.md
```diff
@@ -7,6 +7,7 @@
 * Allow configuring labels and annotations for JMX authentication secrets
 * Enable Cruise Control anomaly.detection configurations
 * Add support for building connector images from the Maven coordinates
+* Allow Kafka Connect Build artifacts to be downloaded from insecure servers
```
PaulRMellor, line 10:
```suggestion
* Allow Kafka Connect Build artifacts to be downloaded from insecure servers (#5542)
```

api/src/main/java/io/strimzi/api/kafka/model/connect/build/DownloadableArtifact.java
```diff
@@ -53,4 +54,15 @@ public String getSha512sum() {
     public void setSha512sum(String sha512sum) {
         this.sha512sum = sha512sum;
     }
+
+    @Description("By default, every connection using TLS is verified to be secure - that the server's certificate is trusted and contains the server name. " +
```
PaulRMellor, line 58:
```suggestion
    @Description("By default, connections using TLS are verified to check they are secure. A server certificate must be trusted and contain the server name. " +
```

api/src/main/java/io/strimzi/api/kafka/model/connect/build/DownloadableArtifact.java
```diff
@@ -53,4 +54,15 @@ public String getSha512sum() {
...
+            "By setting this option to `true`, all TLS verifications wil be disabled and the artifact will be downloaded even from a server otherwise considered insecure.")
```
PaulRMellor, line 59:
```suggestion
            "By setting this option to `true`, all TLS verifications are disabled and the artifact will be downloaded from a server, even when the server is considered insecure.")
```

PaulRMellor (approved):
Looks good. Couple of suggestions

sknot-rh (approved):

PR #5547, opened 2021-09-12
Fix incomplete Connect build URL regex

fvaleri:
Without the ending $, the string is considered valid if the expression matches anywhere within the string.

https://json-schema.org/understanding-json-schema/reference/string.html#regular-expressions


scholzj (commented):
Can you please update the description?

tombentley:
Can we review the other uses of pattern, because I wouldn't be surprised if this wasn't the only incorrect usage.

scholzj (approved):

fvaleri:
> Can we review the other uses of pattern, because I wouldn't be surprised if this wasn't the only incorrect usage.

Right, I found few other places where `^...$` is missing. I'm going to commit a fix for them too.

tombentley (approved):
LGTM, but one nit: Can we add something to the Javadoc for `@Pattern` to note the behaviour?

fvaleri:
Sure, let me add that note.

scholzj:
Thanks for the PR.

Issue #5524, opened 2021-09-03
Rolling update for KafkaConnect and KafkaMirrorMaker2


Pradeep2103:
**Is your feature request related to a problem? Please describe.**
When the clients CA certificate is renewed, the user certificates are also renewed but, the cluster operator does not trigger rolling update for the KafkaConnect and KafkaMirrormaker2 deployments and it continues to use the older truststore and keystore.

**Describe the solution you'd like**
Trigger a rolling update (similar to kafka brokers and entity operator) for the deployments when the mounted user secret is updated with a new certificate.

scholzj:
This should ideally also apply to MM1 and Bridge.

tombentley:
@sknot-rh is proposing to do this by computing a hash of the certs (and I guess sasl passwords/ oauth credentials) obtained from the Secrets consumed by the client pods via volume mounts, and annotating the pod template with it, so that the Deployment controller will rolls the pods when any of them. 

I put a bit of thought into this, because although it's simple and should work, might not be ideal...

For TLS certificates and keys [KIP-687](https://cwiki.apache.org/confluence/display/KAFKA/KIP-687%3A+Automatic+Reloading+of+Security+Store) _may_ apply (I think it's currently only intended for broker configs, but a case could be made for supporting this clients too). In one sense it would be desirable to leverage that since avoiding restarts is always a good thing. But there is no equivalent mechanism in Kafka for other credentials, such as SASLpassword / OAuth credentials, which aren't read from a separate file. So I think it would make sense to ignore KIP-687 for now and focus on a mechanism that requires a rolling restart. 

Another consideration is related to #5079/#5277, i.e. should we assume that these sensitive credentials are always provided via a mount of some kubernetes resource that can be accessed by the operator (and thus the operator can determine when the values changed by reading them)? It is possible to imagine some agent or sidecar which observed the sensitive credentials from within the client pod and exposed to the operator only the fact that they'd changed, allowing the operator (via the deployment controller) to roll the Deployment. Although that's pretty complicated it does leave the door open to other secret stores than kube and I kinda like how the CO could operate a `KafkaConnect` without the ability to connect to the Kafka cluster. 

There's also the point that really this is something that kube itself should provide support for. See https://github.com/kubernetes/kubernetes/issues/22368.

[...]

PR #5549, opened 2021-09-13
Roll pod when cert changed

sknot-rh:
### Type of change
- Enhancement

### Description
For now, Kafka does not support reloading the certificates dynamically. Thus, the rolling update is needed when the certificate is changed.

Fixes https://github.com/strimzi/strimzi-kafka-operator/issues/5524

### Checklist
- [x] Write tests
- [ ] Update documentation
- [ ] Update CHANGELOG.md


systemtest/src/test/java/io/strimzi/systemtest/connect/ConnectST.java
```diff
@@ -1238,6 +1239,79 @@ void testConfigureDeploymentStrategy(ExtensionContext extensionContext) {
...
+    @Tag(INTERNAL_CLIENTS_USED)
+    // changing the password in secret should cause the RU of connect pod
+    void testKafkaConnectWithPlainAndScramShaAuthenticationRolledAfterPasswordChanged(ExtensionContext extensionContext) {
+        final String namespaceName = StUtils.getNamespaceBasedOnRbac(NAMESPACE, extensionContext);
+        final String clusterName = mapWithClusterNames.get(extensionContext.getDisplayName());
+        final String userName = mapWithTestUsers.get(extensionContext.getDisplayName());
+        final String topicName = mapWithTestTopics.get(extensionContext.getDisplayName());
+        final String kafkaClientsName = mapWithKafkaClientNames.get(extensionContext.getDisplayName());
+
+        // Use a Kafka with plain listener disabled
```
scholzj, line 1252: Does this comment still apply? It looks like you use the plain listener.

systemtest/src/test/java/io/strimzi/systemtest/connect/ConnectST.java
```diff
@@ -1238,6 +1239,79 @@ void testConfigureDeploymentStrategy(ExtensionContext extensionContext) {
...
+        KafkaConnectUtils.waitUntilKafkaConnectRestApiIsAvailable(namespaceName, kafkaConnectPodName);
+
+        Map<String, String> connectSnapshot = DeploymentUtils.depSnapshot(namespaceName, KafkaConnectResources.deploymentName(clusterName));
+        String newPassword = "bmVjb0ppbmVob05lelNwcmF2bnlQYXNzd29yZA==";
+        Secret passwordSecret = new SecretBuilder()
+                .withNewMetadata()
+                    .withName(userName)
+                .endMetadata()
+                .addToData("password", newPassword)
+                .build();
```
scholzj, line 1303: I think it would be cleaner to use a custom password and provide it to the `KAfkaUSer` using the (new) feature for it and not just rewrite the generated user secret.

operator-common/src/main/java/io/strimzi/operator/common/operator/resource/SecretOperator.java
```diff
@@ -29,4 +36,28 @@ public SecretOperator(Vertx vertx, KubernetesClient client) {
     protected MixedOperation<Secret, SecretList, Resource<Secret>> operation() {
         return client.secrets();
     }
+
+
```
scholzj, line 40: Double empty line?

operator-common/src/main/java/io/strimzi/operator/common/operator/resource/SecretOperator.java
```diff
@@ -29,4 +36,28 @@ public SecretOperator(Vertx vertx, KubernetesClient client) {
...
+            return this.getAsync(namespace, ((KafkaClientAuthenticationPlain) auth).getPasswordSecret().getSecretName())
+                    .compose(secret -> Future.succeededFuture(secret.getData().get(((KafkaClientAuthenticationPlain) auth).getPasswordSecret().getPassword())));
+        }
+        if (auth instanceof KafkaClientAuthenticationScramSha512) {
+            return this.getAsync(namespace, ((KafkaClientAuthenticationScramSha512) auth).getPasswordSecret().getSecretName())
+                    .compose(secret -> Future.succeededFuture(secret.getData().get(((KafkaClientAuthenticationScramSha512) auth).getPasswordSecret().getPassword())));
+        } else {
+            return Future.failedFuture("Auth type " + auth.getType() + " does not have a password property");
+        }
+    }
```
scholzj, line 62: Do we really want to have these here? It introduces the `api` module into this quite heavily which was not the case before. They are also very specific where as the SecretOperator purpose is very generic.

scholzj (commented):
Few nits but LGTM otherwise. 

Do we everywhere check that all the different secrets exist? If not, we might want to couple these two features together, so that we:
* Provide nice error messages (your code currently seems to expect they exist and if the don't will probably just throw some ugly exceptions)
* Get the secrets only once

operator-common/src/main/java/io/strimzi/operator/common/operator/resource/SecretOperator.java
sknot-rh: @tombentley Do we want to pass certificate, key, and password as a String parameter?

scholzj: The `api` is just part of it ... what is the value of having this here instead of in the Util class?

scholzj (approved):

operator-common/src/main/java/io/strimzi/operator/common/operator/resource/SecretOperator.java
tombentley: Currently we've failed to really encapsulate and abstract our certificate handling. What I mean by that is everywhere we pass around a `Secret` which we _happen to know_ contains a key store (broker certificate and key), or truststore (ca certificate). But that requires that the consumer of that `Secret` knows what _type_ of secret is it, and how to decode the values within it into something it can use. 

I think there is value is trying to better encapsulate these concepts, so rather than passing around Secrets we pass around something higher level. For example a `AdminClientSupplier` might accept some Strimzi `Truststore` class and a Strimzi `Keystore` class. Providing Resource Operators for those things (`Truststore` and `Keystore`) would be a logical thing to do. This would provide a path towards being able to abstract where certain credentials are stored (e.g. option to store keys in Vault), and should make the KAO slightly simpler to understand, because there are fewer generic Secrets floating around.

At least that was my motivation for suggesting this, and I'd be interested to know whether people think that's a desirable goal or not (and if they have specific reasons to think it wouldn't work out).

tombentley:
@sknot-rh I'm fine with the code, but would like to find a consensus on the [abstraction](https://github.com/strimzi/strimzi-kafka-operator/pull/5549#discussion_r714675992) idea before we decide to merge this in its current form.

tombentley (approved):

PR #5564, opened 2021-09-15
Await preferred leadership rolling

tombentley:
### Type of change

- Enhancement

### Description

This PR changes how we do rolling restarts. Rather than just waiting for a broker to be available (which can lead to the next broker being restarted before the last has caught up, also also leaves the cluster unbalanced at the end of a rolling restart, since the last-rolled broker won't be leading any partitions) we required availability and then leadership of all the partitions of which the just-rolled broker is the preferred leader.

This will (obviously) slow down rolling in the general case, but frequently the restarted broker will catch up fairly quickly.

The 30s wait time is pretty much arbitrary and I'm open to tuning this.

### Checklist

- [ ] Write tests
- [ ] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -158,7 +158,7 @@
         <module>mirror-maker-agent</module>
         <module>tracing-agent</module>
         <module>test</module>
-        <module>test-container</module>
+        <!-- <module>test-container</module> -->
         <module>crd-annotations</module>
         <module>crd-generator</module>
         <module>api</module>
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/resource/KafkaAvailability.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/resource/KafkaAvailability.java
@@ -14,6 +14,7 @@ import org.apache.kafka.clients.admin.ConfigEntry;
 import org.apache.kafka.clients.admin.ListTopicsOptions;
 import org.apache.kafka.clients.admin.TopicDescription;
 import org.apache.kafka.common.Node;
+import org.apache.kafka.common.TopicPartition;
 import org.apache.kafka.common.TopicPartitionInfo;
 import org.apache.kafka.common.config.ConfigResource;
 import org.apache.kafka.common.config.TopicConfig;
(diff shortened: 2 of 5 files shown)


tombentley:
Note: I still have to write a unit test for this.

scholzj:
So, if I get this right ... after the broker is rolled, this waits for the pod to be ready, partitions to be available and then it uses the Admin APi to move the leadership of the preferred leaders? What does it do when it timeouts? Does the next reconciliation follow up on this? Or does it continue s if no rolling update happened before?

tombentley:
This is a bit tricky, because the code I've written so far:

1. Doesn't have a timeout 
2. Adding a naive one (at the `// TODO`) means the outer retry could blow up the overall roll time. 

So perhaps we should use the outer retry mechanism, rather than retrying as I've coded it so far. A possible problem with that is the outer retry might be configured to give up quite quickly. And the outer loop is difficult to reason about.

scholzj:
> So perhaps we should use the outer retry mechanism, rather than retrying as I've coded it so far. A possible problem with that is the outer retry might be configured to give up quite quickly. And the outer loop is difficult to reason about.

So, what is the actual issue? I'm not sure I understand the main reasons for this. Is it just the unequal distribution of replica leaders? 

If yes, I wonder whether:
* it really is big enough problem to not be optimistic and say _if it works it works, if it fails we just ignore it and Kafka will reassign it sooner or later anyway_?.
* Or if we thing it is an issue to avoid ... should have a step outside of the rolling update which would every reconciliation check the leaders and reassign them if needed? The unequal distribution of leaders can happen even after a regular pod crash etc., so if it is important, it would catch it also outside of rolling updates and it could also continue the work from previous reconciliation if needed.

tombentley:
> > So perhaps we should use the outer retry mechanism, rather than retrying as I've coded it so far. A possible problem with that is the outer retry might be configured to give up quite quickly. And the outer loop is difficult to reason about.
> 
> So, what is the actual issue? I'm not sure I understand the main reasons for this. Is it just the unequal distribution of replica leaders?

It's not just the balancing of leadership. For `__consumer_offsets` and `__transaction_state` becoming leader means the respective coordinators need to load state, which they do asynchronously. While that's happening clients can discover the new coordinator but will get `COORDINATOR_LOAD_IN_PROGRESS` and have to retry. With 3-way replication on brokers `1,2,3` currently led by `1` and roll order `1,2,3` we might get the sequence of coordinators `1,2,3,1` (where the last `1` is due to Kafka's automatic reelection, or a manual reelection), thus 3 changes of coordinator, where, due to the asynchrony, and the speed at which we currently roll there might only be small windows during which a broker is able to act as coordinator (i.e. when `1` is first rolls and `2` is elected leader it might only actually have loaded the state for a short while before it is rolled). By always electing the preferred leader after each restart we'd get the sequence of coordinators `1,2,1`, so only 2 changes of coordinator and less disruption to clients.

Doing this for other partitions than `__consumer_offsets` and `__transaction_state` is also beneficial because, again, there's less need for clients to spend time discovering new short-lived leaders and fewer resends needed overall. It should have a consequent reduction in tail latency (which is dominated by changes in leadership).

Coming back to my point:

> > And the outer loop is difficult to reason about.

I think the code has got to the point where this needs to be dealt with _anyway_.

scholzj:
@tombentley Thanks for the explanation. I understand it better now, thanks.

pom.xml
```diff
@@ -158,7 +158,7 @@
         <module>mirror-maker-agent</module>
         <module>tracing-agent</module>
         <module>test</module>
-        <module>test-container</module>
+        <!-- <module>test-container</module> -->
```
scholzj, line 161: Should this be uncommented again?

scholzj (commented):
I guess this file `cluster-operator/.attach_pid450387` should not be here?

PR #5565, opened 2021-09-15
Expose connector metrics

ivalexa:
Signed-off-by: alexivsn <alex@securenative.com>

### Type of change

Enhancement / Bugfix

### Description

Currently, for Kafka Connect JMX metrics, connector status is not exposed, instead, the status of the task are exposed under the name of kafka_connect_connector_status which is misleading and when the connector fails to start there is JMX metrics available since the task is down, I have exposed metrics from connector-metrics MBean based on official confluent documentation and updated existing connector task status name from  kafka_connect_connector_status to kafka_connect_connector_task_status

https://github.com/strimzi/strimzi-kafka-operator/discussions/5551

The changes have been validated with JMX exporter, screenshot are attached below:
image
image

### DIFF
--- a/examples/metrics/kafka-connect-metrics.yaml
+++ b/examples/metrics/kafka-connect-metrics.yaml
@@ -91,15 +91,25 @@ data:
       help: "Kafka $1 JMX metric type $2"
       type: GAUGE
 
+    #kafka.connect:type=connector-metrics,connector="{connector}"
+    - pattern: 'kafka.(.+)<type=connector-metrics, connector=(.+)><>(connector-class|connector-type|connector-version|status): (.+)'
+      name: kafka_connect_connector_$3
+      value: 1
+      labels:
+        connector: "$2"
+        $3: "$4"
+      help: "Kafka Connect $3 JMX metric type connector"
+      type: GAUGE
+      
     #kafka.connect:type=connector-task-metrics,connector="{connector}",task="{task}<> status"
     - pattern: 'kafka.connect<type=connector-task-metrics, connector=(.+), task=(.+)><>status: ([a-z-]+)'
-      name: kafka_connect_connector_status
+      name: kafka_connect_connector_task_status
       value: 1
       labels:
         connector: "$1"
         task: "$2"
         status: "$3"
-      help: "Kafka Connect JMX Connector status"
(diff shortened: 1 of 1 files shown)


examples/metrics/kafka-connect-metrics.yaml
```diff
@@ -91,15 +91,25 @@ data:
...
+      labels:
+        connector: "$2"
+        $3: "$4"
+      help: "Kafka Connect $3 JMX metric type connector"
+      type: GAUGE
+      
     #kafka.connect:type=connector-task-metrics,connector="{connector}",task="{task}<> status"
     - pattern: 'kafka.connect<type=connector-task-metrics, connector=(.+), task=(.+)><>status: ([a-z-]+)'
-      name: kafka_connect_connector_status
+      name: kafka_connect_connector_task_status
```
scholzj, line 106: So, from the diff it looks like the `kafka_connect_connector_task_status` replaces the `kafka_connect_connector_status`. Doesn't that mean that the `kafka_connect_connector_status` will be missing now?

scholzj (commented):
Thanks for the PR. Could you please do the change to the file `packaging/examples/metrics/kafka-connect-metrics.yaml` instead of `examples/metrics/kafka-connect-metrics.yaml`? The `examples/` file are form the last release and should not change.

examples/metrics/kafka-connect-metrics.yaml
scholzj: Sorry, I read the description once again ... so are you saying that there is basically no `kafka_connect_connector_status` ... it needs to be replced by `kafka_connect_connector_task_status` with the right task label?

In that case, do we need to also update the Grafana chart? Or is it not used there?

sknot-rh: I do not see this metric to be used in the Grafana dashboard examples.

ivalexa: Hey sorry for the delay, the task status metrics, is unused by the current Grafana dashboard, I have updated the Grafana dashboard template and moved the changes to the packaging folder as it's required by build.
https://github.com/strimzi/strimzi-kafka-operator/pull/5583

PR #5249, opened 2021-07-03
Add TLS-External authentication to User Operator

scholzj:
### Type of change

- Enhancement / new feature

### Description

The User Operator manages authentication, authorization and quotas for Kafka users. However, in some cases, users want to use it only for some of these. This is in general easy when they want to use it for authentication but not for ACLs or Quotas. But it is not always so easy when they want to use it for ACLs and Quotas but not for authentication. Especially if they want to use TLS authentication but issue the user certificates on their own and the ACLs and Quotas need to be set for the `CN=<username>` account. This is something raised multiple times with the users.

This PR adds new Authentication type to the User Operator / `KafkaUser` CR - it is called `tls-external`. When the KafkaUser is configures with `tls-external`, it will tel the operator that it:
* Should not generate the user TLS certificate and create the secret
* But it should treat the user as TLS user and use the `CN=<username>` account when setting up ACLs and Quotas

This should close #5077

### Checklist

- [x] Write tests
- [x] Make sure all tests pass
- [x] Try your changes from Pod inside your Kubernetes and OpenShift cluster, not just locally
- [x] Reference relevant issue(s) and close them after merging
- [x] Update CHANGELOG.md

PR #5576, opened 2021-09-17
[MO] - [system test] -> KafkaUser tls external

see-quick:

### Type of change

- Enhancement / new feature
- Refactoring

### Description

This PR adding a test case for the `tls-external` property used in KafkaUser. The verification process has three steps where:

1. Verify Quotas for `CN=<username>`
2. Verify that secret is not generated for specific `username`
3. Verify that ACLs are properly set.
 
More information [here](https://github.com/strimzi/strimzi-kafka-operator/pull/5249).

### Checklist

- [x] Write tests

### DIFF
--- a/systemtest/src/main/java/io/strimzi/systemtest/utils/kafkaUtils/KafkaUserUtils.java
+++ b/systemtest/src/main/java/io/strimzi/systemtest/utils/kafkaUtils/KafkaUserUtils.java
@@ -92,7 +92,7 @@ public class KafkaUserUtils {
 
     /**
      * Wait until KafkaUser is in desired state
-     * @param namespaceNAme Namespace name
+     * @param namespaceName Namespace name
      * @param userName name of KafkaUser
      * @param state desired state
      */
--- a/systemtest/src/main/java/io/strimzi/systemtest/utils/kafkaUtils/KafkaUserUtils.java
+++ b/systemtest/src/main/java/io/strimzi/systemtest/utils/kafkaUtils/KafkaUserUtils.java
@@ -14,7 +14,6 @@ import io.strimzi.systemtest.resources.ResourceOperation;
 import io.strimzi.systemtest.resources.crd.KafkaUserResource;
 import io.strimzi.systemtest.utils.kubeUtils.objects.SecretUtils;
 import io.strimzi.test.TestUtils;
-import kafka.security.auth.Read;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 
--- a/systemtest/src/main/java/io/strimzi/systemtest/utils/kafkaUtils/KafkaUserUtils.java
+++ b/systemtest/src/main/java/io/strimzi/systemtest/utils/kafkaUtils/KafkaUserUtils.java
@@ -14,6 +14,7 @@ import io.strimzi.systemtest.resources.ResourceOperation;
 import io.strimzi.systemtest.resources.crd.KafkaUserResource;
 import io.strimzi.systemtest.utils.kubeUtils.objects.SecretUtils;
 import io.strimzi.test.TestUtils;
+import kafka.security.auth.Read;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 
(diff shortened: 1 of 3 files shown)


systemtest/src/main/java/io/strimzi/systemtest/utils/kafkaUtils/KafkaUserUtils.java
```diff
@@ -92,16 +92,30 @@ public static void waitUntilKafkaUserStatusConditionIsPresent(String namespaceNa
 
     /**
      * Wait until KafkaUser is in desired state
+     * @param namespaceNAme Namespace name
```
im-konge, line 95:
```suggestion
     * @param namespaceName Namespace name
```

systemtest/src/test/java/io/strimzi/systemtest/operators/user/UserST.java
```diff
@@ -369,6 +383,86 @@ void testCreatingUsersWithSecretPrefix(ExtensionContext extensionContext) {
...
+
+        // verify (c) - ACLs are correctly set for CN=<username>
+        List<AclRule> acls = ((KafkaUserAuthorizationSimple) user.getSpec().getAuthorization()).getAcls();
+
+        assertThat(acls.size(), is(2));
+        assertThat(acls.get(0), is(writeRule));
+        assertThat(acls.get(1), is(describeRule));
+    }
+
+
```
im-konge, line 465: Extra empty line

im-konge (approved):
Two minor comments, LGTM otherwise

systemtest/src/test/java/io/strimzi/systemtest/operators/user/UserST.java
```diff
@@ -218,38 +230,40 @@ void testUserTemplate(ExtensionContext extensionContext) {
...
         String command = "bin/kafka-configs.sh --bootstrap-server localhost:9092 --describe --entity-type users";
         LOGGER.debug("Command for kafka-configs.sh {}", command);
 
-        ExecResult result = cmdKubeClient(NAMESPACE).execInPod(KafkaResources.kafkaPodName(userClusterName, 0), "/bin/bash", "-c", command);
+        ExecResult result = cmdKubeClient(namespaceName).execInPod(KafkaResources.kafkaPodName(clusterName, 0), "/bin/bash", "-c", command);
         assertThat(result.out().contains("Quota configs for user-principal '" + userName + "' are"), is(true));
         assertThat(result.out().contains("request_percentage=" + reqPerc), is(true));
         assertThat(result.out().contains("producer_byte_rate=" + prodRate), is(true));
         assertThat(result.out().contains("consumer_byte_rate=" + consRate), is(true));
         assertThat(result.out().contains("controller_mutation_rate=" + mutRate), is(true));
```
scholzj, line 258: Not sure it has to be fixed in this PR and maybe it is not an issue at all and I just missed somehting. But what I wonder about is:
* You run all the tests in parallel against the same cluster, so there will be multiple users
* Your `command` lists quotas for all users as it seems

So, how do you know that for example `assertThat(result.out().contains("request_percentage=" + reqPerc), is(true))` actually asserts quota from this particular user and not from another user?

systemtest/src/test/java/io/strimzi/systemtest/operators/user/UserST.java
```diff
@@ -369,6 +383,86 @@ void testCreatingUsersWithSecretPrefix(ExtensionContext extensionContext) {
...
+            .endAclRuleTopicResource()
+            .withOperation(AclOperation.DESCRIBE)
+            .build();
+
+        // exercise (a) - create Kafka cluster with support for authorization
+        resourceManager.createResource(extensionContext, KafkaTemplates.kafkaEphemeral(clusterName, 1, 1)
+            .editSpec()
+                .editKafka()
+                    .withNewKafkaAuthorizationSimple()
+                        .withSuperUsers("CN=" + userName)
```
scholzj, line 412: Why do you set the user as super user? is that needed?

systemtest/src/test/java/io/strimzi/systemtest/operators/user/UserST.java
see-quick: That's a good point. I think I should create a separate Kafka cluster for each KafkaUser. (I can open another PR to fix that...)

scholzj: Can't we somehow add the username to the command instead? From the perspective of the execution time it is of course great to share the clusters for this ... we just need to make sure the quotas are evaluated for given user.

see-quick: https://github.com/strimzi/strimzi-kafka-operator/pull/5584

systemtest/src/test/java/io/strimzi/systemtest/operators/user/UserST.java
see-quick: We don't need that...

scholzj (approved):

PR #5606, opened 2021-09-23
[systemtest] Upgrade - fix check if Kafka version is in supported Kafka versions list

im-konge:

### Type of change

- Bugfix

### Description

This PR fixes issue in our upgrade tests - in check, if Kafka version, with which we deployed our Kafka cluster, is in list of supported versions - until now we checked that Kafka version is in list of all Kafka versions inside the `kafka-versions.yaml`, so it was always true and we always waited for rolling update - that's wrong, because when we upgrade CO to version, where certain Kafka version is not supported, we are waiting for RU instead of changing Kafka version to supported one.

### Checklist

### DIFF
--- a/systemtest/src/test/java/io/strimzi/systemtest/upgrade/StrimziUpgradeST.java
+++ b/systemtest/src/test/java/io/strimzi/systemtest/upgrade/StrimziUpgradeST.java
@@ -267,7 +267,7 @@ public class StrimziUpgradeST extends AbstractUpgradeST {
         logPodImages(clusterName);
         changeClusterOperator(testParameters, NAMESPACE, extensionContext);
 
-        if (TestKafkaVersion.containsVersion(getDefaultKafkaVersionPerStrimzi(testParameters.getString("fromVersion")).version())) {
+        if (TestKafkaVersion.supportedVersionsContainsVersion(getDefaultKafkaVersionPerStrimzi(testParameters.getString("fromVersion")).version())) {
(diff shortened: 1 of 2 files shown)


systemtest/src/test/java/io/strimzi/systemtest/upgrade/StrimziUpgradeST.java
```diff
@@ -267,7 +267,7 @@ private void performUpgrade(JsonObject testParameters, ExtensionContext extensio
         logPodImages(clusterName);
         changeClusterOperator(testParameters, NAMESPACE, extensionContext);
 
-        if (TestKafkaVersion.containsVersion(getDefaultKafkaVersionPerStrimzi(testParameters.getString("fromVersion")).version())) {
+        if (TestKafkaVersion.supportedVersionsContainsVersion(getDefaultKafkaVersionPerStrimzi(testParameters.getString("fromVersion")).version())) {
```
scholzj, line 270: So I assume this change will cause the errors in the CO (about unsupported Kafka version) to be ignored and move directly to the upgrade?

im-konge: Yes, if the Kafka version is unsupported, we will not wait and we'll move directly to upgrade, how you wrote.

scholzj (approved):

Frawless (approved):

PR #5610, opened 2021-09-24
docs: kubernetes upgrade guidance

PaulRMellor:

**Documentation**

Adds an outline of the Kubernetes upgrade process in the context of Strimzi.
Added as preliminary step in the _required upgrade sequence_

### Checklist

- [ ] Write tests
- [ ] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/documentation/assemblies/upgrading/assembly-upgrade-kafka.adoc
+++ b/documentation/assemblies/upgrading/assembly-upgrade-kafka.adoc
@@ -7,17 +7,19 @@
 
 To upgrade brokers and clients without downtime, you _must_ complete the Strimzi upgrade procedures in the following order:
 
-. . When upgrading from {ConvertAfterProductVersion} or earlier, update existing custom resources to support the `v1beta2` API version.
+. Update your Kubernetes cluster to a supported version.
++
+--
+* xref:con-upgrade-cluster-{context}[]
+--
+
+. When upgrading from {ConvertAfterProductVersion} or earlier, update existing custom resources to support the `v1beta2` API version.
 +
 --
 * xref:assembly-upgrade-resources-{context}[]
 --
 
 . Update your Cluster Operator to a new Strimzi version.
-+
---
-* xref:proc-upgrading-the-co-{context}[]
---
 +
 The approach you take depends on how you xref:cluster-operator-{context}[deployed the Cluster Operator].
 +
--- a/documentation/modules/deploying/con-deploy-cluster-operator-from-operator-hub.adoc
+++ b/documentation/modules/deploying/con-deploy-cluster-operator-from-operator-hub.adoc
@@ -11,3 +11,8 @@ It offers you an alternative way to install stable versions of Strimzi using the
 The {OperatorLifecycleManager} is used for the installation and management of all Operators published on {OperatorHub}.
 
 To install Strimzi from {OperatorHub}, locate the _Strimzi Kafka Operator_ and follow the instructions provided.
+
+WARNING: Make sure you use the appropriate update channel.
+If you are on a supported version of the Kubernetes, installing Strimzi from the default stable channel is safe.
+However, if you are using a version of Kubernetes that is unsupported, installing Strimzi from the stable channel is unsafe,
+especially when automatic updates are enabled, as the cluster will receive automatic updates with new components that are unsupported by the Kubernetes release.
--- a/documentation/modules/upgrading/con-upgrade-cluster.adoc
+++ b/documentation/modules/upgrading/con-upgrade-cluster.adoc
@@ -1,0 +1,40 @@
+// This module is included in the following assemblies:
+//
+// upgrading/assembly-upgrade.adoc
+
+[id='con-upgrade-cluster-{context}']
+= Upgrading Kubernetes
+
+[role="_abstract"]
+If you are upgrading your Kubenerates cluster to a later supported version, do it before upgrading Strimzi.
+Refer to the Kubernetes upgrade documentation to check the upgrade path and the steps to upgrade your nodes correctly.
+
+You’ll need admin privileges to be able to upgrade your Kubernetes cluster.
+As a Strimzi admin with permission to perform the upgrade, the basic steps are as follows:
+
+. Backup your Kubernetes components
+. Check and provision for any backwards incompatible changes in the release notes of the target version
+. Upgrade the Kubernetes cluster manually or automatically
+
+You'll probably want to upgrade the `kubectl` command-line tool to the latest version too.
+
+You can upgrade Kubernetes with minimal downtime for Strimzi.
+Consult the latest Kubernetes documentation for guidance on the upgrade process.
+For a manual upgrade, you can choose to upgrade nodes, or remove old nodes and replace them with new nodes.
+
+After you have verified that the Kubernetes cluster upgrade is successful, you can upgrade Strimzi.
+
+If you used the Operator Lifecycle Manager (OLM) to install the Strimzi Operator using the _stable_ channel, Strimzi updates are discovered automatically after the Kubernetes upgrade.
+If you have specified automatic OLM updates, the Strimzi Operator will automatically upgrade to the version of Strimzi that was released for the upgraded Kubernetes cluster version.
+If you are manually updating using OLM, you’ll need to approve the upgrade.
+
+Otherwise, follow the required upgrade sequence depending on the method used to deploy Strimzi.
+
+WARNING: We recommend using OLM's manual update policy for Strimzi.
+You can then perform any required upgrade preparation steps before the upgrade.
+An automatic upgrade can occur before performing these steps, which can cause issues with the performance of Strimzi.
+
+[role="_additional-resources"]
+.Additional resources
+* {cluster-docs}[Kubernetes documentation^]
+* {OLMOperatorDocs}
--- a/documentation/shared/attributes.adoc
+++ b/documentation/shared/attributes.adoc
@@ -58,6 +58,7 @@
 
 // External links
 :aws-ebs: link:https://aws.amazon.com/ebs/[Amazon Elastic Block Store (EBS)^]
+:cluster-docs: link:https://kubernetes.io/docs/home/
 :docs-okd: link:https://docs.okd.io/3.11/dev_guide/builds/index.html[builds^]
 :kafkaDoc: link:https://kafka.apache.org/documentation/[Apache Kafka documentation^]
 :KafkaRacks: link:https://kafka.apache.org/documentation/#basic_ops_racks[Kafka racks documentation^]
--- a/documentation/assemblies/upgrading/assembly-upgrade.adoc
+++ b/documentation/assemblies/upgrading/assembly-upgrade.adoc
@@ -63,6 +63,8 @@ A reduction in cluster availability increases the chance that a broker failure w
 
 include::assembly-upgrade-kafka.adoc[leveloffset=+1]
 
+include::modules/con-upgrade-cluster.adoc[leveloffset=+1]
+
 include::assembly-upgrade-resources.adoc[leveloffset=+1]
 
 include::modules/proc-upgrade-cluster-operator.adoc[leveloffset=+1]


PaulRMellor:
@kyguy This PR has the Kubernetes upgrade content we discussed. Not sure if we need to say something on graceful restarts. Or strategies that involve things like `terminationGracePeriodSeconds`, `PodDisruptionBudget` or `drain`

documentation/modules/upgrading/con-upgrade-cluster.adoc
```diff
@@ -0,0 +1,40 @@
+// This module is included in the following assemblies:
+//
+// upgrading/assembly-upgrade.adoc
+
+[id='con-upgrade-cluster-{context}']
+= Upgrading Kubernetes
+
+[role="_abstract"]
+If you are upgrading your Kubenerates cluster to a later supported version, do it before upgrading Strimzi.
```
laidan6000, line 9:
```suggestion
If you want to upgrade your Kubernetes cluster to a later version supported by Strimzi, do this before upgrading Strimzi.
```

documentation/modules/upgrading/con-upgrade-cluster.adoc
```diff
@@ -0,0 +1,40 @@
...
+Refer to the Kubernetes upgrade documentation to check the upgrade path and the steps to upgrade your nodes correctly.
+
+You’ll need admin privileges to be able to upgrade your Kubernetes cluster.
+As a Strimzi admin with permission to perform the upgrade, the basic steps are as follows:
+
+. Backup your Kubernetes components
```
laidan6000, line 15:
```suggestion
. Back up your Kubernetes components
```

documentation/modules/upgrading/con-upgrade-cluster.adoc
```diff
@@ -0,0 +1,40 @@
...
+
+You'll probably want to upgrade the `kubectl` command-line tool to the latest version too.
+
+You can upgrade Kubernetes with minimal downtime for Strimzi.
+Consult the latest Kubernetes documentation for guidance on the upgrade process.
+For a manual upgrade, you can choose to upgrade nodes, or remove old nodes and replace them with new nodes.
+
+After you have verified that the Kubernetes cluster upgrade is successful, you can upgrade Strimzi.
+
+If you used the Operator Lifecycle Manager (OLM) to install the Strimzi Operator using the _stable_ channel, Strimzi updates are discovered automatically after the Kubernetes upgrade.
```
laidan6000, line 27:
```suggestion
If you used the Operator Lifecycle Manager (OLM) to install the Strimzi Operator from the _stable_ channel, Strimzi updates are discovered automatically after the Kubernetes upgrade.
```

documentation/modules/upgrading/con-upgrade-cluster.adoc
```diff
@@ -0,0 +1,40 @@
...
+If you have specified automatic OLM updates, the Strimzi Operator will automatically upgrade to the version of Strimzi that was released for the upgraded Kubernetes cluster version.
```
laidan6000, line 28:
```suggestion
If you have specified _automatic_ OLM updates, the Strimzi Operator will automatically upgrade to the version of Strimzi that was released for the upgraded Kubernetes cluster version.
```

laidan6000 (changes requested):
Looks great. Just some minor comments.

documentation/assemblies/upgrading/assembly-upgrade-kafka.adoc
```diff
@@ -7,17 +7,19 @@
 
 To upgrade brokers and clients without downtime, you _must_ complete the Strimzi upgrade procedures in the following order:
 
-. . When upgrading from {ConvertAfterProductVersion} or earlier, update existing custom resources to support the `v1beta2` API version.
+. Update your Kubernetes cluster to a supported version.
```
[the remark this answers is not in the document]
kyguy: Emphasis on "supported" too! But articulating what is supported and not supported is difficult! For example, Strimzi 0.22 users wanting to upgrade to k8s v1.22 will need to migrate their Strimzi custom resources to `v1beta2` before the k8s upgrade since `apiextensions/v1beta1` custom resources are not supported in k8s v1.22. After doing the CRD migration, they can safely upgrade their k8s cluster.  So Strimzi 0.22 which is the last Strimzi version to use `apiextensions/v1beta1` is not fully supported on k8s 1.22.

documentation/modules/deploying/con-deploy-cluster-operator-from-operator-hub.adoc
```diff
@@ -11,3 +11,8 @@ It offers you an alternative way to install stable versions of Strimzi using the
 The {OperatorLifecycleManager} is used for the installation and management of all Operators published on {OperatorHub}.
 
 To install Strimzi from {OperatorHub}, locate the _Strimzi Kafka Operator_ and follow the instructions provided.
+
+WARNING: Make sure you use the appropriate update channel.
+If you are on a supported version of the Kubernetes, installing Strimzi from the default stable channel is safe.
+However, if you are using a version of Kubernetes that is unsupported, installing Strimzi from the stable channel is unsafe,
+especially when automatic updates are enabled, as the cluster will receive automatic updates with new components that are unsupported by the Kubernetes release.
```
kyguy, line 18: It is a little tricky here, even on supported a k8s version, automatic updates from the `stable` channel are unsafe. One example is the automatic upgrade from Strimzi 0.22 -> 0.23, this is unsafe even on supported k8s versions.
(since Strimzi 0.23 is supported on k8s 1.16 and above but we must do a special API conversion from Strimzi  0.22 to 0.23).

Maybe we could say something like  "installing Strimzi from the default stable channel is generally safe. However, we must be mindful to preform required API conversions between particular Strimzi releases, particularly between Strimzi 0.22 and Strimzi 0.23. For this reason it is highly encouraged to avoid OLM automatic updates which will skip this API conversion process"

documentation/modules/upgrading/con-upgrade-cluster.adoc
```diff
@@ -0,0 +1,40 @@
+// This module is included in the following assemblies:
+//
+// upgrading/assembly-upgrade.adoc
+
+[id='con-upgrade-cluster-{context}']
+= Upgrading Kubernetes
```
scholzj, line 6: I think this is a bit weird:
* The first part focus on how to upgrade Kubernetes cluster. It seems to be a strange mixture of specific and unspecific steps which risk that you forget something and does not really help in any case. I think the whole part should be covered by this single sentence:
  ```
  Refer to your Kubernetes distribution upgrade documentation to check the upgrade path and the steps to upgrade your nodes correctly.
  ```
  I would not leave anything else.
* The second part seems focus on OLM upgrades. But these are completely unrelated to Kubernetes upgrades. Strimzi supports wide variety of Kube versions, so the probability that Kubernetes upgrade enables some new Strimzi upgrade is highly unlikely. In fact it happened only once so far - in a specific constellation when you upgraded to Kube 1.16 after Strimzi 0.22 was released. So I do not think this is useful either.

----

Instead we IMHO need to focus on how to keep the existing clusters fully available during the upgrade. I would expect that almost any upgrade procedure would include shutting down nodes => i.e. first draining them. So the focus here should be on:
* Configuring Pod Disruption Budgets
* Manually rolling the pods while making sure the topics are in-sync / using the Drain Cleaner and operator to do this automatically

documentation/shared/attributes.adoc
```diff
@@ -58,6 +58,7 @@
 
 // External links
 :aws-ebs: link:https://aws.amazon.com/ebs/[Amazon Elastic Block Store (EBS)^]
+:cluster-docs: link:https://kubernetes.io/docs/home/
```
scholzj, line 61: Maybe calling it `kubernetes-docs` would be more self-explaining?

documentation/assemblies/upgrading/assembly-upgrade-kafka.adoc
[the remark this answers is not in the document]
PaulRMellor: I've changed to _Make sure..._
And added a line for latest:

image

documentation/modules/deploying/con-deploy-cluster-operator-from-operator-hub.adoc
PaulRMellor: Thanks @kyguy I've changed to this:

image

documentation/modules/upgrading/con-upgrade-cluster.adoc
PaulRMellor: @scholzj I've changed to be less prescriptive. And added a link to the Supported versions page. (generic link attribute name so a downstream link can use the same)

image

documentation/modules/upgrading/con-upgrade-cluster.adoc
```diff
@@ -0,0 +1,40 @@
...
+
+[role="_abstract"]
+If you are upgrading your Kubenerates cluster to a later supported version, do it before upgrading Strimzi.
+Refer to the Kubernetes upgrade documentation to check the upgrade path and the steps to upgrade your nodes correctly.
+
+You’ll need admin privileges to be able to upgrade your Kubernetes cluster.
+As a Strimzi admin with permission to perform the upgrade, the basic steps are as follows:
```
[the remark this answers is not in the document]
PaulRMellor: Reduced to _The basic upgrade steps are as follows:_

documentation/modules/deploying/con-deploy-cluster-operator-from-operator-hub.adoc
```diff
@@ -11,3 +11,8 @@ It offers you an alternative way to install stable versions of Strimzi using the
 The {OperatorLifecycleManager} is used for the installation and management of all Operators published on {OperatorHub}.
 
 To install Strimzi from {OperatorHub}, locate the _Strimzi Kafka Operator_ and follow the instructions provided.
+
+WARNING: Make sure you use the appropriate update channel.
```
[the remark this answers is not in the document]
PaulRMellor: @scholzj This is what we have now

image

documentation/modules/upgrading/con-upgrade-cluster.adoc
PaulRMellor: Yes. I made a comment about this pod disruption and draining when I pushed up the PR. Thanks for confirming.

documentation/modules/deploying/con-deploy-cluster-operator-from-operator-hub.adoc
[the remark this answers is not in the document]
scholzj: The OLM channel has nothing to do with the Kubernetes version. I do not think it should be part of this

documentation/modules/upgrading/con-upgrade-cluster.adoc
scholzj: I would leave this just at "Check the supported Kubernetes versions for your Strimzi version" and vice-versa. I do not think there is any link between the upgrades and we should not suggest that the upgrades might need to be done together or anything like that. these are really two separate things.

documentation/modules/deploying/con-deploy-cluster-operator-from-operator-hub.adoc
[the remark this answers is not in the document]
PaulRMellor: Okay. If we take that out, we have:

image

documentation/shared/attributes.adoc
PaulRMellor: I went for generic for downstream. But I don't suppose it matters, so updated.

PaulRMellor:
@scholzj I've added an outline of pod disruption and rolling updates inc. drain cleaner. Not too much detail on drain cleaner, as it will have it's own section.

documentation/modules/upgrading/con-upgrade-cluster.adoc
```diff
@@ -0,0 +1,102 @@
...
+        key1: label1
+        key2: label2
+      annotations:
+        key1: label1
+        key2: label2
+    maxUnavailable: 1
+# ...
+----
+
+== Updating pods while keeping topics available
```
scholzj, line 51: As mentioned above, this still requires the PDB to be set to 0.

documentation/modules/upgrading/con-upgrade-cluster.adoc
```diff
@@ -0,0 +1,102 @@
...
+
+During an upgrade, you can trigger a rolling update of pods through the Cluster Operator.
+Using `Pod` resources, rolling updates restart the pods of resources with new pods.
```
scholzj, line 54: We should stress out here that the user needs to watch which pods need to be drained => and then annotate the correct pod.

documentation/modules/upgrading/con-upgrade-cluster.adoc
```diff
@@ -0,0 +1,102 @@
...
+
+A pod disruption budget allows only a specified number of pods to be unavailable at a given time.
+During planned maintenance of Kafka broker pods, a pod disruption budget ensures Kafka continues to run in a highly available environment.
+
+You specify a pod disruption budget using a `template` customization for a Kafka component.
+Labels and annotations identify resources.
+By default, pod disruption budgets allow only a single pod to be unavailable at a given time.
+You can increase the number of unavailable pods allowed by changing the default value of the `maxUnavailable` property.
+
+If you used the Cluster Operator to deploy Kafka, you can use the Strimzi Drain Cleaner tool to evict nodes during an upgrade.
```
scholzj, line 31: Do you expect to get to the Drain Cleaner docs soon? If not, can we add a link to the GitHUb repo as a stop-gap?

documentation/modules/upgrading/con-upgrade-cluster.adoc
PaulRMellor: This is our intro now:

image

documentation/modules/upgrading/con-upgrade-cluster.adoc
```diff
@@ -0,0 +1,102 @@
...
+
+== Configuring pod disruption budgets
+
+A pod disruption budget allows only a specified number of pods to be unavailable at a given time.
+During planned maintenance of Kafka broker pods, a pod disruption budget ensures Kafka continues to run in a highly available environment.
+
+You specify a pod disruption budget using a `template` customization for a Kafka component.
+Labels and annotations identify resources.
+By default, pod disruption budgets allow only a single pod to be unavailable at a given time.
+You can increase the number of unavailable pods allowed by changing the default value of the `maxUnavailable` property.
```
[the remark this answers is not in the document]
PaulRMellor: I've removed this

documentation/modules/upgrading/con-upgrade-cluster.adoc
PaulRMellor: Yes. I think I can start looking at that.

documentation/modules/upgrading/con-upgrade-cluster.adoc
```diff
@@ -0,0 +1,102 @@
...
+
+Refer to your Kubernetes distribution upgrade documentation to check the upgrade path and the steps to upgrade your nodes correctly.
+
+NOTE: Before upgrading, {supported-configurations}[check supported Kubernetes versions^].
+
+When performing your upgrade, you'll want to keep your Strimzi clusters available.
+You can employ the following strategies to do this.
+
+* Configuring pod disruption budgets
+* Updating pods while keeping topics available
```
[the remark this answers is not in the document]
PaulRMellor: All the drain cleaner content now sits under the _Configuring pod disruption budgets_

documentation/modules/upgrading/con-upgrade-cluster.adoc
```diff
@@ -0,0 +1,95 @@
...
+[role="_abstract"]
+If you are upgrading Kubernetes, refer the Kubernertes upgrade documentation to check the upgrade path and the steps to upgrade your nodes correctly.
+Before upgrading Kubernetes, link:{supported-configurations}[check the supported versions for your version of Strimzi^].
+
+When performing your upgrade, you'll want to keep your Kafka clusters available.
+
+You can employ one of the following strategies:
+
+* Configuring pod disruption budgets
+* Manually updating pods while keeping topics available
```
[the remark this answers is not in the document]
PaulRMellor: I changed to:

image

scholzj (approved):

documentation/modules/upgrading/con-upgrade-cluster.adoc
```diff
@@ -0,0 +1,107 @@
...
+kind: Kafka
+metadata:
+  name: my-cluster
+  namespace: myproject
+spec:
+  kafka:
+    # ...
+    template:
+      podDisruptionBudget:
+        maxUnavailable: 1
```
sknot-rh, line 75: I know this is just an example. However on line 59 you mention that value 0 should be set. Should we keep this consistent?

sknot-rh (approved):
One nit. LGTM otherwise

PR #5614, opened 2021-09-26
Support for External IPs (Kafka) and Persistent Storage (Kafka Connect)

lawwk98:
### Type of change

- Bugfix
X Enhancement / new feature
- Refactoring
- Documentation

### Description

Added external IPs to Kafka external listener -- both bootstrap and brokers, to enable cross cluster traffic.
Added persistent storage to Kafka Connect, to enable storage sharing across Connect replicas / pods.
Added test cases.

### Checklist

- [ X] Write tests
- [ X] Make sure all tests pass
- [ X] Update documentation
- [ X] Try your changes from Pod inside your Kubernetes and OpenShift cluster, not just locally
- [ ] Update CHANGELOG.md

### DIFF
--- a/api/src/main/java/io/strimzi/api/kafka/model/KafkaConnectSpec.java
+++ b/api/src/main/java/io/strimzi/api/kafka/model/KafkaConnectSpec.java
@@ -9,6 +9,7 @@ import com.fasterxml.jackson.annotation.JsonProperty;
 import com.fasterxml.jackson.annotation.JsonPropertyOrder;
 import io.strimzi.api.kafka.model.authentication.KafkaClientAuthentication;
 import io.strimzi.api.kafka.model.connect.build.Build;
+import io.strimzi.api.kafka.model.storage.SingleVolumeStorage;
 import io.strimzi.crdgenerator.annotations.Description;
 import io.strimzi.crdgenerator.annotations.DescriptionFile;
 import io.sundr.builder.annotations.Buildable;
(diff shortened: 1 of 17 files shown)


scholzj (changes requested):
First of all, thanks for the PR.

The support for persistent volumes in Kafka Connect has been discussed multiple times in the past. And there were many concerns about handling the storage with multiple replicas, and the limitations storage to supported storage. So I think the best way for that would be to start by opening a proposal (https://github.com/strimzi/proposals) which would allow us to clarify and discuss all the possible issues. You can check one of the existing approved proposals as an example of how such proposal might look like.

As for the external IPs ... not sure that would require a proposal, looks like it might not. But can you explain a bit more about it and the use-cases? In particular, how does it differ from `loadBalancerIP` we already support? How will the external IPs work with advertised hostnames in Kafka? It sounds like that would be an issue, that is why the value and use-case for this is not completely clear.

Apart from the comments above, also some general comments:
* The PR should be split into two PRs - each PR should focus on one thing only
* The mount path for the volumes should probably not be configurable -> at least not in the Kafka brokers and ZooKeeper
* The inheritance int he `api` module might need to be reworked to not have the volume configuration in the Mirror Maker 2 API
* Why not support JBOD storage in Connect? Are there any challenges with that? Having more volumes might be equally useful to having one volume
* Both features will probably require system test to make sure they will be working even in the future
* The Connect volumes would also require some documentation, possibly the external IPs as well.
* The DCO signoff would need to be fixed (the instructions are under the _Details_ link next to the DCO status in the PR status section)
* You might need to run `make crd_install` to apply the CRD changes also to the Helm Chart.

scholzj:
@lawwk98 Have you seen my comment? Do you plan to continue with this PR?

PR #5622, opened 2021-09-28
Update dependencies to deal with CVEs

tomncooper:
### Type of change

- Bugfix

### Description

Bumps dependencies to deal with the following CVEs:

- CVE-2020-13956
- CVE-2021-36090
- CVE-2021-35515
- CVE-2021-35517

### Checklist

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -119,14 +119,15 @@
         <opentracing-kafka.version>0.1.13</opentracing-kafka.version>
         <strimzi-oauth.version>0.8.1</strimzi-oauth.version>
         <commons-codec.version>1.13</commons-codec.version>
+        <commons-compress.version>1.21</commons-compress.version>
         <registry.version>1.3.2.Final</registry.version>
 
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -793,6 +794,12 @@
                 <artifactId>junit-platform-launcher</artifactId>
                 <version>${junit.platform.version}</version>
             </dependency>
+            <!-- transitive dep; override here to avoid buggy version -->
```
scholzj, line 797: Can you add which is using this as transitive dependency?

scholzj (approved):
One nit. LGTM otherwise.

pom.xml
tomncooper: Good point. We should probably remove these overrides at some point and see if they are still relevant?

tomncooper:
These test failures were due to timeouts. How do I re-run them?

tombentley (approved):

scholzj:
@tombentley The original regression is still running. Why did you start a new one?

PR #5623, opened 2021-09-28
Cold backup script fixes

fvaleri:
- Switch to ubi8 image
- Store backup data inside target dir instead of /tmp
- Ignore the sporadic file changed error
- Remove incremental backup option (not working)

These fixes are the result of a real use case, where the script was failing mainly because of the tar issue, but they also didn't have enough free space on /tmp that is used as staging area.

### DIFF
--- a/tools/cold-backup/README.md
+++ b/tools/cold-backup/README.md
@@ -1,10 +1,12 @@
 ## Strimzi backup
 
-Bash script for cold/offline backups of Kafka clusters on Kubernetes/OpenShift. Only local file system is supported. Make sure to have enough free space in the target directory.
+Bash script for offline (_cold_) backups of Kafka clusters on Kubernetes or OpenShift. The script only supports a local file system. Make sure you have enough free disk space in the target directory.
 
-If you think you do not need a backup strategy for Kafka because of its embedded data replication, then consider the impact of a misconfiguration, bug or security breach that deleted all your data. For hot/online backups, you can use storage snapshotting or streaming into object storage.
+If you think you don't need a backup strategy for Kafka because of its embedded data replication, then consider the impact of a misconfiguration, bug, or security breach that deletes all of your data. For online (_hot_) backups, you can use storage snapshotting or stream into object storage.
 
-To run the script, the Kubernetes user must have permission to work with PVC and Strimzi custom resources. The procedure will stop the Cluster Operator and selected cluster for the duration of the backup. Before restoring the Kafka cluster you need to make sure to have the right version of Strimzi CRDs installed. If you have a single cluster-wide Cluster Operator, then you need to scale it down manually. You can run backup and restore procedures for different Kafka clusters in parallel. Only the local file system is supported. Consumer group offsets are included, but not Kafka Connect, MirrorMaker and Kafka Bridge custom resources.
+Run the script as a Kubernetes user with permission to work with PVC and Strimzi custom resources. 
+
+The procedure will stop the Cluster Operator and selected cluster for the duration of the backup. Before restoring the Kafka cluster you need to make sure to have the right version of Strimzi CRDs installed. If you have a single cluster-wide Cluster Operator, then you need to scale it down manually. You can run backup and restore procedures for different Kafka clusters in parallel. Consumer group offsets are included, but not Kafka Connect, MirrorMaker and Kafka Bridge custom resources.
 
 ## Requirements
 
--- a/tools/cold-backup/README.md
+++ b/tools/cold-backup/README.md
@@ -1,18 +1,10 @@
-# Strimzi backup
+## Strimzi backup
 
-Bash script for cold/offline backups of Kafka clusters on Kubernetes/OpenShift.
-Only local file system is supported. Make sure to have enough free space in the target directory.
+Bash script for cold/offline backups of Kafka clusters on Kubernetes/OpenShift. Only local file system is supported. Make sure to have enough free space in the target directory.
 
-If you think you do not need a backup strategy for Kafka because of its embedded data replication, then consider the 
-impact of a misconfiguration, bug or security breach that deleted all your data. For hot/online backups, you can use 
-storage snapshotting or streaming into object storage.
+If you think you do not need a backup strategy for Kafka because of its embedded data replication, then consider the impact of a misconfiguration, bug or security breach that deleted all your data. For hot/online backups, you can use storage snapshotting or streaming into object storage.
 
-To run the script, the Kubernetes user must have permission to work with PVC and Strimzi custom resources. The procedure 
-will stop the Cluster Operator and selected cluster for the duration of the backup. Before restoring the Kafka cluster 
-you need to make sure to have the right version of Strimzi CRDs installed. If you have a single cluster-wide Cluster 
-Operator, then you need to scale it down manually. You can run backup and restore procedures for different Kafka 
-clusters in parallel. Consumer group offsets are included, but not Kafka Connect, MirrorMaker and Kafka Bridge custom 
-resources.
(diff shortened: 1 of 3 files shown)


tools/cold-backup/README.md
```diff
@@ -1,18 +1,21 @@
...
-restoring the Kafka cluster you need to make sure to have the right version of Strimzi CRDs installed.
-If you have a single cluster-wide Cluster Operator, then you need to scale it down manually. You can run
-backup and restore procedures for different Kafka clusters in parallel. Only the local file system is supported.
-Consumer group offsets are included, but not Kafka Connect, MirrorMaker and Kafka Bridge custom resources.
+To run the script, the Kubernetes user must have permission to work with PVC and Strimzi custom resources. The procedure 
+will stop the Cluster Operator and selected cluster for the duration of the backup. Before restoring the Kafka cluster 
+you need to make sure to have the right version of Strimzi CRDs installed. If you have a single cluster-wide Cluster 
+Operator, then you need to scale it down manually. You can run backup and restore procedures for different Kafka 
+clusters in parallel. Consumer group offsets are included, but not Kafka Connect, MirrorMaker and Kafka Bridge custom 
+resources.
```
scholzj, line 15: If you use sentence per line as we do in most other document files, you will avoid weird re-formatings like this which just make it harder to review.

tools/cold-backup/README.md
```diff
@@ -22,36 +25,35 @@ Consumer group offsets are included, but not Kafka Connect, MirrorMaker and Kafk
 - enough disk space
 
 ## Test procedure
+
 ```sh
+CLIENT_IMAGE="quay.io/strimzi/kafka:latest-kafka-2.8.0"
+krun() { kubectl run client -it --rm=true --restart=Never --image=$CLIENT_IMAGE -- $@; }
```
scholzj, line 31: I'm not sure I like this. This probably makes it much harder to understand what the test procedure does. I also don't understand why the image name should be defined in environment variable instead of directly in the field. `krun` is AFAIK also an alias many people already have for `kubectl run`, so it might conflict.

tools/cold-backup/README.md
```diff
@@ -22,36 +25,35 @@ Consumer group offsets are included, but not Kafka Connect, MirrorMaker and Kafk
 - enough disk space
 
 ## Test procedure
+
 ```sh
+CLIENT_IMAGE="quay.io/strimzi/kafka:latest-kafka-2.8.0"
```
scholzj, line 30: Should this used the `latest` image? Wouldn't it be better to use some released image which will not change?

tools/cold-backup/run.sh
```diff
@@ -117,9 +117,7 @@ __stop_cluster() {
 }
 
 __cleanup() {
-    if [[ $INCREMENTAL == false ]]; then
-        rm -rf $__TMP/$NAMESPACE/$CLUSTER_NAME
-    fi
+    rm -rf $BACKUP_PATH/$NAMESPACE/$CLUSTER_NAME
```
scholzj, line 120: Is this secure?

tools/cold-backup/templates/patch.json
```diff
@@ -2,8 +2,9 @@
     "spec": {
         "containers": [
             {
-                "name": "dummy",
-                "image": "centos:7",
+                "name": "strimzi-backup",
+                "image": "registry.access.redhat.com/ubi8/ubi:8.4",
```
scholzj, line 6: Would `ubi-minimal` image be sufficient? Or does it need the regular `ubi` image?

tools/cold-backup/README.md
fvaleri: Ok, will do. Anyway, I just added this phrase: Make sure to have enough free space in the target directory.

tools/cold-backup/README.md
fvaleri: Ok. I can revert that.

tools/cold-backup/README.md
fvaleri: Yep.

tools/cold-backup/templates/patch.json
fvaleri: No. There is no tar package in there.

tools/cold-backup/run.sh
fvaleri: Good point. We can replace all rm commands with an non-empty dir check.

fvaleri:
Thanks for the feedback @scholzj. All your concerns should be addressed now. I'm also using "sentence per line" approach, sorry about that.

scholzj (approved):
LGTM. I will run the STs on it.

PS: You actually seem to use paragraph per line not sentence per line ... but that is ok, at least it does not reformat everything to some line length, so that works for me.

tools/cold-backup/README.md
```diff
@@ -1,18 +1,13 @@
-# Strimzi backup
-Bash script for cold/offline backups of Kafka clusters on Kubernetes/OpenShift.
+## Strimzi backup
 
-If you think you do not need a backup strategy for Kafka because of its embedded data replication,
-then consider the impact of a misconfiguration, bug or security breach that deleted all your data.
-For hot/online backups, you can use storage snapshotting or streaming into object storage.
+Bash script for cold/offline backups of Kafka clusters on Kubernetes/OpenShift. Only local file system is supported. Make sure to have enough free space in the target directory.
```
PaulRMellor, line 3:
```suggestion
Bash script for offline (_cold_) backups of Kafka clusters on Kubernetes or OpenShift. The script only supports a local file system. Make sure you have enough free space in the target directory.
```

tools/cold-backup/README.md
```diff
@@ -1,18 +1,13 @@
...
 
-To run the script, the Kubernetes user must have permission to work with PVC and Strimzi custom resources.
-The procedure will stop the Cluster Operator and selected cluster for the duration of the backup. Before
-restoring the Kafka cluster you need to make sure to have the right version of Strimzi CRDs installed.
-If you have a single cluster-wide Cluster Operator, then you need to scale it down manually. You can run
-backup and restore procedures for different Kafka clusters in parallel. Only the local file system is supported.
-Consumer group offsets are included, but not Kafka Connect, MirrorMaker and Kafka Bridge custom resources.
+If you think you do not need a backup strategy for Kafka because of its embedded data replication, then consider the impact of a misconfiguration, bug or security breach that deleted all your data. For hot/online backups, you can use storage snapshotting or streaming into object storage.
```
PaulRMellor, line 5:
```suggestion
If you think you don't need a backup strategy for Kafka because of its embedded data replication, then consider the impact of a misconfiguration, bug, or security breach that deletes all of your data. For online (_hot_) backups, you can use storage snapshotting or stream into object storage.
```

tools/cold-backup/README.md
```diff
@@ -1,18 +1,13 @@
...
+
+To run the script, the Kubernetes user must have permission to work with PVC and Strimzi custom resources. The procedure will stop the Cluster Operator and selected cluster for the duration of the backup. Before restoring the Kafka cluster you need to make sure to have the right version of Strimzi CRDs installed. If you have a single cluster-wide Cluster Operator, then you need to scale it down manually. You can run backup and restore procedures for different Kafka clusters in parallel. Only the local file system is supported. Consumer group offsets are included, but not Kafka Connect, MirrorMaker and Kafka Bridge custom resources.
```
PaulRMellor, line 7:
```suggestion
Run the script as a Kubernetes user with permission to work with PVC and Strimzi custom resources. 

The procedure will stop the Cluster Operator and a specified Kafka cluster for the duration of the backup. Before restoring the Kafka cluster, you need to make sure to have the right version of Strimzi CRDs installed. If you have a single cluster-wide Cluster Operator, then you need to scale it down manually. You can run backup and restore procedures for different Kafka clusters in parallel. Consumer group offsets are included, but not Kafka Connect, MirrorMaker and Kafka Bridge custom resources.
```

PaulRMellor (approved):
Looks good. Few suggestions. _Only the local file system is supported_ was repeated, so I thought we could lose one.

fvaleri:
Thanks @PaulRMellor.

tools/cold-backup/README.md
fvaleri: ok

tools/cold-backup/README.md
fvaleri: ok

tools/cold-backup/README.md
fvaleri: ok

Frawless (approved):

PR #5626, opened 2021-09-29
[DOC] We should not suggest 1 partition is enough

scholzj:
### Type of change

- Bugfix
- Documentation

### Description

Right now, our [docs](https://strimzi.io/docs/operators/latest/full/using.html#topic_operator_topic_replication_and_scaling) suggest that one partition per topic is sufficient. I think this is wrong and we should not suggest that.

### Checklist

- [x] Update documentation

### DIFF
--- a/documentation/modules/operators/con-topic-operator-store.adoc
+++ b/documentation/modules/operators/con-topic-operator-store.adoc
@@ -69,7 +69,7 @@ spec:
     min.insync.replicas: 2 <3>
   #...
 ----
-<1> The number of partitions for the topic. Generally, 1 partition is sufficient.
+<1> The number of partitions for the topic.
 <2> The number of replica topic partitions. Currently, this cannot be changed in the `KafkaTopic` resource, but it can be changed using the `kafka-reassign-partitions.sh` tool.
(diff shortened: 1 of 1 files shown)


documentation/modules/operators/con-topic-operator-store.adoc
```diff
@@ -69,7 +69,7 @@ spec:
     min.insync.replicas: 2 <3>
   #...
 ----
-<1> The number of partitions for the topic. Generally, 1 partition is sufficient.
```
PaulRMellor, line 72: Do we want to change the value in the example?

PaulRMellor (approved):
Looks good. Maybe we want to change the value in the example.

documentation/modules/operators/con-topic-operator-store.adoc
PaulRMellor: Do we want to give any guidance on best practice for the number of partitions?

scholzj: > Do we want to give any guidance on best practice for the number of partitions?

I'm sure that would not hurt. But it is quite complex topic. So I would say it is way beyond this PR ;-)

tombentley (approved):

PR #5631, opened 2021-09-29
Ability to specify pull secret for KC builds on OpenShift

jcechace:
Enhancement providing means to specify the pull secret used to pull base image when performing KC builds as described in [ENTMQST-3282](https://issues.redhat.com/browse/ENTMQST-3282)

This is likely only good enough as a better description of desired behaviour.


scholzj (commented):
I had some more thoughts about this and the API change here.

On Kubernetes, we use the Kaniko project to build the container image. That actually uses the credentials provided in `/kaniko/.docker/config.json`. This is used for both pulling images and pushing images. We want to avoid merging different credentials (and more or less non-standardised JSONs). So I think the preferred way to deal with this is that the user uses the push secret to provide the credentials for both pulling and pushing. This can be credentials for different registries in one  secret if needed.

OCP Build Config uses separate pull and push secret. In addition, the pull secret might be required also when using ImageStream output which does not require push secret and does not have a field for it. So here, we cannot really use the push secret we with Kubernetes / Kaniko. But having a separate pull secret field in the build section which would work only on OCP seems weird. I think we should put this into the existing template for the BuildConfig (`.spec.template.buildConfig`) => that way it would be specific for the OCP build and not confuse Kubernetes users. And given this seems to be rarely used, I think having it in template should not be a big issue.

@strimzi/maintainers WDYT?

@jcechace Let's see what others think about it. If needed, I can do the work as well.

Frawless:
@scholzj your proposed solution sounds better to me

jcechace:
@scholzj  sounds all right. The ``buildConfig`` field currently uses``ResourceTemplate`` type which can have only metadata. Does it mean a new type should be created?

scholzj:
@jcechace Yes, it would require a new class for BuildConfig template.

scholzj:
@strimzi/maintainers This should be now ready for review.

PR #5634, opened 2021-09-30
Set KStreams RF

sknot-rh:

### Type of change
- Enhancement

### Description
The `REPLICATION_FACTOR_CONFIG` is never set and thus it is always set to broker's default. Setting this value to `min(kafkaReplicas, 3)`. Value 3 is from this [recommendation](https://kafka.apache.org/22/documentation/streams/developer-guide/config-streams#replication-factor).

### Checklist

- [ ] Write tests
- [ ] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/api/src/main/java/io/strimzi/api/kafka/model/EntityTopicOperatorSpec.java
+++ b/api/src/main/java/io/strimzi/api/kafka/model/EntityTopicOperatorSpec.java
@@ -151,7 +151,7 @@ public class EntityTopicOperatorSpec implements UnknownPropertyPreserving, Seria
     }
 
 
-    @Description("Logging configuration")
+    @Description("Logging configuration.")
     @JsonInclude(value = JsonInclude.Include.NON_NULL)
     public Logging getLogging() {
         return logging;
--- a/api/src/main/java/io/strimzi/api/kafka/model/EntityUserOperatorSpec.java
+++ b/api/src/main/java/io/strimzi/api/kafka/model/EntityUserOperatorSpec.java
@@ -129,7 +129,7 @@ public class EntityUserOperatorSpec implements UnknownPropertyPreserving, Serial
         this.readinessProbe = readinessProbe;
     }
 
-    @Description("Logging configuration")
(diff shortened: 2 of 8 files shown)


scholzj (approved):

alesj:
I remember we had an issue where this wasn't set to -1 (hence wasn't set to broker's defaults ...)
@tomncooper do you remember where was this ^^
(as we set this afterwards ...)

scholzj:
```
[ERROR] Failures: 
[ERROR] io.strimzi.systemtest.metrics.MetricsST.testKafkaTopicPartitions
[ERROR]   Run 1: MetricsST.testKafkaTopicPartitions:133 Topic partitions count doesn't match expected value
Expected: is <423.0>
     but: was <425.0>
[ERROR]   Run 2: MetricsST.testKafkaTopicPartitions:133 Topic partitions count doesn't match expected value
Expected: is <423.0>
     but: was <425.0>
[ERROR]   Run 3: MetricsST.testKafkaTopicPartitions:133 Topic partitions count doesn't match expected value
Expected: is <423.0>
```

Looks like some STs to fix @sknot-rh

tomncooper:
We were setting it to 1 which was obviously no great, so we changed it to be the broker default topic RF (`-1`) which strikes me as a better default than `3` which a customer may not want (for whatever reason). The best option IMHO is to set the default streams RF to `-1` unless the user set the config explicitly themselves.

scholzj:
@tomncooper The cluster default for RF is 1. So `-1` will in most cases default to `1`. Is that the desired number of replicas for this topic?

alesj:
Hmm, now as I see it from the code, it will default to 3 in Strimzi ... I might be wrong reading this snippets though ...

scholzj:
Discussed on the community call. This should be closed. As discussed, we want to get the users to set the default setting in their clusters properly and not try to solve this issue in this place.

@sknot-rh Does that make sense? Are you fine with it?

sknot-rh:
Sounds good to me.

PR #5650, opened 2021-10-04
[systemtest] Add test for advertised hostnames + ports in internal and external listeners

im-konge:

### Type of change

- New test

### Description

We should have test checking if added advertised host names and ports will also appear to SAN of each broker `.crt` certificate -> for both external and internal listeners (with TLS). This PR also checking changes from #5460 .

### Checklist

- [x] Write test

### DIFF
--- a/systemtest/src/test/java/io/strimzi/systemtest/kafka/listeners/ListenersST.java
+++ b/systemtest/src/test/java/io/strimzi/systemtest/kafka/listeners/ListenersST.java
@@ -2164,7 +2164,6 @@ public class ListenersST extends AbstractST {
     }
 
     @ParallelNamespaceTest
-    @Tag(NODEPORT_SUPPORTED)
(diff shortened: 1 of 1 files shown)


systemtest/src/test/java/io/strimzi/systemtest/kafka/listeners/ListenersST.java
```diff
@@ -2155,6 +2163,113 @@ void testMessagesTlsScramShaWithPredefinedPassword(ExtensionContext extensionCon
         );
     }
 
+    @ParallelNamespaceTest
+    @Tag(NODEPORT_SUPPORTED)
```
scholzj, line 2167: You are not really connecting to it. So I think you don't needs this. the nodeports are always supported by Kube ... just sometimes you cannot connect to them.

scholzj (commented):
One nit. LGTM otherwise.

scholzj (approved):

Frawless (approved):

PR #5651, opened 2021-10-04
[systemtest] Change ST default kafka version to 3.0.0

im-konge:

### Type of change

- Bugfix

### Description

This PR fixes issue with hard-coded default version of Kafka used in STs and disables tests containing issues, which will be fixed in future PRs.

### Checklist

### DIFF
--- a/systemtest/src/main/java/io/strimzi/systemtest/Environment.java
+++ b/systemtest/src/main/java/io/strimzi/systemtest/Environment.java
@@ -134,7 +134,7 @@ public class Environment {
     /**
      * Defaults
      */
-    private static final String ST_KAFKA_VERSION_DEFAULT = TestKafkaVersion.getLatestSupportedKafkaVersion();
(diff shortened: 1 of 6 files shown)


scholzj:
Does this mean that all the system tests till now were using Kafka 2.8.0?

scholzj:
Or what exactly does this flag do?

im-konge:
Yes, unfortunately. If the `ST_KAFKA_VERSION` env is not set, the default is used - until now `2.8.0`. Maybe it would be better to always get latest version instead of hardcoding it here.

scholzj:
That is fucked up. Why does it not use the Kafka version from the example files? It takes almost everything from there. I think this needs to be rewritten to either take the version from examples or use the default Kafka version as default.

im-konge:
Yes, that was what I was thinking about. Let me rewrite it a bit.

Frawless (approved):
LGTM

PR #5657, opened 2021-10-04
[MO] - elimination test cases, which repeat code path (CFG) in dyn. conf

see-quick:

### Type of change

- Enhancement / new feature
- Refactoring

### Description

This PR eliminates most of the Kafka Dyn. configurations, which we currently test in `DynamicConfigurationSharedST`. It selects 3 test cases randomly and then verifies how we use them. We also discussed offline with @sknot-rh, that this approach is fine because the way we tested it was just repeating the same path of code, which is only time-consuming.

For more details see `@stochasticSelection()` method.

For imagination on Azure, these tests consume 2h of execution time see [1] and with this change, this test suite will last only for 6-8 minutes.

[1] image

### Checklist

- [ ] Write tests


scholzj (commented):
I'm not sure I like the randomness area. It means that failures will not be reproducible. We tend to often just restart tests after failure, so it can easily mask the errors. Also, aren't the STs restarted 3 times even directly in a single test run? So again, that would tend to just hide issues.

see-quick:
I see your point @scholzj but the failure arises only if there is a problem with the Dynamic configuration implementation. For instance, after we make an update the Kafka Pods will rollout... It does not matter, which value of Kafka dynamic configuration we set from my POV. So it makes sense to do this random stuff. And I know your point that the variant values could have different consequences but that's bad I think...Semantically, they must have similar in the context of Dynamic configuration...

> Also, aren't the STs restarted 3 times even directly in a single test run.
Yes, they do.

systemtest/src/test/java/io/strimzi/systemtest/kafka/dynamicconfiguration/DynamicConfigurationSharedST.java
```diff
@@ -179,6 +186,41 @@
...
+    /**
+     * Method, which randomly choose 3 dynamic properties for verification from @see{testCases}. In this case we are ok
+     * with stochastic selection, because we don't care, which configuration is used. Furthermore, it's the same path
+     * of code (i.e., same CFG (control flow graph), which does not triggers RollingUpdate).
+     * @param testCases test cases, where each consist of one dynamic property
+     * @return List 3 chosen dynamic properties
+     */
+    private static List<String> stochasticSelection(final Map<String, Object> testCases) {
+        final List<String> testCaseKeys = new ArrayList<>(testCases.keySet());
+        final Random generator = new Random();
```
sknot-rh, line 198: Shouldn't this be a class member?

systemtest/src/test/java/io/strimzi/systemtest/kafka/dynamicconfiguration/DynamicConfigurationSharedST.java
```diff
@@ -179,6 +186,41 @@
...
+            int stochasticNumber = generator.nextInt(testCaseKeys.size());
+            String chosenDynConfiguration = testCaseKeys.get(stochasticNumber);
+
+            if (chosenDynConfigurations.contains(chosenDynConfiguration)) {
+                while (chosenDynConfigurations.contains(chosenDynConfiguration)) {
+                    LOGGER.debug("List of dynamic configurations already have that type of configuration:\n{}", chosenDynConfigurations.toArray());
+                    stochasticNumber = generator.nextInt(testCaseKeys.size());
+                    chosenDynConfiguration = testCaseKeys.get(stochasticNumber);
+
+                    chosenDynConfigurations.add(chosenDynConfiguration);
```
sknot-rh, line 211: So you are sure you won't pick the same (or already present) key twice?

systemtest/src/test/java/io/strimzi/systemtest/kafka/dynamicconfiguration/DynamicConfigurationSharedST.java
see-quick: In this case, it does not matter because this pseudo-generator is instantiated only once and does not breaks the mathematical foundation. (i.e., do not depends on seed, which is not random). But I will move it to class if in the future one will use it in another part of code...

systemtest/src/test/java/io/strimzi/systemtest/kafka/dynamicconfiguration/DynamicConfigurationSharedST.java
see-quick: I am 100% sure, but its greedy algorithm (which is not good), because it can happen that will iterate over the collection of Dyn. Confs and pick the same twice (but it will not add to the list), which results in another iteration in a while loop. Re-writed to:
```
final int stochasticNumber = rng.nextInt(testCaseKeys.size());
final String chosenDynConfiguration = testCaseKeys.get(stochasticNumber);
LOGGER.debug("New configuration in List of dynamic configuration:\n{}", chosenDynConfigurations.toArray());
chosenDynConfigurations.add(chosenDynConfiguration);
// remove it from `testCaseKeys` list to not include it twice
testCaseKeys.remove(chosenDynConfiguration);
```

Frawless:
In case the execution takes 2 hours, does it mean in case of failure it will take for example 4 hours? or only the failed one will be executed?

see-quick:
> In case the execution takes 2 hours, does it mean in case of failure it will take for example 4 hours? or only the failed one will be executed?

Yes, that's true. (but it is `mvn rerun` functionality)

Frawless (approved):

sknot-rh (approved):

PR #5659, opened 2021-10-05
[systemtest] Fixes for tests and add `max.poll.records` property to consumer

im-konge:

### Type of change

- Test fixes

### Description

This PR fixes flaky/failing tests.

I'm adding `max.poll.records` to consumer properties, as we need to use it in some testcases, where we want to poll more than 500 messages and I'm also adding finalizers to tests where we are using LoadBalancers and didn't contain it already.

### Checklist


Frawless (approved):

scholzj (commented):
Can you please explain why this should be needed? Why do you need to get 1000 records in a single poll? This sound more like some issue with how the client is written. WDYT @tombentley?

im-konge:
From MPOV if you want to check, that you can consume all messages already produced, with (for example) new certificate and you know that you produced more than 500 messages (which is default value for `max.poll.records`), you'll just change that property. 

If we check less than sent number of messages (so we produce 800 messages and we will try to consume 500), there can be possible issues, which we will not find.

Or is my opinion wrong?

im-konge:
Or do you mean polling 1000 messages in stacks - like 10 messages per poll? And poll all 1000 messages in 100 polls?

scholzj:
So if you would need to consume 100000000 messages you would set it to 100000000? You would normally just call the poll in a loop instead.

im-konge:
Yeah how I mentioned in next comment, I get it. Polling all messages in one poll doesn't make sense. 

Well I guess I can use here our test-clients finally or fix the clients used in tests for now.

Thanks @scholzj for spotting this.

scholzj (approved):

PR #5685, opened 2021-10-08
[systemtest] Add check for pausing KafkaConnector

im-konge:

### Type of change

- New test check

### Description

Because we don't have a check for pausing `KafkaConnector`, I'm adding it to already existing test - we will deploy `KafkaConnector` with `pause: true` in `spec`,  producer/consume messages and wait for messages will appear to text file - which will not happen because of paused connector. After replacing it to `pause: false`, messages should appear to file sink file.

### Checklist


im-konge:
@strimzi-ci run tests test_only profile=regression testcase=ConnectST

im-konge:
@strimzi-ci run tests test_only profile=regression testcase=ConnectST

PR #5693, opened 2021-10-09
[MO] - Azure balance our pipelines execution time

see-quick:

### Type of change

- Enhancement / new feature

### Description

This PR balance our regression pipelines (execution time). I have compares all times for Test suites and somehow re-arranged them in the best way possible. Result:

image

Note:
After the OPA issue is resolved, we can enable this test suite and it will increase the execution time of the First pipeline which is currently the most un-balanced but still ok from MPOV.
image

### Checklist


see-quick:
I checked manually all regressions...everything is as excepted!

Issue #5594, opened 2021-09-21
Tool to get log dumps


tombentley:
`report.sh` uses `kubectl` (or `oc` if `kubctl` is not installed) to get broker and zookeeper logs for attaching to bug reports. 
When investigating problems around transactions or consumer groups it can be very helpful to get dumps of the `__transaction_state` and `__consumer_offsets` topics, as well as the relevant application partitions. I think it would be way, way too much to always include these for every partition in the cluster, but it would be good if there was a tool (possiblly `report.sh` or possibly something else) which supported a way of selecting partitions whose logs should be dumped (possibly between given times, to reduce the amount of data returned). For example, something like `report.sh --namespace=mynamespace --cluster=mycluster --transactionalId=mytransactionalid --after <date> --before <date> --partition=mypartition-1`. This would figure out the `__transaction_state` partition and dump the logs from brokers replicating that, as well as for `mypartition-1`. Similarly `--groupId` for consumer groups. There could be multiple occurrances of `--partition`.

tombentley:
To avoid having a single issue about two separate ideas I've opened #5603 for the first point. I'll amend the description and summary so this issue is specifically about the second idea.

fvaleri:
I also think that a separate tool would be better. This tool could also provide some statistics on these files and identify pending and empty transactions. All of this would help the analysis phase in case of issues. Finally, we would also need to specify that payloads are not included in these dumps, but only metadata.

tombentley:
@fvaleri I think there are two things which could be improved: 

1. The extraction of log dumps (this issue)
2. The later analysis of those dumps, which often also requires correlation with broker logs. 

I think 2 is probably a different tool.

PR #5694, opened 2021-10-11
Add Kafka log dump tool

fvaleri:
This is related to issue #5594.

I'm also attaching the output after running 1 transaction with this [test application](https://github.com/fvaleri/integ-examples/tree/main/misc/kafka/kafka-trans) on a new cluster.

We may need to add a system test, but I first wanted to get some feedback.

[my-dump.zip](https://github.com/strimzi/strimzi-kafka-operator/files/7320347/my-dump.zip)

### DIFF
--- a/tools/log-dump/run.sh
+++ b/tools/log-dump/run.sh
@@ -246,20 +246,23 @@ Usage: $0 [command] [params]
     --namespace     Kubernetes namespace
     --cluster       Kafka cluster name
     --group-id      Consumer group id
-    --tot-part      Consumer offsets partitions (default: $TOT_PART)
+    --tot-part      Number of __consumer_offsets partitions (default: $TOT_PART)
     --out-path      Output path (default: $OUT_PATH)
+    --dry-run       Run without dumping (default: $DRY_RUN)
   
   txn_state       Dump transactions state by transactional.id
     --namespace     Kubernetes namespace
     --cluster       Kafka cluster name
     --txn-id        Transactional id
-    --tot-part      Transaction state partitions (default: $TOT_PART)
+    --tot-part      Number of __transaction_state partitions (default: $TOT_PART)
     --out-path      Output path (default: $OUT_PATH)
+    --dry-run       Run without dumping (default: $DRY_RUN)
     
   cluster_meta    Dump cluster metadata (KRaft)
     --namespace     Kubernetes namespace
     --cluster       Kafka cluster name
     --out-path      Output path (default: $OUT_PATH)
+    --dry-run       Run without dumping (default: $DRY_RUN)
 "
 readonly PARAMS="${@}"
 readonly PARRAY=($PARAMS)
--- a/tools/log-dump/run.sh
+++ b/tools/log-dump/run.sh
@@ -54,6 +54,17 @@ get_kafka_setup() {
   fi
 }
 
+coord-partition() {
+  local id="$1"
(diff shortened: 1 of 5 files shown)


fvaleri:
@tombentley @scholzj

fvaleri:
@tombentley what we want to do with this one?

tombentley:
On reflection, I think it would be best to take out the dependency on `klog`, and just let people supply the partitions themselves. It makes it slightly less easy to use, but I don't really want to do a release of `klog` at this point, and having this tool depend on some binary that someone's had to build themselves doesn't feel like a great UX either.

fvaleri:
Understood, but I don't like the idea of removing that functionality, so in my last commit I'm proposing what could be a cleaner alternative.

tools/log-dump/run.sh
```diff
@@ -0,0 +1,340 @@
...
+    --segment       Filter by segment name
+    --out-path      Output path (default: $OUT_PATH)
+    --dry-run       Run without dumping (default: $DRY_RUN)
+    --data          Include record payload (default: $DATA)
+
+  cg_offsets      Dump consumer group offsets by group.id
+    --namespace     Kubernetes namespace
+    --cluster       Kafka cluster name
+    --group-id      Consumer group id
+    --tot-part      Consumer offsets partitions (default: $TOT_PART)
```
tombentley, line 249: From this description I don't really know what `tot-part` is supposed to _be_.
```suggestion
    --tot-part      The number of __consumer_offsets partitions (default: $TOT_PART)
```

tools/log-dump/run.sh
```diff
@@ -0,0 +1,340 @@
...
+    --out-path      Output path (default: $OUT_PATH)
+  
+  txn_state       Dump transactions state by transactional.id
+    --namespace     Kubernetes namespace
+    --cluster       Kafka cluster name
+    --txn-id        Transactional id
+    --tot-part      Transaction state partitions (default: $TOT_PART)
```
tombentley, line 256: Same comment, 
```suggestion
    --tot-part      The number of __transaction_state partitions (default: $TOT_PART)
```

tools/log-dump/run.sh
```diff
@@ -0,0 +1,340 @@
...
+Usage: $0 [command] [params]
+
+  partition       Dump partition
+    --namespace     Kubernetes namespace
+    --cluster       Kafka cluster name
+    --topic         Data topic name
+    --partition     Partition number (zero-based)
+    --segment       Filter by segment name
+    --out-path      Output path (default: $OUT_PATH)
+    --dry-run       Run without dumping (default: $DRY_RUN)
```
tombentley, line 242: Why is it only this subcommand that has `--dry-run`?

tombentley (approved):
A few last nits, but I think using `jshell` is a good compromise here.

tools/log-dump/run.sh
fvaleri: ok

tools/log-dump/run.sh
fvaleri: ok

tools/log-dump/run.sh
fvaleri: Good catch. In reality it is implemented for all commands. Let me fix the help.

scholzj (approved):

PR #5774, opened 2021-10-25
[MO] - @ParallelSuites -> CruiseControl support 

see-quick:

### Type of change

- Bugfix
- Enhancement / new feature

### Description

This PR adds a few things:

1. @ParallelSuite (support to run cruise control package in parallel mode with other suites)
2. fixing bad usage of @ParallelTest annotation (In `CruiseControlApiST` one created Kafka cluster in `INFRA_NAMESPACE` and does not use @ParalllelNamespaceTest as we do in all cases)

### Checklist

- [x] Write tests

### DIFF
--- a/systemtest/src/main/java/io/strimzi/systemtest/BeforeAllOnce.java
+++ b/systemtest/src/main/java/io/strimzi/systemtest/BeforeAllOnce.java
@@ -7,6 +7,7 @@ package io.strimzi.systemtest;
 import io.strimzi.systemtest.parallel.ParallelNamespacesSuitesNames;
 import io.strimzi.systemtest.parallel.ParallelSuiteController;
 import io.strimzi.systemtest.resources.operator.SetupClusterOperator;
+import io.strimzi.systemtest.utils.StUtils;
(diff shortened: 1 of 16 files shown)


see-quick:
Hmmm, I found out that if we execute `cruisecontrol` package (in `RBAC=NAMESPACE` mode) some of them fails because the Kubernetes namespace context is set to `INFRA_NAMESPACE` and not `NAMESPACE` for the specific test suite. Here are two solutions for that:

1. Set `final String namespaceName = StUtils.getNamespaceBasedOnRbac(Constants.NAMESPACE, extensionContext);` to the final String namespaceName = StUtils.getNamespaceBasedOnRbac(Constants.INFRA_NAMESPACE, extensionContext);
2. Implement a mechanism that we always use `namespace-<counter-int>` in @ParallelNamespaceTest that we always copy new RoleBinding and with that ensures that tests can run in parallel.

see-quick:
Rebase -> https://github.com/strimzi/strimzi-kafka-operator/pull/5826 after this is merged

Issue #5757, opened 2021-10-21
Listener in status reporting name as type of listener


ppatierno:
It seems that the listeners information in the `status.listeners` of the `Kafka` CR show the name of the listener but in the `type` field.
For example, the `spec` has:

```yaml
listeners:
    - name: plain
      port: 9092
      tls: false
      type: internal
    - name: tls
      port: 9093
      tls: true
      type: internal
```

the `status` has:

```yaml
listeners:
  - addresses:
    - host: my-cluster-kafka-bootstrap.kafka.svc
      port: 9092
    bootstrapServers: my-cluster-kafka-bootstrap.kafka.svc:9092
    type: plain
  - addresses:
    - host: my-cluster-kafka-bootstrap.kafka.svc
      port: 9093
    bootstrapServers: my-cluster-kafka-bootstrap.kafka.svc:9093
    certificates:
    - |
      -----BEGIN CERTIFICATE-----
      ...
      -----END CERTIFICATE-----
    type: tls
```

so from listener `name: plain` in spec we have `type: plain` in status instead of reflecting `type: internal`.
We should have a new `name` field in the listener status API definition in order to bring the right field from listener declaration and setting the `type` to the actual internal, route, loadbalancer and so on.
At some point we could decide to deprecate and remove the `type` field (it's enough from the name in the listener status check the corresponding type from the listener spec declaration) or just leave it.

PR #5777, opened 2021-10-26
Listener in status reporting name as type of listener

ShubhamRwt:

### Type of change

_Select the type of your PR_

- Enhancement 

### Description

This PR addresses the issue mentioned in #5757. This PR adds the `name` field in the listener status which have the same as the field `type`.

### Checklist

- [ ] Write tests
- [ ] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@
 * Update Open Policy Agent authorizer to 1.4.0 and add support for enabling metrics 
 * Added the option `createBootstrapService` in the Kafka Spec to disable the creation of the bootstrap service for the Load Balancer Type Listener. It will save the cost of one load balancer resource, specially in the public cloud.
 
+### Changes, deprecations and removals
+
+* The `type` field in `ListenerStatus` has been deprecated and will be removed in the future.
+
 ## 0.27.0
 
 * Multi-arch container images with support for x86_64 / AMD64 and AArch64 / ARM64 platforms
--- a/api/src/main/java/io/strimzi/api/kafka/model/status/ListenerStatus.java
+++ b/api/src/main/java/io/strimzi/api/kafka/model/status/ListenerStatus.java
@@ -6,6 +6,7 @@ package io.strimzi.api.kafka.model.status;
 
 import com.fasterxml.jackson.annotation.JsonInclude;
 import com.fasterxml.jackson.annotation.JsonPropertyOrder;
+import io.strimzi.api.annotations.DeprecatedProperty;
 import io.strimzi.api.kafka.model.Constants;
 import io.strimzi.api.kafka.model.UnknownPropertyPreserving;
 import io.strimzi.crdgenerator.annotations.Description;
@@ -28,25 +29,36 @@ import static java.util.Collections.emptyMap;
         builderPackage = Constants.FABRIC8_KUBERNETES_API
 )
 @JsonInclude(JsonInclude.Include.NON_NULL)
-@JsonPropertyOrder({ "type", "addresses", "bootstrapServers", "certificates" })
+@JsonPropertyOrder({ "type", "name", "addresses", "bootstrapServers", "certificates" })
 @EqualsAndHashCode
 public class ListenerStatus implements UnknownPropertyPreserving, Serializable {
     private static final long serialVersionUID = 1L;
 
-    private String type;
+    private String name;
     private List<ListenerAddress> addresses;
     private String bootstrapServers;
     private List<String> certificates;
     private Map<String, Object> additionalProperties;
 
-    @Description("The type of the listener. " +
-            "Can be one of the following three types: `plain`, `tls`, and `external`.")
+    @Description("The name of the listener.")
+    @DeprecatedProperty(movedToPath = "name")
+    @Deprecated
     public String getType() {
-        return type;
+        return name;
     }
 
-    public void setType(String type) {
-        this.type = type;
+    @Deprecated
+    public void setType(String name) {
+        this.name = name;
+    }
+
+    @Description("The name of the listener.")
+    public String getName() {
+        return name;
+    }
+
+    public void setName(String name) {
+        this.name = name;
     }
 
     @Description("A list of the addresses for this listener.")
--- a/api/src/test/java/io/strimzi/api/kafka/model/KafkaTest.java
+++ b/api/src/test/java/io/strimzi/api/kafka/model/KafkaTest.java
@@ -8,10 +8,15 @@ import io.fabric8.kubernetes.api.model.ObjectMeta;
 import io.strimzi.api.kafka.model.listener.arraylistener.GenericKafkaListener;
 import io.strimzi.api.kafka.model.listener.arraylistener.GenericKafkaListenerBuilder;
 import io.strimzi.api.kafka.model.listener.arraylistener.KafkaListenerType;
+import io.strimzi.api.kafka.model.status.ConditionBuilder;
+import io.strimzi.api.kafka.model.status.ListenerAddressBuilder;
+import io.strimzi.api.kafka.model.status.ListenerStatus;
+import io.strimzi.api.kafka.model.status.ListenerStatusBuilder;
 import io.strimzi.test.TestUtils;
 import org.junit.jupiter.api.Test;
 
 import java.net.URISyntaxException;
+import java.text.ParseException;
 import java.util.Collections;
 import java.util.List;
 import java.util.Objects;
@@ -130,6 +135,75 @@ public class KafkaTest extends AbstractCrdTest<Kafka> {
 
         assertThat(listeners.get(0).getAuth().getType(), is("scram-sha-512"));
         assertThat(listeners.get(1).getAuth().getType(), is("tls"));
+    }
+
+    @Test
+    public void testListenerTypeAndNameInStatus() throws ParseException, URISyntaxException {
+
+        Kafka kafka = new KafkaBuilder()
+                .withNewMetadata()
+                    .withName("my-cluster")
+                    .withNamespace("my-namespace")
+                    .withGeneration(2L)
+                .endMetadata()
+                .withNewSpec()
+                    .withNewKafka()
+                        .withReplicas(3)
+                        .withListeners(new GenericKafkaListenerBuilder()
+                            .withName("plain")
+                            .withPort(9092)
+                            .withType(KafkaListenerType.INTERNAL)
+                            .withTls(false)
+                            .build())
+                        .withNewEphemeralStorage()
+                        .endEphemeralStorage()
+                    .endKafka()
+                    .withNewZookeeper()
+                         .withReplicas(3)
+                         .withNewEphemeralStorage()
+                         .endEphemeralStorage()
+                    .endZookeeper()
+                .endSpec()
+                .withNewStatus()
+                    .withObservedGeneration(1L)
+                    .withConditions(new ConditionBuilder()
+                            .withType("Ready")
+                            .withStatus("True")
+                            .build())
+                    .withListeners(new ListenerStatusBuilder()
+                                   .withName("plain")
+                                   .withAddresses(new ListenerAddressBuilder()
+                                           .withHost("my-service.my-namespace.svc")
+                                           .withPort(9092)
+                                           .build())
+                                   .build(),
+                        new ListenerStatusBuilder()
+                                   .withName("external")
+                                   .withAddresses(new ListenerAddressBuilder()
+                                           .withHost("my-route-address.domain.tld")
+                                           .withPort(443)
+                                           .build())
+                                   .build())
+                .endStatus()
+                .build();
+
+        String path = Objects.requireNonNull(this.getClass().getResource("Kafka-listener-name-and-status.yaml")).toURI().getPath();
+        assertThat(TestUtils.toYamlString(kafka), is(TestUtils.getFileAsString(path)));
+    }
+
+    @Test
+    public void testListenersTypeAndName()    {
+        Kafka model = TestUtils.fromYaml("Kafka-listener-name-and-status" + ".yaml", Kafka.class);
+
+        assertThat(model.getStatus().getListeners(), is(notNullValue()));
+        assertThat(model.getStatus().getListeners().size(), is(2));
+
+        List<ListenerStatus> listeners = model.getStatus().getListeners();
+
+        assertThat(listeners.get(0).getType(), is("plain"));
+        assertThat(listeners.get(1).getType(), is("external"));
+        assertThat(listeners.get(0).getName(), is("plain"));
+        assertThat(listeners.get(1).getName(), is("external"));
     }
 
     public void rt(String resourceName) {
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaAssemblyOperator.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaAssemblyOperator.java
@@ -2111,7 +2111,7 @@ public class KafkaAssemblyOperator extends AbstractAssemblyOperator<KubernetesCl
                 String bootstrapAddress = getInternalServiceHostname(ListenersUtils.backwardsCompatibleBootstrapServiceName(name, listener), useServiceDnsDomain);
 
                 ListenerStatus ls = new ListenerStatusBuilder()
-                        .withType(listener.getName())
+                        .withName(listener.getName())
                         .withAddresses(new ListenerAddressBuilder()
                                 .withHost(bootstrapAddress)
                                 .withPort(listener.getPort())
@@ -2233,7 +2233,7 @@ public class KafkaAssemblyOperator extends AbstractAssemblyOperator<KubernetesCl
                     return CompositeFuture.join(perPodFutures);
                 }).compose(res -> {
                     ListenerStatus ls = new ListenerStatusBuilder()
-                        .withType(listener.getName())
+                        .withName(listener.getName())
                         .withAddresses(bootstrapListenerAddressList.stream()
                                 .map(listenerAddress -> new ListenerAddressBuilder().withHost(listenerAddress)
                                         .withPort(listener.getPort())
@@ -2399,7 +2399,7 @@ public class KafkaAssemblyOperator extends AbstractAssemblyOperator<KubernetesCl
                                 }
 
                                 ListenerStatus ls = new ListenerStatusBuilder()
-                                        .withType(listener.getName())
+                                        .withName(listener.getName())
                                         .withAddresses(new ArrayList<>(statusAddresses))
                                         .build();
                                 addListenerStatus(ls);
@@ -2438,7 +2438,7 @@ public class KafkaAssemblyOperator extends AbstractAssemblyOperator<KubernetesCl
                             kafkaBootstrapDnsName.add(bootstrapAddress);
 
                             ListenerStatus ls = new ListenerStatusBuilder()
-                                    .withType(listener.getName())
+                                    .withName(listener.getName())
                                     .withAddresses(new ListenerAddressBuilder()
                                             .withHost(bootstrapAddress)
                                             .withPort(kafkaCluster.getRoutePort())
@@ -2523,7 +2523,7 @@ public class KafkaAssemblyOperator extends AbstractAssemblyOperator<KubernetesCl
                             kafkaBootstrapDnsName.add(bootstrapAddress);
 
                             ListenerStatus ls = new ListenerStatusBuilder()
-                                    .withType(listener.getName())
+                                    .withName(listener.getName())
                                     .withAddresses(new ListenerAddressBuilder()
                                             .withHost(bootstrapAddress)
                                             .withPort(kafkaCluster.getRoutePort())
@@ -2601,7 +2601,7 @@ public class KafkaAssemblyOperator extends AbstractAssemblyOperator<KubernetesCl
                             kafkaBootstrapDnsName.add(bootstrapAddress);
 
                             ListenerStatus ls = new ListenerStatusBuilder()
-                                    .withType(listener.getName())
+                                    .withName(listener.getName())
                                     .withAddresses(new ListenerAddressBuilder()
                                             .withHost(bootstrapAddress)
                                             .withPort(kafkaCluster.getRoutePort())
@@ -4150,7 +4150,7 @@ public class KafkaAssemblyOperator extends AbstractAssemblyOperator<KubernetesCl
             List<ListenerStatus> listeners = kafkaStatus.getListeners();
 
             if (listeners != null) {
-                ListenerStatus listener = listeners.stream().filter(listenerType -> type.equals(listenerType.getType())).findFirst().orElse(null);
+                ListenerStatus listener = listeners.stream().filter(listenerType -> type.equals(listenerType.getName())).findFirst().orElse(null);
 
                 if (listener != null) {
                     listener.setCertificates(singletonList(certificate));
--- a/packaging/helm-charts/helm3/strimzi-kafka-operator/crds/040-Crd-kafka.yaml
+++ b/packaging/helm-charts/helm3/strimzi-kafka-operator/crds/040-Crd-kafka.yaml
@@ -5956,7 +5956,10 @@ spec:
                     properties:
                       type:
                         type: string
-                        description: 'The type of the listener. Can be one of the following three types: `plain`, `tls`, and `external`.'
+                        description: The name of the listener.
+                      name:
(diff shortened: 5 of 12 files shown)


CHANGELOG.md
```diff
@@ -11,6 +11,7 @@
   When upgrading from Strimzi 0.22 or earlier, you have to disable the `ControlPlaneListener` feature gate when upgrading the cluster operator to make sure the Kafka cluster stays available during the upgrade.
   When downgrading to Strimzi 0.22 or earlier, you have to disable the `ControlPlaneListener` feature gate before downgrading the cluster operator to make sure the Kafka cluster stays available during the downgrade.
 * Update to Cruise Control version 2.5.74
+* The `type` field in `ListenerStatus` have been deprecated and will be removed when updating to CRV v1
```
ppatierno, line 14: typo "CRD"

api/src/main/java/io/strimzi/api/kafka/model/status/ListenerStatus.java
```diff
@@ -28,19 +28,20 @@
...
     private String type;
+    private String name;
     private List<ListenerAddress> addresses;
     private String bootstrapServers;
     private List<String> certificates;
     private Map<String, Object> additionalProperties;
 
-    @Description("The type of the listener. " +
-            "Can be one of the following three types: `plain`, `tls`, and `external`.")
+    @Description("The `type` field in `ListenerStatus` have been deprecated and will be removed when updating to CRV v1" +
```
ppatierno, line 43: typo "CRD"

api/src/main/java/io/strimzi/api/kafka/model/status/ListenerStatus.java
```diff
@@ -49,6 +50,16 @@ public void setType(String type) {
...
+    @Description("The name of the listener. " +
+            "Can be one of the following three types: `plain`, `tls`, and `external`.")
+    public String getName() {
+        return name;
+    }
+
+    public void setName(String name) {
+        this.name = name;
+    }
+
```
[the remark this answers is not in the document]
ppatierno: @ShubhamRwt what's the status of this investigation, is that possible?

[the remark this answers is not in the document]
ShubhamRwt: @ppatierno Hello, I did tried out several ways out of what I found but still I was not able to make it use only one setter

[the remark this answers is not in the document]
ShubhamRwt: I am not a 100% sure if its not possible but I tried my best to remove it :(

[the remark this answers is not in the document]
ppatierno: @scholzj do you think it's a must have? From my understand from Shubham it's not possible so we should live with both.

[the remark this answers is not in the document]
scholzj: I don't know, I think we need to figure out something. Calling the setter twice everywere seems ... weird I guess.

api/src/main/java/io/strimzi/api/kafka/model/status/ListenerStatus.java
```diff
@@ -49,6 +50,16 @@ public void setType(String type) {
         this.type = type;
     }
 
+    @Description("The name of the listener. " +
+            "Can be one of the following three types: `plain`, `tls`, and `external`.")
```
ppatierno, line 54: I would avoid saying this here. Maybe it's better saying that for now it's reflecting the same value as type but when it will be deprecated, this field will report the actual name of the listener. @tombentley @scholzj wdyt?

scholzj: This is obviously wrong even for the `type` field and should be fixed.

ppatierno: Right this is still describing the older meaning of the `type` field.

ShubhamRwt: Do I need to change the description of the `name` field only or do I need to change description of `type` field also?

ShubhamRwt: @scholzj @ppatierno

scholzj: You should change both. They mean the same.

api/src/main/java/io/strimzi/api/kafka/model/status/ListenerStatus.java
```diff
@@ -28,19 +28,20 @@
...
+    private String name;
     private List<ListenerAddress> addresses;
     private String bootstrapServers;
     private List<String> certificates;
     private Map<String, Object> additionalProperties;
 
-    @Description("The type of the listener. " +
-            "Can be one of the following three types: `plain`, `tls`, and `external`.")
+    @Description("The `type` field in `ListenerStatus` have been deprecated and will be removed when updating to CRD v1" +
+            "The type of the listener can be one of the following three types: `plain`, `tls`, and `external`.")
```
scholzj, line 44: This is not true and should be removed. You should instead say that this it the name of the listener.

api/src/main/java/io/strimzi/api/kafka/model/status/ListenerStatus.java
ShubhamRwt: @scholzj @ppatierno does the description for both now look fine?

api/src/main/java/io/strimzi/api/kafka/model/status/ListenerStatus.java
```diff
@@ -49,6 +50,15 @@ public void setType(String type) {
         this.type = type;
     }
 
+    @Description("The `name` field reflects the same value as the `type` field. Once the `type` field gets removed, this will reflect the actual name of the listener.")
```
scholzj, line 53: You should explain what it is and not what it replaces or what has the same value. Just say that this is the name of the listener.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaAssemblyOperator.java
```diff
@@ -2000,6 +2000,7 @@ String zkConnectionString(int connectToReplicas, Function<Integer, String> zkNod
 
                 ListenerStatus ls = new ListenerStatusBuilder()
                         .withType(listener.getName())
+                        .withName(listener.getName())
```
scholzj, line 2003: From the community call, it sounded to me like we have two getters and we do not have to call this twice but instead always just call `withName` and it sets both name and type. So I'm a bit confused if this is still needed or not.

ShubhamRwt: I thought that on the community call we agreed on that we can have both the `setters` and these setter generate the method `withName` and `withType` which are then being used in the code

ShubhamRwt:
> * You need to make the `type` field deprecated. At least in the description or something like that.
> * * Thsi should have a note in the deprecation section in CHANGELOG.

@scholzj do we need this ?

scholzj:
> @scholzj do we need this ?

Which part of it?

ShubhamRwt:
Do we need the deprecation in both the description and the changelog?

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaAssemblyOperator.java
scholzj: Ok, maybe I forgot it. But for the record, it is bad implementation because sooner or later, someone will not set one of them and break the API one way or another. If this is our best way to do this, I do not think we should wait for a new CRD version but stop using it much sooner to get rid of this.

scholzj:
> Do we need the deprecation in both the description and the changelog?

Yes, we do.

ShubhamRwt:
So should we have the description of the `type` field like --> `"The `type` field in `ListenerStatus` have been deprecated and will be removed when updating to CRD v1" + " The `type` field reflects the name of the listener?

scholzj:
> So should we have the description of the type field like --> "The `type` field in `ListenerStatus` have been deprecated and will be removed when updating to CRD v1" + " The `type` field reflects the name of the listener?

Well, as I said above ... if we do not have a good implementation to get rid of the two setters, we should maybe stop using it sooner - before next version of the CRDs. We need to decide on this and it needs to reflect in the descriptions.

Plus, we do not know if next CRD version will be v1, so it anyway should be generic I guess.

ShubhamRwt:
Okay. Making it generic we can say -> "The `type` field in `ListenerStatus` have been deprecated and will be removed. +The `type` field reflects the name of the listener".

CHANGELOG.md
```diff
@@ -15,6 +15,7 @@
   When upgrading from Strimzi 0.22 or earlier, you have to disable the `ControlPlaneListener` feature gate when upgrading the cluster operator to make sure the Kafka cluster stays available during the upgrade.
   When downgrading to Strimzi 0.22 or earlier, you have to disable the `ControlPlaneListener` feature gate before downgrading the cluster operator to make sure the Kafka cluster stays available during the downgrade.
 * Update to Cruise Control version 2.5.74
+* The `type` field in `ListenerStatus` have been deprecated and will be removed when updating to CRD v1
```
tombentley, line 18:
```suggestion
* The `type` field in `ListenerStatus` has been deprecated and will be removed when updating to CRD v1
```

api/src/main/java/io/strimzi/api/kafka/model/status/ListenerStatus.java
tombentley:
```suggestion
    @Description("The name of the listener.")
```

@ShubhamRwt the reason to do it like this is it make the eventual removal PR much simpler if you only have to remove the getter and setter for type. If you also have to go changing the docs for name as well (so they no longer refer to the now removed `type`) then it's more work.

api/src/main/java/io/strimzi/api/kafka/model/status/ListenerStatus.java
tombentley, line 44: You should use the `@DeprecatedProperty` annotation. 

```suggestion
    @Description("The name of the listener.")
    @DeprecatedProperty(description="The `type` property is deprecated and has been replaced with the `name` property.", removalVersion="v1", movedToPath="name")
```

api/src/main/java/io/strimzi/api/kafka/model/status/ListenerStatus.java
```diff
@@ -28,19 +28,20 @@
...
 )
 @JsonInclude(JsonInclude.Include.NON_NULL)
-@JsonPropertyOrder({ "type", "addresses", "bootstrapServers", "certificates" })
+@JsonPropertyOrder({ "type", "name", "addresses", "bootstrapServers", "certificates" })
 @EqualsAndHashCode
 public class ListenerStatus implements UnknownPropertyPreserving, Serializable {
     private static final long serialVersionUID = 1L;
 
     private String type;
+    private String name;
```
tombentley, line 37: I still don't really understand exactly why we need two fields.

api/src/main/java/io/strimzi/api/kafka/model/status/ListenerStatus.java
ShubhamRwt: @tombentley Actually I tried using `@DeprecatedProperty earlier but the issue with it is that have to use @Deprecated annotation too then. And if I use it the test eventually starts failing since an extra condition gets added --> ```Condition(status=True, reason=DeprecatedFields, message=In API version null the type property at path status.listeners.type has been deprecated. This field has been deprecated and will be removed when upgrading to CRD
 v1., type=Warning, lastTransitionTime=2021-11-08T06:43:24.066635110Z, additionalProperties={})```

api/src/main/java/io/strimzi/api/kafka/model/status/ListenerStatus.java
ShubhamRwt: I guess some discussion was happened around this -->```After some offline discussion, the type field have to be as it is today so still reporting the listener name for backward compatibility.```

api/src/main/java/io/strimzi/api/kafka/model/status/ListenerStatus.java
tombentley: Why can't you change the test to expect that warning?

ShubhamRwt: I had a discussion with Jakub regarding this and he told that we might not want it

api/src/main/java/io/strimzi/api/kafka/model/status/ListenerStatus.java
ShubhamRwt: I tried to do the same using only one setter for both but I wasn't able to do so since there weremethods like `withType`  which are being used in the code, or we can say that if we don't create getter and setter for `type` field, we don't get that `withType` Method.

ShubhamRwt: I can use the `withName` method  which gets generated due to the setter of `Name` field which have same working but at the same time it will mean that the use of `type` field will be mostly over

tombentley (commented):
Just had an out-of-band conversation with @ShubhamRwt about this. I think we should have getters and setter for both "type" and "name" properties but they get and set the same `name` field, for the following reasons:

* Keeping `getType` means we don't break existing code which might be using it, allowing people to upgrade in their own time before the method is eventually removed.
* Keeping `setType` also means any test code they're written also continues to work. 
* Having `getName` and `setName` provides the method they need to migrate to in their main and test code. 

Having both pairs of getter and setters should work with sundrio and jackson, since they should both be looking at properties not fields. The main drawback is that the sundrio-generated `withType` won't end up with `@java.lang.Deprecated`, so will be a bit harder for people to spot. 

`setType` and `setName` should be annotated with with `@java.lang.Deprecated` and also our own `@DeprecatedProperty` annotations. If `@DeprecatedProperty` breaks tests either those tests should be fixed to allow for the possibility of deprecated properties, or we're saying `@DeprecatedProperty` itself is not fit for purpose and we need to rethink it's use everywhere it's currently being used and remove it (in a separate PR). 

WDYT @scholzj @ppatierno ?

scholzj:
@tombentley I think the important part is that the use both the same field => so that you do not need to do `.withName(name).withType(name)` in the code.

tombentley:
Yeah, right, just `withName`.

api/src/main/java/io/strimzi/api/kafka/model/status/ListenerStatus.java
```diff
@@ -28,25 +29,37 @@
...
     private List<ListenerAddress> addresses;
     private String bootstrapServers;
     private List<String> certificates;
     private Map<String, Object> additionalProperties;
 
-    @Description("The type of the listener. " +
-            "Can be one of the following three types: `plain`, `tls`, and `external`.")
+    @Description("The type of the listener can be one of the following three types: `plain`, `tls`, and `external`.")
+    @DeprecatedProperty(description = "The `type` field getter in `ListenerStatus` has been deprecated")
```
tombentley, line 44: You should be able to use `movedToPath = "name"`, which will be included in the generated docs.

tombentley (approved):
One nit, but otherwise LGTM.

CHANGELOG.md
scholzj: The comment from Tom here still stands I guess. plus you need to more this to the 0.28.0 release.

systemtest/src/main/java/io/strimzi/systemtest/kafkaclients/AbstractKafkaClient.java
```diff
@@ -202,6 +202,7 @@ public void verifyProducedAndConsumedMessages(int producedMessages, int consumed
         }
     }
 
+    @SuppressWarnings("deprecation")
```
scholzj, line 205: Why not update it instead of keeping the deprecated method?

systemtest/src/main/java/io/strimzi/systemtest/utils/kafkaUtils/KafkaUtils.java
```diff
@@ -145,6 +145,7 @@ public static void waitForZkMntr(String namespaceName, String clusterName, Patte
         }
     }
 
+    @SuppressWarnings("deprecation")
```
scholzj, line 148: Why not update it instead of keeping the deprecated method?

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaAssemblyOperator.java
```diff
@@ -4079,6 +4079,7 @@ void addListenerStatus(ListenerStatus ls)    {
             return Future.succeededFuture(this);
         }
 
+        @SuppressWarnings("deprecation")
```
scholzj, line 4082: Why not update it instead of keeping the deprecated method?

api/src/main/java/io/strimzi/api/kafka/model/status/ListenerStatus.java
```diff
@@ -28,25 +29,37 @@
...
-    private String type;
+    private String name;
     private List<ListenerAddress> addresses;
     private String bootstrapServers;
     private List<String> certificates;
     private Map<String, Object> additionalProperties;
 
-    @Description("The type of the listener. " +
-            "Can be one of the following three types: `plain`, `tls`, and `external`.")
+    @Description("The type of the listener can be one of the following three types: `plain`, `tls`, and `external`.")
```
scholzj, line 43: The description is wrong -> it clearly can have other values then just the three. Can you please update it?

scholzj (commented):
Should we have some serialization / deserialization test which checks that the values are properly set with both setters and returned with both getters?

ShubhamRwt:
> Should we have some serialization / deserialization test which checks that the values are properly set with both setters and returned with both getters?

Just a small question, In the classs KafkaStatusTest we are also checking             `assertThat(status.getListeners().get(0).getType(), is("plain"));`

So do we still require seperate tests?

systemtest/src/main/java/io/strimzi/systemtest/utils/kafkaUtils/KafkaUtils.java
ShubhamRwt: If I change `getType()` here and change it to `getName() , Should I remove it from the other places also and also from the tests ?

ppatierno (approved):
I think what Tom said makes sense. No other comments from me. LGTM.

systemtest/src/main/java/io/strimzi/systemtest/utils/kafkaUtils/KafkaUtils.java
scholzj: I guess - of course not where the `getType` is specifically what the test tests. But I don't know what the other places are. I think I commented mainly on those where I saw the deprcation suppression.,

scholzj:
> Just a small question, In the class `KafkaStatusTest` we are also checking `assertThat(status.getListeners().get(0).getType(), is("plain"));`
> 
> So do we still require seperate tests? @scholzj

I do not think that does any serialization / deserialization. That just tests that the getter gives you what was set on the same object. But I might misremember what the test does. So if it does (de)serialization as well, feel free to point it out.

api/src/main/java/io/strimzi/api/kafka/model/status/ListenerStatus.java
```diff
@@ -28,25 +29,37 @@
...
-    private String type;
+    private String name;
     private List<ListenerAddress> addresses;
     private String bootstrapServers;
     private List<String> certificates;
     private Map<String, Object> additionalProperties;
 
-    @Description("The type of the listener. " +
-            "Can be one of the following three types: `plain`, `tls`, and `external`.")
+    @Description("The type of the listener.")
```
scholzj, line 43:
```suggestion
    @Description("The name of the listener.")
```

api/src/main/java/io/strimzi/api/kafka/model/status/ListenerStatus.java
```diff
@@ -28,25 +29,37 @@
...
+    @DeprecatedProperty(movedToPath = "name")
+    @Deprecated
     public String getType() {
-        return type;
+        return name;
     }
 
-    public void setType(String type) {
-        this.type = type;
+    @DeprecatedProperty(description = "The `type` field  setter in `ListenerStatus` has been deprecated")
```
scholzj, line 50: I don't think this is used anywhere.

api/src/test/java/io/strimzi/api/kafka/model/KafkaTest.java
```diff
@@ -132,6 +137,74 @@ public void testListeners()    {
...
+                                .build(),
+                        new ListenerStatusBuilder()
+                                .withName("external")
+                                .withAddresses(new ListenerAddressBuilder()
+                                        .withHost("my-route-address.domain.tld")
+                                        .withPort(443)
+                                        .build())
+                                .build())
+                .endStatus()
+                .build();
```
scholzj, line 188: If you indent this along the `withNew...()` and `end...()` methods, it will be much more readable.

api/src/test/java/io/strimzi/api/kafka/model/KafkaTest.java
```diff
@@ -132,6 +137,74 @@ public void testListeners()    {
...
+
+        String path = Objects.requireNonNull(this.getClass().getResource("Kafka-listener-name-and-status.yaml")).toURI().getPath();
+        assertThat(TestUtils.toYamlString(kafka), is(TestUtils.getFileAsString(path)));
+        System.out.println(path);
```
scholzj, line 192: Thsi can be deleted I guess?

api/src/test/java/io/strimzi/api/kafka/model/KafkaTest.java
```diff
@@ -132,6 +137,74 @@ public void testListeners()    {
...
+    public void testListenersTypeAndName()    {
+        Kafka model = TestUtils.fromYaml("Kafka-listener-name-and-status" + ".yaml", Kafka.class);
+
+        assertThat(model.getStatus().getListeners(), is(notNullValue()));
+        assertThat(model.getStatus().getListeners().size(), is(2));
+
+        List<ListenerStatus> listeners = model.getStatus().getListeners();
+
+        assertThat(listeners.get(0).getType(), is("plain"));
+        assertThat(listeners.get(1).getName(), is("external"));
```
scholzj, line 205:
```suggestion
        assertThat(listeners.get(0).getType(), is("plain"));
        assertThat(listeners.get(0).getName(), is("plain"));
        assertThat(listeners.get(1).getType(), is("external"));
        assertThat(listeners.get(1).getName(), is("external"));
```

packaging/helm-charts/helm3/strimzi-kafka-operator/crds/040-Crd-kafka.yaml
```diff
@@ -5930,7 +5930,10 @@ spec:
                     properties:
                       type:
                         type: string
-                        description: 'The type of the listener. Can be one of the following three types: `plain`, `tls`, and `external`.'
+                        description: 'The type of the listener can be one of the following three types: `plain`, `tls`, and `external`.'
```
scholzj, line 5933: The CRD was not updated I guess.

scholzj (commented):
The test looks good. I left some more nits.

api/src/main/java/io/strimzi/api/kafka/model/status/ListenerStatus.java
ShubhamRwt: So we don't need a Deprecated tag over it ?

scholzj: Well, I'm talking about the `@DeprecatedProperty` => do you see it used for anything? It does not seem to. I think it would need to be on the getter and not on the setter.

ShubhamRwt: Done

scholzj (approved):

scholzj:
(needs regression run after the DCO is fixed)

Issue #5758, opened 2021-10-21
Add topic configuration settings to KafkaTopic CR status


katheris:
**Is your feature request related to a problem? Please describe.**
As a user when I create a KafkaTopic it would be useful to see what configuration has been applied to the topic as some of these values may be set based on either Kafka or Strimzi defaults. This issue is being created based on a discussion in the community call on 21st October 2021.

**Describe the solution you'd like**
Add configuration details such as replication factor and minimum insync replicas into the KafkaTopic CR status field.

**Describe alternatives you've considered**
As an alternative developers could specify these values themselves in the CR so that they don't use any defaults

**Additional context**

scholzj:
@katheris Did you had a chance to look into it? Is this still something we want / need?

katheris:
Hey @scholzj, I've tested this and without the topic operator enabled I can create a `KafkaTopic` with no `partitions` or `replicas`, then once the topic operator comes up it puts in the values. So I think Tom is right that we don't need this because the user can just look at the `KafkaTopic` spec to determine the partitions and replicas

PR #5779, opened 2021-10-26
Add 'replicas' and 'config' to KT status

sknot-rh:

### Type of change
- Enhancement

### Description
Add 'replicas' and 'config' to KT status. Fixes https://github.com/strimzi/strimzi-kafka-operator/issues/5758

### Checklist

- [ ] Write tests
- [ ] Update documentation
- [ ] Update CHANGELOG.md


scholzj:
Discussed on the Community call => it does not seem to make sense for the TO to update both status and spec since the TO is updating the spec as well. So this should be closed.

@sknot-rh Does that make sense? Are you fine with this being closed?

sknot-rh:
Yes, as Tom mentioned, TO works in a bidirectional way. So change in kafka is projected into the k8s resource. Thus duplicate the configuration in the spec is not usefull.

PR #5806, opened 2021-10-29
[systemtest] - SecurityST - disable checks for KE and CC in renewal tests

im-konge:

### Type of change

- Disable test checks

### Description

It seems that there is some issue, which causing `KafkaExporter`, `CruiseControl` and maybe `EntityOperator` roll even on clients CA renewal - this PR disables test checks, verifying if `KE` and `CC` pods were rolled, even they shouldn't - after the issue will be fixed, we will enable it again.

### Checklist


im-konge:
@strimzi-ci run tests test_only profile=regression testcase=SecurityST

Frawless (approved):

Issue #5767, opened 2021-10-25
Kafka upgrade, with inter broker protocol and log message format as M.m.p, fails with misleading error


ppatierno:
Starting from a `Kafka` CR having `spec.kafka.version: 2.8.0` and the corresponding inter broker protocol and log message format as following:

```yaml
config:
      log.message.format.version: "2.8"
      inter.broker.protocol.version: "2.8"
```

doing an upgrade to `3.0.0`, everything works fine.
If "by mistake", we use something like this as starting point instead:

```yaml
config:
      log.message.format.version: "2.8.0"
      inter.broker.protocol.version: "2.8.0"
```

So not just M.m but M.m.p for above versions, the Kafka upgrade to 3.0.0 fails with following condition:

```yaml
conditions:
    - lastTransitionTime: '2021-10-25T09:25:28.802Z'
      message: >-
        log.message.format.version (2.8.0) and inter.broker.protocol.version
        (2.8.0) used by the brokers have to be lower or equal to the Kafka
        broker version we upgrade from (2.8.0)
      reason: KafkaUpgradeException
      status: 'True'
      type: NotReady
```

which of course doesn't make sense because 2.8.0 is actually lower or equal to 2.8.0.

NOTE:
At same time using M.m.p versions, it seems that during Kafka cluster creation they are considered valid values and Kafka brokers work without any issues.

sknot-rh:
I agree that message does not make sense. It is obvious by comparing the strings `2.8` and `2.8.0`. The validation does not catch the Mmp version because we use this regex to get just Mm version: https://github.com/strimzi/strimzi-kafka-operator/blob/main/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/resource/KafkaSpecChecker.java#L34
The solution could be making this regex more strict (to check only x.y version is provided), or perform a new (validation) check around here https://github.com/strimzi/strimzi-kafka-operator/blob/main/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaAssemblyOperator.java#L1018

PR #5828, opened 2021-11-02
Fix message when invalid log message or inter protocol version is provided

sknot-rh:

### Type of change
- Bugfix

### Description
Fixes https://github.com/strimzi/strimzi-kafka-operator/issues/5767
the bug described in the issue ^

Fixed by considering `x.y.z` as an invalid format. The only format is `x.y` or `x.y-IVz`.

### Checklist

- [ ] Write tests
- [ ] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaVersion.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaVersion.java
@@ -436,7 +436,8 @@ public class KafkaVersion implements Comparable<KafkaVersion> {
                 return 1;
             }
         }
-        return components.length - otherComponents.length;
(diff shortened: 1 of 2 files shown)


scholzj (commented):
Can you maybe explain in more detail what you did and why? I don't think it is really clear to me.

scholzj (commented):
Also, as was suggested in the related discussion, the full version is not necessarily wrong. So is this the right way to fix it?

ppatierno:
Agree with Jakub, the issue is not about restrict the user to use M.m because M.m.p is a valid value. The issue should be about allowing CO to proced with the upgrade when a M.m.p is specified and not raising a condition error.

scholzj (approved):
I guess this looks good. MAybe you can fix the description of the PR?

PR #5832, opened 2021-11-02
[MO] - @PackageParallelism -> eliminate `*` and always specify namespaces to watch

see-quick:
### Type of change

- Enhancement / new feature

### Description

This PR adds for the support of `RBAC=NAMESPACE` parallelism. Also, remove two distinctions when we run cluster-wide we set `STRIMZI_NAMESPACE='*'`. In this PR, I always specify a set of namespaces. 

Moreover, I have also created a new class (`NamespaceWatcher`), which takes care of `@ParallelNamespaceTest` namespaces and always use a specific amount for instance 10 namespaces. Each thread then picks one of the `free-namespaces` and lock it by setting the HashMap value to the `true`.

```
namespace-0 -> false
namespace-1 -> true (already assignemt to some thread)
namespace-2 -> false 
...
```
If the test case is done all resources are deleted except namespace. 

### Checklist


see-quick:
Waiting for https://github.com/fabric8io/kubernetes-client/pull/3559 to be merged.

PR #5850, opened 2021-11-06
Make ZooKeeper rolling independent on StatefulSets

scholzj:
### Type of change

- Enhancement / new feature

### Description

Currently, rolling of ZooKeeper clusters is tightly coupled with StatefulSets. That is something what needs to be dealt with as part of the ZooKeeper removal work. In order to avoid too big PRs, this PR refactors the ZooKeeper rolling update to not depend on the stateful set. It adds a new `ZooKeeperRoller` class which uses the same logic as we used so far but instead of depending on stateful sets it uses pod selector to find the pods and roll if needed. It also removed the rolling update logic from the StatefulSet operator and its Kafka and Zoo version since neither one uses it anymore.

### Checklist

- [x] Write tests

### DIFF
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/resource/ZooKeeperRoller.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/resource/ZooKeeperRoller.java
@@ -98,7 +98,7 @@ public class ZooKeeperRoller {
                                 }
                             }
 
-                            // Check if we have a leader and if it needs tolling
+                            // Check if we have a leader and if it needs rolling
                             if (ZookeeperLeaderFinder.UNKNOWN_LEADER.equals(leader) || podsToRoll.get(leader) == null || podsToRoll.get(leader).isEmpty()) {
                                 return fut;
                             } else {
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/resource/ZooKeeperRoller.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/resource/ZooKeeperRoller.java
@@ -1,0 +1,141 @@
+/*
+ * Copyright Strimzi authors.
+ * License: Apache License 2.0 (see the file LICENSE or http://apache.org/licenses/LICENSE-2.0.html).
+ */
+package io.strimzi.operator.cluster.operator.resource;
+
+import io.fabric8.kubernetes.api.model.Pod;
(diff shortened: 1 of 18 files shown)


cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/resource/ZooKeeperRoller.java
```diff
@@ -0,0 +1,141 @@
...
+                                        // roll the pod and wait until it is ready
+                                        // this prevents rolling into faulty state (note: this applies just for ZK pods)
+                                        fut = fut.compose(ignore -> restartPod(reconciliation, podEntry.getKey(), podEntry.getValue()));
+                                    } else {
+                                        LOGGER.debugCr(reconciliation, "Deferring restart of leader {}", podEntry.getKey());
+                                    }
+                                }
+                            }
+
+                            // Check if we have a leader and if it needs tolling
```
sknot-rh, line 101: typo in the comment (tolling)

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/resource/ZookeeperLeaderFinder.java
```diff
@@ -30,32 +26,27 @@
...
  * Helper class for finding the leader of a ZK cluster
  */
 public class ZookeeperLeaderFinder {
-
     private static final ReconciliationLogger LOGGER = ReconciliationLogger.create(ZookeeperLeaderFinder.class);
 
     private static final Pattern LEADER_MODE_PATTERN = Pattern.compile("^Mode: leader$", Pattern.MULTILINE);
 
-    public static final int UNKNOWN_LEADER = -1;
+    public static final String UNKNOWN_LEADER = "-1";
```
sknot-rh, line 43: Can you please explain, why is it better to operate with `String` instead of `int`?

sknot-rh (approved):
LGTM, just one question

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/resource/ZookeeperLeaderFinder.java
scholzj: The methods don't return pod indexes from StatefulSet anymore but pod names => i.e. Strings. So they cannot return an integer when the leader is not found. So I made the `-1` a String as well so that it could be returned. I'm not sure if there is some better way to do this.

Issue #5579, opened 2021-09-17
Strimzi kafka operator pod getting evicted EmptyDir volume "strimzi-tmp" exceeds the limit


nokernel:
**Describe the bug**
Strimzi kafka operator pod getting evicted `Usage of EmptyDir volume "strimzi-tmp" exceeds the limit "1Mi".`

Related to https://github.com/strimzi/strimzi-kafka-operator/pull/5375

**To Reproduce**
1. Create kafka clusters that can't be scheduled ( causing alot of java error in operator logs )
2. Wait

**Expected behavior**
Allow customisation of strimzi-tmp limit

**Environment (please complete the following information):**
 - Strimzi version: 0.25.0
 - Installation method: Helm chart
 - Kubernetes cluster: Kubernetes 1.19.1

scholzj:
Can you get the content of the directory as it evolves over the time (at least listing the files)? Or provide more exact steps to reproduce it? Because the ones you have there are very generic and do not seem to do anything for me (maybe you need to explain what you did so that it cannot be scheduled).

In general, for the operator, you can increase the directory size just by editing the deployment.

nokernel:
I will try to get more info on what is filling up the tmp mount, maybe that will uncover another issue.

sknot-rh:
Having a reproducer would be helpful. I created the Kafka cluster with invalid configuration, CO logs are full of errors, but the content of the `tmp` mount does not change.

mdharper:
I've run into the same issue today with a KafkaConnect resource. It looks like the JDBC Connector/JTDS driver uses the temp directory on its own:
`drwxr-xr-x 2 kafka root       60 Sep 23 19:22 hsperfdata_kafka
-rw-r--r-- 1 kafka root 77426688 Sep 23 19:23 jtds10409044170123727212.tmp
-rw-r--r-- 1 kafka root 36978914 Sep 23 19:22 jtds6814623151674711811.tmp
drwxr-xr-x 2 kafka root       40 Sep 23 19:22 kafka
-rw-r--r-- 1 kafka root      960 Sep 23 19:22 strimzi-connect.properties`

It might be that I'm seeing a lot more volume and that's the trigger. Has the use of strimzi-tmp changed in recent operator versions?

scholzj:
What is JTDS driver? Does it have some way to tell it how much stuff should it create? I guess we can in theory make the size configurable. But creating (fairly big) random files in `/tmp` is always asking for trouble.

[...]

PR #5851, opened 2021-11-06
Fix leaking keystores / truststores in ZooKeeperScaler

scholzj:
### Type of change

- Bugfix

### Description

The `ZooKeeperScaler` is currently creating the keystore and truststore every scale operation. But it deletes them only once. As a result, if user scales Zoo up or down by more than one node, it will create multiple sets of the certificate stores but delete only one. This is causing issues such as #5579 when the temp dir gets full.

This PR moves the certificate store connection to the constructor where the certificate secrets are passed as well. Thanks to that they are created only once regardless how many nodes you add and deletes them.

Since the leaking of files is a bigger issue now with the size-limited temp directory, I also changed it to WARN level when the file deletion fails for any reason to make sure we catch it.

### Checklist


ppatierno (approved):
Lgtm

PR #5864, opened 2021-11-09
remove numbered prefix from helm templates

themarcelor:
### Type of change

- Refactoring

### Description

There is no need for numbered prefixes as Helm is smart enough to know in which order the k8s artifacts need to be applied against the control plane.
```
var InstallOrder SortOrder = []string{
	"Namespace",
	"ResourceQuota",
	"LimitRange",
	"PodSecurityPolicy",
	"Secret",
	"ConfigMap",
	"StorageClass",
	"PersistentVolume",
	"PersistentVolumeClaim",
	"ServiceAccount",
	"CustomResourceDefinition",
	"ClusterRole",
	"ClusterRoleBinding",
	"Role",
	"RoleBinding",
	"Service",
	"DaemonSet",
	"Pod",
	"ReplicationController",
	"ReplicaSet",
	"Deployment",
	"StatefulSet",
	"Job",
	"CronJob",
	"Ingress",
	"APIService",
}
```
source: https://github.com/helm/helm/blob/release-2.10/pkg/tiller/kind_sorter.go#L29

### Test

The following command renders the final Helm chart successfully:
```
% helm template --namespace myproject strimzi-kafka-operator
```


scholzj:
Is there any actual benefit for removing them?

The numbering is useful when pointing to the right files and working with them in general. It comes from the YAML installation files, which is Strimzi primary mean of installation => where the installation order is anyway not the main reason for using it (hint: with things being asynchronous, the order of the files doesn't matter that much anyway).

I don't care much about the numbering in Helm Charts, but to make this work, you would need to:
* Do the changes to the Helm Chart files in `./packaging/` - the Helm Chart which you changes is from the last release and you cannot change it.
* Make sure the `make helm_install` works properly and generates numbered installation files in the `packaging/install` folder.

themarcelor:
Hi @scholzj , looks like some environment variable with the expected checksum needs to be updated 🤷🏼 
```
if this is part of a release instead update the checksum i.e.
HELM_CHART_CHECKSUM="4d0330a8ea4fe55d2afb3207afcad251c04b29e4  -"
->
HELM_CHART_CHECKSUM="3d71b5f92f0b9db433c5a040ca5edfe0dcc304d2  -"
```

scholzj:
> Hi @scholzj , looks like some environment variable with the expected checksum needs to be updated 

But you are not doing a release, so you should not touch those files.

themarcelor:
@scholzj , I now realize that Helm can only be smart enough to handle native Kubernetes resources but it cannot figure out the order of CRDs.
Hence, we would only be able to remove the numbered prefixes from YAMLs with native k8s abstractions declared in the "Kind" parameter.

I'm closing this PR as this is not really critical, it was just an idea.

PR #5870, opened 2021-11-10
Refactor Pod and PVC cleanup to make it less dependent on StatefulSets

scholzj:
### Type of change

- Refactoring

### Description

This PR refactors the the Pod and PVC cleanup to make it less dependent on StatefulSets and make it easier to replace the StatefulSets in the future. The implementation logic is the same. But the old implementation used the StatefulSet across all the methods. The refactored implementation isolates use of StatefulSets into a single method and makes it easier to replace it with something else (or use something else as an alternative). It also slightly reduces the number of Kube API calls needed when no cleanup is requested (which is most of the reconciliations).

To allow non-cascading deletes of Pods or PVCs, the `deleteAsync` method which was before only in the ` StatefulSetOperator` is moved to `AbstractResourceOperator`. It just delegates to the `internalDelete` method already used when you use `reconcile` with `null` value, but allows to configure cascading which `reconcile` does not support.

### Checklist

### DIFF
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaAssemblyOperator.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaAssemblyOperator.java
@@ -2921,7 +2921,7 @@ public class KafkaAssemblyOperator extends AbstractAssemblyOperator<KubernetesCl
         Future<ReconciliationState> manualPodCleaning(String ctlrResourceName, String podName, Labels selector, List<PersistentVolumeClaim> desiredPvcs) {
             return pvcOperations.listAsync(namespace, selector)
                     .compose(existingPvcs -> {
-                        // Find out which PVC need to be deleted
+                        // Find out which PVCs need to be deleted
                         List<PersistentVolumeClaim> deletePvcs;
 
                         if (existingPvcs != null) {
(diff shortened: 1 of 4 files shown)


cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaAssemblyOperator.java
```diff
@@ -2882,170 +2869,159 @@ StatefulSet getKafkaStatefulSet()   {
...
+                        return cleanPodPvcAndStatefulSet(ctlrResourceName, podName, createPvcs, deletePvcs);
+                    })
+                    .map(this);
+        }
 
-            return Future.succeededFuture();
+        /**
+         * Handles the deletion and recreation of the StatefulSet controlling the pod which should be cleaned. In order
+         * to clean the pod and its PVCs, we first need to delete the StatefulSet (non-cascading). Otherwise the
+         * StatefulSet will break the process by recreating the pods or PVCs. This method first deletes the SattefulSet
```
sknot-rh, line 2950: Typo in the word `SattefulSet`

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaAssemblyOperator.java
```diff
@@ -2882,170 +2869,159 @@ StatefulSet getKafkaStatefulSet()   {
...
-                                        List<PersistentVolumeClaim> deletePvcs;
+        Future<ReconciliationState> maybeManualPodCleaning(String ctlrResourceName, Labels selector, List<PersistentVolumeClaim> desiredPvcs) {
+            return podOperations
+                    .listAsync(namespace, selector)
+                    .compose(pods -> {
+                        // Only one pod per reconciliation is rolled
+                        Pod podToClean = pods
+                                .stream()
+                                .filter(pod -> Annotations.booleanAnnotation(pod, AbstractScalableResourceOperator.ANNO_STRIMZI_IO_DELETE_POD_AND_PVC, false))
+                                .findFirst()
```
sknot-rh, line 2897: Just a question. Why can't we get all such pods and execute all the operations in single reconciliation?

sknot-rh (approved):
Just a nit about the typo and one question. It is not clear to me why the cleaning has to happen in multiple reconciliations.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaAssemblyOperator.java
scholzj: I really followed the previous logic here. IIRC, the reason why it was implemented this way was to avoid causing too much disruption by deleting too many things at the same time.

sknot-rh: Ah, thank you.

PR #5880, opened 2021-11-12
Topic config validation

sknot-rh:
### Type of change
- Enhancement / new feature

### Description
In this PR the validation of `spec.config` for KafkaTopic type is added. To be able check the config, we need to generate topic config model, similarly as we do for brokers config model.
When we have the config model, we can check the expected and actual data type of provided config entries.
The config model is tied to the kafka version, so we need let the TO know, what version of kafka it is used in the brokers. It is done in this PR by adding an envVar (open for better suggestions). 
Because until now we used config classes in the CO, I needed to do some refactoring.

### Checklist
- [ ] Write tests
- [ ] Update documentation
- [x] Update CHANGELOG.md


cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaAssemblyOperator.java
```diff
@@ -3160,6 +3160,7 @@ StatefulSet getKafkaStatefulSet()   {
 
             if (entityOperator != null) {
                 EntityTopicOperator topicOperator = entityOperator.getTopicOperator();
+                topicOperator.setKafkaVersion(kafkaCluster.getKafkaVersion().version());
```
sknot-rh, line 3163: @tombentley I do not like this but I was not able to find a better way :( WDYT?

topic-operator/src/main/java/io/strimzi/operator/topic/TopicSerialization.java
```diff
@@ -45,34 +49,30 @@
...
-                    } else {
-                        msg += " but was of type " + v.getClass().getName();
-                    }
-                    throw new InvalidTopicException(kafkaTopic, "KafkaTopic's spec.config has invalid entry: " +
-                            "The key '" + key + "' of the topic config is invalid: " + msg);
-                }
-            }
-            return result;
+            KafkaTopicConfiguration cfg = new KafkaTopicConfiguration(reconciliation, kafkaTopic.getSpec().getConfig().entrySet());
+            validateConfiguration(reconciliation, kafkaTopic, kafkaVersion, cfg);
```
sknot-rh, line 74: @tombentley Can you see a better place to perform the validation?

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/AbstractModel.java
```diff
@@ -140,6 +141,7 @@
     public static final String ENV_VAR_STRIMZI_JAVA_SYSTEM_PROPERTIES = "STRIMZI_JAVA_SYSTEM_PROPERTIES";
     public static final String ENV_VAR_STRIMZI_JAVA_OPTS = "STRIMZI_JAVA_OPTS";
     public static final String ENV_VAR_STRIMZI_GC_LOG_ENABLED = "STRIMZI_GC_LOG_ENABLED";
+    public static final String ENV_VAR_STRIMZI_KAFKA_VERSION = "STRIMZI_KAFKA_VERSION";
```
scholzj, line 144: Should this be defined in the EnttityTopicOperator class is if it is used only there?

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaAssemblyOperator.java
scholzj, line 3163: Shouldn't this be handled via fromCrd method?

scholzj (commented):
I left some nits. But should be reviewed by Tom I guess. Also, I did not noticed any new tests for the topic config validation.

cluster-operator/src/test/java/io/strimzi/operator/cluster/ResourceUtils.java
```diff
@@ -173,6 +173,7 @@ public static Kafka createKafka(String namespace, String name, int replicas,
                         .withLivenessProbe(probe)
                         .withReadinessProbe(probe)
                         .withStorage(new EphemeralStorage())
+                        .withVersion("3.0.0")
```
tombentley, line 176: Consider using `KafkaVersionTestUtils` to avoid hard coding a particular version.

cluster-operator/src/test/java/io/strimzi/operator/cluster/model/EntityTopicOperatorTest.java
```diff
@@ -116,6 +116,7 @@
         expected.add(new EnvVarBuilder().withName(EntityTopicOperator.ENV_VAR_SECURITY_PROTOCOL).withValue(EntityTopicOperatorSpec.DEFAULT_SECURITY_PROTOCOL).build());
         expected.add(new EnvVarBuilder().withName(EntityTopicOperator.ENV_VAR_TLS_ENABLED).withValue(Boolean.toString(true)).build());
         expected.add(new EnvVarBuilder().withName(EntityTopicOperator.ENV_VAR_STRIMZI_GC_LOG_ENABLED).withValue(Boolean.toString(AbstractModel.DEFAULT_JVM_GC_LOGGING_ENABLED)).build());
+        expected.add(new EnvVarBuilder().withName(EntityTopicOperator.ENV_VAR_STRIMZI_KAFKA_VERSION).withValue("3.0.0").build());
```
tombentley, line 119: `KafkaVersionTestUtils` again

config-model-generator/src/main/java/io/strimzi/build/kafka/metadata/KafkaConfigModelGenerator.java
```diff
@@ -21,52 +16,37 @@
...
+public class KafkaConfigModelGenerator extends CommonConfigModelGenerator {
 
-    private static Map<String, ConfigModel> configs() throws NoSuchMethodException, IllegalAccessException, InvocationTargetException {
+    @Override
+    protected Map<String, ConfigModel> configs() throws NoSuchMethodException, IllegalAccessException, InvocationTargetException {
         ConfigDef def = brokerConfigs();
         Map<String, String> dynamicUpdates = brokerDynamicUpdates();
         Method getConfigValueMethod = def.getClass().getDeclaredMethod("getConfigValue", ConfigDef.ConfigKey.class, String.class);
-        getConfigValueMethod.setAccessible(true);
+        doPrivileged((PrivilegedAction) () -> {
```
tombentley, line 39: I don't think we need the whole `doPrivileged` thing, since the config model generator operates at build time => We're in control of whether there's a security manager.

topic-operator/src/main/java/io/strimzi/operator/topic/TopicSerialization.java
```diff
@@ -45,52 +48,47 @@
...
 
+    private static void validateConfiguration(Reconciliation reconciliation, KafkaTopic kafkaTopic, String desiredVersion, KafkaTopicConfiguration configuration) {
+        List<String> errorsInConfig = configuration.validate(desiredVersion);
+
+        if (!errorsInConfig.isEmpty()) {
+            for (String error : errorsInConfig) {
+                LOGGER.warnCr(reconciliation, "KafkaTopic {}/{} has invalid spec.config: {}",
+                        kafkaTopic.getMetadata().getNamespace(),
+                        kafkaTopic.getMetadata().getName(),
+                        error);
```
tombentley, line 59: Is there much value in logging, when we're throwing the exception. Won't the catcher of the exception both log the message and also set the CR condition to include the message? Not logging here would avoid needing to pass the reconciliation so deep.

topic-operator/src/main/java/io/strimzi/operator/topic/TopicSerialization.java
```diff
@@ -45,52 +48,47 @@
...
         }
     }
 
-
     /**
      * Create a Topic to reflect the given KafkaTopic resource.
      * @throws InvalidTopicException
      */
-    public static Topic fromTopicResource(KafkaTopic kafkaTopic) {
+    public static Topic fromTopicResource(KafkaTopic kafkaTopic, Reconciliation reconciliation, String kafkaVersion) {
```
tombentley, line 84: I guess that, although it's obtainable from the environment, you're using an explicit parameter for testing purposes? It would make the patch smaller to just overload `fromTopicResource(KafkaTopic)` with `fromTopicResource(KafkaTopic, String)` (I'm assuming that the logging and `KafkaTopicConfiguration` comments mean that you don't need to pass the Reconciliation either).

config-model-generator/src/main/java/io/strimzi/build/kafka/metadata/KafkaConfigModelGenerator.java
sknot-rh: spotbugs is complaining about this. Should I suppress the warning?

config-model-generator/src/main/java/io/strimzi/build/kafka/metadata/KafkaTopicConfigModelGenerator.java
```diff
@@ -0,0 +1,98 @@
...
+                descriptor.setPattern(".+");
+            } else if (key.validator instanceof ThrottledReplicaListValidator$)   {
+                continue;
+            } else if (key.validator != null) {
+                throw new IllegalStateException("Invalid validator class " + key.validator.getClass() + " for option " + configName);
+            }
+
+            result.put(configName, descriptor);
+        }
+        return result;
```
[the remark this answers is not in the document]
sknot-rh: There is just one difference, different class at line 75

topic-operator/src/main/java/io/strimzi/operator/topic/TopicSerialization.java
sknot-rh: I just wanted to be consistent with the broker (and eventually connect) config validation.

sknot-rh: What do others think? Should we keep the validation of the Topic configuration as simple as possible or inherit it from the broker validation? @strimzi/maintainers

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 * Added the option to configure the Cluster Operator's Zookeeper admin client session timeout via an new env var: `STRIMZI_ZOOKEEPER_ADMIN_SESSION_TIMEOUT_MS`
 * The `ControlPlaneListener` and `ServiceAccountPatching` feature gates are now in the _beta_ phase and are enabled by default.
 * Allow setting any extra environment variables for the Cluster Operator container through Helm using a new `extraEnvs` value.
+* Validate `spec.config` in the KafkaTopic resources
```
scholzj, line 8: I think this deserves a bit more quality

```suggestion
* Validate `spec.config` in the KafkaTopic resources against available Kafka topic configuration options
```

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/EntityTopicOperator.java
```diff
@@ -246,6 +248,7 @@ public static EntityTopicOperator fromCrd(Reconciliation reconciliation, Kafka k
                 if (topicOperatorSpec.getLivenessProbe() != null) {
                     result.setLivenessProbe(topicOperatorSpec.getLivenessProbe());
                 }
+                result.setKafkaVersion(kafkaAssembly.getSpec().getKafka().getVersion());
```
scholzj, line 251: Does this in any way depend also on the `inter.broker.protocol.version`? Also, how but risk is the asynchronicity of the upgrade (Kafka is rolled first).

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/EntityTopicOperator.java
```diff
@@ -358,4 +362,7 @@ public String createLog4jProperties(OrderedProperties properties) {
         return super.createLog4jProperties(properties);
     }
 
+    public void setKafkaVersion(String kafkaVersion) {
```
scholzj, line 365: Do we really need the setter? Can't we just set the variable in `fromCrd`? If you have a setter, should we have a getter?

cluster-operator/src/test/java/io/strimzi/operator/cluster/ResourceUtils.java
```diff
@@ -173,6 +173,7 @@ public static Kafka createKafka(String namespace, String name, int replicas,
                         .withLivenessProbe(probe)
                         .withReadinessProbe(probe)
                         .withStorage(new EphemeralStorage())
+                        .withVersion(KafkaVersionTestUtils.LATEST_KAFKA_VERSION)
```
scholzj, line 176: Why does this need to be added here? Version is not required field. So if it is required by your changes, you need to fix it.

topic-operator/src/test/java/io/strimzi/operator/topic/TopicOperatorBaseIT.java
```diff
@@ -404,7 +405,7 @@ protected void awaitTopicConfigInKube(String resourceName, String key, String ex
         waitFor(() -> {
             KafkaTopic topic = operation().inNamespace(NAMESPACE).withName(resourceName).get();
             LOGGER.info("Polled topic {}, waiting for config change", resourceName);
-            String gotValue = TopicSerialization.fromTopicResource(topic).getConfig().get(key);
+            String gotValue = TopicSerialization.fromTopicResource(topic, "3.0.0").getConfig().get(key);
```
scholzj, line 408: I think having the version hardcoded here will cause issues int he future. Applies also to the one above and below.

topic-operator/src/test/java/io/strimzi/operator/topic/TopicOperatorIT.java
```diff
@@ -193,14 +192,14 @@ public void testKafkaTopicAddedWithHigherMinIsrThanBrokers() throws InterruptedE
...
+               "KafkaTopic topic-operator-it/test-resource-created-with-invalid-config has invalid spec.config: message.format.version has value 'zebra' which does not match the required pattern: \\Q0.8.0\\E(\\.[0-9]+)*|\\Q0.8.0\\E|\\Q0.8.1\\E(\\.[0-9]+)*|\\Q0.8.1\\E|\\Q0.8.2\\E(\\.[0-9]+)*|\\Q0.8.2\\E|\\Q0.9.0\\E(\\.[0-9]+)*|\\Q0.9.0\\E|\\Q0.10.0\\E(\\.[0-9]+)*|\\Q0.10.0-IV0\\E|\\Q0.10.0-IV1\\E|\\Q0.10.1\\E(\\.[0-9]+)*|\\Q0.10.1-IV0\\E|\\Q0.10.1-IV1\\E|\\Q0.10.1-IV2\\E|\\Q0.10.2\\E(\\.[0-9]+)*|\\Q0.10.2-IV0\\E|\\Q0.11.0\\E(\\.[0-9]+)*|\\Q0.11.0-IV0\\E|\\Q0.11.0-IV1\\E|\\Q0.11.0-IV2\\E|\\Q1.0\\E(\\.[0-9]+)*|\\Q1.0-IV0\\E|\\Q1.1\\E(\\.[0-9]+)*|\\Q1.1-IV0\\E|\\Q2.0\\E(\\.[0-9]+)*|\\Q2.0-IV0\\E|\\Q2.0-IV1\\E|\\Q2.1\\E(\\.[0-9]+)*|\\Q2.1-IV0\\E|\\Q2.1-IV1\\E|\\Q2.1-IV2\\E|\\Q2.2\\E(\\.[0-9]+)*|\\Q2.2-IV0\\E|\\Q2.2-IV1\\E|\\Q2.3\\E(\\.[0-9]+)*|\\Q2.3-IV0\\E|\\Q2.3-IV1\\E|\\Q2.4\\E(\\.[0-9]+)*|\\Q2.4-IV0\\E|\\Q2.4-IV1\\E|\\Q2.5\\E(\\.[0-9]+)*|\\Q2.5-IV0\\E|\\Q2.6\\E(\\.[0-9]+)*|\\Q2.6-IV0\\E|\\Q2.7\\E(\\.[0-9]+)*|\\Q2.7-IV0\\E|\\Q2.7-IV1\\E|\\Q2.7-IV2\\E|\\Q2.8\\E(\\.[0-9]+)*|\\Q2.8-IV0\\E|\\Q2.8-IV1\\E|\\Q3.0\\E(\\.[0-9]+)*|\\Q3.0-IV0\\E|\\Q3.0-IV1\\E");
```
scholzj, line 202: Is this still a good test case given this field is deprecated? Should we test on something what will be used more commonly?

topic-operator/src/test/java/io/strimzi/operator/topic/TopicOperatorMockTest.java
```diff
@@ -99,7 +99,8 @@ public void setup(VertxTestContext context) throws Exception {
             Config.ZOOKEEPER_CONNECTION_TIMEOUT_MS.key, "30000",
             Config.NAMESPACE.key, "myproject",
             Config.CLIENT_ID.key, "myproject-client-id",
-            Config.FULL_RECONCILIATION_INTERVAL_MS.key, "10000"
+            Config.FULL_RECONCILIATION_INTERVAL_MS.key, "10000",
+            Config.KAFKA_VERSION.key, "3.0.0"
```
scholzj, line 103: Again, hardcoded version ... probably not good.

topic-operator/src/test/java/io/strimzi/operator/topic/TopicOperatorTest.java
```diff
@@ -83,6 +83,7 @@
         MANDATORY_CONFIG.put(Config.CLIENT_ID.key, "default-client-id");
         // Not mandatory, but makes the time test quicker
         MANDATORY_CONFIG.put(Config.TOPIC_METADATA_MAX_ATTEMPTS.key, "3");
+        MANDATORY_CONFIG.put(Config.TC_KAFKA_VERSION, "3.0.0");
```
scholzj, line 86: Same as above.

topic-operator/src/test/java/io/strimzi/operator/topic/TopicSerializationTest.java
```diff
@@ -24,16 +24,20 @@
...
 import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.matchesPattern;
 import static org.junit.jupiter.api.Assertions.fail;
 
 public class TopicSerializationTest {
 
     private final Labels labels = new Labels(
             "app", "strimzi");
 
+    String kafkaVersion = "3.0.0";
```
scholzj, line 39: Same as above and everywhere else.

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/EntityTopicOperator.java
sknot-rh: I do not think it is dependent. TO uses Kafka version just to load the related config model for topics.

cluster-operator/src/test/java/io/strimzi/operator/cluster/ResourceUtils.java
scholzj: So, do we want to keep this even if it is not needed anymore? I guess it might be useful without it to help with the next change which will miss the point that version is not required ;-)

topic-operator/src/test/java/io/strimzi/operator/topic/TopicSerializationTest.java
scholzj: Does this still need to be fixed for the `Utils.getLatestKafkaVersion()`?

topic-operator/src/test/resources/test.properties
```diff
@@ -0,0 +1 @@
+kafka.version=${kafka.version}
```
scholzj, line 1: Where is this taking the version from?

sknot-rh: Main pom

scholzj: Ok, thanks.

scholzj (approved):
LGTM. But should be reviewed by @tombentley as well.

config-model-generator/src/main/java/io/strimzi/build/kafka/metadata/CommonConfigModelGenerator.java
```diff
@@ -0,0 +1,202 @@
...
+
+    protected Field getField(Class<?> cls, String fieldName) {
+        return AccessController.doPrivileged((PrivilegedAction<Field>) () -> {
+            try {
+                Field f1 = cls.getDeclaredField(fieldName);
+                f1.setAccessible(true);
+                return f1;
+            } catch (ReflectiveOperationException e) {
+                throw new RuntimeException(e);
+            }
```
tombentley, line 140: It would be cleaner to catch the `NoSuchFieldException` here and return `null`, rather than catching `NoSuchFieldException` in `getOneOfFields`. (I'm not really a fan of using exceptions to indicate something that's not an error condition, but could be reported via the method result).

config-model-generator/src/main/java/io/strimzi/build/kafka/metadata/CommonConfigModelGenerator.java
```diff
@@ -0,0 +1,202 @@
...
+import java.lang.reflect.InvocationTargetException;
+import java.security.AccessController;
+import java.security.PrivilegedAction;
+import java.util.List;
+import java.util.Map;
+import java.util.Properties;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+public abstract class CommonConfigModelGenerator {
```
tombentley, line 30: Can we add a sentence of Javadoc?

config-model-generator/src/main/java/io/strimzi/build/kafka/metadata/CommonConfigModelGenerator.java
```diff
@@ -0,0 +1,202 @@
...
+
+        String minStr = matcher.group(1);
+        if (minStr.contains("e") || minStr.contains(".") && !minStr.contains("..")) {
+            descriptor.setMinimum(Double.parseDouble(minStr));
+        } else if (!"...".equals(minStr)) {
+            descriptor.setMinimum(Long.parseLong(minStr));
+        }
+    }
+
+    protected void openRange(Matcher matcher, ConfigModel descriptor) {
```
tombentley, line 187: Can be private

config-model-generator/src/main/java/io/strimzi/build/kafka/metadata/KafkaConfigModelGenerator.java
tombentley: Yes, I would.

config-model-generator/src/main/java/io/strimzi/build/kafka/metadata/KafkaTopicConfigModelGenerator.java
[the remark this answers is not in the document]
tombentley: There's still ~50 lines of duplicated code here. Surely there's scope to refactor and apply the abstract method pattern? I'd expect at least the `if/else if` within the loop could be factored out. Does it matter if there are validators use for Kafka configs and not for Topic configs? It just means we don't always hit that branch in the topic case, but that's not a problem, right?

topic-operator/src/main/java/io/strimzi/operator/topic/TopicSerialization.java
```diff
@@ -45,52 +50,82 @@
...
+     */
+    private static List<String> validate(Map<String, String> topicConfiguration, String kafkaVersion) {
+        List<String> errors = new ArrayList<>();
+        Map<String, ConfigModel> models = readConfigModel(kafkaVersion);
+        for (Map.Entry<String, String> entry: topicConfiguration.entrySet()) {
+            String key = entry.getKey();
+            String value = entry.getValue();
+            ConfigModel config = models.get(key);
+            if (config != null) {
+                errors.addAll(config.validate(key, value));
```
tombentley, line 87: Do we need to keep the type coercion code, in order to avoid regression? Or are we expecting the bidirectionality means that previously even if the user's initial CR used a type that got coerced, it would have been converted to some canonical type following reconciliation?

topic-operator/src/test/java/io/strimzi/operator/topic/TopicOperatorTest.java
```diff
@@ -153,14 +154,14 @@ public void testOnKafkaTopicAdded_ignorable(VertxTestContext context) {
...
-                .withMetadata(new ObjectMetaBuilder().withName("invalid").withLabels(labels.labels()).build())
+                .withMetadata(new ObjectMetaBuilder().withNamespace("test").withName("invalid").withLabels(labels.labels()).build())
                 .withNewSpec()
                     .withReplicas(1)
                     .withPartitions(1)
                     .withConfig(singletonMap(null, null))
                 .endSpec()
             .build();
-        String errorMessage = "KafkaTopic's spec.config has invalid entry: The key 'null' of the topic config is invalid: The value corresponding to the key must have a string, number or boolean value but the value was null";
+        String errorMessage = "KafkaTopic test/invalid has invalid spec.config: null with value 'null' is not one of the known options";
```
tombentley, line 164: This looks suspicious, like we haven't handled something being null.

topic-operator/src/test/java/io/strimzi/operator/topic/Utils.java
```diff
@@ -49,4 +52,18 @@ public static TopicMetadata getTopicMetadata(Topic kubeTopic) {
                 partitions), new Config(configs));
     }
 
+    public static String getLatestKafkaVersion() {
```
tombentley, line 55: Wouldn't it be better to move `io.strimzi.operator.cluster.KafkaVersionTestUtils` to common, than reinvent similar functionality?

config-model-generator/src/main/java/io/strimzi/build/kafka/metadata/CommonConfigModelGenerator.java
```diff
@@ -0,0 +1,259 @@
...
+            return null;
+        });
+        return method;
+    }
+
+    protected abstract ConfigDef configsFromReflectivity();
+
+    protected abstract ConfigModel createDescriptor(Method getConfigValueMethod, ConfigDef def, ConfigDef.ConfigKey key, Map<String, String> dynamicUpdates) throws InvocationTargetException, IllegalAccessException;
+
+    protected void addConfigModel(Method getConfigValueMethod, ConfigDef.ConfigKey key, ConfigDef def, Map<String, String> dynamicUpdates, Map<String, ConfigModel> result) throws InvocationTargetException, IllegalAccessException {
```
tombentley, line 227: When we have long parameter list can we use parameter per line?

config-model-generator/src/main/java/io/strimzi/build/kafka/metadata/CommonConfigModelGenerator.java
```diff
@@ -0,0 +1,259 @@
...
+    protected Method methodFromReflection(Class<ConfigDef> origin, String methodName, Class... paramClasses) throws NoSuchMethodException {
+        Method method = origin.getDeclaredMethod(methodName, paramClasses);
+        doPrivileged((PrivilegedAction<ConfigDef>) () -> {
+            method.setAccessible(true);
+            return null;
+        });
+        return method;
+    }
+
+    protected abstract ConfigDef configsFromReflectivity();
```
tombentley, line 223: Think about the method naming: Does this class know how subclasses are obtaining the `ConfigDef`? Probably not, so it probably shouldn't mention the reflection implementation detail in the method name. Calling it `configDefs` would be enough.

topic-operator/src/main/java/io/strimzi/operator/topic/TopicSerialization.java
sknot-rh: The coercion of type is preserved.

tombentley (approved):
LGTM, thanks!

im-konge (approved):

scholzj:
Should this be merged?

sknot-rh:
Yes, I think it can be merged.

PR #5884, opened 2021-11-12
Remove StatefulSets from ZooKeeper

scholzj:
### Type of change

- Enhancement / new feature

### Description

This is a PoC PR of the StatefulSet removal. It adds a new feature gate `StatefulSetRemoval` and new CRD `KafkaNodeSet`. When the feature gate is enabled, it will use the node sets instead of the StatefulSets to run ZooKeeper. It implements the controller to manage the pods based on the node sets. It should support all ZooKeeper functionality right now:
* rolling updates
* scaling up or down
* manual rolling updates
* CA renewals
* pod and pvc deletions

The controller is implemented as a lightweight controller with a worker queue running in its own thread and not depending on Vert.x. More details about the motivation and some more insights are described in the corresponding proposal [SP#31](https://github.com/strimzi/proposals/blob/main/031-statefulset-removal.md).

It is currently a draft PR and it is missing any tests and documentation. Assuming the code and the proposal are well received, these will done before merging the PR. The Kafka cluster is currently not affected in any way and is always using StatefulSets. That would be done in a separate PR (assuming the PoC and the proposal are accepted).

### Checklist

- [ ] Write tests
- [ ] Update documentation
- [ ] Update CHANGELOG.md


cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaNodeSetController.java
```diff
@@ -0,0 +1,478 @@
...
+     * @param nodeSet   KafkaNodeSet which should be checked
+     *
+     * @return          True if the KafkaNodeSet's Kafka cluster matches the selector labels
+     */
+    private boolean matchesCrSelector(KafkaNodeSet nodeSet)    {
+        if (nodeSet.getMetadata().getLabels() != null
+                && nodeSet.getMetadata().getLabels().get(Labels.STRIMZI_CLUSTER_LABEL) != null) {
+            // We find the matching Kafka cluster and check the CR selector
+            String kafkaClusterName = nodeSet.getMetadata().getLabels().get(Labels.STRIMZI_CLUSTER_LABEL);
+            Kafka kafka = kafkaLister.namespace(nodeSet.getMetadata().getNamespace()).get(kafkaClusterName);
```
tombentley, line 195: AFAICS this doesn't really depend on this resource being a `Kafka` CR. Any resource could be used, it's just we only care about the Kafka case, right?

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaNodeSetController.java
```diff
@@ -0,0 +1,478 @@
...
+
+                if (podWithOwnerReference.getMetadata().getOwnerReferences() != null)   {
+                    //List<OwnerReference> owners = new ArrayList<>()
+                    podWithOwnerReference.getMetadata().getOwnerReferences().add(owner);
+                } else {
+                    podWithOwnerReference.getMetadata().setOwnerReferences(List.of(owner));
+                }
+
+                podOperator.client().inNamespace(reconciliation.namespace()).patch(podWithOwnerReference);
+            }
```
tombentley, line 317: Is it possible for two Kafka CRs to map to the same `KafkaNodeSet` (e.g. if the selector labels of one are a subset of the other)?

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaNodeSetController.java
```diff
@@ -0,0 +1,478 @@
...
+    public void run() {
+        LOGGER.infoOp("Starting KafkaNodeSet controller for namespace {}", watchedNamespace);
+
+        LOGGER.infoOp("Waiting for informers to sync");
+        while (!podInformer.hasSynced() && !kafkaNodeSetInformer.hasSynced() && kafkaInformer.hasSynced())   {
+            // Nothing to do => just loop
+        }
+        LOGGER.infoOp("Informers are in-sync");
+
+        while (!stop) {
```
tombentley, line 385: `stop` will need to be volatile, since it will get changed from another thread.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaNodeSetController.java
```diff
@@ -0,0 +1,478 @@
...
+
+    /**
+     * The run loop of the controller thread. It picks reconciliations from the work queue and executes them.
+     */
+    @Override
+    public void run() {
+        LOGGER.infoOp("Starting KafkaNodeSet controller for namespace {}", watchedNamespace);
+
+        LOGGER.infoOp("Waiting for informers to sync");
+        while (!podInformer.hasSynced() && !kafkaNodeSetInformer.hasSynced() && kafkaInformer.hasSynced())   {
```
tombentley, line 380: Probably want `stop` in here too, just to be sure.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaNodeSetController.java
```diff
@@ -0,0 +1,478 @@
...
+
+    /**
+     * Stops the controller: this method sets the stop flag and the run loop will exit when the flag is set.
+     */
+    public void stop()  {
+        this.stop = true;
+    }
+
+    /**
+     * Helped class to track the pod counts during reconciliation and to pass through different methods. This is used to
```
tombentley, line 409:
```suggestion
     * Helper class to track the pod counts during reconciliation and to pass through different methods. This is used to
```

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaNodeSetController.java
```diff
@@ -0,0 +1,478 @@
...
+     * count the numbers for the KafkaNodeSet status subresource.
+     */
+    static class PodCounter    {
+        int pods = 0;
+        int readyPods = 0;
+        int currentPods = 0;
+    }
+
+    /**
+     * Helped class to track the pod counts during reconciliation and to pass through different methods. This simplified
```
tombentley, line 419:
```suggestion
     * Helper class to track the pod counts during reconciliation and to pass through different methods. This simplified
```

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaNodeSetController.java
```diff
@@ -0,0 +1,478 @@
...
+    private static final ObjectMapper MAPPER = new ObjectMapper();
+
+    private boolean stop = false;
+
+    private final PodOperator podOperator;
+    private final CrdOperator<KubernetesClient, KafkaNodeSet, KafkaNodeSetList> kafkaNodeSetOperator;
+    private final Optional<LabelSelector> crSelector;
+    private final String watchedNamespace;
+
+    private final BlockingQueue<SimplifiedReconciliation> workQueue = new ArrayBlockingQueue<>(1024);
```
tombentley, line 61: What are the advantages of using our own work queue and worker thread over using a single threaded `Executor`?

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaNodeSetController.java
scholzj: Well, right now, the parent of the KafkaNodeSet is always Kafka CR and that is where we need to check the label selector. We do not use it with any other resources. If we one day use it also with Connect or Bridge etc., we would need to extend this. But not sure we need to invest into that now.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaNodeSetController.java
scholzj: The selector is currently among other things  based on the name of the KafkaNodeSet. So no, it is not possible. Plus the same selector is used today with StatefulSets or Deployments where it would cause issues as well if there were any conflicts.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaNodeSetController.java
scholzj: I didn't really consider `Executor` or `ExecutorService`. I follow this is a more common controller pattern. The only advantage I see would be if we use a thread pool to parallelize this. But that brings additional complexity around locking etc. So I'm not sure I see any benefits in that. Do you see any benefits in using Executor / ExecutorService?

cluster-operator/src/main/java/io/strimzi/operator/cluster/FeatureGates.java
```diff
@@ -18,10 +18,12 @@
 
     private static final String CONTROL_PLANE_LISTENER = "ControlPlaneListener";
     private static final String SERVICE_ACCOUNT_PATCHING = "ServiceAccountPatching";
+    private static final String STATEFULSET_REMOVAL = "StatefulSetRemoval";
```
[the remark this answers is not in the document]
ppatierno: Yeah that was my same feeling and a point I made on the proposal about using something like `KafkaNodeSetEnabled` (but even `UseKafkaNodeSet` sounds fine to me).

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaNodeSetController.java
ppatierno: @tombentley do you see any specific use case for doing that?

tombentley: No, I was just asking in case it was something we needed specifically to handle. But it doesn't sound like it.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaNodeSetController.java
tombentley: Nothing completely convincing:

1. Not having to think so hard about threading issues (e.g. around stopping the worker thread).
2. Built-in support for cancellation (right now it doesn't look like we need that).
3. An abstraction which could be useful (e.g. for testing purposes).

cluster-operator/src/main/java/io/strimzi/operator/cluster/FeatureGates.java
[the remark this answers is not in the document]
scholzj: Ok, lets change it once we agree how should the resource be called :-D

scholzj:
@ppatierno @tombentley This should now be ready for review!

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/AbstractModel.java
```diff
@@ -369,6 +377,17 @@ protected Labels getLabelsWithStrimziName(String name, Map<String, String> addit
...
     }
 
+    /**
+     * @param name the value for the {@code strimzi.io/name} key
+     * @param podName the value for the {@code strimzi.io/pod-name} key
+     * @param additionalLabels a nullable map of additional labels to be added to this instance of Labels
+     *
+     * @return Labels object with the default labels merged with the provided additional labels and the new {@code strimzi.io/name} label
+     */
+    protected Labels getLabelsWithStrimziName(String name, String podName, Map<String, String> additionalLabels) {
```
ppatierno, line 387: isn't it better having just `getLabelsWithStrimzi`? The `Name` part seems to imply that you look for only `strimzi.io/name` label (and you are using `withStrimziName` inside) not for `PodName` as well (`strimzi.io/pod-name` label).

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/AbstractModel.java
```diff
@@ -1075,6 +1094,137 @@ protected StatefulSet createStatefulSet(
...
+            Function<String, List<Volume>> volumes,
+            Affinity affinity,
+            List<Container> initContainers,
+            List<Container> containers,
+            List<LocalObjectReference> imagePullSecrets,
+            boolean isOpenShift)  {
+        List<Map<String, Object>> pods = new ArrayList<>(replicas);
+
+        for (int i = 0; i < replicas; i++)  {
+            String podName = ZookeeperCluster.zookeeperPodName(cluster, i);
```
ppatierno, line 1125: I know that this PR is just for removing SS for ZooKeeper and not for Kafka but could we try to make this method more general for using for the future SS removal for Kafka as well? It's even in the `AbstractModel` class that should not have any specific references to ZK or Kafka. Maybe we could pass a Java `Function` to provide the pod name?

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/StrimziPodSetController.java
```diff
@@ -0,0 +1,490 @@
...
+    /**
+     * Creates the StrimziPodSet controller. The controller should normally exist once per operator for cluster-wide mode
+     * or once per namespace for namespaced mode.
+     *
+     * @param watchedNamespace      Namespace which should be watched. Use * for all namespaces.
+     * @param crSelectorLabels      Selector labels for custom resource managed by this operator instance. This is used
+     *                              to check that the pods belong to a Kafka cluster matching these labels.
+     * @param kafkaOperator         Kafka Operator for getting the Kafka custom resources
+     * @param strimziPodSetOperator  StrimziPodSet Operator used to manage the StrimziPodSet resources - get them, update their status etc.
+     * @param podOperator           Pod oeprator for managing pods
```
ppatierno, line 78: typo "oeprator"

api/src/main/java/io/strimzi/api/kafka/model/StrimziPodSet.java
```diff
@@ -0,0 +1,172 @@
...
+                        ),
+                    @Crd.Spec.AdditionalPrinterColumn(
+                            name = "Current Pods",
+                            description = "Number of up-to-date pods managed by the StrimziPodSetController",
+                            jsonPath = ".status.currentPods",
+                            type = "integer"
+                        ),
+                    @Crd.Spec.AdditionalPrinterColumn(
+                            name = "Age",
+                            description = "Age of the StrimziPodSetController",
```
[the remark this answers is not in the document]
scholzj: The `StrimziPodSetController` is probably some bad copy paste from when I was renaming the custom resource. It should be just `StrimziPodSet`. It is age of the `StrimziPodSet` resource. It is the same as when you do for example `kubectl get pods`

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/AbstractModel.java
scholzj: I renamed it to `getLabelsWithStrimziNameAndPodName` -> I think that is in line with `getLabelsWithStrimziNameAndDiscovery` and other methods. The Pod name in the label is significant as it is needed for the StrimziPodSet pods only.

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/AbstractModel.java
```diff
@@ -1075,6 +1094,137 @@ protected StatefulSet createStatefulSet(
...
+                    .build();
+        }
+
+        Pod pod = new PodBuilder()
+                .withNewMetadata()
+                    .withName(podName)
+                    .withLabels(getLabelsWithStrimziName(name, podName, templatePodLabels).withStatefulSetPod(podName).withStrimziPodSetController(strimziPodSetName).toMap())
+                    .withNamespace(namespace)
+                    .withAnnotations(Util.mergeLabelsOrAnnotations(podAnnotations, templatePodAnnotations))
+                    //.withOwnerReferences(createOwnerReference())
```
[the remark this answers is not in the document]
scholzj: I deleted the line. The OwnerReference is managed by the `StrimziPodSetController` when creating the pods. That way, it is not part of the revision, but it also makes it easier for the controller to set it to pods when something in the owner reference changes etc.

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/AbstractModel.java
```diff
@@ -1406,6 +1556,28 @@ protected PodDisruptionBudget createPodDisruptionBudget()   {
                 .build();
     }
 
+    /**
+     * Creates a PodDisruptionBudget for use with Custom pod controller. Custom pod controllers can use only PDBs with
```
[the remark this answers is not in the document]
scholzj: I expanded the comment to make it more clear. Basically, custom controller is any non-built-in controller which managed pods. The built-in controllers are StatefulSets, DaemonSets, ReplicaSets etc. Our controller trough StrimziPodSets is custom controller.

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/AbstractModel.java
scholzj: Good point. Fixed.

cluster-operator/src/main/java/io/strimzi/operator/cluster/ClusterOperator.java
```diff
@@ -95,6 +97,10 @@ public void start(Promise<Void> start) {
         // Configure the executor here, but it is used only in other places
         getVertx().createSharedWorkerExecutor("kubernetes-ops-pool", config.getOperationsThreadPoolSize(), TimeUnit.SECONDS.toNanos(120));
 
+        strimziPodSetController = new StrimziPodSetController(namespace, config.getCustomResourceSelector(), resourceOperatorSupplier.kafkaOperator, resourceOperatorSupplier.strimziPodSetOperator, resourceOperatorSupplier.podOperations);
+        strimziPodSetControllerThread = new Thread(strimziPodSetController, "StrimziPodSetController");
+        strimziPodSetControllerThread.start();
```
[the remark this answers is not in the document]
scholzj: I think it is better this way ... the party running the controller controls how it runs -> gives you more flexibility how to handle things etc. But I do not feel strongly about it.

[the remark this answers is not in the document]
ppatierno: I think putting the thread inside the controller is just about hiding the logic but you are still providing `start/stop` to the part running the controller; should it really know that it's a thread or something different? isn't it enough providing it an API to do so?

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/AbstractModel.java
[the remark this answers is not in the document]
ppatierno: Ok that makes sense. Thanks.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/resource/PodRevision.java
```diff
@@ -0,0 +1,106 @@
...
+     * @param desiredPodSet    The StrimziPodSet resource with the desired Pods
+     *
+     * @return                  True if the revision changed. False otherwise.
+     */
+    public static boolean hasChanged(Pod currentPod, StrimziPodSet desiredPodSet)   {
+        Pod desiredPod = desiredPodSet
+                .getSpec()
+                .getPods()
+                .stream()
+                .map(pod -> MAPPER.convertValue(pod, Pod.class))
```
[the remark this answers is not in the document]
scholzj: It is not used because it cannot be translated into CRD with our `crd-generator` and writing our custom model for the whole Pod structure seemed unnecessary. The only thing we loose by this is validation -> that is anyway semi-functional and not so important given it is an internal resource only.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/StrimziPodSetController.java
```diff
@@ -0,0 +1,491 @@
...
+     */
+    @Override
+    public void run() {
+        LOGGER.infoOp("Starting StrimziPodSet controller for namespace {}", watchedNamespace);
+
+        LOGGER.infoOp("Waiting for informers to sync");
+        while (!stop
+                && !podInformer.hasSynced()
+                && !strimziPodSetInformer.hasSynced()
+                && kafkaInformer.hasSynced())   {
```
ppatierno, line 389: should you wait for `kafkaInformer` to be synced? so the condition `!kafkaInformer.hasSynced()` ?

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaAssemblyOperator.java
```diff
@@ -3530,11 +3708,26 @@ private boolean isCustomCertUpToDate(StatefulSet sts, Pod pod, String annotation
...
 
+            if (ctrlResource instanceof StatefulSet) {
+                StatefulSet sts = (StatefulSet) ctrlResource;
+
+                if (!isPodUpToDate(sts, pod)) {
+                    reasons.add("Pod has old generation");
+                }
+
+                if (!isCustomCertUpToDate(sts, pod, KafkaCluster.ANNO_STRIMZI_CUSTOM_LISTENER_CERT_THUMBPRINTS)) {
+                    reasons.add("custom certificate one or more listeners changed");
```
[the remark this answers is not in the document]
scholzj: At this point, we support only ZooKeeper. And ZooKeeper has no custom listener certificates. I expect that this will need to be handled when we add support for Kafka. But adding it now makes little sense since it cannot be tested etc.

api/src/main/java/io/strimzi/api/kafka/model/StrimziPodSetSelector.java
```diff
@@ -0,0 +1,51 @@
...
+import java.util.Map;
+
+@Buildable(
+        editableEnabled = false,
+        builderPackage = Constants.FABRIC8_KUBERNETES_API
+)
+@JsonInclude(JsonInclude.Include.NON_NULL)
+@JsonPropertyOrder({"matchLabels"})
+@EqualsAndHashCode
+public class StrimziPodSetSelector implements UnknownPropertyPreserving, Serializable {
```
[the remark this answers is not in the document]
scholzj: Deleted, I didn't used it at the end.

cluster-operator/src/main/java/io/strimzi/operator/cluster/ClusterOperator.java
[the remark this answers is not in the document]
scholzj: Ok, I changed it. Please have a look.

ppatierno (approved):
I have no more relevant issues with this. I tried and played with it and seems to work. Thanks!

tombentley (approved):
Thanks, LGTM

PR #5889, opened 2021-11-15
Fix Grafana panels using filtered count queries

nickgooding:
### Type of change

- Bugfix

### Description

Fixes two Grafana panels that could show out-of-date data. Since Prometheus will return an empty vector rather than 0 with a query like `count(x < 1)` when all samples are filtered out, these Grafana panels would erroneously display the most recent non-zero value within the selected time range.

### Checklist

- [ ] Write tests
- [ ] Update documentation
- [ ] Update CHANGELOG.md


scholzj (approved):
Thanks for the PR.

sknot-rh (approved):

ppatierno (approved):

PR #5909, opened 2021-11-18
Added SCRAM-SHA-256 authentication for Kafka clients

abaus-vc:
### Type of change

Enhancement / new feature

### Description

This modification adds SCRAM-SHA-256 authentication support for Kafka clients. See discussion https://github.com/strimzi/strimzi-kafka-operator/discussions/5855

### Checklist

- [ ] Write tests - not done (see thread)
- [ ] Make sure all tests pass - we had problems running the integration tests against a local minikube, but the same tests fail in the main branch, so it is unlikely this change is responsible
- [x] Update documentation
- [ ] Check RBAC rights for Kubernetes / OpenShift roles - does not apply
- [ ] Reference relevant issue(s) and close them after merging - issue does not exist (yet)
- [x] Update CHANGELOG.md
- [ ] Supply screenshots for visual changes, such as Grafana dashboards - does not apply

### DIFF
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/model/AuthenticationUtils.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/model/AuthenticationUtils.java
@@ -112,8 +112,8 @@ public class AuthenticationUtils {
                 KafkaClientAuthenticationPlain passwordAuth = (KafkaClientAuthenticationPlain) authentication;
                 addNewVolume(volumeList, volumeNamePrefix, passwordAuth.getPasswordSecret().getSecretName(), isOpenShift);
             } else if (authentication instanceof KafkaClientAuthenticationScram) {
-                KafkaClientAuthenticationScram passwordAuth = (KafkaClientAuthenticationScram) authentication;
-                addNewVolume(volumeList, volumeNamePrefix, passwordAuth.getPasswordSecret().getSecretName(), isOpenShift);
+                KafkaClientAuthenticationScram scramAuth = (KafkaClientAuthenticationScram) authentication;
+                addNewVolume(volumeList, volumeNamePrefix, scramAuth.getPasswordSecret().getSecretName(), isOpenShift);
             } else if (authentication instanceof KafkaClientAuthenticationOAuth) {
                 KafkaClientAuthenticationOAuth oauth = (KafkaClientAuthenticationOAuth) authentication;
                 volumeList.addAll(configureOauthCertificateVolumes(oauthVolumeNamePrefix, oauth.getTlsTrustedCertificates(), isOpenShift));
@@ -183,8 +183,8 @@ public class AuthenticationUtils {
                 KafkaClientAuthenticationPlain passwordAuth = (KafkaClientAuthenticationPlain) authentication;
                 volumeMountList.add(VolumeUtils.createVolumeMount(volumeNamePrefix + passwordAuth.getPasswordSecret().getSecretName(), passwordVolumeMount + passwordAuth.getPasswordSecret().getSecretName()));
             } else if (authentication instanceof KafkaClientAuthenticationScram) {
-                KafkaClientAuthenticationScram passwordAuth = (KafkaClientAuthenticationScram) authentication;
-                volumeMountList.add(VolumeUtils.createVolumeMount(volumeNamePrefix + passwordAuth.getPasswordSecret().getSecretName(), passwordVolumeMount + passwordAuth.getPasswordSecret().getSecretName()));
+                KafkaClientAuthenticationScram scramAuth = (KafkaClientAuthenticationScram) authentication;
+                volumeMountList.add(VolumeUtils.createVolumeMount(volumeNamePrefix + scramAuth.getPasswordSecret().getSecretName(), passwordVolumeMount + scramAuth.getPasswordSecret().getSecretName()));
             } else if (authentication instanceof KafkaClientAuthenticationOAuth) {
                 KafkaClientAuthenticationOAuth oauth = (KafkaClientAuthenticationOAuth) authentication;
                 volumeMountList.addAll(configureOauthCertificateVolumeMounts(oauthVolumeNamePrefix, oauth.getTlsTrustedCertificates(), oauthCertsVolumeMount));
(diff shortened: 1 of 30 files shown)


cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaMirrorMaker2AssemblyOperator.java
```diff
@@ -367,6 +368,9 @@ private static void addClusterToMirrorMaker2ConnectorConfig(Map<String, Object>
             if (KafkaClientAuthenticationPlain.TYPE_PLAIN.equals(clientAuthType)) {
                 saslMechanism = "PLAIN";
                 jaasConfig = "org.apache.kafka.common.security.plain.PlainLoginModule required username=\"" + authProperties.get(AuthenticationUtils.SASL_USERNAME) + "\" password=\"${file:" + CONNECTORS_CONFIG_FILE + ":" + cluster.getAlias() + ".sasl.password}\";";        
+            } else if (KafkaClientAuthenticationScramSha256.TYPE_SCRAM_SHA_256.equals(clientAuthType)) {
+                saslMechanism = "SCRAM-SHA-256";
+                jaasConfig = "org.apache.kafka.common.security.scram.ScramLoginModule required username=\"" + authProperties.get(AuthenticationUtils.SASL_USERNAME) + "\" password=\"${file:" + CONNECTORS_CONFIG_FILE + ":" + cluster.getAlias() + ".sasl.password}\";";
```
scholzj, line 373: I'm a bit puzzled why is this change needed only in the MirrorMaker2 related class. Why is it not needed for Connect or Bridge or Mirror Maker 1? Are you sure this is the only place where this is needed? (not saying it is, I'm really just wondering)

scholzj (commented):
Thanks for the PR. It looks quite good. I left one questions inline, but two more points:
* Could you please fix the DCO sing-off? The instructions should be under the _Details_ link next to the DCO status
* Writing system tests for this might be quite hard. But I think we should have at least some unit tests that the right configuration is generated etc. I do not see any in this PR. So myybe that is something to look into.

scholzj:
PS: I also looks like you might need to run `make crd_install` to copy the new CRDs to different places.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaMirrorMaker2AssemblyOperator.java
abaus-vc: The other cases are handled by polymorphism (inheriting from the common superclass ```io.strimzi.api.kafka.model.authentication.KafkaClientAuthenticationScram```)

abaus-vc:
I addressed the issues you raised (added unittests, fixed the DCO sign-off and invoked the crd_install make target).

scholzj (commented):
The code looks good. I think the unit tests are still missing.

abaus-vc:
Ah, you are correct, the commit containing the added unittests was still unpushed, sorry. Fixed, please check again.

scholzj (approved):
Great, thanks for the PR.

scholzj:
@abaus-vc I approved it, but while looking into some similar work, I realized this probably doesn't work. Inside the container images, there are shell scripts generating the configuration files. They will need to be updated as well and I do not see it in this PR. As an example, you can have a look here: https://github.com/strimzi/strimzi-kafka-operator/blob/main/docker-images/kafka/scripts/kafka_connect_config_generator.sh#L61-L64 ... but there will be more scripts with similar content generating the configs for the different compoenents.

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/AuthenticationUtils.java
```diff
@@ -254,11 +255,11 @@ public static void configureClientAuthenticationEnvVars(KafkaClientAuthenticatio
                 properties.put(SASL_USERNAME, passwordAuth.getUsername());
                 properties.put(SASL_PASSWORD_FILE, String.format("%s/%s", passwordAuth.getPasswordSecret().getSecretName(), passwordAuth.getPasswordSecret().getPassword()));
                 properties.put(SASL_MECHANISM, KafkaClientAuthenticationPlain.TYPE_PLAIN);
-            } else if (authentication instanceof KafkaClientAuthenticationScramSha512) {
-                KafkaClientAuthenticationScramSha512 passwordAuth = (KafkaClientAuthenticationScramSha512) authentication;
+            } else if (authentication instanceof KafkaClientAuthenticationScram) {
+                KafkaClientAuthenticationScram passwordAuth = (KafkaClientAuthenticationScram) authentication;
```
ppatierno, line 259: I know it was in the code before but `passwordAuth` sounds strange to me ... can we use something like `scramAuth` as we have the `tlsAuth` above? The same comments applies in all other places where we have this `passwordAuth` when it's SCRAM in reality.

scholzj: BTW: SCRAM is password based authentication.

ppatierno: @scholzj I know that of course, but `passwordAuth` doesn't sound to be good enough to me.

abaus-vc:
Regarding the config generator scripts: I only found 2 more locations in addition to the one already pointed out and updated all of them accordingly.

scholzj:
@abaus-vc Could you please rebase it? Some of the shell scripts changed the location around since you opened the PR.

abaus-vc:
I have rebased the code onto the current main.

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/AuthenticationUtils.java
scholzj: Was this renamed as @ppatierno requested?

abaus-vc: It is now.

ppatierno (approved):

scholzj:
Thanks for the PR.

Neustradamus:
@abaus-vc: Thanks a lot for all about SCRAM.

Linked to:
- https://github.com/scram-xmpp/info/issues/1

PR #5914, opened 2021-11-18
Fix incompatible types issue with IBM Semeru JVM

kun-lu20:
Replace withNew*(String) methods with with*(String)

### Type of change

_Select the type of your PR_

- Bugfix

### Description

The incompatible types issue with IBM Semeru 11 Runtime (previously known as AdoptOpenJDK 11 Openj9 JVM) is triggered again by recent commits. Based on  #5494 , this PR is to fix the issue by replacing certain deprecated `withNew*(String)` methods with `with*(String)` methods, so that this project will be successfully compiled and built across different JVMs. 

### Checklist

- [ ] Write tests
- [ ] Update documentation
- [ ] Update CHANGELOG.md


scholzj (approved):

sknot-rh (approved):

scholzj:
@kun-lu20 I'm fine to merge this. But I wonder if it would be more interesting to look into the core issue => which this is a problem for the IBM JDK but works fine on OpenJDK (and as far as I know also for the Oracle JDK). These are generated methods => so why do they not generated on the IBM JDK?

kun-lu20:
Hi @scholzj , when using IBM Semeru runtime to compile, only `withNew*(byte[])` methods will be generated in `Sundrio` produced classes such as `*Builder` classes.

kun-lu20:
@scholzj The root cause might be in here:
https://github.com/sundrio/sundrio/blob/533e4b9b6ecc6ed2c9580c20162d296378e7a6f0/annotations/builder/src/main/java/io/sundr/builder/internal/utils/BuilderUtils.java#L312

The code just chooses the first constructor which has only one argument. Probably on IBM Semeru runtime the String(byte[]) constructor comes first.

Furthermore, the overloaded withNew*() methods are deprecated in recent `Sundrio` releases and we could use with*(String) methods instead which would not cause the issue.

scholzj:
> Hi @scholzj , when using IBM Semeru runtime to compile, only `withNew*(byte[])` methods will be generated in `Sundrio` produced classes such as `*Builder` classes.

Right. So maybe the best way to fix this is to go after the root cause. It is either some issue in Sundrio or in the IBM JDK. Fixing it there will do that once and for all. As long as we are fixing just the usage of these methods, they will be always creeping in because most developers and the CI do not use the IBM JDK.

kun-lu20:
Thanks @scholzj , you're right. The code url I mentioned above could give us some insights of the cause of this issue. We will investigate more and try to fix the root cause so that it won't happen again.

kun-lu20:
Thanks @scholzj @sknot-rh

PR #5918, opened 2021-11-20
Add support for intra-broker disk balancing

kyguy:
### Type of change

- Enhancement / new feature

### Description

Adds the ability to preform intra-broker partition balancing using the `KafkaRebalance` resource so users can balance partitions between disks of a individual Kafka broker with JBOD storage.

In the past. Strimzi would always generate a Cruise Control capacity configuration like this:
```
{
  "brokerCapacities":[
    {
      "brokerId": "-1",
      "capacity": {
        "DISK": "100000",
        "CPU": "100",
        "NW_IN": "10000",
        "NW_OUT": "10000"
      },
      "doc": "This is the default capacity. Capacity unit used for disk is in MB, cpu is 
              in percentage, network throughput is in KB."
    }
  ]
}
```
With this PR,  Strimzi now generates Cruise Control JBOD capacity configuraton when managing a Kafka cluster with JBOD storage
```
{
  "brokerCapacities":[
    {
      "brokerId": "0",
      "capacity": {
        "DISK": {
          "/var/lib/kafka/data-0/kafka-log0" : "1024.0"
          "/var/lib/kafka/data-1/kafka-log0" : "1024.0"
        },
        "CPU": "100",
        "NW_IN": "10000",
        "NW_OUT": "10000"
      },
      "doc": "Capacity for Broker 0"
    },
    {
      "brokerId": "1",
      "capacity": {
        "DISK": {
          "/var/lib/kafka/data-0/kafka-log1" : "1024.0"
          "/var/lib/kafka/data-1/kafka-log1" : "1024.0"
        },
        "CPU": "100",
        "NW_IN": "10000",
        "NW_OUT": "10000"
      },
      "doc": "Capacity for Broker 1"
    },
    {
      "brokerId": "1",
      "capacity": {
        "DISK": {
          "/var/lib/kafka/data-0/kafka-log2" : "1024.0"
          "/var/lib/kafka/data-1/kafka-log2" : "1024.0"
        },
        "CPU": "100",
        "NW_IN": "10000",
        "NW_OUT": "10000"
      },
      "doc": "Capacity for Broker 2"
    }
  ]
}

```
This allows Cruise Control to locate and monitor JBOD disks for preforming partition rebalaning amongst them.

### Checklist

- [x] Write tests
- [x] Update documentation
- [ ] Update CHANGELOG.md


cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaAssemblyOperator.java
```diff
@@ -391,7 +391,7 @@ ReconciliationState createReconciliationState(Reconciliation reconciliation, Kaf
     class ReconciliationState {
         private final String namespace;
         private final String name;
-        private final Kafka kafkaAssembly;
+        private Kafka kafkaAssembly;
```
scholzj, line 394: I don't think you should not do this. There is a reason why this is final. Having it not final can break many different things.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaAssemblyOperator.java
```diff
@@ -3594,7 +3607,6 @@ private boolean isPodUpToDate(StatefulSet sts, Pod pod) {
                             this.cruiseControlMetricsAndLogsConfigMap = logAndMetricsConfigMap;
                             this.cruiseControl = cruiseControl;
                             this.ccDeployment = cruiseControl.generateDeployment(pfa.isOpenshift(), annotations, imagePullPolicy, imagePullSecrets);
-
```
scholzj, line 3597: Why remove this?

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaAssemblyOperator.java
```diff
@@ -1748,6 +1751,16 @@ String zkConnectionString(int connectToReplicas, Function<Integer, String> zkNod
...
 
+        private Kafka updateAssemblyKafkaStorage(Kafka kafkaAssembly, KafkaCluster kafkaCluster) {
+            return new KafkaBuilder(kafkaAssembly)
+                .editSpec()
+                    .editKafka()
+                        .withStorage(kafkaCluster.getStorage())
+                    .endKafka()
+                .endSpec()
+                .build();
+        }
```
scholzj, line 1762: You should not modify it like this. Certainly not here because that might screw up other things. You should use `kafkaCluster.getStorage()` where you need the storage.

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/cruisecontrol/Capacity.java
```diff
@@ -11,36 +11,97 @@
...
     /**
-     * Generate disk capacity configuration from the supplied storage configuration
+     * Generate broker capacity entry for capacity configuration.
+     *
+     * @param brokerId Id of broker
+     * @param diskCapacity Disk capacity configuration
+     * @param doc Documentation for broker entry
+     * @return Broker entry as a JsonObject
+     */
+    private JsonObject generateBrokerCapacity(int brokerId, Object diskCapacity, String doc) {
```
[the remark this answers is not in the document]
kyguy: I was having trouble with the type here since it was messing up the syntax of the returned JsonObject. I found that the diskCapacity needed to be a:
-  String when there are no JBOD disks
-  JsonObject when there are JBOD
to prevent malformed JSON

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/cruisecontrol/Capacity.java
```diff
@@ -11,36 +11,97 @@
...
 import java.util.List;
 
 import static io.strimzi.operator.cluster.model.StorageUtils.parseMemory;
 
 public class Capacity {
-    public static final double DEFAULT_BROKER_DISK_MIB_CAPACITY = 100_000;  // in MiB
-    public static final int DEFAULT_BROKER_CPU_UTILIZATION_CAPACITY = 100;  // as a percentage (0-100)
-    public static final double DEFAULT_BROKER_INBOUND_NETWORK_KIB_PER_SECOND_CAPACITY = 10_000;  // in KiB/s
-    public static final double DEFAULT_BROKER_OUTBOUND_NETWORK_KIB_PER_SECOND_CAPACITY = 10_000;  // in KiB/s
+    private static final int DEFAULT_BROKER_ID = -1;
```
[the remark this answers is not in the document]
kyguy: When using a non-JBOD disk configuration, we can generate a generic capacity configuration that contains a single broker entry, the "default" broker configuration, which will apply to all brokers. CC designates the id of this default broker entry as  `-1`

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/cruisecontrol/Capacity.java
```diff
@@ -11,36 +11,97 @@
...
+    private static final String DEFAULT_BROKER_DOC = "This is the default capacity. Capacity unit used for disk is in MB, cpu is in percentage, network throughput is in KB.";
```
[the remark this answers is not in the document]
kyguy: The CC capacity configuration requires a description for every broker entry, this is just the description used for the default broker entry which will be used for all brokers if a non-JBOD disk configuration is provided.

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/cruisecontrol/Capacity.java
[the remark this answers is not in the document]
scholzj: Can't the String for non-JBOD disks be shared as some JsonObject as well?

api/src/main/java/io/strimzi/api/kafka/model/KafkaRebalanceSpec.java
```diff
@@ -62,6 +63,17 @@ public void setSkipHardGoalCheck(boolean skipHardGoalCheck) {
         this.skipHardGoalCheck = skipHardGoalCheck;
     }
 
+    @Description("Whether to balance load between disks within brokers, known as intra-broker disk balancing (requires JBOD Kafka deployment. ) " +
```
scholzj, line 66: I did not noticed it before ... but what does `load` stand for here? Is it load as in amount of IOPS? Or load as in disk space utilization? Or both? Might be good to clarify this.

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/cruisecontrol/Capacity.java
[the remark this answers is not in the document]
scholzj: Ok, makes sense. Can you add some explanatory comment? It will make the code more readable and understandable even for people who are not CC experts. ;-)

api/src/main/java/io/strimzi/api/kafka/model/KafkaRebalanceSpec.java
```diff
@@ -62,6 +63,17 @@ public void setSkipHardGoalCheck(boolean skipHardGoalCheck) {
...
+            "Default is false.")
```
scholzj, line 67: One more stupid question ... I know is this is a new feature. But is there a reason to disable it by default? It sounds like something useful what would not do any harm if enabled. But maybe I'm missing something,.

kyguy: Never are stupid questions, they are helping me be more thorough! The reason it is disabled by default is because an exception will be thrown if a rebalance proposal is requested with it enabled when JBOD storage is disabled

api/src/main/java/io/strimzi/api/kafka/model/KafkaRebalanceSpec.java
kyguy: Load stand for disk space utilization here, so I have update the description!

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/cruisecontrol/Capacity.java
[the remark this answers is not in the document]
kyguy: I need to play around with this a bit more, I'll get back to you

api/src/main/java/io/strimzi/api/kafka/model/KafkaRebalanceSpec.java
kyguy: Yeah you are right, there are just a couple of things I need to check in the CC code but I don't think this should have any negative consequences. I'll add something for it in the next commit

ppatierno: I would agree with Jakub. Enabled by default and checking if JBOD is used, then at least logging a WARN message to make the user aware that this CC feature won't be used because the Kafka instance is not using JBOD.

PaulRMellor (approved):
Looking good. Some suggestions on the content.

api/src/main/java/io/strimzi/api/kafka/model/KafkaRebalanceSpec.java
kyguy: Good idea guys, I have just added some logic to handle this

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/cruisecontrol/Capacity.java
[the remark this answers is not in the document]
kyguy: Done!

kyguy:
@PaulRMellor thanls for all of your edits, I merged them and added a few tweaks on top of them to address some recent logic changes. Let me know what you think!

api/src/main/java/io/strimzi/api/kafka/model/KafkaRebalanceSpec.java
```diff
@@ -62,6 +63,17 @@ public void setSkipHardGoalCheck(boolean skipHardGoalCheck) {
         this.skipHardGoalCheck = skipHardGoalCheck;
     }
 
+    @Description("Enables intra-broker disk balancing, which balances disk space utilization between disks on the same broker. Only applies to Kafka deployments that use JBOD storage. ) " +
```
scholzj, line 66: You have the closing `)` but no opening `(`.

api/src/main/java/io/strimzi/api/kafka/model/KafkaRebalanceSpec.java
```diff
@@ -62,6 +63,17 @@ public void setSkipHardGoalCheck(boolean skipHardGoalCheck) {
...
+    public boolean isRebalanceDisk() {
+        return rebalanceDisk;
+    }
+
+    public void setRebalanceDisk(boolean rebalanceDisk) {
+        this.rebalanceDisk = rebalanceDisk;
+    }
+
     @Description("A regular expression where any matching topics will be excluded from the calculation of optimization proposals. " +
             "This expression will be parsed by the java.util.regex.Pattern class; for more information on the supported formar " +
```
scholzj, line 78:
```suggestion
            "This expression will be parsed by the java.util.regex.Pattern class; for more information on the supported format " +
```

api/src/main/java/io/strimzi/api/kafka/model/KafkaRebalanceSpec.java
```diff
@@ -62,6 +63,17 @@ public void setSkipHardGoalCheck(boolean skipHardGoalCheck) {
         this.skipHardGoalCheck = skipHardGoalCheck;
     }
 
+    @Description("Enables intra-broker disk balancing, which balances disk space utilization between disks on the same broker. Only applies to Kafka deployments that use JBOD storage. ) " +
+            "Default is true when using a Kafka deployment with JBOD storage and false otherwise.")
+    @JsonInclude(JsonInclude.Include.NON_NULL)
```
scholzj, line 68: Since it defaults to true, it should probably be NON_DEFAULT and not NOT_NULL.

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/cruisecontrol/Capacity.java
```diff
@@ -11,36 +11,100 @@
...
+        String size = "";
+
+        for (SingleVolumeStorage volume : ((JbodStorage) storage).getVolumes()) {
+            name = VolumeUtils.getVolumePrefix(volume.getId());
+            path = AbstractModel.KAFKA_MOUNT_PATH + "/" + name + "/" + AbstractModel.KAFKA_LOG_DIR + idx;
+
+            if (volume instanceof PersistentClaimStorage) {
+                size =  ((PersistentClaimStorage) volume).getSize();
+            } else if (volume instanceof EphemeralStorage) {
+                size = ((EphemeralStorage) volume).getSizeLimit();
```
scholzj, line 96: What if the size limit is null? It is not required value.

ppatierno (approved):
LGTM. I have no additional comments than the ones raised by Jakub.

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/cruisecontrol/Capacity.java
kyguy: Not sure if I tried this, let me get back to you on this

kyguy: I have  added a conditional to set the size to the `DEFAULT_BROKER_DISK_MIB_CAPACITY` when the provided size limit is null.

scholzj (approved):
Ping me to start the STs after the regular build completes.

kyguy:
@scholzj There is just one last thing I need to double check with the new system test I added, I'll ping you once I have confirmed it is fine and the regular build completes!

api/src/main/java/io/strimzi/api/kafka/model/KafkaRebalanceSpec.java
kyguy: Alright after much experimentation and investigation of the Cruise Control codebase I have found that inter-broker balancing and intra-broker balancing can only happen as separate requests. This means that JBOD users would need two separate sequential requests to balance between brokers and JBOD disks. For this reason, I am going to revert the `rebalanceDisk` to default to `false` since CC only preforms an intra-broker balance when `rebalanceDisk` is set to `true`!

TL;DR In case I forget, I am going to write down how the goals are chosen when `rebalanceDisks` parameter is set to true. First there is a check to make sure no other goals are specified in the request, if there are CC will throw an error, if not CC will add the intra-broker goals to the goal list [1]. Next, CC will create the `goalsByPriority` list using this goal list [2]. Then the GoalOptimizer will iterate through this `goalsByPriority` list (which now only includes intra-broker goals) to come up with an optimization proposal based on those goals [3]

[1] https://github.com/linkedin/cruise-control/blob/migrate_to_kafka_2_4/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/parameters/ParameterUtils.java#L776-L780
[2] https://github.com/linkedin/cruise-control/blob/migrate_to_kafka_2_4/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/KafkaCruiseControlUtils.java#L469-L474
[3] https://github.com/linkedin/cruise-control/blob/migrate_to_kafka_2_4/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/analyzer/GoalOptimizer.java#L443

scholzj: I guess it should be added to the docs directly here since you already modify them?

documentation/modules/cruise-control/con-rebalance-performance.adoc
```diff
@@ -44,6 +44,13 @@ These strategies can be configured as a sequence.
 The first strategy attempts to compare two partition reassignments using its internal logic. 
 If the reassignments are equivalent, then it passes them to the next strategy in the sequence to decide the order, and so on.
 
+[discrete]
+== Intra-broker disk balancing
+
+Moving a large amount of data between disks on the same broker has less impact than between separate brokers
+If you are running a Kafka deployment that uses JBOD storage, Cruise Control can balance partitions between disks on the same broker. 
+To preform an intra-broker disk balance set `rebalanceDisk` to `true` under the `KafkaRebalance.spec`.
```
PaulRMellor, line 52:
```suggestion
To preform an intra-broker disk balance, set `rebalanceDisk` to `true` under the `KafkaRebalance.spec`.
```

PaulRMellor (approved):
Looks good. :+1:

documentation/modules/cruise-control/ref-cruise-control-configuration.adoc
```diff
@@ -92,10 +92,13 @@ Capacity limits can be set for the following broker resources:
 * `inboundNetwork`  - Inbound network throughput in byte units per second (Default: 10000KiB/s)
 * `outboundNetwork` - Outbound network throughput in byte units per second (Default: 10000KiB/s)
 
-For disk storage, use a number value with either standard Kubernetes byte units (K, M, G, or T), their bibyte (power of two) equivalents (Ki, Mi, Gi, or Ti), or a byte value with or without E notation.
-For network throughput, use an integer value with standard Kubernetes byte units (K, M, G) or their bibyte (power of two) equivalents (Ki, Mi, Gi) per second.
```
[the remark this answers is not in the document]
kyguy: Whoops, I accidentally deleted when rebasing sorry!! Just reverted the changes so it is back now,  nice catch!

api/src/main/java/io/strimzi/api/kafka/model/KafkaRebalanceSpec.java
kyguy: Added some new docs here and in the documentation, going to ping for another review once the tests pass

kyguy:
Since intra-broker balancing and inter-broker balancing can only happen as separate balancing events I had to revert some changes I made last week that enabled intra-broker balancing by default when JBOD storage was enabled. I also updated the documentation to explain this, please take another look and let me know what you think!

documentation/modules/cruise-control/ref-cruise-control-configuration.adoc
```diff
@@ -95,7 +95,13 @@ Capacity limits can be set for the following broker resources:
 For disk storage, use a number value with either standard Kubernetes byte units (K, M, G, or T), their bibyte (power of two) equivalents (Ki, Mi, Gi, or Ti), or a byte value with or without E notation.
 For network throughput, use an integer value with standard Kubernetes byte units (K, M, G) or their bibyte (power of two) equivalents (Ki, Mi, Gi) per second.
 
-Because Strimzi Kafka brokers are homogeneous, Cruise Control applies the same capacity limits to every broker it is monitoring.
+For Kafka clusters that do not use JBOD storage, Strimz generates a generic Cruise Control capacity configuration entry that applies to every broker.
```
scholzj, line 98:
```suggestion
For Kafka clusters that do not use JBOD storage, Strimzi generates a generic Cruise Control capacity configuration entry that applies to every broker. 
```

documentation/modules/cruise-control/con-rebalance-performance.adoc
```diff
@@ -44,6 +44,15 @@ These strategies can be configured as a sequence.
 The first strategy attempts to compare two partition reassignments using its internal logic. 
 If the reassignments are equivalent, then it passes them to the next strategy in the sequence to decide the order, and so on.
 
+[discrete]
+== Intra-broker disk balancing
+
+Moving a large amount of data between disks on the same broker has less impact than between separate brokers
+If you are running a Kafka deployment that uses JBOD storage, Cruise Control can balance partitions between disks on the same broker.
```
scholzj, line 51: I think you need to distinguish two different things => use JBOD storage versus has multiple disks. We recommend everyone to use JBOD storage (as a Strimzi storage type) even if they use it with single disk because they can add more later. So to which part does this apply? I assume users using JBOD storage with a single disk should be fine? I think we need to clarify this both here and in other places in the docs / APi docs in this PR.

kyguy: JBOD storage with a single disk should be fine, I assume if a user executes an intra-broker disk balancing while using one disk will just result in a proposal with 0 partition movements but I will test it out just to make sure and get back to you.

scholzj: Yeah, i would expect it to be fine since Cruise Cotnrol does not know different. But the docs are not clear about this. I think you need to always say somehting like _JBOD with multiple disks per broker_ or soemthing like that,

PaulRMellor, line 51:
```suggestion
If you are running a Kafka deployment that uses JBOD storage with multiple disks on the same broker, Cruise Control can balance partitions between the disks. 
```

PaulRMellor: _NOTE: If you are using JBOD storage with a single disk, intra-broker disk balancing between disks is not applicable._  Note after the next sentence

documentation/modules/cruise-control/con-rebalance-performance.adoc
```diff
@@ -44,6 +44,15 @@ These strategies can be configured as a sequence.
 The first strategy attempts to compare two partition reassignments using its internal logic. 
 If the reassignments are equivalent, then it passes them to the next strategy in the sequence to decide the order, and so on.
 
+[discrete]
+== Intra-broker disk balancing
+
+Moving a large amount of data between disks on the same broker has less impact than between separate brokers
+If you are running a Kafka deployment that uses JBOD storage, Cruise Control can balance partitions between disks on the same broker. 
+To preform an intra-broker disk balance and only an intra-broker disk balance, set `rebalanceDisk` to `true` under the `KafkaRebalance.spec`.
+When setting `rebalanceDisk` to `true`, do not set a `goals` field in the `KafkaRebalance.spec`, Cruise Control will automatically set the intra-broker goals and ignore the inter-broker goals.
```
PaulRMellor, line 53:
```suggestion
When setting `rebalanceDisk` to `true`, do not set a `goals` field in the `KafkaRebalance.spec`, as Cruise Control will automatically set the intra-broker goals and ignore the inter-broker goals.
```

api/src/main/java/io/strimzi/api/kafka/model/KafkaRebalanceSpec.java
```diff
@@ -62,8 +63,19 @@ public void setSkipHardGoalCheck(boolean skipHardGoalCheck) {
         this.skipHardGoalCheck = skipHardGoalCheck;
     }
 
+    @Description("Enables intra-broker disk balancing, which balances disk space utilization between disks on the same broker. Only applies to Kafka deployments that use JBOD storage. " +
```
PaulRMellor, line 66:
```suggestion
    @Description("Enables intra-broker disk balancing, which balances disk space utilization between disks on the same broker. Only applies to Kafka deployments that use JBOD storage with multiple disks. " +
```

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceAssemblyOperator.java
```diff
@@ -329,6 +331,14 @@ private String rebalanceStateConditionType(KafkaRebalanceStatus status) {
         if (kafkaRebalanceSpec.isSkipHardGoalCheck()) {
             rebalanceOptionsBuilder.withSkipHardGoalCheck();
         }
+        if (!usingJbodStorage) {
+            if (kafkaRebalanceSpec.isRebalanceDisk()) {
+                LOGGER.warnOp("Intra-broker balancing only applies to Kafka deployments that use JBOD storage.");
```
PaulRMellor, line 336:
```suggestion
                LOGGER.warnOp("Intra-broker balancing only applies to Kafka deployments that use JBOD storage with multiple disks.");
```

documentation/modules/cruise-control/con-rebalance-performance.adoc
```diff
@@ -81,6 +90,7 @@ For the server setting, use a comma separated string with the fully qualified na
 .2+| `BaseReplicaMovementStrategy`
 | `replicaMovementStrategies`
 
+| `rebalanceDisk`   .1+|  Enables intra-broker disk balancing, which balances disk space utilization between disks on the same broker. Only applies to Kafka deployments that use JBOD storage.               .1+| false
```
PaulRMellor, line 93:
```suggestion
| `rebalanceDisk`   .1+|  Enables intra-broker disk balancing, which balances disk space utilization between disks on the same broker. Only applies to Kafka deployments that use JBOD storage with multiple disks.               .1+| false
```

PaulRMellor (approved):
Looks good, Kyle. I suggested a note for the JBOD/one disk situation. Not sure how much detail we want to go into there.

documentation/modules/cruise-control/con-rebalance-performance.adoc
```diff
@@ -44,6 +44,15 @@ These strategies can be configured as a sequence.
 The first strategy attempts to compare two partition reassignments using its internal logic. 
 If the reassignments are equivalent, then it passes them to the next strategy in the sequence to decide the order, and so on.
 
+[discrete]
+== Intra-broker disk balancing
+
+Moving a large amount of data between disks on the same broker has less impact than between separate brokers
```
ppatierno, line 50: Missing full stop at the end of the sentence.

documentation/modules/cruise-control/con-rebalance-performance.adoc
ppatierno: I would not use the wording "is not applicable" (at least as it sounds as non native speaker). I would specify, as Kyle mentioned, that it would result in a proposal with 0 partition movements. Using "is not applicable" sounds to me as it doesn't work at all and could be confusing imho.

documentation/modules/cruise-control/con-rebalance-performance.adoc
```diff
@@ -44,6 +44,15 @@ These strategies can be configured as a sequence.
...
+If you are running a Kafka deployment that uses JBOD storage, Cruise Control can balance partitions between disks on the same broker. 
+To preform an intra-broker disk balance and only an intra-broker disk balance, set `rebalanceDisk` to `true` under the `KafkaRebalance.spec`.
```
ppatierno, line 52: typo "perform"

ppatierno: Other than the typo, I didn't get what the sentence mean ... is there a repetition?

"To preform an intra-broker disk balance and only an intra-broker disk balance" ... what do you mean?

documentation/modules/cruise-control/con-rebalance-performance.adoc
```diff
@@ -44,6 +44,15 @@ These strategies can be configured as a sequence.
...
+When setting `rebalanceDisk` to `true`, do not set a `goals` field in the `KafkaRebalance.spec`, Cruise Control will automatically set the intra-broker goals and ignore the inter-broker goals.
+Cruise Control does not prefrom inter-broker and intra-broker balancing at the same time.
```
ppatierno, line 54: typo "perform"

documentation/modules/cruise-control/con-rebalance-performance.adoc
kyguy: I meant to imply that there is no inter-broker balance happening when an intra-broker balance it preformed but it was not clear so I removed it!

kyguy:
Thanks for the feedback everyone, I made some edits to the docs based on the it

api/src/main/java/io/strimzi/api/kafka/model/KafkaRebalanceSpec.java
scholzj: > Since it defaults to true, it should probably be NON_DEFAULT and not NOT_NULL.

This still applies I guess.

documentation/modules/cruise-control/con-rebalance-performance.adoc
```diff
@@ -81,6 +91,7 @@ For the server setting, use a comma separated string with the fully qualified na
 .2+| `BaseReplicaMovementStrategy`
 | `replicaMovementStrategies`
 
+| `rebalanceDisk`   .1+|  Enables intra-broker disk balancing, which balances disk space utilization between disks on the same broker. Only applies to Kafka deployments that use JBOD storage.               .1+| false
```
scholzj, line 94: `Only applies to Kafka deployments that use JBOD storage with multiple disks.`

documentation/modules/cruise-control/ref-cruise-control-configuration.adoc
```diff
@@ -95,7 +95,13 @@ Capacity limits can be set for the following broker resources:
 For disk storage, use a number value with either standard Kubernetes byte units (K, M, G, or T), their bibyte (power of two) equivalents (Ki, Mi, Gi, or Ti), or a byte value with or without E notation.
 For network throughput, use an integer value with standard Kubernetes byte units (K, M, G) or their bibyte (power of two) equivalents (Ki, Mi, Gi) per second.
 
-Because Strimzi Kafka brokers are homogeneous, Cruise Control applies the same capacity limits to every broker it is monitoring.
+For Kafka clusters that do not use JBOD storage, Strimzi generates a generic Cruise Control capacity configuration entry that applies to every broker.
```
scholzj, line 98: ???

```suggestion
For Kafka clusters that use only one disk per broker, Strimzi generates a generic Cruise Control capacity configuration entry that applies to every broker.
```

documentation/modules/cruise-control/ref-cruise-control-configuration.adoc
```diff
@@ -95,7 +95,13 @@ Capacity limits can be set for the following broker resources:
...
+
+For Kafka clusters that use JBOD storage, do not set a disk capacity limit.
```
scholzj, line 100:
```suggestion
For Kafka clusters that use JBOD storage with multiple disks, do not set a disk capacity limit.
```

documentation/modules/cruise-control/ref-cruise-control-configuration.adoc
```diff
@@ -95,7 +95,13 @@ Capacity limits can be set for the following broker resources:
...
+Strimzi generates a specific disk capacity configuration for this type of cluster. 
+This is required because the path names of the Kafka JBOD volumes include a broker ID. 
+The configuration allows for intra-broker disk balancing.
+If you set a disk capacity limit, it will override the capacity configuration for the cluster with a standard configuration.
```
scholzj, line 104: Why does it sound as with with one disk, user has to configure something manually while with multiple disks we configure it? Seems strange and user-unfriendly.

documentation/modules/cruise-control/ref-cruise-control-configuration.adoc
kyguy: For Kafka clusters that do not use JBOD storage, Strimzi will generate a capacity configuration like this:
```
{
  "brokerCapacities":[
    {
      "brokerId": "-1",
      "capacity": {
        "DISK": "100000",
        "CPU": "100",
        "NW_IN": "10000",
        "NW_OUT": "10000"
      },
      "doc": "This is the default capacity. Capacity unit used for disk is in MB, cpu is 
              in percentage, network throughput is in KB."
    }
  ]
}
```
Where the default broker entry `-1` is applied to every broker in the cluster.

For Kafka clusters that use JBOD storage, Strimzi will generate a capacity configuration like this:
```
{
  "brokerCapacities":[
    {
      "brokerId": "0",
      "capacity": {
        "DISK": {
          "/var/lib/kafka/data-0/kafka-log0" : "1024.0"
          "/var/lib/kafka/data-1/kafka-log0" : "1024.0"
        },
        "CPU": "100",
        "NW_IN": "10000",
        "NW_OUT": "10000"
      },
      "doc": "Capacity for Broker 0"
    },
    {
      "brokerId": "1",
      "capacity": {
        "DISK": {
          "/var/lib/kafka/data-0/kafka-log1" : "1024.0"
          "/var/lib/kafka/data-1/kafka-log1" : "1024.0"
        },
        "CPU": "100",
        "NW_IN": "10000",
        "NW_OUT": "10000"
      },
      "doc": "Capacity for Broker 1"
    },
    {
      "brokerId": "1",
      "capacity": {
        "DISK": {
          "/var/lib/kafka/data-0/kafka-log2" : "1024.0"
          "/var/lib/kafka/data-1/kafka-log2" : "1024.0"
        },
        "CPU": "100",
        "NW_IN": "10000",
        "NW_OUT": "10000"
      },
      "doc": "Capacity for Broker 2"
    }
  ]
}

```
Even if it is a JBOD configuration that only has one disk:

```
{
  "brokerCapacities":[
    {
      "brokerId": "0",
      "capacity": {
        "DISK": {
          "/var/lib/kafka/data-0/kafka-log0" : "1024.0"
        },
        "CPU": "100",
        "NW_IN": "10000",
        "NW_OUT": "10000"
      },
      "doc": "Capacity for Broker 0"
    },
    {
      "brokerId": "1",
      "capacity": {
        "DISK": {
          "/var/lib/kafka/data-0/kafka-log1" : "1024.0"
        },
        "CPU": "100",
        "NW_IN": "10000",
        "NW_OUT": "10000"
      },
      "doc": "Capacity for Broker 1"
    },
    {
      "brokerId": "1",
      "capacity": {
        "DISK": {
          "/var/lib/kafka/data-0/kafka-log2" : "1024.0"
        },
        "CPU": "100",
        "NW_IN": "10000",
        "NW_OUT": "10000"
      },
      "doc": "Capacity for Broker 2"
    }
  ]
}

```

documentation/modules/cruise-control/ref-cruise-control-configuration.adoc
kyguy: Strimzi will auto configure the storage capacity config in all case whether it running a Kafka cluster deployment using non-JBOD, JBOD with multiple disks, JBOD with single disk. What I am trying to articulate is that we don't want users manually setting the disk capacity when using JBOD storage since it will override the special JBOD config generated by Strimzi with something like this:
```
{
  "brokerCapacities":[
    {
      "brokerId": "-1",
      "capacity": {
        "DISK": <Manually set disk capacity value>,
        "CPU": "100",
        "NW_IN": "10000",
        "NW_OUT": "10000"
      },
      "doc": "This is the default capacity. Capacity unit used for disk is in MB, cpu is 
              in percentage, network throughput is in KB."
    }
  ]
}
```
which will prevent intra-broker balancing

documentation/modules/cruise-control/ref-cruise-control-configuration.adoc
scholzj: How should people know that this is what _generic_ means?

documentation/modules/cruise-control/ref-cruise-control-configuration.adoc
kyguy: Makes me think now that maybe we should ditch this explanation and just prevent users from setting the disk capacity when using JBOD storage. It doesn't make sense to allow them to override it in the first place as it will only limit the features they can use

kyguy: It will be easier for me to fix the code to prevent this then explaining in the docs for users to avoid it :D

scholzj: Why would people configure the disk in the first place when we generate the configuration automatically? Why would I care when you use the _generic_ configuration and when you use the per-broker configuration? I mean, this is a part of Strimzi I do not understand much. And I do not think this really helps me to understand the difference and what it means for me.

scholzj: > Makes me think now that maybe we should ditch this explanation and just prevent users from setting the disk capacity when using JBOD storage. It doesn't make sense to allow them to override it in the first place as it will only limit the features they can use

I'm not sure. Why would the user want to configure the disk capacity with a single disk? And if they want to configure it for a single disk, why would they not want that for multiple disks?

scholzj: I mean, the text right now sounds to me like user of single disk has to configure more manually than a user with multiple disks. But you now make it sounds like user with a single disk has actually more options. But it is not clear why these options matter.

documentation/modules/cruise-control/ref-cruise-control-configuration.adoc
kyguy: I am not sure but I feel like I am getting users into the weeds with this, I am going to add some logic in the Capacity class to prevent users from misconfiguration. It will be much easier for us and the users to that then explain and have them understand!

documentation/modules/cruise-control/ref-cruise-control-configuration.adoc
kyguy: > Why would people configure the disk in the first place when we generate the configuration automatically?

You are right, I confused myself on this. Users shouldn't configure the disk because Strimzi automatically generates it. And even if they set the disk capacity the generated JBOD configuration will be unaffected/safe so all we can remove this whole paragraph!

kyguy:
As discussed in the comments above, since the disk capacity is generated automatically by Strimzi, we do not need users to ever set it. I have added a note explaining this and removed some of the docs surrounding setting it. I will deprecate `spec.cruiseControl.capacity.disk` in a future PR

scholzj (approved):
Nice work 👍

Frawless (approved):

kyguy:
@scholzj finally ready for ST barrage!

kyguy:
Ready for merge!

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/resource/cruisecontrol/RebalanceOptions.java
```diff
@@ -130,6 +139,11 @@ public RebalanceOptionsBuilder withSkipHardGoalCheck() {
             return this;
         }
 
+        public RebalanceOptionsBuilder withRebalanceDisk(boolean rebalanceDisk) {
```
tomncooper, line 142: Why does this need the boolean argument? The default is false, if someone calls rebalanceBuilder.withRebalanceDisk() we set the value to true. If not it stays false?

tomncooper (approved):
Good work @kyguy! This is a great feature to have!

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/cruisecontrol/Capacity.java
```diff
@@ -11,36 +11,100 @@
...
-    public static final int DEFAULT_BROKER_CPU_UTILIZATION_CAPACITY = 100;  // as a percentage (0-100)
-    public static final double DEFAULT_BROKER_INBOUND_NETWORK_KIB_PER_SECOND_CAPACITY = 10_000;  // in KiB/s
-    public static final double DEFAULT_BROKER_OUTBOUND_NETWORK_KIB_PER_SECOND_CAPACITY = 10_000;  // in KiB/s
+    // CC allows specifying a generic "default" broker entry in the capacity configuration to apply to all brokers without a specific broker entry.
+    // CC designates the id of this default broker entry as "-1".
+    // When using a non-JBOD disk configuration, we use a "default" broker entry to configure all brokers
+    private static final int DEFAULT_BROKER_ID = -1;
+    private static final String DEFAULT_BROKER_DOC = "This is the default capacity. Capacity unit used for disk is in MB, cpu is in percentage, network throughput is in KB.";
+
+    private static final double DEFAULT_BROKER_DISK_MIB_CAPACITY = 100_000;  // in MiB
```
mimaison, line 30: I find having the unit at the end of names slightly more readable, for example `DEFAULT_BROKER_DISK_CAPACITY_MIB`. WDYT?

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/cruisecontrol/Capacity.java
```diff
@@ -11,36 +11,100 @@
...
+     * Generate JBOD disk capacity configuration for a broker using the supplied storage configuration
      *
      * @param storage Storage configuration for Kafka cluster
-     * @return Disk capacity configuration value as a Double
+     * @param idx Index of the broker
+     * @return Disk capacity configuration value as a JsonObject for broker idx
+     */
+    private JsonObject generateJbodDiskCapacity(Storage storage, int idx) {
+        JsonObject json = new JsonObject();
+        String name = "";
```
mimaison, line 85: This could be defined directly in the loop below

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceAssemblyOperator.java
```diff
@@ -154,6 +155,7 @@
     private final SecretOperator secretOperations;
     private final PlatformFeaturesAvailability pfa;
     private final Optional<LabelSelector> kafkaSelector;
+    boolean usingJbodStorage = false;
```
mimaison, line 158: Just reviewing in Github, so it's hard to fully check but it looks like this field could be private

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/cruisecontrol/Capacity.java
kyguy: How about `DEFAULT_BROKER_DISK_CAPACITY_IN_MIB`?

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/cruisecontrol/Capacity.java
kyguy: Should I move `path` down there as well? I thought it was more efficient to define the variables before the loop? Or does the compiler optimize make this not matter?

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceAssemblyOperator.java
see-quick: Also, you do not need to explicitly set false here. Javac compiler sets it for you (default state)

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/cruisecontrol/Capacity.java
mimaison: I don't think we necessarily need `IN` but I think it's ok so I leave it up to you if you want to keep it.

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/cruisecontrol/Capacity.java
kyguy: I guess it the time saved (if any) is negligible , your suggestion makes it  more readable, I'll update it

systemtest/src/test/java/io/strimzi/systemtest/cruisecontrol/CruiseControlST.java
```diff
@@ -250,4 +253,43 @@ void testCruiseControlReplicaMovementStrategy(ExtensionContext extensionContext)
         ccConfFileContent = cmdKubeClient(namespaceName).execInPodContainer(ccPodName, Constants.CRUISE_CONTROL_CONTAINER_NAME, "cat", Constants.CRUISE_CONTROL_CONFIGURATION_FILE_PATH).out();
         assertThat(ccConfFileContent, containsString(newReplicaMovementStrategies));
     }
+
+    @ParallelNamespaceTest
+    void testCruiseControlIntraBrokerBalancing(ExtensionContext extensionContext) {
+        final String namespaceName = StUtils.getNamespaceBasedOnRbac(namespace, extensionContext);
+        String clusterName = mapWithClusterNames.get(extensionContext.getDisplayName());
```
see-quick, line 260: Could you use TestStorage class? You can look at `testCruiseControlBasicAPIRequests` test case.
```java
final TestStorage testStorage = new TestStorage(extensionContext);
```

kyguy: @see-quick just made the changes, let me now what you think!

see-quick (approved):

kyguy:
@scholzj ready for another barrage!

PR #5919, opened 2021-11-20
Update `default.replication.factor` and `min.insync.replicas` in system tests based on number of nodes

scholzj:
### Type of change

- Bugfix

### Description

The #5916 added the `default.replication.factor` and `min.insync.replicas` fields to the example files. The example files are being used by the system tests and now system tests are failing, because they do not adapt the numbers for these options but they change the number of replicas.

This PR changes the `default.replication.factor` and `min.insync.replicas` values in the same way as they are changed to the other configuration options included in the example YAMLs. That should help to fix the system tests.

### Checklist


scholzj:
This should be now ready for review.

sknot-rh (approved):

Frawless (approved):

PR #5923, opened 2021-11-22
Remove the `ps` dependency from system tests

scholzj:
### Type of change

- Enhancement / new feature

### Description

This PR removes the `ps` utility from system tests and the `procps` package from our base image.

### Checklist

### DIFF
--- a/docker-images/base/Dockerfile
+++ b/docker-images/base/Dockerfile
@@ -5,7 +5,7 @@ ARG TARGETPLATFORM
 USER root
 
 RUN microdnf update \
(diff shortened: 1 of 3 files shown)


michalxo:
This does not seem to work properly on rhel. For some strange reason, data in `/proc/<pid>/cmdline` converts space to some weird character and parsing is wrong. 

Is there any reason why can't we use jps/jcmd commands? I am investigating this route..

scholzj:
@michalxo There is no `jcmd` in upstream images. So your command does not work.

scholzj:
FYI: @michalxo confirmed this works fine for him.

Frawless (approved):

PR #5941, opened 2021-11-24
Admin client should use PEM format of secrets (TO)

sknot-rh:
### Type of change
- Enhancement

### Description
The Topic operator still uses the PKCS12 format of certificates. It should use PEM format instead. Also, TO does not currently use the `AdminClientProvider` API to create the AdminClient instance.

### Checklist
- [ ] Write tests
- [ ] Update documentation
- [ ] Update CHANGELOG.md


CHANGELOG.md
```diff
@@ -12,6 +12,7 @@
   When upgrading from Strimzi 0.22 or earlier, you have to disable the `ControlPlaneListener` feature gate when upgrading the cluster operator to make sure the Kafka cluster stays available during the upgrade.
   When downgrading to Strimzi 0.22 or earlier, you have to disable the `ControlPlaneListener` feature gate before downgrading the cluster operator to make sure the Kafka cluster stays available during the downgrade.
 * Update to Cruise Control version 2.5.74
+* Use PEM format for UO, CO and TO
```
scholzj, line 15: I think this is internal implementation detail So I would not mention it here. Plus it is not completely true -> it applies only to Kafka Admin client.

documentation/modules/deploying/proc-deploy-topic-operator-standalone.adoc
```diff
@@ -121,27 +121,21 @@ The default value is "PLAINTEXT".
...
-  - name: STRIMZI_TRUSTSTORE_PASSWORD <2>
-    value: "__TRUSTSTORE-PASSWORD__"
-  - name: STRIMZI_KEYSTORE_LOCATION <3>
-    value: "/path/to/keystore.p12"
-  - name: STRIMZI_KEYSTORE_PASSWORD <4>
-    value: "__KEYSTORE-PASSWORD__"
+  - name: STRIMZI_CLUSTER_CA_CERT_SECRET_NAME <1>
+    value: "publicKeysSecret"
+  - name: STRIMZI_EO_KEY_SECRET_NAME <2>
+    value: "privateKeySecret"
```
scholzj, line 134: This is in a way API change. Can we keep the old way? If not, this needs to be mentioned in the CHANGELOG in the breaking changes section. But if you are not using the Admin Client provider from CO and UO, maybe you can keep old options as well at least as deprecated for some time?

scholzj (commented):
Two general comments. The code needs to be reviewed by SMEs.

CHANGELOG.md
```diff
@@ -12,6 +12,7 @@
   When upgrading from Strimzi 0.22 or earlier, you have to disable the `ControlPlaneListener` feature gate when upgrading the cluster operator to make sure the Kafka cluster stays available during the upgrade.
   When downgrading to Strimzi 0.22 or earlier, you have to disable the `ControlPlaneListener` feature gate before downgrading the cluster operator to make sure the Kafka cluster stays available during the downgrade.
 * Update to Cruise Control version 2.5.74
+* Using of PKCS format is now deprecated in TO. PEM format should be used instead.
```
scholzj, line 15: Was it using PKCS12? Did it also support JKS?

documentation/modules/deploying/proc-deploy-topic-operator-standalone.adoc
```diff
@@ -121,12 +121,26 @@ The default value is "PLAINTEXT".
 You can set the value to `PLAINTEXT`, `SSL`, `SASL_PLAINTEXT`, or `SASL_SSL`.
 
 . If you want to connect to Kafka brokers that are using certificates from a public certificate authority, set `STRIMZI_PUBLIC_CA` to `true`. Set this property to `true`, for example, if you are using Amazon AWS MSK service.
-. If you enabled TLS with the `STRIMZI_TLS_ENABLED` environment variable, specify the keystore and truststore used to authenticate connection to the Kafka cluster.
+. If you enabled TLS with the `STRIMZI_TLS_ENABLED` environment variable, specify the keystore and truststore (deprecated) or specify names of the secrets containing public and private keys used to authenticate connection to the Kafka cluster.
```
scholzj, line 124: Why do we specify secrets? Why not mount the certs into env var?

scholzj: The problem is that specifying just secret name enforces the secret structure on the user. That sucks. Mounting the files directly into env vars is on the other hand much more flexible. Doesn't depend how man secrets you have, what are the field names etc.

sknot-rh: I was not sure that is safe.

scholzj: It would also make sure the TO does not need any RBAC rights to read the secrets => the code would AFAIK not work right now?

sknot-rh: I did some manual testing and it worked. AFICS these [RBAC](https://github.com/strimzi/strimzi-kafka-operator/blob/main/packaging/install/cluster-operator/031-ClusterRole-strimzi-entity-operator.yaml#L36) rules are used

scholzj: But that is just because the user operator which works with secrets. The standalone TO does not have this RBAC rights.

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/EntityTopicOperator.java
```diff
@@ -283,6 +291,11 @@ public static EntityTopicOperator fromCrd(Reconciliation reconciliation, Kafka k
         varList.add(buildEnvVar(ENV_VAR_SECURITY_PROTOCOL, EntityTopicOperatorSpec.DEFAULT_SECURITY_PROTOCOL));
         varList.add(buildEnvVar(ENV_VAR_TLS_ENABLED, Boolean.toString(true)));
         varList.add(buildEnvVar(ENV_VAR_STRIMZI_GC_LOG_ENABLED, String.valueOf(gcLoggingEnabled)));
+
+        varList.add(buildEnvVarFromSecret(ENV_VAR_EO_CERTIFICATE_CHAIN_DATA, EntityOperator.secretName(cluster), "entity-operator.crt"));
+        varList.add(buildEnvVarFromSecret(ENV_VAR_EO_KEY_SECRET_DATA, EntityOperator.secretName(cluster), "entity-operator.key"));
+        varList.add(buildEnvVar(ENV_VAR_CLUSTER_CA_CERT_SECRET_DATA, Util.certsToPemString(clusterCaCertSecret)));
```
scholzj, line 297: This should be mounted from secret.

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/EntityTopicOperator.java
```diff
@@ -59,6 +61,10 @@
 
     public static final String ENV_VAR_TLS_ENABLED = "STRIMZI_TLS_ENABLED";
 
+    public static final String ENV_VAR_CLUSTER_CA_CERT_SECRET_DATA = "STRIMZI_CLUSTER_CA_CERT_SECRET_DATA";
+    public static final String ENV_VAR_EO_KEY_SECRET_DATA = "STRIMZI_EO_KEY_DATA";
+    public static final String ENV_VAR_EO_CERTIFICATE_CHAIN_DATA = "STRIMZI_EO_CERTIFICATE_CHAIN_DATA";
+
```
scholzj, line 67: The `DATA` or `SECRET_DATA` sounds a bit weird in the name. Why don't you follow the same naming you used in the TO code? `STRIMZI_PRIVATE_KEY`, `STRIMZI_CERTIFICATE_CHAIN`, `STRIMZI_TRUSTED_CERTIFICATES`.

scholzj (commented):
Did it lost somewhere the backwards compatibility again?

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/EntityTopicOperator.java
sknot-rh: Yeah, I know. This is a workaround as this secret can contain multiple `*.crt` properties. The `buildEnvVarFromSecret` method allows us to load just one key, so we would need envar per key(?). I did not find any better way how to implement this :(

scholzj: Well, I think you need to make sure we have one file with all somewhere first then I guess. This is not just a TO problem. That seems to be a general problem for all clients.

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/EntityTopicOperator.java
tombentley: +1. Conceptually there is a very important distinction between a certificate that ended up in a trust store and the certificates in the chain. They're all certificates, but the former implies a trust relationship, whereas the latter do not. Ensuring the code that manipulates certs maintains this distinction is important. So using `trust` in variable and constant names is a good idea.

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/EntityTopicOperator.java
tombentley: What's preventing concatenating the PEM encoded certs?

documentation/modules/deploying/proc-deploy-topic-operator-standalone.adoc
sknot-rh: I think this needs to wait for https://github.com/strimzi/proposals/pull/46

scholzj: Then I guess we should maybe close it in the mean time? A long way before that is implemented.

Issue #3057, opened 2020-05-19
[Enhancement] Cruise Control: Provide better log message when incorrect annotation is used


tombentley:
When you provide a wrong value for the `strimzi.io/rebalance` annotation on a `KafkaRebalance` you get this message logged:

2020-05-19 09:50:27 WARN  KafkaRebalanceAssemblyOperator:954 - Wrong annotation value strimzi.io/rebalance=approved on myproject/my-rebalance

The annotation gets removed from the CR and the `Status` condition remains `ProposalReady`. 

It would be better if the logged warning message included the valid values.

scholzj:
As I said before ... this should be also raised in the conditions of the status. I do not think warning in CO logs is enough.

tombentley:
I agree.

ShubhamRwt:
Is this issue still valid @scholzj @tombentley ?

scholzj:
The CC experts should probably know if this was fixed or not.

ppatierno:
@ShubhamRwt have you tried if this still happens but I am not sure it was already fixed. It would be worth a try and then providing valid values for the annotation based on the state of the `KafkaRebalance` resource.

ShubhamRwt:
Not yet @ppatierno , but yes it is on my agenda for today. I will try to replicate it, and see if it still exists.

ShubhamRwt:
Looks like the issue still persists @scholzj @ppatierno

ShubhamRwt:
@scholzj @ppatierno I was facing an issue while I was trying get the warning message in the condition. Actually what happens is when annotate our KR resource, even if its wrong what happens is that the annotations are removed and made `none`. The issue which I am facing is that I was making a check on the basis of annotation but, after a reconciliation the annotation is removed and when the status is built again then the warning doesn't show up since the resource is now updated and `none` annonation is there

ShubhamRwt:
Should I do something like not changing the annotation to `none` if wrong annotation is passed

ppatierno:
Just improving the warning message as Tom suggested would be pretty simple; what Jakub is asking for, which makes sense, is improving the condition in the `KafkaRebalance` status.
[...]

PR #5948, opened 2021-11-25
Added warning condition to status when wrong annotation is provided to rebalance resource

ShubhamRwt:

_Select the type of your PR_

- Bugfix
- Enhancement / new feature

### Description

This PR resolves the issue mentioned in #3057. This PR improves the warning message shown in cluster operator and also adds warning message to the condition of the status when a wrong annotation is passed.

### Checklist

- [ ] Write tests
- [ ] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/api/src/main/java/io/strimzi/api/kafka/model/balancing/KafkaRebalanceState.java
+++ b/api/src/main/java/io/strimzi/api/kafka/model/balancing/KafkaRebalanceState.java
@@ -4,6 +4,8 @@
  */
 package io.strimzi.api.kafka.model.balancing;
 
+import java.util.List;
+
 public enum KafkaRebalanceState {
     /**
      * The resource has not been observed by the operator before.
@@ -14,7 +16,7 @@ public enum KafkaRebalanceState {
      *     <dt>NotReady</dt><dd>If the resource is invalid and a request could not be made.</dd>
      * </dl>
      */
-    New,
+    New(List.of()),
     /**
      * A proposal has been requested from Cruise Control, but is not ready yet.
      * Transitions to:
@@ -24,7 +26,7 @@ public enum KafkaRebalanceState {
      *     <dt>NotReady</dt><dd>If Cruise Control returned an error</dd>
      * </dl>
      */
-    PendingProposal,
+    PendingProposal(List.of(KafkaRebalanceAnnotation.stop)),
     /**
      * A proposal is ready and waiting for approval.
      * Transitions to:
@@ -34,7 +36,7 @@ public enum KafkaRebalanceState {
      *     <dt>ProposalReady</dt><dd>When the user sets annotation strimzi.io/rebalance=refresh and the proposal is already ready.</dd>
      * </dl>
      */
-    ProposalReady,
+    ProposalReady(List.of(KafkaRebalanceAnnotation.approve, KafkaRebalanceAnnotation.refresh)),
     /**
      * Cruise Control is doing the rebalance for an approved proposal.
      * Transitions to:
@@ -44,7 +46,7 @@ public enum KafkaRebalanceState {
      *     <dt>Ready</dt><dd>Once the rebalancing is complete.</dd>
      * </dl>
      */
-    Rebalancing,
+    Rebalancing(List.of(KafkaRebalanceAnnotation.stop)),
     /**
      * The user has stopped the actual rebalancing by setting annotation strimzi.io/rebalance=stop
      * May transition back to:
@@ -53,7 +55,7 @@ public enum KafkaRebalanceState {
      *     <dt>ProposalReady</dt><dd>If the user sets annotation strimzi.io/rebalance=refresh and the proposal is already ready.</dd>
      * </dl>
      */
-    Stopped,
+    Stopped(List.of(KafkaRebalanceAnnotation.refresh)),
     /**
      * There's been some error.
      * Transitions to:
@@ -61,15 +63,29 @@ public enum KafkaRebalanceState {
      *     <dt>New</dt><dd>If the error was caused by the resource itself that was fixed by the user.</dd>
      * </dl>
      */
-    NotReady,
+    NotReady(List.of(KafkaRebalanceAnnotation.refresh)),
     /**
      * The rebalance is complete and there is no transition from this state.
      * The resource is eligible for garbage collection after a configurable delay.
      * There is no transition from this state to a new one.
      */
-    Ready,
+    Ready(List.of(KafkaRebalanceAnnotation.refresh)),
     /**
      * The user paused reconciliations by setting annotation strimzi.io/pause-reconciliation="true".
      */
-    ReconciliationPaused
+    ReconciliationPaused(List.of());
+
+    private final List<KafkaRebalanceAnnotation> validAnnotations;
+
+    KafkaRebalanceState(List<KafkaRebalanceAnnotation> validAnnotations) {
+        this.validAnnotations = validAnnotations;
+    }
+
+    public boolean isValidateAnnotation(KafkaRebalanceAnnotation annotation) {
+        return validAnnotations.contains(annotation);
+    }
+
+    public List<KafkaRebalanceAnnotation> getValidAnnotations() {
+        return validAnnotations;
+    }
 }
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceAssemblyOperator.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceAssemblyOperator.java
@@ -61,6 +61,7 @@ import io.vertx.core.json.JsonArray;
 import io.vertx.core.json.JsonObject;
 
 import java.util.ArrayList;
+import java.util.HashSet;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.HashMap;
@@ -400,18 +401,21 @@ public class KafkaRebalanceAssemblyOperator
                                                         ANNO_STRIMZI_IO_REBALANCE,
                                                         rawRebalanceAnnotation(updatedKafkaRebalance));
                                                 if (hasRebalanceAnnotation(updatedKafkaRebalance)) {
-                                                    LOGGER.debugCr(reconciliation, "Removing annotation {}={}",
-                                                            ANNO_STRIMZI_IO_REBALANCE,
-                                                            rawRebalanceAnnotation(updatedKafkaRebalance));
-                                                    // Updated KafkaRebalance has rebalance annotation removed as
-                                                    // action specified by user has been completed.
-                                                    KafkaRebalance patchedKafkaRebalance = new KafkaRebalanceBuilder(updatedKafkaRebalance)
-                                                            .editMetadata()
-                                                                .removeFromAnnotations(ANNO_STRIMZI_IO_REBALANCE)
-                                                            .endMetadata()
-                                                            .build();
-
-                                                    return kafkaRebalanceOperator.patchAsync(reconciliation, patchedKafkaRebalance);
+                                                    if (currentState != KafkaRebalanceState.New && rebalanceAnnotation != KafkaRebalanceAnnotation.none && !currentState.isValidateAnnotation(rebalanceAnnotation)) {
+                                                        return Future.succeededFuture();
+                                                    } else {
+                                                        LOGGER.infoCr(reconciliation, "Removing annotation {}={}",
+                                                                ANNO_STRIMZI_IO_REBALANCE,
+                                                                rawRebalanceAnnotation(updatedKafkaRebalance));
+                                                        // Updated KafkaRebalance has rebalance annotation removed as
+                                                        // action specified by user has been completed.
+                                                        KafkaRebalance patchedKafkaRebalance = new KafkaRebalanceBuilder(updatedKafkaRebalance)
+                                                                .editMetadata()
+                                                                    .removeFromAnnotations(ANNO_STRIMZI_IO_REBALANCE)
+                                                                .endMetadata()
+                                                                .build();
+                                                        return kafkaRebalanceOperator.patchAsync(reconciliation, patchedKafkaRebalance);
+                                                    }
                                                 } else {
                                                     LOGGER.debugCr(reconciliation, "No annotation {}", ANNO_STRIMZI_IO_REBALANCE);
                                                     return Future.succeededFuture();
@@ -470,7 +474,6 @@ public class KafkaRebalanceAssemblyOperator
         conditions.addAll(validation);
         Condition currentState = rebalanceStateCondition(currentStatus);
         conditions.add(currentState);
-
         return new KafkaRebalanceStatusBuilder()
                 .withSessionId(currentStatus.getSessionId())
                 .withOptimizationResult(currentStatus.getOptimizationResult())
@@ -731,7 +734,9 @@ public class KafkaRebalanceAssemblyOperator
             return onNew(reconciliation, host, apiClient, kafkaRebalance, rebalanceOptionsBuilder);
         } else {
             // Stay in the current NotReady state, returning null as next state
-            return Future.succeededFuture(new MapAndStatus<>(null, null));
+            Set<Condition> conditions = validate(reconciliation, kafkaRebalance);
+            validateAnnotation(conditions, KafkaRebalanceState.NotReady, rebalanceAnnotation);
+            return Future.succeededFuture(buildRebalanceStatus(null, KafkaRebalanceState.NotReady, conditions));
         }
     }
 
@@ -795,12 +800,16 @@ public class KafkaRebalanceAssemblyOperator
                                     vertx.cancelTimer(t);
                                     p.complete(buildRebalanceStatus(null, KafkaRebalanceState.Stopped, validate(reconciliation, currentKafkaRebalance)));
                                 } else {
-
                                     requestRebalance(reconciliation, host, apiClient, kafkaRebalance, true, rebalanceOptionsBuilder,
                                             currentKafkaRebalance.getStatus().getSessionId())
                                         .onSuccess(rebalanceMapAndStatus -> {
                                             // If the returned status has an optimization result then the rebalance proposal
                                             // is ready, so stop the polling
+                                            KafkaRebalanceStatus status = rebalanceMapAndStatus.getStatus();
+                                            Set<Condition> conditions = new HashSet<>();
+                                            validateAnnotation(conditions, KafkaRebalanceState.PendingProposal, rebalanceAnnotation(reconciliation, currentKafkaRebalance));
+                                            status.addConditions(conditions);
+                                            rebalanceMapAndStatus.setStatus(status);
                                             if (rebalanceMapAndStatus.getStatus().getOptimizationResult() != null &&
                                                     !rebalanceMapAndStatus.getStatus().getOptimizationResult().isEmpty()) {
                                                 vertx.cancelTimer(t);
@@ -872,7 +881,10 @@ public class KafkaRebalanceAssemblyOperator
                 return requestRebalance(reconciliation, host, apiClient, kafkaRebalance, true, rebalanceOptionsBuilder);
             default:
                 LOGGER.warnCr(reconciliation, "Ignore annotation {}={}", ANNO_STRIMZI_IO_REBALANCE, rebalanceAnnotation);
-                return configMapOperator.getAsync(kafkaRebalance.getMetadata().getNamespace(), kafkaRebalance.getMetadata().getName()).compose(loadmap -> Future.succeededFuture(new MapAndStatus<>(loadmap, buildRebalanceStatusFromPreviousStatus(kafkaRebalance.getStatus(), validate(reconciliation, kafkaRebalance)))));
+                Set<Condition> conditions = validate(reconciliation, kafkaRebalance);
+                validateAnnotation(conditions, KafkaRebalanceState.ProposalReady, rebalanceAnnotation);
+                return configMapOperator.getAsync(kafkaRebalance.getMetadata().getNamespace(), kafkaRebalance.getMetadata().getName())
+                        .compose(loadmap -> Future.succeededFuture(new MapAndStatus<>(loadmap, buildRebalanceStatusFromPreviousStatus(kafkaRebalance.getStatus(), conditions))));
         }
     }
 
@@ -926,6 +938,8 @@ public class KafkaRebalanceAssemblyOperator
                                         });
                                 } else {
                                     LOGGER.infoCr(reconciliation, "Getting Cruise Control rebalance user task status");
+                                    Set<Condition> conditions = validate(reconciliation, kafkaRebalance);
+                                    validateAnnotation(conditions, KafkaRebalanceState.Rebalancing, rebalanceAnnotation(reconciliation, currentKafkaRebalance));
                                     apiClient.getUserTaskStatus(host, CruiseControl.REST_API_PORT, sessionId)
                                         .onSuccess(cruiseControlResponse -> {
                                             JsonObject taskStatusJson = cruiseControlResponse.getJson();
@@ -935,7 +949,7 @@ public class KafkaRebalanceAssemblyOperator
                                                     vertx.cancelTimer(t);
                                                     LOGGER.infoCr(reconciliation, "Rebalance ({}) is now complete", sessionId);
                                                     p.complete(buildRebalanceStatus(
-                                                            kafkaRebalance, null, KafkaRebalanceState.Ready, taskStatusJson, validate(reconciliation, kafkaRebalance)));
+                                                            kafkaRebalance, null, KafkaRebalanceState.Ready, taskStatusJson, conditions));
                                                     break;
                                                 case COMPLETED_WITH_ERROR:
                                                     // TODO: There doesn't seem to be a way to retrieve the actual error message from the user tasks endpoint?
@@ -944,7 +958,7 @@ public class KafkaRebalanceAssemblyOperator
                                                     //       details of any issues while rebalancing.
                                                     LOGGER.errorCr(reconciliation, "Rebalance ({}) optimization proposal has failed to complete", sessionId);
                                                     vertx.cancelTimer(t);
-                                                    p.complete(buildRebalanceStatus(sessionId, KafkaRebalanceState.NotReady, validate(reconciliation, kafkaRebalance)));
+                                                    p.complete(buildRebalanceStatus(sessionId, KafkaRebalanceState.NotReady, conditions));
                                                     break;
                                                 case IN_EXECUTION: // Rebalance is still in progress
                                                     // We need to check that the status has been updated with the ongoing optimisation proposal
@@ -957,7 +971,7 @@ public class KafkaRebalanceAssemblyOperator
                                                         // Cancel the timer so that the status is returned and updated.
                                                         vertx.cancelTimer(t);
                                                         p.complete(buildRebalanceStatus(
-                                                                kafkaRebalance, sessionId, KafkaRebalanceState.Rebalancing, taskStatusJson, validate(reconciliation, kafkaRebalance)));
+                                                                kafkaRebalance, sessionId, KafkaRebalanceState.Rebalancing, taskStatusJson, conditions));
                                                     }
                                                     ccApiErrorCount.set(0);
                                                     // TODO: Find out if there is any way to check the progress of a rebalance.
@@ -1027,8 +1041,9 @@ public class KafkaRebalanceAssemblyOperator
         if (rebalanceAnnotation == KafkaRebalanceAnnotation.refresh) {
             return requestRebalance(reconciliation, host, apiClient, kafkaRebalance, true, rebalanceOptionsBuilder);
         } else {
-            LOGGER.warnCr(reconciliation, "Ignore annotation {}={}", ANNO_STRIMZI_IO_REBALANCE, rebalanceAnnotation);
-            return Future.succeededFuture(buildRebalanceStatus(null, KafkaRebalanceState.Stopped, validate(reconciliation, kafkaRebalance)));
+            Set<Condition> conditions = validate(reconciliation, kafkaRebalance);
+            validateAnnotation(conditions, KafkaRebalanceState.Stopped, rebalanceAnnotation);
+            return Future.succeededFuture(buildRebalanceStatus(null, KafkaRebalanceState.Stopped, conditions));
         }
     }
 
@@ -1053,8 +1068,10 @@ public class KafkaRebalanceAssemblyOperator
         if (rebalanceAnnotation == KafkaRebalanceAnnotation.refresh) {
             return requestRebalance(reconciliation, host, apiClient, kafkaRebalance, true, rebalanceOptionsBuilder);
         } else {
-            LOGGER.warnCr(reconciliation, "{}: Ignore annotation {}={}", ANNO_STRIMZI_IO_REBALANCE, rebalanceAnnotation);
-            return configMapOperator.getAsync(kafkaRebalance.getMetadata().getNamespace(), kafkaRebalance.getMetadata().getName()).compose(loadmap -> Future.succeededFuture(new MapAndStatus<>(loadmap, buildRebalanceStatusFromPreviousStatus(kafkaRebalance.getStatus(), validate(reconciliation, kafkaRebalance)))));
+            Set<Condition> conditions = validate(reconciliation, kafkaRebalance);
+            validateAnnotation(conditions, KafkaRebalanceState.Ready, rebalanceAnnotation);
+            return configMapOperator.getAsync(kafkaRebalance.getMetadata().getNamespace(), kafkaRebalance.getMetadata().getName())
+                    .compose(loadmap -> Future.succeededFuture(new MapAndStatus<>(loadmap, buildRebalanceStatusFromPreviousStatus(kafkaRebalance.getStatus(), conditions))));
         }
     }
 
@@ -1214,12 +1231,18 @@ public class KafkaRebalanceAssemblyOperator
     private KafkaRebalanceAnnotation rebalanceAnnotation(Reconciliation reconciliation, KafkaRebalance kafkaRebalance) {
         String rebalanceAnnotationValue = rawRebalanceAnnotation(kafkaRebalance);
         KafkaRebalanceAnnotation rebalanceAnnotation;
+
         try {
-            rebalanceAnnotation = rebalanceAnnotationValue == null ?
-                    KafkaRebalanceAnnotation.none : KafkaRebalanceAnnotation.valueOf(rebalanceAnnotationValue);
+            if (rebalanceAnnotationValue == null) {
+                rebalanceAnnotation = KafkaRebalanceAnnotation.none;
+            } else if (rebalanceAnnotationValue.equals("none")) {
+                rebalanceAnnotation = KafkaRebalanceAnnotation.unknown;
+            } else {
+                rebalanceAnnotation = KafkaRebalanceAnnotation.valueOf(rebalanceAnnotationValue);
+            }
         } catch (IllegalArgumentException e) {
             rebalanceAnnotation = KafkaRebalanceAnnotation.unknown;
-            LOGGER.warnCr(reconciliation, "Wrong annotation value {}={} on {}/{}",
+            LOGGER.warnCr(reconciliation, "Wrong annotation value {}={} on {}/{}. The rebalance resource supports the following annotations: approve, refresh, stop.",
                     ANNO_STRIMZI_IO_REBALANCE, rebalanceAnnotationValue,
                     kafkaRebalance.getMetadata().getNamespace(), kafkaRebalance.getMetadata().getName());
         }
@@ -1237,6 +1260,22 @@ public class KafkaRebalanceAssemblyOperator
         return hasRebalanceAnnotation(kafkaRebalance) ?
                 kafkaRebalance.getMetadata().getAnnotations().get(ANNO_STRIMZI_IO_REBALANCE) : null;
 
+    }
+
+    /**
+     * Return warning condition if the rebalance annotation wrong corresponding to the state
+
+     * @param state Contains the KafkaRebalanceState for which we are checking the valid annotation
+     * @param rebalanceAnnotation Contains the rebalance annotation
+     * @return the value for the strimzio.io/rebalance annotation on the provided KafkaRebalance resource instance
+     */
+    private void validateAnnotation(Set<Condition> conditions, KafkaRebalanceState state, KafkaRebalanceAnnotation rebalanceAnnotation) {
+        if (rebalanceAnnotation != KafkaRebalanceAnnotation.none) {
+            if (!state.isValidateAnnotation(rebalanceAnnotation)) {
+                conditions.add(StatusUtils.buildWarningCondition("InvalidAnnotation", "Wrong annotation value: " + rebalanceAnnotation
+                        + ". The rebalance resource supports the following annotations in " + state.toString() + " state: " + state.getValidAnnotations()));
+            }
+        }
     }
 
     /**
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceAssemblyOperator.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceAssemblyOperator.java
@@ -401,7 +401,7 @@ public class KafkaRebalanceAssemblyOperator
                                                         ANNO_STRIMZI_IO_REBALANCE,
                                                         rawRebalanceAnnotation(updatedKafkaRebalance));
                                                 if (hasRebalanceAnnotation(updatedKafkaRebalance)) {
-                                                    if (currentState != KafkaRebalanceState.New && rebalanceAnnotation != KafkaRebalanceAnnotation.none && !currentState.isValidateAnnotation(rebalanceAnnotation)) {
+                                                    if (currentState != KafkaRebalanceState.ReconciliationPaused && rebalanceAnnotation != KafkaRebalanceAnnotation.none && !currentState.isValidateAnnotation(rebalanceAnnotation)) {
                                                         return Future.succeededFuture();
                                                     } else {
                                                         LOGGER.infoCr(reconciliation, "Removing annotation {}={}",
--- a/cluster-operator/src/test/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceAssemblyOperatorTest.java
+++ b/cluster-operator/src/test/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceAssemblyOperatorTest.java
@@ -587,7 +587,7 @@ public class KafkaRebalanceAssemblyOperatorTest {
                             refreshKr);
                 })
                 .onComplete(context.succeeding(v -> {
-                    // the resource moved from Ready to PropsalPending
+                    // the resource moved from Ready to ProposalReady
                     assertState(context, kubernetesClient, CLUSTER_NAMESPACE, RESOURCE_NAME, KafkaRebalanceState.ProposalReady);
                     checkpoint.flag();
                 }));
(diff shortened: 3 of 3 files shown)


ShubhamRwt:
Status:
  Conditions:
    Last Transition Time:  2021-11-25T20:38:11.643022Z
    Status:                True
    Type:                  ProposalReady
    Last Transition Time:  2021-11-25T20:38:11.554458Z
    Message:               The rebalance resource supports the following annotations : none, approve, refresh, stop.
    Reason:                Unsupported/Unknown annotation value
    Status:                True
    Type:                  Warning
  Observed Generation:     1


@ppatierno @tomncooper @scholzj @kyguy Is it fine to have both the condition or should I get only one condition although that would me to create a new method in StatusUtils class?

packaging/install/cluster-operator/060-Deployment-strimzi-cluster-operator.yaml
```diff
@@ -115,4 +115,4 @@ spec:
               cpu: 200m
               memory: 384Mi
   strategy:
-    type: Recreate
+    type: Recreate
```
scholzj, line 118: Why did you changed this?

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceAssemblyOperator.java
```diff
@@ -837,7 +840,9 @@ public MapAndStatus(T loadMap, K status) {
                 return requestRebalance(reconciliation, host, apiClient, kafkaRebalance, true, rebalanceOptionsBuilder);
             default:
                 LOGGER.warnCr(reconciliation, "Ignore annotation {}={}", ANNO_STRIMZI_IO_REBALANCE, rebalanceAnnotation);
-                return Future.succeededFuture(new MapAndStatus<>(null, buildRebalanceStatusFromPreviousStatus(kafkaRebalance.getStatus(), validate(reconciliation, kafkaRebalance))));
+                Set<Condition> conditions = validate(reconciliation, kafkaRebalance);
+                conditions.add(StatusUtils.buildWarningCondition("Unsupported/Unknown annotation value", "The rebalance resource supports the following annotations : none, approve, refresh, stop."));
```
scholzj, line 844: What does the `none` mean? Also, I don't think you should have space before `:`.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceAssemblyOperator.java
```diff
@@ -1180,7 +1185,7 @@ private KafkaRebalanceAnnotation rebalanceAnnotation(Reconciliation reconciliati
                     KafkaRebalanceAnnotation.none : KafkaRebalanceAnnotation.valueOf(rebalanceAnnotationValue);
         } catch (IllegalArgumentException e) {
             rebalanceAnnotation = KafkaRebalanceAnnotation.unknown;
-            LOGGER.warnCr(reconciliation, "Wrong annotation value {}={} on {}/{}",
+            LOGGER.warnCr(reconciliation, "Wrong annotation value {}={} on {}/{}. The rebalance resource supports the following annotations : none, approve, refresh, stop.",
```
scholzj, line 1188: Same as above: What does the `none` mean? Also, I don't think you should have space before `:`.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceAssemblyOperator.java
```diff
@@ -837,7 +840,9 @@ public MapAndStatus(T loadMap, K status) {
                 return requestRebalance(reconciliation, host, apiClient, kafkaRebalance, true, rebalanceOptionsBuilder);
             default:
                 LOGGER.warnCr(reconciliation, "Ignore annotation {}={}", ANNO_STRIMZI_IO_REBALANCE, rebalanceAnnotation);
-                return Future.succeededFuture(new MapAndStatus<>(null, buildRebalanceStatusFromPreviousStatus(kafkaRebalance.getStatus(), validate(reconciliation, kafkaRebalance))));
+                Set<Condition> conditions = validate(reconciliation, kafkaRebalance);
```
scholzj, line 843: What does the `validate` method do? Shouldn't it validate the annotation and set the condition?

cluster-operator/src/test/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceAssemblyOperatorTest.java
```diff
@@ -1224,6 +1273,15 @@ private void assertValidationCondition(VertxTestContext context, KubernetesClien
         });
     }
 
+    private void assertAnnotationValidationCondition(VertxTestContext context, KubernetesClient kubernetesClient, String namespace, String resource, String reason) {
+        context.verify(() -> {
+            KafkaRebalance kafkaRebalance = Crds.kafkaRebalanceOperation(kubernetesClient).inNamespace(namespace).withName(resource).get();
+            assertThat(kafkaRebalance, StateMatchers.hasState());
+            kafkaRebalance.getStatus().getConditions();
```
scholzj, line 1280: What does this line do?

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceAssemblyOperator.java
ppatierno: As mentioned offline, `none` is not a possible value, it's just an enum value which represents no value on the annotation that was needed for the state machine. It should not be listed here.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceAssemblyOperator.java
ShubhamRwt: The `validate` method is used to check for deprecated and unknown properties in the CR

packaging/install/cluster-operator/060-Deployment-strimzi-cluster-operator.yaml
ShubhamRwt: Ahh it might have changed when I was setting my docker images. Sorry

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceAssemblyOperator.java
```diff
@@ -456,6 +459,7 @@ private KafkaRebalanceStatus buildRebalanceStatusFromPreviousStatus(KafkaRebalan
         conditions.addAll(validation);
         Condition currentState = rebalanceStateCondition(currentStatus);
         conditions.add(currentState);
+        System.out.println(conditions);
```
tomncooper, line 462: Remember to remove these in the final PR.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceAssemblyOperator.java
```diff
@@ -837,7 +841,9 @@ public MapAndStatus(T loadMap, K status) {
                 return requestRebalance(reconciliation, host, apiClient, kafkaRebalance, true, rebalanceOptionsBuilder);
             default:
                 LOGGER.warnCr(reconciliation, "Ignore annotation {}={}", ANNO_STRIMZI_IO_REBALANCE, rebalanceAnnotation);
-                return Future.succeededFuture(new MapAndStatus<>(null, buildRebalanceStatusFromPreviousStatus(kafkaRebalance.getStatus(), validate(reconciliation, kafkaRebalance))));
+                Set<Condition> conditions = validate(reconciliation, kafkaRebalance);
+                conditions.add(StatusUtils.buildWarningCondition("Unsupported/Unknown annotation value", "The rebalance resource supports the following annotations : approve, refresh, stop."));
```
tomncooper, line 845: Is it possible to add the actual annotation text the user has attached to the warning message?

ShubhamRwt: I am not sure since the annotation is converted to unknown by this step. But I can try that for sure

sknot-rh: I think Tom meant something like 
`conditions.add(StatusUtils.buildWarningCondition("Unsupported/Unknown annotation value" + rebalanceAnnotation, "The rebalance resource supports the following annotations : approve, refresh, stop."));`

ShubhamRwt: @sknot-rh Actually the  annotation is changed to `unknown`. For eg If we pass `Dave` as an annotation then it will get converted to unknown and then this if condition gets executed.

ShubhamRwt: I think maybe Tom meant that we should have that wrong annotation in the message which is `Dave` in the above eg

tomncooper: You could add a field to the Rebalance Annotation enum to store the supplied annotation text? Just an idea.

kyguy (approved):
Looks great @ShubhamRwt!

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceAssemblyOperator.java
```diff
@@ -1180,7 +1185,7 @@ private KafkaRebalanceAnnotation rebalanceAnnotation(Reconciliation reconciliati
                     KafkaRebalanceAnnotation.none : KafkaRebalanceAnnotation.valueOf(rebalanceAnnotationValue);
         } catch (IllegalArgumentException e) {
             rebalanceAnnotation = KafkaRebalanceAnnotation.unknown;
-            LOGGER.warnCr(reconciliation, "Wrong annotation value {}={} on {}/{}",
+            LOGGER.warnCr(reconciliation, "Wrong annotation value {}={} on {}/{}. The rebalance resource supports the following annotations : approve, refresh, stop.",
```
scholzj, line 1188: Don't use space before `:`:

```suggestion
            LOGGER.warnCr(reconciliation, "Wrong annotation value {}={} on {}/{}. The rebalance resource supports the following annotations: approve, refresh, stop.",
```

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceAssemblyOperator.java
```diff
@@ -837,7 +840,9 @@ public MapAndStatus(T loadMap, K status) {
                 return requestRebalance(reconciliation, host, apiClient, kafkaRebalance, true, rebalanceOptionsBuilder);
             default:
                 LOGGER.warnCr(reconciliation, "Ignore annotation {}={}", ANNO_STRIMZI_IO_REBALANCE, rebalanceAnnotation);
-                return Future.succeededFuture(new MapAndStatus<>(null, buildRebalanceStatusFromPreviousStatus(kafkaRebalance.getStatus(), validate(reconciliation, kafkaRebalance))));
+                Set<Condition> conditions = validate(reconciliation, kafkaRebalance);
+                conditions.add(StatusUtils.buildWarningCondition("InvalidAnnotation", "Unsupported/Unknown annotation value: " + rawRebalanceAnnotation(kafkaRebalance) + ". The rebalance resource supports the following annotations : approve, refresh, stop."));
```
scholzj, line 844: Don't use space before `:`:

```suggestion
                conditions.add(StatusUtils.buildWarningCondition("InvalidAnnotation", "Unsupported/Unknown annotation value: " + rawRebalanceAnnotation(kafkaRebalance) + ". The rebalance resource supports the following annotations: approve, refresh, stop."));
```

scholzj: It looks like we set this condition only when the proposal is Ready. Is that intentional? When happens when a wrong annotation is used when the proposal is in a different state?

ShubhamRwt: @scholzj The annotations that cruise control support are generally applied in `ProposalReady` state mainly. As per the code I think if we are on `Ready` or maybe `Stop` state, we can only use `refresh` annotation and other annotations are ignored in that case basically. So I guess in that case, even other two correct annotations are wrong so I can add a warning regarding that ?

scholzj: I think you need to consult the CC experts what is the best approach. This is just something what seemed strange to me.

ppatierno: @ShubhamRwt IIRC we agreed offline to set a warning condition in any case a not supported annotation was used on the `KafkaRebalance` any state it is. So for example, in `Stop` state we are accepting `refresh` only but we are ignoring the others logging the warning. I would say that we should add a warning condition in that case as well. The same for `Ready`state.
The rule should be: in any state we are if an annotation is not supported, we should add a warning condition and not just logging the warning as we are doing today.

ShubhamRwt: @ppatierno I have added the warning in other states in the latest commit

ppatierno: Ok so I guess the @scholzj comment was stale and you made the change, I see it but ...
What I see in other states (i.e. onNotReady) is that a local `conditions` is created but never used in the sense that it's not reflected in a built status for the KafkaRebalance resource.

```java
} else {
            // Stay in the current NotReady state, returning null as next state
            Set<Condition> conditions = validate(reconciliation, kafkaRebalance);
            conditions.add(StatusUtils.buildWarningCondition("InvalidAnnotation", "Wrong annotation value: " + rawRebalanceAnnotation(kafkaRebalance) + ". The rebalance resource supports the following annotations in `NotReady` state: refresh"));
            return Future.succeededFuture(new MapAndStatus<>(null, null));
        }
```

I guess you would need to build the status as well using the warning conditions.

ShubhamRwt: @ppatierno @scholzj I have added the warning test now with some change in how I was writing the test. They seem to work fine now and applied on all states

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceAssemblyOperator.java
```diff
@@ -891,6 +897,8 @@ public MapAndStatus(T loadMap, K status) {
                                         });
                                 } else {
                                     LOGGER.infoCr(reconciliation, "Getting Cruise Control rebalance user task status");
+                                    Set<Condition> conditions = validate(reconciliation, kafkaRebalance);
+                                    conditions.add(StatusUtils.buildWarningCondition("InvalidAnnotation", "Wrong annotation value: " + rebalanceAnnotation(reconciliation, currentKafkaRebalance) + ". The rebalance resource supports the following annotations in 'Rebalancing' state: stop."));
```
scholzj, line 901: How do you know there is some rebalance annotation set? The If-else logic before shows that the annotation is not `KafkaRebalanceAnnotation.stop` ... but how do you know it is not null?

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceAssemblyOperator.java
```diff
@@ -1019,7 +1029,9 @@ public MapAndStatus(T loadMap, K status) {
             return requestRebalance(reconciliation, host, apiClient, kafkaRebalance, true, rebalanceOptionsBuilder);
         } else {
             LOGGER.warnCr(reconciliation, "{}: Ignore annotation {}={}", ANNO_STRIMZI_IO_REBALANCE, rebalanceAnnotation);
-            return configMapOperator.getAsync(kafkaRebalance.getMetadata().getNamespace(), kafkaRebalance.getMetadata().getName()).compose(loadmap -> Future.succeededFuture(new MapAndStatus<>(loadmap, buildRebalanceStatusFromPreviousStatus(kafkaRebalance.getStatus(), validate(reconciliation, kafkaRebalance)))));
+            Set<Condition> conditions = validate(reconciliation, kafkaRebalance);
+            conditions.add(StatusUtils.buildWarningCondition("InvalidAnnotation", "Wrong annotation value: " + rebalanceAnnotation + ". The rebalance resource supports the following annotations in `Ready` state: refresh"));
```
scholzj, line 1033: Same as above, how do you know any annotation is set?

ShubhamRwt: Resolved this issue in new changes

scholzj:
@ShubhamRwt It looks like you need to fix the Spotbugs errors:

```
[ERROR] Switch statement found in io.strimzi.operator.cluster.operator.assembly.KafkaRebalanceAssemblyOperator.rebalanceAnnotation(Reconciliation, KafkaRebalance) where default case is missing [io.strimzi.operator.cluster.operator.assembly.KafkaRebalanceAssemblyOperator, io.strimzi.operator.cluster.operator.assembly.KafkaRebalanceAssemblyOperator, io.strimzi.operator.cluster.operator.assembly.KafkaRebalanceAssemblyOperator, io.strimzi.operator.cluster.operator.assembly.KafkaRebalanceAssemblyOperator] At KafkaRebalanceAssemblyOperator.java:[lines 1219-1221]Another occurrence at KafkaRebalanceAssemblyOperator.java:[lines 1227-1231]Another occurrence at KafkaRebalanceAssemblyOperator.java:[lines 1236-1238]Another occurrence at KafkaRebalanceAssemblyOperator.java:[lines 1243-1246] SF_SWITCH_NO_DEFAULT
```

ShubhamRwt:
> @ShubhamRwt It looks like you need to fix the Spotbugs errors:
> 
> ```
> [ERROR] Switch statement found in io.strimzi.operator.cluster.operator.assembly.KafkaRebalanceAssemblyOperator.rebalanceAnnotation(Reconciliation, KafkaRebalance) where default case is missing [io.strimzi.operator.cluster.operator.assembly.KafkaRebalanceAssemblyOperator, io.strimzi.operator.cluster.operator.assembly.KafkaRebalanceAssemblyOperator, io.strimzi.operator.cluster.operator.assembly.KafkaRebalanceAssemblyOperator, io.strimzi.operator.cluster.operator.assembly.KafkaRebalanceAssemblyOperator] At KafkaRebalanceAssemblyOperator.java:[lines 1219-1221]Another occurrence at KafkaRebalanceAssemblyOperator.java:[lines 1227-1231]Another occurrence at KafkaRebalanceAssemblyOperator.java:[lines 1236-1238]Another occurrence at KafkaRebalanceAssemblyOperator.java:[lines 1243-1246] SF_SWITCH_NO_DEFAULT
> ```

Resolved @scholzj

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceAssemblyOperator.java
```diff
@@ -1174,13 +1208,52 @@ public MapAndStatus(T loadMap, K status) {
      */
     private KafkaRebalanceAnnotation rebalanceAnnotation(Reconciliation reconciliation, KafkaRebalance kafkaRebalance) {
         String rebalanceAnnotationValue = rawRebalanceAnnotation(kafkaRebalance);
+        KafkaRebalanceState state = state(kafkaRebalance);
         KafkaRebalanceAnnotation rebalanceAnnotation;
         try {
             rebalanceAnnotation = rebalanceAnnotationValue == null ?
                     KafkaRebalanceAnnotation.none : KafkaRebalanceAnnotation.valueOf(rebalanceAnnotationValue);
+            if (state != null) {
```
ppatierno, line 1216: maybe this code could be simplified a lot. We can do the other way around ... switch on the annotation because we are interested in just three values `approve`, `stop` and `refresh`, the default would be just returning an  `unknown`.
On each annotation case, you can just filter the right (or wrong) state with a single if and returning `unknown` in case, otherwise the current annotation.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceAssemblyOperator.java
```diff
@@ -388,18 +388,21 @@ private String rebalanceStateConditionType(KafkaRebalanceStatus status) {
...
                                                 if (hasRebalanceAnnotation(updatedKafkaRebalance)) {
-                                                    LOGGER.debugCr(reconciliation, "Removing annotation {}={}",
-                                                            ANNO_STRIMZI_IO_REBALANCE,
-                                                            rawRebalanceAnnotation(updatedKafkaRebalance));
-                                                    // Updated KafkaRebalance has rebalance annotation removed as
-                                                    // action specified by user has been completed.
-                                                    KafkaRebalance patchedKafkaRebalance = new KafkaRebalanceBuilder(updatedKafkaRebalance)
-                                                            .editMetadata()
+                                                    if (rebalanceAnnotation.equals(KafkaRebalanceAnnotation.unknown)) {
+                                                        return kafkaRebalanceOperator.patchAsync(reconciliation, updatedKafkaRebalance);
```
ppatierno, line 392: we are leaving annotation no why are we patching? any change to the KR resource?

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceAssemblyOperator.java
```diff
@@ -717,7 +719,13 @@ public MapAndStatus(T loadMap, K status) {
             return onNew(reconciliation, host, apiClient, kafkaRebalance, rebalanceOptionsBuilder);
         } else {
             // Stay in the current NotReady state, returning null as next state
-            return Future.succeededFuture(new MapAndStatus<>(null, null));
+            if (rawRebalanceAnnotation(kafkaRebalance) == null) {
```
ppatierno, line 722: if raw annonation is `null`, it means to be `none` as enum, why are we anyway doing this check here? 
Anyway the code can be simpler. You are anyway calling `validate` to get the conditions but then adding the invalid annotation condition only if it's not `none` (so any other different from the valid `refresh`).
So in the `else` branch, you could have just something like this? 

```java
Set<Condition> conditions = validate(reconciliation, kafkaRebalance);
            if (rebalanceAnnotation != KafkaRebalanceAnnotation.none) {
                conditions.add(StatusUtils.buildWarningCondition("InvalidAnnotation", "Wrong annotation value: " + rawRebalanceAnnotation(kafkaRebalance) + ". The rebalance resource supports the following annotations in `NotReady` state: refresh"));
            }
            return Future.succeededFuture(buildRebalanceStatus(null, KafkaRebalanceState.NotReady, conditions));
```

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceAssemblyOperator.java
```diff
@@ -760,12 +768,16 @@ public MapAndStatus(T loadMap, K status) {
...
                                     p.complete(buildRebalanceStatus(null, KafkaRebalanceState.Stopped, validate(reconciliation, currentKafkaRebalance)));
                                 } else {
-
                                     requestRebalance(reconciliation, host, apiClient, kafkaRebalance, true, rebalanceOptionsBuilder,
                                             currentKafkaRebalance.getStatus().getSessionId())
                                         .onSuccess(rebalanceMapAndStatus -> {
                                             // If the returned status has an optimization result then the rebalance proposal
                                             // is ready, so stop the polling
+                                            KafkaRebalanceStatus status = rebalanceMapAndStatus.getStatus();
+                                            if (rebalanceAnnotation(reconciliation, currentKafkaRebalance) != null) {
```
ppatierno, line 777: aren't we trying to always return an enum (and not `null`) from the `rebalanceAnnotation` method? what would `null` mean?

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceAssemblyOperator.java
```diff
@@ -836,8 +848,14 @@ public MapAndStatus(T loadMap, K status) {
...
+                    return configMapOperator.getAsync(kafkaRebalance.getMetadata().getNamespace(), kafkaRebalance.getMetadata().getName()).compose(loadmap -> Future.succeededFuture(new MapAndStatus<>(loadmap, buildRebalanceStatusFromPreviousStatus(kafkaRebalance.getStatus(), validate(reconciliation, kafkaRebalance)))));
+                } else {
+                    LOGGER.warnCr(reconciliation, "Ignore annotation {}={}", ANNO_STRIMZI_IO_REBALANCE, rebalanceAnnotation);
+                    Set<Condition> conditions = validate(reconciliation, kafkaRebalance);
+                    conditions.add(StatusUtils.buildWarningCondition("InvalidAnnotation", "Wrong annotation value: " + rawRebalanceAnnotation(kafkaRebalance) + ". The rebalance resource supports the following annotations in 'ProposalReady' state: approve, refresh, stop."));
+                    return configMapOperator.getAsync(kafkaRebalance.getMetadata().getNamespace(), kafkaRebalance.getMetadata().getName()).compose(loadmap -> Future.succeededFuture(new MapAndStatus<>(loadmap, buildRebalanceStatusFromPreviousStatus(kafkaRebalance.getStatus(), conditions))));
+                }
```
ppatierno, line 858: even in this piece of code you can always get the conditions from validate method and the adding the invalid annotation condition only if needed and having just one common code (see my suggestion before). 
Other than this I don't understand the check on  `null`, we should always check on annotations values. What `null` means here?

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceAssemblyOperator.java
```diff
@@ -891,6 +909,10 @@ public MapAndStatus(T loadMap, K status) {
                                         });
                                 } else {
                                     LOGGER.infoCr(reconciliation, "Getting Cruise Control rebalance user task status");
+                                    Set<Condition> conditions = validate(reconciliation, kafkaRebalance);
+                                    if (rawRebalanceAnnotation(currentKafkaRebalance) != null) {
```
ppatierno, line 913: again, let's check annotation value not `null`, or what does it mean?

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceAssemblyOperator.java
```diff
@@ -992,8 +1014,14 @@ public MapAndStatus(T loadMap, K status) {
...
-            LOGGER.warnCr(reconciliation, "Ignore annotation {}={}", ANNO_STRIMZI_IO_REBALANCE, rebalanceAnnotation);
-            return Future.succeededFuture(buildRebalanceStatus(null, KafkaRebalanceState.Stopped, validate(reconciliation, kafkaRebalance)));
+            if (rawRebalanceAnnotation(kafkaRebalance) == null) {
+                return Future.succeededFuture(buildRebalanceStatus(null, KafkaRebalanceState.Stopped, validate(reconciliation, kafkaRebalance)));
+            } else {
+                LOGGER.warnCr(reconciliation, "Ignore annotation {}={}", ANNO_STRIMZI_IO_REBALANCE, rawRebalanceAnnotation(kafkaRebalance));
+                Set<Condition> conditions = validate(reconciliation, kafkaRebalance);
+                conditions.add(StatusUtils.buildWarningCondition("InvalidAnnotation", "Wrong annotation value: " + rawRebalanceAnnotation(kafkaRebalance) + ". The rebalance resource supports the following annotations in `Stopped` state: refresh"));
+                return Future.succeededFuture(buildRebalanceStatus(null, KafkaRebalanceState.Stopped, conditions));
+            }
```
ppatierno, line 1024: all the same code to be simplified and why `null` check

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceAssemblyOperator.java
```diff
@@ -1018,8 +1046,14 @@ public MapAndStatus(T loadMap, K status) {
...
+            if (rawRebalanceAnnotation(kafkaRebalance) == null) {
+                return configMapOperator.getAsync(kafkaRebalance.getMetadata().getNamespace(), kafkaRebalance.getMetadata().getName()).compose(loadmap -> Future.succeededFuture(new MapAndStatus<>(loadmap, buildRebalanceStatusFromPreviousStatus(kafkaRebalance.getStatus(), validate(reconciliation, kafkaRebalance)))));
+            } else {
+                LOGGER.warnCr(reconciliation, "{}: Ignore annotation {}={}", ANNO_STRIMZI_IO_REBALANCE, rawRebalanceAnnotation(kafkaRebalance));
+                Set<Condition> conditions = validate(reconciliation, kafkaRebalance);
+                conditions.add(StatusUtils.buildWarningCondition("InvalidAnnotation", "Wrong annotation value: " + rawRebalanceAnnotation(kafkaRebalance) + ". The rebalance resource supports the following annotations in `Ready` state: refresh"));
+                return configMapOperator.getAsync(kafkaRebalance.getMetadata().getNamespace(), kafkaRebalance.getMetadata().getName()).compose(loadmap -> Future.succeededFuture(new MapAndStatus<>(loadmap, buildRebalanceStatusFromPreviousStatus(kafkaRebalance.getStatus(), conditions))));
+            }
```
ppatierno, line 1056: ditto as above in other places.

ppatierno (changes requested):

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceAssemblyOperator.java
ShubhamRwt: While applying the suggestion, One edge case slipped into my brain. What if we are on `Ready` state and a user passes `none` annotation? It will make the condition to be false and we will not get the warning

ShubhamRwt: Should we treat `none` as `null` only then ? ie if someone passes `none`, we don't do anything and dont add the warning either?

ppatierno: `none` is not an annotation value :-) ... `none` is the enum val corresponding to the fact that there is NO annotation on the `KafkaRebalance` resource (so raw annotation is `null` because getting annotation from the custom resource return `null` because it doesn't exist).

ppatierno:
@ShubhamRwt finally I got your edge case about a user specifying `strimzi.io/rebalance=none` as annotation that in our current code is translated into the `KafkaRebalanceAnnotation.none` enum value which has an actual different meaning for us ... it's NO annotation and not an annotation with `none` as a value.
So my thought is that we should completely get rid of the `KafkaRebalanceAnnotation.none`.
We need a non ambiguous way to know that there is NO annotation.
We could go in two possible directions:

* using the `rawRebalanceAnnotation` and if it returns `null` to check there is NO annotation. Today, this method is used just for logging or extracting the raw value to convert into the corresponding `KafkaRebalanceAnnotation` enum value. It should be used everywhere we have a check on `KafkaRebalanceAnnotation.none` (as `if` condition or a `case` in a switch). 
* changing the `rebalanceAnnotation` method by returning an `Optional<KafkaRebalanceAnnotation>`. It should return an empty one when there is NO annotation. Checking if it's empty should be used everywhere we have a check on `KafkaRebalanceAnnotation.none` (as `if` condition or a `case` in a switch).

I would prefer the second one because it's more clean imho.
We are passing a `KafkaRebalanceAnnotation` parameter in the reconcile and down the road to the methods for computing the next state. It would mean more refactoring in the `KafkaRebalanceAssemblyOperator` and corresponding tests but it should end and a cleaner solution.

I would like to hear from @kyguy and @tomncooper as well, wdyt?

kyguy:
For something quick and dirty, could we throw an `IllegalArgumentException` in the `rebalanceAnnotation` method when a user provides a rebalanceAnnotationValue with a value of "none" to force the edge case to be correctly labeled as `KafkaRebalanceAnnotation.unknown` instead of `KafkaRebalanceAnnotation.none`? 

If not, and we are keen on stripping out the KafkaRebalanceAnnotation.none enum, I am happy with Paolo's suggestion: 

> changing the `rebalanceAnnotation` method by returning an `Optional<KafkaRebalanceAnnotation>. It should return an empty one when there is NO annotation. Checking if it's empty should be used everywhere we have a check on `KafkaRebalanceAnnotation.none` (as `if` condition or a `case` in a switch).

ppatierno:
@kyguy suggestion could be a quick hack but I am not sure we need it. I would be more for handling it properly. Let's see what @tomncooper thinks as well ... then we can have a final decision.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceAssemblyOperator.java
```diff
@@ -388,18 +388,21 @@ private String rebalanceStateConditionType(KafkaRebalanceStatus status) {
...
+                                                    if (rebalanceAnnotation.equals(KafkaRebalanceAnnotation.unknown)) {
+                                                        return kafkaRebalanceOperator.patchAsync(reconciliation, updatedKafkaRebalance);
+                                                    } else {
+                                                        LOGGER.debugCr(reconciliation, "Removing annotation {}={}",
+                                                                ANNO_STRIMZI_IO_REBALANCE,
+                                                                rawRebalanceAnnotation(updatedKafkaRebalance));
+                                                        // Updated KafkaRebalance has rebalance annotation removed as
+                                                        // action specified by user has been completed.
+                                                        KafkaRebalance patchedKafkaRebalance = new KafkaRebalanceBuilder(updatedKafkaRebalance)
+                                                                .editMetadata()
```
tomncooper, line 400: If you are not editing the metadata you don't need to call `editMetadata`?

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceAssemblyOperator.java
tomncooper: This is a good point, can you not just return `updatedKafkaRebalance` (or whatever results in essentially a no-op)?

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceAssemblyOperator.java
```diff
@@ -466,8 +468,8 @@ private KafkaRebalanceStatus buildRebalanceStatusFromPreviousStatus(KafkaRebalan
 
     private MapAndStatus<ConfigMap, KafkaRebalanceStatus> buildRebalanceStatus(String sessionID, KafkaRebalanceState cruiseControlState, Set<Condition> validation) {
         List<Condition> conditions = new ArrayList<>();
-        conditions.add(StatusUtils.buildRebalanceCondition(cruiseControlState.toString()));
         conditions.addAll(validation);
+        conditions.add(StatusUtils.buildRebalanceCondition(cruiseControlState.toString()));
```
tomncooper, line 472: Not that I think it matters, but why have you swapped the order of these?

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceAssemblyOperator.java
tomncooper: I think Paolo's suggestion is a good one. It would be better to switch on the annotation.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceAssemblyOperator.java
```diff
@@ -1174,13 +1208,52 @@ public MapAndStatus(T loadMap, K status) {
      */
     private KafkaRebalanceAnnotation rebalanceAnnotation(Reconciliation reconciliation, KafkaRebalance kafkaRebalance) {
         String rebalanceAnnotationValue = rawRebalanceAnnotation(kafkaRebalance);
+        KafkaRebalanceState state = state(kafkaRebalance);
         KafkaRebalanceAnnotation rebalanceAnnotation;
         try {
             rebalanceAnnotation = rebalanceAnnotationValue == null ?
                     KafkaRebalanceAnnotation.none : KafkaRebalanceAnnotation.valueOf(rebalanceAnnotationValue);
```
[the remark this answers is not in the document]
tomncooper: This line could be inside the rawRebalanceAnnotation method

tomncooper (changes requested):
This looks like a good start but I think the questions raised by @ppatierno need to be addressed.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceAssemblyOperator.java
tomncooper: ??

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceAssemblyOperator.java
```diff
@@ -717,7 +719,11 @@ public MapAndStatus(T loadMap, K status) {
             return onNew(reconciliation, host, apiClient, kafkaRebalance, rebalanceOptionsBuilder);
         } else {
             // Stay in the current NotReady state, returning null as next state
-            return Future.succeededFuture(new MapAndStatus<>(null, null));
+            Set<Condition> conditions = validate(reconciliation, kafkaRebalance);
+            if (rebalanceAnnotation != KafkaRebalanceAnnotation.none) {
```
tomncooper, line 723: Can we no apply a `Refresh` annotation to a `NotReady` KafkaRebalance? Is that not valid / no-op?

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceAssemblyOperator.java
```diff
@@ -760,12 +766,16 @@ public MapAndStatus(T loadMap, K status) {
...
                                 } else {
-
                                     requestRebalance(reconciliation, host, apiClient, kafkaRebalance, true, rebalanceOptionsBuilder,
                                             currentKafkaRebalance.getStatus().getSessionId())
                                         .onSuccess(rebalanceMapAndStatus -> {
                                             // If the returned status has an optimization result then the rebalance proposal
                                             // is ready, so stop the polling
+                                            KafkaRebalanceStatus status = rebalanceMapAndStatus.getStatus();
+                                            if (rebalanceAnnotation(reconciliation, currentKafkaRebalance) != KafkaRebalanceAnnotation.none) {
+                                                status.addCondition(StatusUtils.buildWarningCondition("InvalidAnnotation", "Wrong annotation value: " + kafkaRebalance.getMetadata().getAnnotations() + ". The rebalance resource supports the following annotations in 'PendingProposal' state: approve, refresh, stop."));
```
tomncooper, line 776: I think the error message is wrong. We don't support `approve` and `refresh` annotations on a `ProposalPending` state? It think we only support `stop`, at least that is what the code suggests? 

Also, can you split this over a few lines to make it easier to read on thinner screens.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceAssemblyOperator.java
```diff
@@ -1018,8 +1037,12 @@ public MapAndStatus(T loadMap, K status) {
...
-            return configMapOperator.getAsync(kafkaRebalance.getMetadata().getNamespace(), kafkaRebalance.getMetadata().getName()).compose(loadmap -> Future.succeededFuture(new MapAndStatus<>(loadmap, buildRebalanceStatusFromPreviousStatus(kafkaRebalance.getStatus(), validate(reconciliation, kafkaRebalance)))));
+            Set<Condition> conditions = validate(reconciliation, kafkaRebalance);
+            if (rebalanceAnnotation != KafkaRebalanceAnnotation.none) {
+                LOGGER.warnCr(reconciliation, "{}: Ignore annotation {}={}", ANNO_STRIMZI_IO_REBALANCE, rawRebalanceAnnotation(kafkaRebalance));
+                conditions.add(StatusUtils.buildWarningCondition("InvalidAnnotation", "Wrong annotation value: " + kafkaRebalance.getMetadata().getAnnotations() + ". The rebalance resource supports the following annotations in `Ready` state: refresh"));
+            }
+            return configMapOperator.getAsync(kafkaRebalance.getMetadata().getNamespace(), kafkaRebalance.getMetadata().getName()).compose(loadmap -> Future.succeededFuture(new MapAndStatus<>(loadmap, buildRebalanceStatusFromPreviousStatus(kafkaRebalance.getStatus(), conditions))));
```
tomncooper, line 1045: I know this was like this in the original code but can we breakup these huge lines a bit.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceAssemblyOperator.java
```diff
@@ -1174,13 +1197,42 @@ public MapAndStatus(T loadMap, K status) {
...
-                    KafkaRebalanceAnnotation.none : KafkaRebalanceAnnotation.valueOf(rebalanceAnnotationValue);
+            if (rebalanceAnnotationValue == null) {
+                rebalanceAnnotation = KafkaRebalanceAnnotation.none;
+            } else if (rebalanceAnnotationValue.equals("none")) {
+                rebalanceAnnotation = KafkaRebalanceAnnotation.unknown;
+            } else {
+                rebalanceAnnotation = KafkaRebalanceAnnotation.valueOf(rebalanceAnnotationValue);
+            }
+
+            if (rebalanceAnnotation != KafkaRebalanceAnnotation.none) {
```
tomncooper, line 1212: Why is this switch statement here? You already have code to exclude and highlight inappropriate annotation types in each of the state methods (`onProposalReady`, `onRebalancing` etc). This seems redundant? It also means you have maintain the mapping of what annotations are valid in each state in two places (this switch and in the state method). 

I suggest removing this switch statement and keeping the logic of what annotations are valid inside each state method.

tomncooper (changes requested):
A few suggested changes but this is nearly there.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceAssemblyOperator.java
ShubhamRwt: No particular reason. I have restored it back to the way it was

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceAssemblyOperator.java
ShubhamRwt: Resolved

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceAssemblyOperator.java
ShubhamRwt: I guess we might require that since we are setting the rebalance annotation to `unknown` here

tomncooper: Ok, after going through this with @ShubhamRwt he has convinced me this actually needed. However, I am still concerned that we have 2 locations to maintain valid annotations for each state. I wonder if it would be better, instead of doing the checking in the rebalanceAnnotations method and have that value given to `reconcile`, to have a generic `validateAnnotation` method that gets called from within each `onState` method. Something like:
```java
List<KafkaRebalanceAnnotations> validAnnotations = Arrays.asList(KafkaRebalanceAnnotation.stop);
KafkaRebalanceAnnotation rebalanceAnnotation = validateAnnotation("onRebalancing", validAnnotations);
```
That way, the list and logic for valid annotations for each state are kept only in the `onState` methods. 

WDYT @ppatierno? Am I over thinking this?

ppatierno: I took a "static" look at the code and I think that what @tomncooper is saying should work. As an example ...
Imagine that someone set the `stop` annotation but we are in the `ProposalReady` state. This method would translate the `stop` into `unknown` but let's assume we remove the switch statement here and the annotation will remain `stop`. When the `onProposalReady` is executed, the `stop` will be handled in the `default` case of the switch and warning about the wrong annotation.
Even for the other `onXXXXX` method, any kind of not allowed annotation seems to be handled the same way by printing the warning.
Am I right? If yes, it could assume that the switch here can be removed as Tom's suggestion.

ShubhamRwt: @tomncooper @ppatierno I just have a small doubt. The generic methodwill check all if the rebalance annotation is correct or not as per the state, right ? But Iam not actually overriding the `rebalanceAnnotation` to `unknown` anywhere. So when I get back to the `reconcile` method, wouldn't this condition fail -->
```          if (rebalanceAnnotation.equals(KafkaRebalanceAnnotation.unknown)) {
                                                    } ```

Comment (ShubhamRwt) [reply cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceAssemblyOperator.java:1219]:
``` if (rebalanceAnnotation.equals(KafkaRebalanceAnnotation.unknown)) { 
}```

File: api/src/main/java/io/strimzi/api/kafka/model/balancing/KafkaRebalanceState.java [hunk 27]
```diff
@@ -53,23 +55,35 @@
      *     <dt>ProposalReady</dt><dd>If the user sets annotation strimzi.io/rebalance=refresh and the proposal is already ready.</dd>
      * </dl>
      */
-    Stopped,
+    Stopped(List.of(KafkaRebalanceAnnotation.stop, KafkaRebalanceAnnotation.refresh)),
```
tomncooper, line 58: Can we stop a stopped Rebalance? I think it is just refresh?

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceAssemblyOperator.java
```diff
@@ -717,7 +719,12 @@ public MapAndStatus(T loadMap, K status) {
             return onNew(reconciliation, host, apiClient, kafkaRebalance, rebalanceOptionsBuilder);
         } else {
             // Stay in the current NotReady state, returning null as next state
-            return Future.succeededFuture(new MapAndStatus<>(null, null));
+            Set<Condition> conditions = validate(reconciliation, kafkaRebalance);
+            if (!KafkaRebalanceState.NotReady.validateAnnotation(rebalanceAnnotation)) {
+                conditions.add(StatusUtils.buildWarningCondition("InvalidAnnotation", "Wrong annotation value: "
+                        + kafkaRebalance.getMetadata().getAnnotations() + ". The rebalance resource supports the following annotations in `NotReady` state: refresh"));
```
tomncooper, line 725: If you want to futureproof this you can get the valid annotations from the KafkaRebalanceState enum itself by adding a `validStateString` method to the enum that returns a string of the valid annotations for that state? Then you can call that in the message. That way if you add new valid annotations to a state you don't have to change this string.

tomncooper: Also this batch of code is basically the same in all the onState methods. Could you make it a generic method that just accepts a KafkaRebalanceState and the rebalanceAnnotation?

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceAssemblyOperator.java
```diff
@@ -760,12 +767,17 @@ public MapAndStatus(T loadMap, K status) {
...
                                     p.complete(buildRebalanceStatus(null, KafkaRebalanceState.Stopped, validate(reconciliation, currentKafkaRebalance)));
                                 } else {
-
                                     requestRebalance(reconciliation, host, apiClient, kafkaRebalance, true, rebalanceOptionsBuilder,
                                             currentKafkaRebalance.getStatus().getSessionId())
                                         .onSuccess(rebalanceMapAndStatus -> {
                                             // If the returned status has an optimization result then the rebalance proposal
                                             // is ready, so stop the polling
+                                            KafkaRebalanceStatus status = rebalanceMapAndStatus.getStatus();
+                                            if (!KafkaRebalanceState.PendingProposal.validateAnnotation(rebalanceAnnotation(reconciliation, currentKafkaRebalance))) {
```
tomncooper, line 776: Again this is repeated code, can you make it a generic method that takes a state and the annotation?

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceAssemblyOperator.java
ShubhamRwt: Generic method inside the KRAO class itself, right?

tomncooper: Yes, just a private method so you don't end up repeating the same code pattern over and over.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceAssemblyOperator.java
```diff
@@ -837,7 +843,11 @@ public MapAndStatus(T loadMap, K status) {
                 return requestRebalance(reconciliation, host, apiClient, kafkaRebalance, true, rebalanceOptionsBuilder);
             default:
                 LOGGER.warnCr(reconciliation, "Ignore annotation {}={}", ANNO_STRIMZI_IO_REBALANCE, rebalanceAnnotation);
-                return configMapOperator.getAsync(kafkaRebalance.getMetadata().getNamespace(), kafkaRebalance.getMetadata().getName()).compose(loadmap -> Future.succeededFuture(new MapAndStatus<>(loadmap, buildRebalanceStatusFromPreviousStatus(kafkaRebalance.getStatus(), validate(reconciliation, kafkaRebalance)))));
+                Set<Condition> conditions = validate(reconciliation, kafkaRebalance);
+                conditions.add(StatusUtils.buildWarningCondition("InvalidAnnotation", "Wrong annotation value: " +
```
tomncooper, line 847: Why aren't you using the `validateAnnotation` method here?

ShubhamRwt: In `ProposalReady` state every annotaion is managed already in the cases if I am not wrong and the last case is when we apply some wrong annotation

ShubhamRwt: I guess I can still do it here, since the annotation will always be wrong in this case

ShubhamRwt: Added @tomncooper

tomncooper:
@ppatierno Could you have another pass over this, there have been quite a few changes since your lass look.

api/src/main/java/io/strimzi/api/kafka/model/balancing/KafkaRebalanceState.java
```diff
@@ -53,23 +55,37 @@
...
+
+    KafkaRebalanceState(List<KafkaRebalanceAnnotation> validAnnotations) {
+        this.validAnnotations = validAnnotations;
+    }
+
+    public boolean validateAnnotation(KafkaRebalanceAnnotation annotation) {
+        return validAnnotations.contains(annotation);
+    }
+
+    public String validAnnotationString() {
```
ppatierno, line 88: why not overriding the `toString` instead of naming a new method?
Something like:

```java
@Override
    public String toString() {
        return "KafkaRebalanceState{" +
                "validAnnotations=" + validAnnotations +
                '}';
    }
```

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceAssemblyOperator.java
```diff
@@ -1200,6 +1220,15 @@ private String rawRebalanceAnnotation(KafkaRebalance kafkaRebalance) {
...
     }
 
+    private Condition validateAnnotation(KafkaRebalanceState state, KafkaRebalanceAnnotation rebalanceAnnotation) {
+        Condition condition = null;
+        if (!KafkaRebalanceState.Ready.validateAnnotation(rebalanceAnnotation)) {
+            condition =  StatusUtils.buildWarningCondition("InvalidAnnotation", "Wrong annotation value: " + rebalanceAnnotation
+                    + ". The rebalance resource supports the following annotations in `Ready` state: " + state.validAnnotationString());
+        }
+        return condition;
+    }
```
ppatierno, line 1230: not sure I understand this method. 
AFAIU, it looks like It should validate that the `rebalanceAnnotation` is valid against the `state` (both are provided as parameter) and then returning a warning condition in the case it was not valid. Instead, it's checking that the `rebalanceAnnotation` is valid for just the `Ready` state? Why?
This method is called in a lot of places where the `Ready` state is not involved.
Other than this, I would add Javadoc as well.

ppatierno:
@tomncooper @ShubhamRwt I agree it's in the right direction but I left a suggestion and a comment that could raise a potential mistake ... or just me not getting what the code is doing ;-)

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceAssemblyOperator.java
ShubhamRwt: Thanks for catching it Paolo. I might have not changed it

api/src/main/java/io/strimzi/api/kafka/model/balancing/KafkaRebalanceState.java
ShubhamRwt: Hi, I guess we are making use of the normal toString() method at some places so overriding it might cause errors.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceAssemblyOperator.java
ShubhamRwt: Resolved this issue now

api/src/main/java/io/strimzi/api/kafka/model/balancing/KafkaRebalanceState.java
ShubhamRwt: for eg. -> `KafkaRebalanceState.valueOf(rebalanceStateType)` If Iam not wrong valueOf() does call the `toString` method, right?

ppatierno: Not sure about the `valueOf()` but what I know is that `toString()` calls `Enum.name()` by default. Can you try maybe?

ShubhamRwt: I did tried to override the `toString()` method but I was getting some errors and  this `KafkaRebalanceState.valueOf(rebalanceStateType)` was in the stacktrace

ppatierno: As agreed offline, let's stay as we are without overriding `toString()` but making `validAnnotationString()` a getter for the `validAnnotations` instead.

ppatierno (approved):
LGTM!

tomncooper (approved):

scholzj (approved):
LGTM asuming the tests pass

ShubhamRwt:
Done @scholzj

kyguy (approved):

scholzj:
@ShubhamRwt Can you check the failing system test `ReconciliationST.testPauseReconciliationInKafkaRebalanceAndTopic`? It seems to be related to your PR changes.

ShubhamRwt:
Strange. The test passed fine for me locally. Let me see again whats causing this

api/src/main/java/io/strimzi/api/kafka/model/balancing/KafkaRebalanceState.java
```diff
@@ -34,7 +36,7 @@
      *     <dt>ProposalReady</dt><dd>When the user sets annotation strimzi.io/rebalance=refresh and the proposal is already ready.</dd>
      * </dl>
      */
-    ProposalReady,
+    ProposalReady(List.of(KafkaRebalanceAnnotation.approve, KafkaRebalanceAnnotation.refresh, KafkaRebalanceAnnotation.none)),
```
tomncooper, line 39: Why is `none` an allowed annotation? It doesn't serve any purpose? I thought you had explicit logic to handle the (very unlikely case) the user actually put `annotation=none`?

api/src/main/java/io/strimzi/api/kafka/model/balancing/KafkaRebalanceState.java
```diff
@@ -53,23 +55,37 @@
...
      */
-    NotReady,
+    NotReady(List.of(KafkaRebalanceAnnotation.refresh)),
     /**
      * The rebalance is complete and there is no transition from this state.
      * The resource is eligible for garbage collection after a configurable delay.
      * There is no transition from this state to a new one.
      */
-    Ready,
+    Ready(List.of(KafkaRebalanceAnnotation.refresh, KafkaRebalanceAnnotation.none)),
```
tomncooper, line 72: Again, this enum is the source of truth for what annotations are valid. `none` is not a valid annotation.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceAssemblyOperator.java
```diff
@@ -1191,12 +1208,18 @@ public MapAndStatus(T loadMap, K status) {
     private KafkaRebalanceAnnotation rebalanceAnnotation(Reconciliation reconciliation, KafkaRebalance kafkaRebalance) {
         String rebalanceAnnotationValue = rawRebalanceAnnotation(kafkaRebalance);
         KafkaRebalanceAnnotation rebalanceAnnotation;
+
         try {
-            rebalanceAnnotation = rebalanceAnnotationValue == null ?
-                    KafkaRebalanceAnnotation.none : KafkaRebalanceAnnotation.valueOf(rebalanceAnnotationValue);
+            if (rebalanceAnnotationValue == null) {
+                rebalanceAnnotation = KafkaRebalanceAnnotation.none;
+            } else if (rebalanceAnnotationValue.equals("none")) {
```
tomncooper, line 1215: So you do have logic for dealing with `none`. Why did you add that option to the enum?

api/src/main/java/io/strimzi/api/kafka/model/balancing/KafkaRebalanceState.java
ShubhamRwt: Sorry @tomncooper. Looks like I forgot to stash these enum changes. Let me do it

ShubhamRwt: Removed

scholzj:
```
[ERROR] Errors: 
[ERROR] io.strimzi.systemtest.operators.ReconciliationST.testPauseReconciliationInKafkaRebalanceAndTopic(ExtensionContext)
[ERROR]   Run 1: ReconciliationST.testPauseReconciliationInKafkaRebalanceAndTopic:182 » Wait Ti...
[ERROR]   Run 2: ReconciliationST.testPauseReconciliationInKafkaRebalanceAndTopic:182 » Wait Ti...
[ERROR]   Run 3: ReconciliationST.testPauseReconciliationInKafkaRebalanceAndTopic:182 » Wait Ti...
```

Looks related to this PR.

ShubhamRwt:
Yes @scholzj and I guess the fix I just pushed will fix the failure

ShubhamRwt:
Just making final checks

PR #5960, opened 2021-11-29
Topic config validation

sknot-rh:
### Type of change
- Enhancement / new feature

### Description
In this PR the validation of `spec.config` for KafkaTopic type is added. To be able check the config, we need to generate topic config model, similarly as we do for brokers config model.
When we have the config model, we can check the expected and actual data type of provided config entries.
The config model is tied to the kafka version, so we need let the TO know, what version of kafka it is used in the brokers. It is done in this PR by adding an envVar (open for better suggestions). 
Because until now we used config classes in the CO, I needed to do some refactoring.

We need constants from `KafkaVersionTestUtils` accessible in TO. We cannot move the entire class as it is using KafkaVersion classes. So I created a new class `KafkaVersionTestConstants` containing desired constants.

//hopefully fixed version of https://github.com/strimzi/strimzi-kafka-operator/pull/5880

### Checklist
- [ ] Write tests
- [ ] Update documentation
- [x] Update CHANGELOG.md


operator-common/src/main/java/io/strimzi/operator/cluster/model/KafkaVersionTestConstants.java
```diff
@@ -0,0 +1,41 @@
+/*
+ * Copyright Strimzi authors.
+ * License: Apache License 2.0 (see the file LICENSE or http://apache.org/licenses/LICENSE-2.0.html).
+ */
+package io.strimzi.operator.cluster.model;
+
+public class KafkaVersionTestConstants {
```
scholzj, line 7: I'm not convinced this is a good idea. These are semi-artificial data used by some specific tests in the CO. They are designed for the tests which use them for upgrade testing etc. The do not necessarily need to correspond to what is in the actual cluster. For example as you can see, they are missing the 2.8.1 release. In different situations they will be missing other versions as well. It seems better to me to keep these as they were. 

For the TO tests, is there any really any reason to not load the data from the actual versions file?

cluster-operator/src/test/java/io/strimzi/operator/cluster/operator/assembly/ConnectorMockTest.java
```diff
@@ -24,9 +24,9 @@
 import io.strimzi.operator.PlatformFeaturesAvailability;
 import io.strimzi.operator.cluster.ClusterOperatorConfig;
 import io.strimzi.operator.cluster.FeatureGates;
-import io.strimzi.operator.cluster.KafkaVersionTestUtils;
 import io.strimzi.operator.cluster.ResourceUtils;
 import io.strimzi.operator.cluster.model.KafkaConnectCluster;
+import io.strimzi.operator.cluster.KafkaVersionTestUtils;
```
scholzj, line 29: Can we keep the original ordering in this and the other files? It helps to minimize the footprint of this PR.

cluster-operator/src/test/java/io/strimzi/operator/cluster/operator/resource/KafkaSpecCheckerTest.java
```diff
@@ -12,11 +12,11 @@
 import io.strimzi.api.kafka.model.storage.EphemeralStorage;
 import io.strimzi.api.kafka.model.storage.EphemeralStorageBuilder;
 import io.strimzi.api.kafka.model.storage.JbodStorageBuilder;
-import io.strimzi.operator.cluster.KafkaVersionTestUtils;
 import io.strimzi.operator.cluster.ResourceUtils;
 import io.strimzi.operator.cluster.model.KafkaCluster;
 import io.strimzi.operator.cluster.model.KafkaConfiguration;
 import io.strimzi.operator.cluster.model.KafkaVersion;
+import io.strimzi.operator.cluster.KafkaVersionTestUtils;
```
scholzj, line 19: Same as above, maybe the changes in this file are not needed? Both here and below?

scholzj (commented):
Some more nits, but looks mostly good otherwise.

scholzj (approved):
Looks good in general. But the TO related part has to be reviewed by some TO expert.

sknot-rh:
> Looks good in general. But the TO related part has to be reviewed by some TO expert.

@tombentley Could you, please?

config-model-generator/src/main/java/io/strimzi/build/kafka/metadata/CommonConfigModelGenerator.java
```diff
@@ -0,0 +1,263 @@
...
+import java.util.Properties;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+import static java.security.AccessController.doPrivileged;
+
+/**
+ * Abstract class for loading a config model from Kafka classes.
+ */
+public abstract class CommonConfigModelGenerator {
```
tombentley, line 43: Common → Abstract?

config-model-generator/src/main/java/io/strimzi/build/kafka/metadata/CommonConfigModelGenerator.java
```diff
@@ -0,0 +1,263 @@
...
+                    throw new IllegalStateException(str);
+                }
+            } else {
+                throw new IllegalStateException(str);
+            }
+            return descriptor;
+        } catch (RuntimeException e) {
+            throw new RuntimeException("Invalid range " + str + " for key " + key.name, e);
+        }
+    }
```
tombentley, line 175: Do all of these need to be `protected`?

systemtest/src/test/java/io/strimzi/systemtest/operators/CustomResourceStatusST.java
```diff
@@ -605,8 +605,9 @@ void assertKafkaTopicWrongMinInSyncReplicasStatus(String topicName, String inval
         assertThat(kafkaTopicStatus.getConditions().stream()
             .anyMatch(condition -> condition.getType().equals(NotReady.toString())), is(true));
         assertThat(kafkaTopicStatus.getConditions().stream()
-            .anyMatch(condition -> condition.getReason().equals("InvalidRequestException")), is(true));
+            .anyMatch(condition -> condition.getReason().equals("InvalidTopicException")), is(true));
```
tombentley, line 608: Technically this is a compatibility break.

tombentley (commented):
Nits, except the compatibility break. Can we revert the name of the exception used in the condition?

systemtest/src/test/java/io/strimzi/systemtest/operators/CustomResourceStatusST.java
sknot-rh: Ah. Ok. Anyway, I think this PR needs to be eventually rebased to https://github.com/strimzi/strimzi-kafka-operator/pull/5988

CHANGELOG.md
```diff
@@ -13,6 +13,7 @@
 * Update OPA Authorizer to 1.3.0
 * Update to Cruise Control version 2.5.79
 * Update Log4j2 to 2.17.0
+* Validate `spec.config` in the KafkaTopic resources against available Kafka topic configuration options
```
scholzj, line 16: This needs to be moved to 0.28.0!

scholzj (commented):
Stupid question ... but, if the validation requires a Kafka version to be passed to the container, it probably should:
* Be added to the standalone TO installation?
* Be documented for the standalone TO installation?

scholzj:
@sknot-rh Sorry, could you please do one more rebase? I think I somehow missed the notification about the changes.

topic-operator/src/main/java/io/strimzi/operator/topic/Config.java
```diff
@@ -210,6 +211,9 @@ private Value(String key, Type<? extends T> type, boolean required) {
     /** Do we use old ZooKeeper based TopicStore */
     public static final Value<Boolean> USE_ZOOKEEPER_TOPIC_STORE = new Value<>(TC_USE_ZOOKEEPER_TOPIC_STORE, BOOLEAN, "false");
 
+    /** The version of Kafka */
+    public static final Value<String> KAFKA_VERSION = new Value<>(TC_KAFKA_VERSION, STRING, "");
```
scholzj, line 215: What happens when the version is not specified? Will it use the latest or something like that?

sknot-rh: I'd bet that `RuntimeException` `"Error reading from classpath resource /kafka--topic-config-model.json"` will be thrown.

scholzj: I guess I merged this too early because it completely breaks the backwards compatibility for the TO. Maybe we should revert it? Can't it simple use the latest Kafka version as default?

PR #5966, opened 2021-11-30
docs: drain cleaner image

PaulRMellor:

**Documentation**
Adds drain cleaner image details to _Pushing container images to your own registry_ section in _Deploying Strimzi_

### Checklist

- [ ] Write tests
- [x] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/documentation/modules/con-common-configuration-properties.adoc
+++ b/documentation/modules/con-common-configuration-properties.adoc
@@ -304,10 +304,6 @@ spec:
     # ...
 ----
 
-*Strimzi Drain Cleaner image*
-
-The image for the Strimzi Drain Cleaner is specified in the `Deployment` installation file.
-
 [id='con-common-configuration-healthchecks-{context}']
 === `livenessProbe` and `readinessProbe` healthchecks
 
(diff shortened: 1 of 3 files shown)


documentation/modules/con-common-configuration-properties.adoc
```diff
@@ -304,6 +304,10 @@ spec:
     # ...
 ----
 
+*Strimzi Drain Cleaner image*
```
PaulRMellor, line 307: Not sure we need this here.

PaulRMellor:
Cherry-pick to 0.26 branch when merged

documentation/modules/con-common-configuration-properties.adoc
scholzj: I would leave this out => it is something completely unrelated to the API.

scholzj (commented):
TBH, I'm not sure this is really needed. The Drain Cleaner is very loosely coupled. Plus I would be afraid that we will forget to update these once we do new releases.

scholzj (approved):

sknot-rh (approved):

PaulRMellor:
@scholzj  could you please cherry-pick to 0.26

scholzj:
0.26.0 does not use Drain Cleaner 0.2.0 - that is used in main branch only. It uses 0.1.0. So we would probably need a separate PR for it directly into the 0.26.0 branch.

PaulRMellor:
> 0.26.0 does not use Drain Cleaner 0.2.0 - that is used in main branch only. It uses 0.1.0. So we would probably need a separate PR for it directly into the 0.26.0 branch.

Ah. Okay, I'll send one up.

PR #5978, opened 2021-12-02
Fixed date command to rename CA cert with date/time

ppatierno:
While running the procedure to renew my own CA cert, the command to rename the old certificate with expiration date/time failed with this error:

```shell
date: extra operand ‘15:38:23’
```

It seems that the output produced by the openssl command is something like this:

```shell
Thu  2 Dec 15:38:23 UTC 2021
```

and the subsequent date command consider that time as an extra operand so the need for double quotes.

This PR should addresses the problem.
Running this on Fedora 33.

### DIFF
--- a/documentation/modules/security/proc-renewing-your-own-ca-certificates.adoc
+++ b/documentation/modules/security/proc-renewing-your-own-ca-certificates.adoc
@@ -53,7 +53,7 @@ where _DATE_ is the certificate expiry date in the format _YEAR-MONTH-DAYTHOUR-M
 For example `ca-2018-09-27T17-32-00Z.crt`.
 +
 [source,shell,subs="+quotes"]
-mv ca.crt ca-$(date -u -d$(openssl x509 -enddate -noout -in ca.crt | sed 's/.*=//') +'%Y-%m-%dT%H-%M-%SZ').crt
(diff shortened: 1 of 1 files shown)


scholzj:
None of these seem to work on MacOS :-/. Is the exact data really important? IIRC it is not. So maybe we can use just a general instruction?

```
* Rename the ca.crt file to ca-<CertificateIssueDate.crt
```

But no strong feelings abotu it,

ppatierno:
damn! I have not MacOS to try, any change you get one working that I can test on Fedora? Otherwise I can go with your suggestion.

scholzj:
I don't really know TBH ... if the one you have works at least on other Linuxes it is probably fine to go with it if you want.

ppatierno:
Well I have just Fedora but I would assume that date could work same way other Linuxes. I would say let's merge this and see if someone else complain, at least we know it works fine on Fedora and I would assume on RHEL.

scholzj:
> Well I have just Fedora

Yeah, I thought that maybe someone else with some other distro can give it a try. But if you wanna just merge it, I can approve it ;-)

ppatierno:
@strimzi/maintainers anyone using a different Linux distrubution than Fedora who can try this?

scholzj (approved):

Issue #1050, opened 2018-10-31
topic-operator fails liveness 7x per hour


BradyWiseman:
Running strimzi/topic-operator:0.8.2 standalone on Oracle Cloud Infrastructure/Oracle Kubernetes Engine.
Using manifests as shipped in examples/install/topic-operator.
Once started, topic-operator fails liveness probe 5-7 times per hour and is restarted by k8s.  Pod logs show normal operation (using DEBUG flag), no exceptions or errors.

We have increased both cpu & memory requests, which seemed to help, then back to the previous pattern.

scholzj:
Hmm, interesting. Can you supply the full log from the Topic Operator? If you are using the `examples/install/topic-operator` installation, you are using it with your own Kafka cluster which is not deployed by the Strimzi Cluster Operator, right? Is this cluster stable? Are other applications connected to the Kafka and Zookeeper nodes running stable?

BradyWiseman:
Correct - kafka cluster is dead simple, not deployed by strimzi.  Yes, it is stable.  Nothing else is connected to it.
I will send logs after next lengthy run.

wmorgan6796:
I am having this issue on a much larger scale as well. I use the vanilla option as well
[logs-from-topic-operator-in-will-cluster-entity-operator-67ccb64c86-hrkv9.txt](https://github.com/strimzi/strimzi-kafka-operator/files/2555860/logs-from-topic-operator-in-will-cluster-entity-operator-67ccb64c86-hrkv9.txt)

scholzj:
@wmorgan6796 For how long can you see this repeat in the log file? On the startup, this can sometimes happen before all the components get in sync and up and running.

wmorgan6796:
I’m seeing this for hours straight. Today over a four hour period it had
over 20 restarts.

scholzj:
Hmm, that is strange. Could you please also share the log from the tls-sidecar / stunnel container in the same pod? Do you use persistent or ephemeral Zookeeper?

sjbodzo:
I am also seeing the same issue. We had a 5-node cluster deployed for ~6 hours that restarted 20+ times while sitting idle.

wmorgan6796:
Here are logs from the tls sidecar and the topic operator at the same time
[tls-sidecar-topic-operator.log](https://github.com/strimzi/strimzi-kafka-operator/files/2557816/tls-sidecar-topic-operator.log)
[...]

PR #5981, opened 2021-12-02
chore: bump topic-operator resource requests/limits

hobbsh:
### Type of change

Bugfix

### Description

The standalone topic-operator chokes when it starts up with a fairly obscure error. 

```
2021-12-02 22:14:51 INFO  KafkaStreamsTopicStoreService:52 - Starting ...
2021-12-02 22:14:52 WARN  BlockedThreadChecker: - Thread Thread[vert.x-eventloop-thread-0,5,main] has been blocked for 2299 ms, time limit is 2000 ms
2021-12-02 22:14:53 WARN  BlockedThreadChecker: - Thread Thread[vert.x-eventloop-thread-0,5,main] has been blocked for 3299 ms, time limit is 2000 ms
2021-12-02 22:14:54 WARN  BlockedThreadChecker: - Thread Thread[vert.x-eventloop-thread-0,5,main] has been blocked for 4299 ms, time limit is 2000 ms
```

This issue brought to light that the resource limits might be too tight, and it was correct.
https://github.com/strimzi/strimzi-kafka-operator/issues/1050

### Checklist

- [ ] Write tests
- [ ] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/install/topic-operator/05-Deployment-strimzi-topic-operator.yaml
+++ b/install/topic-operator/05-Deployment-strimzi-topic-operator.yaml
@@ -55,10 +55,10 @@ spec:
             periodSeconds: 30
           resources:
             limits:
-              memory: 96Mi
-              cpu: 100m
+              memory: 512Mi
+              cpu: 500m
             requests:
-              memory: 96Mi
+              memory: 256Mi
               cpu: 100m
   strategy:
     type: Recreate


scholzj (commented):
Thanks for the PR. The errors might (but might not) be caused by resources. But it would be good to understand why do you think exactly this amount of resources is the right one and why should everyone else use it. 

Also, You should probably not use the limit memory for Java which is not guaranteed and can cause problems -> the limit memory should be the same as the request memory.

Also as described in the README file in the `install/` folder, the files there should not be changed outside of release. If the resources should be really changed, you should do it in `packaging/install/...`.

@tombentley @sknot-rh Please have a look at the values

hobbsh:
> Thanks for the PR. The errors might (but might not) be caused by resources. But it would be good to understand why do you think exactly this amount of resources is the right one and why should everyone else use it.

I did not spend much time figuring out the minimum values, but it was immediately clear that the existing ones are too low. So I based the values off of the `user-operator` values. All of the operators include limits so I'm not sure what you mean - are you saying the requests should equal the limits?

scholzj:
> All of the operators include limits so I'm not sure what you mean - it seems like you are you saying only the memory requests should equal the memory limits (and not cpu)?

Yeah, they have limits. But the memory request and limit are always the same -> this is to make sure the memory is available. With CPU, it is easier to use the limit, but with memory it does not always work that well.

tombentley:
Is there any evidence that those threads were blocked by gc? How many `KafkaTopics` did you have?

hobbsh:
> Is there any evidence that those threads were blocked by gc? How many `KafkaTopics` did you have?

I couldn't tell you, I don't think metrics were being shipped. I was creating just 4 Topics.

scholzj:
Discussed on the community call on 13th January. This should be closed and if needed, we should do more testing about what the appropriatte resources would be.

Issue #3475, opened 2020-08-07
[Enhancement] Add extra Secret + Password checks to KafkaConnect


samuel-hawker:
**Is your feature request related to a problem? Please describe.**

For Kafka Connect, one common problem I've seen is people putting the actual password in their `KafkaConnect` Custom Resource (instead of the name of the password field in the secret)
It'd be a great enhanement if the operator updated the status of the KafkaConnect CR when you do that, so it says "you asked for password 'blah' but there is no key 'blah' in your secret" or something helpful and friendly like that.

Similarly, there has been some confusion around certs. Where a user has been using a KafkaUser cert in `spec.tls.trustedCertificates` where the KafkaConnect spec was expecting a cluster ca cert
It wasn't at all obvious to the user why his connect cluster was broken.

If we could do anything to tell if we've been given the right type of cert, it'd be a huge usability improvement.

An examples:
```
kind: KafkaConnectS2I
metadata:
  name: my-connect-cluster
spec:
  # ...
  authentication:
    passwordSecret:
      password: my-password
      secretName: my-secret
```

**Describe the solution you'd like**
The solution should be if the operator can, at runtime infer any reasoning why it is not working, this should be propagated into the status of the custom resource.

**Describe alternatives you've considered**
The only other alternative is changing the format of the CR to be clearer the password is a key, i.e.
```
    passwordSecret:
     # key: my-password
     or
      #passwordKey: my-password
      secretName: my-secret
```
I think documentation and examples already exist for this that is more than sufficient, but sometimes a user may still make these mistakes

samuel-hawker:
Up for discussion how it would be implemented, but I think it best belongs in the `status.conditions` as an error or warning.

ppatierno:
> Similarly, there has been some confusion around certs. Where a user has been using a KafkaUser cert in spec.tls.trustedCertificates where the KafkaConnect spec was expecting a cluster ca cert
It wasn't at all obvious to the user why his connect cluster was broken.

Tbh maybe it's just a lack of knowledge on how TLS works for people who make the mistake.
When you say "trusted certificates store" it should be clear that it's something you use for trusting what you are getting from the remote peer and you have to verify and trust. 
In a TLS handshake, you always get a server certificate and you can trust it:

* having the certificate used for signing the server certificate
* having the server certificate itself

The above has to be in the trust store.
Why using a KafkaUser certificate when it's for client authentication and why they should think that it's a "trusted" cert? 
[...]

PR #5983, opened 2021-12-03
validating password key of a secret for auth scram and plaintext. (#3475)

fjbecerra:
Signed-off-by: Francisco Becerra <fran.bcra@gmail.com>

### Type of change

Enhancement 

### Description

Validates a password key of a secret for authentication Scram and Plaintext while reconciling.  It throws a meaningful error message if it's given a wrong password key, mentioning both the secret and the password key passed. 

### Checklist

- [ ] Write tests
- [ ] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/operator-common/src/main/java/io/strimzi/operator/common/Util.java
+++ b/operator-common/src/main/java/io/strimzi/operator/common/Util.java
@@ -666,13 +666,31 @@ public class Util {
     private static Future<String> getPasswordAsync(SecretOperator secretOperator, String namespace, KafkaClientAuthentication auth) {
         if (auth instanceof KafkaClientAuthenticationPlain) {
             return secretOperator.getAsync(namespace, ((KafkaClientAuthenticationPlain) auth).getPasswordSecret().getSecretName())
-                    .compose(secret -> secret == null ? Future.failedFuture("Secret " + ((KafkaClientAuthenticationPlain) auth).getPasswordSecret().getSecretName() + " not found") :
-                            Future.succeededFuture(secret.getData().get(((KafkaClientAuthenticationPlain) auth).getPasswordSecret().getPassword())));
+            .compose(secret -> {
+                if (secret == null) {
+                    return Future.failedFuture("Secret " + ((KafkaClientAuthenticationPlain) auth).getPasswordSecret().getSecretName() + " not found");
+                } else {
+                    String passwordKey = ((KafkaClientAuthenticationPlain) auth).getPasswordSecret().getPassword();
+                    if (!secret.getData().containsKey(passwordKey)) {
+                        return Future.failedFuture(String.format("Secret %s does not contain key %s", ((KafkaClientAuthenticationPlain) auth).getPasswordSecret().getSecretName(), passwordKey));
+                    }
+                    return Future.succeededFuture(secret.getData().get(passwordKey));
+                }   
+            });
         }
         if (auth instanceof KafkaClientAuthenticationScramSha512) {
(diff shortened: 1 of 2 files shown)


operator-common/src/main/java/io/strimzi/operator/common/Util.java
```diff
@@ -666,13 +666,31 @@ public static void delete(Path key) {
     private static Future<String> getPasswordAsync(SecretOperator secretOperator, String namespace, KafkaClientAuthentication auth) {
         if (auth instanceof KafkaClientAuthenticationPlain) {
             return secretOperator.getAsync(namespace, ((KafkaClientAuthenticationPlain) auth).getPasswordSecret().getSecretName())
-                    .compose(secret -> secret == null ? Future.failedFuture("Secret " + ((KafkaClientAuthenticationPlain) auth).getPasswordSecret().getSecretName() + " not found") :
-                            Future.succeededFuture(secret.getData().get(((KafkaClientAuthenticationPlain) auth).getPasswordSecret().getPassword())));
+            .compose(secret -> {
+                if (secret == null) {
+                    return Future.failedFuture("Secret " + ((KafkaClientAuthenticationPlain) auth).getPasswordSecret().getSecretName() + " not found");
+                } else {
+                    String password = ((KafkaClientAuthenticationPlain) auth).getPasswordSecret().getPassword();
```
scholzj, line 673: Maybe calling the variable `passwordKey` will make it more readable -> it will be more clear that it is just the key to it and not the password itself.

operator-common/src/main/java/io/strimzi/operator/common/Util.java
```diff
@@ -666,13 +666,31 @@ public static void delete(Path key) {
...
         }
         if (auth instanceof KafkaClientAuthenticationScramSha512) {
             return secretOperator.getAsync(namespace, ((KafkaClientAuthenticationScramSha512) auth).getPasswordSecret().getSecretName())
-                    .compose(secret -> secret == null ? Future.failedFuture("Secret " + ((KafkaClientAuthenticationScramSha512) auth).getPasswordSecret().getSecretName() + " not found") :
-                            Future.succeededFuture(secret.getData().get(((KafkaClientAuthenticationScramSha512) auth).getPasswordSecret().getPassword())));
+            .compose(secret -> {
+                if (secret == null) {
+                    return Future.failedFuture("Secret " + ((KafkaClientAuthenticationScramSha512) auth).getPasswordSecret().getSecretName() + " not found");
+                } else {
+                    String password = ((KafkaClientAuthenticationScramSha512) auth).getPasswordSecret().getPassword();
```
scholzj, line 687: Same as above ... Maybe calling the variable `passwordKey` will make it more readable -> it will be more clear that it is just the key to it and not the password itself.

scholzj (commented):
Thanks for the PR. I left two nits which can IMHO make the code ore readable. But otherwise it looks great.

ppatierno (approved):
Thanks for the PR! I approved but I agree with Jakub comments, so I would rename the `password` var to `passwordKey` as well before merging.

operator-common/src/main/java/io/strimzi/operator/common/Util.java
fjbecerra: yes, good catch, updated!

operator-common/src/main/java/io/strimzi/operator/common/Util.java
fjbecerra: updated

scholzj (approved):
Thanks for the PR.

sknot-rh (approved):

PR #5986, opened 2021-12-03
Improve the build and release process

scholzj:
### Type of change

- Refactoring

### Description

This PR improves the build and release process of the Strimzi operators. It follows the ideas discussed in the past during the community calls. The main targets are:
* Add new release pipelines to facilitate better release process for release candidates
* Make it easier to do CVE re-spins
* Make it easier to do multi-arch builds

This PR currently provides a better separation of the release artifacts so that they can be used in the build and release pipelines. It separates the Java build from the container builds and uses the artifacts as the point which connects them.


scholzj:
Closing as it needs to be opened from internal branch to add new pipelines 🙄  🤦‍♂️

PR #5988, opened 2021-12-04
Generate Kafka configuration models through Maven and not through Make

scholzj:
### Type of change

- Refactoring

### Description

The `config-model-generator` Maven module generates the Kafka configuration models for the different Kafka versions supported by Strimzi. These models are used by the Cluster Operator when changing and validating Kafka configurations etc.

Currently, it is only the `make` build which generates the models. It runs a shell script and the shell script calls the Maven build to generate the model. This is because to generate the model for different Kafka versions, the Maven build needs to be called multiple times with different Kafka versions as dependencies. When you run only the Maven build for the whole Strimzi project, the models are not generated and the operators might not work -> only the Make build does that.

This PR changes how this is called. It uses Maven profiles to generate the model. One profile - the default one - calls the shell script which calls the Maven again multiple times to generate models for the different versions. Thanks to that, if you now do something like `mvn install` on the top level (or in the `config-model-generator` module itself), it will now generate the models. Of course, as with other modules, the Make build calls the Maven build as well. So when you use the Make build instead of the Maven build, the models will be still generated as before. This should help to build the project correctly even in situations when you might not want to use Make build for example to avoid building container images etc.

To make this work, the `config-model` needs to be installed. I do not think that should be an issue in general, but I needed to change the CodeQL build to use `mvn install` instead of `mvn package`.


sknot-rh (approved):

scholzj:
@sknot-rh @tombentley I reworked this a bit in a way which I think is even better -> I merged the `config-model` and `config-model-generator` modules. So now even commands like `mvn verify` etc. work. This allows us to simplify the build pipelines. Please have another look. Thanks.

sknot-rh (approved):

config-model/pom.xml
```diff
@@ -11,14 +11,61 @@
 
     <artifactId>config-model</artifactId>
 
+    <properties>
+        <kafka-metadata-version>${kafka.version}</kafka-metadata-version>
+        <config-model-file>../cluster-operator/src/main/resources/kafka-${kafka-metadata-version}-config-model.json</config-model-file>
+    </properties>
+
+    <dependencyManagement>
+        <!-- Manage versions to allow for override via the CLI with system property. -->
```
tombentley, line 20: Can you add to the comment the why we need to be able to override with system props?

config-model/src/main/java/io/strimzi/kafka/config/model/generator/KafkaConfigModelGenerator.java
```diff
@@ -2,7 +2,7 @@
  * Copyright Strimzi authors.
  * License: Apache License 2.0 (see the file LICENSE or http://apache.org/licenses/LICENSE-2.0.html).
  */
-package io.strimzi.build.kafka.metadata;
+package io.strimzi.kafka.config.model.generator;
```
tombentley, line 5: This means that the generator will be present in the classpath at runtime, which wasn't the case before. I think this means that kafka itself appears to be a (transitive) runtime dependency of the operator. I think we want to avoid that, if possible, because it means someone will end up using that dependency in operator code. We don't want that kind of coupling.

scholzj: Sorry not sure I follow exactly what do you mean with this. Can you elaborate a bit more on this and explain ehat issues do you expect from it?

scholzj: I mean, yes, this moves the generator to runtime. The rest is not completely clear to me.

tombentley: Previously Kafka was not a dependency of the operator. Changing that allows operator code to make use of Kafka APIs, and it would be easy to miss in code review. Making use of Kafka APIs risks operator behaviour becoming coupled to some particular _version_ of Kafka (which then could mean weird incompatiblities in the operator, depending on when the dependency was upgraded). It would also make it easier for the operator to depend on internal Kafka APIs (again, easily missed in code review). Avoiding these kinds of things was the reason we came up with the config model in the first place. Otherwise we could have just had the operator depend directly on Kafka's `KafkaConfig` and `Validator` classes, which would be much simpler, but prone so subtle bugs or compatibility problems if those changed between Kafka versions.

scholzj: Ok, understand now. But, well, the previous model was broken - I would argue that it was broken way more than introducing the Kafka dependency. 

So what other alternatives do you see? The previous iteration of this had the modules separate, but that requires you to always to `mvn install` instead of `mvn test` or `mvn verify` for example. I can probably live with it and it is better than the current state. But it was not perfect either.

tombentley: I think keeping the modules separate would be preferable. I hear your build pain though. I think the right way to solve that would be to factor the generator into something (maybe a custom maven build plugin?) which the operator build could simply depend on (not have to build). But that's pretty costly, since it would require a release and push to maven central. Or maybe scripting.

scholzj: Ok, well, I pushed the alternative implementation with separate modules then. Please have a look.

tombentley (approved):
LGTM, but one thought: Would it work to `mvn install` the model generator in the shell script?

scholzj:
> LGTM, but one thought: Would it work to mvn install the model generator in the shell script?

In theory, yes. But practically, I think you would need to also install the parent project etc. So it would be more complicated and not sure how well would it work.

PR #5991, opened 2021-12-05
Improve build and release system

scholzj:
### Type of change

- Refactoring

### Description

This PR improves the build and release process of the Strimzi operators. It follows the ideas discussed in the past during the community calls and similar pipeline setups for Drain Cleaner and Kafka Bridge. The main targets are:
* Add new release pipelines to facilitate better release process for release candidates
  * Release artifacts (docs, release YAMLs, Helm Chart ZIP) should be produced by the pipeline
  * The same build / containers should be used for RCs as for GAs
* Make it easier to do CVE re-spins
  * Take the released binaries and just rebuild the containers
* Add support for multi-arch builds

To provide this, our build had to be refactored to make it easier to separate the binaries used to build the container images. They are now collected in one place (`./docker-images/artifacts`) and can be easily passed between pipelines. It also separates the `make` targets for Java and container builds to be able to build these separately with the artifacts being the shared point between them. This change should have only a minimal impact on the users building Strimzi locally.

The Azure pipelines were refactored. Apart form adding the new pipelines for release and CVE re-spins, I also refactored the different templates used by build and system test pipelines. The build pipeline has also been redesigned and consists from multiple jobs with different tasks:

Screenshot 2021-12-05 at 20 59 48

This PR also adds support for ARM64 / AArch64 images (-> it is part of this PR instead of separate PR since it allows to properly test the multi-arch build capability).


.azure/cve-pipeline.yaml
```diff
@@ -0,0 +1,89 @@
...
+    default: true
+  - name: releaseSuffix
+    displayName: Release Suffix
+    type: number
+  - name: sourcePipelineId
+    displayName: Pipeline ID of the source build
+    type: number
+    default: 16
+    values:
+      - 16
```
Frawless, line 22: just out of curiosity - why 16?

.azure/templates/jobs/build/push_containers.yaml
```diff
@@ -0,0 +1,64 @@
...
+          DOCKER_PASS: $(QUAY_PASS)
+          DOCKER_REGISTRY: "quay.io"
+      - bash: "make docker_delete_manifest"
+        displayName: "Delete existing container manifests"
+        env:
+          BUILD_REASON: $(Build.Reason)
+          BRANCH: $(Build.SourceBranch)
+          DOCKER_REGISTRY: "quay.io"
+          DOCKER_ORG: "strimzi"
+          DOCKER_TAG: '${{ parameters.dockerTag }}'
```
Frawless, line 46: wouldn't be better to set those env vars for the whole job? If it's possible of course, I am not sure now, but I think it should be

systemtest/pom.xml
```diff
@@ -230,15 +230,22 @@
     <profiles>
 
         <profile>
-            <id>all</id>
+            <id>none</id>
             <properties>
-                <skipTests>false</skipTests>
+                <skipITs>true</skipITs>
```
Frawless, line 235: shouldn't be there `-skipTests`?

scholzj: No, we want to run the UTs as part of the regular UTs. But the ITs will run the STs -> which is what we don't want in regular build. LEts hope we will never have any actual ITs in the systemtests module ;-).

.azure/cve-pipeline.yaml
scholzj: This is the ID of the pipeline the artifacts are taken from -> and that is in this case so far only the operators `build` pipeline which has ID 16. There might be more options in the future, but even if not, having it here helps to have it in one place.

.azure/templates/jobs/build/push_containers.yaml
scholzj: I'm not aware of any way to do that. If there is, I can do it. But it might be a bit confusing since some env-vars might be defined more centrally, others which to for example define secrets would be only locally for the steps which actually need them for example.

.azure/cve-pipeline.yaml
```diff
@@ -0,0 +1,89 @@
...
+        parameters:
+          dockerTag: '${{ parameters.releaseVersion }}-${{ parameters.releaseSuffix }}'
+          artifactSource: 'current'
+          artifactProject: 'strimzi'
+          artifactPipeline: ''
+          artifactRunVersion: ''
+          artifactRunId: ''
+          architectures: ['amd64', 'arm64']
+  - stage: manual_validation
+    displayName: Validate container before pushing container as ${{ parameters.releaseVersion }}
```
ppatierno, line 56: I don't remember we have this manual validation in Drain Cleaner and canary for example. Why do we need it here? Is that anything that we should think of the other repo as well?

.azure/cve-pipeline.yaml
Frawless: oh, I see, maybe we can add there comment with that explanation?

ppatierno (approved):
lgtm

.azure/templates/jobs/build/push_containers.yaml
Frawless: ye that makes sense...it should be possible with `variables` key on same level as `steps` but keep it as it is

Frawless (approved):

.azure/cve-pipeline.yaml
scholzj: I added the comments.

Issue #5951, opened 2021-11-26
CruiseControl CPU Utilization Computation


roland-haeusler:
**Describe the bug**
When trying to do a rebalance with Cruise Control, we always get an OptimizationFailureException saying the CpuCapacityGoal cannot be satisfied and we should add more brokers. The Utilization Computation seems off. The CruiseControl REST API Call to /kafkacruisecontrol/load shows that it assumes our brokers have 1 core, which may be the cause of the malcomputation.

In CruiseControl you can set a num.cores for broker capacity (https://github.com/linkedin/cruise-control/blob/migrate_to_kafka_2_4/config/capacityCores.json), but in Strimzi, this is not possible (https://strimzi.io/docs/operators/latest/using.html#type-CruiseControlSpec-reference)

**To Reproduce**
1. Add cruiseControl to Kafka CR cruiseControl: {}
2. Create KafkaRebalance with spec: {}
3. Wait to see Status:
`Error for request:
        cluster-main-cruise-control.kafka-devl.svc:9090/kafkacruisecontrol/rebalance?json=true&dryrun=true&verbose=true&skip_hard_goal_check=false.
        Server returned: Error processing POST request '/rebalance' due to:
        'com.linkedin.kafka.cruisecontrol.exception.OptimizationFailureException:
        [CpuCapacityGoal] Insufficient capacity for cpu (Utilization 858.79,
        Allowed Capacity 600.00, Threshold: 1.00). Add at least 3 brokers with
        the same cpu capacity (100.00) as broker-0. Add at least 3 brokers with
        the same cpu capacity (100.00) as broker-0.'.`

**Expected behavior**
Expect to see a "ProposalReady" Status in the KafkaRebalance CR, something like:
`Status:
  Conditions:
    Last Transition Time:  2020-05-19T13:50:12.533Z
    Status:                ProposalReady
    Type:                  State
  Observed Generation:     1
  Optimization Result:
    Data To Move MB:  0
    Excluded Brokers For Leadership:
    Excluded Brokers For Replica Move:
    Excluded Topics:
    Intra Broker Data To Move MB:         0
    Monitored Partitions Percentage:      100
    Num Intra Broker Replica Movements:   0
    Num Leader Movements:                 0
    Num Replica Movements:                26
    On Demand Balancedness Score After:   81.8666802863978
    On Demand Balancedness Score Before:  78.01176356230222
    Recent Windows:                       1
[...]

PR #5992, opened 2021-12-05
Improve CC documentation about CPU capacity and utilization

fvaleri:
This is to add a couple of important notes from the discussion in #5951.

### DIFF
--- a/documentation/modules/cruise-control/proc-generating-optimization-proposals.adoc
+++ b/documentation/modules/cruise-control/proc-generating-optimization-proposals.adoc
@@ -143,7 +143,7 @@ com.linkedin.kafka.cruisecontrol.exception.OptimizationFailureException:
 
 [NOTE]
 ====
-The error shows CPU capacity as a percentage rather than CPU cores. For that reason, it does not directly map to the number of CPUs configured in `Kafka.spec.kafka.resources`. It is like having a single _virtual_ CPU per broker, which has the cycles of one or more CPUs. This has no effect on the rebalance behavior, since the ratio between CPU utilization and capacity remains the same.
+The error shows CPU capacity as a percentage rather than CPU cores. For that reason, it does not directly map to the number of CPUs configured in Kafka CR. It is like having a single _virtual_ CPU per broker, which has the cycles of `Kafka.spec.kafka.resources.limits.cpu` CPUs. This has no effect on the rebalance behavior, since the ratio between CPU utilization and capacity remains the same.
 ====
 
 .What to do next
--- a/documentation/modules/cruise-control/proc-generating-optimization-proposals.adoc
+++ b/documentation/modules/cruise-control/proc-generating-optimization-proposals.adoc
@@ -129,9 +129,9 @@ For descriptions of each property, see xref:contents-optimization-proposals[Cont
 
 .Insufficient CPU capacity
 
-If a Kafka cluster is overloaded in terms of CPU utilization, you might see an insufficient CPU capacity error in the `KafkaRebalance` status. 
-This might be because of excluded topics configuration (`excludedTopics`). 
-Optimization proposals ignore excluded topics, but CPU utilization calculations consider their replicas.
+If a Kafka cluster is overloaded in terms of CPU utilization, you might see an insufficient CPU capacity error in the `KafkaRebalance` status.
+It's worth noting that this utilization value is unaffected by the `excludedTopics` configuration. 
+Although optimization proposals will not reassign replicas of excluded topics, their load is still considered in the utilization calculation.
 
 .Example CPU utilization error
 [source,shell,subs="+quotes"]
--- a/documentation/modules/cruise-control/proc-generating-optimization-proposals.adoc
+++ b/documentation/modules/cruise-control/proc-generating-optimization-proposals.adoc
@@ -129,7 +129,9 @@ For descriptions of each property, see xref:contents-optimization-proposals[Cont
 
 .Insufficient CPU capacity
 
-If a Kafka cluster is overloaded in terms of CPU utilization, you might see an insufficient CPU capacity error in the `KafkaRebalance` status. This might be because of excluded topics configuration (`excludedTopics`). Optimization proposals ignore excluded topics, but CPU utilization calculations consider their replicas.
+If a Kafka cluster is overloaded in terms of CPU utilization, you might see an insufficient CPU capacity error in the `KafkaRebalance` status. 
+This might be because of excluded topics configuration (`excludedTopics`). 
+Optimization proposals ignore excluded topics, but CPU utilization calculations consider their replicas.
 
 .Example CPU utilization error
 [source,shell,subs="+quotes"]
@@ -143,7 +145,9 @@ com.linkedin.kafka.cruisecontrol.exception.OptimizationFailureException:
 
 [NOTE]
 ====
-The error shows CPU capacity as a percentage rather than CPU cores. For that reason, it does not directly map to the number of CPUs configured in Kafka CR. It is like having a single _virtual_ CPU per broker, which has the cycles of `Kafka.spec.kafka.resources.limits.cpu` CPUs. This has no effect on the rebalance behavior, since the ratio between CPU utilization and capacity remains the same.
+The error shows CPU capacity as a percentage rather than CPU cores. For that reason, it does not directly map to the number of CPUs configured in Kafka CR.
+It is like having a single _virtual_ CPU per broker, which has the cycles of `Kafka.spec.kafka.resources.limits.cpu` CPUs.
+This has no effect on the rebalance behavior, since the ratio between CPU utilization and capacity remains the same.
 ====
 
 .What to do next
--- a/documentation/modules/cruise-control/proc-generating-optimization-proposals.adoc
+++ b/documentation/modules/cruise-control/proc-generating-optimization-proposals.adoc
@@ -127,20 +127,23 @@ Status:
 The properties in the `Optimization Result` section describe the pending cluster rebalance operation. 
 For descriptions of each property, see xref:contents-optimization-proposals[Contents of optimization proposals].
 
-.Insufficient capacity for cpu
+.Insufficient CPU capacity
 
-If the Kafka cluster is overloaded in terms of CPU utilization and needs some additional brokers, you may see an error like the following in the `KafkaRebalance` status. This may happen because, while proposals exclude the `excludedTopics`, the CPU utilization still considers replicas of the `excludedTopics`.
+If a Kafka cluster is overloaded in terms of CPU utilization, you might see an insufficient CPU capacity error in the `KafkaRebalance` status. This might be because of excluded topics configuration (`excludedTopics`). Optimization proposals ignore excluded topics, but CPU utilization calculations consider their replicas.
 
+.Example CPU utilization error
 [source,shell,subs="+quotes"]
+----
 com.linkedin.kafka.cruisecontrol.exception.OptimizationFailureException:
         [CpuCapacityGoal] Insufficient capacity for cpu (Utilization 615.21,
         Allowed Capacity 420.00, Threshold: 0.70). Add at least 3 brokers with
         the same cpu capacity (100.00) as broker-0. Add at least 3 brokers with
         the same cpu capacity (100.00) as broker-0.
+----
 
 [NOTE]
 ====
-These numbers do not make sense if you look at `Kafka.spec.kafka.resources` because Strimzi generates a CPU percentage Cruise Control configuration, instead of a CPU core configuration. It is like having a single "virtual" CPU which has the cycles of 1 or more CPUs (the number configured in `Kafka.spec.kafka.resources.limits.cpu`). This is counterintuitive, but it has no effect on the rebalance behavior since the ratio between the utilization and capacity remains the same.
+The error shows CPU capacity as a percentage rather than CPU cores. For that reason, it does not directly map to the number of CPUs configured in `Kafka.spec.kafka.resources`. It is like having a single _virtual_ CPU per broker, which has the cycles of one or more CPUs. This has no effect on the rebalance behavior, since the ratio between CPU utilization and capacity remains the same.
 ====
 
 .What to do next
--- a/documentation/modules/cruise-control/proc-generating-optimization-proposals.adoc
+++ b/documentation/modules/cruise-control/proc-generating-optimization-proposals.adoc
@@ -125,7 +125,23 @@ Status:
 ----
 +
 The properties in the `Optimization Result` section describe the pending cluster rebalance operation. 
-For descriptions of each property, see xref:contents-optimization-proposals[Contents of optimization proposals]. 
+For descriptions of each property, see xref:contents-optimization-proposals[Contents of optimization proposals].
+
+.Insufficient capacity for cpu
+
+If the Kafka cluster is overloaded in terms of CPU utilization and needs some additional brokers, you may see an error like the following in the `KafkaRebalance` status. This may happen because, while proposals exclude the `excludedTopics`, the CPU utilization still considers replicas of the `excludedTopics`.
+
+[source,shell,subs="+quotes"]
+com.linkedin.kafka.cruisecontrol.exception.OptimizationFailureException:
+        [CpuCapacityGoal] Insufficient capacity for cpu (Utilization 615.21,
+        Allowed Capacity 420.00, Threshold: 0.70). Add at least 3 brokers with
+        the same cpu capacity (100.00) as broker-0. Add at least 3 brokers with
+        the same cpu capacity (100.00) as broker-0.
+
+[NOTE]
+====
+These numbers do not make sense if you look at `Kafka.spec.kafka.resources` because Strimzi generates a CPU percentage Cruise Control configuration, instead of a CPU core configuration. It is like having a single "virtual" CPU which has the cycles of 1 or more CPUs (the number configured in `Kafka.spec.kafka.resources.limits.cpu`). This is counterintuitive, but it has no effect on the rebalance behavior since the ratio between the utilization and capacity remains the same.
+====
 
 .What to do next
 
--- a/documentation/modules/cruise-control/ref-cruise-control-configuration.adoc
+++ b/documentation/modules/cruise-control/ref-cruise-control-configuration.adoc
@@ -94,6 +94,11 @@ Capacity limits can be set for the following broker resources, using the standar
 
 Because Strimzi Kafka brokers are homogeneous, Cruise Control applies the same capacity limits to every broker it is monitoring.
 
+[NOTE]
+====
+Cruise Control assumes brokers resources to be homogeneous. Therefore, it is recommended to set both `Kafka.spec.kafka.resources.limits.cpu` and `Kafka.spec.kafka.resources.limits.memory`, so that Kubernetes will ensure broker pods are limited to the same capacity. This is especially important when using nodes with heterogeneous resources (e.g. CPU cores).
+====
+
 .An example Cruise Control brokerCapacity configuration using bibyte units
 [source,yaml,subs="attributes+"]
 ----
@@ -103,6 +108,12 @@ metadata:
   name: my-cluster
 spec:
   # ...
+  kafka:
+    # ...
+    resources:
+      limits:
+        cpu: 1000m
+        memory: 1Gi
   cruiseControl:
     # ...
     brokerCapacity:
--- a/documentation/modules/cruise-control/ref-cruise-control-configuration.adoc
+++ b/documentation/modules/cruise-control/ref-cruise-control-configuration.adoc
@@ -96,7 +96,7 @@ Because Strimzi Kafka brokers are homogeneous, Cruise Control applies the same c
 
 [NOTE]
 ====
-Cruise Control assumes brokers resources to be homogeneous. Therefore, it is recommended to set both `Kafka.spec.kafka.resources.limits.cpu` and `Kafka.spec.kafka.resources.limits.memory`, so that Kubernetes will ensure broker pods are limited to the same capacity. This is especially important when using nodes with heterogeneous resources (e.g. CPU cores).
+Cruise Control assumes brokers resources to be homogeneous. Therefore, it is recommended to set value in `Kafka.spec.kafka.resources.requests` equal to `Kafka.spec.kafka.resources.limits`, so that Kubernetes will ensure broker pods have the same capacity and Guaranteed QoS, which improves stability. This is especially important when using nodes with heterogeneous resources (e.g. CPU cores).
 ====
 
 .An example Cruise Control brokerCapacity configuration using bibyte units
@@ -108,12 +108,6 @@ metadata:
   name: my-cluster
 spec:
   # ...
-  kafka:
-    # ...
-    resources:
-      limits:
-        cpu: 1000m
-        memory: 1Gi
   cruiseControl:
     # ...
     brokerCapacity:
--- a/documentation/modules/cruise-control/ref-cruise-control-configuration.adoc
+++ b/documentation/modules/cruise-control/ref-cruise-control-configuration.adoc
@@ -96,7 +96,7 @@ Because Strimzi Kafka brokers are homogeneous, Cruise Control applies the same c
 
 [NOTE]
 ====
-Cruise Control assumes brokers resources to be homogeneous. Therefore, it is recommended to set value in `Kafka.spec.kafka.resources.requests` equal to `Kafka.spec.kafka.resources.limits`, so that Kubernetes will ensure broker pods have the same capacity and Guaranteed QoS, which improves stability. This is especially important when using nodes with heterogeneous resources (e.g. CPU cores).
+Cruise Control assumes brokers resources to be homogeneous. Therefore, it is recommended to set value in `Kafka.spec.kafka.resources.requests` equal to `Kafka.spec.kafka.resources.limits`, so that Kubernetes will ensure broker pods have the same capacity and Guaranteed QoS, which also improves stability. This is especially important when using nodes with heterogeneous resources (CPU cores for example).
 ====
 
 .An example Cruise Control brokerCapacity configuration using bibyte units
--- a/documentation/modules/cruise-control/ref-cruise-control-configuration.adoc
+++ b/documentation/modules/cruise-control/ref-cruise-control-configuration.adoc
@@ -96,7 +96,7 @@ Because Strimzi Kafka brokers are homogeneous, Cruise Control applies the same c
 
 [NOTE]
 ====
-Cruise Control assumes brokers resources to be homogeneous. Therefore, it is recommended to set value in `Kafka.spec.kafka.resources.requests` equal to `Kafka.spec.kafka.resources.limits`, so that Kubernetes will ensure broker pods have the same capacity and Guaranteed QoS, which also improves stability. This is especially important when using nodes with heterogeneous resources (CPU cores for example).
+We recommend that `Kafka.spec.kafka.resources.requests` and `Kafka.spec.kafka.resources.limits` are set to equal values. With these settings, broker pods will have the same capacity and quality of service is guaranteed. The settings can improve stability. This is especially important when using nodes with heterogeneous resources, such as CPU cores.
 ====
 
 .An example Cruise Control brokerCapacity configuration using bibyte units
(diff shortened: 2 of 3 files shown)


fvaleri:
cc @kyguy

packaging/examples/cruise-control/kafka-cruise-control.yaml
```diff
@@ -23,6 +23,10 @@ spec:
       min.insync.replicas: 2
       log.message.format.version: "3.0"
       inter.broker.protocol.version: "3.0"
+    resources:
+      limits:
+        cpu: 1000m
+        memory: 1Gi
```
scholzj, line 29: Why is this being added? I don't think we want any specific values in the example resources as they don't necessarily work.

documentation/modules/cruise-control/ref-cruise-control-configuration.adoc
```diff
@@ -103,6 +108,12 @@ metadata:
   name: my-cluster
 spec:
   # ...
+  kafka:
+    # ...
+    resources:
+      limits:
```
scholzj, line 114: Should this be request rather than limit?

documentation/modules/cruise-control/ref-cruise-control-configuration.adoc
```diff
@@ -94,6 +94,11 @@ Capacity limits can be set for the following broker resources, using the standar
 
 Because Strimzi Kafka brokers are homogeneous, Cruise Control applies the same capacity limits to every broker it is monitoring.
 
+[NOTE]
+====
+Cruise Control assumes brokers resources to be homogeneous. Therefore, it is recommended to set both `Kafka.spec.kafka.resources.limits.cpu` and `Kafka.spec.kafka.resources.limits.memory`, so that Kubernetes will ensure broker pods are limited to the same capacity. This is especially important when using nodes with heterogeneous resources (e.g. CPU cores).
```
scholzj, line 99: We for sure do not recommend setting memory limits much -> it is not guaranteed and can cause problems when the memory needs to be returned.

packaging/examples/cruise-control/kafka-cruise-control.yaml
fvaleri: We recommend to set a limit, but our own example has no limit, this is in contrast. What about adding a comment on top of resources which says "set your own values here"?

documentation/modules/cruise-control/ref-cruise-control-configuration.adoc
fvaleri: Yeah, I mean, the full recommendation would be to set requests == limits. Should I add this point too?

packaging/examples/cruise-control/kafka-cruise-control.yaml
scholzj: None of our examples have any requests or limits, because it causes problems and misunderstandings: you do not know which setting is the right for the user and which actually works in his environment. And you also do not want to suggest to anyone that `cpu: 1000m` and `memory: 1Gi` is what you recommend them to use.

documentation/modules/cruise-control/ref-cruise-control-configuration.adoc
fvaleri: This sets requests == limits.

scholzj: Yes, but it is not clear to everyone. Setting only request also means that it will be the limit but it is much more clear what it means to everyone.

packaging/examples/cruise-control/kafka-cruise-control.yaml
fvaleri: Ok, agree. Will remove it.

documentation/modules/cruise-control/ref-cruise-control-configuration.adoc
fvaleri: Ok.

fvaleri:
Did some changes based on your feedback. Thanks.

scholzj:
PS: It seems the build is failing because it does not like the `e.g.` - `for example` should be used instead.

documentation/modules/cruise-control/proc-generating-optimization-proposals.adoc
```diff
@@ -125,7 +125,23 @@ Status:
 ----
 +
 The properties in the `Optimization Result` section describe the pending cluster rebalance operation. 
-For descriptions of each property, see xref:contents-optimization-proposals[Contents of optimization proposals]. 
+For descriptions of each property, see xref:contents-optimization-proposals[Contents of optimization proposals].
+
+.Insufficient capacity for cpu
```
PaulRMellor, line 130:
```suggestion
.Insufficient CPU capacity
```

fvaleri: Yeah, that was actually taken from the error message. But I'll do the change.

documentation/modules/cruise-control/proc-generating-optimization-proposals.adoc
```diff
@@ -125,7 +125,23 @@ Status:
...
+[source,shell,subs="+quotes"]
+com.linkedin.kafka.cruisecontrol.exception.OptimizationFailureException:
+        [CpuCapacityGoal] Insufficient capacity for cpu (Utilization 615.21,
+        Allowed Capacity 420.00, Threshold: 0.70). Add at least 3 brokers with
+        the same cpu capacity (100.00) as broker-0. Add at least 3 brokers with
+        the same cpu capacity (100.00) as broker-0.
+
+[NOTE]
+====
+These numbers do not make sense if you look at `Kafka.spec.kafka.resources` because Strimzi generates a CPU percentage Cruise Control configuration, instead of a CPU core configuration. It is like having a single "virtual" CPU which has the cycles of 1 or more CPUs (the number configured in `Kafka.spec.kafka.resources.limits.cpu`). This is counterintuitive, but it has no effect on the rebalance behavior since the ratio between the utilization and capacity remains the same.
```
PaulRMellor, line 143:
```suggestion
The error suggests adding brokers with CPU capacity shown as a percentage rather than CPU cores. CPU `resources` are configured as integers representing cores in Strimzi. You can ignore the percentages. This has no effect on the rebalance behavior, since the ratio between CPU utilization and capacity remains the same.
```

PaulRMellor: I wasn't sure what were trying to say in this note. I found it difficult to relate to the error message. I've made a suggestion, assuming the main point is that we can ignore the numbers for capacity shown in the error.

documentation/modules/cruise-control/ref-cruise-control-configuration.adoc
```diff
@@ -94,6 +94,11 @@ Capacity limits can be set for the following broker resources, using the standar
 
 Because Strimzi Kafka brokers are homogeneous, Cruise Control applies the same capacity limits to every broker it is monitoring.
 
+[NOTE]
+====
+Cruise Control assumes brokers resources to be homogeneous. Therefore, it is recommended to set value in `Kafka.spec.kafka.resources.requests` equal to `Kafka.spec.kafka.resources.limits`, so that Kubernetes will ensure broker pods have the same capacity and Guaranteed QoS, which also improves stability. This is especially important when using nodes with heterogeneous resources (CPU cores for example).
```
PaulRMellor, line 99:
```suggestion
We recommend that `Kafka.spec.kafka.resources.requests` and `Kafka.spec.kafka.resources.limits` are set to equal values. With these settings, broker pods will have the same capacity and quality of service. The settings can improve stability. This is especially important when using nodes with heterogeneous resources, such as CPU cores.
```

PaulRMellor: Don't think we need the homogeneous line,as it's written in the sentence before the note.

PaulRMellor (commented):
Hi Federico. Nice addition. I've made some suggestions. I found some of the sentences hard to read, so I've tried to simplify.

documentation/modules/cruise-control/proc-generating-optimization-proposals.adoc
```diff
@@ -125,7 +125,23 @@ Status:
 ----
 +
 The properties in the `Optimization Result` section describe the pending cluster rebalance operation. 
-For descriptions of each property, see xref:contents-optimization-proposals[Contents of optimization proposals]. 
+For descriptions of each property, see xref:contents-optimization-proposals[Contents of optimization proposals].
+
+.Insufficient CPU capacity
+
+If the Kafka cluster is overloaded in terms of CPU utilization and needs some additional brokers, you may see an error like the following in the `KafkaRebalance` status. This may happen because, while proposals exclude the `excludedTopics`, the CPU utilization still considers replicas of the `excludedTopics`.
```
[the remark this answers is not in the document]
fvaleri: It reads better, but it is not quite right.

documentation/modules/cruise-control/ref-cruise-control-configuration.adoc
kyguy: "Cruise Control" -> "Strimzi Cruise Control"

It is recommended to set `Kafka.spec.kafka.resources.limits.cpu` so Kubernetes will ensure broker pods have the same CPU capacity, otherwise the CPU utilization for brokers can get under or overestimated when working with nodes with heterogeneous CPU resources. 

I need to think more about the setting of `Kafka.spec.kafka.resources.limits.cpu`. From what I understand, Cruise Control doesn't have any explicit protections against heterogeneous memory capacities but memory is considered when calculating broker load by using certain metrics (e.g. by using metrics like bytes in and bytes out in its CPU and Network utilization calculations). Values like bytes in and bytes out could differ between brokers with different memory capacities. I am not sure how big of an impact it would have or if it is covered by the other metrics but I would think it would be a good idea to set `Kafka.spec.kafka.resources.limits.cpu` (except for the caveats that Jakub mentioned!)

documentation/modules/cruise-control/proc-generating-optimization-proposals.adoc
fvaleri: It reads better, but it is not quite right.

documentation/modules/cruise-control/ref-cruise-control-configuration.adoc
fvaleri: Perfect. Thanks.

fvaleri:
@PaulRMellor Thanks for your suggestions. 

I rephrased a bit some of them, because the meaning was not exactly the same after your changes. I hope it still reads well.

kyguy (commented):
Hey @fvaleri thanks for the PR!

I left a couple of notes, let me know what you think!

One other thing, from what I remember we enforce having one sentence per line (to make it easier to leave feedback in the PR review phase)

documentation/modules/cruise-control/proc-generating-optimization-proposals.adoc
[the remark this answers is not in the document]
fvaleri: The paragraph describes under which conditions you may see this error, so that you can verify if it matches with your setup. Instead, the note specifically addresses any doubts you may have looking at numbers. I would prefer the note to focus only on that aspect, as we know that it can be misleading.

fvaleri:
Thanks @PaulRMellor and @kyguy for your feedback. I did some changes based on that.

documentation/modules/cruise-control/proc-generating-optimization-proposals.adoc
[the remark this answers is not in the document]
kyguy: I felt that the second sentence starting with "This may happen because" makes it sound as if `excludedTopics` could be the cause of the brokers being overloaded. From what I understand, `excludedTopics` does not affect the load on the brokers, it affects what partitions are allowed to be moved?

I thought the mention of `excludeTopics` was to warn users who are expecting `excludeTopics` to affect the load that `excludeTopics` has no affect on the load. Was that the intention or am I misunderstanding something?

documentation/modules/cruise-control/proc-generating-optimization-proposals.adoc
```diff
@@ -125,7 +125,26 @@ Status:
...
+com.linkedin.kafka.cruisecontrol.exception.OptimizationFailureException:
+        [CpuCapacityGoal] Insufficient capacity for cpu (Utilization 615.21,
+        Allowed Capacity 420.00, Threshold: 0.70). Add at least 3 brokers with
+        the same cpu capacity (100.00) as broker-0. Add at least 3 brokers with
+        the same cpu capacity (100.00) as broker-0.
+----
+
+[NOTE]
+====
+The error shows CPU capacity as a percentage rather than CPU cores. For that reason, it does not directly map to the number of CPUs configured in `Kafka.spec.kafka.resources`. It is like having a single "virtual" CPU per broker, which has the cycles of one or more CPUs. This has no effect on the rebalance behavior, since the ratio between CPU utilization and capacity remains the same.
```
PaulRMellor, line 146:
```suggestion
The error shows CPU capacity as a percentage rather than CPU cores. For that reason, it does not directly map to the number of CPUs configured in `Kafka.spec.kafka.resources`. It is like having a single _virtual_ CPU per broker, which has the cycles of one or more CPUs. This has no effect on the rebalance behavior, since the ratio between CPU utilization and capacity remains the same.
```

documentation/modules/cruise-control/ref-cruise-control-configuration.adoc
```diff
@@ -94,6 +94,11 @@ Capacity limits can be set for the following broker resources, using the standar
 
 Because Strimzi Kafka brokers are homogeneous, Cruise Control applies the same capacity limits to every broker it is monitoring.
 
+[NOTE]
+====
+We recommend that `Kafka.spec.kafka.resources.requests` and `Kafka.spec.kafka.resources.limits` are set to equal values. With these settings, broker pods will have the same capacity and quality of service. The settings can improve stability. This is especially important when using nodes with heterogeneous resources, such as CPU cores.
```
PaulRMellor, line 99:
```suggestion
We recommend that `Kafka.spec.kafka.resources.requests` and `Kafka.spec.kafka.resources.limits` are set to equal values. With these settings, broker pods will have the same capacity and quality of service is guaranteed. The settings can improve stability. This is especially important when using nodes with heterogeneous resources, such as CPU cores.
```

PaulRMellor (approved):
:+1: Couple of small suggestions.

documentation/modules/cruise-control/ref-cruise-control-configuration.adoc
```diff
@@ -94,6 +94,11 @@ Capacity limits can be set for the following broker resources, using the standar
 
 Because Strimzi Kafka brokers are homogeneous, Cruise Control applies the same capacity limits to every broker it is monitoring.
 
+[NOTE]
+====
+We recommend that `Kafka.spec.kafka.resources.requests` and `Kafka.spec.kafka.resources.limits` are set to equal values. With these settings, broker pods will have the same capacity and quality of service is guaranteed. The settings can improve stability. This is especially important when using nodes with heterogeneous resources, such as CPU cores.
+====
+
```
scholzj, line 101: I don't think we recommend this. The only thing we recommend is to make sure that the Java memory of the Kafka brokers fits into the memory request.

fvaleri: OK let me rephrase it.

documentation/modules/cruise-control/ref-cruise-control-configuration.adoc
```diff
@@ -94,6 +94,11 @@ Capacity limits can be set for the following broker resources, using the standar
 
 Because Strimzi Kafka brokers are homogeneous, Cruise Control applies the same capacity limits to every broker it is monitoring.
 
+[NOTE]
+====
+We recommend that the Java memory of the Kafka brokers fits into the `Kafka.spec.kafka.resources.requests.memory`, so that brokers will have roughly the same capacity.
+====
+
```
scholzj, line 101: Sorry, maybe my comment should have been clearer last time as it could have been read as suggesting to fix the text. Apologies for that. This is more or less correct now, but how does it fit here? Nothing in here talks about memory management or JVM memory configuration. 

Also, in general, the fact that all the brokers have the same capacity is a limitation of Strimzi and not its goal. The test seems to suggest it is the end-goal. The main reason to have the Java memory fit into the memory request is also not because of capacity planning but to achieve stability and avoid running out of memory.

TBH, I would just remove this part and leave this part of the docs as it was.

fvaleri: Yeah, I think there was some confusion due to multiple edits. 

What I wanted to note here is that CC assumes same CPU/memory capacity on all brokers. It simply does not support per-broker values yet. In order to have a stable RebalanceProposal, which is still valid few minutes later when it is applied, it would be good to set requests == limits in Kafka CR, so that resources would be reserved upfront.

scholzj: @fvaleri Ok, I guess that got lost somewhere between the different review comments. I think it is ok to cover that in that case. But it should not be presented as recommendation because it is really a limitation. So maybe it could say something like `CPU goals in Cruise Control currently don't support CPU limits. For properly working CPU goals, you should always use the same CPU limit ad CPU request in the Kafka brokers configuration` (doesn't need to be exactly this of course). That makes it clear it is not a general recommendation but rather a limitation of Cruise Control. Also - I assume it is only the CPU goals which care about it - so it makes it clear that you can use CPU limits and just use other goals.

@kyguy @tomncooper Is it the case that Cruise Control doesn't handle CPU limits? If yes, what are the plans to fix it? It sounds like quite significant limitation to me.

kyguy: Cruise Control cannot accurately compare CPU utilization between broker pods on different nodes with heterogeneous CPU cores without either:
(1) Broker pod CPU limit being set to a homogenous value through K8s `LimitRange` [1] object or  `Kafka` resource limits
(2) Node specific CPU information in the `spec.cruiseControl.capacity` config in `Kafka` resource. 

Currently, Strimzi does not support (2) and therefore must rely on (1) for accurate CPU utilization comparison between brokers. I feel this follows Kubernetes principles of abstracting physical resources from the applications and users. Otherwise, we would need Strimzi or Cruise Control to be able to gather node information to generate (2) or have users manually track and update (2) whenever a broker pod was scheduled to a different node. 

[1] https://kubernetes.io/docs/concepts/policy/limit-range/

fvaleri: Thanks @kyguy. So this is actually a Strimzi limitation at the moment.

If you agree, I would phrase that note like this:

"Strimzi does not allow to configure per-broker CPU capacity. For that reason, Cruise Control cannot accurately compare CPU utilization between broker pods on different Kubernetes nodes. To overcome this, we suggest to set CPU limits equal to requests in `Kafka.spec.kafka.resources`. This would allow to reserve the required CPU resources upfront and generate accurate RebalanceProposals."

Then, there is also the workaround of excluding CPU goals from the preset hard/default goals, but the configuration is verbose and I don't think we should even mention it in the documentation.

scholzj: * You cannot phrase some Cruise Control specific limitation as general recommendation. It is not really recommendation, it is a workaround for some limitation. It needs to be phrased as such - so we do not suggest it to anyone, but it is a workaround people can use if they want.
* _Strimzi does not allow to configure per-broker CPU capacity._ I'm not sure how this is related. Per-broker CPU configuration has nothing to do with requests versus limits.
* As far as I understood, it affects Cruise Control only when CPU goals are used. That needs to be clear from it as well.
* _So this is actually a Strimzi limitation at the moment._ ... I do not see how this is a Strimzi limitation. The CPU limit is  something what needs to be understood by Cruise Control. They it can be either set in the CR or it can be detected by the CC metrics reporter. That is my understanding of the issue.

fvaleri: CC allows to set CPU capacity for each broker:
https://github.com/linkedin/cruise-control/blob/migrate_to_kafka_2_4/config/capacityCores.json

Strimzi does not allow to use this type of configuration, as you cannot specify a different CPU limit for each broker. We have the concept of a single "virtual" CPU core that has the cycles of `Kafka.spec.kafka.resources.limits.cpu` cores. 

If you don't set requests == limits. The real/current CPU capacity would be different when broker pods are running on different Kubernetes nodes. It may be due to other workloads scheduled on the same nodes or simply because they are over committed. This makes the RebalanceProposals unreliable with regards to CPU goals (enabled by default).

As an enhancement, we may think to detect the real/current CPU capacity for each broker at runtime (e.g. using metrics as you suggests) and reconcile the CC capacity configuration periodically.

Do we agree with that?

scholzj: But this is not about setting CPU cores per broker. Strimzi currently doesn't allow this anyway, although it will probably in the future. The issue is that when you set the limit, the actual CPU available can change every second. You configure the resources as request=4cpu and limit=8cpu. That means that in one second, it might have 8cpu, next it might have 6cpu and the next one it might have 4cpu. So being able to set the cpu per-broker in some Cruise Control settings will not help you in anyway because the actual usage might change every second.

fvaleri: The point is: if you set requests == limits in Kafka CR, then the CPU capacity would be guaranteed and rebalance proposal accurate. This is what I would like to recommend when using CC with CPU goals enabled.

scholzj: Yes, but as I said, there are many possible pros and cons of having requests == limits. So I don't think we should recommend it. It should be just described as workaround.

kyguy: >  It is not really recommendation, it is a workaround for some limitation. It needs to be phrased as such - so we do not suggest it to anyone, but it is a workaround people can use if they want.

Although I originally wanted this as a recommendation, I understand where Jakub is coming from here. The limit setting is only needed for CC to work in specific circumstances (it is not needed for Strimzi as a whole) so I see how characterizing the limit setting as a workaround would make sense.

> The CPU limit is something what needs to be understood by Cruise Control. They it can be either set in the CR or it can be detected by the CC metrics reporter. That is my understanding of the issue.

As discussed above, we cannot set it in the CR at the moment but I don't think we should allow it anyway because it would introduce a user maintenance burden. It would be nice if the CC metrics reporter detected and reported broker capacity and I am not sure why it doesn't  but I am sure there is a good reason. I'll find out why

> Is it the case that Cruise Control doesn't handle CPU limits? If yes, what are the plans to fix it? It sounds like quite significant limitation to me.

The advertised approach to solve this issue without limit setting is to write our own BrokerCapacityConfigResolver [1] I planned to investigate this anyway this to address related network I/O capacity issues, we can look into solving the CPU capacity issue at the same time. 

As for this PR, I am fine with stating limit setting as a workaround for accurate CPU balances when pods are running on nodes with heterogeneous CPU resources.

[1] https://github.com/linkedin/cruise-control/wiki/Pluggable-Components#broker-capacity-config-resolver

fvaleri: > As for this PR, I am fine with stating limit setting as a workaround for accurate CPU balances when pods are running on nodes with heterogeneous CPU resources.

@kyguy if you set `limits=3 < requests=1` in Kafka CR, then you are still not guaranteed to have accurate CPU balances, as each broker real CPU capacity would be somewhere between 1 and 3 cores. The configured limit of 3 cores could not be satisfied for long time, for example because the K8s node does not have capacity. 

Am I wrong?

kyguy: Sorry @fvaleri what I meant to say I am fine with your workaround of setting of `requests==limits`, you are correct!

fvaleri:
@kyguy @scholzj I think the last commit reflects our great discussion. I'm open to further improvements or clarifications.

documentation/modules/cruise-control/ref-cruise-control-configuration.adoc
```diff
@@ -94,6 +94,11 @@ Capacity limits can be set for the following broker resources, using the standar
 
 Because Strimzi Kafka brokers are homogeneous, Cruise Control applies the same capacity limits to every broker it is monitoring.
 
+[NOTE]
+====
+In order to guarantee accurate RebalanceProposals when using CPU goals, you can set requests equal to limits in Kafka CR. That way, all CPU resources are reserved upfront and the broker's full capacity is always available, even when the Kubernetes node is under pressure. This is a tradeoff, as it comes at the cost of some unused CPU cycles.
+====
+
```
scholzj, line 101:
```suggestion
[NOTE]
====
In order to guarantee accurate rebalance proposal when using CPU goals, you can set CPU requests equal to CPU limits in `spec.kafka.resources` in the Kafka CR. 
That way, all CPU resources are reserved upfront and are always available.
That allows Cruise Control to properly evaluate the CPU utilization when preparing the rebalance proposals based on CPU goals.
====

```

documentation/modules/cruise-control/proc-generating-optimization-proposals.adoc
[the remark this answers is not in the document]
kyguy: I am fine with leaving the mention of `excludeTopics` up here in the first paragraph to prevent user confusion but I think we should rephrase so something like: 

> It's worth noting that this utilization value is unaffected by the excluded topics configuration (`excludedTopics`). Although optimization proposals will not reassign replicas of excluded topics, their load is still considered in the utilization calculation.

Also we will want one sentence per line in this file as well!

scholzj (approved):

documentation/modules/cruise-control/proc-generating-optimization-proposals.adoc
```diff
@@ -125,7 +125,30 @@ Status:
 ----
 +
 The properties in the `Optimization Result` section describe the pending cluster rebalance operation. 
-For descriptions of each property, see xref:contents-optimization-proposals[Contents of optimization proposals]. 
+For descriptions of each property, see xref:contents-optimization-proposals[Contents of optimization proposals].
+
+.Insufficient CPU capacity
+
+If a Kafka cluster is overloaded in terms of CPU utilization, you might see an insufficient CPU capacity error in the `KafkaRebalance` status. 
+This might be because of excluded topics configuration (`excludedTopics`).
```
kyguy, line 133: Adding comments from an "outdated" thread here for visibility:

I felt that this sentence makes it sound as if excludedTopics could be the cause of the brokers being overloaded. From what I understand, excludedTopics does not affect the load on the brokers, it affects what partitions are allowed to be moved?

I thought the mention of excludeTopics was to warn users who are expecting excludeTopics to affect the load that excludeTopics has no affect on the load. Was that the intention or am I misunderstanding something?

documentation/modules/cruise-control/proc-generating-optimization-proposals.adoc
```diff
@@ -125,7 +125,30 @@ Status:
...
 +
 The properties in the `Optimization Result` section describe the pending cluster rebalance operation. 
-For descriptions of each property, see xref:contents-optimization-proposals[Contents of optimization proposals]. 
+For descriptions of each property, see xref:contents-optimization-proposals[Contents of optimization proposals].
+
+.Insufficient CPU capacity
+
+If a Kafka cluster is overloaded in terms of CPU utilization, you might see an insufficient CPU capacity error in the `KafkaRebalance` status. 
+This might be because of excluded topics configuration (`excludedTopics`). 
+Optimization proposals ignore excluded topics, but CPU utilization calculations consider their replicas.
```
kyguy, line 134: I am fine with leaving the mention of excludeTopics up here in the first paragraph to prevent user confusion but I think we should rephrase so something like:

> It's worth noting that this utilization value is unaffected by the excluded topics configuration (excludedTopics). Although optimization proposals will not reassign replicas of excluded topics, their load is still considered in the utilization calculation.

documentation/modules/cruise-control/proc-generating-optimization-proposals.adoc
fvaleri: I agree it is not clear enough.

AFAIK, when you set an `excludeTopics` regular expression, all topic names matching that expression are excluded from partition replica/leader movements in the optimization proposal. For that reason you, may never hit the CPU goal. Obviously, that would be a misconfiguration.

documentation/modules/cruise-control/proc-generating-optimization-proposals.adoc
fvaleri: Much better, thanks!

kyguy (approved):
Looks great @fvaleri!

scholzj:
Thanks for the PR @fvaleri

PR #6002, opened 2021-12-08
Add s390x support to multi-arch images on Azure CI

kun-lu20:

### Type of change

_Select the type of your PR_

- Enhancement / new feature

### Description

This PR is to add s390x support to Operator images on Azure CI.

### Checklist

- [ ] Write tests
- [ ] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/.azure/build-pipeline.yaml
+++ b/.azure/build-pipeline.yaml
@@ -53,7 +53,7 @@ stages:
           artifactPipeline: ''
           artifactRunVersion: ''
           artifactRunId: ''
-          architectures: ['amd64', 'arm64']
+          architectures: ['amd64', 'arm64', 's390x']
 
   # Push Strimzi containers -> run only on main branch
   - stage: push_containers
@@ -73,7 +73,7 @@ stages:
           artifactPipeline: ''
           artifactRunVersion: ''
           artifactRunId: ''
-          architectures: ['amd64', 'arm64']
+          architectures: ['amd64', 'arm64', 's390x']
 
   # Publish Strimzi docs to the website -> run only on main branch
   - stage: public_docs
--- a/.azure/cve-pipeline.yaml
+++ b/.azure/cve-pipeline.yaml
@@ -37,7 +37,7 @@ stages:
           artifactPipeline: '${{ parameters.sourcePipelineId }}'
           artifactRunVersion: 'specific'
           artifactRunId: '${{ parameters.sourceBuildId }}'
-          architectures: ['amd64', 'arm64']
+          architectures: ['amd64', 'arm64', 's390x']
   - stage: containers_publish_with_suffix
     displayName: Publish Containers for ${{ parameters.releaseVersion }}-${{ parameters.releaseSuffix }}
     dependsOn:
@@ -52,7 +52,7 @@ stages:
           artifactPipeline: ''
           artifactRunVersion: ''
           artifactRunId: ''
-          architectures: ['amd64', 'arm64']
+          architectures: ['amd64', 'arm64', 's390x']
   - stage: manual_validation
     displayName: Validate container before pushing container as ${{ parameters.releaseVersion }}
     dependsOn:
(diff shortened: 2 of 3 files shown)


scholzj:
I will just say what I said in the other PR for the Kafka Bridge. We should have some testing strategy before adding support for s390x.

kun-lu20:
Hi @scholzj , I am considering to add `.travis.yml` to the root directory for creating a CI workflow on Travis to build and run s390x tests, like what we did in Strimzi Kafka Bridge. Are you ok with that? Thanks!

scholzj:
What do you want to use it for? We have in general very bad experience with Travis. It is not good enough to run the system tests. It wasn't really that stable even with unit tests. I think the the operators are a bit different from the Bridge. In the Bridge the IT tests should give quite a good coverage. For the operators, the core of the coverage is really in the system tests which we need to run.

rposts:
Hi @scholzj , I am working with @kun-lu20 to come up with a test strategy as recommended by you earlier.  We have had success with TravisCI on `s390x` and hope that with a better system we may be able to achieve the end results on `s390x`.  I'd like to better understand some of the stumbling blocks you encountered using Travis CI towards system testing.

I also notice Jenkins CI used in the repo to run System Tests.  Admittedly, I am unfamiliar when this gets triggered but based on the description it seems to happen when a PR is pushed. I wonder if this is something we could consider for `s390x` arch.

I fully understand your concerns around having a testing strategy in place for `s390x`.  I would appreciate your guidance on how we can achieve this given some of the challenges encountered above.

scholzj:
@rposts The main issue with Travis were insufficient resources. Not just what was offered on paper, but mainly what was actually available. You have to keep in mind that the system tests need to run Kubernetes with one or more Kafka clusters with Connect, Mirror Maker etc. So it has different requirements than for example running the unit tests.

kun-lu20:
Hi @scholzj , hope all is well.

Reg adding s390x system tests to CI workflow, we're also working on the enablement of `Keycloak` and `Keycloak-operator` images which are dependencies of some system tests. The Keycloak community informed us that they will introduce new Quarkus based Keycloak images very soon and they suggested us to add s390x support on the new images.

I'd like to ask for your opinion concerning if we can enable part of s390x system tests on CI first. Once s390x support is ready on new Keycloak images, we can add the system tests which depend on them to our CI.

Thank you very much!

scholzj:
I do not think it would be a big issue if individual tests such as Keycloak or Open Policy Agent would be skipped.They are nto really key functionality, so I think we can live with those being disabled on s390x.

kun-lu20:
Thanks @scholzj

rposts:
Hi @scholzj,

As per your suggestion we investigated running system tests as part of s390x CI enablement. Here is a summary of our findings:

Attempts with Jenkins CI to run system tests was far more productive. Based on existing Jenkins-pr pipeline in the repo we were able to successfully run acceptance tests. This involved some modifications to existing scripts.

We would like to understand if Jenkins CI can be an option to run system tests on s390x. This would allow us to bridge the testing gap raised earlier.  Access to dedicated s390x community VMs can be provided for s390x Jenkins node.

Let us know your thoughts.

Thanks!

scholzj:
I do not think Jenkins itself is an issue. But the Jenkins server we use sometimes for some of the tests is located in a closed network and I'm not sure it can be used for this. @Frawless and @im-konge might have better idea, but I guess we might need to have somewhere the Jenkins server for this as well, not just the workers.

Frawless:
Ye, Jenkins is completely fine, however as @scholzj mentioned, our Jenkins is in a closed network and because of that, we moved most of the testing to azure, which is open to every contributor. Is your Jenkins instance exposed to outside or it's in a closed network?

rposts:
@Frawless I conducted my tests on internal Jenkins server/node.  Wonder if it is possible for your Jenkins instance to talk to a publicly hosted node?  Do you think this is something we can explore?

Frawless:
I am not sure if this is possible due to security. However, the Jenkins itself should be available to the community so it shouldn't be in a private network.

scholzj (approved):

kun-lu20:
Hi @Frawless , could you please have a look at this PR? Thank you very much!

Frawless (approved):

kun-lu20:
Thanks @scholzj @Frawless

Issue #5791, opened 2021-10-27
User and Topic operators should have separate users


scholzj:
The User and Topic operator share today a single user as _Entity Operator_. This makes it hard to distinguish them in some metrics and monitoring for example such as the case with #5790. We should have a separate user for each of them.

yyang48:
Is this _user_ from the **SASL_USERNAME** property? Where is this name set up in the `Entity Operator`?

scholzj:
No, when deployed as part of the Kafka cluster, the User and Topic operators use TLS client authentication. So it is about using separate certificates for each of them.

yyang48:
I see. Thanks for clarify!

yyang48:
I checked this part, it seems like the "*-entity-operator-certs" is _currently_ used by three containers: topic operator, user operator and tls sidecar. Does it make sense to create two more new certificates: "*-entity-**user**-operator-certs" and "-entity-**topic**-operator-certs"? In this way, each container has its own certificates:

1. topic operator use the new "-entity-topic-operator-certs".
2. user operator use the new "-entity-user-operator-certs".
3. TLS sidecar from the entity-operator deployment use the existing "-entity-operator-certs"

WDYT?

scholzj:
The TLS sidecar is now used only by the Topic Operator. So maybe it can just use the `-entity-topic-operator-certs` certs? But if it is somehow significantly easier to use `-entity-operator-certs` it might be fine as well.

yyang48:
@scholzj should we close this issue now, because the PR is merged?

scholzj:
@yyang48 Good point. Thanks for noticing.

PR #6006, opened 2021-12-08
[Issue 5791] Use separate certificates for the topic operator and user operator.

yyang48:
NOT A CONTRIBUTION

This is to address the following issue:
https://github.com/strimzi/strimzi-kafka-operator/issues/5791

Signed-off-by: Yaodong Yang <yyang48@apple.com>

### Type of change

_Select the type of your PR_

- Enhancement / new feature

### Description

Currently we use the same certificates for both Topic Operator and User Operator. As a result, they are using the same User/KafkaPrincipal when communicate to Kafka Brokers. From the monitoring perspective, there is no way to differentiate the activities from both of these operators, as described in the original issue.

This PR will create a separate certificate to the Topic Operator and User Operator. With this change, they will use two different user/KafkaPrincipal when talk to Kafka Brokers. The TLS sidecar will still share the certificate with the Topic Operator, as it did today.
 
### Checklist

- [ ] Write tests
- [ ] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaAssemblyOperator.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaAssemblyOperator.java
@@ -3547,7 +3547,8 @@ public class KafkaAssemblyOperator extends AbstractAssemblyOperator<KubernetesCl
         }
 
         Future<ReconciliationState> entityTopicOperatorSecret(Supplier<Date> dateSupplier) {
-            return updateCertificateSecretWithDiff(EntityTopicOperator.secretName(name), entityOperator == null || entityOperator.getTopicOperator() == null ? null : entityOperator.getTopicOperator().generateSecret(clusterCa, isMaintenanceTimeWindowsSatisfied(dateSupplier)))
+            // Since the TLS Sidecar uses the same certificate as the topic operator, we need to create the Topic Operator Certificate even if Only the User Operator is configured.
+            return updateCertificateSecretWithDiff(EntityTopicOperator.secretName(name), entityOperator == null ? null : entityOperator.generateSecret(clusterCa, isMaintenanceTimeWindowsSatisfied(dateSupplier)))
                     .map(changed -> {
                         existingEntityTopicOperatorCertsChanged = changed;
                         return this;
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaAssemblyOperator.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaAssemblyOperator.java
@@ -347,7 +347,8 @@ public class KafkaAssemblyOperator extends AbstractAssemblyOperator<KubernetesCl
                 .compose(state -> state.entityOperatorUserOpRoleBindingForRole())
                 .compose(state -> state.entityOperatorTopicOpAncillaryCm())
                 .compose(state -> state.entityOperatorUserOpAncillaryCm())
-                .compose(state -> state.entityOperatorSecret(this::dateSupplier))
+                .compose(state -> state.entityTopicOperatorSecret(this::dateSupplier))
+                .compose(state -> state.entityUserOperatorSecret(this::dateSupplier))
                 .compose(state -> state.entityOperatorDeployment())
                 .compose(state -> state.entityOperatorReady())
 
@@ -443,7 +444,8 @@ public class KafkaAssemblyOperator extends AbstractAssemblyOperator<KubernetesCl
         private boolean existingZookeeperCertsChanged = false;
         private boolean existingKafkaCertsChanged = false;
         private boolean existingKafkaExporterCertsChanged = false;
-        private boolean existingEntityOperatorCertsChanged = false;
+        private boolean existingEntityTopicOperatorCertsChanged = false;
+        private boolean existingEntityUserOperatorCertsChanged = false;
         private boolean existingCruiseControlCertsChanged = false;
 
         // Custom Listener certificates
@@ -3515,7 +3517,7 @@ public class KafkaAssemblyOperator extends AbstractAssemblyOperator<KubernetesCl
                 }).compose(recon -> {
                     if (recon instanceof ReconcileResult.Noop)   {
                         // Lets check if we need to roll the deployment manually
-                        if (existingEntityOperatorCertsChanged) {
+                        if (existingEntityTopicOperatorCertsChanged || existingEntityUserOperatorCertsChanged) {
                             return entityOperatorRollingUpdate();
                         }
                     }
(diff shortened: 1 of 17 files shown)


scholzj (commented):
Thanks for the PR. The code looks good to me, but the regression tests seem to be failing. Most of the tests seem to be failing becauze the User Operator does not seem to work. It looks like it has some issues connecting to Kafka:

```
  - type: "NotReady"
    status: "True"
    lastTransitionTime: "2021-12-09T13:41:47.976100Z"
    reason: "SslAuthenticationException"
    message: "Failed to process post-handshake messages"
```

I'm not sure exactly why. One of the things missing is probably to fix the user configuration in the Kafka brokers: https://github.com/strimzi/strimzi-kafka-operator/blob/main/cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaBrokerConfigurationBuilder.java#L471-L474. But I think the error would be something else as that fails before it gets to fail with this. 

Please have a look at it and feel free to ping me if you need any help.

yyang48:
Another surprised found in my local test, which caused the previous failure.

With this new commit, I think we should not see any current failures at least (finger crossed). @scholzj could you help me to trigger the pipeline whenever you get a chance, thanks!

user-operator/src/main/java/io/strimzi/operator/user/UserOperatorConfig.java
```diff
@@ -21,7 +21,7 @@
     public static final String STRIMZI_CA_CERT_SECRET_NAME = "STRIMZI_CA_CERT_NAME";
     public static final String STRIMZI_CA_KEY_SECRET_NAME = "STRIMZI_CA_KEY_NAME";
     public static final String STRIMZI_CLUSTER_CA_CERT_SECRET_NAME = "STRIMZI_CLUSTER_CA_CERT_SECRET_NAME";
-    public static final String STRIMZI_EO_KEY_SECRET_NAME = "STRIMZI_EO_KEY_SECRET_NAME";
+    public static final String STRIMZI_EUO_KEY_SECRET_NAME = "STRIMZI_EUO_KEY_SECRET_NAME";
```
scholzj, line 24: There are users using the User Operator _standalone_ with Kafka clusters not deployed by Strimzi. They use the files for it from here: https://github.com/strimzi/strimzi-kafka-operator/tree/main/packaging/install/user-operator. Changing the environment variable name is basically a backwards-incompatible change of API for them. It is also covered in the docs https://strimzi.io/docs/operators/latest/full/deploying.html#deploying-the-user-operator-standalone-str.

I think the best thing would be to keep the old environment variable as it was. That way, there is no need to change the docs and no breaking-change for the standalone user operator users.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaAssemblyOperator.java
```diff
@@ -347,7 +347,8 @@ public KafkaAssemblyOperator(Vertx vertx, PlatformFeaturesAvailability pfa,
                 .compose(state -> state.entityOperatorUserOpRoleBindingForRole())
                 .compose(state -> state.entityOperatorTopicOpAncillaryCm())
                 .compose(state -> state.entityOperatorUserOpAncillaryCm())
-                .compose(state -> state.entityOperatorSecret(this::dateSupplier))
+                .compose(state -> state.entityTopicOperatorSecret(this::dateSupplier))
+                .compose(state -> state.entityUserOperatorSecret(this::dateSupplier))
```
scholzj, line 351: The upgrade tests seem to pass fine. But I do not see any place where we delete the old certificate which is not completely nice since it would basically stay there forever. It would be nice if we can delete it as well here. Basically just adding some step with `secretOperations.reconcile(..., ..., ..., null)` and some comment explaining why are we deleting it.

scholzj (commented):
It looks good and the tests are passing. I just found two more comments. Sorry for not finding them earlier.

tombentley (approved):
Apart from Jakub's existing comments this LGTM. Thanks!

user-operator/src/main/java/io/strimzi/operator/user/UserOperatorConfig.java
yyang48: Sounds good! I will bring it back. Basically, I will do this in the UserOperatorConfig.java
```
        String euoKeySecretName = map.get(UserOperatorConfig.STRIMZI_EUO_KEY_SECRET_NAME);
        if (euoKeySecretName == null || euoKeySecretName.isEmpty()) {
            // It's needed for the backward compatible with the standAlone deployment of the User Operator.
            euoKeySecretName = map.get(UserOperatorConfig.STRIMZI_EO_KEY_SECRET_NAME);
        }
```

I guess this should work in both modes: StandAlone mode vs Strimzi mode.

scholzj: In general, yes ... but wouldn't it be just easier to use the original name?

yyang48: ah, yes! Let me change it.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaAssemblyOperator.java
yyang48: @scholzj I looked at the code again, and it seems like we have the code to delete the certificates when the topic Operator or user Operator is deleted. Check this:
```
Future<ReconciliationState> entityTopicOperatorSecret(Supplier<Date> dateSupplier) {
            return updateCertificateSecretWithDiff(EntityTopicOperator.secretName(name), entityOperator == null || entityOperator.getTopicOperator() == null ? null : entityOperator.getTopicOperator().generateSecret(clusterCa, isMaintenanceTimeWindowsSatisfied(dateSupplier)))
        .map(changed -> {
            existingEntityTopicOperatorCertsChanged = changed;
            return this;
        });
}
```

Basically the Secret object will be null, whenever the topic Operator is removed. Within the secretOperations.getAsync() method, there is a call to the secretOperations.reconcile(). -> This method does take a null as the last argument.

I also manually run this test: `testRemoveUserOperatorFromEntityOperator`. I do see the secret for the user operator was deleted and then recreated in the test.

Does this make sense to you?

user-operator/src/main/java/io/strimzi/operator/user/UserOperatorConfig.java
yyang48: done.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaAssemblyOperator.java
yyang48: I meant this

`entityOperator == null || entityOperator.getTopicOperator() == null ? null : entityOperator.getTopicOperator().generateSecret(clusterCa, isMaintenanceTimeWindowsSatisfied(dateSupplier))`

The secret will become null if the topic operator is not defined. Otherwise, try to generate the secret

scholzj: That is correct. But what I meant here is something else. If you run Strimzi 0.26, it will maintain a secret named `<my-cluster>-entity-operator-certs` with certificate inside. Now once your PR is merged and users upgrade to it, there will be two new secrets: `<my-cluster>-entity-user-operator-certs` and `<my-cluster>-entity-topic-operator-certs`. But nobody will delete the old `<my-cluster>-entity-operator-certs` secret which will not be used anywhere but will still exist. Or is the old secret somewhere deleted and I just missed it?

ppatierno (approved):
LGTM. Thanks!

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaAssemblyOperator.java
yyang48: I see what you mean now:-). You are right! We don't clean up the old certificate now, which we should. Let me add it in this PR.

scholzj (approved):
Great, thanks for the PR. I will run the system tests once more just to check nothing broke since the last time and merge it.

PR #6012, opened 2021-12-09
Test items in Kafka Connect and its external configuration

### LINKED ISSUE #2072: Revert changes from #1212
The issue in kubernetes (https://github.com/kubernetes/kubernetes/issues/68466) which was addressed in https://github.com/strimzi/strimzi-kafka-operator/issues/1212 has been fixed, so once our CI is using a fixed version of kube we should be able to revert those changes.

scholzj:
### Type of change

- Task

### Description

Some time ago, we had to disable the use of the `items` field in the Kafka Connect external configuration due to a bug in Kubernetes. This bug is fixed from 1.16, so we should now be able to use it in the tests again.

This should close #2072

### Checklist


scholzj:
Closing as it is not really fixed in Kube 1.16 but only in 1.17 🙄

PR #6027, opened 2021-12-12
Download Kafka binaries from ASF CDN

fvaleri:
The function `download_kafka_binaries_from_mirror` is constantly failing during the build, as `mirrors.cgi` does not exists anymore. As a result, we are always falling back to the slower archive URL.

Recently, the [ASF has moved to CDN](https://blogs.apache.org/foundation/entry/apache-software-foundation-moves-to) and the [mirrors system is no longer needed/available](https://www.apache.org/dyn/closer.cgi).

For these reasons, I'm proposing to download from CDN first and fallback to archive only when needed (i.e. Kafka 2.8.0 is only available in archive, as it is replaced by Kafka 2.8.1).

Tested with a full build.

### DIFF
--- a/docker-images/artifacts/build.sh
+++ b/docker-images/artifacts/build.sh
@@ -105,7 +105,7 @@ function fetch_and_unpack_kafka_binaries {
         if [ $get_file -gt 0 ]
         then
             echo "Fetching Kafka $kafka_version binaries from: $binary_file_url"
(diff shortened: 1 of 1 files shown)


tombentley (approved):

scholzj (approved):

scholzj:
Thanks for the PR.

PR #6029, opened 2021-12-13
Refactoring: add STRIMZI_DNS_CACHE_TTL to ClusterOperatorConfig

fvaleri:
No change in behavior.


scholzj (changes requested):
Why are you changing it in a backwards incompatible way?

fvaleri:
> Why are you changing it in a backwards incompatible way?

Do you mean seconds vs millis right?

scholzj:
> Do you mean seconds vs millis or variable name?

Well, it is both really. I'm not sure if this option needs to be formalized - I do not remember using or needing it for a long time. But if we do it, we might want to do it in a backwards compatible fashion and keek the old name and format.

scholzj:
Apart from that, an explanation when to use it wld be good for the docs I guess.

fvaleri:
Sure let me fix that, thanks.

scholzj (approved):
Thanks for the PR. LGTM

ppatierno (approved):

PR #6005, opened 2021-12-08
Integrate client timeouts for `oauth` authentication and `keycloak` authorization

mstruk:
Signed-off-by: Marko Strukelj <marko.strukelj@gmail.com>

### Type of change

- Enhancement / new feature
- Documentation

### Description

Integrate client timeouts when connecting to authorization server introduced in strimzi-kafka-oauth 0.10.0

### Checklist

_Please go through this checklist and make sure all applicable tasks have been done_

- [ ] Write tests
- [ ] Make sure all tests pass
- [ ] Update documentation
- [ ] Check RBAC rights for Kubernetes / OpenShift roles
- [ ] Try your changes from Pod inside your Kubernetes and OpenShift cluster, not just locally
- [ ] Reference relevant issue(s) and close them after merging
- [ ] Update CHANGELOG.md
- [ ] Supply screenshots for visual changes, such as Grafana dashboards

PR #6032, opened 2021-12-13
Groups extraction for OAuth authentication

mstruk:
### Type of change

- Enhancement / new feature
- Documentation

### Description

This PR integrates the groups extraction from JWT token or the result of the introspection endpoint to be used by a custom authorizer.

### Checklist

- [ ] Write tests
- [ ] Update documentation
- [ ] Update CHANGELOG.md


mstruk:
This PR depends on #6005

docker-images/artifacts/kafka-thirdparty-libs/2.8.x/pom.xml
```diff
@@ -16,7 +16,7 @@
     </licenses>
 
     <properties>
-        <strimzi-oauth.version>0.8.1</strimzi-oauth.version>
+        <strimzi-oauth.version>0.10.0-rc1-pre</strimzi-oauth.version>
```
scholzj, line 19: Will this work with Kafka 2.8.0? I remember we did changes to the authorizer part in 3.0.0.

mstruk: It will work, but it does slightly change the logging. So if someone is using `keycloak` authorization and is parsing the Kafka logs applying some regexes, then their regexes may no longer catch what they were catching before. Upgrading to 3.0.x would result in the same change. Thus in previous version there was a slight difference in logging when using `keycloak` authorization with 2.8.x vs. 3.0.x. By using the new oauth version on both that difference will now be gone. And it is necessary anyway in order to provide the new features (client timeouts, group extraction).

scholzj: > It will work, but it does slightly change the logging. So if someone is using keycloak authorization and is parsing the Kafka logs applying some regexes, then their regexes may no longer catch what they were catching before.

Thanks, If it is only logging, I don't think that would be a problem. At least not for me :-o.

mstruk:
While this is still in draft to prevent it from merging before strimzi-kafka-oauth project is released, the code itself is ready for review. 
This PR build on #6005. Once that one is merged this one will be rebased again on top of main. Until then, for the purposes of reviewing, use the filter that only shows the last commit: https://github.com/strimzi/strimzi-kafka-operator/pull/6032/commits/0a28e3b3005015bdf2ce793dd629b6de2f38061e

scholzj (approved):

ppatierno (approved):

PR #6040, opened 2021-12-15
Cc cpu util

sknot-rh:
### Type of change
- Enhancement

### Description
The capacity configuration (that allows Cruise Control to balance resources) is currently set manually by the user, which can easily be missed in setting up the Kafka CR. The Strimzi Operator should set these values automatically using the resource limits in the pod templates.

### Checklist
- [x] Write tests
- [ ] Update documentation
- [ ] Update CHANGELOG.md


cluster-operator/src/main/java/io/strimzi/operator/cluster/model/cruisecontrol/Capacity.java
```diff
@@ -31,7 +32,9 @@ public Capacity(KafkaSpec spec) {
         BrokerCapacity bc = spec.getCruiseControl().getBrokerCapacity();
 
         this.diskMiB = bc != null && bc.getDisk() != null ? getSizeInMiB(bc.getDisk()) : generateDiskCapacity(spec.getKafka().getStorage());
-        this.cpuUtilization = bc != null && bc.getCpuUtilization() != null ? bc.getCpuUtilization() : DEFAULT_BROKER_CPU_UTILIZATION_CAPACITY;
+        this.cpuUtilization = bc != null && bc.getCpuUtilization() != null ? bc.getCpuUtilization() :
+                (spec.getKafka().getResources() == null || spec.getKafka().getResources().getLimits() == null ||
+                        spec.getKafka().getResources().getLimits().get("cpu") == null) ? DEFAULT_BROKER_CPU_UTILIZATION_CAPACITY : Quantities.parseCpuAsMilliCpus(spec.getKafka().getResources().getLimits().get("cpu").toString()) / 10;
```
scholzj, line 37: Why do you use limit and not request? Also, why do you divide it by 10?

kyguy (commented):
Hey Standa, I think there may be a misunderstanding here. The Cruise Control CPU capacity config, e.g. `spec.cruiseControl.capacity` is used to configure the percentage of CPU resources a broker can use out of the total CPU resources available to that broker, not for setting the CPU capacity itself.  By default, the CPU capacity config is set at 100, meaning that 100% of CPU resources available to the broker pod can be used. When `spec.kafka.resource.limit` is set, Cruise Control will automatically set the broker CPU capacity to `spec.kafka.resource.limit`. When a user sets the Cruise Control CPU capacity to 50, Cruise Control will  set the broker CPU capacity to 50% of `spec.kafka.resource.limit`.

The `spec.cruiseControl.capacity` is for users to customize how much of the total CPU capacity available to a broker pod that they want that broker to use. We do not need to set `spec.cruiseControl.capacity` unless we are looking to address the UI issue mentioned here [3], in which case we would want to use a **core count** CPU capacity config, see below:

Cruise Control only accepts a CPU config as a
-  **Percentage**: an integer value between [0, 100] inclusive. This value represents the percentage of CPU resources a broker can use out of the total CPU resources available to that broker, see example config here [1]
-  **Core count**: integer value in a JSON object format. This represent the number of cores a broker can use at 100% , see example config here [2] 

Cruise Control will crash if we provide values outside this **percentage** interval or **core** format. 

Is this PR meant to address the UI issue mentioned here [3] ?

[1] https://github.com/linkedin/cruise-control/blob/migrate_to_kafka_2_5/config/capacity.json#L7
[2] https://github.com/linkedin/cruise-control/blob/migrate_to_kafka_2_5/config/capacityCores.json#L7
[3] https://github.com/strimzi/strimzi-kafka-operator/issues/5951#issuecomment-983953897

cluster-operator/src/test/java/io/strimzi/operator/cluster/model/CruiseControlTest.java
```diff
@@ -379,6 +380,45 @@ public void testImagePullPolicy() {
...
+
+        // value without m
+        k = createKafka(cruiseControlSpec);
+        resources = new ResourceRequirementsBuilder()
+                .withLimits(singletonMap("cpu", new Quantity("2")))
+                .build();
+        k.getSpec().getKafka().setResources(resources);
+        generatedCapacity = new Capacity(k.getSpec());
+        assertThat(getCapacityConfigurationFromEnvVar(k, ENV_VAR_BROKER_CPU_UTILIZATION_CAPACITY), is(Integer.toString(generatedCapacity.getCpuUtilization())));
+        assertThat(getCapacityConfigurationFromEnvVar(k, ENV_VAR_BROKER_CPU_UTILIZATION_CAPACITY), is(Integer.toString(200)));
```
kyguy, line 409: We cannot specify percentage values outside  the interval [0,100] or else Cruise Control will crash when it reads the generated capacity config file we generate.

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/cruisecontrol/Capacity.java
kyguy: Cruise Control will crash if we provide values outside the interval [0,100], see my comment below

tomncooper:
Sorry @sknot-rh I may have sent you down the wrong path here. Out of interest @kyguy can you point to where in the code: "Cruise Control will automatically set the broker CPU capacity to spec.kafka.resource.limit"?

kyguy:
> can you point to where in the code: "Cruise Control will automatically set the broker CPU capacity to spec.kafka.resource.limit"?

Sorry @tomncooper I misspoke, what I meant to say is that by default the Cruise Control Broker CPU capacity config will be set to 100 which means that a broker will not be considered over capacity if it is using 100% of the CPU resources available to it (which would be `spec.kafka.resource.limit` if set). 

@sknot-rh I was also wrong about the **percentage** interval Cruise Control takes. After testing it today I found the CPU **percentage** config can take double values outside the [0, 100] interval without Cruise Control crashing. However, from what I gathered from the code [1], if we provide percentage values outside the [0,100] interval, the CPU utilization calculations get messed up. 

For example, if we look at the how Cruise Control calculates CPU utilization with respect to its defined capacity here [1]
```
UNIT_INTERVAL_TO_PERCENTAGE * ACTUAL_CPU_UTIL / CAPACITY
```
and we assume the following values for the following variables:
`UNIT_INTERVAL_TO_PERCENTAGE`= 100 [2]
`ACTUAL_CPU_UTIL` = 100 
then try different `CAPACITY` values: 
```
When we set CAPACITY = 100, the broker is using 1x or 100% of its CPU capacity
100 * 100 / 100 = 100

When we set CAPACITY =  50, the broker is using 2x or 200% of its CPU capacity
100 * 100 / 50 =  200
```
The problem arises when we set CAPACITY values > 100
```
When we set CAPACITY = 200, the broker is using 0.5x or 50% of its CPU capacity? 
100 * 100 / 200 = 50
```
This doesn't make sense because  ACTUAL_CPU_UTIL = 100, meaning that the broker is using 100% of its CPU resources, so the broker has no capacity left!

[1] https://github.com/linkedin/cruise-control/blob/2.5.73/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/response/stats/BasicStats.java#L75
[2] https://github.com/linkedin/cruise-control/blob/migrate_to_kafka_2_5/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/monitor/MonitorUtils.java#L52

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/cruisecontrol/Capacity.java
kyguy: @sknot-rh please ignore my previous comment, Cruise Control will not crash if we provide values outside the interval [0,100] but the utilization numbers don't seem to make sense when we do

sknot-rh:
> Sorry @tomncooper I misspoke, what I meant to say is that by default the Cruise Control Broker CPU capacity config will be set to 100 which means that a broker will not be considered over capacity if it is using 100% of the CPU resources available to it (which would be spec.kafka.resource.limit if set).

It seems CO already copes with setting of the CC CPU utilization. Closing.

Issue #6048, opened 2021-12-16
Cluster Reconciliation Deadlock on Invalid inter.broker.protocol.version / log.message.format.version


fuchstim:
**Describe the bug**
Whilst upgrading Kafka to v2.8.0, due to a misunderstanding I entered `2.8-IV1` as the `inter.broker.protocol.version` and `log.message.format.version` config (see YAML extract below).

Being an invalid version number, this resulted in the operator being unable to complete reconciliation (see stack trace below).

However, trying to roll back / fix this mistake proved a bit difficult, because even after updating and applying the Kafka Object to fix the bad version string, the operator still refused to complete reconciliation.

Finally what resolved this issue for us was to manually edit the `strimzi.io/log-message-format-version` and `strimzi.io/inter-broker-protocol-version` annotations on the existing Kafka Broker pods (edited to be a valid version number), allowing the operator to successfully complete reconciliation and roll back the change.

**To Reproduce**
1. Deploy a Strimzi Kafka resource with an invalid inter.broker.protocol.version / log.message.format.version version number
2. Wait for the operator to trigger a periodic reconciliation
3. Reconciliation fails, as expected
4. Rolling back config change is impossible without manual intervention

**Expected behavior**
Strimzi should validate the version number being configured on the `strimzi.io/log-message-format-version` and `strimzi.io/inter-broker-protocol-version` annotations *before* starting to update resources, to avoid this deadlock scenario

**Environment (please complete the following information):**
 - Strimzi version: 0.22.1, 0.23.0, 0.26.0, 0.26.1
 - Installation method: Helm Chart
 - Kubernetes cluster: Kubernetes v1.21.2-eks
 - Infrastructure: EKS

**YAML files and logs**
Kafka Cluster YAML:

```yaml
apiVersion: kafka.strimzi.io/v1beta2
kind: Kafka
metadata:
  name: {{ .Release.Name }}-cluster
  namespace: {{ .Release.Namespace }}
spec:
  kafka:
    version: 2.8.0
    replicas: 3
    listeners:
      - name: tls
        port: 9093
        type: internal
        tls: true
        authentication:
          type: tls
    authorization:
      type: simple
      superUsers: []
    config:
      offsets.topic.replication.factor: 3
      transaction.state.log.replication.factor: 3
      transaction.state.log.min.isr: 2
      log.message.format.version: "2.8-IV1"
      inter.broker.protocol.version: "2.8-IV1"
    storage:
      type: ephemeral
[...]

PR #6057, opened 2021-12-17
Fix kafka version comparing when message format is used instead

sknot-rh:

### Type of change
- Bugfix

### Description
`compareDottedVersions` does not cope with versions like `3.0-IV1` so it cannot be used for comparing `inter.broker.protocol.version`. 
If the version contains `-` character, only its prefix is used. So if we provide `2.8-IV1`, the version `2.8` will be used.

Fixes https://github.com/strimzi/strimzi-kafka-operator/issues/6048

### Checklist
- [ ] Write tests
- [ ] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaVersion.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaVersion.java
@@ -449,6 +449,7 @@ public class KafkaVersion implements Comparable<KafkaVersion> {
      * 1 if version1 &gt; version2.
      */
     public static int compareDottedIVVersions(String version1, String version2) {
+        // Due to validation in KafkaConfiguration.validate only values like 2.8 or 2.8-IV0 can get there
         String trimmedVersion1 = version1.split("-")[0];
         String trimmedVersion2 = version2.split("-")[0];
         return compareDottedVersions(trimmedVersion1, trimmedVersion2);
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaVersion.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaVersion.java
@@ -423,9 +423,6 @@ public class KafkaVersion implements Comparable<KafkaVersion> {
      * 1 if version1 &gt; version2.
      */
     public static int compareDottedVersions(String version1, String version2) {
-        // Version like X.Y-IVZ can get through validation
-        version1 = version1.split("-")[0];
-        version2 = version2.split("-")[0];
         String[] components = version1.split("\\.");
         String[] otherComponents = version2.split("\\.");
         for (int i = 0; i < Math.min(components.length, otherComponents.length); i++) {
@@ -441,6 +438,20 @@ public class KafkaVersion implements Comparable<KafkaVersion> {
         }
         // mismatch was not found, but the versions are of different length, e.g. 2.8 and 2.8.0
         return 0;
+    }
+
+    /**
+     * Compare two decimal version strings, e.g. 3.0 &gt; 2.7-IV1. Ignores the IV part
+     * @param version1 The first version.
+     * @param version2 The second version.
+     * @return Zero if version1 == version2;
+     * -1 if version1 &lt; version2;
+     * 1 if version1 &gt; version2.
+     */
+    public static int compareDottedIVVersion(String version1, String version2) {
+        String trimmedVersion1 = version1.split("-")[0];
+        String trimmedVersion2 = version2.split("-")[0];
+        return compareDottedVersions(trimmedVersion1, trimmedVersion2);
     }
 
     @Override
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaVersion.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaVersion.java
@@ -423,6 +423,9 @@ public class KafkaVersion implements Comparable<KafkaVersion> {
      * 1 if version1 &gt; version2.
      */
     public static int compareDottedVersions(String version1, String version2) {
+        // Version like X.Y-IVZ can get through validation
+        version1 = version1.split("-")[0];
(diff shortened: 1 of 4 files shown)


cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaVersion.java
```diff
@@ -423,6 +423,9 @@ public int compareVersion(String version) {
      * 1 if version1 &gt; version2.
      */
     public static int compareDottedVersions(String version1, String version2) {
+        // Version like X.Y-IVZ can get through validation
+        version1 = version1.split("-")[0];
+        version2 = version2.split("-")[0];
```
scholzj, line 428: Isn't the same method also used to compare the protocol versions?

scholzj:
Maybe you can elaborate a bit more on:
* What is the problem and what happens today when it fails
* How does this fix it

cluster-operator/src/test/java/io/strimzi/operator/cluster/model/KafkaVersionTest.java
```diff
@@ -31,6 +31,20 @@ private Reader getKafkaVersionsReader(String kafkaVersions) {
         return new InputStreamReader(KafkaVersion.class.getResourceAsStream("/" + kafkaVersions), StandardCharsets.UTF_8);
     }
 
+    @ParallelTest
+    public void parsingInvalidVersionTest() {
```
scholzj, line 35: I'm not sure I get the test => you added new method `compareDottedIVVersion`. But you do not seem to use it in the test. Plus the `compareDottedVersion` which I would expect to fail seems to be passing here for `2.8-IV1`.

scholzj (commented):
The code looks good. But I do not really understand the test you added. I think it would be also good to add test using the IV versions in LMFV and IBPV into `KafkaUpgradeDowngradeMockTest`.

cluster-operator/src/test/java/io/strimzi/operator/cluster/model/KafkaVersionTest.java
```diff
@@ -31,6 +31,24 @@ private Reader getKafkaVersionsReader(String kafkaVersions) {
...
+
+        assertThrows(NumberFormatException.class, () -> {
+            KafkaVersion kvFail = new KafkaVersion("why", "you", "little", "3.6.9", false, true, "");
+            KafkaVersion.compareDottedIVVersions("2.8.0", kvFail.protocolVersion());
+        });
+
+        assertThrows(NumberFormatException.class, () -> {
+            KafkaVersion kvFail = new KafkaVersion("2.8.1-IV1", "you", "little", "3.6.9", false, true, "");
+            KafkaVersion.compareDottedVersions("2.8.0", kvFail.protocolVersion());
+        });
```
scholzj, line 48: BTW: The `KafkaVersion` instances are managed by us. So they would not contain anything stupid. They will IMHO always be something like `new KafkaVersion("2.8.0", "2.8", "2.8", "3.6.9", false, true, "")`. It is the version you compare it to which will be usually user provided? So maybe the tests make more sense the other way around? Having a valid `KafkaVersion` and comparing it with invalid values?

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaVersion.java
```diff
@@ -440,6 +440,20 @@ public static int compareDottedVersions(String version1, String version2) {
...
+     * Compare two decimal version strings, e.g. 3.0 &gt; 2.7-IV1. Ignores the IV part
+     * @param version1 The first version.
+     * @param version2 The second version.
+     * @return Zero if version1 == version2;
+     * -1 if version1 &lt; version2;
+     * 1 if version1 &gt; version2.
+     */
+    public static int compareDottedIVVersions(String version1, String version2) {
+        String trimmedVersion1 = version1.split("-")[0];
+        String trimmedVersion2 = version2.split("-")[0];
```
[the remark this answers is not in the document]
sknot-rh: This should be already validated by `KafkaConfiguration.validate(KafkaVersion)`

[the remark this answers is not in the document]
tombentley: Could you add a comment then?

tombentley (approved):

cluster-operator/src/test/java/io/strimzi/operator/cluster/operator/assembly/KafkaUpgradeDowngradeMockTest.java
```diff
@@ -211,6 +211,40 @@ public void testUpgradeWithMessageAndProtocolVersions(VertxTestContext context)
...
+    @Test
+    public void testUpgradeWithIVMessageAndProtocolVersions(VertxTestContext context)  {
+        String iv = "-IV0";
+        Kafka initialKafka = kafkaWithVersions(KafkaVersionTestUtils.PREVIOUS_KAFKA_VERSION,
+                KafkaVersionTestUtils.PREVIOUS_FORMAT_VERSION + iv,
+                KafkaVersionTestUtils.PREVIOUS_PROTOCOL_VERSION + iv);
+
+        Kafka updatedKafka = kafkaWithVersions(KafkaVersionTestUtils.LATEST_KAFKA_VERSION,
+                KafkaVersionTestUtils.PREVIOUS_FORMAT_VERSION,
+                KafkaVersionTestUtils.PREVIOUS_PROTOCOL_VERSION);
```
scholzj, line 225: Should it use the `-IV0` version here as well? Essentially, but not using it here, you limit what is tested here. The test as it is right now suggests that the user has Kafka configured with for example:
* version 2.8.0
* IBPV 2.8-IV0
* LMFV 2.8-IV0

An suddenly changes to for example:
* version 3.0.0
* IBPV 2.8
* LMFV 2.8

This seems not likely.

scholzj (approved):
Thanks- LGTM assuming the regression tests pass.

PR #6062, opened 2021-12-17
Use the same model for printing executed commands in all containers

scholzj:
### Type of change

- Bugfix

### Description

We have currently in each container different policy of which commands executed during the startup are or aren't printed using the `set -x` and `set +x` fields. This PR tries to unify it to:
* Always print the main command we are starting
* Don't print all the preparation of the config files etc. (it still prints intentionally logged messages using `echo`)

This should help to create a cleaner log output when the containers start while still showing the important part which is the main command.

### Checklist

### DIFF
--- a/docker-images/jmxtrans/docker-entrypoint.sh
+++ b/docker-images/jmxtrans/docker-entrypoint.sh
@@ -1,5 +1,6 @@
 #!/usr/bin/env bash
 set -e
+set +x
 
 EXEC="-jar $JAR_FILE -e -j $JSON_DIR -s $SECONDS_BETWEEN_RUNS -c $CONTINUE_ON_ERROR $ADDITIONAL_JARS_OPTS"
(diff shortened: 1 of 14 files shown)


docker-images/jmxtrans/docker-entrypoint.sh
```diff
@@ -1,5 +1,6 @@
 #!/usr/bin/env bash
 set -e
+set +x
```
tombentley, line 3: Do you think there's some debug value in being able to conditionally echo the commands executed in the preparation, e.g. if some env var is set?

scholzj: I don't think I see the need. I guess it could be aded later if we find it missing In many containers it was anyway disabled already before this PR, so it is more cleanup then radical change.

scholzj: @tombentley So, is it ok with you like this? Or do you want some changes?

tombentley (approved):

PR #6063, opened 2021-12-18
Fix InternalKafkaClient and VerifiableClient

fvaleri:
Even though the ColdBackupScriptST was running fine, there were two errors printed to stderr. One of them revealed a couple of bugs in the ST framework (who tests the testing framework?).

### 1s error (not a problem)

```log
ls: cannot access '/home/vsts/work/1/s/systemtest/target/infra-namespace/my-cluster-95134611': No such file or directory
```

The fact that we have non empty stderr does not necessary mean that the script has failed, it could be expected like in this case. We only have one `ls` command in the entire script. As I show below, the exit code is 0 when the directory is not found. We can suppress the stderr, but that would hide other important errors (e.g. not having permissions to list files).

```
$ if [[ ! -z "$(ls -A foo)" ]]; then
 echo "Non empty directory"
fi
ls: cannot access 'foo': No such file or directory
$ echo $?
0
```

### 2nd error (ST framework bugs)

```log
Missing required argument "[bootstrap-server]"...
```

##### 1st bug
 
This happens two times when getting the offsets map, before the backup and after the restore. Instead of returning a runtime error, the `InternalKafkaClient.getCurrentOffsets()` returns an empty map, so the final assertion which compares them still holds true (added an additional check to avoid this).

##### 2nd bug

The `InternalKafkaClient` uses the `VerifiableClient.allowParameter()` to choose between `--bootstrap-server` and the deprecated `--broker-list` parameters. Recently we moved from Kafka 2.x to Kafka 3.x, and the following expression used in that method does not work anymore.

```java
return Integer.parseInt(current.group("major")) >= Integer.parseInt(minimal.group("major"))
    && Integer.parseInt(current.group("minor")) >= Integer.parseInt(minimal.group("minor"))
    && Integer.parseInt(current.group("micro")) >= Integer.parseInt(minimal.group("micro"));
```

### DIFF
--- a/systemtest/src/main/java/io/strimzi/systemtest/kafkaclients/internalClients/InternalKafkaClient.java
+++ b/systemtest/src/main/java/io/strimzi/systemtest/kafkaclients/internalClients/InternalKafkaClient.java
@@ -312,6 +312,9 @@ public class InternalKafkaClient extends AbstractKafkaClient<InternalKafkaClient
 
         boolean hasPassed = consumerGroups.run(timeoutMs);
         LOGGER.info("ConsumerGroups finished correctly: {}", hasPassed);
+        if (!hasPassed) {
+            throw new RuntimeException("VerifiableClient has failed. Check stderr for more details.");
+        }
 
         // output parsing
         Map<String, String> currentOffsets = new HashMap<>();
--- a/systemtest/src/main/java/io/strimzi/systemtest/kafkaclients/internalClients/VerifiableClient.java
+++ b/systemtest/src/main/java/io/strimzi/systemtest/kafkaclients/internalClients/VerifiableClient.java
@@ -330,9 +330,16 @@ public class VerifiableClient {
         Matcher current = pattern.matcher(clientVersion);
         Matcher minimal = pattern.matcher(minimalVersion);
         if (current.find() && minimal.find()) {
-            return Integer.parseInt(current.group("major")) >= Integer.parseInt(minimal.group("major"))
-                    && Integer.parseInt(current.group("minor")) >= Integer.parseInt(minimal.group("minor"))
-                    && Integer.parseInt(current.group("micro")) >= Integer.parseInt(minimal.group("micro"));
+            int major = Integer.parseInt(current.group("major")), 
+                minor = Integer.parseInt(current.group("minor")),
+                micro = Integer.parseInt(current.group("micro"));
+            int minMajor = Integer.parseInt(minimal.group("major")), 
+                minMinor = Integer.parseInt(minimal.group("minor")),
+                minMicro = Integer.parseInt(minimal.group("micro")); 
+            return 
+                    (major > minMajor)
+                    || (major == minMajor && minor > minMinor)
+                    || (major == minMajor && minor == minMinor && micro >= minMicro);
         }
         return false;
     }
--- a/systemtest/src/main/java/io/strimzi/systemtest/kafkaclients/internalClients/VerifiableClient.java
+++ b/systemtest/src/main/java/io/strimzi/systemtest/kafkaclients/internalClients/VerifiableClient.java
@@ -4,7 +4,6 @@
  */
 package io.strimzi.systemtest.kafkaclients.internalClients;
 
-import static io.strimzi.systemtest.resources.ResourceManager.kubeClient;
 import static io.strimzi.test.k8s.KubeClusterResource.cmdKubeClient;
 
 import java.io.IOException;
@@ -121,19 +120,11 @@ public class VerifiableClient {
         this.clientArgumentMap.put(ClientArgument.MAX_MESSAGES, Integer.toString(maxMessages));
         if (kafkaUsername != null) this.clientArgumentMap.put(ClientArgument.USER, kafkaUsername.replace("-", "_"));
 
-        String image = kubeClient().namespace(podNamespace).getPod(podNamespace, this.podName).getSpec().getContainers().get(0).getImage();
-        String clientVersion = image.substring(image.length() - 5);
-
-        this.clientArgumentMap.put(allowParameter("2.5.0", clientVersion) ? ClientArgument.BOOTSTRAP_SERVER : ClientArgument.BROKER_LIST, bootstrapServer);
+        this.clientArgumentMap.put(ClientArgument.BOOTSTRAP_SERVER, bootstrapServer);
 
         if (clientType == ClientType.CLI_KAFKA_VERIFIABLE_CONSUMER) {
             this.consumerGroupName = verifiableClientBuilder.consumerGroupName;
             this.clientArgumentMap.put(ClientArgument.GROUP_ID, consumerGroupName);
-
-            if (allowParameter("2.3.0", clientVersion)) {
-                this.consumerInstanceId = verifiableClientBuilder.consumerInstanceId;
-                this.clientArgumentMap.put(ClientArgument.GROUP_INSTANCE_ID, this.consumerInstanceId);
-            }
         }
 
         if (clientType == ClientType.CLI_KAFKA_CONSUMER_GROUPS) {
--- a/systemtest/src/main/java/io/strimzi/systemtest/kafkaclients/internalClients/VerifiableClient.java
+++ b/systemtest/src/main/java/io/strimzi/systemtest/kafkaclients/internalClients/VerifiableClient.java
@@ -12,8 +12,6 @@ import java.util.Arrays;
 import java.util.List;
 import java.util.Objects;
 import java.util.concurrent.ExecutionException;
-import java.util.regex.Matcher;
-import java.util.regex.Pattern;
 
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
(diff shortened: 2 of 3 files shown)


fvaleri:
@see-quick

systemtest/src/main/java/io/strimzi/systemtest/kafkaclients/internalClients/VerifiableClient.java
```diff
@@ -330,9 +330,16 @@ private boolean allowParameter(String minimalVersion, String clientVersion) {
...
+            int major = Integer.parseInt(current.group("major")), 
+                minor = Integer.parseInt(current.group("minor")),
+                micro = Integer.parseInt(current.group("micro"));
+            int minMajor = Integer.parseInt(minimal.group("major")), 
+                minMinor = Integer.parseInt(minimal.group("minor")),
+                minMicro = Integer.parseInt(minimal.group("micro")); 
+            return 
+                    (major > minMajor)
+                    || (major == minMajor && minor > minMinor)
+                    || (major == minMajor && minor == minMinor && micro >= minMicro);
```
scholzj, line 342: Could you please provide more details why this is needed or desired? Doesn't have to be a comment in the code, but PR description would be nice.

fvaleri: Sure, let me update the description.

systemtest/src/main/java/io/strimzi/systemtest/kafkaclients/internalClients/InternalKafkaClient.java
```diff
@@ -312,6 +312,9 @@ public String getPodName() {
 
         boolean hasPassed = consumerGroups.run(timeoutMs);
         LOGGER.info("ConsumerGroups finished correctly: {}", hasPassed);
+        if (!hasPassed) {
+            throw new RuntimeException("VerifiableClient has failed. Check stderr for more details.");
```
see-quick, line 316: Is it possible that here the race condition can occur? Did you try to execute this test 4-5 times? It's passing?

see-quick (commented):
Thanks for this PR :) I have just one question.

systemtest/src/main/java/io/strimzi/systemtest/kafkaclients/internalClients/InternalKafkaClient.java
fvaleri: Thanks for the review @see-quick. 

Yes I did, and it looks fine. In ColdBackupScriptST I create my own instance of InternalKafkaClient, so I don't see any race condition. I also execute all acceptance STs locally. All green, apart a couple of errors due to timeouts. If you have any additional test in mind, please let me know and I'll run.

I see that there are other build errors here, but I don't think they are related to my changes. WDYT?

fvaleri: Unit and integration test run on my machine:

```shell
$ mvn -e -V -B -Dmaven.javadoc.skip=true -Dsurefire.rerunFailingTestsCount=5 -Dfailsafe.rerunFailingTestsCount=2 install
...
[INFO] Reactor Summary for Strimzi - Apache Kafka on Kubernetes and OpenShift 0.27.0-SNAPSHOT:
[INFO] 
[INFO] Strimzi - Apache Kafka on Kubernetes and OpenShift . SUCCESS [  1.914 s]
[INFO] kafka-agent ........................................ SUCCESS [  0.929 s]
[INFO] mirror-maker-agent ................................. SUCCESS [  0.171 s]
[INFO] tracing-agent ...................................... SUCCESS [  0.158 s]
[INFO] test ............................................... SUCCESS [  0.351 s]
[INFO] crd-annotations .................................... SUCCESS [  0.686 s]
[INFO] crd-generator ...................................... SUCCESS [  3.022 s]
[INFO] api ................................................ SUCCESS [02:07 min]
[INFO] mockkube ........................................... SUCCESS [  3.728 s]
[INFO] config-model ....................................... SUCCESS [  0.648 s]
[INFO] config-model-generator ............................. SUCCESS [ 15.968 s]
[INFO] certificate-manager ................................ SUCCESS [  5.392 s]
[INFO] operator-common .................................... SUCCESS [02:48 min]
[INFO] topic-operator ..................................... SUCCESS [30:02 min]
[INFO] cluster-operator ................................... SUCCESS [07:31 min]
[INFO] user-operator ...................................... SUCCESS [ 11.128 s]
[INFO] kafka-init ......................................... SUCCESS [  2.499 s]
[INFO] systemtest ......................................... SUCCESS [  0.747 s]
[INFO] ------------------------------------------------------------------------
[INFO] BUILD SUCCESS
[INFO] ------------------------------------------------------------------------
[INFO] Total time:  43:18 min
[INFO] Finished at: 2021-12-19T16:20:03+01:00
[INFO] ------------------------------------------------------------------------
```

scholzj:
> The InternalKafkaClient uses the VerifiableClient.allowParameter() to choose between --bootstrap-server and the deprecated --broker-list parameters. Recently we moved from Kafka 2.x to Kafka 3.x, and the following expression used in that method does not work anymore.

Sorry, but ... I still don't get it. Why did it worked before and doesn't work now? 

But also:
* If it does not work for Kafka 3.0.0, why are no tests failing because of it?
* Do we still need any support for the `--broker-list` option? I wonder if it can be just removed.

fvaleri:
@scholzj 

> Sorry, but ... I still don't get it. Why did it worked before and doesn't work now?

As part of the backup/restore checks, we call `getCurrentOffsets()` two times inside the ColdBackupScriptST. Due to the above bugs, we are getting an empty offsets map in both cases, instead of a runtime exception. With two empty maps, the following assertion still pass:

```
assertThat("Current consumer group offsets are not the same as before the backup", offsetsAfterRestore, is(offsetsBeforeBackup));
```

In addition to fixing the ST framework bugs, I added an extra check (offsets map should no be empty).

> If it does not work for Kafka 3.0.0, why are no tests failing because of it?

That's because currently only the `ColdBackupScriptST` is using the `CLI_KAFKA_CONSUMER_GROUPS="/opt/kafka/bin/kafka-consumer-groups.sh"`. In Kafka 3 `kafka-verifiable-consumer.sh` and `kafka-verifiable-producer.sh` support both `--bootstrap-server` and `--broker-list` (deprecated), while `kafka-consumer-groups.sh` only supports `--bootstrap-server `.

> Do we still need any support for the --broker-list option? I wonder if it can be just removed.

I think it is needed  for testing with Kafka versions older than 2.5.0.

scholzj:
> > Do we still need any support for the --broker-list option? I wonder if it can be just removed.
>
> I think this is needed for testing with Kafka versions older than 2.5.0.

Right. But we do not test any versions older than 2.5.0. We support now only 2.8.0 and newer in the STs. So it sounds like it can be just removed?

fvaleri:
> So it sounds like it can be just removed?

Well, in the same class there is also a check against 2.3.0. Probably @see-quick and @Frawless have more background about this. I think we can eventually address this in another PR. Would this be fine?

scholzj:
> Well, in the same class there is also a check against 2.3.0. Probably @see-quick and @Frawless have more background on this. I think we can eventually address this in another PR. Would this be fine?

Well, we do not support 2.3.0 for a long time. So I think that is all some old leftover and can be removed. But sure, we can wait for @see-quick and @Frawless to confirm.

scholzj:
@see-quick @Frawless As we are trying to not use the internal clients and the test-client image anyway and use the test clients from the separate repository, maybe there is more work needed around this as well?

see-quick:
> So I think that is all some old leftover and can be removed

Yes, that's true. :) I think this piece of code can be removed:
```java
this.clientArgumentMap.put(allowParameter("2.5.0", clientVersion) ? ClientArgument.BOOTSTRAP_SERVER : ClientArgument.BROKER_LIST, bootstrapServer);
``` 
and also this one:
```java
if (allowParameter("2.3.0", clientVersion)) {
                this.consumerInstanceId = verifiableClientBuilder.consumerInstanceId;
                this.clientArgumentMap.put(ClientArgument.GROUP_INSTANCE_ID, this.consumerInstanceId);
            }
```

Of course, the best way is to use [`test-clients`](https://github.com/strimzi/test-clients) but I think in test-clients we should do more work on the additional method (`getOffsets`).

fvaleri:
Ok, I'll remove them as part of this PR.

systemtest/src/main/java/io/strimzi/systemtest/kafkaclients/internalClients/VerifiableClient.java
```diff
@@ -330,9 +321,16 @@ private boolean allowParameter(String minimalVersion, String clientVersion) {
...
```
scholzj, line 333: So, is this still used from anywhere else? It seems to be used only from the two places which you removed.

fvaleri: Nope, do you also prefer to get rid of that?

scholzj: Yeah, I think we should delete it instead of changing the code.

fvaleri: Ok, but I changed it before, to make it work :)

scholzj (approved):
Thanks for the PR.

see-quick (approved):
Thanks for the PR

scholzj:
Thanks for the PR!

Issue #6058, opened 2021-12-17
The `strimzi_resources{kind="Kafka"}` metric is not removed when the Kafka related namespace is deleted


ppatierno:
Assuming to have a cluster operator deployment (version 0.26.0) watching ALL namespaces on a Kubernetes/OpenShift cluster, it seems that the `strimzi_resources{kind="Kafka"}` metric is not deleted when when the namespace where the `Kafka` CR lives is deleted (so the user deleted the resource and then the namespace).

It seems to be related to how these metrics are exported on each periodic reconcile here:

https://github.com/strimzi/strimzi-kafka-operator/blob/main/operator-common/src/main/java/io/strimzi/operator/common/Operator.java#L74

If across two reconcile cycles, a namespace with a `Kafka` CR is deleted so the corresponding Kafka instance, the namespace is not anymore in the `desiredNames` list and the corresponding metric is not removed; it will hang forever as a "zombie".

scholzj:
> If across two reconcile cycles, a namespace with a Kafka CR is deleted so the corresponding Kafka instance, the namespace is not anymore in the desiredNames list and the corresponding metric is not removed; it will hang forever as a "zombie".

I do not think you need to delete the namespace. It should be enough if there is no other Kafka cluster in that namespace. Then the desiredNames will not have the namespace even if it still exists.

ppatierno:
The use case I had was about deleting the namespace directly hence the `Kafka` CR was deleted, but I do agree that it doesn't matter.

PR #6070, opened 2021-12-20
Fix decrement metric when resource is deleted

sknot-rh:

### Type of change
- Bugfix

### Description
If all namespaces are watched, one verticle is used so we can reset all the records in the metrics map. If multiple namespaces are used, each of them has to reset its own part of the metrics map.

Fixes https://github.com/strimzi/strimzi-kafka-operator/issues/6058

### Checklist

- [ ] Write tests
- [ ] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/operator-common/src/main/java/io/strimzi/operator/common/AbstractOperator.java
+++ b/operator-common/src/main/java/io/strimzi/operator/common/AbstractOperator.java
@@ -252,6 +252,12 @@ public abstract class AbstractOperator<
 
                 return createOrUpdate.future();
             } else {
+                if (resourceCounter(namespace).get() > 0) {
+                    resourceCounter(namespace).getAndDecrement();
+                }
+                if (pausedResourceCounter(namespace).get() > 0) {
+                    pausedResourceCounter(namespace).getAndDecrement();
+                }
                 LOGGER.infoCr(reconciliation, "{} {} should be deleted", kind, name);
                 return delete(reconciliation).map(deleteResult -> {
                     if (deleteResult) {
--- a/operator-common/src/main/java/io/strimzi/operator/common/AbstractOperator.java
+++ b/operator-common/src/main/java/io/strimzi/operator/common/AbstractOperator.java
@@ -107,6 +107,12 @@ public abstract class AbstractOperator<
     }
 
     @Override
+    public void resetCounters() {
+        resourceCounterMap.entrySet().forEach(entry -> entry.getValue().set(0));
+        pausedResourceCounterMap.entrySet().forEach(entry -> entry.getValue().set(0));
+    }
+
+    @Override
     public String kind() {
         return kind;
     }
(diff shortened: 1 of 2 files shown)


operator-common/src/main/java/io/strimzi/operator/common/Operator.java
```diff
@@ -36,6 +36,8 @@
  */
 public interface Operator {
 
+    static final ReconciliationLogger LOGGER = ReconciliationLogger.create(Operator.class);
```
scholzj, line 39: Why is this needed? You do not seem to use it.

operator-common/src/main/java/io/strimzi/operator/common/Operator.java
```diff
@@ -71,16 +72,19 @@ default void reconcileAll(String trigger, String namespace, Handler<AsyncResult<
...
 
             for (NamespaceAndName resourceRef : desiredNames) {
-                resourceCounter(resourceRef.getNamespace()).getAndIncrement();
                 Reconciliation reconciliation = new Reconciliation(trigger, kind(), resourceRef.getNamespace(), resourceRef.getName());
                 futures.add(reconcile(reconciliation));
             }
             CompositeFuture.join(futures).map((Void) null).onComplete(handler);
         } else {
             resourceCounter(namespace).set(0);
+            pausedResourceCounter(namespace).set(0);
```
scholzj, line 87: How does this work with operator watching all namespaces? Line 76-77 will reset it for namespaces which have some resources. But these two lines would reset it for namespace `*` I suspect.

scholzj (commented):
TBH, I'm not sure this really fixes it. I would expect that you will have to stream through the metrics map and reset it for each namespace that way.

operator-common/src/main/java/io/strimzi/operator/common/AbstractOperator.java
```diff
@@ -175,6 +175,7 @@ public String kind() {
                     return Future.succeededFuture();
                 }
 
+                resourceCounter(namespace).getAndIncrement();
```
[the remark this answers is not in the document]
sknot-rh: Ah, I did not know the metric should not be updated when it is created/removed.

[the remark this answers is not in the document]
scholzj: The problem is that you cannot count it. You might get modified multiple times between the timer based reconciliations.

operator-common/src/main/java/io/strimzi/operator/common/Operator.java
sknot-rh: Correct

operator-common/src/main/java/io/strimzi/operator/common/Operator.java
```diff
@@ -81,6 +86,7 @@ default void reconcileThese(String trigger, Set<NamespaceAndName> desiredNames,
...
```
scholzj, line 89: when running cluster wide operator, this would create metrics with `namespace=*`. I do not think we want that. I guess the question is whether we want to emit the metrics of 0 resources (since I guess no metric would also count as 0). And if we do, we should again if-it to set this only when it doesn't watch all namespaces.

scholzj: Actually, these two lines 88+89 are not needed at all, or? They just set the 0 again already set on 72-77.

scholzj (approved):

ppatierno (approved):
LGTM

PR #6078, opened 2021-12-21
Fixed CO passing JVM options to the bridge

### LINKED ISSUE #6075: Cluster operator missing Java options to be passed to Kafka bridge
The cluster operator, through the `KafkaBridgeCluster` model, just sets the GC logging enabled and Java System properties on the bridge (from the `JvmOptions` API) but it doesn't provide other Java optiojns like Xmx, Xms and XX.
We already fixed the bridge on https://github.com/strimzi/strimzi-kafka-bridge/issues/571 in order to get `STRIMZI_JAVA_SYSTEM_PROPERTIES` and `STRIMZI_JAVA_OPTS` but the last one is not actually filled by cluster operator.

ppatierno:
### Type of change

- Bugfix

### Description

This PR fixes #6075 having the CO passing the Xmx, Xms and XX Java options to the bridge configuration via the `STRIMZI_JAVA_OPTS` env var and the Java system properties via `STRIMZI_JAVA_SYSTEM_PROPERTIES` env var.

### Checklist

- [x] Write tests


cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaBridgeCluster.java
```diff
@@ -167,6 +167,7 @@ public static KafkaBridgeCluster fromCrd(Reconciliation reconciliation, KafkaBri
         if (spec.getJvmOptions() != null) {
             kafkaBridgeCluster.setJavaSystemProperties(spec.getJvmOptions().getJavaSystemProperties());
         }
```
scholzj, line 169: This is not needed anymore I guess?

ppatierno: The corresponding field `javaSystemProperties` is still used here (line 350):

```java
ModelUtils.javaOptions(varList, getJvmOptions(), javaSystemProperties);
```

but of course I can avoid to set the field and passing directly it from the jvm options if not null.
Taking a quick look we have this kind of way to handle it across all the other models (i.e. EntityTopicOperator as well).

scholzj: I don't know. It does already quite a lot of refactoring here. But ok I guess.

scholzj (approved):

PR #6121, opened 2021-12-29
Add support for Kafka 3.1.0

scholzj:
### Type of change

- Enhancement / new feature

### Description

This PR adds support for Kafka 3.1.0 and removes Kafka 2.8.0 and 2.8.1.

### Checklist

- [x] Write tests
- [x] Update documentation
- [x] Update CHANGELOG.md


laurovenancio:
Why to remove the kafka 2.8 support? It would be available just two versions (3.0.0 and 3.1.0) without minor versions fixes. It is risky to use them in a production environment. It would be great if strimzi could support images until the second minor version (2.8.2, 3.0.2, 3.1.2).

scholzj:
@laurovenancio The Strimzi policy with regards to Kafka versions is to support the 2 latest minor versions => which right now will be 3.0 and 3.1. We do not have the resources to support and maintain the code for all the different Kafka versions. There is also no guarantee about the timing of the Kafka patch releases which makes it basically impossible to commit to anything like that.

scholzj:
@strimzi/maintainers This should be now ready for review. The system test failures are not related to this PR.

ppatierno (approved):

Frawless (approved):

Issue #6107, opened 2021-12-24
The strimzi_resources{kind="Kafka"} metric is not removed when the Kafka related namespace is deleted and CO is watching specific namespaces


ppatierno:
It seems that the #6058 is not totally fixed.
I am running a CO watching a couple of namespaces, so configured this way:

```yaml
env:
            - name: STRIMZI_NAMESPACE
              value: "kafka-1,kafka-2"
```

Creating two Kafka instances (one for each namespace), everything works fine and the corresponding `strimzi_resources` metric is updated as expected:

```shell
strimzi_resources{kind="Kafka",namespace="kafka-2",selector="",} 1.0
strimzi_resources{kind="Kafka",namespace="kafka-1",selector="",} 1.0
```

By deleting the `Kafka` CR in the `kafka-1` namespace for example, the corresponding metric is updated as expected to 0 but if the namespace is deleted directly, it's not and still stays at 1.0.
At some time the CO is printing the following errors which kind of makes sense because the namespace is not there anymore

```shell
2021-12-24 09:26:23 INFO  OperatorWatcher:38 - Reconciliation #53(watch) Kafka(kafka-1/my-cluster): Kafka my-cluster in namespace kafka-1 was DELETED
2021-12-24 09:26:23 INFO  AbstractOperator:261 - Reconciliation #53(watch) Kafka(kafka-1/my-cluster): Kafka my-cluster should be deleted
2021-12-24 09:26:23 INFO  AbstractOperator:266 - Reconciliation #53(watch) Kafka(kafka-1/my-cluster): Assembly my-cluster or some parts of it will be deleted by garbage collection
2021-12-24 09:26:23 INFO  AbstractOperator:517 - Reconciliation #53(watch) Kafka(kafka-1/my-cluster): reconciled
2021-12-24 09:26:59 WARN  WatcherWebSocketListener:73 - Exec Failure: HTTP 403, Status: 403 - kafkamirrormakers.kafka.strimzi.io is forbidden: User "system:serviceaccount:kafka:strimzi-cluster-operator" cannot watch resource "kafkamirrormakers" in API group "kafka.strimzi.io" in the namespace "kafka-1"
[...]

PR #6133, opened 2022-01-03
Reset counters when namespace is not reconcilable

sknot-rh:

### Type of change
- Bugfix

### Description
If the namespace is removed, the CO is not able to set the correct resource count (0). Checking the result of `allResourceNames` for failure can be used for such a case.

Fixes https://github.com/strimzi/strimzi-kafka-operator/issues/6107
### Checklist
- [ ] Write tests
- [ ] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/operator-common/src/main/java/io/strimzi/operator/common/Operator.java
+++ b/operator-common/src/main/java/io/strimzi/operator/common/Operator.java
@@ -61,6 +61,8 @@ public interface Operator {
                 reconcileThese(trigger, ar.result(), namespace, handler);
                 periodicReconciliationsCounter(namespace).increment();
             } else {
+                resourceCounter(namespace).set(0);
+                pausedResourceCounter(namespace).set(0);
                 handler.handle(ar.map((Void) null));
             }
         });
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/ClusterOperator.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/ClusterOperator.java
@@ -105,7 +105,7 @@ public class ClusterOperator extends AbstractVerticle {
                 kafkaConnectAssemblyOperator, kafkaBridgeAssemblyOperator, kafkaMirrorMaker2AssemblyOperator));
         for (AbstractOperator<?, ?, ?, ?> operator : operators) {
             watchFutures.add(operator.createWatch(namespace, operator.recreateWatch(namespace)).compose(w -> {
-                LOGGER.info("Opened watch for {} operator", operator.kind());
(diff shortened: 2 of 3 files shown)


scholzj (commented):
I'm personally not sure we should be solving this. This is a user configuration error and I do not think the metrics really matter in that case. But if nothing else, you should add unit tests for this.

operator-common/src/main/java/io/strimzi/operator/common/Operator.java
```diff
@@ -61,6 +61,8 @@ default void reconcileAll(String trigger, String namespace, Handler<AsyncResult<
                 reconcileThese(trigger, ar.result(), namespace, handler);
                 periodicReconciliationsCounter(namespace).increment();
             } else {
+                resourceCounter(namespace).set(0);
+                pausedResourceCounter(namespace).set(0);
```
scholzj, line 65: So, how does this work with:
* Failures to all-namespaces operator? It will IMHO at any failure set the metrics for namespace `*` to 0 and never remove them. And at the same time it will keep the old metrics for namespaces which were maybe deleted.
* How do you know the count is 0 because the namespace was deleted? How do you know it wasn't just some temp glitch in Kube API or RBAC issue but exactly deleted namespace? You should probably remove the metrics rarther than setting them to 0.

I think that by fixing one very edge case you break several other very edge cases? I really think that we should accept that when the operator is broken due to things such as RBAC issues or missing namespaces etc., it will not provide reliable metrics since it does not really know the actual values. Otherwise you might take forever fixing all weird cases.

CC @ppatierno 

_As a sidenote => in few months, nobody will remember what was this piece of code do or the test you added test. You should definitely explain in places like this in comments why are they there._

ppatierno: I would agree with Jakub on this. Handling the scenario here means catching not just the case I raised but even something more ... and the cases could be a lot. 
I thought more about it:

* the metrics seem to be reset when the CO is configured to watch ALL namespaces but you delete one. So it's fine.
* if you instead configure the CO to watch specific namespaces but then delete one you are kind of breaking CO configuration so I agree with Jakub, we have just to accept it.

In the end we could just close the issue with no fix.

sknot-rh: Yes, I agree with closing this. Should the CO at least print some warnings if the code flow goes thtough https://github.com/strimzi/strimzi-kafka-operator/blob/main/operator-common/src/main/java/io/strimzi/operator/common/Operator.java#L64 ?

scholzj: I would expect the errors to be logged on other level. But you would need to double check it if you want.

PR #6138, opened 2022-01-04
[systemtest] Separate Kafka version of test-clients from broker Kafka version

im-konge:

### Type of change

- Enhancement / new feature

### Description

Until now we had only one Kafka version set in STs - via `ST_KAFKA_VERSION`. We also set the used Kafka version in test-clients with this variable, which can be a problem when we want to run tests against new (and not released) Kafka version, for which we don't have test-clients built.

This PR separates Kafka version of test-clients from version used for Kafka brokers.

### Checklist

### DIFF
--- a/systemtest/src/main/java/io/strimzi/systemtest/Environment.java
+++ b/systemtest/src/main/java/io/strimzi/systemtest/Environment.java
@@ -138,7 +138,6 @@ public class Environment {
     /**
      * Defaults
      */
-    private static final String ST_KAFKA_VERSION_DEFAULT = TestKafkaVersion.getDefaultSupportedKafkaVersion();
     public static final String STRIMZI_ORG_DEFAULT = "strimzi";
     public static final String STRIMZI_TAG_DEFAULT = "latest";
     public static final String STRIMZI_REGISTRY_DEFAULT = "quay.io";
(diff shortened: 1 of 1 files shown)


systemtest/src/main/java/io/strimzi/systemtest/Environment.java
```diff
@@ -153,6 +157,8 @@
     private static final ClusterOperatorInstallType CLUSTER_OPERATOR_INSTALL_TYPE_DEFAULT = ClusterOperatorInstallType.BUNDLE;
     private static final boolean LB_FINALIZERS_DEFAULT = false;
     private static final String STRIMZI_FEATURE_GATES_DEFAULT = "";
+
+    private static final String CLIENTS_KAFKA_VERSION_DEFAULT = "3.0.0";
```
Frawless, line 161: Maybe you can move `ST_KAFKA_VERSION` next to it and also add prefix `ST_` ? Or at least move it

im-konge: Okay, I wanted test-clients relative defaults to be together. But I'll move it and rename it, thanks :)

Frawless (approved):

scholzj (approved):

Issue #3636, opened 2020-09-13
[Bug] KafkaRebalance resource created before CC is enabled is never accepted


scholzj:
When a `KafkaRebalance` resource is created while Cruise Control is not enabled in `Kafka` CR, CO throws a warning that CC needs to be enabled and this is also set in the `KafkaRebalance` status. However, after CC is enabled in `Kafka` CR and deployed, it should pick up the `KafkaRebalance` resource in periodical reconciliation and use it for `KafkaRebalance`. This does not seem to happen currently.

ppatierno:
Yeah what I see is that it's reconciled but in a wrong way. The `KafkaRebalance` operator reconcile the resource but doesn't take any meaningful action and leaves the last NotReady state as follow.

```shell
2020-10-08 10:02:12 INFO  AbstractOperator:217 - Reconciliation #41(timer) KafkaRebalance(default/my-rebalance): KafkaRebalance my-rebalance will be checked for creation or modification
2020-10-08 10:02:12 INFO  KafkaRebalanceAssemblyOperator:319 - Reconciliation #41(timer) KafkaRebalance(default/my-rebalance): Rebalance action from state [NotReady]
2020-10-08 10:02:12 INFO  KafkaRebalanceAssemblyOperator:333 - Reconciliation #41(timer) KafkaRebalance(default/my-rebalance): State updated to [NotReady] with annotation strimzi.io/rebalance=null 
2020-10-08 10:02:12 INFO  AbstractOperator:455 - Reconciliation #41(timer) KafkaRebalance(default/my-rebalance): reconciled
```

I will look into this.

ppatierno:
So actually, it's not a bug but how it was designed. 
I mean, when the resource it's in `NotReady` state, the user has to annotate with `strimzi.io/rebalance=refresh` for pushing the operator to reconsider the rebalance resource once the error has been fixed.
I guess we did it on purpose for not waiting for a reconcile cycle when the error is not in the resource itself but kind of external like in your case the CC not deployed at the beginning.
Without the annotation, the rebalance resource would be reconciled immediately only if there is an error on it and the user updates the resource itself triggering the watch; it could immediately move from `NotReady` to `New` to be reprocessed.
But when the error is kind of external, the only way is waiting for the next reconcile cycle or doing a change on the resource (which doesn't really need), in this case with an annotation.
I see this as the only reason for the need for the "refresh" annotation otherwise we can just say the user to wait for the next reconcile.

[...]

PR #6146, opened 2022-01-04
Remove the `refresh` annotation when CC is not enabled or Cluster label is missing/wrong.

ShubhamRwt:

### Type of change

_Select the type of your PR_

- Bugfix

### Description

This PR address the issue mentioned in #3636. This removes the use of `refresh` annotation when we apply KR and CC is not enabled or the cluster label is missing/wrong.

Code Breakdown :-

- This code basically uses an extra variable  `disabled` which acts as a switch.
- It gets true when we hit the cases like CC is not enabled or maybe cluster label was missing etc.
- The `disabled` variable gets `true`, we can get a new CC request in the next periodic reconcile and if the request is performed and CC is also up then we can again set disabled as `false` since now CC is Up and running/ we have made the labels correct  

Alternative :

- We can also have a separate state named `Disabled` to resolve the problem 

### Checklist

- [ ] Write tests
- [ ] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceAssemblyOperator.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceAssemblyOperator.java
@@ -154,6 +154,7 @@ public class KafkaRebalanceAssemblyOperator
     private final SecretOperator secretOperations;
     private final PlatformFeaturesAvailability pfa;
     private final Optional<LabelSelector> kafkaSelector;
+    private static boolean disabled;
 
     private final ConfigMapOperator configMapOperator;
     /**
@@ -171,6 +172,7 @@ public class KafkaRebalanceAssemblyOperator
         this.kafkaOperator = supplier.kafkaOperator;
         this.configMapOperator = supplier.configMapOperations;
         this.secretOperations = supplier.secretOperations;
+        disabled = false;
     }
 
     /**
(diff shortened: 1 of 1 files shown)


ShubhamRwt:
@ppatierno @kyguy @scholzj Hello, I have tried to remove the `refresh` annoation without the use of an extra and just use a `variable` which can act as a switch. I needed you suggestions regarding this approach if it looks fine or maybe I should resolve it in the alternative manner

scholzj (commented):
This seems wrong. There is only one instance of KafkaRebalanceAssemblyOperator per installation. So it is shared for different namespaces and different Kafka clusters. So I don't think you can use a single static field to store any information.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceAssemblyOperator.java
```diff
@@ -711,8 +713,10 @@ public MapAndStatus(T loadMap, K status) {
                                                     KafkaRebalance kafkaRebalance,
                                                     KafkaRebalanceAnnotation rebalanceAnnotation,
                                                     RebalanceOptions.RebalanceOptionsBuilder rebalanceOptionsBuilder) {
-        if (rebalanceAnnotation == KafkaRebalanceAnnotation.refresh) {
-            // The user has fixed the error on the resource and want to 'refresh'
+
+        if (rebalanceAnnotation == KafkaRebalanceAnnotation.refresh || disabled) {
```
kyguy, line 717: Wouldn't we want this conditional to be:
```
rebalanceAnnotation == KafkaRebalanceAnnotation.refresh && !disabled
``` 
So that a new request isn't made when CC is disabled?

kyguy (commented):
From what I understand, we want the KafkaRebalanceAssemblyOperator to call `onNew()` once CC is up and running, we don't want to wait for a `refresh` annotation to be supplied unless the KafkaRebalance resource is malformed. But I would start with picking up the `KafkaRebalance` resource once CC is ready and expand upon there!

ShubhamRwt:
Thanks @scholzj and @kyguy for the suggestions. I will try the other alternative approach since it might be more better and good way to respond towards this issue.

scholzj:
@ShubhamRwt Should we close this then and open a new PR once you have the new code?

ShubhamRwt:
Sure @scholzj , I will create a new one by today probably.

Issue #6145, opened 2022-01-04
extraEnvs doesn't work for cluster operator helm charts


spenczar:
**Describe the bug**

`extraEnvs` is intended to set extra environment variables used by the cluster operator when deploying with Helm charts. The implementation in #5859 is incorrect - it uses `toYaml` which raises a confusing error, possibly due to typing issues with the Helm YAML.

**To Reproduce**

0. (if this is the first time using Strimzi with Helm) set up your helm system to use strimzi charts:

    ```
    helm repo add strimzi https://strimzi.io/charts/
   ```

1. Update Strimzi repo

    ```
    helm repo update
    ```

2. Use this `values.yaml`:

    ```yaml
    extraEnvs:
      - name: JAVA_OPTS
        value: "-Xms=256m -Xmx=256m"
    ```

3. Run this command:

    ```
    helm template strimzi strimzi/strimzi-kafka-operator --dry-run --version 0.27.0 --debug --values valules.yaml
    ```

**Expected behavior**

Rendered resource definitions in YAML, and a zero exit code, including the following deployment:

```yaml
# Source: strimzi-kafka-operator/templates/060-Deployment-strimzi-cluster-operator.yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: strimzi-cluster-operator
  labels:
    app: strimzi
    chart: strimzi-kafka-operator-0.27.0
    component: deployment
    release: strimzi-release
    heritage: Helm
spec:
  replicas: 1
  selector:
    matchLabels:
      name: strimzi-cluster-operator
      strimzi.io/kind: cluster-operator
  template:
    metadata:
      labels:
        name: strimzi-cluster-operator
        strimzi.io/kind: cluster-operator
    spec:
      serviceAccountName: strimzi-cluster-operator
      volumes:
        - name: strimzi-tmp
          emptyDir:
            medium: Memory
            sizeLimit: 1Mi
        - name: co-config-volume
          configMap:
            name: strimzi-cluster-operator
      containers:
        - name: strimzi-cluster-operator
          image: quay.io/strimzi/operator:0.27.0
          ports:
            - containerPort: 8080
              name: http
          args:
            - /opt/strimzi/bin/cluster_operator_run.sh
          volumeMounts:
            - name: strimzi-tmp
              mountPath: /tmp
            - name: co-config-volume
              mountPath: /opt/strimzi/custom-config/
          env:
            - name: STRIMZI_NAMESPACE
[...]

PR #6147, opened 2022-01-05
[Fix] extraEnvs doesn't work for cluster operator helm charts 

ShubhamRwt:

_Select the type of your PR_

- Bugfix

### Description

This PR fixes the issue mentioned in #6145. It removes the use the use if `toYaml` which was giving error and makes use of `range`.

### Checklist

- [ ] Write tests
- [ ] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/packaging/helm-charts/helm3/strimzi-kafka-operator/templates/060-Deployment-strimzi-cluster-operator.yaml
+++ b/packaging/helm-charts/helm3/strimzi-kafka-operator/templates/060-Deployment-strimzi-cluster-operator.yaml
@@ -129,7 +129,10 @@ spec:
               value: {{ .Values.connectBuildTimeoutMs | quote }}
             {{- end }}
             {{- if .Values.extraEnvs }}
-            {{ toYaml .Values.extraEnvs | indent 12 }}
+            {{- range $key, $val := .Values.extraEnvs }}
+            - name: {{ $val.name }}
+              value: {{ $val.value | quote }}
+            {{- end }}
             {{- end }}
           livenessProbe:
             httpGet:
--- a/packaging/helm-charts/helm3/strimzi-kafka-operator/templates/060-Deployment-strimzi-cluster-operator.yaml
+++ b/packaging/helm-charts/helm3/strimzi-kafka-operator/templates/060-Deployment-strimzi-cluster-operator.yaml
@@ -129,10 +129,7 @@ spec:
               value: {{ .Values.connectBuildTimeoutMs | quote }}
             {{- end }}
             {{- if .Values.extraEnvs }}
-            {{- range $key, $val := .Values.extraEnvs }}
-            - name: {{ $val.name }}
-              value: {{ $val.value | quote }}
-            {{- end }}
+{{ toYaml .Values.extraEnvs | indent 12 }}
             {{- end }}
           livenessProbe:
             httpGet:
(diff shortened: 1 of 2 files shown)


scholzj (commented):
Could you please also add an example of the value to the README file or add it commented out to the values.yaml file?

This was of doing it also doesn't allow to reference secrets or confimaps etc. -> is that something we want? If we don't want it, does the format for passing the values still make sense? Would map be easier to use?

ShubhamRwt:
@scholzj I tried to use `toYaml` with `nindent 12` and I was able to get through the error and get something like this 

```            - name: STRIMZI_FEATURE_GATES
              value: ""
            
            - name: JAVA_OPTS
              value: -Xms=256m -Xmx=256m
```

ShubhamRwt:
I wonder why `indent` is not working fine while `nindent` is. It might be because
 ```           
            - name: STRIMZI_FEATURE_GATES
              value: {{ .Values.featureGates | quote }} ``` . We are not checking any `if` here and we are getting feature gates in our resources even when we dont use any

ShubhamRwt:
I guess I got the fix

ShubhamRwt:
Sorry looks like I made a mistake

scholzj:
> We are not checking any `if` here and we are getting feature gates in our resources even when we dont use any

That alone is not wrong. We do not have to do if it makes things easier. But the empty value means the default behaviour => so it is not a problem on its own.

ShubhamRwt:
Got it @scholzj . The changes I have pushed resolves the error which was coming.

spenczar (approved):
Duh, of course! Thanks for the fix.

`nindent` makes this a little cleaner to read, in my opinion, but it's no big deal.

ShubhamRwt:
> Duh, of course! Thanks for the fix.
> 
> `nindent` makes this a little cleaner to read, in my opinion, but it's no big deal.

`nindent` will add an extra line too.

```- name: STRIMZI_FEATURE_GATES
              value: ""
            
            - name: JAVA_OPTS
              value: -Xms=256m -Xmx=256m
```

I guess we might not need it. What do you suggest @scholzj ?

scholzj:
> I guess we might not need it. What do you suggest @scholzj ?

I do not know Helm, so I have no idea.

scholzj:
@ppatierno Can you have a look at this if you have few minutes? Thanks.

packaging/helm-charts/helm3/strimzi-kafka-operator/templates/060-Deployment-strimzi-cluster-operator.yaml
```diff
@@ -129,7 +129,7 @@ spec:
               value: {{ .Values.connectBuildTimeoutMs | quote }}
             {{- end }}
             {{- if .Values.extraEnvs }}
-            {{ toYaml .Values.extraEnvs | indent 12 }}
+{{ toYaml .Values.extraEnvs | indent 12 }}
```
ppatierno, line 132: not an Helm expert but why the change is about changing the indentation of this line?

ShubhamRwt: It is due to the `indent 12`. I guess it starts counting the indentation space from the start

ppatierno (approved):

rajha-korithrien:
Is not the proper Helm way to deal with this to use the "-" operator after the `{{`?

Notice that there is a difference between `{{` and `{{-`.

I believe that `{{-` instructs helm to remove all the leading whitespace of this line when rendering the template. Doing so will allow us to retain the 12 spaces while causing the results of the toYaml function to be correctly rendered into the resulting deployment yaml file.

PR #6151, opened 2022-01-06
docs: removes duplicate procedures from deploying guide

PaulRMellor:

**Documentation**

Removes the following procedures from _Deploying_ guide.

- Performing a restart of a Kafka connector
- Performing a restart of a Kafka connector task

The procedures are present in the _Using_ (Configuring) guide
Links to the procedures replaces the duplication.

- [ ] Write tests
- [x] Update documentation
- [ ] Update CHANGELOG.md


scholzj (commented):
Well, if we are renaming the Using guide to Configuring guide => doesn't this make more sense in the Deploying guide? It is not really a configuration, it is an control action. Or what is the split what does into which guide?

PaulRMellor:
> Well, if we are renaming the Using guide to Configuring guide => doesn't this make more sense in the Deploying guide? It is not really a configuration, it is an control action. Or what is the split what does into which guide?

We spoke about this and decided on this way around. _Deploying_ is the doc for deploying the components. The Configuring guide has the content for configuring, where we show how to enable KafkaConnectors. The guide also has content for managing a deployment too.  That said, it's not completely obvious where these procedures should go. If we keep in the Configuring guide, maybe I can make it clearer in the prereqs that the procedure applies if you've enabled KafkaConnector and used it to deploy connectors. And maybe we drop it further down in the guide to the _Managing_section?

scholzj (approved):
Ok, I guess I forgot about it then. Still wonder if it is really intuitive.

PR #6158, opened 2022-01-08
[DOC] Fix secret names in security documentation

fvaleri:
This is required after PR #6006.

### DIFF
--- a/documentation/modules/configuring/ref-list-of-kafka-cluster-resources.adoc
+++ b/documentation/modules/configuring/ref-list-of-kafka-cluster-resources.adoc
@@ -60,7 +60,8 @@ These resources are only created if the Entity Operator is deployed using the Cl
 `_cluster-name_-entity-operator-_random-string_`:: Pod created by the Entity Operator deployment.
 `_cluster-name_-entity-topic-operator-config`:: ConfigMap with ancillary configuration for Topic Operators.
 `_cluster-name_-entity-user-operator-config`:: ConfigMap with ancillary configuration for User Operators.
-`_cluster-name_-entity-operator-certs`:: Secret with Entity Operator keys for communication with Kafka and ZooKeeper.
+`_cluster-name_-entity-topic-operator-certs`:: Secret with Topic Operator keys for communication with Kafka and ZooKeeper.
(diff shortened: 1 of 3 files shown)


scholzj (commented):
The changes to `documentation/modules/security/ref-certificates-and-secrets.adoc` are good. 

But the changes to `documentation/modules/deploying/proc-deploy-user-operator-standalone.adoc` suggest that we broker the backwards compatibility. So we probably need to look into it if it is really needed. Also, if we keep it, similar changes would need to be done to the TO documentation I guess. We would also need to add it to the CHANGELOG.md.

Maybe the easiest way to do right now would be to just remove the changes to `documentation/modules/deploying/proc-deploy-user-operator-standalone.adoc` until we investigate it.

fvaleri:
> But the changes to documentation/modules/deploying/proc-deploy-user-operator-standalone.adoc suggest that we broker the backwards compatibility.

Well, I don't think we broke anything here. The value of `STRIMZI_EO_KEY_SECRET_NAME` is just an example, which happens to be equal to the old secret's name. Instead, in `proc-deploy-topic-operator-standalone.adoc` we use general examples like the following:

```
env:
  - name: STRIMZI_TRUSTSTORE_LOCATION (1)
    value: "/path/to/truststore.p12"
```

So maybe we can use something like `user-operator-certs-secret-name`, or leave it as it is. I'm fine with both.

scholzj:
> I don't think we broke anything here.

The secret name is not the problem. But the keys inside the secret break the compatibility.

fvaleri:
> The secret name is not the problem. But the keys inside the secret break
the compatibility.

You are right, we broke it here (changing from `entity-operator` to
`entity-user-operator`):

https://github.com/strimzi/strimzi-kafka-operator/blob/main/user-operator/src/main/java/io/strimzi/operator/user/Main.java#L128

CHANGELOG note added. We also need an update in
`ref-list-of-kafka-cluster-resources.adoc`. Should be all, but if you can
double check at your convenience it would be better.

scholzj:
Yeah, but the right thing to do is to fix it so that it is not breaking change. Adding something to CHANGELOG is an option only if it can be fixed which I doubt.

scholzj (approved):

Frawless (approved):

scholzj:
Thanks for the PR!

PR #6162, opened 2022-01-10
Custom authentication support

Sam-Kramer:
### Type of change

Enhancement / new feature

### Description

Adds support for custom authentication for Kafka deployed by the Strimzi operator. 

### Checklist

- [x] Write tests
- [x] Update documentation
- [ ] Update CHANGELOG.md


api/src/main/java/io/strimzi/api/kafka/model/listener/KafkaListenerAuthenticationCustom.java
```diff
@@ -0,0 +1,76 @@
...
+
+    public static final String FORBIDDEN_PREFIXES = "ssl.keystore.location, ssl.keystore.password, ssl.keystore.type";
+
+    private static final long serialVersionUID = 1L;
+
+    public static final String TYPE_CUSTOM = "custom";
+
+    private Map<String, Object> listenerConfig;
+    private boolean sasl = true;
+    private List<CertSecretSource> tlsTrustedCertificates;
```
Sam-Kramer, line 37: remove

api/src/main/java/io/strimzi/api/kafka/model/listener/KafkaListenerAuthenticationCustom.java
```diff
@@ -0,0 +1,76 @@
...
+/**
+ * Configures a listener to use mutual TLS authentication.
+ */
+@Buildable(
+        editableEnabled = false,
+        builderPackage = Constants.FABRIC8_KUBERNETES_API
+)
+@JsonInclude(JsonInclude.Include.NON_NULL)
+@EqualsAndHashCode
+public class KafkaListenerAuthenticationCustom extends KafkaListenerAuthentication {
```
Sam-Kramer, line 27: I decided to leave out `tlsTrustedCertificates` despite it being listed in the proposal, due to the primary reason being the user can easily? define the truststore themself, reference it as a secret, and specify it as a config.

api/src/main/java/io/strimzi/api/kafka/model/listener/KafkaListenerAuthenticationCustom.java
```diff
@@ -0,0 +1,76 @@
...
+
+    public static final String FORBIDDEN_PREFIXES = "ssl.keystore.location, ssl.keystore.password, ssl.keystore.type";
```
Sam-Kramer, line 29: Should we ban these prefixes, as they're generated by the parent Listener class?  This currently doesn't do anything, but wanted to add for discussion. 

Ultimately I _think_ we do want to do this, as we want to control the public/private key the same way as the other listeners. Although, this may be cumbersome for some users, as they may want full control over these configuration values.

api/src/main/java/io/strimzi/api/kafka/model/listener/KafkaListenerAuthenticationCustom.java
```diff
@@ -0,0 +1,76 @@
...
+    @JsonInclude(JsonInclude.Include.NON_NULL)
+    @Override
+    public String getType() {
+        return TYPE_CUSTOM;
+    }
+
+    @Description("Enable or disable SASL on this listener. Default value is true.")
+    @JsonInclude(JsonInclude.Include.NON_NULL)
+    public boolean isSasl() {
+        return sasl;
```
scholzj, line 50: I wonder if this should be required instead of defaulting to true 🤔 . It would make it more obvious I guess.

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaCluster.java
```diff
@@ -1443,6 +1446,11 @@ private void setDataVolumesClaimsAndMountPaths(Storage storage) {
                 KafkaListenerAuthenticationOAuth oauth = (KafkaListenerAuthenticationOAuth) listener.getAuth();
                 volumeList.addAll(AuthenticationUtils.configureOauthCertificateVolumes("oauth-" + ListenersUtils.identifier(listener), oauth.getTlsTrustedCertificates(), isOpenShift));
             }
+
+            if (isListenerWithCustomAuth(listener)) {
+                KafkaListenerAuthenticationCustom custom = (KafkaListenerAuthenticationCustom) listener.getAuth();
+                volumeList.addAll(AuthenticationUtils.configureGenericSecretVolumes("custom-" + ListenersUtils.identifier(listener), custom.getSecrets(), isOpenShift));
```
scholzj, line 1452: Can we fit something like `custom-listener-` here? Eventually some shortcut of it? `custom` sounds a bit too generic => it might prove to be limiting int he future, but it will not be easy to change if already used by users.

documentation/api/io.strimzi.api.kafka.model.authentication.KafkaClientAuthenticationCustom.adoc
```diff
@@ -0,0 +1,90 @@
...
+* SASL = False, TLS = False -> PLAINTEXT
+
+.Listener Config
+All configuration listed under `listener-config` will be prefixed with `listener.name.<listener-name>-<port>`.
+Therefore, the above `sasl.enabled.mechanisms` becomes `listener.name.<listener-name>-<port>.sasl.enabled.mechanisms`.
+
+.Secrets
+`Secrets` will be mounted to `/opt/palantir/custom-auth-secrets/custom-<listener-name>-<port>/<secret-name>` in the Kafka broker nodes' containers.
+
+For example, the mounted secret `example` in the above setup would be located at `/opt/palantir/custom-auth-secrets/custom-oauth-bespoke-9093/example`.
```
scholzj, line 47: I don't think the paths correspond to the code (+ even if I read the code wrong, should be `/opt/kafka` ;-))

cluster-operator/src/test/java/io/strimzi/operator/cluster/model/KafkaBrokerConfigurationBuilderTest.java
```diff
@@ -1410,6 +1411,122 @@ public void testOAuthOptions()  {
...
+    @ParallelTest
+    public void testCustomConfigSetProtocolMapCorrectlyForsSslSasl() {
+        GenericKafkaListener listener = new GenericKafkaListenerBuilder()
+                .withName("CUSTOM-LISTENER")
+                .withPort(9092)
+                .withType(KafkaListenerType.INTERNAL)
+                .withTls(true)
+                .withNewKafkaListenerAuthenticationCustomAuth()
+                .withSasl(true)
+                .withListenerConfig(ImmutableMap.of())
```
scholzj, line 1423: Please use `Map.of()` instead (here and in other places) or build the Map separately (it does not need to be immutable). This seems to pull Guava dependency which we don't really want just because of `ImmutableMap`. This is also what failed the CI build.

scholzj (commented):
Thanks for the PR. I did a quick pass through it and left some comments. I might have some more later, but it already now looks quite good.

To pass the build, you will need to run `make crd_install` to copy the CRDs to some more places. It would be also good to add some CHANGELOG.md record about this new feature.

And it make sure this keeps working, I think we should have some system test for this as well. I was thinking that maybe the system test can just use the custom authentication to setup `SCRAM-SHA-512` authentication (but using `type: custom`). That would be easy to implement as it does not need any special software of clients. But it gives us at least basic system test. @Frawless @im-konge WDYT?

api/src/main/java/io/strimzi/api/kafka/model/listener/KafkaListenerAuthenticationCustom.java
scholzj: Yeah ... that should be fine => I thought that that was also what we wanted in the proposal. Maybe we just forgot to update it. Feel free to remove it from the code as far as I'm concerned.

scholzj:
PS: You will also need to fix the DCO signoff => the details how to do it should be under the _Details_ link next to `DCO` in the status section of the PR.

Frawless:
@scholzj  ye, sounds good to me to have test you described and eventually we can add more in another PR

im-konge:
I agree with Jakub, if it will be needed, we can add more options to clients and then use it and add more STs. But some simple ST would be great for now

documentation/api/io.strimzi.api.kafka.model.authentication.KafkaClientAuthenticationCustom.adoc
Sam-Kramer: woops! my apologies, silly mistake on my behalf :)

api/src/main/java/io/strimzi/api/kafka/model/listener/KafkaListenerAuthenticationCustom.java
Sam-Kramer: Yup agreed, think that works better by forcing the user to define it.

Sam-Kramer:
> Thanks for the PR. I did a quick pass through it and left some comments. I might have some more later, but it already now looks quite good.
> 
> To pass the build, you will need to run `make crd_install` to copy the CRDs to some more places. It would be also good to add some CHANGELOG.md record about this new feature.
> 
> And it make sure this keeps working, I think we should have some system test for this as well. I was thinking that maybe the system test can just use the custom authentication to setup `SCRAM-SHA-512` authentication (but using `type: custom`). That would be easy to implement as it does not need any special software of clients. But it gives us at least basic system test. @Frawless @im-konge WDYT?

Yup agreed SystemTest makes sense / is what I promised in the proposal! Will get cracking on that ASAP.

api/src/main/java/io/strimzi/api/kafka/model/listener/KafkaListenerAuthenticationCustom.java
Sam-Kramer: Yup, my only hesitation here is that then we should add the `tlsTrustedCertificates` here, as otherwise users cannot configure their truststore at all for their listener.

Sam-Kramer: My ignorance surrounding `But the broker truststore should be IMHO configured through the standard Strimzi mechanism` is probably what's leading to my confusion. Currently I only see `brokerCertChainAndKey` which helps a user get their key/cert into the keystore, but I do not see anything for a truststore.

documentation/api/io.strimzi.api.kafka.model.authentication.KafkaClientAuthenticationCustom.adoc
```diff
@@ -0,0 +1,90 @@
+To configure custom authentication, set the `type` property to `custom`.
```
PaulRMellor, line 1: Not sure we need this file. If I understand correctly, this is custom Kafka listener/authentication option. In which case, this be added under the `authentication` heading in _documentation/api/io.strimzi.api.kafka.model.listener.arraylistener.GenericKafkaListener.adoc_

documentation/api/io.strimzi.api.kafka.model.authentication.KafkaClientAuthenticationCustom.adoc
```diff
@@ -0,0 +1,90 @@
...
+
+Custom authentication allows for any type of kafka supported authentication to be used.
```
PaulRMellor, line 3:
```suggestion
Custom authentication allows for any type of kafka-supported authentication to be used.
```

scholzj: `any type` sounds a bit weird. Should it be more about `custom type`?

api/src/main/java/io/strimzi/api/kafka/model/listener/KafkaListenerAuthenticationCustom.java
scholzj: Well, not directly. But it is controlled using the Clients CA which influences what is in the truststore.

systemtest/src/test/java/io/strimzi/systemtest/kafka/listeners/ListenersST.java
```diff
@@ -389,6 +389,90 @@ void testSendMessagesTlsScramSha(ExtensionContext extensionContext) {
...
+        LOGGER.info("Checking if generated password has {} characters", passwordLength);
+        String password = kubeClient().namespace(namespaceName).getSecret(kafkaUsername).getData().get("password");
+        String decodedPassword = new String(Base64.getDecoder().decode(password));
+
+        assertEquals(decodedPassword.length(), passwordLength);
+
+        Service kafkaService = kubeClient(namespaceName).getService(namespaceName, KafkaResources.bootstrapServiceName(clusterName));
+        String kafkaServiceDiscoveryAnnotation = kafkaService.getMetadata().getAnnotations().get("strimzi.io/discovery");
+        JsonArray serviceDiscoveryArray = new JsonArray(kafkaServiceDiscoveryAnnotation);
+        assertThat(serviceDiscoveryArray, is(StUtils.expectedServiceDiscoveryInfo(9096, "kafka", "scram-sha-512", true)));
```
scholzj, line 473: Same as above => this seems to be just a copy paste from another test which is not needed here I guess?

documentation/api/io.strimzi.api.kafka.model.authentication.KafkaClientAuthenticationCustom.adoc
```diff
@@ -0,0 +1,90 @@
...
+
+. The specified principal builder class exists on the image.
+If you're using the open source Strimzi images, you'll need to rebuild your own with your desired classes.
+*Note --* It's worth opening a Github issue on this particular problem _before_ building your own, as one may already exist.
+. No other listener is using `oauth` type authentication.
+This is because the Oauth listener forcefully appends its own principlar builder to the kafka config.
+. The specified principal builder is compatible with Strimzi.
+
+Custom principal builders need to support peer certificates for authentication, as Strimzi uses this to manage the Kafka cluster.
+This is expected to be identical, or very similar to https://github.com/strimzi/strimzi-kafka-oauth/blob/main/oauth-server/src/main/java/io/strimzi/kafka/oauth/server/OAuthKafkaPrincipalBuilder.java[Strimzi's OAuth library].
```
PaulRMellor, line 62:
```suggestion
ifdef::Downloading[]
A custom OAuth principal builder might be identical or very similar to the Strimzi https://github.com/strimzi/strimzi-kafka-oauth/blob/main/oauth-server/src/main/java/io/strimzi/kafka/oauth/server/OAuthKafkaPrincipalBuilder.java[OAuth principal builder].
endif::Downloading[]
```

PaulRMellor: I've added a condition here, which we normally do with links to github repo

documentation/api/io.strimzi.api.kafka.model.authentication.KafkaClientAuthenticationCustom.adoc
```diff
@@ -0,0 +1,90 @@
...
+In particular, it leverages https://github.com/apache/kafka/blob/trunk/clients/src/main/java/org/apache/kafka/common/security/authenticator/DefaultKafkaPrincipalBuilder.java#L73-L79[Kafka's default principal builder class], which allows the building of a principal based upon the name of the of peer certificate.
+Thus, it is expected that the custom principal builder provides a principal of type "user" with the name being that of the ssl peer certificate.
```
PaulRMellor, line 64:
```suggestion
The custom principal builder should provide a principal of type "user" using the name of the SSL peer certificate.
```

documentation/api/io.strimzi.api.kafka.model.authentication.KafkaClientAuthenticationCustom.adoc
```diff
@@ -0,0 +1,90 @@
...
+
+An example of such a principal builder which satisfies Strimzi’s auth requirements:
```
PaulRMellor, line 66:
```suggestion
The following example shows a custom principal builder that satisfies the OAuth requirements of Strimzi.
```

documentation/api/io.strimzi.api.kafka.model.authentication.KafkaClientAuthenticationCustom.adoc
```diff
@@ -0,0 +1,90 @@
...
+
+[source,java,subs="attributes+"]
```
PaulRMellor, line 68:
```suggestion
.Example principal builder for custom OAuth configuration
[source,java,subs="attributes+"]
```

PaulRMellor (commented):
Doc looking good. I left a comment about the location of the content. Hope I understood correctly and it makes sense

api/src/main/java/io/strimzi/api/kafka/model/listener/KafkaListenerAuthenticationCustom.java
Sam-Kramer: Gotcha, and that means we need to add the following to config for all custom listener types:
```
            writer.println(String.format("listener.name.%s.ssl.truststore.location=/tmp/kafka/clients.truststore.p12", listenerNameInProperty));
            writer.println(String.format("listener.name.%s.ssl.truststore.password=${CERTS_STORE_PASSWORD}", listenerNameInProperty));
            writer.println(String.format("listener.name.%s.ssl.truststore.type=PKCS12", listenerNameInProperty));
```

scholzj: I don't know. That would mean it is using TLS client authentication. That might not be the case - at least not for everyone. I would say that if you want TLS client authentication, you should use `type: tls` and not `type: custom`.

Sam-Kramer: Ok, let's leave this out for now then. If this is needed down the road for some other use-case, this is a trivial PR to fix.

systemtest/src/test/java/io/strimzi/systemtest/kafka/listeners/ListenersST.java
```diff
@@ -389,6 +389,90 @@ void testSendMessagesTlsScramSha(ExtensionContext extensionContext) {
...
+                .editOrNewTemplate()
+                .editOrNewUserOperatorContainer()
+                .addToEnv(
+                        new ContainerEnvVarBuilder()
+                                .withName("STRIMZI_SCRAM_SHA_PASSWORD_LENGTH")
+                                .withValue(String.valueOf(passwordLength))
+                                .build())
+                .endUserOperatorContainer()
+                .endTemplate()
+                .endEntityOperator()
```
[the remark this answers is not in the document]
Sam-Kramer: I thought this would be needed for
 ```        internalKafkaClient.checkProducedAndConsumedMessages(
                internalKafkaClient.sendMessagesTls(),
                internalKafkaClient.receiveMessagesTls()
        );
```

File: documentation/api/io.strimzi.api.kafka.model.authentication.KafkaClientAuthenticationCustom.adoc (hunk 8 above)
Comment (Sam-Kramer) [reply documentation/api/io.strimzi.api.kafka.model.authentication.KafkaClientAuthenticationCustom.adoc]: Woops, that was an accident on my behalf! I renamed + linked from `documentation/api/io.strimzi.api.kafka.model.listener.arraylistener.GenericKafkaListener.adoc`

File: api/src/main/java/io/strimzi/api/kafka/model/listener/KafkaListenerAuthenticationCustom.java (hunk 3 above)
Comment (scholzj) [reply api/src/main/java/io/strimzi/api/kafka/model/listener/KafkaListenerAuthenticationCustom.java:28]: Agree. We can always add that later if we want. So, I guess that means this should then change to `ssl.`, or?

File: api/src/main/java/io/strimzi/api/kafka/model/listener/KafkaListenerAuthenticationCustom.java [hunk 16]
```diff
@@ -0,0 +1,74 @@
...
+    @JsonInclude(JsonInclude.Include.NON_NULL)
+    public Map<String, Object> getListenerConfig() {
+        return listenerConfig;
+    }
+
+    public void setListenerConfig(Map<String, Object> config) {
+        this.listenerConfig = config;
+    }
+
+    @Description("Secrets to be mounted to /mnt/strimzi/custom-auth-secrets/<listener-name>/<secret-key>")
```
scholzj, line 65: I think this is another wrong path. FRom the code it seems like `/opt/kafka/custom-auth-secrets/custom-<listener-name>/<secret-name>` or something like that. I think you have the right path in `io.strimzi.api.kafka.model.listener.KafkaListenerAuthenticationCustom.adoc`, just need to fix it here as well.

scholzj (commented):
One more nit which I found.

api/src/main/java/io/strimzi/api/kafka/model/listener/KafkaListenerAuthenticationCustom.java
```diff
@@ -0,0 +1,74 @@
...
+    @JsonInclude(JsonInclude.Include.NON_NULL)
+    public boolean isSasl() {
+        return sasl;
+    }
+
+    public void setSasl(boolean enabled) {
+        this.sasl = enabled;
+    }
+
+    @Description("Configuration to be used for a specific listener. All values are prefixed with lister.name.<lister-name>.")
```
PaulRMellor, line 55:
```suggestion
    @Description("Configuration to be used for a specific listener. All values are prefixed with listener.name._<listener_name>_.")
```

documentation/api/io.strimzi.api.kafka.model.listener.arraylistener.GenericKafkaListener.adoc
```diff
@@ -251,6 +251,7 @@ Authentication for the listener can be specified as:
 * Mutual TLS (`tls`)
 * SCRAM-SHA-512 (`scram-sha-512`)
 * Token-based OAuth 2.0 (`oauth`).
```
PaulRMellor, line 253: If you could remove this period, that would be great!

PaulRMellor (approved):
Great. :+1:

api/src/main/java/io/strimzi/api/kafka/model/listener/KafkaListenerAuthenticationCustom.java
Sam-Kramer: Yeah re-thinking this, should we mount it to `/opt/kafka/custom-auth-secrets/custom-listener-<listener-name>/<secret-name>` instead? (Only difference, is the folder name `custom-<listener-name>` vs `custom-listener-<listener-name>`

Sam-Kramer: Also, what do we think is more clear, `custom-<listener-name>-<port>` or `custom-<listener-name>`? Technically, it's the former, but the <listener-name> maybe assumed to be listener name + port?

scholzj: I'm flexible TBH. I think the `/opt/kafka/custom-auth-secrets` is important since that defines what secrets these are. Having both the listener name and port in the name helps to orient your self better from my opinion. I think other paths use the `listener-name-port` format as well.

api/src/main/java/io/strimzi/api/kafka/model/listener/KafkaListenerAuthenticationCustom.java
```diff
@@ -0,0 +1,99 @@
...
+
+    public Map<String, Object> getFilteredListenerConfig() {
+        if (listenerConfig == null) {
+            return listenerConfig;
+        }
+        HashMap<String, Object> filteredConfig = new HashMap<>();
+        listenerConfig.forEach((key, value) -> {
+            if (key != null) {
+                String cleanedKey = key.toLowerCase().trim();
+                if (FORBIDDEN_PREFIXES.stream().anyMatch(prefix -> cleanedKey.startsWith(prefix.toLowerCase()))) {
```
scholzj, line 89: I think these trigger some spotbugs error (you should use some charset here).

api/src/main/java/io/strimzi/api/kafka/model/listener/KafkaListenerAuthenticationCustom.java
```diff
@@ -0,0 +1,99 @@
...
+    public List<GenericSecretSource> getSecrets() {
+        return secrets;
+    }
+
+    public void setSecrets(List<GenericSecretSource> secrets) {
+        this.secrets = secrets;
+    }
+
+
+    public Map<String, Object> getFilteredListenerConfig() {
```
scholzj, line 81: I think this would maybe have a better place in the `cluster-operator` module than in the `api` module. I also wonder if any of the existing methods can be reused for this => this pattern is already used in most of the `config` fields. Reusing that would probably also give it the proper validation against the Kafka configuration models.

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/ListenersValidator.java
```diff
@@ -560,4 +564,20 @@ private static void validateOauth(Set<String> errors, GenericKafkaListener liste
...
+     * Filters configuration provided in any custom authentication listeners to remove any forbidden properties
+     * @param reconciliation Current reconciliation attempt
+     * @param listener Listener to filter
+     */
+    private static void filterCustomAuthConfiguration(Reconciliation reconciliation, GenericKafkaListener listener) {
+        if (isListenerWithCustomAuth(listener)) {
+            KafkaListenerAuthenticationCustom customAuth = (KafkaListenerAuthenticationCustom) listener.getAuth();
+            if (customAuth.getListenerConfig() != null) {
+                KafkaListenerCustomAuthConfiguration config = new KafkaListenerCustomAuthConfiguration(reconciliation, customAuth.getListenerConfig().entrySet());
+                customAuth.setListenerConfig(Map.copyOf(config.asOrderedProperties().asMap()));
```
scholzj, line 579: I think modifying the custom resource is fairly dangerous. Can't this be used instead in `KafkaBrokerConfigurationBuilder.java` when passing the configuration to the broker config file? There you can remove the forbidden options without modifying the resource and storing the anywhere.

cluster-operator/src/test/java/io/strimzi/operator/cluster/model/KafkaBrokerConfigurationBuilderTest.java
```diff
@@ -1412,6 +1412,121 @@ public void testOAuthOptions()  {
...
+    @ParallelTest
+    public void testCustomConfigPrefixesUserProvidedConfig() {
+        GenericKafkaListener listener = new GenericKafkaListenerBuilder()
+                .withName("CUSTOM-LISTENER")
+                .withPort(9092)
+                .withType(KafkaListenerType.INTERNAL)
+                .withTls(true)
+                .withNewKafkaListenerAuthenticationCustomAuth()
+                .withSasl(true)
+                .withListenerConfig(Map.of("oauthbearer.sasl.client.callback.handler.class", "client.class",
```
scholzj, line 1484: Can we add some `ssl...` option here to test that it will not be passed into the configuration?

scholzj (commented):
LGTM, few last comments. It also seems you need to update the custom resources and docs after the changes to the `api` module. And you will also need to fix the DCO sign-off I mentioned before.

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/ListenersValidator.java
Sam-Kramer: @scholzj the only issue is that the `Reconciliation` object has no reference here, which is required for the AbstractConfig class. I'm happy to make a constructor that doesn't require that though, or simply pass in the `Reconciliation` object to builder alternatively?

cluster-operator/src/test/java/io/strimzi/operator/cluster/model/KafkaBrokerConfigurationBuilderTest.java
Sam-Kramer: In this present case, I've moved the test to `KafkaClusterTest` (https://github.com/strimzi/strimzi-kafka-operator/pull/6162/files#diff-a6f8b697177e22f1da12401caa919e111ac8d6b204b437022576893e04732784R3698-R3722) as the removal of forbidden prefixes happens in `KafkaCluster` itself, rather than `KafkaBrokerConfigurationBuilder`. That said, if we move the removal to `KafkaBrokerConfigurationBuilder` we can then test here!

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/ListenersValidator.java
scholzj: I doubt you can easily create a constructor without the reconciliation parameter because it is passed to the parent constructor. So I guess passing the reconciliation to the `KafkaBrokerConfigurationBuilder` class, probably in its constructor might be the easier way.

scholzj (commented):
Ok, I think the code looks good to me. I started the regression test suite as well. You still need to fix the DCO.

scholzj (approved):
Thanks. LGTM.

scholzj:
@ppatierno @tombentley Do you have something more to this?

ppatierno (approved):
LGTM. There are some conflicts to address but thanks for the PR!

Sam-Kramer:
Rebased, should be good to go. Thank you everyone (@scholzj / @ppatierno / @PaulRMellor) for your time/patience/collaboration on this :)

PR #6156, opened 2022-01-07
Add log-dump script ST and refactoring

fvaleri:
This ST is much lighter than the other one (cold-backup). 

I also did some refactoring to make the two script styles consistent with each other, and improved a couple of functions (`wait_for()` in cold-backup and `get_kafka_setup()` in log-dump). Anyway, there is no interface or behavior change.

The small change to VerifiableClient is required by find-bugs.

PR #6165, opened 2022-01-11
Update spotbugs and checkstyle

sknot-rh:

### Type of change
- Chores

### Description
Updated spotbugs and checkstyle. Spotbugs now checks access to the private members from the static methods. So using singleton pattern triggers `*_EXPOSE_REP*` warnings (also public method getting or setting private field). I added those to spotbugs exclude.

### Checklist

- [ ] Write tests
- [ ] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/operator-common/src/test/java/io/strimzi/operator/common/operator/resource/AbstractNonNamespacedResourceOperatorTest.java
+++ b/operator-common/src/test/java/io/strimzi/operator/common/operator/resource/AbstractNonNamespacedResourceOperatorTest.java
@@ -100,7 +100,7 @@ public abstract class AbstractNonNamespacedResourceOperatorTest<C extends Kubern
         HasMetadata hasMetadata = mock(HasMetadata.class);
         when(mockResource.get()).thenReturn(resource);
 
-        when(mockResource.withPropagationPolicy(true ? DeletionPropagation.FOREGROUND : DeletionPropagation.ORPHAN)).thenReturn(mockR);
+        when(mockResource.withPropagationPolicy(DeletionPropagation.FOREGROUND)).thenReturn(mockR);
         when(mockR.patch((T) any())).thenReturn(hasMetadata);
 
         NonNamespaceOperation mockNameable = mock(NonNamespaceOperation.class);
@@ -132,7 +132,7 @@ public abstract class AbstractNonNamespacedResourceOperatorTest<C extends Kubern
         T resource = resource();
         Resource mockResource = mock(resourceType());
         when(mockResource.get()).thenReturn(resource);
-        when(mockResource.withPropagationPolicy(true ? DeletionPropagation.FOREGROUND : DeletionPropagation.ORPHAN)).thenReturn(mockResource);
+        when(mockResource.withPropagationPolicy(DeletionPropagation.FOREGROUND)).thenReturn(mockResource);
         when(mockResource.patch(any())).thenReturn(resource);
 
         NonNamespaceOperation mockNameable = mock(NonNamespaceOperation.class);
--- a/systemtest/src/main/java/io/strimzi/systemtest/kafkaclients/internalClients/VerifiableClient.java
+++ b/systemtest/src/main/java/io/strimzi/systemtest/kafkaclients/internalClients/VerifiableClient.java
@@ -123,8 +123,9 @@ public class VerifiableClient {
 
         if (clientType == ClientType.CLI_KAFKA_VERIFIABLE_CONSUMER) {
             this.consumerGroupName = verifiableClientBuilder.consumerGroupName;
-            this.clientArgumentMap.put(ClientArgument.GROUP_ID, consumerGroupName);
             this.consumerInstanceId = verifiableClientBuilder.consumerInstanceId;
+            this.clientArgumentMap.put(ClientArgument.GROUP_ID, consumerGroupName);
+            this.clientArgumentMap.put(ClientArgument.GROUP_INSTANCE_ID, consumerInstanceId);
         }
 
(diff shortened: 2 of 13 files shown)


operator-common/src/test/java/io/strimzi/operator/common/operator/resource/AbstractNonNamespacedResourceOperatorTest.java
```diff
@@ -100,7 +100,7 @@ public void testCreateWhenExistsWithChangeIsAPatch(VertxTestContext context) {
         HasMetadata hasMetadata = mock(HasMetadata.class);
         when(mockResource.get()).thenReturn(resource);
 
-        when(mockResource.withPropagationPolicy(true ? DeletionPropagation.FOREGROUND : DeletionPropagation.ORPHAN)).thenReturn(mockR);
+        when(mockResource.withPropagationPolicy(DeletionPropagation.FOREGROUND)).thenReturn(mockR);
```
scholzj, line 103: Why change this?

operator-common/src/test/java/io/strimzi/operator/common/operator/resource/AbstractNonNamespacedResourceOperatorTest.java
```diff
@@ -132,7 +132,7 @@ public void testCreateWhenExistsWithoutChangeIsNotAPatch(VertxTestContext contex
         T resource = resource();
         Resource mockResource = mock(resourceType());
         when(mockResource.get()).thenReturn(resource);
-        when(mockResource.withPropagationPolicy(true ? DeletionPropagation.FOREGROUND : DeletionPropagation.ORPHAN)).thenReturn(mockResource);
+        when(mockResource.withPropagationPolicy(DeletionPropagation.FOREGROUND)).thenReturn(mockResource);
```
scholzj, line 135: Same as above.

systemtest/src/main/java/io/strimzi/systemtest/kafkaclients/internalClients/VerifiableClient.java
```diff
@@ -122,8 +122,9 @@ public VerifiableClient(VerifiableClientBuilder verifiableClientBuilder) {
 
         if (clientType == ClientType.CLI_KAFKA_VERIFIABLE_CONSUMER) {
             this.consumerGroupName = verifiableClientBuilder.consumerGroupName;
-            this.clientArgumentMap.put(ClientArgument.GROUP_ID, consumerGroupName);
             this.consumerInstanceId = verifiableClientBuilder.consumerInstanceId;
+            this.clientArgumentMap.put(ClientArgument.GROUP_ID, consumerGroupName);
+            this.clientArgumentMap.put(ClientArgument.GROUP_INSTANCE_ID, consumerInstanceId);
```
scholzj, line 127: Why this change? Seems unrelated to spotbugs or checkstyle.

systemtest/src/main/java/io/strimzi/systemtest/logs/TestExecutionWatcher.java
```diff
@@ -20,6 +20,8 @@
 
 public class TestExecutionWatcher implements TestExecutionExceptionHandler, LifecycleMethodExecutionExceptionHandler {
 
+    private static final Random RANDOM = new Random();
```
scholzj, line 23: Why do you use once static field and once non-static? Shouldn't it use everywhere the same?

scholzj (commented):
> Spotbugs now checks access to the private members from the static methods. So using singleton pattern triggers *_EXPOSE_REP* warnings. I added those to spotbugs exclude.

Perhaps you can elaborate some more on this and provide some examples of what it doesn't like and why should it be excluded?

operator-common/src/test/java/io/strimzi/operator/common/operator/resource/AbstractNonNamespacedResourceOperatorTest.java
sknot-rh: Well, `true ? DeletionPropagation.FOREGROUND : DeletionPropagation.ORPHAN` is equal to 
```
if (true) {
    doSomething();`
    } else {
    doSomethingElse();
}
``` 
so it never gets to the else branch, does it?

scholzj: Good point. I didn't noticed it was just `true` and not some actual condition. Weird code :-o.

systemtest/src/main/java/io/strimzi/systemtest/kafkaclients/internalClients/VerifiableClient.java
sknot-rh: Spotbugs reported `consumerGroupName` as unused. Looking at the code I thing it might a bug.
It was partially fixed in https://github.com/strimzi/strimzi-kafka-operator/pull/6156/files#diff-f93d2fac0903e6c67c6a2111f4b65fa4b61a2fd27690459d80d0ae4c6561a70eR126
@fvaleri @see-quick Could you please take a look?

scholzj (approved):

systemtest/src/main/java/io/strimzi/systemtest/kafkaclients/internalClients/VerifiableClient.java
fvaleri: Yeah, that actually revealed a bug and I forgot to add it to the arguments map. Thanks.

Issue #4433, opened 2021-02-17
Optional Load Balancer (+ external IP) for bootstrap server when using type "loadbalancer"


trietsch:
**Is your feature request related to a problem? Please describe.**
When using type `loadbalancer` for a Kafka listener, `n+1` Load Balancers are created in Kubernetes, `n` for the amount of brokers and 1 for the bootstrap server.

**Describe the solution you'd like**
The bootstrap server is only for convenience (as discussed [here](https://github.com/strimzi/strimzi-kafka-operator/discussions/4430)), which may not be always necessary. Providing a flag in the configuration for the `loadbalancer` type listener stating whether to provision a Load Balancer for the bootstrap server would be a nice addition.

**Describe alternatives you've considered**
The only alternative now is to always provision a bootstrap Load Balancer service.

Sh1n:
I am also looking for a way to disable the +1 bootstrap server on a given listener.

scholzj:
FYI: I have this on my TODO list, but do not expect to get to it in 0.22 ... probably will have some time to look into it in 0.23.

yyang48:
Just checked, we have two bootstrap services in a single Kafka Cluster. For instance, in a running Kafka Cluster managed by strimzi in AWS EKS, we have the following k8s services created by strimzi: 
```test-kafka-01-cruise-control
 test-kafka-01-kafka-0
 test-kafka-01-kafka-1
 test-kafka-01-kafka-2
 test-kafka-01-kafka-3
 test-kafka-01-kafka-4
 test-kafka-01-kafka-5
 test-kafka-01-kafka-bootstrap
 test-kafka-01-kafka-brokers
 test-kafka-01-kafka-external-bootstrap
 test-kafka-01-zookeeper-client
 test-kafka-01-zookeeper-nodes
```

It seems like the strimzi will create one Load Balancer for the  `test-kafka-01-kafka-external-bootstrap` service, not the `test-kafka-01-kafka-bootstrap`. The `test-kafka-01-kafka-bootstrap` is used internally by the cruise control. 

So this issue is to give an option NOT to create the `test-kafka-01-kafka-external-bootstrap` service? With this, the Load Balancer for this service will not be created as well. There is no change for the ` test-kafka-01-kafka-bootstrap`

Is this correct?

scholzj:
[...]

PR #6170, opened 2022-01-12
Add the option to skip the creation of the bootstrap service in the Load

yyang48:
Balancer Kafka Listener.

This is the fix for the issue: https://github.com/strimzi/strimzi-kafka-operator/issues/4433

Signed-off-by: Yaodong Yang <yyang48@apple.com>

### Type of change

_Select the type of your PR_

- Enhancement / new feature

### Description

### Checklist

- [x] Write tests
- [x] Update documentation
- [x] Update CHANGELOG.md

Screen Shot 2022-01-20 at 8 53 19 AM

### DIFF
--- a/api/src/main/java/io/strimzi/api/kafka/model/listener/arraylistener/GenericKafkaListenerConfiguration.java
+++ b/api/src/main/java/io/strimzi/api/kafka/model/listener/arraylistener/GenericKafkaListenerConfiguration.java
@@ -32,7 +32,7 @@ import static java.util.Collections.emptyMap;
  */
 @DescriptionFile
 @JsonPropertyOrder({"brokerCertChainAndKey", "ingressClass", "preferredAddressType", "externalTrafficPolicy",
-    "loadBalancerSourceRanges", "bootstrap", "brokers", "ipFamilyPolicy", "ipFamilies"})
+    "loadBalancerSourceRanges", "bootstrap", "brokers", "ipFamilyPolicy", "ipFamilies", "createBootstrapService"})
 @JsonInclude(JsonInclude.Include.NON_NULL)
 @Buildable(
     editableEnabled = false,
@@ -55,6 +55,7 @@ public class GenericKafkaListenerConfiguration implements Serializable, UnknownP
     private Integer maxConnectionCreationRate;
     private IpFamilyPolicy ipFamilyPolicy;
     private List<IpFamily> ipFamilies;
+    private Boolean createBootstrapService = true;
 
     private Map<String, Object> additionalProperties = new HashMap<>(0);
 
@@ -227,6 +228,17 @@ public class GenericKafkaListenerConfiguration implements Serializable, UnknownP
 
     public void setIpFamilies(List<IpFamily> ipFamilies) {
         this.ipFamilies = ipFamilies;
+    }
+
+    @Description("Whether to create the bootstrap service or not." +
+            "This field can be used with the `loadBalancer` type listener.")
+    @JsonInclude(JsonInclude.Include.NON_NULL)
+    public Boolean getCreateBootstrapService() {
+        return createBootstrapService;
+    }
+
+    public void setCreateBootstrapService(Boolean createBootstrapService) {
+        this.createBootstrapService = createBootstrapService;
     }
 
     @Override
--- a/api/src/main/java/io/strimzi/api/kafka/model/listener/arraylistener/GenericKafkaListenerConfiguration.java
+++ b/api/src/main/java/io/strimzi/api/kafka/model/listener/arraylistener/GenericKafkaListenerConfiguration.java
@@ -231,6 +231,7 @@ public class GenericKafkaListenerConfiguration implements Serializable, UnknownP
     }
 
     @Description("Whether to create the bootstrap service or not. " +
+            "The bootstrap service is created by default (if not specified differently). " +
             "This field can be used with the `loadBalancer` type listener.")
     @JsonInclude(JsonInclude.Include.NON_DEFAULT)
     public Boolean getCreateBootstrapService() {
--- a/api/src/main/java/io/strimzi/api/kafka/model/listener/arraylistener/GenericKafkaListenerConfiguration.java
+++ b/api/src/main/java/io/strimzi/api/kafka/model/listener/arraylistener/GenericKafkaListenerConfiguration.java
@@ -230,9 +230,9 @@ public class GenericKafkaListenerConfiguration implements Serializable, UnknownP
         this.ipFamilies = ipFamilies;
     }
 
-    @Description("Whether to create the bootstrap service or not." +
+    @Description("Whether to create the bootstrap service or not. " +
             "This field can be used with the `loadBalancer` type listener.")
-    @JsonInclude(JsonInclude.Include.NON_NULL)
+    @JsonInclude(JsonInclude.Include.NON_DEFAULT)
     public Boolean getCreateBootstrapService() {
         return createBootstrapService;
     }
--- a/api/src/test/java/io/strimzi/api/kafka/model/KafkaTest.java
+++ b/api/src/test/java/io/strimzi/api/kafka/model/KafkaTest.java
@@ -7,11 +7,13 @@ package io.strimzi.api.kafka.model;
 import io.fabric8.kubernetes.api.model.ObjectMeta;
 import io.strimzi.api.kafka.model.listener.arraylistener.GenericKafkaListener;
 import io.strimzi.api.kafka.model.listener.arraylistener.GenericKafkaListenerBuilder;
+import io.strimzi.api.kafka.model.listener.arraylistener.GenericKafkaListenerConfiguration;
 import io.strimzi.api.kafka.model.listener.arraylistener.KafkaListenerType;
 import io.strimzi.test.TestUtils;
 import org.junit.jupiter.api.Test;
 
 import java.net.URISyntaxException;
+import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
 import java.util.Objects;
@@ -80,14 +82,26 @@ public class KafkaTest extends AbstractCrdTest<Kafka> {
 
     @Test
     public void testNewListenerSerialization() throws URISyntaxException {
-        List<GenericKafkaListener> listeners = Collections.singletonList(
-                new GenericKafkaListenerBuilder()
+
+        List<GenericKafkaListener> listeners = new ArrayList<>(2);
+        listeners.add(new GenericKafkaListenerBuilder()
                         .withName("lst")
                         .withPort(9092)
                         .withType(KafkaListenerType.INTERNAL)
                         .withTls(true)
                 .build()
         );
+
+        GenericKafkaListenerConfiguration genericKafkaListenerConfiguration = new GenericKafkaListenerConfiguration();
+        genericKafkaListenerConfiguration.setCreateBootstrapService(false);
+
+        listeners.add(new GenericKafkaListenerBuilder()
+                .withName("snd")
+                .withPort(9094)
+                .withType(KafkaListenerType.LOADBALANCER)
+                .withTls(true)
+                .withConfiguration(genericKafkaListenerConfiguration)
+                .build());
 
         Kafka kafka = new KafkaBuilder()
                 .withNewMetadata()
--- a/api/src/test/java/io/strimzi/api/kafka/model/KafkaTest.java
+++ b/api/src/test/java/io/strimzi/api/kafka/model/KafkaTest.java
@@ -7,13 +7,11 @@ package io.strimzi.api.kafka.model;
 import io.fabric8.kubernetes.api.model.ObjectMeta;
 import io.strimzi.api.kafka.model.listener.arraylistener.GenericKafkaListener;
 import io.strimzi.api.kafka.model.listener.arraylistener.GenericKafkaListenerBuilder;
-import io.strimzi.api.kafka.model.listener.arraylistener.GenericKafkaListenerConfiguration;
 import io.strimzi.api.kafka.model.listener.arraylistener.KafkaListenerType;
 import io.strimzi.test.TestUtils;
 import org.junit.jupiter.api.Test;
 
 import java.net.URISyntaxException;
-import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
 import java.util.Objects;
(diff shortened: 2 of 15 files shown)


cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaAssemblyOperator.java
```diff
@@ -2053,6 +2054,13 @@ String zkConnectionString(int connectToReplicas, Function<Integer, String> zkNod
             for (GenericKafkaListener listener : loadBalancerListeners) {
                 String bootstrapServiceName = ListenersUtils.backwardsCompatibleBootstrapServiceName(name, listener);
 
+                // When the listener is load balancer and useCreateBootStrapService is false,
+                // we will skip the creation of the external load balancer to save the cost.
+                if (KafkaListenerType.LOADBALANCER == listener.getType() && listener.getConfiguration() != null
+                        && !listener.getConfiguration().getCreateBootstrapService()) {
+                    continue;
+                }
```
scholzj, line 2062: I'm afraid this would be too simple :-/. It skips the creation of the ListenerStatus which we definitely do not want to skip. We would want it at least without the address, but ideally with the address constructed from the broker services / loadbalancers.

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaCluster.java
```diff
@@ -790,6 +790,10 @@ public Service generateService() {
         List<Service> services = new ArrayList<>(externalListeners.size());
 
         for (GenericKafkaListener listener : externalListeners)   {
+            if (KafkaListenerType.LOADBALANCER == listener.getType() && listener.getConfiguration() != null && !listener.getConfiguration().getCreateBootstrapService()) {
```
scholzj, line 793: MAybe this `if` can have some method in https://github.com/strimzi/strimzi-kafka-operator/blob/main/cluster-operator/src/main/java/io/strimzi/operator/cluster/model/ListenersUtils.java ... you already seem to be using it in 2 places.

api/src/main/java/io/strimzi/api/kafka/model/listener/arraylistener/GenericKafkaListenerConfiguration.java
```diff
@@ -229,6 +230,17 @@ public void setIpFamilies(List<IpFamily> ipFamilies) {
         this.ipFamilies = ipFamilies;
     }
 
+    @Description("Whether to create the bootstrap service or not." +
+            "This field can be used with the `loadBalancer` type listener.")
```
scholzj, line 234: Since we support this only with load balancers, you should also add some validation to https://github.com/strimzi/strimzi-kafka-operator/blob/main/cluster-operator/src/main/java/io/strimzi/operator/cluster/model/ListenersValidator.java

scholzj (commented):
Thanks for the PR. I left some initial comments.

api/src/test/java/io/strimzi/api/kafka/model/KafkaTest.java
```diff
@@ -80,15 +82,27 @@ public void testCertificationAuthorityBuilderAndInts() throws URISyntaxException
...
+        GenericKafkaListenerConfiguration genericKafkaListenerConfiguration = new GenericKafkaListenerConfiguration();
+        genericKafkaListenerConfiguration.setCreateBootstrapService(false);
+
+        listeners.add(new GenericKafkaListenerBuilder()
+                .withName("snd")
+                .withPort(9094)
+                .withType(KafkaListenerType.LOADBALANCER)
+                .withTls(true)
+                .withConfiguration(genericKafkaListenerConfiguration)
+                .build());
```
scholzj, line 104: I guess it is fine to keep this ... but TBH it is not really needed. I do not think this option affects the serialization in any significant way which is  what is really tested here.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaAssemblyOperator.java
yyang48: Sorry, I don't understand what you mentioned above.  You mean we want to keep the headless service for the bootstrap service, but don't assign a load balancer IP for it? Here is my test example:

If we skip the bootstrap service, I got the following services. The external bootstrap service is skipped.

image

Do we still want to create the external-bootstrap service, but without an External Load Balancer assigned? I guess my confusion is where is the ListenerStatus defined.

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaCluster.java
yyang48: done.

api/src/test/java/io/strimzi/api/kafka/model/KafkaTest.java
yyang48: Yes, it's just a boolean value, not a complex object. I take a second thought before adding them here next time

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaAssemblyOperator.java
yyang48: put it in another way, what should the expect output be.

scholzj: When you look at the deployed Kafka cluster, in the status section you should see something like this:

```yaml
    - addresses:
      - host: 192.168.1.72
        port: 9094
      bootstrapServers: 192.168.1.72:9094
      certificates:
      - |
        -----BEGIN CERTIFICATE-----
        ...
        -----END CERTIFICATE-----
      type: external
```

That is all generated here on the lines 2077-2084. Your code basiclly just skips this, so there would be no record about this listener in the status. I think we should think about how to handle that. Ideally, we would want there multiple addresses instead of the _per-broker_ load balancers. Not sure how easy that would be. But if nothing else, we would probably want to at least have the information abott the listener without the address.

yyang48: I got your point now, thanks for the explanation! Let me look at the code again.

yyang48: So we would like to change the listener status to this (if we skip the creation of the bootstrap service)? use the above example:

``` 
     - addresses:
      bootstrapServers: 192.168.1.72:9094 
      certificates:
      - |
        -----BEGIN CERTIFICATE-----
        ...
        -----END CERTIFICATE-----
      type: external
```

It seems like a better idea as it aligns with Apache Kafka's bootstrap services, so it's widely recognized by the kafka users.

Am I correct?

scholzj: Yes. I think I would be flexible if it would be too complicated. But ideally we would probably want something like this when the bootstrap service is disabled:

```yaml
    - addresses:
        - host: 192.168.1.72
          port: 9094
        - host: 192.168.1.73
          port: 9094
        - host: 192.168.1.74
          port: 9094
      bootstrapServers: 192.168.1.72:9094,192.168.1.73:9094,192.168.1.74:9094 
      certificates:
        - |
          -----BEGIN CERTIFICATE-----
          ...
          -----END CERTIFICATE-----
      type: external
```

Where `192.168.1.72:9094,192.168.1.73:9094,192.168.1.74:9094 ` are the addresses of the loadbalancers corresponding to the individual brokers.

yyang48: Got it, thanks! It shouldn't be hard. let me try again and update the PR later.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaAssemblyOperator.java
```diff
@@ -2053,78 +2053,88 @@ String zkConnectionString(int connectToReplicas, Function<Integer, String> zkNod
...
-                            String bootstrapAddress;
-
-                            if (svc.getStatus().getLoadBalancer().getIngress().get(0).getHostname() != null)    {
-                                bootstrapAddress = svc.getStatus().getLoadBalancer().getIngress().get(0).getHostname();
-                            } else {
-                                bootstrapAddress = svc.getStatus().getLoadBalancer().getIngress().get(0).getIp();
-                            }
-
-                            LOGGER.debugCr(reconciliation, "Found address {} for Service {}", bootstrapAddress, bootstrapServiceName);
+                List<String> listenerAddressList = new ArrayList<>(kafkaCluster.getReplicas() + 1);
```
yyang48, line 2056: @scholzj I guess I probably need your help to take a look at this section...

I tested a couple of times about my PR and here is what I found:
1. In the new PR, the operator didn't wait for the completion of generating these Load Balancer Addresses, and it continue to generate the broker certificate and even create the pods with an empty Load Balancer address. Therefore the Kafka Broker can't start because of the lack of the Load Balancer address.
2. My suspicious is my understanding of the vert.x future is not correct, do I need a join after get the bootstrap address in line 2080?

Thanks!

scholzj: Yeah. The way you have it written, the `perListenerFut` will complete either on line 2057 or 2077. The whole block between lines 2082-2137 => it will not wait for it because it will be already complete. If you want, I can try to rewrite it tomorrow and share the code with you to compare it. Or we can have some call to go through it via screen sharing tomorrow (or when you have time). Whatever you prefer.

yyang48: Thank you for your time and the help. I tested locally, it works! Updating the PR now and adding some tests on it.

yyang48: @scholzj do you have any suggestions for the testing? I'm thinking to add the test in this file: https://github.com/strimzi/strimzi-kafka-operator/blob/main/cluster-operator/src/test/java/io/strimzi/operator/cluster/operator/assembly/KafkaAssemblyOperatorTest.java. It doesn't have the load balancer listener today, I guess I can add this type and test it with `createBootstrapService` disabled. Or it's better to do it in other places?

scholzj: No, please don't add it there. That parametrized test is quite bad since any test you add there will run in many unrelated combinations. Maybe `KafkaAssemblyOperatorNonParametrizedTest` would be a good place?

yyang48:
@scholzj could you please take another look for this PR? It's ready for the review. Thanks!

api/src/main/java/io/strimzi/api/kafka/model/listener/arraylistener/GenericKafkaListenerConfiguration.java
```diff
@@ -229,6 +230,17 @@ public void setIpFamilies(List<IpFamily> ipFamilies) {
...
+    @JsonInclude(JsonInclude.Include.NON_NULL)
```
scholzj, line 235: Can you set this to NON_DEFAULT? That way it should not always render into the YAML unless it would be set to false (and thus not change the YAMLs which already exists without the value.)

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/ListenersValidator.java
```diff
@@ -196,6 +198,19 @@ private static void validateServiceDnsDomain(Set<String> errors, GenericKafkaLis
...
 
+    /**
+     * Validates that createBootstrapService is used only with Load Balancer type listener
+     *
+     * @param errors    List where any found errors will be added
+     * @param listener  Listener which needs to be validated
+     */
+    private static void validateCreateBootstrapService(Set<String> errors, GenericKafkaListener listener) {
+        if (KafkaListenerType.LOADBALANCER != listener.getType()
+                && listener.getConfiguration() != null && !listener.getConfiguration().getCreateBootstrapService())    {
```
scholzj, line 209: Only a nit ... but might be easier to read if you split it into two lines:

```java
        if (KafkaListenerType.LOADBALANCER != listener.getType()
                && listener.getConfiguration() != null 
                && !listener.getConfiguration().getCreateBootstrapService())    {
```

api/src/main/java/io/strimzi/api/kafka/model/listener/arraylistener/GenericKafkaListenerConfiguration.java
```diff
@@ -229,6 +230,17 @@ public void setIpFamilies(List<IpFamily> ipFamilies) {
         this.ipFamilies = ipFamilies;
     }
 
+    @Description("Whether to create the bootstrap service or not." +
```
scholzj, line 233: You need to add a space after the `.`. Otherwise in the docs this is then rendered as `Whether to create the bootstrap service or not.This field can be used with the loadBalancer type listener.` without the space between the sentences.

scholzj (commented):
I left some nits. But it looks great otherwise. I will also trigger the regression tests.

scholzj:
PS: The regression failure is not related to this PR.

scholzj:
It looks like it needs to update the CRDs:

```
ERROR: Uncommitted changes in derived resources:
M	documentation/modules/appendix_crds.adoc
M	packaging/helm-charts/helm3/strimzi-kafka-operator/crds/040-Crd-kafka.yaml
M	packaging/install/cluster-operator/040-Crd-kafka.yaml
Run the following to add up-to-date resources:
  mvn clean verify -DskipTests -DskipITs \
    && make crd_install \
    && make helm_install \
    && git add packaging/install/ packaging/helm-charts/ documentation/modules/appendix_crds.adoc cluster-operator/src/main/resources/cluster-roles \
    && git commit -s -m 'Update derived resources'
```

yyang48:
> It looks like it needs to update the CRDs:
> 
> ```
> ERROR: Uncommitted changes in derived resources:
> M	documentation/modules/appendix_crds.adoc
> M	packaging/helm-charts/helm3/strimzi-kafka-operator/crds/040-Crd-kafka.yaml
> M	packaging/install/cluster-operator/040-Crd-kafka.yaml
> Run the following to add up-to-date resources:
>   mvn clean verify -DskipTests -DskipITs \
>     && make crd_install \
>     && make helm_install \
>     && git add packaging/install/ packaging/helm-charts/ documentation/modules/appendix_crds.adoc cluster-operator/src/main/resources/cluster-roles \
>     && git commit -s -m 'Update derived resources'
> ```

Yes, I just fixed it. btw, I really like this error message, it even give me the command to fix it :-)

scholzj (approved):

scholzj:
(ST failures are not related to this PR)

api/src/main/java/io/strimzi/api/kafka/model/listener/arraylistener/GenericKafkaListenerConfiguration.java
```diff
@@ -229,6 +230,17 @@ public void setIpFamilies(List<IpFamily> ipFamilies) {
         this.ipFamilies = ipFamilies;
     }
 
+    @Description("Whether to create the bootstrap service or not. " +
+            "This field can be used with the `loadBalancer` type listener.")
```
ppatierno, line 234: Maybe it's worth adding that it's created by default (if not specified differently).

ppatierno (approved):
LGTM. Just a nit.

scholzj:
Thanks for the PR. Great work!

PR #6180, opened 2022-01-13
Added fix for renewal of user provided CAs

### LINKED ISSUE #5466: The renewal process with custom CA does not work
### Discussed in https://github.com/strimzi/strimzi-kafka-operator/discussions/5437

<div type='discussions-op-text'>

<sup>Originally posted by **FrankWang1108** August 17, 2021</sup>
**Describe the bug**
This happens when we try to install our own CA when the strimzi is running and the strimzi ca is in use. It could also happen when updating the CA.

It happens after uploading new CA and certs into the *-cluster-ca-certs.
Following the [renew own CA document](https://strimzi.io/docs/operators/in-development/using.html#renewing-your-own-ca-certificates-str), we have both ca.crt and ca-dates.crt in the cluster-ca-certs, and because the keystore and old ca-dates.crt can trust each other, there should still have connection. This is what is expected and the connection between zk and broker are stable because of this.

However the ClusterOperator only uses ca.crt to check for connection between zk and broker. Therefore after uoloading the new CA and certs. The ClusterOperator loses connection to zk and broker, it tries to connect to it multiple times until eventually it will perform a rolling update for zk and broker.
Once the zk updated, it will pickup the new keystore, then it will be able to connect to ClusterOperator again.

**To Reproduce**
1. bring up a cluster
2. change the cluster-ca and cluster-ca-cert to your own ca
3. check the log of ClusterOperator to see it trying to connect to zk and broker.

**Expected behavior**
I would expect when the ClusterOperator trying to connect to zk and broker, it will use all .crt files in the cluster-ca-cert to check for connection, instead of just using ca.crt.
I would expect it to stay connected and trigger a rolling update because it sees new certs updates, not because it was unable to connect.

 - Strimzi version: 0.25.1 (but I think the bug is still there for newer versions)
 - Installation method: YAML files

</div>

ppatierno:
### Type of change

- Bugfix

### Description

This PR fixes #5466.
Today, when a user provides his own CA certificate (cluster or client) but then he tries to follow the [renew process](https://strimzi.io/docs/operators/in-development/using.html#renewing-your-own-ca-certificates-str) as we documented, the procedure doesn't work at all.
The current code doesn't have logic for the CO to detect that a new CA certificate was provided by the user (by following the documentation and having a `Secret` with both the new CA certificate together with the old one CA-<expiration-date>).
With this PR, this first detection issue is resolved by adding the thumbprint of the cluster CA certificate, as an annotation, on all the `Secret`(s) that contains server certificates signed by that CA certificate. For example, the `<cluster>-zookeeper-nodes`, `<cluster>-kafka-brokers` and the same for the CO and EO (TO and UO) related `Secret`(s).

```yaml
annotations:
    clients-ca-thumbprint: hCyUdeqWt6gXDUKYdAtaFYDiPoZlWAwCjW7liqtWkps=
```

Just for the Kafka brokers related `Secret`(s) there is also a corresponding annotation with the thumbprint of the clients CA certificate.

```yaml
annotations:
    clients-ca-thumbprint: hCyUdeqWt6gXDUKYdAtaFYDiPoZlWAwCjW7liqtWkps=
    cluster-ca-thumbprint: DoNswf/bRfs6nF1IVx7PhbOVXKvJvjv/wOxuQY8eoTM=
```

During the periodic reconcile, the CO is able to detect that the thumbprint of the provided CA certificate is different from the one stored in such `Secret`(s).
When the change is detected, it just sets the renewal type as `REPLACE_KEY`, starting the same flow which runs when the CA certificates are handled by Strimzi itself (of course a part the CA certificate generation that is skipped).
So it runs the right pods rolls to first trust the new CA certificate and then generating the new server certificates for the different components (CO itself, Kafka, ZooKeeper, TO, UO).

In order to do so there are a couple of things to take into account that needs the documentation to be updated because the user has to follow them in order to renew the CA certificate properly:

1. The CN (Common Name) of the current CA certificate and the renewed one have to be different. This is because when they will be both trusted, the openssl tool used by stunnel in the tls-sidecar (in the TO) is not able to use both of them to trust the ZooKeeper server certificates but just one of them, so the verification would fail. When the Strimzi auto-creation of CA certificate is used, this is exactly what happens; a new CA is created with a suffix of `v1` (compared to the previous `v0`) in the CN to make them different, so manual renewal should follow same path.
2. The `Secret` containing the CA certificate has to have the `strimzi.io/ca-cert-generation` annotation as well as the one containing the key has to have the `strimzi.io/ca-key-generation`. This generation should start from 0 and increased on each renewal; even this time this is exactly how the auto-renewal process works in Strimzi doing that for you.

So the above steps need to be documented to make "handmade" renewal the same as the "auto" renewal from Strimzi operator.

Fixing this way, I made a couple of assumptions:

* when renewing the CA certificate, the user will always create a new key (as we describe in the documentation) so the reason why the fix inject into the flow with a `REPLACE_KEY` renewal type.
* the CA-<expiration-date> is not deleted by the CO automatically when it expires (as it happens for Strimzi generated CA certificate). It's the user to delete it from the `Secret`.

Documentation will come with a different PR.

### Checklist

- [x] Update CHANGELOG.md

### DIFF
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/model/ClusterCa.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/model/ClusterCa.java
@@ -242,18 +242,18 @@ public class ClusterCa extends Ca {
     }
 
     @Override
-    protected String caCertThumbprintAnnotation() {
-        return ANNO_STRIMZI_IO_CLUSTER_CA_THUMBPRINT;
+    protected String caCertGenerationAnnotation() {
+        return ANNO_STRIMZI_IO_CLUSTER_CA_CERT_GENERATION;
     }
 
     @SuppressWarnings("BooleanExpressionComplexity")
     @Override
-    protected boolean isCaCertThumbprintChanged() {
+    protected boolean isCaCertGenerationChanged() {
         // at least one Secret has a different cluster CA certificate thumbprint.
         // it is useful when a renewal cluster CA certificate process needs to be recovered after an operator crash
-        return isCaCertThumbprintChanged(zkNodesSecret) || isCaCertThumbprintChanged(brokersSecret) ||
-                isCaCertThumbprintChanged(entityTopicOperatorSecret) || isCaCertThumbprintChanged(entityUserOperatorSecret) ||
-                isCaCertThumbprintChanged(kafkaExporterSecret) || isCaCertThumbprintChanged(cruiseControlSecret) ||
-                isCaCertThumbprintChanged(clusterOperatorSecret);
+        return isCaCertGenerationChanged(zkNodesSecret) || isCaCertGenerationChanged(brokersSecret) ||
+                isCaCertGenerationChanged(entityTopicOperatorSecret) || isCaCertGenerationChanged(entityUserOperatorSecret) ||
+                isCaCertGenerationChanged(kafkaExporterSecret) || isCaCertGenerationChanged(cruiseControlSecret) ||
+                isCaCertGenerationChanged(clusterOperatorSecret);
     }
 }
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/model/ClusterCa.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/model/ClusterCa.java
@@ -248,12 +248,12 @@ public class ClusterCa extends Ca {
 
     @SuppressWarnings("BooleanExpressionComplexity")
     @Override
-    protected boolean isCaCertGenerationChanged() {
+    protected boolean hasCaCertGenerationChanged() {
         // at least one Secret has a different cluster CA certificate thumbprint.
         // it is useful when a renewal cluster CA certificate process needs to be recovered after an operator crash
-        return isCaCertGenerationChanged(zkNodesSecret) || isCaCertGenerationChanged(brokersSecret) ||
-                isCaCertGenerationChanged(entityTopicOperatorSecret) || isCaCertGenerationChanged(entityUserOperatorSecret) ||
-                isCaCertGenerationChanged(kafkaExporterSecret) || isCaCertGenerationChanged(cruiseControlSecret) ||
-                isCaCertGenerationChanged(clusterOperatorSecret);
+        return hasCaCertGenerationChanged(zkNodesSecret) || hasCaCertGenerationChanged(brokersSecret) ||
+                hasCaCertGenerationChanged(entityTopicOperatorSecret) || hasCaCertGenerationChanged(entityUserOperatorSecret) ||
+                hasCaCertGenerationChanged(kafkaExporterSecret) || hasCaCertGenerationChanged(cruiseControlSecret) ||
+                hasCaCertGenerationChanged(clusterOperatorSecret);
     }
 }
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/model/ClusterCa.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/model/ClusterCa.java
@@ -241,4 +241,19 @@ public class ClusterCa extends Ca {
             isMaintenanceTimeWindowsSatisfied);
     }
 
+    @Override
+    protected String caCertThumbprintAnnotation() {
+        return ANNO_STRIMZI_IO_CLUSTER_CA_THUMBPRINT;
+    }
+
+    @SuppressWarnings("BooleanExpressionComplexity")
+    @Override
+    protected boolean isCaCertThumbprintChanged() {
+        // at least one Secret has a different cluster CA certificate thumbprint.
+        // it is useful when a renewal cluster CA certificate process needs to be recovered after an operator crash
+        return isCaCertThumbprintChanged(zkNodesSecret) || isCaCertThumbprintChanged(brokersSecret) ||
+                isCaCertThumbprintChanged(entityTopicOperatorSecret) || isCaCertThumbprintChanged(entityUserOperatorSecret) ||
+                isCaCertThumbprintChanged(kafkaExporterSecret) || isCaCertThumbprintChanged(cruiseControlSecret) ||
+                isCaCertThumbprintChanged(clusterOperatorSecret);
+    }
 }
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaCluster.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaCluster.java
@@ -1357,9 +1357,11 @@ public class KafkaCluster extends AbstractModel {
      * internal communication with Zookeeper.
      * It also contains the related Kafka brokers private keys.
      *
+     * @param clusterCa The CA for cluster certificates
+     * @param clientsCa The CA for clients certificates
      * @return The generated Secret
      */
-    public Secret generateBrokersSecret() {
+    public Secret generateBrokersSecret(ClusterCa clusterCa, ClientsCa clientsCa) {
 
         Map<String, String> data = new HashMap<>(replicas * 4);
         for (int i = 0; i < replicas; i++) {
@@ -1369,7 +1371,10 @@ public class KafkaCluster extends AbstractModel {
             data.put(KafkaCluster.kafkaPodName(cluster, i) + ".p12", cert.keyStoreAsBase64String());
             data.put(KafkaCluster.kafkaPodName(cluster, i) + ".password", cert.storePasswordAsBase64String());
         }
-        return createSecret(KafkaCluster.brokersSecretName(cluster), data);
+        Map<String, String> annotations = Map.of(
+                clusterCa.caCertThumbprintAnnotation(), clusterCa.currentCaCertThumbprint(),
+                clientsCa.caCertThumbprintAnnotation(), clientsCa.currentCaCertThumbprint());
+        return createSecret(KafkaCluster.brokersSecretName(cluster), data, annotations);
     }
 
     /**
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaCluster.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaCluster.java
@@ -1371,9 +1371,10 @@ public class KafkaCluster extends AbstractModel {
             data.put(KafkaCluster.kafkaPodName(cluster, i) + ".p12", cert.keyStoreAsBase64String());
             data.put(KafkaCluster.kafkaPodName(cluster, i) + ".password", cert.storePasswordAsBase64String());
         }
+
         Map<String, String> annotations = Map.of(
-                clusterCa.caCertThumbprintAnnotation(), clusterCa.currentCaCertThumbprint(),
-                clientsCa.caCertThumbprintAnnotation(), clientsCa.currentCaCertThumbprint());
+                clusterCa.caCertGenerationAnnotation(), String.valueOf(clusterCa.certGeneration()),
+                clientsCa.caCertGenerationAnnotation(), String.valueOf(clientsCa.certGeneration()));
         return createSecret(KafkaCluster.brokersSecretName(cluster), data, annotations);
     }
 
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/model/ModelUtils.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/model/ModelUtils.java
@@ -144,7 +144,9 @@ public class ModelUtils {
             reasons.add("certificate doesn't exist yet");
             shouldBeRegenerated = true;
         } else {
-            if (clusterCa.keyCreated() || clusterCa.certRenewed() || (isMaintenanceTimeWindowsSatisfied && clusterCa.isExpiring(secret, keyCertName + ".crt"))) {
+            if (clusterCa.keyCreated() || clusterCa.certRenewed() ||
+                    (isMaintenanceTimeWindowsSatisfied && clusterCa.isExpiring(secret, keyCertName + ".crt")) ||
+                    clusterCa.isCaCertThumbprintChanged(secret)) {
                 reasons.add("certificate needs to be renewed");
                 shouldBeRegenerated = true;
             }
@@ -191,7 +193,8 @@ public class ModelUtils {
             data.put(keyCertName + ".password", certAndKey.storePasswordAsBase64String());
         }
 
-        return createSecret(secretName, namespace, labels, ownerReference, data, emptyMap(), emptyMap());
+        return createSecret(secretName, namespace, labels, ownerReference, data,
+                Collections.singletonMap(clusterCa.caCertThumbprintAnnotation(), clusterCa.currentCaCertThumbprint()), emptyMap());
     }
 
     public static Secret createSecret(String name, String namespace, Labels labels, OwnerReference ownerReference,
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/model/ModelUtils.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/model/ModelUtils.java
@@ -146,7 +146,7 @@ public class ModelUtils {
         } else {
             if (clusterCa.keyCreated() || clusterCa.certRenewed() ||
                     (isMaintenanceTimeWindowsSatisfied && clusterCa.isExpiring(secret, keyCertName + ".crt")) ||
-                    clusterCa.isCaCertGenerationChanged(secret)) {
+                    clusterCa.hasCaCertGenerationChanged(secret)) {
                 reasons.add("certificate needs to be renewed");
                 shouldBeRegenerated = true;
             }
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/model/ModelUtils.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/model/ModelUtils.java
@@ -146,7 +146,7 @@ public class ModelUtils {
         } else {
             if (clusterCa.keyCreated() || clusterCa.certRenewed() ||
                     (isMaintenanceTimeWindowsSatisfied && clusterCa.isExpiring(secret, keyCertName + ".crt")) ||
-                    clusterCa.isCaCertThumbprintChanged(secret)) {
+                    clusterCa.isCaCertGenerationChanged(secret)) {
                 reasons.add("certificate needs to be renewed");
                 shouldBeRegenerated = true;
             }
@@ -194,7 +194,7 @@ public class ModelUtils {
         }
 
         return createSecret(secretName, namespace, labels, ownerReference, data,
-                Collections.singletonMap(clusterCa.caCertThumbprintAnnotation(), clusterCa.currentCaCertThumbprint()), emptyMap());
+                Collections.singletonMap(clusterCa.caCertGenerationAnnotation(), String.valueOf(clusterCa.certGeneration())), emptyMap());
     }
 
     public static Secret createSecret(String name, String namespace, Labels labels, OwnerReference ownerReference,
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaAssemblyOperator.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaAssemblyOperator.java
@@ -611,6 +611,7 @@ public class KafkaAssemblyOperator extends AbstractAssemblyOperator<KubernetesCl
                         Secret clusterCaKeySecret = null;
                         Secret clientsCaCertSecret = null;
                         Secret clientsCaKeySecret = null;
+                        Secret brokersSecret = null;
                         List<Secret> clusterSecrets = secretOperations.list(reconciliation.namespace(), selectorLabels);
                         for (Secret secret : clusterSecrets) {
                             String secretName = secret.getMetadata().getName();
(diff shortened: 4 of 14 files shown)


cluster-operator/src/main/java/io/strimzi/operator/cluster/model/ClusterCa.java
```diff
@@ -241,4 +241,15 @@ public Secret cruiseControlSecret() {
             isMaintenanceTimeWindowsSatisfied);
     }
 
+    @Override
+    protected boolean isCaCertThumbprintChanged() {
+        if (zkNodesSecret != null) {
+            String clusterCaThumbprintAnno = zkNodesSecret.getMetadata().getAnnotations().get("cluster-ca-thumbprint");
```
see-quick, line 247: Just curious...the method invocation of `isCaCertThumbprintChanged` cannot produce here NPE? (there will always be in annotations `cluster-ca-thumbprint` when the method is invoked?)

ppatierno: Yeah when the Secret is created, it always has the annotation.

scholzj: And what about upgrade?

ppatierno: Damn ... good point, coming from an older release of the Strimzi operator! I will check this, thanks guys.

scholzj:
> The Secret containing the CA certificate has to have the strimzi.io/ca-cert-generation annotation as well as the one containing the key has to have the strimzi.io/ca-key-generation. This generation should start from 0 and increased on each renewal; even this time this is exactly how the auto-renewal process works in Strimzi doing that for you.

So, how does this deal with existing clusters which will not have the generation set? You can easily tell new users to set it. But we need to make sure that when an existing user wich custom CA secret without the generation upgrades, it doesn't fall all over.

.checkstyle/checkstyle.xml
```diff
@@ -107,7 +107,7 @@
 
         <module name="ClassFanOutComplexity">
             <!-- default is 20 -->
-            <property name="max" value="44"/>
+            <property name="max" value="45"/>
```
scholzj, line 110: Would using an annotation to suppress the issue make more sense than changing the defaults?

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/ClusterCa.java
```diff
@@ -241,4 +241,15 @@ public Secret cruiseControlSecret() {
             isMaintenanceTimeWindowsSatisfied);
     }
 
+    @Override
+    protected boolean isCaCertThumbprintChanged() {
```
scholzj, line 245: This seems weird. The method has a generic name, but it really checks only the ZooKeeper secret. Is that intentional? What about the other secrets? Also, this is a bit of a liability once we get rid of ZooKeeper.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaAssemblyOperator.java
```diff
@@ -2652,7 +2655,7 @@ String getCertificateThumbprint(Secret certSecret, CertAndKeySecretSource custom
         }
 
         Future<ReconciliationState> kafkaBrokersSecret() {
-            return updateCertificateSecretWithDiff(KafkaCluster.brokersSecretName(name), kafkaCluster.generateBrokersSecret())
+            return updateCertificateSecretWithDiff(KafkaCluster.brokersSecretName(name), kafkaCluster.generateBrokersSecret(clusterCa, clientsCa))
```
scholzj, line 2658: Why are you not changing it also for the other compoenents? UO, TO, KE, CC, ...

scholzj:
> Just for the Kafka brokers related Secret(s) there is also a corresponding annotation with the thumbprint of the clients CA certificate.

This seems a bit weird ... since the secret has nothing to do with the Clients CA. Will it stay in sync? Because the secret will not change when the Clients CA is renewed.

.checkstyle/checkstyle.xml
ppatierno: Agree. I reverted it back, but the suppress annotation seems to be not needed anymore. I guess at some point I changed the code and increasing the fan out complexity was not needed anymore :-)

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/ClusterCa.java
scholzj: I think it is not just the Zoo removal what would be the problem here. One thing to keep in mind is restartability of the process. I.e., what happens if the operator fails in the middle before updating and rolling all components.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaAssemblyOperator.java
ppatierno: So ... UO, TO and KE uses `ModeUtils.buildSecret` to build their Secrets and I updated that. Good point about CC and the others I will check.

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/ClusterCa.java
ppatierno: Good point. It means that currently if after the first ZK rolls, the new CA certificate is trusted by ZK pods and the ZK Secret is updated with the new thumbprint but the CO crashes, it won't recover to roll K and the other pods. The simpler solution I see is about checking the thumbprint of CA certificate on all the Secrets in the `ClusterCA` to be sure that if at least one is different, the `renewalType=REPLACE_KEY` is injected correctly and the flow can proceed. Does it make sense?

scholzj: In general => but I do not remember the code good enough to know when exactly are the secrets updated. The part it will likely fail are the rolling updates, bot the secret updates. So checking of the secrets works fine if they are updated only right before the roll.

ppatierno:
> This seems a bit weird ... since the secret has nothing to do with the Clients CA. Will it stay in sync? Because the secret will not change when the Clients CA is renewed.

But the thumbprint is update; the flow follows the Secret creation as usual by creating the same one but with a different thumbprint. I just added that to the usual build Secret creation in this case.

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/ClusterCa.java
ppatierno: So on first roll, for trusting the new CA certificate, Secret is not update. On the second roll, the Secret is updated with the new CA certificate thumbprint right after rolling the pods.

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/ModelUtils.java
```diff
@@ -144,7 +144,7 @@ public static Secret buildSecret(Reconciliation reconciliation, ClusterCa cluste
             reasons.add("certificate doesn't exist yet");
             shouldBeRegenerated = true;
         } else {
-            if (clusterCa.keyCreated() || clusterCa.certRenewed() || (isMaintenanceTimeWindowsSatisfied && clusterCa.isExpiring(secret, keyCertName + ".crt"))) {
+            if (clusterCa.keyCreated() || clusterCa.certRenewed() || (isMaintenanceTimeWindowsSatisfied && clusterCa.isExpiring(secret, keyCertName + ".crt")) || clusterCa.isCaCertThumbprintChanged(secret)) {
```
tombentley, line 147: Can we break this long line?

operator-common/src/main/java/io/strimzi/operator/cluster/model/Ca.java
```diff
@@ -403,6 +405,12 @@ public CertAndKey generateSignedCert(String commonName, String organization) thr
                 reasons.add("certificate added");
             }
 
+            /*
+            if (isCaCertThumbprintChanged(secret)) {
+                reasons.add("new certificate generated");
+            }
+            */
```
tombentley, line 412: I guess this should be removed?

operator-common/src/main/java/io/strimzi/operator/cluster/model/Ca.java
```diff
@@ -520,6 +528,7 @@ public void createRenewOrReplace(String namespace, String clusterName, Map<Strin
         if (!generateCa) {
             certData = caCertSecret != null ? caCertSecret.getData() : emptyMap();
             keyData = caKeySecret != null ? singletonMap(CA_KEY, caKeySecret.getData().get(CA_KEY)) : emptyMap();
+            renewalType = isCaCertThumbprintChanged() ? RenewalType.REPLACE_KEY : RenewalType.NOOP;
```
tombentley, line 531: Can we not use the policy defined in the CR?

operator-common/src/main/java/io/strimzi/operator/cluster/model/Ca.java
```diff
@@ -733,6 +742,21 @@ public String currentCaCertBase64() {
         return decoder.decode(caKeySecret().getData().get(CA_KEY));
     }
 
+    /**
+     * @return The base64 encoded thumbprint of the current CA certificate.
+     */
+    public String currentCaCertThumbprint() {
+        try {
+            X509Certificate cert = Ca.x509Certificate(currentCaCertBytes());
+            byte[] signature = MessageDigest.getInstance("SHA-256").digest(cert.getEncoded());
```
tombentley, line 751: Is there a reason to use SHA-256? Openssl and other tools use SHA-1. Maybe there's some value in our thumbprints being the same as theirs?

operator-common/src/main/java/io/strimzi/operator/cluster/model/Ca.java
```diff
@@ -1011,4 +1035,21 @@ private void renewCaCert(Subject subject, Map<String, String> certData) {
             throw new RuntimeException(e);
         }
     }
+
+    protected abstract String caCertThumbprintAnnotation();
+
+    protected abstract boolean isCaCertThumbprintChanged();
+
+    protected boolean isCaCertThumbprintChanged(Secret secret) {
```
tombentley, line 1043: Javadoc please

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/ClusterCa.java
```diff
@@ -241,4 +241,19 @@ public Secret cruiseControlSecret() {
             isMaintenanceTimeWindowsSatisfied);
     }
 
+    @Override
+    protected String caCertThumbprintAnnotation() {
+        return "cluster-ca-thumbprint";
```
scholzj, line 246: Should this be defined as constanct somewhere in `Annotations`?

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaCluster.java
```diff
@@ -1311,7 +1313,10 @@ public Secret generateBrokersSecret() {
             data.put(KafkaCluster.kafkaPodName(cluster, i) + ".p12", cert.keyStoreAsBase64String());
             data.put(KafkaCluster.kafkaPodName(cluster, i) + ".password", cert.storePasswordAsBase64String());
         }
-        return createSecret(KafkaCluster.brokersSecretName(cluster), data);
+        Map<String, String> annotations = Map.of(
+                clusterCa.caCertThumbprintAnnotation(), clusterCa.currentCaCertThumbprint(),
+                clientsCa.caCertThumbprintAnnotation(), clientsCa.currentCaCertThumbprint());
+        return createSecret(KafkaCluster.brokersSecretName(cluster), data, annotations);
```
scholzj, line 1319: We seem to be adding the annotation here, for Zoo and for CC. But not for EO or for KE. Is the implementation of the secrets different there? Or why don't we need to do it?

operator-common/src/main/java/io/strimzi/operator/cluster/model/ClientsCa.java
```diff
@@ -37,6 +40,20 @@ public static Secret adapt060ClientsCaSecret(Secret clientsCaKey) {
         return clientsCaKey;
     }
 
+    public void initCaSecrets(Secret secret) {
+        this.brokersSecret = secret;
+    }
+
+    @Override
+    protected String caCertThumbprintAnnotation() {
+        return "clients-ca-thumbprint";
```
scholzj, line 49: Same as earlier. I think this would be good in `Annotations` to have all annos centrally. (not necessarily the method => just the string definition)

operator-common/src/main/java/io/strimzi/operator/cluster/model/Ca.java
ppatierno: We actually don't know what the user will do, if generating the new CA using the same key or using a new key. Setting the more strict policy (so REPLACE_KEY) means that we should get enough rolling of our pods for the overall renewal process. We are "blind" here, so it's playing safer this way.

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/ClusterCa.java
ppatierno: Yeah it makes sense even if I preferred to add the annotations this way in the `Ca` where other CA management related annotations live:

```java
public static final String ANNO_STRIMZI_IO_CLUSTER_CA_THUMBPRINT = Annotations.STRIMZI_DOMAIN + "cluster-ca-thumbprint";
    public static final String ANNO_STRIMZI_IO_CLIENTS_CA_THUMBPRINT = Annotations.STRIMZI_DOMAIN + "clients-ca-thumbprint";
```

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaCluster.java
ppatierno: EO and KE use the `ModeUtils#buildSecret` method to build the Secret and it's updated to add the annotations.

operator-common/src/main/java/io/strimzi/operator/cluster/model/ClientsCa.java
ppatierno: Yeah, ditto as above.

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/ClusterCa.java
scholzj: If the other annos are there already then yea, it might make more sense to add them there.

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/ClusterCa.java
```diff
@@ -241,4 +241,19 @@ public Secret cruiseControlSecret() {
...
+    @SuppressWarnings("BooleanExpressionComplexity")
+    @Override
+    protected boolean isCaCertThumbprintChanged() {
+        // at least one Secret has a different cluster CA certificate thumbprint.
+        // it is useful when a renewal cluster CA certificate process needs to be recovered after an operator crash
+        return isCaCertThumbprintChanged(zkNodesSecret) || isCaCertThumbprintChanged(brokersSecret) ||
+                isCaCertThumbprintChanged(entityTopicOperatorSecret) || isCaCertThumbprintChanged(entityUserOperatorSecret) ||
+                isCaCertThumbprintChanged(kafkaExporterSecret) || isCaCertThumbprintChanged(cruiseControlSecret) ||
+                isCaCertThumbprintChanged(clusterOperatorSecret);
+    }
```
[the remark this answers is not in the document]
ppatierno: I was thinking about putting the CA cert thumbprint annotation on the ZK and K pods (so on `StatefulSet` and through the `StrimziPodSet`) as it happens today for the CA cert generation. This could help the `KafkaAssemblyOperator#getReasonsToRestartPod` to check that even if `isCaCertThumbprintChanged` is true, we have to check specific pods annotation because them could be already updated so no need to roll.
It's also true that during the first roll about trusting the new CA, the thumbprint annotation is not updated but it happens on the second one on certificates generation. :-(

operator-common/src/main/java/io/strimzi/operator/cluster/model/ClientsCa.java
```diff
@@ -37,6 +40,20 @@ public static Secret adapt060ClientsCaSecret(Secret clientsCaKey) {
         return clientsCaKey;
     }
 
+    public void initCaSecrets(Secret secret) {
+        this.brokersSecret = secret;
+    }
```
scholzj, line 45: Its `initCaSecrets` in plural. So, should it mirror the ClusterCA method and take a list of secrets despite currently needing only one secret from the list? The fact that you use the same name but different parameter and behaviour seems weird. I would either rename ti to match the actual purpose (e.g. `initBrokerSecret` or `setBrokerSecret`) or mirror the behaviour of the ClusterCa class and just pick the broker secret.

scholzj (approved):
One nit wich I think you should address. LGTM otherwise.

see-quick (approved):
The errors from Azure are not related to this PR and should be fixed here https://github.com/strimzi/strimzi-kafka-operator/pull/6245

operator-common/src/main/java/io/strimzi/operator/cluster/model/ClientsCa.java
ppatierno: As discussed offline, It's kind of problematic doing the same as ClusterCa because we would need the cluster name to filter the corresponding broker Secret name. It would be ok if ClientsCa would be used in the CO only but it's used in the UO which can run as standalone as well so we cannot rely on the `strimzi.io/cluster` label selector for getting such an info. I will go to just rename the method in a better way.

tombentley (commented):
Reading the companion doc PR (https://github.com/strimzi/strimzi-kafka-operator/pull/6240) made me wonder whether we should really be using an extra annotation for the thumbprint when we already have the generation annotation. I'd like to understand if the generation on its own is enough to track the changing dependency (and if not why not).

ppatierno:
@tombentley it's a good point ... thinking more about it, the generation annotation is on the CA certificate Secret only so when it's replaced by a new CA with a newer generation we lost that information and the CO cannot infer that CA is changed. The thumbprint annotation is on the Secrets (ZK, K, EO, CO, ...) containing server certificates signed with the CA certificate so that on reconcile we can compare the new CA thumbprint with the one on those Secrets. Using the generation could work but not out of box, we would need to store it on the Secrets (instead of the thumbprint) as well to be able to make a comparison and detect the CA certificate update.

ppatierno:
@tombentley @scholzj I changed the code by saving the `strimzi.io/cluster-ca-cert-generation` and `strimzi.io/clients-ca-cert-generation` (only for Kafka) on Secrets instead of the thumbprint. I run some tests and of course CO is able to detect change because the user increments the generation on renewal. I don't see any drawbacks.

cluster-operator/src/test/java/io/strimzi/operator/cluster/model/CaRenewalTest.java
```diff
@@ -199,6 +219,16 @@ protected CertAndKey generateSignedCert(Subject subject,
...
+
+            @Override
+            protected boolean isCaCertGenerationChanged() {
+                return false;
+            }
+
+            @Override
+            protected String caCertGenerationAnnotation() {
+                return null;
+            }
```
tombentley, line 231: Aside: I wish there were a better alternative to using anonymous subclasses of Ca for this test. It makes it harder to see how the tests are different. Could you see if there's scope for collecting some of them together into named subclasses at least, rather than having a different subclass per test?

operator-common/src/main/java/io/strimzi/operator/cluster/model/Ca.java
```diff
@@ -733,6 +736,21 @@ public String currentCaCertBase64() {
         return decoder.decode(caKeySecret().getData().get(CA_KEY));
     }
 
+    /**
+     * @return The base64 encoded thumbprint of the current CA certificate.
+     */
+    public String currentCaCertThumbprint() {
```
tombentley, line 742: This is now unused.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaAssemblyOperator.java
```diff
@@ -654,14 +657,13 @@ ReconciliationState createReconciliationState(Reconciliation reconciliation, Kaf
...
                                 ModelUtils.getRenewalDays(clusterCaConfig),
                                 clusterCaConfig == null || clusterCaConfig.isGenerateCertificateAuthority(), clusterCaConfig != null ? clusterCaConfig.getCertificateExpirationPolicy() : null);
+                        this.clusterCa.initCaSecrets(clusterSecrets);
                         clusterCa.createRenewOrReplace(
                                 reconciliation.namespace(), reconciliation.name(), caLabels.toMap(),
                                 clusterCaCertLabels, clusterCaCertAnnotations,
                                 clusterCaConfig != null && !clusterCaConfig.isGenerateSecretOwnerReference() ? null : ownerRef,
                                 isMaintenanceTimeWindowsSatisfied(dateSupplier));
 
-                        this.clusterCa.initCaSecrets(clusterSecrets);
```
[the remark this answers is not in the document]
ppatierno: because the `createRenewOrReplace` is going to call `hasCaCertGenerationChanged` which relies on Secret(s) storing the CA certificate generation to check if it's changed and needs to start the renewal procedure.

tombentley (approved):
Thanks @ppatierno

cluster-operator/src/test/java/io/strimzi/operator/cluster/model/CaRenewalTest.java
ppatierno: Ok. Just done. I factored out a `MockedCa` inner class that should work as you suggested.

scholzj (approved):

PR #6181, opened 2022-01-13
fix indenting in helm deployment cluster operator

BobClaerhout:
### Type of change

_Select the type of your PR_

- Bugfix

### Description

When using the strimi-cluster-operator as a dependency in another chart, you cannot configure the extraEnvs because the indenting is wrong.  
In the meantime I also updated any existing `indent` usages in the file.

### Checklist

- [ ] Write tests
- [ ] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/packaging/helm-charts/helm3/strimzi-kafka-operator/templates/060-Deployment-strimzi-cluster-operator.yaml
+++ b/packaging/helm-charts/helm3/strimzi-kafka-operator/templates/060-Deployment-strimzi-cluster-operator.yaml
@@ -129,7 +129,7 @@ spec:
               value: {{ .Values.connectBuildTimeoutMs | quote }}
             {{- end }}
             {{- if .Values.extraEnvs }}
-{{ toYaml .Values.extraEnvs | indent 12 }}
+{{ toYaml .Values.extraEnvs | nindent 12 }}
             {{- end }}
           livenessProbe:
             httpGet:
@@ -147,7 +147,7 @@ spec:
           securityContext: {{ toYaml .| nindent 12 }}
           {{- end }}
           resources:
-{{ toYaml .Values.resources | indent 12 }}
+{{ toYaml .Values.resources | nindent 12 }}
       {{- with .Values.nodeSelector }}
       nodeSelector:
         {{- toYaml . | nindent 8 }}


scholzj (commented):
Thanks for the PR. However, you cannot change the released files outside of a proper release. You should do any changes to the Helm Chart which is in the `packaging/` folder. Thanks.

BobClaerhout:
> Thanks for the PR. However, you cannot change the released files outside of a proper release. You should do any changes to the Helm Chart which is in the `packaging/` folder. Thanks.

Ah, didn't read the `DO_NOT_EDIT.md` file properly... Thanks for the quick response. I updated it. Do you think it's possible to patch the latest helm release?

ShubhamRwt:
@BobClaerhout Hi, Doesn't `nindent` add a new line ? I guess `indent` should have worked fine. Can you tell me the error you got ?

scholzj (commented):
Can you please elaborate a bit more on what the problem is and how does this fix it? Maybe provide some easy to reproduce examples of what does not work?

In 0.27.0, there is indeed a bug around the `extraEnvs`. But that was already fixed with #6147 by @ShubhamRwt. I'm not aware of any error with the resources (which have defaults and if they would not work, it would fail for everyone using the Helm Chart, but unlike the `extraEnvs`,  nobody reported anything).

When I run the changes from your PR, I do not see any difference from what was generated before. The resources are exactly the same. So I think we need to really understand what does this actually fix.

Thanks.

BobClaerhout:
`nindent` indeed adds a new line. However, that doesn't matter in helm charts. More importantly, it makes sure that the indent is actually what you want. When you are using the strimi-kafka-operator helm chart as a dependency, you end up using following values files:
```
strimzi-kafka-operator:
  extraEnvs:
  - name: "myEnv"
    value: "myValue"
```
Mind the 2 additional spaces in front of the array values (line 3 and 4). These 2 additional spaces are causing some issues. When you use `nindent` the 2 additional spaces are removed.  
Additionally, you can indent the reference again as it was before https://github.com/strimzi/strimzi-kafka-operator/pull/6147 just for readability and "elegance".

ShubhamRwt:
Can you share those error logs? If we try to deploy operator using helm with the current code. I am not getting any issues

ShubhamRwt:
If we see the yaml template generating right now it looks like this:

            - name: STRIMZI_DEFAULT_JMXTRANS_IMAGE
              value: quay.io/strimzi/jmxtrans:latest
            - name: STRIMZI_DEFAULT_KANIKO_EXECUTOR_IMAGE
              value: quay.io/strimzi/kaniko-executor:latest
            - name: STRIMZI_DEFAULT_MAVEN_BUILDER
              value: quay.io/strimzi/maven-builder:latest
            - name: STRIMZI_OPERATOR_NAMESPACE
              valueFrom:
                fieldRef:
                  fieldPath: metadata.namespace
            
            - name: STRIMZI_FEATURE_GATES
              value: ""
            - name: JAVA_OPTS
              value: -Xms=256m -Xmx=256m

scholzj:
> When you are using the strimi-kafka-operator helm chart as a dependency ...

Sorry, I don't get it. Why should it do something else when used as a dependency? That sounds like a Helm issue to me.

BobClaerhout:
I have following files:
Chart.yaml:
```
apiVersion: v2
name: this-chart-dependes-on-strimzi
description: My chart for kubernetes
type: application
version: 1.0.0
appVersion: 1.0.0

dependencies:
- name: strimzi-kafka-operator
  repository: "https://strimzi.io/charts/"
  version: 0.27.0
```

values.yaml:
```
strimzi-kafka-operator:
  watchAnyNamespace: true
  extraEnvs:
  - name: JAVA_OPTS
    value: "-Dlog4j2.formatMsgNoLookups=true"
```
I do the following things:
```
helm dependency update
helm template ./ -f values.yaml --debug
```
As a result, in the charts folder, following file is downloaded: `strimzi-kafka-operator-helm-3-chart-0.27.0.tgz` and the template is generated. I'm getting following exception:
`Error: YAML parse error on iot-hub-kafka/charts/strimzi-kafka-operator/templates/060-Deployment-strimzi-cluster-operator.yaml: error converting YAML to JSON: yaml: line 100: did not find expected key`

When untar the strimzi-kafka-chart, execute the change which is included in this PR, tar and overwrite the existing chart.tgz file. and running `helm template ./ -f values.yaml --debug` again, I'm getting following result in for the deployment env variables:
```
....
            - name: STRIMZI_OPERATOR_NAMESPACE
              valueFrom:
                fieldRef:
                  fieldPath: metadata.namespace
            
            - name: STRIMZI_FEATURE_GATES
              value: ""
            
            - name: JAVA_OPTS
              value: -Dlog4j2.formatMsgNoLookups=true
.....
```
The difference between a "normal" helm template and a template using a dependency is the indenting on the values file. If you would install the helm chart directly and not using the dependency, you avoid the indent and you do not encounter the issue, hereby the `values.yaml` file when installing directly:
```
watchAnyNamespace: true
extraEnvs:
- name: JAVA_OPTS
  value: "-Dlog4j2.formatMsgNoLookups=true"
```

ShubhamRwt:
@BobClaerhout The bug was present in 2.7.0 and it was fixed in [#6417](https://github.com/strimzi/strimzi-kafka-operator/pull/6147). Did you tried the changes made in the this PR?

BobClaerhout:
Ah ok, yes. this was fixed in that PR as well. Same bug indeed. Sorry for the trouble

scholzj:
Ok, so if #6147 fixes it, that is great => I picked that up for a 0.27.1 release. So it should hopefully be fixed there for you as well.

Issue #3570, opened 2020-08-31
Report tool staleness


sknot-rh:
Pods running newer images of kafka and zookeeper do not contain tls-sidecar containers anymore. The report tool still tries to get their logs.
Also we are missing MM2 & Cruise control checks.

PR #6234, opened 2022-01-19
Fix tls-sidecar errors in report.sh

fvaleri:
In addition to what mentioned in the title, there are also some minor improvements.

### DIFF
--- a/tools/report.sh
+++ b/tools/report.sh
@@ -5,6 +5,7 @@ if [[ $(uname -s) == "Darwin" ]]; then
   alias echo="gecho"; alias grep="ggrep"; alias sed="gsed"; alias date="gdate"
 fi
 
+{ # this ensures the entire script is downloaded #
 NAMESPACE=""
 CLUSTER=""
 KUBECTL_INSTALLED=false
@@ -193,12 +194,12 @@ get_pod_logs() {
     if [[ $exists == true ]]; then
       local logs
       # get current logs
-      if [[ -n $con ]]; then logs="$($KUBE_CLIENT -n "$NAMESPACE" logs "$pod" -c "$con")"
-      else logs="$($KUBE_CLIENT -n "$NAMESPACE" logs "$pod")"; fi
+      if [[ -n $con ]]; then logs="$($KUBE_CLIENT -n "$NAMESPACE" logs "$pod" -c "$con" ||true)"
+      else logs="$($KUBE_CLIENT -n "$NAMESPACE" logs "$pod" ||true)"; fi
       if [[ -n $logs && -n $con ]]; then echo "$logs" > "$TMP"/reports/podlogs/"$pod"-"$con".log;
       elif [[ -n $logs && -z $con ]]; then echo "$logs" > "$TMP"/reports/podlogs/"$pod".log; fi
       # get previous logs if available
-      if [[ -n $con ]]; then logs="$($KUBE_CLIENT -n "$NAMESPACE" logs "$pod" -c "$con" -p 2>/dev/null ||true)"
+      if [[ -n $con ]]; then logs="$($KUBE_CLIENT -n "$NAMESPACE" logs "$pod" -p -c "$con" 2>/dev/null ||true)"
       else logs="$($KUBE_CLIENT -n "$NAMESPACE" logs "$pod" -p 2>/dev/null ||true)"; fi
       if [[ -n $logs && -n $con ]]; then echo "$logs" > "$TMP"/reports/podlogs/"$pod"-"$con".log.0;
       elif [[ -n $logs && -z $con ]]; then echo "$logs" > "$TMP"/reports/podlogs/"$pod".log.0; fi
@@ -294,3 +295,4 @@ zip -qr "$FILENAME".zip ./reports/
 cd "$OLD_DIR" || exit
 mv "$TMP"/"$FILENAME".zip ./
 echo "Report file $FILENAME.zip created"
+} # this ensures the entire script is downloaded #
--- a/tools/report.sh
+++ b/tools/report.sh
@@ -5,18 +5,18 @@ kubectl_installed=false
 platform="kubectl"
 secrets="hidden"
 
-oc &>/dev/null
-
-if [ $? -eq 0 ]; then
-	oc_installed=true
-	platform="oc"
-fi
+# sed delimiter
+readonly sd=$(echo -en "\001")
 
 kubectl &>/dev/null
-
-if [ $? -eq 0 ]; then # we will use kubectl with priority (?)
+if [ $? -eq 0 ]; then
 	kubectl_installed=true
-	platform="kubectl"
+else
+  oc &>/dev/null
+  if [ $? -eq 0 ]; then
+  	oc_installed=true
+  	platform="oc"
+  fi
 fi
 
 if [[ $oc_installed = false && $kubectl_installed = false ]]; then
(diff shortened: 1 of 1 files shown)


scholzj:
I don't know if it is the right time for this. The tool is not versioned. So it needs to work also with older versions. That is why we were postponing it. The TLS sidecars in Kafka and Zoo are gone for quite long now. So maybe removing those is fine. But the other sidecars are still there and should be kept. And how does it work with versions not supporting `StrimziPodSets`? Does it actually proceed? Or fail?

It is also missing Cruise Control and Mirror Maker 2: https://github.com/strimzi/strimzi-kafka-operator/issues/3570

fvaleri:
OK, I wrongly assumed it was versioned. 

I'm now fixing the "container not valid" errors on newer versions, and adding missing bits.

I also testing it on main and 0.22.1 versions.

scholzj:
Ok, thanks.

fvaleri:
There are other things that can be improved, but at least this fixes these ugly errors.

tools/report.sh
```diff
@@ -107,44 +106,43 @@ resources_to_fetch=(
...
 	)
 
 if [ "$secrets" = "off" ]; then
-  resources_to_fetch=( "${resources_to_fetch[@]/secrets}" )
+  resources_to_fetch=("${resources_to_fetch[@]/secrets}")
 fi
 
 get_masked_secrets() {
-	mkdir -p $direct/reports/"$1"
+	mkdir -p $direct/reports/$1
```
scholzj, line 119: Doublequoting is normally recommended everywhere. What is your reason for removing it? (not just here but in many other places) The script was written before we used shellcheck and does not pass currently. But IIRC this would be one of the things it would complain about. If we one day get to use it here as well, we might need to revert all these back?

fvaleri: So why not using quotes also for $direct? For me it's more about the context in which it is used that matters. Anyway, I understand the project standard, so I'll run shellcheck on it and fix any warning/error for future proof.

scholzj: Well, as I said, it does not pass shellcheck currently (even before your PR I mean). And you don't have to follow it to fix all the points. But we should not move in the oposite direction either.

fvaleri: Yeah, I agree.

fvaleri:
Now it should pass shell check and run smoothly on current and older releases.

tools/report.sh
```diff
@@ -1,276 +1,296 @@
...
+  echo "clusteroperator"
+  echo "    $CO_DEPLOY"
+  $KUBE_CLIENT get deploy strimzi-cluster-operator -o yaml -n "$NAMESPACE" > "$TMP"/reports/deployments/cluster-operator.yaml
+  $KUBE_CLIENT get po -l strimzi.io/kind=cluster-operator -o yaml -n "$NAMESPACE" > "$TMP"/reports/pods/cluster-operator.yaml
+  CO_POD=$($KUBE_CLIENT get po -l strimzi.io/kind=cluster-operator -o name -n "$NAMESPACE" --ignore-not-found)
+  if [[ -n $CO_POD ]]; then
+    echo "    $CO_POD"
+    CO_POD=$(echo "$CO_POD" | cut -d "/" -f 2) && readonly CO_POD
+    $KUBE_CLIENT logs "$CO_POD" -n "$NAMESPACE" > "$TMP"/reports/podlogs/cluster-operator.log
+    $KUBE_CLIENT logs "$CO_POD" -p -n "$NAMESPACE" 2>/dev/null > "$TMP"/reports/podlogs/cluster-operator.log.0
```
scholzj, line 242: Why didn't you kept the `previous-` naming? Also, unless I missed something ... why do you now capture the previous log only for the Cluster Operator and for the deployments but not for the statefulsets? (+ why does it use different mechanism in each place?)

scholzj (commented):
Some of the `sed` commands do not seem to work on MacOS. I get following errors:

```
secrets
    secret/my-cluster-clients-ca
    secret/my-cluster-clients-ca-cert
sed: 1: "s\001\s*.crt":"LS0tLS1CRUd ...": unterminated substitute pattern
    secret/my-cluster-cluster-ca
    secret/my-cluster-cluster-ca-cert
sed: 1: "s\001\s*.crt":"LS0tLS1CRUd ...": unterminated substitute pattern
    secret/my-cluster-cluster-operator-certs
sed: 1: "s\001\s*uster-operator.crt ...": unterminated substitute pattern
    secret/my-cluster-cruise-control-api
    secret/my-cluster-cruise-control-certs
sed: 1: "s\001\s*uise-control.crt": ...": unterminated substitute pattern
    secret/my-cluster-entity-topic-operator-certs
sed: 1: "s\001\s*tity-operator.crt" ...": unterminated substitute pattern
    secret/my-cluster-entity-user-operator-certs
sed: 1: "s\001\s*tity-operator.crt" ...": unterminated substitute pattern
    secret/my-cluster-kafka-brokers
sed: 1: "s\001\s*-cluster-kafka-0.c ...": unterminated substitute pattern
    secret/my-cluster-kafka-exporter-certs
sed: 1: "s\001\s*fka-exporter.crt": ...": unterminated substitute pattern
    secret/my-cluster-zookeeper-nodes
sed: 1: "s\001\s*-cluster-zookeeper ...": unterminated substitute pattern
```

and some of the secrets files are empty. Replacing `sed` with `gsed` seems to fix it. So I think you should load the `multi-platform-support.sh` and use it.

scholzj:
> Replacing sed with gsed seems to fix it. So I think you should load the multi-platform-support.sh and use it.

Correction. Using `gsed` gets rid of the errors, but it keeps a lot of values in the files. I guess that might be ok for `crt` files. But not for `p12 files which contain the private keys (the `.key` files are masked, but it helps a little if the p12 file is there including the password). Not sure if this was bug in the old version of if this is something what this PR breaks.

tools/report.sh
fvaleri: I used one of the common convention for naming rolled log files, with `.n` suffix (e.g. my.log, my.log.0, my.log.1, etc.). There are others, no preference at all.

fvaleri:
> > Replacing sed with gsed seems to fix it. So I think you should load the multi-platform-support.sh and use it.
> 
> Correction. Using `gsed` gets rid of the errors, but it keeps a lot of values in the files. I guess that might be ok for `crt` files. But not for `p12 files which contain the private keys (the `.key` files are masked, but it helps a little if the p12 file is there including the password). Not sure if this was bug in the old version of if this is something what this PR breaks.

I think `get_masked_secrets()` was already broken. I can fix that. How this should behave exactly?

tools/report.sh
fvaleri: There was no log capturing for the statefulsets in the previous revision. Should I add them?

fvaleri:
What about using the alias technique which you can find in cold-backup and log-dump scripts instead of multi-platform-support? I think it is much better than having to use variables.

fvaleri:
There is also another problem with using labels. For example, if I have connect deployed, logs are not downloaded because the pod is filtered out by this:

```shell
PODS=$($KUBE_CLIENT get pods -l strimzi.io/cluster="$CLUSTER" -o name -n "$NAMESPACE" | cut -d "/" -f 2) && readonly PODS
```

Example:

```shell
$ kubectl get po -l strimzi.io/cluster="my-cluster"
NAME                                          READY   STATUS    RESTARTS   AGE
my-cluster-entity-operator-6f6c7f6b58-lsnkw   3/3     Running   0          3h36m
my-cluster-kafka-0                            1/1     Running   0          3h38m
my-cluster-kafka-1                            1/1     Running   0          3h38m
my-cluster-kafka-2                            1/1     Running   0          3h38m
my-cluster-kafka-exporter-84589bf9fd-tlxwh    1/1     Running   0          3h35m
my-cluster-zookeeper-0                        1/1     Running   0          3h39m
my-cluster-zookeeper-1                        1/1     Running   0          3h39m
my-cluster-zookeeper-2                        1/1     Running   0          3h39m
```

Instead, it works with something like:

```shell
$ kubectl get po --show-labels | grep "strimzi.io" | awk '{print $1}'
my-cluster-entity-operator-6f6c7f6b58-lsnkw
my-cluster-kafka-0
my-cluster-kafka-1
my-cluster-kafka-2
my-cluster-kafka-exporter-84589bf9fd-tlxwh
my-cluster-zookeeper-0
my-cluster-zookeeper-1
my-cluster-zookeeper-2
my-connect-connect-6f766f7fb4-wx6g2 <----------
```

scholzj:
> What about using the alias technique which you can find in cold-backup and log-dump scripts instead of multi-platform-support? I think it is much better than having to use variables.

Doing different things in every script means things are much harder up to update, debug, fix, understnd. I think using the same what is everywhere else is always an advantage over creating custom solutions.

> There is also another problem with using labels. 

The tool is designed to capture a logs of a single cluster. That is why it also has the `--cluster` option which is required. Rewriting it to capture everything is potentially dangerous since it might capture things which people do not want to share / are not involved in the problem. I'm also not sure this really supports the other deployments and captures everything needed for them. It is not just about the pod logs.

tools/report.sh
scholzj: > I used one of the common conventions for naming rolled log files, with .n suffix (e.g. my.log, my.log.0, my.log.1, etc.). There are others, no preference at all.

TBH, I'm not sure where is that common. It might be eaiser to follow what we had here as common and maintain continuity unless there is a really strong reason for changing it => for example because tool processing the output would nor handle the `previous-` prefix etc. But I do not think there is anything like that.

> What do you mean with statefulset logs? I now created a single function to get pod/container log, including previous one if present.

Sorry, with StatfefulSet logs I meant Kafka and ZooKeeper logs. They did not seemed to be captured (the previous log).

fvaleri:
Ok, it should be fine now.

tools/report.sh
fvaleri: I see previous logs in my tests, but only when there is any. I'm avoiding to create empty files. Try to delete a pod.

fvaleri:
@scholzj the problem with sourcing multi-platform-support is that the script is no more self-contained. For that reason, I can't run the script from the GH raw URL, which was really convenient.

```
curl -sLko run https://raw.githubusercontent.com/fvaleri/strimzi-kafka-operator/fix-report/tools/report.sh \
  && chmod +x run && ./run --namespace=test --cluster=my-cluster
```

scholzj:
> @scholzj the problem with sourcing multi-platform-support.sh is that the report script is not self-contained as before. For that reason, I can't run it from the GH raw URL, which was really convenient.

I guess that is very unsecure, but fair point, so ok, lets define it directly in the script.

scholzj (approved):

scholzj:
Hey @strimzi/maintainers Can anyone have a look at this?

Frawless (approved):

scholzj:
Thanks for the PR @fvaleri

PR #6238, opened 2022-01-20
Fix #6019 moving log4j to test scope

### LINKED ISSUE #6019: Minimize api dependencies
The api module has very broad compile dependencies, including on log4j and vert.x core.

The log4j dependencies can be test instead.  The vert.x dependency seems to only be for a single annotation - but the dependency has a large transitive footprint.  It doesn't appear that vert.x offers another narrow dependency, maybe they could be persuaded to release one.

Similarly it looks like an upstream issue is warranted against fabric8 to depend only on a client api without pulling in all the transitive dependencies.  I'll make sure something for that is captured for fabric8 6.  The only thing that seems problematic is the direct usage of KubernetesDeserializer.registerCustomKind.

shawkins:
### Type of change

- Refactoring

### Description

Initial change towards #6019 to minimize dependencies exposed by the api module.

### DIFF
--- a/api/pom.xml
+++ b/api/pom.xml
@@ -71,10 +71,12 @@
         <dependency>
             <groupId>org.apache.logging.log4j</groupId>
             <artifactId>log4j-core</artifactId>
+            <scope>test</scope>
         </dependency>
         <dependency>
             <groupId>org.apache.logging.log4j</groupId>
             <artifactId>log4j-slf4j-impl</artifactId>
+            <scope>test</scope>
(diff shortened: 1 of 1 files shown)


scholzj:
Can you fix the DCO please @shawkins?

shawkins:
> Can you fix the DCO please @shawkins?

Yes, no problem.

scholzj (approved):
Thanks for the PR

Frawless (approved):

PR #6239, opened 2022-01-20
Refactor storage handling in model classes to minimize duplicate logic

scholzj:
### Type of change

- Refactoring

### Description

While working on the StatefulSet removal, I noticed that some of the methods for working with storage in `KafkaCluster` and `ZooKeeperCluster` were to some extent duplicated. I moved these classes to a single place (most of it to `VolumeUtils`, with the exception of the `createPersistentVolumeClaims` method which is too much dependent on the `AbstractModel` class). These methods are now shared by both Kafka and ZooKeeper. I also centralized the check for the mandatory IDs for JBOD volumes so that it is done in one place only.

Additionally, I also:
* added better tests to cover all the different situations
* used more consistent method names
* added Javadoc to most of the methdos

This PR should make sure that most of the logic is in one place and has better test coverage.

### Checklist

- [x] Write tests


scholzj:
(the CI failure is not related to this PR)

ppatierno (approved):

PR #6240, opened 2022-01-21
Fix documentation about install and renew your own CA certificates

ppatierno:
### Type of change

- Documentation

### Description

This is the documentation related update for PR #6180 which fixes issues #5466.
The PR adds more information about certification generations to be added to the Secret for handling your own CA certificates and corresponding renewal.

### Checklist

- [x] Update documentation


ppatierno:
This PR can be merged only after #6180

documentation/modules/security/proc-installing-your-own-ca-certificates.adoc
```diff
@@ -11,7 +11,7 @@ This procedure describes how to install your own CA certificates and keys instea
 The Cluster Operator automatically generates and renews the following secrets:
 
 `_CLUSTER-NAME_-cluster-ca`:: The cluster secret that contains the private key for the cluster CA.
-`_CLUSTER-NAME_-cluster-ca-cert`:: The cluster secret that contains a cluster CA certificate. The certificate contains a public key to validate the identity of Kafka brokers.
+`_CLUSTER-NAME_-cluster-ca-cert`:: The cluster secret that contains a cluster CA certificate. The certificate contains a public key to validate the identity of Kafka brokers and ZooKeeper nodes.
```
scholzj, line 14: This is user facing docs => users do not connect to ZooKeeper ... so I think this is confusing. I think it should remain as it was.

documentation/modules/security/proc-renewing-your-own-ca-certificates.adoc
```diff
@@ -54,6 +54,11 @@ For example `ca-2018-09-27T17-32-00Z.crt`.
 +
 [source,shell,subs="+quotes"]
 mv ca.crt ca-$(date -u -d"$(openssl x509 -enddate -noout -in ca.crt | sed 's/.*=//')" +'%Y-%m-%dT%H-%M-%SZ').crt
++
+Saving the current CA certificate has to be done only if you are using a different key for generating the new certificate.
+In this case the CN (Common Name) of the new CA certificate should be different from the previous one.
+For example, when Strimzi cluster operator takes care of the renewal automatically it adds a suffix like "vX" to identify a version "X" (that is, v0, v1, and so on).
+If you are using the same key for generating the new CA certificate saving the current one is not necessary because only one CA certificate at time can be trusted.
```
scholzj, line 61: IIRC, you said that saving the old here caused failures in the scripts. So maybe this should use stronger wording?

documentation/modules/security/proc-renewing-your-own-ca-certificates.adoc
```diff
@@ -105,3 +110,16 @@ kubectl create secret generic _CA-KEY-SECRET_ --from-file=ca.key=_CA-KEY-SECRET-
...
++
+[source,shell,subs="+quotes"]
+----
+kubectl annotate secret _CA-CERTIFICATE-SECRET_ strimzi.io/ca-cert-generation=_CA-CERTIFICATE-GENERATION_
+kubectl annotate secret _CA-KEY-SECRET_ strimzi.io/ca-key-generation=_CA-KEY-GENERATION_
+----
++
+When the Strimzi generated CA certificates are used, the generation increment is automatically handled by the cluster operator.
+In this case, with manual renewal of your own CA certificates, you have to take care to increment the generation on renewal accordingly.
+The generation annotations have to have the next incremental value of the current ones from the Secret.
```
scholzj, line 124: Should it be the next value or just higher value? I wonder if it could cause some confusion when you for example due to some errors try to do another renewal before the first one finished etc.

documentation/modules/security/proc-renewing-your-own-ca-certificates.adoc
ppatierno: Yeah openssl is not able to deal with two CA having same CN. I will be clearer about this.

documentation/modules/security/proc-renewing-your-own-ca-certificates.adoc
ppatierno: Good point. It has to be an higher value, no need to be exactly the next one.

documentation/modules/security/proc-renewing-your-own-ca-certificates.adoc
```diff
@@ -105,3 +110,16 @@ kubectl create secret generic _CA-KEY-SECRET_ --from-file=ca.key=_CA-KEY-SECRET-
...
+[source,shell,subs="+quotes"]
+----
+kubectl annotate secret _CA-CERTIFICATE-SECRET_ strimzi.io/ca-cert-generation=_CA-CERTIFICATE-GENERATION_
+kubectl annotate secret _CA-KEY-SECRET_ strimzi.io/ca-key-generation=_CA-KEY-GENERATION_
+----
++
+When the Strimzi generated CA certificates are used, the generation increment is automatically handled by the cluster operator.
+In this case, with manual renewal of your own CA certificates, you have to take care to increment the generation on renewal accordingly.
+The generation annotations have to have a values higher than the current ones from the Secret.
+This is needed by the Strimzi cluster operator to roll the pods for updating the certificates.
```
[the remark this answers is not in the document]
PaulRMellor: Maybe an example can help here.

PaulRMellor (commented):
Looking good. I wasn't entirely clear on where we're suggesting to add an incremental value in the doc. Thought an example might help

documentation/modules/security/proc-renewing-your-own-ca-certificates.adoc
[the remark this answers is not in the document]
ppatierno: @PaulRMellor I don't understand why we should say "add" ("add an incremental value to the annotation.") we are not adding ... we setting the annotation with an incremental value (`strimzi.io/ca-cert-generation=1`, then with a new renewal `strimzi.io/ca-cert-generation=2` and so on).

documentation/modules/security/proc-installing-your-own-ca-certificates.adoc
```diff
@@ -122,6 +122,24 @@ kubectl label secret _CA-KEY-SECRET_ strimzi.io/kind=Kafka strimzi.io/cluster=_C
...
++
+[source,shell,subs="+quotes"]
+----
+kubectl annotate secret _CA-KEY-SECRET_ strimzi.io/ca-key-generation=_CA-KEY-GENERATION_
+----
++
+* Annotation `strimzi.io/ca-cert-generation=_CA-CERTIFICATE-GENERATION_` defines the new generation of the CA certificate.
+* Annotation `strimzi.io/ca-key-generation=_CA-KEY-GENERATION_` defines the new generation of the CA key.
++
+If you are replacing CA certificates automatically generated by the Strimzi cluster operator then you should use the next incremental value as new generation from the corresponding existing annotation.
```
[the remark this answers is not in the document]
ppatierno: @PaulRMellor didn't get what you mean sorry ... which procedure are you referring to?

documentation/modules/security/proc-renewing-your-own-ca-certificates.adoc
```diff
@@ -105,3 +110,16 @@ kubectl create secret generic _CA-KEY-SECRET_ --from-file=ca.key=_CA-KEY-SECRET-
...
+
+. Annotate the secrets with the generation annotations:
++
+[source,shell,subs="+quotes"]
+----
+kubectl annotate secret _CA-CERTIFICATE-SECRET_ strimzi.io/ca-cert-generation=_CA-CERTIFICATE-GENERATION_
+kubectl annotate secret _CA-KEY-SECRET_ strimzi.io/ca-key-generation=_CA-KEY-GENERATION_
+----
++
+When the Strimzi generated CA certificates are used, the generation increment is automatically handled by the cluster operator.
```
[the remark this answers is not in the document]
PaulRMellor: Got it. I've updated the suggestion

documentation/modules/security/proc-renewing-your-own-ca-certificates.adoc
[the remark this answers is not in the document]
PaulRMellor: Right. Updated.

documentation/modules/security/proc-installing-your-own-ca-certificates.adoc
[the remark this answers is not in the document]
ppatierno: Isn't the user confused that he should set always generation 1? I mean if the current has generation 0 then 1 is ok ... but even 2, 3, 10, 50, ... any number greater than the current.

documentation/modules/security/proc-renewing-your-own-ca-certificates.adoc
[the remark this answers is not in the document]
ppatierno: Do you mean  ...

For example, change `strimzi.io/ca-cert-generation=1` to `strimzi.io/ca-cert-generation=2`

Right?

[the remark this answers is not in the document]
PaulRMellor: Yes! Updated

documentation/modules/security/proc-installing-your-own-ca-certificates.adoc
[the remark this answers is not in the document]
PaulRMellor: I added the example we use in the other step.

documentation/modules/security/proc-installing-your-own-ca-certificates.adoc
```diff
@@ -122,6 +122,24 @@ kubectl label secret _CA-KEY-SECRET_ strimzi.io/kind=Kafka strimzi.io/cluster=_C
...
+[source,shell,subs="+quotes"]
+----
+kubectl annotate secret _CA-KEY-SECRET_ strimzi.io/ca-key-generation=_CA-KEY-GENERATION_
+----
++
+* Annotation `strimzi.io/ca-cert-generation=_CA-CERTIFICATE-GENERATION_` defines the generation of a new CA certificate.
+* Annotation `strimzi.io/ca-key-generation=_CA-KEY-GENERATION_` defines the generation of a new CA key.
++
+If you are replacing CA certificates automatically generated by the Strimzi cluster operator then you should use the next incremental value as new generation from the corresponding existing annotation.
+If you are generating the CA certificates for the first time, you can start from 0 (zero) as the incremental value.
```
PaulRMellor, line 141:
```suggestion
If you are generating the CA certificates for the first time, you can start from 0 (zero) as the incremental value (`strimzi.io/ca-cert-generation=0`). This value must be incremented when you renew the certificates.

**Review by PaulRMellor — APPROVED** [review]
Nice :+1:

File: documentation/modules/security/proc-installing-your-own-ca-certificates.adoc [hunk 8]
```diff
@@ -122,6 +122,24 @@ kubectl label secret _CA-KEY-SECRET_ strimzi.io/kind=Kafka strimzi.io/cluster=_C
...
++
+[source,shell,subs="+quotes"]
+----
+kubectl annotate secret _CA-KEY-SECRET_ strimzi.io/ca-key-generation=_CA-KEY-GENERATION_
+----
++
+* Annotation `strimzi.io/ca-cert-generation=_CA-CERTIFICATE-GENERATION_` defines the generation of a new CA certificate.
+* Annotation `strimzi.io/ca-key-generation=_CA-KEY-GENERATION_` defines the generation of a new CA key.
++
+If you are replacing CA certificates automatically generated by the Cluster Operator, use the next higher incremental value from the existing annotation.
```
scholzj, line 140: I'm confused. Does this refer to migration from Strimzi generated CA to custom CA?

documentation/modules/security/proc-renewing-your-own-ca-certificates.adoc
```diff
@@ -54,6 +54,13 @@ For example `ca-2018-09-27T17-32-00Z.crt`.
 +
 [source,shell,subs="+quotes"]
 mv ca.crt ca-$(date -u -d"$(openssl x509 -enddate -noout -in ca.crt | sed 's/.*=//')" +'%Y-%m-%dT%H-%M-%SZ').crt
++
+If you are using a different key to generate a new certificate, retain the current CA certificate.
```
scholzj, line 58: I think this deserves some more explanation => might be useful to explain why but also what does retaining men.

documentation/modules/security/proc-renewing-your-own-ca-certificates.adoc
```diff
@@ -54,6 +54,13 @@ For example `ca-2018-09-27T17-32-00Z.crt`.
...
+The CN (Common Name) of the new CA certificate must also be different from the current one.
+For example, when the Cluster Operator renews certificates automatically it adds a _v<version_number>_ suffix to identify a version.
+Do the same with your own certificates.
+If you are using the same key for generate a new CA certificate, don't retain the old one in the secret.
```
scholzj, line 62: I think this deserves some more explanation => might be useful to explain why but also what does retaining men.

documentation/modules/security/proc-renewing-your-own-ca-certificates.adoc
```diff
@@ -54,6 +54,13 @@ For example `ca-2018-09-27T17-32-00Z.crt`.
 +
 [source,shell,subs="+quotes"]
 mv ca.crt ca-$(date -u -d"$(openssl x509 -enddate -noout -in ca.crt | sed 's/.*=//')" +'%Y-%m-%dT%H-%M-%SZ').crt
++
+If you are using a different key to generate a new certificate, retain the current CA certificate.
+The CN (Common Name) of the new CA certificate must also be different from the current one.
+For example, when the Cluster Operator renews certificates automatically it adds a _v<version_number>_ suffix to identify a version.
+Do the same with your own certificates.
```
scholzj, line 61: What certificates? I think this needs to be more specific or left out.

documentation/modules/security/proc-renewing-your-own-ca-certificates.adoc
```diff
@@ -105,3 +112,15 @@ kubectl create secret generic _CA-KEY-SECRET_ --from-file=ca.key=_CA-KEY-SECRET-
...
+. Annotate the secrets with the generation annotations:
++
+[source,shell,subs="+quotes"]
+----
+kubectl annotate secret _CA-CERTIFICATE-SECRET_ strimzi.io/ca-cert-generation=_CA-CERTIFICATE-GENERATION_
+kubectl annotate secret _CA-KEY-SECRET_ strimzi.io/ca-key-generation=_CA-KEY-GENERATION_
+----
++
+When Strimzi generates certificates, the certificate generation annotation is automatically incremented by the Cluster Operator.
+For manual renewal of your own CA certificates, set the annotations with a higher incremental value. For example, change `strimzi.io/ca-cert-generation=1` to `strimzi.io/ca-cert-generation=2`.
```
scholzj, line 125: So, which annotation should be changed when? I mean with regards to key-renewal & cett renewal?

scholzj (commented):
TBH, I'm not sure I understand what exactly should be done for the renewal. It should probably have more steps and more examples.

documentation/modules/security/proc-installing-your-own-ca-certificates.adoc
ppatierno: Yes. I thought it was important to highlight this if user starts with Strimzi CAs and then wants to move to his own CA certs. For triggering the update, increasing the generation is needed.

ppatierno:
> TBH, I'm not sure I understand what exactly should be done for the renewal. It should probably have more steps and more examples.

I just updated the procedure with the additional stuff needed. I think the current way works why re-writing it again if it was approved in the past?

documentation/modules/security/proc-installing-your-own-ca-certificates.adoc
scholzj, line 140: Sorry to be picky. I think the idea to cover the migration from Strimzi generated to custom CA is good and useful. But I think it needs to point to the renewal guide and not suggest to follow the _installing your own CA_ guide. This guide goes through the creation of new secret before deploying cluster. You cannot do that for existing cluster with Strimzi's CA. You would also need to follow the _new key_ renewal pattern to ensure continuity. Just creating a brand new secret with new certs would break the cluster ragardless what generation it uses.

It would also need to include the changes to the Kafka configuration. So maybe it needs separate procedure?

documentation/modules/security/proc-installing-your-own-ca-certificates.adoc
```diff
@@ -122,6 +122,24 @@ kubectl label secret _CA-KEY-SECRET_ strimzi.io/kind=Kafka strimzi.io/cluster=_C
...
+If you are generating the CA certificates for the first time, you can start from 0 (zero) as the incremental value (`strimzi.io/ca-cert-generation=0`). Set a higher incremental value when you renew the certificates.
```
scholzj, line 141:
```suggestion
If you are generating the CA certificates for the first time, you can start from 0 (zero) as the incremental value (for example `strimzi.io/ca-cert-generation=0`).
```

documentation/modules/security/proc-renewing-your-own-ca-certificates.adoc
```diff
@@ -105,3 +112,17 @@ kubectl create secret generic _CA-KEY-SECRET_ --from-file=ca.key=_CA-KEY-SECRET-
 kubectl label secret _CA-CERTIFICATE-SECRET_ strimzi.io/kind=Kafka strimzi.io/cluster=_KAFKA-CLUSTER-NAME_
 kubectl label secret _CA-KEY-SECRET_ strimzi.io/kind=Kafka strimzi.io/cluster=_KAFKA-CLUSTER-NAME_
 ----
+
+. Annotate the secrets with the generation annotations:
++
+[source,shell,subs="+quotes"]
+----
+kubectl annotate secret _CA-CERTIFICATE-SECRET_ strimzi.io/ca-cert-generation=_CA-CERTIFICATE-GENERATION_
+kubectl annotate secret _CA-KEY-SECRET_ strimzi.io/ca-key-generation=_CA-KEY-GENERATION_
```
scholzj, line 121: What happens when reconciliation hits between updating the secret and updating the annotations? Shouldn't this instead of using `kubectl label` and `kubectl annotate` work with editing the original secret?

documentation/modules/security/proc-renewing-your-own-ca-certificates.adoc
```diff
@@ -105,3 +112,17 @@ kubectl create secret generic _CA-KEY-SECRET_ --from-file=ca.key=_CA-KEY-SECRET-
...
+----
++
+When Strimzi generates certificates, the certificate generation annotation is automatically incremented by the Cluster Operator.
+For manual renewal of your own CA certificates, set the annotations with a higher incremental value. For example, change `strimzi.io/ca-cert-generation=1` to `strimzi.io/ca-cert-generation=2`.
+The annotations need a value higher than the ones from the current secret so the Cluster Operator can roll the pods and update the certificates.
+The `strimzi.io/ca-cert-generation` has to be incremented on each CA certificate renewal.
```
scholzj, line 127: So, when renewing key, this does not need to be bumped? Or does it need to be bumped in both cases? I think `each CA certificate renewal` is not really clear.

scholzj (commented):
TBH, I'm not sure I think this is really understandable. What about:
* Splitting the key replacement and CA renewal into two separate procedures?
* Adding more examples which would include a YAMLs of the secrets to show exactly how they should look like? I think that will make it much clearer and you can use callouts to point out what should be where, what needs to be changed etc. Editing the secrets would also help to deal with the delayed annotations etc.

ppatierno:
Ok then. I will re-write this PR by having two procedures:

* Renewing your own CA certificates
* Replacing private keys used by your own CA certificates

It will follow the same pattern we have for the ones related to auto-generated CA by Strimzi.

ppatierno:
@scholzj @PaulRMellor I made a big change on this PR with two different procedures (if key is replaced or not).

documentation/modules/security/proc-installing-your-own-ca-certificates.adoc
scholzj: I think this is still valid comment.

documentation/modules/security/proc-installing-your-own-ca-certificates.adoc
scholzj: This as well.

documentation/modules/security/proc-renewing-your-own-ca-certificates.adoc
```diff
@@ -5,7 +5,8 @@
 [id='renewing-your-own-ca-certificates-{context}']
 = Renewing your own CA certificates
 
-This procedure describes how to renew CA certificates and keys you installed yourself, instead of using the certificates generated by the Cluster Operator.
+This procedure describes how to renew CA certificates you installed yourself, instead of using the certificates generated by the Cluster Operator.
+Follow this procedure when you are not changing the private keys used to sign the CA certificates, but your new CA certificates are signed with same keys.
```
scholzj, line 9: The CA does not have to be always self-signed. It can be just an intermediate CA. So, does this apply when:
* the private key which signed the CA is changed?

or

* When the CA private key is changed?

I assume the second is valid for sure, but never tried the first. Either way should probably make it more clear.

documentation/modules/security/proc-renewing-your-own-ca-certificates.adoc
```diff
@@ -16,92 +17,68 @@ The procedure describes the renewal of CA certificates in PEM format.
 
 * The Cluster Operator is running.
 * xref:installing-your-own-ca-certificates-{context}[Your own CA certificates and private keys are installed].
-* You have new cluster and clients X.509 certificates and keys in PEM format.
-
-These could be generated using an `openssl` command, such as:
-
-[source,shell,subs="+quotes"]
-openssl req -x509 -new -days _NUMBER-OF-DAYS-VALID_ --nodes -out ca.crt -keyout ca.key
+* You have new cluster and clients X.509 certificates in PEM format.
```
scholzj, line 20: Should this be `or`? I assume you can replace only one of them since they are completely independent?

documentation/modules/security/proc-renewing-your-own-ca-certificates.adoc
```diff
@@ -16,92 +17,68 @@ The procedure describes the renewal of CA certificates in PEM format.
...
+    strimzi.io/ca-cert-generation: "0" <2>
+  labels:
+    strimzi.io/cluster: my-cluster
+    strimzi.io/kind: Kafka
+  name: my-cluster-cluster-ca-cert
+  #...
+type: Opaque
+----
+<1> Current CA certificate base64 encoded
+<2> Current CA certificate generation
```
scholzj, line 56: Seems a bit weord to have the before example and no after example => should we have both before and after (shows the bumped generation and new CRT file)?

documentation/modules/security/proc-renewing-your-own-ca-certificates.adoc
```diff
@@ -16,92 +17,68 @@ The procedure describes the renewal of CA certificates in PEM format.
...
-.. Delete the existing secret:
-+
-[source,shell,subs="+quotes"]
-kubectl delete secret _CA-KEY-SECRET_
-+
-_CA-KEY-SECRET_ is the name of CA key, which is `_KAFKA-CLUSTER-NAME_-cluster-ca` for the cluster CA key and `_KAFKA-CLUSTER-NAME_-clients-ca` for the clients CA key.
+When Strimzi generates certificates, the certificate generation annotation is automatically incremented by the Cluster Operator.
+For manual renewal of your own CA certificates, set the annotations with a higher incremental value.
+The annotations need a value higher than the ones from the current Secret so the Cluster Operator can roll the pods and update the certificates.
+The `strimzi.io/ca-cert-generation` has to be incremented on each CA certificate renewal.
```
scholzj, line 80: Is this still needed? wasn't that covered above?

documentation/modules/security/proc-replacing-your-own-private-keys.adoc
```diff
@@ -0,0 +1,136 @@
+// Module included in the following assemblies:
+//
+// assembly-security.adoc
+
+[id='proc-replacing-your-own-private-keys-{context}']
+= Replacing private keys used by your own CA certificates
+
+This procedure describes how to renew CA certificates and keys you installed yourself, instead of using the certificates generated by the Cluster Operator.
+Follow this procedure when you are changing the private keys used to sign the new CA certificates.
```
scholzj, line 9: Same as above, is it changing the private keys used to sign the CA? Or the CA private key? Not the same thing.

documentation/modules/security/proc-replacing-your-own-private-keys.adoc
```diff
@@ -0,0 +1,136 @@
...
+    strimzi.io/ca-cert-generation: "0" <2>
+  labels:
+    strimzi.io/cluster: my-cluster
+    strimzi.io/kind: Kafka
+  name: my-cluster-cluster-ca-cert
+  #...
+type: Opaque
+----
+<1> Current CA certificate base64 encoded
+<2> Current CA certificate generation
```
scholzj, line 57: Sema as in the other procedure => showing both and after secret would be useful. Especially since the modification here goes beyond just changing the generation.

documentation/modules/security/proc-replacing-your-own-private-keys.adoc
```diff
@@ -0,0 +1,136 @@
...
++
+Copy the new CA key base64 encoded got from the first step into the `ca.key` field of the `data` section in the `Secret`.
++
+.. Increase the CA key generation.
++
+Updated the `strimzi.io/ca-key-generation` annotation with a higher incremental value.
+For example, change `strimzi.io/ca-key-generation=0` to `strimzi.io/ca-key-generation=1`.
++
+
+. Save changes and exit.
```
scholzj, line 129: I guess this can still screw up the cluster if only one secret is updated when a recocniliation starts? This process is f*cked up :-/. Should we tell the users to pause the reconciliation or something?

documentation/modules/security/proc-renewing-your-own-ca-certificates.adoc
ppatierno: Ok. I will add the after example in the section related to "Save changes and exit".

documentation/modules/security/proc-renewing-your-own-ca-certificates.adoc
ppatierno: This was just a kind of summary about generation and how it differs from the auto-renewal. If you think it sounds like a step to do and could be misleading, I can remove it. Wdyt?

documentation/modules/security/proc-replacing-your-own-private-keys.adoc
ppatierno: Good point. Yeah it could happen that CA cert is updated but not corresponding key. During rolls it will screw up stuff. 
I think that pausing and resuming is the only good solution we have. I will add a step at the beginning and the end of the procedure.

documentation/modules/security/proc-renewing-your-own-ca-certificates.adoc
ppatierno: Ok this sentence is worded in the wrong way. The CA key is used to sign the server certificates we generate, while I was wrongly mentioning key used to sign the CA which is absolutely a different thing.

documentation/modules/security/proc-replacing-your-own-private-keys.adoc
ppatierno: I just removed the sentence in this case, because the previous one and even the title should be clear about you using a new CA key.

scholzj (commented):
nice touch with the pausing or the reconciliation.

documentation/modules/security/proc-renewing-your-own-ca-certificates.adoc
```diff
@@ -16,92 +17,85 @@ The procedure describes the renewal of CA certificates in PEM format.
...
+Save the changes to the `Secret` and exit.
 
-.. Recreate the secret with the new CA key:
-+
-[source,shell,subs="+quotes"]
-kubectl create secret generic _CA-KEY-SECRET_ --from-file=ca.key=_CA-KEY-SECRET-FILENAME_
+When Strimzi generates certificates, the certificate generation annotation is automatically incremented by the Cluster Operator.
+For manual renewal of your own CA certificates, set the annotations with a higher incremental value.
+The annotations need a value higher than the ones from the current Secret so the Cluster Operator can roll the pods and update the certificates.
+The `strimzi.io/ca-cert-generation` has to be incremented on each CA certificate renewal.
```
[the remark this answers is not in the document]
ppatierno: Agree.

documentation/modules/security/proc-replacing-your-own-private-keys.adoc
```diff
@@ -0,0 +1,223 @@
...
++
+[source,shell,subs="+quotes"]
+----
+kubectl annotate Kafka _NAME-OF-CUSTOM-RESOURCE_ strimzi.io/pause-reconciliation-
+----
+
+When Strimzi generates certificates, the certificate and key generation annotations are automatically incremented by the Cluster Operator.
+For manual renewal of your own CA certificates together with a new CA key, set the annotations with a higher incremental value.
+The annotations need a value higher than the ones from the current Secret so the Cluster Operator can roll the pods and update the certificates and keys.
+The `strimzi.io/ca-cert-generation` and `strimzi.io/ca-key-generation` have to be incremented on each CA certificate renewal.
```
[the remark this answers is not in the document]
ppatierno: Agree.

documentation/modules/security/proc-replacing-your-own-private-keys.adoc
```diff
@@ -0,0 +1,223 @@
...
+.. Save changes and exit.
++
+Following the previous example of the `Secret` related to the cluster CA certificate for a Kafka cluster named `my-cluster` with the new CA certificate, and the bumped generation.
++
+[source,yaml,subs=attributes+]
+----
+apiVersion: v1
+kind: Secret
+data:
+  ca.crt: GCa6LS3RTHeKFiFDGBOUDYFAZ0F... <1>
```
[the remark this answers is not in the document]
ppatierno: Right!

scholzj (approved):

PaulRMellor (approved):
I've reviewed and pushed up some edits

tombentley (approved):

documentation/modules/security/proc-renewing-your-own-ca-certificates.adoc
```diff
@@ -17,21 +20,22 @@ The procedure describes the renewal of CA certificates in PEM format.
 
 * The Cluster Operator is running.
 * xref:installing-your-own-ca-certificates-{context}[Your own CA certificates and private keys are installed].
-* You have new cluster or clients X.509 certificates in PEM format.
+* You have new cluster or client X.509 certificates in PEM format.
```
ppatierno, line 23: It's actually "clients" ... as we call it (it's the CA certificate to sign clients related certs)

documentation/modules/security/proc-renewing-your-own-ca-certificates.adoc
```diff
@@ -48,34 +52,32 @@ metadata:
...
+Copy the base64-encoded CA certificate from the previous step as the value for the `ca.crt` property under `data`.
 +
-.. Increase the CA certificate generation.
+. Increase the value of the CA certificate generation annotation.
 +
 Update the `strimzi.io/ca-cert-generation` annotation with a higher incremental value.
 For example, change `strimzi.io/ca-cert-generation=0` to `strimzi.io/ca-cert-generation=1`.
-If the `Secret` is missing the annotation, it is treated as generation `0` so add the annotation with `1` as value.
-
+If the `Secret` is missing the annotation, the value is treated as `0`, so change the value to `1`.
```
ppatierno, line 71: why "change"? if it's missing you have to add the annotation, you cannot change it.

documentation/modules/security/proc-replacing-your-own-private-keys.adoc
```diff
@@ -5,36 +5,40 @@
...
+For example, when the Cluster Operator renews certificates automatically it adds a _v<version_number>_ suffix to identify a version.
+Do the same with your own CA certificate by adding a different suffix on each renewal.
+By using a different key to generate a new CA certificate, you retain the current CA certificate stored in the `Secret`.
+
 .Prerequisites
 
 * The Cluster Operator is running.
 * xref:installing-your-own-ca-certificates-{context}[Your own CA certificates and private keys are installed].
-* You have new cluster or clients X.509 certificates and keys in PEM format.
+* You have new cluster or client X.509 certificates and keys in PEM format.
```
ppatierno, line 28: ditto as above, it should be "clients"

documentation/modules/security/proc-replacing-your-own-private-keys.adoc
```diff
@@ -77,40 +81,38 @@ metadata:
...
 +
-.. Increase the CA certificate generation.
+.. Increase the value of the CA certificate generation annotation.
 +
-Updated the `strimzi.io/ca-cert-generation` annotation with a higher incremental value.
+Update the `strimzi.io/ca-cert-generation` annotation with a higher incremental value.
 For example, change `strimzi.io/ca-cert-generation=0` to `strimzi.io/ca-cert-generation=1`.
-If the `Secret` is missing the annotation, it is treated as generation `0` so add the annotation with `1` as value.
-
+If the `Secret` is missing the annotation, the value is treated as `0`, so change the value to `1`.
```
ppatierno, line 106: ditto as above, it's missing so you have to add, can't change.

documentation/modules/security/proc-replacing-your-own-private-keys.adoc
```diff
@@ -128,23 +130,22 @@ metadata:
...
-Save the changes to the `Secret` and exit.
+<1> New base64-encoded CA certificate
+<2> Old base64-encoded CA certificate
+<3> New CA certificate generation annotation value
 
-. Update the corresponding `Secret` with the new CA key and generation.
+. Update the `Secret` for the CA key used to sign your new CA certificate.
 +
-.. Edit the existing `Secret`:
+.. Edit the existing secret to add the new CA key and update the certificate generation annotation value.
```
ppatierno, line 139: it's the key generation annotation. It's about the key not the certificate.

documentation/modules/security/proc-replacing-your-own-private-keys.adoc
```diff
@@ -161,34 +162,32 @@ metadata:
...
 +
-.. Increase the CA key generation.
+.. Increase the value of the CA key generation annotation.
 +
-Updated the `strimzi.io/ca-key-generation` annotation with a higher incremental value.
+Update the `strimzi.io/ca-key-generation` annotation with a higher incremental value.
 For example, change `strimzi.io/ca-key-generation=0` to `strimzi.io/ca-key-generation=1`.
-If the `Secret` is missing the annotation, it is treated as generation `0` so add the annotation with `1` as value.
-
+If the `Secret` is missing the annotation, it is treated as `0`, so change the value to `1`.
```
ppatierno, line 181: same as before, more "add" than "change".

documentation/modules/security/proc-replacing-your-own-private-keys.adoc
```diff
@@ -161,34 +162,32 @@ metadata:
...
 For manual renewal of your own CA certificates together with a new CA key, set the annotation with a higher incremental value.
-The annotation needs a value higher than the ones from the current Secret so the Cluster Operator can roll the pods and update the certificates and keys.
+The annotation needs a higher value than the one from the current secret so that the Cluster Operator can roll the pods and update the certificates and keys.
 The `strimzi.io/ca-key-generation` has to be incremented on each CA certificate renewal.
-+
 
-. Save changes and exit.
-+
-Following the previous example of the `Secret` related to the cluster CA key for a Kafka cluster named `my-cluster` with the new CA key, and the bumped generation.
+. Save the secret with the new CA key and certificate generation annotation value.
```
ppatierno, line 188: it's the key generation annotation

PR #6241, opened 2022-01-21
[systemtest] Check real logs for pattern in logging tests

im-konge:

### Type of change

- Enhancement

### Description

Currently we are checking in our logging tests (`LoggingChangeST`) that pod contains `log4j` properties file, which is changing while we changing CMs of each components. It can be problem when the `log4j` is not used in images (by mistake) - the file itself is changed, but the logging not. This PR adds checks, that log from certain pod contains `log4j` pattern.

Also fixes issue with JSON format logging test - even when we didn't get a JSON object from log we returned `true`, which lead to wrong assertions in our checks. We had also some edge cases, where regex wasn't matching any records. I'm adding parser to `StUtils`, which processes block of text char by char, creating `JsonObject`s and then `JsonArray`.

### Checklist

### DIFF
--- a/systemtest/src/main/java/io/strimzi/systemtest/utils/StUtils.java
+++ b/systemtest/src/main/java/io/strimzi/systemtest/utils/StUtils.java
@@ -19,6 +19,7 @@ import io.strimzi.systemtest.Constants;
 import io.strimzi.systemtest.Environment;
 import io.strimzi.test.TestUtils;
 import io.strimzi.test.k8s.KubeClusterResource;
+import io.vertx.core.json.DecodeException;
 import io.vertx.core.json.JsonArray;
 import io.vertx.core.json.JsonObject;
 import org.apache.logging.log4j.LogManager;
@@ -56,8 +57,6 @@ public class StUtils {
 
     private static final Pattern VERSION_IMAGE_PATTERN = Pattern.compile("(?<version>[0-9.]+)=(?<image>[^\\s]*)");
 
-    private static final Pattern BETWEEN_JSON_OBJECTS_PATTERN = Pattern.compile("}[\\n\\r]+\\{");
-    private static final Pattern ALL_BEFORE_JSON_PATTERN = Pattern.compile("(.*\\s)}, \\{", Pattern.DOTALL);
     private static final BiFunction<String, ExtensionContext, Boolean> CONTAINS_ANNOTATION =
         (annotationName, extensionContext) -> Arrays.stream(extensionContext.getElement().get().getAnnotations()).filter(
             annotation -> annotation.annotationType().getName()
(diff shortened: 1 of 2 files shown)


systemtest/src/main/java/io/strimzi/systemtest/utils/StUtils.java
```diff
@@ -434,4 +411,49 @@ public static void copyImagePullSecret(String namespace) {
                 .withData(Collections.singletonMap(".dockerconfigjson", pullSecret.getData().get(".dockerconfigjson")))
                 .build());
     }
+
+    private static JsonArray getJsonArrayFromLog(String log) {
```
see-quick, line 415: JFYI: If you are making the parser the great way how to make it quickly only by defining excepted structure and 
 then just generate it by `ANTLR`...

see-quick, line 415: Would be great to put here some JavaDoc what is the excepted structure of the input `log` :)

systemtest/src/test/java/io/strimzi/systemtest/log/LoggingChangeST.java
```diff
@@ -140,6 +143,9 @@ void testJSONFormatLogging(ExtensionContext extensionContext) {
         String configMapKafkaName = "json-layout-kafka";
         String configMapCOName = Constants.STRIMZI_DEPLOYMENT_NAME;
 
+        String originalCoLoggers = kubeClient().getClient().configMaps()
+            .inNamespace(INFRA_NAMESPACE).withName(configMapCOName).get().getData().get("log4j2.properties");
```
see-quick, line 147:
```suggestion
.inNamespace(clusterOperator.getDeploymentNamespace()).withName(configMapCOName).get().getData().get("log4j2.properties");
```
This needs to change because when you try OLM here (it won't work)

systemtest/src/test/java/io/strimzi/systemtest/log/LoggingChangeST.java
```diff
@@ -240,12 +246,16 @@ void testJSONFormatLogging(ExtensionContext extensionContext) {
         StUtils.checkLogForJSONFormat(namespaceName, zkPods, "zookeeper");
         StUtils.checkLogForJSONFormat(namespaceName, eoPods, "topic-operator");
         StUtils.checkLogForJSONFormat(namespaceName, eoPods, "user-operator");
+
+        // set loggers of CO back to original
+        configMapCO.getData().put("log4j2.properties", originalCoLoggers);
+        kubeClient().getClient().configMaps().inNamespace(INFRA_NAMESPACE).createOrReplace(configMapCO);
```
see-quick, line 252: same here

see-quick (commented):
Looks good ;) 👌 just few comments  🦥

systemtest/src/test/java/io/strimzi/systemtest/log/LoggingChangeST.java
im-konge: Completely forgot that! Good catch 🪲

systemtest/src/test/java/io/strimzi/systemtest/log/LoggingChangeST.java
im-konge: 🪲🤲

systemtest/src/main/java/io/strimzi/systemtest/utils/StUtils.java
im-konge: Good to know 🤔 thanks!

systemtest/src/test/java/io/strimzi/systemtest/log/LoggingChangeST.java
im-konge: I also changed it everywhere else in test

im-konge:
@strimzi-ci run tests test_only profile=regression testcase=LoggingChangeST

see-quick (approved):

Frawless (approved):

im-konge:
Failures in AZP are not related to this PR

PR #6244, opened 2022-01-22
Fix recovery of pending pods when PodSets are used

### LINKED ISSUE #6203: StrimziPodSets should be able to handle recovery from incorrectly configured resources
It can happen that the user deploying the Kafka cluster might incorrectly configure the resources for the cluster. This might result in the pods being unschedulable / pending. After the user updates the resources to the correect values, the operator needs to update the pods with the new resources to make them schedulable.

Currently, when using StrimziPodSets, we do not seem to handle this. This needs to be fixed and the test `RecoveryIsolatedST.testRecoveryFromImpossibleMemoryRequest` should be updated to run even with Strimzi Pod Sets enabled.

scholzj:
### Type of change

- Bugfix

### Description

When PodSets are being used and the Kafka pods get into the `Pending` state, the operator currently cannot recover them from such state. This PR fixes it and re-enables the system tests covering this scenario.

The fix consists of fixing the condition of the Kafka Roller. But also handling better the rolling update to add or remove volumes, which right now was being always running through all brokers and was interfering. This PR conditions this rolling update and runs it only when any storage changes actually happen.

This should resolve #6203

### Checklist

- [x] Write tests

### DIFF
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/resource/KafkaRoller.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/resource/KafkaRoller.java
@@ -455,7 +455,9 @@ public class KafkaRoller {
                 "PodScheduled".equals(ps.getType())
                         && "Unschedulable".equals(ps.getReason())
                         && "False".equals(ps.getStatus()));
-        if (podStuck && !reasonToRestartPod.contains("Pod has old generation")) {
+        if (podStuck
+                && !reasonToRestartPod.contains("Pod has old generation")   // "Pod has old generation" is used with StatefulSets
+                && !reasonToRestartPod.contains("Pod has old revision")) {  // "Pod has old revision" is used with PodSets
             // If the pod is unschedulable then deleting it, or trying to open an Admin client to it will make no difference
             // Treat this as fatal because if it's not possible to schedule one pod then it's likely that proceeding
             // and deleting a different pod in the meantime will likely result in another unschedulable pod.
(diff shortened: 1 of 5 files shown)


cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/resource/KafkaRoller.java
```diff
@@ -455,7 +455,9 @@ private RestartPlan restartPlan(int podId, Pod pod, RestartContext restartContex
                 "PodScheduled".equals(ps.getType())
                         && "Unschedulable".equals(ps.getReason())
                         && "False".equals(ps.getStatus()));
-        if (podStuck && !reasonToRestartPod.contains("Pod has old generation")) {
+        if (podStuck
+                && !reasonToRestartPod.contains("Pod has old generation")   // "Pod has old generation" is used with StatefulSets
+                && !reasonToRestartPod.contains("Pod has old revision")) {  // "Pod has old revision" is used with PodSets
```
ppatierno, line 460: If I got the PR right. This is the only change needed for fixing #6203, right? All the other changes are for improving how handling changes on volumes for rolling pods.

scholzj: No, that is not the case. Both changes are needed. The rolling update for adding / removing volumes is before the regular rolling update. When it runs, it fails on the pending pods since it doesn't roll the pod to fix resource configuration, just to add or remove storage. So adding the check if this it really needed, helps to get to the regular rolling update which fixes the pending pods. So they are both part of the fix.

ppatierno: Ok, thanks.

ppatierno (approved):

PR #6245, opened 2022-01-22
[MO] - @IsolatedSuite creation of co

see-quick:

### Type of change

- Bugfix

### Description

This PR fix problem where one run only one specific `@IsolatedSuite` by which shared Cluster Operator is not deployed and thus there are expectations to run co with a different configuration.

### Checklist

### DIFF
--- a/systemtest/src/test/java/io/strimzi/systemtest/security/oauth/OauthAuthorizationIsolatedST.java
+++ b/systemtest/src/test/java/io/strimzi/systemtest/security/oauth/OauthAuthorizationIsolatedST.java
@@ -652,9 +652,6 @@ public class OauthAuthorizationIsolatedST extends OauthAbstractST {
 
     @BeforeAll
     void setUp(ExtensionContext extensionContext)  {
-        clusterOperator.unInstall();
-        clusterOperator.defaultInstallation().createInstallation().runInstallation();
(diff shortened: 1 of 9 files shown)


systemtest/src/test/java/io/strimzi/systemtest/security/oauth/OauthAuthorizationIsolatedST.java
```diff
@@ -652,6 +652,8 @@ void testKeycloakAuthorizerToDelegateToSimpleAuthorizer(ExtensionContext extensi
 
     @BeforeAll
     void setUp(ExtensionContext extensionContext)  {
+        clusterOperator.unInstall();
+        clusterOperator.defaultInstallation().createInstallation().runInstallation();
```
im-konge, line 656: Why don't you add it to `setupCoAndKeycloak` method? Currently it's only deploying Keycloak. (so it has wrong name or CO installation should be inside it)

see-quick: Yes, make sense 👌

im-konge (approved):

Frawless (approved):

PR #6262, opened 2022-01-26
docs: adds oauth plain authentication flows

PaulRMellor:

**Documentation**

Adds a description of the OAuth Plain authentication flows to the _Configuring_ guide, with diagrams to describe the flows.

### Checklist

- [ ] Write tests
- [x] Update documentation
- [ ] Update CHANGELOG.md


scholzj (approved):
One nit. Should be reviewed by Marko as an OAuth expert.

documentation/modules/oauth/con-oauth-authentication-flow.adoc
```diff
@@ -13,12 +14,25 @@ We recommend that you configure clients to use OAUTHBEARER whenever possible.
...
+[source,yaml,subs="attributes+"]
+----
+  # ...
+  authentication:
+    type: oauth
+    # ...
+    enableOauthBearer: true
+----
+
+The OAUTHBEARER authentication mechanism must be explicitly configured.
```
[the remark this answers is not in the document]
PaulRMellor: Bit vague. I think we're talking about the actual _config_ rather than enabling. It applies to both mechanisms anyway, so I've moved up into the intro. I'll see what @mstruk thinks.

> You configure Kafka broker listeners to use OAuth 2.0 authentication for connecting clients.
> If necessary, you can use the OAUTHBEARER and PLAIN mechanisms on the same `oauth` listener.
> The properties to support each mechanism must be explicitly specified in the `oauth` listener configuration.

documentation/modules/oauth/con-oauth-authentication-client-options.adoc
```diff
@@ -3,33 +3,34 @@
...
-In this section, we explain and visualize the communication flow between Kafka client, Kafka broker, and authorization server during Kafka session initiation.
-The flow depends on the client and server configuration.
+[role="_abstract"]
+OAuth 2.0 authentication flows depend on the underlying Kafka client and Kafka broker configuration.
+The flows must also be supported by the authorization server used.
 
-When a Kafka client sends an access token as credentials to a Kafka broker, the token needs to be validated.
+Kafka broker listener configuration determines how clients authenticate using an access token.
+The client can pass a client ID and secret to request an access token.
+If a listener is configured to use PLAIN authentication, you can use a client username and password to request or pass access tokens.
```
mstruk, line 14: This doesn't sound right. there is no such thing as 'client username'. There is either `clientId` or `username`. The way it sounds here is as if we support password authentication, where a user can configure the client with some user account's username and password. That is not supported even when PLAIN authentication is used. What is supported is to authenticate with clientId and secret or with username and access token and passing these as `username` and `password` parameters of the SASL/PLAIN authentication mechanism.  
```suggestion
If a listener is configured to use PLAIN authentication, PLAIN mechanism can be used on the client to authenticate with `clientId` and `secret`, or `username` and `access token` by passing these as `username` and `password` parameters of the PLAIN mechanism.
```

documentation/modules/oauth/con-oauth-authentication-client-options.adoc
```diff
@@ -3,33 +3,34 @@
...
-Using fast local token validation requires the authorization server to provide a JWKS endpoint with public certificates that are used to validate signatures on the tokens.
-
-Another option is to use an OAuth 2.0 introspection endpoint on the authorization server.
-Each time a new Kafka broker connection is established, the broker passes the access token received from the client to the authorization server, and checks the response to confirm whether or not the token is valid.
-
-Kafka client credentials can also be configured for:
+Kafka client credentials can also be configured for the following types of authentication:
 
 * Direct local access using a previously generated long-lived access token
 * Contact with the authorization server for a new access token to be issued
```
mstruk, line 27: Maybe add here a bit more detail?
```suggestion
* Contact with the authorization server for a new access token to be issued (by using `clientId` and a `secret` or a `refresh token`)
```

documentation/modules/oauth/con-oauth-authentication-client-options.adoc
```diff
@@ -3,33 +3,34 @@
...
 
 NOTE: An authorization server might only allow the use of opaque access tokens, which means that local token validation is not possible.
```
mstruk, line 29: I would move this after line 23.

documentation/modules/oauth/con-oauth-authentication-client-options.adoc
```diff
@@ -3,33 +3,34 @@
...
 
-== Example client authentication flows
+== Example client authentication flows using the SASL OAUTHBEARER mechanism
 
-Here you can see the communication flows, for different configurations of Kafka clients and brokers, during Kafka session authentication.
+You can use the following communication flows for Kafka authentication using the OAuth SASL OAUTHBEARER mechanism.
```
mstruk, line 33:
```suggestion
You can use the following communication flows for Kafka authentication using the SASL OAUTHBEARER mechanism.
```

documentation/modules/oauth/con-oauth-authentication-flow.adoc
```diff
@@ -13,14 +14,26 @@ We recommend that you configure clients to use OAUTHBEARER whenever possible.
...
+  # ...
+  authentication:
+    type: oauth
+    # ...
+    enableOauthBearer: true
+----
+
+Many Kafka client tools use libraries that provide basic support for OAUTHBEARER at the protocol level.
+To support application development, Strimzi provides an _OAuth callback handler_ for the upstream Kafka Client Java libraries (but not for other libraries).
 Therefore, you do not need to write your own callback handlers for such clients.
```
mstruk, line 37: Due to the last part in backets:
> (but not for other libraries)

One might interpret 'such clients' to refer to these other libraries?

```suggestion
Therefore, you do not need to write your own callback handlers.
```

documentation/modules/oauth/con-oauth-authentication-flow.adoc
```diff
@@ -65,29 +81,14 @@ When used with the OAuth 2.0 over PLAIN callbacks, Kafka clients authenticate wi
 
 * A long-lived access token, obtained manually at configuration time
 
-The client must be enabled to use PLAIN authentication, and provide a `username` and `password`.
-If the password is prefixed with `$accessToken:` followed by the value of the access token, the Kafka broker will interpret the password as the access token.
+For both methods, the client must provide a `username` and `password` to pass credentials to the Kafka broker.
```
mstruk, line 84:
```suggestion
For both methods, the client must provide `username` and `password` PLAIN parameters to pass credentials to the Kafka broker.
```

documentation/modules/oauth/con-oauth-authentication-flow.adoc
```diff
@@ -65,29 +81,14 @@ When used with the OAuth 2.0 over PLAIN callbacks, Kafka clients authenticate wi
...
+If the password is prefixed with `$accessToken:` followed by the value of an access token, the Kafka broker will interpret the password as the access token.
```
mstruk, line 85:
```suggestion
When listener is configured with `tokenEndpointUri`, if the password is prefixed with `$accessToken:` followed by the value of an access token, the Kafka broker will interpret the password as the access token.
```

documentation/modules/oauth/con-oauth-authentication-flow.adoc
```diff
@@ -65,29 +81,14 @@ When used with the OAuth 2.0 over PLAIN callbacks, Kafka clients authenticate wi
...
 Otherwise, the Kafka broker will interpret the `username` as the client ID and the `password` as the client secret.
```
mstruk, line 87:
```suggestion
When listener has no `tokenEndpointUri` configured, the `password` parameter is never interpreted as a `secret` since it can't be exchanged for an access token without a `tokenEndpoinyUri`. In this case `secret` is always interpreted as an unprefixed `access token`. 
```

documentation/modules/oauth/proc-oauth-authentication-broker-config.adoc
```diff
@@ -133,11 +133,9 @@ Depending on how you apply OAuth 2.0 authentication, and the type of authorizati
...
-<9> Additional configuration for PLAIN mechanism to allow clients to authenticate by passing `clientId` and `secret` as `username` and `password` as described in the previous point. If not specified the clients can authenticate over PLAIN only by passing an access token as `password` parameter.
+<7> Set this to `false` to disable the OAUTHBEARER mechanism on the listener. At least one of PLAIN or OAUTHBEARER has to be enabled. Default is `true`.
+<8> Set to `true` to enable PLAIN authentication on the listener, which is supported for clients on all platforms.
+<9> Additional configuration for the PLAIN mechanism. Clients can authenticate by passing a client ID and secret as a `username` and `password`. Default is `false`. If not specified, clients can authenticate over PLAIN only by passing an access token as the `password` using an `$accessToken:` prefix.
```
mstruk, line 138: I'm not sure what 9 refers to without seeing a rendered doc, but the only setting I can think of here is `tokenEndpointUri` which is not true / false.

```suggestion
<9> Additional configuration for the PLAIN mechanism. Clients can authenticate by passing a client ID and secret as a `username` and `password`, or by passing account name and access token with an `$accessToken:` prefix. If not specified, clients can authenticate over PLAIN only by passing an unprefixed access token as the `password`.
```

mstruk (changes requested):
Found some more issues.

documentation/modules/oauth/con-oauth-authentication-client-options.adoc
PaulRMellor: I've updated as follows:

> If a listener is configured to use PLAIN authentication, the client can authenticate with a client ID and secret or username and access token. These values are passed as the `username` and `password` properties of the PLAIN mechanism.

documentation/modules/oauth/con-oauth-authentication-flow.adoc
PaulRMellor: Tried to break a very long sentence up a bit.

> For both methods, the client must provide the PLAIN properties `username` and `password` to pass credentials to the Kafka broker.
> For the Kafka broker to interpret the password as an access token, it requires an `$accessToken:` prefix followed by the value of the access token. The listener configuration must have a token endpoint (`tokenEndpointUri`) configured for authentication. 
> Otherwise, the Kafka broker will interpret the `username` as the client ID and the `password` as the client secret.
>

documentation/modules/oauth/con-oauth-authentication-flow.adoc
PaulRMellor: Not sure I understand what we're saying here. Is this to replace the preceding sentence, which seems to contradict it?

What we have so far:

> For both methods, the client must provide the PLAIN `username` and `password` properties to pass credentials to the Kafka broker.
> For the Kafka broker to interpret a password as an access token, the following is required:
> 
> * The listener configuration must have a token endpoint (`tokenEndpointUri`) configured 
> * The password requires an `$accessToken:` prefix followed by the value of the access token
> 
> Otherwise, the Kafka broker will interpret the `username` as the client ID and the `password` as the client secret.

documentation/modules/oauth/proc-oauth-authentication-broker-config.adoc
PaulRMellor: Yes. It's the callout for the tokenEndpointUri. 
Maybe this is enough:

> Additional configuration for the PLAIN mechanism. If specified, clients can authenticate over PLAIN by passing an access token as a `password` using an `$accessToken:` prefix.

documentation/modules/oauth/con-oauth-authentication-flow.adoc
mstruk: > If a token endpoint is not configured on the broker listener, ...

Doesn't that sound strange, to first make a categorical claim in the previous paragraph that two conditions are required to use access token for authentication, but in the next sentence add another case (negating that the two conditions are required) where we say that if the first condition is not fulfilled then it's also possible to use access token. Maybe someone else can chime in (@scholzj ?). I would prefer to start from server config being with tokenEndpointUri or without tokenEndpointUri and depending on that explain the options for credentials, rather than trying to explain how to pass access token which is the angle from which the explanation goes here. But maybe that's just me.
 
> the password should not include

Maybe use `the password parameter` or `the password property` to make it clear we have _that_ password in mind.

documentation/modules/oauth/con-oauth-authentication-client-options.adoc
mstruk: Sounds better, yes.

documentation/modules/oauth/proc-oauth-authentication-broker-config.adoc
mstruk: I don't know, maybe. I like to repeat myself in different places because usually we don't read docs from start to finish but go look for a solution. Here being reminded that how you pass access token depends on whether this is configured or not.

documentation/modules/oauth/con-oauth-authentication-client-options.adoc
```diff
@@ -3,33 +3,36 @@
...
+= OAuth 2.0 client authentication flows
 
-In this section, we explain and visualize the communication flow between Kafka client, Kafka broker, and authorization server during Kafka session initiation.
-The flow depends on the client and server configuration.
+[role="_abstract"]
+OAuth 2.0 authentication flows depend on the underlying Kafka client and Kafka broker configuration.
+The flows must also be supported by the authorization server used.
 
-When a Kafka client sends an access token as credentials to a Kafka broker, the token needs to be validated.
+Kafka broker listener configuration determines how clients authenticate using an access token.
```
joaedwar, line 12: The Kafka broker listener ...

documentation/modules/oauth/con-oauth-authentication-client-options.adoc
```diff
@@ -3,33 +3,36 @@
...
-Depending on the authorization server used, and the configuration options available, you may prefer to use:
+If a listener is configured to use PLAIN authentication, the client can authenticate with a client ID and secret or username and access token.
+These values are passed as the `username` and `password` properties of the PLAIN mechanism.
 
-* Fast local token validation based on JWT signature checking and local token introspection, without contacting the authorization server
-* An OAuth 2.0 introspection endpoint provided by the authorization server
+Listener configuration supports the following token validation options:
 
-Using fast local token validation requires the authorization server to provide a JWKS endpoint with public certificates that are used to validate signatures on the tokens.
+* Fast local token validation based on JWT signature checking and local token introspection, without contacting an authorization server.
```
joaedwar, line 20: Refrain from using fragments.

Maybe try...

You can use fast local token validation based on JWT signature checking and local token introspection, without contacting an authorization server.

documentation/modules/oauth/con-oauth-authentication-client-options.adoc
```diff
@@ -3,33 +3,36 @@
...
+The authorization server provides a JWKS endpoint with public certificates that are used to validate signatures on the tokens.
+* A call to a token introspection endpoint provided by an authorization server.
```
joaedwar, line 22: Please rewrite so it is not an incomplete sentence.

documentation/modules/oauth/con-oauth-authentication-client-options.adoc
```diff
@@ -41,39 +44,74 @@ Here you can see the communication flows, for different configurations of Kafka
...
+* xref:oauth-plain-access-token-{context}[Client using a long-lived access token without a client ID and secret]
+
+[id='oauth-plain-client-id-{context}']
+.Client using a client ID and secret, with the broker obtaining the access token for the client
+
+image:oauth-plain-client-id.png[Client using a client ID and secret, with the broker obtaining the access token for the client]
+
+. The Kafka client passes a `clientId` as a username and a `secret` as a password.
+. The Kafka broker uses a token endpoint to pass the `clientId` and `secret` to the authorization server.
+. The authorization server returns a fresh access token in exchange or an error if the client credentials aren't valid.
```
joaedwar, line 98: please remove 'in exchange'

Also, I think it is safer to not use the contraction aren't here. I know the style guide has updated rules on this, but it still poses some issues with translation and localization.

documentation/modules/oauth/con-oauth-authentication-client-options.adoc
```diff
@@ -41,39 +44,74 @@ Here you can see the communication flows, for different configurations of Kafka
...
+. The Kafka broker validates the token in one of the following ways:
+.. If a token introspection endpoint is specified, the Kafka broker validates the access token by calling the endpoint on the authorization server.
+A session is established if the token validation was successful.
```
joaedwar, line 101: I think 

A session is established if the token validation is successful.

documentation/modules/oauth/con-oauth-authentication-client-options.adoc
```diff
@@ -41,39 +44,74 @@ Here you can see the communication flows, for different configurations of Kafka
...
+.. If local token introspection is used, there is no request made to the authorization server.
```
joaedwar, line 102: I think a possible rewrite. 

If local token introspection is used, a request is not made to the authorization server.

documentation/modules/oauth/con-oauth-authentication-client-options.adoc
```diff
@@ -41,39 +44,74 @@ Here you can see the communication flows, for different configurations of Kafka
...
+. The password is passed with or without an `$accessToken:` string prefix depending on whether or not the Kafka broker listener is configured with a token endpoint for authentication.
+.. If the token endpoint is configured, the password should be prefixed by `$accessToken:` to let the broker know that the password parameter contains an access token rather than a client secret. The Kafka broker interprets the username as the account username.
+.. If the token endpoint is not configured on the Kafka broker listener (enforcing a `no-client-credentials mode`), the password should provide the access token without the prefix. The Kafka broker interprets the username as the account username.
+In this mode, the client doesn't use a client ID and secret, and the `password` parameter is always interpreted as a raw access token.
+. The Kafka broker validates the token in one of the following ways:
+.. If a token introspection endpoint is specified, the Kafka broker validates the access token by calling the endpoint on the authorization server. Session is established if token validation was successful.
```
joaedwar, line 116:
```suggestion
If a token introspection endpoint is specified, the Kafka broker validates the access token by calling the endpoint on the authorization server. A session is established if token validation is successful.
```

documentation/modules/oauth/con-oauth-authentication-flow.adoc
```diff
@@ -65,28 +81,21 @@ When used with the OAuth 2.0 over PLAIN callbacks, Kafka clients authenticate wi
...
-  # ...
-  authentication:
-    type: oauth
-    # ...
-    enablePlain: true
-    tokenEndpointUri: https://_OAUTH-SERVER-ADDRESS_/auth/realms/external/protocol/openid-connect/token
-----
+If the `password` is set as the access token, the `username` must be set to the same principal name that the Kafka broker obtains from the access token.
+The process depends on how you configure username extraction using `userNameClaim`, `fallbackUserNameClaim`, `fallbackUsernamePrefix`, or `userInfoEndpointUri`.
+It also depends on your authorization server; in particular, how it maps client IDs to account names.
```
joaedwar, line 98: 'It' should be specified. Sentence is vague

joaedwar (commented):
ltgm. 


Just a few suggestions

documentation/modules/oauth/con-oauth-authentication-flow.adoc
PaulRMellor: Thnaks, @mstruk. Let's rearrange a little:

> For both methods, the client must provide the PLAIN `username` and `password` properties to pass credentials to the Kafka broker. The client uses these properties to pass a client ID and secret or username and access token.
> 
> Client IDs and secrets are used to obtain access tokens.
> 
> Access tokens are passed as `password` property values. 
> You pass the access token with or without an `$accessToken:` prefix. 
> 
> * If you configure a token endpoint (`tokenEndpointUri`) in the listener configuration, you need the prefix.
> * If you don't configure a token endpoint (`tokenEndpointUri`) in the listener configuration, you don't need the prefix. The Kafka broker interprets the password as a raw access token.

documentation/modules/oauth/proc-oauth-authentication-broker-config.adoc
PaulRMellor: Agreed. But this a callout. I don't think we need to be verbose. We have the options covered in a couple of other places.
I've added a link in the _Before you start_ pf this procedure that goes to the _Oauth 2.0 authentication mechanisms_ section as reference

documentation/modules/oauth/con-oauth-authentication-flow.adoc
PaulRMellor: Changed to:

> You can specify username extraction options in your listener using the `userNameClaim`, `fallbackUserNameClaim`, `fallbackUsernamePrefix`, and `userInfoEndpointUri` properties.
> The username extraction process also depends on your authorization server; in particular, how it maps client IDs to account names.

documentation/modules/oauth/con-oauth-authentication-flow.adoc
mstruk: Yeah I think that's better.

mstruk (approved):

joaedwar (approved):
lgtm

joaedwar (approved):

PR #6271, opened 2022-01-27
[systemtest] Fix NPEs when running STs with CO installed with Helm

im-konge:

### Type of change

- Bugfix

### Description

Currently on every execution of `createInstallation()` method inside the `SetupClusterOperatorBuilder` we are completely changing the `instanceHolder`, but without adding `helmResource` and `olmResource`.  In cases, where we are checking equality of default installation and current installation, invocation of new instance rewrites both of the resource variables to null (because they are instance members and not set by the builder in the SetupClusterOperator class), which leads to NPE inside the unInstall() method call..

This PR changes the fields to be `static`, so they will not be overrided and will only change on CO installation.

### Checklist

### DIFF
--- a/systemtest/src/main/java/io/strimzi/systemtest/resources/operator/SetupClusterOperator.java
+++ b/systemtest/src/main/java/io/strimzi/systemtest/resources/operator/SetupClusterOperator.java
@@ -142,7 +142,6 @@ public class SetupClusterOperator {
         if (this.clusterOperatorRBACType == null) {
             this.clusterOperatorRBACType = ClusterOperatorRBACType.CLUSTER;
         }
-        instanceHolder = this;
     }
 
     /**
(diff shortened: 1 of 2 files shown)


see-quick (approved):

Frawless (approved):

im-konge:
@strimzi-ci run tests test_only profile=acceptance env=CLUSTER_OPERATOR_INSTALL_TYPE=helm

sknot-rh (approved):

im-konge:
@strimzi-ci run tests test_only profile=acceptance env=CLUSTER_OPERATOR_INSTALL_TYPE=helm

im-konge:
@strimzi-ci run tests test_only profile=acceptance env=CLUSTER_OPERATOR_INSTALL_TYPE=helm

PR #6290, opened 2022-02-01
Annotate constructors with @Buildable insted of the whole class

Frawless:

### Type of change

- Refactoring

### Description

This PR moves annotation for sundrio to building our model builders from class to constructor level which removes `.setKind()` call in builder constructor and removes https://github.com/fabric8io/kubernetes-client/blob/master/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/CustomResource.java#L172 log which occurred significantly during testing.

### Checklist

### DIFF
--- a/api/src/main/java/io/strimzi/api/kafka/model/Kafka.java
+++ b/api/src/main/java/io/strimzi/api/kafka/model/Kafka.java
@@ -107,6 +107,7 @@ public class Kafka extends CustomResource<KafkaSpec, KafkaStatus> implements Nam
     private Map<String, Object> additionalProperties = new HashMap<>(0);
     private KafkaStatus status;
 
+    @Deprecated
     public Kafka() {}
 
     @Buildable(
--- a/api/src/main/java/io/strimzi/api/kafka/model/Kafka.java
+++ b/api/src/main/java/io/strimzi/api/kafka/model/Kafka.java
@@ -110,6 +110,10 @@ public class Kafka extends CustomResource<KafkaSpec, KafkaStatus> implements Nam
     @Deprecated
     public Kafka() {}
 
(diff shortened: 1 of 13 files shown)


tombentley (commented):
What is the impact on existing callers of the default constructor -- I assume existing code continues to work, only with the warning? 

Should we deprecate the old 0-parameter constructors?

Frawless:
@tombentley  It seems that 0-param constructor is used only in tests now. So I guess it makes sense to deprecate it and remove it from the tests (use the new one) and remove it in 0.29 for example. I didn't find any issues connected to this change yet. I am not sure if someone is using it, in systemtests we don't use it.

tombentley:
@Frawless bear in mind that other people than us use the classes in the `api` module, so we can't just remove the 0-parameter constructor without breaking their code. We can deprecate it of course. But I think we need tests that jackson can still deserialize json when there isn't a default constructor.

Frawless:
Ye, I don't want to remove it completely for the next release. Regarding the tests...you mean tests that will load Kafka yaml and map it to class or?

Frawless:
So I deprecated the old constructors and used new ones in tests. Tests for serialization/deserialization are already in place.

Regression pipeline (just reference) - https://dev.azure.com/cncf/strimzi/_build/results?buildId=100479&view=results

Frawless:
@tombentley could you please do another review?

tombentley (approved):
Should we overload the constructor for people using our API to create their CR instances (where they'll never have a status)? It's minor, but it would let them about passing an explicit `null` argument when instantiating. 

But I'm happy for you to merge it in any case. Thanks.

Frawless:
That's a good point. @scholzj  wdyt?

scholzj:
That is fine with me. Although I'm not sure how many of such users we actually have :-D.

scholzj (approved):

PR #6292, opened 2022-02-01
[systemtest] Refactor of example clients -> test-clients

im-konge:
### Type of change

- Refactoring

### Description

This PR refactors our example-clients to be our internal clients, together with removing builders -> for builder and fluent classes we will use `sundrio`, which will generate everything for us. Thanks to usage of `sundrio` the code is cleaner and much more readable. Also, this is prerequisite for completely removing `test-client` from our tests.

### Checklist


systemtest/src/main/java/io/strimzi/systemtest/kafkaclients/internalClients/KafkaAdminClient.java
```diff
@@ -28,101 +29,64 @@
...
-    }
+    private void checkParametersForBuilder(KafkaAdminClientsBuilder builder) {
+        if (builder.getTopicOperation() == null || builder.getTopicOperation().isEmpty()) {
+            throw new InvalidParameterException("TopicOperation must be set.");
+        }
 
-    public KafkaAdminClient.Builder newBuilder() {
-        return new KafkaAdminClient.Builder();
-    }
+        if (builder.getBootstrapAddress() == null || builder.getBootstrapAddress().isEmpty()) {
```
see-quick, line 82: This is not needed I think. :) When you invoke `super(builder);` constructor it will call also the method from `BaseClients. checkParametersForBuilder`

systemtest/src/main/java/io/strimzi/systemtest/kafkaclients/internalClients/KafkaAdminClient.java
```diff
@@ -28,101 +29,64 @@
...
-            .withBootstrapAddress(getBootstrapAddress())
-            .withTopicOperation(getTopicOperation())
-            .withTopicCount(getTopicCount())
-            .withTopicOffset(getTopicOffset())
-            .withPartitions(getPartitions())
-            .withReplicationFactor(getReplicationFactor())
-            .withKafkaAdminClientName(getAdminName())
-            .withTopicName(getTopicName())
-            .withNamespaceName(getNamespaceName());
+        if ((builder.getTopicName() == null || builder.getTopicName().isEmpty())
```
see-quick, line 86: Here you just need `(builder.getTopicOperation().equals("help") || builder.getTopicOperation().equals("list"))`

see-quick (commented):
Looks good few nits )

systemtest/src/main/java/io/strimzi/systemtest/kafkaclients/internalClients/KafkaAdminClient.java
im-konge: completely forgot about this :) 👍🤲

systemtest/src/main/java/io/strimzi/systemtest/kafkaclients/internalClients/KafkaAdminClient.java
im-konge: thanks! good catch :)

Frawless (approved):

systemtest/src/main/java/io/strimzi/systemtest/kafkaclients/internalClients/KafkaAdminClient.java
im-konge: Discussed offline, this has to be like it is.

systemtest/pom.xml
```diff
@@ -200,10 +200,44 @@
             <artifactId>bcprov-jdk15on</artifactId>
             <version>${bouncycastle.version}</version>
         </dependency>
+        <dependency>
+            <groupId>io.sundr</groupId>
+            <artifactId>builder-annotations</artifactId>
+            <scope>provided</scope>
+        </dependency>
```
tombentley, line 207: In the api module this is the only sundr dependency required. We don't need to explicitly declare a dep on `sundr-core` or `sundr-codegen`

systemtest/src/main/java/io/strimzi/systemtest/kafkaclients/internalClients/BridgeClients.java
```diff
@@ -0,0 +1,155 @@
...
+
+    public int getPort() {
+        return port;
+    }
+
+    public void setPort(int port) {
+        this.port = port;
+    }
+
+    public JobBuilder producerStrimziBridge() {
```
tombentley, line 40: I wasn't clear to me before if we really needed to be returning `JobBuilder` and not `Job` all the call sites except one that I saw just called `build` on the `JobBuilder` returned from these APIs. For the one exception we could either just `new JobBuilder(producerStrimziBridge())` and edit the things we needed, or possibly isn't it a bit of an indication that our builders weren't sufficiently expressive that we needed to go changing the Job returned from our builder in that one call site?

systemtest/src/main/java/io/strimzi/systemtest/kafkaclients/internalClients/BaseClients.java
```diff
@@ -0,0 +1,56 @@
...
+ */
+package io.strimzi.systemtest.kafkaclients.internalClients;
+
+import io.strimzi.systemtest.resources.ResourceManager;
+import io.sundr.builder.annotations.Buildable;
+
+import java.security.InvalidParameterException;
+
+@Buildable(editableEnabled = false)
+public class BaseClients {
```
tombentley, line 13: Should this be `abstract`?

im-konge: changed to `abstract`, makes sense, thanks :)

systemtest/src/main/java/io/strimzi/systemtest/kafkaclients/internalClients/BridgeClients.java
im-konge: Yeah that's right, I guess we kept it to be `JobBuilder` for few occurrences, but now it's just one, which I resolved with `new JobBuilder(..)`. Good catch! :)

see-quick (approved):
Nice job 🦥👌 This looks great )

tombentley (approved):

Frawless (approved):

PR #6295, opened 2022-02-01
[MO] - [system test] -> tests fixes, instance holder, annotation lock…

see-quick:
… changed

### Type of change

- Bugfix
- Enhancement / new feature
- Refactoring
- Documentation

### Description

This PR adds a few fixes:
1. bad namespace handling in our Utils class
2. Resource lock changed to `global` because it could happen a situation where @ParallelTest and IsolatedTest lock their own locks and thus Isolated test can run with parallel one. By specifying the same lock this situation will not happen anymore.
3.  Our diff method in cluster operator could provide NPE because the cluster operator instance was not using the current configuration, which results in NPE when two instances were compared (default and current).
### Checklist


see-quick:
@strimzi-ci run tests test_only test_count_running_in_parallel=4

Frawless (approved):

PR #6319, opened 2022-02-07
Add Vertx shutdown hook to Main class

fvaleri:
`Verticle.stop()` methods are not executed if you don't call `Vertx.close()`. Vertx registers a shutdown hook for that, but only if you use its Launcher as main class.

This PR adds our own shutdown hook, which is similar to what Vertx has in its Launcher, but that we can further customize if needed. As a consequence, the ClusterOpertator.stop() is now executed on JVM's regular shutdown.

### DIFF
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/Main.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/Main.java
@@ -22,10 +22,6 @@ import io.strimzi.operator.common.PasswordGenerator;
 import io.strimzi.operator.common.Reconciliation;
 import io.strimzi.operator.common.Util;
 import io.strimzi.operator.common.operator.resource.ClusterRoleOperator;
-import io.vertx.core.CompositeFuture;
-import io.vertx.core.Future;
-import io.vertx.core.Promise;
-import io.vertx.core.Vertx;
 
 import java.io.BufferedReader;
 import java.io.IOException;
@@ -38,6 +34,10 @@ import java.util.List;
 import java.util.Map;
 import java.util.stream.Collectors;
 
+import io.vertx.core.CompositeFuture;
+import io.vertx.core.Future;
+import io.vertx.core.Promise;
+import io.vertx.core.Vertx;
 import io.vertx.core.VertxOptions;
 import io.vertx.micrometer.MicrometerMetricsOptions;
 import io.vertx.micrometer.VertxPrometheusOptions;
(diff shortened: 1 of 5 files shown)


cluster-operator/src/main/java/io/strimzi/operator/cluster/Main.java
```diff
@@ -47,8 +47,17 @@
 @SuppressFBWarnings("DM_EXIT")
 public class Main {
     private static final Logger LOGGER = LogManager.getLogger(Main.class.getName());
+    private static Vertx vertx = Vertx.vertx(
+        new VertxOptions().setMetricsOptions(
+            new MicrometerMetricsOptions()
+                .setPrometheusOptions(new VertxPrometheusOptions().setEnabled(true))
+                .setJvmMetricsEnabled(true)
+                .setEnabled(true)));
```
scholzj, line 55: IIRC, this did not worked well in general to maker this static. It was causing all kind of problems during tests.

cluster-operator/src/main/java/io/strimzi/operator/cluster/ShutdownHook.java
```diff
@@ -0,0 +1,47 @@
...
+
+import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
+import io.vertx.core.Vertx;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.TimeUnit;
+
+@SuppressFBWarnings("RCN_REDUNDANT_NULLCHECK_OF_NONNULL_VALUE")
```
scholzj, line 15: Why do we need it? Might be good to add some explanation comments.

cluster-operator/src/main/java/io/strimzi/operator/cluster/Main.java
fvaleri: OK, will remove static.

cluster-operator/src/main/java/io/strimzi/operator/cluster/ShutdownHook.java
fvaleri: This is to avoid SB error on `if (vertx != null)`, which could be null.

scholzj (approved):

ppatierno (approved):

PR #6323, opened 2022-02-08
[systemtest] Remove StatefulSets/StrimziPodSets related workarounds from STs

### LINKED ISSUE #5956: [ST] Remove temporary workaround from System Tests and StrimziPodSet handling
As part of #5884, there are several workarounds in the System Tests to handle the StrimziPodSets handling => they either skip the tests or disable some parts of them when the `UseStrimziPodSets` feature gate is used. The are marked with comments in the code referencing to this issue. Once (if) #5884 is merged, these workarounds should be removed.

im-konge:
### Type of change

- Enhancement

### Description

This PR removes workarounds when running STs with FG `+UseStrimziPodSets`. Instead of skipping some checks when we don't have StatefulSets, we will get all needed resources/do certain operations based on environment variable.

Methods located in `StUtils`, which are checking/deleting/getting resources from StatefulSets/StrimziPodSets will be removed together with StatefulSet in the future.

Closes #5956 

### Checklist


systemtest/src/main/java/io/strimzi/systemtest/Environment.java
```diff
@@ -234,6 +234,10 @@ public static boolean isNamespaceRbacScope() {
         return STRIMZI_RBAC_SCOPE_NAMESPACE.equals(STRIMZI_RBAC_SCOPE);
     }
 
+    public static boolean isStrimziPodSetEnabled() {
+        return STRIMZI_FEATURE_GATES.contains(Constants.USE_STRIMZI_POD_SET);
```
scholzj, line 238: I'm not sure there is a way to do this better. But once the feature gate transitions to beta and is enabled by default, this method will stop working. I'm not sure it can be done in a more resilient way, but having some comment abotu that might be helpful.

scholzj (approved):
One nit. Otherwise LGTM assuming it works.

systemtest/src/main/java/io/strimzi/systemtest/Environment.java
im-konge: I will think about it more. This was first and easiest way, how to do it, so I just picked it. Anyway I totally agree, this is not sufficient for the future.

Frawless: Is there an option to get it directly from CO? Maybe from some constant class, but I don't want to introduce a dependency on `cluster-operator` package, but if it's available in `operator-common` we can work with it?

Frawless (approved):
LGTM assuming it works and we finish the discussion in above comment

systemtest/src/main/java/io/strimzi/systemtest/Environment.java
im-konge: Discussed offline -> there is no easy or clear way, how to check this. The comment about this is added and we will have to keep in mind that once `StrimziPodSet` FG will be moved to beta, we need to change this.

PR #6334, opened 2022-02-09
Make KafkaRoller work with list of pod names instead of number of replicas

ShubhamRwt:

### Type of change

_Select the type of your PR_

- Task

### Description

The KafkaRoller currently expects to get as parameter number of Kafka replicas. This PR tries to remove the use of replicas in Kafka Roller and introduce the use of pod list.'

### Checklist

- [ ] Write tests
- [ ] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaAssemblyOperator.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaAssemblyOperator.java
@@ -1202,7 +1202,7 @@ public class KafkaAssemblyOperator extends AbstractAssemblyOperator<KubernetesCl
                                 1_000,
                                 operationTimeoutMs,
                                 () -> new BackOff(250, 2, 10),
-                                replicas,
+                                KafkaCluster.generatePodList(reconciliation.name(), replicas),
                                 compositeFuture.resultAt(0),
                                 compositeFuture.resultAt(1),
                                 adminClientProvider,
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaAssemblyOperator.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaAssemblyOperator.java
@@ -769,7 +769,7 @@ public class KafkaAssemblyOperator extends AbstractAssemblyOperator<KubernetesCl
                                             if (podSet != null) {
                                                 return Future.succeededFuture(KafkaCluster.generatePodList(reconciliation.name(), podSet.getSpec().getPods().size()));
                                             } else {
-                                                return Future.succeededFuture(KafkaCluster.generatePodList(reconciliation.name(), 0));
+                                                return Future.succeededFuture(new ArrayList<String>());
                                             }
                                         });
                             } else {
@@ -778,7 +778,7 @@ public class KafkaAssemblyOperator extends AbstractAssemblyOperator<KubernetesCl
                                             if (sts != null)    {
                                                 return Future.succeededFuture(KafkaCluster.generatePodList(reconciliation.name(), sts.getSpec().getReplicas()));
                                             } else {
-                                                return Future.succeededFuture(KafkaCluster.generatePodList(reconciliation.name(), 0));
+                                                return Future.succeededFuture(new ArrayList<String>());
                                             }
                                         });
                             }
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaAssemblyOperator.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaAssemblyOperator.java
@@ -769,7 +769,7 @@ public class KafkaAssemblyOperator extends AbstractAssemblyOperator<KubernetesCl
                                             if (podSet != null) {
                                                 return Future.succeededFuture(KafkaCluster.generatePodList(reconciliation.name(), podSet.getSpec().getPods().size()));
                                             } else {
-                                                return Future.succeededFuture(new ArrayList<String>());
+                                                return Future.succeededFuture(List.<String>of());
                                             }
                                         });
                             } else {
@@ -778,7 +778,7 @@ public class KafkaAssemblyOperator extends AbstractAssemblyOperator<KubernetesCl
                                             if (sts != null)    {
                                                 return Future.succeededFuture(KafkaCluster.generatePodList(reconciliation.name(), sts.getSpec().getReplicas()));
                                             } else {
-                                                return Future.succeededFuture(new ArrayList<String>());
+                                                return Future.succeededFuture(List.<String>of());
                                             }
                                         });
                             }
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaAssemblyOperator.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaAssemblyOperator.java
@@ -766,19 +766,19 @@ public class KafkaAssemblyOperator extends AbstractAssemblyOperator<KubernetesCl
                             if (featureGates.useStrimziPodSetsEnabled())   {
                                 return strimziPodSetOperator.getAsync(namespace, KafkaResources.kafkaStatefulSetName(name))
                                         .compose(podSet -> {
-                                            if (podSet != null)    {
-                                                return Future.succeededFuture(podSet.getSpec().getPods().size());
+                                            if (podSet != null) {
+                                                return Future.succeededFuture(KafkaCluster.generatePodList(reconciliation.name(), podSet.getSpec().getPods().size()));
                                             } else {
-                                                return Future.succeededFuture(0);
+                                                return Future.succeededFuture(KafkaCluster.generatePodList(reconciliation.name(), 0));
                                             }
                                         });
                             } else {
                                 return stsOperations.getAsync(namespace, KafkaResources.kafkaStatefulSetName(name))
                                         .compose(sts -> {
                                             if (sts != null)    {
-                                                return Future.succeededFuture(sts.getSpec().getReplicas());
+                                                return Future.succeededFuture(KafkaCluster.generatePodList(reconciliation.name(), sts.getSpec().getReplicas()));
                                             } else {
-                                                return Future.succeededFuture(0);
+                                                return Future.succeededFuture(KafkaCluster.generatePodList(reconciliation.name(), 0));
                                             }
                                         });
                             }
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/resource/KafkaAvailability.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/resource/KafkaAvailability.java
@@ -58,7 +58,7 @@ class KafkaAvailability {
      * Determine whether the given broker can be rolled without affecting
      * producers with acks=all publishing to topics with a {@code min.in.sync.replicas}.
      */
-    Future<Boolean> canRoll(int  podId) {
+    Future<Boolean> canRoll(int podId) {
         LOGGER.debugCr(reconciliation, "Determining whether broker {} can be rolled", podId);
         return canRollBroker(descriptions, podId);
     }
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/resource/KafkaAvailability.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/resource/KafkaAvailability.java
@@ -58,7 +58,7 @@ class KafkaAvailability {
      * Determine whether the given broker can be rolled without affecting
      * producers with acks=all publishing to topics with a {@code min.in.sync.replicas}.
      */
-    Future<Boolean> canRoll(int podId) {
+    Future<Boolean> canRoll(int  podId) {
         LOGGER.debugCr(reconciliation, "Determining whether broker {} can be rolled", podId);
         return canRollBroker(descriptions, podId);
     }
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/resource/KafkaRoller.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/resource/KafkaRoller.java
@@ -263,7 +263,7 @@ public class KafkaRoller {
         RestartContext ctx = podToContext.computeIfAbsent(podRef.getPodName(),
             k -> new RestartContext(backoffSupplier));
         singleExecutor.schedule(() -> {
-            LOGGER.debugCr(reconciliation, "Considering restart of pod {} after delay of {} {}", podRef.getPodName(), delay, unit);
+            LOGGER.debugCr(reconciliation, "Considering restart of pod {} after delay of {} {}", podRef, delay, unit);
             try {
                 restartIfNecessary(podRef, ctx);
                 ctx.promise.complete();
@@ -272,7 +272,7 @@ public class KafkaRoller {
                 Thread.currentThread().interrupt();
             } catch (FatalProblem e) {
                 LOGGER.infoCr(reconciliation, "Could not restart pod {}, giving up after {} attempts. Total delay between attempts {}ms",
-                        podRef.getPodName(), ctx.backOff.maxAttempts(), ctx.backOff.totalDelayMs(), e);
+                        podRef, ctx.backOff.maxAttempts(), ctx.backOff.totalDelayMs(), e);
                 ctx.promise.fail(e);
                 singleExecutor.shutdownNow();
                 podToContext.forEachValue(Integer.MAX_VALUE, f -> {
@@ -281,14 +281,14 @@ public class KafkaRoller {
             } catch (Exception e) {
                 if (ctx.backOff.done()) {
                     LOGGER.infoCr(reconciliation, "Could not roll pod {}, giving up after {} attempts. Total delay between attempts {}ms",
-                            podRef.getPodName(), ctx.backOff.maxAttempts(), ctx.backOff.totalDelayMs(), e);
+                            podRef, ctx.backOff.maxAttempts(), ctx.backOff.totalDelayMs(), e);
                     ctx.promise.fail(e instanceof TimeoutException ?
                             new io.strimzi.operator.common.operator.resource.TimeoutException() :
                             e);
                 } else {
                     long delay1 = ctx.backOff.delayMs();
                     LOGGER.infoCr(reconciliation, "Could not roll pod {} due to {}, retrying after at least {}ms",
-                            podRef.getPodName(), e, delay1);
+                            podRef, e, delay1);
                     schedule(podRef, delay1, TimeUnit.MILLISECONDS);
                 }
             }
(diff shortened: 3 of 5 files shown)


cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaAssemblyOperator.java
```diff
@@ -728,6 +728,18 @@ public void checkCustomCaSecret(CertificateAuthority ca, Secret certSecret, Secr
...
+
+            ArrayList<String> podNames = new ArrayList<>(replicas);
+
+            for (int podId = 0; podId < replicas; podId++) {
+
+                podNames.add(KafkaCluster.kafkaPodName(reconciliation.name(), podId));
+
+            }
+            return podNames;
+        }
```
scholzj, line 741: I guess having this in KafkaCluster would make more sense.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaAssemblyOperator.java
```diff
@@ -761,25 +773,25 @@ public void checkCustomCaSecret(CertificateAuthority ca, Secret certSecret, Secr
                             if (featureGates.useStrimziPodSetsEnabled())   {
                                 return strimziPodSetOperator.getAsync(namespace, KafkaCluster.kafkaClusterName(name))
                                         .compose(podSet -> {
-                                            if (podSet != null)    {
-                                                return Future.succeededFuture(podSet.getSpec().getPods().size());
+                                            if (podSet != null) {
+                                                return Future.succeededFuture(generatePodList(podSet.getSpec().getPods().size()));
                                             } else {
-                                                return Future.succeededFuture(0);
+                                                return Future.succeededFuture(generatePodList(0));
```
scholzj, line 779: Maybe `List.of()` would be cleaner and less confusing?

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/resource/KafkaRoller.java
```diff
@@ -195,15 +195,15 @@ private boolean initAdminClient() {
 
         Promise<Void> result = Promise.promise();
         singleExecutor.submit(() -> {
-            List<Integer> podIds = new ArrayList<>(numPods);
+            List<Integer> podIds = new ArrayList<>(podList.size());
```
scholzj, line 198: I do not think they are podIds anymore => they are now just indexes in the list. So I think it should be names appropriately here and in th other places (in the for loop etc.).

cluster-operator/src/test/java/io/strimzi/operator/cluster/operator/resource/KafkaRollerTest.java
```diff
@@ -206,11 +207,21 @@ public void controllerNotReadyAfterRolling(VertxTestContext testContext) throws
...
 
+    public List<String> addPodNames(int replicas) {
+        ArrayList<String> podNames = new ArrayList<>(replicas);
+
+        for (int podId = 0; podId < replicas; podId++) {
+            podNames.add(KafkaCluster.kafkaPodName(clusterName(), podId));
+        }
+
+        return podNames;
+    }
```
scholzj, line 218: I think (one of the) proper tests here would be to try it with disconnected list of names. E.g. `my-cluster-kafka-0`, `my-cluster-kafka-1000`, `my-cluster-kafka-controller-2000` for example and check they were all rolled.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/resource/KafkaRoller.java
```diff
@@ -195,15 +195,15 @@ private boolean initAdminClient() {
...
                 // Order the podIds unready first otherwise repeated reconciliations might each restart a pod
                 // only for it not to become ready and thus drive the cluster to a worse state.
-                podIds.add(podOperations.isReady(namespace, podName(podId)) ? podIds.size() : 0, podId);
+                podIds.add(podOperations.isReady(namespace, podList.get(podId)) ? podIds.size() : 0, podId);
             }
             LOGGER.debugCr(reconciliation, "Initial order for rolling restart {}", podIds);
-            List<Future> futures = new ArrayList<>(numPods);
+            List<Future> futures = new ArrayList<>(podList.size());
             for (Integer podId: podIds) {
                 futures.add(schedule(podId, 0, TimeUnit.MILLISECONDS));
```
scholzj, line 208: Does this really work? I do not see you changing the schedule method => so how will it get the pod name from the pod ID?

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaAssemblyOperator.java
```diff
@@ -761,25 +761,25 @@ public void checkCustomCaSecret(CertificateAuthority ca, Secret certSecret, Secr
                             if (featureGates.useStrimziPodSetsEnabled())   {
                                 return strimziPodSetOperator.getAsync(namespace, KafkaCluster.kafkaClusterName(name))
                                         .compose(podSet -> {
-                                            if (podSet != null)    {
-                                                return Future.succeededFuture(podSet.getSpec().getPods().size());
+                                            if (podSet != null) {
+                                                return Future.succeededFuture(KafkaCluster.generatePodList(reconciliation.name(), podSet.getSpec().getPods().size()));
                                             } else {
-                                                return Future.succeededFuture(0);
+                                                return Future.succeededFuture(KafkaCluster.generatePodList(reconciliation.name(), 0));
```
scholzj, line 767: Should this just be `List.of()`?

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaAssemblyOperator.java
```diff
@@ -761,25 +761,25 @@ public void checkCustomCaSecret(CertificateAuthority ca, Secret certSecret, Secr
...
                                         });
                             } else {
                                 return stsOperations.getAsync(namespace, KafkaCluster.kafkaClusterName(name))
                                         .compose(sts -> {
                                             if (sts != null)    {
-                                                return Future.succeededFuture(sts.getSpec().getReplicas());
+                                                return Future.succeededFuture(KafkaCluster.generatePodList(reconciliation.name(), sts.getSpec().getReplicas()));
                                             } else {
-                                                return Future.succeededFuture(0);
+                                                return Future.succeededFuture(KafkaCluster.generatePodList(reconciliation.name(), 0));
```
scholzj, line 776: Same as above ... should this be just `List.of()`?

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/resource/KafkaAvailability.java
```diff
@@ -58,7 +58,7 @@
      * Determine whether the given broker can be rolled without affecting
      * producers with acks=all publishing to topics with a {@code min.in.sync.replicas}.
      */
-    Future<Boolean> canRoll(int podId) {
+    Future<Boolean> canRoll(int  podId) {
```
scholzj, line 61: The additional space is probably not needed?

cluster-operator/src/test/java/io/strimzi/operator/cluster/operator/resource/KafkaRollerTest.java
scholzj: > I think (one of the) proper tests here would be to try it with disconnected list of names. E.g. my-cluster-kafka-0, my-cluster-kafka-1000, my-cluster-kafka-controller-2000 for example and check they were all rolled.

Might be still useful to try this - at least in some tests ^^^

scholzj (commented):
I left some comments. Stupid question which came to my mind ... any place ordering in Map / List might play a role and screw things up? Like when requing the controller to be rolled last etc.?

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaAssemblyOperator.java
ShubhamRwt: List.of() will return me a list of `object` type(List<E>) while the `generatePodList` podList method is returning me a List of string type(List<String) and since this if else is later on followed by a compose so the `future` returned should be of one type (List<E> or List<String>)

scholzj: Fair enough ... but I guess you can still do something like `new ArrayList<String>(0)`?

ShubhamRwt:
```any place ordering in Map / List might play a role and screw things up? Like when requing the controller to be rolled last etc``` @scholzj Hi, I did had a look again but I am not able to look for a place where we are doing something in regards to the ordering.

File: cluster-operator/src/test/java/io/strimzi/operator/cluster/operator/resource/KafkaRollerTest.java (hunk 4 above)
Comment (ShubhamRwt) [reply cluster-operator/src/test/java/io/strimzi/operator/cluster/operator/resource/KafkaRollerTest.java:218]: Should I create a separate test which do this?

Comment (scholzj) [reply cluster-operator/src/test/java/io/strimzi/operator/cluster/operator/resource/KafkaRollerTest.java:218]: I think as a separate test it would be fine, used in the existing tests is fine as well.

File: cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/resource/KafkaRoller.java [hunk 9]
```diff
@@ -124,21 +124,21 @@
...
-                       long pollingIntervalMs, long operationTimeoutMs, Supplier<BackOff> backOffSupplier,
-                       int replicas, Secret clusterCaCertSecret, Secret coKeySecret,
+                       long pollingIntervalMs, long operationTimeoutMs, Supplier<BackOff> backOffSupplier, List<String> podList,
+                        Secret clusterCaCertSecret, Secret coKeySecret,
                        AdminClientProvider adminClientProvider,
                        String kafkaConfig, String kafkaLogging, KafkaVersion kafkaVersion, boolean allowReconfiguration) {
         this.namespace = reconciliation.namespace();
         this.cluster = reconciliation.name();
-        this.numPods = replicas;
+        this.podList = podList;
```
tombentley, line 141: Let's add a check that the names in `podList` are unique: `if (podList.size() != podList.stream().distinct().count()) throw new IAE`.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/resource/KafkaRoller.java
```diff
@@ -192,20 +192,19 @@ private boolean initAdminClient() {
...
         this.podNeedsRestart = podNeedsRestart;
-
         Promise<Void> result = Promise.promise();
         singleExecutor.submit(() -> {
-            List<Integer> podIds = new ArrayList<>(numPods);
+            List<String> podNames = new ArrayList<>(podList.size());
 
-            for (int podId = 0; podId < numPods; podId++) {
-                // Order the podIds unready first otherwise repeated reconciliations might each restart a pod
+            for (int podName = 0; podName < podList.size(); podName++) {
```
tombentley, line 199: It's not a pod name, it's a `podIndex`

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/resource/KafkaRoller.java
```diff
@@ -486,36 +485,40 @@ private RestartPlan restartPlan(int podId, Pod pod, RestartContext restartContex
...
                 needsReconfig = true;
             }
         } else if (needsRestart) {
-            LOGGER.infoCr(reconciliation, "Pod {} needs to be restarted. Reason: {}", podId, reasonToRestartPod);
+            LOGGER.infoCr(reconciliation, "Pod {} needs to be restarted. Reason: {}", podName, reasonToRestartPod);
         }
         return new RestartPlan(needsRestart, needsReconfig, podStuck, diff, loggingDiff);
     }
 
+    int getPodIndexFromPodName(String podName)  {
```
tombentley, line 511: No need for the `get`! It's a function; it's allowed to return a value without needing to make a song and dance about it by prefixing with `get`.

`indexOfPod`?

cluster-operator/src/test/java/io/strimzi/operator/cluster/operator/resource/KafkaRollerTest.java
tombentley: I'd prefer changing the existing test.

tombentley (commented):
A new nits.

I wonder whether, rather than just replacing `int` with `String` we should have a `PodRef` class (it could be an inner class of `KafkaRoller`). This would allow you to have both the pod name and the index in one instance, passed around within `KafkaRoller` and thus avoid needing to compute the index in several places. It should also make it easier to refactor in the future when pods needn't be in the same kube cluster, as it would allow you to add some reference to the cluster+namespace containing the pod.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/resource/KafkaRoller.java
```diff
@@ -808,4 +810,23 @@ public String toString() {
...
+        }
+
+        public String getPodName() {
+            return podName;
+        }
+
+        public int getPodIndex() {
+            return podIndex;
+        }
+
```
tombentley, line 830: You can simplify all you logging calls by overriding `toString`. You should probably also override `equals` and `hashCode` too, in terms of the `podName`.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/resource/KafkaRoller.java
```diff
@@ -808,4 +810,23 @@ public String toString() {
             });
     }
 
+    public class PodRef {
+
+        private String podName;
+        private int podIndex;
```
tombentley, line 816: Can both be `final`

tombentley: And perhaps adding `namespace` allows you to simplify on of the `isReady` methods too?

tombentley (approved):
LGTM, assuming the test all pass following the rebase. Thanks @ShubhamRwt!

ShubhamRwt:
> @ShubhamRwt Could you please rebase this so that we can run the tests on it?
Done @scholzj

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/resource/KafkaRoller.java
ShubhamRwt: I tried adding the `namespace` but the namespace is being used directly and easily at the places where I am passing `podRef`

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/resource/KafkaRoller.java
ShubhamRwt: Should I create the `hashcode` and `equals` method ? They are not being used in the current code

tombentley: I would, personally.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/resource/KafkaRoller.java
tombentley: OK

scholzj (approved):
LGTM. Thanks.

scholzj:
@tombentley @ShubhamRwt It seems that you are still discussing some details of the comments. Please let me know when its done.

ShubhamRwt:
@tombentley @scholzj Hi, I have made the changes.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaAssemblyOperator.java
```diff
@@ -766,19 +766,19 @@ public void checkCustomCaSecret(CertificateAuthority ca, Secret certSecret, Secr
                             if (featureGates.useStrimziPodSetsEnabled())   {
                                 return strimziPodSetOperator.getAsync(namespace, KafkaResources.kafkaStatefulSetName(name))
                                         .compose(podSet -> {
-                                            if (podSet != null)    {
-                                                return Future.succeededFuture(podSet.getSpec().getPods().size());
+                                            if (podSet != null) {
+                                                return Future.succeededFuture(KafkaCluster.generatePodList(reconciliation.name(), podSet.getSpec().getPods().size()));
                                             } else {
-                                                return Future.succeededFuture(0);
+                                                return Future.succeededFuture(new ArrayList<String>());
```
tombentley, line 772: `List.of()`

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaAssemblyOperator.java
```diff
@@ -766,19 +766,19 @@ public void checkCustomCaSecret(CertificateAuthority ca, Secret certSecret, Secr
...
                                         });
                             } else {
                                 return stsOperations.getAsync(namespace, KafkaResources.kafkaStatefulSetName(name))
                                         .compose(sts -> {
                                             if (sts != null)    {
-                                                return Future.succeededFuture(sts.getSpec().getReplicas());
+                                                return Future.succeededFuture(KafkaCluster.generatePodList(reconciliation.name(), sts.getSpec().getReplicas()));
                                             } else {
-                                                return Future.succeededFuture(0);
+                                                return Future.succeededFuture(new ArrayList<String>());
```
tombentley, line 781: `List.of()`

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/resource/KafkaRoller.java
```diff
@@ -397,6 +399,10 @@ private boolean podWaitingBecauseOfAnyReasons(Pod pod, Set<String> reasons) {
         return false;
     }
 
+    private int indexOfPod(String podName)  {
```
tombentley, line 402: This should be `idOfPod`.

tombentley (approved):
A few last nits to fit, but apart from this LGTM. Thanks @ShubhamRwt!

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaAssemblyOperator.java
ShubhamRwt: List.of() will return me a list of object type(List) while the generatePodList podList method is returning me a List of string type(List<String) and since this if-else is later on followed by a compose so the future returned should be of one type either List<E> or List<String>

tombentley: `List.<String>of()` then if type arg inference doesn't work.

scholzj:
Nice job @ShubhamRwt ... thanks.

PR #6336, opened 2022-02-09
[MO] - [system test] -> LogCollector TS & TC labels + @ParallelSuite …

see-quick:
### Type of change

- Enhancement / new feature
- Refactoring

### Description

This PR adds LogCollector class two additional features.

1. **Distinquisted logs between test cases** = currently when we enable parallelism (f.e. 5 in parallel) and ∀ these 5 test cases we @ParallelTest, it will mean that everything will be deployed inside one namespace. Thus when something wrong happens in one test case, it will collect ∀ logs from every pod, even un-related to the test case.  
2. **Handle logs in auxiliary namespaces of `@ParallelSuite`.** = for instance for TracingST we create a new namespace `tracing-st` and currently LogCollector collects only the `infra-namespace` namespace.  By this change, it would be able to collect also `tracing-st` namespace and others.

### Checklist

- [x] BeforeAll fail
- [x] BeforeEach fail
- [x] IsolatedTest fail
- [x] IsolatedTest fail
- [x] ParallelTest fail
- [x] ParallelNamespaceTest fail
- [x] AfterEach fail
- [x] AfterAll fail
- [x] Distinquisted logs between test cases 
- [x] Handle logs in auxiliary namespaces of `@ParallelSuite`


systemtest/src/main/java/io/strimzi/systemtest/logs/LogCollector.java
```diff
@@ -152,6 +180,32 @@ public synchronized void collect() {
         });
     }
 
+    private final void collectLogsForTestSuite(final Pod pod) {
+        if (pod.getMetadata().getLabels().containsKey(Constants.TEST_SUITE_NAME_LABEL)) {
+            if (pod.getMetadata().getLabels().get(Constants.TEST_SUITE_NAME_LABEL).equals(this.collectorElement.getTestClassName())) {
+                LOGGER.trace("Collecting logs for TestSuite: {}, and Pod: {}", this.collectorElement.getTestClassName(), pod.getMetadata().getName());
```
Frawless, line 186: wouldn't debug better option? OR maybe even info? Same in the bellow usages

see-quick: I could set it to `DEBUG` for sure.

see-quick:
@scholzj It does not matter if parallelism is enabled or disabled in this case, because when we always create auxiliary namespaces. The same applies to labelling test cases with test names when creating resources.
---
I can double-check :) but overall I think it will 100% work.

scholzj:
> @scholzj It does not matter if parallelism is enabled or disabled in this case, because when we always create auxiliary namespaces. The same applies to labelling test cases with test names when creating resources.

Well, then I guess the question is if you doublechecked that logging still works on Azure as well :-)

see-quick:
> Well, then I guess the question is if you doublechecked that logging still works on Azure as well :-)

Log collecting works for Azure as well. But I have to solve the problem with labels names, which are more than 63 characters. :)

PR #6353, opened 2022-02-11
docs: new feature gate versions section

PaulRMellor:

**Documentation**
Updates the _Using Strimzi Operators_ section of the _Configuring_ guide.
Separates the lifecycle stages and Strimzi versioning information from the feature gate descriptions to make clearer.
Pulls the sections up a level to make the content easier to locate

- [ ] Write tests
- [x] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/documentation/modules/operators/ref-operator-cluster-feature-gate-releases.adoc
+++ b/documentation/modules/operators/ref-operator-cluster-feature-gate-releases.adoc
@@ -17,9 +17,9 @@ Beta stage features are well tested and their functionality is not likely to cha
 GA stage features are stable and should not change in the future.
 Alpha and beta stage features are removed if they do not prove to be useful.
 
-* The `ControlPlaneListener` feature gate moved to beta stage in Strimzi 0.27.0 and is expected to remain in the beta stage until Strimzi 0.31.
-* The `ServiceAccountPatching` feature gate moved to beta stage in Strimzi 0.27.0 and is expected to remain in the beta stage until Strimzi 0.30.
-* The `UseStrimziPodSets` feature gate is currently planned to move to the beta stage in Strimzi 0.30.0.
+* The `ControlPlaneListener` feature gate moved to beta stage in Strimzi 0.27 and is expected to remain in the beta stage until Strimzi 0.31.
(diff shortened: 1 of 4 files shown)


documentation/modules/operators/ref-operator-cluster-feature-gate-releases.adoc
```diff
@@ -0,0 +1,77 @@
...
+
+¦`ServiceAccountPatching`
+¦-
+¦-
+
+¦`UseStrimziPodSets`
+¦-
+¦0.27 and earlier
+
+|===
```
scholzj, line 77: I don't think this is helpful too much => or rather not sufficient. This information needs to be in the part about the individual feature gates. Otherwise people will miss it. It probably also doesn't need the feature gate to which it doesn't apply in this table.

PaulRMellor: Won't this table grow potentially? And other feature gates need this kind of information. Maybe it's good to have it in one place for easy maintenance and reference, with an _important_ note and link from the related feature gate descriptions section:

image

Otherwise, if we want this info with the feature gate descriptions, can we use attributes?

scholzj: Well, the question is for whom it is good to have it in one place. It is certainly easier for maintenance. But I'm not sure it is easier for the user.

My expectation is that users will be interested in particular feature gate. So the chapter about the specific feature gate should:
* explain whether it is enabled or disabled
* cover all important aspects such as upgrade or downgrade limitations
* possibly suggest when it might move to the next phase
* I do not think it necessarily needs to list when the phase changed last time

I think these things are too important to hide behind some generic link, they need to be placed directly.

When referring someone to it, you will again refer them to an exact feature gate because of some discussion about something. Not so some list of some upgrade issues. I do not think people will come and ask _Can you please point me to the list of FeatureGates with some special upgrade / downgrade issues?_. They will ask for some feature they need and you will point them to a specific Feature Gate related to it.

To be clear, I do not have problem to keep the table -> but the information needs to be in the other place as well.

PaulRMellor: Thanks, @scholzj. I've adjusted the descriptions of each feature gate.

scholzj (approved):

PR #6364, opened 2022-02-14
Removed unused metric options from AbstractOpertorTest

vbusch:
### Type of change

- Bugfix

### Description

I had a branch where the unit test classes were being run in a different order.  It is easily reproduced by adding `<runOrder>alphabetical</runOrder>`
If AbstractOperatorTest runs before OperatorMetricsTest, then OperatorMetricsTest will fail with `createCleanMetricsProvider:516 » NullPointer`

AbstractOperatorTest does not need to setup the metricsOptions.  And starting the metrics and then closing it via `vertx.close` seems to be causing trouble for the other test (OperatorMetricTest).

### Checklist

- [ X] Make sure all tests pass


scholzj (approved):
Thanks for the PR.

ppatierno (approved):

scholzj:
@vbusch Thanks for the PR.

PR #6382, opened 2022-02-16
[DOC] Make Kafka Listener docs a bit more clear

scholzj:
### Type of change

- Bugfix
- Documentation

### Description

This PR cleans up the docs about the Kafka listener certificates:
* Remove the list of listener types where it can be used which is not needed since the list contains all listeners with TLS
* Makes it hopefully a bit more clear that this is about the server certificate and has no relation ship to TLS client authentication

This is based on discussion from #6355


PaulRMellor (approved):
Looks good :+1:

Issue #6367, opened 2022-02-14
StrimziTestContainer code style improvements 


see-quick:
### Description

Tom has some suggestions in https://github.com/strimzi/strimzi-kafka-operator/pull/6337 but currently, we cannot apply them, because we would need to release a new version of the test container. 

### Steps

1. https://github.com/strimzi/test-container/pull/27
2. https://github.com/strimzi/test-container/pull/28
3. https://github.com/strimzi/test-container/pull/30
4. release `0.101.0` test-container

PR #6400, opened 2022-02-20
[MO] - optimize unnecessary 3-node cluster in ITs 

see-quick:
### Type of change

- Enhancement / new feature
- Refactoring

### Description

This PR optimizing ITs, where we used unnecessary three nodes Kafka clusters. I replaced ∀ them to use single-node ∧ thus reducing preparation time 👌🦥. Moreover it resolves this issue [1].

❗️ Before merging this PR we need to release test-container `0.101.0`.

[1] - https://github.com/strimzi/strimzi-kafka-operator/issues/6367

### Checklist

### DIFF
--- a/topic-operator/src/test/java/io/strimzi/operator/topic/TopicOperatorBaseIT.java
+++ b/topic-operator/src/test/java/io/strimzi/operator/topic/TopicOperatorBaseIT.java
@@ -17,6 +17,7 @@ import io.strimzi.api.kafka.model.status.Condition;
 import io.strimzi.api.kafka.model.status.KafkaTopicStatus;
 import io.strimzi.test.TestUtils;
 import io.strimzi.test.container.StrimziKafkaCluster;
+import io.strimzi.test.container.StrimziKafkaContainer;
 import io.strimzi.test.k8s.KubeClusterResource;
 import io.strimzi.test.k8s.cluster.KubeCluster;
 import io.strimzi.test.k8s.exceptions.NoClusterException;
@@ -39,6 +40,7 @@ import org.apache.kafka.common.errors.UnknownTopicOrPartitionException;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.junit.jupiter.api.TestInstance;
+import org.testcontainers.lifecycle.Startable;
 
 import java.io.IOException;
 import java.util.HashMap;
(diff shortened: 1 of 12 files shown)


scholzj:
@see-quick  Is this still relevant after #6401?

see-quick:
> @see-quick Is this still relevant after #6401?

Yes, it is :). That PR just moved Kafka container and topic operator to be shared. This PR replaces multiple node clusters with only Kafka container (single node) if possible. But I need to release the test containers to proceed with this...

scholzj (approved):

topic-operator/src/test/java/io/strimzi/operator/topic/TopicOperatorBaseIT.java
```diff
@@ -226,10 +238,21 @@ protected void startTopicOperator(StrimziKafkaCluster kafkaCluster) throws Inter
         LOGGER.info("Started Topic Operator");
     }
 
-    protected Map<String, String> topicOperatorConfig(StrimziKafkaCluster kafkaCluster) {
-        Map<String, String> m = new HashMap<>();
-        m.put(Config.KAFKA_BOOTSTRAP_SERVERS.key, kafkaCluster.getBootstrapServers());
-        m.put(Config.ZOOKEEPER_CONNECT.key, kafkaCluster.getZookeeper().getHost() + ":" + kafkaCluster.getZookeeper().getFirstMappedPort());
+    protected <T extends Startable> Map<String, String> topicOperatorConfig(final T strimziContainer) {
```
tombentley, line 241: Do we really need the type parameter? Can't the value parameter have type `Startable`?

It any case it feels a bit rubbish to require a dependency on `testcontainers`. We don't actually need any of the methods of `Startable`, we're just using it because it's a convenient supertype of both `StrimziKafkaCluster` and `StrimziKafkaContainer`. Do you think it's worth having a common superinterface for `StrimziKafkaCluster` and `StrimziKafkaContainer` in `strimzi-test-container`, to allow precisely this kind of abstraction in users of `strimzi-test-container`? It would also be more typesafe, since trying to call this method with any old `Startable` would then be a compile time error.

tombentley: And looking at how you're using `strimziContainer` it seems this interface could just expose a `getBootstrapServers` and `getZookeeperConnectString()` (and maybe some methods for the kraft case?) thus avoiding the `instanceof` entirely.

see-quick: Yes, I think this is a good idea to create an interface for such containers, but it would mean another release to get this PR merged. But overall I think it could wait until next release of test container :)

PR #6407, opened 2022-02-22
Shellcheck compliant

fvaleri:
Make log-dump and cold-backup scripts compliant with shellcheck, and add them to the shellcheck script in our CI.

### DIFF
--- a/.azure/scripts/shellcheck.sh
+++ b/.azure/scripts/shellcheck.sh
@@ -12,9 +12,9 @@ docker-images/kafka-based/kafka/stunnel-scripts/*.sh
 docker-images/kafka-based/kafka/exporter-scripts/*.sh
 docker-images/jmxtrans/*.sh
 tools/olm-bundle/*.sh
-tools/*.sh
-tools/cold-backup/run.sh
-tools/log-dump/run.sh"
+tools/cold-backup/*.sh
+tools/log-dump/*.sh
+tools/*.sh"
 
 for SCRIPTS in $SCRIPT_DIRS; do
     shellcheck -a -P $(dirname "$SCRIPTS") -x "$SCRIPTS"
--- a/.azure/scripts/shellcheck.sh
+++ b/.azure/scripts/shellcheck.sh
@@ -12,7 +12,9 @@ docker-images/kafka-based/kafka/stunnel-scripts/*.sh
 docker-images/kafka-based/kafka/exporter-scripts/*.sh
 docker-images/jmxtrans/*.sh
 tools/olm-bundle/*.sh
-tools/*.sh"
+tools/*.sh
+tools/cold-backup/run.sh
+tools/log-dump/run.sh"
 
 for SCRIPTS in $SCRIPT_DIRS; do
     shellcheck -a -P $(dirname "$SCRIPTS") -x "$SCRIPTS"
--- a/systemtest/src/test/java/io/strimzi/systemtest/backup/ColdBackupScriptIsolatedST.java
+++ b/systemtest/src/test/java/io/strimzi/systemtest/backup/ColdBackupScriptIsolatedST.java
@@ -21,7 +21,6 @@ import io.strimzi.test.annotations.IsolatedTest;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.apache.logging.log4j.Level;
-import org.junit.jupiter.api.BeforeAll;
 import org.junit.jupiter.api.Tag;
 
 import io.strimzi.systemtest.AbstractST;
@@ -36,11 +35,13 @@ import org.junit.jupiter.api.extension.ExtensionContext;
 @Tag(INTERNAL_CLIENTS_USED)
 @IsolatedSuite
 public class ColdBackupScriptIsolatedST extends AbstractST {
-
     private static final Logger LOGGER = LogManager.getLogger(ColdBackupScriptIsolatedST.class);
 
     @IsolatedTest
     void backupAndRestore(ExtensionContext context) {
+        clusterOperator.unInstall();
+        clusterOperator = clusterOperator.defaultInstallation().createInstallation().runInstallation();
+        
         String clusterName = mapWithClusterNames.get(context.getDisplayName());
         String groupId = "my-group", newGroupId = "new-group";
         int firstBatchSize = 100, secondBatchSize = 10;
@@ -125,11 +126,5 @@ public class ColdBackupScriptIsolatedST extends AbstractST {
                 .withMessageCount(batchSize)
                 .build();
         return clients;
-    }
-
-    @BeforeAll
-    void setUp() {
-        clusterOperator.unInstall();
-        clusterOperator.defaultInstallation().createInstallation().runInstallation();
     }
 }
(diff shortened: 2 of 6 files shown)


.azure/scripts/shellcheck.sh
```diff
@@ -12,7 +12,9 @@ docker-images/kafka-based/kafka/stunnel-scripts/*.sh
 docker-images/kafka-based/kafka/exporter-scripts/*.sh
 docker-images/jmxtrans/*.sh
 tools/olm-bundle/*.sh
-tools/*.sh"
+tools/*.sh
+tools/cold-backup/run.sh
+tools/log-dump/run.sh"
```
scholzj, line 17: Should this instead of `run.sh` be `*.sh` to cover any scripts added int he future?

systemtest/src/test/java/io/strimzi/systemtest/backup/ColdBackupScriptIsolatedST.java
```diff
@@ -36,11 +35,13 @@
 @Tag(INTERNAL_CLIENTS_USED)
 @IsolatedSuite
 public class ColdBackupScriptIsolatedST extends AbstractST {
-
     private static final Logger LOGGER = LogManager.getLogger(ColdBackupScriptIsolatedST.class);
 
     @IsolatedTest
     void backupAndRestore(ExtensionContext context) {
+        clusterOperator.unInstall();
+        clusterOperator = clusterOperator.defaultInstallation().createInstallation().runInstallation();
```
scholzj, line 43: Why was this change needed?

CC @see-quick

tools/cold-backup/README.md
```diff
@@ -31,25 +29,22 @@ kubectl create -f ./examples/kafka/kafka-persistent.yaml
...
+  bin/kafka-consumer-groups.sh --bootstrap-server :9092 --group my-group --describe > /tmp/offsets.txt
 
 # send additional 12345 messages
-kubectl run kafka-producer-perf-test -it --image="$CLIENT_IMAGE" \
-  --rm="true" --restart="Never" -- bin/kafka-producer-perf-test.sh \
-  --topic my-topic --record-size 1000 --num-records 12345 --throughput -1 \
-  --producer-props acks=1 bootstrap.servers="my-cluster-kafka-bootstrap:9092"
+kubectl exec -it my-cluster-kafka-0 -c kafka -- \
+  bin/kafka-producer-perf-test.sh --topic my-topic --record-size 1000 --num-records 12345 \
+  --throughput -1 --producer-props acks=1 bootstrap.servers="my-cluster-kafka-bootstrap:9092"
```
scholzj, line 47: Why change to `kubectl exec`? That is a bad pattern which should not be encouraged by our docs. It can eaily cause the brokers to get killed etc.

tools/cold-backup/run.sh
```diff
@@ -331,32 +327,37 @@ restore() {
...
   if [[ $CONFIRM == true ]]; then
     echo "Restore of cluster $NAMESPACE/$CLUSTER_NAME"
     confirm
   fi
-  uncompress $SOURCE_FILE $tmp
-  source $tmp/env
+  
+  local tmp="$BACKUP_PATH/$NAMESPACE/$CLUSTER_NAME"
+  uncompress "$SOURCE_FILE" "$tmp"
+  # shellcheck disable=SC1091
```
scholzj, line 343: Maybe you can add a short comment why is it needed into the code (I assume it cannot follow the code since it does nto exist)

scholzj (commented):
Some nits. LGTM otherwise. I will trigger the regression.

systemtest/src/test/java/io/strimzi/systemtest/backup/ColdBackupScriptIsolatedST.java
fvaleri: The error I had consistently on ColdBackup ST is "ZK PVC is already created", or something like that. So I just did what I found in the LogDump ST and it worked.

tools/cold-backup/README.md
fvaleri: Yeah, this is just to avoid updating the image version at every release. I can add a note if you think it is needed.

tools/cold-backup/run.sh
fvaleri: https://github.com/koalaman/shellcheck/wiki/SC1091

.azure/scripts/shellcheck.sh
fvaleri: OK, but what about any sub dir containing scripts?

tools/cold-backup/README.md
scholzj: Do we need to update the image every release? I assume it will keep working in general. You can also use a placeholder for the image and let users to fill it in themself.

tools/cold-backup/run.sh
scholzj: Well, that is what I assumed. But would be still good to have it in the code.

.azure/scripts/shellcheck.sh
scholzj: Well, I didnt said it would be perfect ... but it would be better. I'm not sure if it accepts something like `tools/log-dump/**/*.sh` or soemthing like thate.

systemtest/src/test/java/io/strimzi/systemtest/backup/ColdBackupScriptIsolatedST.java
fvaleri: I think there has been some changes recently. Looking at the new test name, I assume this runs in isolation. Maybe @see-quick can shed a light on this.

tools/cold-backup/README.md
fvaleri: ok

tools/cold-backup/run.sh
fvaleri: ok

systemtest/src/test/java/io/strimzi/systemtest/backup/ColdBackupScriptIsolatedST.java
see-quick: Additionally, if only one test case will be inside `LogDumpScriptIsolatedST` that's okay and we do not need to do such a thing but practically speaking it would not an ideal...

fvaleri: @see-quick this is the error when I use before all. It happens on restore step. 

```
Error from server (AlreadyExists): error when creating "STDIN": persistentvolumeclaims "data-my-cluster-1435185444-zookeeper-0" already exists
error: no matching resources found
```

[strimzi-debug.log](https://github.com/strimzi/strimzi-kafka-operator/files/8117439/strimzi-debug.log)

see-quick: Hmmm, I see...in `@BeforeAll` seems to be missing assigning the default installation to the cluster operator instance so instead of this:
```java
@BeforeAll
    void setUp() {
        clusterOperator.unInstall();
        clusterOperator.defaultInstallation().createInstallation().runInstallation();
    }
```
you should correct it to:
```java
@BeforeAll
    void setUp() {
        clusterOperator.unInstall();
        clusterOperator = clusterOperator.defaultInstallation().createInstallation().runInstallation();
    }

fvaleri: ok, fix pushed

see-quick (approved):

scholzj (approved):

ppatierno (approved):

scholzj:
Thanks for the PR.

PR #6409, opened 2022-02-22
[systemtest] Add documentation for upgrade/downgrade jsons and SPS to upgrade/downgrade tests 

im-konge:

### Type of change

- Enhancement

### Description

This PR adds checks for enabling/disabling SPS feature gate during upgrade/downgrade and also removing version of `0.24.x` from upgrade tests. Some of the checks had to be edited for "supporting" both STS and SPS. 

Also, this PR adds documentation about our upgrade/downgrade JSONs -> to be clear, which fields should be updated if we want to update/add new test scenario.

### Checklist


see-quick (approved):

systemtest/src/test/resources/upgrade/StrimziUpgradeST.json
```diff
@@ -32,42 +32,8 @@
...
-    },
-    "environmentInfo": {
-      "maxK8sVersion": "latest",
-      "status": "stable",
-      "flakyEnvVariable": "none",
-      "reason" : "Test is working on all environment used by QE."
-    },
-    "strimziFeatureGatesFlags": ""
+    "strimziFeatureGatesFlagsBefore": "",
+    "strimziFeatureGatesFlagsAfter": "-ControlPlaneListener"
```
scholzj, line 36: These options are real misterious. Do we have some docs describing it?

systemtest/src/test/resources/upgrade/StrimziUpgradeST.json
```diff
@@ -102,6 +68,7 @@
       "flakyEnvVariable": "none",
       "reason" : "Test is working on all environment used by QE."
     },
-    "strimziFeatureGatesFlags": ""
+    "strimziFeatureGatesFlagsBefore": "-UseStrimziPodSets",
+    "strimziFeatureGatesFlagsAfter": "+UseStrimziPodSets"
```
scholzj, line 72: Same as above.

scholzj (commented):
Stupid question -> how does this ensure we run both variants of the tests?
* Upgrade form STS to STS
* Upgrade form STS to SPS

im-konge:
I meant it this way -> `0.22.1` to `HEAD` is upgrade STS to STS, `0.28.0` to `HEAD` is upgrade STS to SPS. But, I can definitely add there to run both STS to SPS and STS to STS for `0.28.0` version.

`strimziFeatureGatesFlagsBefore` is for setting the `STRIMZI_FEATURE_GATES` env var when we initially deploy the CO (so when deploying CO with version `0.28.0` f.e.) and `strimziFeatureGatesFlagsAfter` is for setting the FG env var when upgrading to `HEAD`.

scholzj:
> I meant it this way -> 0.22.1 to HEAD is upgrade STS to STS, 0.28.0 to HEAD is upgrade STS to SPS. But, I can definitely add there to run both STS to SPS and STS to STS for 0.28.0 version.

I do not think we need STS-to-SPS upgrade for 0.22. But we definitely need to have both for the latest version.

> strimziFeatureGatesFlagsBefore is for setting the STRIMZI_FEATURE_GATES env var when we initially deploy the CO (so when deploying CO with version 0.28.0 f.e.) and strimziFeatureGatesFlagsAfter is for setting the FG env var when upgrading to HEAD.

The question is whether it is documented. Nobody will find it here. JSON sucks because it IMHO doesn't support comments. Since that would be ideal way to explain what it actually does directly in the file.

im-konge:
```
I do not think we need STS-to-SPS upgrade for 0.22. But we definitely need to have both for the latest version.
```
Okay, I will add there the STS-to-STS for the latest version

```
The question is whether it is documented. Nobody will find it here. JSON sucks because it IMHO doesn't support comments. Since that would be ideal way to explain what it actually does directly in the file.
```
Nothing here is really documented IIRC -> we already discussed this offline.  I can add some doc about the upgrade/downgrade jsons. Also, we wanted to refactor the upgrade tests little bit and use `yaml`s instead of `JSON`s.

see-quick (approved):

scholzj (approved):

PR #6411, opened 2022-02-22
[MO] - [system test] -> propage test suite and test case controller l…

see-quick:
…abels into pod template

### Type of change

- Enhancement / new feature
- Refactoring

### Description

This PR adds support for propagation Controllers (i.e., Job, Deployment) labels in our system tests. This is needed for LogCollector when for instance `KafkaClient` object will create producer (i.e., Job object) and for such Job will be spawned associated Pod. However, this Pod would not be collected because these Controllers does not propagate labels and thus we have to do it manually.

### Checklist 

### DIFF
--- a/systemtest/src/main/java/io/strimzi/systemtest/resources/ResourceManager.java
+++ b/systemtest/src/main/java/io/strimzi/systemtest/resources/ResourceManager.java
@@ -13,6 +13,8 @@ import io.fabric8.kubernetes.api.model.PodCondition;
 import io.fabric8.kubernetes.api.model.PodTemplateSpec;
 import io.fabric8.kubernetes.api.model.admissionregistration.v1.ValidatingWebhookConfiguration;
 import io.fabric8.kubernetes.api.model.apiextensions.v1.CustomResourceDefinition;
+import io.fabric8.kubernetes.api.model.apps.Deployment;
+import io.fabric8.kubernetes.api.model.batch.v1.Job;
 import io.fabric8.kubernetes.api.model.rbac.ClusterRole;
 import io.fabric8.kubernetes.api.model.rbac.ClusterRoleBinding;
 import io.fabric8.kubernetes.client.CustomResource;
@@ -61,6 +63,7 @@ import io.strimzi.test.k8s.HelmClient;
 import io.strimzi.test.k8s.KubeClient;
 import io.strimzi.test.k8s.KubeClusterResource;
 import io.strimzi.test.k8s.cmdClient.KubeCmdClient;
+import org.apache.commons.collections.map.UnmodifiableMap;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.junit.jupiter.api.extension.ExtensionContext;
@@ -147,7 +150,7 @@ public class ResourceManager {
     }
 
     @SafeVarargs
-    public final <T extends HasMetadata> void createResource(ExtensionContext testContext, boolean waitReady, T... resources) {
+        public final <T extends HasMetadata> void createResource(ExtensionContext testContext, boolean waitReady, T... resources) {
         for (T resource : resources) {
             ResourceType<T> type = findResourceType(resource);
             LOGGER.info("Create/Update {} {} in namespace {}",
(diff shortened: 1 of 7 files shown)


systemtest/src/main/java/io/strimzi/systemtest/resources/ResourceType.java
```diff
@@ -6,6 +6,7 @@
 
 import io.fabric8.kubernetes.api.model.HasMetadata;
 
+
```
im-konge, line 9: Extra line

im-konge (approved):

systemtest/src/main/java/io/strimzi/systemtest/resources/ResourceManager.java
```diff
@@ -265,6 +266,33 @@ public static HelmClient helmClient() {
...
+     * Auxiliary method for copying Controller {@link Constants.TEST_SUITE_NAME_LABEL} and {@link Constants.TEST_CASE_NAME_LABEL} labels
+     * into PodTemplate ensuring that in case of failure {@link io.strimzi.systemtest.logs.LogCollector} will collect all
+     * related Pods, which corespondents to such Controller (i.e., Job, Deployment)
+     *
+     * @param resource controller resource from which we copy test suite or test case labels
+     * @param resourcePodTemplate {@link PodTemplateSpec} of the specific resource
+     * @param <T> resource, which sings contract with {@link HasMetadata} interface
+     * @param <R> {@link PodTemplateSpec}
+     */
+    public <T extends HasMetadata, R extends PodTemplateSpec> void copyTsOrTcControllerLabelsIntoPodTemplate(T resource, R resourcePodTemplate) {
```
scholzj, line 279: Any less cryptic name? Also, from the code it looks like `TsAndTc` -> not `Or`.

systemtest/src/main/java/io/strimzi/systemtest/resources/ResourceManager.java
```diff
@@ -265,6 +266,33 @@ public static HelmClient helmClient() {
         return resourceReady[0];
     }
 
+    /**
+     * Auxiliary method for copying Controller {@link Constants.TEST_SUITE_NAME_LABEL} and {@link Constants.TEST_CASE_NAME_LABEL} labels
```
scholzj, line 270: What makes these labels _controller_ labels? What is the controller?

systemtest/src/main/java/io/strimzi/systemtest/resources/ResourceManager.java
```diff
@@ -265,6 +266,33 @@ public static HelmClient helmClient() {
...
+     *
+     * @param resource controller resource from which we copy test suite or test case labels
+     * @param resourcePodTemplate {@link PodTemplateSpec} of the specific resource
+     * @param <T> resource, which sings contract with {@link HasMetadata} interface
+     * @param <R> {@link PodTemplateSpec}
+     */
+    public <T extends HasMetadata, R extends PodTemplateSpec> void copyTsOrTcControllerLabelsIntoPodTemplate(T resource, R resourcePodTemplate) {
+        // 1. fetch Controller and PodTemplate labels
+        final Map<String, String> controllerLabels = resource.getMetadata().getLabels();
+        final Map<String, String> podLabels = resourcePodTemplate.getMetadata().getLabels();
```
scholzj, line 282: Is this safe with regards to no labels / unmodifiable map of labels?

scholzj (commented):
TBH, I'm a bit puzzled by this. The place where you create the Deployment / Job seems to know the right labels. So why does it not set them in the PodTemplate in the first place? This seems like a weird hack to me which doesn't fix it at the core.

systemtest/src/main/java/io/strimzi/systemtest/resources/ResourceManager.java
see-quick: I think I could eliminate `Controller` and just enumerate `TEST_SUITE_NAME_LABEL` and `TEST_CASE_NAME_LABEL` labels. The original idea was that Controller is (Job, Deployment etc.) so that's why controller labels.

systemtest/src/main/java/io/strimzi/systemtest/resources/ResourceManager.java
see-quick: Now it is :)

scholzj (approved):

PR #6420, opened 2022-02-23
Try to improve the default memory configuration of the operators

### LINKED ISSUE #4471: Cluster operator does not have sufficient memory: it uses up >98% of memory
**Describe the bug**
The memory is not sufficient for strimzi-cluster-operator after running two weeks. It is even causing alerts!

image

This is the strimzi-cluster-operator's memory usage:

image

Notice it is even using Swap space now:

image

cross-validating the memory usage:

image

**To Reproduce**
1. Run strimzi
2. wait for two weeks
3. see the memory problem

**Expected behavior**
no mem problem

 - Strimzi version: [e.g. master, 0.11.1] 0.19.0
 - Installation method: [e.g. YAML files, Helm chart, OperatorHub.io] yaml
 - Kubernetes cluster: [e.g. Kubernetes 1.14, OpenShift 3.9] latest
 - Infrastructure: [e.g. Amazon EKS, Minikube] cloud

**YAML files and logs**
N/A

**Additional context**
N/A

scholzj:
### Type of change

- Enhancement / new feature

### Description

I spent some time trying to tune the way the operators use the memory to make it run OoM less often. So far my finding is that using the following options seems to significantly reduce the number of restarts: `-XX:MinRAMPercentage=10 -XX:MaxRAMPercentage=20 -XX:InitialRAMPercentage=10`. The `RAMPercentage` options were chosen over `-Xm*` options since they should better scale with the container resources. I'm not 100% sure how efficient this will be since with the default resources (384Mi) a big chunk of the memory is consumed by things such as CodeHeap etc.

This PR adds it to all 3 operators as the new default. It is set only when none of these options is configured by the user. This might seem a bit weird. However, it seemed to be the best way how to combine the existing JAVA_OPTS` values based by the user to the CO with the options from the `jvmOptions` sections in the KAfka CR for TO and UO. If anyone has any more elegant idea how to integrate it, I'm open to suggestions.

One worry is that it is quite hard to test this in various environments. My observation is based on my long-running environment with 1 Kafka cluster, 1 Connect cluster, 6 connectors, 4 users and 32 topics. It is hard to guarantee that these values won't make things worse for some other users. In any case, users can override these values if needed.

This should (hopefully) resolve #4471 

### Checklist

### DIFF
--- a/docker-images/operator/scripts/launch_java.sh
+++ b/docker-images/operator/scripts/launch_java.sh
@@ -23,6 +23,11 @@ JAVA_OPTS="${JAVA_OPTS} $(get_gc_opts)"
 # Deny illegal access option is supported only on Java 9 and higher
 JAVA_OPTS="${JAVA_OPTS} --illegal-access=deny"
 
+# Default memory options used when the user didn't configured any of these options, we set the defaults
+if [[ "$JAVA_OPTS" != *"MinRAMPercentage"* && "$JAVA_OPTS" != *"MaxRAMPercentage"* && "$JAVA_OPTS" != *"InitialRAMPercentage"* ]]; then
+  JAVA_OPTS="${JAVA_OPTS} -XX:MinRAMPercentage=10 -XX:MaxRAMPercentage=20 -XX:InitialRAMPercentage=10"
+fi
+
 # Disable FIPS if needed
 if [ "$FIPS_MODE" = "disabled" ]; then
     JAVA_OPTS="${JAVA_OPTS} -Dcom.redhat.fips=false"


scholzj:
@tombentley @ppatierno It would be great if you can have a look at this and let me know what you think.

ppatierno:
I am not sure how much I can judge these changes. It seems to be promising to work in terms of percentage and not absolute values. The problem is about testing as Jakub already raised. It would be interesting to see what happens when the same operator has to reconcile 30-40 or more Kafka instances?

scholzj:
> I am not sure how much I can judge these changes. It seems to be promising to work in terms of percentage and not absolute values. The problem is about testing as Jakub already raised. It would be interesting to see what happens when the same operator has to reconcile 30-40 or more Kafka instances?

I do not have capacity for such test. In my environment, it now runs 33 days without OOM. But I understand that gives us only a limited picture. That said:
* This still leaves things configurable, so users should be able to override it when needed
* We are in a situation when the operator is running OOM without any significant load, so I'm not sure there is too much this can screw-up

tombentley (approved):
This seems a reasonable quality of life improvement. Looks like it would also apply to `kafak-init` as well as the operators, but that's not a problem.

scholzj:
> This seems a reasonable quality of life improvement. Looks like it would also apply to `kafak-init` as well as the operators, but that's not a problem.

Actually, at least the Topic Operator has the same issue. So it is intentionally applied to all of them.

ppatierno (approved):
I would merge it and see if it's going to be a real improvement for the community by using it.

PR #6430, opened 2022-02-25
[systemtest] Test clients exchange - backup, bridge, connect, cruisecontrol and dump packages

im-konge:
### Type of change

- Enhancement

### Description

This PR is first of few for test clients exchange.

List of changes: 
- add new `ScraperTemplate`, which contains deployment of, lets say, dummy pod. The main purpose of scraper pod is currently to do requests on APIs. But there should be more in the future.
- add `tls` and `scram-sha` methods to our clients.
- add scraper and user names to `TestStorage`
- use `TestStorage` in tests, that contain work with clients 
- add jobs deletion to `ClientUtils.waitForClientSuccess` and `ClientUtils.waitForClients.Success` - so we don't have to delete jobs in each test after completition
- delete `deleteJobWithWait` methods from few tests

### Checklist

### DIFF
--- a/systemtest/src/main/java/io/strimzi/systemtest/Constants.java
+++ b/systemtest/src/main/java/io/strimzi/systemtest/Constants.java
@@ -61,6 +61,14 @@ public interface Constants {
     int GLOBAL_RECONCILIATION_COUNT = (int) ((RECONCILIATION_INTERVAL / GLOBAL_POLL_INTERVAL) + GLOBAL_STABILITY_OFFSET_COUNT);
 
     /**
+     * Scraper pod labels
+     */
+    String SCRAPER_LABEL_KEY = "user-test-app";
+    String SCRAPER_LABEL_VALUE = "scraper";
+
+    String SCRAPER_NAME = "scraper";
+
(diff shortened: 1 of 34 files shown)


systemtest/src/test/resources/junit-platform.properties
```diff
@@ -1,9 +1,9 @@
 # These are default values if @ParallelTest or @ParallelSuite is not specified
 # junit.jupiter.execution.parallel.mode.default <==> ParallelTest = concurrent
 # junit.jupiter.execution.parallel.mode.classes.default <==> ParallelSuite = concurrent
-junit.jupiter.execution.parallel.enabled=false
+junit.jupiter.execution.parallel.enabled=true
```
see-quick, line 4: By the way you can set up these options via InteliJ in VM options (configuration).

see-quick (approved):
LGTM, until Azure is green (there are a few errors related to parameter handling of KafkaClients) so it should be easy to fix.

Frawless (approved):
Assuming it will work after rebase

PR #6443, opened 2022-02-28
Filter forbidden options in KafkaConnector configurations

### LINKED ISSUE #6438: Connector gets restarted on every reconciliation when tasks.max is set
**Describe the bug**

When setting the `tasks.max` configuration key under `config` the operator issues a warning ("Configuration parameter tasks.max in KafkaConnector.spec.config will be ignored") but still seems to update the config in a way 
that makes the connector restart.

Every 2 minutes, connectors that are configured with `config.action.reload = restart` (which I believe is the default) will get restarted.

**To Reproduce**
1. Add `tasks.max: 10` in the `config` of any `KafkaConnector`
2. See that it restarts on reconciliation (both on change trigger and timer)

**Expected behavior**
Connectors do not restart if `tasks.max` is set by mistake.

 - Strimzi version: 0.27.0
 - Kubernetes cluster: Kubernetes 1.21
 - Infrastructure: Amazon EKS

scholzj:
### Type of change

- Bugfix

### Description

The connector operator currently does not properly handle the forbidden properties:
* They are defined in the API class, but never used to filter the options, just for docs
* The forbidden options are actually filtered / overwritten before aplying the configuration to Connect, but it has its own definition of the forbidden fields
* The forbidden options are not properly filtered when the operator decides if connector configuration needs to be updated or not. This causes unnecessary connector updates and connector reloades when the forbidden options are used as described in #6438 

One of the problems here is that the Connector operator does not use the shared Configuration classes used everywhere else. This PR uses it for the connectors as well. That ensures the forbidden options are filtered in the same way as for the other configurations. The configuration class also does the warnings of forbidden configuration options automatically.

This should close #6438 

### Checklist

### DIFF
--- a/api/src/main/java/io/strimzi/api/kafka/model/AbstractConnectorSpec.java
+++ b/api/src/main/java/io/strimzi/api/kafka/model/AbstractConnectorSpec.java
@@ -27,7 +27,8 @@ import java.util.Map;
 @EqualsAndHashCode
 public abstract class AbstractConnectorSpec extends Spec {
     private static final long serialVersionUID = 1L;
-    private static final String FORBIDDEN_PARAMETERS = "connector.class, tasks.max";
+
+    public static final String FORBIDDEN_PARAMETERS = "connector.class, tasks.max";
 
     private Integer tasksMax;
     private Boolean pause;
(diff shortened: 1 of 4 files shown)


scholzj:
(The CI failure is some Azure caching problem not related to this PR)

ppatierno (approved):

PR #6458, opened 2022-03-03
[MO] - [system test] -> [minor] bump OLM version

see-quick:

### Type of change

- Enhancement / new feature

### Description

This PR just actualize the current OLM version for our system tests

### Checklist

### DIFF
--- a/systemtest/src/main/java/io/strimzi/systemtest/Constants.java
+++ b/systemtest/src/main/java/io/strimzi/systemtest/Constants.java
@@ -367,7 +367,6 @@ public interface Constants {
      */
     String NAMESPACE_KEY = "NAMESPACE_NAME";
     String PREPARE_OPERATOR_ENV_KEY = "PREPARE_OPERATOR_ENV";
-    String PARALLEL_CLASS_COUNT = "PARALLEL_CLASS_COUNT";
 
     /**
      * Auxiliary variable for cluster operator deployment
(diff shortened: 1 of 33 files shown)


im-konge:
I was thinking about bumping the OLM version too, the thing is that I would like to have this automated (not for this PR I guess) -> for example, add it to `make` commands after release -> when we bumping our main branch to `x.y.z-SNAPSHOT` version, we can bump this too? WDYT @scholzj @see-quick ?

scholzj:
> I was thinking about bumping the OLM version too, the thing is that I would like to have this automated (not for this PR I guess) -> for example, add it to make commands after release -> when we bumping our main branch to x.y.z-SNAPSHOT version, we can bump this too? WDYT @scholzj @see-quick ?

That sounds too complicated. I think we first need to understand what it is useful for. I don't understand why do you have such version there. Is it for upgrade tests?

im-konge:
AFAIR we didn't build metadata image for each push to the main brach and we didn't have any `latest` image in OLM, maybe it's different now. We always had images for latest released version - f.e. 0.27.1, 0.28.0 and so on. If we have latest image pushed from the main branch, we can change it to `latest` tag.

scholzj:
Well, first of all, we have the `stable` channel in OLM. Doesn't that help to avoid using the versions? Can't you just get latest from stable channel installed? You do not need the version in the OLM Subscription resource when you have the channel.

Second, non of it explains how is this useful. Why does main branch care about 0.28.0 release of Strimzi? The only thing where it might matter are upgrade tests but I'm not aware of any OLM tests there.

im-konge:
> Well, first of all, we have the stable channel in OLM. Doesn't that help to avoid using the versions? Can't you just get latest from stable channel installed? You do not need the version in the OLM Subscription resource when you have the channel.

Yeah well, IIRC I tried that in the past and it worked, so we can change it.

> Second, non of it explains how is this useful. Why does main branch care about 0.28.0 release of Strimzi? The only thing where it might matter are upgrade tests but I'm not aware of any OLM tests there.

It seems that we are not using it during upgrade. In this case I think it's useless to have it here.

Frawless:
I agree that as a default value we can keep it empty in the subscription.

see-quick:
> I agree that as a default value we can keep it empty in the subscription.

Closing because it was done with https://github.com/strimzi/strimzi-kafka-operator/pull/6468

PR #6462, opened 2022-03-03
aws-msk-iam-auth support for standalone topic operator

eapri:
### Type of change

- Enhancement / new feature

### Description

Add aws-msk-iam-auth configuration support for standalone strimzi topic operator.

- Add aws-msk-iam-auth dependency to POM.XML (https://github.com/aws/aws-msk-iam-auth)
- Add small amount of code changes

### Checklist

- [ ] Write tests
- [X] Update documentation
- [ ] Update CHANGELOG.md


scholzj (commented):
Thanks for the PR.

As discussed multiple times in the past, I'm not sure we want to bundle any authenticators specific to a single provider like this directly in our images and APIs. It does not scale.

So I think that this should be written to  allow _custom_ authentication so that everyone would be able to use it with their own authentication - add their desired JAR(s) to the container images and configure the authenticator in some generic way. Probably in the form of a proposal: https://github.com/strimzi/proposals

Even if we decide to accept this, we would need:
* Integration tests so that we can ensure it works
* Some guarantees that the library will support latest Kafka versions so that we can make sure it does not block us in the future

Issue #6470, opened 2022-03-04
Cold-backup script does not work on MacOS


scholzj:
The cold-backup tool does not work on MacOS. It fails with following error:

```
tar: Option --no-check-device is not supported
Usage:
  List:    tar -tf <archive-filename>
  Extract: tar -xf <archive-filename>
  Create:  tar -cf <archive-filename> [filenames...]
  Help:    tar --help
```

This is probably because of code like this:

```
    kubectl -n "$NAMESPACE" exec -i "$RSYNC_POD_NAME" -- bash -c "tar $flags -C /data -c ." \
      | bash -c "tar $flags -C $target -xv -f - && if [[ $? == 1 ]]; then exit 0; fi"
```

I suppose that the aliases set at the start of the script are no used when a new bash is started.

scholzj:
CC @fvaleri

fvaleri:
OK, I don't have a MacOS machine to test it anymore. I added that subshell commands to make shellcheck happy. Without `bash -c "..."` should work, but let's try to add the export command just before that row:

```    
    export flags target
    kubectl -n "$NAMESPACE" exec -i "$RSYNC_POD_NAME" -- bash -c "tar $flags -C /data -c ." \
      | bash -c "tar $flags -C $target -xv -f - && if [[ $? == 1 ]]; then exit 0; fi"
```

scholzj:
The `flags` and `target` are fine I think because in the error the `--no-check-device` got passes to the `tar` executable`. But the aliases mapping `tar` to `gtar` IMHO don't work because it is calling the original `tar` and not `gtar`.

fvaleri:
I think you are right. Aliases are not inherited in sub processes.

PR #6473, opened 2022-03-04
Fix tar alias expansion on Mac

fvaleri:
Fix for #6470. There may be other ways, like exporting a function instead of alias, but they are very ugly and hard to read.


scholzj (approved):
This seems to work fine. Thanks.

ppatierno (approved):
Don't have a Mac to test but I trust you both with Mac ;-)

PR #6457, opened 2022-03-02
Update Jackson to include CVE fixed in version 2.12.6

pratimsc:
### Type of change

Bugfix - Security Vulnerability: Strimzi Operator

- Bugfix

### Description

This fix includes changes to mitigate CVE reported at - https://github.com/FasterXML/jackson-databind/issues/3328

### Checklist

_Please go through this checklist and make sure all applicable tasks have been done_

- [x] Write tests
- [x] Make sure all tests pass
- [x] Update documentation
- [ ] Check RBAC rights for Kubernetes / OpenShift roles
- [ ] Try your changes from Pod inside your Kubernetes and OpenShift cluster, not just locally
- [ ] Reference relevant issue(s) and close them after merging
- [ ] Update CHANGELOG.md
- [ ] Supply screenshots for visual changes, such as Grafana dashboards

PR #6474, opened 2022-03-04
Update Jackson to include CVE fixed in version 2.12.6 for 3rd party artifacts

pratimsc:
Signed-off-by: Pratim S Chaudhuri <pratim.sunilkumar.chaudhuri@mercer.com>

### Type of change

- Bugfix

### Description

This is linked to : https://github.com/strimzi/strimzi-kafka-operator/pull/6457

### Checklist

- [ ] Write tests
- [ ] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/docker-images/artifacts/kafka-thirdparty-libs/3.1.x/pom.xml
+++ b/docker-images/artifacts/kafka-thirdparty-libs/3.1.x/pom.xml
@@ -339,12 +339,12 @@
         <dependency>
             <groupId>com.fasterxml.jackson.dataformat</groupId>
             <artifactId>jackson-dataformat-yaml</artifactId>
-            <version>2.12.3</version>
+            <version>2.12.6</version>
         </dependency>
         <dependency>
             <groupId>com.fasterxml.jackson.datatype</groupId>
             <artifactId>jackson-datatype-jsr310</artifactId>
-            <version>2.12.3</version>
+            <version>2.12.6</version>
         </dependency>
         <!-- END: To avoid conflicts between Jackson 2.12 used by Kafka and 2.13 used by Fabric8, we have to exclude these libraries and include the right versions -->
         <!-- EnvVar Configuration Provider for Apache Kafka -->


docker-images/artifacts/kafka-thirdparty-libs/3.1.x/pom.xml
```diff
@@ -339,12 +339,12 @@
...
             <artifactId>jackson-dataformat-yaml</artifactId>
-            <version>2.12.3</version>
+            <version>2.12.6</version>
         </dependency>
         <dependency>
             <groupId>com.fasterxml.jackson.datatype</groupId>
             <artifactId>jackson-datatype-jsr310</artifactId>
-            <version>2.12.3</version>
+            <version>2.12.6</version>
         </dependency>
```
scholzj, line 348: Isn't Kafka 3.1 using 2.12.3? This should stick with the Kafka version to keep things in sync.

pratimsc: Kafka has already applied 2.12.6 : 
Kafka 3.0: https://github.com/apache/kafka/commit/ac91fecbd2c1fad7a80a9e9845aae10621756fd7
Kafka 3.1: https://github.com/apache/kafka/commit/f514b9e44649b56e0d436dba9f332b5e7569f93f

Also please refer the ticket at https://issues.apache.org/jira/browse/KAFKA-13658

One option could be to wait until 3.0.1 and 3.1.1 are released, and then include them. 

What is the principle/rules/relation between Strimzi release and Kafka release cycle?

scholzj: Yeah, but these commits mean nothing. They are not in the 3.1.0 or 3.0.0 releases which Strimzi uses.

pratimsc: I will change this PR to only have changes related to Cruise Control, so that it by default include Kafka 3.x versions. And remove the explicit Jackson version upgrades.

What would be the process when version 3.0.1 and 3.1.1 of Kafka are released? Will they be included as part of 0.28.1 version of Strimzi, or version 0.29.0?

scholzj: > I will change this PR to only have changes related to Cruise Control, so that it by default includes Kafka 3.x versions. And remove the explicit Jackson version upgrades.

Please do not modify the dependencies randomly just to fix some report. The software is somehow released and tested. The changes we do to it should be only very limited.

scholzj: > What would be the process when version 3.0.1 and 3.1.1 of Kafka are released? Will they be included as part of 0.28.1 version of Strimzi, or version 0.29.0?

They need to be released first. Then they can be included in the next release.

pratimsc:
Putting a pause to this, and tie it to release of Kafka 3.0.1 and 3.1.1.

Kafka will be updated 3.0.1 as part of https://github.com/strimzi/strimzi-kafka-operator/pull/6482

scholzj:
Ok, than lets close it for the time being and we can discuss it again later.

Issue #6477, opened 2022-03-04
Cold backup tool fails to handle different `-t` parameters


scholzj:
The Cold backup tool seems to have problems handling different values of the `-t` parameter to specify the target file. For example:

1) When calling the tool like this:
    ```
    tools/cold-backup/run.sh backup -n myproject -c my-cluster -t xxx.zip -y
    ```
    With just `-t xxx.zip`, it fails with the following error:
    ```
    ...
    ./kafka-log2/.kafka_cleanshutdown
    pod "cold-backup" deleted
    Compressing ./myproject/my-cluster to xxx.zip
    mv: xxx.zip and ./xxx.zip are identical                                <------- ERROR
    Starting cluster my-cluster
    statefulset.apps/my-cluster-zookeeper scaled
    ...
    ```

2) When I pass a path to subdirectory like this:
    ```
    tools/cold-backup/run.sh backup -n myproject -c my-cluster -t xxx/xxx.zip -y
    ```
    With `-t xxx/xxx.zip`, it seems to succeed:
    ```
    ./kafka-log2/.kafka_cleanshutdown
    pod "cold-backup" deleted
    Compressing xxx/myproject/my-cluster to xxx/xxx.zip
    Starting cluster my-cluster
    statefulset.apps/my-cluster-zookeeper scaled
    ```
    But the ZIP file does not seem to be created and the work files are still there:
    ```
    $ tree xxx
    xxx
    └── myproject
        └── my-cluster
            ├── data
            │   ├── ...
            ├── log-dump-env
            ├── resources
            │   ├── ...
            └── xxx
    
    178 directories, 863 files
    ```
    The `xxx/myproject/my-cluster/xxx` file seems to be the ZIP file. But it is in a wrong place and has no extension.

This is from MacOS - no idea if the same issue is on Linux or if this is MacOS only issue.

scholzj:
CC @fvaleri

Issue #6476, opened 2022-03-04
Cold back tool should make it more clear if it succeeded or not


scholzj:
When the Cold backup tool fails, it always restores the cluster into the original state. That is correct. But it also makes it quite hard to understand if the backup succeeded or failed. You often find out only by checking the exit code. There is no message about it succeeding or failing. And it usually ends with _positive_ messages about the pod successfully running again. So it is quite confusing.

For example with the output like this:
```
./kafka-log0/__consumer_offsets-32/leader-epoch-checkpoint
./kafka-log0/__consumer_offsets-32/00000000000000000000.log
./kafka-log0/__consumer_offsets-32/00000000000000000000.index
./kafka-log0/__consumer_offsets-32/00000000000000000000.timeindex
./kafka-log0/__consumer_offsets-32/partition.metadata
./kafka-log0/__strimzi-topic-operator-kstreams-topic-store-changelog-0/
./kafka-log0/__strimzi-topic-operator-kstreams-topic-store-changelog-0/leader-epoch-checkpoint
./kafka-log0/__strimzi-topic-operator-kstreams-topic-store-changelog-0/00000000000000000000.log
./kafka-log0/__strimzi-topic-operator-kstreams-topic-store-changelog-0/00000000000000000000.index
./kafka-log0/__strimzi-topic-operator-kstreams-topic-store-changelog-0/00000000000000000000.timeindex
command terminated with exit code 1
./kafka-log0/__strimzi-topic-operator-kstreams-topic-store-changelog-0/partition.metadata
./kafka-log0/__strimzi-topic-operator-kstreams-topic-store-changelog-0/00000000000000000003.snapshot
pod "cold-backup" deleted
Starting cluster my-cluster
deployment.apps/my-cluster-entity-operator scaled
Waiting for "condition=Ready" on "pod -l strimzi.io/name=my-cluster-entity-operator" in namespace "myproject"
pod/my-cluster-entity-operator-687ffd85f6-4z475 condition met
deployment.apps/my-cluster-kafka-exporter scaled
Waiting for "condition=Ready" on "pod -l strimzi.io/name=my-cluster-kafka-exporter" in namespace "myproject"
pod/my-cluster-kafka-exporter-577bdbb77-lw8qp condition met
deployment.apps/strimzi-cluster-operator scaled
Waiting for "condition=Ready" on "pod -l strimzi.io/kind=cluster-operator" in namespace "myproject"
[...]

Issue #6471, opened 2022-03-04
Cold-backup script does nto work when StrimziPodSets are used


scholzj:
The Cold backup tool does not support the StrimziPodSets. When the StatefulSets don't exist,it tries to do the backup without scaling down the pods. But this does not work because it does not et a clear access to the storage while the brokers / ZooKeepers are running. It will either never mount the volume (with RWO volumes) or it will mount it (with RWM volumes), but usually fail during backup when any of the files change.

scholzj:
CC @fvaleri

scholzj:
The PR #6472 disabled running the Cold backup STs with StrimziPodSets. It should be re-enabled when this is fixed.

fvaleri:
Thanks for raising this @scholzj. I'll look into it.

PR #6481, opened 2022-03-05
Add support for PodSets and other improvements

fvaleri:
Fixes and improvements related to #6476, #6477, #6471.

### DIFF
--- a/tools/cold-backup/README.md
+++ b/tools/cold-backup/README.md
@@ -49,7 +49,7 @@ kubectl run producer-perf -it --image="$CLIENT_IMAGE" --rm="true" --restart="Nev
     --throughput -1 --producer-props acks=1 bootstrap.servers="my-cluster-kafka-bootstrap:9092"
 
 # run backup procedure
-./tools/cold-backup/run.sh backup -n myproject -c my-cluster -t /tmp/my-cluster.zip
+./tools/cold-backup/run.sh backup -n myproject -c my-cluster -t /tmp/my-cluster.tgz
 
 # recreate the namespace and deploy the operator
 kubectl delete ns myproject
@@ -57,7 +57,7 @@ kubectl create ns myproject
 kubectl create -f ./install/cluster-operator
 
 # run restore procedure and wait for provisioning
-./tools/cold-backup/run.sh restore -n myproject -c my-cluster -s /tmp/my-cluster.zip
+./tools/cold-backup/run.sh restore -n myproject -c my-cluster -s /tmp/my-cluster.tgz
 
 # check consumer group offsets (expected: current-offset match)
 kubectl run consumer-groups -it --image="$CLIENT_IMAGE" --rm="true" --restart="Never" -- \
--- a/tools/cold-backup/README.md
+++ b/tools/cold-backup/README.md
@@ -1,12 +1,18 @@
 ## Strimzi backup
 
-Bash script for offline (_cold_) backups of Kafka clusters on Kubernetes or OpenShift. The script only supports a local file system. Make sure you have enough free disk space in the target directory.
+Bash script for offline (_cold_) backups of Kafka clusters on Kubernetes or OpenShift. 
+The script only supports a local file system, and you need to make sure to have enough free disk space in the target directory.
 
-If you think you don't need a backup strategy for Kafka because of its embedded data replication, then consider the impact of a misconfiguration, bug, or security breach that deletes all of your data. For online (_hot_) backups, you can use storage snapshotting or stream into object storage.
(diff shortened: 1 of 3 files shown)


scholzj (commented):
Can you please rebase it? It seems ot have the issue with the `tar`alias.

scholzj:
It looks like after the backup, the collected files are tared, but the temporary directory structure and its data remains (in my case `myproject/my-cluster` - `namespace/cluster-name`). Is that intentional? I guess it should be deleted, or?

fvaleri:
> It looks like after the backup, the collected files are tared, but the temporary directory structure and its data remains (in my case `myproject/my-cluster` - `namespace/cluster-name`). Is that intentional? I guess it should be deleted, or?

It is like this from the start. There is no `rm` command in the entire script. I would prefer the user to explicitly delete that staging folder before running a new backup. Initially, I also thought about supporting incremental backups with `rsync`, but this does not work well in practice. A possible improvements my be to use a more advanced tool such as `rclone`, but this is outside the scope of this PR.

scholzj:
> It is like this from the start. There is no rm command in the entire script. I would prefer the user to explicitly delete that staging folder before running a new backup.

Ok, I din't remembered it from before, so I was wondering if it is some bug. But if it is intentional its fine.

scholzj:
Some more comments ... I observed them when testing with StrimziPodSets -> not sure if they apply to StatefulSets
* It does not wait for the pods to be terminated. As a result, it tries to delete the Kafka pods first, but because if does not wait, it deletes Zoo right after. That Zoo shutdowns faster than Kafka and leaves the Kafka pod in unclean shutdown. which is bad and should be avoided (I can open a new issue for it if you would want to deal with it separately).
* The cluster does not seem to recover after the backup. It tries to first recover the EO. But that will not work before the Zoo/Kafka pods are ready. And they will not be ready unless you enable the CO first. But without them, the EO never gets ready and the whole backup fails with this:
   ```
   error: timed out waiting for the condition on pods/my-cluster-entity-operator-8554c4bf6c-rxtn9
   backup failed at 144: $con_cmd
   ```
   So I think the right approach here would be to first enable the CO and let the CO recover the cluster. That should work both  with StatefulSets as well as without StatefulSets.
* On one occasion, I got following error:
   ```
    ...
    ./kafka-log2/__strimzi-topic-operator-kstreams-topic-store-changelog-0/00000000000000000000.index
    ./kafka-log2/__strimzi-topic-operator-kstreams-topic-store-changelog-0/00000000000000000000.timeindex
    gtar: Unexpected EOF in archive
    gtar: Unexpected EOF in archive
    gtar: Error is not recoverable: exiting now
    pod "cold-backup" deleted
    Compressing xxx/myproject/my-cluster to xxx/xxx.tgz
    ...
   ```
   Not sure what caused this -> raising it just in case you saw it in the past.

The changes related to the other 2 issues seem to work fine here.

fvaleri:
I agree on the first two points, and I'll commit required changes in this PR. I haven't seen any tar error since first experiments. Feel free to open a separate issue in case you can reproduce it consistently.

scholzj (approved):

scholzj:
> I agree on the first two points, and I'll commit the required changes in this PR. I haven't seen any tar error since first experiments. Feel free to open a separate issue in case you can reproduce it consistently.

I saw it just once so far. Maybe it was some network glitch between my server and local machine running the tool or something.

scholzj:
@fvaleri I forgot about it yesterday -> but since it should now work with StrimziPodSets, can you please also remove the `@StatefulSetTest` annotation from the System Tests? So that I can run them?

fvaleri:
Sure, and thanks for taking the time to review this.

tools/cold-backup/README.md
```diff
@@ -49,15 +55,15 @@ kubectl run producer-perf -it --image="$CLIENT_IMAGE" --rm="true" --restart="Nev
     --throughput -1 --producer-props acks=1 bootstrap.servers="my-cluster-kafka-bootstrap:9092"
 
 # run backup procedure
-./tools/cold-backup/run.sh backup -n myproject -c my-cluster -t /tmp/my-cluster.zip
+./tools/cold-backup/run.sh backup -n myproject -c my-cluster -t /tmp/my-cluster.tgz
```
ppatierno, line 58: by changing to tgz I would say that requirements in the README could be changed by removing the zip and unzip as needed?

ppatierno (approved):
Just one nit but LGTM.

tools/cold-backup/README.md
fvaleri: Right, thanks for spotting this.

scholzj:
Thanks for the PR @fvaleri

PR #6482, opened 2022-03-05
Add support for Kafka 3.0.1

scholzj:
### Type of change

- Enhancement / new feature

### Description

This PR adds support for Kafka 3.0.1. To avoid dependency conflict, I also had to update the 3rd party libs to exclude some more Jackson libraries.

_(This PR is currently draft based on the RC)_

### Checklist

- [x] Update CHANGELOG.md

### DIFF
--- a/docker-images/artifacts/kafka-thirdparty-libs/3.0.x/pom.xml
+++ b/docker-images/artifacts/kafka-thirdparty-libs/3.0.x/pom.xml
@@ -340,11 +340,35 @@
             <groupId>com.fasterxml.jackson.dataformat</groupId>
             <artifactId>jackson-dataformat-yaml</artifactId>
             <version>2.12.3</version>
+            <exclusions>
+                <exclusion>
+                    <groupId>com.fasterxml.jackson.core</groupId>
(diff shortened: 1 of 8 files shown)


docker-images/artifacts/kafka-thirdparty-libs/3.0.x/pom.xml
```diff
@@ -340,11 +340,35 @@
             <groupId>com.fasterxml.jackson.dataformat</groupId>
             <artifactId>jackson-dataformat-yaml</artifactId>
             <version>2.12.3</version>
```
pratimsc, line 342: @scholzj  - I think this should be updated to `2.12.6` as the version 3.0.1 uses `2.12.6` version. 

Ref: https://cwiki.apache.org/confluence/display/KAFKA/Release+Plan+3.0.1

scholzj: We still support 3.0.0. So it is not so easy.

scholzj:
@ppatierno This should be now ready-for-review. Thanks.

ppatierno (approved):

PR #6487, opened 2022-03-08
Per-broker configuration with Kafka PodSets

### LINKED ISSUE #1211: Do we really need to roll the old Kafka pods when scaling up and using loadbalancers and nodeports?
Do we really need to roll the old Kafka pods when scaling up and using loadbalancers and nodeports?

### LINKED ISSUE #4913: Avoid rolling updates for scale down and scale up operations
It would be nice if it was possible to scale the cluster up and down without also having to endure a rolling upgrade. Rolling upgrades are not the end of the world, but they are disruptive to clients and causes latency spikes(loses jit, caches are flushed, etc).  

Scaling the cluster does not trigger a rolling upgrade 

**Additional context**
Some situations where this behavior can be dangerous:
- A cluster that needs a "urgent" scale up because it's overloaded. An unnecessary rolling upgrade can aggravate the situation 
- Big clusters with many partitions. Cleanly shutting down kafka may take some time when there are many partitions, multiply this time by the number of brokers and the rolling operation will be really long. This means that the disruption of the rolling restart will also persist for some time.

It seems that the root cause of the rolling is caused by the change in `advertised-hostnames.config`. The brokers are added or removed from this entry which changes the hash in `strimzi.io/broker-configuration-hash`. 

In https://github.com/strimzi/strimzi-kafka-operator/issues/4090 it was mentioned that there's "long story" regarding the motivation behind this. Can you guys give more context on why the brokers need to be statically added to that config? Couldn't `advertised.listeneners` be configured on a per broker basis using a dynamic script with something like `hostname -A` ?

scholzj:
### Type of change

- Enhancement / new feature

### Description

Currently, we have a single configuration generated for all Kafka brokers. It is the same for all of them with the exception of some replaceable values which are replaced using `envsubst` only in the running broker container. This does not allow us to configure each broker separately to for example give it a separate role (e.g. Kraft controller versus broker etc.). It also means that all brokers need to be rolled  when scaling up / down because some parts of the config change. It will be also limiting once we possibly have different node pools with different Kafka configurations.

This PR changes how the Kafka brokers are configured when using StrimziPodSets. It uses a new per-broker configuration file from its own ConfigMap. It also reduces the way `envsubst` are used in the containers. It now passes only secrets (e.g. OAuth client secrets) and things not known up-front (Rack-id, Node port hostnames) as placeholders. All other information is already filled in in the config file including broker ID or advertised hostnames (with the exception of node port hostnames) and ports.

Long term, this should allow us to have different configuration per-broker which is important for example for Kraft. But already this PR provides some improvements such as reduced number of rolled pods when scaling Kafka up/down or listener address changes for some brokers. (In these situations for StatefulSet we always have to roll all brokers)

When StatefulSets are used, everything works as before. Unlike StrimziPodSets, StatefulSets use the same ConfigMap for all brokers. If we would generate into it different records with per-broker configuration, we would face risks that for bigger clusters, we run out of the available space in the ConfigMap. So I decided to not attempt anything like that. With StrimziPodSets, each broker has its own ConfigMap so this risk is not an issue.

Changes to some of the STs about rolling update during scale-up/down are too complex to change in this PR. They are disabled for now for StrimziPodSets and will be fixed as part of #6493.

This resolves #1211 and resolves #4913.

### Checklist

- [x] Write tests
- [x] Update documentation

### DIFF
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaBrokerConfigurationBuilder.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaBrokerConfigurationBuilder.java
@@ -14,16 +14,16 @@ import io.strimzi.api.kafka.model.KafkaAuthorizationOpa;
 import io.strimzi.api.kafka.model.KafkaAuthorizationSimple;
 import io.strimzi.api.kafka.model.KafkaResources;
 import io.strimzi.api.kafka.model.Rack;
-import io.strimzi.api.kafka.model.listener.KafkaListenerAuthenticationCustom;
-import io.strimzi.api.kafka.model.listener.arraylistener.GenericKafkaListener;
 import io.strimzi.api.kafka.model.listener.KafkaListenerAuthentication;
+import io.strimzi.api.kafka.model.listener.KafkaListenerAuthenticationCustom;
 import io.strimzi.api.kafka.model.listener.KafkaListenerAuthenticationOAuth;
 import io.strimzi.api.kafka.model.listener.KafkaListenerAuthenticationScramSha512;
 import io.strimzi.api.kafka.model.listener.KafkaListenerAuthenticationTls;
+import io.strimzi.api.kafka.model.listener.arraylistener.GenericKafkaListener;
 import io.strimzi.api.kafka.model.listener.arraylistener.GenericKafkaListenerConfiguration;
 import io.strimzi.kafka.oauth.server.ServerConfig;
-import io.strimzi.operator.cluster.operator.resource.cruisecontrol.CruiseControlConfigurationParameters;
 import io.strimzi.kafka.oauth.server.plain.ServerPlainConfig;
+import io.strimzi.operator.cluster.operator.resource.cruisecontrol.CruiseControlConfigurationParameters;
 import io.strimzi.operator.common.Reconciliation;
 
 import java.io.PrintWriter;
@@ -32,6 +32,8 @@ import java.time.Duration;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Locale;
+import java.util.function.Function;
+import java.util.function.Supplier;
 import java.util.stream.Collectors;
 
 /**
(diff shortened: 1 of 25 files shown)


im-konge (approved):

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaBrokerConfigurationBuilder.java
```diff
@@ -41,10 +43,34 @@
...
+    // * ${STRIMZI_NODEPORT_EXTERNALDNS_ADDRESS}
+    // * ${STRIMZI_NODEPORT_INTERNALIP_ADDRESS}
+    // * ${STRIMZI_NODEPORT_INTERNALDNS_ADDRESS}
+    // * ${STRIMZI_NODEPORT_HOSTNAME_ADDRESS}
+
     private final StringWriter stringWriter = new StringWriter();
     private final PrintWriter writer = new PrintWriter(stringWriter);
     private final Reconciliation reconciliation;
 
+    private String brokerId = PLACEHOLDER_BROKER_ID;
```
ppatierno, line 72: Why are you assigning the placeholder constant to a variable and using it through the code instead of using the placeholder itself as done with the others?

ppatierno (approved):
LGTM. Just a nit.

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaBrokerConfigurationBuilder.java
scholzj: Depending on whether you generate per-broker configuration or shared configuration, the `brokerId` value will be either the placeholder or the actual broker ID. If I wanted to use the placeholder constant directly, the code would be full of something like `if (brokerId == null) use-placeholder else use-actual-id`. This makes it easier -> I just set the `brokerId` either to the placeholder or to the actual broker ID and use the `brokerId` variable everywhere.

ppatierno: Yeah I missed that. Thanks.

PR #6516, opened 2022-03-12
Use only one hash function everywhere

scholzj:
### Type of change

- Refactoring

### Description

We use currently several utility methods to create hashes or hash stubs to track changes to different configurations and trigger rolling updates. This PR unifies them to use all the same function instead of multiple different ones.

### Checklist


fvaleri:
This is great. Given that we are touching this, shouldn't we use a stronger hash function such as sha256? I know this is not used for security, but sha1 hash collisions have been demonstrated. Performance impact shouldn't be noticeable.

scholzj:
@fvaleri I think SHA-1 is fine for this for the time being. We anyway take only the first 8 characters,

ppatierno (approved):

fvaleri:
ok +1 for me :)

PR #6518, opened 2022-03-13
Clean-up of the `KafkaCluster` class and its methods

scholzj:
### Type of change

- Refactoring

### Description

This PR tries to clean-up little bit the `KafkaCluster` class. It:
* Removes the static methods which only refer to another static methods. This should make the class smaller and code more readable overall with less methods whcih just redirect to other methods.
* Remove the getters / setters which are used only internally => in most cases we internally use the class fields directly. Only in few cases we did it through setters without any aditional logic or through getters.

There might be more stuff to clean here in the future, but I do not want to make the PRs too big and too complicated to review.

### Checklist

### DIFF
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaCluster.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaCluster.java
@@ -116,8 +116,8 @@ public class KafkaCluster extends AbstractModel {
     protected static final String CONTROLPLANE_PORT_NAME = "tcp-ctrlplane"; // port name is up to 15 characters
 
     // Ingress and Route listeners advertise port 443 regardless what port is used in Kafka, so we store them here
-    protected static final int ROUTE_PORT = 443;
-    protected static final int INGRESS_PORT = 443;
+    public static final int ROUTE_PORT = 443;
+    public static final int INGRESS_PORT = 443;
 
     protected static final String KAFKA_NAME = "kafka";
     protected static final String CLUSTER_CA_CERTS_VOLUME = "cluster-ca";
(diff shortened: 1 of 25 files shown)


cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaCluster.java
```diff
@@ -241,10 +241,10 @@
      */
     private KafkaCluster(Reconciliation reconciliation, HasMetadata resource) {
         super(reconciliation, resource, APPLICATION_NAME);
-        this.name = kafkaClusterName(cluster);
-        this.serviceName = serviceName(cluster);
-        this.headlessServiceName = headlessServiceName(cluster);
-        this.ancillaryConfigMapName = metricAndLogConfigsName(cluster);
+        this.name = KafkaResources.kafkaStatefulSetName(cluster);
```
ppatierno, line 244: This `name` IIRC is used for the `StrimziPodSet` as well (of course for the `StatefulSet` now). Should we try to change the method name `kafkaStatefulSetName` with something more generic (which apply on both "sets")? It's actually the `kafkaClusterName` or? Maybe we can reuse this method name in the `KafkaResources` class as you removed the method in this class.

ppatierno (commented):
Thanks for this Jakub. I was tempted a lot of times doing this to avoid usage of useless getters/setters. It LGTM but I left one comment.

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaCluster.java
scholzj: I think this is complicated - it is used to name many things (although I actually expect the StrimziPodSets to use this only temporarily and not long term). However, there is nothing like a `kafkaClusterName` in reality. The _Kafka cluster_ is pretty much the whole Kafka CR and this is not its name. So that name would be quite confusing.

So I guess the question is whether we have any better name ... and whether it really belongs to this PR since that is really part of the `KafkaResources` class and not `Kafkacluster` class anyway.

Frawless (approved):

ppatierno (approved):

PR #6528, opened 2022-03-16
docs: describe tasksmax for mirrormaker

PaulRMellor:

**Documentation**
Adds a new section to [Kafka MirrorMaker 2.0 cluster configuration](https://strimzi.io/docs/operators/in-development/configuring.html#assembly-mirrormaker-str) in the _configuring_ guide. The section provides some context to how and when you use the `tasksMax` property.

### Checklist

- [ ] Write tests
- [x] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/documentation/modules/configuring/con-config-mirrormaker2-tasks-max.adoc
+++ b/documentation/modules/configuring/con-config-mirrormaker2-tasks-max.adoc
@@ -1,0 +1,43 @@
+// Module included in the following assemblies:
+//
+// assembly-config-mirrormaker2.adoc
+
+[id='con-mirrormaker-tasks-max-{context}']
+= Specifying a maximum number of tasks
+
+[role="_abstract"]
+Connectors create the tasks that are responsible for moving data in and out of Kafka.
+Each connector comprises one or more tasks that are distributed across a group of worker pods that run the tasks.
+
+Tasks run in parallel.
+If there are more tasks than workers, workers handle multiple tasks.
+
+You can specify the maximum number of connector tasks in your MirrorMaker configuration using the `tasksMax` property.
+Without specifying a maximum number of tasks, the default setting is a single task.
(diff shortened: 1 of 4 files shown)


PaulRMellor:
Hi @mimaison -- PR for `tasksMax` doc. If you could take a look, that would be great.

documentation/modules/configuring/con-config-mirrormaker2-tasks-max.adoc
```diff
@@ -0,0 +1,43 @@
...
+
+[id='con-mirrormaker-tasks-max-{context}']
+= Specifying a maximum number of tasks
+
+[role="_abstract"]
+Connectors create the tasks that are responsible for moving data in and out of Kafka.
+Each connector comprises one or more tasks that are distributed across a group of worker pods that run the tasks.
+
+Tasks run in parallel.
+If there are more tasks than workers, workers handle multiple tasks.
```
scholzj, line 13: I think the text is good. But I think it is worth mentioning that a single task will run only within one worker pod. (i.e. having replicas=10 but keeping the tasksMax=1 is useless). I think it kind of follows from the text, but might not hurt to be explicit about this I think

scholzj (commented):
One nit, otherwise it looks good.

documentation/modules/configuring/con-config-mirrormaker2-tasks-max.adoc
PaulRMellor: Changed to:

> Tasks run in parallel.
> Workers are assigned one or more tasks.
> A single task is handled by one worker pod, so you don't need more worker pods than tasks.
> If there are more tasks than workers, workers handle multiple tasks.

scholzj (approved):

PR #6529, opened 2022-03-16
[MO] - [system test] -> Using 3 ZooKeeper replicas instead of 2

see-quick:

### Type of change

- Bugfix

### Description

This PR fixes problem in RecoveryST, SecurityST ∧ KafkaST where we use 2 ZooKeeper replicas and replace them with 3 (using an even number of ZooKeeper rnodes is not recommended).

Eventually, the Kafka cluster is deployed but it took 8 minutes to do so...
```
recovery-cluster-1461276912-kafka-0         0/1     Running   0               28s
recovery-cluster-1461276912-kafka-1         0/1     Running   0               28s
recovery-cluster-1461276912-kafka-2         0/1     Running   0               28s
recovery-cluster-1461276912-zookeeper-0     1/1     Running   7 (3m42s ago)   10m
recovery-cluster-1461276912-zookeeper-1     1/1     Running   7 (3m42s ago)   10m
strimzi-cluster-operator-6499b64c6c-sfn9p   1/1     Running   0               10m
```
With this change, I eliminate such problems.

### Checklist

### DIFF
--- a/systemtest/src/test/java/io/strimzi/systemtest/operators/RecoveryIsolatedST.java
+++ b/systemtest/src/test/java/io/strimzi/systemtest/operators/RecoveryIsolatedST.java
@@ -14,7 +14,6 @@ import io.strimzi.api.kafka.model.KafkaResources;
 import io.strimzi.systemtest.AbstractST;
 import io.strimzi.systemtest.BeforeAllOnce;
 import io.strimzi.systemtest.Constants;
-import io.strimzi.systemtest.Environment;
 import io.strimzi.systemtest.annotations.IsolatedSuite;
 import io.strimzi.systemtest.annotations.StrimziPodSetTest;
 import io.strimzi.systemtest.resources.operator.SetupClusterOperator;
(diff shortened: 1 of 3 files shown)


systemtest/src/test/java/io/strimzi/systemtest/operators/RecoveryIsolatedST.java
```diff
@@ -53,6 +52,8 @@
 class RecoveryIsolatedST extends AbstractST {
 
     static String sharedClusterName;
+    private static final int KAFKA_REPLICAS = 3;
+    private static final int ZOOKEEPER_REPLICAS = 3;
```
Frawless, line 56: Why we can't keep the default values and override them with the same value?

Frawless (approved):

scholzj (commented):
Wait wait wait ... having 2 Zoo nodes is stupid. So I think the change is good. But the explanation in the description does not make sense. The error you list there has nothing to do with the number of Zoo nodes. In fact, I guess it might be self corrected few lines later. So the reasoning is IMHO wrong.

scholzj (approved):

PR #6535, opened 2022-03-17
docs: securing mm2

PaulRMellor:

**Documentation**
- Adds a new procedure called _Securing a Kafka MirrorMaker 2.0 deployment_ to the MirrorMaker2 section of the _Configuring_ guide which describes in outline how to secure connections to a source and target cluster when using MirrorMaker2
- Updates attributes for `inter.broker.protocol.version`  (used in new procedure) and `log.message.format.version`
- Moves a couple of files under the `modules/configuring` directory

- [ ] Write tests
- [x] Update documentation
- [ ] Update CHANGELOG.md


documentation/modules/configuring/proc-config-mirrormaker2-securing-connection.adoc
```diff
@@ -0,0 +1,257 @@
...
+
+NOTE: In this procedure, the certificates generated by the Cluster Operator are used, but you can replace them by xref:installing-your-own-ca-certificates-str[installing your own certificates].
+You can also configure your listener to xref:kafka-listener-certificates-str[use a Kafka listener certificate managed by an external Certificate Authority].
+
+.Before you start
+Before starting this procedure, take a look at the link:{BookURLDeploying}#deploy-examples-{context}[example configuration files^] provided by Strimzi.
+They include examples for securing a deployment of MirrorMaker 2.0 using TLS or SCRAM-SHA-512 authentication.
+The examples specify internal listeners for connecting within a Kubernetes cluster.
+
+The examples also show the ACLs needed by MirrorMaker 2.0 to allow operations on the source and target Kafka clusters.
```
PaulRMellor, line 29: Not sure if we want to list the ACLs required by the connectors in the doc

documentation/modules/configuring/proc-config-mirrormaker2-securing-connection.adoc
```diff
@@ -0,0 +1,257 @@
...
+. Create or update the `Kafka` resources.
++
+[source,shell,subs=+quotes]
+----
+kubectl apply -f _<kafka_configuration_file>_
+----
++
+The Cluster Operator creates the listeners and sets up the cluster and client certificate authority (CA) certificates to enable authentication within the Kafka cluster.
++
+The certificates are created with the same name as the Kafka resource.
```
scholzj, line 99: I'm not sure this makes sense. I guess you  might want to say something like

```suggestion
The certificates are created in the secret `<cluster-name>-cluster-ca-cert`.
```

The certificates do not really have a name and the Secret name differs.

documentation/modules/configuring/proc-config-mirrormaker2-securing-connection.adoc
```diff
@@ -0,0 +1,257 @@
...
+
+.. Configure the ACLs needed by MirrorMaker 2.0 to allow operations on the source and target Kafka clusters.
++
+The ACLs are used by the internal MirrorMaker connectors, and by the underlying Kafka Connect framework.
+--
++
+.Example user configuration for TLS client authentication
+[source,yaml,subs="attributes+"]
+----
+apiVersion: {KafkaUserApiVersion}
```
scholzj, line 118: Same as above, the users are actually different. So listing only one is weird and can be confusing. I would either list none or both.

scholzj (commented):
I left some comments. One thing to think about are the namespaces ... deploying everything into single namespace makes things easier. But not realistic.
* The Topic operator does not work on both clusters in the same namespace. The UO has some limitations as well. Should we mention it?
* Within the same namespace, the secrets can be just loaded directly. With 2 namespaces / 2 Kube clusters, you would need to copy them. This should be probably at least mentioned.

documentation/modules/configuring/proc-config-mirrormaker2-securing-connection.adoc
```diff
@@ -0,0 +1,257 @@
...
+.Procedure
+
+. Configure two `Kafka` resources, one to secure the source Kafka cluster and one to secure the target Kafka cluster.
++
+You can add listener configuration for authentication and enable authorization.
++
+In this example, an internal listener is configured for a Kafka cluster with TLS encryption and authentication.
+Kafka `simple` authorization is enabled.
++
+.Example source Kafka cluster configuration with TLS authentication
```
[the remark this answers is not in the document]
PaulRMellor: I've added both.

documentation/modules/configuring/proc-config-mirrormaker2-securing-connection.adoc
PaulRMellor: I've added both

documentation/modules/configuring/proc-config-mirrormaker2-securing-connection.adoc
```diff
@@ -86,22 +92,64 @@ spec:
...
+          size: 100Gi
+          deleteClaim: false
+  zookeeper:
+    replicas: 1
+    storage:
+      type: persistent-claim
+      size: 100Gi
+      deleteClaim: false
+  entityOperator:
+    userOperator: {}
```
scholzj, line 137: Since you have different namespaces, you can have `topicOperator: {}` here.

documentation/modules/configuring/proc-config-mirrormaker2-securing-connection.adoc
```diff
@@ -162,6 +210,146 @@ spec:
...
+      - resource:
+          type: topic
+          name: heartbeats
+        operation: Describe
+      - resource:
+          type: topic
+          name: heartbeats
+        operation: Write
+----
++
```
scholzj, line 352: We need to make sure it is clear to the users that each user has to be create in different namespace.

documentation/modules/configuring/proc-config-mirrormaker2-securing-connection.adoc
```diff
@@ -239,8 +427,10 @@ spec:
       topicsPattern: ".*"
       groupsPattern: ".*"
 ----
-<1> The TLS certificates for the source Kafka cluster.
+<1> The TLS certificates for the source Kafka cluster. If they are in a separate namespace, copy the cluster secrets from the namespace of the Kafka cluster.
 <2> The user authentication for accessing the source Kafka cluster using the xref:type-KafkaClientAuthenticationTls-reference[TLS mechanism].
+<1> The TLS certificates for the target Kafka cluster.
+<2> The user authentication for accessing the target Kafka cluster.
 
 . Create or update the `KafkaMirrorMaker2` resource.
```
scholzj, line 435: We should probably point the user into which namespace should this be deployed? (the namespace of the target cluster)

scholzj:
> I've assumed that we want to steer the user to do the right or least problematic things, so I've added separate namespaces as a prereq, with an explanation as to why.
I also specified -n _<namespace>_ in the step to apply the kafka clusters. Not sure if we want to add more to the steps along these lines. I haven't done the same for the users. I did mention in the MM config that you might need to copy the cluster secrets from the namespace of the Kafka cluster.

I think this is a hard decision. Using the same namespace is the _easiest_. But it is in reality also useless. With two different namespaces it is a bit more realistic although even that will not be what will most use-cases be about. In most cases it will be between different clusters. But the more realistic the use-case, the _harder_ to document it.

documentation/modules/configuring/proc-config-mirrormaker2-securing-connection.adoc
```diff
@@ -0,0 +1,450 @@
...
+          type: tls
+    authorization:
+      type: simple
+    config:
+      offsets.topic.replication.factor: 1
+      transaction.state.log.replication.factor: 1
+      transaction.state.log.min.isr: 1
+      default.replication.factor: 1
+      min.insync.replicas: 1
+      inter.broker.protocol.version: "3.1"
```
scholzj, line 122:
```suggestion
      inter.broker.protocol.version: "{DefaultInterBrokerVersion}"
```

scholzj (approved):
One it, LGTM otherwise.

PaulRMellor:
Peer review requested from @joaedwar

scholzj:
@PaulRMellor BTW: There seem to be some bugs in the changes failing the build:

```
asciidoctor -v --failure-level WARN -t -dbook -a ProductVersion=latest -a BridgeVersion=0.21.4 -a GithubVersion=main -a OAuthVersion=0.10.0 documentation/configuring/configuring.adoc -o documentation/html/configuring.html
asciidoctor: WARNING: modules/configuring/proc-config-mirrormaker2-securing-connection.adoc: line 433: callout list item index: expected 3, got 1
asciidoctor: WARNING: modules/configuring/proc-config-mirrormaker2-securing-connection.adoc: line 434: callout list item index: expected 4, got 2
```

PR #6538, opened 2022-03-18
docs: review edits

PaulRMellor:

**Documentation**

Introduces a few minor doc edits following review.
Includes:
- Text edits
- Table formatting change
- Updated links

### Checklist

- [ ] Write tests
- [x] Update documentation
- [ ] Update CHANGELOG.md


documentation/shared/attributes.adoc
```diff
@@ -87,7 +87,7 @@
 :KubernetesExternalDNS: link:https://github.com/kubernetes-incubator/external-dns[External DNS^]
 :ApacheKafkaBrokerConfig: link:https://kafka.apache.org/documentation/#brokerconfigs[Apache Kafka documentation^]
 :ApacheKafkaConnectConfig: link:https://kafka.apache.org/documentation/#connectconfigs[Apache Kafka documentation^]
-:ApacheZookeeperConfig: link:https://zookeeper.apache.org/doc/r3.5.8/zookeeperAdmin.html[ZooKeeper documentation^]
+:ApacheZookeeperConfig: link:https://zookeeper.apache.org/doc/current/[ZooKeeper documentation^]
```
scholzj, line 90: `current` here links to ZooKeeper 3.8. Kafka is using 3.6.3. I think we should point to a specific version.

documentation/modules/cruise-control/con-rebalance-performance.adoc
```diff
@@ -56,41 +58,44 @@ Cruise Control does not perform inter-broker and intra-broker balancing at the s
...
+* The individual rebalance performance configurations can be set under `KafkaRebalance.spec`.
 
-|============================================================================================================================
-| Server and `KafkaRebalance` Configuration                | Description                                          | Default Value
+The relevant configurations are summarized in the following table.
 
-| `num.concurrent.partition.movements.per.broker`   .2+| 
-  The maximum number of inter-broker partition movements in each partition reassignment batch              .2+| 5 
-| `concurrentPartitionMovementsPerBroker` 
+.Rebalance performance tuning configuration
```
scholzj, line 66: This table seems to be wrong:
* I don't think it makes it really clear what is what. I think that using 4 columns (One for the configuration key in `Kafka.spec.cruiseControl.config`, one for `KafkaRebalance.spec`, one for description and one for default value) would be much better. If the problem is that it would be too wide, maybe it should be split into two separate tables. The problem of the current table design shows for example with the `rebalanceDisk` row where it is absolutely not clear where it belongs-
* Your changes of the `KafkaRebalance.spec` fields seem wrong - they are not supported by the [API](https://strimzi.io/docs/operators/latest/full/configuring.html#type-KafkaRebalanceSpec-reference)

documentation/shared/attributes.adoc
PaulRMellor: I was trying to avoid version, but updated to `https://zookeeper.apache.org/doc/r3.6.3/`

documentation/modules/cruise-control/con-rebalance-performance.adoc
PaulRMellor: I've changed to four columns, switched them around and modified the titles
I've changed the KafkaBalance properties

scholzj (approved):

documentation/modules/configuring/con-config-mirrormaker2-tasks-max.adoc
```diff
@@ -35,11 +35,11 @@ spec:
   # ...
 ----
 
-For sink and source connectors, the maximum number of tasks that can run in parallel depends on a number of factors.
+For connectors, the maximum number of tasks that can run in parallel depends on a number of factors.
```
PaulRMellor, line 38: @mimaison - I've updated the tasksMax section

mimaison: For MirrorMaker Source the maximum number of tasks is the number of partition being replicated. The number of tasks that will be started is the minimum value between that and tasks.max. There are no other factors.

For Checkpoint, it's similar but using the number of consumer groups being replicated.

For Heartbeat it's always using a single task.

I think we can remove all the rest as it's for other connectors.

PR #6541, opened 2022-03-18
`StrimziPodSet` reconciliation only switch

scholzj:
### Type of change

- Enhancement / new feature

### Description

This PR implements the [`StrimziPodSets`-only reconciliation switch](https://github.com/strimzi/proposals/blob/main/031-statefulset-removal.md#strimzipodsets-only-switch) as defined in the Strimzi Proposal 031. 

With StatefulSets, users can switch off the Cluster Operator and the Kubernetes StatefulSet controller will still manage the pods, recreate them etc. But with StrimziPodsSets, it is not easy to just _switch-off_ the operator because it also manages the pods. this switch allows to disable the reconciliation of all the `Kafka*` custom resources, but still manage the Pods based on the existing StrimziPodSet resources.

### Checklist

- [x] Write tests
- [x] Update documentation

### DIFF
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/ClusterOperator.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/ClusterOperator.java
@@ -101,27 +101,33 @@ public class ClusterOperator extends AbstractVerticle {
             strimziPodSetController.start();
         }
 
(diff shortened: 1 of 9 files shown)


ppatierno (approved):

scholzj:
_Note: Test failures are not related to this PR but to #6544 / #6529_

PR #6542, opened 2022-03-18
Fix non-cascading deletion of the `StrimziPodSet` resources

scholzj:
### Type of change

- Bugfix

### Description

When the StrimziPodSet is deleted with `orphan` cascading policy, the Pods should be kept. This was not always the case and sometimes, the non-cascading deletion resulted in the Pod deletion as well. 

The Pod deletion with StrimziPodSets relies on Kubernetes Garbage Collection. So when the StrimziPodSet resource is deleted, the StrimziPodSet controller actually doesn't do anything and waits for Kubernetes to delete the PodSet as well as the Pods. This works fine for normal deletion. But not for non-cascading deletion. For non-cascading deletion, Kubernetes does following things:
* Marks the PodSet as deleting (sets the deletion timestamp)
* Modifies the Pods and removes the ownerReferences from them so that they are not deleted
* Actually deletes the PodSet

What happens is that the PodSet controller gets several events when the PodSet or the Pods are modified. And sometimes it happens that it reconciles the pods when the PodSet still exists but the ownerReferences are already removed from Pods => and as a result, it adds the ownerReferences back. So when later the PodSet is deleted, the Pods are deleted as well.

This PR updates the PodSet controller. It now checks whether the PodSet isn't being already deleted. And if it is, it will not touch anything and just skip the reconciliation. Only when the PodSet is not being deleted, we actually reconcile it. That way we ensure that when the PodSet is being deleted, we leave it to Kubernetes and Kubernetes do its job - either delete the pods or leave them be.

### Checklist

- [x] Write tests

### DIFF
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/StrimziPodSetController.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/StrimziPodSetController.java
@@ -211,6 +211,17 @@ public class StrimziPodSetController implements Runnable {
     }
 
     /**
+     * Checks whether the StrimziPodSet is being deleted or not. This is needed to handle non-cascading deletions.
+     *
+     * @param podSet    StrimziPodSet which needs to be checked whether it is being deleted
+     *
+     * @return          True if the PodSet is being deleted. False otherwise.
+     */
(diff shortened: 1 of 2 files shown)


ppatierno (approved):

scholzj:
_Note: Test failures are not related to this PR but to #6544 / #6529_

PR #6549, opened 2022-03-22
Separate Kafka Exporter reconciliation from `KafkaAssemblyOperator` class

scholzj:
### Type of change

- Refactoring

### Description

The `KafkaAssemblyOperator` class is currently very monolithic. It contains the reconciliation steps and state for all sub-components which are part of the Kafka custom resource. This PR tries to peel-off Kafka Exporter reconciliation and the state these steps use.

This does not do much on its own, but it seems fairly easy and it reduces the `KafkaAssemblyOperator` footprint. It should make it a bit more clear which parts of the state Kafka Exporter actually needs, it should make it easier to edit the Kafka Exporter recocniliation and if needed it should also make it easier to test it (this PR does not add any new tests - the testing is currently done through the `KafkaAssemblyOperator` testing. But if / when needed, it can be tested separately).

If this is accepted, I think we can similarly handle also the JMXTrans, Cruise Control and EntityOperator reconciliation. It might be much harder to use this to separate ZooKeeper and Kafka.

### Checklist

### DIFF
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaExporter.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaExporter.java
@@ -67,8 +67,6 @@ public class KafkaExporter extends AbstractModel {
     protected String logging;
     protected String version;
 
-    private boolean isDeployed;
-
     protected List<ContainerEnvVar> templateContainerEnvVars;
     protected SecurityContext templateContainerSecurityContext;
 
@@ -103,12 +101,20 @@ public class KafkaExporter extends AbstractModel {
 
     }
 
+    /**
+     * Builds the KafkaExporter model from the Kafka custom resource. If KafkaExporter is not enabled, it will return null.
+     *
+     * @param reconciliation    Reconciliation marker for logging
+     * @param kafkaAssembly     The Kafka CR
+     * @param versions          The list of supported Kafka versions
+     *
+     * @return                  KafkaExporter model object when Kafka Exporter is enabled or null if it is disabled.
+     */
     public static KafkaExporter fromCrd(Reconciliation reconciliation, Kafka kafkaAssembly, KafkaVersion.Lookup versions) {
-        KafkaExporter kafkaExporter = new KafkaExporter(reconciliation, kafkaAssembly);
-
         KafkaExporterSpec spec = kafkaAssembly.getSpec().getKafkaExporter();
+
         if (spec != null) {
-            kafkaExporter.isDeployed = true;
+            KafkaExporter kafkaExporter = new KafkaExporter(reconciliation, kafkaAssembly);
 
             kafkaExporter.setResources(spec.getResources());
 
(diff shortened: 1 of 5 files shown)


scholzj:
@tombentley @ppatierno I wonder what would you think about this. It is not much, but it might help to reduce the `KafkaAssemblyOperator` class at least a little bit.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaAssemblyOperator.java
```diff
@@ -4463,66 +4410,33 @@ String getInternalServiceHostname(String serviceName, boolean useServiceDnsDomai
...
-                Future<Deployment> future = deploymentOperations.getAsync(namespace, this.kafkaExporter.getName());
-                return future.compose(dep -> {
-                    return withVoid(deploymentOperations.waitForObserved(reconciliation, namespace, this.kafkaExporter.getName(), 1_000, operationTimeoutMs));
-                }).compose(dep -> {
-                    return withVoid(deploymentOperations.readiness(reconciliation, namespace, this.kafkaExporter.getName(), 1_000, operationTimeoutMs));
-                }).map(i -> this);
-            }
-            return withVoid(Future.succeededFuture());
+        /**
+         * Run the reconciliation pipeline for the Kafka Exporter cluster
```
tomncooper, line 4432: Nit: is it and Exporter "Cluster"?

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaAssemblyOperator.java
```diff
@@ -4463,66 +4410,33 @@ String getInternalServiceHostname(String serviceName, boolean useServiceDnsDomai
...
+         *
+         * @return                      Future with Reconciliation State
```
tomncooper, line 4434: Nit: Tabs vs spaces?

tomncooper (approved):
Couple of doc nits. Otherwise LGTM. I wonder how much more of this we can do to the KAO?

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaExporterReconciler.java
```diff
@@ -0,0 +1,201 @@
...
+
+import java.util.Date;
+import java.util.List;
+import java.util.function.Supplier;
+
+/**
+ * Class used for reconciliation of Kafka Exporter. This class contains both the steps of the Kafka Exporter
+ * reconciliation pipeline and is also used to store the state between them.
+ */
+public class KafkaExporterReconciler {
```
ppatierno, line 32: Compared to the naming in operator sdks (the Go [1] and the Java [2] ones) maybe we should call this as `KafkaExporterController`? It takes care of everything related to the Kafka Exporter inside the overall Strimzi cluster operator. The only difference with "controllers" in other operators is that this one doesn't really handle a `KafkaExporter` custom resource but more a part of the `Kafka` one.

[1] https://github.com/operator-framework/operator-sdk-samples/blob/master/go/memcached-operator/controllers/memcached_controller.go
[2] https://github.com/java-operator-sdk/samples/blob/main/webserver/src/main/java/io/javaoperatorsdk/operator/sample/WebServerController.java

scholzj: It does not really follow the Kubernetes controller pattern too much. It follows our _assembly operator_ pattern. So I intentionally wanted to avoid calling it a controller. I was thinking about calling it `KafkaExporterAssemblyOperator` ... but it might be confusing with not being a high level `*AssemblyOperator` as the other ones. So I chose `Reconciler`. I'm open to discussion about the name -> but I do not want to call it Controller.

ppatierno (approved):

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaExporter.java
```diff
@@ -104,11 +102,10 @@ protected KafkaExporter(Reconciliation reconciliation, HasMetadata resource) {
     }
 
     public static KafkaExporter fromCrd(Reconciliation reconciliation, Kafka kafkaAssembly, KafkaVersion.Lookup versions) {
```
tombentley, line 104: Document that it can return null?

tombentley (approved):
LGTM, and thanks for this long over due effort.

PR #6552, opened 2022-03-22
Removed the needs for system properties field

ppatierno:
### Type of change

- Refactoring

### Description

This PR does a little bit of refactoring by removing the need for the `javaSystemProperties` field in the `AbstractModel` and using the same approach for other JVM related options.

### Checklist

- [ ] Write tests
- [ ] Update documentation
- [ ] Update CHANGELOG.md


scholzj (commented):
Why do you once use `ModelUtils.javaOptions` and once `jvmSystemProperties` with `ModelUtils.getJavaSystemPropertiesToString`? Wouldn't it make sense to unify it completely?

If not, should `jvmSystemProperties(...)` really ba part of the `AbstractModel`? I think this is one of the issues of the `AbstractModel` class. It is full of methods used in half of the model classes while other half does something completely different.

ppatierno:
> Why do you once use ModelUtils.javaOptions and once jvmSystemProperties with ModelUtils.getJavaSystemPropertiesToString? Wouldn't it make sense to unify it completely?

I was expecting the question :-)
The `javaOptions` method is really a duplicate of others in the `AbstractModel` class and I am talking about `heapOptions`, `jvmPerformanceOptions` and `jvmSystemProperties`. One problem I saw it that the `heapOptions` seem to be really tight to Kafka (it sets the KAFKA_HEAP_OPTS env var) while `javaOptions` is more on STRIMZI_JAVA_OPTS. The `heapOptions` also has the heap fraction and max bytes parameters that configure other env vars not really used in other places (talking where `javaOptions` is used). I would agree that a better refactoring could be done, but I would also start with this step first.

> If not, should jvmSystemProperties(...) really ba part of the AbstractModel? I think this is one of the issues of the AbstractModel class. It is full of methods used in half of the model classes while other half does something completely different.

That's fine, I can move to the `ModeUtils` but then what about the others: `heapOptions` and `jvmPerformanceOptions`. I think they should be moved as well.

scholzj:
> I was expecting the question :-)

Maybe having it in the comments would help if it is expected question :-)

> what about the others: heapOptions and jvmPerformanceOptions. I think they should be moved as well.

That is a bit hard to say for me without seeing how they are used etc. If the same applies to them as well then I guess yes.

scholzj:
PS: i can see how removing the unnecessary field might be useful anyway even without doing some further movements. So if you want to stick with what you have in the PR now, its fine.

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/AbstractModel.java
```diff
@@ -1567,6 +1561,20 @@ protected void jvmPerformanceOptions(List<EnvVar> envVars) {
...
 
+    /**
+     * Adds STRIMZI_JAVA_SYSTEM_PROPERTIES variable to the EnvVar list if any system properties were specified.
+     *
+     * @param envVars List of Environment Variables to add to
+     */
+    protected void jvmSystemProperties(List<EnvVar> envVars) {
+        if (jvmOptions != null) {
+            String jvmSystemPropertiesString = ModelUtils.getJavaSystemPropertiesToString(jvmOptions.getJavaSystemProperties());
+            if (!jvmSystemPropertiesString.isEmpty()) {
```
scholzj, line 1572: I think you need null check here as well.

ppatierno:
@scholzj I moved all the methods we mentioned in the `ModelUtils` class which sounds better. After merging this PR I will take a look to unify with `javaOptions` method to avoid the code duplication we have today.

PR #6558, opened 2022-03-23
[MO] - [system test] -> deployment resource not specifying namespace …

see-quick:
…when get

### Type of change

- Bugfix

### Description

This PR fixes namespace handling in DeploymentResource, where in some edge cases the Kubernetes client could not get the resource in the correct namespace.

### Checklist

### DIFF
--- a/systemtest/src/main/java/io/strimzi/systemtest/resources/kubernetes/DeploymentResource.java
+++ b/systemtest/src/main/java/io/strimzi/systemtest/resources/kubernetes/DeploymentResource.java
@@ -20,7 +20,7 @@ public class DeploymentResource implements ResourceType<Deployment> {
     }
     @Override
     public Deployment get(String namespace, String name) {
-        return ResourceManager.kubeClient().namespace(namespace).getDeployment(namespace, name);
+        return ResourceManager.kubeClient().getDeployment(namespace, name);
     }
     @Override
     public void create(Deployment resource) {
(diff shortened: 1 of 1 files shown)


systemtest/src/main/java/io/strimzi/systemtest/resources/kubernetes/DeploymentResource.java
```diff
@@ -20,7 +20,7 @@ public String getKind() {
     }
     @Override
     public Deployment get(String namespace, String name) {
-        return ResourceManager.kubeClient().namespace(namespace).getDeployment(name);
+        return ResourceManager.kubeClient().namespace(namespace).getDeployment(namespace, name);
```
Frawless, line 23: Does it point to some problem in fabric8? Or why do we need to specify namespace twice?

see-quick: No, but in the case of parallel execution, there is a potential error to have a race condition. Because this piece of code is not synchronized, thus multiple threads could access it (with different namespaces) and it can change the `shared` state of the Kube client. Thus we need it explicitly specify `namespace` inside `getDeployment(...)` method.

see-quick: What I can do about it is remove `unnecessary` switching between namespaces from:
```java
return ResourceManager.kubeClient().namespace(namespace).getDeployment(namespace, name);
```
to
```java
return ResourceManager.kubeClient().getDeployment(namespace, name);
```

Frawless: Ye, that change would be nice.

see-quick: Changed )

Frawless (approved):

PR #6561, opened 2022-03-24
Remove the refresh annotation when CC is not enabled or Cluster label is missing/wrong.

ShubhamRwt:

### Type of change

- Task

### Description

This PR address the issue mentioned in #3636. This removes the use of refresh annotation when we apply KR and CC is not enabled or the cluster label is missing/wrong.

### Checklist

- [ ] Write tests
- [ ] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceAssemblyOperator.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceAssemblyOperator.java
@@ -724,9 +724,9 @@ public class KafkaRebalanceAssemblyOperator
                                                     KafkaRebalance kafkaRebalance,
                                                     KafkaRebalanceAnnotation rebalanceAnnotation,
                                                     RebalanceOptions.RebalanceOptionsBuilder rebalanceOptionsBuilder) {
-        if (rebalanceAnnotation == KafkaRebalanceAnnotation.refresh) {
-            // The user has fixed the error on the resource and want to 'refresh'
-            // This actually requests a new rebalance proposal
+
+        if (shouldRenewRebalance(kafkaRebalance, rebalanceAnnotation)) {
+            // CC is activated from the disabled state or the user has fixed the error on the resource and want to 'refresh'
             return onNew(reconciliation, host, apiClient, kafkaRebalance, rebalanceOptionsBuilder);
         } else {
             // Stay in the current NotReady state, returning null as next state
@@ -1083,11 +1083,13 @@ public class KafkaRebalanceAssemblyOperator
         String clusterName = kafkaRebalance.getMetadata().getLabels() == null ? null : kafkaRebalance.getMetadata().getLabels().get(Labels.STRIMZI_CLUSTER_LABEL);
         String clusterNamespace = kafkaRebalance.getMetadata().getNamespace();
         if (clusterName == null) {
-            LOGGER.warnCr(reconciliation, "Resource lacks label '{}': No cluster related to a possible rebalance.", Labels.STRIMZI_CLUSTER_LABEL);
-            return updateStatus(reconciliation, kafkaRebalance, new KafkaRebalanceStatus(),
-                    new InvalidResourceException("Resource lacks label '"
-                            + Labels.STRIMZI_CLUSTER_LABEL
-                            + "': No cluster related to a possible rebalance.")).mapEmpty();
+
+            String errorString = "Resource lacks label '" + Labels.STRIMZI_CLUSTER_LABEL + "': No cluster related to a possible rebalance.";
+            LOGGER.warnCr(reconciliation, errorString);
+            KafkaRebalanceStatus status = new KafkaRebalanceStatus();
+            status.addCondition(StatusUtils.buildWarningCondition(CruiseControlIssues.clusterLabelMissing.getReason(), errorString));
+            return updateStatus(reconciliation, kafkaRebalance, status,
+                    new InvalidResourceException(errorString)).mapEmpty();
         }
 
         // Get associated Kafka cluster state
@@ -1104,10 +1106,12 @@ public class KafkaRebalanceAssemblyOperator
                         LOGGER.debugCr(reconciliation, "{} {} in namespace {} belongs to a Kafka cluster {} which does not match label selector {} and will be ignored", kind(), kafkaRebalance.getMetadata().getName(), clusterNamespace, clusterName, kafkaSelector.get().getMatchLabels());
                         return Future.succeededFuture();
                     } else if (kafka.getSpec().getCruiseControl() == null) {
-                        LOGGER.warnCr(reconciliation, "Kafka resource lacks 'cruiseControl' declaration : No deployed Cruise Control for doing a rebalance.");
-                        return updateStatus(reconciliation, kafkaRebalance, new KafkaRebalanceStatus(),
-                                new InvalidResourceException("Kafka resource lacks 'cruiseControl' declaration "
-                                        + ": No deployed Cruise Control for doing a rebalance.")).mapEmpty();
+                        String errorString = "Kafka resource lacks 'cruiseControl' declaration " + ": No deployed Cruise Control for doing a rebalance.";
+                        LOGGER.warnCr(reconciliation, errorString);
+                        KafkaRebalanceStatus status = new KafkaRebalanceStatus();
+                        status.addCondition(StatusUtils.buildWarningCondition(CruiseControlIssues.cruiseControlDisabled.getReason(), errorString));
+                        return updateStatus(reconciliation, kafkaRebalance, status,
+                                new InvalidResourceException(errorString)).mapEmpty();
                     }
 
                     if (kafka.getSpec().getKafka().getStorage() instanceof JbodStorage) {
@@ -1283,6 +1287,50 @@ public class KafkaRebalanceAssemblyOperator
     private boolean hasRebalanceAnnotation(KafkaRebalance kafkaRebalance) {
         return kafkaRebalance.getMetadata().getAnnotations() != null &&
                 kafkaRebalance.getMetadata().getAnnotations().containsKey(ANNO_STRIMZI_IO_REBALANCE);
+    }
+
+    private boolean shouldRenewRebalance(KafkaRebalance kafkaRebalance, KafkaRebalanceAnnotation rebalanceAnnotation) {
+        // check if refresh annotation applied or request are made when CC is not active
+        if (rebalanceAnnotation == KafkaRebalanceAnnotation.refresh || kafkaRebalance.getStatus().getConditions().stream().anyMatch(condition -> "ConnectException".equals(condition.getReason()))) {
+            return true;
+        } else {
+            return CruiseControlIssues.checkForMatch(kafkaRebalance.getStatus().getConditions());
+        }
+    }
+
+    /**
+     * Contains the issues that will be picked up in the periodic reconciliation, once corrected.
+     */
+    private enum CruiseControlIssues {
+
+        /**
+         * The Kafka cluster label is missing .
+         */
+        clusterLabelMissing("Cluster Label missing"),
+
+        /**
+         * Cruise Control was not declared in the Kafka Resource
+         */
+        cruiseControlDisabled("Cruise Control disabled");
+
+        public String getReason() {
+            return reason;
+        }
+
+        private final String reason;
+
+        CruiseControlIssues(String reason) {
+            this.reason = reason;
+        }
+
+        public static boolean checkForMatch(List<Condition> conditions) {
+            for (CruiseControlIssues issue : CruiseControlIssues.values()) {
+                if (conditions.stream().anyMatch(condition -> issue.getReason().equals(condition.getReason()))) {
+                    return true;
+                }
+            }
+            return false;
+        }
     }
 
     private KafkaRebalanceState state(KafkaRebalance kafkaRebalance) {
--- a/systemtest/src/test/java/io/strimzi/systemtest/cruisecontrol/CruiseControlST.java
+++ b/systemtest/src/test/java/io/strimzi/systemtest/cruisecontrol/CruiseControlST.java
@@ -8,8 +8,11 @@ import io.fabric8.kubernetes.api.model.Container;
 import io.fabric8.kubernetes.api.model.Quantity;
 import io.fabric8.kubernetes.api.model.ResourceRequirementsBuilder;
 import io.strimzi.api.kafka.model.CruiseControlResources;
+import io.strimzi.api.kafka.model.CruiseControlSpec;
 import io.strimzi.api.kafka.model.KafkaRebalance;
 import io.strimzi.api.kafka.model.KafkaTopicSpec;
+import io.fabric8.kubernetes.api.model.LabelSelector;
+import io.strimzi.api.kafka.model.KafkaResources;
 import io.strimzi.api.kafka.model.status.KafkaRebalanceStatus;
 import io.strimzi.api.kafka.model.status.KafkaStatus;
 import io.strimzi.api.kafka.model.storage.JbodStorage;
@@ -31,12 +34,14 @@ import io.strimzi.systemtest.templates.crd.KafkaClientsTemplates;
 import io.strimzi.systemtest.templates.crd.KafkaRebalanceTemplates;
 import io.strimzi.systemtest.templates.crd.KafkaTemplates;
 import io.strimzi.systemtest.templates.crd.KafkaTopicTemplates;
+import io.strimzi.systemtest.utils.RollingUpdateUtils;
 import io.strimzi.systemtest.utils.StUtils;
 import io.strimzi.systemtest.utils.kafkaUtils.KafkaRebalanceUtils;
 import io.strimzi.systemtest.utils.kafkaUtils.KafkaTopicUtils;
 import io.strimzi.systemtest.utils.kafkaUtils.KafkaUtils;
 import io.strimzi.systemtest.utils.kubeUtils.controllers.DeploymentUtils;
 import io.strimzi.test.k8s.KubeClusterResource;
+import io.strimzi.systemtest.utils.kubeUtils.objects.PodUtils;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.junit.jupiter.api.Tag;
@@ -159,16 +164,27 @@ public class CruiseControlST extends AbstractST {
     void testCruiseControlWithRebalanceResourceAndRefreshAnnotation(ExtensionContext extensionContext) {
         String clusterName = mapWithClusterNames.get(extensionContext.getDisplayName());
 
-        resourceManager.createResource(extensionContext, KafkaTemplates.kafkaWithCruiseControl(clusterName, 3, 3)
-            .editMetadata()
-            .withNamespace(namespace)
-            .endMetadata()
-            .build());
-        resourceManager.createResource(extensionContext, KafkaRebalanceTemplates.kafkaRebalance(clusterName)
-            .editMetadata()
-            .withNamespace(namespace)
-            .endMetadata()
-            .build());
+        resourceManager.createResource(extensionContext, KafkaTemplates.kafkaEphemeral(clusterName, 3, 3)
+                .editMetadata()
+                    .withNamespace(namespace)
+                .endMetadata()
+                .build());
+        resourceManager.createResource(extensionContext, false,  KafkaRebalanceTemplates.kafkaRebalance(clusterName)
+                .editMetadata()
+                    .withNamespace(namespace)
+                .endMetadata()
+                .build());
+
+        final LabelSelector kafkaSelector = KafkaResource.getLabelSelector(clusterName, KafkaResources.kafkaStatefulSetName(clusterName));
+
+        KafkaRebalanceUtils.waitForKafkaRebalanceCustomResourceState(namespace, clusterName, KafkaRebalanceState.NotReady);
+
+        Map<String, String> kafkaPods = PodUtils.podSnapshot(namespace, kafkaSelector);
+
+        // Cruise Control spec is now enabled
+        KafkaResource.replaceKafkaResourceInSpecificNamespace(clusterName, kafka -> kafka.getSpec().setCruiseControl(new CruiseControlSpec()), namespace);
+
+        RollingUpdateUtils.waitTillComponentHasRolled(namespace, kafkaSelector, 3, kafkaPods);
 
         KafkaRebalanceUtils.doRebalancingProcess(new Reconciliation("test", KafkaRebalance.RESOURCE_KIND, namespace, clusterName), namespace, clusterName);
 
--- a/cluster-operator/src/test/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceAssemblyOperatorTest.java
+++ b/cluster-operator/src/test/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceAssemblyOperatorTest.java
@@ -1056,8 +1056,43 @@ public class KafkaRebalanceAssemblyOperatorTest {
      * 2. The operator checks that the Kafka cluster specified in the KafkaRebalance resource (via label) doesn't have Cruise Control configured
      * 4. The KafkaRebalance resource moves to NotReady state
      */
+
     @Test
     public void testCruiseControlDisabled(VertxTestContext context) {
+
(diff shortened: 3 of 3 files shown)


cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceAssemblyOperator.java
```diff
@@ -723,7 +723,11 @@ public MapAndStatus(T loadMap, K status) {
                                                     KafkaRebalance kafkaRebalance,
                                                     KafkaRebalanceAnnotation rebalanceAnnotation,
                                                     RebalanceOptions.RebalanceOptionsBuilder rebalanceOptionsBuilder) {
-        if (rebalanceAnnotation == KafkaRebalanceAnnotation.refresh) {
+
+        if (rebalanceAnnotation == KafkaRebalanceAnnotation.refresh
```
tomncooper, line 727: I think it would be better to move this check out to a private method (something like `shouldRenewRebalance` or similar) that returns a boolean. So we just have `if shouldRenewRebalance(rebalanceAnnotation)`.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceAssemblyOperator.java
```diff
@@ -723,7 +723,11 @@ public MapAndStatus(T loadMap, K status) {
...
+                || kafkaRebalance.getStatus().getConditions().stream().anyMatch(condition -> "Cluster Label missing".equals(condition.getReason()))
```
tomncooper, line 728: Having this set of conditions that constitute a renewal hard coded here feels wrong? Should we have these in a list defined somewhere centrally? Or some kind of enum (I love enums).

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceAssemblyOperator.java
```diff
@@ -1048,7 +1052,11 @@ public MapAndStatus(T loadMap, K status) {
         String clusterNamespace = kafkaRebalance.getMetadata().getNamespace();
         if (clusterName == null) {
             LOGGER.warnCr(reconciliation, "Resource lacks label '{}': No cluster related to a possible rebalance.", Labels.STRIMZI_CLUSTER_LABEL);
-            return updateStatus(reconciliation, kafkaRebalance, new KafkaRebalanceStatus(),
+            KafkaRebalanceStatus status = new KafkaRebalanceStatus();
+            status.addCondition(StatusUtils.buildWarningCondition("Cluster Label missing", "Resource lacks label '"
```
tomncooper, line 1056: This condition is a repeat of the exception string below (with slightly different wording). Could you not make a single error string and pass it to both the condition and the exception.

tomncooper, line 1056: The string `Cluster Label missing` *has* to match one of the strings in the `onNew` method and if you change it you have to remember to change it in multiple locations. These condition strings should be move to variables or enum options so that there is a single source of truth for them.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceAssemblyOperator.java
```diff
@@ -1069,7 +1077,9 @@ public MapAndStatus(T loadMap, K status) {
                         return Future.succeededFuture();
                     } else if (kafka.getSpec().getCruiseControl() == null) {
                         LOGGER.warnCr(reconciliation, "Kafka resource lacks 'cruiseControl' declaration : No deployed Cruise Control for doing a rebalance.");
-                        return updateStatus(reconciliation, kafkaRebalance, new KafkaRebalanceStatus(),
+                        KafkaRebalanceStatus status = new KafkaRebalanceStatus();
+                        status.addCondition(StatusUtils.buildWarningCondition("Cruise Control disabled", "Kafka resource lacks 'cruiseControl' declaration"));
```
tomncooper, line 1081: Again, this condition string `Cruise Control disabled` needs to moved to a variable/enum referenced here and in the `onNew` method. Again the condition, exception and log warning message should probably be the same.

tomncooper (changes requested):
A bit of refactoring needed.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceAssemblyOperator.java
```diff
@@ -1227,6 +1232,45 @@ private boolean hasRebalanceAnnotation(KafkaRebalance kafkaRebalance) {
...
+        return  rebalanceAnnotation == KafkaRebalanceAnnotation.refresh
+                || kafkaRebalance.getStatus().getConditions().stream().anyMatch(condition -> CCIssues.clusterLabelMissing.getReason().equals(condition.getReason()))
+                || kafkaRebalance.getStatus().getConditions().stream().anyMatch(condition -> CCIssues.cruiseControlDisabled.getReason().equals(condition.getReason()))
+                || kafkaRebalance.getStatus().getConditions().stream().anyMatch(condition -> CCIssues.connectException.getReason().equals(condition.getReason()));
+    }
+
+    /**
+     * Contains the issues that will be picked up in the periodic reconciliation, once corrected.
+     */
+    private enum CCIssues {
```
ppatierno, line 1245: I would be explicit as `CruiseControlIssues`

kyguy (commented):
Looking good @ShubhamRwt! I would just add a unit test and say it is good to go!

ShubhamRwt:
@ppatierno @kyguy Shouldn't we have a system test instead if a unit test test in this case?

ShubhamRwt:
Added a unit test now

kyguy (commented):
I think you are right, a system test in this case wouldn't hurt too!

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceAssemblyOperator.java
```diff
@@ -1227,6 +1232,45 @@ private boolean hasRebalanceAnnotation(KafkaRebalance kafkaRebalance) {
                 kafkaRebalance.getMetadata().getAnnotations().containsKey(ANNO_STRIMZI_IO_REBALANCE);
     }
 
+    private boolean shouldRenewRebalance(KafkaRebalance kafkaRebalance, KafkaRebalanceAnnotation rebalanceAnnotation) {
+        return  rebalanceAnnotation == KafkaRebalanceAnnotation.refresh
+                || kafkaRebalance.getStatus().getConditions().stream().anyMatch(condition -> CruiseControlIssues.clusterLabelMissing.getReason().equals(condition.getReason()))
```
tomncooper, line 1237: This feels very verbose. You could move this checking logic into the enum so that you have something like:
```java
CruiseControlIssues.clusterLabelMissing.check(kafkaRebalance.getStatus().getConditions())
```
But the issue is you are streaming all the conditions 3 times (and will add more, if you add more condition in future). It would be more efficient to do one pass over the conditions (in a good old fashioned for loop or a stream if you want to be fancy) and check if each condition's reason matches one of the list of CruiseControlIssues instances?

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceAssemblyOperator.java
```diff
@@ -723,7 +723,8 @@ public MapAndStatus(T loadMap, K status) {
                                                     KafkaRebalance kafkaRebalance,
                                                     KafkaRebalanceAnnotation rebalanceAnnotation,
                                                     RebalanceOptions.RebalanceOptionsBuilder rebalanceOptionsBuilder) {
-        if (rebalanceAnnotation == KafkaRebalanceAnnotation.refresh) {
+
+        if (shouldRenewRebalance(kafkaRebalance, rebalanceAnnotation)) {
             // The user has fixed the error on the resource and want to 'refresh'
```
tomncooper, line 728: Update this comment with the options for CC being activated when it was previously deactivated.

tomncooper (changes requested):
Also agree a system test is the best way to ensure this works and continues to work.

ShubhamRwt:
Added the system test

kyguy (approved):
Nice work @ShubhamRwt! Just need to fix some checkstyle errors

systemtest/src/test/java/io/strimzi/systemtest/cruisecontrol/CruiseControlST.java
```diff
@@ -312,4 +313,28 @@ void testCruiseControlIntraBrokerBalancingWithoutSpecifyingJBODStorage(Extension
...
+                .build());
+
+        KafkaRebalanceUtils.waitForKafkaRebalanceCustomResourceState(namespace, clusterName, KafkaRebalanceState.NotReady);
+
+        //Cruise Control spec is now enabled
+        KafkaResource.replaceKafkaResourceInSpecificNamespace(clusterName, kafka -> kafka.getSpec().setCruiseControl(new CruiseControlSpec()), namespace);
+
+        // The resource will move from `Not Ready` to `ProposalReady`
+        KafkaRebalanceUtils.waitForKafkaRebalanceCustomResourceState(namespace, clusterName, KafkaRebalanceState.ProposalReady);
+    }
```
im-konge, line 339: TBH, maybe you can put it into the `testCruiseControlWithRebalanceResourceAndRefreshAnnotation` -> just change the test a bit to save some time. 

Instead of how it is done now, you can just deploy Kafka without CC, deploy Rebalance, wait for `NotReady` state, put the CC to Kafka -> else should be same as it is now.

kyguy: Could it be even faster if we created the Rebalance resource first and then the Kafka cluster with CC enabled? Would this work?

im-konge: @kyguy good idea, it should work similar way, or?

kyguy: @im-konge This way (assuming we can create a KR resource before a existing Kafka cluster without problems) we save ourselves from having the brokers to be rolled to include the CC metric reporters. I'll leave that up to @ShubhamRwt to decide.

ShubhamRwt: @kyguy @im-konge If I remember correctly, if we apply a KR resource before a Kafka resource or cluster present, then probably we go into the `NotReady` state and stay there. Until or unless we apply a refresh annotation

ShubhamRwt: And the behavior I want to check here is that if cc is not enabled in Kafka resource then once it is enabled the resource will itself move into `Ready` state without the need of `refresh` annotation

im-konge: In both cases you don't have CC deployed, so I thought that it is looking for CC, not for Kafka cluster also. Anyway, you can still do the thing which I proposed above

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceAssemblyOperator.java
```diff
@@ -1227,6 +1231,51 @@ private boolean hasRebalanceAnnotation(KafkaRebalance kafkaRebalance) {
...
+
+        /**
+         * Cruise Control was not declared in the Kafka Resource
+         */
+        cruiseControlDisabled("Cruise Control disabled"),
+
+        /**
+         * Requests made to Cruise COntrol while it was starting
+         */
+        connectException("ConnectException");
```
ppatierno, line 1256: is this not used anywhere?

ShubhamRwt: This exception occurs only when we apply the KR resource and CC pod is started but not in ready state.

ShubhamRwt: But yeah Iam not generating this warning/exception myself

ppatierno: Yeah but I don't see this enum value to be used somewhere, is it really needed? If yes, let's use it.

ShubhamRwt: Yeah I guess this enum would be needed since when we enable the CC then the pods wouldn't become ready instantaneously and the status would contain `ConnectException` in that case and resource will become `NotReady`. So if we check that if the kafka rebalance status contains `ConnectException` then it need to handled in periodic reconciliation

ShubhamRwt: I did tried that, But I was not able to know where this `ConnectException` is being added to the status. TBH i I think this is an exception which is being added in status as a warning.

ShubhamRwt: The reason I made this enum value was because when we run the this method:
        public static boolean checkForMatch(List<Condition> conditions) {
            for (CruiseControlIssues issue : CruiseControlIssues.values()) {
                if (conditions.stream().anyMatch(condition -> issue.getReason().equals(condition.getReason()))) {
                    return true;
                }
            }
            return false;
        }

We need this value to see if our rebalance status contains this as reason

ShubhamRwt: We need this value to see if our rebalance status contains this as reason

ShubhamRwt: `CruiseControlIssues.values()` should contain the `ConnectException` enum value. so when we can compare it with the condition present in the rebalance status

ppatierno: But for the `ConnectException` to be in the rebalance status, it means someone puts it there by catching any connection exception to Cruise Control when it's not ready. I don't see any code building a rebalance status condition with such a condition, or? what am I missing?

ShubhamRwt: Maybe this method is giving the Connect Exception

ShubhamRwt:
```    private void httpExceptionHandler(Promise<? extends CruiseControlResponse> result, Throwable t) {
        if (t instanceof TimeoutException) {
            // Vert.x throws a NoStackTraceTimeoutException (inherits from TimeoutException) when the request times out
            // so we catch and raise a TimeoutException instead
            result.fail(new TimeoutException(t.getMessage()));
        } else if (t instanceof ConnectException) {
            // Vert.x throws a AnnotatedConnectException (inherits from ConnectException) when the request times out
            // so we catch and raise a ConnectException instead
            result.fail(new ConnectException(t.getMessage()));
        } else {
            result.fail(t);
        ```

Comment (ShubhamRwt) [reply cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceAssemblyOperator.java:1256]: `_But for the ConnectException to be in the rebalance status, it means someone puts it there by catching any connection exception to Cruise Control when it's not ready. I don't see any code building a rebalance status condition with such a condition, or? what am I missing?_` Yeah that's exactly I was trying to figure out but was unable to find out. I was just able to see the warning in the status

File: systemtest/src/test/java/io/strimzi/systemtest/cruisecontrol/CruiseControlST.java [hunk 10]
```diff
@@ -136,16 +137,21 @@ void testCruiseControlWithApiSecurityDisabled(ExtensionContext extensionContext)
...
+        resourceManager.createResource(extensionContext, KafkaTemplates.kafkaEphemeral(clusterName, 3, 3)
+                .editMetadata()
+                .withNamespace(namespace)
+                .endMetadata()
+                .build());
+        resourceManager.createResource(extensionContext, false,  KafkaRebalanceTemplates.kafkaRebalance(clusterName)
+                .editMetadata()
+                .withNamespace(namespace)
+                .endMetadata()
+                .build());
```
im-konge, line 149: Indent in both methods please.

systemtest/src/test/java/io/strimzi/systemtest/cruisecontrol/CruiseControlST.java
```diff
@@ -136,16 +137,21 @@ void testCruiseControlWithApiSecurityDisabled(ExtensionContext extensionContext)
...
+
+        KafkaRebalanceUtils.waitForKafkaRebalanceCustomResourceState(namespace, clusterName, KafkaRebalanceState.NotReady);
+
+        //Cruise Control spec is now enabled
```
im-konge, line 153:
```suggestion
        // Cruise Control spec is now enabled
```

systemtest/src/test/java/io/strimzi/systemtest/cruisecontrol/CruiseControlST.java
```diff
@@ -136,16 +137,21 @@ void testCruiseControlWithApiSecurityDisabled(ExtensionContext extensionContext)
...
+        KafkaResource.replaceKafkaResourceInSpecificNamespace(clusterName, kafka -> kafka.getSpec().setCruiseControl(new CruiseControlSpec()), namespace);
```
im-konge, line 154: after this line you should wait for Kafka rolling update and then go through the rebalancing process

systemtest/src/test/java/io/strimzi/systemtest/cruisecontrol/CruiseControlST.java
kyguy: > If I remember correctly, if we apply a KR resource before a Kafka resource or cluster present, then probably we go into the NotReady state and stay there. Until or unless we apply a refresh annotation

@ShubhamRwt  not sure what happens with the reconciliation when a topic or user CR is created before the Kafka cluster is deployed but I imagine we would want consistent behavior for the KafkaRebalance resource. Let me know what you think! 

As for what the system test tests, I'll leave that too you!

systemtest/src/test/java/io/strimzi/systemtest/cruisecontrol/CruiseControlST.java
ShubhamRwt: `KafkaUtils.waitForKafkaReady(namespace, clusterName);` you mean using something like this ?

im-konge: Before replacing the resource, I would pick the pod snapshots

```
 Map<String, String> zkPods = PodUtils.podSnapshot(namespaceName, zkSelector);
 Map<String, String> kafkaPods = PodUtils.podSnapshot(namespaceName, kafkaSelector);
 Map<String, String> eoPod = DeploymentUtils.depSnapshot(namespaceName, KafkaResources.entityOperatorDeploymentName(clusterName));
 ```
 then, after the replace, I would add something like this:
 ```
 RollingUpdateUtils.waitTillComponentHasRolled(namespaceName, zkSelector, 3, zkPods);
 RollingUpdateUtils.waitTillComponentHasRolled(namespaceName, kafkaSelector, 3, kafkaPods);
 DeploymentUtils.waitTillDepHasRolled(namespaceName, KafkaResources.entityOperatorDeploymentName(clusterName), 1, eoPod);
 ```
 IIRC the whole Kafka cluster should be rolled, that's why you need to check all the rolling updates.

**Review by im-konge — APPROVED** [review]
LGTM - the test part :)

File: cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceAssemblyOperator.java (hunk 9 above)
Comment (tomncooper) [reply cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceAssemblyOperator.java:1256]: What was the outcome of this conversation?

Comment (tomncooper) [reply cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceAssemblyOperator.java:1256]: From what I can see this Exception is in the enum so that the KRCR will be continually be put into `New` state every time it is seen, so that it will keep trying to reconcile until CC appears?

Comment (ShubhamRwt) [reply cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceAssemblyOperator.java:1256]: Yes. I guess it would be required because when we bring cruise control up it will take time to restart and if requests are made during that point of time then we will move to 'NotReady' state due to the `ConnectException`

File: cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceAssemblyOperator.java [hunk 13]
```diff
@@ -1227,6 +1231,49 @@ private boolean hasRebalanceAnnotation(KafkaRebalance kafkaRebalance) {
                 kafkaRebalance.getMetadata().getAnnotations().containsKey(ANNO_STRIMZI_IO_REBALANCE);
     }
 
+    private boolean shouldRenewRebalance(KafkaRebalance kafkaRebalance, KafkaRebalanceAnnotation rebalanceAnnotation) {
+        // check if request are made when CC is not active
+        if (kafkaRebalance.getStatus().getConditions().stream().anyMatch(condition -> "ConnectException".equals(condition.getReason()))) {
+            return true;
+        }
+        return  rebalanceAnnotation == KafkaRebalanceAnnotation.refresh || CruiseControlIssues.checkForMatch(kafkaRebalance.getStatus().getConditions());
```
tomncooper, line 1239: Can we split these into separate branches, it would improve readability

ShubhamRwt: Done

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceAssemblyOperator.java
```diff
@@ -1227,6 +1231,52 @@ private boolean hasRebalanceAnnotation(KafkaRebalance kafkaRebalance) {
                 kafkaRebalance.getMetadata().getAnnotations().containsKey(ANNO_STRIMZI_IO_REBALANCE);
     }
 
+    private boolean shouldRenewRebalance(KafkaRebalance kafkaRebalance, KafkaRebalanceAnnotation rebalanceAnnotation) {
+        // check if request are made when CC is not active
+        if (kafkaRebalance.getStatus().getConditions().stream().anyMatch(condition -> "ConnectException".equals(condition.getReason()))) {
+            return true;
+        } else if (rebalanceAnnotation == KafkaRebalanceAnnotation.refresh) {
```
ppatierno, line 1238: why this is not in OR with the previous one? They both return `true`.

ppatierno (changes requested):

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceAssemblyOperator.java
ShubhamRwt: Done @ppatierno

ppatierno (approved):

tomncooper (approved):

scholzj:
@ShubhamRwt Can you please rebase this? Thanks.

ShubhamRwt:
Done @scholzj

PR #6582, opened 2022-03-28
 Refactored and fixed the mechanism for paused reconciliations

ShubhamRwt:

### Type of change

_Select the type of your PR_

- Bugfix
- Refactoring

### Description

This PR refactors the implementation of reconciliations paused mechanism and also fixes the bug in its working. Earlier when the resource was `paused`, the state was set to `New` while the correct behaviour should be that state stays as ` ReconciliationPaused` and stays like it until the resource is unpaused.

### Checklist

- [ ] Write tests
- [ ] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceAssemblyOperator.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceAssemblyOperator.java
@@ -363,17 +363,6 @@ public class KafkaRebalanceAssemblyOperator
 
         LOGGER.infoCr(reconciliation, "Rebalance action from state [{}]", currentState);
 
-        if (Annotations.isReconciliationPausedWithAnnotation(kafkaRebalance)) {
-            // we need to do this check again because it was triggered by a watcher
-            KafkaRebalanceStatus status = new KafkaRebalanceStatus();
-
-            Set<Condition> unknownAndDeprecatedConditions = validate(reconciliation, kafkaRebalance);
-            unknownAndDeprecatedConditions.add(StatusUtils.getPausedCondition());
-            status.setConditions(new ArrayList<>(unknownAndDeprecatedConditions));
-
-            return updateStatus(reconciliation, kafkaRebalance, status, null).compose(i -> Future.succeededFuture());
-        }
-
         if (kafkaRebalance.getSpec().isRebalanceDisk() && !usingJbodStorage) {
             String error = "Cannot set rebalanceDisk=true for Kafka clusters with a non-JBOD storage config. " +
                 "Intra-broker balancing only applies to Kafka deployments that use JBOD storage with multiple disks.";
@@ -385,6 +374,9 @@ public class KafkaRebalanceAssemblyOperator
 
         return computeNextStatus(reconciliation, host, apiClient, kafkaRebalance, currentState, rebalanceAnnotation, rebalanceOptionsBuilder)
            .compose(desiredStatusAndMap -> {
+               if (Annotations.isReconciliationPausedWithAnnotation(kafkaRebalance)) {
+                   return updateStatus(reconciliation, kafkaRebalance, desiredStatusAndMap.status, null).compose(i -> Future.succeededFuture());
+               }
                // More events related to resource modification might be queued with a stale state. (potentially updated by the rebalance holding the lock)
                // Due to possible long rebalancing operations that take the lock for the entire period,
                // do a new get to retrieve the current resource state.
@@ -455,6 +447,8 @@ public class KafkaRebalanceAssemblyOperator
             case Ready:
                 // Rebalance Complete
                 return onReady(reconciliation, host, apiClient, kafkaRebalance, rebalanceAnnotation, rebalanceOptionsBuilder);
+            case ReconciliationPaused:
+                return onReconciliationPaused(reconciliation, host, apiClient, kafkaRebalance, rebalanceAnnotation, rebalanceOptionsBuilder);
             case NotReady:
                 // Error case
                 return onNotReady(reconciliation, host, apiClient, kafkaRebalance, rebalanceAnnotation, rebalanceOptionsBuilder);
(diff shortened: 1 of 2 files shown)


cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceAssemblyOperator.java
```diff
@@ -385,6 +374,9 @@ private String rebalanceStateConditionType(KafkaRebalanceStatus status) {
 
         return computeNextStatus(reconciliation, host, apiClient, kafkaRebalance, currentState, rebalanceAnnotation, rebalanceOptionsBuilder)
            .compose(desiredStatusAndMap -> {
+               if (Annotations.isReconciliationPausedWithAnnotation(kafkaRebalance)) {
+                   return updateStatus(reconciliation, kafkaRebalance, desiredStatusAndMap.status, null).compose(i -> Future.succeededFuture());
+               }
```
scholzj, line 379: The reconciliation being paused is not question of the status only. The Operator should not do anything when it is paused. So does it really make sense to do this only after the `computeNextStatus` etc.?

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceAssemblyOperator.java
```diff
@@ -363,17 +363,6 @@ private String rebalanceStateConditionType(KafkaRebalanceStatus status) {
...
-            // we need to do this check again because it was triggered by a watcher
-            KafkaRebalanceStatus status = new KafkaRebalanceStatus();
-
-            Set<Condition> unknownAndDeprecatedConditions = validate(reconciliation, kafkaRebalance);
-            unknownAndDeprecatedConditions.add(StatusUtils.getPausedCondition());
-            status.setConditions(new ArrayList<>(unknownAndDeprecatedConditions));
-
-            return updateStatus(reconciliation, kafkaRebalance, status, null).compose(i -> Future.succeededFuture());
-        }
-
```
kyguy, line 376: If this was kept (as mentioned on the comment below), would we still need the `onReconciliationPaused()` method? Is the purpose of `onReconciliationPaused()` to retain the ConfigMap of the previous state?

kyguy (commented):
Looks good @ShubhamRwt, just had a small question! 

Also there is a typo in the title "Refqcotored" -> "Refactored"!

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceAssemblyOperator.java
ShubhamRwt: I guess maybe we can get rid of the --> `onReconciliationPaused()` and handle the case for `ReconciliationPaused` in the `computeNextStatus` method itself

ShubhamRwt: case ReconciliationPaused:
                return requestRebalance(reconciliation, host, apiClient, kafkaRebalance, true, rebalanceOptionsBuilder);

ShubhamRwt: Something like this maybe. What do you think @kyguy?

ShubhamRwt: I guess we need to handle it because we don't want the state to be shown as `New` when `reconciliation is paused` (previous behaviour). It should be `ReconciliationPaused`

kyguy: Looks good to me!

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceAssemblyOperator.java
tomncooper: As I said before and as Jakub points out, we need to avoid doing any work we don't need too when in the paused state. So we should avoid triggering `computeNextStatus` if we are paused. 

I think the state was set to New when pauses so that, when the pause annotation was removed, the Kafka Rebalance CR would be reconciled automatically without any further intervention.

ShubhamRwt: The issue is that once the reconciliation is unpaused --> we stay in the `New` state and I guess no matter on which state we pause our resource will always continue form `Proposal Ready` state

ShubhamRwt: This is because one the resource is unpaused then the current state is `New` and then the next steps will happen accordingly

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceAssemblyOperator.java
```diff
@@ -1108,8 +1116,10 @@ public MapAndStatus(T loadMap, K status) {
                                             KafkaRebalanceStatus kafkaRebalanceStatus = currentKafkaRebalance.getStatus();
                                             KafkaRebalanceState currentState;
                                             // cluster rebalance is new or it is in one of the others states
-                                            if (kafkaRebalanceStatus == null || kafkaRebalanceStatus.getConditions().stream().filter(cond -> "ReconciliationPaused".equals(cond.getType())).findAny().isPresent()) {
+                                            if (kafkaRebalanceStatus == null) {
                                                 currentState = KafkaRebalanceState.New;
+                                            } else if (kafkaRebalanceStatus.getConditions().stream().filter(cond -> "ReconciliationPaused".equals(cond.getType())).findAny().isPresent()) {
```
tomncooper, line 1121: Isn't the string `"ReconciliationPaused"` referenced in an enum somewhere? I get nervous with these not being a constants.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceAssemblyOperator.java
```diff
@@ -733,6 +735,15 @@ public MapAndStatus(T loadMap, K status) {
         }
     }
 
+
+    private Future<MapAndStatus<ConfigMap, KafkaRebalanceStatus>> onReconciliationPaused(Reconciliation reconciliation,
```
ppatierno, line 739: Add Javadoc as other methods

ShubhamRwt: Added the docs

ShubhamRwt:
> Looks awesome @ShubhamRwt! Should we add some tests for this too?

@kyguy Do you mean system test or unit test?

kyguy:
@ShubhamRwt I think a unit test would suffice here

ppatierno:
I don't remember the specific details after so long time but I guess that when designing the state-machine we decided that a new "paused" state was not needed but the `New` one was enough (I think it's raised by @tomncooper in another comment here as well). Why do you see the need for a new state? Is there any bug/issue you are trying to fix with this PR?

ppatierno:
So it doesn't sound a real bug but just how it's logged? Can we print a more useful log getting the fact that state is `New` but we asked for pause via the annotation instead of having a new state? It's possible I am not taking some specific reasons into account on why a specific state is needed but I remember that we decided it was not necessary in the past. @tomncooper do you remember that?

ppatierno (approved):

tomncooper (approved):

kyguy (approved):

scholzj (approved):

PR #6584, opened 2022-03-28
Better attribution of Apache marks

tombentley:
See https://www.apache.org/foundation/marks/faq/#attribution


scholzj (approved):
You will need to fix the DCO.

ppatierno (approved):

PR #6588, opened 2022-03-29
[MO] - [system test] -> SecurityST race condition fixes and correct R…

see-quick:
…ollingUpdates

### Type of change

- Bugfix

### Description

This PR fixes the functionality of some security system tests (introduced in #4369). Specifically:

1. testClusterCACertRenew = this test case is checked after Cluster CA renews the Kafka Pods rolls. This doesn't seem right because Cluster Operator must perform a Rolling Update of ZooKeeper, Kafka and Entity Operators pods. 
2. testClientsCACertRenew = in this test case, we checked only Entity Operator, which is again wrong! The correct way is that Cluster Operator must perform a Rolling Update of Kafka and then Entity Operator pods

This also applies to the testCustomClusterCACertificateRenew test case in CustomCaST.

Many of you also face the issue with the regression pipeline (f.e., these tests cases are flaky). Because we only checked the Kafka or Entity Pods (before ZooKeeper or Kafka pods), which results in TimeoutException. This PR eliminates such flakiness.

### Checklist


see-quick:
Build passed :), it's some UI error of Github...
image

PR #6593, opened 2022-03-29
Refactor Cruise Control reconciliation out of the `KafkaAssemblyOperator` class

scholzj:
### Type of change

- Refactoring

### Description

This is another step in the `KafkaAssemblyOperator` refactoring. This PR handles the Cruise Control. It moves the reconciliation and all its state out from `KafkaAssemblyOperator` into a new class `CruiseControlReconciler`. It also does the usual clean-up of the `CruiseControl` model class (removing unnecessary setters / getters, static methods calling another static methods etc.). It also tries to create more clear separation between the classes related to the KafkaRebalance recocniliatin and the `CruiseControl` model => the model should be mainly about the actual Cruise Control deployment and its details should not be needed for KafkaRebalance reconciliation => as a result some fields and methods were moved.

### Checklist

- [x] Write tests


ppatierno (approved):
LGTM. Just one nit.

PR #6599, opened 2022-03-30
Moved to use Clock to check certificate expiration

### LINKED ISSUE #6595: Move from `System.currentTimeMillis()` to `Clock.mills()` for checking CA certificate needs renewal
When the https://github.com/strimzi/strimzi-kafka-operator/pull/6580 will be merged, we should align the `Ca.certNeedsRenewal` to not using the `System.currentTimeMillis()` anymore but the `Clock.mills()` which has the same purpose.
This is going to be useful even for testing, because by mocking the `Clock`, it's possible to get the time we want and checking that the "needs renewal" feature works correctly. Something not simple to test today due to the fact that the current system time is used.

ppatierno:
### Type of change

- Refactoring

### Description

This PR fixes #6595 by using the `Clock` instead of the `System.currentTimeMillis()` to check certificate expiration by changing the code to do that as well.

### Checklist

- [x] Write tests


scholzj (approved):
Looks like the build failed -> but LGTM apart from that.

ppatierno:
Fixed! Damn checkstyle!

PR #6604, opened 2022-03-30
[MO] - [system test] -> use scraper pod instead of kafka clients to c…

see-quick:
…all KC api

### Type of change

- Bugfix

### Description

This PR fixes problem with the following test cases:
1. testRackAwareConnectWrongDeployment
2. testRackAwareConnectCorrectDeployment

The problem is that currently`kafkaClientsPodName` **can't** call KafkaConnect API and only our `Scraper Pod` is allowed to do so. In other words, these test cases always fail because KafkaClients pods are not allowed to call KafkaConnect API.

### Checklist

### DIFF
--- a/systemtest/src/test/java/io/strimzi/systemtest/specific/SpecificIsolatedST.java
+++ b/systemtest/src/test/java/io/strimzi/systemtest/specific/SpecificIsolatedST.java
@@ -6,6 +6,8 @@ package io.strimzi.systemtest.specific;
 
 import io.fabric8.kubernetes.api.model.Affinity;
 import io.fabric8.kubernetes.api.model.Event;
+import io.fabric8.kubernetes.api.model.LabelSelector;
+import io.fabric8.kubernetes.api.model.LabelSelectorBuilder;
 import io.fabric8.kubernetes.api.model.NodeAffinity;
 import io.fabric8.kubernetes.api.model.NodeSelectorRequirement;
 import io.fabric8.kubernetes.api.model.Pod;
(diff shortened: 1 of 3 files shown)


systemtest/src/test/java/io/strimzi/systemtest/specific/SpecificIsolatedST.java
```diff
@@ -334,7 +342,14 @@ void testRackAwareConnectCorrectDeployment(ExtensionContext extensionContext) {
         assertThat(connectPodNodeSelectorRequirement.getKey(), is(rackKey));
         assertThat(connectPodNodeSelectorRequirement.getOperator(), is("Exists"));
 
-        KafkaConnectUtils.sendReceiveMessagesThroughConnect(connectPodName, topicName, kafkaClientsPodName, clusterOperator.getDeploymentNamespace(), clusterName);
+        // fetch scraper Pod name
+        final LabelSelector scraperLabelSelector = new LabelSelectorBuilder()
+            .addToMatchLabels(Constants.SCRAPER_LABEL_KEY, Constants.SCRAPER_LABEL_VALUE)
+            .build();
+        final String scraperPodName = kubeClient(clusterOperator.getDeploymentNamespace())
+            .listPods(clusterOperator.getDeploymentNamespace(), scraperLabelSelector).get(0).getMetadata().getName();
```
Frawless, line 350: don't we have a method for that? if not, shouldn't we have one? I think there will be more occurrences of this.

see-quick: We don't but I added it. )

PR #6605, opened 2022-03-30
[MO] - [system test] -> concurrent exception deletion mulitple namesp…

see-quick:
…aces at once

### Type of change

- Bugfix

### Description

This PR fixes problem using `ConcurrentModificationException`, where we delete multiple namespaces created by Cluster Operator. We resolve such a problem by instantiating a new map which will eliminate this exception.

### Checklist

### DIFF
--- a/systemtest/src/main/java/io/strimzi/systemtest/parallel/TestSuiteNamespaceManager.java
+++ b/systemtest/src/main/java/io/strimzi/systemtest/parallel/TestSuiteNamespaceManager.java
@@ -19,6 +19,7 @@ import java.io.File;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Locale;
 import java.util.Map;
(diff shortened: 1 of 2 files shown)


scholzj (commented):
I'm not sure I understand the issue and the solution. You tak about `ConcurrentModificationException` => but that is something you would normally solve with some concurrent map or some synchronization. But I'm not sure I see nay of that here.

see-quick:
The problem was in 

> I'm not sure I understand the issue and the solution. You tak about `ConcurrentModificationException` => but that is something you would normally solve with some concurrent map or some synchronization. But I'm not sure I see nay of that here.

I know this problem is not obvious at first glance, but we deleted namespaces wrongly. In (https://github.com/strimzi/strimzi-kafka-operator/pull/6605/files#diff-885fe3fd1158ac2dc6f2d1f828f5a2eb4d8d70380e4a2cce244e3e75950c7553L174-L182), we fetch the reference of HashMap for which we iterate. But in (https://github.com/strimzi/strimzi-kafka-operator/pull/6605/files#diff-02ba9aff32b6ba24508f0f82ed1e786b931df96a6523a4b85108677d45946c15L395) we use the same reference and deletes item (during iteration), which results in `ConcurrentModificationException`. I have created a new set to create a new reference and thus eliminate this exception.

scholzj (approved):
Ok then. Should be reviewed. by some SME (@Frawless) as well.

Frawless (approved):

PR #6616, opened 2022-04-01
Clarify that podDisruptionBudget=0 is set so that Kube does not interfere with Drain Cleaner

emmanuelbernard:
Signed-off-by: Emmanuel Bernard <emmanuel@hibernate.org>

### Type of change

- Documentation

### Description

Clarify that the pod disruption budget set to zero is to avoid conflict between the drain cleaner work and the kube scheduler.

CC @PaulRMellor


documentation/assemblies/managing/assembly-drain-cleaner.adoc
```diff
@@ -8,7 +8,8 @@
 [role="_abstract"]
 Kafka and ZooKeeper pods might be evicted during Kubernetes upgrades, maintenance or pod rescheduling.
 If your Kafka broker and ZooKeeper pods were deployed by Strimzi, you can use the Strimzi Drain Cleaner tool to handle the pod evictions.
-You need to set the `podDisruptionBudget` for your Kafka deployment to `0` (zero) for the Strimzi Drain Cleaner to work.
+Since the Strimzi Drain Cleaner will handle the eviction instead of Kubernetes, you need to set the `podDisruptionBudget` for your Kafka deployment to `0` (zero).
+Kubernetes will then no longer interfere.
```
PaulRMellor, line 12:
```suggestion
Kubernetes will then no longer be involved in the eviction process.
```

documentation/modules/managing/proc-drain-cleaner-deploying.adoc
```diff
@@ -86,7 +86,7 @@ spec:
 ----
 +
 Reducing the maximum pod disruption budget to zero prevents Kubernetes from automatically evicting the pods in case of voluntary disruptions,
-so pods must be evicted by the Strimzi Drain Cleaner.
+leaving the floor for pods to be evicted by the Strimzi Drain Cleaner without conflict.
```
PaulRMellor, line 89:
```suggestion
leaving the Strimzi Drain Cleaner free to evict pods without conflict.
```

PaulRMellor (approved):
Couple of suggestions. I thought _interfere_ was maybe a bit strong :slightly_smiling_face:

documentation/modules/managing/proc-drain-cleaner-deploying.adoc
fvaleri: It is the joint work of Drain Cleaner, Cluster Operator and K8s Scheduler to actually move pods away from the cordoned node.

documentation/assemblies/managing/assembly-drain-cleaner.adoc
fvaleri: Kubernetes will still try to evict, but receiving an error because of the PDB settings. For that reason I think "no longer be able to evict pods" is better.

scholzj: I think saying something like

```suggestion
Kubernetes will then no longer be allowed to evict the pod on its own.
```

might be better.

documentation/modules/managing/proc-drain-cleaner-deploying.adoc
scholzj: Yeah, I think Federico has a point. Maybe something like

```suggestion
leaving the Strimzi Drain Cleaner and Strimzi Cluster Operator to roll the pod which will be scheduled by Kubernetes on a different worker node.
```

emmanuelbernard:
Thanks, I've applied your suggestions, squashed and rebased off main.

PR #6627, opened 2022-04-05
Added removal of older cluster CA certificates after renewal by new CA private key

### LINKED ISSUE #1979: [Bug] Remove the old certificates after a renewal of Cluster CA
Currently, when the CA is renewed the old certificates remain in the secrets and in use until they expire. This potential security risk. We should clear the certificate once it is not needed (ie. once all components were moved to new certificates). 

This is not so easy with Clients CA where Strimzi does not control the clients connecting based on it. However, even in that case we should at least add a warning to the docs that users should remove the keys. This warning needs to be especially in the docs around the forced renewal which might be cause by the original certificates being compromised.

ppatierno:
### Type of change

- Bugfix

### Description

This PR fixes #1979.
As usual, when a new private cluster CA key is generated, the cluster CA renewal happens and the older one is stored in the corresponding Secret as `ca-<expiration_date_time>.crt`.
The operator runs two rollings: the first one to trust the new cluster CA cert, the second one to generate the new server certs signed by the new cluster CA cert.
With this PR a third rolling is added after checking that a removal of the older CA cert happened.
This check is based on the fact that the first two rolling happened and all Kafka components related pods have the cluster CA cert generation equals to the one of the new cluster CA cert. After that, the older cluster CA cert can be removed.
The effect of such a removal is a new rolling of the pods that anyway happens on the next reconcile loop.

### Checklist

- [x] Write tests

### DIFF
--- a/operator-common/src/main/java/io/strimzi/operator/cluster/model/Ca.java
+++ b/operator-common/src/main/java/io/strimzi/operator/cluster/model/Ca.java
@@ -47,6 +47,8 @@ import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
 import java.util.function.Function;
+import java.util.function.Predicate;
+import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 
 import static java.time.temporal.ChronoField.DAY_OF_MONTH;
@@ -66,6 +68,10 @@ import static java.util.Collections.singletonMap;
 public abstract class Ca {
 
     protected static final ReconciliationLogger LOGGER = ReconciliationLogger.create(Ca.class);
+
+    private static final Pattern OLD_CA_CERT_PATTERN = Pattern.compile(
+            "^ca-\\d{4}-\\d{2}-\\d{2}T\\d{2}-\\d{2}-\\d{2}Z.crt$"
+    );
 
     private static final DateTimeFormatter DATE_TIME_FORMATTER = new DateTimeFormatterBuilder()
             .appendValue(YEAR, 4, 10, SignStyle.EXCEEDS_PAD)
@@ -557,7 +563,7 @@ public abstract class Ca {
                         addCertCaToTrustStore(CA_CRT, certData);
                     }
             }
-            this.caCertsRemoved = removeExpiredCerts(certData) > 0;
+            this.caCertsRemoved = removeCerts(certData, this::removeExpiredCert) > 0;
         }
         SecretCertProvider secretCertProvider = new SecretCertProvider();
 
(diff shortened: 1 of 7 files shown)


operator-common/src/main/java/io/strimzi/operator/cluster/model/Ca.java
```diff
@@ -794,32 +800,66 @@ public int keyGeneration() {
...
+     */
+    public void maybeDeleteOldCerts() {
+        this.caCertsRemoved = removeCerts(this.caCertSecret.getData(), entry -> OLD_CA_CERT_PATTERN.matcher(entry.getKey()).matches()) > 0;
+        if (this.caCertsRemoved) {
+            LOGGER.infoCr(reconciliation, "{}: Old CA certificates removed", this);
+            /*
+            this.caCertSecret = new SecretBuilder(this.caCertSecret)
+                    .withData(this.caCertSecret.getData())
+                    .build();
+             */
```
scholzj, line 815: Is this some left over which can be deleted?

scholzj (commented):
One nit, LGTM otherwise. Have you tried the regression tests? They might need some updating for the triple rooling update as well.

operator-common/src/main/java/io/strimzi/operator/cluster/model/Ca.java
ppatierno: Yeah :-(

ppatierno:
I will wait for the build, then I will start regression here to check if something is failing. Maybe it will not but anyway some tests could need some update to wait for the third rolling before exiting with success.

scholzj:
> I will wait for the build, then I will start regression here to check if something is failing. Maybe it will not but anyway some tests could need some update to wait for the third rolling before exiting with success.

Its a bit tricky ... since they might not fail but stop testing after 2 rolls.

ppatierno:
> Its a bit tricky ... since they might not fail but stop testing after 2 rolls.

Yeah exactly what I said. Anyway I will check them locally. I wanted just to use the AZ pipeline to see if something else was broken :-)

ppatierno:
@im-konge I have changed the STs related to the CA key replacement, can you take a look please?
@scholzj same for you thanks! I will start regression tests as soon as you both are ok and the current build ends.

im-konge (approved):

scholzj (approved):

ppatierno:
@scholzj it seems that the failures are not related to my changes. @see-quick is working to fix them via https://github.com/strimzi/strimzi-kafka-operator/pull/6652
I would say we can merge this one, is it fine with you?

PR #6632, opened 2022-04-05
Move ZooKeeper reconciliation into separate classes

scholzj:
### Type of change

- Refactoring

### Description

This PR refactors the KafkaAssemblyOperator and moves ZooKeeper reconciliation an its state into a separate class `ZooKeeperRecocniler`. Unlike the previous refactorings for EO, KE or CC, the ZooKeeper and Kafka reconciliation have a lot more in common including some shared code. To minimize the code duplication and simplify testing, this Pr moves the shared code into several additional classes:
* `PvcReconciler` for reconciliation, resizing and deletion of PVCs
* `ManualPodCleaner` for manual pod and PVC cleanups triggered by annotation
* `ReconcilerUtils` with some shared functions such as pod readiness checks, loading of client secrets or creating the lists with restart reasons

It also does some smaller improvements to the `ZookeeperCluster` model and splits the `KafkaSpecChecker` into `KafkaSpecChecker` and `ZooKeeperSpecChecker`.

There are some further refactorings which can be done but which I didn't wanted to do to not make this PR even bigger. For example simplifying the big constructor of the `ZooKeeperReconciler` and use only the smaller one which is used only in tests today etc. But these would just make it harder to review and can be done later as well.

### Checklist

- [x] Write tests


scholzj:
> Love this, the KAO has been too big for too long!

Thanks. The only problem is that even after all of this it still has over 3000 lines ... still more work to do I guess :-o

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/ClusterCa.java
```diff
@@ -151,23 +151,20 @@ public Secret cruiseControlSecret() {
             isMaintenanceTimeWindowsSatisfied);
     }
 
-    public Map<String, CertAndKey> generateZkCerts(Kafka kafka, boolean isMaintenanceTimeWindowsSatisfied) throws IOException {
-        String cluster = kafka.getMetadata().getName();
-        String namespace = kafka.getMetadata().getNamespace();
-
-        DnsNameGenerator zkDnsGenerator = DnsNameGenerator.of(namespace, KafkaResources.zookeeperServiceName(cluster));
-        DnsNameGenerator zkHeadlessDnsGenerator = DnsNameGenerator.of(namespace, KafkaResources.zookeeperHeadlessServiceName(cluster));
+    public Map<String, CertAndKey> generateZkCerts(String namespace, String kafkaName, int replicas, boolean isMaintenanceTimeWindowsSatisfied) throws IOException {
```
ppatierno, line 154: why did you prefer to pass `namespace` and `kafkaName` instead the current only `Kafka` instance?
are we using `kafkaName` in any other places? or just `name` or `clusterName`?

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaCluster.java
```diff
@@ -94,12 +94,12 @@
 import java.util.function.Function;
 import java.util.stream.Collectors;
 
+import static io.strimzi.operator.cluster.model.CruiseControl.CRUISE_CONTROL_METRIC_REPORTER;
 import static io.strimzi.operator.cluster.model.ListenersUtils.isListenerWithCustomAuth;
 import static io.strimzi.operator.cluster.model.ListenersUtils.isListenerWithOAuth;
 import static java.util.Collections.addAll;
 import static java.util.Collections.singletonList;
 import static java.util.Collections.singletonMap;
-import static io.strimzi.operator.cluster.model.CruiseControl.CRUISE_CONTROL_METRIC_REPORTER;
```
ppatierno, line 102: the changes here are just about imports order ... but yours make more sense :-)

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/ZooKeeperSpecChecker.java
```diff
@@ -0,0 +1,67 @@
...
+ */
+public class ZooKeeperSpecChecker {
+    private final ZookeeperCluster zoo;
+
+    /**
+     * @param zoo The model generated based on the spec. This is requested so that default
+     *                     values not included in the spec can be taken into account, without needing
+     *                     this class to include awareness of what defaults are applied.
+     */
+    public ZooKeeperSpecChecker(ZookeeperCluster zoo) {
```
ppatierno, line 27: maybe we are using `zk` somewhere else instead of `zoo`

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/ClusterCa.java
scholzj: Because the Kafka instance is _big_ and it would need to be stored through the reconciliation (today we use it only on the beginning to create the model, so there was not other reason to keep it in the state). And at the same time, it turned out it wasn't really needed.

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaCluster.java
scholzj: Sorry, I tried to not change the imports order. But I guess IDEA sometimes changes it when you organize imports to remove unused things. I can revert it if you want (I'm just not sure you want it reverted if you say it makes more sense this way :-o)

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/ZooKeeperSpecChecker.java
scholzj: I think I used `zoo` on most new things I wrote here. I can rename them to `zk` if you want?

ppatierno (approved):

PR #6637, opened 2022-04-06
feat: Allow users to configure Cruise Control topic names

katheris:
### Type of change

- Enhancement / new feature

### Description

Allow users to configure the names of the Cruise Control topics. If the user changes the name in an existing deployment the old topics are left for the user to remove manually.

### Checklist

- [x] Write tests
- [x] Update documentation
- [ ] Update CHANGELOG.md


scholzj (commented):
I think this is incomplete. It does not cover the configuration of the CC metrics reporter.

ppatierno:
> With your PR I would expect that your CC never gets any metrics when you change the topic name and cannot work properly.

Yeah, Jakub is right. Metrics reporter (running on Kafka) will keep sending metrics on the "hard-coded" topic but CC will get metrics on the new name topic so not working properly in the end.

katheris:
@scholzj @ppatierno I think this change should fix it. The fromCrd function in the CruiseControl model class does already contain logic that applies the default configs and then adds the user config, however that function requires many other arguments and didn't really feel like the right place. So I've made the KafkaBrokerConfigurationBuilder check for user defined config itself and if not take the default. I figured for this change it is simpler and perhaps in future if we have other configs that need checking in that class we can update to have a more central mechanism, perhaps in CruiseControlConfigurationParameters. What do you think?

I have tested with custom topics, left it running for a while and created a RebalanceCR and no other metric topic was created and the proposal was created successfully so this seems to be the only other place that was referencing it.

ppatierno:
@katheris yeah agree. I think that this solution works fine.

> I have tested with custom topics, left it running for a while and created a RebalanceCR and no other metric topic was created and the proposal was created successfully so this seems to be the only other place that was referencing it.

You mean no other metrics created on the old topic but created on the new one?

ppatierno (approved):

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaBrokerConfigurationBuilder.java
```diff
@@ -128,7 +130,10 @@ public KafkaBrokerConfigurationBuilder withCruiseControl(String clusterName, Cru
                                                              String replicationFactor, String minInSyncReplicas)   {
         if (cruiseControl != null) {
             printSectionHeader("Cruise Control configuration");
-            writer.println(CruiseControlConfigurationParameters.METRICS_TOPIC_NAME + "=strimzi.cruisecontrol.metrics");
+            Map<String, Object> config = cruiseControl.getConfig();
+            String metricsTopicName = Optional.ofNullable(config.get(CruiseControlConfigurationParameters.CRUISE_CONTROL_METRIC_REPORTER_TOPIC_NAME.getValue()))
+                            .orElse(CruiseControlConfigurationParameters.DEFAULT_METRIC_REPORTER_TOPIC_NAME).toString();
+            writer.println(CruiseControlConfigurationParameters.METRICS_TOPIC_NAME + "=" + metricsTopicName);
```
scholzj, line 136: Can't you use `getOrDefault` instead of `Optional.ofNullable`? I think it would be more readable. Also, IIRC, the `ofNullable` does not catch NPEs => so you need to check if `cruiseConfig.getConfig()` is not null.

scholzj (commented):
This might be worth adding to CHANGELOG.md as well?

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaBrokerConfigurationBuilder.java
katheris: Good point about the null config, I've changed it and made use of getOrDefault, see what you think

katheris:
> You mean no other metrics created on the old topic but created on the new one?

Yes exactly

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaBrokerConfigurationBuilder.java
```diff
@@ -128,7 +131,9 @@ public KafkaBrokerConfigurationBuilder withCruiseControl(String clusterName, Cru
                                                              String replicationFactor, String minInSyncReplicas)   {
         if (cruiseControl != null) {
             printSectionHeader("Cruise Control configuration");
-            writer.println(CruiseControlConfigurationParameters.METRICS_TOPIC_NAME + "=strimzi.cruisecontrol.metrics");
+            Map<String, Object> config = Optional.ofNullable(cruiseControl.getConfig()).orElseGet(HashMap::new);
+            String metricsTopicName = config.getOrDefault(CruiseControlConfigurationParameters.CRUISE_CONTROL_METRIC_REPORTER_TOPIC_NAME.getValue(), CruiseControlConfigurationParameters.DEFAULT_METRIC_REPORTER_TOPIC_NAME).toString();
+            writer.println(CruiseControlConfigurationParameters.METRICS_TOPIC_NAME + "=" + metricsTopicName);
```
scholzj, line 136: I wonder if using `if` instead of creating a new map would be better from memory perspective. Or maybe we can at least use `Map.of()` or something to create small unmodifiable map?

scholzj (approved):
The code looks good. I just wonder a bit if the implementation is optimal from the memory perspective ... maybe something to think about.

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaBrokerConfigurationBuilder.java
katheris: I wondered about:
```
String metricsTopicName = Optional.ofNullable(cruiseControl.getConfig())
                    .map(config -> config.get(CruiseControlConfigurationParameters.CRUISE_CONTROL_METRIC_REPORTER_TOPIC_NAME.getValue()))
                    .map(Object::toString)
                    .orElse(CruiseControlConfigurationParameters.DEFAULT_METRIC_REPORTER_TOPIC_NAME);
```

but didn't know if that would also be less readable?

scholzj: It is. But maybe if it saves memory there is at least a reason for it. Do the two `.map()` after each other work? Doesn't that throw NPE?

katheris: I don't think it throws an NPE, my understanding is that the `.map()` is clever enough that if any return null it would drop down to the `.orElse()`

scholzj:
Thanks for the PR

PR #6647, opened 2022-04-07
[MO] - [system test] -> OauthPlainIsolatedST removing unnecessary kafka clients

see-quick:
…case

### Type of change

- Enhancement
- Refactoring

### Description

This PR changes the number of replicas that are used in the Kafka cluster during Oauth verification. Moreover, I have removed the unnecessary deployment of Kafka clients. 

This also removed the flakiness of `testProducerConsumerConnect` test case because when I run it locally before this change it fails with a 50% chance (but you had to run the whole test suite, not a single test case). With such a change I was unable to reproduce the error...

### Checklist


Frawless (approved):

scholzj:
Is this test run in Acceptance profile? It seems to run in Regression only ...

Issue #6597, opened 2022-03-30
Remove methods with signature that does not specifies namespace.


see-quick:
### Description 

Currently in our STs we have numerous methods with signature:

```java
(1) public static void method(final String namespaceName, final String clusterName...) {
	// code
}
```

```java
(2) public static void method(final String clusterName...) {
	// code
}
```

This could potentially insert an unintentional bug, which is at first glance hidden. In sequence mode, everything is working fine, but in class-wide parallelisation, the context of the Kubernetes client could be different from what we need. For instance, two parallel suites (i.e., @ParallelSuites) are executed in parallel, and one test case is correctly using the signature (1) method. And the second one is using the signature (2) method, which results in an error (if the namespaces of (1) and (2) differ). 

We should eliminate such signatures from our STs to reduce this error.

jankalinic:
Should be finally resolved by https://github.com/strimzi/strimzi-kafka-operator/pull/7555

PR #6656, opened 2022-04-08
[systemtest] OauthScopeIsolatedST - fix namespace handling in ClientUtils methods

im-konge:

### Type of change

- Bugfix

### Description

This PR fixes handling of namespaces inside the `ClientUtils` methods, as the `kubeClient().namespace(namespace)` is not ensuring, that the resources are taken from the `namespace` - so it needs to be specified in `inNamespace(namespace)` method.

Also, I'm removing some of the variables, which are not used inside the `OauthScopeIsolatedST`

### Checklist

### DIFF
--- a/test/src/main/java/io/strimzi/test/k8s/KubeClient.java
+++ b/test/src/main/java/io/strimzi/test/k8s/KubeClient.java
@@ -635,6 +635,10 @@ public class KubeClient {
         return checkSucceededJobStatus(getNamespace(), jobName, 1);
     }
 
+    public boolean checkSucceededJobStatus(String namespace, String jobName) {
+        return checkSucceededJobStatus(namespace, jobName, 1);
+    }
+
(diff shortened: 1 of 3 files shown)


see-quick (approved):
Thanks

test/src/main/java/io/strimzi/test/k8s/KubeClient.java
```diff
@@ -631,6 +631,10 @@ public boolean checkSucceededJobStatus(String jobName) {
         return checkSucceededJobStatus(getNamespace(), jobName, 1);
```
Frawless, line 631: Guess we should remove calls like this because it's not reliable from my pov.

see-quick: Yes, we do but I think it would be a big change (https://github.com/strimzi/strimzi-kafka-operator/issues/6597)

Frawless (approved):

scholzj (approved):

Issue #6569, opened 2022-03-25
Operator metrics are missing number of paused `KafkaConnector` resources


scholzj:
The Cluster Operator metrics seem to provide counters for number of custom resources with paused reconciliation for most CRs:
```
$ curl -s localhost:8080/metrics | grep strimzi_resources_paused
# HELP strimzi_resources_paused Number of custom resources the operator sees but does not reconcile due to paused reconciliations
# TYPE strimzi_resources_paused gauge
strimzi_resources_paused{kind="KafkaConnect",namespace="myproject",selector="",} 0.0
strimzi_resources_paused{kind="Kafka",namespace="myproject",selector="",} 0.0
strimzi_resources_paused{kind="KafkaMirrorMaker",namespace="myproject",selector="",} 0.0
strimzi_resources_paused{kind="KafkaBridge",namespace="myproject",selector="",} 0.0
strimzi_resources_paused{kind="KafkaMirrorMaker2",namespace="myproject",selector="",} 0.0
strimzi_resources_paused{kind="KafkaRebalance",namespace="myproject",selector="",} 0.0
```

However, it does not seem to provide it for the `KafkaConnector` resources:
```
$ curl -s localhost:8080/metrics | grep KafkaConnector
strimzi_reconciliations_failed_total{kind="KafkaConnector",namespace="myproject",selector="",} 15.0
strimzi_reconciliations_duration_seconds_bucket{kind="KafkaConnector",namespace="myproject",selector="",le="1.0",} 48.0
strimzi_reconciliations_duration_seconds_bucket{kind="KafkaConnector",namespace="myproject",selector="",le="5.0",} 48.0
strimzi_reconciliations_duration_seconds_bucket{kind="KafkaConnector",namespace="myproject",selector="",le="10.0",} 48.0
strimzi_reconciliations_duration_seconds_bucket{kind="KafkaConnector",namespace="myproject",selector="",le="30.0",} 48.0
strimzi_reconciliations_duration_seconds_bucket{kind="KafkaConnector",namespace="myproject",selector="",le="60.0",} 48.0
strimzi_reconciliations_duration_seconds_bucket{kind="KafkaConnector",namespace="myproject",selector="",le="120.0",} 48.0
strimzi_reconciliations_duration_seconds_bucket{kind="KafkaConnector",namespace="myproject",selector="",le="300.0",} 48.0
[...]

PR #6669, opened 2022-04-11
Add paused KafkaConnector resources metric (#6569)

ortex:
Signed-off-by: abushmin <diskbu@yandex.ru>

### Type of change

- Bugfix

### Description

Add metric for KafkaConnect resources with paused reconciliation. #6569
I've tested it manually and I'm not sure if I need to write a test for such changes.

### Checklist

- [x] Write tests
- [ ] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/AbstractConnectOperator.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/AbstractConnectOperator.java
@@ -124,6 +124,7 @@ public abstract class AbstractConnectOperator<C extends KubernetesClient, T exte
     private Map<String, Counter> connectorsFailedReconciliationsCounterMap = new ConcurrentHashMap<>(1);
     private Map<String, Counter> connectorsSuccessfulReconciliationsCounterMap = new ConcurrentHashMap<>(1);
     private Map<String, AtomicInteger> connectorsResourceCounterMap = new ConcurrentHashMap<>(1);
+    private Map<String, AtomicInteger> pausedConnectorsResourceCounterMap = new ConcurrentHashMap<>(1);
     private Map<String, Timer> connectorsReconciliationsTimerMap = new ConcurrentHashMap<>(1);
 
     public AbstractConnectOperator(Vertx vertx, PlatformFeaturesAvailability pfa, String kind,
@@ -364,6 +365,8 @@ public abstract class AbstractConnectOperator<C extends KubernetesClient, T exte
             connectStatus.setConnectorPlugins(connectorPlugins);
 
             connectorsResourceCounter(namespace).set(desiredConnectors.size());
+            long pausedConnectorsCount = desiredConnectors.stream().filter(Annotations::isReconciliationPausedWithAnnotation).count();
+            pausedConnectorsResourceCounter(namespace).set((int) pausedConnectorsCount);
 
             Set<String> deleteConnectorNames = new HashSet<>(runningConnectorNames);
             deleteConnectorNames.removeAll(desiredConnectors.stream().map(c -> c.getMetadata().getName()).collect(Collectors.toSet()));
@@ -884,6 +887,12 @@ public abstract class AbstractConnectOperator<C extends KubernetesClient, T exte
         return Operator.getGauge(namespace, KafkaConnector.RESOURCE_KIND, METRICS_PREFIX + "resources",
                 metrics, null, connectorsResourceCounterMap,
                 "Number of custom resources the operator sees");
+    }
+
+    public AtomicInteger pausedConnectorsResourceCounter(String namespace) {
+        return Operator.getGauge(namespace, KafkaConnector.RESOURCE_KIND, METRICS_PREFIX + "resources.paused",
+                metrics, null, pausedConnectorsResourceCounterMap,
+                "Number of connectors the connect operator sees but does not reconcile due to paused reconciliations");
     }
 
     public Timer connectorsReconciliationsTimer(String namespace) {
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/AbstractConnectOperator.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/AbstractConnectOperator.java
@@ -84,7 +84,6 @@ import java.util.Optional;
 import java.util.Set;
 import java.util.TreeMap;
 import java.util.concurrent.ConcurrentHashMap;
-import java.util.concurrent.atomic.AtomicInteger;
 import java.util.function.BiFunction;
 import java.util.function.Function;
 import java.util.stream.Collectors;
@@ -123,7 +122,6 @@ public abstract class AbstractConnectOperator<C extends KubernetesClient, T exte
     private Map<String, Counter> connectorsReconciliationsCounterMap = new ConcurrentHashMap<>(1);
     private Map<String, Counter> connectorsFailedReconciliationsCounterMap = new ConcurrentHashMap<>(1);
     private Map<String, Counter> connectorsSuccessfulReconciliationsCounterMap = new ConcurrentHashMap<>(1);
-    private Map<String, AtomicInteger> connectorsResourceCounterMap = new ConcurrentHashMap<>(1);
     private Map<String, Timer> connectorsReconciliationsTimerMap = new ConcurrentHashMap<>(1);
 
     public AbstractConnectOperator(Vertx vertx, PlatformFeaturesAvailability pfa, String kind,
@@ -192,12 +190,6 @@ public abstract class AbstractConnectOperator<C extends KubernetesClient, T exte
                     String connectorKind = kafkaConnector.getKind();
                     String connectName = kafkaConnector.getMetadata().getLabels() == null ? null : kafkaConnector.getMetadata().getLabels().get(Labels.STRIMZI_CLUSTER_LABEL);
                     String connectNamespace = connectorNamespace;
-
-                    if (action == Action.ADDED) {
-                        connectOperator.connectorsResourceCounter(connectorNamespace).incrementAndGet();
-                    } else if (action == Action.DELETED) {
-                        connectOperator.connectorsResourceCounter(connectorNamespace).decrementAndGet();
-                    }
 
                     switch (action) {
                         case ADDED:
@@ -882,12 +874,6 @@ public abstract class AbstractConnectOperator<C extends KubernetesClient, T exte
     public Counter connectorsSuccessfulReconciliationsCounter(String namespace) {
         return Operator.getCounter(namespace, KafkaConnector.RESOURCE_KIND, METRICS_PREFIX + "reconciliations.successful", metrics, null, connectorsSuccessfulReconciliationsCounterMap,
                 "Number of reconciliations done by the operator for individual resources which were successful");
-    }
-
-    public AtomicInteger connectorsResourceCounter(String namespace) {
-        return Operator.getGauge(namespace, KafkaConnector.RESOURCE_KIND, METRICS_PREFIX + "resources",
-                metrics, null, connectorsResourceCounterMap,
-                "Number of custom resources the operator sees");
     }
 
     public Timer connectorsReconciliationsTimer(String namespace) {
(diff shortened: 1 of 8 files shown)


scholzj (commented):
Thanks for the PR. Yes, this should ideally have some unit test.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/AbstractConnectOperator.java
```diff
@@ -364,6 +365,8 @@ protected Future<Void> reconcileConnectors(Reconciliation reconciliation, T conn
             connectStatus.setConnectorPlugins(connectorPlugins);
 
             connectorsResourceCounter(namespace).set(desiredConnectors.size());
+            long pausedConnectorsCount = desiredConnectors.stream().filter(Annotations::isReconciliationPausedWithAnnotation).count();
+            pausedConnectorsResourceCounter(namespace).set((int) pausedConnectorsCount);
```
scholzj, line 369: TBH, looking at the code, I wonder if the metrics work at all. It seems to count the number of connectors per-Connect instance. But there might be multiple instances of Connect per namespace which I guess will be overwriting each others counts? Is that something you tried while working on this by any chance?

ortex: Yeah, it doesn't work. The same problem with `strimzi_resources{kind="KafkaConnector"` metric.

ortex: Aggregation could be easily done if it's possible to extend `MetricsProvider` interface. But probably you have only AtomicInteger-based gauge for some reason.
That is how it could be looks like: https://github.com/ortex/strimzi-kafka-operator/commit/60ff7378a970abb2d64a7f0b1aa0ba8c3f799ffb

I didn't add this commit to PR because not sure if we go with this.

Btw, with per-connect approach connectors with the wrong `strimzi.io/cluster` label don't count. Perhaps could be fixed.

Anyway, I'm going to look more deeply at the watcher and how per-connector counting could be done.

scholzj: That is an interesting approach. I do not remember any specific reason why we have only the Atomic Integer. I would say it was all we needed. TBH, I didn't even know you can aggregate the metric like you do it in your branch.

I had a brief look at the connector watch. It would be easy to fix the total resource counter there. You can just increment / decrement the integer when the watch is called for created / deleted connectors. But it would not be completely easy to handle the paused connectors since it will make it hard to handle paused / unpaused connectors to increment / decrement the counter. So I guess your approach is better.

In theory, I guess one can also combine it? Count the total resources in the watch (that way it will include all resources even with wrong `strimzi.io/cluster` label) but the paused resources only in the Connect reconciliation? The connectors with wrong label are not really reconciled anyway, so the paused / unpaused state doesn't make much sense anyway.

ortex: Yeah, I'll try a combined approach

ortex:
Ready for review.
Per-connect counting for paused connectors doesn't work since for paused connect reconciliation stops and paused connectors are not counted anymore.
The current implementation relies on kafkaConnect status.

cluster-operator/src/test/java/io/strimzi/operator/cluster/operator/assembly/ConnectorMockTest.java
```diff
@@ -1550,4 +1613,385 @@ public void testConnectorReconciliationPausedUnpaused() {
...
+        Gauge resourcesPaused = meterRegistry.get("strimzi.resources.paused").tags(tags).gauge();
+        assertThat(resourcesPaused.value(), is(2.0));
+
+        Crds.kafkaConnectorOperation(client).inNamespace(NAMESPACE).delete(connector1);
+        Crds.kafkaConnectorOperation(client).inNamespace(NAMESPACE).delete(connector4);
+        waitForConnectorDeleted(connectorName1);
+        waitForConnectorDeleted(connectorName4);
+
+        assertThat(resources.value(), is(2.0));
+        waitFor("metrics", 50, 5000, () -> resourcesPaused.value() == 1);
```
ortex, line 1703: Couldn't find a better way to do it.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/AbstractConnectOperator.java
```diff
@@ -165,6 +170,27 @@ protected Future<Boolean> delete(Reconciliation reconciliation) {
         }).map(ignored -> Boolean.FALSE);
     }
 
+    @Override
+    public void reconcileAll(String trigger, String namespace, Handler<AsyncResult<Void>> handler) {
+        super.reconcileAll(trigger, namespace, handler);
+
+        connectorOperator.listAsync(namespace, selector())
```
scholzj, line 177: The selector is not applied on the Connector resources but on the connect resources IIRC. So this would need to select all, query the Connects and check their selectors. Which would be IMHO fairly expensive, but not sure there isa better solution.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/AbstractConnectOperator.java
```diff
@@ -243,6 +286,9 @@ public void eventReceived(Action action, KafkaConnector kafkaConnector) {
                                                 + Labels.STRIMZI_CLUSTER_LABEL
                                                 + "': No connect cluster in which to create this connector."),
                                         kafkaConnector, connectOperator.connectorOperator);
+                                if (isPaused(kafkaConnector.getStatus())) {
+                                    connectOperator.pausedConnectorsResourceCounter(connectorNamespace).decrementAndGet();
+                                }
```
scholzj, line 291: Same as with the two comments at the start of the event watch => and explanatory comment will make sure everyone better understands the logic in the future.

operator-common/src/main/java/io/strimzi/operator/common/MicrometerMetricsProvider.java
```diff
@@ -27,6 +27,14 @@ public MicrometerMetricsProvider() {
         this.metrics = BackendRegistries.getDefaultNow();
     }
 
+    /**
+     * Constructor of the Micrometer metrics provider
```
scholzj, line 31: Might be worth adding that this is here mainly for test purposes?

scholzj (commented):
I went through it and left some comments. In some cases I just think it would be wroth adding some explanatory comments tot he code to make it easier to remember and understand it in the future (I have to admit it took me some time to understand why/how it works - not because the code would be badly readable, but because the logic is complicated).

I wonder about some things:
* For most part, you seem to handle everything in the watch. Why is it that we need both the `reconcileAll` method and event watch to handle this? I think for some of the other resources, we do the counting only in the `reconcileAll` method and it simply takes more time for the metrics to update. In your case, the `reconcileAll` seems to be missing the `resetCounters` logic (see `AbstractOperator`) to reset the counters for deleted resources. That seems to be handled by the event watch. But the watch counting seems otherwise self-sufficient and should inclrement and decrement the counters as needed?
* I do not think it handles the CR selector well. The `reconcileAll` applies it on the Connector resources and not on the Connect resources. The watch does not seem to care about it => so it will not decrement when the Connect cluster moves to another operator. I guess this will need to be fixed in some way.

ortex:
Thanks for review!
I totally agree that logic is complicated. Probably too complicated for such a feature.

> Why is it that we need both the reconcileAll method and event watch to handle this?

Firstly I wrote all counting in the watch but then realized that after strimzi-cluster-operator pod restart the paused connectors couldn't be counted. They have both `ReconciliationPaused` condition and `strimzi.io/pause-reconciliation` annotation.

> I think for some of the other resources, we do the counting only in the reconcileAll method and it simply takes more time for the metrics to update. 

Originally `strimzi_resources` metric was counted in `reconcileConnectors` method and it's impossible to count `strimzi_resources_paused` correctly there, but could be counted in `reconcileAll` of course. I've missed that originally.

So, maybe we should go with a combined approach, as discussed earlier?
* count `strimzi_resources_paused` only in `reconcileAll`
* there wouldn't be any race condition
* logic would be much simpler
* most of your comments will be fixed with just code deletion :)
* question is: when counting `strimzi_resources_paused` should we rely on `ReconciliationPaused` condition or `strimzi.io/pause-reconciliation` annotation?


Or I could finish work on the current approach, as far as understand there is not much left to do.
* add comments
* there will be possible race condition for `strimzi_resources_paused`. Perhaps could be fixed. But as far as I can tell changes are small and discrepancies would be fixed within the next reconciliation.

I guess I prefer counting `strimzi_resources_paused` only in `reconcileAll` because logic will be simpler and the solution will be more stable.

> I do not think it handles the CR selector well. The reconcileAll applies it on the Connector resources and not on the Connect resources. The watch does not seem to care about it => so it will not decrement when the Connect cluster moves to another operator. I guess this will need to be fixed in some way.

I'll look more deeply into it.

Fixed a couple comments.

Btw, I'll squash commits before merging.

operator-common/src/main/java/io/strimzi/operator/common/MicrometerMetricsProvider.java
ortex: added

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/AbstractConnectOperator.java
```diff
@@ -165,6 +170,27 @@ protected Future<Boolean> delete(Reconciliation reconciliation) {
...
+                        resourceCounters.compute(connectorNamespace, (ns, counter) -> Objects.requireNonNullElse(counter, 0) + 1);
+                        if (isPaused(c.getStatus())) {
+                            pausedResourceCounters.compute(connectorNamespace, (ns, counter) -> Objects.requireNonNullElse(counter, 0) + 1);
+                        }
+                    });
+
+                    resourceCounters.forEach((ns, count) -> connectorsResourceCounter(ns).set(count));
+                    pausedResourceCounters.forEach((ns, count) -> pausedConnectorsResourceCounter(ns).set(count));
+                });
+    }
```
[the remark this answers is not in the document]
ortex: Moved to `Kafka Connect Assembly Operator`

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/AbstractConnectOperator.java
```diff
@@ -193,6 +219,12 @@ public void eventReceived(Action action, KafkaConnector kafkaConnector) {
                     String connectName = kafkaConnector.getMetadata().getLabels() == null ? null : kafkaConnector.getMetadata().getLabels().get(Labels.STRIMZI_CLUSTER_LABEL);
                     String connectNamespace = connectorNamespace;
 
+                    if (action == Action.ADDED) {
+                        connectOperator.connectorsResourceCounter(connectorNamespace).incrementAndGet();
+                    } else if (action == Action.DELETED) {
+                        connectOperator.connectorsResourceCounter(connectorNamespace).decrementAndGet();
+                    }
```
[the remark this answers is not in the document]
ortex: Removed counting `strimzi_resources` from `reconcileAll`. Now counting only here.
But race condition could be possible for `strimzi_resources_paused`.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/AbstractConnectOperator.java
ortex: I've used selector match expression and select connectors by all connect names.

ortex:
> I do not think it handles the CR selector well. The reconcileAll applies it on the Connector resources and not on the Connect resources. The watch does not seem to care about it => so it will not decrement when the Connect cluster moves to another operator. I guess this will need to be fixed in some way.

I moved counting `strimzi_resources` to `reconcileThese` as well. Even if it's possible to detect moving Connect cluster to another operator in watch to increment/decrement metric there is no simple way to do it and I would end up with too complex logic 😄 

> PS: Sorry, I think this seemed much simpler originally, but it doesn't really look like a good-start issue anymore to me 😔

That's ok 😄 

Ready for review

scholzj (approved):
Thanks for the PR.

ppatierno (approved):

scholzj:
Thanks for the PR!

PR #6670, opened 2022-04-12
Refactor Kafka reconciliation out of the `KafkaAssemblyOperator` class

scholzj:
### Type of change

- Refactoring

### Description

This PR refactors the Kafka reconciliation out from the `KafkaAssemblyOperator` (KAO) class. Doing that, it introduces several new classes:
* `KafkaListenerReconciler` handles the reconciliation of the Kafka broker listeners and the Kubernetes resources they need
* `KafkaReconciler` reconciles the Kafka cluster (everything left apart from the listeners)
* `VersionChangeCreator` prepares the `KafkaVersionChange` object based on analyzing the StatefulSet, PodSet and the broker Pods and prepares the plan for the upgrade
* With most of the KAO gone, it didn't needed anymore all of the configs and operators. So I moved it to store directly the `ClusterOperatorConfig` and `ResourceSupplier`. That allows for much smaller constructors for all the reconciler classes.
* The `currentReplicas` fields were changed from `Integer` to `int`. The original code kept them as `Integer`, but always set them to `0` if the cluster was new. That was confusing in the code with some parts requiring `int`, others using `Integer` and checking it for `null` etc.

### Checklist

- [x] Write tests


cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaAssemblyOperator.java
```diff
@@ -782,2198 +611,267 @@ Future<Void> rollDeploymentIfExists(String deploymentName, String reasons)  {
...
-                }
-            });
+        private int currentReplicas(StatefulSet sts)  {
+            int replicas = 0;
+
+            if (sts != null && sts.getSpec() != null)   {
+                replicas = sts.getSpec().getReplicas();
+            }
+
+            return replicas;
```
ppatierno, line 674: blank lines really needed?

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaAssemblyOperator.java
```diff
@@ -782,2198 +611,267 @@ Future<Void> rollDeploymentIfExists(String deploymentName, String reasons)  {
...
-        Future<Void> maybeRollKafka(int replicas, Function<Pod, List<String>> podNeedsRestart) {
-            return maybeRollKafka(replicas, podNeedsRestart, true);
+        private int currentReplicas(StrimziPodSet podSet)  {
+            int replicas = 0;
+
+            if (podSet != null && podSet.getSpec() != null && podSet.getSpec().getPods() != null)   {
+                replicas = podSet.getSpec().getPods().size();
+            }
+
+            return replicas;
```
ppatierno, line 691: As before but if you prefer having them no problem for me :-)

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaListenersReconciler.java
```diff
@@ -0,0 +1,1040 @@
...
+import java.util.Set;
+import java.util.stream.Collectors;
+
+import static java.util.Collections.singletonList;
+
+/**
+ * Class used for reconciliation of Kafka listeners. This class contains both the steps of the Kafka
+ * reconciliation pipeline related to the listeners and is also used to store the state between them.
+ */
+public class KafkaListenersReconciler {
```
ppatierno, line 56: Trying to figure out why a separate reconciler (from the Kafka one) is needed for the listeners. Just a design choice to make the `KafkaReconciler` less filled of actions or mandatory decision because of any issues?

ppatierno (approved):
LGTM. It's a great and useful refactoring. I left just a couple of nits and a question.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaListenersReconciler.java
scholzj: It could have been in one class, yes. But my thoughts were following:
* The combined class would have had over 2000 lines of code, so fairly long
* The listeners reconciliation is fairly long but relatively self-contained, so it can be easily extracted. Yet, its functions are related in terms of functionality and code and rely on each other (e.g. the Routes or Ingresses are created on top of the Services)

So it seemed like a good candidate to split a bit more. 

I also didn't really do a complete refactoring of the unit tests. I really just made them work as they were but testing the new code. But once we get to that as well, there are a many things to test just with the listeners and their different types. So this should also make the test easier in the future because you will test only the smaller class for the listeners and not the main KafkaRecocniler.

PR #6680, opened 2022-04-13
[systemtest] Rewrite throttling quotas tests

im-konge:

### Type of change

- Refactoring

### Description

The tests inside `ThrottlingQuotaST` were pretty flaky, which caused longer execution time inside AZP, or complete failure of the pipeline.

Issues which I found:
- 3 tests checking 3 different topic operations, 4th was mix of 3 tests before
- we were not waited long enough for throttling quota exception
- 1 Kafka and 1 ZK is not enough for these operations
- duplicates of code
- stuff which should have method were separated there
- creating too many topics - it's not scalability test and the quota can be hit differently
- passing Admin operation strangely - we have Enum for operations and we are passing it via String to the builder - why?
- when tests were running in parallel, there were like 1500 topics created
- steps which can be done one time (like creating same user) appeared in each tests

The quotas can be hit differently than creating big amount of topics - how is it written in [KIP-599](https://cwiki.apache.org/confluence/display/KAFKA/KIP-599%3A+Throttle+Create+Topic%2C+Create+Partition+and+Delete+Topic+Operations) it matters how many mutations will be done during the operation (f.e. we want to create 20 topics with 20 partitions and 1 replication factor -> 20 * 20 * 1 = 400 mutations) and the time spent on the operations.

This PR rewrites the whole suite:
- one test case for all operations
- 3 Kafka brokers and 3 ZKs
- proper waits for messages
- less topics (25 topics)

### Checklist

- [x] Write tests

### DIFF
--- a/systemtest/src/main/java/io/strimzi/systemtest/Constants.java
+++ b/systemtest/src/main/java/io/strimzi/systemtest/Constants.java
@@ -60,6 +60,8 @@ public interface Constants {
     // it is replacement instead of checking logs for reconciliation using dynamic waiting on some change for some period of time
     int GLOBAL_RECONCILIATION_COUNT = (int) ((RECONCILIATION_INTERVAL / GLOBAL_POLL_INTERVAL) + GLOBAL_STABILITY_OFFSET_COUNT);
 
+    long THROTTLING_EXCEPTION_TIMEOUT = Duration.ofMinutes(10).toMillis();
+
     /**
      * Scraper pod labels
      */
@@ -380,6 +382,7 @@ public interface Constants {
     String SCRAPER_KEY = "SCRAPER_NAME";
     String PRODUCER_KEY = "PRODUCER_NAME";
     String CONSUMER_KEY = "CONSUMER_NAME";
+    String ADMIN_KEY = "ADMIN_NAME";
     String USER_NAME_KEY = "USER_NAME";
     String KAFKA_CLIENTS_POD_KEY = "KAFKA_CLIENTS_POD_NAME";
     String KAFKA_TRACING_CLIENT_KEY = "KAFKA_TRACING_CLIENT";
--- a/systemtest/src/main/java/io/strimzi/systemtest/cli/KafkaCmdClient.java
+++ b/systemtest/src/main/java/io/strimzi/systemtest/cli/KafkaCmdClient.java
@@ -6,6 +6,7 @@ package io.strimzi.systemtest.cli;
 
 import io.strimzi.api.kafka.model.KafkaResources;
 import io.strimzi.systemtest.utils.kafkaUtils.KafkaTopicUtils;
+import org.apache.logging.log4j.Level;
 
 import java.util.Arrays;
 import java.util.List;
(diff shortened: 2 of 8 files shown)


systemtest/src/main/java/io/strimzi/systemtest/kafkaclients/internalClients/KafkaAdminClients.java
```diff
@@ -73,15 +73,15 @@ public void setReplicationFactor(int replicationFactor) {
         this.replicationFactor = replicationFactor <= 0 ? 1 : replicationFactor;
     }
 
-    public String getTopicOperation() {
+    public AdminClientOperations getTopicOperation() {
```
scholzj, line 76: Is it _topic operation_ if it returns `AdminClientOperations` which seems more generic? Maybe better name would make it more reusable.

systemtest/src/main/java/io/strimzi/systemtest/Constants.java
```diff
@@ -60,6 +60,8 @@ public interface Constants {
     // it is replacement instead of checking logs for reconciliation using dynamic waiting on some change for some period of time
     int GLOBAL_RECONCILIATION_COUNT = (int) ((RECONCILIATION_INTERVAL / GLOBAL_POLL_INTERVAL) + GLOBAL_STABILITY_OFFSET_COUNT);
 
+    long THROTTLING_EXCEPTION_TIMEOUT = Duration.ofMinutes(10).toMillis();
```
scholzj, line 63: Seems rather long. What is the reason for it?

im-konge: Actual implementation of the admin clients block the thread until some timeout, which can be, f.e. 5 minutes (we need to improve this in the next PR to the test-clients).

Currently, there is 5 minutes timeout for the throttling quota exception, which can be easily hit and the test then fails, even though the job fails right after 2 seconds. I added there 10 minutes just to be 100% sure.

After we fix the admin client behavior, I wanted to decrease this timeout.

Maybe some comment here would be good.

systemtest/src/main/java/io/strimzi/systemtest/kafkaclients/internalClients/KafkaAdminClients.java
im-konge: Yes, make sense. Good catch.

scholzj: Even five minutes seem to be like awfully lot of waiting.

systemtest/src/main/java/io/strimzi/systemtest/Constants.java
im-konge: @scholzj 
> Even five minutes seem to be like awfully lot of waiting.

Yeah, it's awful for me as well, but without rewriting the whole admin client, I cannot do anything with it..

scholzj: Ok :-(

see-quick (approved):

systemtest/src/main/java/io/strimzi/systemtest/cli/KafkaCmdClient.java
```diff
@@ -68,4 +69,19 @@ public static String updateTopicPartitionsCountUsingPodCli(String namespaceName,
...
+        );
+
+        return cmdKubeClient().namespace(namespaceName).execInPod(Level.DEBUG, kafkaPodName, "/opt/kafka/bin/kafka-topics.sh",
+                "--list",
+                "--bootstrap-server",
+                bootstrapName,
+                "--command-config",
+                "/tmp/config.properties")
+            .out();
+    }
```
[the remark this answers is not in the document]
im-konge: Which one do you mean? We were removing the usage of ZK pods for these checks, other pods don't have the script. If we would use the test-client pod (which implementation is being removed), we will have to move to something else in the few weeks.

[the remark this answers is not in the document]
scholzj: Well, most likely you should start a new pod for it anc connect to Kafka.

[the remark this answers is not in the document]
scholzj: Well, I'm saying the same for several years ... do not run stuff inside the Zoo/Kafka pods. If you wanna contrinue that way fine, but I think it is not helping.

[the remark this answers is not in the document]
im-konge: We don't want to continue with that, I will do it in another PR.
I just created #6689 which tracks this issue.

Do you agree?

scholzj (approved):

PR #6684, opened 2022-04-14
[MO] - [system test] -> ListenersST not specified namespace in KafkaC…

see-quick:
…lients resulting in NPE

### Type of change

- Bugfix

### Description

During the replacement of Kafka clients, we forget to specify the namespace in some test cases which result in NPE. This PR fixes the problem.

### Checklist


scholzj (approved):

im-konge (approved):
Thanks :)

PR #6685, opened 2022-04-14
Removed duplicated Cruise Control constants

ppatierno:
Trivial PR about removing some duplicated Cruise Control constants and renaming others following the pattern with the `CRUISE_CONTROL_` suffix.

### DIFF
--- a/operator-common/src/main/java/io/strimzi/operator/cluster/operator/resource/cruisecontrol/CruiseControlConfigurationParameters.java
+++ b/operator-common/src/main/java/io/strimzi/operator/cluster/operator/resource/cruisecontrol/CruiseControlConfigurationParameters.java
@@ -5,15 +5,10 @@
 package io.strimzi.operator.cluster.operator.resource.cruisecontrol;
 
 public enum CruiseControlConfigurationParameters {
-    CONCURRENT_PARTITION_MOVEMENTS("num.concurrent.partition.movements.per.broker"),
-    CONCURRENT_INTRA_PARTITION_MOVEMENTS("num.concurrent.intra.broker.partition.movements"),
-    CONCURRENT_LEADER_MOVEMENTS("num.concurrent.leader.movements"),
-    REPLICATION_THROTTLE("default.replication.throttle"),
-    BROKER_METRICS_WINDOWS("num.broker.metrics.windows"),
-    BROKER_METRICS_WINDOW_MS("broker.metrics.window.ms"),
-    PARTITION_METRICS_WINDOWS("num.partition.metrics.windows"),
-    PARTITION_METRICS_WINDOW_MS("partition.metrics.window.ms"),
-    COMPLETED_USER_TASK_RETENTION_MS("completed.user.task.retention.time.ms"),
+    CRUISE_CONTROL_CONCURRENT_PARTITION_MOVEMENTS("num.concurrent.partition.movements.per.broker"),
+    CRUISE_CONTROL_CONCURRENT_INTRA_PARTITION_MOVEMENTS("num.concurrent.intra.broker.partition.movements"),
+    CRUISE_CONTROL_CONCURRENT_LEADER_MOVEMENTS("num.concurrent.leader.movements"),
+    CRUISE_CONTROL_REPLICATION_THROTTLE("default.replication.throttle"),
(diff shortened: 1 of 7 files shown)


kyguy (approved):

operator-common/src/main/java/io/strimzi/operator/cluster/operator/resource/cruisecontrol/CruiseControlConfigurationParameters.java
```diff
@@ -5,15 +5,10 @@
...
-    REPLICATION_THROTTLE("default.replication.throttle"),
-    BROKER_METRICS_WINDOWS("num.broker.metrics.windows"),
-    BROKER_METRICS_WINDOW_MS("broker.metrics.window.ms"),
-    PARTITION_METRICS_WINDOWS("num.partition.metrics.windows"),
-    PARTITION_METRICS_WINDOW_MS("partition.metrics.window.ms"),
-    COMPLETED_USER_TASK_RETENTION_MS("completed.user.task.retention.time.ms"),
+    CRUISE_CONTROL_CONCURRENT_PARTITION_MOVEMENTS("num.concurrent.partition.movements.per.broker"),
+    CRUISE_CONTROL_CONCURRENT_INTRA_PARTITION_MOVEMENTS("num.concurrent.intra.broker.partition.movements"),
+    CRUISE_CONTROL_CONCURRENT_LEADER_MOVEMENTS("num.concurrent.leader.movements"),
+    CRUISE_CONTROL_REPLICATION_THROTTLE("default.replication.throttle"),
```
scholzj, line 11: Do we really want the prefix? The class is `CruiseControlConfigurationParameters` => so `CruiseControlConfigurationParameters.CONCURRENT_PARTITION_MOVEMENTS` is shorter and is still self-explaining omapred to `CruiseControlConfigurationParameters.CRUISE_CONTROL_CONCURRENT_PARTITION_MOVEMENTS`.

kyguy:
Since these fields are constants and will always be referenced with `CruiseControlConfigurationParameters` prefix, would it be less verbose to remove the `CRUISE_CONTROL_` prefix from the constant names instead?

operator-common/src/main/java/io/strimzi/operator/cluster/operator/resource/cruisecontrol/CruiseControlConfigurationParameters.java
ppatierno: @scholzj @kyguy I totally agree and I wanted to do that but I think that the initial design was to distinguish with other constants starting with `METRICS_REPORTER_` and `METRICS_TOPIC_` within the same class. So the idea (@kyguy can you confirm?) was about differentiate between keys related to more cruise control generic config parameters vs specific ones for the metrics reporter and topic. I am just guessing but I think that it was the rationale behind. But if you want I can remove the `CRUISE_CONTROL_` part.

scholzj: TBH, if you wanna keep the long names, I'm fine and I can approve it. Not a big deal. That said, as someone not that much familiar with the CC implementation, I don't think I understand the logic and I don't think it is intuitive. How are `METRICS_TOPIC_` or `METRICS_REPORTER_` options different from the other options? It all seems Cruise Control to me. But as I said, not much difference for me as I anyway don't work much with the CC code. So if it is helpful for SMEs, let's keep it.

scholzj (approved):

operator-common/src/main/java/io/strimzi/operator/cluster/operator/resource/cruisecontrol/CruiseControlConfigurationParameters.java
kyguy: @ppatierno I don't remember but it sounds reasonable to me! Let's keep it as is in that case!

ppatierno:
I think I am going to remove the `CRUISE_CONTROL_` prefix. The others like `METRICS_REPORTER_` and `METRICS_TOPIC_` should stay imho. It's clear they are CC related (because of the class they leave) but the prefix makes more clear to what they are referring to.

tomncooper:
I think it make sense to go with the shorter names.

PR #6699, opened 2022-04-18
Fail recocniliation when invalid storage unit is used

scholzj:
### Type of change

- Bugfix

### Description

When handling storage, `InvalidArgumentException` can be thrown when invalid unit is used. When the `PvcReconciler` was   considering storage resizing, if was doing so in the `onComplete` handler. And when the exeption was thrown, it was logged, but it did not fail the reconciliation. This PR improves the flow of the `PvcRecocniler` to not do things in the `onComplete` handlers and use `compose` pattern instead. That makes sure the exception is caught and fails the reconciliation properly. It also adds a test to cover this.

Note: The way how to reproduce the issue is to for example use storage with capacity of `0.2Ti`. That is converted to `millibytes` by Kubernetes which throws the `InvalidArgumentException` in the next reconciliation. This PR solves only the stuck reconciliation, but not handling of millibyte units which might be more complicated. #6700 was opened for this.

### Checklist

- [x] Write tests

### DIFF
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/PvcReconciler.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/PvcReconciler.java
@@ -5,7 +5,6 @@
 package io.strimzi.operator.cluster.operator.assembly;
 
 import io.fabric8.kubernetes.api.model.PersistentVolumeClaim;
-import io.fabric8.kubernetes.api.model.storage.StorageClass;
 import io.strimzi.operator.cluster.model.AbstractModel;
 import io.strimzi.operator.cluster.model.StorageUtils;
 import io.strimzi.operator.common.Annotations;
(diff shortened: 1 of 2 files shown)


cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/PvcReconciler.java
```diff
@@ -67,48 +65,43 @@ public Future<Collection<String>> resizeAndReconcilePvcs(Function<Integer, Strin
...
-                        Long desiredSize = StorageUtils.parseMemory(desiredPvc.getSpec().getResources().getRequests().get("storage"));
-
-                        if (!currentSize.equals(desiredSize))   {
-                            // The sizes are different => we should resize (shrinking will be handled in StorageDiff, so we do not need to check that)
-                            resizePvc(currentPvc, desiredPvc).onComplete(resultPromise);
-                        } else  {
-                            // size didn't change, just reconcile
-                            pvcOperator.reconcile(reconciliation, reconciliation.namespace(), desiredPvc.getMetadata().getName(), desiredPvc)
-                                    .onComplete(r -> resultPromise.complete());
+            //Promise<Void> resultPromise = Promise.promise();
```
ppatierno, line 68: is this a leftover?

ppatierno (approved):
LGTM. Just a nit.

PR #6705, opened 2022-04-19
Avoid unnecessary rolling updates of the Cruise Control deployment

scholzj:
### Type of change

- Bugfix

### Description

The `CruiseControlConfiguration` class contains a static Map with the default configuration options. This Map is created using `Map.ofEntries` method. The map is used to add the default values to the user provided CC configuration. But the ordering of the Map created with `Map.ofEntries` is non-deterministic and changes with every restart of the CO. And that causes the Cruise Control configuration in the `CRUISE_CONTROL_CONFIGURATION` environment variable to be reordered and triggers roll of the CC deployment.

This PR changes the map with the default values to be an unmodifiable sorted map. Thanks to that, the ordering is deterministic with every restart of the Cluster Operator and the unnecessary rolling updates are not triggered.

### Checklist


ppatierno (approved):

tomncooper (approved):
Nice catch. Will have to watch out for this in future.

PR #6707, opened 2022-04-20
[ST] Test CruiseControl Resources/JVM options

michalxo:
Signed-off-by: Michal T <mtoth@redhat.com>

Added Test to check CruiseControl resources and JVM arguments.

### DIFF
--- a/systemtest/src/test/java/io/strimzi/systemtest/cruisecontrol/CruiseControlST.java
+++ b/systemtest/src/test/java/io/strimzi/systemtest/cruisecontrol/CruiseControlST.java
@@ -4,6 +4,8 @@
  */
 package io.strimzi.systemtest.cruisecontrol;
 
+import io.fabric8.kubernetes.api.model.Quantity;
+import io.fabric8.kubernetes.api.model.ResourceRequirementsBuilder;
 import io.strimzi.api.kafka.model.CruiseControlResources;
 import io.strimzi.api.kafka.model.KafkaRebalance;
 import io.strimzi.api.kafka.model.KafkaTopicSpec;
@@ -68,7 +70,7 @@ public class CruiseControlST extends AbstractST {
     private final String namespace = testSuiteNamespaceManager.getMapOfAdditionalNamespaces().get(CruiseControlST.class.getSimpleName()).stream().findFirst().get();
 
     @IsolatedTest
-    void testAutoCreationOfCruiseControlTopics(ExtensionContext extensionContext) {
+    void testAutoCreationOfCruiseControlTopicsWithResources(ExtensionContext extensionContext) {
         final String clusterName = mapWithClusterNames.get(extensionContext.getDisplayName());
 
         resourceManager.createResource(extensionContext, KafkaTemplates.kafkaWithCruiseControl(clusterName, 3, 3)
@@ -79,8 +81,27 @@ public class CruiseControlST extends AbstractST {
                 .editKafka()
                     .addToConfig("auto.create.topics.enable", "false")
                 .endKafka()
+                .editCruiseControl()
+                    .withResources(new ResourceRequirementsBuilder()
+                        .addToLimits("memory", new Quantity("256Mi"))
+                        .addToLimits("cpu", new Quantity("1"))
+                        .addToRequests("memory", new Quantity("256Mi"))
+                        .addToRequests("cpu", new Quantity("0.8"))
+                        .build())
+                    .withNewJvmOptions()
+                        .withXmx("512M")
+                        .withXms("256M")
+                        .withXx(Map.of("UseG1GC", "true"))
+                    .endJvmOptions()
+                .endCruiseControl()
             .endSpec()
             .build());
+
+        String ccPodName = kubeClient().listPodsByPrefixInName(namespace, CruiseControlResources.deploymentName(clusterName)).get(0).getMetadata().getName();
+        assertResources(namespace, ccPodName, "cruise-control",
+                "256Mi", "1", "256Mi", "800m");
+        assertExpectedJavaOpts(namespace, ccPodName, "cruise-control",
+                "-Xmx512M", "-Xms256M", "-XX:+UseG1GC");
 
         KafkaTopicUtils.waitForKafkaTopicReady(namespace, CRUISE_CONTROL_METRICS_TOPIC);
         KafkaTopicSpec metricsTopic = KafkaTopicResource.kafkaTopicClient().inNamespace(namespace)
--- a/systemtest/src/test/java/io/strimzi/systemtest/cruisecontrol/CruiseControlST.java
+++ b/systemtest/src/test/java/io/strimzi/systemtest/cruisecontrol/CruiseControlST.java
@@ -4,6 +4,8 @@
  */
 package io.strimzi.systemtest.cruisecontrol;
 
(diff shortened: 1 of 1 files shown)


see-quick:
Why such a test case is needed? It seems to me that this can be verified on the `UNIT` level. Is there some hidden motivation for this, which I don't see? :)

scholzj (commented):
Is this really worth separate test instead of adding it to some existing test? Also, I wonder what are the resources based on. They seem to be fairly high given our resource issues in STs.

systemtest/src/test/java/io/strimzi/systemtest/cruisecontrol/CruiseControlST.java
```diff
@@ -67,6 +70,38 @@ public class CruiseControlST extends AbstractST {
...
+            .endSpec()
+            .build());
+
+        String ccPodName = kubeClient().listPodsByPrefixInName(namespace, CruiseControlResources.deploymentName(clusterName)).get(0).getMetadata().getName();
+        assertResources(namespace, ccPodName, "cruise-control",
+                "1.5Gi", "1", "1Gi", "800m");
+        assertExpectedJavaOpts(namespace, ccPodName, "cruise-control",
+                "-Xmx1G", "-Xms256M", "-XX:+UseG1GC");
+    }
+
```
im-konge, line 104: The test itself is good, but I think you can add it to another test, which is already written. My point is that you will wait for example 5 minutes for Kafka and CC will be up and then you just check that resources are correct - which is little bit small check for the extra time.

scholzj:
> Why such a test case is needed? It seems to me that this can be verified on the UNIT level. Is there some hidden motivation for this, which I don't see? :)

The asserting of the Java options is not something what is really tested in unit tests since that is about actually running the operand. How critical it is to cover that with ST, that is another discussion I guess.

systemtest/src/test/java/io/strimzi/systemtest/cruisecontrol/CruiseControlST.java
michalxo: Yup, already added it into `testAutoCreationOfCruiseControlTopicsWithResources` (and changed name). 
I think it won't affect original test in any way. Agree?

systemtest/src/test/java/io/strimzi/systemtest/cruisecontrol/CruiseControlST.java
```diff
@@ -79,9 +81,28 @@ void testAutoCreationOfCruiseControlTopics(ExtensionContext extensionContext) {
...
                 .endKafka()
+                .editCruiseControl()
+                    .withResources(new ResourceRequirementsBuilder()
+                        .addToLimits("memory", new Quantity("256Mi"))
+                        .addToLimits("cpu", new Quantity("1"))
+                        .addToRequests("memory", new Quantity("256Mi"))
+                        .addToRequests("cpu", new Quantity("0.8"))
+                        .build())
+                    .withNewJvmOptions()
+                        .withXmx("512M")
```
scholzj, line 92: you cannot fit Xmx512M into 256Mi. You probably cannot fit -Xms256M either.

systemtest/src/test/java/io/strimzi/systemtest/cruisecontrol/CruiseControlST.java
```diff
@@ -79,9 +81,28 @@ void testAutoCreationOfCruiseControlTopics(ExtensionContext extensionContext) {
                 .editKafka()
                     .addToConfig("auto.create.topics.enable", "false")
                 .endKafka()
+                .editCruiseControl()
+                    .withResources(new ResourceRequirementsBuilder()
+                        .addToLimits("memory", new Quantity("256Mi"))
+                        .addToLimits("cpu", new Quantity("1"))
+                        .addToRequests("memory", new Quantity("256Mi"))
+                        .addToRequests("cpu", new Quantity("0.8"))
+                        .build())
```
scholzj, line 90: Maybe leaving out the CPU completely would be better -> lower chance of conflict or of running out of resources.

systemtest/src/test/java/io/strimzi/systemtest/cruisecontrol/CruiseControlST.java
michalxo: Aaah, I see now. Of course it's connected to container limits.

systemtest/src/test/java/io/strimzi/systemtest/cruisecontrol/CruiseControlST.java
michalxo: I will leave there 0.1 CPU for both, as we don't have such test for CC. Would you approve it with that? Else I would need to duplicate code from `assertResources()` and also possibly write new test as well for that CPU checks :-)

michalxo: Update: Will remove CPU, as CC is not able to start with 0.1 CPU.

systemtest/src/test/java/io/strimzi/systemtest/cruisecontrol/CruiseControlST.java
```diff
@@ -79,9 +84,27 @@ void testAutoCreationOfCruiseControlTopics(ExtensionContext extensionContext) {
...
+                        .withXx(Map.of("UseG1GC", "true"))
+                    .endJvmOptions()
+                .endCruiseControl()
             .endSpec()
             .build());
 
+        String ccPodName = kubeClient().listPodsByPrefixInName(namespace, CruiseControlResources.deploymentName(clusterName)).get(0).getMetadata().getName();
+        Container container = (Container) KubeClusterResource.kubeClient(namespace).getPod(namespace, ccPodName).getSpec().getContainers().stream().filter(c -> c.getName().equals("cruise-control")).findFirst().get();
+        assertThat(container.getResources().getLimits().get("memory"), CoreMatchers.is(new Quantity("300Mi")));
+        assertThat(container.getResources().getRequests().get("memory"), CoreMatchers.is(new Quantity("300Mi")));
```
im-konge, line 104: `CoreMatchers.is()` -> `is()`

im-konge (approved):
LGTM, one nit :)

systemtest/src/test/java/io/strimzi/systemtest/cruisecontrol/CruiseControlST.java
```diff
@@ -79,9 +83,27 @@ void testAutoCreationOfCruiseControlTopics(ExtensionContext extensionContext) {
...
                 .endKafka()
+                .editCruiseControl()
+                    .withResources(new ResourceRequirementsBuilder()
+                        .addToLimits("memory", new Quantity("300Mi"))
+                        .addToRequests("memory", new Quantity("300Mi"))
+                        .build())
+                    .withNewJvmOptions()
+                        .withXmx("200M")
+                        .withXms("128M")
+                        .withXx(Map.of("UseG1GC", "true"))
```
see-quick, line 94:
```suggestion
                        .withXx(Collections.singletonMap("UseG1GC", "true"))
```

see-quick (approved):
Looks good, just a nuance.

scholzj (approved):

PR #6660, opened 2022-04-09
Add client rack-awareness support to Mirror Maker 2

fvaleri:
### Type of change

_Select the type of your PR_

- Enhancement / new feature

### Description

This PR adds new property `client.rack` for Mirror Maker 2 cluster, which allows to use fetch from closest replica feature.

It implements https://github.com/strimzi/strimzi-kafka-operator/issues/6611.

### Checklist

- [x] Write tests
- [x] Make sure all tests pass
- [x] Update documentation
- [x] Check RBAC rights for Kubernetes / OpenShift roles
- [x] Try your changes from Pod inside your Kubernetes and OpenShift cluster, not just locally
- [x] Reference relevant issue(s) and close them after merging
- [x] Update CHANGELOG.md

PR #6710, opened 2022-04-20
[ST] Add Connect/MM2 rack awareness system tests

fvaleri:
Add Connect/MM2 rack awareness system tests related to PR [#6611](https://github.com/strimzi/strimzi-kafka-operator/pull/6660).


scholzj:
@fvaleri You seem to have some unused import there:
```
[ERROR] /home/vsts/work/1/s/systemtest/src/test/java/io/strimzi/systemtest/mirrormaker/MirrorMaker2IsolatedST.java:58:8: Unused import - java.net.UnknownHostException. [UnusedImports]
```

systemtest/src/test/java/io/strimzi/systemtest/connect/ConnectIsolatedST.java
```diff
@@ -1318,6 +1319,31 @@ void testKafkaConnectWithScramShaAuthenticationRolledAfterPasswordChanged(Extens
...
+        assertThat(podJson, hasJsonPath(StUtils.initGlobalVariableJsonPathBuilder(0, "RACK_TOPOLOGY_KEY"), hasItem(topologyKey)));
+        assertThat(podJson, hasJsonPath(StUtils.nodeSelectorKeyJsonPathBuilder(topologyKey), hasItem(topologyKey)));
+
+        String podLogs = kubeClient(storage.getNamespaceName()).logs(podName);
+        String podNodeName = PodUtils.getPodByName(storage.getNamespaceName(), podName).getSpec().getNodeName();
+        String hostname = podNodeName.contains(".") ? podNodeName.substring(0, podNodeName.indexOf(".")) : podNodeName;
+        String clientRackConfig = "consumer.client.rack=" + hostname;
+        assertTrue(podLogs.contains(clientRackConfig), "Pod logs contain " + clientRackConfig);
+    }
+
```
im-konge, line 1346: This is already tested in `testRackAwareConnectCorrectDeployment`, so you can remove it.

systemtest/src/test/java/io/strimzi/systemtest/mirrormaker/MirrorMaker2IsolatedST.java
```diff
@@ -1633,6 +1635,39 @@ void testKMM2RollAfterSecretsCertsUpdateTLS(ExtensionContext extensionContext) {
...
+        String podName = PodUtils.getPodNameByPrefix(namespaceName, KafkaMirrorMaker2Resources.deploymentName(clusterName));
+        String podJson = TestUtils.toJsonString(kubeClient(namespaceName).getPod(namespaceName, podName));
+        assertThat(podJson, hasJsonPath(StUtils.initGlobalVariableJsonPathBuilder(0, "RACK_TOPOLOGY_KEY"), hasItem(topologyKey)));
+        assertThat(podJson, hasJsonPath(StUtils.nodeSelectorKeyJsonPathBuilder(topologyKey), hasItem(topologyKey)));
+
+        String podLogs = kubeClient(namespaceName).logs(podName);
+        String podNodeName = PodUtils.getPodByName(namespaceName, podName).getSpec().getNodeName();
+        String hostname = podNodeName.contains(".") ? podNodeName.substring(0, podNodeName.indexOf(".")) : podNodeName;
+        String clientRackConfig = "consumer.client.rack=" + hostname;
+        assertTrue(podLogs.contains(clientRackConfig), "Pod logs contain " + clientRackConfig);
```
im-konge, line 1668: Why are you using the JSON output for assertion etc? It can be done using the fabric8 api as in other rack awareness tests.

systemtest/src/test/java/io/strimzi/systemtest/mirrormaker/MirrorMaker2IsolatedST.java
```diff
@@ -1633,6 +1635,39 @@ void testKMM2RollAfterSecretsCertsUpdateTLS(ExtensionContext extensionContext) {
         LOGGER.info("Messages successfully mirrored");
     }
 
+    @ParallelNamespaceTest
+    void testMirrorMaker2RackAwareness(ExtensionContext extensionContext) {
+        final String namespaceName = StUtils.getNamespaceBasedOnRbac(INFRA_NAMESPACE, extensionContext);
+        String clusterName = mapWithClusterNames.get(extensionContext.getDisplayName());
+        String kafkaClusterSourceName = clusterName + "-source";
+        String kafkaClusterTargetName = clusterName + "-target";
```
im-konge, line 1643: You can use here `TestStorage` as well -> for Kafka names then use just `testStorage.getClusterName() + "-source"/"-target"`

im-konge (commented):
Thanks for the PR, I left couple of comments

systemtest/src/test/java/io/strimzi/systemtest/connect/ConnectIsolatedST.java
fvaleri: That method is not checking `consumer.client.rack` configuration, which is the core of this feature. We can integrate that part in `testRackAwareConnectCorrectDeployment`. WDYT?

TBH the test class name doesn't make it easy to find out that it is dealing with rack awareness. I would suggest `RackAwarenessIsolatedST`.

systemtest/src/test/java/io/strimzi/systemtest/mirrormaker/MirrorMaker2IsolatedST.java
fvaleri: OK, I'll look at that. Thanks.

systemtest/src/test/java/io/strimzi/systemtest/mirrormaker/MirrorMaker2IsolatedST.java
fvaleri: OK. Thanks.

systemtest/src/test/java/io/strimzi/systemtest/connect/ConnectIsolatedST.java
im-konge: Yeah I agree that name is not making it easy to find it out. The `SpecificIsolatedST` has more then just rack awareness tests, so maybe moving the tests to separate class and renaming the test would make it clear?

fvaleri: Yeah I'll try to do that.

im-konge: If you'll need any help, feel free to ping me :)

fvaleri:
@im-konge I extracted the rack awareness test logic into dedicated per-component classes, grouped by a new `RACK_AWARENESS` tag. The `SpecificIsolatedST.testRackAware` was not using a valid `topologyKey`, so I also fixed that before moving it out. I ran all tests both on my local Minikube VM and on a remote OpenShift 4.7 cluster.

```shell
mvn verify -pl systemtest -am -Pall \
  -Dit.test="KafkaRackST,ConnectRackIsolatedST,MirrorMaker2RackST" \
  -Dmaven.javadoc.skip=true -Dskip.surefire.tests=true \
  -DfailIfNoTests=false -Dfailsafe.rerunFailingTestsCount=2
```

All new test classes except `ConnectRackIsolatedST` have parallel execution mode, which I tested running `KafkaST` and `KafkaRackST` with package-wide parallelism enabled.

```shell
mvn verify -pl systemtest -am -Pall -Dit.test="KafkaRackST,KafkaST" \
  -Dmaven.javadoc.skip=true -Dskip.surefire.tests=true \
  -DfailIfNoTests=false -Dfailsafe.rerunFailingTestsCount=2 \
  -Djunit.jupiter.execution.parallel.enabled=true \
  -Djunit.jupiter.execution.parallel.config.fixed.parallelism=5
```

im-konge:
@fvaleri thanks for this change
TBH, I don't think that we need new suite for every resource, which can use the rack awareness.
I would create one ST - something like `RackAwarenessST` - that will contain all tests related to rack awareness.

This way we would save a little bit of execution time, because resources like Kafka can be used even for the Connect and MM2 test (at least that's my POV)

Second option is to have the test for each resource in the resource's suite - Kafka rack awareness test in `KafkaST` etc.

fvaleri:
@im-konge I agree, based on testing :)

Will commit that change shortly.

jankalinic (approved):
Looks good to me, let's see what @im-konge thinks about the suggested one test file change.

systemtest/src/test/java/io/strimzi/systemtest/specific/RackAwarenessIsolatedST.java
```diff
@@ -0,0 +1,421 @@
...
+
+    @IsolatedTest
+    void testMirrorMaker2RackAwareness(ExtensionContext extensionContext) {
+        assumeFalse(Environment.isNamespaceRbacScope());
+
+        TestStorage storage = new TestStorage(extensionContext);
+        String namespace = storage.getNamespaceName();
+        String clusterName = storage.getClusterName();
+        String kafkaClusterSourceName = clusterName + "-source";
+        String kafkaClusterTargetName = clusterName + "-target";
```
im-konge, line 377: Same as above

systemtest/src/test/java/io/strimzi/systemtest/specific/RackAwarenessIsolatedST.java
```diff
@@ -0,0 +1,421 @@
...
+        String topologyKey = "kubernetes.io/hostname";
```
im-konge, line 378: Can you create some constant or something for this? It's repeated in every test.

systemtest/src/test/java/io/strimzi/systemtest/specific/RackAwarenessIsolatedST.java
```diff
@@ -0,0 +1,421 @@
...
+
+        resourceManager.createResource(extensionContext,
+                KafkaTemplates.kafkaEphemeral(kafkaClusterSourceName, 1, 1).build());
+        resourceManager.createResource(extensionContext,
+                KafkaTemplates.kafkaEphemeral(kafkaClusterTargetName, 1, 1).build());
```
im-konge, line 383: Same as above - 3/3

systemtest/src/test/java/io/strimzi/systemtest/specific/RackAwarenessIsolatedST.java
```diff
@@ -0,0 +1,421 @@
...
+                KafkaMirrorMaker2Templates.kafkaMirrorMaker2(clusterName, kafkaClusterTargetName, kafkaClusterSourceName, 1, false)
+                        .editSpec()
+                            .withNewRack(topologyKey)
+                        .endSpec()
+                        .build());
+
+        LOGGER.info("MirrorMaker2 cluster deployed successfully");
+        String deployName = KafkaMirrorMaker2Resources.deploymentName(clusterName);
+        String podName = PodUtils.getPodNameByPrefix(namespace, deployName);
+        Pod pod = PodUtils.getPodByName(namespace, podName);
```
im-konge, line 395: Same as above

im-konge (commented):
I know that there are few things, that are not _your_ changes - inside the older tests. But can you please take a look at the comments which I left? Thanks for the changes!

systemtest/src/test/java/io/strimzi/systemtest/specific/RackAwarenessIsolatedST.java
```diff
@@ -0,0 +1,421 @@
...
+    @IsolatedTest
+    void testKafkaRackAwareness(ExtensionContext extensionContext) {
+        assumeFalse(Environment.isNamespaceRbacScope());
+
+        TestStorage storage = new TestStorage(extensionContext);
+        String namespace = storage.getNamespaceName();
+        String clusterName = storage.getClusterName();
+        String topologyKey = "kubernetes.io/hostname";
+
+        resourceManager.createResource(extensionContext, KafkaTemplates.kafkaEphemeral(clusterName, 1, 1)
```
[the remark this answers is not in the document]
fvaleri: Can you elaborate more on this? We have isolated tests here, so I think you are referring to a possible switch to parallel tests, which I'm actually working on and it seems to work fine.

systemtest/src/test/java/io/strimzi/systemtest/specific/RackAwarenessIsolatedST.java
```diff
@@ -0,0 +1,421 @@
...
+        resourceManager.createResource(extensionContext, KafkaConnectTemplates.kafkaConnect(extensionContext, storage.getClusterName(), 1)
+                .editSpec()
+                    .withNewRack(topologyKey)
+                .endSpec()
+                .build());
+
+        LOGGER.info("Connect cluster deployed successfully");
+        String deployName = KafkaConnectResources.deploymentName(clusterName);
+        String podName = PodUtils.getPodNameByPrefix(namespace, deployName);
+        Pod pod = PodUtils.getPodByName(namespace, podName);
```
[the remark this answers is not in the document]
fvaleri: In this case, I actually need these variables in the following code.

systemtest/src/test/java/io/strimzi/systemtest/specific/RackAwarenessIsolatedST.java
fvaleri: Yes, that's a good idea.

systemtest/src/test/java/io/strimzi/systemtest/specific/RackAwarenessIsolatedST.java
```diff
@@ -0,0 +1,421 @@
...
+class RackAwarenessIsolatedST extends AbstractST {
+    private static final Logger LOGGER = LogManager.getLogger(RackAwarenessIsolatedST.class);
+
+    @IsolatedTest
+    void testKafkaRackAwareness(ExtensionContext extensionContext) {
+        assumeFalse(Environment.isNamespaceRbacScope());
+
+        TestStorage storage = new TestStorage(extensionContext);
+        String namespace = storage.getNamespaceName();
+        String clusterName = storage.getClusterName();
```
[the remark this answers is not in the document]
fvaleri: I don't see any possible problem given that the context is relative to the current test, and I don't change these values during the test. Can you explain why this is an anti-pattern?

systemtest/src/test/java/io/strimzi/systemtest/specific/RackAwarenessIsolatedST.java
```diff
@@ -0,0 +1,421 @@
...
+@Tag(CONNECT)
+@Tag(CONNECT_COMPONENTS)
+@Tag(INTERNAL_CLIENTS_USED)
+@IsolatedSuite
+class RackAwarenessIsolatedST extends AbstractST {
+    private static final Logger LOGGER = LogManager.getLogger(RackAwarenessIsolatedST.class);
+
+    @IsolatedTest
+    void testKafkaRackAwareness(ExtensionContext extensionContext) {
+        assumeFalse(Environment.isNamespaceRbacScope());
```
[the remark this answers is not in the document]
fvaleri: Sure, and it works.

systemtest/src/main/java/io/strimzi/systemtest/utils/kubeUtils/objects/PodUtils.java
```diff
@@ -129,6 +129,12 @@ public static String getPodNameByPrefix(String namespaceName, String prefix) {
             .findFirst().orElseThrow().getMetadata().getName();
     }
 
+    public static Pod getPodByName(String namespace, String name) {
+        return kubeClient(namespace).listPods(namespace).stream()
+                .filter(pod -> pod.getMetadata().getName().equals(name))
+                .findFirst().orElseThrow();
+    }
+
```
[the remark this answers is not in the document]
fvaleri: Actually, we already have `public Pod getPod(String namespaceName, String name)` there, so I'll remove this new method.

systemtest/src/test/java/io/strimzi/systemtest/specific/RackAwarenessIsolatedST.java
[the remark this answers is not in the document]
im-konge: Why would you use something, that already contains the names in variables and then pass it to another variables? In that case you don't need the test storage or?

systemtest/src/test/java/io/strimzi/systemtest/specific/RackAwarenessIsolatedST.java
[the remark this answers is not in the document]
im-konge: Why it should be connected to parallel tests? When you use 1 Kafka and 1 ZK, the cluster is not stable - even in the warnings you can see that it's not recommended to use 1 Kafka and 1 ZK - rolling updates, message exchange, different operations.

What exactly is problem to set the 3/3 replicas?

systemtest/src/test/java/io/strimzi/systemtest/specific/RackAwarenessIsolatedST.java
```diff
@@ -0,0 +1,421 @@
...
+                    .editKafka()
+                        .withNewRack(topologyKey)
+                    .endKafka()
+                .endSpec()
+                .build());
+
+        LOGGER.info("Kafka cluster deployed successfully");
+        String ssName = KafkaResources.kafkaStatefulSetName(clusterName);
+        String podName = PodUtils.getPodNameByPrefix(namespace, ssName);
+        Pod pod = PodUtils.getPodByName(namespace, podName);
```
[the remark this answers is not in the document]
im-konge: I don't think that's easier to read - it's much better to specify, which pod do you want (at least that's what we are doing in most of tests).

systemtest/src/test/java/io/strimzi/systemtest/specific/RackAwarenessIsolatedST.java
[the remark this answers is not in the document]
im-konge: You will keep readability even when you use the methods from `TestStorage` :)

systemtest/src/test/java/io/strimzi/systemtest/specific/RackAwarenessIsolatedST.java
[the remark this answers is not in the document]
fvaleri: You talked about race conditions specifically, this is what I'm interested about. I know it's not recommended, but here I'm primarily testing the configuration and it's faster with just one replica. Switching to parallel execution is even faster (I did some changes to allow complete namespace isolation).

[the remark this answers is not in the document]
im-konge: Yeah, it's obviously faster. The thing is that when you are doing some additional checks, they can sometimes fail, because the cluster wasn't stable enough - that's what I mean when I'm saying _race conditions_.

Also, you can use one Kafka with 3/3 in `BeforeAll` as a shared cluster and then, in each test, deploy different resources (and additional Kafka for MM2 test)

fvaleri:
@im-konge this is the parallel version of previous tests. I ran it multiple times on my local minikube and on a remote OpenShift cluster. The previous isolated suite took around 26min to complete, while now we are down to 8min (with parallelism=5). I think this is good, also considering that we would need to add tests for other rack-aware components in the future (i.e. Bridge, TO).

I removed the part which uses `KafkaConnectUtils.sendReceiveMessagesThroughConnect` for two reasons: 1st this is not strictly related to rack awareness, and 2nd file connector will not be loaded anymore in Kafka 3.2.0, see [KAFKA-13748](https://issues.apache.org/jira/browse/KAFKA-13748). We will need a fix to explicitly add it to `plugin.path` to avoid breaking some Connect system tests. I can open an issue to track this.

im-konge:
@fvaleri yes, this looks good. How about the shared Kafka cluster, as you are creating the same cluster for each test.

If you could create an issue to track that, it would be really great. Also, thanks for spotting this :)

fvaleri:
I created that issue. About the shared cluster, I would prefer to keep it like this. The first test has some additional configuration, and we may need to customize the Kafka cluster in future tests. In the end, we would trade flexibility for a small speedup. Instead, I could merge the two Connect tests into one, so that we have one tagged test for each component, what about this?

im-konge:
> I could merge the two Connect tests into one, so that we have one tagged test for each component, what about this?

Okay, that would be great

fvaleri:
I did some minor changes and rebased, so I don't think this IT failure is due to my latest commit.

systemtest/src/test/java/io/strimzi/systemtest/specific/RackAwarenessST.java
```diff
@@ -0,0 +1,206 @@
...
+    void setUp() {
+        assumeFalse(Environment.isNamespaceRbacScope());
+        clusterOperator.unInstall();
+        clusterOperator = clusterOperator
+                .defaultInstallation()
+                .withOperationTimeout(CO_OPERATION_TIMEOUT_SHORT)
+                .withReconciliationInterval(Constants.RECONCILIATION_INTERVAL)
+                .createInstallation()
+                .runInstallation();
+    }
```
see-quick, line 205: Just a minor comment before my whole review. You cannot modify the configuration of the `shared` Cluster Operator when you label the whole test suite as `@ParallelSuite`. It should be `@IsolatedSuite` and the name has to be `RackAwarenessIsolatedST`.

fvaleri: Correct, thanks @see-quick for spotting this. We may actually don't need these configurations, as in `testConnectRackAwareness` we can simply check that the pod is unschedulable with an invalid `topologyKey`, without waiting for the CO timeout. Let me do and test this change.

As a side note, the CO customization in parallel suites could be a common pitfall, especially when converting from isolated test suites (like in this case). I think we should prevent that from happening in the first place by raising a runtime error at the ST framework level. WDYT?

see-quick: > I think we should prevent that from happening in the first place by raising a runtime error at the ST framework level. WDYT?

Yes, I totally agree. Plus if there is no need for CO configuration, you can remove it completely because the shared CO for `@ParallelSuites` is deployed at the start of the execution in `BeforeAllOnce.class`.

see-quick (approved):
Thanks for the PR @fvaleri.

systemtest/src/test/java/io/strimzi/systemtest/specific/RackAwarenessST.java
```diff
@@ -0,0 +1,184 @@
...
+        String podNodeName = pod.getSpec().getNodeName();
+        String hostname = podNodeName.contains(".") ? podNodeName.substring(0, podNodeName.indexOf(".")) : podNodeName;
+        String commandOut = cmdKubeClient(namespace).execInPod(podName, "/bin/bash", "-c", "cat /tmp/strimzi-connect.properties | grep consumer.client.rack").out().trim();
+        assertThat(commandOut.equals("consumer.client.rack=" + hostname), is(true));
+    }
+
+    @BeforeEach
+    void createTestResources(ExtensionContext extensionContext) {
+        storageMap.put(extensionContext, new TestStorage(extensionContext));
+    }
```
im-konge, line 183: Why do you need this?

systemtest/src/test/java/io/strimzi/systemtest/specific/RackAwarenessST.java
```diff
@@ -0,0 +1,184 @@
...
+@ParallelSuite
+class RackAwarenessST extends AbstractST {
+    private static final Logger LOGGER = LogManager.getLogger(RackAwarenessST.class);
+    private static final String TOPOLOGY_KEY = "kubernetes.io/hostname";
+
+    @ParallelNamespaceTest
+    void testKafkaRackAwareness(ExtensionContext extensionContext) {
+        TestStorage storage = storageMap.get(extensionContext);
+        String namespace = storage.getNamespaceName();
+        String clusterName = storage.getClusterName();
```
im-konge, line 51: So, what about this? In all other tests we are using just the `TestStorage`. The idea why was `TestStorage` created is, that we wanted to get rid of all kind of variables inside the tests (variables like `clusterName`, `namespaceName`, `kafkaLabelSelector` etc.). When you create new variables, to which you pass the `TestStorage` stuff, you will have two duplicates, which is not correct FMPOV.

So, how I mentioned earlier, you have two options:

1. use the `TestStorage` and the getters
2. remove the `TestStorage` usage and pass it directly to the variables, to not have duplicates etc.

systemtest/src/test/java/io/strimzi/systemtest/specific/RackAwarenessST.java
```diff
@@ -0,0 +1,184 @@
...
+                        .withNewRack(TOPOLOGY_KEY)
+                        .addToConfig("replica.selector.class", "org.apache.kafka.common.replica.RackAwareReplicaSelector")
+                    .endKafka()
+                .endSpec()
+                .build());
+
+        LOGGER.info("Kafka cluster deployed successfully");
+        String ssName = KafkaResources.kafkaStatefulSetName(clusterName);
+        String podName = PodUtils.getPodNameByPrefix(namespace, ssName);
+        Pod pod = kubeClient().getPod(namespace, podName);
```
im-konge, line 65: About this, it would be much clearer to use just `Pod pod = kubeClient().getPod(namespace, KafkaResources.kafkaPodName(testStorage.getClusterName(), 0);` then doing these extra operations.

systemtest/src/test/java/io/strimzi/systemtest/specific/RackAwarenessST.java
```diff
@@ -0,0 +1,184 @@
...
+@Tag(REGRESSION)
+@ParallelSuite
+class RackAwarenessST extends AbstractST {
+    private static final Logger LOGGER = LogManager.getLogger(RackAwarenessST.class);
+    private static final String TOPOLOGY_KEY = "kubernetes.io/hostname";
+
+    @ParallelNamespaceTest
+    void testKafkaRackAwareness(ExtensionContext extensionContext) {
+        TestStorage storage = storageMap.get(extensionContext);
+        String namespace = storage.getNamespaceName();
```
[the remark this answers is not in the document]
fvaleri: Fair enough. I removed some of these references.

systemtest/src/test/java/io/strimzi/systemtest/specific/RackAwarenessST.java
fvaleri: Yeah, that's for consistency with the other tests, which actually requires them to avoid repetition.

im-konge (approved):

scholzj (approved):

PR #6716, opened 2022-04-21
[systemtest] Add test for podset reconciliation only switch

im-konge:
### Type of change

- New test

### Description

This PR adds new test for the "podset reconciliation only" feature, which was added in #6541 .

I created new test suite - `PodSetIsolatedST` - which will contain all tests related to StrimziPodSets. The test itself is checking that when the `STRIMZI_POD_SET_RECONCILIATION_ONLY` env is used (together with SPS FG on), any change to the Kafka CR are not reflected, thus the Kafka and ZK will not be rolled. But, if the Kafka or ZK pods are deleted, they should be properly recreated.

Also, this PR adds new method `replaceDeployment` into the `DeploymentResource`, so we can easily edit deployments the same way how we are editing resources like Kafka, KafkaConnect etc.

### Checklist

- [x] Write tests

### DIFF
--- a/systemtest/src/test/java/io/strimzi/systemtest/operators/PodSetIsolatedST.java
+++ b/systemtest/src/test/java/io/strimzi/systemtest/operators/PodSetIsolatedST.java
@@ -1,25 +1,114 @@
+/*
+ * Copyright Strimzi authors.
+ * License: Apache License 2.0 (see the file LICENSE or http://apache.org/licenses/LICENSE-2.0.html).
+ */
 package io.strimzi.systemtest.operators;
 
+import io.fabric8.kubernetes.api.model.EnvVar;
+import io.strimzi.api.kafka.model.KafkaResources;
+import io.strimzi.api.kafka.model.ProbeBuilder;
 import io.strimzi.systemtest.AbstractST;
 import io.strimzi.systemtest.BeforeAllOnce;
 import io.strimzi.systemtest.Constants;
-import io.strimzi.systemtest.annotations.StrimziPodSetTest;
+import io.strimzi.systemtest.Environment;
+import io.strimzi.systemtest.annotations.IsolatedSuite;
+import io.strimzi.systemtest.annotations.IsolatedTest;
+import io.strimzi.systemtest.resources.crd.KafkaResource;
+import io.strimzi.systemtest.resources.kubernetes.DeploymentResource;
 import io.strimzi.systemtest.resources.operator.SetupClusterOperator;
+import io.strimzi.systemtest.storage.TestStorage;
+import io.strimzi.systemtest.templates.crd.KafkaTemplates;
+import io.strimzi.systemtest.utils.RollingUpdateUtils;
+import io.strimzi.systemtest.utils.kubeUtils.controllers.DeploymentUtils;
+import io.strimzi.systemtest.utils.kubeUtils.objects.PodUtils;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
 import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.Tag;
 import org.junit.jupiter.api.extension.ExtensionContext;
 
-import static io.strimzi.systemtest.Constants.INFRA_NAMESPACE;
+import java.util.Collections;
+import java.util.List;
+import java.util.Map;
 
-@StrimziPodSetTest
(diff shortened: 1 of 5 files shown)


systemtest/src/test/java/io/strimzi/systemtest/operators/PodSetIsolatedST.java
```diff
@@ -0,0 +1,115 @@
...
+import java.util.Map;
+
+import static io.strimzi.systemtest.Constants.INFRA_NAMESPACE;
+import static io.strimzi.systemtest.Constants.REGRESSION;
+import static io.strimzi.systemtest.Constants.STRIMZI_DEPLOYMENT_NAME;
+import static io.strimzi.systemtest.resources.ResourceManager.kubeClient;
+
+@Tag(REGRESSION)
+@IsolatedSuite
+public class PodSetIsolatedST extends AbstractST {
```
see-quick, line 41: What about some Javadoc to this new TestSuite (I know it's related to the PodSet but for the newcomers, it could be a good starting point)

systemtest/src/test/java/io/strimzi/systemtest/operators/PodSetIsolatedST.java
```diff
@@ -0,0 +1,115 @@
...
+        LOGGER.info("Because the configuration was changed, pods should be rolled");
+
+        RollingUpdateUtils.waitTillComponentHasRolledAndPodsReady(testStorage.getNamespaceName(), testStorage.getZookeeperSelector(), replicas, kafkaPods);
+        RollingUpdateUtils.waitTillComponentHasRolledAndPodsReady(testStorage.getNamespaceName(), testStorage.getKafkaSelector(), replicas, zkPods);
+    }
+
+    @BeforeAll
+    void setup() {
+        clusterOperator.unInstall();
+        clusterOperator = new SetupClusterOperator.SetupClusterOperatorBuilder()
```
see-quick, line 108: instead of pure instantiating with the `new` keyword, you can use something like this:
```java
clusterOperator = clusterOperator
            .defaultInstallation()
            .withExtraEnvVars(INITIAL_ENV_VARS)
            .createInstallation()
            .runInstallation();
```
the context and namespace will be fulfilled in `.defaultInstallation` step.

see-quick (approved):
Overall looks good, I have just a few a nits... :)

systemtest/src/test/java/io/strimzi/systemtest/operators/PodSetIsolatedST.java
im-konge: Good idea, thanks :)

systemtest/src/test/java/io/strimzi/systemtest/operators/PodSetIsolatedST.java
```diff
@@ -0,0 +1,117 @@
...
+
+        LOGGER.info("Changing Kafka resource configuration, the pods should not be rolled");
+
+        KafkaResource.replaceKafkaResourceInSpecificNamespace(testStorage.getClusterName(),
+            kafka -> kafka.getSpec().getZookeeper().setReadinessProbe(new ProbeBuilder().withTimeoutSeconds(probeTimeoutSeconds).build()), testStorage.getNamespaceName());
+        RollingUpdateUtils.waitForNoRollingUpdate(testStorage.getNamespaceName(), testStorage.getZookeeperSelector(), zkPods);
+
+        KafkaResource.replaceKafkaResourceInSpecificNamespace(testStorage.getClusterName(),
+            kafka -> kafka.getSpec().getKafka().setReadinessProbe(new ProbeBuilder().withTimeoutSeconds(probeTimeoutSeconds).build()), testStorage.getNamespaceName());
+        RollingUpdateUtils.waitForNoRollingUpdate(testStorage.getNamespaceName(), testStorage.getKafkaSelector(), kafkaPods);
```
scholzj, line 82: Any chance you can change both at once and wait once? That would save some execution time and the value will be the same.

scholzj (approved):
One possible improvement. But LGTM otherwise.

systemtest/src/test/java/io/strimzi/systemtest/operators/PodSetIsolatedST.java
im-konge: Makes sense, good catch :) thanks

PR #6727, opened 2022-04-22
MockKube2: Mock Kubernetes environment based on Fabric8 mock server

scholzj:
### Type of change

- Enhancement / new feature

### Description

This PR adds _MockKube2_ -> Kubernetes mock server based on the Fabric8 Kubernetes Mock. Fabric8 Kubernetes Mock provides the core functionality. The MockKube2 adds additional support for our CRDs and also mocked Kubernetes  controllers for StatefulSets, Pods, Deployments or Services. Unlike the original Strimzi MockKube, this does not use the Fabric8 client mocked with Mockito but a proper HTTP server. So while the Kubernetes controllers are not real Kubernetes, but just very simplified utilities creating pods from StatefulSets, making them ready etc., for the Strimzi code this is a proper HTTP connection.

The advantages of this approach is more realistic test results and also better support for advanced features such as watches or informers (informers are not supported in our MockKube). One of the disadvantages is speed of execution. In some simple testes, the Fabric8 mock server is significantly faster. But in others significantly slower. The overall execution is definitely several minutes slower. This is not necessarily suprising, because everything happens over HTTP and not just through mocks and maps internally. Also, the mock Kubernetes controllers are using the regular Kubernetes client and HTTP, so they are also slower than the original MockKube which also mocks these controllers, but only internally in memory and not over an Kube client connected over HTTP. So the result is not completely surprising.

This PR does not remove the original MockKube. It is still used by `KafkaRebalanceAssemblyOperatorTest` which is based on a very weird combination of using the MockKube but mixing it heavily with inline mocks inside the tests and really using MockKube only for subset of resources. This would need to be rewritten before we can remove the original code.

### Checklist

- [x] Write tests


mockkube/src/main/java/io/strimzi/test/mockkube2/MockKube2.java
```diff
@@ -0,0 +1,160 @@
...
+import io.strimzi.test.mockkube2.controllers.AbstractMockController;
+import io.strimzi.test.mockkube2.controllers.MockDeploymentController;
+import io.strimzi.test.mockkube2.controllers.MockPodController;
+import io.strimzi.test.mockkube2.controllers.MockServiceController;
+import io.strimzi.test.mockkube2.controllers.MockStatefulSetController;
+
+import java.util.ArrayList;
+import java.util.List;
+
+public class MockKube2 {
```
tombentley, line 33: Can you add a little javadoc explaining how it works?

mockkube/src/main/java/io/strimzi/test/mockkube2/MockKube2.java
```diff
@@ -0,0 +1,160 @@
...
+            mock.addController(new MockPodController(client));
+            return this;
+        }
+
+        public MockKube2Builder withServiceController()  {
+            mock.addController(new MockServiceController(client));
+            return this;
+        }
+
+        public MockKube2Builder withKafkaCrd()  {
```
tombentley, line 98: For the CRDs I wonder if we shouldn't:

1. Register them all automatically?
2. Provide an API in the `api` module for discovering all the CRDs. Something as simple as a `static` method that returned a list of some record class of group, Kind, CR class, List class? Using that here any appropriately elsewhere would make it easier to add new CRDs.

mockkube/src/main/java/io/strimzi/test/mockkube2/controllers/AbstractMockController.java
```diff
@@ -0,0 +1,18 @@
+/*
+ * Copyright Strimzi authors.
+ * License: Apache License 2.0 (see the file LICENSE or http://apache.org/licenses/LICENSE-2.0.html).
+ */
+package io.strimzi.test.mockkube2.controllers;
+
+import io.fabric8.kubernetes.client.KubernetesClient;
+
+public abstract class AbstractMockController {
```
tombentley, line 9: A sentence of javadoc on why this abstraction is needed/what it does

tombentley (approved):
Thanks for this Jakub!

mockkube/src/main/java/io/strimzi/test/mockkube2/MockKube2.java
scholzj: > Register them all automatically?

TBH, it seems unnecessary. We never use more than 2 CRDs really in the tests. Plus this actually creates the CRDs in the mock server, so I thought that maybe it is better to save the time.

> Provide an API in the api module for discovering all the CRDs. Something as simple as a static method that returned a list of some record class of group, Kind, CR class, List class? Using that here any appropriately elsewhere would make it easier to add new CRDs.

We actually have some part of this. We have for example this [array](https://github.com/strimzi/strimzi-kafka-operator/blob/198e1e639a165934d9e8df1bff1dc1986a5e7521/api/src/main/java/io/strimzi/api/kafka/Crds.java#L40) (although currently private). We have also a method for registering all CRDs for deserialisation. However, the problem is that to register CRD for the mock server, we also need the YAMLs of the CRDs. And this does not exists in the `api` module (as it is only generated by a combination of the `api` module and the `crd-generator`). So I cannot use this directly. I would need to pair it somewhere with the YAMLs first. Any ideas how to do it in some elegant way?

tombentley: We could have a new module (`crds`) that:

1. Depended on the `api` module
2. Included the generated CRDs as classpath resources
3. Included the kind of discovery API I'm talking about

I'm not even sure whether it needs to be a separate module; since we generate the CRDs when building the `api` module we could just add them to the `src/main/resources`. Same-jar is attractive because it makes it impossible for the YAMLs to end up being a different version than the classes (e.g. due to messed up poms).

We might (eventually) want some kind of way to check/assert that the classes in the api module are compatible with the CRDs existing in an apiserver. I'm not meaning Kubernetes' idea of CRD schema compatibility here. It would be to prevent people using more recent jars against and API server with older installed CRDs. A simple way of achieving this would be to annotate the CRDs with an integer version which we increment with each release, and also include that integer in the api module, then check that `crd version from api server <= crd version from jar`.

tombentley: Yes, that would be fine @scholzj. Thanks.

scholzj: Raised #6767

Issue #6650, opened 2022-04-07
Healthcheck for Kubernetes cluster


see-quick:
### Problem description

Currently, in our system tests, we do not have any kind of mechanism, which would check the health of the Kubernetes cluster. This could potentially cause a problem. One possible scenario is that cluster is from start `healthy` but during verification, it will be not responsible at all (because of the OOM problem). 

**Describe the solution you'd like**

`1. approach`

We should implement such health checks in our system tests. Specifically in the all`@*All/Each` scopes. Moreover, we can use the already implement a solution by Kubernetes [1] and just call this command:

```java
kubectl get --raw='/readyz?verbose'
```

`2.approach`

Use already implemented method from `io.strimzi.test.k8s.cluster.KubeCluster` interface `isClusterUp()`. The method will be used in ∀ scopes (i.e., `@*All/Each`) because the problem with network stability or OOM issue may occur anywhere.

**Additional context**

[1] -> https://kubernetes.io/docs/reference/using-api/health-checks/

see-quick:
Done in https://github.com/strimzi/strimzi-kafka-operator/pull/6748

PR #6748, opened 2022-04-29
[MO] - [system test] -> Cluster stability check

see-quick:

### Type of change

- Enhancement / new feature
- Refactoring

### Description

Implements `2.approach` from https://github.com/strimzi/strimzi-kafka-operator/issues/6650.

This PR solves issues when for instance during verification a cluster crashes and our tests continue deploying resources, which results in waiting. Now before and after all, test cases or test suites we check if the cluster is available (i.e., we can use Kubernetes API without any problem). If the cluster crashes, then we throw `KubernetesClusterUnstableException` and we also skip the collecting of the Pods (which in this case is not reasonable).
 
### Checklist

### DIFF
--- a/systemtest/src/main/java/io/strimzi/systemtest/exceptions/KubernetesClusterUnstableException.java
+++ b/systemtest/src/main/java/io/strimzi/systemtest/exceptions/KubernetesClusterUnstableException.java
@@ -1,0 +1,17 @@
+/*
+ * Copyright Strimzi authors.
+ * License: Apache License 2.0 (see the file LICENSE or http://apache.org/licenses/LICENSE-2.0.html).
+ */
+package io.strimzi.systemtest.exceptions;
+
+/**
+ * Provides an exception, which is used in cases where cluster may not be responding and its most likely broken.
+ * This could be caused by network or out of memory problem.
+ */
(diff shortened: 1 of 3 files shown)


Frawless (approved):
Looks good, you should verify it with OCP and Minikube as well to be sure that everything is working fine.

see-quick:
> Looks good, you should verify it with OCP and Minikube as well to be sure that everything is working fine.

Yes, I have tried it and it works. Meaning I execute the acceptance profile in parallel and start with fresh cluster and then just delete it (minikube) or remove the session (ocp), which do the job.

scholzj (commented):
Again, better description in the PR would not hurt. It would be great to explain for example what will it do when the cluster is not ready and what to expect after merging this.

see-quick:
> Again, a better description in the PR would not hurt. It would be great to explain for example what will it do when the cluster is not ready and what to expect after merging this.

Done :) I thought that referencing the issue is enough.

see-quick:
> Or will it try for each test / test suite again and again?

It will try for each test suite again, but that's a matter of 30s.

scholzj (approved):
Thanks for the explanation 👍

PR #6749, opened 2022-04-29
[ST Enhancement] Deletion of remaining pvcs after tests.

jankalinic:
Signed-off-by: jkalinic <jkalinic@redhat.com>

### Type of change

_Select the type of your PR_

- Bugfix
- Enhancement / new feature
- Refactoring
- Documentation

### Description

This PR consists of simple enhancement for system tests. PVCs that have annotation 'deletionClaim' set to false, won't be deleted by default after the tests. So this PR contains deletion strategy for those PVCs as we want to clear every resources afterwards.

### Checklist

- [ ] Write tests
- [ ] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/systemtest/src/main/java/io/strimzi/systemtest/resources/crd/KafkaResource.java
+++ b/systemtest/src/main/java/io/strimzi/systemtest/resources/crd/KafkaResource.java
@@ -25,6 +25,7 @@ import org.apache.logging.log4j.Logger;
 import org.junit.platform.commons.util.Preconditions;
 
 import java.util.HashMap;
+import java.util.List;
 import java.util.Map;
 import java.util.function.Consumer;
 import java.util.function.Predicate;
@@ -79,9 +80,14 @@ public class KafkaResource implements ResourceType<Kafka> {
             resource.getMetadata().getName()).withPropagationPolicy(DeletionPropagation.FOREGROUND).delete();
 
         // additional deletion of pvcs with specification deleteClaim set to false which were not deleted prior this method
-        for (PersistentVolumeClaim pvc : kubeClient().listPersistentVolumeClaims(namespaceName, clusterName)) {
-            kubeClient().deletePersistentVolumeClaim(namespaceName, pvc.getMetadata().getName());
-            PersistentVolumeClaimUtils.waitForPersistentVolumeClaimDeletion(namespaceName, pvc.getMetadata().getName());
+        List<PersistentVolumeClaim> persistentVolumeClaimsList = kubeClient().listPersistentVolumeClaims(namespaceName, clusterName);
+
+        for (PersistentVolumeClaim persistentVolumeClaim : persistentVolumeClaimsList) {
+            kubeClient().deletePersistentVolumeClaim(namespaceName, persistentVolumeClaim.getMetadata().getName());
+        }
+
+        for (PersistentVolumeClaim persistentVolumeClaim : persistentVolumeClaimsList) {
+            PersistentVolumeClaimUtils.waitForPersistentVolumeClaimDeletion(namespaceName, persistentVolumeClaim.getMetadata().getName());
         }
     }
 
--- a/systemtest/src/main/java/io/strimzi/systemtest/resources/crd/KafkaResource.java
+++ b/systemtest/src/main/java/io/strimzi/systemtest/resources/crd/KafkaResource.java
@@ -80,8 +80,8 @@ public class KafkaResource implements ResourceType<Kafka> {
 
         // additional deletion of pvcs with specification deleteClam set to false which were not deleted prior this method
         for (PersistentVolumeClaim pvc : kubeClient().listPersistentVolumeClaims(namespaceName, clusterName)) {
-            kubeClient().deletePVC(namespaceName, pvc.getMetadata().getName());
-            PersistentVolumeClaimUtils.waitForPVCDeletion(namespaceName, pvc.getMetadata().getName());
+            kubeClient().deletePvc(namespaceName, pvc.getMetadata().getName());
+            PersistentVolumeClaimUtils.waitForPvcDeletion(namespaceName, pvc.getMetadata().getName());
         }
     }
 
--- a/systemtest/src/main/java/io/strimzi/systemtest/resources/crd/KafkaResource.java
+++ b/systemtest/src/main/java/io/strimzi/systemtest/resources/crd/KafkaResource.java
@@ -81,7 +81,7 @@ public class KafkaResource implements ResourceType<Kafka> {
         // additional deletion of pvcs with specification deleteClaim set to false which were not deleted prior this method
         for (PersistentVolumeClaim pvc : kubeClient().listPersistentVolumeClaims(namespaceName, clusterName)) {
             kubeClient().deletePersistentVolumeClaim(namespaceName, pvc.getMetadata().getName());
-            PersistentVolumeClaimUtils.waitForPvcDeletion(namespaceName, pvc.getMetadata().getName());
+            PersistentVolumeClaimUtils.waitForPersistentVolumeClaimDeletion(namespaceName, pvc.getMetadata().getName());
         }
     }
 
--- a/systemtest/src/main/java/io/strimzi/systemtest/resources/crd/KafkaResource.java
+++ b/systemtest/src/main/java/io/strimzi/systemtest/resources/crd/KafkaResource.java
@@ -78,7 +78,7 @@ public class KafkaResource implements ResourceType<Kafka> {
         kafkaClient().inNamespace(namespaceName).withName(
             resource.getMetadata().getName()).withPropagationPolicy(DeletionPropagation.FOREGROUND).delete();
 
-        // additional deletion of pvcs with specification deleteClam set to false which were not deleted prior this method
+        // additional deletion of pvcs with specification deleteClaim set to false which were not deleted prior this method
         for (PersistentVolumeClaim pvc : kubeClient().listPersistentVolumeClaims(namespaceName, clusterName)) {
             kubeClient().deletePersistentVolumeClaim(namespaceName, pvc.getMetadata().getName());
             PersistentVolumeClaimUtils.waitForPvcDeletion(namespaceName, pvc.getMetadata().getName());
--- a/systemtest/src/main/java/io/strimzi/systemtest/resources/crd/KafkaResource.java
+++ b/systemtest/src/main/java/io/strimzi/systemtest/resources/crd/KafkaResource.java
@@ -7,6 +7,7 @@ package io.strimzi.systemtest.resources.crd;
 import io.fabric8.kubernetes.api.model.DeletionPropagation;
 import io.fabric8.kubernetes.api.model.LabelSelector;
 import io.fabric8.kubernetes.api.model.LabelSelectorBuilder;
+import io.fabric8.kubernetes.api.model.PersistentVolumeClaim;
 import io.fabric8.kubernetes.client.dsl.MixedOperation;
 import io.fabric8.kubernetes.client.dsl.Resource;
 import io.strimzi.api.kafka.Crds;
@@ -18,6 +19,7 @@ import io.strimzi.systemtest.Constants;
 import io.strimzi.systemtest.resources.ResourceManager;
 import io.strimzi.systemtest.resources.ResourceOperation;
 import io.strimzi.systemtest.resources.ResourceType;
+import io.strimzi.systemtest.utils.kubeUtils.objects.PersistentVolumeClaimUtils;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.junit.platform.commons.util.Preconditions;
@@ -28,6 +30,7 @@ import java.util.function.Consumer;
 import java.util.function.Predicate;
 
 import static io.strimzi.systemtest.enums.CustomResourceStatus.Ready;
+import static io.strimzi.test.k8s.KubeClusterResource.kubeClient;
 
 public class KafkaResource implements ResourceType<Kafka> {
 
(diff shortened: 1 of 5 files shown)


test/src/main/java/io/strimzi/test/k8s/KubeClient.java
```diff
@@ -232,6 +232,14 @@ public List<String> listPodNamesInSpecificNamespace(String namespaceName, String
             .collect(Collectors.toList());
     }
 
+    public PersistentVolumeClaim getPVC(String namespaceName, String pvcName) {
```
see-quick, line 235: Camel case and same for delete
```suggestion
    public PersistentVolumeClaim getPvc(String namespaceName, String pvcName) {
```

systemtest/src/main/java/io/strimzi/systemtest/utils/kubeUtils/objects/PersistentVolumeClaimUtils.java
```diff
@@ -66,12 +67,24 @@ public static void waitUntilPVCAnnotationChange(String namespaceName, String clu
...
 
-    public static void waitForPVCDeletion(String namespaceName, int volumesCount, JbodStorage jbodStorage, String clusterName) {
+    public static void waitForPVCDeletion(String namespaceName, String pvcName) {
+        TestUtils.waitFor("Wait for PVC deletion", Constants.POLL_INTERVAL_FOR_RESOURCE_DELETION, Duration.ofMinutes(3).toMillis(), () -> {
+            if (kubeClient(namespaceName).getPVC(namespaceName, pvcName) == null) {
+                return true;
+            } else {
+                LOGGER.warn("PVC {} is not deleted yet! Triggering force delete by cmd client!", pvcName);
+                cmdKubeClient(namespaceName).deleteByName("pvc", pvcName);
+                return false;
```
see-quick, line 77: Do we need to trigger this operation multiple times? :) Are there some scenarios, where PVCs are stuck or why do need it? :)

test/src/main/java/io/strimzi/test/k8s/KubeClient.java
```diff
@@ -232,6 +232,14 @@ public List<String> listPodNamesInSpecificNamespace(String namespaceName, String
...
+        return client.persistentVolumeClaims().inNamespace(namespaceName).withName(pvcName).get();
+    }
+
+    public boolean deletePVC(String namespaceName, String pvcName) {
```
see-quick, line 239:
```suggestion
    public boolean deletePvc(String namespaceName, String pvcName) {
```

see-quick (commented):
Overall looks good, I just have a few nits.

test/src/main/java/io/strimzi/test/k8s/KubeClient.java
im-konge: Why camel case here? It's an abbreviation for `PersistentVolumeClaim`. For `StrimziPodSet` it's also SPS.

systemtest/src/main/java/io/strimzi/systemtest/utils/kubeUtils/objects/PersistentVolumeClaimUtils.java
Frawless: do we have this in "force" delete command `--force --grace-period=0 ` ? It might speed up the deletion a lot

systemtest/src/main/java/io/strimzi/systemtest/utils/kubeUtils/objects/PersistentVolumeClaimUtils.java
```diff
@@ -66,12 +67,24 @@ public static void waitUntilPVCAnnotationChange(String namespaceName, String clu
...
+        });
+    }
+
+    public static void waitForJbodDeletion(String namespaceName, int volumesCount, JbodStorage jbodStorage, String clusterName) {
         int numberOfPVCWhichShouldBeDeleted = jbodStorage.getVolumes().stream().filter(
             singleVolumeStorage -> ((PersistentClaimStorage) singleVolumeStorage).isDeleteClaim()
         ).collect(Collectors.toList()).size();
 
-        TestUtils.waitFor("Wait for PVC deletion", Constants.POLL_INTERVAL_FOR_RESOURCE_DELETION, Duration.ofMinutes(6).toMillis(), () -> {
+        TestUtils.waitFor("Wait for JBOD deletion", Constants.POLL_INTERVAL_FOR_RESOURCE_DELETION, Duration.ofMinutes(6).toMillis(), () -> {
```
Frawless, line 87:
```suggestion
        TestUtils.waitFor("Wait for JBOD storage deletion", Constants.POLL_INTERVAL_FOR_RESOURCE_DELETION, Duration.ofMinutes(6).toMillis(), () -> {
```

mozna zmenit ja nazev te metody?

systemtest/src/main/java/io/strimzi/systemtest/utils/kubeUtils/objects/PersistentVolumeClaimUtils.java
Frawless: Anyway, it should be part of different PR, because it could influence much more parts than current changes

systemtest/src/main/java/io/strimzi/systemtest/utils/kubeUtils/objects/PersistentVolumeClaimUtils.java
```diff
@@ -66,12 +67,24 @@ public static void waitUntilPVCAnnotationChange(String namespaceName, String clu
...
 
-    public static void waitForPVCDeletion(String namespaceName, int volumesCount, JbodStorage jbodStorage, String clusterName) {
+    public static void waitForPvcDeletion(String namespaceName, String pvcName) {
+        TestUtils.waitFor("Wait for PVC deletion", Constants.POLL_INTERVAL_FOR_RESOURCE_DELETION, Duration.ofMinutes(3).toMillis(), () -> {
+            if (kubeClient(namespaceName).getPvc(namespaceName, pvcName) == null) {
+                return true;
+            } else {
+                LOGGER.warn("PVC {} is not deleted yet! Triggering force delete by cmd client!", pvcName);
+                cmdKubeClient(namespaceName).deleteByName("pvc", pvcName);
+                return false;
```
im-konge, line 77:
```suggestion
            if (kubeClient(namespaceName).getPvc(namespaceName, pvcName) != null)  {
                LOGGER.warn("PVC {} is not deleted yet! Triggering force delete by cmd client!", pvcName);
                cmdKubeClient(namespaceName).deleteByName("pvc", pvcName);
                return false;
            }
            
            return true;
```

test/src/main/java/io/strimzi/test/k8s/KubeClient.java
```diff
@@ -232,6 +232,14 @@ public List<String> listPodNamesInSpecificNamespace(String namespaceName, String
...
+        return client.persistentVolumeClaims().inNamespace(namespaceName).withName(pvcName).get();
+    }
+
+    public boolean deletePvc(String namespaceName, String pvcName) {
+        return client.persistentVolumeClaims().inNamespace(namespaceName).withName(pvcName).delete();
+    }
+
     public List<PersistentVolumeClaim> listPersistentVolumeClaims(String namespaceName, String clusterName) {
         return client.persistentVolumeClaims().inNamespace(namespaceName).list().getItems().stream()
             .filter(persistentVolumeClaim -> persistentVolumeClaim.getMetadata().getName().contains(clusterName))
```
im-konge, line 245: One method has name `listPersistentVolumeClaims`, other two have `getPvc` or `deletePvc` -> I would either keep the abbreviation or long name, not both.

see-quick (approved):

test/src/main/java/io/strimzi/test/k8s/KubeClient.java
```diff
@@ -232,6 +232,14 @@ public List<String> listPodNamesInSpecificNamespace(String namespaceName, String
             .collect(Collectors.toList());
     }
 
+    public PersistentVolumeClaim getPersistentVolumeClaim(String namespaceName, String pvcName) {
+        return client.persistentVolumeClaims().inNamespace(namespaceName).withName(pvcName).get();
+    }
+
+    public boolean deletePersistentVolumeClaims(String namespaceName, String pvcName) {
```
im-konge, line 239:
```suggestion
    public boolean deletePersistentVolumeClaim(String namespaceName, String pvcName) {
```

systemtest/src/main/java/io/strimzi/systemtest/utils/kubeUtils/objects/PersistentVolumeClaimUtils.java
```diff
@@ -66,12 +67,24 @@ public static void waitUntilPVCAnnotationChange(String namespaceName, String clu
         LOGGER.info("PVC annotation has changed {}", newAnnotation.toString());
     }
 
-    public static void waitForPVCDeletion(String namespaceName, int volumesCount, JbodStorage jbodStorage, String clusterName) {
+    public static void waitForPvcDeletion(String namespaceName, String pvcName) {
+        TestUtils.waitFor("Wait for PVC deletion", Constants.POLL_INTERVAL_FOR_RESOURCE_DELETION, Duration.ofMinutes(3).toMillis(), () -> {
+            if (kubeClient(namespaceName).getPersistentVolumeClaim(namespaceName, pvcName) != null) {
```
im-konge, line 72:
```suggestion
            if (kubeClient().getPersistentVolumeClaim(namespaceName, pvcName) != null) {
```
you can remove it from the `kubeClient()` itself, as it will take only the `namespaceName` from the method parameter (in `getPersistentVolumeClaim`)

systemtest/src/main/java/io/strimzi/systemtest/utils/kubeUtils/objects/PersistentVolumeClaimUtils.java
```diff
@@ -66,12 +67,24 @@ public static void waitUntilPVCAnnotationChange(String namespaceName, String clu
         LOGGER.info("PVC annotation has changed {}", newAnnotation.toString());
     }
 
-    public static void waitForPVCDeletion(String namespaceName, int volumesCount, JbodStorage jbodStorage, String clusterName) {
+    public static void waitForPvcDeletion(String namespaceName, String pvcName) {
+        TestUtils.waitFor("Wait for PVC deletion", Constants.POLL_INTERVAL_FOR_RESOURCE_DELETION, Duration.ofMinutes(3).toMillis(), () -> {
```
im-konge, line 71: Also, can we take a timeout from `Constants` or, if there will be any matching, add it there this one?

im-konge (approved):
Few small nits. LGTM otherwise :)

systemtest/src/main/java/io/strimzi/systemtest/resources/crd/KafkaResource.java
```diff
@@ -74,6 +77,12 @@ public void delete(Kafka resource) {
 
         kafkaClient().inNamespace(namespaceName).withName(
             resource.getMetadata().getName()).withPropagationPolicy(DeletionPropagation.FOREGROUND).delete();
+
+        // additional deletion of pvcs with specification deleteClam set to false which were not deleted prior this method
```
michalxo, line 81: deleteClam

systemtest/src/main/java/io/strimzi/systemtest/resources/crd/KafkaResource.java
```diff
@@ -74,6 +77,12 @@ public void delete(Kafka resource) {
 
         kafkaClient().inNamespace(namespaceName).withName(
             resource.getMetadata().getName()).withPropagationPolicy(DeletionPropagation.FOREGROUND).delete();
+
+        // additional deletion of pvcs with specification deleteClaim set to false which were not deleted prior this method
+        for (PersistentVolumeClaim pvc : kubeClient().listPersistentVolumeClaims(namespaceName, clusterName)) {
+            kubeClient().deletePersistentVolumeClaim(namespaceName, pvc.getMetadata().getName());
+            PersistentVolumeClaimUtils.waitForPersistentVolumeClaimDeletion(namespaceName, pvc.getMetadata().getName());
+        }
```
scholzj, line 85: Why not do it in two cycles?
1) Request deletion of all PVCs in one loop
2) Wait for deletion in the second loop

That way, if PVC deletion takes 10 seconds and you delete 10 PVCs, you will nto wait 100 seconds but only ~10 seconds.

scholzj (commented):
I left on nit. But I wonder why this is needed. AFAIU the tests now run in a namespace per test-suite. And the namespace should be delete after the test suite is complete. That should delete the PVCs as well. So why is this needed in addition to that?

systemtest/src/main/java/io/strimzi/systemtest/resources/crd/KafkaResource.java
see-quick: > I left on nit. But I wonder why this is needed. AFAIU the tests now run in a namespace per test-suite. And the namespace should be delete after the test suite is complete. That should delete the PVCs as well. So why is this needed in addition to that?

Because, there are edge cases, where for instance you install Strimzi by OLM. When you install Strimzi by OLM and you want to have a cluster-wide setting, it has to be installed in the specific namespace where global `OperatorGroup` is installed. Such namespace we do not delete because it would also delete global `OperatorGroup` and thus we wouldn't be able to deploy Operator.

scholzj:
> Because, there are edge cases, where for instance you install Strimzi by OLM. When you install Strimzi by OLM and you want to have a cluster-wide setting, it has to be installed in the specific namespace where global OperatorGroup is installed. Such namespace we do not delete because it would also delete global OperatorGroup and thus we wouldn't be able to deploy Operator.

@see-quick But I would not expect you to deploy any Kafka clusters into the OLM namespace with cluster-wide operators, or? So there should not be any PVCs in there.

see-quick:
> @see-quick But I would not expect you to deploy any Kafka clusters into the OLM namespace with cluster-wide operators, or? So there should not be any PVCs in there.

We could possibly change that but currently, we are deploying resources inside that namespace when we execute almost all @IsolatedSuites (which does not create its own namespace).

systemtest/src/main/java/io/strimzi/systemtest/resources/crd/KafkaResource.java
jankalinic: re two cycles: I haven't bumped into a situation where deletion takes this much time. If there is a problem with deletion it's always caught in waitingForDeletion. Deletion is almost instantly completed in every situation. So in my opinion it should not be necessary.

scholzj: If it is instant, you do not need to wait for it ... if you need to wait for it, you should split it to save time ;-)

Frawless (approved):

systemtest/src/main/java/io/strimzi/systemtest/resources/crd/KafkaResource.java
jankalinic: You are right, I'm adding proposed solution for time save. Thank you for the suggestion.

scholzj (approved):

PR #6751, opened 2022-04-29
Unify refactored reconciler classes

scholzj:
### Type of change

- Refactoring

### Description

The refactoring of the `KafkaAssemblyOperator` class was done in multiple different steps and as a result, each reconciler looks a little bit different. This PR tries to unify them little bit more. It focuses on two things:
* Remove unnecessary prefixes in method names. E.g. in the `CruiseControlReconciler`, it is fairly obvious that we are reconciling Cruise Control resources. So we do not need to name every method as `cruiseControl...` (for example `cruiseControlDeployment`). We can leave out the `cruiseControl` part since the class makes it obvious what it reconciles. This was done out of the box in ZooKeeper and Kafka recocnilers. But not in the smaller reconcilers done at the start of the refactoring.
* Unifies the access modifiers. The top-level recocniler methods are often overwritten in tests when mocing the reconcilers and are set to `protected`. Intrnal methods are set as `private` unless required differently for some specific reasons (usually tests). Again, this was done later out of the box and is change only in the newer classes.

This really jsut unifies the code. There is no new functionality.

### Checklist


ppatierno (approved):
Thanks Jakub. These changes make more sense!

PR #6752, opened 2022-04-30
Move CA reconciliation from `KafkaAssemblyOperator` into a separate class

scholzj:
### Type of change

- Refactoring

### Description

This PR continues the refactoring of the `KafkaAssemblyOperator` and moves the reconciliation of the Cluster and Clients CAs into a separate calls called `CaReconciler`. It handles the reconciliation of both CAs as well as the Cluster Operator certificates. This should complete the refactoring of the `KafkaAssemblyOperator`.

### Checklist

- [x] Write tests

### DIFF
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/CaReconciler.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/CaReconciler.java
@@ -1,0 +1,515 @@
+/*
+ * Copyright Strimzi authors.
+ * License: Apache License 2.0 (see the file LICENSE or http://apache.org/licenses/LICENSE-2.0.html).
+ */
+package io.strimzi.operator.cluster.operator.assembly;
+
+import io.fabric8.kubernetes.api.model.OwnerReference;
+import io.fabric8.kubernetes.api.model.OwnerReferenceBuilder;
+import io.fabric8.kubernetes.api.model.Pod;
+import io.fabric8.kubernetes.api.model.Secret;
+import io.strimzi.api.kafka.model.CertificateAuthority;
+import io.strimzi.api.kafka.model.CruiseControlResources;
+import io.strimzi.api.kafka.model.Kafka;
+import io.strimzi.api.kafka.model.KafkaExporterResources;
+import io.strimzi.api.kafka.model.KafkaResources;
+import io.strimzi.certs.CertManager;
+import io.strimzi.operator.cluster.ClusterOperator;
+import io.strimzi.operator.cluster.ClusterOperatorConfig;
+import io.strimzi.operator.cluster.FeatureGates;
+import io.strimzi.operator.cluster.model.AbstractModel;
+import io.strimzi.operator.cluster.model.Ca;
+import io.strimzi.operator.cluster.model.ClientsCa;
+import io.strimzi.operator.cluster.model.ClusterCa;
+import io.strimzi.operator.cluster.model.InvalidResourceException;
+import io.strimzi.operator.cluster.model.KafkaCluster;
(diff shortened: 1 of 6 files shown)


cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/CaReconciler.java
```diff
@@ -0,0 +1,515 @@
...
+        if (kafkaCr.getSpec().getKafka() != null
+                && kafkaCr.getSpec().getKafka().getTemplate() != null
+                && kafkaCr.getSpec().getKafka().getTemplate().getClusterCaCert() != null
+                && kafkaCr.getSpec().getKafka().getTemplate().getClusterCaCert().getMetadata() != null
+                && kafkaCr.getSpec().getKafka().getTemplate().getClusterCaCert().getMetadata().getLabels() != null) {
+            return kafkaCr.getSpec().getKafka().getTemplate().getClusterCaCert().getMetadata().getLabels();
+        } else {
+            return Map.of();
+        }
+    }
```
ppatierno, line 172: Maybe this method could be something like this to avoid all this explicit `null` checks?

```java
return Optional.ofNullable(kafkaCr.getSpec().getKafka())
		.map(k -> k.getTemplate())
		.map(t -> t.getClusterCaCert()
		.map(r -> r.getMetadata())
		.map(m -> r.getLabels())
		.orElse(Map.of());
```

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/CaReconciler.java
```diff
@@ -0,0 +1,515 @@
...
+        if (kafkaCr.getSpec().getKafka() != null
+                && kafkaCr.getSpec().getKafka().getTemplate() != null
+                && kafkaCr.getSpec().getKafka().getTemplate().getClusterCaCert() != null
+                && kafkaCr.getSpec().getKafka().getTemplate().getClusterCaCert().getMetadata() != null
+                && kafkaCr.getSpec().getKafka().getTemplate().getClusterCaCert().getMetadata().getAnnotations() != null) {
+            return kafkaCr.getSpec().getKafka().getTemplate().getClusterCaCert().getMetadata().getAnnotations();
+        } else {
+            return Map.of();
+        }
+    }
```
ppatierno, line 191: Ditto as above, but for annotations?

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/CaReconciler.java
scholzj: TBH, I think the `if` above is way more readable and easier to understand what it actually does.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/CaReconciler.java
scholzj: TBH, I think the `if` above is way more readable and easier to understand what it actually does.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/CaReconciler.java
ppatierno: I can see your point, that's the good old way but too verbose. Btw no need to change. It was just a suggestion.

ppatierno (approved):

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/CaReconciler.java
scholzj: Thanks.

PR #6764, opened 2022-05-04
[systemtest] Use KafkaConnect build for building images with FileSink plugin

im-konge:

### Type of change

- Enhancement 

### Description

In Kafka 3.2.0 and 3.1.1, the `FileStreamSourceConnector` and `FileStreamSinkConnector` will be removed from the default `CLASSPATH` and `plugin.path`. That means that we will not be able to simply create "file sink connectors" from the main Kafka image, as we are doing it now in our tests.

This PR adds usage of KafkaConnect build feature for building the image with FileSink plugin, which will ensure that we are able to use the file sink connectors in our tests, after it will be removed in future Kafka versions.

### Checklist

### DIFF
--- a/systemtest/src/main/java/io/strimzi/systemtest/Environment.java
+++ b/systemtest/src/main/java/io/strimzi/systemtest/Environment.java
@@ -6,9 +6,11 @@ package io.strimzi.systemtest;
 
 import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.ObjectMapper;
+import io.fabric8.kubernetes.api.model.Service;
 import io.strimzi.systemtest.enums.ClusterOperatorInstallType;
 import io.strimzi.systemtest.utils.TestKafkaVersion;
 import io.strimzi.test.TestUtils;
+import io.strimzi.test.k8s.KubeClusterResource;
 import io.strimzi.test.k8s.cluster.OpenShift;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
@@ -21,6 +23,8 @@ import java.util.Locale;
 import java.util.Map;
 import java.util.Objects;
 import java.util.function.Function;
+
+import static io.strimzi.test.k8s.KubeClusterResource.kubeClient;
 
 /**
  * Class which holds environment variables for system tests.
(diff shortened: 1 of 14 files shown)


systemtest/src/main/java/io/strimzi/systemtest/templates/crd/KafkaConnectTemplates.java
```diff
@@ -115,6 +121,42 @@ private static KafkaConnectBuilder deployKafkaConnectWithNetworkPolicy(Extension
...
+
+    public static KafkaConnectBuilder kafkaConnectWithFileSinkPlugin(ExtensionContext extensionContext, String clusterName, int replicas, boolean allowNP) {
+        return kafkaConnectWithFileSinkPlugin(extensionContext, clusterName, clusterName, replicas, allowNP);
+    }
+
+    public static KafkaConnectBuilder kafkaConnectWithFileSinkPlugin(ExtensionContext extensionContext, String name, String clusterName, int replicas) {
+        return kafkaConnectWithFileSinkPlugin(extensionContext, name, clusterName, replicas, true);
+    }
+
+    public static KafkaConnectBuilder kafkaConnectWithFileSinkPlugin(ExtensionContext extensionContext, String name, String clusterName, int replicas, boolean allowNP) {
```
scholzj, line 136: Javadoc would not hurt. For example what is `allowNP`?

systemtest/src/main/java/io/strimzi/systemtest/Environment.java
```diff
@@ -141,6 +145,8 @@ public class Environment {
      */
     public static final String STRIMZI_POD_SET_RECONCILIATION_ONLY_ENV = "STRIMZI_POD_SET_RECONCILIATION_ONLY";
 
+    public static final String ST_FILE_SINK_PLUGIN_URL_ENV = "ST_FILE_SINK_PLUGIN_URL";
```
scholzj, line 148: I'm fine if you wanna keep the current naming both here as well as for the methods. But strictly speaking, you are adding both the File Sink and File Source connectors. So maybe the names can take that int account?

scholzj (approved):
I left two nits. But LGTM assuming the tests pass.

systemtest/src/main/java/io/strimzi/systemtest/Environment.java
im-konge: Maybe I can use just `ST_FILE_PLUGIN_URL_ENV` as the package is named `connect-file` and also the `connect-file` plugin contains all addons related to file operations? In that case it will be also shorter than adding the `_AND_SOURCE` to the name.

Frawless (approved):

Issue #6265, opened 2022-01-26
Cruise Control CPU + Network I/O capacity detection


kyguy:
**Is your feature request related to a problem? Please describe.**

We need a way to detect the following resources of our nodes:
- CPU 
- Network  I/0

and configure those resources in our Cruise Control capacity configuration.

Otherwise, when balancing partitions across nodes with heterogeneous resource capacities, Cruise Control could under or overestimated broker resources leading to balances that leave our cluster in a worse state than it was before the balance! 

This problem affects the following goals:

- CpuUsageDistributionGoal
- CpuCapacityDistributionGoal
- NetworkInboundUsageDistributionGoal
- NetworkInboundCapacityGoal
- NetworkOutboundUsageDistributionGoal
- NetworkOutboundCapacityGoal

The current workaround for CPU goals is setting: 
```
resource.requests.cpu == resource.limits.cpu
```
What this does is:

- Reserves the same CPU resources upfront, essentially setting the same CPU capacity for every broker.

However, it also:

- Sidelines those CPU resources only to ever be used by the broker, even when those resources are not being used.
- Throttles the broker when it  tries to use more CPU resources than allotted, even when there are spare cycles available on the node.

**Describe the solution you'd like**
Ideally, we patch the upstream Cruise Control metric reporter to detect the resource capacity of the operating environment similar to how it detects CPU utilization e.g. how many CPU cores are available, how much network bandwidth is available  etc. Then adjusting the Cruise Control capacity configuration generation in Strimzi to account for the new upstream changes.

**Describe alternatives you've considered**
Writing our own BrokerCapacityConfigResolver [2] that determines the resource capacity of nodes that pods are running on. However, this would require the Cruise Control pod to have special privileges to extract resource information from the nodes which broker pods are running on.

**Additional context**
[1] https://github.com/strimzi/strimzi-kafka-operator/issues/5951
[2] https://github.com/linkedin/cruise-control/wiki/Pluggable-Components#broker-capacity-config-resolver

fvaleri:
Hi @kyguy, thanks for raising this. I would say "reconcile resource capacity" instead of detect, as this is something that should be done periodically.

PR #6769, opened 2022-05-04
Add network capacity overrides for CC capacity config

kyguy:
### Type of change

- Enhancement / new feature

### Description

For accurate rebalances between brokers running on nodes with heterogeneous network resources, Cruise Control must know the network capacity limit of individual brokers. This PR allows users to specify capacity limit overrides for lists of individual Kafka brokers in the `overrides` property in `Kafka.spec.cruiseControl.brokerCapacity`.

This PR addresses the network capacity issues of https://github.com/strimzi/strimzi-kafka-operator/issues/6265.  The CPU overrides will be handled in a separate PR.

```
apiVersion: {KafkaApiVersion}
kind: Kafka
metadata:
  name: my-cluster
spec:
  # ...
  cruiseControl:
    # ...
    brokerCapacity:
      inboundNetwork: 10000KiB/s
      outboundNetwork: 10000KiB/s
      overrides:
      - brokers: [0]
        inboundNetwork: 20000KiB/s
        outboundNetwork: 20000KiB/s
      - brokers: [1, 2]
        inboundNetwork: 30000KiB/s
        outboundNetwork: 30000KiB/s
```

### Checklist

- [x] Write tests
- [x] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/model/cruisecontrol/Broker.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/model/cruisecontrol/Broker.java
@@ -1,0 +1,77 @@
+/*
+ * Copyright Strimzi authors.
+ * License: Apache License 2.0 (see the file LICENSE or http://apache.org/licenses/LICENSE-2.0.html).
+ */
+package io.strimzi.operator.cluster.model.cruisecontrol;
+
+public class Broker {
+    // CC allows specifying a generic "default" broker entry in the capacity configuration to apply to all brokers without a specific broker entry.
+    // CC designates the id of this default broker entry as "-1".
+    public static final int DEFAULT_BROKER_ID = -1;
+    public static final String DEFAULT_BROKER_DOC = "This is the default capacity. Capacity unit used for disk is in MB, cpu is in percentage, network throughput is in KB.";
+
+    public static final String DEFAULT_CPU_UTILIZATION_CAPACITY = "100";  // as a percentage (0-100)
+    public static final String DEFAULT_DISK_CAPACITY_IN_MIB = "100000";  // in MiB
+    public static final String DEFAULT_INBOUND_NETWORK_CAPACITY_IN_KIB_PER_SECOND = "10000";  // in KiB/s
+    public static final String DEFAULT_OUTBOUND_NETWORK_CAPACITY_IN_KIB_PER_SECOND = "10000";  // in KiB/s
+
+    private Integer id;
+    private String cpu;
+    private String disk;
+    private String inboundNetworkKiBPerSecond;
+    private String outboundNetworkKiBPerSecond;
+    private String doc;
+
+    public Broker(int brokerId, String cpu, String disk, String inboundNetworkKiBPerSecond, String outboundNetworkKiBPerSecond) {
+        this.id = brokerId;
+        this.cpu = cpu;
+        this.disk = disk;
+        this.inboundNetworkKiBPerSecond = inboundNetworkKiBPerSecond;
+        this.outboundNetworkKiBPerSecond = outboundNetworkKiBPerSecond;
+        this.doc = brokerId == -1 ? DEFAULT_BROKER_DOC : "Capacity for Broker " + brokerId;
+    }
+
+    public Integer getId() {
+        return id;
+    }
+
+    public String getCpu() {
+        return cpu;
+    }
+
+    public String getDisk() {
+        return disk;
+    }
+
+    public String getInboundNetworkKiBPerSecond() {
+        return inboundNetworkKiBPerSecond;
+    }
+
+    public String getOutboundNetworkKiBPerSecond() {
+        return outboundNetworkKiBPerSecond;
+    }
+
+    public String getDoc() {
+        return doc;
+    }
+
+    public void setId(Integer id) {
+        this.id = id;
+    }
+
+    public void setCpu(String cpu) {
+        this.cpu = cpu;
+    }
+
+    public void setDisk(String disk) {
+        this.disk = disk;
+    }
+
+    public void setInboundNetworkKiBPerSecond(String inboundNetworkKiBPerSecond) {
+        this.inboundNetworkKiBPerSecond = inboundNetworkKiBPerSecond;
+    }
+
+    public void setOutboundNetworkKiBPerSecond(String outboundNetworkKiBPerSecond) {
+        this.outboundNetworkKiBPerSecond = outboundNetworkKiBPerSecond;
+    }
+}
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/model/cruisecontrol/BrokerCapacity.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/model/cruisecontrol/BrokerCapacity.java
@@ -18,16 +18,16 @@ public class BrokerCapacity {
     private int id;
     private String cpu;
     private String disk;
-    private String inboundNetworkKiBPerSecond;
-    private String outboundNetworkKiBPerSecond;
+    private String inboundNetwork;
+    private String outboundNetwork;
     private final String doc;
 
-    public BrokerCapacity(int brokerId, String cpu, String disk, String inboundNetworkKiBPerSecond, String outboundNetworkKiBPerSecond) {
+    public BrokerCapacity(int brokerId, String cpu, String disk, String inboundNetwork, String outboundNetwork) {
         this.id = brokerId;
         this.cpu = cpu;
         this.disk = disk;
-        this.inboundNetworkKiBPerSecond = inboundNetworkKiBPerSecond;
-        this.outboundNetworkKiBPerSecond = outboundNetworkKiBPerSecond;
+        this.inboundNetwork = inboundNetwork;
+        this.outboundNetwork = outboundNetwork;
         this.doc = brokerId == -1 ? DEFAULT_BROKER_DOC : "Capacity for Broker " + brokerId;
     }
 
@@ -43,12 +43,12 @@ public class BrokerCapacity {
         return disk;
     }
 
-    public String getInboundNetworkKiBPerSecond() {
-        return inboundNetworkKiBPerSecond;
+    public String getInboundNetwork() {
+        return inboundNetwork;
     }
 
-    public String getOutboundNetworkKiBPerSecond() {
-        return outboundNetworkKiBPerSecond;
+    public String getOutboundNetwork() {
+        return outboundNetwork;
     }
 
     public String getDoc() {
@@ -67,11 +67,11 @@ public class BrokerCapacity {
         this.disk = disk;
     }
 
-    public void setInboundNetworkKiBPerSecond(String inboundNetworkKiBPerSecond) {
-        this.inboundNetworkKiBPerSecond = inboundNetworkKiBPerSecond;
+    public void setInboundNetwork(String inboundNetwork) {
+        this.inboundNetwork = inboundNetwork;
     }
 
-    public void setOutboundNetworkKiBPerSecond(String outboundNetworkKiBPerSecond) {
-        this.outboundNetworkKiBPerSecond = outboundNetworkKiBPerSecond;
+    public void setOutboundNetwork(String outboundNetwork) {
+        this.outboundNetwork = outboundNetwork;
     }
 }
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/model/cruisecontrol/BrokerCapacity.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/model/cruisecontrol/BrokerCapacity.java
@@ -17,12 +17,12 @@ public class BrokerCapacity {
 
     private int id;
     private String cpu;
-    private Object disk;
+    private DiskCapacity disk;
     private String inboundNetwork;
     private String outboundNetwork;
     private final String doc;
 
-    public BrokerCapacity(int brokerId, String cpu, Object disk, String inboundNetwork, String outboundNetwork) {
+    public BrokerCapacity(int brokerId, String cpu, DiskCapacity disk, String inboundNetwork, String outboundNetwork) {
         this.id = brokerId;
         this.cpu = cpu;
         this.disk = disk;
@@ -39,7 +39,7 @@ public class BrokerCapacity {
         return cpu;
     }
 
-    public Object getDisk() {
+    public DiskCapacity getDisk() {
         return disk;
     }
 
@@ -63,7 +63,7 @@ public class BrokerCapacity {
         this.cpu = cpu;
     }
 
-    public void setDisk(Object disk) {
+    public void setDisk(DiskCapacity disk) {
         this.disk = disk;
     }
 
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/model/cruisecontrol/BrokerCapacity.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/model/cruisecontrol/BrokerCapacity.java
@@ -4,25 +4,25 @@
  */
 package io.strimzi.operator.cluster.model.cruisecontrol;
 
-public class Broker {
+public class BrokerCapacity {
     // CC allows specifying a generic "default" broker entry in the capacity configuration to apply to all brokers without a specific broker entry.
     // CC designates the id of this default broker entry as "-1".
     public static final int DEFAULT_BROKER_ID = -1;
-    public static final String DEFAULT_BROKER_DOC = "This is the default capacity. Capacity unit used for disk is in MB, cpu is in percentage, network throughput is in KB.";
+    public static final String DEFAULT_BROKER_DOC = "This is the default capacity. Capacity unit used for disk is in MiB, cpu is in percentage, network throughput is in KiB.";
 
     public static final String DEFAULT_CPU_UTILIZATION_CAPACITY = "100";  // as a percentage (0-100)
-    public static final String DEFAULT_DISK_CAPACITY_IN_MIB = "100000";  // in MiB
-    public static final String DEFAULT_INBOUND_NETWORK_CAPACITY_IN_KIB_PER_SECOND = "10000";  // in KiB/s
-    public static final String DEFAULT_OUTBOUND_NETWORK_CAPACITY_IN_KIB_PER_SECOND = "10000";  // in KiB/s
+    public static final String DEFAULT_DISK_CAPACITY_IN_MIB = "100000";
+    public static final String DEFAULT_INBOUND_NETWORK_CAPACITY_IN_KIB_PER_SECOND = "10000";
+    public static final String DEFAULT_OUTBOUND_NETWORK_CAPACITY_IN_KIB_PER_SECOND = "10000";
 
-    private Integer id;
+    private int id;
     private String cpu;
     private String disk;
     private String inboundNetworkKiBPerSecond;
     private String outboundNetworkKiBPerSecond;
-    private String doc;
+    private final String doc;
 
-    public Broker(int brokerId, String cpu, String disk, String inboundNetworkKiBPerSecond, String outboundNetworkKiBPerSecond) {
+    public BrokerCapacity(int brokerId, String cpu, String disk, String inboundNetworkKiBPerSecond, String outboundNetworkKiBPerSecond) {
         this.id = brokerId;
         this.cpu = cpu;
         this.disk = disk;
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/model/cruisecontrol/BrokerCapacity.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/model/cruisecontrol/BrokerCapacity.java
@@ -17,12 +17,12 @@ public class BrokerCapacity {
 
     private int id;
     private String cpu;
-    private String disk;
+    private Object disk;
     private String inboundNetwork;
     private String outboundNetwork;
     private final String doc;
 
-    public BrokerCapacity(int brokerId, String cpu, String disk, String inboundNetwork, String outboundNetwork) {
+    public BrokerCapacity(int brokerId, String cpu, Object disk, String inboundNetwork, String outboundNetwork) {
         this.id = brokerId;
         this.cpu = cpu;
         this.disk = disk;
@@ -39,7 +39,7 @@ public class BrokerCapacity {
         return cpu;
     }
 
-    public String getDisk() {
+    public Object getDisk() {
         return disk;
     }
 
@@ -63,7 +63,7 @@ public class BrokerCapacity {
         this.cpu = cpu;
     }
 
-    public void setDisk(String disk) {
+    public void setDisk(Object disk) {
         this.disk = disk;
     }
 
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/model/cruisecontrol/Capacity.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/model/cruisecontrol/Capacity.java
@@ -118,6 +118,15 @@ public class Capacity {
 
     private final TreeMap<Integer, BrokerCapacity> capacityEntries;
 
+    public static final String CAPACITIES_KEY = "brokerCapacities";
+    private static final String BROKER_ID_KEY = "brokerId";
+    public static final String CAPACITY_KEY = "capacity";
+    public static final String DISK_KEY = "DISK";
+    private static final String CPU_KEY = "CPU";
+    private static final String INBOUND_NETWORK_KEY = "NW_IN";
+    private static final String OUTBOUND_NETWORK_KEY = "NW_OUT";
+    private static final String DOC_KEY = "doc";
+
     private final int replicas;
     private final Storage storage;
 
@@ -135,7 +144,7 @@ public class Capacity {
         return BrokerCapacity.DEFAULT_CPU_UTILIZATION_CAPACITY;
     }
 
-    public static Object processDisk(Storage storage, int brokerId) {
+    public static DiskCapacity processDisk(Storage storage, int brokerId) {
         if (storage instanceof JbodStorage) {
             return generateJbodDiskCapacity(storage, brokerId);
         } else {
@@ -168,10 +177,10 @@ public class Capacity {
      *
      * @param storage Storage configuration for Kafka cluster
      * @param brokerId Id of the broker
-     * @return Disk capacity configuration value as a JsonObject for broker brokerId
+     * @return Disk capacity configuration value for broker brokerId
      */
-    private static JsonObject generateJbodDiskCapacity(Storage storage, int brokerId) {
-        JsonObject disks = new JsonObject();
+    private static DiskCapacity generateJbodDiskCapacity(Storage storage, int brokerId) {
+        DiskCapacity disks = new DiskCapacity();
         String size = "";
 
         for (SingleVolumeStorage volume : ((JbodStorage) storage).getVolumes()) {
@@ -183,7 +192,7 @@ public class Capacity {
             } else if (volume instanceof EphemeralStorage) {
                 size = ((EphemeralStorage) volume).getSizeLimit();
             }
-            disks.put(path, String.valueOf(getSizeInMiB(size)));
+            disks.add(path, String.valueOf(getSizeInMiB(size)));
         }
         return disks;
     }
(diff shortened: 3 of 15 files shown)


cluster-operator/src/main/java/io/strimzi/operator/cluster/model/cruisecontrol/Capacity.java
```diff
@@ -14,62 +15,151 @@
...
+ cruiseControl:
+   brokerCapacity:
+    inboundNetwork: 10000KB/s
+    outboundNetwork: 10000KB/s
+    overrides:
+      - brokers: [0]
+      outboundNetwork: 40000KB/s
+      - brokers: [1, 2]
+      inboundNetwork: 60000KB/s
+      outboundNetwork: 20000KB/s
```
ppatierno, line 55: just a nit, fix indentation

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/cruisecontrol/Capacity.java
```diff
@@ -178,8 +228,120 @@ public static Double getSizeInMiB(String size) {
...
         String size = throughput.substring(0, throughput.indexOf("B"));
-        return StorageUtils.convertTo(size, "Ki");
+        return String.valueOf(StorageUtils.convertTo(size, "Ki"));
+    }
+
+    public static void main(String[] args) {
+        System.out.println(getThroughputInKiB("10000KiB/s"));
+        System.out.println(removeKibSuffix("10000KiB/s"));
+
+    }
```
ppatierno, line 240: what's this `main` here?

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/cruisecontrol/Capacity.java
```diff
@@ -178,8 +228,120 @@ public static Double getSizeInMiB(String size) {
...
+            // A capacity configuration for a cluster with a JBOD configuration
+            // requires a distinct broker capacity entry for every broker because the
+            // Kafka volume paths are not homogeneous across brokers and include
+            // the broker pod index in their names.
+            for (int id = 0; id < replicas; id++) {
+                disk = processDisk(storage, id);
+                Broker broker = new Broker(id, cpu, disk, inboundNetwork, outboundNetwork);
+                capacityEntries.put(id, broker);
+            }
+        }
```
ppatierno, line 266: Maybe I am missing something here but ... it seems that if JBOD is used, you are already filling the entries map will ALL the brokers in the cluster. So it seems that even if you don't have any override, you are anyway filling this entries list with all brokers. Is that correct?

Another point is that on line 285, you have the `processDisk` again and inside that method you check again if storage is JBOD and apply the corresponding logic. Isn't there any duplication here?

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/cruisecontrol/Capacity.java
```diff
@@ -14,62 +15,151 @@
...
 import io.strimzi.operator.cluster.model.VolumeUtils;
+import io.strimzi.operator.common.ReconciliationLogger;
 import io.vertx.core.json.JsonArray;
 import io.vertx.core.json.JsonObject;
 
+import java.util.HashSet;
 import java.util.List;
+import java.util.TreeMap;
 
+/*
```
scholzj, line 26: Is this intentionally not marked as Javadoc?

PaulRMellor (approved):
Great. kyle. :+1: 
I left a couple of suggestions.

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/cruisecontrol/Capacity.java
kyguy: > Maybe I am missing something here but ... it seems that if JBOD is used, you are already filling the entries map will ALL the brokers in the cluster. So it seems that even if you don't have any override, you are anyway filling this entries list with all brokers. Is that correct?

If JBOD is used then we must have an entry for every broker. Otherwise, we only need a broker entry per override. 

> Another point is that on line 285, you have the processDisk again and inside that method you check again if storage is JBOD and apply the corresponding logic. Isn't there any duplication here?

Yes, you are right, this is a duplicate call that is not needed, just removed it now!

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/cruisecontrol/Capacity.java
kyguy: Whoops I meant to remove that,  just removed it now

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/cruisecontrol/Capacity.java
kyguy: That was an oversight, now fixed

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/cruisecontrol/Capacity.java
ppatierno: Wait ... I thought that we could remove the big block of code here not the call on line 285.
So now I have two doubts:

Why, if JBOD is used, we MUST HAVE an entry for every broker? Can you elaborate or giving me a reference with reasons.

If JBOD is not used, this block doesn't run, but now you removed the `processDisk` on line 285, so disk info won't be set anymore?

kyguy: > Why, if JBOD is used, we MUST HAVE an entry for every broker? Can you elaborate or giving me a reference with reasons.

A capacity configuration for a cluster with a JBOD configuration requires a distinct broker capacity entry for every broker because the Kafka volume paths are not homogeneous across brokers and include the broker pod index in their names like `/var/lib/kafka<volume_number>/kafka-log<broker_id>`. For Cruise Control to find the JBOD volumes we must list them by path in the `DISK` sections of the broker entries like this:
```
     {
        "brokerId": "0",
        "capacity": {
          "DISK": {
              "/var/lib/kafka0/kafka-log0": "100000", 
              "/var/lib/kafka1/kafka-log0": "200000"
            },
            ...
        },
        "doc": "Capacity for Broker 0"
      }
        "brokerId": "1",
        "capacity": {
          "DISK": {
              "/var/lib/kafka0/kafka-log1": "100000",
              "/var/lib/kafka1/kafka-log1": "200000"
            },
            ...
        },
        "doc": "Capacity for Broker 1"
      },
 ```

> If JBOD is not used, this block doesn't run, but now you removed the processDisk on line 285, so disk info won't be set anymore?

If JBOD is not used then all brokers will have the same disk capacity configuration as set at the beginning of the method on line 239.  The overriden brokers will have their own broker entries with this disk value and the other brokers will rely on the same disk value as they use the default broker entry configuration

File: cluster-operator/src/main/java/io/strimzi/operator/cluster/model/cruisecontrol/Capacity.java [hunk 5]
```diff
@@ -178,8 +229,109 @@ public static Double getSizeInMiB(String size) {
...
      * @return The equivalent number of kibibytes.
      */
-    public static Double getThroughputInKiB(String throughput) {
+    public static String getThroughputInKiB(String throughput) {
         String size = throughput.substring(0, throughput.indexOf("B"));
-        return StorageUtils.convertTo(size, "Ki");
+        return String.valueOf(StorageUtils.convertTo(size, "Ki"));
+    }
+
+    private void processCapacityEntries(BrokerCapacity bc, Storage s) {
```
tomncooper, line 237: This is a nit, but I really don't like short acronyms as variable names. It makes it hard to figure out what they are later on.

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/cruisecontrol/Capacity.java
```diff
@@ -178,8 +229,109 @@ public static Double getSizeInMiB(String size) {
...
+                        List<Integer> ids = override.getBrokers();
+                        inboundNetwork = processInboundNetwork(bc, override);
+                        outboundNetwork = processOutboundNetwork(bc, override);
+                        for (Integer id : ids) {
+                            if (id == Broker.DEFAULT_BROKER_ID) {
+                                LOGGER.warnOp("Ignoring broker capacity override with illegal broker id -1.");
+                            } else {
+                                if (capacityEntries.containsKey(id)) {
+                                    if (overrideIds.contains(id)) {
+                                        LOGGER.warnOp("Duplicate broker id %d found in overrides, using last occurrence.", id);
```
tomncooper, line 278: Shouldn't we skip if there is a duplicate, save doing the processing and say "first occurrence will be used"?

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/cruisecontrol/Broker.java
```diff
@@ -0,0 +1,77 @@
+/*
```
tomncooper, line 1: This class should really be called `BrokerCapacity`. I know you have a class called that in the API module but we should call a spade a spade.

cluster-operator/src/test/java/io/strimzi/operator/cluster/model/CruiseControlTest.java
```diff
@@ -244,6 +246,53 @@ public void testBrokerCapacities() {
...
         capacity = new Capacity(resource.getSpec(), jbodStorage);
         assertThat(getCapacityConfigurationFromEnvVar(resource, ENV_VAR_CRUISE_CONTROL_CAPACITY_CONFIGURATION), is(capacity.generateCapacityConfig()));
+
+        // Test capacity overrides
+        String inboundNetwork = "50000KB/s";
+        String inboundNetworkOverride0 = "25000KB/s";
+        String inboundNetworkOverride1 = "10000KiB/s";
+        String outboundNetworkOverride1 = "15000KB/s";
+
+        Integer broker0 = 0;
```
mimaison, line 256: What about using the values directly? Do you think the variables help readability?

systemtest/src/test/java/io/strimzi/systemtest/cruisecontrol/CruiseControlConfigurationST.java
```diff
@@ -63,38 +65,87 @@ public class CruiseControlConfigurationST extends AbstractST {
...
 
-        LOGGER.info("We got only one configuration of broker-capacities");
-        assertThat(cruiseControlCapacityFileContent.getJsonArray("brokerCapacities").size(), is(1));
+        LOGGER.info("We have 4 entries in broker-capacities");
+        assertThat(brokerCapacities.size(), is(4));
 
-        LOGGER.info("Verifying cruise control configuration.");
+        JsonObject defaultBroker = brokerCapacities.getJsonObject(0);
+        assertThat(defaultBroker.getString("brokerId"), is("-1"));
+        assertThat(defaultBroker.getString("doc"), not(nullValue()));
```
mimaison, line 127: Can we use `notNullValue()`?

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/cruisecontrol/Capacity.java
```diff
@@ -103,72 +194,32 @@ private JsonObject generateJbodDiskCapacity(Storage storage, int idx) {
...
-        }
-        JsonObject config = new JsonObject();
-        config.put("brokerCapacities", brokerList);
-
-        return config.encodePrettily();
-    }
-
     /*
      * Parse a K8S-style representation of a disk size, such as {@code 100Gi},
      * into the equivalent number of mebibytes represented as a Double.
```
mimaison, line 213: Should it be `represented as a String` now?
Same below

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/cruisecontrol/Capacity.java
```diff
@@ -178,8 +229,109 @@ public static Double getSizeInMiB(String size) {
...
+            for (int id = 0; id < replicas; id++) {
+                disk = processDisk(storage, id);
+                Broker broker = new Broker(id, cpu, disk, inboundNetwork, outboundNetwork);
+                capacityEntries.put(id, broker);
+            }
+        }
+
+        if (bc != null) {
+            // For checking for duplicate brokerIds
+            HashSet<Integer> overrideIds = new HashSet<>();
```
mimaison, line 261: We can use `Set` on the left

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/cruisecontrol/Capacity.java
```diff
@@ -178,8 +229,109 @@ public static Double getSizeInMiB(String size) {
...
+                    for (BrokerCapacityOverride override : overrides) {
+                        List<Integer> ids = override.getBrokers();
+                        inboundNetwork = processInboundNetwork(bc, override);
+                        outboundNetwork = processOutboundNetwork(bc, override);
+                        for (Integer id : ids) {
+                            if (id == Broker.DEFAULT_BROKER_ID) {
+                                LOGGER.warnOp("Ignoring broker capacity override with illegal broker id -1.");
+                            } else {
+                                if (capacityEntries.containsKey(id)) {
+                                    if (overrideIds.contains(id)) {
```
mimaison, line 277: Since `overrideIds` is a `Set` we could call `add()` unconditionally and use the return value to print or not the warning.

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/cruisecontrol/Capacity.java
```diff
@@ -178,8 +229,109 @@ public static Double getSizeInMiB(String size) {
...
+            if (overrides != null) {
+                if (overrides.isEmpty()) {
+                    LOGGER.warnOp("Ignoring empty overrides list");
+                } else {
+                    for (BrokerCapacityOverride override : overrides) {
+                        List<Integer> ids = override.getBrokers();
+                        inboundNetwork = processInboundNetwork(bc, override);
+                        outboundNetwork = processOutboundNetwork(bc, override);
+                        for (Integer id : ids) {
+                            if (id == Broker.DEFAULT_BROKER_ID) {
```
mimaison, line 273: This works because `Broker.DEFAULT_BROKER_ID` is an `int` and causes `id` to auto unbox. It's usually preferable to avoid relying on this behavior in equality checks.

mimaison (commented):
Thanks for the PR, I left a few suggestions

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/cruisecontrol/Capacity.java
```diff
@@ -14,62 +15,152 @@
...
-                .put("NW_OUT", Double.toString(outboundNetworkKiBPerSecond))
-            )
-            .put("doc", doc);
-        return brokerCapacity;
+    public static String processCpu() {
+        return Broker.DEFAULT_CPU_UTILIZATION_CAPACITY;
+    }
+
+    public static String processDisk(Storage storage, int brokerId) {
+        if (storage instanceof  JbodStorage) {
```
im-konge, line 139:
```suggestion
        if (storage instanceof JbodStorage) {
```

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/cruisecontrol/Capacity.java
```diff
@@ -14,62 +15,152 @@
...
+    }
+
+    public static String processInboundNetwork(BrokerCapacity bc, BrokerCapacityOverride override) {
+        if (override != null && override.getInboundNetwork() != null) {
+            return getThroughputInKiB(override.getInboundNetwork());
+        } else if (bc != null && bc.getInboundNetwork() != null) {
+            return getThroughputInKiB(bc.getInboundNetwork());
+        } else {
+            return Broker.DEFAULT_INBOUND_NETWORK_CAPACITY_IN_KIB_PER_SECOND;
+        }
```
im-konge, line 153: Do you need the `else` block here? It's just a nit, but I think you can do 
```suggestion
        if (override != null && override.getInboundNetwork() != null) {
            return getThroughputInKiB(override.getInboundNetwork());
        } else if (bc != null && bc.getInboundNetwork() != null) {
            return getThroughputInKiB(bc.getInboundNetwork());
        }
        return Broker.DEFAULT_INBOUND_NETWORK_CAPACITY_IN_KIB_PER_SECOND;
```

maybe same for the `processOutboundNetwork`

systemtest/src/test/java/io/strimzi/systemtest/cruisecontrol/CruiseControlConfigurationST.java
```diff
@@ -63,38 +65,87 @@ public class CruiseControlConfigurationST extends AbstractST {
 
     private final String namespace = testSuiteNamespaceManager.getMapOfAdditionalNamespaces().get(CruiseControlConfigurationST.class.getSimpleName()).stream().findFirst().get();
 
+    private static Double removeKibSuffix(String s) {
+        return Double.valueOf(s.substring(0, s.length() - 5));
+    }
+
```
im-konge, line 71: This method should be in `StUtils` or something like that FMPOV.

systemtest/src/test/java/io/strimzi/systemtest/cruisecontrol/CruiseControlConfigurationST.java
```diff
@@ -63,38 +65,87 @@ public class CruiseControlConfigurationST extends AbstractST {
...
+                            .withBrokers(overrideList1)
+                            .withInboundNetwork(inboundNetworkOverride1)
+                            .withOutboundNetwork(outboundNetworkOverride1)
+                        .endOverride()
+                    .endBrokerCapacity()
+                .endCruiseControl()
+            .endSpec()
+            .build();
+
+        resourceManager.createResource(extensionContext, kr);
```
im-konge, line 111:
```suggestion
        resourceManager.createResource(extensionContext, KafkaTemplates.kafkaWithCruiseControl(clusterName, 3, 3)
            .editOrNewSpec()
                .editCruiseControl()
                    .withNewBrokerCapacity()
                        .withInboundNetwork(inboundNetwork)
                        .addNewOverride()
                            .withBrokers(overrideList0)
                            .withInboundNetwork(inboundNetworkOverride0)
                        .endOverride()
                        .addNewOverride()
                            .withBrokers(overrideList1)
                            .withInboundNetwork(inboundNetworkOverride1)
                            .withOutboundNetwork(outboundNetworkOverride1)
                        .endOverride()
                    .endBrokerCapacity()
                .endCruiseControl()
            .endSpec()
            .build());
```
as you don't use the `kr` anywhere

systemtest/src/test/java/io/strimzi/systemtest/cruisecontrol/CruiseControlConfigurationST.java
```diff
@@ -63,38 +65,87 @@ public class CruiseControlConfigurationST extends AbstractST {
...
+        double cpu = 100;
+        String inboundNetwork = "10000KiB/s";
+        String outboundNetwork = "10000KiB/s";
+        String inboundNetworkOverride0 = "20000KiB/s";
+        String inboundNetworkOverride1 = "40000KiB/s";
+        String outboundNetworkOverride1 = "400000KiB/s";
+
+        Integer broker0 = 0;
+        Integer broker1 = 1;
+        Integer broker2 = 2;
```
im-konge, line 87: I guess we can go with primitive `int` here - or maybe pass it directly to the list

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/cruisecontrol/Broker.java
```diff
@@ -0,0 +1,77 @@
...
+ */
+package io.strimzi.operator.cluster.model.cruisecontrol;
+
+public class Broker {
+    // CC allows specifying a generic "default" broker entry in the capacity configuration to apply to all brokers without a specific broker entry.
+    // CC designates the id of this default broker entry as "-1".
+    public static final int DEFAULT_BROKER_ID = -1;
+    public static final String DEFAULT_BROKER_DOC = "This is the default capacity. Capacity unit used for disk is in MB, cpu is in percentage, network throughput is in KB.";
+
+    public static final String DEFAULT_CPU_UTILIZATION_CAPACITY = "100";  // as a percentage (0-100)
```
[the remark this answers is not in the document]
kyguy: I'll be doing more work with this in my next PR which will involve CPU capacity overrides, if it's ok with you I would like to keep this field type as a String so it is consistent with the other field types and I don't have to cast it later when make the brokerCapacity configuration string

cluster-operator/src/test/java/io/strimzi/operator/cluster/model/CruiseControlTest.java
kyguy: I really think it helps with the readability, especially where it is used with the assertions statements, I would like to keep these but if you feel strongly I will remove them!

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/cruisecontrol/Capacity.java
kyguy: I think so, if we do the String conversion here then we don't have to do it in the calling class

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/cruisecontrol/Capacity.java
kyguy: Good idea! Can you double check what I changed to make sure I got it right?

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/cruisecontrol/Capacity.java
kyguy: We don't need the `else` block here but I think it is good to list it explicitly to help avoid any possible confusion

systemtest/src/test/java/io/strimzi/systemtest/cruisecontrol/CruiseControlConfigurationST.java
kyguy: Since this is very specific to this test method and not used in any other classes I really think we should leave it here but if you feel strongly about the move I will move it!

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/cruisecontrol/Capacity.java
```diff
@@ -79,22 +170,22 @@ private JsonObject generateBrokerCapacity(int brokerId, JsonObject diskCapacity,
      * @param idx Index of the broker
      * @return Disk capacity configuration value as a JsonObject for broker idx
      */
-    private JsonObject generateJbodDiskCapacity(Storage storage, int idx) {
-        JsonObject json = new JsonObject();
+    private static JsonObject generateJbodDiskCapacity(Storage storage, int idx) {
```
mimaison, line 173: In Kafka we typically talk about broker id rather than broker index. Should we rename the argument to `brokerId`?

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/cruisecontrol/Capacity.java
mimaison: So let's update the comments for the methods

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/cruisecontrol/Capacity.java
im-konge: Okay, makes sense :)

systemtest/src/test/java/io/strimzi/systemtest/cruisecontrol/CruiseControlConfigurationST.java
im-konge: Yeah, I know. The thing is that when someone will write, for example, similar test, he will create new method (if he wouldn't know that there is similar method already "implemented") -> for these things we are adding them to `*Utils` classes. We have also `CruiseControlUtils`, if you don't want to add it to generic `StUtils`.

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/cruisecontrol/Capacity.java
mimaison: Done, it looks good!

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/cruisecontrol/Capacity.java
mimaison: This one still says `represented as a Double.` even though we changed the return type to `String`

mimaison (approved):

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/cruisecontrol/BrokerCapacity.java
```diff
@@ -0,0 +1,77 @@
...
+    public static final String DEFAULT_CPU_UTILIZATION_CAPACITY = "100";  // as a percentage (0-100)
+    public static final String DEFAULT_DISK_CAPACITY_IN_MIB = "100000";
+    public static final String DEFAULT_INBOUND_NETWORK_CAPACITY_IN_KIB_PER_SECOND = "10000";
+    public static final String DEFAULT_OUTBOUND_NETWORK_CAPACITY_IN_KIB_PER_SECOND = "10000";
+
+    private int id;
+    private String cpu;
+    private String disk;
+    private String inboundNetworkKiBPerSecond;
+    private String outboundNetworkKiBPerSecond;
```
ppatierno, line 22: Maybe we want to align this field names to the corresponding one in the `BrokerCapacity` from the api?

ppatierno (approved):
LGTM. I left a couple of nits but no need to change if you want to stay as it is.

im-konge (approved):

scholzj:
@kyguy FYI: I guess the regression error is related to your PR.

tomncooper (approved):
LGTM

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/cruisecontrol/BrokerCapacity.java
```diff
@@ -17,12 +17,12 @@ public class BrokerCapacity {
 
     private int id;
     private String cpu;
-    private String disk;
+    private Object disk;
```
kyguy, line 20: I changed this to an Object type because the disk configuration can either be a String or JsonObject, depending on whether it is a non-JBOD or JBOD configuration is used respectively. I could force this type to JsonObject but makes the code a little messier IMO, let me know if I should do it anyway!

kyguy:
@scholzj regression failure was due to some refactoring I did with the disk capacity values, I have one question about the changes I made, let me know what you think!

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/cruisecontrol/BrokerCapacity.java
scholzj: I think we had this discussion somewhere already in the previous PR about intra-broker balancing but did not come up with any better idea. So I guess I can live with it.

@ppatierno @tomncooper Any idea how to make this better?

scholzj (approved):

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/cruisecontrol/BrokerCapacity.java
tomncooper: Well, if you want to avoid using Object, you could create a new class that wraps the logic for dealing with this. Something like `CruiseControlDiskCapacity` that has 2 constructors or takes an Object and figures it out? Then you would call `CruiseControlDiskCapacity.getCapacity()` to convert to what you need.

kyguy: I have just added a unit test for checking the disk configuration in the same way CC reads it in. It catches the error that was flagged by system tests. It needs some refactoring bur feel free to play with it in the meantime.

tomncooper: `CruiseControlDiskCapacity.getCapacity()` could return whatever `JsonObject.put()` requires (or have another method return the json friendly type `CruiseControlDiskCapacity.getJsonObject()`)?

ppatierno: Yeah I was tweaking along these lines, the `JsonObject.put()` wants an `Object` in the end.
In the end we are just wrapping for avoiding an `Object disk` declaration but a fancier `CruiseControlDiskCapacity disk`.
I am playing with it, @kyguy let's sync on this.

ppatierno: I created this new `CapacityDisk` class https://gist.github.com/ppatierno/0ca502bd307dffef4bb5eac70988de83

That should be used where we are using `Object` and then when writing the JSON we should have something like this:

```java
private JsonObject generateBrokerCapacity(BrokerCapacity brokerCapacity) {
        return new JsonObject()
            .put("brokerId", brokerCapacity.getId())
            .put("capacity", new JsonObject()
                .put("DISK", brokerCapacity.getDisk().getCapacity())
                .put("CPU", brokerCapacity.getCpu())
                .put("NW_IN", brokerCapacity.getInboundNetwork())
                .put("NW_OUT", brokerCapacity.getOutboundNetwork())
            )
            .put("doc", brokerCapacity.getDoc());
    }
```

To highlight the call at `.put("DISK", brokerCapacity.getDisk().getCapacity())`.

@kyguy let's sync offline on this to maybe getting improvements on the class. @tomncooper wdyt?

tomncooper: LGTM

kyguy: I have committed Paolo's changes + some refactoring changes, let me know what you think!

ppatierno: LGTM! Thanks Kyle!

scholzj:
@kyguy Is this worth a record in the CHANGELOG.md file?

kyguy:
@scholzj yes, just added record now

PR #6779, opened 2022-05-07
[ST] Fix error handling when internal registry service is missing

scholzj:
### Type of change

- Bugfix

### Description

When the internal registry service is missing, the code currently throws an NPE instead of some proper error message. This PR fixes it to throw a new `RuntimeException`. It also renames the method `isNotKubernetes` to more appropriate `isOpenShift`.


im-konge (approved):
Thanks @scholzj

ppatierno (approved):

PR #6805, opened 2022-05-12
Add support for Kafka 3.1.1

scholzj:
### Type of change

- Enhancement / new feature

### Description

This PR adds support for Kafka 3.1.1 to Strimzi.

### Checklist

- [x] Update CHANGELOG.md


ppatierno (approved):
LGTM. Just one question. Cannot we remove the support for 3.0.0 and 3.1.0 (leaving the patch releases) just to shrink to 3 supported versions and not 5?

scholzj:
> LGTM. Just one question. Cannot we remove the support for 3.0.0 and 3.1.0 (leaving the patch releases) just to shrink to 3 supported versions and not 5?

That would make upgrades more complicated. We can remove 3.0.0 and 3.0.1 completely. But given the 0.29.0 release is imminent, it seems like something what not help users who want to keep using 3.0 and it would not help us much either. My plan was to remove 3.0.0 and 3.0.1 right after we do 0.29.0.

tomncooper (approved):

PR #6800, opened 2022-05-11
Adding `KafkaRebalance` add/remove broker modes for rebalancing after/before scale up/down

ppatierno:
### Type of change

- Enhancement / new feature

### Description

This PR is about implementing the proposal [#35](https://github.com/strimzi/proposals/blob/main/035-rebalance-types-scaling-brokers.md).
It is about adding the new rebalancing modes `add-broker` and `remove-broker` to the `KafkaRebalance` custom resource (using the `full-rebalance` as the default one when missing for backward compatibility).
It also adds the handling of the above modes used when scaling up and down a cluster and the needs for doing a rebalance (moving replicas to the newly added broker after scaling up, moving replicas out of the broker to remove before the scaling down).

The documentation about this new rebalancing modes and usage will be provided in a separated PR.

[UPDATE]

During the review, we decided to go with different names for the `mode` by using just `full` and then plural for the others, `add-brokers` and `remove-brokers`. This is the only difference referring to the approved proposal.

### Checklist

_Please go through this checklist and make sure all applicable tasks have been done_

- [x] Write tests
- [x] Make sure all tests pass
- [x] Check RBAC rights for Kubernetes / OpenShift roles
- [x] Try your changes from Pod inside your Kubernetes and OpenShift cluster, not just locally
- [x] Reference relevant issue(s) and close them after merging
- [x] Update CHANGELOG.md

PR #6807, opened 2022-05-13
Added documentation about rebalancing modes

ppatierno:
### Type of change

- Documentation

### Description

This documentation is related to the new rebalancing modes added for rebalancing together with scale up and down operations.
This is related to the corresponding implementation https://github.com/strimzi/strimzi-kafka-operator/pull/6800 and proposal [35](https://github.com/strimzi/proposals/blob/main/035-rebalance-types-scaling-brokers.md).

### Checklist

- [x] Update documentation

### DIFF
--- a/documentation/modules/cruise-control/con-optimization-proposals.adoc
+++ b/documentation/modules/cruise-control/con-optimization-proposals.adoc
@@ -239,7 +239,7 @@ To change the cached optimization proposal refresh interval, edit the `proposal.
 Consider a shorter interval for fast changing clusters, although this increases the load on the Cruise Control server.
 
 [id='con-optimization-proposals-modes-{context}']
-== Rebalancing modes with broker scaling
+== Rebalancing modes
 
 When using the `KafkaRebalance` custom resource for running a rebalancing operation, you have three available modes:
 
--- a/documentation/modules/cruise-control/con-optimization-proposals.adoc
+++ b/documentation/modules/cruise-control/con-optimization-proposals.adoc
@@ -238,6 +238,39 @@ If you generate an optimization proposal using the default optimization goals, C
 To change the cached optimization proposal refresh interval, edit the `proposal.expiration.ms` setting in the Cruise Control deployment configuration.
 Consider a shorter interval for fast changing clusters, although this increases the load on the Cruise Control server.
 
+[id='con-optimization-proposals-modes-{context}']
+== Rebalancing modes with broker scaling
+
+When using the `KafkaRebalance` custom resource for running a rebalancing operation, you have three available modes:
+
+* `full`
+* `add-brokers`
+* `remove-brokers`
+
+You specify a mode using the `spec.mode` property of the `KafkaRebalance` custom resource.
+
+`full`:: The `full` mode runs a full rebalance by moving replicas across all the brokers in the cluster.
+This is the default mode if the `spec.mode` property is not defined in the `KafkaRebalance` custom resource.
+
+`add-brokers`:: The `add-brokers` mode is used after scaling up a Kafka cluster by adding one or more brokers.
+Normally, after scaling up a Kafka cluster, new brokers are used to host only the partitions of newly created topics.
+If no new topics are created, the newly added brokers are not used and the existing brokers remain under the same load.
+By using the `add-brokers` mode immediately after adding brokers to the cluster, the rebalancing operation moves replicas from existing brokers to the newly added brokers.
+You specify the new brokers as a list using the `spec.brokers` property of the `KafkaRebalance` custom resource.
+
+`remove-brokers`:: The `remove-brokers` mode is used before scaling down a Kafka cluster by removing one or more brokers.
+If you scale down a Kafka cluster, brokers are shut down even if they host replicas.
+This can lead to under-replicated partitions and possibly result in some partitions being under their minimum ISR (in-sync replicas).
+To avoid this potential problem, the `remove-brokers` mode moves replicas off the brokers that are going to be removed.
+When these brokers are not hosting replicas anymore, you can safely run the scaling down operation.
+You specify the brokers you're removing as a list in the `spec.brokers` property in the `KafkaRebalance` custom resource.
+
+In general, use the `full` rebalance mode to rebalance a Kafka cluster by spreading the load across brokers.
+Use the `add-brokers` and `remove-brokers` modes only if you want to scale your cluster up or down and rebalance the replicas accordingly.
+
+The procedure to run a rebalance is actually the same across the three different modes.
+The only difference is with specifying a mode through the `spec.mode` property and, if needed, listing brokers that have been added or will be removed through the `spec.brokers` property.
+
 .Additional resources
 
 * xref:con-optimization-goals-{context}[]
--- a/documentation/modules/cruise-control/con-rebalance-performance.adoc
+++ b/documentation/modules/cruise-control/con-rebalance-performance.adoc
@@ -9,33 +9,6 @@
 You can adjust several performance tuning options for cluster rebalances.
 These options control how partition replica and leadership movements in a rebalance are executed, as well as the bandwidth that is allocated to a rebalance operation.
 
-== Rebalancing modes
-
-When using the `KafkaRebalance` custom resource for running a rebalancing operation, there are three available modes: `full`, `add-brokers` and `remove-brokers`.
-The mode is specified through the corresponding `spec.mode` field in the `KafkaRebalance` custom resource.
-
-The `full` mode runs a full rebalancing by moving replicas across all the brokers in the cluster.
-This is the default mode if the `spec.mode` field is not defined in the `KafkaRebalance` custom resource.
-
-The `add-brokers` mode can be used after scaling up the Apache Kafka cluster by adding one or more brokers.
-After a scaling up, the new brokers are used to host only partitions of new created topics.
-It means that the newly added brokers are actually not used even for long time if no new topics are created.
-By using the `add-brokers` mode, right after a scaling up, the rebalancing operation moves already existing replicas to the newly added brokers as well.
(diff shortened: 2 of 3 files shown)


documentation/modules/cruise-control/con-rebalance-performance.adoc
```diff
@@ -9,6 +9,33 @@
 You can adjust several performance tuning options for cluster rebalances.
 These options control how partition replica and leadership movements in a rebalance are executed, as well as the bandwidth that is allocated to a rebalance operation.
 
+== Rebalancing modes
```
tomncooper, line 12: Rebalance modes are not really a performance tuning option? They are a level above this. I think this would actually be better as a whole new section below "Why use cruise control" that explains the different types of rebalance that are possible. WDYT @PaulRMellor?

tomncooper: Also looking through the whole CC section in the docs we may need to clarify how each section changes (if at all) with respect to the full/add/remove.

documentation/modules/cruise-control/con-rebalance-performance.adoc
```diff
@@ -9,6 +9,33 @@
...
+
+When using the `KafkaRebalance` custom resource for running a rebalancing operation, there are three available modes: `full`, `add-brokers` and `remove-brokers`.
+The mode is specified through the corresponding `spec.mode` field in the `KafkaRebalance` custom resource.
+
+The `full` mode runs a full rebalancing by moving replicas across all the brokers in the cluster.
+This is the default mode if the `spec.mode` field is not defined in the `KafkaRebalance` custom resource.
+
+The `add-brokers` mode can be used after scaling up the Apache Kafka cluster by adding one or more brokers.
+After a scaling up, the new brokers are used to host only partitions of new created topics.
```
tomncooper, line 21: "Normally, after scaling up the Kafka cluster, the new brokers...."

documentation/modules/cruise-control/con-rebalance-performance.adoc
ppatierno: > Rebalance modes are not really a performance tuning option? They are a level above this. I think this would actually be better as a whole new section below "Why use cruise control" that explains the different types of rebalance that are possible.

I had a chat with @PaulRMellor yesterday and this section should be a more general concepts section. I would agree that the title about performance tuning option is not the best. Maybe we should just try to change it?

> Also looking through the whole CC section in the docs we may need to clarify how each section is effected (if at all) with respect to the full/add/remove.

We went through all sections and I didn't find anything affected unless the "Generating optimization proposals" with the procedure. We thought that the best way was not having three different procedure with a lot of duplication but relying on the single one we have today. The first step just explains how to use the 3 modes. The other steps are actually the same for all the 3 modes.

documentation/modules/cruise-control/con-rebalance-performance.adoc
```diff
@@ -9,6 +9,33 @@
...
+It means that the newly added brokers are actually not used even for long time if no new topics are created.
```
tomncooper, line 22: "This means that, if no new topics are created, the newly added brokers are not used and existing broker remain under the same load."

documentation/modules/cruise-control/con-rebalance-performance.adoc
```diff
@@ -9,6 +9,33 @@
...
+By using the `add-brokers` mode, right after a scaling up, the rebalancing operation moves already existing replicas to the newly added brokers as well.
```
tomncooper, line 23: "By using the `add-brokers` mode, immediatly after adding brokers to the cluster, the rebalancing operation moves replicas from existing brokers to the newly added brokers."

documentation/modules/cruise-control/con-rebalance-performance.adoc
```diff
@@ -9,6 +9,33 @@
...
+The new brokers have to be specified as a list in the `spec.brokers` field in the `KafkaRebalance` custom resource.
+
+The `remove-brokers` mode can be used before scaling down the Apache Kafka cluster by removing one or more brokers.
+If you scale down an Apache Kafka cluster, one or more brokers are shut down even if they host some replicas.
```
tomncooper, line 27: This won't be true after @ShubhamRwt's PR so we will need to remember to change this line.

documentation/modules/cruise-control/con-rebalance-performance.adoc
```diff
@@ -9,6 +9,33 @@
...
+It can lead to have under-replicated partitions and maybe some of them under the min-ISR.
```
tomncooper, line 28: "This can lead to under-replicated partitions and possibly result in some partitions being under their min-ISR."

documentation/modules/cruise-control/con-rebalance-performance.adoc
```diff
@@ -9,6 +9,33 @@
...
+To avoid this problem, the `remove-brokers` mode can be used to move replicas off the brokers that are going to be removed.
+When these brokers are not hosting replicas anymore, you can safely run the scaling down operation.
+The brokers to be removed have to be specified as a list in the `spec.brokers` field in the `KafkaRebalance` custom resource.
+
+In general, you should use a `full` rebalancing mode to rebalance your running Apache Kafka cluster spreading the load across brokers based on your goals.
```
tomncooper, line 33: This implies that the add/remove endpoints don't also use goals? They can and also pretty much all the same performance tuning options. This is why I think adding a new section explaining the full/add/remove rebalances at the top of the CC chapter makes sense.

documentation/modules/cruise-control/proc-generating-optimization-proposals.adoc
```diff
@@ -31,7 +37,58 @@ metadata:
...
+  name: my-rebalance
+  labels:
+    strimzi.io/cluster: my-cluster
+spec:
+  mode: add-brokers
+  brokers: [3, 4] <1>
+----
+<1> List of newly added brokers by the scale up operation. This is mandatory.
++
+If you want safely scale down the Apache Kafka cluster, consider to run a rebalance before to do so in order to move replicas off the brokers that are going to be removed.
```
tomncooper, line 73: "If you want safely scale down the Apache Kafka cluster you should run a remove brokers rebalance before hand in order to move replicas off the brokers that are going to be removed."

documentation/modules/cruise-control/con-rebalance-performance.adoc
ppatierno: Maybe I just miss used the sentence "based on your goals". It wasn't related to the "goals" for the proposal :-)

documentation/modules/cruise-control/con-rebalance-performance.adoc
ppatierno: Agree.

documentation/modules/cruise-control/proc-generating-optimization-proposals.adoc
```diff
@@ -31,7 +37,58 @@ metadata:
...
+  name: my-rebalance
+  labels:
+    strimzi.io/cluster: my-cluster
+spec:
+  mode: remove-brokers
+  brokers: [3, 4] <1>
+----
+<1> List of brokers to be removed by the scale down operation. This is mandatory.
++
+All the steps following in this procedure are actually the same whatever is the mode you are using.
```
tomncooper, line 91: "All the steps described in the following procedures are the same regardless of the rebalance mode you are using."

documentation/modules/cruise-control/con-rebalance-performance.adoc
PaulRMellor: I can take a look at this when I do an editorial review

tomncooper (changes requested):
The decription of the modes is good but it is in the wrong place. It is not a tuning option it is more fundermental than that, it should be its own section at the top of the CC chapter IMHO.

PaulRMellor (approved):
Looks good to me

ShubhamRwt (approved):

documentation/modules/cruise-control/con-rebalance-performance.adoc
ppatierno: About the title ... "scaling" makes sense only with the two new modes and not with the "full" one. I think it's misleading to be there.
Fine about where positioning the section.

documentation/modules/cruise-control/con-optimization-proposals.adoc
```diff
@@ -238,6 +238,39 @@ If you generate an optimization proposal using the default optimization goals, C
 To change the cached optimization proposal refresh interval, edit the `proposal.expiration.ms` setting in the Cruise Control deployment configuration.
 Consider a shorter interval for fast changing clusters, although this increases the load on the Cruise Control server.
 
+[id='con-optimization-proposals-modes-{context}']
+== Rebalancing modes with broker scaling
```
tomncooper, line 242: I agree this is probably a better place than in the tuning options and maybe for now we can leave it here. But I do think that we need to do a bit of an overhaul of the CC docs to introduce the modes nearer the top with a more comprehensive introduction. I will open an issue to cover that.

tomncooper (approved):
Happy with description. But think we probably need to improve the introduction to the CC chapter to describe the multiple modes.

documentation/modules/cruise-control/con-optimization-proposals.adoc
ppatierno: @tomncooper out of curiosity ... what do you think we should add to have "a more comprehensive introduction". Isn't the current one covering everything? Should we already improve it?
Other question is about the title here ... I don't think that "with broker scaling" part is right, because when talking about "full" mode, it doesn't involve scaling.

kyguy (approved):
Looks great Paolo!

PR #6814, opened 2022-05-14
Fix Connect and Cruise Control examples

scholzj:
### Type of change

- Bugfix

### Description

I found several issues in the examples for Connect and Cruise Control:
* `inter.broker.protocol.version` in Kafka examples
* The Connect Build example uses Kafka 3.1.0, probably because it was don ein parallel while Kafka 3.2.0 was being added
* The Cruise Control example use singular for the mode instead of plural and don't work

These changes should be picked for the 0.29.0 release.


ppatierno (approved):
Damn! Latest changes from singular to plural on the rebalancing modes. Good catch! Thanks!

PR #6815, opened 2022-05-14
[ST] Skip rack-awareness tests in namespaced system test runs

scholzj:
### Type of change

- Bugfix

### Description

The #6710 PR removed the assumption which made sure that the rack-awareness tests were skipped when running the system tests in a namespaced mode. As a result, the `namespace-rbac-scope-regression` job is failing now in these tests. This PR adds the assumption back to the tests to make sure they are skipped.


fvaleri:
Thanks @scholzj.

Frawless (approved):

im-konge (approved):

PR #6834, opened 2022-05-18
Added metrics configuration to the Cruise Control configuration doc

ppatierno:
### Type of change

- Documentation

### Description

Currently we have some information about metrics configuration for Cruise Control in the deploying guide [here](https://strimzi.io/docs/operators/in-development/deploying.html#con-metrics-cruise-control-str).
Anyway, as a user I would expect some information in the more specific Cruise Control [configuration](https://strimzi.io/docs/operators/latest/configuring.html#ref-cruise-control-configuration-str) section, where we have logging and other stuff but metrics section is missing.
This trivial PR adds this short section to provide an introduction and linking to the common metrics config which is the same for all the components.

### Checklist

- [x] Update documentation

### DIFF
--- a/documentation/modules/cruise-control/ref-cruise-control-configuration.adoc
+++ b/documentation/modules/cruise-control/ref-cruise-control-configuration.adoc
@@ -171,6 +171,34 @@ spec:
     # ...
 ----
 
+[[metrics-configuration]]
+[discrete]
+== Metrics configuration
+
+Cruise Control expose metrics to provide information about the Kafka cluster as well as its internal components.
+Metrics about rebalancing executions, disk and broker failures, optimization proposals and REST API requests could be valuable to monitor your cluster.
+You can find more information on the official Cruise Control https://github.com/linkedin/cruise-control/wiki/Sensors[Sensors] Wiki page.
+
+You can find more information see the xref:con-common-configuration-prometheus-reference[metricsConfig].
+
+.Metrics configuration
+[source,yaml,subs="+quotes,attributes"]
+----
+apiVersion: {KafkaApiVersion}
+kind: Kafka
+# ...
+spec:
+  cruiseControl:
+    # ...
+    metricsConfig:
+      type: jmxPrometheusExporter
+      valueFrom:
+        configMapKeyRef:
+          name: cruise-control-metrics
+          key: metrics-config.yml
+    # ...
+----
+
 [[api-security-configuration]]
 [discrete]
 == Cruise Control REST API security


scholzj (commented):
My 2 cents ... we should keep this information centralised and not copy it into every single resource which supports it.

ppatierno:
It's centralized in the common section I linked but the difference is that, while for Kafka, KafkaConnect, KafkaMirrorMaker all the examples have a `metricsConfig` field and pointing to the common configuration section, we are missing this in the Cruise Control related guide so no one knows you can configure metrics in Cruise Control.
And the deployment guide is, maybe, not the right place to talk about them.

scholzj:
> It's centralized in the common section I linked but the difference is that, while for Kafka, KafkaConnect, KafkaMirrorMaker all the examples have a metricsConfig field and pointing to the common configuration section, we are missing this in the Cruise Control related guide so no one knows you can configure metrics in Cruise Control.
And the deployment guide is, maybe, not the right place to talk about them.

I don't really follow where do you think Cruise Control is not mentioned. It seemed to me to mention equally the metrics for all operands.

ppatierno:
As I mentioned the other components have bigger examples with `metricsConfig` inside and a callout explaining what it is and linking to the common metrics config. The same about logging. With Cruise Control we have distinct section and not a bigger example with everything in. That's the difference. For this reason, we don't mention metrics in the Cruise Control configuration section. I don't want to push for this PR but at least we should agree in a way to mention/show how to configure metrics for Cruise Control.

scholzj:
> As I mentioned the other components have bigger examples with metricsConfig inside and a callout explaining what it is and linking to the common metrics config. The same about logging. With Cruise Control we have distinct section and not a bigger example with everything in. That's the difference. For this reason, we don't mention metrics in the Cruise Control configuration section. I don't want to push for this PR but at least we should agree in a way to mention/show how to configure metrics for Cruise Control.

So, why don't you do the same? The 8.5 chapter already has the example. You can just add it there and add the callout, or?

ppatierno:
The 8.5 has different examples because we are splitting the Cruise Control Configuration in more sections (i.e. Cross-Origin Resource Sharing configuration, Capacity configuration, Logging configuration, Cruise Control REST API security ...) while as far as I can see the other components have a "huge" example with everything and callout to explain everything (i.e. see the Configuring Kafka https://strimzi.io/docs/operators/latest/configuring.html#proc-config-kafka-str).
If we want to do the same we should revisit the entire Cruise Control Configuration section doing the same: a bigger example and callout.

ppatierno:
@PaulRMellor maybe that's the right way to go imho. I will close this PR then. The work is more than this.

PR #6836, opened 2022-05-18
docs: updates from review of tuning content

PaulRMellor:

**Documentation updates**

Updates the [Tuning Kafka configuration](https://strimzi.io/docs/operators/in-development/configuring.html#assembly-tuning-config-str) content in the _Configuring guide_ following a review 

### Checklist

- [ ] Write tests
- [x] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/documentation/assemblies/managing/assembly-tuning-config.adoc
+++ b/documentation/assemblies/managing/assembly-tuning-config.adoc
@@ -23,7 +23,7 @@ The following tools help with Kafka tuning:
 
 * xref:cruise-control-concepts-str[Cruise Control] generates optimization proposals that you can use to assess and implement a cluster rebalance
 * xref:proc-setting-broker-limits-str[Kafka Static Quota plugin] sets limits on brokers
-* xref:type-Rack-reference[Rack configuration] spreads broker partitions across racks
+* xref:type-Rack-reference[Rack configuration] spreads broker partitions across racks and allows consumers to fetch data from the nearest replica
 
 //Tips to optimize the performance of brokers, and producer and consumer clients
 include::modules/con-managed-broker-config-properties.adoc[leveloffset=+1]
--- a/documentation/modules/managing/con-broker-config-properties.adoc
+++ b/documentation/modules/managing/con-broker-config-properties.adoc
@@ -109,10 +109,11 @@ As an exception, you might want to reduce the settings in a single-broker _test_
 Network threads handle requests to the Kafka cluster, such as produce and fetch requests from client applications.
 Produce requests are placed in a request queue. Responses are placed in a response queue.
 
-The number of network threads should reflect the replication factor and the levels of activity from client producers and consumers interacting with the Kafka cluster.
+The number of network threads per listener should reflect the replication factor and the levels of activity from client producers and consumers interacting with the Kafka cluster.
 If you are going to have a lot of requests, you can increase the number of threads, using the amount of time threads are idle to determine when to add more threads.
 
-To reduce congestion and regulate the request traffic, you can limit the number of requests allowed in the request queue before the network thread is blocked.
+To reduce congestion and regulate the request traffic, you can limit the number of requests allowed in the request queue.
+When the request queue is full, all incoming traffic is blocked.
 
 I/O threads pick up requests from the request queue to process them.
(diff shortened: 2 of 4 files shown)


documentation/modules/managing/con-broker-config-properties.adoc
```diff
@@ -109,10 +109,11 @@ As an exception, you might want to reduce the settings in a single-broker _test_
 Network threads handle requests to the Kafka cluster, such as produce and fetch requests from client applications.
 Produce requests are placed in a request queue. Responses are placed in a response queue.
 
-The number of network threads should reflect the replication factor and the levels of activity from client producers and consumers interacting with the Kafka cluster.
+The number of network threads per listener should reflect the replication factor and the levels of activity from client producers and consumers interacting with the Kafka cluster.
```
PaulRMellor, line 112: @mimaison -- do we we need to say something about prefixing with listener name?

documentation/assemblies/managing/assembly-tuning-config.adoc
```diff
@@ -23,7 +23,7 @@ The following tools help with Kafka tuning:
 
 * xref:cruise-control-concepts-str[Cruise Control] generates optimization proposals that you can use to assess and implement a cluster rebalance
 * xref:proc-setting-broker-limits-str[Kafka Static Quota plugin] sets limits on brokers
-* xref:type-Rack-reference[Rack configuration] spreads broker partitions across racks
+* xref:type-Rack-reference[Rack configuration] spreads broker partitions across racks and allows consumers to fetch data from the nearest replica
```
scholzj, line 26: The Rack-awareness is just one of many things which need to be achieved to consume from nearest replica. So I think we need to be careful about this to make sure it does not suggest that it is all what is needed.

The distribution of the replicas across the partitions is also significantly more important for reliability and avauilability. So I think we need to make sure it does not get lost in this.

documentation/modules/managing/con-broker-config-properties.adoc
scholzj: I don't think we currently allow to configure this differently for different listeners.

mimaison: It's new in 3.2 so yes we don't have it yet.

PaulRMellor: Okay, thanks

documentation/assemblies/managing/assembly-tuning-config.adoc
PaulRMellor: > The Rack-awareness is just one of many things which need to be achieved to consume from nearest replica. So I think we need to be careful about this to make sure it does not suggest that it is all what is needed.

@scholzj Should we mention `affinity` rules and `topologySpreadConstraints` here? Or were you thinking of something else?

PaulRMellor: > The distribution of the replicas across the partitions is also significantly more important for reliability and avauilability. So I think we need to make sure it does not get lost in this.

We could add a note under these links:

_If you're considering rack configuration, make sure that topic  replication is set up for high availability. Having replicated topics across brokers, you can use rack awareness to distribute the replicas across racks._

scholzj: TBH, I would be fine to leave it as it was. The chapter it links to has the details. But if you want to change this, then it should be clear to focus on the main use-case for rack awareness which is availability and reliability. The fetching from closest replica is a niche use-case with a lot of gotchas and caveats.

scholzj: > We could add a note under these links:
> _If you're considering rack configuration, make sure that topic replication is set up for high availability. Having replicated topics across brokers, you can use rack awareness to spread partition replicas across racks._

Its not like this is wrong. It is more that I don't feel like it adds too much nd at the end you still need to go to https://strimzi.io/docs/operators/in-development/full/configuring.html#type-Rack-reference and read the whole stuff there.

PaulRMellor: Thanks. I'll leave it as it is.

scholzj (approved):

PR #6839, opened 2022-05-19
[systemtest] Add tests for `add_broker` and `remove_broker` features in CC

im-konge:

### Type of change

- New tests

### Description

This PR adds two new tests for the new CC features - `add_broker` and `remove_broker` mods.
First tests checking that we are able to use the `add_broker`/`remove_broker` endpoints (as other tests in the `CruiseControlApiST`), second is actually using the `KafkaRebalance` resource during Kafka scaling up and down.

Also, this PR changes image of `Scraper` pod, which will now contain our Kafka image - for using Kafka scripts from different pod than the Kafka one.

### Checklist

- [x] Write tests

### DIFF
--- a/systemtest/src/test/java/io/strimzi/systemtest/cruisecontrol/CruiseControlST.java
+++ b/systemtest/src/test/java/io/strimzi/systemtest/cruisecontrol/CruiseControlST.java
@@ -357,7 +357,7 @@ public class CruiseControlST extends AbstractST {
         assertThat(kafkaRebalanceStatus.getConditions().get(0).getReason(), is("InvalidResourceException"));
     }
 
-    @ParallelNamespaceTest
+    @IsolatedTest
     void testCruiseControlDuringBrokerScaleUpAndDown(ExtensionContext extensionContext) {
         final TestStorage testStorage = new TestStorage(extensionContext);
         final int initialReplicas = 3;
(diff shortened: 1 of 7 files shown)


systemtest/src/test/java/io/strimzi/systemtest/cruisecontrol/CruiseControlST.java
```diff
@@ -348,4 +349,57 @@ void testCruiseControlIntraBrokerBalancingWithoutSpecifyingJBODStorage(Extension
...
+                .build()
+        );
+
+        KafkaRebalanceUtils.waitForKafkaRebalanceCustomResourceState(testStorage.getNamespaceName(),  testStorage.getClusterName(), KafkaRebalanceState.ProposalReady);
+        KafkaRebalanceUtils.doRebalancingProcess(new Reconciliation("test", KafkaRebalance.RESOURCE_KIND, testStorage.getNamespaceName(), testStorage.getClusterName()), testStorage.getNamespaceName(), testStorage.getClusterName());
+
+        LOGGER.info("Scaling Kafka down to {}", initialReplicas);
+
+        KafkaResource.replaceKafkaResourceInSpecificNamespace(testStorage.getClusterName(), kafka -> kafka.getSpec().getKafka().setReplicas(initialReplicas), testStorage.getNamespaceName());
+        RollingUpdateUtils.waitForComponentScaleUpOrDown(testStorage.getNamespaceName(), testStorage.getKafkaSelector(), initialReplicas, kafkaPods);
```
Frawless, line 403: Shouldn't we have here some check that everything is fine?

im-konge: What do you prefer?
The `remove_broker` rebalancing is done before scaling down and the `waitForComponentScaleUpOrDown` ensures that pods are ready.
Maybe we can do full rebalance, but it will add more execution time and I think that it is not necessary here.

im-konge: The check if the KafkaRebalance is in `Ready` state is inside the `doRebalancingProcess` method - any other check is not needed (as discussed offline)

ppatierno (changes requested):
I think that the test needs to be more complete.
What's missing is how we are checking that a rebalancing happened on scaling up and down and replicas were moved accordingly.
What I usually do for testing that rebalancing is working fine is the following.

Having a Kafka cluster with 3 nodes initially ...

1. Create a topic "my-topic" with 10 partitions and 3 as replication factor.
2. Scaling up the cluster to 5 brokers
3. Check that the my-topic has still replicas only on the first 3 brokers. I usually use the `kafka-topics.sh` tool with the related `describe` command, I guess that we can use Java Admin Client API here to get this information. Check that they are all on brokers 0,1,2 (even if we scaled up to 5).
4. Run the rebalancing with `add-brokers`.
5. Check that now at least one replica of my-topic is on 3 or 4.

On scaling down ...

1. Run  rebalance with `remove-brokers`
2. Check that the my-topic doesn't have replicas on brokers 3 and 4 anymore
3. Do the scaling down

Having more partitions for topic my-topic would help to have more possibilities that other brokers are used to place them. So you could use even more than 10 partitions or using the __consumer_offset topic instead which has 50 partitions.

systemtest/src/test/java/io/strimzi/systemtest/cruisecontrol/CruiseControlST.java
```diff
@@ -348,4 +352,71 @@ void testCruiseControlIntraBrokerBalancingWithoutSpecifyingJBODStorage(Extension
...
+        Map<String, String> kafkaPods = PodUtils.podSnapshot(testStorage.getNamespaceName(), testStorage.getKafkaSelector());
+
+        LOGGER.info("Scaling Kafka up to {}", scaleTo);
+
+        KafkaResource.replaceKafkaResourceInSpecificNamespace(testStorage.getClusterName(), kafka -> kafka.getSpec().getKafka().setReplicas(scaleTo), testStorage.getNamespaceName());
+        kafkaPods = RollingUpdateUtils.waitForComponentScaleUpOrDown(testStorage.getNamespaceName(), testStorage.getKafkaSelector(), scaleTo, kafkaPods);
+
+        LOGGER.info("Checking that {} topic has replicas on first 3 brokers", testStorage.getTopicName());
+        List<String> topicReplicas = KafkaTopicUtils.getKafkaTopicReplicasForEachPartition(testStorage.getNamespaceName(), testStorage.getTopicName(), KafkaResources.kafkaPodName(testStorage.getClusterName(), 0), KafkaResources.plainBootstrapAddress(testStorage.getClusterName()));
+        assertEquals(0, (int) topicReplicas.stream().filter(line -> line.contains("3") || line.contains("4")).count());
```
ppatierno, line 376: Let's add this test even before the scaling up when the topic was created.

ppatierno (approved):
One suggestion but LGTM overall. Thanks!

systemtest/src/test/java/io/strimzi/systemtest/cruisecontrol/CruiseControlST.java
im-konge: Makes sense, done :)

Frawless (approved):

PR #6855, opened 2022-05-23
[systemtest] `ConnectIsolatedST` - fixups

im-konge:

### Type of change

- Tests fixes

### Description

This PR fixes several tests inside `ConnectIsolatedST`, where we are checking Connect during scaling up and down - because we are now using Connect Build for building connect image with `File` plugin, the number of pods is not matching - because of extra `build` pods.

Now we will gather pods containing prefix without pods, which contains `exclude` substring

### Checklist


Frawless (changes requested):
IT would be much more better to filter pods by labels:

```kubeClient(namespaceName).listPods(Annotations.STRIMZI_DOMAIN + "name", KafkaConnectResources.deploymentName(clusterName))``` this works fine

im-konge:
@Frawless forgot about the labels, good catch :)

Frawless (approved):

PR #6859, opened 2022-05-24
Add support for UserOperator and Authorization to KRaft mode

scholzj:
### Type of change

- Enhancement / new feature

### Description

Kafka 3.2.0 added the support for `StandardAuhtorizer` - a new authorizer which instead of ZooKeeper stores the data in the KRaft metadata. The interface is the same as for the original ZooKeeper based authorizer, just the internal functionality changed.

This PR adds support for this authorizer to the Cluster Operator when KRaft mode is enabled. It also _enables_ User Operator in the KRaft mode (previously, the whole Entity Operator was unsupported in KRaft, now it is only the Topic Operator). The User Operator on its own does not work out of the box because Kafka 3.2.0 still does not support SCRAM-SHA authentication. So a new flag is passed to the User Operator to indecate if the SCRAM-SHA APIs are available. If they are notavailable (i.e. in the KRaft mode, all calls related to them will be disabled). This is expected to be enabled again once SCRAM-SHA is supported by Kafka in KRaft mode (possibly Kafka 3.3?).

### Checklist

- [x] Write tests
- [x] Update documentation
- [x] Update CHANGELOG.md

### DIFF
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/model/EntityOperator.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/model/EntityOperator.java
@@ -109,9 +109,11 @@ public class EntityOperator extends AbstractModel {
      * @param reconciliation The reconciliation
      * @param kafkaAssembly desired resource with cluster configuration containing the Entity Operator one
      * @param versions The versions.
+     * @param kraftEnabled Indicates whether KRaft is used in the Kafka cluster
+     *
      * @return Entity Operator instance, null if not configured in the ConfigMap
      */
-    public static EntityOperator fromCrd(Reconciliation reconciliation, Kafka kafkaAssembly, KafkaVersion.Lookup versions) {
+    public static EntityOperator fromCrd(Reconciliation reconciliation, Kafka kafkaAssembly, KafkaVersion.Lookup versions, boolean kraftEnabled) {
         EntityOperatorSpec entityOperatorSpec = kafkaAssembly.getSpec().getEntityOperator();
 
         if (entityOperatorSpec != null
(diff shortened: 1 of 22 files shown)


cluster-operator/src/main/java/io/strimzi/operator/cluster/model/EntityOperator.java
```diff
@@ -109,9 +109,11 @@ protected EntityOperator(Reconciliation reconciliation, HasMetadata resource) {
      * @param reconciliation The reconciliation
      * @param kafkaAssembly desired resource with cluster configuration containing the Entity Operator one
      * @param versions The versions.
+     * @param scramShaAdminApiSupported Indicates whether the SCRAM-SHA API is supported
+     *
      * @return Entity Operator instance, null if not configured in the ConfigMap
      */
-    public static EntityOperator fromCrd(Reconciliation reconciliation, Kafka kafkaAssembly, KafkaVersion.Lookup versions) {
+    public static EntityOperator fromCrd(Reconciliation reconciliation, Kafka kafkaAssembly, KafkaVersion.Lookup versions, boolean scramShaAdminApiSupported) {
```
ppatierno, line 116: my only concern is about the naming of the new parameter. I know why it's named this way and I can see your point but at same time we pass that parameter this way `!config.featureGates().useKRaftEnabled()`, so we are actually saying if KRaft mode is enabled or not implying that it means SCRAM-SHA supported or not. Can't we imply this inside the EO instead of the API. I mean, having `kraftEnabled` as parameter in the API which translate to not having SCRAM-SHA internally for now. Not a strong opinion, I thought it just worths a discussion :-)

scholzj: I'm not sure it makes any difference one way or another. The question is, where would you wan to change it. The way it is done now, it decides early about the specific of what exactly is not supported. Alternative would be to pass the _KRaft_ flag up to the user operator and decide this only there. So, which way would you prefer?

ppatierno: Yeah that is what I would do. Passing to UO if KRaft is enabled or not and encapsulate into the UO logic what is supported or not (i.e. SCRAM-SHA) based on the mode flag.

ppatierno (approved):

tomncooper (approved):

PR #6870, opened 2022-05-26
[systemtest] Remove NetworkPolicies from KafkaConnect templates

im-konge:

### Type of change

- Refactoring

### Description

Until now we had creation of `Scraper` pod inside the `KafkaConnectTemplates` class, as part of every deployment of `KafkaConnect` (if we didn't specify that we don't want to create `Scraper` pod), together with deploying `NetworkPolicies` to allow access to `KafkaConnect` API from the `Scraper` pod. This approach contains few issues: we have to pass the `extensionContext` everytime we want to deploy `KafkaConnect`, steps, which are executed, are not enough clear (why we are deploying a `Scraper` pod when we want to deploy KC) etc.

This PR removes `Scraper` pod creation and deployment of NPs from the `KafkaConnectTemplates` and does a small refactor to our template classes.

### Checklist


Frawless (approved):

im-konge:
The test failures are not connected to my PR - it seems that `TopicOperator` allows forbidden options to be set - such as decreasing partitions.

This will be fixed in one of the future PR (either STs or TO behavior)

PR #6890, opened 2022-06-02
Handled `AnnotatedRouteNoHostException` causing the `KafkaRebalance` to get into `NotReady` state when CC is enabled form Disabled state

ShubhamRwt:

### Type of change

- Bugfix

### Description

This PR fixes the issue for `KafkaRebalance` going into `NotReady` state when CC is enabled from Disabled state due to the `AnnotatedRouteNoHostException`

### Checklist

- [ ] Write tests
- [ ] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceAssemblyOperator.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceAssemblyOperator.java
@@ -1365,7 +1365,7 @@ public class KafkaRebalanceAssemblyOperator
         // We check for `CruiseControlRetriableConnectionException` if request are made when CC was not active since not handling this case will move the KafkaRebalance to `NotReady`
         // and it will stay in `NotReady` state until we apply `refresh`
         if (rebalanceAnnotation == KafkaRebalanceAnnotation.refresh
-                || kafkaRebalance.getStatus().getConditions().stream().anyMatch(condition -> "CruiseControlRetriableConnectException".equals(condition.getReason()))) {
+                || kafkaRebalance.getStatus().getConditions().stream().anyMatch(condition -> "CruiseControlRetriableConnectionException".equals(condition.getReason()))) {
             return true;
         } else {
             return CruiseControlIssues.checkForMatch(kafkaRebalance.getStatus().getConditions());
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceAssemblyOperator.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceAssemblyOperator.java
@@ -1364,7 +1364,7 @@ public class KafkaRebalanceAssemblyOperator
         // check if refresh annotation applied or request are made when CC is not active
         if (rebalanceAnnotation == KafkaRebalanceAnnotation.refresh
                 || kafkaRebalance.getStatus().getConditions().stream().anyMatch(condition -> "ConnectException".equals(condition.getReason()))
-                || kafkaRebalance.getStatus().getConditions().stream().anyMatch(condition -> "AnnotatedNoRouteToHostException".equals(condition.getReason()))) {
+                || kafkaRebalance.getStatus().getConditions().stream().anyMatch(condition -> "NoRouteToHostException".equals(condition.getReason()))) {
             return true;
         } else {
             return CruiseControlIssues.checkForMatch(kafkaRebalance.getStatus().getConditions());
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceAssemblyOperator.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceAssemblyOperator.java
@@ -1362,7 +1362,9 @@ public class KafkaRebalanceAssemblyOperator
 
     private boolean shouldRenewRebalance(KafkaRebalance kafkaRebalance, KafkaRebalanceAnnotation rebalanceAnnotation) {
         // check if refresh annotation applied or request are made when CC is not active
-        if (rebalanceAnnotation == KafkaRebalanceAnnotation.refresh || kafkaRebalance.getStatus().getConditions().stream().anyMatch(condition -> "ConnectException".equals(condition.getReason()))) {
+        if (rebalanceAnnotation == KafkaRebalanceAnnotation.refresh
+                || kafkaRebalance.getStatus().getConditions().stream().anyMatch(condition -> "ConnectException".equals(condition.getReason()))
+                || kafkaRebalance.getStatus().getConditions().stream().anyMatch(condition -> "AnnotatedNoRouteToHostException".equals(condition.getReason()))) {
             return true;
         } else {
             return CruiseControlIssues.checkForMatch(kafkaRebalance.getStatus().getConditions());
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceAssemblyOperator.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceAssemblyOperator.java
@@ -1362,7 +1362,7 @@ public class KafkaRebalanceAssemblyOperator
 
     private boolean shouldRenewRebalance(KafkaRebalance kafkaRebalance, KafkaRebalanceAnnotation rebalanceAnnotation) {
         // check if refresh annotation applied or request are made when CC is not active
-        // We check for `CruiseControlRetriableConnectException` if request are made when CC was not active since not handling this case will move the KafkaRebalance to `NotReady`
+        // We check for `CruiseControlRetriableConnectionException` if request are made when CC was not active since not handling this case will move the KafkaRebalance to `NotReady`
         // and it will stay in `NotReady` state until we apply `refresh`
         if (rebalanceAnnotation == KafkaRebalanceAnnotation.refresh
                 || kafkaRebalance.getStatus().getConditions().stream().anyMatch(condition -> "CruiseControlRetriableConnectException".equals(condition.getReason()))) {
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceAssemblyOperator.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceAssemblyOperator.java
@@ -1362,9 +1362,10 @@ public class KafkaRebalanceAssemblyOperator
 
     private boolean shouldRenewRebalance(KafkaRebalance kafkaRebalance, KafkaRebalanceAnnotation rebalanceAnnotation) {
         // check if refresh annotation applied or request are made when CC is not active
+        // We check for `CruiseControlRetriableConnectException` if request are made when CC was not active since not handling this case will move the KafkaRebalance to `NotReady`
+        // and it will stay in `NotReady` state until we apply `refresh`
         if (rebalanceAnnotation == KafkaRebalanceAnnotation.refresh
-                || kafkaRebalance.getStatus().getConditions().stream().anyMatch(condition -> "ConnectException".equals(condition.getReason()))
-                || kafkaRebalance.getStatus().getConditions().stream().anyMatch(condition -> "NoRouteToHostException".equals(condition.getReason()))) {
+                || kafkaRebalance.getStatus().getConditions().stream().anyMatch(condition -> "CruiseControlRetriableConnectException".equals(condition.getReason()))) {
             return true;
         } else {
             return CruiseControlIssues.checkForMatch(kafkaRebalance.getStatus().getConditions());
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceAssemblyOperator.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceAssemblyOperator.java
@@ -1361,9 +1361,9 @@ public class KafkaRebalanceAssemblyOperator
     }
 
     private boolean shouldRenewRebalance(KafkaRebalance kafkaRebalance, KafkaRebalanceAnnotation rebalanceAnnotation) {
-        // check if refresh annotation applied or request are made when CC is not active
-        // We check for `CruiseControlRetriableConnectionException` if request are made when CC was not active since not handling this case will move the KafkaRebalance to `NotReady`
-        // and it will stay in `NotReady` state until we apply `refresh`
+        // Checks if `refresh` annotation applied or request are made when CC is not active
+        // We check for `CruiseControlRetriableConnectionException` which is thrown if requests are made when CC was not active. Not handling this case will move the KafkaRebalance to `NotReady` state
+        // and it will stay in `NotReady` state until we apply `refresh` annotation
         if (rebalanceAnnotation == KafkaRebalanceAnnotation.refresh
                 || kafkaRebalance.getStatus().getConditions().stream().anyMatch(condition -> "CruiseControlRetriableConnectionException".equals(condition.getReason()))) {
             return true;
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/resource/cruisecontrol/CruiseControlApiImpl.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/resource/cruisecontrol/CruiseControlApiImpl.java
@@ -437,7 +437,7 @@ public class CruiseControlApiImpl implements CruiseControlApi {
             // so we catch and raise a TimeoutException instead
             result.fail(new TimeoutException(t.getMessage()));
         } else if (t instanceof NoRouteToHostException) {
-            // Netty throws a AnnotatedNoRouteToHostException when the request times out
+            // Netty throws a AnnotatedNoRouteToHostException when it cannot resolve the host
             //so we catch and raise a NoRouteToHostException instead
             result.fail(new NoRouteToHostException(t.getMessage()));
         } else if (t instanceof ConnectException) {
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/resource/cruisecontrol/CruiseControlApiImpl.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/resource/cruisecontrol/CruiseControlApiImpl.java
@@ -439,7 +439,7 @@ public class CruiseControlApiImpl implements CruiseControlApi {
         } else if (t instanceof NoRouteToHostException || t instanceof ConnectException) {
             // Netty throws a AnnotatedNoRouteToHostException (inherits from NoRouteToHostException) and  when it cannot resolve the host
             // Vert.x throws a AnnotatedConnectException (inherits from ConnectException) when the request times out
-            // so we catch and raise a CruiseControlRestException instead since
+            // so we catch and raise a CruiseControlRetriableConnectException instead since
             result.fail(new CruiseControlRetriableConnectException(t.getMessage()));
         }  else {
             result.fail(t);
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/resource/cruisecontrol/CruiseControlApiImpl.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/resource/cruisecontrol/CruiseControlApiImpl.java
@@ -437,9 +437,9 @@ public class CruiseControlApiImpl implements CruiseControlApi {
             // so we catch and raise a TimeoutException instead
             result.fail(new TimeoutException(t.getMessage()));
         } else if (t instanceof NoRouteToHostException || t instanceof ConnectException) {
-            // Netty throws a AnnotatedNoRouteToHostException (inherits from NoRouteToHostException) and  when it cannot resolve the host
+            // Netty throws a AnnotatedNoRouteToHostException (inherits from NoRouteToHostException) when it cannot resolve the host
             // Vert.x throws a AnnotatedConnectException (inherits from ConnectException) when the request times out
-            // so we catch and raise a CruiseControlRetriableConnectionException instead since
+            // so we catch and raise a CruiseControlRetriableConnectionException instead
             result.fail(new CruiseControlRetriableConnectionException(t));
         }  else {
             result.fail(t);
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/resource/cruisecontrol/CruiseControlApiImpl.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/resource/cruisecontrol/CruiseControlApiImpl.java
@@ -22,6 +22,7 @@ import io.vertx.core.json.JsonObject;
 import io.vertx.core.net.PemTrustOptions;
 
 import java.net.ConnectException;
+import java.net.NoRouteToHostException;
 import java.nio.charset.StandardCharsets;
 import java.util.concurrent.TimeoutException;
 
@@ -435,6 +436,10 @@ public class CruiseControlApiImpl implements CruiseControlApi {
             // Vert.x throws a NoStackTraceTimeoutException (inherits from TimeoutException) when the request times out
             // so we catch and raise a TimeoutException instead
             result.fail(new TimeoutException(t.getMessage()));
+        } else if (t instanceof NoRouteToHostException) {
+            // Netty throws a AnnotatedNoRouteToHostException when the request times out
+            //so we catch and raise a NoRouteToHostException instead
+            result.fail(new NoRouteToHostException(t.getMessage()));
         } else if (t instanceof ConnectException) {
             // Vert.x throws a AnnotatedConnectException (inherits from ConnectException) when the request times out
             // so we catch and raise a ConnectException instead
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/resource/cruisecontrol/CruiseControlApiImpl.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/resource/cruisecontrol/CruiseControlApiImpl.java
@@ -436,15 +436,12 @@ public class CruiseControlApiImpl implements CruiseControlApi {
             // Vert.x throws a NoStackTraceTimeoutException (inherits from TimeoutException) when the request times out
             // so we catch and raise a TimeoutException instead
             result.fail(new TimeoutException(t.getMessage()));
-        } else if (t instanceof NoRouteToHostException) {
-            // Netty throws a AnnotatedNoRouteToHostException (inherits from NoRouteToHostException) when it cannot resolve the host
-            //so we catch and raise a NoRouteToHostException instead
-            result.fail(new NoRouteToHostException(t.getMessage()));
-        } else if (t instanceof ConnectException) {
+        } else if (t instanceof NoRouteToHostException || t instanceof ConnectException) {
+            // Netty throws a AnnotatedNoRouteToHostException (inherits from NoRouteToHostException) and  when it cannot resolve the host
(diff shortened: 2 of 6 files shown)


cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceAssemblyOperator.java
```diff
@@ -1362,7 +1362,9 @@ private boolean hasRebalanceAnnotation(KafkaRebalance kafkaRebalance) {
 
     private boolean shouldRenewRebalance(KafkaRebalance kafkaRebalance, KafkaRebalanceAnnotation rebalanceAnnotation) {
         // check if refresh annotation applied or request are made when CC is not active
-        if (rebalanceAnnotation == KafkaRebalanceAnnotation.refresh || kafkaRebalance.getStatus().getConditions().stream().anyMatch(condition -> "ConnectException".equals(condition.getReason()))) {
+        if (rebalanceAnnotation == KafkaRebalanceAnnotation.refresh
+                || kafkaRebalance.getStatus().getConditions().stream().anyMatch(condition -> "ConnectException".equals(condition.getReason()))
+                || kafkaRebalance.getStatus().getConditions().stream().anyMatch(condition -> "AnnotatedNoRouteToHostException".equals(condition.getReason()))) {
```
ppatierno, line 1367: Tbh I suggested to use the base class `NoRouteToHostException` as we catch the `ConnectException` and not the derived `AnnotatedConnectException`. They are part of the Java base and not Netty. @ShubhamRwt @im-konge can you try with using `NoRouteToHostException` if it works for you?

ppatierno (changes requested):

im-konge (approved):

kyguy (approved):

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/resource/cruisecontrol/CruiseControlApiImpl.java
```diff
@@ -435,6 +436,10 @@ private void httpExceptionHandler(Promise<? extends CruiseControlResponse> resul
             // Vert.x throws a NoStackTraceTimeoutException (inherits from TimeoutException) when the request times out
             // so we catch and raise a TimeoutException instead
             result.fail(new TimeoutException(t.getMessage()));
+        } else if (t instanceof NoRouteToHostException) {
+            // Netty throws a AnnotatedNoRouteToHostException when the request times out
```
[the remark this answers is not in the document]
ShubhamRwt: Fixed @ppatierno. Sorry for not pushing it early

ppatierno (approved):
LGTM

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceAssemblyOperator.java
```diff
@@ -1362,7 +1362,9 @@ private boolean hasRebalanceAnnotation(KafkaRebalance kafkaRebalance) {
 
     private boolean shouldRenewRebalance(KafkaRebalance kafkaRebalance, KafkaRebalanceAnnotation rebalanceAnnotation) {
         // check if refresh annotation applied or request are made when CC is not active
-        if (rebalanceAnnotation == KafkaRebalanceAnnotation.refresh || kafkaRebalance.getStatus().getConditions().stream().anyMatch(condition -> "ConnectException".equals(condition.getReason()))) {
+        if (rebalanceAnnotation == KafkaRebalanceAnnotation.refresh
+                || kafkaRebalance.getStatus().getConditions().stream().anyMatch(condition -> "ConnectException".equals(condition.getReason()))
+                || kafkaRebalance.getStatus().getConditions().stream().anyMatch(condition -> "NoRouteToHostException".equals(condition.getReason()))) {
```
scholzj, line 1367: It is probably worth adding some comment explaining this. Why does it look for this in the conditions? Also, why does it not use the actual exception classes?

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/resource/cruisecontrol/CruiseControlApiImpl.java
```diff
@@ -435,6 +436,10 @@ private void httpExceptionHandler(Promise<? extends CruiseControlResponse> resul
             // Vert.x throws a NoStackTraceTimeoutException (inherits from TimeoutException) when the request times out
             // so we catch and raise a TimeoutException instead
             result.fail(new TimeoutException(t.getMessage()));
+        } else if (t instanceof NoRouteToHostException) {
+            // Netty throws a AnnotatedNoRouteToHostException when it cannot resolve the host
```
scholzj, line 440: You talk about `AnnotatedNoRouteToHostException` buit you do not seem to use it. So ids the code wrong? Or the comment wrong?

scholzj (commented):
Should there be some unit test for this?

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/resource/cruisecontrol/CruiseControlApiImpl.java
ShubhamRwt: Hi Jakub, Actually we do get a `AnnotatedNoRouteToHostException` but we catch `NoRouteToHostException` instead since `AnnotatedNoRouteToHostException` (inherits from NoRouteToHostException)

scholzj: So the comment is wrong?

ShubhamRwt:
Added the tests @scholzj . I refactored the `createKafkaRebalance` method in `KafkaRebalanceStateMachineTest` class so that I can add Exception to the status and see the state change

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/resource/cruisecontrol/CruiseControlApiImpl.java
```diff
@@ -435,6 +436,10 @@ private void httpExceptionHandler(Promise<? extends CruiseControlResponse> resul
             // Vert.x throws a NoStackTraceTimeoutException (inherits from TimeoutException) when the request times out
             // so we catch and raise a TimeoutException instead
             result.fail(new TimeoutException(t.getMessage()));
+        } else if (t instanceof NoRouteToHostException) {
+            // Netty throws a AnnotatedNoRouteToHostException (inherits from NoRouteToHostException) when it cannot resolve the host
+            //so we catch and raise a NoRouteToHostException instead
```
scholzj, line 441: I still don't get from this why are you not checking for `t instanceof AnnotatedNoRouteToHostException` => that is what the comment needs to explain. Because you say you do something else than what you actually do.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceAssemblyOperator.java
scholzj: As commented before, this might deserve an explanation why this checks for the errors in conditions, because that seems wierd.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/resource/cruisecontrol/CruiseControlApiImpl.java
ShubhamRwt: We are catching `NoRouteToHostException` just like we catch the ConnectException and not the derived AnnotatedConnectException. I guess in the comments I wrote the same reason that it catch and raise a `NoRouteToHostException`  instead.

scholzj: But the comment does not really explain it. This seems weird:
* Why do you need to raise another exception when it inherits?
* If you need to raise the `NoRouteToHostException` instead, why do you check for `instanceof NoRouteToHostException` and not `instanceof AnnotatedNoRouteToHostException`

None of that is explained there. The comment just says what it does, but it needs to say why it does it.

ShubhamRwt: Actually if you see the other if conditions there also we are doing the same thing. We try to catch the `Parent Exception`. We are doing this because `ConnectException`, `NoRouteToHostException` all these exceptions are basically part of of Java base.

scholzj: I'm not sure that is a justification - that some other place does the same for sure not. And if an exception from some undesired transitive dependency is raised, it is probably a bug somewhere. But in any case, it is not what the comment says. So even if the code is kept as it is, the comment should be updated.

ShubhamRwt: Yeah I agree with you. I guess its better to catch `AnnotatedNoRouteHostException` itself and then raise a `NoRouteToHostException`. I was just trying to be consistent. I will fix the comment here

scholzj: I think you should find out who wrote the original code and have them explain why they did what they did.

ppatierno: @tomncooper do you remember why we went this way? Was it just for handling the Java base exceptions up the chain instead of handling the Netty related ones?

tomncooper: I do not remember why this was done? Git blame shows I wrote and Paolo was the last to change it. It doesn't seem to make sense to rethrow the parent Exception?

ppatierno: Maybe it was done for making the code which catches that exception more general and not relying on some specific exception coming from Netty but the base one coming from Java?

tomncooper: Well if you want to handle this you could still catch the more generic exception? I tried deleting this code and running the tests and nothing bad occured (except one test that exects the connection exceception not the netty one to be thrown). Maybe the system tests would show more, but this does seem to be a pointless operation?

ShubhamRwt: Should I catch the exception which are thrown itself and not the parent Exceptions then @ppatierno @tomncooper . Also I will test if it works fine and if it does then maybe we can keep it that way?

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceAssemblyOperator.java
tomncooper: I agree a better description is needed an that this checking could be made more robust. 

@scholzj what would be a better way to check for these types of exceptions?

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/resource/cruisecontrol/CruiseControlApiImpl.java
tomncooper: +1 for not doing this re-throwing of the parent. I tested the change and nothing broke so I am for removing this.

The only reason for doing this I can see is not exposing the dependency exception types in the API but they inherit from base Java exceptions so this is nor eeally an issue as far as I am concerned.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceAssemblyOperator.java
scholzj: @tomncooper Well, I do not know why it does what it does -> so it is hard to say what is a _better_ to do it.

But from my PoV ... the controller / operator pattern is based on retrying things in every reconciliation. So why does it need to look into the conditions what the error is? Shouldn't it just know what the error is because it just got the exception from some operation?

scholzj: As I said ... if this is the best way how to deal with something, fine, but it seems unusual at best, so I think it should be explained in the comment. And I also think that this `why does it not use the actual exception classes` is a valid comment even if the logic stays.

scholzj:
@ShubhamRwt Can you please fix the build and the DCO signoff?

ShubhamRwt:
Sure. Actually we need some discussion on this PR and yesterday Tom and I decided to have a meeting today regarding this

ShubhamRwt:
I have made the new changes @ppatierno @tomncooper

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/resource/cruisecontrol/CruiseControlApiImpl.java
```diff
@@ -435,11 +436,12 @@ private void httpExceptionHandler(Promise<? extends CruiseControlResponse> resul
...
             // so we catch and raise a TimeoutException instead
             result.fail(new TimeoutException(t.getMessage()));
-        } else if (t instanceof ConnectException) {
+        } else if (t instanceof NoRouteToHostException || t instanceof ConnectException) {
+            // Netty throws a AnnotatedNoRouteToHostException (inherits from NoRouteToHostException) and  when it cannot resolve the host
             // Vert.x throws a AnnotatedConnectException (inherits from ConnectException) when the request times out
-            // so we catch and raise a ConnectException instead
-            result.fail(new ConnectException(t.getMessage()));
-        } else {
+            // so we catch and raise a CruiseControlRestException instead since
```
ppatierno, line 442: raise a CruiseControlRetriableConnectException not CruiseControlRestException

cluster-operator/src/test/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceStateMachineTest.java
```diff
@@ -96,12 +97,13 @@ public void resetServer() {
...
      * @param userAnnotation An annotation to be applied after the reconcile has started, for example "approve" or "stop".
      * @param rebalanceSpec A custom rebalance specification. If null a blank spec will be used.
+     * @param exceptionString
      * @return A KafkaRebalance instance configured with the supplied parameters.
      */
     private KafkaRebalance createKafkaRebalance(KafkaRebalanceState currentState,
                                                 String currentStatusSessionID,
                                                 String userAnnotation,
-                                                KafkaRebalanceSpec rebalanceSpec) {
+                                                KafkaRebalanceSpec rebalanceSpec, String exceptionString) {
```
ppatierno, line 106: I see it sets the reason on the condition, why not calling it `reason` instead of `exceptionString`?

ppatierno (changes requested):

cluster-operator/src/test/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceStateMachineTest.java
```diff
@@ -97,11 +98,12 @@ public void resetServer() {
      * @param userAnnotation An annotation to be applied after the reconcile has started, for example "approve" or "stop".
      * @param rebalanceSpec A custom rebalance specification. If null a blank spec will be used.
      * @return A KafkaRebalance instance configured with the supplied parameters.
+     * @return Reason for the condition
```
ppatierno, line 101: I think it's wrong. You have got two `@return` I guess you meant to add a `@param reason`

ShubhamRwt: Oops Sorry, Let me correct it

ShubhamRwt: Fixed

ppatierno (approved):
Just a final nit, but LGTM overall to me.

kyguy (approved):

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/resource/cruisecontrol/CruiseControlRetriableConnectException.java
```diff
@@ -0,0 +1,12 @@
+/*
+ * Copyright Strimzi authors.
+ * License: Apache License 2.0 (see the file LICENSE or http://apache.org/licenses/LICENSE-2.0.html).
+ */
+package io.strimzi.operator.cluster.operator.resource.cruisecontrol;
+
+public class CruiseControlRetriableConnectException extends RuntimeException {
```
tomncooper, line 7: I wonder if a better name for this would have been `CruiseControlRetriableConnectionException` to be explicit and avoid any possible confusion with Kafka Connect.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/resource/cruisecontrol/CruiseControlApiImpl.java
```diff
@@ -435,11 +436,12 @@ private void httpExceptionHandler(Promise<? extends CruiseControlResponse> resul
...
             result.fail(new TimeoutException(t.getMessage()));
-        } else if (t instanceof ConnectException) {
+        } else if (t instanceof NoRouteToHostException || t instanceof ConnectException) {
+            // Netty throws a AnnotatedNoRouteToHostException (inherits from NoRouteToHostException) and  when it cannot resolve the host
             // Vert.x throws a AnnotatedConnectException (inherits from ConnectException) when the request times out
-            // so we catch and raise a ConnectException instead
-            result.fail(new ConnectException(t.getMessage()));
-        } else {
+            // so we catch and raise a CruiseControlRetriableConnectException instead since
+            result.fail(new CruiseControlRetriableConnectException(t.getMessage()));
```
tomncooper, line 443: So I think this is the right approach. However, now the KRAO is expecting **any** implementation of the `CruiseControlApi` interface to fail a future, if it is a transient connection error, with a `CruiseControlRetriableConnectException` from many/all its methods. 

Ideally, we would put this exception handler in an an abstract class so other implementation could use it, but maybe it would be acceptable to simply update the documentation of the interface class to explicitly say that any transient connection error should be failed with a `CruiseControlRetriableConnectException`?

@ppatierno WDYT?

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/resource/cruisecontrol/CruiseControlRetriableConnectException.java
ppatierno: Don't know ... I kind of like the relationship with the `ConnectException` we catch in the code to make this.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/resource/cruisecontrol/CruiseControlApiImpl.java
ppatierno: >  other implementation could use it

Which "other implementation" are you referring to?

scholzj, line 443: Should this inherit the exception instead of copying only the message? That way, it should keep the whole stack etc. and I do not see what problems would it cause.

```suggestion
            result.fail(new CruiseControlRetriableConnectException(t));
```

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/resource/cruisecontrol/CruiseControlRetriableConnectException.java
scholzj: I would agree with @tomncooper here. The `CruiseControlRetriableConnectException` seems weird to me as well. Double confusing with Kafka Connect being around.

scholzj: Also, it is thrown for `NoRouteToHostException` as well, not only for `ConnectException` which might add to another confusion.

im-konge (approved):

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/resource/cruisecontrol/CruiseControlApiImpl.java
ShubhamRwt: I tried doing this but I was getting some issues since the message which is being set in the condition seems to come out as `null` if I do that so probably up in the stack the message we are passing is being used

ShubhamRwt:
@tomncooper @ppatierno Does this PR need more changes?

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/resource/cruisecontrol/CruiseControlApiImpl.java
scholzj: But reusing the exception is important since that gives you the stack trace etc. I do not understand how the message being null relates to it.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceAssemblyOperator.java
```diff
@@ -1362,7 +1362,10 @@ private boolean hasRebalanceAnnotation(KafkaRebalance kafkaRebalance) {
 
     private boolean shouldRenewRebalance(KafkaRebalance kafkaRebalance, KafkaRebalanceAnnotation rebalanceAnnotation) {
         // check if refresh annotation applied or request are made when CC is not active
-        if (rebalanceAnnotation == KafkaRebalanceAnnotation.refresh || kafkaRebalance.getStatus().getConditions().stream().anyMatch(condition -> "ConnectException".equals(condition.getReason()))) {
+        // We check for `CruiseControlRetriableConnectException` if request are made when CC was not active since not handling this case will move the KafkaRebalance to `NotReady`
+        // and it will stay in `NotReady` state until we apply `refresh`
+        if (rebalanceAnnotation == KafkaRebalanceAnnotation.refresh
+                || kafkaRebalance.getStatus().getConditions().stream().anyMatch(condition -> "CruiseControlRetriableConnectException".equals(condition.getReason()))) {
```
ppatierno, line 1368: should it be "CruiseControlRetriableConnectionException" now? Even in the comment above?

cluster-operator/src/test/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceStateMachineTest.java
```diff
@@ -60,6 +60,7 @@ public class KafkaRebalanceStateMachineTest {
     private static final String RESOURCE_NAME = "my-rebalance";
     private static final String CLUSTER_NAMESPACE = "cruise-control-namespace";
     private static final String CLUSTER_NAME = "kafka-cruise-control-test-cluster";
+    private static final String REASON = "CruiseControlRetriableConnectException";
```
ppatierno, line 63: Using "CruiseControlRetriableConnectionException" as reason here as well?

ppatierno (changes requested):

ppatierno (changes requested):

cluster-operator/src/test/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceStateMachineTest.java
```diff
@@ -873,6 +877,43 @@ private void testStateWithWrongAnnotation(Vertx vertx, VertxTestContext context,
...
+
+        kcRebalance = createKafkaRebalance(KafkaRebalanceState.Ready, null, null, ADD_BROKER_KAFKA_REBALANCE_SPEC, REASON);
+        this.testStateWithException(vertx, context, 0, CruiseControlEndpoints.ADD_BROKER, kcRebalance, KafkaRebalanceState.NotReady, KafkaRebalanceAnnotation.none);
+
+        kcRebalance = createKafkaRebalance(KafkaRebalanceState.Ready, null, null, REMOVE_BROKER_KAFKA_REBALANCE_SPEC, REASON);
+        this.testStateWithException(vertx, context, 0, CruiseControlEndpoints.REMOVE_BROKER, kcRebalance, KafkaRebalanceState.NotReady, KafkaRebalanceAnnotation.none);
+    }
+
+
+    private void testStateWithException(Vertx vertx, VertxTestContext context,
```
ppatierno, line 905: this methos seems to be exactly the same as `testStateWithWrongAnnotation` but just with a different name. How is it testing that the connection exception is recovered?

cluster-operator/src/test/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceStateMachineTest.java
```diff
@@ -873,6 +877,43 @@ private void testStateWithWrongAnnotation(Vertx vertx, VertxTestContext context,
...
+
+        kcRebalance = createKafkaRebalance(KafkaRebalanceState.Ready, null, null, ADD_BROKER_KAFKA_REBALANCE_SPEC, REASON);
+        this.testStateWithException(vertx, context, 0, CruiseControlEndpoints.ADD_BROKER, kcRebalance, KafkaRebalanceState.NotReady, KafkaRebalanceAnnotation.none);
+
+        kcRebalance = createKafkaRebalance(KafkaRebalanceState.Ready, null, null, REMOVE_BROKER_KAFKA_REBALANCE_SPEC, REASON);
+        this.testStateWithException(vertx, context, 0, CruiseControlEndpoints.REMOVE_BROKER, kcRebalance, KafkaRebalanceState.NotReady, KafkaRebalanceAnnotation.none);
+    }
+
+    @Test
+    public void testReadyAfterConnectException(Vertx vertx, VertxTestContext context) throws IOException, URISyntaxException {
```
ppatierno, line 893: Let's use "ConnectionException" in the method name for consistency

ppatierno (changes requested):

cluster-operator/src/test/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceStateMachineTest.java
ShubhamRwt: Should we name it 'testReadywithCruiseControlRetriableException' itself? Since we are addressing two errors under the same exception?

cluster-operator/src/test/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceStateMachineTest.java
ShubhamRwt: The `testStateWithException` method is always taking the current state as `ProposalReady` and checking the transition from `NotReady` to `ProposalReady` while for the `testStatewithWrongAnnotations` we have to check for multiple next states since its being used by multiple other tests

ShubhamRwt: @ppatierno I have removed the `test` which I was adding named `testStateWithNoRouteToHostEception` since now `CruiseControlRetriableConnectionException` is being thrown in both cases(ConnectExcpetion && testStateWithNoRouteToHostEception) so I guess only one test is enough

ppatierno (approved):
LGTM overall.

scholzj (approved):

PR #6900, opened 2022-06-07
Fix the byte-wise sorting on Linux/Fedora

### LINKED ISSUE #6899: The `sort` used for checksum processing isn't byte-wise on Linux/Fedora OS
After changing the files in the `install` folder in relation to the #6897 , I had to update the `.checksums` file here: https://github.com/strimzi/strimzi-kafka-operator/blob/main/.checksums

Running the `make checksum_install` on my Fedora laptop seems to return a different value (for the SHA) compared to what is in the CI pipeline and on a MacOS (I asked @scholzj to do a test for me).

After some investigation, the problem seems to be in the `sort` command and more specifically the locale settings (depending on the OS).
Without processing the last SHAs, but just listing the SHAs for all the files in the folder, Jakub seems to have an output where the order is something like this:

```shell
358647c16a544b973eea06e436b7afd948c9d812  ./install/DO_NOT_EDIT.md
bd0c95bcc3cbd7b06a4ecdcec222392f9af25334  ./install/canary/000-ServiceAccount.yaml
....
....
```

while in my case

```shell
...
...
fbb47c316438fbf34584c14626da279d5582c0f1  ./install/cluster-operator/060-Deployment-strimzi-cluster-operator.yaml
358647c16a544b973eea06e436b7afd948c9d812  ./install/DO_NOT_EDIT.md
...
```

The `D` of the `DO_NOT_EDIT` is handled as coming before the `c` of the `canary` folder in Jakub case and not for me.
It seems to be related to how the sort is behaving and doing the comparison. The `D` ASCII code comes before the `c` ASCII code, so the sort command on MacOS is byte-wise taking into account ASCII code.
To have same result I needed to set `LC_ALL=C` to force the sorting to be byte-wise. Only in this way I get same ordering result.
So the command would be something like:

```bash
checksum_install:
	@$(FIND) ./install/ -type f -print0 | LC_ALL=C $(SORT) -z | $(XARGS) -0 $(SHA1SUM) | $(SHA1SUM)
```

Trying running this and then `make release_files_check` seems to be successful.
I would like to open a PR to fix this, @scholzj would you mind to try this change locally?

ppatierno:
This PR fixes #6899 by forcing the byte-wise (based on ASCII code) sorting when processing the checksums on install and helm related folders.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -200,12 +200,12 @@ prerequisites_check:
 	SED=$(SED) ./tools/prerequisites-check.sh
 
 checksum_examples:
-	@$(FIND) ./examples/ -type f -print0 | $(SORT) -z | $(XARGS) -0 $(SHA1SUM) | $(SHA1SUM)
+	@$(FIND) ./examples/ -type f -print0 | LC_ALL=C $(SORT) -z | $(XARGS) -0 $(SHA1SUM) | $(SHA1SUM)
 
 checksum_install:
-	@$(FIND) ./install/ -type f -print0 | $(SORT) -z | $(XARGS) -0 $(SHA1SUM) | $(SHA1SUM)
+	@$(FIND) ./install/ -type f -print0 | LC_ALL=C $(SORT) -z | $(XARGS) -0 $(SHA1SUM) | $(SHA1SUM)
 
 checksum_helm:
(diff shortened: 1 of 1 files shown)


scholzj (approved):
LGTM assuming it works also on the CI.

PR #6901, opened 2022-06-07
docs: adds rack awareness callout to the mm2 config example

PaulRMellor:

**Documentation**
Adds rack awareness config to the [Configuring Kafka MirrorMaker 2.0](https://strimzi.io/docs/operators/in-development/configuring.html#proc-mirrormaker-replication-str) example.  Callout links to the [Rack schema reference](https://strimzi.io/docs/operators/in-development/configuring.html#type-Rack-reference) as with [Configuring Kafka Connect](https://strimzi.io/docs/operators/in-development/configuring.html#proc-kafka-connect-config-str)

### Checklist

- [ ] Write tests
- [x] Update documentation
- [ ] Update CHANGELOG.md


documentation/modules/configuring/proc-config-mirrormaker2-replication.adoc
```diff
@@ -234,10 +236,11 @@ To configure topic offset synchronization, this property must also be set for th
 <39> xref:con-common-configuration-healthchecks-reference[Healthchecks] to know when to restart a container (liveness) and when a container can accept traffic (readiness).
 <40> xref:con-common-configuration-jvm-reference[JVM configuration options] to optimize performance for the Virtual Machine (VM) running Kafka MirrorMaker.
 <41> ADVANCED OPTION: xref:con-common-configuration-images-reference[Container image configuration], which is recommended only in special situations.
-<42> xref:assembly-customizing-kubernetes-resources-str[Template customization]. Here a pod is scheduled with anti-affinity, so the pod is not scheduled on nodes with the same hostname.
-<43> Environment variables are also xref:ref-tracing-environment-variables-str[set for distributed tracing using Jaeger].
-<44> xref:assembly-distributed-tracing-str[Distributed tracing is enabled for Jaeger].
-<45> xref:type-ExternalConfiguration-reference[External configuration] for a Kubernetes Secret mounted to Kafka MirrorMaker as an environment variable.
+<42> xref:type-Rack-reference[Rack awareness] is configured to spread replicas across different racks. A `topologykey` must match the label of a cluster node.
```
fvaleri, line 239: I would say: "across different racks or availability zones" and "must match a node label containing the rack id, like the well-known [topology.kubernetes.io/zone](https://kubernetes.io/docs/reference/labels-annotations-taints/#topologykubernetesiozone)."

scholzj (commented):
My view ... if you want to make it more visible like this, you should add a proper explanation when it is useful given how niche feature this is. Otherwise it might easily confuse users into thinking that it is useful while in most cases it will not do much good.

PaulRMellor:
@scholzj 

> My view ... if you want to make it more visible like this, you should add a proper explanation when it is useful given how niche feature this is. Otherwise it might easily confuse users into thinking that it is useful while in most cases it will not do much good.

We have this callout for Kafka, Kafka Connect and MirrorMaker 2.0. I've updated as follows. Hopefully, this makes it clear that it's for use in specific circumstances.

> RACK OPTION: [Rack awareness](file:///home/pmellor/Documents/git/strimzi-kafka-operator/documentation/configuring/build/tmp/en-US/html-single/index.html#type-Rack-reference) configuration for the deployment. Use this option only if you need to spread replicas across different racks, data centers, or availability zones. The `topologyKey` must match a node label containing the rack ID. This configuration specifies a zone using the standard `topology.kubernetes.io/zone` label.

fvaleri:
LGTM

scholzj:
Its not about the example. It is about when it makes sense to use this feature and when not.

documentation/modules/configuring/proc-config-kafka-connect.adoc
```diff
@@ -175,7 +175,7 @@ You can also use _configuration provider plugins_ to xref:assembly-loading-confi
...
 <16> xref:con-common-configuration-jvm-reference[JVM configuration options] to optimize performance for the Virtual Machine (VM) running Kafka Connect.
 <17> ADVANCED OPTION: xref:con-common-configuration-images-reference[Container image configuration], which is recommended only in special situations.
-<18> RACK OPTION: xref:type-Rack-reference[Rack awareness] configuration for the deployment. Use this option only if you need to spread replicas across different racks, data centers, or availability zones. The `topologyKey` must match a node label containing the rack ID. This configuration specifies a zone using the standard `topology.kubernetes.io/zone` label.  
+<18> SPECIALIZED OPTION: xref:type-Rack-reference[Rack awareness] configuration for the deployment. This is a specialized option intended for a local deployment. Use this option if you want connectors to consume from the closest replica rather than the leader replica. If suitable, rack awareness can improve network utilization. The `topologyKey` must match a node label containing the rack ID. The example used in this configuration specifies a zone using the standard `topology.kubernetes.io/zone` label.
```
scholzj, line 178: I think we need to explain what local deployment means, as that will be not clear to people.

```suggestion
<18> SPECIALIZED OPTION: xref:type-Rack-reference[Rack awareness] configuration for the deployment. This is a specialized option intended for a local deployment. Use this option if you want connectors to consume from the closest replica rather than the leader replica. In some case, consuming from closest replica can improve network utilization or save costs. The `topologyKey` must match a node label containing the rack ID. The example used in this configuration specifies a zone using the standard `topology.kubernetes.io/zone` label. To consume from the closest replica, you need to also enable the `RackAwareReplicaSelector` in the Kafka broker configuration.
```

documentation/modules/configuring/proc-config-mirrormaker2-replication.adoc
```diff
@@ -236,7 +236,7 @@ To configure topic offset synchronization, this property must also be set for th
...
 <40> xref:con-common-configuration-jvm-reference[JVM configuration options] to optimize performance for the Virtual Machine (VM) running Kafka MirrorMaker.
 <41> ADVANCED OPTION: xref:con-common-configuration-images-reference[Container image configuration], which is recommended only in special situations.
-<42> RACK OPTION: xref:type-Rack-reference[Rack awareness] configuration for the deployment. Use this option only if you need to spread replicas across different racks, data centers, or availability zones. The `topologyKey` must match a node label containing the rack ID. This configuration specifies a zone using the standard `topology.kubernetes.io/zone` label.
+<42> SPECIALIZED OPTION: xref:type-Rack-reference[Rack awareness] configuration for the deployment. This is a specialized option intended for a local deployment. Use this option if you want connectors to consume from the closest replica rather than the leader replica. If suitable, rack awareness can improve network utilization. The `topologyKey` must match a node label containing the rack ID. The example used in this configuration specifies a zone using the standard `topology.kubernetes.io/zone` label.
```
scholzj, line 239: Same as above for Connect.

documentation/modules/configuring/proc-config-kafka-connect.adoc
fvaleri: If I get this right, local deployment basically means in the same racks/DCs/AZs set or subset of the source Kafka cluster. With rack awareness enabled, a connector task would be able to consume messages from a partition replica hosted in the same rack where it is running (broker.rack==client.rack), thus avoiding inter rack traffic, which is slower and may be charged. If such replica is not available, then it would fall back and consume from the leader replica, wherever it is hosted.

scholzj: That is the general idea. The _local deployment_ is important because:
* If you connect to some remote Kafka cluster, you are obviously loosing the advantage from avoiding the inter-rack traffic. Since the inter-region traffic is probably a bigger bottleneck but also doesn't create the intra-rack traffic in the first place.
* In many public clouds, the zones are randomized. So `us-east-1a` in my AWS account is not the same data centre as `us-east-1a` in your AWS account. So forcing Kafka to consume from some Kafka broker in the same zone in another AWS account means nothing.

At the same time, consuming from closest replica has drawbacks such as increased latency etc. So you do not want to use it _just in case it maybe helped_.

What exactly local deployment means is complicated and will depend on the infrastructure. It does not necessarily mean _in the same Kubernetes cluster_. But it will be a little bit individual for each user.

fvaleri: Thanks @scholzj for the additional insights.

scholzj (approved):

Issue #6854, opened 2022-05-23
The operator restarts periodically


mpuch12:
**Describe the bug**
After deploy cluster operator restart periodically (typically after 15-20 minutes), without any error in logs 

**Environment:**
 - Strimzi version: 0.29.0
 - Installation method: Helm chart
 - Kubernetes cluster: Kubernetes 1.22.2
 - Infrastructure: Kubadm
 - Kafka version: 3.1.0

**YAML files and logs**

Deploy using:
`helm upgrade 
  kafka-operator strimzi/strimzi-kafka-operator 
  --namespace kafka 
  --version 0.29.0 
  --install 
  --create-namespace 
  --wait 
  --timeout 300s 
  --set resources.requests.cpu=50m`

Operator last logs:
```
[...]

PR #6915, opened 2022-06-08
Update launch_java.sh to support cgroups v1 and v2

joseacl:
In order to avoid the problem spotted here https://github.com/strimzi/strimzi-kafka-operator/issues/6854#issuecomment-1149989213

Basically, we've calculated the value of the property -Xmx based on the limits in the container (compatible with cgroups v1 and v2)

### Type of change

_Select the type of your PR_

- Enhancement / new feature

### Description

Dynamic calculation of -Xmx value if the user does not specify it. 
This calculation is based on the limits set in the container and it is compatible with cgroups v1 and v2 versions.

### Checklist

- [ ] Write tests
- [ ] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/documentation/modules/cruise-control/proc-approving-optimization-proposal.adoc
+++ b/documentation/modules/cruise-control/proc-approving-optimization-proposal.adoc
@@ -25,58 +25,62 @@ Cruise Control will then apply the optimization proposal to the Kafka cluster, r
 
 .Procedure
 
-Perform these steps for the optimization proposal that you want to approve:
+Perform these steps for the optimization proposal that you want to approve.
 
 . Unless the optimization proposal is newly generated, check that it is based on current information about the state of the Kafka cluster.
 To do so, refresh the optimization proposal to make sure it uses the latest cluster metrics:
 
-.. Annotate the `KafkaRebalance` resource in Kubernetes with `refresh`:
+.. Annotate the `KafkaRebalance` resource in Kubernetes with `strimzi.io/rebalance=refresh`:
 +
 [source,shell,subs="+quotes"]
 ----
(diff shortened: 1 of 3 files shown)


scholzj (commented):
Thanks for the PR. I think that if nothing else, you have to make oit backwards ccompatible and make sure it accepts the user-passed options. It would be good to also add more comments explaining what are you doing in why.

scholzj (commented):
You will also need to fix the DCO sign-off -> the instructions are under the _Details_ link next to the DCO status.

ivansjg:
> Thanks for the PR. I think that if nothing else, you have to make oit backwards ccompatible and make sure it accepts the user-passed options. It would be good to also add more comments explaining what are you doing in why.

Basically, if your kubernetes cluster is based in cgroups v2, and JVM version is lower than version 15, "XMaxRAMPercentage" is not gonna work and JVM running into the container will try to take all memory in the k8s node. That will lead to k8s kill the pod because is trying to take more memory than resources provided, which leads into the pod is continuous restarting.
So, this commit, afaik, is calculating maximum RAM for heap based on cgroups files (supporting cgroups v1 and v2).

scholzj:
I get that, although this generic description is something you can add to the PR description. But the process you have there seems fairly complicated. So describibg the steps with comments in the code what does it do and why is IMHO useful. This also works for many users, so it is important to not change the behaviour for them. You can for example use your calculation only when the cgroups v2 are used or when no other Java options are set.

joseacl:
Hi @scholzj !
Sorry, but we had a problem trying with the sign-off of commits from this PR and we decided to create a new one which the suggestions commented, you can check here:
https://github.com/strimzi/strimzi-kafka-operator/pull/6920

Thanks and sorry for the inconvenience!

PR #6917, opened 2022-06-09
[systemtest] Use different pod than Kafka for executing all Kafka scripts

### LINKED ISSUE #6689: [systemtest] Remove all commands running in Kafka pods from STs
**Current situation**

Currently, in our STs, we are executing some commands from Kafka pods - mainly in `KafkaCmdClient` and other.
This is not right, we should execute commands from different pod - such as the `scraper` pod - to prevent issue with compatibility (the scripts can be changed during versions) etc.

It would be great to use the `scraper` pod for most of the commands - like API requests - and for other operations, for which is Admin API needed, use the Admin client image and create the pod - from there create/update/list/delete topics or users.

This will need refactor of the Admin client inside the [test-clients repository](https://github.com/strimzi/test-clients), as the actual implementation doesn't allow us to use the image for nothing else than just `Job`.

im-konge:

### Type of change

- Refactoring

### Description

How was already multiple times mentioned, we should use different pod than Kafka/ZK pod for executing the Kafka scripts. In #6839 the `Scraper` deployment was edited to use our Kafka image, so the pod will contain all needed scripts, which we can use in our tests.

In this PR I'm refactoring our utils classes and tests to use `Scraper` pod for all Kafka scripts execution and also bringing all the commands from utils classes to one place - `KafkaCmdClient`.

Closes #6689 

### Checklist

### DIFF
--- a/systemtest/src/main/java/io/strimzi/systemtest/cli/KafkaCmdClient.java
+++ b/systemtest/src/main/java/io/strimzi/systemtest/cli/KafkaCmdClient.java
@@ -4,35 +4,27 @@
  */
 package io.strimzi.systemtest.cli;
 
-import io.strimzi.api.kafka.model.KafkaResources;
 import io.strimzi.systemtest.utils.kafkaUtils.KafkaTopicUtils;
 import org.apache.logging.log4j.Level;
 
 import java.util.Arrays;
 import java.util.List;
+import java.util.Map;
+import java.util.stream.Collectors;
+import java.util.stream.IntStream;
 
-import static io.strimzi.systemtest.resources.ResourceManager.kubeClient;
 import static io.strimzi.test.k8s.KubeClusterResource.cmdKubeClient;
 
 public class KafkaCmdClient {
 
-    private static final int PORT = 9092;
-
-
-    public static List<String> listTopicsUsingPodCli(String namespaceName, String clusterName, int kafkaPodId) {
-        String podName = KafkaResources.kafkaPodName(clusterName, kafkaPodId);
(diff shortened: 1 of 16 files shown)


systemtest/src/main/java/io/strimzi/systemtest/cli/KafkaCmdClient.java
```diff
@@ -4,84 +4,112 @@
...
-
-    public static List<String> listTopicsUsingPodCli(String namespaceName, String clusterName, int kafkaPodId) {
-        String podName = KafkaResources.kafkaPodName(clusterName, kafkaPodId);
-        return Arrays.asList(cmdKubeClient(namespaceName).execInPod(podName, "/bin/bash", "-c",
-            "bin/kafka-topics.sh --list --bootstrap-server localhost:" + PORT).out().split("\\s+"));
-    }
-
-    public static List<String> listTopicsUsingPodCli(String clusterName, int kafkaPodId) {
-        return listTopicsUsingPodCli(kubeClient().getNamespace(), clusterName, kafkaPodId);
+    public static List<String> listTopicsUsingPodCli(String namespaceName, String scraperPodName, String bootstrapServer) {
```
Frawless, line 20: Do we want to use `scraperPodName` or something more generic?

Frawless (approved):

systemtest/src/main/java/io/strimzi/systemtest/cli/KafkaCmdClient.java
im-konge: More generic will be better. Thanks :)

PR #6920, opened 2022-06-09
Added support for cgroups v2

joseacl:
Signed-off-by: joseacl <joseacl@gmail.com>

### Type of change

_Select the type of your PR_

- Enhancement / new feature

### Description

This PR fixed the problem spotted here https://github.com/strimzi/strimzi-kafka-operator/issues/6854#issuecomment-1149989213.

Java versions older than 15 don't support cgroups v2 when trying to figure out the memory limits (using MinRAMPercentage, MaxRAMPercentage), so in this PR we've added the modifications needed to calculate the value of -Xmx based on cgroups v2 memory file.

### Checklist

- [ ] Write tests
- [ ] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/docker-images/operator/scripts/launch_java.sh
+++ b/docker-images/operator/scripts/launch_java.sh
@@ -27,13 +27,7 @@ calc_max_memory() {
   local mem_limit
   mem_limit="$(cat ${mem_file_cgroups_v2})"
 
-   if [ "${mem_limit}" -le 314572800 ]; then
-    # Restore the one-fourth default heap size instead of the one-half below 300MB threshold
-    # See https://docs.oracle.com/javase/8/docs/technotes/guides/vm/gctuning/parallel.html#default_heap_size
-    calc_maximum_size_opt "${mem_limit}" "50"
-  else
-    calc_maximum_size_opt "${mem_limit}" "75"
-  fi
+  calc_maximum_size_opt "${mem_limit}" "20"
 }
 
 export MALLOC_ARENA_MAX=2
@@ -51,6 +45,8 @@ JAVA_OPTS="${JAVA_OPTS} --illegal-access=deny"
 if [[ ! -r "${mem_file_cgroups_v2}" && "$JAVA_OPTS" != *"MinRAMPercentage"* && "$JAVA_OPTS" != *"MaxRAMPercentage"* && "$JAVA_OPTS" != *"InitialRAMPercentage"* ]]; then
   JAVA_OPTS="${JAVA_OPTS} -XX:MinRAMPercentage=10 -XX:MaxRAMPercentage=20 -XX:InitialRAMPercentage=10"
 elif [[ -r "${mem_file_cgroups_v2}" && "$JAVA_OPTS" != *"Xmx"* ]]; then
+  # This workaround for cgroups v2 must be removed once java version used by the operator is updated.
+  # https://developers.redhat.com/articles/2022/04/19/java-17-whats-new-openjdks-container-awareness#
   # Calculate -Xmx java option
   JAVA_OPTS="${JAVA_OPTS} $(calc_max_memory)"
 fi


docker-images/operator/scripts/launch_java.sh
```diff
@@ -12,6 +14,70 @@ function get_gc_opts {
...
+    return
+  fi
+
+  if [ "${CONTAINER_MAX_MEMORY}" -le 314572800 ]; then
+    # Restore the one-fourth default heap size instead of the one-half below 300MB threshold
+    # See https://docs.oracle.com/javase/8/docs/technotes/guides/vm/gctuning/parallel.html#default_heap_size
+    calc_mem_opt "${CONTAINER_MAX_MEMORY}" "50" "mx"
+  else
+    calc_mem_opt "${CONTAINER_MAX_MEMORY}" "75" "mx"
+  fi
```
scholzj, line 78: Why these numbers? There is a reason why we set the MaxRamPercentage to 20% => because it does not work well with higher numbers. So shouldn't this follow the same logic?

joseacl: Taking a look at the deployment that used this code,

https://github.com/strimzi/strimzi-kafka-operator/blob/main/helm-charts/helm3/strimzi-kafka-operator/templates/060-Deployment-strimzi-cluster-operator.yaml#L50

There is only one container in the pod deployed by as part of that deployment. So we don't see the need to use only 20% of the memory available by the container for that java application (operator).

docker-images/operator/scripts/launch_java.sh
```diff
@@ -12,6 +14,70 @@ function get_gc_opts {
...
+}
+
+calc_mem_opt() {
+  local max_mem="$1"
+  local fraction="$2"
+  local mem_opt="$3"
+
+  local val
+  val=$(calc "round($1*$2/100/1048576)" "${max_mem}" "${fraction}")
+  echo "-X${mem_opt}${val}m"
```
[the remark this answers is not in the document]
joseacl: You're right! we'll modify it as you suggested. Thanks!

docker-images/operator/scripts/launch_java.sh
```diff
@@ -24,8 +90,13 @@ JAVA_OPTS="${JAVA_OPTS} $(get_gc_opts)"
...
 # Default memory options used when the user didn't configured any of these options, we set the defaults
-if [[ "$JAVA_OPTS" != *"MinRAMPercentage"* && "$JAVA_OPTS" != *"MaxRAMPercentage"* && "$JAVA_OPTS" != *"InitialRAMPercentage"* ]]; then
+if [[ ! -r "${mem_file_cgroups_v2}" && "$JAVA_OPTS" != *"MinRAMPercentage"* && "$JAVA_OPTS" != *"MaxRAMPercentage"* && "$JAVA_OPTS" != *"InitialRAMPercentage"* ]]; then
   JAVA_OPTS="${JAVA_OPTS} -XX:MinRAMPercentage=10 -XX:MaxRAMPercentage=20 -XX:InitialRAMPercentage=10"
+elif [[ -r "${mem_file_cgroups_v2}" && "$JAVA_OPTS" != *"Xmx"* ]]; then
+  # Set env vars reflecting memory limits
+  init_memory_limit_env_vars
+  # Calculate -Xmx java option
+  JAVA_OPTS="${JAVA_OPTS} $(calc_max_memory)"
 fi
```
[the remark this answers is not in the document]
joseacl: Perhaps I'm missing something, but the else code is only executed if the pod is deployed in a cluster using cgroups v2 (XX:MinRAMPercentage , XX:MinRAMPercentage, don't work in that case) and the user does not specify -Xmx.

Since we're going to simplify the process of getting that value, calc_max_memory should always return a value specified in /sys/fs/cgroup/memory.max

What do you think?

docker-images/operator/scripts/launch_java.sh
```diff
@@ -12,6 +14,70 @@ function get_gc_opts {
...
+  local max_mem_meminfo_kb
+  max_mem_meminfo_kb="$(< /proc/meminfo awk '/MemTotal/ {print $2}')"
+  local max_mem_meminfo
+  max_mem_meminfo=$((max_mem_meminfo_kb*1024))
+
+   max_mem_cgroup="$(cat ${mem_file_cgroups_v2})"
+
+  if [ "${max_mem_cgroup:-0}" != -1 ] && [ "${max_mem_cgroup:-0}" -lt ${max_mem_meminfo:-0} ]; then
+    echo "${max_mem_cgroup}"
+  fi
```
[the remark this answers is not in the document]
joseacl: The comparison I've mentioned in the comment below 
https://github.com/strimzi/strimzi-kafka-operator/pull/6920#discussion_r893784816]

docker-images/operator/scripts/launch_java.sh
```diff
@@ -12,6 +14,41 @@ function get_gc_opts {
...
+
+# Calculate the value of -Xmx options base on cgroups_v2 values
+calc_max_memory() {
+  local mem_limit
+  mem_limit="$(cat ${mem_file_cgroups_v2})"
+
+   if [ "${mem_limit}" -le 314572800 ]; then
+    # Restore the one-fourth default heap size instead of the one-half below 300MB threshold
+    # See https://docs.oracle.com/javase/8/docs/technotes/guides/vm/gctuning/parallel.html#default_heap_size
+    calc_maximum_size_opt "${mem_limit}" "50" "mx"
```
scholzj, line 46: I guess the `mx` parameter is not needed anymore.

docker-images/operator/scripts/launch_java.sh
```diff
@@ -12,6 +14,41 @@ function get_gc_opts {
...
+  else
+    calc_maximum_size_opt "${mem_limit}" "75" "mx"
```
scholzj, line 48: Same as above, I guess the mx parameter is not needed anymore.

docker-images/operator/scripts/launch_java.sh
scholzj: The 20% value is based on our experience (likely with cgroups v1). Why do you think with cgroups v2 the Java requirements will suddenly change and it will be able to use 50% or 75%?

docker-images/operator/scripts/launch_java.sh
```diff
@@ -12,6 +14,41 @@ function get_gc_opts {
   fi
 }
 
+# Generic formula evaluation based on awk
```
scholzj, line 17: I think the comment should explain what the formula calculates rather than stating how it calculates it.

docker-images/operator/scripts/launch_java.sh
```diff
@@ -12,6 +14,41 @@ function get_gc_opts {
...
+    function round(x) {
+      return int(x + 0.5)
+    }
+    {print '"int(${formula})"'}
+  '
+}
+
+calc_maximum_size_opt() {
+  local max_mem="$1"
+  local fraction="$2"
```
scholzj, line 31: Maybe you should call it percentage rather than fraction? As that is what you seem to pass in.

docker-images/operator/scripts/launch_java.sh
joseacl: More likely I'm missing something, but as I see it, It's not a matter of using cgroups v1 or v2, it's a matter of the point of view of the application inside k8s.

The deployment of the operator only has 1 container inside the pod created by k8s, strimzi-cluster-operator, which is a Java Application, so the question could be:
Why only give that application 20% of the memory requested to k8s? There isn't anything else that will be executed in that pod.

So perhaps the value that should be re-evaluated is the default request to k8s. 
This one:
https://github.com/strimzi/strimzi-kafka-operator/blob/main/helm-charts/helm3/strimzi-kafka-operator/values.yaml#L142

docker-images/operator/scripts/launch_java.sh
joseacl: Absolutely right! I'll remove them, thanks!

docker-images/operator/scripts/launch_java.sh
joseacl: I've simplified the process by removing that function, which does give us much value (the goal of that function was to round a number)

docker-images/operator/scripts/launch_java.sh
joseacl: Thanks for the explanation. For us, those values are working fine.

But since you have more information about the behaviour of the operator, we are going to keep this value aligned with MaxRAMPercentage.

calc_mem_opt "${CONTAINER_MAX_MEMORY}" "20"

Also, bear in mind that this workaround must be removed once the java version used by the operator is updated with a version without that issue.

scholzj (approved):
I started the system tests (although I assume they run on cgroups v1 and not much should change for them). LGTM otherwise. T-hanks for the PR and for incorporating the comments.

scholzj:
@ppatierno @tombentley Can you please review this?

tombentley (approved):
LGTM, thanks!

PR #6935, opened 2022-06-13
Set `zookeeper.electionPortBindRetry` to infinite retry

scholzj:
### Type of change

- Enhancement / new feature

### Description

Sometimes, when a new cluster is being deployed, the ZooKeeper pods first roll few times before they find each other. This PR adds new option `zookeeper.electionPortBindRetry` and sets it to `0` which means infinite retry. This should help the ZooKeeper nodes to find each other without unnecessary restarts (and based on my observations it does help to avoid the restarts).

This option is available only as a system property, so it is set directly in the `zookeeper_run.sh` script.

### Checklist


ppatierno (approved):
Good catch! Thanks!

PR #6945, opened 2022-06-15
fix: failing user reconciliation

### LINKED ISSUE #5691: Takes too long for KafkaUser to be ready
**Describe the bug**
When creating Kafka users say 1-5 users it works just fine, KafkaUsers are in Ready status in no time. But when I try to create bulk KafkaUser - say 50 users at a time, some users can take almost 10 mins to be "Ready", so this means client cannot publish/consume messages using these KafkaUsers - wait can be very long as the number of KafkaUsers on a Kafka cluster increase.

**To Reproduce**
1. To repro - you need 1. Kafka cluster, 1. topic already created.
2. Create 50 KafkaUsers on Strimzi Kafka cluster at the same time - via a  script (I create it via K8 api)
3. Right away with in a minute trying to use all these users to publish/read from using these users and you will get Topic Authorization Failed, Group Authorization Failed errors, because users are not yet "Ready"

2. Create Custom Resource 'Kafka User' using the yaml below:

```apiVersion: kafka.strimzi.io/v1beta2
kind: KafkaUser
metadata:
  name: cog-reader
  labels:
    strimzi.io/cluster: pod-kafka-cluster
spec:
  authentication:
    type: scram-sha-512
  authorization:
    type: simple
    acls:
      - resource:
          type: topic
          name: pod-notifications
          patternType: literal
        operation: Read
      - resource:
          type: topic
          name: pod-notifications
          patternType: literal
        operation: Describe
      - resource:
          type: group
          name: cog-group
          patternType: literal
        operation: Read
```

3. Authorization simple, authentication SASL_SSL/SCRAM-SHA-512
4. Apply 50 such users in parallel to Strimzi cluster.
5. Check the logs of the user-operator and it is cluttered with TimeoutExceptions

```
2021-10-09 06:55:40 DEBUG KafkaAdminClient:815 - [AdminClient clientId=adminclient-1] Call(callName=describeAcls, deadlineMs=1633762539991, tries=1, nextAllowedTryMs=1633762540099) timed out at 1633762539999 after 1 attempt(s)
java.lang.Exception: TimeoutException: Timed out waiting to send the call. Call: describeAcls
	at org.apache.kafka.clients.admin.KafkaAdminClient$Call.failWithTimeout(KafkaAdminClient.java:816) [org.apache.kafka.kafka-clients-2.8.0.jar:?]
	at org.apache.kafka.clients.admin.KafkaAdminClient$Call.fail(KafkaAdminClient.java:789) [org.apache.kafka.kafka-clients-2.8.0.jar:?]
	at org.apache.kafka.clients.admin.KafkaAdminClient$TimeoutProcessor.handleTimeouts(KafkaAdminClient.java:912) [org.apache.kafka.kafka-clients-2.8.0.jar:?]
	at org.apache.kafka.clients.admin.KafkaAdminClient$AdminClientRunnable.timeoutCallsToSend(KafkaAdminClient.java:993) [org.apache.kafka.kafka-clients-2.8.0.jar:?]
	at org.apache.kafka.clients.admin.KafkaAdminClient$AdminClientRunnable.processRequests(KafkaAdminClient.java:1301) [org.apache.kafka.kafka-clients-2.8.0.jar:?]
	at org.apache.kafka.clients.admin.KafkaAdminClient$AdminClientRunnable.run(KafkaAdminClient.java:1264) [org.apache.kafka.kafka-clients-2.8.0.jar:?]
	at java.lang.Thread.run(Thread.java:829) [?:?]
```
```
2021-10-09 06:55:40 DEBUG KafkaAdminClient:815 - [AdminClient clientId=adminclient-1] Call(callName=alterUserScramCredentials, deadlineMs=1633762539992, tries=1, nextAllowedTryMs=1633762540099) timed out at 1633762539999 after 1 attempt(s)
java.lang.Exception: TimeoutException: Timed out waiting to send the call. Call: alterUserScramCredentials
	at org.apache.kafka.clients.admin.KafkaAdminClient$Call.failWithTimeout(KafkaAdminClient.java:816) [org.apache.kafka.kafka-clients-2.8.0.jar:?]
	at org.apache.kafka.clients.admin.KafkaAdminClient$Call.fail(KafkaAdminClient.java:789) [org.apache.kafka.kafka-clients-2.8.0.jar:?]
	at org.apache.kafka.clients.admin.KafkaAdminClient$TimeoutProcessor.handleTimeouts(KafkaAdminClient.java:912) [org.apache.kafka.kafka-clients-2.8.0.jar:?]
	at org.apache.kafka.clients.admin.KafkaAdminClient$AdminClientRunnable.timeoutCallsToSend(KafkaAdminClient.java:993) [org.apache.kafka.kafka-clients-2.8.0.jar:?]
	at org.apache.kafka.clients.admin.KafkaAdminClient$AdminClientRunnable.processRequests(KafkaAdminClient.java:1301) [org.apache.kafka.kafka-clients-2.8.0.jar:?]
	at org.apache.kafka.clients.admin.KafkaAdminClient$AdminClientRunnable.run(KafkaAdminClient.java:1264) [org.apache.kafka.kafka-clients-2.8.0.jar:?]
	at java.lang.Thread.run(Thread.java:829) [?:?]
```
```TimeoutExecption
rg.apache.kafka.common.errors.TimeoutException: Call(callName=describeClientQuotas, deadlineMs=1633758992264, tries=1, nextAllowedTryMs=1633758992365) timed out at 1633758992265 after 1 attempt(s)
Caused by: org.apache.kafka.common.errors.TimeoutException: Timed out waiting to send the call. Call: describeClientQuotas
2021-10-09 05:56:32 DEBUG StatusDiff:46 - Status differs: {"op":"add","path":"/conditions/0/reason","value":"TimeoutException"}
2021-10-09 05:56:32 DEBUG StatusDiff:48 - Desired Status path /conditions/0/reason has value "TimeoutException"
org.apache.kafka.common.errors.TimeoutException: Call(callName=describeClientQuotas, deadlineMs=1633758992264, tries=1, nextAllowedTryMs=1633758992365) timed out at 1633758992265 after 1 attempt(s)
Caused by: org.apache.kafka.common.errors.TimeoutException: Timed out waiting to send the call. Call: describeClientQuotas
java.lang.Exception: TimeoutException: Timed out waiting to send the call. Call: describeClientQuotas
java.lang.Exception: TimeoutException: Timed out waiting to send the call. Call: describeClientQuotas

java.lang.Exception: TimeoutException: Timed out waiting to send the call. Call: describeClientQuotas
        at org.apache.kafka.clients.admin.KafkaAdminClient$Call.failWithTimeout(KafkaAdminClient.java:816) [org.apache.kafka.kafka-clients-2.8.0.jar:?]
        at org.apache.kafka.clients.admin.KafkaAdminClient$Call.fail(KafkaAdminClient.java:789) [org.apache.kafka.kafka-clients-2.8.0.jar:?]
        at org.apache.kafka.clients.admin.KafkaAdminClient$TimeoutProcessor.handleTimeouts(KafkaAdminClient.java:912) [org.apache.kafka.kafka-clients-2.8.0.jar:?]
        at org.apache.kafka.clients.admin.KafkaAdminClient$AdminClientRunnable.timeoutCallsToSend(KafkaAdminClient.java:993) [org.apache.kafka.kafka-clients-2.8.0.jar:?]
        at org.apache.kafka.clients.admin.KafkaAdminClient$AdminClientRunnable.processRequests(KafkaAdminClient.java:1301) [org.apache.kafka.kafka-clients-2.8.0.jar:?]
        at org.apache.kafka.clients.admin.KafkaAdminClient$AdminClientRunnable.run(KafkaAdminClient.java:1264) [org.apache.kafka.kafka-clients-2.8.0.jar:?]
        at java.lang.Thread.run(Thread.java:829) [?:?]
```

**Expected behavior**
Creating Kafka user with ACL's should be a straightforward call and it should have deterministic output when 5 users are created or 50 users are created 

 - Strimzi version: [e.g. main, 0.25.0]
 - Installation method: 
    - kubectl get ns kafka || kubectl create ns kafka
    - helm repo add strimzi https://strimzi.io/charts/
    - helm upgrade --install strimzi-kafka strimzi/strimzi-kafka-operator --version="0.25.0" -n kafka

 - Kubernetes cluster:  EKS 1.18
 - Infrastructure: Amazon EKS

**YAML files and logs**
Kafka -cluster yaml 

```
kind: Kafka
metadata:
  name: pod-kafka-cluster
  namespace: kafka
spec:
  kafka:
    version: 2.8.0
    replicas: 3
    listeners:
      - name: plain
        port: 9092
        tls: false
        type: internal
      - name: external
        authentication:
          type: scram-sha-512
        port: 9094
        tls: true
        type: loadbalancer
    authorization:
      type: simple
    config:
      log.message.format.version: "2.8"
    storage:
      type: persistent-claim
      size: 50Gi
      deleteClaim: true
  zookeeper:
    replicas: 3
    storage:
      type: persistent-claim
      size: 50Gi
      deleteClaim: true
  entityOperator:
    userOperator: 
      reconciliationIntervalSeconds: 157680000
      logging:
        type: inline
        loggers:
          rootLogger.level: DEBUG      
    topicOperator: 
      reconciliationIntervalSeconds: 157680000
      logging:
        type: inline
        loggers:
          rootLogger.level: DEBUG      
```
Topic yaml  
```
kind: KafkaTopic
metadata:
  name: pod-notifications
  labels:
    strimzi.io/cluster: pod-kafka-cluster
spec:
  partitions: 1
  replicas: 3
```
Publisher user (create multiple - this is just an example)
```
kind: KafkaUser
metadata:
  name: pod-agent-writer
  labels:
    strimzi.io/cluster: pod-kafka-cluster
spec:
  authentication:
    type: scram-sha-512
  authorization:
    type: simple
    acls:
      - resource:
          type: topic
          name: pod-notifications
          patternType: literal
        operation: Write
      - resource:
          type: topic
          name: pod-notifications
          patternType: literal
        operation: Describe
```

Consumer - create multiple this is jus an example
```
kind: KafkaUser
metadata:
  name: cog-reader
  labels:
    strimzi.io/cluster: pod-kafka-cluster
spec:
  authentication:
    type: scram-sha-512
  authorization:
    type: simple
    acls:
      - resource:
          type: topic
          name: pod-notifications
          patternType: literal
        operation: Read
      - resource:
          type: topic
          name: pod-notifications
          patternType: literal
        operation: Describe
      - resource:
          type: group
          name: cog-group
          patternType: literal
        operation: Read

```

**Additional context**
Use case: I have need to create Kafka. users/topics on demand, so I have a micro service that invokes K8's api that in-turn invoke Strimzi API's to create Kafka resources. topic operator works fine and topics are in ready status. KafkaUsers can take too long to be ready - and this causes some components that rely on this micro service to fail. 

Earlier periodic reconciliation was causing too much issues, so I turned periodic reconciliation off. Now as the number of users have increased all those timeout errors are back. I don't know. if Kafka. Admin client is already busy and ignoring all the calls? 

Any help here is appreciated. Any pointers,

ctron:
closes #5691

### Type of change

_Select the type of your PR_

- Bugfix

### Description

This change performs the full/periodic reconciliation
of the user resource sequentially. This prevent this
operator from spiking a lot of requests and constantly
failing user resources.

### Checklist

- [ ] Write tests
- [ ] Update documentation
- [ ] Update CHANGELOG.md


ppatierno:
I have concerns about this change, mainly because it has an impact on all the other operators because it's the common code across all of them.
For example, with this change (that maybe could help with the UO, but I said "maybe") we could have problems when reconciling a lot of Kafka CRs (think about a cluster operator handling even 50-60 Kafka CRs or more). Waiting each reconcile one by one could mean ages to complete a cycle across all the Kafka CRs (I could think reconciliation where rolling pods are involved for example, a Kafka version upgrade, certificates renewals, ...). Meanwhile a periodic reconcile could be triggered as well (queued again and again?).
Said that not sure what the best solution would be on the UO unless increasing the resources.

ctron:
> Said that not sure what the best solution would be on the UO unless increasing the resources.

That is not a solution. Adding more users increases the spike of requests that this produces, and massively increases the chance of failing (all). Despite the fact that the user operator and kafka broker consume much less resources then they got allocated.

The only "solution" I saw in the other issue was to disable periodic reconciliation and create a single user at a time: https://github.com/strimzi/strimzi-kafka-operator/issues/5691#issuecomment-1049507903

As to the concern of influencing other operators, that could easily be addresses by making this behavior specific to the `KafkaUserOperator`. I amended the PR accordingly.

> Meanwhile a periodic reconcile could be triggered as well (queued again and again?).

How do you currently guard against this issue?

If you don't, then I would suggest to add a check into the `reconcileAll` method, which could check if a current reconciliation is in progress, and only execute one if there is none running, or schedule one after the completion of the current (if not already scheduled).

Does that address your concerns?

ctron:
This PR fixes the issue for us, for the case of the periodic reconciles. Still, when a watch is created/re-created, this will overload the operator again.

Which somehow proves to me that the approach is working just fine.

The creation of the watch works differently, as it feeds events one by one into the reconciler. As this happens a less frequently than the periodic reconcile, it still is an improvement. I will try to get the other case fixed as well.

ctron:
> It does not set any boundaries for the queue, it does not handle the events etc.

Well there are no boundaries for the Set either? And maybe you could elaborate on which "events" you mean.

> This needs a proper solution and not a half-backed code which you dump on someone else to deal with the issues and problems it causes.

What part of this would be "half-backed"? And who said something about dumping the code? And what problems does it cause?

Maybe, instead of just grumbling, you do an actual review? Or provide some useful feedback.

scholzj:
> Well there are no boundaries for the Set either? And maybe you could elaborate on which "events" you mean.

Events means events from Kuberetes watches => resources being created, updated or deleted. With boundaries I mean that you just queue more and more stuff without any limit.

ctron:
You did notice this is a draft PR, right? So it is not supposed to be complete.

And regarding the watch creation I said that:

> I will try to get the other case fixed as well.

About:

> With boundaries I mean that you just queue more and more stuff without any limit.

This is what happens right now. There are no limits to the reconciliation process that I saw. Maybe there are, I just didn't see them.

But also regarding that, I gave a proposal on how to fix this: https://github.com/strimzi/strimzi-kafka-operator/pull/6945#issuecomment-1158631121 … I just didn't hear any feedback from you if this would be an acceptable approach.

I am trying to figure out if it is worth my time to complete this PR. I have a solution for our case. But I think it would make sense to fix this issue for others too.

ppatierno:
@ctron I and Jakub had an offline discussion about this issue. Jakub has a plan to address the problem which is going to work for all the use cases without "breaking" the behaviour of other components. To be honest there is not ETA for it but that should be across July/August.
If you are ok I would close this PR and we could wait for the future fix. 
Meanwhile, if you really need it, you could use the image you already built with your fix. What do you think?

ctron:
Yea, I wanted to update this PR today anyway. We switched away from using the entity operator now. We still have the option to integrate with it, but focus on the Kafka Admin API now. So, no need to rush.

PR #7003, opened 2022-06-29
Check if cgroups v2 file contains number or not

im-konge:

### Type of change

_Select the type of your PR_

- Bugfix
- Enhancement / new feature
- Refactoring
- Documentation

### Description

### Checklist

- [ ] Write tests
- [ ] Update documentation
- [ ] Update CHANGELOG.md


docker-images/operator/scripts/launch_java.sh
```diff
@@ -30,6 +30,21 @@ calc_max_memory() {
...
+  local calc_cgroups_mem=false
+  if [[ -r "${mem_file}" ]]; then
+    value=$(cat "${mem_file}")
+    if [[ $value =~ $num_pattern ]]; then
+      calc_cgroups_mem=true
+    fi
+  fi
+
+  echo $calc_cgroups_mem
+}
```
scholzj, line 46: This check seems wrong based on the docs, I would expect you to check specifically for `max`.

im-konge:
@scholzj that was just my assumption, as other pods, which don't contain multiple containers contain numerical values.

If we should go with `memory.high`, then I can change it

docker-images/operator/scripts/launch_java.sh
im-konge: Well, I thought that there could be also different values than just `max` - that's why I'm checking, if the file contains a number.

im-konge: Okay, I will check what the `memory.high` contains when I will not specify any resources limits inside the deployment and let you know my findings

im-konge:
@scholzj yeah that's what this should actually do - but I'm checking, that there is a number. So I will change it to check `max` instead of number. Thanks both :)

im-konge:
@scholzj @joseacl I changed it to look for the `max` value, can you please have another look at this?

scholzj (commented):
I think the code will be much cleaner if you follow the previous structure.
* Check if it is using Cgroups v2
    * Handle in that code branch whether and how to set the limits. This should also make it easy to avoid any additional procedures whihc just complicate the code and made it harder to read and test it.
* Keep the Cgroups v1 in its own code section.
* Add some comments explaining what the code does and why

That way, it will be much more transparent what are you doing and easier to figure out what the code does when needed.

ppatierno (approved):
LGTM

PR #7006, opened 2022-06-30
docs: updates for MM2 handling large message volumes

PaulRMellor:

**Documentation**

Updates [Handling high volumes of messages](https://strimzi.io/docs/operators/in-development/configuring.html#con-mirrormaker-high-volume-messages-str) in the _Configuring_ guide.

MirrorMaker2 used to fail when handling large volumes of messages because of a high producer buffer or low offset commit timeout. This caused a failed to flush error. The underlying issue has now been fixed.

The section has been updated to explain how you can tune the MM2 config to handle high volumes of messages, rather than workaround the error.

### Checklist

- [ ] Write tests
- [x] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/documentation/modules/configuring/con-config-mirrormaker2-high-volume-messages.adoc
+++ b/documentation/modules/configuring/con-config-mirrormaker2-high-volume-messages.adoc
@@ -6,16 +6,18 @@
 = Handling high volumes of messages
 
 [role="_abstract"]
-If your MirrorMaker 2.0 deployment is going to be handling a high volume of messages, you might need to adjust its configuration to support it. 
-The configuration can help manage throughput and issues with latency. 
+If your MirrorMaker 2.0 deployment is going to be handling a high volume of messages, you might need to adjust its configuration configuration to help manage throughput and issues with latency. 
+Operations like tracking web activity can generate a high volume of messages. 
+But even a source cluster with moderate throughput can produce a high volume of messages if it has a lot of existing data. 
 
-MirrorMaker 2.0 uses an internal producer to replicate data from the source to the target Kafka cluster. 
-To control the streaming of high volumes of data, you can tune the producer batch size configuration for optimal throughput. 
+MirrorMaker 2.0 fetches data from the source cluster and hands it to the Kafka Connect runtime producers so that it's replicated to the target cluster.
+
+To control the streaming of high volumes of data, you can tune the configuration of the Kafka Connect runtime producers for optimal throughput.
 
 In certain circumstances, you might want to increase the size of the message batches sent in a single produce request.
-You increase the batch size using the `batch.size` producer configuration to reduce the number of outstanding messages ready to be sent.
+You increase the batch size using the `batch.size` producer configuration.
+A larger batch size reduces the number of outstanding messages ready to be sent and the size of the backlog in the message queue
 Messages being sent to the same partition are batched together.
-A larger batch size reduces the number of requests and the size of the backlog in the message queue.
 A produce request is sent to the target cluster when the batch size is reached.
 By increasing the batch size, produce requests are delayed and more messages are added to the batch and sent to brokers at the same time.  
 This can improve throughput when you have just a few topic partitions that handle large numbers of messages.  
@@ -23,7 +25,7 @@ This can improve throughput when you have just a few topic partitions that handl
 Consider the number and size of the records that the producer handles for a suitable producer batch size. 
 
 Use `linger.ms` to add a wait time in milliseconds to delay produce requests when producer load decreases. 
-What the delay means is that more records can be added to batches if they are under the maximum batch size.  
+The delay means that more records can be added to batches if they are under the maximum batch size.  
 
 Configure the `batch.size` and `linger.ms` values at the source connector level (`producer.override.*`), as they relate to the specific producer that sends topic messages to the target Kafka cluster.
 
@@ -35,12 +37,12 @@ The data replication pipeline to the target Kafka cluster is as follows:
 The producer sends messages in its buffer to topics in the target Kafka cluster.
 While this is happening, Kafka Connect tasks continue to poll topics in the source Kafka cluster to add messages to the source message queue.
 
-The size of the buffer is set using the `producer.buffer.memory` property. 
+The size of the producer buffer for the source connector is set using the `producer.override.buffer.memory` property.
 Tasks wait for a specified timeout period (`offset.flush.timeout.ms`) before the buffer is flushed. 
 This should be enough time for the sent messages to be acknowledged by the brokers and offset data committed. 
 The source task does not wait for the producer to empty the message queue before committing offsets, except during shutdown.
 
-If the producer is unable to keep up with the throughput of messages in the source message queue, buffering is blocked until there is space available in the buffer within a time period allocated by `max.block.ms`.
+If the producer is unable to keep up with the throughput of messages in the source message queue, buffering is blocked until there is space available in the buffer within a time period bounded by `max.block.ms`.
 Any unacknowledged messages still in the buffer are sent during this period.
 New messages are not added to the buffer until these messages are acknowledged and flushed.
 
--- a/documentation/modules/configuring/con-config-mirrormaker2-high-volume-messages.adoc
+++ b/documentation/modules/configuring/con-config-mirrormaker2-high-volume-messages.adoc
@@ -38,11 +38,7 @@ While this is happening, Kafka Connect tasks continue to poll topics in the sour
 The size of the buffer is set using the `producer.buffer.memory` property. 
 Tasks wait for a specified timeout period (`offset.flush.timeout.ms`) before the buffer is flushed. 
 This should be enough time for the sent messages to be acknowledged by the brokers and offset data committed. 
-
-NOTE: Kafka 3.1.0 and newer versions do not wait for the producer to empty the message queue before committing offsets, except during shutdown.
-
-Try to avoid a situation where there are too many messages in the buffer, so they can't all be flushed before the offset flush timeout is reached.
-The offset flush timeout period needs to be sufficient for the size of the buffer.
+Kafka does not wait for the producer to empty the message queue before committing offsets, except during shutdown.
 
 If the producer is unable to keep up with the throughput of messages in the source message queue, buffering is blocked until there is space available in the buffer within a time period allocated by `max.block.ms`.
 Any unacknowledged messages still in the buffer are sent during this period.
@@ -84,7 +80,7 @@ spec:
       tasksMax: 2
       config:
         producer.override.batch.size: 327680
-        producer.override.linger.ms: 5
+        producer.override.linger.ms: 100
     # ...
   resources: 
     requests:
--- a/documentation/modules/configuring/con-config-mirrormaker2-high-volume-messages.adoc
+++ b/documentation/modules/configuring/con-config-mirrormaker2-high-volume-messages.adoc
@@ -9,7 +9,7 @@
 If your MirrorMaker 2.0 deployment is going to be handling a high volume of messages, you might need to adjust its configuration to support it. 
 The configuration can help manage throughput and issues with latency. 
 
-MirrorMaker 2.0 uses an internal producer client to replicate data from the source to the target Kafka cluster. 
+MirrorMaker 2.0 uses an internal producer to replicate data from the source to the target Kafka cluster. 
 To control the streaming of high volumes of data, you can tune the producer batch size configuration for optimal throughput. 
 
 In certain circumstances, you might want to increase the size of the message batches sent in a single produce request.
@@ -20,16 +20,16 @@ A produce request is sent to the target cluster when the batch size is reached.
 By increasing the batch size, produce requests are delayed and more messages are added to the batch and sent to brokers at the same time.  
 This can improve throughput when you have just a few topic partitions that handle large numbers of messages.  
 
-Consider the number and size of records for a suitable producer batch size. 
-Use something like the following formula: `max.poll.records * record-size-avg * 1024`. 
-`max.poll.records` is the maximum number of messages to return in a poll. 
-`record-size-avg` is a metric that gives the average size of a records. 
+Consider the number and size of the records that the producer handles for a suitable producer batch size. 
 
-Use `producer.linger.ms` to add a wait time in milliseconds to delay produce requests when producer load decreases. 
-What the delay means is that more records can be added to batches if they are under the maximum batch size.       
+Use `linger.ms` to add a wait time in milliseconds to delay produce requests when producer load decreases. 
+What the delay means is that more records can be added to batches if they are under the maximum batch size.  
 
-The data replication pipeline is as follows:
+Configure the `batch.size` and `linger.ms` values at the source connector level (`producer.override.*`), as they relate to the specific producer that sends topic messages to the target Kafka cluster.
 
+The data replication pipeline to the target Kafka cluster is as follows:
+
+.Data replication pipeline
 *source topic -> (Kafka Connect tasks) source message queue -> producer buffer -> target topic*.
 
 The producer sends messages in its buffer to topics in the target Kafka cluster.
@@ -76,14 +76,15 @@ spec:
   - alias: "my-cluster-target"
     config:
       offset.flush.timeout.ms: 10000
-      producer.batch.size: 327680
-      producer.linger.ms: 5
     bootstrapServers: my-cluster-target-kafka-bootstrap:9092
   mirrors:
   - sourceCluster: "my-cluster-source"
     targetCluster: "my-cluster-target"
     sourceConnector:
       tasksMax: 2
+      config:
+        producer.override.batch.size: 327680
+        producer.override.linger.ms: 5
     # ...
   resources: 
     requests:
--- a/documentation/modules/configuring/con-config-mirrormaker2-high-volume-messages.adoc
+++ b/documentation/modules/configuring/con-config-mirrormaker2-high-volume-messages.adoc
@@ -6,26 +6,53 @@
 = Handling high volumes of messages
 
 [role="_abstract"]
-If your MirrorMaker 2.0 deployment is going to be handling a high volume of messages, you might need to adjust its configuration to support it.
+If your MirrorMaker 2.0 deployment is going to be handling a high volume of messages, you might need to adjust its configuration to support it. The configuration can help manage throughput and issues with latency. 
 
-The flush pipeline for data replication is *source topic -> (Kafka Connect) source message queue -> producer buffer -> target topic*.
-An offset flush timeout period (`offset.flush.timeout.ms`) is the time to wait for the producer buffer (`producer.buffer.memory`) to flush and offset data to be committed.
-Try to avoid a situation where a large producer buffer and an insufficient offset flush timeout period causes a _failed to flush_ or _failed to commit offsets_ type of error.
+MirrorMaker 2.0 uses an internal producer client (`KafkaProducer`) to replicate data from the source to target Kafka cluster. 
+To control the streaming of high volumes of data, you can tune the following producer configuration for optimal throughput: 
 
-This type of error means that there are too many messages in the producer buffer, so they can't all be flushed before the offset flush timeout is reached.
+* Producer batch size 
+* Producer buffer size 
 
-If you are getting this type of error, try the following configuration changes:
+.Producer batch size
+In certain circumstances, you might want to increase the size of the message batches sent in a single produce request.
+You increase the batch size using the `batch.size` producer configuration to reduce the number of outstanding messages ready to be sent.
+A larger batch size reduces the number of requests and the backlog in the message queue.
+A produce request is sent to the target cluster when the batch size is reached.
+By increasing the batch size, produce requests are delayed and more messages are added to the batch and sent to brokers at the same time.  
+This can improve throughput when you have just a few topic partitions that handle large numbers of messages. 
+
+NOTE: To be able to increase the batch size, the `connector.client.config.override.policy` must be set to `All`. This is the default setting. The property enables connector client configurations.   
+
+.Producer buffer size
+The data replication pipeline is as follows:
+
+*source topic -> (Kafka Connect tasks) source message queue -> producer buffer -> target topic*.
+
+The producer sends messages in its buffer to topics in the target Kafka cluster.
+While this is happening, Kafka Connect tasks continue to poll topics in the source Kafka cluster to add messages to the source message queue.
+
+The size of the buffer is set using the `producer.buffer.memory` property. 
+The producer waits for a specified timeout period (`offset.flush.timeout.ms`) before the buffer is flushed. 
+This should be enough time for the sent messages to be acknowledged by the brokers and offset data committed.
+
+Try to avoid a situation where there are too many messages in the buffer, so they can't all be flushed before the offset flush timeout is reached.
+The offset flush timeout period needs to be sufficient for the size of the buffer.
+
+If the producer is unable to keep up with the throughput of messages in the source message queue, buffering is blocked until there is space available in the buffer within a time period allocated by `max.block.ms`.
+Any unacknowledged messages still in the buffer are sent during this period.
+New messages are not added to the buffer until these messages are acknowledged and flushed.
+
+You can try the following configuration changes to keep the underlying source message queue of outstanding messages at a manageable size:
 
 * Decreasing the default value in bytes of the `producer.buffer.memory`
 * Increasing the default value in milliseconds of the `offset.flush.timeout.ms`
+* Ensuring that there are enough xref:con-common-configuration-resources-reference[CPU and memory resources]
+* Increasing the number of tasks that run in parallel by doing the following:
+** xref:con-mirrormaker-tasks-max-{context}[Increasing the number of tasks] using the `tasksMax` property
+** Increasing the number of nodes for the workers that run tasks using the `replicas` property
 
-The changes should help to keep the underlying Kafka Connect queue of outstanding messages at a manageable size.
-You might need to adjust the values to have the desired effect.
-
-If these configuration changes don't resolve the error, you can try increasing the number of tasks that run in parallel by doing the following.
-
-* xref:con-mirrormaker-tasks-max-{context}[Increasing the number of tasks] using the `tasksMax` property
-* Increasing the number of nodes for the workers that run tasks using the `replicas` property
+You might need to keep adjusting the configuration values until they have the desired effect.
 
 .Example MirrorMaker 2.0 configuration for handling high volumes of messages
 [source,yaml,subs="+quotes,attributes"]
@@ -45,12 +72,21 @@ spec:
     config:
       offset.flush.timeout.ms: 10000
       producer.buffer.memory: 8388608
+      producer.batch.size: 327680
     bootstrapServers: my-cluster-target-kafka-bootstrap:9092
   mirrors:
   - sourceCluster: "my-cluster-source"
     targetCluster: "my-cluster-target"
     sourceConnector:
       tasksMax: 10
+    # ...
+  resources: 
+    requests:
+      cpu: "1"
+      memory: Gi
+    limits:
+      cpu: "2"
+      memory: 4Gi      
 ----
 
 == Checking the message flow
--- a/documentation/modules/configuring/con-config-mirrormaker2-high-volume-messages.adoc
+++ b/documentation/modules/configuring/con-config-mirrormaker2-high-volume-messages.adoc
@@ -14,10 +14,19 @@ To control the streaming of high volumes of data, you can tune the producer batc
 
 In certain circumstances, you might want to increase the size of the message batches sent in a single produce request.
 You increase the batch size using the `batch.size` producer configuration to reduce the number of outstanding messages ready to be sent.
-A larger batch size reduces the number of requests and the backlog in the message queue.
+Messages being sent to the same partition are batched together.
+A larger batch size reduces the number of requests and the size of the backlog in the message queue.
 A produce request is sent to the target cluster when the batch size is reached.
 By increasing the batch size, produce requests are delayed and more messages are added to the batch and sent to brokers at the same time.  
 This can improve throughput when you have just a few topic partitions that handle large numbers of messages.  
+
+Consider the number and size of records for a suitable producer batch size. 
+Use something like the following formula: `max.poll.records * record-size-avg * 1024`. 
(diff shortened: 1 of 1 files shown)


documentation/modules/configuring/con-config-mirrormaker2-high-volume-messages.adoc
```diff
@@ -45,12 +72,21 @@ spec:
     config:
       offset.flush.timeout.ms: 10000
       producer.buffer.memory: 8388608
+      producer.batch.size: 327680
```
PaulRMellor, line 75: @fvaleri -- is this the right property name?

documentation/modules/configuring/con-config-mirrormaker2-high-volume-messages.adoc
```diff
@@ -6,26 +6,53 @@
...
+In certain circumstances, you might want to increase the size of the message batches sent in a single produce request.
+You increase the batch size using the `batch.size` producer configuration to reduce the number of outstanding messages ready to be sent.
+A larger batch size reduces the number of requests and the backlog in the message queue.
+A produce request is sent to the target cluster when the batch size is reached.
+By increasing the batch size, produce requests are delayed and more messages are added to the batch and sent to brokers at the same time.  
+This can improve throughput when you have just a few topic partitions that handle large numbers of messages. 
 
-* Decreasing the default value in bytes of the `producer.buffer.memory`
-* Increasing the default value in milliseconds of the `offset.flush.timeout.ms`
+NOTE: To be able to increase the batch size, the `connector.client.config.override.policy` must be set to `All`. This is the default setting. The property enables connector client configurations.
```
mimaison, line 25: It depends where the override is set. If it is set in the runtime configuration, via `producer.batch.size` then you don't need this to be set to `All`. However if the override in in the connector configuration, via `producer.override.batch.size`, then this is needed. 

Below we seem to set `producer.batch.size`. Assuming it is done in the runtime configuration (not entirely sure) then we don't need `connector.client.config.override.policy` to be set to any particular value.

documentation/modules/configuring/con-config-mirrormaker2-high-volume-messages.adoc
```diff
@@ -6,26 +6,53 @@
 = Handling high volumes of messages
 
 [role="_abstract"]
-If your MirrorMaker 2.0 deployment is going to be handling a high volume of messages, you might need to adjust its configuration to support it.
+If your MirrorMaker 2.0 deployment is going to be handling a high volume of messages, you might need to adjust its configuration to support it. The configuration can help manage throughput and issues with latency. 
 
-The flush pipeline for data replication is *source topic -> (Kafka Connect) source message queue -> producer buffer -> target topic*.
-An offset flush timeout period (`offset.flush.timeout.ms`) is the time to wait for the producer buffer (`producer.buffer.memory`) to flush and offset data to be committed.
-Try to avoid a situation where a large producer buffer and an insufficient offset flush timeout period causes a _failed to flush_ or _failed to commit offsets_ type of error.
+MirrorMaker 2.0 uses an internal producer client (`KafkaProducer`) to replicate data from the source to target Kafka cluster.
```
scholzj, line 11: I don't really undertsand the meaning of the `(KafkaProducer)` here.

documentation/modules/configuring/con-config-mirrormaker2-high-volume-messages.adoc
```diff
@@ -6,26 +6,53 @@
...
 
-The flush pipeline for data replication is *source topic -> (Kafka Connect) source message queue -> producer buffer -> target topic*.
-An offset flush timeout period (`offset.flush.timeout.ms`) is the time to wait for the producer buffer (`producer.buffer.memory`) to flush and offset data to be committed.
-Try to avoid a situation where a large producer buffer and an insufficient offset flush timeout period causes a _failed to flush_ or _failed to commit offsets_ type of error.
+MirrorMaker 2.0 uses an internal producer client (`KafkaProducer`) to replicate data from the source to target Kafka cluster. 
+To control the streaming of high volumes of data, you can tune the following producer configuration for optimal throughput: 
 
-This type of error means that there are too many messages in the producer buffer, so they can't all be flushed before the offset flush timeout is reached.
+* Producer batch size 
+* Producer buffer size
```
fvaleri, line 15: I would remove `buffer.size`, as it is a sub-optimal solution.
You may fix the problem by reducing this value, but the overall throughput would be bad.
Anyway, if we decide to keep, then we should explain how to set it.

documentation/modules/configuring/con-config-mirrormaker2-high-volume-messages.adoc
fvaleri: It is set at the runtime level in the following example, so we can get rid of it.

documentation/modules/configuring/con-config-mirrormaker2-high-volume-messages.adoc
```diff
@@ -6,26 +6,53 @@
...
+
+*source topic -> (Kafka Connect tasks) source message queue -> producer buffer -> target topic*.
+
+The producer sends messages in its buffer to topics in the target Kafka cluster.
+While this is happening, Kafka Connect tasks continue to poll topics in the source Kafka cluster to add messages to the source message queue.
 
-The changes should help to keep the underlying Kafka Connect queue of outstanding messages at a manageable size.
-You might need to adjust the values to have the desired effect.
+The size of the buffer is set using the `producer.buffer.memory` property. 
+The producer waits for a specified timeout period (`offset.flush.timeout.ms`) before the buffer is flushed.
```
fvaleri, line 36: Rather than "the producer", I would use "the task" as subject here.

This is not true after the recent fix, where the offsets commit happens asynchronously in another thread. Rather than be completely blocked, MM2 would make very slow progress in case of high volume.

documentation/modules/configuring/con-config-mirrormaker2-high-volume-messages.adoc
PaulRMellor: removed

documentation/modules/configuring/con-config-mirrormaker2-high-volume-messages.adoc
PaulRMellor: We don't need it. Removed

documentation/modules/configuring/con-config-mirrormaker2-high-volume-messages.adoc
PaulRMellor: Actually meaning the producer buffer memory here, which we describe later. I've changed to that

scholzj (approved):
LGTM. Should be also approved by the SMEs (@mimaison and @fvaleri).

documentation/modules/configuring/con-config-mirrormaker2-high-volume-messages.adoc
```diff
@@ -6,25 +6,24 @@
...
+If your MirrorMaker 2.0 deployment is going to be handling a high volume of messages, you might need to adjust its configuration to support it. 
+The configuration can help manage throughput and issues with latency. 
 
-MirrorMaker 2.0 uses an internal producer client (`KafkaProducer`) to replicate data from the source to target Kafka cluster. 
+MirrorMaker 2.0 uses an internal producer client to replicate data from the source to the target Kafka cluster. 
 To control the streaming of high volumes of data, you can tune the following producer configuration for optimal throughput: 
 
 * Producer batch size 
-* Producer buffer size 
+* Producer buffer memory
```
fvaleri, line 16: I would remove all buffer memory mentions (lines 16, 26, 47).

Decreasing this value doesn't directly address the issue (slow producer) and it is also complicated, as you would need to take into account the avg message size.

documentation/modules/configuring/con-config-mirrormaker2-high-volume-messages.adoc
```diff
@@ -78,7 +78,7 @@ spec:
   - sourceCluster: "my-cluster-source"
     targetCluster: "my-cluster-target"
     sourceConnector:
-      tasksMax: 10
+      tasksMax: 2
```
fvaleri, line 81: That's ok, but I would also set `spec.replicas: 1` instead of 5 to keep the example with minimal resources.

I think it is fine because we have notes that suggest to increase these values according to available resources.

documentation/modules/configuring/con-config-mirrormaker2-high-volume-messages.adoc
```diff
@@ -6,25 +6,24 @@
...
 A larger batch size reduces the number of requests and the backlog in the message queue.
 A produce request is sent to the target cluster when the batch size is reached.
 By increasing the batch size, produce requests are delayed and more messages are added to the batch and sent to brokers at the same time.  
-This can improve throughput when you have just a few topic partitions that handle large numbers of messages. 
+This can improve throughput when you have just a few topic partitions that handle large numbers of messages.  
 
-NOTE: To be able to increase the batch size, the `connector.client.config.override.policy` must be set to `All`. This is the default setting. The property enables connector client configurations.   
-
-.Producer buffer size
+.Producer buffer memory
```
fvaleri, line 26: I thought we just wanted to provide some tuning advice here, but if we want to include the root cause analysis, then it would be something like this, with the note about recent changes in that space:

When there is a high volume of messages, mirroring may be slow or not working at all.
This happens because the underline **Producer is not optimized for throughput**.

At every iteration, the MirrorMaker2 flush pipeline is the following:

```
source-topic -> Connect outstanding messages queue (# records) -> producer buffer (bytes) -> target-topic
```

The `MirrorSourceTask` periodically `poll()` from the source topic, even before the previous batch of records has been acknowledged.
In the meantime, a different thread tries to commit the source topic offsets every `offset.flush.timeout.ms`.
One key point here is that the commit process waits until the outstanding messages queue is empty to proceed.

When there is a high volume of messages coming in, the outstanding messages queue can grow unbounded (there is no hard limit).
This causes the Producer's buffer to fill up and the `send()` operation to become synchronous, which also prevents the `MirrorSourceTask` from polling.
Eventually, the Producer becomes unable to free that queue before the commit times out and fails.

>Since Kafka 3.1.0 (Strimzi 0.28.0), the offsets commit does not wait for the producer to empty the queue and the `offset.flush.timeout.ms` only applies to the actual writing of offset data to the internal offsets topic.
> Mirroring can still be very slow when having a high volume, so the optimization proposed here still applies.

PaulRMellor: Thanks @federico. I think we cover the config that handles the pipeline in enough detail for context, but please add any specific suggestions if you think we are missing something.

fvaleri: There is still no mention of recent changes (see the quote at the end of my previous message). We should also remove `producer.buffer.memory: 8388608` from the example.

fvaleri:
I'm wondering if we should add a formula to determine the `producer.batch.size`, something like `max.poll.records * record-size-avg * 1024` (currently we have default*20). We also probably need to add a value for `producer.linger.ms`, to avoid big latency when the load decreases.

WDYT?

documentation/modules/configuring/con-config-mirrormaker2-high-volume-messages.adoc
PaulRMellor: Thanks @fvaleri  -- I added a note and removed `producer.buffer.memory` from the example

tyamashi-oss:
MirrorSourceConnector has the role of transferring the message volumes, and ideally, the configuration should be configured only for its consumer/producer. To do so, the following configurations can be used:
- MirrorSourceConnector's producer    mirrors.sourceConnector.config: producer.override.\*
- MirrorSourceConnector's consumer   mirrors.sourceConnector.config: consumer.\*
I think the above method is rather better. 

As far as I researched/tested, MirrorMaker2 uses 6 producers and 5 consumers internally as of now. If use {target-cluster}.config: producer.buffer.memory/producer.batch.size as the example in the documentation, it affects MirrorMaker2-specific 3 producers(, not only MirrorSourceConnector's producer but also MirrorCheckpointConnector's and MirrorHeartbeatConnector's one). its effects sometimes lead to unexpected behaviors and errors. 

I understand that this complexity is caused by MirrorMaker2 not disclosing the internal producer/consumer. And "override" with and without depends on whether the producer/consumer was created with the Kafka Connect framework or MirrorMaker2. I have described the results of my investigation at https://issues.redhat.com/browse/ENTMQST-4116. The above method is also not perfect(I am discouraged rather), but I believe it is safer than {target-cluster}.config: producer.buffer.memory.

documentation/modules/configuring/con-config-mirrormaker2-high-volume-messages.adoc
PaulRMellor: Also added linger

PaulRMellor:
> MirrorSourceConnector has the role of transferring the message volumes, and ideally, the configuration should be configured only for its consumer/producer. To do so, the following configurations can be used:
> 
> * MirrorSourceConnector's producer    mirrors.sourceConnector.config: producer.override.*
> * MirrorSourceConnector's consumer   mirrors.sourceConnector.config: consumer.*
>   I think the above method is rather better.
> 
> As far as I researched/tested, MirrorMaker2 uses 6 producers and 5 consumers internally as of now. If use {target-cluster}.config: producer.buffer.memory/producer.batch.size as the example in the documentation, it affects MirrorMaker2-specific 3 producers(, not only MirrorSourceConnector's producer but also MirrorCheckpointConnector's and MirrorHeartbeatConnector's one). its effects sometimes lead to unexpected behaviors and errors.
> 
> I understand that this complexity is caused by MirrorMaker2 not disclosing the internal producer/consumer. And "override" with and without depends on whether the producer/consumer was created with the Kafka Connect framework or MirrorMaker2. I have described the results of my investigation at https://issues.redhat.com/browse/ENTMQST-4116. The above method is also not perfect(I am discouraged rather), but I believe it is safer than {target-cluster}.config: producer.buffer.memory.


Thanks @[tyamashi-oss](https://github.com/tyamashi-oss) --

We've taken out the `producer.buffer.memory` config from this update. 
Additionally, we're looking at adding a section that describes the producer/consumer config for MM2 in a separate PR.
In terms of this PR, you're saying that the `producer.batch.size` config for this section should actually be under `sourceConnector` ?

@fvaleri @mimaison FYI^^

documentation/modules/configuring/con-config-mirrormaker2-high-volume-messages.adoc
fvaleri: Great, it looks almost good.

For `producer.linger.ms` I was searching for a confirmation, but probably it is not directly related to the problem at hand, so I would leave it out.

fvaleri:
> it affects MirrorMaker2-specific 3 producers(, not only MirrorSourceConnector's producer but also MirrorCheckpointConnector's and MirrorHeartbeatConnector's one). its effects sometimes lead to unexpected behaviors and errors.

Hi @tyamashi-oss, I've seen no issue setting `buffer.size` at the runtime level.

Can you be more specific about this? Which unexpected behaviors and errors?

fvaleri:
> As far as I know, MM2 has a problem where, under some conditions, the consumer offset can get out of sync due to timing issues with sending/receiving messages 

We had a bug there, which is fixed now. 

With regards to `producer.buffer.size` optimization, I don't see how this is related, but if you have a reproducer I will be happy to look at it.

> For more efficient data transfer by MirrorSourceConnector, it makes sense to increase buffer.memory and batch.size.

Actually, if you don't increase `producer.buffer.size`, increasing `producer.buffer.memory` would make things worst in pre Kafka 3.1 releases, as you will reach the flush timeout faster (where no data is committed anymore). This issue is now fixed, but mirroring may still be slow and the most effective tuning is `producer.buffer.size`. 

As with all Kafka default values, I would not increase `producer.buffer.memory`, unless there is a good reason which is effective on most environments.

You can read more on this here:

https://www.javierholguera.com/2020/06/16/kafka-connect-offset-commit-errors-ii

tyamashi-oss:
Hello @fvaleri,

As a conclusion, I have no objection to using {target-cluster}.config: producer.* in the documentation as in the past, because I just shared a risk, just in case.

To configure the MirrorSourceConnector producer, my personal opinion is that it is preferable to use mirrors.sourceConnector.config: producer.override.* rather {target-cluster}.config: producer.* because it does not affect other connectors such as MirrorCheckpointConnector and MirrorCheckpointConnector. However, I think mirrors.sourceConnector.config: producer.override.* is also not perfect because MirrorMaker2 does not disclose the internal producer/consumer. I have no objection to using {target-cluster}.config: producer.* in the documentation as in the past. Thank you.

PaulRMellor:
> Hello @fvaleri,
> 
> As a conclusion, I have no objection to using {target-cluster}.config: producer.* in the documentation as in the past, because I just shared a risk, just in case.
> 
> To configure the MirrorSourceConnector producer, my personal opinion is that it is preferable to use mirrors.sourceConnector.config: producer.override.* rather {target-cluster}.config: producer.* because it does not affect other connectors such as MirrorCheckpointConnector and MirrorCheckpointConnector. However, I think mirrors.sourceConnector.config: producer.override.* is also not perfect because MirrorMaker2 does not disclose the internal producer/consumer. I have no objection to using {target-cluster}.config: producer.* in the documentation as in the past. Thank you.

Thanks @tyamashi-oss  -- we could add a note in this section to say that it's also possible to add the producer config at the connector level? Something like this

> NOTE: Producer configuration at the cluster level applies to the producers of all connectors. As the change relates specifically to the producer of the source connector, you can add the configuration at the connector level by adding properties under `mirrors.sourceConnector.config.producer.*`.  

@fvaleri @mimaison -- any further comments or are you happy to approve?

fvaleri:
@PaulRMellor let me test the connector level setting and I'll report back.

fvaleri:
Ok, the connector level setup works fine both with small and big messages.
I would also remove the `batch.size` formula because it's not tested.

```
apiVersion: {KafkaMirrorMaker2ApiVersion}
kind: KafkaMirrorMaker2
metadata:
  name: my-mirror-maker2
spec:
  version: {DefaultKafkaVersion}
  replicas: 1
  connectCluster: "my-cluster-target"
  clusters:
  - alias: "my-cluster-source"
    bootstrapServers: my-cluster-source-kafka-bootstrap:9092
  - alias: "my-cluster-target"
    bootstrapServers: my-cluster-target-kafka-bootstrap:9092
  mirrors:
  - sourceCluster: "my-cluster-source"
    targetCluster: "my-cluster-target"
    sourceConnector:
      tasksMax: 2
      config:
          # source connector producer tuning
          producer.override.batch.size: 327680
    # ...
  resources:
    requests:
      cpu: "1"
      memory: Gi
    limits:
      cpu: "2"
      memory: 4Gi
```

Thanks @tyamashi-oss for raising this.

documentation/modules/configuring/con-config-mirrormaker2-high-volume-messages.adoc
```diff
@@ -6,26 +6,58 @@
...
+While this is happening, Kafka Connect tasks continue to poll topics in the source Kafka cluster to add messages to the source message queue.
 
-The changes should help to keep the underlying Kafka Connect queue of outstanding messages at a manageable size.
-You might need to adjust the values to have the desired effect.
+The size of the buffer is set using the `producer.buffer.memory` property. 
+Tasks wait for a specified timeout period (`offset.flush.timeout.ms`) before the buffer is flushed. 
+This should be enough time for the sent messages to be acknowledged by the brokers and offset data committed. 
 
-If these configuration changes don't resolve the error, you can try increasing the number of tasks that run in parallel by doing the following.
+NOTE: Kafka 3.1.0 and newer versions do not wait for the producer to empty the message queue before committing offsets, except during shutdown.
```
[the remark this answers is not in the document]
fvaleri: Yeah, I was also wondering that in a previous comment. The tuning proposed here is still useful in case of high latency, but after that release, which includes the fix, you can't be in a situation where MM2 appears as blocked (Failed to flush).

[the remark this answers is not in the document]
PaulRMellor: I've removed the notes and incorporated into the body text. I've also changed the config of `linger`. 
I also removed the sentence about too many messages in the buffer so they can't all be flushed. I'm assuming the other references to flush -- in the context of handling large volumes of messages -- are okay

documentation/modules/configuring/con-config-mirrormaker2-high-volume-messages.adoc
```diff
@@ -38,11 +38,7 @@ While this is happening, Kafka Connect tasks continue to poll topics in the sour
 The size of the buffer is set using the `producer.buffer.memory` property. 
 Tasks wait for a specified timeout period (`offset.flush.timeout.ms`) before the buffer is flushed. 
 This should be enough time for the sent messages to be acknowledged by the brokers and offset data committed. 
-
-NOTE: Kafka 3.1.0 and newer versions do not wait for the producer to empty the message queue before committing offsets, except during shutdown.
-
-Try to avoid a situation where there are too many messages in the buffer, so they can't all be flushed before the offset flush timeout is reached.
-The offset flush timeout period needs to be sufficient for the size of the buffer.
+Kafka does not wait for the producer to empty the message queue before committing offsets, except during shutdown.
```
fvaleri, line 41: Instead of "Kafka does not wait", I would say the "The source task does not wait".

fvaleri (commented):
Left a minor comment, but looks good.

PaulRMellor:
@fvaleri -- Updated from minor comment. Please okay if you're okay to merge
@mimaison -- any more comments?

documentation/modules/configuring/con-config-mirrormaker2-high-volume-messages.adoc
```diff
@@ -6,26 +6,54 @@
 = Handling high volumes of messages
 
 [role="_abstract"]
-If your MirrorMaker 2.0 deployment is going to be handling a high volume of messages, you might need to adjust its configuration to support it.
+If your MirrorMaker 2.0 deployment is going to be handling a high volume of messages, you might need to adjust its configuration to support it.
```
mimaison, line 9: I wonder if we should clarify what to consider a high volume of data? This could be a source cluster with average throughput but a lot of existing data.

documentation/modules/configuring/con-config-mirrormaker2-high-volume-messages.adoc
```diff
@@ -6,26 +6,54 @@
 = Handling high volumes of messages
 
 [role="_abstract"]
-If your MirrorMaker 2.0 deployment is going to be handling a high volume of messages, you might need to adjust its configuration to support it.
+If your MirrorMaker 2.0 deployment is going to be handling a high volume of messages, you might need to adjust its configuration to support it. 
+The configuration can help manage throughput and issues with latency.
```
mimaison, line 10: Could we merge this sentence with the one above? It looks a bit strange bit itself

documentation/modules/configuring/con-config-mirrormaker2-high-volume-messages.adoc
```diff
@@ -6,26 +6,54 @@
...
 
 [role="_abstract"]
-If your MirrorMaker 2.0 deployment is going to be handling a high volume of messages, you might need to adjust its configuration to support it.
+If your MirrorMaker 2.0 deployment is going to be handling a high volume of messages, you might need to adjust its configuration to support it. 
+The configuration can help manage throughput and issues with latency. 
 
-The flush pipeline for data replication is *source topic -> (Kafka Connect) source message queue -> producer buffer -> target topic*.
-An offset flush timeout period (`offset.flush.timeout.ms`) is the time to wait for the producer buffer (`producer.buffer.memory`) to flush and offset data to be committed.
-Try to avoid a situation where a large producer buffer and an insufficient offset flush timeout period causes a _failed to flush_ or _failed to commit offsets_ type of error.
+MirrorMaker 2.0 uses an internal producer to replicate data from the source to the target Kafka cluster.
```
mimaison, line 12: It sounds like the producer is internal to MM2. The producer is from the Connect runtime.

What about something like
```
MirrorMaker 2.0 fetches data from the source cluster and hands it to the Connect runtime producers so it's replicated to the target cluster.

To control the streaming of high volumes of data, you can tune the Connect runtime producers configuration for optimal throughput.
```

documentation/modules/configuring/con-config-mirrormaker2-high-volume-messages.adoc
```diff
@@ -6,26 +6,54 @@
...
-An offset flush timeout period (`offset.flush.timeout.ms`) is the time to wait for the producer buffer (`producer.buffer.memory`) to flush and offset data to be committed.
-Try to avoid a situation where a large producer buffer and an insufficient offset flush timeout period causes a _failed to flush_ or _failed to commit offsets_ type of error.
+MirrorMaker 2.0 uses an internal producer to replicate data from the source to the target Kafka cluster. 
+To control the streaming of high volumes of data, you can tune the producer batch size configuration for optimal throughput. 
 
-This type of error means that there are too many messages in the producer buffer, so they can't all be flushed before the offset flush timeout is reached.
+In certain circumstances, you might want to increase the size of the message batches sent in a single produce request.
+You increase the batch size using the `batch.size` producer configuration to reduce the number of outstanding messages ready to be sent.
+Messages being sent to the same partition are batched together.
+A larger batch size reduces the number of requests and the size of the backlog in the message queue.
```
mimaison, line 18: This says pretty much the same as line 16. Should we remove one of them?

documentation/modules/configuring/con-config-mirrormaker2-high-volume-messages.adoc
```diff
@@ -6,26 +6,54 @@
...
+By increasing the batch size, produce requests are delayed and more messages are added to the batch and sent to brokers at the same time.  
+This can improve throughput when you have just a few topic partitions that handle large numbers of messages.  
 
-If you are getting this type of error, try the following configuration changes:
+Consider the number and size of the records that the producer handles for a suitable producer batch size. 
 
-* Decreasing the default value in bytes of the `producer.buffer.memory`
-* Increasing the default value in milliseconds of the `offset.flush.timeout.ms`
+Use `linger.ms` to add a wait time in milliseconds to delay produce requests when producer load decreases. 
+What the delay means is that more records can be added to batches if they are under the maximum batch size.
```
mimaison, line 26: I find sentences starting with `What` harder to read. What about:
```
The delay means that more records ....
```

documentation/modules/configuring/con-config-mirrormaker2-high-volume-messages.adoc
```diff
@@ -6,26 +6,54 @@
...
+.Data replication pipeline
+*source topic -> (Kafka Connect tasks) source message queue -> producer buffer -> target topic*.
 
-The changes should help to keep the underlying Kafka Connect queue of outstanding messages at a manageable size.
-You might need to adjust the values to have the desired effect.
+The producer sends messages in its buffer to topics in the target Kafka cluster.
+While this is happening, Kafka Connect tasks continue to poll topics in the source Kafka cluster to add messages to the source message queue.
 
-If these configuration changes don't resolve the error, you can try increasing the number of tasks that run in parallel by doing the following.
+The size of the buffer is set using the `producer.buffer.memory` property.
```
mimaison, line 38: Depends where it's set from. If it's from the connector config, you must use `producer.override.buffer.memory`.

documentation/modules/configuring/con-config-mirrormaker2-high-volume-messages.adoc
```diff
@@ -6,26 +6,54 @@
...
+The producer sends messages in its buffer to topics in the target Kafka cluster.
+While this is happening, Kafka Connect tasks continue to poll topics in the source Kafka cluster to add messages to the source message queue.
 
-If these configuration changes don't resolve the error, you can try increasing the number of tasks that run in parallel by doing the following.
+The size of the buffer is set using the `producer.buffer.memory` property. 
+Tasks wait for a specified timeout period (`offset.flush.timeout.ms`) before the buffer is flushed. 
+This should be enough time for the sent messages to be acknowledged by the brokers and offset data committed. 
+The source task does not wait for the producer to empty the message queue before committing offsets, except during shutdown.
+
+If the producer is unable to keep up with the throughput of messages in the source message queue, buffering is blocked until there is space available in the buffer within a time period allocated by `max.block.ms`.
```
mimaison, line 43: I'd replace `allocated` by `bounded`

documentation/modules/configuring/con-config-mirrormaker2-high-volume-messages.adoc
PaulRMellor: I wasn't sure if we needed this. But I guess an example or two might help.

I added:

> Operations like tracking web activity can generate a high volume of messages. 
> But even a source cluster with moderate throughput can produce a high volume of messages if it has a lot of existing data.

documentation/modules/configuring/con-config-mirrormaker2-high-volume-messages.adoc
PaulRMellor: :+1: 
Works better with the new examples too

documentation/modules/configuring/con-config-mirrormaker2-high-volume-messages.adoc
PaulRMellor: And this is probably where we should set it. 
I changed to:

> The size of the buffer is set using the `producer.override.buffer.memory` property for the source connector.

documentation/modules/configuring/con-config-mirrormaker2-high-volume-messages.adoc
mimaison: I should have said `affirmative sentences`! Otherwise why would I use a sentence starting with `What` just after saying I don't like them 🤣

documentation/modules/configuring/con-config-mirrormaker2-high-volume-messages.adoc
```diff
@@ -6,24 +6,26 @@
 = Handling high volumes of messages
 
 [role="_abstract"]
-If your MirrorMaker 2.0 deployment is going to be handling a high volume of messages, you might need to adjust its configuration to support it. 
-The configuration can help manage throughput and issues with latency. 
+If your MirrorMaker 2.0 deployment is going to be handling a high volume of messages, you might need to adjust its configuration configuration to help manage throughput and issues with latency.
```
fvaleri, line 9: configuration configuration -> configuration

PR #7020, opened 2022-07-06
Helm: Using release name instead of hard code name

haoxins:
### Type of change

_Select the type of your PR_

- Enhancement / new feature

### Description

Using Release name instead of the hard code name in the Helm chart

### Checklist

- [ ] Write tests
- [ ] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/packaging/helm-charts/helm3/strimzi-kafka-operator/templates/010-ServiceAccount-strimzi-cluster-operator.yaml
+++ b/packaging/helm-charts/helm3/strimzi-kafka-operator/templates/010-ServiceAccount-strimzi-cluster-operator.yaml
@@ -1,11 +1,10 @@
 apiVersion: v1
 kind: ServiceAccount
 metadata:
-  name: strimzi-cluster-operator
+  name: {{ .Release.Name }}
   labels:
     app: {{ template "strimzi.name" . }}
     chart: {{ template "strimzi.chart" . }}
     component: service-account
     release: {{ .Release.Name }}
     heritage: {{ .Release.Service }}
-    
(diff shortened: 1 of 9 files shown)


scholzj (commented):
Thanks for the PR. Some quick comments:
* All changes done to the Helm Chart outside of the regular releae have to be done to ther `packaging/helm-charts` directory. The `helm-charts` changes only during a release. So you would need to update your PR.
* That said, I do not think your changes work anyway. The Operator is creating binding for some of these roles using the fixed names. So when you change the names, you break the operator. So I do not think this can done.
* It would be great to have a proper description explaining also the motivation of this and what does it help with.

scholzj:
Well, as I said, it will not work anyway. The roles need to have their original names. 

Even if it worked, I'm also not sure why will the names be more meaningful this way. Your system will be full of resources for different versions which will IMHO not get deleted. I also don't understand how would upgrade work -> who will deelte the old deployment for example?

haoxins:
> I also don't understand how would upgrade work -> who will deelte the old deployment for example?

Yes, we can't, but this will be useful for the new users.
This is a broken change anyway.

scholzj:
That could make the Helm chart configurable but backwards compatible, yes. But it will not solve the issue that the operator will not work. You would need to extend the whole operator code to be ready for this as wel - to have the (cluster)role names configurable inside the operator. So it would be a a bit more complicated than just the Helm Chart.

haoxins:
> But it will not solve the issue that the operator will not work. 

I'm new to `Strimzi`, after take a glance at the code and pods, I know what you said.

Will consider to contribute this feature in the future.

PR #7025, opened 2022-07-07
docs: adds section that describes the MM2 producers and consumers

PaulRMellor:

**Documentation**
Adds a new section to the [Kafka MirrorMaker cluster configuration](https://strimzi.io/docs/operators/in-development/configuring.html#assembly-deployment-configuration-kafka-mirror-maker-str) section of the _Configuring_ guide

- Describes the MM2 producers and consumers for each of the MM2 connectors
- Shows where you configure the producers in the MM2 resource
- Provides an example showing possible configurations

### Checklist

- [ ] Write tests
- [x] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/documentation/modules/configuring/con-config-mirrormaker2-producers-consumers.adoc
+++ b/documentation/modules/configuring/con-config-mirrormaker2-producers-consumers.adoc
@@ -54,7 +54,7 @@ The following tables describe the producers and consumers for each of the connec
 
 |===
 
-IMPORTANT: The consumer configuration changes depending on the location of the `offset-syncs` topic. The default location is the source Kafka cluster. You can set `offset-syncs.topic.location` to `target` to use the target Kafka cluster. If you use the target Kafka cluster, use `mirrors.checkpointConnector.config: target.consumer.*` for the consumer configuration. 
+NOTE: You can set `offset-syncs.topic.location` to `target` to use the target Kafka cluster as the location of the `offset-syncs` topic. 
 
 .Heartbeat connector producer
 [cols="1,1a,3m",options="header"]
@@ -69,6 +69,8 @@ IMPORTANT: The consumer configuration changes depending on the location of the `
 |mirrors.heartbeatConnector.config: producer.override.*
 
 |===
+
+The following example shows how you configure the producers and consumers. 
 
 .Example configuration for connector producers and consumers
 [source,yaml,subs="+quotes,attributes"]
@@ -87,18 +89,18 @@ spec:
       tasksMax: 5
       config:
         producer.override.batch.size: 327680
-        producer.override.linger.ms: 5
-        producer.request.timeout.ms: 40000
-        consumer.fetch.max.bytes: 73400320
+        producer.override.linger.ms: 100
+        producer.request.timeout.ms: 30000
+        consumer.fetch.max.bytes: 52428800
         # ...
     checkpointConnector:
       config:
-        producer.override.request.timeout.ms: 40000
-        consumer.max.poll.interval.ms: 180000
+        producer.override.request.timeout.ms: 30000
+        consumer.max.poll.interval.ms: 300000
         # ...
     heartbeatConnector:
       config:
-        producer.override.request.timeout.ms: 40000
+        producer.override.request.timeout.ms: 30000
         # ...      
 ----
 
--- a/documentation/modules/configuring/con-config-mirrormaker2-producers-consumers.adoc
+++ b/documentation/modules/configuring/con-config-mirrormaker2-producers-consumers.adoc
@@ -6,7 +6,7 @@
 = Connector producer and consumer configuration
 
 [role="_abstract"]
-MirrorMaker 2.0 connectors uses internal producers and consumers.
+MirrorMaker 2.0 connectors use internal producers and consumers.
 If needed, you can configure these producers and consumers to override the default settings. 
 
 For example, you can increase the `batch.size` for the source producer that sends topics to the target Kafka cluster to better accommodate large volumes of messages.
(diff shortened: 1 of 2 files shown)


documentation/modules/configuring/con-config-mirrormaker2-producers-consumers.adoc
```diff
@@ -0,0 +1,110 @@
+// Module included in the following assemblies:
+//
+// assembly-config-mirrormaker2.adoc
+
+[id='con-mirrormaker-producers-consumers-{context}']
+= Connector producer and consumer configuration
+
+[role="_abstract"]
+MirrorMaker 2.0 connectors uses internal producers and consumers.
+If needed, you can configure these producers and consumers to override the default settings.
```
PaulRMellor, line 10: Maybe we need to say something about config exceptions?

documentation/modules/configuring/con-config-mirrormaker2-producers-consumers.adoc
```diff
@@ -0,0 +1,110 @@
...
+        producer.override.request.timeout.ms: 40000
+        consumer.max.poll.interval.ms: 180000
+        # ...
+    heartbeatConnector:
+      config:
+        producer.override.request.timeout.ms: 40000
+        # ...      
+----
+
+NOTE: You can configure producers and consumers at the cluster level using `spec.clusters.config.producer.\*` or `spec.clusters.config.consumer.*` properties.
```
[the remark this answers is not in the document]
tyamashi-oss: I think it would be better to remove it. 

`spec.cluster.config:producer.*` affects not only MirrorSourceConnector’s producer but also MirrorCheckpointConnector’s and MirrorHeartbeatConnector’s one, It can be described by explaining which producers are affected, however, it is somewhat hard to understand and there will be no use case where the same value is set for these three producers.

`spec.cluster.config:consumer.*` affects only three consumers, which load data in xxxBackingStore,  in Kafka Connect(, not xxxConnector) in MirrorMaker2. It is trivial to configure and does not need to be described.

documentation/modules/configuring/con-config-mirrormaker2-producers-consumers.adoc
tyamashi-oss: This statement appears to be fine. I think it is safe to leave.

documentation/modules/configuring/con-config-mirrormaker2-producers-consumers.adoc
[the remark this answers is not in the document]
PaulRMellor: I removed this note after discussing with @tyamashi-oss

PaulRMellor:
@tyamashi-oss -- do you have and further comments or are we ready to merge?
@mimaison -- do you have any comments for this section on MM2 producer and consumers?

documentation/modules/configuring/con-config-mirrormaker2-producers-consumers.adoc
```diff
@@ -0,0 +1,109 @@
...
+|Emits consumer offset checkpoints.
+|mirrors.checkpointConnector.config: producer.override.* 
+
+|Consumer
+|Loads the `offset-syncs` topic.
+|mirrors.checkpointConnector.config: consumer.*
+
+|===
+
+IMPORTANT: The consumer configuration changes depending on the location of the `offset-syncs` topic. The default location is the source Kafka cluster. You can set `offset-syncs.topic.location` to `target` to use the target Kafka cluster. If you use the target Kafka cluster, use `mirrors.checkpointConnector.config: target.consumer.*` for the consumer configuration.
```
tyamashi-oss, line 57: I think we can remove this sentence. The following is the same as described in [ENTMQST-4116](https://issues.redhat.com/browse/ENTMQST-4116): 
```
I think the users should use only `mirrors.checkpointConnector.config: consumer.*` to configure the consumer in MirrorCheckpointConnector, and they should not use {source_or_target} prefix.
There is no other consumers in checkpointConnector, and it would be misleading to use different source and target depending on the value of offset-syncs.topic.location, so it is not necessary to specify further whether it is source or target cluster.

I think it is sufficient to explain that only `mirrors.checkpointConnector.config: consumer.*` is used to configure the consumer in MirrorCheckpointConnector.
```

documentation/modules/configuring/con-config-mirrormaker2-producers-consumers.adoc
```diff
@@ -0,0 +1,109 @@
...
+|Configuration
+
+|Producer
+|Emits heartbeats.
+|mirrors.heartbeatConnector.config: producer.override.*
+
+|===
+
+.Example configuration for connector producers and consumers
+[source,yaml,subs="+quotes,attributes"]
```
tyamashi-oss, line 74: A certain number of users try to use the configurations described in the documentation as they are considering them to be the recommended values (somewhat surprisingly). It might be safer to simply use the default values for each configurations in this example. 

I think we can use the sourceConnector configurations discussed in https://github.com/strimzi/strimzi-kafka-operator/pull/7006. However, it would be safe to simply use the default values for checkpointConnector and HeartbeatConnector configuration. Here are the default values:
```
request.timeout.ms: 30000
max.poll.interval.ms: 300000
```

mimaison: Yes the default values are suitable in many cases. I think this example should be described as "this is how you do it" rather than "here are some good values you should use"

documentation/modules/configuring/con-config-mirrormaker2-producers-consumers.adoc
PaulRMellor: I thought it might be useful to mention `offset-syncs.topic.location` here, so changed to:

> NOTE: You can set `offset-syncs.topic.location` to `target` to use the target Kafka cluster as the location of the `offset-syncs` topic.

documentation/modules/configuring/con-config-mirrormaker2-producers-consumers.adoc
PaulRMellor: I changed to the default values and introduced the example with the following:

> The following example shows how you configure the producers and consumers.

PR #7029, opened 2022-07-07
Future-proof Cruise Control integration for Strimzi PodSet migration

kyguy:
### Type of change

- Enhancement / new feature

### Description

Future-proofing Cruise Control integration for Strimzi PodSet migration. The Cruise Control capacity configuration generated for a cluster with a JBOD configuration requires a distinct broker capacity entry for every broker because the Kafka volume paths are not homogeneous across brokers and include the broker pod index in their names. The changes in this PR ensure that the proper broker indexes will be used even if/when we switch to using non-sequential broker ids as part of the PodSet migration.

### Checklist

- [ ] Write tests
- [ ] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/resource/KafkaRoller.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/resource/KafkaRoller.java
@@ -392,7 +392,7 @@ public class KafkaRoller {
         return false;
     }
 
-    private int idOfPod(String podName)  {
+    public static int idOfPod(String podName)  {
         return Integer.parseInt(podName.substring(podName.lastIndexOf("-") + 1));
     }
 
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/model/CruiseControl.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/model/CruiseControl.java
@@ -204,7 +204,7 @@ public class CruiseControl extends AbstractModel {
 
             // To avoid illegal storage configurations provided by the user,
             // we rely on the storage configuration provided by the KafkaAssemblyOperator
-            cruiseControl.capacity = new Capacity(kafkaCr.getSpec(), storage);
+            cruiseControl.capacity = new Capacity(reconciliation, kafkaCr.getSpec(), storage);
 
             // Parse different types of metrics configurations
             ModelUtils.parseMetrics(cruiseControl, ccSpec);
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/model/cruisecontrol/Capacity.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/model/cruisecontrol/Capacity.java
@@ -14,9 +14,12 @@ import io.strimzi.api.kafka.model.storage.PersistentClaimStorage;
 import io.strimzi.api.kafka.model.storage.SingleVolumeStorage;
 import io.strimzi.api.kafka.model.storage.Storage;
 import io.strimzi.operator.cluster.model.AbstractModel;
+import io.strimzi.operator.cluster.model.KafkaCluster;
 import io.strimzi.operator.cluster.model.StorageUtils;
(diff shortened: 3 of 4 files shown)


cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/resource/KafkaRoller.java
```diff
@@ -392,7 +392,7 @@ private boolean podWaitingBecauseOfAnyReasons(Pod pod, Set<String> reasons) {
         return false;
     }
 
-    private int idOfPod(String podName)  {
+    public static int idOfPod(String podName)  {
```
scholzj, line 395: I think that instead of making this public and sharing the KafkaRoller, you should move it to some `*Util` class.

scholzj (commented):
One nit, but LGTM otherwise.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/resource/KafkaRoller.java
scholzj: Is there any reason why it should be in `operator-common`? It not, `ModelUtils` in `cluster-operator` would be IMHO better.

kyguy: Just moved to `ModelUtils`

scholzj (approved):
LGTM. The regression tests passed, only storing the logs failed.

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/ModelUtils.java
```diff
@@ -739,4 +739,15 @@ public static boolean hasOwnerReference(HasMetadata resource, OwnerReference own
     public static int caCertGeneration(Ca ca) {
         return Annotations.intAnnotation(ca.caCertSecret(), Ca.ANNO_STRIMZI_IO_CA_CERT_GENERATION, Ca.INIT_GENERATION);
     }
+
+    /**
+     * Returns the id of a pod given the pod name
+     *
```
ShubhamRwt, line 745: Is this space required?

ShubhamRwt (approved):

ShubhamRwt:
LGTM

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/ModelUtils.java
scholzj: I think we normally have an empty lines between the description block, params block and return.

ppatierno (approved):
LGTM

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/ModelUtils.java
```diff
@@ -739,4 +739,15 @@ public static boolean hasOwnerReference(HasMetadata resource, OwnerReference own
...
+     * @param podName   Name of pod
+     *
+     * @return          Id of the pod
+     */
+    public static int idOfPod(String podName)  {
+        return Integer.parseInt(podName.substring(podName.lastIndexOf("-") + 1));
```
tomncooper, line 751: Seem to remember that the PodSets proposal suggested that the brokers pods could use any id scheme (a, b, c or 1st, 2nd, 3rd)? Is there any chance these ids won't be an int? Does it matter if they stay a string?

scholzj: The ultimate Kafka broker ID has to be int AFAIK. So there will be a way how to convert it to int.

tomncooper: Thanks Jakub, thought that might be the case, I was just being pedantic.

kyguy:
@tomncooper  It hasn't been decided yet but I would assume it would be an int!

PR #7034, opened 2022-07-08
Bump CC version to pick up NumCore type change in BrokerStats

kyguy:
### Type of change

- Bugfix

### Description

Related to recent Cruise Control CPU capacity work, upstream Cruise Control changed the CPU core type from int -> double to accommodate millicore values but missed the change in the BrokerStats which causes the CPU values to be truncated. This PR is to bump to the CC release which  fixes that

### Checklist

- [ ] Write tests
- [ ] Update documentation
- [ ] Update CHANGELOG.md


ppatierno (approved):
LGTM. Tests seem to be fine.

PR #7067, opened 2022-07-16
Add default value for kubernetes-client backwards compatibility

fvaleri:
The default value of kubernetes-client backwards compatibility has recently been changed:
https://github.com/fabric8io/kubernetes-client/blob/a5285fbefb1781c290052b35670a796e06110174/doc/MIGRATION-v6.md#backwards-compatibility-interceptor

This commit enables kubernetes-client backwards compatibility by default to avoid a change in behavior when switching to 6.x.
The kubernetes-client specific configuration is set as CO system property and passed as environment variable to all operands.

We use the value of KUBERNETES_BACKWARDSCOMPATIBILITYINTERCEPTOR_DISABLE enviroment variable if valid and present in CO deployment at install time.

### DIFF
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/ClusterOperatorConfig.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/ClusterOperatorConfig.java
@@ -411,6 +411,24 @@ public class ClusterOperatorConfig {
     }
 
     /**
+     * Retrieve kubernetes-client backwards compatibility configuration.
+     *
+     * @param env true to get the environment rather than system configuration
+     *
+     * @return backwards compatibility configuration
+     */
+    public static String[] getKubeClientBackwardsCompatibilityConfig(boolean env) {
+        String[] config = {
+            env ? "KUBERNETES_BACKWARDSCOMPATIBILITYINTERCEPTOR_DISABLE"
+                    : "kubernetes.backwardsCompatibilityInterceptor.disable",
+                System.getenv().getOrDefault(
+                        "KUBERNETES_BACKWARDSCOMPATIBILITYINTERCEPTOR_DISABLE",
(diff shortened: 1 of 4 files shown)


scholzj:
Sorry, but why is this needed? I this we want to follow the Fabric8 defaults and not keep the old behavior. Plus anyone who wants to configure can configure it already today without any addiitonal code complexity.

fvaleri:
> Sorry, but why is this needed? I this we want to follow the Fabric8 defaults and not keep the old behavior. Plus anyone who wants to configure can configure it already today without any addiitonal code complexity.

Today, you have to set that var in all operator deployments in order to have consistency. With this you only do it in the CO deployment. Anyway, I'm not against following the new kubernetes-client default, so we can close this.

scholzj:
> Today, you have to set that var in all operator deployments in order to have consistency. With this you only do it in the CO deployment. Anyway, I'm not against following the new kubernetes-client default, so we can close this.

Sorry, but I do not think this explains why is this needed. What value does it add? This is IMHO just complexity and code without any added value or use-case. I even doubt any operands actually use the interceptor.

fvaleri:
> > Today, you have to set that var in all operator deployments in order to have consistency. With this you only do it in the CO deployment. Anyway, I'm not against following the new kubernetes-client default, so we can close this.
> 
> Sorry, but I do not think this explains why is this needed. What value does it add? This is IMHO just complexity and code without any added value or use-case. I even doubt any operands actually use the interceptor.

The added value is to avoid the change in behavior when switching to 6.x, but if that change is desiderable, then I'm fine to close the PR.

scholzj:
I personally don't see any reasons why we should not follow Fabric8 and its defaults.

fvaleri:
Yeah, ok. I'm going to close it.

PR #7087, opened 2022-07-21
docs(mm2): Describes adding tasks for consumer group offset sync

PaulRMellor:

**Documentation**
Updates MM2 section [Specifying a maximum number of tasks](https://strimzi.io/docs/operators/in-development/configuring.html#con-mirrormaker-tasks-max-str) to describe how to add more tasks for checkpoint connector and why you might want to. 

### Checklist

- [ ] Write tests
- [x] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/documentation/modules/configuring/con-config-mirrormaker2-tasks-max.adoc
+++ b/documentation/modules/configuring/con-config-mirrormaker2-tasks-max.adoc
@@ -8,6 +8,7 @@
 [role="_abstract"]
 Connectors create the tasks that are responsible for moving data in and out of Kafka.
 Each connector comprises one or more tasks that are distributed across a group of worker pods that run the tasks.
+Increasing the number of tasks can help with performance issues when replicating a large number of partitions or synchronizing the offsets of a large number of consumer groups. 
 
 Tasks run in parallel.
 Workers are assigned one or more tasks.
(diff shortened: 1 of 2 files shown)


PaulRMellor:
I add some info on metrics currently available from the dashboard  that might be useful.
I don't know if we want to say anything on logs or pulling info from REST API.
Can any connector/task info be returned using _my-mm2-cluster-mirrormaker2-api_, like with _my-connect-cluster-connect-api_ with Kafka Connect cluster?

documentation/modules/configuring/con-config-mirrormaker2-tasks-max.adoc
```diff
@@ -16,9 +17,19 @@ If there are more tasks than workers, workers handle multiple tasks.
...
 You can specify the maximum number of connector tasks in your MirrorMaker configuration using the `tasksMax` property.
 Without specifying a maximum number of tasks, the default setting is a single task.
-If the infrastructure supports the processing overhead, increasing the number of tasks can improve throughput.
 
-.tasksMax configuration for a MirrorMaker connector
+The heartbeat connector always uses a single task.
+
+The number of tasks that are started for the source and checkpoint connectors is the lower value between the maximum number of possible tasks and the value for `tasksMax`.
+For the source connector, the maximum number of tasks possible is one for each partition being replicated from the source cluster.
+For the checkpoint connector, the maximum number of tasks possible is one for each consumer group being replicated from the source cluster.
```
scholzj, line 25: What are the _possible tasks_ here? What determines how many tasks are possible? Do we explain it somewhere?

PaulRMellor: We're suggesting that the max possible is the lower value of one for each partition or the value set for `tasksMax` value, which defaults to 1. 
But, as Kafka suggest, the value used depends _on the available hardware resources and the total number of topic-partitions to be replicated._

scholzj (approved):

fvaleri:
In the discussion for this PR we also noted the following that may be worth adding:

> By default MM2 refreshes groups every 10 minutes, but some user set this as low as 30 seconds. In that case an application regularly generating dynamic groups can really impact the connector.

mimaison (approved):

PaulRMellor:
> In the discussion for this PR we also noted the following that may be worth adding:
> 
> > By default MM2 refreshes groups every 10 minutes, but some user set this as low as 30 seconds. In that case an application regularly generating dynamic groups can really impact the connector.

Thanks @fvaleri  -- I've added a paragraph on this

fvaleri:
LGTM.

Thanks @PaulRMellor

PR #7089, opened 2022-07-21
Pluggable Pod Security Providers implementation

scholzj:
### Type of change

- Enhancement / new feature

### Description

This PR implements the Strimzi Proposal [SP#37](https://github.com/strimzi/proposals/blob/main/037-pluggable-pod-security-profiles.md) and adds a mechanism for pluggable PodSecurityProviders. 

The only notable difference between the proposal and the PR is in the `PodSecurityProviderContext` interface. Only during the implementation I realised that the user supplied context has two different types. I decided to create two interfaces:

```java
interface PodSecurityProviderContext {
    Storage storage();
    PodSecurityContext userSuppliedContext();
}
```

and 

```java
interface ContainerSecurityProviderContext {
    Storage storage();
    SecurityContext userSuppliedContext();
}
```

_(The other option was to have a single interface with two methods where one would always return `null`, but two interfaces seemed better.)_

This PR provides only minimal documentation. Full documentation will be done in a a separate PR.

### Checklist

- [x] Write tests
- [x] Update documentation
- [x] Update CHANGELOG.md


ppatierno:
I had a first quick pass but should we have Javadoc on new interfaces and related methods at least?

scholzj:
> I had a first quick pass but should we have Javadoc on new interfaces and related methods at least?

Good point. I didn't wanted to waste time with them in the prototype before the proposal was approved and then I forgot them :-/. I will add them.

api/src/main/java/io/strimzi/plugin/security/profiles/ContainerSecurityProviderContext.java
```diff
@@ -0,0 +1,13 @@
+/*
+ * Copyright Strimzi authors.
+ * License: Apache License 2.0 (see the file LICENSE or http://apache.org/licenses/LICENSE-2.0.html).
+ */
+package io.strimzi.plugin.security.profiles;
+
+import io.fabric8.kubernetes.api.model.SecurityContext;
+import io.strimzi.api.kafka.model.storage.Storage;
+
+public interface ContainerSecurityProviderContext {
```
tombentley, line 10: Can you add some Javadocs?

api/src/main/java/io/strimzi/plugin/security/profiles/PodSecurityProviderContext.java
```diff
@@ -0,0 +1,13 @@
+/*
+ * Copyright Strimzi authors.
+ * License: Apache License 2.0 (see the file LICENSE or http://apache.org/licenses/LICENSE-2.0.html).
+ */
+package io.strimzi.plugin.security.profiles;
+
+import io.fabric8.kubernetes.api.model.PodSecurityContext;
+import io.strimzi.api.kafka.model.storage.Storage;
+
+public interface PodSecurityProviderContext {
```
tombentley, line 10: Can you add some Javadocs?

api/src/main/java/io/strimzi/plugin/security/profiles/PodSecurityProvider.java
```diff
@@ -0,0 +1,125 @@
...
+ * Copyright Strimzi authors.
+ * License: Apache License 2.0 (see the file LICENSE or http://apache.org/licenses/LICENSE-2.0.html).
+ */
+package io.strimzi.plugin.security.profiles;
+
+import io.fabric8.kubernetes.api.model.PodSecurityContext;
+import io.fabric8.kubernetes.api.model.SecurityContext;
+import io.strimzi.platform.PlatformFeatures;
+
+public interface PodSecurityProvider {
```
tombentley, line 11: Can you add some Javadocs?

api/src/main/java/io/strimzi/plugin/security/profiles/ContainerSecurityProviderContext.java
tombentley: And possibly a `package-info.java` to summarize the purpose of the package (since it's in `api`).

tombentley (approved):

scholzj:
> Anyway I see that with a lot of `if` statement you have the opening brackets far away from the condition ... more than just one space, maybe something bad with your IDE?

I think I'm always using tab and not a single space. I do not have a strong preference, but if we want to use just a single space, we need to have a checkstyle for it.

PR #7131, opened 2022-07-31
Add support for Kafka 3.2.1

scholzj:
### Type of change

- Enhancement / new feature

### Description

This PR adds support for Kafka 3.2.1. It is used as a new default version. It also updates all the examples to use 3.2.1 as well.

### Checklist

- [x] Update documentation
- [x] Update CHANGELOG.md


ppatierno (approved):
I am fine with it, was just wondering if we could get rid of at least 3.1.1 version support or maybe even 3.1.0 (to avoid having 4 Kafka versions). For people coming from 0.29.0 is it not feasible for them to upgrade to Kafka 3.1.1 or 3.2.0 first and before moving to Strimzi 0.30.0?

scholzj:
> I am fine with it, was just wondering if we could get rid of at least 3.1.1 version support or maybe even 3.1.0 (to avoid having 4 Kafka versions). For people coming from 0.29.0 is it not feasible for them to upgrade to Kafka 3.1.1 or 3.2.0 first and before moving to Strimzi 0.30.0?

We [promise](https://github.com/strimzi/strimzi-kafka-operator/blob/main/KAFKA_VERSION_SUPPORT.md) to support at least two last minor versions. I guess in theory we can remove 3.1.0 and keep only 3.1.1. But normally we keep both of them until 3.3.0 is released. And then we drop both 3.1.0 and 3.1.1 support.

xin-hao-awx:
Maybe we should cherry-pick this commit into branch `release-0.30.x` ?

scholzj:
I will try to find some time to see if it can be cherry-picked. But we are currently not planning to include this in any new patch release.

PR #7164, opened 2022-08-05
Add user-agent to Kubernetes Client (#7163)

### LINKED ISSUE #7163: Strimzi should identify itself via its User-Agent in calls to the Kubernetes API
I am trying to determine which system is calling certain K8s APIs to see who is using a deprecated API before I do a K8s upgrade. The only data available about the caller is the User-Agent.

Add the strimzi component name and version to the User-Agent string used by the Kubernetes client.

_None considered_

**Additional context**
_N/A_

adamzr:
### Type of change

- Enhancement / new feature

### Description

Set User-Agent in Kubernetes Clients to identify itself with the following information:

- That we are using the Fabric8 client.
- The version of the Fabric8 client that we are using.
- The name of the Strimzi component.
- The version of the Strimzi component.

Fixes #7163 

### Checklist

- [x] Write tests
- [x] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/kafka-init/src/main/java/io/strimzi/kafka/init/Main.java
+++ b/kafka-init/src/main/java/io/strimzi/kafka/init/Main.java
@@ -4,11 +4,9 @@
  */
 package io.strimzi.kafka.init;
 
-import io.fabric8.kubernetes.client.Config;
-import io.fabric8.kubernetes.client.ConfigBuilder;
 import io.fabric8.kubernetes.client.KubernetesClient;
-import io.fabric8.kubernetes.client.KubernetesClientBuilder;
-import io.fabric8.kubernetes.client.Version;
+import io.strimzi.operator.common.OperatorKubernetesClientBuilder;
+
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 
@@ -22,10 +20,7 @@ public class Main {
         LOGGER.info("Init-kafka {} is starting", strimziVersion);
         InitWriterConfig config = InitWriterConfig.fromMap(System.getenv());
 
-        final String userAgent = "fabric8-kubernetes-client/" + Version.clientVersion() 
-                                + " strimzi-init-kafka/" + strimziVersion;
-        final Config kubernetesClientConfig = new ConfigBuilder().withUserAgent(userAgent).build();
-        KubernetesClient client = new KubernetesClientBuilder().withConfig(kubernetesClientConfig).build();
+        final KubernetesClient client = new OperatorKubernetesClientBuilder("strimzi-kafka-init", strimziVersion).build();
 
         LOGGER.info("Init-kafka started with config: {}", config);
 
--- a/kafka-init/src/main/java/io/strimzi/kafka/init/Main.java
+++ b/kafka-init/src/main/java/io/strimzi/kafka/init/Main.java
@@ -4,8 +4,11 @@
  */
 package io.strimzi.kafka.init;
 
+import io.fabric8.kubernetes.client.Config;
+import io.fabric8.kubernetes.client.ConfigBuilder;
 import io.fabric8.kubernetes.client.KubernetesClient;
 import io.fabric8.kubernetes.client.KubernetesClientBuilder;
+import io.fabric8.kubernetes.client.Version;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 
@@ -15,10 +18,14 @@ public class Main {
 
     public static void main(String[] args) {
 
-        LOGGER.info("Init-kafka {} is starting", Main.class.getPackage().getImplementationVersion());
+        final String strimziVersion = Main.class.getPackage().getImplementationVersion();
+        LOGGER.info("Init-kafka {} is starting", strimziVersion);
         InitWriterConfig config = InitWriterConfig.fromMap(System.getenv());
 
-        KubernetesClient client = new KubernetesClientBuilder().build();
+        final String userAgent = "fabric8-kubernetes-client/" + Version.clientVersion() 
+                                + " strimzi-init-kafka/" + strimziVersion;
+        final Config kubernetesClientConfig = new ConfigBuilder().withUserAgent(userAgent).build();
+        KubernetesClient client = new KubernetesClientBuilder().withConfig(kubernetesClientConfig).build();
 
         LOGGER.info("Init-kafka started with config: {}", config);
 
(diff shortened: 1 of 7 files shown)


kafka-init/src/main/java/io/strimzi/kafka/init/Main.java
```diff
@@ -15,10 +18,14 @@ public class Main {
...
-        LOGGER.info("Init-kafka {} is starting", Main.class.getPackage().getImplementationVersion());
+        final String strimziVersion = Main.class.getPackage().getImplementationVersion();
+        LOGGER.info("Init-kafka {} is starting", strimziVersion);
         InitWriterConfig config = InitWriterConfig.fromMap(System.getenv());
 
-        KubernetesClient client = new KubernetesClientBuilder().build();
+        final String userAgent = "fabric8-kubernetes-client/" + Version.clientVersion() 
+                                + " strimzi-init-kafka/" + strimziVersion;
+        final Config kubernetesClientConfig = new ConfigBuilder().withUserAgent(userAgent).build();
+        KubernetesClient client = new KubernetesClientBuilder().withConfig(kubernetesClientConfig).build();
```
scholzj, line 28: Would it make sense to have this in one place in some _util_ class inside the `operator-common` module?

kafka-init/src/main/java/io/strimzi/kafka/init/Main.java
```diff
@@ -15,10 +18,14 @@ public class Main {
 
     public static void main(String[] args) {
 
-        LOGGER.info("Init-kafka {} is starting", Main.class.getPackage().getImplementationVersion());
+        final String strimziVersion = Main.class.getPackage().getImplementationVersion();
+        LOGGER.info("Init-kafka {} is starting", strimziVersion);
         InitWriterConfig config = InitWriterConfig.fromMap(System.getenv());
 
-        KubernetesClient client = new KubernetesClientBuilder().build();
+        final String userAgent = "fabric8-kubernetes-client/" + Version.clientVersion()
```
scholzj, line 25: Why have you chosen the user agent to look like this?
* I guess it creates overhead on every request?
* Fabric8 client version is normally based on the Strimzi version
* If you need to include Fabric8 version, why not include also the version of the HTTP client?

kafka-init/src/main/java/io/strimzi/kafka/init/Main.java
scholzj, line 28: Would it be better to avoid creating the variables in `main` which are only used once or twice at startup?

scholzj (commented):
Thanks for the PR. I left few comments.

Why is the service account information not sufficient? It identifies who made the call and from there you can find the rest normally. (I do not have problem with this change in general apart form the comments - but I'm curious :-o).

Also, it would probably make sense to do this also for the Cluster Operator which makes most of the calls? (https://github.com/strimzi/strimzi-kafka-operator/blob/main/cluster-operator/src/main/java/io/strimzi/operator/cluster/Main.java)

kafka-init/src/main/java/io/strimzi/kafka/init/Main.java
adamzr: Yes. I committed a change to do that.

kafka-init/src/main/java/io/strimzi/kafka/init/Main.java
adamzr: Good point. I changed it to just have the Strimzi component name and version. You can determine the Fabric8 and HTTP client version based on that.

adamzr:
@scholzj Thank you for the review! We were looking at a report from GKE warning us of components in the cluster using deprecated APIs. This report only has user-agent on it. It might be possible to get a better report using the service-accounts of the callers. However, having a more useful User-Agent is a good idea in my opinion.

operator-common/src/main/java/io/strimzi/operator/common/OperatorKubernetesClientBuilder.java
```diff
@@ -0,0 +1,42 @@
...
+
+import io.fabric8.kubernetes.client.Config;
+import io.fabric8.kubernetes.client.ConfigBuilder;
+import io.fabric8.kubernetes.client.KubernetesClient;
+import io.fabric8.kubernetes.client.KubernetesClientBuilder;
+
+/**
+ * Class for generating Kubernetes Clients for Operators.
+ */
+public class OperatorKubernetesClientBuilder {
```
scholzj, line 15: This looks good. But would a single static method which is just called be enough instead of having a full class for it? You can for example name it `KubernetesClientUtils` and call the static method `clientWithUserAgent` or something.

WDYT @tombentley @ppatierno?

scholzj (commented):
Did you test if the client auto-configures in the same way as before when you create the configuration like this?  E.g. when the user configures is through environment variables or system properties?

operator-common/src/main/java/io/strimzi/operator/common/OperatorKubernetesClientBuilder.java
adamzr: I'm not used to Strimzi's coding style, I didn't know what your opinion on statics and *Util classes are, and I copied the style from PasswordGenerator. Maybe that was a bad model for this.

Let me know if you prefer the Utils class with the static method and I can switch to it.

adamzr:
@scholzj I use a similar approach to configuring the user-agent in my own project and the auto-configuration works just fine.

operator-common/src/main/java/io/strimzi/operator/common/OperatorKubernetesClientBuilder.java
```diff
@@ -0,0 +1,42 @@
...
+     */
+    public OperatorKubernetesClientBuilder(final String componentName, final String version) {
+        this.componentName = componentName;
+        this.version = version;
+    }
+
+    /**
+     * Builds the KubernetesClient.
+     *
+     * @return String with new password
```
ppatierno, line 35: The Javadoc is wrong I guess :-)

operator-common/src/main/java/io/strimzi/operator/common/OperatorKubernetesClientBuilder.java
ppatierno: Tbh I like the idea of having the same approach as the `KubernetesClientBuilder` so I am fine with the way it's done now. But it's not a strong opinion, if others prefer the Util classes I am fine too.

ppatierno (approved):

operator-common/src/main/java/io/strimzi/operator/common/OperatorKubernetesClientBuilder.java
scholzj: Ok, if Paolo is fine with it lets stick with it. So I guess its just the Javadoc comment he left. Thanks for your patience.

tombentley (approved):
LGTM, thanks!

scholzj (approved):
Thanks.

PR #7165, opened 2022-08-06
Add shutdown hook to user and topic operators.

fvaleri:
This change is required to ensure that Verticle.stop() is called during a clean shutdown,
so that watches can be closed and any inflight reconciliation has a chance to complete.

I'm also moving the hook to the operator-common module as it is a shared need.

### DIFF
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/ClusterOperator.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/ClusterOperator.java
@@ -92,31 +92,27 @@ public class ClusterOperator extends AbstractVerticle {
         // Configure the executor here, but it is used only in other places
         getVertx().createSharedWorkerExecutor("kubernetes-ops-pool", config.getOperationsThreadPoolSize(), TimeUnit.SECONDS.toNanos(120));
 
-        if (config.featureGates().useStrimziPodSetsEnabled()) {
-            strimziPodSetController = new StrimziPodSetController(namespace, config.getCustomResourceSelector(), resourceOperatorSupplier.kafkaOperator, resourceOperatorSupplier.strimziPodSetOperator, resourceOperatorSupplier.podOperations, config.getPodSetControllerWorkQueueSize());
-            strimziPodSetController.start();
-        }
-
         @SuppressWarnings({ "rawtypes" })
-        List<Future> watchFutures = new ArrayList<>(8);
+        List<Future> startFutures = new ArrayList<>(8);
+        startFutures.add(maybeStartStrimziPodSetController());
 
         if (!config.isPodSetReconciliationOnly()) {
             List<AbstractOperator<?, ?, ?, ?>> operators = new ArrayList<>(asList(
                     kafkaAssemblyOperator, kafkaMirrorMakerAssemblyOperator,
                     kafkaConnectAssemblyOperator, kafkaBridgeAssemblyOperator, kafkaMirrorMaker2AssemblyOperator));
             for (AbstractOperator<?, ?, ?, ?> operator : operators) {
-                watchFutures.add(operator.createWatch(namespace, operator.recreateWatch(namespace)).compose(w -> {
+                startFutures.add(operator.createWatch(namespace, operator.recreateWatch(namespace)).compose(w -> {
                     LOGGER.info("Opened watch for {} operator", operator.kind());
                     watchByKind.put(operator.kind(), w);
                     return Future.succeededFuture();
                 }));
             }
 
-            watchFutures.add(AbstractConnectOperator.createConnectorWatch(kafkaConnectAssemblyOperator, namespace, config.getCustomResourceSelector()));
-            watchFutures.add(kafkaRebalanceAssemblyOperator.createRebalanceWatch(namespace));
+            startFutures.add(AbstractConnectOperator.createConnectorWatch(kafkaConnectAssemblyOperator, namespace, config.getCustomResourceSelector()));
+            startFutures.add(kafkaRebalanceAssemblyOperator.createRebalanceWatch(namespace));
         }
 
-        CompositeFuture.join(watchFutures)
+        CompositeFuture.join(startFutures)
                 .compose(f -> {
                     LOGGER.info("Setting up periodic reconciliation for namespace {}", namespace);
                     this.reconcileTimer = vertx.setPeriodic(this.config.getReconciliationIntervalMs(), res2 -> {
@@ -131,6 +127,23 @@ public class ClusterOperator extends AbstractVerticle {
                 .onComplete(start);
     }
 
+    public Future<Void> maybeStartStrimziPodSetController() {
+        Promise<Void> handler = Promise.promise();
+        vertx.executeBlocking(future -> {
+            try {
+                if (config.featureGates().useStrimziPodSetsEnabled()) {
+                    strimziPodSetController = new StrimziPodSetController(namespace, config.getCustomResourceSelector(), resourceOperatorSupplier.kafkaOperator,
+                            resourceOperatorSupplier.strimziPodSetOperator, resourceOperatorSupplier.podOperations, config.getPodSetControllerWorkQueueSize());
+                    strimziPodSetController.start();
+                }
+                future.complete();
+            } catch (Throwable e) {
+                LOGGER.error("StrimziPodSetController start failed");
+                future.fail(e);
+            }
+        }, handler);
+        return handler.future();
+    }
 
     @Override
     public void stop(Promise<Void> stop) {
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/Main.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/Main.java
@@ -24,7 +24,7 @@ import io.strimzi.operator.common.MetricsProvider;
 import io.strimzi.operator.common.MicrometerMetricsProvider;
 import io.strimzi.operator.common.PasswordGenerator;
 import io.strimzi.operator.common.Reconciliation;
-import io.strimzi.operator.common.VertxShutdownHook;
+import io.strimzi.operator.common.ShutdownHook;
 import io.strimzi.operator.common.Util;
 import io.strimzi.operator.common.operator.resource.ClusterRoleOperator;
 
@@ -86,7 +86,7 @@ public class Main {
                 .setJvmMetricsEnabled(true)
                 .setEnabled(true));
         Vertx vertx = Vertx.vertx(options);
-        Runtime.getRuntime().addShutdownHook(new Thread(new VertxShutdownHook(vertx)));
+        Runtime.getRuntime().addShutdownHook(new Thread(new ShutdownHook(vertx)));
 
         // Setup Micrometer Metrics provider
         MetricsProvider metricsProvider = new MicrometerMetricsProvider();
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/Main.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/Main.java
@@ -24,6 +24,7 @@ import io.strimzi.operator.common.MetricsProvider;
 import io.strimzi.operator.common.MicrometerMetricsProvider;
 import io.strimzi.operator.common.PasswordGenerator;
 import io.strimzi.operator.common.Reconciliation;
+import io.strimzi.operator.common.VertxShutdownHook;
 import io.strimzi.operator.common.Util;
 import io.strimzi.operator.common.operator.resource.ClusterRoleOperator;
 
@@ -85,10 +86,7 @@ public class Main {
                 .setJvmMetricsEnabled(true)
                 .setEnabled(true));
         Vertx vertx = Vertx.vertx(options);
-
-        // Verticle.stop() methods are not executed if you don't call Vertx.close()
-        // Vertx registers a shutdown hook for that, but only if you use its Launcher as main class
-        Runtime.getRuntime().addShutdownHook(new Thread(new ShutdownHook(vertx)));
+        Runtime.getRuntime().addShutdownHook(new Thread(new VertxShutdownHook(vertx)));
 
         // Setup Micrometer Metrics provider
         MetricsProvider metricsProvider = new MicrometerMetricsProvider();
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/Main.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/Main.java
@@ -203,7 +203,6 @@ public class Main {
                         }
                     } else {
                         LOGGER.error("Cluster Operator verticle in namespace {} failed to start", namespace, res.cause());
-                        System.exit(1);
                     }
                     prom.handle(res);
                 });
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/StrimziPodSetController.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/StrimziPodSetController.java
@@ -82,58 +82,63 @@ public class StrimziPodSetController implements Runnable {
      * @param podSetControllerWorkQueueSize Indicates the size of the StrimziPodSetController work queue
      */
     public StrimziPodSetController(String watchedNamespace, Labels crSelectorLabels, CrdOperator<KubernetesClient, Kafka, KafkaList> kafkaOperator, StrimziPodSetOperator strimziPodSetOperator, PodOperator podOperator, int podSetControllerWorkQueueSize) {
-        this.podOperator = podOperator;
-        this.strimziPodSetOperator = strimziPodSetOperator;
-        this.crSelector = (crSelectorLabels == null || crSelectorLabels.toMap().isEmpty()) ? Optional.empty() : Optional.of(new LabelSelector(null, crSelectorLabels.toMap()));
-        this.watchedNamespace = watchedNamespace;
-        this.workQueue = new ArrayBlockingQueue<>(podSetControllerWorkQueueSize);
+        try {
+            this.podOperator = podOperator;
+            this.strimziPodSetOperator = strimziPodSetOperator;
+            this.crSelector = (crSelectorLabels == null || crSelectorLabels.toMap().isEmpty()) ? Optional.empty() : Optional.of(new LabelSelector(null, crSelectorLabels.toMap()));
+            this.watchedNamespace = watchedNamespace;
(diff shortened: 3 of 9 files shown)


scholzj (commented):
I don't follow why is this needed. There is no waiting for anything to finish anyway.

fvaleri:
The UO just closes its watch, but the TO actually [waits some time for inflight reconciliations](https://github.com/strimzi/strimzi-kafka-operator/blob/main/topic-operator/src/main/java/io/strimzi/operator/topic/Session.java#L148). The problem is that this logic never execute in case of clean shutdown. We should have something similar in the UO.

scholzj:
> The UO just closes its watch, but [the TO waits some time for inflight reconciliations](https://github.com/strimzi/strimzi-kafka-operator/blob/main/topic-operator/src/main/java/io/strimzi/operator/topic/Session.java#L148). The problem is that this logic is never executed in case of clean shutdown. We should have something similar in the UO.

Well, you seem to be adding it to all operators but you say that only the Topic Operator. needs it. That seems weird.

fvaleri:
> Well, you seem to be adding it to all operators but you say that only the Topic Operator. needs it. That seems weird.

What's the point of putting any logic in Verticle.stop() if we never call it?

The CO already has it, the TO needs it. It would be good if the CO and UO could handle inflight reconciliation in a similar way.

scholzj:
> > Well, you seem to be adding it to all operators but you say that only the Topic Operator. needs it. That seems weird.
> 
> What's the point of putting any logic in Verticle.stop() if it is never called?
> 
> The CO already has it, the TO needs it. It would be good if the CO and UO could handle inflight reconciliation in a similar way.

I don't know how TO handles this. But in general, you have limited time to shutdown so there is no time to wait for reconciliation to finish. The operator is stateless -> it should be able to stop and recover in other instance at any time.

fvaleri:
I think the `terminationGracePeriodSeconds` is 30 seconds by default, which gives some chance to complete an inflight reconciliation. Of course the user can increase or decrease it as needed.

operator-common/src/main/java/io/strimzi/operator/common/VertxShutdownHook.java
```diff
@@ -2,23 +2,28 @@
...
 
 import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
 import io.vertx.core.Vertx;
-import org.apache.logging.log4j.LogManager;
-import org.apache.logging.log4j.Logger;
-
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.TimeUnit;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
```
scholzj, line 12: Can we keep the original import order?

scholzj:
PS: The shutdown hook does not seem to work well when the verticles don't start. Is always waits for the 2 minute timeout. Maybe it should be set-up only after the startup?

fvaleri:
Thanks. I'll look into that. Do you get the "Failure in stopping Vertx" error message with the stack trace? Can you post here?

scholzj:
No, I get only
```
2022-08-08 08:31:00 INFO  ShutdownHook:27 - Shutting down
```
And a lot of waiting. I did not measured it, but it could be the ~2 minutes. I guess the last message might be lost without a proper logging stack easily, so not sure about that.

scholzj:
PS: The way I run into it (while testing something else)
* Deploy the operator as usuallly
* Edit the deployment and configure it to watch two namespaces 
  * In my case one was the original one with all RBAC present and the other was just a random namespace without any RBAC rights
  * => this should send the operator crash looping because of the missing RBAC

fvaleri:
@scholzj I think the delay is due to a Verticle.stop() which never completes or is very slow to do so. As you say, this consumes all the fixed ShutdownHook timeout (2 minutes) during a clean shutdown. Let me explain.

This is the thread which is blocking `System.exit()`:

```
"vert.x-eventloop-thread-0" #24 prio=5 os_prio=0 cpu=62.20ms elapsed=113.63s tid=0x000055c33c3e6000 nid=0x28 in Object.wait()  [0x00007f53b0394000]
   java.lang.Thread.State: WAITING (on object monitor)
	at java.lang.Object.wait(java.base@11.0.16/Native Method)
	- waiting on <0x00000000fd6edbc0> (a java.lang.Thread)
	at java.lang.Thread.join(java.base@11.0.16/Thread.java:1300)
	- waiting to re-lock in wait() <0x00000000fd6edbc0> (a java.lang.Thread)
	at java.lang.Thread.join(java.base@11.0.16/Thread.java:1375)
	at java.lang.ApplicationShutdownHooks.runHooks(java.base@11.0.16/ApplicationShutdownHooks.java:107)
	at java.lang.ApplicationShutdownHooks$1.run(java.base@11.0.16/ApplicationShutdownHooks.java:46)
	at java.lang.Shutdown.runHooks(java.base@11.0.16/Shutdown.java:130)
	at java.lang.Shutdown.exit(java.base@11.0.16/Shutdown.java:174)
	- locked <0x00000000fd275ea0> (a java.lang.Class for java.lang.Shutdown)
	at java.lang.Runtime.exit(java.base@11.0.16/Runtime.java:116)
	at java.lang.System.exit(java.base@11.0.16/System.java:1752)
	at io.strimzi.operator.cluster.Main.lambda$run$3(Main.java:168)
	at io.strimzi.operator.cluster.Main$$Lambda$290/0x0000000840406c40.handle(Unknown Source)
	at io.vertx.core.impl.future.FutureImpl$3.onFailure(FutureImpl.java:153)
	at io.vertx.core.impl.future.FutureBase.emitFailure(FutureBase.java:75)
	at io.vertx.core.impl.future.FutureImpl.tryFail(FutureImpl.java:230)
	at io.vertx.core.impl.future.Mapping.onFailure(Mapping.java:45)
	at io.vertx.core.impl.future.FutureBase.emitFailure(FutureBase.java:75)
	at io.vertx.core.impl.future.FutureImpl.tryFail(FutureImpl.java:230)
	at io.vertx.core.impl.future.PromiseImpl.tryFail(PromiseImpl.java:23)
	at io.vertx.core.impl.future.PromiseImpl.onFailure(PromiseImpl.java:54)
	at io.vertx.core.impl.future.PromiseImpl.handle(PromiseImpl.java:43)
	at io.vertx.core.impl.future.PromiseImpl.handle(PromiseImpl.java:23)
	at io.vertx.core.impl.DeploymentManager.lambda$reportResult$2(DeploymentManager.java:136)
	at io.vertx.core.impl.DeploymentManager$$Lambda$311/0x000000084041c840.handle(Unknown Source)
	at io.vertx.core.impl.AbstractContext.dispatch(AbstractContext.java:100)
	at io.vertx.core.impl.AbstractContext.dispatch(AbstractContext.java:63)
	at io.vertx.core.impl.EventLoopContext.lambda$runOnContext$0(EventLoopContext.java:38)
	at io.vertx.core.impl.EventLoopContext$$Lambda$294/0x0000000840418840.run(Unknown Source)
	at io.netty.util.concurrent.AbstractEventExecutor.runTask(AbstractEventExecutor.java:174)
	at io.netty.util.concurrent.AbstractEventExecutor.safeExecute(AbstractEventExecutor.java:167)
	at io.netty.util.concurrent.SingleThreadEventExecutor.runAllTasks(SingleThreadEventExecutor.java:470)
	at io.netty.channel.nio.NioEventLoop.run(NioEventLoop.java:503)
	at io.netty.util.concurrent.SingleThreadEventExecutor$4.run(SingleThreadEventExecutor.java:995)
	at io.netty.util.internal.ThreadExecutorMap$2.run(ThreadExecutorMap.java:74)
	at io.netty.util.concurrent.FastThreadLocalRunnable.run(FastThreadLocalRunnable.java:30)
	at java.lang.Thread.run(java.base@11.0.16/Thread.java:829)
``` 

We know that the Verticle is failing to start. This triggers Vertx `DeploymentManager.rollback()`, which ends up calling `DeploymentManager.reportResult()` (reference in the stack). If you look at this method, you'll see that the `DeploymentManager.DeploymentImpl.doUndeploy()` handler is called, which calls `verticleHolder.verticle.stop(stopPromise)`.

The issue: if a `Verticle.stop()` implementation fails to call the future, the stop will never complete. This is also noted in `DeploymentManager.undeployAll()`, which is called when you do `Vertx.close()`. This exactly why I added that 2 minutes timeout in the ShutdownHook. In the last commit, I made the timeout configurable, so that we can also specify a different value for each operator.

Do you think it is possible that `ClusterOperator.stop()` may fail or take long time to call `stop.complete()` under this conditions?

scholzj:
I don't think the configurability solves it. What is the right timeout for something what never timeouts? 0? Why not setup the hook only after a successful start as I suggested? That will at least eliminate this issues when it fails at startup?

fvaleri:
> I don't think the configurability solves it. 

I didn't say that Configurability solves the issue, it is just an enhancement. 



> Why not setup the hook only after a successful start as I suggested?  That will at least eliminate this issues when it fails at startup?

Yes, but not on a clean shutdown where, for whatever reason (future bug?), the stop does not completes.

> What is the right timeout for something what never timeouts? 0?

My point is we should fix the bug.

scholzj:
> My point is that we should fix the bug.

Sorry, what do you mean with the bug? You mean something in Vert.x?

fvaleri:
So [Vertx should timeout if the verticle stop takes too long](https://github.com/eclipse-vertx/vert.x/blob/master/src/main/java/io/vertx/core/impl/DeploymentManager.java#L95), but we should guarantee to always call `stop.complete()` in `ClusterOperator.stop()`. We can simply comment the stop logic to see if my theory holds. Let me do that.

operator-common/src/main/java/io/strimzi/operator/common/ShutdownHook.java
```diff
@@ -12,14 +12,27 @@
...
-    private static final Logger LOGGER = LogManager.getLogger(ShutdownHook.class.getName());
+    private static final Logger LOGGER = LogManager.getLogger(ShutdownHook.class);
     
     private Vertx vertx;
-    
+    private long timeoutMs;
+
     public ShutdownHook(Vertx vertx) {
         this.vertx = vertx;
+        this.timeoutMs = 120_000;
```
tombentley, line 30: Better imho to delegate to the 2-parameter ctor `this(vertx, timeoutMs)`

operator-common/src/main/java/io/strimzi/operator/common/ShutdownHook.java
```diff
@@ -29,19 +42,18 @@ public void run() {
         if (vertx != null) {
```
tombentley, line 42: We can do the null check in the constructor, so we don't need it here.

fvaleri:
I did a quick test commenting out all CO stop logic except `stop.complete()` and now I don't have the following blocked thread. 

```
"SIGTERM handler" #28 daemon prio=9 os_prio=0 cpu=0.71ms elapsed=29.91s tid=0x00007f53cd163000 nid=0x3c waiting for monitor entry  [0x00007f53b0d9e000]
   java.lang.Thread.State: BLOCKED (on object monitor)
	at java.lang.Shutdown.exit(java.base@11.0.16/Shutdown.java:173)
	- waiting to lock <0x00000000fd275ea0> (a java.lang.Class for java.lang.Shutdown)
	at java.lang.Terminator$1.handle(java.base@11.0.16/Terminator.java:51)
	at jdk.internal.misc.Signal$1.run(java.base@11.0.16/Signal.java:220)
	at java.lang.Thread.run(java.base@11.0.16/Thread.java:829)
```

Moreover the pod restart counter is increasing, and this did not happened before (deadlock). Now the 2 minutes ShutdowHook timeout is used, which increases the counter every two minutes. Will double check this tomorrow.

tombentley (approved):
A couple of minor points, but this LGTM.

operator-common/src/main/java/io/strimzi/operator/common/ShutdownHook.java
fvaleri: Right, thanks.

operator-common/src/main/java/io/strimzi/operator/common/ShutdownHook.java
fvaleri: Much better. Thanks.

fvaleri:
@scholzj the bug is in `Vertx.close()`, because `CO.stop()` is never called when the Verticle fails at startup. The workaround to make it fail fast at startup is to replace `System.exit(1)` with `Runtime.getRuntime().halt(1)`, so that we do not trigger `Vertx.close()` in deploy verticle handler. My investigation on why `Vertx.close()` hangs continue...

scholzj:
> @scholzj the bug could be in `Vertx.close()`, because `CO.stop()` is never called when the CO verticle fails at startup. One ugly workaround to make it fail fast is to replace `System.exit(1)` with `Runtime.getRuntime().halt(1)`, so that we do not trigger `Vertx.close()` in that specific case. My investigation on why `Vertx.close()` hangs continue...

Well, we can just remove the shutdown hook and keep using the current way to stopping the operator if we want to skip closing the Vertx. I also still don't understand why we cannot just configure the shutdown hook after all verticles are started which would bypass at least some of these cases.

fvaleri:
@scholzj the issue should be fixed now. 

It is actually due to a Vertx bug, which hangs when you call close with no Verticle deployed. This is exactly the state we have when the CO fails to start because of the RBAC issue.

I also added try-catch inside the StrimziPodSetController, so that it is clear where the CO is failing without having to look at the stack. Let me know if you are good with that.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/StrimziPodSetController.java
```diff
@@ -82,58 +82,63 @@ public class StrimziPodSetController implements Runnable {
...
-        }, 10 * 60 * 1000);
+                @Override
+                public void onDelete(Pod pod, boolean deletedFinalStateUnknown) {
+                    enqueuePod(pod, "DELETED");
+                }
+            }, 10 * 60 * 1000);
 
-        controllerThread = new Thread(this, "StrimziPodSetController");
+            controllerThread = new Thread(this, "StrimziPodSetController");
+        } catch (Throwable e) {
```
scholzj, line 138: Why is this needed?

cluster-operator/src/main/java/io/strimzi/operator/cluster/Main.java
```diff
@@ -73,9 +74,6 @@ public static void main(String[] args) {
                 .setJvmMetricsEnabled(true)
                 .setEnabled(true));
         Vertx vertx = Vertx.vertx(options);
-
-        // Verticle.stop() methods are not executed if you don't call Vertx.close()
-        // Vertx registers a shutdown hook for that, but only if you use its Launcher as main class
```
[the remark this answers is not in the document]
fvaleri: This is in the ShutdownHook javadoc now.

operator-common/src/main/java/io/strimzi/operator/common/ShutdownHook.java
```diff
@@ -0,0 +1,60 @@
...
+                    latch.countDown();
+                });
+                if (!latch.await(timeoutMs, TimeUnit.MILLISECONDS)) {
+                    LOGGER.error("Timed out waiting for shutdown to complete");
+                }
+            }
+            LOGGER.info("Shutdown completed");
+        } catch (Throwable e) {
+            LOGGER.error("Shutdown error, stopping the JVM");
+            Runtime.getRuntime().halt(1);
```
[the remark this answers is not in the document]
fvaleri: Right. Let me get rid of it.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/StrimziPodSetController.java
fvaleri: That way I can print an error log which tells the user that the OP failed during initialization, without having to look at the stack to figure it out.

scholzj: Maybe you can wrap it on the outside then instead of doing this in the constructor? I do not really like the idea of catching something here just to not do anything and just log it.

cluster-operator/src/main/java/io/strimzi/operator/cluster/Main.java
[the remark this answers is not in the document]
scholzj: Ok then I guess.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/StrimziPodSetController.java
fvaleri: Yes, much better. Thanks.

cluster-operator/src/main/java/io/strimzi/operator/cluster/ClusterOperator.java
```diff
@@ -96,31 +96,27 @@ public void start(Promise<Void> start) {
...
         getVertx().createSharedWorkerExecutor("kubernetes-ops-pool", config.getOperationsThreadPoolSize(), TimeUnit.SECONDS.toNanos(120));
 
-        if (config.featureGates().useStrimziPodSetsEnabled()) {
-            strimziPodSetController = new StrimziPodSetController(namespace, config.getCustomResourceSelector(), resourceOperatorSupplier.kafkaOperator, resourceOperatorSupplier.strimziPodSetOperator, resourceOperatorSupplier.podOperations, config.getPodSetControllerWorkQueueSize());
-            strimziPodSetController.start();
-        }
-
         @SuppressWarnings({ "rawtypes" })
-        List<Future> watchFutures = new ArrayList<>(8);
+        List<Future> startFutures = new ArrayList<>(8);
```
scholzj, line 100: Is this 9 now?

fvaleri: I count 8 futures in total, including the new one related to the PodSetController. Am I wrong?

scholzj: Ok. I did not count them, I just saw that you added additional item to it but did not changed the number. So I guess it was wrong already before maybe?

Can you please rebase it to fix the conflict and I will run the regression tests on it. Thanks

fvaleri: I deployed an all ops were working fine.

Yes, I'm going to rebase.

scholzj (approved):
LGTM I guess.

PR #7166, opened 2022-08-06
Start the health and metrics webserver only once per operator pod

scholzj:
### Type of change

- Bugfix

### Description

Today, we start the Vert.x webserver which is used for health checks and for metrics as part of the `ClusterOperator` class / verticle. This does not seem to be right:
* It has nothing to do with the verticle itself, it does not use anything from it
* The `ClusterOperator` verticle is deployed multiple times when watching multiple namespaces. And in that case, each of them creates its own webserver. Somehow, Vert.x deals with it and it does not complain about the port being already used. But it seems weird because the handlers keep existing and the health checks randomly call one of the other handler. You can see it if you enhance the logging:
  ```yaml
  2022-08-06 19:40:23 INFO  ClusterOperator:189 - /metrics for namespace myproject accessed
  2022-08-06 19:40:25 INFO  ClusterOperator:186 - /ready for namespace myproject2 accessed
  2022-08-06 19:40:25 INFO  ClusterOperator:183 - /healthy for namespace myproject accessed
  2022-08-06 19:40:53 INFO  ClusterOperator:189 - /metrics for namespace myproject accessed
  2022-08-06 19:40:55 INFO  ClusterOperator:183 - /healthy for namespace myproject2 accessed
  2022-08-06 19:40:55 INFO  ClusterOperator:186 - /ready for namespace myproject accessed
  ```

This PR improves the logic and moves the webserver into the `Main` class. That way, it is always needed to start only once. This needed only little bit of refactoring to make the MetricsProvider available to it. In the future, this will be important to be able to possibly use leases and leader election.

It also does a minor refactoring of the `main` method in `Main` class to avoid the deep-nesting of the futures and instead compose them. And it adds some more Javadoc to the `Main` class.

### Checklist

### DIFF
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/ClusterOperator.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/ClusterOperator.java
@@ -126,7 +126,6 @@ public class ClusterOperator extends AbstractVerticle {
                         }
                     });
 
-                    //start.complete();
                     return Future.succeededFuture((Void) null);
                 })
                 .onComplete(start);
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/ClusterOperator.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/ClusterOperator.java
@@ -10,19 +10,20 @@ import io.strimzi.operator.cluster.operator.assembly.AbstractConnectOperator;
 import io.strimzi.operator.cluster.operator.assembly.KafkaAssemblyOperator;
 import io.strimzi.operator.cluster.operator.assembly.KafkaBridgeAssemblyOperator;
 import io.strimzi.operator.cluster.operator.assembly.KafkaConnectAssemblyOperator;
+import io.strimzi.operator.cluster.operator.assembly.KafkaMirrorMaker2AssemblyOperator;
 import io.strimzi.operator.cluster.operator.assembly.KafkaMirrorMakerAssemblyOperator;
 import io.strimzi.operator.cluster.operator.assembly.KafkaRebalanceAssemblyOperator;
 import io.strimzi.operator.cluster.operator.assembly.StrimziPodSetController;
 import io.strimzi.operator.cluster.operator.resource.ResourceOperatorSupplier;
 import io.strimzi.operator.common.AbstractOperator;
-import io.strimzi.operator.cluster.operator.assembly.KafkaMirrorMaker2AssemblyOperator;
 import io.vertx.core.AbstractVerticle;
 import io.vertx.core.AsyncResult;
 import io.vertx.core.CompositeFuture;
 import io.vertx.core.Future;
 import io.vertx.core.Handler;
 import io.vertx.core.Promise;
-import io.vertx.core.http.HttpServer;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
 
 import java.util.ArrayList;
 import java.util.List;
@@ -31,9 +32,6 @@ import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.TimeUnit;
 
 import static java.util.Arrays.asList;
-import io.micrometer.prometheus.PrometheusMeterRegistry;
-import org.apache.logging.log4j.LogManager;
-import org.apache.logging.log4j.Logger;
 
 /**
  * An "operator" for managing assemblies of various types <em>in a particular namespace</em>.
@@ -46,8 +44,6 @@ public class ClusterOperator extends AbstractVerticle {
 
     private static final String NAME_SUFFIX = "-cluster-operator";
     private static final String CERTS_SUFFIX = NAME_SUFFIX + "-certs";
-
-    private static final int HEALTH_SERVER_PORT = 8080;
 
     private final KubernetesClient client;
     private final String namespace;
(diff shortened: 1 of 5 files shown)


cluster-operator/src/main/java/io/strimzi/operator/cluster/ClusterOperator.java
```diff
@@ -129,7 +125,9 @@ public void start(Promise<Void> start) {
                             reconcileAll("timer");
                         }
                     });
-                    return startHealthServer().map((Void) null);
+
+                    //start.complete();
```
ppatierno, line 129: remove comment?

cluster-operator/src/main/java/io/strimzi/operator/cluster/Main.java
```diff
@@ -77,39 +88,66 @@ public static void main(String[] args) {
...
-                                LOGGER.error("Unable to start operator for 1 or more namespace", ar.cause());
-                                System.exit(1);
-                            }
-                        });
-                    } else {
-                        LOGGER.error("Failed to gather environment facts", pfa.cause());
+        maybeCreateClusterRoles(vertx, config, client)
+                .compose(i -> startHealthServer(vertx, metricsProvider))
+                .compose(i -> createPlatformFeaturesAvailability(vertx, client))
+                .compose(pfa -> deployClusterOperatorVerticles(vertx, client, metricsProvider, pfa, config))
```
ppatierno, line 100: I was wondering if the health server should be the last one to be started when the cluster operator verticle is up ad running so that the K8S platform gets the readiness of the operator when it's really started.

ppatierno (commented):
Good catch Jakub! I left a couple of comments.

cluster-operator/src/main/java/io/strimzi/operator/cluster/Main.java
scholzj: The webserver just returns HTTP OK all the time. I does not say anything about the ClusterOperator instance(s). But having it before the `deployClusterOperatorVerticles` will be important while implementing a leadership eelction into the operator (which is actually how I found this bug), where the `ClusterOperator` start will be possibly delayed for when the operator instance becomes a leader.

cluster-operator/src/main/java/io/strimzi/operator/cluster/ClusterOperator.java
scholzj: Good catch. Thanks.

cluster-operator/src/main/java/io/strimzi/operator/cluster/Main.java
ppatierno: > I does not say anything about the ClusterOperator instance(s)

Not for this PR but maybe this is a problem, or? Should we have a readiness returning OK only when the CO is really working and "watching" for Kafka related custom resources?

> But having it before the deployClusterOperatorVerticles will be important while implementing a leadership eelction

Can you elaborate a little bit why the HTTP server is useful to run earlier when leadership election will be in place?

scholzj: > Can you elaborate a little bit why the HTTP server is useful to run earlier when leadership election will be in place?

When you have a leadership election in place and run multiple instances of the operator, only one of them will be a leader and will _operate_. The others need to sit and wait if they ever become leaders. And during this time, the health checks need to pass as well. Otherwise Kubernetes will see these processes as failing and will crash loop them. That beats the purpose of the leader election for example as discussed in #7158.

> Not for this PR but maybe this is a problem, or? Should we have a readiness returning OK only when the CO is really working and "watching" for Kafka related custom resources?

Well, we do not do that today and I don't think anyone has any easy solution for how this would work (I.e. how to check that the verticles are really working). So I would say we should get there when we actually have something for that.

ppatierno: >  And during this time, the health checks need to pass as well. Otherwise Kubernetes will see these processes as failing and will crash loop them.

Ok it makes more sense but does it make sense to have the `liveness` passing but not the `readiness` until the leader is actually elected? Just for reference in the future even referring to your second comment.

ppatierno (approved):

cluster-operator/src/main/java/io/strimzi/operator/cluster/Main.java
scholzj: I would say that it is ready in both cases. Just when you are not a leader, readiness does not mean you are operating resources. From my experience, having something not ready for a long time triggers all kind of monitoring tools etc.. So I would avoid it.

ppatierno: I think that "readiness" concept for an operator is kind of ... not well defined.
For a "normal" application exposed via HTTP for example, the readiness is important for avoiding that request are coming through the corresponding K8S `Service` to you: AFAIK, only when your probe reports that you are "ready", the request is forwarded through the `Service` to your application.
For an operator, "readiness" sounds to me that the operator is "watching" custom resources or able to do a periodic reconcile, so it's ready to handle a custom resource creation/update/deletion.

scholzj: From my point of view -> readiness should mean whether the operator is ready. When it is ready as a _warm stand-by_  and waiting to become a leader, it is from my point of view ready as well even when it does not watch resources. That is at the end the point of the leadership election to have something what is _ready_ to take over immediately when the other instance of the operator fails. I'm not sure the HTTP example is easily applicable here because it does not describe the same concept of _warm stand-by_.

I do not think the liveness versus readiness concept is used in general to indicate this - for example for databases which often use the active / passive replica concepts. I haven't seen anything what would be not-ready intentionally for weeks.

PR #7169, opened 2022-08-08
Integrate OAuth 0.11.0 features (OAuth metrics, password grants)

mstruk:
### Type of change

- Enhancement / new feature
- Documentation

### Description

This PR integrates Strimzi OAuth 0.11.0 features into Strimzi custom resources, and configuration generation
 
### Checklist

- [ ] Write tests
- [ ] Update documentation
- [ ] Update CHANGELOG.md


api/src/main/java/io/strimzi/api/kafka/model/listener/KafkaListenerAuthenticationOAuth.java
```diff
@@ -230,6 +233,16 @@ public void setJwksExpirySeconds(Integer jwksExpirySeconds) {
         this.jwksExpirySeconds = jwksExpirySeconds;
     }
 
+    @Description("Enable or disable whether the 'use' attribute of `key` declarations in JWKS endpoint response should be ignored. Default value is false.")
```
PaulRMellor, line 236:
```suggestion
    @Description("Flag to ignore the 'use' attribute of `key` declarations in a JWKS endpoint response. Default value is `false`.")
```

api/src/main/java/io/strimzi/api/kafka/model/listener/KafkaListenerAuthenticationOAuth.java
```diff
@@ -416,4 +429,23 @@ public void setTokenEndpointUri(String tokenEndpointUri) {
...
+    @JsonInclude(JsonInclude.Include.NON_DEFAULT)
+    public boolean isEnableMetrics() {
+        return enableMetrics;
+    }
+
+    public void setEnableMetrics(boolean enableMetrics) {
+        this.enableMetrics = enableMetrics;
+    }
+
+    @Description("Enable or disable if potentially recoverable runtime errors during startup should result in Kafka Broker process termination. Default value is true.")
```
PaulRMellor, line 442:
```suggestion
    @Description("Enable or disable termination of Kafka broker processes due to potentially recoverable runtime errors during startup. Default value is `true`.")
```

documentation/api/io.strimzi.api.kafka.model.authentication.KafkaClientAuthenticationOAuth.adoc
```diff
@@ -58,6 +59,51 @@ authentication:
     key: access-token
 ----
 
+.Username and password
+This way of authenticating clients uses the OAuth `Resource Owner Password Grant` mechanism which has been deprecated, and is only supported to enable integration in environments where the `Client Credentials` mechanism cannot be used due to the existing IAM solutions not supporting it or for some reason requiring the use of user accounts for the authentication.
```
PaulRMellor, line 63:
```suggestion
OAuth username and password configuration uses the OAuth _Resource Owner Password Grant_ mechanism. The mechanism is deprecated, and is only supported to enable integration in environments where client credentials (ID and secret) cannot be used. You might need to use user accounts if your access management system does not support another approach or user accounts are required for authentication.
```

documentation/api/io.strimzi.api.kafka.model.authentication.KafkaClientAuthenticationOAuth.adoc
```diff
@@ -58,6 +59,51 @@ authentication:
...
+
+Typical use for this is to create a special user account in your IAM solution (the authorization server) that represents your client application. Give it a long randomly generated password and a very limited set of permissions - for example, it is only allowed to connect to your Kafka cluster but not allowed to use any other services or do any kind of UI login.
```
PaulRMellor, line 65:
```suggestion
A typical approach is to create a special user account in your authorization server that represents your client application. You then give the account a long randomly generated password and a very limited set of permissions. For example, the account can only connect to your Kafka cluster, but is not allowed to use any other services or login to the user interface.
```

documentation/api/io.strimzi.api.kafka.model.authentication.KafkaClientAuthenticationOAuth.adoc
```diff
@@ -58,6 +59,51 @@ authentication:
...
+
+Consider using a refresh token mechanism first.
+
+You can configure the address of your authorization server in the `tokenEndpointUri` property together with the client ID, username and the password used in authentication.
+The OAuth client will connect to the OAuth server, authenticate using the username, the password, the client ID, and optionally even the client secret to obtain an access token which it will use to authenticate with the Kafka broker.
+
+In the `passwordSecret` property, specify a link to a `Secret` containing the password.
+
+Normally, you also have to configure a `clientId` using a public OAuth client. If using a confidential OAuth client, you also have to configure a `clientSecret`.
```
PaulRMellor, line 74:
```suggestion
Normally, you also have to configure a `clientId` using a public OAuth client. 
If you are using a confidential OAuth client, you also have to configure a `clientSecret`.
```

documentation/modules/oauth/con-oauth-authentication-client.adoc
```diff
@@ -25,3 +25,4 @@ The Kafka client exchanges either:
 
 * Client ID and Secret
 * Client ID, refresh token, and (optionally) a Secret
+* Username and password, with Client ID and (optionally) a Secret
```
PaulRMellor, line 28:
```suggestion
* Client ID, refresh token, and (optionally) a secret
* Username and password, with client ID and (optionally) a secret
```

PaulRMellor: Unless we're specifying the resource, we don't need to capitalize

PaulRMellor (commented):
Looks good
Left a few suggestions. As discussed, we'll need some additional doc updates for grant and metrics.

api/src/main/java/io/strimzi/api/kafka/model/KafkaAuthorizationKeycloak.java
```diff
@@ -44,6 +45,7 @@ public class KafkaAuthorizationKeycloak extends KafkaAuthorization {
     private Integer connectTimeoutSeconds;
     private Integer readTimeoutSeconds;
     private List<String> superUsers;
+    private boolean enableMetrics = false;
```
scholzj, line 48: Stupid question ... how does this work? The `enableMetrics` enables the metrics mechanism inside the OAuth libraries in server. And then the user has to also configure the export rules in the `Kafka` CR to tell the Prometheus JMX exporter to export these JMX metrics as Prometheus metrics. Is that correct?

ppatierno (approved):
LGTM

api/src/main/java/io/strimzi/api/kafka/model/KafkaAuthorizationKeycloak.java
mstruk: That is correct. There's always a possibility or concern that metrics collection may result in greater memory usage or poorer performance so it has to be enabled explicitly. For Prometheus JMX Exporter it integrates via the same mechanism as other metrics. I'm preparing an example OAuth enabled + metrics Kafka cluster deployment, which I'm still testing.

scholzj: Ok, thanks for the explanation. Yeah, I think it makes sense -> just wanted to confirm my understanding is right.

scholzj (approved):

PR #7177, opened 2022-08-09
Add leader election to the Cluster Operator

### LINKED ISSUE #7174: Allow the Cluster Operator to run in multiple replicas
With the introduction of the StrimziPodSets, the availability of the Cluster Operator is more important than before since it is also responsible for (re)starting the Kafka pods. 

In general, the operator itself is:
* Stateless and can easily continue after a restart
* Fast to start when the pod is killed or deleted

So under normal circumstances, there should be no need to run it in multiple replicas. But in situations when multiple worker nodes (e.g. whole AZ) crash, it can easily happen that many pods go down including Kafka / ZooKeeper pods and the cluster operator pod. In such cases, a lot of new pods might need to be scheduled. That can cause congestions or other issues with insufficient resources when the operator cannot be scheduled and the operands and the StrimziPodSets are left un-operated.

Having multiple operator instances running before hand as a warm stand-by might be useful, because the backup instance of the operator will be already running and will be able to take over and handle the things almost immediately.

This can be implemented using the Lease resource and the support for LeaderElection which is part of the Fabric8 Kubernetes client.

scholzj:
### Type of change

- Enhancement / new feature

### Description

This PR adds support for Leader Election. It uses the Fabric8 Leader Election API and the Kubernetes `Lease` resources. When the Cluster Operator starts, it will use the leader election to have one instance of it elected as a leader. When the operator becomes a leader, it will proceed to start the operator verticles and start operate resources. The instances which will not be a leader will be sitting as stand-by and waiting for their chance. The issue #7174 describes the possible reasons why multiple replicas might be useful.

While in the code, the leader election is disabled by default (for backwards compatibility), the installation YAML files will have it enabled. Even if the user runs the operator with a single replica (which remains the default), the leader election will be used => the single replica will be simply always the leader.

The use of the Kubernetes `Lease` resource requires the additional RBAC rights. The configuration in our YAML files uses the `Lease` names `strimzi-cluster-operator` in the namespace where the operator runs. To facilitate this, it has its own `ClusterRole` and `RoleBinding` (a separate RBAC files are used because this RBAC needs to target the namespace where the operator runs and not the one which it watches).

This should close #7174 

### Checklist

- [x] Write tests
- [x] Update documentation
- [x] Update CHANGELOG.md

### DIFF
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/leaderelection/LeaderElectionManager.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/leaderelection/LeaderElectionManager.java
@@ -1,0 +1,123 @@
+/*
+ * Copyright Strimzi authors.
+ * License: Apache License 2.0 (see the file LICENSE or http://apache.org/licenses/LICENSE-2.0.html).
+ */
+package io.strimzi.operator.cluster.leaderelection;
+
+import io.fabric8.kubernetes.client.KubernetesClient;
+import io.fabric8.kubernetes.client.extended.leaderelection.LeaderCallbacks;
+import io.fabric8.kubernetes.client.extended.leaderelection.LeaderElectionConfigBuilder;
+import io.fabric8.kubernetes.client.extended.leaderelection.resourcelock.LeaseLock;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+
+import java.util.function.Consumer;
+
+/**
+ * LeaderElectionManager class is responsible for the leader election process. It has its own long-running thread to
+ * handle the servicing of the Kubernetes Lease lock. It waits for it until it get available and when it becomes the
+ * leader, it also periodically updates the lock to make sure it stays the leader as well. It offers callbacks which
+ * the operators can use to integrate with it.
+ */
+public class LeaderElectionManager implements Runnable  {
+    private static final Logger LOGGER = LogManager.getLogger(LeaderElectionManager.class);
+
+    private final KubernetesClient client;
+    private final LeaderElectionManagerConfig config;
+    private final Runnable startLeadershipCallback;
+    private final Runnable stopLeadershipCallback;
+    private final Consumer<String> leadershipChangeCallback;
+    private final Thread managerThread;
+
+    private boolean isLeader = false;
+
+    /**
+     * LeaderElectionManager constructor
+     *
+     * @param client                    Kubernetes client
+     * @param config                    LeaderElectionManager configuration
+     * @param startLeadershipCallback   Callback which is called when this instance becomes a leader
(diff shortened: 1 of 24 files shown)


cluster-operator/src/main/java/io/strimzi/operator/cluster/leaderelection/LeaderElectionManager.java
```diff
@@ -0,0 +1,123 @@
...
+     */
+    public void start() {
+        managerThread.start();
+    }
+
+    /**
+     * Stop method to stop the election thread
+     */
+    public void stop()  {
+        managerThread.interrupt();
```
ppatierno, line 92: is there any `InterruptedException` thrown in the run that we should catch and handle gracefully instead of having it in the log? Just asking because this exception is not always thrown anyway, so asking if you noticed it by any chance.

documentation/modules/operators/ref-operator-cluster.adoc
```diff
@@ -217,6 +217,50 @@ Enables or disables features and functionality controlled by xref:ref-operator-c
...
+[source,yaml,options="nowrap"]
+----
+env:
+  - name: STRIMZI_LEADER_ELECTION_IDENTITY
+    valueFrom:
+      fieldRef:
+        fieldPath: metadata.name
+----
+
+`STRIMZI_LEADER_ELECTION_LEASE_DURATION_MS`:: Optional, default `15` seconds.
```
ppatierno, line 255: I see are describing the defaults for other `_MS` env var in ms. For example ...

> STRIMZI_FULL_RECONCILIATION_INTERVAL_MS
Optional, default is 120000 ms. The interval between periodic reconciliations, in milliseconds.

I would do the same here as 15000 ms and for the others.

ppatierno (commented):
LGTM overall, I left just a couple of nits.

documentation/modules/operators/ref-operator-cluster.adoc
scholzj: Makes sense. I fixed it for all 3 `_MS` options I added.

cluster-operator/src/main/java/io/strimzi/operator/cluster/leaderelection/LeaderElectionManager.java
scholzj: The Fabric8 method catches the interrupts and does not throw them. So we cannot catch it, but it just finishes the thread basically. And when that happens it looses the leadership and exits.

ppatierno: Yes it's what I was wondering. So we are covered.

ppatierno (approved):

scholzj:
@tombentley I'm going to merge this. If you get to this later and have any comments, please let me know and I will apply them in a separate PR.

PR #7187, opened 2022-08-13
Improve the operator shutdown hook

fvaleri:
I did some more investigation on the hanging shutdown issue, and came out with this improved
shutdown hook which completely avoids it. The previous version does not cover the scenario
where System.exit is invoked from an event-loop thread after a successful verticle deployment.

The problem:

System.exit is a blocking call that waits all registered shutdown-hook threads to complete
before stopping the VM. Vertx.close stops all deployed verticles and waits for the event-loop
threads to complete their current iteration. We have a deadlock when invoking System.exit
from an event-loop thread which triggers a shutdown hook invoking Vertx.close. There is no
issue when invoking System.exit from a non event-loop thread.

### DIFF
--- a/operator-common/src/main/java/io/strimzi/operator/common/ShutdownHook.java
+++ b/operator-common/src/main/java/io/strimzi/operator/common/ShutdownHook.java
@@ -5,19 +5,26 @@
 package io.strimzi.operator.common;
 
 import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
+import io.vertx.core.Promise;
+import io.vertx.core.Verticle;
 import io.vertx.core.Vertx;
+import io.vertx.core.impl.Deployment;
+import io.vertx.core.impl.VertxImpl;
 import static java.util.Objects.requireNonNull;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
-
-import java.util.concurrent.CountDownLatch;
+import java.util.Set;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
 import java.util.concurrent.TimeUnit;
 
 /**
- * This shutdown hook ensure that {@code Vertx.close()} is called on a clean shutdown,
- * which in turn calls the stop method of all running Verticles.
- * <p>
- * We add a fixed timeout because Vertx has none for stopping running Verticles.
+ * Shutdown hook that retrieves and stops all deployed verticles without invoking Vertx.close.
+ * <br><br>
+ * System.exit is a blocking call that waits all registered shutdown-hook threads to complete before stopping the VM.
+ * Vertx.close stops all deployed verticles and waits for the event-loop threads to complete their current iteration.
+ * We have a deadlock when invoking System.exit from an event-loop thread which triggers a shutdown hook invoking Vertx.close.
+ * There is no issue when invoking System.exit from a non event-loop thread.
  */
 @SuppressFBWarnings("RCN_REDUNDANT_NULLCHECK_OF_NONNULL_VALUE")
 public class ShutdownHook implements Runnable {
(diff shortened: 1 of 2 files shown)


operator-common/src/main/java/io/strimzi/operator/common/ShutdownHook.java
```diff
@@ -5,19 +5,26 @@
 package io.strimzi.operator.common;
 
 import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
+import io.vertx.core.Promise;
+import io.vertx.core.Verticle;
 import io.vertx.core.Vertx;
+import io.vertx.core.impl.Deployment;
+import io.vertx.core.impl.VertxImpl;
```
scholzj, line 12: Is this really something what should be used? Sounds not like a part of the public API.

operator-common/src/main/java/io/strimzi/operator/common/ShutdownHook.java
```diff
@@ -38,22 +45,48 @@ public ShutdownHook(Vertx vertx, long timeoutMs) {
...
+        vertx.deploymentIDs()
+                .stream()
+                .map(id -> ((VertxImpl) vertx).getDeployment(id))
+                .forEach(deployment -> stopVerticles(deployment));
+        LOGGER.info("Shutdown completed");
+    }
+
+    private void stopVerticles(Deployment deployment) {
+        Set<Verticle> verticles = deployment.getVerticles();
+        ExecutorService executor = Executors.newFixedThreadPool(verticles.size());
```
scholzj, line 57: Is it really the right choice to start starting new executors when you are shutting down?

operator-common/src/main/java/io/strimzi/operator/common/ShutdownHook.java
fvaleri: You are right. Let me see if there is a way to achieve the same using the public API.

operator-common/src/main/java/io/strimzi/operator/common/ShutdownHook.java
fvaleri: That is to speed up the shutdown in case of deployments with multiple verticles. Let me also review this one.

scholzj: I think the shutdown was completely fine before the shutdown hook. It seems really weird to now _shutdown_ by starting more new things.

scholzj:
@tombentley @ppatierno I leave this one to you, I don't understand why do we need the shutdown hook at all.

fvaleri:
It's the same reason we need verticle stop. This is simply to ensure these methods are called on a clean shutdown and execute whatever cleanup logic is in there (e.g. stop timers, finish inflight reconciliations), with a max timeout. No less, no more.

fvaleri:
@scholzj I see your point that the operators are stateless, so it's not a big deal if the verticle stop methods are not called, but I think it's still important to execute the cleanup logic whenever is possible, because not doing so may cause some undesired effect on remote systems we are connecting to (resources may remain committed to the connection for an indeterminate time).

This code is very fast, if you are concerned that this may delay the operator shutdown because of some issue in the cleanup logic, we can set a shorter timeout for cluster and user operators whose logic is pretty simple (e.g. 30s as the default K8s terminationGracePeriodSeconds). 

I'm open to improvements, so please let me know if you think that there is a better way to implement the same.

scholzj:
My concern is that this whole think does not seem to solve an actual issue but has risks connected to it. You talk about clean-up logic, but there is only little actual cleanup logic implemented. So what you are doing is a super clean Vert.x shutdown but it is not obvious how and why that matters.

tombentley:
Thanks @fvaleri, I took a look at this and couldn't quite follow the reasoning. The TO doesn't call `System.exit`, so I don't quite follow how it can be involved in a deadlock.

I think @scholzj's point in https://github.com/strimzi/strimzi-kafka-operator/pull/7165 was that the TO ought to be crash-safe. i.e. it shouldn't be necessary for the shutdown hook to have run in order for the TO to recover correctly the next time it runs (this is because in general the JVM can terminate without shutdown hooks being executed, e.g. `kill -9`). The TO does have _some_ persistent state (the topic store), but the way the reconciliation works should be safe if that doesn't get updated in a reconciliation (so long as what ever other update are needed _happen before_ the Topic Store update).

So I wonder if we're not progressing in the wrong direction here. We need the TO to be crash safe anyway, even without the shutdown hook. So perhaps rather than trying to fix the shutdown hook to not deadlock it would be better to ensure the crash-safety.

fvaleri:
@tombentley yes, it was the CO that was calling System.exit, not the TO. That said, every verticle could potentially call it in the future.

I see the point of crash safety and the shutdown hook is not intended as a way to ensure that, but only to make sure that whatever cleanup code we have in verticle stop is actually executed on clean shutdowns. The reasons are that we are maintaining code that is never used and closing clients helps external systems to free some resources faster.

I agree that the stop/cleanup logic should be quick and the TO is doing too much waiting some time for inflight reconciliations. Then, at some point, Jakub pointed out that the hook was holding the CO shutdown when it was failing because of a startup issue. That was true, it was consuming the whole 2 minutes timeout and the reason is what I describe in this PR.

This new shutdown hook implementation avoids using Vertx.close and simply retrieves deployment IDs, which are then used to undeploy all verticles in parallel. The undeploy API triggers the verticle stop execution. In my tests, this is pretty fast, but I'm good to reduce the 2 minutes timeout to whatever value you think is more appropriate.

scholzj:
Discussed on the Community call on 3rd November: Should be closed (details in the call recording).

PR #7191, opened 2022-08-16
[MO] - [system test] -> Pod Security policies 

see-quick:
### Type of change

- Enhancement / new feature

### Description

This PR adds a test cases for the `Pluggable Pod Security Providers` feature (i.e., https://github.com/strimzi/strimzi-kafka-operator/pull/7089). 

PodSecurityPoliciesST provides tests for Pod Security policies. In short, Pod security profiles are a mechanism used in Pods or containers, which may prohibit some operations (e.g., running only as a non-root user, allowing only some Volume types etc.).

Reason why is this test suite has to run in complete isolation (i.e., {@link IsolatedSuite}) is that we need to modify Cluster Operator configuration, specifically env {@code STRIMZI_POD_SECURITY_PROVIDER_CLASS} to restricted.

Test cases are designed to verify the standard behaviour of Pod Security policies. Specifically, (i.) we check if containers such as Kafka, ZooKeeper, EntityOperator, and KafkaBridge have correctly set .securityContext (ii.) then we check if these resources working and are stable with exchanging messages.

### Checklist

- [x] Write tests

### DIFF
--- a/systemtest/src/main/java/io/strimzi/systemtest/annotations/RequiredMinKubeOrOpenshiftVersionCondition.java
+++ b/systemtest/src/main/java/io/strimzi/systemtest/annotations/RequiredMinKubeOrOpenshiftVersionCondition.java
@@ -1,0 +1,44 @@
+/*
+ * Copyright Strimzi authors.
+ * License: Apache License 2.0 (see the file LICENSE or http://apache.org/licenses/LICENSE-2.0.html).
+ */
+package io.strimzi.systemtest.annotations;
+
+import io.strimzi.test.k8s.KubeClusterResource;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+import org.junit.jupiter.api.extension.ConditionEvaluationResult;
+import org.junit.jupiter.api.extension.ExecutionCondition;
+import org.junit.jupiter.api.extension.ExtensionContext;
+
+import java.util.Optional;
+
+import static org.junit.platform.commons.support.AnnotationSupport.findAnnotation;
+
+public class RequiredMinKubeOrOpenshiftVersionCondition implements ExecutionCondition  {
+
+    private static final Logger LOGGER = LogManager.getLogger(RequiredMinKubeOrOpenshiftVersionCondition.class);
+
+    @Override
+    public ConditionEvaluationResult evaluateExecutionCondition(ExtensionContext extensionContext) {
+        final Optional<RequiredMinKubeOrOcpBasedKubeVersion> annotation = findAnnotation(extensionContext.getElement(), RequiredMinKubeOrOcpBasedKubeVersion.class);
+        final KubeClusterResource cluster = KubeClusterResource.getInstance();
+        final double ocpBasedKubeVersion = annotation.get().ocpBasedKubeVersion();
+        final double kubeVersion = annotation.get().kubeVersion();
+        final boolean isOcp = cluster.isOpenShift();
+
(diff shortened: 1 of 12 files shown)


im-konge (approved):

scholzj (commented):
Pod Security Policies is the name of the Kubernetes feature which is being removed (https://kubernetes.io/docs/concepts/security/pod-security-policy/). The new feautre is called Pod Security Admission, eventually Pod Security Standards / Pod Security Profiles refers to the actual profiles the pods might match. Or Pod Security Providers is how we called the Strimzi plugins. We should name it accordingly to avoid any confusion wth the deprecated Kube APIs.

systemtest/src/main/java/io/strimzi/systemtest/annotations/RequiredMinKubeOrOpenshiftVersionCondition.java
```diff
@@ -0,0 +1,44 @@
...
+        if ((isOcp && Double.parseDouble(cluster.client().clusterKubernetesVersion()) >= ocpBasedKubeVersion) ||
+            Double.parseDouble(cluster.client().clusterKubernetesVersion()) >= kubeVersion) {
+            return ConditionEvaluationResult.enabled("Test is enabled");
+        } else {
+            LOGGER.info("{} is @RequiredMinKubeOrOpenshiftVersion with type of cluster: {} and version {}, but the running on cluster with {}: Ignoring {}",
+                extensionContext.getDisplayName(),
+                cluster.cluster().getClass().toString(),
+                isOcp ? ocpBasedKubeVersion : kubeVersion,
+                cluster.client().clusterKubernetesVersion(),
+                extensionContext.getDisplayName()
```
Frawless, line 39: Guess we don't need it twice in one log message

Frawless (approved):
Just one nit + Jakub's comment

see-quick:
> Pod Security Policies is the name of the Kubernetes feature which is being removed (https://kubernetes.io/docs/concepts/security/pod-security-policy/). The new feautre is called Pod Security Admission, eventually Pod Security Standards / Pod Security Profiles refers to the actual profiles the pods might match. Or Pod Security Providers is how we called the Strimzi plugins. We should name it accordingly to avoid any confusion wth the deprecated Kube APIs.

I have changed it to **Pod Security profiles**.

scholzj (commented):
I do not see you to configure the enforcement of the restricted profile on the namespace level. So does this purely test the security context from the pods? Should we also enforce the profile? (https://kubernetes.io/docs/concepts/security/pod-security-admission/#pod-security-admission-labels-for-namespaces)

test/src/main/java/io/strimzi/test/k8s/KubeClient.java
```diff
@@ -91,6 +91,10 @@ public void createNamespace(String name) {
         client.namespaces().resource(ns).createOrReplace();
     }
 
+    public void createNamespace(final Namespace namespace) {
```
scholzj, line 94: Should it be called `createOrReplaceNamespace` then?

scholzj (commented):
One nit. LGTM otherwise.

test/src/main/java/io/strimzi/test/k8s/KubeClient.java
see-quick: > Should it be called `createOrReplaceNamespace` then?

Yes, will do.

scholzj (approved):
LGTM assuming the tests pass.

PR #7195, opened 2022-08-17
Deprecate Jaeger / OpenTracing based tracing support

scholzj:
### Type of change

- Task

### Description

Currently, we support tracing using Jaeger clients and the OpenTracing API. However, the Jaeger clients are retired and the OpenTracing project is archived. Especially the OpenTracing archiving is a problem for us since there are no new releases of the OpenTracing Kafka tooling. It already isn't compatible with Kafka Streams API in Kafka 3.0+ and it might happen that in the future it also isn't compatible anymore with the Consumer / Producer APIs. Therefore, I do not think we are able to guarantee that it will work with the _next_ Kafka release and should deprecate it.

This PR deprecates the API and adds a note to the docs. It does not plan the removal yet -> we can keep supporting it while it works (especially since there is no replacement right now).

We plan to support tracing based on OpenTelemetry in the future. However, this work is not complete yet, so there is no replacement at this point. However, due to the risks of the future compatibility around OpenTracing, it seems fair to deprecate it even without replacement.

### Checklist

- [x] Update documentation
- [x] Update CHANGELOG.md

### DIFF
--- a/documentation/assemblies/tracing/assembly-distributed-tracing.adoc
+++ b/documentation/assemblies/tracing/assembly-distributed-tracing.adoc
@@ -19,6 +19,13 @@ Support for tracing is built in to the following components:
 * MirrorMaker 2.0
 * Strimzi Kafka Bridge
 
+[NOTE]
+====
+The `type: jaeger` tracing support based on Jaeger clients and OpenTracing API is deprecated.
+Because the Jaeger clients are retired and the OpenTracing project is archived, we cannot guarantee their support for future Kafka versions.
+In the future, we plan to replace it with a new tracing feature based on the OpenTelemetry project.
+====
+
 You enable and configure tracing for these components using template configuration properties in their custom resources.
 
 To enable tracing in Kafka producers, consumers, and Kafka Streams API applications, you _instrument_ application code using the link:https://github.com/opentracing-contrib/java-kafka-client/blob/master/README.md[OpenTracing Apache Kafka Client Instrumentation^] library (included with Strimzi). When instrumented, clients generate trace data; for example, when producing messages or writing offsets to the log.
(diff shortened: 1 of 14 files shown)


ppatierno (approved):
LGTM

documentation/assemblies/tracing/assembly-distributed-tracing.adoc
```diff
@@ -19,6 +19,13 @@ Support for tracing is built in to the following components:
 * MirrorMaker 2.0
 * Strimzi Kafka Bridge
 
+[NOTE]
```
PaulRMellor, line 22: We haven't mentioned Jaeger yet, so maybe we should add a sentence before the note.

_By default, Strimzi uses Jaeger tracing. It is enabled by setting the `spec.tracing.type` property to `jaeger` in the component configuration._

documentation/assemblies/tracing/assembly-distributed-tracing.adoc
```diff
@@ -19,6 +19,13 @@ Support for tracing is built in to the following components:
...
+====
+The `type: jaeger` tracing support based on Jaeger clients and OpenTracing API is deprecated.
+Because the Jaeger clients are retired and the OpenTracing project is archived, we cannot guarantee their support for future Kafka versions.
+In the future, we plan to replace it with a new tracing feature based on the OpenTelemetry project.
```
PaulRMellor, line 26:
```suggestion
Support for `type: jaeger` tracing is deprecated.
The Jaeger clients are now retired and the OpenTracing project archived. 
As such, we cannot guarantee their support for future Kafka versions.
In the future, we plan to introduce a new tracing feature based on the OpenTelemetry project.
```

PaulRMellor (approved):
Looks good. A couple of suggestions, including maybe adding a line to give some content to the note.

PR #7203, opened 2022-08-18
[MO] - [system test] -> Un-necessary creation of additional namespace in ParallelSuites

see-quick:

### Type of change

- Bugfix
- Enhancement / new feature
- Refactoring

### Description

When the @ParallelSuite test suite contains only @ParallelNamespaceTests it creates an unnecessary namespace. When one runs the tests, we need to dynamically check if such a test suite contains @ParallelTest or @IsolatedTest, and then we need to create an additional namespace. Otherwise not. This PR solves such a problem.

### Checklist

- [x] Write tests

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -793,6 +793,11 @@
                 <artifactId>junit-platform-launcher</artifactId>
                 <version>${junit.platform.version}</version>
             </dependency>
+            <dependency>
+                <groupId>org.junit.platform</groupId>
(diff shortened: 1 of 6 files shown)


Frawless (approved):

im-konge (approved):

scholzj (approved):
LGTM if the regression tests pass.

PR #7214, opened 2022-08-18
Move CR validation from AbstractOperator to StatusUtils

scholzj:
### Type of change

- Refactoring

### Description

The `validate` method in `AbstractOperator` class is used to get warnings from the Custom Resource (for example due to the use of deprecated fields) for the conditions section of the custom resource status. This PR moves this method from `AbstractOperator` class to `StatusUtils` where it can be used as a static method and available to other classes as well. It moves in the sme direction also the method `addWarningsToStatus` for adding additional conditions to the existing `Status` object. (This is related to my UserOperator refactoring, just want to do this as a separate PR to make reviews easier)

This PR also removes some warnings and unused methods from the `StatusUtils` class. 

It also does refactoring to the `OperatorMetricsTest`. This was needed to get the tests passing with the moved `validate` method => the nested `Foo` classes were not accessible to the method from a different class. This refactoring seems to also simplify the code in general.

### Checklist

### DIFF
--- a/operator-common/src/main/java/io/strimzi/operator/common/operator/resource/StatusUtils.java
+++ b/operator-common/src/main/java/io/strimzi/operator/common/operator/resource/StatusUtils.java
@@ -160,11 +160,11 @@ public class StatusUtils {
      * Adds additional conditions to te status (this expects)
      *
      * @param status                            The Status instance where additonal conditions should be added
-     * @param unknownAndDeprecatedConditions    The Set with the new conditions
+     * @param conditions    The Set with the new conditions
      */
-    public static void addConditionsToStatus(Status status, Set<Condition> unknownAndDeprecatedConditions)   {
+    public static void addConditionsToStatus(Status status, Set<Condition> conditions)   {
         if (status != null)  {
-            status.addConditions(unknownAndDeprecatedConditions);
+            status.addConditions(conditions);
         }
     }
 }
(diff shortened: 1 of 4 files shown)


operator-common/src/main/java/io/strimzi/operator/common/operator/resource/StatusUtils.java
```diff
@@ -136,4 +128,43 @@ public static Condition getPausedCondition() {
...
+        return Collections.emptySet();
+    }
+
+    /**
+     * Adds additional conditions to te status (this expects)
+     *
+     * @param status                            The Status instance where additonal conditions should be added
+     * @param unknownAndDeprecatedConditions    The Set with the new conditions
+     */
+    public static void addConditionsToStatus(Status status, Set<Condition> unknownAndDeprecatedConditions)   {
```
ppatierno, line 165: is the variable name as `unknownAndDeprecatedConditions` still valid? This method was for adding warning conditions before but now it was renamed as a more general to add conditisions, so maybe the param could just be `conditions`?

ppatierno (approved):
LGTM. I just left a nit.

operator-common/src/main/java/io/strimzi/operator/common/operator/resource/StatusUtils.java
scholzj: Good catch. I renamed the method to give it more generic name but not the param. Fixed.

PR #7215, opened 2022-08-19
Fixed Kafka Exporter Grafana dashboard with N/A for empty numeric fields

### LINKED ISSUE #2628: Kafka Exporter reporting some wrong values when no data traffic
The current provided Kafka Exporter dashboard has its main fields based on a specific exporter metric, `kafka_consumergroup_current_offset`, which is not exposed by the exporter itself until there is no real data exchange between clients (it's not enough having a consumer joined to a consumer group, but a producer sending messages is needed as well).
This means that when the user takes a look at the dashboard when no data are exchanged, the dashboard looks like the following one:

Screenshot-20200302154210-1904x970

It means that, even if you have topics (so partitions), but no data is exchanged between clients, the values are still 0 which doesn't reflect the real status of the cluster (same for replicas, insync replicas, ...).
I guess that this dashboard was provided in order to work on its own without relying on metrics exported by the Kafka brokers. So it means that even if you create a `Kafka` resource without `metrics` section you can get some metrics from the Kafka exporter dashboard anyway.
Was it done on purpose? I think that we should rely on the integration with Kafka `metrics` so using some metrics from brokers to fill fields like topics, replicas and so on in order to avoid this 0 values in the exporter dashboard. Of course it means that the exporter dashboard works fine only if you deploy the `Kafka` resource with `metrics` enabled.

ppatierno:
This PR fixes #2628 
It changes the Kafka Exporter Grafana dashboard in order to show `N/A` instead of `0` when metrics are not available to show current data.
It also explains the meaning of this in the corresponding documentation.

### DIFF
--- a/documentation/modules/metrics/con_kafka-exporter-lag.adoc
+++ b/documentation/modules/metrics/con_kafka-exporter-lag.adoc
@@ -13,6 +13,8 @@ Kafka Exporter extracts additional metrics data from Kafka brokers related to of
 The metrics data is used, for example, to help identify slow consumers.
 Lag data is exposed as Prometheus metrics, which can then be presented in Grafana for analysis.
 
+The Grafana dashboard for Kafka Exporter, provided as an example, is available in the xref:ref-metrics-dashboards-{context}[Example Grafana dashboards] list.
+
 IMPORTANT: Kafka Exporter provides only additional metrics related to consumer lag and consumer offsets.
 For regular Kafka metrics, you have to configure the Prometheus metrics in xref:proc-metrics-kafka-deploy-options-{context}[Kafka brokers].
 
--- a/documentation/modules/metrics/con_kafka-exporter-lag.adoc
+++ b/documentation/modules/metrics/con_kafka-exporter-lag.adoc
@@ -13,7 +13,7 @@ Kafka Exporter extracts additional metrics data from Kafka brokers related to of
 The metrics data is used, for example, to help identify slow consumers.
 Lag data is exposed as Prometheus metrics, which can then be presented in Grafana for analysis.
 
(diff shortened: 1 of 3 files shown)


scholzj (approved):

documentation/modules/metrics/con_kafka-exporter-lag.adoc
```diff
@@ -13,6 +13,8 @@ Kafka Exporter extracts additional metrics data from Kafka brokers related to of
 The metrics data is used, for example, to help identify slow consumers.
 Lag data is exposed as Prometheus metrics, which can then be presented in Grafana for analysis.
 
+The Grafana dashboard for Kafka Exporter, provided as an example, is available in the xref:ref-metrics-dashboards-{context}[Example Grafana dashboards] list.
```
PaulRMellor, line 16:
```suggestion
A Grafana dashboard for Kafka Exporter is one of a number of xref:ref-metrics-dashboards-{context}[example Grafana dashboards] provided by Strimzi.
```

documentation/modules/metrics/ref_metrics-dashboards.adoc
```diff
@@ -49,3 +49,5 @@ The dashboards are populated with a representative set of metrics for monitoring
 |`strimzi-kafka-exporter.json`
 
 |===
+
+NOTE: When metrics are not available to Kafka Exporter, because there is no traffic in the cluster yet, the corresponding Grafana dashboard will show `N/A` for numeric fields and `No data to show` for graphs.
```
PaulRMellor, line 53:
```suggestion
NOTE: When metrics are not available to the Kafka Exporter, because there is no traffic in the cluster yet, the Kafka Exporter Grafana dashboard will show `N/A` for numeric fields and `No data to show` for graphs.
```

PaulRMellor (approved):
LGTM :+1:

PR #7223, opened 2022-08-22
Add support for IPv6 addresses in our TLS certificates

### LINKED ISSUE #7217: Kafka cluster cannot be deployed with IPv6 advertised names
_Please use this to only for bug reports. For questions or when you need help, you can use the [GitHub Discussions](https://github.com/strimzi/strimzi-kafka-operator/discussions), our [#strimzi Slack channel](https://slack.cncf.io/) or out [user mailing list](https://lists.cncf.io/g/cncf-strimzi-users/topics)._

Assigning IPv6 to advertised names for kafka nodes or using IPv6 load balancers causes strimzi operator fails with error.

**To Reproduce**
1. Create kafka cluster which assigns IPv6 addresses to listeners's LB
2. Configure ipFamilies as IPv6 only
3. Try to deploy cluster

**Expected behavior**
Cluster is starting normally

 - Strimzi version: 0.30.0
 - Kubernetes cluster: Kubernetes 1.21
 - Infrastructure: Yandex Cloud

**YAML files and logs**
`2022-08-20 07:27:50 INFO  ClusterOperator:128 - Triggering periodic reconciliation for namespace *
2022-08-20 07:27:51 INFO  AbstractOperator:226 - Reconciliation #3604(timer) Kafka(kafka/kafka): Kafka kafka will be checked for creation or modification
2022-08-20 07:27:55 ERROR AbstractOperator:247 - Reconciliation #3604(timer) Kafka(kafka/kafka): createOrUpdate failed
java.lang.IllegalArgumentException: Invalid DNS name: MY_IPV6_ADDRESS
        at io.strimzi.certs.Subject$Builder.addDnsName(Subject.java:47) ~[io.strimzi.certificate-manager-0.30.0.jar:0.30.0]
        at io.strimzi.operator.cluster.model.ClusterCa.lambda$generateBrokerCerts$4(ClusterCa.java:204) ~[io.strimzi.cluster-operator-0.30.0.jar:0.30.0]
        at io.strimzi.operator.cluster.model.Ca.maybeCopyOrGenerateCerts(Ca.java:430) ~[io.strimzi.operator-common-0.30.0.jar:0.30.0]
        at io.strimzi.operator.cluster.model.ClusterCa.generateBrokerCerts(ClusterCa.java:224) ~[io.strimzi.cluster-operator-0.30.0.jar:0.30.0]
        at io.strimzi.operator.cluster.model.KafkaCluster.generateCertificatesSecret(KafkaCluster.java:1259) ~[io.strimzi.cluster-operator-0.30.0.jar:0.30.0]
        at io.strimzi.operator.cluster.operator.assembly.KafkaReconciler.lambda$certificateSecret$46(KafkaReconciler.java:729) ~[io.strimzi.cluster-operator-0.30.0.jar:0.30.0]
        at io.vertx.core.impl.future.Composition.onSuccess(Composition.java:38) ~[io.vertx.vertx-core-4.2.4.jar:4.2.4]
        at io.vertx.core.impl.future.FutureBase.emitSuccess(FutureBase.java:60) ~[io.vertx.vertx-core-4.2.4.jar:4.2.4]
        at io.vertx.core.impl.future.FutureImpl.tryComplete(FutureImpl.java:211) ~[io.vertx.vertx-core-4.2.4.jar:4.2.4]
        at io.vertx.core.impl.future.PromiseImpl.tryComplete(PromiseImpl.java:23) ~[io.vertx.vertx-core-4.2.4.jar:4.2.4]
        at io.vertx.core.impl.future.PromiseImpl.onSuccess(PromiseImpl.java:49) ~[io.vertx.vertx-core-4.2.4.jar:4.2.4]
        at io.vertx.core.impl.future.FutureBase.lambda$emitSuccess$0(FutureBase.java:54) ~[io.vertx.vertx-core-4.2.4.jar:4.2.4]
        at io.netty.util.concurrent.AbstractEventExecutor.runTask(AbstractEventExecutor.java:174) ~[io.netty.netty-common-4.1.77.Final.jar:4.1.77.Final]
        at io.netty.util.concurrent.AbstractEventExecutor.safeExecute(AbstractEventExecutor.java:167) ~[io.netty.netty-common-4.1.77.Final.jar:4.1.77.Final]
        at io.netty.util.concurrent.SingleThreadEventExecutor.runAllTasks(SingleThreadEventExecutor.java:470) ~[io.netty.netty-common-4.1.77.Final.jar:4.1.77.Final]
        at io.netty.channel.nio.NioEventLoop.run(NioEventLoop.java:503) ~[io.netty.netty-transport-4.1.77.Final.jar:4.1.77.Final]
        at io.netty.util.concurrent.SingleThreadEventExecutor$4.run(SingleThreadEventExecutor.java:995) ~[io.netty.netty-common-4.1.77.Final.jar:4.1.77.Final]
        at io.netty.util.internal.ThreadExecutorMap$2.run(ThreadExecutorMap.java:74) ~[io.netty.netty-common-4.1.77.Final.jar:4.1.77.Final]
        at io.netty.util.concurrent.FastThreadLocalRunnable.run(FastThreadLocalRunnable.java:30) ~[io.netty.netty-common-4.1.77.Final.jar:4.1.77.Final]
        at java.lang.Thread.run(Thread.java:829) ~[?:?]`

Attach or copy paste the custom resources you used to deploy the Kafka cluster and the relevant YAMLs created by the Cluster Operator.
Attach or copy and paste also the relevant logs.

```
kind: Kafka
metadata:
  name: kafka
  namespace: kafka
  annotations:
    strimzi.io/use-connector-resources: "true"
spec:
  kafka:
    version: 3.2.0
    image: quay.io/strimzi/kafka:0.30.0-kafka-3.2.0
    replicas: 5
    rack:
      topologyKey: kubernetes.io/hostname
    jvmOptions:
      -Xms: 8192m
      -Xmx: 8192m
    listeners:
      - name: plain
        port: 9092
        type: internal
        tls: false
        authentication:
          type: scram-sha-512
        configuration:
          useServiceDnsDomain: true
      - name: tls
        port: 9093
        type: internal
        tls: true
        authentication:
          type: scram-sha-512
        configuration:
          useServiceDnsDomain: true
      - name: external
        port: 9095
        type: loadbalancer
        tls: true
        authentication:
          type: scram-sha-512
        configuration:
          ipFamilies:
          - IPv6
          bootstrap:
            loadBalancerIP: fc01::8d1c
          brokers:
            - broker: 0
            - broker: 1
            - broker: 2
            - broker: 3
            - broker: 4
    authorization:
      type: simple
      superUsers:
        - root
        - apicurio
    config:
      offsets.topic.replication.factor: 3
      transaction.state.log.replication.factor: 3
      transaction.state.log.min.isr: 2
      default.replication.factor: 3
      min.insync.replicas: 2
      log.message.format.version: "3.2"
      inter.broker.protocol.version: "3.2"

      auto.create.topics.enable: false
      topic.creation.enable: true
      log.cleanup.policy: compact
      log.retention.ms: -1
      message.max.bytes: 1073741824
      log.cleaner.max.compaction.lag.ms: 84600000 # 1 day`

```
*To easily collect all YAMLs and logs, you can use our [report script](https://github.com/strimzi/strimzi-kafka-operator/blob/main/tools/report.sh) which will automatically collect all files and prepare a ZIP archive which can be easily attached to this issue.
The usage of this script is:
`./report.sh --namespace <string> --cluster <string> [--bridge <string>] [--connect <string>] [--mm2 <string>]`*

scholzj:
### Type of change

- Bugfix
- Enhancement / new feature

### Description

Addresses added to certificate SANs have to be added either as IP addresses or as DNS addresses. So Strimzi does a validation to decide if the address is IP or DNS and adds them. However, we currently support only IPv4 addresses. And if user tries to add IPv6 address, it will be rejected as invalid.

This PR adds support for IPv6 addresses to Strimzi. It is non-trivial task because the format of IPv6 addresses is more complicated compared to IPv4. And also, we need to normalize the IPv6 address, because the certificate management tool (currently OpenSSL) will do that as well and we need to compare the different addresses against each other. To achieve this, we add:
* New class `IpAndDnsValidation` where we centralize the logic for validating IP and DNS addresses, which was currently duplicated between multiple different classes. So now it is in one central class.
* Support for validating IPv6 addresses in both compressed and uncompressed formats
* Support for normalizing the IPv6 address. We use lowercase uncompressed address without any leading zeros. This allows us to compare the existing certificates and decide if new certificates should be issued.
* Tests for the IP address validation and normalization and also a test which compares the OpenSSL generated certificate with our configuration to make sure the normalization works.

This should close #7217.

### Checklist

- [x] Write tests

### DIFF
--- a/certificate-manager/src/main/java/io/strimzi/certs/IpAndDnsValidation.java
+++ b/certificate-manager/src/main/java/io/strimzi/certs/IpAndDnsValidation.java
@@ -12,32 +12,37 @@ import java.util.regex.Pattern;
  */
 public class IpAndDnsValidation {
     // Pattern for matching IPv4 address (requires additional check for segments to be equal or less than 255 - this is done in the isValidIpv4Address method)
-    private static final Pattern IPV4_ADDRESS = Pattern.compile("\\d{1,3}\\.\\d{1,3}\\.\\d{1,3}\\.\\d{1,3}");
+    private static final Pattern IPV4_ADDRESS = Pattern.compile("^\\d{1,3}\\.\\d{1,3}\\.\\d{1,3}\\.\\d{1,3}$");
 
     // Pattern for uncompressed IPv6 address -> case-insensitive, tolerates leading 0s
     private static final Pattern IPV6_ADDRESS = Pattern.compile("^(?:[\\da-fA-F]{1,4}:){7}[\\da-fA-F]{1,4}$");
 
-    // Pattern for compressed IPv6 address -> case-insensitive, tolerates leading 0s
+    // Pattern for compressed IPv6 address -> case-insensitive, tolerates leading 0s (requires additional check for number of segments)
     private static final Pattern IPV6_COMPRESSED_ADDRESS = Pattern.compile("^((?:[\\dA-Fa-f]{1,4}(?::[\\dA-Fa-f]{1,4})*)?)::((?:[\\dA-Fa-f]{1,4}(?::[\\dA-Fa-f]{1,4})*)?)$");
 
     // Pattern for DNS names
-    private static final Pattern DNS_NAME = Pattern.compile("^(" +
-            // a single char dns name
-            "[a-zA-Z\\d]|" +
-            // can't begin or end with -                followed by more labels of same
-            "[a-zA-Z\\d][a-zA-Z\\d\\-]{0,61}[a-zA-Z\\d])(\\.([a-zA-Z\\d]|[a-zA-Z\\d][a-zA-Z\\d\\-]{0,61}[a-zA-Z\\d]))*$");
+    private static final Pattern DNS_NAME;
+    static {
+        String dnsLabel = "(" +
+                // a single char dns name (can't begin or end with)
+                "[a-zA-Z\\d]|" +
+                // followed by more labels of same
+                "[a-zA-Z\\d][a-zA-Z\\d\\-]{0,61}[a-zA-Z\\d])";
+        DNS_NAME = Pattern.compile("^" + dnsLabel + "(\\." + dnsLabel + ")*");
+    }
 
     // Pattern used to normalize the segments of the IPv6 address
     private static final Pattern IPV6_LEADING_ZEROS_TRIMMING = Pattern.compile("(^|:)0+([\\da-f]+)(:|$)");
 
     /**
-     * Returns true if the name is a valid DNS name. That means that it matches the DNS_NAME pattern
+     * Returns true if the name is a valid DNS name or a wildcard DNS name. That means that it matches the DNS_NAME
+     * pattern or starts with *. followed up with a string which matches the DNS_PATTERN.
      *
      * @param name   Name which should be validated
      *
      * @return  True if the name is valid DNS name. False otherwise.
      */
-    public static boolean isValidDnsName(String name) {
+    public static boolean isValidDnsNameOrWildcard(String name) {
         return name.length() <= 255
                 && (DNS_NAME.matcher(name).matches()
                 || (name.startsWith("*.") && DNS_NAME.matcher(name.substring(2)).matches()));
@@ -85,23 +90,26 @@ public class IpAndDnsValidation {
      * @return  True if the name is valid IPv4 address. False otherwise.
      */
     public static boolean isValidIpv6Address(String ip) {
-        return IPV6_ADDRESS.matcher(ip).matches() || IPV6_COMPRESSED_ADDRESS.matcher(ip).matches();
+        return IPV6_ADDRESS.matcher(ip).matches()
+                // Includes additional segment count check for the compressed address
+                || (IPV6_COMPRESSED_ADDRESS.matcher(ip).matches() && ip.split(":").length > 0 && ip.split(":").length <= 7);
     }
 
     /**
-     * Normalizes the IPv6 address. Normalized address in Strimzi means that the IPv6 address is:
+     * Normalizes the IPv6 address. The address passed to thus method should be already validated. Normalized address
+     * in Strimzi means that the IPv6 address is:
      *     - uncompressed
      *     - without leading 0
      *     - lowercase
      * This is based on the OpenSSL output. OpenSSL will normalize the IPv6 addresses in SANs this way. And we need to
      * be able to compare the SANs to know when the SANs changed and a new cert should be generated.
      *
-     * @param ip    IPv6 address which should be normalized
+     * @param validIp    IPv6 address which should be normalized. The IPv6 address passed here should be already validated.
      *
      * @return  Normalized IPv6 address
      */
-    public static String normalizeIpv6Address(String ip)   {
-        String normalized = ip;
+    public static String normalizeIpv6Address(String validIp)   {
+        String normalized = validIp;
 
         // Normalize compressed address
         // We insert the required number of 0 segments instead of :: to uncompress the address
--- a/certificate-manager/src/main/java/io/strimzi/certs/IpAndDnsValidation.java
+++ b/certificate-manager/src/main/java/io/strimzi/certs/IpAndDnsValidation.java
@@ -1,0 +1,130 @@
+/*
+ * Copyright Strimzi authors.
+ * License: Apache License 2.0 (see the file LICENSE or http://apache.org/licenses/LICENSE-2.0.html).
+ */
+package io.strimzi.certs;
+
+import java.util.Locale;
+import java.util.regex.Pattern;
+
+/**
+ * Helper class to help with validation of IP addresses and DNS addresses
+ */
+public class IpAndDnsValidation {
+    // Pattern for matching IPv4 address (requires additional check for segments to be equal or less than 255 - this is done in the isValidIpv4Address method)
+    private static final Pattern IPV4_ADDRESS = Pattern.compile("\\d{1,3}\\.\\d{1,3}\\.\\d{1,3}\\.\\d{1,3}");
+
+    // Pattern for uncompressed IPv6 address -> case-insensitive, tolerates leading 0s
+    private static final Pattern IPV6_ADDRESS = Pattern.compile("^(?:[\\da-fA-F]{1,4}:){7}[\\da-fA-F]{1,4}$");
+
+    // Pattern for compressed IPv6 address -> case-insensitive, tolerates leading 0s
+    private static final Pattern IPV6_COMPRESSED_ADDRESS = Pattern.compile("^((?:[\\dA-Fa-f]{1,4}(?::[\\dA-Fa-f]{1,4})*)?)::((?:[\\dA-Fa-f]{1,4}(?::[\\dA-Fa-f]{1,4})*)?)$");
+
+    // Pattern for DNS names
+    private static final Pattern DNS_NAME = Pattern.compile("^(" +
+            // a single char dns name
+            "[a-zA-Z\\d]|" +
+            // can't begin or end with -                followed by more labels of same
+            "[a-zA-Z\\d][a-zA-Z\\d\\-]{0,61}[a-zA-Z\\d])(\\.([a-zA-Z\\d]|[a-zA-Z\\d][a-zA-Z\\d\\-]{0,61}[a-zA-Z\\d]))*$");
+
+    // Pattern used to normalize the segments of the IPv6 address
+    private static final Pattern IPV6_LEADING_ZEROS_TRIMMING = Pattern.compile("(^|:)0+([\\da-f]+)(:|$)");
+
+    /**
+     * Returns true if the name is a valid DNS name. That means that it matches the DNS_NAME pattern
+     *
+     * @param name   Name which should be validated
+     *
+     * @return  True if the name is valid DNS name. False otherwise.
+     */
+    public static boolean isValidDnsName(String name) {
+        return name.length() <= 255
+                && (DNS_NAME.matcher(name).matches()
+                || (name.startsWith("*.") && DNS_NAME.matcher(name.substring(2)).matches()));
+    }
+
+    /**
+     * Returns true if the name is a valid IPv4 or IPv6 address.
+     *
+     * @param ip    IP address which should be validated
+     *
+     * @return  True if the name is valid IP address. False otherwise.
+     */
+    public static boolean isValidIpAddress(String ip) {
+        return isValidIpv4Address(ip) || isValidIpv6Address(ip);
+    }
+
+    /**
+     * Returns true if the name is a valid IPv4 address. Valid IPv4 address is an address which matches the IPV6_ADDRESS
+     * or IPV6_COMPRESSED_ADDRESS pattern.
+     *
+     * @param ip    IP address which should be validated
+     *
+     * @return  True if the name is valid IPv6 address. False otherwise.
+     */
+    public static boolean isValidIpv4Address(String ip) {
+        boolean matches = IPV4_ADDRESS.matcher(ip).matches();
+        if (matches) {
+            String[] split = ip.split("\\.");
+            for (String num : split) {
+                int i = Integer.parseInt(num);
+                if (i > 255) {
+                    return false;
+                }
+            }
+        }
+        return matches;
+    }
+
+    /**
+     * Returns true if the name is a valid IPv6 address. Valid IPv6 address is an address which matches the IPV4_ADDRESS
+     * pattern and each segment is equal or less than 255.
+     *
+     * @param ip    IP address which should be validated
+     *
+     * @return  True if the name is valid IPv4 address. False otherwise.
+     */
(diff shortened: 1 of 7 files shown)


scholzj:
_Note: This can be partially tested on any Kube cluster -> you can just add the IPv6 address as a `alternativeName` for example. In addition to that, I also tested this in my AWS cluster with IPv6 support._

certificate-manager/src/main/java/io/strimzi/certs/IpAndDnsValidation.java
```diff
@@ -0,0 +1,130 @@
...
+
+    // Pattern for compressed IPv6 address -> case-insensitive, tolerates leading 0s
+    private static final Pattern IPV6_COMPRESSED_ADDRESS = Pattern.compile("^((?:[\\dA-Fa-f]{1,4}(?::[\\dA-Fa-f]{1,4})*)?)::((?:[\\dA-Fa-f]{1,4}(?::[\\dA-Fa-f]{1,4})*)?)$");
+
+    // Pattern for DNS names
+    private static final Pattern DNS_NAME = Pattern.compile("^(" +
+            // a single char dns name
+            "[a-zA-Z\\d]|" +
+            // can't begin or end with -                followed by more labels of same
+            "[a-zA-Z\\d][a-zA-Z\\d\\-]{0,61}[a-zA-Z\\d])(\\.([a-zA-Z\\d]|[a-zA-Z\\d][a-zA-Z\\d\\-]{0,61}[a-zA-Z\\d]))*$");
```
tombentley, line 28: This is a monster! I wonder if it would be more readable if you defined a 
```java
static {
    String dnsLabel = "(" +
            // a single char dns name
            "[a-zA-Z\\d]|" +
            // can't begin or end with -                followed by more labels of same
            "[a-zA-Z\\d][a-zA-Z\\d\\-]{0,61}[a-zA-Z\\d])"; 
    DNS_NAME = Pattern.compile("^" + dnsLabel + "(\\." + dnsLabel + ")*");
}

File: certificate-manager/src/main/java/io/strimzi/certs/IpAndDnsValidation.java [hunk 2]
```diff
@@ -0,0 +1,130 @@
...
+ * Helper class to help with validation of IP addresses and DNS addresses
+ */
+public class IpAndDnsValidation {
+    // Pattern for matching IPv4 address (requires additional check for segments to be equal or less than 255 - this is done in the isValidIpv4Address method)
+    private static final Pattern IPV4_ADDRESS = Pattern.compile("\\d{1,3}\\.\\d{1,3}\\.\\d{1,3}\\.\\d{1,3}");
+
+    // Pattern for uncompressed IPv6 address -> case-insensitive, tolerates leading 0s
+    private static final Pattern IPV6_ADDRESS = Pattern.compile("^(?:[\\da-fA-F]{1,4}:){7}[\\da-fA-F]{1,4}$");
+
+    // Pattern for compressed IPv6 address -> case-insensitive, tolerates leading 0s
```
tombentley, line 20: Like the IPv4 case, I think it's worth commenting what this pattern doesn't validate. Presumably there's a bound on the number of groups before and after the `::`?

certificate-manager/src/main/java/io/strimzi/certs/IpAndDnsValidation.java
```diff
@@ -0,0 +1,130 @@
...
+    private static final Pattern IPV6_COMPRESSED_ADDRESS = Pattern.compile("^((?:[\\dA-Fa-f]{1,4}(?::[\\dA-Fa-f]{1,4})*)?)::((?:[\\dA-Fa-f]{1,4}(?::[\\dA-Fa-f]{1,4})*)?)$");
+
+    // Pattern for DNS names
```
tombentley, line 23: Again, useful to mention what's not validated by this pattern.

certificate-manager/src/main/java/io/strimzi/certs/IpAndDnsValidation.java
```diff
@@ -0,0 +1,130 @@
...
+    private static final Pattern DNS_NAME = Pattern.compile("^(" +
+            // a single char dns name
+            "[a-zA-Z\\d]|" +
+            // can't begin or end with -                followed by more labels of same
+            "[a-zA-Z\\d][a-zA-Z\\d\\-]{0,61}[a-zA-Z\\d])(\\.([a-zA-Z\\d]|[a-zA-Z\\d][a-zA-Z\\d\\-]{0,61}[a-zA-Z\\d]))*$");
+
+    // Pattern used to normalize the segments of the IPv6 address
+    private static final Pattern IPV6_LEADING_ZEROS_TRIMMING = Pattern.compile("(^|:)0+([\\da-f]+)(:|$)");
```
tombentley, line 31: A segment cannot be all zeros?

certificate-manager/src/main/java/io/strimzi/certs/IpAndDnsValidation.java
```diff
@@ -0,0 +1,130 @@
...
+
+    /**
+     * Returns true if the name is a valid DNS name. That means that it matches the DNS_NAME pattern
```
tombentley, line 34: Add a link the the RFC that defines the grammar for DNS names, perhaps?

certificate-manager/src/main/java/io/strimzi/certs/IpAndDnsValidation.java
```diff
@@ -0,0 +1,130 @@
...
+     *
+     * @param name   Name which should be validated
+     *
+     * @return  True if the name is valid DNS name. False otherwise.
+     */
+    public static boolean isValidDnsName(String name) {
+        return name.length() <= 255
+                && (DNS_NAME.matcher(name).matches()
+                || (name.startsWith("*.") && DNS_NAME.matcher(name.substring(2)).matches()));
```
tombentley, line 43: Add to Javadoc that is matches wildcards. And probably the method name should be something like `isDnsNameOrWildcard`

certificate-manager/src/test/java/io/strimzi/certs/IpAndDnsValidationTest.java
```diff
@@ -0,0 +1,61 @@
...
+package io.strimzi.certs;
+
+import org.junit.jupiter.api.Test;
+
+import static org.hamcrest.CoreMatchers.is;
+import static org.hamcrest.MatcherAssert.assertThat;
+
+public class IpAndDnsValidationTest {
+    @Test
+    public void testIPv6()   {
```
tombentley, line 14: Other cases:
* A group with > 4 chars
* Presence of a triple `:::`
* Too many/few groups in presence of `::`

certificate-manager/src/test/java/io/strimzi/certs/IpAndDnsValidationTest.java
```diff
@@ -0,0 +1,61 @@
...
+        assertThat(IpAndDnsValidation.isValidIpv6Address("::1"), is(true));
+        assertThat(IpAndDnsValidation.isValidIpv6Address("fc01::8d1c"), is(true));
+        assertThat(IpAndDnsValidation.isValidIpv6Address("::fc01:8d1c"), is(true));
+        assertThat(IpAndDnsValidation.isValidIpv6Address("fc01:8d1c::"), is(true));
+        assertThat(IpAndDnsValidation.isValidIpv6Address("1762:0:0:0:0:B03:1:AF18"), is(true));
+
+        assertThat(IpAndDnsValidation.isValidIpv6Address("fc01::8j1c"), is(false));
+        assertThat(IpAndDnsValidation.isValidIpv6Address("fc01::176j::8j1c"), is(false));
```
tombentley, line 22: You tested a non-hex char above. I think here you want to test with valid hex chars only and that > 1 `::` is rejected.

certificate-manager/src/test/java/io/strimzi/certs/IpAndDnsValidationTest.java
```diff
@@ -0,0 +1,61 @@
...
+        assertThat(IpAndDnsValidation.isValidIpv6Address("176J:0:0:0:0:B03:1:AF18"), is(false));
+        assertThat(IpAndDnsValidation.isValidIpv6Address("1762:0:0:0:0:1:B03:1:AF18"), is(false));
+    }
+
+    @Test
+    public void testIPv4()   {
+        assertThat(IpAndDnsValidation.isValidIpv4Address("127.0.0.1"), is(true));
+        assertThat(IpAndDnsValidation.isValidIpv4Address("123.123.123.123"), is(true));
+
+        assertThat(IpAndDnsValidation.isValidIpv4Address("127.0.0.0.1"), is(false));
```
tombentley, line 32: Also test a triplet

certificate-manager/src/main/java/io/strimzi/certs/IpAndDnsValidation.java
```diff
@@ -0,0 +1,130 @@
...
+     *     - without leading 0
+     *     - lowercase
+     * This is based on the OpenSSL output. OpenSSL will normalize the IPv6 addresses in SANs this way. And we need to
+     * be able to compare the SANs to know when the SANs changed and a new cert should be generated.
+     *
+     * @param ip    IPv6 address which should be normalized
+     *
+     * @return  Normalized IPv6 address
+     */
+    public static String normalizeIpv6Address(String ip)   {
```
tombentley, line 103: Is it a precondition of this method that `ip` has been validated already? If so we should document that (and maybe call `ip` `validId` instead). Else perhaps we should check it here?

certificate-manager/src/test/java/io/strimzi/certs/IpAndDnsValidationTest.java
```diff
@@ -0,0 +1,61 @@
...
+        assertThat(IpAndDnsValidation.isValidIpv4Address("321.321.321.321"), is(false));
+        assertThat(IpAndDnsValidation.isValidIpv4Address("some.domain.name"), is(false));
+    }
+
+    @Test
+    public void testDnsNames()   {
+        assertThat(IpAndDnsValidation.isValidDnsName("example"), is(true));
+        assertThat(IpAndDnsValidation.isValidDnsName("example.com"), is(true));
+
+        assertThat(IpAndDnsValidation.isValidDnsName("example:com"), is(false));
```
tombentley, line 42: Test too many chars

certificate-manager/src/main/java/io/strimzi/certs/IpAndDnsValidation.java
scholzj: This is used in the normalization. In general, the segment can be `:0:`, `:00:` or `:0000:` and it all means the same. Similarly, you can be `:00b3`, :`0b3:` or `:b3:` and they are the same. So the normalization removes the leading zeros to end up always with e.g. `:0:` or `:b3`. This allows us to compare the addresses and decide if new cert is needed.

certificate-manager/src/main/java/io/strimzi/certs/IpAndDnsValidation.java
scholzj: I'm a bit reluctant to do this. The IPv4 validation or the DNS validation are not new. They were just moved to not have them across 3 different classes. I addressed the other comments you had for them. But I do not really know what relation does this have to the RFC and how well it covers it.

certificate-manager/src/main/java/io/strimzi/certs/IpAndDnsValidation.java
tombentley: But don't _think_ this will match `:00:` because of the `[\\da-f]+`. The `0+` matches greedily, so the `\\d` doesn't match the 2nd `0`.

scholzj: I have test with `00` I can add one more with `000` as well if needed. It seems to work fine. My understanding of the reg-exp is that yes, `0+` matches greedily. But the `[\\da-f]+` does as well. So if all the zeros would go to `0+`, it would not match the expression at all. So when you have only multiple zeroes and nothing else, the last zero always goes to `[\\da-f]+`. E.g. with `:000:`, the first two zeroes will go to `0+` and the last one stays with `[\\da-f]+`.

scholzj: I added the test here: https://github.com/scholzj/strimzi-kafka-operator/blob/6ad18065ba278796984e23442fad5a9aca012c3d/certificate-manager/src/test/java/io/strimzi/certs/IpAndDnsValidationTest.java#L66

tombentley: OK, I believe you. I must have messed up my "test" (which was to use the regex search in IDEA).

tombentley (approved):
Thanks Jakub!

PR #7242, opened 2022-08-26
Add metrics to StrimziPodSetController

scholzj:
### Type of change

- Enhancement / new feature

### Description

This PR adds metrics to the `StrimziPodSetController` The metrics follow the metrics already provided by out `*AssemblyOperator` classes and deviate only where necessary:
* The Controllers do not use resource locking. So this metric is not used for StrimziPodSets
* The Controller use work queue and it might happen that the resource whcih should be enqueued for some event is already in the queue. There is a new metric provided for this event. Similarly to locks, this might indicate a problems, but can also happen in a regular run.
* The resource state metrics are not exposed. First of all because they make the code very complicated (I might follow with a proposal to remove the existing ones in operators later) and second because the StrimziPodSets do not have a clear ready / not ready state. Some pods might be ready, some not etc.

This Pr also refactors how metrics are done in the `*AssemblyOperator` classes and in the `Operator` interface. The metrics used to be hardcoded in these classes. In this PR, I have created a new _metrics holder_ pattern. `AbstractMetricsHolder` and its subclasses are used to hold the metrics in a separate class. Since the metrics differ a bit between different operators and controllers, it creates a hierarchy:

```mermaid
classDiagram
    AbstractMetricsHolder <|-- ControllerMetricsHolder
    AbstractMetricsHolder <|-- OperatorMetricsHolder
    OperatorMetricsHolder <|-- ConnectOperatorMetricsHolder
    AbstractMetricsHolder: Shared metrics
    ControllerMetricsHolder: Already Queued metric
    OperatorMetricsHolder: Without Lock metric
    ConnectOperatorMetricsHolder: Connector metrics
```

I believe this nicely separates the metrics from the operator and controller logic. It also allows us to share the code between the classes.

This PR also does some minor refactoring of the `StrimziPodSetController`. Apart from adding the metrics, it also separates some calls to new methods, setups the informer handlers only in later phase etc. The StrimziPodSet metrics were developed as part of the User Operator rewrite. A separate PR is opened for the StrimziPodSet metrics to simplify review and testing via more smaller PRs. So these changes are not completely related to the metrics themselves, but are generally useful and separating them into yet another PR would be too complicated.

### Checklist

- [x] Write tests
- [x] Update CHANGELOG.md

### DIFF
--- a/operator-common/src/main/java/io/strimzi/operator/common/metrics/AbstractMetricsHolder.java
+++ b/operator-common/src/main/java/io/strimzi/operator/common/metrics/AbstractMetricsHolder.java
@@ -1,0 +1,230 @@
+/*
+ * Copyright Strimzi authors.
+ * License: Apache License 2.0 (see the file LICENSE or http://apache.org/licenses/LICENSE-2.0.html).
+ */
+package io.strimzi.operator.common.metrics;
+
+import io.micrometer.core.instrument.Counter;
+import io.micrometer.core.instrument.Tag;
+import io.micrometer.core.instrument.Tags;
+import io.micrometer.core.instrument.Timer;
+import io.strimzi.operator.common.MetricsProvider;
+import io.strimzi.operator.common.model.Labels;
+
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.atomic.AtomicInteger;
+import java.util.function.Function;
+
+/**
+ * Util class which holds the different metrics used by controllers
+ */
+public abstract class AbstractMetricsHolder {
+    protected static final String METRICS_PREFIX = "strimzi.";
+
+    protected final String kind;
+    protected final Labels selectorLabels;
+    protected final MetricsProvider metricsProvider;
+
+    protected final Map<String, AtomicInteger> resourceCounterMap = new ConcurrentHashMap<>(1);
+    protected final Map<String, AtomicInteger> pausedResourceCounterMap = new ConcurrentHashMap<>(1);
+    private final Map<String, Counter> periodicReconciliationsCounterMap = new ConcurrentHashMap<>(1);
+    private final Map<String, Counter> reconciliationsCounterMap = new ConcurrentHashMap<>(1);
+    private final Map<String, Counter> failedReconciliationsCounterMap = new ConcurrentHashMap<>(1);
+    private final Map<String, Counter> successfulReconciliationsCounterMap = new ConcurrentHashMap<>(1);
+    private final Map<String, Timer> reconciliationsTimerMap = new ConcurrentHashMap<>(1);
+
+    /**
+     * Constructs the metrics holder
+     *
+     * @param kind              Kind of the resources for which these metrics apply
+     * @param selectorLabels    Selector labels to select the controller resources
+     * @param metricsProvider   Metrics provider
+     */
+    public AbstractMetricsHolder(String kind, Labels selectorLabels, MetricsProvider metricsProvider) {
+        this.kind = kind;
+        this.selectorLabels = selectorLabels;
+        this.metricsProvider = metricsProvider;
+    }
+
+    /**
+     * Metrics provider used for the metrics by this holder class
+     *
+     * @return  Metrics provider
+     */
+    public MetricsProvider metricsProvider()    {
+        return metricsProvider;
+    }
+
+    ////////////////////
+    // Methods for individual counters
+    ////////////////////
+
+    /**
+     * Counter metric for number of periodic reconciliations. It should be incremented only once per timer-trigger. It
+     * should not be incremented for every resource found by the periodical reconciliation.
+     *
+     * @param namespace     Namespace of the resources being reconciled
+     *
+     * @return  Metrics counter
+     */
+    public Counter periodicReconciliationsCounter(String namespace) {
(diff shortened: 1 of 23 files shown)


operator-common/src/main/java/io/strimzi/operator/common/metrics/AbstractMetricsHolder.java
```diff
@@ -0,0 +1,230 @@
...
+import io.strimzi.operator.common.MetricsProvider;
+import io.strimzi.operator.common.model.Labels;
+
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.atomic.AtomicInteger;
+import java.util.function.Function;
+
+/**
+ * Util class which holds the different metrics used by controllers
```
tombentley, line 20:
```suggestion
 * Abstract base class holding common metrics used by operators and controllers.
 * Subclasses can add more specialized metrics. 
```

operator-common/src/main/java/io/strimzi/operator/common/metrics/AbstractMetricsHolder.java
```diff
@@ -0,0 +1,230 @@
...
+     *
+     * @return  Metrics counter
+     */
+    public Counter successfulReconciliationsCounter(String namespace) {
+        return getCounter(namespace, kind, METRICS_PREFIX + "reconciliations.successful", metricsProvider, selectorLabels, successfulReconciliationsCounterMap,
+                "Number of reconciliations done by the operator for individual resources which were successful");
+    }
+
+    /**
+     * Counter metric for number of resources.
```
tombentley, line 114: Can we clarify tha this is the number of consumed resources (i.e. CRs), rather than some counter over all resources (CRs plus the output/derived resources).

operator-common/src/main/java/io/strimzi/operator/common/metrics/AbstractMetricsHolder.java
```diff
@@ -0,0 +1,230 @@
...
+     * @param kind              Kind of the resource
+     * @param metricName        Name of the metric
+     * @param metrics           Metrics provider
+     * @param selectorLabels    Selector labels used to filter the resources
+     * @param counterMap        Map with counters
+     * @param metricHelp        Help description of the metric
+     *
+     * @return  Counter metric
+     */
+    public static Counter getCounter(String namespace, String kind, String metricName, MetricsProvider metrics, Labels selectorLabels, Map<String, Counter> counterMap, String metricHelp) {
```
tombentley, line 193: Do these factory methods need to be `public`, or could they be `protected`?

operator-common/src/main/java/io/strimzi/operator/common/metrics/ControllerMetricsHolder.java
```diff
@@ -0,0 +1,44 @@
...
+
+import io.micrometer.core.instrument.Counter;
+import io.strimzi.operator.common.MetricsProvider;
+import io.strimzi.operator.common.model.Labels;
+
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+
+/**
+ * Util class which holds the different metrics used by controllers
```
tombentley, line 15:
```suggestion
 * A metrics holder for controllers. 
```

operator-common/src/main/java/io/strimzi/operator/common/metrics/OperatorMetricsHolder.java
```diff
@@ -0,0 +1,52 @@
...
+
+import io.micrometer.core.instrument.Counter;
+import io.strimzi.operator.common.MetricsProvider;
+import io.strimzi.operator.common.model.Labels;
+
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+
+/**
+ * Util class which holds the different metrics used by operators
```
tombentley, line 15:
```suggestion
 * A metrics holder for operators.
```

operator-common/src/main/java/io/strimzi/operator/common/metrics/AbstractMetricsHolder.java
```diff
@@ -0,0 +1,230 @@
...
+
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.atomic.AtomicInteger;
+import java.util.function.Function;
+
+/**
+ * Util class which holds the different metrics used by controllers
+ */
+public abstract class AbstractMetricsHolder {
```
tombentley, line 22: IDK whether we really need to put `Abstract` in the class name. I.e. why not just `MetricsHolder`?

operator-common/src/main/java/io/strimzi/operator/common/operator/resource/ResourceDiff.java
```diff
@@ -12,9 +12,13 @@
 
 import java.util.regex.Pattern;
 
-class ResourceDiff<T extends HasMetadata> extends AbstractJsonDiff {
+public class ResourceDiff<T extends HasMetadata> extends AbstractJsonDiff {
     private static final ReconciliationLogger LOGGER = ReconciliationLogger.create(ResourceDiff.class.getName());
 
+    public static final Pattern DEFAULT_IGNORABLE_PATHS = Pattern.compile(
```
tombentley, line 18: Does anything actually use this field?

scholzj:
@tombentley Thanks. Your comments should be now all incorporated.

tombentley (approved):
Thanks!

Issue #7138, opened 2022-08-02
Make it easier to wait for custom resource readiness when using the Strimzi `api` module


scholzj:
When using the Strimzi `api` module from other parts of Strimzi or from outside applications just using it as an dependency, one of the common tasks is to wait for the custom resource to get ready. This can be easily done fairly easily directly in the code. For example for a Kafka cluster, it would be:

```java
Crds.kafkaOperation(client).inNamespace(NAMESPACE).withName(NAME).waitUntilCondition(k -> {
    if (k.getStatus() != null && k.getStatus().getConditions() != null) {
        return updatedKafka.getMetadata().getGeneration() == k.getStatus().getObservedGeneration()
                && k.getStatus().getConditions().stream().anyMatch(c -> "Ready".equals(c.getType()) && "True".equals(c.getStatus()));
    } else {
        return false;
    }
}, 5, TimeUnit.MINUTES);
```

But it might be nicer to have this directly in the `api` module itself to simplify the application code and allow the users of the `api` module to call for example only something like:

```java
Crds.kafkaOperation(client).inNamespace(NAMESPACE).withName(NAME).waitUntilCondition(Kafka.readyCondition(), 5, TimeUnit.MINUTES);
```

This will make it this code part of Strimzi so we can improve it if something changes in the Strimzi itself. It will be also easier to use for users. We can also cover different states (e.g. the `KafkaRebalance` CR can support different conditions for different states such as ProposalReady etc.)

scholzj:
Triaged on 25.8.2022: Should be kept and implemented. Might be a good first issue.

PR #7247, opened 2022-08-28
Add CustomResourceConditions containing predicates to await CRD readiness

robobario:
### Type of change

- Enhancement / new feature

### Description

Waiting for CRDs to become ready takes a few lines of boilerplate code. This should make it easier for users to wait for programmatically created resources to become ready (or enter other desired states)

See [strimzi-api-examples](https://github.com/scholzj/strimzi-api-examples/compare/main...robobario:strimzi-api-examples:wait-condition-api?expand=1) fork for usage (updated to 0.31.0-SNAPSHOT, fixed up some fabric8 6.0.0 changes, added cruisecontrol rebalance code).

fixes issue #7138

### Checklist

- [x] Write tests
- [ ] Update documentation
- [x] Update CHANGELOG.md
- [NA] Supply screenshots for visual changes, such as Grafana dashboards

### DIFF
--- a/api/src/main/java/io/strimzi/api/kafka/model/CustomResourceConditions.java
+++ b/api/src/main/java/io/strimzi/api/kafka/model/CustomResourceConditions.java
@@ -32,33 +32,13 @@ class CustomResourceConditions {
     private CustomResourceConditions() {
     }
 
-    /**
-     * Returns a predicate that determines if a CustomResource is ready. A CustomResource is
-     * ready if the observedGeneration of its status is equal to the generation of its metadata
-     * and any of the conditions of its status has type:"Ready" and status:"True"
-     *
-     * @param <Y> the type of the custom resources Status
-     * @param <T> the type of the custom resource
-     * @return a predicate that checks if a CustomResource is ready
-     */
     static <Y extends Status, T extends CustomResource<?, Y>> Predicate<T> isReady() {
         return isLatestGenerationAndAnyConditionMatches("Ready", "True");
     }
 
-    /**
-     * Returns a predicate that determines if a CustomResource's status is the latest generation
-     * and any of the conditions of its status matches the type and status specified.
-     *
-     * @param <Y>    the type of the custom resources Status
-     * @param <T>    the type of the custom resource
-     * @param type   the Type of the condition we expect to be present in the status
-     * @param status the Status of the condition we expect to be present in the status
-     * @return a predicate that checks if a CustomResource is ready
-     */
     static <Y extends Status, T extends CustomResource<?, Y>> Predicate<T> isLatestGenerationAndAnyConditionMatches(String type, String status) {
         return isStatusLatestGenerationAndMatches(anyCondition(type, status));
     }
-
 
     private static <Y extends Status> Predicate<Y> anyCondition(String expectedType, String expectedStatus) {
         return (status) -> {
--- a/api/src/main/java/io/strimzi/api/kafka/model/CustomResourceConditions.java
+++ b/api/src/main/java/io/strimzi/api/kafka/model/CustomResourceConditions.java
@@ -21,10 +21,10 @@ import java.util.function.Predicate;
  * Crds.kafkaOperation(client).inNamespace(NAMESPACE).resource(kafka).create();
  * Crds.kafkaOperation(client).inNamespace(NAMESPACE).withName(NAME).waitUntilCondition(Kafka.isReady(), 5, TimeUnit.MINUTES);
  * </pre>
- * and to wait on a specific KafkaRebalance state using {@link io.strimzi.api.kafka.model.KafkaRebalance#isState(KafkaRebalanceState) KafkaRebalance::isState}:
+ * and to wait on a specific KafkaRebalance state using {@link io.strimzi.api.kafka.model.KafkaRebalance#isInState(KafkaRebalanceState) KafkaRebalance::isInState}:
  * <pre>
  * Crds.kafkaRebalanceOperation(client).inNamespace(NAMESPACE).withName(REBALANCE_NAME)
- *     .waitUntilCondition(KafkaRebalance.isState(KafkaRebalanceState.ProposalReady), 5, TimeUnit.MINUTES);
+ *     .waitUntilCondition(KafkaRebalance.isInState(KafkaRebalanceState.ProposalReady), 5, TimeUnit.MINUTES);
  * </pre>
  */
 class CustomResourceConditions {
--- a/api/src/main/java/io/strimzi/api/kafka/model/CustomResourceConditions.java
+++ b/api/src/main/java/io/strimzi/api/kafka/model/CustomResourceConditions.java
@@ -5,11 +5,32 @@
 package io.strimzi.api.kafka.model;
 
 import io.fabric8.kubernetes.client.CustomResource;
+import io.strimzi.api.kafka.model.balancing.KafkaRebalanceState;
 import io.strimzi.api.kafka.model.status.Status;
 
 import java.util.function.Predicate;
 
-public class CustomResourceConditions {
+/**
+ * CustomResourceConditions supplies convenience predicates that are intended to reduce boilerplate
+ * code when using fabric8 clients to wait for a CRD to enter some anticipated state. This class provides
+ * a generic implementation that can be used internally, public Predicates are exposed on the CRD
+ * classes.
+ * <p>
+ * An example usage of one of the public predicates ({@link io.strimzi.api.kafka.model.Kafka#isReady() Kafka::isReady}):
+ * <pre>
+ * Crds.kafkaOperation(client).inNamespace(NAMESPACE).resource(kafka).create();
+ * Crds.kafkaOperation(client).inNamespace(NAMESPACE).withName(NAME).waitUntilCondition(Kafka.isReady(), 5, TimeUnit.MINUTES);
+ * </pre>
+ * and to wait on a specific KafkaRebalance state using {@link io.strimzi.api.kafka.model.KafkaRebalance#isState(KafkaRebalanceState) KafkaRebalance::isState}:
+ * <pre>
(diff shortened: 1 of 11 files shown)


robobario:
As part of this I've exposed a signature like `public static <Y extends Status, T extends CustomResource<?, Y>> Predicate<T> isReady()`. This can be used across the different CustomResource types, but I'm wondering if it's a bit too tricky of a signature to be public; can we rely on the types continuing to extend CustomResource, what if the fabric8 api changed. 

Maybe it would be better to have a method on each CRD type (as suggested in the issue) like `Predicate<Kafka> isReady()` on `Kafka` etc. Offering up a Predicate<Kafka> might be easier to support? We can still use this generic code, but hide it away.

api/src/main/java/io/strimzi/api/kafka/model/CustomResourceConditions.java
```diff
@@ -0,0 +1,46 @@
...
+import java.util.function.Predicate;
+
+public class CustomResourceConditions {
+
+    public static <Y extends Status, T extends CustomResource<?, Y>> Predicate<T> isReady() {
+        return isStatusLatestGenerationAndMatches(anyCondition("Ready", "True"));
+    }
+
+
+    public static Predicate<KafkaRebalance> isProposalReady() {
```
scholzj, line 19: I wonder if we should give it a more clear name ... `isKafkaRebalanceProposalReady` => to make it more clear what it does, since ProposalReady applies only there.

scholzj (commented):
Thanks for the PR. I left one nit, but looks good otherwise.

Could you please fix the DCO sign-off? The instructions should be under the _Details_ link next to the DCO status.

scholzj:
One more thing, could you please add some Javadoc at least to the public methods?

robobario:
Added DCO signoff, javadoc and renamed that method to `isKafkaRebalanceProposalReady`

scholzj (approved):
Thanks

api/src/main/java/io/strimzi/api/kafka/model/CustomResourceConditions.java
```diff
@@ -0,0 +1,61 @@
...
+ * License: Apache License 2.0 (see the file LICENSE or http://apache.org/licenses/LICENSE-2.0.html).
+ */
+package io.strimzi.api.kafka.model;
+
+import io.fabric8.kubernetes.client.CustomResource;
+import io.strimzi.api.kafka.model.status.Status;
+
+import java.util.function.Predicate;
+
+public class CustomResourceConditions {
```
tombentley, line 12: We probably want a `private` no-args constructor so no one can instantiate this class.

api/src/main/java/io/strimzi/api/kafka/model/CustomResourceConditions.java
```diff
@@ -0,0 +1,61 @@
...
+    /**
+     * Returns a predicate that determines if a CustomResource is ready. A CustomResource is
+     * ready if the observedGeneration of its status is equal to the generation of its metadata
+     * and any of the conditions of its status has type:"Ready" and status:"True"
+     *
+     * @param <Y> the type of the custom resources Status
+     * @param <T> the type of the custom resource
+     * @return a predicate that checks if a CustomResource is ready
+     */
+    public static <Y extends Status, T extends CustomResource<?, Y>> Predicate<T> isReady() {
```
tombentley, line 23: This could be used with _any_ custom resource, even non-Strimzi ones (which might not have `Ready`, and therefore the returned predicate might never return true). To avoid this perhaps we should only expose public methods for Strimzi resources?

api/src/main/java/io/strimzi/api/kafka/model/CustomResourceConditions.java
tombentley, line 12: This class could do with at least a line of Javadoc explaining what it holds. But ideally also a code snippet so people can see+copy the intended usage.

api/src/main/java/io/strimzi/api/kafka/model/CustomResourceConditions.java
tombentley: I also wonder having these as `static` methods on the CR classes (or perhaps their status classes) might make for better discoverability?

api/src/main/java/io/strimzi/api/kafka/model/CustomResourceConditions.java
```diff
@@ -0,0 +1,61 @@
...
+    }
+
+    /**
+     * Returns a predicate that determines if a KafkaRebalance is in a ProposalReady state. A KafkaRebalance is
+     * in proposal ready if the observedGeneration of its status is equal to the generation of its metadata
+     * and any of the conditions of its status has type:"ProposalReady" and status:"True"
+     *
+     * @return a predicate that checks if a KafkaRebalance is in ProposalReady state
+     */
+    public static Predicate<KafkaRebalance> isKafkaRebalanceProposalReady() {
```
tombentley, line 34: Why are we only exposing something for the ProposalReady state and not for all the states of `KafkaRebalanceState`?

api/src/main/java/io/strimzi/api/kafka/model/CustomResourceConditions.java
robobario: Having a signature like `static Predicate<Kafka> isReady()` on the `Kafka` class could be easier to update in future if your definition of readiness changes or needed to be specialised per CR. I think it can't go on the `Status` classes as we need access to `metadata` to get at the `generation`.

api/src/main/java/io/strimzi/api/kafka/model/CustomResourceConditions.java
robobario: True, how about making the signature `isKafkaRebalanceState(KafkaRebalanceState state)` and using the enum name? Or if we moved it to KafkaRebalance, `isState`.

Also I was wondering if there should be any smarts around other terminal states, like you are waiting for `ProposalReady` but it enters `NotReady` instead due to an error. Then fabric8 would sit around until it hits the timeout.

tombentley: > True, how about making the signature `isKafkaRebalanceState(KafkaRebalanceState state)` and using the enum name? Or if we moved it to KafkaRebalance, `isState`.

Yeah, I think having a `state` parameter is a good idea. @scholzj how do you feel about putting these methods on the CR classes themselves?

> 
> Also I was wondering if there should be any smarts around other terminal states, like you are waiting for `ProposalReady` but it enters `NotReady` instead due to an error. Then fabric8 would sit around until it hits the timeout.

Doing that we would end up encoding assumptions about the current state machine within the predicate. Such assumptions could, _in general_, be invalidated by in the future by changes to the state macine. So I think we want to be a bit careful. 

The _specific_ suggestion you're making is: Should the predicate throw an exception if the SM enters a terminal state that's not the one for which the predicate will return true. The compatibility semantics of that depends on whether any states change their terminal-ness in the future. The Javadocs for `NotReady` suggest it's not a terminal state (can transition back to `New`). Overall I'm inclined to think it safest to let users handle this (e.g. by combining the predicates `ProposalReady && !NotReady`). But I wonder if @tomncooper @ppatierno have any thoughts about this.

scholzj: > Yeah, I think having a `state` parameter is a good idea. @scholzj how do you feel about putting these methods on the CR classes themselves?

I assume you mean classes such as `io.strimzi.api.kafka.model.Kafka`? I think that is fine with me. I don't have problem with that. I think the point you made about it being too generic was good.

robobario: I will refactor to that

api/src/main/java/io/strimzi/api/kafka/model/CustomResourceConditions.java
robobario: Added private constructor in latest commit

robobario: Added javadoc with a couple of example usages

api/src/main/java/io/strimzi/api/kafka/model/CustomResourceConditions.java
robobario: Moved the predicates to public static methods on the CR classes and made CustomResourceConditions package private

api/src/main/java/io/strimzi/api/kafka/model/KafkaRebalance.java
```diff
@@ -154,4 +156,8 @@ public String toString() {
             throw new RuntimeException(e);
         }
     }
+
+    public static Predicate<KafkaRebalance> isState(KafkaRebalanceState state) {
```
tombentley, line 160: `inInState()` ?

tombentley (approved):
The public methods on the CR classes still need Javadoc (you should be able to move the existing docs from `CustomResourceConditions`, which is no longer `public`).

Otherwise this LGTM. 

Thanks!

scholzj:
@robobario Just FYI: I resolved the conflict so that the tests can run. You might need to pull the updated branch first.

api/src/main/java/io/strimzi/api/kafka/model/KafkaRebalance.java
robobario: That makes more sense, changing now.

robobario: Done

robobario:
> The public methods on the CR classes still need Javadoc (you should be able to move the existing docs from `CustomResourceConditions`, which is no longer `public`).

javadoc moved to the public methods.

I'm finished for now. Do you want it rebased into a single commit before merge?

scholzj:
> I'm finished for now. Do you want it rebased into a single commit before merge?

You can if you want, but its not needed. We will squash-merge it.

PR #7260, opened 2022-09-01
[MO] - [system test] -> Client configured correctly in PSP and new te…

see-quick:
…st case

### Type of change

- Bugfix
- Enhancement / new feature

### Description

We do not configure our clients correctly to use `restricted` Pod Security profiles. This PR fixes problems with Clients and also adds one negative test case.

### Checklist

- [x] Write tests


Frawless (approved):

scholzj (approved):
LGTM assuming the tests pass.

see-quick:
> LGTM assuming the tests pass.

Yes, they did both on Minikube and OCP.
image

Plus I think there is no point in running Azure regression because we use their 1.16 Kubernetes version.

PR #7263, opened 2022-09-01
ST: Propagate configuration for pod policy provider into helm and olm resources during STs

Frawless:

### Type of change

- Bugfix

### Description

This PR propagate extra env vars into helm and OLM resources used within system tests module. This should fix issues with running system tests with pod security policy when we specify different than default policy also for helm and OLM.

### Checklist


scholzj (approved):

scholzj (approved):
LGTM assuming the tests pass.

Issue #1546, opened 2019-04-16
Consistent colours for pods in Grafana


tombentley:
**Describe the bug**

The provided grafana dashboards don't use a consistent colour across the charts for the same zookeeper node or kafka broker. This can make reading and correlating the charts more difficult than it needs to be. See https://github.com/grafana/grafana/issues/6662

**To Reproduce**
1. Just look at the default grafana dashboard provided for a ZK cluster with ≥ 2 nodes.

**Expected behavior**

You'll see zookeeper-_n_ has different colours in the different timeseries charts. So cyan might be zookeeper-0 in one chart but zookeeper-2 on another.

scholzj:
Triaged on 2.8.2022: This is still an issue in the current dashboards.

PR #7243, opened 2022-08-26
Fix/dashboards ds prometheus

giany:
### Type of change

_Select the type of your PR_

- Bugfix

### Description

When trying to add the Grafana dashboards programatically (via helm chart) this error appears:

```
Failed to upgrade legacy queries Datasource ${DS_PROMETHEUS} was not found
```
Seems it's related to this [issue](https://github.com/grafana/grafana/issues/10786). 
### Checklist

_Please go through this checklist and make sure all applicable tasks have been done_

Tested on EKS. No visual changes done.

PR #7299, opened 2022-09-09
Unify colours in Grafana graphs for pod records

Frawless:

### Type of change

- Bugfix

### Description

This PR unifies the colors across Grafana dashboards for all pods returned by Prometheus queries. 

**Kafka Dashboard**
image

**Zookeeper Dashboard**
image

**KafkaConnect Dashboard**
image

**KafkaMirrorMaker2 Dashboard**
image

Solves https://github.com/strimzi/strimzi-kafka-operator/issues/1546

### Checklist

### DIFF
--- a/packaging/examples/metrics/grafana-dashboards/strimzi-kafka-connect.json
+++ b/packaging/examples/metrics/grafana-dashboards/strimzi-kafka-connect.json
@@ -263,9 +263,9 @@
       "steppedLine": false,
       "targets": [
         {
-          "expr": "sum(rate(kafka_consumer_incoming_byte_total{strimzi_io_kind=~\"KafkaConnect.*\",strimzi_io_cluster=\"$strimzi_connect_cluster_name\"}[1m])) by (kubernetes_pod_name)",
+          "expr": "sum(rate(kafka_consumer_incoming_byte_total{strimzi_io_kind=~\"KafkaConnect.*\",strimzi_io_cluster=\"$strimzi_connect_cluster_name\"}[1m])) by (pod)",
           "interval": "",
-          "legendFormat": "{{kubernetes_pod_name}}",
+          "legendFormat": "{{pod}}",
           "refId": "A"
         }
       ],
@@ -346,9 +346,9 @@
       "steppedLine": false,
       "targets": [
         {
-          "expr": "sum(rate(kafka_producer_outgoing_byte_total{strimzi_io_kind=~\"KafkaConnect.*\",strimzi_io_cluster=\"$strimzi_connect_cluster_name\"}[1m])) by (kubernetes_pod_name)",
+          "expr": "sum(rate(kafka_producer_outgoing_byte_total{strimzi_io_kind=~\"KafkaConnect.*\",strimzi_io_cluster=\"$strimzi_connect_cluster_name\"}[1m])) by (pod)",
           "interval": "",
-          "legendFormat": "{{kubernetes_pod_name}}",
+          "legendFormat": "{{pod}}",
           "refId": "A"
         }
       ],
@@ -518,11 +518,11 @@
       "steppedLine": false,
       "targets": [
         {
-          "expr": "sum without(area)(jvm_memory_bytes_used{strimzi_io_kind=~\"KafkaConnect.*\",strimzi_io_cluster=\"$strimzi_connect_cluster_name\"})",
+          "expr": "sum (jvm_memory_bytes_used{strimzi_io_kind=~\"KafkaConnect.*\",strimzi_io_cluster=\"$strimzi_connect_cluster_name\"}) by (pod)",
           "format": "time_series",
           "hide": false,
           "intervalFactor": 2,
-          "legendFormat": "{{kubernetes_pod_name}}",
+          "legendFormat": "{{pod}}",
           "metric": "",
           "refId": "A",
           "step": 4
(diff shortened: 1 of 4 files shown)


packaging/examples/metrics/grafana-dashboards/strimzi-kafka.json
```diff
@@ -1115,10 +1115,10 @@
       "steppedLine": false,
       "targets": [
         {
-          "expr": "sum(jvm_memory_bytes_used{namespace=\"$kubernetes_namespace\",kubernetes_pod_name=~\"$strimzi_cluster_name-$kafka_broker\",strimzi_io_name=\"$strimzi_cluster_name-kafka\"}) by (kubernetes_pod_name)",
+          "expr": "sum(jvm_memory_bytes_used{namespace=\"$kubernetes_namespace\",kubernetes_pod_name=~\"$strimzi_cluster_name-$kafka_broker\",strimzi_io_name=\"$strimzi_cluster_name-kafka\"}) by (pod)",
           "format": "time_series",
           "intervalFactor": 1,
-          "legendFormat": "{{kubernetes_pod_name}}",
+          "legendFormat": "{{pod}}",
```
ppatierno, line 1121: why this change again, even considering you are filtering on `kubernetes_pod_name` attribute in the query?

packaging/examples/metrics/grafana-dashboards/strimzi-kafka-connect.json
```diff
@@ -263,9 +263,9 @@
       "steppedLine": false,
       "targets": [
         {
-          "expr": "sum(rate(kafka_consumer_incoming_byte_total{strimzi_io_kind=~\"KafkaConnect.*\",strimzi_io_cluster=\"$strimzi_connect_cluster_name\"}[1m])) by (kubernetes_pod_name)",
+          "expr": "sum(rate(kafka_consumer_incoming_byte_total{strimzi_io_kind=~\"KafkaConnect.*\",strimzi_io_cluster=\"$strimzi_connect_cluster_name\"}[1m])) by (pod)",
           "interval": "",
-          "legendFormat": "{{kubernetes_pod_name}}",
+          "legendFormat": "{{pod}}",
```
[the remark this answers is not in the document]
Frawless: From my POV yes, at least I didn't see any issues when I applied these changes to my grafana dashboards. I think that if we will keep `kubernetes_pod_name`, the colours will be different in this panel, or at least it was during my testing.

packaging/examples/metrics/grafana-dashboards/strimzi-kafka.json
Frawless: `kubernetes_pod_name` was not available in all queries for KafkaConnect AFAIR so that's the reason why I choose `pod`

packaging/examples/metrics/grafana-dashboards/strimzi-kafka-connect.json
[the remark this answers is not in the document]
ppatierno: > I think that if we will keep kubernetes_pod_name, the colours will be different in this panel, or at least it was during my testing.

The question would be ... why? What's the difference between these two attributes? Are you trying on OpenShift or Kubernetes? I think `kubernetes_pod_name` is coming from Prometheus but if on OpenShift maybe you are relaying on its own Prometheus instance that could add `pod` attribute as well? I am just guessing trying to understand the difference between the two.

packaging/examples/metrics/grafana-dashboards/strimzi-kafka.json
ppatierno: again the question here is ... why? More investigation needed from my perspective.

packaging/examples/metrics/grafana-dashboards/strimzi-kafka-connect.json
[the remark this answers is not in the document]
Frawless: I am trying it on OpenShift. Monitoring stack available with OpenShift installation does not inject `kubernetes_pod_name` for all available metrics we show in the dashboards. You can check for example `container_cpu_usage_seconds_total` metric which is shown in KC or MM2 dashboards. This query does not return `kubernetes_pod_name`, guess this was also the reason for changes in https://github.com/strimzi/strimzi-kafka-operator/pull/4149/ from `kubernetes_pod_name` to `pod`. I will try to double-check it on Kubernetes cluster, but I assume it will work as it works on OpenShift.

packaging/examples/metrics/grafana-dashboards/strimzi-kafka.json
scholzj: Well, I expect that might be injected by our Prometheus configs. But the question is how it works on our example setup and not on OpenShift. Same as I said on #7243, I think we need to explain why is `pod` the right one and why does it not impact the other users. I think we need more then just _it works for me_.

When I tried the Kafka dashboard on my cluster, it seems to work fine. That said, ther eseem to be some inconsistences in the changes. For example this query:

```
sum(rate(jvm_gc_collection_seconds_sum{namespace=\"$kubernetes_namespace\",kubernetes_pod_name=~\"$strimzi_cluster_name-$kafka_broker\",strimzi_io_name=\"$strimzi_cluster_name-kafka\"}[5m])) by (pod)
```

=> This PR changes the grouping to be based on `pod` instead of `kubernetes_pod_name`. Yet, it seems to use `kubernetes_pod_name` in the `sum` command it self. So right now, I guess to make it work we actually need both `kubernetes_pod_name` and `pod` labels to be present. That seems weird. 

So I guess we need to figure out:
* What adds the `pod` and `kubernetes_pod_name` labels and where can they be expected
* Pick the best one and use it consistently

Frawless: It seems that we inject `kubernetes_pod_name` for each metric which is scraped from the strimzi related pods - https://github.com/strimzi/strimzi-kafka-operator/blob/main/examples/metrics/prometheus-install/strimzi-pod-monitor.yaml#L81 . So I guess `pod` is added by kubernetes/prometheus configuration. I will try to find some details.

Frawless: That actually makes sense, because `container_cpu_usage_seconds_total` looks to me like a kubernetes metric so that's the reason why it doesn't have `kubernetes_pod_name` in

Frawless:
it seems that `container_memory_usage_bytes` shows a little bit strange numbers and I am not sure if we want to show them. For example, I set `limits` and `requests` to 5Gi. Kafka brokers memory usage was around that value but grafana shows 14Gi +/- (current memory usage + disk swap or something?). On the other hand, `container_memory_working_set_bytes` shows the same value as kubernetes/openshift shows in memory consumption per pod. @tombentley ideas?

im-konge (approved):
LGTM

see-quick (approved):

scholzj (approved):

ppatierno (approved):

PR #7313, opened 2022-09-12
KafkaConnector Ready status reflects connector and task state

katheris:
### Type of change

_Select the type of your PR_

- Enhancement / new feature

### Description

The KafkaConnector status "Ready" condition should not be there if one of the connector or the tasks are failed.

### Checklist

- [x] Write tests
- [ ] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/AbstractConnectOperator.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/AbstractConnectOperator.java
@@ -745,7 +745,7 @@ public abstract class AbstractConnectOperator<C extends KubernetesClient, T exte
             List<String> failedTaskIds;
             if (connectorHasFailed(statusResultJson)) {
                 connectorReadiness = Future.failedFuture(new Throwable("Connector has failed, see connectorStatus for more details."));
-            } else if ((failedTaskIds = failedTasks(statusResultJson)).size() > 0) {
+            } else if ((failedTaskIds = failedTaskIds(statusResultJson)).size() > 0) {
                 connectorReadiness = Future.failedFuture(new Throwable(String.format("The following tasks have failed: %s, see connectorStatus for more details.", String.join(", ", failedTaskIds))));
             }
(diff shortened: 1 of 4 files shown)


scholzj (commented):
So, was even the connector state ignored before? And not just the task states?

The code looks good to me. But I think this is a big change in the semantics of the Readiness state. So it would deserve CHANGELOG.md entry => but a regular entry and also adding the `Changes, deprecations and removals` section and listing it there as well.

katheris:
> So, was even the connector state ignored before? And not just the task states?

Yes, although in reality most connectors don't have any reason to fail. If you set invalid config then the REST call to create the connector fails so that would fail the reconcile loop and give a NotReady condition, otherwise if the config is valid most connectors just have code in them to set the task config so there is little code that would actually result in an exception.

Good point about the `CHANGELOG.md` I'll make that update

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/AbstractConnectOperator.java
```diff
@@ -762,6 +772,22 @@ Future<Void> maybeUpdateConnectorStatus(Reconciliation reconciliation, KafkaConn
             });
     }
 
+    private boolean connectorHasFailed(JsonObject statusResult) {
+        JsonObject connectorStatus = statusResult.getJsonObject("connector");
+        return connectorStatus != null && "FAILED".equals(connectorStatus.getString("state"));
+    }
+
+    private List<String> failedTasks(JsonObject statusResult) {
```
tombentley, line 780: Might `failedTaskIds` be a better name?

tombentley (approved):
Thanks, this LGTM. 

I'm surprised that we didn't implement this originally, but I don't recall any reasons why we consciously decided not to.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/AbstractConnectOperator.java
katheris: Yes, I've updated it

PR #7292, opened 2022-09-08
Fix issue with running KafkaConnect build with a sidecar

im-konge:
Signed-off-by: Lukas Kral <lukywill16@gmail.com>

### Type of change

- Bugfix

### Description

How mentioned in #4604 , when running KafkaConnect build with sidecar, the build is basically never completed, as we are checking always the first container status (which could be status of the sidecar).

This PR fixes this behavior - adding possibility to run KafkaConnect build with the sidecar.

Fixes #4604 

### Checklist

- [ ] Make sure all tests pass
- [ ] Try your changes from Pod inside your Kubernetes and OpenShift cluster, not just locally

PR #7319, opened 2022-09-13
Use correct container name for KafkaConnect build on Openshift

im-konge:

### Type of change

- Bugfix

### Description

In #7292 the checks for right container for KafkaConnect build were added, but it's only functional on K8s. OCP uses different name of the build container than Kaniko, which ends in `NotReady` state for every KafkaConnect build on OCP.

### Checklist

### DIFF
--- a/api/src/main/java/io/strimzi/api/kafka/model/KafkaConnectResources.java
+++ b/api/src/main/java/io/strimzi/api/kafka/model/KafkaConnectResources.java
@@ -153,16 +153,4 @@ public class KafkaConnectResources {
     public static String buildName(String clusterName, Long buildVersion) {
         return buildConfigName(clusterName) + "-" + buildVersion;
     }
-
-    /**
-     * Returns the name of the Kafka Connect {@code Build} container.
-     *
-     * @param clusterName  The {@code metadata.name} of the {@code KafkaConnect} resource.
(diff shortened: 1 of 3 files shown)


api/src/main/java/io/strimzi/api/kafka/model/KafkaConnectResources.java
```diff
@@ -153,4 +153,16 @@ public static String buildConfigName(String clusterName) {
...
+     * Returns the name of the Kafka Connect {@code Build} container.
+     *
+     * @param clusterName  The {@code metadata.name} of the {@code KafkaConnect} resource.
+     * @param isOpenshift Flag for determining, if we are running on Openshift
+     *
+     * @return The name of the corresponding Kafka Connect {@code Build} Container.
+     */
+    public static String buildContainerName(String clusterName, boolean isOpenshift) {
+        return isOpenshift ? "docker-build" : buildPodName(clusterName);
+    }
```
scholzj, line 167: I think you should keep this internally in the Utils class somewhere and not add it to the public API.

im-konge: Good point. Done

scholzj (approved):

ppatierno (approved):

Frawless (approved):

PR #7325, opened 2022-09-14
Uniform to use UTC Zone from Clock

ppatierno:
Trivial PR to use the UTC ZoneId coming directly from using Clock instead of parsing the "UTC" string.


scholzj (approved):

devguyio (approved):
Nice trick, it also bothered me to hard code the "UTC" string.

PR #7329, opened 2022-09-14
docs(storage): Adds some context to the data storage options for Kafka

PaulRMellor:

**Documentation**
Updates [Kafka and ZooKeeper storage types](https://strimzi.io/docs/operators/in-development/configuring.html#assembly-storage-str) to provide some more context to the storage types that can be used by Kafka. 

Makes the overviews for ephemeral, persistent and JBOD storage consistent.

### Checklist

- [ ] Write tests
- [x] Update documentation
- [ ] Update CHANGELOG.md


documentation/assemblies/configuring/assembly-storage.adoc
```diff
@@ -3,16 +3,15 @@
...
+As stateful applications, Kafka and ZooKeeper store data on disk. 
+Strimzi supports three storage types for this data:
 
-* Ephemeral
+* Ephemeral (Recommended for development only)
 * Persistent
-* JBOD storage
-
-NOTE: JBOD storage is supported only for Kafka, not for ZooKeeper.
+* JBOD (Suitable for *Kafka only* not ZooKeeper)
```
scholzj, line 14: It is not about suitability. JBOD is supported by Kafka only.

documentation/modules/configuring/con-considerations-for-data-storage.adoc
```diff
@@ -3,18 +3,18 @@
...
+For Strimzi to work well, an efficient data storage infrastructure is essential.
 
 Block storage is required.
 File storage, such as NFS, does not work with Kafka.
 
 Choose from one of the following options for your block storage:
 
 * Cloud-based block storage solutions, such as {aws-ebs}
-* {K8sLocalPersistentVolumes}
+* A mounted {K8sLocalPersistentVolumes}
```
scholzj, line 17: I would not add this. the way I would read it is that the user has to go and mount it. That is not really the case. So I think this is a bit confusing.

documentation/modules/configuring/ref-storage-ephemeral.adoc
```diff
@@ -5,14 +5,26 @@
 [id='ref-ephemeral-storage-{context}']
 = Ephemeral storage
 
+[role="_abstract"]
+Ephemeral data storage is transient. 
+All pods on a node share a local ephemeral storage space. 
+Data is retained for as long as the pod that uses it is running.
+The data is lost when a pod restarts.
```
scholzj, line 12:
```suggestion
The data is lost when a pod is deleted. 
```

documentation/modules/configuring/ref-storage-ephemeral.adoc
```diff
@@ -5,14 +5,26 @@
...
+
 Ephemeral storage uses `{K8sEmptyDir}` volumes to store data.
-To use ephemeral storage, set the `type` field to `ephemeral`.
+An `emptyDir` volume is created when a pod is assigned to a node. 
+You can set the total amount of storage for the `emptyDir` using the `sizeLimit` property .
 
-IMPORTANT: `emptyDir` volumes are not persistent and the data stored in them is lost when the pod is restarted.
+IMPORTANT: `emptyDir` volumes are not persistent and the data stored in them is lost when a pod is restarted.
 After the new pod is started, it must recover all data from the other nodes of the cluster.
 Ephemeral storage is not suitable for use with single-node ZooKeeper clusters or for Kafka topics with a replication factor of 1. This configuration will cause data loss.
```
scholzj, line 23: It seems like we basically repeat the content here within a 15 lines of text. The `IMPORTANT` block basically says the same as bit chunk of the first paragraph. MAybe we can simplify it and keep only the `IMPORTANT` part or something like that?

documentation/modules/configuring/ref-storage-ephemeral.adoc
```diff
@@ -5,14 +5,26 @@
...
 = Ephemeral storage
 
+[role="_abstract"]
+Ephemeral data storage is transient. 
+All pods on a node share a local ephemeral storage space. 
+Data is retained for as long as the pod that uses it is running.
+The data is lost when a pod restarts. 
+Although a pod can recover data in a highly available environment.  
+
+Because of its transient nature, ephemeral storage is not recommended for production.
```
scholzj, line 15: Should we instead of `not recommended for production` say something like `should be used only for development and/or CIs`

documentation/modules/configuring/ref-storage-ephemeral.adoc
```diff
@@ -32,9 +44,9 @@ spec:
     # ...
 ----
 
-== Log directories
+== Ephemeral storage for log directories
```
scholzj, line 47: I think this title is a bit confusing. I would definitely not expect the content under it based on the name. Maybe something like this would be better?

```suggestion
== Mount path of Kafka log directories
```

documentation/modules/configuring/ref-storage-jbod.adoc
```diff
@@ -3,17 +3,21 @@
...
+= JBOD storage
 
 [role="_abstract"]
-You can configure Strimzi to use JBOD, a data storage configuration of multiple disks or volumes. JBOD is one approach to providing increased data storage for Kafka brokers. It can also improve performance.
+You can configure Strimzi to use JBOD, a data storage configuration of multiple disks or volumes. 
+JBOD is one approach to providing increased data storage for Kafka brokers. 
+It can also improve performance.
 
-A JBOD configuration is described by one or more volumes, each of which can be either xref:ref-ephemeral-storage-{context}[ephemeral] or xref:ref-persistent-storage-{context}[persistent]. The rules and constraints for JBOD volume declarations are the same as those for ephemeral and persistent storage. For example, you cannot decrease the size of a persistent storage volume after it has been provisioned, or you cannot change the value of `sizeLimit` when type=ephemeral.
+NOTE: JBOD storage is suitable for *Kafka only* not ZooKeeper.
```
scholzj, line 13: Maybe supported or available is better then _suitable_.

```suggestion
NOTE: JBOD storage is supported for *Kafka only* not ZooKeeper.
```

documentation/modules/configuring/ref-storage-jbod.adoc
```diff
@@ -31,22 +35,26 @@ storage:
...
-When persistent storage is used to declare JBOD volumes, the naming scheme of the resulting Persistent Volume Claims is as follows:
+When persistent storage is used to declare JBOD volumes, it creates a PVC with the following name:
 
 `data-_id_-_cluster-name_-kafka-_idx_`::
 
-Where `_id_` is the ID of the volume used for storing data for Kafka broker pod `_idx_`.
+PVC for the volume used for storing data for the Kafka broker pod `_idx_`.
+The `_id_` is the ID of the volume used for storing data for Kafka broker pod.
+
+== JBOD storage for log directories
```
scholzj, line 51: Again, maybe this should say something about mount path or something like that?

documentation/modules/configuring/ref-storage-persistent.adoc
```diff
@@ -5,36 +5,48 @@
 [id='ref-persistent-storage-{context}']
 = Persistent storage
 
-Persistent storage uses {K8sPersistentVolumeClaims} to provision persistent volumes for storing data.
-Persistent Volume Claims can be used to provision volumes of many different types, depending on the {K8SStorageClass} which will provision the volume.
-The data types which can be used with persistent volume claims include many types of SAN storage as well as {K8sLocalPersistentVolumes}.
+[role="_abstract"]
+Persistent data storage retains data even when a pod that uses the storage restarts.
```
scholzj, line 9: It should not keep them just between restarts but make them persisted in general, survive all kind of failures and disasters etc.. So maybe we should make this statement more wide than just pod restarts?

documentation/modules/configuring/ref-storage-persistent.adoc
```diff
@@ -5,36 +5,48 @@
...
+Persistent data storage uses a storage framework called persistent volumes (PV).
```
scholzj, line 10: MAybe we should call it _Dynamic Volume Provisioning_? https://kubernetes.io/docs/concepts/storage/dynamic-provisioning/

documentation/modules/configuring/ref-storage-persistent.adoc
```diff
@@ -5,36 +5,48 @@
...
+Persistent data storage uses a storage framework called persistent volumes (PV). 
+PVs are storage resources, independent of the pods that use them.
+Pod configuration uses {K8sPersistentVolumeClaims} (PVCs) to make storage requests on the PVs.
+The PVC requests the amount of storage required when a pod is being created.
+The underlying storage infrastructure of the PV does not need to be understood. 
+If a PV matches the storage criteria, the PVC is bound to the PV.
 
-To use persistent storage, the `type` has to be set to `persistent-claim`.
-Persistent storage supports additional configuration options:
+Because of its permanent nature, persistent storage is recommended for production.
```
scholzj, line 17: To avoid confusion here ... 
* `type: persistent-claim` is recommended for production for ZooKeeper
* `type: jbod` with one or more `type: persistent-claim` volumes is recommended for production for Kafka

documentation/modules/configuring/ref-storage-persistent.adoc
```diff
@@ -5,36 +5,48 @@
...
-Storage identification number. This option is mandatory for storage volumes defined in a JBOD storage declaration.
+A storage identification number. This option is mandatory for storage volumes defined in a JBOD storage declaration.
 Default is `0`.
 
 `size` (required)::
-Defines the size of the persistent volume claim, for example, "1000Gi".
+The size of the persistent volume claim, for example, "1000Gi".
 
 `class` (optional)::
 The Kubernetes {K8SStorageClass} to use for dynamic volume provisioning.
```
scholzj, line 35: Maybe we could add that additional storage _configuration_ can be usually done as part of the Storage class configuration. For example things such as encryption, iops, file system type (etx4 versus xfs) etc.

documentation/modules/configuring/ref-storage-persistent.adoc
```diff
@@ -126,28 +139,29 @@ spec:
...
 `data-_cluster-name_-kafka-_idx_`::
-Persistent Volume Claim for the volume used for storing data for the Kafka broker pod `_idx_`.
+PVC for the volume used for storing data for the Kafka broker pod `_idx_`.
 
 `data-_cluster-name_-zookeeper-_idx_`::
-Persistent Volume Claim for the volume used for storing data for the ZooKeeper node pod `_idx_`.
+PVC for the volume used for storing data for the ZooKeeper node pod `_idx_`.
 
-== Log directories
+== Log directories for persistent storage
```
scholzj, line 164: Again, maybe we can make it more clear we talk about mount path here?

documentation/modules/configuring/ref-storage-ephemeral.adoc
PaulRMellor: I just retained the specific configuration info in the IMPORTANT note. I think the general stuff is better in the intro.

documentation/modules/configuring/ref-storage-persistent.adoc
PaulRMellor: changed to:

> Persistent data storage retains data in the event of system disruption.
> For pods that use persistent data storage, data is persisted across failures and restarts.

documentation/modules/configuring/ref-storage-persistent.adoc
PaulRMellor: I think we can use that term
I've rewritten and reordered these lines a little

> A dynamic provisioning framework enables clusters to be created with persistent storage. 
> Pod configuration uses {K8sPersistentVolumeClaims} (PVCs) to make storage requests on persistent volumes (PVs).
> PVs are storage resources that represent a storage volume.
> PVs are independent of the pods that use them.

PaulRMellor:
This commit updates the persistent config examples to show jbod config for Kafka
https://github.com/strimzi/strimzi-kafka-operator/pull/7329/commits/2def2a884bce5ed600483f4798136f1fdbbc4251

documentation/modules/configuring/ref-storage-persistent.adoc
```diff
@@ -5,48 +5,91 @@
...
+PVs are independent of the pods that use them.
+The PVC requests the amount of storage required when a pod is being created.
+The underlying storage infrastructure of the PV does not need to be understood. 
+If a PV matches the storage criteria, the PVC is bound to the PV.
+
+Because of its permanent nature, persistent storage is recommended for production.
+
+PVCs can request different types of persistent storage by specifying a {K8SStorageClass}.
+Storage classes define storage profiles and dynamically provision PVs.  
+If a storage class is not specified, a standard profile is used.
```
scholzj, line 24:
```suggestion
If a storage class is not specified, the default storage class is used.
```

scholzj (approved):
One nit. LGTM otherwise.

templarfelix (approved):

scholzj:
@ppatierno Can you please have a look at this?

ppatierno (approved):
LGTM

Issue #6610, opened 2022-03-31
Add client rack-awareness support to Strimzi Bridge pods to allow consuming from the closest replica


scholzj:
The issue #3426 added support for _client rack-awareness_ to Kafka Connect. The _client rack-awareness_ allows to configure the `client.rack` option of the Kafka consumers to have them consume from the closest replica instead of from the leader replica. In some cases, this might for example lead to lower costs while things such as latency might not matter. We should add this feature to Strimzi Bridge deployment as well. The implementation for KAfka Connect might serve as an example how to implement it.

PR #7339, opened 2022-09-16
Add client rack-awareness support to Strimzi Bridge pods

robobario:
### Type of change

- Enhancement / new feature
- Documentation

### Description

Implements #6610

If you configure a rack topologyKey for a KafkaBridge CRD then an init container will be added to the bridge pod to retrieve the value of rack.topologyKey and write it into a file in `/opt/strimzi/init`. This `/opt/strimzi/init` volume is also mounted into the bridge container so it can retrieve the rack.id at runtime and use it for consumer configuration.

There is a corresponding PR for the bridge to add the rack id to it's consumer configuration: https://github.com/strimzi/strimzi-kafka-bridge/pull/659

Why:
You could configure your Kafka Bridge to prefer to consume from a broker in the same zone, potentially saving on inter-zone transfer costs. This may cost in overall message latency as you may be consuming from a follower for a topic-partition.

`/opt/strimzi/init` was selected as the mount because strimzi-bridge startup scripts use a STRIMZI_HOME instead of KAFKA_HOME, this keeps the configuration within /opt/strimzi instead of introducting /opt/kafka to the image.

### Checklist

- [x] Write tests
- [ ] Update documentation
- [x] Update CHANGELOG.md

### DIFF
--- a/api/src/main/java/io/strimzi/plugin/security/profiles/PodSecurityProvider.java
+++ b/api/src/main/java/io/strimzi/plugin/security/profiles/PodSecurityProvider.java
@@ -241,18 +241,6 @@ public interface PodSecurityProvider {
 
 
     /**
-     * Provides the (container) security context for the Kafka Bridge init containers. The default implementation just
-     * returns the security context configured by the user in the template section or null (no security context).
-     *
-     * @param context   Provides the context which can be used to generate the security context
-     *
-     * @return  Security context which will be set for the Kafka Bridge init containers
-     */
-    default SecurityContext kafkaBridgeInitContainerSecurityContext(ContainerSecurityProviderContext context) {
-        return securityContextOrNull(context);
-    }
-
-    /**
      * Provides the Pod security context for the Kafka Connect Build (Kaniko) pod. The default implementation just returns
      * the security context configured by the user in the template section or null (no Pod security context).
      *
@@ -323,6 +311,19 @@ public interface PodSecurityProvider {
     default SecurityContext bridgeContainerSecurityContext(ContainerSecurityProviderContext context) {
         return securityContextOrNull(context);
     }
+    
+    /**
+     * Provides the (container) security context for the Kafka Bridge init containers. The default implementation just
+     * returns the security context configured by the user in the template section or null (no security context).
+     *
+     * @param context   Provides the context which can be used to generate the security context
+     *
+     * @return  Security context which will be set for the Kafka Bridge init containers
+     */
+    default SecurityContext bridgeInitContainerSecurityContext(ContainerSecurityProviderContext context) {
+        return securityContextOrNull(context);
+    }
+
 
     /**
      * Internal method to handle the default Pod security context.
--- a/api/src/main/java/io/strimzi/plugin/security/profiles/PodSecurityProvider.java
+++ b/api/src/main/java/io/strimzi/plugin/security/profiles/PodSecurityProvider.java
@@ -239,6 +239,19 @@ public interface PodSecurityProvider {
         return securityContextOrNull(context);
     }
 
+
+    /**
+     * Provides the (container) security context for the Kafka Bridge init containers. The default implementation just
+     * returns the security context configured by the user in the template section or null (no security context).
+     *
+     * @param context   Provides the context which can be used to generate the security context
+     *
+     * @return  Security context which will be set for the Kafka Bridge init containers
+     */
+    default SecurityContext kafkaBridgeInitContainerSecurityContext(ContainerSecurityProviderContext context) {
+        return securityContextOrNull(context);
+    }
+
     /**
      * Provides the Pod security context for the Kafka Connect Build (Kaniko) pod. The default implementation just returns
      * the security context configured by the user in the template section or null (no Pod security context).
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/model/AbstractModel.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/model/AbstractModel.java
@@ -130,8 +130,6 @@ public abstract class AbstractModel {
     protected static final String INIT_VOLUME_NAME = "rack-volume";
     protected static final String INIT_VOLUME_MOUNT = "/opt/kafka/init";
     protected static final String ENV_VAR_KAFKA_INIT_RACK_TOPOLOGY_KEY = "RACK_TOPOLOGY_KEY";
-
-    protected static final String ENV_VAR_KAFKA_INIT_INIT_FOLDER_KEY = "INIT_FOLDER";
     protected static final String ENV_VAR_KAFKA_INIT_NODE_NAME = "NODE_NAME";
 
     public static final String ANCILLARY_CM_KEY_METRICS = "metrics-config.json";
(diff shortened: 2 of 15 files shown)


cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaBridgeCluster.java
```diff
@@ -236,7 +234,45 @@ public static KafkaBridgeCluster fromCrd(Reconciliation reconciliation, KafkaBri
 
         return kafkaBridgeCluster;
     }
-    
+
+
+    private static void fromCrdTemplate(final KafkaBridgeCluster kafkaBridgeCluster, final KafkaBridgeSpec spec) {
```
robobario, line 239: extracted method due to checkstyle complaint about the method complexity when adding more `if` statements

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaBridgeCluster.java
```diff
@@ -112,7 +120,7 @@ public class KafkaBridgeCluster extends AbstractModel {
     protected static final String ENV_VAR_KAFKA_BRIDGE_CORS_ALLOWED_METHODS = "KAFKA_BRIDGE_CORS_ALLOWED_METHODS";
 
     protected static final String CO_ENV_VAR_CUSTOM_BRIDGE_POD_LABELS = "STRIMZI_CUSTOM_KAFKA_BRIDGE_LABELS";
-
+    protected static final String INIT_VOLUME_MOUNT = "/opt/strimzi/init";
```
robobario, line 123: As mentioned in the commit message I thought it was better to mount it under the STRIMZI_HOME in the bridge container, where Connect and Mirrormaker put it in `/opt/kafka`

documentation/modules/appendix_crds.adoc
```diff
@@ -2792,42 +2792,46 @@ include::../api/io.strimzi.api.kafka.model.KafkaBridgeSpec.adoc[leveloffset=+1]
 
 [options="header"]
 |====
-|Property                 |Description
-|replicas          1.2+<.<a|The number of pods in the `Deployment`.
```
robobario, line 2796: This generated file ended up indented differently, not sure what I've done wrong

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaBridgeCluster.java
```diff
@@ -508,10 +550,101 @@ protected void setBootstrapServers(String bootstrapServers) {
...
     }
 
+
+    protected void setRack(Rack rack) {
+        this.rack = rack;
+    }
+
+    protected void setInitImage(String initImage) {
+        this.initImage = initImage;
+    }
```
scholzj, line 560: Do we need the setters here? I think we normally tend to just set the values without the use of setters when the setter is not called from outside which does not seem to be the case.

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaBridgeCluster.java
```diff
@@ -508,10 +550,101 @@ protected void setBootstrapServers(String bootstrapServers) {
...
+
     public KafkaBridgeHttpConfig getHttp() {
         return this.http;
     }
 
+    public String getInitContainerClusterRoleBindingName() {
+        return KafkaBridgeResources.initContainerClusterRoleBindingName(cluster, namespace);
+    }
```
scholzj, line 568: Can't the place which is calling this call `KafkaBridgeResources.initContainerClusterRoleBindingName` directly? We tried to avoid these _call hierarchies_ lately.

scholzj (commented):
Thanks for the PR. This looks mostly good. I left some smaller comments.

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaBridgeCluster.java
scholzj: I think that is fine. It makes sense here the way you did it 👍

documentation/modules/appendix_crds.adoc
scholzj: It indents according to the longest field. So this is OK.

scholzj:
@robobario It looks like some of your changes to the CRD were not propagated through -> that is what failed the build. This should help to fix it:
```
ERROR: Uncommitted changes in derived resources:
M	packaging/helm-charts/helm3/strimzi-kafka-operator/crds/046-Crd-kafkabridge.yaml
Run the following to add up-to-date resources:
  mvn clean verify -DskipTests -DskipITs \
    && make crd_install \
    && make helm_install \
    && git add packaging/install/ packaging/helm-charts/ documentation/modules/appendix_crds.adoc cluster-operator/src/main/resources/cluster-roles \
    && git commit -s -m 'Update derived resources'
```

api/src/main/java/io/strimzi/plugin/security/profiles/PodSecurityProvider.java
```diff
@@ -239,6 +239,19 @@ default SecurityContext kafkaConnectInitContainerSecurityContext(ContainerSecuri
...
+     * returns the security context configured by the user in the template section or null (no security context).
+     *
+     * @param context   Provides the context which can be used to generate the security context
+     *
+     * @return  Security context which will be set for the Kafka Bridge init containers
+     */
+    default SecurityContext kafkaBridgeInitContainerSecurityContext(ContainerSecurityProviderContext context) {
+        return securityContextOrNull(context);
+    }
+
```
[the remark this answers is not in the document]
robobario: Moved it down under the existing bridge methods and renamed to bridgeInitContainerSecurityContext to match existing methods.

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/AbstractModel.java
```diff
@@ -130,6 +130,8 @@ public abstract class AbstractModel {
     protected static final String INIT_VOLUME_NAME = "rack-volume";
     protected static final String INIT_VOLUME_MOUNT = "/opt/kafka/init";
     protected static final String ENV_VAR_KAFKA_INIT_RACK_TOPOLOGY_KEY = "RACK_TOPOLOGY_KEY";
+
+    protected static final String ENV_VAR_KAFKA_INIT_INIT_FOLDER_KEY = "INIT_FOLDER";
```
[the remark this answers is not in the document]
robobario: I put it up here only because the other env var keys for the init container were defined here, I have pushed it down into KafkaBridgeCluster now.

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaBridgeCluster.java
robobario: Setters removed

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaBridgeAssemblyOperator.java
```diff
@@ -131,4 +137,8 @@ Future<ReconcileResult<ServiceAccount>> kafkaBridgeServiceAccount(Reconciliation
                 KafkaBridgeResources.serviceAccountName(bridge.getCluster()),
                 bridge.generateServiceAccount());
     }
+
+    protected Future<ReconcileResult<ClusterRoleBinding>> bridgeInitClusterRoleBinding(Reconciliation reconciliation, String crbName, ClusterRoleBinding crb) {
+        return withIgnoreRbacError(reconciliation, clusterRoleBindingOperations.reconcile(reconciliation, crbName, crb), crb);
+    }
```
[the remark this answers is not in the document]
robobario: Done

scholzj (approved):
LGTM. Thanks. I will start the regression for it.

PR #7350, opened 2022-09-19
 refactor Metric assertions in TopicOperatorTests

SamBarker:
### Type of change

- Refactoring

### Description

Reviewing https://github.com/strimzi/strimzi-kafka-operator/pull/7338 I noticed a lot of repeated code about asserting values of various metrics. In each block asserting metrics there is a lot of "noise" around finding the appropriate metrics in the repository and then matching the value. 

So this change pulls out some custom assertion methods so that its easier to see what each assertion is trying to prove about the metrics. 

It might be judicious to wait for #7338 to be merged before merging this change as that ensures all the tests actually pass before refactoring them. However I based the PR on main as there is no direct dependency between them.

### Checklist

- [X] Write tests
- [ ] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/topic-operator/src/test/java/io/strimzi/operator/topic/TopicOperatorTest.java
+++ b/topic-operator/src/test/java/io/strimzi/operator/topic/TopicOperatorTest.java
@@ -170,22 +170,20 @@ public class TopicOperatorTest {
                 mockTopicStore.assertEmpty(context);
                 assertNotReadyStatus(context, new InvalidTopicException(null, throwable.getMessage()));
                 context.verify(() -> {
-                    MeterRegistry registry = metrics.meterRegistry();
-
                     assertCounterValueIsZero("reconciliations");
                     assertCounterValueIsZero("reconciliations.successful");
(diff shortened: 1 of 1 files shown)


scholzj (approved):
Thanks.

scholzj:
@tombentley @katheris Can you please have a look at this?

katheris:
The changes look good to me, but in our investigations into the failing tests under PR #7338 we noticed some problems with the metric assertions, for example tests asserting metric results where the code that drives those metrics isn't being exercised. So I think it would probably make sense to hold off on this change until we have resolved those problems. Also until that is merged this change is refactoring code that often isn't being exercised as many tests are completing early (which is what 7338 is addressing) so another reason to hold off until we can see that these changes don't break the tests for certain.

scholzj:
Ok, so, should we change it to Draft? Or close it for the time being?

tombentley (approved):
Many thanks for this @SamBarker, LGTM.

katheris (approved):
LGTM, definitely more readable now

PR #7364, opened 2022-09-21
[systemtest] Add test for LeaderElection feature

im-konge:

### Type of change

- Enhancement / new feature

### Description

This PR adds test for LeaderElection feature, which checks if the leader is correctly  change during failover - in this case wrong image and `CrashLoopBackOff`.

### Checklist

- [ ] Write tests

### DIFF
--- a/systemtest/src/test/java/io/strimzi/systemtest/operators/LeaderElectionIsolatedST.java
+++ b/systemtest/src/test/java/io/strimzi/systemtest/operators/LeaderElectionIsolatedST.java
@@ -12,24 +12,30 @@ import io.strimzi.systemtest.AbstractST;
 import io.strimzi.systemtest.Constants;
 import io.strimzi.systemtest.Environment;
 import io.strimzi.systemtest.annotations.IsolatedSuite;
+import io.strimzi.systemtest.resources.operator.BundleResource;
+import io.strimzi.systemtest.resources.operator.specific.HelmResource;
 import io.strimzi.systemtest.storage.TestStorage;
 import io.strimzi.systemtest.utils.StUtils;
 import io.strimzi.systemtest.utils.kubeUtils.objects.PodUtils;
 import io.strimzi.test.annotations.IsolatedTest;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
+import org.junit.jupiter.api.AfterEach;
 import org.junit.jupiter.api.BeforeAll;
 import org.junit.jupiter.api.Tag;
 import org.junit.jupiter.api.extension.ExtensionContext;
 
-import java.util.Arrays;
-import java.util.List;
+import java.nio.file.Files;
+import java.nio.file.Paths;
+import java.util.Collections;
 
-import static io.strimzi.systemtest.Constants.INFRA_NAMESPACE;
 import static io.strimzi.systemtest.Constants.REGRESSION;
 import static io.strimzi.systemtest.resources.ResourceManager.kubeClient;
-import static org.junit.jupiter.api.Assertions.assertNotEquals;
-import static org.junit.jupiter.api.Assertions.assertTrue;
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.containsString;
+import static org.hamcrest.Matchers.equalTo;
+import static org.hamcrest.Matchers.not;
+import static org.hamcrest.Matchers.nullValue;
 import static org.junit.jupiter.api.Assumptions.assumeTrue;
 
 
@@ -51,47 +57,20 @@ public class LeaderElectionIsolatedST extends AbstractST {
 
     private static final Logger LOGGER = LogManager.getLogger(LeaderElectionIsolatedST.class);
 
-    /*
-     These env variables are already inside packaging/../060-Deployment-strimzi-cluster-operator.yaml, but to be sure that
-     we are setting everything correctly and the Leader Election is really enabled, we are adding it to additional envs
-     */
-    final EnvVar leaderEnabledEnv = new EnvVarBuilder()
+    final EnvVar leaderDisabledEnv = new EnvVarBuilder()
         .withName("STRIMZI_LEADER_ELECTION_ENABLED")
-        .withValue("true")
+        .withValue("false")
         .build();
 
-    final EnvVar leaseNameEnv = new EnvVarBuilder()
-        .withName("STRIMZI_LEADER_ELECTION_LEASE_NAME")
-        .withValue(Constants.STRIMZI_DEPLOYMENT_NAME)
-        .build();
-
-    final EnvVar leaseNamespaceEnv = new EnvVarBuilder()
-        .withName("STRIMZI_LEADER_ELECTION_LEASE_NAMESPACE")
-        .withNewValueFrom()
-            .withNewFieldRef()
-                .withFieldPath("metadata.namespace")
-            .endFieldRef()
-        .endValueFrom()
-        .build();
-
-    final EnvVar leaderIdentityEnv = new EnvVarBuilder()
-        .withName("STRIMZI_LEADER_ELECTION_IDENTITY")
-        .withNewValueFrom()
-            .withNewFieldRef()
-                .withFieldPath("metadata.name")
-            .endFieldRef()
-        .endValueFrom()
-        .build();
-
-    final List<EnvVar> envList = Arrays.asList(leaderEnabledEnv, leaseNameEnv, leaseNamespaceEnv, leaderIdentityEnv);
+    final String leaderMessage = "I'm the new leader";
 
     @IsolatedTest
     void testLeaderElection(ExtensionContext extensionContext) {
-        TestStorage testStorage = new TestStorage(extensionContext);
+        final TestStorage testStorage = new TestStorage(extensionContext);
 
+        // create CO with 2 replicas, wait for Deployment readiness and leader election
         clusterOperator = clusterOperator.defaultInstallation()
             .withExtensionContext(extensionContext)
-            .withExtraEnvVars(envList)
             .withReplicas(2)
             .createInstallation()
             .runInstallation();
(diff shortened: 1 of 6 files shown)


systemtest/src/test/java/io/strimzi/systemtest/operators/LeaderElectionIsolatedST.java
```diff
@@ -0,0 +1,129 @@
...
+/**
+ * Suite for testing Leader Election feature <br>
+ *
+ * The feature allows users to run Cluster operator in more than one replica <br>
+ *
+ * There will be always one leader, other replicas will stay in "standby" mode <br>
+ *
+ * The whole procedure of deploying CO with Leader Election enabled and many more is described in
+ *
+ * <a href="https://strimzi.io/docs/operators/in-development/configuring.html#assembly-using-multiple-cluster-operator-replicas-str">the documentation</a>
```
see-quick, line 45: Hmmm, I guess that referring to doc as a specific URL is not great because it could change anytime... :) Wouldn't be better to reference just `https://strimzi.io/docs/operators/in-development/configuring.html`?

systemtest/src/test/java/io/strimzi/systemtest/operators/LeaderElectionIsolatedST.java
```diff
@@ -0,0 +1,129 @@
...
+
+@Tag(REGRESSION)
+@IsolatedSuite
+public class LeaderElectionIsolatedST extends AbstractST {
+
+    private static final Logger LOGGER = LogManager.getLogger(LeaderElectionIsolatedST.class);
+
+    /*
+     These env variables are already inside packaging/../060-Deployment-strimzi-cluster-operator.yaml, but to be sure that
+     we are setting everything correctly and the Leader Election is really enabled, we are adding it to additional envs
```
see-quick, line 56: > These env variables are already inside packaging/../060-Deployment-strimzi-cluster-operator.yaml, but to be sure that we are setting everything correctly and the Leader Election is really enabled, we are adding it to additional envs

well, I see your point about overriding these envs. But we relay on `packaging/../060-Deployment-strimzi-cluster-operator.yaml` file. Meaning that Leader Election is enabled by default (with 1 replica) and thus we expect that such Deployment will have these ENVs already correctly configured. By overriding, we could simply miss a real bug. Maybe we can easily do a pre-check (i.e., check if such ENVs are present in `060-Deployment-strimzi-cluster-operator.yaml`? If so we continue, otherwise throw an Exception?

systemtest/src/test/java/io/strimzi/systemtest/operators/LeaderElectionIsolatedST.java
```diff
@@ -0,0 +1,129 @@
...
+
+        String logFromNewLeader = StUtils.getLogFromPodByTime(INFRA_NAMESPACE, currentLeaderPodName, Constants.STRIMZI_DEPLOYMENT_NAME, "300s");
+
+        assertTrue(logFromNewLeader.contains("I'm the new leader"));
+        assertNotEquals(oldLeaderPodName, currentLeaderPodName);
+    }
+
+    @BeforeAll
+    void setup() {
+        assumeTrue(!Environment.isOlmInstall());
```
see-quick, line 126: Adding some comment why OLM installation is not supported for this feature would be great.

systemtest/src/test/java/io/strimzi/systemtest/operators/LeaderElectionIsolatedST.java
```diff
@@ -0,0 +1,129 @@
...
+                .withFieldPath("metadata.name")
+            .endFieldRef()
+        .endValueFrom()
+        .build();
+
+    final List<EnvVar> envList = Arrays.asList(leaderEnabledEnv, leaseNameEnv, leaseNamespaceEnv, leaderIdentityEnv);
+
+    @IsolatedTest
+    void testLeaderElection(ExtensionContext extensionContext) {
+        TestStorage testStorage = new TestStorage(extensionContext);
```
see-quick, line 90:
```suggestion
        final TestStorage testStorage = new TestStorage(extensionContext);
```

systemtest/src/test/java/io/strimzi/systemtest/operators/LeaderElectionIsolatedST.java
```diff
@@ -0,0 +1,129 @@
...
+
+        PodUtils.waitUntilPodIsInCrashLoopBackOff(testStorage.getNamespaceName(), oldLeaderPodName);
+
+        Lease currentLease = kubeClient().getClient().leases().inNamespace(testStorage.getNamespaceName()).withName(Constants.STRIMZI_DEPLOYMENT_NAME).get();
+        String currentLeaderPodName = currentLease.getSpec().getHolderIdentity();
+
+        String logFromNewLeader = StUtils.getLogFromPodByTime(INFRA_NAMESPACE, currentLeaderPodName, Constants.STRIMZI_DEPLOYMENT_NAME, "300s");
+
+        assertTrue(logFromNewLeader.contains("I'm the new leader"));
+        assertNotEquals(oldLeaderPodName, currentLeaderPodName);
```
see-quick, line 121: So are we starting to use `jupiter.api` ? :D

see-quick (approved):
I have a few comments but apart from them, it looks great. Good job! :) Just one question, what about testing a negative scenario where we configure 2 or potentially more replicas, and we set `STRIMZI_LEADER_ELECTION_ENABLED` to false? I assume that CO will crash will some kind of error message.

systemtest/src/test/java/io/strimzi/systemtest/operators/LeaderElectionIsolatedST.java
im-konge: Makes sense

scholzj: I would keep it. In the worst case, if the anchor is removed or renamed, you will still get to https://strimzi.io/docs/operators/in-development/configuring.html

systemtest/src/test/java/io/strimzi/systemtest/operators/LeaderElectionIsolatedST.java
scholzj: I agree with @see-quick. There is no plan to disable it in the 060-... file. And by using what it has, you don't test only the code but also the installation files which adds more value.

systemtest/src/test/java/io/strimzi/systemtest/operators/LeaderElectionIsolatedST.java
```diff
@@ -0,0 +1,129 @@
...
+ *
+ * There will be always one leader, other replicas will stay in "standby" mode <br>
+ *
+ * The whole procedure of deploying CO with Leader Election enabled and many more is described in
+ *
+ * <a href="https://strimzi.io/docs/operators/in-development/configuring.html#assembly-using-multiple-cluster-operator-replicas-str">the documentation</a>
+ */
+
+@Tag(REGRESSION)
+@IsolatedSuite
```
scholzj, line 49: Do we need to disable it for OLM tests? I don't think OLM lets you easily override the number of replicas or?

systemtest/src/test/java/io/strimzi/systemtest/operators/LeaderElectionIsolatedST.java
```diff
@@ -0,0 +1,129 @@
...
+
+        clusterOperator = clusterOperator.defaultInstallation()
+            .withExtensionContext(extensionContext)
+            .withExtraEnvVars(envList)
+            .withReplicas(2)
+            .createInstallation()
+            .runInstallation();
+
+        Lease oldLease = kubeClient().getClient().leases().inNamespace(testStorage.getNamespaceName()).withName(Constants.STRIMZI_DEPLOYMENT_NAME).get();
+        String oldLeaderPodName = oldLease.getSpec().getHolderIdentity();
```
scholzj, line 100: How do you know the operator is ready and has a leader elected at this point?

systemtest/src/test/java/io/strimzi/systemtest/operators/LeaderElectionIsolatedST.java
```diff
@@ -0,0 +1,129 @@
...
+        LOGGER.info("Changing image of the leader pod: {} to not available image - to cause CrashLoopBackOff and change of leader to second pod (failover)", oldLeaderPodName);
+
+        kubeClient().editPod(testStorage.getNamespaceName(), oldLeaderPodName).edit(pod -> new PodBuilder(pod)
+            .editOrNewSpec()
+                .editContainer(0)
+                    .withImage("wrong-image/name:latest")
+                .endContainer()
+            .endSpec()
+            .build()
+        );
```
scholzj, line 111: So, you have 2 replicas. And when you change the image, you expect that one will be deleted and the new pod will not start because of faulty image. But how do you know that it will be the leader who is deleted and fails to start and not the follower?

systemtest/src/test/java/io/strimzi/systemtest/operators/LeaderElectionIsolatedST.java
im-konge: I'm disable it in the `@BeforeAll`, just will add there comment why it is disabled - I tried to pass the replicas there, but it `config` of the subscription lets you configure just few things.

systemtest/src/test/java/io/strimzi/systemtest/operators/LeaderElectionIsolatedST.java
im-konge: We are waiting for the deployment readiness there. So I assume that when the deployment and pods are up and running, the leader is also elected.

systemtest/src/test/java/io/strimzi/systemtest/operators/LeaderElectionIsolatedST.java
im-konge: I have the `oldLeaderPodName` variable, which holds the name of the pod which was leader before. So I'm editing pod which _is_ leader actually. That way we'll know that it will move to the second pod.

systemtest/src/test/java/io/strimzi/systemtest/operators/LeaderElectionIsolatedST.java
im-konge: Okay, thanks both :)

systemtest/src/test/java/io/strimzi/systemtest/operators/LeaderElectionIsolatedST.java
scholzj: What exactly is _there_?

systemtest/src/test/java/io/strimzi/systemtest/operators/LeaderElectionIsolatedST.java
scholzj: Ahh, ok. So ... you edit the Pod directly under the Deployments hands? I wonder how stable that is, but I guess as long as it works 🤷‍♂️

systemtest/src/test/java/io/strimzi/systemtest/operators/LeaderElectionIsolatedST.java
im-konge: On line 97 - `runInstallation()` - it deploys the CO and wait for readiness of `Deployment` and its pods

systemtest/src/test/java/io/strimzi/systemtest/operators/LeaderElectionIsolatedST.java
im-konge: I let it be like 10 minutes and it was in the crash, I thought K8s will roll it somehow. We'll see how it will work on Azure

systemtest/src/test/java/io/strimzi/systemtest/operators/LeaderElectionIsolatedST.java
scholzj: Ok, fine.

But I think you guys really need to learn to name the test methods with what they actually do. The code would be much more readable for everyone if it was called for example `installAndWaitForReadiness`. This way, this is a hidden knowledge and makes it really hard to read and review the code because nobody without the knowledge knows whether this is a bug, an issue or what exactly happens there. Not saying you need to rename it in this PR, but this is not the first time I see it in the STs.

im-konge: I agree. For this I created - https://github.com/strimzi/strimzi-kafka-operator/issues/7368 . We will look into in one of the future PRs. Thanks @scholzj

systemtest/src/test/java/io/strimzi/systemtest/operators/LeaderElectionIsolatedST.java
Frawless: I think we agreed to use hamcrest for asserts some time ago. Guess we should continue with it and possibly refactor jupiter occurrences or agree on different way

scholzj (approved):
LGTM. Thanks.

Frawless (approved):

im-konge:
@strimzi-ci run tests test_only profile=regression install_type=helm testcase=LeaderElectionIsolatedST

im-konge:
Test failures are not connected to my PR.

PR #7365, opened 2022-09-21
clusterIP external listener to expose kafka through ingress controller using tcp config

maknihamdi:
add clusterIP listener type. Can be used to expose kafka through ingress controller using tcp config

### Type of change

- Enhancement / new feature

### Description

If you 

- don't want to manage dynamic broker adresses with nodeport services
- don't want to bring new LoadBalancer and external IP by broker (+ bootstrap)
- can't use tls in your existing kafka clients
- cant enable ssl passthrough in your ingress controller

we don't have another solution today for an external access to your kafka cluster.

The proposal is to use ingress controller TCP configuration to expose brokers on different tcp ports. TLS are optional, and no http ingress objects are managed. 

strimzi will create (like with ingress listener type) cluster ip services, and manually we should care about ingress tcp configurations.

### Checklist

- [x] Write tests
- [x] Update documentation
- [x] Update CHANGELOG.md


scholzj (commented):
Thanks for the PR. I think support for Ingress-TCP might be an interesting feature. Although I wonder whether it isn't better to focus on Gateway API. 

However, my issue with this PR is that it does not seem to really implement TCP Ingress listener. If I got it right from the code and from your description, it seems to implement some kind of _different_ internal listener which creates per-broker services but does not create any Ingress resources.

Now, I can see why this is useful and in a way it is even better and more versatile since this comes handy to also plug other non-Nginx-Ingress controllers or the Gateway API mentioned above. So it is not necessarily bad thing. But calling it ingress-tcp is simply misleading. So I think we need to find some new name for it which better describes what it does. TBH, I do not have a great idea for the name ... maybe `internal-clusterip`? Or just `clusterip`? @strimzi/maintainers Any thoughts about the name?

Apart form the name, some other things which might need changing:
* It should probably use the services as advertised hostnames by default if no override for the advertised hostname is specified + the related tests
* I think this would also need the docs to be added here: https://github.com/strimzi/strimzi-kafka-operator/blob/main/documentation/api/io.strimzi.api.kafka.model.listener.arraylistener.GenericKafkaListener.adoc to explain more detail of how this works, how to use it and when is it useful.
* A system test might be nice. Just something simple here: https://github.com/strimzi/strimzi-kafka-operator/blob/main/systemtest/src/test/java/io/strimzi/systemtest/kafka/listeners/ListenersST.java - for example similar to `testSendMessagesTlsAuthenticated` just with the new listener without any overrides and ingresses.
* Add record to CHANGELOG.md

maknihamdi:
completely agree with you, it's not really an ingress implementation. there is no Ingress objects manager with this listener. But it is a custom implementation if we want to have an ingress tcp routing. BTW, ther is another problem with this name, we cant have an all-in-the-box implementation like ingress listener, because ingress tcp config are generally shared with other services, with nginx ingress controller we should edit LoadBalancer services to define exposed ports and a special configmap to define port/services mappings.
We may create a configmap "template" to help uses with a simple copy/past to create the nginx configmap.

my first intention was to call it "custom" or generic listener (like the class used by the operator) because we only need to have a custom config and a cluster ip service by broker. I prefer your suggestion: clusterIp listener.

keep open for other suggestions!

thank you fr your review! I consider your changing requests in another commit

api/src/main/java/io/strimzi/api/kafka/model/listener/arraylistener/GenericKafkaListener.java
```diff
@@ -84,7 +84,8 @@ public void setPort(int port) {
             "* `route` type uses OpenShift Routes to expose Kafka.\n" +
             "* `loadbalancer` type uses LoadBalancer type services to expose Kafka.\n" +
             "* `nodeport` type uses NodePort type services to expose Kafka.\n" +
-            "* `ingress` type uses Kubernetes Nginx Ingress to expose Kafka.\n")
+            "* `ingress` type uses Kubernetes Nginx Ingress to expose Kafka, with passthrought.\n" +
```
PaulRMellor, line 87:
```suggestion
            "* `ingress` type uses Kubernetes Nginx Ingress to expose Kafka with passthrough.\n" +
```

PaulRMellor:
Something like `cluster-ip` sounds fine to me.
@maknihamdi - I can take another look when you've made the updates.
I'm interested in the doc changes, as I'm involved in maintaining them

We'll need something in the following places:

- https://github.com/strimzi/strimzi-kafka-operator/blob/main/documentation/api/io.strimzi.api.kafka.model.listener.arraylistener.GenericKafkaListener.adoc
- https://strimzi.io/docs/operators/in-development/configuring.html#assembly-accessing-kafka-outside-cluster-str

And possibly in other places where we talk about listeners. Maybe if we look at these two sections and then I can follow up with a PR that updates elsewhere as necessary?

scholzj:
`type: service` is another option which came to my mind. As in ...
* `type: route` creates Routes
* `type: ingress` creates Ingresses
* `type: service` creates Services 🤷

maknihamdi:
LoadBalancer/NodePort are also a Service ...
it can be a source of confusion

scholzj:
Yeah, true to that. I will try to get some more feedback on this from other maintainers to avoid renaming it five times.

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/ListenersValidator.java
```diff
@@ -186,6 +189,39 @@ private static void validateIngress(Set<String> errors, int replicas, GenericKaf
...
+                    }
+                }
+            } else {
+                errors.add("listener " + listener.getName() + " is missing a broker configuration with advertised host names which is required for ClusterIP based listeners");
+            }
+
+
+        } else {
+            errors.add("listener " + listener.getName() + " is missing a configuration with advertised host names which is required for ClusterIP based listeners");
+        }
```
scholzj, line 222: I don't think you should make the advertised host / ports required. Instead, you should update the `KafkaListenersReconciler` class and have it by defualt advertise the DNS names of the ClusterIP services / bootstrap service and add them to the TLS certificates. You can check how its done for example here: https://github.com/strimzi/strimzi-kafka-operator/blob/9978b91fa8c714df61b63b1ccd348b94871b9098/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaListenersReconciler.java#L404 or in the other methods in this class. That way, it will work out of the box and you can optionally override it.

scholzj (commented):
Ok, nobody replied regarding the name. So I guess we go with `type: cluster-ip`. I left some more comments.

maknihamdi:
thank you @scholzj for your review!
I also have the systemtest to do, and I think miss some documentations requested by @PaulRMellor 
come back with other commits with all of this

cluster-operator/src/test/java/io/strimzi/operator/cluster/model/KafkaBrokerConfigurationBuilderTest.java
```diff
@@ -1426,6 +1426,101 @@ public void testWithExternalListenersIngress()  {
                 "listener.name.external-9094.ssl.keystore.type=PKCS12"));
     }
 
+    @ParallelTest
+    public void testWithExternalListenersClusterIPWithTLS()  {
```
tombentley, line 1430: The rest of the code mostly uses strict camel case ⇒ `testWithExternalListenersClusterIpWithTls`

cluster-operator/src/test/java/io/strimzi/operator/cluster/model/KafkaBrokerConfigurationBuilderTest.java
```diff
@@ -1426,6 +1426,101 @@ public void testWithExternalListenersIngress()  {
...
+                "sasl.enabled.mechanisms=",
+                "ssl.secure.random.implementation=SHA1PRNG",
+                "ssl.endpoint.identification.algorithm=HTTPS",
+                "listener.name.external-9094.ssl.keystore.location=/tmp/kafka/cluster.keystore.p12",
+                "listener.name.external-9094.ssl.keystore.password=${CERTS_STORE_PASSWORD}",
+                "listener.name.external-9094.ssl.keystore.type=PKCS12"));
+    }
+
+    @ParallelTest
+    public void testWithExternalListenersClusterIPWithoutTLS()  {
```
tombentley, line 1479: `testWithExternalListenersClusterIpWithoutTls()`

documentation/api/io.strimzi.api.kafka.model.listener.arraylistener.GenericKafkaListener.adoc
```diff
@@ -213,9 +213,109 @@ spec:
           preferredNodePortAddressType: InternalDNS
     #...
 ----
-+
 NOTE: TLS hostname verification is not currently supported when exposing Kafka clusters using node ports.
 
+cluster-ip:: Configures an external listener to expose Kafka `cluserIP` type `Services`.
```
tombentley, line 218:
```suggestion
cluster-ip:: Configures an external listener to expose Kafka `clusterIP` type `Services`.
```

documentation/api/io.strimzi.api.kafka.model.listener.arraylistener.GenericKafkaListener.adoc
```diff
@@ -213,9 +213,109 @@ spec:
...
++
+This listener type can be helpful if you:
+
+* don't want to manage dynamic broker adresses with nodeport services
+* don't want to bring new LoadBalancer and external IP by broker (+ bootstrap)
+* can't use tls in your existing kafka clients
+* can't enable ssl passthrough in your ingress controller
```
tombentley, line 225:
```suggestion
* can't enable SSL passthrough in your ingress controller
```

documentation/api/io.strimzi.api.kafka.model.listener.arraylistener.GenericKafkaListener.adoc
```diff
@@ -213,9 +213,109 @@ spec:
...
-+
 NOTE: TLS hostname verification is not currently supported when exposing Kafka clusters using node ports.
 
+cluster-ip:: Configures an external listener to expose Kafka `cluserIP` type `Services`.
++
+This listener type can be helpful if you:
+
+* don't want to manage dynamic broker adresses with nodeport services
+* don't want to bring new LoadBalancer and external IP by broker (+ bootstrap)
+* can't use tls in your existing kafka clients
```
tombentley, line 224:
```suggestion
* can't use TLS in your existing kafka clients
```

documentation/api/io.strimzi.api.kafka.model.listener.arraylistener.GenericKafkaListener.adoc
```diff
@@ -213,9 +213,109 @@ spec:
...
     #...
 ----
-+
 NOTE: TLS hostname verification is not currently supported when exposing Kafka clusters using node ports.
 
+cluster-ip:: Configures an external listener to expose Kafka `cluserIP` type `Services`.
++
+This listener type can be helpful if you:
+
+* don't want to manage dynamic broker adresses with nodeport services
```
tombentley, line 222:
```suggestion
* don't want to manage dynamic broker adresses with NodePort services
```

documentation/api/io.strimzi.api.kafka.model.listener.arraylistener.GenericKafkaListener.adoc
```diff
@@ -213,9 +213,109 @@ spec:
...
+Using only one LoadBalancer, brokers could be exposed using ingress controller.
++
+A new clusterIP service is created for each Kafka broker pod.
++
+An additional clusterIP is created to serve as a Kafka _bootstrap_ address.
++
+The client connects on the `advertisedPort`, 9190 for broker 0, on `advertisedHost`, ingress.mycompany.com in the following example, but traffic is then routed to the port you configure,
+which is `9096` in the following example.
++
+`advertisedHost` ingress.mycompany.com should be a dns name routed to ingress LoadBalancer service. In the Ingress Controller LoadBalancer service,
```
tombentley, line 238:
```suggestion
`advertisedHost` ingress.mycompany.com should be a DNS name routed to an ingress-based LoadBalancer service. In the Ingress Controller LoadBalancer service,
```

documentation/api/io.strimzi.api.kafka.model.listener.arraylistener.GenericKafkaListener.adoc
```diff
@@ -213,9 +213,109 @@ spec:
...
+          brokers:
+            - broker: 0
+              advertisedHost: ingress.mycompany.com
+              advertisedPort: 9190
+            - broker: 1
+              advertisedHost: ingress.mycompany.com
+              advertisedPort: 9191
+            - broker: 2
+              advertisedHost: ingress.mycompany.com
+              advertisedPort: 9192
```
tombentley, line 263: One concern about this API is it makes it harder to scale the cluster up, since you have to update this configuration with new brokers prior to the actual scale up. But then you also have to update the Service and the ConfigMap, so perhaps it's inevitable that this can't really work easily with scale up.

documentation/api/io.strimzi.api.kafka.model.listener.arraylistener.GenericKafkaListener.adoc
```diff
@@ -213,9 +213,109 @@ spec:
...
+    #...
+
+----
+In your ingress controller configuration you should set properly port->service mapping.
+
+Each broker (clusterIP service) should have different port number to give to ingress controller a selector to choose `Service` to forward to.
```
tombentley, line 269:
```suggestion
Each broker (clusterIP service) must have different port number to give to ingress controller a selector to choose `Service` to forward to.
```

tombentley (approved):
Thanks for the PR. `cluster-ip` seems like an OK name to me. 

I left a few trivial comments, but I guess it LGTM.

cluster-operator/src/test/java/io/strimzi/operator/cluster/model/KafkaBrokerConfigurationBuilderTest.java
maknihamdi: cause of my golang developper practice and golint rules. terms like Id, Ip, Http are always used as all capital letter form (ID, IP, HTTP) in a composite function/type/var name
I can change it if it does not respect the project rules.

documentation/api/io.strimzi.api.kafka.model.listener.arraylistener.GenericKafkaListener.adoc
maknihamdi: sure, since the ingress exposition are a manual operation, it will be hard to have a scalable solution. We still need manuel and hard configuration to make it working

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/ListenersValidator.java
maknihamdi: do you think is good idea to add ClusterIp listener type to ListenersUtils.internalListeners method? because when I say internalServicesReady method I dont think I need more with ClusterIP listeners. the other solution is to add a similar method (clusterIPServicesReady?) with the same content as internalServicesReady in the beginning

scholzj: I think you can do a new method for this listener type. That is what I would probably do. But the `internalServicesReady` and the other `*Ready` methods should give you the idea what would be in your new method.

see-quick:
> thank you @scholzj for your review! I also have the systemtest to do, and I think miss some documentations requested by @PaulRMellor come back with other commits with all of this

If you have any trouble creating an ST you can ask me :) but as Jakub already mentioned, you can do very similar to this one https://github.com/strimzi/strimzi-kafka-operator/blob/main/systemtest/src/test/java/io/strimzi/systemtest/kafka/listeners/ListenersST.java#L158-L210.

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/ListenersValidator.java
```diff
@@ -360,12 +360,12 @@ private static void validateBootstrapLabelsAndAnnotations(Set<String> errors, Ge
                 && !KafkaListenerType.INGRESS.equals(listener.getType())) {
             if (listener.getConfiguration().getBootstrap().getLabels() != null
                     && !listener.getConfiguration().getBootstrap().getLabels().isEmpty()) {
-                errors.add("listener " + listener.getName() + " cannot configure bootstrap.labels because it is not LoadBalancer, NodePort, Route, or Ingress based listener");
+                errors.add("listener " + listener.getName() + " cannot configure bootstrap.labels because it is not LoadBalancer, NodePort, Route or Ingress based listener");
```
scholzj, line 363: Why would these not be allowed for ClusterIP listeners? They have their own services, so I think it should be allwoed.

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/ListenersValidator.java
```diff
@@ -360,12 +360,12 @@ private static void validateBootstrapLabelsAndAnnotations(Set<String> errors, Ge
...
             }
 
             if (listener.getConfiguration().getBootstrap().getAnnotations() != null
                     && !listener.getConfiguration().getBootstrap().getAnnotations().isEmpty()) {
-                errors.add("listener " + listener.getName() + " cannot configure bootstrap.annotations because it is not LoadBalancer, NodePort, Route, or Ingress based listener");
+                errors.add("listener " + listener.getName() + " cannot configure bootstrap.annotations because it is not LoadBalancer, NodePort, Route or Ingress based listener");
```
scholzj, line 368: Same as above: Why would these not be allowed for ClusterIP listeners? They have their own services, so I think it should be allwoed.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaListenersReconciler.java
```diff
@@ -442,6 +443,63 @@ protected Future<Void> internalServicesReady()   {
...
+     *      2) Collects the clusterIP addresses for certificates and advertised hostnames
+     *
+     * @return  Future which completes clusterIP service addresses are collected
+     */
+    protected Future<Void> clusterIPServicesReady()   {
+        for (GenericKafkaListener listener : ListenersUtils.clusterIPListeners(kafka.getListeners())) {
+            boolean useServiceDnsDomain = (listener.getConfiguration() != null && listener.getConfiguration().getUseServiceDnsDomain() != null)
+                    ? listener.getConfiguration().getUseServiceDnsDomain() : false;
+
+            String bootstrapAddress = listener.getConfiguration().getBootstrap().getHost();
```
scholzj, line 459: I do not think you should take the host. `host` fields should be probably disabled for the ClusterIP listeenrs, or?

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaListenersReconciler.java
```diff
@@ -442,6 +443,63 @@ protected Future<Void> internalServicesReady()   {
...
+            result.listenerStatuses.add(ls);
+
+            // Set advertised hostnames and ports
+            for (int brokerId = 0; brokerId < kafka.getReplicas(); brokerId++) {
+                String brokerAddress;
+
+                if (useServiceDnsDomain) {
+                    brokerAddress = DnsNameGenerator.podDnsNameWithClusterDomain(reconciliation.namespace(), KafkaResources.brokersServiceName(reconciliation.name()), KafkaResources.kafkaStatefulSetName(reconciliation.name()) + "-" + brokerId);
+                } else {
+                    brokerAddress = DnsNameGenerator.podDnsNameWithoutClusterDomain(reconciliation.namespace(), KafkaResources.brokersServiceName(reconciliation.name()), KafkaResources.kafkaStatefulSetName(reconciliation.name()) + "-" + brokerId);
```
scholzj, line 486: This looks like it would not direct the traffic through the ClusterIP service but instead try to use it as headless service. I think you need to use here `getInternalServiceHostname` method as well. Just feed it the name of the per-broker service, which I guess is generated through `ListenersUtils.backwardsCompatibleBrokerServiceName`.

scholzj (commented):
Some more things I found. But I think we are getting close.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaListenersReconciler.java
maknihamdi: right! I confuse it with bootstrap url used by client to reach bootstrap service. We should ignore host field

maknihamdi:
> If you have any trouble creating an ST you can ask me :) but as Jakub already mentioned, you can do very similar to this one https://github.com/strimzi/strimzi-kafka-operator/blob/main/systemtest/src/test/java/io/strimzi/systemtest/kafka/listeners/ListenersST.java#L158-L210.

I think I got it. however I couldn't run the test with my Idea. It seam like the container running on operator pod in my minikube is not synced with source code.

im-konge:
> > If you have any trouble creating an ST you can ask me :) but as Jakub already mentioned, you can do very similar to this one https://github.com/strimzi/strimzi-kafka-operator/blob/main/systemtest/src/test/java/io/strimzi/systemtest/kafka/listeners/ListenersST.java#L158-L210.
> 
> 
> 
> I think I got it. however I couldn't run the test with my Idea. It seam like the container running on operator pod in my minikube is not synced with source code.

Hey,
You will need to build the operator image, push it to some public registries and then change it in `packaging/install/cluster-operator/060-Deployment-strimzi-cluster-operator.yaml`. After that you will be able to run it with your changes.

cluster-operator/src/test/java/io/strimzi/operator/cluster/model/KafkaClusterTest.java
```diff
@@ -3142,6 +3142,96 @@ public void testExternalIngressMissingConfiguration() {
...
+            assertThat(srv.getMetadata().getName(), is(KafkaResources.kafkaStatefulSetName(CLUSTER) + "-" + i));
+            assertThat(srv.getSpec().getType(), is("ClusterIP"));
+            assertThat(srv.getSpec().getSelector().get(Labels.KUBERNETES_STATEFULSET_POD_LABEL), is(KafkaResources.kafkaPodName(CLUSTER, i)));
+            assertThat(srv.getSpec().getPorts(), is(Collections.singletonList(kc.createServicePort(ListenersUtils.BACKWARDS_COMPATIBLE_EXTERNAL_PORT_NAME, 9094, 9094, "TCP"))));
+            checkOwnerReference(kc.createOwnerReference(), srv);
+        }
+
+    }
+    @ParallelTest
+    public void testExternalClusterIPMissingConfiguration() {
```
scholzj, line 3213: MAybe you can rename it to just `testClusterIPMissingConfiguration`?

cluster-operator/src/test/java/io/strimzi/operator/cluster/model/KafkaClusterTest.java
```diff
@@ -3142,6 +3142,96 @@ public void testExternalIngressMissingConfiguration() {
         assertThrows(InvalidResourceException.class, () -> KafkaCluster.fromCrd(Reconciliation.DUMMY_RECONCILIATION, kafkaAssembly, VERSIONS));
     }
 
+
+    @ParallelTest
+    public void testExternalClusterIP() {
```
scholzj, line 3147: MAybe you can rename it to just `testClusterIP`?

documentation/api/io.strimzi.api.kafka.model.listener.arraylistener.GenericKafkaListener.adoc
```diff
@@ -213,9 +213,35 @@ spec:
           preferredNodePortAddressType: InternalDNS
     #...
 ----
-+
 NOTE: TLS hostname verification is not currently supported when exposing Kafka clusters using node ports.
 
+cluster-ip:: Create a listener which uses per-broker `Services` instead of the headless Service.
++
+This listener type can be useful to expose the cluster outside of Kuebrnetes with some custom mechanisms such as TCP Ingress or other Ingress controllers etc
+
```
scholzj, line 221: I think this empty line should not be there to connect the paragraphs.

documentation/api/io.strimzi.api.kafka.model.listener.arraylistener.GenericKafkaListener.adoc
```diff
@@ -213,9 +213,35 @@ spec:
...
++
+A new clusterIP service is created for each Kafka broker pod.
++
+An additional clusterIP is created to serve as a Kafka _bootstrap_ address.
++
+
```
scholzj, line 227: I think this empty line should not be there to connect the paragraphs.

systemtest/src/test/java/io/strimzi/systemtest/kafka/listeners/ListenersST.java
```diff
@@ -668,6 +668,92 @@ void testLoadBalancerTls(ExtensionContext extensionContext) {
         );
     }
 
+    @ParallelNamespaceTest
+    @Tag(SANITY)
+    @Tag(ACCEPTANCE)
+    @Tag(EXTERNAL_CLIENTS_USED)
```
scholzj, line 674: I think this is not needed since we don't really use it as external listener here. We should base it on one of the internal listener tests I think. Or @im-konge

systemtest/src/test/java/io/strimzi/systemtest/kafka/listeners/ListenersST.java
```diff
@@ -668,6 +668,92 @@ void testLoadBalancerTls(ExtensionContext extensionContext) {
...
+    void testClusterIP(ExtensionContext extensionContext) {
+        final String namespaceName = StUtils.getNamespaceBasedOnRbac(clusterOperator.getDeploymentNamespace(), extensionContext);
+        final String clusterName = mapWithClusterNames.get(extensionContext.getDisplayName());
+        final String topicName = mapWithTestTopics.get(extensionContext.getDisplayName());
+
+        resourceManager.createResource(extensionContext, KafkaTemplates.kafkaEphemeral(clusterName, 3)
+                .editSpec()
+                .editKafka()
+                .withListeners(new GenericKafkaListenerBuilder()
+                        .withName(Constants.EXTERNAL_LISTENER_DEFAULT_NAME)
```
scholzj, line 684: Maybe we should have some new name as it is not external listener?

systemtest/src/test/java/io/strimzi/systemtest/kafka/listeners/ListenersST.java
```diff
@@ -668,6 +668,92 @@ void testLoadBalancerTls(ExtensionContext extensionContext) {
...
+                externalKafkaClient.sendMessagesPlain(),
+                externalKafkaClient.receiveMessagesPlain()
+        );
+
+    }
+
+    @ParallelNamespaceTest
+    @Tag(SANITY)
+    @Tag(ACCEPTANCE)
+    @Tag(EXTERNAL_CLIENTS_USED)
```
scholzj, line 714: Same as above.

systemtest/src/test/java/io/strimzi/systemtest/kafka/listeners/ListenersST.java
```diff
@@ -668,6 +668,92 @@ void testLoadBalancerTls(ExtensionContext extensionContext) {
...
+        final String namespaceName = StUtils.getNamespaceBasedOnRbac(clusterOperator.getDeploymentNamespace(), extensionContext);
+        final String clusterName = mapWithClusterNames.get(extensionContext.getDisplayName());
+        final String topicName = mapWithTestTopics.get(extensionContext.getDisplayName());
+        final String userName = mapWithTestUsers.get(extensionContext.getDisplayName());
+
+        resourceManager.createResource(extensionContext, KafkaTemplates.kafkaEphemeral(clusterName, 3)
+                .editSpec()
+                .editKafka()
+                .withListeners(new GenericKafkaListenerBuilder()
+                        .withName(Constants.EXTERNAL_LISTENER_DEFAULT_NAME)
```
scholzj, line 725: Same as above.

systemtest/src/test/java/io/strimzi/systemtest/kafka/listeners/ListenersST.java
```diff
@@ -668,6 +668,92 @@ void testLoadBalancerTls(ExtensionContext extensionContext) {
         );
     }
 
+    @ParallelNamespaceTest
+    @Tag(SANITY)
+    @Tag(ACCEPTANCE)
```
scholzj, line 673: I think we do not want it in the ACCEPTANCE test suite since it would make it too long. I think this should say `@Tag(REGRESSION)` for both of the tests, or @im-konge

scholzj (commented):
LGTM. I have few more last nits about some naming in the tests and in the docs. But the actual code seems good to me. I also left some comments on the ST, but I will start the tests here to see if it passes or not.

systemtest/src/test/java/io/strimzi/systemtest/kafka/listeners/ListenersST.java
im-konge: I agree with @scholzj , `SANITY` profile contains tests for each component (just a sanity check that everything is working). When you take a look on a `testNodePort` and other, we are running it just for regression. So I would go this way here too.

You don't need to add the `REGRESSION` tag, it will be taken from the whole class.

systemtest/src/test/java/io/strimzi/systemtest/kafka/listeners/ListenersST.java
```diff
@@ -668,6 +668,92 @@ void testLoadBalancerTls(ExtensionContext extensionContext) {
...
+                .withListeners(new GenericKafkaListenerBuilder()
+                        .withName(Constants.EXTERNAL_LISTENER_DEFAULT_NAME)
+                        .withPort(9102)
+                        .withType(KafkaListenerType.CLUSTER_IP)
+                        .withTls(false)
+                        .build())
+                .withConfig(Collections.singletonMap("default.replication.factor", 3))
+                .endKafka()
+                .endSpec()
+                .build());
```
im-konge, line 692: Can you please fix the indents as we have in our other tests?

systemtest/src/test/java/io/strimzi/systemtest/kafka/listeners/ListenersST.java
im-konge: If it isn't external listener and we can use the internal listener here, let's go with it.

systemtest/src/test/java/io/strimzi/systemtest/kafka/listeners/ListenersST.java
```diff
@@ -668,6 +668,92 @@ void testLoadBalancerTls(ExtensionContext extensionContext) {
...
+                .withNamespaceName(namespaceName)
+                .withClusterName(clusterName)
+                .withMessageCount(MESSAGE_COUNT)
+                .withListenerName(Constants.EXTERNAL_LISTENER_DEFAULT_NAME)
+                .build();
+
+        externalKafkaClient.verifyProducedAndConsumedMessages(
+                externalKafkaClient.sendMessagesPlain(),
+                externalKafkaClient.receiveMessagesPlain()
+        );
```
im-konge, line 707: If it is not an external listener (as it is not used as an external listener), do we need the external clients instead of "internal"? You can check the `testSendMessagesPlainAnonymous`, which uses the `KafkaClients`

systemtest/src/test/java/io/strimzi/systemtest/kafka/listeners/ListenersST.java
```diff
@@ -668,6 +668,92 @@ void testLoadBalancerTls(ExtensionContext extensionContext) {
...
+                        .withName(Constants.EXTERNAL_LISTENER_DEFAULT_NAME)
+                        .withPort(9103)
+                        .withType(KafkaListenerType.CLUSTER_IP)
+                        .withTls(true)
+                        .withAuth(new KafkaListenerAuthenticationTls())
+                        .build())
+                .withConfig(Collections.singletonMap("default.replication.factor", 3))
+                .endKafka()
+                .endSpec()
+                .build());
```
im-konge, line 734: Indent (as above)

systemtest/src/test/java/io/strimzi/systemtest/kafka/listeners/ListenersST.java
```diff
@@ -668,6 +668,92 @@ void testLoadBalancerTls(ExtensionContext extensionContext) {
...
+                .withMessageCount(MESSAGE_COUNT)
+                .withKafkaUsername(userName)
+                .withSecurityProtocol(SecurityProtocol.SSL)
+                .withListenerName(Constants.EXTERNAL_LISTENER_DEFAULT_NAME)
+                .build();
+
+        externalKafkaClient.verifyProducedAndConsumedMessages(
+                externalKafkaClient.sendMessagesTls(),
+                externalKafkaClient.receiveMessagesTls()
+        );
```
im-konge, line 753: Same as above (with the `KafkaClients`)

systemtest/src/test/java/io/strimzi/systemtest/kafka/listeners/ListenersST.java
im-konge: You can add it to the `Constants` class.

im-konge (commented):
Few minor comments, thanks for adding the STs :) good job

systemtest/src/test/java/io/strimzi/systemtest/kafka/listeners/ListenersST.java
maknihamdi: I dont thins we can use ssl with regular KafkaClient, I didn't find equivalent to ".withSecurityProtocol(SecurityProtocol.SSL)"

systemtest/src/test/java/io/strimzi/systemtest/kafka/listeners/ListenersST.java
maknihamdi: finally I have to keep SANITY and remove ACCEPTANCE, right?

systemtest/scripts/results.json
```diff
@@ -0,0 +1 @@
+{"body":":heavy_exclamation_mark: **Systemtests Failed (no tests results are present)** :heavy_exclamation_mark:"}
```
scholzj, line 1: This probably should not be committed?

scholzj (commented):
The code LGTM - but I think there is one file which should not be there. I will run the regression tests.

@im-konge Can you please review the ST?

systemtest/scripts/results.json
maknihamdi: oops. I remove it and add it to gitingore

systemtest/src/test/java/io/strimzi/systemtest/kafka/listeners/ListenersST.java
```diff
@@ -668,6 +668,90 @@ void testLoadBalancerTls(ExtensionContext extensionContext) {
...
+            .withListeners(new GenericKafkaListenerBuilder()
+                .withName(Constants.CLUSTER_IP_LISTENER_DEFAULT_NAME)
+                .withPort(9102)
+                .withType(KafkaListenerType.CLUSTER_IP)
+                .withTls(false)
+                .build())
+            .withConfig(Collections.singletonMap("default.replication.factor", 3))
+            .endKafka()
+            .endSpec()
+            .build());
```
im-konge, line 691:
```suggestion
        resourceManager.createResource(extensionContext, KafkaTemplates.kafkaEphemeral(clusterName, 3)
            .editSpec()
                .editKafka()
                    .withListeners(new GenericKafkaListenerBuilder()
                        .withName(Constants.CLUSTER_IP_LISTENER_DEFAULT_NAME)
                        .withPort(9102)
                        .withType(KafkaListenerType.CLUSTER_IP)
                        .withTls(false)
                        .build())
                    .withConfig(Collections.singletonMap("default.replication.factor", 3))
                .endKafka()
            .endSpec()
            .build());
```

systemtest/src/test/java/io/strimzi/systemtest/kafka/listeners/ListenersST.java
im-konge: Remove the `SANITY` too please, as these don't need to be in this group of sanity tests

systemtest/src/test/java/io/strimzi/systemtest/kafka/listeners/ListenersST.java
im-konge: Makes sense

systemtest/src/test/java/io/strimzi/systemtest/kafka/listeners/ListenersST.java
```diff
@@ -668,6 +668,90 @@ void testLoadBalancerTls(ExtensionContext extensionContext) {
...
+            .withNamespaceName(namespaceName)
+            .withClusterName(clusterName)
+            .withMessageCount(MESSAGE_COUNT)
+            .withListenerName(Constants.CLUSTER_IP_LISTENER_DEFAULT_NAME)
+            .build();
+
+        externalKafkaClient.verifyProducedAndConsumedMessages(
+                externalKafkaClient.sendMessagesPlain(),
+                externalKafkaClient.receiveMessagesPlain()
+        );
```
im-konge, line 706: Why don't you use the `KafkaClients`? Or do you want some help with it?

systemtest/src/test/java/io/strimzi/systemtest/kafka/listeners/ListenersST.java
```diff
@@ -668,6 +668,90 @@ void testLoadBalancerTls(ExtensionContext extensionContext) {
...
+                .withName(Constants.CLUSTER_IP_LISTENER_DEFAULT_NAME)
+                .withPort(9103)
+                .withType(KafkaListenerType.CLUSTER_IP)
+                .withTls(true)
+                .withAuth(new KafkaListenerAuthenticationTls())
+                .build())
+            .withConfig(Collections.singletonMap("default.replication.factor", 3))
+            .endKafka()
+            .endSpec()
+            .build());
```
im-konge, line 732:
```suggestion
        resourceManager.createResource(extensionContext, KafkaTemplates.kafkaEphemeral(clusterName, 3)
            .editSpec()
                .editKafka()
                    .withListeners(new GenericKafkaListenerBuilder()
                        .withName(Constants.CLUSTER_IP_LISTENER_DEFAULT_NAME)
                        .withPort(9103)
                        .withType(KafkaListenerType.CLUSTER_IP)
                        .withTls(true)
                        .withAuth(new KafkaListenerAuthenticationTls())
                        .build())
                    .withConfig(Collections.singletonMap("default.replication.factor", 3))
                .endKafka()
            .endSpec()
            .build());
```

systemtest/src/test/java/io/strimzi/systemtest/kafka/listeners/ListenersST.java
```diff
@@ -668,6 +668,90 @@ void testLoadBalancerTls(ExtensionContext extensionContext) {
...
+            .withMessageCount(MESSAGE_COUNT)
+            .withKafkaUsername(userName)
+            .withSecurityProtocol(SecurityProtocol.SSL)
+            .withListenerName(Constants.CLUSTER_IP_LISTENER_DEFAULT_NAME)
+            .build();
+
+        externalKafkaClient.verifyProducedAndConsumedMessages(
+                externalKafkaClient.sendMessagesTls(),
+                externalKafkaClient.receiveMessagesTls()
+        );
```
im-konge, line 751: Same as above

systemtest/src/test/java/io/strimzi/systemtest/kafka/listeners/ListenersST.java
```diff
@@ -668,6 +668,90 @@ void testLoadBalancerTls(ExtensionContext extensionContext) {
         );
     }
 
+    @ParallelNamespaceTest
+    @Tag(SANITY)
+    @Tag(INTERNAL_CLIENTS_USED)
+    void testClusterIP(ExtensionContext extensionContext) {
```
im-konge, line 674:
```suggestion
    void testClusterIp(ExtensionContext extensionContext) {
```

systemtest/src/test/java/io/strimzi/systemtest/kafka/listeners/ListenersST.java
```diff
@@ -668,6 +668,90 @@ void testLoadBalancerTls(ExtensionContext extensionContext) {
...
+                externalKafkaClient.sendMessagesPlain(),
+                externalKafkaClient.receiveMessagesPlain()
+        );
+
+    }
+
+    @ParallelNamespaceTest
+    @Tag(SANITY)
+    @Tag(INTERNAL_CLIENTS_USED)
+    void testClusterIPTls(ExtensionContext extensionContext) {
```
im-konge, line 713:
```suggestion
    void testClusterIpTls(ExtensionContext extensionContext) {
```

im-konge (commented):
Few comments, if you'll need any help with `KafkaClients`, let me know.

systemtest/src/test/java/io/strimzi/systemtest/kafka/listeners/ListenersST.java
```diff
@@ -668,6 +668,102 @@ void testLoadBalancerTls(ExtensionContext extensionContext) {
...
+    @Tag(INTERNAL_CLIENTS_USED)
+    void testClusterIp(ExtensionContext extensionContext) {
+        final TestStorage testStorage = new TestStorage(extensionContext);
+
+        final String namespaceName = testStorage.getNamespaceName();
+        final String clusterName = testStorage.getClusterName();
+        final String topicName = testStorage.getTopicName();
+        final String userName = testStorage.getUserName();
+        final String producerName = testStorage.getProducerName();
+        final String consumerName = testStorage.getConsumerName();
```
im-konge, line 681: You can use the `TestStorage` directly and "save some place" not creating new variables for that. (jfyi)

systemtest/src/test/java/io/strimzi/systemtest/kafka/listeners/ListenersST.java
```diff
@@ -668,6 +668,102 @@ void testLoadBalancerTls(ExtensionContext extensionContext) {
...
+        resourceManager.createResource(extensionContext, kafkaClients.producerStrimzi());
+        ClientUtils.waitForClientSuccess(testStorage.getProducerName(), testStorage.getNamespaceName(), MESSAGE_COUNT);
+
+        resourceManager.createResource(extensionContext, kafkaClients.consumerStrimzi());
+        ClientUtils.waitForClientSuccess(testStorage.getConsumerName(), testStorage.getNamespaceName(), MESSAGE_COUNT);
+
+    }
+
+    @ParallelNamespaceTest
+    @Tag(SANITY)
```
im-konge, line 718: Can you please remove the `SANITY` tag here as well?

im-konge (approved):
Just minor comments. For systemtest POV, this LGTM. Thanks a lot for your contribution!

systemtest/src/test/java/io/strimzi/systemtest/kafka/listeners/ListenersST.java
maknihamdi: aha :) I guessed it, but I figured if Lukas hadn't mentioned it in the last reviews, maybe we should keep it.

I'll remove it

im-konge: Sorry about that! My bad :)

systemtest/src/test/java/io/strimzi/systemtest/kafka/listeners/ListenersST.java
```diff
@@ -668,6 +668,102 @@ void testLoadBalancerTls(ExtensionContext extensionContext) {
...
+                        .withPort(9102)
+                        .withType(KafkaListenerType.CLUSTER_IP)
+                        .withTls(false)
+                        .build())
+                    .withConfig(Collections.singletonMap("default.replication.factor", 3))
+                .endKafka()
+            .endSpec()
+            .build());
+
+        ServiceUtils.waitUntilAddressIsReachable(KafkaResource.kafkaClient().inNamespace(namespaceName).withName(clusterName).get().getStatus().getListeners().get(0).getAddresses().get(0).getHost());
```
maknihamdi, line 697: @im-konge do you think is necessary to have this check with an internal listener?

im-konge: IIRC this is check for the external listeners mainly
So you can remove it. good catch :)

maknihamdi: I admit it was a "stupid" copy from other tests. (like port number, btw I have to fix it also)

maknihamdi:
@im-konge @scholzj it will be something like:

`String serviceName = String.format("%s-kafka-%s-bootstrap", clusterName, Constants.CLUSTER_IP_LISTENER_DEFAULT_NAME);`
`String bootstrapServiceName = String.format("%s.%s.svc", serviceName, namespaceName);`

do you prefer to extract this 2 lines of code to a special method in https://github.com/strimzi/strimzi-kafka-operator/blob/main/api/src/main/java/io/strimzi/api/kafka/model/KafkaResources.java ? It will be useful for a couple of tests. It can be `    public static String customListenerBootstrapServiceName(String namespaceName, String clusterName, String listenerName) {...}
`
we have some special thinks with this listener, it's not an plain internal one, neither an external one, this is why when reconciler compute bootstrap address, we have the listener name in the middle (cf https://github.com/strimzi/strimzi-kafka-operator/blob/main/cluster-operator/src/main/java/io/strimzi/operator/cluster/model/ListenersUtils.java#L280)

im-konge:
@maknihamdi in STs we are sometimes taking the bootstrap server name from the Kafka CR, when the Kafka is deployed and ready (that means after `resourceManager.createResource();` method). I don't know if we should have some method in the `api` module, that's something for @scholzj . But if it would be used just in tests, I guess that it's not really necessary to add it there.

scholzj:
FYI: I'm traveling this week. If needed I can look more into why the test is failing next week when I get back.

systemtest/src/test/java/io/strimzi/systemtest/kafka/listeners/ListenersST.java
```diff
@@ -668,6 +668,100 @@ void testLoadBalancerTls(ExtensionContext extensionContext) {
         );
     }
 
+    private String clusterIpBootstrapServiceName(String namespaceName, String clusterName, String listenerName) {
+        String serviceName = String.format("%s-kafka-%s-bootstrap", clusterName, listenerName);
+        String bootstrapServiceName = String.format("%s.%s.svc", serviceName, namespaceName);
+        return bootstrapServiceName;
+    }
```
im-konge, line 675: Okay, this looks good.
Can you move it to `KafkaUtils` please?
Also, do you think that it will be same for every environment? As I mentioned in previous comment - to be sure that the name will be correct, we are getting it from the Kafka status (you can check it in `AbstractKafkaClient` in `getBootstrapServerFromStatus` method). Do you think it will be better to get if from status, to prevent some future issues with different bootstrap name?

maknihamdi: Yes! Sure!
I can add a similar method to KafkaClient, or make KafkaClient implementing AbstractKafkaClient (there is a reason why internalClients.KafkaClient didn't implement AbstractKafkaClient?) to take advantage of this class

maknihamdi: something like
 `    public String computeBootstrapAddressFromStatus(String clusterName, String listenerName) { 
//same code as AbstractKafkaClient.getBootstrapServerFromStatus }
` 
returning bootstrapAddress, which can be used to set bootstrapAddress field on client. The method can be added to internalClients.KafkaClient or internalClients.BaseClient to be used by other clients (bridge, ...)

what do you prefer?

im-konge: I think that it would be out of scope of this PR, I think in this PR you can just add that method to the `KafkaUtils` and use it in those two tests.
External and internal clients were a bit different in the past, I think we can make it more abstract in the future.
Anyway, thanks for the changes :)

maknihamdi: done.

ListenersST test run successfully in my local env

thanks for your review !

systemtest/src/main/java/io/strimzi/systemtest/utils/kafkaUtils/KafkaUtils.java
```diff
@@ -502,4 +502,26 @@ public static String namespacedTlsBootstrapAddress(String clusterName, String na
...
+            LOGGER.info("Listener name is not specified. Picking the first one from the Kafka Status.");
+            return listenerStatusList.get(0).getBootstrapServers();
+        }
+
+        return listenerStatusList.stream().filter(listener -> listener.getName().equals(listenerName))
+                .findFirst()
+                .orElseThrow(RuntimeException::new)
+                .getBootstrapServers();
+
+
```
im-konge, line 524: Can you please remove those two lines? Thanks!

im-konge (approved):
Just one small nit, thank you for your changes to the STs :)

systemtest/src/main/java/io/strimzi/systemtest/utils/kafkaUtils/KafkaUtils.java
maknihamdi: removed ;)

scholzj:
@maknihamdi I wanted to merge it and I built it and tried it locally just because of it. Unfortunately, I found two more issues.

1) The bootstrap service does not registr all the proper DNS names for the certificate SANs. It seems to register only the one with the format `<service-name>.<namespace>.svc`. It should also register `<service-name>.<namespace>`, `<service-name>` and `<service-name>.<namespace>.svc.<cluster-DNS-suffix>` as done for example here: https://github.com/strimzi/strimzi-kafka-operator/blob/ba1fe15383510c061be1c64189c57c7c41b95c5f/cluster-operator/src/main/java/io/strimzi/operator/cluster/model/ClusterCa.java#L188-L192

    This is not necessarily a blocker -> I can do this in a follow-up PR if needed ... however, the second issue is a blocker and needs to be fixed.

2) The broker DNS name is not registered anywhere and added tot he SAN. So even if you use the bootstrap address whihc is part of the certificate, you still run into this:

    ```
    Caused by: javax.net.ssl.SSLHandshakeException: No subject alternative DNS name matching my-cluster-kafka-tls-2.myproject.svc found.
	    at java.base/sun.security.ssl.Alert.createSSLException(Alert.java:131)
	    at java.base/sun.security.ssl.TransportContext.fatal(TransportContext.java:349)
	    at java.base/sun.security.ssl.TransportContext.fatal(TransportContext.java:292)
	    at java.base/sun.security.ssl.TransportContext.fatal(TransportContext.java:287)
	    at java.base/sun.security.ssl.CertificateMessage$T13CertificateConsumer.checkServerCerts(CertificateMessage.java:1357)
	    at java.base/sun.security.ssl.CertificateMessage$T13CertificateConsumer.onConsumeCertificate(CertificateMessage.java:1232)
	    at java.base/sun.security.ssl.CertificateMessage$T13CertificateConsumer.consume(CertificateMessage.java:1175)
	    at java.base/sun.security.ssl.SSLHandshake.consume(SSLHandshake.java:392)
	    at java.base/sun.security.ssl.HandshakeContext.dispatch(HandshakeContext.java:443)
	    at java.base/sun.security.ssl.SSLEngineImpl$DelegatedTask$DelegatedAction.run(SSLEngineImpl.java:1074)
	    at java.base/sun.security.ssl.SSLEngineImpl$DelegatedTask$DelegatedAction.run(SSLEngineImpl.java:1061)
	    at java.base/java.security.AccessController.doPrivileged(Native Method)
	    at java.base/sun.security.ssl.SSLEngineImpl$DelegatedTask.run(SSLEngineImpl.java:1008)
	    at org.apache.kafka.common.network.SslTransportLayer.runDelegatedTasks(SslTransportLayer.java:430)
	    at org.apache.kafka.common.network.SslTransportLayer.handshakeUnwrap(SslTransportLayer.java:514)
	    at org.apache.kafka.common.network.SslTransportLayer.doHandshake(SslTransportLayer.java:368)
	    at org.apache.kafka.common.network.SslTransportLayer.handshake(SslTransportLayer.java:291)
	    at org.apache.kafka.common.network.KafkaChannel.prepare(KafkaChannel.java:178)
	    at org.apache.kafka.common.network.Selector.pollSelectionKeys(Selector.java:543)
	    at org.apache.kafka.common.network.Selector.poll(Selector.java:481)
	    at org.apache.kafka.clients.NetworkClient.poll(NetworkClient.java:561)
	    at org.apache.kafka.clients.consumer.internals.ConsumerNetworkClient.poll(ConsumerNetworkClient.java:265)
	    at org.apache.kafka.clients.consumer.internals.ConsumerNetworkClient.poll(ConsumerNetworkClient.java:236)
	    at org.apache.kafka.clients.consumer.internals.ConsumerNetworkClient.poll(ConsumerNetworkClient.java:227)
	    at org.apache.kafka.clients.consumer.internals.ConsumerNetworkClient.awaitMetadataUpdate(ConsumerNetworkClient.java:164)
	    at org.apache.kafka.clients.consumer.internals.AbstractCoordinator.ensureCoordinatorReady(AbstractCoordinator.java:257)
	    at org.apache.kafka.clients.consumer.internals.ConsumerCoordinator.poll(ConsumerCoordinator.java:480)
	    at org.apache.kafka.clients.consumer.KafkaConsumer.updateAssignmentMetadataIfNeeded(KafkaConsumer.java:1261)
	    at org.apache.kafka.clients.consumer.KafkaConsumer.poll(KafkaConsumer.java:1230)
	    at org.apache.kafka.clients.consumer.KafkaConsumer.poll(KafkaConsumer.java:1210)
	    at io.vertx.kafka.client.consumer.impl.KafkaReadStreamImpl.lambda$pollRecords$6(KafkaReadStreamImpl.java:154)
	    at java.base/java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:515)
	    at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:264)
	    at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128)
	    at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628)
	    at java.base/java.lang.Thread.run(Thread.java:829)
    Caused by: java.security.cert.CertificateException: No subject alternative DNS name matching my-cluster-kafka-tls-2.myproject.svc found.
	    at java.base/sun.security.util.HostnameChecker.matchDNS(HostnameChecker.java:212)
	    at java.base/sun.security.util.HostnameChecker.match(HostnameChecker.java:103)
	    at java.base/sun.security.ssl.X509TrustManagerImpl.checkIdentity(X509TrustManagerImpl.java:455)
	    at java.base/sun.security.ssl.X509TrustManagerImpl.checkIdentity(X509TrustManagerImpl.java:415)
	    at java.base/sun.security.ssl.X509TrustManagerImpl.checkTrusted(X509TrustManagerImpl.java:283)
	    at java.base/sun.security.ssl.X509TrustManagerImpl.checkServerTrusted(X509TrustManagerImpl.java:141)
	    at java.base/sun.security.ssl.CertificateMessage$T13CertificateConsumer.checkServerCerts(CertificateMessage.java:1335)
	    ... 31 more
    ```

    which renders it unusable for any client which tries to use it with TLS. We need to register the DNS name also when the user override is not configured (in this case, `<service-name>.<namespace>.svc` should be enough because this is the follow-up connection based on the advertised address). This is a blocker because without it it basically doesn't work.

We plan to start a new release this week. So please let me know if you need any help with this so tht we can close it in time for the new release and don't have to wait for the next one.

scholzj:
Thanks for the PR.

PR #7366, opened 2022-09-21
Add OpenTelemetry support

### LINKED ISSUE #7021: Add support for OpenTelemetry
The OpenTracing project was archived by CNCF and the new tracing project in the landscape is now [OpenTelemetry](https://opentelemetry.io/).
We should add the support for OpenTelemetry on Kafka Connect, Mirror Maker 1 and 2 while maintaining the support for OpenTracing for backward compatibility but then removing it finally in one of the coming releases.

ppatierno:
### Type of change

- Enhancement / new feature

### Description

This PR fixes #7021 adding the support for OpenTelemetry in the operator by enabling it on Kafka Connect, MM and MM2.
Documentation will be addressed in a different PR.

### Checklist

- [x] Write tests
- [x] Update CHANGELOG.md

### DIFF
--- a/api/src/main/java/io/strimzi/api/kafka/model/tracing/Tracing.java
+++ b/api/src/main/java/io/strimzi/api/kafka/model/tracing/Tracing.java
@@ -23,6 +23,7 @@ import java.util.Map;
         property = "type")
 @JsonSubTypes({
     @JsonSubTypes.Type(name = JaegerTracing.TYPE_JAEGER, value = JaegerTracing.class),
+    @JsonSubTypes.Type(name = OpenTelemetryTracing.TYPE_OPENTELEMETRY, value = OpenTelemetryTracing.class),
 })
 @JsonInclude(JsonInclude.Include.NON_NULL)
 @EqualsAndHashCode
@@ -32,8 +33,8 @@ public abstract class Tracing implements UnknownPropertyPreserving, Serializable
     private Map<String, Object> additionalProperties;
 
     @Description("Type of the tracing used. " +
-            "Currently the only supported type is `jaeger` for Jaeger tracing. " +
-            "The Jaeger tracing is deprecated")
+            "Currently the only supported types are `jaeger` for OpenTracing (Jaeger) tracing and `opentelemetry` for OpenTelemetry tracing. " +
+            "The OpenTracing (Jaeger) tracing is deprecated")
     public abstract String getType();
 
     @Override
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaConnectCluster.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaConnectCluster.java
@@ -53,6 +53,8 @@ import io.strimzi.api.kafka.model.connect.ExternalConfigurationEnv;
 import io.strimzi.api.kafka.model.connect.ExternalConfigurationEnvVarSource;
 import io.strimzi.api.kafka.model.connect.ExternalConfigurationVolumeSource;
 import io.strimzi.api.kafka.model.template.KafkaConnectTemplate;
+import io.strimzi.api.kafka.model.tracing.JaegerTracing;
+import io.strimzi.api.kafka.model.tracing.OpenTelemetryTracing;
 import io.strimzi.api.kafka.model.tracing.Tracing;
 import io.strimzi.operator.cluster.ClusterOperatorConfig;
 import io.strimzi.operator.cluster.model.securityprofiles.ContainerSecurityProviderContextImpl;
@@ -197,7 +199,7 @@ public class KafkaConnectCluster extends AbstractModel {
      * from the instantiation of the (subclass of) KafkaConnectCluster,
      * thus permitting reuse of the setter-calling code for subclasses.
      */
-    @SuppressWarnings({"checkstyle:CyclomaticComplexity", "checkstyle:NPathComplexity"})
+    @SuppressWarnings({"checkstyle:CyclomaticComplexity", "checkstyle:NPathComplexity", "deprecation"})
     protected static <C extends KafkaConnectCluster> C fromSpec(Reconciliation reconciliation,
                                                                 KafkaConnectSpec spec,
                                                                 KafkaVersion.Lookup versions,
@@ -211,8 +213,13 @@ public class KafkaConnectCluster extends AbstractModel {
             kafkaConnect.setConfiguration(config);
         }
         if (kafkaConnect.tracing != null)   {
-            config.setConfigOption("consumer.interceptor.classes", "io.opentracing.contrib.kafka.TracingConsumerInterceptor");
-            config.setConfigOption("producer.interceptor.classes", "io.opentracing.contrib.kafka.TracingProducerInterceptor");
+            if (JaegerTracing.TYPE_JAEGER.equals(kafkaConnect.tracing.getType())) {
+                config.setConfigOption("consumer.interceptor.classes", JaegerTracing.CONSUMER_INTERCEPTOR_CLASS_NAME);
+                config.setConfigOption("producer.interceptor.classes", JaegerTracing.PRODUCER_INTERCEPTOR_CLASS_NAME);
+            } else if (OpenTelemetryTracing.TYPE_OPENTELEMETRY.equals(kafkaConnect.tracing.getType())) {
+                config.setConfigOption("consumer.interceptor.classes", OpenTelemetryTracing.CONSUMER_INTERCEPTOR_CLASS_NAME);
+                config.setConfigOption("producer.interceptor.classes", OpenTelemetryTracing.PRODUCER_INTERCEPTOR_CLASS_NAME);
+            }
         }
 
         if (kafkaConnect.getImage() == null) {
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaMirrorMakerCluster.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaMirrorMakerCluster.java
@@ -25,6 +25,8 @@ import io.strimzi.api.kafka.model.KafkaMirrorMakerSpec;
 import io.strimzi.api.kafka.model.Probe;
 import io.strimzi.api.kafka.model.ProbeBuilder;
 import io.strimzi.api.kafka.model.template.KafkaMirrorMakerTemplate;
+import io.strimzi.api.kafka.model.tracing.JaegerTracing;
+import io.strimzi.api.kafka.model.tracing.OpenTelemetryTracing;
 import io.strimzi.api.kafka.model.tracing.Tracing;
 import io.strimzi.operator.cluster.model.securityprofiles.ContainerSecurityProviderContextImpl;
 import io.strimzi.operator.cluster.model.securityprofiles.PodSecurityProviderContextImpl;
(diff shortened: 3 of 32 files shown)


cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaMirrorMakerCluster.java
```diff
@@ -311,7 +312,12 @@ private KafkaMirrorMakerConsumerConfiguration getConsumerConfiguration() {
         KafkaMirrorMakerConsumerConfiguration config = new KafkaMirrorMakerConsumerConfiguration(reconciliation, consumer.getConfig().entrySet());
 
         if (tracing != null) {
-            config.setConfigOption("interceptor.classes", "io.opentracing.contrib.kafka.TracingConsumerInterceptor");
+            // using hard-coded string because of the -Werror flag
+            if ("jaeger".equals(tracing.getType())) {
```
see-quick, line 316: Hmm, I wonder if using switch could resolve your problem with `-Werror` and then use just
```java
switch (tracing.getType) { 
    case OpenTelemetryTracing.TYPE_OPENTELEMETRY:
       ....
    case JaegerTracing.TYPE_JAEGER:
      ....
```

ppatierno: Actually I was not having the problem with the same code in the `KafkaMirrorMaker2AssemblyOperator` then I noticed that there was a `@SuppressWarnings("deprecation")` there already. I will use the same here.

tracing-agent/src/main/java/io/strimzi/tracing/agent/TracingAgent.java
```diff
@@ -4,34 +4,30 @@
...
-            if (jaegerServiceName != null) {
-                LOGGER.info("Initializing Jaeger tracing with service name {}", jaegerServiceName);
-                Tracer tracer = Configuration.fromEnv().getTracer();
-                GlobalTracer.registerIfAbsent(tracer);
-            } else {
-                LOGGER.error("Jaeger tracing cannot be initialized because JAEGER_SERVICE_NAME environment variable is not defined");
-            }
+        Tracing tracing;
+        switch (agentArgs) {
+            case "jaeger":
```
im-konge, line 20: I see that you are using "hardcoded" values on few places, would it make sense to use some constants, for example from the `api` module (like `TYPE_JAEGER`)?

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaConnectCluster.java
```diff
@@ -211,8 +213,13 @@ protected static <C extends KafkaConnectCluster> C fromSpec(Reconciliation recon
             kafkaConnect.setConfiguration(config);
         }
         if (kafkaConnect.tracing != null)   {
-            config.setConfigOption("consumer.interceptor.classes", "io.opentracing.contrib.kafka.TracingConsumerInterceptor");
-            config.setConfigOption("producer.interceptor.classes", "io.opentracing.contrib.kafka.TracingProducerInterceptor");
+            if (JaegerTracing.TYPE_JAEGER.equals(kafkaConnect.tracing.getType())) {
+                config.setConfigOption("consumer.interceptor.classes", "io.opentracing.contrib.kafka.TracingConsumerInterceptor");
+                config.setConfigOption("producer.interceptor.classes", "io.opentracing.contrib.kafka.TracingProducerInterceptor");
```
im-konge, line 218: Can we have some constants for those classes (config options)?

tracing-agent/src/main/java/io/strimzi/tracing/agent/TracingAgent.java
ppatierno: Tbh I wanted to avoid to add the `api` module dependency to the agent just for this ...

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaConnectCluster.java
ppatierno: I don't see big value on using constants here. The Jaeger part will go away at some point soon anyway.

im-konge: My point was that when you'll have for example one class which would hold these information for the Jaeger part, once we'll remove it, it will be much easier and we will be sure that we didn't forget anything 🤷 . But yeah, it was just an idea.

tracing-agent/src/main/java/io/strimzi/tracing/agent/TracingAgent.java
im-konge: Ah okay, make sense. Thanks :)

scholzj (commented):
Nice job. I left some smaller comments. But LGTM overall. What is missing is the record for CHANGELO.md which this surely deserves.

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaConnectCluster.java
ppatierno: I can create constants for these classes and the corresponding ones for OpenTelemetry. The open question is where?
These classes are used mostly in the model and assembly related stuff. We could think of using the `ModelUtils` as a central place or having them in the `api` module in each `xxxxTracing` class definition but it would be exposed to users who use our api. I have no strong opinion. @scholzj what's your take on this?

api/src/main/java/io/strimzi/api/kafka/model/tracing/Tracing.java
```diff
@@ -32,7 +33,7 @@ public abstract class Tracing implements UnknownPropertyPreserving, Serializable
     private Map<String, Object> additionalProperties;
 
     @Description("Type of the tracing used. " +
-            "Currently the only supported type is `jaeger` for Jaeger tracing. " +
+            "Currently the only supported types are `jaeger` for Jaeger tracing and `opentelemetry` for OpenTelemetry tracing. " +
```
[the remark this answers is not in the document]
ppatierno: Yeah it makes sense. But the only idea I have is replacing " ... for Jaeger tracing ...." with something like " ... for OpenTracing (Jaeger) tracing ..." so highlighting that the tracing system is OpenTracing but leaving a Jaeger reference as it is the value for the configuration.

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaConnectCluster.java
```diff
@@ -197,7 +199,7 @@ public static KafkaConnectCluster fromCrd(Reconciliation reconciliation, KafkaCo
      * from the instantiation of the (subclass of) KafkaConnectCluster,
      * thus permitting reuse of the setter-calling code for subclasses.
      */
-    @SuppressWarnings({"checkstyle:CyclomaticComplexity", "checkstyle:NPathComplexity"})
+    @SuppressWarnings({"checkstyle:CyclomaticComplexity", "checkstyle:NPathComplexity", "deprecation"})
```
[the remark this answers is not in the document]
ppatierno: I don't get you ... the suppression is on the `fromSpec` method. What am I missing?

cluster-operator/src/test/java/io/strimzi/operator/cluster/model/KafkaBridgeClusterTest.java
```diff
@@ -1096,18 +1097,39 @@ public void testProbeConfiguration()   {
...
+                throw new IllegalArgumentException("The '" + type + "' is not a valid tracing type");
+        }
+        KafkaBridge resource = builder.build();
 
         KafkaBridgeCluster kb = KafkaBridgeCluster.fromCrd(Reconciliation.DUMMY_RECONCILIATION, resource, VERSIONS);
         Deployment deployment = kb.generateDeployment(new HashMap<>(), true, null, null);
         Container container = deployment.getSpec().getTemplate().getSpec().getContainers().get(0);
 
-        assertThat(AbstractModel.containerEnvVars(container).get(KafkaBridgeCluster.ENV_VAR_STRIMZI_TRACING), is("jaeger"));
+        assertThat(AbstractModel.containerEnvVars(container).get(KafkaBridgeCluster.ENV_VAR_STRIMZI_TRACING), is(type));
```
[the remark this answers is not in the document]
ppatierno: If I got your comment right, are you referring to the other env vars that the user has to set through the `template` (i.e. JAEGER_SERVICE_NAME or OTEL_SERVICE_NAME and so on)? In this case we need to add them in the test code to add them but does it make sense? Or which config values are you referring to?

pom.xml
```diff
@@ -124,6 +124,7 @@
         <jaeger.version>1.8.1</jaeger.version>
         <opentracing.version>0.33.0</opentracing.version>
         <opentracing-kafka.version>0.1.13</opentracing-kafka.version>
+        <opentelemetry.alpha-version>1.18.0-alpha</opentelemetry.alpha-version>
```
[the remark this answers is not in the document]
ppatierno: So let me explain the gist of this better ...
Right now in the OpenTelemetry land, some libraries have a "definitive" version, while others are still in "alpha" because evolving quickly. 
For bridge support I preferred to use `opentelemetry.version` property to refer the "definitive" version, while `opentelemetry.alpha-version` for the "alpha" ones that I would expect to disappear at some point.
For example, I am using `opentelemetry.version` property in the 3rd library pom files where the `opentelemetry-exporter-jaeger` dependency is a final release but at same time I have `opentelemetry.alpha-version` used for the `opentelemetry-kafka-clients-2.6` and `opentelemetry-sdk-extension-autoconfigure` which are still marked as "alpha". The same rule applies here.

ppatierno:
@scholzj Thanks for the review! I replied questions and feedback. Regarding the CHANGELOG, I totally agree. I have it written down yet in 0.32.0 because it needs anyway doc which I should work together with @PaulRMellor and I am not sure we'll meet 0.32.0 or the next one. It depends on the dates.

scholzj:
> @scholzj Thanks for the review! I replied questions and feedback. Regarding the CHANGELOG, I totally agree. I have it written down yet in 0.32.0 because it needs anyway doc which I should work together with @PaulRMellor and I am not sure we'll meet 0.32.0 or the next one. It depends on the dates.

The CHANGELOG is normally based on the feature being added and not on the docs. So I think if you plan to merge it for 0.32 (which I assume you do -> there are still few weeks until 0.32), it should be IMHO added there in this PR.

cluster-operator/src/test/java/io/strimzi/operator/cluster/model/KafkaBridgeClusterTest.java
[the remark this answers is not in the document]
scholzj: I assume you are also passing some options for the bridge configuration? Or is this the only option? Check for example the Connect test which also checks the COnnect config for the interceptors etc.

pom.xml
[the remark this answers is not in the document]
scholzj: Seems a bit weird if you have only the `alpha` version. Maybe it should be at least called `opentelemetry-alpha.version`? Anyway, I can live with this, so feel free to keep it.

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaConnectCluster.java
ppatierno: Ok, `api` module works for me. I just noticed the usage with Authorization classes, so it makes sense doing the same way.

cluster-operator/src/test/java/io/strimzi/operator/cluster/model/KafkaBridgeClusterTest.java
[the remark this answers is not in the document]
ppatierno: The bridge is different because based on the `bridge.tracing` property (that operator sets through `STRIMZI_TRACING` env var), it's the bridge code itself setting the interceptor properties with corresponding classes. So nothing we can check at this stage of the test. I guess this is the reason why it wasn't there before even with just OpenTracing supported.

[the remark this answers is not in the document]
scholzj: Ok, in that case this is probably fine.

scholzj (approved):

ppatierno:
@see-quick @im-konge anything more to add?

im-konge (approved):
LGTM. Thanks :)

see-quick (approved):

PR #7379, opened 2022-09-22
Set the OwnerReference in Pods managed by StrimziPodSets as controllers

scholzj:
### Type of change

- Enhancement / new feature

### Description

The StrimziPodSet controller currently sets the StrimziPodSet as the owner of the Pod resources. But it does not configure itself as controller there. The issue with that is that `kubectl drain` does not like it when draining the nodes:

```
kubectl drain ip-10-0-142-210.ec2.internal --delete-emptydir-data --ignore-daemonsets --timeout=6000s
node/ip-10-0-142-210.ec2.internal cordoned
error: unable to drain node "ip-10-0-142-210.ec2.internal" due to error:cannot delete Pods declare no controller (use --force to override): myproject/my-cluster-kafka-0, myproject/my-cluster-zookeeper-2, continuing command...
There are pending nodes to be drained:
 ip-10-0-142-210.ec2.internal
cannot delete Pods declare no controller (use --force to override): myproject/my-cluster-kafka-0, myproject/my-cluster-zookeeper-2
```

This is not a major issue -> it can be overridden by adding the `--force` flag. But it should not be needed, since the pods have controller. This PR changes the owner reference used for the Pods managed by the StrimziPodSets to mark itself as controller. That allows the node draining even without the `--force` flag. That should make it easier to drain nodes through various autoscalers which might not use the `--force` flag or for the Strimzi Pods to coexist with other applications using bare pods (so that the `--force` flag needed for Strimzi does not have to kick out other pods where it might not be desired).

### Checklist

- [x] Write tests

### DIFF
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/model/ModelUtils.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/model/ModelUtils.java
@@ -675,19 +675,19 @@ public class ModelUtils {
      * Creates the OwnerReference based on the resource passed as parameter
      *
      * @param owner         The resource which should be the owner
-     * @param controller    Indicates whether the owner acts also as the controller. This value is used in the
+     * @param isController  Indicates whether the owner acts also as the controller. This value is used in the
      *                      controller flag which is part of the OwnerReference object.
      *
      * @return          The new OwnerReference
      */
-    public static OwnerReference createOwnerReference(HasMetadata owner, boolean controller)   {
+    public static OwnerReference createOwnerReference(HasMetadata owner, boolean isController)   {
         return new OwnerReferenceBuilder()
                 .withApiVersion(owner.getApiVersion())
                 .withKind(owner.getKind())
                 .withName(owner.getMetadata().getName())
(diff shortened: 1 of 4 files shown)


cluster-operator/src/main/java/io/strimzi/operator/cluster/model/ModelUtils.java
```diff
@@ -674,18 +674,20 @@ public static void parseMetrics(AbstractModel model, HasConfigurableMetrics reso
...
      *
-     * @param owner     The resource which should be the owner
+     * @param owner         The resource which should be the owner
+     * @param controller    Indicates whether the owner acts also as the controller. This value is used in the
+     *                      controller flag which is part of the OwnerReference object.
      *
      * @return          The new OwnerReference
      */
-    public static OwnerReference createOwnerReference(HasMetadata owner)   {
+    public static OwnerReference createOwnerReference(HasMetadata owner, boolean controller)   {
```
ppatierno, line 683: maybe `isController`?

Frawless (approved):

PR #7382, opened 2022-09-23
docs(security context): Adding a pod security label

PaulRMellor:

**Documentation**

Adds an optional step to add the `pod-security` label to a namespace as an added check when using the `restricted` security context.
Plus some minor edits from peer review

### Checklist

- [ ] Write tests
- [x] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/documentation/assemblies/configuring/assembly-security-providers.adoc
+++ b/documentation/assemblies/configuring/assembly-security-providers.adoc
@@ -13,7 +13,7 @@ For example, permissions can control runtime operations or access to resources.
 ifdef::Section[]
 //how security context and security providers work
 include::../../modules/configuring/con-config-security-providers.adoc[leveloffset=+1]
-//implementing your own provider
+//enabling the provider
 include::../../modules/configuring/proc-config-restricted-security-provider.adoc[leveloffset=+1]
 //implementing your own provider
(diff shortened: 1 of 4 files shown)


see-quick (approved):
nice job, thanks Paul! :)

scholzj (commented):
TBH, I do not think we should document this. This is purely Kubernetes thing, nothing to do with Strimzi. And this might not work that way long term or accross all distributions.

PaulRMellor:
> TBH, I do not think we should document this. This is purely Kubernetes thing, nothing to do with Strimzi. And this might not work that way long term or accross all distributions.

H'm. I liked the idea of providing this check. @see-quick Jakub's not convinced.

see-quick:
> > TBH, I do not think we should document this. This is purely Kubernetes thing, nothing to do with Strimzi. And this might not work that way long term or accross all distributions.
> 
> H'm. I liked the idea of providing this check. @see-quick Jakub's not convinced.

Well, I see Jakub's point on this. At first it seems great but if you look from the maintaining perspective it could lead to a scenario, where something in Kubernetes changed, and we could potentially miss it. Therefore I think we should let users find such info in up-to-date Kubernetes DOCs, not ours.

scholzj (approved):

PR #7390, opened 2022-09-24
Split the cluster roles to allow cluster-wide operators with only limited access to some namespaces

### LINKED ISSUE #7378: Restrict cluster operator priviledges when in cluster wide mode
When the cluster operator is installed in cluster wide mode (watching all namespaces), it has to be granted some cluster roles through cluster role bindings. In particular, the cluste role strimzi-cluster-operator-namespaced. The problem is that this cluster role contains quite sensitive k8s resources like secrets and rolebindings. As a consequence, the cluster operator service account is granted rights on secrets in ALL namespaces of the k8s cluster, which is a security concern.

We would like the service account to have rights at cluster scope only for necessary resources (principle of least priviledge). To do so, we propose to have strimzi-cluster-operator-namespaced split in 2 distinct cluster roles. One containing non sensitive resources that the cluster operator requires at cluster scope. Another one with sensitive resources that the cluster operator requires only for namespace hosting a strimzi resource.

The first cluster role would be bound with a cluster role binding once for all while the second with role bindings only in namespaces hosting strimzi resources

scholzj:
### Type of change

- Enhancement / new feature

### Description

Currently, Strimzi can be installed in two different modes:
* **Watching one or more individual namespaces.** In this case, most of the RBAC resources need to be set-up only for the namespaces the operator watches. Only the RBAC rights for the cluster-wide resources such as reading the nodes or storage classes or managing cluster role bindings need to be _cluster-wide_. This works fine in general. But given how the Kubernetes API is designed, the operator here deployes a separate _verticle_ per namespaces with its own watches etc. Due to this, this mode can consume a lot of resources when you want too many namespaces and does not scale well.
* **Watching the whole cluster.** In this mode, the operator is watching the whole cluster and has access to the whole cluster. All RBAC resources are set-up as cluster wide. This mode works well and does not have any scalability issues related to the number of namespaces in the cluster. This is because it uses only a single operator _verticle_ and only one set watches for the whole cluster. However, in this mode, the operator has also access tot he whole cluster and in theory, it can for example read the secrets in all namespaces. This is something users sometimes wnat to avoid for security reasons.

This PR tries to make it easier for users to install the operator as cluster-wide but limit the RBAC rights the operator has. It splits the existing role `strimzi-cluster-operator-namespaced` into too roles:
* The original `strimzi-cluster-operator-namespaced`
* The new `strimzi-cluster-operator-watched`

The original role now covers the RBAC rights which are needed only in the namespaces where some custom resource is created. The new role contains the RBAC rights which are required as cluster wide for the watches and informers and also to do some basic management of the custom resources (get them, update their status etc.).

Thanks to this split, the users can do the following:
* Deploy the CO as cluster-wide
* Bind the new role `strimzi-cluster-operator-watched` as cluster-wide
* Bind the `strimzi-cluster-operator-namespaced` role with role-binding onl for the namespaces, where the user plans to have some operands.

This adds additional complexities for users who want to take this effort. They will be resonsible for managing the RBAC rights and creating the required `RoleBindings` for the `strimzi-cluster-operator-namespaced` role. If the user fails to do it, the operator will run into RBAC errors (which will be reflected in the state of the CR in the namespace with missing RBAC). But at the same time, the operator will have less access rights to completely unrelated namespaces, so it might be more secure.

This option does not replace the regular cluster-wide installation. It will be only an _expert_ option for those who want to use this. There will be no special tolling provided for such installation as it is expected the users using this will deal with this in their own way. The split of the cluster roles just makes it much easier for them to install the operator this way.

This should fix #7378 

### Checklist

- [x] Update documentation
- [x] Update CHANGELOG.md

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,9 +7,15 @@
 * The `ControlPlaneListener` feature gate moves to GA
 * Add client rack-awareness support to Strimzi Bridge pods
 
-### Deprecations and removals
+### Changes, deprecations and removals
 
 * A connector or task failing triggers a 'NotReady' condition to be added to the KafkaConnector CR status. This is different from previous versions where the CR would report 'Ready' even if the connector or a task had failed.
+* The `ClusterRole` from file `020-ClusterRole-strimzi-cluster-operator-role.yaml` was split into two separate roles:
+  * The original `strimzi-cluster-operator-namespaced` `ClusterRole` in the file `020-ClusterRole-strimzi-cluster-operator-role.yaml` contains the rights related to the resources created based on some Strimzi custom resources.
+  * The new `strimzi-cluster-operator-watched` `ClusterRole` in the file `023-ClusterRole-strimzi-cluster-operator-role.yaml` contains the rights required to watch and manage the Strimzi custom resources.
+  
+  When deploying the Strimzi Cluster Operator as cluster-wide, the `strimzi-cluster-operator-watched` `ClusterRole` needs to be always granted at the cluster level.
+  But the `strimzi-cluster-operator-namespaced` `ClusterRole` might be granted only for the namespaces where any custom resources are created.
 * The `ControlPlaneListener` feature gate moves to GA. 
   Direct upgrade from Strimzi 0.22 or earlier is not possible anymore.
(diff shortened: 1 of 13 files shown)


tombentley (approved):
LGTM, thanks!

Issue #7381, opened 2022-09-23
Quickstart lists step that is not needed (and not possible)


v42net:
**Suggestion / Problem**
The Quickstart section [`Installing Strimzi`](https://strimzi.io/docs/operators/latest/quickstart.html#proc-install-product-str) step 6 instructs to edit `install/cluster-operator/060-Deployment-strimzi-cluster-operator.yaml`. This entry now uses a `valueFrom`, so I suspect that it does not need to be modified anymore.

**Documentation Link**
https://strimzi.io/docs/operators/latest/quickstart.html#proc-install-product-str

scholzj:
You mean the value for `STRIMZI_NAMESPACE`? That always used `vaueFrom` in the default installation files. But it defaults to the namespace where the operator runs while the example tells it to watch another namespace. So it needs to be modified.

v42net:
Indeed the value of `STRIMZI_NAMESPACE`. From the quickstart instructions I understood that I had to replace `value: <old-value>` by `value: <new-value>`. So as a newbe, finding a `valueFrom` entry I suspected that this instruction was outdated. Thanks for the explanation.

scholzj:
Ahh, ok. I opened https://github.com/strimzi/strimzi-kafka-operator/pull/7392 => do you think that would make it more clear for you?

PR #7392, opened 2022-09-25
Make the STRIMZI_NAMESPACE change in the Quicksrt guide a bit more clear

scholzj:
### Type of change

- Documentation

### Description

This PR tries to make the Quickstrat bit more clear based on the feedback we got in #7381 -> it lists from what value should the `STRIMZI_NAMESPACE` be changed to what value.

### Checklist

- [x] Update documentation

### DIFF
--- a/documentation/modules/quickstart/proc-installing-product.adoc
+++ b/documentation/modules/quickstart/proc-installing-product.adoc
@@ -52,7 +52,7 @@ kubectl create ns my-kafka-project
 ----
 
 . Edit the `install/cluster-operator/060-Deployment-strimzi-cluster-operator.yaml` file and set the `STRIMZI_NAMESPACE` environment variable to the namespace `my-kafka-project`.
-You can do that by changing the environment variable configuration from:
+Change the environment variable configuration from this:
 +
 [source, yaml, subs=+quotes ]
 ----
@@ -65,7 +65,7 @@ env:
 # ...
 ----
 +
-To:
+To this:
 +
 [source, yaml, subs=+quotes ]
 ----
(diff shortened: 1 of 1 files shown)


ppatierno (approved):
LGTM but tbh it was clear to me even before. You are actually already providing the section ...
```yaml
env:
- name: STRIMZI_NAMESPACE
  value: my-kafka-project
# ...
```
which shows how you should set the env var. Don't understand why we should show from where we are moving ...

scholzj:
> LGTM but tbh it was clear to me even before. You are actually already providing the section ...

Well, it was clear to me as well. But we are not the target group, so if this makes it more clear to new users, it might be worth it..

documentation/modules/quickstart/proc-installing-product.adoc
```diff
@@ -52,6 +52,20 @@ kubectl create ns my-kafka-project
 ----
 
 . Edit the `install/cluster-operator/060-Deployment-strimzi-cluster-operator.yaml` file and set the `STRIMZI_NAMESPACE` environment variable to the namespace `my-kafka-project`.
+You can do that by changing the environment variable configuration from:
```
PaulRMellor, line 55:
```suggestion
Change the environment variable configuration from this:
```

documentation/modules/quickstart/proc-installing-product.adoc
```diff
@@ -52,6 +52,20 @@ kubectl create ns my-kafka-project
...
+# ...
+env:
+- name: STRIMZI_NAMESPACE
+  valueFrom:
+    fieldRef:
+      fieldPath: metadata.namespace
+# ...
+----
++
+To:
```
PaulRMellor, line 68:
```suggestion
To this:
```

PaulRMellor (approved):

Issue #7320, opened 2022-09-13
Remove support for Ingress & Events v1beta1 after bumping up minimum Kubernetes version to 1.19


devguyio:
After changing the minimum supported Kubernetes version to `1.19` (See #7184) support for `v1beta1.Events` & `v1beta1.Ingress` is no longer needed since support for v1 of Events & Ingress are supported from Kubetnetes 1.19+. So we should be able to remove the related code from Strimzi operators.

PR #7393, opened 2022-09-26
Remove Events v1beta specific code

devguyio:
### Type of change

_Select the type of your PR_

- Refactoring

### Description

Remove the code specific to v1beta1/Event APIs and refactor existing code to use v1/Event by default. 

Part of #7320 

### Checklist

- [ ] Write tests
- [ ] ~Update documentation~
- [ ] ~Check RBAC rights for Kubernetes / OpenShift roles
- [ ] ~Update CHANGELOG.md~
- [ ] ~Supply screenshots for visual changes, such as Grafana dashboards~


devguyio:
> I guess we could also remove the `PlatformFeaturesAvailability.hasEventsApiV1()` method? It should not be used anymore after this PR.

Done

scholzj (approved):
LGTM. Thanks for the PR.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/resource/events/KubernetesRestartEventPublisher.java
```diff
@@ -37,30 +36,24 @@ public abstract class KubernetesRestartEventPublisher {
...
      */
-    public static KubernetesRestartEventPublisher createPublisher(KubernetesClient client, String operatorName, boolean hasEventsV1) {
-        Clock clock = Clock.systemDefaultZone();
-        if (hasEventsV1) {
-            return new V1RestartEventPublisher(clock, client, operatorName);
-        } else {
-            return new V1Beta1RestartEventPublisher(clock, client, operatorName);
-        }
+    public KubernetesRestartEventPublisher(KubernetesClient client, String operatorName) {
+        this(client, operatorName, Clock.systemDefaultZone());
```
ppatierno, line 50: just wondering if this is right. It's setting the clock by using the default zone of the system it's running that could be any zone around the world. At same time, we log everything using UTC. Should we use UTC clock for the events time as well?

devguyio: good catch! this was not meant to be like that.

devguyio: Ok, I understand more now, I had the impression that this is something I introduced, however after reading the code now it seems it's the way it has been so far.

I would say for this PR because it's a refactoring, let's keep it as it is, and I'd be happy to check this out in a separate PR  to be changed and tested on its own I believe.

ppatierno: Fine with me, it's important to take a note of it and having an explanation. Maybe there is a good one and I am wrong. Or maybe we should really use UTC there.

ppatierno (approved):

PR #7395, opened 2022-09-26
Remove Ingress v1beta1 specific code

devguyio:
### Type of change

- Refactoring

### Description

Remove the code specific to v1beta1/Ingress APIs and refactor existing code to use v1/Ingress by default.

Part of #7320

### Checklist

- [ ] ~Write tests~
- [ ] ~Update documentation~
- [ ] ~Check RBAC rights for Kubernetes / OpenShift roles~
- [ ] ~Update CHANGELOG.md~
- [ ] ~Supply screenshots for visual changes, such as Grafana dashboards~


devguyio:
This is to validate the ITs are running after removing the v1beta1 Ingress, will update it once I'm sure everything is passing and #7393 is merged.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaListenersReconciler.java
```diff
@@ -242,9 +235,6 @@ protected Future<Void> routes() {
      * @return  Future which completes when all ingresses are created or deleted.
      */
     protected Future<Void> ingresses() {
-        if (!pfa.hasIngressV1()) {
-            return Future.succeededFuture();
-        }
```
scholzj, line 238: Should we remove the empty line here?

cluster-operator/src/test/java/io/strimzi/operator/cluster/operator/assembly/KafkaListenerReconcilerIngressV1Beta1Test.java
```diff
@@ -227,17 +129,14 @@ public MockKafkaListenersReconciler(
...
-                IngressV1Beta1Operator ingressV1Beta1Operator) {
-            super(reconciliation, kafka, null, pfa, 300_000L, secretOperator, serviceOperator, routeOperator, ingressOperator, ingressV1Beta1Operator);
+                IngressOperator ingressOperator) {
+            super(reconciliation, kafka, null, pfa, 300_000L, secretOperator, serviceOperator, routeOperator, ingressOperator);
         }
 
         @Override
         public Future<ReconciliationResult> reconcile()  {
             return ingresses()
-                    .compose(i -> ingressesV1Beta1())
```
scholzj, line 238: I think we need to think about this test class. Its main purpose is that with Kube 1.16 it uses the v1beta1 methods and with Kube 1.22 it is using the v1 methods. It does not seem to check anything else. After removing the v1beta1 support, I don't think the v1 test makes any sense now. So maybe we should remove it completely?

scholzj (commented):
Some nits, LGTM otherwise.

scholzj (approved):
Thanks. LGTM.

ppatierno (approved):

PR #7412, opened 2022-09-30
[systemtest] Enable multiple COs in one namespace in our tests after LeaderElection feature

im-konge:

### Type of change

- "Bugfix"

### Description

After the LeaderElection feature was introduced, one of our tests - `MultipleClusterOperatorsIsolatedST#testKafkaCCAndRebalanceWithMultipleCOs` - had to be changed to run two COs in different namespaces - because of the `ClusterRole`, `Role` and `RoleBinding` which wasn't configured to allow us to edit it with different lease name.

This PR enables these resources to be configurable and changes back `MultipleClusterOperatorsIsolatedST#testKafkaCCAndRebalanceWithMultipleCOs` to run COs in one namespace (which was intention).

### Checklist

### DIFF
--- a/systemtest/src/main/java/io/strimzi/systemtest/resources/operator/SetupClusterOperator.java
+++ b/systemtest/src/main/java/io/strimzi/systemtest/resources/operator/SetupClusterOperator.java
@@ -13,7 +13,10 @@ import io.fabric8.kubernetes.api.model.ServiceAccountBuilder;
 import io.fabric8.kubernetes.api.model.apiextensions.v1.CustomResourceDefinition;
 import io.fabric8.kubernetes.api.model.rbac.ClusterRole;
 import io.fabric8.kubernetes.api.model.rbac.ClusterRoleBinding;
+import io.fabric8.kubernetes.api.model.rbac.ClusterRoleBuilder;
 import io.fabric8.kubernetes.api.model.rbac.Role;
+import io.fabric8.kubernetes.api.model.rbac.RoleBinding;
+import io.fabric8.kubernetes.api.model.rbac.RoleBindingBuilder;
 import io.fabric8.kubernetes.api.model.rbac.RoleBuilder;
 import io.strimzi.systemtest.BeforeAllOnce;
 import io.strimzi.systemtest.Constants;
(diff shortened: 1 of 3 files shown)


systemtest/src/main/java/io/strimzi/systemtest/resources/operator/SetupClusterOperator.java
```diff
@@ -462,6 +465,63 @@ public void prepareEnvForOperator(ExtensionContext extensionContext, String clie
         prepareEnvForOperator(extensionContext, clientNamespace, Collections.singletonList(clientNamespace));
     }
 
+    private String changeLeaseNameInResourceIfNeeded(String yamlPath) {
+        final EnvVar leaseEnvVar = extraEnvVars.stream().filter(envVar -> envVar.getName().equals("STRIMZI_LEADER_ELECTION_LEASE_NAME")).findFirst().orElse(null);
+        if (leaseEnvVar != null && yamlPath.contains("022-")) {
```
Frawless, line 470: Wouldn't be better to create a list of resource names and test path against it instead of this magic constant?

systemtest/src/main/java/io/strimzi/systemtest/resources/operator/SetupClusterOperator.java
```diff
@@ -462,6 +465,63 @@ public void prepareEnvForOperator(ExtensionContext extensionContext, String clie
...
+            try {
+                File tmpFile = File.createTempFile(yamlPath.replace(Constants.STRIMZI_DEPLOYMENT_NAME + ".yaml", leaseEnvVar.getValue()), "yaml");
+                String tmpFileContent = "";
+                final String resourceName = leaseEnvVar.getValue() + "-leader-election";
+                final String resourceType = yamlPath.replace(TestUtils.USER_PATH + "/../packaging/install/cluster-operator/", "").split("-")[1];
```
Frawless, line 475: don't we have `packaging/install...` in some constant?

systemtest/src/main/java/io/strimzi/systemtest/resources/operator/SetupClusterOperator.java
see-quick, line 470: I am little bit sceptical about that condition (i.e., yamlPath.contains("022-")) but something I can live with :D Maybe being explicit in this case would be better. We don't know if another file with 022- prefix will be added in the future, which could potentially break this. But it is just my though...

see-quick (approved):
Looks good just one comment 👍.

systemtest/src/main/java/io/strimzi/systemtest/resources/operator/SetupClusterOperator.java
im-konge: Yeah I know, as @Frawless mentioned I'll create a list of resource with proper test path rather than this :) good point!

systemtest/src/main/java/io/strimzi/systemtest/resources/operator/SetupClusterOperator.java
im-konge: Nope, I search through the module and I didn't find anything
We are usually working with `TestUtils.USER_PATH + "...path"`

I think I can add it to `Constants` class and changed it everywhere, but I think that's something for different PR

Frawless (approved):

PR #7421, opened 2022-10-04
ZookeeperRoller considers unready pods

### LINKED ISSUE #1001: ZooKeeper rolling update should deal with unready pods
Further to my comments on #969 I think there's an extra problem with how we're handling zookeeper rolling, based on what I've observed looking at Strimzi upgrading.

Even if the readiness is based on `stat` and looking for `leader`, `follower` (or `standalone`) in the output we can still break the cluster. The problem is that if there is a problem in the node that was restarted we wait for it to become read, but it doesn't and eventually the wait times out. On the next reconciliation we decide not to roll that pod based on the metadata and proceed to delete the next pod in the SS. If that fails to come back up the same timeout happens there too. Each reconciliation takes out another ZK pod until the whole cluster is down. At that point the Kafka brokers without any functioning zookeeper and the whole cluster grinds to a halt.

robobario:
### Type of change

- Bugfix

### Description

This fix makes the ZookeeperRoller:
1. re-order the pods so that non-ready pods are evaluated first
2. wait for pods that do not require a restart to become ready

This means that in the a scenario with a pod stuck in a non-ready state it will evaluate the non-ready pod first on each run and await readiness. Likely ending in an exception.

Why:
Previously a bad pattern could occur where ZookeeperRoller restarted a pod and it did not become ready. This eventually hit a timeout that failed the roll but on the next periodic reconciliation that pod would not be eligible for a restart and the roller would then restart the next node. This would eventually take the whole Zookeeper cluster offline if, for example, the pods couldn't be scheduled.

Note:
There are further specialisations in the KafkaRoller that we could add/share with the ZookeeperRoller, like a short-circuit for pods that are stuck in pending due to them being unschedulable.

Fixes #1001

### Checklist

- [x] Write tests
- [x] Update CHANGELOG.md

### DIFF
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/resource/ZooKeeperRoller.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/resource/ZooKeeperRoller.java
@@ -89,12 +89,12 @@ public class ZooKeeperRoller {
                             for (ZookeeperPodContext podContext : clusterRollContext.getPodContextsWithNonReadyFirst())  {
                                 final String podName = podContext.getPodName();
                                 if (podContext.requiresRestart() && !podName.equals(leader)) {
-                                        LOGGER.debugCr(reconciliation, "Restarting non-leader pod {}", podName);
-                                        // roll the pod and wait until it is ready
-                                        // this prevents rolling into faulty state (note: this applies just for ZK pods)
-                                        fut = fut.compose(ignore -> restartPod(reconciliation, podName, podContext.reasonsToRestart));
+                                    LOGGER.debugCr(reconciliation, "Restarting non-leader pod {}", podName);
+                                    // roll the pod and wait until it is ready
+                                    // this prevents rolling into faulty state (note: this applies just for ZK pods)
+                                    fut = fut.compose(ignore -> restartPod(reconciliation, podName, podContext.reasonsToRestart));
                                 } else {
-                                    if(podContext.requiresRestart()) {
+                                    if (podContext.requiresRestart()) {
                                         LOGGER.debugCr(reconciliation, "Deferring restart of leader {}", podName);
                                     }
                                     LOGGER.debugCr(reconciliation, "Pod {} does not need to be restarted", podName);
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/resource/ZooKeeperRoller.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/resource/ZooKeeperRoller.java
@@ -88,16 +88,16 @@ public class ZooKeeperRoller {
                             // Then roll each non-leader pod => the leader is rolled last
                             for (ZookeeperPodContext podContext : clusterRollContext.getPodContextsWithNonReadyFirst())  {
                                 if (podContext.requiresRestart() && !podContext.getPodName().equals(leader)) {
-                                    LOGGER.debugCr(reconciliation, "Restarting non-leader pod {}", podContext.getPodName());
+                                    LOGGER.debugCr(reconciliation, "Pod {} needs to be restarted", podContext.getPodName());
                                     // roll the pod and wait until it is ready
                                     // this prevents rolling into faulty state (note: this applies just for ZK pods)
                                     fut = fut.compose(ignore -> restartPod(reconciliation, podContext.getPodName(), podContext.reasonsToRestart));
                                 } else {
                                     if (podContext.requiresRestart()) {
                                         LOGGER.debugCr(reconciliation, "Deferring restart of leader {}", podContext.getPodName());
+                                    } else {
+                                        LOGGER.debugCr(reconciliation, "Pod {} does not need to be restarted", podContext.getPodName());
                                     }
-                                    LOGGER.debugCr(reconciliation, "Pod {} does not need to be restarted", podContext.getPodName());
-                                    LOGGER.debugCr(reconciliation, "Waiting for non-restarted pod {} to become ready", podContext.getPodName());
                                     fut = fut.compose(ignore -> podOperator.readiness(reconciliation, reconciliation.namespace(), podContext.getPodName(), READINESS_POLLING_INTERVAL_MS, operationTimeoutMs));
                                 }
                             }
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/resource/ZooKeeperRoller.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/resource/ZooKeeperRoller.java
@@ -13,16 +13,20 @@ import io.strimzi.operator.common.operator.resource.PodOperator;
 import io.vertx.core.Future;
 import io.vertx.core.Promise;
 
-import java.util.HashMap;
+import java.util.ArrayList;
 import java.util.List;
-import java.util.Map;
+import java.util.Set;
 import java.util.function.Function;
+import java.util.stream.Collectors;
 
 /**
  * ZooKeeperRoller helps to roll ZooKeeper cluster. It uses the ZooKeeperLeaderFinder to find the leader which is
- * rolled first.
+ * rolled last.
  */
 public class ZooKeeperRoller {
+
+    private static final long READINESS_POLLING_INTERVAL_MS = 1_000;
+
     private static final ReconciliationLogger LOGGER = ReconciliationLogger.create(ZooKeeperRoller.class.getName());
 
     private final PodOperator podOperator;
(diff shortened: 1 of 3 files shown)


scholzj (approved):
LGTM, thanks for the PR. I think that there is one more possible hole in it when the Pod does not exist at all. But I think we should proceed with this PR and if needed, we should get back to it in the future.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/resource/ZooKeeperRoller.java
```diff
@@ -128,14 +135,76 @@ public Future<Void> maybeRollingUpdate(Reconciliation reconciliation, Labels sel
...
                 .compose(pod -> podOperator.restart(reconciliation, pod, operationTimeoutMs))
                 .compose(ignore -> {
                     LOGGER.debugCr(reconciliation, "Waiting for readiness of pod {}", podName);
-                    return podOperator.readiness(reconciliation, reconciliation.namespace(), podName, pollingIntervalMs, operationTimeoutMs);
+                    return podOperator.readiness(reconciliation, reconciliation.namespace(), podName, READINESS_POLLING_INTERVAL_MS, operationTimeoutMs);
                 });
     }
+
+
+    private static class ZookeeperClusterRollContext {
```
SamBarker, line 148: I'm not seeing anything `Zookeeper` specific about the context classes, I wonder if it would be better to remove the prefix and inflate them to top level classes?

cluster-operator/src/test/java/io/strimzi/operator/cluster/operator/resource/ZooKeeperRollerTest.java
```diff
@@ -78,6 +80,63 @@ public void testAllPodsAreRolled(VertxTestContext context)  {
...
+        ZookeeperLeaderFinder leaderFinder = mock(ZookeeperLeaderFinder.class);
+        when(leaderFinder.findZookeeperLeader(any(), any(), any(), any())).thenReturn(Future.succeededFuture("my-cluster-zookeeper-2"));
+
+        MockZooKeeperRoller roller = new MockZooKeeperRoller(podOperator, leaderFinder, 300_00L);
+
+        Checkpoint async = context.checkpoint();
+        roller.maybeRollingUpdate(Reconciliation.DUMMY_RECONCILIATION, DUMMY_SELECTOR, shouldRoll, new Secret(), new Secret())
+              .onComplete(context.succeeding(v -> context.verify(() -> {
+                  assertThat(roller.podRestarts.size(), is(3));
+                  assertThat(roller.podRestarts.pollFirst(), is("my-cluster-zookeeper-1"));
```
SamBarker, line 103: Feels like this is quite a subtle to assert the order.

cluster-operator/src/test/java/io/strimzi/operator/cluster/operator/resource/ZooKeeperRollerTest.java
```diff
@@ -78,6 +80,63 @@ public void testAllPodsAreRolled(VertxTestContext context)  {
...
+
+        ZookeeperLeaderFinder leaderFinder = mock(ZookeeperLeaderFinder.class);
+        when(leaderFinder.findZookeeperLeader(any(), any(), any(), any())).thenReturn(Future.succeededFuture("my-cluster-zookeeper-2"));
+
+        MockZooKeeperRoller roller = new MockZooKeeperRoller(podOperator, leaderFinder, 300_00L);
+
+        Checkpoint async = context.checkpoint();
+        roller.maybeRollingUpdate(Reconciliation.DUMMY_RECONCILIATION, DUMMY_SELECTOR, shouldRoll, new Secret(), new Secret())
+              .onComplete(context.failing(v -> context.verify(() -> {
+                  assertThat(roller.podRestarts.size(), is(0));
```
SamBarker, line 135: Size 0 is a bit of a surprise, I was expecting `my-cluster-zookeeper-0` to be restartable. Is it zero because `my-cluster-zookeeper-1` is false and thus rolling any nodes would break the cluster quorum?

cluster-operator/src/test/java/io/strimzi/operator/cluster/operator/resource/ZooKeeperRollerTest.java
```diff
@@ -78,6 +80,63 @@ public void testAllPodsAreRolled(VertxTestContext context)  {
...
+        when(podOperator.isReady(any(), eq("my-cluster-zookeeper-1"))).thenReturn(false);
+        when(podOperator.isReady(any(), eq("my-cluster-zookeeper-2"))).thenReturn(true);
+        when(podOperator.listAsync(any(), any(Labels.class))).thenReturn(Future.succeededFuture(PODS));
+
+        ZookeeperLeaderFinder leaderFinder = mock(ZookeeperLeaderFinder.class);
+        when(leaderFinder.findZookeeperLeader(any(), any(), any(), any())).thenReturn(Future.succeededFuture("my-cluster-zookeeper-2"));
+
+        MockZooKeeperRoller roller = new MockZooKeeperRoller(podOperator, leaderFinder, 300_00L);
+
+        Checkpoint async = context.checkpoint();
```
SamBarker, line 99: Do we need the `checkpoint` or can we just use `context.complete()` on line 107?

cluster-operator/src/test/java/io/strimzi/operator/cluster/operator/resource/ZooKeeperRollerTest.java
```diff
@@ -78,6 +80,63 @@ public void testAllPodsAreRolled(VertxTestContext context)  {
...
+        when(podOperator.listAsync(any(), any(Labels.class))).thenReturn(Future.succeededFuture(PODS));
+        when(podOperator.readiness(any(), any(), any(), anyLong(), anyLong())).thenReturn(Future.failedFuture("failure"));
+
+
+        ZookeeperLeaderFinder leaderFinder = mock(ZookeeperLeaderFinder.class);
+        when(leaderFinder.findZookeeperLeader(any(), any(), any(), any())).thenReturn(Future.succeededFuture("my-cluster-zookeeper-2"));
+
+        MockZooKeeperRoller roller = new MockZooKeeperRoller(podOperator, leaderFinder, 300_00L);
+
+        Checkpoint async = context.checkpoint();
```
SamBarker, line 132: Do we need the `checkpoint` or can we just use `context.complete` on line 136?

cluster-operator/src/test/java/io/strimzi/operator/cluster/operator/resource/ZooKeeperRollerTest.java
robobario: Correct, `my-cluster-zookeeper-1` is non-ready, so it is sorted to the front when we are iterating over the pods. We then await readiness on it, which is mocked to fail. This failure stops the chain and nothing is restarted.

Previously it calculated that the non-ready pods had no restart reasons so they were skipped over and the next eligible pod would be restarted. This can eventually take 100% of the ZK pods offline.

cluster-operator/src/test/java/io/strimzi/operator/cluster/operator/resource/ZooKeeperRollerTest.java
robobario: Pushed up a refactor to use hamcrest `contains` which gives you a failure like:
```
Expected: iterable containing ["my-cluster-zookeeper-0", "my-cluster-zookeeper-1", "my-cluster-zookeeper-2"]
     but: item 0: was "my-cluster-zookeeper-1"
```
If they're out of order

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/resource/ZooKeeperRoller.java
robobario: Removing the prefix sounds good, but I'd hesitate on pulling them up given it's a first cut. Best to keep them trapped in here at first. I don't want to help the `List<String>` of restart reasons to proliferate. The `KafkaRoller` is working with the `RestartReasons` class which would be better.

cluster-operator/src/test/java/io/strimzi/operator/cluster/operator/resource/ZooKeeperRollerTest.java
robobario: I'll also note down that maybe there is another hole/race, if the leader is somehow non-ready we could restart one of the non-leader nodes because we always restart the leader last.

We could change the initial loop to also await readiness of the leader, or put it off to another PR since it's an existing problem.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/resource/ZooKeeperRoller.java
SamBarker: Ok Sounds reasonable

scholzj: I think there is no real value from separating them right now. So I would keep it inside this class.

cluster-operator/src/test/java/io/strimzi/operator/cluster/operator/resource/ZooKeeperRollerTest.java
robobario: Replaced the checkpoints in the test class with `context.completeNow()`

cluster-operator/src/test/java/io/strimzi/operator/cluster/operator/resource/ZooKeeperRollerTest.java
robobario: Replaced the checkpoints in the test class with context.completeNow()

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/resource/ZooKeeperRoller.java
```diff
@@ -53,60 +57,61 @@ public Future<Void> maybeRollingUpdate(Reconciliation reconciliation, Labels sel
...
-                                if (podEntry.getValue() != null && !podEntry.getValue().isEmpty()) {
-                                    if (!podEntry.getKey().equals(leader)) {
-                                        LOGGER.debugCr(reconciliation, "Restarting non-leader pod {}", podEntry.getKey());
-                                        // roll the pod and wait until it is ready
-                                        // this prevents rolling into faulty state (note: this applies just for ZK pods)
-                                        fut = fut.compose(ignore -> restartPod(reconciliation, podEntry.getKey(), podEntry.getValue()));
-                                    } else {
-                                        LOGGER.debugCr(reconciliation, "Deferring restart of leader {}", podEntry.getKey());
+                            for (ZookeeperPodContext podContext : clusterRollContext.getPodContextsWithNonReadyFirst())  {
+                                final String podName = podContext.getPodName();
```
scholzj, line 90: Why do we need this variable? Its just a getter which should be easy to call directly, or?

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/resource/ZooKeeperRoller.java
```diff
@@ -128,14 +133,76 @@ public Future<Void> maybeRollingUpdate(Reconciliation reconciliation, Labels sel
...
+
+        private ZookeeperPodContext(final Pod pod, final List<String> reasonsToRestart, final boolean ready) {
+            this.pod = pod;
+            this.ready = ready;
+            if (reasonsToRestart != null) {
+                this.reasonsToRestart.addAll(reasonsToRestart);
+            }
+        }
+
+
```
scholzj, line 197: Do we need the two empty lines?

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/resource/ZooKeeperRoller.java
```diff
@@ -128,14 +133,76 @@ public Future<Void> maybeRollingUpdate(Reconciliation reconciliation, Labels sel
...
+        public boolean requiresRestart() {
+            return podContexts.stream().anyMatch(ZookeeperPodContext::requiresRestart);
+        }
+
+
+        public Set<String> podNames() {
+            return podContexts.stream().map(ZookeeperPodContext::getPodName).collect(Collectors.toSet());
+        }
+
+
```
scholzj, line 174: Do we need the two empty lines?

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/resource/ZooKeeperRoller.java
```diff
@@ -128,14 +133,76 @@ public Future<Void> maybeRollingUpdate(Reconciliation reconciliation, Labels sel
...
+        public void add(final ZookeeperPodContext podContext) {
+            podContexts.add(podContext);
+        }
+
+
+        public boolean requiresRestart() {
+            return podContexts.stream().anyMatch(ZookeeperPodContext::requiresRestart);
+        }
+
+
```
scholzj, line 169: Do we need the two empty lines?

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/resource/ZooKeeperRoller.java
```diff
@@ -128,14 +133,76 @@ public Future<Void> maybeRollingUpdate(Reconciliation reconciliation, Labels sel
...
         return podOperator.getAsync(reconciliation.namespace(), podName)
                 .compose(pod -> podOperator.restart(reconciliation, pod, operationTimeoutMs))
                 .compose(ignore -> {
                     LOGGER.debugCr(reconciliation, "Waiting for readiness of pod {}", podName);
-                    return podOperator.readiness(reconciliation, reconciliation.namespace(), podName, pollingIntervalMs, operationTimeoutMs);
+                    return podOperator.readiness(reconciliation, reconciliation.namespace(), podName, READINESS_POLLING_INTERVAL_MS, operationTimeoutMs);
                 });
     }
+
+
```
scholzj, line 145: Do we need the two empty lines everwhere? 😂

cluster-operator/src/test/java/io/strimzi/operator/cluster/operator/resource/ZooKeeperRollerTest.java
robobario: I changed it to also await readiness of the leader when doing the first iteration of the pods.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/resource/ZooKeeperRoller.java
```diff
@@ -53,60 +57,61 @@ public Future<Void> maybeRollingUpdate(Reconciliation reconciliation, Labels sel
...
+                                    LOGGER.debugCr(reconciliation, "Restarting non-leader pod {}", podName);
+                                    // roll the pod and wait until it is ready
+                                    // this prevents rolling into faulty state (note: this applies just for ZK pods)
+                                    fut = fut.compose(ignore -> restartPod(reconciliation, podName, podContext.reasonsToRestart));
+                                } else {
+                                    if (podContext.requiresRestart()) {
+                                        LOGGER.debugCr(reconciliation, "Deferring restart of leader {}", podName);
                                     }
+                                    LOGGER.debugCr(reconciliation, "Pod {} does not need to be restarted", podName);
+                                    LOGGER.debugCr(reconciliation, "Waiting for non-restarted pod {} to become ready", podName);
```
scholzj, line 101: Did you checked the log messages in a real pod? Doesn't the `readiness` method cover the logging?

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/resource/ZooKeeperRoller.java
robobario: It had 6 usages in the block, so it was a reflex to pull it out. Happy to inline.

scholzj: It does not do any complex computation. So I would just call the method inline.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/resource/ZooKeeperRoller.java
robobario: This bit of logging was following the KafkaRoller logging, I figured it would be good if the logs were similar. 

Looking at it again the ZookeeperRoller logging is misleading anyway like it logs about `Restarting` while it's looping and constructing the composed future, it's not actually restarting pods.

For example in the test where no pods are restarted you see:

```
Pod my-cluster-zookeeper-0 should be rolled due to [Should restart]
Pod my-cluster-zookeeper-1 does not need to be rolled
Pod my-cluster-zookeeper-2 should be rolled due to [Should restart]
Zookeeper leader is pod my-cluster-zookeeper-2
Pod my-cluster-zookeeper-1 does not need to be restarted
 Waiting for non-restarted pod my-cluster-zookeeper-1 to become ready
Restarting non-leader pod my-cluster-zookeeper-0
Deferring restart of leader my-cluster-zookeeper-2
Pod my-cluster-zookeeper-2 does not need to be restarted
 Waiting for non-restarted pod my-cluster-zookeeper-2 to become ready
 ```

Should we move the logging into the futures?

Comment (scholzj) [reply cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/resource/ZooKeeperRoller.java:101]: Yeah, but the tests don't call the real methods, right? You use mocks. The `readiness` method (and the methods it calls) IMHO logs as well on a debug level. So I wonder if this really adds something new.

Comment (robobario) [reply cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/resource/ZooKeeperRoller.java:101]: Yeah it doesn't add anything, I'll remove that log. 

But what about the ZookeeperRoller logging in general? It's writing out misleading logs like in the above: `Restarting non-leader pod my-cluster-zookeeper-0` is logged while building up the composed future but we never restart due to an earlier failure in the future chain.

File: cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/resource/ZooKeeperRoller.java (hunk 10 above)
Comment (robobario) [reply cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/resource/ZooKeeperRoller.java:145]: No :( codestyle from my last job sorry, will fix.

File: cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/resource/ZooKeeperRoller.java (hunk 11 above)
Comment (scholzj) [reply cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/resource/ZooKeeperRoller.java:101]: Hmm, I never noticed that. But I guess you can fix that if you want.

Comment (robobario) [reply cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/resource/ZooKeeperRoller.java:101]: Since it's not new and shuffling the logs around into futures will complicate this PR more I'd put it off to another PR. Maybe a quick fix for now would be to change the messages a little to indicate it's a planned action.

Comment (robobario) [reply cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/resource/ZooKeeperRoller.java]: Actually, we could just delete it since the restart future info logs:
```
        LOGGER.infoCr(reconciliation, "Rolling pod {} due to {}", podName, reasons);
```

scholzj: Well, some of them such as the readiness seems new. But in general it is fine to leave the old issues for another PR of course.

robobario: Okay pushed up logging changes. Now on the first iteration of the pods you will get one of:

- Pod x needs to be restarted
- Deferring restart of leader x
- Pod x does not need to be restarted

logged for each pod.

scholzj: Sounds good. Thanks.

SamBarker (approved):
lgtm

ppatierno (approved):
LGTM. Thanks!

scholzj:
Thanks for the Pr @robobario

robobario:
Thanks for the reviews :bow:

PR #7424, opened 2022-10-04
Replace sensitive information mask with "omitted"

fvaleri:
Using * chars for sensible information mask (passwords and keys from report.sh) breaks YAML parsers. The * char is used as part of the anchors and aliases feature in YAML. The parser raises and error when it finds it and it is not expected. We can use quotes so that the parser sees the mask as astring, or use a different replacement as proposed here.

### DIFF
--- a/tools/report.sh
+++ b/tools/report.sh
@@ -16,7 +16,7 @@ SECRETS_OPT="hidden"
 # sed non-printable text delimiter
 SD=$(echo -en "\001") && readonly SD
 # sed sensitive information filter expression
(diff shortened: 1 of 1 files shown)


scholzj (approved):
I think we use `[hidden]` normally in te container shell scripts for example. But I can for sure live with `omitted` as well. Thanks for the PR.

fvaleri:
Let's keep it consistent and use [hidden]. Thanks.

ppatierno (approved):

PR #7425, opened 2022-10-04
Add optional --out-dir parameter to the report script

fvaleri:
This is useful if you don't have enough /tmp space and you want to use a different dir. If you don't specify the out dir, then the current behavior is maintained.

### DIFF
--- a/tools/report.sh
+++ b/tools/report.sh
@@ -14,7 +14,7 @@ MM2=""
 KUBECTL_INSTALLED=false
 OC_INSTALLED=false
 KUBE_CLIENT="kubectl"
-OUT_DIR="$(mktemp -d)"
+OUT_DIR=""
 SECRETS_OPT="hidden"
 
 # sed non-printable text delimiter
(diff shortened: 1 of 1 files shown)


tools/report.sh
```diff
@@ -8,9 +8,13 @@ fi
 { # this ensures the entire script is downloaded #
 NAMESPACE=""
 CLUSTER=""
+BRIDGE=""
+CONNECT=""
+MM2=""
 KUBECTL_INSTALLED=false
 OC_INSTALLED=false
 KUBE_CLIENT="kubectl"
+OUT_DIR="$(mktemp -d)"
```
scholzj, line 17: Should this be called only when `--out-dir` was not specified?

fvaleri: This is a default value, which is overwritten when you specify `--out-dir`.

scholzj: Right. But you call `mktemp -d` then even if you end up not needing it.

fvaleri: Agree. We want to avoid creating folders when they are not used. Let me fix that.

scholzj (approved):
Thanks.

ppatierno (approved):

PR #7438, opened 2022-10-06
docs(downgrade): Note on OLM and downgrade constraints

PaulRMellor:

**Documentation**

Adds a note to [Downgrading Strimzi](https://strimzi.io/docs/operators/in-development/deploying.html#assembly-downgrade-str) to clarify that the downgrade steps use the install artifacts. And downgrading using the OLM is not possible. 

### Checklist

- [ ] Write tests
- [x] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/documentation/assemblies/upgrading/assembly-downgrade.adoc
+++ b/documentation/assemblies/upgrading/assembly-downgrade.adoc
@@ -9,7 +9,7 @@
 If you are encountering issues with the version of Strimzi you upgraded to,
 you can revert your installation to the previous version.
 
-If you used the installation artifacts to install Strimzi, you can use the same artifacts to perform the following downgrade procedures:
+If you used the YAML installation files to install Strimzi, you can use the YAML installation files from the previous release to perform the following downgrade procedures:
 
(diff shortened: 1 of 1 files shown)


documentation/assemblies/upgrading/assembly-downgrade.adoc
```diff
@@ -5,20 +5,22 @@
 [id='assembly-downgrade-{context}']
 = Downgrading Strimzi
 
+[role="_abstract"]
 If you are encountering issues with the version of Strimzi you upgraded to,
 you can revert your installation to the previous version.
 
-You can perform a downgrade to:
+If you used the installation artifacts to install Strimzi, you can use the same artifacts to perform the following downgrade procedures:
```
scholzj, line 12: I think this is a bit cryptic. I assume the _installation artifacts_ here mean the YAML installation files? Also, you cannot really use the same for the downgrade. You need to use the same type of installation but from the release you downgrade to. Maybe something like this makes it more clear?

```suggestion
If you used the YAML installation files to install Strimzi, you can use the YAML installation files from the previous release to perform the following downgrade procedures:
```

PR #7444, opened 2022-10-07
docs(mm2): Updates to concepts from review

PaulRMellor:

**Documentation**

Updates MM2 concepts in the docs from review by @mimaison 

### Checklist

- [ ] Write tests
- [x] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/documentation/modules/configuring/con-config-mirrormaker2-replication.adoc
+++ b/documentation/modules/configuring/con-config-mirrormaker2-replication.adoc
@@ -15,11 +15,15 @@ MirrorMaker 2.0 uses:
 
 MirrorMaker 2.0 is based on the Kafka Connect framework, _connectors_ managing the transfer of data between clusters.
 
+You configure a `KafkaMirrorMaker2` custom resource to define the Kafka Connect deployment, including the connection details of the source and target clusters, and then run a set of MirrorMaker 2.0 connectors to make the connection.
+
 MirrorMaker 2.0 consists of the following connectors:
 
-`MirrorSourceConnector`:: The source connector replicates topics from a source cluster to a target cluster. It also replicates ACLs and is necessary for the `MirrorCheckpointConnector` to run.
+`MirrorSourceConnector`:: The source connector replicates topics from a source cluster to a target cluster. It also replicates ACLs and is necessary for the `MirrorCheckpointConnector` to run. 
 `MirrorCheckpointConnector`:: The checkpoint connector periodically tracks offsets. If enabled, it also synchronizes consumer group offsets between the source and target cluster.
 `MirrorHeartbeatConnector`:: The heartbeat connector periodically checks connectivity between the source and target cluster.
+
+NOTE: If you are using the User Operator to manage ACLs, ACL replication through the connector is not possible.    
 
 The process of _mirroring_ data from a source cluster to a target cluster is asynchronous.
 Each MirrorMaker 2.0 instance mirrors data from one source cluster to one target cluster. 
@@ -31,11 +35,11 @@ image::mirrormaker.png[MirrorMaker 2.0 replication]
 By default, a check for new topics in the source cluster is made every 10 minutes.
 You can change the frequency by adding `refresh.topics.interval.seconds` to the source connector configuration.
 
-= Cluster configuration
+== Cluster configuration
 
 You can use MirrorMaker 2.0 in _active/passive_ or _active/active_ cluster configurations.
 
-active/active cluster configuration:: An active/active configuration has two active clusters replicating data bidirectionally. Applications can use either cluster. Each cluster can provide the same data. In this way,  you can make the same data available in different geographical locations.
+active/active cluster configuration:: An active/active configuration has two active clusters replicating data bidirectionally. Applications can use either cluster. Each cluster can provide the same data. In this way,  you can make the same data available in different geographical locations. As consumer groups are active in both clusters, consumer offsets for replicated topics are not synchronized back to the source cluster. 
 active/passive cluster configuration:: An active/passive configuration has an active cluster replicating data to a passive cluster. The passive cluster remains on standby. You might use the passive cluster for data recovery in the event of system failure.
 
 The expectation is that producers and consumers connect to active clusters only.
@@ -76,6 +80,8 @@ MirrorMaker 2.0 monitors the source topics.
 MirrorMaker 2.0 detects and propagates changes to the source topics to the remote topics.
 Changes might include automatically creating missing topics and partitions.
 
+NOTE: In most cases you write to local topics and read from remote topics. Though write operations are not prevented on remote topics, they should be avoided. 
+
 == Offset tracking
 MirrorMaker 2.0 tracks offsets for consumer groups using internal topics.
 
@@ -83,8 +89,6 @@ MirrorMaker 2.0 tracks offsets for consumer groups using internal topics.
 `checkpoints` topic:: The `checkpoints` topic maps the last committed offset in the source and target cluster for replicated topic partitions in each consumer group.
 
 As they used internally by MirrorMaker 2.0, you do not interact directly with these topics. 
-
-TIP: You can use Kafka's `RemoteClusterUtils.java` utility in your client application to retrieve the remote offsets for a consumer group from the `checkpoints` topic.
 
 `MirrorCheckpointConnector` emits _checkpoints_ for offset tracking.
 Offsets for the `checkpoints` topic are tracked at predetermined intervals through configuration.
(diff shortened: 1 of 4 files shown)


documentation/modules/configuring/con-config-mirrormaker2-replication.adoc
```diff
@@ -72,19 +70,21 @@ With this configuration applied, topics retain their original names.
 
 == Topic configuration synchronization
 
-Topic configuration is automatically synchronized between source and target clusters.
-By synchronizing configuration properties, the need for rebalancing is reduced.
-
-== Data integrity
-
-MirrorMaker 2.0 monitors source topics and propagates any configuration changes to remote topics, checking for and creating missing partitions.
-Only MirrorMaker 2.0 can write to remote topics.
```
PaulRMellor, line 81: @mimaison - I removed this rather than it possibly confusing matters. I don't know if tehre's anything we should add about constraints on writing to topics within the MM2 framework?

documentation/modules/configuring/con-config-mirrormaker2-replication.adoc
```diff
@@ -14,33 +15,30 @@ MirrorMaker 2.0 uses:
 
 MirrorMaker 2.0 is based on the Kafka Connect framework, _connectors_ managing the transfer of data between clusters.
 
-MirrorMaker 2.0 uses the following connectors:
+MirrorMaker 2.0 consists of the following connectors:
 
-`MirrorSourceConnector`:: The source connector replicates topics from a source cluster to a target cluster.
+`MirrorSourceConnector`:: The source connector replicates topics from a source cluster to a target cluster. It also replicates ACLs and is necessary for the `MirrorCheckpointConnector` to run.
```
scholzj, line 20: Should we add a note that the ACL replication does not work with User Operator enabled?

documentation/modules/configuring/con-config-mirrormaker2-replication.adoc
```diff
@@ -14,33 +15,30 @@ MirrorMaker 2.0 uses:
...
 You can change the frequency by adding `refresh.topics.interval.seconds` to the source connector configuration.
-However, increasing the frequency of the operation might affect overall performance.
 
 = Cluster configuration
 
 You can use MirrorMaker 2.0 in _active/passive_ or _active/active_ cluster configurations.
 
-* In an _active/active_ configuration, both clusters are active and provide the same data simultaneously, which is useful if you want to make the same data available locally in different geographical locations.
-* In an _active/passive_ configuration, the data from an active cluster is replicated in a passive cluster, which remains on standby, for example, for data recovery in the event of system failure.
+active/active cluster configuration:: An active/active configuration has two active clusters replicating data bidirectionally. Applications can use either cluster. Each cluster can provide the same data. In this way,  you can make the same data available in different geographical locations.
```
scholzj, line 38: The confusing thing for many users here is how the offset configuration (does not) works in active-active clusters. I.e. that consumer offsets for the mirrored topics are nit synchronized back tot he original source cluster. It might be worth adding it here.

documentation/modules/configuring/con-config-mirrormaker2-replication.adoc
```diff
@@ -72,19 +70,21 @@ With this configuration applied, topics retain their original names.
...
 MirrorMaker 2.0 tracks offsets for consumer groups using internal topics.
 
-* The `offset-syncs` topic maps the source and target offsets for replicated topic partitions from record metadata
-* The `checkpoints` topic maps the last committed offset in the source and target cluster for replicated topic partitions in each consumer group
+`offset-syncs` topic:: The `offset-syncs` topic maps the source and target offsets for replicated topic partitions from record metadata.
+`checkpoints` topic:: The `checkpoints` topic maps the last committed offset in the source and target cluster for replicated topic partitions in each consumer group.
+
+As they used internally by MirrorMaker 2.0, you do not interact directly with these topics. 
+
+TIP: You can use Kafka's `RemoteClusterUtils.java` utility in your client application to retrieve the remote offsets for a consumer group from the `checkpoints` topic.
```
scholzj, line 87: As this is not needed anymore, is calling it a `TIP` appropriatte? I would call it different and explain more when it is useful.

documentation/modules/overview/con-configuration-points-mirrormaker.adoc
```diff
@@ -36,10 +37,10 @@ Topic replication is defined using regular expression patterns to include or exc
 
 The following MirrorMaker 2.0 connectors and related internal topics help manage the transfer and synchronization of data between the clusters.
 
-MirrorSourceConnector:: A _MirrorSourceConnector_ creates remote topics from the source cluster.
+MirrorSourceConnector:: A _MirrorSourceConnector_ creates remote topics from the source cluster. It also replicates ACLs and is necessary for the `MirrorCheckpointConnector` to run.
```
scholzj, line 40: This seems to mirror the content from previous module. Same comment about ACL applies.

documentation/modules/overview/con-configuration-points-mirrormaker.adoc
```diff
@@ -24,8 +25,8 @@ MirrorMaker 2.0 uses:
 
 You can use MirrorMaker 2.0 in _active/passive_ or _active/active_ cluster configurations.
 
-* In an _active/active_ configuration, both clusters are active and provide the same data simultaneously, which is useful if you want to make the same data available locally in different geographical locations.
-* In an _active/passive_ configuration, the data from an active cluster is replicated in a passive cluster, which remains on standby, for example, for data recovery in the event of system failure.
+active/active cluster configuration:: An active/active configuration has two active clusters replicating data bidirectionally. Applications can use either cluster. Each cluster can provide the same data. In this way,  you can make the same data available in different geographical locations.
```
scholzj, line 28: This seems to mirror the previous module. Is that intentional? In any case, the same about the offset synchronization applies as I commented abotu above.

documentation/modules/configuring/con-config-mirrormaker2-replication.adoc
mimaison: Yes that makes sense.

mimaison: MM2 does not prevent any actions on remote topics but in many cases users should avoid producing to remote topics.

documentation/modules/configuring/con-config-mirrormaker2-replication.adoc
mimaison: Is there a technical reason or is this potentially a feature that could be added?

documentation/modules/configuring/con-config-mirrormaker2-replication.adoc
mimaison: I think it's still useful. The sync feature is not enabled by default and `RemoteClusterUtils` has a couple of other features.

documentation/modules/configuring/con-config-mirrormaker2-replication.adoc
scholzj: The User Operator has the KafkaUser CRs as the only source of truth. So if it is enabled, only two things will happen:
* It will rewrite the ACLs synced by MM2 with what is in the KafkaUser CRs
* The `KafkaUsers` will have the exact same ACLs configured, in which case they will be set by the User Operator and the ACL mirroring is not needed.

documentation/modules/configuring/con-config-mirrormaker2-replication.adoc
scholzj: From my experience ... I think they usually understand that they should not produce into them. But what they usually don't understand is that when the consumed from them, the consumed offsets are not synced back into the original cluster.

documentation/modules/configuring/con-config-mirrormaker2-replication.adoc
scholzj: I'm not saying it is not useful. But I think most users would prefer to use the automatic sync which is also more versatile since it does not rely on Java. So the context and when to use it should be better explained.

documentation/modules/configuring/con-config-mirrormaker2-replication.adoc
mimaison: I see, thanks for the explanation!

documentation/modules/configuring/con-config-mirrormaker2-replication.adoc
PaulRMellor: Added this: _As consumer groups are active in both clusters, consumer offsets for replicated topics are not synchronized back to the source cluster. _

documentation/modules/configuring/con-config-mirrormaker2-replication.adoc
PaulRMellor: I've changed to _NOTE_ and moved under Synchronizing consumer group offsets. Not sure if we want to say more about where it's useful (or how it might be)

> NOTE: If you have an application written in Java, you can use the `RemoteClusterUtils.java` utility to synchronize offsets from the application. The utility fetches remote offsets for a consumer group from the `checkpoints` topic.

documentation/modules/overview/con-configuration-points-mirrormaker.adoc
PaulRMellor: Yes. This is in the _Overview_ guide. I'm going to split the replicated content and reuse it...

documentation/modules/overview/con-configuration-points-mirrormaker.adoc
PaulRMellor: Same comment as previous.

scholzj (approved):

mimaison (approved):

PR #5204, opened 2021-06-27
Remove warnings from test cases touched in PR #5174

LiamClarkeNZ:
### Type of change

- Refactoring

### Description

Remove warnings from test files touched in PR #5174. 

### Checklist

_Please go through this checklist and make sure all applicable tasks have been done_

- [ ] Write tests
- [X] Make sure all tests pass
- [ ] Update documentation
- [ ] Check RBAC rights for Kubernetes / OpenShift roles
- [ ] Try your changes from Pod inside your Kubernetes and OpenShift cluster, not just locally
- [ ] Reference relevant issue(s) and close them after merging
- [ ] Update CHANGELOG.md
- [ ] Supply screenshots for visual changes, such as Grafana dashboards

PR #7451, opened 2022-10-10
Upgrade vertx to 4.3.4

### LINKED ISSUE #6741: Re-enable `SessionStartupDoesNotBlockMainThreadTest` tests after Vert.x 4.3.0
The tests in `SessionStartupDoesNotBlockMainThreadTest` use detection of the main thread being blocked. Right now, they use system properties to configure `vertx.logger-delegate-factory-class-name` and override the logger to detect it. This seems to be flaky and failing in some environments (e.g. on MaxOS with OPenJDK, when running the tests from CLI with `mvn install`). They were therefore disabled for the time being.

Vert.x 4.3.0 brinds a new possibility to configure `blockedThreadChecker` (see eclipse-vertx/vert.x#4283). We should move to that once we upgrade to Vert.x 4.3.0 and re-enable the tests.

robobario:
### Type of change

- Enhancement / new feature

### Description

* Upgrade to vertx-4.3.4
* Refactor to keep references to shared `WorkerExecutor`'s in fields so the pool isn't automatically closed. There was a behaviour change in vertx 4.3.0. SharedWorkerPool now gets shutdown if all it's WorkerExecutors are finalised or closed. See:
  - https://github.com/eclipse-vertx/vert.x/issues/4345
  - https://github.com/eclipse-vertx/vert.x/pull/4352
* Refactor `SessionStartupDoesNotBlockMainThreadTest` to use a custom blocked thread handler instead of using logging to sniff blocking behaviour and re-enable the test

Fixes #6741

### Checklist

- [x] Update CHANGELOG.md


cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/CaReconciler.java
```diff
@@ -116,6 +118,7 @@ public CaReconciler(
     ) {
         this.reconciliation = reconciliation;
         this.vertx = vertx;
+        this.workerExecutor = vertx.createSharedWorkerExecutor("kubernetes-ops-pool");
```
scholzj, line 121: I don't get this. Why is this needed? The reference is kept in `ClusterOperator` class. This should not create a new executor, it just gets an existing one. The same should apply to most of the other production classes apart from `ClusterOperator` and `UserOperator`

robobario: This way the unit tests didn't need to change, like `KubernetesRestartEventsMockTest` has direct calls to the reconciler. I'm trying to avoid the failures occuring again in future because class X is used in a new unit test or context.

robobario: Oh and `ZkImpl` uses a different pool, so that one needs a field.

robobario: Alternatively we could make all the vertx tests use some common code to create and hold a worker executor?

cluster-operator/src/test/java/io/strimzi/operator/cluster/operator/resource/KafkaRollerTest.java
```diff
@@ -158,7 +158,7 @@ public void pod1NotReadyAfterRolling(VertxTestContext testContext) throws Interr
         clearRestarted();
         doFailingRollingRestart(testContext, kafkaRoller,
                 asList(2, 3, 4),
-                KafkaRoller.FatalProblem.class, "Error while waiting for non-restarted pod c-kafka-1 to become ready",
+                KafkaRoller.FatalProblem.class, "Error while waiting for restarted pod c-kafka-1 to become ready",
```
robobario, line 161: I changed two assertions in this test because the vertx-junit5 upgrade revealed that the tests were false positives. Debugging them on vertx-4.2.4 showed that the assertions inside `testContext.failing` were ignored.

The test expected them to be non-restarted but the roller decides they need reconfiguration due to logging config difference.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/CaReconciler.java
scholzj: My personal preference would be to keep the production code clean of this and do it in the rest code. Not sure a common code can be used there, since each test will handle it differently.

robobario: Changed it to create one field per operator, one field in `ZkImpl` and then do the rest in the test classes.

scholzj (approved):
LGTM. Thanks for implementing the changes. I restarted the CI which failed, but I wasn§t sure it was related to this PR. I will also run the STs.

showuon (approved):

cluster-operator/src/test/java/io/strimzi/operator/cluster/operator/assembly/CertificateRenewalTest.java
```diff
@@ -102,6 +109,18 @@ public class CertificateRenewalTest {
 
     private List<Secret> secrets = new ArrayList<>();
 
+    @BeforeAll
+    public static void before() {
+        vertx = Vertx.vertx();
```
katheris, line 114: I think by putting this in BeforeAll rather than BeforeEach it effectively changes the scope of the vertx object. My understanding is that if you pass the vertx object as an argument to each test you will get a new instance for each test, whereas here the result would be one instance shared by all tests. Perhaps not a problem but worth making sure we are happy with that.

cluster-operator/src/test/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceStateMachineTest.java
```diff
@@ -319,8 +319,9 @@ private void krNewBadGoalsError(Vertx vertx, VertxTestContext context, CruiseCon
                         } else {
                             context.failNow(new RuntimeException("This operation failed with an unexpected error:" + result.cause().getMessage()));
                         }
+                    } else {
```
katheris, line 322: There's an even neater way to do this, since we have the test context you can have:

```
.onComplete(context.failing(throwable -> {
    if (throwable.getMessage().contains("java.lang.IllegalArgumentException: Missing hard goals")) {
        context.completeNow();
    } else {
        context.failNow(new RuntimeException("This operation failed with an unexpected error:" + throwable.getMessage()));
    });
```

Then if the call didn't fail the test vertx will fail the call automatically.

cluster-operator/src/test/java/io/strimzi/operator/cluster/operator/resource/KafkaRollerTest.java
katheris: This test class looks like it has similar logic to the one in Topic Operator where functions inside the test function are passed the test context and create checkpoints and call context.CompleteNow. That's dangerous because it's likely that the test is completed when the first call to doFailingRollingRestart is called and it never gets to the second call. This is the PR where I made changes to the similar topic tests: https://github.com/strimzi/strimzi-kafka-operator/pull/7338

katheris (commented):
Generally look good, just one comment on the changes in terms of impact. For the KafkaRoller tests I wonder if a wider refactor is needed separate from this PR.

cluster-operator/src/test/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceStateMachineTest.java
robobario: Cool, pushed up the simplification. I also set the throwable as the cause of the new RuntimeException in case of null message.

cluster-operator/src/test/java/io/strimzi/operator/cluster/operator/assembly/CertificateRenewalTest.java
robobario: Correct, I checked and it's constructing a new instance per test.

I don't think it's a problem, in that the tests are passing. It was changed from static to using the extension to inject [as part of a broad test tidy up PR](https://github.com/strimzi/strimzi-kafka-operator/pull/5204/files#diff-42e49bd8eda8159c300cf6bab463e00bdd7f8c5ad67c2e6643fde3b0fdddf16e). I couldn't see a discussion on it so at a guess @LiamClarkeNZ was removing some code that could be handled for us by the framework. Do you remember at all @LiamClarkeNZ?

Maybe we could come back around with another PR to pick a style (static vs injected-per-test) and potentially pull out a base test class.

LiamClarkeNZ: @robobario @katheris As this is running under the `VertxExtension`, and there are no test specific config changes being made in `beforeAll()`, manually instantiating a `Vertx()` isn't necessary. 

Just add a `Vertx` argument to test methods that need to interact with a Vertx instance, and the extension will inject it automatically.

topic-operator/src/test/java/io/strimzi/operator/topic/SessionStartupDoesNotBlockMainThreadTest.java
```diff
@@ -179,8 +163,12 @@ void ensureSlowK8sWatcherCreationDoesNotBlockThreads() throws ExecutionException
...
-        BlockedThreadWarnings warnings = BlockedThreadWarnings.getInstance();
-        assertTrue(warnings.isPickedUpByVertx(), "Sanity check that Vert.x was using my gammy logger for the BlockedThread Checker");
-        assertEquals(0, warnings.blockedThreadWarningsCount(), "If the BlockedThreadChecker logged anything at WARN, then start-up is still blocking");
+        assertEquals(0, warnings.get(), "If the BlockedThreadChecker logged anything at WARN, then start-up is still blocking");
+    }
+
+    private AtomicInteger installBlockedThreadHandler() {
+        AtomicInteger warnings = new AtomicInteger();
+        ((VertxInternal) vertx).blockedThreadChecker().setThreadBlockedHandler(event -> warnings.incrementAndGet());
+        return warnings;
```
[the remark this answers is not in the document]
LiamClarkeNZ: This upgrade & refactoring makes me very happy 😊

cluster-operator/src/test/java/io/strimzi/operator/cluster/operator/assembly/CertificateRenewalTest.java
robobario: I rolled it back to use injection, we still need to keep an executor reference per test so the field is created in a BeforeEach and closed in an AfterEach

katheris (approved):
LGTM

CHANGELOG.md
```diff
@@ -14,6 +14,7 @@
 * Updated JMX Exporter dependency to 0.17.2
 * ZookeeperRoller considers unready pods
 * Support multiple operations per ACLRule
+* Upgrade vertx to 4.3.4
```
ppatierno, line 17: Just minor thing ... `Vert.x`

ppatierno (approved):
LGTM. Just a minor thing in the changelog.

scholzj:
Thanks for the PR @robobario

PR #7455, opened 2022-10-10
Fix the `RoleBinding` creation in the Quickstart guide

scholzj:
### Type of change

- Bugfix

### Description

In #7390 I missed to update the Quickstart docs which now doesn't work because the new Role Binding is not created. Luckily it was noticed by @devguyio. This PR should fix it.


ppatierno (approved):

devguyio (approved):
thanks! lgtm

PR #7460, opened 2022-10-11
docs(build fix): Format fix for annotations used to highlight and describe deprecated properties

PaulRMellor:

**Documentation**

The [AclRule schema properties](https://strimzi.io/docs/operators/in-development/configuring.html#type-AclRule-schema-reference) and [AclRuleTopicResource schema reference](https://strimzi.io/docs/operators/in-development/configuring.html#type-AclRuleTopicResource-reference) are not presenting correctly in the api reference.

This PR introduces a format fix so that the sections are generated correctly. 

### Checklist

- [ ] Write tests
- [ ] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/api/src/main/java/io/strimzi/api/kafka/model/AclRule.java
+++ b/api/src/main/java/io/strimzi/api/kafka/model/AclRule.java
@@ -84,7 +84,7 @@ public class AclRule implements UnknownPropertyPreserving, Serializable {
         this.host = host;
     }
 
-    @Description("**The `operation` property has been deprecated, and should now be configured using `spec.authorization.acls[*].operations`.** Operation which will be allowed or denied. " +
+    @Description("Operation which will be allowed or denied. " +
             "Supported operations are: Read, Write, Create, Delete, Alter, Describe, ClusterAction, AlterConfigs, DescribeConfigs, IdempotentWrite and All.")
     @DeprecatedProperty(movedToPath = "spec.authorization.acls[*].operations")
     @Deprecated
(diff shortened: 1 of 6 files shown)


scholzj (commented):
The change is fine. But it looks like the `documentation/modules/appendix_crds.adoc` file is not properly generated. So I think you will need to regenerate it.

PaulRMellor:
@scholzj  - I've tried to rebuild, but I get the old and updated line when I do. 

> |operation   1.2+<.<a|*The `operation` property has been deprecated, and should now be configured using `spec.authorization.acls[*].operations`.* **The `operation` property has been deprecated, and should now be configured using `spec.authorization.acls[*].

I tried changing the Java description to something different, but still the old line is reinserted. Even removed the `. m2`. Any ideas?

scholzj:
Well, the first part is generated by the annotations. So that will be always there. The second part is what you added whcih should not be there. That is driven by the annotation.

PaulRMellor:
> Well, the first part is generated by the annotations. So that will be always there. The second part is what you added whcih should not be there. That is driven by the annotation.

Ah, right. Thanks!

scholzj:
@PaulRMellor I guess this would be the place to fix it if there should be `**` instead of `*`: https://github.com/strimzi/strimzi-kafka-operator/blob/f6255b4394afeceacbd5e9bce2779e4bb8c1badf/crd-generator/src/main/java/io/strimzi/crdgenerator/DocGenerator.java#L295-L310

scholzj:
You would need to change it on line 296 and 304 I guess.

scholzj (approved):
Maybe you can rename the PR to better match what it does now? It might help in the future when we need to dig through some history.

PR #7461, opened 2022-10-11
docs(producer config): Updates the section on data durability

PaulRMellor:

**Documentation**

Updates the [Data durability](https://strimzi.io/docs/operators/in-development/configuring.html#data_durability) section for data durability. Makes the acknowledgment of all in-sync replicas for `acks=all` clearer.

### Checklist

- [ ] Write tests
- [x] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/documentation/modules/managing/con-producer-config-properties.adoc
+++ b/documentation/modules/managing/con-producer-config-properties.adoc
@@ -55,11 +55,11 @@ acks=all <1>
 # ...
 ----
 
-<1> `acks=all` forces a partition leader to replicate messages to a certain number of followers before
+<1> `acks=all` forces a leader replica to replicate messages to a certain number of followers before
 acknowledging that the message request was successfully received. 
 
 The `acks=all` setting offers the strongest guarantee of delivery, but it will increase the latency between the producer sending a message and receiving acknowledgment.
-If you don't require such strong guarantees, a setting of `acks=0` or `acks=1` provides either no delivery guarantees or only acknowledgment that the leader partition has written the record to its log.
+If you don't require such strong guarantees, a setting of `acks=0` or `acks=1` provides either no delivery guarantees or only acknowledgment that the leader replica has written the record to its log.
 
 With `acks=all`, the leader waits for all in-sync replicas to acknowledge message delivery.
 A topic's `min.insync.replicas` configuration sets the minimum required number of in-sync replica acknowledgements.
@@ -67,28 +67,33 @@ The number of acknowledgements include that of the leader and followers.
 
 A typical starting point is to use the following configuration:
 
-* Producer
+* Producer configuration:
 ** `acks=all` (default)
-* Broker/Topic:
+* Broker configuration for topic replication:
 ** `default.replication.factor=3` (default = `1`)
 ** `min.insync.replicas=2` (default = `1`)
 
-The following table describes how this configuration operates depending on the availability of replica brokers.
+When you create a topic, you can override the default replication factor.
+You can also override `min.insync.replicas` at the topic level in the topic configuration. 
 
-.Replica broker availability
+Strimzi uses this configuration in the example configuration files for multi-node deployment of Kafka. 
+
+The following table describes how this configuration operates depending on the availability of followers that replicate the leader replica.
+
+.Follower availability
 [cols="2,3,2",options="header"]
 |===
 
-|Number of replica brokers available and in-sync
+|Number of followers available and in-sync
 |Acknowledgements 
-|Producer operable?
+|Producer can send messages?
 
 |2
-|The leader waits for 2 follower replica broker acknowledgements
+|The leader waits for 2 follower acknowledgements
 |Yes
 
 |1
-|The leader waits for 1 follower replica broker acknowledgement
+|The leader waits for 1 follower acknowledgement
 |Yes
 
 |0
@@ -97,10 +102,10 @@ The following table describes how this configuration operates depending on the a
 
 |===
 
-A topic replication factor of 3 creates one leader partition and two followers.
-In this configuration, the producer can continue if a single replica broker is unavailable.
+A topic replication factor of 3 creates one leader replica and two followers.
+In this configuration, the producer can continue if a single follower is unavailable.
 Some delay can occur whilst a failed broker is removed from the in-sync replicas or a new leader is created.
-If a second replica broker becomes unavailable, the producer won’t receive acknowledgments and won’t be able to produce more messages.
+If a second follower becomes unavailable, the producer won’t receive acknowledgments and won’t be able to produce more messages.
 
 NOTE: If the system fails, there is a risk of unsent data in the buffer being lost.
 
--- a/documentation/modules/managing/con-producer-config-properties.adoc
+++ b/documentation/modules/managing/con-producer-config-properties.adoc
@@ -104,8 +104,11 @@ The following table describes how this configuration operates depending on the a
 
 A topic replication factor of 3 creates one leader replica and two followers.
 In this configuration, the producer can continue if a single follower is unavailable.
-Some delay can occur whilst a failed broker is removed from the in-sync replicas or a new leader is created.
-If a second follower becomes unavailable, the producer won’t receive acknowledgments and won’t be able to produce more messages.
+Some delay can occur whilst removing a failed broker from the in-sync replicas or a creating a new leader.
+If the second follower is also unavailable, message delivery will not be successful.
+Instead of acknowledging successful message delivery, the leader sends an error (_not enough replicas_) to the producer.  
+The producer raises an equivalent exception.
+With `retries` configuration, the producer can resend the failed message request.
 
 NOTE: If the system fails, there is a risk of unsent data in the buffer being lost.
 
--- a/documentation/modules/managing/con-producer-config-properties.adoc
+++ b/documentation/modules/managing/con-producer-config-properties.adoc
@@ -61,37 +61,25 @@ acknowledging that the message request was successfully received.
 The `acks=all` setting offers the strongest guarantee of delivery, but it will increase the latency between the producer sending a message and receiving acknowledgment.
 If you don't require such strong guarantees, a setting of `acks=0` or `acks=1` provides either no delivery guarantees or only acknowledgment that the leader partition has written the record to its log.
 
-With `acks=all`, a leader waits for all in-sync replicas to acknowledge message delivery.
+With `acks=all`, the leader waits for all in-sync replicas to acknowledge message delivery.
 A topic's `min.insync.replicas` configuration sets the minimum required number of in-sync replica acknowledgements.
+The number of acknowledgements include that of the leader and followers. 
 
 A typical starting point is to use the following configuration:
 
 * Producer
-** `acks=all`
-* Topic:
-** `default.replication.factor=3`
-** `min.insync.replicas=2`
+** `acks=all` (default)
+* Broker/Topic:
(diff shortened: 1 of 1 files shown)


scholzj (approved):
LGTM. But I wonder if it would be good to have also someone like @tombentley or @mimaison go through this to doublecheck it is ok from Kafka perspective.

documentation/modules/managing/con-producer-config-properties.adoc
```diff
@@ -55,25 +56,59 @@ acks=all <1>
...
+A topic replication factor of 3 creates one leader partition and two followers.
+In this configuration, the producer can continue if a single replica broker is unavailable.
+Some delay can occur whilst a failed broker is removed from the in-sync replicas or a new leader is created.
+If a second replica broker becomes unavailable, the producer won’t receive acknowledgments and won’t be able to produce more messages.
+
+.Replica broker availability
+[cols="2,3,2",options="header"]
+|===
+
+|Number of replica brokers available
```
mimaison, line 94: I think you mean available and in-sync for this partition

documentation/modules/managing/con-producer-config-properties.adoc
```diff
@@ -55,25 +56,59 @@ acks=all <1>
...
-A typical starting point is to have a topic replication factor of 3, with two in-sync replicas on other brokers.
-In this configuration, the producer can continue unaffected if a single broker is unavailable.
-If a second broker becomes unavailable, the producer won’t receive acknowledgments and won’t be able to produce more messages.
+With `acks=all`, a leader waits for all in-sync replicas to acknowledge message delivery.
+A topic's `min.insync.replicas` configuration sets the minimum required number of in-sync replica acknowledgements.
+
+A typical starting point is to use the following configuration:
+
+* Producer
+** `acks=all`
```
mimaison, line 70: Since 3.0 this is now the default

documentation/modules/managing/con-producer-config-properties.adoc
```diff
@@ -55,25 +56,59 @@ acks=all <1>
...
+* Topic:
```
mimaison, line 71: Should we say broker instead of topic?

documentation/modules/managing/con-producer-config-properties.adoc
```diff
@@ -55,25 +56,59 @@ acks=all <1>
...
 
 The `acks=all` setting offers the strongest guarantee of delivery, but it will increase the latency between the producer sending a message and receiving acknowledgment.
 If you don't require such strong guarantees, a setting of `acks=0` or `acks=1` provides either no delivery guarantees or only acknowledgment that the leader partition has written the record to its log.
 
-The number of brokers which need to have appended the messages to their logs before the acknowledgment is sent to the producer is determined by the topic's `min.insync.replicas` configuration.
-A typical starting point is to have a topic replication factor of 3, with two in-sync replicas on other brokers.
-In this configuration, the producer can continue unaffected if a single broker is unavailable.
-If a second broker becomes unavailable, the producer won’t receive acknowledgments and won’t be able to produce more messages.
+With `acks=all`, a leader waits for all in-sync replicas to acknowledge message delivery.
+A topic's `min.insync.replicas` configuration sets the minimum required number of in-sync replica acknowledgements.
```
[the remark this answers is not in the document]
PaulRMellor: I think it's worth adding. Added _The number of acknowledgements include that of the leader and followers._

documentation/modules/managing/con-producer-config-properties.adoc
PaulRMellor: I've moved around to make it more obvious the example config is related to the table

documentation/modules/managing/con-producer-config-properties.adoc
PaulRMellor: We say earlier, but I thought it might be useful to show all defaults here.

> * Producer
> ** `acks=all` (default)
> * Topic:
> ** `default.replication.factor=3` (default = `1`)
> ** `min.insync.replicas=2` (default = `1`)

documentation/modules/managing/con-producer-config-properties.adoc
PaulRMellor: Maybe. What about Broker/Topic?

documentation/modules/managing/con-producer-config-properties.adoc
```diff
@@ -55,24 +56,51 @@ acks=all <1>
...
+** `default.replication.factor=3` (default = `1`)
+** `min.insync.replicas=2` (default = `1`)
+
+The following table describes how this configuration operates depending on the availability of replica brokers.
+
+.Replica broker availability
+[cols="2,3,2",options="header"]
+|===
+
+|Number of replica brokers available and in-sync
```
scholzj, line 82: What does `replica brokers` mean? I do not think that is an recognized term. The same applies to the above.

documentation/modules/managing/con-producer-config-properties.adoc
```diff
@@ -55,24 +56,51 @@ acks=all <1>
...
+|Acknowledgements 
+|Producer operable?
```
scholzj, line 84: As a non-native speaker, `operable` sounds really weird.

documentation/modules/managing/con-producer-config-properties.adoc
```diff
@@ -55,24 +56,51 @@ acks=all <1>
...
-# ...
-----
-<1> Use `2` in-sync replicas. The default is `1`.
+A typical starting point is to use the following configuration:
+
+* Producer
+** `acks=all` (default)
+* Broker/Topic:
+** `default.replication.factor=3` (default = `1`)
+** `min.insync.replicas=2` (default = `1`)
```
scholzj, line 74: I wonder if the `default = 1` would be a bit confusing. Most of our examples actually set the default to 2 and 3 already. E.g. in https://github.com/strimzi/strimzi-kafka-operator/blob/main/examples/kafka/kafka-persistent.yaml#L22-L23. It also doesn't really cover the `Broker/Topic` relationship which IMHO needs more explanation if you want to cover it both.

documentation/modules/managing/con-producer-config-properties.adoc
```diff
@@ -55,24 +56,51 @@ acks=all <1>
...
+|The leader waits for 1 follower replica broker acknowledgement
+|Yes
+
+|0
+|The leader raises an exception
+|No
+
+|===
+
+A topic replication factor of 3 creates one leader partition and two followers.
```
scholzj, line 100:
```suggestion
A topic replication factor of 3 creates one leader replica and two followers.
```

documentation/modules/managing/con-producer-config-properties.adoc
```diff
@@ -55,24 +56,51 @@ acks=all <1>
...
+In this configuration, the producer can continue if a single replica broker is unavailable.
+Some delay can occur whilst a failed broker is removed from the in-sync replicas or a new leader is created.
+If a second replica broker becomes unavailable, the producer won’t receive acknowledgments and won’t be able to produce more messages.
```
scholzj, line 103: Again, `replica broker` makes no sense. `broker replica` would be somewhat understandable, but I expect it would mean something completely different (ie. number of brokers, not topic replicas).

documentation/modules/managing/con-producer-config-properties.adoc
PaulRMellor: I've changed there to _follower_ and lead into the table with: _The following table describes how this configuration operates depending on the availability of followers that replicate the leader partition._

documentation/modules/managing/con-producer-config-properties.adoc
PaulRMellor: I think it's okay to have them, to show that they are not default.
But I've also added this after: _Strimzi uses this configuration in the example configuration files for multi-node deployment of Kafka._

documentation/modules/managing/con-producer-config-properties.adoc
PaulRMellor: Changes all instances of this

documentation/modules/managing/con-producer-config-properties.adoc
PaulRMellor: These are all followers now

scholzj (approved):

documentation/modules/managing/con-producer-config-properties.adoc
```diff
@@ -44,35 +44,68 @@ Compression is useful for improving throughput and reducing the load on storage,
...
+|0
+|The leader raises an exception
+|No
+
+|===
+
+A topic replication factor of 3 creates one leader replica and two followers.
+In this configuration, the producer can continue if a single follower is unavailable.
+Some delay can occur whilst a failed broker is removed from the in-sync replicas or a new leader is created.
+If a second follower becomes unavailable, the producer won’t receive acknowledgments and won’t be able to produce more messages.
```
tyamashi-oss, line 108: I don't think it is a wrong, but I had the impression as if there was no response from the leader. If a second replica broker becomes unavailable, the producer won’t receive acknowledgments **`to be considered successful`** from the leader. 

FYI, the leader will respond error code for NotEnoughReplicas or NotEnoughReplicasAfterAppend. And the producer with 'acks=all' will raise either NotEnoughReplicas or NotEnoughReplicasAfterAppend exception on the producer side. And those exception are retryable exceptions, so the producer can repeat retries if there is retry configuration.

> [Kafka doc min.insync.replicas] https://kafka.apache.org/documentation/
> When a producer sets acks to "all" (or "-1"), min.insync.replicas specifies the minimum number of replicas that must acknowledge a write for the write to be considered successful. If this minimum cannot be met, then the producer will raise an exception (either NotEnoughReplicas or NotEnoughReplicasAfterAppend).

documentation/modules/managing/con-producer-config-properties.adoc
```diff
@@ -55,25 +56,59 @@ acks=all <1>
...
+With `acks=all`, a leader waits for all in-sync replicas to acknowledge message delivery.
+A topic's `min.insync.replicas` configuration sets the minimum required number of in-sync replica acknowledgements.
+
+A typical starting point is to use the following configuration:
+
+* Producer
+** `acks=all`
+* Topic:
+** `default.replication.factor=3`
+** `min.insync.replicas=2`
```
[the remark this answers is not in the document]
PaulRMellor: Yes. I added this from Jakub's comments:

> When you create a topic, you can override the default replication factor.
> You can also override `min.insync.replicas` at the topic level in the topic configuration.

documentation/modules/managing/con-producer-config-properties.adoc
PaulRMellor: It might be worth saying something on the error/exception in a general way. Updated as follows:

> A topic replication factor of 3 creates one leader replica and two followers. In this configuration, the producer can continue if a single follower is unavailable. Some delay can occur whilst removing a failed broker from the in-sync replicas or a creating a new leader. If the second follower is also unavailable, message delivery will not be successful. Instead of acknowledging successful message delivery, the leader sends an error (_not enough replicas_) to the producer. The producer raises an equivalent exception. With retries configuration, the producer can resend the failed message request.

tyamashi-oss: LGTM. Thank you for updating.

tyamashi-oss (approved):

mimaison (approved):

PR #7475, opened 2022-10-13
Docs: Super users require full cert subject w/o User Operator

### LINKED ISSUE #7427: superUsers requires full certificate Subject
**Suggestion / Problem**
In the Strimzi docs, this is what is mentioned regarding specifying mTLS super user principals:
> If a user uses TLS client authentication, their username is the common name from their certificate subject prefixed with CN=

However, from testing, this is not the case.
When using the following in the Kafka CR:
```
    authorization:
      type: simple
      superUsers:
        - CN=strimzi-kafka
```
And using a AdminClient to perform operations on the cluster with a certificate with the following Subject:
`CN=strimzi-kafka,OU=foo,O=bar,L=baz,ST=qux,C=XX`
The AdminClient is denied operations.

When the CR is updated to the following:
```
    authorization:
      type: simple
      superUsers:
        - CN=strimzi-kafka,OU=foo,O=bar,L=baz,ST=qux,C=XX
```
Then the AdminClient is authorized as expected. This indicates that the documentation may be making assumptions about the Subject contained in users' certificates only containing a CN and I have checked that `ssl.principal.mapping.rules` is not used anywhere to do this either.

Thus I suggest that the documentation be updated for all versions that supported the `superUsers` field to mention that the entire Subject (Distinguished Name) of the certificate is treated as the username, as per https://docs.confluent.io/platform/current/kafka/authorization.html#tls-ssl-principal-user-names
I.e.
> By default, the name of the principal identified by a TLS/SSL certificate is the DN (X.500 Distinguished Name) of that certificate (also known as the Subject), which uses the form CN=writeuser,OU=Unknown,O=Unknown,L=Unknown,ST=Unknown,C=Unknown.

**Documentation Link**
https://strimzi.io/docs/operators/latest/configuring.html#super_users

Manicben:
Resolves #7427

### Type of change

- Documentation

### Description

The current documentation on adding super users with TLS client auth assumes the User Operator is used to provision client certs. When using custom client certs without the User Operator, the entire cert subject is required, as per https://docs.confluent.io/platform/current/kafka/authorization.html#tls-ssl-principal-user-names

### Checklist

- [ ] Write tests
- [X] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/documentation/modules/security/con-securing-kafka-authorization.adoc
+++ b/documentation/modules/security/con-securing-kafka-authorization.adoc
@@ -27,6 +27,7 @@ and are supported by all authorization mechanisms.
 
 To designate super users for a Kafka cluster, add a list of user principals to the `superUsers` property.
 If a user uses TLS client authentication, their username is the common name from their certificate subject prefixed with `CN=`.
+Otherwise, if not using the User Operator and using your own certificates for TLS client authentication, their username is their full certificate subject, which uses the form `CN=writeuser,OU=Unknown,O=Unknown,L=Unknown,ST=Unknown,C=Unknown`.
 
 .An example configuration with super users
 [source,yaml,subs="attributes+"]
@@ -45,5 +46,6 @@ spec:
         - CN=client_1
         - user_2
         - CN=client_3
+        - CN=client_4,OU=my_ou,O=my_org,L=my_location,ST=my_state,C=US
     # ...
 ----
--- a/documentation/modules/security/con-securing-kafka-authorization.adoc
+++ b/documentation/modules/security/con-securing-kafka-authorization.adoc
@@ -18,7 +18,7 @@ Supported authorization options:
 * xref:type-KafkaAuthorizationCustom-reference[Custom authorization]
 
 .Kafka cluster authorization options
-image::kafka-authorization-config-options.png[options for kafks authorization configuration]
+image::kafka-authorization-config-options.png[options for kafka authorization configuration]
 
 == Super users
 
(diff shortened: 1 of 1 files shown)


documentation/modules/security/con-securing-kafka-authorization.adoc
```diff
@@ -27,6 +27,7 @@ and are supported by all authorization mechanisms.
 
 To designate super users for a Kafka cluster, add a list of user principals to the `superUsers` property.
 If a user uses TLS client authentication, their username is the common name from their certificate subject prefixed with `CN=`.
+Otherwise, if not using the User Operator and using your own certificates for TLS client authentication, their username is their full certificate subject, which uses the form `CN=writeuser,OU=Unknown,O=Unknown,L=Unknown,ST=Unknown,C=Unknown`.
```
scholzj, line 30: Sorry, but I think this is confusing. It will look like this if you generate a certificate with all these fields set to `Unknown`. That IMHO doesn't make much sense. And it is not clear If certificate with subject `CN=writeuser,O=MyOrg` is `CN=writeuser,OU=Unknown,O=MyOrg,L=Unknown,ST=Unknown,C=Unknown` or not. I think you should use some better and more realistic example to make it more clear.

Manicben: I've updated the wording and added a couple more examples to show the case where a certificate may have less fields in the subject.
Let me know if it would be preferred to have an more realistic example in the example YAML (using fictional company, location names, etc)

scholzj (approved):
Great, thanks a lot for the PR.

ppatierno (approved):

Issue #2621, opened 2020-02-28
Automatic restarts of Kafka Connectors by the operator


iXmonvi33:
Hello,

I am using strimzi 0.16.2 and i see problem when mssql in K8S die and recreate, strimzi doesn't connect once again and it's failing (SQL connectors doesn't work)

Is there any settings to tell kafka to reconnect to database ?

scholzj:
Strimzi doesn't connect to any database. So I assume you are talking about some connector in Kafka Connect? We are not providing any connectors, so I think you need to raise it with whoever is author of the connector to find out how it is or isn't re-connecting?

Or what would you expect Strimzi to do to get the connector to reconnect?

patsadow2:
I think Strimzi should observer somehow Connectors and rerun it, when it's failed ?

scholzj:
Right now you are expected to use the REST APi to restart the connector if needed. I guess restarting it automatically makes some sense. But it has its pros and cons. For example if the operator does it, it will be restarting it every _2 minutes_ regardess of of your MSSQL state. This might not be what everyone wants.

patsadow2:
The main problem is that kafka-connector fail always after 5-10min and i have to redeploy kafka-connector for mssql manual ;/

scholzj:
So why does it fail every 5-10 minutes?

patsadow2:
I am not sure i was thought that it is some kind of connection problem but i dont know why it happens always after a few minutes, for a few minutes it is working perfectly ;/

patsadow2:
It's from kafkaconnector

![image](https://user-images.githubusercontent.com/19226186/75621929-f616c500-5b9a-11ea-8d5e-769fcbd4a370.png)

patsadow2:
Do you have any script that check kubectl describe kafkaconnector inventory-connector-mssql and then redepoly when fails ?

scholzj:
If such error happens regularly without any reason such as MSSQL issues etc., I suggest you raise it with Debezium, maybe they will have an idea. But I think we should keep this open as an issue for automatically restarting the connectors / tasks. We would have anyway discussed that, so we can use it for tracking.

I do not have any cript which would restart the connector. I have alwaysdone it nly manually using curl as a one of thing.

patsadow2:
Would be great because i cannot bypass it and i am not able to sit and check connection every 5 minutes ;d

patsadow2:
[...]

PR #7500, opened 2022-10-19
Feat/auto restart connectors

ThomasDangleterre:
### Type of change

- Enhancement / new feature

### Description

Implements [Automatically restarting FAILED connectors and tasks -  proposal 007](https://github.com/strimzi/proposals/blob/main/007-restarting-kafka-connect-connectors-and-tasks.md#automatically-restarting-failed-connectors-and-tasks)

Related issue : #2621 

### Checklist

- [x] Write tests
- [x] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/api/src/main/java/io/strimzi/api/kafka/model/AutoRestart.java
+++ b/api/src/main/java/io/strimzi/api/kafka/model/AutoRestart.java
@@ -26,11 +26,10 @@ import java.util.Map;
 @EqualsAndHashCode
 public class AutoRestart implements UnknownPropertyPreserving, Serializable {
     private static final long serialVersionUID = 1L;
-    private boolean enabled = false;
+    private boolean enabled;
     private Map<String, Object> additionalProperties = new HashMap<>(0);
 
-    @Description("Whether automatic restart for failed connectors and tasks should be enabled or disabled. Defaults to false")
-    @JsonInclude(JsonInclude.Include.NON_DEFAULT)
+    @Description("Whether automatic restart for failed connectors and tasks should be enabled or disabled")
     public boolean isEnabled() {
         return enabled;
     }
--- a/api/src/main/java/io/strimzi/api/kafka/model/AutoRestart.java
+++ b/api/src/main/java/io/strimzi/api/kafka/model/AutoRestart.java
@@ -1,0 +1,52 @@
+/*
+ * Copyright Strimzi authors.
+ * License: Apache License 2.0 (see the file LICENSE or http://apache.org/licenses/LICENSE-2.0.html).
+ */
+package io.strimzi.api.kafka.model;
+
+import com.fasterxml.jackson.annotation.JsonInclude;
+import io.strimzi.crdgenerator.annotations.Description;
+import io.strimzi.crdgenerator.annotations.DescriptionFile;
+import io.sundr.builder.annotations.Buildable;
+import lombok.EqualsAndHashCode;
+
+import java.io.Serializable;
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * Representation of the AutoRestart configuration.
+ */
+@DescriptionFile
+@Buildable(
+    editableEnabled = false,
+    builderPackage = Constants.FABRIC8_KUBERNETES_API
+)
+@JsonInclude(JsonInclude.Include.NON_NULL)
+@EqualsAndHashCode
+public class AutoRestart implements UnknownPropertyPreserving, Serializable {
+    private static final long serialVersionUID = 1L;
+    private boolean enabled = true;
+    private Map<String, Object> additionalProperties = new HashMap<>(0);
+
+
+    @Description("Enable/Disable auto restart, default to true")
+    @JsonInclude(JsonInclude.Include.NON_DEFAULT)
+    public boolean isEnabled() {
+        return enabled;
+    }
+
+    public void setEnabled(boolean enabled) {
+        this.enabled = enabled;
+    }
+
+    @Override
+    public Map<String, Object> getAdditionalProperties() {
+        return this.additionalProperties;
+    }
+
+    @Override
+    public void setAdditionalProperty(String name, Object value) {
+        this.additionalProperties.put(name, value);
+    }
+}
--- a/api/src/main/java/io/strimzi/api/kafka/model/AutoRestart.java
+++ b/api/src/main/java/io/strimzi/api/kafka/model/AutoRestart.java
@@ -26,7 +26,7 @@ import java.util.Map;
 @EqualsAndHashCode
 public class AutoRestart implements UnknownPropertyPreserving, Serializable {
     private static final long serialVersionUID = 1L;
-    private boolean enabled;
+    private boolean enabled = true;
     private Map<String, Object> additionalProperties = new HashMap<>(0);
 
     @Description("Whether automatic restart for failed connectors and tasks should be enabled or disabled")
--- a/api/src/main/java/io/strimzi/api/kafka/model/AutoRestart.java
+++ b/api/src/main/java/io/strimzi/api/kafka/model/AutoRestart.java
@@ -29,7 +29,7 @@ public class AutoRestart implements UnknownPropertyPreserving, Serializable {
     private boolean enabled = true;
     private Map<String, Object> additionalProperties = new HashMap<>(0);
 
-    @Description("Whether automatic restart should be enabled or disabled, default to true")
+    @Description("Whether automatic restart for failed connectors and tasks should be enabled or disabled. Defaults to true")
     @JsonInclude(JsonInclude.Include.NON_DEFAULT)
     public boolean isEnabled() {
         return enabled;
--- a/api/src/main/java/io/strimzi/api/kafka/model/AutoRestart.java
+++ b/api/src/main/java/io/strimzi/api/kafka/model/AutoRestart.java
@@ -29,7 +29,7 @@ public class AutoRestart implements UnknownPropertyPreserving, Serializable {
     private boolean enabled = true;
     private Map<String, Object> additionalProperties = new HashMap<>(0);
 
-    @Description("Whether automatic restart should be enable or disable, default to true")
+    @Description("Whether automatic restart should be enabled or disabled, default to true")
     @JsonInclude(JsonInclude.Include.NON_DEFAULT)
     public boolean isEnabled() {
         return enabled;
--- a/api/src/main/java/io/strimzi/api/kafka/model/AutoRestart.java
+++ b/api/src/main/java/io/strimzi/api/kafka/model/AutoRestart.java
@@ -29,8 +29,7 @@ public class AutoRestart implements UnknownPropertyPreserving, Serializable {
     private boolean enabled = true;
     private Map<String, Object> additionalProperties = new HashMap<>(0);
 
-
-    @Description("Enable/Disable auto restart, default to true")
+    @Description("Whether automatic restart should be enable or disable, default to true")
     @JsonInclude(JsonInclude.Include.NON_DEFAULT)
     public boolean isEnabled() {
         return enabled;
--- a/api/src/main/java/io/strimzi/api/kafka/model/AutoRestart.java
+++ b/api/src/main/java/io/strimzi/api/kafka/model/AutoRestart.java
@@ -26,10 +26,10 @@ import java.util.Map;
 @EqualsAndHashCode
 public class AutoRestart implements UnknownPropertyPreserving, Serializable {
     private static final long serialVersionUID = 1L;
-    private boolean enabled = true;
+    private boolean enabled = false;
     private Map<String, Object> additionalProperties = new HashMap<>(0);
 
-    @Description("Whether automatic restart for failed connectors and tasks should be enabled or disabled. Defaults to true")
+    @Description("Whether automatic restart for failed connectors and tasks should be enabled or disabled. Defaults to false")
     @JsonInclude(JsonInclude.Include.NON_DEFAULT)
     public boolean isEnabled() {
         return enabled;
--- a/api/src/main/java/io/strimzi/api/kafka/model/KafkaConnectorSpec.java
+++ b/api/src/main/java/io/strimzi/api/kafka/model/KafkaConnectorSpec.java
@@ -23,8 +23,6 @@ public class KafkaConnectorSpec extends AbstractConnectorSpec {
 
     private String className;
 
-    private AutoRestart autoRestart = new AutoRestart();
-
     @Description("The Class for the Kafka Connector")
     @JsonProperty("class")
     public String getClassName() {
@@ -35,13 +33,4 @@ public class KafkaConnectorSpec extends AbstractConnectorSpec {
         this.className = className;
     }
 
-    @Description("Automatic restart of connector and tasks configuration")
-    @JsonInclude(JsonInclude.Include.NON_EMPTY)
-    public AutoRestart getAutoRestart() {
-        return autoRestart;
-    }
-
-    public void setAutoRestart(AutoRestart autoRestart) {
-        this.autoRestart = autoRestart;
-    }
 }
--- a/api/src/main/java/io/strimzi/api/kafka/model/KafkaConnectorSpec.java
+++ b/api/src/main/java/io/strimzi/api/kafka/model/KafkaConnectorSpec.java
@@ -16,12 +16,14 @@ import lombok.EqualsAndHashCode;
         builderPackage = Constants.FABRIC8_KUBERNETES_API
 )
 @JsonInclude(JsonInclude.Include.NON_NULL)
-@JsonPropertyOrder({"class", "tasksMax", "config"})
+@JsonPropertyOrder({"class", "tasksMax", "autoRestart", "config"})
 @EqualsAndHashCode(callSuper = true)
 public class KafkaConnectorSpec extends AbstractConnectorSpec {
     private static final long serialVersionUID = 1L;
 
     private String className;
+
+    private AutoRestart autoRestart = new AutoRestart();
 
     @Description("The Class for the Kafka Connector")
     @JsonProperty("class")
@@ -31,5 +33,15 @@ public class KafkaConnectorSpec extends AbstractConnectorSpec {
 
     public void setClassName(String className) {
         this.className = className;
+    }
+
+    @Description("Whether the connector should restart when tasks fail ")
+    @JsonInclude(JsonInclude.Include.NON_EMPTY)
+    public AutoRestart getAutoRestart() {
+        return autoRestart;
+    }
+
+    public void setAutoRestart(AutoRestart autoRestart) {
+        this.autoRestart = autoRestart;
     }
 }
--- a/api/src/main/java/io/strimzi/api/kafka/model/KafkaConnectorSpec.java
+++ b/api/src/main/java/io/strimzi/api/kafka/model/KafkaConnectorSpec.java
@@ -32,5 +32,4 @@ public class KafkaConnectorSpec extends AbstractConnectorSpec {
     public void setClassName(String className) {
         this.className = className;
     }
-
 }
--- a/api/src/main/java/io/strimzi/api/kafka/model/KafkaConnectorSpec.java
+++ b/api/src/main/java/io/strimzi/api/kafka/model/KafkaConnectorSpec.java
@@ -35,7 +35,7 @@ public class KafkaConnectorSpec extends AbstractConnectorSpec {
         this.className = className;
     }
 
-    @Description("Whether the connector should restart when tasks fail ")
+    @Description("Automatic restart of connector and tasks configuration")
     @JsonInclude(JsonInclude.Include.NON_EMPTY)
     public AutoRestart getAutoRestart() {
         return autoRestart;
--- a/api/src/main/java/io/strimzi/api/kafka/model/KafkaMirrorMaker2ConnectorSpec.java
+++ b/api/src/main/java/io/strimzi/api/kafka/model/KafkaMirrorMaker2ConnectorSpec.java
@@ -6,6 +6,7 @@ package io.strimzi.api.kafka.model;
 
 import com.fasterxml.jackson.annotation.JsonInclude;
 import com.fasterxml.jackson.annotation.JsonPropertyOrder;
+import io.strimzi.crdgenerator.annotations.Description;
 import io.sundr.builder.annotations.Buildable;
 import lombok.EqualsAndHashCode;
 
@@ -14,8 +15,20 @@ import lombok.EqualsAndHashCode;
         builderPackage = Constants.FABRIC8_KUBERNETES_API
 )
 @JsonInclude(JsonInclude.Include.NON_NULL)
-@JsonPropertyOrder({"tasksMax", "config"})
+@JsonPropertyOrder({"tasksMax", "autoRestartConnectorsAndTasks", "config"})
 @EqualsAndHashCode(callSuper = true)
 public class KafkaMirrorMaker2ConnectorSpec extends AbstractConnectorSpec {
     private static final long serialVersionUID = 1L;
+
+    private AutoRestart autoRestartConnectorsAndTasks = new AutoRestart();
+
+    @Description("Whether the connector should restart when tasks fail ")
+    @JsonInclude(JsonInclude.Include.NON_EMPTY)
+    public AutoRestart getAutoRestartConnectorsAndTasks() {
+        return autoRestartConnectorsAndTasks;
+    }
+
+    public void setAutoRestartConnectorsAndTasks(AutoRestart autoRestart) {
+        this.autoRestartConnectorsAndTasks = autoRestart;
+    }
 }
--- a/api/src/main/java/io/strimzi/api/kafka/model/KafkaMirrorMaker2ConnectorSpec.java
+++ b/api/src/main/java/io/strimzi/api/kafka/model/KafkaMirrorMaker2ConnectorSpec.java
@@ -14,7 +14,7 @@ import lombok.EqualsAndHashCode;
         builderPackage = Constants.FABRIC8_KUBERNETES_API
 )
 @JsonInclude(JsonInclude.Include.NON_NULL)
-@JsonPropertyOrder({"tasksMax", "autoRestartConnectorsAndTasks", "config"})
+@JsonPropertyOrder({"tasksMax", "config"})
 @EqualsAndHashCode(callSuper = true)
 public class KafkaMirrorMaker2ConnectorSpec extends AbstractConnectorSpec {
     private static final long serialVersionUID = 1L;
--- a/api/src/main/java/io/strimzi/api/kafka/model/KafkaMirrorMaker2ConnectorSpec.java
+++ b/api/src/main/java/io/strimzi/api/kafka/model/KafkaMirrorMaker2ConnectorSpec.java
@@ -6,7 +6,6 @@ package io.strimzi.api.kafka.model;
 
 import com.fasterxml.jackson.annotation.JsonInclude;
 import com.fasterxml.jackson.annotation.JsonPropertyOrder;
-import io.strimzi.crdgenerator.annotations.Description;
 import io.sundr.builder.annotations.Buildable;
 import lombok.EqualsAndHashCode;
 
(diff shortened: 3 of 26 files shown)


api/src/main/java/io/strimzi/api/kafka/model/KafkaMirrorMaker2ConnectorSpec.java
```diff
@@ -14,8 +15,20 @@
         builderPackage = Constants.FABRIC8_KUBERNETES_API
 )
 @JsonInclude(JsonInclude.Include.NON_NULL)
-@JsonPropertyOrder({"tasksMax", "config"})
+@JsonPropertyOrder({"tasksMax", "autoRestartConnectorsAndTasks", "config"})
 @EqualsAndHashCode(callSuper = true)
 public class KafkaMirrorMaker2ConnectorSpec extends AbstractConnectorSpec {
     private static final long serialVersionUID = 1L;
+
+    private AutoRestart autoRestartConnectorsAndTasks = new AutoRestart();
```
scholzj, line 23: I think we deviate here from the proposal. The proposal suggests to have it once per `KafkaMirrorMaker2` CR - so that would be added to `KafkaMirrorMaker2Spec` class. But here you put it per connector.

In general, I do not have problem with that. But:
* It should be discussed and made clear since it deviates from the approved proposal
* If this is the way we want to go, then you can make the code simple and just have the
  ```yaml
   autoRestart:
     enabled: false
  ```
  directly in the `AbstractConnectorSpec` and use the same APi for both of them.

CC @strimzi/maintainers @ajborley

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/AbstractConnectOperator.java
```diff
@@ -824,6 +883,13 @@ protected JsonObject asJson(KafkaConnectorSpec spec, KafkaConnectorConfiguration
         return connectorConfigJson.put("connector.class", spec.getClassName());
     }
 
+    protected Future<KafkaConnectorStatus> getPreviousKafkaConnectorStatus(CrdOperator<KubernetesClient, KafkaConnector, KafkaConnectorList> resourceOperator,
```
scholzj, line 886: Didn't we got the resource with its latest status on the beginning? Can't we use that instead of querying the API server again?

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/AbstractConnectOperator.java
```diff
@@ -578,10 +582,60 @@ private Future<Void> pauseResume(Reconciliation reconciliation, String host, Kaf
...
+    private boolean shouldAutoRestart(AutoRestartStatus autoRestartStatus) {
+        if (autoRestartStatus == null) {
+            return true;
+        }
+        var count =  autoRestartStatus.getCount();
+        var minutesSinceLastRestart = StatusUtils.unitDifferenceUntilNow(StatusUtils.iso8601(autoRestartStatus.getLastRestartTimestamp()), ChronoUnit.MINUTES);
+
+        var nextRestart = count  * (count - 1);
+
+        return count <  7 &&  minutesSinceLastRestart  >= nextRestart;
```
scholzj, line 632: So, what happens after we restart it 7 times? Do we stop restarting it? I would expect it to restart it again after the longest interval (e.g. after 0, 2, 6, 12, 20, 30, 30, 30, 30 ... minutes0.)

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/AbstractConnectOperator.java
```diff
@@ -578,10 +582,60 @@ private Future<Void> pauseResume(Reconciliation reconciliation, String host, Kaf
...
+                         return Future.succeededFuture(status);
+                     }
+                 });
+        } else {
+            return Future.succeededFuture(status);
+        }
+    }
+
+    /**
+     * Connector or tasks failed should only restart at minute 0, 0, 2, 6, 12, 20 and 30
```
scholzj, line 619: Is the double 0 (`0, 0`) just a typo? Or will it actually restart for some reason twice immediately? I guess it would based on the algorithm.

documentation/api/io.strimzi.api.kafka.model.AutoRestart.adoc
```diff
@@ -0,0 +1,35 @@
...
+
+[source,yaml,subs="attributes+"]
+----
+apiVersion: strimzi.io/v1alpha1
+kind: KafkaMirrorMaker2
+metadata:
+  name: my-mm2-cluster
+spec:
+  autoRestartConnectorsAndTasks:
+    enabled: false
```
scholzj, line 34: As mentioned above, this matchs the proposal. But not your current implementation.

api/src/main/java/io/strimzi/api/kafka/model/status/KafkaMirrorMaker2Status.java
```diff
@@ -33,6 +33,8 @@ public class KafkaMirrorMaker2Status extends KafkaConnectStatus {
 
     private List<Map<String, Object>> connectors = new ArrayList<>(3);
 
+    private List<AutoRestartStatus> autoRestartStatuses = new ArrayList<>();
```
scholzj, line 36: Does this need to use some list with clear sorting? Or is it not needed in this place? (if the fields are generated in random order, they will change order and will trigger constant upgrades of the CR in Kubernetes)

scholzj (commented):
Thanks for the PR and sorry it took so long to get to it. I left some comments which we need to look into. Also, the tests failed => I restarted them in case it were just some flaky tests, but they seeemd related.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/AbstractConnectOperator.java
```diff
@@ -509,6 +511,7 @@ protected Future<ConnectorStatusAndConditions> maybeCreateOrUpdateConnector(Reco
                         .compose(conditions ->
                             apiClient.statusWithBackOff(reconciliation, new BackOff(200L, 2, 10), host, port, connectorName)
                                 .compose(createConnectorStatusAndConditions(conditions)))
+                        .compose(status -> autoRestartFailedConnectorAndTasks(reconciliation, host, apiClient, connectorName, connectorSpec, status, resource))
```
katheris, line 514: Is this the right place in the flow to restart failed connectors and tasks? Since we call createConnectorStatusAndConditions on the previous line will this mean that after we have restarted any failing connectors or tasks the connector status will still show it as failed? Do we need a new call to get the status?

documentation/api/io.strimzi.api.kafka.model.AutoRestart.adoc
```diff
@@ -0,0 +1,35 @@
+Connectors and tasks that are in the FAILED state are automatically restarted.
+
+A back-off algorithm is applied to automatic restarts of each connector and it's tasks.
+When a connector is failing an auto restart attempt will happen at first and second reconcile then attempts while be delayed
```
katheris, line 4:
```suggestion
When a connector is failing an auto restart attempt will happen at first and second reconcile then attempts will be delayed,
```

documentation/api/io.strimzi.api.kafka.model.AutoRestart.adoc
```diff
@@ -0,0 +1,35 @@
...
+incrementing the time between automatic restarts by 2 minutes up to a maximum of 30 minutes, at which point no further
+automatic restarts are attempted, and the connector or tasks are left in the FAILED state.
```
katheris, line 6: Should we add a sentence here to explain what the user should do next? Is the expectation that they then would add an annotation to request the connector be restarted? Once the connector/task has been successfully restarted will automatic restarts work again if it later fails?

documentation/api/io.strimzi.api.kafka.model.AutoRestart.adoc
```diff
@@ -0,0 +1,35 @@
...
+
+
+If you are using a kafka version <3.0.0, only the connector will be restarted not failed tasks.
```
katheris, line 9:
```suggestion
If you are using a Kafka version <3.0.0, only the connector will be restarted not failed tasks.
```

api/src/main/java/io/strimzi/api/kafka/model/status/AutoRestartStatus.java
```diff
@@ -0,0 +1,83 @@
...
+
+    private int count;
+
+    private String connectorName;
+
+    private String lastRestartTimestamp;
+    private Map<String, Object> additionalProperties;
+
+    @JsonInclude(JsonInclude.Include.NON_NULL)
+    @Description("The number of time the element restarted.")
```
katheris, line 41:
```suggestion
    @Description("The number of times the element restarted.")
```

katheris (commented):
Great to see this new feature being added, I had a few small suggested changes to wording etc. The main question I had though was around where in the reconciliation flow we are doing the auto restart and what that means for the connector status. It would be useful to have a test where the connector fails initially, is restarted successfully and we see the status is now ready.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/AbstractConnectOperator.java
ThomasDangleterre: As far as I saw, Kafka Connector statuses are recreated from scratch at each reconcile.  
We get connector and tasks statuses from Connect API before auto restarting but it is lacking the auto restart status as it is stored in the resource and the status of the resource is null when we call `maybeCreateOrUpdateConnector` function.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/AbstractConnectOperator.java
ThomasDangleterre: > between automatic restarts by 2 minutes, up to a maximum of 30 minutes, at which point no further automatic restarts are attempted, and the connector or tasks are left in the FAILED state.

I followed the proposal, should it keep restarting after 30 minutes ?

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/AbstractConnectOperator.java
ThomasDangleterre: I did it to restart at next reconciliation not matter how much time past between them, but it was not in the proposal I will revert this.

api/src/main/java/io/strimzi/api/kafka/model/status/KafkaMirrorMaker2Status.java
ThomasDangleterre: I missed that part, I will implement clear sorting.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/AbstractConnectOperator.java
ThomasDangleterre: I set the auto restart after the api status call to get failed connectors and tasks. As is, failed connectors and tasks will stay in failed state until next reconciliation even if the restart was successful. I don't know if I should add an another call to fetch connectors and tasks statuses after auto restarting or it is OK to wait until next reconciliation to have the status.

documentation/api/io.strimzi.api.kafka.model.AutoRestart.adoc
ThomasDangleterre: Yes I missed parts of what user should expect after auto restart and what should they do next. I will add them.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/AbstractConnectOperator.java
scholzj: Hmm, I missed this in the proposal. I guess it is good in that case. Although I'm not sure I understand the motivation behind it anymore.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/AbstractConnectOperator.java
scholzj: They are. But on the beginning, there is always a full `KafkaConnector` resource including the status. I'm not sure how easy or hard would it be to pass it around (currently, I think only the `.spec` part is passed around) -> maybe it would not be possible because of the duality with the Mirror Maker 2 connectors.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/AbstractConnectOperator.java
katheris: I guess the motivation would be that if the auto-restart fails 7 times we can probably assume that there is a problem with the config of the connector or task, so continuing to restart it forever will not help fix the problem.

scholzj: Ok, lets stick with it.

mimaison:
I know I come pretty late in the process but after reviewing the proposal, I've got a few questions:
1. Should we rename `autoRestartConnectorsAndTasks` to `autoRestartFailedConnectorsAndTasks` to be explicit it's not about restarting "stopped" tasks but instead only connectors and tasks in the `FAILED` state. Same for the MM2 flag.
2. Are we sure we want this enabled by default? Not entirely sure what the policy is around backwards compatibility but I see this as a significant change of behavior.

I've not checked the PR yet, I'll try to take a look tomorrow.

scholzj:
Yeah, good point. It should not be anabled by default. The API classes should not have the default object but should be null by default.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/AbstractConnectOperator.java
tombentley: Yes, I forgot about `RESTARTING`, that would be a much better fit.

documentation/api/io.strimzi.api.kafka.model.AutoRestart.adoc
```diff
@@ -0,0 +1,35 @@
...
+apiVersion: strimzi.io/v1alpha1
+kind: KafkaConnector
+metadata:
+    name: my-source-connector
+spec:
+    autoRestart:
+        enabled: false
+----
+
+Similarly, a new field in the KafkaMirrorMaker2 spec named autoRestartConnectorsAndTasks with an enabled field value of false can be applied, to disable auto-restarting connectors and tasks.
```
PaulRMellor, line 24:
```suggestion
For MirrorMaker 2.0, use the `autoRestartConnectorsAndTasks` property of the `KafkaMirrorMaker2` resource to disable automatic restarts of failed connectors and tasks.
```

documentation/api/io.strimzi.api.kafka.model.AutoRestart.adoc
```diff
@@ -0,0 +1,35 @@
...
+
+[source,yaml,subs="attributes+"]
```
PaulRMellor, line 26:
```suggestion
.Disabling automatic restarts of failed connectors for MirrorMaker 2.0
[source,yaml,subs="attributes+"]
```

api/src/main/java/io/strimzi/api/kafka/model/AutoRestart.java
```diff
@@ -0,0 +1,51 @@
...
+    builderPackage = Constants.FABRIC8_KUBERNETES_API
+)
+@JsonInclude(JsonInclude.Include.NON_NULL)
+@EqualsAndHashCode
+public class AutoRestart implements UnknownPropertyPreserving, Serializable {
+    private static final long serialVersionUID = 1L;
+    private boolean enabled = true;
+    private Map<String, Object> additionalProperties = new HashMap<>(0);
+
+    @Description("Whether automatic restart should be enabled or disabled, default to true")
```
PaulRMellor, line 32:
```suggestion
    @Description("Whether automatic restart for failed connectors and tasks should be enabled or disabled. Defaults to true")
```

api/src/main/java/io/strimzi/api/kafka/model/status/AutoRestartStatus.java
```diff
@@ -0,0 +1,83 @@
...
+
+    private int count;
+
+    private String connectorName;
+
+    private String lastRestartTimestamp;
+    private Map<String, Object> additionalProperties;
+
+    @JsonInclude(JsonInclude.Include.NON_NULL)
+    @Description("The number of times the element restarted.")
```
PaulRMellor, line 41:
```suggestion
    @Description("The number of times the connector or task is restarted.")
```

api/src/main/java/io/strimzi/api/kafka/model/status/AutoRestartStatus.java
```diff
@@ -0,0 +1,83 @@
...
+        return connectorName;
+    }
+
+    public void setConnectorName(String connectorName) {
+        this.connectorName = connectorName;
+    }
+
+    @JsonInclude(JsonInclude.Include.NON_NULL)
+    @Description("Last time the  auto restart was attempted. " +
+        "The required format is 'yyyy-MM-ddTHH:mm:ssZ', in the UTC time zone")
```
PaulRMellor, line 61:
```suggestion
        "The required format is 'yyyy-MM-ddTHH:mm:ssZ' in the UTC time zone.")
```

api/src/main/java/io/strimzi/api/kafka/model/status/AutoRestartStatus.java
```diff
@@ -0,0 +1,83 @@
...
+    public String getConnectorName() {
+        return connectorName;
+    }
+
+    public void setConnectorName(String connectorName) {
+        this.connectorName = connectorName;
+    }
+
+    @JsonInclude(JsonInclude.Include.NON_NULL)
+    @Description("Last time the  auto restart was attempted. " +
```
PaulRMellor, line 60:
```suggestion
    @Description("The last time the automatic restart was attempted. " +
```

PaulRMellor (commented):
Looks great! I made some suggestions on the doc. Can re-review for any additions and changes.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/AbstractConnectOperator.java
katheris: Yeah that works well, better than waiting around to see if it restarted. It's then clear to the user that the operator has taken an action. So we just need to make sure at the end of the reconcile loop that the status we got back is included in the status update. That might mean you still need to have this call higher up.

api/src/main/java/io/strimzi/api/kafka/model/KafkaMirrorMaker2ConnectorSpec.java
```diff
@@ -14,7 +14,7 @@
         builderPackage = Constants.FABRIC8_KUBERNETES_API
 )
 @JsonInclude(JsonInclude.Include.NON_NULL)
-@JsonPropertyOrder({"tasksMax", "config"})
+@JsonPropertyOrder({"tasksMax", "autoRestartConnectorsAndTasks", "config"})
```
scholzj, line 17: Should this still be needed? I guess it now lives inside the connector specifications.

api/src/main/java/io/strimzi/api/kafka/model/KafkaConnectorSpec.java
```diff
@@ -32,4 +32,5 @@ public String getClassName() {
     public void setClassName(String className) {
         this.className = className;
     }
+
```
scholzj, line 35: Maybe the new empty line is not needed?

api/src/main/java/io/strimzi/api/kafka/model/AutoRestart.java
```diff
@@ -0,0 +1,51 @@
...
+@DescriptionFile
+@Buildable(
+    editableEnabled = false,
+    builderPackage = Constants.FABRIC8_KUBERNETES_API
+)
+@JsonInclude(JsonInclude.Include.NON_NULL)
+@EqualsAndHashCode
+public class AutoRestart implements UnknownPropertyPreserving, Serializable {
+    private static final long serialVersionUID = 1L;
+    private boolean enabled = false;
```
scholzj, line 29: Wasn't this set to `true` originally? I would do this in one of the following ways:
* Default to `true` here. But leaving the `autoRestart` field in `AbstractConnectorSpec` default to null (as you already have it). That way, it will be disabled by default (as autoRestart is not set by default). But it would allow to just set it to `autoRestart: {}` to enable the auto-restart.
* Or I would simply make it required without any default.

I would do one of the above because it seems a bit weird to me that `autoRestart: {}` would actually mean disabled auto-restart. Seems counterintuitive.

documentation/api/io.strimzi.api.kafka.model.AutoRestart.adoc
```diff
@@ -0,0 +1,42 @@
...
+.Enabling automatic restarts of failed connectors for MirrorMaker 2.0
+[source,yaml,subs="attributes+"]
+----
+apiVersion: strimzi.io/v1alpha1
+kind: KafkaMirrorMaker2
+metadata:
+  name: my-mm2-cluster
+spec:
+  autoRestart:
+    enabled: true
```
scholzj, line 41: This should be updated to match the current API layout with per-connector configurations I guess?

packaging/helm-charts/helm3/strimzi-kafka-operator/crds/048-Crd-kafkamirrormaker2.yaml
```diff
@@ -244,6 +244,13 @@ spec:
                             type: integer
                             minimum: 1
                             description: The maximum number of tasks for the Kafka Connector.
+                          autoRestartConnectorsAndTasks:
```
scholzj, line 247: I guess the CRDs were not regenerated after you changed the field name to `autoRestart`?

api/src/main/java/io/strimzi/api/kafka/model/AutoRestart.java
```diff
@@ -0,0 +1,50 @@
...
+@DescriptionFile
+@Buildable(
+    editableEnabled = false,
+    builderPackage = Constants.FABRIC8_KUBERNETES_API
+)
+@JsonInclude(JsonInclude.Include.NON_NULL)
+@EqualsAndHashCode
+public class AutoRestart implements UnknownPropertyPreserving, Serializable {
+    private static final long serialVersionUID = 1L;
+    private boolean enabled;
```
scholzj, line 29: `boolean` defaults to false. So I think you should set it explicitely to `true`.

scholzj:
> Does anyone know how I can test this method ?
https://github.com/strimzi/strimzi-kafka-operator/pull/7500/files#diff-b5b2f23d9bfdb2ded843e763e2a7dc4a438fa18c93820a742a1633e683d55d18R622-R641
It seems a bad idea to wait 2 minutes during IT tests. Maybe I should make this method static and create a unit test class for It ?

Two ideas come to my mind ... make it possible to configure the _current_ time into the method. So instead of testing in a realtime, you can call the method with some _artificial_ timestamp. E.g. always pass `9:00:00`, `9:02:00`, `9:05:00`, `9:06:00` etc. and always check if it will or will not restart. I think that is how we test other things which depend on time such as certificate expirations etc. I guess other option might be to make the _restart strategy_ more configurble and for example test it with restart after 200 milliseconds instead of 2 minutes or something like that. But the time based manipulation would be what I would try first I guess. @tombentley @katheris any better ideaa?

> I didn't find any related tests for the underlying MirrorMaker2's connectors. Is there something I'm missing or Kafka Connector tests are enough ?

TBH, I'm not sure. IMHO the Mirror Maker connectors are passed into the connector operator. So maybe there are no other tests?

operator-common/src/main/java/io/strimzi/operator/common/operator/resource/StatusUtils.java
```diff
@@ -48,6 +50,25 @@ public static String iso8601(Instant instant) {
         return ZonedDateTime.ofInstant(instant, ZoneOffset.UTC).format(DateTimeFormatter.ISO_INSTANT);
     }
 
+    /**
+     *  Returns an Instant from a string date in ISO 8601 format
```
tombentley, line 54:
```suggestion
     * Returns an Instant from a string date in ISO 8601 format
```

operator-common/src/main/java/io/strimzi/operator/common/operator/resource/StatusUtils.java
```diff
@@ -48,6 +50,25 @@ public static String iso8601(Instant instant) {
...
+     * @param date a string representing a date, for example "2019-07-23T09:08:12.356Z"
+     * @return an Instant
+     */
+    public static Instant iso8601(String date)  {
```
tombentley, line 58: 8601 also defined formats for pure dates, pure times and date times without timezones, so the name `iso8601` is a bit ambiguous. I wonder if `isoUtcDatetime` would be better.

operator-common/src/main/java/io/strimzi/operator/common/operator/resource/StatusUtils.java
```diff
@@ -48,6 +50,25 @@ public static String iso8601(Instant instant) {
...
+        return Instant.parse(date);
+    }
+
+    /**
+     * Get an amount of time unit between a date and now
+     * @param date the date to start from
+     * @param unit the time unit of the interval, for example ChronoUnit.MINUTES
+     * @return long amount of time
+     */
+    public static long unitDifferenceUntilNow(Instant date,  ChronoUnit unit) {
```
tombentley, line 68: Since you only call it with `unit` as minutes we might as well push that into the method body and get rid of the parameter. Then we could call method this `numMinutesAgo`.

tombentley (approved):
Left a few nits, but overall I'm happy with this. Thanks!

scholzj (commented):
LGTM. The build failed because of an unused import:
```
[ERROR] /home/vsts/work/1/s/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/AbstractConnectOperator.java:79:8: Unused import - java.time.temporal.ChronoUnit. [UnusedImports]
```
Can you please fix it?

scholzj:
@katheris Do you have something more for this PR?

api/src/main/java/io/strimzi/api/kafka/model/KafkaMirrorMaker2ConnectorSpec.java
ajborley: I don't have a problem with that either - I think it's unlikely that a user would want to switch autoRestart off for one MM2 connector and not the others, but at least this way they have the option.

(thanks for tagging me in here Jakub, and thanks @ThomasDangleterre for picking this up! :) )

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/AbstractConnectOperator.java
ajborley: Yes, that's right

> if the auto-restart fails 7 times we can probably assume that there is a problem with the config of the connector or task, so continuing to restart it forever will not help fix the problem.

Yes, that's the idea. The corollary here is that if the restart works and the connector is running happily for a while after restarting, but then fails again, it should start a new series of restarts. 

For example, I could imagine a connector that uses an external service, and that external service occasionally becomes unresponsive for a short time. We wouldn't want the connector to go into permanently failed state if it happens to have been restarted 7 times over a few days..

scholzj (approved):

scholzj:
@ThomasDangleterre Thanks a lot for this PR, this is a great addition!

kmcrawford:
@scholzj is this live in 0.32 or do we need to wait until 0.33?

im-konge:
Hi @kmcrawford,
this feature is available from 0.33.0 - which is already released (latest released version is 0.33.2).

PR #7515, opened 2022-10-24
Removed unused Vert.x components

ppatierno:
It seems that `vertx-web-common` and `vertx-web-client` are not referenced (so not used) in any of the projects.
This PR removes them from the principal pom.xml.


scholzj (approved):
LGTM assuming it works without them.

ppatierno:
Regression failures are not related to this PR and @im-konge is looking into it, so we can merge this one.

PR #7523, opened 2022-10-26
Update Keycloak Auth README.md

maxemann96:
### Type of change
Documentation

### Description
Fixed a broken docs hyperlink. 
Added an additional hint to another docs and examples.

### Checklist
Documentation only change with no pre created / referenced issue


scholzj (commented):
This is a good catch. Thanks. However, the `examples/` files are updated only during an release. This needs to be fixed in `packaging/examples`. There seems to be also one more place in this file when the URL is used. It will also need to be fixed in the release tooling here: https://github.com/strimzi/strimzi-kafka-operator/blob/main/packaging/examples/Makefile#L15

Please let us know if you can update it. Otherwise I can open a separate Pr for it myself if needed.

scholzj (approved):
Great, thanks for the PR.

ppatierno (approved):
Thanks!

PR #7530, opened 2022-10-27
Try to find better balance between OOM killer and being out of heap space

scholzj:
### Type of change

- Bugfix

### Description

This PR continues to try to find the right balance between the container memory and heap space. Originally, the operator was very often running over the container memory limits and was being killed by OOM killer in Kubernetes. Then we limited more the available heap memory using the `-XX:MinRAMPercentage`. That from my experience helped to reduce the OOM kills (I haven't seen any in my environments). But it introduced very frequent crashes because Java cannot allocate memory on the heap. Lately, this seems to be happening very often even with just one Kafka cluster and one Connect cluter.

This PR continues the tuning attempts and changes the `-XX:MinRAMPercentage` from 10% to 15% to see if it helps. My tests so far show that it helped to minimize the OutOfHeap issues I had very commonly during a regular work. It also does not seem to introduce any frequent OOM kills (but that might require longer observation after the next release). So hopefully this PR helps to bring in a bit more stability.

### Checklist

### DIFF
--- a/docker-images/operator/scripts/launch_java.sh
+++ b/docker-images/operator/scripts/launch_java.sh
@@ -28,7 +28,7 @@ JAVA_OPTS="${JAVA_OPTS} -XX:+ExitOnOutOfMemoryError"
 
 # Default memory options used when the user didn't configured any of these options, we set the defaults
 if [[ "$JAVA_OPTS" != *"MinRAMPercentage"* && "$JAVA_OPTS" != *"MaxRAMPercentage"* ]]; then
-  JAVA_OPTS="${JAVA_OPTS} -XX:MinRAMPercentage=10 -XX:MaxRAMPercentage=20"
+  JAVA_OPTS="${JAVA_OPTS} -XX:MinRAMPercentage=15 -XX:MaxRAMPercentage=20"
 fi
 
 # Disable FIPS if needed


ppatierno (approved):

docker-images/operator/scripts/launch_java.sh
```diff
@@ -28,7 +28,7 @@ JAVA_OPTS="${JAVA_OPTS} -XX:+ExitOnOutOfMemoryError"
 
 # Default memory options used when the user didn't configured any of these options, we set the defaults
 if [[ "$JAVA_OPTS" != *"MinRAMPercentage"* && "$JAVA_OPTS" != *"MaxRAMPercentage"* ]]; then
-  JAVA_OPTS="${JAVA_OPTS} -XX:MinRAMPercentage=10 -XX:MaxRAMPercentage=20"
+  JAVA_OPTS="${JAVA_OPTS} -XX:MinRAMPercentage=15 -XX:MaxRAMPercentage=20"
```
tombentley, line 31: I don't think this (or what we had previously, for that matter) is what we want. According to https://www.baeldung.com/java-jvm-parameters-rampercentage: 

> The MinRAMPercentage parameter, unlike its name, allows setting the maximum heap size for a JVM running with a small amount of memory (less than 200MB).

and 

> The MaxRAMPercentage parameter allows setting the maximum heap size for a JVM running with a large amount of memory (greater than 200 MB).

In saying `-XX:MinRAMPercentage=15` for a container with <200MB of RAM we would use a heap of 0.15×200 = 30MB. This seems:
1. incredibly small
2. not the behaviour we want for small RAM, where we should use proportionately _more_ of container RAM for the Java Heap (and thus less available for page cache, but if you're going to run in a container with such little RAM then that's a choice you've made, even if unknowingly).

Something like `-XX:MinRAMPercentage=50 -XX:MaxRAMPercentage=20` kind of fits better, except it would result in a cliff effect for containers with RAM in the 200M region: We'd go from ~100MB heap to 40MB heap @ 200MB container memory, and only get above 100MB again when container memory was 1GB.

scholzj: I don't think it works this way. When you have less container memory, the percentage needs to be smaller because the other then heap memory areas such as the code cache or how is it called will remain the same. So you have a block of fixed size memory for the JVM which is the same regardless whether you give it 200Mi or 1000Mi of memory. And the rest is left for things such as heap -> so with bigger overall memory, you can use bigger percentage for heap.

I also don't remember the exact numbers. But AFAIK the default settings we have (384Mi) map to the `-XX:MinRAMPercentage` and not to `-XX:MaxRAMPercentage`.

PR #4096, opened 2020-12-12
Kafka connect: add jmxOptions

garmes-gdev:
### Type of change

- Enhancement / new feature

### Description

add jmxOptions to KafkaConnect 

### Checklist

- [x] Write tests
- [x] Make sure all tests pass
- [ ] Update documentation
- [ ] Check RBAC rights for Kubernetes / OpenShift roles
- [ ] Try your changes from Pod inside your Kubernetes and OpenShift cluster, not just locally
- [ ] Reference relevant issue(s) and close them after merging
- [x] Update CHANGELOG.md
- [ ] Supply screenshots for visual changes, such as Grafana dashboards

PR #7534, opened 2022-10-27
Fix JMX port not added to KafkaCluster

michael-barker:
Add the JXM port to the list of container ports when JMX is enabled on a KafkaCluster.

### Type of change

- Bugfix

### Description

When setting `jmxOptions: {}` on a KafkaCluster, the JMX port is not exposed. I noticed on t[he PR to add JMX to KafkaConnectCluster](https://github.com/strimzi/strimzi-kafka-operator/pull/4096/files#diff-9401e8484d93180261dbb8e40efde42a935698119bb824142ad87594ad9a0620R354) that the JMX port is add to the list of container ports. I believe this may need to be added to KafkaCluster as well so that the JMX port is exposed.

### Checklist

- [ ] Write tests
- [ ] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/cluster-operator/src/test/java/io/strimzi/operator/cluster/model/KafkaClusterTest.java
+++ b/cluster-operator/src/test/java/io/strimzi/operator/cluster/model/KafkaClusterTest.java
@@ -456,10 +456,11 @@ public class KafkaClusterTest {
                     .endKafka()
                 .endSpec()
                 .build();
+
         KafkaCluster kc = KafkaCluster.fromCrd(Reconciliation.DUMMY_RECONCILIATION, kafka, VERSIONS);
-        ContainerPort jmxContainerPort = kc.createContainerPort(JMX_PORT_NAME, JMX_PORT, "TCP");
         List<Container> containers = kc.getContainers(ImagePullPolicy.IFNOTPRESENT);
 
+        ContainerPort jmxContainerPort = kc.createContainerPort(JMX_PORT_NAME, JMX_PORT, "TCP");
         assertThat(containers.get(0).getPorts().contains(jmxContainerPort), is(true));
     }
 
--- a/cluster-operator/src/test/java/io/strimzi/operator/cluster/model/KafkaClusterTest.java
+++ b/cluster-operator/src/test/java/io/strimzi/operator/cluster/model/KafkaClusterTest.java
@@ -94,6 +94,8 @@ import java.util.Map;
 import java.util.Set;
 import java.util.stream.Collectors;
 
+import static io.strimzi.operator.cluster.model.AbstractModel.JMX_PORT;
+import static io.strimzi.operator.cluster.model.AbstractModel.JMX_PORT_NAME;
 import static io.strimzi.test.TestUtils.set;
 import static java.util.Arrays.asList;
 import static java.util.Collections.emptyList;
@@ -443,6 +445,22 @@ public class KafkaClusterTest {
         assertThat(headless.getMetadata().getLabels().containsKey(Labels.STRIMZI_DISCOVERY_LABEL), is(false));
 
         checkOwnerReference(kc.createOwnerReference(), headless);
+    }
+
+    @ParallelTest
+    public void testExposesJmxContainerPortWhenJmxEnabled() {
+        Kafka kafka = new KafkaBuilder(KAFKA)
+                .editSpec()
+                    .editKafka()
+                        .withJmxOptions(new KafkaJmxOptionsBuilder().build())
+                    .endKafka()
+                .endSpec()
+                .build();
+        KafkaCluster kc = KafkaCluster.fromCrd(Reconciliation.DUMMY_RECONCILIATION, kafka, VERSIONS);
+        ContainerPort jmxContainerPort = kc.createContainerPort(JMX_PORT_NAME, JMX_PORT, "TCP");
+        List<Container> containers = kc.getContainers(ImagePullPolicy.IFNOTPRESENT);
+
+        assertThat(containers.get(0).getPorts().contains(jmxContainerPort), is(true));
     }
 
     @SuppressWarnings({"checkstyle:MethodLength"})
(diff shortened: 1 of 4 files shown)


scholzj:
This seems a bit weird. How can anyone connect to it if it is not exposed on the container?

scholzj:
In any case, the same problem seems to be also in ZooKeeperCluster in that case. We should probably also have some unit test for that.

scholzj (commented):
Thanks for the PR. 
* Can you please fix the DCO signoff (instructions should be under the "Details" link in the PR's status section). 
* Also, can you fix the same for ZooKeeperCluster as well so that we can fix it all at once?
* It might be also good to have some unit tests for this in `ZooKeeperClusterTest` and `KafkaClusterTest` classes to hve it covered.

scholzj:
I think ZooKeeper has exactly the same support as the Kafka brokers: https://strimzi.io/docs/operators/latest/full/configuring.html#type-KafkaJmxOptions-reference. So it should work exactly the same - including the missing container port. But ok, I can fix it for ZooKeeper separately, that isno problem. Mybe you can add the unit tests for Kafka it self? It also seems there is no method `isJmxEnabled()`. Just a field. So you should just use the field in the `if`.

michael-barker:
I was able to add a unit test. `isJmxEnabled()` is used in other places in `KafkaCluster` and doesn't look like the field is referenced directly anywhere other than the getter and setter.

michael-barker:
Sorry about that. I think I created my branch off of master instead of main so it was pretty out of date. That also explains why I thought it was so much more work to do the same change for Zookeeper :laughing: 

I pushed some changes for both `KafkaCluster` and `ZookeeperCluster` and a test for each that the JMX port is exposed.

scholzj (commented):
Hmm, maybe should think about deleting the `master` branch to not confuse anyone.

It seems the CI is missing one import:

```
[ERROR] /home/vsts/work/1/s/cluster-operator/src/test/java/io/strimzi/operator/cluster/model/ZookeeperClusterTest.java:[277,9] cannot find symbol
  symbol:   class ContainerPort
  location: class io.strimzi.operator.cluster.model.ZookeeperClusterTest
```

But otherwise it looks good.

michael-barker:
Should be good now :+1:

scholzj (approved):

cluster-operator/src/test/java/io/strimzi/operator/cluster/model/KafkaClusterTest.java
```diff
@@ -445,6 +447,22 @@ public void testGenerateHeadlessServiceWithJmxMetrics() {
...
+        Kafka kafka = new KafkaBuilder(KAFKA)
+                .editSpec()
+                    .editKafka()
+                        .withJmxOptions(new KafkaJmxOptionsBuilder().build())
+                    .endKafka()
+                .endSpec()
+                .build();
+        KafkaCluster kc = KafkaCluster.fromCrd(Reconciliation.DUMMY_RECONCILIATION, kafka, VERSIONS);
+        ContainerPort jmxContainerPort = kc.createContainerPort(JMX_PORT_NAME, JMX_PORT, "TCP");
+        List<Container> containers = kc.getContainers(ImagePullPolicy.IFNOTPRESENT);
```
ppatierno, line 461: I am not sure this test is right. You are creating a `KafkaCluster` model from the `Kafka` resource where you enabled JMX which would be enough to set `isJmxEnabled` to true internally. Then you create the container port on the resource, so actually "forcing" the port creation while it should happen because of the set in the resource.
Instead you should just get the container port list via `getContainerPortList()` right after the `fromCrd` and verify that the JMX port is present in that list.

michael-barker: I'm creating `jmxContainerPort` so that and I can check that the list of container ports contains that port. I had tested this by commenting out the new conditional in `KafkaCluster` which adds the JMX port. When the conditional is commented out then this test fails. Would it be clearer if we moved `jmxContianerPort` down where the assertion happens?

ppatierno: Using the `getContainerPortList()` also checks that the `isJmxEnabled` flag is set correctly by the `fromCrd` (depending on the `Kafka` CR) and that based on that flag, the JMX port is created.

ppatierno (changes requested):

cluster-operator/src/test/java/io/strimzi/operator/cluster/model/KafkaClusterTest.java
ppatierno: Ah sorry, I got what you mean now! But moving it later sounds better yes.

ppatierno (approved):

cluster-operator/src/test/java/io/strimzi/operator/cluster/model/KafkaClusterTest.java
michael-barker: Are you suggesting to call `kc.getContainerPortList()` instead of `kc.getContainers()`? I went to make this change but `getContainerPortList` is private on `ZookeeperCluster`.

I think we also commented at the same time so wasn't sure if you saw my previous message.

ppatierno: Yes we commented at same time. Calling `getContainers` is right because it's going to call `getContainerPortList`. I would just create the JMX port right before the assert for clarity. Anyway I already approved it ;-)

michael-barker: I went ahead and moved `jmxContainerPort` next to the assertion. Hopefully that improves the clarity of what this test is asserting.

ppatierno: Thanks @michael-barker !

ppatierno (approved):

cluster-operator/src/test/java/io/strimzi/operator/cluster/model/KafkaClusterTest.java
michael-barker: :rofl: Did it again. At least we're on the same page when commenting at the same time.

scholzj:
Thanks for the PR!

PR #7537, opened 2022-10-27
Add all DNS names to the certificate for the to Cluster-IP listener bootstrap service

scholzj:
### Type of change

- Enhancement / new feature

### Description

The PR #7365 added a new listener type `cluster-ip` which is an internal listener based on per-broker services instead of an headless service and its DNS names. The original PR adds only the `<service-name>.<namespace>.svc` DNS name to the server certificate SANs. So when a user tried to use one of the other DNS names available for the service (e.g. `<service-name>.<namespace>` or just `<service-name>`), then the TLS hostname verification fails.

This PR adds all the possible DNS names of the service to the certificate so that users can use any format for the bootstrap service. Since this is done also om the `ClusterCa` class for the regular services used for `type: internal` listeners or for the replication and control plane listeners, it also refactors the code to have these service name generated in one place instead of many.

### Checklist

- [x] Write tests

### DIFF
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/model/ClusterCa.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/model/ClusterCa.java
@@ -177,22 +177,14 @@ public class ClusterCa extends Ca {
 
     public Map<String, CertAndKey> generateBrokerCerts(String namespace, String cluster, int replicas, Set<String> externalBootstrapAddresses,
                                                        Map<Integer, Set<String>> externalAddresses, boolean isMaintenanceTimeWindowsSatisfied) throws IOException {
-        DnsNameGenerator kafkaDnsGenerator = DnsNameGenerator.of(namespace, KafkaResources.bootstrapServiceName(cluster));
-        DnsNameGenerator kafkaHeadlessDnsGenerator = DnsNameGenerator.of(namespace, KafkaResources.brokersServiceName(cluster));
-
         Function<Integer, Subject> subjectFn = i -> {
             Subject.Builder subject = new Subject.Builder()
                     .withOrganizationName("io.strimzi")
                     .withCommonName(KafkaResources.kafkaStatefulSetName(cluster));
 
-            subject.addDnsName(KafkaResources.bootstrapServiceName(cluster));
-            subject.addDnsName(String.format("%s.%s", KafkaResources.bootstrapServiceName(cluster), namespace));
-            subject.addDnsName(kafkaDnsGenerator.serviceDnsNameWithoutClusterDomain());
-            subject.addDnsName(kafkaDnsGenerator.serviceDnsName());
-            subject.addDnsName(KafkaResources.brokersServiceName(cluster));
-            subject.addDnsName(String.format("%s.%s", KafkaResources.brokersServiceName(cluster), namespace));
-            subject.addDnsName(kafkaHeadlessDnsGenerator.serviceDnsNameWithoutClusterDomain());
-            subject.addDnsName(kafkaHeadlessDnsGenerator.serviceDnsName());
+            subject.addDnsNames(ModelUtils.generateAllServiceDnsNames(namespace, KafkaResources.bootstrapServiceName(cluster)));
+            subject.addDnsNames(ModelUtils.generateAllServiceDnsNames(namespace, KafkaResources.brokersServiceName(cluster)));
+
             subject.addDnsName(DnsNameGenerator.podDnsName(namespace, KafkaResources.brokersServiceName(cluster), KafkaResources.kafkaPodName(cluster, i)));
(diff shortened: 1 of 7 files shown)


cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaListenersReconciler.java
```diff
@@ -403,7 +404,8 @@ protected Future<Void> clusterIPServicesReady()   {
             String bootstrapAddress = getInternalServiceHostname(reconciliation.namespace(), ListenersUtils.backwardsCompatibleBootstrapServiceName(reconciliation.name(), listener), useServiceDnsDomain);
 
             if (listener.isTls()) {
-                result.bootstrapDnsNames.add(bootstrapAddress);
+                ModelUtils.generateAllServiceDnsNames(reconciliation.namespace(), ListenersUtils.backwardsCompatibleBootstrapServiceName(reconciliation.name(), listener))
```
maknihamdi, line 407: I think if this list bootstrapDnsNames is used only with certs, we can avoid to fill it here when ClusterCa.generateBrokerCerts already generate the same list and add dns names to subject. We do it twice now, here https://github.com/scholzj/strimzi-kafka-operator/blob/use-all-DNS-names-for-cluster-ip-listener-bootstrap-servive/cluster-operator/src/main/java/io/strimzi/operator/cluster/model/ClusterCa.java#L185 and here https://github.com/scholzj/strimzi-kafka-operator/blob/use-all-DNS-names-for-cluster-ip-listener-bootstrap-servive/cluster-operator/src/main/java/io/strimzi/operator/cluster/model/ClusterCa.java#L196

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/ClusterCa.java
```diff
@@ -177,22 +177,14 @@ public Map<String, CertAndKey> generateZkCerts(String namespace, String kafkaNam
...
 
-            subject.addDnsName(KafkaResources.bootstrapServiceName(cluster));
-            subject.addDnsName(String.format("%s.%s", KafkaResources.bootstrapServiceName(cluster), namespace));
-            subject.addDnsName(kafkaDnsGenerator.serviceDnsNameWithoutClusterDomain());
-            subject.addDnsName(kafkaDnsGenerator.serviceDnsName());
-            subject.addDnsName(KafkaResources.brokersServiceName(cluster));
-            subject.addDnsName(String.format("%s.%s", KafkaResources.brokersServiceName(cluster), namespace));
-            subject.addDnsName(kafkaHeadlessDnsGenerator.serviceDnsNameWithoutClusterDomain());
-            subject.addDnsName(kafkaHeadlessDnsGenerator.serviceDnsName());
+            subject.addDnsNames(ModelUtils.generateAllServiceDnsNames(namespace, KafkaResources.bootstrapServiceName(cluster)));
```
maknihamdi, line 185: in case of clusterIp listener, I think this list is already filled in externalBootstrapAddresses

scholzj: The certificate exists once per broker regardless the number of listeners. In this place, it adds the dns names for the default services. That is needed regardless any user-configured-listeners since it is used in any case for replication and control-plane listener. So this handles the default services. So for example with cluster named `my-cluster`, this will add DNS names for services `my-cluster-kafka-brokers` and `my-cluster-kafka-bootstrap`.

The place in the KafkaListenerReconciler handles the DNS names for the listeners. When I add a `type: cluster-ip` listener named `tls`, it would for example add the DNS names for the bootstrap service `my-cluster-kafka-tls-bootstrap`. So these will be different from those added in the `ClusterCA` class.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaListenersReconciler.java
scholzj: _Please see the reply on the other comment_

ppatierno (approved):
Just a nit but LGTM.

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/ModelUtils.java
```diff
@@ -733,4 +733,29 @@ public static int caCertGeneration(Ca ca) {
     public static int idOfPod(String podName)  {
         return Integer.parseInt(podName.substring(podName.lastIndexOf("-") + 1));
     }
+
+    /**
+     * Generates all possible DNS names for a Kubernetes service:
+     *     - service-name
+     *     - service-name.namespace
+     *     - service-name.namespace.svc
+     *     - service-name.namespace.svc.dns.suffix
```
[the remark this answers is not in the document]
scholzj: I did not wanted to use the `<...>` format anywhere because you cannot use it directly in JavaDocs. You need to escape it as `&lt;` and `&gt;`. So it it is hard to make it compile but also readable as JavaDoc in the IDE. That is why I didn't used it for the DNS suffix, service name or namespace.

PR #7542, opened 2022-10-28
Improve cold-backup script for dev/test clusters

fvaleri:
This PR adds the following improvements with no interface change.

- Skip internal topics so that the TO can safely reinitialize from Kafka on restore
- Cluster stop and start functions
- Support for clusters using StrimziPodSets
- Readme and help content

### DIFF
--- a/tools/cold-backup/README.md
+++ b/tools/cold-backup/README.md
@@ -1,18 +1,25 @@
-## Strimzi backup
+## Cold backup
 
-Bash script for offline (_cold_) backups of Kafka clusters on Kubernetes or OpenShift. 
-The script only supports a local file system, and you need to make sure to have enough free disk space in the target directory.
+Bash script for running cold backups of test or development Kafka clusters deployed on Kubernetes.
 
-If you think you don't need a backup strategy for Kafka because of its embedded data replication, then consider the impact of a misconfiguration, bug, or security breach that deletes all of your data. 
-For online (_hot_) backups, you can use storage snapshotting or stream into object storage.
+The backup procedure will stop the cluster operator and selected cluster for the entire duration of the backup.
+Run the script as a Kubernetes user with permission to work with PVC and Strimzi custom resources.
+The script only supports local file system, and you need to make sure to have enough free disk space in the target directory.
 
-Run the script as a Kubernetes user with permission to work with PVC and Strimzi custom resources. 
+Before restoring the Kafka cluster you need to make sure to have the right version of Strimzi CRDs installed.
+The restored cluster is actually the same Kafka cluster, because volumes contain the same Kafka cluster ID.
 
-The procedure will stop the Cluster Operator and selected cluster for the duration of the backup. 
-Before restoring the Kafka cluster you need to make sure to have the right version of Strimzi CRDs installed. 
-If you have a single cluster-wide Cluster Operator, then you need to scale it down manually. 
-You can run backup and restore procedures for different Kafka clusters in parallel. 
-Consumer group offsets are included, but not Kafka Connect, MirrorMaker and Kafka Bridge custom resources.
+If one or more ZooKeeper nodes are crash looping because there is no snapshot, you can apply the following patch to recover them.
+
+```sh
+kubectl patch k my-cluster --type merge -p '
+  spec:
+    zookeeper:
+      jvmOptions:
+        javaSystemProperties:
+          - name: zookeeper.snapshot.trust.empty
+            value: "true"'
+```
 
 ## Requirements
 
@@ -20,7 +27,6 @@ Consumer group offsets are included, but not Kafka Connect, MirrorMaker and Kafk
 - tar 1.33+ (GNU)
 - kubectl 1.19+ (K8s CLI)
 - yq 4.6+ (YAML processor)
-- enough disk space
 
 ## Usage example
 
--- a/tools/cold-backup/README.md
+++ b/tools/cold-backup/README.md
@@ -2,24 +2,12 @@
 
 Bash script for running cold backups of test or development Kafka clusters deployed on Kubernetes.
 
-The backup procedure will stop the cluster operator and selected cluster for the entire duration of the backup.
+The script will pause the reconciliation and stop the selected cluster for the entire backup duration.
 Run the script as a Kubernetes user with permission to work with PVC and Strimzi custom resources.
 The script only supports local file system, and you need to make sure to have enough free disk space in the target directory.
 
 Before restoring the Kafka cluster you need to make sure to have the right version of Strimzi CRDs installed.
 The restored cluster is actually the same Kafka cluster, because volumes contain the same Kafka cluster ID.
-
-If one or more ZooKeeper nodes are crash looping because there is no snapshot, you can apply the following patch to recover them.
-
-```sh
-kubectl patch k my-cluster --type merge -p '
-  spec:
-    zookeeper:
-      jvmOptions:
-        javaSystemProperties:
-          - name: zookeeper.snapshot.trust.empty
-            value: "true"'
-```
 
 ## Requirements
 
--- a/tools/cold-backup/run.sh
+++ b/tools/cold-backup/run.sh
@@ -39,6 +39,8 @@ check_kube_conn() {
   kubectl version -o yaml --request-timeout=5s 1>/dev/null
 }
 
+# This function can be used to wait for some resource condition in a specific namespace.
+# Here you can find some examples on how to use it:
 # wait_for "condition=Ready" "kafkas.kafka.strimzi.io/my-cluster" "my-namespace"
 # wait_for "condition=Ready" "pod -l strimzi.io/kind=cluster-operator" "my-namespace"
 # wait_for "delete" "kafkas.kafka.strimzi.io/my-cluster" "my-namespace"
--- a/tools/cold-backup/run.sh
+++ b/tools/cold-backup/run.sh
@@ -5,11 +5,11 @@ if [[ $(uname -s) == "Darwin" ]]; then
   alias echo="gecho"; alias dirname="gdirname"; alias grep="ggrep"; alias readlink="greadlink"
   alias tar="gtar"; alias sed="gsed"; alias date="gdate"; alias ls="gls"
 fi
-BASE="" && pushd "$(dirname "$(readlink -f "${BASH_SOURCE[0]}")")" >/dev/null \
-  && { BASE=$PWD; popd >/dev/null; } && readonly BASE
+SCRIPT_DIR="" && pushd "$(dirname "$(readlink -f "${BASH_SOURCE[0]}")")" >/dev/null \
+  && { SCRIPT_DIR=$PWD; popd >/dev/null; } && readonly SCRIPT_DIR
 trap 'failure "$LINENO" "$BASH_COMMAND"' ERR
-
 readonly RSYNC_POD_NAME="cold-backup"
+
 ZK_REPLICAS=0
 ZK_PVC=()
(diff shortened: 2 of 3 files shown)


tools/cold-backup/README.md
```diff
@@ -1,26 +1,32 @@
-## Strimzi backup
+## Cold backup
 
-Bash script for offline (_cold_) backups of Kafka clusters on Kubernetes or OpenShift. 
-The script only supports a local file system, and you need to make sure to have enough free disk space in the target directory.
+Bash script for running cold backups of test or development Kafka clusters deployed on Kubernetes.
```
scholzj, line 3: I don't think this makes much sense. 
* Why do you need backup for development and test clusters?
* Why is this not good enough for production (assuming you can live with the outage)?

tools/cold-backup/README.md
```diff
@@ -1,26 +1,32 @@
...
 
-If you think you don't need a backup strategy for Kafka because of its embedded data replication, then consider the impact of a misconfiguration, bug, or security breach that deletes all of your data. 
-For online (_hot_) backups, you can use storage snapshotting or stream into object storage.
+The backup procedure will stop the cluster operator and selected cluster for the entire duration of the backup.
```
scholzj, line 5: Should you pause it instead to avoid disruption of other clusters?

tools/cold-backup/README.md
```diff
@@ -1,26 +1,32 @@
...
-Run the script as a Kubernetes user with permission to work with PVC and Strimzi custom resources. 
+Before restoring the Kafka cluster you need to make sure to have the right version of Strimzi CRDs installed.
+The restored cluster is actually the same Kafka cluster, because volumes contain the same Kafka cluster ID.
 
-The procedure will stop the Cluster Operator and selected cluster for the duration of the backup. 
-Before restoring the Kafka cluster you need to make sure to have the right version of Strimzi CRDs installed. 
-If you have a single cluster-wide Cluster Operator, then you need to scale it down manually. 
-You can run backup and restore procedures for different Kafka clusters in parallel. 
-Consumer group offsets are included, but not Kafka Connect, MirrorMaker and Kafka Bridge custom resources.
+If one or more ZooKeeper nodes are crash looping because there is no snapshot, you can apply the following patch to recover them.
```
scholzj, line 12: Is the tool not backing-up the snapshots? Or why is this a concern?

tools/cold-backup/run.sh
```diff
@@ -30,109 +30,94 @@ FILE_NAME=""
...
+  exit 1
+}
+
+check_kube_conn() {
+  kubectl version -o yaml --request-timeout=5s 1>/dev/null
+}
+
+# wait_for "condition=Ready" "kafkas.kafka.strimzi.io/my-cluster" "my-namespace"
+# wait_for "condition=Ready" "pod -l strimzi.io/kind=cluster-operator" "my-namespace"
+# wait_for "delete" "kafkas.kafka.strimzi.io/my-cluster" "my-namespace"
```
scholzj, line 44: Is this an example how to use it? Or some left over code? If it is left here intentionally, some explanation might be good.

tools/cold-backup/run.sh
```diff
@@ -30,109 +30,94 @@ FILE_NAME=""
...
     echo "Stopping cluster $CLUSTER_NAME"
     local co_deploy && co_deploy="$(kubectl -n "$NAMESPACE" get deploy strimzi-cluster-operator -o name --ignore-not-found)"
-    if [[ -n $co_deploy ]]; then
-      kubectl -n "$NAMESPACE" scale "$co_deploy" --replicas 0
-      wait_for "delete" "pod -l strimzi.io/kind=cluster-operator"
-    else
-      echo "No local operator found, make sure no cluster-wide operator is watching this namespace"
-      confirm
+    if [[ -z $co_deploy ]]; then
+      confirm "No local operator found, make sure no operator is watching this Kafka cluster"
```
scholzj, line 79: Does this mean the operator has to run in the same namespace as the cluster? I guess one more reason to use pausing of the resources rather then scaling down the operator.

tools/cold-backup/README.md
fvaleri: > Why do you need backup for development and test clusters?

Maybe you did a fairly complex setup that you want to backup and restore as needed. In some testing/demo environment, K8s clusters are automatically deleted after some time. Here I'm thinking about support teams that works with multiple customers.

> Why is this not good enough for production (assuming you can live with the outage)?

Well, the backup/restore procedures are pretty slow because they use tar that does not support parallel execution. I'm sure that there is a way to make it run all volumes in parallel, but it would be too complex for a bash script. I would say, if users find this tool useful, we could implement it with Quarkus for example. WDYT?

tools/cold-backup/README.md
fvaleri: Yes, that's actually what it does now. I just forgot to remove that sentence.

tools/cold-backup/README.md
fvaleri: This can sometimes happen when you backup and restore a freshly deployed clusters with very few or no data.

tools/cold-backup/run.sh
fvaleri: Yes, these examples are intentionally left. Let me improve the comment.

tools/cold-backup/run.sh
fvaleri: No. It is already pausing the reconciliation :)

tools/cold-backup/README.md
scholzj: I can understand that it has its limitations for production. But TBH:
* What you say there basically says do not use this in production
* I'm not sold on the idea how useful tool it is for development and testing. Especially if it is _slow_ as you describe it.
* A good practice for demos is to setup things from scratch to show how things work. Not just restore some backup.

So, honestly, if it is not good enough for production ... I have to wonder if we are just wasting our time on this and if it should be cut from the project. Because it seems to have very little value.

tools/cold-backup/README.md
scholzj: This can happen in general, yes. But I do not see why to cover it here.
* The user can be affected by this even without using the backup tool - for example during a rolling update. So this is not unique to the backup tool.
* Going back to the first comment ... what is really the value of restoring a brand new cluster in test and development? (other then testing this tool of course)

tools/cold-backup/README.md
fvaleri: It has been useful in a very limited set of use cases, so I'm ok with cutting it. Should I open a PR for that?

scholzj: It probably should be discussed on spme community call to see what others think.

fvaleri: OK, fine. Thanks.

scholzj:
Discussed ont he community call on 3rd November:
* @fvaleri Do you have any more detailed use cases when this would be used in development and test? And how would it be useful? If yes, can you please share them?

fvaleri:
I played with this a little more and improved the script to the point we can run it without stopping the cluster. The actual slowness was coming from the final compression step, where gzip uses only one core. Replacing gzip with pigz makes it a lot faster.  

Cluster data are copied using the GNU tar utility, which sequentially stores files in the output. This is not a problem, because the bottleneck is the network and/or the hard drive.

As an example, it takes 30 minutes to produce a 90 GiB backup of a 3 nodes cluster with 146 GiB of data (minikube). The exact same test on a remote cluster in the EU region takes 80 minutes (OCP with 3x m5.4xlarge workers and EBS).

That said, I still think this doesn't justify the maintenance burden (bash is not a good fit for this kind of complexity). We can simply remove it from Strimzi and I'll make it available as a public repository for whoever is interested. 

If you agree, I'll open a PR to remove it.

@scholzj @tombentley

scholzj:
As I said before ... you would need to be able to guarantee that the _online_ backup is recoverable in 100% of the cases and I'm not sure you are able to guarantee it. I guess @tombentley might know if that is something what can be guaranteed or not.

Would the _online_ backup if it works make it production ready? It take that it would not from the last paragraph, correct?

fvaleri:
> As I said before ... you would need to be able to guarantee that the _online_ backup is recoverable in 100% of the cases and I'm not sure you are able to guarantee it. I guess @tombentley might know if that is something what can be guaranteed or not.
> 

In all the tests I did, I was always able to restore the backup without any data loss if this is what you mean. With the *online* mode was the same so far, but I guess we would need much more tests.

> Would the _online_ backup if it works make it production ready? It take that it would not from the last paragraph, correct?

For me that would imply having a suite of unit/integration tests. I'll share the last version and, if we find it correct and valuable, I can explore a Java implementation of the same algorithm.

scholzj:
> For me that would imply having a suite of unit/integration tests.

There is a system test for this IIRC.

tombentley:
> In all the tests I did, I was always able to restore the backup without any data loss if this is what you mean. With the _online_ mode was the same so far, but I guess we would need much more tests.

There are a bunch of difficult-to-system-test corner cases to consider. These are just the ones from the top of my head:

* What happens if a broker restarts during the backup and truncates its log? Could we back up data-that-never-was?
* What about compacted topics where the log cleaner is executing while we're taking the backup (I _guess_ they'd be ok, since they'll be in a temporary segment which we can ignore)? 
* What about backing up partitions of topics which are deleted and re-created during the backup -- do we end up with segments mixed from both old and new topics?
* What are the effects of backing up on disk and network bandwidth? Do we need throttling? How do we orchestrate that to avoid the backup never completing, or taking too long?
* If we back up a `__transaction_state` segment before a data partition is added to a transaction, but back up the data partition after a transaction was started can we end up with a hanging transaction upon restore?

>  production ready

No one's day is improved when the backup they faithfully took fails to restore. To say it's production ready we'd need to be very confident that we can always restore correctly from a backup, which means we need to have answers to all of the above questions. That's because if people believe this backup is "production ready" they will likely make different (possibly more risky) decisions than if they believe they have a "best effort, no guarantees" backup.

fvaleri:
Hi @tombentley. Thanks for having a look.

> What happens if a broker restarts during the backup and truncates its log? Could we back up data-that-never-was?

If that happens because of a data corruption that was already there, then I guess the truncation will happen again, as soon as the cluster is restored and the broker pod started. Instead, if it happens because of a disk corruption (bad blocks), then I think you could detect that, replace the volume use the backup to restore corrupted data.

> What about compacted topics where the log cleaner is executing while we're taking the backup (I guess they'd be ok, since they'll be in a temporary segment which we can ignore)?

Yeah, haven't thought about that. Currently I'm copying all files, but you are right, we need to filter out that kind of temporary files.

> What about backing up partitions of topics which are deleted and re-created during the backup -- do we end up with segments mixed from both old and new topics?

Yes, I think that's the risk, because the layout on disk is based on topic name, not id. That's a fundamental problem IMO.

> What are the effects of backing up on disk and network bandwidth? Do we need throttling? How do we orchestrate that to avoid the backup never completing, or taking too long?

Good point. We would need to add some flags to set and update throttling.

> If we back up a __transaction_state segment before a data partition is added to a transaction, but back up the data partition after a transaction was started can we end up with a hanging transaction upon restore?

In theory this should be covered, because we could have an inflight transaction when all the brokers fail at once (i.e. power failure). That's said, I won't bet on it.

To summarize: I see a fundamental problem with topic recreations and risks with transactions. Probably a connector that periodically syncs with S3 or another high latency storage is a much better option for backups, and that would also work for on-prem.

fvaleri:
I'm closing this PR and will open a new one to remove this tool based on the above discussion.

Thanks.

PR #7543, opened 2022-10-28
Disabled bridge metrics test

ppatierno:
The `MetricsIsolatedST.testKafkaBridgeMetrics` started to fail on checking the `system_cpu_count` metric which gets a different value depending on the platform (i.e. in AZP fails with 1 != 2, on my laptop 1 != 12, ...)
Waiting for investigating why it's now failing but moving forward with the next release, this PR disables the test for now.

### DIFF
--- a/systemtest/src/test/java/io/strimzi/systemtest/metrics/MetricsIsolatedST.java
+++ b/systemtest/src/test/java/io/strimzi/systemtest/metrics/MetricsIsolatedST.java
@@ -60,6 +60,7 @@ import io.strimzi.test.executor.Exec;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.junit.jupiter.api.BeforeAll;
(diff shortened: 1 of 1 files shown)


systemtest/src/test/java/io/strimzi/systemtest/metrics/MetricsIsolatedST.java
```diff
@@ -431,6 +432,7 @@ void testMirrorMaker2Metrics() {
         assertThat(values.stream().mapToDouble(i -> i).sum(), is((double) 1));
     }
 
+    @Disabled("Started to fail on the system_cpu_count metric check which is different depending on the environment")
```
scholzj, line 435: You should probably open an issue for fixing and re-enabling it and add link to it here in a comment?

ppatierno: Done.

scholzj (approved):

PR #7558, opened 2022-10-31
Fix flaky `StrimziPodSetControllerMockTest.testFailedPodRecovery` unit test

scholzj:
### Type of change

- Bugfix

### Description

The `StrimziPodSetControllerMockTest.testFailedPodRecovery` unit test is sometimes flaky on Azure. This PR tries to fix the NPE which is happening due to a race condition when the status of the pod is not yet set by the _mock pod controller_.

### Checklist

### DIFF
--- a/cluster-operator/src/test/java/io/strimzi/operator/cluster/operator/assembly/StrimziPodSetControllerMockTest.java
+++ b/cluster-operator/src/test/java/io/strimzi/operator/cluster/operator/assembly/StrimziPodSetControllerMockTest.java
@@ -686,7 +686,9 @@ public class StrimziPodSetControllerMockTest {
                     10_000,
                     () -> {
                         Pod p = client.pods().inNamespace(NAMESPACE).withName(podName).get();
-                        return p != null && !resourceVersion.equals(p.getMetadata().getResourceVersion());
+                        return p != null
+                                && !resourceVersion.equals(p.getMetadata().getResourceVersion())
+                                && p.getStatus() != null; // We need to also wait for the Pod status to be set to avoid NullPointerExceptions
                     },
                     () -> context.failNow("Test timed out waiting for pod recreation!"));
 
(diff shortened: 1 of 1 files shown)


Frawless (approved):

cluster-operator/src/test/java/io/strimzi/operator/cluster/operator/assembly/StrimziPodSetControllerMockTest.java
```diff
@@ -686,7 +686,9 @@ public void testFailedPodRecovery(VertxTestContext context) {
                     10_000,
                     () -> {
                         Pod p = client.pods().inNamespace(NAMESPACE).withName(podName).get();
-                        return p != null && !resourceVersion.equals(p.getMetadata().getResourceVersion());
+                        return p != null
+                                && !resourceVersion.equals(p.getMetadata().getResourceVersion())
+                                && p.getStatus() != null; // We need to also wait for the Pod status to be set to avoid NullPointerExceptions
```
tombentley, line 691: Where do the NPEs manifest?

scholzj: In the assert below this `waitFor` call. Basically:
* `StrimziPodSetController` recreates the Pod without status (which normally Kubernetes would manage)
* When using MockKube, the `MockPodController` will set the Pod as ready instead of Kubernetes

As this is asynchronous, there is a time window when the Pod is recreated but is missing the status which causes the NPE in the assert below. By waiting for the status to be present as well, we eliminate the race condition and make sure the status is there and does not throw an NPE.

An alternative would be to do something like this:

```java
assertThat(recreatedPod.getStatus() == null || !"Failed".equals(recreatedPod.getStatus().getPhase()), is(true));
```

But improving the wait condition seemed to make the code a bit more readable.

scholzj: @tombentley I updated the comment to make it a bit more clear.

tombentley (approved):
Thanks @scholzj

PR #7559, opened 2022-11-01
Delete the `StrimziPodSet` or `StatefulSet` first when migrating between them

scholzj:
### Type of change

- Bugfix

### Description

When the `UseStrimziPodSets` feature gate is enabled or disabled, the operator needs to migrate between `StrimziPodSets` (SPS) and `StatefulSets` (STS) for the existing clusters. Right now, we have two methods which are executed one after another. The first one either creates or deletes the STS resource. The other one either creates or deletes the SPS resource. 

Because the STS method is first, it means that when moving from SPS to STS, when the new STS is created, the old SPS will still exist. That used to work fine and the STS took over the pods and become ready immediately. But probably since we set the `controller` flag in the Pod's owner reference when managing them through SPS (which is important for other reasons), the StatefulSet does not take over the pods while the SPS exists and does not become ready. 

And because the method which creates the STS waits for it to be ready, it means it never completes and the reconciliation timeouts instead of moving to STS. After the timeout, the next reconciliation picks up and completes the migration. So this is not a blocking issue. But because of the timeouts, it takes much longer than needed to move from STS to SPS.

This PR splits the methods in the reconciliation. It makes sure that if the transition from STS to SPS or the other way around is happening, that the old resource is deleted first before creating the new one. That helps to make the transition smoother and faster.

### Checklist


Frawless (approved):

see-quick (approved):

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaReconciler.java
```diff
@@ -259,6 +259,8 @@ public Future<Void> reconcile(KafkaStatus kafkaStatus, Clock clock)    {
                 .compose(i -> jmxSecret())
                 .compose(i -> podDisruptionBudget())
                 .compose(i -> podDisruptionBudgetV1Beta1())
+                .compose(i -> maybeDeleteStatefulSet())
+                .compose(i -> maybeDeletePodSet())
```
tombentley, line 263: It's a shame that the names don't reflect that the deletions are specific to migration. I can't think of a better suggestion though. `migrateFromStatefulSetController` and `migrateFromPodSetController` perhaps?

tombentley (approved):

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaReconciler.java
scholzj: Sounds like a good idea. I will rename them.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaReconciler.java
```diff
@@ -852,6 +853,19 @@ protected Future<Void> statefulSet() {
                         return Future.succeededFuture();
                     });
         } else {
+            return Future.succeededFuture();
+        }
+    }
+
+    /**
+     * Helps with the migration from StatefulSets to StrimziPodSets when the cluster is switching between them. When the
+     * switch happens, it deletes the old StatefulSet. It should happen before the new PodSet StatefulSet is created to
```
ppatierno, line 862: just `PodSet` here? Removing `StatefulSet`?

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/ZooKeeperReconciler.java
```diff
@@ -522,12 +524,34 @@ protected Future<Void> podDisruptionBudgetV1Beta1() {
...
+                        return Future.succeededFuture();
+                    });
+        } else {
+            return Future.succeededFuture();
+        }
+    }
+
+    /**
+     * Helps with the migration from StatefulSets to StrimziPodSets when the cluster is switching between them. When the
+     * switch happens, it deletes the old StatefulSet. It should happen before the new PodSet StatefulSet is created to
```
ppatierno, line 549: Ditto as above?

ppatierno (approved):
LGTM. I left just a nit.

PR #7562, opened 2022-11-01
[systemtest] Refactor metrics tests

### LINKED ISSUE #7544: Fix and re-enable `MetricsIsolatedST.testKafkaBridgeMetrics` test
The `MetricsIsolatedST.testKafkaBridgeMetrics` is disabled with https://github.com/strimzi/strimzi-kafka-operator/pull/7543.
After digging into it and fixing the issue we should re-enable the test again.

im-konge:

### Type of change

- Bugfix
- Refactoring

### Description

This PR completely refactors the whole `MetricsIsolatedST` suite, which is flaky. As part of this refactoring, I added `collectedData` into the `MetricsCollector`, so each instance holds latest version of metrics get from each `resource` or `pod`. When the `collectedData` doesn't contain specific pattern, the `waitFor` is used here for specific metric to appear. For each poll, `collectedData` is updated with new value. This way I removed flakiness in case the metric wasn't present just in the moment of assert.

Also, this PR removes some checks, which weren't working before - for example checking metrics of resources, which doesn't exist.

As part of this PR, I removed few metrics, which were not relevant (because it depended or environment or resource and it's state):

`system_cpu_count`
`strimzi_reconciliations_locked_total`
`strimzi_reconciliations_failed_total` - if reconciliation doesn't fail, this metric is not present in metrics
`strimzi_reconciliations_periodical_total` for KafkaConnector, which is not present

Fixes #7544 

### Checklist

### DIFF
--- a/systemtest/src/main/java/io/strimzi/systemtest/metrics/MetricsCollector.java
+++ b/systemtest/src/main/java/io/strimzi/systemtest/metrics/MetricsCollector.java
@@ -39,6 +39,8 @@ public class MetricsCollector {
 
     private static final Logger LOGGER = LogManager.getLogger(MetricsCollector.class);
 
+    private static final Object LOCK = new Object();
+
     private String namespaceName;
     private String scraperPodName;
     private ComponentType componentType;
@@ -46,6 +48,7 @@ public class MetricsCollector {
     private int metricsPort;
     private String metricsPath;
     private LabelSelector componentLabelSelector;
+    private Map<String, String> collectedData;
 
     public static class Builder {
         private String namespaceName;
@@ -112,6 +115,10 @@ public class MetricsCollector {
 
     public int getMetricsPort() {
         return metricsPort;
+    }
+
+    public Map<String, String> getCollectedData() {
+        return collectedData;
     }
 
     protected MetricsCollector.Builder newBuilder() {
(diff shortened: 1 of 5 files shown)


Frawless (approved):

systemtest/src/main/java/io/strimzi/systemtest/metrics/MetricsCollector.java
```diff
@@ -208,21 +215,51 @@ private int getDefaultMetricsPortForComponent() {
...
     }
 
+    /**
+     * Method checks already collected metrics data for Pattern containing desired metric
+     * @param pattern Pattern of metric which is desired
+     *
+     * @return ArrayList of values collected from the metrics
+     */
+    public ArrayList<Double> waitForSpecificMetricAndCollect(Pattern pattern) {
+        synchronized (LOCK) {
```
see-quick, line 242: You can use synchronized keyword in the defining method scope:
```java
public ArrayList<Double> waitForSpecificMetricAndCollect(Pattern pattern) 
```
--
```java
public synchronized ArrayList<Double> waitForSpecificMetricAndCollect(Pattern pattern) 
```

see-quick (approved):
Great work! 🕺👍 One nit...

systemtest/src/main/java/io/strimzi/systemtest/metrics/MetricsCollector.java
im-konge: Good point, thanks! :)

PR #7566, opened 2022-11-03
Add liveness and readiness checks for KRaft mode

katheris:
### Type of change

_Select the type of your PR_

- Enhancement / new feature

### Description

Add liveness and readiness checks for KRaft mode.

**KRaft Controller probes:**
 - Liveness: listening on 9090 (the port set on the listener specified under controller.listener.names)
 - Readiness: listening on 9090 (the port set on the listener specified under controller.listener.names)

The readiness probe and liveness probe for a KRaft Controller are identical because there is no clear state change that would indicate the controller is "Ready" rather than simply alive. The current-state metric (https://kafka.apache.org/documentation/#kraft_quorum_monitoring) indicates whether this member is part of a quorum, however a pod should be marked as ready based on it's own status, not the status of other pods. So this metric is not useful for the readiness probe.

**KRaft Broker probes:**
 - Liveness: broker state < 3 || listening 9091 (replication port)
 - Readiness: broker state >= 3 && broker state != 127 (127 represents unknown state)

The expected broker state transitions are (https://github.com/apache/kafka/blob/trunk/metadata/src/main/java/org/apache/kafka/metadata/BrokerState.java):
NOT_RUNNING (0)
    ↓
STARTING (1)
    ↓
RECOVERY (2)
    ↓
RUNNING (3)
    ↓
PENDING_CONTROLLED_SHUTDOWN (6)
    ↓
SHUTTING_DOWN (7)

The other possible value is 127 which represents UNKNOWN.

For the liveness probe, if the broker is starting or in recovery mark as alive so that Kubernetes does not kill the pod while it is still starting up. Once the broker is in RUNNING state we should expect it to be listening on the replication port so it can communicate with the other brokers.

For the readiness probe, only mark the broker as ready once it reaches the RUNNING stage. We should tolerate PENDING_CONTROLLED_SHUTDOWN and SHUTTING_DOWN states, since it may still be doing work in this time. We should not tolerate an UNKNOWN status as this means the broker is likely not in a good state.

### Checklist

- [ ] Write tests
- [ ] Update documentation
- [x] Update CHANGELOG.md


kafka-agent/src/main/java/io/strimzi/kafka/agent/KafkaAgent.java
```diff
@@ -33,17 +33,24 @@ public class KafkaAgent {
     private static final String YAMMER_METRICS_IN_KAFKA_3_2_AND_EARLIER = "kafka.metrics.KafkaYammerMetrics";
 
     private final File sessionConnectedFile;
-    private File brokerReadyFile;
+    private final File brokerReadyFile;
     private MetricName brokerStateName;
     private Gauge brokerState;
     private MetricName sessionStateName;
     private Gauge sessionState;
+    private Boolean inKRaftMode = false;
```
scholzj, line 41: Do we need `Boolean` instead of `boolean`? Also, should we set it to `false` in the existing constructor and mark it as final? (no idea if it adds anything, just wondering since you seem to do it to the other fields :-o)

scholzj (commented):
I left some comments inline.  But in general:
* I think the description would deserve a proper explanation of why does it do what it does. For example, it is completely unclear why does controller check only for ports being listened to. While broker checks for additional states. I think it is really important to track the reasoning behind it and how it works. We do not have proposal for this, so we should track it here. If some of that leaks into comments in the shell scripts, I don't think it would hurt either.
* You seem to have controller check. You seem to have broker check (although currently not used as we do not support clear brokers or controllers). But it seems to be missing checks for the mixed mode. This is a bit weird since that is the mode that we currently have.

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaCluster.java
```diff
@@ -1575,6 +1575,9 @@ protected List<Container> getInitContainers(ImagePullPolicy imagePullPolicy) {
 
     @Override
     protected List<Container> getContainers(ImagePullPolicy imagePullPolicy) {
+        String livenessCommand = useKRaft ? "/opt/kafka/kafka_liveness_kraft.sh" : "/opt/kafka/kafka_liveness.sh";
+        // In ZooKeeper mode kafka-agent will create /var/opt/kafka/kafka-ready in the container
+        String[] readinessCommand = useKRaft ? new String[]{"/opt/kafka/kafka_readiness_kraft.sh"} : new String[]{"test", "-f", "/var/opt/kafka/kafka-ready"};
```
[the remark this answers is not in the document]
katheris: I hadn't, I hadn't spotted the `STRIMZI_KRAFT_ENABLED` env var, so if we can use that it would mean keeping all the ZooKeeper vs KRaft logic in the scripts which I agree would be cleaner

docker-images/kafka-based/kafka/scripts/kafka_run.sh
```diff
@@ -63,8 +63,9 @@ if [ "$STRIMZI_KRAFT_ENABLED" = "true" ]; then
     echo "Kraft storage is already formatted"
   fi
 
-  touch /var/opt/kafka/kafka-ready
-  touch /var/opt/kafka/zk-connected
+  rm -f /var/opt/kafka/kafka-ready 2> /dev/null
+    KAFKA_OPTS="${KAFKA_OPTS} -javaagent:$(ls "$KAFKA_HOME"/libs/kafka-agent*.jar)=/var/opt/kafka/kafka-ready"
+    export KAFKA_OPTS
```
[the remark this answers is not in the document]
katheris: No, I'll address it

kafka-agent/src/main/java/io/strimzi/kafka/agent/KafkaAgent.java
katheris: I have no idea why I randomly chose Boolean here, that definitely wasn't an intentional choice, and yeah I think using a final variable would be better, I'll update it

katheris:
@scholzj thanks for your review and comments, I've responded to each. For your more general comments:
> I think the description would deserve a proper explanation of why does it do what it does. For example, it is completely unclear why does controller check only for ports being listened to. While broker checks for additional states. I think it is really important to track the reasoning behind it and how it works. We do not have proposal for this, so we should track it here. If some of that leaks into comments in the shell scripts, I don't think it would hurt either.

I'm assuming by description you mean the commit description? Yes I think you're right that we should make sure it's clear why we've made these choices so we can refer back later.

> You seem to have controller check. You seem to have broker check (although currently not used as we do not support clear brokers or controllers). But it seems to be missing checks for the mixed mode. This is a bit weird since that is the mode that we currently have.

My intention was for us to have a controller check and a broker check separately, then if we are running a combined node (broker and controller), which just run both checks and if they both succeed then the node is marked as ready. So since we currently don't have a mechanism to indicate whether we are running in controller mode, broker mode or combined node I kept the different checks separate, but in the top level KRaft readiness/liveness scripts I just run both checks. My expectation was that at some point that file would include logic to decide which checks we need to run. Does that make sense?

scholzj:
> I'm assuming by description you mean the commit description? Yes I think you're right that we should make sure it's clear why we've made these choices so we can refer back later.

I mean the PR description by it.

> My intention was for us to have a controller check and a broker check separately, then if we are running a combined node (broker and controller), which just run both checks and if they both succeed then the node is marked as ready. So since we currently don't have a mechanism to indicate whether we are running in controller mode, broker mode or combined node I kept the different checks separate, but in the top level KRaft readiness/liveness scripts I just run both checks. My expectation was that at some point that file would include logic to decide which checks we need to run. Does that make sense?

Ok, makes sense. 👍

kafka-agent/src/main/java/io/strimzi/kafka/agent/KafkaAgent.java
```diff
@@ -153,11 +162,19 @@ public void run() {
...
-                Integer running = Integer.valueOf(3);
+                Integer running = 3;
+                Integer unknown = 127;
                 Object value = brokerState.value();
 
-                if ((value instanceof Integer && running.equals(value))
-                        || (value instanceof Byte && running.equals(((Byte) value).intValue()))) {
+                boolean stateIsRunning = false;
+                if (value instanceof Integer) {
+                    stateIsRunning = inKRaftMode ? running.compareTo((Integer) value) <= 0 && !unknown.equals(value) : running.equals(value);
```
showuon, line 171: For the broker readiness probe: `broker state >=3 && broker state != 127`, I think it applied to both ZK and KRaft mode brokers. @tombentley , could you confirm?

kafka-agent/src/main/java/io/strimzi/kafka/agent/KafkaAgent.java
```diff
@@ -208,8 +230,15 @@ private void touch(File file) throws IOException {
...
+            File brokerReadyFile = new File(agentArgs);
+            if (brokerReadyFile.exists() && !brokerReadyFile.delete()) {
+                LOGGER.error("Broker readiness file already exists and could not be deleted: {}", brokerReadyFile);
+                System.exit(1);
+            } else {
+                LOGGER.info("Starting KafkaAgent with brokerReadyFile={}", brokerReadyFile);
+                new KafkaAgent(brokerReadyFile).run();
+            }
         } else {
             File brokerReadyFile = new File(agentArgs.substring(0, index));
```
showuon, line 243: Should we add a debug log here to show `Running in ZK mode` ?

showuon (commented):
Thanks for the PR. Left some comments.

kafka-agent/src/main/java/io/strimzi/kafka/agent/KafkaAgent.java
katheris: I'm not following the comments here, @showuon and @tombentley are you saying that the behaviour I have here is correct, where for KRaft mode we mark a node with `broker state >=3 && broker state != 127` as ready, but with ZK-based broker we only mark ready if `broker state == 3`? Or are you saying I should update the ZK-based check to also check for >=3  && != 127.

Yeah I think those variable renames will make it clearer

tombentley: > I should update the ZK-based check to also check for >=3 && != 127.

That is what I'm saying.

kafka-agent/src/main/java/io/strimzi/kafka/agent/KafkaAgent.java
katheris: Yes, added

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaCluster.java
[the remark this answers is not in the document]
katheris: I've updated this to use a single script for liveness and a single script for readiness.

kafka-agent/src/main/java/io/strimzi/kafka/agent/KafkaAgent.java
katheris: Ok, I've updated the PR to use the same check for both ZK-based and KRaft mode when checking the state and renamed the variables.

katheris:
@showuon @scholzj @tombentley I think I've addressed all of your comments

tombentley (approved):
LGTM

showuon:
@SamBarker @robobario and I found the liveness probe seems not check the broker state < 3, is that right? This is important because this would allow us mark the pod as liveness when broker is doing log recovery, which is broker state 2.

katheris:
@tombentley thinking about this, for liveness of a KRaft broker, we currently check if the state is < 3 or if it is listening on the replication port. Does this mean if the broker has a state of 0 for NOT_RUNNING, then we would still mark it as alive? Should we be verifying that it has a state of greater than 0 to be able to marked as alive?

tombentley:
> Should we be verifying that it has a state of greater than 0 to be able to marked as alive?

Great question. If you look at the [ZK broker code](https://github.com/tombentley/kafka/blob/7d67ddce22cedb48295da16e6b5d4b6800b61d28/core/src/main/scala/kafka/server/KafkaServer.scala#L200) you can see that the transition to STARTING happens very early on during `startup()`. There's very little that could go wrong and what could go wrong will (should) result in an exception and process exit. 

For KRaft there's a lot more that happens (and could possibly go wrong) before we get to `STARTING`. That means we're more likely to observe `NOT_RUNNING` in practice, but that on its own doesn't indicate a problem. Assuming the probe had a `failureThreshold > 1` and didn't have a tiny `initialDelaySeconds` then excluding `NOT_RUNNING` could make sense. But TBH I'm inclined to keep it simple until we've had reports that that doesn't work.

kafka-agent/src/main/java/io/strimzi/kafka/agent/KafkaAgent.java
```diff
@@ -167,13 +185,18 @@ boolean handleBrokerState() {
                     ready = true;
 
                 } else if (i++ % 60 == 0) {
-                    LOGGER.debug("Metric {} = {} (type: {})", brokerStateName, value, value.getClass());
+                    LOGGER.debug("Metric {} = {} (type: {})", brokerStateName, observedState, observedState.getClass());
```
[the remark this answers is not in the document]
katheris: I assume you mean if `brokerState.value()` returns a null value? I can't seem to see whether that can return a null value or not since we are using an abstract type here

docker-images/kafka-based/kafka/scripts/kafka_liveness.sh
```diff
@@ -1,11 +1,19 @@
...
-  # Not yet ready, so test ZK connected state
+elif [ "$STRIMZI_KRAFT_ENABLED" != "true" ]; then
+  # Not yet ready, so if in ZK mode test ZK connected state
   test -f /var/opt/kafka/zk-connected
 fi
+
+if [ "$STRIMZI_KRAFT_ENABLED" = "true" ]; then
+  # Test KRaft controller liveness
+  . ./kafka_controller_liveness_readiness.sh
+fi
```
scholzj, line 19: This seems a bit weird TBH. 

In Zoo mode ... 
* If Kafka is ready, we test port 9091. 
* If it isn't ready, we test if it connected to ZooKeeper

In KRaft mode:
* If Kafka is ready, we test port 9091 and 9090
* If Kafka is not ready, we test only 9090

Does Kafka actually treat these listeners differently? It did not used to in the past and IMHO the original probe is simply using 9091 because that is what we had. 9090 was added only later. does it really make sense to test both ports?

docker-images/kafka-based/kafka/scripts/kafka_readiness.sh
```diff
@@ -0,0 +1,12 @@
+#!/usr/bin/env bash
+set -e
+
+# Test ZK-based broker and KRaft broker readiness
+# The kafka-agent will create /var/opt/kafka/kafka-ready in the container when the broker
+# state is >= 3 && != 127 (UNKNOWN state)
+test -f /var/opt/kafka/kafka-ready
```
scholzj, line 7: Can the state `>= 3 && != 127` be reached without listening on the listeners? Also, in KRaft mode, if it would not be listening on port 9090, I assume the liveness probe will kill it. So that might be a second reason why is it not needed here.

The original assumption of the probes for sure was that the internal Kafka state and the liveness probe determine that listeners are alive and we do not need to check it again.

docker-images/kafka-based/kafka/scripts/kafka_run.sh
```diff
@@ -63,8 +63,9 @@ if [ "$STRIMZI_KRAFT_ENABLED" = "true" ]; then
     echo "Kraft storage is already formatted"
   fi
 
-  touch /var/opt/kafka/kafka-ready
-  touch /var/opt/kafka/zk-connected
+  rm -f /var/opt/kafka/kafka-ready 2> /dev/null
+  KAFKA_OPTS="${KAFKA_OPTS} -javaagent:$(ls "$KAFKA_HOME"/libs/kafka-agent*.jar)=/var/opt/kafka/kafka-ready"
+  export KAFKA_OPTS
```
scholzj, line 68: If we want to keep this code as is, I'm fine and can live with it. But I wonder if it would be nicer to have a flag in the readiness probe to indicate KRaft (or read it from the environment variable) and configure the probe just once. But as I said, if you prefer this method, I'm fine with it.

kafka-agent/src/main/java/io/strimzi/kafka/agent/KafkaAgent.java
```diff
@@ -208,9 +231,17 @@ private void touch(File file) throws IOException {
     public static void premain(String agentArgs) {
         int index = agentArgs.indexOf(':');
         if (index == -1) {
-            LOGGER.error("Unable to parse arguments {}", agentArgs);
-            System.exit(1);
+            LOGGER.debug("Running in KRaft mode");
```
ppatierno, line 234: what do you think about adding a comment on what are the parameters passed to the agent or how it's called in the two different ways. I know it's in the kafka_run script but would be helpful instead of jumping to the script file having some info here.

docker-images/kafka-based/kafka/scripts/kafka_liveness.sh
katheris: I think combining the broker and controller checks in the above makes the comparison seem weird. A fairer comparison is:

In Zoo mode:
 - If Kafka is ready, we test port 9091.
 - If it isn't ready, we test if it connected to ZooKeeper

In KRaft mode:
For brokers:
 - If Kafka is ready, we test port 9091.
 - If it isn't ready we just mark it as alive.
For controllers:
 - We test for port 9090

So my understanding is the 9090 port is specific to KRaft controllers and is not relevant to KRaft brokers. Does that make more sense? I tried to explain it clearly in the PR description.

scholzj: Yeah, but right now this is for a combined node. Do we really want to test both port 9090 and 9091 at the same time on combined nodes?

docker-images/kafka-based/kafka/scripts/kafka_readiness.sh
katheris: The listening check below is to verify that the KRaft controller is ready. My understanding was that if this node was running as a controller only and not a broker, then the broker state metric would not be present. So that is the reason why we have two separate checks. At the moment we run both since we are running in combined mode, but in future we would have an if statement here that selects which checks to run based on the mode the node is running in.

scholzj: Sorry, but I think approach is really weird. I would expect this PR to do one of the following:
* Handle what we have today => in which case it should handle the combined nodes only
* Build for the future => in whcih case there should be a clear path for controller nodes, broker nodes and combined nodes. The paths for the controller and broker nodes will not be used today for anything, but we will need it one day so why not add it now. 

But instead this seems to ship some weird hybrid. We have code which doesn't distinguish between the node types. But it seems to have a code which is apparently there because one day we will have the node types. I think you need to pick one of those options and follow it.

So if this should follow the first option and create a solution only for the combined nodes today ... do we really expect the broker to reach one of the desired states but not listen on 9090?

If this delivers for the future, then there should be some clear code path which distinguishes the different modes and does only what is expected to be done on them.

docker-images/kafka-based/kafka/scripts/kafka_liveness.sh
ppatierno: > If it isn't ready we just mark it as alive

This sounds weird to me but I have not deeper knowledge of brokers in KRaft mode but how can say broker is alive but not ready?

> Yeah, but right now this is for a combined node. Do we really want to test both port 9090 and 9091 at the same time on combined nodes?

Can it happen that a combined node is listening on the replication port 9091 but not on 9090 so it cannot be considered working properly?

katheris: The ports are used for two different things, so I think for a combined mode to work correctly it would need to be listening on both. I don't know enough about how KRaft mode works under the covers to know for sure whether a combined node could get into a state where it has one listener working and one not.

I did a very basic test where I tried starting a combined node with the replication port already in use, and then the controller port already in use and in both cases the combined node fully shuts down. It looks like it attempts to start the controller first, then if that is successful it starts the broker. If the broker then hits an error that causes a shutdown it shuts down first and then the controller shuts down.

If we were to only check one port I guess the replication port would be the one to check, since it appears to listen on that port second.

@tombentley @showuon what do you think?

kafka-agent/src/main/java/io/strimzi/kafka/agent/KafkaAgent.java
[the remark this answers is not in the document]
k-wall: You interpreted by managed comment correctly.   I'd suggest erring on the side of caution and being null safe.

docker-images/kafka-based/kafka/scripts/kafka_liveness.sh
SamBarker: >> If it isn't ready we just mark it as alive
>
> This sounds weird to me but I have not deeper knowledge of brokers in KRaft mode but how can say broker is alive but not ready?

I don't think it's really about KRaft mode, it's more about Kubernetes views on status. My shorthand description:  
- `liveness` is the pod doing something useful. 
- `readiness` can it accept traffic. 

So as an example we would want to mark a Kafka broker as alive in `broker_state == 2` a.k.a LogRecovery but we wouldn't mark it as ready to accept traffic as the listeners are not available yet.

Apologies if this is all to basic I just want to make sure we are all on the same page. 

In combined mode we should probably mark the pod as alive if the controller port is listening and ready when the listener port is available. 

Is port availability enough to mark a controller as ready? Should we be waiting for it become a member of the quorum? Can we tell when it is?

katheris:
As an update to this issue I have worked out I believe why the systemtests have failures.

In combined mode during normal startup the following happens:
 - All the pods are started at the same time
 - The controller processes start listening on 9090 so the first part of the readiness check passes
 - The broker processes start up and start talking to the controller quorum so the controller unfences them and they move to STARTED and therefore the second part of the readiness check passes
 - All pods are now ready
 
 However, if I start the pods initially with affinity constraints so they are all in pending, then remove the constraint this is what happens:
 - The first pod is scheduled
 - The controller process starts up and is listening on 9090 so the first part of the readiness check passes
 - The broker process starts up but since the other controllers are not started yet it does not move to STARTED, it stays in STARTING. This means the second part of the readiness check fails
 - Because this pod doesn't become ready the other pods aren't scheduled, so the pod just sits in crash loop backoff
 
 One workaround I thought of was that we could check for whether the controller is a leader/follower yet, if it is we could only then check for readiness in the broker pod. This would allow the pods to become ready when the broker process isn't up yet but mean once the brokers should be running properly we do still check that they are.

The problem with the proposed workaround is the metric we use to check if the controller is a leader/follower isn't easily accessible to the KafkaAgent.

Another option is in combined mode to only do checks for the broker and controller processes listening on their respective ports, and abandon the check for STARTED.

showuon:
> The problem with the proposed workaround is the metric we use to check if the controller is a leader/follower isn't easily accessible to the KafkaAgent.

Actually, the quorum state info is stored in file within `$LOG_DIR/__cluster_metadata-0/quorum-state`, which containing leader or follower info. Do you think we can read the controller info from there?

I like the idea of checking controller leader/follower status. But I also agree what @scholzj said before:

> In general, the readiness probe is the state of the node not of the KRaft quorum. So just because the other nodes are no ready and there is not enough nodes to form a quorum, it does not mean the node itself is not ready.

So, maybe we just adopt the 2nd option you provided?

scholzj:
So, if I got it right, there is no state to indicate that the controller is _ready_ to be member of the quorum or that it is trying to join / form the quorum (regardless of whether it currently succeeded or not). There is only a state which indicates that it is now part of the quorum?

If that is the case, I don't think it should be part of the readiness or liveness either before or after some event. You would not want to get all pods looping just because the quorum is lost, or? Or at least I would hope that Kafka is smarter than that and can recreate the quorum without restarts.

Also - this is an experience from how ZooKeeper works - once you get all the pods crash-looping, there is a big risk that they never find each other anymore and you will need to go and delete all of them manually to get it fixed. This is because they might never have a time window when they are all running and can find each other. Each of them will run at a separate time with its own crash-loop backoff.

So it is really important that they don't loop forever but rather sit and wait. The readiness probe itself should not restart the pod. But my view is that its semantics indicate that the Pod is ready to work. Not that the quorum is formed. So again, it should be ready when the broker is ready to become member of the quorum and not when it joined the quorum.

Any additional information such as whether it is part of a quorum, I think that can be exposed through some webserver or metric as discussed for example as part of the Kafka roller improvements.

So, if I understood it right, maybe checking if it listens to a port is the best we can do here? It is super lousy liveness and readiness check. But maybe this is something what has to be improved in Kafka and that will not happen in short term. So I guess we should do the best we can today and improve later if needed.

katheris:
> There is only a state which indicates that it is now part of the quorum?

That is my understanding yes.

I'm definitely leaning towards us saying that in combined mode we only check the ports are listening. Although it feels like it isn't a full readiness check for the broker to only check for a port listening, and I guess also for the controller, I know from experience that having a complicated readiness check that is very conditional can cause more problems than it solves.

Separately we need to also decide whether we would include the BrokerState metric in the readiness check for a node that is broker only, but that can be decided outside of this PR.

katheris:
I've raised a proposal to discuss the liveness and readiness probes to try to get to an agreed approach: https://github.com/strimzi/proposals/pull/66

katheris:
Closing this PR. I will open a new PR once the related proposal is approved with changes that match that proposal.

PR #7579, opened 2022-11-05
[systemtest] Make `testUpgradeKafkaWithoutVersion` to run with parameters from yaml

im-konge:

### Type of change

- Enhancement

### Description

Currently the `testUpgradeKafkaWithoutVersion` runs with hardcoded values, which points to Strimzi `0.23.0`. But that version wasn't built for other architectures like `ppc64le` or `s390x`, which makes running this test on different arch environment impossible.

This PR changes the `testUpgradeKafkaWithoutVersion` to run with parameters from upgrade yaml.

As part of this, I added `UpgradeKafkaVersion`, which just represents Kafka with version, LMFV and IBPV. I could go with `TestKafkaVersion`, but that contains a lot more variables than we need and also it's kinda impossible to create that variable with specific values like `null` or `""`, which is needed in our tests.

We will need to refactor the upgrade tests, with all the names of each field in the yamls (and corresponding variable names inside the `UpgradeDowngradeData`), but it's for different PR and it will need some better idea, how to do it.

### Checklist


systemtest/src/main/java/io/strimzi/systemtest/upgrade/UpgradeKafkaVersion.java
```diff
@@ -0,0 +1,71 @@
+/*
+ * Copyright Strimzi authors.
+ * License: Apache License 2.0 (see the file LICENSE or http://apache.org/licenses/LICENSE-2.0.html).
+ */
+package io.strimzi.systemtest.upgrade;
+
+import io.strimzi.systemtest.utils.TestKafkaVersion;
+
+public class UpgradeKafkaVersion {
```
see-quick, line 9: Explaining, what is the purpose of this class would be great :)

see-quick (approved):
LGTM...just one comment and assuming the upgrade will be green. 🤟

Frawless (approved):

PR #7595, opened 2022-11-08
Use the new `operations` field in `KafkaUser` ACLs in our tests

scholzj:
### Type of change

- Bugfix

### Description

In strimzi 0.32.0 we added new field `operations` to `KafkaUser` ACL specification and deprecated the old `operation` field. But we still kept using the `operation` field in User Operator tests and in some System Tests. This PR updates them to use the new field only (apart from the tests designed to test the backwards comaptibility of course).

### Checklist


im-konge (approved):
Thanks for the fix :)

PR #7596, opened 2022-11-08
[systemtest] Add test for upgrading Kafka without LMFV

im-konge:

### Type of change

- New test

### Description

This PR adds test for upgrading Kafka without log message format version, which is not needed to be specified from Kafka `3.0.0`. The tests were already updated, so instead of 3 rolling updates, we are checking just 2, but we didn't add any test for upgrading Kafka without the LMFV.

### Checklist

- [ ] Write tests


Frawless (approved):

see-quick (approved):

im-konge:
OCP: OpenTracingST
image

PR #7597, opened 2022-11-09
Moving from Jaeger exporter to OTLP exporter usage by default

### LINKED ISSUE #7592: Migrating from OpenTelemetry Jaeger exporter to OTLP exporter
Referring to #7570, the current Jaeger Exporter, used by default when OpenTelemetry is enabled, doesn't have support for TLS via SDK environment variables.
At same time, the Jaeger Exporter is going to be deprecated (details [here](https://github.com/open-telemetry/opentelemetry-specification/pull/2858)) and the OpenTelemetry community is pushing for using the OTLP exporter which has TLS support and it's anyway usable with a Jaeger backend which has [support for OTLP protocol](https://www.jaegertracing.io/docs/1.35/apis/#opentelemetry-protocol-stable) since 1.35 version.
We should migrate to OTLP Exporter as soon as possible, which means:

* removing the Jaeger Exporter dependency `opentelemetry-exporter-jaeger` and replacing it with the `opentelemetry-exporter-otlp`
* making the new OTLP Exporter the default one by setting the `OTEL_TRACES_EXPORTER=otlp` in our images script or just leaving that env var empty because it will be the default behavior for OpenTelemetry

The migration should be done before the next release, so it also means that the #7588 can be closed and documentation doesn't need any statement around TLS.

ppatierno:
This PR fixes #7592 moving from the Jaeger exporter to the OLTP exporter usage by default when using OpenTelemetry.

This PR cannot be merged yet and is in draft fot two main reason:

* it needs a new bridge 0.23.0 having OTLP as well, otherwise it will deploy KC, MM and MM2 with OTLP but old bridge 0.22.3 with Jaeger exporter
* the tracing system tests have to be updated by using Jaeger 1.35 as minimum version because it's the first version providing an OTLP endpoint.


CHANGELOG.md
```diff
@@ -2,7 +2,7 @@
 
 ## 0.33.0
 
-* n/a
+* Moved from using the Jaeger exporter to OTLP exporter by default
```
scholzj, line 5: Maybe this deserves a section about deprecations and breaking changes to note the Jaeger OTel client removal? Especially since the user actually needs to change the URL etc where the traces are send or upgrade Jaeger server etc.

docker-images/artifacts/kafka-thirdparty-libs/3.2.x/pom.xml
```diff
@@ -130,14 +130,16 @@
...
+            <artifactId>opentelemetry-exporter-otlp</artifactId>
             <version>${opentelemetry.version}</version>
+            <!--
             <exclusions>
                 <exclusion>
                     <groupId>com.fasterxml.jackson.core</groupId>
                     <artifactId>jackson-core</artifactId>
                 </exclusion>
             </exclusions>
+            -->
```
scholzj, line 142: This should be deleted or user?

docker-images/artifacts/kafka-thirdparty-libs/3.3.x/pom.xml
```diff
@@ -130,14 +130,16 @@
...
+            <artifactId>opentelemetry-exporter-otlp</artifactId>
             <version>${opentelemetry.version}</version>
+            <!--
             <exclusions>
                 <exclusion>
                     <groupId>com.fasterxml.jackson.core</groupId>
                     <artifactId>jackson-core</artifactId>
                 </exclusion>
             </exclusions>
+            -->
```
scholzj, line 142: Same as above

CHANGELOG.md
ppatierno: Yes, good point. What about it now?

docker-images/artifacts/kafka-thirdparty-libs/3.2.x/pom.xml
ppatierno: Yeah, I commented just to check that the build was ok this time but then forgot to remove.

CHANGELOG.md
scholzj: Should it also cover that the OTel Jaeger client is not part of the images anymore and if soeone wants to use it they need to add the binaries?

ppatierno: good point. just added.

scholzj (approved):

ppatierno:
Current status:

* for Kafka Connect and Mirror Maker system tests I already updated them to use the OTLP endpoint as collector on the Jaeger backend side
* for Bridge related test I haven't change the endpoint yet because we need a new bridge release with the OTLP exporter in (so I cannot test this PR yet)
* regarding the used Kafka clients (see `KafkaTracingClients` class), it seems they use the image coming from the `test-clients` project ... can you confirm this @im-konge ? These test clients have to be update by using the `opentelemetry-exporter-otlp` and we should coordinate. If we change it in the test-clients to allow this PR to work (in terms of system tests), then the regression tests could fail for other PRs still using Kafka images with old Jaeger exporter (until this one will be merged and the migration to OTLP exporter finished). Wdyt @scholzj ?

im-konge:
@ppatierno yes, the `KafkaTracingClients` are using the test-clients images, but it has fixed version, so I can add PR for the `opentelemetry-exporter-otlp` and after that do a micro release.
After I will release it, you can use it in this PR
So we will not break anything with PR to the main branch of test-clients

im-konge:
So, after the test-clients will be released (the micro version `0.4.1`), here are all the updates needed in the code to make it work
https://gist.github.com/im-konge/1c512e5592b21e7cea533eab678ebbb7
Maybe the `OTEL_TRACES_EXPORTER` will not be needed, if bridge is able to have `otlp` as default without need of setting that env variable

ppatierno:
@im-konge I have updated the PR using your suggestions. Could you have another pass on the PR please?

ppatierno:
@strimzi-ci run tests test_only profile=regression testcase=OpenTelemetryST,OpenTracingST

im-konge (approved):

ppatierno:
Regression includes the tracing STs and they are green, so we are good to merge. Not sure why the Jenkins one failed. Wdyt @scholzj @im-konge ?

im-konge:
I don't think that it's related to PR, Jenkins was down or something like that in the time you triggered it

I will check it

im-konge:
@ppatierno it failed just because we didn't have Java 17 on our containers, that should be OK now.

ppatierno:
@im-konge thanks for checking! My point was that it's now useless running the Jenkins pipeline because you already ran the regression(s) and they include the tracing STs. Everything is green. So I can merge it.

PR #7599, opened 2022-11-09
[systemtest] Update to Jaeger 1.35.0

### LINKED ISSUE #7598: Upgrading Jaeger backend to 1.35 (at least) on tracing system tests
Related to #7592 and the corresponding opened draft PR #7597 we need the Jaeger backend version used for the tracing tests to be updated to minimum 1.35. Right now it's using 1.20 and 1.25.
This upgrade is needed because 1.35 is the minimum Jaeger backend version having the OTLP endpoint.
We should also check that this Jaeger version still works fine with Kubernetes 1.19.

im-konge:

### Type of change

- Enhancement 

### Description

This PR upgrades to Jaeger `1.35.0`, so we could use OTLP exporter for our tests.
As part of this PR, I removed `1.2x.0` versions from our STs and added cert manager, which is required for running new versions of Jaeger.

Fixes #7598 

### Checklist


see-quick (approved):
LGTM, assuming the tests will pass for all platforms.

systemtest/src/test/java/io/strimzi/systemtest/tracing/TracingAbstractST.java
```diff
@@ -543,26 +545,21 @@ void doTestKafkaBridgeServiceWithHttpTracing(final ExtensionContext extensionCon
...
-                cmdKubeClient(this.namespace).applyContent(yamlContent);
-            }
-        } else {
-            throw new FileNotFoundException("Folder with Jaeger files is empty or doesn't exist");
-        }
+    private void deployJaegerContent() throws IOException {
+        this.certManager = new File(certManagerPath);
+
+        LOGGER.info("Deploying CertManager from {}", certManagerPath);
+        cmdKubeClient("").apply(this.certManager);
```
Frawless, line 556: Why you don't pass namespace name or leave it empty to use namespace from kubeClusterResource? You do it a few lines bellow.

systemtest/src/test/java/io/strimzi/systemtest/tracing/TracingAbstractST.java
```diff
@@ -78,10 +78,12 @@ public abstract class TracingAbstractST extends AbstractST {
 
     private static final Logger LOGGER = LogManager.getLogger(TracingAbstractST.class);
 
-    private Stack<String> jaegerConfigs = new Stack<>();
+    private String jaegerConfigs;
+    private File certManager;
```
Frawless, line 82: You can `create`or `apply` directly based on String path, any specific reason for this? I don't see any changes done in the file so for deletion you can use the same file as well?

Frawless (approved):
Just two nits, more like questions. LGTM otherwise.

systemtest/src/test/java/io/strimzi/systemtest/tracing/TracingAbstractST.java
im-konge: `certManager` is installed in `kube-system` namespace and one more (don't remember the name).
When you specify the namespace there, it will have conflicts and it will simply not be created.
When you leave it empty, there is still used `namespacedCommand`, so `kubeClient().getNamespace()` is used there.
Because I didn't want to make bigger refactoring, I've just pass empty `String` there. I will add comment why I added it there.

systemtest/src/test/java/io/strimzi/systemtest/tracing/TracingAbstractST.java
im-konge: Good point, `certManager` is not necessary there.
`jaegerConfigs` contains modified file, so instead of "creating" it again in deletion method, with all the replacements, I thought it will be simpler to store it like this.

im-konge:
OCP 4.11

OpenTracingST
image

OpenTelemetryST
image

systemtest/src/test/java/io/strimzi/systemtest/tracing/TracingAbstractST.java
im-konge: Removed the `certManager` variable

PR #7603, opened 2022-11-09
System test for cluster role split for cluster wide operator with lim…

see-quick:
…ited access

### Type of change

- Enhancement / new feature
- Refactoring

### Description

This PR aims **(a)** cover https://github.com/strimzi/strimzi-kafka-operator/pull/7390. I have added one test case, which in summary does the following steps:

1. Define our own Roles, ClusterRoles, RoleBindings and ClusterRoleBindings
2. Deploy cluster-wide Cluster Operator
3. During deployment of ClusterOperator, create ClusterRole (-watched) bound to a cluster-wide configuration
4. During deployment of ClusterOperator, create ClusterRole (-namespaced) bound to some specific namespace (i.e., `example-1`)
5. Deploy Kafka cluster into example-namespace
6. Deploy Kafka cluster into infra-namespace
7. Check that Kafka CR is present in infra-namespace but fails (it can be checked via CR Status)
8. Check that Kafka CR is present and successfully deployed in example-namespace 

(b) adding functionality to our SetupClusterOperator (i.e., a user can now add his own Role, RoleBinding, ClusterRole and ClusterRoleBinding resources and pass it via SetupClusterOperator builder. 

### Checklist

- [x] Write tests

### DIFF
--- a/systemtest/src/main/java/io/strimzi/systemtest/resources/operator/SetupClusterOperator.java
+++ b/systemtest/src/main/java/io/strimzi/systemtest/resources/operator/SetupClusterOperator.java
@@ -264,8 +264,8 @@ public class SetupClusterOperator {
         // check if namespace is already created
         createCONamespaceIfNeeded();
         prepareEnvForOperator(extensionContext, namespaceInstallTo, bindingsNamespaces);
-        // if a user want to customise one of the Role, ClusterRole, RoleBindings and ClusterRoleBinding he must do it
-        // everything by himself in scope of RBAC permissions otherwise we apply the default one
+        // if we manage directly in the individual test one of the Role, ClusterRole, RoleBindings and ClusterRoleBinding we must do it
+        // everything by ourselves in scope of RBAC permissions otherwise we apply the default one
         if (this.isRolesAndBindingsManagedByAnUser()) {
             final List<HasMetadata> listOfRolesAndBindings = Stream.of(
                     this.roles, this.roleBindings, this.clusterRoles, this.clusterRoleBindings)
--- a/systemtest/src/main/java/io/strimzi/systemtest/resources/operator/SetupClusterOperator.java
+++ b/systemtest/src/main/java/io/strimzi/systemtest/resources/operator/SetupClusterOperator.java
@@ -275,6 +275,7 @@ public class SetupClusterOperator {
                 ResourceManager.getInstance().createResource(extensionContext, itemRoleOrBinding);
             }
         } else {
+            LOGGER.info("Install default bindings.");
             this.applyDefaultBindings();
         }
 
@@ -570,7 +571,7 @@ public class SetupClusterOperator {
 
             switch (resourceType) {
                 case Constants.ROLE:
-                    if (this.isRolesAndBindingsManagedByAnUser()) {
+                    if (!this.isRolesAndBindingsManagedByAnUser()) {
                         Role role = TestUtils.configFromYaml(createFile, Role.class);
                         ResourceManager.getInstance().createResource(extensionContext, new RoleBuilder(role)
                             .editMetadata()
@@ -580,7 +581,7 @@ public class SetupClusterOperator {
                     }
                     break;
                 case Constants.CLUSTER_ROLE:
-                    if (this.isRolesAndBindingsManagedByAnUser()) {
+                    if (!this.isRolesAndBindingsManagedByAnUser()) {
                         ClusterRole clusterRole = TestUtils.configFromYaml(changeLeaseNameInResourceIfNeeded(createFile.getAbsolutePath()), ClusterRole.class);
                         ResourceManager.getInstance().createResource(extensionContext, clusterRole);
                     }
--- a/systemtest/src/main/java/io/strimzi/systemtest/resources/operator/SetupClusterOperator.java
+++ b/systemtest/src/main/java/io/strimzi/systemtest/resources/operator/SetupClusterOperator.java
@@ -274,23 +274,6 @@ public class SetupClusterOperator {
             for (final HasMetadata itemRoleOrBinding : listOfRolesAndBindings) {
                 ResourceManager.getInstance().createResource(extensionContext, itemRoleOrBinding);
             }
-
-//            // a) deploy specific Roles
-//            for (final Role role : this.roles) {
-//                ResourceManager.getInstance().createResource(extensionContext, role);
-//            }
-//            // b) specific ClusterRole
-//            for (final ClusterRole clusterRole : this.clusterRoles) {
-//                ResourceManager.getInstance().createResource(extensionContext, clusterRole);
-//            }
-//            // c) RoleBindings
-//            for (final RoleBinding roleBinding : this.roleBindings) {
-//                ResourceManager.getInstance().createResource(extensionContext, roleBinding);
-//            }
-//            // d) ClusterRoleBindings
-//            for (final ClusterRoleBinding clusterRoleBinding : this.clusterRoleBindings) {
-//                ResourceManager.getInstance().createResource(extensionContext, clusterRoleBinding);
-//            }
         } else {
             this.applyDefaultBindings();
         }
--- a/systemtest/src/main/java/io/strimzi/systemtest/resources/operator/SetupClusterOperator.java
+++ b/systemtest/src/main/java/io/strimzi/systemtest/resources/operator/SetupClusterOperator.java
@@ -8,6 +8,7 @@ import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
 import io.fabric8.kubernetes.api.model.ConfigMap;
 import io.fabric8.kubernetes.api.model.ConfigMapBuilder;
 import io.fabric8.kubernetes.api.model.EnvVar;
+import io.fabric8.kubernetes.api.model.HasMetadata;
 import io.fabric8.kubernetes.api.model.ServiceAccount;
 import io.fabric8.kubernetes.api.model.ServiceAccountBuilder;
 import io.fabric8.kubernetes.api.model.apiextensions.v1.CustomResourceDefinition;
@@ -46,6 +47,7 @@ import java.io.File;
 import java.io.IOException;
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.Collection;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
@@ -53,6 +55,7 @@ import java.util.Map;
 import java.util.Objects;
 import java.util.function.Predicate;
 import java.util.stream.Collectors;
+import java.util.stream.Stream;
 
 import static io.strimzi.test.k8s.KubeClusterResource.kubeClient;
 import static org.junit.jupiter.api.Assumptions.assumeTrue;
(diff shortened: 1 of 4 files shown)


systemtest/src/test/java/io/strimzi/systemtest/specific/SpecificIsolatedST.java
```diff
@@ -193,6 +207,97 @@ void testLoadBalancerSourceRanges(ExtensionContext extensionContext) {
...
+
+        final List<Role> roles = Arrays.asList(strimziClusterOperator020);
+        final List<ClusterRole> clusterRoles = Arrays.asList(strimziClusterOperator021, strimziClusterOperator022,
+                strimziClusterOperator023, strimziClusterOperator030, strimziClusterOperator031, strimziClusterOperator033);
+
+        // ---- b) defining RoleBindings
+        final RoleBinding strimziClusterOperator020Namespaced = TestUtils.configFromYaml(SetupClusterOperator.getInstanceHolder().switchClusterRolesToRolesIfNeeded(new File(Constants.PATH_TO_PACKAGING_INSTALL_FILES + "/cluster-operator/020-RoleBinding-strimzi-cluster-operator.yaml"), true), RoleBinding.class);
+        final RoleBinding strimziClusterOperator022LeaderElection = TestUtils.configFromYaml(SetupClusterOperator.getInstanceHolder().changeLeaseNameInResourceIfNeeded(new File(Constants.PATH_TO_PACKAGING_INSTALL_FILES + "/cluster-operator/022-RoleBinding-strimzi-cluster-operator.yaml").getAbsolutePath()), RoleBinding.class);
```
im-konge, line 235:
```suggestion
        final RoleBinding strimziClusterOperator022LeaderElection = TestUtils.configFromYaml(SetupClusterOperator.getInstanceHolder().changeLeaseNameInResourceIfNeeded(new File(Constants.PATH_TO_LEASE_ROLE_BINDING).getAbsolutePath()), RoleBinding.class);
```

systemtest/src/test/java/io/strimzi/systemtest/specific/SpecificIsolatedST.java
```diff
@@ -193,6 +207,97 @@ void testLoadBalancerSourceRanges(ExtensionContext extensionContext) {
...
+            .runBundleInstallation();
+
+        // implicit verification that a user is able to deploy Kafka cluster in namespace <example-1>, where we are allowed
+        // to create Custom Resources because of `*-namespaced Role`
+        resourceManager.createResource(extensionContext, false, KafkaTemplates.kafkaEphemeral(testStorage.getClusterName(), 3)
+            .editMetadata()
+            // this should work
+            .withNamespace(namespaceWhereCreationOfCustomResourcesIsApproved)
+            .endMetadata()
+            .build());
```
im-konge, line 279: Indent
Also, if it should work, shouldn't we wait for readiness of Kafka?
Or deploy it in parallel and wait for first Kafka to be ready and second Kafka to contain message about failure?

systemtest/src/test/java/io/strimzi/systemtest/specific/SpecificIsolatedST.java
```diff
@@ -193,6 +207,97 @@ void testLoadBalancerSourceRanges(ExtensionContext extensionContext) {
...
+
+        resourceManager.createResource(extensionContext, false, KafkaTemplates.kafkaEphemeral(testStorage.getClusterName(), 3)
+            .editMetadata()
+                // this should not work
+                .withNamespace(clusterOperator.getDeploymentNamespace())
+            .endMetadata()
+            .build());
+
+        // verify that in `infra-namespace` we are not able to deploy Kafka cluster
+        KafkaUtils.waitForKafkaStatusUpdate(clusterOperator.getDeploymentNamespace(), testStorage.getClusterName());
```
im-konge, line 289: Would it be more stable if we go with `waitUntilKafkaStatusConditionContainsMessage()`?
I'm afraid that there can be few moments when we don't get desired failure message

systemtest/src/test/java/io/strimzi/systemtest/specific/SpecificIsolatedST.java
```diff
@@ -193,6 +207,97 @@ void testLoadBalancerSourceRanges(ExtensionContext extensionContext) {
...
+        final ClusterRole strimziClusterOperator022 = TestUtils.configFromYaml(SetupClusterOperator.getInstanceHolder().changeLeaseNameInResourceIfNeeded(new File(Constants.PATH_TO_PACKAGING_INSTALL_FILES + "/cluster-operator/022-ClusterRole-strimzi-cluster-operator-role.yaml").getAbsolutePath()), ClusterRole.class);
+        final ClusterRole strimziClusterOperator023 = TestUtils.configFromYaml(new File(Constants.PATH_TO_PACKAGING_INSTALL_FILES + "/cluster-operator/023-ClusterRole-strimzi-cluster-operator-role.yaml"), ClusterRole.class);
+        final ClusterRole strimziClusterOperator030 = TestUtils.configFromYaml(new File(Constants.PATH_TO_PACKAGING_INSTALL_FILES + "/cluster-operator/030-ClusterRole-strimzi-kafka-broker.yaml"), ClusterRole.class);
+        final ClusterRole strimziClusterOperator031 = TestUtils.configFromYaml(new File(Constants.PATH_TO_PACKAGING_INSTALL_FILES + "/cluster-operator/031-ClusterRole-strimzi-entity-operator.yaml"), ClusterRole.class);
+        final ClusterRole strimziClusterOperator033 = TestUtils.configFromYaml(new File(Constants.PATH_TO_PACKAGING_INSTALL_FILES + "/cluster-operator/033-ClusterRole-strimzi-kafka-client.yaml"), ClusterRole.class);
```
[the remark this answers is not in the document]
im-konge: Maybe having this kind of method also for `RoleBinding`s would make also sense - then we can remove from list what we don't need

im-konge (commented):
Few comments, but this looks really good :) Thank you!

systemtest/src/main/java/io/strimzi/systemtest/resources/operator/SetupClusterOperator.java
```diff
@@ -253,7 +264,19 @@ private void bundleInstallation() {
         // check if namespace is already created
         createCONamespaceIfNeeded();
         prepareEnvForOperator(extensionContext, namespaceInstallTo, bindingsNamespaces);
-        applyBindings();
+        // if a user want to customise one of the Role, ClusterRole, RoleBindings and ClusterRoleBinding he must do it
+        // everything by himself in scope of RBAC permissions otherwise we apply the default one
```
scholzj, line 268: This is not something a _user_ would do in system tests. Should it instead say _if they are managed directly by the individual test_?

systemtest/src/test/java/io/strimzi/systemtest/specific/SpecificIsolatedST.java
```diff
@@ -193,6 +207,97 @@ void testLoadBalancerSourceRanges(ExtensionContext extensionContext) {
...
+        KafkaUtils.waitForKafkaStatusUpdate(clusterOperator.getDeploymentNamespace(), testStorage.getClusterName());
+
+        final Condition condition = KafkaResource.kafkaClient().inNamespace(clusterOperator.getDeploymentNamespace()).withName(testStorage.getClusterName()).get().getStatus().getConditions().stream().findFirst().get();
+
+        assertThat(condition.getReason(), CoreMatchers.is("KubernetesClientException"));
+        assertThat(condition.getStatus(), CoreMatchers.is("True"));
+        assertThat(condition.getMessage(), CoreMatchers.containsString("Forbidden!Configured service account doesn't have access"));
+
+        // rollback to the default configuration
+        clusterOperator.rollbackToDefaultConfiguration();
```
scholzj, line 298: What exactly does this mean?

scholzj (commented):
I do not see anywhere the deletion of the RBAC resource you create in the test. Are they deleted somewhere automatically?

systemtest/src/test/java/io/strimzi/systemtest/specific/SpecificIsolatedST.java
see-quick: > Also, if it should work, shouldn't we wait for readiness of Kafka?

Yes, that's leftover...

> Or deploy it in parallel and wait for first Kafka to be ready and second Kafka to contain message about failure?

Well, I will re-order the deployment of these Kafka's (and we can achieve that kind of parallelism with additional readiness check).

systemtest/src/test/java/io/strimzi/systemtest/specific/SpecificIsolatedST.java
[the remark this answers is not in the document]
see-quick: Well, in this case, I want explicitly to have ClusterRole and not Role (as we implemented in the SetupClusterOperator).

systemtest/src/main/java/io/strimzi/systemtest/resources/operator/SetupClusterOperator.java
see-quick: Yes, that's better :) I will update it.

systemtest/src/test/java/io/strimzi/systemtest/specific/SpecificIsolatedST.java
see-quick: Well, this actually un-install the Cluster Operator with the specific configuration, which we used in the test case, and then deploys the default Cluster Operator.

see-quick:
> I do not see anywhere the deletion of the RBAC resource you create in the test. Are they deleted somewhere automatically?

That's handled by `clusterOperator.rollbackToDefaultConfiguration();`. It completely uninstalls the current installation of SetupClusterOperator (i.e., custom Roles...) and deploys the one with the default configuration

systemtest/src/test/java/io/strimzi/systemtest/specific/SpecificIsolatedST.java
scholzj: Maybe the comment can be more explicit about it then? How does it know what to uninstall if you install the RBAC bindings in your own way? How do you make sure they are not left behind?

see-quick: > Maybe the comment can be more explicit about it then? 

Yes, I will more elaborate on this comment :) thanks

> How does it know what to uninstall if you install the RBAC bindings in your own way? How do you make sure they are not left behind?

TL;DR; It is handled by ResourceManager inside SetupClusterOperator.
--
When you create a resource by ResourceManager using the generic method `resourceManager.createResource(extensionContent, <resource>)` it will do the following:
1. Find resource type; meaning if such resource is Kafka, KMM2 etc.
2. Create a resource 
3. and then we store info to the `Map<String, Stack<ResourceItem>>`, the key identifier is the test case name and the Stack contains multiple resources, which will be deleted after the test case because we call `ResourceManager.getInstance().deleteResources(extensionContext);` method, which retrieve all of the resources specific to test case thanks to extensionContent metadata.

In this scenario `clusterOperator.rollbackToDefaultConfiguration();` calls at first unInstall() method, which completely deletes all resources (this is handled by ResourceManager) and all namespaces. Thus also removing RBAC bindings deployed in this test case.

scholzj: Ok, thanks for the explanation. 👍

scholzj (approved):

Frawless (approved):

im-konge (approved):

scholzj:
@see-quick Does this also work with Helm / OLM etc.?

PR #7615, opened 2022-11-13
Refactoring listing the artifacts map

ppatierno:
Trivial PR to refactor how going through the artifacts map.


scholzj (approved):
Not sure what value does this add. But I guess why not.

ppatierno:
I see it as more readable and in the end also fixes an IDE warning about listing the map this way.

im-konge (approved):

ppatierno:
Going to merge because the regression failures are not related to these changes. See https://github.com/strimzi/strimzi-kafka-operator/pull/7623

PR #7633, opened 2022-11-15
docs(upgrade): Updates the prereqs for the kafka upgrade and downgrade procedures

PaulRMellor:

**Documentation**

Removes the misleading sentence _"The Cluster Operator, which supports both versions of Kafka, is up and running."_ from the prereqs of [Upgrading Kafka brokers and client applications](https://strimzi.io/docs/operators/in-development/deploying.html#proc-upgrading-brokers-newer-kafka-str) and [Downgrading Kafka brokers and client applications](https://strimzi.io/docs/operators/in-development/deploying.html#proc-downgrading-brokers-older-kafka-str). 

Minor edits to the abstracts or each procedure.

### Checklist

- [ ] Write tests
- [x] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/documentation/modules/upgrading/proc-downgrade-brokers-older-kafka.adoc
+++ b/documentation/modules/upgrading/proc-downgrade-brokers-older-kafka.adoc
@@ -7,16 +7,16 @@
 = Downgrading Kafka brokers and client applications
 
 [role="_abstract"]
-This procedure describes how you can downgrade a Strimzi Kafka cluster to a lower (previous) version of Kafka, such as downgrading from {KafkaVersionHigher} to {KafkaVersionLower}.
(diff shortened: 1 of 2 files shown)


scholzj (approved):
LGTM. I wonder a bit if we should use a weaker statement: `The Cluster Operator, which supports the target versions of Kafka, is up and running.` ... that would be still valid. But I guess it is obvious and does not need to be clarified.

ppatierno (approved):

PaulRMellor:
> LGTM. I wonder a bit if we should use a weaker statement: `The Cluster Operator, which supports the target versions of Kafka, is up and running.` ... that would be still valid. But I guess it is obvious and does not need to be clarified.

I did consider something like that.  But I thought we might be introducing an ambiguity that might raise questions again.

Issue #5680, opened 2021-10-07
Kafka Streams API is not working with tracing and Kafka 3.0.0


im-konge:
When we try to use Kafka Streams application together with Jaeger tracing and Kafka 3.0.0, it always ends with exception:
```
2021-10-05 07:01:55.458.81512692563118 DEBUG InternalTopologyBuilder:1339 - No source topics using pattern subscription found, initializing consumer's subscription collection.
Exception in thread "main" java.lang.UnsupportedOperationException: Implementations of KafkaClientSupplier should implement the getAdmin() method.
	at org.apache.kafka.streams.KafkaClientSupplier.getAdmin(KafkaClientSupplier.java:40)
	at org.apache.kafka.streams.KafkaStreams.<init>(KafkaStreams.java:846)
	at org.apache.kafka.streams.KafkaStreams.<init>(KafkaStreams.java:711)
	at Main.main(Main.java:48)
```

It seems that OpenTracing doesn't support Kafka 3.0.0

im-konge:
PR with adding support for Kafka 3.0.0 opened by @scholzj  [Opentracing-contrib/java-kafka-client#91](https://github.com/opentracing-contrib/java-kafka-client/pull/91).

hsargsyan:
Any updates on this?

scholzj:
Triaged on 21.7.2022: OpenTracing moved into OpenTelemetry and it looks like the PR with the Kafka 3.0 support will not be merged/released anymore. The plan is to migrate to OpenTelemetry which is the replacement for OpenTracing. Strimzi does not provide any direct support for OpenTracing in its operands, it was only used in system tests. This PR should be closed and over time we should migrate to OpenTelemetry.

PR #7639, opened 2022-11-16
[systemtest] Enable KafkaStreams checks for OpenTelemetry and enable both `OpenTracingST` and `OpenTelemetryST` suites after fixes 

im-konge:

### Type of change

- Enhancement 

### Description

Until now, we were skipping checks for `KafkaStreams` in the tracing tests, because of #5680 . This is still the issue, but we can enable those checks for `OpenTelemetry`.

Also, I'm enabling back the `OpenTelemetryST` and `OpenTracingST`. The issue was in deletion of Cert-manager and it's namespace - there were some leftover webhooks from previous instance of Jaeger operator and we didn't delete them (neither the webhook or whole namespace). But, in #7631 I added the whole namespace to our `Stack`, which is being deleted after all tests. This should fix the issues we saw in AZP (and everywhere else).

### Checklist

### DIFF
--- a/systemtest/src/test/java/io/strimzi/systemtest/tracing/OpenTelemetryST.java
+++ b/systemtest/src/test/java/io/strimzi/systemtest/tracing/OpenTelemetryST.java
@@ -8,7 +8,6 @@ import io.strimzi.api.kafka.model.tracing.OpenTelemetryTracing;
 import io.strimzi.api.kafka.model.tracing.Tracing;
 import io.strimzi.systemtest.annotations.ParallelNamespaceTest;
 import io.strimzi.systemtest.annotations.ParallelSuite;
-import org.junit.jupiter.api.Disabled;
 import org.junit.jupiter.api.Tag;
 import org.junit.jupiter.api.extension.ExtensionContext;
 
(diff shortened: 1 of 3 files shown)


scholzj (commented):
The code looks good. But you should probably improve the Pr name to make it a bit more obvious that it enabled the tests again.

im-konge:
> The code looks good. But you should probably improve the Pr name to make it a bit more obvious that it enabled the tests again.

I thought that `enable ... and both tracing suites` will be enough 😄 but I changed the name, hopefully it's more obvious now

scholzj:
> `Enable KafkaStreams checks for OpenTelemetry, OpenTracingST and OpenTelemetryST`

I think this is still not completely clear. The tests never run for OpenTelemetry before and you are not enabling them for OpenTracing as it is not supported. You are also enabling the completely disabled tests (unrelated from Kafka Streams). So ideally, that should be clear to people going through the commit list.

If you already decided to mix this into a single PR, you should name and describe it properly to cover what it does. Even it it means longer name.

ppatierno (approved):

scholzj (approved):

im-konge:
It seems that the tracing tests are working
I will try it once more to see, if it was just luck or it is actually working.

im-konge:
@strimzi-ci run tests test_only profile=regression testcase=OpenTelemetryST,OpenTracingST

Frawless (approved):

PR #7651, opened 2022-11-18
Add option to change default image for docker registry deployed inside minikube

Frawless:

### Type of change

- Enhancement / new feature

### Description

From time to time we face the issue that we are not able to pull image `registry` from dockerhub due to pull rate limits. This PR introduces an option that allows us to pull different image from our private registry. The image will be specified via env variable `MINIKUBE_REGISTRY_IMAGE`. The default value will remain the same (`registry`) so this change shouldn't affect any pipeline on Strimzi azure.

I also added the same option for s390x with env var name `MINIKUBE_REGISTRY_IMAGE_S390X` .

### Checklist

### DIFF
--- a/.azure/scripts/setup-kubernetes.sh
+++ b/.azure/scripts/setup-kubernetes.sh
@@ -4,8 +4,8 @@ set -xe
 rm -rf ~/.kube
 
 KUBE_VERSION=${KUBE_VERSION:-1.19.0}
-REGISTRY_IMAGE=${REGISTRY_IMAGE:-"registry"}
-REGISTRY_IMAGE_S390X=${REGISTRY_IMAGE_S390X:-"s390x/registry:2.8.0-beta.1"}
+MINIKUBE_REGISTRY_IMAGE=${REGISTRY_IMAGE:-"registry"}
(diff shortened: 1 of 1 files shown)


im-konge (approved):

see-quick (approved):

.azure/scripts/setup-kubernetes.sh
```diff
@@ -4,6 +4,8 @@ set -xe
 rm -rf ~/.kube
 
 KUBE_VERSION=${KUBE_VERSION:-1.19.0}
+MINIKUBE_REGISTRY_IMAGE=${REGISTRY_IMAGE:-"registry"}
+MINIKUBE_REGISTRY_IMAGE_S390X=${REGISTRY_IMAGE_S390X:-"s390x/registry:2.8.0-beta.1"}
```
scholzj, line 8: Stupid question ... why do we need two options? Can't the `MINIKUBE_REGISTRY_IMAGE` be used to set the S390x version of the registry when running on s390x?

Frawless: Good point. I will change it.

Frawless: Done

scholzj (approved):

PR #7654, opened 2022-11-18
Add broker certificate annotation per pod in SPS

### LINKED ISSUE #3913: [Bug] Rolling update does not recover and roll all pods after timeout when certificates are updated
**Describe the bug**
The main problem is when we want to run `testClusterCaRemovedTriggersRollingUpdate` with CO's operationTimeout set to 30s. The test itself (the procedure itself too) is successful on OCP3.11, but fails on OCP4.x. So we were forced to changed the timeout of CO to 5mins.

If we delete cluster certs and waiting for ZK rolling update to be done, just one pod is rolled and any Kafka pod is rolled. It's "fixed" when we set the operationTimeout of CO to be more than minute.

Something similar is with `testTriggerRollingUpdateAfterOverrideBootstrap` in `RollingUpdateST`. 

**To Reproduce**
For example run the `testClusterCaRemovedTriggersRollingUpdate` test or try this:

(on OCP4.x)
1. Deploy Strimzi with operationTimeout to 30s
2. Deploy the Kafka cluster with internal/external listeners
3. Add the `kafka-test.XXXX.azure.XXXX.net` to "new bootstrap alternative name"
4. The Kafka pods will not roll in time and this issue happen

**Expected behavior**
Rolling update of ZK and Kafka is successful with `operationTimeout` set to 30s.

 - Strimzi version: all version & master
 - Installation method: yaml
 - Kubernetes cluster: Openshift 4.x

`oc get pod`
```
my-cluster-entity-operator-66c9c4c7cc-ksnxm   3/3       Running   0          10m
my-cluster-kafka-0                            2/2       Running   0          11m
my-cluster-kafka-1                            2/2       Running   0          11m
my-cluster-kafka-2                            2/2       Running   0          11m
my-cluster-kafka-clients-67f75b4c76-vc825     1/1       Running   0          10m
my-cluster-zookeeper-0                        1/1       Running   0          8m55s
my-cluster-zookeeper-1                        1/1       Running   0          12m
my-cluster-zookeeper-2                        1/1       Running   0          12m
strimzi-cluster-operator-7c57c55bbb-29d4l     1/1       Running   0          13m
```

devguyio:
### Type of change

_Select the type of your PR_

- Bugfix

Fixes #3913 

### Description

Add an annotation per pod that contains the broker certificate thumbprint. That way in case of a certificate update, the Kafka Roller would detect if the pods were not rolled consistently since pods would still be annotated by the old certificate thumbprint .

This PR also introduces a small enhancement to avoid duplicating hardcoded key names in the Secret all over the code and encapsulate them inside an enum. Not all code was refactored to use the `SecretEntries` enum to not pollute the PR. 

### Checklist

- [X] Write tests
- [ ] Update documentation
- [ ] Update CHANGELOG.md


cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaCluster.java
```diff
@@ -1282,6 +1287,13 @@ public Secret generateCertificatesSecret(ClusterCa clusterCa, ClientsCa clientsC
         );
     }
 
+    /**
+     * Get the broker certificate information stored in the broker secret.
+     */
+    public CertAndKey getBrokerCertificates(Secret secret, int brokerId) {
+        return  Ca.asCertAndKey(secret, KafkaResources.kafkaPodName(cluster, brokerId));
```
scholzj, line 1294: Just a nit ... but if you do some other changes, maybe we can get rid of the doublespace?

```suggestion
        return Ca.asCertAndKey(secret, KafkaResources.kafkaPodName(cluster, brokerId));
```

certificate-manager/src/main/java/io/strimzi/certs/CertAndKey.java
```diff
@@ -66,4 +67,15 @@ public String storePassword() {
...
+    @Override
+    public String toString() {
+        return "CertAndKey{" +
+                "key=" + Arrays.toString(key) +
+                ", cert=" + Arrays.toString(cert) +
+                ", trustStore=" + Arrays.toString(trustStore) +
+                ", keyStore=" + Arrays.toString(keyStore) +
+                ", storePassword='" + storePassword + '\'' +
+                '}';
+    }
```
scholzj, line 80: Does this possibly lead to leaking sensitive information in logs, errors and exceptions? Would it be better to have some other method used to give you the String for the checksum (or even directly the checksum itself) to avoid the leakage know or by mistake in the future?

Also, does the hash need to be based on all the fields? The pods don't really use all of them for the broker certificates (Just wodnering -> not sure we really shoudln't use all of them).

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaReconciler.java
```diff
@@ -903,11 +927,14 @@ protected Future<Void> podSet() {
                 replicas = kafka.getReplicas();
             }
 
-            StrimziPodSet kafkaPodSet = kafka.generatePodSet(replicas, pfa.isOpenshift(), imagePullPolicy, imagePullSecrets, brokerId -> kafkaPodAnnotations(brokerId, false));
-            return strimziPodSetOperator.reconcile(reconciliation, reconciliation.namespace(), KafkaResources.kafkaStatefulSetName(reconciliation.name()), kafkaPodSet)
-                    .compose(rr -> {
-                        podSetDiff = rr;
-                        return Future.succeededFuture();
+            return secretOperator.getAsync(reconciliation.namespace(), KafkaResources.kafkaSecretName(reconciliation.name()))
```
scholzj, line 930: Is the secret not available in the KafkaCluster class somewhere? Do we really need to get it from Kube API? (I'm not saying on 100% it is ... just wondering).

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaReconciler.java
```diff
@@ -1102,10 +1129,14 @@ protected Future<Void> scaleUp() {
             //   => we need to do scale-up
             LOGGER.infoCr(reconciliation, "Scaling Kafka up from {} to {} replicas", currentReplicas, kafka.getReplicas());
 
-            if (featureGates.useStrimziPodSetsEnabled())   {
-                StrimziPodSet kafkaPodSet = kafka.generatePodSet(kafka.getReplicas(), pfa.isOpenshift(), imagePullPolicy, imagePullSecrets, brokerId -> kafkaPodAnnotations(brokerId, false));
-                return strimziPodSetOperator.reconcile(reconciliation, reconciliation.namespace(), KafkaResources.kafkaStatefulSetName(reconciliation.name()), kafkaPodSet)
-                        .map((Void) null);
+            if (featureGates.useStrimziPodSetsEnabled()) {
+
+                return secretOperator.getAsync(reconciliation.namespace(), KafkaResources.kafkaSecretName(reconciliation.name()))
```
scholzj, line 1134: Same as above.

scholzj (commented):
I left some nits, but it looks mostly good. I also don't see any test actually covering this, only updates to the existing tests. Should we have one? Or do you think it is not needed?

devguyio:
> I left some nits, but it looks mostly good. I also don't see any test actually covering this, only updates to the existing tests. Should we have one? Or do you think it is not needed?

I think I should indeed add a test for this. Will work on it. The existing tests will fail in case the secret object wasn't reconciled and if it had garbage data, but there're no tests that verify cases like "one broker certificate changed, make sure it's rolled" etc.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaReconciler.java
```diff
@@ -810,12 +810,41 @@ protected Future<Void> podDisruptionBudgetV1Beta1() {
...
      *
-     * @return  Map with Pod annotations
+     * @return Map with Pod annotations
      */
-    private Map<String, String> kafkaPodAnnotations(int brokerId, boolean storageAnnotation)    {
+    private Map<String, String> podSetPodAnnotations(int brokerId, Secret kafkaSecret) {
+        Map<String, String> podAnnotations = commonKafkaPodAnnotations(brokerId);
+
+        // Annotation of broker certificate hash
+        podAnnotations.put(KafkaCluster.ANNO_STRIMZI_BROKER_CERT_HASH, Util.hashStub(kafka.getBrokerCertificates(kafkaSecret, brokerId).toString()));
```
tombentley, line 822: It's a bit weird to use the `toString` to calculate the hash here. If I understand the problem correctly we only _really care_ when the certificate itself changes (true, right now the rest of the CertAndKey only changes when the cert changes, but let's not build in that assumption). I think it would be better to use the actual broker "thumbprint". As in:

* Get the certificate's ASN1 encoded form (which you can get from `X509Certificate#getEncoded()`)
* Hash it using SHA1.

This is the same thumbprint/fingerprint that things like openssl use so it's actually meaningful more broadly than what we're using here

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaCluster.java
```diff
@@ -165,6 +165,11 @@ public class KafkaCluster extends AbstractModel {
      */
     public static final String ANNO_STRIMZI_BROKER_CONFIGURATION_HASH = Annotations.STRIMZI_DOMAIN + "broker-configuration-hash";
 
+    /**
+     * Records the hash of the Kafka broker. Each pod will have a different hash since each broker has a different certificate.
```
tombentley, line 169: > Records the hash of the Kafka broker.

I don't know what that's supposed to mean.

tombentley (commented):
Thanks, I think this is along the right lines.

ppatierno:
I have a bad memory but I remember that when working on a related thing, I started with a thumbprint and then ended with using the certificate generation annotation that was enough. So I was wondering why the generation was not enough and we need one more annotation.

tombentley:
@ppatierno While that works for the CA certs (which IIRC live in their own Secret) the broker end entity certs all currently live in a single `Secret`, so it's not very specific to an individual broker.

operator-common/src/main/java/io/strimzi/operator/cluster/model/Ca.java
```diff
@@ -67,6 +67,28 @@
...
+        public static String getDataForPod(Secret secret, String podName, SecretEntry entry) {
+            return secret.getData().get(getNameForPod(podName, entry));
+        }
+
+        public static String getNameForPod(String podName, SecretEntry entry) {
+            return podName + entry.suffix;
+        }
+
+    }
+
```
ppatierno, line 91: please add Javadoc on the enum declaration, to all the enum values and public methods. We are going to have strict checks when moving to Java 17 so it's better to address missing doc right now. Thanks!

operator-common/src/main/java/io/strimzi/operator/cluster/model/Ca.java
```diff
@@ -439,6 +456,20 @@ protected Map<String, CertAndKey> maybeCopyOrGenerateCerts(
...
     }
 
+    private boolean isNewVersion(Secret secret, String podName) {
+        String store = SecretEntry.getDataForPod(secret, podName, SecretEntry.P12_KEYSTORE);
+        String password = SecretEntry.getDataForPod(secret, podName, SecretEntry.P12_KEYSTORE_PASSWORD);
+
+        return store != null && !store.isEmpty() && password != null && !password.isEmpty();
+    }
+
+    public static CertAndKey asCertAndKey(Secret secret, String podName) {
```
ppatierno, line 466: Javadoc (see my comment about Java 17)

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaReconciler.java
devguyio: I can't find it in the KafkaCluster class

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaCluster.java
```diff
@@ -165,6 +165,11 @@ public class KafkaCluster extends AbstractModel {
...
+     */
+    public static final String ANNO_STRIMZI_BROKER_CERT_HASH = Annotations.STRIMZI_DOMAIN + "broker-certs-hash";
```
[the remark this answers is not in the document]
devguyio: I agree that we need to review the terminology to account for KRaft , I'd be happy to create an issue to track this

cluster-operator/src/test/java/io/strimzi/operator/cluster/ResourceUtils.java
```diff
@@ -193,6 +195,27 @@ public static Kafka createKafka(String namespace, String name, int replicas,
                 .build();
     }
 
+    public static Secret createClusterSecret(String clusterNamespace, String clusterName, int replicas, String secretName,
+                                             String caCert, String caKey, String caStore, String caStorePassword) {
+        Map<String, String> data = new HashMap<>(replicas * 4);
```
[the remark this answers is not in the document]
devguyio: can you please elaborate more? not sure what you mean by the load factor here

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaReconciler.java
devguyio: Done

devguyio:
I've addressed the review comments, can you please take another look?

cluster-operator/src/test/java/io/strimzi/operator/cluster/ResourceUtils.java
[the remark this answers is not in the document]
tombentley: By default `java.util.HashMap` has a load factor of 0.75. The Javadocs describe the relationship between capacity and load factor like this:

>  When the number of entries in the hash table exceeds the product of the load factor and the current capacity, the hash table is rehashed (that is, internal data structures are rebuilt) so that the hash table has approximately twice the number of buckets. 

So if you're expecting `data` to end up with `replicas * 4` keys in it and you asked for an initial capacity of `replicas * 4` then while populating `data` you'll end up rehashing once you've added `0.75 * replicas * 4` keys. To avoid rehashing the initial capacity would need to be `ceil(replicas * 4 / 0.75)`, or `(int) (1 + replicas * 4 /0.75)` (i.e. I got it wrong too in my initial comment).

This is something that we get wrong lots elsewhere in Strimzi. If we were serious about this we'd add a utility function to encapsulate the logic.

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaCluster.java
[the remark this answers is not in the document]
tombentley: Perhaps there's value in creating an issue to look at existing usage, but here you're adding a new annotation, which is a public API of sorts, using the term "broker" which isn't the right term for what we're doing here and now. Can we change this to `server-certs-hash`?

operator-common/src/main/java/io/strimzi/operator/cluster/model/Ca.java
```diff
@@ -1187,4 +1252,33 @@ protected boolean hasCaCertGenerationChanged(Secret secret) {
...
+     *
+     * @param certSecret    Secrets with the certificate
+     * @param key           Key under which the certificate is stored in the Secret
+     *
+     * @return              Hash stub of the certificate
+     */
+    public static String getCertificateThumbprint(Secret certSecret, String key) {
+        try {
+            X509Certificate cert = Ca.cert(certSecret, key);
+            return Util.hashStub(cert.getEncoded());
```
tombentley, line 1267: IIRC (but might be wrong) the OpenSSL fingerprint is the whole SHA1 of the DER, not just the prefix. If that's correct let's avoid using a truncation here and return the whole thing.

tombentley (approved):
A couple of nits, but otherwise LGTM. Thanks!

operator-common/src/main/java/io/strimzi/operator/cluster/model/Ca.java
devguyio: I'm fine with that.

Note that this is already in use for the `ANNO_STRIMZI_CUSTOM_LISTENER_CERT_THUMBPRINTS` annotation, but I don't think this would a very far reaching breaking change. Users will end up with bigger value in that annotation after upgrade , so pods will be rolled even though the certs are the same, but that's just during the upgrade.

cluster-operator/src/test/java/io/strimzi/operator/cluster/ResourceUtils.java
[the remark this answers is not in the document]
devguyio: Thanks! I found another constructor which accepts a load factor. That sounds more readable in this case IMHO.

`new HashMap<>(replicas * 4, 1)`

wdyt?

operator-common/src/main/java/io/strimzi/operator/cluster/model/Ca.java
devguyio: @tombentley after some discussion with @scholzj , it seems using the short 8-characters SHA1 hash is a pattern used in multiple places in the project, and here it might have the added benefit of easier diffing when changed, is there a particular benefit that you see that would make us not use `hashStub` here and just generate the full SHA1 hash?

devguyio:
@scholzj @ppatierno a gentle reminder

devguyio:
is it expected to update CHANGELOG.md for such a change (i.e. adding a new pod annotation) ?

scholzj:
> is it expected to update CHANGELOG.md for such a change (i.e. adding a new pod annotation) ?

No, this would not be something for the changelog from my point of view.

operator-common/src/main/java/io/strimzi/operator/cluster/model/Ca.java
tombentley: > is there a particular benefit that you see that would make us not use `hashStub` here

Only to the extent that a `SHA1(DER)` is a thing which is spat out by other tools in the SSL world, so there is some very minor utility being consistent with that. Using a truncated version of that still works, but it's no longer "the certificate fingerprint". I can live with it either way.

operator-common/src/main/java/io/strimzi/operator/cluster/model/Ca.java
```diff
@@ -1187,4 +1252,33 @@ protected boolean hasCaCertGenerationChanged(Secret secret) {
...
+
+    /**
+     * Generates hash stub of the broker certificate which is used to track when the certificate changes and rolling update needs to be triggered.
+     *
+     * @param certSecret    Secrets with the certificate
+     * @param podName       Broker pod name
+     *
+     * @return              Hash stub of the certificate
+     */
+    public static String getCertificateThumbprintForBroker(Secret certSecret, String podName) {
```
scholzj, line 1281: Same as above. Should it really be here? Given it is directly ried to brokers, then maybe if nothing else `ClusterCa` is a better place for this? But strictly speaking, it might not be that much CA related as just a utility method?

operator-common/src/main/java/io/strimzi/operator/cluster/model/Ca.java
```diff
@@ -512,6 +558,25 @@ protected Map<String, CertAndKey> maybeCopyOrGenerateCerts(
...
+        String password = SecretEntry.getDataForPod(secret, podName, SecretEntry.P12_KEYSTORE_PASSWORD);
+
+        return store != null && !store.isEmpty() && password != null && !password.isEmpty();
+    }
+
+    /** Return given secret for pod as a CertAndKey object
+     * @param secret Kubernetes Secret
+     * @param podName Name of the pod
+     * @return CertAndKey instance
+     */
```
scholzj, line 572: A nit, but maybe we can keep the formatting as for other Javadoc comments?

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaReconciler.java
scholzj: The `certificateSecret` is patching the secret. It even already has some logic to decide whether the certificates changed or not. So I guess the secret has to be known there and you can possibly just store the hash stubs there?

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaReconciler.java
scholzj: Same as above.

operator-common/src/test/java/io/strimzi/operator/common/operator/MockCertManager.java
```diff
@@ -122,16 +122,104 @@ public class MockCertManager implements CertManager {
             "QtlIdmFljGSaGGY6aJjUvUdgoPp1yQPa5oS+afr5g9gaEp4lxP6mc+Li\n" +
             "-----END CERTIFICATE-----\n";
 
+    private static final String ENTITY_CERT = "-----BEGIN CERTIFICATE-----\n" +
```
scholzj, line 125: What does `ENTITY` stand for here?

operator-common/src/test/java/io/strimzi/operator/common/operator/MockCertManager.java
```diff
@@ -122,16 +122,104 @@ public class MockCertManager implements CertManager {
...
+            "T8efO2ITPIdPjSQnQtTdIrJzh46JRmXRM+3EEyb94BnvR7BB9Brqam5BIHxmgpDK\n" +
+            "4QTF7I+K4jclRwlR0aJmtW5PqEi7ZSsdY9/xWYMjrtblkCiN5ukzkcFZFFolWJvN\n" +
+            "J8IkceVTKw9KzpUNxORDj8pXOO2SCqReu4u3yPiYy7Eed7zTgmmDIGcQdy4kmAvc\n" +
+            "hqsCggEAWs3tK7cgOYJbbPGrW8rvtneK8+BH+15fyFjEZLv+0kDAu3pux5CBhsn8\n" +
+            "xHYe5sTUJnSMv4RFnGN777R56NKdqz+g6cOBwX8Wa9HJgUv37xQ0y+bpWwRGEcWa\n" +
+            "b40yeuAnBPwaGFwfm+IvB/h/Dovblwb8IDhQ3H0A/Xw94HWhEDhOfnDozd2zSaIS\n" +
+            "gPuuXmbT+5Nnefg6DupRxhOLUASG5eyaIJKGaBvoc2Ylmq5BXPyCnMIoUNS9e85q\n" +
+            "tUt1ndVBV2e7HV4BMXB6cCvE62H2ks8WoytpxOyg8GfwIeePe+seJOQAj6HuG4R8\n" +
+            "nEhbTwU8z4HvDeru2iLQfOv0JyKaXg==\n" +
+            "-----END PRIVATE KEY-----";
```
scholzj, line 208: Maybe you can give it a better validity? Eg. 100 years as with the other certs here?

operator-common/src/test/java/io/strimzi/operator/common/operator/MockCertManager.java
```diff
@@ -235,7 +336,8 @@ public void addCertToTrustStore(File certFile, String certAlias, File trustStore
     @Override
     public void addKeyAndCertToKeyStore(File keyFile, File certFile, String alias, File keyStoreFile, String keyStorePassword) throws IOException {
         // never called during the tests which use this MockCertManager
-        write(keyStoreFile, "key store");
+        //write(keyStoreFile, "key store");
+        Files.write(keyStoreFile.toPath(), ENTITY_CERT_STORE);
```
scholzj, line 340: Is this really needed? The commented out original code makes it confusing (and if nothing else, it should not be here I guess).

But also, your certificate is not a valid proper certificate for where this is used. So I wonder what advantages does this really bring since this is something what can potentially cause a lot of confusion later. While `key store` is obviously dummy certificate, an actual certificate might not be so clear as dummy value.

operator-common/src/test/java/io/strimzi/operator/common/operator/MockCertManager.java
```diff
@@ -253,17 +355,17 @@ public void deleteFromTrustStore(List<String> aliases, File trustStoreFile, Stri
      */
     @Override
     public void generateCsr(File keyFile, File csrFile, Subject sbj) throws IOException {
-        write(keyFile, "key file");
-        write(csrFile, "csr file");
+        write(keyFile, ENTITY_KEY);
+        write(csrFile, ENTITY_CERT);
```
scholzj, line 359: Same as above.

operator-common/src/test/java/io/strimzi/operator/common/operator/MockCertManager.java
```diff
@@ -253,17 +355,17 @@ public void deleteFromTrustStore(List<String> aliases, File trustStoreFile, Stri
...
     @Override
     public void generateCert(File csrFile, File caKey, File caCert, File crtFile, Subject sbj, int days) throws IOException {
-        write(crtFile, "crt file");
+        write(crtFile, ENTITY_CERT);
     }
 
     @Override
     public void generateCert(File csrFile, byte[] caKey, byte[] caCert, File crtFile, Subject sbj, int days) throws IOException {
-        write(crtFile, "crt file");
+        write(crtFile, ENTITY_CERT);
```
scholzj, line 369: Same as above.

user-operator/src/test/java/io/strimzi/operator/user/model/KafkaUserModelTest.java
```diff
@@ -226,9 +226,9 @@ public void testGenerateSecretGeneratesCertificateWhenNoSecretExistsProvidedByUs
         Secret generated = model.generateSecret();
 
         assertThat(new String(model.decodeFromSecret(generated, "ca.crt")), is("clients-ca-crt"));
-        assertThat(new String(model.decodeFromSecret(generated, "user.crt")), is("crt file"));
-        assertThat(new String(model.decodeFromSecret(generated, "user.key")), is("key file"));
-        assertThat(new String(model.decodeFromSecret(generated, "user.p12")), is("key store"));
+        assertThat(generated.getData().get("user.crt"), is(MockCertManager.entityCert()));
+        assertThat(generated.getData().get("user.key"), is(MockCertManager.entityKey()));
+        assertThat(generated.getData().get("user.p12"), is(MockCertManager.entityCertStore()));
```
scholzj, line 231: This nicely demonstrates the points I made on `MockCertManager`. You replaced something clearly mocked with something what is not valid anyway, but in a very sneaky way.

operator-common/src/main/java/io/strimzi/operator/cluster/model/Ca.java
```diff
@@ -67,6 +68,56 @@
 @SuppressWarnings("checkstyle:CyclomaticComplexity")
 public abstract class Ca {
 
+    /**
+     * A certificate entry in a ConfigMap. Each entry contains an entry name and data.
+     */
+    public enum SecretEntry {
```
scholzj, line 74: What is the value we get here from using Enum instead of just constants?

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaCluster.java
```diff
@@ -177,6 +177,11 @@ public class KafkaCluster extends AbstractModel {
      */
     public static final String ANNO_STRIMZI_BROKER_CONFIGURATION_HASH = Annotations.STRIMZI_DOMAIN + "broker-configuration-hash";
 
+    /**
+     * Annotation for keeping Kafka servers' certificate thumbprints.
+     */
+    public static final String ANNO_STRIMZI_SERVER_CERT_HASH = Annotations.STRIMZI_DOMAIN + "server-certs-hash";
```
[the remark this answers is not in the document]
devguyio: ack, this was overlooked

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaReconciler.java
devguyio: I've changed this to cache the cert hash in a HashMap that gets populated once when the secret is retrieved.

operator-common/src/main/java/io/strimzi/operator/cluster/model/Ca.java
devguyio: changed to the full sha1.

operator-common/src/main/java/io/strimzi/operator/cluster/model/Ca.java
```diff
@@ -1187,4 +1252,33 @@ protected boolean hasCaCertGenerationChanged(Secret secret) {
...
+
+    /**
+     * Generates hash stub of the certificate which is used to track when the certificate changes and rolling update needs to be triggered.
+     *
+     * @param certSecret    Secrets with the certificate
+     * @param key           Key under which the certificate is stored in the Secret
+     *
+     * @return              Hash stub of the certificate
+     */
+    public static String getCertificateThumbprint(Secret certSecret, String key) {
```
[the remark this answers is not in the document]
devguyio: Done

operator-common/src/main/java/io/strimzi/operator/cluster/model/Ca.java
devguyio: Done

operator-common/src/main/java/io/strimzi/operator/cluster/model/Ca.java
devguyio: An Enum has two added benefits:
- It acts as a set of the possible keys in the secret. 
- It has methods which encapsulate the logic of the mapping of the secret key names to the broker.

operator-common/src/test/java/io/strimzi/operator/common/operator/MockCertManager.java
```diff
@@ -23,110 +23,220 @@
...
-            "W/8LMGkSFfO4quAJAgMBAAGjUzBRMB0GA1UdDgQWBBQUwNmfsNj+PM240pVPxYx9\n" +
-            "Q9eQhDAfBgNVHSMEGDAWgBQUwNmfsNj+PM240pVPxYx9Q9eQhDAPBgNVHRMBAf8E\n" +
-            "BTADAQH/MA0GCSqGSIb3DQEBCwUAA4IBAQAnhbNKwMmnayHsT6kKgyyDV6RUUYs6\n" +
-            "nYf3nx+GIQWSw4c5TOHDcTWdKpOxVnLNXYKQoSkb1RBoSMLBdQwidZ5K2DB5eXaG\n" +
-            "rcfEbKNBc5ZCFgFEAyy35pitJOmU/KzCdKyvx+TR5hIgGoKajYX5JZxj+1rTPGKO\n" +
-            "ePT9iFp1ZbzHjgw6vFeJ+D2ov6HfW6C/KuK9Y6xUpvRQLVjMJYCyzxkxQAxZvu/0\n" +
-            "0HVYYH6UJ7kuWywFMWoBdZ8US/vuUSBYyCGNL9p6ol+h9rsz3cIWBVBjx8C3qKki\n" +
-            "QtlIdmFljGSaGGY6aJjUvUdgoPp1yQPa5oS+afr5g9gaEp4lxP6mc+Li\n" +
-            "-----END CERTIFICATE-----\n";
+    private static final String CLUSTER_KEY = """
```
devguyio, line 26: I changed these to a text block since I was adding `END_ENTITY_CERT` and it made sense to be consistent.

operator-common/src/test/java/io/strimzi/operator/common/operator/MockCertManager.java
```diff
@@ -234,14 +344,14 @@ public void addCertToTrustStore(File certFile, String certAlias, File trustStore
 
     @Override
     public void addKeyAndCertToKeyStore(File keyFile, File certFile, String alias, File keyStoreFile, String keyStorePassword) throws IOException {
-        // never called during the tests which use this MockCertManager
-        write(keyStoreFile, "key store");
+        write(keyStoreFile, END_ENTITY_KEY_STORE);
```
devguyio, line 347: @scholzj the string was split into a constant to be consistent with the rest. 

For MockCertManager, a real Certificate is needed for the cluster operator tests to pass since the annotation generation uses the X509Certificate class. 

And since a real cert is used, it made sense to also use the real corresponding key to not lose it since it shouldn't have any impact on the tests performance.

The real key store is the one that's not needed anywhere so instead of wringing and reading a binary keystore I kept it as a string.

user-operator/src/test/java/io/strimzi/operator/user/model/KafkaUserModelTest.java
```diff
@@ -226,9 +226,9 @@ public void testGenerateSecretGeneratesCertificateWhenNoSecretExistsProvidedByUs
         Secret generated = model.generateSecret();
 
         assertThat(new String(model.decodeFromSecret(generated, "ca.crt")), is("clients-ca-crt"));
-        assertThat(new String(model.decodeFromSecret(generated, "user.crt")), is("crt file"));
-        assertThat(new String(model.decodeFromSecret(generated, "user.key")), is("key file"));
-        assertThat(new String(model.decodeFromSecret(generated, "user.p12")), is("key store"));
+        assertThat(new String(model.decodeFromSecret(generated, "user.crt")), is(MockCertManager.userCert()));
```
devguyio, line 229: > This nicely demonstrates the points I made on MockCertManager. You replaced something clearly mocked with something what is not valid anyway, but in a very sneaky way.

@scholzj for this comment , I'm hoping that `MockCertManager.userCert()` makes it clear this is a mock, given that having a real certificate is needed so I opted in for unifying how the tests get the expected values from the MockCertManager instead of hardcoding them sometimes and using getters in others.

operator-common/src/test/java/io/strimzi/operator/common/operator/MockCertManager.java
devguyio: It's now called `END_ENTITY_CERT` which would be reused as for user cert and broker cert. I added a comment to clarify.

operator-common/src/test/java/io/strimzi/operator/common/operator/MockCertManager.java
devguyio: Done

operator-common/src/main/java/io/strimzi/operator/common/Util.java
```diff
@@ -556,12 +556,20 @@ public static String hashStub(String toBeHashed)   {
...
+        byte[] digest = getSHA1(toBeHashed);
+        return String.format("%040x", new BigInteger(1, digest)).substring(0, 8);
+    }
+
+    /**
+     * Get a SHA-1 hash of the provided byte array
+     * @param toBeHashed    Byte array for which the hash will be returned
+     * @return              SHA-1 hash
+     */
+    public static byte[] getSHA1(byte[] toBeHashed) {
```
scholzj, line 568: It is not a getter. So you can maybe call it `sha1Digest()`? Assuming this method is really needed.

operator-common/src/main/java/io/strimzi/operator/cluster/model/CertUtils.java
```diff
@@ -0,0 +1,55 @@
...
+            X509Certificate cert = Ca.cert(certSecret, Ca.SecretEntry.getNameForPod(podName, Ca.SecretEntry.CRT));
+            if (cert != null) {
+                hash = String.format("%040x", new BigInteger(1, Util.getSHA1(cert.getEncoded())));
+            }
+            return hash;
+        } catch (CertificateEncodingException e) {
+            throw new RuntimeException("Failed to get certificate thumbprint of " + podName + " from Secret " + certSecret.getMetadata().getName(), e);
+        }
+    }
+}
```
scholzj, line 55: Why do we need this in addition to `getCertificateThumbprint`? It seems to do basically the same. It can easily just call `getCertificateThumbprint` => either translate the pod name into the key here and delegate or just do it directly in the two places this method is now called.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaReconciler.java
```diff
@@ -810,12 +815,41 @@ protected Future<Void> podDisruptionBudgetV1Beta1() {
     }
 
     /**
-     * Prepares annotations for Kafka pods which are known only in the KafkaAssemblyOperator level. These are later
-     * passed to KafkaCluster where there are used when creating the Pod definitions.
+     * Prepares annotations for Kafka pods within a StrimziPodSet which are known only in the KafkaAssemblyOperator level.
+     * These are later passed to KafkaCluster where there are used when creating the Pod definitions.
      *
-     * @return  Map with Pod annotations
+     * @return Map with Pod annotations
```
scholzj, line 821: A nit ... but maybe you can add the missing param description to the Javadoc?

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaReconciler.java
```diff
@@ -810,12 +815,41 @@ protected Future<Void> podDisruptionBudgetV1Beta1() {
...
+        podAnnotations.put(ANNO_STRIMZI_IO_STORAGE, ModelUtils.encodeStorageToJson(kafka.getStorage()));
+
+        return podAnnotations;
+    }
+
+    /**
+     * Prepares the Kafka pods' annotations that are common between StatefulSets and StrimziPodSets.
+     *
+     * @return Map with Pod annotations
+     */
```
scholzj, line 851: A nit ... but maybe you can add the missing param description to the Javadoc?

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaReconciler.java
```diff
@@ -1101,10 +1130,14 @@ protected Future<Void> scaleUp() {
...
-                return strimziPodSetOperator.reconcile(reconciliation, reconciliation.namespace(), KafkaResources.kafkaStatefulSetName(reconciliation.name()), kafkaPodSet)
-                        .map((Void) null);
+            if (featureGates.useStrimziPodSetsEnabled()) {
+
+                return secretOperator.getAsync(reconciliation.namespace(), KafkaResources.kafkaSecretName(reconciliation.name()))
+                        .compose(secret -> {
+                            StrimziPodSet kafkaPodSet = kafka.generatePodSet(kafka.getReplicas(), pfa.isOpenshift(), imagePullPolicy, imagePullSecrets, this::podSetPodAnnotations);
+                            return strimziPodSetOperator.reconcile(reconciliation, reconciliation.namespace(), KafkaResources.kafkaStatefulSetName(reconciliation.name()), kafkaPodSet)
+                                    .map((Void) null);
+                        });
```
scholzj, line 1140: Why is this change needed? You do not seem to use the `secret` for anything here. So not sure why it needs to be queried.

operator-common/src/main/java/io/strimzi/operator/cluster/model/Ca.java
```diff
@@ -67,6 +67,56 @@
...
+
+        /**
+         * Get the name of secret entry of given SecretEntry type for podName
+         * @param podName Name of the pod which secret entry is looked for
+         * @param entry The SecretEntry type
+         * @return the name of the secret entry
+         */
+        public static String getNameForPod(String podName, SecretEntry entry) {
+            return podName + entry.suffix;
+        }
```
scholzj, line 116: It seems a bit weird to have these in the Enum. Would it make more sense to have them in the Ca class?

operator-common/src/main/java/io/strimzi/operator/cluster/model/CertUtils.java
```diff
@@ -0,0 +1,55 @@
...
+import io.strimzi.operator.common.Util;
+
+import java.math.BigInteger;
+import java.security.cert.CertificateEncodingException;
+import java.security.cert.X509Certificate;
+
+/**
+ * Certificate utility methods
+ */
+public class CertUtils {
```
scholzj, line 17: Any reason this is in `operator-common` and not in `cluster-operator`? Does not seem to be used anywhere else then in `cluster-operator` from a quick look.

operator-common/src/test/resources/END_ENTITY_CERT.pem
```diff
@@ -0,0 +1,25 @@
+-----BEGIN CERTIFICATE-----
```
scholzj, line 1: Isn't this the same as what you have in the constants in `MockCertManager`? Why do we keep it twice?

operator-common/src/test/resources/END_ENTITY_KEY.pem
```diff
@@ -0,0 +1,52 @@
+-----BEGIN PRIVATE KEY-----
```
scholzj, line 1: Same as above.

scholzj (commented):
I left some more comments. I also don't understand why are the changes to the User Operator tests needed. What changed the outcome of the tests?

scholzj:
> I also don't understand why are the changes to the User Operator tests needed. What changed the outcome of the tests?

I guess this was a stupid question. You changed the asserts only, so I guess it is already using the MockCertManager.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaReconciler.java
devguyio: Good catch! This was overlooked and should've been reverted after I've introduced the HashMap cache.

operator-common/src/main/java/io/strimzi/operator/cluster/model/Ca.java
devguyio: The idea here is to encapsulate all the logic around the name composition in a single type. I'm fine with moving it to the Ca as well given such Secret related naming logic exists (e.g. annotation names etc, which IMHO are a good candidate for being converted into an ENUM as well).

Lemme know your preference.

operator-common/src/main/java/io/strimzi/operator/cluster/model/CertUtils.java
devguyio: > Does not seem to be used anywhere else then in cluster-operator from a quick look.
Ack. Will move it.

operator-common/src/main/java/io/strimzi/operator/cluster/model/Ca.java
scholzj: For me, Enum should cover a finite set of options. You clearly have here (for various reasons - some valid, some probably just our stupidity) an infinite set of key names. So at least from my personal point of view -> keep the Enum where it applies and move the other methods away. Or - change it to a regular class with some constants and some methods to generate other names dynamically. IMHO, the main advantage of the Enum would be the type enforcement. But you do not have it anyway if you somewhere use the Enum and somewhere just a generated String.

Feel free to check with Tom Bentley if you want - he might have a better idea of how common is to mix unrelated methods into the Enum like this.

operator-common/src/test/resources/END_ENTITY_CERT.pem
devguyio: I added them to follow the other Cluster & Client CA certs in the `MockCertManager` which exist as pem files under test resources as well.

I didn't add the key store though since it's not loaded or used in the `MockCertManager`.

scholzj: So, is it used anywhere? They do not seem to be referenced in the code.

operator-common/src/main/java/io/strimzi/operator/common/Util.java
devguyio: Done

operator-common/src/test/resources/END_ENTITY_CERT.pem
devguyio: I didn't see them used anywhere. I assumed they're added there for a reason, I can only making them available as files allows folks to check them out with openssl cli easily without having to copy / paste from the IDE. 

do you want me to remove only the ones I added? or all of them?

scholzj: Well, I did not check how the others are used. I guess @ppatierno added them so he might know how they are used. But in any case, if we do not know how and what for are they used I would not add the new ones just because I guess.

operator-common/src/main/java/io/strimzi/operator/cluster/model/Ca.java
devguyio: Another problem of moving those methods outside is that the encapsulation of the suffix would break, since this code relies on the knowledge that there's a "suffix" , which will need to have a getter and IMHO it would look ugly.

However from the perspective of this Enum usage, this code makes more sense to me:

```java
String store = SecretEntry.getDataForPod(secret, podName, SecretEntry.P12_KEYSTORE);
String password = SecretEntry.getDataForPod(secret, podName, SecretEntry.P12_KEYSTORE_PASSWORD);
```

Compared to e.g.

```java
String store = Ca.getSecretDataForPod(secret, podName, SecretEntry.P12_KEYSTORE);
String password = Ca.getSecretDataForPod(secret, podName, SecretEntry.P12_KEYSTORE_PASSWORD);
```

Again, it's just an improvement which I can revert and hard code the names as they were.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaReconciler.java
devguyio: Done

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaReconciler.java
devguyio: Done

operator-common/src/main/java/io/strimzi/operator/cluster/model/Ca.java
scholzj: The two code snippets look almost identical to me. I do not see anything broken about it.

devguyio: I thought maybe it's easier if we talk about a specific code so I've pushed a commit with the refactoring to a separate branch https://github.com/devguyio/strimzi-kafka-operator/commit/24db7e751ae447cba0732da6f9b20de7c5058303

Is that the change you feel we should go with?

scholzj: That code looks mostly good to me. I would move the Enum up to p or down to the bottom of the class. A nd the methods with other methods. But that is just the placement in the file. The code it self would work for me.

devguyio: ack, I just used the IDE to refactor and get a change out. Will change it properly and push it. Thanks!

devguyio: Done

operator-common/src/main/java/io/strimzi/operator/cluster/model/Ca.java
```diff
@@ -512,6 +543,26 @@ protected Map<String, CertAndKey> maybeCopyOrGenerateCerts(
         return certs;
     }
 
+    private boolean isNewVersion(Secret secret, String podName) {
```
scholzj, line 546: Can you add some Javadoc here explaining what it does and why is it named like this? Without the comment on line 497, it is completely unclear what version it is talking about in the method name. Only the comment on line 497 actually makes it a bit clear.

Something like _Secrets created by older operator versions did not contain all the fields. This method checks if they are present._ or at least something simple like that.

scholzj (commented):
One more nit so that we know what it does next year ;-). Plus the resource files. but LGTM otherwise. I will start the regression to see what that thinks about it.

operator-common/src/main/java/io/strimzi/operator/cluster/model/Ca.java
devguyio: Done

operator-common/src/test/resources/END_ENTITY_CERT.pem
ppatierno: I went through my PR where I added those .pem files and they are not used at all ... of course it's 3 years ago :-) Tbh I think we can delete them. There is no code where they are loaded as resources but the hard-coded ones are used instead.

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/CertUtils.java
```diff
@@ -0,0 +1,44 @@
...
+    /**
+     * Generates a short SHA1-hash (a hash stub) of the certificate which is used to track when the certificate changes and rolling update needs to be triggered.
+     *
+     * @param certSecret    Secrets with the certificate
+     * @param key           Key under which the certificate is stored in the Secret
+     * @return              Hash stub of the certificate
+     */
+    public static String getCertificateShortThumbprint(Secret certSecret, String key) {
+        var thumbprint = getCertificateThumbprint(certSecret, key);
+        return thumbprint == null ? null : thumbprint.substring(0, 8);
```
ppatierno, line 26: out of curiosity, what's the rationale behind `substring(0,8)` and not bigger or shorter?

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaReconciler.java
```diff
@@ -149,6 +150,7 @@ public class KafkaReconciler {
     private String logging = "";
     private String loggingHash = "";
     private final Map<Integer, String> brokerConfigurationHash = new HashMap<>();
+    private final Map<Integer, String> kafkaServerCertificateHash = new HashMap<>();
```
ppatierno, line 153: is it on purpose, maybe related to KRaft naming things (kafka servers, controllers, ...), that we are moving from using `broker` prefix to `kafkaServer` here?

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/CertUtils.java
scholzj: We use the 8 character hash stub for ages. But for whatever reason this Pr has to do it differently and use the full has for the server certificate 🙄

ppatierno: Do we have a constant define somewhere for the 8 character size hash stub?

scholzj: I don't think so. Before this PR it was actually just in one place anyway I think.

ppatierno: So with two places now, a constant would be better.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaReconciler.java
devguyio: yes, @tombentley suggested above that since we're introducing a new user facing API ,i.e. the annotation, we stay away from broker and use something more generic. So server was chosen.

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/CertUtils.java
devguyio: Done

operator-common/src/test/resources/END_ENTITY_CERT.pem
devguyio: @scholzj @ppatierno I deleted the ones added in this PR, we can delete the other in a separate PR I believe if you wanted.

scholzj (approved):

ppatierno (approved):

PR #7664, opened 2022-11-21
Improve JavaDoc comments before moving to Java 17

scholzj:
### Type of change

- Task

### Description

Java 17 has much stricter validation of JavaDoc comments - especially when JavaDoc warnings should be treated as errors. To move to Java 17 without disabling treating of warnings as errors in the JavaDoc Maven plugin, we need to add a lot of JavaDoc comments. 

In order to make the PR reviews easier with smaller PRs, this PR is a _preparation_ for Java 17 and adds some of the JavaDoc comments upfront while they are not needed yet. Where needed, it adds the comments or where possible changes the visibility of the methods or removes unused classes (such as `OpenShiftOriginLinker`).


api/src/main/java/io/strimzi/api/kafka/model/AbstractConnectorSpec.java
```diff
@@ -36,40 +39,73 @@ public abstract class AbstractConnectorSpec extends Spec {
 
     private AutoRestart autoRestart;
 
+    /**
+     * @return  Maximal number of tasks
```
ppatierno, line 43: Why Maximal and not maximum as the description?

api/src/main/java/io/strimzi/api/kafka/model/AbstractConnectorSpec.java
```diff
@@ -36,40 +39,73 @@ public abstract class AbstractConnectorSpec extends Spec {
...
     @Description("The maximum number of tasks for the Kafka Connector")
     @Minimum(1)
     public Integer getTasksMax() {
         return tasksMax;
     }
 
+    /**
+     * Sets the maximum number of tasks
+     *
+     * @param tasksMax  Maximal number of tasks
```
ppatierno, line 54: Maxium?

config-model/src/main/java/io/strimzi/kafka/config/model/Scope.java
```diff
@@ -4,8 +4,22 @@
...
 package io.strimzi.kafka.config.model;
 
+/**
+ * Scope of the configuration option
+ */
 public enum Scope {
+    /**
+     * Per-broker option
+     */
     PER_BROKER,
```
ppatierno, line 14: when I did this on the bridge I preferred to have this kind of comments on one line to avoid the enum to be longer to scroll. But it's just my preference for readability.

ppatierno (approved):
LGTM overall. Of course, it's huge so I went through faster :-)
Just a couple of nits.

config-model/src/main/java/io/strimzi/kafka/config/model/Scope.java
scholzj: TBH, it seems better readable to me with the full comment.

ppatierno: But you have to scroll more for the full list of enums in the editor. Anyway no need to change of course.

PR #7677, opened 2022-11-22
Add ability to create REST API users for Cruise Control

kyguy:
### Type of change

- Enhancement / new feature

### Description

This PR gives ability  to create REST API users for  the Cruise Control REST API. 
This allows developers and third-party applications to define roles and permissions to access to the Cruise Control REST API without having to disable HTTP basic authentication. 

The rough plan is to add an `apiUsers` section to the `cruiseControl.spec` where an API users can be defined by name and predefined Cruise Control role [1]. As of right now, the schema will look something like this:
```
spec:
  cruiseControl:
    apiUsers:
      - name: user
        password:
          valueFrom:
            secretKeyRef:
              name: my-secret (1)
              key: my-password (2)
        role: USER (3)
     ...
```
(1) The name of the secret containing the predefined password.
(2) The key for the password stored inside the secret.
(3) By default Cruise Control defines three roles: VIEWER, USER and ADMIN.

    VIEWER role: has access to the most lightweight kafka_cluster_state, user_tasks and review_board endpoints.
    USER role: has access to all the GET endpoints except bootstrap and train.
    ADMIN role: has access to all endpoints.

 https://github.com/linkedin/cruise-control/wiki/Security#authorization

### Checklist

- [ ] Write tests
- [ ] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/resource/cruisecontrol/CruiseControlApiImpl.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/resource/cruisecontrol/CruiseControlApiImpl.java
@@ -81,7 +81,7 @@ public class CruiseControlApiImpl implements CruiseControlApi {
         }
     }
 
-    private static HTTPHeader generateAuthHttpHeader(String user, String password) {
+    public static HTTPHeader generateAuthHttpHeader(String user, String password) {
         String headerName = "Authorization";
         String headerValue = "Basic " + Util.encodeToBase64(String.join(":", user, password));
 
@@ -460,5 +460,9 @@ public class CruiseControlApiImpl implements CruiseControlApi {
         } else {
             result.fail(t);
         }
+    }
+
+    public void setAuthHttpHeader(HTTPHeader authHttpHeader) {
+        this.authHttpHeader = authHttpHeader;
     }
 }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,7 @@
 * Moved from using the Jaeger exporter to OTLP exporter by default
 * Kafka Exporter support for `Recreate` deployment strategy
 * `ImageStream` validation for Kafka Connect builds on OpenShift
-* Add read-only REST API user for Cruise Control
+* Add ability to create REST API users for Cruise Control
 
 ### Changes, deprecations and removals
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
 * Moved from using the Jaeger exporter to OTLP exporter by default
 * Kafka Exporter support for `Recreate` deployment strategy
 * `ImageStream` validation for Kafka Connect builds on OpenShift
+* Add read-only REST API user for Cruise Control
 
 ### Changes, deprecations and removals
 
--- a/api/src/main/java/io/strimzi/api/kafka/model/CruiseControlSpec.java
+++ b/api/src/main/java/io/strimzi/api/kafka/model/CruiseControlSpec.java
@@ -8,6 +8,7 @@ import com.fasterxml.jackson.annotation.JsonInclude;
 import com.fasterxml.jackson.annotation.JsonPropertyOrder;
 import io.fabric8.kubernetes.api.model.ResourceRequirements;
 import io.strimzi.api.annotations.DeprecatedProperty;
+import io.strimzi.api.kafka.model.balancing.ApiUser;
 import io.strimzi.api.kafka.model.balancing.BrokerCapacity;
 import io.strimzi.api.kafka.model.template.CruiseControlTemplate;
 import io.strimzi.crdgenerator.annotations.Description;
@@ -16,6 +17,7 @@ import io.strimzi.crdgenerator.annotations.KubeLink;
 import io.sundr.builder.annotations.Buildable;
 import java.io.Serializable;
 import java.util.HashMap;
+import java.util.List;
 import java.util.Map;
 import lombok.EqualsAndHashCode;
 
@@ -28,7 +30,7 @@ import lombok.EqualsAndHashCode;
 @JsonInclude(JsonInclude.Include.NON_NULL)
 @JsonPropertyOrder({
     "image", "tlsSidecar", "resources", "livenessProbe", "readinessProbe", "jvmOptions", "logging", "template",
-    "brokerCapacity", "config", "metricsConfig"})
+    "brokerCapacity", "apiUsers", "config", "metricsConfig"})
 @EqualsAndHashCode
 public class CruiseControlSpec implements HasConfigurableMetrics, UnknownPropertyPreserving, Serializable {
     private static final long serialVersionUID = 1L;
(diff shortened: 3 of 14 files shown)


cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/resource/cruisecontrol/CruiseControlApiImpl.java
```diff
@@ -65,7 +65,7 @@ public HttpClientOptions getHttpClientOptions() {
         }
     }
 
-    private static HTTPHeader generateAuthHttpHeader(String user, String password) {
+    public static HTTPHeader generateAuthHttpHeader(String user, String password) {
```
scholzj, line 68: Does it really need to be public? Wouldn't some lower level be sufficient? If it hs to be public, please add Javadoc.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/resource/cruisecontrol/CruiseControlApiImpl.java
```diff
@@ -445,4 +445,8 @@ private void httpExceptionHandler(Promise<? extends CruiseControlResponse> resul
             result.fail(t);
         }
     }
+
+    public void setAuthHttpHeader(HTTPHeader authHttpHeader) {
```
scholzj, line 449: Same as above. Does it really need to be public? If yes, add Javadoc please.

scholzj (commented):
It seems to give the access just out of the box without enabling it in any way. Can you explain why that is not a security risk?

kyguy:
@scholzj It does expose information relating to cluster state like distribution of leader/follower information, offline disks, load of partitions.  I didn't think this information could be harmful but maybe it could be used for nefarious purposes and therefore should not be exposed out of the box.

As an alternative, what if we added a section to the cruiseControl spec of the Kafka resource, that would allow people to enable read-only Cruise Control API users? e.g.:
```
spec:
  cruiseControl:
    apiUsers:
      - name: user
        password: user
        role: USER (1)
```
(1) The `USER` roll grants access to the `GET` endpoints of Cruise Control API

A question for @mimaison @ppatierno @ShubhamRwt as well! Maybe this is something that should be separated into a proposal? What do you think? 


[1] https://github.com/linkedin/cruise-control/wiki/Security#authorization

scholzj:
> It does expose information relating to cluster state like distribution of leader/follower information, offline disks, load of partitions. I didn't think this information could be harmful but maybe it could be used for nefarious purposes and therefore should not be exposed out of the box.

For some users, things such as broker addresses or topic names would for sure be confidential. Not sure what Cruise Control really offers, but it probably should be considered. Lets see what other think.

> As an alternative, what if we added a section to the cruiseControl spec of the Kafka resource, that would allow people to enable read-only Cruise Control API users? e.g.:

Should it be `role: USER` or something like `role: READ-ONLY` to leave space for pottential non-read-only users? Also, if you want the user to specify a password, it would need to reference a secret. Or I guess you could just generate it.

kyguy:
> Should it be role: USER or something like role: READ-ONLY to leave space for pottential non-read-only users? 

Currently, the DefaultRoleSecurityProvider of Cruise Control only has three roles defined: `VIEWER`, `USER`, `ADMIN` [1]. We would need to create a custom RoleSecurityProvider to create additional roles. That being said, it would probably be better user experience if we offer additional role options like`READ_ONLY` through the Strimzi custom resource, even if these role options end up mapping to one of the predefined Cruise Control roles of the DefaultRoleSecurityProvider.

> Also, if you want the user to specify a password, it would need to reference a secret. Or I guess you could just generate it.

We could offer the option for both, I'll look into it

[1] https://github.com/linkedin/cruise-control/wiki/Security#authorization

ppatierno:
I think having a way to declare users in the custom resource is a good idea. Right now, I would just stick with the roles we have in Cruise Control unless we have clear in mind what are the kind of permissions/rights we want to assign and how to segregate into new roles. Maybe this would need a proposal anyway?

scholzj:
I guess sticking with the existing roles makes sense. Although not sure what the difference between them is TBH. Viewer sounds like read-only user. Admin sounds like a user who can do something. But not sure what would user do in between.

mimaison:
As Kyle mentioned above, you can find the description of all the default Cruise Control roles in https://github.com/linkedin/cruise-control/wiki/Security#authorization:
```
By default Cruise Control defines three roles: VIEWER, USER and ADMIN.

    VIEWER role: has access to the most lightweight kafka_cluster_state, user_tasks and review_board endpoints.
    USER role: has access to all the GET endpoints except bootstrap and train.
    ADMIN role: has access to all endpoints.
```

According to the [endpoint list](https://github.com/linkedin/cruise-control/blob/migrate_to_kafka_2_5/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/CruiseControlEndPoint.java#L40-L48), `USER` gives access to the following `GET` endpoints:
- `/kafkacruisecontrol/load`: To query the cluster load [0]
- `/kafkacruisecontrol/partition_load`: To query the partition load [1]
- `/kafkacruisecontrol/proposals`: To query optimization proposals [2]
- `/kafkacruisecontrol/state`: To get the state of Cruise Control [3]
- `/kafkacruisecontrol/kafka_cluster_state`: To get the Kafka cluster state [4]
- `/kafkacruisecontrol/user_tasks`: To get the list of tasks [5]
- `/kafkacruisecontrol/review_board`: To get the requests in the Review Board [6]

0: https://github.com/linkedin/cruise-control/wiki/REST-APIs#query-the-current-cluster-load
1: https://github.com/linkedin/cruise-control/wiki/REST-APIs#query-partition-resource-utilization
2: https://github.com/linkedin/cruise-control/wiki/REST-APIs#get-optimization-proposals 
3: https://github.com/linkedin/cruise-control/wiki/REST-APIs#query-the-state-of-cruise-control
4: https://github.com/linkedin/cruise-control/wiki/REST-APIs#query-partition-and-replica-state
5: https://github.com/linkedin/cruise-control/wiki/REST-APIs#query-the-user-request-result
6: https://github.com/linkedin/cruise-control/wiki/2-step-verification-for-POST-requests#checking-the-requests-that-are-pending_review-approved-submitted-or-discarded

Currently, with the default Cruise Control security provider, `DefaultRoleSecurityProvider`, both `USER` and `VIEWER` roles are effectively read-only and only have access to `GET` endpoints. I think the default roles are satisfactory for our use cases so we should just reuse them instead of mapping custom roles (like read-only) onto them (or by creating a custom security provider).

scholzj:
@kyguy What is the state of this? Is it still a work in progress? There wasn't any change for some time.

kyguy:
Hey @scholzj, this is still a work in progress. I am going to change the direction of this PR based on the feedback. I have converted the PR back to `DRAFT` mode and updated the description to reflect the new direction the PR will take. I am hoping the changes will be small enough that we can do any potential design iteration here and not need a separate proposal. Apologies for the delay!

scholzj:
Should be re-opened when it is ready.

PR #7688, opened 2022-11-22
Move to Java 17 as a language level

scholzj:
### Type of change

- Task

### Description

The PR #7624 moved Strimzi operators to Java 17 as the runtime. This PR moves the Strimzi Operators to Java 17 as a language level as well.

The exception to this are the following modules - `api`, `test`, `crd-generator` and `crd-annotations` which stay on the Java 11 language level. The reason for that is that the `api` module is a published artifact which we expect users to use and we want it to be still usable in Java 11. The other 3 modules are dependnecies of the `api` module so they need to stay on Java 11 as well.

Java 17 also does a much more strict validation of the JavaDocs. When configured to raise warnings as errors, it requires Javadocs for all public classes, methods or fields. The PR #7664 already fixed Javadocs for most modules. But there are still hundreds of errors in the `api`, `operator-common`, `cluster-operator` and `topic-operator` modules.  This PR for the time being disables the raising of warnings as errors for the time being. We should try to fix the Javadocs for `operator-common`, `cluster-operator` and `topic-operator` in the future. But I would suggest to keep this disabled for `api` permanently. The `api` classes are mostly getters and setters without any real functionality. Documenting all of them will not add much value but will be a lot of effort. In addition, the missing comments are also on the generated classes where we cannot add the comments easily but we want to have them in the Javadocs as well.

This PR also updates the Azure pipelines to use Java 17 and updates various Maven plugins to make them compatible with Java 17.

### Checklist

- [x] Update documentation
- [x] Update CHANGELOG.md

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 * Support for automatically restarting failed Connect or Mirror Maker 2 connectors
 * Redesign of Strimzi User Operator to improve its scalability
-* Use Java 17 as the runtime for all containers
+* Use Java 17 as the runtime for all containers and language level for most modules
 * Upgrade Vert.x to 4.3.5
 
 ### Changes, deprecations and removals
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 * Support for automatically restarting failed Connect or Mirror Maker 2 connectors
 * Redesign of Strimzi User Operator to improve its scalability
-* Use Java 17 as the runtime for all containers and language level for most modules
+* Use Java 17 as the runtime for all containers and language level for all modules except `api`, `crd-generator`, `crd-annotations`, and `test`
 * Upgrade Vert.x to 4.3.5
 
 ### Changes, deprecations and removals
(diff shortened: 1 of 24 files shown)


CHANGELOG.md
```diff
@@ -4,7 +4,7 @@
 
 * Support for automatically restarting failed Connect or Mirror Maker 2 connectors
 * Redesign of Strimzi User Operator to improve its scalability
-* Use Java 17 as the runtime for all containers
+* Use Java 17 as the runtime for all containers and language level for most modules
```
tombentley, line 7: Maybe this should be "all modules except..." to it's more precise.

certificate-manager/src/test/java/io/strimzi/certs/OpenSslCertManagerIT.java
```diff
@@ -78,7 +78,7 @@ public void testGenerateRootCaCertWithDays() throws Exception {
         X509Certificate x509Certificate1 = loadCertificate(cert);
         assertTrue(selfVerifies(x509Certificate1),
                 "Unexpected self-verification");
-        assertEquals(x509Certificate1.getSubjectDN(), x509Certificate1.getIssuerDN(), "Unexpected self-signedness");
```
tombentley, line 81: I assume these methods were removed?

tombentley (approved):

ppatierno (approved):

certificate-manager/src/test/java/io/strimzi/certs/OpenSslCertManagerIT.java
scholzj: They are deprecated as they expose some internal class or something. So where possible, I replaced it with the alternative which should be used now.

PR #7705, opened 2022-11-24
Allow ClusterIP listeners to configure host

ThomasK33:
### Type of change

- Bugfix
- Refactoring

### Description

With the merge of #7365, one can manually expose Kafka in one's preferred way.

When manually configuring ingress TCP configurations, e.g., with Traefik's IngressRouteTCP CRD, one will get confronted with a `No subject alternative DNS name matching` error message.

To fix this, one would have to be able to configure the bootstrap's and brokers' host names, yet this currently is only possible for `Ingress` or `Route` listeners.

(I've also fixed a typo in the error message changing `ot` to `or.`)

### Checklist

- [X] Write tests
- [X] Update documentation
- [X] Check RBAC rights for Kubernetes / OpenShift roles
- [ ] Update CHANGELOG.md

### DIFF
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/model/ListenersValidator.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/model/ListenersValidator.java
@@ -187,7 +187,7 @@ public class ListenersValidator {
     }
 
     /**
-     * Validates that useServiceDnsDomain is used only with internal type listener
+     * Validates that useServiceDnsDomain is used only with internal or cluster-ip type listener
      *
      * @param errors    List where any found errors will be added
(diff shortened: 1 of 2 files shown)


scholzj (commented):
Thanks for the PR. However, this is not how it should work. If you want to add some additional SANs to the certificates, you can use the `alternativeNames`: https://strimzi.io/docs/operators/latest/full/configuring.html#property-listener-config-altnames-reference

The host field is something what is used for the actual Route or Ingress resource in the `route` or `ingress` listeners.

ThomasK33:
Thanks a lot, Jakub.
Somehow, I completely missed this.

PR #7711, opened 2022-11-25
Update docs and release notes with improved FIPS support

scholzj:
### Type of change

- Documentation

### Description

Thsi PR updates the CHANGELOG and the documentation with the information about the improved FIPS support.

### Checklist

- [x] Update documentation
- [x] Update CHANGELOG.md


documentation/modules/operators/proc-disabling-fips-mode-cluster-operator.adoc
```diff
@@ -0,0 +1,51 @@
+// Module included in the following assemblies:
+//
+// assembly-using-the-cluster-operator.adoc
+
+[id='proc-disabling-fips-mode-cluster-operator-{context}']
+= Disabling FIPS mode
+
+[role="_abstract"]
+If you want, you can also disable the FIPS mode by setting the `FIPS_MODE` environment variable to `disabled` in the deployment configuration for the Cluster Operator.
```
PaulRMellor, line 9:
```suggestion
Strimzi automatically switches to FIPS mode when running on a FIPS-enabled Kubernetes cluster.
Disable FIPS mode by setting the `FIPS_MODE` environment variable to `disabled` in the deployment configuration for the Cluster Operator.
```

documentation/modules/operators/proc-disabling-fips-mode-cluster-operator.adoc
```diff
@@ -0,0 +1,51 @@
...
+With disabled FIPS mode, Strimzi will automatically disable FIPS in the OPenJDK in all components.
```
PaulRMellor, line 10:
```suggestion
With FIPS mode disabled, Strimzi automatically disables FIPS in the OpenJDK for all components.
```

documentation/modules/operators/proc-disabling-fips-mode-cluster-operator.adoc
```diff
@@ -0,0 +1,51 @@
...
+Such Strimzi deployment won't be FIPS compliant, and the Strimzi operators as well as all of its operands will be running in the same way as if they were running on an Kubernetes cluster without FIPS enabled.
```
PaulRMellor, line 11:
```suggestion
With FIPS mode disabled, Strimzi is not FIPS compliant.
The Strimzi operators, as well as all operands, run in the same way as if they were running on an Kubernetes cluster without FIPS enabled.
```

documentation/modules/operators/proc-disabling-fips-mode-cluster-operator.adoc
PaulRMellor: Might be useful to say why you might want to do this?

PaulRMellor (commented):
Great! I wonder if the password constraints should be as a prereq in the procedure

documentation/modules/operators/proc-disabling-fips-mode-cluster-operator.adoc
scholzj: > Might be useful to say why you might want to do this?

To be honest, I do not know. But I want to keep the option _just in case_ and since the doc is already there, though tit might be easiest to keep it 🤷

scholzj:
> Great! I wonder if the password constraints should be as a prereq in the procedure

Thanks for the comments. They should be applied now. The password constraint do not apply to the FIPS disabled mode so I do not think it should be in the procedure.

ppatierno (approved):

PaulRMellor:
> > Great! I wonder if the password constraints should be as a prereq in the procedure
> 
> Thanks for the comments. They should be applied now. The password constraint do not apply to the FIPS disabled mode so I do not think it should be in the procedure.

Thanks @scholzj -  I thought about that suggestion after making it. Didn't make sense considering we're disabling.

PaulRMellor (approved):
:+1:

PR #7732, opened 2022-11-30
Use special disabled operators when some functionality is disabled

scholzj:
### Type of change

- Enhancement / new feature

### Description

Today - in the User Operator - it is possible that some features are disabled. This includes:
* ACL management is disabled when authorization is disabled in the KAfka cluster or when _incompatible_ authorizer is used which does not support the Kafka Admin APi for managing ACLs
* SCRAM-SHA credentials management is disabled in KRaft mode which currently doesn't support SCRAM-SHA authentication

Right now, when these features are disabled, it is handled in the following way:
* The `KafkaUser` custom resources are validated to not use the disabled features and are rejected rigth away if they do
* All calls to the disabled operations are only conditional and are not executed when the feature is disabled. This is needed because for example even if KafkaUser does not specify any ACL rules, we normally check that the user doesn't have any ACL rules set.

Despite this, we create the `ScramCredentialsOperator` and `SimpleAclOperator` as we do when the feature is enabled. They create the Cache and the Batch reconciler, but do not starts them. So they are nto running, but they exist and use some memory.

This PR creates new `DisabledSimpleAclOperator` and `DisabledScramCredentialsOperator` classes. These classes do not have any logic and when any of the methods in them is called they throw `UnsupportedOperationException` error. When the SCRAM-SHA or ACLs are not supported, these classes are created instead of the `ScramCredentialsOperator` and `SimpleAclOperator` classes to be more efficient. But it should also help to detect any bugs in the handling of the disabled features because if they are somewhere called by mistake, the exceptions will be visible in the logs.

Other alternative I considered  wasto have these _dummy_ classes always return succeeded futures without doing anything. This might have helped us to eliminate the conditional logic when the operators should or should not be called (see for example [here](https://github.com/strimzi/strimzi-kafka-operator/blob/0c1f7ef79aed58d257421dd430598b16c75acff8/user-operator/src/main/java/io/strimzi/operator/user/operator/KafkaUserOperator.java#L197-L199)). But it would make it harder to see if they are somewhere called by mistake. So at the end, I decided to implement the solution where they throw the exceptions.

### Checklist

### DIFF
--- a/user-operator/src/main/java/io/strimzi/operator/user/Main.java
+++ b/user-operator/src/main/java/io/strimzi/operator/user/Main.java
@@ -22,6 +22,8 @@ import io.strimzi.operator.common.MetricsProvider;
 import io.strimzi.operator.common.MicrometerMetricsProvider;
 import io.strimzi.operator.common.OperatorKubernetesClientBuilder;
 import io.strimzi.operator.common.Util;
+import io.strimzi.operator.user.operator.DisabledScramCredentialsOperator;
+import io.strimzi.operator.user.operator.DisabledSimpleAclOperator;
 import io.strimzi.operator.user.operator.KafkaUserOperator;
 import io.strimzi.operator.user.operator.QuotasOperator;
 import io.strimzi.operator.user.operator.ScramCredentialsOperator;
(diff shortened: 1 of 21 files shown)


tombentley (approved):
This seems a reasonable approach to me. Have you tested that the exception does get logged?

scholzj:
@tombentley Thanks. Yes, it is logged. It fails the reconciliation. I added a test for it as well.

PR #7734, opened 2022-11-30
add log for troubleshooting when init admin client error

showuon:
### Type of change

- Bugfix

### Description

When initAdminClient got error, we'll force restarting pods. But we didn't log any exception while the error happened, which will be difficult to identify the reason why we didn't successfully create admin client. Log it in this PR when exception thrown.

### Checklist

- [ ] Write tests
- [ ] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/resource/KafkaRoller.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/resource/KafkaRoller.java
@@ -179,6 +179,7 @@ public class KafkaRoller {
             try {
                 this.allClient = adminClient(IntStream.range(0, podList.size()).boxed().collect(Collectors.toList()), false);
             } catch (ForceableProblem | FatalProblem e) {
+                LOGGER.warnCr(reconciliation, "Failed to create adminClient.", e);
                 return false;
             }
         }


cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/resource/KafkaRoller.java
```diff
@@ -179,6 +179,7 @@ private boolean initAdminClient() {
             try {
                 this.allClient = adminClient(IntStream.range(0, podList.size()).boxed().collect(Collectors.toList()), false);
             } catch (ForceableProblem | FatalProblem e) {
+                LOGGER.infoCr(reconciliation, "Failed to create adminClient.", e);
```
scholzj, line 182: Should it be logged as a warning message if it as an error which we do not expect to happen?

scholzj (commented):
Thaks for the PR. I like the idea of logging it ... but should it be a warning message?

scholzj:
Also, can you please fix the DCO sign-off? The instructions should be under the _Details_ link next to the DCO status in the PR's status section

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/resource/KafkaRoller.java
showuon: OK, updated. Thanks.

tombentley (approved):

scholzj:
Thanks for the PR @showuon

PR #7740, opened 2022-12-01
Fix "container not found" error when trying to get strimzi.properties from crashing kafka/zk pods.

fvaleri:
I'm also moving the log collection at the end because it takes longer than the others and if there is any issue you would still get the other data.


scholzj (approved):

ppatierno (approved):

scholzj:
Thanks for the PR @fvaleri

PR #7751, opened 2022-12-04
Use shared executor for async operations using the KAfka Admin API and Kubernetes API

### LINKED ISSUE #7611: Use custom executor for async operations in User Operator instead of using JVMs default
As discussed in https://github.com/strimzi/strimzi-kafka-operator/pull/7583#discussion_r1019406901 - the redesigned User Operator is currently running the async calls on the JVM's default thread pool. It might be better to use our own thread pool / executor for it.

scholzj:
### Type of change

- Enhancement / new feature

### Description

Currently, the User Operator makes different `handleAsync` calls when using the Kubernetes or Kafka Admin APIs during the user reconciliations. These are in the `KafkaUserOperator` class and in the Admin API Operators used by it. Right now, these calls are using the JVM's default thread pool.

This PR creates a new shared thread pool executor for these operations. By default, it uses 4 threads. If needed, the pool size is user configurable to tune it for the best performance.

This should close #7611 

### Checklist

- [x] Update documentation

### DIFF
--- a/user-operator/src/main/java/io/strimzi/operator/user/Main.java
+++ b/user-operator/src/main/java/io/strimzi/operator/user/Main.java
@@ -33,6 +33,8 @@ import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 
 import java.security.Security;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
 
 /**
  * The main class of the Strimzi User Operator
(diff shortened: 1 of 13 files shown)


user-operator/src/main/java/io/strimzi/operator/user/Main.java
```diff
@@ -70,13 +72,15 @@ public static void main(String[] args) {
         UserOperatorConfig config = UserOperatorConfig.fromMap(System.getenv());
         KubernetesClient client = new OperatorKubernetesClientBuilder("strimzi-user-operator", Main.class.getPackage().getImplementationVersion()).build();
         Admin adminClient = createAdminClient(config, client, new DefaultAdminClientProvider());
+        ExecutorService kafkaUserOperatorExecutor = Executors.newFixedThreadPool(config.getUserOperationsThreadPoolSize(), r -> new Thread(r, "operator-thread-pool"));
```
tombentley, line 75: Doesn't this mean all the threads will have the same name? IF so, that could result in some confusing logs. Better to use names which are more likely to be unique (e.g. using an `AtomicInteger` to generate a more unique name on thread creation.

tombentley (approved):

PR #7755, opened 2022-12-06
Set 1 ZK replica for Kafka ephemeral single example

fvaleri:
Trivial PR for fixing the Kafka ephemeral single example.

### DIFF
--- a/packaging/examples/kafka/kafka-ephemeral-single.yaml
+++ b/packaging/examples/kafka/kafka-ephemeral-single.yaml
@@ -25,7 +25,7 @@ spec:
     storage:
       type: ephemeral
   zookeeper:
-    replicas: 3
+    replicas: 1
     storage:
       type: ephemeral
   entityOperator:


ppatierno:
IIRC, having 3 ZK replicas was done on purpose in order to have your ephemeral cluster still working if one ZK node goes down. @scholzj do you remember?

fvaleri:
@ppatierno but there is another ephemeral example with 3 ZKs (kafka-ephemeral.yaml).

ppatierno:
@fvaleri the difference between them is the number of Kafka replicas, 1 vs 3 ... but both should have 3 replicas for ZK.

scholzj (commented):
As said by Paolo, without 3 ZooKeeper replicas, this is essentially useless because it does not survive any rolling update or pod deletion etc. This way, it at least helps a bit from suers who take the regular ephemeral example, change the replicas to 1 and then wonder why does it not work.

fvaleri:
Ok, I see the point. I would not consider this for anything more than quick client tests. Thanks for your comments.

PR #7769, opened 2022-12-08
Add support for Batch reconciliation in the Kubernetes resource operators

scholzj:
### Type of change

- Enhancement / new feature

### Description

In some situations, we need to reconcile a _list_ of Kubernetes resources based on some selector. We want to:
* Create or patch the desired resources
* Delete the resources matching the selector but not desired anymore

We use this logic in several places for example when managing services, routes or ingresses. Moving the logic to the resource operators simplifies the code and makes it easier to use in the future.

### Checklist

- [x] Write tests


Frawless (approved):
I am not an expert but LGTM

ppatierno (approved):
Good idea, thanks!

PR #7797, opened 2022-12-13
Refactor resource operators to have a single shared abstract class

scholzj:
### Type of change

- Enhancement / new feature

### Description

Today, the resource operators we have in `operator-common` do not have a single common ancestor. Instead, there is `AbstractNonNamespacedOperator` for non-namespaced resources and `AbstractResourceOperator` for namespaced resources. While these for two separate inheritance trees, they obviously share many common aspects.

This PR restructures them. It renames the `AbstractResourceOperator` to `AbstractNamespacedResourceOperator` and creates a _new_ `AbstractResourceOperator` whihc is used as an abstract parent for both `AbstractNamespacedResourceOperator` and `AbstractNonNamespacedOperator` and moves the shared code to it.

It also refactors the methods for listing resources to have the _list_ logic copied in less different places which should make it easier to add `ListOptions` to the `list(...)` calls. This is related to #7794 to make it easier to configure the `LIST` operation to optimise the consumed resources on the Kubernetes control plane side. (However, this PR itself does not change ayn of the `list` call options => that would be subject to some later PR)

### Checklist

### DIFF
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/ClusterOperatorConfig.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/ClusterOperatorConfig.java
@@ -14,7 +14,7 @@ import io.strimzi.operator.cluster.model.UnsupportedVersionException;
 import io.strimzi.operator.common.InvalidConfigurationException;
 import io.strimzi.operator.common.Util;
 import io.strimzi.operator.common.model.Labels;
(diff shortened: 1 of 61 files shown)


scholzj:
This should be ready for review. The failing system test is related to #7808 and #7809.

ppatierno (approved):
A lot of changes :-) ... but mostly related to renaming. LGTM!

PR #7808, opened 2022-12-14
Fix flaky `testCRBDeletionErrorIsIgnoredWhenRackAwarenessIsNotEnabled` ST

scholzj:
### Type of change

- Bugfix

### Description

This Pr should fix the flaky `testCRBDeletionErrorIsIgnoredWhenRackAwarenessIsNotEnabled` system test from the `ClusterOperatorRbacIsolatedST` class. The main problem of the test is that the second assert checks only last 50 messages. That is very unstable because it depends on the timing of the different reconciliations and the amount of logged messages. It is also unreliable because while the second assert should check for the Connect related message, it can actually match the Kafka message as well.

This PR changes it and asserts for longer log message including the reconciliation marker instead of searching only the last 50 log lines. That should make sure that we actually check both Kafka and Connect log messages and since we do not need to search only the last 50 lines, we have a better chance to find the right message.

### Checklist


im-konge (approved):
LGTM, thanks

Frawless (approved):

PR #7818, opened 2022-12-15
Fix kafka pod address used to check if replication port is listening

### LINKED ISSUE #7798: KafkaRoller tcp probe uses incorrect host
In the KafkaRoller if we fail to connect an admin client connection to a broker pod (when determining if it is a controller) we try to make a TCP connection to the replication port. If that fails then we force restart the pod. I don't think the hostname we are connecting to is correct, we are connecting on `${podName}:${replicationPort}` [here](https://github.com/strimzi/strimzi-kafka-operator/blob/ef1bf282d97ecfa2ff3aa3a8ee759a11c52e9a49/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/resource/KafkaRoller.java#L803).

Debugging locally I can see that the admin client bootstraps using:
`my-cluster-kafka-0.my-cluster-kafka-brokers.myproject.svc.cluster.local:9091`

But the tcp probe tries to connect using:
`my-cluster-kafka-0:9091` which doesn't resolve to anything (edit: on minikube, does this host work in other environments?)

robobario:
Fixes #7798

### Type of change

- Bugfix

### Description

The KafkaRoller was trying to connect to `${podName}:${replicaPort}` which doesn't resolve to anything from the operator.

### Checklist

- [x] Write tests
- [ ] Update CHANGELOG.md

### DIFF
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/resource/KafkaRoller.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/resource/KafkaRoller.java
@@ -800,7 +800,7 @@ public class KafkaRoller {
             try {
                 LOGGER.debugCr(reconciliation, "Probing TCP port due to previous problems connecting to pod {}", podRef);
                 // do a tcp connect and close (with a short connect timeout)
-                tcpProbe(podRef.getPodName(), KafkaCluster.REPLICATION_PORT);
+                tcpProbe(DnsNameGenerator.podDnsName(namespace, KafkaResources.brokersServiceName(cluster), podRef.getPodName()), KafkaCluster.REPLICATION_PORT);
             } catch (IOException connectionException) {
                 throw new ForceableProblem("Unable to connect to " + podRef.getPodName() + ":" + KafkaCluster.REPLICATION_PORT, executionException.getCause(), true);
             }
@@ -817,7 +817,7 @@ public class KafkaRoller {
      * @param port The port
      * @throws IOException if anything went wrong.
      */
-    void tcpProbe(String hostname, int port) throws IOException {
+    /*test*/ void tcpProbe(String hostname, int port) throws IOException {
         Socket socket = new Socket();
         try {
             socket.connect(new InetSocketAddress(hostname, port), 5_000);
--- a/cluster-operator/src/test/java/io/strimzi/operator/cluster/operator/resource/KafkaRollerTest.java
+++ b/cluster-operator/src/test/java/io/strimzi/operator/cluster/operator/resource/KafkaRollerTest.java
@@ -12,6 +12,7 @@ import io.strimzi.operator.cluster.KafkaVersionTestUtils;
 import io.strimzi.operator.cluster.model.RestartReason;
 import io.strimzi.operator.cluster.model.RestartReasons;
 import io.strimzi.operator.cluster.operator.resource.events.KubernetesRestartEventPublisher;
+import io.strimzi.operator.common.AdminClientProvider;
 import io.strimzi.operator.common.BackOff;
 import io.strimzi.operator.common.DefaultAdminClientProvider;
 import io.strimzi.operator.common.Reconciliation;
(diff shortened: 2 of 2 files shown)


robobario:
I'm wondering if we're better off without the unit test changes as they are quite ugly. The problem is that the tests currently use this subclass, TestKafkaRoller which overrides the `controller` method.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/resource/KafkaRoller.java
```diff
@@ -817,7 +817,7 @@ private void maybeTcpProbe(PodRef podRef, Exception executionException, RestartC
      * @param port The port
      * @throws IOException if anything went wrong.
      */
-    void tcpProbe(String hostname, int port) throws IOException {
+    protected void tcpProbe(String hostname, int port) throws IOException {
```
scholzj, line 820: Does it really need to be `protected`? Wouldn't default (with `/* test */` marker) be enough?

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/resource/KafkaRoller.java
```diff
@@ -826,6 +826,10 @@ void tcpProbe(String hostname, int port) throws IOException {
         }
     }
 
+    private String podDnsName(String podName) {
+        return DnsNameGenerator.podDnsName(namespace, KafkaResources.brokersServiceName(cluster), podName);
+    }
+
```
scholzj, line 832: Do we really need a method which just calls another method and is used from 2 places? TBH, I would say no.

(And if we would decide that this should stay ... maybe it can at least include the port as well then?)

scholzj (commented):
Not sure abotu the test ... I think I will leave it to @tombentley to comment.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/resource/KafkaRoller.java
SamBarker: Given the root issue is inconsistent (and incorrect) addressing of pods a single method that gets it right seem appropriate to me. 

Including the port won't work for the TCP probe as the socket requires both address and port as separate arguments (either directly or via `InetSocketAddress`).

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/resource/KafkaRoller.java
robobario: Yep, not sure why I changed it. Changed to default with the marker.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/resource/KafkaRoller.java
scholzj: There is a single consistent method ... its called `DnsNameGenerator.podDnsName`. It will not help with consistency if every class hides it into a separate layer of obfuscation.

robobario: I've changed it to introduce an `Endpoint` record to pass back the host/port and make the new method do a little more work.

scholzj:
So, what is the plan for this? I still think this should be simplified and not add unnecessary methods and classes which just make it less readable. And nobody else from @strimzi/maintainers seemed to have jumped in and approved it either.

robobario:
Happy to rollback and re-introduce the duplication. I agree it's a bit much now.

scholzj:
I would certainly just use the DnsGenerator directly.

scholzj (approved):
Thanks @robobario

scholzj:
@tombentley Do you think you can have a look at this one? Thanks.

tombentley (approved):

scholzj:
Thanks for the fix @robobario

PR #7826, opened 2022-12-19
Add early compute resources validation for Kafka and ZK

### LINKED ISSUE #7683: Resources should validate correctness of new configuration
**Describe the bug**

During testing of #7421 , I have tried scenario with 3 Kafka pods and 3 Zookeeper pods, where user will specify some invalid configuration - f.e. setting request bigger than the limit. When the Zookeper is edited with this invalid configuration, first pod is deleted and after a while, another two pods are deleted, with 1 pod which is trying to be deployed - ending in `CrashLoopBackOff` state. But, this is not problem just for Zookeeper, but also for all other resources.

**To Reproduce**

1. Deploy CO
2. Deploy Kafka cluster with 3 Kafka pods and 3 ZK pods
3. Edit Zookeeper part of the Kafka CR, setting the invalid configuration
4. Check the rolling update of Zookeeper pods

**Expected behavior**

Instead of deleting/rolling the whole Zookeeper cluster, it would maybe throw `InvalidResourceException` and prevent the rolling update to happen

 - Strimzi version: main
 - Installation method: YAML
 - Kubernetes cluster: Kubernetes 1.19 and above
 - Infrastructure: Minikube

**YAML files and logs**

`kubectl get pods`
```
NAME                                                 READY   STATUS    RESTARTS   AGE
my-cluster-61d4c97e-entity-operator-c85cbdb6-6grc4   3/3     Running   0          16m
my-cluster-61d4c97e-kafka-0                          1/1     Running   0          16m
my-cluster-61d4c97e-kafka-1                          1/1     Running   0          11m
my-cluster-61d4c97e-kafka-2                          1/1     Running   0          11m
my-cluster-61d4c97e-zookeeper-0                      0/1     Running   1          17m
```

kafka.yaml
https://gist.github.com/im-konge/92cd4c9bf14fe63b2fde509e45ba16db

fvaleri:
This change avoids disruptive rolling update in case of invalid configuration. All the other components fails fast without any service disruption.

Can't use CustomResourceValidationExpressions (Kube 1.25+) because we still support Kube 1.19+.

Should fix #7683.


operator-common/src/main/java/io/strimzi/operator/common/Util.java
```diff
@@ -858,4 +862,32 @@ public static boolean isMaintenanceTimeWindowsSatisfied(Reconciliation reconcili
...
+     * Validate container's common compute resources.
+     * 
+     * @param resources Compute resources configuration.
+     * @param type Compute resource type (cpu or memory).
+     * @param owner Compute resource owner.
+     * 
+     * @return Error message or empty (no error found).
+     */
+    public static Optional<String> validateComputeResources(ResourceRequirements resources, String type, String owner) {
+        if (resources != null && resources.getRequests() != null && resources.getLimits() != null
```
scholzj, line 876: I don't really get the `Arrays.asList("cpu", "memory").contains(type)` ... I would expect this to apply to all resource types as a general rule. If you want to use this only for `cpu` and `memory`, I'm fine. But it should reflect in the method name and you should maybe consider:
* Throwing some exception if wrong type is passed?
* Hide the type behind two public methods - one for CPU and on for Memory - calling thi private method with a given type

But I would prefer to just keep it generic and let it handle any type.

operator-common/src/main/java/io/strimzi/operator/common/Util.java
```diff
@@ -858,4 +862,32 @@ public static boolean isMaintenanceTimeWindowsSatisfied(Reconciliation reconcili
     public static String encodeToBase64(String encode)  {
         return Base64.getEncoder().encodeToString(encode.getBytes(StandardCharsets.US_ASCII));
     }
+
+    /**
+     * Validate container's common compute resources.
```
scholzj, line 867: Plase explin what exactly you validate. You seem to validate some specific use cases - but there are other things validated elsewhere and possibly other unvalidated things. So we should be clear about what it does. Maybe you can also make it more clear in the method name that the focus is on validating limits against requests.

operator-common/src/main/java/io/strimzi/operator/common/Util.java
```diff
@@ -858,4 +862,32 @@ public static boolean isMaintenanceTimeWindowsSatisfied(Reconciliation reconcili
...
+            if (cpuRequest != null && cpuRequest.getNumericalAmount().compareTo(BigDecimal.ZERO) != 1) {
+                return Optional.of(String.format("Invalid %s request for %s: must be > 0", type, owner));
+            }
+            if (cpuLimit != null && cpuLimit.getNumericalAmount().compareTo(BigDecimal.ZERO) != 1) {
+                return Optional.of(String.format("Invalid %s limit for %s: must be > 0", type, owner));
+            }
+            if (cpuRequest != null && cpuLimit != null && cpuRequest.getNumericalAmount().compareTo(cpuLimit.getNumericalAmount()) > 0) {
+                return Optional.of(String.format("Invalid %s request for %s: must be <= limit", type, owner));
+            }
+            return Optional.empty();
```
scholzj, line 889: Should this be `if ... else if ... else if ... else if ... else` instead? It would make it more clear what happens when I guess.

operator-common/src/main/java/io/strimzi/operator/common/Util.java
```diff
@@ -858,4 +862,32 @@ public static boolean isMaintenanceTimeWindowsSatisfied(Reconciliation reconcili
...
+        }
+        return Optional.empty();
```
scholzj, line 891: Having this in `else` block would make sure that nobody will forget to add a `return` statement in future modification as it will be validated by IDE / Java. This way, if someone forgets it, it will just return the empty value wihtout any warnings.

scholzj (commented):
Thanks for the PR.

Why not have this part of the model classes and trigger it from the `fromCrd` methods? That would make more sense to me then have it as part of the reconciliation. If the resources are invalid, you can throw `InvalidRecourceException`. It should be also ideally executed for all the models and not just Kafka / Zoo.

I also thing that using a static method for it and calling it from the model classes is fine and should simplify testing. But this is ultimately Cluster Operator only thing. So it should be there  - for example in the `ModelUtils` class - and not in the `operator-common` module.

It would be also great to have tests covering the different valid and invalid resource combinations.

fvaleri:
Hi @scholzj, thanks for the suggestions.

I still need to add the missing resource type logic and test combinations, but I thinks the last commit should look similar to what you had in mind.

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaCluster.java
```diff
@@ -621,6 +623,19 @@ private static void configureCruiseControlMetrics(Kafka kafkaAssembly, KafkaClus
         }
     }
 
+    /**
+     * Early resources validation to avoid triggering any pod operation with invalid configuration.
+     * 
+     * @param resources Resources configuration.
+     */
+    private static void validateComputeResources(ResourceRequirements resources) {
+        Set<String> errors = ModelUtils.validateCpuResources(resources, "kafka");
```
scholzj, line 632: Maybe instead of `kafka` you can use the YAML path? E.g. `.spec.kafka.resources` or `.spec.kafka.zookeeper`?

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/ModelUtils.java
```diff
@@ -812,4 +815,60 @@ public static List<String> generateAllServiceDnsNames(String namespace, String s
 
         return dnsNames;
     }
+
+    /**
+     * Validate CPU requests and limits.
```
scholzj, line 820: I would expand on what do you validate.

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/ModelUtils.java
```diff
@@ -812,4 +815,60 @@ public static List<String> generateAllServiceDnsNames(String namespace, String s
...
+     * @param owner Compute resources owner.
+     * 
+     * @return Error set.
+     */
+    public static Set<String> validateCpuResources(ResourceRequirements resources, String owner) {
+        return validateComputeResources(resources, "cpu", owner);
+    }
+
+    /**
+     * Validate memory requests and limits.
```
scholzj, line 832: Same as above ... expand on what do you validate.

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/ModelUtils.java
```diff
@@ -812,4 +815,60 @@ public static List<String> generateAllServiceDnsNames(String namespace, String s
...
+     * @param owner Compute resources owner.
+     * 
+     * @return Error set.
+     */
+    public static Set<String> validateMemoryResources(ResourceRequirements resources, String owner) {
+        return validateComputeResources(resources, "memory", owner);
+    }
+
+    /**
+     * Validate compute resources requests and limits.
```
scholzj, line 844: Expand on what do you validate ;-).

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/ModelUtils.java
```diff
@@ -812,4 +815,60 @@ public static List<String> generateAllServiceDnsNames(String namespace, String s
...
+     * 
+     * TODO: handle any type, not just cpu and memory.
```
scholzj, line 846: If you plan to do it, you should open an issue. TODO like this are from my experience useless ;-). Realistically, we do not have any real use case for this anyway right now I guess. So we can also just simply clarity that we handle only CPU and Memory here?

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/ModelUtils.java
```diff
@@ -812,4 +815,60 @@ public static List<String> generateAllServiceDnsNames(String namespace, String s
...
+     * @param type Compute resources type.
+     * @param owner Compute resources owner.
+     * 
+     * @return Error set.
+     */
+    private static Set<String> validateComputeResources(ResourceRequirements resources, String type, String owner) {
+        Set<String> errors = new HashSet<>();
+        String ownerId = owner != null && !owner.isBlank() ? owner : "unknown";
+        if (resources != null && resources.getRequests() != null) {
+            Quantity cpuRequest = resources.getRequests().get(type);
```
scholzj, line 858: Its a generic method, so maybe a generic name instead of `cpuRequest` would be better?

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/ModelUtils.java
```diff
@@ -812,4 +815,60 @@ public static List<String> generateAllServiceDnsNames(String namespace, String s
...
+            if (cpuRequest != null && cpuRequest.getNumericalAmount().compareTo(BigDecimal.ZERO) != 1) {
+                errors.add(String.format("Invalid %s request for %s: must be > 0", type, ownerId));
+            }
+            if (resources.getLimits() != null) {
+                Quantity cpuLimit = resources.getLimits().get(type);
```
scholzj, line 863: Its a generic method, so maybe a generic name instead of `cpuLimit` would be better?

scholzj (commented):
This looks more closely to what I meant, yes. But you now have the method `validateComputeResources` twice there with minimal differences. Why not embed all of that into a single ModelUtils method including generating the exception and just do `validateComputeResources(kafkaClusterSpec.getResources(), "kafka");` in the `fromCrd` methods?

fvaleri:
@scholzj I think this is now ready for another review when you have time.

cluster-operator/src/test/java/io/strimzi/operator/cluster/model/KafkaClusterResourcesTest.java
```diff
@@ -0,0 +1,137 @@
...
+                .build();
+
+        if (combo.error() != null) {
+            InvalidResourceException ex = assertThrows(InvalidResourceException.class, () -> KafkaCluster.fromCrd(Reconciliation.DUMMY_RECONCILIATION, kafka, VERSIONS));
+            assertThat(ex.getMessage(), equalTo(combo.error()));
+        } else {
+            KafkaCluster kc = KafkaCluster.fromCrd(Reconciliation.DUMMY_RECONCILIATION, kafka, VERSIONS);
+            assertThat(kc.getContainers(null).get(0), notNullValue());
+        }   
+    }
```
scholzj, line 136: This is more or less good ... but you are testing a static method from `ModelUtils`. You should make your life (and the life of anyone changing this in the future) and move this to `ModelUtilsTest` and test the static methods directly. That way, it will be resilient against any changes to the `KafkaCluster` class.

cluster-operator/src/test/java/io/strimzi/operator/cluster/model/KafkaClusterResourcesTest.java
```diff
@@ -0,0 +1,137 @@
...
+                    .endJbodStorage()
+                    .withConfig(Map.of("log.message.format.version", "3.0", "inter.broker.protocol.version", "3.0"))
+                .endKafka()
+            .endSpec()
+            .build();
+
+    private record ResourcesCombo(String cpuRequest, String cpuLimit, String memoryRequest, String memoryLimit, String error) { }
+    
+    private static List<ResourcesCombo> resourcesCombos() {
+        return Arrays.asList(
```
scholzj, line 71: One particular area I'm missing here is testing different units against each other. E.g.:
* Memory request 2048Mi versus Memory limit 4Gi and the other way around
* CPU request 500m versus CPU limit 2 and the other way around

cluster-operator/src/test/java/io/strimzi/operator/cluster/model/KafkaClusterResourcesTest.java
```diff
@@ -0,0 +1,137 @@
...
+import java.util.List;
+import java.util.Map;
+
+import static org.hamcrest.CoreMatchers.equalTo;
+import static org.hamcrest.CoreMatchers.notNullValue;
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+
+@SuppressWarnings({"checkstyle:ClassDataAbstractionCoupling", "checkstyle:ClassFanOutComplexity", "checkstyle:JavaNCSS"})
+public class KafkaClusterResourcesTest {
```
scholzj, line 29: Wouldn't it be easier to just add it into the existing test class instead of adding a new one? In case you decide to keep the new separate class ... are the `SuppressWarnings` really needed for such a small class?

cluster-operator/src/test/java/io/strimzi/operator/cluster/model/KafkaClusterResourcesTest.java
```diff
@@ -0,0 +1,137 @@
...
+                            new GenericKafkaListenerBuilder()
+                                    .withName("tls")
+                                    .withPort(9093)
+                                    .withType(KafkaListenerType.INTERNAL)
+                                    .withTls(true)
+                                    .build())
+                    .withNewJbodStorage()
+                        .withVolumes(new PersistentClaimStorageBuilder().withId(0).withSize("100Gi").withDeleteClaim(false).build())
+                    .endJbodStorage()
+                    .withConfig(Map.of("log.message.format.version", "3.0", "inter.broker.protocol.version", "3.0"))
```
scholzj, line 63: I would remove this ... irrelevant to the tests and less changes might be needed in the future.

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/ModelUtils.java
```diff
@@ -812,4 +813,62 @@ public static List<String> generateAllServiceDnsNames(String namespace, String s
...
+
+    /**
+     * Validate cpu and memory resources.
+     * Early resources validation avoids triggering any pod operation with invalid configuration.
+     * 
+     * For both cpu and memory, this method checks that request is greater than zero, 
+     * limit is greater than zero and limit is greater than or equal to the request.
+     * 
+     * @param resources Resources configuration.
+     * @param owner Resources owner.
```
scholzj, line 825: Perhaps you can be a bit clear in the comment what the `owner` is or rename it to something like `resourcePath`? Both here in the private method. TBH, initially I was quite confused what does owner reference have to do with this.

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/ModelUtils.java
```diff
@@ -812,4 +813,62 @@ public static List<String> generateAllServiceDnsNames(String namespace, String s
...
+                    && resources.getLimits() != null && resources.getLimits().get(type) != null) {
+                Quantity limit = resources.getLimits().get(type);
+                Quantity request = resources.getRequests().get(type);
+                if (request != null && request.getNumericalAmount().compareTo(limit.getNumericalAmount()) > 0) {
+                    errors.add(String.format("%s %s request must be <= limit", oid, type));
+                }
+            }
+        }
+        return errors;
+    }
```
scholzj, line 873: Maybe you can consider better spacing for improved readablity? I would for example space it like this:

```suggestion
    private static List<String> validateComputeResources(ResourceRequirements resources, String type, String owner) {
        List<String> errors = new ArrayList<>();
        String oid = owner != null && !owner.isBlank() ? owner : "component";
        
        if (resources != null) {
            if (resources.getRequests() != null && resources.getRequests().get(type) != null) {
                Quantity request = resources.getRequests().get(type);
                
                if (request != null && request.getNumericalAmount().compareTo(BigDecimal.ZERO) <= 0) {
                    errors.add(String.format("%s %s request must be > zero", oid, type));
                }
            }
            
            if (resources.getLimits() != null && resources.getLimits().get(type) != null) {
                Quantity limit = resources.getLimits().get(type);
                
                if (limit.getNumericalAmount().compareTo(BigDecimal.ZERO) <= 0) {
                    errors.add(String.format("%s %s limit must be > zero", oid, type));
                }
            }
            
            if (resources.getRequests() != null && resources.getRequests().get(type) != null
                    && resources.getLimits() != null && resources.getLimits().get(type) != null) {
                Quantity limit = resources.getLimits().get(type);
                Quantity request = resources.getRequests().get(type);
                
                if (request != null && request.getNumericalAmount().compareTo(limit.getNumericalAmount()) > 0) {
                    errors.add(String.format("%s %s request must be <= limit", oid, type));
                }
            }
        }
        
        return errors;
    }
```

Just a suggestion whcih might improve the code look.

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/ModelUtils.java
```diff
@@ -812,4 +813,62 @@ public static List<String> generateAllServiceDnsNames(String namespace, String s
...
+     * 
+     * @param resources Compute resources configuration.
+     * @param type Compute resources type.
+     * @param owner Compute resources owner.
+     * 
+     * @return Error set.
+     */
+    private static List<String> validateComputeResources(ResourceRequirements resources, String type, String owner) {
+        List<String> errors = new ArrayList<>();
+        String oid = owner != null && !owner.isBlank() ? owner : "component";
```
scholzj, line 849: What are the values when this would not be passed? It seems fairly unnecessary to me. It is a private method which I would expect to be always called in the right way.

scholzj (commented):
Few mor enits. But I think the code looks mostly good. Thanks.

cluster-operator/src/test/java/io/strimzi/operator/cluster/model/KafkaClusterResourcesTest.java
fvaleri: I can do that, but I would still need tocreate a dedicated test class as I cannot mix `@ParameterizedTest` with `@Test`.

scholzj: Ok - I still think it makes sense to test it directly, even if it is a separate class.

fvaleri: Me too.

fvaleri:
@scholzj can you please run regression tests again when the build completes? Thanks.

scholzj (approved):

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaCluster.java
```diff
@@ -327,6 +327,7 @@ public static KafkaCluster fromCrd(Reconciliation reconciliation, Kafka kafkaAss
             result.clusterId = getOrGenerateKRaftClusterId(kafkaAssembly);
         }
 
+        ModelUtils.validateComputeResources(kafkaClusterSpec.getResources(), ".spec.kafka.resources");
```
ppatierno, line 330: I was wondering if we want to pass the `kafkaClusterSpec` here as for the `validateIntConfigProperty`. I also like the parameters order (first spec then what to check). Maybe in a different PR we could change such order for the `validateIntConfigProperty` as well.

ppatierno (approved):
LGTM. Just a nit.

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaCluster.java
fvaleri: Thanks Paolo. 

> I was wondering if we want to pass the kafkaClusterSpec here as for the validateIntConfigProperty.

Here I'm passing the nearest spec object that contains the values to be validated. In that sense, I think it is consistent with `validateIntConfigProperty`.

> I also like the parameters order (first spec then what to check). Maybe in a different PR we could change such order for the validateIntConfigProperty as well.

Agree.

PR #7839, opened 2022-12-21
Fix Javadocs issues in the topic-operator module

### LINKED ISSUE #7703: Fix Javadocs in the `topic-operator` module
By default we configure the Maven Javadoc plugin to raise any warnings as errors when validating the Javadocs in our code. That helps us to keep the comments valid and up-to-date. But Java 17 brings more strict checks for missing Javadocs. So as a result, we have lot of missing Javadoc errors. We decided to proceed with move to Java 17 and for now, we disabled the `failOnWarnings` flag for the `topic-operator` module. But we should get back to it, fix it and re-enable it again.

ShubhamRwt:

### Type of change

- Refactoring

### Description

This PR fixes #7703. This PR fixes the Javadocs issues in the topic-operator module  adding the missing Javadoc or making the methods / fields not public where they don't need to be public.
It also contains refactoring of some methods which are now not used.

- [ ] Write tests
- [ ] Update documentation
- [ ] Update CHANGELOG.md


scholzj (approved):
Assuming the tests pass, it is OK I guess. @tombentley as a TO SME would need to comments on the detailed changes.

topic-operator/src/main/java/io/strimzi/operator/topic/ConflictingChangesException.java
```diff
@@ -11,6 +11,12 @@
  * and the topic in Kafka had conflicting changes.
  */
 public class ConflictingChangesException extends OperatorException {
+
+    /** Constructor
```
tombentley, line 15: Eugh, can we put this on the following line? (And same for other instances, you might be able to fix this with a suitable `sed` invocation).

tombentley (commented):
This looks fine to me except the non-standard `/** Blah blah` lines where the doc is not just a one-liner (i.e. where the line doesn't end `*/`).

topic-operator/src/main/java/io/strimzi/operator/topic/ConflictingChangesException.java
ShubhamRwt: Fixed @tombentley

tombentley (approved):
Thanks @ShubhamRwt

PR #7854, opened 2022-12-23
Sample Prometheus configuration change

tgib23:
### Type of change

- Bugfix

### Description

In the sample Prometheus configuration, ServiceAcount `prometheus-server` is created without namespace specified, but it's used with namespace `myproject`.

### Checklist

- [ ] Write tests
- [ ] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/documentation/modules/metrics/proc_metrics-deploying-prometheus.adoc
+++ b/documentation/modules/metrics/proc_metrics-deploying-prometheus.adoc
@@ -28,18 +28,6 @@ NOTE: By default, the Prometheus Operator only supports jobs that include an `en
 
 .Procedure
 
-. Modify the Prometheus installation file (`prometheus.yaml`) according to the namespace Prometheus is going to be installed into:
-+
-On Linux, use:
-+
-[source,shell,subs="+quotes,attributes"]
-sed -i 's/namespace: .*/namespace: _my-namespace_/' prometheus.yaml
-+
-On MacOS, use:
-+
-[source,shell,subs="+quotes,attributes"]
-sed -i '' 's/namespace: .*/namespace: _my-namespace_/' prometheus.yaml
-
 . Edit the `PodMonitor` resource in `strimzi-pod-monitor.yaml` to define Prometheus jobs that will scrape the metrics data from pods.
 +
 Update the `namespaceSelector.matchNames` property with the namespace where the pods to scrape the metrics from are running.
--- a/examples/metrics/prometheus-install/prometheus.yaml
+++ b/examples/metrics/prometheus-install/prometheus.yaml
@@ -43,7 +43,6 @@ roleRef:
 subjects:
   - kind: ServiceAccount
     name: prometheus-server
-    namespace: myproject
 
 ---
 apiVersion: monitoring.coreos.com/v1
--- a/examples/metrics/prometheus-install/prometheus.yaml
+++ b/examples/metrics/prometheus-install/prometheus.yaml
@@ -43,6 +43,7 @@ roleRef:
 subjects:
   - kind: ServiceAccount
     name: prometheus-server
+    namespace: default
 
 ---
 apiVersion: monitoring.coreos.com/v1
@@ -68,7 +69,7 @@ spec:
       app: strimzi
   alerting:
     alertmanagers:
-    - namespace: myproject
+    - namespace: default
       name: alertmanager
       port: alertmanager
   additionalScrapeConfigs:


scholzj (commented):
Sorry, but I do not think this is how it works. I do not think your change is valid. Did you tried to use it?

The `ServiceAccount` is without a namespace because it takes the namespace from whatever is your default or passed from `kubectl` through `-n` or `--namespace`. So that is intentional. But the binding has to contain the namespace of the subject. So you have to keep it and change it in case you use some other namesapce.

tgib23:
Sorry, should have added more explanation.
I added this change to follow what is written in Strimzi docs.

https://strimzi.io/docs/operators/in-development/deploying.html#proc-metrics-deploying-prometheus-str
image

Because the sample command does not pass the namespace, this creates ServiceAccount in default namespace.
So, just following the docs doesn't complete metrics configuration and some trouble shooting is necessary, adding namespace to ServiceAccount creation, deleting namespace from binding part (this PR), or add -n in running those commands, etc.

As you mentioned, adding ServiceAccount in default namespace is not a good practice? Then, maybe it's better to change ServiceAccount creation part or document?

scholzj:
Fine, then you should open a PR to fix the docs here in `./documentation` if they are wrong and don't work. But the change you do here in this PR is not valid. The resource is not a valid ClusterRoleBinding and does not apply. So this is not a solution to anything I guess.

scholzj:
(A side note - any changes would need to be anyway done to `packaging/examples/...` - but I do not think that really matters here.)

tgib23:
>  But the binding has to contain the namespace of the subject.

MY BAD! Totally missed this point, so I just fixed the PR.

I also fixed the binding of alertmanager so that it's more aligned with the current document.
Then, modifying the namespace in prometheus.yaml is not necessary, so I deleted them, too.
With this change, I hope my intention is a bit more clear (Sorry for confusing!)

I also checked what changes would be necessary if we keep using non-default namespace (`namespace: myproject`) in prometheus.yaml, and following is the minimum requirement AFAIK

* Add `-n myproject` to following commands
```
kubectl apply -f prometheus-additional.yaml
kubectl apply -f strimzi-pod-monitor.yaml
kubectl apply -f prometheus-rules.yaml
kubectl apply -f prometheus.yaml
```
* Use `http://prometheus-operated.myproject:9090` instead of `http://prometheus-operated:9090` for Prometheus server URL

Not sure which is better solution as Strimzi docs. Any comment please?

scholzj:
Sorry, but no, this does not work.
* You just changed the namespace from `myproject` to `default`. That does not really have any real meaning. Sure, for someone `default` might be preferred, but `default` is a special namespace that should not be really used in many distributions because it has special properties. And there will be many people for whom it does not work anyway. Also, all of Strimzi uses `myproject` as the default namespace.
* You also changed the docs and removed the section to update the namespace. Again, this will be needed for most users as most users do not use `default` namespace.
* `-n myproject` does not fix these issues. if you want to ad it to some commands in the docs for more clarity, fine. But it only affects the namespace where the resource is created. It does not change he namespace in the specification inside the resources. So no, this is not sufficient on its own neither.

scholzj:
Closing as there was no feedback on this for 2 weeks.

tgib23:
@scholzj 

Sorry, I've been busy with year end/start family stuff, so away from this for a while

> You just changed the namespace from myproject to default. That does not really have any real meaning. Sure, for someone default might be preferred

It's not about me or others' preference. I thought the docs and sample files are intended to use `default` namespace for some objects, because they don't specify `-n [namespace]` and use `http://prometheus-operated:9090`, meaning Prometheus is working in `default` namespace, so I thought my PR would be the minimum changes to let all the metrics components work. But sure, I understand we should not specify `default` namespace to use.

>  Also, all of Strimzi uses myproject as the default namespace.

I think that's fair enough. Then, the rest of of my proposal would be to change `http://prometheus-operated:9090` to `http://prometheus-operated.myproject:9090 (replace myproject to your namespace)`. If you agree, I'd create another PR for that.

tgib23:
https://github.com/strimzi/strimzi-kafka-operator/blame/a229ec9f5bdb50405c3994acc6799d6392106b6f/documentation/modules/metrics/proc_metrics-grafana-dashboard.adoc#L79

scholzj:
Isn't the expectation that they run in the same namespace?

I don't know. But TBH, this is a never-ending story. Tomorrow someone will come and insist that there needs to be a command of how to find out what exactly _your namespace_ is and so on. I would love to see users think a bit when they follow the docs and not just blindly copy things and expect us to cover every single detail. Running something like Kafka on Kubernetes is a fairly advanced task so people should be able to adapt these things to their environment and specific needs without super granular guidance.

Sorry for the rant. If you want, feel free to open the PR to add the namespace to the URL.

PR #7858, opened 2022-12-23
docs(oauth): Provides example config for authentication methods

PaulRMellor:

**Documentation**

Updates [Configuring Kafka Java clients to use OAuth 2.0](https://strimzi.io/docs/operators/in-development/configuring.html#proc-oauth-client-config-str) to include example configuration for different authentication methods.

### Checklist

- [ ] Write tests
- [ ] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/documentation/modules/oauth/proc-oauth-client-config.adoc
+++ b/documentation/modules/oauth/proc-oauth-client-config.adoc
@@ -5,9 +5,18 @@
 [id='proc-oauth-client-config-{context}']
 = Configuring Kafka Java clients to use OAuth 2.0
 
-This procedure describes how to configure Kafka producer and consumer APIs to use OAuth 2.0 for interaction with Kafka brokers.
+[role="_abstract"]
+Configure Kafka producer and consumer APIs to use OAuth 2.0 for interaction with Kafka brokers.
+Add a callback plugin to your client `pom.xml` file, then configure your system properties for OAuth 2.0.
 
-Add a client callback plugin to your _pom.xml_ file, and configure the system properties.
+
+How you configure your system properties depends on the method you are using to authenticate your client with the Kafka brokers.
+You can use one of the following methods: 
+
+* The OAuth 2.0 client credentials mechanism
+* The (deprecated) OAuth 2.0 password grant mechanism
+* An access token
+* A refresh token
 
 .Prerequisites
 
@@ -28,10 +37,12 @@ Add a client callback plugin to your _pom.xml_ file, and configure the system pr
 </dependency>
 ----
 
-. Configure the system properties for the callback:
+. Configure the system properties for the callback.
 +
 For example:
 +
+--
+.Client credentials mechanism properties 
 [source,env, subs="+quotes,attributes"]
 ----
 System.setProperty(ClientConfig.OAUTH_TOKEN_ENDPOINT_URI, “_https://<auth-server-address>/auth/realms/master/protocol/openid-connect/token_”); <1>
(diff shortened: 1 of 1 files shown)


documentation/modules/oauth/proc-oauth-client-config.adoc
```diff
@@ -47,8 +58,49 @@ System.setProperty(ClientConfig.OAUTH_AUDIENCE, "_<audience_value_") <5>
...
++
+--
+.Password grants mechanism properties 
+[source,env, subs="+quotes,attributes"]
+----
+System.setProperty(ClientConfig.OAUTH_TOKEN_ENDPOINT_URI, “_https://<auth-server-address>/auth/realms/master/protocol/openid-connect/token_”);
+System.setProperty(ClientConfig.OAUTH_CLIENT_ID, "_<client_name>_"); <1>
+System.setProperty(ClientConfig.OAUTH_CLIENT_SECRET, "_<client_secret>_"); <2>
+System.setProperty(ClientConfig.OAUTH_PASSWORD_GRANT_USERNAME, "__<username>__"); <3>
+System.setProperty(ClientConfig.OAUTH_PASSWORD_GRANT_PASSWORD, "__<password>__"); <4>
```
scholzj, line 71: I find this way a bit weird. Why would someone really use this way to set up the properties? I guess the more logical configuration options are to use environment variables, pass system properties from outside using `-D` option or configure it directly in the `sasl.jaas.config` option in the Kafka configuration. E.g.

```
sasl.jaas.config=org.apache.kafka.common.security.oauthbearer.OAuthBearerLoginModule required \
  oauth.client.id="team-a-client" \
  oauth.client.secret="team-a-client-secret" \
  ...
  ...;
```

The mechanism to use system properties but configured from inside the application should technically work, but does seem like something one would use in production-grade applications. This applies to this whole section basically.

scholzj (approved):
Sjould be reviewed by Marko as the SME. But I'm fine with it in general. That said, I left a comment for consideration about something what seems a bit weird to me.

documentation/modules/oauth/proc-oauth-client-config.adoc
PaulRMellor: Like we do in the later examples. @mstruk -- should we change how we add the config here?  We might also want to mention the various options.

mstruk: Yes, I guess let's make it the way we ourselves configure it as @scholzj suggested.

I'm guessing that we chose this style because it's less error prone. Setting `sasl.jaas.config` requires following a strict formatting, it's supposed to be a 'single line' (which `\n` at the end of each line ensures), the config parameter names _must not_ be quoted, but all the values _have to be_ quoted. It has to start exactly like it does, contain the keyword `required`, and it has to end with `;`.


In this case, for client credentials it would be:
```
sasl.jaas.config=org.apache.kafka.common.security.oauthbearer.OAuthBearerLoginModule required \
  oauth.token.endpoint.uri="_<token_endpoint_url>_" \
  oauth.client.id="_<client_id>_"  <1> \
  oauth.client.secret="_<client_secret>_" <2> ;
```

For 'password grant' we would have:
```
sasl.jaas.config=org.apache.kafka.common.security.oauthbearer.OAuthBearerLoginModule required \
  oauth.token.endpoint.uri="_<token_endpoint_url>_" \
  oauth.client.id="_<client_id>_"  <1> \
  oauth.client.secret="_<client_secret>_" <2> \
  oauth.password.grant.username="__<username>__"  <3> \
  oauth.password.grant.password="__<password>__"  <4> ;
```

For access token we would have:
```
sasl.jaas.config=org.apache.kafka.common.security.oauthbearer.OAuthBearerLoginModule required \
  oauth.token.endpoint.uri="_<token_endpoint_url>_" \
  oauth.access.token="_<access_token>_"  <1> ;
```

And for refresh token it would be:
```
sasl.jaas.config=org.apache.kafka.common.security.oauthbearer.OAuthBearerLoginModule required \
  oauth.token.endpoint.uri="_<token_endpoint_url>_" \
  oauth.client.id="_<client_id>_"  <1> \
  oauth.client.secret="_<client_secret>_" <2> \
  oauth.refresh.token="_<refresh_token>_"  <3> ;
```

mstruk (commented):
It makes sense to just load the whole client properties file in one go, rather than set the properties one by one:

```
Properties props = new Properties();
try (FileReader reader = new FileReader("client.properties", StandardCharsets.UTF_8)) {
  props.load(propsReader);
}
```

Where `client.properties` file contains:

```
security.protocol=SASL_PLAINTEXT
sasl.mechanism=OAUTHBEARER
sasl.jaas.config=org.apache.kafka.common.security.oauthbearer.OAuthBearerLoginModule required \
  oauth.token.endpoint.uri="_<token_endpoint_url>_" \
  ... ;
```

PaulRMellor:
> Properties props = new Properties();
> try (FileReader reader = new FileReader("client.properties", StandardCharsets.UTF_8)) {
>   props.load(reader);
> }

Thanks @mstruk. I've updated the procedure to show the use of properties file, then loading as suggested. I've also added a bit more of an explanation as to what we are configuring and what options are available in the introduction.

mstruk:
Looks good. Another question ...

Do we cover somewhere else the SASL_SSL options:
`ssl.truststore.location`
`ssl.truststore.password`
`ssl.truststore.type`

These are needed when `security.protocol=SASL_SSL` is set.

And the OAuth config parameters:
`oauth.ssl.truststore.location`
`oauth.ssl.truststore.password`
`oauth.ssl.truststore.type`

Needed when secure connection (`https://`) to authorization server is used, which should always be the case for production deployment.

documentation/modules/oauth/proc-oauth-client-config.adoc
```diff
@@ -28,46 +57,101 @@ Add a client callback plugin to your _pom.xml_ file, and configure the system pr
...
+security.protocol=SASL_SSL
+sasl.mechanism=OAUTHBEARER
+sasl.jaas.config=org.apache.kafka.common.security.oauthbearer.OAuthBearerLoginModule required \
+  oauth.token.endpoint.uri="_<token_endpoint_url>_" \
+  oauth.client.id="_<client_id>_" \ # <1>
+  oauth.client.secret="_<client_secret>_" \ # <2>
+  oauth.password.grant.username="_<username>_" \ # <3> 
+  oauth.password.grant.password="_<password>_" \ # <4> 
+  oauth.scope="_<scope>_" \
+  oauth.audience="_<audience>_" ;
```
mstruk, line 105: Maybe add comments here like in line 78.

PaulRMellor: Not sure we need to repeat. We haven't repeated the callouts of `security.protocol` or `sasl.mechanism`

PaulRMellor:
> ssl.truststore.location

@mstruk - I've added the truststore config. More descriptive on the authorization server config, as we describe the client truststore config for brokers elsewhere in the docs

mstruk (approved):

PR #7874, opened 2023-01-05
Refactor labels handling in model classes

scholzj:
### Type of change

- Refactoring

### Description

This PR refactors how we handle labels in the `AbstractModel` class and in the classes extending it. It tries to address the following things:
* The `strimzi.io/name` label seems to be currently used in a very messy way. Some resources use it for their exact name (e.g. Deployments, StrimziPodSets, StatefulSets, or some Services). Some use it to reference the Deployment / StatefulSet / StrimziPodSet they belong to -> this applies to the Pods, some Services, some Secrets, or Config Maps. Others for some reason have the name simply set to `strimzi`. The label seems to be broken. But since in some cases it is used in selectors, it is hard to simply remove it. So this PR at least tries to give it some consistency and sets it to the name of the _compoenent_ everywhere (this in general is the name of the StatefulSet / StrimziPodSet / Deployment - e.g. `my-cluster-kafka`, `my-cluster-cruise-control` etc.). This was already the case for places where we use selectors, so they should not be broken by this.
* Adds a new label `strimzi.io/component` which has the type o the component the resource belongs to. This is for example `kafka`, `zookeeper`, `cruise-control` etc. We were missing such label before, but it is useful for example for pod scheduling to target the pods from the same components but belonging to other clusters. E.g. when you want to make sure your Kafka broker pod does not share the node with any other Kafka broker pod even from a different Kafka cluster.
* Using the labels consistently allows cleaning some of the different auxiliary methods in the `AbstractModel` class to get the different label sets. You can now in most cases use the `labels` field directly or just with the additional labels from the template. This is important, because it will make it easier to refactor the `AbstractModel` in the future into multiple smaller utils classes with static method and pass there just the `Labels` instance.
* It also moves the `name` field into the `AbstractModel` class. This field was used in all the other classes, so there is no reason to not have it in the `AbstractModel` class directly and it is used for the labels.

### Checklist

- [x] Write tests


cluster-operator/src/main/java/io/strimzi/operator/cluster/model/AbstractModel.java
```diff
@@ -321,15 +321,17 @@ public abstract class AbstractModel {
     /**
      * Constructor
      *
-     * @param reconciliation   The reconciliation
-     * @param resource         Kubernetes resource with metadata containing the namespace and cluster name
-     * @param applicationName  Name of the application that the extending class is deploying
+     * @param reconciliation    The reconciliation marker
+     * @param resource          Custom resource with metadata containing the namespace and cluster name
+     * @param name              Name of the Strimzi component usually consisting from the cluster name and component type
+     * @param componentType     Type of the application that the extending class is deploying
```
ppatierno, line 327: is "application" referring to the "Strimzi component" above? I know you are referring to Kafka, ZooKeeper, Cruise Control as "applications" but maybe we want to have consistency with "Strimzi component"? So changing in one place or the other?

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/CruiseControl.java
```diff
@@ -165,8 +165,8 @@ public class CruiseControl extends AbstractModel {
      * @param resource  Kubernetes resource with metadata containing the namespace and cluster name
      */
     protected CruiseControl(Reconciliation reconciliation, HasMetadata resource) {
-        super(reconciliation, resource, APPLICATION_NAME);
-        this.name = CruiseControlResources.deploymentName(cluster);
+        super(reconciliation, resource, CruiseControlResources.deploymentName(resource.getMetadata().getName()), APPLICATION_NAME);
```
ppatierno, line 168: should we change the `APPLICATION_NAME` constant to `COMPONENT_TYPE` now? or vice versa, kind of related to my comment on Javadoc before?

operator-common/src/main/java/io/strimzi/operator/common/model/Labels.java
```diff
@@ -48,22 +48,26 @@ public class Labels {
...
-     * The name of the K8S resource.
-     * This is often the same name as the cluster
-     * (i.e. the same as {@code strimzi.io/cluster})
-     * but is different in some cases (e.g. headful and headless services)
+     * Name of the Strimzi component to which given resource belongs. E.g. Kafka or ZooKeeper. This label does not
+     * depend on the name of the cluster. This is useful to identify resources which belong to the same component but
+     * different clusters which is useful for example for scheduling (e.g. when you do not want this broker to be
+     * scheduled on a node where any other Kafka broker is running).
+     */
+    public static final String STRIMZI_COMPONENT_LABEL = STRIMZI_DOMAIN + "component";
```
ppatierno, line 61: is this the label for the component type? I am a little bit lost across ... component, component type and application :-(

scholzj:
@ppatierno I tried to rename more of the fields to make it a bit more consistent. Hope that works for you. It probably created some noise in the diffs, but you are probably right that it makes sense to do it. Please have a look and let me know if this looks better or not.

ppatierno (approved):
Thanks Jakub. Makes a lot of more sense to me now. LGTM!

PR #7878, opened 2023-01-06
KRaft liveness and readiness phase 1

katheris:
### Type of change

- Enhancement / new feature

### Description

KRaft liveness and readiness phase 1
* If KRaft mode enabled, check controller process is running for both liveness and readiness.
* Move Kafka readiness logic to kafka_readiness.sh script.
* Update KafkaAgent to create ready file only if broker state is > 3 and != 127.

### Checklist

- [ ] Write tests
- [ ] Update documentation
- [x] Update CHANGELOG.md

### DIFF
--- a/kafka-agent/src/main/java/io/strimzi/kafka/agent/KafkaAgent.java
+++ b/kafka-agent/src/main/java/io/strimzi/kafka/agent/KafkaAgent.java
@@ -32,6 +32,9 @@ public class KafkaAgent {
     // KafkaYammerMetrics class in Kafka 3.2-
     private static final String YAMMER_METRICS_IN_KAFKA_3_2_AND_EARLIER = "kafka.metrics.KafkaYammerMetrics";
 
+    private static final Integer BROKER_RUNNING_STATE = 3;
+    private static final Integer BROKER_UNKNOWN_STATE = 127;
+
     private final File sessionConnectedFile;
     private final File brokerReadyFile;
     private MetricName brokerStateName;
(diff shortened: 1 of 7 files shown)


kafka-agent/src/main/java/io/strimzi/kafka/agent/KafkaAgent.java
```diff
@@ -159,11 +159,19 @@ public void run() {
             boolean handleBrokerState() {
                 LOGGER.trace("Polling {}", brokerStateName);
                 boolean ready = false;
-                Integer running = Integer.valueOf(3);
-                Object value = brokerState.value();
+                Integer runningState = 3;
+                Integer unknownState = 127;
```
scholzj, line 163: Would it make sense to have these as some static final constants?

scholzj (commented):
Thanks for the PR. One nit, LGTM otherwise.

kafka-agent/src/main/java/io/strimzi/kafka/agent/KafkaAgent.java
katheris: Yes, there's no reason for them to be set here

kafka-agent/src/main/java/io/strimzi/kafka/agent/KafkaAgent.java
```diff
@@ -159,11 +162,17 @@ public void run() {
...
                 LOGGER.trace("Polling {}", brokerStateName);
                 boolean ready = false;
-                Integer running = Integer.valueOf(3);
-                Object value = brokerState.value();
+                Object observedState = brokerState.value();
 
-                if ((value instanceof Integer && running.equals(value))
-                        || (value instanceof Byte && running.equals(((Byte) value).intValue()))) {
+                boolean stateIsRunning = false;
+                if (observedState instanceof Integer) {
```
fvaleri, line 168: Maybe I'm missing something, but the broker state is a byte value, so why not directly use `BrokerState` enum from Kafka, getting rid of `instanceof` checks and int conversions?

scholzj: Less dependencies?

fvaleri: OK, you can use byte constants if you prefer, but the `observedState` will never be an int (unless Kafka changes the public enum), so this code is effectively unreachable.

scholzj (approved):
LGTM. Thanks @katheris

kafka-agent/src/main/java/io/strimzi/kafka/agent/KafkaAgent.java
scholzj: I guess Kate (or someone else if you volunteer 😉) would need to test it with the supported versions if any of them might use the integer part. It might be a leftover from Kafka 2.5.0. (This is also the reason why pulling in `BrokerState` is not a good idea. This stuff is not exactly public API which is not changing.)

fvaleri: Sure, I'll do that.

ppatierno (approved):
LGTM. Thanks @katheris!

fvaleri:
@katheris @scholzj I did that change and tested on all supported Kafka versions (3.2.0, 3.2.1, 3.2.3, 3.3.1) with Minikube and OpenShift 4.11.

scholzj:
Thanks @katheris

showuon (commented):
Sorry for being late. LGTM!

PR #7884, opened 2023-01-09
Using one HttpClient for asking user task status and rebalance per `KafkaRebalance` reconcile

### LINKED ISSUE #6739: Creating one HTTP client per `KafkaRebalance` reconcile
With the current implementation in the `HttpClientUtils` class, a new `HttpClient` instance is created and closed every time.

https://github.com/strimzi/strimzi-kafka-operator/blob/main/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/resource/HttpClientUtils.java#L27

This is mostly used in the Kafka Connect and Cruise Control API.
Mostly related to the Cruise Control API, it means that after asking for a rebalance, the periodic check on the user task status, done every 5 secs, creates a new HTTP client every time. So assuming one rebalance request a new client is created, then assuming some time for having the rebalance done one to N clients created, during the same reconcile.
This could lead to use a lot of more resources if the cluster operator is handling more and more Kafka instances.

One improvement could be creating just one HTTP client per reconcile (in the `KafkaRebalanceAssemblyOperator.reconcileRebalance` so that the same is used for asking a rebalance and checking the user task status.

ShubhamRwt:

### Type of change

- Enhancement / new feature

### Description

This PR fixes #6739. With these changes we will no longer have to create the HttpClient every time we ask for a rebalance or we need to check the user task status hence saving resources.

### Checklist

- [ ] Write tests
- [ ] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceAssemblyOperator.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceAssemblyOperator.java
@@ -761,7 +761,8 @@ public class KafkaRebalanceAssemblyOperator
 
     private Future<MapAndStatus<ConfigMap, KafkaRebalanceStatus>> onNew(Reconciliation reconciliation,
                                                                         String host, CruiseControlApi apiClient, KafkaRebalance kafkaRebalance,
-                                                                        AbstractRebalanceOptions.AbstractRebalanceOptionsBuilder<?, ?> rebalanceOptionsBuilder, HttpClient httpClient) {
+                                                                        AbstractRebalanceOptions.AbstractRebalanceOptionsBuilder<?, ?> rebalanceOptionsBuilder,
+                                                                        HttpClient httpClient) {
         return requestRebalance(reconciliation, host, apiClient, kafkaRebalance, true, rebalanceOptionsBuilder, httpClient);
     }
 
@@ -782,7 +783,8 @@ public class KafkaRebalanceAssemblyOperator
                                                                              String host, CruiseControlApi apiClient,
                                                                              KafkaRebalance kafkaRebalance,
                                                                              KafkaRebalanceAnnotation rebalanceAnnotation,
-                                                                             AbstractRebalanceOptions.AbstractRebalanceOptionsBuilder<?, ?> rebalanceOptionsBuilder, HttpClient httpClient) {
+                                                                             AbstractRebalanceOptions.AbstractRebalanceOptionsBuilder<?, ?> rebalanceOptionsBuilder,
+                                                                             HttpClient httpClient) {
 
         if (shouldRenewRebalance(kafkaRebalance, rebalanceAnnotation)) {
             // CC is activated from the disabled state or the user has fixed the error on the resource and want to 'refresh'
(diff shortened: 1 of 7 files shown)


scholzj (commented):
Sorry, I don't follow the logic here. The `withHttpClient` method works as a wrapper which makes sure that:
* The client is created for the operation
* It is used within the operation
* It is closed regardless how the operation ends

Your `withHttpClientCruiseControl` variant does not seem to make sense: 
* It gets the client from outside 
* It does not care about closing it after the operation is complete (makes sense)
* Closes the client in case of errors (makes no sense -> the client was created elsewhere, what if someone else uses it but you already closed it here?)

Maybe I missed something, but if you want to share the client across more operations, you should stick with the existing method but move it to an upper level in the operation tree. Following the model of the `withHttpClient` method is also important to make sure we don't leak the client as it was happening in the past for example.

ppatierno:
My initial idea was to have the Http client self-contained in the CruiseControlApiImpl following its lifecycle.
I mean one instance of the client used by the CC API implementation for each operation.
I think the `withHttpClient` method was thought as an util to get a client, use and then close it for kind of more one shot, rare operations, not something like an HTTP polling we need with CC usage.

scholzj:
> My initial idea was to have the Http client self-contained in the CruiseControlApiImpl following its lifecycle.
I mean one instance of the client used by the CC API implementation for each operation.
I think the withHttpClient method was thought as an util to get a client, use and then close it for kind of more one shot, rare operations, not something like an HTTP polling we need with CC usage.

Well, if you want to do it differently, you can. But I think the current method nicely wraps it and makes sure it is closed. The new one does not guarantee it. So you would need to be super careful to make sure it is properly closed. Can't you just use the `withHttpClient` but use it at a higher level above where you create the `CruiseControlApiImpl`? Or is the `CruiseControlApiImpl` used as the resource operator for example and exists the whole lifecycle of the operator?

ppatierno:
The `withHttpClient` creates, uses and closes the Http client instance everytime, so wherever we use it (even at higher level), it behaves this way and the Http client cannot be reused.
About the other point, I checked and we are instatiating the `CruiseControlApiImpl` on each reconcile which makes all this discussion kind of useless. I agree that we could retain the Http client instance if we use just one instance of the `CruiseControlApiImpl` for the entire lifecycle of the operator but I am not sure we can. On creation, it gets Secrets with certificates to connect to Cruise Control and during the cluster lifecycle they can change.
If we decided to have just one instance of the `CruiseControlApiImpl` (with one Http client) we should carefully handle the scenario where certificates change and we need the recreation of that client.
So not sure at this point if the change is worth the effort.

scholzj:
> The withHttpClient creates, uses and closes the Http client instance everytime, so wherever we use it (even at higher level), it behaves this way and the Http client cannot be reused.

Well, you pass the operation to it. It is up to you whether the operation is just a single HTTP query or whether it is the whole KafkaRebalance reconciliation. So you can for example wrap the whole `KafkaRebalanceAssemblyOperator.reconcileRebalance` method if you want and share the client through it, or?

kyguy (commented):
Hey @ShubhamRwt we could address the issue and cut down A LOT of lines by calling `CruiseControlApiImpl` directly in the `KafkaRebalanceAssemblyOperator` class here [1] and passing the `vertx` instance as a parameter. Then we could create the HTTP client in the constructor of the CruiseControlApiImpl class and use it for all of our API calls. See my comments above.

This refactoring would reduce us to one HTTP client per reconcile without having to pass an `HTTP client` around

[1] https://github.com/strimzi/strimzi-kafka-operator/blob/0.32.0/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceAssemblyOperator.java#L1201

ShubhamRwt:
Closing this since it needs to be done differently.

PR #7896, opened 2023-01-11
Reset the auto-restart status if the connector is running after the restart

### LINKED ISSUE #7891: Connector auto-restart counter does not reset back to 0
When auto-restarting connectors, we want to avoid trying to restart a failing connector again and again. So we use a counter with a timestamp to do a limited backoff. The connector is restarted only up to 7 times and each restart is done later after the previous one. (E.g. first restart is immediate, the second restart will not be done sooner than 2 minutes after the first, the third will be done at least 6 minutes after the second, etc.) The latest state is tracked in the status section of the `KafkaConnector` CR - it stores the number of restarts already done and the timestamp of the last restart.

This works as designed. If your connector is broken and fails all the time, we will not keep restarting it forever and force you to deal with it. But also, if you have a longer issue - e.g. a 10-minute network outage - the backoff will try to restart it not only immediately but also a bit later.

However, there is currently no reset to the tracked restart state. If your connector fails for example, once per week but works fine in between, it will still accumulate the restarts in its status. And after 7 weeks, it will reach the maximal count of 7 restarts and will not be considered for restarts anymore.

We need to find a middle ground - a mechanism that after some time decides the connector is fixed and resets the restart counter. That way, the next restart sequence will start from scratch. That way, the auto-restart can be useful even for longer time periods. For example, if the connector is running fine for N minutes, the operator would reset the counter to 0. The time might be configurable in the auto-restart section in `.spec` to handle different requirements for different situations.

scholzj:
### Type of change

- Bugfix

### Description

As described in #7891, we currently never reset the auto-restart status during normal operations. The only changing the configuration resets the status. So after your connector is restarted 7 times - regardless of during which period, it could be even weeks or months - it will not be restarted anymore until some configuration change happens and resets it. This PR adds functionality to reset the auto-restart status when the connector is running some time after the restart. 

Instead of an interval configurable in the auto-restart spec as originally suggested in the issue, it uses the same back-off mechanism as used for the restart. For example:
* if the connector was restarted just once, then the auto-restart status will be reset if it is running at least 2 minutes after the restart
* if the connector was restarted already trice, then the auto-restart status will be reset if it is running at least 12 minutes after the restart

Currently, the reset mechanism does not check that the connector is running the whole back-off interval. It checks only that it is in a running state in the next reconciliation after the corresponding back-off interval is over. I tried to implement it, but it is prone to cycle the operator with misbehaving connectors which are often changing the status as it ends up looping through updating the status fields tracking for how long it runs without interruption. That is why I in the end decided to abandon this and simply check if it is running after the interval. I think the solution implemented here at the end offers a good combination of simplicity, and fault-tolerance and at the same time delivers value in most use cases.
This should close #7891

### Checklist

- [x] Write tests

### DIFF
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/AbstractConnectOperator.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/AbstractConnectOperator.java
@@ -35,6 +35,7 @@ import io.strimzi.api.kafka.model.KafkaConnectorSpec;
 import io.strimzi.api.kafka.model.KafkaMirrorMaker2;
 import io.strimzi.api.kafka.model.connect.ConnectorPlugin;
 import io.strimzi.api.kafka.model.status.AutoRestartStatus;
+import io.strimzi.api.kafka.model.status.AutoRestartStatusBuilder;
 import io.strimzi.api.kafka.model.status.Condition;
 import io.strimzi.api.kafka.model.status.KafkaConnectStatus;
 import io.strimzi.api.kafka.model.status.KafkaConnectorStatus;
@@ -613,8 +614,22 @@ public abstract class AbstractConnectOperator<C extends KubernetesClient, T exte
         }
     }
 
+    /**
+     * Handles auto-restarting of the connectors and managing the auto-restart status. It checks that current state of
+     * the connector and its tasks. If it is failing, it will restart them in periodic intervals with backoff. If the connector is stable after the restart, it resets that auto-restart status.
+     *
+     * @param reconciliation    Reconciliation marker
+     * @param host              Kafka Connect host
+     * @param apiClient         Kafka Connect REST API client
+     * @param connectorName     Name of the connector
+     * @param connectorSpec     Spec of the connector
+     * @param status            The new/future status of the connector
+     * @param resource          The custom resource used to get the existing status of the connector
+     *
+     * @return  Future with connector status and conditions which completes when the auto-restart is handled
+     */
     @SuppressWarnings({ "rawtypes" })
-    private  Future<ConnectorStatusAndConditions> autoRestartFailedConnectorAndTasks(Reconciliation reconciliation, String host, KafkaConnectApi apiClient, String connectorName, KafkaConnectorSpec connectorSpec, ConnectorStatusAndConditions status, CustomResource resource) {
+    /* test */ Future<ConnectorStatusAndConditions> autoRestartFailedConnectorAndTasks(Reconciliation reconciliation, String host, KafkaConnectApi apiClient, String connectorName, KafkaConnectorSpec connectorSpec, ConnectorStatusAndConditions status, CustomResource resource) {
         JsonObject statusResultJson = new JsonObject(status.statusResult);
         if (connectorSpec.getAutoRestart() != null && connectorSpec.getAutoRestart().isEnabled()) {
             return getPreviousKafkaConnectorStatus(connectorOperator, resource)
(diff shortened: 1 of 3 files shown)


tombentley:
The code changes LGTM.

2 minutes isn't long for someone to observe from the status the fact that an auto-restart happened. I wonder if we should have a metric for the number of times we've auto-restarted the connector or a task. That would provided a more persistent history, enough to get a timestamp from which to look at the logs. I know we already have the count of the number of successul/unsuccessful reconciliations, but that's not exactly the same thing (TBH it's not immediately clear to me whether a reconciliation in which we auto-restarted a connector is counted as successful or not wrt that metric: If it counts as unsuccessful then I guess we wouldn't need a metric).

scholzj:
> 2 minutes isn't long for someone to observe from the status the fact that an auto-restart happened. I wonder if we should have a metric for the number of times we've auto-restarted the connector or a task. That would provided a more persistent history, enough to get a timestamp from which to look at the logs. I know we already have the count of the number of successul/unsuccessful reconciliations, but that's not exactly the same thing (TBH it's not immediately clear to me whether a reconciliation in which we auto-restarted a connector is counted as successful or not wrt that metric: If it counts as unsuccessful then I guess we wouldn't need a metric).

IIRC it does not fail the reconciliation. The connector operator does not consider a failing connector as a reconciliation failure. It just updates the status. (This is not a new behavior related to the auto-restarts. I think this is like it since the beginning)

I do not think the auto-restart status is the best source for monitoring this really. So it did not come to my mind to see this issue. And to raise awareness I moved the related log messages to the INFO level from DEBUG. But I think you have a good point. I raised #7900 to track the metrics. I think it would be a useful addition and allows you to build alerting etc.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/AbstractConnectOperator.java
```diff
@@ -652,29 +677,89 @@ private  Future<ConnectorStatusAndConditions> autoRestartFailedConnectorAndTasks
...
+     * @param host                          Kafka Connect host
+     * @param apiClient                     Kafka Connect REST API client
+     * @param connectorName                 Name of the connector
+     * @param status                        The new/future status of the connector
+     * @param previousAutoRestartStatus     Previous auto-restart status used to generate the new status (increment the restart counter etc.)
+
+     * @return  Future with connector status and conditions which completes when the connector is restarted
+     */
+    private Future<ConnectorStatusAndConditions> autoRestartConnector(Reconciliation reconciliation, String host, KafkaConnectApi apiClient, String connectorName, ConnectorStatusAndConditions status, AutoRestartStatus previousAutoRestartStatus) {
+        LOGGER.infoCr(reconciliation, "Auto restarting connector {}", connectorName);
```
katheris, line 692: Noting that this is an increase in logging level a previously this log was at debug level. I prefer having it at info level as I think it is useful for the user to know when strimzi has restarted a connector.

cluster-operator/src/test/java/io/strimzi/operator/cluster/operator/assembly/KafkaConnectAssemblyOperatorTest.java
```diff
@@ -962,47 +964,396 @@ public void testShouldAutoRestartConnector(VertxTestContext context) {
...
             .withCount(7)
             .withLastRestartTimestamp(ZonedDateTime.now(ZoneOffset.UTC).minusDays(1).format(DateTimeFormatter.ISO_INSTANT))
             .build();
-        assertThat(op.shouldAutoRestart(autoRestartStatus, status), is(false));
+        assertThat(op.shouldAutoRestart(autoRestartStatus), is(false));
 
         context.completeNow();
     }
 
+    @Test
```
katheris, line 1005: You could do this with a parametrized test if you wanted to have different tests for each case. It would potentially make it easier to see which one failed and less code overall, but would mean it instantiates the operator each time.

katheris (approved):
LGTM - I noted a change in the logging level, but prefer the new level. Also had a suggestion on the tests, but I notice the existing tests are not parametrized either.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/AbstractConnectOperator.java
scholzj: That was intentional. I think it is important enough information to be logged at INFO which is what most users use.

cluster-operator/src/test/java/io/strimzi/operator/cluster/operator/assembly/KafkaConnectAssemblyOperatorTest.java
scholzj: Yeah, it is basically a copy-paste of the previous test. I think it is fine the way it is - its just a few variations.

Issue #2266, opened 2019-12-02
[Enhancement] Allow inter-broker TLS to be disabled


dalelane:
**Is your feature request related to a problem? Please describe.**
Disabling TLS for Kafka's inter-broker connections would improve the performance. For installs where security isn't a concern, such as those running on private isolated secure clusters, inter-broker TLS may not be needed.

We'd like a way to disable TLS for these cases, so users can decide whether they want to prioritise performance or security.

**Describe the solution you'd like**
A new boolean attribute in KafkaClusterSpec that enables/disables inter-broker TLS. 

e.g. 

```
apiVersion: kafka.strimzi.io/v1beta1
kind: Kafka
metadata:
  name: my-cluster
spec:
  kafka:
    version: 2.3.1
    replicas: 3
    interBrokerTls: false
```

The new attribute would default to `true` to maintain current behaviour. Existing CR definitions need not be affected, but there would be an option here to edit an existing cluster by setting this to `false`. (Rolling updates that modify this would require persistence, as new brokers wouldn't be able to talk to the previous generation brokers so replication wouldn't be possible during the rolling upgrade.)

**Describe alternatives you've considered**
Alternatively, we could treat this as a new type of listener, that can be configured - calling it something like `interbroker` or `replication`:

e.g.
```
apiVersion: kafka.strimzi.io/v1beta1
kind: Kafka
metadata:
  name: my-cluster
spec:
  kafka:
    version: 2.3.1
    replicas: 3
    listeners:
      plain: {}
      tls: {}
      interbroker:
        tls: false
```

If we think we'll ever want to make other configuration changes to the inter-broker listeners, this might be a preferable approach. Plus, semantically it feels like a good fit to configure it here rather than as an attribute of the Kafka cluster overall. 

**Additional context**
None.

scholzj:
I think the `spec.kafka.listeners.interbroker` looks good in the current API state - it fits well and allows to add more options in the future. However, we had some thoughts (and user requests) about giving users more freedom to define the listeners. E.g. to change to something more "free":

```yaml
listeners:
 - name: plain
    type: internal
    port: 9092
    tls: false
    authentication:
      type: scram-sha522
 - name: tls
    type: internal
    tls: true
    port: 9093
    authentication:
      type: tls
 - name: external1
    type: external
    tls: true
    port: 9094
    authentication:
      type: tls
 - name: external2
    type: external
    tls: true
    port: 9095
    authentication:
      type: tls
```

[...]

PR #7901, opened 2023-01-12
[Enhancement] Add new interBrokerTls boolean flag

frankgrimes97:
The new flag defaults to 'true' to preserve the existing behavior of configuring the ControlPlane and Replication listeners with TLS. When set to 'false', they are created in PLAINTEXT protocol mode.

https://github.com/strimzi/strimzi-kafka-operator/issues/2266

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@
 * `ImageStream` validation for Kafka Connect builds on OpenShift
 * Support for configuring the metadata for the Role / RoleBinding of Entity Operator
 * Add liveness and readiness probes specifically for nodes running in KRaft combined mode
+* Add new interBrokerTls boolean flag [#2266](https://github.com/strimzi/strimzi-kafka-operator/issues/2266)
 
 ### Changes, deprecations and removals
 
--- a/api/src/main/java/io/strimzi/api/kafka/model/KafkaClusterSpec.java
+++ b/api/src/main/java/io/strimzi/api/kafka/model/KafkaClusterSpec.java
@@ -35,7 +35,8 @@ import java.util.Map;
 @JsonInclude(JsonInclude.Include.NON_NULL)
 @JsonPropertyOrder({
     "version", "replicas", "image", "listeners", "config", "storage", "authorization", "rack", "brokerRackInitImage",
-    "livenessProbe", "readinessProbe", "jvmOptions", "jmxOptions", "resources", "metricsConfig", "logging", "template"})
+    "livenessProbe", "readinessProbe", "jvmOptions", "jmxOptions", "resources", "metricsConfig", "logging", "template",
+    "interBrokerTls"})
 @EqualsAndHashCode
 public class KafkaClusterSpec implements HasConfigurableMetrics, UnknownPropertyPreserving, Serializable {
 
@@ -71,6 +72,7 @@ public class KafkaClusterSpec implements HasConfigurableMetrics, UnknownProperty
     private List<GenericKafkaListener> listeners;
     private KafkaAuthorization authorization;
     private KafkaClusterTemplate template;
+    private boolean interBrokerTls = true;
     private Map<String, Object> additionalProperties = new HashMap<>(0);
 
     @Description("The kafka broker version. Defaults to {DefaultKafkaVersion}. " +
@@ -249,6 +251,16 @@ public class KafkaClusterSpec implements HasConfigurableMetrics, UnknownProperty
 
     public void setTemplate(KafkaClusterTemplate template) {
         this.template = template;
+    }
+
+    @Description("Inter-broker communication should be over TLS. Defaults to 'true'." +
+            "Affects CONTROLPLANE and REPLICATION listener creation.")
+    public boolean getInterBrokerTls() {
+        return interBrokerTls;
+    }
+
+    public void setInterBrokerTls(boolean interBrokerTls) {
+        this.interBrokerTls = interBrokerTls;
     }
 
     @Override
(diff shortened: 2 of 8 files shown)


scholzj (changes requested):
Thanks for the PR. This would be a nice addition. However, I'm afraid this is a bit over simplified. For a start, this is significant enough feature to require a proposal (https://github.com/strimzi/proposals) as indicated in the issue.

I (personally) also thing that for it to make sense it would need to cover following features:
* Proper thought through API design to make sure this makes some long term sense because once you add something it cannot be changed and the design in this PR is definitely not something what can be built on in the future
* ZooKeeper
* All the other components (the way this PR is done, the Topic and User operator, Cruise Control, KAfka Exporter will be all broken)
* Alternative form of authentication for the communication because without it you loose a lot of features such as authorization. And even if it would be without any security, you would need to write some validation to make sure what is and isn't supported in this mode.
* The secondary resources such as dealing with CA, TLS certificates etc. when TLS is not used
* System tests
* Documentation

There is probably more, but these are just the things which come to my mind as something the proposal should cover and explain how they will be implemented.

frankgrimes97:
@scholzj Hi, first of all thank you for the quick and informative response!

For context, we have been successfully running Strimzi at scale for a number of months now, and like others have an interest in being able to run it without TLS for inter-broker communication both for increased throughput performance and also operational simplicity in our isolated environments.

Being new to the Strimzi codebase itself I don't think we (yet) have a broad enough grasp of all the interdependencies to come up with a detailed/comprehensive up-front proposal.
Nonetheless, I decided to try to dive in to (naively) implement the feature as initially proposed with the following purposes in mind:

1) Get familiar with the Strimzi codebase, its development and test suite
2) Prompt a revival of interest/attention in the requested feature enhancement
3) Signal our willingness to engage in trying to get it implemented

I see from your response that simply changing the CONTROLPLANE and REPLICATION listeners to PLAINTEXT is not sufficient and has cascading repercussions not obvious from simply running the local test suite.

Perhaps a good place to start is for us to understand which of the components you listed (ZooKeeper, Topic and User operator, Cruise Control, Kafka Exporter) connect to each of those listeners?

Many thanks!

scholzj:
> For context, we have been successfully running Strimzi at scale for a number of months now, and like others have an interest in being able to run it without TLS for inter-broker communication both for increased throughput performance and also operational simplicity in our isolated environments.

There is certainly some interest in it. I would say that the interest is not _big enough_ for it to be on top of the priority list for anyone in the core team. But it is high enough to make it interesting contribution. However, we have to make sure that the way it is done is sustainable, that it can be supported and maintained going forward, and that it can be built on in the future (this feature would be for example a good basis for some tighter integration with some service mesh, etc.). So that is why the proposal is important and that is why it is sometimes necessary to implement a lot more to get a contribution like this accepted and merged than just some smaller part which would be sufficient for you but might not allow us to move forward easily.

> Being new to the Strimzi codebase itself I don't think we (yet) have a broad enough grasp of all the interdependencies to come up with a detailed/comprehensive up-front proposal. Nonetheless, I decided to try to dive in to (naively) implement the feature as initially proposed with the following purposes in mind:
> 
> 1. Get familiar with the Strimzi codebase, its development and test suite
> 2. Prompt a revival of interest/attention in the requested feature enhancement
> 3. Signal our willingness to engage in trying to get it implemented

To be honest, this is probably one of the most complex tasks we have in the backlog. I don't think it is something that can be done in a few days (it is at the end one of the reasons why nobody has done it so far). That said - if you want to work on it, it does not mean that you are expected to come up with the complete proposal on your own out of the blue. We can help to guide you through it, we can have some calls to discuss the details of the design, etc. Judging from this PR, it seems you got a fairly good grip on the codebase. So it might be more about identifying all the missing pieces, deciding how to configure it, etc.

So if you would be interested to go through this, then we will try our best to help. 

> Perhaps a good place to start is for us to understand which of the components you listed (ZooKeeper, Topic and User operator, Cruise Control, Kafka Exporter) connect to each of those listeners?

Well, ZooKeeper is a bit separate. Kafka connects to ZooKeeper and not ZooKeeper to Kafka. However, it probably makes little sense to keep the security tight around ZooKeeper and remove it from Kafka. So I think it would make sense to remove the security there as well.

The rest of the components all connect to the replication interface. In addition, the Cluster Operator is using it as well to manage the Kafka cluster.

frankgrimes97:
So, to start, I've been trying to familiarize myself with how the Zookeeper cluster gets initialized.

If the config needs to affect Zookeeper cluster creation as well, then perhaps KafkaCluster/KafkaClusterSpec isn't the right place for the `interBrokerTls` flag.
I see that KafkaAssemblyOperator handles creating both the Zookeeper and Kafka clusters but it's not immediately clear to me how that can take in the flag.
Should we consider putting `interBrokerTls` behind a Feature Gate so it can be read by the misc. clusters/services being created?

scholzj:
Well, whether it needs a feature flag or not depends on how it is implemented and designed. If it is something you have to enable through some option in the custom resource, then feature gate might not make much sense as you have the API configuration. But the API configuration probably should not be just a simple flag in order to be able to extend it, build on it, add different authentication types for the internal communication etc.

chenzhiwei:
Thank you for your PR, this is a long-awaited feature.

As this operator supports deploying multiple Kafka clusters, feature gate will force all Kafka clusters to have same behavior.

frankgrimes97:
We've dug some more into the code and now believe that the complexity of this new feature is beyond what we can reasonably commit to at this time.

Hronom:
@frankgrimes97 is there any workaround you find?

@scholzj why not help with this? I think for maintainers who is familiar with code base it will be easy to set the API or at least provide a way to workaround issue instead of chatting with people in PR's.

This is the blocker for service mesh setup. And from conversations here and there I see that maintainers actually blocking this instead of helping to resolve this issue:
https://github.com/strimzi/strimzi-kafka-operator/issues/8746
https://github.com/strimzi/strimzi-kafka-operator/issues/2266

scholzj:
I think the reasons why this PR was not merged are described above. The role of the maintainer is not to accept any PR someone opens. And it is also not to spend their time on whatever you think is the highest priority. The role is to make sure that the project development is sustainable. That includes making sure that any changes are right from the API design perspective, that we have the resources to maintain them, and that they can be built-on (and yes, disabling the TLS on the internal listeners is only a very small part of proper integration into service mesh, so there needs to be a proper plan how to do things), etc. We also didn't block anything. There is an open issue that tracks the enhancement request and I offered several times help to people interested in working on this.

Hronom:
My comment was in context then we target on result. Also I have assumption that for who is in context of how this repository maintained it will be simple change to do, comparing to people that run around and hit this wall.

PR #7903, opened 2023-01-13
Simplify `OwnerReference` handling in model classes

scholzj:
### Type of change

- Refactoring

### Description

This PR continues refactoring the `AbstractModel` class and its subclasses. It moves the creation of the OwnerReference to the `AbstractModel` constructor instead of calling the `createOwnerReference` method again and again to build the same reference. This also removes the need to call the `setOwnerReference` method from every subclass (not sure why we did it like that - maybe some regression failures will remind me 🤔).

Since the owner reference is now stored just as a field, it could be more easily passed to static methods for generating the resources in the future.

As part of these changes, I also slightly improved the tests to use one shared method for checking the owner reference and to compere it against the custom resource rather than just comparing the same field against each other.

### Checklist


ppatierno (approved):
LGTM!

PR #7916, opened 2023-01-14
Move PDB creation to `PodDisruptionBudgetUtils` class

scholzj:
### Type of change

- Refactoring

### Description

This PR continues refactoring the `AbstractModel` class and its subclasses. It moves the creation of Pod Disruption Budgets into static methods in a new class `PodDisruptionBudgetUtils`. The new methods take the full `PodDisruptionBudgetTemplate` object, so it is not necessary anymore to parse it in the `fromCrd` methods and we can also optimize the fields used to hold the PDB options in the model classes.

To make dealing with labels and annotations from the template easier, this PR also introduces a new interface `HasMetadataTemplate` which is used for all the template objects which have the metadata template. That allows us to create a new `TemplateUtils` class with static methods for parsing the labels and annotations out of the different templates. These handle all the null checks and make the code in the methods using them easier to read since it does not have to deal with them.

The Cruise Control model seemed to have the methods for using PDBs. But it actually does not use PDB. So these were removed without replacement.

### Checklist

- [x] Write tests

### DIFF
--- a/api/src/main/java/io/strimzi/api/kafka/model/template/HasMetadataTemplate.java
+++ b/api/src/main/java/io/strimzi/api/kafka/model/template/HasMetadataTemplate.java
@@ -1,0 +1,14 @@
+/*
+ * Copyright Strimzi authors.
+ * License: Apache License 2.0 (see the file LICENSE or http://apache.org/licenses/LICENSE-2.0.html).
+ */
+package io.strimzi.api.kafka.model.template;
+
+/**
+ * Interface used for template objects which allow to configure Kubernetes metadata. This is used to have a shard
+ * methods for getting the template values.
+ */
+public interface HasMetadataTemplate {
+    MetadataTemplate getMetadata();
+    void setMetadata(MetadataTemplate metadata);
+}
--- a/api/src/main/java/io/strimzi/api/kafka/model/template/HasMetadataTemplate.java
+++ b/api/src/main/java/io/strimzi/api/kafka/model/template/HasMetadataTemplate.java
@@ -9,6 +9,17 @@ package io.strimzi.api.kafka.model.template;
  * methods for getting the template values.
  */
 public interface HasMetadataTemplate {
+    /**
+     * Gets the metadata template
+     *
+     * @return  Metadata template
+     */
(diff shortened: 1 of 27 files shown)


api/src/main/java/io/strimzi/api/kafka/model/template/HasMetadataTemplate.java
```diff
@@ -0,0 +1,14 @@
...
+ */
+package io.strimzi.api.kafka.model.template;
+
+/**
+ * Interface used for template objects which allow to configure Kubernetes metadata. This is used to have a shard
+ * methods for getting the template values.
+ */
+public interface HasMetadataTemplate {
+    MetadataTemplate getMetadata();
+    void setMetadata(MetadataTemplate metadata);
```
ppatierno, line 13: should they have Javadoc even to avoid Java 17 complaining?

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/TemplateUtils.java
```diff
@@ -0,0 +1,48 @@
...
+
+import java.util.Map;
+
+/**
+ * Shared methods for working with Strimzi API templates
+ */
+public class TemplateUtils {
+    /**
+     * Extracts custom labels configured through the Strimzi API resource templates. This method deals the null checks
+     * and makes the code using it more eas to read.
```
ppatierno, line 17: typo `easy`

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/TemplateUtils.java
```diff
@@ -0,0 +1,48 @@
...
+                && template.getMetadata() != null) {
+            return template.getMetadata().getLabels();
+        } else {
+            return null;
+        }
+    }
+
+    /**
+     * Extracts custom annotations configured through the Strimzi API resource templates. This method deals the null
+     * checks and makes the code using it more eas to read.
```
ppatierno, line 34: typo `easy`

ppatierno (approved):
LGTM. Just a couple of nits.

PR #7920, opened 2023-01-15
Move PVC creation out of `AbstractModel` class

scholzj:
### Type of change

- Refactoring

### Description

This PR continues to refactor `AbstractModel` and its subclasses. It moves the methods used to generate PVCs which are used only by the Kafka and ZooKeeper clusters into a separate utils class. It also moves the Storage annotations from the `AbstractModel` to the `Annotations` class where most of our other annotations are and does some smaller improvements to other methods dealing with storage in `AsbtractModel` (such as using new Java 17 features to simplify the code).

### Checklist

- [x] Write tests


ppatierno (approved):
LGTM

PR #7923, opened 2023-01-16
Improve keycloak authorization example docs

mstruk:
Signed-off-by: Marko Strukelj <marko.strukelj@gmail.com>

### Type of change

- Documentation

### Description

Improve instructions for obtaining the CA certificates for the Kafka cluster, and the SSO.

### Checklist

- [ ] Write tests
- [x] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/documentation/modules/oauth/proc-oauth-authorization-keycloak-example.adoc
+++ b/documentation/modules/oauth/proc-oauth-authorization-keycloak-example.adoc
@@ -285,17 +285,6 @@ STOREPASS=storepass
 
 echo "Q" | openssl s_client -showcerts -connect $KAFKA_HOST_PORT 2>/dev/null | awk ' /BEGIN CERTIFICATE/,/END CERTIFICATE/ { print $0 } ' > /tmp/my-cluster-kafka.pem
 ----
-
-+
-If the client accesses the Kafka cluster over the ingress or OpenShift route, the Kafka bootstrap hostname is different, and you may need to specify it as an extra argument to the `openssl` command in order to get the certificate containing the proper SNA records:
-+
-[source,shell]
-----
-KAFKA_BOOTSTRAP_EXTERNAL=kafka-bootstrap-addr.my-domain.ext
-STOREPASS=storepass
-
-echo "Q" | openssl s_client -showcerts -servername $KAFKA_BOOTSTRAP_EXTERNAL -connect $KAFKA_HOST_PORT 2>/dev/null | awk ' /BEGIN CERTIFICATE/,/END CERTIFICATE/ { print $0 } ' > /tmp/my-cluster-kafka.pem
-----
 +
 The obtained `.pem` file is usually not one single certificate, but a certificate chain. You only have to provide the top-most issuer CA which is listed as last in the `/tmp/my-cluster-kafka.pem`.
 You can extract it manually or using the following command:
@@ -309,6 +298,10 @@ cp $(ls kafka-* | sort -r | head -n 1) my-cluster-kafka-ca.crt
 ----
 +
 NOTE: The proper way to obtain the trusted CA certificate is from a trusted source, and not by using the `openssl` command.
+      For this example we assume the client is running in a pod in the same namespace where the Kafka cluster was deployed.
+      If the client accessed the Kafka cluster from outside the Kubernetes cluster you would have to first determine the bootstrap address.
+      In that case you can also get the cluster certificate directly from the Kubernetes Secret, no need for `openssl`.
+      See: https://strimzi.io/docs/operators/latest/configuring.html#assembly-accessing-kafka-outside-cluster-str for more information.
 
 . Add the certificate for the Kafka broker to the truststore.
 +
--- a/documentation/modules/oauth/proc-oauth-authorization-keycloak-example.adoc
+++ b/documentation/modules/oauth/proc-oauth-authorization-keycloak-example.adoc
@@ -172,16 +172,27 @@ SSO_HOST=_SSO-HOSTNAME_
 SSO_HOST_PORT=$SSO_HOST:443
 STOREPASS=storepass
 
-echo "Q" | openssl s_client -showcerts -connect $SSO_HOST_PORT 2>/dev/null | awk ' /BEGIN CERTIFICATE/,/END CERTIFICATE/ { print $0 } ' > /tmp/sso.crt
+echo "Q" | openssl s_client -showcerts -connect $SSO_HOST_PORT 2>/dev/null | awk ' /BEGIN CERTIFICATE/,/END CERTIFICATE/ { print $0 } ' > /tmp/sso.pem
 ----
 +
 The certificate is required as Kubernetes ingress is used to make a secure (HTTPS) connection.
++
+Usually there is not one single certificate, but a certificate chain. You only have to provide the top-most issuer CA which is listed as last in the `/tmp/sso.pem`.
+You can extract it manually or using the following command:
++
+[source,shell,subs="+quotes"]
+----
+split -p "-----BEGIN CERTIFICATE-----" sso.pem sso-
+for f in $(ls sso-*); do mv $f $f.pem; done
+cp $(ls sso-* | sort -r | head -n 1) sso-ca.crt
+----
+
 
 . Deploy the certificate to Kubernetes as a secret.
 +
 [source,shell]
 ----
-kubectl create secret generic oauth-server-cert --from-file=/tmp/sso.crt -n $NS
+kubectl create secret generic oauth-server-cert --from-file=/tmp/sso-ca.crt -n $NS
 ----
 
 . Set the hostname as an environment variable
@@ -239,14 +250,24 @@ SSO_HOST=_SSO-HOSTNAME_
 SSO_HOST_PORT=$SSO_HOST:443
 STOREPASS=storepass
 
-echo "Q" | openssl s_client -showcerts -connect $SSO_HOST_PORT 2>/dev/null | awk ' /BEGIN CERTIFICATE/,/END CERTIFICATE/ { print $0 } ' > /tmp/sso.crt
+echo "Q" | openssl s_client -showcerts -connect $SSO_HOST_PORT 2>/dev/null | awk ' /BEGIN CERTIFICATE/,/END CERTIFICATE/ { print $0 } ' > /tmp/sso.pem
 ----
-
++
+Usually there is not one single certificate, but a certificate chain. You only have to provide the top-most issuer CA which is listed as last in the `/tmp/sso.pem`.
+You can extract it manually or using the following command:
++
+[source,shell,subs="+quotes"]
+----
+split -p "-----BEGIN CERTIFICATE-----" sso.pem sso-
+for f in $(ls sso-*); do mv $f $f.pem; done
+cp $(ls sso-* | sort -r | head -n 1) sso-ca.crt
+----
++
 . Create a truststore for TLS connection to the Kafka brokers.
 +
 [source,shell,subs="+quotes"]
 ----
-keytool -keystore /tmp/truststore.p12 -storetype pkcs12 -alias sso -storepass $STOREPASS -import -file /tmp/sso.crt -noprompt
+keytool -keystore /tmp/truststore.p12 -storetype pkcs12 -alias sso -storepass $STOREPASS -import -file /tmp/sso-ca.crt -noprompt
 ----
 
 . Use the Kafka bootstrap address as the hostname of the Kafka broker and the `tls` listener port (9093) to prepare a certificate for the Kafka broker.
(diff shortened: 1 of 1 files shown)


documentation/modules/oauth/proc-oauth-authorization-keycloak-example.adoc
```diff
@@ -256,14 +277,34 @@ keytool -keystore /tmp/truststore.p12 -storetype pkcs12 -alias sso -storepass $S
 KAFKA_HOST_PORT=my-cluster-kafka-bootstrap:9093
 STOREPASS=storepass
 
-echo "Q" | openssl s_client -showcerts -connect $KAFKA_HOST_PORT 2>/dev/null | awk ' /BEGIN CERTIFICATE/,/END CERTIFICATE/ { print $0 } ' > /tmp/my-cluster-kafka.crt
+echo "Q" | openssl s_client -showcerts -connect $KAFKA_HOST_PORT 2>/dev/null | awk ' /BEGIN CERTIFICATE/,/END CERTIFICATE/ { print $0 } ' > /tmp/my-cluster-kafka.pem
 ----
 
+If the client accesses the Kafka cluster over the Ingress or the OpenShift Route the Kafka bootstrap hostname will be different, and you may need to specify it as an extra argument to the `openssl` command in order to get the certificate containing the proper SNA records:
```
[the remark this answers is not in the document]
PaulRMellor: We need a couple of `+`  in the markdown to bring the content inline with the step:

image

documentation/modules/oauth/proc-oauth-authorization-keycloak-example.adoc
```diff
@@ -256,14 +277,34 @@ keytool -keystore /tmp/truststore.p12 -storetype pkcs12 -alias sso -storepass $S
...
+[source,shell]
+----
+KAFKA_BOOTSTRAP_EXTERNAL=kafka-bootstrap-addr.somedomain.ext
+STOREPASS=storepass
+
+echo "Q" | openssl s_client -showcerts -servername $KAFKA_BOOTSTRAP_EXTERNAL -connect $KAFKA_HOST_PORT 2>/dev/null | awk ' /BEGIN CERTIFICATE/,/END CERTIFICATE/ { print $0 } ' > /tmp/my-cluster-kafka.pem
+----
```
PaulRMellor, line 290:
```suggestion
+
[source,shell]
----
KAFKA_BOOTSTRAP_EXTERNAL=kafka-bootstrap-addr.my-domain.ext
STOREPASS=storepass

echo "Q" | openssl s_client -showcerts -servername $KAFKA_BOOTSTRAP_EXTERNAL -connect $KAFKA_HOST_PORT 2>/dev/null | awk ' /BEGIN CERTIFICATE/,/END CERTIFICATE/ { print $0 } ' > /tmp/my-cluster-kafka.pem
----
```

documentation/modules/oauth/proc-oauth-authorization-keycloak-example.adoc
```diff
@@ -256,14 +277,34 @@ keytool -keystore /tmp/truststore.p12 -storetype pkcs12 -alias sso -storepass $S
...
++
+The obtained `.pem` file is usually not one single certificate, but a certificate chain. You only have to provide the top-most issuer CA which is listed as last in the `/tmp/my-cluster-kafka.pem`.
+You can extract it manually or using the following command:
++
+[source,shell,subs="+quotes"]
+----
+split -p "-----BEGIN CERTIFICATE-----" /tmp/my-cluster-kafka.pem kafka-
+for f in $(ls kafka-*); do mv $f $f.pem; done
+cp $(ls kafka-* | sort -r | head -n 1) my-cluster-kafka-ca.crt
+----
```
PaulRMellor, line 300: Same as previous suggestion

documentation/modules/oauth/proc-oauth-authorization-keycloak-example.adoc
```diff
@@ -256,14 +277,34 @@ keytool -keystore /tmp/truststore.p12 -storetype pkcs12 -alias sso -storepass $S
...
+
+
 . Add the certificate for the Kafka broker to the truststore.
```
PaulRMellor, line 303: This shows as 1 in the [file view](https://github.com/strimzi/strimzi-kafka-operator/blob/9d3fc4739d80947244614dbe0e3418d47963d3a2/documentation/modules/oauth/proc-oauth-authorization-keycloak-example.adoc). The `+` fix above should sort it out.

PaulRMellor (approved):
Thanks @mstruk. Looks good. I guess a concern is that the examples don't work for everyone or that they become less useful over time (and we are not constantly checking they still work)

documentation/modules/oauth/proc-oauth-authorization-keycloak-example.adoc
scholzj, line 283: In such case, the port will be different as well ... and so it will be with node ports, but then you do not need the TLS-SNI of course.

documentation/modules/oauth/proc-oauth-authorization-keycloak-example.adoc
```diff
@@ -239,14 +250,24 @@ SSO_HOST=_SSO-HOSTNAME_
...
++
 . Create a truststore for TLS connection to the Kafka brokers.
 +
 [source,shell,subs="+quotes"]
 ----
-keytool -keystore /tmp/truststore.p12 -storetype pkcs12 -alias sso -storepass $STOREPASS -import -file /tmp/sso.crt -noprompt
+keytool -keystore /tmp/truststore.p12 -storetype pkcs12 -alias sso -storepass $STOREPASS -import -file /tmp/sso-ca.crt -noprompt
 ----
 
 . Use the Kafka bootstrap address as the hostname of the Kafka broker and the `tls` listener port (9093) to prepare a certificate for the Kafka broker.
```
scholzj, line 273: Unlike with the OAuth server, we actually know pretty well how the TLS is designed for the Kafka cluster since that is our work. The user should either use the Cluster CA public key or the CA used to sign the custom listener certificate if they use it. Such a description would be much more reliable than trying to dig the broker certificates through `openssl`.

mstruk: It is reliable to the extent that you can simply map `my-cluster-clients-ca-cert` secret by keys `ca.p12` and `ca.password` to the env vars in the yaml.

Here in the example we are inside the shell, as we use the Kafka CLI tools, and  in order to show how one would do it even from outside the cluster.
```
NS=sso
kubectl run -ti --restart=Never --image={DockerKafkaImageCurrent} kafka-cli -n $NS -- /bin/sh
```
I bet there are other places in the docs that clearly describe how to mount the secret for your pod and I prefer to simply link to those descriptions, and to emphasise in the documentation that using `openssl` to get the certificate is not a proper safe way. Rather one should get the trusted certificate from kubernetes cluster or from cluster administrator.

documentation/modules/oauth/proc-oauth-authorization-keycloak-example.adoc
mstruk: How should this be fixed to be correct?

documentation/modules/oauth/proc-oauth-authorization-keycloak-example.adoc
```diff
@@ -172,16 +172,28 @@ SSO_HOST=_SSO-HOSTNAME_
...
 STOREPASS=storepass
 
-echo "Q" | openssl s_client -showcerts -connect $SSO_HOST_PORT 2>/dev/null | awk ' /BEGIN CERTIFICATE/,/END CERTIFICATE/ { print $0 } ' > /tmp/sso.crt
+echo "Q" | openssl s_client -showcerts -connect $SSO_HOST_PORT 2>/dev/null | awk ' /BEGIN CERTIFICATE/,/END CERTIFICATE/ { print $0 } ' > /tmp/sso.pem
 ----
 +
 The certificate is required as Kubernetes ingress is used to make a secure (HTTPS) connection.
++
+Usually there is not one single certificate, but a certificate chain. You only have to provide the top-most issuer CA which is listed as last in the `/tmp/sso.pem`.
+You can extract it manually or using the following commands:
```
mstruk, line 181: Here I removed the OpenSSL reference because it's incorrect.

mstruk:
I added the example subtitles as you suggested:
https://github.com/strimzi/strimzi-kafka-operator/blob/9acca473feaf71143f84052e6b97dc34b0cde00b/documentation/modules/oauth/proc-oauth-authorization-keycloak-example.adoc

documentation/modules/oauth/proc-oauth-authorization-keycloak-example.adoc
scholzj: But what you have here basically works in like 20% of cases only. Everyone else will be just confused as well about why the super detailed guide does not work.

documentation/modules/oauth/proc-oauth-authorization-keycloak-example.adoc
```diff
@@ -256,14 +283,38 @@ keytool -keystore /tmp/truststore.p12 -storetype pkcs12 -alias sso -storepass $S
...
++
+If the client accesses the Kafka cluster over the ingress or OpenShift route, the Kafka bootstrap hostname is different, and you may need to specify it as an extra argument to the `openssl` command in order to get the certificate containing the proper SNA records:
++
+[source,shell]
+----
+KAFKA_BOOTSTRAP_EXTERNAL=kafka-bootstrap-addr.my-domain.ext
+STOREPASS=storepass
+
+echo "Q" | openssl s_client -showcerts -servername $KAFKA_BOOTSTRAP_EXTERNAL -connect $KAFKA_HOST_PORT 2>/dev/null | awk ' /BEGIN CERTIFICATE/,/END CERTIFICATE/ { print $0 } ' > /tmp/my-cluster-kafka.pem
+----
```
scholzj, line 298: This does not make sense because above you assume _`tls` listener port (9093)_. If you wanna make it general, you should do it properly, and cover all the listener types and all the possible situations from the beginning. So you need to start from the beginning, how do they find the bootstrap address in the status section of the Kafka YAML. (i.e. how to set `KAFKA_HOST_PORT` properly in the first place).

documentation/modules/oauth/proc-oauth-authorization-keycloak-example.adoc
scholzj: You need to explain how to find the proper bootstrap address.

documentation/modules/oauth/proc-oauth-authorization-keycloak-example.adoc
mstruk: What makes you think it only works for 20% of the cases?

SSO is pretty simple - it's TLS protected web server. All we need is a local truststore with the CA.
How does the above procedure not work for someone?

documentation/modules/oauth/proc-oauth-authorization-keycloak-example.adoc
mstruk: I can add a link to: https://strimzi.io/docs/operators/latest/configuring.html#proc-accessing-kafka-using-nodeports-str

documentation/modules/oauth/proc-oauth-authorization-keycloak-example.adoc
mstruk: Let's differentiate between example related docs, and non-example related docs. This example runs in an interactive pod in the same namespace where the Kafka cluster is deployed by Strimzi. Access for the example is not via a Route.
This extra fragment tries to be helpful for people who want to access it from outside. It's not about being consistent, it's not about covering all the possible listener types ... Contecting to Kafka cluster in Strimzi through different listener types should be covered elsewhere in the documentation. I suggest I include here the link to such a chapter.

scholzj: > This example runs in an interactive pod in the same namespace where the Kafka cluster is deployed by Strimzi. Access for the example is not via a Route.

Fine, then remove the stuff which does not belong there. Because the other stuff suggests it is a general guide that works for all configurations.

> This extra fragment tries to be helpful for people who want to access it from outside. It's not about being consistent, it's not about covering all the possible listener types ... Connecting to Kafka cluster in Strimzi through different listener types should be covered elsewhere in the documentation. I suggest I include here the link to such a chapter.

Fine, then do it properly and cover the access from outside in the previous steps and for all listeners.

----

Pick one and do it properly, when you do both, it ultimately doesn't work for anyone.

documentation/modules/oauth/proc-oauth-authorization-keycloak-example.adoc
scholzj: I do not care about the SSO part. I talk about the Kafka part in all my comments

documentation/modules/oauth/proc-oauth-authorization-keycloak-example.adoc
```diff
@@ -256,14 +283,31 @@ keytool -keystore /tmp/truststore.p12 -storetype pkcs12 -alias sso -storepass $S
...
+split -p "-----BEGIN CERTIFICATE-----" /tmp/my-cluster-kafka.pem kafka-
+for f in $(ls kafka-*); do mv $f $f.pem; done
+cp $(ls kafka-* | sort -r | head -n 1) my-cluster-kafka-ca.crt
+----
++
+NOTE: The proper way to obtain the trusted CA certificate is from a trusted source, and not by using the `openssl` command.
+      For this example we assume the client is running in a pod in the same namespace where the Kafka cluster was deployed.
+      If the client accessed the Kafka cluster from outside the Kubernetes cluster you would have to first determine the bootstrap address.
+      In that case you can also get the cluster certificate directly from the Kubernetes Secret, no need for `openssl`.
+      See: https://strimzi.io/docs/operators/latest/configuring.html#assembly-accessing-kafka-outside-cluster-str for more information.
```
mstruk, line 304: I removed the external access certificate fetching and added additional comments and a link into NOTE section.
The link itself probably has to be done differently.

scholzj (approved):

PR #7930, opened 2023-01-18
[systemtest] Update scalability tests for UO and update tags

im-konge:
### Type of change

- Enhancement

### Description

This PR adds new `UserScalabilityIsolatedST`, which contains tests for creating and updating 1000  users with TLS and SCRAM-SHA with ACLs.

Also, I'm changing other tests containing `scalability` tag (which is more about scaling up and down) to `componentscaling`.

### Checklist

- [x] Write tests


systemtest/src/main/java/io/strimzi/systemtest/enums/UserAuthType.java
```diff
@@ -0,0 +1,9 @@
+/*
+ * Copyright Strimzi authors.
+ * License: Apache License 2.0 (see the file LICENSE or http://apache.org/licenses/LICENSE-2.0.html).
+ */
+package io.strimzi.systemtest.enums;
+public enum UserAuthType {
```
Frawless, line 6: Don't we have it in API module?

Frawless (approved):

systemtest/src/main/java/io/strimzi/systemtest/enums/UserAuthType.java
im-konge: I'm afraid that there is nothing which we can use - I mean, nothing in some `common` module etc. At least I don't know about some enum or class.

systemtest/src/test/java/io/strimzi/systemtest/operators/user/UserScalabilityIsolatedST.java
```diff
@@ -0,0 +1,148 @@
...
+
+import java.util.ArrayList;
+import java.util.List;
+
+import static io.strimzi.systemtest.Constants.SCALABILITY;
+
+@Tag(SCALABILITY)
+public class UserScalabilityIsolatedST extends AbstractST {
+    private static final Logger LOGGER = LogManager.getLogger(UserScalabilityIsolatedST.class);
+    private static String clusterName = "";
```
see-quick, line 36: Why do you explicitly init these variables (i.e., cluster and topic name) with empty string? By default, these string vars will have null is there any hidden logic  I am overlooking? :)

see-quick (approved):
LGTM, just one question 👍...

systemtest/src/test/java/io/strimzi/systemtest/operators/user/UserScalabilityIsolatedST.java
im-konge: No big logic behind it, I just didn't want to have it `null`. I will change it 😄  thanks

im-konge:
@strimzi-ci run tests test_only profile=scalability

scholzj (approved):
Some nits. LGTM otherwise.

systemtest/src/test/java/io/strimzi/systemtest/connect/ConnectIsolatedST.java
```diff
@@ -399,7 +399,7 @@ void testJvmAndResources(ExtensionContext extensionContext) {
     }
 
     @ParallelNamespaceTest
-    @Tag(SCALABILITY)
+    @Tag(COMPONENT_SCALING)
```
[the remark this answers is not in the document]
im-konge: It is still under regression, only change is when we want to run just those scale up/down tests.

im-konge:
@strimzi-ci run tests test_only profile=scalability

im-konge:
The tests need to be executed on bigger cluster, they are quite unstable on Minikube, so here are the results for the tests running on OCP cluster:

image

PR #7934, opened 2023-01-19
Move creation of workload resources from the `AbstractModel` class

scholzj:
### Type of change

- Refactoring

### Description

This PR refactors the creation of workload resources from the `AbstractModel` class to a separate `WorkloadUtils` class. This includes StrimziPodSets, StatefulSets, Deployments, and standalone Pods. This is done in one step as they use in many cases the same fields and moving only part of them does not allow to identify and remove the unneeded things from `AbstractModel`.

The methods look relatively big with a lot of options. There might be things we can further improve in the future. But right now I think it is most important that to clean the `AbstractModel` class and entangle which field or method belongs to what. that should help us with any possible future imrpovements.

_Note: This is a drat PR which currently does not contain any new unit tests._

### Checklist

- [ ] Write tests

### DIFF
--- a/api/src/main/java/io/strimzi/api/kafka/model/template/CruiseControlTemplate.java
+++ b/api/src/main/java/io/strimzi/api/kafka/model/template/CruiseControlTemplate.java
@@ -32,7 +32,7 @@ import java.util.Map;
 public class CruiseControlTemplate implements Serializable, UnknownPropertyPreserving {
     private static final long serialVersionUID = 1L;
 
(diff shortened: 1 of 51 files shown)


scholzj:
@tombentley @ppatierno Sorry for the review noise ... I moved wrong Pr out of review. The one to look at is #7959

scholzj:
This should be now ready for review.

ppatierno (approved):
LGTM.

PR #7949, opened 2023-01-20
Add a unit test case for validateCreateBootstrapService method.

see-quick:
### Type of change

- Enhancement / new feature

### Description

This PR adds a unit test case for the `validateCreateBootstrapService` method. 

### Checklist

- [x] Write tests

### DIFF
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/model/ListenersValidator.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/model/ListenersValidator.java
@@ -208,7 +208,7 @@ public class ListenersValidator {
     private static void validateCreateBootstrapService(Set<String> errors, GenericKafkaListener listener) {
         if (KafkaListenerType.LOADBALANCER != listener.getType()
                 && listener.getConfiguration() != null
-                && !listener.getConfiguration().getCreateBootstrapService()) {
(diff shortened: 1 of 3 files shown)


scholzj (commented):
I think you misread the code. This option is _true_ by default. The bootstrap service is always created. On `type: loadbalancer` listener, you can disable it (because LBs cost money so you have more motivation to do so ;-)).

So what the validator does is it checks that id the listener is not LB type, the `createBootstrapService` has to be true. Which I think works in the original code and you break it.

see-quick:
> I think you misread the code. This option is _true_ by default. The bootstrap service is always created. On `type: loadbalancer` listener, you can disable it (because LBs cost money so you have more motivation to do so ;-)).
> 
> So what the validator does is it checks that id the listener is not LB type, the `createBootstrapService` has to be true. Which I think works in the original code and you break it.

I see my bad...should I at least add a unit for validation of this part `errors.add("listener " + listener.getName() + " cannot configure createBootstrapService because it is not load balancer listener");`, because I don't see that is covered.

scholzj (commented):
Adding test is fine. But you should clean the unrelated changes.

see-quick:
> Adding test is fine. But you should clean the unrelated changes.

Yes, and I have also changed the name of the PR.

scholzj (approved):

PR #7959, opened 2023-01-23
Move from OkHttp client to JDK HTTP client (for use in Fabric8 Kubernetes client)

scholzj:
### Type of change

- Refactoring

### Description

This PR moves away from using the OkHttp client which is used in an old version (3.x) which is not updated anymore and moves to use the JDK HTTP Client. It also fixes some related tests which were failing with the change.

### Checklist

- [x] Update CHANGELOG.md

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## 0.34.0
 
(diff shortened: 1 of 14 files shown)


ppatierno (approved):
LGTM

tombentley (approved):

PR #7975, opened 2023-01-25
Redesigned the `UseOperatorConfig` class to make it more flexible and clean 

ShubhamRwt:

### Type of change

- Enhancement

### Description

This PR redesign the `UserOperatorConfig` class. This PR gets rid of a lot of parameter declaration which was happening earlier and this new design make the  `UserOperatorConfig` cleaner. We have introduced a new inner class named `ConfigParameter` which wraps all the parameter corresponding to a `config` parameter for eg. its type, default value, key.
For different types we have created an abstract class named `Type` which takes care of the types of the config parameter

### Checklist

- [ ] Write tests
- [ ] Update documentation
- [ ] Update CHANGELOG.md


user-operator/src/main/java/io/strimzi/operator/user/UserOperatorConfig.java
```diff
@@ -364,194 +296,64 @@ private static boolean getBooleanProperty(Map<String, String> map, String name,
...
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) return true;
+        if (o == null || getClass() != o.getClass()) return false;
+        UserOperatorConfig config = (UserOperatorConfig) o;
+        return Objects.equals(map, config.map);
     }
 
     @Override
-    public String toString() {
```
scholzj, line 529: I guess we need some proper `toString` implementation for the logs.

scholzj (commented):
I had a quick look through this. My thoughts ...

* I like how the options are defined

... but ...

* I do not like how much harder to read is the resulting code (e.g. doing `config.get(UserOperatorConfig.KRAFT_ENABLED))` instead of just `config.isKraftEnabled()`
* It does not seem to address the issues we have in tests where changing different values for different test cases is very complicated and ugly.

I guess for the first point, we could keep the getters. For the second, some kind of builder to _rebuild_ the final config with just changing one or two options would be needed I guess.

I wonder what others think.

user-operator/src/main/java/io/strimzi/operator/user/UserOperatorConfig.java
```diff
@@ -47,313 +141,151 @@ public class UserOperatorConfig {
...
-     * @param controllerThreadPoolSize Size of the pool of the controller threads used to reconcile the users
-     * @param cacheRefresh Refresh interval for the cache storing the resources from the Kafka Admin API
-     * @param batchQueueSize Maximal queue for requests when micro-batching the Kafka Admin API requests
-     * @param batchMaxBlockSize Maximal batch size for micro-batching the Kafka Admin API requests
-     * @param batchMaxBlockTime Maximal batch time for micro-batching the Kafka Admin API requests
-     * @param userOperationsThreadPoolSize Size of the thread pool for user operations done by KafkaUserOperator and
-     *                                     the classes used by it
+     * @param map  Map containing configurations and their respective values
      */
     @SuppressWarnings({"checkstyle:ParameterNumber"})
```
tombentley, line 235: I don't think this is still needed

user-operator/src/main/java/io/strimzi/operator/user/UserOperatorConfig.java
```diff
@@ -47,313 +141,151 @@ public class UserOperatorConfig {
...
 
-        String clusterCaCertSecretName = map.get(UserOperatorConfig.STRIMZI_CLUSTER_CA_CERT_SECRET_NAME);
-
-        String euoKeySecretName = map.get(UserOperatorConfig.STRIMZI_EO_KEY_SECRET_NAME);
-
-        String caNamespace = map.get(UserOperatorConfig.STRIMZI_CA_NAMESPACE);
-        if (caNamespace == null || caNamespace.isEmpty()) {
-            caNamespace = namespace;
+        if (this.map.get(STRIMZI_CA_NAMESPACE) == null || this.map.get(STRIMZI_CA_NAMESPACE).equals("")) {
+            this.map.put(STRIMZI_CA_NAMESPACE, this.map.get(STRIMZI_NAMESPACE));
```
tombentley, line 254: Why are we special casing these when `ConfigParameter` already knows about required-ness?

user-operator/src/main/java/io/strimzi/operator/user/UserOperatorConfig.java
```diff
@@ -364,194 +296,64 @@ private static boolean getBooleanProperty(Map<String, String> map, String name,
...
 
-    /**
-     * @return  Maximal queue for requests when micro-batching the Kafka Admin API requests
-     */
-    public int getBatchQueueSize() {
-        return batchQueueSize;
+        final String s = map.getOrDefault(value.key, value.defaultValue);
+        if (s != null) {
+            if ((value.key.equals(STRIMZI_NAMESPACE) || value.key.equals(STRIMZI_CA_CERT_SECRET_NAME) || value.key.equals(STRIMZI_CA_KEY_SECRET_NAME)) && s.equals("")) {
+                throw new InvalidConfigurationException("Config value: " + value.key + " is mandatory");
```
tombentley, line 320: Why do we need to special case these when the `ConfigParameter` already knows about required-ness?

user-operator/src/main/java/io/strimzi/operator/user/UserOperatorConfig.java
```diff
@@ -14,11 +13,106 @@
...
+import java.util.Collections;
 
 /**
  * Cluster Operator configuration
  */
 public class UserOperatorConfig {
+
+
+    private static abstract class Type<T> {
+        abstract T parse(String s);
```
tombentley, line 28: As an abstraction this probably needs to define:

1. what exception is thrown if `s` cannot be parsed
2. whether the result can ever be null

user-operator/src/main/java/io/strimzi/operator/user/UserOperatorConfig.java
```diff
@@ -14,11 +13,106 @@
...
+     * A kubernetes selector.
+     */
+    private static final Type<? extends Labels> LABEL_PREDICATE = new Type<>() {
+        @Override
+        public Labels parse(String s) {
+            return Labels.fromString(s);
+        }
+    };
+
+    /** Wrapper class for all the configuration parameters */
```
tombentley, line 81: Document the _concept_, or what this class is trying to model. "Wrapper class" is not really a concept. 

e.g.

```
Models a configuration parameter, identified by a unique key, which may be required, and if not may have a default value. Optional parameters without a default value implicitly have a null default. The key is also the name of the environment variable from which the value may be read, when it is read from the environment.
```

That explains to someone coming along later what the intent for this class is.

user-operator/src/main/java/io/strimzi/operator/user/UserOperatorConfig.java
```diff
@@ -47,313 +141,151 @@ public class UserOperatorConfig {
...
+    /** Maximal batch time for micro-batching the Kafka Admin API requests */
+    public static final ConfigParameter<Integer> BATCH_MAXIMUM_BLOCK_TIME_MS = new ConfigParameter<>(STRIMZI_BATCH_MAXIMUM_BLOCK_TIME_MS, INTEGER, "100");
+    /** Size of the thread pool for user operations done by KafkaUserOperator and the classes used by it */
+    public static final ConfigParameter<Integer> USER_OPERATIONS_THREAD_POOL_SIZE = new ConfigParameter<>(STRIMZI_USER_OPERATIONS_THREAD_POOL_SIZE, INTEGER, "4");
+    /** Additional configuration for the Kafka Admin Client */
+    public static final ConfigParameter<String> KAFKA_ADMIN_CLIENT_CONFIGURATION = new ConfigParameter<>(STRIMZI_KAFKA_ADMIN_CLIENT_CONFIGURATION, STRING, false);
+    /** Lit of maintenance windows */
+    public static final ConfigParameter<String> MAINTENANCE_TIME_WINDOWS = new ConfigParameter<>(STRIMZI_MAINTENANCE_TIME_WINDOWS, STRING, false);
+
+    private static final Map<String, ConfigParameter<?>> CONFIG_VALUES = Map.ofEntries(
```
tombentley, line 200: This feels wrong. Having two places to update when adding a parameter is a bug waiting to happen. It should be possible, some how, to build the collection as a side-effect of defining a parameter.

user-operator/src/main/java/io/strimzi/operator/user/UserOperatorConfig.java
```diff
@@ -47,313 +141,151 @@ public class UserOperatorConfig {
...
-        if (caCertSecretName == null || caCertSecretName.isEmpty()) {
-            throw new InvalidConfigurationException(UserOperatorConfig.STRIMZI_CA_CERT_SECRET_NAME + " cannot be null");
+    public UserOperatorConfig(Map<String, String> map) {
+        this.map = new HashMap<>(map.size());
+        for (Map.Entry<String, String> entry : map.entrySet()) {
+            final ConfigParameter<?> configValue = CONFIG_VALUES.get(entry.getKey());
+            if (configValue == null) {
+                throw new IllegalArgumentException("Unknown config key " + entry.getKey());
+            }
+            this.map.put(configValue.key, get(map, configValue));
```
tombentley, line 243: Use getters for accessing the members of  `configValue`.

user-operator/src/main/java/io/strimzi/operator/user/UserOperatorConfig.java
```diff
@@ -14,11 +13,106 @@
...
+     */
+    private static final Type<? extends Labels> LABEL_PREDICATE = new Type<>() {
+        @Override
+        public Labels parse(String s) {
+            return Labels.fromString(s);
+        }
+    };
+
+    /** Wrapper class for all the configuration parameters */
+    public static class ConfigParameter<T> {
```
tombentley, line 82: This can probably be a `record`.

user-operator/src/main/java/io/strimzi/operator/user/UserOperatorConfig.java
ShubhamRwt: This is just to make sure that if `CA_NAMESPACE is null or empty then we should use the `NAMESPACE` value. It is more like a validation that `CA_NAMESPACE` cannot be null.

user-operator/src/main/java/io/strimzi/operator/user/UserOperatorConfig.java
ShubhamRwt: The required-ness is fine but  if the the user gives an empty string "" like this then it will be treated as a correct namespace if we don't include this check

user-operator/src/main/java/io/strimzi/operator/user/UserOperatorConfig.java
ShubhamRwt: Yet to be fixed

user-operator/src/main/java/io/strimzi/operator/user/UserOperatorConfigBuilder.java
```diff
@@ -0,0 +1,202 @@
...
+ */
+public class UserOperatorConfigBuilder {
+
+    private final Map<String, String> map;
+
+    protected  UserOperatorConfigBuilder() {
+        this.map = new HashMap<>();
+    }
+
+    protected  UserOperatorConfigBuilder(UserOperatorConfig config) {
```
scholzj, line 22: The builder looks good. But I wonder if you can maybe more automate it ... 
* You know the config keys, so can you maybe just iterate through the options instead of listing them one by one?
* As this is expected to be used by tests only basically, do we need the specific `.with...()` methods for each option? Most of them might never be used. Maybe we can just have some `.with(ConfigParameter, Value)` method?
* Should this be moved into the `USerOperatorConfig` class to have private access to some of the fields without needing to expose them through public methods?

scholzj (commented):
Looks msotly good. Just left some smaller comments.

user-operator/src/main/java/io/strimzi/operator/user/UserOperatorConfig.java
```diff
@@ -4,21 +4,59 @@
...
+         * Contructor
+         * @param key Configuration parameter name/key
+         * @param type type of the default value
+         * @param defaultValue default value of the configuration parameter
+         * @param required  If the value is required or not
+         */
+        public ConfigParameter {
+            CONFIG_VALUES.put(key, this);
+        }
+    }
```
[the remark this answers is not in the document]
ShubhamRwt: Good catch jakub. I think I can use something like this in CO also

user-operator/src/main/java/io/strimzi/operator/user/UserOperatorConfig.java
```diff
@@ -47,229 +85,151 @@ public class UserOperatorConfig {
     static final String STRIMZI_USER_OPERATIONS_THREAD_POOL_SIZE = "STRIMZI_USER_OPERATIONS_THREAD_POOL_SIZE";
 
     // Default values
-    static final long DEFAULT_FULL_RECONCILIATION_INTERVAL_MS = 120_000;
-    static final String DEFAULT_KAFKA_BOOTSTRAP_SERVERS = "localhost:9091";
+
+    /**
+     * Namespace in which the operator will run and create resources
+     */
+    public static final ConfigParameter<String> NAMESPACE = new ConfigParameter<>(STRIMZI_NAMESPACE, STRING, "", true);
```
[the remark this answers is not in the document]
ShubhamRwt: It is used in some tests

user-operator/src/main/java/io/strimzi/operator/user/UserOperatorConfig.java
```diff
@@ -278,280 +238,292 @@ public static UserOperatorConfig fromMap(Map<String, String> map) {
...
+     * @param value The configuration value that need to be checked
+     * @return The configuration value
      */
-    private static int getIntProperty(Map<String, String> map, String name, int defaultVal) {
-        String value = map.get(name);
-        if (value != null) {
-            return Integer.parseInt(value);
-        } else {
-            return defaultVal;
+    private <T> T get(Map<String, String> map, ConfigParameter<T> value) {
```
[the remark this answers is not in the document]
ShubhamRwt: This `get()` method is used to make sure that we don't parse null values and if any required value is `null` then it throw the exception so I think that this method is required

user-operator/src/main/java/io/strimzi/operator/user/UserOperatorConfig.java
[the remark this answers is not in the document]
scholzj: Ok, in that case we have to keep it I guess.

user-operator/src/main/java/io/strimzi/operator/user/UserOperatorConfig.java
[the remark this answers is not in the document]
ShubhamRwt: @scholzj I tried to move this method to `operator-common` but there is one problem. Iam currently using the contructor to build the Configuration Map as a side effect of initialising the parameters. Now If I move it to `operator-common` it would be hard to do it the way I am doing it currently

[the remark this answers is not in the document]
scholzj: That is fine. If it cannot be done it cannot be done.

operator-common/src/main/java/io/strimzi/operator/common/operator/resource/AbstractConfig.java
```diff
@@ -0,0 +1,74 @@
...
+     */
+    public static final AbstractConfig<? extends Boolean> BOOLEAN = new AbstractConfig<>() {
+        @Override
+        public Boolean parse(String s) {
+            return Boolean.parseBoolean(s);
+        }
+    };
+
+    /**
+     * A kubernetes selector.
```
scholzj, line 65: Kubernetes Selectors have many different forms. Might be good to add here more description and some examples of the expected format.

user-operator/src/main/java/io/strimzi/operator/user/Main.java
```diff
@@ -70,7 +73,9 @@ public static void main(String[] args) {
         Security.setProperty("networkaddress.cache.ttl", dnsCacheTtl);
 
         // Create and log UserOperatorConfig
-        UserOperatorConfig config = UserOperatorConfig.fromMap(System.getenv());
+        Map<String, String> m = new HashMap<>(System.getenv());
+        m.keySet().retainAll(UserOperatorConfig.keyNames());
```
scholzj, line 77: If this is needed, why does it not live in the constructor or in the `buildFromMap` method?

scholzj:
@ShubhamRwt Can you please fix the CI and theDCO signoff before we dive into the next review round?

ShubhamRwt:
@scholzj Hi, Actually it requires more changes so I have changed it to draft

operator-common/src/main/java/io/strimzi/operator/common/operator/resource/AbstractConfig.java
```diff
@@ -0,0 +1,97 @@
...
+            }
+        }
+
+        return kafkaAdminClientConfiguration;
+    };
+
+    /**
+     * A java Long
+     */
+    AbstractConfig<Long> LONG = (AbstractConfig<Long>) Long::parseLong;
```
tombentley: `parseLong` will throw `NumberFormatException`, but the contract says `InvalidConfigurationException` will be used, so you need to catch and rethrow.

operator-common/src/main/java/io/strimzi/operator/common/operator/resource/AbstractConfig.java
```diff
@@ -0,0 +1,97 @@
...
+
+    /**
+     * A Java Integer
+     */
+    AbstractConfig<Integer> INTEGER = (AbstractConfig<Integer>) Integer::parseInt;
+
+    /**
+     * A Java Boolean
+     */
+    AbstractConfig<Boolean> BOOLEAN = (AbstractConfig<Boolean>) Boolean::parseBoolean;
```
tombentley: Again, exception types.

operator-common/src/main/java/io/strimzi/operator/common/operator/resource/AbstractConfig.java
```diff
@@ -0,0 +1,97 @@
...
+
+    /**
+     * A kubernetes selector.
+     */
+    AbstractConfig<Labels> LABEL_PREDICATE = (AbstractConfig<Labels>) Labels::fromString;
```
tombentley: Again, exception types.

user-operator/src/main/java/io/strimzi/operator/user/operator/KafkaUserOperator.java
```diff
@@ -319,6 +319,8 @@ private void maybeGenerateScramCredentials(Reconciliation reconciliation, KafkaU
      * @param userSecret        Secret with existing user credentials or null if the secret doesn't exist yet
      */
     private void maybeGenerateTlsCredentials(Reconciliation reconciliation, KafkaUserModel user, Secret userSecret) {
+        System.out.println(config.getCaNamespace());
```
tombentley, line 322: We should remove this. Use logging if you really want to output something.

operator-common/src/main/java/io/strimzi/operator/common/operator/resource/ConfigParameter.java
```diff
@@ -0,0 +1,73 @@
...
+        map.put(key, this);
+    }
+
+    /**
+     * Generates the configuration map
+     * @param envVarMap    Map containing values entered by user.
+     * @param configValues Map containing all the configuration keys with default values
+     * @return             Generated configuration map
+     */
+    public static Map<String, Object> define(Map<String, String> envVarMap, Map<String, ConfigParameter<?>> configValues) {
```
tombentley, line 42: `configValues` should probably be `configParameters`, since it doesn't contain any actual _values_, but rather the schema for a set of expected values.

operator-common/src/main/java/io/strimzi/operator/common/operator/resource/ConfigParameter.java
```diff
@@ -0,0 +1,73 @@
...
+        // now add all those config (with default value) that weren't in the given map
+        Map<String, ConfigParameter<?>> x = new HashMap<>(configValues);
+        x.keySet().removeAll(envVarMap.keySet());
+        for (ConfigParameter<?> value : x.values()) {
+            generatedMap.put(value.key(), get(envVarMap, configValues, value));
+        }
+        return generatedMap;
+    }
+    private static <T> T get(Map<String, String> map, Map<String, ConfigParameter<?>> configValues, ConfigParameter<T> value) {
+        if (!configValues.containsKey(value.key())) {
```
tombentley, line 59: If `value` is null  (which will happen if an unexpected env var is present in the `envVarMap`) this will NPE. It would be better to make this check at the point where you do the lookup (in `define()`).

operator-common/src/main/java/io/strimzi/operator/common/operator/resource/ConfigParameter.java
```diff
@@ -0,0 +1,73 @@
...
+            throw new InvalidConfigurationException("Unknown config value: " + value.key() + " probably needs to be added to Config.configValues");
```
tombentley, line 60: It's always worth thinking about who (as in, which persona) you expect to see an exception message, and what will make sense to them. In this case do they know what `Config.configValues` is? I suspect this is a reference to an out of date class. You could correct that, but in general it's a good idea to avoid putting class, field or method names in messages. If you _have to_ use a class name use a class literal so can't get out of sync with the actual code, e.g. `"Blah blah " + Foo.class.getSimpleName() + " blah blah"`.

operator-common/src/main/java/io/strimzi/operator/common/operator/resource/ConfigParameter.java
```diff
@@ -0,0 +1,73 @@
...
+        }
+
+        final String s = map.getOrDefault(value.key(), value.defaultValue());
+        if (s != null) {
+            return value.type().parse(s);
```
tombentley, line 65: Here's where you might catch InvalidConfigException and add info about what env var had the problem.

ShubhamRwt:
Thanks for the review @tombentley. Let me add these changes

operator-common/src/main/java/io/strimzi/operator/common/operator/resource/AbstractConfig.java
```diff
@@ -0,0 +1,97 @@
...
+
+/**
+ * Abstract class which contains different type of parameters
+ */
+public interface AbstractConfig<T> {
+
+    /**
+     * Parses the string based on its type
+     *
+     * @param s String value
```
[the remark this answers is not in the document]
ShubhamRwt: Yes `s` can be null

operator-common/src/main/java/io/strimzi/operator/common/operator/resource/AbstractConfig.java
```diff
@@ -0,0 +1,97 @@
...
+     * @return the value based on its type
```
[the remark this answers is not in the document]
ShubhamRwt: Yes it can

ShubhamRwt:
@scholzj Is it feasible for 0.34 release?

scholzj:
If it works and is approved then I guess it should.

operator-common/src/main/java/io/strimzi/operator/common/operator/resource/ConfigParameter.java
```diff
@@ -0,0 +1,74 @@
...
+/**
+ * Models a configuration parameter, identified by a unique key, which may be required, and if not may have a default value.
+ * Optional parameters without a default value implicitly have a null default.
+ * The key is also the name of the environment variable from which the value may be read, when it is read from the environment.
+ *
+ * @param key Configuration parameter name/key
+ * @param <T> Type of object
+ * @param type type of the default value
+ * @param defaultValue default value of the configuration parameter
+ * @param required  If the value is required or not
```
scholzj, line 21: The `map` field seems to be missing here?

Tip: if you align the descriptions with tabs, the JavaDocs will be much more readable.

operator-common/src/main/java/io/strimzi/operator/common/operator/resource/ConfigParameter.java
```diff
@@ -0,0 +1,74 @@
...
+
+        // now add all those config (with default value) that weren't in the given map
+        Map<String, ConfigParameter<?>> x = new HashMap<>(configParameterMap);
+        x.keySet().removeAll(envVarMap.keySet());
+        for (ConfigParameter<?> value : x.values()) {
+            generatedMap.put(value.key(), get(envVarMap, value));
+        }
+        return generatedMap;
+    }
+    private static <T> T get(Map<String, String> map, ConfigParameter<T> value) {
```
scholzj, line 62: Empty line between the method would be good.

operator-common/src/main/java/io/strimzi/operator/common/operator/resource/ConfigParameterParser.java
```diff
@@ -0,0 +1,121 @@
...
+     * @param configValue config value in String format
+     * @throws InvalidConfigurationException if the given configuration value is not supported
+     * @return the value based on its type
+     */
+    T parse(String configValue) throws InvalidConfigurationException;
+
+    /**
+     * A java string
+     */
+
```
scholzj, line 34: Empty line not needed?

operator-common/src/main/java/io/strimzi/operator/common/operator/resource/ConfigParameterParser.java
```diff
@@ -0,0 +1,121 @@
...
+    ConfigParameterParser<String> NON_EMPTY_STRING = configValue -> {
+        if (configValue == null || configValue.isEmpty()) {
+            throw new InvalidConfigurationException("Failed to parse. Value cannot be empty or null");
+        } else {
+            return configValue;
+        }
+    };
+
+    /**
+     * A semicolon-delimited list of strings.
```
scholzj, line 49: What do we use semicolon-delimited lists for? I know we use comma-delimited lists, but not aware of semicolons.

user-operator/src/main/java/io/strimzi/operator/user/Main.java
```diff
@@ -40,6 +40,7 @@
 /**
  * The main class of the Strimzi User Operator
  */
+@SuppressWarnings({"checkstyle:ClassDataAbstractionCoupling"})
```
scholzj, line 43: Is this really needed just for changing `fromMap` to `buildFromMap`?

user-operator/src/main/java/io/strimzi/operator/user/UserOperatorConfig.java
```diff
@@ -46,512 +58,381 @@ public class UserOperatorConfig {
     static final String STRIMZI_BATCH_MAXIMUM_BLOCK_TIME_MS = "STRIMZI_BATCH_MAXIMUM_BLOCK_TIME_MS";
     static final String STRIMZI_USER_OPERATIONS_THREAD_POOL_SIZE = "STRIMZI_USER_OPERATIONS_THREAD_POOL_SIZE";
 
-    // Default values
-    static final long DEFAULT_FULL_RECONCILIATION_INTERVAL_MS = 120_000;
-    static final String DEFAULT_KAFKA_BOOTSTRAP_SERVERS = "localhost:9091";
+    /**
+     * Namespace in which the operator will run and create resources
+     */
+    public static final ConfigParameter<String> NAMESPACE = new ConfigParameter<>(STRIMZI_NAMESPACE, NON_EMPTY_STRING, "", true, CONFIG_VALUES);
```
scholzj, line 64: Are the names of the env vars used also somewhere else? If not, wouldn't the resulting code be more readable if the strings were just inlined and the constants removed? E.g.

```suggestion
    public static final ConfigParameter<String> NAMESPACE = new ConfigParameter<>("STRIMZI_NAMESPACE", NON_EMPTY_STRING, "", true, CONFIG_VALUES);
```

user-operator/src/main/java/io/strimzi/operator/user/UserOperatorConfig.java
```diff
@@ -46,512 +58,381 @@ public class UserOperatorConfig {
...
+    public static Set<String> keyNames() {
+        return Collections.unmodifiableSet(CONFIG_VALUES.keySet());
     }
 
     /**
-     * @return  Clients CA validity in days
+     * Gets the configuration value corresponding to the key
+     * @param <T> type of value
+     * @param value instance of Value class
+     * @return configuration value w.r.t to the key
```
scholzj, line 202: Again, formatting the JavaDocs nicely as it was in the old code can do magic in how the Javadoc is readable. So does better capitalization.

```suggestion
     * Gets the configuration value corresponding to the key
     * 
     * @param <T>     Type of value
     * @param value   Instance of Value class
     * 
     * @return   Configuration value w.r.t to the key
```

user-operator/src/main/java/io/strimzi/operator/user/operator/KafkaUserOperator.java
```diff
@@ -320,6 +320,7 @@ private void maybeGenerateScramCredentials(Reconciliation reconciliation, KafkaU
      * @param userSecret        Secret with existing user credentials or null if the secret doesn't exist yet
      */
     private void maybeGenerateTlsCredentials(Reconciliation reconciliation, KafkaUserModel user, Secret userSecret) {
+
```
scholzj, line 323: Empty line not needed I guess.

user-operator/src/test/java/io/strimzi/operator/user/ResourceUtils.java
```diff
@@ -37,7 +38,9 @@ public class ResourceUtils {
     public static final String CA_KEY_NAME = "ca-key";
     public static final String PASSWORD = "my-password";
 
+
```
scholzj, line 41: Empty line not needed.

operator-common/src/main/java/io/strimzi/operator/common/operator/resource/ConfigParameterParser.java
ShubhamRwt: We are using it for `MAINTENENCE_TIME_WINDOW in the `UserOperatorConfig` class

user-operator/src/main/java/io/strimzi/operator/user/UserOperatorConfig.java
scholzj: You did that for the defaults. Can't it be done similarly? Or use the Builder instead of the maps?

operator-common/src/main/java/io/strimzi/operator/common/operator/resource/ConfigParameterParser.java
ShubhamRwt: Changing it to comma seperated lists was giving error

ShubhamRwt:
I have added the suggestions @scholzj

user-operator/src/main/java/io/strimzi/operator/user/UserOperatorConfig.java
ShubhamRwt: But the tests where the constants are being used are testing if the configuration are coming our correctly from the map. For eg. in the `UserOperatorConfigTest`.

ShubhamRwt:
```
        ENV_VARS.put(UserOperatorConfig.NAMESPACE.key(), "namespace");
        ENV_VARS.put(UserOperatorConfig.RECONCILIATION_INTERVAL_MS.key(), "30000");
```

ShubhamRwt: I can do something like this if its fine?

scholzj: Yeah, I think that is good.

operator-common/src/main/java/io/strimzi/operator/common/operator/resource/ConfigParameterParser.java
scholzj: That is fine, if it simecolon sepaarted it is semicolon separated. But you should not name it just `LIST`.

user-operator/src/main/java/io/strimzi/operator/user/UserOperatorConfig.java
ShubhamRwt: Great, I will make these changes then. Thanks Jakub

ShubhamRwt: Done

scholzj (approved):
LGTM. Thanks

scholzj:
@ShubhamRwt Did you actually try it? It looks like all the regression pipelines fail because the User Operator basically doesn't start at all.

ShubhamRwt:
Sorry, I think I didn’t tried it after the new commits when I made things more generic thinking it shall work(but it didn’t). I will have a look why this is happening.

user-operator/src/main/java/io/strimzi/operator/user/UserOperatorConfig.java
```diff
@@ -4,554 +4,407 @@
...
-    static final int DEFAULT_SCRAM_SHA_PASSWORD_LENGTH = 32;
+    public static final ConfigParameter<String> SECRET_PREFIX = new ConfigParameter<>("STRIMZI_SECRET_PREFIX", STRING, "", false, CONFIG_VALUES);
+    /**
+     * Number of days for which the certificate should be valid
+     */
+    public static final ConfigParameter<Integer> CERTS_VALIDITY_DAYS = new ConfigParameter<>("STRIMZI_CLIENTS_CA_VALIDITY", INTEGER, "30", true, CONFIG_VALUES);
+    /**
+     * How long before the certificate expiration should the user certificate be renewed
+     */
+    public static final ConfigParameter<Integer> CERTS_RENEWAL_DAYS = new ConfigParameter<>("STRIMZI_CLIENTS_CA_RENEWAL", INTEGER, "365", true, CONFIG_VALUES);
```
tombentley, line 80: Can these be negative? I suspect we should force them to be strictly positive. One way to do that would be to write a method in `ConfigParameterParse` something like this:

```java
    static <T extends Number> ConfigParameterParser<T> strictlyPositive(ConfigParameterParser<T> parser) {
        return new ConfigParameterParser<T>() {
            public T parse(String s) { 
                var value = parser.parse(s);
                if (value.longValue() <= 0) {
                    throw new ...
                }
               return value;
            }
        };
    }
```

then you can do

```java
public static final ConfigParameter<Integer> CERTS_VALIDITY_DAYS = new ConfigParameter<>("STRIMZI_CLIENTS_CA_VALIDITY", strictlyPositive(INTEGER), "30", true, CONFIG_VALUES);
```

You could write similar factories methods for decorators which applied other validation (e.g. number in a range, string matching a regex), but we can add those on a case-by-case basis.

user-operator/src/main/java/io/strimzi/operator/user/UserOperatorConfig.java
```diff
@@ -4,554 +4,407 @@
...
-    // Default values
-    static final long DEFAULT_FULL_RECONCILIATION_INTERVAL_MS = 120_000;
-    static final String DEFAULT_KAFKA_BOOTSTRAP_SERVERS = "localhost:9091";
+
+    private static final Map<String, ConfigParameter<?>> CONFIG_VALUES = new HashMap<>();
+
+    /**
+     * Namespace in which the operator will run and create resources
+     */
+    public static final ConfigParameter<String> NAMESPACE = new ConfigParameter<>("STRIMZI_NAMESPACE", NON_EMPTY_STRING, "", true, CONFIG_VALUES);
```
tombentley, line 36: I would really like to see two static factory methods on `ConfigParameter`:

```java
public static ConfigParameter<T> required(String key, ConfigParameterParser<T> type, Map<String, ConfigParameter<?>> map) { 
/* just calls the constructor */ 
}
ConfigParameter<T> optional(String key, ConfigParameterParser type, String defaultValue, Map<String, public static ConfigParameter<?>> map) {
 /* just calls the constructor */
}
````

That will make all these declarations much clearer, e.g.

```java
public static final ConfigParameter<String> NAMESPACE =  ConfigParameter.required("STRIMZI_NAMESPACE", NON_EMPTY_STRING, CONFIG_VALUES);
```

 and also meansthe reader doesn't have to do a double-take when they see a default of `""` for something with `NON_EMPTY_STRING` type. ("Oh, it's OK, the default is never used because it's a required paramete")

user-operator/src/main/java/io/strimzi/operator/user/UserOperatorConfig.java
```diff
@@ -4,554 +4,407 @@
...
+     */
+    public static final ConfigParameter<Integer> CERTS_VALIDITY_DAYS = new ConfigParameter<>("STRIMZI_CLIENTS_CA_VALIDITY", INTEGER, "30", true, CONFIG_VALUES);
+    /**
+     * How long before the certificate expiration should the user certificate be renewed
+     */
+    public static final ConfigParameter<Integer> CERTS_RENEWAL_DAYS = new ConfigParameter<>("STRIMZI_CLIENTS_CA_RENEWAL", INTEGER, "365", true, CONFIG_VALUES);
+    /**
+     * Length used for the Scram-Sha Password
+     */
+    public static final ConfigParameter<Integer> SCRAM_SHA_PASSWORD_LENGTH = new ConfigParameter<>("STRIMZI_SCRAM_SHA_PASSWORD_LENGTH", INTEGER, "32", true, CONFIG_VALUES);
```
tombentley, line 84: This one definitely has to be strictly positive.

ShubhamRwt:
@scholzj Can you run the regression tests on this so I can check if the fix worked

scholzj:
Well, did try it locally? Did it work for you @ShubhamRwt?

ShubhamRwt:
Yes @scholzj. All the tests in ‘UserST’ class were successful

ShubhamRwt:
Hi @scholzj. The test which is failing in the regression tests fails for me locally even with the quay images. Is it a flaky test?

user-operator/src/test/java/io/strimzi/operator/user/operator/KafkaUserOperatorMockTest.java
```diff
@@ -282,7 +280,7 @@ public void testCreateTlsUserWithACLsDisabled() throws ExecutionException, Inter
                     .endQuotas()
                 .endSpec()
                 .build();
-        KafkaUserOperator op = new KafkaUserOperator(ResourceUtils.createUserOperatorConfig(Map.of(), false, false, "32", null), client, mockCertManager, scramOps, quotasOps, aclOps, EXECUTOR);
+        KafkaUserOperator op = new KafkaUserOperator(ResourceUtils.createUserOperatorConfig(Map.of(), false, false, "32", ""), client, mockCertManager, scramOps, quotasOps, aclOps, EXECUTOR);
```
scholzj, line 283: Why did we moved from `null` to `""`?

ShubhamRwt: I did it because earlier I was using a single constructor while defining the ConfigParameter. I think with the new refactoring I can revert this change back

scholzj: I think with `null` it better captures the difference between not set and set to n empty string which is not always the same. This applies to the default values of the configuration options as well.

user-operator/src/main/java/io/strimzi/operator/user/UserOperatorConfig.java
```diff
@@ -155,6 +155,7 @@ public static UserOperatorConfig buildFromMap(Map<String, String> map) {
 
         Map<String, Object> generatedMap = ConfigParameter.define(envMap, CONFIG_VALUES);
 
+        // This is to make sure that if CA namespace is empty or null, then we can use the namespace value
         if (generatedMap.get(CA_NAMESPACE.key()) == null || String.valueOf(generatedMap.get(CA_NAMESPACE.key())).isEmpty()) {
             generatedMap.put(CA_NAMESPACE.key(), generatedMap.get(NAMESPACE.key()));
         }
```
scholzj, line 161: Shouldn't this be handled in the application code rather then here?

user-operator/src/main/java/io/strimzi/operator/user/UserOperatorConfig.java
```diff
@@ -4,554 +4,410 @@
...
+     */
+    public static final ConfigParameter<String> NAMESPACE = new ConfigParameter<>("STRIMZI_NAMESPACE", NON_EMPTY_STRING, CONFIG_VALUES);
+    /**
+     * Name of the secret containing the clients Certification Authority certificate.
+     */
+    public static final ConfigParameter<String> CA_CERT_SECRET_NAME = new ConfigParameter<>("STRIMZI_CA_CERT_NAME", NON_EMPTY_STRING, CONFIG_VALUES);
+    /**
+     * Name of the secret containing the cluster Certification Authority certificate.
+     */
+    public static final ConfigParameter<String> CLUSTER_CA_CERT_SECRET_NAME = new ConfigParameter<>("STRIMZI_CLUSTER_CA_CERT_SECRET_NAME", STRING, "", CONFIG_VALUES);
```
scholzj, line 45: The `null` comment applies also to all places like this ... why is the default `""` instead of `null`?

user-operator/src/main/java/io/strimzi/operator/user/UserOperatorConfig.java
ShubhamRwt: I tried to place a good place for it but it seems to fit here. Since the code we use for the generation of the configuration map, its generic so it will be used for topic and cluster operator. So putting it there wouldn't be much useful.

tombentley (approved):

user-operator/src/main/java/io/strimzi/operator/user/UserOperatorConfig.java
scholzj: The idea was more to handle it where the CA_NAMESPACE is actually used as that place knows best how to handle its values. Another option would be for example to have some method `getCaNamespaceOrNamespace` which will wrap this logic. This will allow you to have it centrally, but not modify the configuration. I think the way you do it right now is questionable because it distorts what is the actual configuration set by the user and how we use the configuration. I think having a clear separation of that is something that we should try to keep.

ShubhamRwt: Yeah that sounds good. I will have a look at where it is used

scholzj: FYI: One of the places uses it incorrectly. I already opened https://github.com/strimzi/strimzi-kafka-operator/pull/8235 to fix it there. so you should just ignore that one.

scholzj (approved):
Nice job. I will run the regression, but LGTM otherwise.

scholzj:
Thanks for the PR @ShubhamRwt

PR #7977, opened 2023-01-26
Allow Kubernetes API servers to list from cache instead of always asking for the latest state

scholzj:
### Type of change

- Refactoring

### Description

By default, when listing Kubernetes resources with the Fabric 8 Kubernetes client, it does not set the `resourceVersion` option in the query, which means we are requesting the most recent information. That might create a significant load on the API server as discussed in #7794. When setting the `resourceVersion` to `0`, we loosen the requirement for the most recent state which allows answering the query for example from an existing cache. (You can read more about the `resourceVersion` impact on the query in the [Kube docs](https://kubernetes.io/docs/reference/using-api/api-concepts/#semantics-for-get-and-list))

This PR tries to use `resourceVErsion=0` in our list queries to create less load on the Kubernetes API servers we communicate with. We should observe if we see any issues, but the expectation is that this should work fine.

### Checklist

### DIFF
--- a/operator-common/src/main/java/io/strimzi/operator/common/operator/resource/AbstractResourceOperator.java
+++ b/operator-common/src/main/java/io/strimzi/operator/common/operator/resource/AbstractResourceOperator.java
@@ -7,6 +7,7 @@ package io.strimzi.operator.common.operator.resource;
 import io.fabric8.kubernetes.api.model.HasMetadata;
 import io.fabric8.kubernetes.api.model.KubernetesResourceList;
 import io.fabric8.kubernetes.api.model.LabelSelector;
+import io.fabric8.kubernetes.api.model.ListOptionsBuilder;
 import io.fabric8.kubernetes.client.KubernetesClient;
 import io.fabric8.kubernetes.client.dsl.FilterWatchListDeletable;
 import io.fabric8.kubernetes.client.dsl.Listable;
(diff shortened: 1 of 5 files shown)


ppatierno:
Without getting the most recent resource, aren't we exposed to get a stale one and maybe making mistakes on our side during the reconcile? It's not clear to me how the Kubernetes cache mechanism works in order to update stale resources in the cache by reading them from ETCD when needed.

scholzj:
> Without getting the most recent resource, aren't we exposed to get a stale one and maybe making mistakes on our side during the reconcile? It's not clear to me how the Kubernetes cache mechanism works in order to update stale resources in the cache by reading them from ETCD when needed.

It seems that it is quite common to use the `resourceVersion=0` when you do not have a good reason to request the latest. The Fabric8 informers are using it as well for example to get started. I think the ask to not push the Kubernetes API server limits too much is reasonable. I also assume that the API server's intention is to not have the cache too stale. TBH, I cannot promise that this will have absolutely no impact. But without trying it, it will never be clear if there is anything or not. That is at the end why I opened it now and not before the 0.33 release.

ppatierno (approved):
LGTM, waiting to see if it will really work as per our discussion.

fvaleri:
Kube doc says;

>It is possible for the request to return data at a much older resource version that the client has previously observed, particularly in high availability configurations, due to partitions or stale caches. Clients that cannot tolerate this should not use this semantic.

I think this could trigger inconsistent behavior if we act based on the annotation or status information returned by the list operation. If this information is stale, we may apply the wrong logic. Could this be a concrete risk?

scholzj:
@fvaleri Sorry, I merged this before your comment showed up. I think some of that risk is already there. There is also no quantification of that risk. So, what I said above is exactly why my answer above says what it says.

PR #7980, opened 2023-01-26
Just testing the bridge endpoints refactoring

ppatierno:
Draft PR for an early testing of the bridge with endpoints refactoring due to https://github.com/strimzi/strimzi-kafka-bridge/pull/754

### DIFF
--- a/bridge.version
+++ b/bridge.version
@@ -1,1 +1,1 @@
-0.24.0
+endpoints-refactoring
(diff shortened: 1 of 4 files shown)


ppatierno:
So the test failure in regression-bundle I is not bridge related.
The test failure in bundle VII was on `HelmChartIsolatedST` and it was trying to push image with tag `endpoints-refactoring` from the strimzi repo (so not from my repo), because of the wrong helm chart template used. I run it locally and it worked, avoiding to run again entire regression here.
Going to close this and merging the https://github.com/strimzi/strimzi-kafka-bridge/pull/754 that  I was testing with this.

PR #7982, opened 2023-01-27
Add bridge pipeline

see-quick:

### Type of change

- Enhancement / new feature

### Description

This PR adds a new pipeline to our Azure (i.e., bridge pipeline). When Paolo is doing a release of Bridge, and he wants to double-check it, he would need to run the whole regression instead of only the Bridge profile. So this should resolve such an issue.

### Checklist

### DIFF
--- a/.azure/bridge-pipeline.yaml
+++ b/.azure/bridge-pipeline.yaml
@@ -1,0 +1,22 @@
+# Triggers
+# This pipeline will be triggered manually for a release or by github comment
+trigger: none
+pr:
+  autoCancel: false
+  branches:
+    include:
+      - '*'
+
+parameters:
+  - name: releaseVersion
+    displayName: Release Version
+    type: string
+    # If releaseVersion == latest then images will be built as part of the pipeline
+    default: "latest"
+
+jobs:
+  - template: 'templates/jobs/system-tests/bridge_jobs.yaml'
+    parameters:
+      releaseVersion: '${{ parameters.releaseVersion }}'
+
+
--- a/.azure/templates/jobs/system-tests/bridge_jobs.yaml
+++ b/.azure/templates/jobs/system-tests/bridge_jobs.yaml
@@ -1,0 +1,9 @@
+jobs:
+  - template: '../../steps/system_test_general.yaml'
+    parameters:
+      name: 'bridge'
+      display_name: 'bridge-bundle'
+      profile: 'bridge'
+      cluster_operator_install_type: 'bundle'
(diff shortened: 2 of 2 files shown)


scholzj (commented):
I'm not sure I follow the need for this. 
* Does it really run all the bridge tests including those which mix many different things such as tracing etc.?
* What difference does it really make? You do not make it easier to run the tests. You just add one more pipeline which will need to be maintained, tested when something changes etc.

If you would want to make things easier, this pipeline should accept the Bridge image as an input so that you can run it without opening a dummy PR and just run the tests manually against some branch with specified bridge image.

ppatierno:
I was having a conversation with Maros on this. Right now to run STs for bridge only we need to run the full 5 hours regression suite which includes all tests we don't need. About including other things, what I see is that tracing tests or oauth related to bridge are using the `bridge` tag so they will run.

scholzj:
@see-quick So, what is the plan for this PR?

see-quick:
> * Does it really run all the bridge tests including those which mix many different things such as tracing etc.?

Yes.

> * What difference does it really make? You do not make it easier to run the tests. You just add one more pipeline which will need to be maintained, tested when something changes etc.

Well, is is basically a sub-set of regression pipeline so if regression works that implicitly mean that bridge pipeline will work also. Meaning there is no need of more maintainability than we have now...

> If you would want to make things easier, this pipeline should accept the Bridge image as an input so that you can run it without opening a dummy PR and just run the tests manually against some branch with specified bridge image.

Yes, this is a good solution. If you want @ppatierno I can do in this way.

ppatierno:
@see-quick I am totally fine whatever is the simple solution for you as well. I want just to save time instead of waiting 5 hours regression. During the today's community call we were wondering if it really saves time. Can you estimate or do you know how much bridge tests will take?

see-quick:
> @see-quick I am totally fine whatever is the simple solution for you as well. I want just to save time instead of waiting 5 hours regression. During the today's community call we were wondering if it really saves time. Can you estimate or do you know how much bridge tests will take?

I have checked on my minikube, which mirrors Azure setup (i.e, 6900MB RAM, 2CPU) and it was done in 1h 20 minutes.

ppatierno:
Which is anyway much better than 5 ;-)

ppatierno:
@see-quick so AFAIU we decided to stay with opening a dummy PR for running the pipeline and not doing what Jakub suggested (to provide an image). Is it right?

Issue #7953, opened 2023-01-20
When KafkaRebalance resource is Ready, it should not transition to due to Kafka Cluster failure


scholzj:
Once the KafkaRebaance resource is in the Ready state, it should not transition to not ready just because the Kafka cluster moves to _unready_. It is already complete, so there seems to be no value in failing it.

scholzj:
Triaged on 26th January 2023: Looks like a bug.

kanha-gupta:
Can you provide some direction ? as i am beginner

scholzj:
@ppatierno I guess you will be able to guide @kanha-gupta through this

kanha-gupta:
@scholzj Hi ! i've tried to resolve this issue, Will improvise according to the directions given in future.
Please review if it solves the bug.
Thanks alot for support :)

ppatierno:
@kanha-gupta I have just replied on the PR where @ShubhamRwt did the same.

haanhvu:
I'll work on this

scholzj:
@haanhvu Thanks. @ShubhamRwt or @ppatierno might be the best guides for this if you need anything.

ppatierno:
@haanhvu I think that looking at the closed #7997 is a good start. It was almost done, we added some feedback and requests.

prashant235913:
Is this issue open....I want to solve it.

haanhvu:
@prashant235913 Yeah please go ahead. I plan to continue this this weekend. But if you want to work on this I'll find something else.

prashant235913:
Thanks @haanhvu for the support...As I am Beginner and it Can be my first contribution...How Can i get Some guidance of waht i am supposed to do.

haanhvu:
@prashant235913 I think the first step should be setting up the project. You can follow the development guide [here](https://github.com/strimzi/strimzi-kafka-operator/blob/main/development-docs/DEV_GUIDE.md).

If you face a problem during setting up that you don't understand or can't google for, you can ask about it in the Slack channels. If you haven't joined CNCF Slack workspace, get an invite [here](https://communityinviter.com/apps/cloud-native/cncf). If you're already in CNCF Slack, then find and join `strimzi` and `strimzi-dev` channels.

[...]

PR #7990, opened 2023-01-28
Fixed Transition happening by kafkaCluster failure

kanha-gupta:
Bugfix

Code uses inReady method if state is in ready condition to avoid transition

issue: 
https://github.com/strimzi/strimzi-kafka-operator/issues/7953

- [ ] Write tests
- [✓] Make sure all tests pass
- [ ] Update documentation
- [ ✓] Check RBAC rights for Kubernetes / OpenShift roles
- [ ] Update CHANGELOG.md


ShubhamRwt:
Thanks for the PR @kanha-gupta. I wonder if it is really the correct place to put the check. Can't we put the check somewhere in `KafkaRebalanceAssemblyOperator` itself? I guess you can place the check somewhere in `reconcileRebalance()` method where we check for the readiness of Kafka Cluster and update status based on that?

ppatierno:
@kanha-gupta Yeah I totally agree with @ShubhamRwt comment. The `api` module is not the right place and I am not even sure that it works. It doesn't take into account if the Kafka cluster is ready or not as defined in the issue.
This has to be fixed across the `reconcileRebalance` methods chain where you are reconciling the `KafkaRebalance` resources and we also already get information about the readiness of the Kafka cluster, other then getting the status of the rebalancing proposal itself. I would recommend to follow Shubham advice, taking a look at how the reconciliation works and it's related state machine.

PR #7993, opened 2023-01-30
Throw warning when `StatefulSets` are being used in Strimzi 0.34

scholzj:
### Type of change

- Task

### Description

StatefulSet support is planned to be removed in Strimzi 0.35. So it might make sense in Strimzi 0.34 to throw warnings about this. This PR adds warnings for any Kafka cluster using StatefulSets in the log and in its status.

Example of the status warning:

```yaml
    conditions:
    - lastTransitionTime: "2023-01-30T02:48:33.176487347Z"
      message: Support for StatefulSets will be removed in Strimzi 0.35. You should
        consider migrating to StrimziPodSets.
      reason: StatefulSetRemoval
      status: "True"
      type: Warning
```

And in the log:

```
2023-01-30 02:48:33 WARN  KafkaReconciler:295 - Reconciliation #1(watch) Kafka(myproject/my-cluster): Support for StatefulSets will be removed in Strimzi 0.35. You should consider migrating to StrimziPodSets.
```

### Checklist

### DIFF
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaReconciler.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaReconciler.java
@@ -19,6 +19,7 @@ import io.strimzi.api.kafka.model.KafkaResources;
 import io.strimzi.api.kafka.model.StrimziPodSet;
 import io.strimzi.api.kafka.model.StrimziPodSetBuilder;
 import io.strimzi.api.kafka.model.listener.arraylistener.GenericKafkaListener;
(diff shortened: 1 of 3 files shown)


ppatierno (approved):
LGTM. Should we have an issue to track to remove this when 0.35 will be released?

scholzj:
> Should we have an issue to track to remove this when 0.35 will be released?

TBH, I don't think it's needed. We will remove the feature gate as part of that => So that will automatically lead us to this code. If we open an issue I would be more afraid we forget to close it.

tombentley (approved):

fvaleri (approved):
LGTM.

PR #7997, opened 2023-01-30
Fixed Transition happening by kafkaCluster failure

kanha-gupta:
- Bugfix

### Description

Changes in KafkaRebalanceAssemblyOperator. Disabled the transition of KafkaRebalance resource when KafkaCluster is not ready

issue: https://github.com/strimzi/strimzi-kafka-operator/issues/7953

### Checklist

- [ ] Write tests
- [✓] Make sure all tests pass
- [ ] Update documentation
- [✓] Reference relevant issue(s) and close them after merging
- [ ] Update CHANGELOG.md


kanha-gupta:
@ShubhamRwt @ppatierno @scholzj 
Hi ! with the help of your guidance, I have made the necessary changes required.
I request you to pass a review & Thanks alot for guidance.
(I had to close previous PR due to a git error by my side).

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceAssemblyOperator.java
```diff
@@ -1170,8 +1170,7 @@ private Future<MapAndStatus<ConfigMap, KafkaRebalanceStatus>> onReady(Reconcilia
                             || kafka.getStatus().getConditions() == null
                             || kafka.getStatus().getConditions().stream().noneMatch(condition -> condition.getType().equals("Ready") && condition.getStatus().equals("True"))) {
                         LOGGER.warnCr(reconciliation, "Kafka cluster is not Ready");
-                        KafkaRebalanceStatus status = new KafkaRebalanceStatus();
-                        return updateStatus(reconciliation, kafkaRebalance, status,
+                        return updateStatus(reconciliation, kafkaRebalance, null,
```
fvaleri, line 1173: Here you are skipping the status update regardless of the current rebalance state, which you need to take into account to only skip the status update when the current state is ready.

fvaleri (changes requested):
Thanks for looking into this. I left a comment.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceAssemblyOperator.java
```diff
@@ -1170,8 +1170,7 @@ private Future<MapAndStatus<ConfigMap, KafkaRebalanceStatus>> onReady(Reconcilia
                             || kafka.getStatus().getConditions() == null
                             || kafka.getStatus().getConditions().stream().noneMatch(condition -> condition.getType().equals("Ready") && condition.getStatus().equals("True"))) {
```
ShubhamRwt, line 1171: @kanha-gupta You are now updating the status as `null` which means that even if `KafkaCluster` was `NotReady` from the start, it will not reflecting any updates in the `KafkaRebalance` CR status. As per the issue we just want to stop the transition of `KafkaRebalance` from `Ready` to `NotReady` state  if the kafka resource becomes `NotReady` and the kafka rebalance resource was already in `Ready` state. So ideally you can place some check in the `if` condition above(line 1169-1171) for checking if the rebalance resource is ready or not.

ShubhamRwt (changes requested):

kanha-gupta:
@ShubhamRwt @fvaleri Hi !
I've undone previous changes & instead added a check in If condition.
if rebalance is in ready state, it will not pass the condition & hence transition would be prevented.
Thanks alot for guidance. Learnt alot in the process :)

fvaleri (commented):
This is fine, but it also requires a new test in KafkaRebalanceAssemblyOperatorTest.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceAssemblyOperator.java
```diff
@@ -1168,7 +1168,7 @@ private Future<MapAndStatus<ConfigMap, KafkaRebalanceStatus>> onReady(Reconcilia
                                         + "' does not exist in namespace " + clusterNamespace + ".")).mapEmpty();
                     } else if (kafka.getStatus() == null
                             || kafka.getStatus().getConditions() == null
-                            || kafka.getStatus().getConditions().stream().noneMatch(condition -> condition.getType().equals("Ready") && condition.getStatus().equals("True"))) {
+                            || kafka.getStatus().getConditions().stream().noneMatch(condition -> condition.getType().equals("Ready") && condition.getStatus().equals("True")) && state(kafkaRebalance) != KafkaRebalanceState.Ready) {
```
ppatierno, line 1171: I think we are missing opening and closing parenthesis to group the "big" OR across the Kafka related status.
The && is just added at the end but because of the boolean operators precedence it's evaluated first so:

`true || false && false ===> true`
`(true || false) && false ===> false`

So I think we should factoring out a method like `isKafkaClusterReady` which evaluate the big chain of OR we have, and then having something like:

```java
if (!isKafkaClusterReady() && state(kafkaRebalance) != KafkaRebalanceState.Ready) {
...
}
```

@fvaleri @ShubhamRwt wdyt?

ppatierno (changes requested):

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceAssemblyOperator.java
ShubhamRwt: @ppatierno I agree with you. The changes are missing the opening and closing parenthesis to group the `OR` which is not correct. Yeah it would be great to have a method to avoid a big `OR` but if its not required then we should at least fix the parenthesis issue

ShubhamRwt:
@kanha-gupta One other advise would be to test the changes locally from minikube. That way you can be sure about whether the changes are really working or not. You can also test it by writing some unit tests

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceAssemblyOperator.java
fvaleri: @ppatierno you are absolutely right. 

I agree with the refactoring you propose, in fact I missed the grouping because it's hard to read. Let me add that the test should cover all permutations.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceAssemblyOperator.java
```diff
@@ -1166,9 +1166,9 @@ private Future<MapAndStatus<ConfigMap, KafkaRebalanceStatus>> onReady(Reconcilia
                                 new NoSuchResourceException("Kafka resource '" + clusterName
                                         + "' identified by label '" + Labels.STRIMZI_CLUSTER_LABEL
                                         + "' does not exist in namespace " + clusterNamespace + ".")).mapEmpty();
-                    } else if (kafka.getStatus() == null
+                    } else if ((kafka.getStatus() == null
                             || kafka.getStatus().getConditions() == null
-                            || kafka.getStatus().getConditions().stream().noneMatch(condition -> condition.getType().equals("Ready") && condition.getStatus().equals("True"))) {
+                            || kafka.getStatus().getConditions().stream().noneMatch(condition -> condition.getType().equals("Ready") && condition.getStatus().equals("True"))) && state(kafkaRebalance) != KafkaRebalanceState.Ready) {
```
ShubhamRwt, line 1171: @kanha-gupta As @ppatierno said, we can create a separate `isKafkaClusterReady` which can help us get rid of this big if block. I think it would be good to use it for cleaner code.

ShubhamRwt (changes requested):

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceAssemblyOperator.java
ppatierno: Yes I think that right now the code is unreadable. What I suggested would help.

ppatierno:
@kanha-gupta but did you test the change with a running Kubernetes and Kafka cluster running (simulating Kafka not ready)?
This is the minimum thing to do. Then we are not in a rush to fix this but we need a unit test as well. Merging without a test (manual or automatic) could break the current status.

fvaleri:
@kanha-gupta I would suggest to start from [here](https://github.com/strimzi/strimzi-kafka-operator/blob/main/development-docs/DEV_GUIDE.md). First check that you have all required software, then try to build the project following the instructions. Finally, you can try to run the KafkaRebalanceAssemblyOperatorTest from your favorite IDE or from command line.

```sh
$ mvn -Dmaven.javadoc.skip=true -Dcheckstyle.skip=true test \
  -pl cluster-operator -Dtest="KafkaRebalanceAssemblyOperatorTest"

[INFO] Tests run: 49, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 167.603 s - in io.strimzi.operator.cluster.operator.assembly.KafkaRebalanceAssemblyOperatorTest
...
[INFO] 
[INFO] Results:
[INFO] 
[INFO] Tests run: 49, Failures: 0, Errors: 0, Skipped: 0
[INFO] 
[INFO] ------------------------------------------------------------------------
[INFO] BUILD SUCCESS
[INFO] ------------------------------------------------------------------------
[INFO] Total time:  02:50 min
[INFO] Finished at: 2023-02-02T09:17:05+01:00
[INFO] ------------------------------------------------------------------------
```

After that, you can study the existing tests on that class and reuse the code to create your own test. You don't need Minikube to run the above test, but it's useful to do some manual testing or experiment with the system and be sure that your change actually works when integrating with all the other components.

There is no rush, so take the time you need. Let us know if something is not clear.

Hope it helps.

ppatierno:
@kanha-gupta why did you close the PR? Something you don't want to work on anymore after our requests for changes?

kanha-gupta:
@ppatierno greetings!
Actually I am encountering Numerous maven errors while setting up The project in my local environment. I've tried Linux too but unfortunately couldn't reach to a Build success.
This is the only reason why i am unable to progress with the testing required & it's been so much delay.
The reason for closing the PR was to give space for any other contributor in the community incase They want to fix this. 
I'll try to setup the project, if i can then I'll reopen PR :)
Thank you !

kanha-gupta:
Thanks to all of you for all the support !
You guys are amazing & I really wanted to be a contributor of this project.
I'll restart my Work once I get over the Setup part :)

PR #7998, opened 2023-01-30
Refactor Service and Config Map generation in AbstractModel and its subclasses

scholzj:
### Type of change

- Refactoring

### Description

This PR continues the refactoring of `AbstractModel` and its subclasses. It moves the Service and Config Map operations into separate utils classes and removes some of the related fields which were defined in the Abstract model but used only in few places.

### Checklist

- [x] Write tests

### DIFF
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/model/ServiceUtils.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/model/ServiceUtils.java
@@ -1,0 +1,244 @@
+/*
+ * Copyright Strimzi authors.
+ * License: Apache License 2.0 (see the file LICENSE or http://apache.org/licenses/LICENSE-2.0.html).
+ */
+package io.strimzi.operator.cluster.model;
+
+import io.fabric8.kubernetes.api.model.OwnerReference;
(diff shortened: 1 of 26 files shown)


cluster-operator/src/main/java/io/strimzi/operator/cluster/model/ServiceUtils.java
```diff
@@ -0,0 +1,244 @@
...
+     * @param template              Template with user's custom metadata for this service
+     * @param ports                 List of service ports
+     * @param selector              Selector for selecting the Pods to route the traffic to
+     * @param type                  Type of the service
+     * @param additionalLabels      Additional labels
+     * @param additionalAnnotations Additional annotations
+     * @param ipFamilyPolicy        IP Family Policy configuration
+     * @param ipFamilies            List of IP familiers
+     *
+     * @return  New discoverable ClusterIP Service
```
ppatierno, line 120: I guess it's enough `New ClusterIP Service` because this is the common method across the previous two. Only with `createDiscoverableClusterIpService` we get  a discoverable one.

ppatierno (approved):
LGTM. I left just a nit.

PR #8026, opened 2023-02-02
Add ZooKeeper certificate annotation per ZK pod in SPS

devguyio:
### Type of change

_Select the type of your PR_

- Bugfix

Followup for #7654 

### Description

To fix avoid the same behavior of #3913 when rolling ZooKeeper, this PR adds an annotation per pod that contains the ZooKeeper certificate thumbprint. That way in case of a certificate update, the ZooKeeper Roller would detect if the pods were not rolled consistently since pods would still be annotated by the old certificate thumbprint .

### Checklist

- [X] Write tests
- [ ] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaReconciler.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaReconciler.java
@@ -95,6 +95,7 @@ import java.util.stream.Collectors;
 import java.util.stream.IntStream;
 
 import static io.strimzi.operator.cluster.model.KafkaCluster.ANNO_STRIMZI_IO_KAFKA_VERSION;
+import static io.strimzi.operator.common.Annotations.ANNO_STRIMZI_SERVER_CERT_HASH;
 
 /**
  * Class used for reconciliation of Kafka. This class contains both the steps of the Kafka
@@ -808,7 +809,7 @@ public class KafkaReconciler {
         Map<String, String> podAnnotations = commonKafkaPodAnnotations(brokerId);
 
         // Annotation of broker certificate hash
-        podAnnotations.put(KafkaCluster.ANNO_STRIMZI_SERVER_CERT_HASH, kafkaServerCertificateHash.get(brokerId));
+        podAnnotations.put(Annotations.ANNO_STRIMZI_SERVER_CERT_HASH, kafkaServerCertificateHash.get(brokerId));
 
         return podAnnotations;
     }
@@ -1086,8 +1087,17 @@ public class KafkaReconciler {
         if (featureGates.useStrimziPodSetsEnabled())   {
             return maybeRollKafka(
                     podSetDiff.resource().getSpec().getPods().size(),
-                    pod -> ReconcilerUtils.reasonsToRestartPod(reconciliation, podSetDiff.resource(), pod, fsResizingRestartRequest, existingCertsChanged, clusterCa, clientsCa),
-                    listenerReconciliationResults.advertisedHostnames,
+                    pod -> ReconcilerUtils.reasonsToRestartPod(
+                            reconciliation,
+                            podSetDiff.resource(),
+                            pod,
+                            fsResizingRestartRequest,
+                            Annotations.hasAnnotationWithValue(
+                                    pod, ANNO_STRIMZI_SERVER_CERT_HASH,
+                                    kafkaServerCertificateHash.get(ANNO_STRIMZI_SERVER_CERT_HASH)),
+                            clusterCa,
+                            clientsCa
+                    ), listenerReconciliationResults.advertisedHostnames,
                     listenerReconciliationResults.advertisedPorts,
                     true
             );
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaReconciler.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaReconciler.java
@@ -619,17 +619,6 @@ public class KafkaReconciler {
     }
 
     /**
-     * Utility method to extract pod index number from pod name
-     *
-     * @param podName   Name of the pod
-     *
-     * @return          Index of the pod
-     */
-    private static int getPodIndexFromPodName(String podName)  {
-        return Integer.parseInt(podName.substring(podName.lastIndexOf("-") + 1));
-    }
-
-    /**
      * Generates and creates the ConfigMaps with per-broker configuration for Kafka brokers used in PodSets. It will
      * also delete the ConfigMaps for any scaled-down brokers (scale down is done before this is called in the
      * reconciliation)
@@ -664,7 +653,7 @@ public class KafkaReconciler {
                     // Create / update the desired config maps
                     for (ConfigMap cm : desiredConfigMaps) {
                         String cmName = cm.getMetadata().getName();
-                        int brokerId = getPodIndexFromPodName(cmName);
+                        int brokerId = ReconcilerUtils.getPodIndexFromPodName(cmName);
 
                         // The advertised hostname and port might change. If they change, we need to roll the pods.
                         // Here we collect their hash to trigger the rolling update. For per-broker configuration,
@@ -809,7 +798,7 @@ public class KafkaReconciler {
         Map<String, String> podAnnotations = commonKafkaPodAnnotations(brokerId);
 
         // Annotation of broker certificate hash
-        podAnnotations.put(Annotations.ANNO_STRIMZI_SERVER_CERT_HASH, kafkaServerCertificateHash.get(brokerId));
+        podAnnotations.put(ANNO_STRIMZI_SERVER_CERT_HASH, kafkaServerCertificateHash.get(brokerId));
 
         return podAnnotations;
     }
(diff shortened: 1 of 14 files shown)


cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/ZooKeeperReconciler.java
```diff
@@ -589,6 +600,20 @@ private Future<Void> podSet(int replicas) {
...
     }
 
+    /**
+     * Prepares annotations for ZooKeeper pods within a StrimziPodSet.
+     *
+     * @param podNum Number of the ZooKeeper pod, the annotations of which are being prepared.
+     * @return Map with Pod annotations
+     */
+    public Map<String, String> zkPodSetPodAnnotations(int podNum) {
+        Map<String, String> podAnnotations = new LinkedHashMap<>(2);
```
tombentley, line 610: To avoid inevitable reallocation I guess you need to change the initial capacity:

```suggestion
        Map<String, String> podAnnotations = new LinkedHashMap<>(1 + (int)(3/0.75));
```

tombentley (commented):
This looks reasonable, but I'd like to see a test which covers this fix.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaReconciler.java
```diff
@@ -1104,8 +1105,17 @@ protected Future<Void> rollingUpdate() {
...
-                    pod -> ReconcilerUtils.reasonsToRestartPod(reconciliation, podSetDiff.resource(), pod, fsResizingRestartRequest, existingCertsChanged, clusterCa, clientsCa),
-                    listenerReconciliationResults.advertisedHostnames,
+                    pod -> ReconcilerUtils.reasonsToRestartPod(
+                            reconciliation,
+                            podSetDiff.resource(),
+                            pod,
+                            fsResizingRestartRequest,
+                            Annotations.hasAnnotationWithValue(
+                                    pod, ANNO_STRIMZI_SERVER_CERT_HASH,
+                                    kafkaServerCertificateHash.get(ANNO_STRIMZI_SERVER_CERT_HASH)),
```
scholzj, line 1115: Why exactly did this changed?

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaReconciler.java
```diff
@@ -826,7 +827,7 @@ private Map<String, String> podSetPodAnnotations(int brokerId) {
         Map<String, String> podAnnotations = commonKafkaPodAnnotations(brokerId);
 
         // Annotation of broker certificate hash
-        podAnnotations.put(KafkaCluster.ANNO_STRIMZI_SERVER_CERT_HASH, kafkaServerCertificateHash.get(brokerId));
+        podAnnotations.put(Annotations.ANNO_STRIMZI_SERVER_CERT_HASH, kafkaServerCertificateHash.get(brokerId));
```
scholzj, line 830: You did static import of the annotation, so do you need the `Annotations.`?

scholzj (commented):
Some nits, but LGTM otherwise. Agree with Tom that a test would be nice.

scholzj:
PS: The regression pipeline failure is not related to your PR (the Build failure likely is)

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaReconciler.java
```diff
@@ -1104,8 +1094,17 @@ protected Future<Void> rollingUpdate() {
...
             return maybeRollKafka(
                     podSetDiff.resource().getSpec().getPods().size(),
-                    pod -> ReconcilerUtils.reasonsToRestartPod(reconciliation, podSetDiff.resource(), pod, fsResizingRestartRequest, existingCertsChanged, clusterCa, clientsCa),
-                    listenerReconciliationResults.advertisedHostnames,
+                    pod -> ReconcilerUtils.reasonsToRestartPod(
+                            reconciliation,
+                            podSetDiff.resource(),
+                            pod,
+                            fsResizingRestartRequest,
+                            !Annotations.hasAnnotationWithValue(
```
devguyio, line 1102: @scholzj this was missed in #7654 , this is to set the proper restart reason, that the cert has changed.

cluster-operator/src/test/java/io/strimzi/operator/cluster/ResourceUtils.java
```diff
@@ -674,7 +674,7 @@ public static ClusterOperatorConfig dummyClusterOperatorConfig(KafkaVersion.Look
     }
 
     public static ClusterOperatorConfig dummyClusterOperatorConfig(KafkaVersion.Lookup versions, long operationTimeoutMs) {
-        return dummyClusterOperatorConfig(versions, ClusterOperatorConfig.DEFAULT_OPERATION_TIMEOUT_MS, "");
+        return dummyClusterOperatorConfig(versions, operationTimeoutMs, "");
```
devguyio, line 677: while playing around with the tests, I noticed that this method ignores the `operationTimeoutMs` argument and I feel this is a mistake? hence this fix.

devguyio:
@scholzj @tombentley I've added two tests and applied the review comments.

cluster-operator/src/test/java/io/strimzi/operator/cluster/operator/assembly/KafkaAssemblyOperatorPodSetTest.java
```diff
@@ -176,25 +176,15 @@ public static void afterAll() {
...
-        StrimziPodSet zkPodSet = zkCluster.generatePodSet(KAFKA.getSpec().getZookeeper().getReplicas(), false, null, null, null);
+        StrimziPodSet zkPodSet = zkCluster.generatePodSet(KAFKA.getSpec().getZookeeper().getReplicas(), false, null, null, podNum -> null);
         KafkaCluster kafkaCluster = KafkaCluster.fromCrd(Reconciliation.DUMMY_RECONCILIATION, KAFKA, VERSIONS);
         StrimziPodSet kafkaPodSet = kafkaCluster.generatePodSet(KAFKA.getSpec().getKafka().getReplicas(), false, null, null, brokerId -> null);
 
         ResourceOperatorSupplier supplier = ResourceUtils.supplierWithMocks(false);
 
         SecretOperator secretOps = supplier.secretOperations;
         when(secretOps.reconcile(any(), any(), any(), any())).thenReturn(Future.succeededFuture());
-        when(secretOps.getAsync(NAMESPACE, KafkaResources.kafkaSecretName(CLUSTER_NAME))).thenReturn(
```
devguyio, line 187: Now you don't really need a mocked secret for the reconciliation to pass, as creation of the annotation will be ignored if no secret exists. So I cleaned up all of the "mocks" added previously since the're not needed now.

cluster-operator/src/test/java/io/strimzi/operator/cluster/ResourceUtils.java
scholzj: I think Shubham dealt with it in #8054?

scholzj (commented):
Sorry for not getting to it earlier while traveling. But it LGTM. Can you please rebase it to get rid of the conflicts so that we can run the tesst as well on it?

scholzj (approved):

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/ZooKeeperReconciler.java
```diff
@@ -432,11 +437,21 @@ protected Future<Void> certificateSecret(Clock clock) {
...
+                                    }
+                                    IntStream.range(0, zk.getReplicas())
+                                            .forEach(podNum -> {
+                                                var podName = KafkaResources.zookeeperPodName(reconciliation.name(), podNum);
+                                                zkCertificateHash.put(
+                                                        podNum,
+                                                        CertUtils.getCertificateThumbprint(patchResult.resource(),
+                                                                Ca.secretEntryNameForPod(podName, Ca.SecretEntry.CRT)
+                                                        ));
+                                            });
```
ppatierno, line 454: not something to change if you like this way but quite often I ask myself why an easy old school `for` loop isn't much better and readable than "complicated" Java Streams API logic (i.e. generating a stream of int instead of just using the `for` index to go). Answer to myself is that `for` loop always wins.

ppatierno (approved):

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/ZooKeeperReconciler.java
devguyio: Probably a personal preference for this small loop. Personally, I'm trying to use more functional style code, and when I'm in that mindset, my mind subconsciously never picks up a for loop specifically as a valid code construct. If was reviewing this PR and it was using a for loop, I probably wouldn't have noticed and would've approved it, it's just my mental state when I'm writing code lately.

ppatierno: I can totally understand the mindset of functional programming style and I totally agree when you have collections you want to work on by using streams API ... but generating a streams of integer instead of a for loop ... is too much to me ... it's diabolic :-P

tombentley: I also think a plain `for` statement would be clearer here. While there's nothing wrong _per se_ with preferring a functional style, if you're constructing a Stream simply to use `forEach` that's not functional (`forEach` is operating by side-effect, which is the opposite of functional).

@devguyio changing the code you _write_ to work around not seeing `for` as a valid code construct is not really a solution, since you still have to _read_ code written by others.

devguyio: I think my comment was misunderstood, I'm totally not saying this is better than the other, I just tried to explain the "faulty" thought process that lead to this loop. 

I changed this and the couple of ones in the tests.

scholzj: You happy with the change @tombentley?

scholzj:
Ok, I guess I'm going to merge it. Thanks for the PR @devguyio

PR #8027, opened 2023-02-02
[systemtest] Remove hardcoded link to `kafka-versions.yaml` from upgrade/downgrade tests

im-konge:

### Type of change

- Enhancement

### Description

This PR removes hardcoded link to `kafka-versions.yaml` from the upgrade/downgrade tests and pass it to `StrimziDowngradeST.yaml` and `StrimziUpgradeST.yaml`. This way we will be able to specify different url from where the `kafka-versions.yaml` should be taken, instead of editing the code.

### Checklist


systemtest/src/test/resources/upgrade/StrimziDowngradeST.yaml
```diff
@@ -30,6 +31,7 @@
   toExamples: strimzi-0.33.0
   urlFrom: HEAD
   urlTo: https://github.com/strimzi/strimzi-kafka-operator/releases/download/0.33.0/strimzi-0.33.0.zip
+  kafkaVersionsUrlFrom: HEAD
```
scholzj, line 34: Should this be a local file path instead?

scholzj (commented):
I think you should think a bit more about the name of the field and the values. It seems weird to have a URL here and some other definitions elsewhere. Ale the name seems weird ... what does `kafkaVersionsUrlFrom` mean? It is either a URL of the file or it is some identifier of a place the file should be taken from (such as `HEAD`)?

systemtest/src/test/resources/upgrade/StrimziDowngradeST.yaml
im-konge: As we have in `urlFrom`, if it is `HEAD`, we are getting it from the local classes or files -> as you can see in `getKafkaWithVersionFromUrl` -> if it is `HEAD`, the specific version from already existing `TestKafkaVersion` is returned insted of grabbing it from the URL.

im-konge:
Yeah I agree that now it is confusing.
So maybe I can changed it here to `beforeKafkaVersionsUrl` or `fromKafkaVersionsUrl` (whatever makes more sense), create an issue for changing it everywhere and handle those changes in next PR?

Frawless:
`from` makes better sense to me. Or completely remove before/from from that field

im-konge:
Changed it to `fromKafkaVersionsUrl`.

jankalinic (approved):
LGTM

im-konge:
Added issue to change all of the fields to be consistent https://github.com/strimzi/strimzi-kafka-operator/issues/8050.

systemtest/src/test/resources/upgrade/StrimziDowngradeST.yaml
```diff
@@ -5,6 +5,7 @@
 #  toExamples: String - Folder name which contains examples to upgrade / downgrade to. Downloaded from "urlTo"
 #  urlFrom: String - url, from which examples are downloaded
 #  urlTo: String - url, from which examples are downloaded during downgrade
+#  kafkaVersionsUrlFrom: String - url, from which kafka-versions.yaml is downloaded
```
scholzj, line 8:
```suggestion
#  fromKafkaVersionsUrl: String - url, from which kafka-versions.yaml is downloaded
```

systemtest/src/test/resources/upgrade/StrimziUpgradeST.yaml
```diff
@@ -3,6 +3,7 @@
 #  fromExamples: String - Folder name which contains examples to deploy. Downloaded from "urlFrom"
 #  oldestKafka: String - version of Kafka, used in initial deployment of Kafka CR
 #  urlFrom: String - url, from which examples are downloaded
+#  kafkaVersionsUrlFrom: String - url, from which is kafka-versions.yaml file downloaded
```
scholzj, line 6:
```suggestion
#  fromKafkaVersionsUrl: String - url, from which is kafka-versions.yaml file downloaded
```

scholzj (approved):

Frawless (approved):

PR #8037, opened 2023-02-03
Removes un-necessary test cases from systemtests

see-quick:

### Type of change

- Refactoring

### Description

This PR removes test cases which we assume are not needed:
1. `testKafkaOffsetsReplicationFactorHigherThanReplicas` - FMPOV, this is specifically a test case for integration level, not a system, and thus we are removing it.
2. `testDrainCleanerWithComponentsDuringNodeDraining` - this test case has been implemented. Still, we are not using it because of the PDB complexity (i.e., if we want to use it, then we will need to implement it for each platform in a different implementation). Moreover, there is also a need for a 9-node cluster.

### Checklist

- [ ] Write tests

### DIFF
--- a/systemtest/src/test/java/io/strimzi/systemtest/kafka/KafkaST.java
+++ b/systemtest/src/test/java/io/strimzi/systemtest/kafka/KafkaST.java
@@ -1593,24 +1593,6 @@ class KafkaST extends AbstractST {
     }
 
     @ParallelNamespaceTest
-    void testKafkaOffsetsReplicationFactorHigherThanReplicas(ExtensionContext extensionContext) {
-        final String namespaceName = StUtils.getNamespaceBasedOnRbac(namespace, extensionContext);
-        final String clusterName = mapWithClusterNames.get(extensionContext.getDisplayName());
-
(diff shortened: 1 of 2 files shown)


Frawless (approved):

im-konge (approved):

scholzj (commented):
> `testKafkaOffsetsReplicationFactorHigherThanReplicas`

So, does it have an integration test? Just because you think it should be an integration test probably does not justify the removal of this.

> `testDrainCleanerWithComponentsDuringNodeDraining`

This looks like the only test actually testing Drain Cleaner. I get that running it on a multinode cluster is not easy. But without that, we do not have any coverage for Drain Cleaner.

see-quick:
> So, does it have an integration test? Because you think it should be an integration test does not justify the removal of this.

We have the UT test case `testReplicasAndRelatedOptionsValidationNok` for verifying such a scenario.

scholzj (approved):
I think you should keep the Drain Cleaner test. But if you wanna remove it, your call I guess.

scholzj:
@see-quick Should this be merged?

PR #8048, opened 2023-02-05
Follow up cleanups after Pod and Container creation refactoring

scholzj:
### Type of change

- Refactoring

### Description

This PR does some follow-up cleanup to the previous refactorings of Pod and Container creation:
* Instead of parsing the Pod template and extracting the security context in every class in many places, we can move it to the `PodSecurityProviderContextImpl` class
* Moves the required environment variables which are passed to every container to `ContainerUtils` to have it with the containers
* Removes the unused `getEnvVars` method from `AbstractModel`

### Checklist

- [x] Write tests


ppatierno (approved):
LGTM

PR #8054, opened 2023-02-06
Minor mistake during overloading methods in the tests

ShubhamRwt:

### Type of change

_Select the type of your PR_

- Task

### Description

In the `dummyClusterOperatorConfig(KafkaVersion.Lookup versions, long operationTimeoutMs)` method we were passing both the values but no using it. Changing the value of `operationTimeoutMs` would lead to error since it was using default values in all cases.

### Checklist

- [ ] Write tests
- [ ] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/cluster-operator/src/test/java/io/strimzi/operator/cluster/ResourceUtils.java
+++ b/cluster-operator/src/test/java/io/strimzi/operator/cluster/ResourceUtils.java
@@ -673,10 +673,6 @@ public class ResourceUtils {
                 ClusterOperatorConfig.DEFAULT_POD_SECURITY_PROVIDER_CLASS, null);
     }
 
-    public static ClusterOperatorConfig dummyClusterOperatorConfig(KafkaVersion.Lookup versions, long operationTimeoutMs) {
-        return dummyClusterOperatorConfig(versions, operationTimeoutMs, "");
-    }
-
     public static ClusterOperatorConfig dummyClusterOperatorConfig(KafkaVersion.Lookup versions) {
         return dummyClusterOperatorConfig(versions, ClusterOperatorConfig.DEFAULT_OPERATION_TIMEOUT_MS, "");
     }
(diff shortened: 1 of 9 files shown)


scholzj (commented):
LGTM assuming it works. But the question is - if this did not worked for so long and nothing was failing, is it really needed?

scholzj (approved):

ShubhamRwt:
This method is currently being used by multiple tests. But I guess we don't need this method since earlier it was using the the default value of `operationTimeoutMs`. So I guess the best solution would be to either get rid of this method since we already have another method which only take `KafkaVersion.Lookup versions` as a argument and keeps the `operationTimeoutMs` as default

ppatierno:
@ShubhamRwt so are you going to get rid of this method updating this PR?

ShubhamRwt:
@ppatierno Yes, i was planning to do that but was just waiting for a suggestion from you or @scholzj

scholzj:
TBH, I do not know what to suggest. The tests clearly intended to use a different timeout. But they seem to pass fine without it. So other than _go through all the tests using it and figure one by one why were they trying to use this_ it is hard to give any suggestions. In any case, it seems at least some of the tests don't like your fix as the CI failed now twice?

ShubhamRwt:
Thanks @scholzj. I will investigate it and post the issue why it is failing now. I think it should pass since the test is probably there to make sure that everything works when we use different timeout. Maybe there is some bug lurking somwhere

scholzj:
Well, the tests now use the timeout they configure which is often much shorter than the default. So it could be that some of them are simply timing out because it is too short. In general, the shorter timeout is important when testing some timeouts etc. - as it detects the failure faster. It should not matter much for tests just waiting for successful completion.

ShubhamRwt:
Shouldn't the alternative that I suggested work in this case. To just remove the method since it is of no use currently since we are only setting one value?

scholzj:
Well, in general yes. As that is what worked until now 🤷

scholzj (approved):

scholzj:
@ppatierno Can you have a look at this?

ppatierno (approved):

PR #8056, opened 2023-02-06
Use default `replication.policy.separator` in docs and examples

scholzj:
### Type of change

- Bugfix

### Description

Our examples and docs covering the `IdentityReplicationPolicy` currently use `replication.policy.separator` set to empty string `""`. This does not seem to make much sense:
* The `IdentityReplicationPolicy` itself does not care about it
* But it is used to name some internal MM2 topics which end up with weird names with the empty string separator

This PR removes `replication.policy.separator` from the examples and updates the docs to use the default value / not suggest it is completely irrelevant when `IdentityReplicationPolicy` is used.

### Checklist

- [x] Update documentation

### DIFF
--- a/documentation/modules/configuring/con-config-mirrormaker2-connectors.adoc
+++ b/documentation/modules/configuring/con-config-mirrormaker2-connectors.adoc
@@ -30,7 +30,7 @@ The following table describes connector properties and the connectors you config
 |✓
 |✓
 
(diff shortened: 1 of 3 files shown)


mimaison (approved):
LGTM

fvaleri (approved):
LGTM. Thanks.

ppatierno (approved):

PaulRMellor (approved):

PR #8059, opened 2023-02-07
Refactor logging configuration in `AbstractModel` hierarchy

scholzj:
### Type of change

- Refactoring

### Description

This PR refactors the logging configuration in `AbstractModel` and its subclasses. As in previous cases, it moves most of the logic to a separate class named `LoggingUtils`  and uses it from the `AbstractModel` and its subclasses.

It also changes the default Log4j / Log4j2 configuration files => it moves them to a new sub-directory to have them better structured and names them after the class name to save storing one additional field which would differ for each class.

### Checklist

- [x] Write tests


ppatierno (approved):
LGTM

PR #8070, opened 2023-02-07
Added patterns to expose KRaft related metrics via JMX exporter

ppatierno:
This PR adds JMX exporter pattern to the Kafka metrics example in order to expose KRaft related metrics via Prometheus.
It conveys all the available metrics (as described here https://kafka.apache.org/documentation/#kraft_monitoring) which will be useful to make a KRaft dedicated dashboard or updating the Kafka one. It will come with a different PR because needs more investigation. Anyway with this PR, the user can start to take advantage of the KRaft metrics, to run their Prometheus queries or build their own Grafana dashboards.

### DIFF
--- a/examples/metrics/kafka-metrics.yaml
+++ b/examples/metrics/kafka-metrics.yaml
@@ -197,6 +197,22 @@ data:
       type: GAUGE
       labels:
         quantile: "0.$4"
+    # KRaft overall related metrics
+    # distinguish between always increasing COUNTER (total and max) and variable GAUGE (all others) metrics
+    - pattern: "kafka.server<type=raft-metrics><>(.+-total|.+-max):"
+      name: kafka_server_raftmetrics_$1
+      type: COUNTER
+    - pattern: "kafka.server<type=raft-metrics><>(.+):"
+      name: kafka_server_raftmetrics_$1
+      type: GAUGE
+    # KRaft "low level" channels related metrics
+    # distinguish between always increasing COUNTER (total and max) and variable GAUGE (all others) metrics
+    - pattern: "kafka.server<type=raft-channel-metrics><>(.+-total|.+-max):"
+      name: kafka_server_raftchannelmetrics_$1
+      type: COUNTER
+    - pattern: "kafka.server<type=raft-channel-metrics><>(.+):"
+      name: kafka_server_raftchannelmetrics_$1
+      type: GAUGE
   zookeeper-metrics-config.yml: |
     # See https://github.com/prometheus/jmx_exporter for more info about JMX Prometheus Exporter metrics
     lowercaseOutputName: true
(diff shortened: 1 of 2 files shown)


scholzj (changes requested):
All changes only to `packagin/...` please!

ppatierno:
damn, you are right! Fixed!

ppatierno:
> I guess the changes look good. But given that the KRaft mode is just an experimental thing at this point, should they really be part of the main configuration? Or should they be for example commented out? What is the performance impact of the rules which have no value?

Not an expert of JMX exporter but what we have in the PR will add 4 patterns that it will try to match. For a non KRaft cluster, there will be no match so no metrics exposed to Prometheus. I don't think this additional matching overloads JMX exporter. Anyway, I could agree on adding them but commented out.

> Also, if we want to keep them, should we have some KRaft dashabord which would use them and build it out as an alternative to the ZooKeeper dashabord for when KRaft is used?

As I mentioned in the description, yes ... we should have a new dashboard or maybe some more stuff on the Kafka one but it will come with a different PR. I don't think they are related. We can provide the way to export the KRaft metrics right now and the users can already build their own dashboard or just query them via Prometheus UI. Then, we'll come with an example later. I would need more investigation about what could make sense to have at least as a starting point.

scholzj:
> Not an expert of JMX exporter but what we have in the PR will add 4 patterns that it will try to match. For a non KRaft cluster, there will be no match so no metrics exposed to Prometheus. I don't think this additional matching overloads JMX exporter. Anyway, I could agree on adding them but commented out.

I think this is an issue that would be good to understand properly given how common complaints about the JMX Exporter being slow are. You say just 4 patterns. But does it not mean that the 4 patterns will be matched against every Kafka metric?

ppatierno:
> I think this is an issue that would be good to understand properly given how common complaints about the JMX Exporter being slow are. You say just 4 patterns. But does it not mean that the 4 patterns will be matched against every Kafka metric?

I can investigate the JMX exporter code if you want but if it works the way you are describing, so slowing down ... next question will be ... aren't we going to add more metrics patterns because JMX exporter is getting slow? So no more KRaft dashboard? In the end we'll need these metrics, or?

scholzj:
We will need these metrics ... but you are adding them basically for everyone at a point when nobody is really expected to use KRaft apart from some development work. So I think that is the main concern.

ppatierno:
Taking a quick look at JMX exporter source code it seems to get all beans and then for each beans try to match all the compiled patterns from the configuration file.

> We will need these metrics ... but you are adding them basically for everyone at a point when nobody is really expected to use KRaft apart from some development work. So I think that is the main concern.

I commented them out, so that if someone is playing with current KRaft mode he can get metrics.

scholzj (approved):

PR #8072, opened 2023-02-07
Reflect that `UseStrimziPodSets` feature gate is enabled by default in the docs

scholzj:
### Type of change

- Bugfix
- Documentation

### Description

The `UseStrimziPodSets` feature gate is now enabled by default. So when referring to StatefulSets versus StrimziPodSets, we should not state that StrimziPodSets are used when the feature gate is enabled but rather that StatefulSets is used when it is disabled. Otherwise, it might confuse users who did not enable anything and yet they now use StrimziPodSets.

### Checklist

- [x] Update documentation


documentation/modules/configuring/ref-list-of-kafka-cluster-resources.adoc
```diff
@@ -54,7 +54,7 @@ The following resources are created by the Cluster Operator in the Kubernetes cl
 `_cluster-name_-kafka-_listener-name_-_pod-id_`:: Service used to route traffic from outside the Kubernetes cluster to individual pods. This resource is created only when an external listener is enabled. The new service name will be used for all other external listeners.
 `_cluster-name_-kafka-_listener-name_-bootstrap`:: Bootstrap route for clients connecting from outside the Kubernetes cluster. This resource is created only when an external listener is enabled and set to type `route`. The new route name will be used for all other external listeners.
 `_cluster-name_-kafka-_listener-name_-_pod-id_`:: Route for traffic from outside the Kubernetes cluster to individual pods. This resource is created only when an external listener is enabled and set to type `route`. The new route name will be used for all other external listeners.
-`_cluster-name_-kafka-config`:: ConfigMap which contains the Kafka ancillary configuration and is mounted as a volume by the Kafka broker pods.
+`_cluster-name_-kafka-config`:: ConfigMap which contains the Kafka ancillary configuration and is mounted as a volume by the Kafka broker pods (if the `UseStrimziPodSets` feature gate is disabled).
```
PaulRMellor, line 57:
```suggestion
`_cluster-name_-kafka-config`:: ConfigMap containing the Kafka ancillary configuration, which is mounted as a volume by the broker pods when the `UseStrimziPodSets` feature gate is disabled.
```

PaulRMellor (approved):
Looks good.

PR #8090, opened 2023-02-11
Stable Pod identities for Kafka Connect and Mirror Maker 2

scholzj:
### Type of change

- Enhancement / new feature

### Description

This PR implements the [Strimzi Proposal 045](https://github.com/strimzi/proposals/blob/main/045-Stable-identities-for-Kafka-Connect-worker-nodes.md). It adds support to Kafka Connect and Kafka Mirror Maker 2 to run with StrimziPodSets and use stable pod names similarly to for example our Kafka brokers. This feature is protected through a Feature Gate which is in the alpha stage (i.e. disabled by default).

The only notable change compared to the proposal is that it does not create a dependency between `UseStrimziPodSets` feature gate and the new `StableConnectIdentities` feature gate. The `UseStrimziPodSets` feature gate is planned to be removed in the next Strimzi release, so we would need to change it again. It would also complicate the regression system tests since it would not be possible to test this in the _feature gates regression_ pipeline which has `USeStrimziPodSets` disabled.

The functionality for rolling updates of Kafka Connect / MM2 and for migration between Deployments and StrimziPodSets is done in separate classes so that it can be easily reused between Connect and MM2. It is non-trivial right now to use a single assembly operator to reconcile both Connect and MM2. That might be something to possibly look at in a different PR.

It also makes some necessary changes to the shared components. The StrimziPodSetController now needs to watch the Connect and MM2 Pods and reconcile them. It also updates the resource operators to allow them to influence the mechanism with which the different resources are updated easily - mainly to choose between patching and replacing as sometimes you might need to use a specific mechanism.

_This PR is currently a draft => the implementation is complete, but there are no unit tests for the new features._

### Checklist

- [ ] Write tests
- [x] Update documentation
- [x] Update CHANGELOG.md

### DIFF
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/ConnectBuildOperator.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/ConnectBuildOperator.java
@@ -77,11 +77,13 @@ public class ConnectBuildOperator {
     /**
      * Asynchronously runs the KafkaConnectBuild (if present) and reconciles the related resources
      *
-     * @param reconciliation    The reconciliation
-     * @param namespace         Namespace of the Connect cluster
-     * @param controllerResource       Name of the Connect cluster
-     * @param connectBuild      KafkaConnectBuild object
-     * @return                  Future for tracking the asynchronous result of the reconciliation steps
+     * @param reconciliation        The reconciliation
+     * @param namespace             Namespace of the Connect cluster
+     * @param controllerResource    The controller resource (Deployment or StrimziPodSet) with annotations describing
+     *                              the current state. Or null if it does not exist yet.
+     * @param connectBuild          KafkaConnectBuild object from the Kafka Connect custom resource
+     *
+     * @return  Future for tracking the asynchronous result of the reconciliation steps
      */
     public Future<BuildInfo> reconcile(Reconciliation reconciliation, String namespace, HasMetadata controllerResource, KafkaConnectBuild connectBuild) {
         if (connectBuild.getBuild() == null) {
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/ConnectBuildOperator.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/ConnectBuildOperator.java
@@ -6,9 +6,9 @@ package io.strimzi.operator.cluster.operator.assembly;
 
 import io.fabric8.kubernetes.api.model.ConfigMap;
 import io.fabric8.kubernetes.api.model.ContainerStateTerminated;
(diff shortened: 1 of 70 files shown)


cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/ConnectBuildOperator.java
```diff
@@ -82,11 +79,11 @@ public ConnectBuildOperator(PlatformFeaturesAvailability pfa, ResourceOperatorSu
      *
      * @param reconciliation    The reconciliation
      * @param namespace         Namespace of the Connect cluster
-     * @param connectName       Name of the Connect cluster
+     * @param controllerResource       Name of the Connect cluster
```
ppatierno, line 82: Javadoc ... it's not just the name anymore

ppatierno (approved):
Wow! What a PR, great work! I went through it and didn't find any clear issues but I didn't run it. Left just one nit. LGTM!

documentation/modules/operators/ref-operator-cluster-feature-gates.adoc
```diff
@@ -96,3 +96,17 @@ To enable the `UseKRaft` feature gate, specify `+UseKRaft` in the `STRIMZI_FEATU
...
 IMPORTANT: The `UseKRaft` feature gate depends on the `UseStrimziPodSets` feature gate.
 When enabling the `UseKRaft` feature gate, make sure that the `USeStrimziPodSets` feature gate is enabled as well.
+
+[id='ref-operator-stable-connect-identities-feature-gate-{context}']
+== StableConnectIdentities feature gate
+
+The `StableConnectIdentities` feature gate has a default state of _disabled_.
+
+The `StableConnectIdentities` feature gate uses `StrimziPodSet` resources to manage the Kafka Connect and Kafka Mirror Maker 2 instead of the Kubernetes Deployment resources.
+`StrimziPodSets` give the pods stable names and stable addresses which do not change during rolling upgrades.
```
PaulRMellor, line 106:
```suggestion
`StrimziPodSets` give the pods stable names and stable addresses, which do not change during rolling upgrades.
```

PaulRMellor (approved):
Great! :+1: 
A couple of typos and suggestions on the doc.

scholzj:
@tombentley Do you have any more comments? Thanks.

tombentley (approved):
The only other comments I have are wrt the "best" strategy for the `KafkaConnectRoller`. I wondered whether we should support logic for:

* trying to determine a best order (e.g. roll workers running fewest tasks first, so as to minimise the amount of disruption)
* or, after restarting a worker, whether we should be waiting for all the tasks which were running on that worker to get resume on some other worker. 

But I don't think we should hold up merging this PR. If we want to do that kind of improvement we can add it later.

PR #8096, opened 2023-02-13
Refactor JMX support in reconcilers and models

scholzj:
### Type of change

- Refactoring

### Description

We have currently 3 operands with JMX support (Kafka, Zoo, and Connect / MM2). Each of them has its own implementation. This PR refactors that and creates a separate `JmxModel` class and `reconcileJmxSecret` method which is shared across all of them. The `JmxModel` instance is created from the `.spec.zookeeper`, `.spec.kafka`, or `.spec` sections and used in the reconciliation. To be able to do this, this Pr also introduces some interfaces to be able to work with the different resource types. It also renames some of the internally used environment variables to use the same variable names in all containers.

### Checklist

- [x] Write tests


cluster-operator/src/main/java/io/strimzi/operator/cluster/model/JmxModel.java
```diff
@@ -0,0 +1,167 @@
...
+
+        if (specSection.getJmxOptions() != null)    {
+            this.isEnabled = true;
+            this.isAuthenticated = specSection.getJmxOptions().getAuthentication() != null
+                    && specSection.getJmxOptions().getAuthentication() instanceof KafkaJmxAuthenticationPassword;
+            this.template = specSection.getTemplate() != null ? specSection.getTemplate().getJmxSecret() : null;
+        } else {
+            this.isEnabled = false;
+            this.isAuthenticated = false;
+            this.template = null;
```
ppatierno, line 72: wondering if the above are just the defaults for boolean and an object so maybe we could avoid this "else" branch. Unless you put it here for clarity, which could make sense.

ppatierno (approved):
LGTM, great work! I left just a couple of nits.

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/JmxModel.java
scholzj: Well, if we want to keep the field `final`, we cannot set it to default value first and then override it. So to remove the `else` part, we would need to remove the `final` as well. So it seems better to me this way.

tombentley (approved):
Nice. Thanks.

PR #8100, opened 2023-02-14
docs(ssl): Updates the ssl and cipher suites config description

PaulRMellor:

**Documentation**

Updates the [ssl](https://strimzi.io/docs/operators/latest/configuring.html#con-common-configuration-ssl-reference) common configuration properties.  Changes example versions and also makes it clearer how `ssl.enabled.protocols` and `ssl.protocol` are used.

### Checklist

- [ ] Write tests
- [x] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/documentation/modules/con-common-configuration-properties.adoc
+++ b/documentation/modules/con-common-configuration-properties.adoc
@@ -44,12 +44,11 @@ A cipher suite is a set of algorithms for secure connection and data transfer.
 For example, a cipher suite specification might use a combination of the following algorithms:
 
 * TLS authentication
-* ECDHE (Elliptic Curve Diffie-Hellman Ephemeral) key exchange
-* RSA authentication
-* AES (Advanced Encryption Standard) encryption
+* AES (Advanced Encryption Standard) encryption (256-bit key)
+* GCM (Galois/Counter Mode) authenticated encryption
 * SHA384 (Secure Hash Algorithm) data integrity protection
 
-The combination is encapsulated in the `TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384` cipher suite specification.
+The combination is encapsulated in the `TLS_AES_256_GCM_SHA384` cipher suite specification.
 
 The `ssl.enabled.protocols` property specifies the available TLS versions that can be used for secure communication between the cluster and its clients. 
 The `ssl.protocol` property sets the default TLS version for all connections, and it must be chosen from the enabled protocols.
@@ -61,7 +60,7 @@ Use the `ssl.endpoint.identification.algorithm` property to enable or disable ho
 # ...
 spec:
   config:
-    ssl.cipher.suites: "TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384" # <1>
+    ssl.cipher.suites: "TLS_AES_256_GCM_SHA384" # <1>
     ssl.enabled.protocols: "TLSv1.3", "TLSv1.2" # <2>
     ssl.protocol: "TLSv1.3" # <3>
     ssl.endpoint.identification.algorithm: HTTPS # <4>
--- a/documentation/modules/con-common-configuration-properties.adoc
+++ b/documentation/modules/con-common-configuration-properties.adoc
@@ -39,6 +39,8 @@ When using Kafka with a Kafka cluster not managed by Strimzi, you can specify th
 
 You can incorporate SSL configuration and cipher suite specifications to further secure TLS-based communication between your client application and a Kafka cluster.
 In addition to the standard TLS configuration, you can specify a supported TLS version and enable cipher suites in the configuration for the Kafka broker.
+You can also add the configuration to your clients if you wish to limit the TLS versions and cipher suites they use. 
+The configuration on the client must only use protocols and cipher suites that are enabled on the broker. 
 
 A cipher suite is a set of algorithms for secure connection and data transfer.
 For example, a cipher suite specification might use a combination of the following algorithms:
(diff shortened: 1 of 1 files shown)


documentation/modules/con-common-configuration-properties.adoc
```diff
@@ -37,28 +37,39 @@ When using Kafka with a Kafka cluster not managed by Strimzi, you can specify th
...
 [source,yaml,subs="attributes+"]
 ----
 # ...
 spec:
   config:
-    ssl.cipher.suites: "TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384" <1>
-    ssl.enabled.protocols: "TLSv1.2" <2>
-    ssl.protocol: "TLSv1.2" <3>
-    ssl.endpoint.identification.algorithm: HTTPS <4>
+    ssl.cipher.suites: "TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384" # <1>
```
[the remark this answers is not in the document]
PaulRMellor: Good spot. I can change that.

documentation/modules/con-common-configuration-properties.adoc
```diff
@@ -37,28 +37,39 @@ When using Kafka with a Kafka cluster not managed by Strimzi, you can specify th
...
 # ...
 ----
-<1> The cipher suite for TLS using a combination of `ECDHE` key exchange mechanism, `RSA` authentication algorithm,
-`AES` bulk encyption algorithm and `SHA384` MAC algorithm.
-<2> The SSl protocol `TLSv1.2` is enabled.
-<3> Specifies the `TLSv1.2` protocol to generate the SSL context.
-Allowed values are `TLSv1.1` and `TLSv1.2`.
+<1> Cipher suite specification enabled.
+<2> TLS versions supported.
+<3> Default TLS version is `TLSv1.3`. If a client only supports `TLSv1.2`, it can still connect and communicate with the broker using that version instead.
```
scholzj, line 72: This seems a bit confusing -> the example is using path `.spec.config` which would suggest it is for clients. But the way the sentence here is phrased suggests it should be a server (a server - i.e. a Kafka broker - would have the path as `.spec.kafka.config`). I think these options can be applied for both sides. So should we make it more generic?
* Rephrase this with something like _Default TLS version is `TLSv1.3`. But if the other side only supports `TLSv1.2`, it can still connect and communicate with it using that version instead._
* Remove the `spec:` part from the example

scholzj (commented):
I guess we should try to use a working combination in the proposal - so I think the comment from Mickael makes sense.

documentation/modules/con-common-configuration-properties.adoc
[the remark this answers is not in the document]
PaulRMellor: Updated to use TLS_AES_256_GCM_SHA384 as the example.

documentation/modules/con-common-configuration-properties.adoc
PaulRMellor: Right. I've mentioned the vice versa aspect. I also thought it would be useful to mention that this config can be on either end.

scholzj (approved):
LGTM. Thanks.

documentation/modules/con-common-configuration-properties.adoc
```diff
@@ -37,28 +37,40 @@ When using Kafka with a Kafka cluster not managed by Strimzi, you can specify th
...
+You can incorporate SSL configuration and cipher suite specifications to further secure TLS-based communication between your client application and a Kafka cluster.
+In addition to the standard TLS configuration, you can specify a supported TLS version and enable cipher suites in the configuration for the Kafka broker.
+You can also add the configuration to your clients if you wish to limit the TLS versions and cipher suites they use. 
+The configuration on the client must only use protocols and cipher suites that are enabled on the broker. 
 
-You can also configure the `ssl.endpoint.identification.algorithm` property to enable or disable hostname verification.
+A cipher suite is a set of algorithms for secure connection and data transfer.
+For example, a cipher suite specification might use a combination of the following algorithms:
+
+* TLS authentication
```
ppatierno, line 48: I am not sure we can say the TLS authentication is an algorithm and it's part of a cipher suite ...

PaulRMellor: changed to:

> For example, the cipher suite 'TLS_AES_256_GCM_SHA384' is composed of the following mechanisms, which are used in conjunction with the TLS protocol:

ppatierno: Much better thanks!

ppatierno (approved):

documentation/modules/con-common-configuration-properties.adoc
```diff
@@ -37,28 +37,39 @@ When using Kafka with a Kafka cluster not managed by Strimzi, you can specify th
...
 ----
 # ...
-spec:
+
   config:
-    ssl.cipher.suites: "TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384" <1>
-    ssl.enabled.protocols: "TLSv1.2" <2>
-    ssl.protocol: "TLSv1.2" <3>
-    ssl.endpoint.identification.algorithm: HTTPS <4>
+    ssl.cipher.suites: "TLS_AES_256_GCM_SHA384" # <1>
```
mimaison, line 64: We only list a TLSv1.3 cipher so I'm still unsure if this will work in case TLSv1.2 is picked. I don't have an environment to test it but it may be simpler to add a TLSv1.2 cipher.

mimaison (approved):

PR #8106, opened 2023-02-15
docs(connect api): Using Kafka Connect REST API securely

PaulRMellor:
### Type of change

- Documentation

New section in the deploying guide describes how to use the Kafka Connect API more securely.
Includes config instructions for the following:

- Setting the `org.apache.kafka.disallowed.login.modules` property to prevent the use of insecure login modules.
- Setting the`connector.client.config.override.policy` property to "None" to prevent connector configurations from being overridden by unauthenticated requests.

### Checklist

- [ ] Write tests
- [x] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/documentation/modules/deploying/con-securing-kafka-connect-api.adoc
+++ b/documentation/modules/deploying/con-securing-kafka-connect-api.adoc
@@ -42,9 +42,9 @@ Only allow trusted login modules and follow the latest advice from Kafka for the
 For example, later versions of Kafka disable several insecure JNDI login modules by default. 
 As a best practice, you should explicitly disallow insecure login modules in your Kafka Connect configuration by using the `org.apache.kafka.disallowed.login.modules` system property.
 
-`connector.client.config.override.policy`:: Set the `connector.client.config.override.policy` property to `none` (default) to prevent connector configurations from being overridden by unauthenticated requests. 
+`connector.client.config.override.policy`:: Set the `connector.client.config.override.policy` property to `None` (default) to prevent connector configurations from being overridden by unauthenticated requests. 
 +
-.Example configuration for prevent configurations being overridden
+.Example configuration to specify connector override policy
 [source,yaml,subs=attributes+]
 ----
 apiVersion: {KafkaConnectApiVersion}
--- a/documentation/modules/deploying/con-securing-kafka-connect-api.adoc
+++ b/documentation/modules/deploying/con-securing-kafka-connect-api.adoc
@@ -12,17 +12,26 @@ Additionally, the user must have authenticated access to the Kubernetes cluster 
 
 It is crucial to restrict access to the Kafka Connect API only to trusted users to prevent unauthorized actions and potential security issues. 
 The Kafka Connect API provides extensive capabilities for altering connector configurations, which makes it all the more important to take security precautions.
+Someone with access to the Kafka Connect API could potentially obtain sensitive information that an administrator may assume is secure.
 
-If you are using the `KafkaConnector` custom resource to manage connectors, Kubernetes RBAC rules limit permission to make changes to Kubernetes cluster administrators by default.
+For example, suppose an organization uses a Kafka Connect cluster and connectors to stream sensitive data from a customer database to a central database. 
+The administrator uses a configuration provider plugin to store store sensitive information related to connecting to the customer database and the central database, such as database connection details and authentication credentials.  
+The configuration provider protects this sensitive information from being exposed to unauthorized users. 
+However, someone who has access to the Kafka Connect API can still obtain access to the customer database without the consent of the administrator.
+They can do this by setting up a fake database and configuring a connector to connect to it. 
+They then modify the connector configuration to point to the customer database, but instead of sending the data to the central database, they send it to the fake database.
+By configuring the connector to connect to the fake database, the login details and credentials for connecting to the customer database are intercepted, even though they are stored securely in the configuration provider.
+
+If you are using the `KafkaConnector` custom resources, then by default the Kubernetes RBAC rules permit only Kubernetes cluster administrators to make changes to connectors.
 You can also xref:adding-users-the-strimzi-admin-role-str[designate non-cluster administrators to manage Strimzi resources].  
 With `KafkaConnector` resources enabled in your Kafka Connect configuration, changes made directly using the Kafka Connect REST API are reverted by the Cluster Operator.
-If you are not using the `KafkaConnector` resource, the default RBAC rules do not limit access to the Kafka Connect API directly.
-If you want to limit access to the Kafka Connect REST API using Kubernetes RBAC, you need to enable and use the `KafkaConnector` resources. 
+If you are not using the `KafkaConnector` resource, the default RBAC rules do not limit access to the Kafka Connect API.
+If you want to limit direct access to the Kafka Connect REST API using Kubernetes RBAC, you need to enable and use the `KafkaConnector` resources. 
 
 For improved security, we recommend configuring the following properties for the Kafka Connect API:
 
 `org.apache.kafka.disallowed.login.modules`:: (Kafka 3.4 or later) Set the `org.apache.kafka.disallowed.login.modules` Java system property to prevent the use of insecure login modules. 
-For example, specifying `org.apache.kafka.common.security.kerberos.KerberosLoginModule` prevents the use of the Kafka `KerberosLoginModule`.
+For example, specifying `com.sun.security.auth.module.JndiLoginModule` prevents the use of the Kafka `JndiLoginModule`.
 +
 .Example configuration for disallowing login modules
 [source,yaml,subs=attributes+]
@@ -41,8 +50,6 @@ spec:
         value: com.sun.security.auth.module.JndiLoginModule, org.apache.kafka.common.security.kerberos.KerberosLoginModule
 # ...      
 ----
-+
-System properties are passed to the JVM using the standard `-D` option.
 +
 Only allow trusted login modules and follow the latest advice from Kafka for the version you are using.
 As a best practice, you should explicitly disallow insecure login modules in your Kafka Connect configuration by using the `org.apache.kafka.disallowed.login.modules` system property.
--- a/documentation/modules/deploying/con-securing-kafka-connect-api.adoc
+++ b/documentation/modules/deploying/con-securing-kafka-connect-api.adoc
@@ -1,0 +1,63 @@
+// This assembly is included in the following assemblies:
+//
+// assembly-deploy-kafka-connect-with-plugins.adoc
+
+[id='con-securing-kafka-connect-api-{context}']
+= Using the Kafka Connect API securely
+
+[role="_abstract"]
+To grant a user access to the Kafka Connect API, a Kafka cluster admin must provide them with the API URL. 
+This includes the hostname or IP address of the Kafka Connect cluster and the port number on which the API is running (by default, port 8083). 
+Additionally, the user must have authenticated access to the Kubernetes cluster and the Kafka Connect REST API endpoint in order to make a request.
+
+It is crucial to restrict access to the Kafka Connect API only to trusted users to prevent unauthorized actions and potential security issues. 
+The Kafka Connect API provides extensive capabilities for altering connector configurations, which makes it all the more important to take security precautions.
+
+For improved security, we recommend configuring the following properties for the Kafka Connect API:
+
+`org.apache.kafka.disallowed.login.modules`:: Set the `org.apache.kafka.disallowed.login.modules` Java system property to prevent the use of insecure login modules. 
+For example, specifying `org.apache.kafka.common.security.plain.PlainLoginModule` prevents the use of the Kafka `PlainLoginModule`.
++
+.Example configuration for disallowing login modules
+[source,yaml,subs=attributes+]
+----
+apiVersion: {KafkaConnectApiVersion}
+kind: KafkaConnect
+metadata:
+  name: my-connect-cluster
+  annotations:
+    strimzi.io/use-connector-resources: "true" 
+spec:
+  # ...
+  jvmOptions:
+    javaSystemProperties:
+      - name: org.apache.kafka.disallowed.login.modules
+        value: org.apache.kafka.common.security.plain.PlainLoginModule
+# ...      
+----
++
+System properties are passed to the JVM using the standard `-D` option.
++
+Only allow trusted login modules and follow the latest advice from Kafka for the version you are using.
+For example, later versions of Kafka disable several insecure JNDI login modules by default. 
+As a best practice, you should explicitly disallow insecure login modules in your Kafka Connect configuration by using the `org.apache.kafka.disallowed.login.modules` system property.
+
+`connector.client.config.override.policy`:: Set the `connector.client.config.override.policy` property to `none` (default) to prevent connector configurations from being overridden by unauthenticated requests. 
++
+.Example configuration for prevent configurations being overridden
+[source,yaml,subs=attributes+]
+----
+apiVersion: {KafkaConnectApiVersion}
+kind: KafkaConnect
+metadata:
+  name: my-connect-cluster
+  annotations:
+    strimzi.io/use-connector-resources: "true" 
+spec:
+  # ...
+  config:
+    connector.client.config.override.policy: None
+# ...      
+----
++
+NOTE: You can also configure `connector.client.config.override.policy` at the connector level. 
--- a/documentation/modules/deploying/con-securing-kafka-connect-api.adoc
+++ b/documentation/modules/deploying/con-securing-kafka-connect-api.adoc
@@ -13,11 +13,11 @@ Additionally, the user must have authenticated access to the Kubernetes cluster 
 It is crucial to restrict access to the Kafka Connect API only to trusted users to prevent unauthorized actions and potential security issues. 
 The Kafka Connect API provides extensive capabilities for altering connector configurations, which makes it all the more important to take security precautions.
 
-If you are using the `KafkaConnector` custom resource to manage connectors, Strimzi's RBAC rules limit permission to make changes to Kubernetes cluster administrators by default.
+If you are using the `KafkaConnector` custom resource to manage connectors, Kubernetes RBAC rules limit permission to make changes to Kubernetes cluster administrators by default.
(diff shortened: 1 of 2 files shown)


PaulRMellor:
Peer review requested from @joaedwar

documentation/modules/deploying/con-securing-kafka-connect-api.adoc
```diff
@@ -0,0 +1,63 @@
...
+metadata:
+  name: my-connect-cluster
+  annotations:
+    strimzi.io/use-connector-resources: "true" 
+spec:
+  # ...
+  jvmOptions:
+    javaSystemProperties:
+      - name: org.apache.kafka.disallowed.login.modules
+        value: org.apache.kafka.common.security.plain.PlainLoginModule
```
scholzj, line 35: I guess this is a bad example because it essentially re-enabled the JNDI module with the CVE?

documentation/modules/deploying/con-securing-kafka-connect-api.adoc
```diff
@@ -0,0 +1,63 @@
...
+
+[role="_abstract"]
+To grant a user access to the Kafka Connect API, a Kafka cluster admin must provide them with the API URL. 
+This includes the hostname or IP address of the Kafka Connect cluster and the port number on which the API is running (by default, port 8083). 
+Additionally, the user must have authenticated access to the Kubernetes cluster and the Kafka Connect REST API endpoint in order to make a request.
+
+It is crucial to restrict access to the Kafka Connect API only to trusted users to prevent unauthorized actions and potential security issues. 
+The Kafka Connect API provides extensive capabilities for altering connector configurations, which makes it all the more important to take security precautions.
+
+For improved security, we recommend configuring the following properties for the Kafka Connect API:
```
scholzj, line 16: Thsi is Kafka 3.4+ only. Should be mentioned there.

documentation/modules/deploying/con-securing-kafka-connect-api.adoc
```diff
@@ -0,0 +1,63 @@
+// This assembly is included in the following assemblies:
+//
+// assembly-deploy-kafka-connect-with-plugins.adoc
+
+[id='con-securing-kafka-connect-api-{context}']
+= Using the Kafka Connect API securely
```
scholzj, line 6: Technically, this is correct. But the reality is that we have a big gap around security here. So I wonder if we can have some more neutral title. Otherwise users will come and ask why does this say nothing about authentication, TLS etc.

documentation/modules/deploying/con-securing-kafka-connect-api.adoc
```diff
@@ -0,0 +1,63 @@
...
+# ...      
+----
++
+System properties are passed to the JVM using the standard `-D` option.
++
+Only allow trusted login modules and follow the latest advice from Kafka for the version you are using.
+For example, later versions of Kafka disable several insecure JNDI login modules by default. 
+As a best practice, you should explicitly disallow insecure login modules in your Kafka Connect configuration by using the `org.apache.kafka.disallowed.login.modules` system property.
+
+`connector.client.config.override.policy`:: Set the `connector.client.config.override.policy` property to `None` (default) to prevent connector configurations from being overridden by unauthenticated requests.
```
mimaison, line 45: Disabling all overrides may restrict use cases and if you are running connectors with overrides, this may change their behavior.

documentation/modules/deploying/con-securing-kafka-connect-api.adoc
```diff
@@ -0,0 +1,63 @@
...
+  annotations:
+    strimzi.io/use-connector-resources: "true" 
+spec:
+  # ...
+  config:
+    connector.client.config.override.policy: None
+# ...      
+----
++
+NOTE: You can also configure `connector.client.config.override.policy` at the connector level.
```
mimaison, line 63: Can you? I thought you could only set this in the runtime so it would always apply to all connectors.

documentation/modules/deploying/con-securing-kafka-connect-api.adoc
PaulRMellor: I swapped for `KerberosLoginModule`

documentation/modules/deploying/con-securing-kafka-connect-api.adoc
```diff
@@ -0,0 +1,63 @@
...
+
+[id='con-securing-kafka-connect-api-{context}']
+= Using the Kafka Connect API securely
+
+[role="_abstract"]
+To grant a user access to the Kafka Connect API, a Kafka cluster admin must provide them with the API URL. 
+This includes the hostname or IP address of the Kafka Connect cluster and the port number on which the API is running (by default, port 8083). 
+Additionally, the user must have authenticated access to the Kubernetes cluster and the Kafka Connect REST API endpoint in order to make a request.
+
+It is crucial to restrict access to the Kafka Connect API only to trusted users to prevent unauthorized actions and potential security issues.
```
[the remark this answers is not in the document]
PaulRMellor: Do we also want to say anything about adding RBAC rules for Connect API? I added this:

> If you are using the `KafkaConnector` custom resource to manage connectors, Strimzi's RBAC rules limit permission to make changes to Kubernetes cluster administrators by default.
> You can also xref:adding-users-the-strimzi-admin-role-str[designate non-cluster administrators to manage Strimzi resources].  
> With `KafkaConnector` resources enabled in your Kafka Connect configuration, changes made directly using the Kafka Connect REST API are reverted by the Cluster Operator.
> If you are not using the`KafkaConnector` resource, the default RBAC rules do not limit access to the Kafka Connect API directly.
> If you want to limit access to the Kafka Connect REST API using Kubernetes RBAC, you need to set up additional RBAC rules. 

If that's accurate/required, maybe we want to add a procedure about setting up RBAC role?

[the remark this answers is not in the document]
scholzj: I think this is fine.

documentation/modules/deploying/con-securing-kafka-connect-api.adoc
```diff
@@ -0,0 +1,63 @@
...
+    javaSystemProperties:
+      - name: org.apache.kafka.disallowed.login.modules
+        value: org.apache.kafka.common.security.plain.PlainLoginModule
+# ...      
+----
++
+System properties are passed to the JVM using the standard `-D` option.
++
+Only allow trusted login modules and follow the latest advice from Kafka for the version you are using.
+For example, later versions of Kafka disable several insecure JNDI login modules by default.
```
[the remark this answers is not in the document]
PaulRMellor: I'm not sure we need this line at all. I've removed it give more general advice that won't become out-of-date. 
Let me know if you think we should be specific here.

documentation/modules/deploying/con-securing-kafka-connect-api.adoc
PaulRMellor: @mimaison  Do we need to say anything on this? I'm not sure how the behavior might change?

documentation/modules/deploying/con-securing-kafka-connect-api.adoc
PaulRMellor: @mimaison Do we need to say anything on this? I'm not sure how the behavior might change?

documentation/modules/deploying/con-securing-kafka-connect-api.adoc
PaulRMellor: Removed

documentation/modules/deploying/con-securing-kafka-connect-api.adoc
```diff
@@ -0,0 +1,66 @@
...
+
+[role="_abstract"]
+To grant a user access to the Kafka Connect API, a Kafka cluster admin must provide them with the API URL. 
+This includes the hostname or IP address of the Kafka Connect cluster and the port number on which the API is running (by default, port 8083). 
+Additionally, the user must have authenticated access to the Kubernetes cluster and the Kafka Connect REST API endpoint in order to make a request.
+
+It is crucial to restrict access to the Kafka Connect API only to trusted users to prevent unauthorized actions and potential security issues. 
+The Kafka Connect API provides extensive capabilities for altering connector configurations, which makes it all the more important to take security precautions.
+
+If you are using the `KafkaConnector` custom resource to manage connectors, Strimzi's RBAC rules limit permission to make changes to Kubernetes cluster administrators by default.
```
scholzj, line 16:
```suggestion
If you are using the `KafkaConnector` custom resource to manage connectors, Kubernetes RBAC rules limit permission to make changes to Kubernetes cluster administrators by default.
```

documentation/modules/deploying/con-securing-kafka-connect-api.adoc
```diff
@@ -0,0 +1,66 @@
...
+You can also xref:adding-users-the-strimzi-admin-role-str[designate non-cluster administrators to manage Strimzi resources].  
+With `KafkaConnector` resources enabled in your Kafka Connect configuration, changes made directly using the Kafka Connect REST API are reverted by the Cluster Operator.
+If you are not using the `KafkaConnector` resource, the default RBAC rules do not limit access to the Kafka Connect API directly.
+If you want to limit access to the Kafka Connect REST API using Kubernetes RBAC, you need to set up additional RBAC rules.
```
scholzj, line 20: There is no way how to link Kubernetes RBAC with the REST API. You would need to use the `KafkaConnector` resources.

```suggestion
If you want to limit access to the Kafka Connect REST API using Kubernetes RBAC, you need to enable the Connector operator and use the `KafkaConnector` resources. 
```

documentation/modules/deploying/con-securing-kafka-connect-api.adoc
```diff
@@ -0,0 +1,66 @@
...
+metadata:
+  name: my-connect-cluster
+  annotations:
+    strimzi.io/use-connector-resources: "true" 
+spec:
+  # ...
+  jvmOptions:
+    javaSystemProperties:
+      - name: org.apache.kafka.disallowed.login.modules
+        value: org.apache.kafka.common.security.kerberos.KerberosLoginModule
```
scholzj, line 41: My main point here was the following ... You show how to set a custom disabled module. But I would expect that to override the default from Kafka which is `com.sun.security.auth.module.JndiLoginModule` and essentialy expose the user again to the original CVE. So I think you need to have here something like this for example:

```suggestion
      - name: org.apache.kafka.disallowed.login.modules
        value: com.sun.security.auth.module.JndiLoginModule,org.apache.kafka.common.security.kerberos.KerberosLoginModule
```

documentation/modules/deploying/con-securing-kafka-connect-api.adoc
```diff
@@ -0,0 +1,66 @@
...
+# ...      
+----
++
+System properties are passed to the JVM using the standard `-D` option.
++
+Only allow trusted login modules and follow the latest advice from Kafka for the version you are using.
+As a best practice, you should explicitly disallow insecure login modules in your Kafka Connect configuration by using the `org.apache.kafka.disallowed.login.modules` system property.
+
+`connector.client.config.override.policy`:: Set the `connector.client.config.override.policy` property to `None` (default) to prevent connector configurations from being overridden by unauthenticated requests.
```
scholzj, line 50: Maybe this would work better?

```suggestion
`connector.client.config.override.policy`:: Set the `connector.client.config.override.policy` property to `None` (default) to prevent connector configurations to override the configuration of Kafka Connect and the consumers and producers used by it.
Before setting it to `None`, make sure that any connectors you have already deployed do not depend on the custom configuration.
```

documentation/modules/deploying/con-securing-kafka-connect-api.adoc
PaulRMellor: Ah, right. I understand what you were saying now

scholzj (approved):

documentation/modules/deploying/con-securing-kafka-connect-api.adoc
```diff
@@ -0,0 +1,66 @@
...
+[id='con-securing-kafka-connect-api-{context}']
+= Limiting access to the Kafka Connect API
+
+[role="_abstract"]
+To grant a user access to the Kafka Connect API, a Kafka cluster admin must provide them with the API URL. 
+This includes the hostname or IP address of the Kafka Connect cluster and the port number on which the API is running (by default, port 8083). 
+Additionally, the user must have authenticated access to the Kubernetes cluster and the Kafka Connect REST API endpoint in order to make a request.
+
+It is crucial to restrict access to the Kafka Connect API only to trusted users to prevent unauthorized actions and potential security issues. 
+The Kafka Connect API provides extensive capabilities for altering connector configurations, which makes it all the more important to take security precautions.
```
tombentley, line 14: I think we need an example here to really make the point that people with the ability to configure connectors can use that information to obtain information the administrator might otherwise think is secure. For example, setting up a fake database and then configuring a connector to connect to it would be used to obtain a database password that is otherwise protected via the config provider mechanism. Therefore the administrator should assume that the people who have the ability to configure connectors in effect have complete control over the connect cluster and resources, such as secrets, that are available to it.

documentation/modules/deploying/con-securing-kafka-connect-api.adoc
```diff
@@ -0,0 +1,66 @@
...
+
+If you are using the `KafkaConnector` custom resource to manage connectors, Kubernetes RBAC rules limit permission to make changes to Kubernetes cluster administrators by default.
```
tombentley, line 16: I found this a little difficult to parse. I think the "by default" being at the end doesn't help:

```suggestion
If you are using the `KafkaConnector` custom resources then by default the Kubernetes RBAC rules permit only Kubernetes cluster administrators to make changes to connectors.
```

documentation/modules/deploying/con-securing-kafka-connect-api.adoc
```diff
@@ -0,0 +1,66 @@
...
+You can also xref:adding-users-the-strimzi-admin-role-str[designate non-cluster administrators to manage Strimzi resources].  
+With `KafkaConnector` resources enabled in your Kafka Connect configuration, changes made directly using the Kafka Connect REST API are reverted by the Cluster Operator.
+If you are not using the `KafkaConnector` resource, the default RBAC rules do not limit access to the Kafka Connect API directly.
```
tombentley, line 19: the "directly" here could be misinterpreted to mean "they do limit access, but in an indirect way". 

```suggestion
If you are not using the `KafkaConnector` resource, the default RBAC rules do not limit direct access to the Kafka Connect API.
```

documentation/modules/deploying/con-securing-kafka-connect-api.adoc
```diff
@@ -0,0 +1,66 @@
...
+If you want to limit access to the Kafka Connect REST API using Kubernetes RBAC, you need to enable and use the `KafkaConnector` resources.
```
tombentley, line 20:
```suggestion
If you want to limit direct access to the Kafka Connect REST API using Kubernetes RBAC, you need to enable and use the `KafkaConnector` resources. 
```

documentation/modules/deploying/con-securing-kafka-connect-api.adoc
```diff
@@ -0,0 +1,66 @@
...
+spec:
+  # ...
+  jvmOptions:
+    javaSystemProperties:
+      - name: org.apache.kafka.disallowed.login.modules
+        value: com.sun.security.auth.module.JndiLoginModule, org.apache.kafka.common.security.kerberos.KerberosLoginModule
+# ...      
+----
++
+System properties are passed to the JVM using the standard `-D` option.
```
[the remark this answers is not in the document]
PaulRMellor: removed

documentation/modules/deploying/con-securing-kafka-connect-api.adoc
PaulRMellor: How about this:

> For example, suppose an organization uses a Kafka Connect cluster and connectors to stream sensitive data from a customer database to a central database. 
> The administrator uses a configuration provider plugin to store store sensitive information related to connecting to the customer database and the central database, such as database connection details and authentication credentials.  
> The configuration provider protects this sensitive information from being exposed to unauthorized users. 
> However, someone who has access to the Kafka Connect API can still obtain access to the customer database without the consent of the administrator.
> They can do this by setting up a fake database and configuring a connector to connect to it. 
> They then modify the connector configuration to point to the customer database, but instead of sending the data to the central database, they send it to the fake database.
> By configuring the connector to connect to the fake database, the login details and credentials for connecting to the customer database are intercepted, even though they are stored securely in the configuration provider.

documentation/modules/deploying/con-securing-kafka-connect-api.adoc
```diff
@@ -0,0 +1,73 @@
...
+//
+// assembly-deploy-kafka-connect-with-plugins.adoc
+
+[id='con-securing-kafka-connect-api-{context}']
+= Limiting access to the Kafka Connect API
+
+[role="_abstract"]
+To grant a user access to the Kafka Connect API, a Kafka cluster admin must provide them with the API URL. 
+This includes the hostname or IP address of the Kafka Connect cluster and the port number on which the API is running (by default, port 8083). 
+Additionally, the user must have authenticated access to the Kubernetes cluster and the Kafka Connect REST API endpoint in order to make a request.
```
tombentley, line 11: The subject is _limiting_ access but we're starting out with how to _give_ access, which is a little weird. 

I think we should start with the "Why?" (which is currently the 2nd paragraph), and then go on to the "How?" (which is like this paragraph, but stated slightly differently):

```suggestion
Anyone with authenticated access to the Kubernetes cluster and knowledge of the Kafka Connect REST API endpoint (URL including hostname/IP address and port number) can use the REST API.
```

Then the example shows how someone with such access is able to use their ability to create a connector to recover a secret which they'd not otherwise have access to.

mimaison (approved):
LGTM

documentation/modules/deploying/con-securing-kafka-connect-api.adoc
PaulRMellor: Thanks @tombentley. I've done the switch and changed to a new second paragraph as follows:

> The Kafka Connect REST API can be accessed by anyone who has authenticated access to the Kubernetes cluster and knows the endpoint URL, which includes the hostname/IP address and port number.

tombentley (approved):
LGTM, thanks Paul!

PR #8118, opened 2023-02-17
Add REST API to KafkaAgent that exposes broker metrics

tinaselenge:
### Type of change
- Enhancement / new feature

### Description

This PR implements [Strimzi Proposal 048](https://github.com/strimzi/proposals/blob/main/048-avoid-broker-restarts-when-in-recovery.md). It adds REST API to KafkaAgent that exposes broker metrics. The KafkaRoller would use this API to determine the broker state and log recovery progress if it's in RECOVERY state. This would allow KafkaRoller to make better decision, when rolling brokers. 

In the future, the  API could also be extended with internal non-TLS and used by the readiness/liveness checks.
### Checklist

- [ ] Write tests
- [ ] Update documentation
- [ ] Update CHANGELOG.md


kafka-agent/src/main/java/io/strimzi/kafka/agent/KafkaAgent.java
```diff
@@ -4,28 +4,59 @@
...
+ *          "remainingLogsToRecover": 456
+ *        }
+ *      }</dd>
+ * </dl>
  */
 public class KafkaAgent {
     private static final Logger LOGGER = LoggerFactory.getLogger(KafkaAgent.class);
+    private static final String BROKER_STATE_PATH = "/v1/broker-state";
+    private static final int HTTPS_PORT = 8443;
+    private static final long GRACEFUL_SHUTDOWN_TIMEOUT_MS = 60 * 1000;
```
scholzj, line 59: The default graceful shutdown period on Kubernetes is 30 seconds. That is what most users will be using. So I think it makes little sense to have this set to anything higher than that.

kafka-agent/src/main/java/io/strimzi/kafka/agent/KafkaAgent.java
```diff
@@ -214,21 +345,32 @@ private void touch(File file) throws IOException {
...
             LOGGER.error("Unable to parse arguments {}", agentArgs);
             System.exit(1);
         } else {
-            File brokerReadyFile = new File(agentArgs.substring(0, index));
-            File sessionConnectedFile = new File(agentArgs.substring(index + 1));
+            File brokerReadyFile = new File(args[0]);
+            File sessionConnectedFile = new File(args[1]);
+            String sslKeyStorePath = args[2];
+            String certStorePass = args[3];
+            String sslTrustStorePath = args[4];
```
scholzj, line 357: It would probably make more sense to me to have a separate password parameter for keystore and truststore.

scholzj (commented):
I left some nits. Maybe you can also clarify in the description that it implementes only a part of the proposal.

kafka-agent/src/main/java/io/strimzi/kafka/agent/KafkaAgent.java
```diff
@@ -135,6 +187,85 @@ private MetricsRegistry metricsRegistry()   {
...
+        Handler handler = new AbstractHandler() {
+            @Override
+            public void handle(String s, Request baseRequest, HttpServletRequest request, HttpServletResponse response) throws IOException {
+                response.setContentType("application/json");
+                response.setCharacterEncoding("UTF-8");
+                baseRequest.setHandled(true);
+
+                Map<String, Object> brokerStateResponse = new HashMap<>();
+                if (brokerState != null) {
+                    if ((byte) brokerState.value() == 2 && remainingLogsToRecover != null && remainingSegmentsToRecover != null) {
```
ppatierno, line 227: can't we have a constant instead of 2, as we have for `BROKER_RUNNING_STATE` and `BROKER_UNKNOWN_STATE`?

kafka-agent/src/main/java/io/strimzi/kafka/agent/KafkaAgent.java
```diff
@@ -214,21 +349,36 @@ private void touch(File file) throws IOException {
...
+            String sslKeyStorePass = args[3];
+            String sslTrustStorePath = args[4];
+            String sslTrustStorePass = args[5];
             if (brokerReadyFile.exists() && !brokerReadyFile.delete()) {
                 LOGGER.error("Broker readiness file already exists and could not be deleted: {}", brokerReadyFile);
                 System.exit(1);
             } else if (sessionConnectedFile.exists() && !sessionConnectedFile.delete()) {
                 LOGGER.error("Session connected file already exists and could not be deleted: {}", sessionConnectedFile);
                 System.exit(1);
+            } else if (sslKeyStorePath.equals("") || sslTrustStorePath.equals("")) {
```
showuon, line 369: nit: can be replaced with `isEmpty` or `isBlank`

kafka-agent/src/main/java/io/strimzi/kafka/agent/KafkaAgent.java
```diff
@@ -135,6 +191,85 @@ private MetricsRegistry metricsRegistry()   {
...
+
+        sslContextFactory.setKeyStorePath(sslKeyStorePath);
+        sslContextFactory.setKeyStorePassword(sslKeyStorePassword);
+        sslContextFactory.setKeyManagerPassword(sslKeyStorePassword);
+
+        sslContextFactory.setTrustStorePath(sslTruststorePath);
+        sslContextFactory.setTrustStorePassword(sslTruststorePassword);
+        sslContextFactory.setNeedClientAuth(true);
+
+        return  sslContextFactory;
```
showuon, line 270: nit: additional space

kafka-agent/src/main/java/io/strimzi/kafka/agent/KafkaAgent.java
```diff
@@ -4,56 +4,113 @@
...
+ * <dl>
+ *     <dt>{@code GET /v1/broker-state}</dt>
+ *     <dd>Reflects the BrokerState metric, returning a JSON response e.g. {"brokerState": 3}.
+ *      If broker state is RECOVERY(2), it includes remainingLogsToRecover and remainingLogsToRecover in the response e.g.
+ *      {"brokerState": 2
+ *       "recovery": {
+ *          "remainingLogsToRecover": 123
+ *          "remainingLogsToRecover": 456
+ *        }
+ *      }</dd>
```
showuon, line 52: nit: if it's a json type, there should be comma "," between each element, right?

scholzj (commented):
I wonder if this can be split into more files to make it more moduler and possible also make it possible to add some tests. E.g.:
* Have a separate class for all the Jetty stuff which could probably have some unit test
* Possible have a separation of how we get the Kafka info to allow us to mock it and have some unit tests for it as well. 

It is just something what came to my mind when going through the code - but maybe it is for some reason not possible or too complicated? WDYT?

kafka-agent/src/main/java/io/strimzi/kafka/agent/KafkaAgent.java
```diff
@@ -214,21 +349,36 @@ private void touch(File file) throws IOException {
      */
     public static void premain(String agentArgs) {
         int index = agentArgs.indexOf(':');
+        String[] args = agentArgs.split(":");
         if (index == -1) {
```
robobario, line 353: nit: we could remove `index` and change this block to test the length of args to make a better exception than ArrayIndexOutOfBounds if we have too few args.

kafka-agent/src/main/java/io/strimzi/kafka/agent/KafkaAgent.java
robobario: should be `remainingSegmentsToRecover` in both places in the comment also

tinaselenge:
> I wonder if this can be split into more files to make it more moduler and possible also make it possible to add some tests. E.g.:
> 
> * Have a separate class for all the Jetty stuff which could probably have some unit test
> * Possible have a separation of how we get the Kafka info to allow us to mock it and have some unit tests for it as well.
> 
> It is just something what came to my mind when going through the code - but maybe it is for some reason not possible or too complicated? WDYT?

Initially I had the handler in a separate class but I couldn't pass the references to recovery metrics to it as they are not initialised at the time of the starting the server. Also, in the future we will probably remove the liveness/readiness logic from KafkaAgent and use the endpoint directly instead, so then the class would be just running Jetty server and expose metrics. It would be good to add some unit tests though, I'm not quite sure how at the moment but maybe we could try adding some tests without separating the class in another PR?

scholzj (approved):

kafka-agent/src/main/java/io/strimzi/kafka/agent/KafkaAgent.java
```diff
@@ -65,28 +122,27 @@ public void onMetricRemoved(MetricName metricName) {
...
+                    brokerStateName = metricName;
+                    brokerState = (Gauge) metric;
+                } else if (isRemainingLogsToRecover(metricName) && metric instanceof Gauge) {
+                    LOGGER.debug("Metric {} added ", metricName);
+                    remainingLogsToRecover = (Gauge) metric;
+                } else if (isRemainingSegmentsToRecover(metricName) && metric instanceof Gauge) {
+                    LOGGER.debug("Metric {} added ", metricName);
+                    remainingSegmentsToRecover = (Gauge) metric;
+                } else if ("SessionExpireListener".equals(metricName.getType())
+                        && "SessionState".equals(metricName.getName())
```
tombentley, line 136: For consistency we could factor out a `isSessionState` method.

kafka-agent/src/main/java/io/strimzi/kafka/agent/KafkaAgent.java
```diff
@@ -135,6 +191,84 @@ private MetricsRegistry metricsRegistry()   {
...
+                    } else {
+                        brokerStateResponse.put("brokerState", brokerState.value());
+                    }
+
+                    response.setStatus(HttpServletResponse.SC_OK);
+                    String json = new ObjectMapper().writeValueAsString(brokerStateResponse);
+                    response.getWriter().print(json);
+                } else {
+                    response.setStatus(HttpServletResponse.SC_SERVICE_UNAVAILABLE);
+                    response.getWriter().print("Broker state metric not found");
```
tombentley, line 246: It seems like 404 would be a better fit?

tombentley (commented):
I think we need to find a way to test this, even if by creating metrics in the metrics registry by hand.

kafka-agent/pom.xml
```diff
@@ -38,6 +38,44 @@
...
+        <dependency>
+            <groupId>junit</groupId>
+            <artifactId>junit</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.httpcomponents</groupId>
+            <artifactId>httpclient</artifactId>
+            <version>4.5.13</version>
+            <scope>test</scope>
+        </dependency>
```
scholzj, line 74: Do we really need this as a dependency? Wouldn't the JDK HTTP client be enough?

tinaselenge: You are right, we don't need this dependency. I have removed it in the last commit.

tinaselenge:
> I think we need to find a way to test this, even if by creating metrics in the metrics registry by hand.

Hi @tombentley, I have added some tests for the handler. Do you think they are sufficient or there should be more scenarios tested?

scholzj:
@tombentley @ppatierno @robobario @showuon Any more comment on this? Now when we are starting a new release cycle would be a great time to get something like this over the line and merge it!

kafka-agent/src/main/java/io/strimzi/kafka/agent/KafkaAgent.java
```diff
@@ -65,28 +136,26 @@ public void onMetricRemoved(MetricName metricName) {
...
+                    LOGGER.debug("Metric {} added ", metricName);
+                    brokerStateName = metricName;
+                    brokerState = (Gauge) metric;
+                } else if (isRemainingLogsToRecover(metricName) && metric instanceof Gauge) {
+                    LOGGER.debug("Metric {} added ", metricName);
+                    remainingLogsToRecover = (Gauge) metric;
+                } else if (isRemainingSegmentsToRecover(metricName) && metric instanceof Gauge) {
+                    LOGGER.debug("Metric {} added ", metricName);
+                    remainingSegmentsToRecover = (Gauge) metric;
+                } else if (isSessionState(metricName)
```
ppatierno, line 149: do we need all the above `LOGGER.debug` to log a metric was added? I see a `LOGGER.trace("Metric added {}", metricName);` at the top which is logging for all of them.

ppatierno (approved):
LGTM. I left just one nit.

kafka-agent/src/main/java/io/strimzi/kafka/agent/KafkaAgent.java
tinaselenge: I guess it was for getting different level of loggings, trace being more verbose than debug which is just logging broker metrics. I'm happy to remove it anyway, as metrics get logged either way.

ppatierno: We could change the common one to be `debug` then. Not sure why it was trace.

kafka-agent/src/test/java/io.strimzi.kafka.agent/KafkaAgentTest.java
```diff
@@ -0,0 +1,109 @@
+/*
+ * Copyright Strimzi authors.
+ * License: Apache License 2.0 (see the file LICENSE or http://apache.org/licenses/LICENSE-2.0.html).
+ */
+package io.strimzi.kafka.agent;
```
tombentley, line 5: This source file appears to be in a directory called `io.strimzi.kafka.agent`, rather than nested directories `io/strimzi/kafka/agent`, which would be the norm for Java.

kafka-agent/src/test/java/io.strimzi.kafka.agent/KafkaAgentTest.java
```diff
@@ -0,0 +1,109 @@
...
+    public void testBrokerMetricNotFound() throws Exception {
+        KafkaAgent agent = new KafkaAgent(null, null, null);
+        context.setHandler(agent.getServerHandler());
+        server.setHandler(context);
+        server.start();
+
+        HttpResponse<String> response = HttpClient.newBuilder()
+                .build()
+                .send(req, HttpResponse.BodyHandlers.ofString());
+        assertEquals(response.statusCode(), HttpServletResponse.SC_NOT_FOUND);
```
tombentley, line 105: The expected value should be the first argument, and the actual value the 2nd. It doesn't matter when the assertion succeeds, but it results in a confusing message if the assertion fails.

tombentley (approved):
A couple of nits, otherwise LGTM. Thanks @tinaselenge.

showuon:
Sorry, @scholzj , I found some issues while running tests under KRaft mode. It looks like some more test update is needed since it is expected to have topic operator and using topic operator to create topic... etc. I'll ping you when it's ready to re-run. Thanks.

scholzj:
> Sorry, @scholzj , I found some issues while running tests under KRaft mode. It looks like some more test update is needed since it is expected to have topic operator and using topic operator to create topic... etc. I'll ping you when it's ready to re-run. Thanks.

@showuon I guess this was intended for #8247?

showuon:
> > Sorry, @scholzj , I found some issues while running tests under KRaft mode. It looks like some more test update is needed since it is expected to have topic operator and using topic operator to create topic... etc. I'll ping you when it's ready to re-run. Thanks.
> 
> @showuon I guess this was intended for #8247?

Oops, wrong PR. Please ignore it.

PR #8147, opened 2023-02-23
docs(rebalancing): Moves content related to scaling and rebalancing

PaulRMellor:
### Type of change

**Documentation**
Content from the Configuring guide is being moved to the Deploying guide.
The Configuring guide is being positioned as a pure schema reference guide.
Instructions to deploy, configure and administer Strimzi will sit together in the Deploying guide.

This PR moves content related to using scaling and rebalancing, including Cruise Control.
The info has been split three ways, so that there is a section for scaling brokers, cruise control and using the `kafka-reassign-partitions.sh tool`

Some rewrites to the intros of the reassign tool sections, so as to make it clearer what they are for. The section is no longer titled _Scaling clusters_, instead _Using the partition reassignment tool_. 

A new _Scaling Kafka clusters by adding or removing brokers_ section describes that config and also the best way to reassign partitions (Cruise Control) .

image

### Checklist

- [ ] Write tests
- [x] Update documentation
- [ ] Update CHANGELOG.md


documentation/modules/configuring/con-scaling-kafka-clusters.adoc
```diff
@@ -3,32 +3,44 @@
...
+----
+
+When you add add or remove brokers, Kafka does not automatically reassign partitions. 
+The best way to do this is using Cruise Control.
+You can use Cruise Control's `add-brokers` and `remove-brokers` modes when scaling a cluster up or down.
+
+* Use the `add-brokers` mode after scaling up a Kafka cluster to move partition replicas from existing brokers to the newly added brokers.
+* Use the `remove-brokers` mode before scaling down a Kafka cluster to move partition replicas off the brokers that are going to be removed.
+
+NOTE: When using a `StatefulSet` to manage Kafka brokers in a Kubernetes cluster, you cannot specify which specific pod to remove from the cluster. Instead, the broker removal process starts from the highest numbered pod in the `StatefulSet`.
```
ppatierno, line 46: I guess that right now it's still valid with `StrimziPodSet` right @scholzj ? Should we specify it here as well?

ppatierno (approved):
LGTM. Just one nit.

documentation/modules/configuring/con-scaling-kafka-clusters.adoc
scholzj: Yes. This is valid for the time being even with StrimziPodSets.

documentation/assemblies/configuring/assembly-reassign-tool.adoc
```diff
@@ -0,0 +1,25 @@
...
+
+[id='assembly-reassign-tool-{context}']
+= Using the partition reassignment tool
+
+[role="_abstract"]
+When scaling a Kafka cluster, you may need to add or remove brokers and update the distribution of partitions or the replication factor of topics. 
+To update partitions and topics, you can use the `kafka-reassign-partitions.sh` tool.
+
+Changing the replication factor of a topic cannot be done using any other Kafka tools. 
+You must use `kafka-reassign-partitions.sh` to do this.
```
scholzj, line 13: I don't think this is true. You can do it using the Kafka Admin API for example. The point here should be more that it is not supported by our Cruise COntrol integration and in Topic Operator.

documentation/modules/configuring/con-scaling-kafka-clusters.adoc
```diff
@@ -3,32 +3,44 @@
...
+
+NOTE: To increase the throughput of a Kafka topic, you can increase the number of partitions for that topic. 
+This allows the load of the topic to be shared between different brokers in the cluster. 
+However, if every broker is constrained by a specific resource (such as I/O), adding more partitions will not increase the throughput. 
+In this case, you need to add more brokers to the cluster.
+
+Change the `Kafka.spec.kafka.replicas` configuration to add or reduce the number of brokers.
+In the following example, the number of brokers is set at 3.
+A replica factor of 3 in Kafka means that each partition of a topic is replicated across three brokers. 
+This provides fault tolerance and ensures that if one of the brokers fails, other brokers are still available.
```
scholzj, line 24: This is a bit confusing I think. Strictly speaking, it is correct. But the `Kafka.spec.kafka.replicas` does not have any direct impact on the actual replication factor. It is only a prerequisite (e.g. to use replication factor 3, you need at least 3 replicas).

I think we should remove this here or clarify this.

documentation/modules/configuring/con-scaling-kafka-clusters.adoc
scholzj: Already raised by Paolo. This note is good, just remove the reference to StatefulSets.

documentation/modules/configuring/proc-adding-volumes-to-jbod-storage.adoc
```diff
@@ -57,8 +57,5 @@ spec:
 kubectl apply -f _<kafka_configuration_file>_
 
 . Create new topics or reassign existing partitions to the new disks.
-
-[role="_additional-resources"]
-.Additional resources
-
-For more information about reassigning topics, see xref:con-partition-reassignment-{context}[].
++
+TIP: Cruise Control is an effective tool for reassigning partitions.
```
scholzj, line 61: Should we add that this needs the `interbroker` reassignment to be enabled?

documentation/modules/configuring/proc-removing-volumes-from-jbod-storage.adoc
```diff
@@ -21,6 +21,8 @@ IMPORTANT: To avoid data loss, you have to move all partitions before removing t
 
 . Reassign all partitions from the disks which are you going to remove.
 Any data in partitions still assigned to the disks which are going to be removed might be lost.
++
+TIP: Cruise Control is an effective tool for reassigning partitions.
```
scholzj, line 25: Actually, Cruise Control currently cannot help with this. This would be another situation where the reassign toll might need to be used.

documentation/modules/configuring/proc-scaling-down-a-kafka-cluster.adoc
```diff
@@ -4,12 +4,12 @@
 
 [id='proc-scaling-down-a-kafka-cluster-{context}']
 
-= Scaling down a Kafka cluster
+= Reassigning partitions before removing brokers
```
scholzj, line 7: Isn't think where we wanted to have Cruise control mentioned / linked?

documentation/modules/configuring/proc-scaling-up-a-kafka-cluster.adoc
```diff
@@ -4,12 +4,12 @@
 
 [id='proc-scaling-up-a-kafka-cluster-{context}']
 
-= Scaling up a Kafka cluster
+= Reassigning partitions after adding brokers
```
scholzj, line 7: Same as above. Didn't we wanted to have the Cruise cotnrol mentioned here?

documentation/assemblies/configuring/assembly-reassign-tool.adoc
PaulRMellor: Okay. I changed to:

> Neither the Strimzi Cruise Control integration nor the Topic Operator support changing the replication factor of a topic. However, you can change the replication factor of a topic using the `kafka-reassign-partitions.sh` tool.

documentation/modules/configuring/con-scaling-kafka-clusters.adoc
PaulRMellor: Assuming I understood what we want to cover, I've tried to clarify as follows:

Adjusting the `Kafka.spec.kafka.replicas` configuration affects the number of brokers in the cluster that act as replicas. 
The actual replication factor for topics is determined by settings for the `default.replication.factor` and `min.insync.replicas`, and the number of available brokers. 
For example, a replication factor of 3 means that each partition of a topic will be replicated across three brokers, ensuring fault tolerance in the event of a broker failure.

documentation/modules/configuring/proc-removing-volumes-from-jbod-storage.adoc
PaulRMellor: Right. I've updated the note

documentation/modules/configuring/proc-scaling-up-a-kafka-cluster.adoc
PaulRMellor: I've added a note

scholzj (approved):

mimaison (approved):

ShubhamRwt (approved):

PR #8175, opened 2023-03-01
Updated the Vertx and Snakeyaml dependency

ShubhamRwt:
### Type of change

_Select the type of your PR_

- Task

### Description

This PR updates the vertx dependency to 4.3.8 and also updates snakeyaml version to 2.0 to fix a CVE present in snakeyaml 1.33. 

### Checklist

- [ ] Write tests
- [ ] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -113,9 +113,9 @@
         <fasterxml.jackson-annotations.version>2.14.1</fasterxml.jackson-annotations.version>
         <!-- This is used to override CVE-2022-25857 and CVE-2022-41854 => should be removed once jackson-dataformat-yaml uses it out of the box -->
         <!-- https://github.com/strimzi/strimzi-kafka-operator/issues/7261 covers the removal of this override -->
-        <snakeyaml.version>1.33</snakeyaml.version>
-        <vertx.version>4.3.7</vertx.version>
-        <vertx-junit5.version>4.3.7</vertx-junit5.version>
+        <snakeyaml.version>2.0</snakeyaml.version>
+        <vertx.version>4.3.8</vertx.version>
+        <vertx-junit5.version>4.3.8</vertx-junit5.version>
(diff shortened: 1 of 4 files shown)


scholzj (approved):
LGTM if it works 🤔

scholzj (commented):
But it doesn't. :-/

ShubhamRwt:
To update the `snakeyaml` version it will require code changes and Iam not sure if it will be backwards compatible so should we leave the snakeyaml version as it is? Wdyt @ppatierno @scholzj.

fvaleri:
https://github.com/fabric8io/kubernetes-client/issues/4933

pom.xml
```diff
@@ -131,7 +131,7 @@
         <jetty.version>9.4.48.v20220622</jetty.version>
         <javax-servlet.version>3.1.0</javax-servlet.version>
         <strimzi-oauth.version>0.11.0</strimzi-oauth.version>
-        <netty.version>4.1.86.Final</netty.version>
+        <netty.version>4.1.89.Final</netty.version>
```
ppatierno, line 134: In the bridge I decided to set 4.1.87 because that's the version that Vert.x 4.38 brings to us in order to avoid having some Netty modules 4.1.87 (because of Vert.x) and other Netty modules on a different version (i.e. 4.1.89 in your case). Maybe it's better to stick with what Vert.x is bringing? @scholzj wdyt? Otherwise I will have to bump Netty on bridge as well to align with the operator at least.

scholzj: I think sticking with what VErt.x uses is a good idea.

ShubhamRwt: Okay, I will revert it to 4.1.87

ppatierno (approved):

scholzj (approved):

Issue #7663, opened 2022-11-21
Clean old features from the CRD generator


scholzj:
Strimzi currently supports only Kubernetes 1.19+. But the CRD generator seems to have lot of left overs from previous Kuberneets CRD versions. It seems to have logic to:
* Support `v1beta1` CRD API
* Support CRDs which do not handle different schema per version

and possibly some more. These should be removed to clean up the code as they are nto needed anymore.

PR #8180, opened 2023-03-01
Clean old features from the CRD generator

omjikush09:
### Type of change

- Refactoring

### Description

Clean old features from the CRD generator [#issue](https://github.com/strimzi/strimzi-kafka-operator/issues/7663)

### Checklist

- [ ] Write tests
- [ ] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/crd-generator/src/main/java/io/strimzi/crdgenerator/CrdGenerator.java
+++ b/crd-generator/src/main/java/io/strimzi/crdgenerator/CrdGenerator.java
@@ -433,12 +433,11 @@ class CrdGenerator {
         if (conversionStrategy instanceof NoneConversionStrategy) {
             conversion.put("strategy", "None");
         } else if (conversionStrategy instanceof WebhookConversionStrategy) {
-            boolean v1Beta1CrdApi = crdApiVersion.compareTo(V1) < 0;
             conversion.put("strategy", "Webhook");
             WebhookConversionStrategy webhookStrategy = (WebhookConversionStrategy) conversionStrategy;
-            ObjectNode webhook = conversion.putObject(v1Beta1CrdApi ? "webhookClientConfig" : "webhook");
-            (v1Beta1CrdApi ? conversion : webhook).putArray("conversionReviewVersions").add("v1").add("v1beta1");
-            ObjectNode webhookClientConfig = (v1Beta1CrdApi ? conversion : webhook).putObject(v1Beta1CrdApi ? "webhookClientConfig" : "clientConfig");
+            ObjectNode webhook = conversion.putObject("webhook");
+            webhook.putArray("conversionReviewVersions").add("v1");
+            ObjectNode webhookClientConfig = (webhook).putObject("clientConfig");
             webhookClientConfig.put("caBundle", webhookStrategy.caBundle);
             if (webhookStrategy.isUrl()) {
                 webhookClientConfig.put("url", webhookStrategy.url);
@@ -1240,7 +1239,7 @@ class CrdGenerator {
                 targetKubeVersions = KubeVersion.parseRange("1.11+");
             }
             if (crdApiVersion == null) {
-                crdApiVersion = ApiVersion.V1BETA1;
+                crdApiVersion = ApiVersion.V1;
             }
             if (conversionServiceName != null) {
                 conversionStrategy = new WebhookConversionStrategy(conversionServiceName, conversionServiceNamespace, conversionServicePath, conversionServicePort, conversionServiceCaBundle);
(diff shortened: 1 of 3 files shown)


omjikush09:
Hi, This PR can take some as I am learning about CRD and I have my exams.. :sob:

scholzj:
@omjikush09 This is not a critical issue, so no worries. I think the changes you did are good so far, but I think there is more stuff to remove. If you want, we can also consider doing it in smaller parts and remove it tep by step few lines in one Pr, few in another one etc.

scholzj (commented):
I think there are some more places which can be cleaned up - try to look for the `crdApiVersion.compareTo` => that seems to be a good identifier for things to either remove completely or just remove the `if`.

@tombentley WDYT ... should we remove the `crdApiVersion` completely as we have only `v1` now? There might be some `v2` version in the future. But we have no idea where it will be needed. So if we keep it without using it, it might not be anyway useful in the future.

omjikush09:
Hi, I tried to do but multiple test cases are failing. Will ping you on slack.

scholzj (commented):
The `CrdGeneratorTest.java` seems to have some additional tests which would need to be updated to use the v1 API and not the v1beta1 API. You also mentioned on Slack that the custom resource sued by the test is not applicable on Kubernetes - can you please provide more details?

omjikush09:
```The CustomResourceDefinition "examples.crdgenerator.strimzi.io" is invalid: 
* spec.validation.openAPIV3Schema.properties[alternatives].oneOf[0].items.type: Forbidden: must be empty to be structural
* spec.validation.openAPIV3Schema.properties[alternatives].oneOf[0].type: Forbidden: must be empty to be structural
* spec.validation.openAPIV3Schema.properties[alternatives].oneOf[1].type: Forbidden: must be empty to be structural
* spec.validation.openAPIV3Schema.properties[alternatives].oneOf[1].x-kubernetes-preserve-unknown-fields: Forbidden: must be false to be structural
* spec.validation.openAPIV3Schema.properties[alternatives].type: Required value: must not be empty for specified object fields
* spec.validation.openAPIV3Schema.properties[either]: Required value: because it is defined in spec.validation.openAPIV3Schema.oneOf[0].properties[either]
* spec.validation.openAPIV3Schema.properties[or]: Required value: because it is defined in spec.validation.openAPIV3Schema.oneOf[1].properties[or]
* spec.validation.openAPIV3Schema.properties[typedAlternatives].oneOf[0].properties[key2].type: Forbidden: must be empty to be structural
* spec.validation.openAPIV3Schema.properties[typedAlternatives].oneOf[0].type: Forbidden: must be empty to be structural
* spec.validation.openAPIV3Schema.properties[typedAlternatives].oneOf[1].properties[key1].type: Forbidden: must be empty to be structural
* spec.validation.openAPIV3Schema.properties[typedAlternatives].oneOf[1].type: Forbidden: must be empty to be structural
* spec.validation.openAPIV3Schema.properties[typedAlternatives].type: Required value: must not be empty for specified object fields
```
Error when applying simpleTest.yaml to minikube and this CRD is generated by test

scholzj:
I will try to have a look at it tomorrow. But I think these things are forbidden in the `v1` schema, so they can be removed.

scholzj:
So, the `alternatives` and `typedAlternatives` ... I think those should be removed. Including all the classes it references, the annotation, the code generating the spec and docs, etc. since that should not be used anymore in the current CRD spec. I'm not entirely sure what would the problem be with the `either` and `or` fields. They should be fine and supported I think. This is also used in the `KafkaMirrorMaker` CRD, so not sure why would it not work here.

To be honest, this issue is quite complicated. And I'm not sure how much more time we will spend if we go through this the way that you run into some issues and I try to investigate it a few days later. Unfortunately, I do not have a prepared list of things that need to be removed as I would need to go through this step by step and actually remove them to figure it out. So I'm afraid I'm not sure what is the best way to proceed here.

scholzj:
@omjikush09 So, should we close this? Or do you want to continue with it and try to proceed with removing the alternatives?

omjikush09:
@scholzj  I tried to remove but no luck. So I think I would rollback back my last commit then we can merge it.

scholzj:
I'm not sure if rollback really helps as it would be still incomplete. We
need to do this properly including the tests and old features etc.

omjikush09:
Hi, @scholzj  I know it will not be complete but as you have told we complete it in multiple parts so I thought we can do it that way.
Ya if you want you can close this PR and create a new PR when the error is resolved.

scholzj:
> Hi, @scholzj I know it will not be complete but as you have told we complete it in multiple parts so I thought we can do it that way.

We can do it in parts, but it would need to have some system. The first commit seems to remove things from the middle. Doing it in parts would be for example to have a PR which removes the alternatives part etc. -> starting from the outside and not from the middle.

So I think it is better to close it and I will try to have a more detailed look at it when I find more time. But anyway, thanks a lot for the effort you put into it!

PR #8213, opened 2023-03-09
Integrate OAuth 0.12.0 features

mstruk:
### Type of change

_Select the type of your PR_

- Bugfix
- Enhancement / new feature
- Documentation

### Description

Intregrate new OAuth features into CRDs config:
* `httpRetries`, `httpRetryPauseMillis` in `authentication` section for `oauth` type
* `httpRetries` in `authorization` section for `keycloak` type 
*  Fix `enableMetrics` in `authorization` section for `keycloak` type (had no effect)

### Checklist

- [ ] Write tests
- [ ] Update documentation
- [ ] Update CHANGELOG.md


api/src/main/java/io/strimzi/api/kafka/model/authentication/KafkaClientAuthenticationOAuth.java
```diff
@@ -38,6 +38,8 @@ public class KafkaClientAuthenticationOAuth extends KafkaClientAuthentication {
     private String tokenEndpointUri;
     private Integer connectTimeoutSeconds;
     private Integer readTimeoutSeconds;
+    private Integer httpRetries;
+    private Integer httpRetryPauseMillis;
```
scholzj, line 42: I think we would normally use the `Ms` suffix. I.e. `httpRetryPauseMs`. For example `expireAfterMs` in OPA authorization. The `ms` suffix is also used in Kafka configs.

api/src/main/java/io/strimzi/api/kafka/model/listener/KafkaListenerAuthenticationOAuth.java
```diff
@@ -66,6 +66,8 @@ public class KafkaListenerAuthenticationOAuth extends KafkaListenerAuthenticatio
     private String customClaimCheck;
     private Integer connectTimeoutSeconds;
     private Integer readTimeoutSeconds;
+    private Integer httpRetries;
+    private Integer httpRetryPauseMillis;
```
scholzj, line 70: Same as in the other class.

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaBrokerConfigurationBuilder.java
```diff
@@ -710,6 +716,7 @@ private void configureKeycloakAuthorization(String clusterName, KafkaAuthorizati
         addOption(writer, "strimzi.authorization.grants.refresh.pool.size", authorization.getGrantsRefreshPoolSize());
         addOption(writer, "strimzi.authorization.connect.timeout.seconds", authorization.getConnectTimeoutSeconds());
         addOption(writer, "strimzi.authorization.read.timeout.seconds", authorization.getReadTimeoutSeconds());
+        addOption(writer, "strimzi.authorization.http.retries", authorization.getHttpRetries());
```
scholzj, line 719: It is not really an issue ... more a curiosity ... why do the OAuth options have the null and >0 checks and the authorization options are just added? Also, what if this is `null` given the type seems to be an Integer?

scholzj (commented):
Left some nits ...

ppatierno (commented):
Nothing to add to what mentioned by Jakub.

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaBrokerConfigurationBuilder.java
mstruk: This is probably just an ommition on my part. It would make sense that checks for configuration that will result in failure if let though should be a reason for aborting at the operator level. I'll have a closer look at this.

api/src/main/java/io/strimzi/api/kafka/model/authentication/KafkaClientAuthenticationOAuth.java
mstruk: I didn't notice, but sure we can have Ms. Intuitively it's just automatic to me that if we have SomethingSeconds, and not SomethingS, then we also have SomethingMillis and not SomethingMs. But I can change it since it's established pattern.
I don't really want to change it in OAuth lib though.

scholzj: I'm not so concerned about the OAuth lib, so feel free to keep it there and change it here. I think the `seconds` versus `ms` is commonly used elsewhere as well ... such as in Kafka itself. I guess `Millis` is not clear, `Milliseconds` is too long, and using just `S` for seconds` would be confusing.

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaBrokerConfigurationBuilder.java
scholzj: Just to be clear, if it works fine, I do not have problem with it ... I was just wondering as you have the checks few lines above.

api/src/main/java/io/strimzi/api/kafka/model/KafkaAuthorizationKeycloak.java
```diff
@@ -171,6 +172,17 @@ public void setReadTimeoutSeconds(Integer readTimeoutSeconds) {
         this.readTimeoutSeconds = readTimeoutSeconds;
     }
 
+    @Description("The maximum number of retries to attempt if an initial request fails. If not set, the default is to not attempt any retries.")
```
PaulRMellor, line 175:
```suggestion
    @Description("The maximum number of retries to attempt if an initial HTTP request fails. If not set, the default is to not attempt any retries.")
```

api/src/main/java/io/strimzi/api/kafka/model/authentication/KafkaClientAuthenticationOAuth.java
```diff
@@ -117,6 +119,26 @@ public void setReadTimeoutSeconds(Integer readTimeoutSeconds) {
         this.readTimeoutSeconds = readTimeoutSeconds;
     }
 
+    @Description("The maximum number of retries to attempt if an initial request fails. If not set, the default is to not attempt any retries.")
```
PaulRMellor, line 122:
```suggestion
    @Description("The maximum number of retries to attempt if an initial HTTP request fails. If not set, the default is to not attempt any retries.")
```

api/src/main/java/io/strimzi/api/kafka/model/authentication/KafkaClientAuthenticationOAuth.java
```diff
@@ -117,6 +119,26 @@ public void setReadTimeoutSeconds(Integer readTimeoutSeconds) {
...
+    @JsonInclude(JsonInclude.Include.NON_NULL)
+    public Integer getHttpRetries() {
+        return httpRetries;
+    }
+
+    public void setHttpRetries(Integer httpRetries) {
+        this.httpRetries = httpRetries;
+    }
+
+    @Description("The pause to take before retrying a failed request. If not set, the default is to not pause at all but to immediately repeat a request.")
```
PaulRMellor, line 132:
```suggestion
    @Description("The pause to take before retrying a failed HTTP request. If not set, the default is to not pause at all but to immediately repeat a request.")
```

api/src/main/java/io/strimzi/api/kafka/model/listener/KafkaListenerAuthenticationOAuth.java
```diff
@@ -163,6 +165,26 @@ public void setReadTimeoutSeconds(Integer readTimeoutSeconds) {
         this.readTimeoutSeconds = readTimeoutSeconds;
     }
 
+    @Description("The maximum number of retries to attempt if an initial request fails. If not set, the default is to not attempt any retries.")
```
PaulRMellor, line 168:
```suggestion
    @Description("The maximum number of retries to attempt if an initial HTTP request fails. If not set, the default is to not attempt any retries.")
```

api/src/main/java/io/strimzi/api/kafka/model/listener/KafkaListenerAuthenticationOAuth.java
```diff
@@ -163,6 +165,26 @@ public void setReadTimeoutSeconds(Integer readTimeoutSeconds) {
...
+    @JsonInclude(JsonInclude.Include.NON_NULL)
+    public Integer getHttpRetries() {
+        return httpRetries;
+    }
+
+    public void setHttpRetries(Integer httpRetries) {
+        this.httpRetries = httpRetries;
+    }
+
+    @Description("The pause to take before retrying a failed request. If not set, the default is to not pause at all but to immediately repeat a request.")
```
PaulRMellor, line 178:
```suggestion
    @Description("The pause to take before retrying a failed HTTP request. If not set, the default is to not pause at all but to immediately repeat a request.")
```

documentation/modules/oauth/proc-oauth-authentication-broker-config.adoc
```diff
@@ -146,8 +148,10 @@ For production, always use `https://` urls.
...
 <13> The connect timeout in seconds when connecting to the authorization server. The default value is 60.
 <14> The read timeout in seconds when connecting to the authorization server. The default value is 60.
-<15> A JsonPath query used to extract groups information from JWT token or introspection endpoint response. Not set by default. This can be used by a custom authorizer to make authorization decisions based on user groups.
-<16> A delimiter used to parse groups information when returned as a single delimited string. The default value is ',' (comma).
+<15> The maximum number of retries to perform if the request to the authorization server fails in unexpected way. The default value is 0 (no retries). When using this option consider reducing the timeout times for `connectTimeoutSeconds` and `readTimeoutSeconds`. Note that retries prevent the current worker thread from being available to other requests. Many of them stalling could make the Kafka broker unresponsive.
```
PaulRMellor, line 151:
```suggestion
<15> The maximum number of times to retry a failed HTTP request to the authorization server. The default value is `0`, meaning that no retries are performed. To use this option effectively, consider reducing the timeout times for the `connectTimeoutSeconds` and `readTimeoutSeconds` options. However, note that retries may prevent the current worker thread from being available to other requests, and if too many requests stall, it could make the Kafka broker unresponsive.
```

documentation/modules/oauth/proc-oauth-authentication-broker-config.adoc
```diff
@@ -146,8 +148,10 @@ For production, always use `https://` urls.
...
+<16> The time to pause before attempting another retry. By default this time is zero (no pause) as many issues are per-request network glitches or proxy glitches, but if your authorization server is under stress you may want to set this to 100 ms or more.
```
PaulRMellor, line 152:
```suggestion
<16> The time to wait before attempting another retry of a failed HTTP request to the authorization server. By default, this time is set to zero, meaning that no pause is applied. This is because many issues that cause failed requests are per-request network glitches or proxy issues that can be resolved quickly. However, if your authorization server is under stress or experiencing high traffic, you may want to set this option to a value of 100 ms or more to reduce the load on the server and increase the likelihood of successful retries.
```

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaBrokerConfigurationBuilder.java
mstruk: Oh I see what you mean. Yeah, this part is not necessarily consistent. 

The easiest would be to perform no checks and to let all values through without any validation.
The minus of that is that users can't look at operator log for configuration errors but have to look in Kafka log and decipher how oauth.* config properties convert back to listener / authorizer CRD config names.

So it's good to have validation at the Operator level. Then next thing is trying to decide between raising an error for every obviously nonsensical config value (which may sometimes be deliberately done during trying out different things), and interpreting a nonsensical value as something to silently ignore (and not set the value at all). The same choice happens inside OAuth when interpreting the configuration. It ignores some value (but logs a warning) and you always can see the actual configuration values used in the log. Of course it depends on the specific configuration option if it can be interpreted clearly. For example if only positive integer makes sense then do we raise an error for negative or zero value or just ignore it and use zero default.

For some options (those a few lines above) I opted in the past to ignore the bad value, letting the OAuth use the default.

documentation/modules/oauth/proc-oauth-authentication-broker-config.adoc
```diff
@@ -146,8 +148,10 @@ For production, always use `https://` urls.
...
+<17> A JsonPath query used to extract groups information from JWT token or introspection endpoint response. Not set by default. This can be used by a custom authorizer to make authorization decisions based on user groups.
```
PaulRMellor, line 153:
```suggestion
<17> A JsonPath query that is used to extract groups information from either the JWT token or the introspection endpoint response. This option is not set by default. By configuring this option, a custom authorizer can make authorization decisions based on user groups.
```

documentation/modules/oauth/proc-oauth-authentication-broker-config.adoc
```diff
@@ -146,8 +148,10 @@ For production, always use `https://` urls.
...
+<18> A delimiter used to parse groups information when returned as a single delimited string. The default value is ',' (comma).
```
PaulRMellor, line 154:
```suggestion
<18> A delimiter used to parse groups information when it is returned as a single delimited string. The default value is ',' (comma).
```

documentation/modules/oauth/proc-oauth-authorization-broker-config.adoc
```diff
@@ -83,6 +84,7 @@ Default is `false`.
 <9> (Optional) The number of threads to use to refresh (in parallel) the grants for the active sessions. The default value is 5.
 <10> (Optional) The connect timeout in seconds when connecting to the Keycloak token endpoint. The default value is 60.
 <11> (Optional) The read timeout in seconds when connecting to the Keycloak token endpoint. The default value is 60.
+<12> (Optional) The maximum number of retries to perform (without pausing) if the initial session grants request to the authorization server fails in unexpected way. The default value is 0 (no retries). When using this option consider reducing the timeout times for `connectTimeoutSeconds` and `readTimeoutSeconds`. Note that retries prevent the current worker thread from being available to other requests. Many of them stalling could make the Kafka broker unresponsive.
```
PaulRMellor, line 87:
```suggestion
<12> (Optional) The maximum number of times to retry (without pausing) a failed HTTP request to the authorization server. The default value is `0`, meaning that no retries are performed. To use this option effectively, consider reducing the timeout times for the `connectTimeoutSeconds` and `readTimeoutSeconds` options. However, note that retries may prevent the current worker thread from being available to other requests, and if too many requests stall, it could make the Kafka broker unresponsive.
```

documentation/modules/oauth/proc-oauth-kafka-config.adoc
```diff
@@ -109,6 +111,8 @@ An authorization server may require a client to specify the audience.
 In this case it is `kafka`.
 <6> (Optional) The connect timeout in seconds when connecting to the authorization server. The default value is 60.
 <7> (Optional) The read timeout in seconds when connecting to the authorization server. The default value is 60.
+<8> (Optional) The maximum number of retries to perform if the request to the authorization server fails in unexpected way. The default value is 0 (no retries). When using this option consider reducing the timeout times for `connectTimeoutSeconds` and `readTimeoutSeconds`. Note that retries prevent the current worker thread from being available to other requests. Many of them stalling could make the Kafka Bridge unresponsive.
```
PaulRMellor, line 114:
```suggestion
<8> (Optional) The maximum number of times to retry a failed HTTP request to the authorization server. The default value is `0`, meaning that no retries are performed. To use this option effectively, consider reducing the timeout times for the `connectTimeoutSeconds` and `readTimeoutSeconds` options. However, note that retries may prevent the current worker thread from being available to other requests, and if too many requests stall, it could make the Kafka broker unresponsive.
```

documentation/modules/oauth/proc-oauth-kafka-config.adoc
```diff
@@ -109,6 +111,8 @@ An authorization server may require a client to specify the audience.
...
+<9> (Optional) The time to pause before attempting another retry. By default this time is zero (no pause) as many issues are per-request network glitches or proxy glitches, but if your authorization server is under stress you may want to set this to 100 ms or more.
```
PaulRMellor, line 115:
```suggestion
<9> (Optional) The time to wait before attempting another retry of a failed HTTP request to the authorization server. By default, this time is set to zero, meaning that no pause is applied. This is because many issues that cause failed requests are per-request network glitches or proxy issues that can be resolved quickly. However, if your authorization server is under stress or experiencing high traffic, you may want to set this option to a value of 100 ms or more to reduce the load on the server and increase the likelihood of successful retries.
```

PaulRMellor (approved):
Great! I've made a few suggestions to the doc.

mstruk:
Thanks @PaulRMellor !

scholzj (commented):
LGTM. Can you please update it to the released version? (remove the staging repos)

ppatierno (approved):

scholzj (approved):

PR #8216, opened 2023-03-10
docs(clients): updates the client upgrade strategies

PaulRMellor:
### Type of change

**Documentation**

Updates **Strategies for upgrading clients** to reflect the changes introduced in Kafka 3.0 or newer. It puts more emphasis on strategies relevant to upgrades from later versions of Kafka (post version 3.0).
Now includes only two strategies: consumer-firsta nd consumer-first using broker-down conversion at the topic level.

The updated section also includes a clarification on the use of the `message.format.version` property when upgrading from Kafka versions earlier than 3.0.0, and how it can be used at the topic-level or broker-level.

It also includes a reference to the conversions metrics that can be used in the context of down-conversion.

### Checklist

- [ ] Write tests
- [x] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/documentation/modules/upgrading/con-upgrade-strategies-for-upgrading-clients.adoc
+++ b/documentation/modules/upgrading/con-upgrade-strategies-for-upgrading-clients.adoc
@@ -7,7 +7,10 @@
 = Strategies for upgrading clients
 
 [role="_abstract"]
-The goal of upgrading Kafka clients is to enable them to consume new message formats produced by updated producers. 
+Upgrading Kafka clients ensures that they benefit from the features, fixes, and improvements that are introduced in new versions of Kafka. 
+Upgraded clients maintain compatibility with other upgraded Kafka components.
+The performance and stability of the clients might also be improved.
+
 The chosen upgrade strategy depends on whether you are upgrading brokers or clients first. 
 If you upgrade clients before brokers, some new features may not work as they are not yet supported by brokers. 
 However, brokers can handle producers and consumers running with different versions and supporting different log message versions.
--- a/documentation/modules/upgrading/con-upgrade-strategies-for-upgrading-clients.adoc
+++ b/documentation/modules/upgrading/con-upgrade-strategies-for-upgrading-clients.adoc
@@ -6,53 +6,31 @@
 
 = Strategies for upgrading clients
 
-The right approach to upgrading your client applications (including Kafka Connect connectors) depends on your particular circumstances.
+[role="_abstract"]
+The goal of upgrading clients is to enable them to consume new message formats produced by updated producers.
+You do this by upgrading all the consumers for a topic _before_ upgrading any of the producers.
+Upgrading consumers first ensures that consuming applications can understand the new message format before any new messages in that format are produced.
+The chosen upgrade strategy depends on a number of factors, such as the number of applications that need to be upgraded and how much downtime is tolerable. 
+You can combine more than one strategy.
 
-Consuming applications need to receive messages in a message format that they understand. You can ensure that this is the case in one of two ways:
+.Consumers first strategy
 
-* By upgrading all the consumers for a topic _before_ upgrading any of the producers.
-* By having the brokers down-convert messages to an older format.
+For the consumers first strategy, upgrade all the consuming applications, and then upgrade all the producing applications. 
 
-Using broker down-conversion puts extra load on the brokers, so it is not ideal to rely on down-conversion for all topics for a prolonged period of time.
-For brokers to perform optimally they should not be down converting messages at all.
+This strategy assumes that all consumers in your organization can be upgraded in a coordinated way, and it does not work for applications that are both consumers and producers.
+If there is a problem with the upgraded clients, there is a risk that messages in the new format might be added to the message log so that you cannot revert to the previous consumer version.
 
-Broker down-conversion is configured in two ways:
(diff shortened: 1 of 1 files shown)


documentation/modules/upgrading/con-upgrade-strategies-for-upgrading-clients.adoc
```diff
@@ -6,65 +6,43 @@
 
 = Strategies for upgrading clients
 
-The right approach to upgrading your client applications (including Kafka Connect connectors) depends on your particular circumstances.
+[role="_abstract"]
+The goal of upgrading clients is to enable them to consume new message formats produced by updated producers.
```
mimaison, line 10: To me, you upgrade clients to benefits from new features, fixes and improvements.

documentation/modules/upgrading/con-upgrade-strategies-for-upgrading-clients.adoc
```diff
@@ -6,65 +6,43 @@
...
+You do this by upgrading all the consumers for a topic _before_ upgrading any of the producers.
```
mimaison, line 11: Brokers are able to handle producers and consumers running with different versions and supporting different log message versions. 
The strategies are upgrading brokers or clients first:
- If you upgrade clients first there's a chance some new features may not work as they are not supported by brokers yet. 
- If you upgrade brokers first, in the past there was a chance client would only support an older message version and it would cause brokers to downconvert records or you would have to keep an older log message version. Since the log message version has not changed in years this has not been an issue for many versions.

documentation/modules/upgrading/con-upgrade-strategies-for-upgrading-clients.adoc
```diff
@@ -6,65 +6,43 @@
...
+It also works for applications that are both consumers and producers of the same topic. 
+The approach ensures that the upgraded producers and consumers are working correctly before you commit to using the new message format version.
 The main drawback of this approach is that it can be complicated to manage in a cluster with many topics and applications.
 
-NOTE: It is also possible to apply multiple strategies.
-For example, for the first few applications and topics the
-"per-topic consumers first, with down conversion" strategy can be used.
-When this has proved successful another, more efficient strategy can be considered acceptable to use instead.
+TIP: You can use Kafka metrics to monitor the performance of message down-conversion. 
+The _message conversion time_ metric measures the time taken to perform message conversion.
```
mimaison, line 47: Let's include the MBean name or a link to the upstream doc.

Maybe it's also worth mentioning the metrics that track client names and versions:
`kafka.server:type=socket-server-metrics,clientSoftwareName=<name>,clientSoftwareVersion=<version>,listener=<listener>,networkProcessor=<processor>`

documentation/modules/upgrading/con-upgrade-strategies-for-upgrading-clients.adoc
PaulRMellor: I've updated without saying too much about versions, just highlighting the potential drawbacks of both.

> The goal of upgrading Kafka clients is to enable them to consume new message formats produced by updated producers. 
> The chosen upgrade strategy depends on whether you are upgrading brokers or clients first. 
> If you upgrade clients before brokers, some new features may not work as they are not yet supported by brokers. 
> However, brokers can handle producers and consumers running with different versions and supporting different log message versions.
> 
> If you upgrade brokers first, there may be a need for down-conversion or retention of older log message versions to ensure compatibility with older applications. 
> The decision to upgrade clients or brokers first depends on several factors, such as the number of applications that need to be upgraded and how much downtime is tolerable.
> 
> Therefore, it is important to carefully consider the best approach for upgrading Kafka clients and brokers to ensure a smooth transition.

documentation/modules/upgrading/con-upgrade-strategies-for-upgrading-clients.adoc
PaulRMellor: Updated as follows:

````
[TIP]
====
The following Kafka broker metrics help monitor the performance of message down-conversion:

* `kafka.network:type=RequestMetrics,name=MessageConversionsTimeMs,request={Produce|Fetch}` provides metrics on the time taken to perform message conversion. 
* `kafka.server:type=BrokerTopicMetrics,name={Produce|Fetch}MessageConversionsPerSec,topic=([-.\w]+)` provides metrics on the number of messages converted over a period of time.  

You can track Kafka client names and versions using this metric:

* `kafka.server:type=socket-server-metrics,clientSoftwareName=<name>,clientSoftwareVersion=<version>,listener=<listener>,networkProcessor=<processor>`  
====
````

scholzj (approved):
LGTM. But I don't really understand it that much ... so approval from @mimaison is more important.

documentation/modules/upgrading/con-upgrade-strategies-for-upgrading-clients.adoc
PaulRMellor: Yeah. Let's get away from strategy here.
Updated as follows:

> Upgrading Kafka clients ensures that they benefit from the features, fixes, and improvements that are introduced in new versions of Kafka. 
>Upgraded clients maintain compatibility with other upgraded Kafka components.
> The performance and stability of the clients might also be improved.

mimaison (approved):

PR #8218, opened 2023-03-11
Refactor metrics and logging in model classes

scholzj:
### Type of change

- Refactoring

### Description

This PR refactors metrics and logging handling in the `AbstractModel` class and its subclasses. It creates 2 new classes: `MetricsModel` and `LoggingModel` which wraps the logic with the help of some static methods in utils classes. This allows us to free the metrics and logging from the `AbstractModel` and use it only in the subclasses where they are actually applicable. (This is similar to how it was already done for JMX)

It also creates a new utils class `MetricsAndLoggingUtils` which groups some static methods to help with the metrics and logging Config Map. While the metrics and logging are seemingly unrelated, they end up in the same ConfigMap for the operands. So these methods deal with both and are grouped together.

### Checklist

- [x] Write tests

### DIFF
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/model/logging/LoggingModel.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/model/logging/LoggingModel.java
@@ -10,7 +10,7 @@ import io.strimzi.api.kafka.model.Logging;
 import io.strimzi.operator.common.Reconciliation;
 
 /**
- * Represents a model for components with enabled JMX access
+ * Represents a model for logging configuration
  */
 public class LoggingModel {
     /**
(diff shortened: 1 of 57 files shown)


cluster-operator/src/main/java/io/strimzi/operator/cluster/model/logging/LoggingModel.java
```diff
@@ -0,0 +1,101 @@
...
+ */
+package io.strimzi.operator.cluster.model.logging;
+
+import io.fabric8.kubernetes.api.model.ConfigMap;
+import io.strimzi.api.kafka.model.HasConfigurableLogging;
+import io.strimzi.api.kafka.model.Logging;
+import io.strimzi.operator.common.Reconciliation;
+
+/**
+ * Represents a model for components with enabled JMX access
```
ppatierno, line 13: JMX?

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/logging/LoggingModel.java
```diff
@@ -0,0 +1,101 @@
...
+     */
+    public static final String LOG4J2_CONFIG_MAP_KEY = "log4j2.properties";
+
+    private final Logging logging;
+    private final String defaultLogConfigBaseName;
+    private final boolean isLog4j2; // Indicates whether Log4j2 is used => if set to false, uses Log4j1
+    private final boolean shouldPatchLoggerAppender;
+
+    /**
+     * Constructs the JMX Model for managing JMX access to Strimzi
```
ppatierno, line 32: again, JMX?

ppatierno (approved):
LGTM. A couple of nits on Javadoc.

PR #8224, opened 2023-03-13
Move JMX classes to a separate package

scholzj:
### Type of change

- Refactoring

### Description

The classes in the `cluster-operator` module, in the `io.strimzi.operator.cluster.model` package are currently mixing many different things together. That makes it sometimes quite hard to separate them, and have a clear idea which parts are genuinely public and used by other components and which are just used internally by a given class, its subclasses, and tests.

Already in the previous PR, we created separate packages for the logging and metrics classes. Thsi PR also refactors the recently added `JmxModel` class and the `SupportsJmx` interface into its own separate `jmx` package.

### Checklist


cluster-operator/src/main/java/io/strimzi/operator/cluster/model/jmx/JmxModel.java
```diff
@@ -26,15 +31,40 @@
...
+     */
+    public static final int JMX_PORT = 9999;
+
+    /**
+     * Key under which the JMX username is stored int he JMX Secret
+     */
+    public static final String JMX_USERNAME_KEY = "jmx-username";
+
+    /**
+     * Key under which the JMX password is stored int he JMX Secret
```
ppatierno, line 50: typo "int he"

ppatierno (approved):
LGTM. Just a typo in the Javadoc.

PR #8225, opened 2023-03-13
Adding KRaft related dashboard

ppatierno:
### Type of change

- Enhancement / new feature

### Description

This PR adds:

~~* a couple of panels on the Kafka dashboard with some KRaft related stuff.~~
* a new KRaft related dashboard.

Following more details.

~~#### Kafka dashboard: additional panels~~

image

~~Additional panels related to the brokers (acting as observers) fetching records from metadata topic (on the quorum controller).
The record offset metric is useful to check how much a broker is keeping the pace with the metadata offset on the leader (see raft-metrics log-end-offset). The corresponding is available on the KRaft dashboard (as log end offset but high-watermark as well). 
The lag in ms is used to process an average in order to know how much lag a broker has when the last record is applied by leader on the metadata topic and fetched by the broker itself.~~

#### KRaft dashboard

<no picture>

Usual memory usage, CPU usage, JVM data and more. It's even true that when in "combined mode" they are a duplicate of the same panels in the Kafka dashboard. We could also discuss about removing them until we'll be able to support not only "combined mode", so that these panels we'll show data about the KRaft quorum nodes.

image

Above data related to the KRaft quorum and metadata exchange across controllers.
The leader of the quorum appends NO_OP_RECORD(s) to the metadata topic continuously as for “showing” its health. The followers will get those records as well (together with all the other metadata related records). Leader should have a consistent append-records-rate (while fetch-records-rate will be 0). All followers should have a consistent fetch-records-rate (while append-records-rate will be 0). When a leader stops (or steps down as leader) and a follower is elected as new leader, its append-records-rate will start to increase until a steady rate while its fetch-records-rate will decrease until 0. When the old leader rejoins as a follower, the metrics will go the other way around, start fetching faster until a steady lower rate with no append. These metrics are in general useful to show if a leader is working fine appending records and followers are fetching and keeping the pace.

The commit latency avg metric is updated for the leader only. It seems to be NaN for all followers (not sure if this is how it's supposed to work). It can provide additional help on checking leader working status when committing records on the metadata topic. High latency could show disk related problems?

image

Above data related to the quorum formation and leader election.
It’s important to check all controllers in the quorum see the same as the current leader.
The current vote could be useful to understand how the quorum controller was elected based on the votes.
Also the current quorum epoch has to be aligned across the controllers in the quorum.

image

Above data related to high-watermark and log end offset for the metadata topic across the controllers.
Useful to get the status and the lag across the followers compared to the leader (and even across the observers, see panels in the Kafka dashboard).

image

Above some data around incoming/outgoing bytes and requests/responses rates on the metadata topic.
No documentation for them, I went through some investigation from the code.
These metrics are exposed by the KRaft client and they are mostly related to the corresponding network client used for communication between followers and leader.
They should show that metadata topic traffic is happening across the quorum controllers.
I would expect an incoming rate on followers and outgoing rate on the leader. What I noticed was incoming and outgoing for followers but nothing (0) for the leader (tbh I am not sure it's working as expected).
They show followers activity about requests/responses to the leader (mostly to fetch metadata records). Expect that for the leader these are 0.

### Checklist

- [ ] Update CHANGELOG.md


scholzj (commented):
I'm not sure I follow the change to the regular Kafka dashboard. What do these panels show in ZooKeeper based cluster?

scholzj (commented):
> Usual memory usage, CPU usage, JVM data and more

If the KRaft dashboard shows these things -> how do you separate the controller nodes from the broker nodes in the future?

showuon:
> What I noticed was incoming and outgoing for followers but nothing (0) for the leader (tbh I am not sure it's working as expected).

It did sound like a bug. I'll take a look.

packaging/examples/metrics/grafana-dashboards/strimzi-kraft.json
```diff
@@ -0,0 +1,2113 @@
...
+          "intervalFactor": 1,
+          "legendFormat": "{{kubernetes_pod_name}}",
+          "refId": "A"
+        }
+      ],
+      "thresholds": [],
+      "timeFrom": null,
+      "timeRegions": [],
+      "timeShift": null,
+      "title": "Metadata Topic Incoming Bytes Rate",
```
showuon, line 1438: I think this should be pure "incoming Bytes", not "incoming bytes rate", is that right? Same question to "outgoing bytes rate"

showuon:
> > What I noticed was incoming and outgoing for followers but nothing (0) for the leader (tbh I am not sure it's working as expected).
> 
> It did sound like a bug. I'll take a look.

I've checked, this works as designed. As you said, this is exposed from raft client, so it only has data while sending/receiving requests/responses via raft client. There will be many data reported from followers/leaders during quorum creation (i.e. voting). After leader elected, there will be only followers/observers sending fetch request via raft client, so that's why no more data reported in leader side.

packaging/examples/metrics/grafana-dashboards/strimzi-kraft.json
ppatierno: Actually the corresponding PromQL query is processing a rate starting from the "total" metrics across the last 60 secs.

```
sum(irate(kafka_server_raftchannelmetrics_incoming_byte_total{namespace="$kubernetes_namespace",strimzi_io_cluster="$strimzi_cluster_name",kubernetes_pod_name=~"$strimzi_cluster_name-$kafka_broker"}[1m])) by (kubernetes_pod_name)
```

ppatierno:
@showuon my understanding was that in the `RaftManager` (running on all controller nodes, leader and followers), a `KafkaRaftClient` instance is used and it's the one even handling inbound requests (i.e. the `FETCH` one) through [`handleRequest`](https://github.com/apache/kafka/blob/35142d43e6aa55bd11484224b8b09a83800d3a22/raft/src/main/java/org/apache/kafka/raft/KafkaRaftClient.java#L1613) not just making requests. It sounded to me it was working in a kind "peer-to-peer" fashion having the Raft client on both sides, leader and followers, using its own network client (where the metrics around bytes are accounted).
Of course, I could be wrong in my understanding.

ppatierno:
> If the KRaft dashboard shows these things -> how do you separate the controller nodes from the broker nodes in the future?

I had a quick thinking about it and was wondering if "tagging" metrics with the actual "process.roles" (broker or controller) would have been a good solution to this if possible to apply.

scholzj:
I do not have a strong opinion on whether long-term KRaft should be a separate dashboard or new charts in the Kafka dashboard. But if it should be one dashboard, I think I would not add it yet given our KRaft support, and would wait for later. Maybe a separate dashboard might be fine in the meantime and we can just copy-paste things later.

> I had a quick thinking about it and was wondering if "tagging" metrics with the actual "process.roles" (broker or controller) would have been a good solution to this if possible to apply.

I don't think you can do that easily. You would need to either parse and modify the whole JMX Prometheus Exporter mapping file or you would need to modify the PodMonitors to do some custom relabeling. That would be quite complicated given the variants (combined nodes etc.).

ppatierno:
> I do not have a strong opinion on whether long-term KRaft should be a separate dashboard or new charts in the Kafka dashboard. But if it should be one dashboard, I think I would not add it yet given our KRaft support, and would wait for later. Maybe a separate dashboard might be fine in the meantime and we can just copy-paste things later.

As I said the additional charts on the `Kafka` dashboard are there because strictly related to brokers acting as observers (for the metadata topic) but I am not sure now how much they are useful.
When in KRaft mode I think the `Kafka` dashboard should have info about brokers node, data traffic, usage and so on as today. The `KRaft` dashboard should be focused on the Raft quorum controllers side. When in "combined" mode the only overlapping between the two I see is about the CPU, memory and so on because we are actually plotting data for the same nodes.

tombentley (approved):

showuon (approved):
LGTM!

ppatierno:
@scholzj ...

Regarding the KRaft dashboard, it includes the main KRaft related metrics which are mostly about controller quorum, voting, leader, and traffic on the metadata topic as you can see in the dashboards on the PR. They make sense for both controllers and brokers (because anyway brokers, as observers, have metadata topic traffic). Broker anyway exposes votes (always -1 because they don't vote) and which is the leader.
I tested and it works for both combined and controllers +  brokers architectures, showing the metrics from all the nodes.
So KRaft metrics for controllers and brokers are on the same dashboard. Unless you want two KRaft dashboards, one for controllers only (like ZK ensamble) and one for brokers (but anyway exposing the same metrics). 
Of course, brokers will still have the Kafka dashboard (because it's not related to KRaft). 

In the Kafka dashboard, as I already mentioned before, I added a couple of graphs for KRaft metrics exposed only by brokers acting as observers (for the metadata topic) but I am not sure now how much they are useful. Maybe we can remove those for the time being and see if the ones in KRaft dashboard are enough.

scholzj:
I think the idea of two dashboards is fine. But without actually seeing them, I probably don't follow how will they work with different brokers and controllers and how do you filter them. For example the CPU and memory metrics you seem to have on both charts. 

> In the Kafka dashboard, as I already mentioned before, I added a couple of graphs for KRaft metrics exposed only by brokers acting as observers (for the metadata topic) but I am not sure now how much they are useful. Maybe we can remove those for the time being and see if the ones in KRaft dashboard are enough.

TBH, I have no idea how useful they would be. I'm more concerned about adding them there today given the current KRaft support state.

scholzj:
PS: Maybe having a screenshots covering how the dahsboards will look like with:
* Mixed controller and broker nodes
* Separate controller and broker nodes

Would help to understand how will they look like. To you have the charts somewhere deployed to make the screenshots?

ppatierno:
> TBH, I have no idea how useful they would be. I'm more concerned about adding them there today given the current KRaft support state.

I am fine removing them for now, I will do.

Regarding the KRaft dashboard, following two screenshots with combined and not-combined mode.
They look the same, just more nodes showing the same metrics.

Controllers and Brokers
kraft

Combined mode
kraft_combined_mode

Regarding the CPU and memory graphs to be in common with Kafka dashboard, it's something I already raised and I agree. The thing is that if we decide to have the KRaft dashboard only about controllers (so that CPU and memory is about controllers and not even brokers anymore), where we are putting the same metrics related to brokers? In the Kafka dashboard them? And when in combined mode?

ppatierno:
> Aren't most of these charts valid only for controllers? Things such as append or fetch metadata etc.? So why does it ist the brokers there?

No. Brokers are actually part of a Kafka KRaft based cluster as "observer"(s) so they are fetching metadata topic as all the "follower" controllers which are part of the quorum instead. So all the above metrics make sense for both controllers and brokers.
The only ones we could argue are "current quorum leader", "current voted leader" and "current quorum epoch" that we could think to be related just to the controllers quorum. The thing is that anyway all of them are exposed by brokers as well because they know the leader, they know the quorum epoch and because they don't vote, the last metric reports always -1.

So considering brokers, in my mind the two dashboards mean:
- Kafka: all "data" related metrics for brokers in terms of topics, partitions, logs, traffic ... as we have today
- KRaft: all "kraft" related metrics for brokers, so mostly requests and traffic on the metadata topic

> Do we also need the general resource usage (Memory, CPU, JVM Memory, GC charts etc.) information there if that is already on the regular Kafka dashboard?

They are in the KRaft dashboard because they should show you info about controllers. Right now they are showing info about brokers as well (duplicating what we have in Kafka, where controllers aren't shown). Maybe we should filter controllers only here? but what when combined mode? In the end the same info will be in the Kafka dashboard as well.

scholzj:
> They are in the KRaft dashboard because they should show you info about controllers. Right now they are showing info about brokers as well (duplicating what we have in Kafka, where controllers aren't shown). Maybe we should filter controllers only here? but what when combined mode? In the end the same info will be in the Kafka dashboard as well.

I guess showing mixed nodes on both make some sense. Or maybe we should consider separating them into a third dashboard? But then the correlation between events in the charts will not be so easy to notice. (e.g. there is a spike in some chart when GC kicks in etc.)

scholzj:
TBH, it's not like I have some amazing ideas on how to do it differently. So I think my main concern is that I would not add the KRaft metrics to the existing dashbaord just yet. Otherwise, I would be fine merging it and we might do more changes in the future if needed.

scholzj (approved):

PR #8249, opened 2023-03-17
Avoid pulling multiple SLF4J versions into Kafka container images

scholzj:
### Type of change

- Bugfix

### Description

This PR updates exclusion on OAuth to exclude the SLF4J API which caused two different versions being present in the container images. It also sets the exclusion to the Kubernetes Config Provider, but that is just for the future as that actually uses the same version as Kafka does and does not pull any new dependencies (so the exclusion is not really needed right now).

### Checklist


fvaleri (approved):
LGTM.

ppatierno (approved):

PR #8253, opened 2023-03-18
Remove JMX Trans support

### LINKED ISSUE #7692: Remove JMX Trans deployment logic
Based on [SP#43](https://github.com/strimzi/proposals/blob/main/043-deprecate-and-remove-jmxtrans.md), we should remove JMX Trans in Strimzi 0.35. This should include:
* Updating the API that JMX Trans is not removed (it has to stay in the API)
* Removing it from the docs
* Removing the container image
* Removing the deployment from the operator code

However, the operator should still check if the JMX Trans resources exist (after upgrade etc.) and delete them if needed (Deployment, services, Service Accounts, Config Maps etc.). These should be removed only in next phase in Strimzi 0.40 as part of #7691

**This issue should be done only for 0.35 as per the proposal!**

scholzj:
### Type of change

- Task

### Description

As proposed and approved on [SP#43](https://github.com/strimzi/proposals/blob/main/043-deprecate-and-remove-jmxtrans.md), this PR removes the support for JMX Trans from Strimzi. It leaves only minimal support required for uninstalling JMX Trans and the APIs which are deprecated and will be ignored, but have to stay present for backwards compatibility reasons.

This PR does following tasks:
* Removes the JMX Trans container image build
* Updates the Helm Chart and YAML installation files and removes JMX Trans from them
* Updates the `JmxTransReconciler` to delete the JMX Trans resources (Deployment, Config Map, Service Account) regardless whether `.spec.jmxTrans` is set or not
* Deprecates the JMX Trans methods from Pod Security Provider interface and removes the implementations in the existing plugins
* Removes JMX Trans from docs (leaves a notice about the removal and the API docs still remains but is marked as deprecated)
* Removes JMX Trans from system tests
* Removes the JMX Trans example
* And more

This should resolve #7692.

### Checklist

- [x] Update documentation
- [x] Update CHANGELOG.md


im-konge (approved):

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/JmxTransReconciler.java
```diff
@@ -4,149 +4,85 @@
...
 import io.vertx.core.Future;
 
-import java.util.List;
-
 /**
- * Class used for reconciliation of JMX Trans deployment. This class contains both the steps of the JMX Trans
- * reconciliation pipeline and is also used to store the state between them.
+ * JMX Trans is not supported anymore. The JmxTransReconciler class is still used to delete the JMXTrans deployment in
+ * case it was used by a previous Strimzi version. The only functionality here is now deleting the JMXTrans deployment.
+ * This will be completely removed in Strimzi 0.40.0.
```
ppatierno, line 18: Looking at the code I was wondering if we should log a WARN message to make clear the user what's going to happen and why JMXTrans related resources were deleted.

scholzj: The deprecation of the API already logs warning messages about this. It also adds it to the `.status` section of the custom resource. I think that is enough.

ppatierno: It logs deprecation which was ok until now. Not sure it's enough when the operator does an actual deletion of stuff if JMX is there. Unless the deprecation warning already mentions that JMX stuff is going to be deleted.

ppatierno: AFAIU, right now the JMX reconciler is going to delete the JMX related stuff right? How about logging it's deleting the stuff. Will this code then be removed in the future, I guess? So no JMX deletion nor logging anymore? Or the reconciliation will be still there?

scholzj: That is covered in the proposal -> 0.35-0.39 will be deleting any existing JMX Trans deployment. From 0.40, it iwll be completely removed and we will not do anything (even delete it if it still exists).

ppatierno: > That is covered in the proposal -> 0.35-0.39 will be deleting any existing JMX Trans deployment.

Imho, we should log (until 0.39 release) that JMX Trans are deleted if there is any. Then we'll remove all the code.
The current `JMXTrans is deprecated and removed in Strimzi 0.35.0` is pretty clear about API deprecation and support removed in 0.35.0 (so you cannot deploy JMX Trans anymore), but it isn't clear that it's going to remove/distrupt your JMX stuff (which instead will stay after 0.40.0 even if it still exists).

scholzj: But when do you want it to be logged? That is what is not clear to me. All the time?

> The current JMXTrans is deprecated and removed in Strimzi 0.35.0 is pretty clear about API deprecation and support removed in 0.35.0 (so you cannot deploy JMX Trans anymore), but it isn't clear that it's going to remove/distrupt your JMX stuff (which instead will stay after 0.40.0 even if it still exists).

TBH, the reality is that when you see any of the warnings, the deployment will be gone. So it will be IMHO pretty clear what `removed` means.

ppatierno: Can we just make it clearer with updating the warning messages as ...

> JMXTrans is deprecated and related resources removed in Strimzi 0.35.0

?

api/src/main/java/io/strimzi/api/kafka/model/KafkaSpec.java
```diff
@@ -96,10 +97,9 @@ public List<String> getMaintenanceTimeWindows() {
     }
 
     @Deprecated
-    @DeprecatedProperty(description = "JMXTrans is deprecated and will be removed in Strimzi 0.35.0")
+    @DeprecatedProperty(description = "JMXTrans is deprecated and removed in Strimzi 0.35.0")
```
PaulRMellor:
```suggestion
    @DeprecatedProperty(description = "JMXTrans is deprecated and removed in Strimzi 0.35.0.")
```

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/JmxTransReconciler.java
scholzj: Ok, should be changed now.

api/src/main/java/io/strimzi/api/kafka/model/KafkaSpec.java
```diff
@@ -96,10 +97,9 @@ public List<String> getMaintenanceTimeWindows() {
...
     @PresentInVersions("v1alpha1-v1beta2")
-    @Description("Configuration for JmxTrans. When the property is present a JmxTrans deployment is created for gathering JMX metrics from each Kafka broker. " +
-            "For more information see https://github.com/jmxtrans/jmxtrans[JmxTrans GitHub]")
+    @Description("JMXTrans is not supported anymore from Strimzi 0.35.0 and this option will be ignored.")
```
PaulRMellor, line 102:
```suggestion
    @Description("As of Strimzi 0.35.0, JMXTrans is not supported and this option is ignored.")
```

PaulRMellor (approved):
LGTM. A couple of suggestions on descriptions.

ppatierno (approved):

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/JmxTransReconciler.java
ppatierno: Thanks Jakub!

PR #8273, opened 2023-03-21
Graduate `UseStrimziPodSets` feature gate and remove `StatefulSet` support

scholzj:
### Type of change

- Task

### Description

This PR implements the proposal [SP#44](https://github.com/strimzi/proposals/blob/main/044-StrimziPodSets-graduation.md) and moves the `UseStrimziPodSets` feature gate and removes StatefulSet support. This includes:
* Removing the feature gate and making StrimziPodSets permanently enabled
* Deprecating the API parts related to StatefulSets (Statefulset templates)
* Deleting the StatefulSets code which is not needed for upgrade purposes
* Removing the parts of the code related only to StatefulSets (such as the shared Kafka configuration etc.)
* Removes tests related to StatefulSets
* Updates the documentation to remove parts related to StatefulSets
* It removes the Feature Gate system tests focused on the removed feature gate

Even after this, we still keep support to upgrade older Strimzi clusters using StatefulSets. So we are able to:
* Detect that StatefulSet is being used
* Migrate to StrimziPodSets

That means that we still have some limited StatefulSet functionality in the code such as getting the StatefulSets, deleting them, and so on. We also keep the RBAC for the time being. This will be removed only at a later stage to support direct upgrades for a longer time

The removal of StatefulSet support in system tests (other then removing the FG STs mentioned earlier) will be done in a separate PR (we should still keep testing upgrades)

### Checklist

- [x] Write tests
- [x] Update documentation
- [x] Update CHANGELOG.md


ppatierno (approved):

Frawless (approved):

documentation/modules/deploying/con-deploy-cluster-operator-watch-options.adoc
```diff
@@ -29,7 +29,7 @@ The Cluster Operator watches for changes to the following resources:
 * `KafkaBridge` for the Kafka Bridge instance.
 * `KafkaRebalance` for the Cruise Control optimization requests.
 
-When one of these resources is created in the Kubernetes cluster, the operator gets the cluster description from the resource and starts creating a new cluster for the resource by creating the necessary Kubernetes resources, such as StatefulSets, Services and ConfigMaps.
+When one of these resources is created in the Kubernetes cluster, the operator gets the cluster description from the resource and starts creating a new cluster for the resource by creating the necessary Kubernetes resources, such as Deployments, Pods, Services and ConfigMaps.
```
PaulRMellor, line 32:
```suggestion
When one of these resources is created in the Kubernetes cluster, the operator gets the cluster description from the resource and starts creating a new cluster for the resource by creating the necessary Kubernetes resources, such as `Deployment`, `Pod`, `Service` and `ConfigMap` resources.
```

documentation/modules/deploying/con-deploy-custom-resources-example.adoc
```diff
@@ -18,7 +18,7 @@ A CRD defines a new `kind` of resource, such as `kind:Kafka`, within a Kubernete
 
 The Kubernetes API server allows custom resources to be created based on the `kind` and understands from the CRD how to validate and store the custom resource when it is added to the Kubernetes cluster.
 
-WARNING: When CRDs are deleted, custom resources of that type are also deleted. Additionally, the resources created by the custom resource, such as pods and statefulsets are also deleted.
+WARNING: When CRDs are deleted, custom resources of that type are also deleted. Additionally, the resources created by the custom resource, such as pods and services are also deleted.
```
PaulRMellor, line 21:
```suggestion
WARNING: When a `CustomResourceDefinition` is deleted, custom resources of that type are also deleted. Additionally, Kubernetes resources created by the custom resource are also deleted, such as `Deployment`, `Pod`, `Service` and `ConfigMap` resources.
```

PaulRMellor (approved):
Great! A couple of minor suggestions

PR #8294, opened 2023-03-26
Use `PodRef` more widely in model and reconciler classes

scholzj:
### Type of change

- Refactoring

### Description

Today, we have many places where we convert broker IDs to pod names or extract broker IDs from pod names. For example, `KafkaCluster` has a method to return a list of pod names which are later converted to broker IDs and later pod name is generated from it again.

Inside the `KafkaRoller` class, we already have a subclass `PodRef` which is used internally in some places to carry the pod name and broker ID. This PR moves it to Java record and into a separate class in the model. It also renames it to `NodeRef` to better correspond to the use (it represents Kafka node)That way it can be used in more places. 

It also already starts using it in the `KafkaCluster` class and in the reconcilers. It also fixes some minor typos and Javadoc issues in the `KafkaRoller` class.

### Checklist

### DIFF
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/model/NodeRef.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/model/NodeRef.java
@@ -13,6 +13,6 @@ package io.strimzi.operator.cluster.model;
 public record NodeRef(String podName, int nodeId) {
     @Override
     public String toString() {
-        return podName;
+        return podName + "/" + nodeId;
     }
 }
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/model/NodeRef.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/model/NodeRef.java
@@ -1,0 +1,18 @@
+/*
+ * Copyright Strimzi authors.
+ * License: Apache License 2.0 (see the file LICENSE or http://apache.org/licenses/LICENSE-2.0.html).
+ */
+package io.strimzi.operator.cluster.model;
+
+/**
+ * Record used to keep reference to a Kafka node through its pod name and node ID.
+ *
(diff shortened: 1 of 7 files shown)


cluster-operator/src/main/java/io/strimzi/operator/cluster/model/NodeRef.java
```diff
@@ -0,0 +1,18 @@
...
+/**
+ * Record used to keep reference to a Kafka node through its pod name and node ID.
+ *
+ * @param podName   Name of the pod which represents this node
+ * @param nodeId    ID of the Kafka node
+ */
+public record NodeRef(String podName, int nodeId) {
+    @Override
+    public String toString() {
+        return podName;
```
ppatierno, line 16: I would expect that the toString represents somehow the whole instance in string format while here it's returning only the podName. Any specific reasons for that?

ppatierno (approved):
LGTM. I left a minor question/comment.

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/NodeRef.java
scholzj: This was already in the `PodRef` class. I can add there something like `<podName>/<nodeId>` if you want. But essentially, the node ID is part of the pod name. We just wanna safe the parsing, but there is no reason for them to be different. So it seemed unnecessry.

ppatierno: I would have your suggestion in place (using `<podName>/<nodeId>`) just for consistency about the general meaning of String representation of a instance ... but it's really up to you I can live with the current one, I was just curious.

scholzj: Ok, I updated it.

PR #8300, opened 2023-03-27
Turn off ExponentialBackoffIntervalCalculator during systemtests deug log

Frawless:
### Type of change

_Select the type of your PR_

- Refactoring

### Description

Debug lof of systemtests is full of lines likes this 
```
2023-03-27 07:23:58 DEBUG [ExponentialBackoffIntervalCalculator:61] Current reconnect backoff is 100 milliseconds (T0)
```
 This PR turn off this logger to make the debug log cleaner

### Checklist

### DIFF
--- a/systemtest/src/main/resources/log4j2.properties
+++ b/systemtest/src/main/resources/log4j2.properties
@@ -30,9 +30,7 @@ logger.clients.level = info
 logger.okhttp.name = okhttp3
 logger.okhttp.level = INFO
 
-logger.cr.name = io.fabric8.kubernetes.client.CustomResource
-logger.cr.level = OFF
-logger.vuu.name = io.fabric8.kubernetes.client.dsl.internal.VersionUsageUtils
-logger.vuu.level = OFF
-logger.ebic.name = io.fabric8.kubernetes.client.utils.ExponentialBackoffIntervalCalculator
-logger.ebic.level = OFF
+logger.fabric8.name = io.fabric8.kubernetes.client
+logger.fabric8.level = OFF
+logger.jayway.name = com.jayway.jsonpath.internal.path.CompiledPath
+logger.jayway.level = OFF
--- a/systemtest/src/main/resources/log4j2.properties
+++ b/systemtest/src/main/resources/log4j2.properties
@@ -3,7 +3,7 @@ name = STConfig
 appender.console.type = Console
 appender.console.name = STDOUT
 appender.console.layout.type = PatternLayout
-appender.console.layout.pattern = %d{yyyy-MM-dd HH:mm:ss}{GMT} [%thread] %highlight{%-5p} [%c{1}:%L] %m%n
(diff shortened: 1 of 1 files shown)


scholzj:
> @scholzj you can check to debug log file from STs [here](https://dev.azure.com/cncf/_apis/resources/Containers/18521087/systemtest-logs?itemPath=systemtest-logs%2Fstrimzi-debug-regression-bundle%20III.%20-%20dynconfig%20%2B%20tracing%20%2B%20watcher%20regression_dynconfig_listeners_tracing_watcher.log) You will find a lot of lines that are actually not from us and are just from fabric8 from some reason. It will remove hundreds of unwanted lines that make debug log unreadable without additional processing.

I don't think that really answers what other messages will get lost by blocking the whole Fabric8 package or the Jayway package. I also wonder why the DEBUG log is needed in the first place if you disable all other components from logging.

Just to be clear - all I'm saying is that I don't feel like I have any competency to review this PR. I'm not saying there is anything wrong with this PR.

Frawless:
> > @scholzj you can check to debug log file from STs [here](https://dev.azure.com/cncf/_apis/resources/Containers/18521087/systemtest-logs?itemPath=systemtest-logs%2Fstrimzi-debug-regression-bundle%20III.%20-%20dynconfig%20%2B%20tracing%20%2B%20watcher%20regression_dynconfig_listeners_tracing_watcher.log) You will find a lot of lines that are actually not from us and are just from fabric8 from some reason. It will remove hundreds of unwanted lines that make debug log unreadable without additional processing.
> 
> I don't think that really answers what other messages will get lost by blocking the whole Fabric8 package or the Jayway package. I also wonder why the DEBUG log is needed in the first place if you disable all other components from logging.
> 
> Just to be clear - all I'm saying is that I don't feel like I have any competency to review this PR. I'm not saying there is anything wrong with this PR.

I am not aware of any logs from fabric8 that we use in case of debug, maybe set it to TRACE level will be a better option instead of putting it OFF. At first, I disabled just specific loggers, but when I had there like 5-7 different classes I made this change and turn it off as a whole. 

@see-quick @im-konge  what are your preferences? Just disable classes directly or disable the whole package?

im-konge:
I would set it for whole package -> we never needed the log from fabric8 in tests. In components like TO, UO, ... , it's useful, but I don't see a benefit of it in tests log.
(just my opinion)

Or do you remember when these lines were actually useful?

I think that every operation that we are doing with fabric8 is logged by our STs (like creation of resources etc.), so that should not be a problem.

Frawless:
> I would set it for whole package -> we never needed the log from fabric8 in tests. In components like TO, UO, ... , it's useful, but I don't see a benefit of it in tests log. (just my opinion)
> 
> Or do you remember when these lines were actually useful?
> 
> I think that every operation that we are doing with fabric8 is logged by our STs (like creation of resources etc.), so that should not be a problem.

No, I don't. Also, these lines are pretty new. I didn't see them in previous changes to the STs logger a few months ago.

jankalinic (approved):

PR #8304, opened 2023-03-27
Initialize OpenTracing Tracer when using OpenTelemetry

Delorien84:
### Type of change

- Bugfix

### Description

Some third party connectors using OpenTracing API to access tracing functionality, eg. Debezium. When using Jaeger the GlobalTracer is initialized but not for OpenTelemetry.

This will allow use to migrate to OpenTelemetry as Jaeger is deprecated.

### Checklist

- [ ] Write tests
- [ ] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/docker-images/artifacts/kafka-thirdparty-libs/3.3.x/pom.xml
+++ b/docker-images/artifacts/kafka-thirdparty-libs/3.3.x/pom.xml
@@ -130,6 +130,11 @@
         </dependency>
         <dependency>
             <groupId>io.opentelemetry</groupId>
+            <artifactId>opentelemetry-opentracing-shim</artifactId>
+            <version>${opentelemetry.alpha-version}</version>
+        </dependency>
+        <dependency>
+            <groupId>io.opentelemetry</groupId>
             <artifactId>opentelemetry-exporter-otlp</artifactId>
             <version>${opentelemetry.version}</version>
         </dependency>
--- a/docker-images/artifacts/kafka-thirdparty-libs/3.4.x/pom.xml
+++ b/docker-images/artifacts/kafka-thirdparty-libs/3.4.x/pom.xml
@@ -130,6 +130,11 @@
         </dependency>
         <dependency>
             <groupId>io.opentelemetry</groupId>
+            <artifactId>opentelemetry-opentracing-shim</artifactId>
+            <version>${opentelemetry.alpha-version}</version>
+        </dependency>
+        <dependency>
+            <groupId>io.opentelemetry</groupId>
             <artifactId>opentelemetry-exporter-otlp</artifactId>
             <version>${opentelemetry.version}</version>
         </dependency>
--- a/pom.xml
+++ b/pom.xml
@@ -127,6 +127,7 @@
         <jaeger.version>1.8.1</jaeger.version>
         <opentracing.version>0.33.0</opentracing.version>
         <opentracing-kafka.version>0.1.13</opentracing-kafka.version>
+        <opentelemetry.version>1.18.0</opentelemetry.version>
         <opentelemetry.alpha-version>1.18.0-alpha</opentelemetry.alpha-version>
         <jetty.version>9.4.48.v20220622</jetty.version>
         <javax-servlet.version>3.1.0</javax-servlet.version>
@@ -699,7 +700,17 @@
             </dependency>
             <dependency>
                 <groupId>io.opentelemetry</groupId>
+                <artifactId>opentelemetry-sdk</artifactId>
+                <version>${opentelemetry.version}</version>
+            </dependency>
+            <dependency>
+                <groupId>io.opentelemetry</groupId>
                 <artifactId>opentelemetry-sdk-extension-autoconfigure</artifactId>
+                <version>${opentelemetry.alpha-version}</version>
+            </dependency>
+            <dependency>
+                <groupId>io.opentelemetry</groupId>
+                <artifactId>opentelemetry-opentracing-shim</artifactId>
                 <version>${opentelemetry.alpha-version}</version>
             </dependency>
             <dependency>
--- a/tracing-agent/pom.xml
+++ b/tracing-agent/pom.xml
@@ -42,6 +42,16 @@
             <artifactId>opentelemetry-sdk-extension-autoconfigure</artifactId>
             <scope>provided</scope>
         </dependency>
+        <dependency>
+            <groupId>io.opentelemetry</groupId>
+            <artifactId>opentelemetry-opentracing-shim</artifactId>
+            <scope>provided</scope>
+        </dependency>
+        <dependency>
+            <groupId>io.opentelemetry</groupId>
+            <artifactId>opentelemetry-sdk</artifactId>
+            <scope>provided</scope>
+        </dependency>
     </dependencies>
 
     <build>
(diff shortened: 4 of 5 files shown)


scholzj (commented):
Can you please explain more about the issue you are trying to address?

Also, how will this work after we drop the OpenTracing & Jaeger completely which is planned in June IIRC? The main issue here is that nobody is willing to merge the PRs for the OpenTracing Kafka support and it is quickly becoming unusable since it doesn't properly support new Kafka versions.

scholzj:
PS: The build failed with this ...

```
[INFO] --- maven-dependency-plugin:3.3.0:analyze-only (analyze) @ tracing-agent ---
[WARNING] Used undeclared dependencies found:
[WARNING]    io.opentelemetry:opentelemetry-sdk:jar:1.18.0:provided
```

That suggests you are using some more dependnecies which you did not add to the pom.xml files.

Delorien84:
Hello,
 I am trying to setup Debezium with tracing support (https://debezium.io/documentation/reference/stable/integrations/tracing.html). After try with error I notice that it works only with Jaeger (`tracing.type = jaeger`). After digging I found that Debezium use OpenTracing API. https://github.com/debezium/debezium/blob/main/debezium-core/src/main/java/io/debezium/transforms/tracing/ActivateTracingSpan.java#L147 . When using Jaeger you already initialize GlobalTracer but not for OpenTelemetry.

Delorien84:
> PS: The build failed with this ...
> 
> ```
> [INFO] --- maven-dependency-plugin:3.3.0:analyze-only (analyze) @ tracing-agent ---
> [WARNING] Used undeclared dependencies found:
> [WARNING]    io.opentelemetry:opentelemetry-sdk:jar:1.18.0:provided
> ```
> 
> That suggests you are using some more dependnecies which you did not add to the pom.xml files.

The error was already present, but I may fix it.

scholzj:
Ok, so the _shim_ will essentially convert the OpenTracing traces from components like Debezium into OpenTelemtry traces? did I get it right?

Delorien84:
> Ok, so the _shim_ will essentially convert the OpenTracing traces from components like Debezium into OpenTelemtry traces? did I get it right?

Yes, it is a little bridge between OpenTracing and OpenTelemetry.

scholzj:
> Yes, it is a little bridge between OpenTracing and OpenTelemetry.

Ok, that makes sense in that case. I wonder if we should have some flag to enable / disable this. But in general, that should be fine.

> The error was already present, but I may fix it.

It certainly wasn't failing any previous builds ;-). I think it's the call to `getOpenTelemetrySdk` causing this. The `OpenTelemetrySdk` class is in `opentelemetry-sdk`. So I don't think the previous version really used that dependency.

scholzj:
PS: You will also need to fix the DCO signoff. The instructions should be under _Details_ link next to the status.

Delorien84:
Thanks, in the meantime I'll try to make MR on Debezium to switch to OpenTelemetry.

Delorien84:
Oh, previously not found but I see bug in Debezium, you were mention here also https://issues.redhat.com/browse/DBZ-2862

Delorien84:
And PR https://github.com/debezium/debezium/pull/4270

ppatierno:
I am trying to understand better how it's going to work ...
AFAICS when you select `tracing.type = opentelemetry`, with your changes it's going to initialize the OpenTelemetry SDK but then use the shim to get a Jaeger tracer and register it. So at this point, is Jaeger used to send the traces? Are we not using OTel anymore? Is this really what we want? It sounds to me that until we don't remove Jaeger definitevely, when people set "opentelemetry" tracing they will still see traces generated by Jaeger. Is my understanding correct?

ppatierno:
Taking a look at shim code, it seems that it's providing a way to use OTel underneath but through OTrace API.
My only doubt is ... Debezium is going to work because it uses OTrace API and the shim helps with using OpenTelemetry underneath.
What happens for a connector which uses OpenTelemetry API directly but we are using the shim to register an OTrace traces?

Delorien84:
> Taking a look at shim code, it seems that it's providing a way to use OTel underneath but through OTrace API. My only doubt is ... Debezium is going to work because it uses OTrace API and the shim helps with using OpenTelemetry underneath. What happens for a connector which uses OpenTelemetry API directly but we are using the shim to register an OTrace traces?

Hello, I do not think that something happened for those connectors. If they are using OpenTelemetry APi directly, they should use `GlobalOpenTelemetry.get()` to retrieve the current instance. And should not have been affected. The connectors in that case should not create new instance of `AutoConfiguredOpenTelemetrySdk.initialize()` because Strimzi already does it.


According to the work in progress of Debezium (https://github.com/debezium/debezium/pull/4270) to use OpenTelemetry API directly we may close this MR. 

My only concern is that Debezium will release the change sooner than you release version with OpenTrace / Jeager removal. But for migration scenario it should be better to have at least one Strimzi version with that change. Then we may do migration without any downtime:

1. Upgrade Strizmi with this change
2. Switch from Jaeger to OpenTelemetry. Setting `tracing.type=opentelemetry` and related ENVs.
3. Upgrade Debezium using OpenTelemetry 
4. Upgrade Strimzi without OpenTrace/Jaeger 


Should I continue and fix the the pending issue? Or we will close this?

Thank you

ppatierno:
@Delorien84 ok I think this PR would make sense for the time being and removing the shim later when we have only OpenTelemetry support.
Anyway you should fix the DCO and checking the build failures.

scholzj (approved):

ppatierno (approved):

scholzj:
Thanks for the PR @Delorien84

vidourem:
Hello,
I'm working on the migration from OpenTracing to OpenTelemetry for the Debezium project [](https://github.com/debezium/debezium/pull/4270). OpenTelemetry doesn't yet provide intrumentation for Kafka Connect and in the meantime I've developed a Kafka interceptor to enable distributed tracability. This interceptor extends the one provided by the OpenTelemetry project, which is still in alpha. As it happens, the OpenTelemetry interceptor package has changed as of version 1.23.0 
Before 1.23.0: 
[io/opentelemetry/instrumentation/kafkaclients/TracingProducerInterceptor.java](https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/v1.22.1/instrumentation/kafka/kafka-clients/kafka-clients-2.6/library/src/main/java/io/opentelemetry/instrumentation/kafkaclients/TracingProducerInterceptor.java#L6)

After 1.23.0: [io/opentelemetry/instrumentation/kafkaclients/v2_6/TracingProducerInterceptor.java](https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/v1.23.0/instrumentation/kafka/kafka-clients/kafka-clients-2.6/library/src/main/java/io/opentelemetry/instrumentation/kafkaclients/v2_6/TracingProducerInterceptor.java#L6)

As a result, I've decided to use vesrion 1.23.0 in Debezium to avoid problems when OpenTelemetry versions are upgraded in Debezium.

At the same time I'm updating the example project [](https://github.com/debezium/debezium-examples/tree/main/outbox)

This project uses a Docker Strimzi image as a Kafka distribution, however the latest version of Strimzi embeds Opentelemetry in version 1.19.0 which doesn't allow the interceptor to work for the reason mentioned above. Do you plan to upgrade Opentelemetry to version 1.23.0+ in your next release?

ppatierno:
@vidourem could you open a discussion to talk about this and referencing this already merged PR please? I would like to avoid polluting a PR with comments after it's merged. We can chat there when you opened it, thanks!

PR #8305, opened 2023-03-27
Use node references instead of replica count in `ClusterCa`

scholzj:
### Type of change

- Refactoring

### Description

The `ClusterCa` class should use a list of `NodeRef` objects instead of a replica count when generating or renewing certificates. This should remove the separate methods for generating pod names from the `ClusterCa` class and use the centrally generated names. It also prepares it for future support of clusters which will not have always a continuous set of node IDs and uniform pod names.

The changed method - `maybeCopyOrGenerateCerts` - is also used by Cruise Control. The calls from Cruise Control have been modified as well. Although in that case, it is not using a real pod name. But that was the case already before when it passed the custom pod name and replica count to this method. This PR just adjusts to the changes done mainly for Kafka and ZooKeeper clusters

This PR also adds additional tests covering more situations in which `maybeCopyOrGenerateCerts` might be called.

### Checklist

- [x] Write tests

### DIFF
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/model/ClusterCa.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/model/ClusterCa.java
@@ -152,7 +152,7 @@ public class ClusterCa extends Ca {
     protected Map<String, CertAndKey> generateCcCerts(String namespace, String kafkaName, boolean isMaintenanceTimeWindowsSatisfied) throws IOException {
         DnsNameGenerator ccDnsGenerator = DnsNameGenerator.of(namespace, CruiseControlResources.serviceName(kafkaName));
 
-        Function<Integer, Subject> subjectFn = i -> {
+        Function<NodeRef, Subject> subjectFn = node -> {
             Subject.Builder subject = new Subject.Builder()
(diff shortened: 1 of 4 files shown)


cluster-operator/src/main/java/io/strimzi/operator/cluster/model/ClusterCa.java
```diff
@@ -285,79 +292,74 @@ protected boolean hasCaCertGenerationChanged() {
...
+
+                CertAndKey certAndKey;
+
+                if (isNewVersion(secret, podName)) {
+                    certAndKey = asCertAndKey(secret, podName);
+                } else {
+                    // coming from an older operator version, the secret exists but without keystore and password
+                    certAndKey = addKeyAndCertToKeyStore(subject.commonName(),
+                            Base64.getDecoder().decode(secretEntryDataForPod(secret, podName, SecretEntry.KEY)),
+                            Base64.getDecoder().decode(secretEntryDataForPod(secret, podName, SecretEntry.CRT)));
```
ppatierno, line 329: I was wondering if at some point we could get rid of this code but we should check the older version which still needs it.

ppatierno (approved):
Good refactoring. Thanks!

PR #8310, opened 2023-03-28
Remove the deprecated Fabric8 method createOrReplace from the project

### LINKED ISSUE #8270: [ST] Replace createOrReplace() calls in the `systemtests` module
From Fabric8 6.5, the `createOrReplace()` calls are deprecated. We still use them a lot in the System Tests. In order to avoid problems in future Fabric8 upgrades, we should replace the use of this call either through detecting the errors and doing our own createOrReplace or even better through fixing the tests using this so that they actually don't need to use createOrReplace in the first place.

im-konge:
### Type of change

- Refactoring

### Description

This PR removes the deprecated Fabric8 method `createOrReplace`.
Instead, we should use the `create()` or `update()` methods depending on the use case.
As part of this PR, I'm also adding few methods for creating and updating various resources and changing it in the tests.
Also, some of the resources like ClusterRoles, CRDs, etc., need to have `createOrUpdate` method for updating the resource when it already exists (because of leftovers from different tests etc.).

Fixes #8270 

### Checklist

### DIFF
--- a/test/src/main/java/io/strimzi/test/k8s/KubeClient.java
+++ b/test/src/main/java/io/strimzi/test/k8s/KubeClient.java
@@ -4,18 +4,7 @@
  */
 package io.strimzi.test.k8s;
 
-import io.fabric8.kubernetes.api.model.ConfigMap;
-import io.fabric8.kubernetes.api.model.DeletionPropagation;
-import io.fabric8.kubernetes.api.model.Event;
-import io.fabric8.kubernetes.api.model.LabelSelector;
-import io.fabric8.kubernetes.api.model.Namespace;
-import io.fabric8.kubernetes.api.model.NamespaceBuilder;
-import io.fabric8.kubernetes.api.model.Node;
-import io.fabric8.kubernetes.api.model.PersistentVolumeClaim;
-import io.fabric8.kubernetes.api.model.Pod;
-import io.fabric8.kubernetes.api.model.Secret;
-import io.fabric8.kubernetes.api.model.Service;
-import io.fabric8.kubernetes.api.model.ServiceAccount;
+import io.fabric8.kubernetes.api.model.*;
 import io.fabric8.kubernetes.api.model.admissionregistration.v1.ValidatingWebhookConfiguration;
 import io.fabric8.kubernetes.api.model.apiextensions.v1.CustomResourceDefinition;
 import io.fabric8.kubernetes.api.model.apps.Deployment;
@@ -28,6 +17,7 @@ import io.fabric8.kubernetes.api.model.rbac.ClusterRole;
 import io.fabric8.kubernetes.api.model.rbac.ClusterRoleBinding;
 import io.fabric8.kubernetes.api.model.rbac.Role;
 import io.fabric8.kubernetes.api.model.rbac.RoleBinding;
+import io.fabric8.kubernetes.api.model.storage.StorageClass;
 import io.fabric8.kubernetes.client.KubernetesClient;
 import io.fabric8.kubernetes.client.KubernetesClientBuilder;
 import io.fabric8.kubernetes.client.VersionInfo;
@@ -55,9 +45,6 @@ import java.util.stream.Collectors;
 
 import static io.strimzi.test.k8s.KubeClusterResource.kubeClient;
 
-// This class is using the deprecated createOrReplace methods which should be removed
-// Removing these deprecated calls is tracked by https://github.com/strimzi/strimzi-kafka-operator/issues/8269
-@SuppressWarnings("deprecation")
 public class KubeClient {
 
     private static final Logger LOGGER = LogManager.getLogger(KubeClient.class);
@@ -95,11 +82,7 @@ public class KubeClient {
 
     public void createNamespace(String name) {
         Namespace ns = new NamespaceBuilder().withNewMetadata().withName(name).endMetadata().build();
-        client.namespaces().resource(ns).createOrReplace();
-    }
-
-    public void createOrReplaceNamespace(final Namespace namespace) {
-        client.namespaces().resource(namespace).createOrReplace();
+        client.namespaces().resource(ns).create();
     }
 
     public void deleteNamespace(String name) {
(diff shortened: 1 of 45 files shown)


im-konge:
I think there are two options for this:

1) check the usages as you mention here and use the create/update depending on that
2) do some encapsulation as you did for secrets (or something like that) -> use create and when an exception about an already existing resource appears, do the update

WDYT?

scholzj:
Well, I think first of all, if you have methods like this: `public void create(KafkaConnect resource)` => then they should clearly do one of the following:
* Use `create()` to create the resource
* Be renamed to correspond to what they actually do in the code

But I think in most cases you should really know whether you want to create something or update something. Only in cases when you really have good reasons why you don't know whether you are creating or updating something you should use what I did in the UO or use the server apply. I thought I had it in my case in UO. But I'm not sure that is the case here => I guess there might be legit cases for it, but somehow this PR did not convince me that it was really considered. 

TBH, if I just wanted to replace the deprecated calls without any further consideration, I would have done it myself and not opened any issues for it.

You should probably also think if server-side apply really does what you want here as it is not the same as replace.

im-konge:
@scholzj based on your comments, I went through the code and used the `create()` and `update()` methods in case where are needed. I will anyway have to run all tests if I didn't break anything else.

I guess this is now correct solution?

scholzj (commented):
This looks better code wise, yes. Thanks.

im-konge:
@scholzj @Frawless @see-quick this is ready for review now.

im-konge:
@strimzi-ci run tests test_only profile=all groups=olm

test/src/main/java/io/strimzi/test/k8s/KubeClient.java
```diff
@@ -604,8 +602,19 @@ public List<ServiceAccount> listServiceAccounts(String namespaceName) {
...
+     * In case that the ServiceAccount is already created, it is being updated.
+     * This can be caused by not cleared ServiceAccounts from other tests on in case we shut down the test before the cleanup
+     * phase. It should not have an impact on the functionality, we just update the ServiceAccount.
+     * @param serviceAccount ServiceAccount that we want to create or update
+     */
+    public void createOrUpdateServiceAccount(ServiceAccount serviceAccount) {
+        try {
+            client.serviceAccounts().inNamespace(serviceAccount.getMetadata().getNamespace()).resource(serviceAccount).create();
+        } catch (KubernetesClientException e) {
+            client.serviceAccounts().inNamespace(serviceAccount.getMetadata().getNamespace()).resource(serviceAccount).update();
```
scholzj, line 616: You should check for the right error here. What is the reasons for SA not being cleaned up?

test/src/main/java/io/strimzi/test/k8s/KubeClient.java
```diff
@@ -663,8 +672,19 @@ public List<ClusterRole> listClusterRoles() {
...
+     * This can be caused by not cleared ClusterRoles from other tests on in case we shut down the test before the cleanup
+     * phase. It should not have an impact on the functionality, we just update the ClusterRole.
+     * @param clusterRole ClusterRole that we want to create or update
+     */
+    public void createOrUpdateClusterRoles(ClusterRole clusterRole) {
+        try {
+            client.rbac().clusterRoles().resource(clusterRole).create();
+        } catch (KubernetesClientException e) {
+            client.rbac().clusterRoles().resource(clusterRole).update();
+        }
```
scholzj, line 687: Same as above. You should check for the right error here. What are the reasons for SA not being cleaned up?

There seem to be a lot of these. I won't repeat this comment anymore.

test/src/main/java/io/strimzi/test/k8s/KubeClient.java
im-konge: We are applying the resources from `packaging/install/cluster-operator` as they are, so we are not changing their names. That is a problem in cases when we are installing multiple operators.

I wanted to create an issue for fixing that, because I think that this is out of the scope of this PR (changing the bundle installation and verify it is working for all cases), that's why I added those try-catches into the code.

And about the error, I think that this is correct. Or you mean that I should check that it's really the `Already exists` error?

scholzj: > We are applying the resources from packaging/install/cluster-operator as they are, so we are not changing their names. That is a problem in cases when we are installing multiple operators.
> 
> I wanted to create an issue for fixing that, because I think that this is out of the scope of this PR (changing the bundle installation and verify it is working for all cases), that's why I added those try-catches into the code.

I'm not entirely sure why just because they are from install means they cannot be deleted. But you should add it to the comment then. Separate Pr for fixing that is fine.

> And about the error, I think that this is correct. Or you mean that I should check that it's really the Already exists error?

You should check the cause, yes. Not blindly retry after seeing a very generic exception.

im-konge: > I'm not entirely sure why just because they are from install means they cannot be deleted. But you should add it to the comment then. Separate Pr for fixing that is fine.

Yeah they are normally deleted when the test(s) are finished with the clean up. But when you stop the test in middle of the execution, the clean up is not done, thus the resources stay there. When you want to run another test, it fails just because the resources already exist. So there are like two issues.

> You should check the cause, yes. Not blindly retry after seeing a very generic exception.

Will do.

scholzj: > Yeah they are normally deleted when the test(s) is(are) finished with the clean up. But when you stop the test in middle of the execution, the clean up is not done, thus the resources stay there. When you want to run another test, it fails just because the resources already exist. So there are like two issues.

Fine, so that should be in the comments I guess.

im-konge: > Fine, so that should be in the comments I guess.

I have there this `on in case we shut down the test before the cleanup phase` -> besides the typo, I thought that it will be sufficient. Maybe adding one or two more sentences would make it clearer.

see-quick (approved):
LGTM, assuming that test will pass

scholzj (approved):
LGTM assuming the tests pass.

Issue #8306, opened 2023-03-27
[Bug] Kafka Exporter dashabord does not work in newer Grafana versions


scholzj:
The Kafka Exporter Grafana dashboard which we provide as part of our examples does not seem to work in newer Grafana versions such as 9.x or even 8.x. Check the attached screenshots => weird decimal numbers seem to be listed in places where they make no sense (e.g. number of topics or partitions) and the tables seem to be broken.

Note: All seems to work fine on older Grafana versions such as 7.x.

![grafana1](https://user-images.githubusercontent.com/5658439/228074566-b819d477-9c8f-4895-b850-5725a8aa9e6d.png)
![grafana2](https://user-images.githubusercontent.com/5658439/228074571-ef5b640a-cc25-47db-95dd-091ebff60cd4.png)

PR #8323, opened 2023-03-30
Add datasource to all panels that missing it in Kafka Bridge and Kafka Exporter dashboards

Frawless:
### Type of change

_Select the type of your PR_

- Bugfix

### Description

KafkaBridge and KafkaExporter were missing datasource configuration in several places what lead to wrong values when dashboards were loaded manually via UI or via JSON config in Grafana CR via Operator. This PR should fix it.

fix https://github.com/strimzi/strimzi-kafka-operator/issues/8306

Tested with Grafana 9.4.7

Before fix:
image

After fix:
image

### Checklist

### DIFF
--- a/packaging/examples/metrics/grafana-dashboards/strimzi-kafka-bridge.json
+++ b/packaging/examples/metrics/grafana-dashboards/strimzi-kafka-bridge.json
@@ -3020,6 +3020,7 @@
       "bars": false,
       "dashLength": 10,
       "dashes": false,
+      "datasource": "${DS_PROMETHEUS}",
       "fill": 1,
       "gridPos": {
         "h": 8,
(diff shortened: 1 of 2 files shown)


ppatierno (commented):
Have you also tested that the problem was in bridge dashboards? And it's now fixed?

Frawless:
> Have you also tested that the problem was in bridge dashboards? And it's now fixed?

Not really as I don't have it deployed now and have other work to do. However, when I deployed dashboard even when there were no bridge metrics available, these 3 have random data filled in the dashboard so from my POV the missing datasource is a bug as I would expect empty charts in that case.

scholzj (approved):
LGTM. Thanks for looking into this.

ppatierno (approved):

PR #8326, opened 2023-03-31
Add readiness check before we check pods for rolling update

scholzj:
### Type of change

- Refactoring

### Description

Today, the `KafkaRoller` does not handle well when the pods are not ready when the rolling update check starts. When the pods are not ready for example because they are still starting, the Roller will _roll_ them just because they are not ready because it cannot really connect to them. 

That has several side effects. One of them is for example described in [SP#048](https://github.com/strimzi/proposals/blob/main/048-avoid-broker-restarts-when-in-recovery.md) and deals with issues when the broker needs a lot of time to recover files after an unclean shutdown. 

Another one - on which this PR focuses - is the split of PodSet update, rolling update, and scale-up. The reconciliation flow today is as follows:
* (N) Update the PodSet, but do not scale it up (i.e. add any new pods) -> update only the YAMLs for the existing pods. If it is the first reconciliation and the pods are being created, we have to wait for readiness in this step before letting the `KafkaRoller` check the pods.
* (N+1) Check the existing pods for rolling update
* (N+2) Scale up the Kafka cluster (update the PodSet again to add new pods)

This logic was reasonably straightforward with StatefulSets where we had a single template and the scale-up or not was just about setting the right number of replicas. But with StrimziPodSets, this requires more complexity because it is not just changing the number of replicas. It actually needs the new pods to be generated. And in the future with things such as non-continuous node IDs and the KRaft architectures, or things such as node pools, this might become even more complex.

This PR tries to simplify the code by moving more responsibility to the `KafkaRoller`. It first checks for the readiness of the pod and waits for it before it actually tries to connect to the Pod and checks if a rolling update is really needed. If the pod does not become ready in time, it will proceed with the KafkaRoller process as before and decide if the pod needs to be rolled or not (so if the pod is for some reason never getting ready, the roller will still get its change to roll through it).

Thanks to this, we can not remove the separate scale-up step and do the scale-up immediately when updating the PodSet and we save the additional step and the complexity needed to understand which pods need to be included in the pod set at which point. This might consume slightly more CPU cycles as the `KafkaRoller` will now check the scaled-up pods. But the overall reconciliation time might be negligible since today we anyway wait for the pod readiness in a different place.

This might also help with situations when the user scaled down a pod that still had data and made the remaining brokers impossible to roll because of the number of in-sync partitions. Now we will scale up before rolling, so the broker will have a chance to start and re-sync the data.

_Note: This is something that is based on my understanding anyway planned in [SP#048](https://github.com/strimzi/proposals/blob/main/048-avoid-broker-restarts-when-in-recovery.md) where it will also check the recovery progress._ 

----

This PR also unifies the check if the pod is stuck (e.g. Pending because it is not-schedulable etc.). It merges the checks which were directly in the `KafkaRoller` code with the `isPodStuck` methods and uses now the method in all places.

### Checklist

### DIFF
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/resource/KafkaRoller.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/resource/KafkaRoller.java
@@ -362,6 +362,17 @@ public class KafkaRoller {
             throw new UnforceableProblem("Error getting pod " + nodeRef.podName(), e);
         }
 
+        if (!isPodStuck(pod)) {
+            // We want to give pods chance to get ready before we try to connect to the or consider them for rolling.
+            // But only in case when they are not stuck as that would
+            LOGGER.debugCr(reconciliation, "Waiting for pod {} to become ready before checking its state", nodeRef.podName());
+            try {
+                await(isReady(pod), operationTimeoutMs, TimeUnit.MILLISECONDS, e -> new RuntimeException(e));
+            } catch (Exception e)   {
+                LOGGER.warnCr(reconciliation, "Pod {} is not ready. We will check if KafkaRoller can do anything about it.", nodeRef.podName());
+            }
+        }
+
         restartContext.restartReasons = podNeedsRestart.apply(pod);
 
         try {
@@ -424,10 +435,10 @@ public class KafkaRoller {
     }
 
     private boolean isPending(Pod pod) {
-        if (pod != null && pod.getStatus() != null && "Pending".equals(pod.getStatus().getPhase())) {
-            return true;
-        }
-        return false;
+        return pod != null
+                && pod.getStatus() != null
+                && "Pending".equals(pod.getStatus().getPhase())
+                && pod.getStatus().getConditions().stream().anyMatch(ps -> "PodScheduled".equals(ps.getType()) && "Unschedulable".equals(ps.getReason()) && "False".equals(ps.getStatus()));
     }
 
     private boolean isPodStuck(Pod pod) {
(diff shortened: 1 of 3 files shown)


cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/resource/KafkaRoller.java
```diff
@@ -362,6 +362,17 @@ private void restartIfNecessary(NodeRef nodeRef, RestartContext restartContext)
             throw new UnforceableProblem("Error getting pod " + nodeRef.podName(), e);
         }
 
+        if (!isPodStuck(pod)) {
+            // We want to give pods chance to get ready before we try to connect to the or consider them for rolling.
+            // But only in case when they are not stuck as that would
```
tombentley, line 367: "as that would..." what?

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/resource/KafkaRoller.java
```diff
@@ -424,10 +435,10 @@ private boolean podWaitingBecauseOfAnyReasons(Pod pod, Set<String> reasons) {
     }
 
     private boolean isPending(Pod pod) {
```
tombentley, line 437: Should this still be called `isPending`? It seems it's more like `isPendingAndUnschedulable`

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/resource/KafkaRoller.java
```diff
@@ -362,6 +362,17 @@ private void restartIfNecessary(NodeRef nodeRef, RestartContext restartContext)
...
         }
 
+        if (!isPodStuck(pod)) {
+            // We want to give pods chance to get ready before we try to connect to the or consider them for rolling.
+            // But only in case when they are not stuck as that would
+            LOGGER.debugCr(reconciliation, "Waiting for pod {} to become ready before checking its state", nodeRef.podName());
+            try {
+                await(isReady(pod), operationTimeoutMs, TimeUnit.MILLISECONDS, e -> new RuntimeException(e));
+            } catch (Exception e)   {
+                LOGGER.warnCr(reconciliation, "Pod {} is not ready. We will check if KafkaRoller can do anything about it.", nodeRef.podName());
```
tombentley, line 372: This message isn't entirely accurate: We can only conclude the pod is not ready is the await timed out, (which gets wrapped in `RuntimeException`). If it failed for some other reason then we don't know if the pod is really ready or not. To be clear: I think it's probably OK to carry on with maybe restarting it in either case.

scholzj:
@tombentley Thanks Tom, your comments should be fixed now.

tombentley (approved):
LGTM.

PR #8329, opened 2023-03-31
[WIP] Redesign the `ClusterOperatorConfig` class to make it more flexible and clean

ShubhamRwt:
### Type of change

_Select the type of your PR_

- Enhancement

### Description

This PR redesign the `ClusterOperatorConfig` class. This PR gets rid of a lot of parameter declaration which was happening earlier and this new design make the `ClusterOperatorConfig` cleaner. We have introduced a new inner class named ConfigParameter which wraps all the parameter corresponding to a config parameter for eg. its type, default value, key.
For different types we have created an abstract class named Type which takes care of the types of the config parameter

### Checklist

- [ ] Write tests
- [ ] Update documentation
- [ ] Update CHANGELOG.md


ShubhamRwt:
@scholzj @tombentley This PR is ready for review now. I have tested it locally and it was working

cluster-operator/src/test/java/io/strimzi/operator/cluster/ResourceUtils.java
```diff
@@ -614,52 +615,39 @@ public static ResourceOperatorSupplier supplierWithMocks(boolean openShift) {
...
+    public static ClusterOperatorConfig dummyClusterOperatorConfig() {
+
+        Map<String, String> envVars = new HashMap<>();
+
+        return ClusterOperatorConfig.buildFromMap(envVars, KafkaVersionTestUtils.getKafkaVersionLookup());
     }
 
     public static ClusterOperatorConfig dummyClusterOperatorConfig(KafkaVersion.Lookup versions, long operationTimeoutMs) {
-        return dummyClusterOperatorConfig(versions, operationTimeoutMs, "", true);
+        return new ClusterOperatorConfigBuilder(dummyClusterOperatorConfig(), versions)
```
scholzj, line 626: Maybe you can go through these methods and for example, remove those which are not used 10+ times or something like that? For example, this one is IMHO not used at all. Some others are used once or twice. Thanks tot he builder pattern, it should be now much easier to do this directly in the place where this is used without a need for a special method.

cluster-operator/src/test/java/io/strimzi/operator/cluster/operator/assembly/KafkaConnectorIT.java
```diff
@@ -152,7 +152,7 @@ public void testConnectorNotUpdatedWhenConfigUnchanged(VertxTestContext context)
         );
 
         KafkaConnectAssemblyOperator operator = new KafkaConnectAssemblyOperator(vertx, pfa, ros,
-                ClusterOperatorConfig.fromMap(Collections.emptyMap(), KafkaVersionTestUtils.getKafkaVersionLookup()),
+                ClusterOperatorConfig.buildFromMap(Collections.emptyMap(), KafkaVersionTestUtils.getKafkaVersionLookup()),
```
scholzj, line 155: I think we tend to use `Map.of` in these places instead of `Collections.emptyMap()` => at least in the newer code. So maybe you can use it in this PR as well (not just here but also below etc.).

operator-common/src/main/java/io/strimzi/operator/common/operator/resource/ConfigParameterParser.java
```diff
@@ -84,6 +93,18 @@
...
 
+
+    /**
+     * A Java Duration
+     */
+    ConfigParameterParser<Duration> DURATION = s -> {
+        try {
+            return Duration.ofMillis(Long.parseLong(s));
+        } catch (NumberFormatException e) {
+            throw new InvalidConfigurationException("Failed to parse. Value is not valid", e);
```
scholzj, line 104: Can you list the value in the exception?

cluster-operator/src/main/java/io/strimzi/operator/cluster/ClusterOperatorConfig.java
```diff
@@ -526,74 +375,135 @@ private static Labels parseLabels(Map<String, String> vars, String configuration
...
         }
     }
 
-    private static LeaderElectionManagerConfig parseLeaderElectionConfig(Map<String, String> envVars)   {
-        boolean enabled = parseBoolean(envVars.get(STRIMZI_LEADER_ELECTION_ENABLED), DEFAULT_LEADER_ELECTION_ENABLED);
+    static ConfigParameterParser<FeatureGates> parseFeatureGates() {
+        return FeatureGates::new;
+    };
+
+
```
scholzj, line 393: Double empty line is not needed.

scholzj (commented):
I left some nits. But it seems to lok mostly good. I will trigger the regression as well.

scholzj (approved):
Two more nits. Btu LGTM otherwise.

operator-common/src/main/java/io/strimzi/operator/common/operator/resource/ConfigParameterParser.java
```diff
@@ -50,7 +59,7 @@
     ConfigParameterParser<List<String>> SEMICOLON_SEPARATED_LIST = configValue -> {
         List<String> windows = null;
         if (configValue != null && !configValue.isEmpty()) {
-            windows = Arrays.asList(configValue.split(";"));
+            windows = asList(configValue.split(";"));
```
tombentley, line 62: Why is the variable called `windows`?

tombentley (approved):

PR #8332, opened 2023-03-31
[systemtest] Add annotation for tests that contain component not supporting FIPS

im-konge:
### Type of change

- Enhancement

### Description

This PR adds annotation for skipping tests in case one of the component inside the test is not supporting FIPS.
It will help us skip those tests when we are running regression on FIPS cluster.

### Checklist


im-konge:
@strimzi-ci run tests test_only profile=all groups=oauth

Frawless (approved):

see-quick (approved):

PR #8348, opened 2023-04-05
docs(connectors): updates the urls for the connector examples

PaulRMellor:
**Documentation**

Updates the URLs of the plugin artifacts used as example connectors in the docs

### Checklist

- [ ] Write tests
- [x] Update documentation
- [ ] Update CHANGELOG.md


scholzj (commented):
My thoughts =>
* Do we really want to have so specific examples?
* If yes, do we need to update them to the _latest_ version?

If the answer is yes, then we should at least make it copy-pastable and keep also the checksums up to date.

scholzj (approved):

PR #8355, opened 2023-04-06
docs(feature gates): updates for adoption of strimzipodset resource

PaulRMellor:
**Documentation**

Updates from the adoption of the `StrimziPodSet` resource

- The deployment verification checks switch to checks on pods rather than deployment resources
- Upgrade/downgrade notes for feature gates made consistent
- The downgrade  section includes a note to check on feature gate downgrade requirements

### Checklist

- [ ] Write tests
- [x] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/documentation/modules/deploying/proc-deploy-kafka-bridge.adoc
+++ b/documentation/modules/deploying/proc-deploy-kafka-bridge.adoc
@@ -38,11 +38,13 @@ kubectl get pods -n _<my_cluster_operator_namespace>_
 .Output shows the deployment name and readiness
 [source,shell,subs="+quotes"]
 ----
-NAME              READY  STATUS   RESTARTS
-my-bridge-bridge  1/1    Running  0
+NAME                       READY  STATUS   RESTARTS
+my-bridge-bridge-<pod_id>  1/1    Running  0
 ----
 +
 `my-bridge` is the name of the Kafka Bridge cluster.
++
+A pod ID identifies each pod created.
 +
 With the default deployment, you install a single Kafka Bridge pod.
 +
--- a/documentation/modules/deploying/proc-deploy-kafka-bridge.adoc
+++ b/documentation/modules/deploying/proc-deploy-kafka-bridge.adoc
@@ -32,20 +32,22 @@ kubectl apply -f examples/bridge/kafka-bridge.yaml
 +
 [source,shell,subs="+quotes"]
 ----
-kubectl get deployments -n _<my_cluster_operator_namespace>_
+kubectl get pods -n _<my_cluster_operator_namespace>_
 ----
 +
 .Output shows the deployment name and readiness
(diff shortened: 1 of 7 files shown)


documentation/modules/operators/ref-operator-cluster-feature-gates.adoc
```diff
@@ -45,8 +45,8 @@ For example, when you change service account labels or annotations using the `te
 The `UseStrimziPodSets` feature gate has moved to GA, which means it is now permanently enabled and cannot be disabled.
 Support for `StatefulSets` has been removed and Strimzi is now always using `StrimziPodSets` to manage Kafka and ZooKeeper pods.
 
-IMPORTANT: Downgrade to Strimzi 0.27 and earlier versions is not possible when using `StrimziPodSets`.
-If you need to downgrade to Strimzi 0.27 or earlier, you have to first downgrade to Strimzi 0.28 - 0.34, disable the `USeStrimziPodSets` feature gate and only then downgrade to Strimzi 0.27 or earlier.
+IMPORTANT: With the `UseStrimziPodSets` feature gate permanently enabled, it is no longer possible to downgrade directly between Strimzi 0.27 or earlier and Strimzi 0.35 and newer.
```
scholzj, line 48: It seems a bit confusing with the older version first when you talk about downgrade. As that starts with the new version. Mybe this would be better?

```suggestion
IMPORTANT: With the `UseStrimziPodSets` feature gate permanently enabled, it is no longer possible to downgrade directly from Strimzi 0.35 and newer to Strimzi 0.27 or earlier.
```

documentation/modules/deploying/proc-deploy-kafka-bridge.adoc
```diff
@@ -32,20 +32,22 @@ kubectl apply -f examples/bridge/kafka-bridge.yaml
...
+kubectl get pods -n _<my_cluster_operator_namespace>_
 ----
 +
 .Output shows the deployment name and readiness
 [source,shell,subs="+quotes"]
 ----
-NAME              READY  UP-TO-DATE  AVAILABLE
-my-bridge-bridge  1/1    1           1
+NAME              READY  STATUS   RESTARTS
+my-bridge-bridge  1/1    Running  0
```
scholzj, line 42: The name here would be more like `my-bridge-bridge-<xxxxxxxx>-<xxxxx>`. Thsi would apply also to the Connect and Mirror MAker 2 pods later when the feature gate is not enabled. So I wonder if we want to use something like `<id> everywhere for the suffix.

PaulRMellor: Yes. I thought to leave it out.  Is something missing from the comment? We could use something like `<pod_id>` that applies to sequential and random? Or I can say something about both for the connect/mm2 steps and we keep the sequential in the example?

PR #8359, opened 2023-04-07
Add unit test for the Kafka Exporter reconciler

scholzj:
### Type of change

- Task

### Description

We do not seem to have currently any direct unit test for the `KafkaExporterReconciler` class. This PR adds such test to check the enabled and disabled KE and their reconciliation as well as reconciliation when network policy generation is globally disabled on the operator level.

### Checklist

- [x] Write tests

### DIFF
--- a/cluster-operator/src/test/java/io/strimzi/operator/cluster/ResourceUtils.java
+++ b/cluster-operator/src/test/java/io/strimzi/operator/cluster/ResourceUtils.java
@@ -614,14 +614,14 @@ public class ResourceUtils {
         return supplier;
     }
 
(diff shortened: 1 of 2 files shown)


devguyio (approved):
LGTM

Just a small nit. It was a little challenging for me when I first tried to understand such unit tests to figure out why the test is testing e.g. that there's a SA or a deployment etc. So I wonder if it'd be a good idea to add some comment on the class that explains a bit what the tests should verify. E.g. `When KafkaExporter is set, the reconciler creates a SA, a secret etc` . 

IMHO that'd be a helpful habit to get into for new contributors, but that can be something separately done as well away from this PR.

ppatierno (approved):
Lgtm

PR #8377, opened 2023-04-11
Use `Set` instead of `List` to pass the `NodeRef` objects around

scholzj:
### Type of change

- Refactoring

### Description

When introducing the NodeRef objects and passing the around in the Cluster Operator instead of always generating again pod names from replicas and node IDs from pod names, I used `List` which was already used in the `KafkaRoller`. However, the nodes and thus also the `NodeRef` objects should be always unique. So it seems to make more sense to use `Set` to pass them around? In particular, `LinkedHashSet` is used to maintain the insertion order.

WDYT @ppatierno? Does Set make more sense here?

### Checklist


ppatierno (approved):
Yes I think it makes more sense as well as retaining the order using the LinkedHashSet. LGTM!

PR #8386, opened 2023-04-14
Refactor handling of health check probes in the model classes

scholzj:
### Type of change

- Refactoring

### Description

This PR does a minor refactoring of the health check probes in the `AbstractModel` and its subclasses.

* The `readinessPath` and `livenessPath` were used only in some classes and even there only partially. So I remove them.
* The reading of the probe options from the custom resources was the same in many classes. I refactored it into the `ProbeUtils` class using an interface to allow reading the probe options from the various CRs.
* Most of the operands share the same default settings. So instead of having them defined multiple times, it is not defined once in `ProbeUtils`.
* The new `ProbeUtils` class was created by renaming the original `ProbeGenerator` and adding additional methods and fields. The `...Utils` name seems to better correspond to its purpose and how the other classes are named

### Checklist

- [x] Write tests


ppatierno (approved):
LGTM

PR #8388, opened 2023-04-14
docs(providers): new single section for loading external config

PaulRMellor:
### Type of change

**Documentation**

Brings the content related to loading external configuration into a single section.
Moves the content (in two places) into the deploying guide.
Introduces the content by describing the configuration providers available.
Describes how to use the `externalConfiguration` property and providers without needing to set up access rights.
Adds new procedures to show how the `FileConfigProvider` and `DirectoryConfigProvider` are used.

image

### Checklist

- [ ] Write tests
- [x] Update documentation
- [ ] Update CHANGELOG.md


documentation/modules/configuring/proc-loading-config-from-env-vars.adoc
```diff
@@ -6,24 +6,44 @@
 = Loading configuration values from environment variables
 
 [role="_abstract"]
-This procedure shows how to use the Environment Variables Configuration Provider plugin.
+Use the `EnvVarConfigProvider` to provide configuration properties as environment variables.
+Environment variables can contain values from config maps or secrets.
 
-In the procedure, environment variables provide configuration properties for a connector.
-A database password is specified as an environment variable.
+In this procedure, environment variables provide configuration properties for a connector to communicate with Amazon AWS.
```
PaulRMellor, line 12: No config provider described in the original env vars content that mentioned AWS, but I thought it would make a better example when amalgamating all the env vars content into this procedure. Unless it is approached differently. In which case, this procedure will need some adjustment.

scholzj: I think that in most cases, the connectors for use with AWS will read these env vars automatically without any additional configuration. But still, I think it is a good example that might be valid in some cases and it would be anyway hard to come up with something better 🤷

ppatierno (approved):

PR #8392, opened 2023-04-16
Refactor the default replicas handling in `AbstractModel` classes

scholzj:
### Type of change

- Refactoring

### Description

This PR refactors the handling of the default number of replicas in the `AbstractModel` and its subclasses. In many cases, we have some `DEFAULT_REPLICAS` field with a default number of replicas. However, this is often used in the constructor and immediately overwritten in the `fromCrd` method, so it actually does not add any real value. This PR tries to remove the unneeded fields.

Unfortunately, some of our CRDs such as KafkaConnect does not set the `.spec.replicas` field as required, so it might not be set by the user. But we can set the default number of replicas in the `api` class by using an integer and setting a default value there. It also updates the docs where it specifies the default values.

### Checklist

- [x] Update documentation

### DIFF
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaConnectCluster.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaConnectCluster.java
@@ -220,7 +220,6 @@ public class KafkaConnectCluster extends AbstractModel implements SupportsMetric
                                                                 KafkaVersion.Lookup versions,
                                                                 C result) {
         result.replicas = spec.getReplicas();
-        //result.replicas = spec.getReplicas() != null && spec.getReplicas() >= 0 ? spec.getReplicas() : DEFAULT_REPLICAS;
         result.tracing = spec.getTracing();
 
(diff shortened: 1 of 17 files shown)


cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaConnectCluster.java
```diff
@@ -221,7 +219,8 @@ protected static <C extends KafkaConnectCluster> C fromSpec(Reconciliation recon
                                                                 KafkaConnectSpec spec,
                                                                 KafkaVersion.Lookup versions,
                                                                 C result) {
-        result.replicas = spec.getReplicas() != null && spec.getReplicas() >= 0 ? spec.getReplicas() : DEFAULT_REPLICAS;
+        result.replicas = spec.getReplicas();
+        //result.replicas = spec.getReplicas() != null && spec.getReplicas() >= 0 ? spec.getReplicas() : DEFAULT_REPLICAS;
```
ppatierno, line 223: remove the commented line?

ppatierno (approved):
LGTM. Just a nit about removing a commented line.

PR #8400, opened 2023-04-17
Fix certificate key replacement issue when Cluster Operator crashes after the trust is established

scholzj:
### Type of change

- Bugfix

### Description

When a Cluster CA private key is replaced, the operator needs to follow through a series of three different rolling updates:
1) First rolling update establishes the trust to the new CA while still using the old server certificates and trusting the old CA as well
2) Second rolling update issues and rolls out the new server certificates signed by the new CA (while still trusting the old CA as well)
3) Third rolling update removes the trust in the old CA

When the operator crashes between the first and second step or during the second step, it will not roll out the new server certificates signed by the new CA, but it will instead just silently bump the CA generation IDs in the secret annotations. That will cause the operands still use the old server certificates even after phase 2. That will work seemingly fine as we still trust the old CA. But in phase 3, we remove the trust in the old CA and thus break the operands.

This PR adds additional check of the secret annotation to trigger the renewal after such CO failure and make it possible to recover from it.

### Checklist

- [x] Write tests

### DIFF
--- a/cluster-operator/src/test/java/io/strimzi/operator/cluster/model/ClusterCaRenewalTest.java
+++ b/cluster-operator/src/test/java/io/strimzi/operator/cluster/model/ClusterCaRenewalTest.java
@@ -20,6 +20,7 @@ import org.junit.jupiter.api.extension.ExtendWith;
 import java.io.File;
 import java.io.IOException;
 import java.util.Base64;
+import java.util.HashMap;
 import java.util.LinkedHashSet;
 import java.util.Map;
 import java.util.Set;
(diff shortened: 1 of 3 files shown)


cluster-operator/src/test/java/io/strimzi/operator/cluster/model/ClusterCaRenewalTest.java
```diff
@@ -560,14 +603,19 @@ public CertAndKey addKeyAndCertToKeyStore(String alias, byte[] key, byte[] cert)
...
-        }
+        //@Override
+        //protected boolean hasCaCertGenerationChanged() {
+        //    return false;
+        //}
+        //
+        //@Override
+        //protected String caCertGenerationAnnotation() {
+        //    return null;
+        //}
```
ppatierno, line 618: remove commented code

ppatierno (approved):
Good catch, thanks Jakub! LGTM, just a nit.

PR #8411, opened 2023-04-19
chore: Update KafkaUser alm-example to use operations field

Kartikey-Mishra1:
### Type of change

_Select the type of your PR_

- Bugfix

### Description

Update Kafka user alm-example in CSV to use the new operations field instead of the old deprecated operation field

### Checklist

- [ ] Write tests
- [ ] Update documentation
- [ ] Update CHANGELOG.md


scholzj (commented):
@Kartikey-Mishra1 Out of curiosity, are you actually using this for something? Because it is actually not used to generate any Strimzi Operator Hub files.

Kartikey-Mishra1:
@scholzj I have probably made the change in the wrong place but when I was using Strimzi on Openshift the Kafka User sample yaml was containing the operation field, so I was hoping to fix that with this. 

image

scholzj:
That would need to be changed in the Operator Hub repos. But TBH I'm not sure you can just change this easily for an existing release. So maybe you should open an issue instead of this PR and we can fix it when added the next Stirmzi version there.

Kartikey-Mishra1:
@scholzj I have raised a bug for this https://github.com/strimzi/strimzi-kafka-operator/issues/8413

scholzj:
@Kartikey-Mishra1 Great, thanks a lot.

PR #8417, opened 2023-04-19
Set pss-defaults compatible with `restricted`

jcpunk:
### Type of change

- Enhancement / new feature

### Description

Set the security context compatible with https://kubernetes.io/docs/concepts/security/pod-security-standards/#restricted

That is the current hardening best practices and shouldn't interfere with the `operator`.  This PR does not cover setting similar fields as a default on the CRD resources.

### Checklist

- [ ] Write tests
- [x] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/packaging/helm-charts/helm3/strimzi-kafka-operator/README.md
+++ b/packaging/helm-charts/helm3/strimzi-kafka-operator/README.md
@@ -110,9 +110,9 @@ the documentation for more details.
 | `fullReconciliationIntervalMs`       | Full reconciliation interval in milliseconds | 120000                                            |
 | `operationTimeoutMs`                 | Operation timeout in milliseconds         | 300000                                               |
 | `operatorNamespaceLabels`            | Labels of the namespace where the operator runs | `nil`                                          |
-| `podSecurityContext`                 | Cluster Operator pod's security context    | `nil`                                               |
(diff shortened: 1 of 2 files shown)


scholzj (commented):
I don't think this is backwards compatible change. While the operator should work under these rules, they can conflict with the various environments where the users use it and result in the deployment being rejected (it is quite common to have tooling which injects the right context instead of having it in the deployment itself, it is also quite common to allow only specific user or group IDs etc.). 

So I think using these settings for your self should work fine with the operator if you rely on the Kubernetes profile. But it should not be the new default.

----

PS: Unrelated to the PR, but if you want the restricted mode for the operands, you should check the Pod Security Provider plugins: https://strimzi.io/blog/2022/09/09/configuring-security-context-in-pods-managed-by-strimzi/

jcpunk:
Thanks for the link to the blog post!

I was thinking folks injecting the settings would be doing so via helm values, so their changes would persist.  I don't really know much about the operator or how folks might want to customize its security values...

Based on the blog post, it looks like compatibility with `restricted` by default for the applications is a future item.

scholzj:
> I was thinking folks injecting the settings would be doing so via helm values, so their changes would persist. I don't really know much about the operator or how folks might want to customize its security values...

You could of course configure it in the custom resources as well -> and if you use Helm to create the custom resources, it is probably reasonably easy. But especially if you have many YAMLs, this way you can easily control it centrally as well. (And you can also write your own custom provider plugin if you have some special requirements - there is another blog post about that). But just to be clear -> this is only about the operands and not about the context for Cluster Operator.

PR #8419, opened 2023-04-20
Consistent logging of the rolling reasons

ppatierno:
This trivial PR just makes consistent the logging of the rolling reasons across different components which could run a roll of pods (ZooKeeper, Kafka, EO, ...).
It takes the `KafkaRoller` as a reference (just moving from singular to plural with "Reasons", as they could be more than one).

### DIFF
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/CaReconciler.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/CaReconciler.java
@@ -447,7 +447,7 @@ public class CaReconciler {
         return deploymentOperator.getAsync(reconciliation.namespace(), deploymentName)
                 .compose(dep -> {
                     if (dep != null) {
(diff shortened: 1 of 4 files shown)


scholzj (commented):
TBH, I find this format weird. You have a complete sentence `Rolling pod {}.` including the `.` ... And then you have what is not a sentence `Reasons: {}`.

I always thought I would prefer one style for all log messages (including either using `.` everywhere or nowhere), but it did not seemed practical to try to align them. But here it is really weird as you mix the styles in a single log message. So to be honest using everywhere `Rolling pod {} due to {}` would seem to make more sense to me.

I don't really care that much about it -> so if you wanna merge this form, I'm fine to approve it if you want. Just feels a bit weird.

ppatierno:
I have no strong opinion I was just trying to make everything consistent with one shape. I can live with "Rolling Pod {} due to {}" and "Rolling Deployment {} due to {}" if you like.

scholzj:
That would seem better to me at least.

scholzj (approved):
Thanks.

im-konge (approved):

PR #8423, opened 2023-04-20
docs(ssl): removes repeated ssl config examples and cleans up config descriptions

PaulRMellor:
**Documentation**

- Removes repeated example configuration for `ssl` TLS versions and cipher suites
- Streamlines and makes consistent the `config` sections for components in the schema reference, and removes the ssl config as we lonk to the common `ssl` config section

### Checklist

- [ ] Write tests
- [x] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/documentation/api/io.strimzi.api.kafka.model.KafkaBridgeConsumerSpec.adoc
+++ b/documentation/api/io.strimzi.api.kafka.model.KafkaBridgeConsumerSpec.adoc
@@ -16,10 +16,16 @@ However, Strimzi takes care of configuring and managing options related to the f
 * Security (encryption, authentication, and authorization)
 * Consumer group identifier
 
-See the description of the `config` property for a list of prefixes for options you cannot change.
+Properties with the following prefixes cannot be set:
+
+* `bootstrap.servers`
+* `group.id`
+* `sasl.`
+* `security.`
+* `ssl.` 
 
 If the `config` property contains an option that cannot be changed, it is disregarded, and a warning message is logged to the Cluster Operator log file.
-All other supported options are forwarded to Kafka Bridge, including the following exception to the options configured by Strimzi:
+All other supported options are forwarded to Kafka Bridge, including the following exceptions to the options configured by Strimzi:
 
 * Any `ssl` configuration for xref:con-common-configuration-ssl-reference[supported TLS versions and cipher suites]
 
--- a/documentation/api/io.strimzi.api.kafka.model.KafkaBridgeConsumerSpec.adoc
+++ b/documentation/api/io.strimzi.api.kafka.model.KafkaBridgeConsumerSpec.adoc
@@ -40,5 +40,5 @@ spec:
 ----
 
 IMPORTANT: The Cluster Operator does not validate keys or values in the `config` object.
-If an invalid configuration is provided, the Kafka Bridge cluster might not start or might become unstable.
+If an invalid configuration is provided, the Kafka Bridge deployment might not start or might become unstable.
 In this case, fix the configuration so that the Cluster Operator can roll out the new configuration to all Kafka Bridge nodes.
--- a/documentation/api/io.strimzi.api.kafka.model.KafkaBridgeConsumerSpec.adoc
+++ b/documentation/api/io.strimzi.api.kafka.model.KafkaBridgeConsumerSpec.adoc
@@ -6,24 +6,22 @@ The values can be one of the following JSON types:
 * Number
 * Boolean
 
-You can specify and configure the options listed in the {ApacheKafkaConsumerConfig} with the exception of those options which are managed directly by Strimzi.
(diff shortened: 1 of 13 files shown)


documentation/api/io.strimzi.api.kafka.model.KafkaBridgeConsumerSpec.adoc
```diff
@@ -6,24 +6,22 @@ The values can be one of the following JSON types:
...
 
-You can specify and configure the options listed in the {ApacheKafkaConsumerConfig} with the exception of those options which are managed directly by Strimzi.
-Specifically, all configuration options with keys equal to or starting with one of the following strings are forbidden:
+*Exceptions*
 
-* `ssl.`
-* `sasl.`
-* `security.`
-* `bootstrap.servers`
-* `group.id`
```
scholzj, line 16: I would probably keep this -> I think it helps some of the users to see the exact options. ... at least unless the list is somewhere else as well.

documentation/api/io.strimzi.api.kafka.model.KafkaBridgeProducerSpec.adoc
```diff
@@ -6,23 +6,22 @@ The values can be one of the following JSON types:
...
 * Boolean
 
-You can specify and configure the options listed in the {ApacheKafkaproducerConfig} with the exception of those options which are managed directly by Strimzi.
-Specifically, all configuration options with keys equal to or starting with one of the following strings are forbidden:
+*Exceptions*
 
-* `ssl.`
-* `sasl.`
-* `security.`
-* `bootstrap.servers`
```
scholzj, line 15: Same as above ... and everywhere else.

documentation/api/io.strimzi.api.kafka.model.KafkaBridgeConsumerSpec.adoc
ppatierno: I agree with Jakub.

documentation/api/io.strimzi.api.kafka.model.KafkaBridgeConsumerSpec.adoc
```diff
@@ -38,9 +36,9 @@ spec:
...
       enable.auto.commit: true
-      ssl.cipher.suites: TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384
-      ssl.enabled.protocols: TLSv1.2
-      ssl.protocol: TLSv1.2
-      ssl.endpoint.identification.algorithm: HTTPS
     # ...
 ----
+
+IMPORTANT: The Cluster Operator does not validate keys or values in the `config` object.
+If an invalid configuration is provided, the Kafka Bridge cluster might not start or might become unstable.
```
ppatierno, line 43: Why we say "cluster" here. I think it's misleading. Wouldn't it be better just "deployment" instead?

documentation/api/io.strimzi.api.kafka.model.KafkaBridgeProducerSpec.adoc
```diff
@@ -37,9 +36,9 @@ spec:
...
       delivery.timeout.ms: 300000
-      ssl.cipher.suites: TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384
-      ssl.enabled.protocols: TLSv1.2
-      ssl.protocol: TLSv1.2
-      ssl.endpoint.identification.algorithm: HTTPS
     # ...
 ----
+
+IMPORTANT: The Cluster Operator does not validate keys or values in the `config` object.
+If an invalid configuration is provided, the Kafka Bridge cluster might not start or might become unstable.
```
ppatierno, line 43: Ditto about "cluster" as above.

documentation/api/io.strimzi.api.kafka.model.KafkaBridgeConsumerSpec.adoc
PaulRMellor: Th exact options are in the table description for `config`, hence the sentence that replaces the list (_See the description of the `config` property for a list of prefixes for options you cannot change._) Not sure why we want to repeat the list here?

image

scholzj: I don't know - to me it seems a bit more readable as an exact list. From my point of view, the list you added instead of it has the same information in a less useful form. So I do not really follow the reasoning for this.

PaulRMellor: It's certainly more readable, but we're just maintaining the same list in two places. I did wonder if we should change the format in the table (@Description) or remove them from the tables, but both options would be reflected in the YAML etc.

Anyway, I've added the lists, with some updates too

ppatierno (approved):

Issue #7685, opened 2022-11-22
All Zookeeper pods are deleted when are rolled with invalid configuration


im-konge:
**Describe the bug**

As part of the [#7683](https://github.com/strimzi/strimzi-kafka-operator/issues/7683), there was also found one more thing about terminating the ZK pods. How it was done in #7421 , when the configuration cannot be fulfilled, it should just roll one ZK pod, to prevent loss of whole ZK cluster. But, instead of this behavior, each ZK pod is being deleted.

**To Reproduce**

Steps to reproduce the behavior are same as for #7683 :
1. Deploy CO
2. Deploy Kafka cluster with 3 Kafka pods and 3 ZK pods
3. Edit Zookeeper part of the Kafka CR, setting the invalid configuration
4. Check the rolling update of Zookeeper pods

**Expected behavior**

Expected behavior is, that only one ZK pod will be rolled and if the roll will not be successful, other ZK pods will not be rolled until the issue (invalid configuration) will be fixed.

**Environment (please complete the following information):**

- Strimzi version: main
- Installation method: YAML
- Kubernetes cluster: Kubernetes 1.19 and above
- Infrastructure: Minikube

**YAML files and logs**

`kubectl get pods`
```
NAME                                                 READY   STATUS    RESTARTS   AGE
my-cluster-61d4c97e-entity-operator-c85cbdb6-6grc4   3/3     Running   0          16m
my-cluster-61d4c97e-kafka-0                          1/1     Running   0          16m
my-cluster-61d4c97e-kafka-1                          1/1     Running   0          11m
my-cluster-61d4c97e-kafka-2                          1/1     Running   0          11m
my-cluster-61d4c97e-zookeeper-0                      0/1     Running   1          17m
```

kafka.yaml
 https://gist.github.com/im-konge/92cd4c9bf14fe63b2fde509e45ba16db

scholzj:
To reproduce this, the invalid configuration needs to be in the Pod spec => e.g. memory limit smaller then memory request. Such invalid configuration will cause the Pod to be not created and not exist at all. That will make the ZooKeeperRoller to not take these _non-existent_ pods into account and roll the remaining running Pods into the same state as well until the whole ZooKeeper cluster is gone.

scholzj:
Triaged on 12.1.2023: Should be tracked and fixed.

fvaleri:
https://github.com/strimzi/strimzi-kafka-operator/pull/7826

PR #8429, opened 2023-04-21
Consider missing pods when rolling the ZooKeeper cluster

scholzj:
### Type of change

- Bugfix

### Description

As described in #7685, today, when a Pod is completely missing (e.g. because the Pod YAML is invalid - this could happen by invalid resources in the past and probably still can today for example due to misconfigured affinity etc.), the ZooKeeper Roller didn't consider it in the rolling process. As a result, it could have taken the whole ZooKeeper cluster down one by one by rolling the issue to next pod in every reconciliation.

This PR updates the ZooKeeper Roller to consider missing pods as well. If the pod is completely missing, it will first wait for it to get ready. Then it will consider unready pods (if there are any) and only last the ready pods will be considered.

Thanks to this, if some configuration issue makes to pod missing, the operator will not roll this out to more pods and instead will stop at the first one.

### Checklist

- [x] Write tests

### DIFF
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/CaReconciler.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/CaReconciler.java
@@ -374,8 +374,10 @@ public class CaReconciler {
     }
 
     /**
-     * Finds the number of current ZooKeeper replicas. We need to find the current number because if new trust should be
-     * rolled out while scaling event is happening, we have to roll out the trust only to the existing pods.
+     * If we need to roll the ZooKeeper cluster to roll out the trust to a new CA private key (i.e. when a CA private
+     * key is replaced), we need to know what the current number of ZooKeeper nodes is. Getting it from the Kafka custom
+     * resource might not be good enough if a scale-up /scale-down is happening at the same time. So we get the
+     * StrimziPodSet and find out the correct number of ZooKeeper nodes from it.
      *
      * @return  Current number of ZooKeeper replicas
      */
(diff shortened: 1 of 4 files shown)


cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/CaReconciler.java
```diff
@@ -372,8 +373,36 @@ Future<Void> rollingUpdateForNewCaKey() {
         }
     }
 
-    // ZooKeeper is rolled only for new Cluster CA key
-    private Future<Void> maybeRollZookeeper(RestartReasons podRestartReasons) {
+    /**
+     * Finds the number of current ZooKeeper replicas. We need to find the current number because if new trust should be
+     * rolled out while scaling event is happening, we have to roll out the trust only to the existing pods.
```
ppatierno, line 378: can you be clear about what's the "trust" we are talking about here? I could think about trusting a new cluster CA but trying to understand its relation with the subject of this PR (the description doesn't mention any "trust").

scholzj: I improved the Javadoc comment here. Please have a look if it is better now.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/CaReconciler.java
```diff
@@ -372,8 +373,38 @@ Future<Void> rollingUpdateForNewCaKey() {
         }
     }
 
-    // ZooKeeper is rolled only for new Cluster CA key
-    private Future<Void> maybeRollZookeeper(RestartReasons podRestartReasons) {
+    /**
+     * If we need to roll the ZooKeeper cluster to roll out the trust to a new CA private key (i.e. when a CA private
```
ppatierno, line 377: "to roll out the trust to a new CA private key (i.e. when a CA private key is replaced)" ... I guess it should be "to roll out the trust to a new CA certificate (i.e. when a CA private key is replaced) because you are "trusting" the new certificate (coming because of a new key) but not the new key itself.

scholzj: Better?

ppatierno (approved):
Yes, thanks! LGTM now!

PR #8432, opened 2023-04-21
[systemtest] Documentation generator for tests

im-konge:
### Type of change

- Enhancement / new feature

### Description

This PR adds a generator for the STs documentation, which is present in the new module - `systemtest-doc`.
The generator, based on the parameters, generates Markdown file for each test class and for each method adds record to the file. To achieve that, new annotations are added -> `@TestDoc`, `@TestDoc.Step`, `@TestDoc.Usecase`, `@TestDoc.Desc`.

Description of each annotation and how the generation works are mentioned in the `README.md` file inside the `systemtest-doc` module.

Additionally, this PR adds new profile to the `systemtest` module, which help us with copying all the needed dependencies to the `target` file of the module, that can be then added to the classpath of the Java exec command.

For this PR, I added an example to `ListenersST#testSendMessagesPlainAnonymous` and the generated file can be seen in `systemtest-doc/documentation/io/strimzi/systemtest/kafka/listeners/ListenersST.md`.

### Checklist

- [ ] Write tests
- [X] Update documentation


scholzj:
Wouldn't a separate repo be better for this?

pom.xml
```diff
@@ -268,6 +269,12 @@
                 <artifactId>kafka-oauth-keycloak-authorizer</artifactId>
                 <version>${strimzi-oauth.version}</version>
             </dependency>
+            <dependency>
+                <groupId>io.strimzi</groupId>
+                <artifactId>systemtest-doc</artifactId>
+                <version>${project.version}</version>
+                <scope>compile</scope>
+            </dependency>
```
scholzj, line 277: Is this really needed if it is part of this repo?

scholzj (commented):
How do you expect this to work? Where do you expect the generted files to be? How do you check if they are up to date etc.?

im-konge:
> Wouldn't a separate repo be better for this?

For the generator or the ST documentation?

I was thinking about it and the generator needs the built classes of the `systemtest` module + all the dependencies, so I decided to keep it in this repository.

> How do you expect this to work?

The build is not forced during the `mvn clean install`, by which it would do some kind of cyclic dependency.
When we edit the documentation of test, we need to manually generate it (I guess it is similarly done in case of the `api` module, but correct me if I'm wrong).

>  Where do you expect the generted files to be?

Currently it is generated inside the `systemtest-doc` module into the `documentation` directory.

> How do you check if they are up to date etc.?

Checking if there are changes in the test that is already documented would be non-trivial FMPOV. Because there are lot of things that can happen and that we would have to watch. Instead, we can add some check that would print some information that there are changes to the test and that the contributor should update the documentation and re-generate it. In the case of new tests, it should be relatively easy to make the build (or some PR check) fail, with an information that the test doesn't contain the documentation.

pom.xml
im-konge: Will check it, thanks :)

jankalinic (approved):

scholzj:
> For the generator or the ST documentation?

For the generator. The ST documentation has to live with the STs. I think there are several reasons for it:
* The lifecycle of the generator does not seem to be tied to Strimzi releases
* It seems like something usable outside this repo => for example in the Drain Cleaner repo, but possible even outside of Strimzi
* It will help to keep the build times in check as well
* It can be probably easily integrated for example as a Maven plugin or similar

> Checking if there are changes in the test that is already documented would be non-trivial FMPOV. Because there are lot of things that can happen and that we would have to watch. Instead, we can add some check that would print some information that there are changes to the test and that the contributor should update the documentation and re-generate it. In the case of new tests, it should be relatively easy to make the build (or some PR check) fail, with an information that the test doesn't contain the documentation.

If you plan to commit the docs into the repo, you should have some checks for it. Similar to how we check the files in docs and packaging for example.

im-konge:
@scholzj yeah that makes sense. Should I create a proposal for adding a new repository to the organization?

scholzj:
You can. But I would even consider having it outside of Strimzi. This seems to be generally useful - and assuming you checked and there is no similar tool that already exists, I think it might be interesting even outside of Strimzi.

im-konge:
Okay, I will move it outside of this repository and keep the documentation and the build here. I'll just convert it to draft for now.

Thanks! :)

scholzj:
Discussed on the community call on 4th May: Should be closed and will be reopened later.

PR #8436, opened 2023-04-24
Remove heartbeat connector from Mirror Maker 2 examples

scholzj:
### Type of change

- Task

### Description

This PR removes the `heartbeat` connector from out Mirror Maker 2 examples. The heartbeat connector in MM2 does not work really well in our examples:
* It connects directly to the target cluster, so it is not really mirrored and does not add much value
* The purpose does not seem clear to the users. They do not really understand that it is optional or how to use it. They often think that this is some mandatory part without which MM2 will not work.
* It has some value in active-active mirroring when it is mirrored. But the user experience seems weird since you configure it in the _wrong_ `KafkaMirrorMaker2` resource

This PR impacts only the examples. It is still part of the code and users can enable it. It also remains to be documented. We should aim to improve the documentation and better explain when it might be useful. But that is out of scope for this PR.

### Checklist


packaging/examples/security/scram-sha-512-auth/mirror-maker-2.yaml
```diff
@@ -202,7 +202,7 @@ spec:
           name: "*"
         operations:
           - Read
-      # MirrorHeartbeatConnector
+      # MirrorHeartbeatConnector => needed only if the connector is enabled
```
mimaison, line 205: Should we remove that too?
Same below

scholzj: I would not remove it -> I think it is useful to track this. But we can for example consider commenting out the ACLs.

mimaison (approved):

PR #8439, opened 2023-04-26
ST: Increase memory for CO to 512Mi in STs on FIPS clusters

Frawless:
### Type of change

- Enhancement / new feature

### Description

Cluster Operator memory is now set to 512Mi when fips are enabled on clusters.

It also adds a check for https://github.com/strimzi/strimzi-kafka-operator/pull/8400

### Checklist

### DIFF
--- a/systemtest/src/main/java/io/strimzi/systemtest/Constants.java
+++ b/systemtest/src/main/java/io/strimzi/systemtest/Constants.java
@@ -455,4 +455,12 @@ public interface Constants {
         ROLE, PATH_TO_LEASE_ROLE,
         ROLE_BINDING, PATH_TO_LEASE_ROLE_BINDING
     );
+
+    /**
+     * Cluster Operator resources config
+     */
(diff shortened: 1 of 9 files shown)


scholzj (commented):
I think we need to do this only where needed - e.g. on FIPS. We do not want to use this as a general setting.

Frawless:
> I think we need to do this only where needed - e.g. on FIPS. We do not want to use this as a general setting.

I agree, I just didn't want to add another env variable. This config basically affects everything except azure (unless user set 
`RESOURCE_ALLOCATION_STRATEGY` )

So should I introduce a new env variable, something like `CO_HIGHER_MEMORY` and when it will be set to `true` just change the memory config?

scholzj:
Don't you already have some FIPS indicators? Aren't some FIPS tests disabled?

Frawless:
Yes, I kinda forgot about this. I will update the PR

scholzj (approved):
LGTM assuming it works.

PR #8455, opened 2023-04-28
ST: Check that operatorVersion is emptyString instead of null after fixing spotbugs

Frawless:
### Type of change

- Bugfix

### Description

When we deploy CO via OLm we check the configuration against initialized values and because I changed the init value for operatorVersion to fix spotbugs error to empty string, the check should count with that as well. This PR fixes the bug when the check resulted in wrong value and subscription contained wrong values.

### Checklist


Frawless:
@strimzi-ci run tests --profile=acceptance --install-type=olm

Frawless:
@strimzi-ci run tests --profile=acceptance --install-type=olm

see-quick (approved):

Frawless:
The failure from jenkins is handled as part of https://github.com/strimzi/strimzi-kafka-operator/pull/8460

PR #8459, opened 2023-05-02
docs(mm2): deciding when to use the heartbeat connector

PaulRMellor:
**Documentation**

Use of the MM2 heartbeat connector is not always needed.
This update introduces a new section to the doc called **Deciding when to use the heartbeat connector**.
The section explains what the heartbeat connector does, and why you might want to use it or not use it.

### Checklist

- [ ] Write tests
- [ ] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/documentation/modules/configuring/con-config-mirrormaker2-replication.adoc
+++ b/documentation/modules/configuring/con-config-mirrormaker2-replication.adoc
@@ -3,7 +3,7 @@
 // assembly-config-mirrormaker2.adoc
 
 [id='con-mirrormaker-{context}']
-= MirrorMaker 2 data replication
+= Replicating data using MirrorMaker 2
 
 [role="_abstract"]
 Data replication across clusters supports scenarios that require:
@@ -69,13 +69,19 @@ Because the synchronization is time-based, any switchover by consumers to a pass
 
 NOTE: If you have an application written in Java, you can use the `RemoteClusterUtils.java` utility to synchronize offsets through the application. The utility fetches remote offsets for a consumer group from the `checkpoints` topic. 
 
-== Connectivity checks
+== Deciding when to use the heartbeat connector
 
-`MirrorHeartbeatConnector` emits _heartbeats_ to check connectivity between clusters.
+The heartbeat connector emits heartbeats to check connectivity between source and target Kafka clusters.
+An internal heartbeat topic is replicated from the source cluster, which means that the heartbeat connector must be connected to the source cluster. 
+A target cluster uses the heartbeat topic to perform the following checks:
 
-An internal `heartbeat` topic is replicated from the source cluster.
-Target clusters use the `heartbeat` topic to check the following:
+* Identify all source clusters it is mirroring data from
+* Check that the Kafka Connect worker managing connectivity between clusters is running
+* Verify that the source cluster is available
 
-* The connector managing connectivity between clusters is running
-* The source cluster is available
+You can use the heartbeat connector to monitor the liveness and latency of the mirroring process.
+This helps to make sure that the process is not stuck or has stopped for any reason. 
+However, using the heartbeat connector is not always necessary and can be left out of the MirrorMaker 2 configuration. 
+For example, if your deployment has low network latency or a small number of topics, you might prefer to monitor the mirroring process using log messages or other monitoring tools. 
+Heartbeats are enabled by default, but you can set the `emit.heartbeats.enabled` property of the heartbeat connector to `false` if you do not want to use them.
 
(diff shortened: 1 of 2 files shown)


documentation/modules/configuring/con-config-mirrormaker2-replication.adoc
```diff
@@ -69,13 +69,19 @@ Because the synchronization is time-based, any switchover by consumers to a pass
 
 NOTE: If you have an application written in Java, you can use the `RemoteClusterUtils.java` utility to synchronize offsets through the application. The utility fetches remote offsets for a consumer group from the `checkpoints` topic. 
 
-== Connectivity checks
+== Deciding when to use the heartbeat connector
 
-`MirrorHeartbeatConnector` emits _heartbeats_ to check connectivity between clusters.
+The heartbeat connector emits heartbeats to check connectivity between source and target Kafka clusters.
+An internal heartbeat topic is replicated from the source cluster, which means that the heartbeat connector must be connected to the source cluster. 
+A target cluster uses the heartbeat topic to perform the following checks:
```
mimaison, line 76: What about:
```suggestion
The heartbeat topic is located on the target cluster and allows to:
```

documentation/modules/configuring/con-config-mirrormaker2-replication.adoc
```diff
@@ -69,13 +69,19 @@ Because the synchronization is time-based, any switchover by consumers to a pass
...
 
-An internal `heartbeat` topic is replicated from the source cluster.
-Target clusters use the `heartbeat` topic to check the following:
+* Identify all source clusters it is mirroring data from
+* Check that the Kafka Connect worker managing connectivity between clusters is running
+* Verify that the source cluster is available
 
-* The connector managing connectivity between clusters is running
-* The source cluster is available
+You can use the heartbeat connector to monitor the liveness and latency of the mirroring process.
```
mimaison, line 82: I wonder if we remove `Check that the Kafka Connect worker managing connectivity between clusters is running` and merge the last bullet point and following sentence into something like: `Verify the liveness and latency of the mirroring process.`

mimaison (commented):
Thanks for the PR. I left a few suggestions

documentation/modules/configuring/con-config-mirrormaker2-replication.adoc
```diff
@@ -69,13 +69,19 @@ Because the synchronization is time-based, any switchover by consumers to a pass
...
+This helps to make sure that the process is not stuck or has stopped for any reason. 
+However, using the heartbeat connector is not always necessary and can be left out of the MirrorMaker 2 configuration. 
+For example, if your deployment has low network latency or a small number of topics, you might prefer to monitor the mirroring process using log messages or other monitoring tools. 
+Heartbeats are enabled by default, but you can set the `emit.heartbeats.enabled` property of the heartbeat connector to `false` if you do not want to use them.
```
scholzj, line 86: Not sure what are we trying to say with this. All the example were modified, so the Heartbeat connector is not enabled by default anymore.

documentation/modules/configuring/con-config-mirrormaker2-replication.adoc
```diff
@@ -69,13 +69,19 @@ Because the synchronization is time-based, any switchover by consumers to a pass
 
 NOTE: If you have an application written in Java, you can use the `RemoteClusterUtils.java` utility to synchronize offsets through the application. The utility fetches remote offsets for a consumer group from the `checkpoints` topic. 
 
-== Connectivity checks
+== Deciding when to use the heartbeat connector
```
scholzj, line 72: I agree with Mickael's comments -> but it seems good in general. However, as we removed the hearbeat connector from the examples -> do we also need to add an example of how to enable it to the docs?

mimaison (approved):

documentation/modules/configuring/con-config-mirrormaker2-replication.adoc
PaulRMellor: We show the heartbeat connector configuration in a later section, so I'm not sure we need it here.
I've intentionally kept the config out of this overview section so that it can be reused in rhel doc.
I did add this line though:

> You enable the heartbeat connector by including its configuration in the MirrorMaker 2 deployment configuration.

documentation/modules/configuring/con-config-mirrormaker2-replication.adoc
```diff
@@ -69,13 +69,18 @@ Because the synchronization is time-based, any switchover by consumers to a pass
...
+* Identify all source clusters it is mirroring data from
+* Verify the liveness and latency of the mirroring process
 
-* The connector managing connectivity between clusters is running
-* The source cluster is available
+This helps to make sure that the process is not stuck or has stopped for any reason. 
+You enable the heartbeat connector by including its configuration in the MirrorMaker 2 deployment configuration.
+However, using the heartbeat connector is not always necessary and can be left out of the MirrorMaker 2 configuration. 
+For example, if your deployment has low network latency or a small number of topics, you might prefer to monitor the mirroring process using log messages or other monitoring tools. 
+If heartbeats are enabled, you can set the `emit.heartbeats.enabled` property of the heartbeat connector to `false` if you do not want to use them.
```
scholzj, line 85: Shouldn't suggest to just removing the connector from the `KafkaMirrorMaker2` instead of setting some disable option? That would also correspond to the examples.

PaulRMellor: Yes. I've updated

PR #8464, opened 2023-05-02
Refactor informer handling in `UserController` and `StrimziPodSetController`

scholzj:
### Type of change

- Refactoring

### Description

This PR refactors the informer handling in the `UserController` and in the `StrimziPodSetController`. It applies the following changes:
* Uses Fabric8 to create _runnable_ informers => _runnable_ informers are not started by default and they are started only by us. That gives us better control over the lifecycle of the informer and allows us to configure things that are not configurable once the informer is started. This allows us also to actually enable the re-sync (which we configured before, but was not really used). 
* Adds synchronous stopping of the informers. Before this PR, we _requested_ the informers to stop, but did not wait for them to actually stop.
* It adds an exception handler to log any possible exception. In previous Strimzi versions, there were some issues reported when the informer stopped working. This might help us to debug it more if it happens again (although it might have been caused by Fabric8 bugs fixed in the meantime as well, so it might not reproduce anymore).
* It creates new `InformerUtils` method shared between the `StrimziPodSetController` and the `UserController` for some of the logic shared by both (`UserController` does not use Vert.x and the Resource Operator classes, so it could not be placed there).
* Refactors the event handlers in both controllers into separate inner classes to make the logic of starting the controllers shorter and easier to read what it does without studying the (sometimes) long event handler code.

This PR does not add any new tests as it just refactors the existing functionality - so it is covered by the existing tests.

### Checklist

### DIFF
--- a/operator-common/src/main/java/io/strimzi/operator/common/InformerUtils.java
+++ b/operator-common/src/main/java/io/strimzi/operator/common/InformerUtils.java
@@ -1,0 +1,56 @@
+/*
+ * Copyright Strimzi authors.
+ * License: Apache License 2.0 (see the file LICENSE or http://apache.org/licenses/LICENSE-2.0.html).
+ */
+package io.strimzi.operator.common;
+
+import io.fabric8.kubernetes.client.informers.SharedIndexInformer;
+
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
+
+/**
+ * Utilities for working with informers
+ */
+public class InformerUtils {
+    private static final ReconciliationLogger LOGGER = ReconciliationLogger.create(InformerUtils.class);
+
+    /**
+     * Logs exceptions in the informers to give us a better overview of what is happening.
+     *
+     * @param type          Type of the informer
+     * @param isStarted     Flag indicating whether the informer is already started
+     * @param throwable     Throwable describing the exception which occurred
+     *
+     * @return  Boolean indicating whether the informer should retry or not.
+     */
+    public static boolean loggingExceptionHandler(String type, boolean isStarted, Throwable throwable) {
+        LOGGER.warnOp("Caught exception in the " + type + " informer which is " + (isStarted ? "started" : "not started"), throwable);
+        // We always want the informer to retry => we just want to log the error
+        return true;
+    }
+
+    /**
+     * Synchronously stops one or more informers. It will stop them and then wait for the specified timeout for them to
+     * actually stop.
+     *
+     * @param timeoutMillis     Timeout in milliseconds for how long we will wait for each informer to stop
+     * @param informers         Informers which should be stopped.
+     */
+    public static void stopAll(long timeoutMillis, SharedIndexInformer<?>... informers) {
+        LOGGER.infoOp("Stopping informers");
+        for (SharedIndexInformer<?> informer : informers)    {
+            informer.stop();
+        }
+
+        try {
+            for (SharedIndexInformer<?> informer : informers)    {
+                informer.stopped().toCompletableFuture().get(timeoutMillis, TimeUnit.MILLISECONDS);
+            }
+        } catch (InterruptedException | TimeoutException | ExecutionException e) {
+            // We just log the error as we are anyway shutting down
+            LOGGER.warnOp("Interrupted while waiting for the informers to stop", e);
+        }
+    }
+}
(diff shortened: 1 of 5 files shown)


operator-common/src/main/java/io/strimzi/operator/common/InformerUtils.java
```diff
@@ -0,0 +1,56 @@
...
+     * Logs exceptions in the informers to give us a better overview of what is happening.
+     *
+     * @param type          Type of the informer
+     * @param isStarted     Flag indicating whether the informer is already started
+     * @param throwable     Throwable describing the exception which occurred
+     *
+     * @return  Boolean indicating whether the informer should retry or not.
+     */
+    public static boolean loggingExceptionHandler(String type, boolean isStarted, Throwable throwable) {
+        LOGGER.warnOp("Caught exception in the " + type + " informer which is " + (isStarted ? "started" : "not started"), throwable);
```
ppatierno, line 29: out of curiosity, what are the scenarios where the informer can throw an exception without being started?

ppatierno (approved):
LGTM. I left just a question.

operator-common/src/main/java/io/strimzi/operator/common/InformerUtils.java
tombentley, line 29: If no exception is expected and this exception indicates the non-delivery of an event it would seem to me that error, rather than warn, is more appropriate.

operator-common/src/main/java/io/strimzi/operator/common/InformerUtils.java
```diff
@@ -0,0 +1,56 @@
...
+            informer.stop();
+        }
+
+        try {
+            for (SharedIndexInformer<?> informer : informers)    {
+                informer.stopped().toCompletableFuture().get(timeoutMillis, TimeUnit.MILLISECONDS);
+            }
+        } catch (InterruptedException | TimeoutException | ExecutionException e) {
+            // We just log the error as we are anyway shutting down
+            LOGGER.warnOp("Interrupted while waiting for the informers to stop", e);
```
tombentley, line 53: Not necessarily interrupted.

operator-common/src/main/java/io/strimzi/operator/common/operator/resource/AbstractNamespacedResourceOperator.java
```diff
@@ -408,52 +409,58 @@ public Future<Void> deleteAsync(Reconciliation reconciliation, String namespace,
...
      *
      * @return          Informer instance
      */
-    public SharedIndexInformer<T> informer(String namespace)   {
-        if (ANY_NAMESPACE.equals(namespace))    {
-            return operation().inAnyNamespace().inform();
-        } else {
-            return operation().inNamespace(namespace).inform();
-        }
+    public SharedIndexInformer<T> informer(String namespace, long resyncInterval)   {
```
tombentley, line 420: `resyncIntervalMs`, just to be clearer.

user-operator/src/main/java/io/strimzi/operator/user/UserController.java
```diff
@@ -172,69 +169,27 @@ protected void stop() {
                 LOGGER.debugOp("Interrupted while stopping controller loop", e);
             }
         });
+
+        InformerUtils.stopAll(5_000L, userInformer, secretInformer);
```
tombentley, line 173: Is it not safer to stop the informers first, and wait for the queues to empty before stopping the controller?

operator-common/src/main/java/io/strimzi/operator/common/InformerUtils.java
```diff
@@ -0,0 +1,56 @@
...
+     * @return  Boolean indicating whether the informer should retry or not.
+     */
+    public static boolean loggingExceptionHandler(String type, boolean isStarted, Throwable throwable) {
+        LOGGER.errorOp("Caught exception in the " + type + " informer which is " + (isStarted ? "started" : "not started"), throwable);
+        // We always want the informer to retry => we just want to log the error
+        return true;
+    }
+
+    /**
+     * Synchronously stops one or more informers. It will stop them and then wait for the specified timeout for them to
```
tombentley, line 35:
```suggestion
     * Synchronously stops one or more informers. It will stop them and then wait for up to the specified timeout for each of them to
```

tombentley (approved):
Thanks.

user-operator/src/main/java/io/strimzi/operator/user/UserController.java
scholzj: Good idea. I added it there. Thanks.

Issue #8255, opened 2023-03-19
[systemtest] Create test class for TopicOperator and move all related tests there


im-konge:
Currently, most of the TO tests are in `KafkaST` (and other test suites).

We should create a test class specifically for TO, with the name `TopicOperatorST, and move all TO related tests there:
Example tests:

- `testForTopicOperator` - this test should be also renamed + should be updated with new way of checking resources etc.
- `testTopicWithoutLabels` - add stability wait + wait for periodic reconciliation of TO

PR #8475, opened 2023-05-04
[ST] Move and Improve testTopicWithoutLabels, from KafkaST to TopicST

henryZrncik:
### Type of change

- Refactoring

### Description

- Migration of  `testTopicWithoutLabels` from `KafkaST`to  `TopicST`
- addition of extra verification in the test that checks whether or not kafka topic is or is not created when `KafkaTopic` CR does not contain necessary label.

### Checklist

- [x] Write tests
- [x] Update documentation

fixes https://github.com/strimzi/strimzi-kafka-operator/issues/8255

### DIFF
--- a/systemtest/src/main/java/io/strimzi/systemtest/utils/kafkaUtils/KafkaTopicUtils.java
+++ b/systemtest/src/main/java/io/strimzi/systemtest/utils/kafkaUtils/KafkaTopicUtils.java
@@ -15,6 +15,7 @@ import io.strimzi.test.TestUtils;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 
+import java.time.Duration;
 import java.util.Arrays;
 import java.util.List;
 import java.util.Map;
@@ -226,25 +227,26 @@ public class KafkaTopicUtils {
 
     /**
      * Verifies that {@code absentTopicName} topic remains absent in {@code clusterName} Kafka cluster residing in {@code namespaceName},
+     * for two times {@code topicOperatorReconciliationSeconds} duration (in seconds) of Topic Operator reconciliation time,
      * by querying the cluster using kafka scripts from {@code queryingPodName} Pod.
      * @param namespaceName Namespace name
      * @param queryingPodName  the name of the pod to query Kafka topics from
      * @param clusterName name of Kafka cluster
-     * @param absentTopicName name of Kafka topic which should not be created.
+     * @param absentTopicName name of Kafka topic which should not be created
+     * @param topicOperatorReconciliationSeconds interval in seconds for Topic Operator to reconcile
      * @throws AssertionError in case topic is created
      */
-    public static void verifyUnchangedTopicAbsence(String namespaceName, String queryingPodName, String clusterName, String absentTopicName) {
+    public static void verifyUnchangedTopicAbsence(String namespaceName, String queryingPodName, String clusterName, String absentTopicName, int topicOperatorReconciliationSeconds) {
 
-        long poolInterval = Constants.POLL_INTERVAL_FOR_RESOURCE_READINESS; // repeat every second
-        long duration = Constants.RECONCILIATION_INTERVAL; // repeat for ~30sec
-        long startTime = System.currentTimeMillis();
-        long endTime = startTime + duration;
-        LOGGER.info("Verify absence of topic '{}' in listed kafka topics for next {} seconds in Namespace/{}", absentTopicName, duration / 1000, namespaceName);
+        long reconciliationDuration = Duration.ofSeconds(topicOperatorReconciliationSeconds).toMillis();
+        long endTime = System.currentTimeMillis() + 2 * reconciliationDuration;
+
+        LOGGER.info("Verify absence of topic '{}' in listed kafka topics for next {} seconds in Namespace/{}", absentTopicName, reconciliationDuration / 1000, namespaceName);
 
         while (System.currentTimeMillis() < endTime) {
             assertThat(KafkaCmdClient.listTopicsUsingPodCli(namespaceName, queryingPodName, KafkaResources.plainBootstrapAddress(clusterName)), not(hasItems(absentTopicName)));
             try {
-                Thread.sleep(poolInterval);
+                Thread.sleep(Constants.POLL_INTERVAL_FOR_RESOURCE_READINESS);
             } catch (InterruptedException e) {
                 e.printStackTrace();
             }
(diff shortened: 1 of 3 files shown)


systemtest/src/main/java/io/strimzi/systemtest/utils/kafkaUtils/KafkaTopicUtils.java
```diff
@@ -219,4 +223,31 @@ public static void waitForTopicWithPrefixDeletion(String namespaceName, String t
...
+     * @param namespaceName Namespace name
+     * @param queryingPodName  the name of the pod to query Kafka topics from
+     * @param clusterName name of Kafka cluster
+     * @param absentTopicName name of Kafka topic which should not be created.
+     * @throws AssertionError in case topic is created
+     */
+    public static void verifyUnchangedTopicAbsence(String namespaceName, String queryingPodName, String clusterName, String absentTopicName) {
+
+        long poolInterval = Constants.POLL_INTERVAL_FOR_RESOURCE_READINESS; // repeat every second
+        long duration = Constants.RECONCILIATION_INTERVAL; // repeat for ~30sec
```
im-konge, line 239: I think you can use the constants directly

systemtest/src/main/java/io/strimzi/systemtest/utils/kafkaUtils/KafkaTopicUtils.java
```diff
@@ -219,4 +223,31 @@ public static void waitForTopicWithPrefixDeletion(String namespaceName, String t
...
+        LOGGER.info("Verify absence of topic '{}' in listed kafka topics for next {} seconds in Namespace/{}", absentTopicName, duration / 1000, namespaceName);
+
+        while (System.currentTimeMillis() < endTime) {
+            assertThat(KafkaCmdClient.listTopicsUsingPodCli(namespaceName, queryingPodName, KafkaResources.plainBootstrapAddress(clusterName)), not(hasItems(absentTopicName)));
+            try {
+                Thread.sleep(poolInterval);
+            } catch (InterruptedException e) {
+                e.printStackTrace();
+            }
+        }
```
im-konge, line 251: You should use `TestUtils.waitFor` method instead of this

henryZrncik: yeah I think that is a bit different thing that i want to test, 
whereas waitFor implies that I am waiting for something (to occur) in this case I am simply making sure that something really remains the way it is (kafka topic "topic-without-label" topic really is not created and never have been from KafkaTopic CR with same name). So this would 
- either force me to either wait for timeout (waiting for it to exist) 
- or pass immediately (waiting for it to not exist) 
- or i would modify it to stability like check which would be actually pointless as i don't want it only to be stable, instead stable immediately => something like unchanged. 

But I am really open to change this implementation.

im-konge: As you mentioned, we are using it for the stability waits also (f.e. `waitUntilPodStabilityReplicasCount`), but yeah, in that case when you don't want to use it as "stability wait", I guess you can keep it

systemtest/src/main/java/io/strimzi/systemtest/utils/kafkaUtils/KafkaTopicUtils.java
henryZrncik: default is 120 seconds (AFAIK) but i explicitly set 10 seconds in  kafka spec for TO reconciliation, so I can pass it (10sec pool interval) as well.

systemtest/src/main/java/io/strimzi/systemtest/utils/kafkaUtils/KafkaTopicUtils.java
```diff
@@ -219,4 +224,32 @@ public static void waitForTopicWithPrefixDeletion(String namespaceName, String t
...
+
+        long reconciliationDuration = Duration.ofSeconds(topicOperatorReconciliationSeconds).toMillis();
+        long endTime = System.currentTimeMillis() + 2 * reconciliationDuration;
+
+        LOGGER.info("Verify absence of topic '{}' in listed kafka topics for next {} seconds in Namespace/{}", absentTopicName, reconciliationDuration / 1000, namespaceName);
+
+        while (System.currentTimeMillis() < endTime) {
+            assertThat(KafkaCmdClient.listTopicsUsingPodCli(namespaceName, queryingPodName, KafkaResources.plainBootstrapAddress(clusterName)), not(hasItems(absentTopicName)));
+            try {
+                Thread.sleep(Constants.POLL_INTERVAL_FOR_RESOURCE_READINESS);
```
henryZrncik, line 249: I removed 2 local variables,  (`constants.POLL_INTERVAL_FOR_RESOURCE_READINESS) is used directly instead of naming it + `startTime` is removed. 

also I pass here  `topicOperatorReconciliationSeconds` which specifies TO reconciliation time in seconds. (I can obtain this value easily as I am the one setting it in test which call this method).

henryZrncik:
@strimzi-ci run tests --testcase=TopicST

Frawless (approved):

systemtest/src/main/java/io/strimzi/systemtest/utils/kafkaUtils/KafkaTopicUtils.java
```diff
@@ -219,4 +224,32 @@ public static void waitForTopicWithPrefixDeletion(String namespaceName, String t
                 }
             });
     }
+
+    /**
+     * Verifies that {@code absentTopicName} topic remains absent in {@code clusterName} Kafka cluster residing in {@code namespaceName},
+     * for two times {@code topicOperatorReconciliationSeconds} duration (in seconds) of Topic Operator reconciliation time,
+     * by querying the cluster using kafka scripts from {@code queryingPodName} Pod.
+     * @param namespaceName Namespace name
```
see-quick, line 232: an empty line between the description and the first `@...` param.

systemtest/src/test/java/io/strimzi/systemtest/operators/topic/TopicST.java
```diff
@@ -580,6 +587,77 @@ void testKafkaTopicChangingMinInSyncReplicas(ExtensionContext extensionContext)
...
+     *  - kafka-topic
+     *  - labels
+     */
+    @ParallelNamespaceTest
+    @KRaftNotSupported("TopicOperator is not supported by KRaft mode and is used in this test class")
+    void testTopicWithoutLabels(ExtensionContext extensionContext) {
+        final String namespaceName = StUtils.getNamespaceBasedOnRbac(namespace, extensionContext);
+        final String clusterName = mapWithClusterNames.get(extensionContext.getDisplayName());
+        final String scraperName = mapWithScraperNames.get(extensionContext.getDisplayName());
+        final String kafkaTopicName = "topic-without-labels";
```
see-quick, line 615: Here you should you `TestStorage` class and then retrieve all these fields from that instance. (I know that you just copy-pasted such code but It just small change, which does not affect anything)

see-quick (approved):
Apart from my 2 comments. Looks good and thanks for it 👍

Frawless (approved):

PR #8490, opened 2023-05-05
docs(logging): adds log levels for specific classes to the logging examples

PaulRMellor:
**Documentation**

 includes examples of how to set log levels for specific classes or loggers in the example logging configuration YAML.

### Checklist

- [ ] Write tests
- [x] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/documentation/api/io.strimzi.api.kafka.model.CruiseControlSpec.adoc
+++ b/documentation/api/io.strimzi.api.kafka.model.CruiseControlSpec.adoc
@@ -264,10 +264,16 @@ spec:
       type: inline
       loggers:
         rootLogger.level: INFO
-        logger.com.linkedin.kafka.cruisecontrol.executor.Executor: TRACE
-        logger.com.linkedin.kafka.cruisecontrol.analyzer.GoalOptimizer: DEBUG
+        logger.exec.name: com.linkedin.kafka.cruisecontrol.executor.Executor # <1>
+        logger.exec.level: TRACE # <2>
+        logger.go.name: com.linkedin.kafka.cruisecontrol.analyzer.GoalOptimizer # <3>
+        logger.go.level: DEBUG # <4>
     # ...
 ----
+<1> Creates a logger for the Cruise Control `Executor` class.
+<2> Sets the logging level for the `Executor` class.
+<3> Creates a logger for the Cruise Control `GoalOptimizer` class.
+<4> Sets the logging level for the `GoalOptimizer` class.
 
 NOTE: When investigating an issue with Cruise Control, it's usually sufficient to change the `rootLogger` to `DEBUG` to get more detailed logs. However, keep in mind that setting the log level to `DEBUG` may result in a large amount of log output and may have performance implications.
 
--- a/documentation/api/io.strimzi.api.kafka.model.EntityTopicOperatorSpec.adoc
+++ b/documentation/api/io.strimzi.api.kafka.model.EntityTopicOperatorSpec.adoc
@@ -40,10 +40,19 @@ spec:
         type: inline
         loggers:
           rootLogger.level: INFO
-          logger.io.strimzi.operator.topic.ZkTopicStore: TRACE
-          logger.io.strimzi.operator.topic.TopicOperator: DEBUG
+          logger.top.name: io.strimzi.operator.topic # <1>
+          logger.top.level: DEBUG # <2> 
+          logger.toc.name: io.strimzi.operator.topic.TopicOperator # <3>
+          logger.toc.level: TRACE # <4>
+          logger.clients.level: DEBUG # <5>
   # ...
 ----
+<1> Creates a logger for the `topic` package.
+<2> Sets the logging level for the `topic` package.
+<3> Creates a logger for the `TopicOperator` class.
+<4> Sets the logging level for the `TopicOperator` class.
+<5> Changes the logging level for the default `clients` logger. The `clients` logger is part of the logging configuration provided with Strimzi.
+By default, it is set to `INFO`.   
 
 NOTE: When investigating an issue with the operator, it's usually sufficient to change the `rootLogger` to `DEBUG` to get more detailed logs. 
 However, keep in mind that setting the log level to `DEBUG` may result in a large amount of log output and may have performance implications.
--- a/documentation/api/io.strimzi.api.kafka.model.EntityUserOperatorSpec.adoc
+++ b/documentation/api/io.strimzi.api.kafka.model.EntityUserOperatorSpec.adoc
@@ -40,10 +40,20 @@ spec:
         type: inline
         loggers:
           rootLogger.level: INFO
-          logger.io.strimzi.operator.user.UserController: TRACE
-          logger.io.strimzi.operator.user.HealthCheckAndMetricsServer: DEBUG
+          logger.uop.name: io.strimzi.operator.user # <1>
+          logger.uop.level: DEBUG # <2> 
+          logger.abstractcache.name: io.strimzi.operator.user.operator.cache.AbstractCache # <3>
+          logger.abstractcache.level: TRACE # <4>
+          logger.jetty.level: DEBUG # <5>
+          
   # ...
 ----
+<1> Creates a logger for the `user` package.
+<2> Sets the logging level for the `user` package.
+<3> Creates a logger for the `AbstractCache` class.
+<4> Sets the logging level for the `AbstractCache` class.
+<5> Changes the logging level for the default `jetty` logger. The `jetty` logger is part of the logging configuration provided with Strimzi.
+By default, it is set to `INFO`.  
 
 NOTE: When investigating an issue with the operator, it's usually sufficient to change the `rootLogger` to `DEBUG` to get more detailed logs. However, keep in mind that setting the log level to `DEBUG` may result in a large amount of log output and may have performance implications.
 
--- a/documentation/api/io.strimzi.api.kafka.model.KafkaBridgeSpec.adoc
+++ b/documentation/api/io.strimzi.api.kafka.model.KafkaBridgeSpec.adoc
@@ -10,9 +10,10 @@ Configuration options relate to:
 
 [id='property-kafka-bridge-logging-{context}']
 === `logging`
+
 Kafka Bridge has its own configurable loggers:
 
-* `logger.bridge`
+* `rootLogger.level`
 * `logger.__<operation-id>__`
 
 You can replace `_<operation-id>_` in the `logger.__<operation-id>__` logger to set log levels for specific operations:
@@ -76,7 +77,7 @@ spec:
   logging:
     type: inline
     loggers:
-      logger.bridge.level: INFO
+      rootLogger.level: INFO
       # enabling DEBUG just for send operation
       logger.send.name: "http.openapi.operation.send"
       logger.send.level: DEBUG


scholzj (commented):
Did you tried all of them? Do they work? I'm a bit surprised the format is the same for all compoenents regardless the logging framework they use.

PaulRMellor:
> Did you tried all of them? Do they work? I'm a bit surprised the format is the same for all compoenents regardless the logging framework they use.

I'm checking again. Most seem okay, but I've asked Kyle about the CC examples. I've set to _draft_ until confirmed.

PaulRMellor:
The latest commit adds better examples.
low-level logs removed from the operators and Cruise Control. Speaking with @fvaleri , he thinks that just changing the level for the root logger is enough. 
Also added a note about switching to `DEBUG`

scholzj:
> low-level logs removed from the operators and Cruise Control. Speaking with @fvaleri , he thinks that just changing the level for the root logger is enough.

I think they are useful for operators. So I would keep them.

fvaleri:
> > low-level logs removed from the operators and Cruise Control. Speaking with @fvaleri , he thinks that just changing the level for the root logger is enough.
> 
> I think they are useful for operators as well as for Cruise Control. So I would keep them.

In general, I would keep them, but setting at info level. That way, we show some useful logging configurations without the risk that some user will cut and paste into their CR without modifications. For example, log4j.logger.kafka.request.logger is very verbose at DEBUG level, but can be useful in some situations.

documentation/api/io.strimzi.api.kafka.model.KafkaMirrorMakerSpec.adoc
```diff
@@ -45,9 +47,13 @@ spec:
     type: inline
     loggers:
       mirrormaker.root.logger: "INFO"
```
fvaleri, line 49: This is for MM1, which is deprecated. This is a minor change so it's fine, but I think we should avoid doing non-security changes.

fvaleri (commented):
Thanks @PaulRMellor.

Configurations seems to work fine. Left some comments.

scholzj:
> In general, I would keep them, but setting at info level. That way, we show some useful logging configurations without the risk that some user will cut and paste into their CR without modifications. For example, log4j.logger.kafka.request.logger is verbose at DEBUG level, but can be useful in some situations.

@fvaleri I'm not sure I get it. Why not apply that logic for the other examples as well then? E.g. the Kafka examples and keep `log4j.logger.kafka.request.logger` at `INFO` as well? 

This part of the docs should not say _this is the logging you should use_ => if that is the case we should change the code (but in general we should IMHO follow Kafka defaults as a Kafka's recommendation). It should show how you change them. I can see how it works even when you show how to set them without changing the original log level in the example. But it should be consistent and follow the same approach everywhere.

fvaleri:
> This part of the docs should not say _this is the logging you should use_ => if that is the case we should change the code (but in general we should IMHO follow Kafka defaults as a Kafka's recommendation). It should show how you change them. I can see how it works even when you show how to set them without changing the original log level in the example. But it should be consistent and follow the same approach everywhere.

I agree on showing extra configurations on all components. 

Given that the default is the recommended setup and we don't show it in the doc, a hasty user may simply cut and paste one of these snippets to add it's own configuration. In that sense, I think it is safer to set them at INFO level or whatever value Kafka and CC recommend.

PaulRMellor:
> > This part of the docs should not say _this is the logging you should use_ => if that is the case we should change the code (but in general we should IMHO follow Kafka defaults as a Kafka's recommendation). It should show how you change them. I can see how it works even when you show how to set them without changing the original log level in the example. But it should be consistent and follow the same approach everywhere.
> 
> I agree on showing extra configurations on all components.
> 
> Given that the default is the recommended setup and we don't show it in the doc, a hasty user may simply cut and paste one of these snippets to add it's own configuration. In that sense, I think it is safer to set them at INFO level or whatever value Kafka and CC recommend.

We now have a note about using debug with each example. I don't think we need to set everything at info just in case.
Besides, we want to show that you can set separate loggers at different levels, and would possibly set debug at a lower level.

scholzj:
> We now have a note about using debug with each example. I don't think we need to set everything at info just in case.
Besides, we want to show that you can set separate loggers at different levels, and would possibly set debug at a lower level.

I'm fine with using DEBUG. Lets just be consistent and do it everywhere the same way.

documentation/api/io.strimzi.api.kafka.model.KafkaConnectSpec.adoc
```diff
@@ -114,9 +116,13 @@ spec:
     type: inline
     loggers:
       connect.root.logger.level: "INFO"
+      log4j.logger.org.apache.kafka.connect.runtime.WorkerSourceTask: "DEBUG"
+      log4j.logger.org.apache.kafka.connect.runtime.WorkerSinkTask: "TRACE"
```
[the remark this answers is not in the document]
PaulRMellor: When we first added these examples, quotes were used for the log4j loggers and no-quotes on the log4j2 loggers. I've updated so none now use quotes

documentation/api/io.strimzi.api.kafka.model.EntityUserOperatorSpec.adoc
```diff
@@ -38,9 +40,13 @@ spec:
         type: inline
         loggers:
           rootLogger.level: INFO
+          logger.io.strimzi.operator.user.UserController: TRACE
+          logger.io.strimzi.operator.user.HealthCheckAndMetricsServer: DEBUG
```
scholzj, line 44: This does not work:
* These classes have no real debug or trace logs so they serve as a bad example
* This configuration changes the log level for an exiting logger. Not for a class. So it results in following error:
  ```
  Exception in thread "Log4j2-TF-1-ConfiguratonFileWatcher-8" org.apache.logging.log4j.core.config.ConfigurationException: No name attribute provided for Logger io
	at org.apache.logging.log4j.core.config.properties.PropertiesConfigurationBuilder.createLogger(PropertiesConfigurationBuilder.java:264)
	at org.apache.logging.log4j.core.config.properties.PropertiesConfigurationBuilder.build(PropertiesConfigurationBuilder.java:180)
	at org.apache.logging.log4j.core.config.properties.PropertiesConfigurationFactory.getConfiguration(PropertiesConfigurationFactory.java:56)
	at org.apache.logging.log4j.core.config.properties.PropertiesConfiguration.reconfigure(PropertiesConfiguration.java:47)
	at org.apache.logging.log4j.core.LoggerContext.onChange(LoggerContext.java:755)
	at org.apache.logging.log4j.core.util.AbstractWatcher$ReconfigurationRunnable.run(AbstractWatcher.java:93)
	at java.base/java.lang.Thread.run(Thread.java:833)
  ```

I do not know if it is possible to configure log level just for a class here. I think you would need to use something like this where you define the logger to have it working and have an example of a class that demonstrates it is working by logging at debug level regularly:

```suggestion
          logger.abstractcache.level: DEBUG
          logger.abstractcache.name: io.strimzi.operator.user.operator.cache.AbstractCache
```

or alternatively something like this to log a whole package which might make sense in some cases as well:

```
        loggers:
          logger.cache.level: DEBUG
          logger.cache.name: io.strimzi.operator.user.operator.cache
```

Defining the loggers like this makes it a bit more complicated and I wonder if we should mention also the existing loggers already defined to avoid conflicts.

This will apply to all components using Log4j2.

documentation/api/io.strimzi.api.kafka.model.EntityTopicOperatorSpec.adoc
```diff
@@ -40,6 +40,8 @@ spec:
         type: inline
         loggers:
           rootLogger.level: INFO
+          logger.io.strimzi.operator.topic.ZkTopicStore: TRACE
+          logger.io.strimzi.operator.topic.TopicOperator: DEBUG
```
fvaleri, line 44: As Jakub said in another comment, this is not working for all log4j2 components. 

This is an example to change a package log level:

```sh
spec:
  entityOperator:
    topicOperator:
      logging:
        type: inline
        loggers:
          rootLogger.level: INFO
          logger.top.name: io.strimzi.operator.topic
          logger.top.level: DEBUG
```

This is an example to change a class log level:

```sh
spec:
  entityOperator:
    topicOperator:
      logging:
        type: inline
        loggers:
          rootLogger.level: INFO
          logger.toc.name: io.strimzi.operator.topic.TopicOperator
          logger.toc.level: DEBUG
```

It is easier for the existing/default loggers that you can find in cluster-operator/src/main/resources/default-logging:

```sh
spec:
  entityOperator:
    topicOperator:
      logging:
        loggers:
          logger.clients.level: DEBUG
```

At runtime, all the above configs are stored the following config map:

```sh
$ kubectl get cm my-cluster-entity-topic-operator-config -o yaml
apiVersion: v1
data:
  log4j2.properties: |
    # Do not change this generated file. Logging can be configured in the corresponding Kubernetes resource.
    name=TOConfig
    appender.console.type=Console
    appender.console.name=STDOUT
    appender.console.layout.type=PatternLayout
    appender.console.layout.pattern=%d{yyyy-MM-dd HH:mm:ss,nnnnn} %-5p [%t] %c{1}:%L - %m%n
    rootLogger.level=INFO
    rootLogger.appenderRefs=stdout
    rootLogger.appenderRef.console.ref=STDOUT
    rootLogger.additivity=false
    logger.clients.name=org.apache.kafka.clients
    logger.clients.level=DEBUG
    logger.clients.additivity=false
    logger.streams.name=org.INFO.kafka.streams
    logger.streams.level=INFO
    logger.streams.additivity=false
    logger.okhttp3.name=okhttp3
    logger.okhttp3.level=INFO
    logger.okhttp3.additivity=false
    logger.toc.level=INFO
    logger.toc.name=io.strimzi.operator.topic.TopicOperator

    monitorInterval=30
kind: ConfigMap
metadata:
  creationTimestamp: "2023-05-17T05:59:05Z"
  labels:
    app.kubernetes.io/instance: my-cluster
    app.kubernetes.io/managed-by: strimzi-cluster-operator
    app.kubernetes.io/name: entity-topic-operator
    app.kubernetes.io/part-of: strimzi-my-cluster
    strimzi.io/component-type: entity-topic-operator
    strimzi.io/kind: Kafka
    strimzi.io/name: my-cluster-entity-topic-operator
  name: my-cluster-entity-topic-operator-config
  namespace: test
  ownerReferences:
  - apiVersion: kafka.strimzi.io/v1beta2
    blockOwnerDeletion: false
    controller: false
    kind: Kafka
    name: my-cluster
    uid: e5793828-703f-4d9a-957d-fa7c0f36c4a4
  resourceVersion: "1252331"
  uid: faeb38c3-179f-4e12-8234-86eebc8e90c8
```

fvaleri (changes requested):

documentation/api/io.strimzi.api.kafka.model.KafkaBridgeSpec.adoc
```diff
@@ -76,10 +76,10 @@ spec:
   logging:
     type: inline
     loggers:
-      logger.bridge.level: "INFO"
+      logger.bridge.level: INFO
       # enabling DEBUG just for send operation
       logger.send.name: "http.openapi.operation.send"
-      logger.send.level: "DEBUG"
+      logger.send.level: DEBUG
```
fvaleri, line 82: This is fine. As for the others, I would also include `rootLogger.level: INFO`. Then, we can remove `logger.bridge.level` because it maps to the base bridge package, so it is essentially the same as root.

fvaleri (changes requested):

scholzj:
> At runtime, all the above configs are stored the following config map:

This is internal implementation detail. It should not be part of the logging docs.

documentation/api/io.strimzi.api.kafka.model.CruiseControlSpec.adoc
```diff
@@ -264,6 +264,8 @@ spec:
       type: inline
       loggers:
         rootLogger.level: INFO
+        logger.com.linkedin.kafka.cruisecontrol.executor.Executor: TRACE
+        logger.com.linkedin.kafka.cruisecontrol.analyzer.GoalOptimizer: DEBUG
```
fvaleri, line 268: This does not work. Correct version:

```sh
spec:
  cruiseControl:
    logging:
      type: inline
      loggers:
        rootLogger.level: INFO
        logger.exec.name: com.linkedin.kafka.cruisecontrol.executor.Executor
        logger.exec.level: TRACE
        logger.go.name: com.linkedin.kafka.cruisecontrol.analyzer.GoalOptimizer
        logger.go.level: DEBUG
```

Here we don't have default loggers.

documentation/api/io.strimzi.api.kafka.model.KafkaConnectSpec.adoc
```diff
@@ -115,9 +115,9 @@ spec:
   logging:
     type: inline
     loggers:
-      connect.root.logger.level: "INFO"
-      log4j.logger.org.apache.kafka.connect.runtime.WorkerSourceTask: "DEBUG"
-      log4j.logger.org.apache.kafka.connect.runtime.WorkerSinkTask: "TRACE"
+      connect.root.logger.level: INFO
+      log4j.logger.org.apache.kafka.connect.runtime.WorkerSourceTask: TRACE
+      log4j.logger.org.apache.kafka.connect.runtime.WorkerSinkTask: DEBUG
```
fvaleri, line 120: This is correct (log4j1).

documentation/api/io.strimzi.api.kafka.model.KafkaMirrorMakerSpec.adoc
```diff
@@ -46,9 +46,9 @@ spec:
   logging:
     type: inline
     loggers:
-      mirrormaker.root.logger: "INFO"
-      log4j.logger.org.apache.kafka.clients.NetworkClient: "TRACE"
-      log4j.logger.org.apache.kafka.common.network.Selector: "DEBUG"
+      mirrormaker.root.logger: INFO
+      log4j.logger.org.apache.kafka.clients.NetworkClient: TRACE
+      log4j.logger.org.apache.kafka.common.network.Selector: DEBUG
```
fvaleri, line 51: This is correct (log4j1).

fvaleri (changes requested):

fvaleri:
> > At runtime, all the above configs are stored the following config map:
> 
> This is internal implementation detail. It should not be part of the logging docs.

This information was for @PaulRMellor benefit.

documentation/api/io.strimzi.api.kafka.model.EntityTopicOperatorSpec.adoc
PaulRMellor: Right. Okay. I'll use these examples and highlight the config for individual packages and classes

documentation/api/io.strimzi.api.kafka.model.EntityUserOperatorSpec.adoc
PaulRMellor: I used a combination of this and Federico's suggestion, with callouts to explain the possibilities. 
I mentioned that there are default loggers, with an example for the operators. Maybe we want to list all the default loggers for each?

fvaleri: I would keep default loggers out. 

Thy are convenient because it's just one line of configuration, but at the same time ambiguous. For example in the TO the config `logger.clients.level` maps to `org.apache.kafka.clients`, but it may refer to some other client package. You have to look at the config map or code to find out. Additionally, it is something we would need to keep aligned in case of changes (easy to forget).

PaulRMellor: Thanks @fvaleri - no more updates here then

scholzj (approved):

fvaleri (approved):
LGTM. Thanks @PaulRMellor.

PR #8491, opened 2023-05-05
Remove old apiestensions v1beta1 features from the CRD Generator

### LINKED ISSUE #7663: Clean old features from the CRD generator

scholzj:
### Type of change

- Task

### Description

The `apiextensions` `v1beta1` API is for a long time not used by Strimzi. But we still had the related code in the CRD Generator and related classes. This PR cleans up the code related tot he `v1beta1` API and leaves only the code relevant to the `v1` API. It also addresses some minor warnings from the IDE.

This should close #7663

### Checklist


scholzj:
> Did we _need_ to remove support for `oneOf` aka `@Alternation`? Or did you just remove it because we don't currently use it?

It is not supported any more in `apiextensions.k8s.io/v1` - this was one of the incompatibilities between `v1` and `v1beta1`. So I do not think there is much sense in keeping it around.

scholzj:
> > Did we _need_ to remove support for `oneOf` aka `@Alternation`? Or did you just remove it because we don't currently use it?
> 
> It is not supported any more in `apiextensions.k8s.io/v1` - this was one of the incompatibilities between `v1` and `v1beta1`. So I do not think there is much sense in keeping it around.

Just to clarify -> the `@Alternation` was support for multiple different types under one field. That is not supported anymore and removed. The `@OneOf` as _use either field `x` or field `y` is still supported by Kubernetes and that was not removed in this PR.

tombentley (approved):

PR #8498, opened 2023-05-09
Move `PlatformFeaturesAvailability` class to `cluster-operator` module

scholzj:
### Type of change

- Refactoring

### Description

The `PlatformFeaturesAvailability` class is part of the `operator-common` module and thus is included in all the operators. However, it is really used only in the `cluster-operator` module and we do not have any plans to use it elsewhere as far as I know -> the User and Topic Operators are not really dependent on the platform versions.

This PR moves the class and its test to the `cluster-operator` module. That should help to minimize the footprint of the other operators by not having it in the shared code-base.

### Checklist


ppatierno (approved):
LGTM.

PR #8500, opened 2023-05-10
Push Helm Chart as OCI artifact to Quay.io during the release

scholzj:
### Type of change

- Taskj

### Description

Helm charts can be pushed into the Quay.io registry as an OCI artifcat. This PR adds a new stage to the release pipeline to push it there as part of the release. This should allow us in the future to install the artifacts directly from Quay.io without pushing them to some other storage (such as GitHub release) - but at this point in time, it is more experimental step to see what we can do with it. One interesting option would be for example if we can get rid of managing the Helm Chart index.yaml etc.

This PR also updates the Helm version used in our CI to use version which supports the OCI artifacts.

### DIFF
--- a/.azure/release-pipeline.yaml
+++ b/.azure/release-pipeline.yaml
@@ -64,3 +64,17 @@ stages:
           artifactRunVersion: 'specific'
           artifactRunId: '${{ parameters.sourceBuildId }}'
           architectures: ['amd64', 'arm64', 's390x']
+  # Publishes the Strimzi Helm Chart as an OCI artifact to Quay.io
+  - stage: helm_as_oci_publish
+    displayName: Publish Helm Chart as OCI artifact
+    dependsOn:
+      - prepare_release_artifacts
+    condition: and(succeeded(), startsWith(variables['build.sourceBranch'], 'refs/heads/release-'))
+    jobs:
+      - template: 'templates/jobs/build/publish_helm_chart_as_oci.yaml'
+        parameters:
+          artifactSource: 'current'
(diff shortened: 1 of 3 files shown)


Frawless (approved):

scholzj:
I called it experimental mostly because I do not fully understand its value yet :-o. Otherwise, I think it should be fine to use. The Helm Chart(s) are right now here (the Drain Cleaner is not there yet): https://quay.io/repository/strimzi-helm/strimzi-kafka-operator?tab=tags

----

But as you say that you are using the OCI charts ... can you maybe share some more background?
* Why are you using them?
* How are you using them?

I know that you can install the Helm Charts from the OCI repo with something like this:

```
helm install strimzi oci://quay.io/strimzi-helm/strimzi-kafka-operator
```

But I do not fully understand if it can completely replace the the Helm Chart repository which we have at `https://strimzi.io/charts/`. That would save us the maintenance effort for the `index.yaml` file for the repo.

rouke-broersma:
OCI completely replaces the old helm repository yes, no need to keep index.yaml (except that it would be a massive breaking change for any current users).

We use them because they are easier to distribute, we already have a container registry why maintain a separate helm registry as well.

We are also seeing that more and more partners are moving to OCI. Another benefit is that it is easy to 'cache' OCI charts in your own registry. In our case we already do this by making all the charts we use a chart dependency, and pushing our own chart to the registry. But for simple charts we could push the OCI chart to our own registry directly. Helm repositories afaik don't really have this built-in cacheability in the same way.

PR #8521, opened 2023-05-17
Returning 204 No Content on status check endpoints

ppatierno:
This PR is about the status check endpoints (ready and healthy) to return `204` instead of `200` because no content is provided (it mostly aligns with what we have on the bridge with this PR https://github.com/strimzi/strimzi-kafka-bridge/pull/800).

### DIFF
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/Main.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/Main.java
@@ -326,9 +326,9 @@ public class Main {
         vertx.createHttpServer()
                 .requestHandler(request -> {
                     if (request.path().equals("/healthy")) {
-                        request.response().setStatusCode(200).end();
+                        request.response().setStatusCode(204).end();
                     } else if (request.path().equals("/ready")) {
-                        request.response().setStatusCode(200).end();
+                        request.response().setStatusCode(204).end();
                     } else if (request.path().equals("/metrics")) {
                         PrometheusMeterRegistry metrics = (PrometheusMeterRegistry) metricsProvider.meterRegistry();
                         request.response().setStatusCode(200)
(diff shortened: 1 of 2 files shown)


ppatierno:
What I noticed is that the UO is behaving differently. It's returning a content like this:

```json
{
  "status": "ok|not-ok"
}
```

so the `200` makes sense. @scholzj is this content really needed? Aren't the returns code (204 and 500) already enough?

scholzj:
> @scholzj is this content really needed? Aren't the returns code (204 and 500) already enough?

I find the content to be useful. But I do not feel strongly enough to ask you to add it to the other operators 🤷

scholzj (approved):

ppatierno:
> I find the content to be useful. But I do not feel strongly enough to ask you to add it to the other operators

My rationale was about to be consistent across them. It could also mean removing the content from the UO, not only adding to the others ;-)

scholzj:
I would not remove it from the UO. And TBH, I did not raised the HTTP 404 versus 500 issue because of consistency -> just because 404 felt simply wrong. The 500 used in UO was mentioned only as an example what might be a better code.

ppatierno:
> I would not remove it from the UO. And TBH, I did not raised the HTTP 404 versus 500 issue because of consistency -> just because 404 felt simply wrong. The 500 used in UO was mentioned only as an example what might be a better code.

My point is not related to 404 vs 500. I agreed with you on moving to use 500 in the bridge. As I mentioned in my PR it was another option I had in mind, didn't know was already used in the operators, so I feel it was right to do that.
My point now is about 200 vs 204. But if you feel that content should be there for UO and consistency with CO and TO is not needed in that regards, then let's leave things how they are. But how much that content is used is not clear to me. The healthy server is there for Kube probes and a meaningful return code should be enough there.

ppatierno:
@scholzj if we want to merge and proceed with this, I don't think we need run regression, or do we?

PR #8530, opened 2023-05-19
Add trailing dot to the DNS name pattern used by CertManager

### LINKED ISSUE #8520: [Bug]: unable to add trailing dot to fqdn
### Bug Description

I am trying to add trailing dots to FQDNs in my strimzi configuration - I need this to address excessive dns searches on kubernetes cluster where I am running kafka producers.  When I try to add trailing dots, the operator throws an exception:

2023-05-17 00:29:52 WARN  AbstractOperator:525 - Reconciliation #555(watch) Kafka(kafka/vcmp): Failed to reconcile
java.lang.IllegalArgumentException: Invalid DNS name: tpavcpaw112s12.vici.verizon.com.
	at io.strimzi.certs.Subject$Builder.addDnsNames(Subject.java:83) ~[io.strimzi.certificate-manager-0.33.2.jar:0.33.2]

### Steps to reproduce

in kafka resource, try to use an fqdn with a trailing dot:

kind: Kafka
metadata:
  name: kafkaresource
  namespace: kafka
spec:
  kafka:
    listeners:
      - name: external
        port: 9094
        type: nodeport
        tls: true
        authentication:
          type: tls
        configuration:
          preferredNodePortAddressType: Hostname
          bootstrap:
            nodePort: 32261
            alternativeNames:
              - 'host.mydomain.com.'

config trimmed for clarity...

### Expected behavior

I would like FQDNs with trailing dots to me allowed by the DNS name validation - I believe it is here: https://github.com/strimzi/strimzi-kafka-operator/blob/9aa9cd3143e57a3e0e9a3138ce680c3a87b2b534/certificate-manager/src/main/java/io/strimzi/certs/IpAndDnsValidation.java#L23

### Strimzi version

0.33.2

### Kubernetes version

v1.26.3

### Installation method

yaml files

### Infrastructure

bare-metal

2023-05-17 00:29:50 INFO  AbstractOperator:239 - Reconciliation #555(watch) Kafka(kafka/vcmp): Kafka vcmp will be checked for creation or modification
2023-05-17 00:29:52 ERROR AbstractOperator:260 - Reconciliation #555(watch) Kafka(kafka/vcmp): createOrUpdate failed
java.lang.IllegalArgumentException: Invalid DNS name: mybroker.mydomain.com.
	at io.strimzi.certs.Subject$Builder.addDnsNames(Subject.java:83) ~[io.strimzi.certificate-manager-0.33.2.jar:0.33.2]
	at io.strimzi.certs.Subject$Builder.addDnsName(Subject.java:66) ~[io.strimzi.certificate-manager-0.33.2.jar:0.33.2]
	at io.strimzi.operator.cluster.model.ClusterCa.lambda$generateBrokerCerts$4(ClusterCa.java:230) ~[io.strimzi.cluster-operator-0.33.2.jar:0.33.2]
	at io.strimzi.operator.cluster.model.Ca.maybeCopyOrGenerateCerts(Ca.java:454) ~[io.strimzi.operator-common-0.33.2.jar:0.33.2]
	at io.strimzi.operator.cluster.model.ClusterCa.generateBrokerCerts(ClusterCa.java:248) ~[io.strimzi.cluster-operator-0.33.2.jar:0.33.2]
	at io.strimzi.operator.cluster.model.KafkaCluster.generateCertificatesSecret(KafkaCluster.java:1212) ~[io.strimzi.cluster-operator-0.33.2.jar:0.33.2]
	at io.strimzi.operator.cluster.operator.assembly.KafkaReconciler.lambda$certificateSecret$48(KafkaReconciler.java:726) ~[io.strimzi.cluster-operator-0.33.2.jar:0.33.2]
	at io.vertx.core.impl.future.Composition.onSuccess(Composition.java:38) ~[io.vertx.vertx-core-4.3.7.jar:4.3.7]
	at io.vertx.core.impl.future.FutureBase.emitSuccess(FutureBase.java:60) ~[io.vertx.vertx-core-4.3.7.jar:4.3.7]
	at io.vertx.core.impl.future.FutureImpl.tryComplete(FutureImpl.java:211) ~[io.vertx.vertx-core-4.3.7.jar:4.3.7]
	at io.vertx.core.impl.future.PromiseImpl.tryComplete(PromiseImpl.java:23) ~[io.vertx.vertx-core-4.3.7.jar:4.3.7]
	at io.vertx.core.impl.future.PromiseImpl.onSuccess(PromiseImpl.java:49) ~[io.vertx.vertx-core-4.3.7.jar:4.3.7]
	at io.vertx.core.impl.future.FutureBase.lambda$emitSuccess$0(FutureBase.java:54) ~[io.vertx.vertx-core-4.3.7.jar:4.3.7]
	at io.netty.util.concurrent.AbstractEventExecutor.runTask(AbstractEventExecutor.java:174) ~[io.netty.netty-common-4.1.86.Final.jar:4.1.86.Final]
	at io.netty.util.concurrent.AbstractEventExecutor.safeExecute(AbstractEventExecutor.java:167) ~[io.netty.netty-common-4.1.86.Final.jar:4.1.86.Final]
	at io.netty.util.concurrent.SingleThreadEventExecutor.runAllTasks(SingleThreadEventExecutor.java:470) ~[io.netty.netty-common-4.1.86.Final.jar:4.1.86.Final]
	at io.netty.channel.nio.NioEventLoop.run(NioEventLoop.java:569) ~[io.netty.netty-transport-4.1.86.Final.jar:4.1.86.Final]
	at io.netty.util.concurrent.SingleThreadEventExecutor$4.run(SingleThreadEventExecutor.java:997) ~[io.netty.netty-common-4.1.86.Final.jar:4.1.86.Final]
	at io.netty.util.internal.ThreadExecutorMap$2.run(ThreadExecutorMap.java:74) ~[io.netty.netty-common-4.1.86.Final.jar:4.1.86.Final]
	at io.netty.util.concurrent.FastThreadLocalRunnable.run(FastThreadLocalRunnable.java:30) ~[io.netty.netty-common-4.1.86.Final.jar:4.1.86.Final]
	at java.lang.Thread.run(Thread.java:833) ~[?:?]

### Additional context

_No response_

fvaleri:
By default, the /etc/resolv.conf of each container looks like this:

```sh
search test.svc.cluster.local svc.cluster.local cluster.local us-east-2.compute.internal
nameserver 172.30.0.10
options ndots:5
```

The ndots option sets a threshold for the number of dots which must appear in the host name before a FQDN search will be made. If the host name contains less than 5 dots, the resolution will sequentially go through all local search domains first and, in case none succeed, will resolve it as a FDQN only at last.

It turns out that adding a trailing dot to the host name forces the FQDN resolution on the upstream DNS server, thus reducing internal DNS server load (kube-dns pods) and application latency.

You can also use the dnsConfig pod property to reduce the ndots value for your application, but that would be impractical when having lots of client applications, where some of them could be by a third party.

Note that TLS handshake with trailing dot only works with recent JDK versions (8u372+, 11.0.19+, 17.0.6+).

https://bugs.openjdk.org/browse/JDK-8065422

This should fix #8520.

### DIFF
--- a/certificate-manager/src/main/java/io/strimzi/certs/IpAndDnsValidation.java
+++ b/certificate-manager/src/main/java/io/strimzi/certs/IpAndDnsValidation.java
@@ -30,7 +30,7 @@ public class IpAndDnsValidation {
                 "[a-zA-Z\\d][a-zA-Z\\d\\-]{0,61}[a-zA-Z\\d])";
         // a trailing dot that can be used to force FQDN resolutions, reducing
         // the internal DNS server load (kube-dns pods) and application latency
-        String trailingDot = "(\\.?)";
+        String trailingDot = "\\.?";
         DNS_NAME = Pattern.compile("^" + dnsLabel + "(\\." + dnsLabel + ")*" + trailingDot);
     }
 
--- a/certificate-manager/src/main/java/io/strimzi/certs/IpAndDnsValidation.java
+++ b/certificate-manager/src/main/java/io/strimzi/certs/IpAndDnsValidation.java
@@ -28,7 +28,10 @@ public class IpAndDnsValidation {
                 "[a-zA-Z\\d]|" +
                 // followed by more labels of same
                 "[a-zA-Z\\d][a-zA-Z\\d\\-]{0,61}[a-zA-Z\\d])";
-        DNS_NAME = Pattern.compile("^" + dnsLabel + "(\\." + dnsLabel + ")*");
+        // a trailing dot that can be used to force FQDN resolutions, reducing
+        // the internal DNS server load (kube-dns pods) and application latency
+        String trailingDot = "(\\.?)";
+        DNS_NAME = Pattern.compile("^" + dnsLabel + "(\\." + dnsLabel + ")*" + trailingDot);
     }
 
     // Pattern used to normalize the segments of the IPv6 address
--- a/certificate-manager/src/test/java/io/strimzi/certs/IpAndDnsValidationTest.java
+++ b/certificate-manager/src/test/java/io/strimzi/certs/IpAndDnsValidationTest.java
@@ -49,6 +49,10 @@ public class IpAndDnsValidationTest {
 
         assertThat(IpAndDnsValidation.isValidDnsNameOrWildcard("example:com"), is(false));
         assertThat(IpAndDnsValidation.isValidDnsNameOrWildcard("veryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryverylongexample.com"), is(false));
+
+        assertThat(IpAndDnsValidation.isValidDnsNameOrWildcard("example.com."), is(true));
+        assertThat(IpAndDnsValidation.isValidDnsNameOrWildcard("example.com.."), is(false));
+        assertThat(IpAndDnsValidation.isValidDnsNameOrWildcard("*.example.com."), is(true));
     }
 
     @Test
--- a/certificate-manager/src/test/java/io/strimzi/certs/OpenSslCertManagerIT.java
+++ b/certificate-manager/src/test/java/io/strimzi/certs/OpenSslCertManagerIT.java
@@ -301,7 +301,8 @@ public class OpenSslCertManagerIT {
                 .withCommonName("MyCommonName")
                 .withOrganizationName("MyOrganization")
                 .addDnsName("example1.com")
-                .addDnsName("example2.com").build();
+                .addDnsName("example2.com")
+                .addDnsName("example3.com.").build();
 
         File cert = Files.createTempFile("crt-", ".crt").toFile();
 
--- a/certificate-manager/src/test/java/io/strimzi/certs/SubjectTest.java
+++ b/certificate-manager/src/test/java/io/strimzi/certs/SubjectTest.java
@@ -77,8 +77,10 @@ public class SubjectTest {
     public void testDnsValidation() {
         new Subject.Builder().addDnsName("example.com");
         new Subject.Builder().addDnsName("*.example.com");
-        assertThrows(IllegalArgumentException.class, () -> new Subject.Builder().addDnsName("foo.*.example.come"));
+        new Subject.Builder().addDnsName("example.io.");
+        assertThrows(IllegalArgumentException.class, () -> new Subject.Builder().addDnsName("foo.*.example.com"));
         assertThrows(IllegalArgumentException.class, () -> new Subject.Builder().addDnsName("54t8g#'/.l"));
+        assertThrows(IllegalArgumentException.class, () -> new Subject.Builder().addDnsName("example.io.."));
     }
 
     @Test


scholzj (commented):
I'm not sure I follow the use-case and value of this. You are changing just the validation ... but it would be still good to understand the use case and value of this to better understand if something more should be done. It also seems to be missing any kind of test.

fvaleri:
The use case and value of this is "reducing internal DNS server load (kube-dns pods) and application latency", which is basically what the user reported.

The full story is described here: https://pracucci.com/kubernetes-dns-resolution-ndots-options-and-why-it-may-affect-application-performances.html

I did manual tests and I can add automated tests, but first wanted to check the approach with you.

scholzj:
> The use case and value of this is "reducing internal DNS server load (kube-dns pods) and application latency" without touching client applications, which is basically what the user reported.

I don't think I follow that. If you run inside Kubernetes, you should use an internal listener and Kubernetes DNS. You should not use external listener and resolve the DNS elsewhere. So why should that matter?

tombentley:
What if your clients are also in a kube cluster, but a different kube cluster than the strimzi cluster?

scholzj:
Ok, that would make some sense I guess ... was it meant like that @fvaleri?

scholzj:
Ok, then let's keep it. But please add some tests for the DNS name validation.

fvaleri:
Sure, will do. Thanks.

fvaleri:
@scholzj @tombentley unit and integration tests added.

Considering the impact that ndot:5 may have on application latency, do we also want to add a documentation note in "Advertised addresses for brokers"?

scholzj:
> Considering the impact that the ndot:5 kube-dns option may have on application latency, do we also want to add a documentation note in "Advertised addresses for brokers"?

TBH, it does not seem to be a general issue otherwise the advertised addresses in many different places should follow that rule in general which they don't. So I'm in for enabling users to use it, but would not really cover it in docs.

scholzj (approved):

certificate-manager/src/main/java/io/strimzi/certs/IpAndDnsValidation.java
```diff
@@ -28,7 +28,10 @@ public class IpAndDnsValidation {
                 "[a-zA-Z\\d]|" +
                 // followed by more labels of same
                 "[a-zA-Z\\d][a-zA-Z\\d\\-]{0,61}[a-zA-Z\\d])";
-        DNS_NAME = Pattern.compile("^" + dnsLabel + "(\\." + dnsLabel + ")*");
+        // a trailing dot that can be used to force FQDN resolutions, reducing
+        // the internal DNS server load (kube-dns pods) and application latency
+        String trailingDot = "(\\.?)";
```
tombentley, line 33: I guess we don't need the parens, because we don't want to capture the match for `.?`, right?

tombentley (approved):
One nit, but lgtm otherwise.

certificate-manager/src/main/java/io/strimzi/certs/IpAndDnsValidation.java
fvaleri: Yep, removed.

scholzj:
@ppatierno Do you have anything for this? Or can it be merged?

ppatierno (approved):
LGTM.

Issue #50, opened 2017-10-15
Kafka should use service DNS names and not Pod IP addresses as advertised host


scholzj:
Kafka is currently using the pod IP address as the advertised host name. I think that is not correct, as the Pod IP address is not necessarily stable. It should be using either the IP address exposed through the headless service or probably even better the hostname exposed through the headless service.

I'm not sure in detail how the different Kafka clients work and how often they recheck the advertised addresses. But having them stable should not hurt, it can only help.

scholzj:
It looks like the change to DNS hostnames caused some problems to users who are using Kafka from outside of OpenShift since the hostnames don't resolve outside of OpenShift. We should rollback this change

This is still a valid issue but we should revisit it later when we have other ways how to make Kafka available outside of OpenShift.

komljen:
I agree that pod IP is not a good choice and something like `${HOSTNAME}.kafka-headless` is probably more suitable. Hostname in StatefulSet is stable.

scholzj:
@komljen It was causing some issues when trying to access Kafka from outside of OpenShift. Basically, if you start the OpenShift cluster on Linux using `oc cluster up` you can access the pods directly trhough IP address also from local applications not running on OpenShift. When the advertised hostnames are changed from Pod IP address to DNS names, these DNS names are not resolved outside of OpenShift and this stops working. But some users are using this right now. 

So we need to first implement better way for exposing Kafka to the outside of OpenShift. And once we do that we will again change to using the hostnames instead of the PodIPs.

Veryclouder:
Kafka operator has adopted your advisement, but I can't use svc DNS name to access kafka from outside

Veryclouder:
I hope you can give me some suggestions. 
For example, when I use kafka operator to deploy kafka, I can modify some parameters to use pod ip instead of use DNS name

scholzj:
You should use the external listener to access the Kafka cluster from outside.

Veryclouder:
thanks, from my heart!

Veryclouder:
I'll think about it later. The external listener is used to expose the services inside the cluster. The main problem with Kafka is that the domain name, not the IP, is used to register the backend pod to zookeeper. The domain name cannot be accessed locally. If the domain name resolution is configured locally, it can be accessed, but if the pod IP changes, the configuration file must be changed!



发自 网易邮箱大师




---- 回复的原邮件 ----
| 发件人 | Jakub ***@***.***> |
[...]

PR #8551, opened 2023-05-23
Initial phase of the Kafka Node Pool implementation

scholzj:
### Type of change

- Enhancement / new feature

### Description

This PR provides the initial phase of the Kafka Node Pool implementation - [Strimzi Proposal #50](https://github.com/strimzi/proposals/blob/main/050-Kafka-Node-Pools.md). The are several follow-up issues tracked through the code (#8590, #8592, #8593). Some of them are created in order to speed things up and minimize the PR size. Others are because System Test changes would need to be done first.

The implementation is using a backwards compatible [_virtual node pool_](https://github.com/strimzi/proposals/blob/main/050-Kafka-Node-Pools.md#virtual-node-pool) implementation to avoid separate code paths. It also uses a feature gate to prevent users from using the `KafkaNodePool` resources (given they are in `alpha` stage and we might decide to not proceed with it). Also, as described in the proposal, the storage and replicas fields in the `Kafka` custom resource remain mandatory as long as the node pools are in `alpha` stage and they have to be specified despite being ignored.

This PR does not contain any documentation and examples which will be delivered in a separate PR (again - mainly in order to speed the process).

### Checklist

- [x] Write tests
- [x] Update CHANGELOG.md


api/src/main/java/io/strimzi/api/kafka/Crds.java
```diff
@@ -24,6 +24,7 @@
 import io.strimzi.api.kafka.model.KafkaRebalance;
 import io.strimzi.api.kafka.model.KafkaTopic;
 import io.strimzi.api.kafka.model.KafkaUser;
+import io.strimzi.api.kafka.model.nodepool.KafkaNodePool;
```
ppatierno, line 27: I don't have a strong opinion, but I am not sure we really need this `nodepoll` layer under `model`.
Making a comparison and thinking about `KafkaRebalance` resource, we don't have it in a `rebalance` layer under `model` for example.

api/src/main/java/io/strimzi/api/kafka/model/KafkaClusterSpec.java
```diff
@@ -71,6 +71,7 @@ public class KafkaClusterSpec implements HasConfigurableMetrics, HasConfigurable
     private List<GenericKafkaListener> listeners;
     private KafkaAuthorization authorization;
     private KafkaClusterTemplate template;
+    //private LabelSelector nodePoolSelector;
```
ppatierno, line 74: to be removed?

api/src/main/java/io/strimzi/api/kafka/model/KafkaClusterSpec.java
```diff
@@ -253,6 +254,19 @@ public void setTemplate(KafkaClusterTemplate template) {
...
+    //@KubeLink(group = "meta", version = "v1", kind = "labelselector")
+    //@JsonProperty(required = true)
+    //public LabelSelector getNodePoolSelector() {
+    //    return nodePoolSelector;
+    //}
+    //
+    //public void setNodePoolSelector(LabelSelector nodePoolSelector) {
+    //    this.nodePoolSelector = nodePoolSelector;
+    //}
+
```
ppatierno, line 269: all the block above to be removed?

api/src/main/java/io/strimzi/api/kafka/model/nodepool/KafkaNodePool.java
```diff
@@ -0,0 +1,141 @@
...
+                        scale = @Crd.Spec.Subresources.Scale(
+                                specReplicasPath = KafkaNodePool.SPEC_REPLICAS_PATH,
+                                statusReplicasPath = KafkaNodePool.STATUS_REPLICAS_PATH,
+                                labelSelectorPath = KafkaNodePool.LABEL_SELECTOR_PATH
+                        )
+                ),
+                additionalPrinterColumns = {
+                    @Crd.Spec.AdditionalPrinterColumn(
+                                name = "Desired replicas",
+                                description = "The desired number of Kafka Connect replicas",
```
ppatierno, line 56: Kafka Connect? :-P

api/src/main/java/io/strimzi/api/kafka/model/nodepool/KafkaNodePoolSpec.java
```diff
@@ -0,0 +1,106 @@
...
+
+/**
+ * Represents the .spec section of a KafkaNodePool resource
+ */
+@Buildable(
+        editableEnabled = false,
+        builderPackage = Constants.FABRIC8_KUBERNETES_API
+)
+@JsonInclude(JsonInclude.Include.NON_NULL)
+@JsonPropertyOrder({"replicas", "storage", "resources", "jvmOptions", "template"})
```
ppatierno, line 31: missing "roles" I think

api/src/test/java/io/strimzi/api/kafka/model/nodepool/KafkaNodePoolCrdIT.java
```diff
@@ -0,0 +1,47 @@
...
+import io.strimzi.api.kafka.model.AbstractCrdIT;
+import io.strimzi.test.TestUtils;
+import org.junit.jupiter.api.AfterAll;
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.Test;
+
+/**
+ * This class tests the encoding and decoding of the KafkaNodePool custom resource
+ */
+public class KafkaNodePoolCrdIT extends AbstractCrdIT {
```
ppatierno, line 16: should we also have a test where the `KafkaNodePool` resource has a wrong/not existing role configured in?

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaBrokerConfigurationBuilder.java
```diff
@@ -239,66 +244,83 @@ public KafkaBrokerConfigurationBuilder withListeners(
...
+                ));
+            }
+
+            securityProtocol.add(CONTROL_PLANE_LISTENER_NAME + ":SSL");
+            configureControlPlaneListener();
+        } else {
+            // But even KRaft broker only node needs it in the security protocol map and the security configuration
+            configureControlPlaneListener();
+            securityProtocol.add(CONTROL_PLANE_LISTENER_NAME + ":SSL");
+        }
```
ppatierno, line 268: two questions:

1. any reasons why the two lines of code above are in a different order in the two `if-else` branches?
2. can we avoid the `else` and having the above two lines out of it because it's anyway in common with the `if` branch?

ppatierno (commented):
I had a first pass leaving some comments but I will need another one :-)

tombentley (commented):
Some initial comments, from a partial first pass.

api/src/main/java/io/strimzi/api/kafka/Crds.java
scholzj: TBH, I think the `io.strimzi.api.kafka.model` package is already loaded with all kinds of classes, often used only once. I think better separation into smaller packages is useful for things that are not shared in any way I would ideally love to split the existing classes into packages as well, but that would break the compatibility for anyone using the API. But from my point of view, this is a good way to go - so unless you have really strong feelings about this, I would prefer to keep it.

ppatierno: I totally agree with refactoring and moving the others to smaller packages, if that was the rationale behind this choice then +1 from me.
I can totally understand your concern about breaking compatibility but I think we should move in that direction asap.

scholzj: Well, ok ... I can open an issue to refactor the existing as well. But I do not think that should be done in this PR anyway.

ppatierno: > But I do not think that should be done in this PR anyway.

Absolutely agree.

scholzj: I opened #8628 for it.

scholzj:
@tombentley @ppatierno Thanks for the comments. All of them should be incorporated now.

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaPool.java
```diff
@@ -0,0 +1,296 @@
...
+                        "result, all storage changes will be ignored. Use DEBUG level logging for more information " +
+                        "about the detected changes.", pool.getMetadata().getNamespace(), pool.getMetadata().getName());
+
+                Condition warning = StatusUtils.buildWarningCondition("KafkaStorage",
+                        "The desired Kafka storage configuration in the KafkaNodePool custom resource " + pool.getMetadata().getNamespace() + "/" + pool.getMetadata().getName() + " contains changes which are not allowed. As a " +
+                        "result, all storage changes will be ignored. Use DEBUG level logging for more information " +
+                        "about the detected changes.");
+                result.warningConditions.add(warning);
+
+                result.setStorage(oldStorage);
```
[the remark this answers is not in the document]
scholzj: We did use setters for everything in the past and then removed them because 99% of them did not do anything -> it just generated a lot of dummy code doing nothing. I you want to introduce them everywhere again, then I think that is way beyond the scope of this PR.

Alternatively, I can also remove the `setStorage` method here and simply do the validation before every time I set it directly => that would not be so hard.

api/src/main/java/io/strimzi/api/kafka/model/nodepool/KafkaNodePoolSpec.java
```diff
@@ -0,0 +1,106 @@
...
+
+    private int replicas;
+    private Storage storage;
+    private List<ProcessRoles> roles;
+    private ResourceRequirements resources;
+    private JvmOptions jvmOptions;
+    private KafkaNodePoolTemplate template;
+
+    @Description("The number of pods in the pool.")
+    @Minimum(0)
```
ppatierno, line 44: I was wondering if we should allow to have an empty node pool. What does it mean? The current minimum `spec.replicas` we allow on Kafka and Zookeeper is 1. How does the node pool differ from this perspective?

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/nodepools/VirtualNodePoolConverter.java
```diff
@@ -0,0 +1,85 @@
...
+import io.strimzi.api.kafka.model.nodepool.KafkaNodePoolTemplateBuilder;
+import io.strimzi.api.kafka.model.template.KafkaClusterTemplate;
+
+import java.util.List;
+import java.util.stream.IntStream;
+
+/**
+ * Class which contains utility methods for converting the .spec.kafka section from the Kafka CR into a virtual KafkaNodePool resource
+ */
+public class VirtualNodePoolConverter   {
```
ppatierno, line 20:
```suggestion
public class VirtualNodePoolConverter {
```

ppatierno (commented):
I had another pass, left just a couple of comments. I have another couple of classes to focus on, but everything looks to be almost good. I will have another pass and accept by the end of the day for sure.

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/nodepools/NodePoolUtils.java
```diff
@@ -0,0 +1,161 @@
...
+                hasController = true;
+            }
+        }
+
+        if (!hasBroker) {
+            errors.add("At least one KafkaNodePool with the broker role is required when KRaft mode is enabled");
+        }
+
+        if (!hasController) {
+            errors.add("At least one KafkaNodePool with the controller role is required when KRaft mode is enabled");
```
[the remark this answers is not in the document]
scholzj: I will add it to `KafkaSpecChecker` where it probably fits better. I will also add a check for a single controller with ephemeral storage.

api/src/main/java/io/strimzi/api/kafka/model/nodepool/KafkaNodePoolSpec.java
scholzj: The main difference is that you can have multiple node pools. So in theory, you can have one with 0 replicas (=> no Kafka nodes) and another one with >0 replicas. Thus the cluster would still have some nodes and can work. Where is if you set replicas to 0 in the Kafka CR, it would mean 0 brokers and that would not really work.

ppatierno: Yeah but I was just wondering about what's the use case of an empty node pool. Thinking more maybe it allows the user to move brokers from a pool to another, making the first empty and then deleting it if needed. Or maybe they can take a node pool definition because will add more brokers later. With node pools (even if empty) they can have a kind of high-level vision of where brokers can be deployed and what are the resources that each pool has (storage, CPU, ..).

scholzj:
@tombentley @ppatierno The latest round of comments should be incorporated now.

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaPool.java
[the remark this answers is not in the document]
tombentley: I don't think we should standardize on relying on the caller to call a validation prior to setting a value: It's too easy to forget. 

It would be better to use getters and setters everywhere, precisely because it means there's a single place to add/fix validation code. But I hear your complaint about having to maintain those methods when 95% of the time they're just boilerplate. We could use Lombok's [@Getter and @Setter](https://projectlombok.org/features/GetterSetter) on the model classes. We already use Lombok in the `api` module for `equals()` and hashCode()`, so this wouldn't be a huge step.

More radically, I don't think the models actually have to be mutable. They're not changed once they've been returned from `fromCrd()`. If that's correct then we could refactor them to be `record`s, validating their fields in the constructor. To avoid the `fromCrd()`  needing many many local variables (prior to instantiating the model record) we could use sundrio to generate a builder. I'm not at all certain that this is actually a good idea, but I'd be interested to know just how much it makes you wince :slightly_smiling_face: 

In any case, I'm not suggesting you need to do either of these changes in _this_ PR, merely that I don't think this is a pattern we actually want to use long term/promote.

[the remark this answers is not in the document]
scholzj: > It would be better to use getters and setters everywhere, precisely because it means there's a single place to add/fix validation code. But I hear your complaint about having to maintain those methods when 95% of the time they're just boilerplate. We could use Lombok's [@Getter and @Setter](https://projectlombok.org/features/GetterSetter) on the model classes. We already use Lombok in the api module for equals() and hashCode()`, so this wouldn't be a huge step.

I do not think the problem is in maintaining these methods. The problem is the code readability because you do not know what such methods actually do. When you have just an assignment, you know it is just an assignment and nothing more. If you have setters, you need to always check what are they actually doing - is there any additional logic? Or is there just an assignment? So the code becomes very unreadable as you need to jump into the implementations of the setters / getters to see what they are doing.

I also wonder about the memory consumption impact, since cleaning all these methods in the past was one of the things which seemed to help us void some of the OoM issues.

> More radically, I don't think the models actually have to be mutable. They're not changed once they've been returned from fromCrd(). If that's correct then we could refactor them to be records, validating their fields in the constructor. To avoid the fromCrd() needing many many local variables (prior to instantiating the model record) we could use sundrio to generate a builder. I'm not at all certain that this is actually a good idea, but I'd be interested to know just how much it makes you wince 🙂

I think this makes some sense. Most of the fields should be essentially final. They are not final only because we do not set them in the constructor but in the `fromCrd` method.

ppatierno (approved):
I had another and final pass. Nothing to add. Thanks for this effort!

tombentley (approved):
A few more nits. Overall this looks good. Thanks, this was a huge piece of work.

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaPool.java
```diff
@@ -0,0 +1,296 @@
...
+
+        KafkaPool result = new KafkaPool(reconciliation, kafka, pool, componentName(kafka, pool), ownerReference, idAssignment);
+
+        result.gcLoggingEnabled = isGcLoggingEnabled(kafka, pool);
+        result.jvmOptions = pool.getSpec().getJvmOptions() != null ? pool.getSpec().getJvmOptions() : kafka.getSpec().getKafka().getJvmOptions();
+        result.resources = pool.getSpec().getResources() != null ? pool.getSpec().getResources() : kafka.getSpec().getKafka().getResources();
+
+        if (useKRaft)   {
+            // KRaft is enabled. The roles are what is set in the KafkaNodePool CR or mixed node by default
+            result.kraftRoles = pool.getSpec().getRoles() != null && !pool.getSpec().getRoles().isEmpty() ? new HashSet<>(pool.getSpec().getRoles()) : Set.of(ProcessRoles.CONTROLLER, ProcessRoles.BROKER);
```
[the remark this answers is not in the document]
scholzj: I think it would make sense to have them required for KRaft. But with ZooKeeper, not sure it makes sense to force this on users to set the role to broker. In the CRD schema, we cannot require the roles only in KRaft. So if we set them as required, it would be required even outside of KRaft -> but without CRD based validation that without a KRaft you have always only `BROKER` selected. 

Is that what we really want? If you think it should be required despite this, I'm fine to change it.

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaPool.java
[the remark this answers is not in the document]
scholzj: Ok, I opened #8658 for this.

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaPool.java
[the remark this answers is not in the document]
tombentley: The cost of making the roles required in the CRD today, AFAIU, is an instruction to the user in the docs for upgrading their ZK cluster to NodePools (and also for creating ZK clusters on NP). That seems small to me. The benefit is we end up with CRs which are much more explicit. So we're trading off ease of use of NodePools with ZK now against requiring explictness when using NodePools in a KRaft only future. One thing about the future is that there's a lot of it, so personally I'm inclined to say let's make it required today.

api/src/main/java/io/strimzi/api/kafka/model/nodepool/KafkaNodePoolStatus.java
```diff
@@ -0,0 +1,76 @@
...
+        return replicas;
+    }
+
+    public void setReplicas(int replicas) {
+        this.replicas = replicas;
+    }
+
+    @JsonInclude(JsonInclude.Include.NON_NULL)
+    @Description("Label selector for pods providing this resource.")
+    public String getLabelSelector() {
```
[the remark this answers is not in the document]
tombentley: OK, I suppose I can live with that.

PR #8566, opened 2023-05-25
Add tag for tests which are not supported on arm

kornys:
### Type of change

- Bugfix

### Description

Due to missing support of dependencies for systemtests like `keycloak` on arm64/aarch64 we need to exclude these tests when we run them on ARM clusters. So I introduced tag `armunsupported` to allow us exclude these tests.

### DIFF
--- a/systemtest/src/main/java/io/strimzi/systemtest/Constants.java
+++ b/systemtest/src/main/java/io/strimzi/systemtest/Constants.java
@@ -363,6 +363,11 @@ public interface Constants {
      */
     String ROUTE = "route";
 
+    /**
+     * Tag for tests, without ARM,AARCH64 support
+     */
+    String ARM_UNSUPPORTED = "armunsupported";
+
     String ISOLATED_TEST = "isolatedtest";
     String PARALLEL_TEST = "paralleltest";
     /**
--- a/systemtest/src/main/java/io/strimzi/systemtest/Constants.java
+++ b/systemtest/src/main/java/io/strimzi/systemtest/Constants.java
@@ -366,7 +366,7 @@ public interface Constants {
     /**
      * Tag for tests, without ARM,AARCH64 support
      */
-    String ARM_UNSUPPORTED = "armunsupported";
+    String ARM64_UNSUPPORTED = "arm64unsupported";
 
     String ISOLATED_TEST = "isolatedtest";
     String PARALLEL_TEST = "paralleltest";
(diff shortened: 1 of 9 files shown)


im-konge (approved):

systemtest/src/test/java/io/strimzi/systemtest/security/OpaIntegrationST.java
```diff
@@ -31,12 +31,14 @@
...
 import java.io.IOException;
 
+import static io.strimzi.systemtest.Constants.ARM_UNSUPPORTED;
 import static io.strimzi.systemtest.Constants.INTERNAL_CLIENTS_USED;
 import static io.strimzi.systemtest.Constants.REGRESSION;
 import static io.strimzi.test.k8s.KubeClusterResource.cmdKubeClient;
 
 @Tag(REGRESSION)
 @Tag(INTERNAL_CLIENTS_USED)
+@Tag(ARM_UNSUPPORTED)
```
scholzj, line 41: Why is this not supported? Does for example https://hub.docker.com/layers/openpolicyagent/opa/latest-static/images/sha256-7c754dc6ac0c9cd8309c9d90f43515c848f61a817d93b97128441ffff01110f3?context=explore not work?

kornys: you are right, fixed

systemtest/src/main/java/io/strimzi/systemtest/Constants.java
```diff
@@ -363,6 +363,11 @@ public interface Constants {
      */
     String ROUTE = "route";
 
+    /**
+     * Tag for tests, without ARM,AARCH64 support
+     */
+    String ARM_UNSUPPORTED = "armunsupported";
```
scholzj, line 369: Would `ARM64_UNSUPPORTED` and `arm64unsupported` make more sense? In case we have some other Arm platforms in the future :-o?

scholzj (approved):
LGTM assuming it works. I guess it would be good to have a tracker for making the OAuth tests usable on Aarch64 as well.

kornys:
> LGTM assuming it works. I guess it would be good to have a tracker for making the OAuth tests usable on Aarch64 as well.

It depends on keycloak operator, keycloak image is built for arm64 but operator does not contain manifests for arm.

scholzj:
> It depends on keycloak operator, keycloak image is built for arm64 but operator does not contain manifests for arm.

I can understand that. However, if it is only the manifest that is missing, that can be worked around on our side if needed (not saying you have to do this now - but it could be done if desired). We should also check from time to time again if the situation changed or not to enable these tests if they add a manifest. So that is why I think an issue for this would be useful.

Frawless (approved):

PR #8571, opened 2023-05-26
Common health server

tombentley:
### Type of change

- Refactoring

### Description

This PR moves the existing `HealthCheckAndMetricsServer` class from `user-operator` to `operator-common` in preparation for it to be used by the [Unidirectional Topic Operator](https://github.com/strimzi/proposals/blob/main/051-unidirectional-topic-operator.md). The actual logic for liveness and readiness is abstracted into `Liveness` and `Readiness` interfaces, instances of which are passed to the `HealthCheckAndMetricsServer` constructor. Both of these and also the `MetricsProvider` may now be null.

I also reduced the visibility of the inner classes, added a bit of logging and improved the coverage of the test.

### Checklist

- [x] Write tests


scholzj (approved):
LGTM. I wonder if `http` is the right package name -> it is not really any generic HTTP stuff but health check and metrics server. But its not ike I have any better ide which would be short 🤔 .

tombentley:
> I wonder if `http` is the right package name

I agree, it's not _quite_ right. Happy to change it we can agree to something better. It's almost `observability`, or `management`, or something, but those suffer from being a bit unspecific.

tombentley:
What about `scrape`?

scholzj:
I don't think anyone without prior knowledge would understand what it means.

operator-common/src/main/java/io/strimzi/operator/common/http/Liveness.java
```diff
@@ -0,0 +1,23 @@
...
+package io.strimzi.operator.common.http;
+
+/**
+ * A liveness check implemented by an application (e.g. an operator) and called by the
+ * {@link HealthCheckAndMetricsServer} when handling a health check request.
+ */
+public interface Liveness {
+
+    /**
+     * Indicates whether the controller is alive or not.
```
ppatierno, line 14: isn't "the application" better? You are being generic later about it and then adding an example usage with a controller, or?

operator-common/src/main/java/io/strimzi/operator/common/http/Readiness.java
```diff
@@ -0,0 +1,22 @@
...
+package io.strimzi.operator.common.http;
+
+/**
+ * A readiness check implemented by an application (e.g. an operator) and called by the
+ * {@link HealthCheckAndMetricsServer} when handling a health check request.
+ */
+public interface Readiness {
+
+    /**
+     * Indicates whether the controller is ready.
```
ppatierno, line 14: Ditto as above.

ppatierno (approved):
LGTM. Just left a couple of nits.

user-operator/src/main/java/io/strimzi/operator/user/Main.java
```diff
@@ -99,7 +100,7 @@ public static void main(String[] args) {
         );
 
         // Create the health check and metrics server
-        HealthCheckAndMetricsServer healthCheckAndMetricsServer = new HealthCheckAndMetricsServer(controller, metricsProvider);
+        HealthCheckAndMetricsServer healthCheckAndMetricsServer = new HealthCheckAndMetricsServer(controller, controller, metricsProvider);
```
fvaleri, line 103: We could create a common interface that extends Liveness, Readiness and MetricsProvider. Then we could use it as an additional one-argument constructor for the HealthCheckAndMetricsServer. WDYT?

PR #8573, opened 2023-05-26
docs(volumes): clarifies no risk to data loss when resizing PVs

PaulRMellor:
### Documentation

Update to docs that makes it clear that resizing persistent volumes presents no risk to data loss

### Checklist

- [ ] Write tests
- [x] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/documentation/modules/configuring/proc-resizing-persistent-volumes.adoc
+++ b/documentation/modules/configuring/proc-resizing-persistent-volumes.adoc
@@ -6,8 +6,9 @@
 = Resizing persistent volumes
 
 [role="_abstract"]
-You can provision increased storage capacity by increasing the size of the persistent volumes used by an existing Strimzi cluster.
-Resizing persistent volumes is supported in clusters that use either a single persistent volume or multiple persistent volumes in a JBOD storage configuration.
+Persistent volumes used by a Strimzi cluster can be resized without any risk of data loss. 
+This capability allows for expanding storage capacity by increasing the size of existing persistent volumes. 
+Resizing is supported in clusters with either a single persistent volume or multiple persistent volumes configured in JBOD storage.
 
 NOTE: You can increase but not decrease the size of persistent volumes.
 Decreasing the size of persistent volumes is not currently supported in Kubernetes.
--- a/documentation/modules/configuring/proc-resizing-persistent-volumes.adoc
+++ b/documentation/modules/configuring/proc-resizing-persistent-volumes.adoc
@@ -10,7 +10,7 @@ Persistent volumes used by a cluster can be resized without any risk of data los
 Following a configuration update to change the size of the storage, Strimzi instructs the storage infrastructure to make the change. 
 Storage expansion is supported in Strimzi clusters that use a single persistent volume per-broker or multiple persistent volumes per-broker that use JBOD storage. 
 
-NOTE: Storage reduction is only possible with JBOD, and only after moving all partitions to other volumes within the same broker (intra-broker) or to other brokers within the same cluster (intra-cluster).
+NOTE: Storage reduction is only possible with JBOD. You can remove a disk after moving all partitions on the disk to other volumes within the same broker (intra-broker) or to other brokers within the same cluster (intra-cluster).
 
 .Prerequisites
 
--- a/documentation/modules/configuring/proc-resizing-persistent-volumes.adoc
+++ b/documentation/modules/configuring/proc-resizing-persistent-volumes.adoc
@@ -6,12 +6,11 @@
 = Resizing persistent volumes
 
 [role="_abstract"]
-Persistent volumes used by a Strimzi cluster can be resized without any risk of data loss. 
-This capability allows for expanding storage capacity by increasing the size of existing persistent volumes. 
-Resizing is supported in clusters with either a single persistent volume or multiple persistent volumes configured in JBOD storage.
+Persistent volumes used by a cluster can be resized without any risk of data loss, as long as the storage infrastructure supports it. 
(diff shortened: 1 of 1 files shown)


documentation/modules/configuring/proc-resizing-persistent-volumes.adoc
```diff
@@ -6,8 +6,9 @@
 = Resizing persistent volumes
 
 [role="_abstract"]
-You can provision increased storage capacity by increasing the size of the persistent volumes used by an existing Strimzi cluster.
-Resizing persistent volumes is supported in clusters that use either a single persistent volume or multiple persistent volumes in a JBOD storage configuration.
+Persistent volumes used by a Strimzi cluster can be resized without any risk of data loss. 
+This capability allows for expanding storage capacity by increasing the size of existing persistent volumes. 
+Resizing is supported in clusters with either a single persistent volume or multiple persistent volumes configured in JBOD storage.
```
fvaleri, line 11:
```suggestion
Storage expansion is supported in clusters with either a single persistent volume per-broker or multiple persistent volumes per-broker configured with JBOD. Storage reduction is only supported with JBOD, after all partitions have been moved to other volumes of the same broker (intra-broker) or other brokers of the same cluster (intra-cluster).
```

documentation/modules/configuring/proc-resizing-persistent-volumes.adoc
```diff
@@ -6,8 +6,9 @@
 = Resizing persistent volumes
 
 [role="_abstract"]
-You can provision increased storage capacity by increasing the size of the persistent volumes used by an existing Strimzi cluster.
-Resizing persistent volumes is supported in clusters that use either a single persistent volume or multiple persistent volumes in a JBOD storage configuration.
+Persistent volumes used by a Strimzi cluster can be resized without any risk of data loss. 
+This capability allows for expanding storage capacity by increasing the size of existing persistent volumes.
```
fvaleri, line 10: I would remove this line which is not complete (see the the other comment).

documentation/modules/configuring/proc-resizing-persistent-volumes.adoc
scholzj, line 11: I think there should be two things mentioned here:
* The resizing needs to be supported by the infrastructure
* Strimzi really just tells the infrastructure to resize the volumes

fvaleri (changes requested):

documentation/modules/configuring/proc-resizing-persistent-volumes.adoc
scholzj: I would not mix removing JBOD volumes into this. It is something completely different than resizing volumes.

fvaleri: > I would not mix removing JBOD volumes into this. It is something completely different than resizing volumes.

Reducing capacity is resizing. I think it is worth mentioning, also considering that the "Removing volumes from JBOD storage" rocedure is explained few paragraph below.

scholzj: Its not resizing of persistent volumes which this procedure is about.

fvaleri: Ok, I see what you mean. Let's remove "Storage reduction..", but I would keep the rest.

PaulRMellor: we do have a note after this intro that talks about decreasing the size of persistent volumes. I can expand that to give this content.

PaulRMellor: I've replaced what we had in the note with this, as it was contradicting what we had originally.

PaulRMellor: See what you think it the update. We can remove the note if it gets in the way.

scholzj: Sorry, but I think this is confusing as it suggests you can empty the disk and then shrink it. You should at least add there that the _reduction_ happens through _removing the whole disk_ and not by shrinking it.

PaulRMellor: I've changed the note. We're talking about resizing in this procedure and showing how to expand, but I think a note on reduction is helpful.

> NOTE: Storage reduction is only possible with JBOD. You can remove a disk after moving all partitions on the disk to other volumes within the same broker (intra-broker) or to other brokers within the same cluster (intra-cluster).

documentation/modules/configuring/proc-resizing-persistent-volumes.adoc
```diff
@@ -6,11 +6,11 @@
 = Resizing persistent volumes
 
 [role="_abstract"]
-You can provision increased storage capacity by increasing the size of the persistent volumes used by an existing Strimzi cluster.
-Resizing persistent volumes is supported in clusters that use either a single persistent volume or multiple persistent volumes in a JBOD storage configuration.
+Persistent volumes used by a cluster can be resized without any risk of data loss, as long as the storage infrastructure supports it. 
+Following a configuration update to change the size of the storage, Strimzi instructs the storage infrastructure to make the change. 
+Storage expansion is supported in Strimzi clusters that use a single persistent volume per-broker or multiple persistent volumes per-broker that use JBOD storage.
```
scholzj, line 11: This is confusing because you can have JBOD storage with a single volume.

```suggestion
Storage expansion is supported in Strimzi clusters that use persistent-claim volumes. 
```

scholzj (approved):

fvaleri (approved):
LGTM. Thanks.

PR #8577, opened 2023-05-29
Fix failing tests when HTTP/HTTPS proxy env vars are set

fvaleri:
Even if you are not behind a proxy, you may have NO_PROXY set. For example, on Fedora 36 with Gnome the default configuration is empty proxy with NO_PROXY set. In general, many corporate developers are behind proxies, so this would be useful for them.

Related test classes:

- KafkaConnectClusterTest
- KafkaExporterTest
- EntityTopicOperatorTest
- EntityUserOperatorTest
- KafkaBridgeClusterTest
- KafkaMirrorMaker2ClusterTest
- KafkaMirrorMakerClusterTest
- CruiseControlTest

Because of that, KafkaConnectAssemblyOperatorPodSetTest.testBuildChangedCluster was also failing with a wrong image build hash.

### DIFF
--- a/cluster-operator/src/test/java/io/strimzi/operator/cluster/TestUtils.java
+++ b/cluster-operator/src/test/java/io/strimzi/operator/cluster/TestUtils.java
@@ -9,9 +9,11 @@ import io.fabric8.kubernetes.api.model.ConfigMapBuilder;
 import io.fabric8.kubernetes.api.model.ConfigMapKeySelectorBuilder;
 import io.fabric8.kubernetes.api.model.Container;
 import io.fabric8.kubernetes.api.model.EnvVar;
+import io.fabric8.kubernetes.api.model.EnvVarBuilder;
 import io.strimzi.api.kafka.model.JmxPrometheusExporterMetrics;
 import io.strimzi.api.kafka.model.JmxPrometheusExporterMetricsBuilder;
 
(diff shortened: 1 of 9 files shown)


cluster-operator/src/test/java/io/strimzi/operator/cluster/TestUtils.java
```diff
@@ -54,4 +56,30 @@ public static Map<String, String> containerEnvVars(Container container) {
                         // On duplicates, last-in wins
                         (u, v) -> v));
     }
+
+    /**
+     * If present, adds the HTTP/HTTPS proxy environment variables to the list.
```
scholzj, line 61: Maybe you can add a bit more here to better explain why and when it is needed => when your environment has these environment variables already set because running behind proxy etc.

scholzj (approved):
It is a bit weird problem -> but it is not easy to ensure a clean test environment when it comes to environment variables. So I guess there is nothing better what one can do to address this issue.

ppatierno (approved):
LGTM.

cluster-operator/src/test/java/io/strimzi/operator/cluster/TestUtils.java
fvaleri: Done. Thanks.

PR #8601, opened 2023-06-01
Avoid rolling broker in log recovery

tinaselenge:
### Type of change

- Enhancement / new feature

### Description

Implement the final part of [Strimzi proposal 48](https://github.com/strimzi/proposals/blob/main/048-avoid-broker-restarts-when-in-recovery.md) to avoid broker restart when in log recovery. The KafkaRoller will now use the KafkaAgent's REST API to determine the broker state and log recovery progress. When the broker is in log recovery, instead of force restarting it, KafkaRoller will wait until the operational timeout is reached. If the log recovery is not completed during a reconciliation, that reconciliation will fail. Reconciliations will keep failing until the log recovery is finished. 

### Checklist

- [X] Write tests
- [ ] Update documentation
- [x] Update CHANGELOG.md


cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/resource/KafkaRoller.java
```diff
@@ -187,6 +207,69 @@ protected Future<Pod> pod(Integer podId) {
...
+            var result = completableFuture.get(3, TimeUnit.SECONDS);
+            brokerstate = (int) result.get("brokerState");
+            if (brokerstate == BROKER_RECOVERY_STATE && result.containsKey("recoveryState")) {
+                LOGGER.infoCr(reconciliation, podName + " is in log recovery. The current recovery state: " + result.get("recoveryState"));
+            }
+        } catch (TimeoutException | InterruptedException | ExecutionException e) {
+            LOGGER.warnCr(reconciliation, "Failed to get broker state", e);
+        }
+        return brokerstate;
+    }
```
scholzj, line 271: I wonder if it would make sense to factor this out into a separate class. Might make the code a bit cleaner and make it easier to mock it in tests etc.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/resource/KafkaRoller.java
```diff
@@ -371,6 +454,9 @@ private void restartIfNecessary(NodeRef nodeRef, RestartContext restartContext)
                 await(isReady(pod), operationTimeoutMs, TimeUnit.MILLISECONDS, e -> new RuntimeException(e));
             } catch (Exception e)   {
                 if (e.getCause() instanceof TimeoutException) {
+                    if (getBrokerState(pod.getMetadata().getName()) == (BROKER_RECOVERY_STATE)) {
+                        throw new UnforceableProblem("Pod " + nodeRef.podName() + " is not ready because the broker is performing log recovery.", e.getCause());
+                    }
```
scholzj, line 459: I wonder if we should try the check for every exception and not just timeout ...

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/resource/KafkaRoller.java
```diff
@@ -371,6 +454,9 @@ private void restartIfNecessary(NodeRef nodeRef, RestartContext restartContext)
                 await(isReady(pod), operationTimeoutMs, TimeUnit.MILLISECONDS, e -> new RuntimeException(e));
             } catch (Exception e)   {
                 if (e.getCause() instanceof TimeoutException) {
+                    if (getBrokerState(pod.getMetadata().getName()) == (BROKER_RECOVERY_STATE)) {
+                        throw new UnforceableProblem("Pod " + nodeRef.podName() + " is not ready because the broker is performing log recovery.", e.getCause());
```
scholzj, line 458: Didn't we planned to log some progress here as well?

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/resource/KafkaRoller.java
```diff
@@ -126,6 +130,14 @@ public class KafkaRoller {
...
     private final boolean allowReconfiguration;
     private Admin allClient;
+    private static final String BROKER_STATE_REST_PATH = "/v1/broker-state";
+    private static final int BROKER_STATE_HTTPS_PORT = 8443;
+    private static final int BROKER_RECOVERY_STATE = 2;
+    private PemTrustOptions pto;
+    private PemKeyCertOptions pkco;
+    private static final String CLUSTER_CA_CERT_SECRET_KEY = "ca.crt";
+    private static final String CO_KEY_SECRET_KEY = "cluster-operator.key";
+    private static final String CO_KEY_CERT_SECRET_KEY = "cluster-operator.crt";
```
scholzj, line 140: Maybe the `private static final` variable can be defined up top with the other `static` variables?

scholzj (commented):
Thanks for the PR.

I think one aspect which is missing is the network policies -> you will need to update the part of the `KafkaCluster` class which generates the network policies to grant the cluster operator access to the 8443 port of the healthcheck.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/resource/KafkaRoller.java
```diff
@@ -187,6 +207,69 @@ protected Future<Pod> pod(Integer podId) {
     private final ConcurrentHashMap<String, RestartContext> podToContext = new ConcurrentHashMap<>();
     private Function<Pod, RestartReasons> podNeedsRestart;
 
+    private HttpClientOptions getHttpClientOptions() {
+        return new HttpClientOptions()
+                .setLogActivity(true)
+                .setVerifyHost(true)
+                .setSsl(true)
+                .setVerifyHost(true)
```
ppatierno, line 215: `setVerifyHost` called twice. Not a big deal but useless.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/resource/KafkaRoller.java
```diff
@@ -187,6 +207,69 @@ protected Future<Pod> pod(Integer podId) {
...
+                .setPemTrustOptions(pto)
+                .setKeyCertOptions(pkco);
+    }
+
+    protected <T> Future<Map<String, Object>> doGet(String host) {
+        LOGGER.debugCr(reconciliation, "Making GET request to {}", BROKER_STATE_REST_PATH);
+        return HttpClientUtils.withHttpClient(vertx, getHttpClientOptions(), (httpClient, result) ->
+                httpClient.request(HttpMethod.GET, BROKER_STATE_HTTPS_PORT, host, BROKER_STATE_REST_PATH, request -> {
+                    if (request.succeeded()) {
+                        request.result().setFollowRedirects(true)
```
ppatierno, line 225: AFAIK the `setFollowRedirects` is true by default?

tinaselenge: It didn't seem to work if I don't set the setFollowRedirects to true when I was testing. Also from vertx doc, I didn't see it mentioning that it is set to true by default but possible I missed it.

tinaselenge:
> Thanks for the PR.
> 
> I think one aspect which is missing is the network policies -> you will need to update the part of the `KafkaCluster` class which generates the network policies to grant the cluster operator access to the 8443 port of the healthcheck.

@scholzj Thank you for reviewing the PR. For network policy, looks like I just need to add an ingress rule for the new port. Is this right?

```
public NetworkPolicy generateNetworkPolicy(String operatorNamespace, Labels operatorNamespaceLabels) {

        rules.add(NetworkPolicyUtils.createIngressRule(BROKERSTATE_PORT, List.of(clusterOperatorPeer)));

```

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/resource/BrokerState.java
```diff
@@ -0,0 +1,51 @@
...
+ * License: Apache License 2.0 (see the file LICENSE or http://apache.org/licenses/LICENSE-2.0.html).
+ */
+package io.strimzi.operator.cluster.operator.resource;
+
+import java.util.Map;
+
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+public class BrokerState {
```
tombentley, line 12: Add Javadoc to explain what this class is for. E.g. "Java representation of the JSON response from the /v1/broker-state endpoint of the KafkaAgent"

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/resource/BrokerStateProvider.java
```diff
@@ -0,0 +1,122 @@
...
+import io.strimzi.operator.common.Util;
+import io.vertx.core.Future;
+import io.vertx.core.Vertx;
+import io.vertx.core.buffer.Buffer;
+import io.vertx.core.http.HttpClientOptions;
+import io.vertx.core.http.HttpMethod;
+import io.vertx.core.net.PemKeyCertOptions;
+import io.vertx.core.net.PemTrustOptions;
+
+public class BrokerStateProvider {
```
tombentley, line 29: This feels more like a `BrokerStateClient` to me. Or even a `KafkaAgentClient` which currently only exposes a method for interacting with the broker state endpoint.

Also, does this need to be `public`?

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/resource/BrokerStateProvider.java
```diff
@@ -0,0 +1,122 @@
...
+    private static final String CO_KEY_SECRET_KEY = "cluster-operator.key";
+    private static final String CO_KEY_CERT_SECRET_KEY = "cluster-operator.crt";
+
+    private Vertx vertx;
+    private Reconciliation reconciliation;
+    private String cluster;
+    private PemTrustOptions pto;
+    private PemKeyCertOptions pkco;
+
+    BrokerStateProvider(Reconciliation reconciliation, Vertx vertx, String cluster, String namespace, Secret clusterCaCertSecret, Secret coKeySecret) {
```
tombentley, line 46: Given that we're trying to move away from using Vertx I wonder if we should use [`java.net.http.HttpClient`](https://docs.oracle.com/en/java/javase/11/docs/api/java.net.http/java/net/http/HttpClient.html). `doGet()` would then be blocking, returning a `String`, rather than a `Future<String>`, but that seems fine since you're already doing a blocking get on line 112, so you're not really benefitting from the fact that it returns a future anyway. Wdyt @ppatierno @scholzj?

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/resource/BrokerStateProvider.java
```diff
@@ -0,0 +1,122 @@
...
+            if (future.succeeded()) {
+                completableFuture.complete(future.result());
+            } else {
+                completableFuture.completeExceptionally(future.cause());
+            }
+        });
+
+        try {
+            var result = completableFuture.get(3, TimeUnit.SECONDS);
+            ObjectMapper mapper = new ObjectMapper();
```
tombentley, line 113: `ObjectMapper` is thread safe, so this could be a constant.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/resource/BrokerStateProvider.java
```diff
@@ -0,0 +1,122 @@
...
+            brokerstate = mapper.readValue(result, BrokerState.class);
+        } catch (TimeoutException | InterruptedException | ExecutionException e) {
+            LOGGER.warnCr(reconciliation, "Failed to get broker state", e);
+        } catch (JsonProcessingException jsonException) {
+            LOGGER.warnCr(reconciliation, "Failed to parse response for broker state", jsonException);
+        }
```
tombentley, line 119: If you're returning `new BrokerState(-1, null)` in the event of these errors you should at least state that in the javadoc of this method, since it's not obvious that that would be the contract.

tombentley (commented):
Thanks! I left a few comments.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/resource/BrokerStateProvider.java
ppatierno: I agree, yes.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/resource/KafkaAgentClient.java
```diff
@@ -0,0 +1,117 @@
...
+     * @param podName Name of the pod to interact with
+     * @return A BrokerState that contains broker state and recovery progress.
+     *         -1 is returned for broker state if the http request failed or returned non 200 response.
+     *         Null value is returned for recovery progress if broker state is not 2 (RECOVERY).
+     */
+    public BrokerState getBrokerState(String podName) {
+        BrokerState brokerstate = new BrokerState(-1, null);
+        String host = DnsNameGenerator.podDnsName(namespace, KafkaResources.brokersServiceName(cluster), podName);
+        try {
+            URI uri = new URI("https://" +  host + ":" + BROKER_STATE_HTTPS_PORT + BROKER_STATE_REST_PATH);
```
tombentley, line 110: You should use the `new URI("https", null, host, BROKER_STATE_HTTPS_PORT, BROKER_STATE_REST_PATH, null, null)` constructor  (which will avoid the URI constructor having to parse a single URI into its constituent parts).

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/resource/BrokerState.java
```diff
@@ -0,0 +1,75 @@
...
+     * @param recoveryState Map that has the number of remaining logs and segments to recover
+     */
+    @JsonCreator
+    public BrokerState(@JsonProperty("brokerState")int code, @JsonProperty("recoveryState") Map<String, Object> recoveryState) {
+        this.code = code;
+        this.recoveryState = recoveryState;
+    }
+
+    /**
+     * Integer that represents the broker state
```
tombentley, line 34:
```suggestion
     * Integer that represents the broker state, or -1 if there was an error when getting the broker state.
```

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/resource/BrokerState.java
```diff
@@ -0,0 +1,75 @@
...
+
+import java.util.Map;
+
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+/**
+ * Java representation of the JSON response from the /v1/broker-state endpoint of the KafkaAgent
+ */
+public class BrokerState {
```
tombentley, line 15: I guess this could also be package access?

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/resource/KafkaAgentClient.java
```diff
@@ -0,0 +1,117 @@
...
+
+
+            HttpRequest req = HttpRequest.newBuilder()
+                    .uri(uri)
+                    .GET()
+                    .build();
+
+            var response =  HttpClient.newBuilder()
+                    .sslContext(sslContext)
+                    .build()
```
tombentley, line 87: I think it would be more natural to configure the HttpClient once, initializing a field in the constructor. Then this `doGet` method just creates the request and handles the response. That would allow to reuse the same `HttpClient` between calls.

tombentley (commented):
Thanks, I left a few more comments.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/resource/KafkaAgentClient.java
```diff
@@ -0,0 +1,131 @@
...
+        this.coKeySecret = coKeySecret;
+    }
+
+    private HttpClient createHttpClient() throws GeneralSecurityException, IOException {
+
+        SSLContext sslContext = SSLContext.getInstance("TLSv1.3");
+        KeyStore trustStore = KeyStore.getInstance("PKCS12");
+        trustStore.load(new ByteArrayInputStream(
+                        Util.decodeFromSecret(clusterCaCertSecret, "ca.p12")),
+                new String(Util.decodeFromSecret(clusterCaCertSecret, "ca.password"), StandardCharsets.UTF_8).toCharArray());
```
scholzj, line 64: Can we use the PEM certificates here in some easy way:?

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/resource/KafkaAgentClient.java
```diff
@@ -0,0 +1,131 @@
...
+        this.reconciliation = reconciliation;
+        this.cluster = cluster;
+        this.namespace = namespace;
+        this.clusterCaCertSecret = clusterCaCertSecret;
+        this.coKeySecret = coKeySecret;
+    }
+
+    private HttpClient createHttpClient() throws GeneralSecurityException, IOException {
+
+        SSLContext sslContext = SSLContext.getInstance("TLSv1.3");
```
scholzj, line 60: Does this in any way conflict with FIPS? When user configures Kafka to use another TLS version in its security context, will the Agent be independent and work fine with TLSv1.3?

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/resource/KafkaAgentClient.java
```diff
@@ -0,0 +1,131 @@
...
+    protected String doGet(URI uri) {
+        try {
+            HttpRequest req = HttpRequest.newBuilder()
+                    .uri(uri)
+                    .GET()
+                    .build();
+
+            if (httpClient == null) {
+                httpClient = createHttpClient();
+            }
```
scholzj, line 96: What is the reason for not doing this in constructor? If there is some, might be good to add a comment with explanation.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/resource/KafkaRoller.java
```diff
@@ -367,6 +368,10 @@ private void restartIfNecessary(NodeRef nodeRef, RestartContext restartContext)
             throw new UnforceableProblem("Error getting pod " + nodeRef.podName(), e);
         }
 
+        if (kafkaAgentClient == null) {
+            this.kafkaAgentClient = initKafkaAgentClient();
+        }
```
scholzj, line 373: Same as in the other class. Why not do this in the constructor?

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/resource/KafkaRoller.java
```diff
@@ -375,7 +380,12 @@ private void restartIfNecessary(NodeRef nodeRef, RestartContext restartContext)
             LOGGER.debugCr(reconciliation, "Waiting for pod {} to become ready before checking its state", nodeRef.podName());
             try {
                 await(isReady(pod), operationTimeoutMs, TimeUnit.MILLISECONDS, e -> new RuntimeException(e));
-            } catch (Exception e)   {
+            } catch (Exception e) {
+                BrokerState brokerState = kafkaAgentClient.getBrokerState(pod.getMetadata().getName());
```
scholzj, line 384: If the `kafkaAgentClient` is used only here and if the `kafkaAgentClient` is not initialized in the constructor -> would it make sense to initialize it only here?

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/resource/KafkaAgentClient.java
tinaselenge: The reason I didn't it put this in the constructor was to handle the exceptions in getBrokerState(). However thinking about this again, we probably should put this in the constructor and throw the exceptions in KafkaRoller. We don't want to throw exceptions in KafkaRoller only when we actually attempt to connect to KafkaAgent for compatibility reason. However, we should be able to able initialise the client whether KafkaAgent is available or not.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/resource/KafkaRoller.java
tinaselenge: Similar reason above applies to this. I wasn't sure if when we can't initialise kafkaAgentClient, that should cause the reconciliation to start. Admin clients are not initialised from the constructor, so kafkaAgentClient probably should not either.

tinaselenge: I can add comments to clarify this if you agree

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/resource/KafkaRoller.java
tinaselenge: yes I agree with this. I will move it to here.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/resource/KafkaRoller.java
scholzj: Ok, thanks.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/resource/KafkaAgentClient.java
tinaselenge: Yes I think we can use the PEM certificates instead. Out of curiosity, can I ask why we prefer to use PEM certificates?

scholzj: It is easier to use and more Kubernetes-friendly format. So using them might be easier for future work around certificate management etc.

scholzj: I mean ... if it would be something super complicated, you should not do it in your PR, we can deal with it later. But if it works easily, it would be nice to use it.

tinaselenge: As you suggested, I will do this in a separate PR because I would have to make quite few changes.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/resource/KafkaAgentClient.java
```diff
@@ -0,0 +1,146 @@
...
+        this.reconciliation = reconciliation;
+        this.cluster = cluster;
+        this.namespace = namespace;
+        this.clusterCaCertSecret = clusterCaCertSecret;
+        this.coKeySecret = coKeySecret;
+        this.httpClient = createHttpClient();
+    }
+
+    //visible for testing
+    KafkaAgentClient(Reconciliation reconciliation, String cluster, String namespace) {
```
scholzj, line 60: We usually do something like this when you need to expose some method for testing:

```suggestion
    /* test */ KafkaAgentClient(Reconciliation reconciliation, String cluster, String namespace) {
```

scholzj (approved):
LGTM.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/resource/KafkaAgentClient.java
scholzj: Ok, can you open an issue for it please so that it is not forgotten?

scholzj:
Can you please resolve the conflict so that I can run the system tests.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/resource/KafkaAgentClient.java
```diff
@@ -0,0 +1,145 @@
...
+                    new String(Util.decodeFromSecret(clusterCaCertSecret, "ca.password"), StandardCharsets.UTF_8).toCharArray());
+
+            String trustManagerFactoryAlgorithm = TrustManagerFactory.getDefaultAlgorithm();
+            TrustManagerFactory trustManagerFactory = TrustManagerFactory.getInstance(trustManagerFactoryAlgorithm);
+            trustManagerFactory.init(trustStore);
+            char[] keyPassword = new String(Util.decodeFromSecret(coKeySecret, "cluster-operator.password"), StandardCharsets.UTF_8).toCharArray();
+            KeyStore coKeyStore = KeyStore.getInstance("PKCS12");
+            coKeyStore.load(new ByteArrayInputStream(
+                            Util.decodeFromSecret(coKeySecret, "cluster-operator.p12")),
+                    keyPassword
```
ppatierno, line 84: why you used a variable here instead of inlining the password recovery as for the truststore?

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/resource/KafkaAgentClient.java
ppatierno: I am confused. Now the it's done in the constructor, do we still need this code here?

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/resource/KafkaAgentClient.java
```diff
@@ -0,0 +1,145 @@
...
+
+            return HttpClient.newBuilder()
+                    .sslContext(sslContext)
+                    .build();
+        } catch (GeneralSecurityException | IOException e) {
+            throw new RuntimeException("Failed to configure HTTP client", e);
+        }
+    }
+
+    protected String doGet(URI uri) {
```
ppatierno, line 101: any specific reason to be `protected`?

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/resource/KafkaAgentClient.java
tinaselenge: created here https://github.com/strimzi/strimzi-kafka-operator/issues/8721

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/resource/KafkaAgentClient.java
tinaselenge: It's also used in line 89 for initialising the keyManagerFactory.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/resource/KafkaAgentClient.java
tinaselenge: ah good spot, I forgot to delete it here.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/resource/KafkaAgentClient.java
ppatierno: Oh thanks, I missed that!

ppatierno (approved):
LGTM!

scholzj:
@tinaselenge Looks like the build is filing because of:
```
[ERROR] /home/vsts/work/1/s/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/resource/KafkaAgentClient.java:101:6: 'method def modifier' has incorrect indentation level 5, expected level should be 4. [Indentation]
Audit done.
```

tinaselenge:
@scholzj  thanks, I thought I fixed it but had not pushed it up.

tombentley (approved):
LGTM. Thanks!

scholzj:
Thanks for the PR.

PR #8603, opened 2023-06-01
Add microshift to list of allowed systemtest clusters

kornys:
### Type of change

Add microshift into allowed list of clusters to systemtest

### Description

Currently I'm unable to trigger tests on microshift due to error 
```
NoCluster Unable to find a cluster; tried [minikube, kubectl, oc]
```
This PR adds microshift into list.


Frawless (approved):
We should also identify tests that are not executable on microshift like connect build, etc, but it is not needed in scope of this PR.

im-konge (approved):
LGTM. Thanks for the PR :)

scholzj:
Should we merge this?

im-konge:
Yeah, will do.

PR #8622, opened 2023-06-06
Fixed recovery of new CA certificate trusting phase on CO stopped

### LINKED ISSUE #8402: Certificate key replacement fails when Cluster Operator crashes before the trust is established
When a Cluster CA private key is replaced, the operator needs to follow through a series of three different rolling updates:

1. First rolling update establishes the trust to the new CA while still using the old server certificates and trusting the old CA as well
2. Second rolling update issues and rolls out the new server certificates signed by the new CA (while still trusting the old CA as well)
3. Third rolling update removes the trust in the old CA

When the operator crashes before the first step is completed, it can happen that the operands will not trust the new CA. however, the operator will not detect it and recover from it and instead proceed with phase 2 and roll out the new certificates. This will not work, because trust will not be established and the operands will not sync during the rolling updates. If the trust is not rolled only to EO, KE or CC, it might still recover later. But if it fails before trust is rolled out to Zoo or Kafka, the clusters might not get back online anymore.

Note: This bug is present in the code already for some time :-/

Note 2: The description above describes the ideal state. But right now, due to #8401, phase 2 will actually pass successfully as it does not renew the certificates and everything will fall apart only in phase 3 when the trust is removed.

ppatierno:
### Type of change

- Bugfix

### Description

This PR fixes #8402.
As discussed in the issue, it works by adding a new `strimzi.io/cluster-ca-key-trusted-generation` annotation on the pods which has to match the `strimzi.io/strimzi.io/cluster-ca-key-generation` in order to consider the "trusting" phase to be completed.
This is done as the first step in the `CaReconciler` to check such annotation on all the pods. If it's not consistent with the new CA key generation, pods should continue to roll for the "trusting" phase. During it, even the CO Secret doesn't have to be updated otherwise if something goes wrong it will already get the new signed certs and won't be able to connect to Kafka and ZooKeeper pods anymore.

### Checklist

- [x] Write tests

### DIFF
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/CaReconciler.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/CaReconciler.java
@@ -64,11 +64,11 @@ import java.util.function.Function;
 public class CaReconciler {
     private static final ReconciliationLogger LOGGER = ReconciliationLogger.create(CaReconciler.class.getName());
 
-    private final Reconciliation reconciliation;
+    /* test */ final Reconciliation reconciliation;
     private final Vertx vertx;
     private final long operationTimeoutMs;
 
-    private final DeploymentOperator deploymentOperator;
+    /* test */ final DeploymentOperator deploymentOperator;
     private final StrimziPodSetOperator strimziPodSetOperator;
     private final SecretOperator secretOperator;
     private final PodOperator podOperator;
@@ -92,6 +92,9 @@ public class CaReconciler {
     private ClusterCa clusterCa;
     private ClientsCa clientsCa;
     private Secret oldCoSecret;
+
+    /* test */ boolean isClusterCaNeedFullTrust;
+    /* test */ boolean isClusterCaFullyUsed;
 
     /**
      * Constructs the CA reconciler which reconciles the Cluster and Client CAs
@@ -196,6 +199,7 @@ public class CaReconciler {
      */
     public Future<CaReconciliationResult> reconcile(Clock clock)    {
         return reconcileCas(clock)
+                .compose(i -> verifyClusterCaFullyTrustedAndUsed())
                 .compose(i -> clusterOperatorSecret(clock))
                 .compose(i -> rollingUpdateForNewCaKey())
                 .compose(i -> maybeRemoveOldClusterCaCertificates())
@@ -328,6 +332,10 @@ public class CaReconciler {
 
     Future<Void> clusterOperatorSecret(Clock clock) {
         oldCoSecret = clusterCa.clusterOperatorSecret();
+        if (oldCoSecret != null && this.isClusterCaNeedFullTrust) {
+            LOGGER.warnCr(reconciliation, "Cluster CA needs to be fully trusted across the cluster, keeping current CO secret and certs");
+            return Future.succeededFuture();
+        }
         Secret secret = ModelUtils.buildSecret(
                 reconciliation,
                 clusterCa,
@@ -353,7 +361,9 @@ public class CaReconciler {
     Future<Void> rollingUpdateForNewCaKey() {
         RestartReasons podRollReasons = RestartReasons.empty();
 
-        if (clusterCa.keyReplaced()) {
+        // cluster CA needs to be fully trusted
+        // it is coming from a cluster CA key replacement which didn't end successfully (i.e. CO stopped) and we need to continue from there
+        if (clusterCa.keyReplaced() || isClusterCaNeedFullTrust) {
             podRollReasons.add(RestartReason.CLUSTER_CA_CERT_KEY_REPLACED);
         }
 
@@ -375,6 +385,77 @@ public class CaReconciler {
     }
 
     /**
+     * Gather the Kafka related components pods for checking CA key trust and CA certificate usage to sign servers certificate.
+     *
+     * Verify that all the pods are already trusting the new CA certificate signed by a new CA key.
+     * It checks each pod's CA key generation, compared with the new CA key generation.
+     * When the trusting phase is not completed (i.e. because CO stopped), it needs to be recovered from where it was left.
+     *
+     * Verify that all pods are already using the new CA certificate to sign server certificates.
+     * It checks each pod's CA certificate generation, compared with the new CA certificate generation.
+     * When the new CA certificate is used everywhere, the old CA certificate can be removed.
+     */
+    /* test */ Future<Void> verifyClusterCaFullyTrustedAndUsed() {
+        isClusterCaNeedFullTrust = false;
+        isClusterCaFullyUsed = true;
+
+        // Building the selector for Kafka related components
+        Labels labels =  Labels.forStrimziCluster(reconciliation.name()).withStrimziKind(Kafka.RESOURCE_KIND);
+
+        return podOperator.listAsync(reconciliation.namespace(), labels)
+                .compose(pods -> {
+
+                    // still no Pods, a new Kafka cluster is under creation
+                    if (pods.isEmpty()) {
+                        isClusterCaFullyUsed = false;
+                        return Future.succeededFuture();
+                    }
+
+                    int clusterCaCertGeneration = clusterCa.certGeneration();
+                    int clusterCaKeyGeneration = clusterCa.keyGeneration();
+
+                    LOGGER.debugCr(reconciliation, "Current cluster CA cert generation {}", clusterCaCertGeneration);
+                    LOGGER.debugCr(reconciliation, "Current cluster CA key generation {}", clusterCaKeyGeneration);
+
+
+                    for (Pod pod : pods) {
+                        // with "RollingUpdate" strategy on Deployment(s) (i.e. the Cruise Control one),
+                        // while the Deployment is reported to be ready, the old pod is still alive but terminating
+                        // this condition is for skipping "Terminating" pods for checks on the CA key and old certificates
+                        if (pod.getMetadata().getDeletionTimestamp() == null) {
+                            int podClusterCaCertGeneration = Annotations.intAnnotation(pod, Ca.ANNO_STRIMZI_IO_CLUSTER_CA_CERT_GENERATION, clusterCaCertGeneration);
+                            LOGGER.debugCr(reconciliation, "Pod {} has cluster CA cert generation {}", pod.getMetadata().getName(), podClusterCaCertGeneration);
+
+                            int podClusterCaKeyGeneration = Annotations.intAnnotation(pod, Ca.ANNO_STRIMZI_IO_CLUSTER_CA_KEY_GENERATION, clusterCaKeyGeneration);
+                            LOGGER.debugCr(reconciliation, "Pod {} has cluster CA key generation {} compared to the Secret CA key generation {}",
+                                    pod.getMetadata().getName(), podClusterCaKeyGeneration, clusterCaKeyGeneration);
+
+                            // only if all Kafka related components pods are updated to the new cluster CA cert generation,
+                            // there is the possibility that we should remove the older cluster CA from the Secret and stores
+                            if (clusterCaCertGeneration != podClusterCaCertGeneration) {
+                                this.isClusterCaFullyUsed = false;
+                            }
+
+                            if (clusterCaKeyGeneration != podClusterCaKeyGeneration) {
+                                this.isClusterCaNeedFullTrust = true;
+                            }
+
+                        } else {
+                            LOGGER.debugCr(reconciliation, "Skipping CA key generation check on pod {}, it's terminating", pod.getMetadata().getName());
+                        }
+
+                        if (isClusterCaNeedFullTrust) {
+                            LOGGER.debugCr(reconciliation, "The new Cluster CA is not yet trusted by all pods");
+                        }
+                        if (!isClusterCaFullyUsed) {
+                            LOGGER.debugCr(reconciliation, "The old Cluster CA is still used by some server certificates and cannot be removed");
+                        }
+                    }
+                    return Future.succeededFuture();
+                });
+    }
+
+    /**
      * If we need to roll the ZooKeeper cluster to roll out the trust to a new CA certificate when a CA private key is
      * being replaced, we need to know what the current number of ZooKeeper nodes is. Getting it from the Kafka custom
      * resource might not be good enough if a scale-up /scale-down is happening at the same time. So we get the
@@ -382,7 +463,7 @@ public class CaReconciler {
      *
      * @return  Current number of ZooKeeper replicas
      */
-    private Future<Integer> getZooKeeperReplicas() {
+    /* test */ Future<Integer> getZooKeeperReplicas() {
         return strimziPodSetOperator.getAsync(reconciliation.namespace(), KafkaResources.zookeeperStatefulSetName(reconciliation.name()))
                 .compose(podSet -> {
                     if (podSet != null
@@ -405,7 +486,7 @@ public class CaReconciler {
      * @return  Future which completes when this step is done either by rolling the ZooKeeper cluster or by deciding
      *          that no rolling is needed.
      */
-    private Future<Void> maybeRollZookeeper(int replicas, RestartReasons podRestartReasons) {
+    /* test */ Future<Void> maybeRollZookeeper(int replicas, RestartReasons podRestartReasons) {
         if (podRestartReasons.contains(RestartReason.CLUSTER_CA_CERT_KEY_REPLACED)) {
             Labels zkSelectorLabels = Labels.EMPTY
                     .withStrimziKind(reconciliation.kind())
@@ -424,7 +505,7 @@ public class CaReconciler {
         }
     }
 
-    private Future<Set<NodeRef>> getKafkaReplicas() {
+    /* test */ Future<Set<NodeRef>> getKafkaReplicas() {
         Labels selectorLabels = Labels.EMPTY
                 .withStrimziKind(reconciliation.kind())
                 .withStrimziCluster(reconciliation.name())
@@ -444,7 +525,7 @@ public class CaReconciler {
                 });
     }
 
-    private Future<Void> rollKafkaBrokers(Set<NodeRef> nodes, RestartReasons podRollReasons) {
+    /* test */ Future<Void> rollKafkaBrokers(Set<NodeRef> nodes, RestartReasons podRollReasons) {
         return new KafkaRoller(
                 reconciliation,
                 vertx,
@@ -484,7 +565,7 @@ public class CaReconciler {
      *
      * @return  Succeeded future if it succeeded, failed otherwise.
      */
-    private Future<Void> rollDeploymentIfExists(String deploymentName, String reason)  {
+    Future<Void> rollDeploymentIfExists(String deploymentName, String reason)  {
         return deploymentOperator.getAsync(reconciliation.namespace(), deploymentName)
                 .compose(dep -> {
                     if (dep != null) {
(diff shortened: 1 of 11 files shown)


cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/CaReconciler.java
```diff
@@ -373,6 +382,47 @@ Future<Void> rollingUpdateForNewCaKey() {
...
+     * Verify that all the pods are already trusting the new CA certificate signed by a new CA key.
+     * It checks each pod's CA key generation, compared with the new CA key generation.
+     * When the trusting phase is not completed (i.e. because CO stopped), it needs to be recovered from where it was left.
+     */
+    Future<Void> isClusterCaFullyTrusted() {
+        // Building the selector for Kafka related components
+        Labels labels =  Labels.forStrimziCluster(reconciliation.name()).withStrimziKind(Kafka.RESOURCE_KIND);
+
+        isClusterCaFullyTrusted = true;
+        return podOperator.listAsync(reconciliation.namespace(), labels)
```
scholzj, line 395: Just a thought ... the list is a relatively expensive operation ... and we will now do it twice for every reconciliation just in the `CaReconciler`. Would it make sense to list the pods once at the beginning to establish both:
* Whether the new CA is fully trusted
* Whether the the old CA can be removed

And then in two separate steps handle the flags?

operator-common/src/main/java/io/strimzi/operator/cluster/model/Ca.java
```diff
@@ -174,6 +174,11 @@ public String getSuffix() {
      */
     public static final String ANNO_STRIMZI_IO_CLIENTS_CA_CERT_GENERATION = Annotations.STRIMZI_DOMAIN + "clients-ca-cert-generation";
 
+    /**
+     * Annotation for tracking the Cluster CA key generation trusted by Kubernetes resources
+     */
+    public static final String ANNO_STRIMZI_IO_CLUSTER_CA_KEY_TRUSTED_GENERATION = Annotations.STRIMZI_DOMAIN + "cluster-ca-key-trusted-generation";
```
scholzj, line 180: As we discussed on the call => since we add the annotation only when the CA KEY generation is actually used an not just trusted, should we instead get back to just use the `CLUSTER_CA_KEY_GENERATION` annotation?

scholzj (commented):
Some nits. But I think it looks mostly good. I wonder if there is some way to have some more tests for this?

operator-common/src/main/java/io/strimzi/operator/cluster/model/Ca.java
ppatierno: Yeah I meant to do the change and forgot :-(

scholzj: 😆

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/CaReconciler.java
ppatierno: That's a good point. My only thought is the following ...
Between the trusting check and the CA cert removal there is the rolling for new CA key. While we don't expect changes in the `Pod` definition, they will be "technically" different pods, even with different names (actually ZK and K pods will have same names but EO, CC, KE will not). We should also discard the pods in "terminating" state when listing because they won't be there anymore during the CA cert removal.

ppatierno:
>  I wonder if there is some way to have some more tests for this?

I thought more about this but didn't find a good way for an integration test at least. I think this is something that could be tested more with system tests, where you can kill the CO pod and check everything will continue "smoothly".

ppatierno:
> I don't think you can test this in STs because you cannto know the right moment to kill the operator.

Well you can force the key replacement, waiting for the first ZK pod to be rolled and then kill the CO, that should be enough.

> But in unit tests, you can mock exactly the right state into which the CaReconciler starts.

Not sure I have got this :-(

scholzj:
> > But in unit tests, you can mock exactly the right state into which the CaReconciler starts.
> 
> Not sure I have got this :-(

You can mock the state of the cluster exactly as you need it in a unit test. you can prepare the Pods with different annotations, you can prepare terminating pods, etc. And then you can run the CaReconciler on it to see that it reacts as you expect.

ppatierno:
@scholzj I have added a couple of tests. One to check the new method which verifies that the CA key is trusted or not. The other one checks that the right rolling reason is specified (to roll nodes) when the CA key is not trusted.

cluster-operator/src/test/java/io/strimzi/operator/cluster/TestUtils.java
```diff
@@ -66,21 +69,34 @@ public static Map<String, String> containerEnvVars(Container container) {
...
+    }
+
+    public static Pod podWithNameAndAnnotations(String name, Map<String, String> annotations) {
+        return new PodBuilder()
+                .withNewMetadata()
+                    .withName(name)
+                    .withAnnotations(annotations)
+                .endMetadata()
+                .build();
+    }
```
scholzj, line 101: There are many different places where similar logic is used. So it makes sense to have it shared. But you should go through all the tests doing this and update them to use this method as well.

cluster-operator/src/test/java/io/strimzi/operator/cluster/operator/assembly/CaReconcilerTest.java
```diff
@@ -1451,4 +1458,195 @@ public void testClientsCASecretsWithoutOwnerReference(Vertx vertx, VertxTestCont
...
+                    .endKafka()
+                    .withNewZookeeper()
+                        .withReplicas(3)
+                        .withNewEphemeralStorage()
+                        .endEphemeralStorage()
+                    .endZookeeper()
+                .endSpec()
+                .build();
+
+
```
scholzj, line 1484: Double empty line probably not needed.

cluster-operator/src/test/java/io/strimzi/operator/cluster/operator/assembly/CaReconcilerTest.java
```diff
@@ -1451,4 +1458,195 @@ public void testClientsCASecretsWithoutOwnerReference(Vertx vertx, VertxTestCont
...
+                supplier, vertx, certManager, passwordGenerator);
+        mockCaReconciler
+                .reconcile(Clock.systemUTC())
+                .onComplete(context.succeeding(c -> context.verify(() -> {
+                    assertThat(mockCaReconciler.isClusterCaFullyTrusted, is(false));
+                    assertThat(mockCaReconciler.zkPodRestartReasons.contains(RestartReason.CLUSTER_CA_CERT_KEY_REPLACED), is(true));
+                    assertThat(mockCaReconciler.kPodRollReasons.contains(RestartReason.CLUSTER_CA_CERT_KEY_REPLACED), is(true));
+                    assertThat(mockCaReconciler.deploymentRollReason.size() == 3, is(true));
+                    for (String reason: mockCaReconciler.deploymentRollReason) {
+                        assertThat(reason.equals(RestartReason.CLUSTER_CA_CERT_KEY_REPLACED.getDefaultNote()), is(true));
```
scholzj, line 1596: Why where the deployments rolled if accroding to the `Kafka` resource none should exist?

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/CaReconciler.java
scholzj: So, what about this comment?

cluster-operator/src/test/java/io/strimzi/operator/cluster/TestUtils.java
ppatierno: I already moved this out from the manual rolling update test class. These methods were declared just there. I moved in this common TestUtils to be reused by my tests. If we want to use the same somewhere else, I don't think it's part of this PR.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/CaReconciler.java
ppatierno: Actually I was waiting for your feedback on my concern. The list of pods at the beginning could be different from the one across the steps (at least the first 2 which are done in the same reconciliation).

scholzj: But is it really possible that that changes the result?

cluster-operator/src/test/java/io/strimzi/operator/cluster/TestUtils.java
scholzj: Anyway, I guess I can live with this if needed.

cluster-operator/src/test/java/io/strimzi/operator/cluster/operator/assembly/CaReconcilerTest.java
ppatierno: I added the deployment(s) get and changed the mocked CaReconcile logic as well.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/CaReconciler.java
ppatierno: maybe not. the two steps you mentioned above are checking a couple annotation that won't change across the reconcile. What bothers me is just that they will really be different pods between step 1 and 2 :-)
but if you like I can make it listing one time and storing in the CaReconciler as a member field.

ppatierno: I have pushed changes with listing the pods just one time at the beginning of the CA reconciliation.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/CaReconciler.java
```diff
@@ -327,6 +333,10 @@ private void checkCustomCaSecret(CertificateAuthority ca, Secret certSecret, Sec
 
     Future<Void> clusterOperatorSecret(Clock clock) {
         oldCoSecret = clusterCa.clusterOperatorSecret();
+        if (!this.isClusterCaFullyTrusted) {
+            LOGGER.warnCr(reconciliation, "Cluster CA not fully trusted, keeping current CO secret and certs");
+            return Future.succeededFuture();
+        }
```
scholzj, line 339: Should there be some check for the secret's existence? I.e. what happens if the secret is deleted during a renewal (if the CA is already expired, that is a common approach to delete all secrets and have the operator recreate them).

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/CaReconciler.java
```diff
@@ -373,6 +385,64 @@ Future<Void> rollingUpdateForNewCaKey() {
...
+                    this.pods.clear();
+                    if (pods.isEmpty()) {
+                        return Future.succeededFuture();
+                    }
+                    for (Pod pod : pods) {
+                        // with "RollingUpdate" strategy on Deployment(s) (i.e. the Cruise Control one),
+                        // while the Deployment is reported to be ready, the old pod is still alive but terminating
+                        // this condition is for skipping "Terminating" pods for checks on the CA key and old certificates
+                        if (pod.getMetadata().getDeletionTimestamp() == null) {
+                            this.pods.add(pod);
```
scholzj, line 409: Wouldn't it make sense to instead of storing all the pods in the object to run the two checks on them and store the booleans instead? It seems like that would safe a lot of memory.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/CaReconciler.java
```diff
@@ -381,7 +451,7 @@ Future<Void> rollingUpdateForNewCaKey() {
      *
      * @return  Current number of ZooKeeper replicas
      */
-    private Future<Integer> getZooKeeperReplicas() {
+    Future<Integer> getZooKeeperReplicas() {
```
scholzj, line 454: Is it used in tests? Should it be in that case
```suggestion
    /* test */ Future<Integer> getZooKeeperReplicas() {
```
as we normally do for other methods?

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/CaReconciler.java
```diff
@@ -404,7 +474,7 @@ private Future<Integer> getZooKeeperReplicas() {
      * @return  Future which completes when this step is done either by rolling the ZooKeeper cluster or by deciding
      *          that no rolling is needed.
      */
-    private Future<Void> maybeRollZookeeper(int replicas, RestartReasons podRestartReasons) {
+    Future<Void> maybeRollZookeeper(int replicas, RestartReasons podRestartReasons) {
```
scholzj, line 477: Same as above and in few more methods below.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/CaReconciler.java
ppatierno: Yeah, I added it.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/CaReconciler.java
ppatierno: Yep done.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/CaReconciler.java
```diff
@@ -373,6 +385,71 @@ Future<Void> rollingUpdateForNewCaKey() {
...
+                    LOGGER.debugCr(reconciliation, "Current cluster CA key generation {}", clusterCaKeyGeneration);
+
+
+                    for (Pod pod : pods) {
+                        // with "RollingUpdate" strategy on Deployment(s) (i.e. the Cruise Control one),
+                        // while the Deployment is reported to be ready, the old pod is still alive but terminating
+                        // this condition is for skipping "Terminating" pods for checks on the CA key and old certificates
+                        if (pod.getMetadata().getDeletionTimestamp() == null) {
+                            int podClusterCaCertGeneration = Annotations.intAnnotation(pod, Ca.ANNO_STRIMZI_IO_CLUSTER_CA_CERT_GENERATION, clusterCaCertGeneration);
+                            LOGGER.debugCr(reconciliation, "Pod {} cluster CA cert generation {}", pod.getMetadata().getName(), podClusterCaCertGeneration);
```
scholzj, line 428: A nit ... but should this have some message or at least some verb?

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/CaReconciler.java
```diff
@@ -373,6 +385,71 @@ Future<Void> rollingUpdateForNewCaKey() {
...
+                                    pod.getMetadata().getName(), podClusterCaKeyGeneration, clusterCaKeyGeneration);
+
+                            // only if all Kafka related components pods are updated to the new cluster CA cert generation,
+                            // there is the possibility that we should remove the older cluster CA from the Secret and stores
+                            if (clusterCaCertGeneration != podClusterCaCertGeneration) {
+                                this.isClusterCaFullyUsed = false;
+                            }
+
+                            if (clusterCaKeyGeneration != podClusterCaKeyGeneration) {
+                                LOGGER.debugCr(reconciliation, "Need to continue trusting new CA key ....");
```
scholzj, line 441: This will be in some situations logged for every pod. Is it really needed? If we want to log something, would it make more sense to log it:
* Once at the end?
* For both flags?

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/CaReconciler.java
```diff
@@ -373,6 +385,71 @@ Future<Void> rollingUpdateForNewCaKey() {
...
+                                this.isClusterCaNeedFullTrust = true;
+                            }
+
+                        } else {
+                            LOGGER.infoCr(reconciliation, "Skipping CA key generation check on pod {}, it's terminating", pod.getMetadata().getName());
```
scholzj, line 446: Does this make sense at INFO level? Seems more like DEBUG to me.

cluster-operator/src/test/java/io/strimzi/operator/cluster/operator/assembly/CaReconcilerTest.java
```diff
@@ -1451,4 +1459,233 @@ public void testClientsCASecretsWithoutOwnerReference(Vertx vertx, VertxTestCont
...
+
+        @Override
+        Future<Void> rollDeploymentIfExists(String deploymentName, String reason) {
+            return deploymentOperator.getAsync(reconciliation.namespace(), deploymentName)
+                    .compose(dep -> {
+                        if (dep != null) {
+                            this.deploymentRollReason.add(reason);
+                        }
+                        return Future.succeededFuture();
+                    });
```
scholzj, line 1683: Wouldn't doing this through Mockito as the other operations be more elegant than doing it here in the override?

ppatierno: The `deploymentOperator.getAsync` is already mocked in the test to return the Deployment(s).
What I am adding here is the rolling reason which is a field in the MockCaReconciler (and via the override I am avoiding the base class behaviour which would rolling the deployment instead).

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/CaReconciler.java
ppatierno: Agree.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/CaReconciler.java
```diff
@@ -373,6 +383,77 @@ Future<Void> rollingUpdateForNewCaKey() {
...
+
+                        } else {
+                            LOGGER.debugCr(reconciliation, "Skipping CA key generation check on pod {}, it's terminating", pod.getMetadata().getName());
+                        }
+
+                        if (isClusterCaNeedFullTrust) {
+                            LOGGER.debugCr(reconciliation, "Need to continue trusting new cluster CA key ....");
+                        }
+                        if (!isClusterCaFullyUsed) {
+                            LOGGER.debugCr(reconciliation, "The cluster CA certificate is not used across all nodes to sign the certificates");
```
scholzj, line 450:
```suggestion
                            LOGGER.debugCr(reconciliation, "The old Cluster CA is still used by some server certificates and cannot be removed");
```

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/CaReconciler.java
```diff
@@ -373,6 +383,77 @@ Future<Void> rollingUpdateForNewCaKey() {
...
+                            if (clusterCaKeyGeneration != podClusterCaKeyGeneration) {
+                                this.isClusterCaNeedFullTrust = true;
+                            }
+
+                        } else {
+                            LOGGER.debugCr(reconciliation, "Skipping CA key generation check on pod {}, it's terminating", pod.getMetadata().getName());
+                        }
+
+                        if (isClusterCaNeedFullTrust) {
+                            LOGGER.debugCr(reconciliation, "Need to continue trusting new cluster CA key ....");
```
scholzj, line 447:
```suggestion
                            LOGGER.debugCr(reconciliation, "The new Cluster CA is not yet trusted by all pods");
```

scholzj (approved):
Some nits to the logs to maybe make them more understandable. But LGTM otherwise - no need to re-review.

PR #51, opened 2017-10-15
Add Kafka healthcheck using kafka-broker-api-versions.sh - Fixes #3

scholzj:

PR #8627, opened 2023-06-06
Unidirectional topic operator

tombentley:
### Type of change

- Enhancement / new feature

### Description

This is an initial implementation of the Unidirectional Topic Operator described in [proposal #51](https://github.com/strimzi/proposals/blob/main/051-unidirectional-topic-operator.md). Some details can be found in the `DESIGN.md`

### Checklist

- [x] Write tests
- [ ] Update CHANGELOG.md

### DIFF
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/FeatureGates.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/FeatureGates.java
@@ -20,10 +20,14 @@ public class FeatureGates {
     private static final String STABLE_CONNECT_IDENTITIES = "StableConnectIdentities";
     private static final String KAFKA_NODE_POOLS = "KafkaNodePools";
 
+    private static final String UNIDIRECTIONAL_TOPIC_OPERATOR = "UnidirectionalTopicOperator";
+
     // When adding new feature gates, do not forget to add them to allFeatureGates() and toString() methods
     private final FeatureGate useKRaft = new FeatureGate(USE_KRAFT, false);
     private final FeatureGate stableConnectIdentities = new FeatureGate(STABLE_CONNECT_IDENTITIES, false);
     private final FeatureGate kafkaNodePools = new FeatureGate(KAFKA_NODE_POOLS, false);
+
+    private final FeatureGate unidirectionalTopicOperator = new FeatureGate(UNIDIRECTIONAL_TOPIC_OPERATOR, false);
 
     /**
      * Constructs the feature gates configuration.
@@ -53,6 +57,9 @@ public class FeatureGates {
                         break;
                     case KAFKA_NODE_POOLS:
                         setValueOnlyOnce(kafkaNodePools, value);
+                        break;
+                    case UNIDIRECTIONAL_TOPIC_OPERATOR:
+                        setValueOnlyOnce(unidirectionalTopicOperator, value);
                         break;
                     default:
                         throw new InvalidConfigurationException("Unknown feature gate " + featureGate + " found in the configuration");
@@ -107,6 +114,14 @@ public class FeatureGates {
     public boolean kafkaNodePoolsEnabled() {
         return kafkaNodePools.isEnabled();
     }
+
+    /**
+     * @return  Returns true when the UnidirectionalTopicOperator feature gate is enabled
+     */
+    public boolean unidirectionalTopicOperatorEnabled() {
+        return unidirectionalTopicOperator.isEnabled();
+    }
+
     /**
      * Returns a list of all Feature gates. Used for testing.
      *
@@ -116,7 +131,8 @@ public class FeatureGates {
         return List.of(
                 useKRaft,
                 stableConnectIdentities,
-                kafkaNodePools
+                kafkaNodePools,
+                unidirectionalTopicOperator
         );
     }
 
@@ -125,7 +141,8 @@ public class FeatureGates {
         return "FeatureGates(" +
                 "UseKRaft=" + useKRaft.isEnabled() + "," +
                 "StableConnectIdentities=" + stableConnectIdentities.isEnabled() + "," +
-                "KafkaNodePools=" + kafkaNodePools.isEnabled() +
+                "KafkaNodePools=" + kafkaNodePools.isEnabled() + "," +
+                "UnidirectionalTopicOperator=" + unidirectionalTopicOperator.isEnabled() +
                 ")";
     }
 
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/model/EntityOperator.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/model/EntityOperator.java
@@ -124,13 +124,31 @@ public class EntityOperator extends AbstractModel {
      * @return Entity Operator instance, null if not configured in the ConfigMap
      */
     public static EntityOperator fromCrd(Reconciliation reconciliation, Kafka kafkaAssembly, KafkaVersion.Lookup versions, SharedEnvironmentProvider sharedEnvironmentProvider) {
+        return fromCrd(reconciliation, kafkaAssembly, versions, sharedEnvironmentProvider, false);
+    }
+    /**
+     * Create an Entity Operator from given desired resource
+     *
+     * @param reconciliation The reconciliation
+     * @param kafkaAssembly desired resource with cluster configuration containing the Entity Operator one
+     * @param versions The versions.
+     * @param sharedEnvironmentProvider     Shared environment provider.
+     * @param unidirectionalTopicOperator Indicates whether the UTO should be used.
+     *
+     * @return Entity Operator instance, null if not configured in the ConfigMap
+     */
+    public static EntityOperator fromCrd(Reconciliation reconciliation,
+                                         Kafka kafkaAssembly,
+                                         KafkaVersion.Lookup versions,
+                                         SharedEnvironmentProvider sharedEnvironmentProvider,
+                                         boolean unidirectionalTopicOperator) {
         EntityOperatorSpec entityOperatorSpec = kafkaAssembly.getSpec().getEntityOperator();
 
         if (entityOperatorSpec != null
                 && (entityOperatorSpec.getUserOperator() != null || entityOperatorSpec.getTopicOperator() != null)) {
             EntityOperator result = new EntityOperator(reconciliation, kafkaAssembly, sharedEnvironmentProvider);
 
-            EntityTopicOperator topicOperator = EntityTopicOperator.fromCrd(reconciliation, kafkaAssembly, sharedEnvironmentProvider);
+            EntityTopicOperator topicOperator = EntityTopicOperator.fromCrd(reconciliation, kafkaAssembly, sharedEnvironmentProvider, unidirectionalTopicOperator);
             EntityUserOperator userOperator = EntityUserOperator.fromCrd(reconciliation, kafkaAssembly, sharedEnvironmentProvider);
 
             result.tlsSidecar = entityOperatorSpec.getTlsSidecar();
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/model/EntityOperator.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/model/EntityOperator.java
@@ -85,6 +85,7 @@ public class EntityOperator extends AbstractModel {
     /* test */ String zookeeperConnect;
     private EntityTopicOperator topicOperator;
     private EntityUserOperator userOperator;
+    /* test */  boolean unidirectionalTopicOperator;
     private TlsSidecar tlsSidecar;
     private String tlsSidecarImage;
 
@@ -154,6 +155,7 @@ public class EntityOperator extends AbstractModel {
 
             result.tlsSidecar = entityOperatorSpec.getTlsSidecar();
             result.topicOperator = topicOperator;
+            result.unidirectionalTopicOperator = unidirectionalTopicOperator;
             result.userOperator = userOperator;
 
             String tlsSideCarImage = entityOperatorSpec.getTlsSidecar() != null ? entityOperatorSpec.getTlsSidecar().getImage() : null;
@@ -247,7 +249,7 @@ public class EntityOperator extends AbstractModel {
         }
 
         // The TLS Sidecar is only used by the Topic Operator. Therefore, when the Topic Operator is disabled, the TLS side should also be disabled.
-        if (topicOperator != null) {
+        if (topicOperator != null && !this.unidirectionalTopicOperator) {
             String tlsSidecarImage = this.tlsSidecarImage;
             if (tlsSidecar != null && tlsSidecar.getImage() != null) {
                 tlsSidecarImage = tlsSidecar.getImage();
@@ -305,8 +307,10 @@ public class EntityOperator extends AbstractModel {
             volumeList.add(VolumeUtils.createSecretVolume(EUO_CERTS_VOLUME_NAME, KafkaResources.entityUserOperatorSecretName(cluster), isOpenShift));
         }
 
-        volumeList.add(VolumeUtils.createTempDirVolume(TLS_SIDECAR_TMP_DIRECTORY_DEFAULT_VOLUME_NAME, templatePod));
-        volumeList.add(VolumeUtils.createSecretVolume(TLS_SIDECAR_CA_CERTS_VOLUME_NAME, AbstractModel.clusterCaCertSecretName(cluster), isOpenShift));
+        if (!unidirectionalTopicOperator) {
+            volumeList.add(VolumeUtils.createTempDirVolume(TLS_SIDECAR_TMP_DIRECTORY_DEFAULT_VOLUME_NAME, templatePod));
+            volumeList.add(VolumeUtils.createSecretVolume(TLS_SIDECAR_CA_CERTS_VOLUME_NAME, AbstractModel.clusterCaCertSecretName(cluster), isOpenShift));
+        }
         return volumeList;
     }
 
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/model/EntityOperator.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/model/EntityOperator.java
@@ -126,6 +126,7 @@ public class EntityOperator extends AbstractModel {
     public static EntityOperator fromCrd(Reconciliation reconciliation, Kafka kafkaAssembly, KafkaVersion.Lookup versions, SharedEnvironmentProvider sharedEnvironmentProvider) {
         return fromCrd(reconciliation, kafkaAssembly, versions, sharedEnvironmentProvider, false);
     }
+    
     /**
      * Create an Entity Operator from given desired resource
      *
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/model/EntityOperator.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/model/EntityOperator.java
@@ -248,7 +248,8 @@ public class EntityOperator extends AbstractModel {
             containers.add(userOperator.createContainer(imagePullPolicy));
         }
 
-        // The TLS Sidecar is only used by the Topic Operator. Therefore, when the Topic Operator is disabled, the TLS side should also be disabled.
+        // The TLS Sidecar is only used by the Bidirectional Topic Operator.
+        // Therefore, when the Topic Operator is disabled, or we're using the Unidirectional TO, the TLS side should also be disabled.
         if (topicOperator != null && !this.unidirectionalTopicOperator) {
             String tlsSidecarImage = this.tlsSidecarImage;
             if (tlsSidecar != null && tlsSidecar.getImage() != null) {
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/model/EntityOperator.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/model/EntityOperator.java
@@ -32,8 +32,8 @@ import io.strimzi.operator.cluster.ClusterOperatorConfig;
 import io.strimzi.operator.cluster.Main;
 import io.strimzi.operator.cluster.model.securityprofiles.ContainerSecurityProviderContextImpl;
 import io.strimzi.operator.cluster.model.securityprofiles.PodSecurityProviderContextImpl;
-import io.strimzi.operator.common.Reconciliation;
 import io.strimzi.operator.cluster.operator.resource.SharedEnvironmentProvider;
+import io.strimzi.operator.common.Reconciliation;
 import io.strimzi.operator.common.Util;
 
 import java.io.BufferedReader;
@@ -309,8 +309,8 @@ public class EntityOperator extends AbstractModel {
 
         if (!unidirectionalTopicOperator) {
             volumeList.add(VolumeUtils.createTempDirVolume(TLS_SIDECAR_TMP_DIRECTORY_DEFAULT_VOLUME_NAME, templatePod));
-            volumeList.add(VolumeUtils.createSecretVolume(TLS_SIDECAR_CA_CERTS_VOLUME_NAME, AbstractModel.clusterCaCertSecretName(cluster), isOpenShift));
         }
+        volumeList.add(VolumeUtils.createSecretVolume(TLS_SIDECAR_CA_CERTS_VOLUME_NAME, AbstractModel.clusterCaCertSecretName(cluster), isOpenShift));
         return volumeList;
     }
 
--- a/cluster-operator/src/test/java/io/strimzi/operator/cluster/model/EntityOperatorTest.java
+++ b/cluster-operator/src/test/java/io/strimzi/operator/cluster/model/EntityOperatorTest.java
@@ -126,8 +126,17 @@ public class EntityOperatorTest {
     private final EntityOperator entityOperator = EntityOperator.fromCrd(new Reconciliation("test", resource.getKind(), resource.getMetadata().getNamespace(), resource.getMetadata().getName()), resource, VERSIONS, SHARED_ENV_PROVIDER);
 
     @ParallelTest
-    public void testGenerateDeployment() {
+    public void testGenerateDeploymentWithBto() {
+        testGenerateDeployment(false);
+    }
 
+    @ParallelTest
+    public void testGenerateDeploymentWithUto() {
+        testGenerateDeployment(true);
+    }
+
+    private void testGenerateDeployment(boolean useUto) {
+        entityOperator.unidirectionalTopicOperator = useUto;
         Deployment dep = entityOperator.generateDeployment(true, null, null);
 
         List<Container> containers = dep.getSpec().getTemplate().getSpec().getContainers();
@@ -137,28 +146,36 @@ public class EntityOperatorTest {
         assertThat(dep.getSpec().getReplicas(), is(1));
         TestUtils.checkOwnerReference(dep, resource);
 
-        assertThat(containers.size(), is(3));
+        assertThat(containers.size(), is(useUto ? 2 : 3));
         // just check names of topic and user operators (their containers are tested in the related unit test classes)
         assertThat(containers.get(0).getName(), is(EntityTopicOperator.TOPIC_OPERATOR_CONTAINER_NAME));
         assertThat(containers.get(1).getName(), is(EntityUserOperator.USER_OPERATOR_CONTAINER_NAME));
         // checks on the TLS sidecar container
-        Container tlsSidecarContainer = containers.get(2);
-        assertThat(tlsSidecarContainer.getImage(), is(image));
-        assertThat(io.strimzi.operator.cluster.TestUtils.containerEnvVars(tlsSidecarContainer).get(EntityOperator.ENV_VAR_ZOOKEEPER_CONNECT), is(KafkaResources.zookeeperServiceName(cluster) + ":" + ZookeeperCluster.CLIENT_TLS_PORT));
-        assertThat(io.strimzi.operator.cluster.TestUtils.containerEnvVars(tlsSidecarContainer).get(ModelUtils.TLS_SIDECAR_LOG_LEVEL), is(TlsSidecarLogLevel.NOTICE.toValue()));
-        assertThat(EntityOperatorTest.volumeMounts(tlsSidecarContainer.getVolumeMounts()), is(map(
-                        EntityOperator.TLS_SIDECAR_TMP_DIRECTORY_DEFAULT_VOLUME_NAME, VolumeUtils.STRIMZI_TMP_DIRECTORY_DEFAULT_MOUNT_PATH,
-                        EntityOperator.TLS_SIDECAR_CA_CERTS_VOLUME_NAME, EntityOperator.TLS_SIDECAR_CA_CERTS_VOLUME_MOUNT,
-                        EntityOperator.ETO_CERTS_VOLUME_NAME, EntityOperator.ETO_CERTS_VOLUME_MOUNT)));
-        assertThat(tlsSidecarContainer.getReadinessProbe().getInitialDelaySeconds(), is(tlsHealthDelay));
-        assertThat(tlsSidecarContainer.getReadinessProbe().getTimeoutSeconds(), is(tlsHealthTimeout));
-        assertThat(tlsSidecarContainer.getLivenessProbe().getInitialDelaySeconds(), is(tlsHealthDelay));
-        assertThat(tlsSidecarContainer.getLivenessProbe().getTimeoutSeconds(), is(tlsHealthTimeout));
+        if (!useUto) {
+            Container tlsSidecarContainer = containers.get(2);
+            assertThat(tlsSidecarContainer.getImage(), is(image));
+            assertThat(io.strimzi.operator.cluster.TestUtils.containerEnvVars(tlsSidecarContainer).get(EntityOperator.ENV_VAR_ZOOKEEPER_CONNECT), is(KafkaResources.zookeeperServiceName(cluster) + ":" + ZookeeperCluster.CLIENT_TLS_PORT));
+            assertThat(io.strimzi.operator.cluster.TestUtils.containerEnvVars(tlsSidecarContainer).get(ModelUtils.TLS_SIDECAR_LOG_LEVEL), is(TlsSidecarLogLevel.NOTICE.toValue()));
+            assertThat(EntityOperatorTest.volumeMounts(tlsSidecarContainer.getVolumeMounts()), is(map(
+                    EntityOperator.TLS_SIDECAR_TMP_DIRECTORY_DEFAULT_VOLUME_NAME, VolumeUtils.STRIMZI_TMP_DIRECTORY_DEFAULT_MOUNT_PATH,
+                    EntityOperator.TLS_SIDECAR_CA_CERTS_VOLUME_NAME, EntityOperator.TLS_SIDECAR_CA_CERTS_VOLUME_MOUNT,
+                    EntityOperator.ETO_CERTS_VOLUME_NAME, EntityOperator.ETO_CERTS_VOLUME_MOUNT)));
+            assertThat(tlsSidecarContainer.getReadinessProbe().getInitialDelaySeconds(), is(tlsHealthDelay));
+            assertThat(tlsSidecarContainer.getReadinessProbe().getTimeoutSeconds(), is(tlsHealthTimeout));
+            assertThat(tlsSidecarContainer.getLivenessProbe().getInitialDelaySeconds(), is(tlsHealthDelay));
+            assertThat(tlsSidecarContainer.getLivenessProbe().getTimeoutSeconds(), is(tlsHealthTimeout));
+        }
 
         List<Volume> volumes = dep.getSpec().getTemplate().getSpec().getVolumes();
         assertThat(volumes.stream().filter(volume -> volume.getName().equals(EntityUserOperator.USER_OPERATOR_TMP_DIRECTORY_DEFAULT_VOLUME_NAME)).findFirst().orElseThrow().getEmptyDir().getSizeLimit(), is(new Quantity("100", "Mi")));
         assertThat(volumes.stream().filter(volume -> volume.getName().equals(EntityTopicOperator.TOPIC_OPERATOR_TMP_DIRECTORY_DEFAULT_VOLUME_NAME)).findFirst().orElseThrow().getEmptyDir().getSizeLimit(), is(new Quantity("100", "Mi")));
-        assertThat(volumes.stream().filter(volume -> volume.getName().equals(EntityOperator.TLS_SIDECAR_TMP_DIRECTORY_DEFAULT_VOLUME_NAME)).findFirst().orElseThrow().getEmptyDir().getSizeLimit(), is(new Quantity("100", "Mi")));
+        if (!useUto) {
+            assertThat(volumes.stream().filter(volume -> volume.getName().equals(EntityOperator.TLS_SIDECAR_TMP_DIRECTORY_DEFAULT_VOLUME_NAME)).findFirst().orElseThrow().getEmptyDir().getSizeLimit(), is(new Quantity("100", "Mi")));
+            assertThat(volumes.stream().filter(volume -> volume.getName().equals(EntityOperator.TLS_SIDECAR_CA_CERTS_VOLUME_NAME)).findFirst().isEmpty(), is(false));
+        } else {
+            assertThat(volumes.stream().filter(volume -> volume.getName().equals(EntityOperator.TLS_SIDECAR_TMP_DIRECTORY_DEFAULT_VOLUME_NAME)).findFirst().isEmpty(), is(true));
+            assertThat(volumes.stream().filter(volume -> volume.getName().equals(EntityOperator.TLS_SIDECAR_CA_CERTS_VOLUME_NAME)).findFirst().isEmpty(), is(true));
+        }
     }
 
     @ParallelTest
--- a/cluster-operator/src/test/java/io/strimzi/operator/cluster/model/EntityOperatorTest.java
+++ b/cluster-operator/src/test/java/io/strimzi/operator/cluster/model/EntityOperatorTest.java
@@ -135,8 +135,8 @@ public class EntityOperatorTest {
         testGenerateDeployment(true);
     }
 
-    private void testGenerateDeployment(boolean useUto) {
-        entityOperator.unidirectionalTopicOperator = useUto;
+    private void testGenerateDeployment(boolean useUnidirectionalTopicOperator) {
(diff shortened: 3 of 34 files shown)


tombentley:
@scholzj @ppatierno and anyone else who's watching this: It is now marked as ready for review.

topic-operator/src/main/java/io/strimzi/operator/topic/v2/Either.java
```diff
@@ -0,0 +1,94 @@
+/*
+ * Copyright Strimzi authors.
+ * License: Apache License 2.0 (see the file LICENSE or http://apache.org/licenses/LICENSE-2.0.html).
+ */
+package io.strimzi.operator.topic.v2;
+
+import java.util.Objects;
+import java.util.function.Function;
+
+class Either<X, Y> {
```
scholzj, line 10: A JavaDoc might be great to explain what it does / why it exists.

fvaleri (commented):
Thanks @tombentley. Great work!

I reviewed the code and tested the UTO on a remote OpenShift 4.13 cluster. I also tested the upgrade from BTO to UTO.

Metrics, user docs and changelog will be added in a follow up PR right?

I'll put here a bug report and some additional notes.

### Bug: number of partitions divergence

- Create a KafkaTopic named "foo" using kafka-topic.yaml from packaging.
- Create a KafkaTopic named "bar" with "topicName: foo" using kafka-topic.yaml from packaging.
- Wait some time and bar will be marked as NOT ready due to ResourceConflict, as expected.
- Edit foo and raise the number of partitions from 1 to 3.
- Wait some time and confirm this change in Kafka using kafka-topics.sh.
- Unmanage foo by using annotation `trimzi.io/managed=false`.
- Delete foo in Kube.
- Wait some more time for bar to become ready.

At this point, the number of partitions inside the KafkaTopic managing foo topic will be different from the actual number of partitions inside Kafka.

```sh
$ kubectl get kt bar -o yaml
kind: KafkaTopic
metadata:
  creationTimestamp: "2023-06-14T07:34:14Z"
  finalizers:
  - strimzi.io/topic-operator
  generation: 5
  labels:
  name: bar
  namespace: test
  resourceVersion: "14373"
  uid: d29072c5-857c-48a2-ba19-114c707d6fbd
spec:
  config:
    retention.ms: 7200000
    segment.bytes: 1073741824
  partitions: 1
  replicas: 1
  topicName: foo
status:
  conditions:
  - lastTransitionTime: "2023-06-14T07:55:57.912184014Z"
    status: "True"
    type: Ready
  observedGeneration: 5
  topicName: foo

$ kubectl exec my-cluster-kafka-0 -- bin/kafka-topics.sh --bootstrap-server my-cluster-kafka-bootstrap:9092 --describe --topic foo
Topic: foo	TopicId: pYgNiFhqRyiwrWSrOFHKoA	PartitionCount: 3	ReplicationFactor: 1	Configs: min.insync.replicas=2,segment.bytes=1073741824,retention.ms=7200000,message.format.version=3.0-IV1
	Topic: foo	Partition: 0	Leader: 0	Replicas: 0	Isr: 0
	Topic: foo	Partition: 1	Leader: 1	Replicas: 1	Isr: 1
	Topic: foo	Partition: 2	Leader: 2	Replicas: 2	Isr: 2
```

### Usability improvement

On the usability side, it would be good to have an additional MANAGED column when listing KafkaTopic resources, something like:

```sh
$ kubectl get kt
NAME   CLUSTER      PARTITIONS   REPLICATION FACTOR   MANAGED   READY
bar    my-cluster   1            1                    True      False
foo    my-cluster   3            1                    True      True
baz    my-cluster   1            1                    False     True
```

### Design document integration

I like the idea of putting important design and implementation considerations near the code, as in the new user operator.

As explained in the proposal, the use of finalizers can prevent other resources from being removed (i.e. the entire namespace). I would add this information to both `DESIGN.md` and user doc, becasue I think it will be a pretty common issue.

The UTO authenticates to Kafka as super user, so when it gets `UNKNOWN_TOPIC_OR_PARTITION` is actually what the UTO expect (the topic doesn't exist). Instead, if we switch to a more restrictive user configuration following the least priviledge principle, then this error could mean that the user has no `Describe` permission on that topic. Maybe we can add a note about that.

topic-operator/src/main/java/io/strimzi/operator/topic/v2/BatchingTopicController.java
```diff
@@ -0,0 +1,989 @@
...
+            } catch (ExecutionException e) {
+                return pair(reconcilableTopic, Either.ofRight(handleAdminException(e)));
+            } catch (InterruptedException e) {
+                throw new UncheckedInterruptedException(e);
+            }
+        }));
+    }
+
+    private static TopicOperatorException handleAdminException(ExecutionException e) {
+        // Note this method is declared to return a RuntimeException purely so that the caller
```
LiamClarkeNZ, line 271: Comment states RuntimeException, but the TopicOperatorException extends Exception?

topic-operator/src/main/java/io/strimzi/operator/topic/v2/TopicOperatorMain.java
```diff
@@ -0,0 +1,171 @@
...
+    /**
+     * Entrypoint for unidirectional TO.
+     * @param args Command line args
+     * @throws Exception If bad things happen
+     */
+    public static void main(String[] args) throws Exception {
+        TopicOperatorConfig topicOperatorConfig = TopicOperatorConfig.buildFromMap(System.getenv());
+        TopicOperatorMain operator = operator(topicOperatorConfig, kubeClient(), Admin.create(topicOperatorConfig.adminClientConfig()));
+        operator.start();
+        LOGGER.infoOp("Returning from main()");
```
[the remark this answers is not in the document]
tombentley: Just to be clear: `operator.start()` returns, rather than blocks, so it's normal to see this log message when the operator starts, and seeing it tells you that startup is complete. The TO process doesn't terminate when the main thread stops because by that point the non-daemon queue processing threads are running. 

I think if we make this `DEBUG` we should probably make many/all of the other log messages in this class `DEBUG` too. But TBH they're all only printed once anyway, so I think info is fine, personally.

topic-operator/src/main/java/io/strimzi/operator/topic/v2/TopicOperatorConfig.java
```diff
@@ -0,0 +1,247 @@
...
+        }
+
+        kafkaClientProps.put(SaslConfigs.SASL_MECHANISM, saslMechanism);
+        kafkaClientProps.put(SaslConfigs.SASL_JAAS_CONFIG, jaasConfig);
+    }
+
+    @Override
+    public String toString() {
+        String mask = "********";
+        return "TopicOperatorConfig{" +
```
tombentley, line 224: @scholzj did we decide multi-line log messages were fine? Should I change this to reflect that?

scholzj: Nobody protested -> so I would say that yes, they are fine.

topic-operator/src/main/java/io/strimzi/operator/topic/v2/Either.java
tombentley: @fvaleri I didn't know that convention. I can refactor it, but I'll do that after fixing other review comments since it might confuse github if there are any comments at call sites.

@scholzj I'll add some Javadoc. I'll also rename the `X` and `Y` to `L` and `R` to better reflect their relationship to `left` and `right`.

FWIW, I considered implementing this as a `sealed interface` with two permitted `final` implementations, but decided the `boolean left` was about as good.

@see-quick did you have any other comments/suggestions on this, since you mentioned it in a private chat?

topic-operator/src/main/java/io/strimzi/operator/topic/v2/DESIGN.md
```diff
@@ -0,0 +1,38 @@
...
+It follows that reducing the cost of those `Admin` operations will allow the steady-state UTO to cope with larger numbers of `KafkaTopic`. 
+Kafka provides a useful mechanism for getting higher throughput for metadata operations: batching.
+
+For this reason the UTO tries to make use of the request batching supported by the Admin client.
+It does this by using a similar heuristic to that used in the Apache Kafka `Producer` client: Trading a bit of latency (in the form of a configurable "linger" duration) to in order to collect a batch of items to achieve higher throughput.
+For this reason the UTO tries to make use of the request batching supported by the `Admin` client.
+It does this by using a similar heuristic to that used in the Apache Kafka `Producer` client: Trading a bit of latency (in the form of a configurable "linger" duration) to in order to collect a batch of items to achieve higher throughput.
+Thus rather than reconciling individual KafkaTopics the UTO reconcilate batches of KafkaTopics. 
+* The batches are created on each iteration of the `BatchingLoop.LoopRunnable`.
+* Once a batch is created, the topics within it are reconciled together through to completion.
```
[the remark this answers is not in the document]
tombentley: "1 reconciliation" is a bit open to interpretation about what exactly a reconciliation is. So I worded it as `1 iteration of `LoopRunnable` => 1 batch => N topic events`. Let me know if that doesn't work for you and we can agree a better wording.

topic-operator/src/main/java/io/strimzi/operator/topic/v2/BatchingTopicController.java
```diff
@@ -0,0 +1,989 @@
...
+
+    private Either<Boolean, TopicOperatorException> validateSingleManagingResource(ReconcilableTopic reconcilableTopic) {
+        String tn = reconcilableTopic.topicName();
+        var existing = topics.computeIfAbsent(tn, k -> new HashSet<>());
+        Ref thisRef = new Ref(reconcilableTopic.kt());
+        existing.add(thisRef);
+        return Either.ofLeft(true);
+    }
+
+    private Either<Boolean, TopicOperatorException> validateSingleManagingResource2(ReconcilableTopic reconcilableTopic) {
```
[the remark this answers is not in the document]
tombentley: I renamed the other one.

topic-operator/src/main/java/io/strimzi/operator/topic/v2/DESIGN.md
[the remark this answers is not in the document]
fvaleri: Fine with me, thanks.

topic-operator/src/main/java/io/strimzi/operator/topic/v2/BatchingLoop.java
```diff
@@ -0,0 +1,302 @@
...
+                    LOGGER.warnOp("[Batch #{}] Item {} couldn't be pushed onto the deque head after rejection from batch", batchId, item);
+                }
+            }
+        }
+
+        private void addToBatch(int batchId, Batch batch, List<TopicEvent> rejected, TopicEvent topicEvent) {
+            // We could add logic here to cope properly with interleaved upserts and deletes
+            // of the same topic in the same batch.
+            // E.g. upset then delete is equivalent to just a delete
+            // It's actually a bit tricky since you have to process the events in reverse order to correctly
```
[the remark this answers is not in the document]
tombentley: There are pros and cons. I've thought about it far too hard. 

Pros:

* It's dead simple
* It means we don't really need fiddly tests to cover the cases

Cons: 

* It means we have to use a deque, which might not perform great
* It can increase the latency for processing many events happening to the same topic

Note that because the deque/queue contains events, not simply topics, we can't easily put a logically correct upper bound on how big it should get (unlike the UserOperator, where the queue holds _unique_ `KafkaUsers`, so you only need a queue that's as big as the number of users you expect). So the latency issue could actually manifest as `offer` failing, and incorrect behaviour (e.g. missed deletion events when finalizers are not in use resulting in topics remaining in Kafka). It would probably be best to exit the whole TO process in that case, forcing the user to increase the size of the deque. Wdyt?

topic-operator/pom.xml
```diff
@@ -22,6 +22,36 @@
...
+            <artifactId>testing-junit5-extension</artifactId>
+            <version>0.3.0</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>io.kroxylicious.testing</groupId>
+            <artifactId>testing-impl</artifactId>
+            <version>0.3.0</version>
+            <scope>test</scope>
+        </dependency>
```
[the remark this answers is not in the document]
tombentley: I would really rather not switch. I'm using some features of Kroxy testing which are not available in our own test containers. 

1. Most of the tests in the IT are using an in-VM cluster rather than using test containers, which means all ~200 tests run in about 2 minutes on my laptop. It would take much longer to run these tests if they were all using test containers, which would slow the development inner loop.
2. Kroxy testing provides a common abstraction over clusters which are in-VM and clusters which are running in containers, which is also a feature I'm using. It's actually very nice to be able to easily switch between the two when you realise that you need to resort to using containers.

In my opinion switching would require a lot of effort to rewrite very large test, which would then run slower for the rest of time, but for what actual benefit?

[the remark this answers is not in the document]
scholzj: I think we have a very bad experience with dependencies like that - e.g. the Debezium Kafka test clusters etc.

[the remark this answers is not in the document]
tombentley: What was the basis of the bad experience, exactly?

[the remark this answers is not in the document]
tombentley: * Almost all our dependencies are from outside projects which we cannot fix ourselves.
* How is a project to become well proven no one uses it? It's useful and it has an acceptable license. 
* The project in question has gone to the trouble of factoring this out as a separate dependency precisely for other projects to use (unlike DBZ ever did). They have specifically released this dependency to maven central. They've been making [regular releases](https://central.sonatype.com/artifact/io.kroxylicious.testing/testing-junit5-extension/0.3.0/versions). They're just about to release a new version with support for 3.5.0. 
* I don't think it's fair to use historical problems with a dependency from one project (Debezium) to make assertions about a dependency from another _completely different_ project. 

Finally, I wrote a lot of the code in this dependency, and work with the community in question, _unlike most of our other dependencies_. While that certainly makes me somewhat biased, it is also reason to doubt that this _really_ represents as much of a problem as you're suggesting.

topic-operator/src/main/java/io/strimzi/operator/topic/v2/BatchingLoop.java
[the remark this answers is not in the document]
fvaleri: > It would probably be best to exit the whole TO process in that case, forcing the user to increase the size of the deque. Wdyt?

I agree with that. The termination message should be something like "Insufficient event queue capacity, please raise by setting STRIMZI_MAX_QUEUE_SIZE environment variable".

That said, the STRIMZI_MAX_QUEUE_SIZE default value seems a bit low thinking about a busy cluster. Should be align with what we have on the user operator (1024)? We should also enforce STRIMZI_MAX_QUEUE_SIZE >= STRIMZI_MAX_BATCH_SIZE.

topic-operator/src/main/java/io/strimzi/operator/topic/v2/BatchingLoop.java
```diff
@@ -0,0 +1,302 @@
...
+                        batchId, topicUpsert);
+                return null;
+            }
+        }
+
+        private void fillBatch(int batchId, Batch batch) throws InterruptedException {
+            LOGGER.traceOp("[Batch #{}] Filling", batchId);
+            List<TopicEvent> rejected = new ArrayList<>();
+
+
```
ppatierno, line 230: double blank lines

topic-operator/src/test/java/io/strimzi/operator/topic/v2/LogCaptor.java
```diff
@@ -0,0 +1,263 @@
...
+ * {@code
+ *         try (var captor = LogCaptor.requireMatch(LOGGER, Level.INFO,
+ *                 le -> "Hello, world".equals(le.getMessage().getFormattedMessage()),
+ *                 50, TimeUnit.MILLISECONDS)) {
+ *             LOGGER.info("Hello, world");
+ *         }
+ * }
+ * </pre>
+ */
+public class LogCaptor implements AutoCloseable {
```
ppatierno, line 41: Am I wrong or this class is never used in the UTO logic? I see it just used in the corresponding tests.

tombentley: You're correct, and that's why it's in `topic-operator/src/test/java`.

topic-operator/src/main/java/io/strimzi/operator/topic/v2/BatchingLoop.java
```diff
@@ -0,0 +1,302 @@
...
+     * The loop is alive if none of the threads have been blocked for more than 2 minutes.
+     * "Blocked" means they're not returned to their outermost loop.
+     * @return True if the loop is alive..
+     */
+    boolean isAlive() {
+        for (var t : threads) {
+            if (!t.isAlive()) {
+                LOGGER.infoOp("isAlive returning false because {} is not alive", t);
+                return false;
+            } else if (t.msSinceLastLoop() > 120_000L) {
```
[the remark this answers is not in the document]
ppatierno: Maybe they are too low level parameters, but from where 2 mins come? Why it's not 1, 3, 5? Just trying to understand the rational behind that number.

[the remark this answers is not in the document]
tombentley: It didn't come from any principled argument. 1 seemed too small, and 3 seemed too long. What are the consequences of getting it wrong? Only that the TO will be restarted when it _might have_ recovered, or won't be restarted as quickly as it would have when it might not have recovered. I.e. even with the above logic added there would be no correct value.

LiamClarkeNZ:
@ppatierno @scholzj @tombentley What are the actual risks of using kroxylicious? What's the worst case scenario if it goes pear shaped? I can't see any negative effects beyond the ongoing effort to handle flakiness, or worst case, rewriting tests that use it.

Tom states it offers iteration speed and additional features over test containers, and frankly, the tests based on on test containers didn't have a stellar reputation for non-flakiness when I was last looking at them.

And besides, there's the eternal implicit social contract of developers about introducing new dependencies - you brought it in, you're the one getting asked to fix it. If Tom thinks the benefits are worth that ongoing responsibility, then I trust his decision, and his commitment to that implicit contract.

In short, there's no evidence of actual risk, just hypothetical maybes, and I'd suggest not letting the perfect be the enemy of the good.

topic-operator/pom.xml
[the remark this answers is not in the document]
LiamClarkeNZ: @ppatierno @scholzj @tombentley What are the actual risks of using kroxylicious? What's the worst case scenario if it goes pear shaped? I can't see any negative effects beyond ongoing effort to handle flakiness, or worst case, rewriting tests that use it. 

Tom states it offers iteration speed and additional features over test containers, faster iteration (red/green etc.) is very compelling in a testing tool, and tbh, the Strimzi tests built around test containers were rather brittle when I last looked at them.

With regards to the risk of needing ongoing effort to handle any flakiness, there's alway been an eternal implicit social contract among programmers when introducing new technologies - you brought it in, you're the one that people will expect to support and fix it. If Tom thinks the benefits of kroxylicious are worth shouldering that responsibility, then I trust his decision and his commitment to that implicit contract.

In short, there seems to be no evidence of actual risk, just hypotheticals and maybes, and I'd suggest not letting the perfect be the enemy of the good.

fvaleri (commented):
@tombentley looks like I can still trigger the bug:

```sh
$ git branch -v
* unidirectional-topic-operator 04ec96247 Avoid logging ERROR when KT deleted but using finalizers
$ docker images | grep quay.io/fvaleri/operator | head -n1
quay.io/fvaleri/operator                        latest                eef4203be514   12 minutes ago   506MB
$ kubectl exec my-cluster-kafka-0 -- bin/kafka-topics.sh --bootstrap-server my-cluster-kafka-bootstrap:9092 --describe --topic foo
Topic: foo	TopicId: 1K1oZNuKRvKgusjou-XhOg	PartitionCount: 3	ReplicationFactor: 1	Configs: min.insync.replicas=2,segment.bytes=1073741824,retention.ms=7200000,message.format.version=3.0-IV1
	Topic: foo	Partition: 0	Leader: 1	Replicas: 1	Isr: 1
	Topic: foo	Partition: 1	Leader: 2	Replicas: 2	Isr: 2
	Topic: foo	Partition: 2	Leader: 0	Replicas: 0	Isr: 0
$ kubectl get kt bar -o yaml
kind: KafkaTopic
metadata:
  creationTimestamp: "2023-06-20T16:57:25Z"
  finalizers:
  - strimzi.io/topic-operator
  generation: 1
  labels:
  name: bar
  namespace: test
  resourceVersion: "121752"
  uid: ca5e2909-4342-45c5-879b-14a2927b1df4
spec:
  config:
    retention.ms: 7200000
    segment.bytes: 1073741824
  partitions: 1
  replicas: 1
  topicName: foo
status:
  conditions:
  - lastTransitionTime: "2023-06-20T16:59:10.080782075Z"
    status: "True"
    type: Ready
  observedGeneration: 1
  topicName: foo
```

From the logs I see that the controller detects the partition decrease when foo is deleted, but then the bar's Ready status is updated to true. This goes on and on for every informer sync (every 2 minutes by default).

```sh
2023-06-21 07:30:46,09350 DEBUG [LoopRunnable-0] BatchingTopicController:877 - Reconciliation #14(upsert) KafkaTopic(test/bar): Partition decrease from 3 to 1
...
2023-06-21 07:30:46,11463 DEBUG [LoopRunnable-0] BatchingTopicController:971 - Reconciliation #14(upsert) KafkaTopic(test/bar): Updating status with KafkaTopicStatus(super=Status(conditions=[Condition(status=True, reason=null, message=null, type=Ready, lastTransitionTime=2023-06-21T07:30:46.113225042Z, additionalProperties={})], observedGeneration=1, additionalProperties={}), topicName=foo)
```

Debugging the issue, I can see that the expected error `Left(TopicOperatorException$NotSupported)` is in the results after partitions check, but it is overwritten by `Right(null)` when we update the status.

topic-operator/pom.xml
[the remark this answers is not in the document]
scholzj: > In short, there seems to be no evidence of actual risk, just hypotheticals and maybes, and I'd suggest not letting the perfect be the enemy of the good.

Seriously? Referring to actual past issues is `hypotheticals and maybes`. But some _wanna be funny_ references to non-existing social contracts are the real deal?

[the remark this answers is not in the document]
LiamClarkeNZ: @scholzj - actual past issues yep, but with a different library, extrapolating from bad experiences with Debezium to kroxylicious is what I deem hypothetical. Subjective, of course, so agree to disagree.

And wasn't trying for humour, Kiwi humour is dry, but it's not _that_ dry, that was my rather long-winded way of saying that "Tom isn't going to drop this dependency into the codebase via a drive-by PR and not support it",  and as such, the risk level is low.

topic-operator/src/main/java/io/strimzi/operator/topic/v2/BatchingLoop.java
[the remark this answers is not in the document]
tombentley: Bumped up the default size to 1024 as you suggest. `STRIMZI_MAX_QUEUE_SIZE >= STRIMZI_MAX_BATCH_SIZE` doesn't necessarily make sense, since the adding to the quieue and the batch creation happen in different threads, so you could actually have a bigger batch than queue and it would work, so long as the queue offer was not rejected.

I quite like the idea of using a termination message (written in `/dev/termination-log`, per [kube defaults](https://kubernetes.io/docs/tasks/debug/debug-application/determine-reason-pod-failure/)). But we'd need a mechanism to be able to test it without actually exiting the test JVM. Also interested in what @scholzj and @ppatierno think about this.

topic-operator/src/test/java/io/strimzi/operator/topic/v2/BatchingTopicControllerTest.java
```diff
@@ -0,0 +1,330 @@
...
+    }
+
+    private static <T> KafkaFuture<T> interruptedFuture() throws ExecutionException, InterruptedException {
+        var future = mock(KafkaFuture.class);
+        doThrow(new InterruptedException()).when(future).get();
+        return future;
+    }
+
+    private String namespace(String ns) {
+        //namespaces.push(ns);
```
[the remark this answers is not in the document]
tombentley: I've removed that commented out code and factored out some common functions into a `TopicOperatorTestUtil`. If you spot any more duplicated functions let me know.

topic-operator/src/main/java/io/strimzi/operator/topic/v2/BatchingLoop.java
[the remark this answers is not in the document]
scholzj: I think simply logging it will be equally if not more effective and a lot less trouble ... unless you plan any automated processing of the error message.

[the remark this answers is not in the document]
tombentley: Do you mean logging prior to exit (as opposed to a proper termination message)? Or logging and carrying on?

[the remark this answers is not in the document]
scholzj: I meant mainly the use of the termination log. TBH, I do not understand the issue well enough to comment on whether the operator should exit or not. I think that depends on:
* Whether it can recover
* How much you want the users to notice it (they are obviously much more likely to notice it when it starts crash looping -> but might not really care if it happens only once per week due to some run of events.). But once it would be restarting too often, the regular log is IMHO fine to find the cause.

I think the termination message might be useful especially if for example we have some further tooling to monitor the TO and suggest some insights to reconfigure it.

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/EntityOperator.java
scholzj, line 1: Should you also skip the TLS sidecar when UTO is enabled? Right now, you seem to always deploy it even if it is not needed/used. (UO does not use it anymore)

topic-operator/src/main/resources/log4j2.properties
```diff
@@ -9,3 +9,13 @@ rootLogger.level = ${env:STRIMZI_LOG_LEVEL:-INFO}
...
+
+## Kafka AdminClient logging is a bit noisy at INFO level
+#logger.kafka.name = org.apache.kafka
+#logger.kafka.level = ${env:STRIMZI_AC_LOG_LEVEL:-WARN}
+#logger.kafka.additivity = false
+#
+## Zookeeper is very verbose on INFO level , set it to WARN by default
+#logger.zookeepertrustmanager.name = org.apache.zookeeper
+#logger.zookeepertrustmanager.level = ${env:STRIMZI_ZOOKEEPER_LOG_LEVEL:-WARN}
+#logger.zookeepertrustmanager.additivity = false
```
scholzj, line 21: Why do we have the commented out code here? Its not like anyone can uncomment it when it is baked into the JARs.

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/EntityOperator.java
tombentley: Yes, I was thinking we can fix that in a follow up PR, but I guess I can also do it here if you'd prefer.

fvaleri:
@tombentley I confirm that the two bugs are now fixed. Thanks! 

Any comment on the last part of my previous comment? Can we make the use of finalizers opt-in (i.e. disabled by default)?

tombentley:
@fvaleri 

> Wdyt about the last part of my previous comment? Can we make finalizers opt-in (i.e. disabled by default)?

That would be different from what was approved in the proposal. I'm not wedded to keeping the current default, but I'd rather not change it on a whim. I think this is something we can get a better feeling for while the feature gate is still alpha. I.e. if people try and we get feedback that the finalizers really are problematic then it's still easily changed, imo.

fvaleri:
> I.e. if people try and we get feedback that the finalizers really are problematic then it's still easily changed, imo.

I have a bad feeling about this, but it can be disabled, so let's see what's the feedback.

fvaleri (approved):
LGTM.

The two failing STs are flaky and unrelated.

cluster-operator/src/test/java/io/strimzi/operator/cluster/model/EntityOperatorTest.java
```diff
@@ -126,8 +126,17 @@ static Map<String, String> volumeMounts(List<VolumeMount> mounts) {
...
+    public void testGenerateDeploymentWithBto() {
+        testGenerateDeployment(false);
+    }
 
+    @ParallelTest
+    public void testGenerateDeploymentWithUto() {
+        testGenerateDeployment(true);
+    }
+
+    private void testGenerateDeployment(boolean useUto) {
```
ppatierno, line 138: not sure `useUto` is really understandable :-) I would use the more "verbose" version.

topic-operator/pom.xml
```diff
@@ -22,6 +22,36 @@
...
+            <version>0.4.0</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <!-- Override since both kroxylicious.testing and strimzi-test-container have it as a dependency
+                 and inheriting the version from strimzi-test-container results in errors -->
+            <groupId>com.github.docker-java</groupId>
+            <artifactId>docker-java-api</artifactId>
+            <version>3.3.0</version>
+            <scope>test</scope>
```
ppatierno, line 53: is there any reason why the all the above dependencies aren't using a property to define the version?

topic-operator/src/main/java/io/strimzi/operator/topic/v2/BatchingLoop.java
```diff
@@ -0,0 +1,309 @@
...
+     * "Blocked" means they're not returned to their outermost loop.
+     * @return True if the loop is alive..
+     */
+    boolean isAlive() {
+        for (var thread : threads) {
+            if (!thread.isAlive()) {
+                LOGGER.infoOp("isAlive returning false because {} is not alive", thread);
+                return false;
+            } else if (thread.msSinceLastLoop() > 120_000L) {
+                LOGGER.infoOp("isAlive returning false because {} appears to be stuck", thread);
```
ppatierno, line 116: would be the above "WARN" instead of "INFO"?

topic-operator/src/main/java/io/strimzi/operator/topic/v2/BatchingLoop.java
```diff
@@ -0,0 +1,309 @@
...
+
+    /**
+     * The loop is ready is all the threads have started and none have stopped.
+     * I.e. all the threads are alive.
+     * @return True if the loop is ready.
+     */
+    boolean isReady() {
+        for (var thread : this.threads) {
+            if (!thread.isAlive()) {
+                LOGGER.infoOp("isReady returning false, because {} is not alive", thread);
```
ppatierno, line 132: Ditto as above.

topic-operator/src/main/java/io/strimzi/operator/topic/v2/BatchingLoop.java
```diff
@@ -0,0 +1,309 @@
...
+                if (batch.size() > 0) {
+                    LOGGER.infoOp("[Batch #{}] Reconciling {} topics", batchId, batch.size());
+                    // perform reconciliation on new batch
+                    if (!batch.toUpdate.isEmpty()) {
+                        controller.onUpdate(batch.toUpdate.stream().map(upsert -> lookup(batchId, upsert)).filter(Objects::nonNull).toList());
+                    }
+                    if (!batch.toDelete.isEmpty()) {
+                        controller.onDelete(batch.toDelete.stream().map(td -> new ReconcilableTopic(new Reconciliation("delete", "KafkaTopic", td.namespace(), td.name()), td.topic(), BatchingTopicController.topicName(td.topic()))).toList());
+                    }
+                    LOGGER.debugOp("[Batch #{}] Reconciled batch", batchId);
```
ppatierno, line 203: this is at "DEBUG" but the corresponding "opening" log above is at "INFO" ... isn't it better to have them both at "DEBUG" to avoid polluting log on reconcile?

topic-operator/src/main/java/io/strimzi/operator/topic/v2/BatchingTopicController.java
```diff
@@ -0,0 +1,1005 @@
...
+        if (kt.getSpec() != null) {
+            tn = kt.getSpec().getTopicName();
+        }
+        if (tn == null) {
+            tn = kt.getMetadata().getName();
+        }
+        return tn;
+    }
+
+    static String rv(KafkaTopic kt) {
```
ppatierno, line 135: `rv` ... isn't it too short? :-D

topic-operator/src/main/java/io/strimzi/operator/topic/v2/BatchingTopicController.java
```diff
@@ -0,0 +1,1005 @@
...
+        return collect;
+    }
+
+    private KafkaTopic addFinalizer(ReconcilableTopic reconcilableTopic) {
+        if (!reconcilableTopic.kt().getMetadata().getFinalizers().contains(FINALIZER)) {
+            LOGGER.debugCr(reconcilableTopic.reconciliation(), "Adding finalizer {}", FINALIZER);
+            long t0 = System.nanoTime();
+            KafkaTopic edit = Crds.topicOperation(kubeClient).resource(reconcilableTopic.kt()).edit(old ->
+                    new KafkaTopicBuilder(old).editOrNewMetadata().addToFinalizers(FINALIZER).endMetadata().build());
+            LOGGER.traceCr(reconcilableTopic.reconciliation(), "Added finalizer {}, took {}ns, resourceVersion now {}", FINALIZER, System.nanoTime() - t0,
```
ppatierno, line 155: out of curiosity, why are you interested in how long it takes to add the finalizer?

topic-operator/src/main/java/io/strimzi/operator/topic/v2/BatchingTopicController.java
```diff
@@ -0,0 +1,1005 @@
...
+        return reconcilableTopic.kt();
+    }
+
+    private KafkaTopic removeFinalizer(ReconcilableTopic reconcilableTopic) {
+        if (reconcilableTopic.kt().getMetadata().getFinalizers().contains(FINALIZER)) {
+            LOGGER.debugCr(reconcilableTopic.reconciliation(), "Removing finalizer {}", FINALIZER);
+            long t0 = System.nanoTime();
+            var result = Crds.topicOperation(kubeClient).resource(reconcilableTopic.kt()).edit(old ->
+                    new KafkaTopicBuilder(old).editOrNewMetadata().removeFromFinalizers(FINALIZER).endMetadata().build());
+            LOGGER.traceCr(reconcilableTopic.reconciliation(), "Removed finalizer {}, took {}ns, resourceVersion now {}", FINALIZER, System.nanoTime() - t0,
```
ppatierno, line 168: Ditto as above.

topic-operator/src/main/java/io/strimzi/operator/topic/v2/ReconcilableTopic.java
```diff
@@ -0,0 +1,18 @@
+/*
+ * Copyright Strimzi authors.
+ * License: Apache License 2.0 (see the file LICENSE or http://apache.org/licenses/LICENSE-2.0.html).
+ */
+package io.strimzi.operator.topic.v2;
+
+import io.strimzi.api.kafka.model.KafkaTopic;
+import io.strimzi.operator.common.Reconciliation;
+
+record ReconcilableTopic(Reconciliation reconciliation, KafkaTopic kt, String topicName) {
```
ppatierno, line 10: Javadoc missing?

topic-operator/src/main/java/io/strimzi/operator/topic/v2/TopicEvent.java
```diff
@@ -0,0 +1,92 @@
...
+ * License: Apache License 2.0 (see the file LICENSE or http://apache.org/licenses/LICENSE-2.0.html).
+ */
+package io.strimzi.operator.topic.v2;
+
+import io.strimzi.api.kafka.model.KafkaTopic;
+
+import java.util.Objects;
+
+/**
+ *
```
ppatierno, line 12:
```suggestion
```

topic-operator/src/main/java/io/strimzi/operator/topic/v2/TopicEvent.java
```diff
@@ -0,0 +1,92 @@
...
+
+    default KubeRef toRef() {
+        return new KubeRef(namespace(), name(), 0);
+    }
+}
+/**
+ * The creation or update of a KafkaTopic.
+ * Note that this may include the change which adds a metadata.deletionTimestamp.
+ */
+record TopicUpsert(long nanos, String namespace, String name, String resourceVersion) implements TopicEvent {
```
ppatierno, line 29: is `nanos` the "timestamp" of the event? Can we make it clearer with a better name?

topic-operator/src/main/java/io/strimzi/operator/topic/v2/TopicEvent.java
```diff
@@ -0,0 +1,92 @@
...
+                ", name='" + name + '\'' +
+                ", resourceVersion='" + resourceVersion + '\'' +
+                ", ageNs='" + ageNs() + '\'' +
+                '}';
+    }
+}
+/**
+ * The deletion of a KafkaTopic
+ */
+record TopicDelete(long nanos, KafkaTopic topic) implements TopicEvent {
```
ppatierno, line 61: Ditto as above about `nanos`.

topic-operator/src/main/java/io/strimzi/operator/topic/v2/TopicOperatorConfig.java
```diff
@@ -0,0 +1,247 @@
...
+                "\n\ttlsEnabled=" + tlsEnabled +
+                "\n\ttruststoreLocation='" + truststoreLocation + '\'' +
+                "\n\ttruststorePassword='" + mask + '\'' +
+                "\n\tkeystoreLocation='" + keystoreLocation + '\'' +
+                "\n\tkeystorePassword='" + mask + '\'' +
+                "\n\tsslEndpointIdentificationAlgorithm='" + sslEndpointIdentificationAlgorithm + '\'' +
+                "\n\tsaslEnabled=" + saslEnabled +
+                "\n\tsaslMechanism='" + saslMechanism + '\'' +
+                "\n\tsaslUsername='" + saslUsername + '\'' +
+                "\n\tsaslPassword='" + mask + '\'' +
```
ppatierno, line 239: what's the purpose of having a `mask` variable to be concatenated with other corresponding "password" related fields instead of having just the asterisks in the string and avoid concatenation (which allocate a new String each time)?

ppatierno (commented):
I had another pass, tbh I haven't tried running the code. So I left some nits and questions.
Also, I was wondering why when using `record` most of the corresponding fields/members are not documented as we do for class constructor arguments instead.

topic-operator/src/main/java/io/strimzi/operator/topic/v2/BatchingTopicController.java
tombentley: It's not the finalizer specifically, there is trace level logging for all communication between both Kafka and Kube. It might be better as a metric, but we don't have metrics yet (see #8650). Collecting this information _somehow_ will enable us to reason in the future about what changes might be needed to support large numbers of topics.

topic-operator/src/main/java/io/strimzi/operator/topic/v2/TopicEvent.java
tombentley: It results from a call to `System.nanoTime()`. Therefore it's a value that has no meaning on it's own (the epoch it not defined). It can only be used with the result of another `System.nanoTime()`, e.g. to compute a duration. So it's a bit tricky to know _what_ to call it. Do you have any suggestions?

cluster-operator/src/main/java/io/strimzi/operator/cluster/FeatureGates.java
```diff
@@ -107,6 +114,14 @@ public boolean stableConnectIdentitiesEnabled() {
     public boolean kafkaNodePoolsEnabled() {
         return kafkaNodePools.isEnabled();
     }
+
+    /**
+     * @return  Returns true when the UnidirectionalTopicOperator feature gate is enabled
```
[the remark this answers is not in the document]
tombentley: I use the same `@return` phrasing as the other methods in this class. Do you want me to change them all?

topic-operator/pom.xml
tombentley: I fixed the kroxy ones. For this one the comment above explains why.

topic-operator/src/main/java/io/strimzi/operator/topic/v2/TopicEvent.java
ppatierno: > It results from a call to System.nanoTime(). Therefore it's a value that has no meaning on it's own

well should it have a meaning? Why this nanoTime() call is made, which is the purpose? I see you are using to process an age ... `System.nanoTime() - nanos` ... but age of what? Of the `TopicUpsert` itself? So is the nanos parameter a kind of born date?

cluster-operator/src/main/java/io/strimzi/operator/cluster/FeatureGates.java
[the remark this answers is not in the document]
ppatierno: Well it's just my preference to see the `@return` as the verb ... so I can just explain what it's returning, no need to say "Returns". Not a big deal of course, just personal preference.

topic-operator/src/main/java/io/strimzi/operator/topic/v2/TopicEvent.java
tombentley: > Why this nanoTime() call is made, which is the purpose?

That's a good catch. We're not currently using it really (apart from the `toString()` output that isn't really called). I was using it to record the total time it took to reconcile an event. We could remove it, but we will be adding a metric (`strimzi.reconciliations.duration`) which will need it, so I'm inclined to keep it.

> So is the nanos parameter a kind of born date?

Yes, a kind of born "date", but relative to an unknown epoch (in particular it's not a milliseconds offset from the unix epoch,  like from `System.currentTimeMillis()`). So call it `bornDate` would be misleading, I think. I suppose `nanosOffset` would be sort-of better.  Wdyt?

ppatierno: >  I suppose nanosOffset would be sort-of better. Wdyt?

Well the parameter would be a `nanosStartOffset` because you are going to get the "elapsed" nanos time via `System.nanoTime() - nanosStartOffset` where the current nanos time would be the end offset.

PR #8636, opened 2023-06-07
Replace Vert.x with standard Java concurrency API in common module

### LINKED ISSUE #7594: Refactor _resource operators_ to be independent on Vert.x
Resource operators (see package `io.strimzi.operator.common.operator.resource` in module `operator-common`) are classes which make it easier to work with Kubernetes resources. They contain several kinds of logic:
* `reconcile` mechanism which decides whether a resource should be created, deleted or patched
* Diffing mechanism which decides whether it is needed to patch the resource (whether there is any difference)
* Async mechanisms which can make some synchronous blocking calls async
* Additional checks such as making sure that deleted resource is already deleted

These classes currently heavily depend on Vert.x. They return Vert.x futures, the use Vert.x futures and blocking execution etc. That makes them impossible to use when not using Vert.x such as from the new User operator or in the PodSet controller. 

The logic in these classes is useful even when not using Vert.x. So we should make them useable also without it. We should investigate:
* If they can be rewritten to use pure Java (e.g. completion stages or completable futures) while still being used form the Vert.x part of the code
* Whether we need to create a parallel classes for use without Vert.x

This should be also implemented.

MikeEdgar:
### Type of change

_Select the type of your PR_

- ~~Bugfix~~
- ~~Enhancement / new feature~~
- Refactoring
- ~~Documentation~~

### Description

Replace Vert.x with standard Java concurrency API in common module. The PR introduces a `CompletableFuture` subclass `StrimziFuture` that also implements the Vert.x `FutureInternal` interface. This accomplishes two things - (1) provides a central location to manage a separate `ForkJoinPool` for async workloads and (2) allows existing interactions with the operator-common module to continue using the `Future` API. The expectation is that `StrimziFuture` may be a temporary/transitional class. However, the ability to manage the work thread pool centrally may prove useful over time. Currently, the pool parallelism is fixed at 20 to match the Vert.x worker pool size.

Closes #7594 

### Checklist

- [ ] Write tests
- [ ] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/operator-common/src/main/java/io/strimzi/operator/common/StrimziFuture.java
+++ b/operator-common/src/main/java/io/strimzi/operator/common/StrimziFuture.java
@@ -1,0 +1,328 @@
+/*
+ * Copyright Strimzi authors.
+ * License: Apache License 2.0 (see the file LICENSE or http://apache.org/licenses/LICENSE-2.0.html).
+ */
+package io.strimzi.operator.common;
+
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.Objects;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.CompletionStage;
+import java.util.concurrent.Executor;
+import java.util.concurrent.ForkJoinPool;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicReference;
+import java.util.function.Function;
+import java.util.function.Supplier;
+
+import io.vertx.core.AsyncResult;
+import io.vertx.core.Future;
+import io.vertx.core.Handler;
+import io.vertx.core.impl.ContextInternal;
+import io.vertx.core.impl.future.FutureInternal;
+import io.vertx.core.impl.future.Listener;
+
+/**
+ * Custom {@linkplain CompletableFuture} that also implements the Vert.x
+ * {@linkplain FutureInternal} interface. May be used as a drop-in for either
+ * type, with the intention of helping transition from Vert.x to the Java
+ * concurrency APIs. Implements FutureInternal rather than just Future because
+ * some Vert.x code assumes (and casts) Futures to FutureInternal.
+ *
+ * @param <T> The result type returned by this future's {@code join} and
+ *            {@code get} methods
+ *
+ * @see CompletableFuture
+ * @see Future
+ * @see FutureInternal
+ */
+public class StrimziFuture<T> extends CompletableFuture<T> implements FutureInternal<T> {
+
+    private static final Executor ASYNC_POOL = new ForkJoinPool(20);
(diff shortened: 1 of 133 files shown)


operator-common/src/main/java/io/strimzi/operator/common/StrimziFuture.java
```diff
@@ -0,0 +1,328 @@
...
+import java.util.concurrent.atomic.AtomicReference;
+import java.util.function.Function;
+import java.util.function.Supplier;
+
+import io.vertx.core.AsyncResult;
+import io.vertx.core.Future;
+import io.vertx.core.Handler;
+import io.vertx.core.impl.ContextInternal;
+import io.vertx.core.impl.future.FutureInternal;
+import io.vertx.core.impl.future.Listener;
```
scholzj, line 24: Would it be possible to isolate the use of Vert.x to the cluster / topic operator? That would allow us to get rid of the Vert.x as dependency in the `operator-common` and thus not pull the dependency into the operators using it which do not use Vert.x.

scholzj: Hmm, I guess optional would be fine, we can control that. If not, maybe we need to reconsider together with @tombentley what does it actually bring. It would for sure still make it possible to use the resource operators without running on Vert.x even if it was still a dependency.

MikeEdgar: It seems that the optional dependency approach isn't going to work without some duplication. I.e. a Java concurrency and a Vert.x version of the resource operators. Would it be fine to keep the transitive dependency on Vert.x, avoiding all direct usage of the APIs other than by the cluster/topic operators? Then, once those two are migrated to use the standard concurrency API, `StrimziFuture` can shed its Vert.x `Future` implementation and the dependency can be made optional, or possibly removed entirely.

scholzj: @tombentley What do you think about this?

scholzj: The option 3) was what I was originally thinking about. But I can see how your approach is much more elegant from the perspective of the CO changes.

One other thing to maybe consider ... I think the User Operator and Topic Operator really manage only 3 resources: KafkaUsers, KafkaTopics and Secrets. I do not think there is anything else. So in theory, one could also simply move the majority of the resource operators to the CO module and address only the Secret operator which is probably the only shared resource among all of them. But I'm not sure how much it really simplifies this since they all inherit from the same parent classes etc.

MikeEdgar:
Closing in favor of #8673

PR #8639, opened 2023-06-07
[systemtest] Two small fixes for flaky tests

im-konge:
### Type of change

- Test fixes

### Description

This small PR fixes two flaky tests:

`ConnectIsolatedST#testScaleConnectWithConnectorToZero` - problem with KafkaConnector and its status - we are checking that the status is `NotReady`, but there was race condition, when from time to time the status of the connector wasn't switched and the test failed

`OauthScopeIsolatedST#testScopeKafkaConnectSetCorrectly` - because the logs of Kafka pods contain many things, the last 30 lines could be filled up with different logs than the needed one.

### Checklist


henryZrncik (approved):
looks good to me.

Frawless (approved):

PR #8659, opened 2023-06-12
Grafana dashboards: fix default datasource

### LINKED ISSUE #8545: [Bug]: Grafana dashboards through Helm chart missing datasource
### Bug Description

Setting `dashboards.enabled: true` adds dashboards to Grafana that give the following error `Failed to upgrade legacy queries Datasource ${DS_PROMETHEUS} was not found` in Grafana 9.5.2.
I suspect this is because the dashboards only set `DS_PROMETHEUS` on import through the UI.

### Steps to reproduce

1. Set `dashboards.enabled` to `true`
2. Wait until dashboards are loaded in Grafana
3. Go to one of the added dashboards
4. Observe the error

### Expected behavior

Same behavior as on manual import; dashboard should use a datasource, preferably the default datasource. 

### Strimzi version

0.35.0

### Kubernetes version

1.26

### Installation method

Helm chart

### Infrastructure

Azure Kubernetes

_No response_

### Additional context

This could probably be solved by removing the `__inputs` part of the dashboard and adding the following to `templating.list`:

```json
      {
        "current": {
          "selected": false,
        },
        "description": null,
        "error": null,
        "hide": 0,
        "includeAll": false,
        "label": null,
        "multi": false,
        "name": "DS_PROMETHEUS",
        "options": [],
        "query": "prometheus",
        "refresh": 1,
        "regex": "",
        "skipUrlSync": false,
        "type": "datasource"
      }
```
Which is what we currently use for some of the Strimzi dashboards that we manually deploy

tjanson:
### Type of change

- Bugfix 

### Description

/fixes #8545

Removes the datasource input variable and leaves the `datasource` fields blank instead, which leads to Grafana correctly choosing the default datasource on its own.

Note: I don't really know what I'm doing with regards to the Grafana dashboard input datasource behavior (and it seems I'm not alone).
However, this change seems to be sufficient to work out-of-the-box with a "standard" Prometheus/Grafana setup, such as the one deployed by [kube-prometheus-stack](https://github.com/prometheus-operator/kube-prometheus) and derivatives.

I'm not sure if this is the _best_ solution, but it certainly seems better than the current state. The discussion in #8545 did not identify a better solution as far as I can tell.

### Checklist

- [ ] Write tests
- [ ] Update documentation
- [ ] Update CHANGELOG.md


tjanson:
Rebased to change files in correct folder, should be good now.

scholzj:
Thanks for the PR.

The original files are maintained in https://github.com/strimzi/strimzi-kafka-operator/tree/main/packaging/examples/metrics/grafana-dashboards -> the files in the Helm Chart are just a copy of them. So your changes to them will be overwritten in the future. However, those files are not used only with kube-prometheus-stack or whatever the name was. So changing them is not as simple because your change likely breaks things for others.

So we need to either - as discussed in #8545 - find a solution that works for everyone. Or you would need to modify the charts either directly in the Helm Chart when packaging them into JSON or for example, modify them in the `Makefile` while they are being copied.

scholzj:
> It's possible that there is no solution that works for everyone, and in particular the current one does not. I believe the change proposed in this PR makes the dashboards compatible with the majority of use cases, indeed hopefully the overwhelming majority, thus making it a worthwhile change imho. (That's also my takeaway from the discussion in #8545.)
> 
> Whether this constitutes a breaking change for some users is a good question, and if so, whether it could be avoided – I'm not sure. In the Helm Chart, the datasource could be made configurable, possibly defaulting to the current value. But as for the JSON dashboard files themselves – I guess there's no way to avoid a breaking change. Maybe it's worth it?

Honestly, the issue has like 4 people who commented about it plus your thumbs-up. The triage on the community call also suggested that this does not seem like the clear thing to just go and remove it without more testing and discussion. 

Also, that discussion is specifically about the Helm Chart Config Maps added in Strimzi 0.35 (that is why I offered as a possible solution to make it the way that it is changed directly in the Helm Chart). Not about the dashboards used by most users which are in the actual source directory. So it hardly constitutes the majority take.

tjanson:
Sorry, I understood now that you suggested modifying the dashboards specifically for the Helm Chart. Do you think that's the way to go? If so, I could go that route.

My impression is that getting rid of the `DS_PROMETHEUS` default would be benefitial for all users, but I could definitely be mistaken. The _impression_ I got from not just #8545 but the related issues on Grafana's side is that the DS_PROMETHEUS thing is at least outdated if not the artefact of a broken (from the end user perspective) export mechanism on the side of Grafana, affecting most people. (But that's just that, my impression.)

tjanson:
My bad, I reread the comments in #8545 and see that leaving the datasource empty has been discussed and deemed at least somewhat problematic (in https://github.com/strimzi/strimzi-kafka-operator/issues/8545#issuecomment-1567426858). Feel free to reject this, perhaps I'll give it a fresh go within the confines of the Helm templating to make it configurable.

PR #8664, opened 2023-06-13
feat(helm): Grafana dashboards: Make datasource configurable

tjanson:
### Type of change

- Bugfix, sort of: #8545
- Enhancement / new feature

### Description

Allows overriding the datasource variable in the Grafana dashboard JSONs when deployed via Helm, via the value `dashboards.datasource`.
This replacement is done via a regex on the JSON file source.

Implementation note: [`regexReplaceAllLiteral`](https://helm.sh/docs/chart_template_guide/function_list/#regexreplaceallliteral-mustregexreplaceallliteral) is used to avoid having to escape the `$` in `${DS_PROMETHEUS}` (as `$$`), which would be either confusing or require more processing.

Limitations: Using a regular expression to process JSON is ugly. However, I'm not aware of a better way.

Decisions: Should we default to `dashboards.datasource: ""`, which should work out-of-the-box when using kube-prometheus-stack?

### Checklist

- [ ] Write tests
- [ ] Update documentation
- [ ] Update CHANGELOG.md


tjanson:
Aside from the regex arguably being a hacky, non-robust way to alter the JSON, I think this is a good solution, since we can stick to the current behavior (i.e., no breaking change), while offering an easy solution for everyone affected by #8545.

scholzj:
> Decisions: Should we default to dashboards.datasource: "", which should work out-of-the-box when using kube-prometheus-stack?

As [mentioned](https://github.com/strimzi/strimzi-kafka-operator/pull/8659#issuecomment-1588154714) yesterday by @jcpunk => it seems to work for some as it is. So the question is ... does removing the data source break it for those users? If not, I think it is fine to have it set to `""` by default.

scholzj (commented):
Thanks for the PR. I think the logic looks good. Two things ...
* Could you maybe add a note about the new option to the [README.md](https://github.com/strimzi/strimzi-kafka-operator/blob/main/packaging/helm-charts/helm3/strimzi-kafka-operator/README.md)
* Can you please fix the DCO signoff? The instructions should be under the _Details_ link next to the DCO status.

tjanson:
> So the question is ... does removing the data source break it for those users? If not, I think it is fine to have it set to `""` by default.

I'm not sure, and that's why I went for the most conservative option (not changing the default) for now – unless there's more feedback.

I'll fix the DCO and README.

scholzj (approved):
Thanks.

Frawless (approved):
LGTM

ppatierno (approved):

scholzj:
Thanks for the PR @tjanson

jcpunk:
Long term it might be better to make the datasource an input to the dashboards for non-helm users.

PR #8668, opened 2023-06-14
[systemtest] Remove YAML desc from upgrade tests log on INFO level

im-konge:
### Type of change

- Enhancement

### Description

This PR changes the log level for methods used in upgrade/downgrade tests to `DEBUG` level from `INFO` level. It will remove the unnecessary description of all YAMLs applied inside the test run.

Also, it adds few lines about resource creation to make the log more clear.

### Checklist


Frawless (approved):

scholzj (approved):
Thanks

PR #8670, opened 2023-06-14
[DOC] Add StrimziPodSet to the API reference

scholzj:
### Type of change

- Documentation

### Description

Originally, we did not include the `StrimziPodSet` specification into the API reference guide as it is only an internal resource. But since people are noticing it anyway and often reading it, it makes sense to have it there to make it easier for them to understand the structure. This PR includes it there with a warning that it is an internal resource and users are not expected to create, modify or delete it.

### Checklist

- [x] Update documentation

### DIFF
--- a/api/src/main/java/io/strimzi/api/kafka/model/status/StrimziPodSetStatus.java
+++ b/api/src/main/java/io/strimzi/api/kafka/model/status/StrimziPodSetStatus.java
@@ -30,7 +30,7 @@ public class StrimziPodSetStatus extends Status {
     private int readyPods;
     private int currentPods;
 
-    @Description("Number of pods managed by the StrimziPodSet controller.")
+    @Description("Number of pods managed by this StrimziPodSet resource.")
     @JsonInclude(value = JsonInclude.Include.NON_NULL)
     public int getPods() {
         return pods;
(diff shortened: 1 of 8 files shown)


ppatierno (approved):

fvaleri (approved):

api/src/main/java/io/strimzi/api/kafka/model/status/StrimziPodSetStatus.java
```diff
@@ -50,7 +50,7 @@ public void setReadyPods(int readyPods) {
         this.readyPods = readyPods;
     }
 
-    @Description("Number of pods managed by the StrimziPodSet controller that have the current revision.")
+    @Description("Number of pods managed by this StrimziPodSet resource that have the current revision.")
```
PaulRMellor, line 53:
```suggestion
    @Description("Number of pods managed by this `StrimziPodSet` resource that have the current revision.")
```

documentation/api/io.strimzi.api.kafka.model.StrimziPodSet.adoc
```diff
@@ -0,0 +1,2 @@
+IMPORTANT: `StrimziPodSet` is an internal Strimzi resource.
+Strimzi users are not expected to create, modify or delete `StrimziPodSet` resources.
```
PaulRMellor, line 2: Is it strong enough? What happens if they do?

Maybe something like:

IMPORTANT: `StrimziPodSet` is an internal Strimzi resource.
Information is provided for reference only.
Do not create, modify or delete `StrimziPodSet` resources as this might cause errors.

PaulRMellor (approved):
Looks good. I left a couple of suggestions

PR #8672, opened 2023-06-14
Updated Kafka node pools examples with 3.4.1 version

ppatierno:
Trivial PR to use the Kafka version 3.4.1 in the node pools examples as it is already used in the others (non node pools based).


scholzj (approved):
I'm not sure it is worth it as it will just cause conflict with the 3.5.0 branch which is changing those to 3.5.0. But if you want to merge it ...

ppatierno:
I don't see any already opened PR with Kafka 3.5.0.

scholzj:
As I said ... you can merge it if you want - that is why I approved it. But it will just mean that I will have to rebase and solve the conflicts later 🤷.

ppatierno:
It will be an easy conflict to fix but I don't want put you in troubles :-)
Going to close this.

PR #8674, opened 2023-06-15
Improved ignored annotations in PvcOperator

scholzj:
### Type of change

- Bugfix

### Description

The `PvcOperator` class already ignores some Kubernetes annotations. But as discussed in #8667, there are some more that we should ignore as well and not patch them. This PR expands the range of ignored annotations.

### Checklist

- [x] Write tests


ppatierno (approved):

fvaleri (approved):
That's another use case relevant for https://github.com/strimzi/proposals/pull/80.

PR #8680, opened 2023-06-16
Change minikube registry port

kornys:
### Type of change

- Bugfix

### Description

Minikube registry should be changed to port 58925 when docker is used as driver.


im-konge (approved):

kornys:
closing it, this is not a way how to fix minikube registry on arm64

PR #8683, opened 2023-06-16
docs(node pools): adds docs for the node pools preview

PaulRMellor:
Documentation

Adds doc for the node pools preview

- Overview: adds a summary of the configuration.
- Deploying and Managing guide:  Adds the following:  
    -  Procedure to deploy node pools
    - Config example
    - Procedure to migrate a Kafka cluster to use node pools
    - Procedure to move nodes between node pools
    - Updates to feature gates section
    - Updates to common config to mention node pools support

### Checklist

- [ ] Write tests
- [x] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/documentation/assemblies/deploying/assembly-deploy-kafka-cluster.adoc
+++ b/documentation/assemblies/deploying/assembly-deploy-kafka-cluster.adoc
@@ -19,7 +19,7 @@ After you have deployed the Cluster Operator, use a `Kafka` resource to deploy t
 When installing Kafka, Strimzi also installs a ZooKeeper cluster and adds the necessary configuration to connect Kafka with ZooKeeper.
 
 If you are trying the preview of the node pools feature, you can deploy a Kafka cluster with one or more node pools.
-Node pools provide configuration for a set of kafka nodes.
+Node pools provide configuration for a set of Kafka nodes.
 By using node pools, nodes can have different configuration within the same Kafka cluster.
 
 Node pools are not enabled by default, so you must xref:ref-operator-kafka-node-pools-feature-gate-{context}[enable the `KafkaNodePools` feature gate] before using them. 
--- a/documentation/assemblies/deploying/assembly-deploy-kafka-cluster.adoc
+++ b/documentation/assemblies/deploying/assembly-deploy-kafka-cluster.adoc
@@ -12,11 +12,17 @@ You can use these files to deploy the Topic Operator and User Operator at the sa
 
 After you have deployed the Cluster Operator, use a `Kafka` resource to deploy the following components:
 
-* xref:deploying-kafka-cluster-{context}[Kafka cluster]
+* xref:deploying-kafka-cluster-{context}[Kafka cluster] or (preview) xref:deploying-kafka-node-pools-{context}[Kafka cluster with node pools]
 * xref:deploying-the-topic-operator-using-the-cluster-operator-{context}[Topic Operator]
 * xref:deploying-the-user-operator-using-the-cluster-operator-{context}[User Operator]
 
 When installing Kafka, Strimzi also installs a ZooKeeper cluster and adds the necessary configuration to connect Kafka with ZooKeeper.
+
+If you are trying the preview of the node pools feature, you can deploy a Kafka cluster with one or more node pools.
+Node pools provide configuration for a set of kafka nodes.
+By using node pools, nodes can have different configuration within the same Kafka cluster.
+
+Node pools are not enabled by default, so you must xref:ref-operator-kafka-node-pools-feature-gate-{context}[enable the `KafkaNodePools` feature gate] before using them. 
 
 If you haven't deployed a Kafka cluster as a `Kafka` resource, you can't use the Cluster Operator to manage it.
 This applies, for example, to a Kafka cluster running outside of Kubernetes.
@@ -25,6 +31,8 @@ You can also deploy and use other Kafka components with a Kafka cluster not mana
 
 //Deploy Kafka cluster with storage option
 include::../../modules/deploying/proc-deploy-kafka-cluster.adoc[leveloffset=+1]
+//Deploy Kafka node pools
+include::../../modules/deploying/proc-deploy-kafka-node-pools.adoc[leveloffset=+1]
 //Include Topic Operator in deployment
 include::../../modules/deploying/proc-deploy-topic-operator-with-cluster-operator.adoc[leveloffset=+1]
 //Include User Operator in deployment
--- a/documentation/modules/configuring/con-config-node-pools.adoc
+++ b/documentation/modules/configuring/con-config-node-pools.adoc
@@ -1,0 +1,105 @@
+// Module included in the following assemblies:
+//
+// assembly-config.adoc
+
+[id='config-node-pools-{context}']
+= (Preview) Configuring node pools
+
+[role="_abstract"]
+Update the `spec` properties of the `KafkaNodePool` custom resource to configure a node pool deployment.
+
+NOTE: The node pools feature is available as a preview. Node pools are not enabled by default, so you must xref:ref-operator-kafka-node-pools-feature-gate-{context}[enable the `KafkaNodePools` feature gate] before using them. 
+
+A node pool refers to a distinct group of Kafka nodes within a Kafka cluster.
+Each pool has its own unique configuration, which includes mandatory settings for the number of replicas, roles, and storage allocation.
+
+Optionally, you can also specify values for the following properties:
+
+* `resources` to specify memory and cpu requests and limits
+* `template` to specify custom configuration for pods and other Kubernetes resources
+* `jvmOptions` to specify custom JVM configuration for heap size, runtime and other options 
+
+The `Kafka` resource represents the configuration for all nodes in the Kafka cluster.
+The `KafkaNodePool` resource represents the configuration for nodes only in the node pool.
+Configuration options not specified in the `KafkaNodePool` resource are inherited from the `Kafka` resource.
+Configuration specified in the `KafkaNodePool` resource takes precedence if set in both resources
+For example, if `jvmOptions` are specified in the node pool configuration and Kafka configuration, the node pool configuration is used.
+
+Node pools can be used with Kafka clusters that operate in KRaft mode (using Kafka Raft metadata) or use ZooKeeper for cluster management.
+If you are using KRaft mode, you can specify roles for all nodes in the node pool to operate as brokers, controllers, or both.
+If you are using ZooKeeper, nodes must be set as brokers only.
+
+IMPORTANT: **KRaft mode is not ready for production in Apache Kafka or in Strimzi.**
+
+For a deeper understanding of the node pool configuration options, refer to the link:{BookURLConfiguring}[Strimzi Custom Resource API Reference^].
+
+NOTE: While the `KafkaNodePools` feature gate that enables node pools is in alpha phase, replica and storage configuration properties in the `KafkaNodePool` resource must also be present in the `Kafka` resource. The configuration in the `Kafka` resource is ignored when node pools are used. Similarly, ZooKeeper configuration properties must also be present in the `Kafka` resource when using KRaft mode. These properties are also ignored.
+
+.Example configuration for a node pool in a cluster using KRaft mode 
+[source,yaml,subs="+attributes"]
+----
+apiVersion: {KafkaNodePoolApiVersion}
+kind: KafkaNodePool
+metadata:
+  name: dual-role # <1>
+  labels:
+    strimzi.io/cluster: my-cluster # <2>
+spec:
+  replicas: 3 # <3>
+  roles: # <4>
+    - controller
+    - broker
+  storage: # <5>
+    type: jbod
+    volumes:
+      - id: 0
+        type: persistent-claim
+        size: 20Gi
+        deleteClaim: false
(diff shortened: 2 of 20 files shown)


PaulRMellor:
I have a couple of questions about this content:
Do we need to say anything about default heap settings in context of node pools? (https://strimzi.io/docs/operators/in-development/configuring.html#con-common-configuration-jvm-reference)
Do we say anything about changing roles -- best approach?

documentation/modules/configuring/con-config-node-pools.adoc
```diff
@@ -0,0 +1,105 @@
...
+
+* `resources` to specify memory and cpu requests and limits
+* `template` to specify custom configuration for pods and other Kubernetes resources
+* `jvmOptions` to specify custom JVM configuration for heap size, runtime and other options 
+
+The `Kafka` resource represents the configuration for all nodes in the Kafka cluster.
+The `KafkaNodePool` resource represents the configuration for nodes only in the node pool.
+Configuration options not specified in the `KafkaNodePool` resource are inherited from the `Kafka` resource.
+Configuration specified in the `KafkaNodePool` resource takes precedence if set in both resources
+For example, if `jvmOptions` are specified in the node pool configuration and Kafka configuration, the node pool configuration is used.
```
scholzj, line 26: Can we go into more details and explain that the options are not combined? To better explain what is covered in this part of the proposal:

> When the optional field is not present in `KafkaNodePool` and is present in `Kafka.spec.kafka`, it will default to the configured value from the `Kafka` CR.
> So when `KafkaNodePool.spec.<field>` exists, `Kafka.spec.kafka.<field>` will be ignored.
> This will be applied only on the highest level as demonstrated in the example below:
> 
> * If the `KafkaNodePool.spec.jvmOptions` does not exist but `Kafka.spec.kafka.jvmOptions` does exist, the values from `Kafka.spec.kafka.jvmOptions` will be used.
> * When `KafkaNodePool.spec.jvmOptions` exists and has `-Xmx: 1024m` and `Kafka.spec.kafka.jvmOptions` exists as well with `-Xms: 512m`, the operator will completely ignore the value from `Kafka.spec.kafka.jvmOptions` and use only the value from `KafkaNodePool.spec.jvmOptions`.
> * When `KafkaNodePool.spec.template` contains only the field `podSet.metadata.labels`, the operator will completely ignore the corresponding values from `Kafka.spec.kafka.template` (for example `podSet.metadata.annotations`, but also for example `pod.metadata.labels`).

documentation/modules/operators/ref-operator-cluster-feature-gates.adoc
```diff
@@ -54,15 +54,23 @@ You have to first downgrade through one of the Strimzi versions in-between, disa
...
-*The `.spec.zookeeper` properties in the Kafka custom resource will be ignored, but still need to be present.*
+*The `.spec.zookeeper` properties in the `Kafka` custom resource are ignored, but still need to be present.*
 The `UseKRaft` feature gate provides an API that configures Kafka cluster nodes and their roles.
 The API is still in development and is expected to change before the KRaft mode is production-ready.
 
+If you are trying Kafka in KRaft mode, you can also enable the `KafkaNodePools` feature gate to manage Kafka node configurations in node pools. 
+
 Currently, the KRaft mode in Strimzi has the following major limitations:
 
 * Moving from Kafka clusters with ZooKeeper to KRaft clusters or the other way around is not supported.
```
scholzj, line 76: We should probably add to the limitations that `controller` only nodes cannot be rolled or updated.

scholzj (commented):
Looks great. I left only some minor comments. Thansk.

documentation/assemblies/deploying/assembly-deploy-kafka-cluster.adoc
```diff
@@ -12,19 +12,27 @@ You can use these files to deploy the Topic Operator and User Operator at the sa
...
 
-* xref:deploying-kafka-cluster-{context}[Kafka cluster]
+* xref:deploying-kafka-cluster-{context}[Kafka cluster] or (preview) xref:deploying-kafka-node-pools-{context}[Kafka cluster with node pools]
 * xref:deploying-the-topic-operator-using-the-cluster-operator-{context}[Topic Operator]
 * xref:deploying-the-user-operator-using-the-cluster-operator-{context}[User Operator]
 
 When installing Kafka, Strimzi also installs a ZooKeeper cluster and adds the necessary configuration to connect Kafka with ZooKeeper.
 
+If you are trying the preview of the node pools feature, you can deploy a Kafka cluster with one or more node pools.
+Node pools provide configuration for a set of kafka nodes.
```
ppatierno, line 22:
```suggestion
Node pools provide configuration for a set of Kafka nodes.
```

documentation/modules/configuring/con-config-node-pools.adoc
```diff
@@ -0,0 +1,105 @@
...
+Optionally, you can also specify values for the following properties:
+
+* `resources` to specify memory and cpu requests and limits
+* `template` to specify custom configuration for pods and other Kubernetes resources
+* `jvmOptions` to specify custom JVM configuration for heap size, runtime and other options 
+
+The `Kafka` resource represents the configuration for all nodes in the Kafka cluster.
+The `KafkaNodePool` resource represents the configuration for nodes only in the node pool.
+Configuration options not specified in the `KafkaNodePool` resource are inherited from the `Kafka` resource.
+Configuration specified in the `KafkaNodePool` resource takes precedence if set in both resources
```
ppatierno, line 25:
```suggestion
Configuration specified in the `KafkaNodePool` resource takes precedence if set in both resources.
```

documentation/modules/configuring/con-config-node-pools.adoc
```diff
@@ -0,0 +1,105 @@
...
+
+NOTE: While the `KafkaNodePools` feature gate that enables node pools is in alpha phase, replica and storage configuration properties in the `KafkaNodePool` resource must also be present in the `Kafka` resource. The configuration in the `Kafka` resource is ignored when node pools are used. Similarly, ZooKeeper configuration properties must also be present in the `Kafka` resource when using KRaft mode. These properties are also ignored.
+
+.Example configuration for a node pool in a cluster using KRaft mode 
+[source,yaml,subs="+attributes"]
+----
+apiVersion: {KafkaNodePoolApiVersion}
+kind: KafkaNodePool
+metadata:
+  name: dual-role # <1>
```
ppatierno, line 44:
```suggestion
  name: kraft-dual-role # <1>
```

ppatierno: Suggested this name change just to make it clear that dual role is about a KRaft example.

documentation/modules/configuring/proc-moving-node-pools.adoc
```diff
@@ -0,0 +1,67 @@
...
+
+.Prerequisites
+
+* xref:deploying-cluster-operator-str[The Cluster Operator must be deployed.]
+
+.Procedure
+
+. Create a new node in the target node pool.
++
+For example, node pool `pool-a` has three replicas. We add a node by increasing the number of replicas:
```
ppatierno, line 26: Can we have a short introduction here? About what we have (pool-a with 3 replicas and pool-b with 4 replicas) and what we want to achieve (moving to pool-a having 4 replicas and scale down pool-b to 3 and moving data).
As it is now, I am going to "discover" what's happening while doing. A short intro could set expectations better while reading imho.

documentation/modules/deploying/proc-deploy-kafka-node-pools.adoc
```diff
@@ -0,0 +1,101 @@
...
+----
++
+.Output shows the node pool names and readiness
+[source,shell,subs="+quotes"]
+----
+NAME                        READY   STATUS    RESTARTS
+my-cluster-entity-operator  3/3     Running   0
+my-cluster-pool-a-kafka-0   1/1     Running   0
+my-cluster-pool-a-kafka-1   1/1     Running   0
+my-cluster-pool-a-kafka-4   1/1     Running   0
```
ppatierno, line 82: Is it really worth showing the output? I mean ... previously you are explaining there are 3 different examples of using node pools but it's not clear to which this output is referring to. We could be more precise and clear or just remove this output?

documentation/modules/operators/ref-operator-cluster-feature-gates.adoc
```diff
@@ -54,15 +54,23 @@ You have to first downgrade through one of the Strimzi versions in-between, disa
 The `UseKRaft` feature gate has a default state of _disabled_.
 
 The `UseKRaft` feature gate deploys the Kafka cluster in the KRaft (Kafka Raft metadata) mode without ZooKeeper.
+ZooKeeper and KRaft are mechanisms used to manage metadata and coordinate operations in Kafka clusters. 
+KRaft mode eliminates the need for an external coordination service like ZooKeeper. 
+In KRaft mode, brokers take on the roles of brokers, controllers, or both.
```
ppatierno, line 59:
```suggestion
In KRaft mode, Kafka nodes take on the roles of brokers, controllers, or both. 
```

documentation/modules/configuring/proc-moving-node-pools.adoc
PaulRMellor: Good idea. I have added a summary to the intro.

documentation/modules/deploying/proc-deploy-kafka-node-pools.adoc
PaulRMellor: We're showing here a couple of things: the additional node is added and running and also we have the note to say that the node IDs might not be sequential. I think this is a useful verification that could anyway be skipped.

documentation/modules/configuring/proc-moving-node-pools.adoc
```diff
@@ -0,0 +1,77 @@
...
+[source,shell,subs="+quotes"]
+----
+kubectl get pods -n _<my_cluster_operator_namespace>_
+----
++
+.Output shows four Kafka nodes in the node pool
+[source,shell,subs="+quotes"]
+----
+NAME                        READY   STATUS    RESTARTS
+my-cluster-entity-operator  3/3     Running   0
```
scholzj, line 54: Should this line be removed? This seems a bit strange as it is right now => it shows some of the pods which will be there but are unrelated (entity operator) and does not show the pods whihc will be there and are relevant (the `pool-b` pods).

I think to make it less confusing, you should consider:
* Listing all pods (EO, Zoo, Pool-a, pool-b) 
* List the pods relevant to the procedure (pool-a and pool-b)
* List only the pool-a pods about which you talk in this particular point

But the current state is a weird mixture between them.

scholzj (commented):
One more nit.

documentation/modules/deploying/proc-deploy-kafka-node-pools.adoc
ppatierno: So .... you are putting this output here after mentioning the above three examples we are providing so I think that should be some relationship between this output and the above examples but ...

- `pool-a` is the name of the pool for the kafka.yaml example only. All the other examples have different pool names.
- deploying the kafka.yaml example you will have 0,1 and 2 as IDs ... not 4

What I mean, if the user just apply the above YAMLs file as you are describing and compare with the output you are showing, they won't match. It could be confusing imho.

joaedwar (approved):

documentation/modules/deploying/proc-deploy-kafka-node-pools.adoc
```diff
@@ -0,0 +1,101 @@
...
+----
++
+* `my-cluster` is the name of the Kafka cluster.
+* `pool-a` is the name of the node pool.
++
+A sequential index number starting with `0` identifies each Kafka pod created.
+If you are using ZooKeeper, you'll also see the ZooKeeper pods.
++
+`READY` shows the number of replicas that are ready/expected.
+The deployment is successful when the `STATUS` shows as `Running`.
```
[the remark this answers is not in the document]
PaulRMellor: Changed all occurrences

PaulRMellor:
@scholzj @ppatierno -- Thanks for the feedback so far.  We just have comments on the `get pods` output. 

We currently show the output of `pool-a`, which has the node added.  I think that might be okay if I remove the operator pod and change the title to _Output shows four Kafka nodes in the ***target*** node pool_. 

And then we can show the output for the scaled down`pool-b` with three nodes later in the procedure and use the title _Output shows three Kafka nodes in the ***source*** node pool_.  How does that sound? 

Regarding Node IDs. I thought it useful to show that they might be out of sequence within a node pool. Isn't that the case? 

NOTE: I wasn't thinking about the deployment examples for this procedure, which are in another part of the guide. I was just using the same naming convention to be consistent.

documentation/modules/deploying/proc-deploy-kafka-node-pools.adoc
PaulRMellor: please see https://github.com/strimzi/strimzi-kafka-operator/pull/8683#issuecomment-1601166058

scholzj:
I'm fine with removing the EO pod and doing hat you suggested.

ppatierno:
@PaulRMellor I am fine with the suggestion

ppatierno (approved):

scholzj (approved):

PR #8696, opened 2023-06-20
Make sure MM2 examples work regardless the `min.insync.replicas` setting on the Kafka cluster

scholzj:
### Type of change

- Bugfix

### Description

In our Mirror Maker 2 examples, we currently set the number of replicas for the Connect part to `-1` to use the default replication factor from the Kafka cluster:

```
      config.storage.replication.factor: -1
      offset.storage.replication.factor: -1
      status.storage.replication.factor: -1
```

But at the same time, we set the `replication.factor` in the connector configurations to simply `1`. So when a user tries the example against the Kafka cluster which has `default.replication.factor: 3` and `min.insync.replicas: 2`, the topics will be created with RF=1 and the minimum number of in-sync replicas of 2. Such a topic will never work.

This PR updates the examples to use `-1` in the connector configurations as well to use the default replication factor of the Kafka cluster. That should make it work even in such situations.

After the release, this should be also updated in the OLM examples which are part of the CSV submission.

### Checklist

### DIFF
--- a/packaging/examples/metrics/kafka-mirror-maker-2-metrics.yaml
+++ b/packaging/examples/metrics/kafka-mirror-maker-2-metrics.yaml
@@ -14,23 +14,27 @@ spec:
     - alias: "my-cluster-target"
       bootstrapServers: my-cluster-target-kafka-bootstrap:9092
       config:
-        config.storage.replication.factor: 1
-        offset.storage.replication.factor: 1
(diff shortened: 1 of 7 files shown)


mimaison (approved):
LGTM

ppatierno (approved):
LGTM

fvaleri (approved):
LGTM. Thanks.

PR #8700, opened 2023-06-20
Add acceptance test profile and trigger them then push to main

kornys:
### Type of change

- Enhancement / new feature

### Description

Added acceptance test profile to a TF and also trigger them when push to main branch happens


Frawless (approved):
We would like to try this (running acceptance tests for every commit into main), in case the tests will not be stable we will revert the changes.

im-konge (approved):
Thanks

PR #8702, opened 2023-06-20
Adding DNS name resolution before ZooKeeper starts

### LINKED ISSUE #8675: ZooKeeper 3.6.4 quorum failing with Kafka 3.4.1
I am using the `main` branch and deploying the operator with `latest` image which includes the support for Kafka 3.4.1 (bringing a new ZooKeeper version 3.6.4).
With minikube (I tried with both an older 1.26.1 and newer 1.30.1 versions), the ZooKeeper ensamble gets stuck, the quorum is not formed, the operator is not able to connect to the ZooKeeper nodes so it cannot proceed with the Kafka pods creation.
The cluster deployment remains in this condition forever.

After long investigation it turned up that during the startup, one or more ZooKeeper pods, are not able to bind the quorum port 3888 to the Cluster IP but during the DNS resolution they get the loopback address instead.
Following a possible log at ZooKeeper startup where you can see the binding at 127.0.0.1:3888 instead of something like 172.17.0.4:3888 (so getting a valid not loopback IP address).

```shell
INFO 3 is accepting connections now, my election bind port: my-cluster-zookeeper-2.my-cluster-zookeeper-nodes.default.svc/127.0.0.1:3888 (org.apache.zookeeper.server.quorum.QuorumCnxManager) [ListenerHandler-my-cluster-zookeeper-2.my-cluster-zookeeper-nodes.default.svc/127.0.0.1:3888]
```

This way, the ZooKeeper node is not accessible from the others for forming a quorum.

It's also important to highlight I was not able to reproduce the issue on OpenShift or AKS instances and it's also not so easy to reproduce on Mac based machine (I am using Fedora).
Anyway the same exact minikube configuration works fine with previous Kafka 3.4.0 (bringing an older ZooKeeper version 3.6.3).

It seems the problem could be related to the newer ZooKeeper version 3.6.4 having a faster startup time when the DNS names registration is not already done at platform level (see minikube) so the name resolution provides the loopback address instead of the Cluster IP we want.

ppatierno:
This PR fixes #8675.
It adds a DNS names resolution check before ZooKeeper starts in order to be sure that ZooKeeper code will be able to resolve hostname and getting right Cluster IP address.
Because of using `nslookup` tool, the PR also adds the `bind-utils` package to the images.
The DNS check can be disabled by setting `ZOOKEEPER_DNS_CHECKS_DISABLED=true` env var, via the `zookeeper.template.zookeeperContainer` field in the `Kafka` custom resource.

### DIFF
--- a/docker-images/base/Dockerfile
+++ b/docker-images/base/Dockerfile
@@ -9,7 +9,7 @@ ARG TARGETARCH
 USER root
 
 RUN microdnf update \
-    && microdnf --setopt=install_weak_deps=0 --setopt=tsflags=nodocs install java-${JAVA_VERSION}-openjdk-headless openssl shadow-utils \
+    && microdnf --setopt=install_weak_deps=0 --setopt=tsflags=nodocs install java-${JAVA_VERSION}-openjdk-headless openssl shadow-utils bind-utils \
     && microdnf reinstall -y tzdata \
     && microdnf clean all
 
--- a/docker-images/base/Dockerfile
+++ b/docker-images/base/Dockerfile
@@ -9,7 +9,7 @@ ARG TARGETARCH
 USER root
 
 RUN microdnf update \
-    && microdnf --setopt=install_weak_deps=0 --setopt=tsflags=nodocs install java-${JAVA_VERSION}-openjdk-headless openssl shadow-utils bind-utils \
+    && microdnf --setopt=install_weak_deps=0 --setopt=tsflags=nodocs install java-${JAVA_VERSION}-openjdk-headless openssl shadow-utils \
     && microdnf reinstall -y tzdata \
     && microdnf clean all
 
--- a/docker-images/kafka-based/kafka/scripts/zookeeper_run.sh
+++ b/docker-images/kafka-based/kafka/scripts/zookeeper_run.sh
@@ -42,12 +42,12 @@ mkdir -p /tmp/zookeeper
 # It is also related to https://issues.apache.org/jira/browse/ZOOKEEPER-4708
 if [ -z "$ZOOKEEPER_DNS_CHECKS_DISABLED" ] || [ "$ZOOKEEPER_DNS_CHECKS_DISABLED" = "false" ]; then
   echo "Trying to resolve ${BASE_HOSTNAME}-$((ZOOKEEPER_ID-1)).${BASE_FQDN} ..."
-  ipaddress=$(nslookup ${BASE_HOSTNAME}-$((ZOOKEEPER_ID-1)).${BASE_FQDN} | grep "Address" | awk '{print $2}' | sed -n 2p)
+  ipaddress=$(nslookup "${BASE_HOSTNAME}"-$((ZOOKEEPER_ID-1))."${BASE_FQDN}" | grep "Address" | awk '{print $2}' | sed -n 2p)
 
-  while [ -z $ipaddress ]; do
+  while [ -z "$ipaddress" ]; do
     sleep 1
     echo "Trying to resolve ${BASE_HOSTNAME}-$((ZOOKEEPER_ID-1)).${BASE_FQDN} ..."
-    ipaddress=$(nslookup ${BASE_HOSTNAME}-$((ZOOKEEPER_ID-1)).${BASE_FQDN} | grep "Address" | awk '{print $2}' | sed -n 2p)
+    ipaddress=$(nslookup "${BASE_HOSTNAME}"-$((ZOOKEEPER_ID-1))."${BASE_FQDN}" | grep "Address" | awk '{print $2}' | sed -n 2p)
   done
 fi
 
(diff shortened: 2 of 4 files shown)


scholzj (commented):
It does not seem to work -> from a quick look at the failing tests and quick try my self, it probably does not work in a single node scenario. (Not 100% if all the tests failing have 1 Zoo only, but looks like it.)

ppatierno:
@scholzj I can confirm that. I am investigating why the 0.0.0.0 binding has a bad effect on a 1 ZK node ensamble.

ppatierno:
I left some updates on the corresponding #8675 issue. We would need to move to a different solution.

docker-images/base/Dockerfile
```diff
@@ -9,7 +9,7 @@ ARG TARGETARCH
 USER root
 
 RUN microdnf update \
-    && microdnf --setopt=install_weak_deps=0 --setopt=tsflags=nodocs install java-${JAVA_VERSION}-openjdk-headless openssl shadow-utils \
+    && microdnf --setopt=install_weak_deps=0 --setopt=tsflags=nodocs install java-${JAVA_VERSION}-openjdk-headless openssl shadow-utils bind-utils \
```
scholzj, line 12: I think you should install this only in the Kafka container image used for ZooKeeepr. Not to all of them.

docker-images/kafka-based/kafka/scripts/zookeeper_run.sh
```diff
@@ -37,6 +37,22 @@ export CERTS_STORE_PASSWORD
...
+  ipaddress=$(nslookup ${BASE_HOSTNAME}-$((ZOOKEEPER_ID-1)).${BASE_FQDN} | grep "Address" | awk '{print $2}' | sed -n 2p)
+
+  while [ -z $ipaddress ]; do
+    sleep 1
+    echo "Resolving ${BASE_HOSTNAME}-$((ZOOKEEPER_ID-1)).${BASE_FQDN} ..."
+    ipaddress=$(nslookup ${BASE_HOSTNAME}-$((ZOOKEEPER_ID-1)).${BASE_FQDN} | grep "Address" | awk '{print $2}' | sed -n 2p)
+  done
+fi
+
+echo "IP address $ipaddress"
```
scholzj, line 54: I think this will make it a bit more understandable for users => that you are not resolving it multiple times but you are tryin to resolve it until it does actually resolve.

```suggestion
if [ -z "$ZOOKEEPER_DNS_CHECKS_DISABLED" ] || [ "$ZOOKEEPER_DNS_CHECKS_DISABLED" = "false" ]; then
  echo "Trying to resolve ${BASE_HOSTNAME}-$((ZOOKEEPER_ID-1)).${BASE_FQDN} ..."
  ipaddress=$(nslookup ${BASE_HOSTNAME}-$((ZOOKEEPER_ID-1)).${BASE_FQDN} | grep "Address" | awk '{print $2}' | sed -n 2p)

  while [ -z $ipaddress ]; do
    sleep 1
    echo "Trying to resolve ${BASE_HOSTNAME}-$((ZOOKEEPER_ID-1)).${BASE_FQDN} ..."
    ipaddress=$(nslookup ${BASE_HOSTNAME}-$((ZOOKEEPER_ID-1)).${BASE_FQDN} | grep "Address" | awk '{print $2}' | sed -n 2p)
  done
fi

echo "Resolved IP address $ipaddress"
```

ppatierno:
@scholzj I changed the PR name and the description with the new solution after discussing on #8675. Can you have a new pass please?

docker-images/base/Dockerfile
ppatierno: Yeah good point, agree.

scholzj (approved):

kornys:
From testing-farm results we can see that issue is fixed on minikube.

Frawless:
I was looking into the failures and `FeatureGatesIsolatedST.testSwitchingConnectStabilityIdentifiesFeatureGateOnAndOff` looks suspicious to me. It failed on Azure and on TF Arm as well (TF x86 passed).

ppatierno:
@im-konge said that these failing tests are well known and not related to the fix in this PR.
I also ran the failed tests locally (still on minikube where I was facing the ZooKeeper issue of course) and they worked fine.
I would assume we can merge this or waiting for some more tests? Do we need more than regression @scholzj ?

scholzj:
I think you should at least restart it and get the tests green.

see-quick:
I have restarted the failed pipelines :).

ppatierno:
> I think you should at least restart it and get the tests green.

Hoping they are just flaky :-) 

> I have restarted the failed pipelines :).

Thanks Maros!

im-konge:
@ppatierno #8716 will fix the FG ST, the SecurityST fail is still something we need to check and fix (JFYI)

PR #8709, opened 2023-06-20
Dashboards can now select datasource dynamically

jcpunk:
### Type of change

_Select the type of your PR_

- Enhancement / new feature
- Refactoring

### Description

Make the grafana dashboard datasource a select-able field of the dashboard itself.
This should provide a much better overall experience for folks as they can now have multiple prometheus instances for different clusters and use a single set of dashboards to track them all.

### Checklist

- [x] Write tests
- [x] Update documentation
- [ ] Update CHANGELOG.md

### Note

The dashboards were all imported into grafana 7.3.7 (the most common version listed in their plugins set) and I left the extra data they wanted to put in (mostly null option sets).  Some of the dashboards had other (even newer versions), but I figured some level of normalization was a good thing.

Before merging I'd like thoughts from @tjanson due to: https://github.com/strimzi/strimzi-kafka-operator/issues/8545 and https://github.com/strimzi/strimzi-kafka-operator/pull/8664

### DIFF
--- a/packaging/examples/metrics/grafana-dashboards/strimzi-cruise-control.json
+++ b/packaging/examples/metrics/grafana-dashboards/strimzi-cruise-control.json
@@ -1,14 +1,4 @@
 {
-  "__inputs": [
-    {
-      "name": "DS_PROMETHEUS",
-      "label": "Prometheus",
-      "description": "",
-      "type": "datasource",
-      "pluginId": "prometheus",
-      "pluginName": "Prometheus"
-    }
-  ],
   "__requires": [
     {
       "type": "grafana",
@@ -52,7 +42,7 @@
   "gnetId": null,
   "graphTooltip": 0,
   "id": null,
-  "iteration": 1596037257491,
+  "iteration": 1687297471450,
   "links": [],
   "panels": [
     {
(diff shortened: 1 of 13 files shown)


scholzj (approved):
I guess this looks good to me. Thanks for the PR. Can you please fix the DCO sign-off?

tjanson:
@jcpunk I did not test this from a 100% clean slate, but I removed the dashboards and regenerated them from your fork with no issues, i.e., it worked out of the box for me. (My default and only `prometheus` datasource is called `Prometheus`.)

ppatierno:
@jcpunk I see a lot of changes which are not related to the PR goal. Are they needed? Are they related to the usage of a newer Grafana version?

ppatierno:
In general I try to reduce the changes to dashboards just focusing on what's really needed.
For example, here I see a lot of `"fieldConfig"` added and they are pretty empty. Maybe we could remove them to avoid the noise. I also see a "pluginVersion" downgraded, why?

jcpunk:
Some of the dashboards look like they were generated in a newer version of grafana.  I did manual imports/exports all from the same grafana container.  I'd be open to the idea of using a newer grafana as the minimum requirements, but I figured that was a subject for a different PR.

The various empty field configs all got populated automatically by grafana.  I suspect they may be unnecessary.  But given that grafana wanted them there for its export, I'm hesitant to remove them.

Frawless (approved):

ppatierno (approved):

scholzj:
Thanks for the PR @jcpunk

PR #8718, opened 2023-06-22
Add validation for cluster IDs between the Kafka CR and KafkaNodePool CRs

### LINKED ISSUE #8605: [NodePools] Validate Kafka cluster ID before Kafka reconciliation
One of the things not delivered in the initial Kafka Node Pools PR #8551 is the validation of the Kafka cluster IDs. The idea is that as part of the node pool validation, we will validate that the Kafka cluster-ID in the `Kafka` CR and in the `KafkaNodePool` CRs is identical (unless it is a new cluster/pool and it is missing of course). This should help to prevent situations, where due to a change of the `strimzi.io/cluster` label, the node pool would be highjacked from one Kafka cluster to another.

scholzj:
### Type of change

- Enhancement / new feature

### Description

This PR implements the validation of the Apache Kafka cluster IDs as included in the [StrimziProposal#50](https://github.com/strimzi/proposals/blob/main/050-Kafka-Node-Pools.md). 

To be able to properly validate the cluster IDs, it has to make sure the cluster-ID is stored properly in the status of the Kafka CR across different situations -> including when the reconciliation fails or when the reconciliation is paused. Before this PR, the cluster-ID was omitted from the status in these situations and it was not possible to validate it properly.

The validation also distinguishes between validation for KRaft- and ZooKeeper-based clusters. This is because, for KRaft clusters, we set the cluster-ID ourselves. But for ZooKeeper-based clusters, it is generated by Kafka (setting it ourselves would be too complicated as we would need to create it in ZooKeeper before starting the Kafka brokers).

So for ZooKeeper-based clusters, we make sure that:
* When the cluster-ID is not set in the Kafka CR, it is also not set in the KafkaNodePool CRs
* When the IDs are set, they are always the same

For KRaft-based clusters, we make sure that:
* The IDs set in the various resources are the same
* If the ID is set at least in one place, it is used as the cluster ID for the cluster
* If the ID is not set anywhere, new ID is generated

This PR closes #8605.

### Checklist

- [x] Write tests

### DIFF
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/model/nodepools/NodePoolUtils.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/model/nodepools/NodePoolUtils.java
@@ -13,6 +13,8 @@ import io.strimzi.operator.cluster.model.KafkaPool;
 import io.strimzi.operator.cluster.model.ModelUtils;
 import io.strimzi.operator.common.Reconciliation;
 import io.strimzi.operator.cluster.operator.resource.SharedEnvironmentProvider;
+import io.strimzi.operator.common.ReconciliationLogger;
+import org.apache.kafka.common.Uuid;
 
 import java.util.ArrayList;
 import java.util.List;
@@ -22,6 +24,8 @@ import java.util.Map;
  * Class which contains different utility methods for working with Node Pools
  */
 public class NodePoolUtils {
+    private static final ReconciliationLogger LOGGER = ReconciliationLogger.create(NodePoolUtils.class.getName());
+
     /**
      * Generates KafkaPool instances from Kafka and KafkaNodePool resources.
      *
@@ -80,7 +84,7 @@ public class NodePoolUtils {
                     )
             );
         } else {
-            validateNodePools(kafka, nodePools, useKRaft);
+            validateNodePools(reconciliation, kafka, nodePools, useKRaft);
 
             // We prepare ID Assignment
             NodeIdAssignor assignor = new NodeIdAssignor(nodePools);
@@ -107,11 +111,12 @@ public class NodePoolUtils {
     /**
      * Validates KafkaNodePools
      *
-     * @param kafka         The Kafka custom resource
-     * @param nodePools     The list with KafkaNodePool resources
-     * @param useKRaft      Flag indicating whether KRaft is enabled or not
+     * @param reconciliation    Reconciliation marker
+     * @param kafka             The Kafka custom resource
+     * @param nodePools         The list with KafkaNodePool resources
+     * @param useKRaft          Flag indicating whether KRaft is enabled or not
      */
-    public static void validateNodePools(Kafka kafka, List<KafkaNodePool> nodePools, boolean useKRaft)    {
+    public static void validateNodePools(Reconciliation reconciliation, Kafka kafka, List<KafkaNodePool> nodePools, boolean useKRaft)    {
         // If there are no node pools, the rest of the validation makes no sense, so we throw an exception right away
         if (nodePools.isEmpty()
                 || nodePools.stream().noneMatch(np -> np.getSpec().getReplicas() > 0))    {
@@ -121,9 +126,17 @@ public class NodePoolUtils {
             List<String> errors = new ArrayList<>();
 
             if (useKRaft) {
+                // Validate process roles
                 errors.addAll(validateKRaftProcessRoles(nodePools));
+
+                // Validate cluster IDs
+                errors.addAll(validateKRaftClusterIds(reconciliation, kafka, nodePools));
             } else {
+                // Validate process roles
                 errors.addAll(validateZooKeeperBasedProcessRoles(nodePools));
+
+                // Validate cluster IDs
+                errors.addAll(validateZooKeeperBasedClusterIds(reconciliation, kafka, nodePools));
             }
 
             // Throw an exception if there are any errors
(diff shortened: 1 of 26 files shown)


cluster-operator/src/main/java/io/strimzi/operator/cluster/model/nodepools/NodePoolUtils.java
```diff
@@ -198,4 +211,140 @@ private static List<String> validateKRaftProcessRoles(List<KafkaNodePool> nodePo
...
+        //     * No cluster ID regardless whether the Kafka CR has some cluster ID or not (e.g. with new node pools)
+        //     * The same cluster ID as the Kafka CR if the Kafka CR has a cluster ID already
+        //
+        // But they are not allowed to have any cluster ID when the Kafka CR does not have one yet.
+        for (KafkaNodePool pool : pools)   {
+            if (pool.getStatus() != null && pool.getStatus().getClusterId() != null)   {
+                if (kafkaCrClusterId == null)   {
+                    hasClusterIdTooEarly = true;
+                    LOGGER.warnCr(reconciliation, "Cluster ID {} found in KafkaNodePool {} is not allowed without the same cluster ID being present in the Kafka custom resource", pool.getStatus().getClusterId(), pool.getMetadata().getName());
+                    //kafkaCrClusterId = pool.getStatus().getClusterId();
```
ppatierno, line 248: leftover?

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/nodepools/NodePoolUtils.java
```diff
@@ -198,4 +211,140 @@ private static List<String> validateKRaftProcessRoles(List<KafkaNodePool> nodePo
...
+        // For ZooKeeper based clusters, the Node Pools are allowed to have:
+        //     * No cluster ID regardless whether the Kafka CR has some cluster ID or not (e.g. with new node pools)
+        //     * The same cluster ID as the Kafka CR if the Kafka CR has a cluster ID already
+        //
+        // But they are not allowed to have any cluster ID when the Kafka CR does not have one yet.
+        for (KafkaNodePool pool : pools)   {
+            if (pool.getStatus() != null && pool.getStatus().getClusterId() != null)   {
+                if (kafkaCrClusterId == null)   {
+                    hasClusterIdTooEarly = true;
+                    LOGGER.warnCr(reconciliation, "Cluster ID {} found in KafkaNodePool {} is not allowed without the same cluster ID being present in the Kafka custom resource", pool.getStatus().getClusterId(), pool.getMetadata().getName());
```
ppatierno, line 247: I would add the name of the Kafka CR as well, to have a single warning line to understand from which CRs the cluster-ids are got and compared.

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/nodepools/NodePoolUtils.java
```diff
@@ -198,4 +211,140 @@ private static List<String> validateKRaftProcessRoles(List<KafkaNodePool> nodePo
...
+                    //kafkaCrClusterId = pool.getStatus().getClusterId();
+                } else if (!kafkaCrClusterId.equals(pool.getStatus().getClusterId())) {
+                    hasConflict = true;
+                    LOGGER.warnCr(reconciliation, "Cluster ID {} found in KafkaNodePool {} is not the same as the cluster ID {} found in the Kafka custom resource", pool.getStatus().getClusterId(), pool.getMetadata().getName(), kafkaCrClusterId);
```
ppatierno, line 251: Ditto as above.

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/nodepools/NodePoolUtils.java
```diff
@@ -198,4 +211,140 @@ private static List<String> validateKRaftProcessRoles(List<KafkaNodePool> nodePo
...
+            return List.of("The KafkaNodePool resources should not have cluster ID set before the Kafka custom resource.");
+        } else {
+            return List.of();
+        }
+    }
+
+    /**
+     * Extracts the cluster IDs found in the Kafka and KafkaNodePool custom resources and validates whether the
+     * resources do not belong to different Kafka cluster. This method is used for KRaft based Kafka clusters. In Kraft
+     * based Kafka clusters, we set the cluster ID our-self. So it might happen that the KafkaNOdePool status has the
```
ppatierno, line 268:
```suggestion
     * based Kafka clusters, we set the cluster ID our-self. So it might happen that the KafkaNodePool status has the
```

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaAssemblyOperator.java
```diff
@@ -126,6 +126,15 @@ public Future<KafkaStatus> createOrUpdate(Reconciliation reconciliation, Kafka k
                 status.setObservedGeneration(kafkaAssembly.getMetadata().getGeneration());
             }
 
+            if (status.getClusterId() == null
+                    && kafkaAssembly.getStatus() != null
+                    && kafkaAssembly.getStatus().getClusterId() != null)  {
+                // If not set in the status prepared by reconciliation but set int he status previously, we copy the
```
ppatierno, line 132:
```suggestion
                // If not set in the status prepared by reconciliation but set in the status previously, we copy the
```

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/nodepools/NodePoolUtils.java
scholzj: @ppatierno I'm not sure that is needed. It has the name of the Kafka CR in the reconciliation marker. WDYT?

ppatierno: Ah right, ok no need to add more.

ppatierno (approved):

PR #8720, opened 2023-06-23
Fix tmt plan stage naming due to new version of testing-farm

kornys:
### Type of change

- Bugfix

### Description

After new version of TF release with new TMT we hit a regression where naming of steps in tmt plan like `foo/bar`is not allowed. This fix remove `/` from step name.


Frawless (approved):

im-konge (approved):
LGTM, assuming it works :) Thanks

PR #8722, opened 2023-06-23
[systemtest] Enable running tests with KafkaNodePools in FG and KRaft modes

im-konge:
### Type of change

- Enhancement

### Description

This PR enables the possibility to run STs with KafkaNodePools - when it is enabled.
The logic behind this is following:

1. If we run tests without KafkaNodePools, the scenario remains the same - we deploy Kafka resource and wait for readiness,...
2. In case that we run tests with KafkaNodePools enabled, in the moment when we are deploying Kafka, the KafkaNodePool with `BROKER` role is created, with all the settings around replicas and storage from Kafka resource (Kafka resource remains the same)
3. In case that we run tests with KafkaNodePools and KRaft enabled, the scenario is the same as in 2., but we add `CONTROLLER` role to the KafkaNodePools - so we have dual role KafkaNodePool

To enable this option I had to edit our `ResourceManager` and especially to `createResource`, which I split into two extra methods for labeling the resource and also setting the namespace if needed + I added the method for creating the KafkaNodePool.

As part of this PR I also added constant to `Annotations.java` for the NodePools annotation.

### Checklist


henryZrncik (approved):

operator-common/src/main/java/io/strimzi/operator/common/Annotations.java
```diff
@@ -150,6 +150,11 @@ public class Annotations {
      */
     public static final String ANNO_DEP_KUBE_IO_REVISION = "deployment.kubernetes.io/revision";
 
+    /**
+     * Annotation for enabling/disabling usage of KafkaNodePools in Kafka CR
+     */
+    public static final String ANNO_STRIMZI_IO_NODE_POOLS = STRIMZI_DOMAIN + "node-pools";
```
scholzj, line 156: You should not add a new annotation. Assuming you need it in the `operator-common` module, you should move here the existing annotation.

scholzj (changes requested):

systemtest/src/main/java/io/strimzi/systemtest/resources/ResourceManager.java
```diff
@@ -254,6 +230,99 @@ public final <T extends HasMetadata> void createResource(ExtensionContext testCo
...
+                    labels.put(Constants.TEST_SUITE_NAME_LABEL, testSuiteName);
+                }
+                resource.getMetadata().setLabels(labels);
+            }
+        }
+    }
+
+    private <T extends HasMetadata> void setNamespaceInResource(ExtensionContext testContext, T resource) {
+        // ignore test context of shared Cluster Operator
+        if (testContext != BeforeAllOnce.getSharedExtensionContext()) {
```
see-quick, line 314: This could be removed because we no longer use shared CO.

scholzj (approved):

im-konge:
@scholzj @Frawless @see-quick this is ready for review now

In the pipelines, there are 2 failures -> one for `HelmChartIsolatedST` (should be fixed by #8773) and `CruiseControlST.testCruiseControlWithSingleNodeKafka` (should be fixed by #8776)

systemtest/src/main/java/io/strimzi/systemtest/resources/ResourceManager.java
```diff
@@ -97,6 +101,7 @@ public class ResourceManager {
 
     private static String coDeploymentName = Constants.STRIMZI_DEPLOYMENT_NAME;
     private static ResourceManager instance;
+    private static final Random RANDOM = new Random();
```
see-quick, line 104: Is this somewhere used?

see-quick (approved):
LGTM, just one nit. Thanks 👍

Frawless (approved):

systemtest/src/main/java/io/strimzi/systemtest/Environment.java
```diff
@@ -264,6 +264,10 @@ public static boolean isKRaftModeEnabled() {
         return STRIMZI_FEATURE_GATES.contains(Constants.USE_KRAFT_MODE);
     }
 
+    public static boolean isKafkaNodePoolEnabled() {
```
scholzj, line 267: Should it be `isKafkaNodePoolsEnabled`? Both the feature gate and your constant use plural.

operator-common/src/main/java/io/strimzi/operator/common/Annotations.java
```diff
@@ -160,6 +160,11 @@ public class Annotations {
      */
     public static final String ANNO_DEP_KUBE_IO_REVISION = "deployment.kubernetes.io/revision";
 
+    /**
+     * Annotation for enabling or disabling the Node Pools. This annotation is used on the Kafka CR
+     */
+    public static final String ANNO_STRIMZI_IO_NODE_POOLS = STRIMZI_DOMAIN + "node-pools";
```
scholzj, line 166: Maybe you can sort it a bit add it after the other strimzi.io annotations?

scholzj (commented):
I left some nits. But to be honest, as a someone who is not a system test SME, I do not really understand what are all the things this does. Does thi sdo only some preparation? Or actually affect the tests we run? If so, can you add a more clear description of what the tests will test now which would be more understandable even for people not so familiar with the STs? E.g. what will now be used in the KRaft regression and feature gates regression.

im-konge:
@scholzj it resolves issues when running FG and KRaft pipelines (currently just FG, I want to add KafkaNodePools to KRaft pipelines in different PR) - it will really use KafkaNodePools, it fixes some checks inside the tests, so it will work with different names etc.

I will add it to the description of the PR. Originally this PR was about adding "basic support" of KafkaNodePools to the STs, but it ended up as a PR which makes the tests work with KafkaNodePools.

Currently, if the `KafkaNodePools` FG is enabled, the Kafka resource will be deployed with the annotation and the KafkaNodePools will be created too (just before the Kafka resource, as it should be done). In the future PRs, we should add more specific tests for KafkaNodePools.

scholzj (approved):

PR #8725, opened 2023-06-24
Add warning conditions tot he Kafka CR if PVCs cannot be resized because of Storage Class

scholzj:
### Type of change

- Enhancement / new feature

### Description

When a user tries to resize the storage but it is not supported due to a storage class (e.g. missing Storage Class or Storage Class which does not support resizing), we issue warnings into logs. But those might be easy to miss for the users. So this PR adds in addition warning conditions to the Kafka CR:

```
    - lastTransitionTime: "2023-06-24T14:49:35.098095357Z"
      message: Storage Class standard does not support resizing of volumes. PVC data-0-my-cluster-kafka-0
        cannot be resized.
      reason: PvcResizingWarning
      status: "True"
      type: Warning
```

This should help to make sure users are aware of the issue.

### Checklist

- [x] Write tests


ppatierno (approved):
LGTM

PR #8731, opened 2023-06-26
Fix nslookup resolving ZooKeeper pod IP address as 127.0.0.1

ppatierno:
After adding the DNS name resolution check on the ZooKeeper pods via `nslookup`, I noticed that it can return `127.0.0.1` as an early resolution. It takes more time to get a "valid" Pod IP.
This would unlock the ZooKeeper application to start but then trying to bind the quorum port to 127.0.0.1 which, of course, makes the ensamble not working.
This PR adds an additional check that during `nslookup` DNS name resolution, the IP address has to be different from `127.0.0.1`.


scholzj:
There is no _Cluster IP_ here. The DNS name resolves to a _Pod IP_. You should probably fix it in the description to make it more clear.

ppatierno:
Makes sense, just fixed.

scholzj (approved):
I guess I'm fine if you want to merge it 🤷. But it does not seem to make any change in my environment - to better or to worse.

fvaleri (approved):
LGTM.

I successfully tested this workaround after rebasing on the same Minikube instance that originally had the issue. I also ran the full unit and integration test suite on it. Now the ZK hostname resolution takes 30 seconds by default on Minikube, unless you reduce the coredns cache timeout.

ppatierno:
Yeah, checking for DNS ready to resolve hostnames adds some delays. On minikube it's related to the cache configuration for coredns which is 30 secs by default. Lowering that value allows the check to be successful faster.
It's also faster on environment like OpenShift or AKS where I was able to try. Anyway, this delay should help ZooKeeper to avoid itself delaying/crashing because DNS names are not resolved yet.

PR #8733, opened 2023-06-26
[ST] Remove unspecified kafkaTopicTemplate methods

jankalinic:
### Type of change
- Refactoring

### Description

This PR should fix problems with non-existent namespace being used for kafkaTopic creation. The problem appears while parallel execution. The resource manager uses last namespace (from prev test), which is deleted after the test. So when tests like testKafkaTopicDifferentStates start with topic creation, it fails.

### Checklist

- [ ] Write tests
- [ ] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/systemtest/src/main/java/io/strimzi/systemtest/templates/crd/KafkaTopicTemplates.java
+++ b/systemtest/src/main/java/io/strimzi/systemtest/templates/crd/KafkaTopicTemplates.java
@@ -8,31 +8,27 @@ import io.strimzi.api.kafka.model.KafkaTopic;
 import io.strimzi.api.kafka.model.KafkaTopicBuilder;
 import io.strimzi.operator.common.model.Labels;
 import io.strimzi.systemtest.Constants;
-import io.strimzi.systemtest.resources.ResourceManager;
+import io.strimzi.systemtest.storage.TestStorage;
 import io.strimzi.test.TestUtils;
 
 public class KafkaTopicTemplates {
 
     private KafkaTopicTemplates() {}
 
-    public static KafkaTopicBuilder topic(String clusterName, String topicName) {
-        return defaultTopic(clusterName, topicName, 1, 1, 1, ResourceManager.kubeClient().getNamespace());
+    public static KafkaTopicBuilder topic(TestStorage testStorage) {
+        return defaultTopic(testStorage.getClusterName(), testStorage.getTopicName(), 1, 1, 1, testStorage.getNamespaceName());
     }
 
     public static KafkaTopicBuilder topic(String clusterName, String topicName, String topicNamespace) {
         return defaultTopic(clusterName, topicName, 1, 1, 1, topicNamespace);
     }
 
-    public static KafkaTopicBuilder topic(String clusterName, String topicName, int partitions) {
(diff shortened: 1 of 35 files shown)


jankalinic:
@strimzi-ci run tests --profile=regression --install-type=bundle --ocp-version=4.13 --env=ST_KAFKA_VERSION=3.4.0

im-konge (approved):
LGTM, assuming it works

systemtest/src/main/java/io/strimzi/systemtest/templates/crd/KafkaTopicTemplates.java
```diff
@@ -8,31 +8,27 @@
...
     }
 
-    public static KafkaTopicBuilder topic(String clusterName, String topicName, int partitions, int replicas) {
-        return defaultTopic(clusterName, topicName, partitions, replicas, replicas, ResourceManager.kubeClient().getNamespace());
-    }
-
-    public static KafkaTopicBuilder topic(String clusterName, String topicName, int partitions, int replicas, int minIsr) {
-        return defaultTopic(clusterName, topicName, partitions, replicas, minIsr, ResourceManager.kubeClient().getNamespace());
+    public static KafkaTopicBuilder topic(String clusterName, String topicName, int partitions, int replicas, String topicNamespace) {
+        return defaultTopic(clusterName, topicName, partitions, replicas, replicas, topicNamespace);
```
Frawless, line 31: Uniting the clusterName, Name, and namespaceName parameters positions would make sense across all resources. Not strictly as part of this PR.

systemtest/src/test/java/io/strimzi/systemtest/kafka/KafkaST.java
```diff
@@ -542,7 +542,7 @@ void testLabelsExistenceAndManipulation(ExtensionContext extensionContext) {
             .endSpec()
             .build());
 
-        resourceManager.createResource(extensionContext, KafkaTopicTemplates.topic(testStorage.getClusterName(), testStorage.getTopicName()).build());
+        resourceManager.createResource(extensionContext, KafkaTopicTemplates.topic(testStorage).build());
```
Frawless, line 545: I am not sure if I like this approach. It makes the code a little bit mysterious as you pass there only testStorage but it is my opinion.

systemtest/src/main/java/io/strimzi/systemtest/templates/crd/KafkaTopicTemplates.java
jankalinic: Will surely do in another PR. Thanks for the tip!

systemtest/src/test/java/io/strimzi/systemtest/kafka/KafkaST.java
jankalinic: We already do this with kafkaUsers, but i agree it's not instantly telling.

jankalinic:
@strimzi-ci run tests --install-type=bundle --ocp-version=4.13 --testcase=TopicST#testKafkaTopicDifferentStates

PR #8738, opened 2023-06-27
Log a warning when a KafkaTopic has no spec

### LINKED ISSUE #8465: `KafkaTopic` without `.spec` section should not be silently ignored
### Discussed in https://github.com/orgs/strimzi/discussions/8463

----

It looks like in the `KafkaTopic` CR, the `.spec` section is not required and users can create the resource without it. However, the operator seems to check it in the `K8sTopicWatcher` class and when it is missing, it silently ignores it. I think it is fine to not process such a resource, but we should at least throw a warning and not skip such a resource silently.

mabulgu:
### Type of change

Bugfix

Fixes #8465 

### Description

Please see the related issue: https://github.com/strimzi/strimzi-kafka-operator/issues/8465
Also related discussion: https://github.com/orgs/strimzi/discussions/8463

### Checklist

- [x] Write tests
~~- [ ] Update documentation~~
~~- [ ] Check RBAC rights for Kubernetes / OpenShift roles~~
- [x] Update CHANGELOG.md
~~- [ ] Supply screenshots for visual changes, such as Grafana dashboards~~

### Notes

**NOTE TO REVIEWER:** 
It is not possible to handle the log output in the test context to validate without any problems, so I did the validation manually. I found a few ways which made me create a WriterAppender to write the logs to a Writer, but the maven verification complained this time by saying "dependency problems found" without any detail. This worked, but it could not pass the verification of maven-dependency-plugin

Here is that commit if you are interested, and further discussion if needed: https://github.com/strimzi/strimzi-kafka-operator/pull/8738/commits/8eb23803dcfd2f5867a179360679fa374ac778c2

Here is the complaining maven verification log: https://dev.azure.com/cncf/strimzi/_build/results?buildId=141431&view=logs&j=78df3800-942a-5b94-093f-0e36f8f8252f&t=1166645e-b435-5805-cc4c-2efe8ac57bb6&l=6323

### DIFF
--- a/topic-operator/src/main/java/io/strimzi/operator/topic/K8sTopicWatcher.java
+++ b/topic-operator/src/main/java/io/strimzi/operator/topic/K8sTopicWatcher.java
@@ -49,8 +49,8 @@ class K8sTopicWatcher implements Watcher<KafkaTopic> {
     public void eventReceived(Action action, KafkaTopic kafkaTopic) {
         ObjectMeta metadata = kafkaTopic.getMetadata();
         Map<String, String> labels = metadata.getLabels();
+        LogContext logContext = LogContext.kubeWatch(action, kafkaTopic).withKubeTopic(kafkaTopic);
         if (kafkaTopic.getSpec() != null) {
-            LogContext logContext = LogContext.kubeWatch(action, kafkaTopic).withKubeTopic(kafkaTopic);
             String name = metadata.getName();
             String kind = kafkaTopic.getKind();
             if (!initReconcileFuture.isComplete()) {
@@ -91,6 +91,8 @@ class K8sTopicWatcher implements Watcher<KafkaTopic> {
                     LOGGER.debugCr(logContext.toReconciliation(), "Ignoring {} to {} {} because metadata.generation==status.observedGeneration", action, kind, name);
                 }
             }
+        } else {
+            LOGGER.warnCr(logContext.toReconciliation(), "Topic {}/{} has no spec", kafkaTopic.getMetadata().getNamespace(), kafkaTopic.getMetadata().getName());
         }
     }
 
--- a/topic-operator/src/main/java/io/strimzi/operator/topic/K8sTopicWatcher.java
+++ b/topic-operator/src/main/java/io/strimzi/operator/topic/K8sTopicWatcher.java
@@ -92,7 +92,7 @@ class K8sTopicWatcher implements Watcher<KafkaTopic> {
                 }
             }
         } else {
-            LOGGER.warnCr(logContext.toReconciliation(), "Topic {}/{} has no spec", kafkaTopic.getMetadata().getNamespace(), kafkaTopic.getMetadata().getName());
+            LOGGER.warnCr(logContext.toReconciliation(), "Topic has no spec");
         }
     }
 
--- a/topic-operator/src/test/java/io/strimzi/operator/topic/TopicOperatorTest.java
+++ b/topic-operator/src/test/java/io/strimzi/operator/topic/TopicOperatorTest.java
@@ -151,6 +151,22 @@ public class TopicOperatorTest {
         context.completeNow();
     }
 
+    /** Test what happens when a KafkaTopic with no spec gets created in kubernetes */
+    @Test
+    public void testOnKafkaTopicAdded_nospec(VertxTestContext context) {
+        KafkaTopic kafkaTopic = new KafkaTopicBuilder()
+                .withMetadata(new ObjectMetaBuilder().withName("nospec")
+                        .withNamespace("my-namespace")
+                        .withLabels(labels.labels()).build())
+                .build();
+        K8sTopicWatcher watcher = new K8sTopicWatcher(topicOperator, Future.succeededFuture(), () -> { });
+        watcher.eventReceived(ADDED, kafkaTopic);
+
+        mockKafka.assertEmpty(context);
+        mockTopicStore.assertEmpty(context);
+        context.completeNow();
+    }
+
     /** Test what happens when a non-topic KafkaTopic gets created in kubernetes */
     @Test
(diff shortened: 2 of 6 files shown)


topic-operator/src/main/java/io/strimzi/operator/topic/K8sTopicWatcher.java
```diff
@@ -91,6 +91,8 @@ public void eventReceived(Action action, KafkaTopic kafkaTopic) {
                     LOGGER.debugCr(logContext.toReconciliation(), "Ignoring {} to {} {} because metadata.generation==status.observedGeneration", action, kind, name);
                 }
             }
+        } else {
+            LOGGER.warnCr(logContext.toReconciliation(), "Topic {}/{} has no spec", kafkaTopic.getMetadata().getNamespace(), kafkaTopic.getMetadata().getName());
```
ppatierno, line 95: shouldn't the reconciliation marker already bring namespace/name so we don't need it logged in the message?

topic-operator/src/test/java/io/strimzi/operator/topic/TopicOperatorTest.java
```diff
@@ -151,6 +151,22 @@ public void testOnKafkaTopicAdded_ignorable(VertxTestContext context) {
         context.completeNow();
     }
 
+    /** Test what happens when a KafkaTopic with no spec gets created in kubernetes */
+    @Test
+    public void testOnKafkaTopicAdded_nospec(VertxTestContext context) {
```
ppatierno, line 156:
```suggestion
    public void testOnKafkaTopicAddedNoSpec(VertxTestContext context) {
```

let's use camelcase ...

mabulgu: Using underscore is also a common format in unit and/or bdd testing. Esp. it is very common in tests formatted as given-when-then (a quick example here: https://medium.com/@stefanovskyi/unit-test-naming-conventions-dd9208eadbea).

I just followed the convention which is already used in the methods here.

I will reword the "nospec" to "noSpec" but wanna leave the _ as the other methods also have it, if you don't mind.

topic-operator/src/main/java/io/strimzi/operator/topic/K8sTopicWatcher.java
mabulgu: Oh let me check. I've seen sth similar used so added the same here. If it is redundant, I will get rid of it.

mabulgu: Updated! Good catch, thanks!

topic-operator/src/test/java/io/strimzi/operator/topic/TopicOperatorTest.java
mabulgu: When the method names include test-specific states (such as given-when-then, or simply before-after states), they might get longer so in order to make readability better it is common to use underscores along with the camel style in Java.

In languages like Python, since _ usage is common, they use two underscores (__) instead. (test_on_kafka_topic_added__no_spec). So the _ which is used in test methods to separate the behavioral situations is sth language independent. It is the space conversions that vary from language to language. Java use camel case, Python uses snake case, and some use kabab case (I guess it was Lisp lang IIRC).

ppatierno: Tbh I have seen this in other languages (as you said in Python) but not with Java. Anyway I am not going to "block" this PR just for this :-) if the others are ok with the naming.

ppatierno: Tbh I have seen this in other languages (as you said in Python) but not with Java. Anyway I am not going to "block" this PR just for this :-) if the others are ok with the naming.

mabulgu: I wanted to make them consistent within the test class. As you might notice, the other methods are named just with camel case. 

Since both methods are not that long and doesnt include a huge scenario normally I would start with a normal convention (like the last one you mention here), and when the scenarios get more complex, I would start using _ for the step state separation.

I just wanted to emphasize that using underscore is sth common even in Java, but in the cases of using scenarios. Here we have a one-two step scenario which doesn't make much difference to use the generic convention or the behavioral-test special one.

If you don't mind, we can reach out to Tom later on and ask him about why he preferred to use that convention in one test class, but not the other one as both seem to include a state change (thus a +1 step scenario)

mabulgu: BTW for the overall style thing, we can discuss and create a subset of team agreement document (I say subset because team agreement docs not only have techie bits but everything related to teams/groups) that would fit the contributor guide of this project so that anyone who contributes would follow the guide. I will try to propose it in one of the community meetings (I guess that would be the right place)

ppatierno: I am not sure it's worth doing the effort on an agreement doc about code style. I was just asking why the difference. Knowing the reason from @tombentley would be enough for me :-)

mabulgu: What I mean is a kinda social contract that includes the small technical bits, which might help resolve conflicts easier (not for just style-wise concerns of course). I believe it is worth doing it, as it is not just for code style but for general consistency.

With all respect, leaving a link here for further reading as anyone might be interested in the topic as well: https://openpracticelibrary.com/practice/social-contract/

tombentley (approved):
LGTM, thanks!

ppatierno (approved):

PR #8741, opened 2023-06-28
Make the KafkaRebalance spec changes reflect in the status when it is updated

### LINKED ISSUE #8104: Update `KafkaRebalance` status if spec is changed
Once a `KafkaRebalance` resource is created and the status is set by the Cluster Operator, the status will never be updated again until the `KafkaRebalance` resource is "refreshed". This can lead to confusing UX when a user updates the `KafkaRebalance` spec and those changes are reflected in the `KafkaRebalance` spec but not in the `KafkaRebalance` status. The `KafkaRebalance` status will remain the same even if the `KafkaRebalance`  spec is changed, possibly leading users to think the status is related to the latest spec.

We should have the Cluster Operator (or KafkaRebalanceOperator) automatically "refresh" the `KafkaRebalance` resource when the spec is changed.

I have considered two other options:

1.  Automatically update the `KafkaRebalance` status with an error when the spec changes and log an error in the Cluster Operator. However, this seems a lot more verbose than automatically "refreshing" the resource to pick up the changes.

2. Prevent the `KafkaRebalance` spec from being updated after creation, display an error that the resource must be recreated when a user attempts to update the spec. This also seems a lot more verbose than automatically "refreshing" the resource to pick up the changes.

ShubhamRwt:
### Type of change

_Select the type of your PR_

- Bugfix
- Enhancement / new feature
- Refactoring
- Documentation

### Description

This PR fixes #8104. Currently, when we update the spec section of the KafkaRebalance resource when resource has already been applied then the status of the KafkaRebalance resource is not updated w.r.t to the edits. This PR fixes this issue.

Let's see how the mechanism works:

1. Deploy the operator and apply the Kafka resource with Cruise Control enabled
2. Now apply the `KafkaRebalance` resource with goals
3. KafkaRebalance resource changes are detected via a watch set up by the Kafka Rebalance Operator after every KafkaRebalance resource status change and state polling period. Every-time  there is an update to the rebalance, the watcher gets that event. 
4. The `MODIFIED` event is received whenever there is some change in the KafkaRebalance resource like spec update or status update etc.
5. Now only judging based on `Modified` event wouldn't help us since it can be any change in the resource so for being sure that we are picking up the `spec changes, I am using `generation` field in metadata  of `KafkaRebalance` which is something coming from  Kubernetes and is updated or changed only when spec has changed. For eg. By default generation is 1 and when spec updates it changes to 2
6. So to compare previous and current generation, I am using ` old generation` annotation which keeps track of the older generation.
7. So now if the current and older generation differs then I am patching the KafkaRebalance resource with `refresh` annotation. 

Here is how the Kafka Rebalance Operator handle “refresh” requests in every state:

PendingProposal -> Requests new proposal.
ProposalReady -> Refresh and generates a fresh proposal
Rebalancing -> Stops ongoing rebalance, then request a new proposal if stopping the rebalance was success
Stopped -> request new proposal.
NotReady -> request new proposal.
Ready -> request new proposal.

### Checklist

- [ ] Write tests
- [ ] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/api/src/main/java/io/strimzi/api/kafka/model/balancing/KafkaRebalanceState.java
+++ b/api/src/main/java/io/strimzi/api/kafka/model/balancing/KafkaRebalanceState.java
@@ -21,7 +21,7 @@ public enum KafkaRebalanceState {
      * A proposal has been requested from Cruise Control, but is not ready yet.
      * Transitions to:
      * <dl>
-     *     <dt>PendingProposal</dt><dd>A rebalance proposal is not ready yet or strimzi.io/rebalance=refresh is applied and a new proposal is requested</dd>
+     *     <dt>PendingProposal</dt><dd>A rebalance proposal is not ready yet or strimzi.io/rebalance=refresh is applied and a new proposal is requested.</dd>
      *     <dt>ProposalReady</dt><dd>Once Cruise Control has a ready proposal.</dd>
      *     <dt>NotReady</dt><dd>If Cruise Control returned an error</dd>
      * </dl>
--- a/api/src/main/java/io/strimzi/api/kafka/model/balancing/KafkaRebalanceState.java
+++ b/api/src/main/java/io/strimzi/api/kafka/model/balancing/KafkaRebalanceState.java
@@ -26,7 +26,7 @@ public enum KafkaRebalanceState {
      *     <dt>NotReady</dt><dd>If Cruise Control returned an error</dd>
      * </dl>
      */
-    PendingProposal(List.of(KafkaRebalanceAnnotation.stop)),
+    PendingProposal(List.of(KafkaRebalanceAnnotation.stop, KafkaRebalanceAnnotation.refresh)),
     /**
      * A proposal is ready and waiting for approval.
      * Transitions to:
--- a/api/src/main/java/io/strimzi/api/kafka/model/balancing/KafkaRebalanceState.java
+++ b/api/src/main/java/io/strimzi/api/kafka/model/balancing/KafkaRebalanceState.java
@@ -21,7 +21,7 @@ public enum KafkaRebalanceState {
      * A proposal has been requested from Cruise Control, but is not ready yet.
      * Transitions to:
      * <dl>
-     *     <dt>PendingProposal</dt><dd>A rebalance proposal is not ready yet or strimzi.io/rebalance=refresh is applied and a new proposal is requested.</dd>
+     *     <dt>PendingProposal</dt><dd>A rebalance proposal is not ready yet.</dd>
      *     <dt>ProposalReady</dt><dd>Once Cruise Control has a ready proposal.</dd>
      *     <dt>NotReady</dt><dd>If Cruise Control returned an error</dd>
      * </dl>
--- a/api/src/main/java/io/strimzi/api/kafka/model/balancing/KafkaRebalanceState.java
+++ b/api/src/main/java/io/strimzi/api/kafka/model/balancing/KafkaRebalanceState.java
@@ -21,7 +21,7 @@ public enum KafkaRebalanceState {
      * A proposal has been requested from Cruise Control, but is not ready yet.
      * Transitions to:
      * <dl>
-     *     <dt>PendingProposal</dt><dd>A rebalance proposal is not ready yet or strimzi.io/rebalance=refresh is applied due to spec changes</dd>
+     *     <dt>PendingProposal</dt><dd>A rebalance proposal is not ready yet or strimzi.io/rebalance=refresh is applied and a new proposal is requested</dd>
      *     <dt>ProposalReady</dt><dd>Once Cruise Control has a ready proposal.</dd>
      *     <dt>NotReady</dt><dd>If Cruise Control returned an error</dd>
      * </dl>
@@ -44,7 +44,7 @@ public enum KafkaRebalanceState {
      *     <dt>Rebalancing</dt><dd>While the actual rebalancing is still ongoing</dd>
      *     <dt>Stopped</dt><dd>If the user sets annotation strimzi.io/rebalance=stop.</dd>
      *     <dt>Ready</dt><dd>Once the rebalancing is complete.</dd>
-     *     <dt>ProposalReady</dt><dd>When annotation strimzi.io/rebalance=refresh is applied on rebalance resource due to spec changes and the proposal is updated.</dd>
+     *     <dt>ProposalReady</dt><dd>When annotation strimzi.io/rebalance=refresh is applied on rebalance resource and the proposal is updated.</dd>
      * </dl>
      */
     Rebalancing(List.of(KafkaRebalanceAnnotation.stop, KafkaRebalanceAnnotation.refresh)),
--- a/api/src/main/java/io/strimzi/api/kafka/model/balancing/KafkaRebalanceState.java
+++ b/api/src/main/java/io/strimzi/api/kafka/model/balancing/KafkaRebalanceState.java
@@ -21,7 +21,7 @@ public enum KafkaRebalanceState {
      * A proposal has been requested from Cruise Control, but is not ready yet.
      * Transitions to:
      * <dl>
-     *     <dt>PendingProposal</dt><dd>A rebalance proposal is not ready yet.</dd>
+     *     <dt>PendingProposal</dt><dd>A rebalance proposal is not ready yet or strimzi.io/rebalance=refresh is applied due to spec changes</dd>
      *     <dt>ProposalReady</dt><dd>Once Cruise Control has a ready proposal.</dd>
      *     <dt>NotReady</dt><dd>If Cruise Control returned an error</dd>
      * </dl>
@@ -44,6 +44,7 @@ public enum KafkaRebalanceState {
      *     <dt>Rebalancing</dt><dd>While the actual rebalancing is still ongoing</dd>
      *     <dt>Stopped</dt><dd>If the user sets annotation strimzi.io/rebalance=stop.</dd>
      *     <dt>Ready</dt><dd>Once the rebalancing is complete.</dd>
+     *     <dt>ProposalReady</dt><dd>When annotation strimzi.io/rebalance=refresh is applied on rebalance resource due to spec changes and the proposal is updated.</dd>
      * </dl>
      */
     Rebalancing(List.of(KafkaRebalanceAnnotation.stop, KafkaRebalanceAnnotation.refresh)),
--- a/api/src/main/java/io/strimzi/api/kafka/model/balancing/KafkaRebalanceState.java
+++ b/api/src/main/java/io/strimzi/api/kafka/model/balancing/KafkaRebalanceState.java
@@ -46,7 +46,7 @@ public enum KafkaRebalanceState {
      *     <dt>Ready</dt><dd>Once the rebalancing is complete.</dd>
      * </dl>
      */
-    Rebalancing(List.of(KafkaRebalanceAnnotation.stop)),
+    Rebalancing(List.of(KafkaRebalanceAnnotation.stop, KafkaRebalanceAnnotation.refresh)),
     /**
      * The user has stopped the actual rebalancing by setting annotation strimzi.io/rebalance=stop
      * May transition back to:
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceAssemblyOperator.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceAssemblyOperator.java
@@ -75,7 +75,9 @@ import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
 import static io.strimzi.operator.cluster.operator.resource.cruisecontrol.CruiseControlApiImpl.HTTP_DEFAULT_IDLE_TIMEOUT_SECONDS;
-import static io.strimzi.operator.common.Annotations.*;
+import static io.strimzi.operator.common.Annotations.ANNO_STRIMZI_IO_REBALANCE;
+import static io.strimzi.operator.common.Annotations.ANNO_STRIMZI_IO_REBALANCE_AUTOAPPROVAL;
+import static io.strimzi.operator.common.Annotations.ANNO_KUBE_GENERATION;
 
 /**
  * <p>Assembly operator for a "KafkaRebalance" assembly, which interacts with the Cruise Control REST API</p>
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceAssemblyOperator.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceAssemblyOperator.java
@@ -238,20 +238,19 @@ public class KafkaRebalanceAssemblyOperator
         if (action == Watcher.Action.MODIFIED) {
             KafkaRebalanceBuilder patchedKafkaRebalance = new KafkaRebalanceBuilder(resource);
 
-            if (!resource.getMetadata().getAnnotations().containsKey("oldGeneration")) {
-                patchedKafkaRebalance.editMetadata().addToAnnotations(Map.of("oldGeneration", "1")).endMetadata();
-            }
-
-            if (!patchedKafkaRebalance.buildMetadata().getAnnotations().get("oldGeneration").equals(resource.getMetadata().getGeneration().toString())) {
+            if (patchedKafkaRebalance.buildStatus() != null
+                    && patchedKafkaRebalance.buildStatus().getObservedGeneration() != resource.getMetadata().getGeneration()) {
                 patchedKafkaRebalance.editMetadata()
                         .addToAnnotations(Map.of(ANNO_STRIMZI_IO_REBALANCE, KafkaRebalanceAnnotation.refresh.toString()))
-                        .addToAnnotations(Map.of("oldGeneration", resource.getMetadata().getGeneration().toString()))
-                        .endMetadata();
+                        .endMetadata().
+                        editStatus()
+                        .withObservedGeneration(resource.getMetadata().getGeneration())
+                        .endStatus();
             }
 
-            kafkaRebalanceOperator.patchAsync(reconciliation, patchedKafkaRebalance.build());
-            withLock(reconciliation, LOCK_TIMEOUT_MS,
-                    () -> reconcileRebalance(reconciliation, patchedKafkaRebalance.build()));
+            kafkaRebalanceOperator.patchAsync(reconciliation, patchedKafkaRebalance.build()).onComplete(
+                    r -> withLock(reconciliation, LOCK_TIMEOUT_MS,
+                            () -> reconcileRebalance(reconciliation, patchedKafkaRebalance.build())));
         } else {
             withLock(reconciliation, LOCK_TIMEOUT_MS,
                     () -> reconcileRebalance(reconciliation, action == Watcher.Action.DELETED ? null : resource));
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceAssemblyOperator.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceAssemblyOperator.java
@@ -851,7 +851,6 @@ public class KafkaRebalanceAssemblyOperator
                             // Safety check as timer might be called again (from a delayed timer firing)
                             if (state(currentKafkaRebalance) == KafkaRebalanceState.PendingProposal) {
                                 if (rebalanceAnnotation(currentKafkaRebalance) == KafkaRebalanceAnnotation.refresh) {
-                                    System.out.println("hitman");
                                     LOGGER.infoCr(reconciliation, "Requesting a new proposal since spec has been updated");
                                     vertx.cancelTimer(t);
                                     requestRebalance(reconciliation, host, apiClient, currentKafkaRebalance, true, rebalanceOptionsBuilder).onSuccess(p::complete);
@@ -1162,21 +1161,21 @@ public class KafkaRebalanceAssemblyOperator
             return Future.succeededFuture();
         }
 
-        KafkaRebalanceBuilder patchedKafkaRebalance = new KafkaRebalanceBuilder(kafkaRebalance);
+        if (kafkaRebalance.getStatus() != null
+                && kafkaRebalance.getStatus().getObservedGeneration() != kafkaRebalance.getMetadata().getGeneration()) {
 
-        if (patchedKafkaRebalance.buildStatus() != null
-                && patchedKafkaRebalance.buildStatus().getObservedGeneration() != kafkaRebalance.getMetadata().getGeneration()) {
+            KafkaRebalanceBuilder patchedKafkaRebalance = new KafkaRebalanceBuilder(kafkaRebalance);
 
-            System.out.println(patchedKafkaRebalance.buildStatus().getObservedGeneration());
-            patchedKafkaRebalance.editMetadata()
-                    .addToAnnotations(Map.of(ANNO_STRIMZI_IO_REBALANCE, KafkaRebalanceAnnotation.refresh.toString()))
+            patchedKafkaRebalance
+                    .editMetadata()
+                         .addToAnnotations(Map.of(ANNO_STRIMZI_IO_REBALANCE, KafkaRebalanceAnnotation.refresh.toString()))
                     .endMetadata()
                     .editStatus()
-                    .withObservedGeneration(kafkaRebalance.getMetadata().getGeneration())
+                         .withObservedGeneration(kafkaRebalance.getMetadata().getGeneration())
                     .endStatus();
 
             kafkaRebalanceOperator.patchAsync(reconciliation, patchedKafkaRebalance.build()).onComplete(
-                    r -> LOGGER.infoCr(reconciliation, "Done"));
+                    r -> LOGGER.infoCr(reconciliation, "The KafkaRebalance resource is updated with refresh annotation since spec is updated"));
         }
 
         String clusterName = kafkaRebalance.getMetadata().getLabels() == null ? null : kafkaRebalance.getMetadata().getLabels().get(Labels.STRIMZI_CLUSTER_LABEL);
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceAssemblyOperator.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceAssemblyOperator.java
@@ -75,8 +75,7 @@ import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
 import static io.strimzi.operator.cluster.operator.resource.cruisecontrol.CruiseControlApiImpl.HTTP_DEFAULT_IDLE_TIMEOUT_SECONDS;
-import static io.strimzi.operator.common.Annotations.ANNO_STRIMZI_IO_REBALANCE;
-import static io.strimzi.operator.common.Annotations.ANNO_STRIMZI_IO_REBALANCE_AUTOAPPROVAL;
+import static io.strimzi.operator.common.Annotations.*;
 
 /**
  * <p>Assembly operator for a "KafkaRebalance" assembly, which interacts with the Cruise Control REST API</p>
@@ -256,7 +255,6 @@ public class KafkaRebalanceAssemblyOperator
             withLock(reconciliation, LOCK_TIMEOUT_MS,
                     () -> reconcileRebalance(reconciliation, action == Watcher.Action.DELETED ? null : resource));
         }
-
     }
 
     /**
@@ -871,7 +869,7 @@ public class KafkaRebalanceAssemblyOperator
                             // Safety check as timer might be called again (from a delayed timer firing)
                             if (state(currentKafkaRebalance) == KafkaRebalanceState.PendingProposal) {
                                 if (rebalanceAnnotation(currentKafkaRebalance) == KafkaRebalanceAnnotation.refresh) {
-                                    LOGGER.debugCr(reconciliation, "Requesting for a new proposal since spec is updated");
+                                    LOGGER.infoCr(reconciliation, "Requesting a new proposal since spec has been updated");
                                     vertx.cancelTimer(t);
                                     requestRebalance(reconciliation, host, apiClient, currentKafkaRebalance, true, rebalanceOptionsBuilder).onSuccess(p::complete);
                                 } else if (rebalanceAnnotation(currentKafkaRebalance) == KafkaRebalanceAnnotation.stop) {
@@ -1022,7 +1020,7 @@ public class KafkaRebalanceAssemblyOperator
                                             p.fail(e.getCause());
                                         });
                                 } else if (rebalanceAnnotation(currentKafkaRebalance) == KafkaRebalanceAnnotation.refresh) {
-                                    LOGGER.debugCr(reconciliation, "Stopping current Cruise Control rebalance user task since spec has been updated and requesting a new one");
+                                    LOGGER.infoCr(reconciliation, "Stopping current Cruise Control rebalance user task since spec has been updated and requesting a new one");
                                     vertx.cancelTimer(t);
                                     apiClient.stopExecution(host, CruiseControl.REST_API_PORT)
                                             .onSuccess(r -> {
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceAssemblyOperator.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceAssemblyOperator.java
@@ -851,7 +851,7 @@ public class KafkaRebalanceAssemblyOperator
                             // Safety check as timer might be called again (from a delayed timer firing)
                             if (state(currentKafkaRebalance) == KafkaRebalanceState.PendingProposal) {
                                 if (rebalanceAnnotation(currentKafkaRebalance) == KafkaRebalanceAnnotation.refresh) {
-                                    LOGGER.infoCr(reconciliation, "Requesting a new proposal since spec has been updated");
+                                    LOGGER.debugCr(reconciliation, "Requesting a new proposal since refresh annotation is applied on the KafkaRebalance resource");
                                     vertx.cancelTimer(t);
                                     requestRebalance(reconciliation, host, apiClient, currentKafkaRebalance, true, rebalanceOptionsBuilder).onSuccess(p::complete);
                                 } else if (rebalanceAnnotation(currentKafkaRebalance) == KafkaRebalanceAnnotation.stop) {
@@ -1002,7 +1002,7 @@ public class KafkaRebalanceAssemblyOperator
                                             p.fail(e.getCause());
                                         });
                                 } else if (rebalanceAnnotation(currentKafkaRebalance) == KafkaRebalanceAnnotation.refresh) {
-                                    LOGGER.infoCr(reconciliation, "Stopping current Cruise Control rebalance user task since spec has been updated and requesting a new one");
+                                    LOGGER.debugCr(reconciliation, "Stopping current Cruise Control rebalance user task since refresh annotation is applied on the KafkaRebalance resource and requesting a new proposal");
                                     vertx.cancelTimer(t);
                                     apiClient.stopExecution(host, CruiseControl.REST_API_PORT)
                                             .onSuccess(r -> {
@@ -1175,7 +1175,7 @@ public class KafkaRebalanceAssemblyOperator
                     .endStatus();
 
             kafkaRebalanceOperator.patchAsync(reconciliation, patchedKafkaRebalance.build()).onComplete(
-                    r -> LOGGER.infoCr(reconciliation, "The KafkaRebalance resource is updated with refresh annotation since spec is updated"));
+                    r -> LOGGER.debugCr(reconciliation, "The KafkaRebalance resource is updated with refresh annotation"));
         }
 
         String clusterName = kafkaRebalance.getMetadata().getLabels() == null ? null : kafkaRebalance.getMetadata().getLabels().get(Labels.STRIMZI_CLUSTER_LABEL);
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceAssemblyOperator.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceAssemblyOperator.java
@@ -241,10 +241,10 @@ public class KafkaRebalanceAssemblyOperator
             if (patchedKafkaRebalance.buildStatus() != null
                     && patchedKafkaRebalance.buildStatus().getObservedGeneration() != resource.getMetadata().getGeneration()) {
                 patchedKafkaRebalance.editMetadata()
-                        .addToAnnotations(Map.of(ANNO_STRIMZI_IO_REBALANCE, KafkaRebalanceAnnotation.refresh.toString()))
-                        .endMetadata().
-                        editStatus()
-                        .withObservedGeneration(resource.getMetadata().getGeneration())
+                             .addToAnnotations(Map.of(ANNO_STRIMZI_IO_REBALANCE, KafkaRebalanceAnnotation.refresh.toString()))
+                        .endMetadata()
+                        .editStatus()
+                             .withObservedGeneration(resource.getMetadata().getGeneration())
                         .endStatus();
             }
 
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceAssemblyOperator.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceAssemblyOperator.java
@@ -235,26 +235,8 @@ public class KafkaRebalanceAssemblyOperator
                 resource.getStatus() != null ? rebalanceStateConditionType(resource.getStatus()) : null,
                 ANNO_STRIMZI_IO_REBALANCE, rawRebalanceAnnotation(resource));
 
-        if (action == Watcher.Action.MODIFIED) {
-            KafkaRebalanceBuilder patchedKafkaRebalance = new KafkaRebalanceBuilder(resource);
-
-            if (patchedKafkaRebalance.buildStatus() != null
-                    && patchedKafkaRebalance.buildStatus().getObservedGeneration() != resource.getMetadata().getGeneration()) {
-                patchedKafkaRebalance.editMetadata()
-                             .addToAnnotations(Map.of(ANNO_STRIMZI_IO_REBALANCE, KafkaRebalanceAnnotation.refresh.toString()))
-                        .endMetadata()
-                        .editStatus()
-                             .withObservedGeneration(resource.getMetadata().getGeneration())
-                        .endStatus();
-            }
-
-            kafkaRebalanceOperator.patchAsync(reconciliation, patchedKafkaRebalance.build()).onComplete(
-                    r -> withLock(reconciliation, LOCK_TIMEOUT_MS,
-                            () -> reconcileRebalance(reconciliation, patchedKafkaRebalance.build())));
-        } else {
             withLock(reconciliation, LOCK_TIMEOUT_MS,
                     () -> reconcileRebalance(reconciliation, action == Watcher.Action.DELETED ? null : resource));
-        }
     }
 
     /**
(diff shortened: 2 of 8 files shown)


scholzj (commented):
You should probably clearly describe here how does it work, especially with regards of what happens when the spec is changed in the different phases. That might be a topic for the docs as well, but having the details in the PR description would be a start I guess.

ShubhamRwt:
@scholzj Good point. Let me make the discription more descriptive

ShubhamRwt:
@scholzj I have added more info. Is it good as a description or do we need more clarification

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceAssemblyOperator.java
```diff
@@ -1001,6 +1025,17 @@ private Future<MapAndStatus<ConfigMap, KafkaRebalanceStatus>> onRebalancing(Reco
                                             LOGGER.errorCr(reconciliation, "Cruise Control stopping execution failed", e.getCause());
                                             p.fail(e.getCause());
                                         });
+                                } else if (rebalanceAnnotation(currentKafkaRebalance) == KafkaRebalanceAnnotation.refresh) {
+                                    LOGGER.debugCr(reconciliation, "Stopping current Cruise Control rebalance user task since spec has been updated and requesting a new one");
+                                    vertx.cancelTimer(t);
+                                    apiClient.stopExecution(host, CruiseControl.REST_API_PORT)
```
kyguy, line 1031: A couple thoughts about the partitions that have already been moved by the rebalance operation which we are stopping. 
- Will the newly requested proposal account for these moved partitions
- Is there any reason we would want to roll back the changes/movements? (Not that we can do anything here as part of this PR concerning this, that functionality would need its own PR)

Make me second guess our direction and think if it would be better if instead of stopping and requesting new proposal in the `Rebalacing` state we just log and update the `KafkaRebalance` status section with an error. Just a thought, let me know what you think!

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceAssemblyOperator.java
```diff
@@ -223,13 +223,32 @@ public void eventReceived(Action action, KafkaRebalance kafkaRebalance) {
...
+
+                        if (!kafkaRebalance.getMetadata().getAnnotations().containsKey("oldGeneration")) {
+                            patchedKafkaRebalance.editMetadata().addToAnnotations(Map.of("oldGeneration", "1")).endMetadata();
+                        }
+
+                        if (!patchedKafkaRebalance.buildMetadata().getAnnotations().get("oldGeneration").equals(kafkaRebalance.getMetadata().getGeneration().toString())) {
+                            patchedKafkaRebalance.editMetadata()
+                                    .addToAnnotations(Map.of(ANNO_STRIMZI_IO_REBALANCE, KafkaRebalanceAnnotation.refresh.toString()))
+                                    .addToAnnotations(Map.of("oldGeneration", kafkaRebalance.getMetadata().getGeneration().toString()))
+                                    .endMetadata();
```
kyguy, line 242: When does the "oldGeneration" annotation value get incremented? Or does it only ever have  one of two values - null or 1?

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceAssemblyOperator.java
```diff
@@ -223,13 +223,32 @@ public void eventReceived(Action action, KafkaRebalance kafkaRebalance) {
...
                             kafkaRebalance.getMetadata().getName(),
                             kafkaRebalance.getStatus() != null ? rebalanceStateConditionType(kafkaRebalance.getStatus()) : null,
                             ANNO_STRIMZI_IO_REBALANCE, rawRebalanceAnnotation(kafkaRebalance));
 
-                    withLock(reconciliation, LOCK_TIMEOUT_MS,
-                        () -> reconcileRebalance(reconciliation, action == Action.DELETED ? null : kafkaRebalance));
+                    if (action == Action.MODIFIED) {
+                        KafkaRebalanceBuilder patchedKafkaRebalance = new KafkaRebalanceBuilder(kafkaRebalance);
+
+                        if (!kafkaRebalance.getMetadata().getAnnotations().containsKey("oldGeneration")) {
```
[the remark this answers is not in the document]
ShubhamRwt: the `oldGeneration` annotations comes into picture when spec has changed and this annotation wouldn't be available if there is no spec change. But I guess maybe when its available a user might be(not sure) able to change it

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceAssemblyOperator.java
```diff
@@ -223,13 +223,32 @@ public void eventReceived(Action action, KafkaRebalance kafkaRebalance) {
...
+                        }
+
+                        if (!patchedKafkaRebalance.buildMetadata().getAnnotations().get("oldGeneration").equals(kafkaRebalance.getMetadata().getGeneration().toString())) {
+                            patchedKafkaRebalance.editMetadata()
+                                    .addToAnnotations(Map.of(ANNO_STRIMZI_IO_REBALANCE, KafkaRebalanceAnnotation.refresh.toString()))
+                                    .addToAnnotations(Map.of("oldGeneration", kafkaRebalance.getMetadata().getGeneration().toString()))
+                                    .endMetadata();
+                        }
+
+                        kafkaRebalanceOperator.patchAsync(reconciliation, patchedKafkaRebalance.build());
```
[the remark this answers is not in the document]
ShubhamRwt: yeah that is something I can look into. Maybe using onSuccess() or onComplete() for this

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceAssemblyOperator.java
```diff
@@ -1001,6 +1025,17 @@ private Future<MapAndStatus<ConfigMap, KafkaRebalanceStatus>> onRebalancing(Reco
...
+                                } else if (rebalanceAnnotation(currentKafkaRebalance) == KafkaRebalanceAnnotation.refresh) {
+                                    LOGGER.debugCr(reconciliation, "Stopping current Cruise Control rebalance user task since spec has been updated and requesting a new one");
+                                    vertx.cancelTimer(t);
+                                    apiClient.stopExecution(host, CruiseControl.REST_API_PORT)
+                                            .onSuccess(r -> {
+                                                requestRebalance(reconciliation, host, apiClient, currentKafkaRebalance, true, rebalanceOptionsBuilder).onSuccess(p::complete);
+                                            })
+                                            .onFailure(e -> {
+                                                LOGGER.errorCr(reconciliation, "Cruise Control stopping execution failed", e.getCause());
+                                                p.fail(e.getCause());
```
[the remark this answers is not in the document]
ShubhamRwt: Yes

api/src/main/java/io/strimzi/api/kafka/model/balancing/KafkaRebalanceState.java
```diff
@@ -26,7 +26,7 @@ public enum KafkaRebalanceState {
      *     <dt>NotReady</dt><dd>If Cruise Control returned an error</dd>
      * </dl>
      */
-    PendingProposal(List.of(KafkaRebalanceAnnotation.stop)),
+    PendingProposal(List.of(KafkaRebalanceAnnotation.stop, KafkaRebalanceAnnotation.refresh)),
```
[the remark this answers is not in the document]
ShubhamRwt: I added `refresh` here to say that `PendingProposal` also supports refresh annotation. Without this we can get log saying that `refresh` is not supported in `PendingProposal` state

scholzj, line 29: Should the docs above be updated?

api/src/main/java/io/strimzi/api/kafka/model/balancing/KafkaRebalanceState.java
```diff
@@ -46,7 +46,7 @@ public enum KafkaRebalanceState {
      *     <dt>Ready</dt><dd>Once the rebalancing is complete.</dd>
      * </dl>
      */
-    Rebalancing(List.of(KafkaRebalanceAnnotation.stop)),
+    Rebalancing(List.of(KafkaRebalanceAnnotation.stop, KafkaRebalanceAnnotation.refresh)),
```
scholzj, line 49: Should the docs above be updated?

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceAssemblyOperator.java
```diff
@@ -223,13 +223,32 @@ public void eventReceived(Action action, KafkaRebalance kafkaRebalance) {
                     Reconciliation reconciliation = new Reconciliation("kafkarebalance-watch", kafkaRebalance.getKind(),
                             kafkaRebalance.getMetadata().getNamespace(), kafkaRebalance.getMetadata().getName());
 
-                    LOGGER.debugCr(reconciliation, "EventReceived {} on {} with status [{}] and {}={}", action,
+                    LOGGER.infoCr(reconciliation, "EventReceived {} on {} with status [{}] and {}={}", action,
```
scholzj, line 226: Is this intentional?

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceAssemblyOperator.java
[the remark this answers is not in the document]
scholzj: What is the `oldGeneration` annotation? Who sets it?

ppatierno:
@ShubhamRwt I had a pass and waiting for outstanding questions from Kyle and Jakub. I have the same kind of questions.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceAssemblyOperator.java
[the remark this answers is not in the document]
ShubhamRwt: It is being setted by me @scholzj . It is used to track the previous `generation` count. The `generation` field is settted by Kubernetes and is updated on every spec change. So I am using `oldGeneration` annoation to track the previous `generation`.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceAssemblyOperator.java
ShubhamRwt: No, I will fix this, Sorry.

api/src/main/java/io/strimzi/api/kafka/model/balancing/KafkaRebalanceState.java
ShubhamRwt: Sure let me do it

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceAssemblyOperator.java
ppatierno: @kyguy it's getting the value from the "generation" assigned to the resource by Kubernetes when the spec changes. Shubham is just saving it in an annotation as a way to know when spec is changed.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceAssemblyOperator.java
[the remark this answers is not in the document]
ppatierno: +1 what Jakub said

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceAssemblyOperator.java
ShubhamRwt: @kyguy I will have to test this use case. But if I am not wrong if we stop the rebalance and ask for a new rebalance. I think it will create a fresh proposal based on the current situation so it should account for the moved partition in my opinion.

scholzj:
> PendingProposal -> Requests new proposal.
ProposalReady -> Stops ongoing rebalance, then request a new proposal if stopping the rebalance was success
Stopped -> request new proposal.
NotReady -> request new proposal.
Ready -> request new proposal.

Isn't there a difference between `ProposalReady` and `Rebalancing`? `ProposalReady` is IMHO waiting for approval and not rebalancing yet.

ShubhamRwt:
> > PendingProposal -> Requests new proposal.
> > ProposalReady -> Stops ongoing rebalance, then request a new proposal if stopping the rebalance was success
> > Stopped -> request new proposal.
> > NotReady -> request new proposal.
> > Ready -> request new proposal.
> 
> Isn't there a difference between `ProposalReady` and `Rebalancing`? `ProposalReady` is IMHO waiting for approval and not rebalancing yet.

I guess its a typo it should be ‘Rebalancing’  and not ‘ProposalReady’

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceAssemblyOperator.java
```diff
@@ -223,13 +224,32 @@ public void eventReceived(Action action, KafkaRebalance kafkaRebalance) {
...
+
+                        if (!kafkaRebalance.getMetadata().getAnnotations().containsKey(ANNO_KUBE_GENERATION)) {
+                            patchedKafkaRebalance.editMetadata().addToAnnotations(Map.of(ANNO_KUBE_GENERATION, "1")).endMetadata();
+                        }
+
+                        if (!patchedKafkaRebalance.buildMetadata().getAnnotations().get(ANNO_KUBE_GENERATION).equals(kafkaRebalance.getMetadata().getGeneration().toString())) {
+                            patchedKafkaRebalance.editMetadata()
+                                    .addToAnnotations(Map.of(ANNO_STRIMZI_IO_REBALANCE, KafkaRebalanceAnnotation.refresh.toString()))
+                                    .addToAnnotations(Map.of(ANNO_KUBE_GENERATION, kafkaRebalance.getMetadata().getGeneration().toString()))
+                                    .endMetadata();
```
scholzj, line 243: I'm not an expert on this. But this seems a bit weird and this can cause fights with any tooling managing the KafkaRebalance. Why don't you store the generation in the observed generation field in the status? Do you really need to self-annotate the resource to refresh it?

api/src/main/java/io/strimzi/api/kafka/model/balancing/KafkaRebalanceState.java
```diff
@@ -21,12 +21,12 @@ public enum KafkaRebalanceState {
      * A proposal has been requested from Cruise Control, but is not ready yet.
      * Transitions to:
      * <dl>
-     *     <dt>PendingProposal</dt><dd>A rebalance proposal is not ready yet.</dd>
+     *     <dt>PendingProposal</dt><dd>A rebalance proposal is not ready yet or strimzi.io/rebalance=refresh is applied due to spec changes</dd>
```
scholzj, line 24: How do you know the `strimzi.io/rebalance=refresh` was added by you because of a spec change and not by the user because he wants a refresh?

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceAssemblyOperator.java
scholzj: Still not fixed I guess?

scholzj:
> I guess its a typo it should be ‘Rebalancing’ and not ‘ProposalReady’

But we have both states, or? So I guess one way or another, one is missing.

ShubhamRwt:
I have added both now @scholzj and fixed the comment

scholzj (commented):
Also, should there be some tests and documentation?

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceAssemblyOperator.java
ShubhamRwt: Sure, Iam testing this at the moment. I guess it should work fine

ShubhamRwt: Works fine

operator-common/src/main/java/io/strimzi/operator/common/Annotations.java
```diff
@@ -165,6 +165,11 @@ public class Annotations {
      */
     public static final String ANNO_DEP_KUBE_IO_REVISION = "deployment.kubernetes.io/revision";
 
+    /**
+     * Annotation for tracking Kubernetes generation in KafkaRebalanceAssemblyOperator
+     */
+    public static final String ANNO_KUBE_GENERATION = STRIMZI_DOMAIN + "kubeGeneration";
```
scholzj, line 171: Is it still used somewhere?

api/src/main/java/io/strimzi/api/kafka/model/balancing/KafkaRebalanceState.java
scholzj: I think this comment still stands. How do you know the annotation is there due to spec change?

api/src/main/java/io/strimzi/api/kafka/model/balancing/KafkaRebalanceState.java
```diff
@@ -44,9 +44,10 @@ public enum KafkaRebalanceState {
      *     <dt>Rebalancing</dt><dd>While the actual rebalancing is still ongoing</dd>
      *     <dt>Stopped</dt><dd>If the user sets annotation strimzi.io/rebalance=stop.</dd>
      *     <dt>Ready</dt><dd>Once the rebalancing is complete.</dd>
+     *     <dt>ProposalReady</dt><dd>When annotation strimzi.io/rebalance=refresh is applied on rebalance resource due to spec changes and the proposal is updated.</dd>
```
scholzj, line 47: Again -> how do you know the refresh anno was applied due to spec changes?

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceAssemblyOperator.java
```diff
@@ -228,8 +228,25 @@ public void eventReceived(Action action, KafkaRebalance kafkaRebalance) {
                             kafkaRebalance.getStatus() != null ? rebalanceStateConditionType(kafkaRebalance.getStatus()) : null,
                             ANNO_STRIMZI_IO_REBALANCE, rawRebalanceAnnotation(kafkaRebalance));
 
-                    withLock(reconciliation, LOCK_TIMEOUT_MS,
-                        () -> reconcileRebalance(reconciliation, action == Action.DELETED ? null : kafkaRebalance));
+                    if (action == Action.MODIFIED) {
+                        KafkaRebalanceBuilder patchedKafkaRebalance = new KafkaRebalanceBuilder(kafkaRebalance);
+
+                        if (patchedKafkaRebalance.buildStatus().getObservedGeneration() != (kafkaRebalance.getMetadata().getGeneration())) {
```
scholzj, line 234:
```suggestion
                        if (patchedKafkaRebalance.buildStatus().getObservedGeneration() != kafkaRebalance.getMetadata().getGeneration()) {
```

Also, what about null checks? How do you know that the status cannot be null?

api/src/main/java/io/strimzi/api/kafka/model/balancing/KafkaRebalanceState.java
ShubhamRwt: I had a go through regarding this. I think if we are `refresh` to trigger this mechanism then its possible for the user to do it manually themselves. I can fix this documentation. Would that be fine or we don't want the user to use `refresh` in these state manually @ppatierno @kyguy @scholzj ?

scholzj: My comment was really only about the Javadoc that did not seem to be right.

ShubhamRwt: Ahh Okay. I will fix it

ShubhamRwt:
Look like I pushed older commit. Let me fix that

systemtest/src/test/java/io/strimzi/systemtest/cruisecontrol/CruiseControlST.java
```diff
@@ -203,6 +203,51 @@ void testCruiseControlWithRebalanceResourceAndRefreshAnnotation(ExtensionContext
...
+                .editMetadata()
+                .withNamespace(clusterOperator.getDeploymentNamespace())
+                .endMetadata()
+                .build());
+
+        KafkaRebalanceUtils.waitForKafkaRebalanceCustomResourceState(clusterOperator.getDeploymentNamespace(), clusterName, KafkaRebalanceState.ProposalReady);
+
+        KafkaRebalanceResource.replaceKafkaRebalanceResourceInSpecificNamespace(clusterName, kafkaRebalance -> kafkaRebalance.getSpec().setGoals(List.of("custom-goal")), clusterOperator.getDeploymentNamespace());
+
+        KafkaRebalanceUtils.waitForKafkaRebalanceCustomResourceState(clusterOperator.getDeploymentNamespace(), clusterName, KafkaRebalanceState.NotReady);
```
scholzj, line 222: Why `NotReady`? Shouldn't it go to ProposalPending or ProposalReady? (as waiting only for ProposalPending only might cause race condition)

systemtest/src/test/java/io/strimzi/systemtest/cruisecontrol/CruiseControlST.java
```diff
@@ -203,6 +203,51 @@ void testCruiseControlWithRebalanceResourceAndRefreshAnnotation(ExtensionContext
...
+
+        KafkaRebalanceUtils.waitForKafkaRebalanceCustomResourceState(clusterOperator.getDeploymentNamespace(), clusterName, KafkaRebalanceState.ProposalReady);
+
+        LOGGER.info("Annotating KafkaRebalance: {} with 'approve' anno", clusterName);
+        KafkaRebalanceUtils.annotateKafkaRebalanceResource(new Reconciliation("test", KafkaRebalance.RESOURCE_KIND, clusterOperator.getDeploymentNamespace(), clusterName), clusterOperator.getDeploymentNamespace(), clusterName, KafkaRebalanceAnnotation.approve);
+        KafkaRebalanceUtils.waitForKafkaRebalanceCustomResourceState(clusterOperator.getDeploymentNamespace(), clusterName, KafkaRebalanceState.Rebalancing);
+
+        KafkaRebalanceResource.replaceKafkaRebalanceResourceInSpecificNamespace(clusterName, kafkaRebalance -> kafkaRebalance.getSpec().setReplicationThrottle(100000), clusterOperator.getDeploymentNamespace());
+
+        KafkaRebalanceUtils.waitForKafkaRebalanceCustomResourceState(clusterOperator.getDeploymentNamespace(), clusterName, KafkaRebalanceState.ProposalReady);
```
scholzj, line 248: Will this be really stable? It sounds like this will be open to race conditions where you miss the `Rebalancing` state.

systemtest/src/test/java/io/strimzi/systemtest/cruisecontrol/CruiseControlST.java
```diff
@@ -203,6 +203,51 @@ void testCruiseControlWithRebalanceResourceAndRefreshAnnotation(ExtensionContext
         KafkaRebalanceUtils.doRebalancingProcess(new Reconciliation("test", KafkaRebalance.RESOURCE_KIND, Constants.TEST_SUITE_NAMESPACE, clusterName), Constants.TEST_SUITE_NAMESPACE, clusterName);
     }
 
+    @IsolatedTest
+    void testCruiseControlSpecChangesReflectedInStatusWhenSpecChanges(ExtensionContext extensionContext) {
```
im-konge, line 207: I guess we can maybe put these two tests together or?

systemtest/src/test/java/io/strimzi/systemtest/cruisecontrol/CruiseControlST.java
```diff
@@ -203,6 +203,51 @@ void testCruiseControlWithRebalanceResourceAndRefreshAnnotation(ExtensionContext
...
+        String clusterName = mapWithClusterNames.get(extensionContext.getDisplayName());
+
+        resourceManager.createResourceWithoutWait(extensionContext, KafkaTemplates.kafkaWithCruiseControl(clusterName, 3, 1).build());
+
+        resourceManager.createResourceWithoutWait(extensionContext,  KafkaRebalanceTemplates.kafkaRebalance(clusterName)
+                .editMetadata()
+                .withNamespace(clusterOperator.getDeploymentNamespace())
+                .endMetadata()
+                .build());
+
```
im-konge, line 217: indent

systemtest/src/test/java/io/strimzi/systemtest/cruisecontrol/CruiseControlST.java
```diff
@@ -203,6 +203,51 @@ void testCruiseControlWithRebalanceResourceAndRefreshAnnotation(ExtensionContext
...
```
im-konge, line 239: indent

systemtest/src/test/java/io/strimzi/systemtest/cruisecontrol/CruiseControlST.java
```diff
@@ -203,6 +203,51 @@ void testCruiseControlWithRebalanceResourceAndRefreshAnnotation(ExtensionContext
...
+        KafkaRebalanceUtils.waitForKafkaRebalanceCustomResourceState(clusterOperator.getDeploymentNamespace(), clusterName, KafkaRebalanceState.NotReady);
+
+        KafkaRebalanceStatus kafkaRebalanceStatus = KafkaRebalanceResource.kafkaRebalanceClient().inNamespace(clusterOperator.getDeploymentNamespace()).withName(clusterName).get().getStatus();
+        assertThat(kafkaRebalanceStatus.getConditions().get(0).getReason(), is("CruiseControlRestException"));
+    }
+
+    @IsolatedTest
+    void testCruiseControlChangesFromRebalancingtoProposalReadyWhenSpecUpdated(ExtensionContext extensionContext) {
+        String clusterName = mapWithClusterNames.get(extensionContext.getDisplayName());
```
im-konge, line 230: in one of the future PRs we will remove the `mapWith*` maps, so you should use 
```
TestStorage testStorage = new TestStorage(extensionContext, Constants.TEST_SUITE_NAMESPACE)
```

systemtest/src/test/java/io/strimzi/systemtest/cruisecontrol/CruiseControlST.java
```diff
@@ -203,6 +203,51 @@ void testCruiseControlWithRebalanceResourceAndRefreshAnnotation(ExtensionContext
...
+
+        resourceManager.createResourceWithoutWait(extensionContext,  KafkaRebalanceTemplates.kafkaRebalance(clusterName)
+                .editMetadata()
+                .withNamespace(clusterOperator.getDeploymentNamespace())
+                .endMetadata()
+                .build());
+
+        KafkaRebalanceUtils.waitForKafkaRebalanceCustomResourceState(clusterOperator.getDeploymentNamespace(), clusterName, KafkaRebalanceState.ProposalReady);
+
+        LOGGER.info("Annotating KafkaRebalance: {} with 'approve' anno", clusterName);
```
kyguy, line 242: Should we spell out annotation here?
```suggestion
        LOGGER.info("Annotating KafkaRebalance: {} with 'approve' annotation", clusterName);
```

systemtest/src/test/java/io/strimzi/systemtest/cruisecontrol/CruiseControlST.java
kyguy: We could most definitely miss the `Rebalancing` state, maybe we should skip the wait on line 244 where we wait for the `Reblancing` state? I am not sure we can guarantee or force the resource will be in `Rebalancing` state when we check, or test this specific scenario reliably. That being said, if we remove the wait on line 244 the resource should end in the `ProposalReady` regardless the outcome of the race condition.

api/src/main/java/io/strimzi/api/kafka/model/balancing/KafkaRebalanceState.java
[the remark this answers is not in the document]
kyguy: Do we need to add a note to the `Ready` state as well, since the annotation could be applied in that state too if the spec is updated?

kyguy (commented):
Solid work @ShubhamRwt, this is a useful enhancement and works smoothly from what I have tested.

With this change, we no longer need to manually refresh the `KafkaRebalance` resource to reuse it, we can just update the `KafkaRebalance` spec. Maybe we should update the docs and logs to reflect this too. For example, updating logs like this:
```
Rebalancing is completed. You can use the `refresh` annotation to ask for a new rebalance request 
```
Let me know what you think!

systemtest/src/test/java/io/strimzi/systemtest/cruisecontrol/CruiseControlST.java
ShubhamRwt: Yes @kyguy , Correct.

ShubhamRwt: I think I should rename this test as to avoid confusion

systemtest/src/test/java/io/strimzi/systemtest/cruisecontrol/CruiseControlST.java
ShubhamRwt: Iam not sure about that since one test is to check if error is propogated correctly and other is to test if new request completes successfully but yeah I can try to merge them

ShubhamRwt: @scholzj @kyguy WDYT? Should I merge these tests or?

systemtest/src/test/java/io/strimzi/systemtest/cruisecontrol/CruiseControlST.java
ShubhamRwt: Great idea @kyguy. Let me fix that

systemtest/src/test/java/io/strimzi/systemtest/cruisecontrol/CruiseControlST.java
scholzj: Not everything needs a system test. So maybe you should cover some of the logic in unit tests.

scholzj: You should not need to use the watch. The watch should just trigger reconciliation, so you can trigger the reconciliation multiple times without the watch. Also, if there is no other way you could use the Mock Kube already used in some KafkaReblanace related tests.

ShubhamRwt: I tried as you suggested @scholzj. Just one issue, The refresh annotation is being applied at the moment in the `eventHandler` method. And what I can see is that we are calling `reconcileRebalance` in the KRAO tests during the reconciliation purpose. So the `refresh` annotation is not being set as it is only applied when `eventHandler` method is called which is I guess a part of `createWatch` method. So to make things easier can I apply `refresh` annotation directly and I can comment on it that it is applied as the spec is updated and should be done automatically?

scholzj: Well, as I said, the Mock Kube cluster can work with events as far as I know. So maybe you can use that. Also, if the code is not unit-testable, maybe you should consider some refactoring to make it testable. I'm afraid I do not know the details of how the Kafka Rebalance operator works, so I'm probably not the right one to give you some more detailed instructions. But I do not think running 10 minutes of system tests just because it is not easy to unit test it is the right choice.

cluster-operator/src/test/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceAssemblyOperatorTest.java
```diff
@@ -246,6 +255,55 @@ private void krNewToProposalReady(VertxTestContext context, int pendingCalls, Cr
                 }));
     }
 
+    private void krNotReadyToProposalReadyOnSpecChange(VertxTestContext context, int pendingCalls, CruiseControlEndpoints endpoint, KafkaRebalance kr) throws IOException, URISyntaxException {
```
scholzj, line 258: Why is this not inlined int he test? It seems to be used only by one test anyway. So no code is shared here.

cluster-operator/src/test/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceAssemblyOperatorTest.java
```diff
@@ -1721,6 +1782,21 @@ private KafkaRebalance annotate(KubernetesClient kubernetesClient, String namesp
         return patchedKr;
     }
 
+    /**
+     * Updates the KafkaRebalance spec
+     */
+    private KafkaRebalance updateRebalanceSpec(KubernetesClient kubernetesClient, String namespace, String resource) {
```
scholzj, line 1788: Same as above, why is this not inlined?

systemtest/src/test/java/io/strimzi/systemtest/cruisecontrol/CruiseControlST.java
```diff
@@ -203,6 +203,30 @@ void testCruiseControlWithRebalanceResourceAndRefreshAnnotation(ExtensionContext
...
+                .endMetadata()
+                .build());
+
+        KafkaRebalanceUtils.waitForKafkaRebalanceCustomResourceState(clusterOperator.getDeploymentNamespace(), testStorage.getClusterName(), KafkaRebalanceState.ProposalReady);
+
+        LOGGER.info("Annotating KafkaRebalance: {} with 'approve' anno", testStorage.getClusterName());
+        KafkaRebalanceUtils.annotateKafkaRebalanceResource(new Reconciliation("test", KafkaRebalance.RESOURCE_KIND, clusterOperator.getDeploymentNamespace(), testStorage.getClusterName()), clusterOperator.getDeploymentNamespace(), testStorage.getClusterName(), KafkaRebalanceAnnotation.approve);
+
+        // updated the spec while the resource was in `Rebalancing` state
+        KafkaRebalanceResource.replaceKafkaRebalanceResourceInSpecificNamespace(testStorage.getClusterName(), kafkaRebalance -> kafkaRebalance.getSpec().setReplicationThrottle(100000), clusterOperator.getDeploymentNamespace());
```
scholzj, line 224: How do you know this happens during `Rebalancing` state?

systemtest/src/test/java/io/strimzi/systemtest/cruisecontrol/CruiseControlST.java
```diff
@@ -203,6 +203,30 @@ void testCruiseControlWithRebalanceResourceAndRefreshAnnotation(ExtensionContext
...
+
+        // the resource moved to `ProposalReady` state
+        KafkaRebalanceUtils.waitForKafkaRebalanceCustomResourceState(clusterOperator.getDeploymentNamespace(), testStorage.getClusterName(), KafkaRebalanceState.ProposalReady);
```
scholzj, line 227: Who moves it to ProposalReady? Shouldn't it need another approval after you refreshed it?

ShubhamRwt: No, it doesn't need other refresh. If we update the spec during the rebalancing process, it will just ask for a new Proposal request after stopping the previous one

ShubhamRwt: So the proposal will move to `ProposalReady` state

systemtest/src/test/java/io/strimzi/systemtest/cruisecontrol/CruiseControlST.java
ShubhamRwt: Since we have annotated the resource with `approve` so the state would have moved to `Rebalancing`

scholzj: But how do you know it will not move to Ready before you get to check it?

systemtest/src/test/java/io/strimzi/systemtest/cruisecontrol/CruiseControlST.java
scholzj: Sorry, I mixed Ready and ProposalReady here. You are right, it should get there.

systemtest/src/test/java/io/strimzi/systemtest/cruisecontrol/CruiseControlST.java
ShubhamRwt: Agreed. Let me fix the comment

ShubhamRwt: Fixed

cluster-operator/src/test/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceAssemblyOperatorTest.java
```diff
@@ -246,6 +246,82 @@ private void krNewToProposalReady(VertxTestContext context, int pendingCalls, Cr
...
+     * 2. The operator requests a rebalance proposal through the Cruise Control REST API
+     * 3. The operator receives a "missing hard goals" error instead of a proposal
+     * 4. The KafkaRebalance resource moves to the 'NotReady' state
+     * 5. The rebalance spec is updated with the 'skip hard goals check' field to "true"
+     * 6. The operator requests a rebalance proposal through the Cruise Control REST API
+     * 7. The rebalance proposal is ready on the first call
+     * 8. The KafkaRebalance resource moves to the 'ProposalReady' state
+     */
+    @Test
+    public void krNotReadyToProposalReadyOnSpecChange(VertxTestContext context) throws IOException, URISyntaxException {
```
scholzj, line 263: Maybe the name could start with `test` prefix to make it easier to find? `testKrNotReadyToProposalReadyOnSpecChange`?

cluster-operator/src/test/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceAssemblyOperatorTest.java
```diff
@@ -1721,6 +1798,21 @@ private KafkaRebalance annotate(KubernetesClient kubernetesClient, String namesp
         return patchedKr;
     }
 
+    /**
+     * Updates the KafkaRebalance spec
+     */
+    private KafkaRebalance updateRebalanceSpec(KubernetesClient kubernetesClient, String namespace, String resource) {
```
scholzj, line 1804: This now seems to be unused? Or did I missed it somewhere?

cluster-operator/src/test/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceAssemblyOperatorTest.java
ShubhamRwt: Done

scholzj (approved):
LGTM I guess. Should be reviewed by the SMEs.

api/src/main/java/io/strimzi/api/kafka/model/balancing/KafkaRebalanceState.java
```diff
@@ -21,12 +21,12 @@ public enum KafkaRebalanceState {
      * A proposal has been requested from Cruise Control, but is not ready yet.
      * Transitions to:
      * <dl>
-     *     <dt>PendingProposal</dt><dd>A rebalance proposal is not ready yet.</dd>
+     *     <dt>PendingProposal</dt><dd>A rebalance proposal is not ready yet or strimzi.io/rebalance=refresh is applied and a new proposal is requested.</dd>
```
ppatierno, line 24: Do we need to specify `or strimzi.io/rebalance=refresh is applied and a new proposal is requested` ? I mean, this is the description of the state and it is actually `A rebalance proposal is not ready yet` ... now if it happens because you create the KafkaRebalance or refresh is not part of the state itself.

api/src/main/java/io/strimzi/api/kafka/model/balancing/KafkaRebalanceState.java
```diff
@@ -44,9 +44,10 @@ public enum KafkaRebalanceState {
      *     <dt>Rebalancing</dt><dd>While the actual rebalancing is still ongoing</dd>
      *     <dt>Stopped</dt><dd>If the user sets annotation strimzi.io/rebalance=stop.</dd>
      *     <dt>Ready</dt><dd>Once the rebalancing is complete.</dd>
+     *     <dt>ProposalReady</dt><dd>When annotation strimzi.io/rebalance=refresh is applied on rebalance resource and the proposal is updated.</dd>
```
ppatierno, line 47: question ... if refreshing while Rebalancing, doesn't the FSM pass for ProposalPending? Does it move to ProposalReady directly? I mean ... the FSM depends on the API call to CC, so do you confirm that CC just returns a proposal so it's not pending?

ShubhamRwt: It moves to ProposalReady Directly. I have tested this multiple times and it was directly moving to proposal ready

ppatierno: Ok so CC API is just returning it's ready.

api/src/main/java/io/strimzi/api/kafka/model/balancing/KafkaRebalanceState.java
ShubhamRwt: Yeah, I will fix this

ppatierno (approved):
LGTM

systemtest/src/test/java/io/strimzi/systemtest/cruisecontrol/CruiseControlST.java
```diff
@@ -203,6 +203,30 @@ void testCruiseControlWithRebalanceResourceAndRefreshAnnotation(ExtensionContext
         KafkaRebalanceUtils.doRebalancingProcess(new Reconciliation("test", KafkaRebalance.RESOURCE_KIND, Constants.TEST_SUITE_NAMESPACE, clusterName), Constants.TEST_SUITE_NAMESPACE, clusterName);
     }
 
+    @IsolatedTest
+    void testCruiseControlChangesFromRebalancingtoProposalReadyWhenSpecUpdated(ExtensionContext extensionContext) {
+        TestStorage testStorage = new TestStorage(extensionContext, Constants.TEST_SUITE_NAMESPACE);
+
+        resourceManager.createResourceWithoutWait(extensionContext, KafkaTemplates.kafkaWithCruiseControl(testStorage.getClusterName(), 3, 1).build());
+
+        resourceManager.createResourceWithoutWait(extensionContext,  KafkaRebalanceTemplates.kafkaRebalance(testStorage.getClusterName())
```
im-konge, line 212: I think it should be `createResourceWithWait`

systemtest/src/test/java/io/strimzi/systemtest/cruisecontrol/CruiseControlST.java
```diff
@@ -203,6 +203,30 @@ void testCruiseControlWithRebalanceResourceAndRefreshAnnotation(ExtensionContext
...
+                .editMetadata()
+                .withNamespace(clusterOperator.getDeploymentNamespace())
+                .endMetadata()
+                .build());
```
im-konge, line 216:
```suggestion
                .editMetadata()
                    .withNamespace(clusterOperator.getDeploymentNamespace())
                .endMetadata()
                .build());
```

im-konge (commented):
Just two nits from ST POV, thanks :)

kyguy (approved):
No other feedback from me, great work @ShubhamRwt!

systemtest/src/test/java/io/strimzi/systemtest/cruisecontrol/CruiseControlST.java
ShubhamRwt: for both of them @im-konge ?

im-konge: Yes, thanks

ShubhamRwt: Done. Thanks @im-konge

im-konge (approved):

PR #8743, opened 2023-06-28
Add support for insecure flag in Maven artifacts in Kafka Connect Build

scholzj:
### Type of change

- Enhancement / new feature

### Description

The `insecure` flag in Kafka Connect Build allows downloading artifacts from sources that have some issues with TLS encryption. For example, use self-signed certificates, have expired certificates etc. Currently, this flag is not available for Maven artifacts.

This PR adds support for it to Maven artifacts as well. It overrides the TLS checks for both the `curl` calls as well as for the Maven downloads. It allows users use the Maven artifacts even with for example private Maven repositories signed by 
custom certificates etc.

### Checklist

- [x] Write tests
- [x] Update documentation
- [x] Update CHANGELOG.md


cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaConnectDockerfile.java
```diff
@@ -154,13 +154,35 @@ private void connectorPluginsPreStage(PrintWriter writer, List<Plugin> plugins)
...
+                                        assembleResourceUrl(repo, mvn, "jar"));
+                    }
+
+                    //Cmd cmd = run("curl", "-f", insecureCurlFlag, "-L", "--create-dirs", "--output", "/tmp/" + artifactDir + "/pom.xml", assembleResourceUrl(repo, mvn, "pom"))
+                    //        .andRun("echo", settingsXml).redirectTo(settingsFile) // Create the settings file
+                    //        .andRun("mvn", "dependency:copy-dependencies", "-s", settingsFile,
+                    //                "-DoutputDirectory=/tmp/artifacts/" + artifactDir, insecureMavenFlags, "-f", "/tmp/" + artifactDir + "/pom.xml")
+                    //        .andRun("curl", "-f", insecureCurlFlag, "-L", "--create-dirs", "--output",
+                    //                "/tmp/artifacts/" + artifactDir + "/" + mvn.getArtifact() + "-" + mvn.getVersion() + ".jar",
+                    //                assembleResourceUrl(repo, mvn, "jar"));
```
ppatierno, line 185: remove all this commented part?

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaConnectDockerfile.java
```diff
@@ -154,13 +154,35 @@ private void connectorPluginsPreStage(PrintWriter writer, List<Plugin> plugins)
...
+                                        assembleResourceUrl(repo, mvn, "jar"));
+                    } else {
+                        cmd = run("curl", "-f", "-L", "--create-dirs", "--output", "/tmp/" + artifactDir + "/pom.xml", assembleResourceUrl(repo, mvn, "pom"))
+                                .andRun("echo", settingsXml).redirectTo(settingsFile) // Create the settings file
+                                .andRun("mvn", "dependency:copy-dependencies", "-s", settingsFile,
+                                        "-DoutputDirectory=/tmp/artifacts/" + artifactDir, "-f", "/tmp/" + artifactDir + "/pom.xml")
+                                .andRun("curl", "-f", "-L", "--create-dirs", "--output",
+                                        "/tmp/artifacts/" + artifactDir + "/" + mvn.getArtifact() + "-" + mvn.getVersion() + ".jar",
+                                        assembleResourceUrl(repo, mvn, "jar"));
+                    }
```
ppatierno, line 177: it's a pity that it's not that simple to factor out some common code here, considering that most of the arguments are actually the same. I see that `run` has a String varargs but anyway building a list of String(s) with arguments and then call the `run` just one time could be not that simple. Not saying you should do that, anyway.

ppatierno (approved):

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaConnectDockerfile.java
scholzj: I tried to have originally the code only once. But it is quite complicated with how the escaping has to be done etc. So at the end I decided it is better to just have it twice like that because any attempt to have the command built only once were always much worse (at least in my opinion) than this :-(.

PR #8751, opened 2023-06-29
User operator: leverage CrdOperator and SecretOperator

MikeEdgar:
### Type of change

_Select the type of your PR_

- ~~Bugfix~~
- ~~Enhancement / new feature~~
- Refactoring
- ~~Documentation~~

### Description

Migrate User Operator to use new `CrdOperator` and `SecretOperator` resource classes and reduce direct interaction with the K8s client.

### Checklist

- [ ] Write tests
- [ ] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/user-operator/src/main/java/io/strimzi/operator/user/Main.java
+++ b/user-operator/src/main/java/io/strimzi/operator/user/Main.java
@@ -7,6 +7,7 @@ package io.strimzi.operator.user;
 import io.fabric8.kubernetes.api.model.Secret;
 import io.fabric8.kubernetes.client.KubernetesClient;
 import io.micrometer.core.instrument.MeterRegistry;
+import io.micrometer.core.instrument.binder.MeterBinder;
 import io.micrometer.core.instrument.binder.jvm.ClassLoaderMetrics;
 import io.micrometer.core.instrument.binder.jvm.JvmGcMetrics;
 import io.micrometer.core.instrument.binder.jvm.JvmMemoryMetrics;
@@ -14,6 +15,8 @@ import io.micrometer.core.instrument.binder.jvm.JvmThreadMetrics;
 import io.micrometer.core.instrument.binder.system.ProcessorMetrics;
 import io.micrometer.prometheus.PrometheusConfig;
 import io.micrometer.prometheus.PrometheusMeterRegistry;
+import io.strimzi.api.kafka.KafkaUserList;
+import io.strimzi.api.kafka.model.KafkaUser;
 import io.strimzi.certs.OpenSslCertManager;
 import io.strimzi.operator.common.AdminClientProvider;
 import io.strimzi.operator.common.DefaultAdminClientProvider;
@@ -22,6 +25,8 @@ import io.strimzi.operator.common.MicrometerMetricsProvider;
 import io.strimzi.operator.common.OperatorKubernetesClientBuilder;
 import io.strimzi.operator.common.Util;
 import io.strimzi.operator.common.http.HealthCheckAndMetricsServer;
+import io.strimzi.operator.common.operator.resource.concurrent.CrdOperator;
+import io.strimzi.operator.common.operator.resource.concurrent.SecretOperator;
 import io.strimzi.operator.user.operator.DisabledSimpleAclOperator;
 import io.strimzi.operator.user.operator.KafkaUserOperator;
 import io.strimzi.operator.user.operator.QuotasOperator;
@@ -34,13 +39,16 @@ import org.apache.logging.log4j.Logger;
 import java.security.Security;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
+import java.util.concurrent.ThreadFactory;
 import java.util.concurrent.atomic.AtomicInteger;
+import java.util.function.Supplier;
+import java.util.stream.Stream;
 
 /**
  * The main class of the Strimzi User Operator
  */
 public class Main {
-    private final static Logger LOGGER = LogManager.getLogger(Main.class);
+    private static final Logger LOGGER = LogManager.getLogger(Main.class);
 
     /**
      * Main method which starts the webserver with healthchecks and metrics and the UserController which is responsible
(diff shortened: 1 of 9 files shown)


user-operator/src/main/java/io/strimzi/operator/user/operator/KafkaUserOperator.java
```diff
@@ -106,6 +109,41 @@ public void stop() {
...
+     * @param selectorLabels   Selector which should be matched by the resources
+     * @param resyncIntervalMs The interval in which the resync of the informer
+     *                         should happen in milliseconds
+     *
+     * @return Informer instance
+     *
+     * @see io.strimzi.operator.common.operator.resource.concurrent.AbstractNamespacedResourceOperator#informer(String,
+     *      Map, long)
+     */
+    public SharedIndexInformer<KafkaUser> informer(String namespace, Map<String, String> selectorLabels, long resyncIntervalMs) {
```
MikeEdgar, line 128: An alternative to this and the following proxy method would be to have `KafkaUserOperator` extend `CrdOperator`.

user-operator/src/main/java/io/strimzi/operator/user/Main.java
```diff
@@ -154,13 +172,30 @@ private static Admin createAdminClient(UserOperatorConfig config, KubernetesClie
...
+         *
+         * See https://checkstyle.sourceforge.io/checks/metrics/classdataabstractioncoupling.html
+         */
+        Stream.<Supplier<MeterBinder>>of(
+                ClassLoaderMetrics::new,
+                JvmMemoryMetrics::new,
+                JvmGcMetrics::new,
+                ProcessorMetrics::new,
+                JvmThreadMetrics::new)
+            .forEach(binder -> binder.get().bindTo(registry));
```
scholzj, line 187: TBH, I'm not sure I see the value in this. If you think this code is better, you should just use it. But the comment suggests you think it is a bad code and you do it just to avoid some Checkstyle warning. If that is the case, I think you should just use an annotation to suppress the warning. But this current way seems a bit strange since the comment suggests the code is bad and is there just to avoid checkstyle. 

(I personally would stick with the old way, nut I'm fine with the new one as well)

user-operator/src/main/java/io/strimzi/operator/user/Main.java
```diff
@@ -154,13 +172,30 @@ private static Admin createAdminClient(UserOperatorConfig config, KubernetesClie
...
 
         return new MicrometerMetricsProvider(registry);
     }
+
+    private static class OperatorWorkThreadFactory implements ThreadFactory {
+
```
scholzj, line 193: Empty line not needed here.

user-operator/src/main/java/io/strimzi/operator/user/operator/KafkaUserOperator.java
scholzj, line 128: Why don't you just pass the `CrdOperator` instance directly to `UserController` and `UserControllerLoop`? It seems a bit weird to proxy it through the `KafkaUserOperator` class which does someching else.

user-operator/src/main/java/io/strimzi/operator/user/operator/KafkaUserOperator.java
```diff
@@ -106,6 +109,41 @@ public void stop() {
...
+     *
+     * @param reconciliation Reconciliation marker
+     * @param kafkaUser      Desired resource with the updated status
+     *
+     * @return Future which completes when the status is patched
+     *
+     * @see io.strimzi.operator.common.operator.resource.concurrent.CrdOperator#updateStatusAsync(Reconciliation,
+     *      KafkaUser)
+     */
+    public CompletionStage<KafkaUser> updateStatusAsync(Reconciliation reconciliation, KafkaUser kafkaUser) {
```
scholzj, line 143: Same as above -> it seems weird to proxy it through this instead of using it there directly.

user-operator/src/main/java/io/strimzi/operator/user/operator/KafkaUserOperator.java
```diff
@@ -290,23 +324,32 @@ private void maybeGenerateCredentials(Reconciliation reconciliation, KafkaUserMo
...
+        if (user.isUserWithDesiredPassword()) {
             // User is a SCRAM-SHA-512 user and requested some specific password instead of generating a random password
-            desiredPasswordSecret = client.secrets().inNamespace(reconciliation.namespace()).withName(user.desiredPasswordSecretName()).get();
-            if (desiredPasswordSecret == null) {
-                throw new InvalidResourceException("Secret " + user.desiredPasswordSecretName() + " in namespace " + reconciliation.namespace() + " with requested password not found");
-            }
+            getRequiredSecret(
+                    reconciliation.namespace(),
+                    user.desiredPasswordSecretName(),
+                    () -> new InvalidResourceException(String.format(MISSING_SECRET_MSG, user.desiredPasswordSecretName(), reconciliation.namespace(), "requested password")))
```
scholzj, line 335: Is this really needed? The secrets are mostly self-explanatory out of the context. So I'm not sure whether this hyper-customizes message isn't just overcomplicating the code compared to a simple error message with name and namespace generated in the `getRequiredSecret` method.

MikeEdgar: This was to allow using `getRequiredSecret` for the CA cert/key as well as the SCRAM-SHA desired password without changing the latter's exception message. If it's fine to remove the description part from all three, that would be better.

user-operator/src/main/java/io/strimzi/operator/user/Main.java
MikeEdgar: I guess my comment has a bit of a negative tone to it :-) I'll revert this and suppress the error :+1:

user-operator/src/main/java/io/strimzi/operator/user/Main.java
MikeEdgar: Blank removed

user-operator/src/main/java/io/strimzi/operator/user/operator/KafkaUserOperator.java
MikeEdgar: Done

user-operator/src/main/java/io/strimzi/operator/user/operator/KafkaUserOperator.java
MikeEdgar: Updated to have simpler message (secret name and namespace missing) for all potential exceptions.

scholzj (approved):
LGTM. Thanks for the PR.

ppatierno (approved):
LGTM

scholzj:
Thanks for the PR @MikeEdgar

PR #8756, opened 2023-06-30
docs(topic operator): updates for the preview of the unidirectional topic operator (UTO)

PaulRMellor:
**Documentation**

Updates the documentation for the preview of the unidirectional topic operator (UTO)

- Updates the Overview description of the Topic Operator (with equivalent changes to the User Operator section)
- Adds a description of the UnidirectionalTopicOperator feature gate to Deploying guide
- Updates the Using the Topic Operator section in the Deploying guide, including:
  - new procedure for Managing KafkaTopic resources without impacting Kafka topics
  - New procedure for Enabling topic management for existing Kafka topics
  - New concepts for Deleting managed topics 
  - New instructions for deploying standalone UTO
  - Updated section describing Handling changes to topics
  - Updated section describing Topic naming conventions

NOTE: Sections relating specifically to UTO are flagged with _(Preview)_

### Checklist

- [ ] Write tests
- [x] Update documentation
- [ ] Update CHANGELOG.md


documentation/modules/configuring/con-configuring-topic-operator.adoc
```diff
@@ -7,6 +7,10 @@
 
 [role="_abstract"]
 Use `topicOperator` properties in `Kafka.spec.entityOperator` to configure the Topic Operator.
+
+NOTE: If you are using the preview of unidirectional topic management, the following properties are not used and will be ignored: `zookeeperSessionTimeoutSeconds` and `topicMetadataMaxAttempts`.
```
scholzj, line 11: Should we use full path here to make it more clear what does it relate to?

```suggestion
NOTE: If you are using the preview of unidirectional topic management, the following properties are not used and will be ignored: `Kafka.spec.entityOperator.topicOperator.zookeeperSessionTimeoutSeconds` and `Kafka.spec.entityOperator.topicOperator.topicMetadataMaxAttempts`.
```

Or maybe make it more general as it applies also when deploying the Topic Operator in the standalone mode.

documentation/modules/deploying/proc-deploy-topic-operator-standalone.adoc
```diff
@@ -52,43 +52,43 @@ spec:
...
+            - name: STRIMZI_JAVA_OPTS # <10>
               value: "-Xmx=512M -Xms=256M"
-            - name: STRIMZI_JAVA_SYSTEM_PROPERTIES <11>
+            - name: STRIMZI_JAVA_SYSTEM_PROPERTIES # <11>
               value: "-Djavax.net.debug=verbose -DpropertyName=value"
-            - name: STRIMZI_PUBLIC_CA <12>
+            - name: STRIMZI_PUBLIC_CA # <12>
               value: "false"
-            - name: STRIMZI_TLS_AUTH_ENABLED <13>
+            - name: STRIMZI_TLS_ENABLED # <13>
```
scholzj, line 81: Is this a backwards incompatible change in the TO? Or was it typo in the docs?

documentation/modules/deploying/proc-deploy-topic-operator-standalone.adoc
```diff
@@ -99,9 +99,9 @@ This does not have to be the name of the Kafka cluster.
 It can be the label assigned to the `KafkaTopic` resource.
 If you deploy more than one Topic Operator, the labels must be unique for each.
 That is, the operators cannot manage the same resources.
-<4> The host and port pair of the address to connect to the ZooKeeper cluster.
+<4> (ZooKeeper) The host and port pair of the address to connect to the ZooKeeper cluster.
 This must be the same ZooKeeper cluster that your Kafka cluster is using.
-<5> The ZooKeeper session timeout, in milliseconds.
+<5> (ZooKeeper) The ZooKeeper session timeout, in milliseconds.
 The default is `18000` (18 seconds).
```
scholzj, line 105: Should we mention here the options ignored in Unidirectional TO as well?

documentation/modules/deploying/proc-deploy-topic-operator-standalone.adoc
```diff
@@ -175,3 +175,49 @@ strimzi-topic-operator  1/1    1           1
...
+Other unused environment variables that can be removed if present:
++
+* `STRIMZI_REASSIGN_THROTTLE`
+* `STRIMZI_REASSIGN_VERIFY_INTERVAL_MS`
+* `STRIMZI_TOPIC_METADATA_MAX_ATTEMPTS`
+* `STRIMZI_TOPICS_PATH`
+* `STRIMZI_STORE_TOPIC`
+* `STRIMZI_STORE_NAME`
+* `STRIMZI_APPLICATION_ID`
+* `STRIMZI_STALE_RESULT_TIMEOUT_MS`
```
scholzj, line 208: Why the 4 env vars above _should be removed_ and these only _can be removed_? Not sure I follow what makes them different.

documentation/modules/deploying/proc-deploy-topic-operator-standalone.adoc
```diff
@@ -175,3 +175,49 @@ strimzi-topic-operator  1/1    1           1
...
+----
+# ...
+env:
+  - name: STRIMZI_USE_FINALIZERS
+    value: "true"
+----
++
+Use this environment variable if you want to use finalizers to control xref:con-deleting-managed-topics-{context}[topic deletion].
+
+. Deploy the standalone Topic Operator in the standard way.
```
scholzj, line 223: I assume that there will be also some change to the startup script needed? Or some additional environment variable?

documentation/modules/deploying/proc-deploy-topic-operator-standalone.adoc
```diff
@@ -175,3 +175,49 @@ strimzi-topic-operator  1/1    1           1
...
+.Additional configuration for unidirectional topic management
+[source,shell,subs=+quotes]
+----
+# ...
+env:
+  - name: STRIMZI_USE_FINALIZERS
+    value: "true"
+----
++
+Use this environment variable if you want to use finalizers to control xref:con-deleting-managed-topics-{context}[topic deletion].
```
scholzj, line 221: Doesn't this default to `true`? So, should it talk about disabling it instead?

documentation/modules/operators/con-application-topic-handling.adoc
```diff
@@ -0,0 +1,38 @@
...
+The type of update depends on the xref:ref-operator-topic-{context}[mode of topic management].
+
+* For bidirectional topic management, configuration changes are synchronized between the Kafka topic and the `KafkaTopic` resource. Incompatible changes prioritize the Kafka configuration, and the `KafkaTopic` resource is adjusted accordingly.
+* For unidirectional topic management (currently in preview), configuration changes only go in one direction: from the `KafkaTopic` resource to the Kafka topic. Any changes to a Kafka topic managed outside the `KafkaTopic` resource are reverted. 
+
+== Automatic creation of topics
+
+Applications can trigger the automatic creation of topics in the Kafka cluster.
+If the Kafka broker configuration `auto.create.topics.enable` is set to `true`, it enables automatic topic creation by the Kafka broker when an application tries to produce or consume from a non-existing topic.   
+It's possible that this might trigger the creation of topics in the cluster before the Topic Operator can react.
```
scholzj, line 31: It is actually triggered _very_ automatically, without doing anything special in the application- Also, `auto.create.topics.enable` is enabled by default. I think the way this is worded gives the impression this needs to be enabled which can be misleading.

documentation/modules/operators/con-application-topic-handling.adoc
```diff
@@ -0,0 +1,38 @@
...
+
+[id='con-application-topic-handling-{context}']
+= Handling changes to topics
+
+[role="_abstract"]
+The Topic Operator is capable of managing changes to topics when there is no single source of truth. 
+The `KafkaTopic` resource and the Kafka topic can undergo independent modifications, where real-time observation of changes may not always be feasible, particularly when the Topic Operator is not operational.
+To handle this, the Topic Operator maintains a topic store that stores relevant information about each topic. 
+It compares the state of the Kafka cluster and Kubernetes with the topic store to determine the necessary changes for synchronization. 
+This evaluation takes place during startup and at regular intervals while the Topic Operator is active.
```
scholzj, line 13: Is the topic store used in UTO? In the previous file, you suggest the related environment variables can be removed.

documentation/modules/operators/con-operator-topic-names.adoc
```diff
@@ -0,0 +1,71 @@
...
+kind: KafkaTopic
+metadata:
+  name: topic-name-1
+  labels:
+    strimzi.io/cluster: my-cluster
+spec:
+  topicName: topic-name-1  
+----
+
+The cluster name is derived from the `Kafka` resource.
```
scholzj, line 23: It is not derived from it -> it is the name of the Kafka resource.

documentation/modules/operators/con-operator-topic-names.adoc
```diff
@@ -0,0 +1,71 @@
...
+If the label does not match the Kafka cluster, the Topic Operator cannot identify the `KafkaTopic`, and the topic is not created.
```
scholzj, line 24:
```suggestion
If the label does not match the Kafka cluster, the Topic Operator will not see the `KafkaTopic`, and the topic is not created.
```

scholzj, line 24: The cluster label is IMHO used when the TO is deployed using the KAfka resource. But Standalone TO might not use it. Might be worth mentioning?

documentation/modules/operators/con-operator-topic-names.adoc
```diff
@@ -0,0 +1,71 @@
...
+apiVersion: {KafkaTopicApiVersion}
+kind: KafkaTopic
+metadata:
+  name: My.Topic.1 # <1>
+spec:
+  topicName: my-topic-1 # <2>
+  # ...
+----
+<1> Kafka topic name uses upper case and periods, which are invalid in Kubernetes.
+<2> A topic name alias that works in Kubernetes.
```
scholzj, line 53: Isn't this the wrong way around? The `.metadata.name` has to be a valid Kubernetes name. `.spec.topicName` can contain the actualKafka name.

documentation/modules/operators/con-operator-topic-names.adoc
```diff
@@ -0,0 +1,71 @@
...
+
+If more than one `KafkaTopic` resource refers to the same Kafka topic, the resource that was created first is considered to be the one managing the topic. 
+The status of the newer resources is updated to indicate a conflict, and their status is changed to `False`.
+
+If a Kafka client application, such as Kafka Streams, automatically creates topics with invalid Kubernetes resource names, the Topic Operator generates a valid `metadata.name`. 
+It replaces invalid characters and appends a hash to the name.
```
scholzj, line 59: Only on BTO -> UTO will not create the KafkaTopic resource for it

documentation/modules/operators/ref-operator-cluster-feature-gates.adoc
```diff
@@ -123,4 +123,20 @@ Examples of the `KafkaNodePool` resources can be found in the xref:config-exampl
...
+The `UnidirectionalTopicOperator` feature gate introduces a unidirectional topic management mode for creating Kafka topics using the `KafkaTopic` resource.
+Unidirectional mode is compatible with using KRaft for cluster management.
+With unidirectional mode, you create Kafka topics using the `KafkaTopic` resource, which are then managed by the Topic Operator.
+Any configuration changes to a topic outside the `KafkaTopic` resource are reverted.
+For more information on topic management, see xref:ref-operator-topic-str[].
+
+.Enabling the UnidirectionalTopicOperator feature gate
+
+To enable the `UnidirectionalTopicOperator` feature gate, specify `+UnidirectionalTopicOperator` in the `STRIMZI_FEATURE_GATES` environment variable in the Cluster Operator configuration.
+The `KafkaTopic` custom resource must also have the annotation `strimzi.io/managed: true` (default).
```
scholzj, line 142: If it is default, do they really `must have this annotation`?

documentation/modules/overview/con-overview-components-topic-operator.adoc
```diff
@@ -6,29 +6,30 @@
...
+The Topic Operator maintains information about each topic in a _topic store_, which is continually synchronized with updates from Kubernetes `KafkaTopic` custom resources or Kafka topics.
+Updates from operations applied to a local in-memory topic store are persisted to a backup topic store on disk.
+
+The Topic Operator operates in the following modes: 
+
+Bidirectional mode:: Bidirectional mode means that the Topic Operator can reconcile changes to a `KafkaTopic` resource to and from a Kafka cluster.
+This means that you can update topics either through the `KafkaTopic` resource or directly in Kafka, and the Topic Operator will ensure that both sources are updated to reflect the changes. This mode can be used with ZooKeeper cluster management. 
+Unidirectional mode (preview):: Unidirectional means that the Topic Operator solely manages topics through the `KafkaTopic` resource. This mode can be used with KRaft cluster management.
+
+NOTE: Unidirectional mode and KRaft cluster management are available to try by enabling the `UnidirectionalTopicOperator` and `UseKRaft` feature gates.
```
scholzj, line 31: I do not think Tom's PR enables this. So this would need an additional follow-up PR. So I wonder if we want to exclude it for the time being to avoid having it merged by mistake.

documentation/modules/deploying/proc-deploy-topic-operator-standalone.adoc
PaulRMellor: My mistake. I confused it with the other variable. This went in as part of this change: https://github.com/strimzi/strimzi-kafka-operator/pull/5201
Reverted.

documentation/modules/deploying/proc-deploy-topic-operator-standalone.adoc
PaulRMellor: I don't think UTO works if the ZK variables are present. I assumed the other variables would be ignored if present.   I'll see what Tom says on this split

documentation/modules/deploying/proc-deploy-topic-operator-standalone.adoc
PaulRMellor: Instead of this, I've added an example config for the UTO.

documentation/modules/deploying/proc-deploy-topic-operator-standalone.adoc
PaulRMellor: I'm assuming the zookeeper variables make the difference, as with change in the [topic_operator_run.sh](https://github.com/strimzi/strimzi-kafka-operator/pull/8627/files#diff-ea49bbb821c918d163e90ec3da950dffae099d3871e3983d7792a2f45584bf96):

_If ZooKeeper connection details are provided then use the old bidirectional TO
Otherwise use the new unidirectional TO_

documentation/modules/deploying/proc-deploy-topic-operator-standalone.adoc
PaulRMellor: Not sure that's the case for standalone preview.  I'll see what Tom says

documentation/modules/operators/con-application-topic-handling.adoc
PaulRMellor: Changed as follows:

> Applications can trigger the automatic creation of topics in the Kafka cluster.
> By default, the Kafka broker configuration `auto.create.topics.enable` is set to `true`, allowing the broker to create topics automatically when an application attempts to produce or consume from a non-existing topic.  
> Automatic topic creation in the cluster can happen before the Topic Operator can react.

documentation/modules/operators/con-application-topic-handling.adoc
PaulRMellor: Yes, it's not used. I've moved the content around and added a sub-heading to make this section a lot clearer.

> [role="_abstract"]
> How the Topic Operator handles changes to topics depends on the xref:ref-operator-topic-{context}[mode of topic management].
> 
> * For bidirectional topic management, configuration changes are synchronized between the Kafka topic and the `KafkaTopic` resource. Incompatible changes prioritize the Kafka configuration, and the `KafkaTopic` resource is adjusted accordingly.
> * For unidirectional topic management (currently in preview), configuration changes only go in one direction: from the `KafkaTopic` resource to the Kafka topic. Any changes to a Kafka topic managed outside the `KafkaTopic` resource are reverted. 
> 
> == Topic store for bidirectional topic management
> 
> For bidirectional topic management, the Topic Operator is capable of handling changes to topics when there is no single source of truth....

documentation/modules/operators/con-operator-topic-names.adoc
PaulRMellor: Right. I've added a note:

> NOTE: The label is only used when deploying the Topic Operator with the `Kafka` resource.
> When deploying the standalone Topic Operator, you specify a connection to a Kafka cluster instead.

documentation/modules/operators/con-operator-topic-names.adoc
PaulRMellor: Yes! I've updated and tried to make clearer.

> The `KafkaTopic` resource allows you to specify the Kafka topic name using the `metadata.name` field. 
> However, if the desired Kafka topic name is not a valid Kubernetes resource name, you can use the `spec.topicName` property to specify the actual name. 
> The `spec.topicName` field is optional, and when it's absent, the Kafka topic name defaults to the `metadata.name` of the topic.
> When a topic is created, the topic name cannot be changed later.
> 
> .Example of supplying a valid Kafka topic name
> [source,yaml,subs="attributes+"]
> apiVersion: {KafkaTopicApiVersion}
> kind: KafkaTopic
> metadata:
>   name: my-topic-1 # <1>
> spec:
>   topicName: My.Topic.1 # <2>
> <1> A valid topic name that works in Kubernetes.
> <2> A Kafka topic name that uses upper case and periods, which are invalid in Kubernetes.

documentation/modules/operators/ref-operator-cluster-feature-gates.adoc
PaulRMellor: I've changed as follows:

> For the `KafkaTopic` custom resource to use this feature, the `strimzi.io/managed` annotation is set to `true` by default.

documentation/modules/operators/con-operator-topic-names.adoc
PaulRMellor: Updated:

> If a Kafka client application, such as Kafka Streams, automatically creates topics with invalid Kubernetes resource names, the Topic Operator generates a valid `metadata.name` when used in bidirectional mode. 
> It replaces invalid characters and appends a hash to the name. 
> However, this behavior does not apply in (preview) unidirectional mode.

documentation/modules/overview/con-overview-components-topic-operator.adoc
PaulRMellor: Right. I've removed from a  couple of places that suggest the support is there in the preview.

documentation/modules/deploying/proc-deploy-topic-operator-standalone.adoc
scholzj: I think we should clarify it in some dedicated paragraph. Because it is not completely clear that it is the presence or absence of the ZK variables which defines which operator is used. The Feature Gates as Tom uses it do not propagate into the Topic Operator. So for standalone TO users this is important information.

tombentley: For the standalone TO it is the absence of the ZK env vars that determines which operator will be used. For the entity topic operator the +UnidirectionalTopicOperator feature gate determines which env vars it will be deployed with.

documentation/modules/operators/con-application-topic-handling.adoc
```diff
@@ -0,0 +1,40 @@
...
+Applications can trigger the automatic creation of topics in the Kafka cluster.
+By default, the Kafka broker configuration `auto.create.topics.enable` is set to `true`, allowing the broker to create topics automatically when an application attempts to produce or consume from a non-existing topic.  
+Automatic topic creation in the cluster can happen before the Topic Operator can react.
+
+For bidirectional topic management, the Topic Operator synchronizes the changes between the topics and `KafkaTopic` resources.
+
+If you are trying the unidirectional topic management preview, this can mean that the topics created for an application deployment are initially created with default topic configuration.
+If the Topic Operator attempts to reconfigure the topics based on `KafkaTopic` resource specifications included with the application deployment, the operation might fail because the required change to the configuration is not allowed.
+For example, if the change means lowering the number of topic partitions.
+For this reason, it is recommended to disable `auto.create.topics.enable` in the Kafka cluster configuration when using unidirectional topic management.
```
tombentley, line 40: This might be too much detail...

```suggestion
For this reason, it is recommended to disable `auto.create.topics.enable` in the Kafka cluster configuration when using unidirectional topic management.
This might result in the Kafka application starting before the Topic Operator has created the topic in Kafka.
Typically a Kafka client would automatically retry if it encountered an `UNKNOWN_TOPIC_OR_PARTITION` error, but even if it crashes and was running as a `Deployment` it will be restarted by Kubernetes and so eventually recover.
```

documentation/modules/operators/con-deleting-managed-topics.adoc
```diff
@@ -0,0 +1,56 @@
...
+  labels:
+    strimzi.io/cluster: my-cluster
+  deletionTimestamp: 20230301T000000.000  
+----
+
+The resource is still present.
+If the deletion fails, it is shown in the status of the resource.
+
+When the finalization tasks are successfully executed, the finalizer is removed from the metadata, and the resource is fully deleted.  
+
```
tombentley, line 56: We should note here that the use of finalizers can prevent other resources from being deleted. For example the namespace that contains a resource that has a finalizer cannot be deleted until the `metadata.finalizers` is empty. This would be the case if the topic operator was not running, or was not able to delete the topic in Kafka.

documentation/modules/operators/con-operator-topic-names.adoc
```diff
@@ -0,0 +1,75 @@
...
+    strimzi.io/cluster: my-cluster
+spec:
+  topicName: topic-name-1  
+----
+
+The label provides the cluster name of the `Kafka` resource.
+If the label does not match the Kafka cluster, the Topic Operator cannot see the `KafkaTopic`, and the topic is not created.
+
+NOTE: The label is only used when deploying the Topic Operator with the `Kafka` resource.
+When deploying the standalone Topic Operator, you specify a connection to a Kafka cluster instead.
```
tombentley, line 27: This is wrong. Labels are not a mechanism for determining where the operator connects. They are a mechanism for determining which `KafkaTopics` it tries to consume. So it's still possible to use labels with the standalone operator.

documentation/modules/operators/con-operator-topic-names.adoc
```diff
@@ -0,0 +1,75 @@
...
+  name: my-topic-1 # <1>
+spec:
+  topicName: My.Topic.1 # <2>
+  # ...
+----
+<1> A valid topic name that works in Kubernetes.
+<2> A Kafka topic name that uses upper case and periods, which are invalid in Kubernetes.
+
+If more than one `KafkaTopic` resource refers to the same Kafka topic, the resource that was created first is considered to be the one managing the topic. 
+The status of the newer resources is updated to indicate a conflict, and their status is changed to `False`.
```
tombentley, line 59:
```suggestion
The status of the newer resources is updated to indicate a conflict, and their `Ready` status is changed to `False`.
```

documentation/modules/operators/proc-converting-non-managed-topics.adoc
```diff
@@ -0,0 +1,86 @@
...
+NOTE: Unidirectional topic management is available as a preview. 
+Unidirectional topic management is not enabled by default, so you must xref:ref-operator-unidirectional-topic-operator-feature-gate-{context}[enable the `UnidirectionalTopicOperator` feature gate] to be able to use it.
+
+.Prerequisites
+
+* xref:deploying-cluster-operator-str[The Cluster Operator must be deployed.]
+
+.Procedure
+
+. Create a `KafkaTopic` resource with a `metadata.name` that is the same as the Kafka topic.
```
tombentley, line 23: Or use `spec.topicName` is the name of the topic in Kafka would not be a legal Kube resource name.

documentation/modules/overview/con-overview-components-topic-operator.adoc
```diff
@@ -6,29 +6,28 @@
...
+You can declare a `KafkaTopic` as part of your application's deployment and the Topic Operator manages the Kafka topic for you.
+
+The Topic Operator maintains information about each topic in a _topic store_, which is continually synchronized with updates from Kubernetes `KafkaTopic` custom resources or Kafka topics.
+Updates from operations applied to a local in-memory topic store are persisted to a backup topic store on disk.
+
+The Topic Operator operates in the following modes: 
+
+Bidirectional mode:: Bidirectional mode means that the Topic Operator can reconcile changes to a `KafkaTopic` resource to and from a Kafka cluster.
+This means that you can update topics either through the `KafkaTopic` resource or directly in Kafka, and the Topic Operator will ensure that both sources are updated to reflect the changes. This mode requires ZooKeeper for cluster management. 
+Unidirectional mode (preview):: Unidirectional means that the Topic Operator solely manages topics through the `KafkaTopic` resource. This mode does not require ZooKeeper.
```
tombentley, line 29: Move this up, before the para starting on line 22.

documentation/modules/overview/con-overview-components-topic-operator.adoc
```diff
@@ -6,29 +6,28 @@
...
-Specifically, if a `KafkaTopic` is:
+When a `KafkaTopic` is created, deleted, or changed, the Topic Operator performs the corresponding action on the Kafka topic.
 
-* Created, the Topic Operator creates the topic
-* Deleted, the Topic Operator deletes the topic
-* Changed, the Topic Operator updates the topic
+You can declare a `KafkaTopic` as part of your application's deployment and the Topic Operator manages the Kafka topic for you.
+
+The Topic Operator maintains information about each topic in a _topic store_, which is continually synchronized with updates from Kubernetes `KafkaTopic` custom resources or Kafka topics.
+Updates from operations applied to a local in-memory topic store are persisted to a backup topic store on disk.
```
tombentley, line 23: Move this down below the para ending on lien 29, and note that topic stores are only used in bidirectional mode.

tombentley (commented):
Left a few nits, but this looks good. Thanks @PaulRMellor!

documentation/modules/deploying/proc-deploy-topic-operator-standalone.adoc
PaulRMellor: Updated with:

> It is the presence or absence of the ZooKeeper variables that defines whether the unidirectional Topic Operator is used.
> Unidirectional topic management does not use ZooKeeper.
> If ZooKeeper environment variables are not present, the unidirectional Topic Operator is used.
> Otherwise, the bidirectional Topic Operator is used.

documentation/modules/operators/con-application-topic-handling.adoc
```diff
@@ -0,0 +1,40 @@
...
+It recognizes that the `KafkaTopic` was created after its last operation. 
+Consequently, the Topic Operator generates the corresponding Kafka topic and saves the metadata in the topic store.
+
+The topic store enables the Topic Operator to manage situations where the topic configuration is altered in both Kafka topics and `KafkaTopic` resources, as long as the changes are compatible.
+When Kafka topic configuration is updated or changes are made to the `KafkaTopic` custom resource, the topic store is updated after reconciling with the Kafka cluster, as long as the changes are compatible.
+
+== Automatic creation of topics
+
+Applications can trigger the automatic creation of topics in the Kafka cluster.
+By default, the Kafka broker configuration `auto.create.topics.enable` is set to `true`, allowing the broker to create topics automatically when an application attempts to produce or consume from a non-existing topic.
```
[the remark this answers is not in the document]
PaulRMellor: changed as follows:

> Applications can trigger the automatic creation of topics in the Kafka cluster.
> By default, the Kafka broker configuration `auto.create.topics.enable` is set to `true`, allowing the broker to create topics automatically when an application attempts to produce or consume from a non-existing topic.  
> Applications might also use the Kafka `AdminClient` to automatically create topics.
> When an application is deployed along with its `KafkaTopic` resources, it is possible that automatic topic creation in the cluster happens before the Topic Operator can react to the `KafkaTopic`.

documentation/modules/operators/con-application-topic-handling.adoc
PaulRMellor: Yeah. Not sure we need to add this. Let's leave it out.

documentation/modules/operators/con-operator-topic-names.adoc
PaulRMellor: changed as follows:

> The label provides the cluster name of the `Kafka` resource.
> The Topic Operator uses the label as a mechanism for determining which `KafkaTopic` resources to manage.
> If the label does not match the Kafka cluster, the Topic Operator cannot see the `KafkaTopic`, and the topic is not created.

documentation/modules/deploying/proc-deploy-topic-operator-with-cluster-operator.adoc
```diff
@@ -7,6 +7,12 @@
 
 [role="_abstract"]
 This procedure describes how to deploy the Topic Operator using the Cluster Operator.
+The Topic Operator can be deployed for use in either bidirectional mode or unidirectional mode.
+In unidirectional mode, topic management is exclusively handled through `KafkaTopic` resources.
```
scholzj, line 11: I think this is a bit confusing. The Topic Operator manages it only through KafkaTopic resources. But users can manage some topics using the Kafka Admin API as well. So I think I would remove this sentence.

documentation/modules/operators/con-application-topic-handling.adoc
```diff
@@ -0,0 +1,81 @@
...
+When upgrading to Strimzi {ProductVersion}, the transition to Topic Operator control of the topic store is seamless.
+Metadata is found and migrated from ZooKeeper, and the old store is deleted.
+
+== Downgrading to a Strimzi version that uses ZooKeeper to store topic metadata
+
+If you are reverting back to a version of Strimzi earlier than 0.22, which uses ZooKeeper for the storage of topic metadata,
+you still downgrade your Cluster Operator to the previous version,
+then downgrade Kafka brokers and client applications to the previous Kafka version as standard.
+
+However, you must also delete the topics that were created for the topic store using a `kafka-admin` command, specifying the bootstrap address of the Kafka cluster.
```
scholzj, line 57:
```suggestion
However, you must also delete the topics that were created for the topic store using a `kafka-topics.sh` command, specifying the bootstrap address of the Kafka cluster.
```

documentation/modules/operators/con-topic-replication.adoc
```diff
@@ -0,0 +1,43 @@
+// Module included in the following assemblies:
+//
+// assembly-using-the-topic-operator.adoc
+
+[id='ref-topic-replication-{context}']
+= Configuring topics for replication and scaling
```
scholzj, line 6: I think scaling is not the right term here since the partition counts might have many different roles.

```suggestion
= Configuring topics for replication and number of partitions
```

scholzj (commented):
Some more nits ...

scholzj (approved):

documentation/modules/operators/con-deleting-managed-topics.adoc
```diff
@@ -0,0 +1,58 @@
...
+Finalizers also prevent related resources from being deleted. 
+For example, the namespace that contains a `KafkaTopic` resource that has a finalizer cannot be deleted until `metadata.finalizer` is empty.
```
tombentley, line 58: I wonder if we want to be a bit more explicit here. Something like:


> This means that if the unidirectional topic operator is not running there is nothing to remove the `metadata.finalizer`. An attempt to delete the containing namespace won't complete until either the operator is restarted, or the finalizer is otherwise removed (for example using `kubectl edit`).

That way we both tell them how this might manifest and what options they have to fix the problem.

documentation/modules/operators/con-topic-replication.adoc
```diff
@@ -0,0 +1,43 @@
...
+    min.insync.replicas: 2 # <3>
+  #...
+----
+<1> The number of partitions for the topic.
+<2> The number of replica topic partitions. Currently, this cannot be changed in the `KafkaTopic` resource, but it can be changed using the `kafka-reassign-partitions.sh` tool.
+<3> The minimum number of replica partitions that a message must be successfully written to, or an exception is raised.
+
+NOTE: In-sync replicas are used in conjunction with the `acks` configuration for producer applications.
+The `acks` configuration determines the number of follower partitions a message must be replicated to before the message is acknowledged as successfully received.
+The Topic Operator runs with `acks=all`, whereby messages must be acknowledged by all in-sync replicas.
```
tombentley, line 31: I think this is misleading. While it's true for how the topic store mechanism works for the BTO it could be easily misinterpreted by a novice user to mean "when you use the TO your applications will be using `acks=all`, which is not the case.

```suggestion
The bidirectional Topic Operator runs with `acks=all` for its internal topics whereby messages must be acknowledged by all in-sync replicas.
```

tombentley (approved):
Thanks Paul, a couple of suggestions but LGTM otherwise.

PR #8758, opened 2023-06-30
Bump up test container to 0.104.0 version

see-quick:
### Type of change

- Enhancement / new feature

### Description

This PR updates the version of the test container to the recently released (i.e., 0.104.0), which brings Kafka 3.5.0 and 3.4.1. 

### Checklist


scholzj (commented):
Do we need to change the Kafka version somewhere as well to test with 3.5.0?

see-quick:
> Do we need to change the Kafka version somewhere as well to test with 3.5.0?

No, it is tested by default with the latest version of Kafka, which the test container has (i.e., 3.5.0). So there is no need to change anything.

scholzj (approved):
Thanks for the work.

PR #8759, opened 2023-06-30
Add support for configuring node IDs for Kafka Node Pools

### LINKED ISSUE #8590: [NodePools] Add advanced features for managing node IDs in node pools
The initial Node Pool PR #8551 do not contain any of the advanced features for a more detailed configuration of the different node IDs assigned to the brokers in different node pools. This includes among other things features such as:

* assigning various node ID ranges to different node pools
* configuring the next IDs which should be used for a new broker
* configuring which broker should be removed at scale down
* (more situations are described in the proposal)

scholzj:
### Type of change

- Enhancement / new feature

### Description

This PR adds support for advanced management of the node IDs of brokers belonging to different Kafka Node Pools.

It introduces two annotations:
* `strimzi.io/next-node-ids`
* `strimzi.io/remove-node-ids`

These annotations (when set) are used to determine the next node ID (broker ID) when a node is being added or removed from the KafkaNodePool. They have a format like this: `[0, 1, 2, 10-20, 30]`. It allows to specify either individual node IDs (e.g. `5`) or ID ranges (e.g. `1000-1999`). When scaling up, the operator will go through the node IDs from the `strimzi.io/next-node-ids` annotation and pick the first available from it and use it for a new broker. During a scale-up, it will go through the IDs from the `strimzi.io/remove-node-ids` annotation and remove the broker(s) with the first used node ID.

When all IDs from the `strimzi.io/next-node-ids` annotation are already used or when no IDs from the `strimzi.io/remove-node-ids` annotation are used, the operator will follow its usual algorithm and remove the highest available node ID when scaling down or the lowest available when scaling up.

The annotations are used only during scaling events. In order to give users some _early feedback_ in case they misformatted the annotation, they will be also validated in every reconciliation and a warning will be thrown in case the format is wrong. This should help to detect it early and fix it. If it is not fixed when some scaling happens, the operator will ignore them and proceed with the scale-up / scale-down.

When the annotations are not set, the operator will pick up the lowest available node ID starting from 0 for scale-up or the highest used node ID for scale-down.

Users are in general expected to use this feature in two ways:
1) Give each node pool a range of IDs. E.g. 0-999 for `pool-a` and 1000-1999 for `pool-b` and use IDs from these ranges for the brokers.
2) Use these annotations to scale down / scale up with a particular node ID. E.g. use annotation `strimzi.io/remove-node-ids: [ 2 ]` to remove the second broker from an existing cluster.

Documentation will be updated in a separate PR.

This should resolve #8590.

### Checklist

- [x] Write tests

### DIFF
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/model/nodepools/NodeIdAssignor.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/model/nodepools/NodeIdAssignor.java
@@ -261,8 +261,8 @@ public class NodeIdAssignor {
     }
 
     /**
-     * Finds the next node ID from a range which is not in use. The IDs are found from range specified in the Node Pool
-     * annotation strimzi.io/next-node-ids.
+     * Finds the next node ID which should be removed from the existing node IDs. The IDs are found from range specified
+     * in the Node Pool annotation strimzi.io/remove-node-ids.
      *
      * @param desired           Set with currently used IDs
      * @param removeIdRange     The NodeIdRange which will pick up the next node ID
@@ -299,7 +299,7 @@ public class NodeIdAssignor {
         try {
             return removeNodeIds == null ? null : new NodeIdRange(removeNodeIds);
         } catch (NodeIdRange.InvalidNodeIdRangeException e) {
-            LOGGER.warnCr(reconciliation, "Failed to use " + Annotations.ANNO_STRIMZI_IO_NEXT_NODE_IDS + " " + removeNodeIds + ". Nodes will be scaled down from the highest node ID.", e);
+            LOGGER.warnCr(reconciliation, "Failed to use " + Annotations.ANNO_STRIMZI_IO_REMOVE_NODE_IDS + " " + removeNodeIds + ". Nodes will be scaled down from the highest node ID.", e);
             return null;
         }
     }
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/model/nodepools/NodeIdAssignor.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/model/nodepools/NodeIdAssignor.java
@@ -5,35 +5,43 @@
 package io.strimzi.operator.cluster.model.nodepools;
 
 import io.strimzi.api.kafka.model.nodepool.KafkaNodePool;
+import io.strimzi.operator.common.Annotations;
 import io.strimzi.operator.common.InvalidConfigurationException;
+import io.strimzi.operator.common.Reconciliation;
+import io.strimzi.operator.common.ReconciliationLogger;
 
 import java.util.Arrays;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.NavigableSet;
 import java.util.TreeSet;
 
 /**
  * Generates the Node ID assignments from / for node pools
  */
 public class NodeIdAssignor {
+    private static final ReconciliationLogger LOGGER = ReconciliationLogger.create(NodePoolUtils.class.getName());
+
     private final Map<String, NodeIdAssignment> assignments = new HashMap<>();
 
     /**
      * Constructs the NodeIdAssignor instance.
      *
-     * @param nodePools     List of KafkaNodePool resources including their statuses
+     * @param reconciliation    Reconciliation marker
+     * @param nodePools         List of KafkaNodePool resources including their statuses
      */
-    /* test */ NodeIdAssignor(KafkaNodePool... nodePools) {
-        this(Arrays.asList(nodePools));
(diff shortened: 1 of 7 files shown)


scholzj:
(The Regression test failures are not related to this PR)

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/nodepools/NodeIdAssignor.java
```diff
@@ -116,32 +133,175 @@ public NodeIdAssignment assignmentForPool(String poolName)   {
...
+        if (removeId == null) {
+            removeId = desired.pollLast();
+        }
+
+        return removeId;
+    }
+
+    /**
+     * Finds the next node ID from a range which is not in use. The IDs are found from range specified in the Node Pool
+     * annotation strimzi.io/next-node-ids.
```
ppatierno, line 265: I think this Javadoc is wrong. It has to refer the id to be removed or? So even the `strimzi.io/remove-node-ids` annotation.

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/nodepools/NodeIdAssignor.java
```diff
@@ -116,32 +133,175 @@ public NodeIdAssignment assignmentForPool(String poolName)   {
...
+     *
+     * @return  The NodeIdRange based on the annotation or null when the annotation is not set or is invalid.
+     */
+    private NodeIdRange removeIdRange(Reconciliation reconciliation, KafkaNodePool pool)   {
+        String removeNodeIds = Annotations.stringAnnotation(pool, Annotations.ANNO_STRIMZI_IO_REMOVE_NODE_IDS, null);
+
+        try {
+            return removeNodeIds == null ? null : new NodeIdRange(removeNodeIds);
+        } catch (NodeIdRange.InvalidNodeIdRangeException e) {
+            LOGGER.warnCr(reconciliation, "Failed to use " + Annotations.ANNO_STRIMZI_IO_NEXT_NODE_IDS + " " + removeNodeIds + ". Nodes will be scaled down from the highest node ID.", e);
```
ppatierno, line 302:
```suggestion
            LOGGER.warnCr(reconciliation, "Failed to use " + Annotations.ANNO_STRIMZI_IO_REMOVE_NODE_IDS + " " + removeNodeIds + ". Nodes will be scaled down from the highest node ID.", e);
```

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/nodepools/NodeIdRange.java
```diff
@@ -0,0 +1,200 @@
...
+            } else if (RANGE_OF_NODE_IDS_PATTERN.matcher(rangeItem).matches())  {
+                String[] startAndEnd = rangeItem.split("\\s*-\\s*");
+
+                if (startAndEnd.length != 2)    {
+                    throw new InvalidNodeIdRangeException("Invalid Node ID range (exactly 2 items must be found): " + rangeItem);
+                }
+
+                try {
+                    start = Integer.parseInt(startAndEnd[0]);
+                    end = Integer.parseInt(startAndEnd[1]);
```
ppatierno, line 131: should we raise an exception if the user specify not consistent start and end? for example, [100-20] instead of [20-100] so using a `start` greater than `end`?

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/nodepools/NodeIdRange.java
```diff
@@ -0,0 +1,200 @@
...
+                    // Start is the end => return start and set it to null since we will be empty soon
+                    Integer returnValue = start;
+                    start = null;
+                    end = null;
+                    return returnValue;
+                } else if (start > end) {
+                    // Start is bigger than end => return start and decrement it (the range is going down)
+                    Integer returnValue = start;
+                    start--;
+                    return returnValue;
```
ppatierno, line 166: related to the previous question, but how much would it be clear to user that in case of, for example, [100-20] it's using 100 and then becoming [99-20]. I mean it's a smart move because anyway you are using the right range, from 20 to 100, but not sure how much it would be clear, instead of just pushing the user to declare a correct [20-100].

scholzj: Please see the answer above.

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/nodepools/NodeIdRange.java
ppatierno: Ok got it. But let's make this clearer in the doc.

ppatierno (approved):

PR #8762, opened 2023-07-03
Extend report.sh with information gathered via Kafka Admin API

gracegrimwood:
### Type of change

- Enhancement / new feature

### Description

Adds information from Kafka Admin API to the data gathered by `report.sh`.

When using `report.sh` to gather cluster information for support purposes, this gives further insight into Kafka instance state and other useful configuration details. All the new Admin API information is able to be turned on/off via flags (in the same way secrets are able to be turned on/down/off currently) so users can gather as much or as little information about their cluster as they want.

Specifically, the script can now gather:
- ACLs
- Consumer Groups
- Topics
- Transactions (including hanging transactions)

#### Some further notes:

As much as possible I tried to follow the existing conventions in report.sh, particularly with the new flags added to control how much information is gathered via the Admin API. However, a colleague raised that the `--acls=off`/`--acls=on`/`--consumers=off`/`--consumers=on`/etc phrasing may not read well to a user on the command line, and that something like `--acls=excluded`/`--acls=included` (or a shorter version like `--acls=none`/`--acls=all`) might read a bit better. I stuck with existing convention for the sake of submitting this PR, but this is something I'm happy to discuss.

I also had the thought that outside of `$USAGE` there's no specific indication that these flags control information coming out of Kafka's Admin API specifically, and did consider changing the flags to represent this (i.e. `--kafka-acls` rather than just `--acls`) but thought it was too verbose.

It's also worth noting that these changes mean that `report.sh` directly interacts with the Kafka instance now, whereas before the closest it got was reading properties files off the pods. This is a notable change of behaviour, but not one I foresee causing issues except perhaps in fringe cases (high load, large amount of information being gathered), and even so this can be mitigated by turning this information gathering off via the provided flags.

### Checklist

- [ ] Write tests
- [ ] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/tools/report.sh
+++ b/tools/report.sh
@@ -16,6 +16,9 @@ OC_INSTALLED=false
 KUBE_CLIENT="kubectl"
 OUT_DIR=""
 SECRETS_OPT="hidden"
+ACLS_OPT="off"
+CONSUMERS_OPT="off"
+TRANSACTIONS_OPT="on"
 
 # sed non-printable text delimiter
 SD=$(echo -en "\001") && readonly SD
@@ -55,6 +58,9 @@ Optional:
   --mm2=<string>                MM2 component name to get pods and logs.
   --secrets=(off|hidden|all)    Secret verbosity. Default is hidden, only the secret key will be reported.
   --out-dir=<string>            Script output directory.
+  --acls=(off|on)               Kafka ACL (Access Control List) verbosity. Default is off, these will not be reported.
+  --consumers=(off|on)          Kafka Consumer information verbosity. Default is off, these will not be reported.
+  --transactions=(off|on)       Kafka Transaction information verbosity. Default is on, high-level transaction information will be reported.
 "
 OPTSPEC=":-:"
 while getopts "$OPTSPEC" optchar; do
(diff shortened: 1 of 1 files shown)


scholzj (commented):
This will not work:
* it does not handle any encryption, authentication, authorization
* it hardcodes listener name which is wrong and will not work in many cases
* It is not really clear where does it run the Kafka commands -> it should definitely not do it in any Kafka broker pods as that is dangerous and can cause outages

I think we should also try to separate the concerns -> collecting any Kafka data should be probably part of a different tool focusing on Kafka which handles the things such as authentication and encryption properly and is executed separately.

scholzj (commented):
I guess the overall usecase is not clear to me as well. ACLs and Topics are likely already covered in Strimzi resources. For transactions and consumers, it is not really clear how they fit into the Strimzi report as Strimzi does nto really do anything about them.

tools/report.sh
```diff
@@ -333,6 +404,15 @@ for POD in $PODS; do
   fi
 done
 
+echo "kafka"
+mkdir -p "$OUT_DIR"/reports/kafka/
+SERVICE=$($KUBE_CLIENT -n "$NAMESPACE" get service  -o name -l "app.kubernetes.io/name=kafka" | grep bootstrap | head -n1) && readonly SERVICE
+BOOTSTRAP_SERVER=$($KUBE_CLIENT get kafka "$CLUSTER" -o=jsonpath='{.status.listeners[?(@.type=="plain")].bootstrapServers}' -n "$NAMESPACE") && readonly BOOTSTRAP_SERVER
```
ppatierno, line 410: a `plain` listener could not exist at all, and the cluster having a TLS based one only. In such a case you would need to get cluster CA certificate to connect (at least for encryption).

gracegrimwood:
Thanks @scholzj and @ppatierno for your feedback here, there are definitely things I missed (like running the commands in the broker! That would have been crazy now that I think about it, and would require baking this into it's own image to work around it).
On further reflection, I think you're right that this is not really the right place for these additions. I've had a chat with the folks who prompted this work, and we've agreed that for the effort required to make this workable, it's just not really worth it right now (and it likely wouldn't belong in Strimzi either way).
Appreciate the thought you gave to this 😄

PR #8763, opened 2023-07-03
Added pod name in the certificate SANs list

ppatierno:
While investigating why the `DynamicConfSharedST` is failing on PR #8752 , it turned out that ZooKeeper is resolving to just the pod name what doing the DNS reverse lookup for hostname verification.
This PR adds the podname to the SANs list as well.

### DIFF
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/model/ClusterCa.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/model/ClusterCa.java
@@ -162,6 +162,7 @@ public class ClusterCa extends Ca {
             subject.addDnsName(ccDnsGenerator.serviceDnsNameWithoutClusterDomain());
             subject.addDnsName(ccDnsGenerator.serviceDnsName());
             subject.addDnsName(CruiseControlResources.serviceName(kafkaName));
+            subject.addDnsName(node.podName());
             subject.addDnsName("localhost");
             return subject.build();
         };
@@ -198,6 +199,7 @@ public class ClusterCa extends Ca {
             subject.addDnsName(zkDnsGenerator.wildcardServiceDnsName());
             subject.addDnsName(zkHeadlessDnsGenerator.wildcardServiceDnsNameWithoutClusterDomain());
             subject.addDnsName(zkHeadlessDnsGenerator.wildcardServiceDnsName());
+            subject.addDnsName(node.podName());
             return subject.build();
         };
 
@@ -228,6 +230,7 @@ public class ClusterCa extends Ca {
 
             subject.addDnsName(DnsNameGenerator.podDnsName(namespace, KafkaResources.brokersServiceName(crName), node.podName()));
             subject.addDnsName(DnsNameGenerator.podDnsNameWithoutClusterDomain(namespace, KafkaResources.brokersServiceName(crName), node.podName()));
+            subject.addDnsName(node.podName());
 
             if (externalBootstrapAddresses != null)   {
                 for (String dnsName : externalBootstrapAddresses) {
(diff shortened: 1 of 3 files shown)


cluster-operator/src/main/java/io/strimzi/operator/cluster/model/ClusterCa.java
```diff
@@ -162,6 +162,7 @@ protected Map<String, CertAndKey> generateCcCerts(String namespace, String kafka
             subject.addDnsName(ccDnsGenerator.serviceDnsNameWithoutClusterDomain());
             subject.addDnsName(ccDnsGenerator.serviceDnsName());
             subject.addDnsName(CruiseControlResources.serviceName(kafkaName));
+            subject.addDnsName(node.podName());
```
scholzj, line 165: Wait ... are you adding it just randomly to all certificates? Why does CC need it?

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/ClusterCa.java
```diff
@@ -228,6 +230,7 @@ protected Map<String, CertAndKey> generateBrokerCerts(
 
             subject.addDnsName(DnsNameGenerator.podDnsName(namespace, KafkaResources.brokersServiceName(crName), node.podName()));
             subject.addDnsName(DnsNameGenerator.podDnsNameWithoutClusterDomain(namespace, KafkaResources.brokersServiceName(crName), node.podName()));
+            subject.addDnsName(node.podName());
```
scholzj, line 233: Same as above, why fo Kafka brokers need this?

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/ClusterCa.java
```diff
@@ -198,6 +199,7 @@ protected Map<String, CertAndKey> generateZkCerts(
             subject.addDnsName(zkDnsGenerator.wildcardServiceDnsName());
             subject.addDnsName(zkHeadlessDnsGenerator.wildcardServiceDnsNameWithoutClusterDomain());
             subject.addDnsName(zkHeadlessDnsGenerator.wildcardServiceDnsName());
+            subject.addDnsName(node.podName());
```
scholzj, line 202: Maybe you can move it somewhere up top to have it sorted from simplest to most complex? E.g. above the two pod names with more complete DNS addresses (i.e. above `subject.addDnsName(DnsNameGenerator.podDnsName(namespace, KafkaResources.zookeeperHeadlessServiceName(crName), node.podName()));`)?

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/ClusterCa.java
ppatierno: I was adding here and in CC just for consistency. I don't think it hurts. But I am fine to remove because for our specific use case, it doesn't matter at all.

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/ClusterCa.java
ppatierno: See below answer.

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/ClusterCa.java
ppatierno: Yeah makes sense.

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/ClusterCa.java
scholzj: There is no consistency already. Zoo has its own set of SANs because of its weird TLS hostname verification. But we should not add that to other components as well.

ppatierno: Yeah, right. I removed from the other places.

scholzj (approved):

PR #8764, opened 2023-07-03
[ST] Add update method for all resources and fix UserScalabilityIsolatedST

jankalinic:
### Type of change

- Refactoring

### Description

This PR introduces UPDATE method for all resources implementing ResourceType. It adds the ability to use update for resources which did not have this opportunity and used replace methods within their respective utils class. This is taken advantage of and respectively fixes UserScalabilityIsolatedST where a large amount of kafka users were created and then altered. This alteration caused issues as users were already created and method in resourceManager provided only creation and not replacement / update, so this was fixed with simple use of update method for the kafka user spec in the ST.

### Checklist

- [ ] Write tests
- [ ] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/systemtest/src/main/java/io/strimzi/systemtest/resources/crd/KafkaUserResource.java
+++ b/systemtest/src/main/java/io/strimzi/systemtest/resources/crd/KafkaUserResource.java
@@ -47,6 +47,9 @@ public class KafkaUserResource implements ResourceType<KafkaUser> {
     public static MixedOperation<KafkaUser, KafkaUserList, Resource<KafkaUser>> kafkaUserClient() {
         return Crds.kafkaUserOperation(ResourceManager.kubeClient().getClient());
     }
+    public void update(KafkaUser kafkaUser) {
+        kafkaUserClient().inNamespace(kafkaUser.getMetadata().getNamespace()).resource(kafkaUser).update();
+    }
 
     public static void replaceUserResourceInSpecificNamespace(String resourceName, Consumer<KafkaUser> editor, String namespaceName) {
         ResourceManager.replaceCrdResource(KafkaUser.class, KafkaUserList.class, resourceName, editor, namespaceName);
--- a/systemtest/src/main/java/io/strimzi/systemtest/resources/crd/KafkaUserResource.java
+++ b/systemtest/src/main/java/io/strimzi/systemtest/resources/crd/KafkaUserResource.java
@@ -47,6 +47,8 @@ public class KafkaUserResource implements ResourceType<KafkaUser> {
     public static MixedOperation<KafkaUser, KafkaUserList, Resource<KafkaUser>> kafkaUserClient() {
         return Crds.kafkaUserOperation(ResourceManager.kubeClient().getClient());
     }
+
+    @Override
     public void update(KafkaUser kafkaUser) {
         kafkaUserClient().inNamespace(kafkaUser.getMetadata().getNamespace()).resource(kafkaUser).update();
     }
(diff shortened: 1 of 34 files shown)


im-konge (approved):

scholzj (commented):
Doing this one by one is normally much slower than doing it as a list operation. Given this is a scalability test, is this sufficient? Isn't it better to update the resource manage and give it update support?

jankalinic:
Sure!  I'll make it so that resource manager updates that with a given array of users.

jankalinic:
I tried to look for a way, but cannot find one where update using API takes an array of resources, even in our resource manager we create resources one by one. The replacement of user resource cannot be done as `user already exists` is thrown - which this PR solves by updating the user instead. I also tested deletion of all users from the list and then their creation via resource manager, but it took even more time than updating them.

jankalinic:
@strimzi-ci run tests --ocp-version=4.13 --install-type=bundle --testcase=UserScalabilityIsolatedST

systemtest/src/test/java/io/strimzi/systemtest/operators/user/UserScalabilityIsolatedST.java
```diff
@@ -112,12 +112,16 @@ private void alterAllUsersInList(ExtensionContext extensionContext, List<KafkaUs
             .endAcl()
             .build();
 
-        listOfUsers.forEach(kafkaUser -> new KafkaUserBuilder(kafkaUser).editOrNewSpec().withKafkaUserAuthorizationSimple(updatedAcl).endSpec().build());
+        listOfUsers.replaceAll(kafkaUser -> new KafkaUserBuilder(kafkaUser)
+                .editSpec()
+                    .withKafkaUserAuthorizationSimple(updatedAcl)
+                .endSpec()
+                .build());
```
im-konge, line 119: indent

jankalinic: Thanks, will do

jankalinic:
Would this approach be sufficient enough ? (updating resources -latest commits)

scholzj (commented):
Is it really intentional to have all these changes here? The description talks about a simple change.

systemtest/src/main/java/io/strimzi/systemtest/resources/crd/KafkaUserResource.java
```diff
@@ -47,6 +47,9 @@ public boolean waitForReadiness(KafkaUser resource) {
     public static MixedOperation<KafkaUser, KafkaUserList, Resource<KafkaUser>> kafkaUserClient() {
         return Crds.kafkaUserOperation(ResourceManager.kubeClient().getClient());
     }
+    public void update(KafkaUser kafkaUser) {
```
scholzj, line 50: Why does this one not `@Override` the parent?

scholzj:
Also, you still seem to be updating them one by one. So my original question if that is sufficient I guess stands.

jankalinic:
> Is it really intentional to have all these changes here? The description talks about a simple change.

Changes regarding update seem relevant to this PR, so i will change description and the title. Thank you

systemtest/src/main/java/io/strimzi/systemtest/resources/crd/KafkaUserResource.java
jankalinic: Sorry my mistake, fixed, thank you for noticing.

jankalinic:
> Also, you still seem to be updating them one by one. So my original question if that is sufficient I guess stands.

If you mean updating one by one the way it is right now in ResourceManager update() ... i could not find a way to make it update all at once and all other resources are processed by ResourceManager the same way one by one.

scholzj (approved):

im-konge (approved):

PR #8771, opened 2023-07-04
[systemtest] Add `CRUISE_CONTROL` tag to tests that use CC

### LINKED ISSUE #8618: [STs] Tests that are using Cruise Control should be marked with the `@Tag(CRUISE_CONTROL)` tag
All tests in the `LogSettingST` class are using Cruise Control because it is set up in the initial deployment of the shared Kafka cluster. However, they are not all marked with the `@Tag(CRUISE_CONTROL)` tag. Only one test which interacts with Cruise Control is marked with it. This is not sufficient, because when the Cruise Control doesn't work, all of these tests will fail during the setup phase even when tests with Cruise Control are excluded using the tag.

This should be addressed ...
* Either by enabling Cruise Control only int he test using it instead of the shared cluster in the setup phase
* Or by marking the whole test suite with the correct tag

The same seems to apply for `MultipleNamespaceIsolatedST` which seems to use setup from `AbstractNamespaceST` which seems to be also using Cruise Control. Also for `AllNamespacesIsolatedST`, `SecurityST`

im-konge:
### Type of change

- Enhancement

### Description

This PR adds tag `CRUISE_CONTROL` to tests that use CC, but miss it somehow.

Fixes #8618 

### Checklist

### DIFF
--- a/systemtest/src/test/java/io/strimzi/systemtest/log/LogSettingST.java
+++ b/systemtest/src/test/java/io/strimzi/systemtest/log/LogSettingST.java
@@ -76,6 +76,7 @@ import static org.hamcrest.CoreMatchers.startsWith;
 import static org.hamcrest.MatcherAssert.assertThat;
 
 @Tag(REGRESSION)
+@Tag(CRUISE_CONTROL)
 @TestMethodOrder(OrderAnnotation.class)
 class LogSettingST extends AbstractST {
     private static final Logger LOGGER = LogManager.getLogger(LogSettingST.class);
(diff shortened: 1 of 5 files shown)


jankalinic (approved):

kornys (approved):

scholzj (approved):
LGTM assuming the tag is set correctly now.

im-konge:
@strimzi-ci run tests --install-type=bundle --groups=cruisecontrol

PR #8648, opened 2023-06-09
[system test] - removal of class-wide parallelism support 

see-quick:
### Type of change

- Refactoring
- Documentation

### Description

This PR removes the order of the test suites. Also, this is the initial PR for removing:
1. support for class-wide parallelism which does not seem to fit our requirements (i.e., OrderTestSuites class, because it is not bounded to any other part)
2. BeforeAllOnce class, and re-place all test classes with branch new deployment of Cluster Operator.
3. Partially remove of ExecutionListener class, SuiteThreadController class and the logic inside AbstractST class
4. Remove completely annotations @ParallelSuite and @IsolatedSuite + StUtils clean-up

The reasons why we are gonna remove class-wide parallelism are:
1. It is hard to maintain (i.e., when one decides to create a test case it simply tests it in sequential mode. But to completely verify that everything works (parallelization) it has to be executed method-wide, where all test cases will be executed in parallel (this is usually done by us system-test owners but I think very few people are doing it apart from us. And lastly to fully prove that everything works one has to run it class-wide, which takes a lot of time where you have to check if there are no conflicts between classes and this is just a small piece of a huge maintaining nightmare.)
5. It is very costly for resources (i.e., running multiple classes takes a lot of resources)
6. It is very hard to troubleshoot a problem for other people than its creator (i.e., thread interleaving, problems that do not occur in sequential mode or method-wide because of conflicts between classes causing race conditions)

### Checklist

- [x] Make sure all tests pass

PR #8773, opened 2023-07-04
[systemtest] Fix issue with HelmChartIsolatedST

im-konge:
### Type of change

- Bugfix

### Description

After #8648 `HelmChartIsolatedST` started to fail, because it was picking last used namespace from the previous test.
This should fix the issue, so the Helm client will always use namespace to which the CO should be installed to.

### Checklist


jankalinic (approved):

scholzj (approved):
LGTM if it works.

PR #8776, opened 2023-07-04
Fix Cruise Control check for a number of nodes to calculate with NodePools

### LINKED ISSUE #8770: [Bug]: CC and KafkaNodePools - operator doesn't reflect number of replicas inside KafkaNodePools
### Bug Description

When we deploy Kafka with 1 replica and we want to deploy CC, the Kafka resource have `NotReady` state with following message:

```
- lastTransitionTime: "2023-07-04T08:07:17.109377416Z"
      message: Kafka namespace-0/my-cluster-7ed0a8ae has invalid configuration. Cruise
        Control cannot be deployed with a single-node Kafka cluster. It requires at
        least two Kafka nodes.
      reason: InvalidResourceException
      status: "True"
      type: NotReady
```
which is expected behavior.
When we scale to 3 replicas, the CC (after Kafka Pods are scaled) is deployed.

But with NodePools, this scenario doesn't work. The replicas for Kafka are set inside the KafkaNodePool resource, which is a source of truth for the Kafka cluster deployment. But, the operator checks Kafka resource for the replicas count.
That means when we deploy KafkaNodePool for Kafka with 1 replica (and also Kafka resource with 1 replica for Kafka), the message like above appears - correct behavior.
But when we scale the replicas inside the KafkaNodePool to 3, it seems that there are no Kafka Pods at all:
```
kubectl get pod -n namespace-0 
                                                                                                                                                                                                                                                                                             
NAME                              READY   STATUS    RESTARTS   AGE
my-cluster-7ed0a8ae-zookeeper-0   1/1     Running   0          13m
```
and the message (as above) is still appearing in logs of CO and also in the Kafka resource's status.

When we then update replicas inside the Kafka CR, the Kafka Pods are created and after that, the CC is deployed as well.

### Steps to reproduce

1. Deploy CO with KafkaNodePools enabled
2. Create KafkaNodePool with 1 replica
3. Deploy Kafka with NodePools enabled, setting the Kafka's replica to 1 as well and CC enabled
4. Check status of Kafka (message as mentioned should appear)
5. Scale KafkaNodePool's replicas to 3
6. Check Pods, logs and Kafka resource's status

### Expected behavior

The Kafka pods should be scaled to 3 based on the NodePool configuration and CC should be deployed, ignoring the number of replicas inside the Kafka CR

### Strimzi version

main

### Kubernetes version

all

### Installation method

YAML

### Infrastructure

_No response_

_No response_

### Additional context

_No response_

scholzj:
### Type of change

- Bugfix

### Description

Currently, we do not support Cruise Control being used with a Kafka cluster having a single Kafka broker. But the validation logic did not take KafkaNodePools into account. This PR updates it to use the KafkaNodePools to execute this check. It also updates the error message to make it more clear it is the broker nodes which matter here. And it adds additional tests with NodePools.

This should resolve #8770 

### Checklist

- [x] Write tests

### DIFF
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaCluster.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaCluster.java
@@ -451,10 +451,11 @@ public class KafkaCluster extends AbstractModel implements SupportsMetrics, Supp
             addAll(metricReporterList, configuration.getConfigOption(KAFKA_METRIC_REPORTERS_CONFIG_FIELD).split(","));
         }
 
-        if (kafkaAssembly.getSpec().getCruiseControl() != null && kafkaAssembly.getSpec().getKafka().getReplicas() < 2) {
+        if (kafkaAssembly.getSpec().getCruiseControl() != null
+                && kafkaCluster.nodes().stream().filter(n -> n.broker()).count() < 2) {
             throw new InvalidResourceException("Kafka " +
                     kafkaAssembly.getMetadata().getNamespace() + "/" + kafkaAssembly.getMetadata().getName() +
-                    " has invalid configuration. Cruise Control cannot be deployed with a single-node Kafka cluster. It requires at least two Kafka nodes.");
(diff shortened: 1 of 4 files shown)


im-konge (approved):
Thanks :)

ppatierno (approved):
LGTM.

PR #8777, opened 2023-07-04
docs(node pools): adds scaling management annotations

PaulRMellor:
**Documentation**

Updates the docs for the nod pools preview.
Adds content related to advanced node ID management for scaling operations.

Includes new and updated procedures:
*  Assigning IDs to node pools for scaling operations
*  Adding nodes to node pools
*  Removing nodes from node pools
*  Moving nodes between  node pools

### Checklist

- [ ] Write tests
- [x] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/documentation/modules/configuring/con-scaling-kafka-clusters.adoc
+++ b/documentation/modules/configuring/con-scaling-kafka-clusters.adoc
@@ -40,11 +40,15 @@ spec:
  # ...
 ----
 
+When adding brokers through the `Kafka` configuration, node IDs start at 0 (zero) and the Cluster Operator assigns the next lowest ID to a new node. 
+The broker removal process starts from the broker pod with the highest ID in the cluster.
+
+If you are managing nodes in the cluster using the the preview of the node pools feature, you adjust the `KafkaNodePool.spec.replicas` configuration to change the number of nodes in the node pool.
+With node pools, you can  xref:proc-managing-node-pools-ids-{context}[assign node IDs for scaling operations].  
+
 When you add add or remove brokers, Kafka does not automatically reassign partitions. 
 The best way to do this is using Cruise Control.
 You can use Cruise Control's `add-brokers` and `remove-brokers` modes when scaling a cluster up or down.
 
 * Use the `add-brokers` mode after scaling up a Kafka cluster to move partition replicas from existing brokers to the newly added brokers.
 * Use the `remove-brokers` mode before scaling down a Kafka cluster to move partition replicas off the brokers that are going to be removed.
-
-NOTE: When scaling down brokers, you cannot specify which specific pod to remove from the cluster. Instead, the broker removal process starts from the highest numbered pod.
--- a/documentation/modules/configuring/con-scaling-kafka-clusters.adoc
+++ b/documentation/modules/configuring/con-scaling-kafka-clusters.adoc
@@ -44,7 +44,7 @@ When adding brokers through the `Kafka` configuration, node IDs start at 0 (zero
 The broker removal process starts from the broker pod with the highest ID in the cluster.
 
 If you are managing nodes in the cluster using the the preview of the node pools feature, you adjust the `KafkaNodePool.spec.replicas` configuration to change the number of nodes in the node pool.
-With node pools, you can  xref:proc-managing-node-pools-ids-{context}[assign node IDs for scaling operations].  
+Additionally, when scaling existing clusters with node pools, you can xref:proc-managing-node-pools-ids-{context}[assign node IDs for the scaling operations].  
 
 When you add add or remove brokers, Kafka does not automatically reassign partitions. 
 The best way to do this is using Cruise Control.
--- a/documentation/modules/configuring/proc-managing-node-pools-ids.adoc
+++ b/documentation/modules/configuring/proc-managing-node-pools-ids.adoc
@@ -1,0 +1,92 @@
+// Module included in the following assemblies:
+//
+// assembly-config.adoc
+
+[id='proc-managing-node-pools-ids-{context}']
+= (Preview) Assigning IDs to node pools for scaling operations
+
+[role="_abstract"]
+This procedure describes how to use annotations for advanced node ID handling by the Cluster Operator when performing scaling operations on node pools. 
+You specify the node IDs to use, rather than the Cluster Operator using the next ID in sequence.
+Management of node IDs in this way gives greater control.
+
+The steps assume that you are using ZooKeeper for cluster management. 
+The operation is not possible for clusters using KRaft mode.
+
+To add a range of IDs, you assign the following annotations to the `KafkaNodePool` resource:
+
+* `strimzi.io/next-node-ids` to add a range of IDs for scaling up
+* `strimzi.io/remove-node-ids` to add a range of IDs for scaling down
+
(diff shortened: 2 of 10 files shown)


documentation/modules/configuring/con-scaling-kafka-clusters.adoc
```diff
@@ -40,11 +40,15 @@ spec:
  # ...
 ----
 
+When adding brokers through the `Kafka` configuration, node IDs start at 0 (zero) and the Cluster Operator assigns the next lowest ID to a new node. 
+The broker removal process starts from the broker pod with the highest ID in the cluster.
+
+If you are managing nodes in the cluster using the the preview of the node pools feature, you adjust the `KafkaNodePool.spec.replicas` configuration to change the number of nodes in the node pool.
+With node pools, you can  xref:proc-managing-node-pools-ids-{context}[assign node IDs for scaling operations].
```
scholzj, line 47: Should we make it more clear here that even creating a new cluster is in a way _scaling operation_ (from nothing to N nodes)? I think some people might confuse this test right now as that they can use it only when scaling existing clusters.

documentation/modules/configuring/proc-managing-node-pools-ids.adoc
```diff
@@ -0,0 +1,92 @@
...
+[id='proc-managing-node-pools-ids-{context}']
+= (Preview) Assigning IDs to node pools for scaling operations
+
+[role="_abstract"]
+This procedure describes how to use annotations for advanced node ID handling by the Cluster Operator when performing scaling operations on node pools. 
+You specify the node IDs to use, rather than the Cluster Operator using the next ID in sequence.
+Management of node IDs in this way gives greater control.
+
+The steps assume that you are using ZooKeeper for cluster management. 
+The operation is not possible for clusters using KRaft mode.
```
scholzj, line 14: This is supported in exactly the same way for ZooKeeper or KRaft. So I think this can be removed.

documentation/modules/configuring/proc-managing-node-pools-ids.adoc
```diff
@@ -0,0 +1,92 @@
...
+
+To add a range of IDs, you assign the following annotations to the `KafkaNodePool` resource:
+
+* `strimzi.io/next-node-ids` to add a range of IDs for scaling up
+* `strimzi.io/remove-node-ids` to add a range of IDs for scaling down
```
scholzj, line 19: As mentioned above, this might make it more clear it applies to any created / removed broker.

```suggestion
* `strimzi.io/next-node-ids` to add a range of IDs that will be used for new brokers
* `strimzi.io/remove-node-ids` to add a range of IDs for removing existing brokers
```

documentation/modules/configuring/proc-managing-node-pools-ids.adoc
```diff
@@ -0,0 +1,92 @@
...
+* Scale down removes the node with the highest available ID in the range. 
+
+If there are gaps in the sequence of node IDs assigned in the node pool, the next node to be added is assigned an ID that fills the gap.
+
+The annotations don't need to be updated after every scaling operation.
+Any unused IDs are still valid for the next scaling event. 
+
+The Cluster Operator allows you to specify a range of IDs in either ascending or descending order, so you can define them in the order the nodes are scaled. 
+For example, when scaling up, you can specify a range such as [`1000-1999`], and the new nodes are assigned the next lowest IDs: `1000`, `1001`, `1002`, `1003`, and so on. 
+Conversely, when scaling down, you can specify a range like [`1999-1000`], ensuring that nodes with the next highest IDs are removed: `1999`, `1998`, `1997`, `1996`, and so on.
```
scholzj, line 44: I wonder if this should have a better example to clarify that it will remove the highest used node ID. E.g. with a node pool with IDs `1000`, `1001`, `1002`, `1003`, it will remove them as `1003`, `1002` etc. I think that might be more clear than starting with `1999` which will in this example be rarely used.

documentation/modules/configuring/proc-scaling-down-node-pools.adoc
```diff
@@ -0,0 +1,61 @@
...
+// assembly-config.adoc
+
+[id='proc-scaling-down-node-pools-{context}']
+= (Preview) Removing nodes from a node pool
+
+[role="_abstract"]
+This procedure describes how to scale down a node pool to remove nodes.
+
+The steps assume that you are using ZooKeeper for cluster management. 
+The operation is not possible for clusters using KRaft mode.
```
scholzj, line 12: I think we can remove this.

ppatierno:
I was wondering why for the scale down example we are using the kafka reassignment tool to remove partition replicas from brokers to be removed instead of `KafkaRebalance` with the `remove-broker` API.

documentation/modules/configuring/con-scaling-kafka-clusters.adoc
PaulRMellor: Changed as follows:

> If you are managing nodes in the cluster using the the preview of the node pools feature, you adjust the `KafkaNodePool.spec.replicas` configuration to change the number of nodes in the node pool.
> Additionally, when scaling existing clusters with node pools, you can xref:proc-managing-node-pools-ids-{context}[assign node IDs for the scaling operations].

documentation/modules/configuring/proc-managing-node-pools-ids.adoc
PaulRMellor: Yeah. I've updated.

documentation/modules/configuring/proc-scaling-down-node-pools.adoc
PaulRMellor: Removed for this. Does that mean we retain for scaling up and moving nodes in node pools?

PaulRMellor:
> I was wondering why for the scale down example we are using the kafka reassignment tool to remove partition replicas from brokers to be removed instead of KafkaRebalance with the remove-broker API.

I don't think we have the support there yet.

documentation/modules/configuring/proc-scaling-down-node-pools.adoc
scholzj: I think with regard to broker nodes, everything should work fine. The controller nodes not really, but that is a different story - I would not include it into these procedures.

scholzj (approved):

ppatierno:
> I don't think we have the support there yet.

@PaulRMellor this is the `remove-brokers` rebalancing mode we already have. https://strimzi.io/docs/operators/latest/deploying.html#con-optimization-proposals-modes-str

documentation/modules/configuring/proc-scaling-down-node-pools.adoc
PaulRMellor: Thanks. I've removed from the other procedures

PaulRMellor:
> > I don't think we have the support there yet.
> 
> @PaulRMellor this is the `remove-brokers` rebalancing mode we already have. https://strimzi.io/docs/operators/latest/deploying.html#con-optimization-proposals-modes-str

Thanks @ppatierno. I assumed from the node pools proposal we were not yet able to use these modes, as the reassign tool is suggested there. After clarification, I've updated.

ppatierno (approved):
LGTM

PR #8794, opened 2023-07-06
Create listener-related resources only for broker nodes

scholzj:
### Type of change

- Enhancement / new feature

### Description

Right now, when running in KRaft mode, we create the listener-related resources (per-broker services, routes, and ingresses) for all Kafka nodes. This includes also controller-only nodes where the regular clients are not connecting. This PR updates the logic in `KafkaCluster` class to generate these resources only for nodes with the broker role (both broker-only as well as mixed). It also updates `KafkaListenerReconciler` to wait for the readiness of these resources only for broker nodes. Plus it also fixes the related listener validations.

As part of this change, I needed to do in many different places checks whether a node pool has broker or controller roles. And I also needed to filter the broker nodes from the list with all nodes. To avoid having this logic copied around in many different places, I created new methods in `KafkaCluster` and `KafkaPool` for these checks. I also renamed the field `kraftRoles` to `processRoles`. Originally, this field was used only in KRaft mode. But as we made it mandatory even with ZooKeeper in the Node Pool API, I think calling it `processRoles` fits better the purpose.

### Checklist

- [x] Write tests


ppatierno (approved):
LGTM.

PR #8796, opened 2023-07-07
Use Cruise control only on Kafka broker nodes when running in KRaft mode

scholzj:
### Type of change

- Enhancement / new feature

### Description

When running in KRaft mode, we currently configure all nodes in Cruise Control capacity configuration even through the controller-only nodes do not really host any regular topics and don't have anything to rebalance. This PR updates the way we configure the Cruise Control capacity to include only the Kafka nodes with the broker-role. It does so by passing only the broker node information to Cruise Control where possible (since nobody else uses it and it does not need the controller info, so we safe some memory).

It also configures the Cruise Control metric reporters only on the broker nodes to not have them started / running on the cntroller-only nodes.

### Checklist

- [x] Write tests

### DIFF
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/model/cruisecontrol/CruiseControlMetricsReporter.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/model/cruisecontrol/CruiseControlMetricsReporter.java
@@ -16,7 +16,7 @@ import io.strimzi.operator.cluster.operator.resource.cruisecontrol.CruiseControl
  * @param topicName             Name of the Metrics Reporter topic
  * @param numPartitions         Number of partitions of the Metrics Reporter topic
  * @param replicationFactor     Replication factor of the Metrics Reporter topic
- * @param minInSyncReplicas     Minimal number of in sync replicas in the of the metrics reporter topic
+ * @param minInSyncReplicas     Minimal number of in sync replicas of the metrics reporter topic
  */
 public record CruiseControlMetricsReporter(String topicName, Integer numPartitions, Integer replicationFactor, Integer minInSyncReplicas) {
     // Configuration field names
@@ -101,7 +101,7 @@ public record CruiseControlMetricsReporter(String topicName, Integer numPartitio
                 throw new IllegalArgumentException(
                         "The Cruise Control metric topic minISR was set to a value (" + userConfiguredMinInSync + ") " +
                                 "which is higher than the number of replicas for that topic (" + configuredCcReplicationFactor + "). " +
-                                "Please ensure that the CC metrics topic minISR is <= to the topic's replication factor."
+                                "Please ensure that the Cruise Control metrics topic minISR is <= to the topic's replication factor."
                 );
             }
         }
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/model/cruisecontrol/CruiseControlMetricsReporter.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/model/cruisecontrol/CruiseControlMetricsReporter.java
@@ -1,0 +1,109 @@
+/*
+ * Copyright Strimzi authors.
+ * License: Apache License 2.0 (see the file LICENSE or http://apache.org/licenses/LICENSE-2.0.html).
+ */
+package io.strimzi.operator.cluster.model.cruisecontrol;
+
+import io.strimzi.api.kafka.model.Kafka;
+import io.strimzi.operator.cluster.model.InvalidResourceException;
+import io.strimzi.operator.cluster.model.KafkaConfiguration;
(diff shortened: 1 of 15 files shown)


kyguy (commented):
Should we disable the Cruise Control metric reporter for the controller nodes as part of this PR as well? [1]

[1] https://github.com/strimzi/strimzi-kafka-operator/blob/main/cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaCluster.java#L326

scholzj:
@kyguy I think that would make sense, but would need a much bigger redesign of how it is configured. So it should be probably left for another PR.

kyguy:
With the Cruise Control metric reporters left in the controller nodes, we may need to add a default broker capacity entry in the Capacity class [1]
```
// Default broker entry
BrokerCapacity defaultBrokerCapacity = new BrokerCapacity(BrokerCapacity.DEFAULT_BROKER_ID, cpu, disk, inboundNetwork, outboundNetwork);
capacityEntries.put(BrokerCapacity.DEFAULT_BROKER_ID, defaultBrokerCapacity);
```
Cruise Control may crash if there are metrics being reported for Kafka nodes that don't have capacity entries defined. The default broker entry would help us avoid this problem (should it exist) until we remove the metric reporters. We will know more when the ST pass or not!

[1] https://github.com/strimzi/strimzi-kafka-operator/blob/main/cluster-operator/src/main/java/io/strimzi/operator/cluster/model/cruisecontrol/Capacity.java#L358

scholzj:
It worked fine in my tests - it simply completely ignored the controller brokers.

ppatierno (approved):
I am fine with this change because STs passed, because I had the same fear as Kyle about the Cruise Control behaviour.
Would you mind to open a GitHub issue to track removing metrics reporter from controller nodes?

scholzj:
@kyguy @ppatierno The next release is soon. So maybe it's better to not do only half of it. I reworked the configuration as well. The code is not completely nice because of how it mixes the user configuration and the Cruise Control configuration. But this is mainly driven by the _API_. Please have another look.

kyguy (approved):
>  I reworked the configuration as well. The code is not completely nice because of how it mixes the user configuration and the Cruise Control configuration. But this is mainly driven by the API. Please have another look.

Looks great Jakub, I understand what you mean about the user and cruise configuration but I don't think there is any better way of doing it, thanks for sorting this out, I nervous about the metric reporters on the controller nodes causing issues

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/cruisecontrol/CruiseControlMetricsReporter.java
```diff
@@ -0,0 +1,109 @@
...
+import io.strimzi.operator.cluster.operator.resource.cruisecontrol.CruiseControlConfigurationParameters;
+
+
+/**
+ * Represents a model for the Cruise Control Metrics Reporter
+ *
+ * @param topicName             Name of the Metrics Reporter topic
+ * @param numPartitions         Number of partitions of the Metrics Reporter topic
+ * @param replicationFactor     Replication factor of the Metrics Reporter topic
+ * @param minInSyncReplicas     Minimal number of in sync replicas in the of the metrics reporter topic
```
ppatierno, line 19:
```suggestion
 * @param minInSyncReplicas     Minimal number of in sync replicas of the metrics reporter topic
```

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/cruisecontrol/CruiseControlMetricsReporter.java
```diff
@@ -0,0 +1,109 @@
...
+        }
+
+        if (minInSyncReplicas == null) { // When it is null, we validate the value from the user configuration
+            int userConfiguredMinInSync = Integer.parseInt(configuration.getConfigOption(CruiseControlConfigurationParameters.METRICS_TOPIC_MIN_ISR.getValue()));
+            int configuredCcReplicationFactor = replicationFactor != null ? replicationFactor : Integer.parseInt(configuration.getConfigOption(CruiseControlConfigurationParameters.METRICS_TOPIC_REPLICATION_FACTOR.getValue()));
+            if (userConfiguredMinInSync > configuredCcReplicationFactor) {
+                throw new IllegalArgumentException(
+                        "The Cruise Control metric topic minISR was set to a value (" + userConfiguredMinInSync + ") " +
+                                "which is higher than the number of replicas for that topic (" + configuredCcReplicationFactor + "). " +
+                                "Please ensure that the CC metrics topic minISR is <= to the topic's replication factor."
```
ppatierno, line 104:
```suggestion
                                "Please ensure that the Cruise Control metrics topic minISR is <= to the topic's replication factor."
```

ppatierno (approved):
Thanks Jakub, just a couple of nits but LGTM.

PR #8808, opened 2023-07-10
Set the constructor and handler method used for test as package private

ppatierno:
Trivial PR to set a package protected the `KafkaAgent` ctor used for testing. It doesn't need to be public.


scholzj (approved):

ppatierno:
I actually made the same change for the method returning the handler for the Jetty server.

PR #8810, opened 2023-07-10
Remove reliance on TestcontainersKafkaCluster in topic operator tests.

k-wall:
### Type of change

_Select the type of your PR_

- Refactoring

### Description

Change the Topic Operator tests that use `TestcontainersKafkaCluster` explicitly to use a default `KafkaCluster` instead.  This will mean, by default, they use in-VM kafka(s) rather than use the containerized native binaries.

Why: we don't have multi-arch container image available for the kafka native and zookeeper native image, so this is causing a problem with Strimzi's release pipeline. This change avoids the dependency.

### Checklist

- [ ] Write tests
- [ ] Update documentation
- [ ] Update CHANGELOG.md


k-wall:
As far as I can tell, the tests themselves don't actually rely on kafka being run with a container.  As these tests use > broker, I suspect the motivation for using it was the speed improvement.

ppatierno (approved):
LGTM. Assuming it works for #8801

scholzj (approved):

PR #82, opened 2017-11-03
Extend makefiles to be able to generate release artifacts

scholzj:
This PR adds a new make target `release` which will create the release artifact. It contains all the YAML files needed for deployment ad README.md as our only docu. This is currently not expected to run as part of the CI. It should be run manually and the result should be attached to the appropriatte GitHub release / tag.

PR #8812, opened 2023-07-10
Added a way to retrieve node config by Kafka Agent and exposing via HTTP

ppatierno:
### Type of change

- Enhancement / new feature

### Description

This PR is about proposal PR [#82](https://github.com/strimzi/proposals/pull/82)

### Checklist

- [ ] Write tests
- [ ] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/docker-images/kafka-based/kafka/scripts/kafka_run.sh
+++ b/docker-images/kafka-based/kafka/scripts/kafka_run.sh
@@ -70,12 +70,16 @@ if [ "$STRIMZI_KRAFT_ENABLED" = "true" ]; then
   fi
 
 else
-  rm -f /var/opt/kafka/kafka-ready /var/opt/kafka/zk-connected 2> /dev/null
-  KEY_STORE=/tmp/kafka/cluster.keystore.p12
-  TRUST_STORE=/tmp/kafka/cluster.truststore.p12
-  KAFKA_OPTS="${KAFKA_OPTS} -javaagent:$(ls "$KAFKA_HOME"/libs/kafka-agent*.jar)=/var/opt/kafka/kafka-ready:/var/opt/kafka/zk-connected:$KEY_STORE:$CERTS_STORE_PASSWORD:$TRUST_STORE:$CERTS_STORE_PASSWORD"
-  export KAFKA_OPTS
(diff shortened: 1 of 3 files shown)


scholzj (commented):
I don't follow this. Why are we exposing a file we have in a config map as a HTTP endpoint?

ppatierno:
> I don't follow this. Why are we exposing a file we have in a config map as a HTTP endpoint?

This is actually a good point Jakub. I don't know the Kafka Roller code in details which already interacts with the Kafka Agent for the broker state. As stated in the proposal, this is going to be a temporary solution so I am not sure this is better to be replaced by the final one (using the Kafka Admin client) or it would be the same just loading info from the ConfigMap. Maybe the effort is just the same.
One difference I see is that in the ConfigMap we have some env vars defined (i.e. `${CERTS_STORE_PASSWORD}`) so not the actual configuration as we are getting from the file system (as used by the node). Is it possible that this can happen for other parameters?

Anyway, I thought that we all agreed together doing this way in our offline call together with @tombentley. It was also described in the corresponding proposal, maybe you should have raised the question there?

scholzj:
> One difference I see is that in the ConfigMap we have some env vars defined (i.e. ${CERTS_STORE_PASSWORD}) so not the actual configuration as we are getting from the file system (as used by the node). Is it possible that this can happen for other parameters?

Well, one thing to keep in mind, the config you will be diffing it against will have the `${CERTS_STORE_PASSWORD}` placeholders as well.

> Anyway, I thought that we all agreed together doing this way in our offline call together with @tombentley. It was also described in the corresponding proposal, maybe you should have raised the question there?

To be honest, my understanding was that it would get the current configuration somehow from Kafka directly (e.g. JMX or some internal APIs). By reading the file, you might also miss the cluster-wide config changes updated internally etc.

ppatierno:
> To be honest, my understanding was that it would get the current configuration somehow from Kafka directly (e.g. JMX or some internal APIs). By reading the file, you might also miss the cluster-wide config changes updated internally etc.

No, the idea was just about reading the file, and I think it was clear. Again as stated in the proposal.
It was accepted that as a temporary solution any dynamic changes would not be reflected to the file, so in general not necessary rollings could happen.

Issue #8060, opened 2023-02-07
Recreate watcher when closed with an exception


kyguy:
**Describe the bug**

Some of the operators do not recreate watches after they are closed with an error:

- AbstractConnectOperator
- KafkaRebalanceAssemblyOperator

This can lead to a custom resources not being watched and updated after the watch on that resource expires [1]. When this happens errors like the following will show in the Cluster Operator log:
```
2022-12-31T20:41:34.080724387Z 2022-12-31 20:41:34 ERROR AbstractWatchManager:315 - Unhandled exception encountered in watcher event handler
2022-12-31T20:41:34.080724387Z io.fabric8.kubernetes.client.KubernetesClientException: too old resource version: 119220794 (119221972)
2022-12-31T20:41:34.080724387Z  at io.strimzi.operator.cluster.operator.assembly.KafkaRebalanceAssemblyOperator$1.onClose(KafkaRebalanceAssemblyOperator.java:233) ~[io.strimzi.cluster-operator-0.29.0.redhat-00014.jar:0.29.0.redhat-00014]
2022-12-31T20:41:34.080724387Z  at io.fabric8.kubernetes.client.utils.WatcherToggle.onClose(WatcherToggle.java:56) ~[io.fabric8.kubernetes-client-5.12.2.redhat-00002.jar:?]
```
We had this same problem with the TopicOperator in the past and fixed it by updating the `K8sTopicWatcher` onClose() method. [2] We also already recreate watches for other operators in the same situation [5] [6] We should look into doing the same for the:

- AbstractConnectOperator [3]
- KafkaRebalanceAssemblyOperator [4]

**Expected behavior**
From what I understand from [1] it is the client's responsibility, or in this case the operators responsibility, to recreate a watch after it expires to avoid fabric8 errors like the following:
```
2022-12-31T20:41:34.080724387Z io.fabric8.kubernetes.client.KubernetesClientException: too old resource version: 119220794 (119221972)
```

**Additional context**
[1] https://stackoverflow.com/questions/61409596/kubernetes-too-old-resource-version
[2] https://github.com/strimzi/strimzi-kafka-operator/issues/3771
[3] https://github.com/strimzi/strimzi-kafka-operator/blob/0.33.0/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/AbstractConnectOperator.java#L287-L291
[...]

PR #8815, opened 2023-07-11
Recreate watcher when closed with an exception

kyguy:
### Type of change

- Bugfix

### Description

Addresses #8060

Some of the operators do not recreate watches after they are closed with an error:

    AbstractConnectOperator
    KafkaRebalanceAssemblyOperator

This can lead to a custom resources not being watched and updated after the watch on that resource expires

### Checklist

- [x] Write tests
- [ ] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/ClusterOperator.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/ClusterOperator.java
@@ -6,8 +6,8 @@ package io.strimzi.operator.cluster;
 
 import io.fabric8.kubernetes.client.KubernetesClient;
 import io.fabric8.kubernetes.client.Watch;
+import io.strimzi.api.kafka.model.KafkaConnector;
 import io.strimzi.api.kafka.model.nodepool.KafkaNodePool;
-import io.strimzi.operator.cluster.operator.assembly.AbstractConnectOperator;
 import io.strimzi.operator.cluster.operator.assembly.KafkaAssemblyOperator;
 import io.strimzi.operator.cluster.operator.assembly.KafkaBridgeAssemblyOperator;
 import io.strimzi.operator.cluster.operator.assembly.KafkaConnectAssemblyOperator;
@@ -117,7 +117,8 @@ public class ClusterOperator extends AbstractVerticle {
         if (!config.isPodSetReconciliationOnly()) {
             List<AbstractOperator<?, ?, ?, ?>> operators = new ArrayList<>(asList(
                     kafkaAssemblyOperator, kafkaMirrorMakerAssemblyOperator,
-                    kafkaConnectAssemblyOperator, kafkaBridgeAssemblyOperator, kafkaMirrorMaker2AssemblyOperator));
+                    kafkaConnectAssemblyOperator, kafkaBridgeAssemblyOperator, kafkaMirrorMaker2AssemblyOperator,
+                    kafkaRebalanceAssemblyOperator));
             for (AbstractOperator<?, ?, ?, ?> operator : operators) {
                 startFutures.add(operator.createWatch(namespace, operator.recreateWatch(namespace)).compose(w -> {
                     LOGGER.info("Opened watch for {} operator", operator.kind());
@@ -135,8 +136,11 @@ public class ClusterOperator extends AbstractVerticle {
                 }));
             }
 
-            startFutures.add(AbstractConnectOperator.createConnectorWatch(kafkaConnectAssemblyOperator, namespace, config.getCustomResourceSelector()));
-            startFutures.add(kafkaRebalanceAssemblyOperator.createRebalanceWatch(namespace));
+            startFutures.add(kafkaConnectAssemblyOperator.createConnectorWatch(namespace, kafkaConnectAssemblyOperator.recreateWatch(namespace)).compose(w -> {
+                LOGGER.info("Opened KafkaConnector watch for {} operator", kafkaConnectAssemblyOperator.kind());
+                watchByKind.put(KafkaConnector.RESOURCE_KIND, w);
+                return Future.succeededFuture();
+            }));
         }
 
         Future.join(startFutures)
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/ClusterOperator.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/ClusterOperator.java
@@ -7,6 +7,7 @@ package io.strimzi.operator.cluster;
 import io.fabric8.kubernetes.client.KubernetesClient;
 import io.fabric8.kubernetes.client.Watch;
 import io.strimzi.api.kafka.model.KafkaConnector;
+import io.strimzi.api.kafka.model.KafkaRebalance;
 import io.strimzi.api.kafka.model.nodepool.KafkaNodePool;
 import io.strimzi.operator.cluster.operator.assembly.KafkaAssemblyOperator;
 import io.strimzi.operator.cluster.operator.assembly.KafkaBridgeAssemblyOperator;
@@ -117,8 +118,7 @@ public class ClusterOperator extends AbstractVerticle {
         if (!config.isPodSetReconciliationOnly()) {
             List<AbstractOperator<?, ?, ?, ?>> operators = new ArrayList<>(asList(
                     kafkaAssemblyOperator, kafkaMirrorMakerAssemblyOperator,
-                    kafkaConnectAssemblyOperator, kafkaBridgeAssemblyOperator, kafkaMirrorMaker2AssemblyOperator,
-                    kafkaRebalanceAssemblyOperator));
+                    kafkaConnectAssemblyOperator, kafkaBridgeAssemblyOperator, kafkaMirrorMaker2AssemblyOperator));
             for (AbstractOperator<?, ?, ?, ?> operator : operators) {
                 startFutures.add(operator.createWatch(namespace, operator.recreateWatch(namespace)).compose(w -> {
                     LOGGER.info("Opened watch for {} operator", operator.kind());
@@ -136,9 +136,15 @@ public class ClusterOperator extends AbstractVerticle {
                 }));
             }
 
-            startFutures.add(kafkaConnectAssemblyOperator.createConnectorWatch(namespace, kafkaConnectAssemblyOperator.recreateWatch(namespace)).compose(w -> {
+            startFutures.add(kafkaConnectAssemblyOperator.createKafkaConnectorWatch(namespace, config.getCustomResourceSelector()).compose(w -> {
                 LOGGER.info("Opened KafkaConnector watch for {} operator", kafkaConnectAssemblyOperator.kind());
                 watchByKind.put(KafkaConnector.RESOURCE_KIND, w);
+                return Future.succeededFuture();
+            }));
+
+            startFutures.add(kafkaRebalanceAssemblyOperator.createRebalanceWatch(namespace, kafkaRebalanceAssemblyOperator.recreateRebalanceWatch(namespace)).compose(w -> {
+                LOGGER.info("Opened watch for {} operator", kafkaRebalanceAssemblyOperator.kind());
+                watchByKind.put(KafkaRebalance.RESOURCE_KIND, w);
                 return Future.succeededFuture();
             }));
         }
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/AbstractConnectOperator.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/AbstractConnectOperator.java
@@ -15,19 +15,15 @@ import io.fabric8.kubernetes.api.model.networking.v1.NetworkPolicy;
 import io.fabric8.kubernetes.api.model.rbac.ClusterRoleBinding;
 import io.fabric8.kubernetes.client.CustomResource;
 import io.fabric8.kubernetes.client.KubernetesClient;
-import io.fabric8.kubernetes.client.KubernetesClientException;
 import io.fabric8.kubernetes.client.Watch;
 import io.fabric8.kubernetes.client.Watcher;
 import io.fabric8.kubernetes.client.WatcherException;
 import io.fabric8.kubernetes.client.dsl.Resource;
 import io.micrometer.core.instrument.Timer;
 import io.netty.channel.ConnectTimeoutException;
-import io.strimzi.api.kafka.KafkaConnectList;
 import io.strimzi.api.kafka.KafkaConnectorList;
 import io.strimzi.api.kafka.model.AbstractKafkaConnectSpec;
-import io.strimzi.api.kafka.model.KafkaConnect;
 import io.strimzi.api.kafka.model.KafkaConnectResources;
-import io.strimzi.api.kafka.model.KafkaConnectSpec;
 import io.strimzi.api.kafka.model.KafkaConnector;
 import io.strimzi.api.kafka.model.KafkaConnectorBuilder;
 import io.strimzi.api.kafka.model.KafkaConnectorSpec;
@@ -52,6 +48,7 @@ import io.strimzi.operator.cluster.operator.resource.ResourceOperatorSupplier;
 import io.strimzi.operator.common.AbstractOperator;
 import io.strimzi.operator.common.Annotations;
 import io.strimzi.operator.common.BackOff;
+import io.strimzi.operator.common.OperatorWatcher;
 import io.strimzi.operator.common.Reconciliation;
 import io.strimzi.operator.common.ReconciliationLogger;
 import io.strimzi.operator.common.Util;
@@ -86,6 +83,7 @@ import java.util.Optional;
 import java.util.Set;
 import java.util.TreeMap;
 import java.util.function.BiFunction;
+import java.util.function.Consumer;
 import java.util.function.Function;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
@@ -128,6 +126,8 @@ public abstract class AbstractConnectOperator<C extends KubernetesClient, T exte
     protected final PlatformFeaturesAvailability pfa;
     protected final ServiceAccountOperator serviceAccountOperations;
     private final int port;
+    private final ClusterOperatorConfig config;
+    private Watcher<KafkaConnector> connectorWatcher;
 
     /**
      * Constructor
@@ -165,6 +165,7 @@ public abstract class AbstractConnectOperator<C extends KubernetesClient, T exte
         this.operatorNamespaceLabels = config.getOperatorNamespaceLabels();
         this.pfa = pfa;
         this.port = port;
+        this.config = config;
     }
 
     @Override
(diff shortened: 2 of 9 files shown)


cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/AbstractConnectOperator.java
```diff
@@ -198,96 +199,87 @@ public ConnectOperatorMetricsHolder metrics()   {
...
-                                // Check whether a KafkaConnect exists
-                                connectOperator.resourceOperator.getAsync(connectNamespace, connectName)
-                                        .compose(connect -> {
-                                            KafkaConnectApi apiClient = connectOperator.connectClientProvider.apply(connectOperator.vertx);
-                                            if (connect == null) {
-                                                Reconciliation r = new Reconciliation("connector-watch", connectOperator.kind(),
-                                                        kafkaConnector.getMetadata().getNamespace(), connectName);
-                                                updateStatus(r, noConnectCluster(connectNamespace, connectName), kafkaConnector, connectOperator.connectorOperator);
-                                                LOGGER.infoCr(r, "{} {} in namespace {} was {}, but Connect cluster {} does not exist", connectorKind, connectorName, connectorNamespace, action, connectName);
+        connectorWatcher =  new OperatorWatcher<KafkaConnector>(this, namespace, onClose) {
```
scholzj, line 212: Does it really make sense to use the OperatorWatcher? You basically throw away the whole logic it has. So IMHO this is a bad reuse. This deserves another logic or another level of abstraction there.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaConnectAssemblyOperator.java
```diff
@@ -359,4 +365,26 @@ Future<Integer> generateAuthHash(String namespace, KafkaConnectSpec kafkaConnect
...
+     * Create Kubernetes watch.
+     *
+     * @param namespace Namespace where to watch.
+     * @param onClose Callback called when the watch is closed.
+     *
+     * @return A future which completes when the watcher has been created.
+     */
+    public Future<Watch> createWatch(String namespace, Consumer<WatcherException> onClose) {
+        watcher = new OperatorWatcher<KafkaConnect>(this, namespace, onClose);
+        return VertxUtil.async(vertx, () -> resourceOperator.watch(namespace, selector(), watcher));
```
scholzj, line 379: Why is this here? It replaces the existing method in the upper classes without any obvious reason.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceAssemblyOperator.java
```diff
@@ -1463,4 +1430,40 @@ protected Future<Boolean> delete(Reconciliation reconciliation) {
...
+    }
+
+    /**
+     * Get watcher.
+     *
+     * @return watcher.
+     */
+    public Watcher getWatcher() {
+        return watcher;
+    }
```
scholzj, line 1468: Same as above, why do you have to override the createWatch method? That breaks the abstraction.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/AbstractConnectOperator.java
scholzj: I'm not really sure which part would you extract. Keep in mind that the logic is shared between Connect and MM2 and the connectors do not really have their own reconciliation. (Not saying that there is nothing to extract - just not sure it is obvious what to extract)

kyguy: > I think we could extract the KafkaConnector reconciliation logic and copy into its own KafkaConnectorAssemblyOperator

Was experimenting with this before, but reverted due to the things Jakub mentioned above. Mess but not impossible. Also had the idea of creating a `KafkaConnectorWatch` class like we have a `NodePoolWatch`.  I'll experiment with this a little more

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceAssemblyOperator.java
kyguy: The inherited `createWatch` method was not being used by the `KafkaRebalanceAssemblyOperator` class, `createRebalanceWatch` was used in its place with some extra logic. Since `createRebalanceWatch` in `KafkaRebalanceAssemblyOperator` is doing the same thing as `createWatch` method in cousin methods e.g.  `kafkaAssemblyOperator` and `kafkaMirrorMakerAssemblyOperator` it seemed like it would make more sense if we removed `createRebalanceWatch`  in favor of `createWatch` with the extra logic there. 

This allows changes like the following in the ClusterOperator class, adding kafkaRebalanceAssemblyOperator to the list of operators instead of treating it differently from the other operators:
```
      if (!config.isPodSetReconciliationOnly()) {
            List<AbstractOperator<?, ?, ?, ?>> operators = new ArrayList<>(asList(
                    kafkaAssemblyOperator, kafkaMirrorMakerAssemblyOperator,
                    kafkaConnectAssemblyOperator, kafkaBridgeAssemblyOperator, kafkaMirrorMaker2AssemblyOperator,
                    kafkaRebalanceAssemblyOperator));
            for (AbstractOperator<?, ?, ?, ?> operator : operators) {
                startFutures.add(operator.createWatch(namespace, operator.recreateWatch(namespace)).compose(w -> {
                  ...
            }
```

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaConnectAssemblyOperator.java
kyguy: It is written this way to extract the watcher used to create the watch. The watcher is extracted for testing purposes, so we can see how the operator behaves with a WatcherException when we run onClose(e) on the watcher

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceAssemblyOperator.java
scholzj: But why not use the original `createWatch`? This way, with its own implementation, it is not like the other operators, bu quite hidden.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaConnectAssemblyOperator.java
scholzj: And how are the other watches tested without storing the watch? Maybe if they don't have a test, you do not need one either. Or you should add it for all of them (but I guess simulating the watch closing with some exception to test it is quite hard so that is why it has no test).

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceAssemblyOperator.java
kyguy: I figured that the extra logic taken from the `createRebalanceWatch` may be important so didn't want to remove it

scholzj: What is the extra logic? Is it really needed? The point of the abstraction is to try to use the same method. Not to override it everywhere - that just makes it harder to find the differences and the places which for example need to be updated. So maybe its better to have a special method for it and have it clear that it is different.

kyguy: > What is the extra logic? 

Just the evenReceived() method changes in the diff above 

> Is it really needed?

Not sure exactly how important it is, I'll take a deeper look into why they were added

> Not to override it everywhere - that just makes it harder to find the differences and the places which for example need to be updated. So maybe its better to have a special method for it and have it clear that it is different.

I understand, I figured overriding it just here and the AbstractConnectAssemblyOperator would be better than having special methods e.g. we know that createWatch() is the method we need to search for to know how the watch is created across operators even if a couple have slightly different implementations

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaConnectAssemblyOperator.java
kyguy: There are not any tests for the other operators but I was thinking of adding one for them. Let me take a closer look into this, if I can align the onCreate() method of the KafkaRebalanceAssemblyOperator and the KafkaConnectorAssemblyOperator to the parent class, we could have one test that covers all operators

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/AbstractConnectOperator.java
```diff
@@ -198,96 +199,87 @@ public ConnectOperatorMetricsHolder metrics()   {
...
+     *
+     * @param namespace Namespace where to watch.
+     * @param onClose Callback called when the watch is closed.
      *
      * @return A future which completes when the watch has been set up.
      */
-    public static Future<Watch> createConnectorWatch(AbstractConnectOperator<KubernetesClient, KafkaConnect, KafkaConnectList, Resource<KafkaConnect>, KafkaConnectSpec, KafkaConnectStatus> connectOperator,
-                                                     String watchNamespaceOrWildcard, Labels selectorLabels) {
+    public Future<Watch> createConnectorWatch(String namespace, Consumer<WatcherException> onClose) {
+        Labels selectorLabels =  config.getCustomResourceSelector();
```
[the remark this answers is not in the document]
kyguy: Ah right, I just wanted to avoid the extra parameter to have the method signature as close to the createWatch() method as possible, but you are right, retaining the extra state it probably not worth it. Since the config is already passed to the constructor maybe it would satisfy us both to just store the selectorLabels in the object

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaConnectAssemblyOperator.java
scholzj: Maybe the tests are not there because it is hard to test it. To be honest, I'm not sure what exactly the tests you have test. You don't really trigger a real watch exception. You just call the onClose handler. But you do not have control over the original watch nor over the new one. So I think your test does not prove anything really. The only way to really test it might be to mock the socket or mock the actual watch code inside Fabric8.

kyguy: But from the partial stack trace and report from the user's cluster in the issue we know that onClose() is called with an exception, more specifically a `KubernetesClientException` and that the watch is not recreated. If we know this, isn't the test enough to show how the the operator recovers from having its onClose() method closed by an exception like this is, even if we are throwing it ourselves in the test?

kyguy: Maybe it would make more sense if we threw a `KubernetesClientException` instead of the `WatcherException` to be more closely represent the issue?

scholzj: It is different whether you call the onClose manually or whether the watch calls it. For example in what the state of the original watch is.

scholzj:
@kyguy I think it is not a bad idea that you are trying to bring these watches closer to how the other work. But sometimes, the things just don't fit and trying to make them fit by force does not always work. I think you can choose to keep them separate and just fix the reconnection. Or maybe the right way is to rework all of them and their abstraction to make the code more reusable and less dependent on the assembly operator code.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/AbstractConnectOperator.java
fvaleri: Yes, KafkaNodePoolWatcher seems a reasonable approach, because the relationship Kafka-NodePools and KafkaConnect-KafkaConnector are similar.

kyguy: Refactored the logic into a `KafkaConnectorWatcher` class, let me know what you think

kyguy:
@scholzj I removed the tests and refactored the KafkaConnector and KafkaRebalance watches a little bit to fix the reconnection. Let me know what you think

fvaleri (commented):
Hi @kyguy I like the custom watcher approach. Left few comments.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceAssemblyOperator.java
```diff
@@ -211,6 +209,78 @@ protected String cruiseControlHost(String clusterName, String clusterNamespace)
...
     }
 
+    /**
+     * Recreates a Kubernetes watch
+     *
+     * @param namespace Namespace which should be watched
+     *
+     * @return  Consumer for a Watched exception
+     */
+    public Consumer<WatcherException> recreateWatch(String namespace) {
```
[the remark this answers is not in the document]
kyguy: Thanks for the catch Fede, that is stale code, just removed

cluster-operator/src/main/java/io/strimzi/operator/cluster/ClusterOperator.java
```diff
@@ -136,11 +136,17 @@ public void start(Promise<Void> start) {
...
             }
 
-            startFutures.add(kafkaConnectAssemblyOperator.createConnectorWatch(namespace, kafkaConnectAssemblyOperator.recreateWatch(namespace)).compose(w -> {
+            startFutures.add(kafkaConnectAssemblyOperator.createKafkaConnectorWatch(namespace, config.getCustomResourceSelector()).compose(w -> {
                 LOGGER.info("Opened KafkaConnector watch for {} operator", kafkaConnectAssemblyOperator.kind());
                 watchByKind.put(KafkaConnector.RESOURCE_KIND, w);
                 return Future.succeededFuture();
             }));
+
+            startFutures.add(kafkaRebalanceAssemblyOperator.createRebalanceWatch(namespace, kafkaRebalanceAssemblyOperator.recreateRebalanceWatch(namespace)).compose(w -> {
```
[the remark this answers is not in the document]
kyguy: Since the logic of the `createRebalanceWatch` and `recreateRebalanceWatch` of the KafkaRebalanceAssemblyOperator is quite different from the `createWatch` and `recreateWatch` of the AbstractAssemblyOperator, I removed it based on Jakub's feedback.

We could of course move the `KafkaRebalanceAssemblyOperator` watch into the standard `createWatch` loop but it would require us to move the logic of:

- `createRebalanceWatch` -> `createWatch`
- `recreateRebalanceWatch` -> `createWatch`

overriding the inherited `createWatch` and `createWatch` methods from the AbstractAssemblyOperator in the KafkaRebalanceAssemblyOperator class. It doesn't matter to me which path we take, just want everyone to agree on the direction we take

If you feel strongly about it, I'll revist with @scholzj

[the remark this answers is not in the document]
fvaleri: Thanks for the explanation. To me it's just a matter of avoiding code duplication. For example, recreateRebalanceWatch and super.recreateWatch looks exactly the same. Then, createRebalanceWatch looks similar to the standard use case, but not exactly the same, so maybe we could address this in another dedicated PR.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/AbstractConnectOperator.java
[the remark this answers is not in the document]
kyguy: Removed

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaConnectAssemblyOperator.java
kyguy: Removed

fvaleri (commented):
@kyguy I think this comment didn't get noticed. The rest is good for me.

> For example, recreateRebalanceWatch and super.recreateWatch looks the same.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaConnectorWatcher.java
```diff
@@ -0,0 +1,163 @@
...
+
+/**
+ * Watcher implementation for watching Kafka connectors and triggering reconciliations based on changes to Kafka
+ * Connector configurations.
+ */
+public class KafkaConnectorWatcher implements Watcher<KafkaConnector> {
+    private static final ReconciliationLogger LOGGER = ReconciliationLogger.create(KafkaNodePoolWatcher.class);
+
+    private final Optional<LabelSelector> selector;
+    private final AbstractConnectOperator<KubernetesClient, KafkaConnect, KafkaConnectList, Resource<KafkaConnect>, KafkaConnectSpec, KafkaConnectStatus> connectOperator;
```
scholzj, line 44: Any reason why this is not `KafkaConnectAssemblyOperator`?

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaConnectorWatcher.java
```diff
@@ -0,0 +1,163 @@
...
+    public static void updateStatus(Reconciliation reconciliation, Throwable error, KafkaConnector kafkaConnector2, CrdOperator<?, KafkaConnector, ?> connectorOperations) {
+        KafkaConnectorStatus status = new KafkaConnectorStatus();
+        StatusUtils.setStatusConditionAndObservedGeneration(kafkaConnector2, status, error);
+        StatusDiff diff = new StatusDiff(kafkaConnector2.getStatus(), status);
+        if (!diff.isEmpty()) {
+            KafkaConnector copy = new KafkaConnectorBuilder(kafkaConnector2).build();
+            copy.setStatus(status);
+            connectorOperations.updateStatusAsync(reconciliation, copy);
+        }
+    }
```
scholzj, line 162: This seems a bit weird. You seem to have the same method in the `connectOperator`. Why don't you just call it there? You seem to duplicate it.

cluster-operator/src/main/java/io/strimzi/operator/cluster/ClusterOperator.java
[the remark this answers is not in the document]
scholzj: I probably don't get what is different here. It should ideally unify and use the `createWatch`. If it needs to be different, then I personally would prefer to not override to make it more visible that it is different.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaConnectorWatcher.java
```diff
@@ -0,0 +1,163 @@
...
+                                        return Future.succeededFuture();
+                                    } else if (connect.getSpec() != null && connect.getSpec().getReplicas() == 0)  {
+                                        LOGGER.infoCr(reconciliation, "{} {} in namespace {} was {}, but Connect cluster {} has 0 replicas", connectorKind, connectorName, connectorNamespace, action, connectName);
+                                        updateStatus(reconciliation, connectOperator.zeroReplicas(connectNamespace, connectName), kafkaConnector, connectOperator.connectorOperator);
+                                        return Future.succeededFuture();
+                                    } else {
+                                        LOGGER.infoCr(reconciliation, "{} {} in namespace {} was {}", connectorKind, connectorName, connectorNamespace, action);
+
+                                        return connectOperator.withLock(reconciliation, LOCK_TIMEOUT_MS,
```
scholzj, line 110: Does this need to call it this way instead of calling just `reconcile`?

scholzj (commented):
I think this would probably work. But it seems like we are just flicking the issues by adding a dirty patch here and there. We will now have several different solutions for this in a single code base:
* Informers in StrimziPodSetController, which is probably ine since that is separate code base.
* Kafka node pool watch as a separate class
* Kafka connector watch as a separate class (following the node pool watch logic, never the less not sharing any code)
* The KafkaRebalance watch which is different then anything else
* The watches used by all the assembly operators other then the KafkaRebalance one

So we will have 5 different ways to watch resources - 4 of them just different solutions based on a watch if we do not want to count the informers.

I think that is quite bad. We should come up with a better solution:
* Use informers everywhere (they automatically reconnect, but we just had major issues with them)?
* Refactor completely how we use the watches. For example something along the direction of:
    * Have one new Abstract watch to handle the reconnecting but abstracts the handler to replace the 4 current implementations
    * Have a pluggable event handler to plug into it (which is really where the code seems to differ most between the different watches)

I guess the main question is if we should try to get directly tot he right solution, or if we should first get this fixed in a fairly hacky way and then get back to it and do the proper solution separately.

One of the reasons for trying to get a proper solution right away is that this seems to change a visibility of many fields and methods and that is something what would be quite hard to fix back.

kyguy:
> For example, recreateRebalanceWatch and super.recreateWatch looks the same.

@fvaleri they are identical except that:
- `recreateRebalanceWatch` calls `rebalanceWatch` 
- `super.recreateWatch` calls `createWatch`

Then `rebalanceWatch` is quite different as it requires a call to `reconcileRebalance`. It would take significant refactor to `KafkaRebalanceAssemblyOperator` to remove the need for this call.

> I guess the main question is if we should try to get directly tot he right solution, or if we should first get this fixed in a fairly hacky way and then get back to it and do the proper solution separately.

@scholzj It doesn't matter to me. TBH I don't see the closed watcher issue happening that often and the workaround for the issue is not that difficult to preform, so delaying the fix for a more suitable, non-hack solution is ok with me.

@scholzj @fvaleri Let me know what you think, if you both agree, I can start investigating one of the alternative solutions

fvaleri:
@kyguy I agree. Using informers with a shared abstraction would be the best.

scholzj:
I think I need to think a bit more about it. The informers have some overhead to make sure they are running. If we want to use them, it would be also good to think a bit more about the use of indexers.

scholzj:
@kyguy @fvaleri I created https://github.com/strimzi/strimzi-kafka-operator/pull/8969 which I think is a slightly better alternative to this and offers a little bit cleaner solution. It should buy us more time to properly think through how informers and indexers might fit in. Please have a look.

scholzj:
Thanks a lot for the effort @kyguy. I think we can now close this.

PR #8817, opened 2023-07-11
Add support for node pools to the report script

fvaleri:
Tested on bash 5 and bash 4 (the older is what we document).
I also ran the script with and without NodePools and also the experimental KRaft mode.
Added a check on bash version.

### DIFF
--- a/tools/report.sh
+++ b/tools/report.sh
@@ -5,7 +5,7 @@ if [[ $(uname -s) == "Darwin" ]]; then
   alias echo="gecho"; alias grep="ggrep"; alias sed="gsed"; alias date="gdate"
 fi
 
-{ # this ensures the entire script is downloaded #
+{ # this ensures that the entire script is downloaded #
 NAMESPACE=""
 CLUSTER=""
 BRIDGE=""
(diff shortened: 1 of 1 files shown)


scholzj (approved):
LGTM. And seems to work fine on MacOS. Thanks for the PR @fvaleri!

ppatierno:
Not related with these changes, but should we remove the `statefulsets` from the list of resources? As per 0.35.0 release we don't have support for them anymore.

ppatierno:
Still not related to these changes but if I try with a not existing namespace, this code seems not working:

```shell
if [[ $($KUBE_CLIENT get ns "$NAMESPACE" &>/dev/null) == "1" ]]; then
  error "Namespace $NAMESPACE not found! Exiting"
fi
```

Good namespace but wrong cluster name works as expected instead.

scholzj:
> Not related with these changes, but should we remove the statefulsets from the list of resources? As per 0.35.0 release we don't have support for them anymore.

We do not version this script. So it will be used wth older versions as well. So I would keep the StatefulSets for few more releases.

> Still not related to these changes but if I try with a not existing namespace, this code seems not working

This seems to be in the original code as well. So not sure we want to fix it in this PR. But I also get errors about a non-existent Kafka cluster and not about a non-existing namespace.

ppatierno (approved):
LGTM.

ppatierno:
> @ppatierno this is an easy fix, let me push it. Thanks for raising.

Thanks! Tried and works ;-)

PR #8820, opened 2023-07-11
Adding thread name to the bridge logging layout pattern

ppatierno:
This trivial PR just adds the thread name to the logging layout pattern for the bridge as already done on the Strimzi HTTP bridge repo as well with this PR https://github.com/strimzi/strimzi-kafka-bridge/pull/818


scholzj (commented):
Maybe you should fix the name / description? You are not really fixing any broker pattern. I this repo you are just adding threads to the output.

PR #8824, opened 2023-07-12
Update shellcheck version

Vaibhav-Nazare:
Fix for report_too.sh 
Expanding an array without an index only gives the element in the index 0.
```
./.azure/scripts/shellcheck.sh

In tools/report.sh line 34:
if [[ "$BASH_VERSINFO" -lt 4 ]]; then
^------------^ SC2128: Expanding an array without an index only gives the first element.
```


scholzj:
@Vaibhav-Nazare What shellcheck version are you using? Maybe you should sync it with the Strimzi CI which uses 0.7.2?

fvaleri (commented):
Jakub is right, you are probably using shellcheck 0.7.1 or lower:

```sh
$ ./shellcheck-v0.7.1/shellcheck ~/Documents/strimzi-kafka-operator/tools/report.sh 

In /home/fvaleri/Documents/strimzi-kafka-operator/tools/report.sh line 34:
if [[ "$BASH_VERSINFO" -lt 4 ]]; then
       ^------------^ SC2128: Expanding an array without an index only gives the first element.

For more information:
  https://www.shellcheck.net/wiki/SC2128 -- Expanding an array without an ind...
```

Vaibhav-Nazare:
I did check this at our end it's higer
```
shellcheck -V
ShellCheck - shell script analysis tool
version: 0.9.0
license: GNU General Public License, version 3
website: https://www.shellcheck.net
```

scholzj:
Right. Then I think there are in general 2 options:
* You downgrade yours to 0.7.2 (instead of fixing the issue 0.9.0 reports). 
* Or you upgrade the one in our CI pipelines to 0.9.0 as well (plus fix the issue it reports). 

But running different versions and opening a PR each time it finds something different does not make sense. It might even happen that one day it has some conflicting issues. So bringing the versions in-sync one way or another will prevent this kind of issue from reappearing again and again.

I do not have a strong preference for which one to use - I guess updating the shellcheck version to 0.9.0 here might make more sense.

Vaibhav-Nazare:
Yes upgrading it too 0.9.0 makes more sense

fvaleri:
I also agree that we should upgrade shellcheck to 0.9.0 and fix the issue.

Note that the same check is also used in `docker-images/artifacts/build.sh` and `docker-images/kafka-based/build.sh`.

scholzj (approved):
Thanks for the changes!

tools/report.sh
```diff
@@ -31,7 +31,7 @@ error() {
 if [[ -z ${BASH_VERSINFO+x} ]]; then
   error "No bash version information available, aborting"
 fi
-if [[ "$BASH_VERSINFO" -lt 4 ]]; then
+if [[ "${BASH_VERSINFO[*]}" -lt 4 ]]; then
```
fvaleri, line 34: This is not correct, as we only care about the major version, so index 0. The same in the other places.

```sh
$ if [[ "${BASH_VERSINFO[*]}" -lt 4 ]]; then
  error "You need bash version >= 4 to run the script"
fi
bash: [[: 5 2 15 1 release x86_64-redhat-linux-gnu: syntax error in expression (error token is "2 15 1 release x86_64-redhat-linux-gnu")
```

fvaleri (changes requested):

ppatierno (approved):
LGTM.Thanks for the PR!

fvaleri (approved):
Thanks.

scholzj:
Thanks for the PR

PR #8826, opened 2023-07-12
[systemtest] Update scalibility profile - add recovery group

im-konge:
### Type of change

- Enhancement

### Description

This PR adds `recovery` group into `scalability` profile and renames the profile to `scalability_recovery`.
The profile is used in our automation, where we are running these two groups together, so I'm adding this change to handle it on a profile level.

### Checklist

### DIFF
--- a/systemtest/pom.xml
+++ b/systemtest/pom.xml
@@ -546,11 +546,12 @@
         </profile>
 
         <profile>
-            <id>scalability</id>
+            <id>scalability_recovery</id>
             <properties>
                 <skipTests>false</skipTests>
                 <groups>
-                    all,scalability
+                    scalability,
+                    recovery
                 </groups>
             </properties>
         </profile>


scholzj (commented):
Does this really make any sense to mix those two groups? Or are you just misusing it for convenience?

im-konge:
> Or are you just misusing it for convenience?

Yeah, mainly for it.
We are running those two groups in a separate pipeline that is executed on a weekly basis.
Maybe having different name would be better

scholzj:
Why can't you specify the groups from the command to run the tests directly? I think it does not make any sense to hardcode what fits a particular CI into the `pom.xml`. Because what would you do if someone comes that they run it differently and want to add profiles for _their setup_? I do not think you want to have the CI details inside the `pom.xml`.

im-konge:
Those two groups are not part of the `regression`, so having a separate profile for them makes sense FMPOV. 
`scalability` group cannot be part of the `regression`, because it needs more resources that in AZP are not available - so the tests fail.
`recovery` contains few tests that don't have to be part of the `regression` and those can be run once in a while

scholzj:
TBH, I think none of the profiles make sense. I just did not notice it before. The regression ones make no sense either. As I said - imagine that someone comes and says they run their own regression pipeline but use just 5 parallel streams and they want to add their own five profiles for their pipelines next to the existing ones. And then someone else comes and says for example that they run scalability, recovery, and bridge tests together and want to add profile names `scalability_recovery_bridge` etc.? Will you merge that and maintain all of these profiles?

im-konge:
The separation of the regression tests is used from the past in AZP, Jenkins and everywhere else.
The profiles just create encapsulation of the groups, included and excluded testcases etc.
When we want to add a package to `components` or `operators` regression profiles, we do it in one place.
If it doesn't make sense, we can remove all the profiles and come back with the solution of executing jobs with groups, included and excluded testcases.
And then, when we need to change something, we will change it on multiple places.

These profiles are used by automation in this repository, so having profiles of different teams that maintain some CI doesn't makes sense to me TBH.

scholzj:
I don't think it matters if the CI is part of this repository or not because you inject its configuration into the actual software configuration. So I think the right solution is to configure the groups directly in the CIs. I'm fine if you wanna merge this and open and issue and remove the profiles later. But I think what we have right now is wrong as it pulls the configuration of a particular CI into the software itself.

im-konge:
Okay and what about AZP? We are using the `azp_*` profiles for running the specific subset of tests.
Should those be removed too and use groups + exclusion of tests?
Or should we add new tags for these subsets directly into the STs?

scholzj:
I think that is the same - it should be part of the AZP files.

scholzj:
As this PR shows, this is not about having a generic profile for running a particular set of tests. This is just artificial configuration to match the requirement of one particular CI where you want to run some tests together. I never saw anything like this in other Maven projects. It has nothing to do with the software itself. It is purely a configuration of the CI (which is not part of the Maven project / Java code).

im-konge:
If it is about the `scalability`/`scalability_recovery` - I agree with removing it.
I would remove also `bridge` profile (and more, I don't want to provide full list here).
I would keep other profiles.

> I never saw anything like this in other Maven projects.

I saw it in other Maven projects that have STs in the same repository as the product/operator.
But also in projects that are just for tests - for the logical separation as I mentioned before.

scholzj:
But when you split the tests into 7 different pipelines because that is what you have in AZP then it is not logical splitting. It is CI configuration.

kornys:
> As this PR shows, this is not about having a generic profile for running a particular set of tests. This is just artificial configuration to match the requirement of one particular CI where you want to run some tests together. I never saw anything like this in other Maven projects. It has nothing to do with the software itself. It is purely a configuration of the CI (which is not part of the Maven project / Java code).

Well it is not purely a configuration of the CI. You can run in locally without specifying bunch of the test or test-groups (labels). So with profile it is much faster way to run maven locally without remembering all test-groups.

Frawless:
I think it makes sense to keep profiles like `upgrade`, `regression`, `operators`, etc. These are pretty common and I don't see any reason why they shouldn't be part of systemtest profiles. However `azp_` profiles are a little bit strange a makes sense to me to keep it as a CI configuration. In case the profiles will makes sense and we will have a proper reason why they shouldn't be there, I would keep them.

scholzj:
Yes, I think it is fine to keep the profiles for running some logical group of tests such as scalability tests or upgrade tests. But it should be based on the logical grouping of the tests and not on the fact that right now the CI wants to run scalability and recovery together.

im-konge:
I agree that scalability + recovery doesn't makes sense as a common profile. So I will close this PR.
And regarding AZP -> I will check if there is a way to keep these profiles in one place and possibly remove it from the `pom.xml`.

Issue #8823, opened 2023-07-11
[Bug]: Getting error with loadBalancerClass


girishms-sentient:
### Bug Description

The loadBalancerClass error occurs when I try to create a Kafka cluster with ArgoCD.

Is there support for loadBalancerClass in strimzi?

### Strimzi version

0.35.0

### Kubernetes version

K8s 1.25

### Installation method

Helm chart with ArgoCD

### Infrastructure

Amazon EKS

### Configuration files and logs

`Failure executing: PATCH at: [https://172.20.0.1:443/api/v1/namespaces/central-logging/services/kafka-kafka-external-bootstrap](https://172.20.0.1/api/v1/namespaces/central-logging/services/kafka-kafka-external-bootstrap). Message: Service "kafka-kafka-external-bootstrap" is invalid: spec.loadBalancerClass: Invalid value: "null": may not change once set. Received status: Status(apiVersion=v1, code=422, details=StatusDetails(causes=[StatusCause(field=spec.loadBalancerClass, message=Invalid value: "null": may not change once set, reason=FieldValueInvalid, additionalProperties={})], group=null, kind=Service, name=kafka-kafka-external-bootstrap, retryAfterSeconds=null, uid=null, additionalProperties={}), kind=Status, message=Service "kafka-kafka-external-bootstrap" is invalid: spec.loadBalancerClass: Invalid value: "null": may not change once set, metadata=ListMeta(_continue=null, remainingItemCount=null, resourceVersion=null, selfLink=null, additionalProperties={}), reason=Invalid, status=Failure, additionalProperties={}).`

scholzj:
You should probably set your Load Balancer Class in the `Kafka` custom resource: https://strimzi.io/docs/operators/latest/full/configuring.html#type-GenericKafkaListenerConfiguration-schema-reference (look for the `class` field).

girishms-sentient:
Are you referring to this?

For example: 
```
    listeners:
      - name: external
        port: 9092
        type: loadbalancer
        loadBalancerClass: service.k8s.aws/nlb
```

scholzj:
Almost I guess. If you check the schema reference the filed name is `class`.

scholzj:
One curious question -> did you configure the Load Balancer class on the service yourself (e.g. manually using `kubectl edit`)? Or did AWS inject it automatically?

girishms-sentient:
@scholzj  AWS injected it automatically. 

[...]

PR #8828, opened 2023-07-12
Don't patch loadBalancerClass field in Service resources

scholzj:
### Type of change

- Bugfix

### Description

The `loadBalancerClass` field might be in some environments set by the platform. In such case, we should reuse it instead of trying to patch it. This PR will patch it if it is set in the existing resource and not in the desired resource to avoid errors such as the one described in #8823.

### Checklist

- [x] Write tests


ppatierno (approved):
LGTM

PR #8832, opened 2023-07-12
[systemtest] Disable whole NamespaceDeletionRecoveryIsolatedST for KRaft (and currently also for regression)

im-konge:
### Type of change

- Bugfix

### Description

This PR disables whole `NamespaceDeletionRecoveryIsolatedST` when running in KRaft mode, as both of the tests need TopicOperator, which is not available in KRaft (UTO is not yet tested, but if the scenario will work, we should enable it back with usage of it).

### Checklist


scholzj (approved):

jankalinic (approved):

jankalinic:
Thanks!

henryZrncik (approved):

PR #8870, opened 2023-07-20
[systemtest] Add assumption for Kafka version to testKRaftMode test

im-konge:
### Type of change

- Bugfix

### Description

This PR adds a simple assumption for `testKRaftMode` test for Kafka version -> the version needs to be `3.5.0` or higher for enabling the test. From Kafka version `3.5.0`, the SCRAM-SHA is available to use in KRaft.
When we enable auth (TLS/SCRAM-SHA) and we want to create a KafkaUser, it needs to create `ScramShaCredentialsCache` in UO. And because test is using TLS, it simply fails on older versions of Kafka:

```
  status:
    conditions:
    - lastTransitionTime: "2023-07-20T04:23:54.939921854Z"
      message: 'java.util.concurrent.CompletionException: java.lang.RuntimeException:
        ScramShaCredentialsCache is not ready!'
      reason: ExecutionException
      status: "True"
      type: NotReady
    observedGeneration: 1
```

### Checklist

### DIFF
--- a/systemtest/src/test/java/io/strimzi/systemtest/operators/FeatureGatesIsolatedST.java
+++ b/systemtest/src/test/java/io/strimzi/systemtest/operators/FeatureGatesIsolatedST.java
@@ -79,6 +79,7 @@ public class FeatureGatesIsolatedST extends AbstractST {
     @Tag(INTERNAL_CLIENTS_USED)
     public void testKRaftMode(ExtensionContext extensionContext) {
         assumeFalse(Environment.isOlmInstall() || Environment.isHelmInstall());
+        // skip test if KRaft mode is enabled and Kafka version is lower than 3.5.0 - https://github.com/strimzi/strimzi-kafka-operator/issues/8806
         assumeTrue(TestKafkaVersion.compareDottedVersions("3.5.0", Environment.ST_KAFKA_VERSION) != 1);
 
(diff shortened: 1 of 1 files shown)


scholzj (approved):

systemtest/src/test/java/io/strimzi/systemtest/operators/FeatureGatesIsolatedST.java
```diff
@@ -77,6 +79,8 @@ public class FeatureGatesIsolatedST extends AbstractST {
     @Tag(INTERNAL_CLIENTS_USED)
     public void testKRaftMode(ExtensionContext extensionContext) {
         assumeFalse(Environment.isOlmInstall() || Environment.isHelmInstall());
+        assumeTrue(TestKafkaVersion.compareDottedVersions("3.5.0", Environment.ST_KAFKA_VERSION) != 1);
```
scholzj, line 82: Maybe you can add the comment as well so that we find it based on the issue number?

im-konge: Yeah, good catch

Frawless (approved):

see-quick (approved):

Issue #8864, opened 2023-07-19
[systemtest] Use Admin client for checking topics in Kafka when UTO is used


im-konge:
In some STs, we check if spec of particular KafkaTopic has correct values -> in CR.
But in situations when the KafkaTopic is created just in Kafka (so there is no KafkaTopic CR), with usage of UTO, the KafkaTopic is not created (synced) from Kafka to KafkaTopic CR.
In these situations, we should check the spec of KafkaTopic directly in Kafka, for example using the test-clients' Admin client.

The test checks are currently disabled for UTO, but we should fix this issue and enable the checks back.
For enabling the checks, search for link of this issue.

fvaleri:
This includes tests with `UTONotSupported` annotation and/or "TODO https://github.com/strimzi/strimzi-kafka-operator/issues/8864" comment.

henryZrncik:
Closed as completed in https://github.com/strimzi/strimzi-kafka-operator/pull/9295 and https://github.com/strimzi/strimzi-kafka-operator/pull/9876

Issue #8863, opened 2023-07-19
[systemtest] Enable tests containing checks for metrics in UTO


im-konge:
Currently, Unidirectional Topic Operator has no metrics.
Until the metrics part of the UTO will not be implemented, STs that contains checks for KafkaTopic's metrics inside TO (UTO) are skipped.
These tests should be enabled again once the feature will be available.

For enabling the tests, search for the link of this issue.

im-konge:
Done in #9240

PR #8874, opened 2023-07-20
[systemtest] Use UTO in KRaft tests

im-konge:
### Type of change

- Enhancement

### Description

This PR enables few tests that were disabled when running in KRaft mode -> and which contain usage of TO.
For enabling the test, I added few annotations, where we are checking, if the UTO is enabled with KRaft or not.

Because the UTO is creating KafkaTopics in one way -> so from KafkaTopic CR to Kafka -> I had to disable few tests (using `@UTONotSupported`). Also there are few changes to the KafkaTopic's status and exceptions + messages, that are in the KafkaTopic's status.

There are also testcases, that will need few more changes and also few tests that are checking TO's metrics (will be implemented in future PRs). For tracking these tests, I created two issues:

https://github.com/strimzi/strimzi-kafka-operator/issues/8863
https://github.com/strimzi/strimzi-kafka-operator/issues/8864

Finally, I added UTO FG to the KRaft pipelines in AZP.

### Checklist

### DIFF
--- a/.azure/templates/jobs/system-tests/kraft_regression_jobs.yaml
+++ b/.azure/templates/jobs/system-tests/kraft_regression_jobs.yaml
@@ -5,7 +5,7 @@ jobs:
       display_name: 'kraft-regression-bundle I. - kafka + oauth'
       profile: 'azp_kafka_oauth'
       cluster_operator_install_type: 'bundle'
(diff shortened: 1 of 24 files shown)


see-quick (approved):

tombentley (approved):
Thanks!

Frawless (approved):

PR #8875, opened 2023-07-20
Fix clean build errors on version bump

fvaleri:
Avoids make clean "could not find artifact" error after a version bump without a previous build.
Make clean command fails also when there is no strimzi artifact in the local Maven repository.

This happens because we miss the config-model artifacts in our local Maven repository.
This commit fixes the issue by first installing required dependencies if not found.

### DIFF
--- a/tools/prerequisites-check.sh
+++ b/tools/prerequisites-check.sh
@@ -38,3 +38,12 @@ if [[ ${YQ_ARRAY[0]} -eq 4  ]] && [[ ${YQ_ARRAY[1]} -eq 2 ]] && [[ ${YQ_ARRAY[2]
   echo -e "$yq_err_msg"
   exit 1
 fi
+
+# avoids make clean "could not find artifact" error after a version bump without a previous build
+# make clean command fails also when there is no strimzi artifact in the local Maven report
+PC_RELVER="$(cat ./release.version)"
+PC_M2HOME="$(mvn help:evaluate -Dexpression=settings.localRepository | grep -v '\[INFO\]')"
+if [[ ! -d "$PC_M2HOME/io/strimzi/config-model/$PC_RELVER" ]]; then
+  echo "Installing parent" && mvn install -N -q
+  echo "Installing config-model" && mvn install -pl config-model -DskipTests -q
+fi


fvaleri:
What's horrible is the dev experience when you first try to build Strimzi. This is just automating what one have to figure out and do manually. That said, I'll try to see if there is a better way.

scholzj:
> when you first try to build Strimzi

You mean when you try to clean the project before building it first, right?

fvaleri:
> > when you first try to build Strimzi
> 
> You mean when you try to clean the project before building it first, right?

Yes, a clean build. For example, one use case is after a rebase, and it fails if the version has been updated (this case). Another example is when testing a candidate release (that's not necessary, but maybe I don't know the project well and wants to avoid missing some cleaning done outside the project directory).

scholzj:
But it is not the build that fails. It is only clean which fails AFAIK.

tools/prerequisites-check.sh
```diff
@@ -38,3 +38,12 @@ if [[ ${YQ_ARRAY[0]} -eq 4  ]] && [[ ${YQ_ARRAY[1]} -eq 2 ]] && [[ ${YQ_ARRAY[2]
   echo -e "$yq_err_msg"
   exit 1
 fi
+
+# avoids make clean "could not find artifact" error after a version bump without a previous build
+# make clean command fails also when there is no strimzi artifact in the local Maven report
+PC_RELVER="$(cat ./release.version)"
+PC_M2HOME="$(mvn help:evaluate -Dexpression=settings.localRepository | grep -v '\[INFO\]')"
```
ppatierno, line 45: out of curiosity what does `PC` stand for?

tools/prerequisites-check.sh
```diff
@@ -38,3 +38,12 @@ if [[ ${YQ_ARRAY[0]} -eq 4  ]] && [[ ${YQ_ARRAY[1]} -eq 2 ]] && [[ ${YQ_ARRAY[2]
...
+if [[ ! -d "$PC_M2HOME/io/strimzi/config-model/$PC_RELVER" ]]; then
+  echo "Installing parent" && mvn install -N -q
```
ppatierno, line 47: aren't you installing everything from the root with this?

tools/prerequisites-check.sh
fvaleri: Prerequisite check, I usually add a simple prefix to limit the risk of overwriting other shared env vars.

tools/prerequisites-check.sh
fvaleri: No, just the parent thanks to the -N flag. This is needed when you run make clean with an empty local Maven repository.

fvaleri:
This hasn't hasn't received much feedback, and it is also creating some issue with the CI pipeline. If anyone is interested, the same workaround can be easily applied to any local dev environment (i.e. using your favorite shell profile).

PR #8877, opened 2023-07-20
[systemtest] Remove TF's sanity check from push to main and enable it in PRs (per-command)

im-konge:
### Type of change

- Enhancement

### Description

This PR removes the Testing Farm's `sanity` check when pushing to `main` branch and enables it for running the `sanity` profile in PRs - using `/packit test --labels sanity` command (so the profile will not be executed automatically).
Currently, there are a few flaky tests, so for every push to main branch the TF's pipelines are failing.
We should firstly stabilize the STs and then we can enable it again. 

### Checklist

### DIFF
--- a/.packit.yaml
+++ b/.packit.yaml
@@ -133,13 +133,15 @@ jobs:
   ###############################################################################################
 
   - job: tests
-    trigger: commit
-    branch: main
+    trigger: pull_request
     identifier: "sanity"
     targets:
       - centos-stream-9-x86_64
       - centos-stream-9-aarch64
     skip_build: true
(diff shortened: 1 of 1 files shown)


jankalinic (approved):

kornys (approved):

scholzj:
> enables it for running the sanity profile in PRs

Does this mean automatically or per-command?

im-konge:
> Does this mean automatically or per-command?
per-command
I will update the description to make it less confusing

Frawless (approved):

PR #8885, opened 2023-07-23
[systemtest] Fix for the upgrade AZP profiles in pom.xml and remove all leftovers after IsolatedSuite removal

im-konge:
### Type of change

- Bugfix

### Description

This PR fixes bug that was added in #8860, where the `Isolated` suites were removed -> this affected naming of the classes and because we are using the names of the STs in AZP profiles, it broke the upgrade pipeline.

-----------------------------------------------------------------------

Also, this PR removes leftovers after the aforementioned PR - `IsolatedSuite`, mentions in Javadoc.
 
### Checklist


scholzj (commented):
If you search the code, there seem to be some more places where `IsolatedST` is still used. E.g. some Javadoc or the testing README. That should be probably fixed as well. There seem to be also some `IsolatedSuite` annotation which is not used and seems to be a left over as well?

im-konge:
Ah yes, I thought that `IsolatedSuite` is used in UTs/ITs also, gonna remove it.
Thanks

see-quick (approved):

scholzj (approved):

scholzj:
(Sorry, I wanted to run upgrade tests but started regression by mistake)

Issue #8850, opened 2023-07-18
[Enhancement]: Liveness / readiness probe for Kraft mode 


tinaselenge:
### Related problem

The current live and readiness probes are based on checking the ZooKeeper connection first and then checking the broker state. Currently, the Strimzi Cluster Operator already supports the KRaft mode (behind a feature gate) and zookeeper is being deprecated soon, therefore checking ZooKeeper is probably not the right approach anymore. 

### Suggested solution

Implement [proposal #046](https://github.com/strimzi/proposals/blob/main/046-kraft-liveness-readiness.md) for liveness and readiness probes in a KRaft Kafka cluster.

The expected outcome based on the proposal for this task is:
```
Controller only node:
Liveness: the Java process in the container is running
Readiness: controller is listening on the port 9090 which is configured in controller.listener.names

Broker only node:
Liveness: broker is listening on the port 9091 which is configured in inter.broker.listener.name
Readiness: the BrokerState metric >=3 && != 127

Combined node:
Liveness: the Java process in the container is running
Readiness: the BrokerState metric >=3 && != 127
```

The proposed liveness probe for broker only mode requires a change because of [KAFKA-14658](https://issues.apache.org/jira/browse/KAFKA-14658). Before this upstream change, Kafka broker port is opened early typically before the log recovery phase. Based on this logic, the proposed liveness probe for broker only mode to listen on the port 9091 which is configured in inter.broker.listener.name. However, after the upstream change in [KAFKA-14658](https://issues.apache.org/jira/browse/KAFKA-14658), the broker starts listening only when it's ready to serve client requests it would mean long log recovery might cause Kube to restart, because the pod wouldn't appear to be alive.

This upstream change does not have an impact on the existing liveness check for zookeeper mode because the check passes if there is a ZK session.

The proposed liveness probe for controller and combined mode to have running Java process in the container. The reason for this was because of very similar issue. For controller mode, the log recovery happens before socket bind, which happens before socket listen (the LISTEN state), so if recovery of the __cluster_metadata log takes a long time then using the port for a liveness check risks Kubernetes killing the controller.

Therefore I think it makes sense to change broker only mode to have the same liveness probe (running Java process in the container) as the controller and combined mode. The expected outcome becomes:
```
Controller only node:
Liveness: the Java process in the container is running
Readiness: controller is listening on the port 9090 which is configured in controller.listener.names

Broker only node:
Liveness: the Java process in the container is running
[...]

PR #46, opened 2017-10-13
Logging

tombentley:
Untested, but some initial work on configuring logging of Kafka and Zookeeper to:

* only log to the console, not to files, and
* use a consistent date format at least.

See #21.

PR #8892, opened 2023-07-24
Liveness / readiness probe for Kraft mode phase 1 and 2

tinaselenge:
### Type of change
- Enhancement / new feature

### Description

This PR implements  [proposal #046](https://github.com/strimzi/proposals/blob/main/046-kraft-liveness-readiness.md) for adding liveness/readiness checks for KRaft mode. 

As explained in https://github.com/strimzi/strimzi-kafka-operator/issues/8850, due to the change in upstream, the proposed liveness probe for broker only mode was changed from "broker is listening on the port 9091" to "the Java process in the container is running", which is consistent with the proposed liveness probe for controller only and combined modes. With this change, phase 2 and phase 3 end up with not much difference therefore both phases are being implemented in this PR. 

This PR is also blocked behind KafkaRoller's update to check controller nodes, otherwise if the pods are initially unschedulable the KafkaRoller won't be able to create the pods successfully once appropriate Kubernetes nodes are available as stated in the proposal. 

Closes https://github.com/strimzi/strimzi-kafka-operator/issues/8850.

### Checklist

- [x] Write tests
- [ ] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/docker-images/kafka-based/kafka/scripts/kafka_liveness.sh
+++ b/docker-images/kafka-based/kafka/scripts/kafka_liveness.sh
@@ -2,8 +2,9 @@
 set -e
 
 if [ "$STRIMZI_KRAFT_ENABLED" = "true" ]; then
-  # Test KRaft controller process is running
-  . ./kafka_controller_liveness.sh
+  for proc in /proc/*[0-9];
+    do if readlink -f "$proc"/exe | grep -q java; then exit 0; fi;
+  done
 else
   # Test ZK-based broker liveness
   # We expect that either the broker is ready and listening on 9091 (replication port)
--- a/docker-images/kafka-based/kafka/scripts/kafka_readiness_kraft.sh
+++ b/docker-images/kafka-based/kafka/scripts/kafka_readiness_kraft.sh
@@ -1,0 +1,14 @@
+#!/usr/bin/env bash
+set -e
+
+file=/tmp/strimzi.properties
+test -f $file
+roles=$(awk -F "process.roles=" '{print $2}' "$file")
+if [[ "$roles" =~ "controller" ]] && [[ ! "$roles" =~ "broker" ]]; then
+  # For controller only mode, check if it is listening on port 9090 (configured in controller.listener.names).
+  netstat -lnt | grep -E 'tcp?[[:space:]]+[0-9]+[[:space:]]+[0-9]+[[:space:]]+[^ ]+:9090+[[:space:]]+[^ ]+[[:space:]]+LISTEN*'
+else
+  # For combined or broker only mode, check readiness via HTTP endpoint exposed by Kafka Agent.
+  # The endpoint returns 200 when broker state is 3 (RUNNING).
+  curl http://localhost:8080/v1/ready/ -v --fail
+fi
--- a/docker-images/kafka-based/kafka/scripts/kafka_readiness_kraft.sh
+++ b/docker-images/kafka-based/kafka/scripts/kafka_readiness_kraft.sh
@@ -9,6 +9,6 @@ if [[ "$roles" =~ "controller" ]] && [[ ! "$roles" =~ "broker" ]]; then
   netstat -lnt | grep -E 'tcp?[[:space:]]+[0-9]+[[:space:]]+[0-9]+[[:space:]]+[^ ]+:9090+[[:space:]]+[^ ]+[[:space:]]+LISTEN*'
 else
   # For combined or broker only mode, check readiness via HTTP endpoint exposed by Kafka Agent.
-  # The endpoint returns 200 when broker state is 3 (RUNNING).
+  # The endpoint returns 204 when broker state is 3 (RUNNING).
   curl http://localhost:8080/v1/ready/ --fail
 fi
--- a/docker-images/kafka-based/kafka/scripts/kafka_readiness_kraft.sh
+++ b/docker-images/kafka-based/kafka/scripts/kafka_readiness_kraft.sh
@@ -10,5 +10,5 @@ if [[ "$roles" =~ "controller" ]] && [[ ! "$roles" =~ "broker" ]]; then
 else
   # For combined or broker only mode, check readiness via HTTP endpoint exposed by Kafka Agent.
   # The endpoint returns 200 when broker state is 3 (RUNNING).
-  curl http://localhost:8080/v1/ready/ -v --fail
+  curl http://localhost:8080/v1/ready/ --fail
 fi
--- a/kafka-agent/src/main/java/io/strimzi/kafka/agent/KafkaAgent.java
+++ b/kafka-agent/src/main/java/io/strimzi/kafka/agent/KafkaAgent.java
@@ -21,6 +21,7 @@ import org.eclipse.jetty.server.ServerConnector;
 import org.eclipse.jetty.server.SslConnectionFactory;
 import org.eclipse.jetty.server.handler.AbstractHandler;
 import org.eclipse.jetty.server.handler.ContextHandler;
+import org.eclipse.jetty.server.handler.ContextHandlerCollection;
 import org.eclipse.jetty.util.ssl.SslContextFactory;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -58,7 +59,7 @@ import java.util.Map;
 public class KafkaAgent {
     private static final Logger LOGGER = LoggerFactory.getLogger(KafkaAgent.class);
     private static final String BROKER_STATE_PATH = "/v1/broker-state";
-    private static final String READINESS_ENDPOINT = "/v1/ready";
+    private static final String READINESS_ENDPOINT_PATH = "/v1/ready";
     private static final int HTTPS_PORT = 8443;
     private static final int HTTP_PORT = 8080;
     private static final long GRACEFUL_SHUTDOWN_TIMEOUT_MS = 30 * 1000;
@@ -121,14 +122,7 @@ public class KafkaAgent {
         pollerThread.setDaemon(true);
 
         try {
-            startReadinessServer();
-        } catch (Exception e) {
-            LOGGER.error("Could not start the server for readiness endpoint: ", e);
-            throw new RuntimeException(e);
-        }
-
-        try {
-            startBrokerStateServer();
+            startHTTPServer();
         } catch (Exception e) {
             LOGGER.error("Could not start the server for broker state: ", e);
             throw new RuntimeException(e);
@@ -222,22 +216,28 @@ public class KafkaAgent {
                 && "SessionExpireListener".equals(name.getType());
     }
 
-    private void startBrokerStateServer() throws Exception {
+    private void startHTTPServer() throws Exception {
         Server server = new Server();
 
         HttpConfiguration https = new HttpConfiguration();
         https.addCustomizer(new SecureRequestCustomizer());
-
-        ServerConnector conn = new ServerConnector(server,
+        ServerConnector httpsConn = new ServerConnector(server,
                 new SslConnectionFactory(getSSLContextFactory(), "http/1.1"),
                 new HttpConnectionFactory(https));
-        conn.setPort(HTTPS_PORT);
+        httpsConn.setPort(HTTPS_PORT);
 
-        ContextHandler context = new ContextHandler(BROKER_STATE_PATH);
-        context.setHandler(getBrokerStateHandler());
+        ContextHandler brokerStateContext = new ContextHandler(BROKER_STATE_PATH);
+        brokerStateContext.setHandler(getBrokerStateHandler());
 
-        server.setConnectors(new Connector[]{conn});
-        server.setHandler(context);
+        ServerConnector httpConn  = new ServerConnector(server);
+        httpConn.setPort(HTTP_PORT);
+
+        ContextHandler readinessContext = new ContextHandler(READINESS_ENDPOINT_PATH);
+        readinessContext.setHandler(getReadinessHandler());
+
+        server.setConnectors(new Connector[] {httpsConn, httpConn});
+        server.setHandler(new ContextHandlerCollection(brokerStateContext, readinessContext));
+
         server.setStopTimeout(GRACEFUL_SHUTDOWN_TIMEOUT_MS);
         server.setStopAtShutdown(true);
         server.start();
@@ -293,19 +293,6 @@ public class KafkaAgent {
         return  sslContextFactory;
     }
 
-    private void startReadinessServer() throws Exception {
-        Server server = new Server();
-        ServerConnector conn = new ServerConnector(server);
-        conn.setPort(HTTP_PORT);
-        server.setConnectors(new Connector[] {conn});
-        ContextHandler readinessContext = new ContextHandler(READINESS_ENDPOINT);
-        readinessContext.setHandler(getReadinessHandler());
-        server.setHandler(readinessContext);
-        server.setStopTimeout(GRACEFUL_SHUTDOWN_TIMEOUT_MS);
-        server.setStopAtShutdown(true);
-        server.start();
-    }
-
     /**
      * Creates a Handler instance to handle incoming HTTP requests for readiness check
      *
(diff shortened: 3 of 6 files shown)


docker-images/kafka-based/kafka/scripts/kafka_readiness_kraft.sh
```diff
@@ -0,0 +1,14 @@
+#!/usr/bin/env bash
+set -e
+
+file=/tmp/strimzi.properties
+test -f $file
```
ppatierno, line 5: do we really need to declare a variable to test the file above?

tinaselenge: thank for pointing it out! I meant to use it on line 6, so will fix it.

docker-images/kafka-based/kafka/scripts/kafka_liveness.sh
```diff
@@ -3,7 +3,7 @@ set -e
 
 if [ "$STRIMZI_KRAFT_ENABLED" = "true" ]; then
   # Test KRaft controller process is running
-  . ./kafka_controller_liveness.sh
+  . ./kafka_liveness_kraft.sh
```
katheris, line 6: Since we are now only calling this script from this one place is it worth moving the logic into this file directly, so we don't need to maintain multiple scripts for liveness?

katheris (commented):
The changes generally look good to me, just had one small comment. Also I think this change needs to be blocked behind the work to update the KafkaRoller to check the mode and whether controller nodes are working when it's rolling pods in KRaft mode. Otherwise if the pods are initially unschedulable the KafkaRoller won't be able to create the pods successfully once appropriate Kubernetes nodes are available.

docker-images/kafka-based/kafka/scripts/kafka_liveness.sh
scholzj: +1 to the comment from Kate. Inlining the script might make sense and would make it easier to read and understand what it does.

kafka-agent/src/main/java/io/strimzi/kafka/agent/KafkaAgent.java
```diff
@@ -280,6 +293,49 @@ private SslContextFactory getSSLContextFactory() {
...
+                    } else {
+                        response.setStatus(HttpServletResponse.SC_NOT_ACCEPTABLE);
+                        response.getWriter().print("Readiness failed: brokerState is " + observedState);
+                    }
+                } else {
+                    response.setStatus(HttpServletResponse.SC_NOT_FOUND);
+                    response.getWriter().print("Broker state metric not found");
+                }
+
+            }
```
scholzj, line 335: Should we have some logging here?

scholzj (commented):
* So, what phase of the proposal does this implement? The phase 2 only given it does not change anything in the KafkaRoller? Would be worth clarifying in the description.
* Why is https://issues.apache.org/jira/browse/KAFKA-14658 not a concern for the liveness check on the ZK mode? Is it because it waits for the ZooKeeper connection first? Again, would be good to explain in the description.
* Would it be better to move this logic from the shell scripts to the agent and just in all cases just call some `alive/` / `ready/` endpoints? That would allow better testability of the logic (currently, this will really be tested only in STs and even there in the ideal cases only through a single path I guess).

tinaselenge:
Thank you all for the feedback.

@scholzj I just would like to clarify on your third point
```
Would it be better to move this logic from the shell scripts to the agent and just in all cases just call some alive/ / ready/ endpoints? That would allow better testability of the logic (currently, this will really be tested only in STs and even there in the ideal cases only through a single path I guess).
```
would it make sense to move all of the readiness and liveness logics to the agent which includes checking of controller listening on the port for controller's readiness and checking of java process running in the container for everyone's liveness? My understanding is that the agent runs in the same process as kafka, so doesn't agent running mean there is a java process?

scholzj:
> Thank you all for the feedback.
> 
> @scholzj I just would like to clarify on your third point
> 
> ```
> Would it be better to move this logic from the shell scripts to the agent and just in all cases just call some alive/ / ready/ endpoints? That would allow better testability of the logic (currently, this will really be tested only in STs and even there in the ideal cases only through a single path I guess).
> ```
> 
> would it make sense to move all of the readiness and liveness logics to the agent which includes checking of controller listening on the port for controller's readiness and checking of java process running in the container for everyone's liveness? My understanding is that the agent runs in the same process as kafka, so doesn't agent running mean there is a java process?

I'm not completely sure how would you check if the listener is listening for example - so I guess you would need to figure it out. But the main point is that the agent is as far as I understand part of the same JVM / Java process. So using the HTTP Agent endpoint for example gives you more value than listing the running processes as you would know it is running AND it actually does something and is not just a dead process in the `proc` filesystem. So I guess in general it might be better for all of them. But the question is how would you implement the different parts.

(I'm also fine to leave this for a future PR -> I think we should consider it, but it might be too much additional scope for this PR.)

scholzj:
@katheris I guess that could be an option as well, yeah.

kafka-agent/src/main/java/io/strimzi/kafka/agent/KafkaAgent.java
```diff
@@ -281,6 +293,49 @@ private SslContextFactory getSSLContextFactory() {
...
+        Server server = new Server();
+        ServerConnector conn = new ServerConnector(server);
+        conn.setPort(HTTP_PORT);
+        server.setConnectors(new Connector[] {conn});
+        ContextHandler readinessContext = new ContextHandler(READINESS_ENDPOINT);
+        readinessContext.setHandler(getReadinessHandler());
+        server.setHandler(readinessContext);
+        server.setStopTimeout(GRACEFUL_SHUTDOWN_TIMEOUT_MS);
+        server.setStopAtShutdown(true);
+        server.start();
```
scholzj, line 306: I thought I raised it before, but I cannot see the review comment anywhere. Why do we need a second server here? Why not have one server and use it for everything? I would expect that Jetty should allow to have HTTPS and HTTP ports in one server.

tinaselenge:
> @scholzj Thinking about this PR again, another approach would be to remove the logic to determine what "Ready" means from the Kafka Agent entirely. Given that the agent now exposes an endpoint with the broker state metric the script could instead query that endpoint and do the check for the value itself. That doesn't help with testability of the logic, but does mean that the Kafka Agent moves to having a single role of exposing metrics (once ZooKeeper goes away), rather than a joint role of exposing metrics and determining what "ready" means.
> 
> It also removes the temptation for the KafkaRoller or other parts of Strimzi to use this endpoint and would prevent us running into the problem we have now where other components (specifically the KafkaRoller) assume the readiness works in a certain way.
> 
> I think that was what I assumed we would do when I originally wrote [proposal #46](https://github.com/strimzi/proposals/blob/main/046-kraft-liveness-readiness.md), but the proposal didn't spell that out since at the time the metric endpoint didn't exist

While I agree on KafkaAgent having a single purpose of exposing metrics, I wonder if it's still better to have the logic in the java agent than managing scripts. It is more testable and if we move the readiness check for the controller, then we could remove the scripts and configure liveness/readiness probes to use the endpoints directly (these would likely to be done as future improvements).

However, if you still think it's better to remove the readiness/liveness logic from the agent, instead of having `/alive` http endpoint, liveness check script could still use the broker's endpoint to confirm that kafka process is running without caring about the returned broker state. I would leave this for future improvement as it might be out of scope for the initial implementation. 

@scholzj @scholzj @ppatierno what do you think?

scholzj:
I do not necessarily think it would be better to move it out of the Agent - but I guess it could be discussed. Maybe we should first finish this PR and then get back to it as it seems there is no clear agreement (sorry, I know I started this whole discussion 🤦 ).

katheris:
I think given that the proposal doesn't specify either way I agree with @scholzj that we can go ahead and get this PR merged with the proposed implementation as I do think the added testability is nice

kafka-agent/src/test/java/io/strimzi/kafka/agent/KafkaAgentTest.java
```diff
@@ -106,4 +106,39 @@ public void testBrokerMetricNotFound() throws Exception {
...
+
+        HttpResponse<String> response = HttpClient.newBuilder()
+                .build()
+                .send(req, HttpResponse.BodyHandlers.ofString());
+
+        assertEquals(HttpServletResponse.SC_OK, response.statusCode());
+    }
+
+    @Test
+    public void testReadinessFail() throws Exception {
```
katheris, line 127: Should we include a test for if the brokerState is 127 (unknown)?

docker-images/kafka-based/kafka/scripts/kafka_readiness_kraft.sh
```diff
@@ -0,0 +1,14 @@
...
+roles=$(awk -F "process.roles=" '{print $2}' "$file")
+if [[ "$roles" =~ "controller" ]] && [[ ! "$roles" =~ "broker" ]]; then
+  # For controller only mode, check if it is listening on port 9090 (configured in controller.listener.names).
+  netstat -lnt | grep -E 'tcp?[[:space:]]+[0-9]+[[:space:]]+[0-9]+[[:space:]]+[^ ]+:9090+[[:space:]]+[^ ]+[[:space:]]+LISTEN*'
+else
+  # For combined or broker only mode, check readiness via HTTP endpoint exposed by Kafka Agent.
+  # The endpoint returns 200 when broker state is 3 (RUNNING).
+  curl http://localhost:8080/v1/ready/ -v --fail
```
ppatierno, line 13: why do we need this call to be `-v` (verbose)? For logging purposes?

kafka-agent/src/main/java/io/strimzi/kafka/agent/KafkaAgent.java
```diff
@@ -38,8 +39,8 @@
  * A very simple Java agent which polls the value of the {@code kafka.server:type=KafkaServer,name=BrokerState}
  * Yammer Metric and once it reaches the value 3 (meaning "running as broker", see {@code kafka.server.BrokerState}),
  * creates a given file.
- * The presence of this file is tested via a Kube "exec" readiness probe to determine when the broker is ready.
- * It also exposes a REST endpoint for broker metrics.
+ * In zookeeper mode, the presence of this file is tested via a Kube "exec" readiness probe to determine when the broker is ready.
```
ppatierno, line 42:
```suggestion
 * In ZooKeeper mode, the presence of this file is tested via a Kube "exec" readiness probe to determine when the broker is ready.
```

kafka-agent/src/main/java/io/strimzi/kafka/agent/KafkaAgent.java
```diff
@@ -116,7 +122,7 @@ private void run() {
         pollerThread.setDaemon(true);
 
         try {
-            startHttpServer();
+            startHTTPServer();
```
ppatierno, line 125: I would have left as it was :-D

kafka-agent/src/main/java/io/strimzi/kafka/agent/KafkaAgent.java
```diff
@@ -281,6 +293,38 @@ private SslContextFactory getSSLContextFactory() {
...
+                baseRequest.setHandled(true);
+                if (brokerState != null) {
+                    byte observedState = (byte) brokerState.value();
+                    boolean stateIsRunning = BROKER_RUNNING_STATE <= observedState && BROKER_UNKNOWN_STATE != observedState;
+                    if (stateIsRunning) {
+                        LOGGER.trace("Broker is in running according to {}. The current state is {}", brokerStateName, observedState);
+                        response.setStatus(HttpServletResponse.SC_OK);
+                    } else {
+                        LOGGER.trace("Broker is not running according to {}. The current state is {}", brokerStateName, observedState);
+                        response.setStatus(HttpServletResponse.SC_NOT_ACCEPTABLE);
```
ppatierno, line 316: as per `406 NOT ACCEPTABLE` HTTP status code (https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/406) I am not sure it's the right use case here.
We are checking a "service" (the broker) which is kind of not available in the state we would like to have ... so maybe `503 SERVICE UNAVAILABLE` is better?
The 4xx status code are client error but here the HTTP client is just asking for the state of the broker, it didn't make any mistake or sent a wrong request. There is nothing the client can do to fix the problem.
It's looks like more a 5xx server error to me.

kafka-agent/src/main/java/io/strimzi/kafka/agent/KafkaAgent.java
```diff
@@ -281,6 +293,38 @@ private SslContextFactory getSSLContextFactory() {
...
+            public void handle(String s, Request baseRequest, HttpServletRequest request, HttpServletResponse response) throws IOException {
+                response.setContentType("application/json");
+                response.setCharacterEncoding("UTF-8");
+                baseRequest.setHandled(true);
+                if (brokerState != null) {
+                    byte observedState = (byte) brokerState.value();
+                    boolean stateIsRunning = BROKER_RUNNING_STATE <= observedState && BROKER_UNKNOWN_STATE != observedState;
+                    if (stateIsRunning) {
+                        LOGGER.trace("Broker is in running according to {}. The current state is {}", brokerStateName, observedState);
+                        response.setStatus(HttpServletResponse.SC_OK);
```
ppatierno, line 313: We are not bringing a body in the HTTP response, I would use `204 NO CONTENT` instead of OK (i.e. this is what we have in our Strimzi HTTP bridge on the readiness/liveness probes).

tinaselenge:
@scholzj testKafkaPodPendingDueToRack seems to pass when running locally. Would you mind running it again?

ppatierno (approved):
LGTM (apart the test failure to be investigated).

docker-images/kafka-based/kafka/scripts/kafka_readiness_kraft.sh
```diff
@@ -0,0 +1,14 @@
...
+
+file=/tmp/strimzi.properties
+test -f $file
+roles=$(awk -F "process.roles=" '{print $2}' "$file")
+if [[ "$roles" =~ "controller" ]] && [[ ! "$roles" =~ "broker" ]]; then
+  # For controller only mode, check if it is listening on port 9090 (configured in controller.listener.names).
+  netstat -lnt | grep -E 'tcp?[[:space:]]+[0-9]+[[:space:]]+[0-9]+[[:space:]]+[^ ]+:9090+[[:space:]]+[^ ]+[[:space:]]+LISTEN*'
+else
+  # For combined or broker only mode, check readiness via HTTP endpoint exposed by Kafka Agent.
+  # The endpoint returns 200 when broker state is 3 (RUNNING).
```
katheris, line 12:
```suggestion
  # The endpoint returns 204 when broker state is 3 (RUNNING).
```

kafka-agent/src/main/java/io/strimzi/kafka/agent/KafkaAgent.java
```diff
@@ -50,12 +51,17 @@
  *          "remainingSegmentsToRecover": 456
  *        }
  *      }</dd>
+ *     <dt>{@code GET /v1/ready}</dt>
+ *     <dd>Returns HTTP code 200 if broker state is RUNNING(3). Otherwise returns non successful HTTP code.
```
katheris, line 55:
```suggestion
 *     <dd>Returns HTTP code 204 if broker state is RUNNING(3). Otherwise returns non successful HTTP code.
```

katheris (approved):
Couple of minor suggestions but other than that looks good to me

tinaselenge:
The same test failed despite passing locally. I think it might be to do with the change in readiness check. 
```
2023-08-10 12:36:54 [main] INFO  [ResourceManager:545] Waiting for Kafka: namespace-16/my-cluster-4c9f7234 will have desired state: Ready

2023-08-10 12:50:54 [main] INFO  [ResourceManager:519] Kafka status:

Conditions:
	Type: NotReady
	Message: Pod is unschedulable or is not starting

Pods with conditions and messages:

my-cluster-4c9f7234-kafka-pool-d5c5a78d-0:
	Type: PodScheduled
	Message: 0/1 nodes are available: 1 node(s) didn't match Pod's node affinity/selector.


my-cluster-4c9f7234-kafka-pool-d5c5a78d-1:
	Type: PodScheduled
	Message: 0/1 nodes are available: 1 node(s) didn't match Pod's node affinity/selector.


my-cluster-4c9f7234-kafka-pool-d5c5a78d-2:
	Type: <EMPTY>
	Message: <EMPTY>

	Type: Ready
	Message: containers with unready status: [kafka]

	Type: ContainersReady
	Message: containers with unready status: [kafka]
```
I'm investigating it.

tinaselenge:
The system test KafkaRollerST#testKafkaPodPendingDueToRack in KRaft mode is failing due to timing out while waiting for the brokers in combined mode to become ready. The scenario matches what was described in the proposal [here](https://github.com/strimzi/proposals/blob/main/046-kraft-liveness-readiness.md#impact-on-the-kafkaroller), where pods are started with affinity constraints therefore put into a pending state by Kubernetes. The test then removes the constraint and waits for the combined pods to become ready. The first pod that got scheduled waits for the controller quorum to form before going into RUNNING state, however quorum cannot be formed as the other pods are still in pending state. 

This is what @katheris mentioned above, that this PR is blocked behind KafkaRoller's update to check KRaft controllers which is partly being implemented via this [proposal in progress](https://github.com/strimzi/proposals/pull/82).

scholzj:
Being blocked by a proposal sounds like something what will not be addressed in a few days. Should we close this PR for the time being or move it to Draft?

showuon:
Sorry for the delay (and thanks for the kindly reminder @tinaselenge !). Will check it tomorrow.

tinaselenge:
@scholzj, @ppatierno can you please trigger the system tests again?

scholzj (approved):

ppatierno (approved):
LGTM. Thanks!

PR #8895, opened 2023-07-24
Update Fabric8 Kubernetes Client to 6.8.0

scholzj:
### Type of change

- Task

### Description

This PR updates the Fabric8 Kubernetes Client to 6.8.0. It also updates Sundrio to keep it in-sync with the version used by Fabric8 and updates the code to match various changes caused by the new Fabric8 version. 

It also disables the check for `BC_UNCONFIRMED_CAST_OF_RETURN_VALUE` in Spotbugs as it seems to generate many false positives through the classes where Fabric8 builder / Fluent classes are being used.

### Checklist


ppatierno:
out of curiosity, is the `@SuppressWarnings("deprecation")` that you added for Mirror Maker and Jaeger related to the fabric8 update as described by the PR? Why not having a different one?

scholzj:
> out of curiosity, is the @SuppressWarnings("deprecation") that you added for Mirror Maker and Jaeger related to the fabric8 update as described by the PR? Why not having a different one?

They were no needed before because the generated code worked somehow differently. Not sure exactly why.

ppatierno (approved):

PR #8897, opened 2023-07-25
[ST] Add ST for KafkaNodePool ID management

jankalinic:
### Type of change

- Enhancement

### Description

This PR adds a basic ST that sets annotation to the deployed KafkaNodePool resource. Then scales it's replicas up and down and checks that correct brokers pods with IDs specified in the annotation are created or deleted. This PR also adds some KafkaNodePoolUtils for ST usage.

### Checklist

- [ ] Write tests

### DIFF
--- a/systemtest/src/main/java/io/strimzi/systemtest/resources/crd/KafkaNodePoolResource.java
+++ b/systemtest/src/main/java/io/strimzi/systemtest/resources/crd/KafkaNodePoolResource.java
@@ -87,7 +87,7 @@ public class KafkaNodePoolResource implements ResourceType<KafkaNodePool> {
 
         String nodePoolName = Constants.KAFKA_NODE_POOL_PREFIX + hashStub(resource.getMetadata().getName());
 
-        KafkaNodePoolBuilder builder = KafkaNodePoolTemplates.defaultKafkaNodePool(nodePoolName, resource.getMetadata().getName(), resource.getSpec().getKafka().getReplicas())
+        KafkaNodePoolBuilder builder = KafkaNodePoolTemplates.defaultKafkaNodePool(resource.getMetadata().getNamespace(), nodePoolName, resource.getMetadata().getName(), resource.getSpec().getKafka().getReplicas())
             .editOrNewMetadata()
                 .withNamespace(resource.getMetadata().getNamespace())
                 .addToLabels(resource.getMetadata().getLabels())
--- a/systemtest/src/main/java/io/strimzi/systemtest/resources/crd/KafkaNodePoolResource.java
+++ b/systemtest/src/main/java/io/strimzi/systemtest/resources/crd/KafkaNodePoolResource.java
@@ -48,6 +48,7 @@ public class KafkaNodePoolResource implements ResourceType<KafkaNodePool> {
     public void delete(KafkaNodePool resource) {
         String namespaceName = resource.getMetadata().getNamespace();
         String clusterName = resource.getMetadata().getLabels().get(Labels.STRIMZI_CLUSTER_LABEL);
+        String podsetName = String.join("-", clusterName, resource.getMetadata().getName());
 
         kafkaNodePoolClient().inNamespace(resource.getMetadata().getNamespace()).resource(resource).delete();
 
@@ -55,7 +56,7 @@ public class KafkaNodePoolResource implements ResourceType<KafkaNodePool> {
         // the Kafka pods will not be deleted during Kafka resource deletion -> so if we would delete PVCs during Kafka deletion
         // they will not be deleted too
         // additional deletion of pvcs with specification deleteClaim set to false which were not deleted prior this method
-        PersistentVolumeClaimUtils.deletePvcsByPrefixWithWait(namespaceName, clusterName);
+        PersistentVolumeClaimUtils.deletePvcsByPrefixWithWait(namespaceName, podsetName);
     }
 
     @Override
--- a/systemtest/src/main/java/io/strimzi/systemtest/templates/crd/KafkaNodePoolTemplates.java
+++ b/systemtest/src/main/java/io/strimzi/systemtest/templates/crd/KafkaNodePoolTemplates.java
@@ -14,9 +14,10 @@ public class KafkaNodePoolTemplates {
 
     private KafkaNodePoolTemplates() {}
 
-    public static KafkaNodePoolBuilder defaultKafkaNodePool(String nodePoolName, String kafkaClusterName, int kafkaReplicas) {
+    public static KafkaNodePoolBuilder defaultKafkaNodePool(String namespaceName, String nodePoolName, String kafkaClusterName, int kafkaReplicas) {
         return new KafkaNodePoolBuilder()
             .withNewMetadata()
+                .withNamespace(namespaceName)
                 .withName(nodePoolName)
                 .withLabels(Map.of(Labels.STRIMZI_CLUSTER_LABEL, kafkaClusterName))
             .endMetadata()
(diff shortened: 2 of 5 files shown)


systemtest/src/main/java/io/strimzi/systemtest/utils/kafkaUtils/KafkaNodePoolUtils.java
```diff
@@ -0,0 +1,85 @@
...
+    public static String annotateKafkaNodePoolRemoveNodeIds(String namespaceName, String resourceName, List<Integer> annotationValue) {
+        return annotateKafkaNodePoolRemoveNodeIds(namespaceName, resourceName, annotationValue.toString());
+    }
+
+    public static void scaleKafkaNodePool(String namespaceName, String kafkaNodePoolName, int scaleToReplicas) {
+        LOGGER.info("Scaling KafkaNodePool: {}/{} to {} replicas", namespaceName, kafkaNodePoolName, scaleToReplicas);
+        KafkaNodePoolResource.replaceKafkaNodePoolResourceInSpecificNamespace(kafkaNodePoolName, nodePool -> {
+            nodePool.getSpec().setReplicas(scaleToReplicas);
+        }, namespaceName);
+    }
```
im-konge, line 62: This is just maybe my feeling, but I would not encapsulate it in some method like this.
I would normally use it directly in the test, so we know what needs to be changed to "scale the NodePool"

im-konge (approved):
I have just one comment, otherwise LGTM

systemtest/src/test/java/io/strimzi/systemtest/kafka/KafkaNodePoolST.java
```diff
@@ -203,6 +205,61 @@ void testKafkaManagementTransferToAndFromKafkaNodePool(ExtensionContext extensio
...
+            .endSpec()
+            .build();
+
+        resourceManager.createResourceWithWait(extensionContext,
+            kafkaNodePool,
+            kafka);
+
+        // Save for later comparison
+        List<Integer> originalNodePoolIds = KafkaNodePoolUtils.getCurrentKafkaNodePoolIds(testStorage.getNamespaceName(), testStorage.getKafkaNodePoolName());
+        List<Integer> expectedNodePoolIds = KafkaNodePoolUtils.parseNodePoolIdsWithRangeStringToIntegerList(nodePoolAnnoIdsWithRange);
```
scholzj, line 240: Why don't you hardcode this instead of having such a complicated method? You already hardcoded the range from which the node IDs will be picked. So I do not think it would matter if you hardcode this as well.

scholzj (commented):
I think the test itself looks mostly good. But I wonder if you have some bigger plan for testing this:
* I think there are many things to test both around this aspect but also how it interacts with multiple pools, that the scaling works beyond just assigning node IDs, how the initial node IDs are assigned, how it interacts when you scale one pool up and other down etc. I you write every single test individually like that, you will have way too many tests which will take ages to execute.
* For a single test, this adds hell of a lot various utility methods to the Utils class. It is not clear if they will be really used elsewhere or not. If not, it is probably not worth it.

So while the code itself looks good to me, it is hard to say how well it fits the overall perspective.

systemtest/src/test/java/io/strimzi/systemtest/kafka/KafkaNodePoolST.java
jankalinic: Yes, i will change that, thank you

systemtest/src/main/java/io/strimzi/systemtest/utils/kafkaUtils/KafkaNodePoolUtils.java
jankalinic: used scale() instead

systemtest/src/main/java/io/strimzi/systemtest/utils/kafkaUtils/KafkaNodePoolUtils.java
```diff
@@ -0,0 +1,40 @@
...
+public class KafkaNodePoolUtils {
+
+    private KafkaNodePoolUtils() {
+    }
+
+    private static final Logger LOGGER = LogManager.getLogger(PodUtils.class);
+
+    public static KafkaNodePool getKafkaNodePool(String namespaceName, String resourceName) {
+        return KafkaNodePoolResource.kafkaNodePoolClient().inNamespace(namespaceName).withName(resourceName).get();
+    }
```
im-konge, line 24: Isn't similar method inside the `KafkaNodePoolResource`?

systemtest/src/test/java/io/strimzi/systemtest/kafka/KafkaNodePoolST.java
```diff
@@ -203,6 +207,98 @@ void testKafkaManagementTransferToAndFromKafkaNodePool(ExtensionContext extensio
...
+        PodUtils.waitUntilPodStabilityReplicasCount(testStorage.getNamespaceName(), KafkaResource.getStrimziPodSetName(testStorage.getClusterName(), nodePoolNameA), 2);
+        PodUtils.waitUntilPodStabilityReplicasCount(testStorage.getNamespaceName(), KafkaResource.getStrimziPodSetName(testStorage.getClusterName(), nodePoolNameB), 6);
+        // Verify NodePool IDs
+        // NodePool B should contain -> [0, 2, 3, 4, 6, 7] as [1, 5] is already taken by NodePool A
+        assertThat(KafkaNodePoolUtils.getCurrentKafkaNodePoolIds(testStorage.getNamespaceName(), nodePoolNameA).equals(Arrays.asList(1, 5)));
+        assertThat(KafkaNodePoolUtils.getCurrentKafkaNodePoolIds(testStorage.getNamespaceName(), nodePoolNameB).equals(Arrays.asList(0, 2, 3, 4, 6, 7)));
+    }
+
+    private void assertThat(boolean equals) {
+    }
```
im-konge, line 300: Why do you have this here?

systemtest/src/main/java/io/strimzi/systemtest/resources/crd/KafkaNodePoolResource.java
```diff
@@ -48,14 +48,15 @@ public void create(KafkaNodePool resource) {
...
+        String podsetName = String.join("-", clusterName, resource.getMetadata().getName());
 
         kafkaNodePoolClient().inNamespace(resource.getMetadata().getNamespace()).resource(resource).delete();
 
         // once KafkaNodePool is deleted, we should delete PVCs and wait for the deletion to be completed
         // the Kafka pods will not be deleted during Kafka resource deletion -> so if we would delete PVCs during Kafka deletion
         // they will not be deleted too
         // additional deletion of pvcs with specification deleteClaim set to false which were not deleted prior this method
-        PersistentVolumeClaimUtils.deletePvcsByPrefixWithWait(namespaceName, clusterName);
+        PersistentVolumeClaimUtils.deletePvcsByPrefixWithWait(namespaceName, podsetName);
```
jankalinic, line 59: This is just fix that is needed in order to clear all resources after the test finishes

henryZrncik (approved):

systemtest/src/main/java/io/strimzi/systemtest/templates/crd/KafkaNodePoolTemplates.java
```diff
@@ -23,4 +24,17 @@ public static KafkaNodePoolBuilder defaultKafkaNodePool(String nodePoolName, Str
...
+        return new KafkaNodePoolBuilder()
+            .withNewMetadata()
+                .withNamespace(namespaceName)
+                .withName(nodePoolName)
+                .withLabels(Map.of(Labels.STRIMZI_CLUSTER_LABEL, kafkaClusterName))
+            .endMetadata()
+            .withNewSpec()
+                .addToRoles(ProcessRoles.BROKER)
+                .withReplicas(kafkaReplicas)
+            .endSpec();
```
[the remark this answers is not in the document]
jankalinic: Sounds better, true

scholzj (approved):

im-konge (approved):

Frawless (approved):

PR #8901, opened 2023-07-25
JaasConfig refactoring code

ShazaAldawamneh:
### Type of change

_Select the type of your PR_

- Enhancement / new feature
- Refactoring

### Description

[Enhancement] Refactor JAAS config generation code into it's own class.

### Checklist

- [ y] Write tests
- [ y] Make sure all tests pass
- [ y] Update documentation
- [ y] Check RBAC rights for Kubernetes / OpenShift roles
- [ y] Try your changes from Pod inside your Kubernetes and OpenShift cluster, not just locally


cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/JaasConfig.java
```diff
@@ -0,0 +1,49 @@
...
+
+import java.util.Map;
+import java.util.Objects;
+import java.util.StringJoiner;
+
+/**
+ * JaasConfig is a Java class that provides a utility method for generating Java Authentication and
+ * Authorization Service (JAAS) configuration strings.
+ */
+public class JaasConfig {
```
tombentley, line 15: Is a new class really necessary, or could we put this code in `AuthenticationUtils`?

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/JaasConfig.java
```diff
@@ -0,0 +1,49 @@
...
+     * @param moduleName The name of the JAAS module to be configured.
+     * @param options A Map containing the options to be set for the JAAS module.
+     *               The options are represented as key-value pairs, where both the key and
+     *               the value must be non-null String objects.
+     * @return A String representing the JAAS configuration.
+     * @throws IllegalArgumentException If the moduleName is empty, or it contains '=' or ';',
+     *                                  or if any key or value in the options map is empty,
+     *                                  or they contain '=' or ';'.
+     */
+    public static String config(String moduleName, Map<String, String> options) {
```
tombentley, line 28: If we did move this to `AuthenticationUtils` it's name would need to change to be more JAAS specific.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/JaasConfig.java
```diff
@@ -0,0 +1,49 @@
...
+                throw new IllegalArgumentException("Keys and values must not contain '=' or ';'");
+            }
+            if (moduleName.contains("=") || moduleName.contains(";") || moduleName.isEmpty()) {
+                throw new IllegalArgumentException("module name must be not empty and must not contain '=' or ';'");
+            } else {
+                String s = key + "=\"" + value + "\"";
+                joiner.add(s);
+            }
+        }
+        var stringStream = joiner.toString();
```
tombentley, line 45: We can inline this.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/JaasConfig.java
```diff
@@ -0,0 +1,49 @@
...
+            String value = entry.getValue();
+            Objects.requireNonNull(key);
+            Objects.requireNonNull(value);
+            if (key.contains("=") || key.contains(";") || value.contains("=") || value.contains(";")) {
+                throw new IllegalArgumentException("Keys and values must not contain '=' or ';'");
+            }
+            if (moduleName.contains("=") || moduleName.contains(";") || moduleName.isEmpty()) {
+                throw new IllegalArgumentException("module name must be not empty and must not contain '=' or ';'");
+            } else {
+                String s = key + "=\"" + value + "\"";
```
tombentley, line 41: We can inline this.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/JaasConfig.java
```diff
@@ -0,0 +1,49 @@
...
+     *                                  or if any key or value in the options map is empty,
+     *                                  or they contain '=' or ';'.
+     */
+    public static String config(String moduleName, Map<String, String> options) {
+        StringJoiner joiner = new StringJoiner(" ");
+        for (Map.Entry<String, String> entry : options.entrySet()) {
+            String key = entry.getKey();
+            String value = entry.getValue();
+            Objects.requireNonNull(key);
+            Objects.requireNonNull(value);
```
tombentley, line 34: `Objects.requireNonNull` returns its argument, so you could write this as:

```suggestion
            String key = Objects.requireNonNull(entry.getKey());
            String value = Objects.requireNonNull(entry.getValue());
```

cluster-operator/src/test/java/io/strimzi/operator/cluster/model/KafkaBrokerConfigurationBuilderTest.java
```diff
@@ -49,9 +49,7 @@
 import static java.util.Arrays.asList;
 import static java.util.Collections.emptyList;
 import static java.util.Collections.singletonList;
-import static org.hamcrest.CoreMatchers.is;
 import static org.hamcrest.MatcherAssert.assertThat;
-import static org.hamcrest.Matchers.equalTo;
 
 @ParallelSuite
 public class KafkaBrokerConfigurationBuilderTest {
```
tombentley, line 55: I'm guessing from all the whitespace/indentation changes in this file that your IDE reformatted it. While some of those changes are OK there are places where it makes the code less readable. Typically when using next builders, for example. So it would be good to revert at least those changes.

cluster-operator/src/test/java/io/strimzi/operator/cluster/model/KafkaBrokerConfigurationBuilderTest.java
```diff
@@ -846,11 +844,11 @@ public void testWithTlsListenersWithCustomCerts()  {
...
-                    .withNewBrokerCertChainAndKey()
-                        .withSecretName("my-secret")
-                        .withKey("my.key")
-                        .withCertificate("my.crt")
-                    .endBrokerCertChainAndKey()
+                .withNewBrokerCertChainAndKey()
+                .withSecretName("my-secret")
+                .withKey("my.key")
+                .withCertificate("my.crt")
+                .endBrokerCertChainAndKey()
```
tombentley, line 851: This is an example of "less readable". We're building a tree of objects here and were using the indentation to reflect the tree structure. Can we revert this one and similar cases?

cluster-operator/src/test/java/io/strimzi/operator/cluster/operator/assembly/JaasConfigTest.java
```diff
@@ -0,0 +1,128 @@
+/*
+ * Copyright Strimzi authors.
+ * License: Apache License 2.0 (see the file LICENSE or http://apache.org/licenses/LICENSE-2.0.html).
+ */
+package io.strimzi.operator.cluster.operator.assembly;
+
+import org.junit.jupiter.api.Test;
+
+//import javax.security.auth.login.Configuration;
+//import java.lang.reflect.Constructor;
```
tombentley, line 10: We try to avoid adding commented out code.

cluster-operator/src/test/java/io/strimzi/operator/cluster/operator/assembly/JaasConfigTest.java
```diff
@@ -0,0 +1,128 @@
...
+    }
+//    public static void main(String[] a) throws Exception {
+//        //org.apache.kafka.common.security.JaasConfig jaasConfig = new org.apache.kafka.common.security.JaasConfig();
+//        Class<?> aClass = Class.forName("org.apache.kafka.common.security.JaasConfig");
+//
+//        Constructor<?> declaredConstructor = aClass.getDeclaredConstructor(String.class, String.class);
+//        declaredConstructor.setAccessible(true);
+//        Configuration o = (Configuration) declaredConstructor.newInstance("==", "foo required foo=\"value\"with\\\"quote\";");
+////        AppConfigurationEntry[] appConfigurationEntry = o.getAppConfigurationEntry();
+//    }
```
tombentley, line 127: Again we usually avoid adding commented code. I think case _I_ understand that we used this to prove/understand what Kafka does and doesn't accept while writing the test. I personally I think there's a case for keeping it, assuming you added a comment explaining what it was for. But I suspect other reviewers might disagree.

tombentley (commented):
Thanks! This looks like a reasonable improvement which centralises how we generate JAAS configs and adds some useful tests. I left a few comments.

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaBrokerConfigurationBuilder.java
```diff
@@ -409,34 +412,40 @@ private void configureAuthentication(String listenerName, List<String> securityP
         if (auth instanceof KafkaListenerAuthenticationOAuth oauth) {
             securityProtocol.add(String.format("%s:%s", listenerName, getSecurityProtocol(tls, true)));
 
-            List<String> options = new ArrayList<>(getOAuthOptions(oauth));
-            options.add("oauth.config.id=\"" + listenerName + "\"");
+            Map<String, String> jaasOptions = new LinkedHashMap<>(getOAuthOptions(oauth));
```
see-quick, line 415: Just curious, why do you use `LinkedHashMap` instead of HashMap? :) I do not see anything related to the order so using HashMap should be sufficient or am I missing something?.

ShazaAldawamneh: because the tests in the KafkaBrokerConfigurationBuilderTest.java is failing when we have HashMap instead of LinkedHashMap and that is because the order matter in this situation

cluster-operator/src/test/java/io/strimzi/operator/cluster/model/AuthenticationUtilsTest.java
```diff
@@ -0,0 +1,129 @@
...
+ * Copyright Strimzi authors.
+ * License: Apache License 2.0 (see the file LICENSE or http://apache.org/licenses/LICENSE-2.0.html).
+ */
+package io.strimzi.operator.cluster.model;
+
+import io.strimzi.operator.cluster.model.AuthenticationUtils;
+import org.junit.jupiter.api.Test;
+
+//import javax.security.auth.login.Configuration;
+//import java.lang.reflect.Constructor;
```
ppatierno, line 11: I guess these can be removed :-)

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaMirrorMaker2AssemblyOperator.java
```diff
@@ -465,7 +465,7 @@ private static void prepareMirrorMaker2ConnectorConfig(Reconciliation reconcilia
         config.putAll(mirror.getAdditionalProperties());
     }
 
-    private static void addClusterToMirrorMaker2ConnectorConfig(Map<String, Object> config, KafkaMirrorMaker2ClusterSpec cluster, String configPrefix) {
+    static void addClusterToMirrorMaker2ConnectorConfig(Map<String, Object> config, KafkaMirrorMaker2ClusterSpec cluster, String configPrefix) {
```
ppatierno, line 468:
```suggestion
    /* test */ static void addClusterToMirrorMaker2ConnectorConfig(Map<String, Object> config, KafkaMirrorMaker2ClusterSpec cluster, String configPrefix) {
```

I see it was changed from `private` to `package` (implicit) to be used in tests. In this case we have the convention to add a comment as I suggested here, in order to make clear why it's not `private`.

ppatierno (approved):
LGTM. Thanks for the PR!

tombentley (approved):
The code changes LGTM. Thanks!

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaBrokerConfigurationBuilder.java
```diff
@@ -563,20 +571,28 @@ private String getSecurityProtocol(boolean tls, boolean sasl)   {
         return options;
     }
 
-    static void addOption(List<String> options, String option, String value) {
-        if (value != null) options.add(String.format("%s=\"%s\"", option, value));
+    static void addOption(Map<String, String> options, String option, String value) {
```
scholzj, line 574: I know you did not changed this. But should it be renamed to `addOptionIfNotNull`?

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaBrokerConfigurationBuilder.java
```diff
@@ -563,20 +571,28 @@ private String getSecurityProtocol(boolean tls, boolean sasl)   {
...
 
-    static void addBooleanOptionIfFalse(List<String> options, String option, boolean value) {
-        if (!value) options.add(String.format("%s=\"%s\"", option, false));
+    static void addBooleanOptionIfFalse(Map<String, String> options, String option, boolean value) {
+        if (!value) {
+            options.put(option, String.valueOf(false));
+        }
     }
 
     static void addOption(PrintWriter writer, String name, Object value) {
```
scholzj, line 592: Same as above ... should it be renamed to `addOptionIfNotNull`?

scholzj (commented):
Two more nits given you probably will need to modify it to fix the tests. LGTM otherwise - apart from the bug somewhere in code failing the STs.

cluster-operator/src/test/java/io/strimzi/operator/cluster/model/KafkaBrokerConfigurationBuilderTest.java
ShazaAldawamneh: thanks Tom, yes that make sense

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaMirrorMaker2AssemblyOperator.java
ShazaAldawamneh: I resolved this in the latest commit

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaBrokerConfigurationBuilder.java
ShazaAldawamneh: I'm wondering why would we do that? as this is not a test method and as not all the methods we have that contain null checker are called that way ? so why would we call this method that way ? if you can clear that to me ?

scholzj: Well, it has the null check. So it adds the option only if it is not null. That is important information as it for example means you do not need to have a null check on the outside of calling this method. So it would be good ot have it clear from the name.

ShazaAldawamneh: I see, thanks for the clarification

fvaleri:
There are a couple of issues on the OAuth JAAS config build:

- In `KafkaMirrorMaker2AssemblyOperator.oauthJaasConfig` the oauthConfig String may and usually contain multiple options, not just one
- We add double quotes for JAAS options in `AuthentictionUtils.addOption`, and then again in `AuthentictionUtils.jaasConfig`

This is how a correct connector's configuration looks like from the main branch:

```sh
offset-syncs.topic.replication.factor=-1,
replication.factor=-1,
sync.topic.acls.enabled=false,
target.cluster.alias=my-cluster-8b74617a-target,
target.cluster.bootstrap.servers=my-cluster-8b74617a-target-kafka-bootstrap:9092,
target.cluster.sasl.login.callback.handler.class=io.strimzi.kafka.oauth.client.JaasClientOauthLoginCallbackHandler,
target.cluster.sasl.mechanism=OAUTHBEARER,
target.cluster.sasl.jaas.config=org.apache.kafka.common.security.oauthbearer.OAuthBearerLoginModule required \
  oauth.client.id="kafka-mirror-maker-2" \
  oauth.password.grant.username="alice" \
  oauth.token.endpoint.uri="http://keycloak-discovery.test-suite-namespace.svc.cluster.local:9595/realms/internal/protocol/openid-connect/token" \
  oauth.connect.timeout.seconds="100" \
  oauth.read.timeout.seconds="100" \
  oauth.client.secret="${file:/tmp/strimzi-mirrormaker2-connector.properties:my-cluster-8b74617a-target.oauth.client.secret}" \
  oauth.password.grant.password="${file:/tmp/strimzi-mirrormaker2-connector.properties:my-cluster-8b74617a-target.oauth.password.grant.password}";,
target.cluster.security.protocol=SASL_PLAINTEXT,
target.cluster.offset.storage.replication.factor=-1,
target.cluster.status.storage.replication.factor=-1,
target.cluster.config.topic.cleanup.policy=compact,
target.cluster.config.storage.replication.factor=-1
```

This is the same connector's configuration created by this branch:

```sh
offset-syncs.topic.replication.factor=-1,
replication.factor=-1,
sync.topic.acls.enabled=false,
target.cluster.alias=my-cluster-758b7d6c-target,
target.cluster.bootstrap.servers=my-cluster-758b7d6c-target-kafka-bootstrap:9092,
target.cluster.sasl.login.callback.handler.class=io.strimzi.kafka.oauth.client.JaasClientOauthLoginCallbackHandler,
target.cluster.sasl.mechanism=OAUTHBEARER,
target.cluster.sasl.jaas.config=org.apache.kafka.common.security.oauthbearer.OAuthBearerLoginModule required \
  oauth.client.secret="${file:/tmp/strimzi-mirrormaker2-connector.properties:my-cluster-758b7d6c-target.oauth.client.secret}" \
  oauth.password.grant.password="${file:/tmp/strimzi-mirrormaker2-connector.properties:my-cluster-758b7d6c-target.oauth.password.grant.password}" \
  oauth.client.id=""kafka-mirror-maker-2" \
  oauth.password.grant.username="alice" \
  oauth.token.endpoint.uri="http://keycloak-discovery.infra-namespace.svc.cluster.local:9595/realms/internal/protocol/openid-connect/token" \
  oauth.connect.timeout.seconds="100" \
  oauth.read.timeout.seconds="100"";
```

The last commit should fix that.

fvaleri (commented):
You can run OauthPasswordGrantsST#testPasswordGrantsKafkaMirrorMaker2 on your local env to confirm it fixes the issue, but there are some conflicts to fix before running the regression test suite.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaMirrorMaker2AssemblyOperator.java
```diff
@@ -532,12 +533,15 @@ private static void prepareMirrorMaker2ConnectorConfig(Reconciliation reconcilia
...
-        if (index != -1) {
-            jaasOptions.put(oauthConfig.substring(0, index), oauthConfig.substring(index + 1));
-        }
+        // convert the sasl.jaas.config string value to a map of jaas options
+        Map<String, String> jaasOptions = Arrays.stream(
+            oauthConfig.replace("\"", "").split(" "))
+                .map(option -> option.split("="))
+                .filter(entry -> entry.length == 2)
+                .collect(Collectors.toMap(entry -> entry[0], entry -> entry[1])
+        );
```
tombentley, line 543: This _works_, but it's really ugly. Before this method is called, `authProperties` gets created with the `OAUTH_CONFIG` key added with a carefully constructed JAAS string, and here we end up parsing it back to keys and values. I was looking at whether we should refactor to avoid formatting+parsing, and pass around something more structured than a `String`. It ends up being a bigger refactoring, but leaving the code slightly more clear and direct.

fvaleri: Yes, I agree. It can be improved. I'll leave it to @ShazaAldawamneh.

ShazaAldawamneh: working on it.

ShazaAldawamneh: thanks for your feedback Tom and Fede

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/AuthenticationUtils.java
```diff
@@ -43,7 +46,10 @@ public class AuthenticationUtils {
     private static final String TLS_AUTH_CERT = "TLS_AUTH_CERT";
     private static final String TLS_AUTH_KEY = "TLS_AUTH_KEY";
     private static final String SASL_PASSWORD_FILE = "SASL_PASSWORD_FILE";
-    private static final String OAUTH_CONFIG = "OAUTH_CONFIG";
+    /**
+     * Key for OAuth config
+     */
+    public static final String OAUTH_CONFIG = "OAUTH_CONFIG";
```
scholzj, line 52: Does it really need to be public? I did not find any use of this outside this class.

cluster-operator/src/test/java/io/strimzi/operator/cluster/operator/assembly/KafkaMirrorMaker2AssemblyOperatorTest.java
```diff
@@ -914,4 +918,165 @@ public void testCreateClusterWithJmxEnabled(VertxTestContext context) {
...
+
+        return mirrorMaker2Captor;
+    }
+
+    private KafkaConnectApi createConnectClientMock() {
+        KafkaConnectApi mockConnectClient = mock(KafkaConnectApi.class);
+        when(mockConnectClient.list(any(), anyString(), anyInt())).thenReturn(Future.succeededFuture(emptyList()));
+        when(mockConnectClient.updateConnectLoggers(any(), anyString(), anyInt(), anyString(), any(OrderedProperties.class))).thenReturn(Future.succeededFuture());
+        return mockConnectClient;
+    }
```
scholzj, line 966: Is this actually used anywhere? Or can it be removed?

cluster-operator/src/test/java/io/strimzi/operator/cluster/model/KafkaBrokerConfigurationBuilderTest.java
```diff
@@ -1944,37 +1944,37 @@ public void testOAuthOptions()  {
...
+        KafkaBrokerConfigurationBuilder.addBooleanOptionIfFalse(expectedOptions, ServerConfig.OAUTH_CHECK_ACCESS_TOKEN_TYPE, false);
+        KafkaBrokerConfigurationBuilder.addOptionIfNotNull(expectedOptions, ServerConfig.OAUTH_VALID_TOKEN_TYPE, "access_token");
+        KafkaBrokerConfigurationBuilder.addOptionIfNotNull(expectedOptions, ServerConfig.OAUTH_SSL_ENDPOINT_IDENTIFICATION_ALGORITHM, "");
+        KafkaBrokerConfigurationBuilder.addOptionIfNotNull(expectedOptions, ServerConfig.OAUTH_CONNECT_TIMEOUT_SECONDS, "30");
+        KafkaBrokerConfigurationBuilder.addOptionIfNotNull(expectedOptions, ServerConfig.OAUTH_READ_TIMEOUT_SECONDS, "60");
+        KafkaBrokerConfigurationBuilder.addOptionIfNotNull(expectedOptions, ServerConfig.OAUTH_HTTP_RETRIES, "2");
+        KafkaBrokerConfigurationBuilder.addOptionIfNotNull(expectedOptions, ServerConfig.OAUTH_HTTP_RETRY_PAUSE_MILLIS, "500");
+        KafkaBrokerConfigurationBuilder.addBooleanOptionIfTrue(expectedOptions, ServerConfig.OAUTH_ENABLE_METRICS, true);
+        KafkaBrokerConfigurationBuilder.addBooleanOptionIfFalse(expectedOptions, ServerConfig.OAUTH_FAIL_FAST, false);
```
scholzj, line 1974: Is this change really necessary? In a way, you create the expected result using the same methods used to create the actual result which is wrong. The expected result should be as simple as possible -> so not calling any of these methods from the production code.

ShazaAldawamneh: I haven't noticed that, thanks Jakub

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/AuthenticationUtils.java
```diff
@@ -49,7 +49,7 @@ public class AuthenticationUtils {
     /**
      * Key for OAuth config
      */
```
scholzj, line 51: Maybe you can remove the comment as well? It is not needed now and it is not really that helpful.

scholzj (approved):

PR #8902, opened 2023-07-25
Fixed getting CO log when multiple replicas are running

ppatierno:
When the CO is running with multiple replicas (leader + standby(s)), the current `report.sh` tool is not able to get the logs from them. No one is returned in the zip file.
This PR fixes the issues, handling the CO pods as an array and looping across them to recover logs (the same way it works for ZooKeeper pods, Kafka pods and so on).


ppatierno:
This turned out from this discussion when the user provided us the report.zip file but it didn't contain the CO log.
Investigating, it was because he was running 2 replicas of the CO, so the above issue in the report script.

scholzj (approved):

fvaleri (approved):

PR #8903, opened 2023-07-25
Fix ZooKeeper quorum port binding all interfaces with at least 3 nodes

ppatierno:
From this discussion https://github.com/orgs/strimzi/discussions/8890 it turned out that when the ZooKeeper ensemble is made by just 2 nodes, the quorum port binding on all interfaces (by using 0.0.0.0) doesn't work properly.
The 2 nodes seem to ping-pong on leading and never reach a quorum.
This PR fixes the issue by using the 0.0.0.0 binding only if the ZooKeeper ensemble is made by at least 3 nodes.


scholzj (approved):

fvaleri (approved):

showuon (approved):

ppatierno:
@scholzj as mentioned in the discussion above, do we want to cherry-pick this for 0.36.1? Wdyt?

PR #8908, opened 2023-07-26
Bumped OpenTelemetry 1.19.0 for the Kafka 3rd party libs

ppatierno:
This trivial PR bumps OpenTelemetry to 1.19.0 in the Kafka 3rd party libs to align with what we have in the root pom and the Strimzi HTTP bridge.


scholzj (approved):
I'm fine with this if this is what you want to use. But didn't we talked about moving to 1.23 in the discussion around Debezium?

ppatierno:
I would like to check the 1.23 on the bridge first to see if there is something more than what was described in that discussion. Let's keep this for now.

PR #8910, opened 2023-07-26
Enabling Kafka Agent to run when KRaft is enabled

ppatierno:
This PR enables the Kafka Agent to run when KRaft is anabled as well.
It's the base for future work about getting Kafka controller configuration via Kafka Agent (see https://github.com/strimzi/strimzi-kafka-operator/pull/8812) as well as improved liveness/readiness probes in KRaft mode (see https://github.com/strimzi/strimzi-kafka-operator/pull/8892).

### DIFF
--- a/docker-images/kafka-based/kafka/scripts/kafka_run.sh
+++ b/docker-images/kafka-based/kafka/scripts/kafka_run.sh
@@ -70,12 +70,16 @@ if [ "$STRIMZI_KRAFT_ENABLED" = "true" ]; then
   fi
 
 else
-  rm -f /var/opt/kafka/kafka-ready /var/opt/kafka/zk-connected 2> /dev/null
-  KEY_STORE=/tmp/kafka/cluster.keystore.p12
-  TRUST_STORE=/tmp/kafka/cluster.truststore.p12
-  KAFKA_OPTS="${KAFKA_OPTS} -javaagent:$(ls "$KAFKA_HOME"/libs/kafka-agent*.jar)=/var/opt/kafka/kafka-ready:/var/opt/kafka/zk-connected:$KEY_STORE:$CERTS_STORE_PASSWORD:$TRUST_STORE:$CERTS_STORE_PASSWORD"
-  export KAFKA_OPTS
+  KAFKA_READY=/var/opt/kafka/kafka-ready
+  ZK_CONNECTED=/var/opt/kafka/zk-connected
+  rm -f $KAFKA_READY $ZK_CONNECTED 2> /dev/null
 fi
+
+KEY_STORE=/tmp/kafka/cluster.keystore.p12
+TRUST_STORE=/tmp/kafka/cluster.truststore.p12
+# when in KRaft mode, the Kafka ready and ZooKeeper connected file paths are empty because not needed to the agent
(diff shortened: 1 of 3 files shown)


ppatierno:
After chatting with @tombentley and @tinaselenge we thought that it's better having this enablement as an atomic PR instead of mixing with other stuff.

tinaselenge (approved):
LGTM!

docker-images/kafka-based/kafka/scripts/kafka_run.sh
```diff
@@ -70,13 +70,17 @@ if [ "$STRIMZI_KRAFT_ENABLED" = "true" ]; then
...
 
 else
-  rm -f /var/opt/kafka/kafka-ready /var/opt/kafka/zk-connected 2> /dev/null
-  KEY_STORE=/tmp/kafka/cluster.keystore.p12
-  TRUST_STORE=/tmp/kafka/cluster.truststore.p12
-  KAFKA_OPTS="${KAFKA_OPTS} -javaagent:$(ls "$KAFKA_HOME"/libs/kafka-agent*.jar)=/var/opt/kafka/kafka-ready:/var/opt/kafka/zk-connected:$KEY_STORE:$CERTS_STORE_PASSWORD:$TRUST_STORE:$CERTS_STORE_PASSWORD"
-  export KAFKA_OPTS
+  KAFKA_READY=/var/opt/kafka/kafka-ready
+  ZK_CONNECTED=/var/opt/kafka/zk-connected
+  rm -f $KAFKA_READY $ZK_CONNECTED 2> /dev/null
```
scholzj, line 75: You define these variables only in this branch of the condition. But you use them everywhere. That seems wrong.

scholzj (commented):
How does this fit with https://github.com/strimzi/proposals/blob/main/046-kraft-liveness-readiness.md?

docker-images/kafka-based/kafka/scripts/kafka_run.sh
ppatierno: They are passed just empty/not existing when KRaft mode is enabled, because the Kafka Agent doesn't need to handle those file in such a mode.

scholzj: They are not passed as empty - they are not defined, which seems quite hacky.

ppatierno: Tbh I don't see it hacky. If you check the code in the `KafkaAgent`, in the end they are got as empty which is what we want. I am not sure it's really worth having lines like

```shell
KAFKA_READY=
ZK_CONNECTED=
```

in the KRaft condition if/else.

scholzj: If nothing else, it would make it clear that this is intentional.

scholzj: I'm quite surprised shell check does not complain about uninitialized variables.

ppatierno: @scholzj I added them as empty in the KRaft mode if/else branch with comment explaining the why.

scholzj (approved):

scholzj:
@ppatierno I'm still not sure I follow this:

> How does this fit with https://github.com/strimzi/proposals/blob/main/046-kraft-liveness-readiness.md?

tinaselenge:
> @ppatierno I'm still not sure I follow this:
> 
> > How does this fit with https://github.com/strimzi/proposals/blob/main/046-kraft-liveness-readiness.md?

Readiness probe for KRaft would be using KafkaAgent's API which is being implemented here https://github.com/strimzi/strimzi-kafka-operator/pull/8892.

PR #8918, opened 2023-07-27
[ST] Add system tests for the report.sh bash script

fvaleri:
This adds system tests for the report.sh bash script.

Running with 3 ZK instances to avoid issues with DNS binding.
Covers deployment with more than one cluster operator instance.

There are also a few changes to the report script:

- Fixed CO replicaset that was missing "-o yaml" flag, thus producing an invalid YAML
- Changed [hidden] with <hidden> as key/pwd placeholder when secrets option is hidden
- Added logic to also get the CO pod YAML content
- Merged CO code blocks into one, like for the other resources
- Changed crds and crs dirs to customresourcedefinitions and customresources
- Moved events.txt from events dir to logs (they are similar)
- Changed podlogs dir name to logs, as this actually hosts container logs


systemtest/src/test/java/io/strimzi/systemtest/report/ReportScriptTest.java
```diff
@@ -0,0 +1,255 @@
...
+import org.junit.jupiter.api.extension.ExtensionContext;
+
+import java.io.File;
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.UUID;
+
+@Tag(REGRESSION)
+public class ReportScriptTest extends AbstractST {
```
scholzj, line 40: Should it follow the namings of all other STs? Methods prefixed with `test`, classes named with `ST`, etc.?

scholzj (commented):
I'm not really sure we want a system test for this TBH. This is a best effort tool with best effort basis maintenance. If we decide to have it, we should maybe have at least more coverage for the actual content of the file and also make it more future proof to cover things such as Node Pools or StableConnectIdentitites?

scholzj (commented):
Also, why is `[hidden]` worse to parse then `hidden`? I do not have problem to change it to something, but `hidden` without any markers will be completely lost when reading the files and logs and I don't think that as a plain english word it is easy to parse. I think we should have some special markers. If for some reason `[]` fo not work, maybe you can use something else.

systemtest/src/test/java/io/strimzi/systemtest/report/ReportScriptTest.java
fvaleri: Sure, thanks for raising.

fvaleri:
> I'm not really sure we want a system test for this TBH. This is a best effort tool with best effort basis maintenance.

Nevertheless it is used by many users, that rely on it to gather information when they need to debug or ask for help. I would prefer to migrate it to Java for better maintenance, but we can do it at some point and run this test to verify compatibility.

> If we decide to have it, we should maybe have at least more coverage for the actual content of the file and also make it more future proof to cover things such as Node Pools or StableConnectIdentitites

I can add NodePools and StableConnectIdentitites. Which file content would you like to check?

> Also, why is [hidden] worse to parse then hidden? 

The parser see this as an Array, but simply adding quotes works fine, so I'm reverting to '[hidden]'.

scholzj:
> > Also, why is [hidden] worse to parse then hidden?
> 
> The parser sees this as an Array, but simply adding quotes works fine, so I'm reverting to '[hidden]'.

But quotes can have other meaning. E.g. it is different whether some option is quoted or unquoted. If `[]` is issue because of arrays, maybe you can use `<>`?

> > If we decide to have it, we should maybe have at least more coverage for the actual content of the file and also make it more future proof to cover things such as Node Pools or StableConnectIdentitites
> 
> I can add NodePools and StableConnectIdentitites. Which file content would you like to check?

Maybe I missed it -> but if there is an ST for this, I would expect it to check all the logs etc. => that they are present and have the right content. The same for all the dumped YAMLs for example.

systemtest/src/test/java/io/strimzi/systemtest/report/ClusterReportST.java
```diff
@@ -0,0 +1,255 @@
...
+    private void assertValidReport(String outPath) {
+        assertThat("Output directory does not exist", Files.exists(Path.of(outPath)));
+        assertThat("Output ZIP does not exist",
+            FileUtils.listFilesWithSuffix(outPath, ".zip").length == 1);
+        assertThat("Kafka CRD does not exist",
+            Files.exists(Path.of(outPath + "/reports/crds/kafkas.kafka.strimzi.io.yaml")));
+        assertThat("ClusterOperator logs are missing", FileUtils.listFilesWithPrefix(
+            outPath + "/reports/podlogs", "strimzi-cluster-operator").length == 2);
+        assertThat("Kafka logs are missing", FileUtils.listFilesWithPrefix(
+            outPath + "/reports/podlogs", testStorage.getClusterName() + "-kafka").length == 3);
```
ppatierno, line 246: my concern is that we are validating just a few of the things that the report is returning, or? What about all the other stuff which are not pod logs?

fvaleri: Yes, I'll add more checks. If you have any suggestion, feel free to raise.

ppatierno: It's bad to say but maybe testing all files it can produce (based on the tests we have here)? I know it could be a lot ...

fvaleri:
> But quotes can have other meaning. E.g. it is different whether some option is quoted or unquoted. If [] is issue because of arrays, maybe you can use <>?

Done.

tools/report.sh
scholzj, line 1: I guess the changes here should be covered in the description.

tools/report.sh
```diff
@@ -283,16 +282,17 @@ if [[ -n $DC_DEPLOY ]]; then
 fi
 
 echo "customresources"
-mkdir -p "$OUT_DIR"/reports/crds "$OUT_DIR"/reports/crs
-CRDS=$($KUBE_CLIENT get crd -l app=strimzi -o name | cut -d "/" -f 2) && readonly CRDS
+mkdir -p "$OUT_DIR"/reports/customresourcedefinitions
```
scholzj, line 285: It says custom resources above. So why is the directory suddenly named custom resource definitions?

tools/report.sh
```diff
@@ -301,11 +301,11 @@ done
 echo "events"
 EVENTS=$($KUBE_CLIENT get event -n "$NAMESPACE" --ignore-not-found) && readonly EVENTS
 if [[ -n $EVENTS ]]; then
-  mkdir -p "$OUT_DIR"/reports/events
-  echo "$EVENTS" > "$OUT_DIR"/reports/events/events.txt
+  mkdir -p "$OUT_DIR"/reports/logs
+  echo "$EVENTS" > "$OUT_DIR"/reports/logs/events.txt
```
scholzj, line 305: Events are not logs. Maybe a separate directoy makes sense?

systemtest/src/test/java/io/strimzi/systemtest/report/ClusterReportST.java
```diff
@@ -0,0 +1,528 @@
...
+            .createInstallation()
+            .runInstallation();
+        this.testStorage = new TestStorage(extensionContext);
+        Kafka kafkaMainCr = KafkaTemplates.kafkaWithCruiseControl(testStorage.getClusterName(), 3, 3)
+            .editOrNewMetadata()
+                .addToAnnotations(Annotations.ANNO_STRIMZI_IO_NODE_POOLS, "enabled")
+            .endMetadata()
+            .build();
+        KafkaNodePool kafkaNodePoolCr =  KafkaNodePoolTemplates
+                .defaultKafkaNodePool("kafka", testStorage.getClusterName(), 3)
```
scholzj, line 221: If you wanna test how it works with node pools, you should probably create multiple non-default node pools. But then you likely need to also test it without the node pools to make sure it works without it as well.

tools/report.sh
fvaleri: Right, let me add it here:

### Fixes

- Fixed CO replicaset that was missing "-o yaml" flag, thus producing an invalid YAML

### Changes

- Changed [hidden] with <hidden> as key/pwd placeholder when secrets option is hidden
- Added logic to also get the CO pod YAML content
- Merged CO code blocks into one, like for the other resources
- Changed crds and crs dirs to customresourcedefinitions and customresources
- Moved events from events to logs directory (see motivation below)
- Changed podlogs dir name to logs, as this actually hosts container logs

tools/report.sh
fvaleri: This code block gets both CRDs and CRs, I didn't change that. It's similar to the cluster operator block that gets different resource types.

tools/report.sh
fvaleri: Well, they have a timestamp, a log level and a message so I thought it could be a good place for it. Additionally, you may want to have a look at them when checking logs. That said, if you prefer to have them in a dedicated dir, then I'm fine.

systemtest/src/test/java/io/strimzi/systemtest/report/ClusterReportST.java
fvaleri: Yeah, that was a compromise to be more future proof, but we can cover both.

tools/report.sh
ppatierno: I would agree with Jakub, let's keep events in a dedicated folder.

systemtest/src/test/java/io/strimzi/systemtest/report/DefaultClusterReportST.java
```diff
@@ -0,0 +1,448 @@
...
+        assertThat("Keys are not hidden in secrets", clusterCaSecret.getData().get(secretKey).equals("<hidden>"));
+
+        assertThat("KafkaBridge CRD does not exist", FileUtils.exists(outPath + "/reports/customresourcedefinitions/kafkabridges.kafka.strimzi.io.yaml"));
+        assertThat("KafkaBridge logs are missing", FileUtils.listFilesWithPrefix(outPath + "/reports/logs", "my-bridge-bridge").length == 1);
+
+        assertThat("KafkaConnect CRD does not exist", FileUtils.exists(outPath + "/reports/customresourcedefinitions/kafkaconnects.kafka.strimzi.io.yaml"));
+        assertThat("KafkaConnect logs are missing", FileUtils.listFilesWithPrefix(outPath + "/reports/logs", "my-connect-connect").length == 1);
+
+        assertThat("KafkaMirrorMaker2 CRD does not exist", FileUtils.exists(outPath + "/reports/customresourcedefinitions/kafkamirrormaker2s.kafka.strimzi.io.yaml"));
+        assertThat("KafkaMirrorMaker2 logs are missing", FileUtils.listFilesWithPrefix(outPath + "/reports/logs", "my-mm2-mirrormaker2").length == 1);
```
ppatierno, line 116: out of curiosity, why checking the "main" CRD so `Kafka` is missing here?

systemtest/src/test/java/io/strimzi/systemtest/report/DefaultClusterReportST.java
```diff
@@ -0,0 +1,448 @@
...
+            "my-bridge-bridge-config.yaml",
+            clusterName + "-cruise-control-config.yaml",
+            clusterName + "-entity-topic-operator-config.yaml",
+            clusterName + "-entity-user-operator-config.yaml",
+            clusterName + "-kafka-0.yaml",
+            clusterName + "-kafka-1.yaml",
+            clusterName + "-kafka-2.yaml",
+            clusterName + "-zookeeper-config.yaml",
+            "my-connect-connect-config.yaml",
+            "my-mm2-mirrormaker2-config.yaml"
```
ppatierno, line 241: do we want to check the `strimzi-cluster-operator` ConfigMap as well?

systemtest/src/test/java/io/strimzi/systemtest/report/DefaultClusterReportST.java
```diff
@@ -0,0 +1,448 @@
...
+            clusterName + "-zookeeper-2.log"
+        )) {
+            assertValidFiles(outPath + "/reports/logs", s, 1);
+        }
+        assertValidFiles(outPath + "/reports/logs", clusterName + "-cruise-control", 1);
+        assertValidFiles(outPath + "/reports/logs", "my-bridge-bridge", 1);
+        assertValidFiles(outPath + "/reports/logs", clusterName + "-entity-operator", 3);
+        assertValidFiles(outPath + "/reports/logs", "my-connect-connect", 1);
+        assertValidFiles(outPath + "/reports/logs", "my-mm2-mirrormaker2", 1);
+        assertValidFiles(outPath + "/reports/logs", "strimzi-cluster-operator", 2);
```
ppatierno, line 446: Maybe you could group the one with num=1 and put them in the for loop as other tests?

systemtest/src/test/java/io/strimzi/systemtest/report/NodePoolsClusterReportST.java
```diff
@@ -0,0 +1,488 @@
...
+        assertThat("Keys are not hidden in secrets", clusterCaSecret.getData().get(secretKey).equals("<hidden>"));
+
+        assertThat("KafkaBridge CRD does not exist", FileUtils.exists(outPath + "/reports/customresourcedefinitions/kafkabridges.kafka.strimzi.io.yaml"));
+        assertThat("KafkaBridge logs are missing", FileUtils.listFilesWithPrefix(outPath + "/reports/logs", "my-bridge-bridge").length == 1);
+
+        assertThat("KafkaConnect CRD does not exist", FileUtils.exists(outPath + "/reports/customresourcedefinitions/kafkaconnects.kafka.strimzi.io.yaml"));
+        assertThat("KafkaConnect logs are missing", FileUtils.listFilesWithPrefix(outPath + "/reports/logs", "my-connect-connect").length == 1);
+
+        assertThat("KafkaMirrorMaker2 CRD does not exist", FileUtils.exists(outPath + "/reports/customresourcedefinitions/kafkamirrormaker2s.kafka.strimzi.io.yaml"));
+        assertThat("KafkaMirrorMaker2 logs are missing", FileUtils.listFilesWithPrefix(outPath + "/reports/logs", "my-mm2-mirrormaker2").length == 1);
```
ppatierno, line 123: same question as other STs class, why 'Kafka` CR check missing here?

systemtest/src/test/java/io/strimzi/systemtest/report/NodePoolsClusterReportST.java
```diff
@@ -0,0 +1,488 @@
...
+            clusterName + "-zookeeper-1.yaml",
+            clusterName + "-zookeeper-2.yaml",
+            "my-connect-connect-0.yaml",
+            "my-mm2-mirrormaker2-0.yaml"
+        )) {
+            assertValidYamls(outPath + "/reports/pods", Pod.class, s, 1);
+        }
+        assertValidYamls(outPath + "/reports/pods", Pod.class, "my-bridge-bridge", 1);
+        assertValidYamls(outPath + "/reports/pods", Pod.class, clusterName + "-cruise-control", 1);
+        assertValidYamls(outPath + "/reports/pods", Pod.class, clusterName + "-entity-operator", 1);
```
ppatierno, line 342: Above three could be in the for loop?

systemtest/src/test/java/io/strimzi/systemtest/report/DefaultClusterReportST.java
```diff
@@ -0,0 +1,448 @@
...
+            clusterName + "-zookeeper-1.yaml",
+            clusterName + "-zookeeper-2.yaml"
+        )) {
+            assertValidYamls(outPath + "/reports/pods", Pod.class, s, 1);
+        }
+        assertValidYamls(outPath + "/reports/pods", Pod.class, "my-bridge-bridge", 1);
+        assertValidYamls(outPath + "/reports/pods", Pod.class, clusterName + "-cruise-control", 1);
+        assertValidYamls(outPath + "/reports/pods", Pod.class, clusterName + "-entity-operator", 1);
+        assertValidYamls(outPath + "/reports/pods", Pod.class, "my-connect-connect", 1);
+        assertValidYamls(outPath + "/reports/pods", Pod.class, "my-mm2-mirrormaker2", 1);
```
ppatierno, line 312: why these are out of the loop so not in the Arrays.asList declaration as done for other tests?
Apart from the next one which has num=2, they could be moved there.

systemtest/src/test/java/io/strimzi/systemtest/report/NodePoolsClusterReportST.java
```diff
@@ -0,0 +1,488 @@
...
+            clusterName + "-zookeeper-0.log",
+            clusterName + "-zookeeper-1.log",
+            clusterName + "-zookeeper-2.log",
+            "my-connect-connect-0.log",
+            "my-mm2-mirrormaker2-0.log"
+        )) {
+            assertValidFiles(outPath + "/reports/logs", s, 1);
+        }
+        assertValidFiles(outPath + "/reports/logs", clusterName + "-cruise-control", 1);
+        assertValidFiles(outPath + "/reports/logs", "my-bridge-bridge", 1);
```
ppatierno, line 484: Ditto as before?

systemtest/src/test/java/io/strimzi/systemtest/report/DefaultClusterReportST.java
fvaleri: Nope, as they leverage the prefix logic here (no stable identities in standard deploy).

systemtest/src/test/java/io/strimzi/systemtest/report/NodePoolsClusterReportST.java
fvaleri: Nope, as they leverage the prefix logic here (no stable identities in standard deploy).

systemtest/src/test/java/io/strimzi/systemtest/report/DefaultClusterReportST.java
fvaleri: That is due to my previous implementation, I'll bring them inside the loop.

tools/report.sh
fvaleri: ok

systemtest/src/test/java/io/strimzi/systemtest/report/DefaultClusterReportST.java
fvaleri: These are actually a refusal from my previous implementation, as they are already covered by previous validate methods. Let me remove them.

systemtest/src/test/java/io/strimzi/systemtest/report/DefaultClusterReportST.java
fvaleri: The script doesn't get this. Do we want to add it in this PR?

ppatierno (approved):
LGTM

fvaleri:
@gracegrimwood fyi

scholzj:
@im-konge Can you please have a look at this?

systemtest/src/test/java/io/strimzi/systemtest/report/DefaultClusterReportST.java
```diff
@@ -0,0 +1,445 @@
...
+        )) {
+            assertValidYamls(outPath + "/reports/poddisruptionbudgets", PodDisruptionBudget.class, s, 1);
+        }
+    }
+
+    private void assertValidPods(String outPath, String clusterName) throws IOException {
+        for (String s : Arrays.asList(
+            clusterName + "-kafka-0.yaml",
+            clusterName + "-kafka-1.yaml",
+            clusterName + "-kafka-2.yaml",
```
im-konge, line 293: I would use methods from API instead of "hardcoding" it here.
So f.e. `KafkaResources.kafkaPodName()` etc.

This applies for all other things -> `KafkaMirrorMaker2Resources.deploymentName(..)`

Also, we have some helper methods for Kafka with/without NodePools, if you want to use them (in `KafkaResource` class).

systemtest/src/test/java/io/strimzi/systemtest/report/NodePoolsClusterReportST.java
```diff
@@ -0,0 +1,486 @@
...
+import java.io.IOException;
+import java.util.Arrays;
+
+import static io.strimzi.systemtest.Constants.REGRESSION;
+import static io.strimzi.test.TestUtils.USER_PATH;
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+
+@Tag(REGRESSION)
+public class NodePoolsClusterReportST extends AbstractClusterReportST {
```
im-konge, line 63: Only difference between those two classes are NodePools, so what if we have just one testsuite with the tests, in which we would switch between Kafka with/without NodePools based on `STRIMZI_FEATURE_GATES` env variable?

Yeah it would need few "hacks" and `if`s to make it work, but more or less all helper methods for naming are already in the suite.

The report.sh tests with NodePools would run during KRaft and FG Azure pipelines, the tests for Kafka without NodePools would run during regular pipelines. We will save execution time and also lines of code, but both modes will be tested.

WDYT?

im-konge (commented):
Thanks for adding the STs for the `report.sh`, I left few comments.
But the scenarios LGTM.

systemtest/src/test/java/io/strimzi/systemtest/report/AbstractClusterReportST.java
```diff
@@ -0,0 +1,64 @@
...
+            testInfo.getTestMethod().get().getName() : UUID.randomUUID().toString();
+        return USER_PATH + "/target/reports/" + clusterName + "/" + methodName;
+    }
+
+    protected static Secret getSecretWithKeyFromFile(String filePath, String key) throws IOException {
+        Secret secret = MAPPER.readValue(new File(filePath), Secret.class);
+        assertThat("The secret {} does not exist", FileUtils.exists(filePath));
+        assertThat("The key {} is not present", secret.getData().get(key) != null);
+        return secret;
+    }
```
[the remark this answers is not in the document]
fvaleri: Yes, makes sense.

systemtest/src/main/java/io/strimzi/systemtest/utils/FileUtils.java
```diff
@@ -116,4 +117,16 @@ public static File updateNamespaceOfYamlFile(String pathToOrigin, String namespa
         }
         return null;
     }
+
+    public static boolean exists(String path) {
```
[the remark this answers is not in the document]
fvaleri: Ok.

systemtest/src/test/java/io/strimzi/systemtest/report/DefaultClusterReportST.java
```diff
@@ -0,0 +1,445 @@
...
+    void setUp(final ExtensionContext extensionContext) {
+        this.clusterOperator = this.clusterOperator
+            .defaultInstallation(extensionContext)
+            .withReplicas(2)
+            .createInstallation()
+            .runInstallation();
+        this.testStorage = new TestStorage(extensionContext);
+        resourceManager.createResourceWithWait(extensionContext,
+            KafkaTemplates.kafkaWithCruiseControl(testStorage.getClusterName(), 3, 3).build(),
+            KafkaTemplates.kafkaEphemeral(testStorage.getClusterName() + "-tgt", 3).build()
```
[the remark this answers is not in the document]
fvaleri: I had problems with the ZK issue while working on this, that's why I used 3 replicas for the target cluster, but I can reduce to 1 replica if you are happy.

systemtest/src/test/java/io/strimzi/systemtest/report/DefaultClusterReportST.java
fvaleri: I'm taking these values from testStorage. Can you be more specific?

systemtest/src/test/java/io/strimzi/systemtest/report/NodePoolsClusterReportST.java
fvaleri: In the NodePools implemenation we also have StableConnectIdentities. Some checks are different, and some are missing. 

In general, I don't like the idea of trading code maintenance for infrastructure convenience. Can't we simply tag them differently, so that they can be activated only when a specific pipeline is running?

systemtest/src/test/java/io/strimzi/systemtest/report/DefaultClusterReportST.java
im-konge: I mean for the file names use methods from API -> for example instead of `clusterName + "-kafka-0.yaml",` use `KafkaResources.kafkaPodName(clusterName, 0) + ".yaml"`

So instead of having those values hardcoded, use methods from API module, so in case of any change, we will not have to fix the test.

systemtest/src/test/java/io/strimzi/systemtest/report/NodePoolsClusterReportST.java
im-konge: Hmm I see. I know that having one class which will run different checks based on env variables can be confusing.
We can definitely have some annotation to trigger these tests when `NodePools` and `StableConnectIdentities` will be enabled, that's not a problem.

Also to remove "extra lines of code", we can have something similar as for tracing tests -> so in the abstract have methods with checks (something like `doTestCreateReport`) and in the two ST classes just call them? I think that there will be common parts of code, which can be reused + maybe changed for handling both NodePools and non NodePools. (it's just an idea)

systemtest/src/test/java/io/strimzi/systemtest/report/DefaultClusterReportST.java
im-konge: The resources are spread across `*Resources` classes, but regarding nodepools, I guess you will have to use the ones in `KafkaResource` class in `systemtest` module

fvaleri: I'm confused now. Before you were mentioning `KafkaResources` in api module, now `KafkaResource` and other `*Resource` classes in systemtest module. 

For example, I can get the first kafka pod name with both `KafkaResources.kafkaPodName(clusterName, 0)` and `KafkaResource.getKafkaPodName("my-cluster", 0)`.

Which classes do I have to use?

im-konge: In case you have testcase, where you either use NodePools mode or non-NodePools mode based on the env variable, you should use `KafkaResource.getKafkaPodName()` in the `systemtest` module. In other case, you can use `KafkaResources` (and other classes) in `api` module

scholzj (changes requested):
Please don't add these names to the `api` module without a proper discussion. This is a public module which should be treated with care.

im-konge:
I didn't say that you should add new classes to `api` module. I said that you can use the methods from `*Resources` classes in `api` module and if you want to handle STs based on env variable (if the tests should run with NodePools enabled/disabled), you can use `KafkaResource` getters.

scholzj:
> > Please don't add these names to the api module without a proper discussion. This is a public module which should be treated with care.
> 
> This change was requested by @im-konge. Do we want to add new methods to *Resource classes in systemtest module?

Sorry for this misunderstanding -> but I think introducing all of these fields into the public API should be probably avoided or at least would require more discussion. These classes are a lot of mess already without it :-().

im-konge:
@fvaleri I can go write suggestions - to help you which methods from `api` you can use (as we are using them in STs) if you want

fvaleri:
After rebasing, the tests fail because there are two namespaces now. The resource manager deploys all components in co-namespace and Strimzi CRs in test-suite-namespace.

> I didn't say that you should add new classes to `api` module...

I thought it was kind of implicit in your request, because there are many names that are missing in *Resources classes. The end result would be a mix of hard coded strings and calls to methods from *Resources classes. Is this what you want?

I can add all the missing resource names to the abstract class. Let me know if this works.

> if you want to handle STs based on env variable (if the tests should run with NodePools enabled/disabled), you can use KafkaResource getters.

I would prefer to keep them as they are for better maintenance. We kind of agreed to use annotations to trigger these tests when needed: https://github.com/strimzi/strimzi-kafka-operator/pull/8918#discussion_r1292526875

im-konge:
> I thought it was kind of implicit in your request, because there are many names that are missing in *Resources classes. The end result would be a mix of hard coded strings and calls to methods from *Resources classes. Is this what you want?

When there are no methods for that in `*Resources` classes, we are adding some constants to the `Constants` class, we have few in `AbstractST` IIRC.
Let me check all the names that are needed - I will then think where we should add those that are missing (I mean where in `systemtest` module), thanks.

> I would prefer to keep them as they are for better maintenance. We kind of agreed to use annotations to trigger these tests when needed: https://github.com/strimzi/strimzi-kafka-operator/pull/8918#discussion_r1292526875

Sorry, I completely forgot that we agreed on that 😄 . Yeah, good point.

> After rebasing, the tests fail because there are two namespaces now. The resource manager deploys all components in co-namespace and Strimzi CRs in test-suite-namespace.

Do you want help with that?

fvaleri:
> Do you want help with that?

I can figure this out, thanks. I want a single namespace for these STs, so I'll use the CO_NAMESPACE constant.

fvaleri:
@im-konge can you have another look when you have time? Thanks

systemtest/src/test/java/io/strimzi/systemtest/report/AbstractClusterReportST.java
```diff
@@ -0,0 +1,161 @@
...
+        return clusterName + "-kafka";
+    }
+
+    protected String zookeeperPodDisruptionBudgetName(String clusterName) {
+        return clusterName + "-zookeeper";
+    }
+
+    protected String entityOperatorNetworkPolicyName(String clusterName) {
+        return clusterName + "-entity-operator";
+    }
```
im-konge, line 156: `KafkaResources.entityOperatorDeploymentName(clusterName)`

systemtest/src/test/java/io/strimzi/systemtest/report/DefaultClusterReportST.java
```diff
@@ -0,0 +1,453 @@
...
+        }
+    }
+
+    private void assertValidPodDisruptionBudgets(String outPath, String clusterName) throws IOException {
+        for (String s : Arrays.asList(
+            bridgePodDisruptionBudgetName("my-bridge") + ".yaml",
+            kafkaPodDisruptionBudgetName(clusterName) + ".yaml",
+            zookeeperPodDisruptionBudgetName(clusterName) + ".yaml",
+            kafkaConnectPodDisruptionBudgetName("my-connect") + ".yaml",
+            kafkaMirrorMaker2PodDisruptionBudgetName("my-mm2") + ".yaml"
```
im-konge, line 291: Shouldn't these `my-bridge`, `my-connect`, `my-mm2` be in some class constants?

systemtest/src/test/java/io/strimzi/systemtest/report/DefaultClusterReportST.java
```diff
@@ -0,0 +1,453 @@
...
+    private void assertValidKafkaUsers(String outPath) throws IOException {
+        assertValidYamls(outPath + "/reports/kafkausers", KafkaUser.class, "my-user.yaml", 1);
+    }
+
+    private void assertValidStrimziPodSets(String outPath, String clusterName) throws IOException {
+        for (String s : Arrays.asList(
+            clusterName + "-kafka.yaml",
+            clusterName + "-tgt-kafka.yaml",
+            clusterName + "-tgt-zookeeper.yaml",
+            clusterName + "-zookeeper.yaml"
```
im-konge, line 412: You can again use some methods from the abstract or?

systemtest/src/test/java/io/strimzi/systemtest/report/DefaultClusterReportST.java
```diff
@@ -0,0 +1,453 @@
...
+    void setUp(final ExtensionContext extensionContext) {
+        this.clusterOperator = this.clusterOperator
+            .defaultInstallation(extensionContext)
+            .withReplicas(2)
+            .createInstallation()
+            .runInstallation();
+        this.testStorage = new TestStorage(extensionContext, Constants.CO_NAMESPACE);
+        resourceManager.createResourceWithWait(extensionContext,
+            KafkaTemplates.kafkaWithCruiseControl(testStorage.getClusterName(), 3, 3).build(),
+            KafkaTemplates.kafkaEphemeral(testStorage.getClusterName() + "-tgt", 1).build()
```
im-konge, line 193: You can create class variable for the target cluster and here you can assign value to it, then use it in all other methods, or?

systemtest/src/test/java/io/strimzi/systemtest/report/DefaultClusterReportST.java
```diff
@@ -0,0 +1,453 @@
...
+import java.io.IOException;
+import java.util.Arrays;
+
+import static io.strimzi.systemtest.Constants.REGRESSION;
+import static io.strimzi.test.TestUtils.USER_PATH;
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+
+@Tag(REGRESSION)
+public class DefaultClusterReportST extends AbstractClusterReportST {
```
im-konge, line 63: To not trigger these tests when NodePools will be enabled, you can add `@NodePoolsNotSupported`

systemtest/src/test/java/io/strimzi/systemtest/report/NodePoolsClusterReportST.java
```diff
@@ -0,0 +1,492 @@
...
```
im-konge, line 69: For skipping this suite when NodePools will be disabled, we will need to create new annotation

im-konge (commented):
I left some comments, looks good overall. Thanks :)

systemtest/src/test/java/io/strimzi/systemtest/report/AbstractClusterReportST.java
```diff
@@ -0,0 +1,161 @@
...
+        return KafkaResources.kafkaPodName(clusterName, podNum);
+    }
+
+    protected String kafkaNodePoolsPodName(String clusterName, String nodePoolName, int podNum) {
+        return clusterName + "-" + nodePoolName + "-" + podNum;
+    }
+
+    protected String kafkaNodePoolsConfigMapName(String clusterName, String nodePoolName, int podNum) {
+        return kafkaNodePoolsPodName(clusterName, nodePoolName, podNum);
+    }
```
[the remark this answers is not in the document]
fvaleri: this works with an env var, so I'm keeping my method

systemtest/src/test/java/io/strimzi/systemtest/report/NodePoolsClusterReportST.java
```diff
@@ -0,0 +1,497 @@
...
+import java.util.Arrays;
+
+import static io.strimzi.systemtest.Constants.NODE_POOLS;
+import static io.strimzi.systemtest.Constants.REGRESSION;
+import static io.strimzi.test.TestUtils.USER_PATH;
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+
+@Tag(REGRESSION)
+@Tag(NODE_POOLS)
```
im-konge, line 70: Sorry for possible confusion, I meant creating annotation like https://github.com/strimzi/strimzi-kafka-operator/blob/main/systemtest/src/main/java/io/strimzi/systemtest/annotations/NodePoolsNotSupported.java

but in the opposite way. So maybe something like `NodePoolsOnly`.
Otherwise we would need to update all the regression pipelines with the new tag and exclude it there, so FMPOV the annotation makes these things easier.

But, if you don't want to add it in this PR, I can do it in another one, no problem at all.

im-konge (commented):
STs looks great, thanks for applying all of the comments I had. Just one comment for the annotation.

Thanks once more! 🙁

im-konge (approved):
Thank you :)

scholzj:
@fvaleri FYI: The CI failure seems to be related to your code.

fvaleri:
I see. I'll have a look. Thanks.

fvaleri:
@im-konge feel free to push the Tag change to this PR, so that we can run the two tests in different pipelines and avoid any resource issue.

fvaleri:
@scholzj CI issues should be fixed now. Once #9088 is merged I'll rebase and add the NodePoolsOnly annotation.

scholzj:
Seems to be still failing. It also seems to be still running both tests, so the change from @im-konge didn't work I guess.

fvaleri:
@im-konge both tests are working consistently on my local environment. I added the `@NodePoolsOnly` to the NodePoolsClusterReportST, but I think this is not enough. We probably need another annotation that says DefaultClusterReportST should run only when NodePools are not used. Wdyt?

1. On CI we have the following error for DefaultClusterReportST.

   ```sh
   2023-09-08T05:21:01.1476023Z java.lang.AssertionError: Missing files of type ConfigMap with  prefix my-cluster-7ded8a15-kafka-0.yaml
   ```

   The issue here is that the pipeline changes some deployed resource name, so we get `my-cluster-7ded8a15-kafka-pool-65678a8c-0`, instead of `my-cluster-7ded8a15-kafka-0`.

2. On CI we have the following error for NodePoolsClusterReportST.

   ```sh
   2023-09-08T05:43:50.3636001Z java.lang.AssertionError: Missing files with prefix 'my-cluster-2524736c-pool-b-10.log'
   ```

   This doesn't make any sense to me, because I see that the pod `my-cluster-2524736c-pool-b-10` is there, and we have `logs-pod-my-cluster-8f08a447-pool-b-10-container-kafka.log` from the pipeline results.

im-konge:
@fvaleri Yeah sorry, I maybe mentioned it somewhere in the comments - could you please add `@NodePoolsNotSupported` annotation to the `DefaultClusterReportST` ? 

> NodePoolsClusterReportST error.

2023-09-08T05:43:50.3636001Z java.lang.AssertionError: Missing files with prefix 'my-cluster-2524736c-pool-b-10.log'

This doesn't make any sense to me, because I see that the pod my-cluster-2524736c-pool-b-10 is there, and we have logs-pod-my-cluster-8f08a447-pool-b-10-container-kafka.log from the pipeline results.

I guess it shouldn't be prefix or? Let me check the code and the results

im-konge:
> I guess it shouldn't be prefix or? Let me check the code and the results

Yeah I guess that the problem is that you have `logs-pod-my-cluster-8f08a447-pool-b-10-container-kafka.log`, but you are listing files with prefix `my-cluster-2524736c-pool-b-10.log` -> you should look for files starting with `logs-pod-my-cluster-2524736c-pool-b-10`, or?

im-konge:
> No, logs-pod-* are in the system test results folder, not in the script output folder. The prefix is correct, and it is working on other bundles, so I guess it's something related to the Kraft bundle, where we shouldn't run (not yet). If you think the annotations are now correct, let's try to rerun the pipelines.

I'm not sure that the annotation you added to the `DefaultClusterReportST` will resolve this issue in `NodePoolsClusterReportST`. ` logs-pod-my-cluster-8f08a447-pool-b-10-container-kafka.log` is log from the STs log collector, right? Sorry, I little bit missed this. Maybe the script wasn't able to gather the logs from the pod?

fvaleri:
> Maybe the script wasn't able to gather the logs from the pod?

I would expect it to fail with all nodepools logs, and the other CI bundles then. Maybe it's because the pipeline was terminated. Not sure about this.

fvaleri:
@im-konge are we good?

#### regression-bundle VII: DefaultClusterReportST success, NodePoolsClusterReportST skipped

```sh
2023-09-12T15:57:55.1028447Z [INFO] Tests run: 8, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 672.305 s - in io.strimzi.systemtest.report.DefaultClusterReportST
```

#### feature-gates-regressionbundle VII: DefaultClusterReportST skipped, NodePoolsClusterReportST success

```sh
2023-09-12T16:05:26.9856515Z [WARNING] Tests run: 9, Failures: 0, Errors: 0, Skipped: 1, Time elapsed: 782.584 s - in io.strimzi.systemtest.report.NodePoolsClusterReportST
```

The error on `regression-bundle IV` seems to be unrelated:

```sh
[ERROR] Errors: 
[ERROR] io.strimzi.systemtest.operators.topic.TopicST.testDeleteTopicEnableFalse(ExtensionContext)
[ERROR]   Run 1: TopicST.testDeleteTopicEnableFalse:263 » NullPointer Cannot invoke "io.strimzi.api.kafka.model.KafkaTopic.getMetadata()" because the return value of "io.fabric8.kubernetes.client.dsl.Resource.get()" is null
[ERROR]   Run 2: TopicST.testDeleteTopicEnableFalse:263 » NullPointer Cannot invoke "io.strimzi.api.kafka.model.KafkaTopic.getMetadata()" because the return value of "io.fabric8.kubernetes.client.dsl.Resource.get()" is null
[ERROR]   Run 3: TopicST.testDeleteTopicEnableFalse:263 » NullPointer Cannot invoke "io.strimzi.api.kafka.model.KafkaTopic.getMetadata()" because the return value of "io.fabric8.kubernetes.client.dsl.Resource.get()" is null
```

im-konge:
@fvaleri it seems alright 

I will run also KRaft pipelines, just to be sure

im-konge:
It seems that this line https://github.com/strimzi/strimzi-kafka-operator/pull/8918/files#diff-0778e9527e7a540a55c8b18a53b3becdd80b4f1b970932ea362946f07bd2a4d2R200 causes issue for KRaft. You will need to create another NodePool to "workaround" this. As by default, `ResourceManager` checks if KRaft is enabled for the run and creates the NodePool with just `broker` role.

fvaleri:
Thanks for looking into that @im-konge.

Can you be more specific about the workaround for NodePoolsClusterReportST?

im-konge:
Sorry that STs are that complicated ... I think that creating one more NodePool for the second Kafka cluster (`my-cluster-XXX-tgt`) should fix the issue.

fvaleri:
Done. Let me know if this works.

Have you thought about giving the possibility to completely disable the entity operator when deploying the Kafka cluster in a ST? In this case, I don't need that for the tgt cluster.

im-konge:
> Have you thought about giving the possibility to completely disable the entity operator when deploying the Kafka cluster in a ST? In this case, I don't need that for the tgt cluster.

TBH no. We are deploying Kafka clusters without EO (or just UO/TO) in few specific cases. But yeah, I can imagine few testcases that could use that. Thanks for the suggestion 👍

fvaleri:
I'm going to close it, since this has been opened for two months, and there is still no agreement on whether we need it or not. In a separate PR, I'm going to extract all the fixes and small changes I did to the `report.sh` while creating the test.

PR #8940, opened 2023-08-02
Improve `LeaderElectionManager` to release the lock at shutdown

scholzj:
### Type of change

- Enhancement / new feature

### Description

Today, when one operator instance is shutdown, it does not release the leader election lock. That means that any other CO instance will need to wait for the lock to expire before it can take it and start operating.

This PR improves the `LeaderElectionManager` to release the lock at the shutdown. To do that, it does several changes:
* It stops running its own thread and instead is using the `CompletableFuture` of the Fabric8 Leader Elector. This allows to easily stop the elector and get it to release the lock.
* It reworks the shutdown hook which was previously shared between TO and CO and was solely focused on shutting down Vert.x. It was rewritten to be able to run several shutdown hooks in a serial fashion, so that we can shutdown things in the right order. For example, first the operator itself, then the Leader Election manager and finally Vert.x itself.

Thanks to these changes, when the operator instance now shuts down, it releases the lock and the other instance can take over immediately if it runs in parallel or right after it starts if it has to be freshly started.

### Checklist

- [x] Write tests

### DIFF
--- a/operator-common/src/main/java/io/strimzi/operator/common/ShutdownHook.java
+++ b/operator-common/src/main/java/io/strimzi/operator/common/ShutdownHook.java
@@ -44,10 +44,10 @@ public class ShutdownHook implements Runnable {
     /**
      * Registers a lambda which should be called during the shutdown
      *
-     * @param c     Function which should be called when shutting down
+     * @param shutdownFunction  Function which should be called when shutting down
      */
-    public void register(Runnable c)    {
-        shutdownStack.push(c);
+    public void register(Runnable shutdownFunction)    {
+        shutdownStack.push(shutdownFunction);
     }
 
     /**
@@ -63,17 +63,17 @@ public class ShutdownHook implements Runnable {
 
         vertx.close(ar -> {
             if (!ar.succeeded()) {
-                LOGGER.error("Vertx close failed", ar.cause());
+                LOGGER.error("Vert.x close failed", ar.cause());
             }
             latch.countDown();
         });
 
         try {
             if (!latch.await(timeoutMs, TimeUnit.MILLISECONDS)) {
-                LOGGER.error("Timed out while waiting for Vertx close");
+                LOGGER.error("Timed out while waiting for Vert.x close");
             }
         } catch (InterruptedException e) {
-            LOGGER.error("Interrupted while waiting for Vertx close");
+            LOGGER.error("Interrupted while waiting for Vert.x close");
         }
 
         LOGGER.info("Shutdown of Vert.x is complete");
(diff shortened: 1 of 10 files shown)


operator-common/src/main/java/io/strimzi/operator/common/ShutdownHook.java
```diff
@@ -5,64 +5,106 @@
...
+        }
+
+        LOGGER.info("Shutdown hook completed");
+    }
+
+    /**
+     * Registers a lambda which should be called during the shutdown
      *
-     * @param vertx     Vert.x instance
+     * @param c     Function which should be called when shutting down
```
ppatierno, line 47: anything better than just `c` here? :-/

operator-common/src/main/java/io/strimzi/operator/common/ShutdownHook.java
```diff
@@ -5,64 +5,106 @@
...
+            }
+            latch.countDown();
+        });
+
+        try {
+            if (!latch.await(timeoutMs, TimeUnit.MILLISECONDS)) {
+                LOGGER.error("Timed out while waiting for Vertx close");
+            }
+        } catch (InterruptedException e) {
+            LOGGER.error("Interrupted while waiting for Vertx close");
```
ppatierno, line 76:
```suggestion
            LOGGER.error("Interrupted while waiting for Vert.x close");
```

operator-common/src/main/java/io/strimzi/operator/common/ShutdownHook.java
```diff
@@ -5,64 +5,106 @@
...
+     * @param vertx         Vert.x instance
+     * @param verticleId    ID of the verticle which should be undeployed
+     * @param timeoutMs     Timeout for the shutdown
+     */
+    public static void undeployVertxVerticle(Vertx vertx, String verticleId, long timeoutMs) {
+        LOGGER.info("Shutting down Vert.x verticle {}", verticleId);
+        CountDownLatch latch = new CountDownLatch(1);
+        vertx.undeploy(verticleId, ar -> {
+            if (!ar.succeeded()) {
+                LOGGER.error("Vertx verticle failed to undeploy", ar.cause());
```
ppatierno, line 94:
```suggestion
                LOGGER.error("Vert.x verticle failed to undeploy", ar.cause());
```

operator-common/src/main/java/io/strimzi/operator/common/ShutdownHook.java
```diff
@@ -5,64 +5,106 @@
...
             }
+
+            latch.countDown();
+        });
+
+        try {
+            if (!latch.await(timeoutMs, TimeUnit.MILLISECONDS)) {
+                LOGGER.error("Timed out while waiting for Vertx verticle to undeploy");
```
ppatierno, line 102:
```suggestion
                LOGGER.error("Timed out while waiting for Vert.x verticle to undeploy");
```

operator-common/src/main/java/io/strimzi/operator/common/ShutdownHook.java
```diff
@@ -5,64 +5,106 @@
...
+            }
+        } catch (InterruptedException e) {
+            LOGGER.error("Interrupted while waiting for Vertx verticle to undeploy");
```
ppatierno, line 105:
```suggestion
            LOGGER.error("Interrupted while waiting for Vert.x verticle to undeploy");
```

PR #8947, opened 2023-08-03
Add model warnings about unknown or deprecated fields to Kafka Node Pools

scholzj:
### Type of change

- Enhancement / new feature

### Description

For most of our custom resources, when the user uses a deprecated or unknown field in its specification, we usually log it as a warning and add it as warning condition to the custom resource. This Pr adds this feature also for KafkaNodePool resources. The warnings are logged and added to the status of the KafkaNodePool and Kafka (since the NodePools are basically part of the Kafka CR) resources.

To make the warnings a bit more clear, this PR also adds an option to include the resource name in them. This is useful for KafkaConnectors or KafkaNodePools for which the warnings would be logged under Kafka / KafkaConnect reconciliation and it would not be obvious to which resource they apply.

### Checklist

- [x] Write tests

### DIFF
--- a/operator-common/src/main/java/io/strimzi/operator/common/operator/resource/StatusUtils.java
+++ b/operator-common/src/main/java/io/strimzi/operator/common/operator/resource/StatusUtils.java
@@ -244,25 +244,22 @@ public class StatusUtils {
      * Validate the Custom Resource. This should log at the WARN level (rather than throwing) if the resource can safely
      * be reconciled (e.g. it merely using deprecated API).
      *
-     * @param <T>                   Custom Resource type
-     * @param <P>                   Custom Resource spec type
-     * @param <S>                   Custom Resource status type
+     * @param <T>               Custom Resource type
+     * @param <P>               Custom Resource spec type
+     * @param <S>               Custom Resource status type
      *
-     * @param reconciliation        The reconciliation
-     * @param resource              The custom resource
-     * @param printResourceName     Flag indicating whether the warning messages should contain the resource
-     *                              identification or not. This is useful when the message is logged as a part of
-     *                              reconciliation of another resource.
-     *
-     * @return  Set of conditions
+     * @param reconciliation    The reconciliation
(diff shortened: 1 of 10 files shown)


operator-common/src/main/java/io/strimzi/operator/common/operator/resource/StatusUtils.java
```diff
@@ -244,22 +244,25 @@ public static Condition getPausedCondition() {
...
+     *                              reconciliation of another resource.
      *
-     * @throws InvalidResourceException if the resource cannot be safely reconciled.
+     * @return  Set of conditions
      *
-     * @return set of conditions
+     * @throws InvalidResourceException if the resource cannot be safely reconciled.
      */
-    public static <T extends CustomResource<P, S>, P extends Spec, S extends Status> Set<Condition> validate(Reconciliation reconciliation, T resource) {
+    public static <T extends CustomResource<P, S>, P extends Spec, S extends Status> Set<Condition> validate(Reconciliation reconciliation, T resource, boolean printResourceName) {
```
ppatierno, line 261: I was wondering if we really need an extra flag to decide to print (or not) the resource name.
When "not needed" it would be just redundant and obvious but does it really hurt? In all other case it would be useful (i.e. KafkaConnector, KafkaNodePool). Having this flag just makes the code more "complicated" imho.

scholzj: I can remove it if you think. But keep in mind that all other cases are the 8 other custom resources basically.

scholzj: Should be done now.

ppatierno: Thanks!

ppatierno (approved):
LGTM!

PR #8958, opened 2023-08-07
Validate number of JBOD disks in KafkaNodePools in KRaft mode

### LINKED ISSUE #8955: [Bug]: Cluster operator allows multiple jbod disks for Kraft with Nodepools
### Bug Description

When user tries to deploy Kraft Kafka with nodepool that contains JBOD storage config with 2 or more disks, the cluster starts rolling updated but the pod end-up in the crashloopbackoff. This is currently known limitation and the yaml should be validated on Operator side. JBOD with 1 disk works fine.

### Steps to reproduce

1. Deploy CO
2. Deploy Kafka with jbod configuration with 2 disks

### Expected behavior

Kafka will be in `NotReady` state as this configuration is not supported.

### Strimzi version

main

### Kubernetes version

OpenShift 4.13.6

### Installation method

YAML files

### Infrastructure

Openstack

```
---
kind: KafkaNodePool
metadata:
  name: broker
  labels:
    strimzi.io/cluster: horus
spec:
  replicas: 5
  roles:
    - broker
  storage:
    type: jbod
    volumes:
      - id: 0
        type: persistent-claim
        size: 100Gi
        deleteClaim: false
      - id: 1
        type: persistent-claim
        size: 100Gi
        deleteClaim: false
```

### Additional context

_No response_

scholzj:
### Type of change

- Bugfix

### Description

Apache Kafka in KRaft mode currently does not support JBOD storage (Strimzi's `type: jbod` itself is fine as long as it has only one disk). Right now, the validation for this is done only in the Kafka CR. But we force users to use KafkaNodePools for KRaft. So the actual storage used is int he KafkaNodePool resources. So the validation is currently not really working.

This PR removes the old validation of JBOD volume count for the Kafka CR which is not needed anymore. And it instead adds  a new validation for KafkaNodePools which causes an exception to be thrown when KRaft is used with more than one JBOD volume in any of the related node pools.

It also fixes some minor typos and indentation issues in the files which are already changed by this PR.

This should resolve #8955.

### Checklist

- [x] Write tests

### DIFF
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KRaftUtils.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KRaftUtils.java
@@ -5,9 +5,7 @@
 package io.strimzi.operator.cluster.model;
 
 import io.strimzi.api.kafka.model.EntityOperatorSpec;
-import io.strimzi.api.kafka.model.KafkaClusterSpec;
 import io.strimzi.api.kafka.model.KafkaSpec;
-import io.strimzi.api.kafka.model.storage.JbodStorage;
 
 import java.util.HashSet;
 import java.util.Set;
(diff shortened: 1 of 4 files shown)


ppatierno (approved):
LGTM

Frawless (approved):

PR #8959, opened 2023-08-08
[ST] Correct namespace handling for Namespace rbac

see-quick:
### Type of change

- Bugfix

### Description

This PR correct how we handle RBAC set to namespace. Some test cases won't work with RBAC set to NAMESPACE due to bad namespace handling. For instance, MirrorMaker2ST, MirrorMakerST and other suites will try to deploy resources into a not-created namespace (i.e., test-suite-namespace) and resulting in an error. 

### Checklist


systemtest/src/main/java/io/strimzi/systemtest/Constants.java
```diff
@@ -410,7 +410,7 @@ public interface Constants {
 
     // main namespace for Cluster Operator deployment
     String CO_NAMESPACE = "co-namespace";
-    String TEST_SUITE_NAMESPACE = "test-suite-namespace";
+    String TEST_SUITE_NAMESPACE = Environment.isNamespaceRbacScope() ? CO_NAMESPACE : "test-suite-namespace";
```
Frawless, line 413: I don't think this is a proper place for it. This line will change Constants file into a file where something is evaluated. Wouldn't be better to move the decision to Abstract or to Env?

see-quick: AbstractsST class is not feasible because we use such value outside of tests so I can move it to the Environment class if that's okay with you?

im-konge (approved):
LGTM, thanks :)

Frawless (approved):

PR #8960, opened 2023-08-08
Add image.imagePullPolicy to values.yaml

rishinair11:
### Type of change
- Documentation

### Description
The operator `Deployment` template contains the provision to customize the `imagePullPolicy`, but that's not mentioned in the `values.yaml`.
This PR makes it clearer.

### Checklist

- [x] Write tests
- [x] Update documentation
- [x] Update CHANGELOG.md

### DIFF
--- a/helm-charts/helm3/strimzi-kafka-operator/values.yaml
+++ b/helm-charts/helm3/strimzi-kafka-operator/values.yaml
@@ -19,6 +19,7 @@ image:
   tag: ""
   # imagePullSecrets:
   #   - name: secretname
+  # imagePullPolicy: IfNotPresent
 logVolume: co-config-volume
 logConfigMap: strimzi-cluster-operator
 logConfiguration: ""


scholzj (commented):
Thanks for the PR. First of all, you are modifying the wrong files. Any changes should be done  in the `./packaging/` folder. Second, I'm not entirely sure what exactly are you trying to achieve with this PR. The option is properly described in the README and is super simple. Why does it need to have a commented out example in the values file?

rishinair11:
@scholzj Hey, thanks for the reply. 
My bad, i’ll make changes in the correct file. 

Yep it’s in the README but sometimes users (like me 😅) may check the values.yaml directly as its faster. 

Let me know if that makes sense.

scholzj:
Well, the README is there for some purpose - to cover and explain the various options. I can probably live with this change if done right, but to be honest, it does not seem to make sense to me. If we follow this approach everywhere, every file and every example we have will be full of commented-out things that are not used. Reading the relevant documentation is the right thing to learn things, not from commented-out lines in the values file.

rishinair11:
Makes sense to me on second thought. Wasnt aware that the field was documented i. the README. 
Closing this PR.

PR #8978, opened 2023-08-11
Removed OpenTracing/Jaeger support

### LINKED ISSUE #7385: Remove support for OpenTracing
When the #7201 (about adding OpenTelemetry support) will be closed, we should remove the OpenTracing support.
The OpenTracing project is now archived and not supported anymore with new releases, bug fixes or CVEs.
The Strimzi operator already deprecates its usage since 0.31.0 release.
By using the future OpenTelemetry addition, users can still continue to use their Jaeger UI to get tracing information, because the operator will use a `jaeger` exporter by default to maitain ba

ppatierno:
### Type of change

- Refactoring

### Description

This PR fixes #7385 which is about removing the OpenTracing support on all the components (Kafka Connect, MM1, MM2 and configuration on the Bridge).
As described in the corresponding issue it was already deprecated in Strimzi 0.31.0 release and in version 0.33.0 we announced that it was going to be removed in a future release after June 2023.

### Checklist

- [x] Write tests
- [x] Update documentation
- [x] Update CHANGELOG.md


api/src/main/java/io/strimzi/api/kafka/model/tracing/JaegerTracing.java
scholzj, line 1: I don't think you can remove the type. That is backward incompatible change. You can just ignoreit.

scholzj (commented):
Should you also remove it from the 3rd party artifacts?

ppatierno:
> Should you also remove it from the 3rd party artifacts?

Damn right! Thanks!

api/src/main/java/io/strimzi/api/kafka/model/tracing/JaegerTracing.java
ppatierno: Do you mean that users should still be able to set `jaeger` as tracing type but we don't care internally? Is this what you mean?

scholzj: If you remove the `type: jaeger`, you will make all CRs which set it invalid -> that can cause problems. So I think the only thing you can do is basically in the CO code `if (type == jaeger) do-nothing` but I don't think you can remove it from the API (you can only add some docs to it that `type: jaeger` will be ignored by the operator).

ppatierno:
@scholzj I reverted back the `jaeger` type to be backward compatible with current `Kafka` CR(s) already in place but just a few things to highlight ...

* the `KafkaBridgeCrdIT` has to be reverted as well because it validates the possible tracing type(s) so even `jaeger` is still there even if it's not usable in reality.
* On the `Tracing.getType()` description I kept my change stating `Currently the only supported type is `opentelemetry` for OpenTelemetry tracing.` So even if you can still put `jaeger` as type (which is ignored), we say that only `opentelemetry` is possible.
* Because `JaegerTracing` is still there (even if ignored) and marked as `@Deprecated` we have the CRD describing it's deprecated while in reality it's not supported anymore.

Regarding the 3rd point, is there any way to make this better?

api/src/main/java/io/strimzi/api/kafka/model/tracing/Tracing.java
```diff
@@ -34,8 +34,7 @@ public abstract class Tracing implements UnknownPropertyPreserving, Serializable
     private Map<String, Object> additionalProperties;
 
     @Description("Type of the tracing used. " +
-            "Currently the only supported types are `jaeger` for OpenTracing (Jaeger) tracing and `opentelemetry` for OpenTelemetry tracing. " +
-            "The OpenTracing (Jaeger) tracing is deprecated")
+            "Currently the only supported type is `opentelemetry` for OpenTelemetry tracing.")
```
scholzj, line 37: I think maybe we could add something like we have for example here? https://github.com/strimzi/strimzi-kafka-operator/blob/a2bae6f2498f187253d189d61c8082a3ea6b8c8b/api/src/main/java/io/strimzi/api/kafka/model/KafkaSpec.java#L102

That `jaeger` will be ignored from 0.37.0 on?

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaConnectCluster.java
```diff
@@ -237,10 +236,7 @@ protected static <C extends KafkaConnectCluster> C fromSpec(Reconciliation recon
             result.configuration = config;
         }
         if (result.tracing != null)   {
-            if (JaegerTracing.TYPE_JAEGER.equals(result.tracing.getType())) {
-                config.setConfigOption("consumer.interceptor.classes", JaegerTracing.CONSUMER_INTERCEPTOR_CLASS_NAME);
-                config.setConfigOption("producer.interceptor.classes", JaegerTracing.PRODUCER_INTERCEPTOR_CLASS_NAME);
-            } else if (OpenTelemetryTracing.TYPE_OPENTELEMETRY.equals(result.tracing.getType())) {
```
scholzj, line 243: Should we keep this and throw a warning here that `type: jaeger` is not supported and will be ignored? Same for the other model classes as well.

cluster-operator/src/test/java/io/strimzi/operator/cluster/model/KafkaBridgeClusterTest.java
```diff
@@ -1234,12 +1233,6 @@ public void testProbeConfiguration()   {
         assertThat(cont.getReadinessProbe().getTimeoutSeconds(), is(32));
     }
 
-    @ParallelTest
-    public void testJaegerTracingConfiguration() {
-        testTracingConfiguration(JaegerTracing.TYPE_JAEGER);
-    }
-
-    @ParallelTest
```
scholzj, line 1242: Did you delete one of these `@ParallelTest` annotations by mistake?

cluster-operator/src/test/java/io/strimzi/operator/cluster/model/KafkaBridgeClusterTest.java
```diff
@@ -1248,11 +1241,6 @@ private void testTracingConfiguration(String type) {
 
         KafkaBridgeBuilder builder = new KafkaBridgeBuilder(this.resource);
         switch (type) {
-            case JaegerTracing.TYPE_JAEGER:
```
scholzj, line 1251: Would it make sense to merge this method directly into `testOpenTelemetryTracingConfiguration`? I guess it is now used only form this one place, or?

cluster-operator/src/test/java/io/strimzi/operator/cluster/model/KafkaConnectClusterTest.java
```diff
@@ -1469,12 +1461,6 @@ public void testOpenTelemetryTracing() {
     public void testTracing(String type, String consumerInterceptor, String producerInterceptor) {
         KafkaConnectBuilder builder = new KafkaConnectBuilder(this.resource);
         switch (type) {
-            case JaegerTracing.TYPE_JAEGER:
```
scholzj, line 1472: Same as above - isn't this method used only from one place? MAybe we can get rid of it? It seems to be also in the other tests.

scholzj (commented):
I left bunch of nits. But it looks good to me overall.

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaConnectCluster.java
ppatierno: Wdym by "throw a warning"? Thorwing an exception and failing the riconcile or just logging a warning? My feeling is the second one which is more likely something like "ignoring" but asking just to confirm.

scholzj: Log a warning, not throw ... sorry.

ppatierno: I am going to have something like this ...

```java
if (JaegerTracing.TYPE_JAEGER.equals(result.tracing.getType())) {
                LOGGER.warnCr(reconciliation, "Tracing type \"{}\" is not supported anymore and will be ignored", JaegerTracing.TYPE_JAEGER);
            }
```

scholzj: Yeah, that looks fine. I guess you can hardcode the type into the log message maybe as it is always the same ... but it is fine even as it is 👍

cluster-operator/src/test/java/io/strimzi/operator/cluster/model/KafkaBridgeClusterTest.java
ppatierno: :-o .... good catch! Yes that was a mistake!

cluster-operator/src/test/java/io/strimzi/operator/cluster/model/KafkaBridgeClusterTest.java
ppatierno: Yeah I was leaving there just in casa we will have a new tracing in the future ... but who knows :-)
Let me merge it.

scholzj (approved):
LGTM. Will the docs be removed in a separate PR?

ppatierno:
No I was planning to update it here while regression was running. This is why it's a draft right now.

im-konge (approved):
LGTM, I will remove the `TracingAbstractST` and refactor the clients a bit in another PR. Thanks :)

ppatierno:
@im-konge the two regression tests failing don't seem to be related to my changes. I ran them locally and they worked fine. Any thoughts?

im-konge:
I have no idea TBH. I can check it tomorrow, or we can try to rerun it

ppatierno:
@strimzi-ci run tests --install-type=bundle --ocp-version=4.13 --testcase=ConnectST#testConnectorTaskAutoRestart,ConnectBuilderST#testBuildWithJarTgzAndZip

documentation/modules/tracing/proc-configuring-tracers-kafka-clients.adoc
```diff
@@ -8,7 +8,7 @@
 [role="_abstract"]
 Initialize a tracer, then instrument your client applications for distributed tracing.
 You can instrument Kafka producer and consumer clients, and Kafka Streams API applications.
-You can initialize a tracer for OpenTracing or OpenTelemetry.
+You can initialize a tracer for OpenTelemetry.
```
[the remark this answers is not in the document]
PaulRMellor: shortens the abstract a bit :point_up:

PaulRMellor (approved):
Looks good. Couple of doc suggestions.

im-konge:
@strimzi-ci run tests --install-type=bundle --ocp-version=4.13 --testcase=ConnectST#testConnectorTaskAutoRestart,ConnectBuilderST#testBuildWithJarTgzAndZip --build-images

ppatierno:
@strimzi-ci run tests --install-type=bundle --ocp-version=4.13 --testcase=ConnectST#testConnectorTaskAutoRestart,ConnectBuilderST#testBuildWithJarTgzAndZip --build-images

PR #8991, opened 2023-08-15
Do not set `broker.id` in KRaft mode

scholzj:
### Type of change

- Bugfix

### Description

Currently, we set the `broker.id` when running in KRaft mode. But in KRaft, `broker.id` is ignored. This PR updates the configuration generator to generate `broker.id` only in ZooKeeper mode and in KRaft mode use `node.id` only.

Note: In ZooKeeper mode, the `node.id` defaults to `broker.id`. So we have to set it since it otherwise causes infinite rolling updates.

### Checklist


ppatierno (approved):
LGTM

PR #8996, opened 2023-08-16
Do not set `inter.broker.protocol.version` in KRaft clusters

scholzj:
### Type of change

- Bugfix

### Description

The `inter.broker.protocol.version` and `log-message.format.version` fields have no meaning anymore in KRaft mode. And in particular `inter.broker.protocol.version` causes many different warnings to be logged by the brokers when it is set in KRaft mode. This PR removes these fields from being set in KRaft in the Pod annotations and in the Kafka node configurations. (It does not change anything in ZooKeeper mode.)

After consideration, I decided to not create any unit test for not setting some value that is not used. If anyone thinks there should be one, please let me know ;-).

### Checklist


ppatierno (approved):
LGTM

PR #8999, opened 2023-08-17
kraft-controllers-rolling branch: Add KafkaControllerConfigurationDiff

katheris:
Change the return type from the KafkaAgentClient's getNodeConfiguration().

### Type of change

_Select the type of your PR_

- Enhancement / new feature

### Description

* Add KafkaControllerConfigurationDiff
* Change the return type from the KafkaAgentClient's getNodeConfiguration().

### Checklist

- [ ] Write tests
- [ ] Update documentation
- [ ] Update CHANGELOG.md


cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/resource/KafkaControllerConfigurationDiff.java
```diff
@@ -0,0 +1,171 @@
...
+import io.strimzi.operator.common.model.OrderedProperties;
+import io.strimzi.operator.common.operator.resource.AbstractJsonDiff;
+
+import java.util.List;
+import java.util.Map;
+
+public class KafkaControllerConfigurationDiff extends AbstractJsonDiff {
+    final boolean configsHaveChanged;
+
+    protected KafkaControllerConfigurationDiff(Map<String, String> controllerConfig, String desired) {
```
ppatierno, line 19: when I was thinking about how to pass the `controllerConfig` I was wonder how much was difficult to build a `Config` instance as we have in the current `KafkaBrokerConfigurationDiff` class. Of course, it's working there because `Config` (coming from Apache Kafka) is returned by the description via the admin client. The idea I had was to build a `Config` instance from what I have got via the Kafka Agent as a JSON. This would have allowed to not change this diffing class when the admin client API for controllers work (so returning a `Config` instance). Of course, building a `Config` instance means adding `ConfigEntry` which could be difficult to create based on how many info we need to know (readonly, default, ...) but I was thinking about referring to the model for that.
It was just an idea I had to have this diffing code to be reusable as much as possible but didn't go through enough to say ... it's crazy and not feasible. Maybe something you could think about.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/resource/KafkaControllerConfigurationDiff.java
```diff
@@ -0,0 +1,171 @@
...
+                "ssl.keystore.type",
+                "ssl.principal.mapping.rules",
+                "ssl.protocol",
+                "ssl.provider",
+                "ssl.secure.random.implementation",
+                "ssl.trustmanager.algorithm",
+                "ssl.truststore.certificates",
+                "ssl.truststore.location",
+                "ssl.truststore.password",
+                "ssl.truststore.type",
```
ppatierno, line 163: take mostly all the `ssl` ones are defined as "ignorable" in the current `KafkaBrokerConfigurationDiff` class, maybe we should take this into account here as well?

katheris: The comment in `KafkaBrokerConfigurationDiff` seems to imply that these are ignorable because they contain some sort of placeholder values, but it doesn't give any more details so it wasn't clear to us whether this applies to the controller config as well. Our plan was to finish implementing the poc and then do some testing of changing the ssl values with and without them included in this list to verify if we need them here or not.

ppatierno: FYI, most of the placeholder should be here https://github.com/strimzi/strimzi-kafka-operator/blob/main/cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaBrokerConfigurationBuilder.java

katheris:
This PR can be merged once it is reviewed, we won't push any more changes to it

ppatierno (approved):
LGTM

Issue #9002, opened 2023-08-18
[Enhancement]: Implement Kafka reconcile and kafka versions field as described in proposal


samuel-hawker:
### Related problem

Implement proposal https://github.com/strimzi/proposals/pull/84 as described

### Suggested solution

Implementation should follow proposal and be properly tested to ensure no regressions

samuel-hawker:
Picking this up as I drove the proposal.
Please message me on here or in slack if you want to get involved!

PR #9003, opened 2023-08-18
feat: Kafka versions in status 

samuel-hawker:
Add new status fields
status.kafkaVersion
status.operatorLastSuccessfulVersion

Add project.version to resources
Edit pre-existing unit tests and add some more.
Note that operator version is reported as x.y.z-SNAPSHOT during development cycle

Contributes to: https://github.com/strimzi/strimzi-kafka-operator/issues/9002

Signed-off-by: Samuel Hawker <sam.b.hawker@gmail.com>
Signed-off-by: Samuel Hawker <samuel.hawker@ibm.com>

### Type of change

_Select the type of your PR_

- Bugfix
- Enhancement / new feature
- Refactoring
- Documentation

### Description

### Checklist

- [ ] Write tests
- [ ] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/cluster-operator/pom.xml
+++ b/cluster-operator/pom.xml
@@ -340,6 +340,11 @@
                 <directory>src/main/resources</directory>
             </resource>
             <resource>
+                <!-- separate folder for filtered resources so that non-filtered files don't have their ${} replaced -->
+                <directory>src/main/resources-filtered</directory>
+                <filtering>true</filtering>
+            </resource>
+            <resource>
                 <directory>..</directory>
                 <includes>
                     <include>kafka-versions.yaml</include>
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaAssemblyOperator.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaAssemblyOperator.java
@@ -76,6 +76,9 @@ public class KafkaAssemblyOperator extends AbstractAssemblyOperator<KubernetesCl
     private static final ReconciliationLogger LOGGER = ReconciliationLogger.create(KafkaAssemblyOperator.class.getName());
 
     private static final Properties PROPERTIES = new Properties();
+    /**
+     * version of the operator, project.version in the pom.xml
+     */
     public static final String OPERATOR_VERSION;
 
     static {
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaAssemblyOperator.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaAssemblyOperator.java
@@ -80,7 +80,7 @@ public class KafkaAssemblyOperator extends AbstractAssemblyOperator<KubernetesCl
     /**
      * version of the operator, project.version in the pom.xml
      */
-    public static final String OPERATOR_VERSION;
+    /* test */ static final String OPERATOR_VERSION;
 
     static {
         InputStream propertiesFile = KafkaAssemblyOperator.class.getResourceAsStream("/.properties");
@@ -159,8 +159,7 @@ public class KafkaAssemblyOperator extends AbstractAssemblyOperator<KubernetesCl
                 status.setClusterId(kafkaAssembly.getStatus().getClusterId());
             }
 
-            if (status.getOperatorLastSuccessfulVersion() == null
-                    && kafkaAssembly.getStatus() != null
+            if (kafkaAssembly.getStatus() != null
                     && kafkaAssembly.getStatus().getOperatorLastSuccessfulVersion() != null
             )  {
                 // If not set in the status prepared by reconciliation but set in the status previously, we copy the
@@ -168,7 +167,6 @@ public class KafkaAssemblyOperator extends AbstractAssemblyOperator<KubernetesCl
                 // reason before setting the cluster ID
                 status.setOperatorLastSuccessfulVersion(kafkaAssembly.getStatus().getOperatorLastSuccessfulVersion());
             }
-
 
             if (status.getKafkaVersion() == null
                     && kafkaAssembly.getStatus() != null
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaAssemblyOperator.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaAssemblyOperator.java
@@ -55,10 +55,12 @@ import io.vertx.core.Future;
 import io.vertx.core.Promise;
 import io.vertx.core.Vertx;
 
+import java.io.IOException;
 import java.time.Clock;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Properties;
 import java.util.Set;
 import java.util.stream.IntStream;
 
@@ -72,6 +74,18 @@ import java.util.stream.IntStream;
  */
 public class KafkaAssemblyOperator extends AbstractAssemblyOperator<KubernetesClient, Kafka, KafkaList, Resource<Kafka>, KafkaSpec, KafkaStatus> {
     private static final ReconciliationLogger LOGGER = ReconciliationLogger.create(KafkaAssemblyOperator.class.getName());
+
+    private static final Properties PROPERTIES = new Properties();
+    public static final String OPERATOR_VERSION;
+
+    static {
+        try {
+            PROPERTIES.load(KafkaAssemblyOperator.class.getResourceAsStream("/.properties"));
+            OPERATOR_VERSION = PROPERTIES.getProperty("version");
+        } catch (IOException e) {
+            throw new RuntimeException(e);
+        }
+    }
 
     /* test */ final ClusterOperatorConfig config;
     /* test */ final ResourceOperatorSupplier supplier;
@@ -131,6 +145,27 @@ public class KafkaAssemblyOperator extends AbstractAssemblyOperator<KubernetesCl
                 status.setClusterId(kafkaAssembly.getStatus().getClusterId());
             }
 
+            if (status.getOperatorLastSuccessfulVersion() == null
+                    && kafkaAssembly.getStatus() != null
+                    && kafkaAssembly.getStatus().getOperatorLastSuccessfulVersion() != null
+            )  {
+                // If not set in the status prepared by reconciliation but set in the status previously, we copy the
+                // operatorLastSuccessfulVersion version into the new status. This is useful for example when the reconciliation fails for some
+                // reason before setting the cluster ID
+                status.setOperatorLastSuccessfulVersion(kafkaAssembly.getStatus().getOperatorLastSuccessfulVersion());
+            }
+
+
+            if (status.getKafkaVersion() == null
+                    && kafkaAssembly.getStatus() != null
+                    && kafkaAssembly.getStatus().getKafkaVersion() != null
+            )  {
+                // If not set in the status prepared by reconciliation but set in the status previously, we copy the
+                // kafka version into the new status. This is useful for example when the reconciliation fails for some
+                // reason before setting the cluster ID
+                status.setKafkaVersion(kafkaAssembly.getStatus().getKafkaVersion());
+            }
+
             if (reconcileResult.succeeded())    {
                 condition = new ConditionBuilder()
                         .withLastTransitionTime(StatusUtils.iso8601(clock.instant()))
@@ -138,6 +173,8 @@ public class KafkaAssemblyOperator extends AbstractAssemblyOperator<KubernetesCl
                         .withStatus("True")
                         .build();
 
+                // successful reconcile, write operator version to successful reconcile field
+                status.setOperatorLastSuccessfulVersion(OPERATOR_VERSION);
                 status.addCondition(condition);
                 createOrUpdatePromise.complete(status);
             } else {
@@ -713,6 +750,14 @@ public class KafkaAssemblyOperator extends AbstractAssemblyOperator<KubernetesCl
         // We copy the cluster ID if set
         if (kafka.getStatus() != null && kafka.getStatus().getClusterId() != null)  {
             status.setClusterId(kafka.getStatus().getClusterId());
+        }
+
+        // We copy the versions if set
+        if (kafka.getStatus() != null && kafka.getStatus().getOperatorLastSuccessfulVersion() != null)  {
+            status.setOperatorLastSuccessfulVersion(kafka.getStatus().getOperatorLastSuccessfulVersion());
+        }
+        if (kafka.getStatus() != null && kafka.getStatus().getKafkaVersion() != null)  {
+            status.setKafkaVersion(kafka.getStatus().getKafkaVersion());
         }
 
         return status;
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaAssemblyOperator.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaAssemblyOperator.java
@@ -56,6 +56,7 @@ import io.vertx.core.Promise;
 import io.vertx.core.Vertx;
 
 import java.io.IOException;
+import java.io.InputStream;
 import java.time.Clock;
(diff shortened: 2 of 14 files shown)


samuel-hawker:
Manual testing suggests this works as intended.

Would anyone have any ideas of potential gaps in unit testing I may have missed.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaAssemblyOperator.java
```diff
@@ -131,13 +159,36 @@ public Future<KafkaStatus> createOrUpdate(Reconciliation reconciliation, Kafka k
...
+                    && kafkaAssembly.getStatus() != null
+                    && kafkaAssembly.getStatus().getOperatorLastSuccessfulVersion() != null
+            )  {
+                // If not set in the status prepared by reconciliation but set in the status previously, we copy the
+                // operatorLastSuccessfulVersion version into the new status. This is useful for example when the reconciliation fails for some
+                // reason before setting the cluster ID
+                status.setOperatorLastSuccessfulVersion(kafkaAssembly.getStatus().getOperatorLastSuccessfulVersion());
+            }
+
+
```
scholzj, line 172: Double empty line not needed?

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaAssemblyOperator.java
```diff
@@ -131,13 +159,36 @@ public Future<KafkaStatus> createOrUpdate(Reconciliation reconciliation, Kafka k
                 status.setClusterId(kafkaAssembly.getStatus().getClusterId());
             }
 
+            if (status.getOperatorLastSuccessfulVersion() == null
```
scholzj, line 162: Can `status.getOperatorLastSuccessfulVersion() == null` really be false at any point? You seem to set it only at line 191 which is behind this point.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaAssemblyOperator.java
```diff
@@ -715,6 +766,14 @@ protected KafkaStatus createStatus(Kafka kafka) {
             status.setClusterId(kafka.getStatus().getClusterId());
         }
 
+        // We copy the versions if set
+        if (kafka.getStatus() != null && kafka.getStatus().getOperatorLastSuccessfulVersion() != null)  {
+            status.setOperatorLastSuccessfulVersion(kafka.getStatus().getOperatorLastSuccessfulVersion());
+        }
+        if (kafka.getStatus() != null && kafka.getStatus().getKafkaVersion() != null)  {
+            status.setKafkaVersion(kafka.getStatus().getKafkaVersion());
+        }
```
scholzj, line 775: If you copy the versions first here ... why is the code on line 162-182 needed? It seems its one or the other.

cluster-operator/src/test/java/io/strimzi/operator/cluster/operator/assembly/KafkaAssemblyOperatorTest.java
```diff
@@ -766,6 +769,18 @@ public void testUpdateClusterNoop(Params params, VertxTestContext context) {
...
+    @MethodSource("data")
+    public void testUpdateClusterKafkaVersion(Params params, VertxTestContext context) {
+        setFields(params);
+        Kafka kafkaAssembly = getKafkaAssembly("bar");
+        kafkaAssembly.setStatus(new KafkaStatusBuilder()
+                .withKafkaVersion("3.4.0")
+                .withOperatorLastSuccessfulVersion("0.30.0")
+                .build());
+        updateCluster(context, getKafkaAssembly("bar"), kafkaAssembly);
+    }
```
scholzj, line 782: Does this really need to be in the parametrized tests? This part seems to have nothing with the paramettrization. So maybe it would fit better in the non-paramterized tests?

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaAssemblyOperator.java
```diff
@@ -73,6 +76,31 @@
 public class KafkaAssemblyOperator extends AbstractAssemblyOperator<KubernetesClient, Kafka, KafkaList, Resource<Kafka>, KafkaSpec, KafkaStatus> {
     private static final ReconciliationLogger LOGGER = ReconciliationLogger.create(KafkaAssemblyOperator.class.getName());
 
+    private static final Properties PROPERTIES = new Properties();
+    /**
+     * version of the operator, project.version in the pom.xml
+     */
+    public static final String OPERATOR_VERSION;
```
scholzj, line 83: Does it need to be public? Can it be private instead or at least be only `/* test */`?

cluster-operator/src/test/java/io/strimzi/operator/cluster/operator/assembly/KafkaReconcilerStatusTest.java
```diff
@@ -181,6 +184,87 @@ public void testKafkaReconcilerStatus(VertxTestContext context) {
...
+                .editOrNewSpec()
+                .editOrNewKafka()
+                .withReplicas(1)
+                .endKafka()
+                .endSpec()
+                .editOrNewStatus()
+                // supportedVersions is an ordered set, get the first element for the lowest kafka version
+                .withKafkaVersion(VERSIONS.supportedVersions().iterator().next())
+                .endStatus()
+                .build();
```
scholzj, line 199: If you indent along the blocks of the custom resource, it would be more readable.

cluster-operator/src/test/java/io/strimzi/operator/cluster/operator/assembly/KafkaReconcilerStatusTest.java
```diff
@@ -181,6 +184,87 @@ public void testKafkaReconcilerStatus(VertxTestContext context) {
...
+    public void testKafkaReconcilerStatusUpdateVersion(VertxTestContext context) {
+        Kafka kafka = new KafkaBuilder(KAFKA)
+                .editOrNewSpec()
+                .editOrNewKafka()
+                .withReplicas(1)
+                .endKafka()
+                .endSpec()
+                .editOrNewStatus()
+                // supportedVersions is an ordered set, get the first element for the lowest kafka version
+                .withKafkaVersion(VERSIONS.supportedVersions().iterator().next())
```
scholzj, line 197: Would it be more readable if you use something like `KafkaVersionTestUtils.PREVIOUS_KAFKA_VERSION`?

cluster-operator/src/test/java/io/strimzi/operator/cluster/operator/assembly/KafkaReconcilerStatusTest.java
```diff
@@ -181,6 +184,87 @@ public void testKafkaReconcilerStatus(VertxTestContext context) {
...
+        // supportedVersions is an ordered set, get the first element for the lowest kafka version
+        String lowestSupportedVersion = VERSIONS.supportedVersions().iterator().next();
+        Kafka kafka = new KafkaBuilder(KAFKA)
+                .editOrNewSpec()
+                .editOrNewKafka()
+                .withVersion(lowestSupportedVersion)
+                .withReplicas(3)
+                .endKafka()
+                .endSpec()
+                .build();
```
scholzj, line 239: Same comments as above ... would it be better to use the constant and indent the builder?

cluster-operator/src/test/java/io/strimzi/operator/cluster/operator/assembly/KafkaReconcilerUpgradeDowngradeTest.java
```diff
@@ -154,7 +155,7 @@ public void testWithAllVersionsInCR(VertxTestContext context) {
         KafkaStatus status = new KafkaStatus();
 
         Checkpoint async = context.checkpoint();
-        reconciler.reconcile(status, Clock.systemUTC()).onComplete(context.succeeding(i -> context.verify(() -> {
+        reconciler.reconcile(status, Clock.systemUTC()).onComplete(context.succeeding(v -> context.verify(() -> {
```
scholzj, line 158: Any reasons to rename it? The `i` suggests that it is _**i**gnored_

cluster-operator/src/test/java/io/strimzi/operator/cluster/operator/assembly/KafkaReconcilerUpgradeDowngradeTest.java
```diff
@@ -122,10 +122,11 @@ public static void afterAll()    {
 
     @Test
     public void testWithAllVersionsInCR(VertxTestContext context) {
+        String kafkaVersion = VERSIONS.defaultVersion().version();
         Kafka kafka = new KafkaBuilder(KAFKA)
                 .editOrNewSpec()
                     .editOrNewKafka()
-                        .withVersion(VERSIONS.defaultVersion().version())
+                        .withVersion(kafkaVersion)
```
scholzj, line 129: Any reason for this change? It seems to have no meaning.

cluster-operator/src/test/java/io/strimzi/operator/cluster/operator/assembly/KafkaReconcilerStatusTest.java
```diff
@@ -181,6 +184,87 @@ public void testKafkaReconcilerStatus(VertxTestContext context) {
         }));
     }
 
+    @Test
+    public void testKafkaReconcilerStatusUpdateVersion(VertxTestContext context) {
```
scholzj, line 188: Should we have a test here that failed reconciliation does not change the status?

cluster-operator/src/test/java/io/strimzi/operator/cluster/operator/assembly/KafkaStatusTest.java
scholzj, line 1: Same as for the Kafka version ... should we somewhere have a test that reconciliation failure uses the previous status?

scholzj (commented):
I left some comments. Also, would it be worth a CHANGELOG.md entry?

cluster-operator/pom.xml
```diff
@@ -334,6 +334,11 @@
             <resource>
                 <directory>src/main/resources</directory>
             </resource>
+            <resource>
+                <!-- separate folder for filtered resources so that non-filtered files don't have their ${} replaced -->
+                <directory>src/main/resources-filtered</directory>
+                <filtering>true</filtering>
+            </resource>
```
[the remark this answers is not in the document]
samuel-hawker: So I originally used this, but for local development `getPackage().getImplementationVersion()` always returns null for unit tests, I read into this and it appears that unless you're running the actual JARs, the information you need for this info is not actually populated, hence why it appears in the main log lines.

Now I can just set the version like this, but then this means from a unit test point of view, unless we do some clever mocking for setting the versions field, or falling back to a default if its null, it will always be setting the key as null which will make for some confusing tests.
Using these templated resources gets around this as the resources will be templated prior to the unit test phase.

If you have any advice/thoughts on how to get `Main.class.getPackage().getImplementationVersion()` to work at unit test level I would absolutely implement the version as this, but I am unsure whether it is wise to do this given it will always be null for unit tests.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaAssemblyOperator.java
samuel-hawker: Apologies, I originally edited this file and didn't revert all changes. Will fix now.

cluster-operator/pom.xml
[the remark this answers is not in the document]
scholzj: Yeah, I guess you are right. That call gets it from the JAR itself. That seems unfortunate :-/.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaAssemblyOperator.java
samuel-hawker: I copied the precedent set by clusterID field (slightly above my changes)
I assume it has been set like this to maintain the behaviour so that it works if someone calls `reconcile` or `createOrUpdate` as if you call `createOrUpdate` directly I belive it skips `createStatus`.

I am happy to keep it to just the one place to do the check, but if that's the case I believe `createOrUpdate` may be the best place, thoughts?

scholzj: I would be careful about touching the cluster ID because it might break the cluster. But I'm not sure why would it need to be done twice.

samuel-hawker:
Forgive my ignorance, but confused why the CICD is erroring with:
```
 [INFO] -------------------------------------------------------------
Error:  COMPILATION ERROR : 
[INFO] -------------------------------------------------------------
Error:  /home/runner/work/strimzi-kafka-operator/strimzi-kafka-operator/cluster-operator/src/test/java/io/strimzi/operator/cluster/operator/assembly/KafkaReconcilerStatusTest.java:[205,55] cannot find symbol
  symbol:   variable ClusterOperator
  location: class io.strimzi.operator.cluster.operator.assembly.KafkaReconcilerStatusTest
Error:  /home/runner/work/strimzi-kafka-operator/strimzi-kafka-operator/cluster-operator/src/test/java/io/strimzi/operator/cluster/operator/assembly/KafkaReconcilerStatusTest.java:[245,55] cannot find symbol
  symbol:   variable ClusterOperator
  location: class io.strimzi.operator.cluster.operator.assembly.KafkaReconcilerStatusTest
Error:  /home/runner/work/strimzi-kafka-operator/strimzi-kafka-operator/cluster-operator/src/test/java/io/strimzi/operator/cluster/operator/assembly/KafkaReconcilerStatusTest.java:[283,55] cannot find symbol
  symbol:   variable ClusterOperator
  location: class io.strimzi.operator.cluster.operator.assembly.KafkaReconcilerStatusTest
[INFO] 3 errors 
```

This all compiles fine locally... any ideas what is going on here?

scholzj:
@samuel-hawker Is it rebased to the latest main branch? The CI runs against the code as it would be when merged. So if not rebased, it might build something slightly different from the branch itself.

samuel-hawker:
Thanks for the tip Jakub, as expected you were right! Rebased and fixed.

cluster-operator/src/test/java/io/strimzi/operator/cluster/operator/assembly/KafkaUpgradeDowngradeMockTest.java
```diff
@@ -205,21 +211,29 @@ public void testUpgradeWithoutMessageAndProtocolVersions(VertxTestContext contex
         initialize(initialKafka)
                 .onComplete(context.succeeding(v -> {
                     context.verify(() -> {
+                        KafkaStatus status = Crds.kafkaOperation(client).inNamespace(NAMESPACE).withName(CLUSTER_NAME).get()
+                                .getStatus();
+                        assertThat(status.getKafkaVersion(), is(KafkaVersionTestUtils.PREVIOUS_KAFKA_VERSION));
+                        assertThat(status.getOperatorLastSuccessfulVersion(), is(KafkaAssemblyOperator.OPERATOR_VERSION));
```
scholzj, line 217: You seem to repeat this code in every test. Should it have some method similar to `assertVersionsInStrimziPodSet`?

scholzj (approved):
One nit to the test. But looks good to me otherwise. I will trigger the system tests as well.

@ppatierno I know you are on PTO and we do not really need your review in order to merge it. But do you want to have a look as well?

ppatierno (approved):
LGTM. Thanks for the PR!

cluster-operator/src/test/java/io/strimzi/operator/cluster/operator/assembly/KafkaUpgradeDowngradeMockTest.java
samuel-hawker: Sure this seems like a good idea. will update the PR.

PR #9005, opened 2023-08-20
Final refactoring of the Model and Operator classes split and import control

### LINKED ISSUE #8997: Enforce separation between the layers of Model and Operator classes
Strimzi consists of 3 different layers (at least in the case of CO - from the bottom layer to the top):
* The _API_ layer which defines the Kubernetes API (essentially the `api` module)
* The _Model_ layer consisting of classes in `io.strimzi.operator.cluster.model` which covers mostly the translation of the API to Kubernetes and Kafka (configs, certs, etc.) resources
* The _Operator_ layer which is responsible for executing the actions (managing resources, talking with Kafka etc.). These are the `io.strimzi.operator.cluster.operator` and `io.strimzi.operator.common.operator` packages.

We should enforce separation between these layers and make sure that the classes from a lower layer do not use directly the classes in the upper layer. For example:
* The API layer classes should not use any classes from the Model or Operator layers
* The Model layer classes should not use any classes from the Operator layer

The first point is easily achieved since the API layer has its own module `api` which does not depend on the other modules. But the Model and Operator classes share the same module. So there is currently nothing preventing the Model classes from using the Operator classes. In fact, there are several such situations. So we should:
* Fix the existing situations where this separation is broken
    * In many cases, this seems to be because classes that should be in Model are in Operators
    * In some cases, refactoring will be needed to fix the separation
* To ensure the separation, we should use the Checkstyle plugin and its import-control feature to prevent Model classes from importing any _Operator_ classes. This is already used for example in Apache Kafka itself.

(If needed, we can split this into additional JIRAs to minimize the scope of the individual PRs.)

scholzj:
### Type of change

- Refactoring

### Description

This PR does the final part of the implementation of #8997.

It does some remaining refactoring of moving classes between the Model and Operator package structures and also moves some classes which stand outside of these but should be in the Model packages.

It also setups the import control as part of the Checkstyle runs. It currently checks only the following parts of the code for allowed imports:
* `io.strimzi.operator.common.model` package
* `io.strimzi.operator.cluster.model` package
* `io.strimzi.operator.user.model` package

It makes sure that these packages import only the allowed classes such as Strimzi API, Fabric8 API Models, regular Java classes etc. It also allows imports of some Strimzi classes which need to be used but do not really fit into the model packages directly such as the reconciliation logger etc.

Checkstyle wants to use a separate file for the import control. But the Maven plugin does not support its proper configuration. So I had to add `strimziRootDirectory` property to all Strimzi `pom.xml` files to make sure the import control works in all the different build modes (using Maven, using Make, in a submodule only, etc.). It is not completely nice, but I spent a lot of time trying to find something better without success. If anyone has a better solution, please let me know.

All other packages are currently unchecked and can import whatever they want.

This resolves #8997.

### Checklist


.checkstyle/import-control.xml
```diff
@@ -0,0 +1,89 @@
+<!DOCTYPE import-control PUBLIC
+        "-//Checkstyle//DTD ImportControl Configuration 1.4//EN"
+        "https://checkstyle.org/dtds/import_control_1_4.dtd">
+
+<!-- We do not have rules for all classes. So we have to first allow all classes import everything
```
ppatierno, line 5: Just to be sure I have got it right, does it mean that if we need to import a new package in a class, we also need to add that "allow" rule here right?

ppatierno (approved):
LGTM. I left just one question.

.checkstyle/import-control.xml
scholzj: The rules apply currently only to the three packages (and its sub-packages) listed in the description. If you change anything else, you do not need to do anything.

If you change something in the model classes for which these rules apply and add some imports, it depends what imports these are. The rules already allow you to import other model classes, API classes, Java classes (such as Maps, Lists, etc.), and so on. So for most changes, you should not need anything. If you add import to something outside of the allowed packages, you should consider whether it is really correct and if yes, you might need to add the package or class here. In general, the thought behind this is mainly that the Model classes should not use any of the operators (resource operators and assembly operators). 

I tried to improve the description in the XML file to make it a bit more clear.

ppatierno: Makes sense thanks

PR #9011, opened 2023-08-21
docs(tls): final tls connections diagram

PaulRMellor:
**Documentation**

Adds the finalized diagram for secure TLS connections

### Checklist

- [ ] Write tests
- [x] Update documentation
- [ ] Update CHANGELOG.md


ppatierno:
I was looking at the picture referring to examples we provides. Usually the port 9092 is for internal connections but it's "plain" and not TLS enabled, while 9093 is still for internal connections and TLS enabled. I was wondering if it's better avoiding confusion with examples, by:

* using 9093 port as for internal clients (because it's TLS enabled in our example)
* using a different port for external clients (with TLS enabled)

PaulRMellor:
> I was looking at the picture referring to examples we provides. Usually the port 9092 is for internal connections but it's "plain" and not TLS enabled, while 9093 is still for internal connections and TLS enabled. I was wondering if it's better avoiding confusion with examples, by:
> 
> * using 9093 port as for internal clients (because it's TLS enabled in our example)
> * using a different port for external clients (with TLS enabled). Not sure which one to use. Usually we have 9094 using OpenShift "routes" but in the end an external client uses 433 and not 9094. Maybe 433 is an option?
> 
> @scholzj wdyt?

I actually considered that, but I thought it was better as it was with the new descriptions beneath the diagram. We go 9090, 9091, 9092 and higher in the descriptions (seems more consistent):

> Listeners for client connections (9092 or higher)
> For TLS-encrypted communication (through configuration of the listener), internal and external clients connect to Kafka brokers. External clients (producers and consumers) connect to the Kafka brokers through the advertised listener port.

https://strimzi.io/docs/operators/in-development/deploying#security-str

scholzj:
I don't know to be honest. The ports are configurable, so whatever number you use will be confusing. Maybe we should have some placeholder for the 9092 and 9093 ports? Or something like `9092+`?

PaulRMellor:
> I think that avoiding specific ports for clients taking into account, as Jakub said, they are configurable is better. I am fine with something like `9092+`. Wdyt @PaulRMellor ?

Thanks @ppatierno. I'll request the change shown in the comment above

ppatierno (approved):
LGTM. I am fine with the latest suggestion from Paul.

PR #9016, opened 2023-08-22
[systemtest] Remove the workaround for OpenTelemetry and TLS on OCP

im-konge:
### Type of change

- Enhancement

### Description

This PR removes the `*.tls.enabled: false` configuration from the `jaeger-instance.yaml`. Previously, we used these because of the issue with TLS connection to the Jaeger endpoint. Now, because we are using OTLP endpoint which is not TLS protected by default, we can remove the workaround.

### Checklist

### DIFF
--- a/systemtest/src/test/resources/tracing/jaeger-instance.yaml
+++ b/systemtest/src/test/resources/tracing/jaeger-instance.yaml
@@ -10,10 +10,6 @@ spec:
       log-level: debug
       query:
         base-path: /jaeger
-      # explicitly disable TLS, because tracing on some platforms automatically enables TLS, which for us is not supported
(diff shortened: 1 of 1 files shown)


im-konge:
@strimzi-ci run tests --profile=all --ocp-version=4.14 --testcase=OpenTelemetryST

ppatierno (approved):
@im-konge I think you can move this PR out of being a draft. We can remove the workaround of disabling TLS on Jaeger when on OpenShift, because the OTLP exporter we are now using the default connect to the OTLP endpoint which is not TLS protected by default (compared to the previous Jaeger one)

scholzj (approved):
LGTM if it works 🤷

PR #9055, opened 2023-08-31
docs(otel): fixes to otel doc from review

PaulRMellor:
**Documentation**

Couple of minor fixes from review of distributed tracing content

### Checklist

- [ ] Write tests
- [x] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/documentation/shared/attributes.adoc
+++ b/documentation/shared/attributes.adoc
@@ -112,6 +112,7 @@
 //distributed tracing versions and links
 :OpenTelemetryAlphaVersion: 1.19.0-alpha
 :OpenTelemetryVersion: 1.19.0
+:OpenTelemetryKafkaClient: 2.6
 
 :OpenTelemetryDocs: link:https://opentelemetry.io/docs/[OpenTelemetry documentation^]
 :JaegerDocs: link:https://www.jaegertracing.io/docs/[Jaeger documentation^]
--- a/documentation/shared/attributes.adoc
+++ b/documentation/shared/attributes.adoc
@@ -112,7 +112,7 @@
 //distributed tracing versions and links
 :OpenTelemetryAlphaVersion: 1.19.0-alpha
 :OpenTelemetryVersion: 1.19.0
-:OpenTelemetryKafkaClient: 2.6
+:OpenTelemetryKafkaClientVersion: 2.6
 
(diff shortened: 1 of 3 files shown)


documentation/shared/attributes.adoc
```diff
@@ -112,6 +112,7 @@
 //distributed tracing versions and links
 :OpenTelemetryAlphaVersion: 1.19.0-alpha
 :OpenTelemetryVersion: 1.19.0
+:OpenTelemetryKafkaClient: 2.6
```
PaulRMellor, line 115: we use this in the otel dependencies:

image

scholzj (approved):

documentation/shared/attributes.adoc
ppatierno: No we should stick with the 1.19 one. That is the OpenTelemetry version. The 2.6 is the Kafka version instead. I think they had one for the older 0.x Kafka and then moved to 2.x (which is compatible with all the new ones 3.x).
It's something I have on my to-do list if we should investigate to update the Otel Kafka instrumentation library with a newer Kafka version but it would mean more workload on us. I don't know about the impact and the time needed for it.

ppatierno (changes requested):
We should still use 1.19 as Otel Kafka instrumentation version. See my inline comment.

scholzj:
@ppatierno The problem is that without adding this option, the XML for the Maven dependencies is broken as it already uses the vairable: https://strimzi.io/docs/operators/latest/full/deploying.html#proc-configuring-tracers-kafka-clients-str

ppatierno:
@scholzj Ah! Ok then we need it but maybe OpenTelemetryKafkaClientVersion is a better name? Just to be clear it's the version of the Kafka client used in the OTel Kafka instrumentation library. The Version suffix is also used in the other two OTel version variables. @PaulRMellor wdyt?

PaulRMellor:
Thanks @ppatierno -- yes, sounds good to me.
I've updated in the latest commit [cfa67c6](https://github.com/strimzi/strimzi-kafka-operator/pull/9055/commits/cfa67c68a459bbe7dbb0492a5557818c1919412a)

ppatierno (approved):

PR #9057, opened 2023-08-31
docs(mm2): updates the section for running multiple instances

PaulRMellor:
**Documentation**

Updates the section describing how to configure Kafka Connect when running multiple instances.
The section now also mentions MM2, which can be configured in the same way for the same reason.
The section moves into the configuring chapter, with  x-references from the sections describing how to deploy Kafka Connect and MM2.

Also includes a few naming updates for MM2.

### Description

### Checklist

- [ ] Write tests
- [x] Update documentation
- [ ] Update CHANGELOG.md


documentation/modules/configuring/proc-config-kafka-connect-user-authorization.adoc
```diff
@@ -5,43 +5,17 @@
...
-When any type of authorization is being used in Kafka, a Kafka Connect user requires read/write access rights to the consumer group
-and the internal topics of Kafka Connect.
+Properties for the Kafka Connect cluster group ID and internal topics are configured by Strimzi by default. 
+Alternatively, you can define them explicitly in the `spec` of the `KafkaConnect` resource. 
+This is useful when xref:con-kafka-connect-mm2-multiple-instances-{context}[configuring Kafka Connect for multiple instances], as the values for the group ID and topics must differ when running multiple Kafka Connect instances.
 
-The properties for the consumer group and internal topics are automatically configured by Strimzi,
-or they can be specified explicitly in the `spec` of the `KafkaConnect` resource.
-
-.Example configuration properties in the `KafkaConnect` resource
```
PaulRMellor, line 16: I removed this as we link to the multiple instances section instead. 
Should we also expand this KC authorization config section to include MM2?

documentation/modules/configuring/con-config-kafka-connect-mm2-multiple-instances.adoc
scholzj, line 1: I'm not sure you can hide it that easily into one file:
* The configuration will be in different places. In Connect, it is `.spec.config` as you have it here. But for MM2, it will be in `config` section of the target cluster configuration. I.e. `.spec.clusters[].config` for the target cluster (bets to refer to it through the `connectCluster` field which is future-proof).
* MM2 also requires a bit more explanation of when it needs to be used. You need to use it when you have multiple MM2s which use the same `connectCluster`. So for example when you have one Kafka cluster used as a source cluster by two diffeent MM2s, you are fine. If you have two MM2s in active-active setup you are also fine as you have different `connectCluster` for each of them. Only when you have multiple MM2s with the same `connectCluster` you need this.

Given these references, I'm not sure whether two separate files one for each aren't better. But I leave that up to you. We just need to explain these differences.

documentation/modules/deploying/proc-deploy-kafka-mirror-maker.adoc
```diff
@@ -21,6 +22,9 @@ In this procedure, we use the following example files:
 * `examples/mirror-maker/kafka-mirror-maker.yaml`
 * `examples/mirror-maker/kafka-mirror-maker-2.yaml`
 
+IMPORTANT: If deploying MirrorMaker 2 clusters to run in parallel, each instance must use unique names for internal Kafka Connect topics. 
+To do this, xref:con-kafka-connect-mm2-multiple-instances-{context}[configure each MirrorMaker 2 instance to replace the defaults].
```
scholzj, line 26: As explain in the other comment. We should be more clear when does it need to be used with MM2.

scholzj (changes requested):

documentation/modules/configuring/con-config-kafka-connect-mm2-multiple-instances.adoc
PaulRMellor: Thanks. I'm trying to keep the content to a minimum where possible, but maybe too much here. I'll take another look for the next commit.

scholzj: To be clear, I don't have any major issue if you want to keep it in a single file if it makes things easier ... but the explanation would need to cover the differences.

scholzj (approved):
LGTM. Thanks.

ppatierno (approved):

PR #9071, opened 2023-09-05
Disable silent Kafka binaries download

ppatierno:
When building the images and it comes to download the Kafka binaries, the curl command is executed in "silent" mode without providing any feedback to the user about the download progress.
Due to network issues, slow ISP connection, CDN or archives Apache site being slow, the user could have the following output for several minutes ...

```shell
Fetching Kafka 3.4.0 binaries from: https://archive.apache.org/dist/kafka/3.4.0/kafka_2.13-3.4.0.tgz
Downloading from CDN: https://dlcdn.apache.org/kafka/3.4.0/kafka_2.13-3.4.0.tgz
CDN HTTP code: 404
Download from CDN failed. Retrying with archive: https://archive.apache.org/dist/kafka/3.4.0/kafka_2.13-3.4.0.tgz
(cursor blinking)
```

He doesn't have any feedback about the download progress.
For example, because I have a very fast ISP connection and I am used to see the download to be pretty fast, I stopped the process after several minutes because I thought there were some problems.
After some checks, the reality was that the archive Apache site was slow instead.

This PR removes the "silent" mode on the curl commands to show the download progress to the user.

### DIFF
--- a/docker-images/artifacts/build.sh
+++ b/docker-images/artifacts/build.sh
@@ -155,12 +155,12 @@ function download_kafka_binaries_from_cdn {
     local cdn_url="https://dlcdn.apache.org/${remote_path}"
     
     echo "Downloading from CDN: ${cdn_url}"
-    local cdn_code=$(curl -Ls -o "${local_path}" -w %{http_code} "${cdn_url}")
+    local cdn_code=$(curl -L -o "${local_path}" -w %{http_code} "${cdn_url}")
     echo "CDN HTTP code: ${cdn_code}"
 
     if [[ "${cdn_code}" != "200" ]]; then
         echo "Download from CDN failed. Retrying with archive: ${archive_url}"
-        local archive_code=$(curl -Ls -o "${local_path}" -w %{http_code} "${archive_url}")
+        local archive_code=$(curl -L -o "${local_path}" -w %{http_code} "${archive_url}")
         echo "Archive HTTP code: ${archive_code}"
     fi
 }


scholzj:
Wouldn't this screw up the logs in Azure with the information?

scholzj:
Also, isn't it capturing the output into the variable? Wouldn't that break it?

ppatierno:
I tried it locally with a simpler script (instead of running an entire make from scratch ... but I can try it as well) ... 

```shell
set -e
set -x

function download_kafka_binaries_from_cdn {
	local cdn_url=$1
	local local_path=$2
	echo "cdn_url: ${cdn_url}"
	echo "local_path: ${local_path}"

	local cdn_code=$(curl -L -o "${local_path}" -w %{http_code} "${cdn_url}")
	echo "CDN HTTP code: ${cdn_code}"
}

download_kafka_binaries_from_cdn $1 $2
```

it prints the following output while running (it updates the value in place) ...

```shell
% Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  9  101M    9 9583k    0     0  58910      0  0:30:09  0:02:46  0:27:23 46608
```

At the end the returned value is the http code printed ...

```shell
++ local cdn_code=200
++ echo 'CDN HTTP code: 200'
CDN HTTP code: 200
```

scholzj:
But the output will look different in CI where the text cannot be updated.
Each progress update will be printed as a separate line. So you can easily
get hundreds of progress line and run out of some capacity limitations. I
think you should maybe try to nakile it customizable as we do it for Maven.

scholzj:
> > you should maybe try to nakile it customizable as we do it for Maven.
> 
> I agree with that. Can you point me where we do it for Maven? I can't find the place :-(

Well, there are several parts ...
* The Makefile passes the `MVN_ARGS` option to every Maven call
* And then you can set it in the different places according to what you want. E.g. in CI we typically set the `-B` option (+ some others) to have the output suitable for non-interactive executions. E.g. https://github.com/strimzi/strimzi-kafka-operator/blob/main/.azure/templates/jobs/build/build_strimzi.yaml#L43

ppatierno:
Going to close this one due to GitHub outage not updating with my commits. Replaced by #9073

PR #9073, opened 2023-09-05
Customize silent curl Kafka binaries download

ppatierno:
When building the images and it comes to download the Kafka binaries, the curl command is executed in "silent" mode without providing any feedback to the user about the download progress.
Due to network issues, slow ISP connection, CDN or archives Apache site being slow, the user could have the following output for several minutes ...

```shell
Fetching Kafka 3.4.0 binaries from: https://archive.apache.org/dist/kafka/3.4.0/kafka_2.13-3.4.0.tgz
Downloading from CDN: https://dlcdn.apache.org/kafka/3.4.0/kafka_2.13-3.4.0.tgz
CDN HTTP code: 404
Download from CDN failed. Retrying with archive: https://archive.apache.org/dist/kafka/3.4.0/kafka_2.13-3.4.0.tgz
(cursor blinking)
```

He doesn't have any feedback about the download progress.
For example, because I have a very fast ISP connection and I am used to see the download to be pretty fast, I stopped the process after several minutes because I thought there were some problems.
After some checks, the reality was that the archive Apache site was slow instead.

This PR removes the "silent" mode on the curl commands to show the download progress to the user.

It replaces #9071 because GitHub had outage, my PR was not updated with commits I was pushing and wasting my time. New PR seems to work here.

### DIFF
--- a/.azure/templates/jobs/build/build_strimzi.yaml
+++ b/.azure/templates/jobs/build/build_strimzi.yaml
@@ -35,6 +35,7 @@ jobs:
         displayName: "Build Java code"
         env:
           MVN_ARGS: "-DskipTests -e -V -B"
+          CURL_ARGS: "-s"
 
       # Run different checks on the code
       - bash: "make spotbugs"
--- a/.azure/templates/jobs/build/build_strimzi.yaml
+++ b/.azure/templates/jobs/build/build_strimzi.yaml
@@ -35,6 +35,7 @@ jobs:
         displayName: "Build Java code"
         env:
           MVN_ARGS: "-DskipTests -e -V -B"
+          RUN_ON_CI: "true"
 
       # Run different checks on the code
       - bash: "make spotbugs"
--- a/.azure/templates/jobs/build/build_strimzi.yaml
+++ b/.azure/templates/jobs/build/build_strimzi.yaml
@@ -16,11 +16,11 @@ jobs:
     steps:
       # Get cached Maven repository
       - template: "../../steps/maven_cache.yaml"
-      - task: Cache@2
-        inputs:
-          key: '"kafka-binaries" | kafka-versions.yaml'
-          path: docker-images/artifacts/binaries/kafka/archives
-        displayName: Kafka binaries cache
+      #- task: Cache@2
+      #  inputs:
+      #    key: '"kafka-binaries" | kafka-versions.yaml'
+      #    path: docker-images/artifacts/binaries/kafka/archives
+      #  displayName: Kafka binaries cache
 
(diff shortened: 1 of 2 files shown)


ppatierno:
@scholzj before I re-enable the Kafka binaries cache, can you have a pass please? So If I need some changes I can easily test it.

.azure/templates/jobs/build/build_strimzi.yaml
```diff
@@ -35,6 +35,7 @@ jobs:
         displayName: "Build Java code"
         env:
           MVN_ARGS: "-DskipTests -e -V -B"
+          RUN_ON_CI: "true"
```
scholzj, line 38: TBH, I would avoid this and follow what we have for Maven with `CURL_PARAM` option that is set to `-s` rather than a flag like this. That gives us much more flexibility and consistency.

ppatierno: While I see your point, curl could be used for other objectives (in the future) and with different parameters. Using a specific env var would tight to a specific list.
Also right now the curl commands we are referring to are ...

`curl -L"${curl_silent}" -o "${local_path}" -w %{http_code} "${cdn_url}"`
and
`curl -L"${curl_silent}" -o "${local_path}" -w %{http_code} "${archive_url}"`

so even in this case different parameters with variables as well (see `-o`)
Which curl parameters would be involved in the `CURL_PARAM`, only `-Ls` in this case? does it make sense just for some of them?

scholzj: I guess `-L` has to always be there. So it is `-s` you need to pass.

scholzj: PS: `CURL_ARGS` probably makes more sense then `CURL_PARAM`. Not sure why I used the wrong one initiially.

scholzj: > While I see your point, curl could be used for other objectives (in the future) and with different parameters. Using a specific env var would tight to a specific list.

Well, this is the same with Maven as well. The `MVN_ARGS` has different values in different places. That is the flexibility it gives you that a single boolean flag cannot.

scholzj (commented):
I guess you still need to set the argument?

ppatierno:
> I guess you still need to set the argument?

Yes yes, I was testing without it first :-)

ppatierno:
@scholzj it's now ready :-)

scholzj (approved):

fvaleri (approved):

see-quick (approved):

PR #9078, opened 2023-09-06
[ST] addition of kafka version check in kafka status when upgrading kafka

henryZrncik:
### Type of change

- Enhancement

### Description

Addition of check whether fields `spec.kafka.version` and `status.version` are matching after kafka version Upgrade. 

### Checklist

- [x] Write tests
- [x] Update documentation

### DIFF
--- a/systemtest/src/main/java/io/strimzi/systemtest/utils/kafkaUtils/KafkaUtils.java
+++ b/systemtest/src/main/java/io/strimzi/systemtest/utils/kafkaUtils/KafkaUtils.java
@@ -109,6 +109,15 @@ public class KafkaUtils {
             });
     }
 
+    public static void waitUntilStatusKafkaVersionMatchesSpecificationKafkaVersion(String clusterName, String namespace) {
+        TestUtils.waitFor("Kafka version in Kafka cluster '" + clusterName + "' to match",
+            Constants.GLOBAL_POLL_INTERVAL, Constants.GLOBAL_STATUS_TIMEOUT, () -> {
+                String kafkaVersionInSpec = KafkaResource.kafkaClient().inNamespace(namespace).withName(clusterName).get().getSpec().getKafka().getVersion();
+                String kafkaVersionInStatus = KafkaResource.kafkaClient().inNamespace(namespace).withName(clusterName).get().getStatus().getKafkaVersion();
+                return kafkaVersionInSpec.equals(kafkaVersionInStatus);
+            });
+    }
+
     public static void waitUntilKafkaStatusConditionContainsMessage(String clusterName, String namespace, String pattern) {
         waitUntilKafkaStatusConditionContainsMessage(clusterName, namespace, pattern, Constants.GLOBAL_STATUS_TIMEOUT);
     }
--- a/systemtest/src/main/java/io/strimzi/systemtest/utils/kafkaUtils/KafkaUtils.java
+++ b/systemtest/src/main/java/io/strimzi/systemtest/utils/kafkaUtils/KafkaUtils.java
@@ -109,12 +109,11 @@ public class KafkaUtils {
             });
     }
 
-    public static void waitUntilStatusKafkaVersionMatchesSpecificationKafkaVersion(String clusterName, String namespace) {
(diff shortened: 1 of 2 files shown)


Frawless (approved):

systemtest/src/main/java/io/strimzi/systemtest/utils/kafkaUtils/KafkaUtils.java
```diff
@@ -109,6 +109,15 @@ public static void waitUntilKafkaStatusConditionContainsMessage(String clusterNa
             });
     }
 
+    public static void waitUntilStatusKafkaVersionMatchesSpecificationKafkaVersion(String clusterName, String namespace) {
```
im-konge, line 112: I don't have problem with this method, but I guess it can produce false-positives or? What if due to some race condition the spec will not be updated in time and you will wait until status of Kafka will contain older/wrong version?

I would pass here the Kafka version we are expecting to be there.

henryZrncik: I see, I originally implemented that method exactly as you suggest, but i came to conslusion that I am basically more interested in fact that it really reflects what was specified in given CR where it should be actually present. there is no use case where i would actually want to compare this version with anything else than newest version. 
Additionally, in given test there is already check of this fact in line https://github.com/strimzi/strimzi-kafka-operator/pull/9078/files#diff-d04f8fc75d2ee2007e6eb5ef26f2779cf38dde7bf75e28e85c1a1f5635997964R322-R323   So i was focusing here more on operator. 

But you are right with possibly false positive assuming e.g. somebody will update version from 3.4.0 to 3.5.0 and also downgrade immedaitelly test would probably pass.

see-quick: Well in such case, you can simply store the previous version of both values and make sure that these values (i.e., `status.version`, `spec.kafka.version`) have been updated. And then such race conditions could not happen. Or?

im-konge: I guess that for other cases when you'll want to use this method, having it exactly defined on what it should wait would make more sense FMPOV. Storing and checking previous values are extra operations, which you don't need, right?
Because even you will store previous values and it will change to something you don't want, you will continue with the test even if the values are wrong.

Yeah for this case when you already have Kafka upgraded and it's checked before, it doesn't matter. But for other test cases and situations, where you will want to wait for specific Kafka version, the race condition can happen and the test(s) will fail.

Also from Sam's proposal, this update to Kafka status should replace checking the libs on the Pod, so we can remove these checks from STs in the future. And then the version will not be checked anywhere, only using this kind of method.

scholzj: > Also from Sam's proposal, this update to Kafka status should replace checking the libs on the Pod, so we can remove these checks from STs in the future. And then the version will not be checked anywhere, only using this kind of method.

No, it should not. You should not test just that the version is set. You need to test that the upgrade was done properly! so watching the rolling updates and so on is still important.

im-konge: By that I meant to remove the method for checking the libs on Kafka Pods in the future, not removing methods for checking rolling updates etc. Or you mean that we should still check the version of libs on the Kafka Pods?

scholzj: Well, especially given you already have it, I would keep it at least unless you need to do some major redesign that it would complicate, block, or something like that.

henryZrncik: all discussed steps are remaining in given test (i.e., we do check if update really happened, pods were rolled version is what it expected), I also added extra parameter to specify exact version which is expected (to avoid mentioned unexpected false-positives or future misuse of given function).

im-konge (approved):

see-quick (approved):

PR #9088, opened 2023-09-07
[systemtest] Add annotation for NodePools only mode in tests

im-konge:
### Type of change

- Enhancement 

### Description

This PR adds annotation for NodePools only mode, with which we can annotate tests that should run only when we have NodePools enabled inside the `STRIMZI_FEATURE_GATES` env. variable.

It will be used as part of #8918 , but we will maybe find other places where we can use it.

### Checklist


Frawless (approved):

fvaleri:
Thanks guys. Can we merge this?

PR #9092, opened 2023-09-07
[ST] Namespace deletion recovery Class, namespace usage fix

henryZrncik:
### Type of change

- Refactoring

### Description
in Class `NamespaceDeletionRecoveryST` After refactor to usage of new constants for namespaces (i.e.,. `Constants.CO_NAMESPACE`, `Constants.TEST_SUITE_NAMESPACE`) namespace for re-creation of cluster operator was used incorrectly, causing CO to be created and watch different namespace. 

This PR fixes used namespace for CO re-creation. 
Additionally, there is switch to use `testStorage` in all manipulation with namespaces to promote usage of the same namespace in case of future changes/refactoring.

### DIFF
--- a/systemtest/src/test/java/io/strimzi/systemtest/operators/NamespaceDeletionRecoveryST.java
+++ b/systemtest/src/test/java/io/strimzi/systemtest/operators/NamespaceDeletionRecoveryST.java
@@ -159,6 +159,9 @@ class NamespaceDeletionRecoveryST extends AbstractST {
 
         LOGGER.info("Recreating Kafka cluster without Topic Operator");
         resourceManager.createResourceWithWait(extensionContext, KafkaTemplates.kafkaPersistent(testStorage.getClusterName(), 3, 3)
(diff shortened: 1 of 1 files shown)


Frawless (approved):

henryZrncik:
@strimzi-ci run tests --profile=all --install-type=bundle --cluster-type=ocp --cluster-version=4.14 --testcase=NamespaceDeletionRecoveryST

see-quick (approved):
LGTM, please check it with:
1. STRIMZI_RBAC_SCOPE=NAMESPACE

henryZrncik:
Although test passes as expected, this is actually a good suggestion to cover all unexpected behaviour, due to discussion https://github.com/strimzi/strimzi-kafka-operator/pull/9092#pullrequestreview-1617002980 (@see-quick) i also added explicit usage of namesapce in kafka re-creation.

im-konge (approved):

PR #9094, opened 2023-09-08
[ST] PodSetSt Reconcilation test, checks addition

henryZrncik:
### Type of change

- Refactoring

### Description

checks added to make sure status of `StrimziPodSet` is updated as expected, and clients can produce and consume data after changes take place.

### DIFF
--- a/systemtest/src/test/java/io/strimzi/systemtest/operators/PodSetST.java
+++ b/systemtest/src/test/java/io/strimzi/systemtest/operators/PodSetST.java
@@ -10,12 +10,16 @@ import io.strimzi.api.kafka.model.ProbeBuilder;
 import io.strimzi.systemtest.AbstractST;
 import io.strimzi.systemtest.Environment;
 import io.strimzi.systemtest.annotations.IsolatedTest;
+import io.strimzi.systemtest.kafkaclients.internalClients.KafkaClients;
+import io.strimzi.systemtest.kafkaclients.internalClients.KafkaClientsBuilder;
 import io.strimzi.systemtest.resources.crd.KafkaResource;
 import io.strimzi.systemtest.resources.kubernetes.DeploymentResource;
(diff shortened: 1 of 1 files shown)


systemtest/src/test/java/io/strimzi/systemtest/operators/PodSetST.java
```diff
@@ -46,6 +50,16 @@ void testPodSetOnlyReconciliation(ExtensionContext extensionContext) {
         final int replicas = 3;
         final int probeTimeoutSeconds = 6;
 
+        // setup clients
+        KafkaClients clients = new KafkaClientsBuilder()
```
see-quick, line 54: I am also thinking about continuous sending messages during update of FG? Do you think it is worth @im-konge @henryZrncik ?

see-quick (approved):
LGTM, just one comment. Anyway looks good! 👍

systemtest/src/test/java/io/strimzi/systemtest/operators/PodSetST.java
henryZrncik: The changes would be, as some of brokers will go down (not only mentioning reconciliation),   config producer to `ack=al`l, broker to create topics by default with appropriate `min.insync.replicas` and `replicaiton` factor). Similar workaround  was done in `SecurityST testCaRenewalBreakInMiddle`

But I do think it is worth adding there.

henryZrncik: I added suggestions into the code base. 👍

henryZrncik:
@strimzi-ci run tests --profile=all --install-type=bundle --cluster-type=ocp --cluster-version=4.14 --testcase=PodSetST

im-konge (approved):

PR #9098, opened 2023-09-08
Add container signing with cosign

scholzj:
### Type of change

- Enhancement / new feature

### Description

This PR adds container signing using our own keyset and the `cosign` tool. The signing is done only in the CI and should not affect any local builds done by the developers. The README.md is updated with the information about how to verify the containers if needed.

### Checklist

- [x] Update CHANGELOG.md

### DIFF
--- a/.azure/templates/jobs/build/push_containers.yaml
+++ b/.azure/templates/jobs/build/push_containers.yaml
@@ -11,6 +11,7 @@ jobs:
       # Install Prerequisites
       - template: "../../steps/prerequisites/install_docker.yaml"
       - template: "../../steps/prerequisites/install_yq.yaml"
+      - template: "../../steps/prerequisites/install_cosign.yaml"
 
       # Get the container archives
       - ${{ each arch in parameters.architectures }}:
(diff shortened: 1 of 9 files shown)


ppatierno:
From where the request to do so is coming? Any issue from the community? What's the purpose? Why it wasn't done until today?

scholzj:
> From where the request to do so is coming? Any issue from the community?

I'm not aware of any issue. But signing the containers is becoming a common security practice, so it seemed useful to have it.

> What's the purpose?

You can verify if the image was really built by Strimzi. While you can verify it manually, the verification can be also done directly on Kubernetes by various tools.

> Why it wasn't done until today?

I guess because nobody did it earlier? 😉 But also, as with many other things ... new things appear all the time and sometimes become common practice over time.

.azure/templates/jobs/build/push_containers.yaml
```diff
@@ -64,3 +65,15 @@ jobs:
...
+        displayName: "Sign container manifests"
+        env:
+          BUILD_REASON: $(Build.Reason)
+          BRANCH: $(Build.SourceBranch)
+          BUILD_ID: $(Build.BuildId)
+          BUILD_COMMIT: $(Build.SourceVersion)
+          DOCKER_REGISTRY: "quay.io"
+          DOCKER_ORG: "strimzi"
+          DOCKER_TAG: '${{ parameters.dockerTag }}'
+          COSIGN_PASSWORD: $(COSIGN_PASSWORD)
```
ppatierno, line 78: what's this env var? I see it's never used, or?

scholzj: It is the password to the private key. It is used by the `cosign` tool automatically, so it has to be passed but you will not see it used directly in our code.

ppatierno: So I don't see it explicitly used here in your changes because that is where cosign expects the password by itself, right?

scholzj: Yes, correct. The `cosign` binary reads this directly.

ppatierno (approved):
LGTM

Issue #7988, opened 2023-01-27
Make label and annotation keys expected to be used by users part of the _public_ API in the `api` module


scholzj:
As originally discussed in #7983 => it would make sense to have the key annotations and labels we expect to be used by users in the `api` module and not in the `operator-common` module. That would make it easier to use them when automating the API. This should for example include the `strimzi.io/cluster` label, the different restart or rolling update annotations etc.

DevPJ9:
Hello @scholzj. I would like to work on this.

scholzj:
I think we need to triage this first and think through where exactly the label and annotation keys should be moved. But I can ping you once it is ready to be worked on.

DevPJ9:
Ok, sure. Thanks for the information.

scholzj:
Triaged on 9.2.2023: Makes sense to move it to the `api` module:
* It should be one class for Labels and one for Annotations
* The constant name in the class should describe what it should be applied to (e.g. Secret, StrimziPodset, etc.)
* We should try to move all _public_ labels and annotations. Not just the `strimzi.io/cluster` label. So e.g. the manual rolling update annotations, CA renewal annotations etc.

scholzj:
@DevPJ9 We added a bit more details about how it might be implemented. So feel free to look at it if you are still interested.

DevPJ9:
ok @scholzj. Thanks for the update. I am looking into it.

scholzj:
I guess that part of the task is to come up with them - at least with the initial batch. Going through the docs and looking for all the procedures where the user is instructed to label or annotate something should be a good start I think.

AnshulAngaria:
@scholzj I would like to work on this.

scholzj:
@AnshulAngaria: As you see above, @DevPJ9 planned to work on this. I'm not sure if this is still the case or not.

DevPJ9:
@scholzj I will start working on it after this week. There are some things which are keeping me engaged for now. If @AnshulAngaria wants he can start working on the task and if he finishes it, that's great. So yeah, @AnshulAngaria you can work on this task.

AnshulAngaria:
@DevPJ9 @scholzj Thanks, I will start working on this.

pratikkolkar:
@scholzj - Can I take up this issue?. If yes, below I have stated my understanding of the issue and having some doubts.

**Problem Statement:**
Move Labels class from ```operator-common``` module to ```api``` module.

[...]

PR #9103, opened 2023-09-12
#7988-added labels and annotation keys into api module

pratikkolkar:
### Type of change

_Select the type of your PR_

- Enhancement 

### Description

As per issue #7988. Made label and annotation keys available in api module by adding `Labels` and `Annotations` class under `io.strimzi.api` package.

### Checklist

- [ ] Write tests
- [ ] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/api/src/main/java/io/strimzi/api/Annotations.java
+++ b/api/src/main/java/io/strimzi/api/Annotations.java
@@ -5,8 +5,7 @@
 package io.strimzi.api;
 
 /**
- * Class for holding some annotation keys and utility methods for handling
- * annotations
+ * Class for holding some annotation keys.
  */
 public class Annotations {
     /**
--- a/api/src/main/java/io/strimzi/api/Annotations.java
+++ b/api/src/main/java/io/strimzi/api/Annotations.java
@@ -1,0 +1,106 @@
+/*
+ * Copyright Strimzi authors.
+ * License: Apache License 2.0 (see the file LICENSE or http://apache.org/licenses/LICENSE-2.0.html).
+ */
+package io.strimzi.api;
+
+/**
+ * Class for holding some annotation keys and utility methods for handling
+ * annotations
+ */
+public class Annotations {
+    /**
+     * The Strimzi domain used in all annotations
+     */
+    public static final String STRIMZI_DOMAIN = "strimzi.io/";
+
+    /**
+     * Annotation used to force rebuild of the container image even if the
+     * dockerfile did not change
+     */
+    public static final String STRIMZI_IO_CONNECT_FORCE_REBUILD = STRIMZI_DOMAIN + "force-rebuild";
+
+    /**
+     * Annotation used to pause resource reconciliation
+     */
+    public static final String ANNO_STRIMZI_IO_PAUSE_RECONCILIATION = STRIMZI_DOMAIN + "pause-reconciliation";
+
+    /**
+     * Annotation to trigger manually rolling updats
+     */
+    public static final String ANNO_STRIMZI_IO_MANUAL_ROLLING_UPDATE = STRIMZI_DOMAIN + "manual-rolling-update";
+
+    /**
+     * This annotation with related possible values (approve, stop, refresh) is set
+     * by the user for interacting
+     * with the rebalance operator in order to start, stop, or refresh rebalancing
+     * proposals and operations.
+     */
+    public static final String ANNO_STRIMZI_IO_REBALANCE = STRIMZI_DOMAIN + "rebalance";
+
+    /**
+     * Use this boolean annotation to auto-approve a rebalance optimization proposal
+     * without the need for the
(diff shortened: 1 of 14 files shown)


scholzj (commented):
Thanks for the PR. The new classes look good. But we do not want to keep these annotations / labels twice. So you would also need to remove the original ones and in the places they were used use these instead.

api/src/main/java/io/strimzi/api/Annotations.java
```diff
@@ -0,0 +1,106 @@
+/*
+ * Copyright Strimzi authors.
+ * License: Apache License 2.0 (see the file LICENSE or http://apache.org/licenses/LICENSE-2.0.html).
+ */
+package io.strimzi.api;
+
+/**
+ * Class for holding some annotation keys and utility methods for handling
```
ppatierno, line 8: There are no utility methods in this class, or?

pratikkolkar: @scholzj - I have a suggestion instead of using the labels/annotations from new class in all the places. Can't we extend the `Labels` and `Annotations` class from `operator-common` module with the new class, the parent class(from `api` module) will be having all the labels which we moved and the child class(from `operator-common` module) can inherit them , this way I believe we can avoid doing the changes to all the places where it is used and duplicates will be handled as well. Let me know your point of view on this.

@ppatierno - You are right no utility methods, I will remove that from comment lines.

pratikkolkar:
> Thanks for the PR. The new classes look good. But we do not want to keep these annotations / labels twice. So you would also need to remove the original ones and in the places they were used use these instead.

@scholzj - I have a suggestion instead of using the labels/annotations from new class in all the places. Can't we extend the `Labels` and `Annotations` class from `operator-common` module with the new class, the parent class(from `api` module) will be having all the labels which we moved and the child class(from `operator-common` module) can inherit them , this way I believe we can avoid doing the changes to all the places where it is used and duplicates will be handled as well. Let me know your point of view on this.

scholzj:
I think that would work, yes. It would also cause less confusion in the imports in places where both can be used. Does that sound reasonable to you as well @ppatierno?

ppatierno:
Yea I am fine with it, thanks @pratikkolkar !

scholzj (commented):
I think the code changes look good. But the build failed because of some unused imports:
```
[ERROR] /home/vsts/work/1/s/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/ManualPodCleaner.java:18:8: Unused import - io.strimzi.operator.common.operator.resource.AbstractScalableNamespacedResourceOperator. [UnusedImports]
[ERROR] /home/vsts/work/1/s/cluster-operator/src/test/java/io/strimzi/operator/cluster/operator/assembly/ManualPodCleanerTest.java:19:8: Unused import - io.strimzi.operator.common.operator.resource.AbstractScalableNamespacedResourceOperator. [UnusedImports]
```
So you will need to remove those.

scholzj:
Hmm, looks like Spotbugs does not like this:
```
[ERROR] High: The class name io.strimzi.operator.common.Annotations shadows the simple name of the superclass io.strimzi.api.Annotations [io.strimzi.operator.common.Annotations] At Annotations.java:[lines 24-288] NM_SAME_SIMPLE_NAME_AS_SUPERCLASS
[ERROR] High: The class name io.strimzi.operator.common.model.Labels shadows the simple name of the superclass io.strimzi.api.Labels [io.strimzi.operator.common.model.Labels] At Labels.java:[lines 108-501] NM_SAME_SIMPLE_NAME_AS_SUPERCLASS
```

I'm sure we can suppress it. But I wonder if it has a point. Should we name the classes in the `api` module somehow differently? `ResourceAnnotations` and `ResourceLabels`? Any thoughts @tombentley & @ppatierno?

ppatierno:
> I'm sure we can suppress it. But I wonder if it has a point. Should we name the classes in the api module somehow differently? ResourceAnnotations and ResourceLabels? Any thoughts @tombentley & @ppatierno?

In general I don't like having classes linked by an inheritance relationship with same name (even if living in different packages). I would do what you are suggesting.

pratikkolkar:
I have renamed the classes as suggested. But getting this Import control error.


image

\
Is it feasible to replace `<allow pkg="io.strimzi.api.annotations" />` and `<allow pkg="io.strimzi.api.kafka" />` with `<allow pkg="io.strimzi.api" />` in import-control.xml ?

scholzj:
I would specifically add the two classes:

```
        <allow class="io.strimzi.api.ResourceLabels" />
        <allow class="io.strimzi.api.ResourceAnotations" />
```

scholzj (approved):
Thanks. This LGTM now.

ppatierno (approved):
LGTM

scholzj:
Thanks for the PR @pratikkolkar ... Nice work!

PR #9106, opened 2023-09-12
[ST] - Add support of provisioning Kind cluster

see-quick:
### Type of change

- Enhancement / new feature
- Refactoring
- Documentation

### Description

Currently, a `minikube` does not have support for IPv6 and thus, it is impossible to test such configuration on Azure pipelines using our current setup (i.e., `minikube`). Fortunately, the [kind](https://kind.sigs.k8s.io/) could solve it because it supports IPv6 or dual stack. 

I have modified our `setup-kubernetes.sh` script to be able also provision the `kind` cluster. In the future, we can also use this as a part of testing farm infrastructure where we can deploy multi-node clusters (because there we have unlimited resources). To contrast in Azure Pipelines we are limited to small machines (i.e., 6GB RAM and 2 CPU) and here we will stick with single-node I assume but (we could easily test dual and IPv6 configurations). 

### Checklist

(IPv4 profiles)
- [x] Check smoke profile 
- [x] Check acceptance profile 


.azure/scripts/setup-kubernetes.sh
```diff
@@ -1,20 +1,25 @@
...
 
-DEFAULT_MINIKUBE_MEMORY=$(free -m | grep "Mem" | awk '{print $2}')
-DEFAULT_MINIKUBE_CPU=$(awk '$1~/cpu[0-9]/{usage=($2+$4)*100/($2+$4+$5); print $1": "usage"%"}' /proc/stat | wc -l)
+DEFAULT_KIND_MEMORY=$(free -m | grep "Mem" | awk '{print $2}')
+DEFAULT_KIND_CPU=$(awk '$1~/cpu[0-9]/{usage=($2+$4)*100/($2+$4+$5); print $1": "usage"%"}' /proc/stat | wc -l)
 
-MINIKUBE_MEMORY=${MINIKUBE_MEMORY:-$DEFAULT_MINIKUBE_MEMORY}
-MINIKUBE_CPU=${MINIKUBE_CPU:-$DEFAULT_MINIKUBE_CPU}
+KIND_MEMORY=${KIND_MEMORY:-$DEFAULT_KIND_MEMORY}
+KIND_CPU=${KIND_CPU:-$DEFAULT_KIND_CPU}
```
Frawless, line 15: I think you shouldn't rename it like that. These options should be common for all clusters available via the script.

.azure/scripts/setup-kubernetes.sh
```diff
@@ -78,47 +143,85 @@ if [ "$TEST_CLUSTER" = "minikube" ]; then
...
-    # Add Docker hub credentials to Minikube
-    if [ "$COPY_DOCKER_LOGIN" = "true" ]
-    then
-      set +ex
+elif [ "$TEST_CLUSTER" = "kind" ]; then
+    install_kubectl
+    install_kubernetes_provisioner
 
-      docker exec "minikube" bash -c "echo '$(cat $HOME/.docker/config.json)'| sudo tee -a /var/lib/kubelet/config.json > /dev/null && sudo systemctl restart kubelet"
+    setup_kube_directory
```
Frawless, line 153: Wouldn't be better to run the common parts before the if statement to avoid redundant lines?

.azure/scripts/setup-kubernetes.sh
see-quick: Yeah, it's just left-over I would name it f.e. `DEFAULT_CLUSTER_MEMORY` etc.

.packit.yaml
```diff
@@ -140,9 +140,45 @@ jobs:
...
+    env: { IP_FAMILY: ipv4 }
     labels:
       - sanity
     tf_extra_params:
       test:
         tmt:
           name: sanity
+
+  ###############################################################################################
+  ##################################    IPv6/dual jobs   ########################################
```
see-quick, line 152: I wonder if we should merge one acceptance job, which will trigger also IPv6 and dual configuration instead of only IPv4

.azure/scripts/setup-kubernetes.sh
```diff
@@ -78,47 +142,108 @@ if [ "$TEST_CLUSTER" = "minikube" ]; then
     fi
 
     minikube addons enable default-storageclass
+    add_docker_hub_credentials_to_kubernetes "$TEST_CLUSTER"
+    setup_registry_proxy
+
+elif [ "$TEST_CLUSTER" = "kind" ]; then
```
im-konge, line 148: Would it make sense to have two separate scripts for `kind` and `minikube` setup, to make this script more clear? I guess that it can be confusing having both in same script file.

systemtest/src/test/java/io/strimzi/systemtest/connect/ConnectBuilderST.java
```diff
@@ -471,6 +473,9 @@ void testBuildOtherPluginTypeWithAndWithoutFileName(ExtensionContext extensionCo
     @Tag(ACCEPTANCE)
     @ParallelTest
     void testBuildPluginUsingMavenCoordinatesArtifacts(ExtensionContext extensionContext) {
+        // using kind we encounter (error building image: deleting file system after stage 0: unlinkat //product_uuid: device or resource busy)
+        assumeFalse(KubeClusterResource.getInstance().isKind());
+
```
im-konge, line 478: Maybe create annotation and skip the whole suite? I guess it will help you also for `regression` pipelines

scholzj (commented):
Do you plan to update all the other repositories as well? If not, it might be worth to keep the original names etc. to make it easier to copy between the CI configurations.

see-quick:
> Do you plan to update all the other repositories as well? If not, it might be worth to keep the original names etc. to make it easier to copy between the CI configurations.

Yeah, but first I want to make it stable in the operator repo. In this PR I have added only acceptance profile but in the next I would also add regression and I think there will be some failures related to IPv6 (I hope not :D)

scholzj:
> > Do you plan to update all the other repositories as well? If not, it might be worth to keep the original names etc. to make it easier to copy between the CI configurations.
> 
> Yeah, but first I want to make it stable in the operator repo. In this PR I have added only acceptance profile but in the next I would also add regression and I think there will be some failures related to IPv6 (I hope not :D)

So, if there are no plans for it right now, maybe you should not change the things in `./azure/templates`. They do not seem to add any value and they seem to just complicate the live when sharing the parts of the CI.

.azure/scripts/setup-kubernetes.sh
see-quick: Well, there is a lot of shared code between those two installations. So I wanted to have it as one script. I could make a `common.sh` and divide these 2 scripts, but I don't know how much value it gives. But if you really want it I can divide it :)

systemtest/src/test/java/io/strimzi/systemtest/connect/ConnectBuilderST.java
see-quick: > Maybe create annotation and skip the whole suite? I guess it will help you also for regression pipelines

I don't know if all KafkaConnectBuilder test cases will fail. I am not fully sure but on Thursday I execute whole suite and only 2-3 test cases failed. So I can for sure create that annotation but I will skip only those, which will fail.

see-quick:
> So, if there are no plans for it right now, maybe you should not change the things in ./azure/templates. They do not seem to add any value and they seem to just complicate the live when sharing the parts of the CI.



> > > Do you plan to update all the other repositories as well? If not, it might be worth to keep the original names etc. to make it easier to copy between the CI configurations.
> > 
> > 
> > Yeah, but first I want to make it stable in the operator repo. In this PR I have added only acceptance profile but in the next I would also add regression and I think there will be some failures related to IPv6 (I hope not :D)
> 
> So, if there are no plans for it right now, maybe you should not change the things in `./azure/templates`. They do not seem to add any value and they seem to just complicate the live when sharing the parts of the CI.



> > > Do you plan to update all the other repositories as well? If not, it might be worth to keep the original names etc. to make it easier to copy between the CI configurations.
> > 
> > 
> > Yeah, but first I want to make it stable in the operator repo. In this PR I have added only acceptance profile but in the next I would also add regression and I think there will be some failures related to IPv6 (I hope not :D)
> 
> So, if there are no plans for it right now, maybe you should not change the things in `./azure/templates`. They do not seem to add any value and they seem to just complicate the live when sharing the parts of the CI.

Yeah, I will undo such changes :).

.azure/scripts/setup-kubernetes.sh
see-quick: I have divided them into two separate scripts ;)

see-quick:
Specific fixes to regression on Testing Farm and IPv6-related stuff I will continue on different PR. The main aim of this PR was achieved and we can easily now test acceptance + smoke pipelines (without ZK issues).

systemtest/src/main/java/io/strimzi/systemtest/annotations/KindNotSupport.java
```diff
@@ -0,0 +1,35 @@
...
+ * <p>If the kind is supported, the test will be disabled. If not, the test will be enabled.</p>
+ *
+ * <p>This annotation leverages the {@link KindNotSupportCondition} to evaluate the execution condition.</p>
+ *
+ * @see KindNotSupportCondition
+ */
+@Target({ElementType.METHOD, ElementType.TYPE})
+@Retention(RetentionPolicy.RUNTIME)
+@ExtendWith(KindNotSupportCondition.class)
+public @interface KindNotSupport {
```
scholzj, line 27: No strong opinions ... but should be be `KindNotSupported`? Or `NoKindSupport`? `KindNotSupport` sounds super weird. (but not a blocker from my point of view)

systemtest/src/main/java/io/strimzi/systemtest/Environment.java
```diff
@@ -287,13 +295,29 @@ public static boolean useLatestReleasedBridge() {
...
+            // we will need a hostname of machine
+            String hostname = "";
+            try {
+                if (Environment.isIpv4Family()) {
+                    hostname = InetAddress.getLocalHost().getHostAddress() + ":5001";
+                }
+            } catch (UnknownHostException e) {
+                throw new RuntimeException(e);
+            }
+            LOGGER.info("Using hostname:{}", hostname);
```
scholzj, line 319: Is that ome left over debug message? If not, should i tbe at least properly formatted and have some meaning?

```suggestion
            LOGGER.info("Using container registry hostname: {}", hostname);
```

scholzj (approved):

systemtest/src/main/java/io/strimzi/systemtest/annotations/KindNotSupport.java
see-quick: Yeah, `KindNotSupported` sounds better thanks for it :)

systemtest/src/main/java/io/strimzi/systemtest/Environment.java
see-quick: Updated :) thanks

.azure/scripts/setup-kind.sh
```diff
@@ -0,0 +1,200 @@
...
+	do
+		echo ${nodeName};
+		kubectl label node ${nodeName} rack-key=zone;
+	done
+}
+
+
+function install_kubernetes_provisioner {
+
+    if [ "$TEST_CLUSTER" = "kind" ]; then
```
im-konge, line 49: It will be always `kind` when using this script, right?

.azure/scripts/setup-kind.sh
```diff
@@ -0,0 +1,200 @@
...
+
+      set -ex
+    fi
+}
+
+setup_kube_directory
+install_kubectl
+install_kubernetes_provisioner
+
+if [ "$TEST_CLUSTER" = "kind" ]; then
```
im-konge, line 90: Here as well

im-konge (approved):
Just two nits, thanks for the PR :) good job

PR #9124, opened 2023-09-15
Add support for Software Bills of Materials (SBOMs) to Strimzi CI

scholzj:
### Type of change

- Enhancement / new feature

### Description

This PR adds support for Software Bills of Materials (SBOMs) to Strimzi CI and Strimzi build. It uses Sysft to generate the SBOMs and publishes them in two ways:
* For all release builds, the SBOMs are pushed to Quay.io in SPDX-JSON format and signed with Cosign (same as our containers)
* For all release and `main` branch builds, the Azure CI produces an archive with the SBOMs in SPDX-JSON and Syft-Table formats signed with Cosign

The reason for not pushing the SBOMs for `main` branch to Quay.io is that it would create lot of records without much value (The SBOMs are pushed under the image digest and not under the tag). We might add it in the future if we see some demand.

The reason for including the SyftTable format is that it is easy for humans to read it:

```
NAME                                              VERSION                                    TYPE         
accessors-smart                                   2.4.9                                      java-archive  
acl                                               2.2.53-1.el8                               rpm           
activation                                        1.1.1                                      java-archive  
alsa-lib                                          1.2.8-2.el8                                rpm           
animal-sniffer-annotations                        1.14                                       java-archive 
...
```

_Testing CI changes affecting only the main branch or release is non-trivial. There is a chance that something will not work. But the logic here is executed at the end of the pipeline, so if there are some issues, we can fix them in follow-up PRs without any real impact other then the pipeline failing after it did everyting it does today._

This should have no impact on Strimzi developers as the changes to the build are not trigger by any of the other commands. There should be also no impact on any users who do not care about SBOMs, they can just ignore all of this.

### Checklist

- [x] Update documentation
- [x] Update CHANGELOG.md

### DIFF
--- a/README.md
+++ b/README.md
@@ -87,6 +87,26 @@ And use it to verify the signature:
 cosign verify --key strimzi.pub quay.io/strimzi/operator:latest --insecure-ignore-tlog=true
 ```
 
+## Software Bill of Materials (SBOM)
+
+From the 0.38.0 release, Strimzi publishes the software bill of materials (SBOM) of our containers.
+The SBOMs are published as an archive with `SPDX-JSON` and `Syft-Table` formats signed using cosign.
+For releases, they are also pushed into the container registry.
+To verify the SBOM signatures, please use the Strimzi public key:
+
+```
+-----BEGIN PUBLIC KEY-----
+MFkwEwYHKoZIzj0CAQYIKoZIzj0DAQcDQgAET3OleLR7h0JqatY2KkECXhA9ZAkC
+TRnbE23Wb5AzJPnpevvQ1QUEQQ5h/I4GobB7/jkGfqYkt6Ct5WOU2cc6HQ==
+-----END PUBLIC KEY-----
+```
+
+You can use it to verify the signature of the SBOM files with the following command:
+
+```
+cosign verify-blob --key cosign.pub --bundle <SBOM-file>.bundle --insecure-ignore-tlog=true <SBOM-file>
+```
+
 ---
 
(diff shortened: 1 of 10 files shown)


README.md
```diff
@@ -87,6 +87,26 @@ And use it to verify the signature:
 cosign verify --key strimzi.pub quay.io/strimzi/operator:latest --insecure-ignore-tlog=true
 ```
 
+## Software Bill of Materials (SBOM)
+
+From the 0.38.0 release, Strimzi publishes the software bill of materials (SBOM) of our containers.
+The SBOMs are published as an archive with `SPDX-JSON` and `Syft-Table` formats signed using cosign.
```
ppatierno, line 93: any reasons for choosing SPDX format? I am new to Syft but I see it can generate other formats as well (i.e. CycloneDX).

scholzj: SPDX is the Linux Foundation format. So I picked it as it is closer to us as a project (CNCF). No other reasons from my side to be honest.

scholzj: Do you see any reasons why to use CycloneDX instead of SPDX?

ppatierno: No tbh and admit my ignorance I was totally new to these formats so I slightly read about them. I was curious about your choice, but I think that it makes sense.

ppatierno (approved):

README.md
scholzj: I will be honest with you. The only reason I made my decision was on the LF / CNCF relationship as I did not have any other preferences one way or the other 🤷. 

The archive also has the Syft-Table format that you can easily read in a text editor as a human. But I had not much else to decide on for the machine-readable formats.

PR #9145, opened 2023-09-21
Fix missing CO replicaset and pod YAMLs in report.sh

fvaleri:
This change includes a couple of small fixes and consistency improvements from #8918.

- Fixed CO replicaset that was missing "-o yaml" flag, thus producing an invalid YAML
- Changed placeholder [hidden] with <hidden> to avoid parsing errors with tools like yq
- Added logic to also get the CO pod YAML content
- Merged CO code blocks into one (consistency)
- Changed crds and crs dirs to customresourcedefinitions and customresources (consistency)
- Changed podlogs dir name to logs, as this actually hosts container logs

### DIFF
--- a/tools/report.sh
+++ b/tools/report.sh
@@ -21,7 +21,7 @@ SECRETS_OPT="hidden"
 # sed non-printable text delimiter
 SD=$(echo -en "\001") && readonly SD
 # sed sensitive information filter expression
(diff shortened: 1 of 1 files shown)


scholzj (approved):
LGTM I guess.

ppatierno (approved):

PR #9153, opened 2023-09-23
Allow manual rolling updates of Connect and MM2 clusters

### LINKED ISSUE #9142: [Enhancement]: Ability to restart kafka-connect instances via the operator
### Related problem

When performing cluster wide rolling restarts, i am unsure how to perform a "safe" restart of connect-instances. Currently i run `kubectl rollout restart deployment` on the connect deployment however i have observed that new pods sometimes get terminated when the operator reconciles leaving me in a position whereby a pod is potentially running an outdated sidecar.

I am aware that i can restart the following:

- kafka - annotate pods with strimzi.io/manual-rolling-update=true
- zookeeper - annotate pods with strimzi.io/manual-rolling-update=true
- connectors - https://github.com/strimzi/proposals/blob/main/007-restarting-kafka-connect-connectors-and-tasks.md

### Suggested solution

Ability to annotate either the deployment or pods such that the operator restarts the connect instances safely (closing down the connectors etc).

### Alternatives

1. run `kubectl rollout restart deployment` multiple times until all old pods are replaced
2. Shotgun solution `kubectl delete pod PODNAME`

### Additional context

https://github.com/orgs/strimzi/discussions/9140

scholzj:
### Type of change

- Enhancement / new feature

### Description

The `StableConnectIdentities` feature gate moved all Connect and MM2 clusters to StrimziPodSets. However, we do not have any way to do a manual rolling update. Before the feature gate, when Deployments were used, users could use `kubectl` to rollout the pods. But this does not work with StrimziPodSets.

This PR extends the existing manual rolling update annotation used for Kafka brokers or Zoo also to Connect and MM2. Users can now roll them manually by annotating the Pods or the StrimziPodSets. The current `KafkaConnectRoller` class was changed similarly to how `KafkaRoller` works (it is of course still significantly simpler) -> it now receives a callable to evaluate of rolling update is needed or not. This way, we can use the same roller logic for both regular rolling updates as well as for manual rolling updates (and in the long-term future for example for CA rolling updates etc.).

This should resolve #9142.

Docs will be done in a separate PR.

### Checklist

- [x] Write tests
- [x] Update CHANGELOG.md

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 * Sign containers using `cosign`
 * Generate and publish Software Bill of Materials (SBOMs) of Strimzi containers
 * Add support for stopping connectors according to [Strimzi Proposal #54](https://github.com/strimzi/proposals/blob/main/054-stopping-kafka-connect-connectors.md)
+* Allow manual rolling of Kafka Connect and Kafka Mirror Maker 2 pods using the `strimzi.io/manual-rolling-update` annotation (supported only when `StableConnectIdentities` feature gate is enabled) 
 * Make sure brokers are empty before scaling them down
 
 ### Changes, deprecations and removals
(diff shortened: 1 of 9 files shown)


ppatierno (approved):
LGTM

ppatierno (approved):
Had another look, it seems to be good.

PR #9156, opened 2023-09-25
Add Unidirectional Topic Operator metrics

### LINKED ISSUE #8650: [Enhancement]: Metrics for Unidirectional Topic Operator
### Related problem

The [initial PR for the UTO](https://github.com/strimzi/strimzi-kafka-operator/pull/8627) does not include metrics. Before we can mark the feature as beta we will need to export metrics that are consistent with the existing BTO via the `/metrics` endpoint.

### Suggested solution

Metrics implementation.

### Alternatives

_No response_

### Additional context

_No response_

fvaleri:
This change adds Prometheus metrics to the Unidirectional Topic Operator (UTO).

In addition to the standard JVM metrics, we provide the following custom metrics:

- `strimzi.resources` (gauge): Number of custom resources the operator sees
- `strimzi.reconciliations` (counter): Number of reconciliations done by the operator for individual resources
- `strimzi.reconciliations.failed` (counter): Number of reconciliations done by the operator for individual resources which failed
- `strimzi.reconciliations.successful` (counter): Number of reconciliations done by the operator for individual resources which were successful
- `strimzi.reconciliations.duration` (timer): The time the reconciliation takes to complete
- `strimzi.reconciliations.paused` (gauge): Number of custom resources the operator sees but does not reconcile due to paused reconciliations
- `strimzi.reconciliations.locked` (counter): Number of reconciliations skipped because another reconciliation for the same resource was still running
- `strimzi.reconciliations.max.queue.size` (gauge): Max size recorded for the shared event queue
- `strimzi.reconciliations.max.batch.size` (gauge): Max size recorded for a single event batch

Should close #8650.

### DIFF
--- a/operator-common/src/main/java/io/strimzi/operator/common/metrics/MetricsHolder.java
+++ b/operator-common/src/main/java/io/strimzi/operator/common/metrics/MetricsHolder.java
@@ -35,8 +35,6 @@ public abstract class MetricsHolder {
     private final Map<String, Counter> successfulReconciliationsCounterMap = new ConcurrentHashMap<>(1);
     private final Map<String, Counter> lockedReconciliationsCounterMap = new ConcurrentHashMap<>(1);
     private final Map<String, Timer> reconciliationsTimerMap = new ConcurrentHashMap<>(1);
-    private final Map<String, AtomicInteger> reconciliationsMaxQueueMap = new ConcurrentHashMap<>(1);
-    private final Map<String, AtomicInteger> reconciliationsMaxBatchMap = new ConcurrentHashMap<>(1);
 
     /**
      * Constructs the metrics holder
@@ -161,30 +159,6 @@ public abstract class MetricsHolder {
     public Counter lockedReconciliationsCounter(String namespace) {
         return getCounter(namespace, kind, METRICS_PREFIX + "reconciliations.locked", metricsProvider, selectorLabels, lockedReconciliationsCounterMap,
                 "Number of reconciliations skipped because another reconciliation for the same resource was still running");
-    }
-
-    /**
-     * Gauge metric for the max size recorded for the event queue.
-     *
-     * @param namespace Namespace of the resources being reconciled
-     *
-     * @return Metrics gauge
-     */
-    public AtomicInteger reconciliationsMaxQueueSize(String namespace) {
-        return getGauge(namespace, kind, METRICS_PREFIX + "reconciliations.max.queue.size",
-            metricsProvider, selectorLabels, reconciliationsMaxQueueMap, "Max size recorded for the shared event queue");
-    }
-
-    /**
-     * Gauge metric for the max size recorded for the event batch.
-     *
-     * @param namespace Namespace of the resources being reconciled
-     *
-     * @return Metrics gauge
-     */
-    public AtomicInteger reconciliationsMaxBatchSize(String namespace) {
-        return getGauge(namespace, kind, METRICS_PREFIX + "reconciliations.max.batch.size",
-            metricsProvider, selectorLabels, reconciliationsMaxBatchMap, "Max size recorded for a single event batch");
     }
 
     ////////////////////
--- a/operator-common/src/main/java/io/strimzi/operator/common/metrics/MetricsHolder.java
+++ b/operator-common/src/main/java/io/strimzi/operator/common/metrics/MetricsHolder.java
@@ -35,6 +35,8 @@ public abstract class MetricsHolder {
     private final Map<String, Counter> successfulReconciliationsCounterMap = new ConcurrentHashMap<>(1);
     private final Map<String, Counter> lockedReconciliationsCounterMap = new ConcurrentHashMap<>(1);
     private final Map<String, Timer> reconciliationsTimerMap = new ConcurrentHashMap<>(1);
+    private final Map<String, AtomicInteger> reconciliationsMaxQueueMap = new ConcurrentHashMap<>(1);
+    private final Map<String, AtomicInteger> reconciliationsMaxBatchMap = new ConcurrentHashMap<>(1);
 
     /**
      * Constructs the metrics holder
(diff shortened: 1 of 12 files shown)


scholzj (commented):
LGTM I guess, but should be reviewed by @tombentley as the SME. Also, the tests do not seem to work in the CI.

topic-operator/src/main/java/io/strimzi/operator/topic/v2/BatchingTopicController.java
```diff
@@ -416,9 +422,11 @@ private void updateInternal(List<ReconcilableTopic> topics) {
             deleteInternal(toBeDeleted, false);
         }
 
+        var remainingAfterDeletions = partitionedByDeletion.get(false);
+        remainingAfterDeletions.forEach(rt -> rt.startTimer(metrics));
```
tombentley, line 426: Why is it correct that we start the timer when processing the batch, rather than from the point we're told about the change to the topic? (i.e. why are we excluding the time in the queue?)

topic-operator/src/main/java/io/strimzi/operator/topic/v2/ReconcilableTopic.java
```diff
@@ -4,16 +4,74 @@
...
+            sample = Timer.start(metrics.metricsProvider().meterRegistry());
+        }
+    }
+
+    /**
+     * Stop the reconciliation timer if present.
+     * @param metrics Metrics holder
+     */
+    public void stopTimer(MetricsHolder metrics) {
+        if (sample != null) {
```
tombentley, line 71: I find this pattern a little odd. I think the RT should be dumb holder of information. The interaction with the `MetricsHolder` could be done in the BTC, with the RT just holding the sample. That way it doesn't need to know about `MetricsHolder`, nor which metric is involved.

tombentley (commented):
Thanks Fede! I had a few points, mostly questions at this stage.

topic-operator/src/main/java/io/strimzi/operator/topic/v2/BatchingLoop.java
```diff
@@ -93,7 +100,8 @@ public void stop() throws InterruptedException {
      */
     public void offer(TopicEvent event) {
         if (queue.offerFirst(event)) {
-            LOGGER.debugOp("Item {} push onto the deque tail", event);
+            LOGGER.debugOp("Item {} push onto the deque head", event);
```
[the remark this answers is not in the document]
fvaleri: We use offerFirst and pollFirst. Looking at the Deque javadoc it looks like they both work with the 1st element of the underline queue, which I think is located in the head/front. Does it make sense?

topic-operator/src/main/java/io/strimzi/operator/topic/v2/BatchingTopicController.java
```diff
@@ -787,6 +796,7 @@ private void deleteInternal(List<ReconcilableTopic> batch, boolean onDeletePath)
                 var e = validate(reconcilableTopic);
                 if (e.isRightEqual(true)) {
                     // adminDelete, removeFinalizer, forgetTopic, updateStatus
+                    reconcilableTopic.startTimer(metrics);
```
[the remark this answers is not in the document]
fvaleri: The reason is to avoid iterating on partitionedByManaged twice, which is also not allowed by the Stream API, so I'll have to clone that stream. I move that to the following map, which I think it's a bit more clean. Wdyt?

topic-operator/src/main/java/io/strimzi/operator/topic/v2/BatchingTopicController.java
fvaleri: The User Operator also has an event queue, and we are excluding time spent there for this metric. We could add a `strimzi.queue.duration` timer that could reveal a pathological situation where lots of changes are done on the same KafkaTopic, starving other resources. That said, we already have the cheaper `strimzi.reconciliations.locked` counter for that. Wdyt?

topic-operator/src/main/java/io/strimzi/operator/topic/v2/ReconcilableTopic.java
fvaleri: I agree, will change that. Thanks.

topic-operator/src/main/java/io/strimzi/operator/topic/v2/BatchingTopicController.java
```diff
@@ -819,6 +829,9 @@ private void deleteInternal(List<ReconcilableTopic> batch, boolean onDeletePath)
                 }
             }
             forgetTopic(pair.getKey());
+            metrics.reconciliationsCounter(namespace).increment();
```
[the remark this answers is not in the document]
fvaleri: You are correct, we should update at the beginning. Thanks.

scholzj (approved):
I have not much clue about the Topic Operator. So it LGTM. 🤷

topic-operator/src/main/java/io/strimzi/operator/topic/v2/BatchingLoop.java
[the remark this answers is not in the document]
tombentley: first/last is fine, but what I was getting at was to update the log message to use something better than tail/head. I struggle to come up with a nice log message using first. `Item {} added to front of queue` is nicer than `Item {} added as first in queue`.

topic-operator/src/main/java/io/strimzi/operator/topic/v2/BatchingTopicController.java
tombentley: If we're being consistent with UO then I guess I'm fine with keeping it like this.

operator-common/src/main/java/io/strimzi/operator/common/metrics/MetricsHolder.java
```diff
@@ -161,6 +163,30 @@ public Counter lockedReconciliationsCounter(String namespace) {
...
+    }
+
+    /**
+     * Gauge metric for the max size recorded for the event batch.
+     *
+     * @param namespace Namespace of the resources being reconciled
+     *
+     * @return Metrics gauge
+     */
+    public AtomicInteger reconciliationsMaxBatchSize(String namespace) {
```
tombentley, line 185: I don't especially like how the batch concept ends up leaking into the common `MetricsHolder` here. Non-batched operators are the norm. 

Another very minor gripe (which is out of scope for this PR) is that `MetricsHolder` (which is kind of a group of metrics) is bringing assumptions about namespaces. What that means for UTO is we end up having to pass around the `namespace` to a few classes which never cared about the namespace before, just so we have it to pass to the `MetricsHolder`. 

These points suggest that perhaps, eventually, we should think about refactoring the MetricsHolder to better cope be more flexible about the differences between different operators. In other words, perhaps there should be a `MetricsGroup` concept and a thing which returns a `MetricsGroup` for a given namespace, and different operators can have different metrics groups.

topic-operator/src/main/java/io/strimzi/operator/topic/v2/BatchingTopicController.java
```diff
@@ -1020,4 +1042,16 @@ private static boolean isDifferentCondition(Condition oldReadyCondition,
                 || !Objects.equals(oldReadyCondition.getReason(), newReason)
                 || !Objects.equals(oldReadyCondition.getMessage(), newMessage);
     }
+
+    private void startTimer(ReconcilableTopic topic) {
+        if (topic.timerSample() == null) {
+            topic.timerSample(Timer.start(metrics.metricsProvider().meterRegistry()));
```
tombentley, line 1048: These names should include what we're timing. E.g. `startReconciliationTimer()` and `topic.reconciliationTimerStample()`. Better convey (when reading the code) what it is that we're timing, and allow us to add other timers later on.

tombentley (approved):
A couple more comments, but LGTM. Thanks again!

operator-common/src/main/java/io/strimzi/operator/common/metrics/MetricsHolder.java
fvaleri: I agree. For the moment, let me introduce a `BatchOperatorMetricsHolder` to host these new metrics. I'm also taking a note to implement the `MetricsGroup` concept as a follow up (along with few other notes I already have on this topic).

fvaleri:
The plan with this PR is to wait for #9163 (which is almost ready), and then add the missing paused reconciliation metric.

fvaleri:
@tombentley and @scholzj I rebased and added `strimzi.reconciliations.paused metric`. Please have a look and run STs. Thanks.

PR #9163, opened 2023-09-26
Add support for pausing KafkaTopic reconciliations with UTO

### LINKED ISSUE #9113: [Bug]: The UnidirectionalTopicOperator does not support reconciliation pausing
### Bug Description

Adding the `strimzi.io/pause-reconciliation: "true"` does not pause the reconciliation when using the UnidirectionalTopicOperator (UTO). This feature is documented for all Strimzi custom resources, and it works on the BidirectionalTopicOperator.

### Steps to reproduce

1. Deploy a cluster with the `UnidirectionalTopicOperator` feature gate enabled
2. Create a new topic t1 and wait for reconciliation
3. Annotate t1 with `strimzi.io/pause-reconciliation: "true"`
4. Edit t1 and increase the partition number

### Expected behavior

No reconciliation should happen when a `KafkaTopic` has the `strimzi.io/pause-reconciliation: "true"` annotation.

### Strimzi version

main

### Kubernetes version

1.25

### Installation method

_No response_

### Infrastructure

_No response_

_No response_

### Additional context

_No response_

fvaleri:
This change adds support for pausing `KafkaTopic` reconciliations with the Unidirectional Topic Operator (UTO). After the `strimzi.io/pause-reconciliation` annotation is set to "true" on a `KafkaTopic`, no configuration change will be reconciled. All valid changes will be reconciled as soon as the annotation is set to "false" or removed. Note that `KafkaTopic` deletions are always reconciled, even when the resource is paused.

Should close #9113.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 * Allow manual rolling of Kafka Connect and Kafka Mirror Maker 2 pods using the `strimzi.io/manual-rolling-update` annotation (supported only when `StableConnectIdentities` feature gate is enabled) 
 * Make sure brokers are empty before scaling them down
 * Update Cruise Control to 2.5.128
+* Add support for pausing `KafkaTopic` reconciliations with the UnidirectionalTopicOperator
 
 ### Changes, deprecations and removals
 
--- a/topic-operator/src/main/java/io/strimzi/operator/topic/v2/BatchingTopicController.java
+++ b/topic-operator/src/main/java/io/strimzi/operator/topic/v2/BatchingTopicController.java
@@ -11,6 +11,7 @@ import io.strimzi.api.kafka.model.KafkaTopic;
 import io.strimzi.api.kafka.model.KafkaTopicBuilder;
 import io.strimzi.api.kafka.model.status.Condition;
 import io.strimzi.api.kafka.model.status.ConditionBuilder;
+import io.strimzi.operator.common.Annotations;
 import io.strimzi.operator.common.Reconciliation;
 import io.strimzi.operator.common.ReconciliationLogger;
 import io.strimzi.operator.common.model.StatusUtils;
@@ -62,7 +63,6 @@ public class BatchingTopicController {
 
     static final String FINALIZER = "strimzi.io/topic-operator";
     static final String MANAGED = "strimzi.io/managed";
-    static final String PAUSED = "strimzi.io/pause-reconciliation";
     static final String AUTO_CREATE_TOPICS_ENABLE = "auto.create.topics.enable";
     private final boolean useFinalizer;
 
@@ -114,10 +114,7 @@ public class BatchingTopicController {
     }
 
     /* test */ static boolean isPaused(KafkaTopic kt) {
-        return kt.getMetadata() != null
-            && kt.getMetadata().getAnnotations() != null
-            && kt.getMetadata().getAnnotations().get(PAUSED) != null
-            && "true".equals(kt.getMetadata().getAnnotations().get(PAUSED));
+        return Annotations.isReconciliationPausedWithAnnotation(kt);
     }
 
     private static boolean isForDeletion(KafkaTopic kt) {
--- a/topic-operator/src/main/java/io/strimzi/operator/topic/v2/BatchingTopicController.java
+++ b/topic-operator/src/main/java/io/strimzi/operator/topic/v2/BatchingTopicController.java
@@ -1001,6 +1001,8 @@ public class BatchingTopicController {
                 || oldStatus.getTopicName() == null
                 || oldReadyCondition == null
                 || isDifferentCondition(oldReadyCondition, condition)) {
+            // the observedGeneration is initialized to 0 when creating a paused user (oldStatus null, paused true)
+            // this will result in metadata.generation: 1 > status.observedGeneration: 0 (not reconciled)
             long observedGeneration = oldStatus != null
                 ? !isPaused(kt) ? kt.getMetadata().getGeneration() : oldStatus.getObservedGeneration()
                 : !isPaused(kt) ? kt.getMetadata().getGeneration() : 0L;
--- a/topic-operator/src/main/java/io/strimzi/operator/topic/v2/BatchingTopicController.java
+++ b/topic-operator/src/main/java/io/strimzi/operator/topic/v2/BatchingTopicController.java
@@ -62,6 +62,7 @@ public class BatchingTopicController {
 
     static final String FINALIZER = "strimzi.io/topic-operator";
     static final String MANAGED = "strimzi.io/managed";
+    static final String PAUSED = "strimzi.io/pause-reconciliation";
     static final String AUTO_CREATE_TOPICS_ENABLE = "auto.create.topics.enable";
     private final boolean useFinalizer;
 
@@ -110,6 +111,13 @@ public class BatchingTopicController {
                 || kt.getMetadata().getAnnotations() == null
                 || kt.getMetadata().getAnnotations().get(MANAGED) == null
                 || !"false".equals(kt.getMetadata().getAnnotations().get(MANAGED));
+    }
+
+    /* test */ static boolean isPaused(KafkaTopic kt) {
+        return kt.getMetadata() != null
+            && kt.getMetadata().getAnnotations() != null
+            && kt.getMetadata().getAnnotations().get(PAUSED) != null
+            && "true".equals(kt.getMetadata().getAnnotations().get(PAUSED));
     }
 
     private static boolean isForDeletion(KafkaTopic kt) {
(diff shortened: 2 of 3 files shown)


topic-operator/src/main/java/io/strimzi/operator/topic/v2/BatchingTopicController.java
```diff
@@ -62,6 +62,7 @@ public class BatchingTopicController {
 
     static final String FINALIZER = "strimzi.io/topic-operator";
     static final String MANAGED = "strimzi.io/managed";
+    static final String PAUSED = "strimzi.io/pause-reconciliation";
```
scholzj, line 65: Should this be in ResourceAnnotations class in `api` module?

Edit: It is there.

CHANGELOG.md
```diff
@@ -6,6 +6,7 @@
 * Generate and publish Software Bill of Materials (SBOMs) of Strimzi containers
 * Add support for stopping connectors according to [Strimzi Proposal #54](https://github.com/strimzi/proposals/blob/main/054-stopping-kafka-connect-connectors.md)
 * Make sure brokers are empty before scaling them down
+* Add support for pausing `KafkaTopic` reconciliations with UTO
```
scholzj, line 9: I would use here full name and not the `UTO` shortcut.

topic-operator/src/main/java/io/strimzi/operator/topic/v2/BatchingTopicController.java
```diff
@@ -112,6 +113,13 @@ public class BatchingTopicController {
                 || !"false".equals(kt.getMetadata().getAnnotations().get(MANAGED));
     }
 
+    /* test */ static boolean isPaused(KafkaTopic kt) {
+        return kt.getMetadata() != null
+            && kt.getMetadata().getAnnotations() != null
+            && kt.getMetadata().getAnnotations().get(PAUSED) != null
+            && "true".equals(kt.getMetadata().getAnnotations().get(PAUSED));
```
scholzj, line 120:
```suggestion
        return Annotations.isReconciliationPausedWithAnnotation(kt);
```

Or maybe even better ... remove the method and use the call directly instead.

topic-operator/src/main/java/io/strimzi/operator/topic/v2/BatchingTopicController.java
```diff
@@ -961,25 +974,39 @@ private void updateStatusForException(ReconcilableTopic reconcilableTopic, Excep
...
         if (oldStatus == null
                 || oldStatus.getObservedGeneration() != kt.getMetadata().getGeneration()
                 || oldStatus.getTopicName() == null
                 || oldReadyCondition == null
-                || isDifferentCondition(oldReadyCondition, newReason, newMessage, newStatus)) {
+                || isDifferentCondition(oldReadyCondition, condition)
+                || isPaused(kt)) {
+            long observedGeneration = oldStatus != null
+                ? !isPaused(kt) ? kt.getMetadata().getGeneration() : oldStatus.getObservedGeneration()
+                : !isPaused(kt) ? kt.getMetadata().getGeneration() : 0L;
```
scholzj, line 1009: I'm not sure I follow this logic. Observed generation indicates that the operator has seen the latest change and processed it. That does not in any way change if it is paused -> the change is still processed and handled accordingly (by ignoring it in the case of paused reconciliation).

fvaleri: Yes, maybe this deserve a comment. Observing the UO I found that the generation is initialized to 0 when creating a paused user (the annotation is already there on creation). This is what the last line does (oldStatus null and paused true).

topic-operator/src/test/java/io/strimzi/operator/topic/v2/TopicControllerIT.java
```diff
@@ -1904,4 +1955,27 @@ public void shouldTerminateIfQueueFull(
         operator = null;
         maybeStartOperator(topicOperatorConfig("ns", kafkaCluster));
     }
-}
\ No newline at end of file
+
+    @Test
+    public void shouldSkipReconciliationsWhenKafkaTopicIsPaused(
```
tombentley, line 1960: We should add tests that deleted paused CRs do get reconciled and one covering CRs created in the paused state.

topic-operator/src/test/java/io/strimzi/operator/topic/v2/TopicControllerIT.java
```diff
@@ -421,6 +448,30 @@ private KafkaTopic createTopic(KafkaCluster kc, KafkaTopic kt) throws ExecutionE
...
+            .editMetadata()
+                .withAnnotations(ANNO_PAUSE_TRUE)
+            .endMetadata()
+            .build()).update();
+        LOGGER.info("Test paused KafkaTopic {} with resourceVersion {}",
+            paused.getMetadata().getName(), BatchingTopicController.resourceVersion(paused));
+        waitUntil(paused, pausedIsTrue());
+    }
+
+    private void updateTopicConfig(String namespace, String topicName, Map<String, Object> configs) {
```
tombentley, line 463: Can we not use the existing `modifyTopicAndAwait()` helper?

topic-operator/src/main/java/io/strimzi/operator/topic/v2/BatchingTopicController.java
```diff
@@ -961,25 +971,39 @@ private void updateStatusForException(ReconcilableTopic reconcilableTopic, Excep
...
         var oldStatus = kt.getStatus();
         Condition oldReadyCondition = oldStatus == null || oldStatus.getConditions() == null ? null : oldStatus.getConditions().stream().findFirst().orElse(null);
-        String newStatus = newReason != null || newMessage != null ? "False" : "True";
         if (oldStatus == null
                 || oldStatus.getObservedGeneration() != kt.getMetadata().getGeneration()
                 || oldStatus.getTopicName() == null
                 || oldReadyCondition == null
-                || isDifferentCondition(oldReadyCondition, newReason, newMessage, newStatus)) {
+                || isDifferentCondition(oldReadyCondition, condition)
+                || isPaused(kt)) {
```
tombentley, line 1003: Won't this mean we do the update every time when a CR is paused? Isn't the `isDifferentCondition()` check enough to mean we will update the status when the annotation is paused/unpaused?

topic-operator/src/main/java/io/strimzi/operator/topic/v2/BatchingTopicController.java
```diff
@@ -418,12 +423,17 @@ private void updateInternal(List<ReconcilableTopic> topics) {
...
-        var partitionedByManaged = partitionedByDeletion.get(false).stream().collect(Collectors.partitioningBy(reconcilableTopic -> isManaged(reconcilableTopic.kt())));
+        var partitionedByManaged = partitionedByDeletion.get(false).stream()
+            .collect(Collectors.partitioningBy(reconcilableTopic -> isManaged(reconcilableTopic.kt())));
         var unmanaged = partitionedByManaged.get(false);
         addOrRemoveFinalizer(useFinalizer, unmanaged).forEach(rt -> putResult(results, rt, Either.ofRight(null)));
 
-        List<ReconcilableTopic> mayNeedUpdate = validateManagedTopics(partitionedByManaged);
+        // skip reconciliation of paused KafkaTopics
+        var partitionedByPaused = validateManagedTopics(partitionedByManaged).stream()
+            .collect(Collectors.partitioningBy(reconcilableTopic -> isPaused(reconcilableTopic.kt())));
```
tombentley, line 433: We've now got three sequential `partitioningBy()`s going on. I wonder if the code would be simpler if we just did a `groupingBy()` instead?

topic-operator/src/test/java/io/strimzi/operator/topic/v2/TopicControllerIT.java
fvaleri: Sure.

topic-operator/src/test/java/io/strimzi/operator/topic/v2/TopicControllerIT.java
fvaleri: Sure, but I need to adapt it to take into account paused state.

topic-operator/src/main/java/io/strimzi/operator/topic/v2/BatchingTopicController.java
fvaleri: I think you are right, but I need to update the type check in `isDifferentCondition()`.

topic-operator/src/main/java/io/strimzi/operator/topic/v2/BatchingTopicController.java
scholzj: I don't remember why we do that in the UO. But yeah, comment explaining what it does might be useful.

fvaleri: I think the reason is that, even in that case, we want to have a status with the observedGeneration, but you don't want to be the same as metadata.generation because the resource is not yet reconciled. I'll add a note.

scholzj (approved):

fvaleri:
@tombentley this is ready for another pass. If it's good, I can then rebase the metrics PR and add the paused reconciliation metric. Thanks.

tombentley (approved):
LGTM

ppatierno (approved):

PR #9168, opened 2023-09-27
Highlight the CA genaration flag status in the CA cert/key replacement doc

ppatierno:
When the user follows this documentation https://strimzi.io/docs/operators/latest/deploying#proc-replacing-your-own-private-keys-str  to replace the CA certificates and private keys generated by the Strimzi operators with his own, we should make clear that the `spec.clusterCa.generateCertificateAuthority` and/or `spec.clientCa.generateCertificateAuthority` need to be set to `false`.


jankalinic (approved):

documentation/modules/security/proc-replacing-your-own-private-keys.adoc
```diff
@@ -52,6 +52,34 @@ kubectl describe Kafka _<name_of_custom_resource>_
 +
 The `type` condition changes to `ReconciliationPaused` at the `lastTransitionTime`.
 
+. Check that your `Kafka` custom resource has the `spec.clusterCa.generateCertificateAuthority` and/or the `spec.clientsCa.generateCertificateAuthority` set to `false`, depending on which CA certificate you are renewing or replacing.
+If the `generateCertificateAuthority` field is on `true` or missing, it means you want to replace the CA certificates and private keys generated by the Strimzi operators with your own.
```
PaulRMellor, line 56:
```suggestion
. Check that the `generateCertificateAuthority` properties in your `Kafka` custom resource are set `false`.
+
If the properties are set to `false`, a CA certificate is not generated by the Cluster Operator. 
You require this option is you are using your own certificates.
```

PaulRMellor (approved):
Looks good. I made a couple of suggestions.
I did wonder if this information might also be a prereq, but step is fine.
I suggested to make the optional config change a separate step.

documentation/modules/security/proc-replacing-your-own-private-keys.adoc
ppatierno: > Check that the `generateCertificateAuthority` properties in your `Kafka` custom resource are set `false`

I was trying to make the point that not both should be set to false, it depends which CA you are replacing. It could be cluster, clients or both. With your suggestion I read as ALL of them has to be set to false.

> If the properties are set to `false`, a CA certificate is not generated by the Cluster Operator. 

Even if the field is totally missing.

PaulRMellor: I've made a change^^

scholzj (approved):

PR #9169, opened 2023-09-27
Fixing logging issue around KafkaRebalanceAssemblyOperator

ShubhamRwt:
### Type of change

_Select the type of your PR_

- Task

### Description

Currently what happen is that if  there is no state change and annotation is applied on resource, then we get the logs as
```
2023-07-10 08:20:49 INFO  KafkaRebalanceAssemblyOperator:475 - Reconciliation #39(kafkarebalance-watch) KafkaRebalance(myproject/my-rebalance):  with annotation PendingProposal=strimzi.io/rebalance applied on the KafkaRebalance resource
```
which is incorrect. I have updated the code to fix this problem

### Checklist

- [ ] Write tests
- [ ] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceAssemblyOperator.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceAssemblyOperator.java
@@ -437,11 +437,10 @@ public class KafkaRebalanceAssemblyOperator
                                                 kafkaRebalance.getMetadata().getName(), desiredStatusAndMap.getLoadMap())
                                                 .compose(i -> updateStatus(reconciliation, currentKafkaRebalance, desiredStatusAndMap.getStatus(), null))
                                                 .compose(updatedKafkaRebalance -> {
-                                                    String message;
                                                     if (currentKafkaRebalance.getStatus() != null
                                                             && updatedKafkaRebalance.getStatus() != null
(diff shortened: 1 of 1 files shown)


cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceAssemblyOperator.java
```diff
@@ -437,18 +437,18 @@ && rawRebalanceAnnotation(kafkaRebalance) == null) {
                                                 kafkaRebalance.getMetadata().getName(), desiredStatusAndMap.getLoadMap())
                                                 .compose(i -> updateStatus(reconciliation, currentKafkaRebalance, desiredStatusAndMap.getStatus(), null))
                                                 .compose(updatedKafkaRebalance -> {
-                                                    String message = "";
+                                                    String message;
```
scholzj, line 440: Why does the `message` obejct need to be defined outside of the `if` scope? It seems to be used only there.

ppatierno: Agree. It was this way before because we had two different `if` at the same level using `message` but now it can be moved inside.

ShubhamRwt: Done

ppatierno (approved):
LGTM

scholzj (approved):

PR #9183, opened 2023-09-30
External IP Configurability for Kafka brokers and External Bootstrap

rao2100:
### Type of change

Enhancement

### Description

Cluster operator does not provide any configurability of the external IPs associated with the nodeport. In other words, user is not able to provide any external IP. We have tried to perform a kubectl patch with the external IP but the setting is not permanent. After a while, the change will be reverted and external IP is gone. To address this, we updated the CRD template with a new option where user can provide external IP for the broker and IP for each of the nodeports. The cluster operator is also updated to retrieve the user provided values and provision them during the init container step.

#### Sample Configuration
```
  external:
    type: nodeport
    tls: false
    overrides:
      bootstrap:
        advertisedHost: 107.250.139.224
        nodePort: 32100
      brokers:
      - broker: 0
        advertisedHost: 107.250.139.225
        nodePort: 32000
      - broker: 1
        advertisedHost: 107.250.139.226
        nodePort: 32001
      - broker: 2
        advertisedHost: 107.250.139.227
        nodePort: 32002
template:
  externalBootstrapService:
    externalTrafficPolicy: Local
    externalIPs:
      - 107.250.139.224
  perPodService:
    externalTrafficPolicy: Local
    externalIPs:
      - 107.250.139.225
      - 107.250.139.226
      - 107.250.139.227
```
```
$ kubectl get svc
NAME                                  TYPE        CLUSTER-IP       EXTERNAL-IP                       PORT(S)                      AGE
my-cluster-kafka-0                    NodePort    10.103.116.212   107.250.139.225                   9094:32000/TCP               107s
my-cluster-kafka-1                    NodePort    10.104.170.243   107.250.139.226                   9094:32001/TCP               107s
my-cluster-kafka-2                    NodePort    10.110.163.100   107.250.139.227                   9094:32002/TCP               107s
my-cluster-kafka-bootstrap            ClusterIP   10.107.234.220   <none>                            9091/TCP,9093/TCP            107s
my-cluster-kafka-brokers              ClusterIP   None             <none>                            9091/TCP,9093/TCP            107s
my-cluster-kafka-external-bootstrap   NodePort    10.102.65.138    107.250.139.224                   9094:32100/TCP               107s
my-cluster-zookeeper-client           ClusterIP   10.106.192.45    <none>                            2181/TCP                     2m9s
my-cluster-zookeeper-nodes            ClusterIP   None             <none>                            
```

### Checklist

- [ ] Write tests
- [ ] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/api/src/main/java/io/strimzi/api/kafka/model/listener/arraylistener/GenericKafkaListenerConfigurationBootstrap.java
+++ b/api/src/main/java/io/strimzi/api/kafka/model/listener/arraylistener/GenericKafkaListenerConfigurationBootstrap.java
@@ -38,6 +38,7 @@ public class GenericKafkaListenerConfigurationBootstrap implements Serializable,
     private Map<String, String> labels = new HashMap<>(0);
     private Integer nodePort;
     private String loadBalancerIP;
+    private List<String> externalIPs;
 
     private Map<String, Object> additionalProperties = new HashMap<>(0);
 
@@ -109,6 +110,16 @@ public class GenericKafkaListenerConfigurationBootstrap implements Serializable,
 
     public void setLoadBalancerIP(String loadBalancerIP) {
         this.loadBalancerIP = loadBalancerIP;
+    }
+
+    @Description("External IPs to the resource.")
+    @JsonInclude(JsonInclude.Include.NON_EMPTY)
+    public List<String> getExternalIPs() {
+        return externalIPs;
+    }
+
+    public void setExternalIP(List<String> externalIPs) {
+        this.externalIPs = externalIPs;
     }
 
     @Override
--- a/api/src/main/java/io/strimzi/api/kafka/model/listener/arraylistener/GenericKafkaListenerConfigurationBroker.java
+++ b/api/src/main/java/io/strimzi/api/kafka/model/listener/arraylistener/GenericKafkaListenerConfigurationBroker.java
@@ -16,6 +16,7 @@ import lombok.EqualsAndHashCode;
 
 import java.io.Serializable;
 import java.util.HashMap;
+import java.util.List;
 import java.util.Map;
 
 import static java.util.Collections.emptyMap;
@@ -42,6 +43,7 @@ public class GenericKafkaListenerConfigurationBroker implements Serializable, Un
     private Map<String, String> labels = new HashMap<>(0);
     private Integer nodePort;
     private String loadBalancerIP;
+    private List<String> externalIPs;
 
     private Map<String, Object> additionalProperties = new HashMap<>(0);
 
@@ -133,6 +135,16 @@ public class GenericKafkaListenerConfigurationBroker implements Serializable, Un
 
     public void setLoadBalancerIP(String loadBalancerIP) {
         this.loadBalancerIP = loadBalancerIP;
+    }
+
+    @Description("External IPs to the resource.")
+    @JsonInclude(JsonInclude.Include.NON_EMPTY)
+    public List<String> getExternalIPs() {
+        return externalIPs;
+    }
+
+    public void setExternalIPs(List<String> externalIPs) {
+        this.externalIPs = externalIPs;
     }
 
     @Override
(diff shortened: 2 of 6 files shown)


scholzj:
Thanks for the PR. However, I think it would be good to:
* Explain why is this useful and what is your use-case. I do not think it is really clear why do you need to set any external IPs. The NodePorts in general listen on the nodes as far as I know and do not care where the traffic comes from. It also seems like the external IPs have little meaning without connection to the advertised addresses etc. So I think you need to explain that.
* Why should the API change be as you suggest it. I would expect the listener configuration to be the right place instead.

scholzj:
Plus your code seems to do something completely different from the description (and is missing any tests, but that can be addressed once it is clear why is this useful).

scholzj:
Triaged on 5th October: We should leave more time to @rao2100 to reply.

rao2100:
@scholzj  please give me a bit more time for me to reply.

rao2100:
Hi @scholzj 

Explain why is this useful and what is your use-case. I do not think it is really clear why do you need to set any external IPs. The NodePorts in general listen on the nodes as far as I know and do not care where the traffic comes from. It also seems like the external IPs have little meaning without connection to the advertised addresses etc. So I think you need to explain that.
- We are allocated a list of external IPs by the cluster admin which are routable thru the entire network. These IPs are used by clients external to the cluster to access the kafka brokers. These are routed using calico and bgp to the cluster node.

Why should the API change be as you suggest it. I would expect the listener configuration to be the right place instead.
- We did not  explore changing the listener configuration, we will explore into this and create a new pull request. 

Thank you

rao2100:
@scholzj  
Latest configuration:

`
kafka:
  replicas: 1
  config:
  listeners:
  - name: plain
    port: 9092
    type: internal
    tls: false 
  - name: external
    port: 9094
    type: nodeport
    tls: false
    configuration:
      preferredNodePortAddressType: ExternalIP
      bootstrap:        
        nodePort: 32100
        externalIPs:
        - 10.15.32.119  
      brokers:         
      - broker: 0        
        nodePort: 32000
        advertisedHost: 10.15.32.119
        externalIPs:
          - 10.15.32.119
  template:  
    externalTrafficPolicy: Local

zookeeper:
  replicas: 1
`

scholzj:
This is how I would expect the API to look like, yes.

The point I don't understand is why the _external IPs_ matter. When you create a NodePort service, it will listen to some port on the worker node. And you can front it with any network equipment and just tell it to route to the port on one of the node addresses. So for example create some load balancer or some kind of proxy, open a port on it and configure it to route the traffic from this port to the node port on the worker node address. And from there, Kube will route it to the Pod. The node port normally does not care about what is the external address of the proxy or the load balancer. It will just take it and forward it. So why do you need to set there the external IP address?

rao2100:
@scholzj 

You are correct that for many scenarios a NodePort service without specifying an ExternalIP would suffice, however, there are certain situations where specifying an ExternalIP might be necessary or beneficial.

For example when you’re running a bare-metal Kubernetes cluster without a load balancer like MetalLB, e.g. when using Calico that is able to talk BGP like MetalLB but with additional features that MetalLB does not provide, Calico however does not support LoadBalancer service types and as a result we use NodePort services.

While NodePort services can often be used without specifying an ExternalIP, there are scenarios where using an ExternalIP can provide additional flexibility or meet specific requirements.

scholzj:
Ok, fair enough. In that case, for a start, you should:
* Fix the DCO signoff. Instructions should be under the Details link.
* Add a validation that this field is used only with the correct listener types and an error is thrown otherwise
* Add some unit test for the validation and for the KafkaCLuster class

rao2100:
Will do that @scholzj

scholzj:
@rao2100 Any plans to get back to it? If you don't plan to get back to it anytime soon, the best way might be to close and open a new PR when it is ready. Thanks.

rao2100:
Hi @scholzj , please give by this week. If not will open a new PR like you have suggested.

rao2100:
will create a new pull request with the requested changes. we are testing the changes internally now.

scholzj:
@rao2100 Thanks, so I take we can close this one if you open a new PR?

PR #9216, opened 2023-10-06
Add operation metrics to the Unidirectional Topic Operator

fvaleri:
The UTO logs some timer metrics for the different Kubernetes and Kafka operations. They are not much useful in logs, so this change proposes to add them to the main set of metrics. There is the possibility that the design will evolve to a batch per Kubernetes event type, but I think we will still be interested in these aggregations.

- addFinalizer: strimzi.add.finalizer.duration
- removeFinalizer: strimzi.remove.finalizer.duration
- updateStatus: strimzi.update.status.duration
- incrementalAlterConfigs: strimzi.alter.configs.duration
- createPartitions: strimzi.create.partitions.duration
- createTopics: strimzi.create.topics.duration
- describeTopics: strimzi.describe.topics.duration
- describeConfigs: strimzi.describe.configs.duration
- deleteTopics: strimzi.delete.topics.duration
- listPartitionReassignments: strimzi.list.reassignments.duration

I'm also attaching the resulting metrics produced by some parallel load with default settings (consumed resources: CPU 150m, memory 457Mi). 

[metrics.txt](https://github.com/strimzi/strimzi-kafka-operator/files/12831900/metrics.txt)


scholzj:
What is the expected value of these metrics?

fvaleri:
> What is the expected value of these metrics?

It is supposed to help with troubleshooting. It's similar to the Kafka request metrics where, for each request type, you can have the time spent on various operations (request queue time, local time, remote time, throttle time, response queue time). In case of a reconciliation slowdown, you can see if the problem is with Kubernetes or Kafka, and if it is a general problem, or only limited to a specific operation.

fvaleri:
> I think tracing would be much more useful for this than metrics.

I think metrics and traces are two complementary tools (observability: logs + metrics + traces). Metrics are aggregated, and you can also set alarms.

ppatierno:
I don't see them as something for tracing ... if we are talking about distributed tracing imho. At the same time, as not being an expert on the UTO, I am not sure how much these values can help with troubleshooting. 
And we have them, does it make sense to have a dashboard showing the data?

fvaleri:
@tombentley this is not urgent, but please have a look when you have some time. Thanks.

tombentley:
Can we tune the buckets? Right now the smallest bucket is `le="1.0"`, but the vast majority of operations take less than this so there isn't much useful resolution. Meanwhile at the other end the difference between 120s and 300s doesn't seem like a distinction worth making for must clusters. Those high end buckets might be useful operations which might take a while (`createTopics`), but not so much for `describeTopics` or `describeConfigs` which should be served from cache.

fvaleri:
Hi @tombentley. I rebased and added a Timer type metric with fine grained histogram buckets called `internalTimer`. We can use it to measure any internal operation with much lower resolution than the default timer (10ms, 20ms, 50ms, 100ms, 500ms, 1s, 5s). Let me know if this is what you intended. Attached you can find a new load test result with updated metrics.

[metrics.txt](https://github.com/strimzi/strimzi-kafka-operator/files/12927839/metrics.txt)

operator-common/src/main/java/io/strimzi/operator/common/metrics/FineGrainedTimerFilter.java
```diff
@@ -0,0 +1,43 @@
...
+package io.strimzi.operator.common.metrics;
+
+import io.micrometer.core.instrument.Meter;
+import io.micrometer.core.instrument.config.MeterFilter;
+import io.micrometer.core.instrument.distribution.DistributionStatisticConfig;
+
+import java.time.Duration;
+
+/**
+ * Timer filter that can be used to set a fine grained histogram buckets.
```
tombentley, line 14:
```suggestion
 * Timer filter that can be used get finer-grained histogram buckets 
  * by overriding the configuration of meters with the given name.
```

operator-common/src/main/java/io/strimzi/operator/common/metrics/MetricsHolder.java
```diff
@@ -242,4 +242,22 @@ protected static AtomicInteger getGauge(String namespace, String kind, String me
...
+     * @param kind              Kind of the resource
+     * @param metricName        Name of the metric
+     * @param metrics           Metrics provider
+     * @param selectorLabels    Selector labels used to filter the resources
+     * @param timerMap          Map with timers
+     * @param metricHelp        Help description of the metric
+     *
+     * @return  Timer metric
+     */
+    protected static Timer getInternalTimer(String namespace, String kind, String metricName, MetricsProvider metrics, Labels selectorLabels, Map<String, Timer> timerMap, String metricHelp) {
```
tombentley, line 260: Again I would question the name.

tombentley (approved):
LGTM. Thanks @fvaleri.

operator-common/src/main/java/io/strimzi/operator/common/MicrometerMetricsProvider.java
```diff
@@ -74,11 +75,32 @@ public Counter counter(String name, String description, Tags tags) {
...
         return Timer.builder(name)
                 .description(description)
-                .sla(Duration.ofMillis(1000), Duration.ofMillis(5000), Duration.ofMillis(10000), Duration.ofMillis(30000), Duration.ofMillis(60000), Duration.ofMillis(120000), Duration.ofMillis(300000))
+                .sla(Duration.ofMillis(1000),
+                    Duration.ofMillis(5000),
+                    Duration.ofMillis(10000),
+                    Duration.ofMillis(30000),
+                    Duration.ofMillis(60000),
+                    Duration.ofMillis(120000),
+                    Duration.ofMillis(300000))
```
ppatierno, line 84: when I did something similar for the canary component, I made the buckets configurable. Doesn't it make sense having the same here?

ppatierno (commented):
I left a comment. Still not sure what's the use of it for troubleshooting. I also asked if having a Grafana dashboard would be helpful to users.

operator-common/src/main/java/io/strimzi/operator/common/MicrometerMetricsProvider.java
fvaleri: That's a good point, which was also raised by Tom. Let me work on that.

fvaleri:
> Still not sure what's the use of it for troubleshooting. 

I explained that in https://github.com/strimzi/strimzi-kafka-operator/pull/9216#issuecomment-1750856601. We can also use them to measure the impact of any change to one or more of there internal operations.

> I also asked if having a Grafana dashboard would be helpful to users.

Yes. I think we need to review the dashboard and maybe update in a dedicated PR, as there are a couple of new metrics related to the new batch implementation. I would not create a dashboard for the metrics included here, unless there is demand from users.

scholzj (commented):
Should all the stuff really reside in the `operator-common`? Shouldn't it be in the Topic Operator? I do not think there is any value for it anywhere else, so it should not be in the shared codebase.

fvaleri:
> Should all the stuff really reside in the `operator-common`? Shouldn't it be in the Topic Operator? I do not think there is any value for it anywhere else, so it should not be in the shared codebase.

Sorry, which stuff? MicrometerMetricsProvider is used in all operators.

scholzj:
> > Should all the stuff really reside in the `operator-common`? Shouldn't it be in the Topic Operator? I do not think there is any value for it anywhere else, so it should not be in the shared codebase.
> 
> Sorry, which stuff? MicrometerMetricsProvider is used in all operators.

All the timer changes used only by the UTO.

fvaleri:
> All the timer changes used only by the UTO.

I think you mean fineGrainedTimer. This can potentially be used by every operator/component, so the common module is the right place IMO.

scholzj:
> > All the timer changes used only by the UTO.
> 
> I think you mean fineGrainedTimer. This can potentially be used by every operator/component, so the common module is the right place IMO.

Everything might be used everywhere potentially. If it actually is used in other operators, it can be moved there. But I do not see it right now given you failed to convince people about its usefulness.

fvaleri:
> you failed to convince people about its usefulness

Well, at least I convinced @tombentley ;)

We use batch operations here, so it is critical to be able to monitor their performance when doing code/configuration changes or when switching to new Kafka releases.

scholzj:
> > you failed to convince people about its usefulness
> 
> Well, at least I convinced @tombentley ;)
> 
> We use batch operations here, so it is critical to be able to monitor their performance when doing code/configuration changes, the environment changes, or when switching to new Kafka releases.

Look, I'm skeptical about how useful it would be in practice and whether anyone would use it. AFAIK, nobody so far did even any basic performance testing to answer questions from users about how many KafkaTopic resources it can handle. So I want to see how someone uses these metrics in every release to test it.

But I do not have a major problem with adding this to the UTO - let's add it and see how it works and is used. But let's not pretend that there is some wide agreement to add this tomorrow to all operators and that is why it should be in operator-common.

fvaleri:
Do we want to make these additional metrics opt-in, considering they were originally at trace?

scholzj:
I think that depends on the performance cost they have.

ppatierno:
I am fine with adding these metrics. You and Tom know the values they bring better than me.
But I agree with Jakub about not having this fine grained timer definition in the operator common.
Don't see a use case right now where it could be reused. I would move to common only if needed in the future imho.

fvaleri:
Thanks all for the feedback. I rebased, moved the fine grained timer stuff to topic-operator module, and added an env var to enable the additional metrics when needed.

tombentley (approved):

ppatierno (approved):

ppatierno:
@fvaleri can you fix the conflicts, please?

@scholzj because you had some comments, there is anything else you would like to add here?

scholzj:
@ppatierno I do not have anything more if you and Tom are fine. But it cannot be merged as it has conflicts.

fvaleri:
@ppatierno rebased.

PR #9218, opened 2023-10-06
Refactoring broker config generation method

ppatierno:
Trivial PR to refactor the `KafkaCluster.generatePerBrokerConfiguration` in order to avoid code duplication and even because it's going in the direction of some changes around the needs for the ZooKeeper to KRaft migration implementation.


scholzj (approved):
LGTM if it works.

PR #9221, opened 2023-10-09
[ST] Add UTO to upgrade/downgrade

im-konge:
### Type of change

- Enhancement

### Description

This PR adds UTO to upgrade/downgrade tests and adds methods for handling KafkaTopic's finalizers into the suite.

In case that we delete UTO and after that we are deleting KafkaTopic(s) - this happens for both upgrade & downgrade tests, because there is no cleanup for all the KafkaTopics created in the test, only for the one that is created from the example YAML - the deletion of the KafkaTopics is stuck (because BTO doesn't delete these), or when the Namespace is being deleted, it is stuck in Terminating state (as there are resources with finalizers that are not cleaned up).

This is resolved with `KafkaTopicUtils.setFinalizersInAllTopicsToNull()`, which sets for all KafkaTopics the finalizers to `null` before they are deleted. This method is then used in `NamespaceUtils.deleteNamespaceWithWait()`, where the finalizers are set to `null` in case that the Namespace is stuck and one of the condition in the status is `SomeFinalizersRemain`.

### Checklist

### DIFF
--- a/systemtest/src/main/java/io/strimzi/systemtest/utils/kubeUtils/objects/NamespaceUtils.java
+++ b/systemtest/src/main/java/io/strimzi/systemtest/utils/kubeUtils/objects/NamespaceUtils.java
@@ -5,16 +5,13 @@
 package io.strimzi.systemtest.utils.kubeUtils.objects;
 
 import io.fabric8.kubernetes.api.model.Namespace;
-import io.strimzi.api.kafka.model.KafkaTopic;
 import io.strimzi.systemtest.Constants;
 import io.strimzi.systemtest.resources.ResourceOperation;
-import io.strimzi.systemtest.resources.crd.KafkaTopicResource;
 import io.strimzi.systemtest.utils.kafkaUtils.KafkaTopicUtils;
 import io.strimzi.test.TestUtils;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 
-import static io.strimzi.test.k8s.KubeClusterResource.cmdKubeClient;
 import static io.strimzi.test.k8s.KubeClusterResource.kubeClient;
 
 public class NamespaceUtils {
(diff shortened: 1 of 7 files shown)


systemtest/src/test/java/io/strimzi/systemtest/upgrade/AbstractUpgradeST.java
```diff
@@ -248,12 +245,14 @@ protected void copyModifyApply(File root, String namespace, ExtensionContext ext
     }
 
     protected void deleteInstalledYamls(File root, String namespace) {
+
```
Frawless, line 248: remove?

systemtest/src/main/java/io/strimzi/systemtest/utils/kubeUtils/objects/NamespaceUtils.java
```diff
@@ -26,4 +31,30 @@ public static void waitForNamespaceDeletion(String name) {
...
+        kubeClient().deleteNamespace(namespaceName);
+
+        TestUtils.waitFor("Namespace: " + namespaceName + "to be deleted", Constants.POLL_INTERVAL_FOR_RESOURCE_DELETION, DELETION_TIMEOUT, () -> {
+            Namespace namespace = kubeClient().getNamespace(namespaceName);
+
+            if (namespace == null) {
+                return true;
+            } else if (namespace.getStatus() != null && namespace.getStatus().getConditions() != null) {
+                if (namespace.getStatus().getConditions().stream().anyMatch(condition -> condition.getReason().contains("SomeFinalizersRemain"))) {
+                    LOGGER.info("There are KafkaTopics with finalizers remaining in Namespace: {}, going to set those finalizers to null", namespaceName);
```
Frawless, line 53: Maybe move it to debug level?

Frawless (approved):
Just two nits, otherwise lgtm

henryZrncik (approved):
LGTM, good job 👍

jankalinic (approved):

PR #9197, opened 2023-10-03
[ST] Add UnidirectionalTopicOperator to FG pipelines in Azure

im-konge:
### Type of change

- Bugfix

### Description

This PR adds `UnidirectionalTopicOperator` to FG pipelines in Azure, as it was missing there and it was only present in the KRaft pipelines.

### Checklist

- [ ] Make sure all tests pass

PR #9224, opened 2023-10-09
[ST] Fix testRecoveryDuringKafkaRollingUpdate

im-konge:
### Type of change

- Bugfix

### Description

In #9197 I accidentally broke the `testRecoveryDuringKafkaRollingUpdate` for BTO because of number of replicas that I'm creating.

This PR reverts those changes and applies different solution for the issue -> creating one consumer before we shut down one node (so the `__consumer_offsets` topic will be successfully created), and then creates another one (with different consumer group), after one of the Kafka Pods is in the `Pending` state.

### Checklist


scholzj (approved):
LGTM assuming the tests pass.

Frawless (approved):

PR #9227, opened 2023-10-10
add command to set the KRaft mode in the operator

ngtszman:
### Type of change

_Select the type of your PR_

- Documentation

### Description

make more clear how to enable KRaft mode (zookeeper-less) 

### Checklist

- [Y] Write tests
- [Y] Make sure all tests pass
- [Y] Update documentation

### DIFF
--- a/examples/kafka/nodepools/README.md
+++ b/examples/kafka/nodepools/README.md
@@ -22,7 +22,12 @@ The [`kafka-with-kraft.yaml`](./kafka-with-kraft.yaml) deploys a Kafka cluster w
 The [`kafka-with-dual-role-kraft-nodes.yaml`](./kafka-with-dual-role-kraft-nodes.yaml) deploys a Kafka cluster with one pool of KRaft nodes that share the _broker_ and _controller_ roles.
 
 _NOTE: To use this example, you have to enable the `UseKRaft` feature gate in addition to the `KafkaNodePools` feature gate._
-
+```
+kubectl set env deployments/strimzi-cluster-operator STRIMZI_FEATURE_GATES=+UseKRaft,+KafkaNodePools -n kafka
+```
+```
+kubectl apply -f kafka-with-kraft.yaml -n kafka
+```
 Please be aware that ZooKeeper-less Kafka is still a work in progress and is still missing many features.
 For example:
 * The controller-only nodes are currently not rolled by Strimzi when their configuration changes due to the limitations of the Kafka Admin API. 


examples/kafka/nodepools/README.md
```diff
@@ -22,7 +22,12 @@ The [`kafka-with-kraft.yaml`](./kafka-with-kraft.yaml) deploys a Kafka cluster w
 The [`kafka-with-dual-role-kraft-nodes.yaml`](./kafka-with-dual-role-kraft-nodes.yaml) deploys a Kafka cluster with one pool of KRaft nodes that share the _broker_ and _controller_ roles.
 
 _NOTE: To use this example, you have to enable the `UseKRaft` feature gate in addition to the `KafkaNodePools` feature gate._
-
+```
+kubectl set env deployments/strimzi-cluster-operator STRIMZI_FEATURE_GATES=+UseKRaft,+KafkaNodePools -n kafka
```
fvaleri, line 26:
```suggestion
kubectl set env deployments/strimzi-cluster-operator STRIMZI_FEATURE_GATES=+UseKRaft,+KafkaNodePools -n <NAMESPACE>
```

examples/kafka/nodepools/README.md
```diff
@@ -22,7 +22,12 @@ The [`kafka-with-kraft.yaml`](./kafka-with-kraft.yaml) deploys a Kafka cluster w
...
+```
```
fvaleri, line 27: There is no need to create two code blocks.
```suggestion
```

fvaleri (commented):
Thanks for the PR. Left some comments.

I wonder if we should also include the `UnidirectionalTopicOperator` feature gate, adding the topic operator to all KRaft examples in this folder.

scholzj (commented):
Thanks for the PR. However, this is a bit oversimplified:
* The operator might be installed using OPeratorHUb or Helm in which case the configuration is completely different.
* The operator might run in a different namespace in which case this will not work.
* And some other situations like this

So this is intentionally left open because you need to figure out what is the right way to do it for you.

There are also several examples and not just `kafka-with-kraft.yaml` so that does not always apply either.

scholzj (commented):
Discussed on the Community call of 19.10.2023: We do not really want to have the specific instructions in the README file. Can you please replace it with a link to the relevant documentation section talking about feature gates and how to enable them?

scholzj:
Discussed on the community call on 30.11.2023: There were no changes t this since October. We should clsoe this and it can be reopened again if needed.

Issue #9233, opened 2023-10-11
Address deprecated Vert.x `executeBlocking` calls


scholzj:
Vert.x 4.4.5 deprecated some `executeBlocking` calls we are using. For the time being, we just marked them to be ignored to not break anything shortly before the 0.38.0 release. But they should be addressed after the release. They will be marked with this issue in a comment to make them easy to find.

See https://github.com/vert-x3/wiki/wiki/4.4.5-Deprecations-and-breaking-changes#deprecation-of-execute-blocking-methods-with-a-handler-of-promise for more details.

scholzj:
Most of the work was done in #9250. But some more complicated uses of `executeBlocking` in TO were left for the TO experts and will be done in another PR.

PR #9234, opened 2023-10-11
Update and clean various dependencies - Vert.x, Jetty, Scala

scholzj:
### Type of change

- Task

### Description

This PR updates the Jetty and Vert.x dependencies before the 0.38.0 release. The Vert.x deprecations to various `executeBlocking` methods are only suppressed in this PR. They will be addressed as part of #9233 after the 0.38 release to avoid any issues.

It also removes the direct dependency on Scala -> we do not seem to use it anywhere as a direct dependency (it will be still a transitive dependency in places where Kafka core JAR is used.)

### Checklist

### DIFF
--- a/cluster-operator/pom.xml
+++ b/cluster-operator/pom.xml
@@ -241,11 +241,6 @@
             <classifier>linux-x86_64</classifier>
         </dependency>
         <dependency>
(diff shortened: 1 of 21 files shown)


ppatierno (approved):
LGTM

PR #9238, opened 2023-10-11
Update ZooKeeper to 3.8.3

scholzj:
### Type of change

- Bugfix

### Description

This PR replaces #9237 and updates ZooKeeper to 3.8.3 to avoid CVE-2023-44981. Unlike the original PR from DependaBot, it syncs the minor ZooKeeper version to Kafka 3.6. It also updates the exclusions.

Strimzi is not really affected by CVE-2023-44981 as we do not use SASL Authentication (we use mTLS). Given the past issues with ZooKeeper 3.6.4 (but those were on the server side!), I wonder if we want clear security scanners in operator images in Strimzi 0.38 or prefer more time to catch any possible issues and want to merge it only in 0.39. WDYT @ppatierno?

### Checklist

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -137,7 +137,7 @@
         <kafka.version>3.6.0</kafka.version>
         <yammer-metrics.version>2.2.0</yammer-metrics.version>
         <scala-library.version>2.13.9</scala-library.version>
-        <zookeeper.version>3.6.4</zookeeper.version>
+        <zookeeper.version>3.8.3</zookeeper.version>
         <zkclient.version>0.11</zkclient.version>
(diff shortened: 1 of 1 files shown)


ppatierno:
> I wonder if we want clear security scanners in operator images in Strimzi 0.38

@scholzj I didn't get what you mean here

scholzj:
@ppatierno What I mean is whether we want to merge it in 0.38 or wait with it for 0.39.

ppatierno:
I think merging now should not hurt. I mean the CVE doesn't impact us as you said.

ppatierno (approved):

PR #9257, opened 2023-10-18
Removed unused bash function

ppatierno:
Trivial PR to remove the `get_option_for_broker` bash function which doesn't seem to be used in any scripts.


scholzj (approved):
LGTM assuming it works.

see-quick (approved):

PR #9258, opened 2023-10-18
[ST] - Add support for multi-node kind cluster, update surefire and failsafe plugin version

see-quick:
### Type of change

_Select the type of your PR_

- Enhancement / new feature

### Description

This PR introduces the capability to run on testing farm multi-node clusters using kind. Plus it updates a surefire and failsafe version.

### Checklist

### DIFF
--- a/.azure/scripts/setup-kind.sh
+++ b/.azure/scripts/setup-kind.sh
@@ -100,9 +100,36 @@ function updateDockerDaemonConfiguration() {
     systemctl restart docker
 }
 
+: '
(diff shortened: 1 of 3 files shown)


Frawless (approved):

im-konge (approved):

see-quick:
Just for the context:

We have decided to let this issue be in terms of implementing some post-processing (i.e., modification of TEST XML), which from our POV is not good because in Kafka 3.7.0 probably such an issue will be resolved. Thus this pipeline will work without any problem. So I would merge it :)

PR #9260, opened 2023-10-18
docs(clients): updates the tuning content for producers and consumers

PaulRMellor:
**Documentation**

Some doc updates related to producer and consumer configuration (based on the feedback for the blog posts on developing producer and consumer applications)

- adds an overview of the configuration options for securing access to Kafka
- adds the key configuration considerations for producer and consumer performance
- updates to the producer and consumer tuning content

### Checklist

- [ ] Write tests
- [x] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/documentation/assemblies/security/assembly-securing-access.adoc
+++ b/documentation/assemblies/security/assembly-securing-access.adoc
@@ -7,15 +7,26 @@
 
 [role="_abstract"]
 Secure your Kafka cluster by managing the access a client has to Kafka brokers.
-Specify configuration options to secure Kafka brokers and clients
-
 A secure connection between Kafka brokers and clients can encompass the following:
 
 * Encryption for data exchange
 * Authentication to prove identity
 * Authorization to allow or decline actions executed by users
 
-The authentication and authorization mechanisms specified for a client must match those specified for the Kafka brokers.
+In Strimzi, securing a connection involves configuring listeners and user accounts:
+
+Listener configuration:: Use the `Kafka` resource to configure listeners for client connections to Kafka brokers.
+Listeners define how clients authenticate, such as using TLS, SCRAM-SHA-512, OAuth 2.0, or custom authentication methods.
+To enhance security, configure TLS encryption to secure communication between Kafka brokers and clients.
+You can further secure TLS-based communication by specifying the supported TLS versions and cipher suites in the Kafka broker configuration.
+For an added layer of protection, you can also specify authorization methods in the listener configuration, such as simple, OAuth 2.0, OPA, or custom authorization.
+
+User Accounts:: Set up user accounts and credentials with `KafkaUser` resources in Strimzi. 
+Users represent your clients and determine how they should authenticate and authorize with the Kafka cluster. 
+The authentication and authorization mechanisms specified in the user configuration must match the Kafka configuration. 
+Additionally, define Access Control Lists (ACLs) to control user access to specific topics and actions for more fine-grained authorization.
+You can also add configuration to limit the TLS versions and cipher suites your client uses.
+
 Strimzi operators automate the configuration process and create the certificates required for authentication.
 The Cluster Operator automatically sets up TLS certificates for data encryption and authentication within your cluster.
 
--- a/documentation/assemblies/security/assembly-securing-access.adoc
+++ b/documentation/assemblies/security/assembly-securing-access.adoc
@@ -16,16 +16,22 @@ A secure connection between Kafka brokers and clients can encompass the followin
 In Strimzi, securing a connection involves configuring listeners and user accounts:
 
 Listener configuration:: Use the `Kafka` resource to configure listeners for client connections to Kafka brokers.
-Listeners define how clients authenticate, such as using TLS, SCRAM-SHA-512, OAuth 2.0, or custom authentication methods.
+Listeners define how clients authenticate, such as using mTLS, SCRAM-SHA-512, OAuth 2.0, or custom authentication methods.
 To enhance security, configure TLS encryption to secure communication between Kafka brokers and clients.
 You can further secure TLS-based communication by specifying the supported TLS versions and cipher suites in the Kafka broker configuration.
-For an added layer of protection, you can also specify authorization methods in the listener configuration, such as simple, OAuth 2.0, OPA, or custom authorization.
++
+For an added layer of protection, use the `Kafka` resource to specify authorization methods for the Kafka cluster, such as simple, OAuth 2.0, OPA, or custom authorization.
 
-User Accounts:: Set up user accounts and credentials with `KafkaUser` resources in Strimzi. 
+User accounts:: Set up user accounts and credentials with `KafkaUser` resources in Strimzi. 
 Users represent your clients and determine how they should authenticate and authorize with the Kafka cluster. 
 The authentication and authorization mechanisms specified in the user configuration must match the Kafka configuration. 
 Additionally, define Access Control Lists (ACLs) to control user access to specific topics and actions for more fine-grained authorization.
-You can also add configuration to limit the TLS versions and cipher suites your client uses.
+To further enhance security, specify user quotas to limit client access to Kafka brokers based on byte rates or CPU utilization.
++
+You can also add producer or consumer configuration to your clients if you wish to limit the TLS versions and cipher suites they use.
+The configuration on the clients must only use protocols and cipher suites that are enabled on the broker.
+
+NOTE: If you are using an OAuth 2.0 to manage client access, user authentication and authorization credentials are managed through the authorization server.
 
 Strimzi operators automate the configuration process and create the certificates required for authentication.
 The Cluster Operator automatically sets up TLS certificates for data encryption and authentication within your cluster.
--- a/documentation/modules/managing/con-consumer-config-properties.adoc
+++ b/documentation/modules/managing/con-consumer-config-properties.adoc
@@ -69,6 +69,7 @@ but it does mean that there are consumers on standby should one stop functioning
 If you can meet throughput goals with fewer consumers, you save on resources.
 
 Consumers within the same consumer group send offset commits and heartbeats to the same broker.
+The consumer sends heartbeats to the Kafka broker to indicate its activity within the consumer group.
 So the greater the number of consumers in the group, the higher the request load on the broker.
 
 [source,env]
@@ -231,16 +232,20 @@ isolation.level=read_committed <1>
 
 == Recovering from failure to avoid data loss
 
-Kafka provides a rebalance protocol which helps with failure detection and recovery within a consumer group.  
-Two useful properties that aid in this recovery process are `session.timeout.ms` and `heartbeat.interval.ms`.
+In the event of failures within a consumer group, Kafka provides a rebalance protocol designed for effective detection and recovery.
+To minimize the potential impact of these failures, one key strategy is to adjust the `max.poll.records` property to balance efficient processing with system stability. 
+This property determines the maximum number of records a consumer can fetch in a single poll.
+Fine-tuning `max.poll.records` helps to maintain a controlled consumption rate, preventing the consumer from overwhelming itself or the Kafka broker.
 
-The `session.timeout.ms` property specifies the maximum amount of time a consumer in a group can go without sending a heartbeat to the Kafka broker. 
-This heartbeat is an indicator that the consumer is active in the group. 
-If a consumer fails to send a heartbeat within the specified session timeout, Kafka considers it as a failure, and the consumer is marked as inactive. 
+Additionally, Kafka offers advanced configuration properties like `session.timeout.ms` and `heartbeat.interval.ms`. 
+These settings are typically reserved for more specialized use cases and may not require adjustment in standard scenarios.
+
+The `session.timeout.ms` property specifies the maximum amount of time a consumer can go without sending a heartbeat to the Kafka broker to indicate it is active within the consumer group. 
+If a consumer fails to send a heartbeat within the session timeout, it is considered inactive.
 A consumer marked as inactive triggers a rebalancing of the partitions for the topic. 
-Setting this value too low can result in false-positive outcomes, while setting it too high can lead to delayed recovery from failures.
+Setting the `session.timeout.ms` property value too low can result in false-positive outcomes, while setting it too high can lead to delayed recovery from failures.
 
-The `heartbeat.interval.ms` property determines how frequently a consumer sends heartbeats to the Kafka broker. 
+The `heartbeat.interval.ms` property determines how frequently a consumer sends heartbeats to the Kafka broker.    
 A shorter interval between consecutive heartbeats allows for quicker detection of consumer failures. 
 The heartbeat interval must be lower, usually by a third, than the session timeout. 
 Decreasing the heartbeat interval reduces the chance of accidental rebalancing, but more frequent heartbeats increases the overhead on broker resources.
--- a/documentation/modules/managing/con-consumer-config-properties.adoc
+++ b/documentation/modules/managing/con-consumer-config-properties.adoc
@@ -97,8 +97,8 @@ The *range* assignment strategy assigns a range of partitions to each consumer, 
 Alternatively, opt for a *round robin* assignment strategy for equal partition distribution among consumers, which is ideal for high-throughput scenarios requiring parallel processing.
 
 For more stable partition assignments, consider the *sticky* and *cooperative sticky* strategies. 
-Sticky strategies maintain assigned partitions during rebalances, when possible. 
-If a consumer was previously assigned certain partitions, the sticky strategies aim to reassign those same partitions back to the same consumer after a rebalance.
+Sticky strategies aim to maintain assigned partitions during rebalances, when possible. 
+If a consumer was previously assigned certain partitions, the sticky strategies prioritize retaining those same partitions with the same consumer after a rebalance, while only revoking and reassigning the partitions that are actually moved to another consumer.
 Leaving partition assignments in place reduces the overhead on partition movements.
 The cooperative sticky strategy also supports cooperative rebalances, enabling uninterrupted consumption from partitions that are not reassigned.
 
--- a/documentation/modules/managing/con-consumer-config-properties.adoc
+++ b/documentation/modules/managing/con-consumer-config-properties.adoc
@@ -231,7 +231,8 @@ isolation.level=read_committed <1>
 
 == Recovering from failure to avoid data loss
 
-Kafka provides mechanisms to detect and recover from failures within consumer groups. Two useful properties that aid in this recovery process are `session.timeout.ms` and `heartbeat.interval.ms`.
+Kafka provides a rebalance protocol which helps with failure detection and recovery within a consumer group.  
+Two useful properties that aid in this recovery process are `session.timeout.ms` and `heartbeat.interval.ms`.
 
 The `session.timeout.ms` property specifies the maximum amount of time a consumer in a group can go without sending a heartbeat to the Kafka broker. 
 This heartbeat is an indicator that the consumer is active in the group. 
--- a/documentation/modules/managing/con-consumer-config-properties.adoc
+++ b/documentation/modules/managing/con-consumer-config-properties.adoc
@@ -6,10 +6,17 @@
 = Kafka consumer configuration tuning
 
 [role="_abstract"]
-Use a basic consumer configuration with optional properties that are tailored to specific use cases.
-
+Use configuration properties to optimize the performance of Kafka consumers.
+You can use standard Kafka consumer configuration options.
 When tuning your consumers your primary concern will be ensuring that they cope efficiently with the amount of data ingested.
 As with the producer tuning, be prepared to make incremental changes until the consumers operate as expected.
+
+When tuning a consumer, consider the following aspects carefully, as they significantly impact its performance and behavior:
+
+Scaling:: Consumer groups enable parallel processing of messages by distributing the load across multiple consumers. Properly configuring consumer groups can enhance scalability and throughput.
+Message ordering:: A consumer observes messages in a single partition in the same order that they were committed to the broker, which means that Kafka only provides ordering guarantees for messages in a single partition. If absolute ordering within a topic is important, use a single-partition topic. While a single-partition topic allows only one consumer instance, you can leverage a pool of worker threads to create a multi-threaded consumer for more efficient processing. When processing messages for specific entities (like users), you can use the ID (like user ID) as the message key and direct all messages to a single partition. This way, you can maintain message ordering for events specific to individual entities. If a new entity is created, you can create a new topic and migrate its data.
+Offset reset policy:: Setting the appropriate offset policy ensures that the consumer consumes messages from the desired starting point and handles message processing accordingly. The default Kafka reset value is `latest`, which starts at the end of the partition, and consequently means some messages might be missed, depending on the consumer's behavior and the state of the partition. Setting `auto.offset.reset` to `earliest` ensures that when connecting with a `new group.id`, all messages are retrieved from the beginning of the log.
+Securing access:: Implement security measures for authentication, encryption, and authorization by setting up user accounts to xref:assembly-securing-access-{context}[manage secure access to Kafka].
 
 == Basic consumer configuration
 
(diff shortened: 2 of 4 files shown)


documentation/assemblies/security/assembly-securing-access.adoc
```diff
@@ -7,15 +7,26 @@
...
 
 * Encryption for data exchange
 * Authentication to prove identity
 * Authorization to allow or decline actions executed by users
 
-The authentication and authorization mechanisms specified for a client must match those specified for the Kafka brokers.
+In Strimzi, securing a connection involves configuring listeners and user accounts:
+
+Listener configuration:: Use the `Kafka` resource to configure listeners for client connections to Kafka brokers.
+Listeners define how clients authenticate, such as using TLS, SCRAM-SHA-512, OAuth 2.0, or custom authentication methods.
```
scholzj, line 19:
```suggestion
Listeners define how clients authenticate, such as using mTLS, SCRAM-SHA-512, OAuth 2.0, or custom authentication methods.
```

documentation/assemblies/security/assembly-securing-access.adoc
```diff
@@ -7,15 +7,26 @@
...
+To enhance security, configure TLS encryption to secure communication between Kafka brokers and clients.
+You can further secure TLS-based communication by specifying the supported TLS versions and cipher suites in the Kafka broker configuration.
+For an added layer of protection, you can also specify authorization methods in the listener configuration, such as simple, OAuth 2.0, OPA, or custom authorization.
```
scholzj, line 22: Kafka supports authorization only at the cluster level. So this is not `in the listener configuration`.

documentation/assemblies/security/assembly-securing-access.adoc
```diff
@@ -7,15 +7,26 @@
...
+
+User Accounts:: Set up user accounts and credentials with `KafkaUser` resources in Strimzi. 
+Users represent your clients and determine how they should authenticate and authorize with the Kafka cluster. 
+The authentication and authorization mechanisms specified in the user configuration must match the Kafka configuration. 
+Additionally, define Access Control Lists (ACLs) to control user access to specific topics and actions for more fine-grained authorization.
+You can also add configuration to limit the TLS versions and cipher suites your client uses.
```
scholzj, line 28: No, you cannot do this in the user account - or at least not in KafkaUser. You can do this only in the producer / consumer configuration.

documentation/assemblies/security/assembly-securing-access.adoc
```diff
@@ -7,15 +7,26 @@
...
+Listener configuration:: Use the `Kafka` resource to configure listeners for client connections to Kafka brokers.
+Listeners define how clients authenticate, such as using TLS, SCRAM-SHA-512, OAuth 2.0, or custom authentication methods.
+To enhance security, configure TLS encryption to secure communication between Kafka brokers and clients.
+You can further secure TLS-based communication by specifying the supported TLS versions and cipher suites in the Kafka broker configuration.
+For an added layer of protection, you can also specify authorization methods in the listener configuration, such as simple, OAuth 2.0, OPA, or custom authorization.
+
+User Accounts:: Set up user accounts and credentials with `KafkaUser` resources in Strimzi. 
+Users represent your clients and determine how they should authenticate and authorize with the Kafka cluster. 
+The authentication and authorization mechanisms specified in the user configuration must match the Kafka configuration. 
+Additionally, define Access Control Lists (ACLs) to control user access to specific topics and actions for more fine-grained authorization.
```
scholzj, line 27: Should it mention quotas as well?

documentation/assemblies/security/assembly-securing-access.adoc
```diff
@@ -7,15 +7,26 @@
...
-The authentication and authorization mechanisms specified for a client must match those specified for the Kafka brokers.
+In Strimzi, securing a connection involves configuring listeners and user accounts:
+
+Listener configuration:: Use the `Kafka` resource to configure listeners for client connections to Kafka brokers.
+Listeners define how clients authenticate, such as using TLS, SCRAM-SHA-512, OAuth 2.0, or custom authentication methods.
+To enhance security, configure TLS encryption to secure communication between Kafka brokers and clients.
+You can further secure TLS-based communication by specifying the supported TLS versions and cipher suites in the Kafka broker configuration.
+For an added layer of protection, you can also specify authorization methods in the listener configuration, such as simple, OAuth 2.0, OPA, or custom authorization.
+
+User Accounts:: Set up user accounts and credentials with `KafkaUser` resources in Strimzi.
```
scholzj, line 24: Should you mention that Oauth users are configured in the OAuth server?

documentation/modules/managing/con-consumer-config-properties.adoc
```diff
@@ -6,11 +6,18 @@
 = Kafka consumer configuration tuning
 
 [role="_abstract"]
-Use a basic consumer configuration with optional properties that are tailored to specific use cases.
-
+Use configuration properties to optimize the performance of Kafka consumers.
+You can use standard Kafka consumer configuration options.
```
scholzj, line 10: TBH, this sentence seems weird. What are the `standard options`? Is this expected to refer to the options that Kafka supports? Or some subset of important or common options? I would probably leave this out.

documentation/modules/managing/con-consumer-config-properties.adoc
```diff
@@ -6,11 +6,18 @@
...
 When tuning your consumers your primary concern will be ensuring that they cope efficiently with the amount of data ingested.
 As with the producer tuning, be prepared to make incremental changes until the consumers operate as expected.
 
+When tuning a consumer, consider the following aspects carefully, as they significantly impact its performance and behavior:
+
+Scaling:: Consumer groups enable parallel processing of messages by distributing the load across multiple consumers. Properly configuring consumer groups can enhance scalability and throughput.
```
scholzj, line 16: I think this needs to mention partitions. Without them, this is confusing as users might not understand how the consumer groups scale.

documentation/modules/managing/con-consumer-config-properties.adoc
```diff
@@ -6,11 +6,18 @@
...
+Message ordering:: A consumer observes messages in a single partition in the same order that they were committed to the broker, which means that Kafka only provides ordering guarantees for messages in a single partition. If absolute ordering within a topic is important, use a single-partition topic. While a single-partition topic allows only one consumer instance, you can leverage a pool of worker threads to create a multi-threaded consumer for more efficient processing. When processing messages for specific entities (like users), you can use the ID (like user ID) as the message key and direct all messages to a single partition. This way, you can maintain message ordering for events specific to individual entities. If a new entity is created, you can create a new topic and migrate its data.
```
scholzj, line 17: This is super hard to comment on because it is one long line. I would switch the first two sentences. Single partition topic allows only `one consumer instance to consumer from it at a point in time`. But you can have more of them. The threading suggestion seems wrong to me => it makes it sound super easy, but without individual-per-message acks it is super hard to implement. The users as the keys are also a bit oversimplified.

documentation/modules/managing/con-consumer-config-properties.adoc
```diff
@@ -6,11 +6,18 @@
...
+Offset reset policy:: Setting the appropriate offset policy ensures that the consumer consumes messages from the desired starting point and handles message processing accordingly. The default Kafka reset value is `latest`, which starts at the end of the partition, and consequently means some messages might be missed, depending on the consumer's behavior and the state of the partition. Setting `auto.offset.reset` to `earliest` ensures that when connecting with a `new group.id`, all messages are retrieved from the beginning of the log.
```
scholzj, line 18: new `groups.id` instead of `new group.id`.

documentation/modules/managing/con-consumer-config-properties.adoc
```diff
@@ -65,6 +72,21 @@ group.id=my-group-id <1>
...
+Select an appropriate partition assignment strategy, which determines how Kafka topic partitions are distributed among consumer instances in a group.
+
+Specify the strategy using the `partition.assignment.strategy` consumer configuration property. 
+The *range* assignment strategy assigns a range of partitions to each consumer, and is useful when you want to process related data together.
+
+Alternatively, opt for a *round robin* assignment strategy for equal partition distribution among consumers, which is ideal for high-throughput scenarios requiring parallel processing.
+
+For more stable partition assignments, consider the *sticky* and *cooperative sticky* strategies. 
+Sticky strategies maintain assigned partitions during rebalances, when possible. 
+The cooperative sticky strategy also supports cooperative rebalances, enabling uninterrupted consumption from partitions that are not reassigned.
```
scholzj, line 86: Based on what do they maintain the stickiness?

documentation/modules/managing/con-consumer-config-properties.adoc
```diff
@@ -65,6 +72,21 @@ group.id=my-group-id <1>
 ----
 <1> Add a consumer to a consumer group using a group id.
 
+== Choosing the right partition assignment strategy
+
+Select an appropriate partition assignment strategy, which determines how Kafka topic partitions are distributed among consumer instances in a group.
```
scholzj, line 77: Should the actual classes be mentioned here by name? Without that nobody can configure it to anything.

documentation/modules/managing/con-producer-config-properties.adoc
```diff
@@ -6,11 +6,18 @@
...
-Use a basic producer configuration with optional properties that are tailored to specific use cases.
-
+Use configuration properties to optimize the performance of Kafka producers.
+You can use standard Kafka producer configuration options.
 Adjusting your configuration to maximize throughput might increase latency or vice versa.
 You will need to experiment and tune your producer configuration to get the balance you need.
 
+When configuring a producer, consider the following aspects carefully, as they significantly impact its performance and behavior:
+
+Compression:: Implementing message compression can reduce network bandwidth usage, conserving resources and improving throughput.
```
scholzj, line 16: I guess you should clarify what resources does it conserve and what is the cost => network and disk storage migh be conserved, but CPU is the cost.

documentation/modules/managing/con-consumer-config-properties.adoc
fvaleri: Agree with @scholzj. We should say that the number of topic partitions determines the max level of parallelism that you can achieve, because one partition can only be assigned to one consumer in a consumer group. That said, a single consumer can offload message processing to a thread pool when ordering is not mandatory, but this is not supported by the Kafka library.

documentation/modules/managing/con-consumer-config-properties.adoc
```diff
@@ -193,17 +215,17 @@ isolation.level=read_committed <1>
 
 == Recovering from failure to avoid data loss
 
-Use the `session.timeout.ms` and `heartbeat.interval.ms` properties to configure the time taken to check and recover from consumer failure within a consumer group.
-
-The `session.timeout.ms` property specifies the maximum amount of time in milliseconds a consumer within a consumer group can be out of contact with a broker before being considered inactive and a _rebalancing_ is triggered between the active consumers in the group.
-When the group rebalances, the partitions are reassigned to the members of the group.
+Kafka provides mechanisms to detect and recover from failures within consumer groups. Two useful properties that aid in this recovery process are `session.timeout.ms` and `heartbeat.interval.ms`.
```
fvaleri, line 218:
```suggestion
Kafka provides a rebalance protocol which helps with failure detection and recovery within a consumer group. Two useful properties that aid in this recovery process are `session.timeout.ms` and `heartbeat.interval.ms`.
```

documentation/modules/managing/con-producer-config-properties.adoc
```diff
@@ -223,7 +230,7 @@ When your application calls `KafkaProducer.send()` the messages are:
 * Added to a batch of messages in a per-partition queue
 
 At which point the `send()` method returns.
-So the time `send()` is blocked is determined by:
+So the time `send()` is blocked is determined by the following:
```
fvaleri, line 233: The send() operation also blocks when buffer.memory is full or metadata are not available. We should also mention these cases for completeness.

fvaleri (commented):
Hi @PaulRMellor, thanks. I think that, for every tuning recommendation, we should always add the relevant metrics that the user can check to see if the configuration change has the desired effect in their use case. We should also say that it is recommended to change one configuration at a time. You can also try them using the simple example we created for the blog posts.

documentation/assemblies/security/assembly-securing-access.adoc
PaulRMellor: I removed the _listener configuration_ bit

documentation/assemblies/security/assembly-securing-access.adoc
PaulRMellor: I added a space and updated as follows to make this clear:

> You can also add producer or consumer configuration to your clients if you wish to limit the TLS versions and cipher suites they use. 
> The configuration on the client must only use protocols and cipher suites that are enabled on the broker.

documentation/assemblies/security/assembly-securing-access.adoc
PaulRMellor: Makes sense.

> To further enhance security, specify user quotas to limit client access to Kafka brokers based on byte rates or CPU utilization.

documentation/modules/managing/con-consumer-config-properties.adoc
PaulRMellor: - Switched the first two lines
- I think we can live without the threading part here -- removed
- Not sure how much more in-depth we should get for using keys here. I changed the text a bit here

> Message ordering:: 
> If absolute ordering within a topic is important, use a single-partition topic. 
> A consumer observes messages in a single partition in the same order that they were committed to the broker, which means that Kafka only provides ordering guarantees for messages in a single partition. 
> It is also possible to maintain message ordering for events specific to individual entities, such as users.
> If a new entity is created, you can create a new topic dedicated to that entity.
> You can use a unique ID, like a user ID, as the message key and route all messages with the same key to a single partition within the topic.

documentation/modules/managing/con-consumer-config-properties.adoc
PaulRMellor: I updated to mention that stickiness is based on previously assigned partitions.

> For more stable partition assignments, consider the *sticky* and *cooperative sticky* strategies. 
> Sticky strategies maintain assigned partitions during rebalances, when possible. 
> If a consumer was previously assigned certain partitions, the sticky strategies aim to reassign those same partitions back to the same consumer after a rebalance.
> Leaving partition assignments in place reduces the overhead on partition movements.
> The cooperative sticky strategy also supports cooperative rebalances, enabling uninterrupted consumption from partitions that are not reassigned.

scholzj:
> > I left some comments. But TBH, I'm not sure this works for me. It tries to cover too much on very little space and that leads to oversimplification. So I'm not sure fro whom is this is really written. People who know Kafka will know this already. For people who do not know Kafka this does not explain enough and might just confuse them.
> 
> It's a fair point. I think the original reason for adding the tuning content was to give an idea of the common ways to tune the components to satisfy specific use cases. The idea is that a user might want to know which properties would be suitable for _Increasing bandwidth for high latency connections_, for example. Our content provides some tips for those kinds of use cases. I imagine that kind of content is searched for a lot. Maybe we could do better on the information we provide. Federico suggests adding metrics info, for example.

TBH, I'm not sure we want to replace the Kafka docs. Maintaining this and testing this for every release would be a lot of load of work. Its not like a blog post where you can say that it was published a few years ago and might not apply anymore. And at least as my personal opinion - I think there is no way you can make this short and simple and yet understandable. You either need to document how Kafka works and how to write the client which is a lot of initial as well as ongoing work and I would argue world will be better of if that is done in Apache Kafka. Or you keep it short by making it arguably too simple and it would be either not helpful or spawn new questions we do not have capacity to answer.

PaulRMellor:
@mimaison , @showuon  - We're currently discussing maintenance of the (broker, producer, consumer)  tuning content in the Strimzi documentation and your advice would be useful in shaping any decision.

The main points of discussion are as follows:

1. **Maintainability:**  Do we want to keep maintaining this type of content?
2. **Depth:** Is the content sufficiently in-depth? If not, is it better to drop the content rather than invest more time in expanding and maintaining it?
3. **Inclusion in Kafka Documentation:** Shouldn't this type of content be referenced from the Kafka documentation? And if there are gaps in the Kafka documentation, can that be rectified? 

Do you think the tuning content is useful to have in the Strimzi documentation, and if so, how can we address the maintainability concerns? Alternatively, do you agree that it might be more appropriate to have such content in the Kafka documentation?

documentation/modules/managing/con-producer-config-properties.adoc
```diff
@@ -246,15 +251,14 @@ batch.size=16384 <2>
 buffer.memory=33554432 <3>
 # ...
 ----
-<1> The `linger` property adds a delay in milliseconds so that larger batches of messages are accumulated and sent in a request. The default is `0'.`
+<1> The `linger` property adds a delay in milliseconds so that larger batches of messages are accumulated and sent in a request. The default is `0`.
```
mimaison, line 254: `linger` -> `linger.ms`

documentation/modules/managing/con-consumer-config-properties.adoc
```diff
@@ -65,6 +79,30 @@ group.id=my-group-id <1>
 ----
 <1> Add a consumer to a consumer group using a group id.
 
+== Choosing the right partition assignment strategy
+
+Select an appropriate partition assignment strategy, which determines how Kafka topic partitions are distributed among consumer instances in a group.
```
mimaison, line 84: @showuon Should we recommend users to set the strategy to `CooperativeStickyAssignor`? I remember a few issues with it, have they been all solved? 
By default consumers still use `RangeAssignor` which in many cases is literally the worst option.

documentation/modules/managing/con-consumer-config-properties.adoc
```diff
@@ -193,17 +231,18 @@ isolation.level=read_committed <1>
 
 == Recovering from failure to avoid data loss
 
-Use the `session.timeout.ms` and `heartbeat.interval.ms` properties to configure the time taken to check and recover from consumer failure within a consumer group.
-
-The `session.timeout.ms` property specifies the maximum amount of time in milliseconds a consumer within a consumer group can be out of contact with a broker before being considered inactive and a _rebalancing_ is triggered between the active consumers in the group.
-When the group rebalances, the partitions are reassigned to the members of the group.
+Kafka provides a rebalance protocol which helps with failure detection and recovery within a consumer group.  
+Two useful properties that aid in this recovery process are `session.timeout.ms` and `heartbeat.interval.ms`.
```
mimaison, line 235: In most cases users should not have to tweak these settings. They can be used for relatively advanced use cases if there's a good justification, in my experience tweaking these early is often a red flag and instead tuning `max.poll.records` is preferred.

documentation/modules/managing/con-producer-config-properties.adoc
```diff
@@ -246,15 +251,14 @@ batch.size=16384 <2>
...
 <2> If a maximum `batch.size` in bytes is used, a request is sent when the maximum is reached, or messages have been queued for longer than `linger.ms` (whichever comes sooner).
 Adding the delay allows batches to accumulate messages up to the batch size.
-<3> The buffer size must be at least as big as the batch size, and be able to accommodate buffering, compression and in-flight requests.
+<3> The buffer size must be at least as big as the batch size, and be able to accommodate buffering, compression, and in-flight requests.
 
 .Increasing throughput
 
-Improve throughput of your message requests by adjusting the maximum time to wait before a message is delivered and completes a send request.
-
+You can improve throughput of your message requests by using the `delivery.timeout.ms` property to adjust the maximum time to wait before a message is delivered and completes a send request.
```
mimaison, line 261: `delivery.timeout.ms` control the maximum amount of time the producer can take a send records (and get the ack back from the broker). It does not `improve throughput`.

mimaison:
I think this content is fine. It brings some context around the configurations and their interactions. It stays relatively high level to give users an idea of the main concepts. Diving deeper is tricky as it often depends on specific use cases. The Kafka docs on the other side are much more a reference with just the descriptions of the configurations.

PaulRMellor:
> I think this content is fine. It brings some context around the configurations and their interactions. It stays relatively high level to give users an idea of the main concepts. Diving deeper is tricky as it often depends on specific use cases. The Kafka docs on the other side are much more a reference with just the descriptions of the configurations.

Thanks @mimaison. I appreciate your perspective. Our goal is to provide practical insights into best practices and common configurations. Regarding maintenance, the content has remained relatively stable since its creation. However, if you feel there are areas we're not addressing, please let me know.

documentation/modules/managing/con-consumer-config-properties.adoc
PaulRMellor: I'v changed the priority of this section to put the emphasis on `max.poll.records`

mimaison (approved):
LGTM, thanks!
There's just [the comment](https://github.com/strimzi/strimzi-kafka-operator/pull/9260#discussion_r1439674381) about the assignment strategy which I'm not sure whether we could make a better recommendation.

documentation/modules/managing/con-consumer-config-properties.adoc
showuon: Yes, AFAIK, there's a bug related to CooperativeStickyAssignor that could cause duplicated consumption still open: https://issues.apache.org/jira/browse/KAFKA-14224 . But yes, rangeAssignor is not a good default one, but I think that's a safe one. Here, I think it's better we don't recommend any assignor and let users to choose it.

showuon (approved):
LGTM! I also think this doc is helpful. You can check what upstream kafka's doc about [important configs](https://kafka.apache.org/documentation/#clientconfig):

image

documentation/modules/managing/con-consumer-config-properties.adoc
PaulRMellor: Thanks. We don't recommend a strategy and also suggest that a custom strategy might work better, so we can leave the section as it is.

scholzj (approved):
Approving after discussion with @PaulRMellor. If anyone has anymore comments, please share them. Otherwise we are going to merge it.

fvaleri (approved):

documentation/modules/managing/con-consumer-config-properties.adoc
```diff
@@ -65,6 +80,30 @@ group.id=my-group-id <1>
...
+* `org.apache.kafka.clients.consumer.CooperativeStickyAssignor`
+
+Specify a class using the `partition.assignment.strategy` consumer configuration property. 
+The *range* assignment strategy assigns a range of partitions to each consumer, and is useful when you want to process related data together.
+
+Alternatively, opt for a *round robin* assignment strategy for equal partition distribution among consumers, which is ideal for high-throughput scenarios requiring parallel processing.
+
+For more stable partition assignments, consider the *sticky* and *cooperative sticky* strategies. 
+Sticky strategies maintain assigned partitions during rebalances, when possible. 
+If a consumer was previously assigned certain partitions, the sticky strategies aim to reassign those same partitions back to the same consumer after a rebalance.
```
ppatierno, line 101: IIRC it's not going to "reassign" in the sense that it revokes and reassigns, leaving the consumer not consuming until the reassignment happens. IIRC it doesn't revoke the partitions which should stick with that consumer. Just revokes the ones which will be moved to another consumer. @mimaison can you confirma that? If yes, @PaulRMellor could we make it clearer here?

ppatierno (approved):
Just a nit. LGTM.

documentation/modules/managing/con-consumer-config-properties.adoc
PaulRMellor: That's a fair point. I don't think reassign is a good way of wording this. I've updated here:  [docs(review): edits to doc from review by PP](https://github.com/strimzi/strimzi-kafka-operator/pull/9260/commits/f8c19a7ce3d24e6f61064f3a81b073034c3883b1)

scholzj: You fine with this @ppatierno? Can we merge it?

ppatierno: Yes it's fine with me. Thanks.

PR #9268, opened 2023-10-19
Fix some minor warnings and typos in several classes

scholzj:
### Type of change

- Bugfix

### Description

This PR fixes some minor warnings and typos in several of our classes.

### Checklist

### DIFF
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/CaReconciler.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/CaReconciler.java
@@ -594,5 +594,5 @@ public class CaReconciler {
     /**
      * Helper class to pass both Cluster and Clients CA as a result of the reconciliation
      */
-    protected record CaReconciliationResult(ClusterCa clusterCa, ClientsCa clientsCa) { }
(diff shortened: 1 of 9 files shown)


cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/CaReconciler.java
```diff
@@ -594,5 +594,5 @@ Future<Void> maybeRemoveOldClusterCaCertificates() {
     /**
      * Helper class to pass both Cluster and Clients CA as a result of the reconciliation
      */
-    protected record CaReconciliationResult(ClusterCa clusterCa, ClientsCa clientsCa) { }
+    public record CaReconciliationResult(ClusterCa clusterCa, ClientsCa clientsCa) { }
```
see-quick, line 597: Why does this need to be public?

see-quick (approved):

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/CaReconciler.java
scholzj: We return it via a public method. So yes, it should be public.

ppatierno (approved):

PR #9273, opened 2023-10-19
Fix UTO NPE with missing partitions and/or replicas

### LINKED ISSUE #9262: [Bug]: TopicOperator doesn't work with KRaft flags
### Bug Description

I have a cluster with a Zookeeper that works properly. The moment I enable the featureGates required for using KRaft (+UseKRaft,+KafkaNodePools,+UnidirectionalTopicOperator) the topic operator stops working. I've not added any annotations to the Kafka resource nor added any Kafka node pools. Just enabled the feature flags.

The topic operator is in Running state and does not fail its liveness or readiness probe. However, it does not work and logs constantly the same Unexpected Exception error:
```
java.lang.NullPointerException: Cannot invoke "java.lang.Integer.intValue()" because "requested" is null   
```

The only config added for the topicOperator in the Kafka resource is the following:
```
    topicOperator:
      jvmOptions:
        -Xms: 128m
        -Xmx: 128m
      resources:
        limits:
          memory: 600Mi
        requests:
          cpu: 100m
          memory: 600Mi
```

### Expected behavior

Is there any required modification to the Kafka resource in order to be able to enable the feature gates? The intention here is to utilize the new feature of having multiple clusters one with Zookeeper and one with KRaft. Currently just enabling the feature gate, stops the TopicOperator of the Zookeeper cluster of working properly.

### Strimzi version

0.38.0

### Kubernetes version

1.27.4

### Installation method

Helm

### Infrastructure

GKE

image

### Additional context

_No response_

fvaleri:
If you create or update a KafaTopic without partitions and/or replicas, the operator fails with a NPE. In all these cases, the operator should apply the default cluster values. Note that empty spec (spec: {}) can be reconciled, but it is different from missing spec, which should not be reconciled. We also need to match the BTO behavior here. In the future we will make the spec required at the schema level.

This should fix #9262.

### DIFF
--- a/topic-operator/src/main/java/io/strimzi/operator/topic/v2/BatchingTopicController.java
+++ b/topic-operator/src/main/java/io/strimzi/operator/topic/v2/BatchingTopicController.java
@@ -299,9 +299,15 @@ public class BatchingTopicController {
     }
 
     private static NewTopic buildNewTopic(KafkaTopic kt, String topicName) {
-        int partitions = kt.getSpec().getPartitions() != null ? kt.getSpec().getPartitions() : BROKER_DEFAULT;
-        short replicas = kt.getSpec().getReplicas() != null ? kt.getSpec().getReplicas().shortValue() : BROKER_DEFAULT;
-        return new NewTopic(topicName, partitions, replicas).configs(buildConfigsMap(kt));
+        return new NewTopic(topicName, partitions(kt), replicas(kt)).configs(buildConfigsMap(kt));
+    }
+
+    private static int partitions(KafkaTopic kt) {
+        return kt.getSpec().getPartitions() != null ? kt.getSpec().getPartitions() : BROKER_DEFAULT;
+    }
+
+    private static short replicas(KafkaTopic kt) {
+        return kt.getSpec().getReplicas() != null ? kt.getSpec().getReplicas().shortValue() : BROKER_DEFAULT;
     }
 
     private static Map<String, String> buildConfigsMap(KafkaTopic kt) {
@@ -436,8 +442,7 @@ public class BatchingTopicController {
         addOrRemoveFinalizer(useFinalizer, unmanaged).forEach(rt -> putResult(results, rt, Either.ofRight(null)));
 
         // skip reconciliation of paused KafkaTopics
-        var partitionedByPaused = validateManagedTopics(partitionedByManaged).stream()
-            .filter(skipKafkaTopicsWithMissingSpec())
+        var partitionedByPaused = validateManagedTopics(partitionedByManaged).stream().filter(hasTopicSpec)
             .collect(Collectors.partitioningBy(reconcilableTopic -> isPaused(reconcilableTopic.kt())));
         partitionedByPaused.get(true).forEach(reconcilableTopic -> putResult(results, reconcilableTopic, Either.ofRight(null)));
 
@@ -465,15 +470,13 @@ public class BatchingTopicController {
         LOGGER.traceOp("Total time reconciling batch of {} KafkaTopics: {}ns", results.size(), System.nanoTime() - t3);
     }
 
-    private static Predicate<ReconcilableTopic> skipKafkaTopicsWithMissingSpec() {
-        return reconcilableTopic -> {
-            var hasSpec = reconcilableTopic.kt().getSpec() != null;
-            if (!hasSpec) {
-                LOGGER.warnCr(reconcilableTopic.reconciliation(), "Topic has no spec.");
-            }
-            return hasSpec;
-        };
-    }
+    private Predicate<ReconcilableTopic> hasTopicSpec = reconcilableTopic -> {
+        var hasSpec = reconcilableTopic.kt().getSpec() != null;
+        if (!hasSpec) {
+            LOGGER.warnCr(reconcilableTopic.reconciliation(), "Topic has no spec.");
+        }
+        return hasSpec;
+    };
 
     private List<ReconcilableTopic> validateManagedTopics(Map<Boolean, List<ReconcilableTopic>> partitionedByManaged) {
         var mayNeedUpdate = partitionedByManaged.get(true).stream().filter(reconcilableTopic -> {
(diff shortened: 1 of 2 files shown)


scholzj (approved):
LGTM if it works for @tombentley

topic-operator/src/main/java/io/strimzi/operator/topic/v2/BatchingTopicController.java
```diff
@@ -438,6 +437,7 @@ private void updateInternal(List<ReconcilableTopic> topics) {
 
         // skip reconciliation of paused KafkaTopics
         var partitionedByPaused = validateManagedTopics(partitionedByManaged).stream()
+            .filter(skipKafkaTopicsWithMissingSpec())
```
tombentley, line 440: Rather than a method that returns a Predicate, just make a method that _is_ the predicate and use a method reference.

But is this behaviour really right? Should we ignore these resources, or reject them with an error in the `status`

topic-operator/src/main/java/io/strimzi/operator/topic/v2/BatchingTopicController.java
```diff
@@ -545,7 +558,8 @@ private void accumulateResults(Map<ReconcilableTopic, Either<TopicOperatorExcept
         var apparentlyDifferentRf = currentStatesOrError.ok().filter(pair -> {
             var reconcilableTopic = pair.getKey();
             var currentState = pair.getValue();
-            return currentState.uniqueReplicationFactor() != reconcilableTopic.kt().getSpec().getReplicas();
+            return reconcilableTopic.kt().getSpec().getReplicas() != null
+                && currentState.uniqueReplicationFactor() != reconcilableTopic.kt().getSpec().getReplicas();
```
tombentley, line 562: I'm not sure this is right. If the `spec.rf` is missing it means "broker default". That has a value even if it's not explicit. And so the actual number of replicas can differ from that.

topic-operator/src/main/java/io/strimzi/operator/topic/v2/BatchingTopicController.java
```diff
@@ -554,7 +568,9 @@ private void accumulateResults(Map<ReconcilableTopic, Either<TopicOperatorExcept
         });
         actuallyDifferentRf.ok().forEach(pair -> {
             var reconcilableTopic = pair.getKey();
-            var partitions = pair.getValue().partitionsWithDifferentRfThan(reconcilableTopic.kt().getSpec().getReplicas());
+            var specPartitions = reconcilableTopic.kt().getSpec().getPartitions() != null
+                ? reconcilableTopic.kt().getSpec().getPartitions() : BROKER_DEFAULT;
```
tombentley, line 572: This duplicate the expression on line 302. Can we factor out a method for it?

topic-operator/src/test/java/io/strimzi/operator/topic/v2/TopicControllerIT.java
```diff
@@ -96,7 +97,7 @@ class TopicControllerIT {
 
     private static final Logger LOGGER = LogManager.getLogger(TopicControllerIT.class);
     public static final Map<String, String> SELECTOR = Map.of("foo", "FOO", "bar", "BAR");
-    private static final String NAMESPACE = "ns";
+    private static final String NAMESPACE = "test-" + UUID.randomUUID();
```
tombentley, line 100: Why was this necessary?

topic-operator/src/test/java/io/strimzi/operator/topic/v2/TopicControllerIT.java
```diff
@@ -514,24 +533,6 @@ public void shouldCreateTopicInKafkaWhenManagedTopicCreatedInKube(KafkaTopic kt,
         createTopicAndAssertSuccess(kafkaCluster, kt);
     }
 
-    @Test
-    public void shouldCreateTopicInKafkaWhenKafkaTopicHasNoSpec(
```
tombentley, line 518: So we're changing a tested behavior. Are we _sure_ this isn't going to be a regression?

tombentley (commented):
Thanks @fvaleri!

topic-operator/src/main/java/io/strimzi/operator/topic/v2/BatchingTopicController.java
fvaleri: > Rather than a method that returns a Predicate, just make a method that is the predicate and use a method reference.

Ok. Thanks.

> But is this behaviour really right? Should we ignore these resources, or reject them with an error in the status

I thought about this. My change here is to exactly match what we have in BTO. There we only raise a warning in logs, without updating the status. In a future major release, we plan to make the spec a required field by changing the schema.

I'm in favor of updating the spec, but I would prefer to do it on a dedicated PR. In the meantime, I can log it as enhancement. Wdyt?

topic-operator/src/main/java/io/strimzi/operator/topic/v2/BatchingTopicController.java
fvaleri: Sure.

topic-operator/src/test/java/io/strimzi/operator/topic/v2/TopicControllerIT.java
fvaleri: This is just to minimize the risk of using an existing test namespace on the test cluster. Additionally, in case of test termination, you can run the suite again and delete the old namespace in parallel.

topic-operator/src/test/java/io/strimzi/operator/topic/v2/TopicControllerIT.java
fvaleri: Sure it is, but we are in alpha stage with very few users, so I guess this is fine. Also see my previous answer related to that.

topic-operator/src/main/java/io/strimzi/operator/topic/v2/BatchingTopicController.java
fvaleri: Let me check this one. Thanks.

topic-operator/src/test/java/io/strimzi/operator/topic/v2/TopicControllerIT.java
tombentley: The problem is you can end up not knowing what namespace it's using. It's annoying if you find you've got lots of `test-...` namespaces in your cluster and you don't know which was the one used for the test that just failed. But :man_shrugging: if you really want this

tombentley (approved):

topic-operator/src/test/java/io/strimzi/operator/topic/v2/TopicControllerIT.java
fvaleri: > It's annoying if you find you've got lots of test-... namespaces in your cluster and you don't know which was the one used

Yes, haven't thought about that. Will change to "uto-test".

ppatierno (approved):

PR #9276, opened 2023-10-20
Increase the size of the `/tmp` volumes to 10Mi

### LINKED ISSUE #9275: [Bug]: strimzi-tmp dir is  too small
### Bug Description

The default emptyDir limit for the strimzi-tmp directory is too small, being only 1MB.

When I use the Java profile function of the deepflow agent, the Java agent exports some files, such as perf.map, which can sometimes exceed 1MB. This results in the pod restarting.

Why is this directory limited to just 1MB? Also, could we use a specific directory like /tmp/strimzi instead of /tmp?

By the way, why do we need a 1MB emptyDir? I'm curious to know.

### Steps to reproduce

1. deploy deepflow agent
2. open java profile function

### Expected behavior

_No response_

### Strimzi version

main

### Kubernetes version

1.25.2

### Installation method

operator

### Infrastructure

k8s

https://strimzi.io/install/latest?namespace=kafka

### Additional context

_No response_

Maple1401:
### Type of change

- Enhancement 

### Description

https://github.com/strimzi/strimzi-kafka-operator/issues/9275

### Checklist

- [ ] Write tests
- [x] Update documentation
- [x] Update CHANGELOG.md

This resolves #9275

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 # CHANGELOG
 
 ## 0.39.0
-
+*  Increase the size of the `/tmp` volumes to 10Mi
 
 ## 0.38.0
 
@@ -9,7 +9,7 @@
 * Sign containers using `cosign`
 * Generate and publish Software Bill of Materials (SBOMs) of Strimzi containers
 * Add support for stopping connectors according to [Strimzi Proposal #54](https://github.com/strimzi/proposals/blob/main/054-stopping-kafka-connect-connectors.md)
-* Allow manual rolling of Kafka Connect and Kafka Mirror Maker 2 pods using the `strimzi.io/manual-rolling-update` annotation (supported only when `StableConnectIdentities` feature gate is enabled) 
+* Allow manual rolling of Kafka Connect and Kafka Mirror Maker 2 pods using the `strimzi.io/manual-rolling-update` annotation (supported only when `StableConnectIdentities` feature gate is enabled)
 * Make sure brokers are empty before scaling them down
 * Update Cruise Control to 2.5.128
 * Add support for pausing reconciliations to the Unidirectional Topic Operator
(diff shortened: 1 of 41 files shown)


scholzj (commented):
I don't think there is any reason to increase the size.

Maple1401:
> I don't think there is any reason to increase the size.

We have used some observability tools, such as Pixie and DeepFlow. They can observe the flame graph of Java programs, but it will cause the Java Virtual Machine to generate symbol table data in the /tmp directory, approximately 1-5MB.

So during the usage, we found that the Pod of the Kafka operator had mounted an emptyDir, which limited /tmp to only 1MB, causing problems with the Pod.

This kind of non-invasive Profile will be of great help to Java programs in the future, so to avoid others also encountering this situation, I think the default value can be changed to 10MB. After all, for modern servers running K8s, 10MB of memory will not have a significant impact.

If there are any omissions in my commit, feel free to point them out. （I am not very proficient at using GitHub.）
image

scholzj:
> We have used some observability tools, such as Pixie and DeepFlow. They can observe the flame graph of Java programs, but it will cause the Java Virtual Machine to generate symbol table data in the /tmp directory, approximately 1-5MB.

That is fine -> as I told you in the issue, that is why you have the option to configure the size of these volumes. But it is not a reason to change this for everyone.

Maple1401:
> > We have used some observability tools, such as Pixie and DeepFlow. They can observe the flame graph of Java programs, but it will cause the Java Virtual Machine to generate symbol table data in the /tmp directory, approximately 1-5MB.
> 
> That is fine -> as I told you in the issue, that is why you have the option to configure the size of these volumes. But it is not a reason to change this for everyone.


Alright, could I understand the reason and purpose for configuring this as 1MB here?

scholzj:
> Alright, could I understand the reason and purpose for configuring this as 1MB here?

It is enough for what Strimzi needs to use it for.

PR #9281, opened 2023-10-20
Remove cluster ID check and move the cluster.id configuration to Config Map

scholzj:
### Type of change

- Bugfix

### Description

As discussed in https://cloud-native.slack.com/archives/C018247K8T0/p1697627591772399, the cluster-ID validation does not really work well with KRaft where have to specify the Cluster ID on our own, but if the storage already exists with another cluster ID (that we have no easy way to find out) we cannot override it. That is causing artificial conflicts.

This PR removes the Cluster ID validation. It also moves the Cluster ID configuration to Config Map instead of using an environment variable. This helps to avoid unnecessary restarts during recovery from existing PVCs or during the KRaft migration.

### Checklist

- [x] Write tests


ppatierno (approved):
LGTM. This really helps my migration related work as well. Thanks!

Issue #8721, opened 2023-06-23
[Enhancement]: Use PEM certificates when connecting to KafkaAgent's REST endpoint


tinaselenge:
### Related problem

KafkaRoller currently uses PKCS12 certificates when connecting to KafkaAgent's REST endpoint to learn about broker state. 

### Suggested solution

We should implement using PEM certificates instead of PKCS12 because it is easier to use and more Kubernetes-friendly format. It may be easier for future work around certificate management.

scholzj:
Triaged on 29.6.2023: This makes sense as we are in general moving towards PEM certificates rather then PKCS12 stores.

PR #9288, opened 2023-10-23
Use PEM certificates when connecting to KafkaAgent REST endpoint

tinaselenge:
### Type of change
- Enhancement

### Description

This PR is for https://github.com/strimzi/strimzi-kafka-operator/issues/8721. 

### Checklist

- [ ] Write tests
- [ ] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/resource/KafkaAgentClient.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/resource/KafkaAgentClient.java
@@ -48,7 +48,8 @@ class KafkaAgentClient {
 
     private static final String BROKER_STATE_REST_PATH = "/v1/broker-state/";
     private static final int BROKER_STATE_HTTPS_PORT = 8443;
-
+    private static final String KEYSTORE_TYPE_JKS = "JKS";
+    private static final String CERT_TYPE_X509 = "X.509";
     private final String namespace;
     private final Reconciliation reconciliation;
     private final String cluster;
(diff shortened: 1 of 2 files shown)


scholzj (commented):
The code LGTM. I will run the STs. Would it make sense to have some unit tests?

operator-common/src/main/java/io/strimzi/operator/common/Util.java
```diff
@@ -166,6 +166,21 @@ public static byte[] decodeFromSecret(Secret secret, String key) {
         return Base64.getDecoder().decode(secret.getData().get(key));
     }
 
+    /**
+     * Decode binary item that holds Private key in PEM format from Kubernetes Secret from base64 into byte array
```
see-quick, line 170:
```suggestion
     * Decode the binary item in a Kubernetes Secret, which holds a private key in PEM format, from base64 to a byte array.
```
Just a nit, won't it be better this way?

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/resource/KafkaAgentClient.java
```diff
@@ -98,6 +98,33 @@ private HttpClient createHttpClient() {
         }
     }
 
+    private KeyStore getTrustStore() throws CertificateException, KeyStoreException, IOException, NoSuchAlgorithmException {
+        final CertificateFactory caCertFactory = CertificateFactory.getInstance("X.509");
```
see-quick, line 102: Don't we have some defined Constants for `X.509` or `JKS`? :) If not should we introduce them to eliminate redundancy in the project?

see-quick (approved):
LGTM, just two nits.

ppatierno (approved):
LGTM.

operator-common/src/main/java/io/strimzi/operator/common/Util.java
tinaselenge: yes, that does sound better. Thanks!

ppatierno:
> Would it make sense to have some unit tests?

@tinaselenge are you planning to write some?

tinaselenge:
> > Would it make sense to have some unit tests?
> 
> @tinaselenge are you planning to write some?

I was thinking what would be the best way to test it. I probably would need to create dummy private key/cert and secrets to check if truststore/keystore can be created from the secret? Or would it make more sense to write system test instead that connects to kafka agent with mutual TLS so that we are testing the whole flow?

scholzj:
> Or would it make more sense to write system test instead that connects to kafka agent with mutual TLS so that we are testing the whole flow?

Well, the mTLS is used all the time, so it does no need new STs I think. I guess the `decodePemPrivateKeyFromSecret` would be easy to unit test. But you are right that relying on the ST makes probably sense.

scholzj (approved):

Issue #7985, opened 2023-01-27
Use `additionalProperties` in the CRD OpenAPI spec for Maps of Strings instead of `x-kubernetes-preserve-unknown-fields`


scholzj:
As discussed in #7979, we should consider using the `additionalProperties` field instead of `x-kubernetes-preserve-unknown-fields` to improve the compatibility with other tools and be more in sync with Kubernetes spec. However, we need to keep in mind:
* That we need to stay backward compatible with existing resources
* We have places where Maps of Objects are used which should stick with ``x-kubernetes-preserve-unknown-fields`

scholzj:
Should be triaged first to see other views!

scholzj:
Triaged on 9.2.2023: This can be implemented, but:
* It seems like it should be used only for places where Map<String, String> is used and not where a deeper objects are (e.g. Connector status section)
* Backwards compatibility should be tested
* There is still some risk

omjikush09:
Hi, Can I work on this?

scholzj:
You can if you want. But I'm not sure how easy or straight forward it is TBH.

omjikush09:
Then I don't think it's a great idea for me to work on this if I get stuck with whom I ask for help. :confused:

PR #9292, opened 2023-10-25
Improvement of Resource Regeneration Due to x-kubernetes-preserve-unknown-fields

### LINKED ISSUE #9291: [Enhancement]: Improvement of Resource Regeneration Due to `x-kubernetes-preserve-unknown-fields`
### Related problem

The resources property of CRDs (Custom Resource Definitions) produced by the model managed in this project all include x-kubernetes-preserve-unknown-fields.
When I use this in conjunction with the terraform tool I'm using, it leads to a problem where Terraform destroys Kafka resources every time it needs to make changes to the resources property and then tries to add them again, resulting in the recreation of Kafka resources.

Simply changing the manifest and sequentially restarting Pods would suffice, but this process takes a long time to apply the changes.
Moreover, while Kafka resources are being recreated, the Kafka cluster becomes unusable.

I believed that to resolve this issue, CRDs needed to be modified, as this behavior is intentional in Terraform.

- https://github.com/hashicorp/terraform-provider-kubernetes/issues/1841#issuecomment-1254825692

### Suggested solution

To address this, my proposed idea was to make CrdGenerator explicitly define subproperties for the limits/requests properties of resources.
This way, Terraform can modify the resources property without recreating Kafka resources, allowing changes to be applied without the need for resource recreation.
However, I'm looking for better ideas to solve this issue!

### Alternatives

I also think that creating a separate model by inheriting from ResourceRequirements could be an alternative approach.

### Additional context

A simple code modification for this idea is similar to the following PR: https://github.com/strimzi/strimzi-kafka-operator/pull/9292

selenehyun:
### Type of change

- Enhancement / new feature

### Description

- This is an idea to address the resource regeneration issue that occurs when modifying the resources property with the same terraform as described in the issue.
- resolve #9291

### Checklist

### DIFF
--- a/crd-generator/src/main/java/io/strimzi/crdgenerator/CrdGenerator.java
+++ b/crd-generator/src/main/java/io/strimzi/crdgenerator/CrdGenerator.java
@@ -753,7 +753,21 @@ class CrdGenerator {
         PropertyType propertyType = property.getType();
         Class<?> returnType = propertyType.getType();
         final ObjectNode schema;
-        if (propertyType.getGenericType() instanceof ParameterizedType
+        if (property.getName().equals("limits")
+                || property.getName().equals("requests")) {
+            ObjectNode properties = nf.objectNode();
+            ObjectNode cpuProperty = nf.objectNode();
+            ObjectNode memoryProperty = nf.objectNode();
+            schema = nf.objectNode();
+            schema.put("type", "object");
+            schema.set("properties", properties);
+
+            cpuProperty.put("type", "string");
+            memoryProperty.put("type", "string");
+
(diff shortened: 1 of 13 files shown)


scholzj (commented):
Thanks for the PR. Please have a look at the issue https://github.com/strimzi/strimzi-kafka-operator/issues/7985 that covers this topic and the related discussion. If you decide to implement it, you should also probably do a more generic implementation and not hardcode particular fields into it.

selenehyun:
@scholzj I'll check it out, and if possible, I'll write satisfactory code and contribute!

scholzj:
@selenehyun Great, thanks.

PR #9298, opened 2023-10-27
Remove offset.lag.max from MirrorMaker examples

mimaison:
### Type of change

- Documentation/examples

### Description

`offset.lag.max` is a configuration from the source connector so we shouldn't see it on checkpoint connector. Also 100 is the default value so we don't need to see it explicitly.

### Checklist

- [ ] Write tests
- [ ] Update documentation
- [ ] Update CHANGELOG.md


scholzj:
It is intentionally configured woth the default value to raise awareness of
the issues it causes. If it is on the wrong place, move it, but not remove
it!

mimaison:
Gotcha! I re-added the configs and just moved it to the sourceConnector config

scholzj (approved):

ppatierno (approved):

PR #9301, opened 2023-10-29
Fixed KRaft storage formating with UUID starting with dash

ppatierno:
While using KRaft and creating a new KRaft based cluster I came in the following error on Kafka pods while formatting the storage ...

```shell
usage: kafka-storage format [-h] --config CONFIG --cluster-id CLUSTER_ID [--ignore-formatted] [--release-version RELEASE_VERSION]
kafka-storage: error: argument --cluster-id/-t: expected one argument
```

Which keeps the cluster to fail and not starting at all.
The issue was related to an autogenerated UUID starting with dash like this `-rmdB0m4T4--Y4thlNXk4Q`.
Because of the Kafka storage tool works by using argparse4j (for parsing argument), it's safer to assign parameters using the `=` sign on the command line, otherwise such an UUID is taken as a new argument and not a corresponding value.

### DIFF
--- a/docker-images/kafka-based/kafka/scripts/kafka_run.sh
+++ b/docker-images/kafka-based/kafka/scripts/kafka_run.sh
@@ -63,6 +63,8 @@ if [ "$STRIMZI_KRAFT_ENABLED" = "true" ]; then
     STRIMZI_CLUSTER_ID=$(cat "$KAFKA_HOME/custom-config/cluster.id")
     echo "Formatting Kraft storage with cluster ID $STRIMZI_CLUSTER_ID"
     mkdir -p "$KRAFT_LOG_DIR"
+    # Using "=" to assign arguments for the Kafka storage tool to avoid issues if the generated
+    # cluster ID starts with a "-". See https://issues.apache.org/jira/browse/KAFKA-15754
     ./bin/kafka-storage.sh format -t="$STRIMZI_CLUSTER_ID" -c=/tmp/strimzi.properties
   else
     echo "Kraft storage is already formatted"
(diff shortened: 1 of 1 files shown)


scholzj:
Isn't this a Kafka bug in the first place?

scholzj (approved):
I mean it is fine to fix this here. But should be raised to Kafka as well I guess?

ppatierno:
> I mean it is fine to fix this here. But should be raised to Kafka as well I guess?

I am also investigating on the Kafka side but I thought that fixing this here as well would have been better/faster.

ppatierno:
@scholzj I took a look on the Kafka side (checking how the `StorageTool` class works) but I think it's strictly related to how the argparse4j library works (it's the library used to parse the command line args). It parses everything starting with `-` as a new argument so in my above example, the `-t` is missing is value because the `-rmdB0m4T4--Y4thlNXk4Q` is handled as a new argument and not the value for `-t`. I checked the library and even unit tests check that it fails in these scenarios. I think the only "workaround" in this cases is using the `-t=<uuid>` form. So not sure we can say it's a Kafka issue.

scholzj:
@ppatierno Well, I wonder if Kafka should not generate such UUIDs in the first place if it causes issues.

ppatierno:
@scholzj it seems that Kafka already has this protection, see here https://github.com/apache/kafka/blob/trunk/clients/src/main/java/org/apache/kafka/common/Uuid.java#L113

So it was our operator code generating the UUID not taking into account this issue (I just discovered).
Now we have two options: leaving the current PR as it is because it should fix the problem, or reverting back here in the run script but generating an UUID as it is done in Kafka itself (see above) or both (but not sure how much it makes sense). My current thinking would be about solution 2 now, changing the UUID generation rather than the run script.
What's your opinion?

ppatierno:
@scholzj one more thing I didn't realized, the operator is using exactly the same `Uuid.randomUuid()` method from Kafka.
I see the commit fixing it here https://github.com/apache/kafka/commit/5c1dd493d6f608b566fdad5ab3a896cb13622bce and it seems to be in Kafka 3.6, so now I am not sure how it ended to generate the UUID I reported in the issue description.
I am using the current main Strimzi branch, so Kafka 3.6.0.

ppatierno:
@scholzj I have found an issue in Kafka :-) I will update you soon!

ppatierno:
@scholzj here we are, I opened this issue in Kafka which explains what we are seeing here: https://issues.apache.org/jira/browse/KAFKA-15754

scholzj:
> I would stick with the proposed solution in this PR. Using the "=" will just work with a UUID starting or not with a "-".

Yeah, it's fine for me to merge this. Maybe it is worth adding some explanatory comment why it uses the `=` and link to the Kafka issue?

PR #9307, opened 2023-10-31
[ST] Make possible to run Connect tests with file-sink plugin using pre-built image (and enable running Connect tests on Microshift)

im-konge:
### Type of change

- Enhancement

### Description

Until now, in testcases where we are using Connect with file-sink plugin, we used Connect Build feature for building the image with missing file-sink plugin. Because we are using this plugin in several testcases, it can extend the test duration.
Also, we were not able to execute these testcases on MicroShift. It is due to missing "build" feature on this environment (which have to be used for building the Connect image with the mentioned plugin). 

This PR adds possibility to deploy Connect with pre-built image containing the file-sink plugin.
It also adds Dockerfile and script for building the image itself.

This new possibility will help us:

- unblock testing of Connect on MicroShift
- speed up our pipelines the Connect image will be built before the test step and then the tests will be just executed with the particular Connect image (it takes around 1minute to build the image using build feature, so in case we have ~30 testcases for Connect with file-sink plugin, it can make significant difference).

### Checklist

### DIFF
--- a/.azure/templates/steps/system_test_general.yaml
+++ b/.azure/templates/steps/system_test_general.yaml
@@ -104,7 +104,7 @@ jobs:
         echo "##vso[task.setvariable variable=connectImage]$(echo $CONNECT_IMAGE_WITH_FILE_SINK_PLUGIN)"
       displayName: "Build KafkaConnect's image with file-sink plugin"
       env:
-        BASE_IMAGE: 'localhost:5000/$(docker_org)/kafka:$(docker_tag)-kafka-KAFKA_VERSION'
+        BASE_IMAGE: '$(docker_registry)/$(docker_org)/kafka:$(docker_tag)-kafka-KAFKA_VERSION'
         KAFKA_VERSION: '${{ parameters.kafkaVersion }}'
 
     - task: Maven@3
--- a/.azure/templates/steps/system_test_general.yaml
+++ b/.azure/templates/steps/system_test_general.yaml
@@ -87,16 +87,6 @@ jobs:
       displayName: "Create dir for Kafka binaries cache"
       condition: ne(variables['docker_tag'], 'latest')
 
-    # We need to set DOCKER_REGISTRY to IP and port of service, which is created by minikube registry addon, port is always 80
-    # Default value for PRs is localhost:5000 because we need to push built images into minikube registry and make them available for pods
-    - script: echo "##vso[task.setvariable variable=docker_registry]$(kubectl get service registry -n kube-system -o=jsonpath='{.spec.clusterIP}'):80"
-      displayName: "Set docker_registry to local registry in minikube"
-      condition: eq(variables['docker_registry'], 'localhost:5000')
-
-    - bash: ".azure/scripts/setup_upgrade.sh"
-      displayName: "Setup environment for upgrade"
-      condition: contains('${{ parameters.profile }}', 'upgrade')
-
     - bash: |
         eval $(minikube docker-env)
         eval $(./systemtest/src/test/resources/connect-build/build-connect-image.sh "$KAFKA_VERSION" "$BASE_IMAGE")
@@ -106,6 +96,16 @@ jobs:
       env:
         BASE_IMAGE: '$(docker_registry)/$(docker_org)/kafka:$(docker_tag)-kafka-KAFKA_VERSION'
         KAFKA_VERSION: '${{ parameters.kafkaVersion }}'
+
+    # We need to set DOCKER_REGISTRY to IP and port of service, which is created by minikube registry addon, port is always 80
+    # Default value for PRs is localhost:5000 because we need to push built images into minikube registry and make them available for pods
+    - script: echo "##vso[task.setvariable variable=docker_registry]$(kubectl get service registry -n kube-system -o=jsonpath='{.spec.clusterIP}'):80"
+      displayName: "Set docker_registry to local registry in minikube"
+      condition: eq(variables['docker_registry'], 'localhost:5000')
+
+    - bash: ".azure/scripts/setup_upgrade.sh"
+      displayName: "Setup environment for upgrade"
+      condition: contains('${{ parameters.profile }}', 'upgrade')
 
     - task: Maven@3
       inputs:
--- a/.azure/templates/steps/system_test_general.yaml
+++ b/.azure/templates/steps/system_test_general.yaml
@@ -99,7 +99,7 @@ jobs:
 
     - bash: |
         eval $(minikube docker-env)
-        eval $(./systemtest/src/test/resources/connect-build/build-connect-image.sh)
+        eval $(./systemtest/src/test/resources/connect-build/build-connect-image.sh ${{ parameters.kafkaVersion }}) $(docker_registry)/$(docker_org)/kafka:$(docker_tag)-kafka-KAFKA_VERSION
         echo "##vso[task.setvariable variable=connectImage]$(echo $CONNECT_IMAGE_WITH_FILE_SINK_PLUGIN)"
       displayName: "Build KafkaConnect's image with file-sink plugin"
 
--- a/.azure/templates/steps/system_test_general.yaml
+++ b/.azure/templates/steps/system_test_general.yaml
@@ -99,9 +99,13 @@ jobs:
 
     - bash: |
         eval $(minikube docker-env)
-        eval $(./systemtest/src/test/resources/connect-build/build-connect-image.sh ${{ parameters.kafkaVersion }}) $(docker_registry)/$(docker_org)/kafka:$(docker_tag)-kafka-KAFKA_VERSION
+        eval $(./systemtest/src/test/resources/connect-build/build-connect-image.sh "$KAFKA_VERSION" "$BASE_IMAGE")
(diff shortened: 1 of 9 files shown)


systemtest/src/test/resources/connect-build/build-connect-image.sh
```diff
@@ -0,0 +1,33 @@
...
+
+RANDOM=$(shuf -i 60000-99999 -n 1)
+
+KAFKA_VERSION=$1
+BASE_IMAGE=$2
+CONNECT_IMAGE=${3:-"strimzi/connect:$RANDOM"}
+
+function getLatestKafkaVersionFromYAML() {
+  local strimziRoot=$(git rev-parse --show-toplevel)
+  KAFKA_VERSION=$(cat $strimziRoot/kafka-versions.yaml | yq eval '.[] | select(.default) | .version' -)
```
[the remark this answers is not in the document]
im-konge: In case the `kafka-versions.yaml` is not in the root folder the script should simply fail -> in case user will not specify any Kafka version and we cannot get it from the YAML, it should not continue -> because we need it for downloading the `.jar` of file-sink.
Or what error handling do you mean? Some check that the `kafka-versions.yaml` exists and if not, `exit 1` and error message? (The error message cannot be there using echo or printf, as we should ideally use `eval $(.script)`, which would fail because "unknown command)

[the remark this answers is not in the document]
see-quick: I was thinking about something like this:
```bash
function getLatestKafkaVersionFromYAML() {
  local strimziRoot=$(git rev-parse --show-toplevel)
  local kafkaVersionsFile="$strimziRoot/kafka-versions.yaml"
  
  # Check if kafka-versions.yaml exists
  if [[ ! -f $kafkaVersionsFile ]]; then
    echo "[ERROR] File $kafkaVersionsFile does not exist."
    exit 1
  fi
  
  # Retrieve the Kafka version
  KAFKA_VERSION=$(cat $kafkaVersionsFile | yq eval '.[] | select(.default) | .version' -)
  
  # Check if the version is retrieved successfully
  if [[ -z $KAFKA_VERSION ]]; then
    echo "[ERROR] Failed to retrieve Kafka version from $kafkaVersionsFile."
    exit 1
  fi
}
```
But when you mention `eval`...I know that current implementation is okay ;)

see-quick (approved):
LGTM

jankalinic (approved):

.azure/templates/steps/system_test_general.yaml
```diff
@@ -97,6 +97,16 @@ jobs:
       displayName: "Setup environment for upgrade"
       condition: contains('${{ parameters.profile }}', 'upgrade')
 
+    - bash: |
```
scholzj, line 100: The description talks about _possibilities_ and Microshift. But you seem to use it here in the regular STs. I don't have a problem with that as long as the Connect Build is tested as well. Btu the description and Pr name should make it more clear I think.

systemtest/src/test/resources/connect-build/Dockerfile
```diff
@@ -0,0 +1,14 @@
+ARG KAFKA_VERSION
+ARG BASE_IMAGE=quay.io/strimzi/kafka:latest-kafka-${KAFKA_VERSION}
```
scholzj, line 2: How does this work for release tests? It should use the GA / RC image in such case.

systemtest/src/test/resources/connect-build/Dockerfile
```diff
@@ -0,0 +1,14 @@
...
+
+FROM ${BASE_IMAGE}
+
+# to make KAFKA_VERSION arg available again after FROM, we need to specify it once more (different build stage)
+ARG KAFKA_VERSION
+
+USER root:root
+RUN mkdir -p /opt/kafka/plugins/file-sink
+
+ADD --chmod=444 https://repo1.maven.org/maven2/org/apache/kafka/connect-file/${KAFKA_VERSION}/connect-file-${KAFKA_VERSION}.jar /opt/kafka/plugins/file-sink/
```
scholzj, line 12: Should you copy it within the image instead of relying on an outside source? It helps to avoid outages, etc.

systemtest/src/test/resources/connect-build/build-connect-image.sh
```diff
@@ -0,0 +1,32 @@
+#!/usr/bin/env bash
+
+RANDOM=$(shuf -i 60000-99999 -n 1)
+
+KAFKA_VERSION=$1
+BASE_IMAGE=$2
+CONNECT_IMAGE=${3:-"strimzi/connect:$RANDOM"}
```
scholzj, line 7: I don't get the random part. If it has some reason, it should have some comment explaining it. But it seems like a stable fixed name would be better. You push this only into the test registry that should be limited to the CI, so no need for randomness.

systemtest/src/test/resources/connect-build/Dockerfile
im-konge: In the AZP I'm passing it via `BASE_IMAGE: 'localhost:5000/$(docker_org)/kafka:$(docker_tag)-kafka-KAFKA_VERSION'` env, which should point to images that were build in one of the previous steps.

.azure/templates/steps/system_test_general.yaml
im-konge: Yeah I think we discussed it before that it would be good to have the image for Connect tests pre-built and don't rely on the Connect build (it will decrease the execution time). I will change the name of the PR and description. Thanks

systemtest/src/test/resources/connect-build/Dockerfile
im-konge: You mean having the jars already in place in the `systemtest/src/test/resources` and copy it in the image, or move the download the download part to the script I created?

systemtest/src/test/resources/connect-build/build-connect-image.sh
im-konge: Yeah, you are right.
At the beginning I thought that it will be better to have the tag randomized, so we will ensure that we are always using the "latest" or "correct" one, but this is the case for Connect build tests and not this one. I will remove that, thanks.

.azure/templates/steps/system_test_general.yaml
im-konge: PR name and description updated

systemtest/src/test/resources/connect-build/Dockerfile
scholzj: But when running the tests for releases, it does not build any images.

systemtest/src/test/resources/connect-build/Dockerfile
scholzj: The `connect-file-<version>.jar` file is IMHO already inside the image. Just not in the classpath. So you can IMHO just do `RUN cp /opt/kafka/libs/connect-file-3.6.0 ...` ... this is important since you use it from the image and do not need to care about the version, need to make the Maven repository configurable for RCs of Kafka etc.

im-konge: Ah okay, I will try it that way, thanks! :)

systemtest/src/test/resources/connect-build/Dockerfile
im-konge: Yeah, I will fix that. Good catch. Thanks

im-konge: So I changed it to `BASE_IMAGE: '$(docker_registry)/$(docker_org)/kafka:$(docker_tag)-kafka-KAFKA_VERSION'`. That should fix the issue with release pipelines. WDYT?

.azure/templates/steps/system_test_general.yaml
```diff
@@ -87,6 +87,16 @@ jobs:
       displayName: "Create dir for Kafka binaries cache"
       condition: ne(variables['docker_tag'], 'latest')
 
+    - bash: |
```
scholzj, line 90: Should you give it some nice _display name_?

scholzj (approved):
One nit. LGTM otherwise if ti works.

systemtest/src/test/resources/connect-build/Dockerfile
scholzj: I'm afraid I cannot say it just like that. You would need to try it.

.azure/templates/steps/system_test_general.yaml
im-konge: I have `Build KafkaConnect's image with file-sink plugin` -> should it be `Build Connect's image with file-sink for STs`?

scholzj: Weird. GitHub does not seem to do the highlighting right, so I did not notice it. Is it well aligned? I guess it is if it works. So I think it is fine - I don't care that much whether it's one or the other wording. I just thought there was nothing because of the highlighting.

im-konge: Yeah it happens to me sometimes as well. It has same alignment as other `displayName` fields. Anyway thanks :)

im-konge:
Test failures are not related to my PR and will be fixed in next one (these failures are present across PRs, not just in this one).

PR #9308, opened 2023-10-31
[ST] kafka node pool controllers configuration changes  and configuration propagation

henryZrncik:
### Type of change

- Enhancement / new feature

### Description
Implementation of new test to verify proper propagation of configuration when creating resource or modifying specification.

in PodSetUtils and StrimziPodSetUtils there are two new methods for waiting for changes/ stability in Pods annotations, we ho with implementation in StrimziPodSetUtils as data In Pods as such are not reflecting reality in case of Controller role when doing modifications. 

the test itself was discused on triage regarding Kafka Node Pools.  

### Checklist

- [x] Write tests

### DIFF
--- a/systemtest/src/main/java/io/strimzi/systemtest/Constants.java
+++ b/systemtest/src/main/java/io/strimzi/systemtest/Constants.java
@@ -5,6 +5,7 @@
 package io.strimzi.systemtest;
 
 import io.fabric8.kubernetes.api.model.LabelSelector;
+import io.strimzi.api.ResourceLabels;
 import io.strimzi.test.TestUtils;
 
 import java.time.Duration;
@@ -84,6 +85,14 @@ public interface Constants {
     String ECHO_SINK_JAR_CHECKSUM = "3f30d48079578f9f2d0a097ed9a7088773b135dff3dc8e70d87f8422c073adc1181cb41d823c1d1472b0447a337e4877e535daa34ca8ef21d608f8ee6f5e4a9c";
     String ECHO_SINK_FILE_NAME = "echo-sink-test.jar";
     String ECHO_SINK_JAR_WRONG_CHECKSUM = "f1f167902325062efc8c755647bc1b782b2b067a87a6e507ff7a3f6205803220";
+
+
+    String NODE_BROKER_CONFIG_HASH_ANNOTATION = ResourceLabels.STRIMZI_DOMAIN + "broker-configuration-hash";
+
+    /**
+     * Node pool labels
+     */
+    String NODE_POOL_LABEL = ResourceLabels.STRIMZI_DOMAIN + "pool-name";
 
     /**
      * Scraper pod labels
--- a/systemtest/src/main/java/io/strimzi/systemtest/Constants.java
+++ b/systemtest/src/main/java/io/strimzi/systemtest/Constants.java
@@ -86,13 +86,7 @@ public interface Constants {
     String ECHO_SINK_FILE_NAME = "echo-sink-test.jar";
     String ECHO_SINK_JAR_WRONG_CHECKSUM = "f1f167902325062efc8c755647bc1b782b2b067a87a6e507ff7a3f6205803220";
 
-
     String NODE_BROKER_CONFIG_HASH_ANNOTATION = ResourceLabels.STRIMZI_DOMAIN + "broker-configuration-hash";
-
-    /**
-     * Node pool labels
-     */
-    String NODE_POOL_LABEL = ResourceLabels.STRIMZI_DOMAIN + "pool-name";
 
     /**
      * Scraper pod labels
--- a/systemtest/src/main/java/io/strimzi/systemtest/resources/crd/KafkaNodePoolResource.java
+++ b/systemtest/src/main/java/io/strimzi/systemtest/resources/crd/KafkaNodePoolResource.java
@@ -23,7 +23,6 @@ import io.strimzi.systemtest.templates.crd.KafkaNodePoolTemplates;
 import io.strimzi.systemtest.utils.kubeUtils.objects.PersistentVolumeClaimUtils;
 
 import java.util.ArrayList;
-import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.function.Consumer;
@@ -81,10 +80,11 @@ public class KafkaNodePoolResource implements ResourceType<KafkaNodePool> {
      * @return a LabelSelector tailored for the specified Kafka cluster and node pool.
      */
     public static LabelSelector getLabelSelector(String clusterName, String kafkaNodePoolName) {
-        Map<String, String> matchLabels = new HashMap<>();
-        matchLabels.put(Labels.STRIMZI_CLUSTER_LABEL, clusterName);
-        matchLabels.put(Labels.STRIMZI_KIND_LABEL, "Kafka");
-        matchLabels.put(Constants.NODE_POOL_LABEL, kafkaNodePoolName);
+        Map<String, String> matchLabels = Map.of(
+            Labels.STRIMZI_CLUSTER_LABEL, clusterName,
+            Labels.STRIMZI_KIND_LABEL, Kafka.RESOURCE_KIND,
+            Labels.STRIMZI_POOL_NAME_LABEL, kafkaNodePoolName
+        );
 
         return new LabelSelectorBuilder()
             .withMatchLabels(matchLabels)
--- a/systemtest/src/main/java/io/strimzi/systemtest/resources/crd/KafkaNodePoolResource.java
+++ b/systemtest/src/main/java/io/strimzi/systemtest/resources/crd/KafkaNodePoolResource.java
@@ -4,6 +4,8 @@
  */
 package io.strimzi.systemtest.resources.crd;
 
+import io.fabric8.kubernetes.api.model.LabelSelector;
+import io.fabric8.kubernetes.api.model.LabelSelectorBuilder;
 import io.fabric8.kubernetes.client.dsl.MixedOperation;
 import io.fabric8.kubernetes.client.dsl.Resource;
 import io.strimzi.api.kafka.Crds;
@@ -21,7 +23,9 @@ import io.strimzi.systemtest.templates.crd.KafkaNodePoolTemplates;
 import io.strimzi.systemtest.utils.kubeUtils.objects.PersistentVolumeClaimUtils;
 
 import java.util.ArrayList;
+import java.util.HashMap;
 import java.util.List;
+import java.util.Map;
 import java.util.function.Consumer;
 
 import static io.strimzi.operator.common.Util.hashStub;
@@ -67,6 +71,24 @@ public class KafkaNodePoolResource implements ResourceType<KafkaNodePool> {
     @Override
     public boolean waitForReadiness(KafkaNodePool resource) {
         return resource != null;
+    }
+
+    /**
+     * Constructs a LabelSelector specific enough to match all Pods Managed by given {@code kafkaNodePoolName}.
+     *
+     * @param clusterName       the name of the Kafka cluster for which the label selector is to be created.
+     * @param kafkaNodePoolName the name of the Kafka node pool within the specified cluster.
+     * @return a LabelSelector tailored for the specified Kafka cluster and node pool.
+     */
+    public static LabelSelector getLabelSelector(String clusterName, String kafkaNodePoolName) {
+        Map<String, String> matchLabels = new HashMap<>();
+        matchLabels.put(Labels.STRIMZI_CLUSTER_LABEL, clusterName);
+        matchLabels.put(Labels.STRIMZI_KIND_LABEL, "Kafka");
+        matchLabels.put(Constants.NODE_POOL_LABEL, kafkaNodePoolName);
+
+        return new LabelSelectorBuilder()
+            .withMatchLabels(matchLabels)
+            .build();
     }
 
     public static MixedOperation<KafkaNodePool, KafkaNodePoolList, Resource<KafkaNodePool>> kafkaNodePoolClient() {
--- a/systemtest/src/main/java/io/strimzi/systemtest/utils/kubeUtils/controllers/StrimziPodSetUtils.java
+++ b/systemtest/src/main/java/io/strimzi/systemtest/utils/kubeUtils/controllers/StrimziPodSetUtils.java
@@ -21,6 +21,7 @@ import io.strimzi.test.TestUtils;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 
+import java.util.Arrays;
 import java.util.Collection;
 import java.util.Map;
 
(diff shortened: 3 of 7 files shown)


henryZrncik:
@strimzi-ci run tests --testcase=KafkaNodePoolST

systemtest/src/main/java/io/strimzi/systemtest/Constants.java
```diff
@@ -85,6 +86,14 @@ public interface Constants {
     String ECHO_SINK_FILE_NAME = "echo-sink-test.jar";
     String ECHO_SINK_JAR_WRONG_CHECKSUM = "f1f167902325062efc8c755647bc1b782b2b067a87a6e507ff7a3f6205803220";
 
+
+    String NODE_BROKER_CONFIG_HASH_ANNOTATION = ResourceLabels.STRIMZI_DOMAIN + "broker-configuration-hash";
```
im-konge, line 90: Extra line + maybe you can move the annotation from `KafkaCluster` class to some common, so we can use it as well (as other annotations). WDYT @scholzj ?

systemtest/src/main/java/io/strimzi/systemtest/Constants.java
```diff
@@ -85,6 +86,14 @@ public interface Constants {
...
+
+    /**
+     * Node pool labels
+     */
+    String NODE_POOL_LABEL = ResourceLabels.STRIMZI_DOMAIN + "pool-name";
```
im-konge, line 95: `Labels.STRIMZI_POOL_NAME_LABEL`

systemtest/src/main/java/io/strimzi/systemtest/resources/crd/KafkaNodePoolResource.java
```diff
@@ -69,6 +73,24 @@ public boolean waitForReadiness(KafkaNodePool resource) {
...
+     * Constructs a LabelSelector specific enough to match all Pods Managed by given {@code kafkaNodePoolName}.
+     *
+     * @param clusterName       the name of the Kafka cluster for which the label selector is to be created.
+     * @param kafkaNodePoolName the name of the Kafka node pool within the specified cluster.
+     * @return a LabelSelector tailored for the specified Kafka cluster and node pool.
+     */
+    public static LabelSelector getLabelSelector(String clusterName, String kafkaNodePoolName) {
+        Map<String, String> matchLabels = new HashMap<>();
+        matchLabels.put(Labels.STRIMZI_CLUSTER_LABEL, clusterName);
+        matchLabels.put(Labels.STRIMZI_KIND_LABEL, "Kafka");
```
im-konge, line 86:
```suggestion
        matchLabels.put(Labels.STRIMZI_KIND_LABEL, Kafka.RESOURCE_KIND);
```

systemtest/src/main/java/io/strimzi/systemtest/utils/kubeUtils/controllers/StrimziPodSetUtils.java
```diff
@@ -52,6 +53,82 @@ public static void waitForStrimziPodSetLabelsDeletion(String namespaceName, Stri
...
+
+        LOGGER.info("Waiting for Pod: {}/{} to change annotation value: {}-{}", namespaceName, podName, annotationKey, annotationValue);
+        final ObjectMapper objectMapper = new ObjectMapper();
+
+
+        TestUtils.waitFor("Pod to change value of annotation" + annotationKey + " with key " + annotationKey, Constants.POLL_INTERVAL_FOR_RESOURCE_READINESS,
+            DELETION_TIMEOUT, () -> {
+                // parse pod from SPS
+                StrimziPodSet sps = StrimziPodSetResource.strimziPodSetClient().inNamespace(namespaceName).withName(strimziPodSetName).get();
+                var podMap = sps.getSpec().getPods();
```
im-konge, line 75: I really don't like using the `var`, when you can determine the type.
I guess that the `sps.getSpec().getPods()` will return `List<Pod>` right? That's better for filtering and you don't need to do all the conversion.

systemtest/src/main/java/io/strimzi/systemtest/utils/kubeUtils/controllers/StrimziPodSetUtils.java
```diff
@@ -52,6 +53,82 @@ public static void waitForStrimziPodSetLabelsDeletion(String namespaceName, Stri
...
+     * Waits for a specific annotation key-value pair to remain stable (from start to beginning) on a given Pod.
+     *
+     * @param namespaceName   the namespace in which the Pod is located.
+     * @param annotationKey   the key of the annotation to be checked.
+     * @param annotationValue the expected value of the annotation to be checked.
+     * @param strimziPodSetName the name of the SPS whose Pod's annotation is being observed.
+     * @param podName the name of the Pod whose annotation is being observed.
+     */
+    public static void waitForPrevailedPodAnnotationKeyValuePairs(String namespaceName, String annotationKey, String annotationValue, String strimziPodSetName, String podName) {
+        LOGGER.info("Waiting for label: {}-{} to keep its value in namespace: {}", namespaceName, annotationKey, annotationValue);
```
im-konge, line 94: I guess you have wrong order of the parameters, or?
`{}-{}` should be key/value and `in namespace: {}` should be the namespace right?

systemtest/src/main/java/io/strimzi/systemtest/utils/kubeUtils/controllers/StrimziPodSetUtils.java
```diff
@@ -52,6 +53,82 @@ public static void waitForStrimziPodSetLabelsDeletion(String namespaceName, Stri
...
+            Constants.GLOBAL_POLL_INTERVAL, Constants.GLOBAL_STATUS_TIMEOUT,
+
+            () -> {
+                // parse pod from SPS
+                StrimziPodSet sps =  StrimziPodSetResource.strimziPodSetClient().inNamespace(namespaceName).withName(strimziPodSetName).get();
+                var podMap = sps.getSpec().getPods();
+
+                Pod[] spsPods = objectMapper.convertValue(podMap, Pod[].class);
+                Pod pod = Arrays.stream(spsPods).filter(p -> p.getMetadata().getName().contains(podName)).findAny().get();
+
```
im-konge, line 108: Same as above, please, don't use the `var`

im-konge: Also, in both methods you are using the same thing, what about creating another method and using it in these two methods?

systemtest/src/main/java/io/strimzi/systemtest/utils/kubeUtils/controllers/StrimziPodSetUtils.java
```diff
@@ -52,6 +53,82 @@ public static void waitForStrimziPodSetLabelsDeletion(String namespaceName, Stri
...
+                        return true;
+                    }
+                } else {
+                    LOGGER.info("Annotations does not have expected value");
+                    return false;
+                }
+                LOGGER.info("Annotation be assumed stable in {} polls", Constants.GLOBAL_STABILITY_OFFSET_COUNT - stableCounter[0]);
+                return false;
+            });
+        LOGGER.info("Pod: {}/{} kept annotation: {}-{} ", namespaceName, podName, annotationKey, annotationValue);
```
im-konge, line 122: I guess you are adding a lot of info log here, which will be really noisy.

systemtest/src/main/java/io/strimzi/systemtest/utils/kubeUtils/objects/PodUtils.java
```diff
@@ -238,6 +238,61 @@ public static void waitUntilPodLabelsDeletion(String namespaceName, String podNa
...
+                        return true;
+                    }
+                } else {
+                    LOGGER.info("Annotations does not have expected value");
+                    return false;
+                }
+                LOGGER.info("Annotation be assumed stable in {} polls", Constants.GLOBAL_STABILITY_OFFSET_COUNT - stableCounter[0]);
+                return false;
+            });
+        LOGGER.info("Pod: {}/{} kept annotation: {}-{} ", namespaceName, observedPodName, annotationKey, annotationValue);
```
im-konge, line 293: Same for the logs here

systemtest/src/main/java/io/strimzi/systemtest/utils/kubeUtils/objects/PodUtils.java
```diff
@@ -238,6 +238,61 @@ public static void waitUntilPodLabelsDeletion(String namespaceName, String podNa
...
+     * Waits for a specific annotation key-value pair to remain stable (from start to beginning) on a given Pod.
+     *
+     * @param namespaceName   the namespace in which the Pod is located.
+     * @param annotationKey   the key of the annotation to be checked.
+     * @param annotationValue the expected value of the annotation to be checked.
+     * @param observedPodName the name of the Pod whose annotation is being observed.
+     *
+     * @throws TimeoutException if the annotation does not remain stable within the {@code Constants.GLOBAL_STATUS_TIMEOUT} period.
+     */
+    public static void waitForPrevailedPodAnnotationKeyValuePairs(String namespaceName, String annotationKey, String annotationValue, String observedPodName) {
```
im-konge, line 271: I guess you don't need to have the same thing on two places -> in `PodUtils` you can have this waits and in `StrimziPodSetUtils` you can have some encapsulation, which will pass the Pod names to the `PodUtils` methods.

systemtest/src/test/java/io/strimzi/systemtest/kafka/KafkaNodePoolST.java
```diff
@@ -308,13 +317,164 @@ void testKafkaNodePoolBrokerIdsManagementUsingAnnotations(ExtensionContext exten
...
+        RollingUpdateUtils.waitForNoRollingUpdate(testStorage.getNamespaceName(), nodePoolBLabelSelector, nodePoolBPodsSnapshot);
+        RollingUpdateUtils.waitTillComponentHasRolled(testStorage.getNamespaceName(), nodePoolALabelSelector, 2, nodePoolAPodsSnapshot);
+        //waitForPodsReady(testStorage.getNamespaceName(), testStorage.getKafkaSelector(), scaledUpBrokerReplicaCount, false);
+
+        // assert that new changed configuration is present in pods under Kafka Node Pool A
+        verifyResourcesConfigurationPropagatedToKnpPods(testStorage, nodePoolNameA, kafkaJvmXmxConfig, kafkaLimitMemoryQuantityConfigModified);
+
+        LOGGER.info("Modifying configuration relevant or irrelevant based on role of given KafkaNodePools");
+
+        final Pod podFromKnpC = PodUtils.getPodsByPrefixInNameWithDynamicWait(testStorage.getNamespaceName(), testStorage.getClusterName() + "-" + nodePoolNameC).get(0);
```
im-konge, line 425: Isn't there some method for constructing the StrimziPodSet name instead of doing `testStorage.getClusterName() + "-" + nodePoolNameC`?

systemtest/src/test/java/io/strimzi/systemtest/kafka/KafkaNodePoolST.java
```diff
@@ -308,13 +317,164 @@ void testKafkaNodePoolBrokerIdsManagementUsingAnnotations(ExtensionContext exten
...
+        final String podCName = podFromKnpC.getMetadata().getName();
+        final String knpCConfigHashAnnotationValue = podFromKnpC.getMetadata().getAnnotations().get(Constants.NODE_BROKER_CONFIG_HASH_ANNOTATION);
+        final String knpCStrimziPodSetName = KafkaNodePoolUtils.getStrimziPodSetName(testStorage.getClusterName(), nodePoolNameC);
+
+        // replacing config which is not relevant to control (also it is part of dynamic config, therefore should not trigger Rolling Update)
+        KafkaResource.replaceKafkaResourceInSpecificNamespace(testStorage.getClusterName(), k -> {
+            k.getSpec().getKafka().getConfig().put("compression.type", "gzip");
+        }, testStorage.getNamespaceName());
```
im-konge, line 433: One liner instead of block of code in lambda?

systemtest/src/test/java/io/strimzi/systemtest/kafka/KafkaNodePoolST.java
```diff
@@ -308,13 +317,164 @@ void testKafkaNodePoolBrokerIdsManagementUsingAnnotations(ExtensionContext exten
...
     void setup(ExtensionContext extensionContext) {
         assumeFalse(Environment.isOlmInstall() || Environment.isHelmInstall());
 
         List<EnvVar> coEnvVars = new ArrayList<>();
         coEnvVars.add(new EnvVar(Environment.STRIMZI_FEATURE_GATES_ENV, "+KafkaNodePools", null));
 
+        if (Environment.isKRaftModeEnabled()) {
+            coEnvVars.add(new EnvVar(Environment.STRIMZI_FEATURE_GATES_ENV, "+UseKRaft,+KafkaNodePools", null));
+        }
+
```
im-konge, line 477: I would maybe add the `+UseKRaft` to the default and run all the tests in KRaft mode -> it will be anyway required or?

systemtest/src/main/java/io/strimzi/systemtest/resources/crd/KafkaNodePoolResource.java
see-quick, line 86: Maybe use Map.of(...) instead of creating mutable map

systemtest/src/main/java/io/strimzi/systemtest/utils/kubeUtils/controllers/StrimziPodSetUtils.java
see-quick: yes, should be like this
```suggestion
        LOGGER.info("Waiting for label: {}-{} to keep its value in namespace: {}", annotationKey, annotationValue, namespaceName);
```

systemtest/src/test/java/io/strimzi/systemtest/kafka/KafkaNodePoolST.java
```diff
@@ -308,13 +317,164 @@ void testKafkaNodePoolBrokerIdsManagementUsingAnnotations(ExtensionContext exten
             KafkaNodePoolUtils.getCurrentKafkaNodePoolIds(testStorage.getNamespaceName(), nodePoolNameB).equals(Arrays.asList(0, 2, 3, 4, 6, 7)));
     }
 
+    @ParallelNamespaceTest
+    void testKNPConfigurationPropagationAndModification(ExtensionContext extensionContext) {
+
```
see-quick, line 322: remove this line?

systemtest/src/test/java/io/strimzi/systemtest/kafka/KafkaNodePoolST.java
```diff
@@ -308,13 +317,164 @@ void testKafkaNodePoolBrokerIdsManagementUsingAnnotations(ExtensionContext exten
...
+     * <p>This method checks the memory limit and JVM heap options of the first pod belonging to the given node pool.
+     *
+     * @param testStorage        the storage object containing test-related information.
+     * @param nodePoolName       the name of the KNP to verify.
+     * @param expectedJvmXmxLimit the expected JVM '-Xmx' heap size limit.
+     * @param expectedMemoryLimit the expected memory limit ('resource.limits.memory') for the pod.
+     *
+     * @throws AssertionError if configurations obtained from Pods do not match the expected values.
+     */
+    public static void verifyResourcesConfigurationPropagatedToKnpPods(TestStorage testStorage, String nodePoolName, String expectedJvmXmxLimit, String expectedMemoryLimit) {
```
see-quick, line 457: There is no need to have this method static and also public...
```suggestion
    private void verifyResourcesConfigurationPropagatedToKnpPods(TestStorage testStorage, String nodePoolName, String expectedJvmXmxLimit, String expectedMemoryLimit) {
```

see-quick (commented):
Few nits. Anyway, this looks very good!

systemtest/src/main/java/io/strimzi/systemtest/utils/kubeUtils/objects/PodUtils.java
henryZrncik: actually i can't as I am writing in the the PR description, if i would look into Pod  (belonging to Kafka Node Pool Controller role, and i would change some `spec.kafka.config` this should cause changes in annotation for config hash, but this annotation is  visibly changed only when we get to this Pod inside SPS CR, not inside Pod as such (i.e. (kubectl get pod my-controller-pod would still have the old value of mentioned annotation) (I think that changing this annotation would actually cause rolling of given pod)

im-konge: Okay, thanks for the explanation :)

systemtest/src/test/java/io/strimzi/systemtest/kafka/KafkaNodePoolST.java
```diff
@@ -52,170 +52,6 @@
...
-     * 6. - Modify Kafka custom resource annotation strimzi.io/node-pool to enable management by KafkaNodePool.
-     *    - new StrimziPodSet is created, replacing former one, Pods are replaced and specification from KafkaNodePool (i.e., changed replica count) has priority over Kafka specification.
-     * 7. - Produce and consume messages in given Kafka Cluster.
-     *    - Clients can produce and consume messages.
-     *
-     * @usecase
-     * - kafka-node-pool
-     */
-    @ParallelNamespaceTest
-    void testKafkaManagementTransferToAndFromKafkaNodePool(ExtensionContext extensionContext) {
```
scholzj, line 79: This seems to be just moved to another class. Would it be better to leave it for a separate PR? Or how does it relate to the topic of this PR?

systemtest/src/test/java/io/strimzi/systemtest/kafka/KafkaNodePoolST.java
```diff
@@ -308,12 +146,159 @@ void testKafkaNodePoolBrokerIdsManagementUsingAnnotations(ExtensionContext exten
             KafkaNodePoolUtils.getCurrentKafkaNodePoolIds(testStorage.getNamespaceName(), nodePoolNameB).equals(Arrays.asList(0, 2, 3, 4, 6, 7)));
     }
 
+    @ParallelNamespaceTest
+    void testKNPConfigurationPropagationAndModification(ExtensionContext extensionContext) {
+        // unless kraft is enabled there is no point testing controllers as only allowed role without KRaft is 'Broker'
+//        assumeTrue(Environment.isKRaftModeEnabled());
```
scholzj, line 152: Is this intentionally commented out?

systemtest/src/test/java/io/strimzi/systemtest/kafka/KafkaNodePoolST.java
```diff
@@ -308,12 +146,159 @@ void testKafkaNodePoolBrokerIdsManagementUsingAnnotations(ExtensionContext exten
...
+        final String podCName = podFromKnpC.getMetadata().getName();
+        final String knpCConfigHashAnnotationValue = podFromKnpC.getMetadata().getAnnotations().get(Constants.NODE_BROKER_CONFIG_HASH_ANNOTATION);
+        final String knpCStrimziPodSetName = KafkaNodePoolUtils.getStrimziPodSetName(testStorage.getClusterName(), nodePoolNameC);
+
+        // replacing config which is not relevant to control (also it is part of dynamic config, therefore should not trigger Rolling Update)
+        KafkaResource.replaceKafkaResourceInSpecificNamespace(testStorage.getClusterName(), k -> k.getSpec().getKafka().getConfig().put("compression.type", "gzip"), testStorage.getNamespaceName());
+
+        StrimziPodSetUtils.waitForPrevailedPodAnnotationKeyValuePairs(testStorage.getNamespaceName(), Constants.NODE_BROKER_CONFIG_HASH_ANNOTATION, knpCConfigHashAnnotationValue, knpCStrimziPodSetName, podCName);
+
+        // replacing config relevant only to controllers
```
scholzj, line 267: Why is this relevant only to controllers?

scholzj (commented):
I left some comments. But TBH, this does not seem right. This tests somethings that is unfinished and will ultimately go away (the fact that the controllers don't really roll). It also relies on internal implementation (the configuration hash annotation) which seems like a bad idea for a system test.

Isn't it better to wait for this to be completed? Or maybe even better, add this test with a proper waiting for RU to the PR from Kate / Tina?

PR #9146, opened 2023-09-21
Add support for KRaft in KafkaRoller

tinaselenge:
### Type of change

- Enhancement / new feature

### Description

This PR implements [proposal](https://github.com/strimzi/proposals/pull/91). 

### Checklist

_Please go through this checklist and make sure all applicable tasks have been done_

- [ ] Write tests
- [ ] Make sure all tests pass
- [ ] Update documentation
- [ ] Check RBAC rights for Kubernetes / OpenShift roles
- [ ] Try your changes from Pod inside your Kubernetes and OpenShift cluster, not just locally
- [ ] Reference relevant issue(s) and close them after merging
- [ ] Update CHANGELOG.md
- [ ] Supply screenshots for visual changes, such as Grafana dashboards

PR #9324, opened 2023-11-06
ZooKeeper to KRaft migration

### LINKED ISSUE #9433: [KRaft] ZooKeeper to KRaft migration
We have got the "ZooKeeper to KRaft migration" [proposal](https://github.com/strimzi/proposals/blob/main/059-zk-kraft-migration.md) approved.
This issue is about implementing the proposal by having the possibility to migrate a ZooKeeper-based cluster to a KRaft-based one through the semi-automatic process in the cluster operator.

ppatierno:
### Type of change

- Enhancement / new feature

### Description

This PR fixes #9433 about implementing the ZooKeeper to KRaft migration process.

The current status is about starting from a ZooKeeper-based cluster, creating a node pool with controllers, applying the `strimzi.io/kraft: migration` annotation on the `Kafka` CR and allowing the operator FSM to start the migration up to the "dual-write" (so creating controllers, rolling brokers, start migration). Then the user can apply the `strimzi.io/kraft: enabled` annotation in order to finilize the process by having the full KRaft-based cluster,
Different things are still missing:

* ~~setting some warnings in the `Kafka` CR status when annotation is applied in non valid states;~~
* ~~checking if readiness/liveness scripts need to evaluate the node role other than the migration state. Maybe it won't be fixed until the new KRaft readiness/liveness will land into the codebase;~~
* ~~checking the values for KAFKA_READY and ZK_CONNECTED env var in the Kafka run script;~~
* ~~the `useKRaft` flag passed around to the `KafkaCluster` is now considered "just" the value of the feature gate (it doesn't take into account `strimzi.io/kraft: enabled` annotation anymore, because now KRaft can be early enabled during migration). Maybe I should use anyway this flag together with all other checks? Not sure ...~~
* unit tests and integration tests for new additions not developed yet;
* ~~logic in the Kafka agent able to get metrics about migration allowing the FSM to know when migration ends and move on to the next phase (which should already work because it would be just a matter of changing the `strimzi.io/kraft` annotation to `enabled` and allowing the operator to update nodes configuration and rolling them). NOTE: rolling controllers is a blocker due to #9146)~~

The PR won't address documentation. There will be a different PR for it.

### Checklist

- [ ] Write tests
- [ ] Update CHANGELOG.md

### DIFF
--- a/.checkstyle/import-control.xml
+++ b/.checkstyle/import-control.xml
@@ -78,6 +78,8 @@ We also control imports only in production classes and not in tests. This is con
         <allow class="io.strimzi.operator.common.Annotations" />
         <allow class="io.strimzi.operator.cluster.ClusterOperatorConfig" />
         <allow class="io.strimzi.operator.cluster.PlatformFeaturesAvailability" />
+        <allow class="io.strimzi.operator.cluster.operator.resource.KafkaMetadataStateManager" />
+        <allow class="io.strimzi.operator.cluster.operator.resource.KafkaMetadataConfigurationState" />
 
         <!-- Other Strimzi projects -->
         <allow pkg="io.strimzi.kafka.oauth" />
--- a/api/src/main/java/io/strimzi/api/kafka/model/Kafka.java
+++ b/api/src/main/java/io/strimzi/api/kafka/model/Kafka.java
@@ -63,6 +63,11 @@ import java.util.function.Predicate;
                 jsonPath = ".status.conditions[?(@.type==\"Ready\")].status",
                 type = "string"),
             @Crd.Spec.AdditionalPrinterColumn(
+                name = "Metadata State",
+                description = "The state of the cluster metadata",
+                jsonPath = ".status.kafkaMetadataState",
+                type = "string"),
+            @Crd.Spec.AdditionalPrinterColumn(
                 name = "Warnings",
                 description = "Warnings related to the custom resource",
                 jsonPath = ".status.conditions[?(@.type==\"Warning\")].status",
--- a/api/src/main/java/io/strimzi/api/kafka/model/status/KafkaStatus.java
+++ b/api/src/main/java/io/strimzi/api/kafka/model/status/KafkaStatus.java
@@ -22,7 +22,7 @@ import java.util.List;
         builderPackage = Constants.FABRIC8_KUBERNETES_API
 )
 @JsonInclude(JsonInclude.Include.NON_NULL)
-@JsonPropertyOrder({ "conditions", "observedGeneration", "listeners", "kafkaNodePools", "clusterId", "operatorLastSuccessfulVersion", "kafkaVersion", "kafkaMetadataVersion" })
+@JsonPropertyOrder({ "conditions", "observedGeneration", "listeners", "kafkaNodePools", "clusterId", "operatorLastSuccessfulVersion", "kafkaVersion", "kafkaMetadataVersion", "kafkaMetadataState" })
 @EqualsAndHashCode
 @ToString(callSuper = true)
 public class KafkaStatus extends Status {
@@ -35,6 +35,7 @@ public class KafkaStatus extends Status {
     private String operatorLastSuccessfulVersion;
     private String kafkaVersion;
     private String kafkaMetadataVersion;
+    private String kafkaMetadataState;
 
     @Description("Addresses of the internal and external listeners")
     public List<ListenerStatus> getListeners() {
@@ -88,5 +89,19 @@ public class KafkaStatus extends Status {
 
     public void setKafkaMetadataVersion(String kafkaMetadataVersion) {
         this.kafkaMetadataVersion = kafkaMetadataVersion;
+    }
+
+    @Description("Defines where cluster metadata are stored. Possible values are: " +
+            "ZooKeeper if the metadata are stored in ZooKeeper " +
+            "KRaftMigration if the controllers are connected to ZooKeeper, together with brokers, and the migration process is running " +
+            "KRaftDualWriting if the migration process finished and the cluster is in dual-write mode " +
+            "KRaftPostMigration if the brokers are fully KRaft-based but controllers being rolled to disconnect from ZooKeeper " +
+            "KRaft if the metadata are stored in KRaft")
+    public String getKafkaMetadataState() {
+        return kafkaMetadataState;
+    }
+
+    public void setKafkaMetadataState(String metadataState) {
+        this.kafkaMetadataState = metadataState;
     }
 }
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaBrokerConfigurationBuilder.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaBrokerConfigurationBuilder.java
@@ -24,6 +24,7 @@ import io.strimzi.api.kafka.model.nodepool.ProcessRoles;
 import io.strimzi.kafka.oauth.server.ServerConfig;
 import io.strimzi.kafka.oauth.server.plain.ServerPlainConfig;
 import io.strimzi.operator.cluster.model.cruisecontrol.CruiseControlMetricsReporter;
+import io.strimzi.operator.cluster.operator.resource.KafkaMetadataConfigurationState;
 import io.strimzi.operator.common.model.cruisecontrol.CruiseControlConfigurationParameters;
 import io.strimzi.operator.common.Reconciliation;
 
@@ -68,21 +69,21 @@ public class KafkaBrokerConfigurationBuilder {
     private final StringWriter stringWriter = new StringWriter();
     private final PrintWriter writer = new PrintWriter(stringWriter);
     private final Reconciliation reconciliation;
-    private final String brokerId;
-    private final boolean useKRaft;
+    private final KafkaMetadataConfigurationState kafkaMetadataConfigState;
+    private final NodeRef node;
 
     /**
      * Broker configuration template constructor
      *
      * @param reconciliation    The reconciliation
-     * @param brokerId          The ID of the broker
-     * @param useKRaft          Indicates whether KRaft is used or not
+     * @param node              NodeRef instance
+     * @param kafkaMetadataConfigState  Represents the state of the Kafka metadata configuration
      */
-    public KafkaBrokerConfigurationBuilder(Reconciliation reconciliation, String brokerId, boolean useKRaft) {
+    public KafkaBrokerConfigurationBuilder(Reconciliation reconciliation, NodeRef node, KafkaMetadataConfigurationState kafkaMetadataConfigState) {
         printHeader();
         this.reconciliation = reconciliation;
-        this.brokerId = brokerId;
-        this.useKRaft = useKRaft;
+        this.node = node;
+        this.kafkaMetadataConfigState = kafkaMetadataConfigState;
 
         // Render the node/broker ID into the config file
         configureNodeOrBrokerId();
@@ -94,13 +95,12 @@ public class KafkaBrokerConfigurationBuilder {
     private void configureNodeOrBrokerId()   {
         printSectionHeader("Node / Broker ID");
 
-        if (useKRaft) {
-            writer.println("node.id=" + brokerId);
-        } else {
-            writer.println("broker.id=" + brokerId);
-            // Node ID is ignored when not using Kraft mode => but it defaults to broker ID when not set.
-            // We set it here in the configuration explicitly to avoid never ending rolling updates.
-            writer.println("node.id=" + brokerId);
+        // Node ID is ignored when not using Kraft mode => but it defaults to broker ID when not set.
+        // We set it here in the configuration explicitly to avoid never ending rolling updates.
+        writer.println("node.id=" + node.nodeId());
+        // only broker in ZooKeeper-mode or during migration needs the Broker ID to be set
+        if (node.broker() && kafkaMetadataConfigState.isZooKeeperOrMigration()) {
+            writer.println("broker.id=" + node.nodeId());
         }
 
         writer.println();
@@ -193,6 +193,20 @@ public class KafkaBrokerConfigurationBuilder {
     }
 
     /**
+     * Configures ZooKeeper migration related flag
+     *
+     * @param isEnabled if ZooKeeper migration is enabled
+     * @return the builder instance
+     */
+    public KafkaBrokerConfigurationBuilder withZooKeeperMigration(boolean isEnabled) {
+        printSectionHeader("Zookeeper migration");
+        writer.println("zookeeper.metadata.migration.enable=" + isEnabled);
+        writer.println();
+
+        return this;
+    }
+
+    /**
      * Adds the KRaft configuration for ZooKeeper-less Kafka. This includes the roles of the broker, the controller
      * listener name and the list of all controllers for quorum voting.
      *
@@ -205,7 +219,9 @@ public class KafkaBrokerConfigurationBuilder {
      */
     public KafkaBrokerConfigurationBuilder withKRaft(String clusterName, String namespace, Set<ProcessRoles> kraftRoles, Set<NodeRef> nodes)   {
         printSectionHeader("KRaft configuration");
-        writer.println("process.roles=" + kraftRoles.stream().map(role -> role.toValue()).sorted().collect(Collectors.joining(",")));
+        if (node.controller() || (node.broker() && kafkaMetadataConfigState.isPostMigrationOrKRaft())) {
+            writer.println("process.roles=" + kraftRoles.stream().map(role -> role.toValue()).sorted().collect(Collectors.joining(",")));
+        }
         writer.println("controller.listener.names=" + CONTROL_PLANE_LISTENER_NAME);
 
         // Generates the controllers quorum list
@@ -238,6 +254,7 @@ public class KafkaBrokerConfigurationBuilder {
      *
      * @return Returns the builder instance
      */
+    @SuppressWarnings({"checkstyle:CyclomaticComplexity"})
     public KafkaBrokerConfigurationBuilder withListeners(
             String clusterName,
             String namespace,
@@ -250,14 +267,14 @@ public class KafkaBrokerConfigurationBuilder {
         List<String> advertisedListeners = new ArrayList<>();
         List<String> securityProtocol = new ArrayList<>();
 
-        boolean isKraftControllerOnly = useKRaft && node.controller() && !node.broker();
-        boolean isKraftBrokerOnly = useKRaft && node.broker() && !node.controller();
+        boolean isKraftControllerOnly = kafkaMetadataConfigState.isKRaftInConfiguration(node) && node.controller() && !node.broker();
 
-        // Control Plane listener => ZooKeeper based brokers and KRaft controllers (including mixed nodes). But not on KRaft broker only nodes
-        if (!isKraftBrokerOnly) {
+        // Control Plane listener is set for pure KRaft controller or combined node, and broker in ZooKeeper mode or in migration state but not when full KRaft.
+        if (node.controller() || (node.broker() && kafkaMetadataConfigState.isZooKeeperOrMigration())) {
             listeners.add(CONTROL_PLANE_LISTENER_NAME + "://0.0.0.0:9090");
 
-            if (!useKRaft) {
+            // Control Plane listener to be advertised only with broker in ZooKeeper-based or migration but NOT when full KRaft only or mixed
+            if (node.broker() && kafkaMetadataConfigState.isZooKeeperOrMigration()) {
                 advertisedListeners.add(String.format("%s://%s:9090",
                         CONTROL_PLANE_LISTENER_NAME,
                         // Pod name constructed to be templatable for each individual ordinal
@@ -277,6 +294,12 @@ public class KafkaBrokerConfigurationBuilder {
         // so that brokers are able to connect to controllers as TLS clients
         configureControlPlaneListener();
 
+        // Replication Listener to be configured on brokers and KRaft controllers only but until post-migration
+        if (node.broker() || node.controller() && kafkaMetadataConfigState.isZooKeeperOrPostMigration()) {
+            securityProtocol.add(REPLICATION_LISTENER_NAME + ":SSL");
+            configureReplicationListener();
+        }
+
         // Non-controller listeners are used only on ZooKeeper based brokers or KRaft brokers (including mixed nodes)
         if (!isKraftControllerOnly) {
             // Replication listener
@@ -286,8 +309,6 @@ public class KafkaBrokerConfigurationBuilder {
                     // Pod name constructed to be templatable for each individual ordinal
                     DnsNameGenerator.podDnsNameWithoutClusterDomain(namespace, KafkaResources.brokersServiceName(clusterName), node.podName())
             ));
-            securityProtocol.add(REPLICATION_LISTENER_NAME + ":SSL");
-            configureReplicationListener();
 
             for (GenericKafkaListener listener : kafkaListeners) {
                 int port = listener.getPort();
@@ -324,10 +345,13 @@ public class KafkaBrokerConfigurationBuilder {
         if (!isKraftControllerOnly) {
             writer.println("advertised.listeners=" + String.join(",", advertisedListeners));
             writer.println("inter.broker.listener.name=" + REPLICATION_LISTENER_NAME);
+        } else if (node.controller() && kafkaMetadataConfigState.isZooKeeperOrPostMigration()) {
+            // needed for KRaft controller only as well until post-migration because it needs to contact brokers
+            writer.println("inter.broker.listener.name=" + REPLICATION_LISTENER_NAME);
         }
 
-        // Control plane listener is on all ZooKeeper based brokers, it's not supported in KRaft mode
-        if (!useKRaft) {
+        // Control plane listener is on all ZooKeeper based brokers, needed during migration as well, when broker still using ZooKeeper but KRaft controllers are ready
+        if (node.broker() && kafkaMetadataConfigState.isZooKeeperOrMigration()) {
             writer.println("control.plane.listener.name=" + CONTROL_PLANE_LISTENER_NAME);
         }
 
@@ -631,7 +655,7 @@ public class KafkaBrokerConfigurationBuilder {
             superUsers.add(String.format("User:CN=%s,O=io.strimzi", "cluster-operator"));
 
             printSectionHeader("Authorization");
-            configureAuthorization(clusterName, superUsers, authorization, useKRaft);
+            configureAuthorization(clusterName, superUsers, authorization, kafkaMetadataConfigState.isKRaftInConfiguration(node));
             writer.println("super.users=" + String.join(";", superUsers));
             writer.println();
         }
@@ -645,11 +669,11 @@ public class KafkaBrokerConfigurationBuilder {
      * @param clusterName Name of the cluster
      * @param superUsers Super-users list who have all the rights on the cluster
      * @param authorization The authorization configuration from the Kafka CR
-     * @param useKRaft      Use KRaft mode in the configuration
+     * @param isKRaft      Use KRaft mode in the configuration
      */
-    private void configureAuthorization(String clusterName, List<String> superUsers, KafkaAuthorization authorization, boolean useKRaft) {
+    private void configureAuthorization(String clusterName, List<String> superUsers, KafkaAuthorization authorization, boolean isKRaft) {
         if (authorization instanceof KafkaAuthorizationSimple simpleAuthz) {
-            configureSimpleAuthorization(simpleAuthz, superUsers, useKRaft);
+            configureSimpleAuthorization(simpleAuthz, superUsers, isKRaft);
         } else if (authorization instanceof KafkaAuthorizationOpa opaAuthz) {
             configureOpaAuthorization(opaAuthz, superUsers);
         } else if (authorization instanceof KafkaAuthorizationKeycloak keycloakAuthz) {
@@ -664,10 +688,10 @@ public class KafkaBrokerConfigurationBuilder {
      *
      * @param authorization     Simple authorization configuration
      * @param superUsers        Super-users list who have all the rights on the cluster
-     * @param useKRaft          Use KRaft mode in the configuration
+     * @param isKRaft          Use KRaft mode in the configuration
      */
-    private void configureSimpleAuthorization(KafkaAuthorizationSimple authorization, List<String> superUsers, boolean useKRaft) {
-        if (useKRaft) {
+    private void configureSimpleAuthorization(KafkaAuthorizationSimple authorization, List<String> superUsers, boolean isKRaft) {
+        if (isKRaft) {
             writer.println("authorizer.class.name=" + KafkaAuthorizationSimple.KRAFT_AUTHORIZER_CLASS_NAME);
         } else {
             writer.println("authorizer.class.name=" + KafkaAuthorizationSimple.AUTHORIZER_CLASS_NAME);
@@ -815,7 +839,7 @@ public class KafkaBrokerConfigurationBuilder {
     public KafkaBrokerConfigurationBuilder withLogDirs(List<VolumeMount> mounts)  {
         // We take all the data mount points and add the broker specific path
         String logDirs = mounts.stream()
-                .map(volumeMount -> volumeMount.getMountPath() + "/kafka-log" + brokerId).collect(Collectors.joining(","));
+                .map(volumeMount -> volumeMount.getMountPath() + "/kafka-log" + node.nodeId()).collect(Collectors.joining(","));
 
         printSectionHeader("Kafka message logs configuration");
         writer.println("log.dirs=" + logDirs);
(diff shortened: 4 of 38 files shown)


cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaBrokerConfigurationBuilder.java
```diff
@@ -70,37 +71,64 @@ public class KafkaBrokerConfigurationBuilder {
...
+        return this.kafkaMetadataConfigState.ordinal() >= KafkaMetadataStateManager.KafkaMetadataConfigurationState.POST_MIGRATION.ordinal();
+    }
+
+    /**
+     * @return if KRaft is going to be configured depending on the node role and the Kafka metadata configuration state
+     */
+    private boolean isKRaftInConfiguration() {
+        return (node.controller() && this.kafkaMetadataConfigState.ordinal() >= KafkaMetadataStateManager.KafkaMetadataConfigurationState.PRE_MIGRATION.ordinal()) ||
+                (node.broker() && this.kafkaMetadataConfigState.ordinal() >= KafkaMetadataStateManager.KafkaMetadataConfigurationState.MIGRATION.ordinal());
+    }
```
scholzj, line 118: Why are these not instance methods of the state?

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaCluster.java
```diff
@@ -240,6 +240,8 @@ public class KafkaCluster extends AbstractModel implements SupportsMetrics, Supp
         }
     }
 
+    private KafkaMetadataStateManager.KafkaMetadataConfigurationState kafkaMetadataConfigState;
+
```
scholzj, line 244: I guess this should be with the other private fields of similar use for example after the line 214?

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaBrokerConfigurationBuilder.java
```diff
@@ -638,11 +690,11 @@ public KafkaBrokerConfigurationBuilder withAuthorization(String clusterName, Kaf
      * @param clusterName Name of the cluster
      * @param superUsers Super-users list who have all the rights on the cluster
      * @param authorization The authorization configuration from the Kafka CR
-     * @param useKRaft      Use KRaft mode in the configuration
+     * @param isKRaft      Use KRaft mode in the configuration
```
scholzj, line 693: We seem to use `useKRaft` in general and not `isKRaft`.

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaBrokerConfigurationBuilder.java
```diff
@@ -657,10 +709,10 @@ private void configureAuthorization(String clusterName, List<String> superUsers,
      *
      * @param authorization     Simple authorization configuration
      * @param superUsers        Super-users list who have all the rights on the cluster
-     * @param useKRaft          Use KRaft mode in the configuration
+     * @param isKRaft          Use KRaft mode in the configuration
```
scholzj, line 712: Same as above.

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaCluster.java
```diff
@@ -1629,13 +1630,67 @@ private String generatePerBrokerConfiguration(NodeRef node, KafkaPool pool, Map<
...
+     * |                | Zk-migration |              |
+     * +----------------+--------------+--------------+
+     * | KRAFT          | KRaft        | KRaft        |
+     * +----------------+--------------+--------------+
+     *
+     * @param node node on which the configuration is applied
+     * @param pool pool hosting the node
+     * @param builder KafkaBrokerConfigurationBuilder instance to use to build the node configuration
+     */
+    private void buildBrokerConfiguration(NodeRef node, KafkaPool pool, KafkaBrokerConfigurationBuilder builder) {
```
scholzj, line 1666: This method is super hard to read. You should at least add some spacing. The INFO messages seem also a bit too much?

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaCluster.java
```diff
@@ -1667,10 +1722,14 @@ public List<ConfigMap> generatePerBrokerConfigurationConfigMaps(MetricsAndLoggin
                 // environment variables are parsed in the container bash scripts
                 data.put(BROKER_LISTENERS_FILENAME, listeners.stream().map(ListenersUtils::envVarIdentifier).collect(Collectors.joining(" ")));
 
-                if (useKRaft) {
+                // controller and broker gets the Cluster ID in different states if during migration
+                // anyway they both get it when in full KRaft-mode
+                if ((node.controller() && this.kafkaMetadataConfigState.ordinal() >= KafkaMetadataStateManager.KafkaMetadataConfigurationState.PRE_MIGRATION.ordinal()) ||
+                        (node.broker() && this.kafkaMetadataConfigState.ordinal() >= KafkaMetadataStateManager.KafkaMetadataConfigurationState.MIGRATION.ordinal())) {
```
scholzj, line 1728: The new line here seems weird. If nothing else, this is how it is done in our other code I think.

```suggestion
                if ((node.controller() && this.kafkaMetadataConfigState.ordinal() >= KafkaMetadataStateManager.KafkaMetadataConfigurationState.PRE_MIGRATION.ordinal())
                        || (node.broker() && this.kafkaMetadataConfigState.ordinal() >= KafkaMetadataStateManager.KafkaMetadataConfigurationState.MIGRATION.ordinal())) {
```

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaCluster.java
```diff
@@ -1800,7 +1859,8 @@ public List<Condition> getWarningConditions() {
      *          broker nodes when KRaft is enabled.
      */
     private Labels brokersSelector()    {
-        if (useKRaft)   {
+        // Starting from the migration phase, brokers should be already selected and used via KRaft
+        if (this.kafkaMetadataConfigState.ordinal() >= KafkaMetadataStateManager.KafkaMetadataConfigurationState.MIGRATION.ordinal()) {
```
scholzj, line 1863: FYI: This looks like the methods in `KafkaBrokerConfigurationBuilder` I suggested should be instance methods instead. It seems they would have a lot of value also in this class.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaReconciler.java
```diff
@@ -196,11 +202,20 @@ public KafkaReconciler(
         this.operationTimeoutMs = config.getOperationTimeoutMs();
         this.kafkaNodePoolCrs = nodePools;
 
+        this.kafkaMetadataStateManager = new KafkaMetadataStateManager(reconciliation, kafkaCr);
+        KafkaMetadataStateManager.KafkaMetadataConfigurationState kafkaMetadataConfigState = this.kafkaMetadataStateManager.computeMetadataConfigurationState();
```
scholzj, line 206: Imagine how many characters you save in all the code if `KafkaMetadataConfigurationState` was a separate class. The code will be a lot more readable.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaReconciler.java
```diff
@@ -196,11 +202,20 @@ public KafkaReconciler(
...
+        LOGGER.infoCr(reconciliation, "KafkaMetadataConfigurationState = {}", kafkaMetadataConfigState);
+
+        // TODO: to evaluate this one, because used only for inter broker protocol version and log message format version below
         boolean isKRaftEnabled = config.featureGates().useKRaftEnabled() && ReconcilerUtils.kraftEnabled(kafkaCr);
+        // KRaft to be considered not only when fully enabled (KRAFT = 4) but also when a migration is about to start (PRE_MIGRATION = 1)
+        // NOTE: this is important to drive the right validation happening in node pools (i.e. roles on node pools, storage, number of controllers, ...)
+        boolean isKRaftInConfiguration = config.featureGates().useKRaftEnabled() &&
```
scholzj, line 213: `isKRaftInConfiguration` means what? It sounds like whether user configured something. I think the name is confusing.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaReconciler.java
```diff
@@ -1040,6 +1057,40 @@ protected Future<Void> updateKafkaVersion(KafkaStatus kafkaStatus) {
...
+        kafkaStatus.setKafkaMetadataState(kafkaMetadataStateManager.computeNextMetadataState().name());
+        return Future.succeededFuture();
+    }
+
+    protected Future<Void> maybeKafkaMetadataMigrationInProgress() {
+        KafkaMetadataStateManager.KafkaMetadataConfigurationState kafkaMetadataConfigState = this.kafkaMetadataStateManager.computeMetadataConfigurationState();
+        // on each reconcile, would be useless to check migration status if it's not going on
+        if (kafkaMetadataConfigState.equals(KafkaMetadataStateManager.KafkaMetadataConfigurationState.MIGRATION)) {
+            LOGGER.infoCr(reconciliation, "Checking ZooKeeper migration state");
+            return ReconcilerUtils.clientSecrets(reconciliation, secretOperator)
```
scholzj, line 1070: How many times are we getting the secrets? Aren't they already stored somewhere?

scholzj (commented):
Some comments from an initial pass ...

.checkstyle/import-control.xml
```diff
@@ -78,6 +78,7 @@ We also control imports only in production classes and not in tests. This is con
         <allow class="io.strimzi.operator.common.Annotations" />
         <allow class="io.strimzi.operator.cluster.ClusterOperatorConfig" />
         <allow class="io.strimzi.operator.cluster.PlatformFeaturesAvailability" />
+        <allow class="io.strimzi.operator.cluster.operator.resource.KafkaMetadataStateManager" />
```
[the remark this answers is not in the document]
scholzj: Looking at the code while working on something else, it looks like the majority of classes like this are in this package. So maybe you can ignore this comment for the time being. I still think the placement makes no sense, but we should probably deal with it _globally_ and not just with this class 🤷.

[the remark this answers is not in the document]
ppatierno: Since the initial development the class was in the `assembly` package as you were suggesting. When I needed to use the `KafkaAgentClient` in it, I noticed that such a class is package-private. Tbh I was not sure to make the `KafkaAgentClient` public so I did the other way around, moving the state manager under `resource`.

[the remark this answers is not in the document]
scholzj: Well, as I said, I do not think it makes any sense for the KafkaAgentCLient to be in the `resource` package. It is not managing any resource. But as suggested, I think you should proceed as it is and we can move all of them later 🤷

api/src/main/java/io/strimzi/api/kafka/model/Kafka.java
```diff
@@ -62,6 +62,11 @@
                 description = "The state of the custom resource",
                 jsonPath = ".status.conditions[?(@.type==\"Ready\")].status",
                 type = "string"),
+            @Crd.Spec.AdditionalPrinterColumn(
```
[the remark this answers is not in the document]
ppatierno: This represents the cluster metadata state, not just about migration (even if it's kind of related of course). It also shows `ZooKeeper` or `KRaft`, while also "migration" related state like (`KRaftMigration` or `KRaftDualWriting`).

api/src/main/java/io/strimzi/api/kafka/model/status/KafkaStatus.java
```diff
@@ -35,6 +35,8 @@ public class KafkaStatus extends Status {
     private String operatorLastSuccessfulVersion;
     private String kafkaVersion;
 
+    private String kafkaMetadataState;
```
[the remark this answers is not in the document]
ppatierno: No need for it :-)

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaBrokerConfigurationBuilder.java
```diff
@@ -70,37 +71,64 @@ public class KafkaBrokerConfigurationBuilder {
...
+     * @param node              NodeRef instance
+     * @param kafkaMetadataConfigState  Represents the state of the Kafka metadata configuration
      */
-    public KafkaBrokerConfigurationBuilder(Reconciliation reconciliation, String brokerId, boolean useKRaft) {
+    public KafkaBrokerConfigurationBuilder(Reconciliation reconciliation, String brokerId, boolean useKRaft, NodeRef node, KafkaMetadataStateManager.KafkaMetadataConfigurationState kafkaMetadataConfigState) {
         printHeader();
         this.reconciliation = reconciliation;
         this.brokerId = brokerId;
         this.useKRaft = useKRaft;
+        this.node = node;
```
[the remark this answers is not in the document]
ppatierno: Yeah, agree.

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaCluster.java
ppatierno: Not sure where you think spacing makes it more readable, if it's really hard to read.
Regarding the INFO logging, it will become TRACE level in the end. Currently, I am using INFO for having a better testing feedback without the need to increase the overall debugging level of the operator.

scholzj: The block has 50 lines of various ifs without a single empty line between them. That makes it hard to read.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaReconciler.java
ppatierno: You can't imagine how much I was struggling and debating with myself on this, the code looks horrible and too long to me as well. My thought was trying to have it as an inner class because strictly related to the state manager. I am going to move it out as a separate class, at least I will have approval from one of the maintainers to do it :-)

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaReconciler.java
```diff
@@ -1040,6 +1057,40 @@ protected Future<Void> updateKafkaVersion(KafkaStatus kafkaStatus) {
...
+                    .compose(compositeFuture -> {
+                        // we should get the quorum controller leader using the Admin Client API describeMetadataQuorum() but
+                        // it's not supported by brokers during migration because they are still connected to ZooKeeper so ...
+                        // picking a random controller from the set, because all expose the needed metrics
+                        NodeRef[] controllers = kafka.controllerNodes().toArray(new NodeRef[kafka.controllerNodes().size()]);
+                        int controllerId = controllers[RANDOM.nextInt(kafka.controllerNodes().size())].nodeId();
```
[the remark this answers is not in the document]
ppatierno: It was a way to avoid picking the same controller consistently if there is any kind of problem with it rolling, restarting or any other problem and not able to provide the metrics we need. Getting the leader would be the best solution but as per comment above it's not possible. Why do you think it doesn't make any sense?

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaBrokerConfigurationBuilder.java
```diff
@@ -70,37 +71,64 @@ public class KafkaBrokerConfigurationBuilder {
...
      */
     private void configureNodeOrBrokerId()   {
         printSectionHeader("Node / Broker ID");
 
-        if (useKRaft) {
-            writer.println("node.id=" + brokerId);
-        } else {
+        // Node ID is ignored when not using Kraft mode => but it defaults to broker ID when not set.
+        // We set it here in the configuration explicitly to avoid never ending rolling updates.
+        writer.println("node.id=" + brokerId);
```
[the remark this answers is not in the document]
ppatierno: In the previous code, the `node.id` was set in both Zk and KRaft clusters and this is going to happen with this code as well, so nothing changed actually. The current `if` is now adding the `broker.id` not only if cluster is Zk but also during migration (because brokers are still connected to Zk and this field is needed).

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaReconciler.java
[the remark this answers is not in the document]
scholzj: Well, it is not consistent when it is random. It is the opposite, irreproducible. If the first node has a problem, I guess I would also wonder what it really means for the cluster. But also, the likelihood of the second or third node having problems is the same as the first one. 

I think for example having a `while` loop over all nodes until you find the answer would be a better approach than the random one and it would also have better guarantees of getting the information from at least some of the nodes. And it would save us keeping the Randon object in the class as well.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaReconciler.java
scholzj: I would probably split it. But I guess you could also consider renaming it if you don't want to split it. You for example do not need to have the `KafkaMetadata` twice there if you keep it as an inner class.

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaBrokerConfigurationBuilder.java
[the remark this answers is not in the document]
scholzj: You are right. I misread the code change.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaReconciler.java
ppatierno: I decided to split it. Let's see if it's better. We can reconsider later or if others will have a different comment on this.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaReconciler.java
ppatierno: In this specific case, we are getting secrets only when there is a "migration" ongoing.
I can't find a place where they are stored tbh. I see that the same way of doing that (so by using `ReconcilerUtils.clientSecrets`) is also done on each reconcile when retrieving the cluster id and during the "maybe" roll Kafka part.
Maybe we could retrieve them one time at the beginning of the reconciliation in order to be used in these three specific steps but ... 
1. not sure if any certificates change could have an impact on using "something" you retrieve at the beginning of the reconcile (even if reconciling certs happens before reconciling Kafka so we should not have such a problem)
2. I would leave this for a different PR because it would have impact on code which is not migration related only

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaCluster.java
ppatierno: Yeah, I will make the changes.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaReconciler.java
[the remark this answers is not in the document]
ppatierno: I have refactored using a loop, let me know what you think please.

documentation/modules/appendix_crds.adoc
```diff
@@ -1628,6 +1628,8 @@ Used in: xref:type-Kafka-{context}[`Kafka`]
 |string
 |kafkaVersion                   1.2+<.<a|The version of Kafka currently deployed in the cluster.
 |string
+|kafkaMetadataState             1.2+<.<a|Defines where cluster metadata are stored: ZooKeeper, KRaft or there is a migration running.
```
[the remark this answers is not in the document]
ppatierno: Yeah, good idea. I listed all of them.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaReconciler.java
ppatierno: Compared to the old code, it's not feature gate enabled AND cluster is full KRaft anymore. 
Now you could have KRaft configured in a node but the cluster is not full KRaft yet because there is a migration going on (i.e. from the PRE_MIGRATION state up to KRAFT). This is the point I am trying to make with this name ... something like "KRaft IS in the configuration but cluster is NOT necessarily KRaft yet".
It means that even in this case the node pools creation stuff for example has to run some validation (i.e. warning about having just 2 controllers, when you create them to migrate).
Do you have any better suggestion?

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaBrokerConfigurationBuilder.java
ppatierno: So as related to the other comment, it's more about "is KRaft in the current configuration" (which doesn't mean it's a full KRaft cluster). I am using the `useKRaft` more as "KRaft is enabled" through the feature gate flag.

Anyway, I am not ANDing this feature gate flag with the rest of checks so maybe something is missing here.
If KRaft feature gate is disabled and you try to deploy KRaft or migration you should got an error ... it's something I have to add.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaAssemblyOperator.java
```diff
@@ -209,6 +209,8 @@ public Future<KafkaStatus> createOrUpdate(Reconciliation reconciliation, Kafka k
     Future<Void> reconcile(ReconciliationState reconcileState)  {
         Promise<Void> chainPromise = Promise.promise();
 
+        // TODO: to think about this, because is it still right to take into account only when KRaft is fully enabled or even
+        //       taking into account when KRaft configuration is enabled furing a migration? (as already happens in other parts)
```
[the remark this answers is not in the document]
ppatierno: Yeah. I moved the metadata state manager to the reconciliation state to be available here and then pushed down to the Kafka reconciler for doing the rest.

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaBrokerConfigurationBuilder.java
scholzj: > So as related to the other comment, it's more about "is KRaft in the current configuration" (which doesn't mean it's a full KRaft cluster).

You can also call it `useKRaft`, be consistent with what we use elsewhere and say that it is about _using KRaft in the configuration_ 😉

fvaleri (commented):
Hi @ppatierno, I left some comments, but looks good.

I also tried to run a migration after merging the Kafka Roller changes. It worked fine until KRaftDualWriting state, where I got this error when it was trying to roll the controllers.

```sh
2023-11-23 18:09:04 WARN  KafkaQuorumCheck:65 - Reconciliation #173(timer) Kafka(test/my-cluster): Error determining the controller quorum leader id
org.apache.kafka.common.errors.UnsupportedVersionException: The broker does not support DESCRIBE_QUORUM
```

After manually restarting the controllers one by one, the migration completed successfully. I verified that broker were connected to the quorum and working.

```sh
NAME         DESIRED KAFKA REPLICAS   DESIRED ZK REPLICAS   READY   METADATA STATE   WARNINGS
my-cluster   3                        3                                              
my-cluster   3                        3                     True    ZooKeeper        
my-cluster   3                        3                     True    ZooKeeper        
my-cluster   3                        3                                              
my-cluster   3                        3                                              
my-cluster   3                        3                     True    KRaftMigration   
my-cluster   3                        3                     True    KRaftDualWriting   
my-cluster   3                        3                     True    KRaftDualWriting   
my-cluster   3                        3                                                
my-cluster   3                        3                                                
my-cluster   3                        3                     True    KRaft  
```

I'm also attaching the CO log for this test.
[zk2kraft.log](https://github.com/strimzi/strimzi-kafka-operator/files/13452789/zk2kraft.log)

ppatierno:
> I also tried to run a migration after merging the Kafka Roller changes.

@fvaleri thanks for testing the two PRs working together! 

> It worked fine until KRaftDualWriting state, where I got this error when it was trying to roll the controllers.

So what you are seeing is expected and it's a problem for us. When it's the time to roll the controllers (so moving from the "dual write" mode to the full KRaft), the controllers (and the brokers as well) are still connected to ZooKeeper and in that state, the `describeQuorum` API doesn't work. This adds an additional complexity to the roller which needs to be able to determine the leader in a different way in such a case than using the `describeQuorum` method. @tinaselenge @katheris I think it's something to think about in your PR, otherwise it's not going to work with migration.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/resource/KafkaMetadataStateManager.java
```diff
@@ -0,0 +1,397 @@
...
+ * It is also used to compute metadata state changes during the migration process from ZooKeeper to KRaft (or rollback)
+ */
+public class KafkaMetadataStateManager {
+
+    private static final ReconciliationLogger LOGGER = ReconciliationLogger.create(KafkaMetadataStateManager.class.getName());
+
+    /**
+     * The annotation value which indicates that the ZooKeeper to KRaft migration is enabled
+     */
+    public static final String MIGRATION_VALUE_STRIMZI_IO_KRAFT = "migration";
```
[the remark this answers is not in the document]
ppatierno: I was actually thinking the other way around. At some point in the future, we won't have migration anymore and only KRaft based cluster. The `KafkaMetadataStateManager` class will be removed and with it all the related metadata stuff, because it will be all in KRaft.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/resource/KafkaMetadataStateManager.java
```diff
@@ -0,0 +1,397 @@
...
+     * Starting from the current internal FSM state and taking into account the strimzi.io/kraft annotation value,
+     * it computes a state representing the desired configuration for nodes (brokers and/or controllers)
+     *
+     * @return the next desired nodes configuration related state
+     */
+    public KafkaMetadataConfigurationState computeMetadataConfigurationState() {
+        switch (this.metadataState) {
+            case ZooKeeper -> {
+                // cluster is still using ZooKeeper, but controllers need to be deployed with ZooKeeper and migration enabled
+                if (this.isKRaftMigration(this.kraftAnno)) {
```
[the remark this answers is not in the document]
ppatierno: Yes, it's part of a cleaning work I have to do on this class. The same applies on all the `onXXXX` private methods. I will make the change for all of them.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/resource/KafkaMetadataStateManager.java
```diff
@@ -0,0 +1,397 @@
...
+    }
+
+    /**
+     * Compute the next desired nodes configuration related state
+     * Starting from the current internal FSM state and taking into account the strimzi.io/kraft annotation value,
+     * it computes a state representing the desired configuration for nodes (brokers and/or controllers)
+     *
+     * @return the next desired nodes configuration related state
+     */
+    public KafkaMetadataConfigurationState computeMetadataConfigurationState() {
```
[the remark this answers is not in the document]
ppatierno: Actually it has a different meaning ...

The `computeNextMetadataState` is really about computing the next state of the internal FSM from the current one.
The `computeMetadataConfigurationState` is about getting the metadata configuration state from the current internal FSM state, so it's not moving the FSM through states. I was wondering if it should be more a `get` instead of a `compute`.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/resource/KafkaMetadataStateManager.java
```diff
@@ -0,0 +1,397 @@
...
+            case KRaftMigration -> onKRaftMigration(reconciliation, currentKraftAnno);
+            case KRaftDualWriting -> onKRaftDualWriting(reconciliation, currentKraftAnno);
+            case KRaftPostMigration -> onKRaftPostMigration(reconciliation, currentKraftAnno);
+        };
+        LOGGER.infoCr(reconciliation, "from [{}] to [{}] with strimzi.io/kraft: [{}]", currentState, this.metadataState, currentKraftAnno);
+        return this.metadataState;
+    }
+
+    /**
+     * Compute the next desired nodes configuration related state
```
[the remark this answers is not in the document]
ppatierno: Not sure about this ... it's more the "node configuration" (so if ZooKeeper and/or KRaft are needed) than the "metadata configuration". I mean from the metadata state, we are getting which node configuration we need, and this is represented by a "state". Maybe the overall naming is bad :-D

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/resource/KafkaMetadataStateManager.java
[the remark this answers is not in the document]
fvaleri: Right, let's put them all together in KafkaMetadataStateManager.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/resource/KafkaMetadataStateManager.java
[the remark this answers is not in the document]
fvaleri: Ok, I think "get" is better.

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaBrokerConfigurationBuilder.java
```diff
@@ -250,14 +270,14 @@ public KafkaBrokerConfigurationBuilder withListeners(
         List<String> advertisedListeners = new ArrayList<>();
         List<String> securityProtocol = new ArrayList<>();
 
-        boolean isKraftControllerOnly = useKRaft && node.controller() && !node.broker();
-        boolean isKraftBrokerOnly = useKRaft && node.broker() && !node.controller();
+        boolean isKraftControllerOnly = useKRaft && kafkaMetadataConfigState.isKRaftInConfiguration(node) && node.controller() && !node.broker();
 
-        // Control Plane listener => ZooKeeper based brokers and KRaft controllers (including mixed nodes). But not on KRaft broker only nodes
-        if (!isKraftBrokerOnly) {
+        // Control Plane listener set for any KRaft controller, only or mixed, and broker in ZooKeeper-based or migration but NOT when full KRaft
```
tinaselenge, line 275:
```suggestion
        // Control Plane listener is set for pure KRaft controller or combined node, and broker in ZooKeeper mode or in migration state. However, it will NOT be set when in full KRaft mode. 
```

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaReconciler.java
```diff
@@ -286,7 +300,9 @@ public Future<Void> reconcile(KafkaStatus kafkaStatus, Clock clock)    {
                 // This has to run after all possible rolling updates which might move the pods to different nodes
                 .compose(i -> nodePortExternalListenerStatus())
                 .compose(i -> addListenersToKafkaStatus(kafkaStatus))
-                .compose(i -> updateKafkaVersion(kafkaStatus));
+                .compose(i -> updateKafkaVersion(kafkaStatus))
+                .compose(i -> maybeKafkaMetadataMigrationInProgress())
```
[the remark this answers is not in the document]
ppatierno: What do you mean by that? Can you elaborate a little bit more please?

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaReconciler.java
ppatierno: That line was removed and simplified.

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaBrokerConfigurationBuilder.java
ppatierno: This parameter was removed from the builder, because not needed anymore.

ppatierno:
Closing this PR which is now replaced by https://github.com/strimzi/strimzi-kafka-operator/pull/9480 having same content but originating from the `zk-kraft-migration` branch living on the Stimzi official repo and not on a fork anymore.

PR #9332, opened 2023-11-07
[DOC] Improve topic replication factor change procedure

fvaleri:
This is still based on the original secure cluster setup (mTLS), as it is common in production. The UTO is not able to reconcile a replication factor change done in Kafka, so I also show how to fix the reconciliation error.

### DIFF
--- a/documentation/modules/configuring/proc-changing-topic-replicas.adoc
+++ b/documentation/modules/configuring/proc-changing-topic-replicas.adoc
@@ -13,103 +13,123 @@ and using a reassignment file to describe how the topic replicas should be chang
 
 This procedure describes a secure process that uses TLS.
 You'll need a Kafka cluster that uses TLS encryption and mTLS authentication.
-A topic called `my-topic` has only 1 replica and we want to increase it to 3 replicas.
 
 .Prerequisites
 
 * You have a running Kafka cluster based on a `Kafka` resource configured with internal TLS encryption and mTLS authentication.
+* You are running an interactive pod container that is connected to the running Kafka broker.
+* You have generated a reassignment JSON file named `reassignment.json`.
 * You are connected as a `KafkaUser` configured with ACL rules that specify permission to manage the Kafka cluster and its topics.
-* You have generated a reassignment JSON file named reassignment.json.
 
 See xref:proc-generating-reassignment-json-files-{context}[Generating reassignment JSON files].
 
+In this procedure, a topic called `my-topic` has 4 replicas and we want to reduce it to 3.
+A JSON file named `topics.json` specifies the topic, and was used to generate the `reassignment.json` file.
+
+.Example JSON file specifies `my-topic`
+[source,json]
+----
+{
+  "version": 1,
+  "topics": [
+    { "topic": "my-topic"}
+  ]
+}
+----
+
 .Procedure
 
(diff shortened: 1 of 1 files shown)


scholzj (commented):
I think you should stick with how it was and update only the parts relevant to UTO. Not rewrtie the commands _your way_.

fvaleri:
> I think you should stick with how it was and update only the parts relevant to UTO. Not rewrtie the commands _your way_.

Wdym? Most of these commands where missing.

scholzj:
The commands as you wrote them will anyway not work for half of the readers. We also do not have the resources to test the commands every release. So I think you should stick with how it was and update only what is relevant to UTO.

fvaleri:
These commands should work for users in that specific use case. Other can easily adjust. I don't think the current procedure makes it particularly easy to do that operation, as you have to jump through various sections, but if you are happy with it.

documentation/modules/configuring/proc-changing-topic-replicas.adoc
```diff
@@ -130,4 +130,12 @@ bin/kafka-topics.sh --bootstrap-server
...
\ No newline at end of file
+----
+
+. Finally, patch the `KafkaTopic` custom resource to match the new replication factor, and let the operator do the rest
+[source,shell,subs=+quotes]
+----
+kubectl patch kt my-topic --type merge -p '
+  spec:
+    replicas: 3'
+----
```
scholzj, line 141: I would just write it in a generic way:

```suggestion
. Finally, update the `KafkaTopic` resource to match the new replication factor (in `.spec.replicas`).
```

Many users use various tooling and cannot do a patch command like this.

scholzj (changes requested):

scholzj (approved):

documentation/modules/configuring/proc-changing-topic-replicas.adoc
```diff
@@ -130,4 +130,18 @@ bin/kafka-topics.sh --bootstrap-server
...
+apiVersion: kafka.strimzi.io/v1beta2
+kind: KafkaTopic
+metadata:
+  name: my-topic
+  labels:
+    strimzi.io/cluster: my-cluster
+spec:
+  partitions: 3
+  replicas: 3
+----
```
PaulRMellor, line 147:
```suggestion
. Finally, edit the `KafkaTopic` custom resource to change `.spec.replicas` to 3, and then wait the reconciliation.
+
[source,shell,subs=+quotes]
----
apiVersion: kafka.strimzi.io/v1beta2
kind: KafkaTopic
metadata:
  name: my-topic
  labels:
    strimzi.io/cluster: my-cluster
spec:
  partitions: 3
  replicas: 3
----
```

PaulRMellor: It needs that + on line 136 to bring the example config inline with the step.

PaulRMellor (approved):
Thanks Fede. Looks good

documentation/modules/configuring/proc-changing-topic-replicas.adoc
fvaleri: Thanks Paul.

scholzj:
Thanks for the PR.

PR #9350, opened 2023-11-15
[ST] Disable two tests and enable OTel test with Connect for MicroShift

im-konge:
### Type of change

- Enhancement 

### Description

This PR disables two tests for MicroShift:

- `DrainCleanerST` - on MicroShift we are not able do the node drain that is used in this suite
- `ConnectST#testConnectorTaskAutoRestart` - this test uses the Connect build feature for creating Connect image with Echo-sink connector plugin

Additionally, this PR enables the `OpenTelemetryST` test for MicroShift (as well for other clusters in case that we have pre-built Connect image with file-sink plugin)

### Checklist


see-quick (approved):
👍

PR #9363, opened 2023-11-19
Remove remaining uses of executeBlocking from Topic Operator

fvaleri:
This should complete #9233.

### DIFF
--- a/topic-operator/src/main/java/io/strimzi/operator/topic/Session.java
+++ b/topic-operator/src/main/java/io/strimzi/operator/topic/Session.java
@@ -25,7 +25,6 @@ import org.apache.kafka.clients.admin.AdminClient;
 import org.apache.kafka.clients.admin.AdminClientConfig;
 import org.apache.kafka.common.config.SslConfigs;
 import org.apache.kafka.common.config.SaslConfigs;
-import org.apache.kafka.common.errors.TimeoutException;
 import org.apache.kafka.streams.StreamsConfig;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
@@ -39,6 +38,9 @@ import java.util.Properties;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.CompletionStage;
 import java.util.concurrent.ExecutionException;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.TimeUnit;
 import java.util.function.BiFunction;
 
 import static io.strimzi.operator.topic.Config.ZOOKEEPER_CONNECT;
@@ -134,71 +136,99 @@ public class Session extends AbstractVerticle {
      */
     @Override
     public void stop(Promise<Void> stop) throws Exception {
+        LOGGER.info("Stopping");
         this.stopped = true;
-        Long timerId = this.timerId;
-        if (timerId != null) {
-            vertx.cancelTimer(timerId);
+        if (this.timerId != null) {
+            vertx.cancelTimer(this.timerId);
         }
-        Promise<Promise<Void>> blockingResult = Promise.promise();
-        vertx.executeBlocking(() -> {
-            long timeout = 120_000L;
-            long deadline = System.currentTimeMillis() + timeout;
-            LOGGER.info("Stopping");
-            LOGGER.debug("Stopping kube watch");
+        ExecutorService shutdownExecutor = Executors.newSingleThreadExecutor();
+        closeKubernetesWatch()
+                .onSuccess(ignored -> LOGGER.debug("Kubernetes watch closed"))
+                .onFailure(cause -> LOGGER.error("Failed to close Kubernetes watch", cause))
+            .compose(ignored -> stopZooKeeperWatcher())
+                .onSuccess(ignored -> LOGGER.debug("ZooKeeper topics watcher stopped"))
+                .onFailure(cause -> LOGGER.error("Failed to stop ZooKeeper topics watcher", cause))
+            .compose(ignored -> stopTopicStoreService())
+                .onSuccess(ignored -> LOGGER.debug("Topic store service stopped"))
+                .onFailure(cause -> LOGGER.error("Failed to stop topic store service", cause))
+            .compose(ignored -> disconnectFromZk(shutdownExecutor))
+                .onSuccess(ignored -> LOGGER.debug("ZooKeeper disconnected"))
+                .onFailure(cause -> LOGGER.error("Failed to disconnect from ZooKeeper", cause))
+            .compose(ignored -> closeKafkaAdminClient(shutdownExecutor))
+                .onSuccess(ignored -> LOGGER.debug("Kafka admin client closed"))
+                .onFailure(cause -> LOGGER.error("Failed to close Kafka admin client", cause))
+            .compose(ignored -> closeHttpHealthServer())
+                .onSuccess(ignored -> LOGGER.debug("HTTP health server closed"))
+                .onFailure(cause -> LOGGER.error("Failed to close HTTP health server", cause))
+            .compose(ignored -> {
+                try {
+                    shutdownExecutor.shutdown();
+                    shutdownExecutor.awaitTermination(10, TimeUnit.SECONDS);
+                } catch (InterruptedException e) {
+                    if (!shutdownExecutor.isTerminated()) {
+                        shutdownExecutor.shutdownNow();
+                    }
+                }
+                LOGGER.info("Stopped");
+                return Future.future(handle -> stop.complete());
+            });
+    }
+
+    private Future<Void> closeKubernetesWatch() {
+        LOGGER.debug("Closing Kubernetes watch");
+        if (topicWatch != null) {
             topicWatch.close();
-            LOGGER.debug("Stopping zk watches");
+        }
+        return Future.succeededFuture();
+    }
+
+    private Future<Void> stopZooKeeperWatcher() {
+        LOGGER.debug("Stopping ZooKeeper watcher");
+        if (topicsWatcher != null) {
             topicsWatcher.stop();
+        }
+        return Future.succeededFuture();
+    }
 
-            Promise<Void> promise = Promise.promise();
-            Handler<Long> longHandler = new Handler<>() {
-                @Override
-                public void handle(Long inflightTimerId) {
-                    if (!topicOperator.isWorkInflight()) {
-                        LOGGER.debug("Inflight work has finished");
-                        promise.complete();
-                    } else if (System.currentTimeMillis() > deadline) {
-                        LOGGER.error("Timeout waiting for inflight work to finish");
-                        promise.complete();
-                    } else {
-                        LOGGER.debug("Waiting for inflight work to finish");
-                        vertx.setTimer(1_000, this);
-                    }
-                }
-            };
-            longHandler.handle(null);
+    private Future<Void> stopTopicStoreService() {
+        LOGGER.debug("Stopping topic store service");
+        if (service != null) {
+            service.stop();
+        }
+        return Future.succeededFuture();
+    }
 
-            promise.future().compose(ignored -> {
-                if (service != null) {
-                    service.stop();
-                }
-                return Future.succeededFuture();
-            });
-
-            promise.future().compose(ignored -> {
-                LOGGER.debug("Disconnecting from zookeeper {}", zk);
-                zk.disconnect(zkResult -> {
-                    if (zkResult.failed()) {
-                        LOGGER.warn("Error disconnecting from zookeeper: {}", String.valueOf(zkResult.cause()));
+    private Future<Void> disconnectFromZk(ExecutorService shutdownExecutor) {
+        LOGGER.debug("Disconnecting from ZooKeeper");
+        Promise<Void> promise = Promise.promise();
+        if (zk != null) {
+            shutdownExecutor.submit(() -> {
+                zk.disconnect(result -> {
+                    if (result.failed()) {
+                        LOGGER.warn("Error disconnecting from ZooKeeper: {}", String.valueOf(result.cause()));
+                        promise.fail(result.cause());
                     }
-                    long timeoutMs = Math.max(1, deadline - System.currentTimeMillis());
-                    LOGGER.debug("Closing AdminClient {} with timeout {}ms", adminClient, timeoutMs);
-                    try {
-                        adminClient.close(Duration.ofMillis(timeoutMs));
-                        HttpServer healthServer = this.healthServer;
-                        if (healthServer != null) {
-                            healthServer.close();
-                        }
-                    } catch (TimeoutException e) {
-                        LOGGER.warn("Timeout while closing AdminClient with timeout {}ms", timeoutMs, e);
-                    } finally {
-                        LOGGER.info("Stopped");
-                        blockingResult.complete();
-                    }
+                    promise.complete();
                 });
-                return Future.succeededFuture();
             });
-            return null;
-        }).onComplete(ar -> stop.complete());
+        }
+        return promise.future();
+    }
+
+    private Future<Void> closeKafkaAdminClient(ExecutorService shutdownExecutor) {
+        LOGGER.debug("Closing Kafka admin client");
+        if (adminClient != null) {
+            shutdownExecutor.submit(() -> adminClient.close(Duration.ofSeconds(2)));
+        }
+        return Future.succeededFuture();
+    }
+
+    private Future<Void> closeHttpHealthServer() {
+        LOGGER.debug("Closing HTTP health server");
+        if (healthServer != null) {
+            healthServer.close();
+        }
+        return Future.succeededFuture();
     }
 
     /**
--- a/topic-operator/src/main/java/io/strimzi/operator/topic/Session.java
+++ b/topic-operator/src/main/java/io/strimzi/operator/topic/Session.java
@@ -228,10 +228,8 @@ public class Session extends AbstractVerticle {
                 zk.disconnect(result -> {
                     if (result.failed()) {
                         LOGGER.warn("Error disconnecting from ZooKeeper: {}", String.valueOf(result.cause()));
-                        promise.fail(result.cause());
-                    } else {
-                        promise.complete();
                     }
+                    promise.complete();
                 });
                 return null;
             });
--- a/topic-operator/src/main/java/io/strimzi/operator/topic/Session.java
+++ b/topic-operator/src/main/java/io/strimzi/operator/topic/Session.java
@@ -133,14 +133,14 @@ public class Session extends AbstractVerticle {
      * Stop the operator.
      */
     @Override
-    @SuppressWarnings("deprecation") // Uses a deprecated executeBlocking call that should be addressed later. This is tracked in https://github.com/strimzi/strimzi-kafka-operator/issues/9233
     public void stop(Promise<Void> stop) throws Exception {
         this.stopped = true;
         Long timerId = this.timerId;
         if (timerId != null) {
             vertx.cancelTimer(timerId);
         }
-        vertx.executeBlocking(blockingResult -> {
+        Promise<Promise<Void>> blockingResult = Promise.promise();
+        vertx.executeBlocking(() -> {
             long timeout = 120_000L;
             long deadline = System.currentTimeMillis() + timeout;
             LOGGER.info("Stopping");
@@ -175,7 +175,6 @@ public class Session extends AbstractVerticle {
             });
 
             promise.future().compose(ignored -> {
-
                 LOGGER.debug("Disconnecting from zookeeper {}", zk);
                 zk.disconnect(zkResult -> {
                     if (zkResult.failed()) {
@@ -198,7 +197,8 @@ public class Session extends AbstractVerticle {
                 });
                 return Future.succeededFuture();
             });
-        }, stop);
+            return null;
+        }).onComplete(ar -> stop.complete());
     }
 
     /**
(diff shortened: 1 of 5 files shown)


scholzj (commented):
I don't see any obvious issues. But the CI failure seems related to this.

topic-operator/src/main/java/io/strimzi/operator/topic/Session.java
```diff
@@ -133,14 +133,14 @@ public void init(Vertx vertx, Context context) {
...
     @Override
-    @SuppressWarnings("deprecation") // Uses a deprecated executeBlocking call that should be addressed later. This is tracked in https://github.com/strimzi/strimzi-kafka-operator/issues/9233
     public void stop(Promise<Void> stop) throws Exception {
         this.stopped = true;
         Long timerId = this.timerId;
         if (timerId != null) {
             vertx.cancelTimer(timerId);
         }
-        vertx.executeBlocking(blockingResult -> {
+        Promise<Promise<Void>> blockingResult = Promise.promise();
```
tombentley, line 142: I don't think this is correct. `stop` is getting completed based on the `Future` returned by `executeBlocking`, not based on when the `blockingResult` is completed (in fact, no code here is dependent on `blockingResult` completing). By my reading of the code, `stop` will be completed once `longHandler.handle()` has been called once, and all the rest of the code here will run _after_ that. 

I think what's happened here (if this _ever_ worked correctly) is that originally we were just using the `executeBlocking` to block on futures in the appropriate order while shutting down. Over time people have reported blocked thread warnings because (for example) sometimes some of those things just happened to take longer than normal. Those have been 'fixed' by (erroneously) making the shutdown of components async, and forgetting about any ordering between them in the process.

The root of this problem is that vertx provides this primitive for blocking execution, but places this arbitrary soft limitation on how long a blocking operation might take. We used that primitive without any of the methods we were calling offering the guarantee of completing within the soft limit vertx has imposed. We ought to have either said "Sometimes you get the warning from vertx, just because closing your admin client/disconnecting from ZK is slower than expected. It's not a bug", or we should have spun up our own executor for shutting things down with as much blocking as needed without the vertx warnings.

Although there's only ever one `Session` within the lifetime of a `Main.main()` we should still try to fix the shutdown so that components are shutdown in an orderly manner, but it's a bit more work than what we have here.

topic-operator/src/main/java/io/strimzi/operator/topic/zk/ZkImpl.java
```diff
@@ -48,69 +49,61 @@ protected ZkImpl(Vertx vertx, ZkClient zkClient) {
...
                 } catch (Throwable t) {
-                    future.fail(t);
+                    result.fail(t);
                 }
-            },
-            handler);
+                return null;
+            }).onComplete(ar -> handler.handle(result.future()));
         return this;
     }
```
tombentley, line 67: Why can't we code it like this:

```java
    @Override
    public Zk create(String path, byte[] data, List<ACL> acls, CreateMode createMode, Handler<AsyncResult<Void>> handler) {
        workerExecutor.<Void>executeBlocking(
            () -> {
                zookeeper.create(path, data == null ? new byte[0] : data, acls, createMode);
                return null;
            }).onComplete(handler);
        return this;
    }
```

topic-operator/src/main/java/io/strimzi/operator/topic/Session.java
fvaleri: Currently, the ShutdownHook sets a hard cap of 10 seconds to complete Vertx shutdown, which I think it's fine. As you say, we can simply try to stop/close all components sequentially, using our own executor where needed.

topic-operator/src/main/java/io/strimzi/operator/topic/zk/ZkImpl.java
fvaleri: Right, thanks.

fvaleri:
@tombentley thanks. I applied the changes you suggested, and also reworked the Session.stop. Please have a look.

topic-operator/src/main/java/io/strimzi/operator/topic/Session.java
```diff
@@ -133,72 +136,102 @@ public void init(Vertx vertx, Context context) {
...
-                        LOGGER.error("Timeout waiting for inflight work to finish");
-                        promise.complete();
-                    } else {
-                        LOGGER.debug("Waiting for inflight work to finish");
-                        vertx.setTimer(1_000, this);
+        ExecutorService shutdownExecutor = Executors.newSingleThreadExecutor();
+        closeKubernetesWatch()
+                .onSuccess(ignored -> LOGGER.debug("Kubernetes watch closed"))
+                .onFailure(cause -> LOGGER.error("Failed to close Kubernetes watch", cause))
+            .compose(ignored -> stopZooKeeperWatcher())
```
tombentley, line 149: I think we could close both kube and ZK watches in parallel. The benefit of doing that is it leaves more time (out of the total 10s hard limit applied by the shutdown hook) to cleanly shutdown the topic store. Note that we don't actually care too much about waiting for things like underlying TCP connections to close. All we really need is that new events stop being delivered to the operator so it can work through in flight events prior to the closing of the topic store.

topic-operator/src/main/java/io/strimzi/operator/topic/Session.java
```diff
@@ -133,72 +136,102 @@ public void init(Vertx vertx, Context context) {
...
+                .onSuccess(ignored -> LOGGER.debug("ZooKeeper topics watcher stopped"))
+                .onFailure(cause -> LOGGER.error("Failed to stop ZooKeeper topics watcher", cause))
+            .compose(ignored -> stopTopicStoreService())
```
tombentley, line 152: I think we only want to stop the topic store once the operator has worked through all existing in-flight events. The old code had that, but I don't see it here in the new code.

topic-operator/src/main/java/io/strimzi/operator/topic/Session.java
```diff
@@ -133,72 +136,102 @@ public void init(Vertx vertx, Context context) {
...
+            .compose(ignored -> closeKafkaAdminClient(shutdownExecutor))
+                .onSuccess(ignored -> LOGGER.debug("Kafka admin client closed"))
+                .onFailure(cause -> LOGGER.error("Failed to close Kafka admin client", cause))
+            .compose(ignored -> closeHttpHealthServer())
+                .onSuccess(ignored -> LOGGER.debug("HTTP health server closed"))
+                .onFailure(cause -> LOGGER.error("Failed to close HTTP health server", cause))
+            .compose(ignored -> {
+                try {
+                    shutdownExecutor.shutdown();
+                    shutdownExecutor.awaitTermination(10, TimeUnit.SECONDS);
```
tombentley, line 167: It feels a bit wrong that we're duplicating knowledge that timeout of the shutdown hook is 10s.

topic-operator/src/main/java/io/strimzi/operator/topic/Session.java
```diff
@@ -133,72 +136,102 @@ public void init(Vertx vertx, Context context) {
...
+                } catch (InterruptedException e) {
+                    if (!shutdownExecutor.isTerminated()) {
```
tombentley, line 169: It's worth adding a comment to explain why this behaviour for handling the InterruptedException is thought to be correct.

topic-operator/src/main/java/io/strimzi/operator/topic/Session.java
```diff
@@ -133,72 +136,102 @@ public void init(Vertx vertx, Context context) {
...
-                        HttpServer healthServer = this.healthServer;
-                        if (healthServer != null) {
-                            healthServer.close();
-                        }
-                    } catch (TimeoutException e) {
-                        LOGGER.warn("Timeout while closing AdminClient with timeout {}ms", timeoutMs, e);
-                    } finally {
-                        LOGGER.info("Stopped");
-                        blockingResult.complete();
+    private Future<Void> stopZooKeeperWatcher() {
```
tombentley, line 186: It's good that you've factored out these separate methods.

topic-operator/src/main/java/io/strimzi/operator/topic/Session.java
```diff
@@ -133,72 +136,102 @@ public void init(Vertx vertx, Context context) {
...
+        return Future.succeededFuture();
+    }
+
+    @SuppressFBWarnings("RV_RETURN_VALUE_IGNORED_BAD_PRACTICE")
+    private Future<Void> disconnectFromZk(ExecutorService shutdownExecutor) {
+        LOGGER.debug("Disconnecting from ZooKeeper");
+        Promise<Void> promise = Promise.promise();
+        if (zk != null) {
+            shutdownExecutor.submit(() -> {
+                zk.disconnect(result -> {
```
tombentley, line 208: We need to be very careful (and should document) on which thread the returned futures complete. In this case that depends on `ZkImpl`, and to be entirely honest I'm not quite sure which thread that will be, because we're calling disconnect from a thread that's not vertx-managed, so I don't know if it's able to recover the/a context thread.

topic-operator/src/main/java/io/strimzi/operator/topic/Session.java
```diff
@@ -133,72 +136,102 @@ public void init(Vertx vertx, Context context) {
...
                 });
-                return Future.succeededFuture();
             });
-        }, stop);
+        }
+        return promise.future();
+    }
+
+    @SuppressFBWarnings("RV_RETURN_VALUE_IGNORED_BAD_PRACTICE")
+    private Future<Void> closeKafkaAdminClient(ExecutorService shutdownExecutor) {
```
tombentley, line 221: This is very deceptive. This method signature suggests that the close is asynchronous (which it is), but by returning a `Future` we're also implying that that future will be _completed when the `close()` is finished_, and that's *not* the case.

This should either return `void`, or the returned `Future` should complete (on a defined thread) once the close has completed.

topic-operator/src/main/java/io/strimzi/operator/topic/Session.java
fvaleri: Factoring out to shared Config.

fvaleri:
@tombentley thanks for the great suggestions. Let me know if anything more is needed.

New shutdown procedure in action:

```sh
2023-11-21 14:53:49,31295 INFO  [Thread-0] ShutdownHook:35 - Shutdown hook started
2023-11-21 14:53:49,31343 INFO  [Thread-0] ShutdownHook:60 - Shutting down Vert.x
2023-11-21 14:53:49,31670 INFO  [vert.x-eventloop-thread-0] Session:138 - Stopping
2023-11-21 14:53:49,31755 DEBUG [vert.x-eventloop-thread-0] Session:168 - Closing Kubernetes watch
2023-11-21 14:53:49,31888 DEBUG [vert.x-worker-thread-1] AbstractWatchManager:323 - Force closing the watch
2023-11-21 14:53:49,31965 DEBUG [-998242313-pool-3-thread-1] Watcher:44 - Watcher closed
2023-11-21 14:53:49,31993 DEBUG [vert.x-worker-thread-1] AbstractWatchManager:160 - Closing the current watch
2023-11-21 14:53:49,32333 DEBUG [vert.x-worker-thread-1] Session:144 - Kubernetes watch closed
2023-11-21 14:53:49,32338 DEBUG [JdkHttpClientFactory-166694583-pool-2-thread-1] WatcherWebSocketListener:65 - WebSocket close received. code: 1000, reason: 
2023-11-21 14:53:49,32350 DEBUG [vert.x-worker-thread-1] Session:186 - Stopping ZooKeeper watcher
2023-11-21 14:53:49,32359 DEBUG [JdkHttpClientFactory-166694583-pool-2-thread-1] AbstractWatchManager:219 - Ignoring already closed/closing connection
2023-11-21 14:53:49,32406 DEBUG [vert.x-worker-thread-1] Session:147 - ZooKeeper topics watcher stopped
2023-11-21 14:53:49,32418 DEBUG [vert.x-worker-thread-1] Session:204 - Stopping topic store service
2023-11-21 14:53:49,32431 DEBUG [vert.x-worker-thread-1] TopicOperator:1390 - Outstanding: {}
2023-11-21 14:53:49,32443 INFO  [vert.x-worker-thread-1] KafkaStreamsTopicStoreService:161 - Stopping services ...
2023-11-21 14:53:49,37438 INFO  [__strimzi-topic-operator-kstreams-4691c8c5-abc5-4b82-984b-ea11325d4d21-StreamThread-1] Metrics:693 - Metrics scheduler closed
2023-11-21 14:53:49,37461 INFO  [__strimzi-topic-operator-kstreams-4691c8c5-abc5-4b82-984b-ea11325d4d21-StreamThread-1] Metrics:697 - Closing reporter org.apache.kafka.common.metrics.JmxReporter
2023-11-21 14:53:49,37477 INFO  [__strimzi-topic-operator-kstreams-4691c8c5-abc5-4b82-984b-ea11325d4d21-StreamThread-1] Metrics:703 - Metrics reporters closed
2023-11-21 14:53:49,37512 INFO  [__strimzi-topic-operator-kstreams-4691c8c5-abc5-4b82-984b-ea11325d4d21-StreamThread-1] AppInfoParser:83 - App info kafka.producer for __strimzi-topic-operator-kstreams-4691c8c5-abc5-4b82-984b-ea11325d4d21-StreamThread-1-producer unregistered
2023-11-21 14:53:49,76191 INFO  [__strimzi-topic-operator-kstreams-4691c8c5-abc5-4b82-984b-ea11325d4d21-StreamThread-1] Metrics:693 - Metrics scheduler closed
2023-11-21 14:53:49,76207 INFO  [__strimzi-topic-operator-kstreams-4691c8c5-abc5-4b82-984b-ea11325d4d21-StreamThread-1] Metrics:697 - Closing reporter org.apache.kafka.common.metrics.JmxReporter
2023-11-21 14:53:49,76223 INFO  [__strimzi-topic-operator-kstreams-4691c8c5-abc5-4b82-984b-ea11325d4d21-StreamThread-1] Metrics:703 - Metrics reporters closed
2023-11-21 14:53:49,76519 INFO  [__strimzi-topic-operator-kstreams-4691c8c5-abc5-4b82-984b-ea11325d4d21-StreamThread-1] AppInfoParser:83 - App info kafka.consumer for __strimzi-topic-operator-kstreams-4691c8c5-abc5-4b82-984b-ea11325d4d21-StreamThread-1-consumer unregistered
2023-11-21 14:53:49,76714 INFO  [__strimzi-topic-operator-kstreams-4691c8c5-abc5-4b82-984b-ea11325d4d21-StreamThread-1] Metrics:693 - Metrics scheduler closed
2023-11-21 14:53:49,76728 INFO  [__strimzi-topic-operator-kstreams-4691c8c5-abc5-4b82-984b-ea11325d4d21-StreamThread-1] Metrics:697 - Closing reporter org.apache.kafka.common.metrics.JmxReporter
2023-11-21 14:53:49,76742 INFO  [__strimzi-topic-operator-kstreams-4691c8c5-abc5-4b82-984b-ea11325d4d21-StreamThread-1] Metrics:703 - Metrics reporters closed
2023-11-21 14:53:49,76968 INFO  [__strimzi-topic-operator-kstreams-4691c8c5-abc5-4b82-984b-ea11325d4d21-StreamThread-1] AppInfoParser:83 - App info kafka.consumer for __strimzi-topic-operator-kstreams-4691c8c5-abc5-4b82-984b-ea11325d4d21-StreamThread-1-restore-consumer unregistered
2023-11-21 14:53:49,77077 INFO  [kafka-admin-client-thread | __strimzi-topic-operator-kstreams-4691c8c5-abc5-4b82-984b-ea11325d4d21-admin] AppInfoParser:83 - App info kafka.admin.client for __strimzi-topic-operator-kstreams-4691c8c5-abc5-4b82-984b-ea11325d4d21-admin unregistered
2023-11-21 14:53:49,77243 INFO  [kafka-admin-client-thread | __strimzi-topic-operator-kstreams-4691c8c5-abc5-4b82-984b-ea11325d4d21-admin] Metrics:693 - Metrics scheduler closed
2023-11-21 14:53:49,77261 INFO  [kafka-admin-client-thread | __strimzi-topic-operator-kstreams-4691c8c5-abc5-4b82-984b-ea11325d4d21-admin] Metrics:697 - Closing reporter org.apache.kafka.common.metrics.JmxReporter
2023-11-21 14:53:49,77275 INFO  [kafka-admin-client-thread | __strimzi-topic-operator-kstreams-4691c8c5-abc5-4b82-984b-ea11325d4d21-admin] Metrics:703 - Metrics reporters closed
2023-11-21 14:53:49,77309 INFO  [__strimzi-topic-operator-kstreams-4691c8c5-abc5-4b82-984b-ea11325d4d21-CloseThread] Metrics:693 - Metrics scheduler closed
2023-11-21 14:53:49,77324 INFO  [__strimzi-topic-operator-kstreams-4691c8c5-abc5-4b82-984b-ea11325d4d21-CloseThread] Metrics:697 - Closing reporter org.apache.kafka.common.metrics.JmxReporter
2023-11-21 14:53:49,77333 INFO  [__strimzi-topic-operator-kstreams-4691c8c5-abc5-4b82-984b-ea11325d4d21-CloseThread] Metrics:703 - Metrics reporters closed
2023-11-21 14:53:49,77365 DEBUG [vert.x-worker-thread-1] Session:150 - Topic store service stopped
2023-11-21 14:53:49,77378 DEBUG [vert.x-worker-thread-1] Session:221 - Disconnecting from ZooKeeper
2023-11-21 14:53:49,77390 DEBUG [vert.x-worker-thread-1] Session:153 - ZooKeeper disconnected
2023-11-21 14:53:49,77399 DEBUG [vert.x-worker-thread-1] Session:243 - Closing Kafka admin client
2023-11-21 14:53:49,77549 INFO  [kafka-admin-client-thread | adminclient-1] AppInfoParser:83 - App info kafka.admin.client for adminclient-1 unregistered
2023-11-21 14:53:49,77952 INFO  [kafka-admin-client-thread | adminclient-1] Metrics:693 - Metrics scheduler closed
2023-11-21 14:53:49,77996 INFO  [kafka-admin-client-thread | adminclient-1] Metrics:697 - Closing reporter org.apache.kafka.common.metrics.JmxReporter
2023-11-21 14:53:49,78019 INFO  [kafka-admin-client-thread | adminclient-1] Metrics:703 - Metrics reporters closed
2023-11-21 14:53:49,78053 DEBUG [vert.x-worker-thread-1] Session:156 - Kafka admin client closed
2023-11-21 14:53:49,78070 DEBUG [vert.x-worker-thread-1] Session:261 - Closing HTTP health server
2023-11-21 14:53:49,78473 DEBUG [vert.x-worker-thread-1] Session:159 - HTTP health server closed
2023-11-21 14:53:49,78508 INFO  [vert.x-worker-thread-1] Session:162 - Stopped
2023-11-21 14:53:49,79294 DEBUG [vert.x-eventloop-thread-0] PoolThreadCache:221 - Freed 2 thread-local buffer(s) from thread: vert.x-eventloop-thread-0
2023-11-21 14:53:49,79799 INFO  [Thread-0] ShutdownHook:79 - Shutdown of Vert.x is complete
2023-11-21 14:53:49,79858 INFO  [Thread-0] ShutdownHook:41 - Shutdown hook completed
```

fvaleri:
@tombentley do you have any other comment?

topic-operator/src/test/java/io/strimzi/operator/topic/TopicOperatorBaseIT.java
```diff
@@ -507,7 +507,7 @@ protected void createAndAlterNumPartitions(String topicName) throws InterruptedE
     protected void waitFor(BooleanSupplier ready, String message) throws TimeoutException, InterruptedException {
         // Note that this timeout for an individual wait must be less than
         // the Vertx @Timeout for the test as a whole
-        long timeout = 120_000;
+        long timeout = 120_000;
```
scholzj, line 510: I guess no need for the empty space here?

scholzj (commented):
Just one nit about unnecessary change in one of the files. LGTM otherwise assuming it works 😉

scholzj (approved):

topic-operator/src/main/java/io/strimzi/operator/topic/Session.java
```diff
@@ -133,72 +134,136 @@ public void init(Vertx vertx, Context context) {
...
-                        LOGGER.debug("Inflight work has finished");
-                        promise.complete();
-                    } else if (System.currentTimeMillis() > deadline) {
-                        LOGGER.error("Timeout waiting for inflight work to finish");
-                        promise.complete();
-                    } else {
-                        LOGGER.debug("Waiting for inflight work to finish");
-                        vertx.setTimer(1_000, this);
-                    }
+        closeKubernetesWatch()
```
katheris, line 143: My understanding of how the vertx futures work is that this future failing will result in all the subsequent `compose` methods not being called and every `onFailure` method being invoked. Is that the intended behaviour here?

fvaleri: Yeah, this is also my understanding. Here we want to stop the internal components in a specific order and that's why we use compose. Each method of this chain completes successfully, regardless of the operation outcome. One could ask why we log failures then. I though that they may capture some framework level error. Does it make sense?

katheris: Doesn't `disconnectFromZk` have the possibility to fail?

katheris: I guess the other question is what we want to happen if a framework level error happens, in that case is it ok to skip stopping the remaining components? Or would we expect to attempt to stop the other components even if an earlier one fails?

fvaleri: > Doesn't disconnectFromZk have the possibility to fail?

Right. That needs to be changed. Thanks.

> Would we expect to attempt to stop the other components even if an earlier one fails

I would say no. All external systems should be able to cope with unclean disconnections due to application failures or network issues, so I see this as a best effort attempt to orderly stop all components before application shutdown, with a configurable hard timeout (10 seconds by default).

fvaleri:
@tombentley @katheris are we good with these changes?

scholzj:
Discussed on the community call on 30.11.2023: @katheris @tombentley Could you please have another look? Thanks

katheris (approved):

topic-operator/src/main/java/io/strimzi/operator/topic/Session.java
katheris: Thanks, that all makes sense to me

PR #9364, opened 2023-11-19
Add support for managing KRaft metadata version

scholzj:
### Type of change

- Enhancement / new feature

### Description

This PR implements the first part of the [SP#61](https://github.com/strimzi/proposals/blob/main/061-kraft-upgrades-and-downgrades.md). It adds support for managing KRaft metadata versions:
* Adds a new field to the Kafka CR to specify the desired metadata version
* Adds the possibility to define the initial metadata version when deploying new clusters
* Adds the possibility to change the metadata versions (to the extent in which it is supported by Kafka itself)

It does not deliver the actual KRaft upgrades yet not any documentation -> that will be done in follow up PR in order to allow for less disruption and faster reviews.

It also adds the initial metadata version to the KRaft example YAMLs. It uses the full format including the subversion. That is important given the significant limitation KRaft currently has when it comes to downgrading the metadata version.

### Checklist

- [x] Write tests
- [x] Update CHANGELOG.md

### DIFF
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaCluster.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaCluster.java
@@ -283,7 +283,7 @@ public class KafkaCluster extends AbstractModel implements SupportsMetrics, Supp
         // This also validates that the Kafka version is supported
         result.kafkaVersion = versions.supportedVersion(kafkaClusterSpec.getVersion());
 
-        // Sets the LRaft metadata version and validates it is supported
+        // Sets the KRaft metadata version and validates it is supported
         String metadataVersion = kafkaClusterSpec.getMetadataVersion() != null ? kafkaClusterSpec.getMetadataVersion() : result.kafkaVersion.metadataVersion();
         KRaftUtils.validateMetadataVersion(metadataVersion);
         result.metadataVersion = metadataVersion;
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaCluster.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaCluster.java
@@ -123,7 +123,6 @@ public class KafkaCluster extends AbstractModel implements SupportsMetrics, Supp
     protected static final int CONTROLPLANE_PORT = 9090;
     protected static final String CONTROLPLANE_PORT_NAME = "tcp-ctrlplane"; // port name is up to 15 characters
 
-
     /**
      * Port used by the Route listeners
      */
@@ -199,12 +198,18 @@ public class KafkaCluster extends AbstractModel implements SupportsMetrics, Supp
      */
     public static final String BROKER_CLUSTER_ID_FILENAME = "cluster.id";
 
+    /**
+     * Key under which the desired Kafka metadata version is stored in Config Map
+     */
+    public static final String BROKER_METADATA_VERSION_FILENAME = "metadata.version";
+
     // Kafka configuration
     private Rack rack;
     private String initImage;
     private List<GenericKafkaListener> listeners;
     private KafkaAuthorization authorization;
     private KafkaVersion kafkaVersion;
+    private String metadataVersion;
     private boolean useKRaft = false;
     private String clusterId;
     private JmxModel jmx;
@@ -277,6 +282,11 @@ public class KafkaCluster extends AbstractModel implements SupportsMetrics, Supp
 
         // This also validates that the Kafka version is supported
         result.kafkaVersion = versions.supportedVersion(kafkaClusterSpec.getVersion());
+
+        // Sets the LRaft metadata version and validates it is supported
+        String metadataVersion = kafkaClusterSpec.getMetadataVersion() != null ? kafkaClusterSpec.getMetadataVersion() : result.kafkaVersion.metadataVersion();
+        KRaftUtils.validateMetadataVersion(metadataVersion);
+        result.metadataVersion = metadataVersion;
 
         // Number of broker nodes => used later in various validation methods
         long numberOfBrokers = result.brokerNodes().size();
@@ -1690,8 +1700,9 @@ public class KafkaCluster extends AbstractModel implements SupportsMetrics, Supp
                 data.put(BROKER_LISTENERS_FILENAME, node.broker() ? listeners.stream().map(ListenersUtils::envVarIdentifier).collect(Collectors.joining(" ")) : null);
 
                 if (useKRaft) {
-                    // In KRaft, we need to pass the Kafka CLuster ID
+                    // In KRaft, we need to pass the Kafka CLuster ID and the metadata version
                     data.put(BROKER_CLUSTER_ID_FILENAME, clusterId);
+                    data.put(BROKER_METADATA_VERSION_FILENAME, metadataVersion);
                 }
 
                 configMaps.add(ConfigMapUtils.createConfigMap(node.podName(), namespace, pool.labels.withStrimziPodName(node.podName()), pool.ownerReference, data));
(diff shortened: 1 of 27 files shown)


cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaCluster.java
```diff
@@ -278,6 +283,11 @@ public static KafkaCluster fromCrd(Reconciliation reconciliation, Kafka kafka, L
         // This also validates that the Kafka version is supported
         result.kafkaVersion = versions.supportedVersion(kafkaClusterSpec.getVersion());
 
+        // Sets the LRaft metadata version and validates it is supported
```
ppatierno, line 286:
```suggestion
        // Sets the KRaft metadata version and validates it is supported
```

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaCluster.java
```diff
@@ -1668,8 +1678,9 @@ public List<ConfigMap> generatePerBrokerConfigurationConfigMaps(MetricsAndLoggin
                 data.put(BROKER_LISTENERS_FILENAME, listeners.stream().map(ListenersUtils::envVarIdentifier).collect(Collectors.joining(" ")));
 
                 if (useKRaft) {
-                    // In KRaft, we need to pass the Kafka CLuster ID
+                    // In KRaft, we need to pass the Kafka CLuster ID and the metadata version
                     data.put(BROKER_CLUSTER_ID_FILENAME, clusterId);
+                    data.put(BROKER_METADATA_VERSION_FILENAME, metadataVersion);
```
ppatierno, line 1683: wondering what's going to happen when, in my migration PR, we are going to set this configuration field on a broker which is migrating so it has both ZooKeeper and KRaft configured. I mean ... if the Kafka code will complain.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KRaftMetadataManager.java
```diff
@@ -0,0 +1,162 @@
...
+import io.vertx.core.Vertx;
+import org.apache.kafka.clients.admin.Admin;
+import org.apache.kafka.clients.admin.FeatureUpdate;
+import org.apache.kafka.clients.admin.UpdateFeaturesOptions;
+import org.apache.kafka.server.common.MetadataVersion;
+
+import java.util.Map;
+
+/**
+ * Utility class for managing KRaft the metadata version. It encapsulates the methods for getting the current metadata
```
ppatierno, line 27:
```suggestion
 * Utility class for managing the KRaft metadata version. It encapsulates the methods for getting the current metadata
```

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KRaftMetadataManager.java
```diff
@@ -0,0 +1,162 @@
...
+ */
+public class KRaftMetadataManager {
+    private static final ReconciliationLogger LOGGER = ReconciliationLogger.create(KRaftMetadataManager.class.getName());
+    /* test */ static final String METADATA_VERSION_KEY = MetadataVersion.FEATURE_NAME;
+
+    /**
+     * Method for managing the KRaft metadata version. This method will get the current metadata version and if needed,
+     * upgrade or downgrade it. It will also update the currently used metadata version in the Kafka CR status. This
+     * method might fail when it fails to communicate with the Kafka cluster and get the metadata version. But it will
+     * not fail when the update of the metadata version fails in order to not fail the whole reconciliation. Such event
```
ppatierno, line 38: I think I have got what you are explaining but the usage of verb "fail" so many times makes it harder to be readable.

docker-images/kafka-based/kafka/scripts/kafka_run.sh
```diff
@@ -61,11 +61,12 @@ if [ "$STRIMZI_KRAFT_ENABLED" = "true" ]; then
...
   if [ ! -f "$KRAFT_LOG_DIR/meta.properties" ]; then
     STRIMZI_CLUSTER_ID=$(cat "$KAFKA_HOME/custom-config/cluster.id")
-    echo "Formatting Kraft storage with cluster ID $STRIMZI_CLUSTER_ID"
+    METADATA_VERSION=$(cat "$KAFKA_HOME/custom-config/metadata.version")
+    echo "Formatting Kraft storage with cluster ID $STRIMZI_CLUSTER_ID and metadata version $METADATA_VERSION"
     mkdir -p "$KRAFT_LOG_DIR"
     # Using "=" to assign arguments for the Kafka storage tool to avoid issues if the generated
     # cluster ID starts with a "-". See https://issues.apache.org/jira/browse/KAFKA-15754
-    ./bin/kafka-storage.sh format -t="$STRIMZI_CLUSTER_ID" -c=/tmp/strimzi.properties
+    ./bin/kafka-storage.sh format -t="$STRIMZI_CLUSTER_ID" -r="$METADATA_VERSION" -c=/tmp/strimzi.properties
```
ppatierno, line 69: When doing a migration from ZooKeeper to KRaft, the formatting doesn't happen because the cluster id is already assigned to the broker (from its previous life as ZooKeeper-based). What does it mean in terms of metadata version then? I mean, the cluster is migrated to KRaft but the storage won't be formatted passing the metadata version.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KRaftMetadataManager.java
scholzj: I tried to reword it. Please let me know if it feels a bit more clear.

docker-images/kafka-based/kafka/scripts/kafka_run.sh
scholzj: I guess you would need to check it next time you are on a migrated cluster. My guess is that one of the following happens:
* It will be migrated to the default metadata version (i.e. Kafka 3.6.0 to 3.6-IV2) => this is what happens if you don't pass the `-r` method to the `kafka-storage.sh` script.
* It will use whatever is in the `inter.broker.protocol.version` => because this is essentially replacement for it.

I do not expect this to cause a major issue. We should just explain in the docs to the users what should they set as `.spec.kafka.metadataVersion`.

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaCluster.java
scholzj: Please see the reply to the comment about the `kafka_run.sh`.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KRaftMetadataManager.java
ppatierno: Yeah, it looks great now thanks!

scholzj: @ppatierno So, can you approve the PR? 😉

docker-images/kafka-based/kafka/scripts/kafka_run.sh
ppatierno: I had a quick look at the impact of this on the migration side. I see that at the end of the migration, the metadata version is set with the default value you mentioned, and it seems to work okay.
Of course, when creating the controllers in my migration PR, the `-r="$METADATA_VERSION"` is still missing. We'll know more when this PR is merged and I can rebase.

ppatierno (approved):

PR #9369, opened 2023-11-20
Remove assumption that is not used anymore

### LINKED ISSUE #8806: [systemtest] Remove assumption in tests when Kafka version is lower than 3.5.0 and KRaft is enabled
Currently, in few of our STs where we are using KafkaUsers, we are skipping the test when KRaft is enabled and Kafka version is lower than 3.5.0.
This is because the SCRAM-SHA feature added in 3.5.0, which is not available in older versions of Kafka we are currently supporting.

Once Kafka versions 3.4.x will be removed, we should also remove these assumptions.

The assumptions are commented with this issue.

scholzj:
### Type of change

- Task

### Description

This PR removes the assumption from some system tests that they should run only on Kafka 3.5 or higher. This assumption is not needed anymore as we do not support any older KAfka versions than 3.5.0.

This should close #8806

### Checklist


im-konge (commented):
Thanks for the PR, I guess there is one more in `FeatureGatesST#testKRaftMode` if you could check it, thanks!

PR #9374, opened 2023-11-21
Reduce number of situations when KRaft controller-only nodes need rolling updates

scholzj:
### Type of change

- Bugfix
- Enhancement / new feature

### Description

In KRaft clusters with dedicated controller nodes, we want to reduce the number of rolling updates of the controller nodes as much as possible. One of the ways to achieve that is to reduce the number of configs pass to them to the minimum needed.

This PR implements some changes to reduce the rolling updates:
* Controller-only certificates should not have SANs for external listeners as any external clients don't connect to them.
* Controller-only init containers do not need to pull the node address for node port listeners.
* Controller-only nodes do not need the list of user-defined listeners (we still pass an empty file in the configuration ConfigMap to avoid complexity in the script generating the Kafka node configuration).
* Controller-only nodes do not need to have open ports for all the various user-defined listeners or for the replication listener. They need only the control plane listener. This helps to avoid rolling updates for changed listener configurations by our users.

These changes should help to reduce the amount of rolls. They should not impact Zoo-based clusters or mixed broker or controller nodes (or broker-only nodes).

### Checklist

- [x] Write tests

### DIFF
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaCluster.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaCluster.java
@@ -1114,19 +1114,37 @@ public class KafkaCluster extends AbstractModel implements SupportsMetrics, Supp
                 emptyMap());
     }
 
-    /* test */ List<ContainerPort> getContainerPortList() {
+    /**
+     * Node Pool for which the ports should be generated. In KRaft, the ports for controller-only nodes might differ
+     * from broker nodes as they do not need all the listeners for clients or for replication.
+     *
+     * @param pool  Pool for which the ports should be generated
+     *
+     * @return  List of container ports
+     */
+    /* test */ List<ContainerPort> getContainerPortList(KafkaPool pool) {
         List<ContainerPort> ports = new ArrayList<>(listeners.size() + 3);
-        ports.add(ContainerUtils.createContainerPort(CONTROLPLANE_PORT_NAME, CONTROLPLANE_PORT));
-        ports.add(ContainerUtils.createContainerPort(REPLICATION_PORT_NAME, REPLICATION_PORT));
 
-        for (GenericKafkaListener listener : listeners) {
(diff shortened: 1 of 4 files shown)


cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaCluster.java
```diff
@@ -1114,19 +1114,37 @@ public Secret generateCertificatesSecret(ClusterCa clusterCa, ClientsCa clientsC
...
+     */
+    /* test */ List<ContainerPort> getContainerPortList(KafkaPool pool) {
         List<ContainerPort> ports = new ArrayList<>(listeners.size() + 3);
-        ports.add(ContainerUtils.createContainerPort(CONTROLPLANE_PORT_NAME, CONTROLPLANE_PORT));
-        ports.add(ContainerUtils.createContainerPort(REPLICATION_PORT_NAME, REPLICATION_PORT));
 
-        for (GenericKafkaListener listener : listeners) {
-            ports.add(ContainerUtils.createContainerPort(ListenersUtils.backwardsCompatiblePortName(listener), listener.getPort()));
+        if (!useKRaft || pool.isController()) {
+            // The control plane listener is on all nodes in ZooKeeper based clusters and on nodes with controller role in KRaft
```
ppatierno, line 1129: just to understand, is it excluding brokers in KRaft? Just to think about how the `!useKRaft` condition will be reflected during a migration where brokers have both KRaft and ZooKeeper configured.

ppatierno (approved):
LGTM. I left a question.

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaCluster.java
ppatierno: I fixed this with commit https://github.com/strimzi/strimzi-kafka-operator/pull/9324/commits/392d2ef548012cb8181e67eda38cabfe8e1b3cc9 as part of my migration PR.

PR #9401, opened 2023-11-27
Added comment about Control Plane listeners on KRaft broker-only nodes

ppatierno:
This PR adds more comments to explain why we still need to configure the CONTROL_PLANE listener on KRaft broker-only nodes when it's actually not used as listener because the "control plane" is made by KRaft controllers out of the brokers.

### DIFF
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaBrokerConfigurationBuilder.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaBrokerConfigurationBuilder.java
@@ -267,7 +267,12 @@ public class KafkaBrokerConfigurationBuilder {
         }
 
         // Security protocol and Control Plane Listener are configured everywhere
+
+        // Control Plane listener is added to security protocol on KRaft broker only nodes as well because validated by Kafka
+        // This is because it will be in the controller.listener.names to allow connecting to controllers
         securityProtocol.add(CONTROL_PLANE_LISTENER_NAME + ":SSL");
+        // Control Plane listener is configured on KRaft broker only nodes as well for allowing TLS certificates keystore generation
+        // so that brokers are able to connect to controllers as TLS clients
         configureControlPlaneListener();
 
         // Non-controller listeners are used only on ZooKeeper based brokers or KRaft brokers (including mixed nodes)
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaBrokerConfigurationBuilder.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaBrokerConfigurationBuilder.java
@@ -268,8 +268,10 @@ public class KafkaBrokerConfigurationBuilder {
 
         // Security protocol and Control Plane Listener are configured everywhere
 
-        // Control Plane listener is added to security protocol on KRaft broker only nodes as well because validated by Kafka
-        // This is because it will be in the controller.listener.names to allow connecting to controllers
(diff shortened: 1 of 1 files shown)


cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaBrokerConfigurationBuilder.java
```diff
@@ -267,7 +267,12 @@ public KafkaBrokerConfigurationBuilder withListeners(
         }
 
         // Security protocol and Control Plane Listener are configured everywhere
+
+        // Control Plane listener is added to security protocol on KRaft broker only nodes as well because validated by Kafka
```
katheris, line 271: This sentence doesn't male sense to me, what do you mean "because validated by Kafka"?

ppatierno: I meant "because the Kafka codebase validates that the control plane listener is in the security protocol list".
So "validated by kafka" is referring to the first part of the sentence to explain why the control plane listener has to be in the security protocol list. Any better suggestion?

fvaleri: this information is used by the controller to know which security protocol to use when connecting to the broker's control plane endpoint

ppatierno: so @fvaleri @katheris any suggestions about how to make this sentence clearer?

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaBrokerConfigurationBuilder.java
```diff
@@ -267,7 +267,12 @@ public KafkaBrokerConfigurationBuilder withListeners(
...
+        // This is because it will be in the controller.listener.names to allow connecting to controllers
         securityProtocol.add(CONTROL_PLANE_LISTENER_NAME + ":SSL");
+        // Control Plane listener is configured on KRaft broker only nodes as well for allowing TLS certificates keystore generation
+        // so that brokers are able to connect to controllers as TLS clients
```
scholzj, line 275: Maybe this would be a bit more clear?

```suggestion
        // Brokers need to know how to connect to the controllers on the Control Plane listener and what security (encryption/authentication) they should use.
        // For that reason, we have to configure the Control Plane listener in the broker-only configuration as well, even though they do not listen at the Control Plane listener port.
        // The brokers use this configuration to detect how to connect to the controllers, what certificates to use etc.
        securityProtocol.add(CONTROL_PLANE_LISTENER_NAME + ":SSL");
```

ppatierno: Just pushed this change, @fvaleri @katheris any more comments?

fvaleri: Fine for me, thanks.

fvaleri (approved):

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaBrokerConfigurationBuilder.java
katheris: Yeah that looks good to me

katheris (approved):

Issue #8658, opened 2023-06-12
Make model classes immutable


scholzj:
Currently, the model classes (such as `KafkaCluster`, `KafkaConnectCluster` etc.) contain a small subset of `final` fields which are set through the constructor. Most of the fields are set from the static `fromCrd` method based on the configuration from the custom resource.

Most of the fields set in the `fromCrd` methods are not changed later. But the fields in these classes cannot be set as final because they are not set in the constructor. So they are mutable although they should really be immutable. We might be able to address this by moving the logic from the `fromCrd` method to the constructor and making these fields final. We can possibly also make these classes just `records`.

It should be investigated whether this is really feasible and whether the number of fields in the constructor is acceptable or not (and whether it possibly should be broken into more records etc.).

This was raised after a discussion in https://github.com/strimzi/strimzi-kafka-operator/pull/8551#discussion_r1226630685

tombentley:
Triaged 15/06/2023: This seems like a good idea. We should try it and see if it will work in practice. It might be possible to do incrementally (i.e. start with one of the simpler APIs before trying to bite off the Kafka model).

scholzj:
While this sounds good on paper, the number of fields in the constructors will be huge. For example:
* `AbstractModel`: 16 final fields. And likely 1 constructor parameter for each of them.
* `KafkaCluster`: 21 final fields (this does not include the 16 from the parent `AbstractModel).
* `CruiseControl`: 10 final fields (+16 from `AbstractModel`)
* `KafkaExporter`: 9 final fields (+16 from `AbstractModel`)
* `KafkaPool`: 10 final fields (+16 from `AbstractModel`)

The number of final fields alone is not the main concern. But the number of the constructor parameters is. One of the main issues is that the `super(...)` has to be called first. That will require the 16 parameters mentioned above. They will have to be prepared / parsed out from the resource before calling the `super(...)`. But because you have to call the `super(...)` as a first thing in the constructor, you cannot do it in the constructor itself. So you would need to do the preparation in the `fromCrd` method and then pass the parameters to the constructor where they will be just set. That likely means that every constructor will have 20-30 (or in some cases even more) parameters. So while I agree the immutable classes are a good idea, I'm not sure this is ready to be done right now by just making it immutable. Maybe it needs some more refactoring to change how these things work to minimize the number of parameters?

WDYT @tombentley?

PR #9410, opened 2023-11-29
Refactor `KafkaVersionChange` class and how it is used in `KafkaRecocniler` and `KafkaCluster`

scholzj:
### Type of change

- Refactoring

### Description

This PR refactors the `KafkaVErsionChange` class and how it is used in the reconcilers and model classes.

As a first change, it changes the KafkaVersionChange from a class to a record. As part of it, it also removes several methods that were used mostly only in tests + in a single place in the actual production code and seem to be unnecessary (e.g. the `isUpgrade()` etc. methods).

It also changes how `KafkaVersionChange` is used in the `KafkaReconciler`. Right now, we extract the version information inside the `KafkaReconciler` constructor and pass it into `KafkaCluster` using setters. This PR moves this logic into the `KafkaCluster.fromCrd(...)` method and removes the setters. That seems to better encapsulate how the version is set and should move us in the direction to achieve https://github.com/strimzi/strimzi-kafka-operator/issues/8658 in the future.

It also changes the `ZooKeeperVersionChangeCreator` class to set the KRaft metadata version as well as it is needed for the ZooKeeper to KRaft migration.

### Checklist

- [x] Write tests

### DIFF
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/ZooKeeperVersionChangeCreator.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/ZooKeeperVersionChangeCreator.java
@@ -366,13 +366,6 @@ public class ZooKeeperVersionChangeCreator implements VersionChangeCreator {
         // For that, we need to set the metadata version even in the ZooKeeperVersionChangeCreator class even through it
         // is not used in KRaft mode. As the controllers will be new, we set it based on the Kafka CR or based on the
         // Kafka version used (which will be always the versionTo on the newly deployed controllers).
-        String metadataVersion;
-        if (metadataVersionFromCr != null)  {
-            metadataVersion = metadataVersionFromCr;
-        } else {
-            metadataVersion = versionTo.metadataVersion();
-        }
-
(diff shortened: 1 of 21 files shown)


cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/ZooKeeperVersionChangeCreator.java
```diff
@@ -360,6 +362,17 @@ && compareDottedIVVersions(interBrokerProtocolVersionFromCr, "3.0") >= 0) {
...
+        // For migration from ZooKeeper to KRaft, we need to configure the metadata version on the new controller nodes.
+        // For that, we need to set the metadata version even in the ZooKeeperVersionChangeCreator class even through it
+        // is not used in KRaft mode. As the controllers will be new, we set it based on the Kafka CR or based on the
+        // Kafka version used (which will be always the versionTo on the newly deployed controllers).
+        String metadataVersion;
+        if (metadataVersionFromCr != null)  {
+            metadataVersion = metadataVersionFromCr;
+        } else {
+            metadataVersion = versionTo.metadataVersion();
+        }
```
ppatierno, line 374: no need to change but what about just one line:

```java
String metadataVersion = metadataVersionFromCr != null ? metadataVersionFromCr : versionTo.metadataVersion()
```

ppatierno (approved):
LGTM. Just a nit but no need to change.

PR #9418, opened 2023-11-30
Fix handling of connector state for Mirror Maker 2 connectors

scholzj:
### Type of change

- Bugfix

### Description

In Strimzi 0.38 we added support for stopping connectors. This includes also MM2. But it does not seem to work in MM2 because the state is not passed to the connector operator. This PR tries to fix it. 

It seems there are no tests covering much of the MM2 reconciliation and the code is not very easy to test. This PR aims to fix the main issue. I opened a separate issue to improve the test coverage and possible refactoring of the `KafkaMirrorMaker2AssemblyOperator` class: #9419

### Checklist


mimaison (approved):
LGTM, thanks Jakub for spotting and fixing this issue!

PR #9420, opened 2023-12-01
Add warning about incompleted upgrade when metadata version is not updated

scholzj:
### Type of change

- Enhancement / new feature

### Description

This PR adds some additional KRaft-related warnings to the Kafka spec checker tool that creates warnings for the status conditions. It adds:
* Warnings when the Kafka cluster is using an older metadata version than what would correspond to the Kafka version. This is similar to the warnings we have today for `inter.broker.protocol.version` and `log.message.format.version` in ZooKeeper-based clusters.
* Adds warnings when `inter.broker.protocol.version` and `log.message.format.version` are configured in a KRaft-based cluster as they are not relevant for KRaft anymore.

### Checklist

- [x] Write tests

### DIFF
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaSpecChecker.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaSpecChecker.java
@@ -55,15 +55,20 @@ public class KafkaSpecChecker {
     List<Condition> run(boolean useKRaft) {
         List<Condition> warnings = new ArrayList<>();
 
-        checkKafkaLogMessageFormatVersion(warnings);
-        checkKafkaInterBrokerProtocolVersion(warnings);
         checkKafkaReplicationConfig(warnings);
         checkKafkaBrokersStorage(warnings);
 
-        // Additional checks done for KRaft clusters
         if (useKRaft)   {
+            // Additional checks done for KRaft clusters
             checkKRaftControllerStorage(warnings);
             checkKRaftControllerCount(warnings);
+            checkKafkaMetadataVersion(warnings);
+            checkInterBrokerProtocolVersionInKRaft(warnings);
+            checkLogMessageFormatVersionInKRaft(warnings);
+        } else {
+            // Additional checks done for ZooKeeper-based clusters
+            checkKafkaLogMessageFormatVersion(warnings);
+            checkKafkaInterBrokerProtocolVersion(warnings);
         }
 
         return warnings;
@@ -180,6 +185,55 @@ public class KafkaSpecChecker {
         } else if (controllerCount % 2 == 0) {
             warnings.add(StatusUtils.buildWarningCondition("KafkaKRaftControllerNodeCount",
                     "Running KRaft controller quorum with an odd number of nodes is recommended."));
+        }
+    }
+
+    /**
+     * Checks if the version of the Kafka brokers matches any custom metadata version config. Updating this is the final
+     * step in upgrading Kafka version in KRaft, so if this doesn't match it is possibly an indication that a user has
+     * updated their Kafka cluster and is unaware that they also should update their metadata version to match.
+     *
+     * @param warnings List to add a warning to, if appropriate.
+     */
+    private void checkKafkaMetadataVersion(List<Condition> warnings) {
+        String metadataVersion = kafkaCluster.getMetadataVersion();
+
+        if (metadataVersion != null) {
+            Matcher m = MAJOR_MINOR_REGEX.matcher(metadataVersion);
+            if (m.matches() && !kafkaBrokerVersion.startsWith(m.group(1))) {
+                warnings.add(StatusUtils.buildWarningCondition("KafkaMetadataVersion",
+                        "Metadata version is older than the Kafka version used by the cluster, which suggests that an upgrade is incomplete."));
+            }
+        }
+    }
+
+    /**
+     * inter.broker.protocol.version should not be used in KRaft. This method checks if it is set and in case it is, it
+     * raises the warning.
+     *
+     * @param warnings List to add a warning to, if appropriate.
+     */
+    private void checkInterBrokerProtocolVersionInKRaft(List<Condition> warnings) {
+        String interBrokerProtocolVersion = kafkaCluster.getConfiguration().getConfigOption(KafkaConfiguration.INTERBROKER_PROTOCOL_VERSION);
+
+        if (interBrokerProtocolVersion != null) {
+            warnings.add(StatusUtils.buildWarningCondition("KafkaInterBrokerProtocolVersionInKRaft",
+                    "inter.broker.protocol.version is not used in KRaft-based Kafka clusters and should be removed from the Kafka custom resource."));
+        }
+    }
+
+    /**
+     * log.message.format.version should not be used in KRaft. This method checks if it is set and in case it is, it
+     * raises the warning.
+     *
+     * @param warnings List to add a warning to, if appropriate.
+     */
+    private void checkLogMessageFormatVersionInKRaft(List<Condition> warnings) {
+        String interBrokerProtocolVersion = kafkaCluster.getConfiguration().getConfigOption(KafkaConfiguration.LOG_MESSAGE_FORMAT_VERSION);
+
+        if (interBrokerProtocolVersion != null) {
+            warnings.add(StatusUtils.buildWarningCondition("KafkaLogMessageFormatVersionInKRaft",
+                    "log.message.format.version is not used in KRaft-based Kafka clusters and should be removed from the Kafka custom resource."));
         }
     }
 }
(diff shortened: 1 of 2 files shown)


fvaleri (approved):

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaSpecChecker.java
```diff
@@ -182,4 +187,53 @@ private void checkKRaftControllerCount(List<Condition> warnings)    {
...
+     *
+     * @param warnings List to add a warning to, if appropriate.
+     */
+    private void checkInterBrokerProtocolVersionInKRaft(List<Condition> warnings) {
+        String interBrokerProtocolVersion = kafkaCluster.getConfiguration().getConfigOption(KafkaConfiguration.INTERBROKER_PROTOCOL_VERSION);
+
+        if (interBrokerProtocolVersion != null) {
+            warnings.add(StatusUtils.buildWarningCondition("KafkaInterBrokerProtocolVersionInKRaft",
+                    "inter.broker.protocol.version is not used in KRaft-based Kafka clusters and should be removed from the Kafka custom resource."));
+        }
```
ppatierno, line 222: We should be aware that this warning will be in the Kafka status during a migration, because IBP is still set on brokers and the cluster is also using KRaft at the same time.
Of course, it would be transient.

scholzj: Why would it be there? It is gated by `useKRaft`, so it should be called only in KRaft mode.

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaSpecChecker.java
```diff
@@ -182,4 +187,53 @@ private void checkKRaftControllerCount(List<Condition> warnings)    {
...
+     *
+     * @param warnings List to add a warning to, if appropriate.
+     */
+    private void checkLogMessageFormatVersionInKRaft(List<Condition> warnings) {
+        String interBrokerProtocolVersion = kafkaCluster.getConfiguration().getConfigOption(KafkaConfiguration.LOG_MESSAGE_FORMAT_VERSION);
+
+        if (interBrokerProtocolVersion != null) {
+            warnings.add(StatusUtils.buildWarningCondition("KafkaLogMessageFormatVersionInKRaft",
+                    "log.message.format.version is not used in KRaft-based Kafka clusters and should be removed from the Kafka custom resource."));
+        }
```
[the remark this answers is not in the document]
scholzj: Same as above, it is gated to be called only in KRaft mode.

see-quick (approved):
LGTM 👍 . Few suggestions

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaSpecChecker.java
```diff
@@ -182,4 +187,53 @@ private void checkKRaftControllerCount(List<Condition> warnings)    {
...
     }
+
+    /**
+     * Checks if the version of the Kafka brokers matches any custom metadata version config. Updating this is the final
+     * step in upgrading Kafka version in KRaft, so if this doesn't match it is possibly an indication that a user has
+     * updated their Kafka cluster and is unaware that they also should update their metadata version to match.
+     *
+     * @param warnings List to add a warning to, if appropriate.
+     */
+    private void checkKafkaMetadataVersion(List<Condition> warnings) {
```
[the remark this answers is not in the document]
scholzj: > What about handling warnings == null?

Heh? `warnings` will never be null. 

> Also what about metadataVersion

What about it?

cluster-operator/src/test/java/io/strimzi/operator/cluster/model/KafkaSpecCheckerWithNodePoolsTest.java
```diff
@@ -202,4 +216,69 @@ public void testSingleKRaftControllerWithEphemeralStorage() {
         assertThat(warnings.get(0).getReason(), is("KafkaStorage"));
         assertThat(warnings.get(0).getMessage(), is("A Kafka cluster with a single controller node and ephemeral storage will lose data after any restart or rolling update."));
     }
+
+    @Test
+    public void testOldMetadataVersion() {
```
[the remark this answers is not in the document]
scholzj: I can change the names if needed. But please suggest names that are at least in sync with the other names used in these tests.

cluster-operator/src/test/java/io/strimzi/operator/cluster/model/KafkaSpecCheckerWithNodePoolsTest.java
```diff
@@ -202,4 +216,69 @@ public void testSingleKRaftControllerWithEphemeralStorage() {
...
+
+        List<Condition> warnings = checker.run(true);
+
+        assertThat(warnings, hasSize(1));
+        assertThat(warnings.get(0).getReason(), is("KafkaMetadataVersion"));
+        assertThat(warnings.get(0).getMessage(), is("Metadata version is older than the Kafka version used by the cluster, which suggests that an upgrade is incomplete."));
+    }
+
+    @Test
+    public void testSameMetadataVersion() {
```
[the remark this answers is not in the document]
scholzj: I can change the names if needed. But please suggest names that are at least in sync with the other names used in these tests.

cluster-operator/src/test/java/io/strimzi/operator/cluster/model/KafkaSpecCheckerWithNodePoolsTest.java
```diff
@@ -202,4 +216,69 @@ public void testSingleKRaftControllerWithEphemeralStorage() {
...
+        KafkaCluster kafkaCluster = KafkaCluster.fromCrd(Reconciliation.DUMMY_RECONCILIATION, kafka, pools, VERSIONS, new KafkaVersionChange(VERSIONS.defaultVersion(), VERSIONS.defaultVersion(), null, null, KafkaVersionTestUtils.LATEST_METADATA_VERSION), true, null, SHARED_ENV_PROVIDER);
+        KafkaSpecChecker checker = new KafkaSpecChecker(kafka.getSpec(), VERSIONS, kafkaCluster);
+
+        List<Condition> warnings = checker.run(true);
+
+        assertThat(warnings, hasSize(0));
+    }
+
+    @Test
+    public void testZooKeeperBasedVersionsInKRaft() {
```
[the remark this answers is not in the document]
scholzj: I can change the names if needed. But please suggest names that are at least in sync with the other names used in these tests.

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaSpecChecker.java
ppatierno: It's not actually true on my migration work. It's because the `useKRaft` passed to the checker is defined like this:

```java
result.warningConditions.addAll(specChecker.run(kafkaMetadataConfigState.isPreMigrationOrKRaft()));
```

So it's true since PRE_MIGRATION phase when the KRaft controllers are deployed, not just at the end when in full KRAFT state. It's because when deploying the controllers, we want the checker validating the first two things: controller storage and controller count. So I had to "adapt" the flag that way.
Anyway I don't think having the warning is a big issue, the opposite I think it's important so the user knows to remove IBP and LMF at the end of migration (or anyway the warning will disappear if the operator added and then removed them because not specified by user).

If we want to avoid that warning earlier, I should change in my branch the signature of the `run` method getting the configuration state and not just a boolean and then running the controller storage and count check if in PRE_MIGRATION while all the others only if in KRAFT.

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaSpecChecker.java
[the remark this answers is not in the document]
ppatierno: See my previous answer.

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaSpecChecker.java
scholzj: I would probably leave the warnings for KRaft mode only. It is just warnings anyway, so if you get them only after the migration is done, that is fine.

scholzj: You could also pass the `kafkaMetadataConfigState` into the spec checker and split the warnings. Do thse from PreMigration:
```
checkKRaftControllerStorage(warnings);
checkKRaftControllerCount(warnings);
checkKafkaMetadataVersion(warnings);
```

and these only in KRaft:

```
checkInterBrokerProtocolVersionInKRaft(warnings);
checkLogMessageFormatVersionInKRaft(warnings);
```

ppatierno: Ah ah ah ... we are overlapping the discussion :-)
I am fine with both approaches. As I said in the end we warn but don't stop the migration anyway (about controllers count and storage).

ppatierno: Yeah, let's leave the check for the end of the migration. Anyway, again we just warn but not stopping the migration.

PR #9435, opened 2023-12-05
docs(kraft): kraft-based upgrade and downgrade instructions

PaulRMellor:
**Documentation**

Updates the upgrade section in the deploying guide.

- Introduces upgrade and downgrade instructions specific to KRaft-based Kafka
- Changes the structure for ZooKeeper and KRaft-related sections
- Updates references and links to the upgrade section

### Checklist

- [ ] Write tests
- [x] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/documentation/modules/upgrading/proc-downgrade-kafka-kraft.adoc
+++ b/documentation/modules/upgrading/proc-downgrade-kafka-kraft.adoc
@@ -50,16 +50,7 @@ spec:
 +
 NOTE: The value of `metadataVersion` must be a string to prevent it from being interpreted as a floating point number.
 
-. Save the change, then wait for rolling updates to complete.
-+
-Check the progress of the rolling updates by watching the pod state transitions:
-+
-[source,shell,subs=+quotes]
-----
-kubectl get pods my-cluster-kafka-0 -o jsonpath='{.spec.containers[0].image}'
-----
-+
-The rolling updates ensure that each pod is using the specified Kafka metadata version.
+. Save the change, and wait for Cluster Operator to update `.status.kafkaMetadataVersion` for the `Kafka` resource. 
 
 . Change the `Kafka.spec.kafka.version` to the previous version.
 +
--- a/documentation/modules/upgrading/con-upgrade-status.adoc
+++ b/documentation/modules/upgrading/con-upgrade-status.adoc
@@ -13,6 +13,7 @@ To ensure that you have the correct versions after completing an upgrade, verify
 
 * `operatorLastSuccessfulVersion` is the version of the Strimzi operator that last performed a successful reconciliation.
 * `kafkaVersion` is the the version of Kafka being used by the Kafka cluster.
+* `kafkaMetadataVersion` is the metadata version used by KRaft-based Kafka clusters 
 
 You can use these values to check an upgrade of Strimzi or Kafka has completed.
 
@@ -28,4 +29,5 @@ status:
   # ...
   kafkaVersion: {DefaultKafkaVersion} 
   operatorLastSuccessfulVersion: {ProductVersion}
+  kafkaMetadataVersion: {DefaultKafkaMetadataVersion}
 ----


documentation/modules/upgrading/proc-downgrade-kafka-kraft.adoc
```diff
@@ -0,0 +1,80 @@
...
+----
+kubectl get pod -w
+----
++
+Check the Cluster Operator logs for an `INFO` level message:
++
+[source,shell,subs=+quotes]
+----
+Reconciliation #_NUM_(watch) Kafka(_NAMESPACE_/_NAME_): Kafka version downgrade from _FROM-VERSION_ to _TO-VERSION_, phase 1 of 1 completed
+----
```
PaulRMellor, line 76: Maybe we prefer to say check resource status here like upgrades

documentation/assemblies/upgrading/assembly-upgrade-zookeeper.adoc
```diff
@@ -0,0 +1,14 @@
+// This assembly is included in the following assemblies:
+//
+// assembly-upgrade.adoc
+
+[id='assembly-upgrade-zookeeper-{context}']
+= Upgrading Kafka when using ZooKeeper
+
+[role="_abstract"]
+If you are using ZooKeeper for cluster management, a kafka upgrade requires an update to the Kafka version and the inter-broker protocol version.
```
scholzj, line 9: Maybe this would sound better? I think `using ZooKeeper for cluster management` might be a bit confusing for some.

```suggestion
If you are using ZooKeeper-based cluster, a Kafka upgrade requires an update to the Kafka version and the inter-broker protocol version.
```

documentation/modules/configuring/con-config-kafka.adoc
```diff
@@ -21,18 +21,14 @@ Configuration options that are particularly important include the following:
...
+* Metadata version for KRaft-based Kafka clusters
+* Inter-broker protocol version for ZooKeeper-based Kafka clusters
 
-For a deeper understanding of the Kafka cluster configuration options, refer to the link:{BookURLConfiguring}[Strimzi Custom Resource API Reference^].
-
-.Kafka versions
-
-The `inter.broker.protocol.version` property for the Kafka `config` must be the version supported by the specified Kafka version (`spec.kafka.version`).
-The property represents the version of Kafka protocol used in a Kafka cluster.
+The `metadataVersion` or `inter.broker.protocol.version` property for the Kafka `config` must be a version supported by the specified Kafka version (`spec.kafka.version`).
```
scholzj, line 27: Can we make it more clear here that `metadataVersion` is a separate field in `.spec.kafka` but `inter.broker.protocol.version` is a field in the `.spec.kafka.config`? I think it will be confusing the way it is written now (as many users might be confused the metadata version is not in the Kafka config anymore and might misread it). Maybe like this?

```suggestion
The `.spec.kafka.metadataVersion` field or `inter.broker.protocol.version` property in `.spec.kafka.config` must be a version supported by the specified Kafka version (`spec.kafka.version`).
```

documentation/modules/configuring/con-config-kafka.adoc
```diff
@@ -21,18 +21,14 @@ Configuration options that are particularly important include the following:
...
+The property represents the Kafka metadata or inter-broker protocol version used in a Kafka cluster.
+If either of these properties is not set in the configuration, the Cluster Operator updates the version to the default for the Kafka version used.
```
scholzj, line 29: Should we maybe add here that:
* The oldest supported metadata version is `3.3`
* Using an older metadata version than the Kafka version might cause some KAfka features to be disabled?

documentation/modules/deploying/proc-deploy-kafka-cluster.adoc
```diff
@@ -10,7 +10,7 @@ This procedure shows how to deploy a Kafka cluster to your Kubernetes cluster us
 
 The deployment uses a YAML file to provide the specification to create a `Kafka` resource.
 
-Strimzi provides the following xref:config-examples-{context}[example files] you can use to create a Kafka cluster:
+Strimzi provides the following xref:config-examples-{context}[example files] to create a Kafka cluster that uses ZooKeeper for cluster management:
```
scholzj, line 13: The example files have KRaft-based examples as well. You do not use them in this section and they are not specifically listed. But I wonder if leaving it more generic is better. Or maybe at least mention that KRaft examples are in `examples/kafka/nodepools`?

documentation/modules/upgrading/con-upgrade-older-clients.adoc
```diff
@@ -0,0 +1,30 @@
+// Module included in the following assemblies:
+//
+// assembly-upgrade-zookeeper.adoc
+
+[id='con-upgrade-older-clients-{context}']
+= Upgrading clients with older message formats
+
+[role="_abstract"]
+Before Kafka 3.0, you would configure a specific message format for brokers using the `log.message.format.version` property (or the `message.format.version` property at the topic level). 
+This allowed brokers to support older Kafka clients that were using an outdated message format.
```
scholzj, line 10: I think this is misleading as even without this field the older client are still supported.

documentation/modules/upgrading/con-upgrade-older-clients.adoc
scholzj, line 1: @mimaison Do you think you can check this description to make sure it is correct from Kafka's perspective? Thanks.

documentation/modules/upgrading/con-upgrade-sequence.adoc
```diff
@@ -14,6 +14,10 @@ Strimzi {ProductVersion} requires Kubernetes {KubernetesVersion}.
...
 You can xref:con-upgrade-cluster-{context}[upgrade Kubernetes with minimal downtime].
 
-. xref:assembly-upgrade-cluster-operator-{context}[Upgrade the Cluster Operator].
+. xref:assembly-upgrade-{context}[Upgrade the Cluster Operator].
 
-. xref:assembly-upgrading-kafka-versions-{context}[Upgrade all Kafka brokers and client applications] to the latest supported Kafka version.
\ No newline at end of file
+. Upgrade Kafka depending on the cluster configuration:
+.. If using Kafka in KRaft mode, use the latest supported Kafka version and `spec.kafka.metadataVersion` to xref:proc-upgrade-kafka-kraft-{context}[upgrade all Kafka brokers and client applications].
+.. If using Kafka with ZooKeeper for cluster management, use the latest supported Kafka version and `inter.broker.protocol.version` to xref:assembly-upgrade-zookeeper-{context}[upgrade all Kafka brokers and client applications].
```
scholzj, line 21: I'm not sure I understand the meaning of the `use the latest supported` in both of these bullets. The point of upgrade is to upgrade to a newer version. Should it be something like `in order to use the latest supported` or something like that?

documentation/modules/upgrading/proc-downgrade-kafka-kraft.adoc
```diff
@@ -0,0 +1,80 @@
+// This module is included in the following assemblies:
+//
+// assembly-downgrade.adoc
+
+[id='proc-downgrade-kafka-kraft-{context}']
+= Downgrading KRaft-based Kafka clusters and client applications
+
+[role="_abstract"]
+Downgrade a KRaft-based Strimzi Kafka cluster to an earlier version.
+When downgrading a KRaft-based Strimzi Kafka cluster to a lower version, like moving from {KafkaVersionHigher} to {KafkaVersionLower}, ensure that the `spec.kafka.metadataVersion` property in the Kafka configuration aligns with a version supported by the specified Kafka version (`spec.kafka.version`).
```
scholzj, line 10: I would write this more generic -> this is not about what is set in the custom resource but what the cluster actually uses. For example, if you change the `.spec.kafka.metadataVerison` to `3.5`, but the Kafka cluster does not accept it, the downgrade still cannot be done.

```suggestion
When downgrading a KRaft-based Strimzi Kafka cluster to a lower version, like moving from {KafkaVersionHigher} to {KafkaVersionLower}, ensure that the metadata version used by the Kafka cluster is a version supported by the Kafka version you want to downgrade to. 
```

documentation/modules/upgrading/proc-downgrade-kafka-kraft.adoc
```diff
@@ -0,0 +1,80 @@
...
+
+NOTE: Consult the Apache Kafka documentation for information regarding the support and limitations associated with KRaft-based downgrades.
+
+.Prerequisites
+
+* The Cluster Operator is up and running.
+* Before you downgrade the Strimzi Kafka cluster, check the following for the `Kafka` resource:
+
+** `Kafka.spec.kafka.config` does not contain options that are not supported by the Kafka version being downgraded to.
+** `Kafka.spec.kafka.config` has a `spec.kafka.metadataVersion` that is supported by the Kafka version being downgraded to.
```
scholzj, line 21:
```suggestion
** `Kafka` custom resource does not contain options that are not supported by the Kafka version being downgraded to.
** `spec.kafka.metadataVersion` is set to a version that is supported by the Kafka version being downgraded to.   
```

documentation/modules/upgrading/proc-downgrade-kafka-kraft.adoc
scholzj, line 1: I left some comments here. But I think this needs to be written bit differently. Either as two separate main steps:
1. Downgrade the metadata version to use metadata version supported by the KAfka version you want to downgrade to
2. Downgrade the Kafka version

Or alternatively:
* Have as a clear prerequisite that the metadata version needs to be already set to version supported by the Kafka version we are downgrading to.

RIght now, you seem to cover both options in what sounds like a single step (single edit to the Kafka CR) and that does not work for downgrades. This is the same for ZooKeeper-based downgrade. But there it seems to be clearer that the inter.broker.protocol.version has to be changed before.

documentation/modules/upgrading/proc-upgrade-kafka-kraft.adoc
```diff
@@ -0,0 +1,130 @@
+// This module is included in the following assemblies:
+//
+// assembly-upgrade.adoc
+
+[id='proc-upgrade-kafka-kraft-{context}']
+= Upgrading KRaft-based Kafka clusters and client applications
+
+[role="_abstract"]
+Upgrade a KRaft-based Strimzi Kafka cluster to the latest supported Kafka version and KRaft metadata version.
```
scholzj, line 9: As people might not always upgrade only to the latest version, maybe this is clearer?

```suggestion
Upgrade a KRaft-based Strimzi Kafka cluster to a newer supported Kafka version and KRaft metadata version.
```

documentation/modules/upgrading/proc-upgrade-kafka-kraft.adoc
```diff
@@ -0,0 +1,130 @@
...
+.Procedure
+
+. Update the Kafka cluster configuration:
++
+[source,shell,subs=+quotes]
+----
+kubectl edit kafka <kafka_configuration_file>
+----
+
+. If configured, check that the `spec.kafka.metadataVersion` is set to the _current_ version.
```
scholzj, line 30: I do not think it is clear what _current_ means. At this point, it should be normally set to a metadata version supported by both the old version (we use today) and new version (we want to upgrade to).

documentation/modules/upgrading/proc-upgrade-kafka-kraft.adoc
```diff
@@ -0,0 +1,130 @@
...
+. Depending on your chosen xref:con-strategies-for-upgrading-clients-{context}[strategy for upgrading clients], upgrade all client applications to use the new version of the client binaries.
++
+If required, set the `version` property for Kafka Connect and MirrorMaker as the new version of Kafka:
++
+.. For Kafka Connect, update `KafkaConnect.spec.version`.
+.. For MirrorMaker, update `KafkaMirrorMaker.spec.version`.
+.. For MirrorMaker 2, update `KafkaMirrorMaker2.spec.version`.
++
+NOTE: If you are using custom images that are built manually, you must rebuild those images to ensure that they are up-to-date with the latest Strimzi base image. 
+For example, if you xref:creating-new-image-from-base-str[created a Docker image from the base Kafka Connect image], update the Dockerfile to point to the latest base image and build configuration.
```
scholzj, line 106:
```suggestion
For example, if you xref:creating-new-image-from-base-str[created a container image from the base Kafka Connect image], update the Dockerfile to point to the latest base image and build configuration.
```

documentation/modules/upgrading/proc-upgrade-kafka-kraft.adoc
```diff
@@ -0,0 +1,130 @@
...
+kind: Kafka
+metadata:
+  name: my-cluster
+spec:
+  kafka:
+    replicas: 3
+    metadataVersion: {KafkaMetadataVersionHigher}
+    version: {KafkaVersionHigher}
+    # ...
+----
```
scholzj, line 126: Now this is a bit tricky ... 

We should have a big warning here saying that Kafka support for downgrading the metadata version is limited and and might not be possible. And we should move here the warning part from line 80 that if you upgrade this and cannot downgrade, you might not be able to downgrade Kafka. Yet, at the same time we should make it clear that they have to eventually update the metadata version and cannot stick with the old one forever.

documentation/shared/attributes.adoc
```diff
@@ -25,6 +25,8 @@
 // Kafka upgrade attributes used in kafka upgrades section
 :DefaultKafkaVersion: 3.6.0
 :DefaultKafkaMetadataVersion: 3.6
+:KafkaMetadataVersionLower: 3.5
+:KafkaMetadataVersionHigher: 3.6
```
scholzj, line 29: I think we should use the IV versions here ... we use them in the examples as well because without it patch version downgrades might be possible in some edge cases.

```suggestion
:KafkaMetadataVersionLower: 3.5-IV2
:KafkaMetadataVersionHigher: 3.6-IV2
```

scholzj (commented):
Thanks for the PR! I left some comments n few places.

documentation/modules/upgrading/proc-downgrade-kafka-kraft.adoc
scholzj: I wonder if these messages are still logged. I cannot find in the code.

documentation/modules/deploying/proc-deploy-kafka-cluster.adoc
PaulRMellor: The node pools examples are mentioned in a separate deploying node pools section after this one: [5.3.2. (Preview) Deploying Kafka node pools](https://strimzi.io/docs/operators/in-development/deploying#deploying-kafka-node-pools-str)

We should take the `preview` off the node pool sections. I'll do that in a separate PR.

documentation/modules/upgrading/con-upgrade-older-clients.adoc
PaulRMellor: I've rewritten as follows:

> Before Kafka 3.0, you could configure a specific message format for brokers using the `log.message.format.version` property (or the `message.format.version` property at the topic level). 
> This allowed brokers to accommodate older Kafka clients that were using an outdated message format. 
> Though Kafka inherently supports older clients without explicitly setting this property, brokers would then need to convert the messages from the older clients, which came with a significant performance cost.

documentation/modules/upgrading/con-upgrade-sequence.adoc
PaulRMellor: Supported by Strimzi. I changed to "update the..."

documentation/modules/upgrading/proc-downgrade-kafka-kraft.adoc
PaulRMellor: I've tried to make it clearer in both procedures by introducing a separate step to update the metadata version or protocol version before the Kafka version.

documentation/modules/upgrading/proc-upgrade-kafka-kraft.adoc
PaulRMellor: changed to:   If configured, check that the current `spec.kafka.metadataVersion` is set to a version supported by the version of Kafka you are upgrading to.

documentation/modules/upgrading/proc-upgrade-kafka-kraft.adoc
PaulRMellor: added the following:

> WARNING: Exercise caution when changing the `metadataVersion`, as downgrading may not be possible. 
> You cannot downgrade Kafka if the `metadataVersion` for the new Kafka version is higher than the Kafka version you wish to downgrade to. 
> However, understand the potential implications on support and compatibility when maintaining an older version.

documentation/modules/upgrading/proc-downgrade-kafka-kraft.adoc
PaulRMellor: I removed all this and pointed to the resource status section.

documentation/modules/upgrading/proc-downgrade-kafka-kraft.adoc
```diff
@@ -0,0 +1,95 @@
...
+    version: {KafkaVersionHigher} # <2>
+    # ...
+----
++
+<1> Metadata version is changed to a version supported by the earlier Kafka version.
+<2> Kafka version is unchanged.
++
+NOTE: The value of `metadataVersion` must be a string to prevent it from being interpreted as a floating point number.
+
+. Save the change, then wait for rolling updates to complete.
```
scholzj, line 53: This will not have rolling update. But the `.status.kafkaMetadataVersion` would be updated by the operator once/if the metadata version is changed.

PaulRMellor: Updated.
I also updated the section on _Checking the status of an upgrade_ so that we now show kafkaMetadataVersion too.

scholzj (approved):
LGTM. Thanks.

mimaison (approved):
LGTM, just left a minor suggestion

PR #9441, opened 2023-12-06
fix(helm-charts): add check for optional env

SagdeevRR:
### Type of change
- Bugfix

### Description
Optional value [featureGates](https://github.com/strimzi/strimzi-kafka-operator/blob/87e973330fc25d5f487f14fd48772e7944057b0b/helm-charts/helm3/strimzi-kafka-operator/values.yaml#L29C1-L29C13) for env `STRIMZI_FEATURE_GATES` in default empty.

### Checklist

- [ ] Write tests
- [ ] Update documentation
- [ ] Update CHANGELOG.md


scholzj (commented):
Thanks for the PR. However, does it really need a new field? You can use the existing `extraEnvs` to specify additional environment variables.

SagdeevRR:
You suggest tottaly removing  `STRIMZI_FEATURE_GATES` in deployment?

Now it looks like this:
```
...
            - name: STRIMZI_FEATURE_GATES
              value: ""
...
```
I think this is incorrect.

scholzj:
Sorry, I guess I misunderstood what the PR does. But what is wrong with:
```
            - name: STRIMZI_FEATURE_GATES
              value: ""
```

I should work just fine.

scholzj:
In any case, if you would want to make this change, you have to do it in the `packaging/helm-charts` folder and not in `helm-charts/`.

scholzj (commented):
Could you please fix the DCO signoff? The instructions should be under the _Details_ link next to the DCO status.

scholzj:
Hmm, this seems to change the main installation files 🤔. Do we want the empty variable there as a guidance for users what to change? WDYT @ppatierno?

ppatierno:
@scholzj Sorry, I didn't get what you are asking for 🤔

scholzj:
@ppatierno What I mean is following:
* Today, we have an empty `STIRMZI_FEATURE_GATES` environment variable in the YAML installation files
* This PR removes it

I wondered if having the empty env var makes it easier for users to configure the feature gates as they already have part of it there. But I guess if you do not know how to configure an environment variable you should maybe not configure the feature gates either 🤷. WDYT?

ppatierno:
> I wondered if having the empty env var makes it easier for users to configure the feature gates as they already have part of it there.

So I think our YAMLs and Helm Charts should be similar. So the env var should be on both or nothing.
This PR just removes it from Helm Charts but not from the YAML. Said that ... I would leave it on both.
I know that we have other env vars which are not defined as empty in the operator deployment but I think the feature gate one is a "key" one. It's better taking it imho.

scholzj:
@ppatierno So, you suggest to reject this PR?

ppatierno:
> @ppatierno So, you suggest to reject this PR?

That's my opinion yes. Of course, it's not just me and you. Maybe we should involve other maintainers and see what they think about this? At least leaving this open for a few time giving them time to review or talking about it during the next Strimzi community call?

scholzj:
Ok.

scholzj:
Discussed in the community call on 14.12.: Thanks for the PR. The feature gates environment variables seems to be useful to us even if it is empty as it makes it easier for the users to set it if needed. While the change in the Helm Chart itself does seem to be minor, merging this PR would require also the change to the YAML installation files which does not seem to be desired.

PR #9442, opened 2023-12-06
[ST] KRaft to KRaft upgrades/downgrades in STs

im-konge:
### Type of change

- Enhancement 

### Description

This PR adds STs for KRaft to KRaft upgrades/downgrades.
I decided to make a copy of our current tests that we have in our suite and create new package `kraft` for the KRaft to KRaft STs. As part of that, I moved the "regular" tests into the `regular` package.

As for the regular tests, for KRaft upgrades/downgrades I'm using the `AbstractKRaftUpgradeST` class, which extends the `AbstractUpgradeST`. To keep the methods same as much as possible, I moved some parts to separate methods, which are then overriden in the `AbstractKRaftUpgradeST`. So even if we are executing some method in the `AbstractUpgradeST`, if there is some method in the `AbstractKRaftUpgradeST` that overrides it, the test uses that one.

`KRaftKafkaUpgradeDowngradeST` works as expected, it does testing of just Kafka upgrade/downgrade. The two classes that are testing the Strimzi & Kafka upgrades/downgrades are implemented as well, the `KRaftStrimziUpgradeST` works, but the `KRaftStrimziDowngradeST` is currently disabled, as `0.38.0` version of Strimzi doesn't support KRaft to KRaft upgrades/downgrades, so that scenario will not currently work. But after `0.39.0` is released, it should be enabled -> and from my testing the tests inside the downgrade suite should work.

As part of this PR I'm changing few classes regarding Kafka version (adding the `metadataVersion`), loaders for the upgrade/downgrade tests (possibility to specify "always enabled" FG -> used for KRaft, UTO, and KafkaNodePools FGs), 
and also adding some helper methods for the NodePools creation to the templates or LabelSelector for getting Pods for particular NodePool.

Finally, I'm adding these STs to AZP and updating our `pom.xml` with `kraft_upgrade` profile + adding `KRaftKafkaUpgradeDowngradeST` to `azp_kafka_upgrade` profile.

### Checklist

- [ ] Write tests

### DIFF
--- a/systemtest/src/test/java/io/strimzi/systemtest/upgrade/kraft/AbstractKRaftUpgradeST.java
+++ b/systemtest/src/test/java/io/strimzi/systemtest/upgrade/kraft/AbstractKRaftUpgradeST.java
@@ -1,0 +1,228 @@
+/*
+ * Copyright Strimzi authors.
+ * License: Apache License 2.0 (see the file LICENSE or http://apache.org/licenses/LICENSE-2.0.html).
+ */
+package io.strimzi.systemtest.upgrade.kraft;
+
+import io.fabric8.kubernetes.api.model.LabelSelector;
+import io.strimzi.api.kafka.model.Kafka;
+import io.strimzi.api.kafka.model.KafkaResources;
+import io.strimzi.api.kafka.model.nodepool.ProcessRoles;
+import io.strimzi.operator.common.Annotations;
+import io.strimzi.systemtest.TestConstants;
+import io.strimzi.systemtest.resources.ResourceManager;
+import io.strimzi.systemtest.resources.crd.KafkaNodePoolResource;
+import io.strimzi.systemtest.templates.crd.KafkaNodePoolTemplates;
+import io.strimzi.systemtest.templates.crd.KafkaTemplates;
+import io.strimzi.systemtest.upgrade.AbstractUpgradeST;
+import io.strimzi.systemtest.upgrade.BundleVersionModificationData;
+import io.strimzi.systemtest.upgrade.CommonVersionModificationData;
+import io.strimzi.systemtest.upgrade.UpgradeKafkaVersion;
+import io.strimzi.systemtest.utils.RollingUpdateUtils;
+import io.strimzi.systemtest.utils.TestKafkaVersion;
+import io.strimzi.systemtest.utils.kafkaUtils.KafkaUserUtils;
+import io.strimzi.systemtest.utils.kafkaUtils.KafkaUtils;
+import io.strimzi.systemtest.utils.kubeUtils.controllers.DeploymentUtils;
+import io.strimzi.systemtest.utils.kubeUtils.objects.PodUtils;
+import io.strimzi.test.TestUtils;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+import org.junit.jupiter.api.extension.ExtensionContext;
+
+import java.io.File;
+import java.io.IOException;
+import java.util.Map;
+
+import static io.strimzi.test.k8s.KubeClusterResource.cmdKubeClient;
+import static io.strimzi.test.k8s.KubeClusterResource.kubeClient;
+import static org.junit.jupiter.api.Assertions.fail;
+
+public class AbstractKRaftUpgradeST extends AbstractUpgradeST {
+
+    private static final Logger LOGGER = LogManager.getLogger(AbstractKRaftUpgradeST.class);
+
+    protected Map<String, String> brokerPods;
+    protected Map<String, String> controllerPods;
+
+    protected static final String CONTROLLER_NODE_NAME = "controller";
+    protected static final String BROKER_NODE_NAME = "broker";
+
+    protected final LabelSelector controllerSelector = KafkaNodePoolResource.getLabelSelector(clusterName, CONTROLLER_NODE_NAME, ProcessRoles.CONTROLLER);
+    protected final LabelSelector brokerSelector = KafkaNodePoolResource.getLabelSelector(clusterName, BROKER_NODE_NAME, ProcessRoles.BROKER);
+
+    // topics that are just present in Kafka itself are not created as CRs in UTO, thus -3 topics in comparison to regular upgrade
+    protected final int expectedTopicCount = upgradeTopicCount;
+
+    protected int getExpectedTopicCount() {
+        return expectedTopicCount;
+    }
+
+    @Override
+    protected void makeSnapshots() {
+        coPods = DeploymentUtils.depSnapshot(TestConstants.CO_NAMESPACE, ResourceManager.getCoDeploymentName());
+        eoPods = DeploymentUtils.depSnapshot(TestConstants.CO_NAMESPACE, KafkaResources.entityOperatorDeploymentName(clusterName));
+        controllerPods = PodUtils.podSnapshot(TestConstants.CO_NAMESPACE, controllerSelector);
+        brokerPods = PodUtils.podSnapshot(TestConstants.CO_NAMESPACE, brokerSelector);
+        connectPods = PodUtils.podSnapshot(TestConstants.CO_NAMESPACE, connectLabelSelector);
+    }
+
+    @Override
+    protected void deployKafkaClusterWithWaitForReadiness(final ExtensionContext extensionContext, final BundleVersionModificationData upgradeData,
+                                                          final UpgradeKafkaVersion upgradeKafkaVersion) {
+        LOGGER.info("Deploying Kafka: {} in Namespace: {}", clusterName, kubeClient().getNamespace());
+
+        if (!cmdKubeClient().getResources(getResourceApiVersion(Kafka.RESOURCE_PLURAL)).contains(clusterName)) {
+            // Deploy a Kafka cluster
+            if (upgradeData.getFromExamples().equals("HEAD")) {
+                resourceManager.createResourceWithWait(extensionContext,
+                    KafkaNodePoolTemplates.kafkaNodePoolWithControllerRoleAndPersistentStorage(TestConstants.CO_NAMESPACE, CONTROLLER_NODE_NAME, clusterName, 3).build(),
+                    KafkaNodePoolTemplates.kafkaNodePoolWithBrokerRoleAndPersistentStorage(TestConstants.CO_NAMESPACE, BROKER_NODE_NAME, clusterName, 3).build(),
+                    KafkaTemplates.kafkaPersistent(clusterName, 3, 3)
+                        .editMetadata()
+                            .addToAnnotations(Annotations.ANNO_STRIMZI_IO_NODE_POOLS, "enabled")
+                            .addToAnnotations(Annotations.ANNO_STRIMZI_IO_KRAFT, "enabled")
+                        .endMetadata()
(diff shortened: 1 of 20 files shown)


systemtest/src/test/java/io/strimzi/systemtest/upgrade/kraft/AbstractKRaftUpgradeST.java
```diff
@@ -0,0 +1,227 @@
...
+
+    protected void changeKafkaAndMetadataVersion(CommonVersionModificationData versionModificationData, ExtensionContext extensionContext) throws IOException {
+        changeKafkaAndMetadataVersion(versionModificationData, false, extensionContext);
+    }
+
+    // replaceEvenIfMissing is current workaround for the situation when `metadataVersion` is not set in Kafka CR -> that's because previous version of operator
+    // doesn't contain this kind of field, so even if we set this field in the Kafka CR, it is removed by the operator
+    // this is needed for correct functionality of the `testUpgradeAcrossVersionsWithUnsupportedKafkaVersion` test
+    @SuppressWarnings("CyclomaticComplexity")
+    protected void changeKafkaAndMetadataVersion(CommonVersionModificationData versionModificationData, boolean replaceEvenIfMissing, ExtensionContext extensionContext) throws IOException {
```
see-quick, line 139: This method is really hard to read in matter of complexity :) Wouldn't be better to somehow refactor part of this method to separate methods to make this more readable? :)

systemtest/src/test/java/io/strimzi/systemtest/upgrade/kraft/KRaftStrimziDowngradeST.java
```diff
@@ -0,0 +1,108 @@
...
+import java.util.List;
+
+import static io.strimzi.systemtest.TestConstants.CO_NAMESPACE;
+import static io.strimzi.systemtest.TestConstants.INTERNAL_CLIENTS_USED;
+import static io.strimzi.systemtest.TestConstants.KRAFT_UPGRADE;
+import static org.junit.jupiter.api.Assumptions.assumeTrue;
+
+@Tag(KRAFT_UPGRADE)
+@Disabled("The tests are currently disabled, as the KRaft to KRaft downgrade (with operator downgrade) is not handled in Strimzi 0.38.0")
+public class KRaftStrimziDowngradeST extends AbstractKRaftUpgradeST {
```
see-quick, line 36: Adding some Java docs to this class would be good. At least what is the purpose of it

systemtest/src/test/java/io/strimzi/systemtest/upgrade/kraft/KRaftKafkaUpgradeDowngradeST.java
```diff
@@ -0,0 +1,324 @@
...
+import java.util.Map;
+
+import static io.strimzi.systemtest.TestConstants.KRAFT_UPGRADE;
+import static io.strimzi.test.k8s.KubeClusterResource.kubeClient;
+import static org.hamcrest.CoreMatchers.is;
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.junit.jupiter.api.Assertions.assertFalse;
+
+@Tag(KRAFT_UPGRADE)
+public class KRaftKafkaUpgradeDowngradeST extends AbstractKRaftUpgradeST {
```
see-quick, line 45: Adding some Javadoc for this class would be great for others when they would add some test cases into such a class.

systemtest/src/test/resources/upgrade/BundleUpgrade.yaml
```diff
@@ -24,26 +24,26 @@
...
+#    userOperator: strimzi/operator:latest
+#  client:
+#    continuousClientsMessages: 500
+#  environmentInfo:
+#    maxK8sVersion: latest
+#    status: stable
+#    flakyEnvVariable: none
+#    reason: Test is working on environment, where k8s server version is < 1.22
+#  featureGatesBefore: ""
+#  featureGatesAfter: ""
```
see-quick, line 46: Hmmm ? :D

see-quick (commented):
Just a few nits :) but this is a huge amount of work! Good Job! 👍

systemtest/src/test/java/io/strimzi/systemtest/upgrade/kraft/AbstractKRaftUpgradeST.java
im-konge: Yeah the same one is in regular upgrade/downgrade and it's also really hard to read. I will have a look on this one and in some next PR I can change the one for regular upgrade.

systemtest/src/test/java/io/strimzi/systemtest/upgrade/kraft/KRaftKafkaUpgradeDowngradeST.java
```diff
@@ -0,0 +1,328 @@
...
+
+        String clusterName = testStorage.getClusterName();
+        String producerName = clusterName + "-producer";
+        String consumerName = clusterName + "-consumer";
+
+        for (int x = sortedVersions.size() - 1; x > 0; x--) {
+            TestKafkaVersion initialVersion = sortedVersions.get(x);
+            TestKafkaVersion newVersion = sortedVersions.get(x - 1);
+
+            // If it is a downgrade then we make sure to use the lower version number for the message format
```
scholzj, line 92:
```suggestion
            // If it is a downgrade then we make sure to use the lower version number for the metadata version
```

systemtest/src/test/java/io/strimzi/systemtest/upgrade/kraft/KRaftKafkaUpgradeDowngradeST.java
```diff
@@ -0,0 +1,328 @@
...
+        List<TestKafkaVersion> sortedVersions = TestKafkaVersion.getSupportedKafkaVersions();
+
+        String producerName = clusterName + "-producer";
+        String consumerName = clusterName + "-consumer";
+
+        for (int x = 0; x < sortedVersions.size() - 1; x++) {
+            TestKafkaVersion initialVersion = sortedVersions.get(x);
+            TestKafkaVersion newVersion = sortedVersions.get(x + 1);
+
+            // If it is an upgrade test we keep the message format as the lower version number
```
scholzj, line 66:
```suggestion
            // If it is an upgrade test we keep the metadata version as the lower version number
```

systemtest/src/test/java/io/strimzi/systemtest/upgrade/kraft/KRaftKafkaUpgradeDowngradeST.java
```diff
@@ -0,0 +1,328 @@
...
+        String clusterName = testStorage.getClusterName();
+        String producerName = clusterName + "-producer";
+        String consumerName = clusterName + "-consumer";
+
+        for (int x = sortedVersions.size() - 1; x > 0; x--) {
+            TestKafkaVersion initialVersion = sortedVersions.get(x);
+            TestKafkaVersion newVersion = sortedVersions.get(x - 1);
+
+            // If it is a downgrade then we make sure to use the lower version number for the message format
+            String metadataVersion = newVersion.metadataVersion();
```
scholzj, line 93: I suspect this is wrong? If you have multiple patch releases as today and loop from newest to oldest, you will:
* Start by deploying 3.6.1 and downgrading to 3.6.0
* 3.6.0 will give you metadata version 3.6
* In next iteration, you try to go from 3.6.0 to 3.5.1 and you will fail to downgrade the metadata?

I guess the matadata version here should be always from `sortedVersions.get(0).metadataVersion()`?

systemtest/src/test/java/io/strimzi/systemtest/upgrade/kraft/KRaftKafkaUpgradeDowngradeST.java
```diff
@@ -0,0 +1,328 @@
...
+
+        // ##############################
+        // Validate that continuous clients finished successfully
+        // ##############################
+        ClientUtils.waitForClientsSuccess(producerName, consumerName, TestConstants.CO_NAMESPACE, continuousClientsMessageCount);
+        // ##############################
+    }
+
+    @IsolatedTest
+    @Disabled("This is currently not possible - missing support for unsafe downgrades")
```
scholzj, line 105: Not sure we really want to have this test -> the operator will not do the unsafe downgrades. The user will have to do them. The _unsafe_ downgrade also means it might not work in some cases.

systemtest/src/test/java/io/strimzi/systemtest/upgrade/kraft/KRaftKafkaUpgradeDowngradeST.java
```diff
@@ -0,0 +1,328 @@
...
+
+        } else {
+            LOGGER.info("Initial Kafka version (" + initialVersion.version() + ") is already ready");
+            controllerPods = PodUtils.podSnapshot(TestConstants.CO_NAMESPACE, controllerSelector);
+            brokerPods = PodUtils.podSnapshot(TestConstants.CO_NAMESPACE, brokerSelector);
+
+            // Wait for metadataVersion change
+            if (!sameMinorVersion
+                && !isUpgrade
+                && !testContext.getDisplayName().contains("DowngradeToOlderMetadataVersion")) {
```
scholzj, line 232: What does this mean?

systemtest/src/test/java/io/strimzi/systemtest/upgrade/kraft/KRaftKafkaUpgradeDowngradeST.java
im-konge: Ah okay, I thought that there will be some kind of option in the future. So I just misread it. Thanks :)

systemtest/src/test/java/io/strimzi/systemtest/upgrade/kraft/KRaftKafkaUpgradeDowngradeST.java
im-konge: The `!testContext.getDisplayName().contains("DowngradeToOlderMetadataVersion")` check is now removed as I removed the test that it mentions.

Otherwise I guess this if block is not relevant, as it should change the `metadataVersion` in case that we are doing downgrade test and we have different minor version (so f.e. downgrade from 3.6 -> 3.5) -> that's not possible.

systemtest/src/test/java/io/strimzi/systemtest/upgrade/kraft/KRaftKafkaUpgradeDowngradeST.java
scholzj: Yes, correct. It has to be 3.5 to be able to downgrade down to 3.5.1

im-konge: Cool, thanks :)

scholzj (approved):
LGTM assuming the tests pass. Should be reviewed by SMEs.

henryZrncik (approved):
great.

see-quick (approved):
LGTM. Thanks for this PR @im-konge ! 💯

PR #9428, opened 2023-12-04
[ST] KafkaNodePoolST add test addRemoveBrokerKNP into Kafka Cluster 

henryZrncik:
### Type of change
- Enhancement 
- Refactoring

### Description

added new test to add and remove KafkaNodePools (now only with `broker` role) into existing Kafka Cluster. 
also allowing KafkaNodePoolST to run without Kraft, (as there is not yet test which would requiere Kraft anyway) and modify tests to run with or without kraft. 
### Checklist

- [x] Write tests
- [x] Make sure all tests pass

PR #9453, opened 2023-12-11
[system test] - KafkaRollerST managing rolling updates across mixed, broker, and controller node

see-quick:
### Type of change

- Enhancement / new feature
- Refactoring

### Description

This PR adds a two test case:
1. testKafkaRollingUpdatesOfControllerNodes - verifies the rolling update behaviour of Kafka controller nodes under specific conditions. It ensures that changes in Kafka configuration and node pool properties affect only the intended node pools, particularly the controller nodes, while leaving others like broker nodes unaffected.
2. testKafkaRollingUpdatesOfMixedNodes - assesses the rolling update behaviour of Kafka nodes in mixed-role configurations.
The main focus is to verify that configuration changes impacting controller roles result in rolling updates of nodes serving mixed roles while ensuring compatibility with KRaft mode and excluding OLM or Helm installations.
### Checklist

- [x] Write tests

### DIFF
--- a/systemtest/src/main/java/io/strimzi/systemtest/resources/ResourceManager.java
+++ b/systemtest/src/main/java/io/strimzi/systemtest/resources/ResourceManager.java
@@ -163,7 +163,7 @@ public class ResourceManager {
     }
 
     @SafeVarargs
-    private final <T extends HasMetadata> void createResource(ExtensionContext testContext, boolean waitReady, T... resources) {
+    private <T extends HasMetadata> void createResource(ExtensionContext testContext, boolean waitReady, T... resources) {
         for (T resource : resources) {
             ResourceType<T> type = findResourceType(resource);
 
--- a/systemtest/src/main/java/io/strimzi/systemtest/resources/ResourceOperation.java
+++ b/systemtest/src/main/java/io/strimzi/systemtest/resources/ResourceOperation.java
@@ -12,7 +12,6 @@ import io.strimzi.api.kafka.model.KafkaMirrorMaker;
 import io.strimzi.api.kafka.model.KafkaMirrorMaker2;
 import io.strimzi.api.kafka.model.StrimziPodSet;
 import io.strimzi.api.kafka.model.balancing.KafkaRebalanceState;
-import io.strimzi.api.kafka.model.nodepool.KafkaNodePool;
 import io.strimzi.systemtest.TestConstants;
 
 import java.time.Duration;
@@ -29,7 +28,6 @@ public class ResourceOperation {
 
         switch (kind) {
             case Kafka.RESOURCE_KIND:
-            case KafkaNodePool.RESOURCE_KIND:
                 timeout = Duration.ofMinutes(14).toMillis();
                 break;
             case KafkaConnect.RESOURCE_KIND:
--- a/systemtest/src/main/java/io/strimzi/systemtest/resources/ResourceOperation.java
+++ b/systemtest/src/main/java/io/strimzi/systemtest/resources/ResourceOperation.java
@@ -12,6 +12,7 @@ import io.strimzi.api.kafka.model.KafkaMirrorMaker;
 import io.strimzi.api.kafka.model.KafkaMirrorMaker2;
 import io.strimzi.api.kafka.model.StrimziPodSet;
 import io.strimzi.api.kafka.model.balancing.KafkaRebalanceState;
+import io.strimzi.api.kafka.model.nodepool.KafkaNodePool;
 import io.strimzi.systemtest.TestConstants;
 
 import java.time.Duration;
@@ -28,6 +29,7 @@ public class ResourceOperation {
 
         switch (kind) {
             case Kafka.RESOURCE_KIND:
+            case KafkaNodePool.RESOURCE_KIND:
                 timeout = Duration.ofMinutes(14).toMillis();
                 break;
             case KafkaConnect.RESOURCE_KIND:
--- a/systemtest/src/main/java/io/strimzi/systemtest/resources/crd/KafkaNodePoolResource.java
+++ b/systemtest/src/main/java/io/strimzi/systemtest/resources/crd/KafkaNodePoolResource.java
@@ -6,11 +6,13 @@ package io.strimzi.systemtest.resources.crd;
 
 import io.fabric8.kubernetes.api.model.LabelSelector;
 import io.fabric8.kubernetes.api.model.LabelSelectorBuilder;
+import io.fabric8.kubernetes.api.model.Pod;
 import io.fabric8.kubernetes.client.dsl.MixedOperation;
 import io.fabric8.kubernetes.client.dsl.Resource;
 import io.strimzi.api.kafka.Crds;
 import io.strimzi.api.kafka.KafkaNodePoolList;
 import io.strimzi.api.kafka.model.Kafka;
+import io.strimzi.api.kafka.model.KafkaResources;
 import io.strimzi.api.kafka.model.nodepool.KafkaNodePool;
 import io.strimzi.api.kafka.model.nodepool.KafkaNodePoolBuilder;
 import io.strimzi.api.kafka.model.nodepool.ProcessRoles;
@@ -20,7 +22,9 @@ import io.strimzi.systemtest.Environment;
 import io.strimzi.systemtest.resources.ResourceManager;
 import io.strimzi.systemtest.resources.ResourceType;
 import io.strimzi.systemtest.templates.crd.KafkaNodePoolTemplates;
+import io.strimzi.systemtest.utils.kafkaUtils.KafkaUtils;
 import io.strimzi.systemtest.utils.kubeUtils.objects.PersistentVolumeClaimUtils;
+import io.strimzi.systemtest.utils.kubeUtils.objects.PodUtils;
 
 import java.util.ArrayList;
 import java.util.HashMap;
@@ -70,7 +74,26 @@ public class KafkaNodePoolResource implements ResourceType<KafkaNodePool> {
 
     @Override
     public boolean waitForReadiness(KafkaNodePool resource) {
-        return resource != null;
+        final LabelSelector kafkaNodePoolSelector = new LabelSelectorBuilder()
+            .withMatchLabels(parseLabelSelector(resource.getStatus().getLabelSelector()))
+            .build();
+
+        // here we wait for readiness of the Kafka pods related to KafkaNodePools
+        PodUtils.waitForPodsReady(resource.getMetadata().getNamespace(), kafkaNodePoolSelector, resource.getSpec().getReplicas(), true);
+
+        return true;
+    }
+
+    private Map<String, String> parseLabelSelector(final String labelSelectorString) {
+        final Map<String, String> labels = new HashMap<>();
+        final String[] labelPairs = labelSelectorString.split(",");
+
+        for (final String label : labelPairs) {
+            String[] keyValue = label.split("=");
+            labels.put(keyValue[0], keyValue[1]);
+        }
+
+        return labels;
     }
 
     public static MixedOperation<KafkaNodePool, KafkaNodePoolList, Resource<KafkaNodePool>> kafkaNodePoolClient() {
--- a/systemtest/src/main/java/io/strimzi/systemtest/resources/crd/KafkaNodePoolResource.java
+++ b/systemtest/src/main/java/io/strimzi/systemtest/resources/crd/KafkaNodePoolResource.java
@@ -6,13 +6,11 @@ package io.strimzi.systemtest.resources.crd;
 
 import io.fabric8.kubernetes.api.model.LabelSelector;
 import io.fabric8.kubernetes.api.model.LabelSelectorBuilder;
-import io.fabric8.kubernetes.api.model.Pod;
 import io.fabric8.kubernetes.client.dsl.MixedOperation;
 import io.fabric8.kubernetes.client.dsl.Resource;
 import io.strimzi.api.kafka.Crds;
 import io.strimzi.api.kafka.KafkaNodePoolList;
 import io.strimzi.api.kafka.model.Kafka;
-import io.strimzi.api.kafka.model.KafkaResources;
 import io.strimzi.api.kafka.model.nodepool.KafkaNodePool;
 import io.strimzi.api.kafka.model.nodepool.KafkaNodePoolBuilder;
 import io.strimzi.api.kafka.model.nodepool.ProcessRoles;
@@ -22,7 +20,6 @@ import io.strimzi.systemtest.Environment;
 import io.strimzi.systemtest.resources.ResourceManager;
 import io.strimzi.systemtest.resources.ResourceType;
 import io.strimzi.systemtest.templates.crd.KafkaNodePoolTemplates;
-import io.strimzi.systemtest.utils.kafkaUtils.KafkaUtils;
 import io.strimzi.systemtest.utils.kubeUtils.objects.PersistentVolumeClaimUtils;
 import io.strimzi.systemtest.utils.kubeUtils.objects.PodUtils;
 
@@ -74,26 +71,12 @@ public class KafkaNodePoolResource implements ResourceType<KafkaNodePool> {
 
     @Override
     public boolean waitForReadiness(KafkaNodePool resource) {
-        final LabelSelector kafkaNodePoolSelector = new LabelSelectorBuilder()
-            .withMatchLabels(parseLabelSelector(resource.getStatus().getLabelSelector()))
-            .build();
+        final LabelSelector kafkaNodePoolSelector = KafkaNodePoolResource.getLabelSelector(resource.getMetadata().getName(), resource.getSpec().getRoles().get(0));
 
         // here we wait for readiness of the Kafka pods related to KafkaNodePools
         PodUtils.waitForPodsReady(resource.getMetadata().getNamespace(), kafkaNodePoolSelector, resource.getSpec().getReplicas(), true);
 
         return true;
-    }
-
-    private Map<String, String> parseLabelSelector(final String labelSelectorString) {
-        final Map<String, String> labels = new HashMap<>();
-        final String[] labelPairs = labelSelectorString.split(",");
-
-        for (final String label : labelPairs) {
-            String[] keyValue = label.split("=");
-            labels.put(keyValue[0], keyValue[1]);
-        }
-
-        return labels;
     }
 
     public static MixedOperation<KafkaNodePool, KafkaNodePoolList, Resource<KafkaNodePool>> kafkaNodePoolClient() {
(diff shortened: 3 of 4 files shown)


systemtest/src/test/java/io/strimzi/systemtest/rollingupdate/KafkaRollerST.java
```diff
@@ -322,6 +326,152 @@ void testKafkaPodPendingDueToRack(ExtensionContext extensionContext) {
...
+        final String nodePoolNameA = testStorage.getKafkaNodePoolName() + "-a";
+        final String nodePoolNameB1 = testStorage.getKafkaNodePoolName() + "-b1";
+        final String nodePoolNameB2 = testStorage.getKafkaNodePoolName() + "-b2";
+        final String nodePoolNameC1 = testStorage.getKafkaNodePoolName() + "-c1";
+        final String nodePoolNameC2 = testStorage.getKafkaNodePoolName() + "-c2";
+
+        final int mixedPoolAReplicas = 1, brokerPoolB1Replicas = 1, brokerPoolB2Replicas = 2, controllerPoolC1Replicas = 2, controllerPoolC2Replicas = 1;
+
+        final Kafka kafka = KafkaTemplates.kafkaPersistent(testStorage.getClusterName(), 1, 1).build();
+        final KafkaNodePool mixedPoolA = KafkaNodePoolTemplates.kafkaBasedNodePoolWithDualRole(nodePoolNameA, kafka, mixedPoolAReplicas).build();
```
henryZrncik, line 361: this part will fail in case we will run this test without `+UseKRaft` FG.
As this would produce Kafka (KafkaNodePool) with role `Controller` , therefore failing on something like: 
```
In a ZooKeeper-based Kafka cluster, the KafkaNodePool role has to be always set only to the 'broker' role   
```

Possible solutions are: 

- use instead `kafkaBasedNodePoolWithFgBasedRole` for this KNP (giving it role broker in any case, and adding controller as well in case kraft is enabled) 
- or make test run only with kraft, 
- or replace `KNP a` with something like   2 KNP `initial broker` and `initial controller`  (with controller being created only if kraft enabled)

henryZrncik (commented):
Looks Good to me, just a single thing to consider.

systemtest/src/main/java/io/strimzi/systemtest/resources/ResourceOperation.java
```diff
@@ -28,6 +29,7 @@ public static long getTimeoutForResourceReadiness(String kind) {
 
         switch (kind) {
             case Kafka.RESOURCE_KIND:
+            case KafkaNodePool.RESOURCE_KIND:
```
im-konge, line 32: We have 14 minutes for Kafka resource because of the ZK, Kafka, and EO pods -> maybe the KafkaNodePool resource can depend on number of replicas that are passed to the resource? Or something like 5 minutes should be sufficient?

systemtest/src/main/java/io/strimzi/systemtest/resources/crd/KafkaNodePoolResource.java
```diff
@@ -118,7 +124,9 @@ public static KafkaNodePool convertKafkaResourceToKafkaNodePool(Kafka resource)
 
     public static LabelSelector getLabelSelector(String clusterName, String poolName, ProcessRoles processRole) {
         Map<String, String> matchLabels = new HashMap<>();
-        matchLabels.put(Labels.STRIMZI_CLUSTER_LABEL, clusterName);
+        if (clusterName != null) {
+            matchLabels.put(Labels.STRIMZI_CLUSTER_LABEL, clusterName);
+        }
```
im-konge, line 129: Just out of curiosity - in which cases it will be null? I mean, the clusterName should be filled or?

systemtest/src/test/java/io/strimzi/systemtest/rollingupdate/KafkaRollerST.java
```diff
@@ -322,6 +326,152 @@ void testKafkaPodPendingDueToRack(ExtensionContext extensionContext) {
...
+     *  5. - Change controller-only configuration and verify that only mixed and controller node pools are affected, while broker nodes remain unchanged.
+     *  6. - Modify broker-only configuration and observe that only mixed-role and broker-role node pools roll, leaving controller-role nodes unaffected.
+     *
+     * @usecase
+     *  - kafka-node-pool-rolling-update
+     *  - kafka-roller
+     *  - kafka-configuration-management
+     */
+    @ParallelNamespaceTest
+    void testKafkaRollerBehaviorUnderVariousScenarios(ExtensionContext extensionContext) {
```
im-konge, line 350: I would name it differently - you can hide a lot under this name. So something around nodepools?

systemtest/src/main/java/io/strimzi/systemtest/resources/ResourceManager.java
```diff
@@ -163,7 +163,7 @@ public final <T extends HasMetadata> void createResourceWithWait(ExtensionContex
     }
 
     @SafeVarargs
-    private final <T extends HasMetadata> void createResource(ExtensionContext testContext, boolean waitReady, T... resources) {
+    private <T extends HasMetadata> void createResource(ExtensionContext testContext, boolean waitReady, T... resources) {
```
im-konge, line 166: Maybe I cannot see it in the code, but why did you changed it?

im-konge (commented):
Few nits, anyway LGTM :) thanks

systemtest/src/test/java/io/strimzi/systemtest/rollingupdate/KafkaRollerST.java
```diff
@@ -322,6 +326,152 @@ void testKafkaPodPendingDueToRack(ExtensionContext extensionContext) {
...
+     *  4. - After creating controller node pools, verify rolling updates across different node pools as they adjust to the new configuration.
+     *     - Check that the controller nodes roll only when their specific pool is modified.
+     *  5. - Change controller-only configuration and verify that only mixed and controller node pools are affected, while broker nodes remain unchanged.
+     *  6. - Modify broker-only configuration and observe that only mixed-role and broker-role node pools roll, leaving controller-role nodes unaffected.
+     *
+     * @usecase
+     *  - kafka-node-pool-rolling-update
+     *  - kafka-roller
+     *  - kafka-configuration-management
+     */
```
scholzj, line 348: Not a strong opinion if all ST SMEs are fine with it. I'm not sure this is optimal ...
* There are several open issues for scaling the controllers (adding a new controller or mixed node pool is scaling controllers)
* Might be worth separating this into two tests => one for dedicated controllers and one for mixed nodes.
* I wonder if separating the testing of adding / removing node pools into separate tests would be worth it as well to create something what is more simple and easier to maintain

systemtest/src/test/java/io/strimzi/systemtest/rollingupdate/KafkaRollerST.java
see-quick: > use instead kafkaBasedNodePoolWithFgBasedRole for this KNP (giving it role broker in any case, and adding controller as well in case kraft is enabled)
I am more favour in of this approach :)

systemtest/src/main/java/io/strimzi/systemtest/resources/crd/KafkaNodePoolResource.java
see-quick: > Just out of curiosity - in which cases it will be null? I mean, the clusterName should be filled or?
Check the next overloaded method :D, 
```java
public static LabelSelector getLabelSelector(final String poolName, final ProcessRoles processRoles) {
        return getLabelSelector(null, poolName, processRoles);
    }
```
Well, clusterName could be retrieved from KNP labels but it is not mandatory to do so. When you see that I am getting this LabelSelector only using the KNP name + role
```
@Override
    public boolean waitForReadiness(KafkaNodePool resource) {
        final LabelSelector kafkaNodePoolSelector = KafkaNodePoolResource.getLabelSelector(resource.getMetadata().getName(), resource.getSpec().getRoles().get(0));

        // here we wait for readiness of the Kafka pods related to KafkaNodePools
        PodUtils.waitForPodsReady(resource.getMetadata().getNamespace(), kafkaNodePoolSelector, resource.getSpec().getReplicas(), true);

        return true;
    }
```

systemtest/src/test/java/io/strimzi/systemtest/rollingupdate/KafkaRollerST.java
see-quick: HMmm what about `testKafkaRollingUpdatesWithMixedBrokerControllerNodePools`?

systemtest/src/test/java/io/strimzi/systemtest/rollingupdate/KafkaRollerST.java
see-quick: Hmmm and what about having:
1. Test case - scale controllers-only nodes 
2. Test case - scale mixed-only nodes
3. Test case - current test case doing both (but removing configuration changes I think to make it simpler)
4. Test case - changing the controller-only and broker-only configuration  (this could be I think one test case)
? :)

im-konge: Yeah that makes sense. I guess it will be more clear what we are testing. At least we will not think every time about what is tested or not based on env variables.

scholzj: Well, yes ... but ...
* Some of them are not working well because of known limitations ... such as 1 and 2
* Rolling updates of mixed nodes are useful as well
* Not sure what exactly you mean with 3 => scaling both mixed and controller nodes together? That seems to make little sense as you would normally use one or the other.

systemtest/src/main/java/io/strimzi/systemtest/resources/ResourceManager.java
see-quick: Well, it is a `private` method, and when one marks a method with `final` then it says that it could not be implemented in the inheriting classes (but it's private so there is no need to use the `final` keyword here). Anyway, I think I forgot to change it during my refactor :/

im-konge: Ah okay, yeah. I thought that you need it somewhere or something.

systemtest/src/test/java/io/strimzi/systemtest/rollingupdate/KafkaRollerST.java
```diff
@@ -322,6 +328,107 @@ void testKafkaPodPendingDueToRack(ExtensionContext extensionContext) {
...
     }
 
+    /**
+     * @description This test case verifies the rolling update behavior of Kafka controller nodes under specific conditions.
+     * It focuses on ensuring that changes in Kafka configuration affect only the intended node pools, particularly
+     * the controller nodes, while leaving others like broker nodes unaffected.
+     *
+     * @steps
+     *  1. - Assume that KRaft mode is enabled and the installation method is bundle only.
+     *  2. - Initialize the test environment and create a Kafka instance with a single replica.
```
scholzj, line 338: Just a nit. But I think this comment is confusing. Given you use node pools and KRaft, the replicas in the Kafka CR is ignored. So You do not really create a Kafka instance with single replica.

systemtest/src/test/java/io/strimzi/systemtest/rollingupdate/KafkaRollerST.java
```diff
@@ -322,6 +328,107 @@ void testKafkaPodPendingDueToRack(ExtensionContext extensionContext) {
...
+     *     - Ensure that only controller nodes undergo a rolling update, while broker nodes remain unaffected.
+     *  6. - Verify the rolling update of controller nodes by comparing the snapshots taken before and after the configuration change.
+     *
+     * @usecase
+     *  - kafka-controller-node-rolling-update
+     *  - kafka-configuration-change-impact
+     *  - kafka-node-pool-management
+     */
+    @ParallelNamespaceTest
+    void testKafkaRollingUpdatesOfControllerNodes(final ExtensionContext extensionContext) {
```
scholzj, line 351: Just an idea ... this is good. But could we add a second rolling update related to some change in the controller node pool (e.g. some affinity, pod label, health check or something like that not related to Kafka configuration)? That will help to trigger also slightly different code path.

systemtest/src/test/java/io/strimzi/systemtest/rollingupdate/KafkaRollerST.java
```diff
@@ -322,6 +328,107 @@ void testKafkaPodPendingDueToRack(ExtensionContext extensionContext) {
...
+    }
+
+    /**
+     * @description This test case assesses the rolling update behavior of Kafka nodes in mixed-role configurations.
+     * The main focus is to verify that configuration changes impacting controller roles result in rolling updates
+     * of nodes serving mixed roles, while ensuring compatibility with KRaft mode and excluding OLM or Helm installations.
+     *
+     * @steps
+     *  1. - Ensure that the environment is running in KRaft mode and is neither an OLM nor Helm installation (only Bundle/YAML!).
+     *  2. - Initialize the test environment by setting up a Kafka instance with a single replica.
```
scholzj, line 394: Same as above, I would probably write it a bit differently.

scholzj (commented):
Some nits and one possible suggestion ... LGTM otherwise.

systemtest/src/test/java/io/strimzi/systemtest/rollingupdate/KafkaRollerST.java
see-quick: Yeah, I see okay removing.

systemtest/src/test/java/io/strimzi/systemtest/rollingupdate/KafkaRollerST.java
```diff
@@ -322,6 +328,107 @@ void testKafkaPodPendingDueToRack(ExtensionContext extensionContext) {
         KafkaUtils.waitForKafkaDeletion(namespaceName, clusterName);
     }
 
+    /**
+     * @description This test case verifies the rolling update behavior of Kafka controller nodes under specific conditions.
+     * It focuses on ensuring that changes in Kafka configuration affect only the intended node pools, particularly
+     * the controller nodes, while leaving others like broker nodes unaffected.
+     *
+     * @steps
+     *  1. - Assume that KRaft mode is enabled and the installation method is bundle only.
```
henryZrncik, line 341: this is a matter of preference i guess but can we mention why we assume this test can run only in KRaft.

systemtest/src/test/java/io/strimzi/systemtest/rollingupdate/KafkaRollerST.java
```diff
@@ -322,6 +328,107 @@ void testKafkaPodPendingDueToRack(ExtensionContext extensionContext) {
...
+     *  2. - Initialize the test environment and create a Kafka instance with a single replica.
+     *  3. - Create and deploy a Kafka node pool with broker roles (brokerPool) and another with controller roles (controllerPool), each with 3 replicas.
```
henryZrncik:
```suggestion
     *  3. - Create and deploy a Kafka node pool with broker role (brokerPool) and another with controller role (controllerPool), each with 3 replicas.
```

systemtest/src/test/java/io/strimzi/systemtest/rollingupdate/KafkaRollerST.java
see-quick: Okay, I extend the test case with another RollinUpdate with PodAffinity :) Thanks for this comment.

scholzj (approved):
Thanks. LGTM assuming it passes in the CIs.

im-konge (approved):
LGTM, thanks :)

systemtest/src/test/java/io/strimzi/systemtest/rollingupdate/KafkaRollerST.java
henryZrncik: as you all mentioned, it seems to me reasonable to have following tests: 
- rolling and config changes related to controllers when there is controller only role (implemented here)
- rolling and config changes related to controllers when there are mixed roles (implemented here) 
- adding removing node pools (right now broker only roles) (implemented here https://github.com/strimzi/strimzi-kafka-operator/pull/9428)  

mentioned things (test improvements) like scaling controllers (adding controller) node pool or additional checks for correct configuration propagation can be added later accordingly.

systemtest/src/test/java/io/strimzi/systemtest/rollingupdate/KafkaRollerST.java
```diff
@@ -322,6 +332,143 @@ void testKafkaPodPendingDueToRack(ExtensionContext extensionContext) {
...
+
+        resourceManager.createResourceWithoutWait(extensionContext, brokerPool, controllerPool, kafka);
+
+        PodUtils.waitForPodsReady(testStorage.getNamespaceName(), brokerPoolSelector, brokerPoolReplicas, true);
+        PodUtils.waitForPodsReady(testStorage.getNamespaceName(), controllerPoolSelector, controllerPoolReplicas, true);
+
+        Map<String, String> brokerPoolPodsSnapshot = PodUtils.podSnapshot(testStorage.getNamespaceName(), brokerPoolSelector);
+        Map<String, String> controllerPoolPodsSnapshot = PodUtils.podSnapshot(testStorage.getNamespaceName(), controllerPoolSelector);
+
+        // change Controller-only configuration inside shared Kafka configuration between KafkaNodePools and see that only mixed and controller pods rolls
```
henryZrncik, line 382:
```suggestion
        // change Controller-only configuration inside shared Kafka configuration between KafkaNodePools and see that only controller pods rolls
```

systemtest/src/test/java/io/strimzi/systemtest/rollingupdate/KafkaRollerST.java
```diff
@@ -322,6 +332,143 @@ void testKafkaPodPendingDueToRack(ExtensionContext extensionContext) {
...
+        final Kafka kafka = KafkaTemplates.kafkaPersistent(testStorage.getClusterName(), 1, 1).build();
+        final KafkaNodePool mixedPool = KafkaNodePoolTemplates.kafkaBasedNodePoolWithDualRole(mixedPoolName, kafka, mixedPoolReplicas).build();
+
+        resourceManager.createResourceWithWait(extensionContext, mixedPool, kafka);
+
+        final LabelSelector mixedPoolSelector = KafkaNodePoolResource.getLabelSelector(testStorage.getClusterName(), mixedPoolName, ProcessRoles.CONTROLLER);
+
+        Map<String, String> mixedPoolPodsSnapshot = PodUtils.podSnapshot(testStorage.getNamespaceName(), mixedPoolSelector);
+
+        // change Controller-only configuration inside shared Kafka configuration between KafkaNodePools and see that only mixed and controller pods rolls
```
henryZrncik, line 464:
```suggestion
        // change Controller-only configuration inside shared Kafka configuration between KafkaNodePools and see that only mixed pods rolls
```

henryZrncik (approved):
Great, thanks you !

systemtest/src/test/java/io/strimzi/systemtest/rollingupdate/KafkaRollerST.java
tinaselenge, line 464:
```suggestion
        // change Controller-only configuration inside shared Kafka configuration between KafkaNodePools and see that all mixed pods rolls
```

tinaselenge (commented):
just a couple of nits and LGTM.

tinaselenge (approved):

systemtest/src/test/java/io/strimzi/systemtest/rollingupdate/KafkaRollerST.java
tinaselenge: also just curious, do we not want to make config change that only applies to broker only pods as well and verify that controller only pods do not roll?

henryZrncik (approved):

Issue #9436, opened 2023-12-05
[KRaft] Prevent mixed nodes changing to controller-only nodes if they still have assigned partition replicas


scholzj:
When a user changes a node pool roles from mixed to controller-only without reassigning all partition replicas assigned to them, it might break the availability of these topics and block rolling updates. We should check this and prevent the change similarly to how we do it when scaling down broker nodes.

scholzj:
Discussed in the community call on 14.12.: Should be implemented as a safeguard.

Issue #9434, opened 2023-12-05
[KRaft] `KafkaRoller` is strugling to  transition controller-only nodes to mixed nodes


scholzj:
When having a controller-only node pool and trying to add a broker role to it, the operator / `KafkaRoller` seem to strugle to roll the nodes to change the configuration. It seems they try to treat the nodes as mixed nodes already, but fail to connect to them on the port 9091 as they are still controllers only.

It looks like the eventually roll them after the timeout expires. But it is not clear if they properly ensure the quorum availability during this. If nothing else, we have to check that this does not break the quorum -> but if possible, we should try to improve the process.

```
2023-12-05 15:05:47 INFO  KafkaAssemblyOperator:844 - Reconciliation #735(watch) Kafka(myproject/my-cluster): KafkaNodePool controllers in namespace myproject was MODIFIED
2023-12-05 15:05:47 INFO  AbstractOperator:265 - Reconciliation #735(watch) Kafka(myproject/my-cluster): Kafka my-cluster will be checked for creation or modification
2023-12-05 15:05:47 INFO  CrdOperator:122 - Reconciliation #735(watch) Kafka(myproject/my-cluster): Status of KafkaNodePool controllers in namespace myproject has been updated
2023-12-05 15:05:47 WARN  NetworkClient:814 - [AdminClient clientId=adminclient-226] Connection to node -7 (my-cluster-controllers-0.my-cluster-kafka-brokers.myproject.svc.cluster.local/172.16.14.234:9091) could not be established. Broker may not be available.
2023-12-05 15:06:17 INFO  KafkaRoller:396 - Reconciliation #735(watch) Kafka(myproject/my-cluster): Will temporarily skip verifying pod my-cluster-controllers-0/0 is up-to-date due to ForceableProblem: Error getting broker config, retrying after at least 250ms
2023-12-05 15:06:47 INFO  AbstractOperator:400 - Reconciliation #735(watch) Kafka(myproject/my-cluster): Reconciliation is in progress
2023-12-05 15:06:47 INFO  KafkaRoller:396 - Reconciliation #735(watch) Kafka(myproject/my-cluster): Will temporarily skip verifying pod my-cluster-controllers-1/1 is up-to-date due to ForceableProblem: Error getting broker config, retrying after at least 250ms
[...]

PR #9458, opened 2023-12-12
Add current roles to the node pool status

scholzj:
### Type of change

- Enhancement / new feature

### Description

This PR adds a new field `roles` to the `status of the `KafkaNodePool` resource which is used to track the roles assigned to given node pool. This field is also set as part of the reconciliation. The information it self is not used for anything in this PR - but it will be used in the future to address #9436 an possibly also #9434. It is done in a separate Pr to get it into the codebase before the 0.39 release to make sure we avoid any issues with upgrade later.

### Checklist

- [x] Write tests

### DIFF
--- a/api/src/main/java/io/strimzi/api/kafka/model/nodepool/KafkaNodePoolStatus.java
+++ b/api/src/main/java/io/strimzi/api/kafka/model/nodepool/KafkaNodePoolStatus.java
@@ -23,7 +23,7 @@ import java.util.List;
         builderPackage = Constants.FABRIC8_KUBERNETES_API
 )
 @JsonInclude(JsonInclude.Include.NON_NULL)
-@JsonPropertyOrder({ "conditions", "observedGeneration", "nodeIds", "clusterId", "replicas", "labelSelector" })
+@JsonPropertyOrder({ "conditions", "observedGeneration", "nodeIds", "clusterId", "roles", "replicas", "labelSelector" })
 @EqualsAndHashCode(callSuper = true)
 @ToString(callSuper = true)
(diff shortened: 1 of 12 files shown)


api/src/main/java/io/strimzi/api/kafka/model/nodepool/KafkaNodePoolStatus.java
```diff
@@ -54,6 +55,15 @@ public void setClusterId(String clusterId) {
         this.clusterId = clusterId;
     }
 
+    @Description("The roles currently assigned to this pool.")
```
ppatierno, line 58: I would be clearer about which roles we are referring to here. Maybe something like this?

```suggestion
    @Description("The KRaft related process roles currently assigned to this pool.")
```

ppatierno (approved):
LGTM. Just a nit.

api/src/main/java/io/strimzi/api/kafka/model/nodepool/KafkaNodePoolStatus.java
scholzj: Well, they are not limited to KRaft. All ZooKeeper-based brokers have the `broker` role.

scholzj: (Plus the roles are required field even for ZooKeeper-based cluster).

ppatierno: Right, so ignore my comment ;-)

PR #9460, opened 2023-12-13
docs(node pools): updates to node pools content to reflect move to beta

PaulRMellor:
**Documentation**

The `KafkaNodePool` feature gate has now reached a beta level of maturity, which means that it is enabled by default. This PR updates the Strimzi docs in which node pools are mentioned to reflect this promotion to beta.

### Checklist

- [ ] Write tests
- [x] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/documentation/modules/configuring/con-config-node-pools.adoc
+++ b/documentation/modules/configuring/con-config-node-pools.adoc
@@ -3,12 +3,10 @@
 // assembly-config.adoc
 
 [id='config-node-pools-{context}']
-= (Preview) Configuring node pools
+= Configuring node pools
 
 [role="_abstract"]
 Update the `spec` properties of the `KafkaNodePool` custom resource to configure a node pool deployment.
-
-NOTE: The node pools feature is available as a preview. Node pools are not enabled by default, so you must xref:ref-operator-kafka-node-pools-feature-gate-{context}[enable the `KafkaNodePools` feature gate] before using them. 
 
 A node pool refers to a distinct group of Kafka nodes within a Kafka cluster.
 Each pool has its own unique configuration, which includes mandatory settings for the number of replicas, roles, and storage allocation.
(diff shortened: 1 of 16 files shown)


documentation/modules/configuring/con-config-node-pools.adoc
```diff
@@ -37,73 +35,72 @@ IMPORTANT: **KRaft mode is not ready for production in Apache Kafka or in Strimz
...
+  roles: # <1>
+    - controller
+    - broker
+  storage:
     type: jbod
     volumes:
       - id: 0
         type: persistent-claim
-        size: 100Gi
+        size: 20Gi
```
scholzj, line 59: Why change the volume size? I do not see any value and 20Gi is in general quite small. I also wonder if we should keep the callouts for the other parts.

documentation/modules/configuring/con-config-node-pools.adoc
```diff
@@ -37,73 +35,72 @@ IMPORTANT: **KRaft mode is not ready for production in Apache Kafka or in Strimz
 
 For a deeper understanding of the node pool configuration options, refer to the link:{BookURLConfiguring}[Strimzi Custom Resource API Reference^].
 
-NOTE: While the `KafkaNodePools` feature gate that enables node pools is in alpha phase, replica and storage configuration properties in the `KafkaNodePool` resource must also be present in the `Kafka` resource. The configuration in the `Kafka` resource is ignored when node pools are used. Similarly, ZooKeeper configuration properties must also be present in the `Kafka` resource when using KRaft mode. These properties are also ignored.
+NOTE: While the `KafkaNodePools` feature gate that enables node pools is in beta phase, replica and storage configuration properties in the `KafkaNodePool` resource must also be present in the `Kafka` resource. The configuration in the `Kafka` resource is ignored when node pools are used. Similarly, ZooKeeper configuration properties must also be present in the `Kafka` resource when using KRaft mode. These properties are also ignored.
```
scholzj, line 38: I think this will likely change before node pools leave beta. So maybe we can reword it something like this?

```suggestion
NOTE: Currently, replica and storage configuration properties in the `KafkaNodePool` resource must also be present in the `Kafka` resource. The configuration in the `Kafka` resource is ignored when node pools are used. Similarly, ZooKeeper configuration properties must also be present in the `Kafka` resource when using KRaft mode. These properties are also ignored.
```

documentation/modules/configuring/proc-migrating-clusters-node-pools.adoc
```diff
@@ -3,13 +3,13 @@
...
 [id='proc-migrating-clusters-node-pools-{context}']
-= (Preview) Migrating existing Kafka clusters to use Kafka node pools
+= Migrating existing Kafka clusters to use Kafka node pools
 
 [role="_abstract"]
 This procedure describes how to migrate existing Kafka clusters to use Kafka node pools.
 After you have updated the Kafka cluster, you can use the node pools to manage the configuration of nodes within each pool.
 
-NOTE: While the `KafkaNodePools` feature gate that enables node pools is in alpha phase, replica and storage configuration in the `KafkaNodePool` resource must also be present in the `Kafka` resource. The configuration is ignored when node pools are being used. 
+NOTE: While the `KafkaNodePools` feature gate that enables node pools is in beta phase, replica and storage configuration in the `KafkaNodePool` resource must also be present in the `Kafka` resource. The configuration is ignored when node pools are being used.
```
scholzj, line 12: Same as above ... maybe this would be better?

```suggestion
NOTE: Currently, replica and storage configuration in the `KafkaNodePool` resource must also be present in the `Kafka` resource. The configuration is ignored when node pools are being used. 
```

documentation/modules/deploying/proc-deploy-kafka-cluster.adoc
```diff
@@ -65,15 +65,33 @@ spec:
...
+Or by editing the Cluster Operator `Deployment` and updating the `STRIMZI_FEATURE_GATES` environment variable:
++
+[source,yaml]
+----
+env
+  - name: STRIMZI_FEATURE_GATES
+    value: -KafkaNodePools
+----
++
+This updates the Cluster Operator.
```
scholzj, line 84: This is not needed and should be removed.

documentation/modules/deploying/proc-deploy-kafka-node-pools.adoc
```diff
@@ -52,23 +49,28 @@ env
...
 +
 [source,shell,subs="attributes+"]
-kubectl apply -f examples/kafka/nodepools/kafka-with-dual-role-kraft-nodes.yaml
+kubectl apply -f examples/kafka/nodepools/kafka-with-kraft.yaml
 
-* To deploy a Kafka cluster in KRaft mode with separate node pools for broker and controller nodes: 
+* To deploy an ephemeral Kafka cluster in KRaft mode with separate node pools for broker and controller nodes: 
 +
 [source,shell,subs="attributes+"]
 kubectl apply -f examples/kafka/nodepools/kafka-with-kraft.yaml
```
scholzj, line 67:
```suggestion
kubectl apply -f examples/kafka/nodepools/kafka-with-kraft-ephemeral.yaml
```

ppatierno (approved):
LGTM. Apart Jakub's comments.

documentation/modules/configuring/con-config-node-pools.adoc
PaulRMellor: That's peculiar. I just changed the order of the examples so that KRaft came first.  I didn't make any changes to the examples themselves. I'll revert this.

scholzj: Ahh, I did not realize that. But especially for the callouts, it makes sense to have them in the first example rather than the second. It looks good now.

scholzj (approved):

Issue #5630, opened 2021-09-29
Improve the certificate handling


sknot-rh:
Currently we've failed to really encapsulate and abstract our certificate handling. What I mean by that is everywhere we pass around a `Secret` which we _happen to know_ contains a key store (broker certificate and key), or truststore (ca certificate). But that requires that the consumer of that `Secret` knows what _type_ of secret is it, and how to decode the values within it into something it can use. 

I think there is value is trying to better encapsulate these concepts, so rather than passing around Secrets we pass around something higher level. For example a `AdminClientSupplier` might accept some Strimzi `Truststore` class and a Strimzi `Keystore` class. Providing Resource Operators for those things (`Truststore` and `Keystore`) would be a logical thing to do. This would provide a path towards being able to abstract where certain credentials are stored (e.g. option to store keys in Vault), and should make the KAO slightly simpler to understand, because there are fewer generic Secrets floating around.

At least that was my motivation for suggesting this, and I'd be interested to know whether people think that's a desirable goal or not (and if they have specific reasons to think it wouldn't work out).

_Originally posted by @tombentley in https://github.com/strimzi/strimzi-kafka-operator/pull/5549#discussion_r714675992_

SDBrett:
Am I correct that you're proposing a singular CR which contains relevant credentials for a Kafka cluster?

I mocked up two manifests to help illustrate my understanding, though I may be way off.


```
apiVersion: v1
kind: KafkaCredential
metadata:
  name: cluster-01-creds
spec:
  truststore:
    dataSource: literal
    truststore: encoded cert
  clientCredentials:
    type: certificate
    certificate:
      dataSource: vault
      secretName: secret/clientCert

---
apiVersion: v1
kind: KafkaCredential
metadata:
  name: cluster-02-creds
spec:
  truststore:
    dataSource: literal
    truststore: encoded cert
  clientCredentials:
    type: plain
    plain:
      dataSource: literal
      userName: user
      password: test123
```

scholzj:
Triaged on 28. 4. 2022: This still makes sense. It is about internal Java APIs in the Strimzi code only and how the certificates are handled. It is not about any new resources or outside representation.

scholzj:
> Hopefully this is enough detail, at least for you to be able to point out if/how it might not work @scholzj.

I think this is good. It should give everyone better idea what it means. Let's get back to it next triage call and decide how to label it etc. Thanks for updating this.

katheris:
[...]

PR #9463, opened 2023-12-14
Refactor cert handling

katheris:
### Type of change
- Refactoring

### Description

Tidy up certificate code:
 - Remove duplicated logic to create Map for Secret data
 - Remove certificate logic from ModelUtils class

This PR contributes to https://github.com/strimzi/strimzi-kafka-operator/issues/5630. I have created this as a separate PR to make it easier to review, this change reduces the number of places that are handling certificates to make it easier to remove the dependence on Kubernetes Secrets in a future PR.

### Checklist

- [x] Write tests
- [ ] Update documentation
- [ ] Update CHANGELOG.md


certificate-manager/src/main/java/io/strimzi/certs/SecretCertProvider.java
```diff
@@ -25,6 +25,117 @@
  */
 public class SecretCertProvider {
 
+    /**
+     * A certificate entry in a ConfigMap. Each entry contains an entry name and data.
```
scholzj, line 29:
```suggestion
     * A certificate entry in a Secret. Each entry contains an entry name and data.
```

scholzj (commented):
The `certificate-manager` module has nothing to do with Secrets. It is only responsible for issuing the certificates. So I'm not sure if we should put any logic related to the Secrets into it. I'm surprised it even builds, because it IMHO should not have any dependency on Fabric8 and should therefore not know the `Secret` class.

katheris:
> The `certificate-manager` module has nothing to do with Secrets. It is only responsible for issuing the certificates. So I'm not sure if we should put any logic related to the Secrets into it. I'm surprised it even builds, because it IMHO should not have any dependency on Fabric8 and should therefore not know the `Secret` class.

The SecretCertProvider class was already in the certificate-manager module and contained methods to create Secrets for certificates so I moved the methods there rather than creating another separate class that deals with certificate Secrets.

It looks like the addSecret methods in that class are currently used from the Ca class which is in operator-common, so the other methods in SecretCertProvider that are there already can't be moved into cluster-operator without a refactor of the Ca class. However I could move the new methods that I am adding to SecretCertProvider to a new class in cluster-operator?

In a follow-up PR I think it would be good to remove any mention of Secrets from the Ca, but I am hesitant to do that as part of this PR as the bigger the refactor the harder it is to review. I did wonder though if as part of that refactor the best outcome would be to have the cluster-operator and user-operator modules have their own separate logic for storing certificates in Secrets, rather than having any of that in operator-common. But I figured that change probably needed an agreed proposal first, what do you think?

scholzj:
> > The `certificate-manager` module has nothing to do with Secrets. It is only responsible for issuing the certificates. So I'm not sure if we should put any logic related to the Secrets into it. I'm surprised it even builds, because it IMHO should not have any dependency on Fabric8 and should therefore not know the `Secret` class.
> 
> The SecretCertProvider class was already in the certificate-manager module and contained methods to create Secrets for certificates so I moved the methods there rather than creating another separate class that deals with certificate Secrets.
> 
> It looks like the addSecret methods in that class are currently used from the Ca class which is in operator-common, so the other methods in SecretCertProvider that are there already can't be moved into cluster-operator without a refactor of the Ca class. However I could move the new methods that I am adding to SecretCertProvider to a new class in cluster-operator?
> 
> In a follow-up PR I think it would be good to remove any mention of Secrets from the Ca, but I am hesitant to do that as part of this PR as the bigger the refactor the harder it is to review. I did wonder though if as part of that refactor the best outcome would be to have the cluster-operator and user-operator modules have their own separate logic for storing certificates in Secrets, rather than having any of that in operator-common. But I figured that change probably needed an agreed proposal first, what do you think?

That is a good point. But maybe the best would be to move the `SecretCertProvider` to `operator-common` rather than add to it? (It also seems to have its own issues as half of the methods in that class are used only by its test, so they probably should not be there at all) WDYT?

scholzj:
It might be also worth thinking about whether all of these things are really needed. `SecretEntry` seems like a weird thing used only in some parts of the project that can be easily replaced by a set of constants.

katheris:
> That is a good point. But maybe the best would be to move the `SecretCertProvider` to `operator-common` rather than add to it? (It also seems to have its own issues as half of the methods in that class are used only by its test, so they probably should not be there at all) WDYT?

That's a good idea. Yeah, I did also notice that. It seemed like this whole module was added in a commit to introduce a separate certificate module, but the commit message implies that there might be follow up changes, but I didn't see any. @ppatierno do you know if there were other changes expected that would use these methods that are currently only used in test? If not then I think we should definitely clean up the methods that aren't being used.

katheris:
> It might be also worth thinking about whether all of these things are really needed. `SecretEntry` seems like a weird thing used only in some parts of the project that can be easily replaced by a set of constants.

My hope is to get to a point that we only have .p12 etc extensions in a single place, rather than constructed differently everywhere, but when I started looking through the changes needed it made sense to do this as a separate PR as there are so many different uses of it.

cluster-operator/src/test/java/io/strimzi/operator/cluster/operator/assembly/CertificateRenewalTest.java
scholzj, line 1: Do all the tests you modified here really test the ModelUtils? Or do they now test the `CertUtils`?

scholzj (commented):
I left bunch of comments. But honestly, I still don't get how you decided what should be moved to SecretCertProvider and what to CertUtils. They seem to originate from `cluster-operator` (at least those I checked - apart from `SecretEntry`) as well. So why not move them to `CertUtils` and keep them in the same module instead of moving it into module that is shared more widely into a class that does not seem to fit the module anyway?

katheris:
> I left bunch of comments. But honestly, I still don't get how you decided what should be moved to SecretCertProvider and what to CertUtils. They seem to originate from `cluster-operator` (at least those I checked - apart from `SecretEntry`) as well. So why not move them to `CertUtils` and keep them in the same module instead of moving it into module that is shared more widely into a class that does not seem to fit the module anyway?

I was trying to keep a separation between the classes that need to understand the structure of how certificates are stored in Kube Secrets and those that don't. However, I think you're right that sharing those methods more widely as a result is not the right choice. I think for this PR keeping them in `CertUtils` makes sense, and then I can hopefully create the better separation in a follow-up. I'll make some changes

scholzj (approved):

scholzj:
Are the CI failures related to the PR? Especially these ones?

```
[ERROR] Errors: 
[ERROR] io.strimzi.operator.cluster.operator.assembly.CertificateRenewalTest.testRenewalOfDeploymentCertificatesDelayedRenewal
[ERROR]   Run 1: CertificateRenewalTest.testRenewalOfDeploymentCertificatesDelayedRenewal:109 » NullPointer Cannot invoke "java.util.Map$Entry.getKey()" because "entries[0]" is null
[ERROR]   Run 2: CertificateRenewalTest.testRenewalOfDeploymentCertificatesDelayedRenewal:109 » NullPointer Cannot invoke "java.util.Map$Entry.getKey()" because "entries[0]" is null
[ERROR]   Run 3: CertificateRenewalTest.testRenewalOfDeploymentCertificatesDelayedRenewal:109 » NullPointer Cannot invoke "java.util.Map$Entry.getKey()" because "entries[0]" is null
[ERROR]   Run 4: CertificateRenewalTest.testRenewalOfDeploymentCertificatesDelayedRenewal:109 » NullPointer Cannot invoke "java.util.Map$Entry.getKey()" because "entries[0]" is null
[ERROR]   Run 5: CertificateRenewalTest.testRenewalOfDeploymentCertificatesDelayedRenewal:109 » NullPointer Cannot invoke "java.util.Map$Entry.getKey()" because "entries[0]" is null
[ERROR]   Run 6: CertificateRenewalTest.testRenewalOfDeploymentCertificatesDelayedRenewal:109 » NullPointer Cannot invoke "java.util.Map$Entry.getKey()" because "entries[0]" is null
[INFO] 
[ERROR] io.strimzi.operator.cluster.operator.assembly.CertificateRenewalTest.testRenewalOfDeploymentCertificatesDelayedRenewalOutsideOfMaintenanceWindow
[ERROR]   Run 1: CertificateRenewalTest.testRenewalOfDeploymentCertificatesDelayedRenewalOutsideOfMaintenanceWindow:142 » NullPointer Cannot invoke "java.util.Map$Entry.getKey()" because "entries[0]" is null
[ERROR]   Run 2: CertificateRenewalTest.testRenewalOfDeploymentCertificatesDelayedRenewalOutsideOfMaintenanceWindow:142 » NullPointer Cannot invoke "java.util.Map$Entry.getKey()" because "entries[0]" is null
[ERROR]   Run 3: CertificateRenewalTest.testRenewalOfDeploymentCertificatesDelayedRenewalOutsideOfMaintenanceWindow:142 » NullPointer Cannot invoke "java.util.Map$Entry.getKey()" because "entries[0]" is null
[ERROR]   Run 4: CertificateRenewalTest.testRenewalOfDeploymentCertificatesDelayedRenewalOutsideOfMaintenanceWindow:142 » NullPointer Cannot invoke "java.util.Map$Entry.getKey()" because "entries[0]" is null
[ERROR]   Run 5: CertificateRenewalTest.testRenewalOfDeploymentCertificatesDelayedRenewalOutsideOfMaintenanceWindow:142 » NullPointer Cannot invoke "java.util.Map$Entry.getKey()" because "entries[0]" is null
[ERROR]   Run 6: CertificateRenewalTest.testRenewalOfDeploymentCertificatesDelayedRenewalOutsideOfMaintenanceWindow:142 » NullPointer Cannot invoke "java.util.Map$Entry.getKey()" because "entries[0]" is null
[INFO] 
[ERROR] io.strimzi.operator.cluster.operator.assembly.CertificateRenewalTest.testRenewalOfDeploymentCertificatesWithNullSecret
[ERROR]   Run 1: CertificateRenewalTest.testRenewalOfDeploymentCertificatesWithNullSecret:43 » NullPointer Cannot invoke "java.util.Map$Entry.getKey()" because "entries[0]" is null
[ERROR]   Run 2: CertificateRenewalTest.testRenewalOfDeploymentCertificatesWithNullSecret:43 » NullPointer Cannot invoke "java.util.Map$Entry.getKey()" because "entries[0]" is null
[ERROR]   Run 3: CertificateRenewalTest.testRenewalOfDeploymentCertificatesWithNullSecret:43 » NullPointer Cannot invoke "java.util.Map$Entry.getKey()" because "entries[0]" is null
[ERROR]   Run 4: CertificateRenewalTest.testRenewalOfDeploymentCertificatesWithNullSecret:43 » NullPointer Cannot invoke "java.util.Map$Entry.getKey()" because "entries[0]" is null
[ERROR]   Run 5: CertificateRenewalTest.testRenewalOfDeploymentCertificatesWithNullSecret:43 » NullPointer Cannot invoke "java.util.Map$Entry.getKey()" because "entries[0]" is null
[ERROR]   Run 6: CertificateRenewalTest.testRenewalOfDeploymentCertificatesWithNullSecret:43 » NullPointer Cannot invoke "java.util.Map$Entry.getKey()" because "entries[0]" is null
[INFO] 
[ERROR] io.strimzi.operator.cluster.operator.assembly.CertificateRenewalTest.testRenewalOfDeploymentCertificatesWithRenewingCa
[ERROR]   Run 1: CertificateRenewalTest.testRenewalOfDeploymentCertificatesWithRenewingCa:76 » NullPointer Cannot invoke "java.util.Map$Entry.getKey()" because "entries[0]" is null
[ERROR]   Run 2: CertificateRenewalTest.testRenewalOfDeploymentCertificatesWithRenewingCa:76 » NullPointer Cannot invoke "java.util.Map$Entry.getKey()" because "entries[0]" is null
[ERROR]   Run 3: CertificateRenewalTest.testRenewalOfDeploymentCertificatesWithRenewingCa:76 » NullPointer Cannot invoke "java.util.Map$Entry.getKey()" because "entries[0]" is null
[ERROR]   Run 4: CertificateRenewalTest.testRenewalOfDeploymentCertificatesWithRenewingCa:76 » NullPointer Cannot invoke "java.util.Map$Entry.getKey()" because "entries[0]" is null
[ERROR]   Run 5: CertificateRenewalTest.testRenewalOfDeploymentCertificatesWithRenewingCa:76 » NullPointer Cannot invoke "java.util.Map$Entry.getKey()" because "entries[0]" is null
[ERROR]   Run 6: CertificateRenewalTest.testRenewalOfDeploymentCertificatesWithRenewingCa:76 » NullPointer Cannot invoke "java.util.Map$Entry.getKey()" because "entries[0]" is null
```

katheris:
> Are the CI failures related to the PR? Especially these ones?

Yeah, looks like changing the  Collections.SingletonMap to Map.of caused some failures related to mock methods, I've added the correct mocking to the tests.

operator-common/src/main/java/io/strimzi/operator/common/model/Ca.java
```diff
@@ -778,6 +761,13 @@ public int certGeneration() {
         return INIT_GENERATION;
     }
 
+    /**
+     * @return the generation of the current CA certificate as an annotation
+     */
+    public Map<String, String> caCertGenerationFullAnnotation() {
```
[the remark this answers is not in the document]
katheris: I've updated to change it to Map.Entry. At the moment all of the clustered components (Cruise Control, ZooKeeper, Kafka) call ModelUtils directly to create certificate Secrets. The other components with a single pod call buildTrustedCertificateSecret which is in CertUtils. So that's why there are call from the different places.

katheris:
@scholzj I've added an extra commit that makes better use of the SecretEntry enum

ppatierno:
> @ppatierno do you know if there were other changes expected that would use these methods that are currently only used in test? If not then I think we should definitely clean up the methods that aren't being used.

It's hard to remember and tell. Maybe they were used before then not anymore and being left overs in the tests. I think that after so long time having this cert manager implementation working, we can clean up by removing these methods.

operator-common/src/main/java/io/strimzi/operator/common/model/Ca.java
```diff
@@ -89,10 +89,21 @@ public enum SecretEntry {
         }
 
         /**
-         * @return The suffix of the entry name in the Secret
+         *
+         * @param prefix to use for the certificate Secret key
+         * @return a certificate Secret key with the correct suffix
```
ppatierno, line 94: I thinks this method deserves more explanation. It's returning the key by concatenating prefix + suffix it's not just about returning with "correct suffix", or?

ppatierno (approved):
LGTM. Thanks for the refactoring!

operator-common/src/main/java/io/strimzi/operator/common/model/Ca.java
katheris: Here by "correct suffix" I mean the correct suffix for the type of data we are storing, e.g. cert vs key vs p12 etc.

What about this:
```suggestion
         /**
         * Build the Kubernetes Secret key to use with this type of SecretEntry.
         *
         * @param prefix to use for the certificate Secret key
         * @return a certificate Secret key with the provided prefix and the suffix of this type of SecretEntry
```

ppatierno: Yes ... I think it's clearer!

katheris: I've updated it

Issue #9465, opened 2023-12-15
[UTO] needs to use log levels in a better way


scholzj:
When using Unidirectional Topic Operator (UTO), the way the log levels are used seems to be very strange. At an INFO level, it logs something like this while running for 2 hours with 11 topics:

```
Preparing trust store certificates for internal communication
Adding /etc/tls-sidecar/cluster-ca-certs/ca.crt to truststore /tmp/topic-operator/replication.truststore.p12 with alias ca
Certificate was added to keystore
Preparing trust store certificates for internal communication is completed
Preparing key store certificates for internal communication
Preparing key store certificates for internal communication is completed
[...]

PR #9468, opened 2023-12-15
Use separate logger for Jetty in Topic Operator

scholzj:
### Type of change

- Bugfix

### Description

This PR adds a separate logger for Jetty in the logging configurations of the Topic Operator. That helps to make sure that when the user enables the DEBUG logging for the root logger of the TO, Jetty will by default stay at the INFO level. This helps to reduce the number of messages that do not seem very useful such as the various Jetty threading logs etc.

This contributes to #9465 (but does not address all parts of it).

### Checklist


fvaleri (approved):
Thanks.

ppatierno (approved):

PR #9486, opened 2023-12-22
Remove Cruise Control's cluster.configs.file and related code

fvaleri:
The `cluster.configs.file` configuration property is never used. The `KafkaTopicConfigProvider` is now deprecated, and `KafkaAdminTopicConfigProvider` is used by default. The `KafkaAdminTopicConfigProvider` retrieves cluster configs via the Kafka admin client rather than reading them from a file.


scholzj:
Maybe we should finish the discussion in #9471 before doing this? As that seems to suggest this is needed and used. I do not think opening more PRs is the right solution how to decide the next steps.

fvaleri:
Ok, I'll mention the possibility to open it in my comment.

fvaleri:
Ok, I'll mention the possibility to open it in my comment.

scholzj:
@fvaleri Maybe we should discuss both PRs in some call in January. Probably something ad-hoc as the next community call is early CET, so @kyguy would not make it there.

PR #9488, opened 2023-12-22
Update the upgrade/downgrade tests to use 0.39.0

scholzj:
### Type of change

- Task

### Description

This PR tries to add the 0.39.0 release to the Upgrade / Downgrade tests.


scholzj:
Closing as it seems to need more changes due to the feature gate changes in 0.39 compared to 0.38. Will be done later by the experts.

PR #9496, opened 2024-01-02
[ST] Create NamespaceManager for handling namespaces in the tests

im-konge:
### Type of change

- Enhancement / Refactoring

### Description

Currently we have creation and deletion of namespaces split across multiple classes, together with copying pull secrets and applying NetworkPolicies. This makes the work with STs complicated, as in case that we would like to change something in behavior of creation/deletion methods for the namespaces, we have to change it on multiple places.

Also, keeping the create/delete methods in the `KubeClusterResource` class makes things complicated - f.e. in case we would like to use methods from `systemtest` module inside the Namespace related methods, we cannot simply do that - as it would create cyclic dependency. One of the example is now, when UTO adds finalizers to KafkaTopics and we are deleting namespaces before deleting the KafkaTopics, the namespace deletion is stuck -> we have method in our `systemtest` module, which we cannot use in `test` module, so we would have to workaround it again.

This PR creates new class - `NamespaceManager`, deletes `NamespaceUtils` class as it's not needed anymore and does a small refactor.

### Checklist


systemtest/src/main/java/io/strimzi/systemtest/resources/NamespaceManager.java
```diff
@@ -0,0 +1,169 @@
...
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+
+import static io.strimzi.test.k8s.KubeClusterResource.kubeClient;
+
+public class NamespaceManager {
```
see-quick, line 29: Javadoc? :)

see-quick (commented):
LGTM. I have only one note that is worth consideration but not the scope of this PR. We should change the way how we store `MAP_WITH_SUITE_NAMESPACES` because, with the removal of class-wide parallelism, we do not need `test class -> {namespaces...}` we need only one dynamic object storing it.

im-konge:
> LGTM. I have only one note that is worth consideration but not the scope of this PR. We should change the way how we store MAP_WITH_SUITE_NAMESPACES because, with the removal of class-wide parallelism, we do not need test class -> {namespaces...} we need only one dynamic object storing it.

Yeah I didn't really want to touch that logic, but I thought about it too. There is another thing -> that the NP are not really applied. But I don't want to mess more things together and I'll rather fix it in different ones.

see-quick (approved):

PR #9499, opened 2024-01-03
Remove Cruise Control's cluster.configs.file and related code

fvaleri:
The Cruise Control's cluster.configs.file configuration property is never used. The KafkaTopicConfigProvider is now deprecated in favor of KafkaAdminTopicConfigProvider, which is used by default. The KafkaAdminTopicConfigProvider retrieves cluster configs via the Kafka admin client rather than reading them from a file. For these reasons, we can get rid of this configuration and related code.


kyguy (approved):
Nice catch @fvaleri, looks good to me!

scholzj (approved):

tombentley (approved):

PR #9500, opened 2024-01-03
Fix typo in import control

fvaleri:
This change fixes a typo in the import-control configuration file.

### DIFF
--- a/.checkstyle/import-control.xml
+++ b/.checkstyle/import-control.xml
@@ -39,7 +39,6 @@ We also control imports only in production classes and not in tests. This is con
         <allow pkg="io.strimzi.api.annotations" />
         <allow pkg="io.strimzi.api.kafka" />
         <allow class="io.strimzi.api.ResourceLabels" />
-        <allow class="io.strimzi.api.ResourceAnnotations" />
         <allow pkg="io.strimzi.certs" />
         <allow pkg="io.strimzi.operator.common.model" />
         <allow class="io.strimzi.operator.common.Reconciliation" />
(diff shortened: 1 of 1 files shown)


scholzj (commented):
I guess the rule is not needed if everything worked despite the typo?

fvaleri:
Unlike `Labels` that is in `operator-common/io/strimzi/operator/common/model`, `Annotations` is in `operator-common/io/strimzi/operator/common`. Is this correct, or we want to move `Annotations` inside model and keep the rule?

scholzj:
> Unlike `Labels` that is in `operator-common/io/strimzi/operator/common/model`, `Annotations` is in `operator-common/io/strimzi/operator/common`. Is this correct, or we want to move `Annotations` inside model and keep the rule?

Sorry, I don't follow what you mean by this.

If the rule refers to a wrong class and nothing is failing, it seems that either the import control does not work properly or the rule is not needed at all.

fvaleri:
We have two similar rules for operator-common:

```sh
<import-control pkg="io.strimzi" strategyOnMismatch="allowed">
    <subpackage name="operator.common.model" strategyOnMismatch="disallowed">
        ...
        <allow class="io.strimzi.api.ResourceLabels" />        <!-- rule 1 -->
        <allow class="io.strimzi.api.ResourceAnnotations" />   <!-- rule 2 -->
```

These rules are supposed to be used by the following classes:

```sh
io.strimzi.operator.common.model.Labels // extends io.strimzi.api.ResourceLabels 
io.strimzi.operator.common.Annotations // extends io.strimzi.api.ResourceAnnotations
```

Nothing is failing because `Annotations` is NOT in the `operator.common.model` package. 

Now, there are two paths here:

1. Simply get rid of rule 2
2. Move the `Annotations` class from `io.strimzi.operator.common.` to `io.strimzi.operator.common.model` and keep the current change.

scholzj:
I would probably remove the rule.

scholzj (approved):

ppatierno (approved):

see-quick (approved):

PR #9504, opened 2024-01-04
[ST] Fix detecting kind in systemtests

jankalinic:
### Type of change

- Refactoring

### Description

This PR is just a small fix to correctly detect the kind cluster type only when kind cluster is present and then act accordingly when running STs. This change should avoid selecting kind during STs run - when no kind cluster was available and tester was on different cluster type.

### Checklist

### DIFF
--- a/test/src/main/java/io/strimzi/test/k8s/cluster/Kind.java
+++ b/test/src/main/java/io/strimzi/test/k8s/cluster/Kind.java
@@ -45,9 +45,9 @@ public class Kind implements KubeCluster {
      */
     @Override
     public boolean isClusterUp() {
-        List<String> cmd = Arrays.asList(CMD, "status");
+        List<String> cmd = Arrays.asList("kubectl", "get", "nodes", "-o", "jsonpath='{.items[*].spec.providerID}'");
         try {
-            return Exec.exec(cmd).exitStatus();
+            return Exec.exec(cmd).out().startsWith("kind");
         } catch (KubeClusterException e) {
             LOGGER.debug("'" + String.join(" ", cmd) + "' failed. Please double check connectivity to your cluster!");
             LOGGER.debug(e);


test/src/main/java/io/strimzi/test/k8s/cluster/Kind.java
```diff
@@ -45,9 +45,9 @@ public boolean isAvailable() {
      */
     @Override
     public boolean isClusterUp() {
-        List<String> cmd = Arrays.asList(CMD, "status");
+        List<String> cmd = Arrays.asList("kubectl", "get", "nodes", "-o", "jsonpath='{.items[*].spec.providerID}'");
```
Frawless, line 48: Wouldn't be better somehow check it with `kind` client?

jankalinic: kind can display nodes and clusters created by kind, which may have different names, but i guess it can be done using `kind get nodes -A` -> with expected output of "No kind nodes for any cluster". Otherwise this way we can be sure of kind presence @Frawless.

Frawless: That sounds more reasonable to me. Kubectl will work, but I think it could pick `kind` even when running cluster will actually be `openshift` due to the following list:

 ```
         if (clusters == null) {
            clusters = new KubeCluster[]{new Minikube(), new Kind(), new Kubernetes(), new OpenShift(), new Microshift()};
        }
        ```

im-konge: But there is the check that the cluster is up. So in case that the provider will not start with `kind`, it will be skipped

Frawless: @im-konge but this check is replaced from `kind` to `kubectl` by @jankalinic  or? There is `isAvailable` executed before `isClusterUp` but `isAvailable` basically says that there is `kind` command available, this still can be true when running on openshift.

im-konge: The `isAvailable` can pass and be true, but the there is `isClusterUp`, which does this call -> and in case that the nodes will not have the specified provider, it should be false, thus skipping Kind or?

```
for (KubeCluster kc : clusters) {
            if (kc.isAvailable()) {
                logger.debug("Cluster {} is installed", kc);
                if (kc.isClusterUp()) {
                    logger.debug("Cluster {} is running", kc);
                    cluster = kc;
                    break;
                } else {
                    logger.debug("Cluster {} is not running!", kc);
                }
            } else {
                logger.debug("Cluster {} is not installed!", kc);
            }
        }
```

Frawless: I missed the part with checking `kind` in the providerID, sorry. In that case it should work I guess. Not sure if it possible to change the field manually or the field is immutable - that could lead to problem described above.

jankalinic: To sum it up, i would like to do it as @Frawless  suggested using kind. The thing is, it really goes by the order, so since we skip the `isAvailable` and focus on `isClusterUp`, then we check `minikube status` which is fine, then the next one is kind, and since it asserted only exitstatus and the result of `kind status` is return of `help` when you have kind installed it asumes you are already on kind. Next thing is that i do not feel 100% sure about just taking status code from `kind get clusters / kind get nodes -A` as it returns **something** (i know it can return error in case of no clusters available, not sure of all the circumstances that can ocure) and `kind` command (from what i see in docs` cannot return something like "kind is up, you are using cluster mycluster". So thats why i went into specs of nodes using kube, as i feel safer with this than with some input..... To support why i think json is the right thing here, this code comes from kind main: `provider-id: "kind://{{.NodeProvider}}/{{.ClusterName}}/{{.NodeName}}"` https://github.com/kubernetes-sigs/kind/blob/40c81f187425254daf2bf84360a6257a278252df/pkg/cluster/internal/kubeadm/config.go which means kind prefix will always be there. If i did not sum it up, or something is missing, please tell me.

see-quick (approved):
LGTM

im-konge (approved):

jankalinic:
@strimzi-ci run tests --cluster-type=ocp --cluster-version=4.14 --profile=all --testcase=ConnectST

Frawless (approved):

test/src/main/java/io/strimzi/test/k8s/cluster/Kind.java
Frawless: That makes sense. Sorry for the noise.

PR #9506, opened 2024-01-04
[ST] Update upgrade/downgrade files after 0.39.0 release and enable KRaftStrimziDowngradeST

im-konge:
### Type of change

- Enhancement

### Description

This PR updates upgrade/downgrade files after the 0.39.0 release. It also enables `KRaftStrimziDowngradeST`, which was blocked until new version of Strimzi, which supports KRaft to KRaft upgrade/(safe) downgrade is not released.

### Checklist

### DIFF
--- a/systemtest/src/main/java/io/strimzi/systemtest/utils/StUtils.java
+++ b/systemtest/src/main/java/io/strimzi/systemtest/utils/StUtils.java
@@ -12,6 +12,7 @@ import com.fasterxml.jackson.dataformat.yaml.YAMLMapper;
 import com.jayway.jsonpath.JsonPath;
 import io.fabric8.kubernetes.api.model.Affinity;
 import io.fabric8.kubernetes.api.model.ConfigMap;
+import io.fabric8.kubernetes.api.model.Container;
 import io.fabric8.kubernetes.api.model.LabelSelector;
 import io.fabric8.kubernetes.api.model.Pod;
 import io.fabric8.kubernetes.api.model.Secret;
@@ -43,6 +44,7 @@ import java.util.Collections;
 import java.util.List;
 import java.util.Locale;
 import java.util.Map;
+import java.util.Optional;
 import java.util.Properties;
 import java.util.function.BiFunction;
 import java.util.function.BooleanSupplier;
@@ -548,5 +550,28 @@ public class StUtils {
     public static void waitUntilSupplierIsSatisfied(final BooleanSupplier sup) {
         TestUtils.waitFor(sup.getAsBoolean() + " is satisfied", TestConstants.GLOBAL_POLL_INTERVAL,
                 TestConstants.GLOBAL_STATUS_TIMEOUT, sup);
+    }
+
+    /**
+     * Checks env variables of Topic operator container (inside EO Pod) and based on that determines, if BTO or UTO is used.
+     * Normally we can determine that based on {@link Environment#isUnidirectionalTopicOperatorEnabled()}, but in case that the
+     * FG is changed during test, we have no other way to check it.
+     *
+     * @param namespaceName name of the Namespace, where the EO Pod is running
+     * @param eoLabelSelector LabelSelector of EO
+     * @return boolean determining if UTO is used or not
+     */
+    public static boolean isUnidirectionalTopicOperatorUsed(String namespaceName, LabelSelector eoLabelSelector) {
+        Optional<Container> topicOperatorContainer = kubeClient().listPods(namespaceName, eoLabelSelector).get(0).getSpec().getContainers()
+            .stream().filter(container -> container.getName().equals("topic-operator")).findFirst();
+
+        if (topicOperatorContainer.isPresent()) {
+            return topicOperatorContainer.get().getEnv()
+                // ZK related env vars are only present in BTO mode -> because of that we can determine which of the TOs is used
+                .stream().noneMatch(envVar -> envVar.getName().contains("ZOOKEEPER"));
+        }
+
+        LOGGER.warn("Cannot determine if UTO is used or not, because the EO Pod doesn't exist, gonna assume that it's not");
+        return false;
     }
 }
--- a/systemtest/src/test/java/io/strimzi/systemtest/upgrade/AbstractUpgradeST.java
+++ b/systemtest/src/test/java/io/strimzi/systemtest/upgrade/AbstractUpgradeST.java
@@ -28,6 +28,7 @@ import io.strimzi.systemtest.kafkaclients.internalClients.KafkaClientsBuilder;
 import io.strimzi.systemtest.resources.ResourceManager;
 import io.strimzi.systemtest.resources.crd.KafkaConnectResource;
 import io.strimzi.systemtest.resources.crd.KafkaResource;
+import io.strimzi.systemtest.resources.crd.KafkaTopicResource;
 import io.strimzi.systemtest.storage.TestStorage;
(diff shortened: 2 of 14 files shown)


systemtest/src/test/java/io/strimzi/systemtest/upgrade/AbstractUpgradeST.java
```diff
@@ -250,6 +251,8 @@ protected void changeClusterOperator(BundleVersionModificationData versionModifi
     }
 
     protected void copyModifyApply(File root, String namespace, final String strimziFeatureGatesValue) {
+        KubeClusterResource.getInstance().setNamespace(namespace);
```
Frawless, line 254: This is not related to the PR, but I think we should really remove options like this and specify namespace in every command that somehow interact with kube api. Partially it is done for fabric8 clients I think. IIUC here it could be passed to binding config directly or just to cmdKubeClient? WDYT?

systemtest/src/test/java/io/strimzi/systemtest/upgrade/kraft/AbstractKRaftUpgradeST.java
```diff
@@ -77,7 +76,7 @@ protected void deployKafkaClusterWithWaitForReadiness(final ExtensionContext ext
                 resourceManager.createResourceWithWait(extensionContext,
                     KafkaNodePoolTemplates.kafkaNodePoolWithControllerRoleAndPersistentStorage(TestConstants.CO_NAMESPACE, CONTROLLER_NODE_NAME, clusterName, 3).build(),
                     KafkaNodePoolTemplates.kafkaNodePoolWithBrokerRoleAndPersistentStorage(TestConstants.CO_NAMESPACE, BROKER_NODE_NAME, clusterName, 3).build(),
-                    KafkaTemplates.kafkaPersistent(clusterName, 3, 3)
+                    KafkaTemplates.kafkaPersistentKRaft(clusterName, 3)
```
Frawless, line 79: isn't it also used in non-kraft tests? If so, the name is a little bit misleading

systemtest/src/test/java/io/strimzi/systemtest/upgrade/regular/StrimziDowngradeST.java
```diff
@@ -95,6 +100,10 @@ void setupEnvironment() {
 
     @AfterEach
     void afterEach() {
+        // delete all topics created in test
+        cmdKubeClient(TestConstants.CO_NAMESPACE).deleteAllByResource(KafkaTopic.RESOURCE_KIND);
+        KafkaTopicUtils.waitForTopicWithPrefixDeletion(TestConstants.CO_NAMESPACE, topicName);
```
Frawless, line 105: Just thinking...all CR topis are created by tests? If so, why ResourceManager doesn't handle them?

systemtest/src/test/java/io/strimzi/systemtest/upgrade/AbstractUpgradeST.java
im-konge: Yeah I know, I'm encountering stuff like this all the time. I have a list of things that should be done, but I don't want to do it in one PR. I wanted to add to all methods the specification of Namespace, but it's not related to this PR. I just needed to fix a problem when the Namespace is recreated and we are watching different one in tests. That is happening after my last PR with NamespaceManager. So it's a solution just for now, I will fix that in another PR if that makes sense.

Frawless: I think the NamespaceManager was good PR to implement it.

systemtest/src/test/java/io/strimzi/systemtest/upgrade/kraft/AbstractKRaftUpgradeST.java
im-konge: FMPOV we should have different templates for KRaft and non-KRaft Kafka CRs and depending on FG env variable pick one or another. I need to have a template that doesn't contain LMFV and IBPV, otherwise it will fail the Kafka containers because of wrong version of those set during the downgrade. If the name is confusing - what would you pick?

systemtest/src/test/java/io/strimzi/systemtest/upgrade/regular/StrimziDowngradeST.java
im-konge: ResourceManager deletes them, but after this AfterEach. Inside this one there is a method for deleting the applied files - even the CRDs. But because of finalizers, you cannot delete the KafkaTopic CRD, thus the process willbe stuck. Removing the KafkaTopics before that will ensure that everything will be cleared without a problem

systemtest/src/test/java/io/strimzi/systemtest/upgrade/AbstractUpgradeST.java
im-konge: Well yes, but it will need another full walkthrought the STs and a lot of lines of code in my opinion. I thought that it will be better to have just one PR for NamespaceManager and another one for adding that possibility

systemtest/src/test/java/io/strimzi/systemtest/upgrade/regular/StrimziDowngradeST.java
im-konge: Ah sorry, I missed the "all CR topics" - most of them are deployed using resourceManager, but I guess one or two are deployed using YAML and the `cmdKubeClient`
So that's why there are extra cleaning steps (to remove YAML files applied during tests before ResourceManager takes care about other CRs)

systemtest/src/main/java/io/strimzi/systemtest/utils/StUtils.java
```diff
@@ -549,4 +549,19 @@ public static void waitUntilSupplierIsSatisfied(final BooleanSupplier sup) {
...
+     * FG is changed during test, we have no other way to check it.
+     *
+     * @param namespaceName name of the Namespace, where the EO Pod is running
+     * @param eoLabelSelector LabelSelector of EO
+     * @return boolean determining if UTO is used or not
+     */
+    public static boolean isUnidirectionalTopicOperatorUsed(String namespaceName, LabelSelector eoLabelSelector) {
+        return kubeClient().listPods(namespaceName, eoLabelSelector).get(0).getSpec().getContainers()
+            .stream().filter(container -> container.getName().equals("topic-operator")).findFirst().get().getEnv()
+            .stream().anyMatch(envVar -> envVar.getName().contains("ZOOKEEPER"));
```
henryZrncik, line 565: I guess this is due to fact that `ZOOKEEPER` is present only in bidirectional but could this be mentioned in comment ?

henryZrncik (approved):
LGTM !

systemtest/src/main/java/io/strimzi/systemtest/utils/StUtils.java
im-konge: good point, will do

see-quick (approved):
Good job! 👍 very nice and clean PR :))

systemtest/src/test/java/io/strimzi/systemtest/upgrade/kraft/AbstractKRaftUpgradeST.java
Frawless: Discussed offline, lgtm

Frawless (approved):

jankalinic (approved):
lgtm

PR #9507, opened 2024-01-05
[ST] Enable connector autoRestart test

jankalinic:
### Type of change

- Refactoring

### Description

While going through STs i've encountered this test which seems to be working now as expected, so i've decided to enable the last checks which seemed to be flaky with current behaviour.

### Checklist

- [ ] Write tests


jankalinic:
@strimzi-ci run tests --cluster-type=ocp --cluster-version=4.14 --profile=all --testcase=ConnectST#testConnectorTaskAutoRestart

henryZrncik (approved):

Frawless (approved):

im-konge (approved):
I'm fine with enabling the check if it works, but I'm really curious what was the cause that it failed all the time before

PR #9508, opened 2024-01-05
Use `additionalProperties` instead of `x-kubernetes-preserve-unknown-fields` for `Map<String, String>` fields

### LINKED ISSUE #7985: Use `additionalProperties` in the CRD OpenAPI spec for Maps of Strings instead of `x-kubernetes-preserve-unknown-fields`

scholzj:
### Type of change

- Enhancement / new feature

### Description

This PR changes how we generate the CRD schema for fields with type `Map<String, String>` where this PR uses the `additionalProeprties` field instead of `x-kubernetes-preserve-unknown-fields`. That should help users using toosl such as Terraform that seem to have some weird behavior when dealing with the `x-kubernetes-preserve-unknown-fields` flag. However, this can be applied only to `Map<String, String>` and not to `Map<String, Object>`, so it is not completely clear how much it will help.

This should resolve #7985.

### Checklist

- [x] Write tests

### DIFF
--- a/crd-generator/src/main/java/io/strimzi/crdgenerator/CrdGenerator.java
+++ b/crd-generator/src/main/java/io/strimzi/crdgenerator/CrdGenerator.java
@@ -798,7 +798,12 @@ class CrdGenerator {
                 || long.class.equals(elementType)) {
             itemResult.put("type", "integer");
         } else if (Map.class.equals(elementType)) {
-            preserveUnknownFields(itemResult);
+            if (isStringStringMap(propertyType)) {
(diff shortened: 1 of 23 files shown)


ppatierno (approved):
LGTM

PR #9514, opened 2024-01-06
Set UTO as default standalone deployment

fvaleri:
The UTO is now in beta phase, which means enabled by default with the Cluster Operator deployment. This change makes the UTO enabled by default also with the standalone deployment.

No need to update the documentation, because the procedure to switch back to standalone BTO is already documented.


scholzj (approved):
Thanks @fvaleri

im-konge (approved):

ppatierno (approved):

PR #9529, opened 2024-01-09
[ST] Upgrade/downgrade - specify file path for Kafka in the YAML files

im-konge:
### Type of change

- Enhancement 

### Description

The file paths to the examples of Kafka are currently hard-coded inside the upgrade/downgrade tests, which can be a problem in case that the file path is changed between versions.

This PR adds new fields to upgrade/downgrade YAML files, making possible to specify full path to Kafka YAML files.

### Checklist

### DIFF
--- a/systemtest/src/main/java/io/strimzi/systemtest/upgrade/CommonVersionModificationData.java
+++ b/systemtest/src/main/java/io/strimzi/systemtest/upgrade/CommonVersionModificationData.java
@@ -4,6 +4,8 @@
  */
 package io.strimzi.systemtest.upgrade;
 
+import java.util.Map;
+
 /**
  * This class contains all the common fields needed for upgrade and downgrade tests from which specific classes
  * inherit these values and extending them further for specific needs of the process.
@@ -17,6 +19,7 @@ public class CommonVersionModificationData {
     private String toExamples;
     private String toUrl;
     private UpgradeKafkaVersion procedures;
+    private Map<String, String> filePaths;
 
     public String getFromVersion() {
         return fromVersion;
@@ -80,6 +83,30 @@ public class CommonVersionModificationData {
 
     public void setProcedures(UpgradeKafkaVersion procedures) {
         this.procedures = procedures;
+    }
+
+    public void setFilePaths(Map<String, String> filePaths) {
+        this.filePaths = filePaths;
+    }
+
+    public Map<String, String> getFilePaths() {
+        return filePaths;
+    }
+
+    public String getKafkaFilePathBefore() {
+        return getFilePaths().get("kafkaBefore");
+    }
+
+    public String getKafkaFilePathAfter() {
+        return getFilePaths().get("kafkaAfter");
+    }
+
+    public String getKafkaKRaftFilePathBefore() {
+        return getFilePaths().get("kafkaKRaftBefore");
+    }
+
+    public String getKafkaKRaftFilePathAfter() {
+        return getFilePaths().get("kafkaKRaftAfter");
     }
 
     @Override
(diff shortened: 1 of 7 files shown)


jankalinic (approved):
Looks smooth

systemtest/src/main/java/io/strimzi/systemtest/upgrade/CommonVersionModificationData.java
```diff
@@ -82,6 +85,30 @@ public void setProcedures(UpgradeKafkaVersion procedures) {
...
+    public void setFilePaths(Map<String, String> filePaths) {
+        this.filePaths = filePaths;
+    }
+
+    public Map<String, String> getFilePaths() {
+        return filePaths;
+    }
+
+    public String getKafkaFilePathBefore() {
+        return getFilePaths().get("kafkaBefore");
```
see-quick, line 97: Not scope of this PR but maybe we should introduce constraints to map the upgrade file attributes? :)

see-quick (approved):
Looks good 👍 LGTM

systemtest/src/main/java/io/strimzi/systemtest/upgrade/CommonVersionModificationData.java
im-konge: Like not null, size etc?
sure we can

systemtest/src/test/java/io/strimzi/systemtest/upgrade/AbstractUpgradeST.java
```diff
@@ -148,23 +149,23 @@ protected void changeKafkaAndLogFormatVersion(CommonVersionModificationData vers
...
 
-        kafkaYaml = new File(examplesPath + "/kafka/kafka-persistent.yaml");
+        kafkaYaml = new File(examplesPath + versionModificationData.getKafkaFilePathAfter());
         LOGGER.info("Deploying Kafka from: {}", kafkaYaml.getPath());
         // Change kafka version of it's empty (null is for remove the version)
         String defaultValueForVersions = kafkaVersionFromCR == null ? null : TestKafkaVersion.getSpecificVersion(kafkaVersionFromCR).messageVersion();
         cmdKubeClient().applyContent(KafkaUtils.changeOrRemoveKafkaConfiguration(kafkaYaml, kafkaVersionFromCR, defaultValueForVersions, defaultValueForVersions));
 
-        kafkaUserYaml = new File(examplesPath + "/user/kafka-user.yaml");
+        kafkaUserYaml = new File(examplesPath + "/examples/user/kafka-user.yaml");
```
Frawless, line 164: why not use `PATH_TO_PACKAGING_EXAMPLES` ?

im-konge: Because it's not always packaging from where we are getting the examples

+ the path for Kafka files is changed now and there can be anything inside the packaging folder. Because I'm changing for Kafka, I need to change it everywhere else.

Frawless: Makes sense. Maybe use constant for `examples` as well and use it in other constants, but maybe it is not needed.

Frawless (approved):

systemtest/src/test/java/io/strimzi/systemtest/upgrade/AbstractUpgradeST.java
im-konge: Maybe it would make sense to have these file paths in one of the classes that collects info for the scenario and add getters, which would then just return the exact paths - it would be better than having it hardcoded on X places.

PR #9540, opened 2024-01-10
[ST] KafkaNodePools Roll only broker role due to broker only config change

henryZrncik:
### Type of change

- Enhancement 

### Description

addition check:
Change of broker role specific configuration in shared `Kafka` custom resource triggers rolling of broker roles. 
 (as this step was previously omitted) 

### Checklist

- [x] Write tests


see-quick (approved):

im-konge (approved):

im-konge:
@henryZrncik the PR that enables KRaft by default is merged, could you please rebase, so we can run the pipelines before merge?

PR #9545, opened 2024-01-11
Abstract client certificate handling

katheris:
### Type of change

_Select the type of your PR_

- Refactoring

### Description

Addresses: https://github.com/strimzi/strimzi-kafka-operator/issues/5630

- Creates `PemTrustSet`, `PemAuthIdentity`, and `ClusterOperatorPKSC12AuthIdentity` to encapsulate the certificates used during mutual TLS authentication.
- Updates KafkaReconciler and ZooKeeperReconciler to fetch clientSecrets once. Since the secret is updated as part of the `CaReconciler` there is no need to fetch it each time we do different parts of the reconcile loop.

### Checklist

- [ ] Write tests
- [ ] Update documentation
- [x] Update CHANGELOG.md


cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaReconciler.java
```diff
@@ -308,6 +315,20 @@ protected Future<Void> modelWarnings(KafkaStatus kafkaStatus) {
         return Future.succeededFuture();
     }
 
+    /**
+     * Initialize the TrustStore and KeyStore suppliers to be used by admin clients during reconciliation
+     *
+     * @return Completes when the TrustStore and KeyStore suppliers have been created
+     */
+    protected Future<Void> initAdminClientKeyStoreTrustStore() {
+        return ReconcilerUtils.clientSecrets(reconciliation, secretOperator)
```
tombentley, line 324: I wonder if we should push knowledge of the `Pem...` down into the ReconcilerUtils, so we have a `Future<PemKeyStoreSupplier> ReconcilerUtils.clientIdentity(Reconciliation reconciliation, SecretOperator secretOperator)` and likewise for the trust.

katheris: Yeah I think that makes sense

operator-common/src/main/java/io/strimzi/operator/common/model/PemKeyStoreSupplier.java
```diff
@@ -0,0 +1,24 @@
+/*
+ * Copyright Strimzi authors.
+ * License: Apache License 2.0 (see the file LICENSE or http://apache.org/licenses/LICENSE-2.0.html).
+ */
+package io.strimzi.operator.common.model;
+
+/**
+ * Interface to be implemented for returning a PEM KeyStore
+ */
+public interface PemKeyStoreSupplier {
```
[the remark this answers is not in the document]
katheris: You're right that it doesn't provide a keystore directly. The thing that I do like about the current name is that it makes it clear what it is intended to be used for. Although `PemPeerIndentity` is a more accurate description, I wonder if it wouldn't be as clear to people trying to use it. What about something more like `PemClientAuthIdentity`? Then `PemTrustSet` for the other one like you suggested elsewhere.

Taking into account your other comment about the keystore methods that the KafkaAgentClient uses, perhaps we could update the "Supplier" class to also have a jksKeyStore() method. If we did that we could drop the `Pem` from the name.

In terms of interface vs record vs class, although currently `ClusterOperatorKeyStoreSupplier` and `UserOperatorKeyStoreSupplier` have different logic to parse the `Secret`, looking at them again I think the logic could be common, it's just the keys in the Secret data that differs.  So I wonder if having `ClientAuthIdentity` as an abstract class makes sense? Then the concrete classes would be `ClusterOperatorClientAuthIdentity` and `UserOperatorClientAuthIdentity` (both quite long names) which each defines the secret key to use for the lookup.

I'll take a look at addressing the null/empty comments. I didn't follow what you meant about the `Optional`, are you suggesting the create method for the suppliers (or whatever we call them) should return an optional? I wonder if the constructor should throw an error if the parsing fails, although I'm not sure what ripple effect that will have on the tests, so will take a look.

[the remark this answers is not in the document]
tombentley: Is the `Client` part really, needed in `PemClientAuthIdentity`? Since a bundle of cert, key and CA cert chain can also be used to identify a server when doing mTLS I suspect we have a `PemAuthIdentity` the root of your abstract class hierarchy, with `PemClientAuthIdentity` as a subtype, and then the CO and UO classes as subtypes of that, if that would be useful. I'm not sure whether we really need distinct `ClusterOperatorClientAuthIdentity` and `UserOperatorClientAuthIdentity` types though. The difference between them appears to me mainly to be how they're constructed (i.e. which Secrets and what the data is expected to contain), rather than wanting to constrain what you can pass instances to, or a requirement for method overloading. This is why I suggested using factory methods. But if you really want separate classes I'm fine with that.

In any case I guess I'm happy with those names you suggest.

My point about Optional was you that with the interface you have to rely on documenting that the key shouldn't be null if the cert is not null, and vice versa. This is a possibility you can completely exclude with a class (because you can enforce rules in the constructor), which simplifies life for callers.

operator-common/src/main/java/io/strimzi/operator/common/model/PemTrustSet.java
```diff
@@ -0,0 +1,112 @@
...
+     */
+    public String trustedCertificatesString() {
+        return trustedCertificateMap.values()
+                .stream()
+                .map(bytes -> new String(bytes, StandardCharsets.US_ASCII))
+                .collect(Collectors.joining("\n"));
+    }
+
+    /**
+     * Validate the set of certificates in this PemTrustSet.
```
tombentley, line 74: This method returns something, which you're not documenting. So this might be better named something like `asX509Certificates()`.

operator-common/src/main/java/io/strimzi/operator/common/model/PemTrustSet.java
```diff
@@ -0,0 +1,112 @@
...
+import java.security.cert.X509Certificate;
+import java.util.Base64;
+import java.util.HashSet;
+import java.util.Map;
+import java.util.Optional;
+import java.util.Set;
+import java.util.stream.Collectors;
+
+/**
+ * Class to represent the set of trusted certificates used by TLS clients
```
tombentley, line 21: I don't think there's anything here which is specific to clients

```suggestion
 * Represents the set of certificates to be trusted by a TLS client or server
```

operator-common/src/main/java/io/strimzi/operator/common/model/PemAuthIdentity.java
```diff
@@ -0,0 +1,112 @@
...
+import io.strimzi.operator.common.Util;
+
+import java.io.ByteArrayInputStream;
+import java.security.cert.Certificate;
+import java.security.cert.CertificateException;
+import java.security.cert.CertificateFactory;
+import java.util.Optional;
+
+/**
+ * Class to represent the identity used during TLS client authentication in the PEM format.
```
tombentley, line 17:
```suggestion
 * Represents the identity used during TLS client authentication.
 * This consists of an X509 end entity certificate, corresponding private key, and a chain of X509 CA certificates, all in PEM format.
```

operator-common/src/main/java/io/strimzi/operator/common/model/PemAuthIdentity.java
```diff
@@ -0,0 +1,112 @@
...
+ */
+public class PemAuthIdentity extends AbstractAuthIdentity {
+
+    private final byte[] pemPrivateKey;
+    private final byte[] pemCertificateChainBytes;
+    private final Certificate pemCertificateChain;
```
tombentley, line 23: Java's `Certificate` is kind of useless. I don't think we're ever expecting to use certificates that are not X509 certificates, so we should downcast and use `X509Certificate` in the API.

operator-common/src/main/java/io/strimzi/operator/common/model/AbstractAuthIdentity.java
```diff
@@ -0,0 +1,86 @@
...
+import io.strimzi.operator.common.Util;
+
+import java.nio.charset.StandardCharsets;
+import java.util.Base64;
+import java.util.Optional;
+
+/**
+ * Abstract class to represent the identity used during TLS client authentication
+ */
+public abstract class AbstractAuthIdentity {
```
tombentley, line 18: This class declares no member methods (just `static` methods). This suggests to me that it's either not finished, or it's a weak abstraction (whatever behaviour it's abstracting cannot be expressed by any member methods).

- If you just need some functions then you can either put them on appropriate other classes or declare a non-`abstract` class with a private constructor to hold them (no need for `abstract`).
- Or, if you really think there's a useful abstraction here ask yourself why there are no member methods. Seemingly there are no call sites which need to manipulate just a `AbstractAuthIdentity`, so do you actually need an abstraction?

operator-common/src/main/java/io/strimzi/operator/common/model/PemAuthIdentity.java
```diff
@@ -0,0 +1,112 @@
...
+    private PemAuthIdentity(Secret secret, String secretKey) {
+        Optional.ofNullable(secret)
+                .map(Secret::getMetadata)
+                .ifPresent(objectMeta -> {
+                    secretName = objectMeta.getName();
+                    secretNamespace = objectMeta.getNamespace();
+                });
+        pemPrivateKey = AbstractAuthIdentity.extractPemPrivateKey(secret, secretKey);
+        pemCertificateChainBytes = AbstractAuthIdentity.extractPemCertificateChain(secret, secretKey);
+        pemCertificateChain = validateCertificateChain(secretKey);
```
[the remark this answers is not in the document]
katheris: I've removed the step to explicitly validate the certificates as it was proving very tricky to integrate with the existing tests. So it now only creates the Certificate object when needed

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/resource/KafkaAgentClient.java
```diff
@@ -100,26 +100,26 @@ private HttpClient createHttpClient() {
...
-                Util.decodeFromSecret(clusterCaCertSecret, "ca.crt")));
+                pemTrustStoreSupplier.pemTrustedCertificates().getBytes(StandardCharsets.US_ASCII)));
         KeyStore trustStore = KeyStore.getInstance(KEYSTORE_TYPE_JKS);
         trustStore.load(null);
         trustStore.setCertificateEntry("ca", caCert);
         return trustStore;
     }
 
-    private KeyStore getKeyStore(char[] keyPassword) throws KeyStoreException, CertificateException, NoSuchAlgorithmException, InvalidKeySpecException, IOException {
+    private KeyStore getKeyStore() throws KeyStoreException, CertificateException, NoSuchAlgorithmException, InvalidKeySpecException, IOException {
```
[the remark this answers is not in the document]
katheris: I agree but given the size of the changes already we should do this in a followup PR, since it's done in a few places in subtly different ways

katheris:
@strimzi/maintainers this is ready for review if any of you have time to take a look

tombentley (approved):
Thanks @katheris, I found a couple of nits, but I think this is a welcome refactoring.

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/ClusterOperatorPKCS12AuthIdentity.java
```diff
@@ -0,0 +1,41 @@
...
+package io.strimzi.operator.cluster.model;
+
+import io.fabric8.kubernetes.api.model.Secret;
+import io.strimzi.operator.common.Util;
+
+/**
+ * Represents the identity used by the cluster operator during TLS client authentication in the PKCS12 format.
+ * Can be used by clients that are unable to use the PEM format provided by PemAuthIdentity.
+ */
+public class ClusterOperatorPKCS12AuthIdentity {
```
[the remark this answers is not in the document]
katheris: This is currently only used by the ZooKeeper reconcile loop, all other components are using PEM, which is also what I think we want moving forwards. So I chose to put it in cluster-operator to limit it's scope, since the norm should be to use the PEM version. Then hopefully it could be removed when ZooKeeper is no longer required. Happy to move it if others also think it is weird to place it here but that was my thought process behind it.

operator-common/src/main/java/io/strimzi/operator/common/Util.java
```diff
@@ -158,26 +185,50 @@ public static File createFileStore(String prefix, String suffix, byte[] bytes) {
...
+     * Decode the private key in PEM format, from base64 to a byte array.
+     * Before decoding it into byte array, it removes the PEM header and footer.
+     * @param privateKey        Key which should be decoded
+     * @return                  Decoded bytes
      */
-    public static byte[] decodePemPrivateKeyFromSecret(Secret secret, String key) {
-        String privateKey = new String(decodeFromSecret(secret, key), StandardCharsets.UTF_8)
+    public static byte[] decodePemPrivateKey(String privateKey) {
+        String decodedPrivateKey = privateKey
                 .replace("-----BEGIN PRIVATE KEY-----", "")
```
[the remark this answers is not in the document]
katheris: I agree that we could have better validation here. However I think we should include it in a follow up PR where we better align the creation of keystores and truststores across the project, since I don't think this should stay in Util ultimately.

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/ClusterOperatorPKCS12AuthIdentity.java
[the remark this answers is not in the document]
scholzj: I think ZooKeeper can use PEM files now. I had it working here: https://github.com/strimzi/strimzi-kafka-operator/compare/main...scholzj:strimzi-kafka-operator:use-pem-files?expand=1#diff-e5e5e71484b286fed9dc8f2d7cf2003356fb2f441d60521cb2da028026bb0a43

Should we first move the ZooKeeper clients to PEM and get rid of this? I can have a look into it @katheris if you want. Or you can do it your self if you want either as a separate PR or as part of this PR.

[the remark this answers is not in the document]
katheris: Thanks @scholzj  that's useful to know. There's a follow up PR I want to do to align the way we create truststores and keystores as we do it a few different ways in different places, shall I include it in that change? Since the migration code was added there are now a few places we construct a ZK client and the code is duplicated currently I believe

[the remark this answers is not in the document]
scholzj: Would it make sense to do that first to have this PR cleaner?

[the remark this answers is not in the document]
katheris: It would make this PR cleaner, but I'm hesitant do it first because this PR touches quite a few commonly changed classes so I've had to rebase it multiple times already, so it would be good to get it in sooner than later. Plus the changes in this PR will make the truststore/keystore refactor simpler to do and review, since in this PR I made some small changes to close the gap a little. For example in `KafkaAgentClient` using the `X509Certificate` class directly, rather than the `Certificate` class.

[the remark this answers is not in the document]
scholzj: Ok, than lets do it later.

CHANGELOG.md
```diff
@@ -13,6 +13,7 @@
 * Prevent removal of the `broker` process role from KRaft mixed-nodes that have assigned partition-replicas
 * Improve broker scale-down prevention to continue in reconciliation when scale-down cannot be executed
 * Added support for Tiered Storage by enabling the configuration of custom storage plugins through the Kafka custom resource.
+* Create classes to represent the trust set and identity for client authentication when creating clients using TLS in operators.
```
scholzj, line 16: Not sure this needs a CHANGELOG record as it is completely internal implementation detail, or?

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/ClusterOperatorPKCS12AuthIdentity.java
```diff
@@ -0,0 +1,41 @@
...
+    private static final String SECRET_KEY = "cluster-operator";
+    private final byte[] keyStore;
+    private final String password;
+
+    /**
+     * Constructs the ClusterOperatorPKCS12AuthIdentity.
+     * @param secret Kubernetes Secret containing the client authentication identity
+     */
+    public ClusterOperatorPKCS12AuthIdentity(Secret secret) {
```
scholzj, line 23: What is the motivation to use a class instead of using a record with static method that creates it from the Secret?

operator-common/src/main/java/io/strimzi/operator/common/Util.java
```diff
@@ -158,26 +185,50 @@ public static File createFileStore(String prefix, String suffix, byte[] bytes) {
...
+                    String name = Optional.ofNullable(secret)
+                            .map(Secret::getMetadata)
+                            .map(ObjectMeta::getName)
+                            .orElse("unknown");
+                    String namespace = Optional.ofNullable(secret)
+                            .map(Secret::getMetadata)
+                            .map(ObjectMeta::getNamespace)
+                            .orElse("unknown");
+                    return Util.missingDataInSecretException(namespace, name, field);
+                });
```
scholzj, line 206: Is this really an efficient and readable way to handle this? I would find using `if` constructs much easier to read and understand what exactly does this actually do.

operator-common/src/main/java/io/strimzi/operator/common/Util.java
```diff
@@ -128,6 +130,31 @@ public static RuntimeException missingSecretException(String namespace, String s
...
+     *
+     * @param namespace     Namespace of the Secret
+     * @param secretName    Name of the Secret
+     * @param keyName       Name of the Secret key
+     * @return              RuntimeException
+     */
+    public static RuntimeException corruptCertificateException(String namespace, String secretName, String keyName) {
+        return new RuntimeException("Bad/corrupt certificate found in data." + keyName + ".crt of Secret "
+                + secretName + " in namespace " + namespace);
+    }
```
scholzj, line 156: These methods seem strange. They do not seem to contain any real logic and seem to be used from a few places only - the places where they are used also do not seem to use the _central facilities_ for whatever reason. Would it be better to just inline this or create a new exception types and use this in their constructors?

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/CaReconciler.java
```diff
@@ -369,10 +380,12 @@ Future<Void> rollingUpdateForNewCaKey() {
         }
 
         if (podRollReasons.shouldRestart()) {
+            PemTrustSet pemTrustSet = new PemTrustSet(clusterCa.caCertSecret());
+            PemAuthIdentity pemAuthIdentity = PemAuthIdentity.clusterOperator(coSecret);
```
scholzj, line 384: Maybe you can name these based on what they are? E.g. `clusterCaTrustSet` and `coAuthIdentity`? I think that can be in general applied to all the parameters and fields -> they should not just copy the type name but they should explain the expected value. E.g. in the `BrokersInUseCheck` above, you have some clear expectation that should be passed in. And I assume it will be the same in many places below.

operator-common/src/main/java/io/strimzi/operator/common/Util.java
```diff
@@ -158,26 +185,50 @@ public static File createFileStore(String prefix, String suffix, byte[] bytes) {
...
+    public static byte[] decodeBase64FieldFromSecret(Secret secret, String field) {
+        return Optional.ofNullable(secret)
+                .map(Secret::getData)
+                .map(data -> data.get(field))
+                .map(value -> Base64.getDecoder().decode(value))
+                .orElseThrow(() -> {
+                    String name = Optional.ofNullable(secret)
+                            .map(Secret::getMetadata)
+                            .map(ObjectMeta::getName)
+                            .orElse("unknown");
```
scholzj, line 200: When would it not have a name?

operator-common/src/main/java/io/strimzi/operator/common/Util.java
```diff
@@ -158,26 +185,50 @@ public static File createFileStore(String prefix, String suffix, byte[] bytes) {
...
+                    String namespace = Optional.ofNullable(secret)
+                            .map(Secret::getMetadata)
+                            .map(ObjectMeta::getNamespace)
+                            .orElse("unknown");
```
scholzj, line 204: When would it not have a namespace?

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaReconciler.java
```diff
@@ -155,6 +157,8 @@ public class KafkaReconciler {
     private String loggingHash = "";
     private final Map<Integer, String> brokerConfigurationHash = new HashMap<>();
     private final Map<Integer, String> kafkaServerCertificateHash = new HashMap<>();
+    private PemTrustSet pemTrustSet;
+    private PemAuthIdentity pemAuthIdentity;
```
scholzj, line 161: Same as everywhere else. It should be named based on what it is and not based on the type. Also, why don't we pass these from the outside through the construtor? M(aybe we discussed it on one of the calls, I might not remember anymore :-().

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/ReconcilerUtils.java
```diff
@@ -95,22 +99,68 @@ public static Future<Void> podsReady(Reconciliation reconciliation, PodOperator
...
+     * @return  Future containing the Kubernetes Secret with the Cluster Operator public and private key in both PEM and PKCS12 format
+     */
+    public static CompositeFuture clientAuthIdentity(Reconciliation reconciliation, SecretOperator secretOperator) {
+        Promise<PemAuthIdentity> pemAuthIdentity = Promise.promise();
+        Promise<ClusterOperatorPKCS12AuthIdentity> pkcs12AuthIdentity = Promise.promise();
+        clientAuthIdentitySecret(reconciliation, secretOperator).onSuccess(secret -> {
+            pemAuthIdentity.complete(PemAuthIdentity.clusterOperator(secret));
+            pkcs12AuthIdentity.complete(new ClusterOperatorPKCS12AuthIdentity(secret));
+        });
+        return Future.join(pemAuthIdentity.future(), pkcs12AuthIdentity.future());
```
scholzj, line 148: This segment seems quite confusing. I don't think they should return a CompositeFuture as that is a meaning less. Why not have some record for returning both? Or wh not use single class representing both the trustset and authentication identity? Are they really used separately anywhere?

operator-common/src/main/java/io/strimzi/operator/common/model/PemAuthIdentity.java
```diff
@@ -0,0 +1,109 @@
...
+ * Represents the identity used during TLS client authentication.
+ * This consists of an X509 end-entity certificate, corresponding private key, and a (possibly empty) chain of X509 intermediate CA certificates, all in PEM format.
+ */
+public class PemAuthIdentity {
+
+    private final byte[] privateKeyAsPemBytes;
+    private final byte[] certificateChainAsPemBytes;
+    private final String secretCertName;
+    private String secretName;
+    private String secretNamespace;
```
scholzj, line 26: Why do we need to keep this? We don't care about it when we have the identity, or?

operator-common/src/main/java/io/strimzi/operator/common/model/PemTrustSet.java
```diff
@@ -0,0 +1,115 @@
...
+import java.util.stream.Collectors;
+
+/**
+ * Represents the set of certificates to be trusted by a TLS client or server
+ */
+public class PemTrustSet {
+    private static final String CERT_SUFFIX = ".crt";
+    private final Map<String, byte[]> trustedCertificateMap;
+    private String secretName;
+    private String secretNamespace;
```
scholzj, line 27: Why is this needed?

operator-common/src/main/java/io/strimzi/operator/common/model/PemAuthIdentity.java
```diff
@@ -0,0 +1,109 @@
...
+     * @param secretCertName Key in the Kubernetes Secret that is associated with the requested identity
+     */
+    private PemAuthIdentity(Secret secret, String secretCertName) {
+        this.secretCertName = secretCertName;
+        Optional.ofNullable(secret)
+                .map(Secret::getMetadata)
+                .ifPresent(objectMeta -> {
+                    secretName = objectMeta.getName();
+                    secretNamespace = objectMeta.getNamespace();
+                });
```
scholzj, line 39: Kube resources should always have their metadata. So I think we should not do this just if present if we really need this.

operator-common/src/main/java/io/strimzi/operator/common/model/PemTrustSet.java
```diff
@@ -0,0 +1,115 @@
...
+    public PemTrustSet(Secret secret) {
+        if (secret == null || secret.getData() == null) {
+            throw new RuntimeException("Cannot extract trust set from null secret");
+        }
+        Optional.of(secret)
+                .map(Secret::getMetadata)
+                .ifPresent(objectMeta -> {
+                    secretName = objectMeta.getName();
+                    secretNamespace = objectMeta.getNamespace();
+                });
```
scholzj, line 42: Kube resources should always have their metadata. So I think we should not do this just if present if we really need this.

operator-common/src/test/java/io/strimzi/operator/common/model/PemAuthIdentityTest.java
```diff
@@ -0,0 +1,64 @@
...
+    public static final String NAMESPACE = "testns";
+    public static final String CLUSTER = "testcluster";
+
+    @Test
+    public void testSecretWithMissingKeyThrowsException() {
+        Secret secretWithMissingClusterOperatorKey = new SecretBuilder()
+                .withNewMetadata()
+                .withName(KafkaResources.secretName(CLUSTER))
+                .withNamespace(NAMESPACE)
+                .endMetadata()
```
scholzj, line 28: If you indent it by the new/end sections, it will be much more readable. Here as well as in the ones below.

operator-common/src/test/java/io/strimzi/operator/common/model/PemTrustSetTest.java
```diff
@@ -0,0 +1,81 @@
...
+    public static final String NAMESPACE = "testns";
+    public static final String CLUSTER = "testcluster";
+
+    @Test
+    public void testSecretCorrupted() {
+        Secret secretWithBadCertificate = new SecretBuilder()
+                .withNewMetadata()
+                .withName(KafkaResources.secretName(CLUSTER))
+                .withNamespace(NAMESPACE)
+                .endMetadata()
```
scholzj, line 29: Same as with the previous file

CHANGELOG.md
katheris: Yeah I can remove it if it's not required

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/ClusterOperatorPKCS12AuthIdentity.java
katheris: No motivation other than habit of creating classes and the other classes (PemTrustSet and PemAuthIdentity) don't work as records do they because the getter methods have some logic in them? I did originally have more logic in the constructors for them but as we discussed the other day that caused issues for the tests.

I am happy to change this to a record though if you want

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/CaReconciler.java
katheris: Good point, I'll review the variable names across the board.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaReconciler.java
katheris: I tried to see if I could do the Secret lookup inside KafkaAssemblyOperator and pass them in, but struggled to get the KafkaAssemblyOperator tests to work with that change

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/ReconcilerUtils.java
katheris: Sometimes we use PemAuthIdentity with ClusterOperatorPKCS12AuthIdentity and sometimes with PemTrustSet instead. As discussed elsewhere though I guess we might be able to remove ClusterOperatorPKCS12AuthIdentity if ZK client can be created with only PEM files

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/ClusterOperatorPKCS12AuthIdentity.java
scholzj: I can live with it if records do not work for the other classes you add for this.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaReconciler.java
scholzj: Might be worth opening a follow-up issue?

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/ReconcilerUtils.java
scholzj: > Sometimes we use PemAuthIdentity with ClusterOperatorPKCS12AuthIdentity and sometimes with PemTrustSet instead.

That seems weird -> PemAuthIdentity and ClusterOperatorPKCS12AuthIdentity both represent the user certificate, or?

I would expect you to use everywhere a pair:
* either PemTrustSet and ClusterOperatorPKCS12AuthIdentity
* or PemTrustSet and PemAuthIdentity

That sounds to me like you could merge the PemTrustSet into both of them (and for example have it null if not set in the user operator or something like that).

scholzj: It would simplify the methods with just one param etc.

katheris: Yeah sorry, that's my mistype, it's as you stated:
- either PemTrustSet and ClusterOperatorPKCS12AuthIdentity
- or PemTrustSet and PemAuthIdentity

The other reason for having them separate is if in future we make it optional to have mutual TLS enabled for the operator connection to the Kafka brokers, we might want to only pass PemTrustSet and no AuthIdentity. But for how the code is today I like your suggestion of only passing the AuthIdentity class and having PemTrustSet be part of it, then we could always refactor to pass one or the other in future if we make that change. What do you think?

katheris: Looking at the DefaultAdminClientProvider it makes sense there at least for them to be separate, so we can build the config accordingly, so I also wonder whether it is strange to have to pass to that method `(pemAuthIdentity.trustSet, pemAuthIdentity)`

operator-common/src/main/java/io/strimzi/operator/common/Util.java
katheris: I wrote these methods here based on the method `missingSecretException` that was already in this class. I can create a new exception type if that is the preferred way to handle these kinds of exceptions.

operator-common/src/main/java/io/strimzi/operator/common/Util.java
katheris: If the secret passed into the method is null

operator-common/src/main/java/io/strimzi/operator/common/Util.java
katheris: It's probably a bit over the top, given that I'm assuming if the secret is not null then we can assume the name, namespace and data should also not be null, so can probably change to something like:

```
        if (secret == null) {
            throw new RuntimeException("Cannot fetch data from null secret");
        }
        String data = secret.getData().get(field);
        if (data != null) {
            return Base64.getDecoder().decode(data);
        } else {
            throw Util.missingDataInSecretException(secret.getMetadata().getNamespace(), secret.getMetadata().getName(), field);
        }
```

operator-common/src/main/java/io/strimzi/operator/common/model/PemAuthIdentity.java
katheris: We use it for the exception if we fail to parse the data into a X509Certificate object

operator-common/src/main/java/io/strimzi/operator/common/model/PemTrustSet.java
katheris: We use it for the exceptions if we fail to parse the data somehow later. It's because we lazily parse the data rather than doing it in the constructor, since I was struggling to get the tests to work if we validate upfront.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/ReconcilerUtils.java
scholzj: Well, I guess:
* You can always have one of them set to null (return null) even within a single class with different parts. If you today have everywhere the two parameters, you will anyway end up with one of them being null. So IMHO the question is more `Do you anywhere pass only one of PemTrustSet, PKCS12AuthIdentity or PemAuthIdentity?` If not, if you always pass one of the two pairings, then I think you do not need two parameters. It does not matter that in `DefaultAdminClientProvider` you sometimes expect one of those to be null as it will still be null whether as a direct parameter or in another class, or? 🤷
* I do not think we have any plans to make mTLS optional. The issue is more with making the whole TLS optional (as the whole TLS might happen on a different layer such as service mesh). But I think if we ever make TLS optional, we will likely replace it with some other kind of identity. So having a single class that in the future might be extended for other types of credentials IMHO plays well with that idea and will give us better results than adding more and more parameters to all the methods. E.g. if we use some `TlsPemIdentity` (and `TlsPKCS12Indetity`) that includes both the trust set and the client cert today, we can easily replace it in the future with some`AbstractIdentity` that might take the form of either the `TlsPemIdentity` or the `SpiffeIdentity` or whatever we use. 

In any case, even if you want to keep the parameters separate, the CompositeFuture is IMO not nice for a public method and it should use some record to wrap both of them at least in this place to pass the credentials out of this method.

operator-common/src/main/java/io/strimzi/operator/common/model/PemAuthIdentity.java
scholzj: I guess I wonder if we should cut our losses and simplify this and simply in some cases rely on the stacktrace?

operator-common/src/main/java/io/strimzi/operator/common/Util.java
scholzj: I guess it would be better to check that separately and possibly throw a different error and not mask it behind some `unknown` namespace?

operator-common/src/main/java/io/strimzi/operator/common/Util.java
scholzj: I guess it would be better to check that separately and possibly throw a different error and not mask it behind some `unknown` name?

scholzj: Maybe it suggests something wrong in the design -> you do not know if the secret is required or not at the point where you process it and that forces you to do this kind of weird pre-processing?

operator-common/src/main/java/io/strimzi/operator/common/Util.java
katheris: I've updated this to be simpler and make use of Objects.requireNonNull

operator-common/src/main/java/io/strimzi/operator/common/Util.java
katheris: I've moved these inline

operator-common/src/main/java/io/strimzi/operator/common/Util.java
katheris: I've updated it to do two steps, so it first throws an exception if the secret is null and then if it fails to get the field it can log knowing the secret name and namespace must be present.

operator-common/src/main/java/io/strimzi/operator/common/model/PemTrustSet.java
katheris: I've removed this

operator-common/src/main/java/io/strimzi/operator/common/model/PemAuthIdentity.java
katheris: I've removed this

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/CaReconciler.java
katheris: I've updated the variable names. I've used `coAuthIdentity` as you suggested and then `clusterCaTrustSet` if the variable is used for both ZooKeeper and Kafka clients, and `kafkaCaTrustSet` and `zkCaTrustSet` for the Kafka and ZooKeeper clients respectively, since it is the trustset for the certificate authority that issued and signed the Kafka/ZooKeeper end entity certificate, which happens to be the cluster CA

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/ReconcilerUtils.java
katheris: I've updated the methods in `ReconcilerUtils` to either return one of `PemTrustSet`, `ClusterOperatorPkcs12AuthIdentity` or `PemAuthIdentity`, or to return a new record that contains the two `AuthIdentity` classes.

Do you want me to also create records to be able to pass the `AuthIdentity` and `TrustSet` classes together rather than separately, or shall I leave them separate?

operator-common/src/main/java/io/strimzi/operator/common/model/PemAuthIdentity.java
katheris: I at least found it useful when debugging why the tests failed, so I could see which secret had the corrupted certificate, so I think it is useful

katheris:
@scholzj Thanks for the review, I think I've responded to or addressed all of your comments

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaReconciler.java
katheris: Sure, I can open one, but I guess the question is whether we would actually do it since it would be quite a complicated task refactoring the tests and perhaps for little gain?

CHANGELOG.md
ppatierno: Yes I agree, it doesn't need to be in the CHANGELOG.

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/ClusterOperatorPKCS12AuthIdentity.java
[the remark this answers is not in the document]
ppatierno: I was also wondering why it's not called just `PKCS12AuthIdentity`. Isn't it the PKCS12 counterpart of the `PemAuthIdentity`? Isn't it good for better abstraction removing that `ClusterOperator` prefix from the class name? I see this class hosts `SECRET_KEY` which is specific to the cluster operator but maybe it should be something in a corresponding derived class? So I mean a PKCS12 auth identity class specific for cluster operator (even if we don't have any more in the end). I see something similar in the `PemAuthIdentity` which makes that abstraction by getting a `secretCertName` parameter to know which key to access in the hosted Secret, or?

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/CaReconciler.java
```diff
@@ -369,10 +380,12 @@ Future<Void> rollingUpdateForNewCaKey() {
         }
 
         if (podRollReasons.shouldRestart()) {
+            PemTrustSet clusterCaTrustSet = new PemTrustSet(clusterCa.caCertSecret());
+            PemAuthIdentity coAuthIdentity = PemAuthIdentity.clusterOperator(coSecret);
```
ppatierno, line 384: could be make explicit that it's about TLS auth from the var name? Something like `coTlsAuthIdentity`? just `AuthIdentity` could be anything, even if we know it's about using TLS between CO and the cluster.
Of course, this comment applies to all places where we have something like this as a variable or a parameter in a method.

ppatierno (commented):
Nice work @katheris . I left a couple of comments.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/ReconcilerUtils.java
scholzj: Yeah, I think having a record for it would be useful since that will make it easier to refactor it in the future if needed to support more different credential types etc.

katheris:
@ppatierno @scholzj I believe I've addressed all your comments so feel free to review again

ppatierno (approved):
Thanks for the PR. LGTM! I started the migration pipeline because of the change to the migration utils, just to be sure everything still works.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/CaReconciler.java
```diff
@@ -479,11 +492,12 @@ Future<Void> rollingUpdateForNewCaKey() {
      *
      * @param replicas              Current number of ZooKeeper replicas
      * @param podRestartReasons     List of reasons to restart the pods
+     * @param zkTlsPemIdentity      Trust set and identity for TLS client authentication for connecting to ZooKeeper
      *
      * @return  Future which completes when this step is done either by rolling the ZooKeeper cluster or by deciding
      *          that no rolling is needed.
      */
-    /* test */ Future<Void> maybeRollZookeeper(int replicas, RestartReasons podRestartReasons) {
+    /* test */ Future<Void> maybeRollZookeeper(int replicas, RestartReasons podRestartReasons, TlsPemIdentity zkTlsPemIdentity) {
```
scholzj, line 500: It is not ZooKeeper identity, or? It is the Cluster operator identity?

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/CaReconciler.java
```diff
@@ -369,10 +381,11 @@ Future<Void> rollingUpdateForNewCaKey() {
         }
 
         if (podRollReasons.shouldRestart()) {
+            TlsPemIdentity tlsPemIdentity = new TlsPemIdentity(new PemTrustSet(clusterCa.caCertSecret()), PemAuthIdentity.clusterOperator(coSecret));
```
scholzj, line 384: `coTlsPemIdentity`?

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/CaReconciler.java
```diff
@@ -522,7 +536,7 @@ Future<Void> rollingUpdateForNewCaKey() {
                 });
     }
 
-    /* test */ Future<Void> rollKafkaBrokers(Set<NodeRef> nodes, RestartReasons podRollReasons) {
+    /* test */ Future<Void> rollKafkaBrokers(Set<NodeRef> nodes, RestartReasons podRollReasons, TlsPemIdentity kafkaTlsPemIdentity) {
```
scholzj, line 539: Same as above -> and possibly in other places as well. This is not a Kafka identity -> it is a Cluster operator identity, or?

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KRaftMetadataManager.java
```diff
@@ -63,28 +63,25 @@ public class KRaftMetadataManager {
     public static Future<Void> maybeUpdateMetadataVersion(
             Reconciliation reconciliation,
             Vertx vertx,
-            SecretOperator secretOperator,
+            TlsPemIdentity kafkaTlsPemIdentity,
```
scholzj, line 66: Same as above - I won't repeat the comment for other places -> this should be the identity of who connects. So CO, or?

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaClusterCreator.java
```diff
@@ -180,7 +181,11 @@ private Future<KafkaCluster> brokerRemovalCheck(Kafka kafkaCr, KafkaCluster kafk
             usedToBeBrokersCheckFailed = false;
             return Future.succeededFuture(kafka);
         } else {
-            return brokerScaleDownOperations.brokersInUse(reconciliation, vertx, secretOperator, adminClientProvider)
+            return Future.join(
+                        ReconcilerUtils.clusterCaPemTrustSet(reconciliation, secretOperator),
+                        ReconcilerUtils.coPemAuthIdentity(reconciliation, secretOperator)
+                    )
+                    .compose(res -> brokerScaleDownOperations.brokersInUse(reconciliation, vertx, new TlsPemIdentity(res.resultAt(0), res.resultAt(1)), adminClientProvider))
```
scholzj, line 188: Would it make sense to have this in ReconcilerUtils as a utility method to provide directly the `TlsPemIdentity`? You have this on multiple places.

cluster-operator/src/test/java/io/strimzi/operator/cluster/operator/assembly/KRaftMetadataManagerTest.java
```diff
@@ -46,6 +44,8 @@
 public class KRaftMetadataManagerTest {
     private static Vertx vertx;
 
+    private static final TlsPemIdentity DUMMY_IDENTITY = new TlsPemIdentity(null, null);
```
scholzj, line 47: Should this be in some shared class instead of being created in every class? ALso, if this should be here, it should be above the `vertx` field as it is `final static`.

cluster-operator/src/test/java/io/strimzi/operator/cluster/operator/assembly/KafkaReconcilerStatusTest.java
```diff
@@ -900,5 +897,43 @@ protected Future<Void> listeners()  {
...
         }
+
+        @Override
+        protected Future<Void> initClientAuthenticationCertificates() {
+            kafkaTlsPemIdentity = new TlsPemIdentity(null, null);
+            return Future.succeededFuture();
+        }
+    }
+
+    static class MockKafkaReconcilerFailsWithVersionUpdate extends KafkaReconciler {
```
scholzj, line 908: Why is this change needed? It does not seem related to the PR itself.

cluster-operator/src/test/java/io/strimzi/operator/cluster/operator/resource/ZookeeperLeaderFinderTest.java
```diff
@@ -63,6 +55,7 @@ public class ZookeeperLeaderFinderTest {
     private static Vertx vertx;
     private final SelfSignedCertificate zkCertificate = SelfSignedCertificate.create();
     private final SelfSignedCertificate coCertificate = SelfSignedCertificate.create();
+    private final static TlsPemIdentity DUMMY_IDENTITY = new TlsPemIdentity(null, null);
```
scholzj, line 58: Same as before. If this is needed, you should place it with the other static final fields up top.

cluster-operator/src/test/java/io/strimzi/operator/cluster/operator/resource/ZookeeperLeaderFinderTest.java
```diff
@@ -228,94 +217,21 @@ public void test1PodClusterReturnsOnlyPodAsLeader(VertxTestContext context) {
...
-                assertThat(e, instanceOf(RuntimeException.class));
-                assertThat(e.getMessage(),
-                        is("The Secret testns/testcluster-cluster-operator-certs is missing the key cluster-operator.key"));
-                a.flag();
-            })));
-
-    }
-
-    @Test
-    public void testSecretsCorrupted(VertxTestContext context) {
```
scholzj, line 275: Do we replace these tests on the higher level?

operator-common/src/main/java/io/strimzi/operator/common/auth/PemAuthIdentity.java
```diff
@@ -0,0 +1,107 @@
...
+public class PemAuthIdentity {
+
+    private final byte[] privateKeyAsPemBytes;
+    private final byte[] certificateChainAsPemBytes;
+    private final String secretCertName;
+    private final String secretName;
+    private final String secretNamespace;
+
+    /**
+     * @param secret Kubernetes Secret containing the Cluster Operator public and private key
```
scholzj, line 29: MAybe you could add some description if you add some Javadoc?

operator-common/src/test/java/io/strimzi/operator/common/DefaultAdminClientProviderTest.java
```diff
@@ -121,25 +99,20 @@ public void testMTlsWithPublicCAConnection() {
 
     @Test
     public void testNullConfig() {
-        InvalidConfigurationException ex = assertThrows(InvalidConfigurationException.class, () -> DefaultAdminClientProvider.adminClientConfiguration("my-kafka:9092", null, USER_SECRET, "user", null));
+        InvalidConfigurationException ex = assertThrows(InvalidConfigurationException.class, () -> DefaultAdminClientProvider.adminClientConfiguration("my-kafka:9092", null, mockPemAuthIdentity(), null));
         assertThat(ex.getMessage(), is("The config parameter should not be null"));
     }
 
-    @Test
-    public void testInvalidCASecret() {
```
scholzj, line 129: Do we cover these tests somewhere else?

cluster-operator/src/test/java/io/strimzi/operator/cluster/operator/assembly/KafkaReconcilerStatusTest.java
katheris: Previously the test testKafkaReconcilerStatusDoesNotUpdateVersionOnFailure relied on the clusterId step in the reconciliation failing due to the certificate secrets being missing, so that the reconciliation loop would fail and the test could check a version update did not occur. Now with the certificates being handled differently the mocked secrets call it was relying on doesn't cause a failure. So I updated the test to explicitly add a step that failed. This makes the test less brittle and focuses it on testing the behaviour it's supposed to be testing.

cluster-operator/src/test/java/io/strimzi/operator/cluster/operator/resource/ZookeeperLeaderFinderTest.java
katheris: Yes, they have been moved to PemAuthIdentityTest, since the logic it is testing has moved to PemAuthIdentity.

operator-common/src/main/java/io/strimzi/operator/common/auth/PemAuthIdentity.java
katheris: I've updated the Javadocs so they all have descriptions

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KRaftMetadataManager.java
katheris: I've changed them to coTlsPemIdentity across the board

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaClusterCreator.java
katheris: Yeah I've moved the Pem ones to a ReconcilerUtils method, I've left the PKCS12 rather than have a record of records

katheris:
@scholzj Thanks for the review, I've addressed your comments

cluster-operator/src/test/java/io/strimzi/operator/cluster/operator/assembly/KafkaReconcilerStatusTest.java
scholzj: I don't get this -> the missing secret should still throw some error, or? Or why does it not fail? That seems like a missing validation?

katheris: I updated the test class so that instead of mocking the secret everywhere I provided an override of the `initClientAuthenticationCertificates` that initializes the PemIdentity variable and returns a SucceededFuture. So for the test `testKafkaReconcilerStatusDoesNotUpdateVersionOnFailure` that expects a failure the equivalent change would be to update MockKafkaReconcilerStatusTasks to have an option where the `initClientAuthenticationCertificates` method fails. However, since the test doesn't really have anything to do with `initClientAuthenticationCertificates` and instead is just testing that some failure somewhere in the reconcile loop means the UpdateVersion does not happen it seemed sensible to decouple the two and instead provide a separate MockKafkaReconciler that has an explicit step in the reconcile loop that fails.

katheris: It looks like I actually forgot to take out the other unnecessary secret mocks from this class, so I'll push up an extra commit to remove them

scholzj: Ok, that makes sense I guess. Thanks.

scholzj (approved):

cluster-operator/src/test/java/io/strimzi/operator/cluster/operator/assembly/KafkaReconcilerStatusTest.java
scholzj: > It looks like I actually forgot to take out the other unnecessary secret mocks from this class, so I'll push up an extra commit to remove them

Too late. Sorry :-(

PR #9553, opened 2024-01-13
Add warnings when unused fields are set when Node Pools or Kraft are used

scholzj:
### Type of change

- Enhancement / new feature

### Description

With `UseKRaft` now being enabled by default, users are not required to set the following sections in the Kafka CR:
* `.spec.kafka.replicas`
* `.spec.kafka.storage`
* `.spec.zookeeper`

When Node Pools are used, the storage and replicas configuration in the KAfka CR are ignored. When KRaft is used in addition to Node Pools, the ZooKeeper section is not used either. This PR adds warning conditions when these ignored sections are set in the custom resource to get the users to remove them.

I originally considered setting those in the `KafkaSpecChecker` class. But that adds warning conditions based on the _model_ classes and not based on the custom resources. It would require passing additional parameters through several classes. That seemed like a lot of effort for something that will eventually go away once ZooKeeper support is dropped completely by Kafka and Strimzi. So at the end, I added them as static methods that are called in the place where the validation of these fields happens in the `KafkaAsssemblyOperator`

### Checklist

- [x] Write tests

### DIFF
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KRaftUtils.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KRaftUtils.java
@@ -4,9 +4,12 @@
  */
 package io.strimzi.operator.cluster.model;
 
+import io.strimzi.api.kafka.model.kafka.Kafka;
 import io.strimzi.api.kafka.model.kafka.KafkaSpec;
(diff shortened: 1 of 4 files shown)


ppatierno (approved):
LGTM

see-quick (approved):
👍

im-konge (approved):

PR #9574, opened 2024-01-19
Update Cruise Control system tests

kyguy:
### Type of change

- Refactoring

### Description

Refactors Cruise Control system tests to remove system tests covered by unit tests and other system tests in attempt to reduce unnecessary test failures and resource usage.

### Checklist

- [x] Write tests
- [ ] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/operator-common/src/main/java/io/strimzi/operator/common/model/cruisecontrol/CruiseControlEndpoints.java
+++ b/operator-common/src/main/java/io/strimzi/operator/common/model/cruisecontrol/CruiseControlEndpoints.java
@@ -36,7 +36,12 @@ public enum CruiseControlEndpoints {
     /**
      * Remove broker
      */
-    REMOVE_BROKER("/kafkacruisecontrol/remove_broker");
+    REMOVE_BROKER("/kafkacruisecontrol/remove_broker"),
+
+    /**
+     * metrics
+     */
+    METRICS("/metrics");
 
     private final String path;
 
--- a/operator-common/src/main/java/io/strimzi/operator/common/model/cruisecontrol/CruiseControlEndpoints.java
+++ b/operator-common/src/main/java/io/strimzi/operator/common/model/cruisecontrol/CruiseControlEndpoints.java
@@ -36,12 +36,7 @@ public enum CruiseControlEndpoints {
     /**
      * Remove broker
      */
-    REMOVE_BROKER("/kafkacruisecontrol/remove_broker"),
-
-    /**
-     * metrics
-     */
-    METRICS("/metrics");
+    REMOVE_BROKER("/kafkacruisecontrol/remove_broker");
 
     private final String path;
 
--- a/systemtest/src/test/java/io/strimzi/systemtest/cruisecontrol/CruiseControlApiST.java
+++ b/systemtest/src/test/java/io/strimzi/systemtest/cruisecontrol/CruiseControlApiST.java
@@ -54,7 +54,7 @@ public class CruiseControlApiST extends AbstractST {
 
         LOGGER.info("----> CRUISE CONTROL DEPLOYMENT STATE ENDPOINT <----");
         CruiseControlUtils.ApiResult response = CruiseControlUtils.callApi(testStorage.getNamespaceName(), CruiseControlUtils.HttpMethod.GET,
-                CruiseControlUtils.Scheme.HTTP, CRUISE_CONTROL_DEFAULT_PORT, CruiseControlEndpoints.STATE, "", false);
+                CruiseControlUtils.Scheme.HTTP, CRUISE_CONTROL_DEFAULT_PORT, CruiseControlEndpoints.STATE.toString(), "", false);
         String responseText = response.getResponseText();
         int responseCode = response.getResponseCode();
 
--- a/systemtest/src/test/java/io/strimzi/systemtest/cruisecontrol/CruiseControlApiST.java
+++ b/systemtest/src/test/java/io/strimzi/systemtest/cruisecontrol/CruiseControlApiST.java
@@ -5,10 +5,7 @@
 package io.strimzi.systemtest.cruisecontrol;
 
 import io.strimzi.operator.common.model.cruisecontrol.CruiseControlEndpoints;
-import io.strimzi.operator.common.model.cruisecontrol.CruiseControlUserTaskStatus;
 import io.strimzi.systemtest.AbstractST;
-import io.strimzi.systemtest.Environment;
-import io.strimzi.systemtest.annotations.KRaftWithoutUTONotSupported;
 import io.strimzi.systemtest.annotations.ParallelNamespaceTest;
 import io.strimzi.systemtest.storage.TestStorage;
 import io.strimzi.systemtest.templates.crd.KafkaTemplates;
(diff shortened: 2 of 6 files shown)


systemtest/src/test/java/io/strimzi/systemtest/cruisecontrol/CruiseControlConfigurationST.java
```diff
@@ -256,6 +261,23 @@ void testConfigurationPerformanceOptions(ExtensionContext extensionContext) thro
                 assertThat(containerConfiguration, hasEntry(key, value.toString())));
     }
 
+    @ParallelNamespaceTest
+    void testKafkaRebalanceAutoApprovalMechanism(ExtensionContext extensionContext) {
```
scholzj, line 265: I don't follow why you moved this here. This test seems to not be related to the other tests in this class. It tests the Cruise Control API (which is the KafkaRebalance resource). So it should have stayed where it was?

(Leaving aside that at least some tests in this class look like they should be unit tests and not system tests)

systemtest/src/test/java/io/strimzi/systemtest/cruisecontrol/CruiseControlApiST.java
```diff
@@ -52,81 +47,59 @@ void testCruiseControlBasicAPIRequests(ExtensionContext extensionContext)  {
...
-        String response = CruiseControlUtils.callApi(testStorage.getNamespaceName(), CruiseControlUtils.SupportedHttpMethods.POST, CruiseControlEndpoints.STATE,  CruiseControlUtils.SupportedSchemes.HTTPS, true);
-
-        assertThat(response, is("Unrecognized endpoint in request '/state'\n" +
-            "Supported POST endpoints: [ADD_BROKER, REMOVE_BROKER, FIX_OFFLINE_REPLICAS, REBALANCE, STOP_PROPOSAL_EXECUTION, PAUSE_SAMPLING, " +
-                "RESUME_SAMPLING, DEMOTE_BROKER, ADMIN, REVIEW, TOPIC_CONFIGURATION, RIGHTSIZE, REMOVE_DISKS]\n"));
-
-        response = CruiseControlUtils.callApi(testStorage.getNamespaceName(), CruiseControlUtils.SupportedHttpMethods.GET, CruiseControlEndpoints.STATE,  CruiseControlUtils.SupportedSchemes.HTTPS, true);
-
         LOGGER.info("Verifying that {} REST API is available", CRUISE_CONTROL_NAME);
+        CruiseControlUtils.ApiResult response = CruiseControlUtils.callApi(testStorage.getNamespaceName(), CruiseControlUtils.HttpMethod.GET, CruiseControlEndpoints.STATE);
```
scholzj, line 51: So, what exactly does this whole test test? And why is it not suported in KRaft mode? I guess that question applies to more of the tests in this class.

systemtest/src/test/java/io/strimzi/systemtest/cruisecontrol/CruiseControlConfigurationST.java
kyguy: > I don't follow why you moved this here. This test seems to not be related to the other tests in this class. It tests the Cruise Control API (which is the KafkaRebalance resource). So it should have stayed where it was?

Whoops, I meant to move that in the `CruiseControlST` class. It seemed that the `CruiseControlST` class has tests related to the `KafkaRebalance` resource where as the `CruiseControlApiST` class seemed to contain tests for hitting/testing the CC API directly.

> (Leaving aside that at least some tests in this class look like they should be unit tests and not system tests)

I'll investigate this class to to see what can be removed

systemtest/src/test/java/io/strimzi/systemtest/cruisecontrol/CruiseControlApiST.java
kyguy: > So, what exactly does this whole test test?

This is just to test basic CC API functionality, that you can get the state, execute rebalances, stop proposal execution, get user task list containing the previously executed tasks

> And why is it not suported in KRaft mode?

Appears to be an oversight, let me clean that up

systemtest/src/test/java/io/strimzi/systemtest/cruisecontrol/CruiseControlConfigurationST.java
scholzj: Yeah, that makes sense I guess.

systemtest/src/test/java/io/strimzi/systemtest/cruisecontrol/CruiseControlApiST.java
scholzj: Do we really need to test these? I would expect that those are:
* Tested by Cruise Controls own tests
* Used in the KafkaRebalance tests

So I would expect here more or less just some check that we _unlocked_ the REST API. But not detailed tests of different CC fetures.

kyguy: > So I would expect here more or less just some check that we unlocked the REST API. But not detailed tests of different CC fetures.

Yeah that sounds fair to me, I'll take a stab at it with the next commit

systemtest/src/test/java/io/strimzi/systemtest/cruisecontrol/CruiseControlST.java
```diff
@@ -489,6 +490,23 @@ void testCruiseControlDuringBrokerScaleUpAndDown(ExtensionContext extensionConte
...
+        final TestStorage testStorage = new TestStorage(extensionContext);
+
+        resourceManager.createResourceWithWait(extensionContext, KafkaTemplates.kafkaWithCruiseControl(testStorage.getClusterName(), 3, 3).build());
+
+        // KafkaRebalance with auto-approval
+        resourceManager.createResourceWithWait(extensionContext, KafkaRebalanceTemplates.kafkaRebalance(testStorage.getClusterName())
+                .editMetadata()
+                .addToAnnotations(Annotations.ANNO_STRIMZI_IO_REBALANCE_AUTOAPPROVAL, "true")
+                .endMetadata()
+                .build());
```
im-konge, line 504:
```suggestion
        resourceManager.createResourceWithWait(extensionContext, KafkaRebalanceTemplates.kafkaRebalance(testStorage.getClusterName())
            .editMetadata()
                .addToAnnotations(Annotations.ANNO_STRIMZI_IO_REBALANCE_AUTOAPPROVAL, "true")
            .endMetadata()
            .build()
        );
```

im-konge (approved):
Thanks for the clean-up, I have just one nit in indent.

systemtest/src/test/java/io/strimzi/systemtest/cruisecontrol/CruiseControlConfigurationST.java
```diff
@@ -62,158 +51,29 @@ public class CruiseControlConfigurationST extends AbstractST {
     private static final Logger LOGGER = LogManager.getLogger(CruiseControlConfigurationST.class);
 
     @ParallelNamespaceTest
-    void testDeployAndUnDeployCruiseControl(ExtensionContext extensionContext) throws IOException {
+    void testMetricReporterConfigurationInKafka(ExtensionContext extensionContext) throws IOException {
```
scholzj, line 54: This test seems to be failing. Also, I'm not sure why you made the changes you did here. 

The original code seemed to be at least checking that the CC pod is removed and that rolling updates to do it were properly happening here. You now just check a config mpa with generated configuration. That is done in the unit tests already and does not need a system test. So we should either reintroduce the original logic or we can remove it completely.

operator-common/src/main/java/io/strimzi/operator/common/model/cruisecontrol/CruiseControlEndpoints.java
```diff
@@ -36,7 +36,12 @@ public enum CruiseControlEndpoints {
     /**
      * Remove broker
      */
-    REMOVE_BROKER("/kafkacruisecontrol/remove_broker");
+    REMOVE_BROKER("/kafkacruisecontrol/remove_broker"),
+
+    /**
+     * metrics
+     */
+    METRICS("/metrics");
```
scholzj, line 44: `metrics` => `Metrics` if we want to keep this. But I'm not sure it is worth to pull this into a production code when it is relevant only to tests.

kyguy: Removed

systemtest/src/test/java/io/strimzi/systemtest/cruisecontrol/CruiseControlConfigurationST.java
kyguy: Reverted to how it was before

scholzj (approved):

scholzj:
Thanks @kyguy

PR #9580, opened 2024-01-23
[ST] ZK to KRaft migration - migration process

im-konge:
### Type of change

- Enhancement

### Description

This PR adds ST for migration process from ZK to KRaft - the "happy path". It goes through the whole process, also during the test the continuous message transmission is done.
The last check - wait for continuous clients to be finished successfully - is currently disabled because of some issue on test-clients side -> there are one or two messages that are not correctly send (due to disconnection from leader), but I have some list of enhancements that I can do on the clients side, so this will not happen (this fail sometimes happen in upgrade/downgrade tests as well).

### Checklist

- [x] Write tests


systemtest/src/test/java/io/strimzi/systemtest/migration/MigrationST.java
```diff
@@ -0,0 +1,265 @@
...
+            .withTopicName(kraftTopicName)
+            .build();
+
+        resourceManager.createResourceWithWait(extensionContext,
+            immediateClients.producerTlsStrimzi(testStorage.getClusterName()),
+            immediateClients.consumerTlsStrimzi(testStorage.getClusterName())
+        );
+
+        ClientUtils.waitForClientsSuccess(testStorage.getProducerName(), testStorage.getConsumerName(), testStorage.getNamespaceName(), testStorage.getMessageCount());
+        // TODO: continuous clients are failing due to missing messages (one or two) during the whole process. This happens from time to time in upgrade/downgrade tests as well
```
see-quick, line 236: I am curious why they are failing here? :)

see-quick (commented):
Looks good to me. Just one nit to consider 👍

systemtest/src/test/java/io/strimzi/systemtest/migration/MigrationST.java
```diff
@@ -0,0 +1,265 @@
...
+
+        LOGGER.info("1. applying the {} annotation with value: {}", Annotations.ANNO_STRIMZI_IO_KRAFT, "migration");
+        KafkaResource.replaceKafkaResourceInSpecificNamespace(testStorage.getClusterName(),
+            kafka -> kafka.getMetadata().getAnnotations().put(Annotations.ANNO_STRIMZI_IO_KRAFT, "migration"), testStorage.getNamespaceName());
+
+        LOGGER.info("2. waiting for controller Pods to be up and running");
+        PodUtils.waitForPodsReady(testStorage.getNamespaceName(), controllerSelector, 3, true);
+        Map<String, String> controllerPodsSnapshot = PodUtils.podSnapshot(testStorage.getNamespaceName(), controllerSelector);
+
+        LOGGER.info("3. waiting for first rolling update of broker Pods");
```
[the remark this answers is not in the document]
im-konge: Getting info from log would cause race conditions FMPOV (the log doesn't contain info in time, it doesn't match the pattern, or whatever). So having a comment here why it's rolling would be better.

systemtest/src/test/java/io/strimzi/systemtest/migration/MigrationST.java
im-konge: I guess that's because in the moment the leader is rolled, the particular message is send - and the producer fails to send the message. IIRC I was trying to fix it for upgrade/downgrade as well - because it seems that there can be issue with configuration of the clients (that's what I remember from discussion with Tom Bentley, but it's quite a long time). So I'm disabling this check until I find the issue (and fix it).

ppatierno: I don't know what "failing" here means but take into account that during brokers rolling you can expect transient errors, producer not able to send, consumer not able to receive (i.e. disconnection, leader partition changing, and more). But it has to recover every time as it happened with my manual tests. No messages are lost.

im-konge: Yeah that's what I actually mean, the producer in our test-clients just tries to send it, it catches some kind of exception and automatically decrease the number of successfully sent messages. So I need to look at this and fix that before enabling this check. Thanks Paolo

systemtest/src/test/java/io/strimzi/systemtest/migration/MigrationST.java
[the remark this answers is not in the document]
see-quick: Cool

systemtest/src/test/java/io/strimzi/systemtest/migration/MigrationST.java
```diff
@@ -0,0 +1,265 @@
...
+     *  19. - Does immediate message transmission to the new KafkaTopic
+     *  20. - Waits until continuous clients are finished successfully
+     *
+     * @usecase
+     *  - zk-to-kraft-migration
+     */
+    @IsolatedTest
+    @SuppressWarnings("checkstyle:MethodLength")
+    void testMigrationFromZkToKRaft(ExtensionContext extensionContext) {
+        TestStorage testStorage = new TestStorage(extensionContext, TestConstants.CO_NAMESPACE);
```
henryZrncik, line 82: i guess that now all resources with end up in `co-namespace` and not  `test-suite-namespace` ?  is this desired ?

systemtest/src/test/java/io/strimzi/systemtest/migration/MigrationST.java
```diff
@@ -0,0 +1,265 @@
...
+
+import java.util.Map;
+
+import static io.strimzi.systemtest.TestConstants.REGRESSION;
+import static io.strimzi.systemtest.resources.ResourceManager.cmdKubeClient;
+import static io.strimzi.systemtest.resources.ResourceManager.kubeClient;
+import static org.hamcrest.CoreMatchers.is;
+import static org.hamcrest.MatcherAssert.assertThat;
+
+@Tag(REGRESSION)
```
henryZrncik, line 42: assuming that this is part of regression should test require precondition of having expected "feature gates" set ? if we run this test with `-KafkaNodePools,-UnidirectionalTopicOperator,-UseKraft` it would fail probably on sanity check ?

systemtest/src/test/java/io/strimzi/systemtest/migration/MigrationST.java
```diff
@@ -0,0 +1,265 @@
...
+            immediateClients.producerTlsStrimzi(testStorage.getClusterName()),
+            immediateClients.consumerTlsStrimzi(testStorage.getClusterName())
+        );
+
+        ClientUtils.waitForClientsSuccess(testStorage.getProducerName(), testStorage.getConsumerName(), testStorage.getNamespaceName(), testStorage.getMessageCount());
+        // TODO: continuous clients are failing due to missing messages (one or two) during the whole process. This happens from time to time in upgrade/downgrade tests as well
+        // it needs to be fixed on test-clients side
+        //  LOGGER.info("8. migration is completed, waiting for continuous clients to finish");
+        //  ClientUtils.waitForClientsSuccess(continuousProducerName, continuousConsumerName, testStorage.getNamespaceName(), continuousMessageCount);
+    }
```
henryZrncik, line 240: Assuming migration went as expected, do we expect Zookeeper `PodSet` and its `Pods` to be removed automatically ?

ppatierno: ZooKeeper removal is not in this PR yet. It's in the oven ... I am cooking it and will be here soon ;-)

systemtest/src/test/java/io/strimzi/systemtest/migration/MigrationST.java
im-konge: This will not change the way, how it is already implemented. By default, everything is installed in `co-namespace` in case of `@IsolatedTest`. But I don't have an answer why it's not installed in `test-suite-namespace` 🤷‍♂️ 
The `TestStorage` will hold there just the value, so the test checks will look into the right namespace.

This behavior can be investigated as part of a different PR

systemtest/src/test/java/io/strimzi/systemtest/migration/MigrationST.java
im-konge: Yeah I wanted to add assumption there. But I also think that it should not really be run just as part of regression 🤔 I mean the tests take ~15-20 minutes, so in case that there will be more than these two (and there will be more), I guess we can move it under something else, so it will not add additional execution time to regression pipelines.

im-konge: Yeah I was thinking about it as well
But in that case, I will have to add it somewhere in AZP or create a dedicated pipeline, which IDK is worth it or not 🤔

systemtest/src/test/java/io/strimzi/systemtest/migration/MigrationST.java
```diff
@@ -0,0 +1,284 @@
...
+        waitForZooKeeperResourcesDeletion(testStorage);
+
+        LOGGER.info("8. Everything related to ZK is deleted, waiting until .status.kafkaMetadataState in Kafka will contain KRaft state");
+
+        KafkaUtils.waitUntilKafkaStatusContainsKafkaMetadataState(testStorage.getNamespaceName(), testStorage.getClusterName(), KafkaMetadataState.KRaft.name());
+
+        LOGGER.info("9. removing LMFV and IBPV from Kafka config -> brokers and controllers should be rolled");
+        KafkaResource.replaceKafkaResourceInSpecificNamespace(testStorage.getClusterName(), kafka -> {
+            kafka.getSpec().getKafka().getConfig().remove("log.message.format.version");
+            kafka.getSpec().getKafka().getConfig().remove("inter.broker.protocol.version");
```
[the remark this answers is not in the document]
im-konge: It's by default encapsulated inside `KafkaTemplates.kafkaPersistent()`, so even if it's not visible for the first sight in the test, it's there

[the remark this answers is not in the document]
ppatierno: Ah ok, thanks! Maybe adding just a comment to explain that they are there. Not in the logged text, just a comment in the code.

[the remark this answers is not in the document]
im-konge: Will do that, yeah

see-quick (approved):

ppatierno (approved):
LGTM! Good first STs!

PR #9585, opened 2024-01-24
Allow reconciliation to continue when scale-down is blocked due to brokers still in use

### LINKED ISSUE #9232: Revert blocked scale-down and continue reconciliation
https://github.com/strimzi/strimzi-kafka-operator/pull/9042 added check to prevent Strimzi scaling down a Kafkabroker that is not empty. It does so by failing the reconciliation and expecting the user to address the issue. We should try to do it in a nicer => revert the scale down internally and proceed with the reconciliation with only warnings being issued tot he user.

scholzj:
### Type of change

- Enhancement / new feature

### Description

Currently, when a user tries to scale down the Kafka cluster but the brokers that would be removed are still in use, we just throw an exception and fail the reconciliation. This works fine in general. But it makes it hard to build any automation on top of it since it essentially blocks any other operation tasks. So for example, if use by mistake scales down the Kafka cluster, the operator starts throwing errors and will not proceed with for example renewing certificates.

This PR tries to address it by reverting the scale-down changes and allowing the reconciliation to proceed as usual. It takes a pragmatic approach and reverts all scale-downs even if some of the brokers would be empty and it would be possible to shut them down. Since Kafka has no fencing, this could mean the brokers that were empty get new topics assigned. But it helps to simplify the logic significantly because it does not need to deal with situations such as when user tries to scale down brokers 3 and 4 where broker 3 is empty and broker 4 is still in use. In such case, we cannot simply correct the replication factor by 1 as that would mean broker 3 (the empty one) is kept and broker 4 is removed => so it will be still blocked with an exception.

From the technical perspective, this PR introduces a new class `KafkaClusterCreator` that contains the logic. This class is used to:
* Create the Kafka cluster model
* Check if it has any scale-down issues
* If it has any scale-down issues, it tries to fix them and recursively creates the Kafka cluster model again (but this time, any scale-down issues are just raised as an error -> in case they were not fixed for the first time)
* It is written in a way to make it easier to extend it in the next steps to check against removing the broker role and possibly other checks (we might consider moving the storage checks to it from the `KafkaCluster.fromCrd(...)` method).

The `KafkaClusterCreator` is used by the `KafkaAssemblyOperator` and the created `KafkaCluster` instance is used to create the `KafkaRecocniler`. I considered some other options as well, but rejected them at the end:
* Doing it in the constructor of `KafkaReconciler` => I rejected it because it seemed to push a lot of logic into the constructor and also had issues with asynchronous code. The advantage of the current approach is that it keeps the constructor clean with only assignments of parameters to object fields.
* Doing it as the first step of the `KafkaReconciler`. This would work quite nicely. But it would mean that the `KafkaCluster` object is not final in the `KafkaReconciler` class. That seemed like something that can be tricky to maintain in the future and is something to avoid.
* Keeping the original `KafkaCluster` object and just modifying the node assignments inside it. This would probably be easier for the initial implementation. But it might be harder to maintain as it would mean the fields inside the `KafkaCluster` will not be final, will be subject to change and we will need to be careful what parts are changed and where and when are the used. Creating a brand new `KafkaCluster` object seemed like a better approach.

_This is currently a draft sued for a preliminary review of the code. It does not contain any new tests and it does not update existing tests to the new code._

This should resolve #9232.

### Checklist

- [ ] Write tests
- [ ] Update documentation
- [ ] Update CHANGELOG.md


scholzj:
@ppatierno Would you mind having a quick look if the solution for this makes sense to you before I write all the tests etc.?

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaClusterCreator.java
```diff
@@ -0,0 +1,267 @@
...
+     *
+     * @return  New Kafka Cluster instance
+     */
+    public Future<KafkaCluster> prepareKafkaCluster(
+            Kafka kafkaCr,
+            List<KafkaNodePool> nodePools,
+            Map<String, Storage> oldStorage,
+            Map<String, List<String>> currentPods,
+            KafkaVersionChange versionChange,
+            boolean tryToFixProblems
```
ppatierno, line 99: why this flag? You envisage scenarios in the code where we don't want to fix problems? (Asking because you are already using this method but of course passing true here).

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaClusterCreator.java
```diff
@@ -0,0 +1,267 @@
...
+                        .endKafka()
+                    .endSpec()
+                    .build();
+
+            return Future.succeededFuture(new KafkaAndNodePools(newKafkaCr, nodePools));
+        } else {
+            // Node pools are used -> we have to fix scale down in the KafkaNodePools
+            List<KafkaNodePool> newNodePools = new ArrayList<>();
+
+            for (KafkaNodePool nodePool : nodePools)    {
```
ppatierno, line 195: could it make sense the nodepools with `controller` role only?

ppatierno (commented):
The overall logic seems to be reasonable to me. I left a couple of questions.
I will start crying about conflicts I see with my migration PR later ...

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaClusterCreator.java
scholzj: Because it is called recursively -> in the first call, we try to fix things. In the second, if there are any errors we failed to fix, it will just the the error and fail the reconciliation. This flag is basically used to indicate which behavior to activate.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaClusterCreator.java
scholzj: Why would it make sense for controller role only? This is something you typically want to do for brokers. But the other changes we want to catch in the future - such as the role changes or possibly storage - would apply to all node pools.

ppatierno: sorry ... I think a missed a term in my question ... I meant "skipping the node pools with controller(s) only" because they don't host topics which impact the scale down.
But I guess you already answered to it about other possible future checks.

scholzj: I guess that can be done, yeah.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaClusterCreator.java
```diff
@@ -0,0 +1,267 @@
...
+            Map<String, Storage> oldStorage,
+            Map<String, List<String>> currentPods,
+            KafkaVersionChange versionChange,
+            boolean tryToFixProblems
+    )   {
+        return createKafkaCluster(kafkaCr, nodePools, oldStorage, currentPods, versionChange)
+                .compose(kafka -> brokerScaleDownCheck(kafkaCr, kafka))
+                .compose(kafka -> {
+                    if (scaleDownCheckFailed && tryToFixProblems)   {
+                        // We have a failure, and should try to fix issues
```
ppatierno, line 105: isn't this comment inconsistent with `false` you are passing later on to the prepare? Or am I missing the meaning of it?

scholzj: It is first called with `tryToFixProblems` set to `true` when called from `KafkaAssemblyOperator`. If we run into any issues in the first run, we try to fix them and call it again, but this time with `tryToFixProblems` set to `false` so that we recreate the `KafkaCluster` object. But if it is still failing, then we just throw up the exception -> so that if there is some issue we cannot fix we don't loop forever.

ppatierno: I think I misleading the comment. It was referring to the "if" branch we are in (so as you said try fix problems is true) while I was looking at how prepare is called later passing false (so you are not trying to fix problems).
Thanks.

scholzj: I will add a comment to make it more clear.

ppatierno (approved):
LGTM. I saw multiple TODO which should be removed.

see-quick (approved):

im-konge (approved):
LGTM, thanks

PR #9589, opened 2024-01-24
[ST] ZK to KRaft migration - rollback process

im-konge:
### Type of change

- New ST

### Description

This PR adds ST for second part of the migration - rollback process. After first part of the migration is finished (we are in the KRaftPostMigration state), we annotate the Kafka resource with `strimzi.io/kraft:rollback` to roll-back to the ZK mode.
The changes are mostly about splitting the parts of the first test into multiple methods, so they can be reusable by other tests (also for those that will come in next PR). The main change is the `doRollback()` method, which executes the whole roll-back process.

### Checklist

- [ ] Write tests

### DIFF
--- a/systemtest/src/main/java/io/strimzi/systemtest/utils/kafkaUtils/KafkaUtils.java
+++ b/systemtest/src/main/java/io/strimzi/systemtest/utils/kafkaUtils/KafkaUtils.java
@@ -594,6 +594,6 @@ public class KafkaUtils {
 
     public static String getKafkaLogFolderNameInPod(String namespaceName, String kafkaPodName) {
         return ResourceManager.cmdKubeClient().namespace(namespaceName)
-            .execInPod(kafkaPodName, "/bin/bash", "-c", "ls /var/lib/kafka/data | grep \"kafka-log[0-9]\" -o").out().trim();
+            .execInPod(kafkaPodName, "/bin/bash", "-c", "ls /var/lib/kafka/data | grep \"kafka-log[0-9]\\+\" -o").out().trim();
     }
 }
--- a/systemtest/src/main/java/io/strimzi/systemtest/utils/kafkaUtils/KafkaUtils.java
+++ b/systemtest/src/main/java/io/strimzi/systemtest/utils/kafkaUtils/KafkaUtils.java
@@ -591,4 +591,9 @@ public class KafkaUtils {
             return k.getStatus().getKafkaMetadataState().equals(desiredKafkaMetadataState);
         });
     }
+
+    public static String getKafkaLogFolderNameInPod(String namespaceName, String kafkaPodName) {
+        return ResourceManager.cmdKubeClient().namespace(namespaceName)
+            .execInPod(kafkaPodName, "/bin/bash", "-c", "ls /var/lib/kafka/data | grep \"kafka-log[0-9]\" -o").out().trim();
+    }
 }
--- a/systemtest/src/test/java/io/strimzi/systemtest/migration/MigrationST.java
+++ b/systemtest/src/test/java/io/strimzi/systemtest/migration/MigrationST.java
@@ -369,7 +369,7 @@ public class MigrationST extends AbstractST {
 
     private void assertThatTopicIsPresentInKRaftMetadata(String namespaceName, LabelSelector controllerSelector, String topicName) {
         String controllerPodName = kubeClient().namespace(namespaceName).listPods(controllerSelector).get(0).getMetadata().getName();
-        String kafkaLogDirName = cmdKubeClient().namespace(namespaceName).execInPod(controllerPodName, "/bin/bash", "-c", "ls /var/lib/kafka/data | grep \"kafka-log[0-9]\\+\" -o").out().trim();
+        String kafkaLogDirName = KafkaUtils.getKafkaLogFolderNameInPod(namespaceName, controllerPodName);
         String commandOutput = cmdKubeClient().namespace(namespaceName).execInPod(controllerPodName, "/bin/bash", "-c", "./bin/kafka-dump-log.sh --cluster-metadata-decoder --skip-record-metadata" +
             " --files /var/lib/kafka/data/" + kafkaLogDirName + "/__cluster_metadata-0/00000000000000000000.log | grep " + topicName).out().trim();
 
--- a/systemtest/src/test/java/io/strimzi/systemtest/migration/MigrationST.java
+++ b/systemtest/src/test/java/io/strimzi/systemtest/migration/MigrationST.java
@@ -259,7 +259,7 @@ public class MigrationST extends AbstractST {
     }
 
     private void doSecondPartOfMigration(ExtensionContext extensionContext, TestStorage testStorage) {
-        LOGGER.info("9. finishing migration - applying the {} annotation with value: {}, controllers should be rolled", Annotations.ANNO_STRIMZI_IO_KRAFT, "enabled");
+        LOGGER.info("Finishing migration - applying the {} annotation with value: {}, controllers should be rolled", Annotations.ANNO_STRIMZI_IO_KRAFT, "enabled");
         KafkaResource.replaceKafkaResourceInSpecificNamespace(testStorage.getClusterName(),
             kafka -> kafka.getMetadata().getAnnotations().put(Annotations.ANNO_STRIMZI_IO_KRAFT, "enabled"), testStorage.getNamespaceName());
 
@@ -290,9 +290,12 @@ public class MigrationST extends AbstractST {
     }
 
     private void doRollback(TestStorage testStorage) {
+        LOGGER.info("Checking that __cluster_metadata topic does exist in Kafka Brokers");
+        assertThatClusterMetadataTopicPresentInBrokerPod(testStorage.getNamespaceName(), brokerSelector, true);
+
         LOGGER.info("From {} state we are going to roll back to ZK", KafkaMetadataState.KRaftPostMigration.name());
 
-        LOGGER.info("Rolling migration back - applying {}: rollback annotation", Annotations.ANNO_STRIMZI_IO_KRAFT);
+        LOGGER.info("Rolling migration back - applying the {} annotation with value: {}", Annotations.ANNO_STRIMZI_IO_KRAFT, "rollback");
         KafkaResource.replaceKafkaResourceInSpecificNamespace(testStorage.getClusterName(),
             kafka -> kafka.getMetadata().getAnnotations().put(Annotations.ANNO_STRIMZI_IO_KRAFT, "rollback"), testStorage.getNamespaceName());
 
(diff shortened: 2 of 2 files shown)


systemtest/src/test/java/io/strimzi/systemtest/migration/MigrationST.java
```diff
@@ -0,0 +1,486 @@
...
+
+        LOGGER.info("10. applying {}: disabled annotation", Annotations.ANNO_STRIMZI_IO_KRAFT);
+        KafkaResource.replaceKafkaResourceInSpecificNamespace(testStorage.getClusterName(),
+            kafka -> kafka.getMetadata().getAnnotations().put(Annotations.ANNO_STRIMZI_IO_KRAFT, "disabled"), testStorage.getNamespaceName());
+
+        LOGGER.info("11. waiting for Broker Pods to be rolled");
+        brokerPodsSnapshot = RollingUpdateUtils.waitTillComponentHasRolledAndPodsReady(testStorage.getNamespaceName(), brokerSelector, 3, brokerPodsSnapshot);
+
+        LOGGER.info("12. waiting until .status.kafkaMetadataState contains {} state", KafkaMetadataState.ZooKeeper.name());
+        KafkaUtils.waitUntilKafkaStatusContainsKafkaMetadataState(testStorage.getNamespaceName(), testStorage.getClusterName(), KafkaMetadataState.ZooKeeper.name());
```
ppatierno, line 431: one thing we should test ...
During migration, when you reach the `KRaftPostMigration` phase, the brokers become so called "observers" and they get a copy of the `__cluster_metadata` topic from the controllers. It's something you could test that this topic exist during the migration as well.
When you rollback and come back to ZooKeeper, we should test that this topic doesn't exist anymore.
The presence should be tested to check the file system of the Kafka broker pod (not sure you can via Kafka CLI tools as it's an internal topic but it's more internal than the __consumer_offset one :-D).

im-konge: I will check that, thanks for the suggestion :)

systemtest/src/test/java/io/strimzi/systemtest/migration/MigrationST.java
```diff
@@ -239,17 +283,64 @@ void testMigrationFromZkToKRaft(ExtensionContext extensionContext) {
...
+        createKafkaTopicAndCheckMetadataWithMessageTransmission(extensionContext, testStorage, kraftTopicName, false);
+
+        LOGGER.info("Migration is completed, waiting for continuous clients to finish");
+        ClientUtils.waitForClientsSuccess(continuousClients.getProducerName(), continuousClients.getConsumerName(), testStorage.getNamespaceName(), continuousClients.getMessageCount());
+    }
+
+    private void doRollback(TestStorage testStorage) {
+        LOGGER.info("From {} state we are going to roll back to ZK", KafkaMetadataState.KRaftPostMigration.name());
+
+        LOGGER.info("Rolling migration back - applying {}: rollback annotation", Annotations.ANNO_STRIMZI_IO_KRAFT);
```
ppatierno, line 295: aren't we using something like `applying the {} annotation with value: {}` in other places?

systemtest/src/test/java/io/strimzi/systemtest/migration/MigrationST.java
```diff
@@ -239,17 +283,64 @@ void testMigrationFromZkToKRaft(ExtensionContext extensionContext) {
...
-        assertThatTopicIsPresentInKRaftMetadata(controllerSelector, kraftTopicName);
+        LOGGER.info("Waiting until .status.kafkaMetadataState contains {} state", KafkaMetadataState.KRaftDualWriting.name());
+        KafkaUtils.waitUntilKafkaStatusContainsKafkaMetadataState(testStorage.getNamespaceName(), testStorage.getClusterName(), KafkaMetadataState.KRaftDualWriting.name());
 
-        LOGGER.info("13b. checking if we are able to do a message transmission on KafkaTopic: {}", kraftTopicName);
+        LOGGER.info("Deleting Controller's NodePool");
+        KafkaNodePool controllerPool = KafkaNodePoolResource.kafkaNodePoolClient().inNamespace(testStorage.getNamespaceName()).withName(testStorage.getControllerPoolName()).get();
+        resourceManager.deleteResource(controllerPool);
+
+        LOGGER.info("Applying {}: disabled annotation", Annotations.ANNO_STRIMZI_IO_KRAFT);
```
ppatierno, line 309: ditto as above

systemtest/src/test/java/io/strimzi/systemtest/migration/MigrationST.java
```diff
@@ -239,17 +283,64 @@ void testMigrationFromZkToKRaft(ExtensionContext extensionContext) {
...
+            kafka -> kafka.getMetadata().getAnnotations().put(Annotations.ANNO_STRIMZI_IO_KRAFT, "disabled"), testStorage.getNamespaceName());
+
+        LOGGER.info("Waiting for Broker Pods to be rolled - rolling back from DualWrite mode to Zookeeper");
+        brokerPodsSnapshot = RollingUpdateUtils.waitTillComponentHasRolledAndPodsReady(testStorage.getNamespaceName(), brokerSelector, 3, brokerPodsSnapshot);
+
+        LOGGER.info("Waiting until .status.kafkaMetadataState contains {} state", KafkaMetadataState.ZooKeeper.name());
+        KafkaUtils.waitUntilKafkaStatusContainsKafkaMetadataState(testStorage.getNamespaceName(), testStorage.getClusterName(), KafkaMetadataState.ZooKeeper.name());
+
+        LOGGER.info("Checking that __cluster_metadata topic does not exist in Kafka Brokers");
+        assertThatClusterMetadataTopicNotPresentInBrokerPod(testStorage.getNamespaceName(), brokerSelector);
```
ppatierno, line 320: while it's important to check this here, what do you think about checking the opposite (that the __cluster_metadata topic exists) but at the beginning of the test, I mean in post-migration before running the rollback?

systemtest/src/test/java/io/strimzi/systemtest/migration/MigrationST.java
```diff
@@ -258,33 +349,43 @@ void testMigrationFromZkToKRaft(ExtensionContext extensionContext) {
...
+    private void assertThatTopicIsPresentInKRaftMetadata(String namespaceName, LabelSelector controllerSelector, String topicName) {
+        String controllerPodName = kubeClient().namespace(namespaceName).listPods(controllerSelector).get(0).getMetadata().getName();
+        String kafkaLogDirName = cmdKubeClient().namespace(namespaceName).execInPod(controllerPodName, "/bin/bash", "-c", "ls /var/lib/kafka/data | grep \"kafka-log[0-9]\" -o").out().trim();
+        String commandOutput = cmdKubeClient().namespace(namespaceName).execInPod(controllerPodName, "/bin/bash", "-c", "./bin/kafka-dump-log.sh --cluster-metadata-decoder --skip-record-metadata" +
             " --files /var/lib/kafka/data/" + kafkaLogDirName + "/__cluster_metadata-0/00000000000000000000.log | grep " + topicName).out().trim();
 
         assertThat(String.join("KafkaTopic: %s is not present KRaft metadata", topicName), commandOutput.contains(topicName), is(true));
     }
 
+    private void assertThatClusterMetadataTopicNotPresentInBrokerPod(String namespaceName, LabelSelector brokerSelector) {
```
ppatierno, line 376: if you get my previous comment, I would make this method more generic to accept a boolean and checking if the metadata topic exist or not depending on that parameter value.

systemtest/src/test/java/io/strimzi/systemtest/migration/MigrationST.java
im-konge: Using that file in the Kafka folder we are checking that the newly created topic is there (that's the first step of the migration -> when the metadata state is in KRaftPostMigration). That means that we _know_ that the file is there. But if you think that it will be worth adding it, I will do that -> we are anyway checking the presence of the topic inside the file in controller Pod

systemtest/src/main/java/io/strimzi/systemtest/utils/kafkaUtils/KafkaUtils.java
```diff
@@ -591,4 +591,9 @@ public static void waitUntilKafkaStatusContainsKafkaMetadataState(String namespa
             return k.getStatus().getKafkaMetadataState().equals(desiredKafkaMetadataState);
         });
     }
+
+    public static String getKafkaLogFolderNameInPod(String namespaceName, String kafkaPodName) {
+        return ResourceManager.cmdKubeClient().namespace(namespaceName)
+            .execInPod(kafkaPodName, "/bin/bash", "-c", "ls /var/lib/kafka/data | grep \"kafka-log[0-9]\" -o").out().trim();
```
see-quick, line 597: `kafka-log[0-9]` that suffix number is based on pod-number I think so if you deploy Kafka with ten replicas it will not match in such an edge case. Either make it to `[0-9]+` or add `kafkaPodNum` as a suffix?

im-konge: Yeah I was thinking that we will actually not have more than 10 replicas. Also the listing the dir instead of parsing the number of Pod from the name looked easier to me

systemtest/src/test/java/io/strimzi/systemtest/migration/MigrationST.java
im-konge: added the check at the beginning of the `doRollback()` method

see-quick (approved):
LGTM 👍 Nice works and thanks for it 💯

systemtest/src/test/java/io/strimzi/systemtest/migration/MigrationST.java
```diff
@@ -87,20 +99,70 @@ public class MigrationST extends AbstractST {
...
+    void testRollbackDuringMigration(ExtensionContext extensionContext) {
+        TestStorage testStorage = new TestStorage(extensionContext, TestConstants.CO_NAMESPACE);
+
+        setupMigrationTestCase(extensionContext, testStorage);
+        doFirstPartOfMigration(extensionContext, testStorage);
+        doRollback(testStorage);
+    }
+
+    private void setupMigrationTestCase(ExtensionContext extensionContext, TestStorage testStorage) {
+        postMigrationTopicName = testStorage.getTopicName() + "-midstep";
```
ppatierno, line 152: we are using `-post-migration` in the current codebase for migration ST as the variable name suggests

systemtest/src/test/java/io/strimzi/systemtest/migration/MigrationST.java
```diff
@@ -239,17 +283,67 @@ void testMigrationFromZkToKRaft(ExtensionContext extensionContext) {
...
+        LOGGER.info("Creating KafkaTopic: {} and checking if the metadata are in both ZK and KRaft", newTopicName);
+        resourceManager.createResourceWithWait(extensionContext, KafkaTopicTemplates.topic(testStorage.getClusterName(), newTopicName, testStorage.getNamespaceName()).build());
+
+        LOGGER.info("Checking if metadata about KafkaTopic: {} are in KRaft controller", newTopicName);
+
+        assertThatTopicIsPresentInKRaftMetadata(testStorage.getNamespaceName(), controllerSelector, newTopicName);
+
+        if (checkZk) {
+            LOGGER.info("Checking if metadata about KafkaTopic: {} are in ZK", newTopicName);
+
```
ppatierno, line 339: is this blank line needed?

ppatierno (approved):
LGTM. Just a couple of nits.

systemtest/src/test/java/io/strimzi/systemtest/migration/MigrationST.java
im-konge: Yeah good catch, I missed this one when I was copying the changes between the branches. Thanks!

ppatierno (approved):

PR #9615, opened 2024-01-30
[ST] Add AZP for migration and few fixups to the migration tests

im-konge:
### Type of change

- Enhancement 

### Description

This PR adds AZP for migration testgroup, which will allow us testing the migration inside the Azure. First idea was to add it to the `regression`, but the execution time is ~30 minutes for those two test cases - so it would make the current execution time of the regression pipelines too long and we could possibly hit the timeout.
Also, it's not maybe something we should run for every regression execution, so we can save some execution time by running just migration test cases.

Finally, this PR adds some fixes to the current STs:

- instead of short names of the resources in the `waitForZooKeeperResourcesDeletion` I'm using the whole names - to remove issues with "No such resource" exceptions
- in two checks for existence of KafkaTopic in ZK/KRaft metadata - instead of direct `assert` I added `waitFor`s that will wait until the KafkaTopic is in metadata, otherwise it fails the test -> that's because of issue in ZK, where we sometimes get "Connection refused" exception

Finally I added assumption to the `beforeAll` to check, so the tests will be skipped in case we are disabling KafkaNodePools or KRaft feature gates (comment from previous PR that I forgot to add).

### Checklist

### DIFF
--- a/.azure/templates/jobs/system-tests/migration_jobs.yaml
+++ b/.azure/templates/jobs/system-tests/migration_jobs.yaml
@@ -6,6 +6,6 @@ jobs:
       profile: 'all'
       groups: 'migration'
       cluster_operator_install_type: 'bundle'
-      timeout: 240
+      timeout: 120
       releaseVersion: '${{ parameters.releaseVersion }}'
       kafkaVersion: '${{ parameters.kafkaVersion }}'
--- a/.azure/templates/jobs/system-tests/migration_jobs.yaml
+++ b/.azure/templates/jobs/system-tests/migration_jobs.yaml
@@ -1,0 +1,11 @@
+jobs:
+  - template: '../../steps/system_test_general.yaml'
+    parameters:
+      name: 'migration'
+      display_name: 'migration-bundle'
+      profile: 'all'
+      groups: 'migration'
+      cluster_operator_install_type: 'bundle'
+      timeout: 240
+      releaseVersion: '${{ parameters.releaseVersion }}'
+      kafkaVersion: '${{ parameters.kafkaVersion }}'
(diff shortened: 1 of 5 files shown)


ppatierno:
@im-konge I was reviewing the PR which looks goot to me but then I noticed that your `/azp list` command doesn't really show the `migration` pipeline and when trying to run it shows "no pipelines". So my question is ... does this addition really work?

see-quick:
> @im-konge I was reviewing the PR which looks goot to me but then I noticed that your `/azp list` command doesn't really show the `migration` pipeline and when trying to run it shows "no pipelines". So my question is ... does this addition really work?

It has to be manually added in Azure UI :)

im-konge:
@ppatierno for now, I just wanted to "show" you that the pipeline works, once it will be merged into your PR (or into the main branch) I will add it to the particular directories, so it will be visible and available in the `azp` command. IIRC it is connected to particular branch and because this branch is on my fork, I didn't wanted to add that one.

ppatierno:
@im-konge @see-quick ah right! got it! thanks!

.azure/templates/jobs/system-tests/migration_jobs.yaml
```diff
@@ -0,0 +1,11 @@
+jobs:
+  - template: '../../steps/system_test_general.yaml'
+    parameters:
+      name: 'migration'
+      display_name: 'migration-bundle'
+      profile: 'all'
+      groups: 'migration'
+      cluster_operator_install_type: 'bundle'
+      timeout: 240
```
see-quick, line 9: do we need such a big timeout for this pipeline btw?

see-quick (approved):
LGTM, just one note.

.azure/templates/jobs/system-tests/migration_jobs.yaml
im-konge: nope, just copy paste :D

PR #9616, opened 2024-01-30
Fix CC not restarted when API secret changes

fvaleri:
The user may need to create new CC API credentials because they may have been compromised. This can be done by simply deleting the secret containing CC API credentials, which is then recreated by the CO. The problem is that CC is not restarted, which leads to the following Rebalance error:

```sh
2024-01-29 17:44:08 ERROR KafkaRebalanceAssemblyOperator:483 - Reconciliation #64(kafkarebalance-watch) KafkaRebalance(test/my-rebalance): Status updated to [NotReady] due to error: Unexpected status code 401 for request to my-cluster-cruise-control.test.svc:9090/kafkacruisecontrol/rebalance?json=true&dryrun=true&verbose=true&skip_hard_goal_check=false&rebalance_disk=false
```

To fix this issue, I'm adding the API secret hash as CC annotation, so that any change will trigger a CC pod restart.

### DIFF
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/CruiseControlReconciler.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/CruiseControlReconciler.java
@@ -78,26 +78,26 @@ public class CruiseControlReconciler {
      * @param reconciliation            Reconciliation marker
      * @param config                    Cluster Operator Configuration
      * @param supplier                  Supplier with Kubernetes Resource Operators
+     * @param passwordGenerator         The password generator for API users
      * @param kafkaAssembly             The Kafka custom resource
      * @param versions                  The supported Kafka versions
      * @param kafkaBrokerNodes          List of the broker nodes which are part of the Kafka cluster
      * @param kafkaBrokerStorage        A map with storage configuration used by the Kafka cluster and its broker pools
      * @param kafkaBrokerResources      A map with resource configuration used by the Kafka cluster and its broker pools
      * @param clusterCa                 The Cluster CA instance
-     * @param passwordGenerator         The password generator for API users
      */
     @SuppressWarnings({"checkstyle:ParameterNumber"})
     public CruiseControlReconciler(
             Reconciliation reconciliation,
             ClusterOperatorConfig config,
             ResourceOperatorSupplier supplier,
+            PasswordGenerator passwordGenerator,
             Kafka kafkaAssembly,
             KafkaVersion.Lookup versions,
             Set<NodeRef> kafkaBrokerNodes,
             Map<String, Storage> kafkaBrokerStorage,
             Map<String, ResourceRequirements> kafkaBrokerResources,
-            ClusterCa clusterCa,
-            PasswordGenerator passwordGenerator
+            ClusterCa clusterCa
     ) {
         this.reconciliation = reconciliation;
         this.cruiseControl = CruiseControl.fromCrd(reconciliation, kafkaAssembly, versions, kafkaBrokerNodes, kafkaBrokerStorage, kafkaBrokerResources, supplier.sharedEnvironmentProvider);
@@ -107,14 +107,14 @@ public class CruiseControlReconciler {
         this.operatorNamespace = config.getOperatorNamespace();
         this.operatorNamespaceLabels = config.getOperatorNamespaceLabels();
         this.isNetworkPolicyGeneration = config.isNetworkPolicyGeneration();
-
+        this.passwordGenerator = passwordGenerator;
+        
         this.deploymentOperator = supplier.deploymentOperations;
         this.secretOperator = supplier.secretOperations;
         this.serviceAccountOperator = supplier.serviceAccountOperations;
         this.serviceOperator = supplier.serviceOperations;
         this.networkPolicyOperator = supplier.networkPolicyOperator;
         this.configMapOperator = supplier.configMapOperations;
-        this.passwordGenerator = passwordGenerator;
     }
 
     /**
@@ -255,7 +255,7 @@ public class CruiseControlReconciler {
                             newSecret.setData(oldSecret.getData());
                         }
                         
-                        this.apiSecretHash = ReconcilerUtils.hashSecretContent(newSecret, "password");
+                        this.apiSecretHash = ReconcilerUtils.hashSecretContent(newSecret);
                         
                         return secretOperator.reconcile(reconciliation, reconciliation.namespace(), CruiseControlResources.apiSecretName(reconciliation.name()), newSecret)
                                 .map((Void) null);
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/CruiseControlReconciler.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/CruiseControlReconciler.java
@@ -26,6 +26,7 @@ import io.strimzi.operator.common.ReconciliationLogger;
 import io.strimzi.operator.common.Util;
 import io.strimzi.operator.common.model.Ca;
 import io.strimzi.operator.common.model.Labels;
+import io.strimzi.operator.common.model.PasswordGenerator;
 import io.strimzi.operator.common.operator.resource.ConfigMapOperator;
 import io.strimzi.operator.common.operator.resource.DeploymentOperator;
 import io.strimzi.operator.common.operator.resource.NetworkPolicyOperator;
@@ -40,7 +41,6 @@ import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
-
 
 /**
  * Class used for reconciliation of Cruise Control. This class contains both the steps of the Cruise Control
@@ -64,12 +64,13 @@ public class CruiseControlReconciler {
     private final ServiceOperator serviceOperator;
     private final NetworkPolicyOperator networkPolicyOperator;
     private final ConfigMapOperator configMapOperator;
+    private final PasswordGenerator passwordGenerator;
 
     private boolean existingCertsChanged = false;
 
     private String serverConfigurationHash = "";
     private String capacityConfigurationHash = "";
-    private String apiCredentialsHash = "";
+    private String apiSecretHash = "";
     
     /**
      * Constructs the Cruise Control reconciler
@@ -83,6 +84,7 @@ public class CruiseControlReconciler {
      * @param kafkaBrokerStorage        A map with storage configuration used by the Kafka cluster and its broker pools
      * @param kafkaBrokerResources      A map with resource configuration used by the Kafka cluster and its broker pools
      * @param clusterCa                 The Cluster CA instance
+     * @param passwordGenerator         The password generator for API users
      */
     @SuppressWarnings({"checkstyle:ParameterNumber"})
     public CruiseControlReconciler(
@@ -94,7 +96,8 @@ public class CruiseControlReconciler {
             Set<NodeRef> kafkaBrokerNodes,
             Map<String, Storage> kafkaBrokerStorage,
             Map<String, ResourceRequirements> kafkaBrokerResources,
-            ClusterCa clusterCa
+            ClusterCa clusterCa,
+            PasswordGenerator passwordGenerator
     ) {
         this.reconciliation = reconciliation;
         this.cruiseControl = CruiseControl.fromCrd(reconciliation, kafkaAssembly, versions, kafkaBrokerNodes, kafkaBrokerStorage, kafkaBrokerResources, supplier.sharedEnvironmentProvider);
@@ -111,6 +114,7 @@ public class CruiseControlReconciler {
         this.serviceOperator = supplier.serviceOperations;
         this.networkPolicyOperator = supplier.networkPolicyOperator;
         this.configMapOperator = supplier.configMapOperations;
+        this.passwordGenerator = passwordGenerator;
     }
 
     /**
@@ -242,7 +246,7 @@ public class CruiseControlReconciler {
         if (cruiseControl != null) {
             return secretOperator.getAsync(reconciliation.namespace(), CruiseControlResources.apiSecretName(reconciliation.name()))
                     .compose(oldSecret -> {
-                        Secret newSecret = cruiseControl.generateApiSecret();
+                        Secret newSecret = cruiseControl.generateApiSecret(passwordGenerator);
 
                         if (oldSecret != null)  {
                             // The credentials should not change with every release
@@ -250,9 +254,9 @@ public class CruiseControlReconciler {
                             // But we use the new secret to update labels etc. if needed
                             newSecret.setData(oldSecret.getData());
                         }
-
-                        this.apiCredentialsHash = String.valueOf(newSecret.hashCode());
-
+                        
+                        this.apiSecretHash = ReconcilerUtils.hashSecretContent(newSecret, "password");
+                        
                         return secretOperator.reconcile(reconciliation, reconciliation.namespace(), CruiseControlResources.apiSecretName(reconciliation.name()), newSecret)
                                 .map((Void) null);
                     });
@@ -289,7 +293,7 @@ public class CruiseControlReconciler {
             podAnnotations.put(Ca.ANNO_STRIMZI_IO_CLUSTER_CA_KEY_GENERATION, String.valueOf(clusterCa.caKeyGeneration()));
             podAnnotations.put(CruiseControl.ANNO_STRIMZI_SERVER_CONFIGURATION_HASH, serverConfigurationHash);
             podAnnotations.put(CruiseControl.ANNO_STRIMZI_CAPACITY_CONFIGURATION_HASH, capacityConfigurationHash);
-            podAnnotations.put(Annotations.ANNO_STRIMZI_AUTH_HASH, apiCredentialsHash);
+            podAnnotations.put(Annotations.ANNO_STRIMZI_AUTH_HASH, apiSecretHash);
             
             Deployment deployment = cruiseControl.generateDeployment(podAnnotations, isOpenShift, imagePullPolicy, imagePullSecrets);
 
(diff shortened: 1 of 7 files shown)


cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/CruiseControlReconciler.java
```diff
@@ -249,6 +251,8 @@ protected Future<Void> apiSecret() {
                             newSecret.setData(oldSecret.getData());
                         }
 
+                        this.apiCredentialsHash = String.valueOf(newSecret.hashCode());
```
scholzj, line 254: How does it work when you hash the whole Secret object? Doesn't it just hash some default `toString` method result? I wonder if it would be better to hash some proper Strings with the values.

fvaleri: This is using the hashCode method generated by Lombok, which uses all non-static, non-transient fields. I think this is fine, but I don't have strong opinion. We can join all values and hash that if you prefer.

scholzj: I think it would be good to have at least some better unit tests that make sure that the value does not change between different reconciliations where the new Secrets are generated. For example, check for the exact hash in the test you have instead of `notNullValue`. Because this depends on the implementation that we do not have under control. I also wonder how much this exposes us to changes to the underlying Fabric8 objects that we might not care about (while we might not care about them in most cases as the pods anyway roll at operator upgrade, it might _surprise_ us one day for some reason).

fvaleri: Ok, makes sense.

fvaleri:
@scholzj so I'm now hashing all passwords in the API secret. Test updated.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/ReconcilerUtils.java
```diff
@@ -381,4 +385,34 @@ public static boolean nodePoolsEnabled(Kafka kafka) {
...
+            }
+        }
+        return sb.toString().length() > 0 
+            ? hash(sb.toString()) 
+            : hash("Keyword not found");
+    }
+    
+    private static String hash(String s) {
+        try {
+            byte[] hash = MessageDigest.getInstance("MD5").digest(s.getBytes());
```
kyguy, line 412: Given the CodeQL check, would "SHA-256" be better here?

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/CruiseControlReconciler.java
```diff
@@ -240,15 +246,17 @@ protected Future<Void> apiSecret() {
...
 
                         if (oldSecret != null)  {
                             // The credentials should not change with every release
                             // So if the secret with credentials already exists, we re-use the values
                             // But we use the new secret to update labels etc. if needed
                             newSecret.setData(oldSecret.getData());
                         }
-
+                        
+                        this.apiSecretHash = ReconcilerUtils.hashSecretContent(newSecret, "password");
```
kyguy, line 258: Will the CC API secret have a key "password"? From what i understand, the secret has the following three keys:
```
cruise-control.apiAdminPassword
cruise-control.apiUserPassword
cruise-control.apiAuthFile
```

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/ReconcilerUtils.java
```diff
@@ -381,4 +385,34 @@ public static boolean nodePoolsEnabled(Kafka kafka) {
...
+        StringBuilder sb = new StringBuilder();
+        for (String key : secret.getData().keySet()) {
+            if (Pattern.compile(keyword, Pattern.CASE_INSENSITIVE).matcher(key).find()) {
+                sb.append(secret.getData().get(key));
+            }
+        }
+        return sb.toString().length() > 0 
+            ? hash(sb.toString()) 
+            : hash("Keyword not found");
+    }
```
kyguy, line 408: Do we only want to create a hash of a single key of the CC API secret? Would there be any benefit of creating a hash of all the keys in the secret?

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/ReconcilerUtils.java
```diff
@@ -381,4 +385,34 @@ public static boolean nodePoolsEnabled(Kafka kafka) {
...
+     */
+    public static String hashSecretContent(Secret secret, String keyword) {
+        if (secret == null) {
+            return hash("Secret not found");
+        }
+        StringBuilder sb = new StringBuilder();
+        for (String key : secret.getData().keySet()) {
+            if (Pattern.compile(keyword, Pattern.CASE_INSENSITIVE).matcher(key).find()) {
+                sb.append(secret.getData().get(key));
+            }
```
kyguy, line 403: Does the if statement here make it so we only create a hash of a single key of the CC API secret here? Would there be any benefit of creating a hash of all the keys in the secret?

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/ReconcilerUtils.java
fvaleri: This is not used for any security related feature, but only for detecting changes.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/CruiseControlReconciler.java
fvaleri: No, the hashSecretContent will has all values of keys containing the keyword. In this case: cruise-control.apiAdminPassword and apiUserPassword.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/ReconcilerUtils.java
fvaleri: I'm creating a hash of all the keys containing "password".

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/CruiseControlReconciler.java
kyguy: Whoops, I misunderstood the code, thanks for the explanation, makes sense!

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/CruiseControlReconciler.java
```diff
@@ -92,7 +96,8 @@ public CruiseControlReconciler(
             Set<NodeRef> kafkaBrokerNodes,
             Map<String, Storage> kafkaBrokerStorage,
             Map<String, ResourceRequirements> kafkaBrokerResources,
-            ClusterCa clusterCa
+            ClusterCa clusterCa,
+            PasswordGenerator passwordGenerator
```
scholzj, line 100: Maybe you could add it after the `ResourceOperatorSupplier` as it is a similar type of class?

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/ReconcilerUtils.java
```diff
@@ -381,4 +385,34 @@ public static boolean nodePoolsEnabled(Kafka kafka) {
...
+
+    /**
+     * Creates a hash from Secret's content.
+     * @param secret Secret with data.
+     * @param keyword String used to find matching keys.
+     * @return Hash of the values whose keys match the keyword.
+     */
+    public static String hashSecretContent(Secret secret, String keyword) {
+        if (secret == null) {
+            return hash("Secret not found");
```
scholzj, line 397: This logic seems weird. I would throw some exception.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/ReconcilerUtils.java
```diff
@@ -381,4 +385,34 @@ public static boolean nodePoolsEnabled(Kafka kafka) {
...
         return KafkaCluster.ENABLED_VALUE_STRIMZI_IO_KRAFT.equals(Annotations.stringAnnotation(kafka, Annotations.ANNO_STRIMZI_IO_KRAFT, "disabled").toLowerCase(Locale.ENGLISH));
     }
+
+    /**
+     * Creates a hash from Secret's content.
+     * @param secret Secret with data.
+     * @param keyword String used to find matching keys.
+     * @return Hash of the values whose keys match the keyword.
+     */
+    public static String hashSecretContent(Secret secret, String keyword) {
```
scholzj, line 395: This seems strange. Why do we need the `keyword` pattern? Why not just hash everything in `.data`?

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/ReconcilerUtils.java
```diff
@@ -381,4 +385,34 @@ public static boolean nodePoolsEnabled(Kafka kafka) {
...
+        }
+        StringBuilder sb = new StringBuilder();
+        for (String key : secret.getData().keySet()) {
+            if (Pattern.compile(keyword, Pattern.CASE_INSENSITIVE).matcher(key).find()) {
+                sb.append(secret.getData().get(key));
+            }
+        }
+        return sb.toString().length() > 0 
+            ? hash(sb.toString()) 
+            : hash("Keyword not found");
```
scholzj, line 407: What happens when you hash empty String? Or why do you have this logic here?

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaAssemblyOperator.java
```diff
@@ -704,7 +704,8 @@ CruiseControlReconciler cruiseControlReconciler()   {
                     kafkaBrokerNodes,
                     kafkaBrokerStorage,
                     kafkaBrokerResources,
-                    clusterCa
+                    clusterCa,
+                    new PasswordGenerator(16)
```
scholzj, line 708: I think we already have a password generator.

```suggestion
                    passwordGenerator
```

cluster-operator/src/test/java/io/strimzi/operator/cluster/operator/assembly/CruiseControlReconcilerTest.java
```diff
@@ -78,14 +80,16 @@ public void reconcileEnabledCruiseControl(VertxTestContext context) {
         ServiceOperator mockServiceOps = supplier.serviceOperations;
         NetworkPolicyOperator mockNetPolicyOps = supplier.networkPolicyOperator;
         ConfigMapOperator mockCmOps = supplier.configMapOperations;
-
+        PasswordGenerator mockPasswordGenerator = mock(PasswordGenerator.class);
```
scholzj, line 83: Why not just do `new PasswordGenerator(10, "a", "a")`?

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/ReconcilerUtils.java
```diff
@@ -381,4 +385,34 @@ public static boolean nodePoolsEnabled(Kafka kafka) {
...
+            return hash("Secret not found");
+        }
+        StringBuilder sb = new StringBuilder();
+        for (String key : secret.getData().keySet()) {
+            if (Pattern.compile(keyword, Pattern.CASE_INSENSITIVE).matcher(key).find()) {
+                sb.append(secret.getData().get(key));
+            }
+        }
+        return sb.toString().length() > 0 
+            ? hash(sb.toString())
```
scholzj, line 406: You should use `Util.hashStub`

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/CruiseControlReconciler.java
fvaleri: Sure.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/ReconcilerUtils.java
fvaleri: Better.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/ReconcilerUtils.java
fvaleri: Yes, I think it's better.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaAssemblyOperator.java
fvaleri: Right. Fixed.

cluster-operator/src/test/java/io/strimzi/operator/cluster/operator/assembly/CruiseControlReconcilerTest.java
fvaleri: Makes sense.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/ReconcilerUtils.java
fvaleri: Yep. Didn't notice that.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/ReconcilerUtils.java
```diff
@@ -381,4 +385,34 @@ public static boolean nodePoolsEnabled(Kafka kafka) {
...
+            if (Pattern.compile(keyword, Pattern.CASE_INSENSITIVE).matcher(key).find()) {
+                sb.append(secret.getData().get(key));
+            }
+        }
+        return sb.toString().length() > 0 
+            ? hash(sb.toString()) 
+            : hash("Keyword not found");
+    }
+    
+    private static String hash(String s) {
```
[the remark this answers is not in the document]
fvaleri: Right.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/ReconcilerUtils.java
```diff
@@ -381,4 +381,23 @@ public static boolean nodePoolsEnabled(Kafka kafka) {
     public static boolean kraftEnabled(Kafka kafka) {
         return KafkaCluster.ENABLED_VALUE_STRIMZI_IO_KRAFT.equals(Annotations.stringAnnotation(kafka, Annotations.ANNO_STRIMZI_IO_KRAFT, "disabled").toLowerCase(Locale.ENGLISH));
     }
+
+    /**
+     * Creates a hash from Secret's data.
```
ppatierno, line 386:
```suggestion
     * Creates a hash from Secret's data.
     
```

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/ReconcilerUtils.java
```diff
@@ -381,4 +381,23 @@ public static boolean nodePoolsEnabled(Kafka kafka) {
...
+     */
+    public static String hashSecretContent(Secret secret) {
+        if (secret == null) {
+            throw new RuntimeException("Secret not found");
+        }
+        StringBuilder sb = new StringBuilder();
+        for (String key : secret.getData().keySet()) {
+            sb.append(secret.getData().get(key));
+        }
+        if (sb.toString().length() == 0) {
```
ppatierno, line 398: what if we do this earlier by checking `secret.getData() != null && secret.getData().size() != 0` instead of creating a new `StringBuilder` (even if empty) which could not be needed at all? Just a suggestion.

ppatierno (approved):
LGTM. Just a couple of nits.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/ReconcilerUtils.java
fvaleri: Makes sense thanks.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/ReconcilerUtils.java
```diff
@@ -381,4 +381,24 @@ public static boolean nodePoolsEnabled(Kafka kafka) {
...
+    public static String hashSecretContent(Secret secret) {
+        if (secret == null) {
+            throw new RuntimeException("Secret not found");
+        }
+        if (secret.getData() == null || secret.getData().isEmpty()) {
+            throw new RuntimeException("Empty secret");
+        }
+        StringBuilder sb = new StringBuilder();
+        for (String key : secret.getData().keySet()) {
+            sb.append(secret.getData().get(key));
```
scholzj, line 400: Should you append the key as well? That way, when the same value is kept but key changes, the hash will catch it.

scholzj (commented):
One nit. LGTM otherwise.

kyguy (approved):
Awesome work, thanks Fede!

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/ReconcilerUtils.java
```diff
@@ -381,4 +381,24 @@ public static boolean nodePoolsEnabled(Kafka kafka) {
...
+        }
+        return Util.hashStub(sb.toString());
+    }
```
scholzj, line 403: A nit, but if you add some spacing, it would be easier to read:

```suggestion
    public static String hashSecretContent(Secret secret) {
        if (secret == null) {
            throw new RuntimeException("Secret not found");
        }
        
        if (secret.getData() == null || secret.getData().isEmpty()) {
            throw new RuntimeException("Empty secret");
        }
        
        StringBuilder sb = new StringBuilder();
        for (String key : secret.getData().keySet()) {
            sb.append(secret.getData().get(key));
        }
        
        return Util.hashStub(sb.toString());
    }
```

fvaleri:
@scholzj I tried `SecurityST#testAutoRenewClientsCaCertsTriggeredByAnno` multiple times with the latest implementation on my machine and it seems to work fine.

scholzj (approved):

PR #9619, opened 2024-01-30
Do not use SLF4J loggers in Topic Operator classes

scholzj:
### Type of change

- Bugfix

### Description

It looks like 2 classes in the Topic Operator use SLF4J loggers without any obvious reasons. This is likely just by mistake because of a wrong import? This PR changes these loggers to use Log4j2 as the other classes in TO. It also removes SLF4J as a direct dependency to avoid this happening again in the future.

(It also deletes one unnecessary warning suppression)

### Checklist

### DIFF
--- a/topic-operator/pom.xml
+++ b/topic-operator/pom.xml
@@ -91,10 +91,6 @@
             <artifactId>kafka-clients</artifactId>
         </dependency>
         <dependency>
-            <groupId>org.slf4j</groupId>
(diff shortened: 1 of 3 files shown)


Frawless (approved):

see-quick (approved):
LGTM, but it should be reviewed by SMEs @tombentley or @fvaleri .

fvaleri (approved):
LGTM. Thanks.

PR #9624, opened 2024-01-31
Add Mock tests for `KafkaMetadataStateManager` class

ShubhamRwt:
### Type of change

_Select the type of your PR_

- Enhancement / new feature

### Description

This PR adds the mock tests for the `KafkaMetadataStateManager` class

### Checklist

- [ ] Write tests
- [ ] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaMetadataStateManager.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaMetadataStateManager.java
@@ -171,13 +171,26 @@ public class KafkaMetadataStateManager {
      * @param controllerPodName Name of the quorum controller leader pod
      */
     public void checkMigrationInProgress(Reconciliation reconciliation, Secret clusterCaCertSecret, Secret coKeySecret, String controllerPodName) {
-        KafkaAgentClient kafkaAgentClient = new KafkaAgentClient(reconciliation, reconciliation.name(), reconciliation.namespace(), clusterCaCertSecret, coKeySecret);
+        KafkaAgentClient kafkaAgentClient = initKafkaAgentClient(reconciliation, clusterCaCertSecret, coKeySecret);
         KRaftMigrationState kraftMigrationState = kafkaAgentClient.getKRaftMigrationState(controllerPodName);
         LOGGER.infoCr(reconciliation, "ZooKeeper to KRaft migration state {} checked on controller {}", kraftMigrationState.state(), controllerPodName);
         if (kraftMigrationState.state() == -1) {
             throw new RuntimeException("Failed to get the ZooKeeper to KRaft migration state");
         }
         this.isMigrationDone = kraftMigrationState.isMigrationDone();
+    }
+
+    /**
+     * Creates the KafkaAgentClient instance
+     *
+     * @param reconciliation    Reconciliation information
+     * @param clusterCaCertSecret   Secret with the Cluster CA public key
+     * @param coKeySecret   Secret with the Cluster CA private key
+     *
+     * @return KafkaAgentClient instance
+     */
+    /* test */ protected KafkaAgentClient initKafkaAgentClient(Reconciliation reconciliation, Secret clusterCaCertSecret, Secret coKeySecret)  {
+        return new KafkaAgentClient(reconciliation, reconciliation.name(), reconciliation.namespace(), clusterCaCertSecret, coKeySecret);
     }
 
     /**
--- a/cluster-operator/src/test/java/io/strimzi/operator/cluster/operator/assembly/KafkaMetadataStateManagerMigrationTest.java
+++ b/cluster-operator/src/test/java/io/strimzi/operator/cluster/operator/assembly/KafkaMetadataStateManagerMigrationTest.java
@@ -1,0 +1,630 @@
+/*
+ * Copyright Strimzi authors.
+ * License: Apache License 2.0 (see the file LICENSE or http://apache.org/licenses/LICENSE-2.0.html).
+ */
+package io.strimzi.operator.cluster.operator.assembly;
+
+import io.fabric8.kubernetes.api.model.Quantity;
+import io.fabric8.kubernetes.api.model.ResourceRequirementsBuilder;
+import io.fabric8.kubernetes.api.model.Secret;
+import io.strimzi.api.kafka.model.kafka.Kafka;
+import io.strimzi.api.kafka.model.kafka.KafkaBuilder;
+import io.strimzi.api.kafka.model.kafka.KafkaResources;
+import io.strimzi.api.kafka.model.kafka.KafkaStatus;
+import io.strimzi.api.kafka.model.kafka.PersistentClaimStorageBuilder;
+import io.strimzi.api.kafka.model.kafka.listener.GenericKafkaListenerBuilder;
+import io.strimzi.api.kafka.model.kafka.listener.KafkaListenerType;
+import io.strimzi.api.kafka.model.nodepool.KafkaNodePool;
+import io.strimzi.api.kafka.model.nodepool.KafkaNodePoolBuilder;
+import io.strimzi.api.kafka.model.nodepool.ProcessRoles;
+import io.strimzi.certs.OpenSslCertManager;
+import io.strimzi.operator.cluster.ClusterOperatorConfig;
+import io.strimzi.operator.cluster.KafkaVersionTestUtils;
+import io.strimzi.operator.cluster.PlatformFeaturesAvailability;
+import io.strimzi.operator.cluster.ResourceUtils;
+import io.strimzi.operator.cluster.model.AbstractModel;
+import io.strimzi.operator.cluster.model.ClusterCa;
+import io.strimzi.operator.cluster.model.KafkaCluster;
+import io.strimzi.operator.cluster.model.KafkaMetadataConfigurationState;
+import io.strimzi.operator.cluster.model.KafkaPool;
+import io.strimzi.operator.cluster.model.KafkaVersion;
+import io.strimzi.operator.cluster.model.MockSharedEnvironmentProvider;
+import io.strimzi.operator.cluster.model.SharedEnvironmentProvider;
+import io.strimzi.operator.cluster.model.nodepools.NodePoolUtils;
+import io.strimzi.operator.cluster.operator.resource.KRaftMigrationState;
+import io.strimzi.operator.cluster.operator.resource.KafkaAgentClient;
+import io.strimzi.operator.cluster.operator.resource.ResourceOperatorSupplier;
+import io.strimzi.operator.common.Annotations;
+import io.strimzi.operator.common.Reconciliation;
+import io.strimzi.operator.common.ReconciliationLogger;
+import io.strimzi.operator.common.model.ClientsCa;
+import io.strimzi.operator.common.model.Labels;
+import io.strimzi.operator.common.model.PasswordGenerator;
+import io.strimzi.operator.common.operator.MockCertManager;
+import io.strimzi.operator.common.operator.resource.SecretOperator;
+import io.strimzi.platform.KubernetesVersion;
+import io.vertx.core.Future;
+import io.vertx.core.Vertx;
+import io.vertx.junit5.Checkpoint;
+import io.vertx.junit5.VertxExtension;
+import io.vertx.junit5.VertxTestContext;
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.ExtendWith;
+
+import java.time.Clock;
+import java.util.List;
+import java.util.Map;
+
+import static io.strimzi.api.kafka.model.kafka.KafkaMetadataState.KRaft;
+import static io.strimzi.api.kafka.model.kafka.KafkaMetadataState.KRaftDualWriting;
+import static io.strimzi.api.kafka.model.kafka.KafkaMetadataState.KRaftMigration;
+import static io.strimzi.api.kafka.model.kafka.KafkaMetadataState.KRaftPostMigration;
+import static io.strimzi.api.kafka.model.kafka.KafkaMetadataState.PreKRaft;
+import static io.strimzi.api.kafka.model.kafka.KafkaMetadataState.ZooKeeper;
+import static org.hamcrest.CoreMatchers.is;
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.ArgumentMatchers.eq;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
+
+@ExtendWith(VertxExtension.class)
+public class KafkaMetadataStateManagerMigrationTest {
+
+    private static final String NAMESPACE = "my-namespace";
+    private static final String CLUSTER_NAME = "my-cluster";
+    private static final ClusterOperatorConfig CONFIG = ResourceUtils.dummyClusterOperatorConfig();
+    private static final KubernetesVersion KUBERNETES_VERSION = KubernetesVersion.MINIMAL_SUPPORTED_VERSION;
+    private static final MockCertManager CERT_MANAGER = new MockCertManager();
+    private static final PasswordGenerator PASSWORD_GENERATOR = new PasswordGenerator(10, "a", "a");
+    private static final KafkaVersion.Lookup VERSIONS = KafkaVersionTestUtils.getKafkaVersionLookup();
+    private static final Reconciliation RECONCILIATION = new Reconciliation("test-trigger", Kafka.RESOURCE_KIND, NAMESPACE, CLUSTER_NAME);
+    private static final SharedEnvironmentProvider SHARED_ENV_PROVIDER = new MockSharedEnvironmentProvider();
+    protected static Vertx vertx;
+
+    @BeforeAll
+    public static void before() {
+        vertx = Vertx.vertx();
+    }
+
+    private final static ClusterCa CLUSTER_CA = new ClusterCa(
+            RECONCILIATION,
+            CERT_MANAGER,
+            PASSWORD_GENERATOR,
+            CLUSTER_NAME,
+            ResourceUtils.createInitialCaCertSecret(NAMESPACE, CLUSTER_NAME, AbstractModel.clusterCaCertSecretName(CLUSTER_NAME), MockCertManager.clusterCaCert(), MockCertManager.clusterCaCertStore(), "123456"),
+            ResourceUtils.createInitialCaKeySecret(NAMESPACE, CLUSTER_NAME, AbstractModel.clusterCaKeySecretName(CLUSTER_NAME), MockCertManager.clusterCaKey())
+    );
+
+    private final static ClientsCa CLIENTS_CA = new ClientsCa(
+            RECONCILIATION,
+            new OpenSslCertManager(),
+            new PasswordGenerator(10, "a", "a"),
+            KafkaResources.clientsCaCertificateSecretName(CLUSTER_NAME),
+            ResourceUtils.createInitialCaCertSecret(NAMESPACE, CLUSTER_NAME, AbstractModel.clusterCaCertSecretName(CLUSTER_NAME), MockCertManager.clusterCaCert(), MockCertManager.clusterCaCertStore(), "123456"),
+            KafkaResources.clientsCaKeySecretName(CLUSTER_NAME),
+            ResourceUtils.createInitialCaKeySecret(NAMESPACE, CLUSTER_NAME, AbstractModel.clusterCaKeySecretName(CLUSTER_NAME), MockCertManager.clusterCaKey()),
+            365,
+            30,
+            true,
+            null
+    );
+
+    private static final Kafka KAFKA = new KafkaBuilder()
+            .withNewMetadata()
+                .withName(CLUSTER_NAME)
+                .withNamespace(NAMESPACE)
+                .withAnnotations(Map.of(Annotations.ANNO_STRIMZI_IO_NODE_POOLS, "enabled"))
+            .endMetadata()
+            .withNewSpec()
+                .withNewKafka()
+                .withListeners(new GenericKafkaListenerBuilder()
+                        .withName("plain")
+                        .withPort(9092)
+                        .withType(KafkaListenerType.INTERNAL)
+                        .withTls(false)
+                        .build())
+            .endKafka()
+            .endSpec()
+            .build();
+
+    private static KafkaCluster createKafkaCluster(Reconciliation reconciliation, Kafka kafka, KafkaMetadataConfigurationState state)   {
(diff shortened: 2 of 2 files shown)


cluster-operator/src/test/java/io/strimzi/operator/cluster/operator/resource/KafkaMetadataStateManagerMockTest.java
```diff
@@ -0,0 +1,640 @@
...
+                .withNewJbodStorage()
+                    .withVolumes(new PersistentClaimStorageBuilder().withId(0).withSize("200Gi").build())
+                .endJbodStorage()
+                .withRoles(ProcessRoles.BROKER)
+                .withResources(new ResourceRequirementsBuilder().withRequests(Map.of("cpu", new Quantity("6"))).build())
+            .endSpec()
+            .build();
+    private static final List<KafkaPool> POOLS = NodePoolUtils.createKafkaPools(Reconciliation.DUMMY_RECONCILIATION, KAFKA, List.of(CONTROLLERS, BROKERS), Map.of(), Map.of(), true, SHARED_ENV_PROVIDER);
+
+    /** Tests migration from Zookeeper State to KRaftPostMigration State.
```
ppatierno, line 180: start the comment on the next line. Please do the same for all the other Javadocs.

```suggestion
    /**
    * Tests migration from Zookeeper State to KRaftPostMigration State.
```

cluster-operator/src/test/java/io/strimzi/operator/cluster/operator/resource/KafkaMetadataStateManagerMockTest.java
```diff
@@ -0,0 +1,640 @@
...
+            30,
+            true,
+            null
+    );
+
+    private static final Kafka KAFKA = new KafkaBuilder()
+            .withNewMetadata()
+                .withName(CLUSTER_NAME)
+                .withNamespace(NAMESPACE)
+                .withAnnotations(Map.of(Annotations.ANNO_STRIMZI_IO_NODE_POOLS, "enabled", Annotations.ANNO_STRIMZI_IO_KRAFT, "migration"))
```
ppatierno, line 117: the `strimzi.io/kraft` annotation should be set during the test and not being here since the creation of the `Kafka` CR.
It depends from where you are starting. I.e. a ZooKeeper based cluster to migrate? The annotation is missing or should be "disabled". You are in the middle of a migration? Then it would be "migration", etc etc

```suggestion
                .withAnnotations(Map.of(Annotations.ANNO_STRIMZI_IO_NODE_POOLS, "enabled"))
```

cluster-operator/src/test/java/io/strimzi/operator/cluster/operator/resource/KafkaMetadataStateManagerMockTest.java
```diff
@@ -0,0 +1,640 @@
...
+         */
+        public MockKafkaMetadataStateManager(Reconciliation reconciliation, Kafka kafkaCr, boolean useKRaftFeatureGateEnabled) {
+            super(reconciliation, kafkaCr, useKRaftFeatureGateEnabled);
+        }
+
+        @Override
+        KafkaAgentClient initKafkaAgentClient(Reconciliation reconciliation, Secret clusterCaCertSecret, Secret coKeySecret) {
+            return mock(KafkaAgentClient.class, invocation -> {
+                if ("getKRaftMigrationState".equals(invocation.getMethod().getName())) {
+                    return new KRaftMigrationState(1);
```
ppatierno, line 612: with this mock, we got that the migration is done as soon as the Kafka agent client is called. What if we want to test that the migration takes long, so for example we need 2-3 reconciliation to get it's done. I think we should cover this use case as well.

cluster-operator/src/test/java/io/strimzi/operator/cluster/operator/resource/KafkaMetadataStateManagerMockTest.java
```diff
@@ -0,0 +1,640 @@
...
+
+        Checkpoint async = context.checkpoint();
+        kr.reconcile(status, Clock.systemUTC()).onComplete(res -> context.verify(() -> {
+            assertThat(res.succeeded(), is(true));
+            assertEquals(status.getKafkaMetadataState(), KRaftMigration.name());
+        })).compose(v -> kr.reconcile(status, Clock.systemUTC())).onComplete(context.succeeding(v -> {
+            assertEquals(status.getKafkaMetadataState(), KRaftDualWriting.name());
+        })).compose(v -> kr.reconcile(status, Clock.systemUTC())).onComplete(context.succeeding(v -> {
+            assertEquals(status.getKafkaMetadataState(), KRaftPostMigration.name());
+            async.flag();
```
ppatierno, line 217: taking into account my first comment about the strimiz.io/kraft annotation on the `KAFKA` member, because you should start from a Kafka CR without annotation, you have to apply it before reconciling.

cluster-operator/src/test/java/io/strimzi/operator/cluster/operator/resource/KafkaMetadataStateManagerMockTest.java
```diff
@@ -0,0 +1,640 @@
...
+     *  The strimzi.io/kraft annotation is set to `enabled`
+     *
+     * @param context Test Context
+     */
+    @Test
+    public void testKRaftPostMigrationToKRaftState(VertxTestContext context) {
+
+        Kafka kafka = new KafkaBuilder(KAFKA)
+                .editMetadata()
+                    .addToAnnotations(Annotations.ANNO_STRIMZI_IO_KRAFT, "enabled")
```
ppatierno, line 231: this is kind of what you should do in the previous test but with the "migration" value :-)

cluster-operator/src/test/java/io/strimzi/operator/cluster/operator/resource/KafkaMetadataStateManagerMockTest.java
```diff
@@ -0,0 +1,640 @@
...
+    public void testWarningIfWrongAnnotationAppliedInKraftState(VertxTestContext context) {
+
+        Kafka kafka = new KafkaBuilder(KAFKA)
+                .editMetadata()
+                .addToAnnotations(Annotations.ANNO_STRIMZI_IO_KRAFT, "rollback")
+                .endMetadata()
+                .withNewStatus()
+                .withKafkaMetadataState(KRaft.name())
+                .endStatus()
+                .build();
```
ppatierno, line 373: wrong indentation

cluster-operator/src/test/java/io/strimzi/operator/cluster/operator/resource/KafkaMetadataStateManagerMockTest.java
```diff
@@ -0,0 +1,640 @@
...
+    public void testWarningIfWrongAnnotationAppliedInZookeeperState(VertxTestContext context) {
+
+        Kafka kafka = new KafkaBuilder(KAFKA)
+                .editMetadata()
+                .addToAnnotations(Annotations.ANNO_STRIMZI_IO_KRAFT, "rollback")
+                .endMetadata()
+                .withNewStatus()
+                .withKafkaMetadataState(ZooKeeper.name())
+                .endStatus()
+                .build();
```
ppatierno, line 466: indentation

cluster-operator/src/test/java/io/strimzi/operator/cluster/operator/resource/KafkaMetadataStateManagerMockTest.java
```diff
@@ -0,0 +1,640 @@
...
+    public void testWarningIfWrongAnnotationAppliedInKraftPostMigrationState(VertxTestContext context) {
+
+        Kafka kafka = new KafkaBuilder(KAFKA)
+                .editMetadata()
+                .addToAnnotations(Annotations.ANNO_STRIMZI_IO_KRAFT, "disabled")
+                .endMetadata()
+                .withNewStatus()
+                .withKafkaMetadataState(KRaftPostMigration.name())
+                .endStatus()
+                .build();
```
ppatierno, line 514: indentation

cluster-operator/src/test/java/io/strimzi/operator/cluster/operator/resource/KafkaMetadataStateManagerMockTest.java
```diff
@@ -0,0 +1,640 @@
...
+    public void testWarningIfWrongAnnotationAppliedInKRaftMigrationState(VertxTestContext context) {
+
+        Kafka kafka = new KafkaBuilder(KAFKA)
+                .editMetadata()
+                .addToAnnotations(Annotations.ANNO_STRIMZI_IO_KRAFT, "rollback")
+                .endMetadata()
+                .withNewStatus()
+                .withKafkaMetadataState(KRaftMigration.name())
+                .endStatus()
+                .build();
```
ppatierno, line 561: indentation

ppatierno (changes requested):

ShubhamRwt:
@ppatierno I see some build failures but they don't look related to this PR

cluster-operator/src/test/java/io/strimzi/operator/cluster/operator/resource/KafkaMetadataStateManagerMockTest.java
```diff
@@ -0,0 +1,657 @@
...
+import static org.junit.jupiter.api.Assertions.assertTrue;
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.ArgumentMatchers.eq;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
+
+@ExtendWith(VertxExtension.class)
+public class KafkaMetadataStateManagerMockTest {
+
+    private static final String NAMESPACE = "my-ns";
```
ppatierno, line 74:
```suggestion
    private static final String NAMESPACE = "my-namespace";
```

cluster-operator/src/test/java/io/strimzi/operator/cluster/operator/resource/KafkaMetadataStateManagerMockTest.java
```diff
@@ -0,0 +1,657 @@
...
+                KafkaMetadataConfigurationState.ZK,
+                null,
+                SHARED_ENV_PROVIDER);
+    }
+
+    private final static KafkaNodePool CONTROLLERS = new KafkaNodePoolBuilder()
+            .withNewMetadata()
+                .withName("controllers")
+                .withNamespace(NAMESPACE)
+                .withGeneration(1L)
```
ppatierno, line 148: is this generation set needed?

cluster-operator/src/test/java/io/strimzi/operator/cluster/operator/resource/KafkaMetadataStateManagerMockTest.java
```diff
@@ -0,0 +1,657 @@
...
+                .endJbodStorage()
+                .withRoles(ProcessRoles.CONTROLLER)
+                .withResources(new ResourceRequirementsBuilder().withRequests(Map.of("cpu", new Quantity("4"))).build())
+            .endSpec()
+            .build();
+    private final static KafkaNodePool BROKERS = new KafkaNodePoolBuilder()
+            .withNewMetadata()
+                .withName("brokers")
+                .withNamespace(NAMESPACE)
+                .withGeneration(1L)
```
ppatierno, line 164: Ditto as above

cluster-operator/src/test/java/io/strimzi/operator/cluster/operator/resource/KafkaMetadataStateManagerMockTest.java
```diff
@@ -0,0 +1,657 @@
...
+        Checkpoint async = context.checkpoint();
+        kr.reconcile(status, Clock.systemUTC()).onComplete(res -> context.verify(() -> {
+            assertThat(res.succeeded(), is(true));
+            assertEquals(status.getKafkaMetadataState(), KRaftMigration.name());
+        })).compose(v -> kr.reconcile(status, Clock.systemUTC())).onComplete(context.succeeding(v -> {
+            // Migration not completed so still in KraftMigration state
+            assertEquals(status.getKafkaMetadataState(), KRaftMigration.name());
+        })).compose(v -> kr.reconcile(status, Clock.systemUTC())).onComplete(context.succeeding(v -> {
+            assertEquals(status.getKafkaMetadataState(), KRaftDualWriting.name());
+            async.flag();
```
ppatierno, line 226: why are you flagging here, shouldn't this happen on the last compose?

cluster-operator/src/test/java/io/strimzi/operator/cluster/operator/resource/KafkaMetadataStateManagerMockTest.java
```diff
@@ -0,0 +1,657 @@
...
+        Checkpoint async = context.checkpoint();
+        kr.reconcile(status, Clock.systemUTC()).onComplete(res -> context.verify(() -> {
+            assertEquals(status.getKafkaMetadataState(), KRaftDualWriting.name());
+            async.flag();
+        }));
+    }
+
+    /**
+     * Tests migration from KRaftDualWriting State to Zookeeper State.
+     * The strimzi.io/kraft annotation is set to `enabled`
```
ppatierno, line 330:
```suggestion
     * The strimzi.io/kraft annotation is set to `disabled`
```

cluster-operator/src/test/java/io/strimzi/operator/cluster/operator/resource/KafkaMetadataStateManagerMockTest.java
```diff
@@ -0,0 +1,657 @@
...
+        }));
+    }
+
+    /**
+     * Tests warning in Kafka status if wrong annotation applied in PreKRaft State.
+     *
+     * @param context Test Context
+     */
+    @Test
+    public void testWarningIfWrongAnnotationAppliedInPreKraftState(VertxTestContext context) {
```
ppatierno, line 427: I think we could remove this test, because `PreKRaft` is not a stable state. It's difficult user changes the annotation on fly.

cluster-operator/src/test/java/io/strimzi/operator/cluster/operator/resource/KafkaMetadataStateManagerMockTest.java
```diff
@@ -0,0 +1,657 @@
...
+            assertEquals(status.getConditions().get(2).getMessage(), "The strimzi.io/kraft annotation can't be set to rollback because the cluster is already ZooKeeper-based." +
+                    "There is no migration ongoing to rollback." +
+                    "If you want to migrate it to be KRaft-based apply the migration value instead.");
+            assertEquals(status.getKafkaMetadataState(), ZooKeeper.name());
+            async.flag();
+        }));
+    }
+
+    /**
+     * Tests warning in Kafka status if wrong annotation applied in Zookeeper State.
```
ppatierno, line 519:
```suggestion
     * Tests warning in Kafka status if wrong annotation applied in KRaftPostMigration State.
```

cluster-operator/src/test/java/io/strimzi/operator/cluster/operator/assembly/KafkaMetadataStateManagerMockTest.java
```diff
@@ -0,0 +1,609 @@
...
+
+        MockKafkaReconciler kr = new MockKafkaReconciler(
+                new Reconciliation("test-trigger", Kafka.RESOURCE_KIND, NAMESPACE, CLUSTER_NAME),
+                vertx,
+                CONFIG,
+                supplier,
+                new PlatformFeaturesAvailability(false, KUBERNETES_VERSION),
+                kafka,
+                List.of(BROKERS, CONTROLLERS),
+                createKafkaCluster(KAFKA),
```
ppatierno, line 212: it's interesting that this doesn't have any impact on the tests but, you are passing `KAFKA` to the `createKafkaCluster` instead of the local variable `kafka`. And that's the same across all the tests. You should fix it.

cluster-operator/src/test/java/io/strimzi/operator/cluster/operator/assembly/KafkaMetadataStateManagerMockTest.java
```diff
@@ -0,0 +1,609 @@
...
+            return agentClient;
+        }
+    }
+
+    /** Mock for KafkaReconciler class */
+    static class MockKafkaReconciler extends KafkaReconciler {
+        private static final ReconciliationLogger LOGGER = ReconciliationLogger.create(MockKafkaReconciler.class.getName());
+
+        public MockKafkaReconciler(Reconciliation reconciliation, Vertx vertx, ClusterOperatorConfig config, ResourceOperatorSupplier supplier, PlatformFeaturesAvailability pfa, Kafka kafka, List<KafkaNodePool> nodePools, KafkaCluster kafkaCluster, ClusterCa clusterCa, ClientsCa clientsCa) {
+            super(reconciliation, kafka, nodePools, createKafkaCluster(kafka), clusterCa, clientsCa, config, supplier, pfa, vertx, new MockKafkaMetadataStateManager(reconciliation, kafka, config.featureGates().useKRaftEnabled()));
```
ppatierno, line 593: I think the `MockKafkaMetadataStateManager` should be passed to the ctor on line 592 and then here.

cluster-operator/src/test/java/io/strimzi/operator/cluster/operator/assembly/KafkaMetadataStateManagerMockTest.java
```diff
@@ -0,0 +1,609 @@
...
+            .build();
+
+    private static KafkaCluster createKafkaCluster(Kafka kafka)   {
+        return KafkaCluster.fromCrd(
+                Reconciliation.DUMMY_RECONCILIATION,
+                kafka,
+                POOLS,
+                VERSIONS,
+                KafkaVersionTestUtils.DEFAULT_ZOOKEEPER_VERSION_CHANGE,
+                KafkaMetadataConfigurationState.ZK,
```
ppatierno, line 142: this should be the state when the reconciliation start, it can't always be `ZK`. It depends on metadata state manager instance and the call of `getMetadataConfigurationState` method on it.

ppatierno (changes requested):

cluster-operator/src/test/java/io/strimzi/operator/cluster/operator/assembly/KafkaMetadataStateManagerMockTest.java
ShubhamRwt: It was because even though I was passing this parameter here, in the constructor It was not being used since I had seperate `createKafkaCluster(kafka)` there which pointed to correct kafka resource.  I have fixed that now

cluster-operator/src/test/java/io/strimzi/operator/cluster/operator/assembly/KafkaMetadataStateManagerMockTest.java
```diff
@@ -0,0 +1,626 @@
...
+                        .withPort(9092)
+                        .withType(KafkaListenerType.INTERNAL)
+                        .withTls(false)
+                        .build())
+            .endKafka()
+            .endSpec()
+            .build();
+
+    private static KafkaCluster createKafkaCluster(Kafka kafka, KafkaMetadataConfigurationState state)   {
+        List<KafkaPool> pools = NodePoolUtils.createKafkaPools(Reconciliation.DUMMY_RECONCILIATION, KAFKA, List.of(CONTROLLERS, BROKERS), Map.of(), Map.of(), true, SHARED_ENV_PROVIDER);
```
ppatierno, line 136:
```suggestion
        List<KafkaPool> pools = NodePoolUtils.createKafkaPools(Reconciliation.DUMMY_RECONCILIATION, kafka, List.of(CONTROLLERS, BROKERS), Map.of(), Map.of(), true, SHARED_ENV_PROVIDER);
```

ppatierno (changes requested):

ppatierno (approved):
LGTM!

cluster-operator/src/test/java/io/strimzi/operator/cluster/operator/assembly/KafkaMetadataStateManagerMockTest.java
scholzj, line 1: `*MockTest` is how we normally call tests using some substantial mock system such as the MockKube or the Connect API mock. We do nto use that name for a regular tests using Mockito. So maybe we can rename this?

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaMetadataStateManager.java
```diff
@@ -172,7 +172,7 @@ public KafkaMetadataConfigurationState getMetadataConfigurationState() {
      */
     public void checkMigrationInProgress(Reconciliation reconciliation, Secret clusterCaCertSecret, Secret coKeySecret, String controllerPodName) {
         if (this.kafkaAgentClient == null) {
-            this.kafkaAgentClient = new KafkaAgentClient(reconciliation, reconciliation.name(), reconciliation.namespace(), clusterCaCertSecret, coKeySecret);
+            this.kafkaAgentClient = initKafkaAgentClient(reconciliation, clusterCaCertSecret, coKeySecret);
```
scholzj, line 175: This seems wrong - although it may not be an issue of this PR, but it inherits the logic already - however:
* This seems to be the only place where the client is used. So why do we have it as a field in the class and not just a local variable or a parameter in the method itself?
* If it needs to be part of the whole class, why not pass the instance in a constructor?

ppatierno: That's a good point. My intention was to avoid creating the `KafkaAgentClient` each time there was the need for check the migration status and a similar approach is used in the `KafkaRoller` but now I realized that, because the check for migration status on the Kafka Agent happens on each reconciliation (and not more times in the same reconcile), a new state manager is created each time, so there is no previous instance of the agent client.
It's enough to be just a local variable here. I will make this change.
The init method was instead useful to Shubham for mocking.

scholzj: Yeah, I'm not strictly against the init method. I'm not strictly against having it as a class field either even if it is used just once. But the way it is it doesn't work together I think.
* If it is a class field, it should be passed in constructor and be final. That way it can be easily passed in tests as a mock as well.
* If it is a local field, I guess I can live with the init method. But I wonder if passing it into the method itself would be nicer and make it easier to test. But I'm not 100% sure whether this is called form in the production code, so maybe the init method is the best solution here.

ppatierno: There is no actual place in the operator where a `KafkaAgentClient` instance is created and provider down the line of different components. It's only used in the `KafkaRoller` which instatiates it internally, so I think that it's the way to go in this case as well, instead of creating outside and passing it down. Its usage is limited.

ppatierno: I made the client local.

scholzj (approved):

PR #9627, opened 2024-01-31
Use KRaft only on Testing Farm to avoid unexpected ZK failures

Frawless:
### Type of change

- Enhancement / new feature

### Description

Because we are closer to KRaft to be production ready, we would like to start using Testing Farm for system tests. This PR removes option to run ZK mode on TF as it was failing with unexpected dns errors caused by ZK bug and move all pipelines to use KRaft mode. 

The pipelines will still be executable only via comments on the PRs.

### Checklist

### DIFF
--- a/test/src/main/java/io/strimzi/test/k8s/cluster/Kind.java
+++ b/test/src/main/java/io/strimzi/test/k8s/cluster/Kind.java
@@ -47,7 +47,7 @@ public class Kind implements KubeCluster {
     public boolean isClusterUp() {
         List<String> cmd = Arrays.asList("kubectl", "get", "nodes", "-o", "jsonpath='{.items[*].spec.providerID}'");
         try {
-            return Exec.exec(cmd).out().contains("kind-cluster");
+            return Exec.exec(cmd).out().contains("kind://");
         } catch (KubeClusterException e) {
             LOGGER.debug("'" + String.join(" ", cmd) + "' failed. Please double check connectivity to your cluster!");
             LOGGER.debug(e);
--- a/test/src/main/java/io/strimzi/test/k8s/cluster/Kind.java
+++ b/test/src/main/java/io/strimzi/test/k8s/cluster/Kind.java
@@ -47,7 +47,9 @@ public class Kind implements KubeCluster {
     public boolean isClusterUp() {
         List<String> cmd = Arrays.asList("kubectl", "get", "nodes", "-o", "jsonpath='{.items[*].spec.providerID}'");
         try {
-            return Exec.exec(cmd).out().startsWith("kind");
+            String cmdOut = Exec.exec(cmd).out();
+            LOGGER.debug("IsKindUp cmd output: " + cmdOut);
+            return cmdOut.contains("kind");
         } catch (KubeClusterException e) {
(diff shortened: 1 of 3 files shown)


kornys:
👍

Frawless:
I am not going to run the full regression as part of this. I will try to check results as part of different PR and fix the failures if any occur.

test/src/main/java/io/strimzi/test/k8s/cluster/Kind.java
```diff
@@ -47,7 +47,7 @@ public boolean isAvailable() {
     public boolean isClusterUp() {
         List<String> cmd = Arrays.asList("kubectl", "get", "nodes", "-o", "jsonpath='{.items[*].spec.providerID}'");
         try {
-            return Exec.exec(cmd).out().startsWith("kind");
+            return Exec.exec(cmd).out().contains("kind-cluster");
```
im-konge, line 50: @see-quick @jankalinic when you were testing the Kind stuff, was the name of the node just "kind" or "kind-cluster"? I'm wondering if there is some kind of label or something on the node that would say for 100% that it's Kind cluster and not something else.

see-quick: It would be based on
```
cat <<EOF | kind create cluster --config=-
    kind: Cluster
    apiVersion: kind.x-k8s.io/v1alpha4
    nodes:
    - role: control-plane
    - role: control-plane
    - role: control-plane
    - role: worker
    - role: worker
    - role: worker
    name: kind-cluster
    containerdConfigPatches:
    - |-
     [plugins."io.containerd.grpc.v1.cri".registry]
       config_path = "/etc/containerd/certs.d"
    networking:
       ipFamily: $IP_FAMILY
```
config defined in `setup-kind.sh`

see-quick: And here you can see it is `kind-cluster`. When you run by default kind on machine I think it is named `kind`

jankalinic: i was dealing with this when setting up STs and what i came up with was that i found official support for nodes to be in form of ` provider-id: "kind://{{.NodeProvider}}/{{.ClusterName}}/{{.NodeName}}"`

jankalinic: it is based on https://github.com/kubernetes-sigs/kind/blob/40c81f187425254daf2bf84360a6257a278252df/pkg/cluster/internal/kubeadm/config.go

im-konge: Maybe obtaining info from apiVersion 
```
apiVersion: kind.x-k8s.io/v1alpha4
```
would make sense?

Because you can anyway (or any user can) have a different name defined there or?

Frawless: ok going to use `kind://` based on kind code.

kornys (approved):

im-konge (approved):
LGTM, thanks

see-quick (approved):

PR #9480, opened 2023-12-21
ZooKeeper to KRaft migration

ppatierno:
### Type of change

- Enhancement / new feature

### Description

This PR fixes #9433 , #9447 and #9448 about implementing the ZooKeeper to KRaft migration process.

The current status is about starting from a ZooKeeper-based cluster, creating a node pool with controllers, applying the `strimzi.io/kraft: migration` annotation on the `Kafka` CR and allowing the operator FSM to start the migration up to the "dual-write" (so creating controllers, rolling brokers, start migration). Then the user can apply the `strimzi.io/kraft: enabled` annotation in order to finilize the process by having the full KRaft-based cluster,

The PR won't address documentation. There will be a different PR for it. See https://github.com/strimzi/strimzi-kafka-operator/pull/9633
~~This PR doesn't cover migration rollback and ZooKeeper removal as well which have different issues and will have different PRs for them.~~
This PR covers migration rollback ~~but taking into account this Kafka issue https://issues.apache.org/jira/browse/KAFKA-16101 and based on deleting the `/controller` znode manually as last step.~~ using `rollback` annotation value from the `KRaftPostMigration` state, the `disabled` from the `KRaftDualWriting`.
This PR also covers the ZooKeeper related resources removal when the migration ends.

### Checklist

_Please go through this checklist and make sure all applicable tasks have been done_

- [x] Write tests
- [x] Make sure all tests pass
- [x] Try your changes from Pod inside your Kubernetes and OpenShift cluster, not just locally
- [x] Reference relevant issue(s) and close them after merging
- [ ] Update CHANGELOG.md

PR #9633, opened 2024-02-02
docs(kraft): adds a procedure for zookeeper to kraft migration

PaulRMellor:
**Documentation**

Adds a new procedure that describes how to migrate from ZooKeeper-based to KRaft-based Kafka cluster.
Also includes steps to roll back, if required.

### Checklist

- [ ] Write tests
- [x] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/documentation/assemblies/upgrading/assembly-upgrade-zookeeper.adoc
+++ b/documentation/assemblies/upgrading/assembly-upgrade-zookeeper.adoc
@@ -8,6 +8,8 @@
 [role="_abstract"]
 If you are using a ZooKeeper-based Kafka cluster, an upgrade requires an update to the Kafka version and the inter-broker protocol version.
 
+If you want to switch from using a ZooKeeper-based Kafka cluster to a KRaft-based cluster as part of your upgrade, see xref:proc-deploy-migrate-kraft-str[].
+
 include::../../modules/upgrading/ref-upgrade-kafka-versions.adoc[leveloffset=+1]
 include::../../modules/upgrading/con-upgrade-older-clients.adoc[leveloffset=+1]
 
--- a/documentation/assemblies/upgrading/assembly-upgrade-zookeeper.adoc
+++ b/documentation/assemblies/upgrading/assembly-upgrade-zookeeper.adoc
@@ -8,7 +8,7 @@
 [role="_abstract"]
 If you are using a ZooKeeper-based Kafka cluster, an upgrade requires an update to the Kafka version and the inter-broker protocol version.
 
-If you want to switch a kafka cluster from using a ZooKeeper for metadata management to operating in KRaft mode, the steps must be performed separately from the upgrade. 
+If you want to switch a Kafka cluster from using a ZooKeeper for metadata management to operating in KRaft mode, the steps must be performed separately from the upgrade. 
 For information on migrating to a KRaft-based cluster, see xref:proc-deploy-migrate-kraft-str[].
 
 include::../../modules/upgrading/ref-upgrade-kafka-versions.adoc[leveloffset=+1]
--- a/documentation/assemblies/upgrading/assembly-upgrade-zookeeper.adoc
+++ b/documentation/assemblies/upgrading/assembly-upgrade-zookeeper.adoc
@@ -8,7 +8,7 @@
 [role="_abstract"]
 If you are using a ZooKeeper-based Kafka cluster, an upgrade requires an update to the Kafka version and the inter-broker protocol version.
 
-If you want to switch from using a ZooKeeper-based Kafka cluster to a KRaft-based cluster, the steps must be performed separately from the upgrade. 
+If you want to switch a kafka cluster from using a ZooKeeper for metadata management to operating in KRaft mode, the steps must be performed separately from the upgrade. 
 For information on migrating to a KRaft-based cluster, see xref:proc-deploy-migrate-kraft-str[].
 
 include::../../modules/upgrading/ref-upgrade-kafka-versions.adoc[leveloffset=+1]
--- a/documentation/assemblies/upgrading/assembly-upgrade-zookeeper.adoc
+++ b/documentation/assemblies/upgrading/assembly-upgrade-zookeeper.adoc
@@ -8,7 +8,7 @@
 [role="_abstract"]
 If you are using a ZooKeeper-based Kafka cluster, an upgrade requires an update to the Kafka version and the inter-broker protocol version.
 
-If you want to switch a Kafka cluster from using a ZooKeeper for metadata management to operating in KRaft mode, the steps must be performed separately from the upgrade. 
+If you want to switch a Kafka cluster from using ZooKeeper for metadata management to operating in KRaft mode, the steps must be performed separately from the upgrade. 
 For information on migrating to a KRaft-based cluster, see xref:proc-deploy-migrate-kraft-str[].
 
 include::../../modules/upgrading/ref-upgrade-kafka-versions.adoc[leveloffset=+1]
--- a/documentation/assemblies/upgrading/assembly-upgrade-zookeeper.adoc
+++ b/documentation/assemblies/upgrading/assembly-upgrade-zookeeper.adoc
@@ -8,7 +8,8 @@
 [role="_abstract"]
 If you are using a ZooKeeper-based Kafka cluster, an upgrade requires an update to the Kafka version and the inter-broker protocol version.
 
-If you want to switch from using a ZooKeeper-based Kafka cluster to a KRaft-based cluster as part of your upgrade, see xref:proc-deploy-migrate-kraft-str[].
+If you want to switch from using a ZooKeeper-based Kafka cluster to a KRaft-based cluster, the steps must be performed separately from the upgrade. 
+For information on migrating to a KRaft-based cluster, see xref:proc-deploy-migrate-kraft-str[].
 
 include::../../modules/upgrading/ref-upgrade-kafka-versions.adoc[leveloffset=+1]
 include::../../modules/upgrading/con-upgrade-older-clients.adoc[leveloffset=+1]
--- a/documentation/modules/deploying/proc-deploy-migrate-kraft.adoc
+++ b/documentation/modules/deploying/proc-deploy-migrate-kraft.adoc
@@ -82,7 +82,7 @@ Errors related to using controllers in a ZooKeeper-based environment are expecte
 +
 [source,shell]
 ----
-kubectl annotate kafka my-cluster strimzi.io/kraft: migration
+kubectl annotate kafka my-cluster strimzi.io/kraft="migration" --overwrite
 ----
 +
 .Enabling KRaft migration
@@ -159,7 +159,7 @@ The migration status is also represented in the `status.kafkaMetadataState` prop
 +
 [source,shell]
 ----
-kubectl annotate kafka my-cluster strimzi.io/kraft: enabled
+kubectl annotate kafka my-cluster strimzi.io/kraft="enabled"
 ----
 +
 .Enabling KRaft migration
@@ -210,7 +210,7 @@ Before the migration is finalized by enabling KRaft in the `Kafka` resource,  an
 +
 [source,shell]
 ----
-kubectl annotate kafka my-cluster strimzi.io/kraft: rollback
+kubectl annotate kafka my-cluster strimzi.io/kraft="rollback"
 ----
 +
 .Rolling back KRaft migration
@@ -240,7 +240,7 @@ kubectl delete KafkaNodePool controller -n my-project
 +
 [source,shell]
 ----
-kubectl annotate kafka my-cluster strimzi.io/kraft: disabled
+kubectl annotate kafka my-cluster strimzi.io/kraft="disabled"
 ----
 +
 .Switching back to using ZooKeeper
--- a/documentation/modules/deploying/proc-deploy-migrate-kraft.adoc
+++ b/documentation/modules/deploying/proc-deploy-migrate-kraft.adoc
@@ -150,12 +150,13 @@ but controllers are managing metadata in Kafka and ZooKeeper.
 Brokers are rolled a second time to remove the flag to enable migration.
 * `KRaftPostMigration` is the state when KRaft mode is enabled for brokers and there is no ZooKeeper involvement. 
 Controllers are still managing in Kafka and ZooKeeper.
-You can xref:proc-deploy-migrate-kraft-rollback-{context}[roll back from this point].
 --
 +
 The migration status is also represented in the `status.kafkaMetadataState` property of the `Kafka` resource. 
 +
-WARNING: The next step is to enable KRaft. Rollback cannot be performed after enabling KRaft.
+WARNING: You can xref:proc-deploy-migrate-kraft-rollback-{context}[roll back to using ZooKeeper from this point]. 
+The next step is to enable KRaft. 
+Rollback cannot be performed after enabling KRaft.
 
 . Enable KRaft in the `Kafka` resource configuration by setting the `strimzi.io/kraft` annotation to `enabled`:
 +
@@ -213,7 +214,7 @@ Before the migration is finalized by enabling KRaft in the `Kafka` resource,  an
 +
 [source,shell]
 ----
-kubectl annotate kafka my-cluster strimzi.io/kraft="rollback"
+kubectl annotate kafka my-cluster strimzi.io/kraft="rollback" --overwrite
 ----
 +
 .Rolling back KRaft migration
--- a/documentation/modules/deploying/proc-deploy-migrate-kraft.adoc
+++ b/documentation/modules/deploying/proc-deploy-migrate-kraft.adoc
@@ -34,7 +34,7 @@ To migrate when the cluster is not using node pools, brokers must be contained i
 Support for node pools is enabled in the `Kafka` resource configuration using the `strimzi.io/node-pools: enabled` annotation.
 
 In this procedure, the Kafka cluster name is `my-cluster`, which is located in the `my-project` namespace. 
-The name of the controller node pool is `controller`.
+The name of the controller node pool created is `controller`, which is required even if you are already using node pools.
 
 .Procedure
 
--- a/documentation/modules/deploying/proc-deploy-migrate-kraft.adoc
+++ b/documentation/modules/deploying/proc-deploy-migrate-kraft.adoc
@@ -22,6 +22,9 @@ Note also, the following:
 .Prerequisites
 
 * You must be using Strimzi 0.40 or newer with Kafka 3.7.0 or newer. If you using an earlier version of Strimzi or Apache Kafka, upgrade before migrating to KRaft mode.
+* Verify that the ZooKeeper-based deployment is operating without the following, as they are not supported in KRaft mode:
+** The Topic Operator running in bidirectional mode. It should either be in unidirectional mode or disabled.
+** JBOD storage. While the `jbod` storage type can be used, the JBOD array must contain only one disk.
 * The Cluster Operator that manages the Kafka cluster is running.
 * The Kafka cluster deployment uses Kafka node pools.
 +
@@ -107,14 +110,16 @@ kubectl get pods -n my-project
 [source,shell]
 ----
 NAME                     READY  STATUS   RESTARTS
-my-cluster-broker-0      1/1    Running  0
-my-cluster-broker-1      1/1    Running  0
-my-cluster-broker-2      1/1    Running  0
+my-cluster-kafka-0       1/1    Running  0
+my-cluster-kafka-1       1/1    Running  0
+my-cluster-kafka-2       1/1    Running  0
 my-cluster-controller-3  1/1    Running  0
 my-cluster-controller-4  1/1    Running  0
 my-cluster-controller-5  1/1    Running  0
 # ...
 ----
++
+Here, the brokers have the name `my-cluster-kafka` as they were contained in a node pool named `kafka` when migrating the cluster to use node pools.
 
 . Check the status of the migration:
 +
--- a/documentation/modules/deploying/proc-deploy-migrate-kraft.adoc
+++ b/documentation/modules/deploying/proc-deploy-migrate-kraft.adoc
@@ -34,7 +34,8 @@ To migrate when the cluster is not using node pools, brokers must be contained i
 Support for node pools is enabled in the `Kafka` resource configuration using the `strimzi.io/node-pools: enabled` annotation.
 
 In this procedure, the Kafka cluster name is `my-cluster`, which is located in the `my-project` namespace. 
-The name of the controller node pool created is `controller`, which is required even if you are already using node pools.
+The name of the controller node pool created is `controller`.
+The node pool for the brokers is called `kafka`.
 
 .Procedure
 
@@ -118,9 +119,6 @@ my-cluster-controller-4  1/1    Running  0
 my-cluster-controller-5  1/1    Running  0
 # ...
 ----
-+
-Here, the brokers have the name `my-cluster-kafka` as they were contained in a node pool named `kafka` when migrating the cluster to use node pools.
-But the brokers can have any name if they are already contained in a node pool.
 
 . Check the status of the migration:
 +
--- a/documentation/modules/deploying/proc-deploy-migrate-kraft.adoc
+++ b/documentation/modules/deploying/proc-deploy-migrate-kraft.adoc
@@ -198,12 +198,23 @@ my-cluster  ...  KRaft
 ----
 +
 --
-* `PreKRaft` is the state when all ZooKeeper-related configurations and resources have been automatically deleted.
+* `PreKRaft` is the state when all ZooKeeper-related resources have been automatically deleted.
 * `KRaft` is the final state (after the controllers have rolled) when the KRaft migration is finalized.
 --
 +
 NOTE: Depending on how `deleteClaim` is configured for ZooKeeper, its Persistent Volume Claims (PVCs) and persistent volumes (PVs) may not be deleted.
 `deleteClaim` specifies whether the PVC is deleted when the cluster is uninstalled. The default is `false`.
+
+. Remove any ZooKeeper-related configuration from the `Kafka` resource.
++
+If present, you can remove the following:
++
+* `log.message.format.version`
+* `inter.broker.protocol.version` 
+* `spec.zookeeper.*` properties 
++
+Removing `log.message.format.version` and `inter.broker.protocol.version` causes the brokers and controllers to roll again.
+Removing ZooKeeper properties removes any warning messages related to ZooKeeper configuration being present in a KRaft-operated cluster.  
 
 [id='proc-deploy-migrate-kraft-rollback-{context}']
 .Performing a rollback on the migration
--- a/documentation/modules/deploying/proc-deploy-migrate-kraft.adoc
+++ b/documentation/modules/deploying/proc-deploy-migrate-kraft.adoc
@@ -10,8 +10,8 @@ If you are using ZooKeeper for metadata management in your Kafka cluster, you ca
 KRaft mode replaces ZooKeeper for distributed coordination, offering enhanced reliability, scalability, and throughput.
 
 During the migration, you install a quorum of controller nodes as a node pool, which replaces ZooKeeper for management of your cluster. 
-You enable KRaft migration in the cluster configuration by applying the `strimzi.io/kraft: migration` annotation.  
-After the migration is complete, you switch the brokers to using KRaft and the controllers out of migration mode using the `strimzi.io/kraft: enabled` annotation.
+You enable KRaft migration in the cluster configuration by applying the `strimzi.io/kraft="migration"` annotation.  
+After the migration is complete, you switch the brokers to using KRaft and the controllers out of migration mode using the `strimzi.io/kraft="enabled"` annotation.
 
 Before starting the migration, verify that your environment can support Kafka in KRaft mode, as there are a number of xref:ref-operator-use-kraft-feature-gate-str[limitations].
 Note also, the following:
@@ -76,7 +76,7 @@ NOTE: For the migration, you cannot use a node pool of nodes that share the brok
 
 . Apply the new `KafkaNodePool` resource to create the controllers.
 +
-Errors related to using controllers in a ZooKeeper-based environment are expected.
+Errors related to using controllers in a ZooKeeper-based environment are expected in the Cluster Operator logs.
 
 . Enable KRaft migration in the `Kafka` resource by setting the `strimzi.io/kraft` annotation to `migration`:
 +
@@ -94,7 +94,7 @@ metadata:
   name: my-cluster
   namespace: my-project
   annotations:
-    strimzi.io/kraft: migration 
+    strimzi.io/kraft="migration" 
 # ...
 ----
 +
@@ -154,12 +154,14 @@ You can xref:proc-deploy-migrate-kraft-rollback-{context}[roll back from this po
 --
 +
 The migration status is also represented in the `status.kafkaMetadataState` property of the `Kafka` resource. 
++
+WARNING: The next step is to enable KRaft. Rollback cannot be performed after enabling KRaft.
 
 . Enable KRaft in the `Kafka` resource configuration by setting the `strimzi.io/kraft` annotation to `enabled`:
 +
 [source,shell]
 ----
-kubectl annotate kafka my-cluster strimzi.io/kraft="enabled"
+kubectl annotate kafka my-cluster strimzi.io/kraft="enabled" --overwrite
 ----
 +
 .Enabling KRaft migration
@@ -171,11 +173,9 @@ metadata:
   name: my-cluster
   namespace: my-project
   annotations:
-    strimzi.io/kraft: enabled 
+    strimzi.io/kraft="enabled" 
 # ...
 ----
-+
-WARNING: Rollback cannot be performed after enabling KRaft.
 
 . Check the status of the move to full KRaft mode:
 +
@@ -200,13 +200,16 @@ my-cluster  ...  KRaft
 * `PreKRaft` is the state when all ZooKeeper-related configurations and resources have been automatically deleted.
 * `KRaft` is the final state (after the controllers have rolled) when the KRaft migration is finalized.
 --
++
+NOTE: Depending on how `deleteClaim` is configured for ZooKeeper, its Persistent Volume Claims (PVCs) and persistent volumes (PVs) may not be deleted.
+`deleteClaim` specifies whether the PVC is deleted when the cluster is uninstalled. The default is `false`.
 
 [id='proc-deploy-migrate-kraft-rollback-{context}']
 .Performing a rollback on the migration
 
 Before the migration is finalized by enabling KRaft in the `Kafka` resource,  and the state has moved to the `KRaft` state, you can perform a rollback operation as follows:
 
-. Apply the `strimzi.io/kraft: rollback` annotation to the `Kafka` resource to roll back the brokers.
+. Apply the `strimzi.io/kraft="rollback"` annotation to the `Kafka` resource to roll back the brokers.
 +
 [source,shell]
 ----
@@ -222,7 +225,7 @@ metadata:
   name: my-cluster
   namespace: my-project
   annotations:
-    strimzi.io/kraft: rollback 
+    strimzi.io/kraft="rollback" 
 # ...
 ----
 +
@@ -236,11 +239,11 @@ The brokers are rolled back so that they can be connected to ZooKeeper again and
 kubectl delete KafkaNodePool controller -n my-project
 ----
 
-. Apply the `strimzi.io/kraft: disabled` annotation to the `Kafka` resource to return the metadata state to `ZooKeeper`.
+. Apply the `strimzi.io/kraft="disabled"` annotation to the `Kafka` resource to return the metadata state to `ZooKeeper`.
 +
 [source,shell]
 ----
-kubectl annotate kafka my-cluster strimzi.io/kraft="disabled"
+kubectl annotate kafka my-cluster strimzi.io/kraft="disabled" --overwrite
 ----
 +
 .Switching back to using ZooKeeper
@@ -252,6 +255,6 @@ metadata:
   name: my-cluster
   namespace: my-project
   annotations:
-    strimzi.io/kraft: disabled 
+    strimzi.io/kraft="disabled" 
 # ...
 ----
--- a/documentation/modules/deploying/proc-deploy-migrate-kraft.adoc
+++ b/documentation/modules/deploying/proc-deploy-migrate-kraft.adoc
@@ -19,17 +19,14 @@ Note also, the following:
 * Migration is only supported on dedicated controller nodes, not on nodes with dual roles as brokers and controllers.
 * Throughout the migration process, ZooKeeper and Controller nodes operate in parallel for a period, requiring sufficient compute resources in the cluster.
 
-WARNING: ZooKeeper to KRaft migration is not yet suitable for production environments.
-
 .Prerequisites
 
-* You must be using Strimzi 0.40 or newer with Kafka 3.6.1 or newer. If you using an earlier version of Strimzi or APache Kafka, upgrade before migrating to KRaft mode.
+* You must be using Strimzi 0.40 or newer with Kafka 3.7.0 or newer. If you using an earlier version of Strimzi or Apache Kafka, upgrade before migrating to KRaft mode.
 * The Cluster Operator that manages the Kafka cluster is running.
 * The Kafka cluster deployment uses Kafka node pools.
 +
 You can xref:proc-migrating-clusters-node-pools-str[migrate your existing Kafka cluster to use node pools]. 
-+
-In the `KafkaNodePool` resource configuration, brokers must be contained in a node pool assigned a `broker` role and with the name `kafka`.
+To migrate, the `KafkaNodePool` resource configuration, brokers must be contained in a node pool assigned a `broker` role and with the name `kafka`.
 Support for node pools is enabled in the `Kafka` resource configuration using the `strimzi.io/node-pools: enabled` annotation.
 
 In this procedure, the Kafka cluster name is `my-cluster`, which is located in `my-project`. 
@@ -128,10 +125,11 @@ kubectl get kafka my-cluster -n my-project -w
 .Updates to the metadata state
 [source,shell]
 ----
-NAME        ...  METADATA STATE   WARNINGS
+NAME        ...  METADATA STATE
 my-cluster  ...  Zookeeper
 my-cluster  ...  KRaftMigration
 my-cluster  ...  KRaftDualWriting
+my-cluster  ...  KRaftPostMigration
 ----
 +
 `METADATA STATE` shows the mechanism used to manage Kafka metadata and coordinate operations.
@@ -139,9 +137,14 @@ At the start of the migration this is `ZooKeeper`.
 +
 --
 * `ZooKeeper` is the initial state when metadata is only stored in ZooKeeper.
-* `KRaftMigration` is the state when the migration is in progress.
-* `KRaftDualWriting` is the state when the migration has ended and the cluster is working as a KRaft cluster. 
-Metadata is managed in Kafka but also replicated in ZooKeeper, so you can xref:proc-deploy-migrate-kraft-rollback-{context}[roll back from this point].
+* `KRaftMigration` is the state when the migration is in progress and brokers are rolled to register with the controllers.
+The migration can take some time at this point depending on the number of topics and partitions in the cluster. 
+* `KRaftDualWriting` is the state when the Kafka cluster is working as a KRaft cluster, 
+but controllers are managing metadata in Kafka and ZooKeeper. 
+Brokers are rolled a second time to remove the ZooKeeper configuration and migration annotation from the `Kafka` resource.
+* `KRaftPostMigration` is the state when KRaft mode is enabled for brokers and there is no ZooKeeper involvement. 
+Controllers are still connected to ZooKeeper.
+You can xref:proc-deploy-migrate-kraft-rollback-{context}[roll back from this point].
 --
 +
 The migration status is also represented in the `status.kafkaMetadataState` property of the `Kafka` resource. 
@@ -178,38 +181,26 @@ kubectl get kafka my-cluster -n my-project -w
 .Updates to the metadata state
 [source,shell]
 ----
-NAME        ...  METADATA STATE        WARNINGS
+NAME        ...  METADATA STATE
 my-cluster  ...  Zookeeper
 my-cluster  ...  KRaftMigration
 my-cluster  ...  KRaftDualWriting
 my-cluster  ...  KRaftPostMigration
-my-cluster  ...  KRaft                 True
+my-cluster  ...  KRaft             
 ----
 +
-`KRaftPostMigration` is the state (after the brokers have rolled) when KRaft mode is enabled and there is no ZooKeeper involvement, though ZooKeeper pods are still running.
-`KRaft` is the final state (after the controllers have rolled) when the KRaft migration has finalized.
+--
+* `KRaft` is the final state (after the controllers have rolled) when the KRaft migration is finalized.
+--
 +
-The warning (`True`) relates to ZooKeeper configuration being present in the `Kafka` resource configuration.
-+
-. Clean up the deployment to remove Zookeeper configuration and resources.
-
-.. Remove `inter.broker.protocol.version`, `log.message.format.version`, and all `spec.zookeeper` configuration properties from the `Kafka` resource.
-.. Apply the changes to the `Kafka` resource configuration.
-.. Delete any resources related to the ZooKeeper deployment:
-+
-[source,shell]
-----
-kubectl delete <resource_type> <resource_name> -n my-project
-----
-+
-For a list of resources created for ZooKeeper, see xref:ref-list-of-kafka-cluster-resources-str[].
+All ZooKeeper-related resources have been automatically deleted.
 
 [id='proc-deploy-migrate-kraft-rollback-{context}']
 .Performing a rollback on the migration
 
-Before the migration is finalized by enabling KRaft in the 'Kafka' resource, you can perform a rollback operation as follows:
+Before the migration is finalized by enabling KRaft in the `Kafka` resource,  and the state has moved to the `KRaft` state, you can perform a rollback operation as follows:
 
-. Apply the `strimzi.io/kraft: rollback` annotation to the `Kafka` resource to roll back the brokers and controllers.
+. Apply the `strimzi.io/kraft: rollback` annotation to the `Kafka` resource to roll back the brokers.
 +
 [source,shell]
 ----
@@ -228,6 +219,8 @@ metadata:
     strimzi.io/kraft: rollback 
 # ...
 ----
++
+The brokers are rolled back so that they can be connected to ZooKeeper again and the state returns to `KRaftDualWriting`.
 
 . Delete the controllers node pool:
 +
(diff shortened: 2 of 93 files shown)


documentation/modules/deploying/proc-deploy-migrate-kraft.adoc
```diff
@@ -0,0 +1,253 @@
+// Module included in the following assemblies:
+//
+// deploying/deploying.adoc
+
+[id='proc-deploy-migrate-kraft-{context}']
+= Migrating to KRaft mode
+
+[role="_abstract"]
+If you are using ZooKeeper for cluster management of your Kafka cluster, you can migrate to using Kafka in KRaft mode.
```
scholzj, line 9: I think maybe this is better? Cluster management that is what Strimzi does 😉.

```suggestion
If you are using ZooKeeper for metadata management in your Kafka cluster, you can migrate to using Kafka in KRaft mode. 
```

documentation/modules/deploying/proc-deploy-migrate-kraft.adoc
```diff
@@ -0,0 +1,253 @@
...
+If you are using ZooKeeper for cluster management of your Kafka cluster, you can migrate to using Kafka in KRaft mode. 
+KRaft mode replaces ZooKeeper for distributed coordination, offering enhanced reliability, scalability, and throughput.
+
+During the migration, you install a quorum of controller nodes as a node pool, which replaces ZooKeeper for management of your cluster. 
+You enable KRaft migration in the cluster configuration by applying the `strimzi.io/kraft: migration` annotation.  
+After the migration is complete, you switch the brokers to using KRaft and the controllers out of migration mode using the `strimzi.io/kraft: enabled` annotation.
+
+Before starting the migration, verify that your environment can support Kafka in KRaft mode, as there are a number of xref:ref-operator-use-kraft-feature-gate-str[limitations].
+
+WARNING: ZooKeeper to KRaft migration is not suitable for production environments.
```
scholzj, line 18: Is this the case? What is the reason?

If nothing else, we should make clear it is not **yet** suitable.

```suggestion
WARNING: ZooKeeper to KRaft migration is not yet suitable for production environments.
```

documentation/modules/deploying/proc-deploy-migrate-kraft.adoc
```diff
@@ -0,0 +1,253 @@
...
+
+[role="_abstract"]
+If you are using ZooKeeper for cluster management of your Kafka cluster, you can migrate to using Kafka in KRaft mode. 
+KRaft mode replaces ZooKeeper for distributed coordination, offering enhanced reliability, scalability, and throughput.
+
+During the migration, you install a quorum of controller nodes as a node pool, which replaces ZooKeeper for management of your cluster. 
+You enable KRaft migration in the cluster configuration by applying the `strimzi.io/kraft: migration` annotation.  
+After the migration is complete, you switch the brokers to using KRaft and the controllers out of migration mode using the `strimzi.io/kraft: enabled` annotation.
+
+Before starting the migration, verify that your environment can support Kafka in KRaft mode, as there are a number of xref:ref-operator-use-kraft-feature-gate-str[limitations].
```
scholzj, line 16: Might be worth noting that:
* Migration can be done only to dedicated controller nodes and not to mixed nodes
* During the migration, users need to run for some time the ZooKeeper and Controller nodes in parallel and should plan to have sufficient compute resources available in their cluster.

documentation/modules/deploying/proc-deploy-migrate-kraft.adoc
```diff
@@ -0,0 +1,253 @@
...
+
+WARNING: ZooKeeper to KRaft migration is not suitable for production environments.
+
+.Prerequisites
+
+* You must be using Strimzi 0.40 (with Kafka 3.7) or newer. If you using an earlier version of AMQ Streams, upgrade before migrating to KRaft mode.
```
scholzj, line 22:
```suggestion
* You must be using Strimzi 0.40 or newer with Kafka 3.6.1 or newer. If you using an earlier version of Strimzi or Apache Kafka, upgrade before migrating to KRaft mode.
```

documentation/modules/deploying/proc-deploy-migrate-kraft.adoc
```diff
@@ -0,0 +1,253 @@
...
+* The Cluster Operator that manages the Kafka cluster is running.
+* The Kafka cluster deployment uses Kafka node pools.
++
+You can xref:proc-migrating-clusters-node-pools-str[migrate your existing Kafka cluster to use node pools]. 
++
+In the `KafkaNodePool` resource configuration, brokers must be contained in a node pool assigned a `broker` role and with the name `kafka`.
+Support for node pools is enabled in the `Kafka` resource configuration using the `strimzi.io/node-pools: enabled` annotation.
+
+NOTE: Currently, replica and storage configuration properties in the `KafkaNodePool` resource must also be present in the `Kafka` resource. The configuration in the `Kafka` resource is ignored when node pools are used.
```
scholzj, line 31: Not the case anymore. Should be deleted.

documentation/modules/deploying/proc-deploy-migrate-kraft.adoc
```diff
@@ -0,0 +1,253 @@
...
+--
+* `ZooKeeper` is the initial state when metadata is only stored in ZooKeeper.
+* `KRaftMigration` is the state when the migration is in progress.
+* `KRaftDualWriting` is the state when the migration has ended and the cluster is working as a KRaft cluster. 
+Metadata is managed in Kafka but also replicated in ZooKeeper, so you can xref:proc-deploy-migrate-kraft-rollback-{context}[roll back from this point].
+--
++
+The migration status is also represented in the `status.KafkaMetadataState` property of the `Kafka` resource. 
+
+. Enable KRaft in the `Kafka` resource configuration by setting the `strimzi.io/kraft` annotation to `enabled`:
```
scholzj, line 147: Please add a clear warning that there is no rollback from here aynmore.

documentation/modules/deploying/proc-deploy-migrate-kraft.adoc
```diff
@@ -0,0 +1,253 @@
...
+`KRaftPostMigration` is the state (after the brokers have rolled) when KRaft mode is enabled and there is no ZooKeeper involvement, though ZooKeeper pods are still running.
+`KRaft` is the final state (after the controllers have rolled) when the KRaft migration has finalized.
++
+The warning (`True`) relates to ZooKeeper configuration being present in the `Kafka` resource configuration.
++
+. Clean up the deployment to remove Zookeeper configuration and resources.
+
+.. Remove `inter.broker.protocol.version`, `log.message.format.version`, and all `spec.zookeeper` configuration properties from the `Kafka` resource.
+.. Apply the changes to the `Kafka` resource configuration.
+.. Delete any resources related to the ZooKeeper deployment:
```
scholzj, line 194: Isn't this done automatically?

documentation/modules/deploying/proc-deploy-migrate-kraft.adoc
```diff
@@ -0,0 +1,253 @@
...
+----
+kubectl delete <resource_type> <resource_name> -n my-project
+----
++
+For a list of resources created for ZooKeeper, see xref:ref-list-of-kafka-cluster-resources-str[].
+
+[id='proc-deploy-migrate-kraft-rollback-{context}']
+.Performing a rollback on the migration
+
+Before the migration is finalized, you can perform a rollback operation as follows:
```
scholzj, line 206: Needs to be more clear what _finalized_ means here.

documentation/modules/deploying/proc-deploy-migrate-kraft.adoc
PaulRMellor: @ppatierno -- do we need this?  (We have the same note in rhel doc)

documentation/modules/deploying/proc-deploy-migrate-kraft.adoc
PaulRMellor: I think Paolo was working towards that. Maybe we're there.

documentation/modules/deploying/proc-deploy-migrate-kraft.adoc
PaulRMellor: we have a few notes in the guide along this line:  "IMPORTANT
KRaft mode is not ready for production in Apache Kafka or in Strimzi."

scholzj: I think it is ready in Apache Kafka (whether it should be or not, that is a different story). If we don't think it is ready yet in Strimzi, we should make clear it will be. But we also need to understand what is missing and when it will be ready and I think this is currently not clear.

ppatierno: From a Strimzi perspective, I consider the migration procedure ready (even for production) but of course the codebase has to be reviewed and approved. From an Apache Kafka perspective, it's defined this way as well but as Jakub says, if it's real or not it's a different story.

documentation/modules/deploying/proc-deploy-migrate-kraft.adoc
ppatierno: Yes we are there now. All the ZooKeeper resources are automatically deleted now.

documentation/assemblies/upgrading/assembly-upgrade-zookeeper.adoc
```diff
@@ -8,6 +8,9 @@
 [role="_abstract"]
 If you are using a ZooKeeper-based Kafka cluster, an upgrade requires an update to the Kafka version and the inter-broker protocol version.
 
+If you want to switch from using a ZooKeeper-based Kafka cluster to a KRaft-based cluster, the steps must be performed separately from the upgrade.
```
ppatierno, line 11: This sentence let me think we are suggesting to switch from a cluster to another. Something like "you can move from using a ZooKeeper-based cluster to a different one which is KRaft-based". I think it's confusing. It should be more referring to the migration operation, making clear you are going to use the same cluster but through migration you are moving out from using ZooKeeper for metadata and using KRaft instead.

documentation/modules/deploying/proc-deploy-migrate-kraft.adoc
```diff
@@ -0,0 +1,257 @@
...
+Note also, the following:
+
+* Migration is only supported on dedicated controller nodes, not on nodes with dual roles as brokers and controllers.
+* Throughout the migration process, ZooKeeper and Controller nodes operate in parallel for a period, requiring sufficient compute resources in the cluster.
+
+WARNING: ZooKeeper to KRaft migration is not yet suitable for production environments.
+
+.Prerequisites
+
+* You must be using Strimzi 0.40 or newer with Kafka 3.6.1 or newer. If you using an earlier version of Strimzi or APache Kafka, upgrade before migrating to KRaft mode.
```
ppatierno, line 26: Actually this should be 3.7.0.

ppatierno, line 26:
```suggestion
* You must be using Strimzi 0.40 or newer with Kafka 3.6.1 or newer. If you using an earlier version of Strimzi or Apache Kafka, upgrade before migrating to KRaft mode.
```

documentation/modules/deploying/proc-deploy-migrate-kraft.adoc
```diff
@@ -0,0 +1,257 @@
...
+* The Cluster Operator that manages the Kafka cluster is running.
+* The Kafka cluster deployment uses Kafka node pools.
++
+You can xref:proc-migrating-clusters-node-pools-str[migrate your existing Kafka cluster to use node pools]. 
++
+In the `KafkaNodePool` resource configuration, brokers must be contained in a node pool assigned a `broker` role and with the name `kafka`.
```
ppatierno, line 32: "and with the name `kafka`" ... from where did you get this "restriction"? We don't have that, the node pool can have the name you prefer (I have been using "broker" as name for the node pool for example, since long time).

documentation/modules/deploying/proc-deploy-migrate-kraft.adoc
```diff
@@ -0,0 +1,257 @@
...
+----
++
+`METADATA STATE` shows the mechanism used to manage Kafka metadata and coordinate operations.
+At the start of the migration this is `ZooKeeper`.
++
+--
+* `ZooKeeper` is the initial state when metadata is only stored in ZooKeeper.
+* `KRaftMigration` is the state when the migration is in progress.
+* `KRaftDualWriting` is the state when the migration has ended and the cluster is working as a KRaft cluster. 
+Metadata is managed in Kafka but also replicated in ZooKeeper, so you can xref:proc-deploy-migrate-kraft-rollback-{context}[roll back from this point].
```
ppatierno, line 144: Actually, along the way, we decided to not stop the migration at `KRaftDualWriting` as stable state but the next one which is the `KRaftPostMigration`.
It means that you have to expect the brokers rolling twice after the controllers are deployed.
The first time because the are configured to be connected to KRaft as well and the ZooKeeper migration flag is enabled. This way they can register to controllers and the migration starts.
Said that I would highlight that the migration can take long depending on how much topics and corresponding metadata need to be migrated. It means that watching the state could end being in the `KRaftMigration` state for long time (this is the actual state when the migration is happening). Checking that the migration is done or not is executed by the operator on reconciliation bases (so it means by default every 2 mins).
When the migration ends, the state machine moves to `KRaftDualWriting` and it triggers another roll of the brokers to remove the ZooKeeper connection and migration flag. So after rolling, the state moves to `KRaftPostMigration` waiting for the user. It's important to say that even in this state, the dual-write is happening.

documentation/modules/deploying/proc-deploy-migrate-kraft.adoc
```diff
@@ -0,0 +1,257 @@
...
+----
+NAME        ...  METADATA STATE        WARNINGS
+my-cluster  ...  Zookeeper
+my-cluster  ...  KRaftMigration
+my-cluster  ...  KRaftDualWriting
+my-cluster  ...  KRaftPostMigration
+my-cluster  ...  KRaft                 True
+----
++
+`KRaftPostMigration` is the state (after the brokers have rolled) when KRaft mode is enabled and there is no ZooKeeper involvement, though ZooKeeper pods are still running.
```
ppatierno, line 189: As explained before, in this state only brokers are out of ZooKeeper. Controllers are still connected to ZooKeeper and doing the dual-write. Only after applying "enabled", they will be rolled to be out of using ZooKeeper.

ppatierno: "though ZooKeeper pods are still running." ... they are not running anymore. I added the automation about deleting everything.

documentation/modules/deploying/proc-deploy-migrate-kraft.adoc
```diff
@@ -0,0 +1,257 @@
...
+----
++
+For a list of resources created for ZooKeeper, see xref:ref-list-of-kafka-cluster-resources-str[].
+
+[id='proc-deploy-migrate-kraft-rollback-{context}']
+.Performing a rollback on the migration
+
+Before the migration is finalized by enabling KRaft in the 'Kafka' resource, you can perform a rollback operation as follows:
+
+. Apply the `strimzi.io/kraft: rollback` annotation to the `Kafka` resource to roll back the brokers and controllers.
```
ppatierno, line 212: from `KRaftPostMigration` you are actually going to rollback brokers only (not the controllers).

documentation/modules/deploying/proc-deploy-migrate-kraft.adoc
```diff
@@ -0,0 +1,257 @@
...
+kind: Kafka
+metadata:
+  name: my-cluster
+  namespace: my-project
+  annotations:
+    strimzi.io/kraft: rollback 
+# ...
+----
+
+. Delete the controllers node pool:
```
ppatierno, line 232: before this paragraph we should make clear that applying the "rollback" annotation is about rolling the brokers (to be connected to ZooKeeper again) and the state machine comes back to `KRaftDualWriting`.
When in this state, you can delete the controllers.

NOTE: actually from this state, the user could change their mind again and applying the "migration" flag to move again forward with migration.

documentation/modules/deploying/proc-deploy-migrate-kraft.adoc
PaulRMellor: Great. I've dropped this step

documentation/assemblies/upgrading/assembly-upgrade-zookeeper.adoc
PaulRMellor: Understood. Changed as follows:

_If you want to switch a kafka cluster from using a ZooKeeper for metadata management to operating in KRaft mode, the steps must be performed separately from the upgrade. 
For information on migrating to a KRaft-based cluster, see xref:proc-deploy-migrate-kraft-str[]._

ppatierno: This sounds great!

documentation/modules/deploying/proc-deploy-migrate-kraft.adoc
PaulRMellor: This is what we have as a requirement when creating the broker node pool: https://strimzi.io/docs/operators/in-development/deploying#proc-migrating-clusters-node-pools-str

ppatierno: But it's not valid when a user has a Kafka cluster which is using ZooKeeper and was created with node pools seems the beginning.
So I think we have two situation here:
1. a user created a Kafka cluster with ZooKeeper but using node pools since beginning and they want to migrate to KRaft now --> their brokers node pool is named as they want
2. a user created a Kafka cluster with ZooKeeper without node pools. They want to migrate but they need to move to use node pools first --> what you are saying applies (following the link you provided first and then KRaft migration).

documentation/modules/deploying/proc-deploy-migrate-kraft.adoc
PaulRMellor: Thanks. I've updated as directed.

documentation/modules/deploying/proc-deploy-migrate-kraft.adoc
PaulRMellor: I've put the paragraph together and adjusted as follows:

```
You can xref:proc-migrating-clusters-node-pools-str[migrate your existing Kafka cluster to use node pools]. 
To migrate, the `KafkaNodePool` resource configuration, brokers must be contained in a node pool assigned a `broker` role and with the name `kafka`.
Support for node pools is enabled in the `Kafka` resource configuration using the `strimzi.io/node-pools: enabled` annotation.
```

ppatierno: I think the prerequisite here is just you have to have a Kafka cluster ZooKeeper-based using node pools.
If it's already using nodepool for brokers, you are ok, you can start to migrate to KRaft.
If not, you have to migrate to use node pools first (as the link we have) then you can start the KRaft migration.
Can we say something like this? It looks clearer to me.

PaulRMellor: Sure. I've updated to be clearer about this

documentation/assemblies/upgrading/assembly-upgrade-zookeeper.adoc
```diff
@@ -8,6 +8,9 @@
 [role="_abstract"]
 If you are using a ZooKeeper-based Kafka cluster, an upgrade requires an update to the Kafka version and the inter-broker protocol version.
 
+If you want to switch a kafka cluster from using a ZooKeeper for metadata management to operating in KRaft mode, the steps must be performed separately from the upgrade.
```
scholzj, line 11:
```suggestion
If you want to switch a Kafka cluster from using a ZooKeeper for metadata management to operating in KRaft mode, the steps must be performed separately from the upgrade. 
```

documentation/modules/deploying/proc-deploy-migrate-kraft.adoc
```diff
@@ -0,0 +1,251 @@
...
+You enable KRaft migration in the cluster configuration by applying the `strimzi.io/kraft: migration` annotation.  
+After the migration is complete, you switch the brokers to using KRaft and the controllers out of migration mode using the `strimzi.io/kraft: enabled` annotation.
+
+Before starting the migration, verify that your environment can support Kafka in KRaft mode, as there are a number of xref:ref-operator-use-kraft-feature-gate-str[limitations].
+Note also, the following:
+
+* Migration is only supported on dedicated controller nodes, not on nodes with dual roles as brokers and controllers.
+* Throughout the migration process, ZooKeeper and Controller nodes operate in parallel for a period, requiring sufficient compute resources in the cluster.
+
+.Prerequisites
```
scholzj, line 22: Should we also list here that the ZooKeeper based cluster:
* Should use Unidirectional Topic Operator or have Topic Operator disabled
* Should not use JBOd storage as JBOD storage is currently not supported in KRaft

documentation/modules/deploying/proc-deploy-migrate-kraft.adoc
```diff
@@ -0,0 +1,251 @@
...
+kubectl get pods -n my-project
+----
++
+.Output shows nodes in broker and controller node pools
+[source,shell]
+----
+NAME                     READY  STATUS   RESTARTS
+my-cluster-broker-0      1/1    Running  0
+my-cluster-broker-1      1/1    Running  0
+my-cluster-broker-2      1/1    Running  0
```
scholzj, line 112: calling these `my-cluster-kafka` would be better as that is the state where you normally get after migrating to NodePools. So this is what most users will have.

documentation/modules/deploying/proc-deploy-migrate-kraft.adoc
ppatierno: > Should not use JBOd storage as JBOD storage is currently not supported in KRaft

Or maybe clear about having just one volume if the `jbod` type is used in the `spec.storage.type` configuration.

ppatierno:
@PaulRMellor I added the "do-not-merge" label because this one should anyway wait for the implementation in #9480 being merged as well.

documentation/modules/deploying/proc-deploy-migrate-kraft.adoc
PaulRMellor: I thought about that, but I wondered if the name might be confusing and less easy to distinguish broker from controller. I've changed as you confirmed my initial thoughts.  I thought an additional line might help:

_Here, the brokers have the name `my-cluster-kafka` as they were contained in a node pool named `kafka` when migrating the cluster to use node pools._

scholzj (approved):
LGTM. But it shoud be approved by Paolo as the migration SME.

documentation/assemblies/upgrading/assembly-upgrade-zookeeper.adoc
```diff
@@ -8,6 +8,9 @@
 [role="_abstract"]
 If you are using a ZooKeeper-based Kafka cluster, an upgrade requires an update to the Kafka version and the inter-broker protocol version.
 
+If you want to switch a Kafka cluster from using a ZooKeeper for metadata management to operating in KRaft mode, the steps must be performed separately from the upgrade.
```
ppatierno, line 11:
```suggestion
If you want to switch a Kafka cluster from using ZooKeeper for metadata management to operating in KRaft mode, the steps must be performed separately from the upgrade. 
```

documentation/modules/deploying/proc-deploy-migrate-kraft.adoc
```diff
@@ -0,0 +1,256 @@
...
+
+During the migration, you install a quorum of controller nodes as a node pool, which replaces ZooKeeper for management of your cluster. 
+You enable KRaft migration in the cluster configuration by applying the `strimzi.io/kraft: migration` annotation.  
+After the migration is complete, you switch the brokers to using KRaft and the controllers out of migration mode using the `strimzi.io/kraft: enabled` annotation.
+
+Before starting the migration, verify that your environment can support Kafka in KRaft mode, as there are a number of xref:ref-operator-use-kraft-feature-gate-str[limitations].
+Note also, the following:
+
+* Migration is only supported on dedicated controller nodes, not on nodes with dual roles as brokers and controllers.
+* Throughout the migration process, ZooKeeper and Controller nodes operate in parallel for a period, requiring sufficient compute resources in the cluster.
```
ppatierno, line 20: why are you using capital "C" for controller here? It was not for other places. I guess just a typo?

documentation/modules/deploying/proc-deploy-migrate-kraft.adoc
```diff
@@ -0,0 +1,256 @@
...
+
+.Prerequisites
+
+* You must be using Strimzi 0.40 or newer with Kafka 3.7.0 or newer. If you using an earlier version of Strimzi or Apache Kafka, upgrade before migrating to KRaft mode.
```
ppatierno, line 24:
```suggestion
* You must be using Strimzi 0.40 or newer with Kafka 3.7.0 or newer. If you are using an earlier version of Strimzi or Apache Kafka, upgrade before migrating to KRaft mode.
```

documentation/modules/deploying/proc-deploy-migrate-kraft.adoc
```diff
@@ -0,0 +1,256 @@
...
+* Verify that the ZooKeeper-based deployment is operating without the following, as they are not supported in KRaft mode:
+** The Topic Operator running in bidirectional mode. It should either be in unidirectional mode or disabled.
+** JBOD storage. While the `jbod` storage type can be used, the JBOD array must contain only one disk.
+* The Cluster Operator that manages the Kafka cluster is running.
+* The Kafka cluster deployment uses Kafka node pools.
++
+If your ZooKeeper-based cluster is already using node pools, it is ready to migrate.
+If not, you can xref:proc-migrating-clusters-node-pools-str[migrate the cluster to use node pools]. 
+To migrate, brokers must be contained in a `KafkaNodePool` resource configuration that is assigned a `broker` role and has the name `kafka`.
```
ppatierno, line 33: Why? The `kafka` name is true if you are migrating from NO node pools to node pools first (by following the link we provide on the previous line) but if your cluster was created with node pools since the beginning, the broker pools could be also named "foo", "bar", whatever.

documentation/modules/deploying/proc-deploy-migrate-kraft.adoc
```diff
@@ -0,0 +1,256 @@
...
+Support for node pools is enabled in the `Kafka` resource configuration using the `strimzi.io/node-pools: enabled` annotation.
+
+In this procedure, the Kafka cluster name is `my-cluster`, which is located in `my-project`. 
+The name of the controller node pool is `controller`.
+
+.Procedure
+
+. For the Kafka cluster, create a node pool with a `controller` role.
++
+Add a quorum of controller nodes to the node pool.
```
ppatierno, line 43: I don't get this sentence. I mean you are adding controller nodes via/using the node pool. What do you mean by "adding TO the node pool"?

documentation/modules/deploying/proc-deploy-migrate-kraft.adoc
```diff
@@ -0,0 +1,256 @@
...
+my-cluster-kafka-0       1/1    Running  0
+my-cluster-kafka-1       1/1    Running  0
+my-cluster-kafka-2       1/1    Running  0
+my-cluster-controller-3  1/1    Running  0
+my-cluster-controller-4  1/1    Running  0
+my-cluster-controller-5  1/1    Running  0
+# ...
+----
++
+Here, the brokers have the name `my-cluster-kafka` as they were contained in a node pool named `kafka` when migrating the cluster to use node pools.
```
ppatierno, line 122: "when migrating the cluster to use node pools" ... well "if migrating ... " 
See my comment above, it's not always true. People could have cluster using node pools for brokers since the beginning and they didn't have the need to migrate (so the pool had a name different from `kafka`).

documentation/modules/deploying/proc-deploy-migrate-kraft.adoc
```diff
@@ -0,0 +1,256 @@
...
+`METADATA STATE` shows the mechanism used to manage Kafka metadata and coordinate operations.
+At the start of the migration this is `ZooKeeper`.
++
+--
+* `ZooKeeper` is the initial state when metadata is only stored in ZooKeeper.
+* `KRaftMigration` is the state when the migration is in progress and brokers are rolled to register with the controllers.
+The migration can take some time at this point depending on the number of topics and partitions in the cluster. 
+* `KRaftDualWriting` is the state when the Kafka cluster is working as a KRaft cluster, 
+but controllers are managing metadata in Kafka and ZooKeeper. 
+Brokers are rolled a second time to remove the ZooKeeper configuration and migration annotation from the `Kafka` resource.
```
ppatierno, line 150: `and migration annotation from the `Kafka` resource.` we are not removing stuff from the Kafka resource.
Maybe you meant the ZooKeeper migration flag from the broker configuration.

documentation/modules/deploying/proc-deploy-migrate-kraft.adoc
```diff
@@ -0,0 +1,256 @@
...
+* `KRaftPostMigration` is the state when KRaft mode is enabled for brokers and there is no ZooKeeper involvement. 
+Controllers are still connected to ZooKeeper.
```
ppatierno, line 152: and they are still doing "dual write", it's worth saying that.

documentation/modules/deploying/proc-deploy-migrate-kraft.adoc
```diff
@@ -0,0 +1,256 @@
...
++
+.Updates to the metadata state
+[source,shell]
+----
+NAME        ...  METADATA STATE
+my-cluster  ...  Zookeeper
+my-cluster  ...  KRaftMigration
+my-cluster  ...  KRaftDualWriting
+my-cluster  ...  KRaftPostMigration
+my-cluster  ...  KRaft
```
ppatierno, line 195:
```suggestion
my-cluster  ...  PreKRaft             
my-cluster  ...  KRaft             
```

I added the new PreKRaft state which is there since the ZooKeeper deletion was implemented.

documentation/modules/deploying/proc-deploy-migrate-kraft.adoc
```diff
@@ -0,0 +1,256 @@
...
+--
++
+All ZooKeeper-related resources have been automatically deleted.
+
+[id='proc-deploy-migrate-kraft-rollback-{context}']
+.Performing a rollback on the migration
+
+Before the migration is finalized by enabling KRaft in the `Kafka` resource,  and the state has moved to the `KRaft` state, you can perform a rollback operation as follows:
+
+. Apply the `strimzi.io/kraft: rollback` annotation to the `Kafka` resource to roll back the brokers.
```
ppatierno, line 209: I think it's important to mention you have to stay in `KRaftPostMigration` for applying this annotation value.

documentation/modules/deploying/proc-deploy-migrate-kraft.adoc
```diff
@@ -0,0 +1,256 @@
...
+** JBOD storage. While the `jbod` storage type can be used, the JBOD array must contain only one disk.
+* The Cluster Operator that manages the Kafka cluster is running.
+* The Kafka cluster deployment uses Kafka node pools.
++
+If your ZooKeeper-based cluster is already using node pools, it is ready to migrate.
+If not, you can xref:proc-migrating-clusters-node-pools-str[migrate the cluster to use node pools]. 
+To migrate, brokers must be contained in a `KafkaNodePool` resource configuration that is assigned a `broker` role and has the name `kafka`.
+Support for node pools is enabled in the `Kafka` resource configuration using the `strimzi.io/node-pools: enabled` annotation.
+
+In this procedure, the Kafka cluster name is `my-cluster`, which is located in `my-project`.
```
see-quick, line 36:
```suggestion
In this procedure, the Kafka cluster name is `my-cluster`, which is located in `my-project` namespace. 
```

documentation/modules/deploying/proc-deploy-migrate-kraft.adoc
```diff
@@ -0,0 +1,256 @@
...
+
+. Apply the new `KafkaNodePool` resource to create the controllers.
++
+Errors related to using controllers in a ZooKeeper-based environment are expected.
+
+. Enable KRaft migration in the `Kafka` resource by setting the `strimzi.io/kraft` annotation to `migration`:
++
+[source,shell]
+----
+kubectl annotate kafka my-cluster strimzi.io/kraft: migration
```
see-quick, line 84: Is this really correct? Shouldn't be this:
```suggestion
kubectl annotate kafka my-cluster strimzi.io/kraft=migration
```

see-quick (commented):
first iteration adding a few comments, I would also do the second one to check all procedures you have add here.

documentation/modules/deploying/proc-deploy-migrate-kraft.adoc
PaulRMellor: I've added the bold to make it clear:

> If your ZooKeeper-based cluster is already using node pools, it is ready to migrate.
> If not, you can xref:proc-migrating-clusters-node-pools-str[migrate the cluster to use node pools]. 
> To migrate **when the cluster is not using node pools**, brokers must be contained in a `KafkaNodePool` resource configuration that is assigned a `broker` role and has the name `kafka`.
> Support for node pools is enabled in the `Kafka` resource configuration using the `strimzi.io/node-pools: enabled` annotation.

documentation/modules/deploying/proc-deploy-migrate-kraft.adoc
PaulRMellor: I've switched ita round. Hopefully, clearer now.

> The node pool adds a quorum of controller nodes to the cluster.

documentation/modules/deploying/proc-deploy-migrate-kraft.adoc
PaulRMellor: It's not always true and I think we're clear on that early on. But, as Jakub pointed out, we're using the migration from no node pools as that would likely be the starting point. I don't think we need to repeat what we said earlier here.

documentation/modules/deploying/proc-deploy-migrate-kraft.adoc
ppatierno: It's not an annotation. It's a configuration applied/removed by the operator on the broker when you migrate/end migration. You are mentioning ZooKeepeer connection here, it's not an annotation as well. The ZooKeeper migration flag comes along with such connection info.

documentation/modules/deploying/proc-deploy-migrate-kraft.adoc
ppatierno: Tbh I am not sure on which basis we can say users will start from no node pools. Do we have any evidence of that. Can't we have users having ZooKeeper with node pools already? I think so. My opinion is having the documentation not biased on that assumption but the other way around and adding that "if you are not using node pools already, please migrate to use them first".

documentation/modules/deploying/proc-deploy-migrate-kraft.adoc
PaulRMellor: I guess it's this `zookeeper.metadata.migration.enable`, but maybe we don't need to mention that. Is it added at `KRaftMigration`?  And in which state is the ZooKeeper configuration and related resources removed?

documentation/modules/deploying/proc-deploy-migrate-kraft.adoc
PaulRMellor: Either okay?  We use `:` throughout the docs for these annotate commands

documentation/modules/deploying/proc-deploy-migrate-kraft.adoc
ppatierno: Yes that flag exactly. It's added when moving to KRftMigration and removed when moving out of KRaftDualWriting. Anyway I am not saying we should mention it but here we have a wrong sentence about "remove the annotation on the Kafka custom resource" which is wrong. We are not removing any annotation.

PaulRMellor: Thanks. I've mentioned when the flag is added and when its removed. I think it's useful.

documentation/modules/deploying/proc-deploy-migrate-kraft.adoc
```diff
@@ -0,0 +1,259 @@
...
+* The Cluster Operator that manages the Kafka cluster is running.
+* The Kafka cluster deployment uses Kafka node pools.
++
+If your ZooKeeper-based cluster is already using node pools, it is ready to migrate.
+If not, you can xref:proc-migrating-clusters-node-pools-str[migrate the cluster to use node pools]. 
+To migrate when the cluster is not using node pools, brokers must be contained in a `KafkaNodePool` resource configuration that is assigned a `broker` role and has the name `kafka`.
+Support for node pools is enabled in the `Kafka` resource configuration using the `strimzi.io/node-pools: enabled` annotation.
+
+In this procedure, the Kafka cluster name is `my-cluster`, which is located in the `my-project` namespace. 
+The name of the controller node pool is `controller`.
```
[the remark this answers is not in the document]
ppatierno: I noticed you are saying it later, but maybe it's useful to mention it now as well at the beginning?

documentation/modules/deploying/proc-deploy-migrate-kraft.adoc
```diff
@@ -0,0 +1,257 @@
...
+      limits:
+        memory: 64Gi
+        cpu: "12"    
+----
++
+NOTE: For the migration, you cannot use a node pool of nodes that share the broker and controller roles.
+
+. Apply the new `KafkaNodePool` resource to create the controllers.
++
+Errors related to using controllers in a ZooKeeper-based environment are expected.
```
fvaleri, line 79:
```suggestion
Errors related to using controllers in a ZooKeeper-based environment are expected in the cluster operator logs.
```

documentation/modules/deploying/proc-deploy-migrate-kraft.adoc
fvaleri: With `:` the command fails. I would also add quotation marks, as annotation values must be strings.

```sh
kubectl annotate kafka my-cluster strimzi.io/kraft="migration"
```

documentation/modules/deploying/proc-deploy-migrate-kraft.adoc
```diff
@@ -0,0 +1,257 @@
...
+kind: Kafka
+metadata:
+  name: my-cluster
+  namespace: my-project
+  annotations:
+    strimzi.io/kraft: enabled 
+# ...
+----
++
+WARNING: Rollback cannot be performed after enabling KRaft.
```
fvaleri, line 178: Should we put this warning before the command to enable?

fvaleri (commented):
Thanks. Left some comments.

documentation/modules/deploying/proc-deploy-migrate-kraft.adoc
```diff
@@ -0,0 +1,257 @@
...
+[source,shell]
+----
+kubectl delete KafkaNodePool controller -n my-project
+----
+
+. Apply the `strimzi.io/kraft: disabled` annotation to the `Kafka` resource to return the metadata state to `ZooKeeper`.
++
+[source,shell]
+----
+kubectl annotate kafka my-cluster strimzi.io/kraft: disabled
```
fvaleri, line 243:
```suggestion
kubectl annotate kafka my-cluster strimzi.io/kraft="disabled" --overwrite
```

documentation/modules/deploying/proc-deploy-migrate-kraft.adoc
```diff
@@ -0,0 +1,257 @@
...
+[id='proc-deploy-migrate-kraft-rollback-{context}']
+.Performing a rollback on the migration
+
+Before the migration is finalized by enabling KRaft in the `Kafka` resource,  and the state has moved to the `KRaft` state, you can perform a rollback operation as follows:
+
+. Apply the `strimzi.io/kraft: rollback` annotation to the `Kafka` resource to roll back the brokers.
++
+[source,shell]
+----
+kubectl annotate kafka my-cluster strimzi.io/kraft: rollback
```
fvaleri, line 213:
```suggestion
kubectl annotate kafka my-cluster strimzi.io/kraft="rollback" --overwrite
```

documentation/modules/deploying/proc-deploy-migrate-kraft.adoc
```diff
@@ -0,0 +1,257 @@
...
+You can xref:proc-deploy-migrate-kraft-rollback-{context}[roll back from this point].
+--
++
+The migration status is also represented in the `status.kafkaMetadataState` property of the `Kafka` resource. 
+
+. Enable KRaft in the `Kafka` resource configuration by setting the `strimzi.io/kraft` annotation to `enabled`:
++
+[source,shell]
+----
+kubectl annotate kafka my-cluster strimzi.io/kraft: enabled
```
fvaleri, line 162:
```suggestion
kubectl annotate kafka my-cluster strimzi.io/kraft="enabled" --overwrite
```

documentation/modules/deploying/proc-deploy-migrate-kraft.adoc
ppatierno: Yeah good catch @fvaleri , in all commands we would need the `--overwrite` because the annotation could be already present on the `Kafka` CR as `disabled` (without overwrite it's not going to work).

documentation/modules/deploying/proc-deploy-migrate-kraft.adoc
ppatierno: I have never tried wit `:` and @fvaleri is right, the command doesn't work on my side either.
Only thing the quotes are not needed around the annotation value but anyway the command works with them as well.

documentation/modules/deploying/proc-deploy-migrate-kraft.adoc
```diff
@@ -0,0 +1,257 @@
...
+but controllers are managing metadata in Kafka and ZooKeeper. 
+Brokers are rolled a second time to remove the flag to enable migration.
+* `KRaftPostMigration` is the state when KRaft mode is enabled for brokers and there is no ZooKeeper involvement. 
+Controllers are still managing in Kafka and ZooKeeper.
+You can xref:proc-deploy-migrate-kraft-rollback-{context}[roll back from this point].
+--
++
+The migration status is also represented in the `status.kafkaMetadataState` property of the `Kafka` resource. 
+
+. Enable KRaft in the `Kafka` resource configuration by setting the `strimzi.io/kraft` annotation to `enabled`:
```
fvaleri, line 158: It may be worth noticing that ZooKeepoer PV and PVC may not be deleted, depending on the `.spec.zookeeper.storage` configuration.

ppatierno: Yeah, let's mention that it will happen or not accordily with the `deleteClaim` flag (true=PVC deleted automatically, flast=PVC not deleted)

ppatierno:
@PaulRMellor while having an offline chat with @fvaleri it could be worth mentioning that if your initial configuration has the inter broker protocol and/or log message format set, you should remove them at the end of the migration. The brokers and controllers will be rolled one more time to update them.
At the same time even when removing the `spec.zookeeper`.
There are warnings in the `Kafka` CR but after the rolling, those are removed.

documentation/modules/deploying/proc-deploy-migrate-kraft.adoc
see-quick: without `"` is also working :)

documentation/modules/deploying/proc-deploy-migrate-kraft.adoc
see-quick: +1

documentation/modules/deploying/proc-deploy-migrate-kraft.adoc
```diff
@@ -0,0 +1,257 @@
...
+The migration can take some time at this point depending on the number of topics and partitions in the cluster. 
+* `KRaftDualWriting` is the state when the Kafka cluster is working as a KRaft cluster, 
+but controllers are managing metadata in Kafka and ZooKeeper. 
+Brokers are rolled a second time to remove the flag to enable migration.
+* `KRaftPostMigration` is the state when KRaft mode is enabled for brokers and there is no ZooKeeper involvement. 
+Controllers are still managing in Kafka and ZooKeeper.
+You can xref:proc-deploy-migrate-kraft-rollback-{context}[roll back from this point].
+--
++
+The migration status is also represented in the `status.kafkaMetadataState` property of the `Kafka` resource.
```
see-quick, line 156: Hmmm I would also here write something that from this standpoint the user could decide if he wants to:
1. go back to ZooKeeper (i.e., use `strimzi.io/kraft: rollback` mechanism) 
2. or continue with migration to KRaft

see-quick (approved):
Just a few comments from me 👍  Looks good

documentation/modules/deploying/proc-deploy-migrate-kraft.adoc
```diff
@@ -0,0 +1,257 @@
...
+* The Kafka cluster deployment uses Kafka node pools.
++
+If your ZooKeeper-based cluster is already using node pools, it is ready to migrate.
+If not, you can xref:proc-migrating-clusters-node-pools-str[migrate the cluster to use node pools]. 
+To migrate when the cluster is not using node pools, brokers must be contained in a `KafkaNodePool` resource configuration that is assigned a `broker` role and has the name `kafka`.
+Support for node pools is enabled in the `Kafka` resource configuration using the `strimzi.io/node-pools: enabled` annotation.
+
+In this procedure, the Kafka cluster name is `my-cluster`, which is located in the `my-project` namespace. 
+The name of the controller node pool created is `controller`.
+The node pool for the brokers is called `kafka`.
```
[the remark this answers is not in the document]
ppatierno: The broker node pool naming was part of along discussion. I don't love that much either but it's coming from the compromise between using a name for people who are coming from node pools migration (and the broker pool will be `kafka` for sure) and the others who already have node pools from the beginning.
I can understand that the annotate command could be confusing but I would expect that users running that command know what `kafka` means there (the CRD not the pool name).

documentation/modules/deploying/proc-deploy-migrate-kraft.adoc
PaulRMellor: Thanks. Actually, we only used `:` in a  couple of places (which I must have referenced). I've gone through the annotations and made consistent. @see-quick -- if you could please check commit 
[docs(review): annotations values update](https://github.com/strimzi/strimzi-kafka-operator/pull/9633/commits/1f44abcd630c1eff0b2e2632410317255243f3e5)

ppatierno:
@PaulRMellor We should take all the updates not related to migration out of this PR but having a new one for them. I see you updated annotation values in all other places, but I would move them for a new PR. Let's take this PR for migration only.

ppatierno (changes requested):
@PaulRMellor We should take all the updates not related to migration out of this PR but having a new one for them. I see you updated annotation values in all other places, but I would move them for a new PR. Let's take this PR for migration only.

documentation/modules/deploying/proc-deploy-migrate-kraft.adoc
PaulRMellor: Thanks. I don't think we need to mention the specific error

documentation/modules/deploying/proc-deploy-migrate-kraft.adoc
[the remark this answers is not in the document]
PaulRMellor: Thanks @see-quick. As Paolo mentions, there was a long discussion on this. The first draft didn't use `kafka`. From Jakub's perspective, most likely case is that the user will be migrating to using node pools as well, so we went with this. I think the explanations we give make this clear.

documentation/modules/deploying/proc-deploy-migrate-kraft.adoc
PaulRMellor: Update: changes outside this migration will be in another PR

ppatierno (approved):
LGTM. Back to green!

fvaleri (approved):
LGTM. Thanks.

ppatierno:
> @PaulRMellor while having an offline chat with @fvaleri it could be worth mentioning that if your initial configuration has the inter broker protocol and/or log message format set, you should remove them at the end of the migration. The brokers and controllers will be rolled one more time to update them.
At the same time even when removing the spec.zookeeper.
There are warnings in the Kafka CR but after the rolling, those are removed.

@PaulRMellor I don't see addressed the above comment here.

documentation/modules/deploying/proc-deploy-migrate-kraft.adoc
```diff
@@ -0,0 +1,272 @@
...
+`METADATA STATE` shows the mechanism used to manage Kafka metadata and coordinate operations.
+At the start of the migration this is `ZooKeeper`.
++
+--
+* `ZooKeeper` is the initial state when metadata is only stored in ZooKeeper.
+* `KRaftMigration` is the state when the migration is in progress.
+The flag to enable ZooKeeper to KRaft migration (`zookeeper.metadata.migration.enable`) is added to the brokers and they are rolled to register with the controllers.
+The migration can take some time at this point depending on the number of topics and partitions in the cluster. 
+* `KRaftDualWriting` is the state when the Kafka cluster is working as a KRaft cluster, 
+but controllers are managing metadata in Kafka and ZooKeeper.
```
scholzj, line 149:
```suggestion
but metadata are being stored in both Kafka and ZooKeeper. 
```

documentation/modules/deploying/proc-deploy-migrate-kraft.adoc
```diff
@@ -0,0 +1,272 @@
...
+--
+* `ZooKeeper` is the initial state when metadata is only stored in ZooKeeper.
+* `KRaftMigration` is the state when the migration is in progress.
+The flag to enable ZooKeeper to KRaft migration (`zookeeper.metadata.migration.enable`) is added to the brokers and they are rolled to register with the controllers.
+The migration can take some time at this point depending on the number of topics and partitions in the cluster. 
+* `KRaftDualWriting` is the state when the Kafka cluster is working as a KRaft cluster, 
+but controllers are managing metadata in Kafka and ZooKeeper. 
+Brokers are rolled a second time to remove the flag to enable migration.
+* `KRaftPostMigration` is the state when KRaft mode is enabled for brokers and there is no ZooKeeper involvement. 
+Controllers are still managing in Kafka and ZooKeeper.
```
scholzj, line 152: Similarly to the comment above. I think `Controllers are still managing in Kafka and ZooKeeper.` makes no sense.

```suggestion
Metadata are being stored in both Kafka and ZooKeeper. 
```

documentation/modules/deploying/proc-deploy-migrate-kraft.adoc
```diff
@@ -0,0 +1,272 @@
...
++
+--
+* `ZooKeeper` is the initial state when metadata is only stored in ZooKeeper.
+* `KRaftMigration` is the state when the migration is in progress.
+The flag to enable ZooKeeper to KRaft migration (`zookeeper.metadata.migration.enable`) is added to the brokers and they are rolled to register with the controllers.
+The migration can take some time at this point depending on the number of topics and partitions in the cluster. 
+* `KRaftDualWriting` is the state when the Kafka cluster is working as a KRaft cluster, 
+but controllers are managing metadata in Kafka and ZooKeeper. 
+Brokers are rolled a second time to remove the flag to enable migration.
+* `KRaftPostMigration` is the state when KRaft mode is enabled for brokers and there is no ZooKeeper involvement.
```
scholzj, line 151: `and there is no ZooKeeper involvement` seems very strange when the next sentence says that it is still used.

PaulRMellor: Yeah. I don't think we lose anything by dropping that.

PR #9636, opened 2024-02-03
[ST] Fixups to few failing tests

im-konge:
### Type of change

- Bugfix

### Description

This PR fixes several failing (or flaky) tests:

- PodSecurityProfilesST.testOperandsWithRestrictedSecurityProfile - issue that the Namespace was not every time correctly labeled - now the Namespace will be labeled in a loop and in case of exception, it just tries again (credit @kornys )

- ConnectST.testConnectorTaskAutoRestart - test which was disabled before, but now "somehow" works - it actually fails from time to time on the infinite restart of EchoSink Connector - there was issue with how we were sending the messages. In case that we send all messages, that are set as a failed count, the exception is thrown and the Connector is restarted. But because the exception is in `put()` method of the `EchoSinkTask`, there is a possibility that the offset of the records is not committed. That causes the "infinite auto-restart".

-  MultipleListenersST.testMultipleRoutes - from time to time the generated name of the Kafka cluster with name of particular NodePool hits the limit of 63 chars, causing issues with Routes. The idea is to change the prefix of the NodePool, so it will be shorter and we will not hit the limit.

- ConnectBuilderST.testUpdateConnectWithAnotherPlugin - issue with `null` plugins in the status. I added a method for waiting until Connect status contains the list of plugins

Also, there are many tests failing because of `assertNoCoErrorsLogged` check, which finds some errors/warnings in CO log - some of the things are just warnings that we should fix in our STs, some of the failures are just wrongly matched to the whitelisted pattern, some of the errors needs fix or addition to whitelist. For now, I disabled this check and created https://github.com/strimzi/strimzi-kafka-operator/issues/9648 to track the fix and enablement.

### Checklist

### DIFF
--- a/systemtest/src/main/java/io/strimzi/systemtest/TestConstants.java
+++ b/systemtest/src/main/java/io/strimzi/systemtest/TestConstants.java
@@ -500,7 +500,7 @@ public interface TestConstants {
     /**
      * KafkaNodePools constants
      */
-    String KAFKA_NODE_POOL_PREFIX = "kafka-pool-";
+    String KAFKA_NODE_POOL_PREFIX = "pool-";
 
     /**
      * Persistent Volume related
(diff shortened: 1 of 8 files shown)


see-quick (approved):
LGTM, and thanks for it! 👍

henryZrncik (approved):

jankalinic (approved):

PR #9637, opened 2024-02-04
[ST] Refactor logs collecting of systemtests and collect more data

kornys:
### Type of change

- Refactoring

### Description

1. Clean LogCollector class
2. Do not call describe nodes for every namespace in case of testFail (it reduces time significantly)
3. Do not create path by string comprehension but use proper Path.resolve() method
4. Use Files.writeString for writing files
5. Collect also configmaps and secrets

### Checklist


im-konge (approved):
LGTM, thanks

see-quick (approved):
Thanks

PR #9638, opened 2024-02-05
[ST] Fix testUpdateConnectWithAnotherPlugin race condition and add wa…

see-quick:
…it until

### Type of change

- Bugfix
- Enhancement / new feature
- Refactoring

### Description

This PR fixes the race condition inside the `testUpdateConnectWithAnotherPlugin` test. I have also added a generic method to cover and fix such problem

### Checklist


im-konge (changes requested):
The fix you are adding here is done in #9636

see-quick:
🤦 oh gosh...I must have overlooked that during that review closing

PR #9649, opened 2024-02-06
ST: Add missing escape characters for log errors whitelist

Frawless:
### Type of change

- Bugfix

### Description

There were missing escape characters for two of the whitelisted errors that caused unexpected failures due to not-matching the whitelisted error. This PR should solve it.

### Checklist

- [ ] Make sure all tests pass

PR #9656, opened 2024-02-07
[ST] Re-enable applying of NetworkPolicies during namespace creation

im-konge:
### Type of change

- Bugfix

### Description

During work on `NamespaceManager` I found out that the NetworkPolicies, that were part of the Namespace creation, are not created at all -> the `NetworkPolicyTemplates.applyDefaultNetworkPolicy()` method, even though named as "apply", just returned the NetworkPolicy object, which was then never actually created (only in `OauthAbstractST`, where it's done by `ResourceManager` from the beginning).

This PR fixes this issue. It adds the correct deployment of NetworkPolicies to `NamespaceManager`, so it is on one place and removes it from anywhere else. Also, it removes old and not used methods after changes, that were made in last PRs.

### Checklist


Frawless (approved):

jankalinic (approved):
LGTM

see-quick (approved):
LGTM, thanks 👍

im-konge:
Test failures in TF are not related to my PR, they are being fixed in #9649 and then in another PR

PR #9657, opened 2024-02-07
Additional authentication type 'serviceaccount-oauth' that extends 'oauth'

mstruk:
### Type of change

- Enhancement / new feature

### Description

Make it easy to configure the usage of Kubernetes internal OIDC server. The new authentication type 'serviceaccount-oauth' is an extension of 'oauth' type for listeners and various managed Kafka clients (KafkaConnector, KafkaBridge, KafkaMirrorMaker2, ...).

### Checklist

- [X] Write tests
- [X] Update documentation
- [X] Update CHANGELOG.md

### DIFF
--- a/api/src/main/java/io/strimzi/api/kafka/model/common/authentication/KafkaClientAuthentication.java
+++ b/api/src/main/java/io/strimzi/api/kafka/model/common/authentication/KafkaClientAuthentication.java
@@ -27,6 +27,7 @@ import java.util.Map;
     @JsonSubTypes.Type(name = KafkaClientAuthenticationScramSha512.TYPE_SCRAM_SHA_512, value = KafkaClientAuthenticationScramSha512.class),
     @JsonSubTypes.Type(name = KafkaClientAuthenticationPlain.TYPE_PLAIN, value = KafkaClientAuthenticationPlain.class),
     @JsonSubTypes.Type(name = KafkaClientAuthenticationOAuth.TYPE_OAUTH, value = KafkaClientAuthenticationOAuth.class),
+    @JsonSubTypes.Type(name = KafkaClientAuthenticationK8sOIDC.TYPE_K8S_OIDC, value = KafkaClientAuthenticationK8sOIDC.class),
 })
 @JsonInclude(JsonInclude.Include.NON_NULL)
 @EqualsAndHashCode
@@ -35,10 +36,11 @@ public abstract class KafkaClientAuthentication implements UnknownPropertyPreser
     private Map<String, Object> additionalProperties;
 
     @Description("Authentication type. " +
-            "Currently the supported types are `tls`, `scram-sha-256`, `scram-sha-512`, `plain`, and 'oauth'. " +
+            "Currently the supported types are `tls`, `scram-sha-256`, `scram-sha-512`, `plain`, 'oauth', and 'k8s-oidc'. " +
             "`scram-sha-256` and `scram-sha-512` types use SASL SCRAM-SHA-256 and SASL SCRAM-SHA-512 Authentication, respectively. " +
             "`plain` type uses SASL PLAIN Authentication. " +
             "`oauth` type uses SASL OAUTHBEARER Authentication. " +
+            "`k8s-oidc` type uses SASL OAUTHBEARER Authentication. " +
             "The `tls` type uses TLS Client Authentication. " +
             "The `tls` type is supported only over TLS connections.")
     public abstract String getType();
--- a/api/src/main/java/io/strimzi/api/kafka/model/common/authentication/KafkaClientAuthentication.java
+++ b/api/src/main/java/io/strimzi/api/kafka/model/common/authentication/KafkaClientAuthentication.java
@@ -40,7 +40,7 @@ public abstract class KafkaClientAuthentication implements UnknownPropertyPreser
             "`scram-sha-256` and `scram-sha-512` types use SASL SCRAM-SHA-256 and SASL SCRAM-SHA-512 Authentication, respectively. " +
             "`plain` type uses SASL PLAIN Authentication. " +
             "`oauth` type uses SASL OAUTHBEARER Authentication. " +
-            "`serviceaccount-oauth` type uses SASL OAUTHBEARER Authentication. " +
+            "`serviceaccount-oauth` type uses SASL OAUTHBEARER Authentication uses SASL OAUTHBEARER Authentication pre-configured with Kubernetes ServiceAccount token file location. " +
             "The `tls` type uses TLS Client Authentication. " +
             "The `tls` type is supported only over TLS connections.")
     public abstract String getType();
--- a/api/src/main/java/io/strimzi/api/kafka/model/common/authentication/KafkaClientAuthentication.java
+++ b/api/src/main/java/io/strimzi/api/kafka/model/common/authentication/KafkaClientAuthentication.java
@@ -27,7 +27,7 @@ import java.util.Map;
     @JsonSubTypes.Type(name = KafkaClientAuthenticationScramSha512.TYPE_SCRAM_SHA_512, value = KafkaClientAuthenticationScramSha512.class),
     @JsonSubTypes.Type(name = KafkaClientAuthenticationPlain.TYPE_PLAIN, value = KafkaClientAuthenticationPlain.class),
     @JsonSubTypes.Type(name = KafkaClientAuthenticationOAuth.TYPE_OAUTH, value = KafkaClientAuthenticationOAuth.class),
-    @JsonSubTypes.Type(name = KafkaClientAuthenticationK8sOIDC.TYPE_K8S_OIDC, value = KafkaClientAuthenticationK8sOIDC.class),
+    @JsonSubTypes.Type(name = KafkaClientAuthenticationServiceAccountOAuth.TYPE_SERVICEACCOUNT_OAUTH, value = KafkaClientAuthenticationServiceAccountOAuth.class),
 })
 @JsonInclude(JsonInclude.Include.NON_NULL)
 @EqualsAndHashCode
@@ -36,11 +36,11 @@ public abstract class KafkaClientAuthentication implements UnknownPropertyPreser
     private Map<String, Object> additionalProperties;
 
     @Description("Authentication type. " +
-            "Currently the supported types are `tls`, `scram-sha-256`, `scram-sha-512`, `plain`, 'oauth', and 'k8s-oidc'. " +
+            "Currently the supported types are `tls`, `scram-sha-256`, `scram-sha-512`, `plain`, 'oauth', and 'serviceaccount-oauth'. " +
             "`scram-sha-256` and `scram-sha-512` types use SASL SCRAM-SHA-256 and SASL SCRAM-SHA-512 Authentication, respectively. " +
             "`plain` type uses SASL PLAIN Authentication. " +
             "`oauth` type uses SASL OAUTHBEARER Authentication. " +
-            "`k8s-oidc` type uses SASL OAUTHBEARER Authentication. " +
+            "`serviceaccount-oauth` type uses SASL OAUTHBEARER Authentication. " +
             "The `tls` type uses TLS Client Authentication. " +
             "The `tls` type is supported only over TLS connections.")
     public abstract String getType();
--- a/api/src/main/java/io/strimzi/api/kafka/model/common/authentication/KafkaClientAuthenticationOAuth.java
+++ b/api/src/main/java/io/strimzi/api/kafka/model/common/authentication/KafkaClientAuthenticationOAuth.java
@@ -48,6 +48,7 @@ public class KafkaClientAuthenticationOAuth extends KafkaClientAuthentication {
     private GenericSecretSource clientSecret;
     private PasswordSecretSource passwordSecret;
     private GenericSecretSource accessToken;
+    private String accessTokenLocation;
     private GenericSecretSource refreshToken;
     private List<CertSecretSource> tlsTrustedCertificates;
     private boolean disableTlsHostnameVerification = false;
@@ -164,6 +165,16 @@ public class KafkaClientAuthenticationOAuth extends KafkaClientAuthentication {
 
     public void setAccessToken(GenericSecretSource accessToken) {
         this.accessToken = accessToken;
+    }
+
+    @Description("Path to the token file containing an access token to be used for authentication.")
+    @JsonInclude(JsonInclude.Include.NON_NULL)
+    public String getAccessTokenLocation() {
+        return accessTokenLocation;
+    }
+
+    public void setAccessTokenLocation(String path) {
+        this.accessTokenLocation = path;
     }
 
     @Description("Link to Kubernetes Secret containing the refresh token which can be used to obtain access token from the authorization server.")
--- a/api/src/main/java/io/strimzi/api/kafka/model/common/authentication/KafkaClientAuthenticationOAuth.java
+++ b/api/src/main/java/io/strimzi/api/kafka/model/common/authentication/KafkaClientAuthenticationOAuth.java
@@ -27,10 +27,11 @@ import java.util.List;
         builderPackage = Constants.FABRIC8_KUBERNETES_API
 )
 @JsonInclude(JsonInclude.Include.NON_NULL)
+// Make sure this is in sync with KafkaClientAuthenticationK8sOIDC
 @JsonPropertyOrder({"type", "clientId", "username", "scope", "audience", "tokenEndpointUri", "connectTimeoutSeconds",
     "readTimeoutSeconds", "httpRetries", "httpRetryPauseMs", "clientSecret", "passwordSecret", "accessToken",
     "refreshToken", "tlsTrustedCertificates", "disableTlsHostnameVerification", "maxTokenExpirySeconds",
-    "accessTokenIsJwt", "enableMetrics", "includeAcceptHeader"})
+    "accessTokenIsJwt", "enableMetrics", "includeAcceptHeader", "accessTokenLocation"})
 @EqualsAndHashCode(callSuper = true)
 @ToString(callSuper = true)
 public class KafkaClientAuthenticationOAuth extends KafkaClientAuthentication {
--- a/api/src/main/java/io/strimzi/api/kafka/model/common/authentication/KafkaClientAuthenticationOAuth.java
+++ b/api/src/main/java/io/strimzi/api/kafka/model/common/authentication/KafkaClientAuthenticationOAuth.java
@@ -27,7 +27,7 @@ import java.util.List;
         builderPackage = Constants.FABRIC8_KUBERNETES_API
 )
 @JsonInclude(JsonInclude.Include.NON_NULL)
-// Make sure this is in sync with KafkaClientAuthenticationK8sOIDC
+// Make sure this is in sync with KafkaClientAuthenticationServiceAccountOAuth
 @JsonPropertyOrder({"type", "clientId", "username", "scope", "audience", "tokenEndpointUri", "connectTimeoutSeconds",
     "readTimeoutSeconds", "httpRetries", "httpRetryPauseMs", "clientSecret", "passwordSecret", "accessToken",
     "refreshToken", "tlsTrustedCertificates", "disableTlsHostnameVerification", "maxTokenExpirySeconds",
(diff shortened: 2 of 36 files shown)


api/src/main/java/io/strimzi/api/kafka/model/common/authentication/KafkaClientAuthentication.java
```diff
@@ -27,6 +27,7 @@
     @JsonSubTypes.Type(name = KafkaClientAuthenticationScramSha512.TYPE_SCRAM_SHA_512, value = KafkaClientAuthenticationScramSha512.class),
     @JsonSubTypes.Type(name = KafkaClientAuthenticationPlain.TYPE_PLAIN, value = KafkaClientAuthenticationPlain.class),
     @JsonSubTypes.Type(name = KafkaClientAuthenticationOAuth.TYPE_OAUTH, value = KafkaClientAuthenticationOAuth.class),
+    @JsonSubTypes.Type(name = KafkaClientAuthenticationK8sOIDC.TYPE_K8S_OIDC, value = KafkaClientAuthenticationK8sOIDC.class),
```
scholzj, line 30: Should we call it `kubernetes`? Or `serviceaccount`?

scholzj: I wonder if we should have a proposal for this to clarify details like this. @strimzi/maintainers Any thoughts on the type name?

Frawless: How about `kubernetes-internal` / `KUBERNETES_INTERNAL` / `K8S_INTERNAL` ? From the discussion during community call it would make sense to me.

see-quick: What about `TYPE_KUBERNETES_OPEN_ID_CONNECT_AUTH`?

PaulRMellor: It doesn't seem too bad to me. Alternatively, just `OIDC` or `KubernetesOIDC`

scholzj: FYI: It is not about the Java variable ... it is about the type `type: k8s-oidc`

im-konge: Few suggestions in case that `k8s` can be the main source for typo:
- `kube-oidc` 
- `kubernetes-oidc` 

Or both abreviations:
- `kube-open-id`
- `kubernetes-open-id`
 
We have also different types like `scram-sha-512`, which are longer and contains dashes.

Also I think that the type name should point directly to the auth, so it will not be confusing for the users, what is really behind that.

ppatierno: I am for the first two suggestions by @im-konge so `kube-iodc` or `kubernetes-iodc`

mstruk: All these are fine with me. I guess the 'k8s' part is the 'too hackish' part so `kube-oidc` is great. I prefer the shorter names, so my second choice then would be `kube-open-id`.

scholzj: > I am for the first two suggestions by @im-konge so `kube-iodc` or `kubernetes-iodc`

`iodc`? That shows the problem I have with it :-D

> I guess the 'k8s' part is the 'too hackish' part so kube-oidc is great.

Is `kube` really less hackish than `k8s`? I guess `kubernetes` would make more sense if that is the case.

mstruk: What about `kubernetes-openid` (slightly better than `kubernetes-open-id`).

mstruk: Or maybe `kubernetes-oauth`, but people may again easily typo it as `kubernetes-auth`.

scholzj: @mstruk What is the expectation from the user? Do you have some example of how will it be used in the client applications? Will the user be really aware of the Oauth / OIDC stuff underneath? Or will it also just point to the service account token? I think it is improtant to make it understandabe to the users what it does. If they have to configure some OAuth / OIDC stuff, than they are aware of it and its fine. But if they just create a service account, it is probably better to make it more obvious from the name?

mstruk: The idea is that on the client you simply use the access token provided by Kubernetes in the standard place. You can use standard Kubernetes ways to configure service account for your Pod. The `authentication` configuration for `KafkaMirrorMaker2` or `KafkaConnect` CRs would be as simple as `type: kubernetes-openid`.
But the mechanism is an extension of `oauth` so you can use any `oauth` config option to override the default values, if needed.

scholzj: So, the appeal to the users is that they should not be worried about OAuth stuff, right? They just create and use a service account. So should that be obvious from the name?

mstruk: That's may actually not be a good idea because searching for it through documentation or search engines will not produce good results.

Frawless: So based on the community call from this week let's move this a little bit. In the comments there were the following options more or less:

- `kubernetes-openid`
- `kubernetes-auth`
- `kubernetes`
- `kubernetes-oidc`
- `kubernetes-internal`
- `kubernetes-sa` maybe (not listed above)?

Any other options?

scholzj: `serviceaccount` or some variant on it? `serviceacount-oidc` or `serviceaccount-oauth`?

Frawless: ye I was thinking about `serviceaccount` as well, but alone it sounds little bit strange to me. `serviceacount-oidc` or `serviceacount-oauth` sounds better I think. I will put it into the list and I guess we could do some vote soon.

scholzj: Discussed on the community call of 16.5.2024: We decided to go with `type: serviceaccount-oauth`.

api/src/main/java/io/strimzi/api/kafka/model/common/authentication/KafkaClientAuthentication.java
```diff
@@ -35,10 +36,11 @@ public abstract class KafkaClientAuthentication implements UnknownPropertyPreser
     private Map<String, Object> additionalProperties;
 
     @Description("Authentication type. " +
-            "Currently the supported types are `tls`, `scram-sha-256`, `scram-sha-512`, `plain`, and 'oauth'. " +
+            "Currently the supported types are `tls`, `scram-sha-256`, `scram-sha-512`, `plain`, 'oauth', and 'serviceaccount-oauth'. " +
             "`scram-sha-256` and `scram-sha-512` types use SASL SCRAM-SHA-256 and SASL SCRAM-SHA-512 Authentication, respectively. " +
             "`plain` type uses SASL PLAIN Authentication. " +
             "`oauth` type uses SASL OAUTHBEARER Authentication. " +
+            "`serviceaccount-oauth` type uses SASL OAUTHBEARER Authentication. " +
```
scholzj, line 43: Maybe we could add something like `uses SASL OAUTHBEARER Authentication in combination with Kubernetes ServiceAccounts`?

api/src/main/java/io/strimzi/api/kafka/model/common/authentication/KafkaClientAuthenticationOAuth.java
```diff
@@ -166,6 +168,16 @@ public void setAccessToken(GenericSecretSource accessToken) {
         this.accessToken = accessToken;
     }
 
+    @Description("Path to the token file containing an access token to be used for authentication.")
```
scholzj, line 171: Is this required when using `type: serviceaccount-oauth`?

api/src/main/java/io/strimzi/api/kafka/model/kafka/listener/KafkaListenerAuthentication.java
```diff
@@ -41,6 +42,7 @@ public abstract class KafkaListenerAuthentication implements UnknownPropertyPres
 
     @Description("Authentication type. " +
             "`oauth` type uses SASL OAUTHBEARER Authentication. " +
+            "`serviceaccount-oauth` type uses SASL OAUTHBEARER Authentication. " +
```
scholzj, line 45: Same as earlier -> maybe you can explain how it differs from `oauth`.

api/src/main/java/io/strimzi/api/kafka/model/kafka/listener/KafkaListenerAuthenticationOAuth.java
```diff
@@ -285,6 +287,16 @@ public void setIntrospectionEndpointUri(String introspectionEndpointUri) {
         this.introspectionEndpointUri = introspectionEndpointUri;
     }
 
+    @Description("Path to the file on the local filesystem that contains a bearer token to be used instead of client_id and secret when authenticating to authorization server")
```
scholzj, line 290: Same as earlier -> is this required for `type: serviceaccount-oauth`? If not, we should maybe add the default. Otherwise say that it has to be used in the description.

cluster-operator/src/test/java/io/strimzi/operator/cluster/model/KafkaMirrorMaker2ClusterTest.java
```diff
@@ -1635,6 +1636,31 @@ public void testPodSetWithOAuthWithAccessToken() {
         });
     }
 
+    @ParallelTest
+    public void testPodSetWithServiceAccountOAuth() {
```
scholzj, line 1640: Should this test also when it is used for the source clusters?

scholzj (commented):
One more thing - this should also have some CHANGELOG record.

api/src/main/java/io/strimzi/api/kafka/model/common/authentication/KafkaClientAuthentication.java
mstruk: Yes, sounds good.

api/src/main/java/io/strimzi/api/kafka/model/common/authentication/KafkaClientAuthenticationOAuth.java
mstruk: It is an underlying capability which may be used with `oauth`. When using `serviceaccount-oauth` this value is automatically configured by the operator to the standard path of the serviceaccount token file.

api/src/main/java/io/strimzi/api/kafka/model/kafka/listener/KafkaListenerAuthentication.java
mstruk: Your proposed string would work here as well. In order to be slightly more descriptive maybe we could say something like: 'uses SASL OAUTHBEARER Authentication pre-configured with Kubernetes ServiceAccount token file location'

cluster-operator/src/test/java/io/strimzi/operator/cluster/model/KafkaMirrorMaker2ClusterTest.java
mstruk: I added the test by pattern-matching other OAuth tests, and they only test with target cluster. I guess the reasoning here was that we really test `KafkaClientAuthenticationOAuthBuilder` here, which is used both for the target and for the source cluster, thus if it works for one why would it fail for the other?

cluster-operator/src/test/java/io/strimzi/operator/cluster/model/KafkaMirrorMakerClusterTest.java
```diff
@@ -1062,6 +1063,27 @@ public void testGenerateDeploymentWithConsumerOAuthWithAccessToken() {
         assertThat(cont.getEnv().stream().filter(var -> KafkaMirrorMakerCluster.ENV_VAR_KAFKA_MIRRORMAKER_OAUTH_CONFIG_CONSUMER.equals(var.getName())).findFirst().orElseThrow().getValue().isEmpty(), is(true));
     }
 
+    @ParallelTest
+    public void testGenerateDeploymentWithConsumerServiceAccountOAuth() {
```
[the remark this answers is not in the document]
mstruk: That's a good catch. There's is a whole series of tests with consumer and with producer there. I missed the `producer` one.

cluster-operator/src/test/java/io/strimzi/operator/cluster/model/KafkaMirrorMaker2ClusterTest.java
scholzj: The env vars work differently for source clusters as there might be many of them. That is why I think it needs to be tested as well. It is possible it was forgotten for the previous part, but you can probably add it at least for this feature.

api/src/main/java/io/strimzi/api/kafka/model/kafka/listener/KafkaListenerAuthentication.java
scholzj: Yeah, I think what you suggest would work fine here.

api/src/main/java/io/strimzi/api/kafka/model/common/authentication/KafkaClientAuthentication.java
mstruk: Updated the text.

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/ListenersValidator.java
```diff
@@ -656,6 +661,31 @@ private static void validateOauth(Set<String> errors, GenericKafkaListener liste
...
+            oAuth.setIncludeAcceptHeader(false);
+        }
+        if (oAuth.getCheckAccessTokenType() == null || oAuth.getCheckAccessTokenType()) {
+            if (oAuth.getCheckAccessTokenType() != null && oAuth.getCheckAccessTokenType()) {
+                LOGGER.warnOp("'checkAccessTokenType' force-set to 'false' for compatibility with 'serviceaccount-oauth'");
+            }
+            oAuth.setCheckAccessTokenType(false);
+        }
+    }
+
```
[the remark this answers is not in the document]
mstruk: I had second thoughts about that but I couldn't find a better way to perform the initialisation of these fields after CR configured values have already been applied. Validation is the logic to always be called at least once after all the values have been set and invoking this method multiple times doesn't change the outcome. This approach is least disruptive and gets the work done (unless I'm missing something very important there that's not completely obvious).

I'm not sure about the proposed alternatives. I guess the `validIssuerUri`, `jwksEndpointUri`, `serverBearerTokenLocation` can be set by overriding setter and getter methods with new default value annotations. Options `includeAcceptHeader` and `checkAccessTokenType` have to be fixed to non-default values, maybe that can be achieved by hardcoding a value into a getter override? But then there is this issue of silently ignoring the attempt to override these values - that's why the WARN logging is logged if an explicit attempt to set the wrong value of these attributes is detected.

Considering the Kafka configuration builder approach - I guess the challenge would again be with `includeAcceptHeader` and `checkAccessTokenType` as we have to make it impossible for the user to override the configuration with a new value - i.e. we have to perform the check and override after all the attributes have been set. These values should be set to defaults and not overridden with a different value.

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/ListenersValidator.java
```diff
@@ -656,6 +661,31 @@ private static void validateOauth(Set<String> errors, GenericKafkaListener liste
         }
     }
 
+    private static void validateAndOverrideServiceAccountOAuth(KafkaListenerAuthenticationOAuth oAuth) {
+        if (oAuth.getValidIssuerUri() == null) {
+            oAuth.setValidIssuerUri("https://kubernetes.default.svc.cluster.local");
+        }
+        if (oAuth.getJwksEndpointUri() == null) {
+            oAuth.setJwksEndpointUri("https://kubernetes.default.svc.cluster.local/openid/v1/jwks");
```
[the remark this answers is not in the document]
mstruk: Ok, I'll take a look at DnsNameGenerator.

documentation/api/io.strimzi.api.kafka.model.kafka.listener.KafkaListenerAuthenticationServiceAccountOAuth.adoc
```diff
@@ -0,0 +1,41 @@
...
+
+ServiceAccount OAuth uses Kubernetes API server's JWKS endpoint to validate service account tokens.
+
+.An example of ServiceAccount OAuth authentication on the listener
+[source,yaml,subs=attributes+]
+----
+authentication:
+  type: serviceaccount-oauth
+  maxSecondsWithoutReauthentication: 3600
+  customClaimCheck: "@.['kubernetes.io'] && @.['kubernetes.io'].['namespace'] in ['example']"
```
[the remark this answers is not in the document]
mstruk: I guess we can do that.

mstruk:
> I left some comments. But I'm afraid this really suffers from the lack of proposal. I guess we screwed this up :-(. Sorry.

I guess we can still make a proposal. The implementation proposal itself is also a way to open up questions we otherwise maybe we wouldn't think of :)

> For the client authentication, we seem to have no unique logic for it compared to a type: oauth authentication with the token path specified (out of curiosity, is the token reloaded from the file periodically? Or how does it work with short-lived tokens?). But by adding a new custom type, it means we need to duplicate a lot of tests. Is it worth it? Should we just have a the docs / examples show how to do it with tape: oauth to save the effort?

That's a fair point, but hard for me to say whether the tradeoff is worth it or not. I see `serviceaccount-oauth` as a zero-conf kind of `oauth`. If it's zero-conf on the server, why not also have zero-conf on the client. No typos, no doubts. But yes, it means extra logic, extra tests.

To answer the token loading question - the token is loaded from the file on demand every time the login callback is called - which is when the relogin procedure happens on the client - some short time before the token expires as handled by the Kafka Client library.

> For the server side, there is a lot more stuff that is auto-configured. I think it is a bit weird to have the exact same API used for it. But I guess the user-friendliness here is worth it and we should keep it and just fix the setting of the defaults in the validation?

Having the same API makes it obvious that it's essentially an OAuth and allows you to set any properties that you can set when configuring OAuth. `customClaimCheck` for example, is one such thing.

Again, I think there is a value for the users in (almost) zero-conf approach. No typos ...

> @mstruk Will you be able to join the next community call? It might be good place to discuss and decide this.

Yes, I'll put it on my calendar.

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/ListenersValidator.java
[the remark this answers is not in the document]
mstruk: By having `KafkaClientAuthenticationServiceAccountOAuth` extend `KafkaClientAuthenticationOAuth`, and `KafkaListenerAuthenticationServiceAccountOAuth` extend `KafkaListenerAuthenticationOAuth` the same code that works to validate 'oauth' can be used to validate 'serviceaccount-oauth'. Otherwise that code would essentially have to be duplicated.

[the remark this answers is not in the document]
mstruk: After thinking some more about this I made an alternative implementation which does away with API duplication.
See: https://github.com/strimzi/strimzi-kafka-operator/pull/10260

mstruk:
We decided at the Strimzi Community Meeting on 27/6/2024 to hold further work on this feature for now, and to provide `oauth` authentication configuration examples in the documentation how to manually configure the use of Kubernetes API server as an authorization server in order to use Kubernetes Service Account on the listeners and the clients.

The additional enabling config options will be part of  #9970.

PR #9663, opened 2024-02-09
Use Integer instead of int in `KafkaClusterSpec` replicas

im-konge:
### Type of change

- Enhancement

### Description

This PR changes the `int` as a type for `replicas` to `Integer`, so we can set it as `null` when the `KafkaBuilder` for the `Kafka` object is used -> for applying using fabric8 API.
It also contains few `null` checks. 

### Checklist

### DIFF
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/model/nodepools/VirtualNodePoolConverter.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/model/nodepools/VirtualNodePoolConverter.java
@@ -13,7 +13,6 @@ import io.strimzi.api.kafka.model.nodepool.KafkaNodePoolTemplateBuilder;
 import io.strimzi.api.kafka.model.nodepool.ProcessRoles;
 
 import java.util.List;
-import java.util.Optional;
 import java.util.stream.IntStream;
 
 /**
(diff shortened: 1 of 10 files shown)


cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaClusterCreator.java
```diff
@@ -208,7 +209,7 @@ private Future<KafkaAndNodePools> revertScaleDown(KafkaCluster kafka, Kafka kafk
         if (scaleDownCheckFailed) {
             if (nodePoolCrs == null || nodePoolCrs.isEmpty()) {
                 // There are no node pools => the Kafka CR is used
-                int newReplicasCount = kafkaCr.getSpec().getKafka().getReplicas() + kafka.removedNodes().size();
+                int newReplicasCount = Optional.ofNullable(kafkaCr.getSpec().getKafka().getReplicas()).orElse(0) + kafka.removedNodes().size();
```
scholzj, line 212: If there are no node pools, it cannot be null.

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/nodepools/VirtualNodePoolConverter.java
```diff
@@ -48,7 +49,7 @@ public static KafkaNodePool convertKafkaToVirtualNodePool(Kafka kafka, Integer e
                     .withLabels(kafka.getMetadata().getLabels())
                 .endMetadata()
                 .withNewSpec()
-                    .withReplicas(kafka.getSpec().getKafka().getReplicas())
+                    .withReplicas(Optional.ofNullable(kafka.getSpec().getKafka().getReplicas()).orElse(0))
```
scholzj, line 52: If there are no node pools, it cannot be null.

cluster-operator/src/test/java/io/strimzi/operator/cluster/ResourceUtils.java
```diff
@@ -232,7 +232,7 @@ public static Secret createInitialCaKeySecret(String clusterNamespace, String cl
                 .build();
     }
 
-    public static Kafka createKafka(String namespace, String name, int replicas,
+    public static Kafka createKafka(String namespace, String name, Integer replicas,
```
scholzj, line 235: Do you really need to change this? Do you expect to pass null somewhere to these methdos?

scholzj (approved):

ppatierno (approved):

PR #9667, opened 2024-02-09
[ST] fixing unnecessary error logs when cleaning resources

henryZrncik:
### Type of change

- Refactoring

### Description
all changes made in order to avoid error logs (mostly null pointers) when removing Kafka resources and namespaces. 

- added one check to make sure kafka is present when cleaning Kafka resource. (solve problems with "Kafka" null pointers)
- added default value for "testMethodName" when setting up cluster Operator namespaces and Logs Collecting, this is expected by the way we are creating them e.g. causing unexpected behaviour when trying to remove namespace.

additionally removal of several explicit kafka deletions in the end of the test as they are not necessary there (and are commonly no longer used as we have cleaning set up).

### DIFF
--- a/systemtest/src/main/java/io/strimzi/systemtest/resources/ResourceManager.java
+++ b/systemtest/src/main/java/io/strimzi/systemtest/resources/ResourceManager.java
@@ -346,6 +346,10 @@ public class ResourceManager {
                 type.delete(resource);
                 assertTrue(waitResourceCondition(resource, ResourceCondition.deletion()),
                         String.format("Timed out deleting %s %s/%s", resource.getKind(), resource.getMetadata().getNamespace(), resource.getMetadata().getName()));
+            } catch (NullPointerException e) {
+                // some of the resources may be actually deleted during the test causing extra error logs when trying to delete resource which is no longer present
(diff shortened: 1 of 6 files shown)


henryZrncik:
the fail of the test in Azure has nothing to do with changes. The test is actually flaky and fails on both this and main branch as well (approx 40% of time) will be fixed in separe from this PR.

see-quick:
Well but for sure you should resolve problem with spotbugs i.e., 
```java
[INFO] BugInstance size is 1
[INFO] Error size is 0
[INFO] Total bugs: 1
[ERROR] Medium: Do not catch NullPointerException like in io.strimzi.systemtest.resources.ResourceManager.deleteResource(HasMetadata[]) [io.strimzi.systemtest.resources.ResourceManager] At ResourceManager.java:[line 349] DCN_NULLPOINTER_EXCEPTION
[INFO] 
```

henryZrncik:
1. (@see-quick) yeah, spotbugs problem should be removed 
2. (@im-konge) Also I will cover similar cases when waiting for resources in other PR as well.

see-quick (approved):

PR #9679, opened 2024-02-12
Set STRIMZI_USE_KRAFT_IN_TESTS=true also for x86 on TF

Frawless:
### Type of change

- Bugfix

### Description

In https://github.com/strimzi/strimzi-kafka-operator/pull/9627 I set the KRaft ST env var only for arm arch. This PR solves it.

### Checklist


im-konge (approved):

kornys (approved):
But my approval means nothing!

see-quick (approved):
👍

PR #9681, opened 2024-02-13
docs(node pools): adds a procedure to describe role transitions

PaulRMellor:
**Documentation**

Updates to the node pool content, to include:

- New procedure for _Transitioning to separate broker and controller roles_
- Limitations on managing node pools with controller role
- Minor consistency changes

### Checklist

- [ ] Write tests
- [x] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/documentation/modules/configuring/con-config-node-pool-roles.adoc
+++ b/documentation/modules/configuring/con-config-node-pool-roles.adoc
@@ -15,7 +15,7 @@ For example, you may have a node pool that contains nodes that perform dual brok
 In this case, you create a new node pool with nodes that act only as brokers, and then reassign partitions from the dual-role nodes to the new brokers.
 You can then switch the old node pool to a controller-only role.
 
-WARNING: Although it is possible to split node pool roles in this way, it not currently possible to transition a dedicated node pool containing only controller or broker nodes to a combined node pool where nodes operate as brokers and controllers.
+WARNING: Although it is possible to split node pool roles in this way, it is not currently possible to transition a dedicated node pool containing only controller or broker nodes to a combined node pool where nodes operate as brokers and controllers.
 
 When removing `broker` roles in the node pool configuration, keep in mind that Kafka does not automatically reassign partitions.
 Before removing the broker role, ensure that nodes changing to controller-only roles do not have any assigned partitions. 
--- a/documentation/modules/configuring/con-config-node-pool-roles.adoc
+++ b/documentation/modules/configuring/con-config-node-pool-roles.adoc
@@ -15,6 +15,8 @@ For example, you may have a node pool that contains nodes that perform dual brok
 In this case, you create a new node pool with nodes that act only as brokers, and then reassign partitions from the dual-role nodes to the new brokers.
 You can then switch the old node pool to a controller-only role.
 
+WARNING: Although it is possible to split node pool roles in this way, it not currently possible to transition a dedicated node pool containing only controller or broker nodes to a combined node pool where nodes operate as brokers and controllers.
+
 When removing `broker` roles in the node pool configuration, keep in mind that Kafka does not automatically reassign partitions.
 Before removing the broker role, ensure that nodes changing to controller-only roles do not have any assigned partitions. 
 If partitions are assigned, the change is prevented.
--- a/documentation/modules/configuring/proc-managing-storage-affinity-node-pools.adoc
+++ b/documentation/modules/configuring/proc-managing-storage-affinity-node-pools.adoc
@@ -109,7 +109,7 @@ spec:
 ----
 
 . Apply the node pool configuration.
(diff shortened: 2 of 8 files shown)


documentation/modules/configuring/proc-managing-storage-affinity-node-pools.adoc
```diff
@@ -109,17 +109,17 @@ spec:
 ----
 
 . Apply the node pool configuration.
-. Check the status of the deployment and wait for the pods in the node pools to be created and have a status of `READY`.
+. Check the status of the deployment and wait for the pods in the node pools to be created and have a status of `Running`.
```
scholzj, line 112: I don't think this is correct. The status running does not mean they are ready. We want the pods to be Ready => i.e. the `READY` column should be `1/1` (Or 2/2 etc. in case of Entitty Operator and so on). This should apply here as well as in the other cases.

documentation/modules/configuring/proc-splitting-node-pool-roles.adoc
```diff
@@ -0,0 +1,145 @@
...
+//
+// assembly-config.adoc
+
+[id='proc-splitting-node-pools-roles-{context}']
+= Transitioning to separate broker and controller roles
+
+[role="_abstract"]
+This procedure describes how to transition to using node pools with separate roles.
+If your Kafka cluster is using a node pool with combined controller and broker roles, you can transition to using two node pools with separate roles.
+To do this, rebalance the cluster to move partition replicas to the node pool with a broker-only role, and then switch the old node pool to a controller-only role.
```
scholzj, line 11:
```suggestion
To do this, add a new broker-only node pool, rebalance the cluster to move partition replicas to the node pool with a broker-only role, and then switch the old node pool to a controller-only role.
```

documentation/modules/configuring/proc-splitting-node-pool-roles.adoc
```diff
@@ -0,0 +1,145 @@
...
+my-cluster-pool-b-3  1/1    Running  0
+my-cluster-pool-b-4  1/1    Running  0
+my-cluster-pool-b-5  1/1    Running  0
+---- 
++
+Node IDs are appended to the name of the node on creation.
+
+. Use the Cruise Control `remove-brokers` mode to reassign partition replicas from the dual-role nodes to the newly added brokers.
++
+The reassignment can take some time depending on the number of topics and partitions in the cluster.
```
scholzj, line 115: Should we add an example here?

```yaml
kind: KafkaRebalance
metadata:
  # ...
spec:
  mode: remove-brokers
  brokers: [0, 1, 2]
```

documentation/modules/configuring/con-config-node-pool-roles.adoc
```diff
@@ -15,6 +15,8 @@ For example, you may have a node pool that contains nodes that perform dual brok
 In this case, you create a new node pool with nodes that act only as brokers, and then reassign partitions from the dual-role nodes to the new brokers.
 You can then switch the old node pool to a controller-only role.
 
+WARNING: Although it is possible to split node pool roles in this way, it not currently possible to transition a dedicated node pool containing only controller or broker nodes to a combined node pool where nodes operate as brokers and controllers.
```
henryZrncik, line 18: Works for me just just mentioning that transitions per se is possible but "semi-functional" , for example adding broker role would take a bit longer time and few failed reconciliations but desired nodes would roll  (in ~10 minutes).

henryZrncik (approved):

documentation/modules/configuring/proc-splitting-node-pool-roles.adoc
scholzj: Ok 👍

documentation/modules/configuring/con-config-node-pool-roles.adoc
scholzj: I don't think we want to mention in the docs that somehting is _semifunctonal_ because it timeouts after a long time 😉

documentation/modules/configuring/proc-splitting-node-pool-roles.adoc
PaulRMellor: If we do, I think we should do the same for the other node pool procedures.
I'll go ahead.

documentation/modules/configuring/con-config-node-pool-roles.adoc
henryZrncik: Right! That "mentioning" in my comment was "me mentioning it" not recommendation to mention it here 😀

PaulRMellor: Yeah. No change on the warning. Thanks @jankalinic

scholzj (approved):

ppatierno (approved):

PR #9687, opened 2024-02-14
[ST] Use ThreadLocal in ResourceManager

kornys:
### Type of change

- Enhancement / new feature
- Refactoring

### Description

Currently due to parallelism of tests every method for creating and work with resources has to pass junit extension context which is overkill and I would say no new developer understand it.
Instead of passing extension context for identifying test context in parallel run, I added ThreadLocal which can hold extenstion context for every thread in which test is running. With this change test suite itself set enxtensionContext of the test Method or Class and test developer does not need to pass extenstion context in methods so code looks more logical and clean. 

### Checklist

- [ ] Update documentation


Frawless (approved):

im-konge (approved):

see-quick (approved):
This is really great improvement! Thanks for that @kornys ! 💯 👍

henryZrncik (approved):

PR #9695, opened 2024-02-15
Fixed OAuth configuration on mixed node test

ppatierno:
This PR fixes the "mixed node" use case in the `testKraftOauthBrokerControllerAndMixedNodes`.
It should be about mixed node (node = 14) but then in the configuration builder it's not set as mixed.
At the same time, the expectation is of course wrong, because being mixed it has to have the CONTROLPLANE-9090 in the listeners.


scholzj (approved):

im-konge (approved):
Thanks for the fix and sorry for the issues 🙂

PR #9700, opened 2024-02-16
[ST] Migration - tests for migration with CO deletion during process, full path with migration and rollback

im-konge:
### Type of change

- Enhancement

### Description

This PR adds two more testcases and does few modification to the `MigrationST`.

Testcases:

- migration process with CO deletion and ZK `deleteClaim: false` - this testcase is similar to the original one, there are two differences -> first one is that during each rolling update the CO Pod is deleted. This will verify that CO is able to recover and finish the rolling update of the particular migration phase without an issue. The second difference is in the `deleteClaim` set in ZK. In the original test-cases (and everywhere else) we are using `deleteClaim` se to `true`, so we are checking, that everything related to ZK (also the PVCs) is deleted as a last step of the migration. With the `deleteClaim: false` the PVCs should be there and CO should not delete them.

- migration -> rollback -> migration - combination of the previous testcases. This testcase verifies, that even if we rollback the migration, we are able to do the full migration without issues. This can be typical use-case when the user would like to migrate from ZK to KRaft, then he changes his mind and rollback the migration, but in the end he decides to migrate it to KRaft (finally)

Together with these two testcases I'm changing the javadoc (re-ordering the steps), adding new checks, refactors current methods.

### Checklist

- [ ] Write tests


systemtest/src/test/java/io/strimzi/systemtest/migration/MigrationST.java
```diff
@@ -103,12 +106,66 @@ public class MigrationST extends AbstractST {
...
+     *  15. - Deletes the ClusterOperator Pod
+     *  16. - Waits for second rolling update of Broker Pods to be finished - removing dependency on ZK
+     *  17. - Checks that Kafka CR has .status.kafkaMetadataState set to KRaftPostMigration
+     *  18. - Creates a new KafkaTopic and checks both ZK and KRaft metadata for presence of the KafkaTopic
+     *  19. - Does immediate message transmission to the new KafkaTopic
+     *  20. - Finishes the migration - annotates the Kafka resource with strimzi.io/kraft:enabled
+     *  21. - Waits for one of the controller Pods to start rolling update
+     *  22. - Deletes the ClusterOperator Pod
+     *  23. - Waits for rolling update of Controller Pods will be finished
+     *  24. - ZK related resources will be deleted
```
ppatierno, line 147: but not the PVCs right? It's better to be clear here.

systemtest/src/test/java/io/strimzi/systemtest/migration/MigrationST.java
```diff
@@ -149,13 +206,90 @@ void testMigrationFromZkToKRaft() {
     @IsolatedTest
     void testRollbackDuringMigration() {
         TestStorage testStorage = new TestStorage(ResourceManager.getTestContext(), TestConstants.CO_NAMESPACE);
+        boolean zkDeleteClaim = true;
+        boolean deleteCoDuringProcess = false;
```
ppatierno, line 210: can't we just pass the boolean(s) instead of declaring these local variables?

systemtest/src/test/java/io/strimzi/systemtest/migration/MigrationST.java
```diff
@@ -149,13 +206,90 @@ void testMigrationFromZkToKRaft() {
...
-        doFirstPartOfMigration(testStorage);
+        setupMigrationTestCase(testStorage, zkDeleteClaim);
+        doFirstPartOfMigration(testStorage, deleteCoDuringProcess);
         doRollback(testStorage);
     }
 
-    private void setupMigrationTestCase(TestStorage testStorage) {
+    /**
+     * @description This testcase is focused on full circle of the migration process. The test starts with the migration,
+     * then, after the first phase, the migration is rolled back. After the rollback process is successfully completed, the test
```
ppatierno, line 219: talking about rollback process completed could be misleading, because in the test you are moving from KRaftPostMigration to KRaftDualWriting but you are not moving back to ZooKeeper, which means that the rollback is really completed. Let's say you are stopping and resuming migration at half-way of a rollback process.

systemtest/src/test/java/io/strimzi/systemtest/migration/MigrationST.java
im-konge: yeah we can, just in terms of the review and clarity I left it there

systemtest/src/test/java/io/strimzi/systemtest/migration/MigrationST.java
im-konge: yeah gonna change the comment here

ppatierno (approved):

PR #9708, opened 2024-02-19
Add topicId to KafkaTopic status

### LINKED ISSUE #9651: [Enhancement]: Add topicId to KafkaTopicStatus
### Related problem

I am working with a component that uses the Kafka `topicId` to identify a topic. Locating a `KafkaTopic` CR for a given `topicId` requires an additional describe operation to obtain the name to then find the CR.

### Suggested solution

Publish the `topicId` in the status section of the KafkaTopic custom resource.

### Alternatives

_No response_

### Additional context

Depending on the timeline for deprecating the bidirectional topic operator, this may be useful for both version of the TO.

fvaleri:
This should close #9651.

### DIFF
--- a/api/src/main/java/io/strimzi/api/kafka/model/topic/KafkaTopicStatus.java
+++ b/api/src/main/java/io/strimzi/api/kafka/model/topic/KafkaTopicStatus.java
@@ -39,7 +39,8 @@ public class KafkaTopicStatus extends Status {
         this.topicName = topicName;
     }
 
-    @Description("Topic id")
+    @Description("The topic's id. For a KafkaTopic with the ready condition, " +
+        "this will change only if the topic gets deleted and recreated with the same name.")
     public String getTopicId() {
         return topicId;
     }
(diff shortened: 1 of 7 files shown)


scholzj (approved):
LGTM. Thanks.

api/src/main/java/io/strimzi/api/kafka/model/topic/KafkaTopicStatus.java
```diff
@@ -37,4 +38,13 @@ public String getTopicName() {
     public void setTopicName(String topicName) {
         this.topicName = topicName;
     }
+
+    @Description("Topic id")
```
tombentley, line 42: We're exposing this to client applications. We should try to define what it means so such applications can use it correctly. This is a bit tricky because we have the possibility of conflicting KafkaTopics. I think this is OK:

```suggestion
    @Description("The topic's id. For a KafkaTopic with the ready condition, this will change only if the topic gets deleted and recreated with the same name.")
```

tombentley (approved):
One comment, but otherwise this LGTM. Thanks @fvaleri.

ppatierno (approved):
LGTM

PR #9709, opened 2024-02-19
[ST] Use correct Kafka resource for each mode (ZK without NP / ZK with NP / KRaft)

im-konge:
### Type of change

- Bugfix / Enhancement

### Description

This PR refactors `KafkaTemplates` class, so during STs the correct Kafka resource is used. Currently, when we are running the STs, we are getting a lot of warnings about specified fields, that are not used:

In ZK with NodePools:
- `.spec.kafka.replicas` is used
- `.spec.kafka.storage` is used

In KRaft then:
- `.spec.zookeeper` is used
- LMFV and IBPV are used in config
- `metadataVersion` is not set (IIRC that's not a warning, but for LMFV and IBPV we are setting a version here, so we should do a same thing in KRaft as well)

Also, `kafkaWithMetrics` and `kafkaWithMetricsAndCruiseControlWithMetrics` contain creation of `ConfigMap`s, which IMHO shouldn't be part of the particular methods, but the creation should be done in `ResourceManager`, thus I removed it from those two methods and created separate one just for the `ConfigMap`s. Those have to applied before applying `Kafka` resource, but that's something what would user anyway do and it's not hidden.

As part of this PR I moved out common code and I created methods that are used in particular `default*` templates.

### Checklist

### DIFF
--- a/systemtest/src/main/java/io/strimzi/systemtest/templates/crd/KafkaTemplates.java
+++ b/systemtest/src/main/java/io/strimzi/systemtest/templates/crd/KafkaTemplates.java
@@ -17,6 +17,7 @@ import io.strimzi.api.kafka.model.kafka.Kafka;
 import io.strimzi.api.kafka.model.kafka.KafkaBuilder;
 import io.strimzi.api.kafka.model.kafka.listener.GenericKafkaListenerBuilder;
 import io.strimzi.api.kafka.model.kafka.listener.KafkaListenerType;
+import io.strimzi.operator.common.Annotations;
 import io.strimzi.systemtest.Environment;
 import io.strimzi.systemtest.TestConstants;
 import io.strimzi.systemtest.utils.TestKafkaVersion;
@@ -39,6 +40,7 @@ public class KafkaTemplates {
     // -------------------------------------------------------------------------------------------
     // Kafka Ephemeral
     // -------------------------------------------------------------------------------------------
+
     public static KafkaBuilder kafkaEphemeral(String clusterName, int kafkaReplicas) {
         return kafkaEphemeral(clusterName, kafkaReplicas, Math.min(kafkaReplicas, 3));
     }
@@ -294,7 +296,6 @@ public class KafkaTemplates {
         applyDefaultConfigurationOfZookeeperKafka(kb, zookeeperReplicas);
         applyDefaultLogging(kb, true);
         applyMemoryRequestsAndLimitsIfNeeded(kb, true);
-        removeFieldsNotRelatedToParticularMode(kb, true);
 
         return kb;
     }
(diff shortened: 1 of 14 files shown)


systemtest/src/main/java/io/strimzi/systemtest/templates/crd/KafkaTemplates.java
```diff
@@ -66,189 +121,247 @@ public static KafkaBuilder kafkaPersistent(String name, int kafkaReplicas, int z
...
-                    .withVersion(Environment.ST_KAFKA_VERSION)
-                    .withReplicas(kafkaReplicas)
-                    .addToConfig("log.message.format.version", TestKafkaVersion.getKafkaVersionsInMap().get(Environment.ST_KAFKA_VERSION).protocolVersion())
+                    .withMetadataVersion(TestKafkaVersion.getKafkaVersionsInMap().get(Environment.ST_KAFKA_VERSION).metadataVersion())
+                .endKafka()
+            .endSpec();
+
+        applyDefaultSpecOfKafka(kb, kafkaReplicas);
+        applyDefaultLogging(kb, false);
+        applyMemoryRequestsAndLimitsIfNeeded(kb, false);
```
Frawless, line 334: would be better to name them with `set` prefix instead of `apply` ?

systemtest/src/main/java/io/strimzi/systemtest/templates/crd/KafkaTemplates.java
```diff
@@ -66,189 +121,247 @@ public static KafkaBuilder kafkaPersistent(String name, int kafkaReplicas, int z
...
+        kb = removeFieldsNotRelatedToParticularMode(kb, false);
```
Frawless, line 335: Why this method returns KafkaBuilder and the above ones no?

systemtest/src/main/java/io/strimzi/systemtest/templates/crd/KafkaTemplates.java
```diff
@@ -304,12 +423,70 @@ private static KafkaBuilder defaultKafka(Kafka kafka, String name, int kafkaRepl
...
+                    .addToConfig("transaction.state.log.min.isr", Math.min(kafkaReplicas, 2))
+                    .addToConfig("transaction.state.log.replication.factor", Math.min(kafkaReplicas, 3))
+                    .addToConfig("default.replication.factor", Math.min(kafkaReplicas, 3))
+                    .addToConfig("min.insync.replicas", Math.min(Math.max(kafkaReplicas - 1, 1), 2))
+                .endKafka()
+            .endSpec();
+    }
+
+    private static KafkaBuilder removeFieldsNotRelatedToParticularMode(KafkaBuilder kafkaBuilder, boolean withZookeeper) {
+        Kafka kafka = kafkaBuilder.build();
```
Frawless, line 470: is this needed? `.withZookeeper(null)` on builder doesn't work?

systemtest/src/main/java/io/strimzi/systemtest/templates/crd/KafkaTemplates.java
see-quick: Because the above methods change the state of KafkaBuilder object but do not set them to `null`. If you set null to for instance ZK without building object it would cause an NPE. Builder does not support nullifying such complex objects I would say.

systemtest/src/main/java/io/strimzi/systemtest/templates/crd/KafkaTemplates.java
im-konge: Hmm I can try that.
I did it this way because IIRC when there was `null`, the default was used.
But it was in case of TopicOperator, so maybe this will be anyway needed for TO.

henryZrncik (approved):
LGTM, thanks for work even warning messages seems to be working as expected.

Frawless (approved):

systemtest/src/main/java/io/strimzi/systemtest/templates/crd/KafkaTemplates.java
Frawless: Makes sense, thanks for explanation

see-quick (approved):

PR #9720, opened 2024-02-21
Added KafkaAssemblyOperator migration-related tests

ppatierno:
This PR adds tests with using KafkaAssemblyOperator related to migration and rollback.

### DIFF
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaAssemblyOperator.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaAssemblyOperator.java
@@ -284,9 +284,9 @@ public class KafkaAssemblyOperator extends AbstractAssemblyOperator<KubernetesCl
     class ReconciliationState {
         private final String namespace;
         private final String name;
-        private final Kafka kafkaAssembly;
-        private final Reconciliation reconciliation;
-        private final KafkaMetadataStateManager kafkaMetadataStateManager;
+        /* test */ final Kafka kafkaAssembly;
+        /* test */ final Reconciliation reconciliation;
+        /* test */ final KafkaMetadataStateManager kafkaMetadataStateManager;
 
         /* test */ KafkaVersionChange versionChange;
 
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaAssemblyOperator.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaAssemblyOperator.java
@@ -284,9 +284,9 @@ public class KafkaAssemblyOperator extends AbstractAssemblyOperator<KubernetesCl
     class ReconciliationState {
         private final String namespace;
         private final String name;
-        /* test */ final Kafka kafkaAssembly;
-        /* test */ final Reconciliation reconciliation;
-        /* test */ final KafkaMetadataStateManager kafkaMetadataStateManager;
+        private final Kafka kafkaAssembly;
+        private final Reconciliation reconciliation;
+        private final KafkaMetadataStateManager kafkaMetadataStateManager;
 
         /* test */ KafkaVersionChange versionChange;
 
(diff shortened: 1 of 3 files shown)


cluster-operator/src/test/java/io/strimzi/operator/cluster/operator/assembly/KRaftMigrationMockTest.java
```diff
@@ -0,0 +1,434 @@
...
+    }
+
+    static class MockKafkaReconciler extends KafkaReconciler {
+
+        public MockKafkaReconciler(Reconciliation reconciliation, Kafka kafkaCr, List<KafkaNodePool> nodePools, KafkaCluster kafka, ClusterCa clusterCa, ClientsCa clientsCa, ClusterOperatorConfig config, ResourceOperatorSupplier supplier, PlatformFeaturesAvailability pfa, Vertx vertx, KafkaMetadataStateManager kafkaMetadataStateManager) {
+            super(reconciliation, kafkaCr, nodePools, kafka, clusterCa, clientsCa, config, supplier, pfa, vertx, kafkaMetadataStateManager);
+        }
+
+        @Override
+        protected KafkaAgentClient initKafkaAgentClient(Reconciliation reconciliation, Secret clusterCaCertSecret, Secret coKeySecret) {
```
scholzj, line 427: Should we use some `AgentClientSupplier` instead of this to make it easier to provide custom implementation?

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaAssemblyOperator.java
```diff
@@ -284,9 +284,9 @@ ReconciliationState createReconciliationState(Reconciliation reconciliation, Kaf
     class ReconciliationState {
         private final String namespace;
         private final String name;
-        private final Kafka kafkaAssembly;
-        private final Reconciliation reconciliation;
-        private final KafkaMetadataStateManager kafkaMetadataStateManager;
+        /* test */ final Kafka kafkaAssembly;
+        /* test */ final Reconciliation reconciliation;
+        /* test */ final KafkaMetadataStateManager kafkaMetadataStateManager;
```
scholzj, line 289: Why is this needed?

cluster-operator/src/test/java/io/strimzi/operator/cluster/operator/assembly/KRaftMigrationMockTest.java
ppatierno: This was exactly on my notes to do. I am happy we are on the same page. I just prefer to not make this change in this PR. It could have more impact and change on other classes, because I would change the ResourceOperatorSupplier. I assume.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaAssemblyOperator.java
ppatierno: Because of the mocking of the reconciliation state when it needs to pass them to the mocked kafka reconciler.

cluster-operator/src/test/java/io/strimzi/operator/cluster/operator/assembly/KRaftMigrationMockTest.java
ppatierno: See other comment.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaAssemblyOperator.java
scholzj: doesn't that beat the purpose of the test? The purpose of the test is to see how it flows -> but if you manipulate the internal states in between it does not fulfill that purpose I guess?

ppatierno: As discussed offline, I will put together a PR on main to provide Kafka agent client via the ResourceSupplier and then using it here.

ppatierno: I opened this one https://github.com/strimzi/strimzi-kafka-operator/pull/9723
When that one is merged, I will simplify the mocking here.

ppatierno:
@scholzj I simplified the test related to how it's using a mocked Kafka Agent client. Can you have another pass, please?

PR #9721, opened 2024-02-21
Fix report tool when getting CO replicasets

fvaleri:
We can have up to 10 replicasets for the cluster operator, depending on the number of updates that have been done using deployment. The report tool fails with the following error when having more than one replicaset.

Error from server (NotFound): replicasets.apps "strimzi-cluster-operator-85b9b646cc\nstrimzi-cluster-operator-86948f6756" not found


scholzj (approved):
LGTM assuming it works.

ppatierno (approved):
LGTM assuming you tested it.

PR #9733, opened 2024-02-22
Added unit tests for `ZookeeperReconciler` class

ShubhamRwt:
### Type of change

_Select the type of your PR_

- Enhancement / new feature

### Description

This PR adds the unit tests for the `ZookeeperReconciler` class

### Checklist

- [ ] Write tests
- [ ] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/ZooKeeperReconciler.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/ZooKeeperReconciler.java
@@ -110,7 +110,7 @@ public class ZooKeeperReconciler {
 
     private String loggingHash = "";
 
-    private final boolean isKRaftMigrationRollback;
+    /*test*/ final boolean isKRaftMigrationRollback;
 
     /**
      * Constructs the ZooKeeper reconciler
@@ -207,7 +207,7 @@ public class ZooKeeperReconciler {
                 .compose(i -> serviceEndpointsReady())
                 .compose(i -> headlessServiceEndpointsReady())
                 .compose(i -> deletePersistentClaims())
-                .compose(i -> maybeDeleteControllerZnode());
+                .compose(i -> this.isKRaftMigrationRollback ? maybeDeleteControllerZnode() : Future.succeededFuture());
     }
 
     /**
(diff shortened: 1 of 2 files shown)


cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/ZooKeeperReconciler.java
```diff
@@ -880,25 +880,19 @@ protected Future<Void> deletePersistentClaims() {
      */
     protected Future<Void> maybeDeleteControllerZnode() {
```
ppatierno, line 881: It's not a "maybe" anymore because it's going to delete the `/controller` znode so I would rename it `deleteControllerZnode`

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/ZooKeeperReconciler.java
```diff
@@ -207,7 +207,7 @@ public Future<Void> reconcile(KafkaStatus kafkaStatus, Clock clock)    {
                 .compose(i -> serviceEndpointsReady())
                 .compose(i -> headlessServiceEndpointsReady())
                 .compose(i -> deletePersistentClaims())
-                .compose(i -> maybeDeleteControllerZnode());
+                .compose(i -> this.isKRaftMigrationRollback ? maybeDeleteControllerZnode() : Future.succeededFuture());
```
ppatierno, line 210: if we rename the current `maybeDeleteControllerZnode` to `deleteControllerZnode` (see next comment), I think we should have a new `maybeDeleteControllerZnode` method containing this code, so something like:

```java
protected Future<Void> maybeDeleteControllerZnode() {
    return this.isKRaftMigrationRollback ? maybeDeleteControllerZnode() : Future.succeededFuture();
}
```

so the compose in the reconcile will stay:

```java
.compose(i -> maybeDeleteControllerZnode());
```

This approach makes more understandable the mocking in the test (see related comment later).

cluster-operator/src/test/java/io/strimzi/operator/cluster/operator/assembly/ZookeeperReconcilerKRaftMigrationTest.java
```diff
@@ -0,0 +1,216 @@
...
+    }
+
+    static class MockZooKeeperReconciler extends ZooKeeperReconciler   {
+        public MockZooKeeperReconciler(Reconciliation reconciliation, Vertx vertx, ClusterOperatorConfig config, ResourceOperatorSupplier supplier, PlatformFeaturesAvailability pfa, Kafka kafkaAssembly, KafkaVersionChange versionChange, Storage oldStorage, int currentReplicas, ClusterCa clusterCa, boolean kraftMigrationRollback) {
+            super(reconciliation, vertx, config, supplier, pfa, kafkaAssembly, versionChange, oldStorage, currentReplicas, clusterCa, kraftMigrationRollback);
+        }
+
+        @Override
+        public Future<Void> reconcile(KafkaStatus kafkaStatus, Clock clock)    {
+            return super.isKRaftMigrationRollback ? maybeDeleteControllerZnode() : Future.succeededFuture();
```
ppatierno, line 208: If you apply my suggestion above, this override will just be ...

```java
maybeDeleteControllerZnode();
```

The current approach is dangerous because if we change the logic in the `ZooKeeperReconciler` we need to change it here as well and we can easily forget. By wrapping everything in the "maybe" method it can't happen.

ppatierno (changes requested):
I had some thinking after our offline discussion and left a few comments.

ppatierno (approved):
LGTM

scholzj (approved):

PR #9751, opened 2024-02-28
[ST] fix CC failing test due to default topic rep. factor

henryZrncik:
### Type of change

- Enhancement 

### Description

tests `testDeployAndUnDeployCruiseControl` and `testAutoCreationOfCruiseControlTopicsWithResources` are assuming 2 replicas on CC related topics, which is no longer true. These topics from now follow default topic replication factor as of https://github.com/strimzi/strimzi-kafka-operator/pull/9471 . 
 
now `default.replication.factor` is set explicitly to the same number is is the number of brokers (just to avoid accidental changes).


im-konge (approved):

see-quick (approved):
LGTM 👍 Good job :)

henryZrncik:
the last regression run was cancelled as there were not much to test with it.

PR #9752, opened 2024-02-28
Fix Connect's dynamic log level changes

### LINKED ISSUE #9067: [Bug]: Connect's dynamic log level changes are only applied to one worker node
### Bug Description

If we deploy the following 3-nodes Connect cluster:

```sh
$ kubectl get po
NAME                                         READY   STATUS    RESTARTS   AGE
my-cluster-entity-operator-dbfcd857c-kfz92   3/3     Running   0          25m
my-cluster-kafka-0                           1/1     Running   0          26m
my-cluster-zookeeper-0                       1/1     Running   0          27m
my-connect-cluster-connect-0                 1/1     Running   0          25m
my-connect-cluster-connect-1                 1/1     Running   0          25m
my-connect-cluster-connect-2                 1/1     Running   0          25m
strimzi-cluster-operator-5dff6ccf95-7tp6p    1/1     Running   0          29m
```

The Cluster Operator also creates a Service that loops through all matching Connect's REST endpoints (one for each worker node).

```sh
$ kubectl get svc my-connect-cluster-connect-api -o yaml | yq '.spec'
clusterIP: 10.97.195.99
clusterIPs:
  - 10.97.195.99
internalTrafficPolicy: Cluster
ipFamilies:
  - IPv4
ipFamilyPolicy: SingleStack
ports:
  - name: rest-api
    port: 8083
    protocol: TCP
    targetPort: 8083
selector:
  strimzi.io/cluster: my-connect-cluster
  strimzi.io/kind: KafkaConnect
  strimzi.io/name: my-connect-cluster-connect
sessionAffinity: None
type: ClusterIP
```

The issue is that dynamic logging level changes are only applied to one worker node, depending on the Service status. Setting root level to OFF in the above example, is only applied to my-connect-cluster-connect-0 worker node.

```sh
$ curl -s -X PUT -H "Content-Type:application/json" "$BASE_URL/admin/loggers/root" -d '{"level": "OFF"}'
....

$ kubectl logs my-connect-cluster-connect-0 | tail -n1
2023-09-04 14:07:54,689 INFO 10.244.0.1 - - [04/Sep/2023:14:07:54 +0000] "GET / HTTP/1.1" 200 91 "-" "kube-probe/1.25" 2 (org.apache.kafka.connect.runtime.rest.RestServer) [qtp338765435-42]
$ kubectl logs my-connect-cluster-connect-1 | tail -n1
2023-09-04 14:27:04,695 INFO 10.244.0.1 - - [04/Sep/2023:14:27:04 +0000] "GET / HTTP/1.1" 200 91 "-" "kube-probe/1.25" 4 (org.apache.kafka.connect.runtime.rest.RestServer) [qtp874981105-74]
$ kubectl logs my-connect-cluster-connect-2 | tail -n1
2023-09-04 14:27:04,697 INFO 10.244.0.1 - - [04/Sep/2023:14:27:04 +0000] "GET / HTTP/1.1" 200 91 "-" "kube-probe/1.25" 3 (org.apache.kafka.connect.runtime.rest.RestServer) [qtp1143968486-39]
```

[KIP-976](https://cwiki.apache.org/confluence/display/KAFKA/KIP-976%3A+Cluster-wide+dynamic+log+adjustment+for+Kafka+Connect) has been recently proposed to allow cluster-wide log level changes with just one request. This is something we could potentially use to easily fix this issue.

### Steps to reproduce

_No response_

### Expected behavior

_No response_

### Strimzi version

main

### Kubernetes version

1.25

### Installation method

_No response_

### Infrastructure

_No response_

_No response_

### Additional context

_No response_

fvaleri:
The loggers endpoint now supports the scope parameter, which can be set to cluster in order to change level in all nodes at once. This change simply adds this parameter that works since Kafka 3.7, and causes no harm to previous versions.

This should close #9067.

### DIFF
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaConnectApiImpl.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaConnectApiImpl.java
@@ -405,7 +405,7 @@ class KafkaConnectApiImpl implements KafkaConnectApi {
                                     .write(buffer.toString());
                             request.result().send(response -> {
                                 if (response.succeeded()) {
-                                    if (response.result().statusCode() == 204) {
+                                    if (List.of(200, 204).contains(response.result().statusCode())) {
                                         response.result().bodyHandler(body -> {
                                             LOGGER.debugCr(reconciliation, "Logger {} updated to level {}", logger, level);
                                             result.complete();
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaConnectApiImpl.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaConnectApiImpl.java
@@ -390,7 +390,7 @@ class KafkaConnectApiImpl implements KafkaConnectApi {
     }
 
     private Future<Void> updateConnectorLogger(Reconciliation reconciliation, String host, int port, String logger, String level) {
-        String path = "/admin/loggers/" + logger;
+        String path = "/admin/loggers/" + logger + "?scope=cluster";
         JsonObject levelJO = new JsonObject();
         levelJO.put("level", level);
         LOGGER.debugCr(reconciliation, "Making PUT request to {} with body {}", path, levelJO);
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaConnectApiImpl.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaConnectApiImpl.java
@@ -405,7 +405,7 @@ class KafkaConnectApiImpl implements KafkaConnectApi {
                                     .write(buffer.toString());
                             request.result().send(response -> {
                                 if (response.succeeded()) {
-                                    if (response.result().statusCode() == 200) {
+                                    if (response.result().statusCode() == 204) {
                                         response.result().bodyHandler(body -> {
                                             LOGGER.debugCr(reconciliation, "Logger {} updated to level {}", logger, level);
                                             result.complete();
--- a/cluster-operator/src/test/java/io/strimzi/operator/cluster/operator/assembly/ConnectCluster.java
+++ b/cluster-operator/src/test/java/io/strimzi/operator/cluster/operator/assembly/ConnectCluster.java
@@ -14,6 +14,7 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicReference;
 
 public class ConnectCluster {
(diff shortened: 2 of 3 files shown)


scholzj:
I guess the test failure is related to this?

scholzj:
Interestingly enough when running the whole test suite I get 500, running just the single test gives me another error but with 204.

fvaleri:
Test fixed. With cluster wide logging level update the [response code changes to 204](https://cwiki.apache.org/confluence/display/KAFKA/KIP-976%3A+Cluster-wide+dynamic+log+adjustment+for+Kafka+Connect#KIP976:ClusterwidedynamiclogadjustmentforKafkaConnect-Scopequeryparameter).

scholzj (commented):
Wait, it is not that simple ... how will this work with Kafka 3.6? I assume the new parameter will be ignored which is fine, but it will give you still HTTP 200, or? So you need to check for both / or check the whole range of HTTP OK codes, or?

fvaleri:
> So you need to check for both / or check the whole range of HTTP OK codes, or?

Right, we still need to support both. Thanks.

ppatierno:
@fvaleri I see the test failing because the 500 error.

scholzj:
Yeah, that is what I pointed out -> the test alone works fine with 204, but when the whole test suite is run it gets 500.

fvaleri:
I'll look into that, but it works fine on my end. Do you also see 500 on your local env?

scholzj:
@fvaleri I did saw 500 locally yesterday when running the whole `KafkaConnectApiIT` suit from IDE. And 204 when running only the single `KafkaConnectApiIT.testChangeLoggers` test.

fvaleri:
That was too easy, indeed.

---

The problem is that, unlike the "worker" scope, the "cluster" scope request writes the logging level configuration into the internal -config topic, in order to propagate this information to the other workers. To do this, it uses a KafkaBasedLog instance, which includes a Kafka [producer initialized to null](https://github.com/apache/kafka/blob/trunk/connect/runtime/src/main/java/org/apache/kafka/connect/util/KafkaBasedLog.java#L175) on instance creation. This [producer is created lazily when the service starts](https://github.com/apache/kafka/blob/trunk/connect/runtime/src/main/java/org/apache/kafka/connect/util/KafkaBasedLog.java#L260).

When we start the test Connect instance, services like the ConfigBackingStore, which uses the KafkaBasedLog, are started asynchronously on a different thread. This means that, when our cluster wide log level configuration request arrives, the producer may not have been created yet, which results in the error "IllegalStateException: This KafkaBasedLog was created in read-only mode and does not support write operations".

This is fixed by using the Connect.isRunning() method to wait for all services initialization before running the tests.

---

I also found a second source of flakiness, this time on task restart test. This is due to "ConnectRestException: Cannot complete request momentarily due to no known leader URL, likely because a rebalance was underway".

This may happens when there is a rebalance in progress, as the error suggests, and confirmed by observing the lop. The workaround here is to only use one worker node for this test, which I think is fine as the focus is on the REST API here.

---

After these changes, I ran KafkaConnectApiIT and KafkaConnectorIT many times without issues. Let me know if you also see the same.

scholzj (approved):
LGTM assuming the tests pass.

ppatierno (approved):

PR #9757, opened 2024-02-29
[DOC] Make logging examples less confusing

scholzj:
### Type of change

- Documentation

### Description

The `logging` section in our API can be placed into different sections depending on the custom resource. E.g. `.spec.logging` in Kafka Connect or Bridge, but `.spec.kafka.logging` for Kafka etc. Some of the example snippets in our docs seem to be confusing as they include the `spec:` marker and suggest this should be placed always directly under `.spec`. This PR removes the `spec:` part to leave it more open where the placement is.

### Checklist

- [x] Update documentation


PaulRMellor (approved):
:+1:

ppatierno (approved):

PR #9773, opened 2024-03-04
Fail test when Connect doesn't start in time and increse the wait time

fvaleri:
This should fix the broken CI pipeline.


ppatierno (approved):
LGTM

PR #9782, opened 2024-03-05
[ST] persistent kafka with 1 rep. instead  ephemeral

### LINKED ISSUE #9748: Do not use single replica nodes with ephemeral storage in System Tests
Some tests such as `OauthAuthorizationST.testSuperUserWithOauthAuthorization` seem to be very flaky because they use (at least in KRaft) configurations with single controller and broker nodes with ephemeral storage. Such a setup looses all data in every single rolling update and should not be used in any system tests.

henryZrncik:
### Type of change

- Refactoring

### Description

stop using ephemeral kafka with 1 replica and use either persistent storage or more replicas.

Fixes #9748

### DIFF
--- a/systemtest/src/test/java/io/strimzi/systemtest/mirrormaker/MirrorMaker2ST.java
+++ b/systemtest/src/test/java/io/strimzi/systemtest/mirrormaker/MirrorMaker2ST.java
@@ -113,19 +113,19 @@ class MirrorMaker2ST extends AbstractST {
         // Deploy source and target kafka
         resourceManager.createResourceWithWait(
             NodePoolsConverter.convertNodePoolsIfNeeded(
-                KafkaNodePoolTemplates.brokerPool(testStorage.getNamespaceName(), testStorage.getSourceBrokerPoolName(), testStorage.getSourceClusterName(), 1).build(),
-                KafkaNodePoolTemplates.controllerPool(testStorage.getNamespaceName(), testStorage.getSourceControllerPoolName(), testStorage.getSourceClusterName(), 3).build()
+                KafkaNodePoolTemplates.brokerPoolPersistentStorage(testStorage.getNamespaceName(), testStorage.getSourceBrokerPoolName(), testStorage.getSourceClusterName(), 1).build(),
+                KafkaNodePoolTemplates.controllerPoolPersistentStorage(testStorage.getNamespaceName(), testStorage.getSourceControllerPoolName(), testStorage.getSourceClusterName(), 1).build()
             )
         );
-        resourceManager.createResourceWithWait(KafkaTemplates.kafkaPersistent(testStorage.getSourceClusterName(), 1, 3).build());
+        resourceManager.createResourceWithWait(KafkaTemplates.kafkaPersistent(testStorage.getSourceClusterName(), 1, 1).build());
 
         resourceManager.createResourceWithWait(
             NodePoolsConverter.convertNodePoolsIfNeeded(
-                KafkaNodePoolTemplates.brokerPool(testStorage.getNamespaceName(), testStorage.getTargetBrokerPoolName(), testStorage.getTargetClusterName(), 1).build(),
-                KafkaNodePoolTemplates.controllerPool(testStorage.getNamespaceName(), testStorage.getTargetControllerPoolName(), testStorage.getTargetClusterName(), 3).build()
+                KafkaNodePoolTemplates.brokerPoolPersistentStorage(testStorage.getNamespaceName(), testStorage.getTargetBrokerPoolName(), testStorage.getTargetClusterName(), 1).build(),
+                KafkaNodePoolTemplates.controllerPoolPersistentStorage(testStorage.getNamespaceName(), testStorage.getTargetControllerPoolName(), testStorage.getTargetClusterName(), 1).build()
             )
         );
-        resourceManager.createResourceWithWait(KafkaTemplates.kafkaPersistent(testStorage.getTargetClusterName(), 1, 3).build());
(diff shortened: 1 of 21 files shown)


henryZrncik:
@strimzi-ci run tests --profile=all --env=STRIMZI_FEATURE_GATES=-KafkaNodePools,-UnidirectionalTopicOperator,-UseKRaft

henryZrncik:
@see-quick yeah good suggestion,  I have tried to avoid this solution in test case where more than 1 kafka is created, but still i wanted to create this situation at least in 2,3 tests to see if it it would really cause some resources problem as i did not hit this problem locally or in past.

systemtest/src/test/java/io/strimzi/systemtest/mirrormaker/MirrorMaker2ST.java
```diff
@@ -632,8 +632,8 @@ void testIdentityReplicationPolicy() {
         );
 
         resourceManager.createResourceWithWait(
-            KafkaTemplates.kafkaEphemeral(testStorage.getSourceClusterName(), 1, 1).build(),
-            KafkaTemplates.kafkaEphemeral(testStorage.getTargetClusterName(), 1, 1).build(),
+            KafkaTemplates.kafkaPersistent(testStorage.getSourceClusterName(), 1, 1).build(),
+            KafkaTemplates.kafkaPersistent(testStorage.getTargetClusterName(), 1, 1).build(),
```
im-konge, line 636: If you are changing the storage from ephemeral to persistent here, should you update the corresponding NodePools as well? (I guess this applies to most of the places)

systemtest/src/test/java/io/strimzi/systemtest/operators/user/UserST.java
```diff
@@ -360,7 +360,7 @@ void testTlsExternalUser() {
                 KafkaNodePoolTemplates.controllerPool(testStorage.getNamespaceName(), testStorage.getControllerPoolName(), testStorage.getClusterName(), 1).build()
             )
         );
-        resourceManager.createResourceWithWait(KafkaTemplates.kafkaEphemeral(testStorage.getClusterName(), 1, 1)
+        resourceManager.createResourceWithWait(KafkaTemplates.kafkaPersistent(testStorage.getClusterName(), 1, 1)
```
im-konge, line 363: I guess that on some places the change between ephemeral and persistent storage doesn't make much sense (at least from MPOV).
It's not about having some problem with resources or something
I just thing that for the tests where no rolling update is happening or we are not depended on the storage, we can use the ephemeral storage

Or maybe I'm wrong 🤷

scholzj: In general, if you don't do rolling updates the likely hood of something going wrong with ephemeral storage is of course much smaller. It might still happen if something restarts due to some issue etc.

systemtest/src/test/java/io/strimzi/systemtest/mirrormaker/MirrorMaker2ST.java
henryZrncik: sure after the discussion this can be done in single PR.

henryZrncik: thanks for bringing this up  👍

systemtest/src/test/java/io/strimzi/systemtest/operators/user/UserST.java
henryZrncik: yeah you both have a point here, 
So as was agreed if there is for some reason (too many resources) only 1 replica , storage is changed to persistent, otherwise I am keeping it as it is so far (I am removing some of increment of replicas due to actually hitting one resource limitation).

henryZrncik:
@strimzi-ci run tests --cluster-type=ocp --cluster-version=4.14 --install-type=bundle --profile=all --testcase=HttpBridgeST,MirrorMaker2ST,RackAwarenessST --parallel-count=1 --env=STRIMZI_FEATURE_GATES=+UnidirectionalTopicOperator;STRIMZI_USE_KRAFT_IN_TESTS=true

systemtest/src/test/java/io/strimzi/systemtest/security/oauth/OauthAuthorizationST.java
```diff
@@ -671,11 +671,11 @@ void setUp()  {
 
         resourceManager.createResourceWithWait(
             NodePoolsConverter.convertNodePoolsIfNeeded(
-                KafkaNodePoolTemplates.brokerPool(Environment.TEST_SUITE_NAMESPACE, KafkaNodePoolResource.getBrokerPoolName(oauthClusterName), oauthClusterName, 1).build(),
-                KafkaNodePoolTemplates.controllerPool(Environment.TEST_SUITE_NAMESPACE, KafkaNodePoolResource.getControllerPoolName(oauthClusterName), oauthClusterName, 1).build()
+                KafkaNodePoolTemplates.brokerPoolPersistentStorage(Environment.TEST_SUITE_NAMESPACE, KafkaNodePoolResource.getBrokerPoolName(oauthClusterName), oauthClusterName, 1).build(),
```
im-konge, line 674: You are changing this, but you should increase the number of replicas to 3/3 just for the rolling update, or do you think that the persistent storage will be sufficient? That's for the `testSuperUserWithOauthAuthorization`

im-konge (approved):
LGTM, thanks

see-quick (approved):

PR #9801, opened 2024-03-08
[ST] Fixups for few failing tests in nightlies

im-konge:
### Type of change

- Bugfix

### Description

This PR fixes few failing tests in our nightlies, mainly for OCP:

- tests that are using `route` as a listener types are sometimes failing on the length of the hostname - that is hit because of long name of the NodePools. I changed it from `broker-`, `control-`, and `mixed-` prefixes to `b-`, `c-`, `m-` prefixes. In case that it will not be that readable, we can try different prefix and possibly shorter the suffix as well.
- changed `my-cluster-` prefix to `cluster-`
- added `ROUTE` tag to one test missing this tag
- created `SkipDefaultNetworkPolicyCreation` for tests that need skip of creation of the default NetworkPolicies -> for example test `testNPGenerationEnvironmentVariable` turns off creation of NetworkPolicies in CO -> and because the "default to deny" global network policy we would not be able to create the Kafka Pods, as it seems that the ZooKeeper throws "connection refused", but in the end it's just the default NetworkPolicies that blocks the connection requests to ZK
- added missing NP creation to HTTP CORS tests and `MultipleClusterOperatorsST`

### Checklist

### DIFF
--- a/systemtest/src/main/java/io/strimzi/systemtest/TestConstants.java
+++ b/systemtest/src/main/java/io/strimzi/systemtest/TestConstants.java
@@ -533,7 +533,7 @@ public interface TestConstants {
     /**
      * NodePool's name prefix based on role
      */
-    String MIXED_ROLE_PREFIX = "mixed-";
-    String BROKER_ROLE_PREFIX = "broker-";
-    String CONTROLLER_ROLE_PREFIX = "control-";
+    String MIXED_ROLE_PREFIX = "m-";
+    String BROKER_ROLE_PREFIX = "b-";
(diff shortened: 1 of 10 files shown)


im-konge:
@strimzi-ci run tests --cluster-type=ocp --cluster-version=4.14 --install-type=bundle --profile=all --groups=route,bridge,networkpolicies

im-konge:
@strimzi-ci run tests --cluster-type=ocp --cluster-version=4.14 --install-type=bundle --profile=all --groups=route,bridge,networkpolicies

Frawless (approved):

PR #9809, opened 2024-03-11
[DOC] Suggest removal of unused fields from Kafka CR after migration to node pools

scholzj:
### Type of change

- Documentation

### Description

As the `.spec.kafka.replicas` and `spec.kafka.storage` fields are not required anymore int he `Kafka` CR, we should instruct users to remove them from the `Kafka` CR when migrating to node pools.

### Checklist

- [x] Update documentation

### DIFF
--- a/documentation/modules/configuring/proc-migrating-clusters-node-pools.adoc
+++ b/documentation/modules/configuring/proc-migrating-clusters-node-pools.adoc
@@ -91,5 +91,5 @@ kubectl apply -f <kafka_configuration_file>
 There is no change or rolling update.
 The resources remain identical to how they were before.
 
-. Once the `KafkaNodePool` resource is in use, you can remove the old fields from the `Kafka` custom resource.
-This includes all the fields that you copied to the `KafkaNodePool` resource in the previous steps such as the `.spec.kafka.replicas` or `.spec.kafka.storage`.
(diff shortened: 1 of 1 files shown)


ppatierno (approved):

documentation/modules/configuring/proc-migrating-clusters-node-pools.adoc
```diff
@@ -89,4 +89,7 @@ kubectl apply -f <kafka_configuration_file>
 ----
 +
 There is no change or rolling update.
-The resources remain identical to how they were before.
\ No newline at end of file
+The resources remain identical to how they were before.
+
+. Once the `KafkaNodePool` resource is in use, you can remove the old fields from the `Kafka` custom resource.
+This includes all the fields that you copied to the `KafkaNodePool` resource in the previous steps such as the `.spec.kafka.replicas` or `.spec.kafka.storage`.
```
PaulRMellor, line 95:
```suggestion
. Remove the replicated properties from the `Kafka` custom resource.
When the `KafkaNodePool` resource is in use, you can remove the properties that you copied to the `KafkaNodePool` resource, such as the `.spec.kafka.replicas` and `.spec.kafka.storage` properties.
```

PaulRMellor (approved):
suggested to make the step a bit more active

fvaleri (approved):

Issue #9734, opened 2024-02-22
Use proper `toString()` implementation in the API classes


scholzj:
The classes in our API module currently seem to mix several ways how to handle the `toString` methods:
* Some use Jackson libraries to render a YAML string
* Some do not do anything
* Some use the Lombok `@ToString` annotation to generate the `toString()` method

We should pick one way (Lombok annotation?) and use it properly in all resources and their subclasses. This would help when debugging issues, with log readability etc.

scholzj:
Triaged on the community call on 7.3.2024: We should use the `@ToString` annotation on all the API classes.

kockas:
I'd  do it. I see quite a lot of inconsistencies also in `@EqualsAndHashCode` when `callSuper = true`  is not used consistently. I would align it, and use `@ToString` inline with `@EqualsAndHashCode`, so both annotation use same `callSuper = true` (or false).
Is that approach ok for you?

scholzj:
I think that makes sense, yes.

PR #9810, opened 2024-03-11
#9734 - added toString to whole model, including Password, PasswordSe…

kockas:
…cretSource and PasswordSource classes

### Type of change

- Bugfix

### Description

Implementation of the issue #9734
All classes in the model package and sub-packages now contain toString generated by Lombok's `@ToString`.
Additionally:
 - each class contains generate hashCode and equals using `@EqualsAndHashCode`
 - when class has a Parent, both `@EqualsAndHashCode` and `@ToString`  are using `callSuper=true`
 - existing toString methods were removed
 - toString is also within classes `Password`, `PasswordSecretSource`, `PasswordSource` as they in fact do not contain secrets
 - class `KubernetesVersion` is left intact with manual hashCode, equals, and toString

### DIFF
--- a/api/pom.xml
+++ b/api/pom.xml
@@ -65,10 +65,6 @@
         </dependency>
         <dependency>
             <groupId>com.fasterxml.jackson.core</groupId>
-            <artifactId>jackson-core</artifactId>
-        </dependency>
-        <dependency>
-            <groupId>com.fasterxml.jackson.core</groupId>
             <artifactId>jackson-annotations</artifactId>
         </dependency>
         <dependency>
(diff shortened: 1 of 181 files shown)


scholzj (commented):
It looks like the scope of the dependnecies changed:
```
[INFO] --- maven-dependency-plugin:3.3.0:analyze-only (analyze) @ api ---
[WARNING] Unused declared dependencies found:
[WARNING]    com.fasterxml.jackson.core:jackson-core:jar:2.16.1:compile
[WARNING] Non-test scoped test only dependencies found:
[WARNING]    com.fasterxml.jackson.dataformat:jackson-dataformat-yaml:jar:2.16.1:compile
```

Can you please update the `api/pom.xml` accordingly?

ppatierno (approved):
Thanks for the PR! LGTM and it seems regressions are all green.

scholzj (approved):
LGTM. Thanks for the PR.

PR #9813, opened 2024-03-11
[ST] Fix and improve QuotasST

### LINKED ISSUE #9754: Test `testKafkaQuotasPluginIntegration` uses wrong Kafka pod name when getting log
The `QuotaST.testKafkaQuotasPluginIntegration` fails with an error like the following:

```shell
2024-02-29 03:52:09 [main] ERROR [TestExecutionWatcher:27] QuotasST - Exception Failure executing: GET at: https://192.168.49.2:8443/api/v1/namespaces/namespace-18/pods/my-cluster-49e5760c-kafka-0/log?pretty=false. Message: pods "my-cluster-49e5760c-kafka-0" not found. Received status: Status(apiVersion=v1, code=404, details=StatusDetails(causes=[], group=null, kind=pods, name=my-cluster-49e5760c-kafka-0, retryAfterSeconds=null, uid=null, additionalProperties={}), kind=Status, message=pods "my-cluster-49e5760c-kafka-0" not found, metadata=ListMeta(_continue=null, remainingItemCount=null, resourceVersion=null, selfLink=null, additionalProperties={}), reason=NotFound, status=Failure, additionalProperties={}). has been thrown in @Test. Going to collect logs from components.
```

This is because at the beginning of the test it's going to use node pools to deploy the Kafka cluster and the brokers node pool is named `broker` but when it comes to get the Kafka log, the pod name for Kafka has a `-kafka-0` suffix while the right one is `-broker-0`.

im-konge:
### Type of change

- Bugfix

### Description

This PR fixes #9754 
with correct broker Pod name. But during the work on the fix I found out that the `QuotasST` doesn't work at all:

- the test-clients were running out of memory on local clusters
- the ST doesn't work on clusters with local storage (e.g. `minikube`), which makes sense in terms of the calculation of the used storage
- the check for `disk is beyond soft limit` is flaky, as we can hit the soft limit really rarely, based on how the messages are send and how much free (or filled) storage we have during the test execution
- we are not running the test in any pipeline and it doesn't contain any tag for easier execution
- there is no note about "how we can actually run the test"

This PR should fix all of the above, it adds new tag `quotasplugin` for easier execution of the ST class, adds new method `isMinikube`, so it will be skipped on `minikube` clusters, and more.

Fixes #9754

### Checklist

### DIFF
--- a/systemtest/src/main/java/io/strimzi/systemtest/TestConstants.java
+++ b/systemtest/src/main/java/io/strimzi/systemtest/TestConstants.java
@@ -399,11 +399,15 @@ public interface TestConstants {
      */
     String ROUTE = "route";
 
-
     /**
      * Tag for tests that focus on migration from ZK to KRaft
      */
     String MIGRATION = "migration";
+
+    /**
+     * Tag for tests that uses Strimzi quotas plugin
+     */
+    String QUOTAS_PLUGIN = "quotasplugin";
 
     /**
      * Tag for tests, without ARM,AARCH64 support
--- a/systemtest/src/test/java/io/strimzi/systemtest/kafka/QuotasST.java
+++ b/systemtest/src/test/java/io/strimzi/systemtest/kafka/QuotasST.java
@@ -27,10 +27,19 @@ import org.junit.jupiter.api.Tag;
 import java.util.Collections;
 
 import static io.strimzi.systemtest.TestConstants.INTERNAL_CLIENTS_USED;
+import static io.strimzi.systemtest.TestConstants.QUOTAS_PLUGIN;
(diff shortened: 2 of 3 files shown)


im-konge:
@strimzi-ci run tests --cluster-type=ocp --cluster-version=4.14 --install-type=bundle --profile=all --groups=quotasplugin

jankalinic (approved):
Looks good to me

henryZrncik:
LGTM, just about execution  is `QUOTAS_PLUGIN` to be executed as part of something else ?

see-quick (approved):
👍

jankalinic (approved):

henryZrncik (approved):

PR #9815, opened 2024-03-12
External ip support

rao2100:
### Type of change

- Enhancement 

### Description

Cluster operator does not provide any configurability of the external IPs associated with the nodeport. In other words, user is not able to provide any external IP. We have tried to perform a kubectl patch with the external IP but the setting is not permanent. After a while, the change will be reverted and external IP is gone. To address this, we updated the CRD template with a new option where user can provide external IP for the broker and IP for each of the nodeports. The cluster operator is also updated to retrieve the user provided values and provision them during the init container step.

NodePort service without specifying an ExternalIP would suffice, however, there are certain situations where specifying an ExternalIP might be necessary or beneficial.

For example when you’re running a bare-metal Kubernetes cluster without a load balancer like MetalLB, e.g. when using Calico that is able to talk BGP like MetalLB but with additional features that MetalLB does not provide, Calico however does not support LoadBalancer service types and as a result we use NodePort services.

While NodePort services can often be used without specifying an ExternalIP, there are scenarios where using an ExternalIP can provide additional flexibility or meet specific requirements.

**Sample Test**

- This test case is to test the strimzi operator customization that allows setting of external IP for Nodeport

**Sample Configuration**

```yaml 
kafka:
  replicas: 1
  config:
  listeners:
  - name: plain
    port: 9092
    type: internal
    tls: false 
  - name: external
    port: 9094
    type: nodeport
    tls: false
    configuration:
      preferredNodePortAddressType: ExternalIP
      bootstrap:        
        nodePort: 32100
        externalIPs:
        - 10.15.32.119  
      brokers:         
      - broker: 0        
        nodePort: 32000
        advertisedHost: 10.15.32.119
        externalIPs:
          - 10.15.32.119
  template:  
    externalTrafficPolicy: Local   
        
zookeeper:
  replicas: 1
```

**Outcome**

-  After deploying all the necessary charts and kafka with the above yaml, verify that the EXTERNAL-IP is set.

```
    $ kubectl get svc

NAME                                   TYPE        CLUSTER-IP       EXTERNAL-IP    PORT(S)                               AGE
odf-cluster-kafka-0                    NodePort    10.102.146.11    10.15.32.119   9094:32000/TCP                        21s
odf-cluster-kafka-bootstrap            ClusterIP   10.106.19.186    <none>         9091/TCP,9092/TCP                     12m
odf-cluster-kafka-brokers              ClusterIP   None             <none>         9090/TCP,9091/TCP,8443/TCP,9092/TCP   12m
odf-cluster-kafka-exporter-metrics     ClusterIP   10.96.170.230    <none>         9404/TCP                              4d22h
odf-cluster-kafka-external-bootstrap   NodePort    10.99.254.0      10.15.32.119   9094:32100/TCP  

```


cluster-operator/src/main/java/io/strimzi/operator/cluster/model/ListenersUtils.java
```diff
@@ -699,4 +699,50 @@ public static String advertisedPortFromOverrideOrParameter(GenericKafkaListener
...
+		                    .map(GenericKafkaListenerConfigurationBroker::getExternalIPs)
+		                    .findAny()
+		                    .orElse(null);
+		        } else {
+		            return null;
+		        }
+		 }else {
+			 throw new UnsupportedOperationException("Broker externalIP are not used with listerner type "+ listener.getType());
+		 }
+    }
```
scholzj, line 747: You should mirror what the other methods do => just return here and do validation in the validation class (`ListenersValidator` or something like that).

scholzj (changes requested):
Thanks for the PR. I think this looks mostly fine. But you should update the listener validation to validate that it is set only with node port listeners. You also need to properly update the CRDs in all locations for the CI to pass.

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/ListenersValidator.java
```diff
@@ -345,6 +347,19 @@ private static void validateBootstrapNodePort(Set<String> errors, GenericKafkaLi
...
         }
     }
+    
+    /**
+     * Validates that bootstrap.nodePort is used only with NodePort type listener
+     *
+     * @param errors    List where any found errors will be added
+     * @param listener  Listener which needs to be validated
+     */
+    private static void validateBootstrapExternalIPs(Set<String> errors, GenericKafkaListener listener) {
```
scholzj, line 357: You should write unit tests (or add it to one of the existing unit tests I guess) for this method as well as the other one you added.

cluster-operator/src/test/java/io/strimzi/operator/cluster/model/ListenersValidatorTest.java
```diff
@@ -1126,4 +1127,49 @@ public void testValidateCreateBootstrapService() {
                 "listener listener2 cannot configure createBootstrapService because it is not load balancer listener",
                 "listener listener3 cannot configure createBootstrapService because it is not load balancer listener"));
     }
+    
+    @ParallelTest
+    public void testValidateBootstrapExternalIPsListenener() {
```
scholzj, line 1132: Yu should probbaly add a third listener with node prot type and see that it doesn't raise any errors.

scholzj (commented):
Looks mostly good. Thanks. I laft one more comment. You should also add a note about this to the `CHANGELOG.md` file.

api/src/main/java/io/strimzi/api/kafka/model/kafka/listener/GenericKafkaListenerConfigurationBootstrap.java
```diff
@@ -114,6 +115,19 @@ public void setLoadBalancerIP(String loadBalancerIP) {
         this.loadBalancerIP = loadBalancerIP;
     }
 
+    @Description("External IPs associated to the nodeport service. " + 
+            "These IPs are used by clients external to the cluster to access the kafka brokers. " +
```
ppatierno, line 119: I guess "cluster" is referring to Kubernetes cluster not Kafka one. It's better to clarify.

api/src/main/java/io/strimzi/api/kafka/model/kafka/listener/GenericKafkaListenerConfigurationBroker.java
```diff
@@ -138,6 +140,19 @@ public void setLoadBalancerIP(String loadBalancerIP) {
         this.loadBalancerIP = loadBalancerIP;
     }
 
+    @Description("External IPs associated to the nodeport service. " + 
+            "These IPs are used by clients external to the cluster to access the kafka brokers. " +
```
ppatierno, line 144: Ditto as above about "cluster"

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaCluster.java
```diff
@@ -713,6 +713,13 @@ public List<Service> generateExternalBootstrapServices() {
                 }
             }
 
+            if (KafkaListenerType.NODEPORT == listener.getType()) {
+                List<String> eips = ListenersUtils.bootstrapExternalIPs(listener);
```
ppatierno, line 717: I would go for a more exhaustive variable name. Something like `externalIps` instead of just `eips`.

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaCluster.java
```diff
@@ -791,6 +798,13 @@ public List<Service> generatePerPodServices() {
                             }
                         }
 
+                        if (KafkaListenerType.NODEPORT == listener.getType()) {
+                            List<String> eips = ListenersUtils.brokerExternalIPs(listener, node.nodeId());
```
ppatierno, line 802: Ditto as above.

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/ListenersUtils.java
```diff
@@ -699,4 +699,34 @@ public static String advertisedPortFromOverrideOrParameter(GenericKafkaListener
...
+
+     /**
+     * Returns broker service external IPs
+     * 
+     * @param listener  Listener for which the external IPs should be found
+     * @param pod       Pod ID for which we should get the configuration option
+     * 
+     * @return          External IPs or null if not specified
+     */
+    public static List<String> brokerExternalIPs(GenericKafkaListener listener, int pod) {
```
ppatierno, line 724: `pod` is effectively a node ID ... so I would use `nodeId` as parameter name

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 * Added support for topic replication factor change to the Unidirectional Topic Operator when Cruise Control integration is enabled.
 * The `KafkaNodePools` feature gate moves to GA stage and is permanently enabled without the possibility to disable it.
   To use the Kafka Node Pool resources, you still need to use the `strimzi.io/node-pools: enabled` annotation on the `Kafka` custom resources.
+* Added support for External IP Configurability for Kafka brokers and External Bootstrap.
```
scholzj, line 8: Maybe something like this better describes what exactly does it do?

```suggestion
* Added support for configuring the `externalIPs` field in node port type services
```

api/src/main/java/io/strimzi/api/kafka/model/kafka/listener/GenericKafkaListenerConfigurationBootstrap.java
```diff
@@ -114,6 +115,19 @@ public void setLoadBalancerIP(String loadBalancerIP) {
         this.loadBalancerIP = loadBalancerIP;
     }
 
+    @Description("External IPs associated to the nodeport service. " + 
+            "These IPs are used by clients external to the kubernetes cluster to access the kafka brokers. " +
```
ppatierno, line 119: Sorry again but we should use "Kubernetes" with the capital "k".
The same applies to all the others places where you have this `@Description` annotation.

scholzj: As well as Kafka with capital `K`.

ppatierno (approved):
LGTM! Thanks!

scholzj (approved):
Thanks for the PR.

Issue #9537, opened 2024-01-10
[Bug]: Kafka Exporter Grafana dashboard too long URL error


jesusmah:
### Bug Description

When there are few topics and few consumer groups, the Kafka Exporter Grafana Dashboard does not work because the API call being executed behind the panels is way too long:

![image](https://github.com/strimzi/strimzi-kafka-operator/assets/4761922/4291808f-b351-408d-87ab-aa590905760e)

The reason for this is that if you select the option "All" for consumer groups and topics it does include all of those as parameters in the URL:

![image](https://github.com/strimzi/strimzi-kafka-operator/assets/4761922/bea43ca6-f1d2-43b6-81be-d2d2792b72c4)

Instead, the regular expression behind those panels should be something like ".*" so that it does not explicitly write all consumer groups and topics as parameters in the URL.

The workaround I managed to put together was to duplicate the dashboard and remove consumer group and topic as variables for the dashboard as well as removed these from the reg expressions behind the panels. The absence of those parameters means that the API call behind will retrieve the data for all consumer groups and topics. However, this implies having a duplicate dashboard for showing all consumer groups and topics.

### Steps to reproduce

1. Import the Kafka Exporter Grafana for a strimzi cluster that contains around 30 topics and 30 consumer groups.
2. Select "All" for consumer groups and topics variables at the top of the dashboard (I think this is the default)
3. See the errors in the dashboard.

### Expected behavior

I would expect the dashboard to work fine and display the data it is meant to regardless of the number of consumer groups and topics just like the dashboard I managed to create as a workaround.

### Strimzi version

0.36.1

### Kubernetes version

OpenShift 4.12.18

### Installation method

Operator

### Infrastructure

VMware

scholzj:
Triaged on a community call on 1.11.2024: This would need to be investigated more and we would need to try to find a better solution than having two dashboards.

maciej-tatarski:
I can work on it if help is still needed

daniel-orlov:
If you need help with Grafana, @maciej-tatarski is the man 💯

maciej-tatarski:
I checked it with grafana `v10.3.3` and prometheus `v2.50.1` with 1000+ topics and issue doesn't seem to appear:
![image](https://github.com/strimzi/strimzi-kafka-operator/assets/135701313/d16d6c1b-ee40-424c-86fd-97fa8c2095eb)

[...]

PR #9817, opened 2024-03-12
Fix value for all topics and groups selected in `Strimzi Kafka Exporter` dashboard

maciej-tatarski:
### Type of change

- Bugfix

### Description

Related to: https://github.com/strimzi/strimzi-kafka-operator/issues/9537

With big number of topics and consumer groups, the Kafka Exporter Grafana Dashboard does not work because the API call being executed behind the panels is too long. Instead of including all topics name in api call, we'll now substitute `all` value with `.*` regex which will have the same result.

image

Tested with grafana `7.5.7`, `v10.3.3`

### Checklist

- [ ] Write tests
- [ ] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/examples/metrics/grafana-dashboards/strimzi-kafka-exporter.json
+++ b/examples/metrics/grafana-dashboards/strimzi-kafka-exporter.json
@@ -1556,7 +1556,7 @@
         "useTags": false
       },
       {
-        "allValue": null,
+        "allValue": ".*",
         "current": {
           "text": "All",
           "value": [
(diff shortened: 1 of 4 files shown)


scholzj (commented):
Thanks for the PR. Can you please do the changes only in the `/packaging/...` folders? The `/install` and `/examples` folders are updated only as part of a release.

maciej-tatarski:
I reverted the changes for releases files

scholzj (approved):
LGTM. Thanks.

ppatierno (approved):

Issue #8732, opened 2023-06-26
Enhance `KafkaBridge` resource with consumer inactivity timeout and HTTP consumer/producer parts enablement


ppatierno:
The current `KafkaBridge` custom resource has an `http` section for setting HTTP related configuration.
Right now it supports only `host` and `port` but the bridge also has:

* http.timeoutSeconds : for deleting inactive consumers after a timeout (disabled by default)
* http.consumer.enabled: to enable/disable the HTTP consumer part (enabled by default)
* http.producer.enabled: to enable/disable the HTTP producer part (enabled by default)

We should allow the above parameters to be configured via the `KafkaBridge` spec.

scholzj:
Triaged on 29.6.2023: We should enable the configuration of these fields. But how should the API look like? Should have a proposal to clarify the API changes.

steffen-karlsson:
If you'd like, I can try to look into this one @ppatierno and @scholzj :)

One comment, you @ppatierno write that `host` is currently supported, but when I look into the source code, it looks like its defaulting to `HTTP_DEFAULT_HOST`: https://github.com/strimzi/strimzi-kafka-operator/blob/main/cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaBridgeCluster.java#L408, also `host` is not defined in the CRD: https://github.com/strimzi/strimzi-kafka-operator/blob/main/install/cluster-operator/046-Crd-kafkabridge.yaml#L223

Just want to make sure that its intentional and no work is needed on the `http.host` property.

Thanks :)

steffen-karlsson:
Hello @ppatierno 

I've now added `host` and `timeoutSeconds` as options to the CRD.
Furthermore, I've decided to add `consumer` and `producer` object with an `enabled` boolean inside the `http` object.

Please see reference here for new CRD:
```
http:
  type: object
  properties:
    host:
      type: string
      description: The host which is the server listening on.
    port:
      type: integer
      minimum: 1023
      description: The port which is the server listening on.
    cors:
      type: object
      properties:
        allowedOrigins:
          type: array
          items:
            type: string
          description: List of allowed origins. Java regular expressions can be used.
        allowedMethods:
          type: array
          items:
            type: string
          description: List of allowed HTTP methods.
      required:
      - allowedOrigins
      - allowedMethods
      description: CORS configuration for the HTTP Bridge.
    timeoutSeconds:
      type: integer
[...]

PR #9820, opened 2024-03-12
#8732 Enhance KafkaBridge resource with consumer inactivity timeout and HTTP consumer/producer parts enablement

steffen-karlsson:
### Type of change
- Enhancement / new feature: https://github.com/strimzi/strimzi-kafka-operator/issues/8732

### Description

Adding CRD Kubernetes API support for following `KafkaBridge` properties:
**http.consumer.timeoutSeconds** - for deleting inactive consumers after a timeout (disabled by default)
**http.consumer.enabled** - to enable/disable the HTTP consumer part (enabled by default)
**http.producer.enabled** - to enable/disable the HTTP producer part (enabled by default)

### Checklist

- [X] Write tests
- [X] Update documentation
- [X] Update CHANGELOG.md


steffen-karlsson:
Missing last part of documentation, will do when discussion in issue is settled: https://github.com/strimzi/strimzi-kafka-operator/issues/8732

api/src/main/java/io/strimzi/api/kafka/model/bridge/KafkaBridgeHttpConfig.java
```diff
@@ -49,8 +55,18 @@ public KafkaBridgeHttpConfig(int port) {
...
 
+    @Description("The host which is the server listening on.")
+    @JsonProperty(defaultValue = HTTP_DEFAULT_HOST)
+    public String getHost() {
+        return host;
+    }
+
+    public void setHost(String host) {
+        this.host = host;
+    }
```
scholzj, line 66: What would be the point of the host configuration? We are here only in Kubernetes context, so I'm not sure what would you use it for.

api/src/main/java/io/strimzi/api/kafka/model/bridge/KafkaBridgeHttpConfig.java
```diff
@@ -38,8 +38,14 @@ public class KafkaBridgeHttpConfig implements UnknownPropertyPreserving, Seriali
...
     public static final int HTTP_DEFAULT_PORT = 8080;
     public static final String HTTP_DEFAULT_HOST = "0.0.0.0";
+    public static final long HTTP_DEFAULT_TIMEOUT = -1L;
+
+    private String host = HTTP_DEFAULT_HOST;
     private int port = HTTP_DEFAULT_PORT;
     private KafkaBridgeHttpCors cors;
+    private long timeoutSeconds = HTTP_DEFAULT_TIMEOUT;
+    private KafkaBridgeProducerEnablement producer = new KafkaBridgeProducerEnablement();
+    private KafkaBridgeConsumerEnablement consumer = new KafkaBridgeConsumerEnablement();
```
scholzj, line 48: I think this needs a bit more discussion and thinking ... if we create this as a new object, then it will be because we expect we might need additional fields there. So it maybe should be called `KafkaBridgeProducerConfig` / `KafkaBridgeConsumerConfig`? But wouldn't it make sense to somehow try to incorporate it into `the .spec.consumer` / `.spec.producer` sections?

api/src/main/java/io/strimzi/api/kafka/model/common/ClientTls.java
```diff
@@ -56,4 +58,11 @@ public void setAdditionalProperty(String name, Object value) {
         }
         this.additionalProperties.put(name, value);
     }
+
+    public String toVarString() {
+        Objects.requireNonNull(this.trustedCertificates, "trustedCertificates cannot be null");
+        return this.trustedCertificates.stream()
+                .map(certSecretSource -> certSecretSource.getSecretName() + "/" + certSecretSource.getCertificate())
+                .collect(Collectors.joining(";"));
+    }
```
scholzj, line 67: I don't really follow why this is added here. It does not seem to belong into the `api` I think.

scholzj (commented):
I left some comments. I the CI failed because of spotbugs. It would also fail because the CRD files are not copied to all the places. You can use `make crd_install` for that.

api/src/main/java/io/strimzi/api/kafka/model/bridge/KafkaBridgeHttpConfig.java
steffen-karlsson: Reason is that @ppatierno was mentioning that `host` is currently supported, but I guess it was not as it was not accessible though Kubernetes API: https://github.com/strimzi/strimzi-kafka-operator/issues/8732#issue-1774631381 also wrote some comments and concerns in the issue.

api/src/main/java/io/strimzi/api/kafka/model/bridge/KafkaBridgeHttpConfig.java
ppatierno: That's the point I was going to make as well.
My view, is that what's under `http` is related to HTTP. So when I refer to something like `http.producer.enabled` and `http.consumer.enabled` I am referring to the HTTP part of the corresponding Kafka producer and consumer. The current status is that I can just enable/disable the usage of consumer or producer on the HTTP side, which means being or not being able to use HTTP to producer and/or consume (so by using the corresponding Kafka part).
Of course, enabling/disabling the HTTP part of it would mean enabling/disabling the corresponding Kafka part which comes to the @scholzj suggestion to use something like `.spec.consumer.enabled` and `.spec.producer.enabled`. That part was mostly there for Kafka related configuration but doesn't mean we can't change it. I agree it seems too much having a dedicated class unless we envisage additional HTTP producer/consumer related configuration which I can't think about now.
Anyway, I would say it needs further discussion in terms of API. I think we also said it would need a proposal first where we can make a decision all together. https://github.com/strimzi/strimzi-kafka-operator/issues/8732#issuecomment-1612639586

@steffen-karlsson would you mind to open a PR with a proposal here https://github.com/strimzi/proposals which could mimic what you have here and starting the discussion on it?

steffen-karlsson: Sure, I raised this to @ppatierno in https://github.com/strimzi/strimzi-kafka-operator/issues/8732 for discussion, if it should be separate object or inside the `spec.consumer` and `spec.producer`

api/src/main/java/io/strimzi/api/kafka/model/common/ClientTls.java
steffen-karlsson: It was to generalise the code and reuse across `Bridge` and `Connector` since a lot of code was duplicated, so I refactored it and wanted to put it elsewhere than in one of those two, any suggestion?

api/src/main/java/io/strimzi/api/kafka/model/bridge/KafkaBridgeHttpConfig.java
ppatierno: I would agree. That was the reason it was taken out.

api/src/main/java/io/strimzi/api/kafka/model/bridge/KafkaBridgeHttpConfig.java
ppatierno: It seems we are having the same discussion on different places :-D
I think a new proposal will be the only definitive one.

api/src/main/java/io/strimzi/api/kafka/model/bridge/KafkaBridgeHttpConfig.java
ppatierno: Maybe I was writing my comment while Steffen was writing his comment :-)
I meant to agree with Jakub, so we should taken it out.

steffen-karlsson: Will revert those changes, no problem :)

api/src/main/java/io/strimzi/api/kafka/model/bridge/KafkaBridgeHttpConfig.java
steffen-karlsson: Sure will fix the other comment from @scholzj and then raise a proposal for further discussion :)

api/src/main/java/io/strimzi/api/kafka/model/common/ClientTls.java
steffen-karlsson: Moved it to cluster-operator to `ModelUtils.java`

api/src/main/java/io/strimzi/api/kafka/model/bridge/KafkaBridgeHttpConfig.java
steffen-karlsson: Hello @scholzj and @ppatierno 

I've raised a proposal with the two implementation options here: https://github.com/strimzi/proposals/pull/111

api/src/main/java/io/strimzi/api/kafka/model/bridge/KafkaBridgeConsumerSpec.java
```diff
@@ -21,16 +21,39 @@
...
+    private boolean enabled = true;
+    private long timeoutSeconds = HTTP_DEFAULT_TIMEOUT;
+
     @Override
     @Description("The Kafka consumer configuration used for consumer instances created by the bridge. Properties with the following prefixes cannot be set: " + FORBIDDEN_PREFIXES + " (with the exception of: " + FORBIDDEN_PREFIX_EXCEPTIONS + ").")
     public Map<String, Object> getConfig() {
         return config;
     }
+
+    @Description("Whether the HTTP consumer should be enabled or disabled")
```
ppatierno, line 42: Let's add the information that it's enabled by default.

api/src/main/java/io/strimzi/api/kafka/model/bridge/KafkaBridgeConsumerSpec.java
```diff
@@ -21,16 +21,39 @@
...
+    public boolean isEnabled() {
+        return enabled;
+    }
+
+    public void setEnabled(boolean enabled) {
+        this.enabled = enabled;
+    }
+
+    @Description("The timeout in seconds for deleting inactive consumers.")
```
ppatierno, line 51: Let's add the information that there is no timeout by default.

api/src/main/java/io/strimzi/api/kafka/model/bridge/KafkaBridgeProducerSpec.java
```diff
@@ -21,16 +21,27 @@
...
 
+    private boolean enabled = true;
+
     @Override
     @Description("The Kafka producer configuration used for producer instances created by the bridge. Properties with the following prefixes cannot be set: " + FORBIDDEN_PREFIXES + " (with the exception of: " + FORBIDDEN_PREFIX_EXCEPTIONS + ").")
     public Map<String, Object> getConfig() {
         return config;
     }
+
+    @Description("Whether the HTTP producer should be enabled or disabled")
```
ppatierno, line 39: Ditto as above for the consumer. Saying it's enabled by default.

api/src/main/java/io/strimzi/api/kafka/model/bridge/KafkaBridgeSpec.java
```diff
@@ -50,8 +50,8 @@ public class KafkaBridgeSpec extends Spec implements HasConfigurableLogging, Has
     private String bootstrapServers;
     private ClientTls tls;
     private KafkaClientAuthentication authentication;
-    private KafkaBridgeConsumerSpec consumer;
-    private KafkaBridgeProducerSpec producer;
+    private KafkaBridgeConsumerSpec consumer = new KafkaBridgeConsumerSpec();
+    private KafkaBridgeProducerSpec producer = new KafkaBridgeProducerSpec();
```
ppatierno, line 54: why the need for this creation here?

cluster-operator/src/test/java/io/strimzi/operator/cluster/model/KafkaBridgeClusterTest.java
```diff
@@ -1300,4 +1309,22 @@ public void testJvmOptions() {
         assertThat(javaOpts.getValue(), containsString("-Xmx256m"));
         assertThat(javaOpts.getValue(), containsString("-XX:InitiatingHeapOccupancyPercent=36"));
     }
+
+    @ParallelTest
+    public void testConsumerProducerOptions() {
+        KafkaBridge resource = new KafkaBridgeBuilder(this.resource)
+                .editSpec()
+                .withConsumer(new KafkaBridgeConsumerSpecBuilder().withTimeoutSeconds(60).build())
+                .withProducer(new KafkaBridgeProducerSpecBuilder().build())
```
ppatierno, line 1318: indentation

CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 ## 0.41.0
 
+* Enhance `KafkaBridge` resource with consumer inactivity timeout and HTTP consumer/producer enablement.
```
scholzj, line 5: As this will be part only of the 0.42 release, can you please move this to the 0.42.0 section?

api/src/main/java/io/strimzi/api/kafka/model/bridge/KafkaBridgeConsumerSpec.java
```diff
@@ -21,16 +21,39 @@
...
+    private long timeoutSeconds = HTTP_DEFAULT_TIMEOUT;
+
     @Override
     @Description("The Kafka consumer configuration used for consumer instances created by the bridge. Properties with the following prefixes cannot be set: " + FORBIDDEN_PREFIXES + " (with the exception of: " + FORBIDDEN_PREFIX_EXCEPTIONS + ").")
     public Map<String, Object> getConfig() {
         return config;
     }
+
+    @Description("Whether the HTTP consumer should be enabled or disabled")
+    public boolean isEnabled() {
```
scholzj, line 43: It might be good to add the `@JsonInclude(JsonInclude.Include.NON_DEFAULT)` annotation.

api/src/main/java/io/strimzi/api/kafka/model/bridge/KafkaBridgeConsumerSpec.java
```diff
@@ -21,16 +21,39 @@
...
+        return enabled;
+    }
+
+    public void setEnabled(boolean enabled) {
+        this.enabled = enabled;
+    }
+
+    @Description("The timeout in seconds for deleting inactive consumers.")
+    public long getTimeoutSeconds() {
```
scholzj, line 52: Same as above, if you want to use `long` with default value, you should add `@JsonInclude(JsonInclude.Include.NON_DEFAULT)`.

api/src/main/java/io/strimzi/api/kafka/model/bridge/KafkaBridgeProducerSpec.java
```diff
@@ -21,16 +21,27 @@
...
+    public boolean isEnabled() {
```
scholzj, line 40: Same as on the other file ... `@JsonInclude(JsonInclude.Include.NON_DEFAULT)`

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaBridgeCluster.java
```diff
@@ -401,10 +404,26 @@ protected List<EnvVar> getEnvVars() {
 
         varList.add(ContainerUtils.createEnvVar(ENV_VAR_KAFKA_BRIDGE_BOOTSTRAP_SERVERS, bootstrapServers));
         varList.add(ContainerUtils.createEnvVar(ENV_VAR_KAFKA_BRIDGE_ADMIN_CLIENT_CONFIG, kafkaBridgeAdminClient == null ? "" : new KafkaBridgeAdminClientConfiguration(reconciliation, kafkaBridgeAdminClient.getConfig().entrySet()).getConfiguration()));
-        varList.add(ContainerUtils.createEnvVar(ENV_VAR_KAFKA_BRIDGE_CONSUMER_CONFIG, kafkaBridgeConsumer == null ? "" : new KafkaBridgeConsumerConfiguration(reconciliation, kafkaBridgeConsumer.getConfig().entrySet()).getConfiguration()));
-        varList.add(ContainerUtils.createEnvVar(ENV_VAR_KAFKA_BRIDGE_PRODUCER_CONFIG, kafkaBridgeProducer == null ? "" : new KafkaBridgeProducerConfiguration(reconciliation, kafkaBridgeProducer.getConfig().entrySet()).getConfiguration()));
         varList.add(ContainerUtils.createEnvVar(ENV_VAR_KAFKA_BRIDGE_ID, cluster));
 
+        if (kafkaBridgeConsumer != null) {
```
scholzj, line 409: What happens when the consumer will not be set? It seems like it would cause an _error_ state where things will fail because the env vars will be missing?

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaBridgeCluster.java
```diff
@@ -401,10 +404,26 @@ protected List<EnvVar> getEnvVars() {
...
+            varList.add(ContainerUtils.createEnvVar(ENV_VAR_KAFKA_BRIDGE_CONSUMER_CONFIG, new KafkaBridgeConsumerConfiguration(reconciliation, kafkaBridgeConsumer.getConfig().entrySet()).getConfiguration()));
+            varList.add(ContainerUtils.createEnvVar(ENV_VAR_KAFKA_BRIDGE_CONSUMER_ENABLED, String.valueOf(kafkaBridgeConsumer.isEnabled())));
+            varList.add(ContainerUtils.createEnvVar(ENV_VAR_KAFKA_BRIDGE_HTTP_CONSUMER_TIMEOUT, String.valueOf(kafkaBridgeConsumer.getTimeoutSeconds())));
+        } else {
+            varList.add(ContainerUtils.createEnvVar(ENV_VAR_KAFKA_BRIDGE_CONSUMER_CONFIG, ""));
+            varList.add(ContainerUtils.createEnvVar(ENV_VAR_KAFKA_BRIDGE_CONSUMER_ENABLED, "true"));
+            varList.add(ContainerUtils.createEnvVar(ENV_VAR_KAFKA_BRIDGE_HTTP_CONSUMER_TIMEOUT, String.valueOf(KafkaBridgeConsumerSpec.HTTP_DEFAULT_TIMEOUT)));
+        }
+
+        if (kafkaBridgeProducer != null) {
```
scholzj, line 419: Same as above.

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaBridgeCluster.java
```diff
@@ -529,6 +548,20 @@ public KafkaBridgeHttpConfig getHttp() {
...
+    public KafkaBridgeConsumerSpec getKafkaBridgeConsumer() {
+        return this.kafkaBridgeConsumer;
+    }
+
+    /**
+     * @return The Producer configuration of the Bridge
+     */
+    public KafkaBridgeProducerSpec getKafkaBridgeProducer() {
+        return this.kafkaBridgeProducer;
+    }
```
scholzj, line 563: I don't think these should be needed.

cluster-operator/src/test/java/io/strimzi/operator/cluster/model/KafkaBridgeClusterTest.java
```diff
@@ -155,6 +161,9 @@ public void testDefaultValues() {
         assertThat(kbc.readinessProbeOptions.getTimeoutSeconds(), is(5));
         assertThat(kbc.livenessProbeOptions.getInitialDelaySeconds(), is(15));
         assertThat(kbc.livenessProbeOptions.getTimeoutSeconds(), is(5));
+        assertThat(kbc.getKafkaBridgeConsumer().isEnabled(), is(true));
+        assertThat(kbc.getKafkaBridgeConsumer().getTimeoutSeconds(), is(-1L));
```
scholzj, line 165: You should really test the environment variables having the right values here. Not the internal fields I think.

documentation/api/io.strimzi.api.kafka.model.bridge.KafkaBridgeConsumerSpec.adoc
```diff
@@ -29,6 +29,16 @@ All other supported options are forwarded to Kafka Bridge, including the followi
 
 * Any `ssl` configuration for xref:con-common-configuration-ssl-reference[supported TLS versions and cipher suites]
 
+[id='property-kafka-bridge-consumer-enabled-config-{context}']
+= `enabled`
+
+Enablement of the producer can be controlled by setting the `enabled` field to `true` or `false`. Default is `true`.
```
scholzj, line 35:
```suggestion
Enablement of the consumer can be controlled by setting the `enabled` field to `true` or `false`. Default is `true`.
```

api/src/main/java/io/strimzi/api/kafka/model/bridge/KafkaBridgeSpec.java
steffen-karlsson: @ppatierno We have two options, either initialising by default, so `enabled` will be true for both of them, or initialise to `null` and then default it to `true` in the `KafkaBridgeCluster` code `getEnvVars()`.

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaBridgeCluster.java
steffen-karlsson: @scholzj initially I did not had these two statements here, as they're both default initiated in `KafkaBridgeSpec`, but SpotBugs failed and pointed out this.

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaBridgeCluster.java
steffen-karlsson: Needed for testing, an alternative will be to set `this.kafkaBridgeConsumer` and `this.kafkaBridgeProducer` as protected.

scholzj: Yeah, but these obejects are just _passed through_ as they are in the custom resource. So that test IMHO does not add any real value. The important part to test is that the environment variables are set accordingly and not that the object passed in the `fromCrd` method still looks the same. so I think the checks of these objects ae not really needed in the tests.

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaBridgeCluster.java
scholzj: Well, as Paolo already commented on the other part ... I would:
* Not initialize the objects, that can cause all kind of problems down the road
* Simply handle it here with if-else ... set the default end vars if it is null, use the real values if it isn't

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaBridgeCluster.java
ppatierno: Yeah I would agree with Jakub on this.

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaBridgeCluster.java
ppatierno: yeah, +1 with Jakub's suggestion.

steffen-karlsson: I've removed the initialisation and keep the if-else statements

cluster-operator/src/test/java/io/strimzi/operator/cluster/model/KafkaBridgeClusterTest.java
steffen-karlsson: `testDefaultValues()` only tests default values for the rest of the props, so wouldn't it be wrong to start testing env values here?

I can definitely add the test of the envs in other suitable tests in the class @scholzj :)

scholzj: * I guess this now throws an NPE as the default value it would return is null?
* But it is also not needed. As you pointed out, you test the environment variables. so this test does not add anything because if the environment variables are set correctly, then:
    * We do not really care about these objects that much
    * We know from the environment variables that these values are there and do not need to test them.


It is not like this check would be somehow harmful, but removing the unnecessary getters from the production code would make it a bit leaner and smaller.

steffen-karlsson: Will remove this test and getters

ppatierno (approved):
LGTM. Thanks!

scholzj:
Thanks for the PR.

PR #9821, opened 2024-03-12
Add refactoring around replicas change handling

fvaleri:
This puts all replicas change logic inside the checkReplicasChange() method. It also improves results handling inside updateInternal() so that results is not passed deep down the call stack.

This is also tested with the ST currently under review.

### DIFF
--- a/topic-operator/src/test/java/io/strimzi/operator/topic/v2/BatchingTopicControllerTest.java
+++ b/topic-operator/src/test/java/io/strimzi/operator/topic/v2/BatchingTopicControllerTest.java
@@ -264,7 +264,7 @@ class BatchingTopicControllerTest {
 
     @ParameterizedTest
     @ValueSource(booleans = { true, false })
-    public void replicasChangeShouldMoveFromNewToPending(boolean cruiseControlEnabled) {
+    public void replicasChangeShouldBeReconciled(boolean cruiseControlEnabled) {
         int replicationFactor = 1;
 
         // setup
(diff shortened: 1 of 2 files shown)


scholzj (approved):
LGTM assuming it works.

topic-operator/src/test/java/io/strimzi/operator/topic/v2/BatchingTopicControllerTest.java
```diff
@@ -309,185 +327,49 @@ public void replicasChangeShouldMoveToPendingFromNew(boolean cruiseControlEnable
...
         } else {
-            assertThat(results.keySet(), empty());
+            assertThat(results.errors().count(), is(1L));
+            assertThat(results.errors().findFirst().get().getValue(), instanceOf(TopicOperatorException.NotSupported.class));
         }
     }
 
-    @ParameterizedTest
-    @ValueSource(booleans = { true, false })
-    public void replicasChangeShouldMoveToOngoingFromPending(boolean cruiseControlEnabled) {
```
tombentley, line 331: Stupid question maybe, but why do we no longer need to test this?

fvaleri: This is covered by replicasChangeShouldMoveFromNewToPending. It doesn't really matter which state transition, as I'm mocking the resulting status (i.e. replicasChangeHandler behavior). What I'm really testing is the controller logic around the replicasChangeHandler invocation, which leads to results.ok() or results.error(). Should we use a better name for this test?

tombentley (approved):
Thanks @fvaleri !

PR #9827, opened 2024-03-13
[DOC] Add note about aggreated metrics produced by the quotas plugin

fvaleri:
The same note is also proposed for the plugin readme, but it is also valid for the current released version.


scholzj (approved):
Thanks.

PaulRMellor (approved):
Thanks. Small suggestion

PR #9833, opened 2024-03-13
Some minor refactoring on migration code

ppatierno:
Trivial PR to apply some minor refactoring to a couple of places related to the migration code.

### DIFF
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaMetadataStateManager.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaMetadataStateManager.java
@@ -14,7 +14,6 @@ import io.strimzi.operator.common.ReconciliationLogger;
 import io.strimzi.operator.common.model.StatusUtils;
 
 import java.util.Locale;
-import java.util.Objects;
 
 /**
  * Class used to reconcile the metadata state which represents where metadata are stored (ZooKeeper or KRaft)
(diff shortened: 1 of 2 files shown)


cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaMetadataStateManager.java
```diff
@@ -66,11 +67,7 @@ public KafkaMetadataStateManager(
         this.kraftAnno = kraftAnnotation(kafkaCr);
         KafkaMetadataState metadataStateFromKafkaCr = kafkaCr.getStatus() != null ? kafkaCr.getStatus().getKafkaMetadataState() : null;
         // missing metadata state means reconciling an already existing Kafka resource with newer operator supporting metadata state or first reconcile
-        if (metadataStateFromKafkaCr == null) {
-            this.metadataState = isKRaftAnnoEnabled() ? KafkaMetadataState.KRaft : KafkaMetadataState.ZooKeeper;
-        } else {
-            this.metadataState = metadataStateFromKafkaCr;
-        }
+        this.metadataState = Objects.requireNonNullElseGet(metadataStateFromKafkaCr, () -> isKRaftAnnoEnabled() ? KafkaMetadataState.KRaft : KafkaMetadataState.ZooKeeper);
```
scholzj, line 70: Why this change? Wasn't the original code easier to read while doing the same?

ppatierno: I see this code shorter for a simple thing and understandable at the same time like the ternary operator.

ppatierno: @strimzi/maintainers any more opinions on this?

see-quick: Well, the old one was more readable because we are used to this I think :) the new one is just a fancy one-liner...I don't have a strong opinion about what is better...but for me, the readability is better in the old one but because I am more used to that form...

im-konge: I agree that we are not much (I found one occurrence in STs) `requireNonNullElseGet` in the project and the previous format was less confusing FMPOV (as mentioned by @see-quick , just because we are not using it regularly).

Also, if you don't want to have the `else` block there, you can do
```
this.metadataState = metadataStateFromKafkaCr;

if (metadataStateFromKafkaCr == null) {
    this.metadataState = isKRaftAnnoEnabled() ? KafkaMetadataState.KRaft : KafkaMetadataState.ZooKeeper;
}
```
which will be even more confusing 😄 but it's an option.

ppatierno: ok, I reverted back to the previous code.

im-konge (approved):

see-quick (approved):

scholzj (approved):

PR #9839, opened 2024-03-15
[ST] Set `STRIMZI_USE_NODE_POOLS_IN_TESTS` to `true` by default

im-konge:
### Type of change

- Enhancement

### Description

When running STs locally, we are assuming that we have NodePools enabled. But currently, those are disabled by default by `STRIMZI_USE_NODE_POOLS_IN_TESTS`. This PR sets this env variable to `true` by default.

### Checklist


fvaleri (approved):
Thanks.

henryZrncik (approved):

scholzj (approved):

see-quick (approved):

PR #9850, opened 2024-03-19
[ST] fix flaky test in MultipleClusterOperatorsST

henryZrncik:
### Type of change

- Refactoring

### Description

test `testKafkaCCAndRebalanceWithMultipleCOs` does now wait for Cruise control to roll after Kafka has its replication factor change, while there is already created KafkaRebalance causing the test to fail as there is only wait for stable cluster, instead of wait for roll of cc Pods.

### DIFF
--- a/systemtest/src/test/java/io/strimzi/systemtest/operators/MultipleClusterOperatorsST.java
+++ b/systemtest/src/test/java/io/strimzi/systemtest/operators/MultipleClusterOperatorsST.java
@@ -241,17 +241,17 @@ public class MultipleClusterOperatorsST extends AbstractST {
      *     - Cluster Operators are successfully deployed.
      *  2. - Set up scrapers and metric collectors for first Cluster Operators.
      *  3. - Deploy Kafka Cluster with 3 Kafka replicas and label 'app.kubernetes.io/operator' pointing to the first Cluster Operator.
-     *  4. - Change Kafka's label selector 'app.kubernetes.io/operator' to point to some none existing operator.
+     *  4. - Change Kafka's label selector 'app.kubernetes.io/operator' to point to not existing Cluster Operator.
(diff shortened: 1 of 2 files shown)


henryZrncik:
@strimzi-ci run tests --profile=all --testcase=MultipleClusterOperatorsST --parallel-count=1

systemtest/src/test/java/io/strimzi/systemtest/operators/MultipleClusterOperatorsST.java
```diff
@@ -371,6 +378,7 @@ void deployCOInNamespace(String coName, String coNamespace, List<EnvVar> extraEn
         LOGGER.info("Creating: {} in Namespace: {}", coName, coNamespace);
 
         clusterOperator = clusterOperator.defaultInstallation()
+            .withReconciliationInterval(TestConstants.FASTER_RECONCILIATION_INTERVAL)
```
im-konge, line 381: The faster reconciliation interval can cause troubles during the rolling update, as not all of the operations will be done and it will start with another reconciliation.

Does it work for all platforms?

henryZrncik: Nice catch. I tested it on minikube and OC 4.14 not sure about rest.

Anywya this is not that important to me so i can remove this part with no problem. WDYT ?

im-konge: If it's not needed, you can remove it

henryZrncik: removed 👍

im-konge (approved):

see-quick (approved):
👍  thanks

Issue #9851, opened 2024-03-19
[Enhancement]: Replace x-kubernetes-preserve-unknown-fields in CRD for Kubernetes Resources object.


alexandrst88:
### Related problem

There is still an issue when provision "Kafka" resource with Terraform Kubernetes Provider.  There is not possible change resources without re-creation of the resource which leads Operator to delete and create cluster. Could you please to add it, or point me the desired design for CRD generator code and i will implement it by myself.

### Suggested solution

https://github.com/strimzi/strimzi-kafka-operator/pull/9508 but for kubernetes resources as well in 0.40.0, it was partially implemented in https://github.com/strimzi/strimzi-kafka-operator/pull/9292 but it was closed.

scholzj:
I think you have to be more specific about what is missing. As far as I know, we did all we could, and even that caused a lot of users some problems. So not sure what more are you asking for.

> Could you please to add it, or point me the desired design for CRD generator code and i will implement it by myself.

The source codes are in this repository. Unsurprisingly, the CRD generator is in the `crd-generator` directory.

msillence:
I've not tested the new CRDs with terraform yet but we are using the kubectl provider https://github.com/gavinbunney/terraform-provider-kubectl with `resource kubectl_manifest` in our terraform to avoid this problem.
Of note the CRDs aren't upgraded by helm when you upgrade the operator you need to do that by hand. Thought I'd mention it in case you still had the old CRDs.

alexandrst88:
Should be fixed in https://github.com/strimzi/strimzi-kafka-operator/pull/9857

alexandrst88:
> I've not tested the new CRDs with terraform yet but we are using the kubectl provider https://github.com/gavinbunney/terraform-provider-kubectl with `resource kubectl_manifest` in our terraform to avoid this problem. You also need to include the provider in the module folder if you are using modules. Of note the CRDs aren't upgraded by helm when you upgrade the operator you need to do that by hand. Thought I'd mention it in case you still had the old CRDs.

I'm using official terraform kubernetes provider - https://registry.terraform.io/providers/hashicorp/kubernetes/latest/docs/resources/manifest. I checked CRD's from 0.40.0 and their still have `x-kubernetes-preserve-unknown-field` for resourcers fields.

scholzj:
[...]

PR #9857, opened 2024-03-20
Handle x-kubernetes-preserve-unknown-fields in CRD for Kubernetes Resources object.

Gartium:
### Type of change

- Enhancement / new feature

### Description

Generate correct schema for x-kubernetes-int-or-string
Related issue https://github.com/strimzi/strimzi-kafka-operator/issues/9851

### Checklist

- [v] Make sure all tests pass

### DIFF
--- a/crd-generator/src/main/java/io/strimzi/crdgenerator/CrdGenerator.java
+++ b/crd-generator/src/main/java/io/strimzi/crdgenerator/CrdGenerator.java
@@ -16,6 +16,7 @@ import com.fasterxml.jackson.databind.node.ObjectNode;
 import com.fasterxml.jackson.databind.node.TextNode;
 import com.fasterxml.jackson.dataformat.yaml.YAMLGenerator;
 import com.fasterxml.jackson.dataformat.yaml.YAMLMapper;
+import io.fabric8.kubernetes.api.model.Quantity;
 import io.fabric8.kubernetes.client.CustomResource;
 import io.strimzi.api.annotations.ApiVersion;
 import io.strimzi.api.annotations.KubeVersion;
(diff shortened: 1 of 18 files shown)


crd-generator/src/test/java/io/strimzi/crdgenerator/ExampleCrd.java
```diff
@@ -54,10 +55,10 @@
         }
     )
 )
-@OneOf({@OneOf.Alternative(@OneOf.Alternative.Property("either")), @OneOf.Alternative(@OneOf.Alternative.Property("or")), @OneOf.Alternative({@OneOf.Alternative.Property("mapStringString"), @OneOf.Alternative.Property("mapStringObject")})})
+@OneOf({@OneOf.Alternative(@OneOf.Alternative.Property("either")), @OneOf.Alternative(@OneOf.Alternative.Property("or")), @OneOf.Alternative({@OneOf.Alternative.Property("mapStringString"), @OneOf.Alternative.Property("mapStringQuantity"), @OneOf.Alternative.Property("mapStringObject")})})
```
scholzj, line 58: Why do you tie it into the alternative fields? I do not understand the relationship of it.

crd-generator/src/main/java/io/strimzi/crdgenerator/CrdGenerator.java
```diff
@@ -885,9 +890,16 @@ private ObjectNode buildArraySchema(ApiVersion crApiVersion, Property property,
...
 
-    private boolean isStringStringMap(PropertyType propertyType) {
+    /**
+     * Utility method to check if Map key-value pair match specific types.
+     * @param propertyType property to check
+     * @param keyType key Class
+     * @param valueType value Class
+     * @return true if key-value types are equal to specified types, false otherwise.
+     */
+    private boolean isSameTypesMap(PropertyType propertyType, Class<?> keyType, Class<?> valueType) {
```
scholzj, line 900: I don't think the name makes sense as the types are not the same. You have different types for key and value. So you should probably find a better name? Maybe `isMapWithTypes`?

scholzj (commented):
I left some nits. It might be also useful to add the resource request / limit to one of the `api` module IT test - e.g. for the Kafka CR or something and test that it is properly envoded / decoded.

Gartium:
Addressed comments

crd-generator/src/test/java/io/strimzi/crdgenerator/ExampleCrd.java
Gartium: Removed, it from alternatives. Wasn't sure if it should be part of it.

crd-generator/src/main/java/io/strimzi/crdgenerator/CrdGenerator.java
Gartium: renamed to `isMapOfTypes`

scholzj (approved):
LGTM. Thanks. Although I still think this would be a good idea:

> It might be also useful to add the resource request / limit to one of the api module IT test - e.g. for the Kafka CR or something and test that it is properly envoded / decoded.

ppatierno (approved):

PR #9876, opened 2024-03-26
[ST] enable all system test disable due to UTO and add admin client utils 

henryZrncik:
### Type of change

- Refactoring

### Description

**Allow all tests disabled due to `UTONotSupported`**
refactor of all checks and tests in ST suites to run with UTO. 
- removing some steps from recovery tests and whole test altogether (`TopicST#testCreateTopicViaKafka`)  which is now (with UTO only) doing nothing
- removal of `UTONotSupported` and `UTONotSupportedCondition`

**Some of Kafka Scripts usages transition to Admin Client**. 
- implementing functionality for adminClient command `describeTopic` which can parse KafkaTopicDescription object with `partitionCount`, `name`, `replicaCount` (which are effectively all we care about the kafka topic in tests so far). 
- adding 8 methods into `AdminClientTemplates` for deploying admin client on any of combination `SSL`, `PLAINTEXT`, `SASL_PLAINTEXT` , `SASL_SSL`  security with/without extra `configuration` parameter.  
- removal of some methods from KafkaTopicUtils which works with kafka scripts  which can be now ignored due to using adminClient. 
- add new label into admin client which would resolve problem of having unique labels in case there are 2 admin clients at the same namespace. 

**Minor refactors**
- admin client deployment.
- util methods in cruise control and their calls in Cruise Control STs.


henryZrncik:
Do not merge before release of new test clients.

systemtest/src/main/java/io/strimzi/systemtest/kafkaclients/internalClients/admin/KafkaAdminException.java
```diff
@@ -0,0 +1,15 @@
+/*
+ * Copyright Strimzi authors.
+ * License: Apache License 2.0 (see the file LICENSE or http://apache.org/licenses/LICENSE-2.0.html).
+ */
+package io.strimzi.systemtest.kafkaclients.internalClients.admin;
+
+class KafkaAdminException extends RuntimeException {
```
see-quick, line 7: Javadoc :)

systemtest/src/main/java/io/strimzi/systemtest/templates/specific/AdminClientTemplates.java
```diff
@@ -25,21 +33,210 @@ public class AdminClientTemplates {
 
     private AdminClientTemplates() {}
 
-    public static Deployment scramShaAdminClient(String namespaceName, String userName, String adminName, String bootstrapName, String additionalConfig) {
-        String finalAdditionalConfig = getAdminClientScramConfig(namespaceName, userName) + "\n" + additionalConfig;
+    ///////////////////////////////////////////
+    //              Tls (SSL)
```
see-quick, line 37:
```suggestion
    //              TLS (SSL)
```

systemtest/src/main/java/io/strimzi/systemtest/templates/specific/AdminClientTemplates.java
```diff
@@ -25,21 +33,210 @@ public class AdminClientTemplates {
...
+     * @param rm The ResourceManager deploying admin client Pod.
+     * @param namespace The Namespace in which to deploy the AdminClient and where Kafka resides.
+     * @param userName The Kafka userName to correctly configure TLS.
+     * @param adminName The name of the AdminClient deployment.
+     * @param clusterName The name of the Kafka cluster to which the AdminClient will connect.
+     * @param bootstrapName The name of the Kafka bootstrap server to use for the initial connection.
+     * @param additionalConfig A string representing additional configuration settings for the AdminClient which will be added into configuration necessary for TLS communication.
+     * @return An instance of AdminClient configured with TLS settings.
+     */
+    public static AdminClient deployTlsAdminClient(ResourceManager rm, String namespace, String userName, String adminName, String clusterName, String bootstrapName, String additionalConfig) {
```
see-quick, line 60: Templates servers as space where you have a definition of all objects, which we use in our test suites (i.e., we have pre-defined resources that repeat many times). Here you break this contract by adding inside `Templates` a ResourceManager, which is bad. You should always use 
```java
AdminClientTemplates.tlsAdminClient(
                namespace,
                userName,
                adminName,
                clusterName,
                bootstrapName,
                additionalConfig
                ));
```
In test case and call `rm.createResourceWithWait(....` also in test case.

systemtest/src/main/java/io/strimzi/systemtest/templates/specific/AdminClientTemplates.java
```diff
@@ -25,21 +33,210 @@ public class AdminClientTemplates {
...
+     * @param rm The ResourceManager deploying admin client Pod.
+     * @param namespace The Namespace in which to deploy the AdminClient and where Kafka resides.
+     * @param userName The Kafka userName to correctly configure SCRAM-SHA.
+     * @param adminName The name of the AdminClient deployment.
+     * @param clusterName The name of the Kafka cluster to which the AdminClient will connect.
+     * @param bootstrapName The name of the Kafka bootstrap server to use for the initial connection.
+     * @param additionalConfig A string representing additional configuration settings for the AdminClient which will be added into configuration necessary for communication.
+     * @return An instance of AdminClient configured with TLS and SCRAM-SHA settings.
+     */
+    public static AdminClient deployScramShaOverTlsAdminClient(ResourceManager rm, String namespace, String userName, String adminName, String clusterName, String bootstrapName, String additionalConfig) {
```
see-quick, line 118: same as previous comment

systemtest/src/main/java/io/strimzi/systemtest/templates/specific/AdminClientTemplates.java
```diff
@@ -25,21 +33,210 @@ public class AdminClientTemplates {
...
+     *
+     * @param rm The ResourceManager deploying admin client Pod.
+     * @param namespace The Namespace in which to deploy the AdminClient and where Kafka resides.
+     * @param userName The Kafka userName to correctly configure SCRAM-SHA.
+     * @param adminName The name of the AdminClient deployment.
+     * @param bootstrapName The name of the Kafka bootstrap server to use for the initial connection.
+     * @param additionalConfig A string representing additional configuration settings for the AdminClient which will be added into configuration necessary for communication.
+     * @return An instance of AdminClient configured with PLAINTEXT and SCRAM-SHA settings.
+     */
+    public static AdminClient deployScramShaOverPlainAdminClient(ResourceManager rm, String namespace, String userName, String adminName, String bootstrapName, String additionalConfig) {
```
see-quick, line 173: same as previous comment

systemtest/src/main/java/io/strimzi/systemtest/templates/specific/AdminClientTemplates.java
```diff
@@ -25,21 +33,210 @@ public class AdminClientTemplates {
...
+     * Deploys an AdminClient with PLAINTEXT communication.
+     *
+     * @param rm The ResourceManager deploying admin client Pod.
+     * @param namespace The Namespace in which to deploy the AdminClient and where Kafka resides.
+     * @param adminName The name of the AdminClient deployment.
+     * @param bootstrapName The name of the Kafka bootstrap server to use for the initial connection.
+     * @param additionalConfig A string representing additional configuration settings for the AdminClient which will be added into configuration necessary for communication.
+     * @return An instance of AdminClient communicating in PLAINTEXT.
+     */
+    public static AdminClient deployPlainAdminClient(ResourceManager rm, String namespace, String adminName, String bootstrapName, String additionalConfig) {
```
see-quick, line 217: same as previous comment

systemtest/src/main/java/io/strimzi/systemtest/utils/kubeUtils/objects/PodUtils.java
```diff
@@ -62,32 +62,42 @@ public static void waitForPodsReady(String namespaceName, LabelSelector selector
...
+     * @param expectPods The expected number of pods that should be ready.
+     * @param containers If true, checks individual container readiness within the pods.
+     * @param onTimeout The action to be executed if the pods do not become ready before the timeout.
+     */
+    public static void waitForPodsReady(String namespace, LabelSelector selector, int expectPods, boolean containers, Runnable onTimeout) {
         TestUtils.waitFor("readiness of all Pods matching: " + selector,
             TestConstants.POLL_INTERVAL_FOR_RESOURCE_READINESS, ResourceOperation.timeoutForPodsOperation(expectPods),
             () -> {
-                List<Pod> pods = kubeClient(namespaceName).listPods(namespaceName, selector);
+                List<Pod> pods = kubeClient(namespace).listPods(namespace, selector);
```
see-quick, line 79: Why this unnecessary change? In the whole `systemtest` we mostly use `namespaceName` for String based.

systemtest/src/test/java/io/strimzi/systemtest/cruisecontrol/CruiseControlConfigurationST.java
```diff
@@ -91,11 +94,15 @@ void testDeployAndUnDeployCruiseControl() throws IOException {
         LOGGER.info("Verifying that there is no configuration to CruiseControl metric reporter in Kafka ConfigMap");
         assertThrows(WaitException.class, () -> CruiseControlUtils.verifyCruiseControlMetricReporterConfigurationInKafkaConfigMapIsPresent(CruiseControlUtils.getKafkaCruiseControlMetricsReporterConfiguration(testStorage.getNamespaceName(), testStorage.getClusterName())));
 
-        // TODO https://github.com/strimzi/strimzi-kafka-operator/issues/8864
-        /*if (!Environment.isKRaftModeEnabled()) {
-            LOGGER.info("Cruise Control Topics will not be deleted and will stay in the Kafka cluster");
-            CruiseControlUtils.verifyThatCruiseControlTopicsArePresent(testStorage.getNamespaceName(), defaultBrokerReplicaCount);
-        }*/
+        final AdminClient adminClient = AdminClientTemplates.deployPlainAdminClient(
```
see-quick, line 97: This is what I don't want. Everything related to resources should go directly through ResourceMananger. And the get it via I think as this is deployment you would need to use `DeploymentResource.get...`similarly as we do for KafkaResource.

systemtest/src/test/java/io/strimzi/systemtest/mirrormaker/MirrorMaker2ST.java
```diff
@@ -177,17 +182,20 @@ void testMirrorMaker2() {
...
-        /*if (!Environment.isKRaftModeEnabled()) {
-            KafkaTopic mirroredTopic = KafkaTopicResource.kafkaTopicClient().inNamespace(testStorage.getNamespaceName()).withName(testStorage.getMirroredSourceTopicName()).get();
-            assertThat(mirroredTopic.getSpec().getPartitions(), is(3));
-            assertThat(mirroredTopic.getMetadata().getLabels().get(Labels.STRIMZI_CLUSTER_LABEL), is(testStorage.getTargetClusterName()));
-
-            // Replace source topic resource with new data and check that mm2 update target topi
-            KafkaTopicResource.replaceTopicResourceInSpecificNamespace(testStorage.getTopicName(), kt -> kt.getSpec().setPartitions(8), testStorage.getNamespaceName());
-            KafkaTopicUtils.waitForKafkaTopicPartitionChange(testStorage.getNamespaceName(), testStorage.getMirroredSourceTopicName(), 8);
-        }*/
+        final AdminClient targetClusterAdminClient = AdminClientTemplates.deployPlainAdminClient(
```
see-quick, line 185: same as previous

systemtest/src/test/java/io/strimzi/systemtest/mirrormaker/MirrorMaker2ST.java
```diff
@@ -304,22 +312,27 @@ void testMirrorMaker2TlsAndTlsClientAuth() {
...
 
-        // TODO https://github.com/strimzi/strimzi-kafka-operator/issues/8864
-        // currently disabled for UTO, as KafkaTopic CR is not created -> we should check it directly in Kafka
-        /*if (!Environment.isKRaftModeEnabled()) {
-            KafkaTopicUtils.waitForKafkaTopicCreation(testStorage.getNamespaceName(), testStorage.getMirroredSourceTopicName());
-            KafkaTopic mirroredTopic = KafkaTopicResource.kafkaTopicClient().inNamespace(testStorage.getNamespaceName()).withName(testStorage.getMirroredSourceTopicName()).get();
+        LOGGER.info("Checking topic is mirrored correctly in target cluster");
+
+        // Deploy kafka admin on communicating with target kafka cluster.
+        final AdminClient targetClusterAdminClient =  AdminClientTemplates.deployTlsAdminClient(
```
see-quick, line 319: you know what I am gonna say... :D

systemtest/src/test/java/io/strimzi/systemtest/mirrormaker/MirrorMaker2ST.java
```diff
@@ -425,16 +438,20 @@ void testMirrorMaker2TlsAndScramSha512Auth() {
...
-        LOGGER.info("Messages successfully mirrored");
 
-        // TODO https://github.com/strimzi/strimzi-kafka-operator/issues/8864
-        /*if (!Environment.isUnidirectionalTopicOperatorEnabled()) {
-            KafkaTopicUtils.waitForKafkaTopicCreation(testStorage.getNamespaceName(), testStorage.getMirroredSourceTopicName());
-            KafkaTopic mirroredTopic = KafkaTopicResource.kafkaTopicClient().inNamespace(testStorage.getNamespaceName()).withName(testStorage.getMirroredSourceTopicName()).get();
+        LOGGER.info("Checking topic is mirrored correctly in target cluster");
+
+        // deploy admin client
+        final AdminClient targetClusterAdminClient = AdminClientTemplates.deployScramShaOverTlsAdminClient(resourceManager,
```
see-quick, line 445: same

see-quick (commented):
Looks good. That's a lot of work done thanks 💯 ! I have a few comments ...

systemtest/src/main/java/io/strimzi/systemtest/templates/specific/AdminClientTemplates.java
```diff
@@ -90,12 +199,76 @@ public static DeploymentBuilder defaultAdminClient(String namespaceName, String
...
+        return  new EnvVarBuilder()
+            .withName("CA_CRT")
+            .withNewValueFrom()
+                .withNewSecretKeyRef()
+                    .withName(KafkaResources.clusterCaCertificateSecretName(clusterName))
+                    .withKey("ca.crt")
+                .endSecretKeyRef()
+            .endValueFrom()
+            .build();
+    }
```
see-quick, line 241: This seems like a generic to use in other resources also.

systemtest/src/main/java/io/strimzi/systemtest/templates/specific/AdminClientTemplates.java
```diff
@@ -90,12 +199,76 @@ public static DeploymentBuilder defaultAdminClient(String namespaceName, String
...
+
+    /**
+     * Generates a list of {@link EnvVar} environment variables for TLS configuration, including the user's certificate and key.
+     * These are extracted from Kubernetes secrets associated with the specified user. This setup is necessary for TLS (only) client
+     * authentication against a Kafka cluster.
+     *
+     * @param userName the name of the user, which corresponds to the Kubernetes secret names containing the user's TLS certificate and key
+     * @return a {@link List} of {@link EnvVar} instances for configuring a Kafka client with TLS
+     */
+    private static List<EnvVar> getTlsEnvironmentVariables(String userName) {
```
see-quick, line 251:
```suggestion
    private static List<EnvVar> buildTLSUserCredentials(String userName) {
```

systemtest/src/main/java/io/strimzi/systemtest/templates/specific/AdminClientTemplates.java
```diff
@@ -90,12 +199,76 @@ public static DeploymentBuilder defaultAdminClient(String namespaceName, String
...
+        EnvVar userCrt = new EnvVarBuilder()
```
see-quick, line 252:
```suggestion
        EnvVar userCertificate = new EnvVarBuilder()
```

systemtest/src/main/java/io/strimzi/systemtest/templates/specific/AdminClientTemplates.java
```diff
@@ -90,12 +199,76 @@ public static DeploymentBuilder defaultAdminClient(String namespaceName, String
...
+            .withName("USER_CRT")
+            .withNewValueFrom()
+                .withNewSecretKeyRef()
+                    .withName(userName)
+                    .withKey("user.crt")
+                .endSecretKeyRef()
+            .endValueFrom()
+            .build();
+
+        EnvVar userKey = new EnvVarBuilder()
```
see-quick, line 262:
```suggestion
        EnvVar userPrivateKey = new EnvVarBuilder()
```

systemtest/src/test/java/io/strimzi/systemtest/cruisecontrol/CruiseControlST.java
```diff
@@ -126,29 +123,16 @@ void testAutoCreationOfCruiseControlTopicsWithResources() {
...
-
-        LOGGER.info("Checking partitions and replicas for {}", CRUISE_CONTROL_PARTITION_METRICS_SAMPLES_TOPIC);
-        assertThat(partitionMetricsTopic.getPartitions(), is(32));
-        assertThat(partitionMetricsTopic.getReplicas(), is(defaultBrokerReplicaCount));
+        resourceManager.createResourceWithWait(KafkaTopicTemplates.topic(testStorage).build());
+        resourceManager.createResourceWithWait(
+            AdminClientTemplates.plainAdminClient(
+                testStorage.getNamespaceName(),
+                testStorage.getAdminName(),
+                KafkaResources.plainBootstrapAddress(testStorage.getClusterName())
```
see-quick, line 131: Don't we have inside `TestStorage` a method that gets such value? Moreover, we could then use just the TestStorage instance to pass it to `plainAdminClient`. But that's for another PR :)

systemtest/src/test/java/io/strimzi/systemtest/mirrormaker/MirrorMaker2ST.java
```diff
@@ -425,16 +443,22 @@ void testMirrorMaker2TlsAndScramSha512Auth() {
...
+                testStorage.getNamespaceName(),
+                testStorage.getTargetUsername(),
+                testStorage.getAdminName(),
+                testStorage.getTargetClusterName(),
+                KafkaResources.tlsBootstrapAddress(testStorage.getTargetClusterName())
+        ));
+        final AdminClient targetClusterAdminClient = AdminClientUtils.getConfiguredAdminClient(testStorage.getNamespaceName(), testStorage.getAdminName());
+
+        AdminClientUtils.waitForTopicPresence(targetClusterAdminClient, testStorage.getMirroredSourceTopicName());
+        assertThat(targetClusterAdminClient.describeTopic(testStorage.getMirroredSourceTopicName()).partitionCount(), is(3));
```
see-quick, line 461: Wouldn't be better to get such value via fabri8? I mean that magic `3`. Or a better way is to factor out that 3 with local variable :) also in creating i.e.,
```diff
- KafkaTopicTemplates.topic(testStorage.getSourceClusterName(), testStorage.getTopicName(), 3, testStorage.getNamespaceName()).build(),
+ KafkaTopicTemplates.topic(testStorage.getSourceClusterName(), testStorage.getTopicName(), sourceTopicPartitions, testStorage.getNamespaceName()).build(),
 ```

**Review by see-quick — COMMENTED** [review]
Looks good thanks., a few suggestions to consider

File: systemtest/src/main/java/io/strimzi/systemtest/templates/specific/AdminClientTemplates.java (hunk 12 above)
Comment (henryZrncik) [reply systemtest/src/main/java/io/strimzi/systemtest/templates/specific/AdminClientTemplates.java:241]: there are only few (2) resources that needs something like this and 1 of them are clients which have this provided internally.

File: systemtest/src/test/java/io/strimzi/systemtest/cruisecontrol/CruiseControlST.java (hunk 16 above)
Comment (henryZrncik) [reply systemtest/src/test/java/io/strimzi/systemtest/cruisecontrol/CruiseControlST.java:131]: not exactly but yeah this seems like good solution in case we will use this extensively.

**Review by see-quick — APPROVED** [review]
LGTM, Assuming that the tests will pass :)

File: systemtest/src/main/java/io/strimzi/systemtest/templates/specific/AdminClientTemplates.java [hunk 18]
```diff
@@ -25,21 +28,127 @@ public class AdminClientTemplates {
...
-    public static DeploymentBuilder defaultAdminClient(String namespaceName, String adminName, String bootstrapName) {
-        return defaultAdminClient(namespaceName, adminName, bootstrapName, "");
+    /**
+     * Creates a Deployment configuration for an AdminClient with TLS settings.
+     */
+    public static Deployment tlsAdminClient(String namespaceName, String userName, String adminName, String clusterName, String bootstrapName, String additionalConfig) {
+        final List<EnvVar> tlsEnvs = buildTLSUserCredentials(userName);
+        final String finalAdditionalConfig = "sasl.mechanism=GSSAPI\n" + "security.protocol=" + SecurityProtocol.SSL + "\n"  + "\n" + additionalConfig;
+
+        return plainAdminClient(namespaceName, adminName, bootstrapName, finalAdditionalConfig)
```
im-konge, line 55: FMPOV it's a bit strange to have method `tlsAdminClient` which returns modified `plainAdminClient`. Maybe it would make sense to have `defaultAdminClient` as we have for other clients?

systemtest/src/main/java/io/strimzi/systemtest/templates/specific/AdminClientTemplates.java
```diff
@@ -25,21 +28,127 @@ public class AdminClientTemplates {
...
+                .editOrNewTemplate()
+                    .editOrNewSpec()
+                        .editFirstContainer()
+                            .addToEnv(getClusterCaCertEnv(clusterName))
+                            .addAllToEnv(tlsEnvs)
+                        .endContainer()
+                    .endSpec()
+                .endTemplate()
+            .endSpec().build();
+    }
```
im-konge, line 66:
```suggestion
                .endTemplate()
            .endSpec()
            .build();
    }
```

systemtest/src/main/java/io/strimzi/systemtest/templates/specific/AdminClientTemplates.java
```diff
@@ -25,21 +28,127 @@ public class AdminClientTemplates {
...
+        return plainAdminClient(namespaceName, adminName, bootstrapName, finalAdditionalConfig)
+            .editOrNewSpec()
+                .editOrNewTemplate()
+                    .editOrNewSpec()
+                        .editFirstContainer()
+                            .addToEnv(getClusterCaCertEnv(clusterName))
+                        .endContainer()
+                    .endSpec()
+                .endTemplate()
+            .endSpec().build();
```
im-konge, line 100:
```suggestion
                .endTemplate()
            .endSpec()
            .build();
```

systemtest/src/main/java/io/strimzi/systemtest/utils/AdminClientUtils.java
```diff
@@ -0,0 +1,87 @@
...
+        LOGGER.info("Waiting for topic: {} to be present in Kafka", topicName);
+        TestUtils.waitFor("Topic: " + topicName + " to be present in Kafka", TestConstants.GLOBAL_POLL_INTERVAL, TestConstants.GLOBAL_TIMEOUT_SHORT,
+            () -> isTopicPresent(adminClient, topicName));
+    }
+
+    public static void waitForTopicAbsence(AdminClient adminClient, String topicName) {
+        LOGGER.info("Waiting for topic: {} to be absent in Kafka", topicName);
+        TestUtils.waitFor("Topic: " + topicName + " to be absent in Kafka", TestConstants.GLOBAL_POLL_INTERVAL, TestConstants.GLOBAL_TIMEOUT_SHORT,
+            () -> !isTopicPresent(adminClient, topicName));
+    }
```
im-konge, line 42: Do these deserve javadoc as well as others?

im-konge (commented):
I left couple of comments, but otherwise this looks good. Thanks a lot :)

systemtest/src/main/java/io/strimzi/systemtest/kafkaclients/internalClients/admin/AdminClient.java
```diff
@@ -16,6 +19,11 @@ public class AdminClient {
     private final String namespaceName;
     private final String podName;
     private final static String CMD = "admin-client";
+    private final static ObjectMapper MAPPER = new ObjectMapper();
+
+    static {
+        MAPPER.configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false);
+    }
```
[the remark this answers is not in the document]
henryZrncik: because this method can be called several times and i did not want to specify it repeatedly; 
also this way it can be configured here for all usages (including future extensions) once only, without need for introducing new methods WDYT ?

systemtest/src/main/java/io/strimzi/systemtest/templates/specific/AdminClientTemplates.java
henryZrncik: good point thanks.

systemtest/src/main/java/io/strimzi/systemtest/kafkaclients/internalClients/admin/AdminClient.java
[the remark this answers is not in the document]
im-konge: well when you will need it in future, I guess you can add it? We had more things like this, which were prepared for the future, but actually they were never used and after few months/years they were removed. So in case that you will need it in the future, you can do it that way.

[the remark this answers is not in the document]
henryZrncik: yeah i see i was mainly focusing on fact that given command (`describe`) can be called several times and we do no need to set that property several times. But Yeah overall for now setting it this way is probably overkill.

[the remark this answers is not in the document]
im-konge: From what I saw I think that it wasn't needed, as you anyway have the whole record class for that result, so there will be no unknown properties I suppose. Anyway, thanks

im-konge (approved):
LGTM, thanks :) good job

PR #9882, opened 2024-03-26
Move unrelated clases from the package with resource operators

scholzj:
### Type of change

- Bugfix

### Description

The `io.strimzi.operator.common.operator.resource` package seems to contain several classes that do not seem to be directly related to the Kubernetes resource operators:
* `TimeoutException` is used by some of the Kubernetes resource operators classes, but is also used by many other classes outside of it such as ZooKeeper clients, ZooKeeepr Eraser, Kafka Roller and so on -> so it makes sense to have it a more general package. I moved it directly to `io.strimzi.operator.common`.
* `ConfigParameter` and `ConfigParameterParser` are used for configuration of our operators and have nothing to do with the Kubernetes Resource Operators. It is not clear why are they in this package. I moved them to a new package `io.strimzi.operator.common.config`.

### Checklist


operator-common/src/main/java/io/strimzi/operator/common/config/ConfigParameterParser.java
```diff
@@ -2,13 +2,14 @@
...
  */
-package io.strimzi.operator.common.operator.resource;
+package io.strimzi.operator.common.config;
 
 
 import io.fabric8.kubernetes.api.model.LocalObjectReference;
 import io.fabric8.kubernetes.api.model.LocalObjectReferenceBuilder;
 import io.strimzi.operator.common.InvalidConfigurationException;
 import io.strimzi.operator.common.model.Labels;
+import io.strimzi.operator.common.operator.resource.AbstractNamespacedResourceOperator;
```
ppatierno, line 12: I was wondering about what you had in the description ...

> ConfigParameter and ConfigParameterParser are used for configuration of our operators and have nothing to do with the Kubernetes Resource Operators.

But then we have a reference to something related to resource operator because of the `ANY_NAMESPACE` constant. Maybe such constant belongs to this class now?

scholzj: Good point. I did not notice that. I will decouple it as this coupling makes no sense.

scholzj: I moved it to the `ConfigParameter` class. I also made use of it in other classes which so far used a separate `"*"` String.

PR #9887, opened 2024-03-27
Remove Vert.x dependency from the `operator-common` module and from UO and TO

scholzj:
### Type of change

- Enhancement / new feature

### Description

After the removal of the BTO, Vert.x is no used only int he Cluster Operator. This PR moves the Vert.x based Kubernetes Resource Operators to the `cluster-operator` module together with the `VertxUtils` class. These were the last Vert.x based classes in the `operator-common` module. So this frees it from the dependency on Vert.x and stops pulling VErt.x also into the UO and TO.

Most of the changes in this PR are just moved classes and changed imports. The Kubernetes resource operators were moved to a new `kubernetes` subpackage since the `resource` operators package in the Cluster OPerator is already full of various Kafka and ZooKeeper stuff, so having a separate package provides better separation. The configuration of the Vert.x cache directory was also moved to be set only in the CO startup shell script.

### Checklist


fvaleri (approved):
Thanks.

see-quick (approved):

Frawless (approved):

ppatierno (approved):
LGTM.

PR #9892, opened 2024-03-28
Pom refactoring on OpenTelemetry related dependencies

ppatierno:
This PR is similar to what is done on the bridge with https://github.com/strimzi/strimzi-kafka-bridge/pull/891

* avoid using a otel version property as variable to declare otel alpha version. OpenTelemetry is moving to deliver different artifacts at difference pace, also some of them will be non alpha at some point. It's better having the explicit version then.
* because guava "jre" is already a direct dependency, it just add an explicit exclusion from the corresponding "androd" from otel and grpc dependencies (NOTE: it's already happening implicitly, but I think it's better having it explicitly for clarity).

### DIFF
--- a/docker-images/artifacts/kafka-thirdparty-libs/3.6.x/pom.xml
+++ b/docker-images/artifacts/kafka-thirdparty-libs/3.6.x/pom.xml
@@ -28,7 +28,7 @@
         <json-smart.version>2.4.9</json-smart.version>
         <jsonevent-layout.version>1.7</jsonevent-layout.version>
         <opentelemetry.version>1.34.1</opentelemetry.version>
-        <opentelemetry-alpha.version>${opentelemetry.version}-alpha</opentelemetry-alpha.version>
+        <opentelemetry-alpha.version>1.34.1-alpha</opentelemetry-alpha.version>
         <opentelemetry.instrumentation.version>1.32.0-alpha</opentelemetry.instrumentation.version>
         <grpc-netty-shaded.version>1.61.0</grpc-netty-shaded.version>
         <guava.version>32.1.3-jre</guava.version>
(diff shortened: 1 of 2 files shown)


docker-images/artifacts/kafka-thirdparty-libs/3.6.x/pom.xml
```diff
@@ -115,29 +115,40 @@
...
             <version>${opentelemetry.version}</version>
             <scope>runtime</scope>
-            <!-- excluded as we end up with conflicting copies in the image detected by find-classpath-collision.sh -->
             <exclusions>
+                <!-- excluded as we end up with conflicting copies in the image detected by find-classpath-collision.sh -->
                 <exclusion>
                     <groupId>com.google.errorprone</groupId>
                     <artifactId>error_prone_annotations</artifactId>
                 </exclusion>
+                <!-- Excluding Guava (which is the android version) because we bring the jre one directly already -->
```
scholzj, line 124: Why is this needed? It does not seem to pull Android build of guava.

ppatierno: Doesn't pull it in because we have direct guava jre dependency. 
If you run a `mvn dependency:tree -Dverbose` you will get this in a couple of place for otel and grpc:

```shell
(com.google.guava:guava:jar:32.1.3-jre:runtime - version managed from 32.1.3-android; omitted for duplicate)
```

If you by any chance remove the guava jre direct dependency, you pull guava android in.
I think that for clarity it's better having the exclusion so it's clear that for otel and grpc we are relying on the guava jre (which makes more sense compared to the android).

ppatierno: Yeah I was trying to make the things clearer, but you raised a good point. I will remove the exclusions, from the corresponding bridge PR as well.

PR #9919, opened 2024-04-04
[ST] Fix `LoggingChangeST#testDynamicallySetConnectLoggingLevels` for Kafka versions below 3.7.x

### LINKED ISSUE #9897: `LoggingChangeST.testDynamicallySetConnectLoggingLevels` should run only on Kafka 3.7.0+?
The system test changes done in #9834 are based on a Apache Kafka change that is available from Kafka 3.7.0 up. And the test does not seem to work properly on Kafka 3.6.x. We should probably fix it to either make the test pass on older versions as well or disable it on Kafka 3.6.x.

im-konge:
### Type of change

- Bugfix

### Description

This PR fixes the issue when running `LoggingChangeST#testDynamicallySetConnectLoggingLevels` with Kafka version (set in `ST_KAFKA_VERSION`) below 3.7.0, where logging change set across multiple connect instances was added.

It adds check for the Kafka version and based on that it assigns either 1 or 3 replicas, with which should be KafkaConnect deployed.

Fixes #9897 

### Checklist

### DIFF
--- a/systemtest/src/test/java/io/strimzi/systemtest/log/LoggingChangeST.java
+++ b/systemtest/src/test/java/io/strimzi/systemtest/log/LoggingChangeST.java
@@ -45,6 +45,7 @@ import io.strimzi.systemtest.templates.crd.KafkaTemplates;
 import io.strimzi.systemtest.templates.specific.ScraperTemplates;
 import io.strimzi.systemtest.utils.RollingUpdateUtils;
 import io.strimzi.systemtest.utils.StUtils;
+import io.strimzi.systemtest.utils.TestKafkaVersion;
 import io.strimzi.systemtest.utils.kafkaUtils.KafkaBridgeUtils;
 import io.strimzi.systemtest.utils.kafkaUtils.KafkaConnectUtils;
 import io.strimzi.systemtest.utils.kafkaUtils.KafkaConnectorUtils;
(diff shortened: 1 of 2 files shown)


systemtest/src/test/java/io/strimzi/systemtest/log/LoggingChangeST.java
```diff
@@ -712,7 +713,10 @@ void testDynamicallySetConnectLoggingLevels() {
         final Pattern log4jPatternDebugLevel = Pattern.compile("^(?<date>[\\d-]+) (?<time>[\\d:,]+) DEBUG (?<message>.+)");
         final Pattern log4jPatternInfoLevel = Pattern.compile("^(?<date>[\\d-]+) (?<time>[\\d:,]+) INFO (?<message>.+)");
 
-        final KafkaConnect connect = KafkaConnectTemplates.kafkaConnect(testStorage.getClusterName(), testStorage.getNamespaceName(), 3)
+        // logging changes on multiple Connect instances are possible from Kafka 3.7.0
```
see-quick, line 716: Should we add a comment to remember that after we support Kafka 3.7.0 and greater then remove this conditional?

see-quick (approved):
LGTM, just one note worth of consideration.. 👍 good job

systemtest/src/test/java/io/strimzi/systemtest/log/LoggingChangeST.java
im-konge: Yeah good point, we have the similar check in `MigrationST`, so I will add comment to those places

PR #9920, opened 2024-04-04
Fix incorrect path in custom authroizer docs and improve them

scholzj:
### Type of change

- Bugfix

### Description

This PR fixed an incorrect path to the 3rd party libs that changed since the docs were written. It also does some additional improvements:
* Adds an option to add the file using a Dockerfile
* Tries to create a better structure to the chapter that seems to mix subchapters followed by a general content

### Checklist

- [x] Update documentation


ppatierno (approved):

PaulRMellor (approved):
Improved. I left a few minor suggestions.

PR #9923, opened 2024-04-05
MirrorMaker2: Add warning for dangerously invalid configuration

### LINKED ISSUE #9905: [Bug]: MM2: Incorrect setting of connectCluster might cause infinite echo when IdentityReplicationPolicy is used
### Bug Description

In MirrorMaker2, you [must](https://strimzi.io/docs/operators/latest/configuring#type-KafkaMirrorMaker2Spec-reference) have the connectCluster setting to point to the same cluster alias as targetCluster in the defined mirrors. When connectCluster is pointing to sourceCluster instead, I would expect MM2 to not start.

With default replication policy, a growing number of topics is created in the cluster set as connectCluster and mirror sourceCluster, with the cluster alias prefixed again and again:
```
kafka1.kafka1.kafka1.kafka1.kafka1.kafka1.kafka1.kafka1.kafka1.kafka1.kafka1.kafka1.kafka1.kafka1.kafka1.kafka1.kafka1.kafka1.heartbeats
```

When replication policy is set to IdentityReplicationPolicy, this causes local echoing, with the source cluster (also configured as connectCluster) getting overrun by recurring mirrored copies of the messages from itself. 

This is incorrect configuration, so it is not a critical bug. However, it is a mistake I myself did with my local lab configuration and it has a potential of corrupting data in the source cluster in a way that's not trivial to recover from.

### Steps to reproduce

Deploy a Strimzi KafkaMirrorMaker2 with connectCluster set to the same value as mirror sourceCluster.

### Expected behavior

If connectCluster is not the same as mirror targetCluster, the correct thing to do would be to not start the MM2 at all.

### Strimzi version

0.40.0

### Kubernetes version

v1.28.3 (minikube v1.32.0)

### Installation method

Helm via ArgoCD

### Infrastructure

_No response_

```
  kind: KafkaMirrorMaker2
  metadata:
    name: mm2-kafka2kafka
  spec:
    connectCluster: "kafka1"
    replicas: 1
    clusters:
    - alias: "kafka1"
      bootstrapServers: my-cluster-kafka-bootstrap.strimzi:9092
    - alias: "kafka2"
      bootstrapServers: my-cluster2-kafka-bootstrap.strimzi:9092
      config:
        offset.storage.topic: mirrormaker2-cluster-offsets
        status.storage.topic: mirrormaker2-cluster-status
        config.storage.topic: mirrormaker2-cluster-configs
        group.id: mirrormaker2-cluster
        replication.factor: 1
        status.storage.replication.factor: 1
        config.storage.replication.factor: 1
        offset.storage.replication.factor: 1
    mirrors:
    - sourceCluster: "kafka1"
      targetCluster: "kafka2"
      heartbeatConnector:
        config:
          # Internal topic 'heartbeats' created at the target cluster
          heartbeats.topic.replication.factor: 3
      sourceConnector:
        tasksMax: 1
        config:
          # -1 means it will use the default replication factor configured in the broker
          replication.factor: -1
          offset-syncs.topic.replication.factor: -1
          sync.topic.acls.enabled: "false"
          replication.policy.class: "io.strimzi.kafka.connect.mirror.IdentityReplicationPolicy"
          refresh.topics.interval.seconds: 600
      checkpointConnector:
        tasksMax: 1
        config:
          # -1 means it will use the default replication factor configured in the broker
          checkpoints.topic.replication.factor: -1
          replication.policy.class: "io.strimzi.kafka.connect.mirror.IdentityReplicationPolicy"
          sync.group.offsets.enabled: "false"
          refresh.groups.interval.seconds: 600
      topicsPattern: ".*"
      groupsPattern: ".*"
```

### Additional context

_No response_

maakuth:
Fixes #9905

### Type of change

- Bugfix

### Description

In MirrorMaker2 configuration model, add check to ensure all mirrors have target set to the connect cluster. This is the only sane option anyway, but instead of raising an error message, it did start and behaved badly. See #9905.

### Checklist

- [x] Write tests
- [ ] Update documentation
- [x] Update CHANGELOG.md

### DIFF
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaMirrorMaker2Connectors.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaMirrorMaker2Connectors.java
@@ -117,7 +117,7 @@ public class KafkaMirrorMaker2Connectors {
                     }
 
                     if (!mirror.getTargetCluster().equals(connectCluster)) {
-                        errorMessages.add("Target cluster alias " + mirror.getTargetCluster() + " is used in a mirror definition, but it is not the same as the connect cluster alias " + connectCluster);
+                        errorMessages.add("Connect cluster alias (currently set to " + connectCluster + ") has to be the same as the target cluster alias " + mirror.getTargetCluster());
                     }
                 }
 
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaMirrorMaker2Connectors.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaMirrorMaker2Connectors.java
@@ -114,6 +114,11 @@ public class KafkaMirrorMaker2Connectors {
                     } else if (!existingClusterAliases.contains(mirror.getTargetCluster())) {
                         errorMessages.add("Target cluster alias " + mirror.getTargetCluster() + " is used in a mirror definition, but cluster with this alias does not exist in cluster definitions");
                     }
+
+                    String connectCluster = kafkaMirrorMaker2.getSpec().getConnectCluster();
+                    if (!mirror.getTargetCluster().equals(connectCluster)) {
+                        errorMessages.add("Target cluster alias " + mirror.getTargetCluster() + " is used in a mirror definition, but it is not the same as the connect cluster alias " + connectCluster);
+                    }
                 }
 
                 if (!errorMessages.isEmpty())   {
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaMirrorMaker2Connectors.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaMirrorMaker2Connectors.java
@@ -101,6 +101,7 @@ public class KafkaMirrorMaker2Connectors {
             } else {
                 Set<String> existingClusterAliases = kafkaMirrorMaker2.getSpec().getClusters().stream().map(KafkaMirrorMaker2ClusterSpec::getAlias).collect(Collectors.toSet());
                 Set<String> errorMessages = new HashSet<>();
+                String connectCluster = kafkaMirrorMaker2.getSpec().getConnectCluster();
 
                 for (KafkaMirrorMaker2MirrorSpec mirror : kafkaMirrorMaker2.getSpec().getMirrors())  {
                     if (mirror.getSourceCluster() == null)  {
@@ -115,7 +116,6 @@ public class KafkaMirrorMaker2Connectors {
                         errorMessages.add("Target cluster alias " + mirror.getTargetCluster() + " is used in a mirror definition, but cluster with this alias does not exist in cluster definitions");
                     }
 
-                    String connectCluster = kafkaMirrorMaker2.getSpec().getConnectCluster();
                     if (!mirror.getTargetCluster().equals(connectCluster)) {
                         errorMessages.add("Target cluster alias " + mirror.getTargetCluster() + " is used in a mirror definition, but it is not the same as the connect cluster alias " + connectCluster);
                     }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 * Added support for configuring the `externalIPs` field in node port type services.
 * The `UnidirectionalTopicOperator` feature gate moves to GA stage and is permanently enabled without the possibility to disable it.
   If the topics whose names start with `strimzi-store-topic` and `strimzi-topic-operator` still exist, you can delete them.
+* Don't allow MirrorMaker2 mirrors with target set to something else than the connect cluster. 
 
 ### Changes, deprecations and removals
 
--- a/cluster-operator/src/test/java/io/strimzi/operator/cluster/model/KafkaMirrorMaker2ConnectorsTest.java
+++ b/cluster-operator/src/test/java/io/strimzi/operator/cluster/model/KafkaMirrorMaker2ConnectorsTest.java
@@ -117,7 +117,7 @@ public class KafkaMirrorMaker2ConnectorsTest {
         assertThat(ex.getMessage(), is("KafkaMirrorMaker2 resource validation failed: " +
                 "[Each MirrorMaker 2 mirror definition has to specify the source cluster alias, " +
                 "Target cluster alias wrong-target is used in a mirror definition, but cluster with this alias does not exist in cluster definitions, " +
-                "Target cluster alias wrong-target is used in a mirror definition, but it is not the same as the connect cluster alias target]"));
+                "Connect cluster alias (currently set to target) has to be the same as the target cluster alias wrong-target]"));
     }
 
     @Test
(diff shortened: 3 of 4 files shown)


maakuth:
This is a trivial change, so I don't think any documentation changes are needed. I am wondering whether the test coverage this way is sufficient, or would it be preferable to add another test case.

I have not run this yet on real clusters, just with test suite.

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaMirrorMaker2Connectors.java
```diff
@@ -114,6 +114,11 @@ public static KafkaMirrorMaker2Connectors fromCrd(Reconciliation reconciliation,
                     } else if (!existingClusterAliases.contains(mirror.getTargetCluster())) {
                         errorMessages.add("Target cluster alias " + mirror.getTargetCluster() + " is used in a mirror definition, but cluster with this alias does not exist in cluster definitions");
                     }
+
+                    String connectCluster = kafkaMirrorMaker2.getSpec().getConnectCluster();
```
scholzj, line 118: If there will be multiple sets of mirrors, you will be extracting this again and again. Would it make sense to get it only once before we start the for loop?

scholzj (commented):
I left one nit. As for the tests:
* Would it make sense to separate the test for the non-existing alias from the test from the connect cluster not being target cluster?
* Maybe it would be also worth having a test with multiple Mirror connectors where some of them have the right target cluster and some do not to see that it validates properly in that case as well.

Also, it seems like some other tests are failing now:

```
[ERROR] Errors: 
[ERROR] io.strimzi.operator.cluster.operator.assembly.KafkaMirrorMaker2AssemblyOperatorPodSetTest.testTopicsGroupsBlacklist(VertxTestContext)
[ERROR]   Run 1: KafkaMirrorMaker2AssemblyOperatorPodSetTest.testTopicsGroupsBlacklist:1115 » InvalidResource KafkaMirrorMaker2 resource validation failed: [Target cluster alias my-cluster-tgt is used in a mirror definition, but it is not the same as the connect cluster alias null]
[ERROR]   Run 2: KafkaMirrorMaker2AssemblyOperatorPodSetTest.testTopicsGroupsBlacklist:1115 » InvalidResource KafkaMirrorMaker2 resource validation failed: [Target cluster alias my-cluster-tgt is used in a mirror definition, but it is not the same as the connect cluster alias null]
[ERROR]   Run 3: KafkaMirrorMaker2AssemblyOperatorPodSetTest.testTopicsGroupsBlacklist:1115 » InvalidResource KafkaMirrorMaker2 resource validation failed: [Target cluster alias my-cluster-tgt is used in a mirror definition, but it is not the same as the connect cluster alias null]
[ERROR]   Run 4: KafkaMirrorMaker2AssemblyOperatorPodSetTest.testTopicsGroupsBlacklist:1115 » InvalidResource KafkaMirrorMaker2 resource validation failed: [Target cluster alias my-cluster-tgt is used in a mirror definition, but it is not the same as the connect cluster alias null]
[ERROR]   Run 5: KafkaMirrorMaker2AssemblyOperatorPodSetTest.testTopicsGroupsBlacklist:1115 » InvalidResource KafkaMirrorMaker2 resource validation failed: [Target cluster alias my-cluster-tgt is used in a mirror definition, but it is not the same as the connect cluster alias null]
[ERROR]   Run 6: KafkaMirrorMaker2AssemblyOperatorPodSetTest.testTopicsGroupsBlacklist:1115 » InvalidResource KafkaMirrorMaker2 resource validation failed: [Target cluster alias my-cluster-tgt is used in a mirror definition, but it is not the same as the connect cluster alias null]
[INFO] 
[ERROR] io.strimzi.operator.cluster.operator.assembly.KafkaMirrorMaker2AssemblyOperatorPodSetTest.testTopicsGroupsExclude(VertxTestContext)
[ERROR]   Run 1: KafkaMirrorMaker2AssemblyOperatorPodSetTest.testTopicsGroupsExclude:1065 » InvalidResource KafkaMirrorMaker2 resource validation failed: [Target cluster alias my-cluster-tgt is used in a mirror definition, but it is not the same as the connect cluster alias null]
[ERROR]   Run 2: KafkaMirrorMaker2AssemblyOperatorPodSetTest.testTopicsGroupsExclude:1065 » InvalidResource KafkaMirrorMaker2 resource validation failed: [Target cluster alias my-cluster-tgt is used in a mirror definition, but it is not the same as the connect cluster alias null]
[ERROR]   Run 3: KafkaMirrorMaker2AssemblyOperatorPodSetTest.testTopicsGroupsExclude:1065 » InvalidResource KafkaMirrorMaker2 resource validation failed: [Target cluster alias my-cluster-tgt is used in a mirror definition, but it is not the same as the connect cluster alias null]
[ERROR]   Run 4: KafkaMirrorMaker2AssemblyOperatorPodSetTest.testTopicsGroupsExclude:1065 » InvalidResource KafkaMirrorMaker2 resource validation failed: [Target cluster alias my-cluster-tgt is used in a mirror definition, but it is not the same as the connect cluster alias null]
[ERROR]   Run 5: KafkaMirrorMaker2AssemblyOperatorPodSetTest.testTopicsGroupsExclude:1065 » InvalidResource KafkaMirrorMaker2 resource validation failed: [Target cluster alias my-cluster-tgt is used in a mirror definition, but it is not the same as the connect cluster alias null]
[ERROR]   Run 6: KafkaMirrorMaker2AssemblyOperatorPodSetTest.testTopicsGroupsExclude:1065 » InvalidResource KafkaMirrorMaker2 resource validation failed: [Target cluster alias my-cluster-tgt is used in a mirror definition, but it is not the same as the connect cluster alias null]
```

You probably just need to fix the custom resource used by these tests that test something different and are now failing this validation.

maakuth:
> * Would it make sense to separate the test for the non-existing alias from the test from the connect cluster not being target cluster?
> 

The non-existing-alias will also keep failing because this added test, but maybe it does make sense to add another test just for this check.

> * Maybe it would be also worth having a test with multiple Mirror connectors where some of them have the right target cluster and some do not to see that it validates properly in that case as well.

This makes sense.

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaMirrorMaker2Connectors.java
```diff
@@ -114,6 +115,10 @@ public static KafkaMirrorMaker2Connectors fromCrd(Reconciliation reconciliation,
                     } else if (!existingClusterAliases.contains(mirror.getTargetCluster())) {
                         errorMessages.add("Target cluster alias " + mirror.getTargetCluster() + " is used in a mirror definition, but cluster with this alias does not exist in cluster definitions");
                     }
+
+                    if (!mirror.getTargetCluster().equals(connectCluster)) {
+                        errorMessages.add("Target cluster alias " + mirror.getTargetCluster() + " is used in a mirror definition, but it is not the same as the connect cluster alias " + connectCluster);
```
scholzj, line 120: I wonder if users would really understand what it means. I wonder if this would make more sense:

```
Target cluster alias " + mirror.getTargetCluster() + " has to be the same as the Connect cluster alias " + connectCluster
```

WDYT @ppatierno @PaulRMellor?

maakuth: You're right, it could be clearer. Maybe like this?
```
Target cluster alias (currently set to " + mirror.getTargetCluster() + ") has to be the same as the Connect cluster alias " + connectCluster

Comment (scholzj) [reply cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaMirrorMaker2Connectors.java:120]: I think that would work. Although I wonder if it would be better to based it the other way => users normally know what their target and source clusters are. They don't realize the target one has to be the connect one. So normally they would I guess change the connect alias. So maybe saying something like this will make it more clear?

```
Connect cluster alias (currently set to " + connectCluster + ") has to be the same as the target cluster alias " + mirror.getTargetCluster()
```

maakuth: Makes sense. I think I parroted the original form from the documentation, but yours is clearer.
> The cluster alias used for Kafka Connect. The value must match the alias of the target Kafka cluster as specified in the spec.clusters configuration.

scholzj (approved):
Thanks. Looks good to me.

PR #9933, opened 2024-04-09
[ST] improve/fix waitForPods util method.

henryZrncik:
### Type of change

- Enhancement 

### Description

Waiting for deployment or other controllers with utils methods `waitForPodsReady` will not work if we don't have selector which are specific enough e.g. (two admin clients cannot be separated by just selector assuming they are in same namespace) with currently present labels. 
- This PR allows specification of pod prefix when looking for Pods of given deployment, with default to empty string `""` which catch all. This is done to not break previous implementation.

There are additional changes in  `AdminClientUtils`  and  `TopicST` to try validity of this change also on admin clients.

### DIFF
--- a/systemtest/src/main/java/io/strimzi/systemtest/utils/AdminClientUtils.java
+++ b/systemtest/src/main/java/io/strimzi/systemtest/utils/AdminClientUtils.java
@@ -1,0 +1,37 @@
+/*
+ * Copyright Strimzi authors.
+ * License: Apache License 2.0 (see the file LICENSE or http://apache.org/licenses/LICENSE-2.0.html).
+ */
+package io.strimzi.systemtest.utils;
+
+import io.strimzi.systemtest.TestConstants;
+import io.strimzi.systemtest.kafkaclients.internalClients.admin.AdminClient;
+import io.strimzi.test.TestUtils;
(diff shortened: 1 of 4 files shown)


im-konge (commented):
I think this is not needed - the Pods in general will have similar (or same) prefix based on the label selector that you specify it. For example if you specify label selector for Kafka, it will pick all Pods like `my-cluster-kafka-(0-x)` and if you specify the prefix `my-cluster-kafka`, it will return the very same Pods.

So I guess, if it's just a matter of AdminClient, you should add more labels to the Deployment, instead of adding the prefix there.

see-quick:
> I think this is not needed - the Pods in general will have similar (or same) prefix based on the label selector that you specify it. For example if you specify label selector for Kafka, it will pick all Pods like `my-cluster-kafka-(0-x)` and if you specify the prefix `my-cluster-kafka`, it will return the very same Pods.
> 
> So I guess, if it's just a matter of AdminClient, you should add more labels to the Deployment, instead of adding the prefix there.

Yeah, agree with that. There is no need to limit it with `prefix`. Adding more labels should solve that :)

henryZrncik:
closing as recommended solution will be part of https://github.com/strimzi/strimzi-kafka-operator/pull/9876

PR #9936, opened 2024-04-09
Add JBOD support to KRaft mode

### LINKED ISSUE #9437: [KRaft] Add support for JBOD storage in KRaft mode
JBOD support is currently missing in Kafka itself. But once it is added, we will need to support it in Strimzi as well. Likely, it might include at least the following changes:
* Remove the validation that rejects clusters with multiple JBOD disks in KRaft mode
* Add handling of not deleting the metadata disks in controller nodes

The details are currently not completely clear until the implementation is finished in Kafka. This issue serves as a tracker for this part of the Strimzi implementation.

scholzj:
### Type of change

- Enhancement / new feature

### Description

This PR adds JBOD support to KRaft-based Apache Kafka cluster. It is implemented based on the Strimzi Proposal [SP#67](https://github.com/strimzi/proposals/blob/main/067-kraft-jbod-support.md). 

Some notable comments about the implementation:
* It does not use an annotation to mark what volume is currently used for the KRaft metadata. It instead detects it from the existing storage configuration stored in an annotation already.
* The check if multiple volumes are marked for KRaft metadata is done in the `StorageDiff` class. While it does not fit into it based on how it is done, having it in the same class allows us to reject this change while continuing the reconciliation process instead of just throwing an exception.
* KRaft examples are updates as discussed in the proposal
* A check for the Kafka version was not explicitly mentioned in the proposal but is implemented in this PR to prevent users from deploying Kafka 3.6.x clusters with JBOD by mistake (unfortunately, this caused lot of changes as a new field describing the Kafka version had to be passed into the validation, but I believe it is worth it)
* Issue #9938 was opened to track the JBOD-based STs that have not yet been adapted to KRaft

This should resolve #9437.

### Checklist

- [x] Write tests
- [x] Update documentation
- [x] Update CHANGELOG.md

### DIFF
--- a/api/src/main/java/io/strimzi/api/kafka/model/kafka/KRaftMetadataStorage.java
+++ b/api/src/main/java/io/strimzi/api/kafka/model/kafka/KRaftMetadataStorage.java
@@ -11,7 +11,7 @@ import com.fasterxml.jackson.annotation.JsonValue;
  * Enum that is used to define if the volume should be used to store the KRaft metadata. Currently, it has only two
  * options:
  * - shared for a volume that should share both KRaft metadata and regular logs
- * - Null for automatic assignment of the KRaft metadata to tne volume with lowest ID
+ * - Null for automatic assignment of the KRaft metadata to the volume with lowest ID
  */
 public enum KRaftMetadataStorage {
     SHARED;
--- a/api/src/main/java/io/strimzi/api/kafka/model/kafka/KRaftMetadataStorage.java
+++ b/api/src/main/java/io/strimzi/api/kafka/model/kafka/KRaftMetadataStorage.java
@@ -1,0 +1,40 @@
+/*
+ * Copyright Strimzi authors.
+ * License: Apache License 2.0 (see the file LICENSE or http://apache.org/licenses/LICENSE-2.0.html).
+ */
+package io.strimzi.api.kafka.model.kafka;
+
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonValue;
+
+/**
+ * Enum that is used to define if the volume should be used to store the KRaft metadata. Currently, it has only two
+ * options:
+ * - shared for a volume that should share both KRaft metadata and regular logs
+ * - Null for automatic assignment of the KRaft metadata to a volume
+ */
+public enum KRaftMetadataStorage {
+    SHARED;
+
+    @JsonCreator
+    public static KRaftMetadataStorage forValue(String value) {
+        //noinspection SwitchStatementWithTooFewBranches
+        switch (value) {
+            case "shared":
+                return SHARED;
+            default:
+                return null;
+        }
+    }
+
+    @JsonValue
+    public String toValue() {
+        //noinspection SwitchStatementWithTooFewBranches
+        switch (this) {
+            case SHARED:
+                return "shared";
+            default:
+                return null;
+        }
+    }
+}
--- a/api/src/main/java/io/strimzi/api/kafka/model/kafka/KRaftMetadataStorage.java
+++ b/api/src/main/java/io/strimzi/api/kafka/model/kafka/KRaftMetadataStorage.java
@@ -11,7 +11,7 @@ import com.fasterxml.jackson.annotation.JsonValue;
  * Enum that is used to define if the volume should be used to store the KRaft metadata. Currently, it has only two
  * options:
  * - shared for a volume that should share both KRaft metadata and regular logs
- * - Null for automatic assignment of the KRaft metadata to a volume
+ * - Null for automatic assignment of the KRaft metadata to tne volume with lowest ID
  */
 public enum KRaftMetadataStorage {
     SHARED;
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/model/nodepools/NodePoolUtils.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/model/nodepools/NodePoolUtils.java
@@ -10,6 +10,8 @@ import io.strimzi.api.kafka.model.kafka.Storage;
 import io.strimzi.api.kafka.model.nodepool.KafkaNodePool;
 import io.strimzi.api.kafka.model.nodepool.ProcessRoles;
 import io.strimzi.operator.cluster.model.KafkaPool;
+import io.strimzi.operator.cluster.model.KafkaVersion;
+import io.strimzi.operator.cluster.model.KafkaVersionChange;
 import io.strimzi.operator.cluster.model.ModelUtils;
 import io.strimzi.operator.cluster.model.SharedEnvironmentProvider;
 import io.strimzi.operator.common.Annotations;
@@ -31,17 +33,18 @@ public class NodePoolUtils {
     /**
      * Generates KafkaPool instances from Kafka and KafkaNodePool resources.
      *
-     * @param reconciliation    Reconciliation marker
-     * @param kafka             Kafka custom resource
-     * @param nodePools         List of node pools belonging to this cluster
-     * @param oldStorage        Maps with old storage configurations, where the key is the name of the controller
-     *                          resource (e.g. my-cluster-pool-a) and the value is the current storage configuration
-     * @param currentPods       Map with current pods, where the key is the name of the controller resource
-     *                          (e.g. my-cluster-pool-a) and the value is a list with Pod names
-     * @param useKRaft          Flag indicating if KRaft is enabled
-     * @param sharedEnvironmentProvider Shared environment provider
+     * @param reconciliation                Reconciliation marker
+     * @param kafka                         Kafka custom resource
+     * @param nodePools                     List of node pools belonging to this cluster
+     * @param oldStorage                    Maps with old storage configurations, where the key is the name of the controller
+     *                                      resource (e.g. my-cluster-pool-a) and the value is the current storage configuration
+     * @param currentPods                   Map with current pods, where the key is the name of the controller resource
+     *                                      (e.g. my-cluster-pool-a) and the value is a list with Pod names
+     * @param versionChange                 Describes Kafka versions used by this cluster
+     * @param useKRaft                      Flag indicating if KRaft is enabled
+     * @param sharedEnvironmentProvider     Shared environment provider
      *
-     * @return  List of KafkaPool instances belonging to given Kafka cluster
+     * @return List of KafkaPool instances belonging to given Kafka cluster
      */
     public static List<KafkaPool> createKafkaPools(
             Reconciliation reconciliation,
@@ -49,9 +52,9 @@ public class NodePoolUtils {
             List<KafkaNodePool> nodePools,
             Map<String, Storage> oldStorage,
             Map<String, List<String>> currentPods,
+            KafkaVersionChange versionChange,
             boolean useKRaft,
-            SharedEnvironmentProvider sharedEnvironmentProvider
-    )    {
+            SharedEnvironmentProvider sharedEnvironmentProvider)    {
         // We create the Kafka pool resources
         List<KafkaPool> pools = new ArrayList<>();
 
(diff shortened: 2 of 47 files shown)


api/src/main/java/io/strimzi/api/kafka/model/kafka/KRaftMetadataStorage.java
```diff
@@ -0,0 +1,40 @@
...
+package io.strimzi.api.kafka.model.kafka;
+
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonValue;
+
+/**
+ * Enum that is used to define if the volume should be used to store the KRaft metadata. Currently, it has only two
+ * options:
+ * - shared for a volume that should share both KRaft metadata and regular logs
+ * - Null for automatic assignment of the KRaft metadata to a volume
```
PaulRMellor, line 14:
```suggestion
 * - Null for automatic assignment of the KRaft metadata to the volume with lowest available ID
```

PaulRMellor (approved):
Looks great. I left a few suggestions.

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/nodepools/NodePoolUtils.java
```diff
@@ -215,21 +219,23 @@ private static List<String> validateKRaftProcessRoles(List<KafkaNodePool> nodePo
...
         List<String> errors = new ArrayList<>();
 
         for (KafkaNodePool pool : nodePools)    {
             if (pool.getSpec().getStorage() != null
                     && pool.getSpec().getStorage() instanceof JbodStorage jbod) {
-                if (jbod.getVolumes().size() > 1) {
-                    errors.add("Using more than one disk in a JBOD storage is currently not supported when the UseKRaft feature gate is enabled (in KafkaNodePool " + pool.getMetadata().getName() + ")");
+                if (KafkaVersion.compareDottedVersions(versionChange.to().version(), "3.7.0") < 0 && jbod.getVolumes().size() > 1) {
+                    // When running Kafka older than 3.7.0, JBOD storage is not supported in KRaft
+                    errors.add("Using more than one disk in a JBOD storage in KRaft mode is supported only with Apache Kafka 3.7.0 and newer (in KafkaNodePool " + pool.getMetadata().getName() + ")");
```
ppatierno, line 238: Is this check going to be removed when the Strimzi operator doesn't support Kafka versions minor than 3.7.0? Should we track it?

cluster-operator/src/test/java/io/strimzi/operator/cluster/model/VolumeUtilsTest.java
```diff
@@ -227,4 +230,57 @@ public void testPodSetVolumesEphemeralOnly()    {
...
+                        .build(),
+                new PersistentClaimStorageBuilder()
+                        .withDeleteClaim(true)
+                        .withId(1)
+                        .withKraftMetadata(KRaftMetadataStorage.SHARED)
+                        .withSize("10Gi")
+                        .build())
+                .build();
+        assertThat(VolumeUtils.kraftMetadataPath(markedJbodStorage), is("/var/lib/kafka/data-1"));
+    }
```
ppatierno, line 285: should we have a test about more than one volume marked for metadata storage?

docker-images/kafka-based/kafka/scripts/kafka_run.sh
```diff
@@ -65,27 +65,30 @@ echo "Using KRaft [${USE_KRAFT}]"
...
-    # Using "=" to assign arguments for the Kafka storage tool to avoid issues if the generated
-    # cluster ID starts with a "-". See https://issues.apache.org/jira/browse/KAFKA-15754
-    ./bin/kafka-storage.sh format -t="$STRIMZI_CLUSTER_ID" -r="$METADATA_VERSION" -c=/tmp/strimzi.properties
+  # Format the KRaft storage
+  STRIMZI_CLUSTER_ID=$(cat "$KAFKA_HOME/custom-config/cluster.id")
+  METADATA_VERSION=$(cat "$KAFKA_HOME/custom-config/metadata.version")
+  echo "Making sure the Kraft storage is formatted with cluster ID $STRIMZI_CLUSTER_ID and metadata version $METADATA_VERSION"
+  # Using "=" to assign arguments for the Kafka storage tool to avoid issues if the generated
+  # cluster ID starts with a "-". See https://issues.apache.org/jira/browse/KAFKA-15754
+  ./bin/kafka-storage.sh format -t="$STRIMZI_CLUSTER_ID" -r="$METADATA_VERSION" -c=/tmp/strimzi.properties -g
```
ppatierno, line 74: wondering if it's worth mentioning that the `-g` option allows to skip if the storage is already formatted.

ppatierno (commented):
I had a first pass adding some comments. I will have a run before adding more feedback (if any).

cluster-operator/src/test/java/io/strimzi/operator/cluster/model/VolumeUtilsTest.java
scholzj: Test for what exactly? Multiple metadata volumes are checked in `StorageDiffTest.testMultipleKraftMetadataVolumes`. This is not supposed to be called with multiple metadata volumes.

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/nodepools/NodePoolUtils.java
scholzj: I opened #9960 and will add a link to the code.

cluster-operator/src/test/java/io/strimzi/operator/cluster/model/VolumeUtilsTest.java
ppatierno: I missed that test in my pass then. Thanks.

systemtest/src/test/java/io/strimzi/systemtest/rollingupdate/AlternativeReconcileTriggersST.java
```diff
@@ -334,7 +334,8 @@ void testManualRollingUpdateForSinglePod() {
      *  - persistent-volume-claims
      */
     @ParallelNamespaceTest
-    @KRaftNotSupported("JBOD is not supported by KRaft mode and is used in this test case.")
+    // This test needs to be adapted to support KRaft: https://github.com/strimzi/strimzi-kafka-operator/issues/9938
+    @KRaftNotSupported("This test has not yet been adapted to support KRaft after JBOD support was added in KRat mode.")
```
tinaselenge, line 338: nit: 
```suggestion
    @KRaftNotSupported("This test has not yet been adapted to support KRaft after JBOD support was added in KRaft mode.")
```

systemtest/src/test/java/io/strimzi/systemtest/kafka/KafkaST.java
```diff
@@ -544,7 +545,8 @@ void testRegenerateCertExternalAddressChange() {
      *  - persistent-storage
      */
     @ParallelNamespaceTest
-    @KRaftNotSupported("JBOD is not supported by KRaft mode and is used in this test case.")
+    // This test needs to be adapted to support KRaft: https://github.com/strimzi/strimzi-kafka-operator/issues/9938
+    @KRaftNotSupported("This test has not yet been adapted to support KRaft after JBOD support was added in KRat mode.")
```
tinaselenge, line 549: nit: 
```suggestion
    @KRaftNotSupported("This test has not yet been adapted to support KRaft after JBOD support was added in KRaft mode.")
```

systemtest/src/test/java/io/strimzi/systemtest/kafka/KafkaST.java
```diff
@@ -400,7 +400,8 @@ void testRemoveComponentsFromEntityOperator() {
      *  - annotations
      */
     @ParallelNamespaceTest
-    @KRaftNotSupported("JBOD is not supported by KRaft mode and is used in this test case.")
+    // This test needs to be adapted to support KRaft: https://github.com/strimzi/strimzi-kafka-operator/issues/9938
+    @KRaftNotSupported("This test has not yet been adapted to support KRaft after JBOD support was added in KRat mode.")
```
tinaselenge, line 404: nit: 
```suggestion
    @KRaftNotSupported("This test has not yet been adapted to support KRaft after JBOD support was added in KRaft mode.")
```

tinaselenge (approved):

ppatierno:
@scholzj while testing this from a KRaft migration perspective I saw the following issue ...
With the current 0.40.0 (but not supporting more than one JBOD disk in KRaft of course) when you are rolling back, the `__cluster_metadata` folder is deleted when you back to ZooKeeper. It's happening with the following snipped of code in the `kafka_run` script when "no KRaft" is detected (see `Using KRaft [false]` log).

```shell
KRAFT_LOG_DIR=$(grep "log\.dirs=" /tmp/strimzi.properties | sed "s/log\.dirs=*//")

  # when in ZooKeeper mode, the __cluster_metadata folder should not exist.
  # if it does, it means a KRaft migration rollback is ongoing and it has to be removed.
  if [ -d "$KRAFT_LOG_DIR/__cluster_metadata-0" ]; then
    echo "Removing __cluster_metadata folder"
    rm -rf "$KRAFT_LOG_DIR/__cluster_metadata-0"
  fi
```

Of course it can't work when the `log.dirs` is set with multiple disks, so the check on the existence of a folder fails (because it's not just a single folder but a list).
So I have locally that part changed this way ...

```shell
# when in ZooKeeper mode, the __cluster_metadata folder should not exist.
 # if it does, it means a KRaft migration rollback is ongoing and it has to be removed.
 # also checking that metadata state is ZK (0), because if it's MIGRATION (2) it means we are rolling back but not finalized yet and KRaft quorum is still in place.
  CURRENT_KRAFT_METADATA_LOG_DIR=$(ls -d /var/lib/kafka/data-*/kafka-log"$STRIMZI_BROKER_ID"/__cluster_metadata-0 2> /dev/null || true)
  if [[ -d "$CURRENT_KRAFT_METADATA_LOG_DIR" ]] && [ "$STRIMZI_KAFKA_METADATA_CONFIG_STATE" -eq 0 ]; then
    echo "Removing __cluster_metadata folder"
    rm -rf "$CURRENT_KRAFT_METADATA_LOG_DIR"
  fi
```

Deleting the `__cluster_metadata` at the end of the rollback also helps us to hit the following [KAFKA-16463](https://issues.apache.org/jira/browse/KAFKA-16463) which is not fixed yet and needs that metadata folder being deleted before restarting a migration again.

The `MigrationST` wasn't able to catch this because it doesn't support multiple JBOD disk yet so the check that `__cluster_metadata` was deleted is hard coded this way.

https://github.com/strimzi/strimzi-kafka-operator/blob/main/systemtest/src/test/java/io/strimzi/systemtest/migration/MigrationST.java#L684

I also think that the check is testing a "wrong" folder, maybe because the tests are not using JBOD at all (even with just one disk) but the persistent storage so the path is just `/var/lib/kafka/data/`.

@im-konge I think the test should be fixed by using JBOD and even with multiple disks support when this PR is merged.

scholzj:
@ppatierno Can you maybe comment on some exact parts of the code? Because the comment is quite confusing and it is not clear to me what parts are referring to what code etc.

Also please keep in mind that migration with JBOD has a separate task and this PR does not really intend to enable migration with JBOD in any way (I assume there are some checks etc.).

ppatierno:
> Also please keep in mind that migration with JBOD has a separate task and this PR does not really intend to enable migration with JBOD in any way

I know but this PR is enabling JBOD disks in KRaft and migration is not going to work from this perspective anymore.
I am anyway fine to have the migration (rollback) back to work with a different PR after this was is merged.
Maybe it's more understandable.

> (I assume there are some checks etc.).

The check about not allowing migration with multiple JBOD disk in 0.40.0 release relies on the `NodePoolUtils.validateNodePools` that was changed by this PR. The 0.40.0 returns the following error ...

```shell
The Kafka cluster my-cluster is invalid: [Using more than one disk in a JBOD storage is currently not supported when the UseKRaft feature gate is enabled (in KafkaNodePool kafka)]                                                       
```

Now that validation additional check `KafkaVersion.compareDottedVersions(versionChange.to().version(), "3.7.0") < 0`  "broke" that validation which passes.

So the current PR allows the migration with multiple JBOD disks and it works fine but has a problem on the rollback which I described.
As I said, we can leave this PR as it is and I will fix migration rollback in a different PR.

ppatierno (approved):
As already written, I can fix the migration rollback with multiple JBOD disks in a different PR so this one LGTM.

PR #9939, opened 2024-04-10
Use a resolvable hostname for JMX RMI instead of IP

omersiar:
### Type of change
- Refactoring

### Description

It is impossible to connect JMX RMI if it advertise an IP address that is non-routable outside of the kubernetes. 

https://github.com/orgs/strimzi/discussions/7469


scholzj (commented):
Thanks for the PR. However ... I guess a good start would be to explain:
* Why does this matter, what does it solve and why / how?
* How can you guarantee that it will not break it for anyone else?

I expect the answer to the second point is that you cannot and that it will break it for people, because:
* The Pod IP address works from the same namespace
* The Pod IP address works from other namespaces
* The Pod IP works in some special network configurations sharing the IP space

The hostname returned by `hostname -s` actually doesn't work from anywhere apart form the same pod I think. And even if you add a fully qualified hostname, it might still break the third situation mentioned above.

omersiar:
Thank you for your interest,

I think you are right,  even if there was enough motivation this should ideally implemented as an opt-in configuration.

Closing it.

PR #9942, opened 2024-04-10
[ST] Move AZP profiles from STs' pom.xml and remove profiles related to Jenkins

im-konge:
### Type of change

- Enhancement

### Description

This PR creates a new settings XML file for the STs profiles for the AZP located in the `.azure/systemtests` folder and remove all the AZP profiles from STs' `pom.xml` file.
The movement of these profiles from the `pom.xml` file was discussed multiple times and as mentioned -> we should not have various profiles of the STs for each CI in the `pom.xml`, but we should follow the path of having the `settings.xml` containing needed profiles in each CI/CD.

This PR also removes few of the profiles (`bridge`, `olm_acceptance_upgrade`) that are used only on the Jenkins (and will be added into the separate `settings.xml` file directly on the Jenkins instance).

The properties from `none` profile are set by default, so I removed that profile as well.

### Checklist


scholzj (commented):
So, is that used on Azure only?

im-konge:
Those profiles with prefix `azp_` are really just profiles used on Azure.
The `bridge` and `olm_acceptance_upgrade` are profiles used in the Jenkins - where will be added as part of the separate settings.xml just for the Jenkins instance.

The `none` profile is not used as the properties are set by default.

scholzj:
Well, I guess it has nothing to do with scripts, so that likely isn't the right place. And if it is no Azure only, what about some more neutral place? E.g. as a resource in system tests? Or next to the pom.xml in system tests? Or something like that?

scholzj:
Well, the `The bridge and olm_acceptance_upgrade are profiles used in the Jenkins` part is a bit confusing. If they are related only to Azure, I think having it under `.azure` is fine. But it is not related to the scripts. So either directly in `.azure` or in some new subdirectory?

im-konge:
I changed the PR name and description - that should make the things clearer.

scholzj (approved):

see-quick (approved):

PR #9943, opened 2024-04-10
Fixed KRaft detection script when broker in POST_MIGRATION

ppatierno:
### Type of change

- Bugfix

### Description

This PR fixes the `kraft_utils.sh` which contains the function to detect the KRaft usage.
During migration, a broker is KRaft when in `POST_MIGRATION` and not since `MIGRATION` state.

### Checklist


scholzj (approved):
LGTM if it works. Thanks.

PR #9945, opened 2024-04-10
Avoid updating connector status again and again when Connect is scaled to 0

scholzj:
### Type of change

- Bugfix

### Description

When Kafka Connect is scaled to 0, we keep updating the status of its connectors again and again. This is because a different status is set by the `KafkaConnect` reconciliation based on the time and by the `KafkaConnector` watch. So the timer updates the connector status, that triggers the watch which updates it again. And next periodic reconciliation, the same repeats.

This PR updates the status update method used to set the _zero replicas_ status in the connector watch to set the same status as the periodical reconciliation does and that way, the status is updated only once. It also removed the default _empty list_ for the used topics and uses null instead when we actually cannot get the real list of topics.

### Checklist


ppatierno (approved):

see-quick (approved):
LGTM, thanks.

PR #9954, opened 2024-04-11
[ST] Fix MigrationST

im-konge:
### Type of change

- Bugfix

### Description

This PR fixes few things failing in the STs in the `MigrationST` class:

- from time to time, producer fails to send one or two messages, so I extended the `delivery.timeout.ms` (together with `request.timeout.ms`) to 30s, which should be sufficient in case of disconnection from the node
- long wait before the continuous clients are finished - changed the delay to 500 ms (from 1000ms)
- issue with deletion of KafkaTopic(s) -> the issue is only with topics from the ZK mode, when we migrated to fully KRaft, from time to time UTO wasn't able to delete the KafkaTopic(s) and because of that the clean-up of our STs was failing. For now I just turned off finalizers in UTO, but this is being investigated
- small timeout for movement of Kafka cluster state to `KRaftDualWriting` and other states -> extended to 5 minutes, however it's still not enough from time to time, depending on the KafkaRoller and when the Pods are rolled. This is being investigated as well
- small timeout for migration AZP -> increased to 240minutes.

Other than this I decreased the wait timeout for "default" resource deletion -> from 3 minutes to 2, as it should be sufficient timeout for resources like CRDs, KafkaTopic, KafkaUser, Secret, ConfigMap, etc.

### Checklist

### DIFF
--- a/systemtest/src/test/java/io/strimzi/systemtest/migration/MigrationST.java
+++ b/systemtest/src/test/java/io/strimzi/systemtest/migration/MigrationST.java
@@ -292,6 +292,7 @@ public class MigrationST extends AbstractST {
         doSecondPartOfMigration(testStorage, deleteCoDuringProcess, zkDeleteClaim);
     }
 
+    @SuppressWarnings({"checkstyle:MethodLength"})
     private void setupMigrationTestCase(TestStorage testStorage, boolean zkDeleteClaim) {
         // we assume that users will have broker NodePool named "kafka", so we will name it completely same to follow this use-case
         String brokerPoolName = "kafka";
@@ -310,7 +311,7 @@ public class MigrationST extends AbstractST {
         brokerSelector = KafkaNodePoolResource.getLabelSelector(testStorage.getClusterName(), brokerPoolName, ProcessRoles.BROKER);
         controllerSelector = KafkaNodePoolResource.getLabelSelector(testStorage.getClusterName(), testStorage.getControllerPoolName(), ProcessRoles.CONTROLLER);
 
-        String clientsAdditionConfiguration = "delivery.timeout.ms=20000\nrequest.timeout.ms=20000\nacks=all\n";
+        String clientsAdditionConfiguration = "delivery.timeout.ms=30000\nrequest.timeout.ms=30000\nacks=all\n";
 
         immediateClients = new KafkaClientsBuilder()
             .withProducerName(testStorage.getProducerName())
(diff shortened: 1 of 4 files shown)


henryZrncik (approved):
Good Job! Increase in `request.timeout.ms` and `delivery.timeout.ms` seems reasonable to me. Also the cleaning logs seems to be clean of any errors now, so thanks for looking on that.

Frawless (approved):

systemtest/src/test/java/io/strimzi/systemtest/migration/MigrationST.java
```diff
@@ -328,7 +329,7 @@ private void setupMigrationTestCase(TestStorage testStorage, boolean zkDeleteCla
             .withBootstrapAddress(KafkaResources.plainBootstrapAddress(testStorage.getClusterName()))
             .withTopicName(continuousTopicName)
             .withMessageCount(continuousMessageCount)
-            .withDelayMs(1000)
+            .withDelayMs(500)
```
ppatierno, line 332: what does exactly this set? Delay between messages? Sorry but it's not clear to me even from the PR description.

see-quick: Yes, delay between messages.

ppatierno: So how does sending messages faster help here?

im-konge: It doesn't help in any way. I just wanted to speed up the test and not wait additional 3 minutes or so before the messages will be send/received.

im-konge: And it's mentioned in here `long wait before the continuous clients are finished - changed the delay to 500 ms (from 1000ms)`

ppatierno: > And it's mentioned in here long wait before the continuous clients are finished - changed the delay to 500 ms (from 1000ms)

Which was exactly the sentence which was not clear to me :-D

ppatierno (approved):
LGTM assuming it works.
At the same time you should plan investigations on the stuff already mentioned in the description. This PR sounds like a temporary workaround to me if I got it correctly.

im-konge:
> At the same time you should plan investigations on the stuff already mentioned in the description. This PR sounds like a temporary workaround to me if I got it correctly.

The only thing that needs investigation is the deletion of topics by UTO.
The timeout for movement to `KRaftDualWriting` state is due to [KAFKA-16563](https://issues.apache.org/jira/browse/KAFKA-16563).
I will check the issue with topics further - to track it, I created https://github.com/strimzi/strimzi-kafka-operator/issues/9975.

PR #9959, opened 2024-04-15
Add more unit tests for `Label.class` related to label filtering

see-quick:
### Type of change

- Enhancement / new feature

### Description

This PR adds more unit tests for `Label.class` to filter labels in all resources (e.g., Kafka, KafkaUser...).

### Checklist

- [x] Write tests
- [x] Update documentation
- [x] Update CHANGELOG.md

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,6 @@
 * The `UnidirectionalTopicOperator` feature gate moves to GA stage and is permanently enabled without the possibility to disable it.
   If the topics whose names start with `strimzi-store-topic` and `strimzi-topic-operator` still exist, you can delete them.
 * Don't allow MirrorMaker2 mirrors with target set to something else than the connect cluster. 
-* Implement functionality to exclude KafkaUser Secret labels through regular expression filtering.
 
 ### Changes, deprecations and removals
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
 * The `UnidirectionalTopicOperator` feature gate moves to GA stage and is permanently enabled without the possibility to disable it.
   If the topics whose names start with `strimzi-store-topic` and `strimzi-topic-operator` still exist, you can delete them.
 * Don't allow MirrorMaker2 mirrors with target set to something else than the connect cluster. 
+* Implement functionality to exclude KafkaUser Secret labels through regular expression filtering.
 
 ### Changes, deprecations and removals
 
--- a/operator-common/src/main/java/io/strimzi/operator/common/model/Labels.java
+++ b/operator-common/src/main/java/io/strimzi/operator/common/model/Labels.java
@@ -154,6 +154,24 @@ public class Labels extends ResourceLabels {
     }
 
     /**
+     * Filters out labels that match the predefined exclusion pattern. This method is used to remove certain
+     * labels that should not be included based on a specific pattern, typically used for internal or sensitive
+     * keys that should not be propagated.
+     *
+     * @param labels        The original map of labels that may contain keys which need to be excluded.
+     * @return              A {@link Labels} object containing only the labels that do not match the exclusion pattern.
+     *                       If the input is null, it returns a new Labels object initialized with null.
+     */
+    public static Labels filterLabels(Map<String, String> labels) {
+        if (labels != null) {
+            labels = labels.entrySet().stream()
+                .filter(entry -> !STRIMZI_LABELS_EXCLUSION_PATTERN.matcher(entry.getKey()).matches())
+                .collect(Collectors.toMap(Map.Entry::getKey, Map.Entry::getValue));
+        }
+        return new Labels(labels);
+    }
+
+    /**
      * @param resource The resource to get the labels of.
      * @return A new instance with filtered labels added from the given {@code resource}.
      */
@@ -161,11 +179,7 @@ public class Labels extends ResourceLabels {
         Map<String, String> additionalLabels = resource.getMetadata().getLabels();
 
         if (additionalLabels != null) {
-            additionalLabels = additionalLabels
-                    .entrySet()
-                    .stream()
-                    .filter(entryset -> !STRIMZI_LABELS_EXCLUSION_PATTERN.matcher(entryset.getKey()).matches())
-                    .collect(Collectors.toMap(Map.Entry::getKey, Map.Entry::getValue));
+            additionalLabels = filterLabels(additionalLabels).toMap();
         }
 
         return additionalLabels(additionalLabels);
--- a/operator-common/src/main/java/io/strimzi/operator/common/model/Labels.java
+++ b/operator-common/src/main/java/io/strimzi/operator/common/model/Labels.java
@@ -167,8 +167,10 @@ public class Labels extends ResourceLabels {
             labels = labels.entrySet().stream()
                 .filter(entry -> !STRIMZI_LABELS_EXCLUSION_PATTERN.matcher(entry.getKey()).matches())
                 .collect(Collectors.toMap(Map.Entry::getKey, Map.Entry::getValue));
+
+            return new Labels(labels);
         }
-        return new Labels(labels);
+        return EMPTY;
     }
 
     /**
--- a/operator-common/src/main/java/io/strimzi/operator/common/model/Labels.java
+++ b/operator-common/src/main/java/io/strimzi/operator/common/model/Labels.java
@@ -154,26 +154,6 @@ public class Labels extends ResourceLabels {
     }
 
     /**
-     * Filters out labels that match the predefined exclusion pattern. This method is used to remove certain
-     * labels that should not be included based on a specific pattern, typically used for internal or sensitive
-     * keys that should not be propagated.
-     *
-     * @param labels        The original map of labels that may contain keys which need to be excluded.
-     * @return              A {@link Labels} object containing only the labels that do not match the exclusion pattern.
-     *                       If the input is null, it returns a new Labels object initialized with null.
-     */
-    public static Labels filterLabels(Map<String, String> labels) {
-        if (labels != null) {
-            labels = labels.entrySet().stream()
-                .filter(entry -> !STRIMZI_LABELS_EXCLUSION_PATTERN.matcher(entry.getKey()).matches())
-                .collect(Collectors.toMap(Map.Entry::getKey, Map.Entry::getValue));
-
-            return new Labels(labels);
-        }
-        return EMPTY;
-    }
-
-    /**
      * @param resource The resource to get the labels of.
      * @return A new instance with filtered labels added from the given {@code resource}.
      */
@@ -181,7 +161,11 @@ public class Labels extends ResourceLabels {
         Map<String, String> additionalLabels = resource.getMetadata().getLabels();
 
         if (additionalLabels != null) {
-            additionalLabels = filterLabels(additionalLabels).toMap();
+            additionalLabels = additionalLabels
+                .entrySet()
+                .stream()
+                .filter(entryset -> !STRIMZI_LABELS_EXCLUSION_PATTERN.matcher(entryset.getKey()).matches())
+                .collect(Collectors.toMap(Map.Entry::getKey, Map.Entry::getValue));
         }
 
         return additionalLabels(additionalLabels);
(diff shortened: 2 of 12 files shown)


user-operator/src/main/java/io/strimzi/operator/user/UserOperatorConfig.java
```diff
@@ -71,6 +73,14 @@ public class UserOperatorConfig {
      * Configures the default prefix of user secrets created by the operator
      */
     public static final ConfigParameter<String> SECRET_PREFIX = new ConfigParameter<>("STRIMZI_SECRET_PREFIX", STRING, "", CONFIG_VALUES);
+
+    /**
+     * Configuration parameter to define a regex pattern used for excluding certain labels
+     * from being considered in processing. This parameter allows for specifying a pattern
+     * that labels must not match to be included. If null, no labels are excluded based on the pattern.
+     */
+    public static final ConfigParameter<Pattern> SECRET_LABELS_EXCLUSION_PATTERN = new ConfigParameter<>("STRIMZI_SECRET_LABELS_EXCLUSION_PATTERN", REGEX_PATTERN, null, CONFIG_VALUES);
```
scholzj, line 82: Isn't the name different in the proposal?

user-operator/src/main/java/io/strimzi/operator/user/model/KafkaUserModel.java
```diff
@@ -387,11 +395,27 @@ protected byte[] decodeFromSecret(Secret secret, String key) {
...
+
+            labels = Util.mergeLabelsOrAnnotations(labels, templateSecretLabels)
+                .entrySet().stream()
+                .filter(entry -> !this.secretLabelExclusionPattern.matcher(entry.getKey()).matches()) // Filtering based on keys
+                .collect(Collectors.toMap(Map.Entry::getKey, Map.Entry::getValue));
+
+            LOGGER.debugOp("Updated labels: {}", labels);
+        } else {
+            LOGGER.debugOp("No Secret Label Exclusion Pattern set. Skipping label filtering.");
+        }
```
scholzj, line 412: I think this logging is a bit over the top. Maybe one debug message when it is set would be sufficient without the need to list any labels.

user-operator/src/main/java/io/strimzi/operator/user/model/KafkaUserModel.java
```diff
@@ -672,4 +696,13 @@ public String getSaslJsonConfig() {
         return getSaslJsonConfig(getScramUserName(name), scramSha512Password);
     }
 
+    /**
+     * Retrieves the compiled regular expression pattern used to exclude certain labels from being included in the secret's metadata.
+     * This pattern is applied to each label key, and labels matching this pattern are not included.
+     *
+     * @return the compiled {@link Pattern} representing the label exclusion criteria, or {@code null} if no exclusion pattern is set.
+     */
+    public Pattern getSecretLabelExclusionPattern() {
```
scholzj, line 705: Doesn't seem to be used anywhere?

user-operator/src/test/java/io/strimzi/operator/user/UserControllerMockTest.java
```diff
@@ -52,6 +52,7 @@
 import static org.mockito.Mockito.when;
 
 public class UserControllerMockTest {
+
```
scholzj, line 55: No need for this.

user-operator/src/test/java/io/strimzi/operator/user/ResourceUtils.java
scholzj, line 1: I'm not sure I follow the changes here. This is a rare option that should be used in few tests. You should not add it to these methods. In the tests where you test it you should use the builder to build a new config with this.

scholzj (commented):
I left some comments. But I think you should consider how `STRIMZI_LABELS_EXCLUSION_PATTERN` inthe `Labels` class is used and try to apply a similar pattern for both.

user-operator/src/main/java/io/strimzi/operator/user/UserOperatorConfig.java
see-quick: I was wondering if `SECRET_LABELS_EXCLUSION_PATTERN` would be more accurate to avoid confusion that it does not exclude labels from `KafkaUser` but from `Secret`. I am fine with naming it `STRIMZI_LABELS_EXCLUSION_PATTERN` if you prefer that way :)

user-operator/src/main/java/io/strimzi/operator/user/model/KafkaUserModel.java
see-quick: It is used in the `testFromCrdTlsExternalUser` test case.

user-operator/src/main/java/io/strimzi/operator/user/model/KafkaUserModel.java
see-quick: Yeah, that's true. This was just for testing purposes...updating

user-operator/src/main/java/io/strimzi/operator/user/model/KafkaUserModel.java
scholzj: Why does the test need to get it from the model? Does it not know what it set in the config? Also, it would probably not require public access, or?

user-operator/src/test/java/io/strimzi/operator/user/ResourceUtils.java
see-quick: Yeah...this is leftover because firstly I have designed a few test cases, but then I eventually decide to remove them because they were not adding any value...I will remove it. Thanks

user-operator/src/main/java/io/strimzi/operator/user/model/KafkaUserModel.java
see-quick:
```java
@Test
    public void testFromCrdTlsExternalUser()   {
        KafkaUserModel model = KafkaUserModel.fromCrd(ResourceUtils.createKafkaUser(ResourceUtils.NAMESPACE, new KafkaUserTlsExternalClientAuthentication()), UserOperatorConfig.SECRET_PREFIX.defaultValue(), Boolean.parseBoolean(UserOperatorConfig.ACLS_ADMIN_API_SUPPORTED.defaultValue()), UserOperatorConfig.SECRET_LABELS_EXCLUSION_PATTERN.type().parse(UserOperatorConfig.SECRET_LABELS_EXCLUSION_PATTERN.defaultValue()));

        assertThat(model.namespace, is(ResourceUtils.NAMESPACE));
        assertThat(model.name, is(ResourceUtils.NAME));
        assertThat(model.labels, is(Labels.fromMap(ResourceUtils.LABELS)
                        .withStrimziKind(KafkaUser.RESOURCE_KIND)
                        .withKubernetesName(KafkaUserModel.KAFKA_USER_OPERATOR_NAME)
                        .withKubernetesInstance(ResourceUtils.NAME)
                        .withKubernetesPartOf(ResourceUtils.NAME)
                        .withKubernetesManagedBy(KafkaUserModel.KAFKA_USER_OPERATOR_NAME)));
        assertThat(model.authentication.getType(), is(KafkaUserTlsExternalClientAuthentication.TYPE_TLS_EXTERNAL));
        assertThat(model.isTlsExternalUser(), is(true));
        assertThat(model.getUserName(), is("CN=" + ResourceUtils.NAME));
        assertThat(model.generateSecret(), is(nullValue()));
        assertThat(model.getSecretLabelExclusionPattern(), is(nullValue()));
    }
    ```
 In that case there is only model so we can't infer such value from config I guess.   
 
 > Also, it would probably not require public access, or
 
Yes, that's right, protected should be fine.

Comment (scholzj) [reply user-operator/src/main/java/io/strimzi/operator/user/model/KafkaUserModel.java]: No, this shows that you are testing the wrong thing. You should test the labels of the secret and not some configuration value.

Comment (see-quick) [reply user-operator/src/main/java/io/strimzi/operator/user/model/KafkaUserModel.java]: Right, I would remove it then.

File: user-operator/src/main/java/io/strimzi/operator/user/model/KafkaUserModel.java [hunk 5]
```diff
@@ -387,11 +387,15 @@ protected byte[] decodeFromSecret(Secret secret, String key) {
...
+        Map<String, String> labels = this.labels.toMap();
+
+        labels = Labels.filterLabels(Util.mergeLabelsOrAnnotations(labels, templateSecretLabels)).toMap();
+
         return new SecretBuilder()
                 .withNewMetadata()
                     .withName(getSecretName())
                     .withNamespace(namespace)
-                    .withLabels(Util.mergeLabelsOrAnnotations(labels.toMap(), templateSecretLabels))
+                    .withLabels(Util.mergeLabelsOrAnnotations(labels, templateSecretLabels))
```
scholzj, line 398: Why is this needed? The labels are taked through the `fromResource` method that already filtered them. That is how CO uses it. Why is it not good enough here?

scholzj: In other words, it already applies the exclusion, doesn't it?

see-quick: Yeah but in the constructor of `KafkaUserModel` we do 
```java
protected KafkaUserModel(String namespace, String name, Labels labels, String secretPrefix) {
        this.namespace = namespace;
        this.name = name;
        this.labels = labels.withKubernetesName(KAFKA_USER_OPERATOR_NAME)
            .withKubernetesInstance(name)
            .withKubernetesPartOf(name)
            .withKubernetesManagedBy(KAFKA_USER_OPERATOR_NAME);
        this.secretPrefix = secretPrefix;
    }
```
So the thing is that we filter labels in the `fromCrd` method scope
```java
public static KafkaUserModel fromCrd(KafkaUser kafkaUser,
                                         String secretPrefix,
                                         boolean aclsAdminApiSupported) {
        KafkaUserModel result = new KafkaUserModel(kafkaUser.getMetadata().getNamespace(),
                kafkaUser.getMetadata().getName(),
                Labels.fromResource(kafkaUser).withStrimziKind(kafkaUser.getKind()),
                secretPrefix);
               ...
               ...
```
but after that in the constructor, we add labels to the model, which seems weird to me. 
---
anyway as you pointed on that line there should be only `labels` which is filtered in L:392.

scholzj: We get the labels from the custom resource, we filter them based on the exclusions ... and then we add our own labels. That is expected. You can never filter all labels. Sure, if you dislike one of the labels we set, you might not be able to change it. But that is life and we might need to live with it. But as I said for the previous variant, I think we should aim for consistent solution everywhere and not doing something else in every part of the project.

see-quick: Well in that case I think filtering in the createSecret is not needed. And labels, which we add i.e., `app.kubernetes.io/...` will be there...and we need to live with it :D. Thus we would have a consistent solution for all projects.

operator-common/src/main/java/io/strimzi/operator/common/model/Labels.java
```diff
@@ -153,6 +153,26 @@ public Labels withAdditionalLabels(Map<String, String> additionalLabels) {
...
+    /**
+     * Filters out labels that match the predefined exclusion pattern. This method is used to remove certain
+     * labels that should not be included based on a specific pattern, typically used for internal or sensitive
+     * keys that should not be propagated.
+     *
+     * @param labels        The original map of labels that may contain keys which need to be excluded.
+     * @return              A {@link Labels} object containing only the labels that do not match the exclusion pattern.
+     *                       If the input is null, it returns a new Labels object initialized with null.
+     */
+    public static Labels filterLabels(Map<String, String> labels) {
```
scholzj, line 165: I guess this is not needed? Or at least not as public method?

CHANGELOG.md
```diff
@@ -11,6 +11,7 @@
 * The `UnidirectionalTopicOperator` feature gate moves to GA stage and is permanently enabled without the possibility to disable it.
   If the topics whose names start with `strimzi-store-topic` and `strimzi-topic-operator` still exist, you can delete them.
 * Don't allow MirrorMaker2 mirrors with target set to something else than the connect cluster. 
+* Implement functionality to exclude KafkaUser Secret labels through regular expression filtering.
```
scholzj, line 14: Technically, we do not add anything, so not sure we want to mention it here? 🤔

scholzj (commented):
I guess all we really need are the beter tests? 🤦

CHANGELOG.md
see-quick: updated

scholzj (approved):

PR #9962, opened 2024-04-15
docs(diagrams): updates digrams to be more kraft friendly

PaulRMellor:
**Documentation**

Updates documentation diagrams to be more KRaft friendly, including changing references from broker to nodes and a diagram showing secure communication in KRaft-based cluster

OAuth diagrams updated for consistency

### Checklist

- [ ] Write tests
- [x] Update documentation
- [ ] Update CHANGELOG.md


documentation/shared/images/oauth-plain-access-token.png
scholzj, line 1: It is a nit and I'm not sure it is worth changing the image. But all the other OAuth images seem to use `Kafka broker`. Only this one seems to use `Kafka node`. Might make sense to use `Kafka broker` here as well?

documentation/shared/images/secure_communication_kraft.png
scholzj, line 1: The pod numbers here are confusing. Maybe we can remove them for both brokers and controllers and just have multiple squares with `Kafka broker pod` and `Kafka controller pod` without the numbers?

(The source of confusion is that you cannot have broker `1` and controller `1` at the same time => so the numbers do not correspond to the actual node IDs. I don't think this is wrong as the number indicates the count, but it might confuse someone. The ZooKeeper based image is fine as the ZooKeeper nodes have their own IDs independent on Kafka. So no need to fix that.)

scholzj (commented):
Some nits, but looks nice otherwise.

documentation/shared/images/oauth-plain-access-token.png
PaulRMellor: I didn't spot that. We'll change that back.

documentation/shared/images/secure_communication_kraft.png
PaulRMellor: I don't think it's confusing. I wouldn't assume to think we're talking IDs here. I coule make this obvious in the surrounding text if we really need to. If we change it, we lose consistency with the zookeeper diagram. What are your thoughts @scholzj?

scholzj: Well, as I said, I would remove the numbers from the picture. I think it can be confusing because the IDs play a key part and the way they work in KRaft with the same series used for everything is already confusing enough with KRaft based on some Slack questions.

ppatierno: I would agree with Jakub and I am fine removing the numbers from ZooKeeper diagram as well.

ppatierno (approved):
LGTM.

scholzj (approved):

PR #9966, opened 2024-04-15
Fix auto-restarting of MM2 connectors

### LINKED ISSUE #9965: MM2 connector auto-restarting does not seem to work
It looks like the MM2 connector auto-restarting does not work (and my guess is that it never did). This is because:
* The autoRestart settings are never passed to the connector operator
* The status of the auto restart is never read correctly

scholzj:
### Type of change

- Bugfix

### Description

This PR fixes the auto-restarting feature for MM2 connectors. It seems that this feature never worked, because:
* The auto-restart configuration was never passed to the ConnectorSpec passed to the connector operator
* The connector operator was reading the wrong auto-restart status from MM2 but always from the KafkaConnector resource. So if the previous step ever worked, it would end up in an endless loop of very rapid restarts.

This PR fixes these issues and adds a unit test for it.

This should fix #9965.

### Checklist

- [x] Write tests

### DIFF
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaMirrorMaker2AssemblyOperator.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaMirrorMaker2AssemblyOperator.java
@@ -251,7 +251,7 @@ public class KafkaMirrorMaker2AssemblyOperator extends AbstractConnectOperator<K
      * @param kafkaMirrorMaker2 The MirrorMaker 2
      * @return A future, failed if any of the connectors could not be reconciled.
      */
-    private Future<Void> reconcileConnectors(Reconciliation reconciliation, KafkaMirrorMaker2 kafkaMirrorMaker2, KafkaMirrorMaker2Cluster mirrorMaker2Cluster, KafkaMirrorMaker2Status mirrorMaker2Status, String desiredLogging) {
+    /* test */ Future<Void> reconcileConnectors(Reconciliation reconciliation, KafkaMirrorMaker2 kafkaMirrorMaker2, KafkaMirrorMaker2Cluster mirrorMaker2Cluster, KafkaMirrorMaker2Status mirrorMaker2Status, String desiredLogging) {
         String host = KafkaMirrorMaker2Resources.qualifiedServiceName(mirrorMaker2Cluster.getCluster(), reconciliation.namespace());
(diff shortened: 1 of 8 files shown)


ppatierno (approved):
LGTM.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaMirrorMaker2AssemblyOperator.java
```diff
@@ -429,6 +429,36 @@ private Future<Void> removeAnnotation(Reconciliation reconciliation, KafkaMirror
...
     }
 
+    /**
+     * Returns the previous auto-restart status with the information about the previous restarts (number of restarts and
+     * last restart timestamp). For Mirror Maker 2, this information is gathered from the KafkaMirrorMaker2 resource
+     * passed to this method.
+     *
+     * @param reconciliation    Reconciliation marker
+     * @param connectorName     Name of the connector for which the restart should be returned
+     * @param resource          The KafkaMirrorMaker2 custom resource that configures the connector)
```
katheris, line 439:
```suggestion
     * @param resource          The KafkaMirrorMaker2 custom resource that configures the connector
```

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaMirrorMaker2AssemblyOperator.java
```diff
@@ -429,6 +429,36 @@ private Future<Void> removeAnnotation(Reconciliation reconciliation, KafkaMirror
...
+            if (kafkaMirrorMaker2.getStatus() != null
+                    && kafkaMirrorMaker2.getStatus().getAutoRestartStatuses() != null)    {
+                AutoRestartStatus ars = kafkaMirrorMaker2.getStatus().getAutoRestartStatuses().stream().filter(c -> connectorName.equals(c.getConnectorName())).findFirst().orElse(null);
+                return Future.succeededFuture(ars);
+            } else {
+                // The status does not exist yet or has no auto-restarts stored
+                return Future.succeededFuture(null);
+            }
+        } else {
+            // Something went wrong and we
```
katheris, line 456: This comment is missing the end of the sentence

katheris (approved):
Couple of nits around javadoc and comments but otherwise looks good to me

PR #9968, opened 2024-04-15
Fix too long timeout in Kubernetes-based IT tests

scholzj:
### Type of change

- Bugfix

### Description

Many of our IT tests use a specific namespace in a real Kubernetes cluster. In my environment, it seems that the deletion through the command line often fails and the namespace deletion gets stuck. However, because of how it is deleted, it waits for the deletion for 480 seconds which is very long. This PR changes how the namespaces are created and deleted to improve the reliability but also to shorten the wait times to fail quickly if needed.

### Checklist


see-quick:
> it seems that the deletion through the command line often fails and the namespace deletion gets stuck

A possible edge case is that some resources are still in that specific namespace and that's one reason why these namespaces are stuck :/. But for sure this seems to the an improvement (shorten timeouts...).

see-quick (approved):

fvaleri (approved):
LGTM. Thanks.

PR #9970, opened 2024-04-16
Integrate additional OAuth 0.15.0 config options

mstruk:
### Type of change

- Enhancement / new feature

### Description

The following options have been added to `authentication` section of type `oauth` of Kafka CR:
- `serverBearerTokenLocation`
- `userNamePrefix`

The following options have been added to `authentication` section of type `oauth` of KafkaMirrorMaker, KafkaMirrorMaker2, KafkaConnectCluster, and KafkaBridge CRs:
- `accessTokenLocation`
- `clientAssertion`
- `clientAssertionLocation`
- `clientAssertionType`
- `saslExtensions`

### Checklist

- [ ] Write tests
- [ ] Update documentation
- [ ] Update CHANGELOG.md


cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaBridgeCluster.java
```diff
@@ -99,6 +99,7 @@ public class KafkaBridgeCluster extends AbstractModel implements SupportsLogging
     protected static final String ENV_VAR_KAFKA_BRIDGE_OAUTH_ACCESS_TOKEN = "KAFKA_BRIDGE_OAUTH_ACCESS_TOKEN";
     protected static final String ENV_VAR_KAFKA_BRIDGE_OAUTH_REFRESH_TOKEN = "KAFKA_BRIDGE_OAUTH_REFRESH_TOKEN";
     protected static final String ENV_VAR_KAFKA_BRIDGE_OAUTH_PASSWORD_GRANT_PASSWORD = "KAFKA_BRIDGE_OAUTH_PASSWORD_GRANT_PASSWORD";
+    protected static final String ENV_VAR_KAFKA_BRIDGE_OAUTH_CLIENT_ASSERTION = "KAFKA_BRIDGE_OAUTH_CLIENT_ASSERTION";
```
scholzj, line 102: This seems to be used only in tests. So why do we need it? Should we move it to the test class? Or should it be actually set somewhere? The same applies to some of the other classes as well.

scholzj (commented):
I thought we would include docs on how to configure it for the Servic Accoutn configuration. Or will that be a separate Pr?

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaBridgeCluster.java
mstruk: I'm just following a pattern here. It's the [kafka_bridge_config_generator.sh](https://github.com/strimzi/strimzi-kafka-bridge/blob/0.29.0/bin/docker/kafka_bridge_config_generator.sh) shell script that uses these same env vars to configure the bridge. I assume they are in this class as they are considered part of the contract. What does @ppatierno think?

mstruk:
> I thought we would include docs on how to configure it for the Servic Accoutn configuration. Or will that be a separate Pr?

I'm making a separate PR for that. But I can also add it to this one if you prefer.

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaBridgeCluster.java
scholzj: So, I'm a bit curios how you set the environment variable if you don't use the name anywhere in the production code. Is it assembled somewhere on the fly from some prefix and suffix? But I'm not aware of any contract that we add there a variable not referenced from the Java code itself - I'm certainly not following such contract in any of my PRs.

ppatierno: So, the above env vars related constant were used here in the past as well IIRC but then we moved to this approach to build them https://github.com/strimzi/strimzi-kafka-operator/blob/main/cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaBridgeCluster.java#L458 
At this point, I guess we can removed them from here and move to test classes.

mstruk: Do you prefer we do the refactoring as part of this PR?
The change is not really related to oauth-0.15.0 configs so it would make more sense as a separate PR which I can prepare.

ppatierno: A separate PR is fine with me. Thanks.

scholzj: I think it is fine to move the others in a separate PR. But we should not introduce new variables like this in this PR.

mstruk: @ppatierno Given that I had to move one constant based on additional feedback, it was actually trivial to move the others as well and avoid the overhead of doing another PR. The idea of another PR was to get this one merged ASAP. But at this point doing another PR is an overhead.

ppatierno: Ok makes sense thanks

ppatierno (approved):

scholzj (approved):

scholzj:
Thanks @mstruk

PR #9974, opened 2024-04-16
Improve tools/report.sh to make only the namespace required

### LINKED ISSUE #9969: [Enhancement]: Improve tools/report.sh to make only the namespace required (make the cluster name optional)
### Related problem

- Currently, the cluster name is required for tools/report.sh. Also, for Bridge, Connect, and MM2, if they are not specified, there is no information about them, and it is not possible to specify multiple them at the same time.
- Many other reporting tools, such as OpenShift, only specify the namespace. However, the current tools/report.sh requires the exact names of the cluster, Bridge, Connect, and MM2 to get those information. So mistakes are common.

### Suggested solution

It would be useful for usability and customer satisfaction if cluster names were not required and only namespace should be specified.

### Alternatives

_No response_

### Additional context

_No response_

tyamashi-oss:
### Type of change
- Enhancement / new feature

### Description

This pull request:
resolve #9969 

Improve tools/report.sh to make only the namespace required (make the cluster name optional) 

[Specification]
```
./tools/report.sh --namespace=my-namespace 
```
If only the namespace is specified, all Kafka clusters, Bridges, Connects, and MM2s associated with Strimzi in that namespace are reported. Previously, the cluster name was also required. 

As before, it is also possible to limit the scope of reporting by specifying Kafka Cluster(--cluster), Bridge(--bridge), Connect(--connect), and MM2(--mm2).

The behavior has changed slightly from previous implementations. Previously, when --namespace and --cluster were specified (the previous common usage), Bridges, Connects, and MM2s were not reported. However, the current implementation reports all information on those Bridges, Connects, and MM2s in the namespace without the explicit option such as --bridge, --connect and --mm2. But I believe this is not a problematic behavior change because, previously, it was inconvenient to have to specify and report bridges, connects, and mm2s one by one, and we can rather consider the previous behavior a bug.

[Implementation]
As implementing considerations, I have tried to implement as few changes from the existing code as possible for backward compatibility.
Although the number of lines changed was large because of the indent changes, the actual changes were not that significant. The use of the uniq command is added, but other than that, no new commands are added. 

[Test]
Tests were run in my local RHEL 8 environment using Strimzi 0.38(based on Kakfa 3.6) and Strimzi 0.24(based on Kakfa 2.8).

### Checklist

- [ ] Write tests
- [ ] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/tools/report.sh
+++ b/tools/report.sh
@@ -56,9 +56,9 @@ Usage: report.sh [options]
 
 Required:
   --namespace=<string>          Kubernetes namespace.
-  --cluster=<string>            Kafka cluster name.
 
 Optional:
+  --cluster=<string>            Kafka cluster name.
   --bridge=<string>             Bridge component name to get pods and logs.
   --connect=<string>            Connect component name to get pods and logs.
   --mm2=<string>                MM2 component name to get pods and logs.
@@ -100,7 +100,7 @@ while getopts "$OPTSPEC" optchar; do
 done
 shift $((OPTIND-1))
 
-if [[ -z $NAMESPACE || -z $CLUSTER ]]; then
+if [[ -z $NAMESPACE ]]; then
   error "$USAGE"
 fi
 
@@ -117,11 +117,6 @@ if [[ "$SECRETS_OPT" != "all" && "$SECRETS_OPT" != "off" && "$SECRETS_OPT" != "h
   error "$USAGE"
 fi
 
-if [[ -z $CLUSTER ]]; then
-  echo "Cluster was not specified. Use --cluster option to specify it."
-  error "$USAGE"
-fi
-
 if [[ -z $NAMESPACE ]]; then
   echo "Namespace was not specified. Use --namespace option to specify it."
   error "$USAGE"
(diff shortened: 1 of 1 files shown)


scholzj (commented):
Thanks for the PR. However, I think this is potentially dangerous as it might mean that way more information is added to the archive than users might want / need and then be all published publicly. I think this should be rejected.

tyamashi-oss:
@scholzj  Thank you for the review. 
I understand your concern that more information is output than the user would expect.

How about making the --cluster option unnecessary only when an option such as --all-components is explicitly specified?

From a technical support engineer's perspective, I believe it would be useful to have output of reports related to Strimzi on a per namespace basis. With this perspective, I have used to draft the following for the AMQ Streams project.
- https://issues.redhat.com/browse/ENTMQST-5853
- https://issues.redhat.com/browse/ENTMQST-5298

If there are concerns with options such as --all-components as well, feel free to reject this pull request.

scholzj:
@tyamashi-oss I personally think this tool is rarely used in Strimzi and is not worth any further investments. But if there is any, we should make sure that they are _secure_ and do not expose unwanted information from the users. That probably requires some careful consideration and discussions which should be done on ethe issue as art of its triage and not in a PR like this.

tyamashi-oss:
@scholzj Thank you for the advice.

From a support engineer's perspective, we first ask the customer for both tools/report.sh and "oc adm inspect ns/${NAMESPACE}" results. Because "oc adm inspect ns/${NAMESPACE}" does not output Strimzi's custom resources, we use tools/report.sh to make up for that.

I had hoped to get a complete set by just using tools/report.sh. However, on the other hand,  if a simple script is created to obtain only Strimzi's custom resources, it would avoid using tools/report.sh by using a combination of "oc adm inspect ns/${NAMESPACE}" as we have done so far. 

My personal opinion is that if it is difficult to invest in tools/report.sh, it is fine to remove it.
Please feel free to reject this pull request.

scholzj:
Discussed on the community call on 18.4.: Thanks for the PR. However, we decided to reject this idea as explained in the #9969 issue.

Issue #9986, opened 2024-04-17
Logging configuration is never updated for Connect when connector operator is disabled


scholzj:
Currently, when the connector operator (connector resources / `strimzi.io/use-connector-resources` annotation) is disabled, we never update the logging configuration - neither dynamically nor through a rolling restart. Not doing the dynamic update is likely caused by the following logic:
* Without the connector operator, users need to manage connectors through REST
* As users need to manage things through REST we cannot setup network policies for it easily as we would block access to it for the users
* Without the network policies, we cannot guarantee that we will have access to the REST API for dynamic configuration

This logic makes sense, but not changing the logging configuration at all due to this seems like a major gap that should be addressed. I guess we have several options:
1. Create the network policy even when the connector operator is disabled and force users to create their own network policies to access the REST API
2. Ignore the missing network policies and risk the reconciliation error when the networking denies access by default
3. Have a separate mechanism and roll the pods when logging changes and connector operator is disabled
4. Ignore it and just document it

(Keep in mind that options 1 and 2 have backward compatibility issues and might break clusters that worked fine until now)

The same logic applies also to the list of available connector plugins. But obviously, that is a lot less serious problem compared to the logging configuration.

----

I guess in general, we should also think about the future of the connector operator ... I think we have many users using the REST API for one reason or another. So we cannot remove that mode. But should we enable the connector operator by default in our examples? Should we treat the disabled connector operator as a legacy mode we do not prefer and just ignore its limitations (i.e. go with the option 4)? Or do we want to fully support it?

scholzj:
Discussed on the community call on 18.4.: We should go with the option 3: `Have a separate mechanism and roll the pods when logging changes and connector operator is disabled`.

PR #9987, opened 2024-04-17
Separate logging configuration and collecting plugin information from the conenctor reconciliation

scholzj:
### Type of change

- Refactoring

### Description

The dynamic update of logging configuration (in both MM2 and Connect) as well as gathering of information about plugins (Connect only) is currently hidden inside the connector reconciliation. That might make some sense for Connect, since these things are done only when the connector operator is enabled. But overall it makes it hard to read the code and understand what is done where. It also prevents us from sharing the code for updating the logging configuration between Connect and MM2.

I also found out we never update logging when connector operator is disabled. This is tracked in #9986 and not addressed in this PR.

This PR separates these actions into separate steps to give better overview of what is happening. But it does not change the basic premise that for Connect, both logging configuration and plugins info gathering happens only with enabled Connector operator.

### Checklist

### DIFF
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/AbstractConnectOperator.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/AbstractConnectOperator.java
@@ -49,6 +49,7 @@ import io.strimzi.operator.common.BackOff;
 import io.strimzi.operator.common.Reconciliation;
 import io.strimzi.operator.common.ReconciliationLogger;
 import io.strimzi.operator.common.model.Labels;
+import io.strimzi.operator.common.model.OrderedProperties;
 import io.strimzi.operator.common.model.StatusDiff;
 import io.strimzi.operator.common.model.StatusUtils;
 import io.strimzi.operator.common.operator.resource.ReconcileResult;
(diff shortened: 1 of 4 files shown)


cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/AbstractConnectOperator.java
```diff
@@ -641,15 +664,15 @@ protected Future<ReconcileResult<ClusterRoleBinding>> connectInitClusterRoleBind
     // Static utility methods and classes
 
     private static Function<Map<String, Object>, Future<ConnectorStatusAndConditions>> createConnectorStatusAndConditions() {
-        return statusResult -> Future.succeededFuture(new ConnectorStatusAndConditions(statusResult, List.of(), List.of(), null));
+        return statusResult -> io.vertx.core.Future.succeededFuture(new ConnectorStatusAndConditions(statusResult, List.of(), List.of(), null));
```
ppatierno, line 667: why we needed the full package? (same for the next lines).

ppatierno (approved):
LGTM. I left just a nit.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/AbstractConnectOperator.java
scholzj: Strange, looks like IntelliJ decided to add them and I did not notice. Thanks for catching it.

katheris (approved):
LGTM

Issue #9231, opened 2023-10-11
[Enhancement]: kafka-exporter make offset lag for disconnected consumers configuratble


msillence:
### Related problem

currently offset.show-all is hardcoded to true

This means we collect stats on consumers that are no longer connected

We would like it to be configurable to turn it off

### Suggested solution

Add a new flag to allow turning off the offset.show-all

### Additional context

We are not interested in consumers that are no longer connected and can use other mechanisms to detect dead consumers if they are still required.

The default for expiring dead consumer groups is 1 week which means these stats hang around with the lag continually growing disrupting our ability to see real issues.

scholzj:
Discussed on the Community call on 19.10.2023: Seems reasonable. Should be implemented.

mabulgu:
Hi @scholzj . My friend @ilkerkocatepe  wants to work on this issue. I will try to guide him as we have been on a mentorship program recently. Would it be ok with you if he takes this one as it seems a good start issue?

scholzj:
@ilkerkocatepe this issue should be available, so feel free to work on it. I think the changes here would be very similar to #8682. So looking at that might be the best start. Feel free to let me know if you would have any questions.

PR #9991, opened 2024-04-17
docs(exporter): adds the showAllOffsets configuration to example Exporter config

PaulRMellor:
**Documentation**

Adds the showAllOffsets property to the example configuration of the Kafka Exporter.
The property is useful for stopping  collecting metrics on disconnected consumers, which can skew metrics collection.
Plus some minor rewrites to the section.

### Checklist

- [ ] Write tests
- [ ] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/documentation/modules/metrics/proc-metrics-kafka-deploy-options.adoc
+++ b/documentation/modules/metrics/proc-metrics-kafka-deploy-options.adoc
@@ -151,7 +151,7 @@ spec:
 <3> A regular expression to specify the topics to include in the metrics.
 <4> A regular expression to specify the consumer groups to exclude in the metrics.
 <5> A regular expression to specify the topics to exclude in the metrics.
-<6> By default, metrics are collected for all consumers regardless of their connection status. Setting `showAllOffsets` to `false` stops collecting metrics on disconnected consumers, preventing potentially misleading metrics from persisting and impairing the ability to detect lag in active consumers.
+<6> By default, metrics are collected for all consumers regardless of their connection status. Setting `showAllOffsets` to `false` stops collecting metrics on disconnected consumers.
 <7> CPU and memory resources to reserve.
 <8> Logging configuration, to log messages with a given severity (debug, info, warn, error, fatal) or above.
 <9> Boolean to enable Sarama logging, a Go client library used by Kafka Exporter.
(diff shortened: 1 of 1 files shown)


documentation/modules/metrics/proc-metrics-kafka-deploy-options.adoc
```diff
@@ -150,12 +151,13 @@ spec:
 <3> A regular expression to specify the topics to include in the metrics.
 <4> A regular expression to specify the consumer groups to exclude in the metrics.
 <5> A regular expression to specify the topics to exclude in the metrics.
-<6> CPU and memory resources to reserve.
-<7> Logging configuration, to log messages with a given severity (debug, info, warn, error, fatal) or above.
-<8> Boolean to enable Sarama logging, a Go client library used by Kafka Exporter.
-<9> Customization of deployment templates and pods.
-<10> Healthcheck readiness probes.
-<11> Healthcheck liveness probes.
+<6> By default, metrics are collected for all consumers regardless of their connection status. Setting `showAllOffsets` to `false` stops collecting metrics on disconnected consumers, preventing potentially misleading metrics from persisting and impairing the ability to detect lag in active consumers.
```
scholzj, line 154: Maybe we should stick to what it does?

```suggestion
<6> By default, metrics are collected for all consumers regardless of their connection status. Setting `showAllOffsets` to `false` stops collecting metrics on disconnected consumers.
```

Because I'm not sure I would say it is _potentially misleading metrics from persisting and impairing the ability to detect lag in active consumers_. I would instead warn about setting it to false because it _hides the problems with the growing lag of currently disconnected consumers_.

PaulRMellor: I thought the problem was that it was hardcoded to true? In the related [issue](https://github.com/strimzi/strimzi-kafka-operator/issues/9231), the disconnected consumers skew the metrics.

scholzj: I think that is a subject to opinion as to whether it being hardcoded to true was a problem or not. For someone apparently yes, but not for others. I would probably stick with what it does and let everyone decide.

scholzj (approved):

ppatierno (approved):
LGTM.

PR #10005, opened 2024-04-19
Don't fail reconciliation when Manual Rolling Update fails

Staniel:
### Type of change

_Select the type of your PR_

- Enhancement / new feature

### Description

Implements https://github.com/strimzi/proposals/blob/main/070-dont-fail-reconciliation-in-manual-rolling-update.md
### Checklist

- [x] Write tests
- [ ] Update documentation
- [ ] Update CHANGELOG.md


cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/AbstractConnectOperator.java
```diff
@@ -216,7 +220,15 @@ protected Future<Void> manualRollingUpdate(Reconciliation reconciliation, KafkaC
                     if (!podNamesToRoll.isEmpty())  {
                         // There are some pods to roll
                         KafkaConnectRoller roller = new KafkaConnectRoller(reconciliation, connect, operationTimeoutMs, podOperations);
-                        return roller.maybeRoll(podNamesToRoll, pod -> RestartReasons.of(RestartReason.MANUAL_ROLLING_UPDATE));
+                        return roller.maybeRoll(podNamesToRoll, pod -> RestartReasons.of(RestartReason.MANUAL_ROLLING_UPDATE))
+                            .recover(error -> {
+                                if (featureGates.continueReconciliationOnManualRollingUpdateFailureEnabled()) {
+                                    LOGGER.warnCr(reconciliation, "Reconciliation will be continued even though manual rolling update failed");
```
scholzj, line 226: I think you should log the error which was raised. Maybe something like:

 ```suggestion
                                    LOGGER.warnCr(reconciliation, "Manual rolling update failed (reconciliation will be continued)", error);
```

The same would apply to the other two places.

File: cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/AbstractConnectOperator.java [hunk 2]
```diff
@@ -87,6 +88,8 @@ public abstract class AbstractConnectOperator<C extends KubernetesClient, T exte
     private static final ReconciliationLogger LOGGER = ReconciliationLogger.create(AbstractConnectOperator.class.getName());
 
     private final boolean isNetworkPolicyGeneration;
+    private final FeatureGates featureGates;
```
scholzj, line 91: Maybe you can keep only boolean and not the whole feature gates class? Something like `boolean continueOnManualRUFauilure`. Again, this might be good in all 3 classes.

cluster-operator/src/main/java/io/strimzi/operator/cluster/FeatureGates.java
```diff
@@ -17,9 +17,12 @@ public class FeatureGates {
     /* test */ static final FeatureGates NONE = new FeatureGates("");
 
     private static final String USE_KRAFT = "UseKRaft";
+    private static final String CONTINUE_RECONCILIATION_ON_MANUAL_ROLLING_UPDATE_FAILURE = "ContinueReconciliationOnManualRollingUpdateFailure";
 
     // When adding new feature gates, do not forget to add them to allFeatureGates() and toString() methods
     private final FeatureGate useKRaft = new FeatureGate(USE_KRAFT, true);
+    private final FeatureGate continueReconciliationOnManualRollingUpdateFailure =
```
scholzj, line 24: I wonder if we should try to shorten it up a bit. We should keep the feature gate name as is (`ContinueReconciliationOnManualRollingUpdateFailure`). But what about using `CONTINUE_ON_MANUAL_RU_FAILURE` and `continueOnManualRUFailure` for the variable and the related method? It will make it a bit easier to read the code and I think it would be still readable.

cluster-operator/src/test/java/io/strimzi/operator/cluster/FeatureGatesTest.java
```diff
@@ -64,12 +64,12 @@ public void testFeatureGatesParsing() {
...
-        //assertThat(new FeatureGates("-UseKRaft,-KafkaNodePools").kafkaNodePoolsEnabled(), is(false));
-        //assertThat(new FeatureGates("  +UseKRaft    ,    +KafkaNodePools").useKRaftEnabled(), is(true));
-        //assertThat(new FeatureGates("  -UseKRaft    ,    -KafkaNodePools").kafkaNodePoolsEnabled(), is(false));
-        //assertThat(new FeatureGates("+KafkaNodePools,-UseKRaft").useKRaftEnabled(), is(false));
-        //assertThat(new FeatureGates("+KafkaNodePools,-UseKRaft").kafkaNodePoolsEnabled(), is(true));
+        assertThat(new FeatureGates("-UseKRaft,-ContinueReconciliationOnManualRollingUpdateFailure").useKRaftEnabled(), is(false));
+        assertThat(new FeatureGates("-UseKRaft,-ContinueReconciliationOnManualRollingUpdateFailure").continueReconciliationOnManualRollingUpdateFailureEnabled(), is(false));
+        assertThat(new FeatureGates("+UseKRaft,+ContinueReconciliationOnManualRollingUpdateFailure").useKRaftEnabled(), is(true));
+        assertThat(new FeatureGates("+UseKRaft,+ContinueReconciliationOnManualRollingUpdateFailure").continueReconciliationOnManualRollingUpdateFailureEnabled(), is(true));
```
scholzj, line 70: These tests are good -> but can we keep the original whitespaces? That is the part they were supposed to test.

cluster-operator/src/test/java/io/strimzi/operator/cluster/operator/assembly/KafkaAssemblyOperatorManualRollingUpdatesTest.java
```diff
@@ -429,6 +430,156 @@ vertx, new PlatformFeaturesAvailability(false, KUBERNETES_VERSION),
...
+                    // Verify Kafka rolling updates
+                    assertThat(kr.maybeRollKafkaInvocations, is(1));
+                    assertThat(kr.kafkaNodesNeedRestart.size(), is(2));
+                    assertThat(kr.kafkaNodesNeedRestart, is(List.of("my-cluster-kafka-0", "my-cluster-kafka-1")));
+                    assertThat(kr.kafkaRestartReasons.apply(podWithName("anyName")), is(RestartReasons.of(RestartReason.MANUAL_ROLLING_UPDATE)));
+
+                    async.flag();
+                })));
+        }
+    }
```
scholzj, line 582: Maybe you could keep an empty line between the different tests?

scholzj (commented):
Thanks for the PR. The code looks good. I left only some smaller comments. Would it be worth to also add some unit tests for Connect / Mirror Maker 2? They have already some maunal rolling update tests in `KafkaMirrorMaker2AssemblyOperatorPodSetTest.java` and `KafkaConnectAssemblyOperatorPodSetTest.java`

scholzj:
One more thing ... could you also enable the feature gate in these two files:
* https://github.com/strimzi/strimzi-kafka-operator/blob/main/.azure/templates/jobs/system-tests/feature_gates_regression_jobs.yaml
* https://github.com/strimzi/strimzi-kafka-operator/blob/main/.azure/templates/jobs/system-tests/feature_gates_regression_namespace_rbac_jobs.yaml

in the `strimzi_feature_gates` parameter => that would allow us to run the tests with the feature gate enabled and see f that works as well. Thanks.

cluster-operator/src/test/java/io/strimzi/operator/cluster/operator/assembly/KafkaAssemblyOperatorManualRollingUpdatesTest.java
Staniel: There is already an empty line between each new test methods.

scholzj: Yeah, but not between 582 and 583, or?

Staniel:
> Thanks for the PR. The code looks good. I left only some smaller comments. Would it be worth to also add some unit tests for Connect / Mirror Maker 2? They have already some maunal rolling update tests in `KafkaMirrorMaker2AssemblyOperatorPodSetTest.java` and `KafkaConnectAssemblyOperatorPodSetTest.java`

I looked at it but didn't find an easy way to simulate a manual roll failure, since it was not using mock operator. Any suggestion?

scholzj:
> I looked at it but didn't find an easy way to simulate a manual roll failure, since it was not using mock operator. Any suggestion?

Maybe you have to write one. You could also try to trigger the error through some mocked method. Maybe if you throw here an error instead of success it would cause an error you can use?
```
when(mockPodOps.deleteAsync(any(), eq(NAMESPACE), startsWith(COMPONENT_NAME), eq(false))).thenReturn(Future.succeededFuture());
```

Let's look sta it next week. Otherwise it looks good.

cluster-operator/src/test/java/io/strimzi/operator/cluster/operator/assembly/KafkaAssemblyOperatorManualRollingUpdatesTest.java
```diff
@@ -429,6 +430,157 @@ vertx, new PlatformFeaturesAvailability(false, KUBERNETES_VERSION),
                 })));
     }
 
+    @Test
+    public void testManualPodRollingUpdateWithPodSetsWithError1(VertxTestContext context) {
```
katheris, line 434: These are effectively parametrized tests, so it might be worth using the actual annotation, it would save having lots of tests with the same name but different numbers at the end. https://junit.org/junit5/docs/current/user-guide/#writing-tests-parameterized-tests-sources-MethodSource

katheris (approved):
LGTM, just had one comment about the tests.

scholzj:
@Staniel I added the unit tests so that we can move forward with this and unblock some other work that would conflict with this.

@ppatierno Could you please review this?

scholzj (approved):

ppatierno (approved):
LGTM. Thanks!

PR #10006, opened 2024-04-22
Define limits and requests schema in CRDs

ilpianista:
### Type of change

Define schema for `limits` and `requests` fields.

- Enhancement / new feature

### Description

I manage Strimzi via terraform and, without schema information, the terraform provider is forced to replace the resource in order to satisfy Terraform's requirements of resource structure invariance.

### Checklist

- [ ] Write tests
- [ ] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/helm-charts/helm3/strimzi-kafka-operator/crds/040-Crd-kafka.yaml
+++ b/helm-charts/helm3/strimzi-kafka-operator/crds/040-Crd-kafka.yaml
@@ -810,10 +810,26 @@ spec:
                               name:
                                 type: string
                         limits:
-                          x-kubernetes-preserve-unknown-fields: true
+                          additionalProperties:
+                            anyOf:
+                            - type: integer
+                            - type: string
+                            pattern: ^(\+|-)?(([0-9]+(\.[0-9]*)?)|(\.[0-9]+))(([KMGTPE]i)|[numkMGTPE]|([eE](\+|-)?(([0-9]+(\.[0-9]*)?)|(\.[0-9]+))))?$
+                            x-kubernetes-int-or-string: true
+                          description: 'Limits describes the maximum amount of compute
+                            resources allowed. More info: https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/'
                           type: object
                         requests:
-                          x-kubernetes-preserve-unknown-fields: true
+                          additionalProperties:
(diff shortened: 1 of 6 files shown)


scholzj (commented):
Thanks for the PR. However:
* You should not change anything in the `helm-charts` directory. That is changed only as part of an release. Any changes should be done in the `packaging` folder.
* The CRD are generated from the Java code in the `api` module. So you cannot change them directly anyway.

ilpianista:
> Thanks for the PR. However:
> 
>     * You should not change anything in the `helm-charts` directory. That is changed only as part of an release. Any changes should be done in the `packaging` folder.
> 
>     * The CRD are generated from the Java code in the `api` module. So you cannot change them directly anyway.

Ouch. Thanks :-D I try to reflect my changes there.

scholzj:
I think this might have been already done by someone, it is just not released. Let's hope it works better than the last changes done for _Terraform_ and we will not need to revert it because of breaking things for others. You should raise this issue with Terraform in the first place to fix how it behaves.

ilpianista:
> I think this might have been already done by someone, it is just not released.

Oh, I should I wait then? At the moment I'm modifying the CrdGenerator class by adding a check on `ResourceRequirements` since that class comes from fabric8.

> Let's hope it works better than the last changes done for _Terraform_ and we will not need to revert it because of breaking things for others. You should raise this issue with Terraform in the first place to fix how it behaves.

Problem is someone already [did](https://github.com/hashicorp/terraform-provider-kubernetes/issues/1841#issuecomment-1254825692) and the short story is that `x-kubernetes-preserve-unknown-fields: true` prevents them to follow the schema.

scholzj:
> Problem is someone already https://github.com/hashicorp/terraform-provider-kubernetes/issues/1841#issuecomment-1254825692 and the short story is that x-kubernetes-preserve-unknown-fields: true prevents them to follow the schema.

No, the short story is that there is no reason for that.

scholzj:
Discussed on the community call on 2.5.2024: This already seems to be implemented and should be closed. If you have anything more, please reopen it or open a new PR when it is ready.

ilpianista:
> This already seems to be implemented and should be closed.

Cool, thanks! Any chance to get a link to the commit/MR?

PR #10010, opened 2024-04-22
Remove events rule from entity-operator

fvaleri:
This rule was only needed by the old implementation of the Topic Operator that has recently been removed.

### DIFF
--- a/cluster-operator/src/main/resources/cluster-roles/020-ClusterRole-strimzi-cluster-operator-role.yaml
+++ b/cluster-operator/src/main/resources/cluster-roles/020-ClusterRole-strimzi-cluster-operator-role.yaml
@@ -88,7 +88,7 @@ rules:
   - apiGroups:
       - "events.k8s.io" # new events api, used by cluster operator
     resources:
-      # The cluster operator needs to be able to create events and delegate permissions to do so
+      # The cluster operator needs to be able to create events
       - events
     verbs:
       - create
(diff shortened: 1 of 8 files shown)


scholzj (commented):
What about `020-ClusterRole-strimzi-cluster-operator-role.yaml`? Shouldn't you remove the core API events from there as well? I think CO itself needs only the `events.k8s.io` IIRC.

cluster-operator/src/main/resources/cluster-roles/020-ClusterRole-strimzi-cluster-operator-role.yaml
```diff
@@ -86,7 +86,6 @@ rules:
       - patch
       - update
   - apiGroups:
-      - "" # legacy core events api, used by topic operator
       - "events.k8s.io" # new events api, used by cluster operator
     resources:
       # The cluster operator needs to be able to create events and delegate permissions to do so
```
scholzj, line 91: Does it still delegate the events? Or are they now used only in the CO?

fvaleri: Both TO and UO do not create any event, so delegation is not required. Do you want me to update the comment?

scholzj: Yeah, I guess that would be perfect. Thanks.

scholzj (commented):
The change looks good, but you need to run `make helm_install` once again 😉

scholzj (approved):

ppatierno (approved):
LGTM

fvaleri:
I think failed tests are unrelated.

Issue #9692, opened 2024-02-15
[KRaft]: Allow KafkaRoller directly connect to controllers 


tinaselenge:
### Related problem

KafkaRoller currently connects to the controller via brokers to get the quorum health information. 

### Suggested solution

With Strimzi supporting Kafka 3.7 which includes [KIP 919](https://cwiki.apache.org/confluence/display/KAFKA/KIP-919%3A+Allow+AdminClient+to+Talk+Directly+with+the+KRaft+Controller+Quorum+and+add+Controller+Registration), we should create a service for controller pods and make KafkaRoller connect to it for quorum information.

scholzj:
Triaged on the community call on 22.2.2024: This should be done once Kafka 3.7 is released. Unless the implementation is trivial, it should have a proposal.

tinaselenge:
When working on this, I discovered that we need to configure controller listeners with hostnames in order for Admin client to bootstrap. This was not clearly stated in the KIP, so I raised an upstream [issue](https://issues.apache.org/jira/browse/KAFKA-16612). 

Reconfiguring the controller listeners do work and allows the operator to talk to the controller directly. I have raised a draft PR #10016. However, this causes controller pods to crash once or twice until the DNS is updated. When a pod gets terminated, pod's IP address changes and the DNS needs to be updated to resolve the hostnames to the new IP address. It looks like controller start up is much quicker than normal brokers, therefore binding happens before the DNS is updated which results in a controller pod crashing a couple of times before it runs ok. 

Despite the crash, the controller pod eventually runs and works fine but this is the side effect we have currently. It's not clear how or whether it would get fixed in Kafka upstream.

tinaselenge:
This issue is blocked until https://issues.apache.org/jira/browse/KAFKA-16781 is completed.

tinaselenge:
We currently cannot update logging configuration dynamically for controllers because we cannot talk to controllers directly. If the logging config is not dynamic, the controllers would not be rolled either.  Looks like the user need to trigger manual rolling update for controllers to change the log level. This can be solved once we can talk to the controller directly. However, in the meantime, should this be opened as a separate issue to track it and potentially add a workaround similar to how we handle controller config changes (we extract the controller-relevant configurations and use it in the configuration annotations)? 
(cc @scholzj)

scholzj:
[...]

PR #10016, opened 2024-04-23
Allow KafkaRoller talk to controller directly

tinaselenge:
### Type of change

_Select the type of your PR_

- Enhancement / new feature

### Description

- Add a method for creating admin client for controllers with BOOTSTRAP_CONTROLLERS_CONFIG set. 
- Make KafkaRoller create admin client against controllers nodes, if they are running 3.9.0 or later. 
- Tidy up: remove ceShouldBeFatal option as it's never set to true.

There will be a follow up PR to dynamically apply configurations for controllers (with version 3.9.0 or later) using the admin client instead of how we currently restart controllers when any config changes. 

Closes https://github.com/strimzi/strimzi-kafka-operator/issues/9692 

### Checklist

- [ ] Write tests
- [ ] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/resource/KafkaRoller.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/resource/KafkaRoller.java
@@ -20,6 +20,7 @@ import io.strimzi.operator.cluster.operator.VertxUtil;
 import io.strimzi.operator.cluster.operator.resource.events.KubernetesRestartEventPublisher;
 import io.strimzi.operator.cluster.operator.resource.kubernetes.PodOperator;
 import io.strimzi.operator.common.AdminClientProvider;
+import io.strimzi.operator.common.Annotations;
 import io.strimzi.operator.common.BackOff;
 import io.strimzi.operator.common.Reconciliation;
 import io.strimzi.operator.common.ReconciliationLogger;
@@ -208,14 +209,14 @@ public class KafkaRoller {
      * Initializes controllerAdminClient if it has not been initialized yet
      * @return true if the creation of AC succeeded, false otherwise
      */
-    private boolean maybeInitControllerAdminClient() {
+    private boolean maybeInitControllerAdminClient(String currentVersion) {
         if (this.controllerAdminClient == null) {
             // Prior to 3.9.0, Kafka did not support directly connecting to controllers nodes
             // via Kafka Admin API when running in KRaft mode.
             // Therefore, use brokers to initialise adminClient for quorum health check
             // when the version is older than 3.9.0.
             try {
-                if (KafkaVersion.compareDottedVersions(kafkaVersion.version(), "3.9.0") >= 0) {
+                if (currentVersion != null && KafkaVersion.compareDottedVersions(currentVersion, "3.9.0") >= 0) {
                     this.controllerAdminClient = controllerAdminClient(nodes);
                 } else {
                     this.controllerAdminClient = brokerAdminClient(Set.of());
@@ -454,9 +455,11 @@ public class KafkaRoller {
         // change and the desired roles still apply.
         boolean isBroker = Labels.booleanLabel(pod, Labels.STRIMZI_BROKER_ROLE_LABEL, nodeRef.broker());
         boolean isController = Labels.booleanLabel(pod, Labels.STRIMZI_CONTROLLER_ROLE_LABEL, nodeRef.controller());
+        // This is relevant when creating admin client for controllers
+        String currentVersion = pod.getMetadata().getAnnotations().get(Annotations.ANNO_STRIMZI_KAFKA_VERSION);
 
         try {
-            checkIfRestartOrReconfigureRequired(nodeRef, isController, isBroker, restartContext);
+            checkIfRestartOrReconfigureRequired(nodeRef, isController, isBroker, restartContext, currentVersion);
             if (restartContext.forceRestart) {
                 LOGGER.debugCr(reconciliation, "Pod {} can be rolled now", nodeRef);
                 restartAndAwaitReadiness(pod, operationTimeoutMs, TimeUnit.MILLISECONDS, restartContext);
@@ -586,7 +589,7 @@ public class KafkaRoller {
      * Determine whether the pod should be restarted, or the broker reconfigured.
      */
     @SuppressWarnings("checkstyle:CyclomaticComplexity")
-    private void checkIfRestartOrReconfigureRequired(NodeRef nodeRef, boolean isController, boolean isBroker, RestartContext restartContext) throws ForceableProblem, InterruptedException, FatalProblem, UnforceableProblem {
+    private void checkIfRestartOrReconfigureRequired(NodeRef nodeRef, boolean isController, boolean isBroker, RestartContext restartContext, String currentVersion) throws ForceableProblem, InterruptedException, FatalProblem, UnforceableProblem {
         RestartReasons reasonToRestartPod = restartContext.restartReasons;
         if (restartContext.podStuck && !reasonToRestartPod.contains(RestartReason.POD_HAS_OLD_REVISION)) {
             // If the pod is unschedulable then deleting it, or trying to open an Admin client to it will make no difference
@@ -609,8 +612,8 @@ public class KafkaRoller {
 
         // if it is a pure controller, initialise the admin client specifically for controllers
         if (isController && !isBroker) {
-            if (!maybeInitControllerAdminClient()) {
-                handleFailedAdminClientForController(nodeRef, restartContext, reasonToRestartPod);
+            if (!maybeInitControllerAdminClient(currentVersion)) {
+                handleFailedAdminClientForController(nodeRef, restartContext, reasonToRestartPod, currentVersion);
                 return;
             }
             restartContext.quorumCheck = quorumCheck(controllerAdminClient, nodeRef);
@@ -674,8 +677,8 @@ public class KafkaRoller {
         restartContext.brokerLoggingDiff = brokerLoggingDiff;
     }
 
-    private void handleFailedAdminClientForController(NodeRef nodeRef, RestartContext restartContext, RestartReasons reasonToRestartPod) throws UnforceableProblem {
-        if (KafkaVersion.compareDottedVersions(kafkaVersion.version(), "3.9.0") >= 0) {
+    private void handleFailedAdminClientForController(NodeRef nodeRef, RestartContext restartContext, RestartReasons reasonToRestartPod, String currentVersion) throws UnforceableProblem {
+        if (currentVersion != null && KafkaVersion.compareDottedVersions(currentVersion, "3.9.0") >= 0) {
             // If the version supports talking to controllers, force restart this pod when the admin client cannot be initialised.
             // There is no reason to continue as we won't be able to connect an admin client to this pod for other checks later.
             LOGGER.infoCr(reconciliation, "KafkaQuorumCheck cannot be initialised for {} because none of the controllers do not seem to responding to connection attempts.", nodeRef);
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/resource/KafkaRoller.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/resource/KafkaRoller.java
@@ -35,10 +35,8 @@ import org.apache.kafka.clients.admin.AlterConfigOp;
 import org.apache.kafka.clients.admin.AlterConfigsResult;
 import org.apache.kafka.clients.admin.Config;
 import org.apache.kafka.clients.admin.DescribeClusterResult;
-import org.apache.kafka.common.KafkaException;
 import org.apache.kafka.common.KafkaFuture;
 import org.apache.kafka.common.Node;
-import org.apache.kafka.common.config.ConfigException;
 import org.apache.kafka.common.config.ConfigResource;
 import org.apache.kafka.common.errors.SslAuthenticationException;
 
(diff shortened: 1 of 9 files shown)


cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaBrokerConfigurationBuilder.java
```diff
@@ -350,12 +348,15 @@ public KafkaBrokerConfigurationBuilder withListeners(
...
         if (!isKraftControllerOnly) {
-            writer.println("advertised.listeners=" + String.join(",", advertisedListeners));
+            // Join broker and controller listeners.
+            // The controller listeners is not empty, if it's a combined node, or broker in ZooKeeper mode or in migration state but not when full KRaft.
+            brokerAdvertisedListeners.addAll(controllerAdvertisedListeners);
+            writer.println("advertised.listeners=" + String.join(",", brokerAdvertisedListeners));
             writer.println("inter.broker.listener.name=" + REPLICATION_LISTENER_NAME);
         } else if (node.controller() && kafkaMetadataConfigState.isZooKeeperToPostMigration()) {
             // needed for KRaft controller only as well until post-migration because it needs to contact brokers
+            writer.println("advertised.listeners=" + String.join(",", controllerAdvertisedListeners));
```
scholzj, line 359: I know this will be used in Kafka 3.9.0. However, it is a forbidden option in the previous Kafka versions and it caused the Kafka node to stop. Unless that changed already in Kafka 3.8.0 to not have it exit when this option is set, you will need to gate it by the Kafka version.

(And assuming it needs the gating, we maybe should consider moving it into a separate PR as it will be needed for any kind of the Kafka 3.9.0 support)

tinaselenge: You are right, I've just realised the changes for removing the restriction of advertised.listeners configuration didn't get included in 3.8.0 (didn't make it before the code freeze). When 3.9.0 is released and supported with Strimzi, either way, this option would need to be gated by the Kafka version, as we would not be able to configure it for the  older versions (3.8.0 and 3.7.0) that are still supported with Strimzi right?

scholzj: I think we will drop 3.7 when 3.8 is added, not 100% sure yet given the strange release cycles. In any case 3.8.0 will be definitely still supported.

tinaselenge:
I will check the regression test failures.

tinaselenge:
@scholzj can you please kick off the regression tests again? The failed system tests seem to pass for me when running locally.

Frawless:
@tinaselenge could you please rebase to avoid running `testCertifiactes` ST that was removed here https://github.com/strimzi/strimzi-kafka-operator/pull/10830 ?

tinaselenge:
thanks @Frawless . I rebased now.

scholzj:
I started the regressions. But the Unit test failure seems related to your PR.

tinaselenge:
The regression tests passed and then I pushed a fix for the failed unit tests.

Frawless:
> The regression tests passed and then I pushed a fix for the failed unit tests.

The regression tests failed - https://dev.azure.com/cncf/strimzi/_build/results?buildId=181799&view=results (passed only for ZK mode - https://dev.azure.com/cncf/strimzi/_build/results?buildId=181800&view=results) and because the previous run for different PR passed I think it could be connected to your changes.

tinaselenge:
Maybe I pushed too soon and it got cancelled? Can please we kick off the regression tests again then?

scholzj:
@tinaselenge 2 of the pipelines timeout after 6 hours. They were not canceled.

scholzj:
> Some of SecurityIT tests failed due to CaReconciler passing a null KafkaVersion to KafkaRoller. With this PR, the KafkaRoller now checks if kafka version is older than 3.9.0 before initialising an admin client to controller nodes. We would have to make CaReconciler pass a valid KafkaVersion. I'm currently testing this locally and will push the change when ready.

I think this needs to be thought through ...
* Should we wait for when we support only 3.9 and 4.0?
* How does it handle upgrades?

scholzj:
Also, what does this PR actually do? What is the connection used for? To reconfigure controllers? Check that they are in-sync?

tinaselenge:
> I think this needs to be thought through ...
Should we wait for when we support only 3.9 and 4.0?
How does it handle upgrades?

I think waiting until we support only 3.9 would certainly make this changes much simpler since we won't need to check Kafka versions and worry about the upgrades. Thinking about this more, actually we don't need to make any changes to CaReconciler.  Instead, we need to check the kafkaVersion from a controller pod itself rather using the kafkaVersion passed to the KafkaRoller.

> Also, what does this PR actually do? What is the connection used for? To reconfigure controllers? Check that they are in-sync?

Yes, this PR makes KafkaRoller talk to controllers so that we can reconfigure controllers (I was planning to make the actual controller reconfigurations changes in a separate PR). Also admin client is used as a connection check and if the roller cannot connect to a node, it force restarts the node. This connection check currently not done for controllers. 

However, given the amount of changes this requires, some of which would need to reverted once only 3.9+ is supported, I'm leaning towards parking this for now. 
What do people think?

scholzj:
> I think waiting until we support only 3.9 would certainly make this changes much simpler since we won't need to check Kafka versions and worry about the upgrades.

So, even if you wait for us to support Kafka 3.9 and newer only, we would generally still need to worry about upgrades as someone can do upgrade from e.g. Strimzi 0.44 / Kafka 3.7 to Strimzi 0.46/Kafka 4.0. This was at least so far expected to be supported. So if we would want to change this and require users to run Strimzi 0.45 / Kafka 3.9 as a prerequisite to upgrading further, we would need to discuss it more.

> Yes, this PR makes KafkaRoller talk to controllers so that we can reconfigure controllers (I was planning to make the actual controller reconfigurations changes in a separate PR). Also admin client is used as a connection check and if the roller cannot connect to a node, it force restarts the node. This connection check currently not done for controllers.

That would be fine ... but worth mentioning in the description to make it clear what to expect.

> However, given the amount of changes this requires, some of which would need to reverted once only 3.9+ is supported, I'm leaning towards parking this for now.
What do people think?

I'm slightly confused if this was meant for the controller reconfiguration or the whole thing.

tinaselenge:
@scholzj I made an update so that we check the current version of the node via the annotation on pod instead of the kafkaVersion passed from the reconciler. This way, we only initialise the admin client for controllers based on the current version rather than desired version. I think this should work for upgrades (I tested in local cluster) and the changes are done only for KafkaRoller.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/resource/KafkaRoller.java
```diff
@@ -454,9 +455,11 @@ private void restartIfNecessary(NodeRef nodeRef, RestartContext restartContext)
         // change and the desired roles still apply.
         boolean isBroker = Labels.booleanLabel(pod, Labels.STRIMZI_BROKER_ROLE_LABEL, nodeRef.broker());
         boolean isController = Labels.booleanLabel(pod, Labels.STRIMZI_CONTROLLER_ROLE_LABEL, nodeRef.controller());
+        // This is relevant when creating admin client for controllers
+        String currentVersion = pod.getMetadata().getAnnotations().get(Annotations.ANNO_STRIMZI_KAFKA_VERSION);
```
scholzj, line 459: Do we know that `pod` is not null here? (you can also use `Annotations.stringAnnotation(...)` to extract it with null checks, but the resource has to be non-null already)

scholzj (commented):
Thanks. I left some questions / nits.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/resource/KafkaRoller.java
```diff
@@ -905,34 +906,48 @@ protected Future<Void> restart(Pod pod, RestartContext restartContext) {
...
         try {
             LOGGER.debugCr(reconciliation, "Creating AdminClient for {}", bootstrapHostnames);
             return adminClientProvider.createAdminClient(bootstrapHostnames, coTlsPemIdentity.pemTrustSet(), coTlsPemIdentity.pemAuthIdentity());
-        } catch (KafkaException e) {
-            if (ceShouldBeFatal && (e instanceof ConfigException
-                    || e.getCause() instanceof ConfigException)) {
-                throw new FatalProblem("An error while try to create an admin client with bootstrap brokers " + bootstrapHostnames, e);
-            } else {
-                throw new ForceableProblem("An error while try to create an admin client with bootstrap brokers " + bootstrapHostnames, e);
-            }
```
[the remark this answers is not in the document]
tinaselenge: ceShouldBeFatal is set to false everywhere so we actually never throw FatalProblem. If we should throw FatalProblem on ConfigException (which we don't currently),  then we should do that regardless of ceShouldBeFatal.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/resource/KafkaRoller.java
tinaselenge: Yes on line 416, we check if pod is null and if it is, we return.

operator-common/src/main/java/io/strimzi/operator/common/Annotations.java
```diff
@@ -34,6 +34,10 @@ public class Annotations extends ResourceAnnotations {
      * annotation we force a restart of the pods managed by StrimziPodSets or Deployments.
      */
     public static final String ANNO_STRIMZI_LOGGING_HASH = STRIMZI_DOMAIN + "logging-hash";
+    /**
+     * Annotation for current Kafka Version
+     */
+    public static final String ANNO_STRIMZI_KAFKA_VERSION = STRIMZI_DOMAIN + "kafka-version";
```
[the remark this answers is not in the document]
tinaselenge: Yes, I agree. I missed that.

operator-common/src/test/java/io/strimzi/operator/common/DefaultAdminClientProviderTest.java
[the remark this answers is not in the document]
tinaselenge: You are right. We previously did not need to test the public method, as it was just calling the internal method. In order to test the public methods, I would need to refactor this test with mocks, as it currently just checks the properties returned from the internal method.

scholzj (approved):
I'm trying to run the STs on this with Kafka 3.8.1. But assuming they pass, this looks good to me. Thanks @tinaselenge.

tinaselenge:
Thank you @scholzj!

operator-common/src/test/java/io/strimzi/operator/common/DefaultAdminClientProviderTest.java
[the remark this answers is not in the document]
tinaselenge: You are right, now public methods have an additional logic, so we probably need to test them but for that I needed to make the adminClientConfiguration() method non static, so that I can mock the class and check the config passed to it. I will push this in a commit. Please let me know what you think. Thanks Jakub!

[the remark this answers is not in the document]
tinaselenge: If you think we shouldn't make this method non static, I can revert the commit. It just tests the logic of different bootstrap config being passed.

operator-common/src/main/java/io/strimzi/operator/common/AdminClientProvider.java
```diff
@@ -26,7 +26,17 @@ public interface AdminClientProvider {
...
     /**
-     * Create a Kafka Admin interface instance
+     * Create a Kafka Admin interface instance for controllers
+     *
+     * @param controllerBootstrapHostnames Kafka controller hostname to connect to for administration operations
+     * @param kafkaCaTrustSet Trust set for connecting to Kafka
+     * @param authIdentity Identity for TLS client authentication for connecting to Kafka
+     * @return Instance of Kafka Admin interface
+     */
+    Admin createControllerAdminClient(String controllerBootstrapHostnames, PemTrustSet kafkaCaTrustSet, PemAuthIdentity authIdentity);
```
ppatierno, line 36: I didn't get why we need new methods for the admin client provider if the creation of an admin client for brokers or controllers differs just for the bootstrap parameter value. Also the corresponding implementation looks exactly the same comparing `createAdminClient` and `createControllerAdminClient`. What am I missing here?

operator-common/src/test/java/io/strimzi/operator/common/DefaultAdminClientProviderTest.java
```diff
@@ -97,10 +103,35 @@ public void testMTlsWithPublicCAConnection() {
...
+    public void testCreateBrokerAdminClient() {
+        DefaultAdminClientProvider defaultAdminClientProvider = mock(DefaultAdminClientProvider.class);
+        when(defaultAdminClientProvider.adminClientConfiguration(any(), any(), any())).thenAnswer(invocation -> {
+            Properties config = invocation.getArgument(2);
+            assertThat(config.size(), is(1));
+            assertThat(config.getProperty(AdminClientConfig.BOOTSTRAP_SERVERS_CONFIG), is("my-kafka:9092"));
+            return config;
+        });
+        defaultAdminClientProvider.createAdminClient("my-kafka:9092", null, null, new Properties());
+    }
```
scholzj, line 133: I think this is an imaginative ay to test it which I kind of like. But you should probably have some kind of CountDownLatch there or something to actually make sure the asserts were called?

tinaselenge: You are right. I think I need to use slightly different approach to make sure the asserts are called. I've updated the last commit I pushed.

scholzj: Thanks.

operator-common/src/main/java/io/strimzi/operator/common/AdminClientProvider.java
ppatierno: Ah right there is a different admin client config to set. Let's take the separated methods.

ppatierno (approved):

scholzj:
Thanks for the PR @tinaselenge

PR #10017, opened 2024-04-23
Improve JBOD storage version validation in KRaft

scholzj:
### Type of change

- Bugfix

### Description

This PR improves the Kafka version validation when JBOD storage is used in KRaft:
* It validates the `from` version instead of the `to` version to better handle upgrades (not allow JBOD storage before the upgrade to 3.7.0 is done)
* Checks that the metadata version is set to `3.7`

### Checklist

- [x] Write tests


ppatierno (approved):
LGTM. Assuming even migration pipeline is green.

PR #10023, opened 2024-04-24
Remove unneeded state (old certificate secrets) from the `Ca` classes

scholzj:
### Type of change

- Refactoring

### Description

The `Ca` classes currently hold the old certificate secrets as a state through the whole reconciliation. This increases memory consumption but could be also misleading as they might not be up to date and might contain old certificates. 

They seem to be used only in two places. One right after the state is set and one later in the reconciliation where we anyway query the secret again. So it seems to be easy to remove it and pass them to the methods directly as done in this PR.

This should also make it easier to split the one big Secret with all Kafka node certificates

### Checklist

### DIFF
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/model/CruiseControl.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/model/CruiseControl.java
@@ -495,22 +495,22 @@ public class CruiseControl extends AbstractModel implements SupportsMetrics, Sup
 
     /**
      * Generate the Secret containing the Cruise Control certificate signed by the cluster CA certificate used for TLS based
-     * internal communication with Kafka
-     * It also contains the related Cruise Control private key.
+     * internal communication with Kafka. It also contains the related Cruise Control private key.
      *
-     * @param namespace Namespace in which the Cruise Control cluster runs
-     * @param kafkaName Name of the Kafka cluster (it is used for the SANs in the certificate)
-     * @param clusterCa The cluster CA.
-     * @param isMaintenanceTimeWindowsSatisfied Indicates whether we are in the maintenance window or not.
-     *                                          This is used for certificate renewals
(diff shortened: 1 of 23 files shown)


katheris (approved):
LGTM

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/CruiseControl.java
```diff
@@ -495,22 +495,22 @@ public Secret generateApiSecret(PasswordGenerator passwordGenerator, Secret oldS
...
+     * @param clusterName                           Name of the Kafka cluster (it is used for the SANs in the certificate)
+     * @param clusterCa                             The cluster CA.
+     * @param existingSecret                        The existing secret with Kafka certificates
+     * @param isMaintenanceTimeWindowsSatisfied     Indicates whether we are in the maintenance window or not.
+     *                                              This is used for certificate renewals
      *
      * @return The generated Secret.
      */
-    public Secret generateCertificatesSecret(String namespace, String kafkaName, ClusterCa clusterCa, boolean isMaintenanceTimeWindowsSatisfied) {
+    public Secret generateCertificatesSecret(String namespace, String clusterName, ClusterCa clusterCa, Secret existingSecret, boolean isMaintenanceTimeWindowsSatisfied) {
```
ppatierno, line 509: I see the method doing the same stuff in other classes is just called `generateSecret` (i.e. EntityTopicOperator) while `generateCertificatesSecret` in others (i.e. here and KafkaCluster). Would you mind having a consistent name? Wdyt? I would be to have `generateCertificatesSecret` on all.

ppatierno (approved):
LGTM. I left just a nit on some methods naming.

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/CruiseControl.java
scholzj: Yeah, that makes sense. Done.

PR #10026, opened 2024-04-25
Relax the KafkaConnect admin/loggers validation

blaghed:
### Type of change

- Enhancement / new feature

### Description

This relaxes the JSON validation done when parsing the response of the `/admin/loggers` call of KafkaConnect.

### Checklist

- [ ] Write tests
- [ ] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaConnectApiImpl.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaConnectApiImpl.java
@@ -443,14 +443,9 @@ class KafkaConnectApiImpl implements KafkaConnectApi {
                                             Map<String, Map<String, String>> fetchedLoggers = mapper.readValue(buffer.getBytes(), MAP_OF_MAP_OF_STRINGS);
                                             Map<String, String> loggerMap = new HashMap<>(fetchedLoggers.size());
                                             for (var e : fetchedLoggers.entrySet()) {
-                                                if (Set.of("level", "last_modified").containsAll(e.getValue().keySet()))   {
-                                                    String level = e.getValue().get("level");
-                                                    if (level != null) {
-                                                        loggerMap.put(e.getKey(), level);
-                                                    }
-                                                } else {
-                                                    result.tryFail(new RuntimeException("Inner map has unexpected keys " + e.getValue().keySet()));
-                                                    break;
+                                                String level = e.getValue().get("level");
+                                                if (level != null) {
+                                                    loggerMap.put(e.getKey(), level);
                                                 }
                                             }
                                             result.tryComplete(loggerMap);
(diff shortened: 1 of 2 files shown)


cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaConnectApiImpl.java
```diff
@@ -443,14 +443,9 @@ public Future<Map<String, String>> listConnectLoggers(Reconciliation reconciliat
...
-                                                    if (level != null) {
-                                                        loggerMap.put(e.getKey(), level);
-                                                    }
-                                                } else {
-                                                    result.tryFail(new RuntimeException("Inner map has unexpected keys " + e.getValue().keySet()));
-                                                    break;
+                                                String level = e.getValue().get("level");
+                                                if (level != null) {
+                                                    loggerMap.put(e.getKey(), level);
                                                 }
```
blaghed, line 449: Was considering leaving the `tryFail` in case `level` doesn't exist at all, but noticed that no other operation on this class does this, so felt more natural to do it the same way.

scholzj:
@tombentley Could you please have a look if you are fine with this when you have a minute? I think you wrote the original code.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaConnectApiImpl.java
```diff
@@ -443,14 +443,9 @@ public Future<Map<String, String>> listConnectLoggers(Reconciliation reconciliat
                                             Map<String, Map<String, String>> fetchedLoggers = mapper.readValue(buffer.getBytes(), MAP_OF_MAP_OF_STRINGS);
                                             Map<String, String> loggerMap = new HashMap<>(fetchedLoggers.size());
                                             for (var e : fetchedLoggers.entrySet()) {
```
tombentley, line 445: while we're here we might as well change `e` to something a bit more meaningful, like `loggerEntry`

tombentley (commented):
Many thanks @blaghed! I left one comment but the change itself looks fine. 

There are already some tests for this class, but not for this method. It would be good to add a test.

Finally, please can you update the PR description with a little more motivation about why we're making this change (e.g. link to an issue)?

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaConnectApiImpl.java
blaghed: Sure, changed to this.

blaghed:
> There are already some tests for this class, but not for this method. It would be good to add a test.

Blindly added some, but note that this project is neither running out of the box nor does it have great code style sharing mechanisms, so adding this is a real pain and I rely on the builds here to even know if the test runs...

> Finally, please can you update the PR description with a little more motivation about why we're making this change (e.g. link to an issue)?

This PR comes from the discussion already linked (strimzi-kafka-operator#9747).

scholzj (approved):

tombentley (approved):
Thanks @blaghed.

PR #10044, opened 2024-04-30
Remove PKCS12 files from the server certificate secrets

scholzj:
### Type of change

- Refactoring

### Description

This PR removes the PKCS12 files from the internal secrets used by components of the Kafka cluster. This includes the server certificates of Kafka, Zoo or CC as well as the client certificates of KE, UO, TO, and CO. We added the KCS12 long time ago. But never really used them in most places. Removing them simplifies the code and its testing and makes the Secrets smaller.

This does nto impact the PKCS12 files int he CA secrets or in the User secrets. Those are expected to be used by users and we want to keep them.

### Checklist

- [x] Write tests

### DIFF
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/model/CertUtils.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/model/CertUtils.java
@@ -143,13 +143,8 @@ public class CertUtils {
      * may have empty key, cert or keystore and null store password.
      */
     public static CertAndKey keyStoreCertAndKey(Secret secret, String keyCertName) {
-        return new CertAndKey(
-                decodeFromSecret(secret, Ca.SecretEntry.KEY.asKey(keyCertName)),
-                decodeFromSecret(secret, Ca.SecretEntry.CRT.asKey(keyCertName)),
-                null,
-                null,
-                null
-        );
+        return new CertAndKey(decodeFromSecret(secret, Ca.SecretEntry.KEY.asKey(keyCertName)),
+                decodeFromSecret(secret, Ca.SecretEntry.CRT.asKey(keyCertName)));
     }
 
     /**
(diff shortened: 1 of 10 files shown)


cluster-operator/src/main/java/io/strimzi/operator/cluster/model/CertUtils.java
```diff
@@ -165,9 +146,9 @@ public static CertAndKey keyStoreCertAndKey(Secret secret, String keyCertName) {
         return new CertAndKey(
                 decodeFromSecret(secret, Ca.SecretEntry.KEY.asKey(keyCertName)),
                 decodeFromSecret(secret, Ca.SecretEntry.CRT.asKey(keyCertName)),
-                new byte[]{},
-                decodeFromSecret(secret, Ca.SecretEntry.P12_KEYSTORE.asKey(keyCertName)),
-                decodeStringFromSecret(secret, Ca.SecretEntry.P12_KEYSTORE_PASSWORD.asKey(keyCertName))
+                null,
```
katheris, line 149: Rather than passing null here, it looks like CertAndKey has a constructor that only takes 2 arguments.

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/ClusterCa.java
```diff
@@ -359,35 +350,31 @@ && secretEntryExists(secret, podName, SecretEntry.KEY) // The secret has the pri
...
+
+        String certData = certificateData.get(SecretEntry.CRT.asKey(podName));
+        if (certData == null) {
+            throw new RuntimeException("Certificate for node " + podName + " is missing the public key");
+        }
+
+        return new CertAndKey(
+                Util.decodeBytesFromBase64(keyData),
+                Util.decodeBytesFromBase64(certData),
+                null,
```
katheris, line 375: Similar to previous comment, can use 2 args constructor.

katheris (approved):
LGTM, just small nit around using different constructor for CertAndKey

ppatierno (approved):
LGTM.

PR #10056, opened 2024-05-03
Cruise Control mock refactoring

fvaleri:
This is part of the refactoring mentioned in the RF change proposal. Next PRs: share CC client logic, replacing Vertx with Java HTTP client.


cluster-operator/src/test/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceAssemblyOperatorTest.java
```diff
@@ -114,16 +121,26 @@ public static void beforeAll() {
...
         vertx = Vertx.vertx();
         sharedWorkerExecutor = vertx.createSharedWorkerExecutor("kubernetes-ops-pool");
+        
+        // Configure Cruise Control mock
+        serverPort = CruiseControl.REST_API_PORT;
+        tlsKeyFile = TestUtils.tempFile(".key");
+        tlsCrtFile = TestUtils.tempFile(".crt");
+
+        new MockCertManager().generateSelfSignedCert(tlsKeyFile, tlsCrtFile,
+            new Subject.Builder().withCommonName("Trusted Test CA").build(), 365);
```
scholzj, line 131: What is the performance impact on this? Working with real certs significantly slows down the tests, so you should preferebly not do it.

operator-common/src/main/java/io/strimzi/operator/common/Util.java
```diff
@@ -539,4 +546,56 @@ public static String fromAsciiBytes(byte[] bytes) {
...
+     * 
+     * @param username Username.
+     * @param password Password.
+     * 
+     * @return Header value.
+     */
+    public static String buildBasicAuthValue(String username, String password) {
+        String credentials = join(":", username, password);
+        return format("Basic %s", Util.encodeToBase64(credentials));
+    }
```
scholzj, line 560: This seems to be used only by TO. So it should probably live there.

operator-common/src/main/java/io/strimzi/operator/common/Util.java
```diff
@@ -539,4 +546,56 @@ public static String fromAsciiBytes(byte[] bytes) {
...
+
+    /**
+     * Get file content as string.
+     * 
+     * @param filePath File path.
+     * 
+     * @return Content as string.
+     */
+    public static String stringFromFile(File filePath) {
```
scholzj, line 569: Seems to be used only in TO tests. So it should live in TO tests.

operator-common/src/main/java/io/strimzi/operator/common/Util.java
```diff
@@ -539,4 +546,56 @@ public static String fromAsciiBytes(byte[] bytes) {
...
+    }
+
+    /**
+     * Get JSON content as string from resource.
+     * 
+     * @param resourcePath Resource path.
+     * 
+     * @return JSON content as string.
+     */
+    public static String jsonFromResource(String resourcePath) {
```
scholzj, line 589: Seems to be used only in tests?

operator-common/src/test/java/io/strimzi/operator/common/operator/resource/cruisecontrol/AbstractMockCruiseControl.java
scholzj, line 1: To be honest I do not understand the existence of this file. It has minimal amount of logic, is in a module not using it and it has two different implementations in different modules. I would understand if you want to have a single shared mock (in which case I think the `test` module would be the right place). But this does not make much sense to me.

scholzj (changes requested):

cluster-operator/src/test/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceAssemblyOperatorTest.java
fvaleri: This is executed one time per test class and it takes around 2 microseconds.

operator-common/src/main/java/io/strimzi/operator/common/Util.java
fvaleri: Ok, moved to TO util. Thanks.

operator-common/src/main/java/io/strimzi/operator/common/Util.java
fvaleri: Ok, moved to TO util. Thanks.

operator-common/src/main/java/io/strimzi/operator/common/Util.java
fvaleri: Ok, moved to test/TestUtils. Thanks.

operator-common/src/test/java/io/strimzi/operator/common/operator/resource/cruisecontrol/AbstractMockCruiseControl.java
fvaleri: The idea was to give a common structure to these mocks, but I agree that it doesn't make much sense given the minimal amount of logic.

cluster-operator/src/test/java/io/strimzi/operator/cluster/operator/resource/cruisecontrol/CruiseControlClientTest.java
```diff
@@ -390,4 +407,124 @@ private void ccBrokerDoesNotExist(Vertx vertx, VertxTestContext context, Abstrac
...
+
+    @Test
+    public void testMockCCServerPendingCallsOverride(Vertx vertx, VertxTestContext context) throws IOException, URISyntaxException {
+        CruiseControlApi client = cruiseControlClientProvider(vertx);
+        String userTaskID = MockCruiseControl.REBALANCE_NO_GOALS_RESPONSE_UTID;
+
+        int pendingCalls1 = 2;
+        Checkpoint firstPending = context.checkpoint(pendingCalls1);
+        int pendingCalls2 = 4;
+        Checkpoint secondPending = context.checkpoint(pendingCalls2);
```
ppatierno, line 439: AFAIU this test runs the same thing but twice ... first time with 2 pending calls and second time with 4 pending calls? Am I right? If yes, why?

fvaleri: TBH, I'm just doing refactoring here, not changing any test. This tests was simply moved from MockCruiseControlTest as it was actually testing the CC client.

ppatierno: Interesting ... Git is blaming me ... who knows why I wrote the test this way. I would leave this out for this PR, I will check the code deeply and if no reasons I will just remove one of them.

ppatierno (approved):
LGTM.

cluster-operator/src/test/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceAssemblyOperatorTest.java
```diff
@@ -114,16 +121,26 @@ public static void beforeAll() {
 
         vertx = Vertx.vertx();
         sharedWorkerExecutor = vertx.createSharedWorkerExecutor("kubernetes-ops-pool");
+        
+        // Configure Cruise Control mock
+        serverPort = CruiseControl.REST_API_PORT;
```
scholzj, line 126: As you are changing this, would it make sense to change it everywhere the same way or not at all?
* Here you create a static variable but assign it a fixed value
* In `KafkaRebalanceStateMachineTest` you keep using `CruiseControl.REST_API_PORT` directly
* In `CruiseControlClientTest` you seem to change it to using static variable and `TestUtils.getFreePort()` => this seems to me like the best variant

If you are alread changing it, it seems like it might make sense to change it the same way everywhere.

topic-operator/src/test/java/io/strimzi/operator/topic/TopicOperatorTestUtil.java
```diff
@@ -130,4 +138,17 @@ static <T> T waitUntilCondition(Resource<T> resource, Predicate<T> condition) {
     static KafkaTopic findKafkaTopicByName(List<KafkaTopic> topics, String name) {
         return topics.stream().filter(kt -> kt.getMetadata().getName().equals(name)).findFirst().orElse(null);
     }
+    
+    static String stringFromFile(File filePath) {
```
scholzj, line 142: Why not just do something like `new String(Files.readAllBytes(Paths.get(fileName)))`?

scholzj (commented):
Few more nits, but LGTM otherwise.

cluster-operator/src/test/java/io/strimzi/operator/cluster/operator/assembly/KafkaRebalanceAssemblyOperatorTest.java
fvaleri: The problem with those tests is that they don't use the CC client directly, but through `KafkaRebalanceAssemblyOperator` where CC port is hard coded to `CruiseControl.REST_API_PORT`. I didn't changed it to keep the PR focused on tests, but I can do it if you prefer.

topic-operator/src/test/java/io/strimzi/operator/topic/TopicOperatorTestUtil.java
fvaleri: `Files.lines` works well with small (this case) and large text files, so I think it's better for a general purpose util method. I just did some improvements.

scholzj (approved):

fvaleri:
@scholzj I added one more small commit to share the basic auth value method, which was in the first commit, but not complete. Sorry.

operator-common/src/main/java/io/strimzi/operator/common/Util.java
```diff
@@ -539,4 +542,16 @@ public static String fromAsciiBytes(byte[] bytes) {
...
 
+    /**
+     * Build basic HTTP authentication header value.
+     *
+     * @param username Username.
+     * @param password Password.
+     *
+     * @return Header value.
+     */
+    public static String buildBasicAuthValue(String username, String password) {
```
scholzj, line 553: I think you should create some `CruiseControlUtils` to keep some order in what is used where.

scholzj (changes requested):

operator-common/src/main/java/io/strimzi/operator/common/Util.java
fvaleri: ok, will come handy in the future

scholzj (approved):

PR #10057, opened 2024-05-03
[ST] MigrationST - add checks for JBOD storage, change the main tests to be parametrized

### LINKED ISSUE #10000: [ST] Add migration system tests with JBOD support in KRaft mode
With the support for JBOD disks in KRaft mode added by https://github.com/strimzi/strimzi-kafka-operator/pull/9936 and it's usage during migration as well (together with a fix by https://github.com/strimzi/strimzi-kafka-operator/pull/9993) we should add additional system tests for migration about migrate and rollback an Apache Kafka cluster using multiple JBOD disks.

im-konge:
### Type of change

- Enhancement

### Description

This PR adds test checks (and tests) for migration/rollback with Kafka containing JBOD storage (in both broker and controller).
Together with this I did a small refactor of the ST class and changed the two main tests (for migration and rollback) to be parametrized tests. That's because there are more parameters + it will be clear which test cases we cover (because we had there one redundant test, so this should make it more clear).

Finally, this PR changes the test documentation to reflect the changes.

Closes #10000 

### Checklist

- [x] Write tests

### DIFF
--- a/systemtest/src/test/java/io/strimzi/systemtest/migration/MigrationST.java
+++ b/systemtest/src/test/java/io/strimzi/systemtest/migration/MigrationST.java
@@ -91,10 +91,11 @@ public class MigrationST extends AbstractST {
      * disrupted.
      *
      * This is a parametrized test with following arguments:
-     *      - ZK delete claim - when set to true, the PVs of ZK should be deleted at the end of the test by the operator; if set to false, the PVs should be kept
-     *      - delete CO during process - inside the test-case there are situations (typically when some rolling update starts) when the CO will be deleted
-     *              -> this will test that the migration will be finished even if the CO will be deleted/rolled
-     *      - use JBOD storage - last argument determines if the Kafka brokers & controllers should use single PV or JBOD storage (with 2 PVs)
+     * <ul>
+     *     <li>$1 - Delete Zookeeper claim: true, false</li>
+     *     <li>$2 - Delete Cluster Operator during the process: false, true</li>
+     *     <li>$3 - Use JBOD storage configuration: true, false</li>
+     * </ul>
      *
      * Based on the checks, the steps can be a bit different.
      *
@@ -156,9 +157,10 @@ public class MigrationST extends AbstractST {
      * disrupted.
      *
      * This is a parametrized test with following arguments:
-     *      - delete CO during process - inside the test-case there are situations (typically when some rolling update starts) when the CO will be deleted
-     *              -> this will test that the migration/rollback will be finished even if the CO will be deleted/rolled
-     *      - use JBOD storage - last argument determines if the Kafka brokers & controllers should use single PV or JBOD storage (with 2 PVs)
+     * <ul>
+     *     <li>$1 - Delete Cluster Operator during the process: false, true</li>
+     *     <li>$2 - Use JBOD storage configuration: true, false</li>
+     * </ul>
      *
      * Based on the checks, the steps can be a bit different.
      * @steps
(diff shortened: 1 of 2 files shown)


systemtest/src/test/java/io/strimzi/systemtest/migration/MigrationST.java
```diff
@@ -654,6 +673,71 @@ private void deleteClusterOperator() {
...
+                .build();
+        }
+
+        return volume;
+    }
+
+    /**
+     * Method returning simple stream of arguments for migration parametrized tests.
+     * First argument determines if the ZK delete claim should be set to true/false, second if
+     * the CO should be deleted during the migration/rollback process, and third if the JBOD storage should be used.
```
see-quick, line 712: Could we write this as 
```
 * <ul>
 *     <li>$1 - Deletes Zookeeper claim: true, false</li>
 *     <li>$2 - Deletes Cluster Operator during the process: false, true</li>
 *     <li>$3 - Uses JBOD storage configuration: true, false</li>
 * </ul>
```
or something similar? :)

see-quick (approved):
LGTM, just a one nit to consider ... Thanks 👍

systemtest/src/test/java/io/strimzi/systemtest/migration/MigrationST.java
im-konge: sure, do you want from me to change it everywhere in the MigrationST to ordered lists in HTML tag? I mean the steps etc.

systemtest/src/test/java/io/strimzi/systemtest/migration/MigrationST.java
```diff
@@ -654,6 +673,71 @@ private void deleteClusterOperator() {
...
+     *
+     * @return  {@link Storage} for the particular testcase (single PV/JBOD)
+     */
+    private Storage getStorageBasedOnTestCase(boolean withJbodStorage) {
+        String pvSize = "1Gi";
+
+        PersistentClaimStorage volume = new PersistentClaimStorageBuilder()
+            .withDeleteClaim(true)
+            .withId(METADATA_VOLUME_ID)
+            .withKraftMetadata(KRaftMetadataStorage.SHARED)
```
ppatierno, line 690: This flag doesn't make sense on brokers when we start from a ZK cluster but luckily it doesn't hurt. I have spoken with Jakub, he will add a warning when it happens but it should have no impact on the STs. Just an heads up for you.

ppatierno (approved):
LGTM. Left just a comment.

systemtest/src/test/java/io/strimzi/systemtest/migration/MigrationST.java
im-konge: I have it there for the Broker pods to have the exact disk where the cluster metadata topic should be present (I thought it will that effect as for the controller Pods).
If it's the case, I don't want to change it during the test, so I just placed it there from the beginning, to make the checks easier.

ppatierno: so yes in general, what's going to happen for a user would be ...

1. start from a ZK based cluster where broker node pool doesn't have kraftMedatata flag on JBOD disks
2. migrate to KRaft
3. on brokers the cluster metadata topic will go to the JBOD disk with ID = 0 (or the minimum one)

In your case you are marking the specific disk to be used after the migration for cluster topic on brokers. It's fine anyway.

im-konge: Okay cool, I just wanted to be sure. I guess I can keep it as it is, but thanks for the heads up! :)

systemtest/src/test/java/io/strimzi/systemtest/migration/MigrationST.java
see-quick: No, I think just part:
```
 * First argument determines if the ZK delete claim should be set to true/false
 ```
 to 
 ```
 $1 - Deletes Zookeeper claim: true, false
 ...
 etc.
 ```
 Would be sufficient enough.

im-konge: ok, will do

PR #10058, opened 2024-05-03
Fix `/controller` znode failure handling during KRaft migration rollback

ppatierno:
### Type of change

- Bugfix

### Description

This PR adds the handling of failures, if any, when the operator deletes the `/controller` znode during a KRaft migration rollback. The current code, just log a warning and continue which is not right. This PR adds correct handling with failed Future to fail the reconciliation.
There are a lot of file changes on several tests classes, because I had to add a way to mock the ZooKeeper admin provider through the resource supplier.

### Checklist

### DIFF
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KRaftMigrationUtils.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KRaftMigrationUtils.java
@@ -5,6 +5,7 @@
 
 package io.strimzi.operator.cluster.operator.assembly;
 
+import io.strimzi.operator.cluster.operator.VertxUtil;
 import io.strimzi.operator.cluster.operator.resource.DefaultZooKeeperAdminProvider;
 import io.strimzi.operator.cluster.operator.resource.KRaftMigrationState;
 import io.strimzi.operator.cluster.operator.resource.KafkaAgentClient;
@@ -14,6 +15,9 @@ import io.strimzi.operator.common.Util;
 import io.strimzi.operator.common.auth.PemAuthIdentity;
 import io.strimzi.operator.common.auth.PemTrustSet;
 import io.strimzi.operator.common.auth.TlsPemIdentity;
+import io.vertx.core.Future;
+import io.vertx.core.Promise;
+import io.vertx.core.Vertx;
 import org.apache.zookeeper.KeeperException;
 import org.apache.zookeeper.admin.ZooKeeperAdmin;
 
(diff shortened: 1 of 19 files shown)


cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KRaftMigrationUtils.java
```diff
@@ -31,41 +35,97 @@ public class KRaftMigrationUtils {
...
-        try (ZooKeeperAdmin admin = new DefaultZooKeeperAdminProvider().createZookeeperAdmin(
+
+        return connectToZooKeeper(reconciliation, vertx, zooKeeperAdminProvider, trustStoreFile, keyStoreFile, operationTimeoutMs, zkConnectionString)
+                .compose(zkAdmin -> {
+                    Promise<Void> znodeDeleted = Promise.promise();
+                    try {
+                        zkAdmin.delete("/controller", -1);
+                        LOGGER.infoCr(reconciliation, "Deleted the '/controller' znode as part of the KRaft migration rollback");
+                        znodeDeleted.complete();
+                    } catch (KeeperException | InterruptedException e)    {
```
scholzj, line 61: What if IllegalArgumentException is thrown?

ppatierno: Which part can raise such exception here? I am catching what the `zkAdmin.delete` could throw.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KRaftMigrationUtils.java
```diff
@@ -31,41 +35,97 @@ public class KRaftMigrationUtils {
...
+
+    private static Future<Void> closeZooKeeperConnection(Reconciliation reconciliation, Vertx vertx, ZooKeeperAdmin zkAdmin, File trustStoreFile, File keyStoreFile, long operationTimeoutMs) {
+        if (zkAdmin != null) {
+            return vertx.executeBlocking(() -> {
+                try {
+                    zkAdmin.close((int) operationTimeoutMs);
+                    return null;
+                } catch (Exception e) {
+                    LOGGER.warnCr(reconciliation, "Failed to close the ZooKeeperAdmin", e);
+                    throw e;
```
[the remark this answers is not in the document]
ppatierno: I would agree. If we cannot close Zoo connection there isn't much we can do. Warning would be just enough.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KRaftMigrationUtils.java
scholzj: `delete(...)` has it in its Javadoc -> when the path does not exist. But I would maybe correct my self -> you should catch anything in here. Because if you don't, you will never complete the future and the reconciliation will likely get _stuck_, or?

ppatierno: Didn't notice it on the Javadoc. But yeah you are right, it's better catching any exception.

PR #10060, opened 2024-05-03
[ST] Fix RackAwarenessST, JmxST, and file-sink image build

im-konge:
### Type of change

- ST fixes

### Description

This PR fixes following issues:

- `RackAwarenessST` - in assertions where we are checking that the hostname is in the particular line of text, we are assuming that there will be just the hostname we are specifying, however for example on AWS, there can be also additional suffix (like region)
- `JmxST` - when running with NP default to deny, we are not able to connect to the API and get the JMX metrics, this PR adds creation of NP for Connect, so we are able to successfully collect the particular beans
- Connect image with file-sink build - when the connect plugin has additional suffix (and not just the Kafka version with `.jar`), the build simply fails, as the jar can be named differently (with different suffix)

### Checklist


Frawless (approved):

see-quick (approved):
LGTM, Thanks 👍

PR #10062, opened 2024-05-04
Move dependency analysis from the root pom.xml to the modules

### LINKED ISSUE #8572: [Enhancement]: Ignored unused declared (etc) applied too broadly
### Related problem

We use `maven-dependency-plugin`'s `analyze-only` goal to analyze poms checking that we're not declaring dependencies which we don't need (or are declaring dependencies which we do need, typically where we have a direct dependency on an artifact but it's pulled in as a transitive dependency of a declared depdenency).

Sometimes this analysis is wrong. For example if there are two direct dependencies (A and B) which depend (possibly transitively) on different versions of a common 3rd dependency (C). Maven will resolve to the version on the shortest path, but that's not guaranteed to work in practice. We fix this by declaring a depdency on C in the pom and selecting the version which we want to use. `analyze-only` sees that as an unused declared, so we have to explicitly ignore it: 

```xml
            <plugin>
                <groupId>org.apache.maven.plugins</groupId>
                <artifactId>maven-dependency-plugin</artifactId>
                <version>${maven.dependency.version}</version>
                <executions>
                    <execution>
                    <id>analyze</id>
                    <goals>
                        <goal>analyze-only</goal>
                    </goals>
                    <configuration>
                        <ignoredUnusedDeclaredDependencies>
                            <ignoredUnusedDeclaredDependency>INSERT COORDINATES FOR C HERE</ignoredUnusedDeclaredDependency>
```

The problem is we've been configuring these overrides in the root pom, so they apply to all submodules by inheritance. That means some submodules may be incorrectly ignoring some warnings which they ought not to.

Furthermore, we would ideally track the reason for ignoring the warnings for each dependency, so we can remove the ignore config is the underlying cause gets fixed in the future. Managing these ignore configs in the root pom makes that really difficult, because each ignored warning might only apply to a single submodule.

### Suggested solution

1. Declare the plugin and its configuration in each submodule. 
2. Use comments to track the need for ignoring these warnings and a related comment on the `<dependency>` to improve the chances that we managed dependencies properly.

### Alternatives

* either: Stick with the status quo because it's easier.
* or: stop using the `maven-dependency-plugin` and let chaos reign.

### Additional context

_No response_

scholzj:
### Type of change

- Refactoring

### Description

This PR moves the Mavane dependency analysis from the root `pom.xml` file to the modules. This allows us to have more granular controller over the dependencies that are ignored unused etc. as we now configure them on the module level and not once for the whole project.

As part of this PR, I also fixed some smaller test issues I discovered thanks to this:
* The `CronTest` was for some reason testing the Log4j cron expression instead of the Quartz one
* The `kafka-agent` tests were not exected at all because they were written for JUnit 4 and not JUnit 5

This PR should resolve #8572 

### Checklist


ppatierno (commented):
LGTM

ppatierno (approved):

PR #10067, opened 2024-05-06
Raise warning when `kraftMetadata` is used in storage configuration for ZooKeeper-based clusters

scholzj:
### Type of change

- Enhancement / new feature

### Description

This PR adds a warning when the `kraftMetadata` option in the storage configuration is used in ZooKeeper-based cluster.

To simplify the code, the warning message is generic without mentioning the exact resource(s) (Kafka or KafkaNodePool) where the issue is. Creating the warning including such a detailed message would significantly increase the complexity as it would require several things:
* Information about the KRaft state (enabled, disabled, etc.)
* Detailed information about the original resources configuration
* The warnings to be passed to the custom resource status

These three are not available in one place and the effort to bring them together does not seem to be worth the effort.

### Checklist

- [x] Write tests


ppatierno (approved):
LGTM! Thanks and sorry for bothering you with this request ;-)

PR #10083, opened 2024-05-08
docs(jbod): kraft: expands description for managing metadata storage

PaulRMellor:
**Documentation**

Adds information on managing the storage volume used for the KRaft-based cluster metadata.
Storage requirements for node types and what happens when the storage volume used for storing metadata changes.  

### Checklist

- [ ] Write tests
- [x] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/documentation/modules/configuring/ref-storage-jbod.adoc
+++ b/documentation/modules/configuring/ref-storage-jbod.adoc
@@ -7,7 +7,7 @@
 
 [role="_abstract"]
 JBOD storage allows you to configure your Kafka cluster to use multiple disks or volumes. 
-This approach provides increased data storage capacity for Kafka brokers, and can lead to performance improvements.
+This approach provides increased data storage capacity for Kafka nodes, and can lead to performance improvements.
 A JBOD configuration is defined by one or more volumes, each of which can be either xref:ref-ephemeral-storage-{context}[ephemeral] or xref:ref-persistent-storage-{context}[persistent]. 
 The rules and constraints for JBOD volume declarations are the same as those for ephemeral and persistent storage. 
 For example, you cannot decrease the size of a persistent storage volume after it has been provisioned, nor can you change the value of `sizeLimit` when the type is `ephemeral`.
@@ -15,7 +15,7 @@ For example, you cannot decrease the size of a persistent storage volume after i
 NOTE: JBOD storage is supported for *Kafka only*, not for ZooKeeper.
 
 To use JBOD storage, you set the storage type configuration in the `Kafka` resource to `jbod`.
-If you are using node pools, you can also specify `jbod` in the storage configuration of individual node pools.
+If you are using node pools, you can also specify `jbod` in the storage configuration for nodes belonging to a specific node pool.
 
 The `volumes` property allows you to describe the disks that make up your JBOD storage array or configuration. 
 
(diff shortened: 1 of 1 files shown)


documentation/modules/configuring/ref-storage-jbod.adoc
```diff
@@ -68,10 +68,19 @@ Where `_id_` is the ID of the volume used for storing data for Kafka broker pod
 
 == Configuring the storage volume used to store the KRaft metadata log
 
-When running a Kafka cluster in a KRaft mode, each node must store the KRaft metadata log on one of its data volumes.
-By default, the KRaft metadata log is stored on the volume with the lowest ID.
+In KRaft mode, a Kafka cluster stores its metadata in a log present on all nodes, including brokers and controllers.
```
scholzj, line 71: I wonder if this would be a bit better wording as not every metadata log is equal and on the brokers it is AFAIU only a local copy and not the main metadata log as with the controllers.

```suggestion
In KRaft mode, a copy of the metadata log is stored on every node, including brokers and controllers.
```

documentation/modules/configuring/ref-storage-jbod.adoc
```diff
@@ -68,10 +68,19 @@ Where `_id_` is the ID of the volume used for storing data for Kafka broker pod
...
+Each node reserves one of its data volumes for the KRaft metadata log.
```
scholzj, line 72: `reserves` suggests to me that it is used only for the metadata which is wrong. Maybe this would sound better?

```suggestion
Each node uses one of its data volumes for the KRaft metadata log. 
```

documentation/modules/configuring/ref-storage-jbod.adoc
```diff
@@ -68,10 +68,19 @@ Where `_id_` is the ID of the volume used for storing data for Kafka broker pod
...
 == Configuring the storage volume used to store the KRaft metadata log
 
-When running a Kafka cluster in a KRaft mode, each node must store the KRaft metadata log on one of its data volumes.
-By default, the KRaft metadata log is stored on the volume with the lowest ID.
+In KRaft mode, a Kafka cluster stores its metadata in a log present on all nodes, including brokers and controllers.
+Each node reserves one of its data volumes for the KRaft metadata log. 
+By default, the log is stored on the volume with the lowest ID. 
 However, you can specify another volume using the `kraftMetadata` property.
 
+For controller-only nodes, which don't handle data, dedicating a single volume to the metadata log is generally sufficient for storage.
```
scholzj, line 76: I would make it more clear that this means JBOD makes little sense here.

```suggestion
For controller-only nodes, which don't handle data, the storage will be used only for the metadata log.
The metadata log will be always stored only on one volume, so using JBOD storage with multiple volumes does not improve the performance or increase the effectively available disk space. 
```

documentation/modules/configuring/ref-storage-jbod.adoc
```diff
@@ -68,10 +68,19 @@ Where `_id_` is the ID of the volume used for storing data for Kafka broker pod
...
 
-When running a Kafka cluster in a KRaft mode, each node must store the KRaft metadata log on one of its data volumes.
-By default, the KRaft metadata log is stored on the volume with the lowest ID.
+In KRaft mode, a Kafka cluster stores its metadata in a log present on all nodes, including brokers and controllers.
+Each node reserves one of its data volumes for the KRaft metadata log. 
+By default, the log is stored on the volume with the lowest ID. 
 However, you can specify another volume using the `kraftMetadata` property.
 
+For controller-only nodes, which don't handle data, dedicating a single volume to the metadata log is generally sufficient for storage. 
+Meanwhile, for broker nodes or nodes combining broker and controller roles, although they may require multiple volumes, they still share the same volume for storing both the metadata log and partition replica data.
```
scholzj, line 77:
```suggestion
Meanwhile, broker nodes or nodes combining broker and controller roles, will share the same volume for storing both the metadata log and partition replica data.
They can also utilize JBOD storage with multiple volumes as one of the volumes will be shared by the metadata log and partition replica data and any additional volumes will be used for partition replica data only.
```

ppatierno (approved):
LGTM (other than Jakub's comments)

scholzj (approved):

PR #10092, opened 2024-05-09
Extract Topic Operator's Cruise Control client

fvaleri:
This is the first step in the direction of having a shared Cruise Control client based on Java HTTP client. For the moment changes are limited to the Topic Operator.

If we agree on the strategy, I can then move the Cruise Control client to operator-common and also use it in CruiseControlApi (different PR).

---

In my view, the shared Cruise Control client shouldn't be concerned with response handling. This client should simply send the request and return a response POJO. Much like Kafka admin and Kube clients do.

The response handling is specific to the functionality we are implementing (the same response could be processed in different ways), so it should be the responsibility of the caller (CruiseControlApi in CO, ReplicasChangeHandler in TO).

That way, each object in the call chain triggered by the controller is focused on doing one thing, and the code is easier to read and maintain.

This would leave a Vertx dependency in CruiseControlApi required by KafkaRebalanceAssemblyOperator. Of course we can also get rid of that, but I would leave it for a dedicated PR.

### DIFF
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/resource/cruisecontrol/CruiseControlApi.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/resource/cruisecontrol/CruiseControlApi.java
@@ -21,11 +21,6 @@ public interface CruiseControlApi {
     String CC_REST_API_PROGRESS_KEY = "progress";
 
     /**
-     * User ID header key
-     */
-    String CC_REST_API_USER_ID_HEADER = "User-Task-ID";
-
-    /**
      *  Gets the state of the Cruise Control server.
      *
      * @param host The address of the Cruise Control server.
--- a/operator-common/src/main/java/io/strimzi/operator/common/model/cruisecontrol/CruiseControlHeaders.java
+++ b/operator-common/src/main/java/io/strimzi/operator/common/model/cruisecontrol/CruiseControlHeaders.java
@@ -5,27 +5,11 @@
 package io.strimzi.operator.common.model.cruisecontrol;
 
 /**
- * Enum with Cruise Control headers
+ * Cruise Control headers.
  */
-public enum CruiseControlHeaders {
+public class CruiseControlHeaders {
     /**
-     * User task id
+     * User task id.
      */
-    USER_TASK_ID_HEADER("User-Task-ID");
-
-    private final String name;
-
-    /**
-     * Creates the Enum from String
-     *
-     * @param name String with the path
-     */
-    CruiseControlHeaders(String name) {
-        this.name = name;
-    }
-
-    @Override
-    public String toString() {
-        return name;
-    }
+    public static final String USER_TASK_ID_HEADER = "User-Task-ID";
 }
--- a/operator-common/src/main/java/io/strimzi/operator/common/model/cruisecontrol/CruiseControlHeaders.java
+++ b/operator-common/src/main/java/io/strimzi/operator/common/model/cruisecontrol/CruiseControlHeaders.java
@@ -1,0 +1,31 @@
+/*
+ * Copyright Strimzi authors.
+ * License: Apache License 2.0 (see the file LICENSE or http://apache.org/licenses/LICENSE-2.0.html).
+ */
+package io.strimzi.operator.common.model.cruisecontrol;
+
+/**
+ * Enum with Cruise Control headers
+ */
+public enum CruiseControlHeaders {
+    /**
+     * User task id
+     */
+    USER_TASK_ID_HEADER("User-Task-ID");
+
+    private final String name;
+
+    /**
+     * Creates the Enum from String
+     *
+     * @param name String with the path
+     */
+    CruiseControlHeaders(String name) {
+        this.name = name;
+    }
+
+    @Override
+    public String toString() {
+        return name;
+    }
+}
--- a/topic-operator/src/main/java/io/strimzi/operator/topic/ReplicasChangeHandler.java
+++ b/topic-operator/src/main/java/io/strimzi/operator/topic/ReplicasChangeHandler.java
@@ -12,7 +12,6 @@ import io.strimzi.operator.common.ReconciliationLogger;
 import io.strimzi.operator.topic.cruisecontrol.CruiseControlClient;
 import io.strimzi.operator.topic.cruisecontrol.CruiseControlClient.TaskState;
 import io.strimzi.operator.topic.cruisecontrol.CruiseControlClient.UserTasksResponse;
-import io.strimzi.operator.topic.cruisecontrol.CruiseControlClientImpl;
 
 import java.io.IOException;
 import java.nio.file.Files;
@@ -42,15 +41,15 @@ import static org.apache.logging.log4j.core.util.Throwables.getRootCause;
 public class ReplicasChangeHandler {
     private static final ReconciliationLogger LOGGER = ReconciliationLogger.create(ReplicasChangeHandler.class);
     
-    private final CruiseControlClient apiClient;
+    private final CruiseControlClient cruiseControlClient;
     
     /**
      * Create a new replicas change handler instance.
      * 
      * @param config Topic Operator configuration.
      */
-    ReplicasChangeHandler(TopicOperatorConfig config) {
-        this.apiClient = new CruiseControlClientImpl(
+    public ReplicasChangeHandler(TopicOperatorConfig config) {
+        this.cruiseControlClient = CruiseControlClient.create(
             config.cruiseControlHostname(),
             config.cruiseControlPort(),
             config.cruiseControlRackEnabled(),
@@ -67,7 +66,7 @@ public class ReplicasChangeHandler {
      */
     public void stop() {
         try {
-            apiClient.close();
+            cruiseControlClient.close();
         } catch (Exception e) {
             throw new RuntimeException(e);
         }
@@ -92,7 +91,7 @@ public class ReplicasChangeHandler {
         try {
             LOGGER.debugOp("Sending topic configuration request, topics {}", topicNames(reconcilableTopics));
             List<KafkaTopic> kafkaTopics = reconcilableTopics.stream().map(rt -> rt.kt()).collect(Collectors.toList());
-            String userTaskId = apiClient.topicConfiguration(kafkaTopics);
+            String userTaskId = cruiseControlClient.topicConfiguration(kafkaTopics);
             updateToOngoing(result, "Replicas change ongoing", userTaskId);
         } catch (Throwable t) {
             updateToFailed(result, format("Replicas change failed, %s", getRootCause(t).getMessage()));
@@ -122,7 +121,7 @@ public class ReplicasChangeHandler {
 
         try {
             LOGGER.debugOp("Sending user tasks request, Tasks {}", groupByUserTaskId.keySet());
-            UserTasksResponse utr = apiClient.userTasks(new ArrayList<>(groupByUserTaskId.keySet()));
+            UserTasksResponse utr = cruiseControlClient.userTasks(groupByUserTaskId.keySet());
 
             if (utr.userTasks().isEmpty()) {
                 // Cruise Control restarted: reset the state because the tasks queue is not persisted
--- a/topic-operator/src/main/java/io/strimzi/operator/topic/ReplicasChangeHandler.java
+++ b/topic-operator/src/main/java/io/strimzi/operator/topic/ReplicasChangeHandler.java
@@ -4,79 +4,73 @@
  */
 package io.strimzi.operator.topic;
 
-import com.fasterxml.jackson.annotation.JsonProperty;
-import com.fasterxml.jackson.core.JsonProcessingException;
-import com.fasterxml.jackson.databind.ObjectMapper;
-import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
 import io.strimzi.api.kafka.model.topic.KafkaTopic;
 import io.strimzi.api.kafka.model.topic.KafkaTopicStatusBuilder;
 import io.strimzi.api.kafka.model.topic.ReplicasChangeStatusBuilder;
 import io.strimzi.operator.common.Reconciliation;
 import io.strimzi.operator.common.ReconciliationLogger;
-import io.strimzi.operator.common.model.cruisecontrol.CruiseControlEndpoints;
-import io.strimzi.operator.common.model.cruisecontrol.CruiseControlParameters;
+import io.strimzi.operator.topic.cruisecontrol.CruiseControlClient;
+import io.strimzi.operator.topic.cruisecontrol.CruiseControlClient.TaskState;
+import io.strimzi.operator.topic.cruisecontrol.CruiseControlClient.UserTasksResponse;
+import io.strimzi.operator.topic.cruisecontrol.CruiseControlClientImpl;
 
-import javax.net.ssl.SSLContext;
-import javax.net.ssl.TrustManagerFactory;
-
-import java.io.ByteArrayInputStream;
 import java.io.IOException;
-import java.net.URI;
-import java.net.URLEncoder;
-import java.net.http.HttpClient;
-import java.net.http.HttpRequest;
-import java.net.http.HttpResponse;
-import java.security.KeyStore;
-import java.security.cert.Certificate;
-import java.security.cert.CertificateFactory;
-import java.time.Duration;
-import java.time.temporal.ChronoUnit;
+import java.nio.file.Files;
+import java.nio.file.Path;
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
-import java.util.Optional;
-import java.util.Set;
-import java.util.concurrent.ExecutorService;
-import java.util.concurrent.Executors;
 import java.util.stream.Collectors;
 
 import static io.strimzi.api.kafka.model.topic.ReplicasChangeState.ONGOING;
 import static io.strimzi.api.kafka.model.topic.ReplicasChangeState.PENDING;
-import static io.strimzi.operator.common.CruiseControlUtil.buildBasicAuthValue;
-import static io.strimzi.operator.topic.TopicOperatorUtil.getFileContent;
 import static io.strimzi.operator.topic.TopicOperatorUtil.hasReplicasChange;
 import static io.strimzi.operator.topic.TopicOperatorUtil.topicNames;
 import static java.lang.String.format;
 import static java.nio.charset.StandardCharsets.UTF_8;
-import static java.util.stream.Collectors.groupingBy;
+import static org.apache.logging.log4j.core.util.Throwables.getRootCause;
 
 /**
  * Replicas change handler that interacts with Cruise Control REST API.
  * <br/><br/>
- * The REST endpoints are {@code topic_configuration} to request replication factor changes and {@code user_tasks} 
- * to check the asynchronous execution result. Cruise Control runs one task execution at a time, additional requests 
- * are queued up to {@code max.active.user.tasks}.
- * <br/><br/>
  * At any given time, a KafkaTopic.spec.replicas change can be in one of the following states:
- * <br/>
  * <ul><li>Pending: Not in Cruise Control's task queue (not yet sent or request error).</li>
  * <li>Ongoing: In Cruise Control's task queue, but execution not started, or not completed.</li>
  * <li>Completed: Cruise Control's task execution completed (target replication factor reconciled).</li></ul>
  */
 public class ReplicasChangeHandler {
     private static final ReconciliationLogger LOGGER = ReconciliationLogger.create(ReplicasChangeHandler.class);
-    private static final String USER_TASK_ID_HEADER = "User-Task-ID";
-    private static final long REQUEST_TIMEOUT_SEC = 60;
     
-    private final TopicOperatorConfig config;
-    private final ExecutorService httpClientExecutor;
-    private final ObjectMapper mapper;
+    private final CruiseControlClient apiClient;
     
+    /**
+     * Create a new replicas change handler instance.
+     * 
+     * @param config Topic Operator configuration.
+     */
     ReplicasChangeHandler(TopicOperatorConfig config) {
-        this.config = config;
-        this.httpClientExecutor = Executors.newCachedThreadPool();
-        this.mapper = new ObjectMapper();
+        this.apiClient = new CruiseControlClientImpl(
+            config.cruiseControlHostname(),
+            config.cruiseControlPort(),
+            config.cruiseControlRackEnabled(),
+            config.cruiseControlSslEnabled(),
+            config.cruiseControlSslEnabled() ? getFileContent(config.cruiseControlCrtFilePath()) : null,
+            config.cruiseControlAuthEnabled(),
+            config.cruiseControlAuthEnabled() ? new String(getFileContent(config.cruiseControlApiUserPath()), UTF_8) : null,
+            config.cruiseControlAuthEnabled() ? new String(getFileContent(config.cruiseControlApiPassPath()), UTF_8) : null
+        );
+    }
+
+    /**
+     * Stop the replicas change handler.
+     */
+    public void stop() {
+        try {
+            apiClient.close();
+        } catch (Exception e) {
+            throw new RuntimeException(e);
+        }
     }
 
     /**
@@ -97,22 +91,11 @@ public class ReplicasChangeHandler {
         
         try {
             LOGGER.debugOp("Sending topic configuration request, topics {}", topicNames(reconcilableTopics));
-            HttpClient client = buildHttpClient();
-            client.sendAsync(buildPostTopicConfigRequest(reconcilableTopics), HttpResponse.BodyHandlers.ofString())
-                .thenAccept(response -> {
-                    if (response.statusCode() != 200) {
-                        updateToFailed(result, format("Replicas change failed (%s)", response.statusCode()), response);
-                        return;
-                    }
-                    try {
-                        String userTaskId = response.headers().firstValue(USER_TASK_ID_HEADER).get();
-                        updateToOngoing(result, "Replicas change ongoing", userTaskId);
-                    } catch (Throwable e) {
-                        updateToFailed(result, format("Failed to get %s header", USER_TASK_ID_HEADER), response);
-                    }
-                }).join();
+            List<KafkaTopic> kafkaTopics = reconcilableTopics.stream().map(rt -> rt.kt()).collect(Collectors.toList());
+            String userTaskId = apiClient.topicConfiguration(kafkaTopics);
+            updateToOngoing(result, "Replicas change ongoing", userTaskId);
         } catch (Throwable t) {
-            updateToFailed(result, format("Replicas change failed, %s", t.getMessage()), null);
+            updateToFailed(result, format("Replicas change failed, %s", getRootCause(t).getMessage()));
         }
         return result;
     }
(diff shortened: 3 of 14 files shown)


operator-common/src/main/java/io/strimzi/operator/common/model/cruisecontrol/CruiseControlHeaders.java
```diff
@@ -0,0 +1,31 @@
+/*
+ * Copyright Strimzi authors.
+ * License: Apache License 2.0 (see the file LICENSE or http://apache.org/licenses/LICENSE-2.0.html).
+ */
+package io.strimzi.operator.common.model.cruisecontrol;
+
+/**
+ * Enum with Cruise Control headers
+ */
+public enum CruiseControlHeaders {
```
ppatierno, line 10: Genuine question, what's the need of a specific class for the headers? Why isn't it ok having them in the `CruiseControlApi` interface?

ppatierno (commented):
I left one comment but not sure the impact it has on the current CC client. I mean, I see just moving a header to a dedicated class and nothing more in the direction of "extract a cruise control client" as the PR name suggests. Maybe it's just a not that great title and PR description which makes me hard the purpose?

operator-common/src/main/java/io/strimzi/operator/common/model/cruisecontrol/CruiseControlHeaders.java
fvaleri: This will be shared with the topic operator code, and it's part of the CC public API.

ppatierno: So it sounds to me you are creating this enum just for having a central/common place where putting the first user task header because you can't/don't want to move the `CruiseControlApi` interface in the operator common now? This is why I am reasoning if a dedicated headers class is really needed.

fvaleri: My plan is to not move the CruiseControlApi, but instead create a new thin CC client based on Java HTTP client, and use that inside CruiseControlApi. This is because CruiseControlApi contains a good amount of the rebalance logic which should not be part of the shared CC client. When all client endpoints will be migrated, we can also think about renaming CruiseControlApi to CruiseControlRebalanceHandler, but that's optional.

ppatierno: What do you mean by thin? The current logic in the CruiseControlApi is for handling the response from the CC REST API and translating them for the upper level (see rebalance operator). Is your client just about sending and receiving HTTP requests/responses? Hasn't the topic operator CC interaction the need for handling them?
I am just wondering if we really need two layers.

fvaleri: By thin I mean that it will only handle request/response and mapping to POJOs (1:1 mapping with the JSON response), nothing more. This is similar to what Kube client or Kafka admin client do. Instead, CruiseControlApi has some extra logic to map between CC response and what rebalance expects, which makes it hard to maintain and share.

fvaleri: > Hasn't the topic operator CC interaction the need for handling them?

Yes, that's why I created a ReplicasChangeHandler that decouples the controller from the CC client. CruiseControlApi will be a similar handler. I think this is a better design overall.

fvaleri: I already tried to migrate user_tasks (but I need this PR to move on), and I see that this approach works quite well. As a result, the CruiseControlApi logic is much easier to read.

fvaleri: Yeah, mapping stuff, I mean that. 

I think it's better to keep that logic isolated in a specific handler (CruiseControlApi in this case). Every component that requires CC integration also needs some kind of custom mapping logic. For example, the topic operator doesn't need the the full response when calling user_tasks, which makes for a faster call.

Additionally CruiseControlResponse depends on Vertx, which would also require RebalanceAssemblyOperator refactoring. This can also be done to get rid of Vertx, but I think should be a follow up PR to not over complicate this task.

ppatierno: I am not totally sure I agree with that. You would like to have one HTTP client which sends requests and get responses from Cruise Control API but then passing to an upper layer the responsability to decode and building a POJO for the response. This layer could be different between cluster operator (because of rebalance) and topic operator (because of replication factor change). You could easily end up with duplicated code because both could need most of the same fields from the received JSON. Why not having one and making a POJO which could be a common denominator for the needs, not adding ALL fields from JSON response but the ones who are useful for our needs on both cluster and topic operator. In the end across the two it's the the user task request to be in common, or? Not sure what the other @strimzi/maintainers think about this.

fvaleri: > Why not having one and making a POJO which could be a common denominator for the needs, not adding ALL fields from JSON response but the ones who are useful for our needs on both cluster and topic operator.

In my view, the shared Cruise Control client shouldn't be concerned with response handling. This client should simply send the request and return a response POJO. Much like Kafka admin and Kube clients do. 

The response handling is specific to the functionality we are implementing (the same response could be processed in different ways), so it should be the responsibility of the caller (CruiseControlApi in CO, ReplicasChangeHandler in TO).

That way, each object involved in the call chain trigger by the controller is focused on doing one thing and the code is easier to read and maintain.

This would leave a Vertx dependency in CruiseControlApi required by KafkaRebalanceAssemblyOperator. Of course we can also get rid of that, but I would leave it for a dedicated PR.

ppatierno: > In my view, the shared Cruise Control client shouldn't be concerned with response handling. This client should simply send the request and return a response POJO. Much like Kafka admin and Kube clients do.

Here it depends on what you mean by "response handling". In my mind the Cruise Control client has to do what you are saying, so using an HTTP client to send a request, get a response and from that building a POJO. At the same time this is what the current `CruiseControlApiImpl` is doing AFAICS. It sends an HTTP request, get the response and put the received JSON into a generic `CruiseControlResponse` POJO (or into a more specific `CruiseControlRebalanceResponse` POJO, depending on the call). So it's mapping the HTTP response body into a POJO as you say. Then the upper level, the `KafkaRebalanceAssemblyOperator` is taking care of extracting from the POJO what it needs. I don't see in the client any rebalancing logic.

> The response handling is specific to the functionality we are implementing (the same response could be processed in different ways), so it should be the responsibility of the caller (CruiseControlApi in CO, ReplicasChangeHandler in TO).

This is where I don't agree. What does it mean `the same response could be processed in different ways` ... building/filling a different POJO for the same HTTP response? I don't agree. Building/filling part of a POJO depending on which part is calling the client? I still don't agree.

As described above, which seems to be what we have today (that could be improved of course), the Cruise Control client should build a POJO from the HTTP response and this POJO is handled by topic operator or cluster operator (rebalance operator). Each of them will get what they need from the POJO. But it will be always the same POJO for both because it's mapped on a corresponding HTTP response.

Example: the user task endpoint returns an HTTP response. It should be mapped to a unique UserTask POJO usable by both topic and cluster operator and not having two different handlers as you say to fill different POJOs.

scholzj, line 10: I do not see any reason for this to use Enum. You should probably just use a regular String constant.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/resource/cruisecontrol/CruiseControlApi.java
```diff
@@ -20,11 +20,6 @@ public interface CruiseControlApi {
      */
     String CC_REST_API_PROGRESS_KEY = "progress";
 
-    /**
-     * User ID header key
-     */
-    String CC_REST_API_USER_ID_HEADER = "User-Task-ID";
```
scholzj, line 26: Are the above string not headers and some other keys instead?

scholzj (commented):
I left some nits. I do not see this PR partiularly controversial. But I wonder if the overall idea should have some higher level discussion or proposal:
* While doing smaller steps is good, it is hard to review the PRs when not understanding the end state
* It seems it could clarify what and how should such client work and what should it do
* I wonder if the shared CruiseControl client should really live in `operator-common` instead of a separate module or maybe even a seprate subproject.

operator-common/src/main/java/io/strimzi/operator/common/model/cruisecontrol/CruiseControlHeaders.java
fvaleri: > In my mind the Cruise Control client has to do what you are saying, so using an HTTP client to send a request, get a response and from that building a POJO. At the same time this is what the current CruiseControlApiImpl is doing AFAICS.

Not exactly. It's not a one to one mapping between CC response fields and POJO properties, but a custom mapping in a format that is suitable for the rebalance operator. 

As an example, the user_tasks response has many fields, but we only have two of them in the CruiseControlResponse (userTaskId and originalResponse). Additionally, the orginalResponse is parsed into a JsonObject, which causes extra delay that may not be needed by other uses cases.

> What does it mean the same response could be processed in different ways ... building/filling a different POJO for the same HTTP response? I don't agree. Building/filling part of a POJO depending on which part is calling the client? I still don't agree.

Absolutely not. Each response has it's own POJO, which contains all the fields we get from the server. Then the caller decides which fields to use, depending on the use case. Of course, we have some optional request flags that determines the presence or not of some extra content. This will be reflected in the client API.

> As described above, which seems to be what we have today (that could be improved of course), the Cruise Control client should build a POJO from the HTTP response and this POJO is handled by topic operator or cluster operator (rebalance operator). Each of them will get what they need from the POJO. But it will be always the same POJO for both because it's mapped on a corresponding HTTP response. 

> Example: the user task endpoint returns an HTTP response. It should be mapped to a unique UserTask POJO usable by both topic and cluster operator and not having two different handlers as you say to fill different POJOs.

That's basically what I'm proposing here. The current client is centered around Vertx, so it requires a rewrite.

fvaleri:
@scholzj thanks for the review.

This started as a simple refactoring, but I now agree on the necessity of a formal proposal before moving on with sharing the CC client with the cluster operator, and I will create one. 

Current changes are mostly confined to the topic operator, so it would still be good to have them for better modularization.

> I wonder if the shared CruiseControl client should really live in operator-common instead of a separate module or maybe even a seprate subproject.

That's an interesting question. A CruiseControl client could probably be useful in other contexts, so it may make sense to create a separate subproject. What other people think about that?

scholzj:
> Current changes are mostly confined to the topic operator, so it would still be good to have them for better modularization.

TBH, I left some comments, but I personally do not find this particular PR somehow controversial. So I do not have problem to proceed with this one right now. My comment was meant more in general.

operator-common/src/main/java/io/strimzi/operator/common/model/cruisecontrol/CruiseControlHeaders.java
ppatierno: > Not exactly. It's not a one to one mapping between CC response fields and POJO properties, but a custom mapping in a format that is suitable for the rebalance operator.

We didn't add all the fields because we didn't need them. It's just a matter of extending with more fields the topic operator would need. Doing a full mapping with fields that won't be used doesn't really make sense imho. We could add other fields in the POJO in the future.

> In my proposal, each response has it's own POJO, which contains all the fields we get from the server. Then the caller decides which fields to use, depending on the use case. Of course, we have some optional request flags that determines the presence or not of some extra content. This will be reflected in the client API.

As said we can avoid having everything but the minimum needed by the different components who use the Cruise Control API. Anyway, it's different from what you said or I understood before: having a different mapping to a different POJO from the same HTTP response, depending on who is going to use that POJO. That was what I didn't like. 

So AFAICS, the `CruiseControlApi` is already doing what we need. It could be improved (removing Vert.x as you said) but it's already mapping HTTP response to POJO where POJO doesn't mean exactly having all fields but bringing a JSON and the component using the API gets the fields it needs from the JSON itself. You could even have a strong typed POJO if you want instead of bringing POJO. 

Anyway I think that this change/refactoring needs a proposal.

fvaleri: I created this enum mainly to share this header definition between CO and TO, but there could be a need to share other headers (not sure about that now, as I only worked on sharing user_tasks as a proof of concept). I can easily revert if you think it is not necessary now.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/resource/cruisecontrol/CruiseControlApi.java
fvaleri: Yes. They are not headers, but field names used by Vertx JsonObject to extract content from the CC response. They wouldn't be needed with the new client, as you would get a POJO with getters.

operator-common/src/main/java/io/strimzi/operator/common/model/cruisecontrol/CruiseControlHeaders.java
ppatierno: If we want somehow takes this PR but, as I mentioned we would need a proposal for the new client imho, I would remove this enum because it sounds unnecessary.

cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/resource/cruisecontrol/CruiseControlApi.java
scholzj: Ok, great. Thanks.

operator-common/src/main/java/io/strimzi/operator/common/model/cruisecontrol/CruiseControlHeaders.java
scholzj: > I created this enum mainly to share this header definition between CO and TO, but there could be a need to share other headers (not sure about that now, as I only worked on sharing user_tasks as a proof of concept). I can easily revert if you think it is not necessary now.

Sorry if I was not clear. I'm not questioning moving it into a shared class. It is the Enum type I'm questioning. The way it is used, it does not seem to me like Enum is the right type and like it adds any value. I would just make it a String constant (`static final String`).

fvaleri: You mean the long term goal of sharing the CC client right? I will raise a proposal for that. Instead, what do you think about this PR?

I explained the motivation around creating the Headers enum [here](https://github.com/strimzi/strimzi-kafka-operator/pull/10092#discussion_r1601290240), but I can revert that change if you think it's not worth.

fvaleri: > Anyway I think that this change/refactoring needs a proposal.

Agree. Thanks.

fvaleri: Ok, let me switch to a string constant.

fvaleri: Done.

scholzj (approved):
LGTM. Should this be also reviewed by @tombentley as the TO SME?

operator-common/src/main/java/io/strimzi/operator/common/model/cruisecontrol/CruiseControlHeaders.java
```diff
@@ -0,0 +1,15 @@
...
+
+/**
+ * Cruise Control headers.
+ */
+public class CruiseControlHeaders {
+    /**
+     * User task id.
+     */
+    public static final String USER_TASK_ID_HEADER = "User-Task-ID";
+}
```
ppatierno, line 15: whatever it's a string within a class or an enum, I am not still convinced we need a dedicated class for this. How many headers do CC have for the REST API? AFAIU, just this one? More?
Tbh I would not remove it from the CruiseControlApi and not touch it until we have a clear plan where we want to go with the new client.

scholzj: You are the Cruise Control SME maintainer here ... so I think your _vote_ should matter more than mine 😉 

But just to make sure ... are you aware the same package has already many other Cruise Control constants defined there: https://github.com/strimzi/strimzi-kafka-operator/tree/main/operator-common/src/main/java/io/strimzi/operator/common/model/cruisecontrol? I personally would be against this if it was only this one constant living in the `operator-common` module. But as have already have many of them there, it seems like it matters lot less to me.

ppatierno: Yeah fair enough.

ppatierno (approved):
LGTM. Approving because it doesn't have any big impact on what we have as CC client within the cluster operator. But to move to a new one, we should discuss everything in a proposal.

topic-operator/src/main/java/io/strimzi/operator/topic/ReplicasChangeHandler.java
```diff
@@ -4,84 +4,73 @@
...
-     * Create a new replicas change client instance.
+     * Create a new replicas change handler instance.
      * 
      * @param config Topic Operator configuration.
      */
     public ReplicasChangeHandler(TopicOperatorConfig config) {
-        this.config = config;
-        this.httpClientExecutor = Executors.newCachedThreadPool();
-        this.mapper = new ObjectMapper();
+        this.apiClient = new CruiseControlClientImpl(
```
tombentley, line 53: Do you want clients of the `CruiseControlClient` interface to have a direct dependency on (and therefore knowledge of) the impl, or should the interface have a static factory for getting an instance of the default impl?

topic-operator/src/main/java/io/strimzi/operator/topic/ReplicasChangeHandler.java
```diff
@@ -4,84 +4,73 @@
...
  */
 public class ReplicasChangeHandler {
     private static final ReconciliationLogger LOGGER = ReconciliationLogger.create(ReplicasChangeHandler.class);
-    private static final String USER_TASK_ID_HEADER = "User-Task-ID";
-    private static final long REQUEST_TIMEOUT_SEC = 60;
     
-    private final TopicOperatorConfig config;
-    private final ExecutorService httpClientExecutor;
-    private final ObjectMapper mapper;
+    private final CruiseControlClient apiClient;
```
tombentley, line 45: `ccClient` is better imo, because that tells you what it's a client _of_.

topic-operator/src/main/java/io/strimzi/operator/topic/cruisecontrol/CruiseControlClient.java
```diff
@@ -0,0 +1,158 @@
...
+    long HTTP_REQUEST_TIMEOUT_SEC = 60;
+
+    /**
+     * Send a POST request to {@code topic_configuration} endpoint.
+     * This can be used to request replication factor changes.
+     * 
+     * @param kafkaTopics List of Kafka topics.
+     * @return The user task id.
+     */
+    String topicConfiguration(List<KafkaTopic> kafkaTopics);
```
tombentley, line 34: 1. Do you really want to couple this interface to `KafkaTopic` like this? Or if this operation is just about RF change (and just happens to be implemented via the topic_configuration endpoint) why not take a `Map<String, Integer>` parameter?

topic-operator/src/main/java/io/strimzi/operator/topic/cruisecontrol/CruiseControlClient.java
```diff
@@ -0,0 +1,158 @@
...
+
+    /**
+     * Send a GET request to {@code user_tasks} endpoint.
+     * This can be used to check the asynchronous task execution result.
+     * 
+     * @param userTaskIds List of user task ids.
+     * @return User tasks response.
+     */
+    UserTasksResponse userTasks(List<String> userTaskIds);
```
tombentley, line 43: This contract is weird. The choice of collection in particular.. It requires a `List` (ordered, dupes allowed), but returns something containing a `List`, but the contract doesn't provide any guarantees about whether/how their orders are correlated. Either:

* The contract should guarantee that the returned List has the same number and ordering as the parameter
* Or the method should return something which contains a `Map<String, UserTask>`, allowing the called to pick out things based on the taskId, in which case this parameter can be relaxed to be a `Collection`.

topic-operator/src/main/java/io/strimzi/operator/topic/cruisecontrol/CruiseControlClient.java
```diff
@@ -0,0 +1,158 @@
...
+     * @return The error message.
+     */
+    Optional<String> errorMessage(HttpResponse<String> response);
+    
+    /**
+     * Topic names grouped by replication factor value.
+     * 
+     * @param topicByReplicationFactor Topic names grouped by replication factor value.
+     */
+    record ReplicationFactor(@JsonProperty("topic_by_replication_factor") Map<Integer, String> topicByReplicationFactor) { }
```
tombentley, line 58: `Map<Integer, String>` but the the value is some topic names (plural)? How does that work? Are the names separated by some character (if so, you should define it!). Why not `Map<Integer, Collection<String>>`?

topic-operator/src/main/java/io/strimzi/operator/topic/cruisecontrol/CruiseControlClient.java
```diff
@@ -0,0 +1,158 @@
...
+            this.value = value;
+        }
+
+        @Override
+        public String toString() {
+            return value;
+        }
+
+        /**
+         * Use this instead of values() to avoid creating a new array each time.
```
tombentley, line 137: Is that actually a problem?

topic-operator/src/main/java/io/strimzi/operator/topic/cruisecontrol/UrlBuilder.java
```diff
@@ -0,0 +1,88 @@
...
+import java.io.UnsupportedEncodingException;
+import java.net.URI;
+import java.util.List;
+
+import static java.lang.String.format;
+
+/**
+ * Cruise Control URL builder.
+ */
+public class UrlBuilder {
```
tombentley, line 19: Does this need to to be `public`?

topic-operator/src/main/java/io/strimzi/operator/topic/ReplicasChangeHandler.java
fvaleri: I just forgot to rename. Thanks for catching this.

topic-operator/src/main/java/io/strimzi/operator/topic/ReplicasChangeHandler.java
fvaleri: The static factory is better. Simple and clean. Thanks.

topic-operator/src/main/java/io/strimzi/operator/topic/cruisecontrol/CruiseControlClient.java
fvaleri: Well, the endpoint is defined as generic. Right now it only supports RF changes, but the goal is to support all topic configurations.

https://github.com/linkedin/cruise-control/wiki/REST-APIs#change-kafka-topic-configuration

IMO, KafkaTopic makes sense if we stay in the context of the Strimzi operators. Instead, we would need something different if we decide to have a general purpose CC client project. If you agree, I would leave it as it is right now, and we can discuss this point in a proposal about sharing the CC client.

topic-operator/src/main/java/io/strimzi/operator/topic/cruisecontrol/CruiseControlClient.java
fvaleri: Looking at CC code, I can say that ordering is preserved and duplicates are filtered out. 

https://github.com/linkedin/cruise-control/blob/2.5.137/cruise-control/src/main/java/com/linkedin/kafka/cruisecontrol/servlet/response/UserTaskState.java#L87-L113

Does ordering really matter here? Or is it just a problem with the contract? If it's the second, I'm switched back to using a Set as input parameter, which was my original choice.

fvaleri:
@tombentley are you ok with the current state? I would need this to complete some other TO changes.

Keep in mind that I'll create a new proposal for sharing the CC client with the CO where we can discuss the API in more details.

topic-operator/src/main/java/io/strimzi/operator/topic/ReplicasChangeHandler.java
```diff
@@ -144,130 +121,40 @@ public List<ReconcilableTopic> requestOngoingChanges(List<ReconcilableTopic> rec
...
-    }
-
-    private void maybeAddBasicAuthHeader(HttpRequest.Builder builder) {
-        if (config.cruiseControlAuthEnabled()) {
-            String apiUsername = new String(getFileContent(config.cruiseControlApiUserPath()), UTF_8);
-            String apiPassword = new String(getFileContent(config.cruiseControlApiPassPath()), UTF_8);
-            builder.header("Authorization", buildBasicAuthValue(apiUsername, apiPassword));
+            return Files.readAllBytes(Path.of(filePath));
+        } catch (IOException ioe) {
+            throw new RuntimeException(format("File not found: %s", filePath));
```
tombentley, line 157: `java.io.UncheckedIOException` would be more specific

topic-operator/src/main/java/io/strimzi/operator/topic/cruisecontrol/CruiseControlClient.java
tombentley: > Yes. In order to support batch requests, we send a JSON payload where, for each RF value, we have a simple regex like `topic1|topic2|topic3` (i.e. a group by operation).

This needs to be in the javadoc.

topic-operator/src/main/java/io/strimzi/operator/topic/cruisecontrol/CruiseControlClient.java
```diff
@@ -0,0 +1,192 @@
...
+     * @param requestURL Request URL.
+     * @param userTaskId User task id.
+     * @param startMs Start time in ms.
+     */
+    record UserTask(
+        @JsonProperty("Status") String status,
+        @JsonProperty("ClientIdentity") String clientIdentity,
+        @JsonProperty("RequestURL") String requestURL,
+        @JsonProperty("UserTaskId") String userTaskId,
+        @JsonProperty("StartMs") String startMs
```
tombentley, line 115: Why is this a string, it looks like it ought to be a `int`/`long`?

topic-operator/src/main/java/io/strimzi/operator/topic/cruisecontrol/CruiseControlClient.java
```diff
@@ -0,0 +1,192 @@
...
+     * @param status Status.
+     * @param clientIdentity Client identity.
+     * @param requestURL Request URL.
+     * @param userTaskId User task id.
+     * @param startMs Start time in ms.
+     */
+    record UserTask(
+        @JsonProperty("Status") String status,
+        @JsonProperty("ClientIdentity") String clientIdentity,
+        @JsonProperty("RequestURL") String requestURL,
```
tombentley, line 113: `java.net.URI` would be more typesafe.

topic-operator/src/main/java/io/strimzi/operator/topic/cruisecontrol/CruiseControlClient.java
tombentley: I think that unless you know it's a perf problem it might be better to leave this out and have more readable code at the call site.

tombentley (commented):
Hi @fvaleri, this is looking pretty good, but I had a few more questions. Thanks!

topic-operator/src/main/java/io/strimzi/operator/topic/cruisecontrol/CruiseControlClient.java
fvaleri: Yeah, but the problem here is that it only returns the path and query string.

fvaleri:
@tombentley should be complete now, thanks for your suggestions.

tombentley (approved):
Thanks @fvaleri

Issue #9921, opened 2024-04-04
[ST] Remove checks for Kafka version higher than 3.7.0 once 3.6.x versions are removed


im-konge:
Currently, there are few places where we are checking if the Kafka version is >= 3.7.0 (to skip test or to change some values).
Once the Kafka version 3.6.x support is removed, we should also remove these checks.

(In order to remove all checks, search for the link to this issue)

PR #10093, opened 2024-05-09
[system test] - fix testJbodMetadataLogRelocation for >= 3.7.0 Kafka … 

see-quick:
…version

### Type of change

- Bugfix

### Description

This PR fixes a problem when we run a release pipeline with the 3.6.2 version this test case failed because it is not intended to run on 3.6.2 because multiple JBOD volumes are not supported (expected). Thus I have updated the assumption condition here.

Should be cherry-picked to `release-0.41.x` :) if possible

### Checklist

- [x] Write tests

### DIFF
--- a/systemtest/src/test/java/io/strimzi/systemtest/rollingupdate/AlternativeReconcileTriggersST.java
+++ b/systemtest/src/test/java/io/strimzi/systemtest/rollingupdate/AlternativeReconcileTriggersST.java
@@ -484,7 +484,7 @@ class AlternativeReconcileTriggersST extends AbstractST {
      */
     @ParallelNamespaceTest
     void testJbodMetadataLogRelocation() {
-        assumeTrue(Environment.isKRaftModeEnabled());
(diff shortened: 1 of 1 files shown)


im-konge (approved):

scholzj (commented):
should we have the same comment / TODO as for the other tests?

```
        // JBOD storage in KRaft is supported only from Kafka 3.7.0 and higher.
        // So we want to run this test when KRaft is disabled or when it is with KRaft and Kafka 3.7.0+
        // TODO: remove once support for 3.6.x is removed - https://github.com/strimzi/strimzi-kafka-operator/issues/9921
```

see-quick:
> should we have the same comment / TODO as for the other tests?
> 
> ```
>         // JBOD storage in KRaft is supported only from Kafka 3.7.0 and higher.
>         // So we want to run this test when KRaft is disabled or when it is with KRaft and Kafka 3.7.0+
>         // TODO: remove once support for 3.6.x is removed - https://github.com/strimzi/strimzi-kafka-operator/issues/9921
> ```

agree :) will do

scholzj (approved):

PR #10097, opened 2024-05-10
docs(refactor): moves update content to separate section

PaulRMellor:
**Documentation**

Doc refactor and edit: Moves the content related to maintenance and rolling updates to a separate section
Part of the process to move content out of the _Managing_ chapter.

Edits: Condenses the maintenance time windows content into a single section and removes three files as part of the rewrite

### Checklist

- [ ] Write tests
- [x] Update documentation
- [ ] Update CHANGELOG.md


documentation/assemblies/deploying/assembly-rolling-updates.adoc
```diff
@@ -3,9 +3,9 @@
 // using/assembly-management-tasks.adoc
 
 [id='assembly-rolling-updates-{context}']
-= Starting rolling updates of Kafka and other operands using annotations
-
-Strimzi supports the use of annotations to manually trigger a rolling update of Kafka and other operands through the Cluster Operator.
+= Managing rolling updates 
+Maintenance time windows allow you to schedule rolling updates of your Kafka and ZooKeeper clusters to start at a convenient time.
```
scholzj, line 7: I think this is misleading as it applies only to the rolling updates triggering completely automatically by the operator - which is basically only certificate renewal. All other rolling udpates are tirggered by user actiona and do not use the maintenance window. So I would remove it from here or explain it in detail.

documentation/modules/configuring/con-maintenance-time-window-definition.adoc
```diff
@@ -3,23 +3,41 @@
...
-= Maintenance time window definition
+= Scheduling maintenance time windows
 
-You configure maintenance time windows by entering an array of strings in the `Kafka.spec.maintenanceTimeWindows` property.
-Each string is a {CronExpression} interpreted as being in UTC (Coordinated Universal Time, which for practical purposes is the same as Greenwich Mean Time).
+[role="_abstract"]
+Schedule updates by the Cluster Operator to Kafka or ZooKeeper clusters for minimal impact on client applications.
 
-The following example configures a single maintenance time window that starts at midnight and ends at 01:59am (UTC), on Sundays, Mondays, Tuesdays, Wednesdays, and Thursdays:
+Updates are usually triggered by changes to the `Kafka` resource.
```
scholzj, line 11: Can we make it clear it clear that these changes are either done by the user or user's tooling?

ppatierno:
LGTM, apart Jakub's comments on which I agree. I will review again when they are addressed.

documentation/assemblies/deploying/assembly-rolling-updates.adoc
PaulRMellor: I did think about moving it to sit with the content on certificate renewal and validity periods, but I thought it might be a bit restrictive. I've moved now and updated some of the content in the section around it. (I think the TLS chapter in general could be refactored a bit more, but that's for another time)

scholzj (approved):

PR #10098, opened 2024-05-12
Improve the role change reversal and properly revert them

scholzj:
### Type of change

- Bugfix

### Description

The current way how the reversal of the role change when a broker role is removed from a node that is not empty is implemented as simply adding the broker role back. But that is not the reversal in case the role is changed from `broker` to `controller`. That leads the operator into a strange _in the middle_ state where the controller role is applied but the broker role is not removed. This PR fixes it by reverting the roles to the original state before this change and maintaining the current state.

### Checklist

- [x] Write tests

### DIFF
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaClusterCreator.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaClusterCreator.java
@@ -249,7 +249,7 @@ public class KafkaClusterCreator {
                         newNodePools.add(
                                 new KafkaNodePoolBuilder(nodePool)
                                         .editSpec()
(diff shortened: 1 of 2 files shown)


ppatierno:
> This PR fixes it by reverting the roles to the original state before this change and maintaining the current state.

@scholzj maybe I haven't got the gist of this but I can't see where this fix is happening. I see the PR having some warning and log changes + test but not a fix somewhere in the code. I am sure there is something I am missing, could you elaborate a little bit more where I should look at, please?

ppatierno:
> @ppatierno In KafkaClusterCreator, line 289 ... addToRoles -> withRoles

:-o ... How could I have missed that??

ppatierno (approved):
LGTM.

PR #10106, opened 2024-05-14
Promote the UseKRaft feature gate to GA

scholzj:
### Type of change

- Task

### Description

In line with the [Strimzi Proposal 62](https://github.com/strimzi/proposals/blob/main/062-UseKRaft-feature-gate-promotion.md#proposed-timeline), this PR moves the `UseKRaft` to GA. It removes it from the code and keeps it permanently enabled.

_Note: This does not mean KRaft is used by default as it still needs to be enabled with the `strimzi.io/kraft: enabled` feature gate. There is just no feature gate anymore._

There are two tasks that might be done in separate PRs as follow-ups:
* The KRaft based ST in the `FeatureGatesST` class seems useful, but not sure it makes sense to keep it in the Feature Gates tests
* The documentation might use further improvements to make KRaft moe prominent also outside of the feature gate description as the list of limitations for example might get a bit lost now.

### Checklist

- [x] Update documentation
- [x] Update CHANGELOG.md

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 ## 0.42.0
 
 * The `UseKRaft` feature gate moves to GA stage and is permanently enabled without the possibility to disable it.
-  To use the KRaft (ZooKeeper-less Apache Kafka), you still need to use the `strimzi.io/kraft: enabled` annotation on the `Kafka` custom resources.
+  To use KRaft (ZooKeeper-less Apache Kafka), you still need to use the `strimzi.io/kraft: enabled` annotation on the `Kafka` custom resources or migrate from an existing ZooKeeper-based cluster using the `strimzi.io/kraft: migration` annotation.
 * Enhance `KafkaBridge` resource with consumer inactivity timeout and HTTP consumer/producer enablement.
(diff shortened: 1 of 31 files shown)


CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ## 0.42.0
 
+* The `UseKRaft` feature gate moves to GA stage and is permanently enabled without the possibility to disable it.
+  To use the KRaft (ZooKeeper-less Apache Kafka), you still need to use the `strimzi.io/kraft: enabled` annotation on the `Kafka` custom resources.
```
ppatierno, line 6: what do you think if we also add something like " ... or migrating from an existing ZooKeeper-based cluster by using strimzi.io/kraft: migration annotation"?

scholzj: Sure, that sounds reasonable.

ppatierno (approved):
LGTM.

im-konge (approved):
LGTM thanks

PaulRMellor (approved):
Thanks. Looks good. Just a couple of minor suggestions.

PR #10123, opened 2024-05-16
Refactor handing and preparatin of trusted certificates

scholzj:
### Type of change

- Refactoring

### Description

Currently, we configure trusted certificates in several places:
* For Kafka clients in Connect, Bridge, MM2, ...
* For OPA authorization
* For OAuth authentication and Keycloak authorization

But in each place we handle them a bit differently. This PR unifies that to use the same code and mechanism in all those places. This is in preparation for implementation of the [Strimz Proposal 72](https://github.com/strimzi/proposals/blob/main/073-improve-handling-of-CA-renewals-and-replacements-in-client-based-operands.md) -> after this refactoring, the implementation will be easier and will need to be done only in one place.

This is also accompanied by the https://github.com/strimzi/strimzi-kafka-bridge/pull/901 PR in the Bridge repo.

### Checklist

- [x] Write tests

### DIFF
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/model/CertUtils.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/model/CertUtils.java
@@ -220,7 +220,7 @@ public class CertUtils {
      * @param prefix            Prefix used to generate the volume name
      */
     private static void maybeAddTrustedCertificateVolume(List<Volume> volumeList, CertSecretSource certSecretSource, boolean isOpenShift, String prefix) {
-        String volumeName = prefix != null ? prefix + '-' + certSecretSource.getSecretName() : certSecretSource.getSecretName();
+        String volumeName = trustedCertificateVolumeName(certSecretSource, prefix);
 
         // skipping if a volume with same name was already added
         if (volumeList.stream().noneMatch(v -> v.getName().equals(volumeName))) {
@@ -267,12 +267,24 @@ public class CertUtils {
      * @param prefix                Prefix used to generate the volume name
      */
     private static void maybeAddTrustedCertificateVolumeMount(List<VolumeMount> volumeMountList, CertSecretSource certSecretSource, String tlsVolumeMountPath, String prefix) {
-        String volumeName = prefix != null ? prefix + '-' + certSecretSource.getSecretName() : certSecretSource.getSecretName();
+        String volumeName = trustedCertificateVolumeName(certSecretSource, prefix);
 
         // skipping if a volume mount with same Secret name was already added
         if (volumeMountList.stream().noneMatch(vm -> vm.getName().equals(volumeName))) {
             volumeMountList.add(VolumeUtils.createVolumeMount(volumeName, tlsVolumeMountPath + certSecretSource.getSecretName()));
         }
+    }
+
+    /**
+     * Generates the volume name that is used in the Volume definition and in the Volume mounts
+     *
+     * @param certSecretSource      Represents a certificate inside a Secret
+     * @param prefix                Prefix used to generate the volume name
+
+     * @return  The generated volume name
+     */
+    private static String trustedCertificateVolumeName(CertSecretSource certSecretSource, String prefix)    {
+        return prefix != null ? prefix + '-' + certSecretSource.getSecretName() : certSecretSource.getSecretName();
     }
 
     /**
--- a/cluster-operator/src/main/java/io/strimzi/operator/cluster/model/CertUtils.java
+++ b/cluster-operator/src/main/java/io/strimzi/operator/cluster/model/CertUtils.java
@@ -211,7 +211,7 @@ public class CertUtils {
 
     /**
      * Creates the Volume used for trusted certificates and adds it to the list. It checks if the Secret where the
-     * certificate is stored has been already used or not and makes sure to mount it only once to avid duplicates and
+     * certificate is stored has been already used or not and makes sure to mount it only once to avoid duplicates and
      * naming conflicts.
      *
      * @param volumeList        List where the volume will be added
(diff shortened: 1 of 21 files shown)


cluster-operator/src/main/java/io/strimzi/operator/cluster/model/CertUtils.java
```diff
@@ -177,4 +180,122 @@ public static boolean doExistingCertificatesDiffer(Secret current, Secret desire
...
+        if (trustedCertificates != null && !trustedCertificates.isEmpty()) {
+            for (CertSecretSource certSecretSource : trustedCertificates) {
+                maybeAddTrustedCertificateVolume(volumeList, certSecretSource, isOpenShift, prefix);
+            }
+        }
+    }
+
+    /**
+     * Creates the Volume used for trusted certificates and adds it to the list. It checks if the Secret where the
+     * certificate is stored has been already used or not and makes sure to mount it only once to avid duplicates and
```
ppatierno, line 214:
```suggestion
     * certificate is stored has been already used or not and makes sure to mount it only once to avoid duplicates and
```

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/CertUtils.java
```diff
@@ -177,4 +180,122 @@ public static boolean doExistingCertificatesDiffer(Secret current, Secret desire
...
+        if (trustedCertificates != null && !trustedCertificates.isEmpty()) {
+            for (CertSecretSource certSecretSource : trustedCertificates) {
+                maybeAddTrustedCertificateVolumeMount(volumeMountList, certSecretSource, tlsVolumeMountPath, prefix);
+            }
+        }
+    }
+
+    /**
+     * Creates the VolumeMount used for trusted TLS certificates and adds it to the list. It checks if the Secret where
+     * the certificate is stored has been already used or not and makes sure to mount it only once to avid duplicates
```
ppatierno, line 261:
```suggestion
     * the certificate is stored has been already used or not and makes sure to mount it only once to avoid duplicates
```

ppatierno (approved):
LGTM. Just a couple of typos.

docker-images/kafka-based/kafka/scripts/tls_utils.sh
```diff
@@ -55,4 +55,30 @@ function find_ca {
...
+# Parameters:
+# $1: Path to the new truststore
+# $2: Truststore password
+# $3: Base path where the certificates are mounted
+# $4: Environment variable defining the certs that should be loaded
+function prepare_truststore {
+    STORE=$1
+    PASSWORD=$2
+    BASEPATH=$3
+    TRUSTED_CERTS=$4
```
ppatierno, line 69: Sorry one thing which came while reviewing the corresponding bridge PR, just about naming. If we know this function is about "truststore", it's clear `STORE` as name, but then why `TRUSTED_CERTS` instead of just `CERTS`. I have no strong opinion but I would go with `STORE` and `CERTS` or `TRUSTED_STORE` and `TRUSTED_CERTS`. Wdyt?

scholzj: (I can rename the `STORE` to `TRUSTSTORE` if you want, I used `STORE` as we seem to use it everywhere else, so I just copied it).

scholzj: I renamed the `STORE` here as well as in the other PR to `TRUSTSTORE`. Please let me know if it works for you and if it is more consistent.

ppatierno: Yeah Thanks Jakub.

cluster-operator/src/main/java/io/strimzi/operator/cluster/model/CertUtils.java
```diff
@@ -177,4 +180,122 @@ public static boolean doExistingCertificatesDiffer(Secret current, Secret desire
...
+     * the certificate is stored has been already used or not and makes sure to mount it only once to avoid duplicates
+     * and naming conflicts.
+     *
+     * @param volumeMountList       List where the volume mount will be added
+     * @param certSecretSource      Represents a certificate inside a Secret
+     * @param tlsVolumeMountPath    Path where the TLS certs should be mounted
+     * @param prefix                Prefix used to generate the volume name
+     */
+    private static void maybeAddTrustedCertificateVolumeMount(List<VolumeMount> volumeMountList, CertSecretSource certSecretSource, String tlsVolumeMountPath, String prefix) {
+        String volumeName = prefix != null ? prefix + '-' + certSecretSource.getSecretName() : certSecretSource.getSecretName();
```
katheris, line 270: Should this line here be extracted to a common method with the adding of the volumes? Since the names need to match (as far as I understand it), won't that mean if we want to change the way we create the volume mount name we would also have to change the volume name?

It may be this will be changed in the future PR that this refactor is preparing for anyway, so not worth doing

cluster-operator/src/test/java/io/strimzi/operator/cluster/model/CertUtilsTest.java
```diff
@@ -106,4 +113,132 @@ public void testExistingCertificatesDiffer()   {
         assertThat(CertUtils.doExistingCertificatesDiffer(defaultSecret, changedScaleUpSecret), is(true));
         assertThat(CertUtils.doExistingCertificatesDiffer(defaultSecret, changedScaleDownSecret), is(true));
     }
+
```
katheris, line 116: Should there be a test for adding multiple volumes/volume mounts with the same secret name?

katheris (approved):
LGTM, just had a couple of questions/suggestions

cluster-operator/src/test/java/io/strimzi/operator/cluster/model/CertUtilsTest.java
scholzj: @katheris The tests do have the same Secret twice -> the `first-certificate` Secret is used in both certificates. So that seems to be covered? Or maybe I did not understand what exactly you meant?

katheris: Ah I somehow missed that, my mistake!

PR #10126, opened 2024-05-17
Use UBI9 instead of UBI8 as a base image

scholzj:
### Type of change

- Refactoring

### Description

As discussed on the community call on 16.5., this PR updates the base image used in Strimzi from UBI8 to UBI9. This updates only the images based on this repo. Separate PRs will be opened for Bridge and Drain Cleaner if this works.

### Checklist

- [x] Update CHANGELOG.md

### DIFF
--- a/docker-images/base/Dockerfile
+++ b/docker-images/base/Dockerfile
@@ -1,4 +1,4 @@
-FROM registry.access.redhat.com/ubi8/ubi-minimal:latest
+FROM registry.access.redhat.com/ubi9/ubi-minimal:latest
 
 LABEL org.opencontainers.image.source='https://github.com/strimzi/strimzi-kafka-operator'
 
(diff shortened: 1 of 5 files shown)


docker-images/base/Dockerfile
```diff
@@ -9,7 +9,7 @@ ARG TARGETARCH
 USER root
 
 RUN microdnf update \
```
fvaleri, line 11: Should we also add `-y` here to avoid the warning?

fvaleri: Unlike dnf, looks like microdnf does not require this flag by default, as it was designed to build containers. Still there is some build tool may raise a warning, even if it shouldn't. I thought you added the flag because of that, or?

scholzj: I added the `-y` flags to places where the build on UBI9 got stuck without them waiting for a user configuration (they were not needed in UBI8). But I never saw any warnings related to that and they were not my motivation for the flags.

fvaleri: Ok, makes sense. Is it not required for the update command?

scholzj: No, doesn't seem to.

fvaleri: Ok, that's consistent with dnf behavior.

Frawless (approved):

im-konge (approved):

ppatierno (approved):

see-quick (approved):

im-konge (approved):

